Issue #3782, opened 2015-05-21
PostgreSQL: sequelize sync hangs within transaction when ENUM field exists


tornillo:
TYPE queries are executed outside of the transaction.

```javascript
var User = sequelize.define('User', {
	type: Sequelize.ENUM('a', 'b')
});

sequelize.transaction()
.then(function (tx) {
	return sequelize.sync({
		force: true,
		transaction: tx
	})
})
.then(function () {
	tx.commit();
});
```

SQL log:
```SQL
Executing (814aa406-47f7-405e-be52-620304464fdc): START TRANSACTION;
Executing (814aa406-47f7-405e-be52-620304464fdc): SET SESSION TRANSACTION ISOLATION LEVEL REPEATABLE READ;
Executing (814aa406-47f7-405e-be52-620304464fdc): SET autocommit = 1;
Executing (814aa406-47f7-405e-be52-620304464fdc): DROP TABLE IF EXISTS "Users" CASCADE;
Executing (default): DROP TYPE IF EXISTS "enum_Users_type";
<-- hangs here with "default"
```

ns3777k:
https://github.com/sequelize/sequelize/blob/master/lib/query-interface.js#L221
only logging & raw params are passed to the query method to drop enum. transaction's missing.
i tried to to pass the transaction either and it went ok but then the error popped up while recreating the table:

```
Executing (b5a9a593-d462-4b4d-8a8c-2cc180325f3e): START TRANSACTION;
Executing (b5a9a593-d462-4b4d-8a8c-2cc180325f3e): SET SESSION TRANSACTION ISOLATION LEVEL REPEATABLE READ;
Executing (b5a9a593-d462-4b4d-8a8c-2cc180325f3e): SET autocommit = 1;
Executing (b5a9a593-d462-4b4d-8a8c-2cc180325f3e): DROP TABLE IF EXISTS "test" CASCADE;
Executing (b5a9a593-d462-4b4d-8a8c-2cc180325f3e): DROP TYPE IF EXISTS "enum_test_type";
Executing (b5a9a593-d462-4b4d-8a8c-2cc180325f3e): DROP TABLE IF EXISTS "test" CASCADE;
Executing (b5a9a593-d462-4b4d-8a8c-2cc180325f3e): DROP TYPE IF EXISTS "enum_test_type";
[...]

PR #3934, opened 2015-06-14
pass transaction dealing with pg enums

ns3777k:
issue https://github.com/sequelize/sequelize/issues/3782 .probably need to add some tests.


mickhansen:
This is fine for a straight merge. Just as a note, only PG supports transactions for DDL statements.

PR #3939, opened 2015-06-15
.equals() should compare dataValues directly

MrVoltz:
Avoids changes created by other's getters

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -993,12 +993,15 @@ Instance.prototype.decrement = function(fields, options) {
  */
 Instance.prototype.equals = function(other) {
   var result = true;
+  
+  if(!other || !other.dataValues)
+    return false;
 
   Utils._.each(this.dataValues, function(value, key) {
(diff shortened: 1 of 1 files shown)


mickhansen:
How are you facing this problem? getters are ideally pure and side-effect free so for two instances with the same values the getters should return the same values aswell.

MrVoltz:
You are comparing non-transformed value with transformed value, so they can be different. You need to either compare transformed with transformed or non-transformed with non-transformed. I patched the function to compare non-transformed with non-transformed.

mickhansen:
It's doing a key lookup on both, so it should be comparing between transformed values.

lib/instance.js
```diff
@@ -993,12 +993,15 @@ Instance.prototype.decrement = function(fields, options) {
  */
 Instance.prototype.equals = function(other) {
   var result = true;
+  
+  if(!other || !other.dataValues)
+    return false;
```
mickhansen: Style: Either isngle line or with brackets, also needs a space after the if.

PR #3941, opened 2015-06-16
Refactored away duplicate code in Instance::get()

Znarkus:
Instance::get() recursively calls itself if getting an array.

```
$ make test

  1387 passing (2m)
  2 pending
```


mickhansen:
There were perf reasons for this IIRC.

pmhpereira:
IMHO, even if there is a minor performance reduction (the difference is only one additional recursive function call), using duplicated code should not be the "correct" way to solve it.

Also, in 2 out of the 3 if-branches, recursion is already being used.

mickhansen:
Yeah it's likely a minor difference if any.

PR #3943, opened 2015-06-16
Fix spelling error in docs --> api --> model.md

sunshinewyin:
WAS: "refererred"
NOW: "referred"


janmeier:
Please change code comments instead

Issue #3850, opened 2015-06-02
Feature Request: Implement removing hooks


ljowen:
Say I add a hook at runtime:

```javascript
model.addHook('afterUpdate', 'myHook', function(instance, cb) {
     //Do stuff
})      
```

It would be great to be able to remove it:

```javascript
 model.removeHook('afterUpdate', 'myHook')
```

ns3777k:
Basically this comes to my mind as

```javascript
/**
   * Remove hook from the model
   *
   * @param {String} hookType
   * @param {String} name
   */
  removeHook: function(hookType, name) {
    hookType = hookAliases[hookType] || hookType;

    if (this.hasHook(hookType)) {
      return this;
    }
    
    this.options.hooks[hookType] = this.options.hooks[hookType].filter(function (hook) {
      // don't remove unnamed hooks
      if (typeof hook === 'function') {
        return true;
      }

      return typeof hook === 'object' && hook.name !== name;
    });

    return this;
  },
```
I'm not totally sure if that would be enough, but works with the provided example. Will double check that tomorrow.
It can hurt performance if there is a lot of hooks in a hookType, so maybe it'd better to make a hash that contains position of a models' hooks? For example (lil pseudo):

```javascript
// addHook:

var hookPosition = this.options.hooks[hookType].push(!!name ? {name: name, fn: fn} : fn);
if !!name && this.options._hookHash[hookType][name] = hookPosition;

// removeHook:

var hookPosition = this.options. _hookHash[hookType][name];
delete this.options.hooks[hookType][hookPosition];
delete this.options. _hookHash[hookType][name];
```

what y'all think?

ljowen:
Awesome work, thanks!

PR #3947, opened 2015-06-16
hooks removing

ns3777k:
issue https://github.com/sequelize/sequelize/issues/3850

Hope I didn't miss anything important :-)

It might hurt performance if there is a lot of hooks in a hookType (doubt anyone can put there so many of 'em :-) ), so maybe it'd better to make a hash that contains position of a models' hooks. M?

There's one more option that's kinda lazy. Calling removeHook we can just mark it to be removed later in runHooks method.


janmeier:
Nice work! Merged in https://github.com/sequelize/sequelize/commit/008312999ffa7e2ee462162573b32e7c60e7f9f7

Issue #3901, opened 2015-06-09
Creating afterUpdate hook using the mixin function causes TypeError


curiousdannii:
```
db.Version.addHook( 'afterUpdate', 'update_CurrentVersion', update_CurrentVersion );
db.Version.afterUpdate( 'update_CurrentVersion', update_CurrentVersion );
```

The first runs correctly, the second causes this error:

```
Unhandled rejection TypeError: Cannot read property 'length' of undefined
    at /home/dannii/i7el/node_modules/sequelize/lib/hooks.js:125:27
From previous event:
    at Function.each (/home/dannii/i7el/node_modules/sequelize/lib/promise.js:21:17)
    at Hooks.runHooks (/home/dannii/i7el/node_modules/sequelize/lib/hooks.js:120:27)
    at self.sequelize.getQueryInterface.apply.then.tap.then.options.fields.forEach.result._previousDataValues.(anonymous function) (/home/dannii/i7el/node_modules/sequelize/lib/instance.js:672:31)
From previous event:
    at Promise.tap (/home/dannii/i7el/node_modules/sequelize/lib/promise.js:21:17)
    at null.<anonymous> (/home/dannii/i7el/node_modules/sequelize/lib/instance.js:669:10)
From previous event:
    at Promise.then (/home/dannii/i7el/node_modules/sequelize/lib/promise.js:21:17)
    at Instance.save (/home/dannii/i7el/node_modules/sequelize/lib/instance.js:558:6)
    at Object.<anonymous> (/home/dannii/i7el/web/controllers/extensions.js:175:16)
From previous event:
    at Promise.spread (/home/dannii/i7el/node_modules/sequelize/lib/promise.js:21:17)
    at create (/home/dannii/i7el/web/controllers/extensions.js:143:5)
    at Layer.handle [as handle_request] (/home/dannii/i7el/node_modules/express/lib/router/layer.js:82:5)
    at next (/home/dannii/i7el/node_modules/express/lib/router/route.js:110:13)
    at onFinish (/home/dannii/i7el/node_modules/multer/index.js:211:16)
    at Busboy.<anonymous> (/home/dannii/i7el/node_modules/multer/index.js:188:9)
    at Busboy.emit (events.js:117:20)
    at Busboy.emit (/home/dannii/i7el/node_modules/multer/node_modules/busboy/lib/main.js:31:35)
    at /home/dannii/i7el/node_modules/multer/node_modules/busboy/lib/types/multipart.js:52:13
[...]

PR #3959, opened 2015-06-17
add optional name arg to .hook function

ns3777k:
issue https://github.com/sequelize/sequelize/issues/3901

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@
 - [FIXED] $or/$and inside a where clause always expects the input to be an array [#3767](https://github.com/sequelize/sequelize/issues/3767)
 - [ADDED] Unique constraints may now include custom error messages
 - [ADDED] It's possible now to remove a hook by name
+- [ADDED] Hook name can be passed via the direct method [#3901](https://github.com/sequelize/sequelize/issues/3901)
 
 # 3.2.0
 - [FEATURE] Add support for new option `targetKey` in a belongs-to relationship for situations where the target key is not the id field.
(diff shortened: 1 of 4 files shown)


janmeier:
It looks to me like this change makes the name required? That should not be the case. 

Also needs a test

ns3777k:
nope.
```js
Model.prototype[hook] = function(name, callback) {
  return this.addHook(hook, name, callback);
};
```

it goes to addHook

```js
addHook: function(hookType, name, fn) {
  if (typeof name === 'function') {
    fn = name;
    name = null;
}
```

Yup. I'll add a test

ns3777k:
Added several tests.

janmeier:
@ns3777k Looks good.

Wow, this hooks test file is just getting bigger and bigger :D. Perhaps hooks are a good candidate for conversion to unit tests @mickhansen ?
Right now we have integration tests that ensures that every type of hook is being called for the relevant functions. I think this could be written more succinctly as a unit test (taking destroy as an example):

```
Stub model.findAll to return some mock data (so that individualHooks are run)

model.destroy({ hooks: true, individualHooks: true })

expect model.runHooks to have been called with beforeBulkDestroy, options
expect model.runHooks to have been called with beforeDestroy, instance, options

expect model.runHooks to have been called with afterDestroy, instance, options
expect model.runHooks to have been called with afterBulkDestroy, options
```

And then another test which ensures that `runHooks` works with both named and unnamed hooks  - which is basically what all the new tests in this PR do ;)

janmeier:
Thanks for this @ns3777k, and sorry for the long wait!

Since my last post I've rewritten large parts of the hooks test (:fire:) and created unit tests instead. I rewrote your tests to https://github.com/sequelize/sequelize/blob/master/test/unit/hooks.test.js#L209-L233

ns3777k:
wanted to do the same but waited for confirmation :-)
good job anyway)

PR #3964, opened 2015-06-18
handle hasOne onDelete cascade when no associate

jrmurad:
hasOne actually means "has zero or one".

The current behavior of {onDelete:'cascade',hooks:true} looks for an associate and fails to execute ${non-existent instance}.destroy().

This patch checks for the non-existence and aborts gracefully.

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -651,6 +651,10 @@ QueryInterface.prototype.delete = function(instance, tableName, identifier, opti
       if (!Array.isArray(instances)) instances = [instances];
 
       return Promise.each(instances, function (instance) {
+        if (!instance) {
+          return Promise.resolve();
+        }
+
         return instance.destroy({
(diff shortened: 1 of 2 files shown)


janmeier:
Looks good @jrmurad - needs a test case :)

lib/query-interface.js
```diff
@@ -651,6 +651,10 @@ QueryInterface.prototype.delete = function(instance, tableName, identifier, opti
       if (!Array.isArray(instances)) instances = [instances];
 
       return Promise.each(instances, function (instance) {
+        if (!instance) {
+          return Promise.resolve();
+        }
+
```
mickhansen: The no-op could probably be a few lines up?

jrmurad:
Input incorporated, thanks. postgres unit+integration tests passing on my machine.

janmeier:
Thanks :)

Issue #3284, opened 2015-03-04
MSSQL insert/update/delete issues when triggers are enabled


shakdoesgithub:
When there is a trigger enabled on a table, the query that is generated is not correct.

For example:

we have a table contactNotes and a trigger on it, when doing an update the query is generated as 

`INSERT INTO [contactnotes] ([CreateDate],[DeleteDate],[LastUpdated],[ContactID],[Notes]) OUTPUT INSERTED.* VALUES ('2012-09-09T23:15:00.000Z','2012-09-09T23:15:00.000Z','2012-09-09T23:15:00.000Z','3097674','test notes');
`

The error that comes back is:
The target table 'contactnotes' of the DML statement cannot have any enabled triggers if the statement contains an OUTPUT clause without INTO clause.

The link: http://blogs.msdn.com/b/sqlprogrammability/archive/2008/07/11/update-with-output-clause-triggers-and-sqlmoreresults.aspx says we need to have the output into a table parameter

One work around was to set the output:false of returnvalues in the dialects file for mssql, but this broke the insert capability.

I think the only way to fix it is by adding the correct syntax that is mentioned in the link above

mickhansen:
@mbroadst

mbroadst:
@shakdoesgithub what version of mssql are you running?

shakdoesgithub:
@mbroadst we are using msSQL 2012

shakdoesgithub:
@mbroadst, I was able to hack a fix together for two queries: Insert and Update.

The queries are modified to include a variable table based on the model.  This only happens if the dialect has output for returnvalues.  The fix checks the column types and ignores virtual types.  It converts NVARCHAR columns to NVARCHAR(MAX), this is a hack for our project.  

Here's the gist file: https://gist.github.com/shakdoesgithub/8942e635f50bed09a3e4



Changes insert query:
https://gist.github.com/shakdoesgithub/8942e635f50bed09a3e4#file-query-generator-js-L172-L186
https://gist.github.com/shakdoesgithub/8942e635f50bed09a3e4#file-query-generator-js-L208-L238
https://gist.github.com/shakdoesgithub/8942e635f50bed09a3e4#file-query-generator-js-L289-L297

Changes for update query:
https://gist.github.com/shakdoesgithub/8942e635f50bed09a3e4#file-query-generator-js-L289-L297
https://gist.github.com/shakdoesgithub/8942e635f50bed09a3e4#file-query-generator-js-L357-L386
[...]

Issue #3967, opened 2015-06-18
MSSQL ENUM VARCHAR(10) default too short


cdmcnamara:
I'm seeing the following error when attempting to persist a record:
`SequelizeDatabaseError: String or binary data would be truncated.`

Upon investigating it would appear as this is related to a column with DataType.ENUM and the translation sync() is making when it actually creates the table:

```
IF OBJECT_ID('[transactions]', 'U') IS NULL
  CREATE TABLE [transactions] (
    [transaction_id] INTEGER IDENTITY (1, 1),
    [type]                  VARCHAR(10)   NULL CHECK (type IN
                                                      ('Payment', 'OfferCreate', 'OfferCancel', 'TrustSet', 'AccountSet', 'SetRegularKey')),
    [result]                NVARCHAR(255) NOT NULL,
    [created_at]            DATETIME2     NOT NULL,
    [updated_at]            DATETIME2     NOT NULL,
    PRIMARY KEY ([transaction_id])
  );
```

The type column is a VARCHAR(10) but the values of the types exceed 10 characters. It looks like the logic that handles the ENUM translation for the sync can be found here: https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/query-generator.js#L357

It seems like it would make sense for this to default to the STRING length of 255 or potenially the length of the longest value in the ENUM array.

janmeier:
Yeah, I think setting this to 255 would be a more sane default - PR welcome, shouldn't require to much change :)

PR #3968, opened 2015-06-18
adds support for table triggers in mssql dialect

shakdoesgithub:
closes https://github.com/sequelize/sequelize/issues/3284

adds a temp table to get the output rows of insert,update and delete
queries when there is a trigger on a table.
Sits behind a flag called hasTrigger on the model
Added tests that cover insert,update and delete queries

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -173,16 +173,24 @@ var QueryGenerator = {
   insertQuery: function(table, valueHash, modelAttributes, options) {
     options = options || {};
 
+
+
     var query
-      , valueQuery = 'INSERT<%= ignore %> INTO <%= table %> (<%= attributes %>)<%= output %> VALUES (<%= values %>)'
-      , emptyQuery = 'INSERT<%= ignore %> INTO <%= table %><%= output %>'
+      , valueQuery = '<%= tmpTable %>INSERT<%= ignore %> INTO <%= table %> (<%= attributes %>)<%= output %> VALUES (<%= values %>)'
+      , emptyQuery = '<%= tmpTable %>INSERT<%= ignore %> INTO <%= table %><%= output %>'
       , outputFragment
       , fields = []
       , values = []
       , key
       , value
       , identityWrapperRequired = false
-      , modelAttributeMap = {};
+      , modelAttributeMap = {}
+      , tmpTable = ''         //tmpTable declaration for trigger
+      , selectFromTmp = ''    //Select statement for trigger
+      , tmpColumns = ''       //Columns for temp table for trigger
+      , outputColumns = ''    //Columns to capture into temp table for trigger
+      , attribute             //Model attribute holder
+      , modelKey;             //key for model
 
     if (modelAttributes) {
       Utils._.each(modelAttributes, function(attribute, key) {
@@ -205,6 +213,35 @@ var QueryGenerator = {
         emptyQuery += ' RETURNING *';
       } else if (!!this._dialect.supports.returnValues.output) {
         outputFragment = ' OUTPUT INSERTED.*';
+
+        //To capture output rows when there is a trigger on MSSQL DB
+        if (modelAttributes && options.hasTrigger && this._dialect.supports.tmpTableTrigger) {
+            tmpTable = 'declare @tmp table (<%= columns %>); ';
+
+            for (modelKey in modelAttributes){
+              attribute = modelAttributes[modelKey];
+              if(!(attribute.type instanceof DataTypes.VIRTUAL)){
+                if (tmpColumns.length > 0){
+                  tmpColumns += ',';
+                  outputColumns += ',';
+                }
+
+                tmpColumns += this.quoteIdentifier(attribute.field) + ' ' + attribute.type.toSql();
+                outputColumns += 'INSERTED.' + this.quoteIdentifier(attribute.field);
+              }
+            }
+
+            var replacement ={
+              columns : tmpColumns
+            };
+
+            tmpTable = Utils._.template(tmpTable)(replacement).trim();
+            outputFragment = ' OUTPUT ' + outputColumns + ' into @tmp';
+            selectFromTmp = ';select * from @tmp';
+
+            valueQuery += selectFromTmp;
+            emptyQuery += selectFromTmp;
+        }
       }
     }
 
@@ -261,7 +298,8 @@ var QueryGenerator = {
       table: this.quoteTable(table),
       attributes: fields.join(','),
       output: outputFragment,
-      values: values.join(',')
+      values: values.join(','),
+      tmpTable: tmpTable
     };
 
     query = (replacements.attributes.length ? valueQuery : emptyQuery) + ';';
@@ -303,9 +341,15 @@ var QueryGenerator = {
     var query
       , values = []
       , outputFragment
-      , modelAttributeMap = {};
+      , modelAttributeMap = {}
+      , tmpTable = ''         //tmpTable declaration for trigger
+      , selectFromTmp = ''    //Select statement for trigger
+      , tmpColumns = ''       //Columns for temp table for trigger
+      , outputColumns = ''    //Columns to capture into temp table for trigger
+      , attribute             //Model attribute holder
+      , modelKey;             //key for model
 
-    query = 'UPDATE <%= table %> SET <%= values %><%= output %> <%= where %>';
+    query = '<%= tmpTable %>UPDATE <%= table %> SET <%= values %><%= output %> <%= where %>';
 
     if (this._dialect.supports['LIMIT ON UPDATE'] && options.limit) {
       query += ' LIMIT ' + this.escape(options.limit) + ' ';
@@ -315,6 +359,34 @@ var QueryGenerator = {
       if (!!this._dialect.supports.returnValues.output) {
         // we always need this for mssql
         outputFragment = ' OUTPUT INSERTED.*';
+
+        //To capture output rows when there is a trigger on MSSQL DB
+        if (attributes && options.hasTrigger && this._dialect.supports.tmpTableTrigger) {
+            tmpTable = 'declare @tmp table (<%= columns %>); ';
+
+            for (modelKey in attributes){
+              attribute = attributes[modelKey];
+              if(!(attribute.type instanceof DataTypes.VIRTUAL)){
+                if (tmpColumns.length > 0){
+                  tmpColumns += ',';
+                  outputColumns += ',';
+                }
+
+                tmpColumns += this.quoteIdentifier(attribute.field) + ' ' + attribute.type.toSql();
+                outputColumns += 'INSERTED.' + this.quoteIdentifier(attribute.field);
+              }
+            }
+
+            var replacement ={
+              columns : tmpColumns
+            };
+
+            tmpTable = Utils._.template(tmpTable)(replacement).trim();
+            outputFragment = ' OUTPUT ' + outputColumns + ' into @tmp';
+            selectFromTmp = ';select * from @tmp';
+
+            query += selectFromTmp;
+        }
       } else if (this._dialect.supports.returnValues && options.returning) {
         query += ' RETURNING *';
       }
@@ -345,7 +417,8 @@ var QueryGenerator = {
       table: this.quoteTable(tableName),
       values: values.join(','),
       output: outputFragment,
-      where: this.whereQuery(where)
+      where: this.whereQuery(where),
+      tmpTable: tmpTable
     };
 
     if (values.length === 0) {
--- a/lib/dialects/mssql/data-types.js
+++ b/lib/dialects/mssql/data-types.js
@@ -153,6 +153,10 @@ var FLOAT = function() {
 };
 util.inherits(FLOAT, BaseTypes.FLOAT);
 
+BaseTypes.ENUM.prototype.toSql = function() {
+  return 'VARCHAR(255)';
+};
+
 module.exports = {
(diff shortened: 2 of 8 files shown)


shakdoesgithub:
this pull request will work for insert/bulkInsert/update and delete queries, but doesn't work for bulkUpdate.  For those to work we need to some more work.  In bulkUpdate we need the instance parameter to gather the model options.

lib/dialects/abstract/query-generator.js
```diff
@@ -205,6 +211,35 @@ var QueryGenerator = {
         emptyQuery += ' RETURNING *';
       } else if (!!this._dialect.supports.returnValues.output) {
         outputFragment = ' OUTPUT INSERTED.*';
+
+        //To capture output rows when there is a trigger on MSSQL DB
+        if (modelAttributes && options.hasTrigger && this._dialect.name === 'mssql') {
```
janmeier: `this._dialect.supports` - defined in `dialect/index`. Probably `this._dialect.supports.tmpTableTrigger` or something similar

lib/dialects/abstract/query-generator.js
```diff
@@ -205,6 +211,35 @@ var QueryGenerator = {
...
+            tmpTableForMSSQL = 'declare @tmp table (<%= columns %>); ';
```
janmeier: No need to add `forMSSQL` to the vars

lib/dialects/mssql/data-types.js
```diff
@@ -153,6 +153,10 @@ var FLOAT = function() {
 };
 util.inherits(FLOAT, BaseTypes.FLOAT);
 
+BaseTypes.ENUM.prototype.toSql = function() {
+  return 'VARCHAR(10)';
```
janmeier: While you're at it, could you make this `VARCHAR(255)` :) https://github.com/sequelize/sequelize/issues/3967

janmeier: There some code for this in mssql query-generator. https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/query-generator.js#L353-L361

I think it fits a lot better here, so could you move the rest of the code into this

lib/query-interface.js
```diff
@@ -497,10 +497,13 @@ QueryInterface.prototype.removeIndex = function(tableName, indexNameOrAttributes
 };
 
 QueryInterface.prototype.insert = function(instance, tableName, values, options) {
+  options = options || {};
+  options.hasTrigger = !!(instance && instance.__options && instance.__options.hasTrigger);
```
janmeier: `instance.__options` will always be true - will `instance` ever be undefined?

test/integration/trigger.test.js
```diff
@@ -0,0 +1,205 @@
+'use strict';
+
+var chai = require('chai')
+  , Sequelize = require('../../index')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../support')
+  , dialect = Support.getTestDialect();
+
+describe(Support.getTestDialectTeaser('Model'), function() {
```
janmeier: You should wrap this whole test file in `if dialect.__supports.tmpTableTrigger` so the test are only run for mssql

janmeier:
Looking good @shakdoesgithub !

Your integration tests would probably fit better as unit tests along the style of https://github.com/sequelize/sequelize/blob/master/test/unit/sql/select.test.js

lib/query-interface.js
mickhansen: use `instance.Model.hasTrigger` if looking for a model option.

shakdoesgithub: instance is undefined in https://github.com/sequelize/sequelize/blob/master/test/integration/query-interface.test.js#L210, the code in my commit handles that gracefully

lib/dialects/mssql/data-types.js
shakdoesgithub: @janmeier, the function attributeToSQL can't be used to get the SQL for the temp table because it adds extra constraints that are not supported in temp tables.  For example, the enum type will add a check constraint to validate the values, but that isn't supported in temp tables.

shakdoesgithub:
@janmeier, added some unit tests as well.

test/integration/trigger.test.js
```diff
@@ -0,0 +1,202 @@
...
+            username: {
+              type: Sequelize.STRING,
+              field:'user_name'
+            }
+          },{
+            hasTrigger:true
+          });
+
+          return User.sync({force: true}).then(function () {
+            if (dialect === 'mssql') {
```
janmeier: You can remove this, since only mssql supports  `tmpTableTrigger`  :)

lib/dialects/abstract/query-generator.js
```diff
@@ -315,6 +360,34 @@ var QueryGenerator = {
       if (!!this._dialect.supports.returnValues.output) {
         // we always need this for mssql
         outputFragment = ' OUTPUT INSERTED.*';
+
+        //To capture output rows when there is a trigger on MSSQL DB
+        if (attributes && options.hasTrigger && this._dialect.name === 'mssql') {
```
janmeier: `this._dialect.supports.tmpTableTrigger`

lib/dialects/abstract/query-generator.js
```diff
@@ -205,6 +213,36 @@ var QueryGenerator = {
         emptyQuery += ' RETURNING *';
       } else if (!!this._dialect.supports.returnValues.output) {
         outputFragment = ' OUTPUT INSERTED.*';
+
+        //To capture output rows when there is a trigger on MSSQL DB
+        if (modelAttributes && options.hasTrigger && this._dialect.supports.tmpTableTrigger) {
+        // if (modelAttributes && options.hasTrigger && this._dialect.name === 'mssql') {
```
janmeier: :fire:

lib/dialects/mssql/query-generator.js
```diff
@@ -354,7 +354,7 @@ var QueryGenerator = {
       if (attribute.type.values && !attribute.values) attribute.values = attribute.type.values;
 
       // enums are a special case
-      template = 'VARCHAR(10) NULL' /* + (attribute.allowNull ? 'NULL' : 'NOT NULL') */;
+      template = 'VARCHAR(255) NULL' /* + (attribute.allowNull ? 'NULL' : 'NOT NULL') */;
```
janmeier: You can just use toSql here - avoids duplicated logic

lib/dialects/mssql/query-generator.js
```diff
@@ -354,7 +354,7 @@ var QueryGenerator = {
...
       template += ' CHECK (' + attribute.field + ' IN(' + Utils._.map(attribute.values, function(value) {
```
janmeier: Could you move this into the datatype? A funtion called something like `checkConstraint` - so this just becomes

`template = attribute.toSql() + attribute.checkConstraint()`

lib/query-interface.js
janmeier: Aah, true - in that case `instance && instance.Model.hasTrigger` should do the trick here

test/integration/trigger.test.js
```diff
@@ -0,0 +1,202 @@
...
+if(current.dialect.supports.tmpTableTrigger)
+{
+  describe(Support.getTestDialectTeaser('Model'), function() {
+    describe('trigger', function() {
+
+
+        it('should return output rows after insert', function() {
+          var self = this;
+
+          var triggerQuery = 'create trigger User_ChangeTracking on [users] for insert,update, delete \n' +
```
janmeier: Place this inside the innermost describe, so you don't have to duplicate it each time

test/integration/trigger.test.js
```diff
@@ -0,0 +1,202 @@
...
+var chai = require('chai')
+  , Sequelize = require('../../index')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../support')
+  , dialect = Support.getTestDialect()
+  , current = Support.sequelize;
+
+
+if(current.dialect.supports.tmpTableTrigger)
+{
```
janmeier: Bracket on the same line please, with space

test/integration/trigger.test.js
```diff
@@ -0,0 +1,202 @@
...
+          var User = self.sequelize.define('user', {
+            username: {
+              type: Sequelize.STRING,
+              field:'user_name'
+            }
+          },{
+            hasTrigger:true
+          });
+
+          return User.sync({force: true}).then(function () {
```
janmeier: `return User.sync({force: true}).bind(this).then(function () {`

Allows you to use `this` inside the promise chain, and remove `var self

test/integration/trigger.test.js
```diff
@@ -0,0 +1,202 @@
...
+            return User.create({
+              username: 'triggertest'
+            });
+          })
+          .then(function(user){
+            return user.destroy();
+          })
+          .then(function (user) {
+            return User.find({username: 'triggertest'}).then(function(user) {
+              expect(user).to.be.a('null');
```
janmeier: `expect(user).to.be.null`

test/unit/sql/insert.js
```diff
@@ -0,0 +1,39 @@
...
+      });
+
+      var options = {
+          returning : true,
+          hasTrigger : true
+      };
+      expectsql(sql.insertQuery(User.tableName,{user_name: 'triggertest'},User.rawAttributes,options),
+      {
+        mssql:'declare @tmp table ([id] INTEGER,[user_name] NVARCHAR(255));INSERT INTO [users] ([user_name]) OUTPUT INSERTED.[id],INSERTED.[user_name] into @tmp VALUES (\'triggertest\');select * from @tmp;',
+        mysql:"INSERT INTO `users` (`user_name`) VALUES ('triggertest');",
```
janmeier: This will also be covered by the default case, so no need to specify it

test/unit/sql/update.js
```diff
@@ -0,0 +1,39 @@
...
+      });
+
+      var options = {
+          returning : true,
+          hasTrigger : true
+      };
+      expectsql(sql.updateQuery(User.tableName,{user_name: 'triggertest'},{id:2},options,User.rawAttributes),
+      {
+        mssql: 'declare @tmp table ([id] INTEGER,[user_name] NVARCHAR(255));UPDATE [users] SET [user_name]=\'triggertest\' OUTPUT INSERTED.[id],INSERTED.[user_name] into @tmp WHERE [id] = 2;select * from @tmp',
+        mysql:"UPDATE `users` SET `user_name`=\'triggertest\' WHERE `id` = 2",
```
janmeier: Same as above

janmeier:
Unit tests look good @shakdoesgithub ! A couple more comments, mostly just style and readabilty and this will be good to merge

janmeier:
And don't forget to squash some of the smaller, non-descriptive commits into one once you are done :)

lib/dialects/mssql/query-generator.js
shakdoesgithub: If I add the checkConstraint to DataTypes, will it have access to attribute.values? 
Also, the checkConstraint relies on escape function which is in query-generator.js, do you want me to copy that over there? it will duplicate it :(  or move it out to another file and both files can require it in

test/unit/sql/insert.js
shakdoesgithub: I don't follow what you're saying, can you explain a bit more?

janmeier: In this case, the expected sql for mysql is exactly the same as the default. So there is no need to have both

shakdoesgithub: Okay this is done as well. If all looks good, I can squash the commits.

lib/dialects/abstract/query-generator.js
janmeier: Please remove the commented out code

test/integration/trigger.test.js
```diff
@@ -0,0 +1,136 @@
...
+                              'select * from inserted\n' +
+                            'end\n' +
+                            'if exists(select 1 from deleted)\n' +
+                            'begin\n' +
+                              'select * from deleted\n' +
+                            'end\n';
+      it('should return output rows after insert', function() {
+
+
+          var User = this.sequelize.define('user', {
```
janmeier: Can you move initialization of the user model and creating the trigger to a `beforeEach` function? That way we dont have to repeat the same setup for each test. 

Sorry I didn't notice this before

janmeier:
Once my two last comments are addressed this is ready for squash and merge.

shakdoesgithub:
@janmeier, all changes are completed and commits have been squashed

janmeier:
Thanks for the good work, and rapid response to feedback @shakdoesgithub ! :)

I took the liberty of refactoring your test a bit - hope it makes sense.

I also removed the test for bulk insert, since it actually doesn't check anything related to triggers by just doing insert + select.

There might still be an issue when doing `bulkInsert(.. , { returning: true })`, since the bulk insert code is specific to mssql and you only changed in the abstract query-generator. But I though it would be easier to address that in a new PR if you feel like it

shakdoesgithub:
No problem! this was a good learning experience.  

Quick question on the refactor of trigger.test.js:
this line: return expect(User.find({username: 'triggertest'})).to.eventually.be.null;
wouldn't that fail when doing a make dialect ? 
Because JSHint doesn't see that line as a function that can be evaluated.

janmeier:
@shakdoesgithub I believe we have an exception in our .jshintrc file for that case - chai assertions have a pretty weird syntax some times :)

Lichtjaeger:
I've added a PR. #7596

PR #3971, opened 2015-06-19
Update index.md

soundyogi:
The Script is not working because of the lonely "return"


mickhansen:
Actually it totally works. But the return doesn't really have any value so fine with removing it.

soundyogi:
Well yeah okay. Anyway my linter cried :D

thanks

PR #3972, opened 2015-06-19
Fix incorrect paths created for "See:" links in docs generator 

cgrinaldi:
None of the current links referenced in the **See:** sections of the [docs](http://docs.sequelizejs.com/en/latest/) are working. For example, clicking on [Sequelize#sync](http://docs.sequelizejs.com/en/latest/api/api/sequelize#sync) under the [Model documentation](http://docs.sequelizejs.com/en/latest/api/model/#sync) for `sync()`:  
screen shot 2015-06-19 at 9 16 23 am

brings you to to a page not found.

The issue is that the link automatically generated has repeated `/api` (see http://docs.sequelizejs.com/en/latest/api/api/sequelize#sync for an example where `/api` is repeated).

This PR removes the repeated `api/` from `docs-generator.js` and re-creates the associated documentation.

### DIFF
--- a/docs/api/associations.md
+++ b/docs/api/associations.md
@@ -1,6 +1,6 @@
 <a name="mixin"></a>
 # Mixin Mixin
-[View code](https://github.com/sequelize/sequelize/blob/cc8687539fe96f7f64887a04ddf5d48f159f5e92/lib/associations/mixin.js#L95)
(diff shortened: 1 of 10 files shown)


janmeier:
Thanks :)!

PR #3984, opened 2015-06-22
Fixed errors in query

pranavrajs:


janmeier:
Thanks! :)

There's actually still an error in the last one, right now the query reads `// UPDATE post SET updatedAt = null WHERE updatedAt NOT NULL;` - can you fix that as well

pranavrajs:
Yeah! I missed that one . Will fix now :)

janmeier:
Thanks :)

PR #3985, opened 2015-06-22
Add support for multiple databases

ackerdev:
Matt originally wrote these changes for MSSQL multiple catalog support at my request.

It is not 100% functional. Some meta queries may not support multiple databases or may be outright broken.

I cannot further this work myself, so I am offering it in the state that mbroadst left it so that maybe someone can give it the love it needs and help it get integrated, and I can stop maintaining it myself. :smile:

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -20,7 +20,8 @@ var QueryGenerator = {
   options: {},
 
   addSchema: function(param) {
-    var self = this;
+    var self = this
+      , database = (param.options && param.options.database ? param.options.database : undefined);
 
     if (!param.$schema) return param.tableName || param;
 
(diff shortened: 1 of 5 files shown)


janmeier:
Unfortunately no-one took up the task :(

alexraju:
@janmeier  This is a much essential feature. Even though I have separate connection, when I do join between tables from different databases, db name is getting appended before the table name causing query to fail. If you could get some priority to this, It will be great. 

It would be great if you could assign someone to this task.

mickhansen:
@alexraju The someone who have to be either me or @janmeier, since we can't really assign anyone from the community (for a lot of reasons).

We've had very few requests for this so whether or not it's essential is highly debateable.
If it's highly essential to you, you can help contribute (that's how me and @janmeier even got started contributing to Sequelize)

AckerApple:
@alexraju We still use the @ackerdev forked version that has Sequelize multidatabase support. The forked version is 3.13 and the current version is 3.23 . The multi database support works great for all general purpose CRUD functionality. This branch as seen above: https://github.com/ackerdev/sequelize/tree/multiple-databases

PR #3986, opened 2015-06-22
Throw ForeignKeyConstraintError in mssql when trying to set an association that doesn't exist

johngiudici:


mickhansen:
Thank you

PR #3991, opened 2015-06-23
Add test case for scope with includes

ackerdev:
There seem to be some issues preventing use of includes inside scope definitions. These tests should reproduce the problems.

WIP, I'll be abusing the PR system for travis-ci checking.

### DIFF
--- a/test/integration/model/scope/associations.test.js
+++ b/test/integration/model/scope/associations.test.js
@@ -12,6 +12,8 @@ describe(Support.getTestDialectTeaser('Model'), function() {
   describe('scope', function () {
     describe('associations', function () {
       beforeEach(function () {
+        var sequelize = this.sequelize;
+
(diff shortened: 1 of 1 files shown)


janmeier:
The problem was that we didn't call conformOptions (converting the include to the form expected later on) when a scope is a function

test/integration/model/scope/associations.test.js
```diff
@@ -32,10 +34,28 @@ describe(Support.getTestDialectTeaser('Model'), function() {
                 username: 'tony'
               }
             },
+            includeActiveProjects: function(){
+              return {
+                include: [{
+                  model: sequelize.models.Company,
+                  include: [sequelize.models.Project.scope('active')]
```
janmeier: Changing this to `.project` - the models are stored under their name :)

PR #3993, opened 2015-06-23
Update moment-timezone to 0.4.0

josephpage:
moment-timezone is the only package out-dated today : https://david-dm.org/sequelize/sequelize

Changelog : https://github.com/moment/moment-timezone/blob/0.4.0/changelog.md
(for main part new IANA TZDB data)

Thanks !


janmeier:
Thanks :-)

PR #3994, opened 2015-06-23
upsert overwrites data with default values on update

FredKSchott:
Currently, a `defaultValue` set on the model will be used in every upsert operation. This is fine for when upsert INSERTs, since that is when default values are used/needed. However, upsert is also using defaultValue on INSERT, which is causing data to be overwritten and lost by the default values. 

~~I've included a test that is currently broken by this bug, I haven't had time to look into a fix yet.~~ EDIT: fix also added

/cc @mickhansen @janmeier (from the last upsert issue we worked on)

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1706,24 +1706,27 @@ Model.prototype.upsert = function (values, options) {
 
   return instance.hookValidate(options).bind(this).then(function () {
     // Map field names
-    values = Utils.mapValueFieldNames(instance.dataValues, options.fields, this);
+    var updatedDataValues = _.pick(instance.dataValues, Object.keys(instance._changed))
+      , insertValues = Utils.mapValueFieldNames(instance.dataValues, options.fields, this)
+      , updateValues = Utils.mapValueFieldNames(updatedDataValues, options.fields, this)
+      , now = Utils.now(this.sequelize.options.dialect);
 
-    var now = Utils.now(this.sequelize.options.dialect);
-
-    if (createdAtAttr && !values[createdAtAttr]) {
-      values[createdAtAttr] = this.$getDefaultTimestamp(createdAtAttr) || now;
+    // Attach createdAt
+    if (createdAtAttr && !updateValues[createdAtAttr]) {
+      insertValues[createdAtAttr] = this.$getDefaultTimestamp(createdAtAttr) || now;
     }
-    if (updatedAtAttr && !values[updatedAtAttr]) {
-      values[updatedAtAttr] = this.$getDefaultTimestamp(updatedAtAttr) || now;
+    if (updatedAtAttr && !insertValues[updatedAtAttr]) {
+      insertValues[updatedAtAttr] = updateValues[updatedAtAttr] = this.$getDefaultTimestamp(updatedAtAttr) || now;
     }
 
     // Build adds a null value for the primary key, if none was given by the user.
     // We need to remove that because of some Postgres technicalities.
     if (!hadPrimary && this.primaryKeyAttribute && !this.rawAttributes[this.primaryKeyAttribute].defaultValue) {
-      delete values[this.primaryKeyField];
+      delete insertValues[this.primaryKeyField];
+      delete updateValues[this.primaryKeyField];
     }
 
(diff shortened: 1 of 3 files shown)


FredKSchott:
I've updated the PR with a proposed fix. `updateValues` are currently created out of the full INSERT data object, and has no way of knowing which fields are defaults and which should be changed. This PR moves this creation up closer to the model where default value and change information is still known.

test/integration/model/upsert.test.js
```diff
@@ -248,6 +249,41 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         });
       });
 
+      it('does not overwrite createdAt time on update', function() {
+        var originalCreatedAt;
+        return this.User.create({ id: 42, username: 'john'}).bind(this).then(function() {
+          return this.User.findById(42);
+        }).then(function(user) {
+          originalCreatedAt = user.createdAt;
+          return this.sequelize.Promise.delay(1000).bind(this).then(function() {
```
janmeier: You can use sinon fake timers here, to artifically advance time http://sinonjs.org/docs/#clock

janmeier:
LGTM  - just a single comment

test/integration/model/upsert.test.js
FredKSchott: In this situation I think we need the actual delay so that sequelize can internally get the proper Date() times on both create & upsert. I've used sinon to fake timers before, but iirc it has weak/no support for modifying the actual system time which is what we need here.

janmeier: We use `new Date()` internally, and that is exactly what

FredKSchott: I must be misremembering then, I remember a lot of trouble last time I used these for anything other than extending timers. I'll experiment a bit tomorrow and see if I can get it to work.

lib/model.js
```diff
@@ -1706,24 +1706,27 @@ Model.prototype.upsert = function (values, options) {
 
   return instance.hookValidate(options).bind(this).then(function () {
     // Map field names
-    values = Utils.mapValueFieldNames(instance.dataValues, options.fields, this);
-
+    var updatedDataValues = _.pick(instance.dataValues, Object.keys(instance._changed));
+    var insertValues = Utils.mapValueFieldNames(instance.dataValues, options.fields, this);
+    var updateValues = Utils.mapValueFieldNames(updatedDataValues, options.fields, this);
     var now = Utils.now(this.sequelize.options.dialect);
```
mickhansen: Purely style: mind using leading commas with a single `var` statement?

FredKSchott: fasho

test/integration/model/upsert.test.js
FredKSchott: @janmeier definitely misremembered, this worked perfectly!

FredKSchott:
@janmeier @mickhansen comments addressed

janmeier:
Thanks @FredKSchott ! :)

FredKSchott:
Woot! thanks for the quick responses :)
Is there any way we could push a patch in the next few days to get this out? It sucks to wait 2-3 weeks for every change to get upsert working.

janmeier:
If you bug @mickhansen then probably yes :)

In the meantime you can point your package.json to `sequelize/sequelize#3d0af74947b71b2523b7cff873963c506e5a7a75`

FredKSchott:
Yea good point, I'll point for now. Thanks @janmeier!

mickhansen:
Released in `3.3.2`

FredKSchott:
Thanks @mickhansen!

telpalbrox:
Hi, since this release, the upsert function is not working properly.
I am trying to upsert an entry with 2 unique fields. Those fields are included in order entries so when I execute the upsert there is a MySQL error "error: uncaughtException: Cannot read property 'uniqueKeys' of undefined" and it is no catched with the .catch method so the app goes down.

janmeier:
@telpalbrox please open a new thread with some code that shows the error

ciberyo16:
thank you @janmeier I opened it here https://github.com/sequelize/sequelize/issues/4059

Issue #3714, opened 2015-05-13
Map fields back to attributes for SELECT queries


sankethkatta:
Add support for mapping fields on a `SELECT` query back to attributes, before loading into a Model.
This would allow Model definitions with using `field` to use raw queries. Demonstration of the issue #3706. 

Currently the docs show using the `callee` feature of raw queries as:

> A second, optional, argument is the _callee_, which is a model. If you pass a model the returned data will be instances of that model.

>```js
// Callee is the model definition. This allows you to easily map a query to a predefined model
sequelize.query('SELECT * FROM projects', Projects).then(function(projects){
  // Each record will now be a instance of Project
})
```

However, this does not actually work properly with all Model definitions (as shown in #3706).

sankethkatta:
Fixed in https://github.com/sequelize/sequelize/pull/3995.

PR #3995, opened 2015-06-24
Map raw field names to attributes.

sankethkatta:
Add feature #3714. 

This also allows the `returning: true` [option](https://github.com/sequelize/sequelize/blob/2c4a9f3cf9887fb33c31e397e758dd4aa3374d01/lib/model.js#L2064) for `update` in postgres to work properly. Currently it [generates](https://github.com/sequelize/sequelize/blob/e948dc8fdc6be90c54441e091472a0c5d72899ce/lib/dialects/abstract/query-generator.js#L212) out `RETURNING *`, which faces the same problem as described with raw queries in #3714.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 - [FIXED] commit/rollback multiple times on same transaction [#4491](https://github.com/sequelize/sequelize/issues/4491)
 - [FIXED] memory leak / options mangle for scopes with include [#4470](https://github.com/sequelize/sequelize/issues/4470)
 - [FIXED] custom `targetKey` for belongsTo on a target with a primary key will now correctly create foreign key constraints [#4455](https://github.com/sequelize/sequelize/issues/4455)
+- [ADDED] Map raw fields back to attributes names when using `mapToModel` or `returning` [#3995](https://github.com/sequelize/sequelize/pull/3995)
 
 # 3.8.0
 - [ADDED] `version` on `Sequelize` returning the current npm/package.json version [#4459](https://github.com/sequelize/sequelize/pull/4459)
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -548,7 +548,19 @@ AbstractQuery.prototype.isUpdateQuery = function() {
 
 AbstractQuery.prototype.handleSelectQuery = function(results) {
   var result = null;
-
+  // Map raw fields to names if a mapping is provided
+  if (this.options.fieldMap) {
+    var fieldMap = this.options.fieldMap;
+    results = Utils._.map(results, function(result) {
+      return Utils._.reduce(fieldMap, function(result, name, field) {
+        if (result[field] !== undefined) {
+          result[name] = result[field];
+          delete result[field];
+        }
+        return result;
+      }, result);
+    });
+  }
   // Raw queries
   if (this.options.raw) {
     result = results.map(function(result) {
@@ -598,7 +610,6 @@ AbstractQuery.prototype.handleSelectQuery = function(results) {
   if (this.options.plain) {
     result = (result.length === 0) ? null : result[0];
   }
-
   return result;
 };
 
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -640,6 +640,8 @@ Sequelize.prototype.import = function(path) {
  * @param {Function}        [options.logging=false] A function that gets executed while running the query to log the sql.
  * @param {Instance}        [options.instance] A sequelize instance used to build the return instance
  * @param {Model}           [options.model] A sequelize model used to build the returned model instances (used to be called callee)
+ * @param {Object}          [options.mapToModel=false] Map returned fields to model's fields if `options.model` or `options.instance` is present. Mapping will occur before building the model instance.
+ * @param {Object}          [options.fieldMap] Map returned fields to arbitrary names for `SELECT` query type.
  *
  * @return {Promise}
  *
@@ -657,6 +659,11 @@ Sequelize.prototype.query = function(sql, options) {
 
   if (options.instance && !options.model) {
     options.model = options.instance.Model;
+  }
+
+  // Map raw fields to model field names using the `fieldAttributeMap`
+  if (options.model && options.mapToModel && !Utils._.isEmpty(options.model.fieldAttributeMap)) {
+    options.fieldMap =  options.model.fieldAttributeMap;
   }
 
   if (Utils._.isPlainObject(sql)) {
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -389,6 +389,8 @@ var QueryGenerator = {
             query += selectFromTmp;
         }
       } else if (this._dialect.supports.returnValues && options.returning) {
+        // ensure that the return output is properly mapped to model fields.
+        options.mapToModel = true;
         query += ' RETURNING *';
       }
     }
--- a/lib/model.js
+++ b/lib/model.js
@@ -870,6 +870,13 @@ Model.prototype.refreshAttributes = function() {
       delete definition.index;
     }
   });
+  // Create a map of field to attribute names
+  this.fieldAttributeMap = Utils._.reduce(this.fieldRawAttributesMap, function(map, value, key) {
+    if (key !== value.fieldName) {
+      map[key] = value.fieldName;
+    }
+    return map;
+  }, {});
 
   this.uniqueKeys = this.options.uniqueKeys;
 
--- a/test/integration/model/update.test.js
+++ b/test/integration/model/update.test.js
@@ -2,12 +2,20 @@
 
 /* jshint -W030 */
 var Support = require(__dirname + '/../support')
-  , DataTypes = require(__dirname + '/../../../lib/data-types');
+  , DataTypes = require(__dirname + '/../../../lib/data-types')
+  , chai = require('chai')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../support')
+  , current = Support.sequelize
+  , _ = require('lodash');
 
 describe(Support.getTestDialectTeaser('Model'), function() {
   describe('update', function () {
-    it('should only update the passed fields', function () {
-      var Account = this.sequelize.define('Account', {
+
+    var Account;
+
+    beforeEach(function() {
+      Account = this.sequelize.define('Account', {
         ownerId: {
           type: DataTypes.INTEGER,
           allowNull: false,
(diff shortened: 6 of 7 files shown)


mickhansen:
I think it would be better if it happened at a lower level than build so it could be reused for bulkCreate etc.

sankethkatta:
Any suggestions where a better place for it would be? Maybe in the Instance constructor or initValues?

Also, not sure if you have any insight into the error the build system is throwing, I am not able to reproduce the error locally (and it seems unrelated to my changes): 
```
 Error: Expected 0 running queries. 1 queries still running in [POSTGRES-NATIVE] Continuation local storage bluebird shims then rejected
```

mickhansen:
Probably in post query result parsing, let the caller pass a field to attribute map that the query result parser can use perhaps. It's something we need a generalized approach to.

sankethkatta:
I attempted to generalize this a bit more and moved the aliasing into `handleSelectQuery`. It uses an option called `aliasMap` which is passed into `Sequelize.prototype.query`.
Let me know what you think.

lib/dialects/abstract/query.js
```diff
@@ -514,7 +514,19 @@ AbstractQuery.prototype.isUpdateQuery = function() {
...
   var result = null;
-
+  // Map raw field names using a map of aliased fields
+  if (this.options.aliasMap) {
+    var aliasMap = this.options.aliasMap;
+    results = Utils._.map(results, function(result) {
+      return Utils._.reduce(aliasMap, function(result, alias, field) {
+        if (field in result) {
+          result[alias] = result[field];
+          delete result[field];
```
mickhansen: `result[field] = undefined` is a bit quicker, can you check if it produces the same result?

lib/dialects/abstract/query.js
```diff
@@ -514,7 +514,19 @@ AbstractQuery.prototype.isUpdateQuery = function() {
 
 AbstractQuery.prototype.handleSelectQuery = function(results) {
   var result = null;
-
+  // Map raw field names using a map of aliased fields
+  if (this.options.aliasMap) {
+    var aliasMap = this.options.aliasMap;
+    results = Utils._.map(results, function(result) {
+      return Utils._.reduce(aliasMap, function(result, alias, field) {
+        if (field in result) {
```
mickhansen: does this syntax deopt @BridgeAR?

mickhansen:
Looking good. Although i'm not sure about the `aliasMap` name.

lib/dialects/abstract/query.js
sankethkatta: It still passes the tests, however, it leaves remnants in the `dataValues` object.

```js
dataValues: { 
  id: 1,
  username: 'john',
  email_address: undefined,
  createdAt: Sun Jan 01 2012 02:10:10 GMT-0800 (PST),
  updatedAt: Sun Jan 01 2012 02:10:10 GMT-0800 (PST),
  emailAddress: 'john@gmail.com' 
}
```

lib/dialects/abstract/query.js
sankethkatta: I can also use lodash's `_.has()` or `result.hasOwnProperty()`. I used the `in` syntax as I had seen in other places in the Sequelize codebase.

lib/dialects/abstract/query.js
mickhansen: Ah yeah of course, we don't want that.

sankethkatta:
Looks like the other parameter names for `.query()` are single words like `replacements`. How about `aliases` or `mappings`?

mickhansen:
That's even less descriptive though :)

sankethkatta:
I see, more descriptive! How about `rawFieldMap`?

mickhansen:
Maybe just `fieldMap`? I think that would be okay.

sankethkatta:
That sounds good to me, updating it now.

lib/dialects/abstract/query.js
```diff
@@ -514,7 +514,19 @@ AbstractQuery.prototype.isUpdateQuery = function() {
 
 AbstractQuery.prototype.handleSelectQuery = function(results) {
   var result = null;
-
+  // Map raw fields to arbitrary names if a mapping is provided
+  if (this.options.fieldMap) {
+    var fieldMap = this.options.fieldMap;
+    results = Utils._.map(results, function(result) {
+      return Utils._.reduce(fieldMap, function(result, name, field) {
+        if (field in result) {
```
BridgeAR: This should be written as 
```js
result.field !== undefined
```

mickhansen: @BridgeAR `in` causes de-opt?

BridgeAR:
LGTM even tough I think it would be nice to improve this at some point to only include the fields needed from the beginning on. But this is only internal stuff.

mickhansen:
@BridgeAR From the beggining? I'm not quite sure i understand? You mean parse and rewrite a raw query to alias the fields?

BridgeAR:
Yes

lib/dialects/abstract/query.js
BridgeAR: This depends on if any field name in result is somehow special or not. If it is e.g. `Strange / field` it'll deopt (there was an issue with someone using slashing in his field names :D ). Otherwise it should not but checking for undefined will be faster.

sankethkatta:
Just pushed the minor change, checking if the field existed in the result using `undefined` as suggested.

mickhansen:
@BridgeAR That does not sound feasible at all, don't think we'll ever do that ;p Rewriting raw queries would be a bit too magic i think. (Why is the input not my actual input?)

lib/dialects/abstract/query.js
```diff
@@ -514,7 +514,19 @@ AbstractQuery.prototype.isUpdateQuery = function() {
 
 AbstractQuery.prototype.handleSelectQuery = function(results) {
   var result = null;
-
+  // Map raw fields to arbitrary names if a mapping is provided
+  if (this.options.fieldMap) {
```
mickhansen: We should probably put this under the `raw` if so it isn't run on all selects, no need to encur the extra loop when it's not needed.

sankethkatta: Currently the fix also works for postgres `returning: true` on `update`. We'd have to put it behind  `this.options.raw || this.options.type === QueryTypes.BULKUPDATE`.

mickhansen: That sounds like a bug, `update` should not result in a handleSelectQuery call.

sankethkatta: I assumed it did this by design as handling a `UPDATE ... RETURNING *` is similar to a select. Here is the call to `handleSelectQuery` in [postgres/query.js](https://github.com/sequelize/sequelize/blob/eaf197fa7e6cb4d6b52d3d13493e6ae81352fdf8/lib/dialects/postgres/query.js#L296).

mickhansen: Huh, yeah maybe you're right. I'm afraid we can't incur that perf overhead for all select queries since they are already perf sensitive and wouldnt take nicely to another nested loop, please do the large if :)

sankethkatta: Hmm, it looks like `this.options.raw` isn't set if a model is passed in [query.js#L671](https://github.com/sequelize/sequelize/blob/master/lib/sequelize.js#L671). Is there a way to know if the query originated from user input on `sequelize.query` directly without using `options.raw`?

mickhansen: Hmm no, no way to know, that is a bit of a problem since actually in this case you don't want a raw result.

sankethkatta: We could introduce a new option to toggle on or off the auto-mapping to model attributes behavior. Right now we assume we should map fields anytime the model option is set. 
But it would require users writing raw queries to set on the new option in addition to setting the model.

sankethkatta:
@mickhansen Trying another approach to this problem. I replaced `fieldMap` with a boolean `mapToModel` option. The mapping will no longer be on the code path for select queries, but will require users of the raw query api to explicitly set `mapToModel: true` to enable the behavior.

Let me know your thoughts!

janmeier:
@sankethkatta I like this approach, it looks pretty clean to me. I'll let @mickhansen comment on this too before merging but LGTM :)

sankethkatta:
@janmeier @mickhansen any update on this?

janmeier:
@sankethkatta Mick is on vacation, I'll bug him about this when he's back ;)

sankethkatta:
Thanks! :smiley:

sankethkatta:
ping! :bell:

mattppp:
Merge please?

mickhansen:
Sorry i was on a trip again.
I'd prefer if we do as little as coupling as possible in the query generator, so ideally generate the field map in `query` that can be used by `handleSelectQuery` perhaps.

mickhansen:
Hmm, nvm the last comment i just deleted.

sankethkatta:
I don't like that the current approach requires the user to set  the `mapToModel` option. However, given the performance limitations of running unnecessary loops and the fact it is not possible to know whether `sequelize.query` was called directly by the user (which we discussed [pull/3995#discussion_r34085726] (https://github.com/sequelize/sequelize/pull/3995#discussion_r34085726)). This seems like the best compromise, and I'd rather err on the side of explicitness anyways.

Let me know if you can think of a better way!

lib/dialects/abstract/query.js
```diff
@@ -514,7 +514,19 @@ AbstractQuery.prototype.isUpdateQuery = function() {
...
+    results = Utils._.map(results, function(result) {
+      return Utils._.reduce(fieldAttributeMap, function(result, name, field) {
+        if (result[field] !== undefined) {
+          result[name] = result[field];
+          delete result[field];
+        }
+        return result;
+      }, result);
+    });
+  }
```
mickhansen: Sorry for being so late to comment.
Ideally i'd like a generic mapper passed to query, or one generated in query and then used here.

The less the query generator and query handler needs to know about `Model` the better.

sankethkatta: Refactored in latest commit.

changelog.md
```diff
@@ -4,6 +4,7 @@
 - [FIXED] Reload doesn't synchronize a null include [#4353](https://github.com/sequelize/sequelize/issues/4353)
 - [FIXED] commit/rollback multiple times on same transaction [#4491](https://github.com/sequelize/sequelize/issues/4491)
 - [FIXED] memory leak / options mangle for scopes with include [#4470](https://github.com/sequelize/sequelize/issues/4470)
+- [FEATURE] Map raw fields back to attributes when model is built [#3714](https://github.com/sequelize/sequelize/issues/3714)
```
mickhansen: `[FEATURE]` > `[ADDED]`
perhaps: "Map raw fields back to attribute names when using `returning` ..."

lib/sequelize.js
```diff
@@ -659,6 +661,11 @@ Sequelize.prototype.query = function(sql, options) {
     options.model = options.instance.Model;
   }
 
+  // Map raw fields to model field names using the `fieldAttributeMap`
+  if (options.model && options.mapToModel && Utils._.size(options.model.fieldAttributeMap) > 0) {
```
mickhansen: Nitpicking, but could use `!_.empty(..)` i believe?

mickhansen:
Please add API docs to `sequelize.query` describing `fieldMap` and `mapToModel` options.
Excellent work!

(please squash commits to a single descriptive commit when ready aswell)

lib/sequelize.js
sankethkatta: Good call!

sankethkatta:
I addressed the comments and squashed the commits. 
As for the documentation, do I need to write somewhere else in addition to the JSDoc? (which I did on [`lib/sequelize#L643`](https://github.com/sequelize/sequelize/pull/3995/files#diff-24f8b5cb05b7e92726ca732f7d16d5baR643).
Thanks!

mickhansen:
Nope jsdoc works :)

mickhansen:
Thank you for your work!

sankethkatta:
Awesome! Thank you for working through this with me! :smiley:

PR #3828, opened 2015-05-29
Geometry Support

mikeringrose:
This was originally based on https://github.com/sequelize/sequelize/pull/3641, however, I fixed some bugs and hopefully made the code more extensible. If you're happy with the path that I am on, I'll add support for MySQL (and more tests).

PR #3699, opened 2015-05-12
Added support for Geometry in Postgres/PostGIS

Ghirro:
Hi there,

So we use sequelize internally and required Geometry support to some degree. I started adding it in a few days ago before I realised that someone had already done a little bit of the work [https://github.com/sequelize/sequelize/pull/3641]. We were actually moving along the same path (using GeoJSON) and that PR had a few approaches that were better than my own and so I integrated them here.

A couple of things to note, I've had to change one test to run locally. I looked it up and it's a known issue (https://github.com/sequelize/sequelize/issues/2966). I also ran into a few issues with the documentation generation suite but have documented what I've done.

The basics are:
***
#### Added support for DataTypes.GEOMETRY(Type, SRID)
This will be stored as a POSTGIS geometry column but the API will always interacts with [GeoJSON](http://geojson.org/geojson-spec.html)
***
#### Added support for new geo based where lookups:
```js
$geoIn: { GeoJson } // Returns when given object contains row
$geoNotIn: { GeoJson } // Returns when given object doesn't contain row
$geoWithin: { point: { GeoJson }, distance: 10 } // Returns when given object is within distance of point
$geoWithout: { point: { GeoJson }, distance: 10 } // Returns when given object is not within distance of point
```
***

I've also had to slightly change the `dropTables` for postgres as this did not work well with the new tests I've introduced (tried to drop POSTGIS views which are vital).

Anyway apologies for a slightly overlapping pull request but I thought it better to mention it.

PR #4013, opened 2015-06-26
Adding GEOMETRY Support

mikeringrose:
This is to replace PR https://github.com/sequelize/sequelize/pull/3828. Same code added to the latest in master.

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -95,7 +95,7 @@ var QueryGenerator = {
   },
 
   showTablesQuery: function() {
-    return "SELECT table_name FROM information_schema.tables WHERE table_schema = 'public';";
+    return "SELECT table_name FROM information_schema.tables WHERE table_schema = 'public' AND table_type LIKE '%TABLE' AND table_name != 'spatial_ref_sys';";
   },
 
   describeTableQuery: function(tableName, schema) {
@@ -897,6 +897,13 @@ var QueryGenerator = {
       return 'ARRAY[' + value.map(function (v) {
         return SqlString.escape(JSON.stringify(v), false, this.options.timezone, this.dialect, field);
       }, this).join(',') + ']::' + jsonType.key + '[]';
+    } else if (value && field && field.type instanceof DataTypes.GEOMETRY) {
+      var s = "ST_GeomFromGeoJSON('" + JSON.stringify(value) + "'";
+      if (field.type.srid) {
+        s += + ", " + field.type.srid;
+      }
+      s += ")";
+      return s;
     }
 
     return SqlString.escape(value, false, this.options.timezone, this.dialect, field);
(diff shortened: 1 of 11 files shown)


mikeringrose:
Locally `make postgres` passes just fine.

>  1568 passing (2m)
>  2 pending

mikeringrose:
Jan if there's anything else that'd you'd like me to change, let me know. Otherwise I'll begin adding MySQL support in another PR.

janmeier:
@mikeringrose Nope, everything looks good! Looking forward to the next PR :)

lib/dialects/postgres/query-generator.js
```diff
@@ -897,6 +897,13 @@ var QueryGenerator = {
       return 'ARRAY[' + value.map(function (v) {
         return SqlString.escape(JSON.stringify(v), false, this.options.timezone, this.dialect, field);
       }, this).join(',') + ']::' + jsonType.key + '[]';
+    } else if (value && field && field.type instanceof DataTypes.GEOMETRY) {
+      var s = "ST_GeomFromGeoJSON('" + JSON.stringify(value) + "'";
+      if (field.type.srid) {
+        s += + ", " + field.type.srid;
```
krisr: hey @mikeringrose, I think you are either missing another term in this statement, or need to remove the leading "+". I tried creating geometry with DataTypes.GEOMETRY("Point", 4326) and ended up having bad queries generated due to this line.

mikeringrose: Thanks for the heads up. I'll be submitting a fix for this issue tomorrow
to address the following bug -
https://github.com/sequelize/sequelize/issues/4054. If you need things to
work for now, you can leave off the SRID. Thanks again.

Mike Ringrose

mshahriarinia:
What is the syntax for sequelize postgres geometry? Is there a HelloWorld example?

janmeier:
@mshahriarinia We will add documentation for this once all dialects are supported.

Until then Id recommend you to look at the tests in this PR to see the syntax

mshahriarinia:
I am using the latest version on github f38d0cae5cb3aaeb22118e0b612264aa963a2298, and I have postgres+postgis. Quick note for folks: on you DB make sure you run `CREATE EXTENSION postgis;` otherwise you get `Error: "geometry" does not exist` from postgres.

mshahriarinia:
Can I do distance calculation between a lat/lon and a point column in potgres in current state or top k closest points to a given point? I went through all sorts of code, [looks like](https://github.com/sequelize/sequelize/pull/3699) somebody introduced the concept of `$geoWithin`, `$geoIn`, but got deleted from pull request?

janmeier:
@mshahriarinia Unfortunately those operators were not integrated into the final pull request.

They would definitely be nice to have, so a PR to re-add them would be very wellcome! The code should be pretty similar to whats in this PR, just needs to be migrated

mikeringrose:
That'd be awesome if we could get those in. :+1:

On Mon, Jul 13, 2015 at 3:17 AM, Jan Aagaard Meier <notifications@github.com

PR #4014, opened 2015-06-26
fixes problem with order in scope when counting

dickie81:
having order option in model’s scope when performing a count kicked out
this error:

SequelizeDatabaseError: column “xxxx” must appear in the GROUP BY
clause or be used in an aggregate function

### DIFF
--- a/test/integration/model.test.js
+++ b/test/integration/model.test.js
@@ -1670,15 +1670,15 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       var options, aggregateSpy;
 
       beforeEach(function () {
-        options = { where: ['username = ?', 'user1']};
+        options = { where: { username: 'user1'}};
 
         aggregateSpy = sinon.spy(this.User, "aggregate");
       });
 
       afterEach(function () {
-        var optsArg = aggregateSpy.args[0][2];
-
-        expect(optsArg.where).to.deep.equal(['username = ?', 'user1']);
+        expect(aggregateSpy).to.have.been.calledWith(
+          sinon.match.any, sinon.match.any,
+          sinon.match.object.and(sinon.match.has('where', { username: 'user1'})));
 
         aggregateSpy.restore();
       });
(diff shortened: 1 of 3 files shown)


janmeier:
This is not specifically related to scopes.

I guess we can remove the order - I can't come up with a case where you need an order for count (at least not when we already remove the limit) @mickhansen ?

mickhansen:
I don't think `order` makes sense with any kind of aggregate (unless you're selecting other fields, which you definitely aren't in our `count` method).

janmeier:
@dickie81 If you add some general count tests instead this should be ready for merge :)

dickie81:
I have added some extra count tests & also nulled offset as this was causing issues too.

test/integration/model.test.js
```diff
@@ -1665,6 +1666,54 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       });
     });
 
+    describe("options sent to aggregate", function () {
+      var options, aggregateSpy;
+
+      beforeEach(function () {
+        options = { where: ['username = ?', 'user1']};
```
janmeier: `where: { username: 'user1' }`

test/integration/model.test.js
```diff
@@ -1665,6 +1666,54 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
+        });
+      });
+
+      it('modifies option "order" by setting it to null', function() {
+        options.order = "username";
+
+        return this.User.count(options).then(function() {
+          var optsArg = aggregateSpy.args[0][2];
+
+          expect(optsArg.order).to.equal(null);
```
janmeier: `expect(aggregateSpy).to.have.been.calledWith(sinon.match.any, sinon.match.any, sinon.match.object.and.has('order', null))`

Might not exactly be prettier, but I think it convey the assertion better, and it doesn't poke directly into the args array

janmeier:
LGTM, just a couple of small comments on tests

dickie81:
comments appreciated - changes implemented

janmeier:
Thanks! :)

Issue #4015, opened 2015-06-26
result.count for findAndCountAll is incorrect when using includes with where


davidlanger:
When using findAndCountAll the count of all rows is incorrect as Sequelize does just count all the rows which are more than the matching models.

``` javascript
var Foo = s.define('Foo', {})
var Bar = s.define('Bar', {m: DataTypes.STRING(40)});

Foo.hasMany(Bar);
Bar.belongsTo(Foo);

// set up (synchronously displayed here
Foo.bulkCreate([{id: 1}, {id: 2}, {id: 3}, {id: 4}, {id: 5}]);
Bar.bulkCreate([{'FooId': 1, m:'yes'}, {'FooId': 1, m:'yes'}, {'FooId': 1, m: 'no'}, {'FooId': 2, m: 'yes'}]);

return Foo.findAndCountAll({
  include: [{ model: Bar, required: true, where: { m: 'yes' } }],
  limit: 1
});
/// results.count is now 3 instead of 2
```

davidlanger:
Thank you, this is great!

PR #4016, opened 2015-06-26
Add failing test case for findAndCountAll with a include and where

davidlanger:
This is a test case for `findAndCountAll` returning an incorrect value for `result.count`, as the where statement in the include makes the `count()` sql function counts all rows instead (3 in this case) of only matching `Foo`s (2 in this case).

See issue #4015


janmeier:
Fixed by adding a distinct to the count part when there are any include on the count

Issue #4011, opened 2015-06-25
Options object is mutated and cannot be reused when calling model.create


sclement41:
I am calling myModel.create passing in an options object which contains my transaction.

I then try to call mySecondModel.create passing in my options object again, but the first call actually mutated my options object which now contains values for "fields", "instance", "model", "returning", etc.

I was expecting the options object to be cloned before being used internally by Sequelize, so that I can reuse it between calls. The problem code is in Instance.prototype.save, where the code is using the options object without cloning it first.

Aukhan:
@mickhansen sorry to be commenting on a closed issue but I need similar help here.
I can see that the create, update methods are not mutating the options object anymore but I am trying to reuse the same options object in hooks (afterCreate, afterUpdate).
I can see save method here
https://github.com/sequelize/sequelize/blob/3e5b8772ef75169685fc96024366bca9958fee63/lib/instance.js#L592
passes the cloned options object into hooks and it contains values for fields, returning etc just like @sclement41  mentions above. When I pass this options object into other getters or query generating functions inside hooks, it causes a field mismatch and throws errors that such and such columns or not part of the table `ER_BAD_FIELD_ERROR: Unknown column 'my_column_name' in 'field list'`

I feel that hooks should also be getting the original options object. Please advice if this is a bad design decision on my end.

Thanks

PR #4023, opened 2015-06-28
option object shouldn't be updated internally.

smihica:
I got "not-null constraint" error because the previous save(opt) call inserted "fields" into opt.
This may cause confusing errors.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -482,7 +482,7 @@ Instance.prototype.save = function(options) {
   if (arguments.length > 1) {
     throw new Error('The second argument was removed in favor of the options object.');
   }
-  options = _.defaults(options || {}, {
+  options = _.defaults({}, options, {
     hooks: true,
     validate: true
   });
@@ -724,7 +724,7 @@ Instance.prototype.save = function(options) {
 * @return {Promise<this>}
 */
 Instance.prototype.reload = function(options) {
-  options = _.defaults({}, options || {}, {
+  options = _.defaults({}, options, {
     where: this.where(),
     limit: 1,
     include: this.options.include || null
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -809,7 +809,7 @@ Instance.prototype.destroy = function(options) {
   options = Utils._.extend({
     hooks: true,
     force: false
-  }, options || {});
+  }, options);
 
   return Promise.bind(this).then(function() {
     // Run before hook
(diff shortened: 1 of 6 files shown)


smihica:
This pr is for the following bug.
https://github.com/sequelize/sequelize/issues/4011

lib/instance.js
```diff
@@ -477,7 +477,7 @@ Instance.prototype.save = function(options) {
   if (arguments.length > 1) {
     throw new Error('The second argument was removed in favor of the options object.');
   }
-  options = _.defaults(options || {}, {
+  options = _.defaults({}, options || {}, {
```
janmeier: I don't think we need  `options || {}`, `options` should be fine

lib/instance.js
```diff
@@ -818,7 +818,7 @@ Instance.prototype.destroy = function(options) {
       for(var i = 0; i < primaryKeys.length; i++) {
           where[this.Model.rawAttributes[primaryKeys[i]].field] = this.get(primaryKeys[i], { raw: true });
       }
-      return this.sequelize.getQueryInterface().delete(this, this.Model.getTableName(options), where, _.defaults(options, { type: QueryTypes.DELETE,limit: null}));
+      return this.sequelize.getQueryInterface().delete(this, this.Model.getTableName(options), where, _.defaults({}, options, { type: QueryTypes.DELETE,limit: null}));
```
janmeier: Can't we just use `options` here?

janmeier:
Looks good @smihica - do you mind adding a few tests. Preferrably unit tests, since they are faster. Something along the lines of 

```js
stub(sequelize, 'query').returns(Promise.resolve([]));

var options = { ... }

model.reload(options);

expect(options).to.deep.equal({... }
```


If you have too many problems trying to stub sequelize.query, integration tests that hit the db will be fine as well

lib/instance.js
smihica: yeah I think so. thank you.

lib/instance.js
smihica: Oh, exactly.

smihica:
Yes, I'll try to add some tests later. thanks reviewing.

smihica:
@janmeier 
OK. I added tests and rebased. and All unit and integration tests were passed in my environment.

janmeier:
Solid work on the tests @smihica ! Thanks for this, and the other PR also :)

PR #4027, opened 2015-06-29
Fixed a bug that wrong sql generated when destroying in pg.

smihica:
Hi.

When I tried like following code, sequelize generated wrong SQL.

    var sequelize = new Sequelize('db', 'user', 'pass', { host: 'localhost', dialect: 'postgres' });

    var User = sequelize.define('test_user', {
      id: {
        type:       Sequelize.INTEGER,
        primaryKey: true,
        field:      "test_user_id",
      }
    }, { freezeTableName: true, timestamps:false });

    User.create({id: 1}).then(function(u) {
      console.log(u.id);
      u.destroy().then(function(){
        console.log('delete success !!');
      }); // error
    });

    /* Log ---
    Executing (default): INSERT INTO "test_user" ("test_user_id") VALUES (1) RETURNING *;
    1
    Executing (default): DELETE FROM "test_user" WHERE "id" IN (SELECT "id" FROM "test_user" WHERE "test_user_id" = 1 LIMIT 1) // *** WRONG SQL HERE
    Unhandled rejection SequelizeDatabaseError: column "id" does not exist
    */

WIth these changes, sequelize generated collect SQL.

    /* Log ---
    Executing (default): INSERT INTO "test_user" ("test_user_id") VALUES (1) RETURNING *;
    1
    Executing (default): DELETE FROM "test_user" WHERE "test_user_id" IN (SELECT "test_user_id" FROM "test_user" WHERE "test_user_id" = 1 LIMIT 1)
    delete success !!
    */

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -395,7 +395,9 @@ var QueryGenerator = {
     primaryKeys[tableName] = primaryKeys[tableName] || [];
 
     if (!!model && primaryKeys[tableName].length < 1) {
-      primaryKeys[tableName] = Object.keys(model.primaryKeys);
+      primaryKeys[tableName] = Utils._.map(Object.keys(model.primaryKeys), function(k){
+        return model.primaryKeys[k].field;
+      });
     }
 
     if (options.limit) {
--- a/test/unit/sql/delete.test.js
+++ b/test/unit/sql/delete.test.js
@@ -1,0 +1,49 @@
+'use strict';
+
+/* jshint -W110 */
+var Support   = require(__dirname + '/../support')
+  , util = require('util')
+  , expectsql = Support.expectsql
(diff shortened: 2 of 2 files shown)


janmeier:
Looks good - could you add a couple of sql unit tests, along the lines of https://github.com/sequelize/sequelize/blob/master/test/unit/sql/select.test.js

smihica:
I see

BridgeAR:
Ping @smihica do you need any help? :)

smihica:
No. Thanks! @BridgeAR ;)

I'm really busy here recently... (*q*)...
I'll write some tests, but may be late.
If you want to fix this bug immediately, I'm afraid but please write some tests by your self.

smihica:
@janmeier 
@BridgeAR 

Hi.
I added a test for this commit and rebased.
And all unit and integration tests were passed in my environment.

Issue #3920, opened 2015-06-11
Unable to exclude primaryKey while performing an Aggregation (sum) with Join


cludden:
I am attempting to perform an aggregation (sum) with a join, and the closest I've gotten is with the below:

```
        Balance.findAll({
            attributes: [
                [app.sequelize.fn('sum', app.sequelize.col('usdAmount')), 'sum']
            ],
            where: {
                date: {
                    $between: ['2015-01-01', '2015-12-31']
                },
                ledgerId: '04'
            },
            include: [
                {
                    model: Accounts,
                    attributes: [],
                    where: {
                        'ebitdaDesc': 'EBITDA'
                    }
                }
            ]
        }).then(function(result) {
            console.log(result);
            res.status(200).send();
        }, function(err) {
            console.log(err);
            res.status(500).send();
        });
```

which generates the following query:
```
'SELECT [Balances].[id], sum([usdAmount]) AS [sum] FROM [GL].[Balances] AS [Balances] INNER JOIN [GL].[Accounts] AS [Account] ON [Balances].[accountId] = [Account].[id] AND [Account].[ebitdaDesc] = \'EBITDA\' WHERE [Balances].[date] BETWEEN \'2015-01-01\' AND \'2015-12-31\' AND [Balances].[ledgerId] = \'04\';'
```

and the following error:
```
'Column \'GL.Balances.id\' is invalid in the select list because it is not contained in either an aggregate function or the GROUP BY clause.'
```

The query executes as expected without including the "[Balances].[id]" in the select statement, but I have not determined how to exclude this.

Any help would be greatly appreciated!

cludden:
As a follow up,
I was able to correct the query by commenting out the following code:

```
/lib/model.js#1181

        if (options.attributes) {
          if (options.attributes.indexOf(this.primaryKeyAttribute) === -1) {
            options.originalAttributes = options.attributes;
            options.attributes = [this.primaryKeyAttribute].concat(options.attributes);
          }
        }
```

Are there any implications in doing this that I'm unaware of (besides having to manually include the primaryKey when I need it)?

mickhansen:
We add the primary key attribute because we need to be able to run deduplication logic.

mickhansen:
Not quite sure what would be the best solution here.

pimterry:
[...]

PR #4029, opened 2015-06-29
Do not automatically add primary key to attributes (or include[].attributes) if raw is set

pimterry:
This bring in the changes discussed in #3920, disabling the automatic addition of the primary key field to the attributes returned by a query, and thereby allowing aggregation in more cases, when raw is set to true.

This includes the changes and tests. Is there any documentation anywhere that needs to be updated in line with this?

### DIFF
--- a/test/integration/include/findAll.test.js
+++ b/test/integration/include/findAll.test.js
@@ -1957,8 +1957,6 @@ describe(Support.getTestDialectTeaser('Include'), function() {
         expect(posts.length).to.equal(1);
 
         var post = posts[0];
-
-        console.log("** " + JSON.stringify(post));
         expect(post.id).not.to.be.ok;
         expect(parseInt(post["comments.commentCount"], 10)).to.equal(3);
       });
--- a/test/integration/include/findAll.test.js
+++ b/test/integration/include/findAll.test.js
@@ -1920,5 +1920,48 @@ describe(Support.getTestDialectTeaser('Include'), function() {
         expect(parseInt(post.get('commentCount'), 10)).to.equal(3);
       });
     });
+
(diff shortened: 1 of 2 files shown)


mickhansen:
Thank you for this PR. The only issue i have is passing `raw` as a single argument, perhaps we could pass the options instead?

test/integration/include/findAll.test.js
```diff
@@ -1920,5 +1920,48 @@ describe(Support.getTestDialectTeaser('Include'), function() {
...
+            }
+          ],
+          raw: true
+        });
+      }).then(function (posts) {
+        expect(posts.length).to.equal(1);
+
+        var post = posts[0];
+
+        console.log("** " + JSON.stringify(post));
```
mickhansen: Please remove the console.log

pimterry:
Both those comments are now fixed.

Build has failed, but that looks to be for other unrelated reasons (issues with transaction serialization), which I suspect is something intermittent, but I don't have permission to retrigger the build. I can't see any way these couple of tiny markups could've caused that.

mickhansen:
Thanks for the contribution :)

hamza-sabri:
Sorry for opening this again. 
I read the article and I still couldn't find a way to do what the article suggested

I had three tables invoice, item, invoice_item(which had a quantity column)
I am doing a join between invoice and item models and would like to apply some aggregate functions (sum, min, max, count, avg, ... etc) on the result of the join or simply add up the quantity field from the result.


But I had the same error described here (https://www.softwire.com/insights/fixing-aggregation-over-relationships-with-sequelize/) and it was the closest I came to a solution but couldn't figure it out. 

could someone help?

PR #4034, opened 2015-06-29
Fix path variable being surrounded by quotes. Fixes #3464

### LINKED ISSUE #3464: Unique violation returns path property surrounded by quotes
When there's a unique violation, the `ValidationErrorItem` is:

```js
{ message: '"quoteId" must be unique',
     type: 'unique violation',
     path: '"quoteId"',
     value: '24902467' }
```

I suppose that's an error?

BridgeAR:
I just ran into this and thought I'll fix it right quick.


janmeier:
I'll let this one pass without a test, but don't get used to it! :)

PR #4041, opened 2015-07-01
Fix  #4040, do not call .clone method on plain objects or arrays.

### LINKED ISSUE #4040: Sequelize is incompattible with collections
The [collections](https://github.com/montagejs/collections) package includes an `array.clone` shim.

As a result Sequelize dies if you try certain calls. Namely this happens if you do a find with an includes that includes a Sequelize class.

For example:
`Model.find({include: [{model: User, as: 'user'}]})`

Basically:
- Sequelize tries to deep clone the method `opts` (with a customizer that avoids cloning Sequelize classes)
- Sequelize's Utils.cloneDeep encounters the array under `include:` and tests it for a `.clone` method.
- Upon discovering collections' `array.clone` it calls the shim and `array.clone` begins a deep clone of the array and any objects in it.
- Upon encountering the Sequelize instance under `model:` the shim attempts to clone it.
- The shim fails to clone the model and throws an `Error: Can't clone [object SequelizeModel:User]` error breaking the whole program.

This will happen to anyone who happens to require the `collections` package **anywhere** in their application and makes use of an array with a Sequelize class nested in it as options to a Sequelize method.

I understand that collections shouldn't be shimming array with a clone method. However blindly trusting a mysterious `clone` method on any object without even testing it doesn't seem like a reliable code pattern.

dantman:


lib/utils.js
```diff
@@ -96,7 +96,7 @@ var Utils = module.exports = {
   cloneDeep: function(obj, fn) {
     return lodash.cloneDeep(obj, function (elem) {
       // Preserve special data-types like `fn` across clones. _.get() is used for checking up the prototype chain
-      if (elem && typeof elem.clone === 'function') {return elem.clone(); }
+      if (elem && !Array.isArray(elem) && !lodash.isPlainObject(elem) && typeof elem.clone === 'function') {return elem.clone(); }
```
mickhansen: Might read better to have seperate ifs for array and plain object and just return undefined

mickhansen: It definitely can't be return elem since then the elements inside won't be cloned.

test/integration/utils.test.js
```diff
@@ -109,6 +109,43 @@ describe(Support.getTestDialectTeaser('Utils'), function() {
     });
   });
 
+  describe('cloneDeep', function() {
+    it('should clone objects', function() {
+      var obj = {foo: 1},
+          clone = Utils.cloneDeep(obj);
```
mickhansen: We use leading commas.

lib/utils.js
dantman: I thought it would be best to just leave whether plain objects get special handling up to the `fn`.

lib/utils.js
```diff
@@ -96,7 +96,15 @@ var Utils = module.exports = {
...
-      if (elem && typeof elem.clone === 'function') {return elem.clone(); }
+      if (elem) {
+        // Allow objects to implement their own clone method
+        if (typeof elem.clone === 'function') {
+          // However do not mistakenly call clone methods on arrays and plain objects
+          if (!Array.isArray(elem) && !lodash.isPlainObject(elem)) {
+            return elem.clone();
+          }
+        }
+      }
```
mickhansen: What i meant was go back to the old style and then add a short circuit at the top like:

```
if (Array.isArray(elem) || lodash.isPlainObject(elem)) return undefined;
```

janmeier:
I fixed a single comma from trailing to leading manually - thanks! :)

PR #4051, opened 2015-07-02
Update querying.md

Verdier:


janmeier:
Hmm, I was sure I already fixed that, but it must have been lost in a restructing - thaks!

PR #4052, opened 2015-07-02
[ci skip] Note about querying and associations

Verdier:


janmeier:
It's a small start, but we definitely need some more include examples in that chapter - thanks!

+1 for ci skip - there's no need to stress travis more than necesary ;)

PR #4053, opened 2015-07-02
Fix typo in docs on Model.schema() method

RebootJeff:


RebootJeff:
I'm not sure if the typo (`hi`) should be corrected to be `this`, `Model`, or `Instance`. It looks like `this` is technically correct, but it's also the least descriptive choice.

Also, I don't know why the file diff shows that I changed line 697 (because I didn't?).

janmeier:
Thanks for pointing this one

RebootJeff:
Oh wow I forgot that the API docs were auto-generated. That explains the `git diff` for line 697 :stuck_out_tongue_closed_eyes:

Issue #4039, opened 2015-07-01
FindOrCreate with CLS and unmanaged transaction


gutobortolozzo:
Hi. I'm using sequelize with mocha to execute my test specs, so using continuation-local-storage and I'm able to create one transaction before test start, propagate this transaction to all my operations and by the end of test rollback the transaction, my configuration is exposed below.

```javascript
var cls = require('continuation-local-storage');
var namespace = cls.createNamespace('my-namespace');
Sequelize.cls = namespace;
```

```javascript
beforeEach(function(done){
    models.sequelize.transaction().then(function(t){
        models.namespace.set("transaction", t);
        done();
    });
});

afterEach(function(){
    models.namespace.get("transaction").rollback();
});
````

But when I'm using findOrCreate, I always need to explicitly pass my transaction as option parameter, what is not required by other methods. After some changes inside my node_module/sequelize/model.js on method findOrCreate I added this silly snippet of code and the method works as expected.

```javascript
if(this.sequelize.constructor.cls){
    options.transaction = this.sequelize.constructor.cls.get('transaction');
}
```

I did some miss configuration or this is the expected behavior of findOrCreate?

Thanks.

janmeier:
The cls transaction is not added before `sequelize.query`, so I see the issue here.

The fix you posted looks good - add a test and you have yourself a PR :)

gutobortolozzo:
@janmeier I will post the PR late this week.

Thanks :)

PR #4062, opened 2015-07-04
FindOrCreate with CLS not using stored transaction

gutobortolozzo:
https://github.com/sequelize/sequelize/issues/4039

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1641,7 +1641,7 @@ Model.prototype.findOrCreate = function(options) {
     );
   }
 
-  if (this.sequelize.constructor.cls) {
+  if (options.transaction === undefined && this.sequelize.constructor.cls) {
     options.transaction = this.sequelize.constructor.cls.get('transaction');
   }
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -1641,6 +1641,10 @@ Model.prototype.findOrCreate = function(options) {
     );
   }
 
+  if(this.sequelize.constructor.cls){
+    options.transaction = this.sequelize.constructor.cls.get('transaction');
+  }
+
   var self = this
     , internalTransaction = !options.transaction
     , values
--- a/lib/model.js
+++ b/lib/model.js
@@ -1641,7 +1641,7 @@ Model.prototype.findOrCreate = function(options) {
     );
   }
 
-  if(this.sequelize.constructor.cls){
+  if (this.sequelize.constructor.cls) {
     options.transaction = this.sequelize.constructor.cls.get('transaction');
   }
 
--- a/test/unit/model/find-or-create.test.js
+++ b/test/unit/model/find-or-create.test.js
@@ -1,0 +1,51 @@
+'use strict';
+
+/* jshint -W030 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support   = require(__dirname + '/../support')
+  , current   = Support.sequelize
+  , cls = require('continuation-local-storage')
+  , sinon = require('sinon')
+  , stub = sinon.stub
+  , Promise = require('bluebird');
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+
+  describe('method findOrCreate', function () {
+
+    before(function () {
+      current.constructor.cls = cls.createNamespace('sequelize');
+    });
+
+    after(function () {
+      delete current.constructor.cls;
+    });
+
+    beforeEach(function(){
+      this.User = current.define('User', {}, {
+        name: 'John'
+      });
+
+      this.stub = stub(this.User.sequelize, 'transaction');
+      this.stub.returns(new Promise(function(){}));
+    });
+
+    afterEach(function(){
+      this.stub.restore();
+    });
+
+    it('should use transaction from cls if available', function () {
+
+      var options = {
+        where : {
+          name : '123'
+        }
+      };
+
+      this.User.findOrCreate(options);
+
+      expect(options).to.have.any.keys('transaction');
+    });
+  });
+});
--- a/test/unit/model/find-or-create.test.js
+++ b/test/unit/model/find-or-create.test.js
@@ -22,16 +22,16 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       delete current.constructor.cls;
     });
 
-    beforeEach(function(){
+    beforeEach(function () {
       this.User = current.define('User', {}, {
         name: 'John'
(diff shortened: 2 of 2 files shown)


gutobortolozzo:
@janmeier As we discussed, here is the PR with a unit test making sure the CLS and method FindOrCreate works fine together.

lib/model.js
```diff
@@ -1641,6 +1641,10 @@ Model.prototype.findOrCreate = function(options) {
     );
   }
 
+  if(this.sequelize.constructor.cls){
```
mickhansen: style: space between if and (: `if (`, also between ) and {, `) {`

gutobortolozzo:
@mickhansen I just pushed the changes you suggested, can you take a look please? :)

test/unit/model/find-or-create.test.js
```diff
@@ -0,0 +1,51 @@
...
+  , sinon = require('sinon')
+  , stub = sinon.stub
+  , Promise = require('bluebird');
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+
+  describe('method findOrCreate', function () {
+
+    before(function () {
+      current.constructor.cls = cls.createNamespace('sequelize');
```
mickhansen: Probably better just to include Sequelize and set it on that.

test/unit/model/find-or-create.test.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+/* jshint -W030 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support   = require(__dirname + '/../support')
+  , current   = Support.sequelize
```
mickhansen: either single space or aligned space, no need to mix.

test/unit/model/find-or-create.test.js
```diff
@@ -0,0 +1,51 @@
...
+      delete current.constructor.cls;
+    });
+
+    beforeEach(function () {
+      this.User = current.define('User', {}, {
+        name: 'John'
+      });
+
+      this.stub = stub(this.User.sequelize, 'transaction');
+      this.stub.returns(new Promise(function () {}));
```
mickhansen: If you're stubbing the transaction method, it won't trigger the CLS setting logic? Or how does this work.

lib/model.js
```diff
@@ -1641,6 +1641,10 @@ Model.prototype.findOrCreate = function(options) {
     );
   }
 
+  if (this.sequelize.constructor.cls) {
```
mickhansen: `&& !options.transaction`

lib/model.js
```diff
@@ -1641,6 +1641,10 @@ Model.prototype.findOrCreate = function(options) {
...
+    options.transaction = this.sequelize.constructor.cls.get('transaction');
```
mickhansen: Please follow the same style as `query`. https://github.com/gutobortolozzo/sequelize/blob/clsFindOrCreate/lib/sequelize.js#L689
Use the Sequelize instance rather than access `constructor`

mickhansen: Hmm although it's probably not possible to access the Sequelize class at this point so your method is probably the only way.

test/unit/model/find-or-create.test.js
gutobortolozzo: It will not trigger the transaction logic, but the test just want to check if even not passing the transaction and having the CLS enabled it will use the transaction provided by CLS.

lib/model.js
gutobortolozzo: I will add the extra check for ```options.transaction === undefined```

test/unit/model/find-or-create.test.js
mickhansen: But you are not providing any transaction to CLS right now, so the test should fail.

test/unit/model/find-or-create.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+      var options = {
+        where : {
+          name : '123'
+        }
+      };
+
+      this.User.findOrCreate(options);
+
+      expect(options).to.have.any.keys('transaction');
```
mickhansen: Probably better to test that the transaction property matches what's in CLS.

gutobortolozzo: Just pushed the check ```options.transaction === undefined``` and assert what's in cls is the same passed on cls.set(...)

test/unit/model/find-or-create.test.js
gutobortolozzo: @mickhansen, Just pushed the changes in the test so it assert against the previous transaction value inserted in the CLS, also pushed the check ```options.transaction === undefined```

test/unit/model/find-or-create.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+        var options = {
+          where : {
+            name : 'John'
+          }
+        };
+
+        self.User.findOrCreate(options);
+
+        expect(options.transaction).to.have.property('id', 123);
```
mickhansen: Are we sure this test does not make any database calls?
Does it reach the `findOne` call inside `findOrCreate`? If so we might want to stub that to ensure no calls are made.

gutobortolozzo: Well, I wasn't able to detected any database hit because the test stubs ```stub(this.User.sequelize, 'transaction')```, so the code inside the ```self.sequelize.transaction(options).bind({})``` is never reached.

mickhansen: Ah right, we call that internally first :)

mickhansen:
LGTM but would like @janmeier to have a look since CLS is his baby.

janmeier:
I like the way you are stubbing (even though it took me a bit to realize how everything works together and how it prevents hitting the DB ;) )

However, I'm not sure asserting that options have been modified is a good idea - generally people would like us not to modify the options object, so at some point it will probably be cloned. So perhaps you could add the assertion in the stub (even though thats not 100% clean)..

mickhansen:
Ah yeah @janmeier is right about that, generally we're moving towards cloning the passed options object so that assertion will fail in the future since the original object will be untouched.

gutobortolozzo:
@mickhansen @janmeier, Just pushed with tests against the stub. Now it's ready to be tested against cloned options ;)
Thanks.

janmeier:
Not the kind of stubbing I though of, but I actually like this better - :+1:

PR #4063, opened 2015-07-04
Fix broken links to api/ pages in the documentation.

sjakthol:
Here's a few fixes to broken links that point to the documents in the api/ directory.

To ease review I didn't run the api doc update script as it would've made the diff quite verbose. Instead I made the changes directly to both the jsdoc comment in lib/model.js and the generated file.


janmeier:
Thanks ! :)

PR #4075, opened 2015-07-08
Standardize `findAndCount` arguments

overlookmotel:
Changes the name of the argument taken by `Model#findAndCount()` to `options`. Was previously `findOptions`.

This makes no functional difference - just brings it into line with the rest of the codebase.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1394,14 +1394,14 @@ Model.prototype.count = function(options) {
  * @return {Promise<Object>}
  * @alias findAndCountAll
  */
-Model.prototype.findAndCount = function(findOptions) { // testhint options:1
-  if (findOptions !== undefined && !_.isPlainObject(findOptions)) {
+Model.prototype.findAndCount = function(options) {
+  if (options !== undefined && !_.isPlainObject(options)) {
     throw new Error('The argument passed to findAndCount must be an options object, use findById if you wish to pass a single primary key value');
   }
 
   var self = this
     // no limit, offset, order, attributes for the options given to count()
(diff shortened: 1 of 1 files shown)


changelog.md
```diff
@@ -3,6 +3,7 @@
 - [BUG] Fix wrong count for `findAndCountAll` with required includes [#4016](https://github.com/sequelize/sequelize/pull/4016)
 - [BUG] Fix problems related to parsing of unique constraint errors [#4017](https://github.com/sequelize/sequelize/issues/4017) and [#4012](https://github.com/sequelize/sequelize/issues/4012)
 - [BUG] Fix postgres path variable being surrounded by quotes to often in unique constraint errors [#4034](https://github.com/sequelize/sequelize/pull/4034)
+- Standardize `findAndCount` arguments - changed `findOptions` to `options`
```
mickhansen: Probably don't need a changelog entry for it, not sure the end users care a great deal.

lib/model.js
```diff
@@ -1394,14 +1394,14 @@ Model.prototype.count = function(options) {
  * @return {Promise<Object>}
  * @alias findAndCountAll
  */
-Model.prototype.findAndCount = function(findOptions) { // testhint options:1
```
mickhansen: You removed the testhint?

overlookmotel: Yes, the hint was only there because the argument wasn't called `options`, so it needed the hint to tell it where to attach `.logging` to.

This is really why I'm submitting this PR - just to allow the hint to be removed, and so to avoid confusion.

overlookmotel:
Have removed the changelog entry. Was trying to be thorough... too thorough!

overlookmotel:
Thanks @mickhansen!

PR #4076, opened 2015-07-08
Update documentation: node-validator is validate.js

gustavpursche:
It seems the author moved it's repository from [node-validator](https://github.com/chriso/node-validator) to [validator.js](https://github.com/chriso/validator.js).


janmeier:
Thanks! :)

PR #4080, opened 2015-07-08
Include the error message in the exception to ease debug

holm:
This simply adds the error message to the exception, so the root problem can more easily be diagnosed.


mickhansen:
What is displayed now, simply `ForeignKeyConstraintError`?

mickhansen:
ping @holm

holm:
Yes, sorry, just the name of the exception was shown, as no message was passed to the constructor. This made it pretty much impossible to figure out the problem.

PR #4083, opened 2015-07-09
adds support for for table triggers in mssql dialect for Model.update

shakdoesgithub:
- adds hasTrigger option in model.js that is then passed to
querygenerator.update
- adds integration test for Model.update


janmeier:
Looking good, but needs a rebase

shakdoesgithub:
looks like there is issues with the mssql server, it is timing out.

janmeier:
Good work @shakdoesgithub  :)

PR #4084, opened 2015-07-09
[Minor refactoring] Removed unnecessary code from instance validator

Znarkus:


Znarkus:
Weird about the tests failing, seems to be a before each hook that fails? 

```
  1) [MSSQL] Model "before each" hook:
     SequelizeDatabaseError: The object '[Users]' does not exist in database 'sequelize_test_93' or is invalid for this operation.

  2)  "before each" hook:
     SequelizeDatabaseError: Could not drop object 'dbo.users' because it is referenced by a FOREIGN KEY constraint.
```

janmeier:
Yeah, thats some silly code :). The error is probably because someone else interfered with the mssql server at that point (mssql servers are hosted on amazon since the travis VM does not run sql server)

janmeier:
Thanks :)!

Znarkus:
Great :)

PR #4087, opened 2015-07-09
Move validator-extends to it's own file

### LINKED ISSUE #4085: Proposal: move Validator with it's extends to own file
Just a short question, before starting a pull request: what do you think about moving all the Validator.extend() related stuff in [instance-validator.js](https://github.com/sequelize/sequelize/blob/master/lib/instance-validator.js) to it's own file (validator-extends.js), so it can be imported from the outside world?

gustavpursche:
This pull requests moves the validator.js code in `instance-validator.js` to `utils/validator-extends.js` to have the ability to reuse the manual extensions, sequelize makes, outside of the lib.

Thank's a lot for the great contribution docs!

Closes #4085


lib/utils/validator-extends.js
```diff
@@ -0,0 +1,85 @@
+'use strict';
+
+var validator = require('validator')
+  , _ = require('lodash');
+
+var extensions = {
+      notEmpty: function(str) {
```
mickhansen: spacing issues

mickhansen:
Spacing issues in the entirety of the one file.
Please name the file 'validator-extras'

lib/utils/validator-extends.js
gustavpursche: What's the issue? No Space between `:` and `function`?

lib/utils/validator-extras.js
```diff
@@ -0,0 +1,85 @@
...
+        return str.indexOf(elem) >= 0 && !!elem;
+      },
+      notContains: function(str, elem) {
+        return !this.contains(str, elem);
+      },
+      is: function(str, pattern, modifiers) {
+        return this.regex(str, pattern, modifiers);
+      },
+    }
+  , extendModelValidations = function(modelInstance) {
```
mickhansen: define this seperately so the methods before can have a ltitle saner indentation for all that code :)

gustavpursche:
Is there something missing here? Travis failure seems to be unrelated to this pull request ...

janmeier:
Rebased and merged in https://github.com/sequelize/sequelize/commit/b0037d67f05a760476b76d2f0a1d2db75e5b58bf

janmeier:
Thanks for your work @gustavpursche  :)

janmeier:
Yeah, the travis failure is related to problems with the mssql servers

Issue #4054, opened 2015-07-02
(bleedingly new) postgis GEOMETRY doesn't support srid


igrayson:
Mostly @mikeringrose.

PR https://github.com/sequelize/sequelize/pull/4013 added `GEOMETRY` support, but its srid handling may have some issues.

1. There's a typo in building `ST_GeomFromGeoJSON` sql, when an srid is present in the field def, causing invalid SQL (example: `ST_GeomFromGeoJSON('{"type":"Point","coordinates":[39.807222,-76.984722]}NaN')`. The issue is the second `+` on https://github.com/sequelize/sequelize/blob/56dbddde6739ecdad049881ae0a63e0f955fcf41/lib/dialects/postgres/query-generator.js#L903.

2. PostGIS' implementation of `ST_GeomFromGeoJSON` doesn't seem to support srid input at all. Neither source nor documentation mention accepting a second (srid) argument. Sequelize output is: `SequelizeDatabaseError: function st_geomfromgeojson(unknown, integer) does not exist`. (mysql `ST_GeomFromGeoJSON` does, fwiw).

mikeringrose:
Thanks. I'll take a look. Obviously I'm also missing a test case. 😠

mikeringrose:
Since our representation is geojson, the coordinate reference system is specified by the 'crs' field, like `{ type: 'Point', coordinates: [39.807222,-76.984722], crs:{ type: 'name', properties : { name: 'EPSG:3857'}}};`. I'll remove the srid from the ST_GeomFromGeoJSON call as 1) postgis doesn't support it and 2) postgis expects it to be specified in the geojson object.

krisr:
@mikeringrose, thanks - I agree removing it is best. I first played around with something like this, but I don't know enough about postgis / gis to feel comfortable with this type of change:

    			if (field.type.srid) {
				s = "ST_SetSRID(" + s + "," + field.type.srid + ")";
			}

I agree it is better to leave it up to the user to specify the crs field in the JSON.

igrayson:
Looks good.

Declaring my column type as `type: DataTypes.GEOMETRY('Point')`, and inserting point values as
```
{
      type: 'Point',
      coordinates: [ lat, long ],
      crs: { type: 'name', properties: { name: 'EPSG:4326'} }
    }
```
works as expected.

PR #4089, opened 2015-07-09
MySQL Geometry Support

mikeringrose:
Adds support for MySQL GEOMETRY support and addresses the issue https://github.com/sequelize/sequelize/issues/4054.


lib/dialects/abstract/query-generator.js
```diff
@@ -2149,7 +2149,7 @@ var QueryGenerator = {
   },
 
   geometrySelect: function(column) {
-    return this.quoteIdentifiers(column);
+    return 'AsText(' + this.quoteIdentifiers(column) + ') AS ' + column;
```
mickhansen: This syntax is supported by all dialects? Otherwise it should be in the dialect specific query generator.

test/unit/sql/data-types.test.js
```diff
@@ -667,28 +667,32 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
       });
     }
 
-    if (current.dialect.supports.ARRAY) {
+    // if (current.dialect.supports.GEOMETRY) {
```
mickhansen: The if should probably still be present.

lib/dialects/abstract/query-generator.js
janmeier: I'm pretty sure sqlite uses `AsText` and `GeomFromText` as well

There is a node module for spatialite (the GIS version of sqlite). But it doesn't seem to be very well maintained https://github.com/zhm/node-spatialite

test/unit/sql/data-types.test.js
mikeringrose: My bad. Totally should have been there.

lib/dialects/abstract/query-generator.js
mikeringrose: I went ahead and moved it into the mysql query-generator as it should be.

mikeringrose:
Errrr...not sure why mssql tests are failing.

janmeier:
Seems to be just a timeout - our mssql servers are very slow at the moment ..

janmeier:
Good work @mikeringrose !! :)

janmeier:
Do you want to take a stab at mssql and sqlite as well, or do you want me to look at those?

mikeringrose:
I can take a stab at them this weekend!

Mike Ringrose

janmeier:
Sweet!

nateonthenet:
There's a bug in here with MySQL/MariaDB if you're using main table aliases (e.g., as the result of a join query). In lib/dialects/abstract/query-generator.js, the following is invoked inside "selectQuery" (~line 994):
```javascript
      if (mainModel && mainModel._isGeometryAttribute(attr)) {
        return self.geometrySelect(attr);
      }
```

This results in queries that look like...
> SELECT `Table`.AsText(`Column`) AS `Table`.`Column`

...and obviously this won't work.

I patched it by setting a flag "fixGeometry" in the if statement above (instead of the return), and modifying lines ~1011-1013 from...
```javascript
      if (options.include && attr.indexOf('.') === -1 && addTable) {
        attr = mainTableAs + '.' + attr;
      }
```
... to ...
```javascript
      if (options.include && attr.indexOf('.') === -1 && addTable) {
        if (fixGeometry) {            
            attr = 'AsText(' + mainTableAs + '.' + attr + ') AS '+attr;
        } else {
            attr = mainTableAs + '.' + attr;
        }
      } else if (fixGeometry) {
          attr = 'AsText(' + attr + ' ) AS '+ attr;
      }
```

This works, but could stand to be properly abstracted back into the mysql dialect.

janmeier:
@nateonthenet I've refactored geometry in https://github.com/sequelize/sequelize/pull/4186 so we select the raw wkx data and parse it in js instead of selecting it as text, so it should be fixed once that is merged

PR #4093, opened 2015-07-11
Fix: add missing comma

uwolfer:


janmeier:
Thanks - I've changed this in the code comments myself so the change persists

PR #4102, opened 2015-07-13
[WIP] Allow built instance as bulkCreate argument

phanect:
This PR allows users to give built instances as `bulkCreate` argument. It works as if it is "`bulkSave`"

Example:

```javascript
var users = [];
users.push(sequelize.models.User.build({
  username: "Alice",
  password: "foobar"
}));

users.push(sequelize.models.User.build({
  username: "Bob",
  password: "tetete"
}));

sequelize.models.User.bulkCreate(users);

```

This PR is not completed yet. (requires docs)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,11 +31,6 @@ before_script:
 script:
   - "make test"
 
-branches:
-  only:
-    - master
-    - 1.7.0
-
 matrix:
   fast_finish: true
   include:
--- a/lib/model.js
+++ b/lib/model.js
@@ -1829,7 +1829,11 @@ Model.prototype.bulkCreate = function(records, options) {
     , now = Utils.now(self.modelManager.sequelize.options.dialect);
 
   var instances = records.map(function(values) {
-    return self.build(values, {isNewRecord: true});
+    if (values instanceof Instance) {
+      return values;
+    } else {
+      return self.build(values, {isNewRecord: true});
+    }
   });
 
   return Promise.try(function() {
--- a/test/integration/model/create.test.js
+++ b/test/integration/model/create.test.js
@@ -1374,6 +1374,39 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       });
     });
 
+    it('should save instances', function() {
+      var User = this.sequelize.define('User', {
+        username: {
+          type: Sequelize.STRING,
+          allowNull: false,
+          validate: {
+            notEmpty: true
+          }
+        }
+      });
+
+      var instances = User.build([
+        { username: 'alice' },
+        { username: 'bob' },
+        { username: 'carol' }
+      ]);
+
+      return User.bulkCreate(instances, {
+        validate: true,
+        individualHooks: true
+      }).then(function(users) {
+        expect(users[0].get('username')).to.equal('alice');
+        expect(users[1].get('username')).to.equal('bob');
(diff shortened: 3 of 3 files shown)


phanect:
I want to hear the opinion about API design.
I considered to create `bulkSave` instead of allowing instance for `bulkCreate`, but currently I didn't. Which is better to create `bulkSave`, or keep current implementation?

phanect:
I want to hear the opinion about API design.
I considered to create bulkSave instead of allowing instance for bulkCreate, but currently I didn't. Which is better to create bulkSave, or keep current implementation?

mickhansen:
What's the reasoning for this exactly? What does it offer over just using an array of values?

test/integration/model.test.js
```diff
@@ -2664,7 +2664,40 @@ describe(Support.getTestDialectTeaser('Model'), function() {
     return this.sequelize.sync({force: true});
   });
 
-  describe('bulkCreate errors', function() {
+  describe('bulkCreate', function() {
```
mickhansen: Probably wanna place this test other than where we test for errors.

phanect:
@mickhansen 

Because I want to use `set[AS]` and `add[AS]`.

Example:
```
var members = [];

var org = sequelize.models.Organization.build({
  name: "xx Foundation"
});

members.push(sequelize.models.Member.build({
  name: "Alice"
}));

members.push(sequelize.models.Member.build({
  name: "Bob"
}));

members.forEach(function(member) {
  org.addMember(member, { save: false });
});

sequelize.models.User.bulkCreate(members);
```

Currently I can do the same thing by adding `organization_id` as follows:
```
{
  name: "Bob",
  organization_id: org.get("id")
}
```
But this way is not clean and might cause bug.

test/integration/model.test.js
phanect: OK.
Where is the best place?
Is it OK to create another `describe` at the same level as `bulkCreate errors` in test/integration/model.test.js?

mickhansen: https://github.com/sequelize/sequelize/blob/master/test/integration/model/create.test.js#L1288

mickhansen:
Makes sense!

mickhansen:
Although i believe we only support `save: false` for belongsTo so you might want to use the setter inversely for 1:M (i.e. add to the belongsTo source).

test/integration/model.test.js
phanect: Oh, missed it! OK.

phanect:
I will restart working for this at 16th or 17th ;)

phanect:
It seems to require heavy modification to realize association before `save()`. (example shown in  [7252891](https://github.com/phanect/sequelize/commit/769894e6b3eaaf1ec39a5c0e154e34362576940b))

[When I run `setOrganization()`](https://github.com/phanect/sequelize/commit/769894e6b3eaaf1ec39a5c0e154e34362576940b#diff-4b224bbc09dfb867f8671d77e298bb8cR1446), `org.id` is `null` and then `null` is assigned to `members[*].OrganizationId`. So it is difficult to determine which `Organization` and `Member`s are associated.

I terminate to work for this and will close this PR...

mickhansen:
Yeah you only have access to the ID post bulkCreate when using `returning: true` with postgres. mysql and sqlite does not support returning multiple insert id's.

PR #4113, opened 2015-07-14
Prefix table names on geometry fields where multiple tables are selected

jimmed:
Say I have two associated tables, `A` and `B`, which both have a `location` field of the Geometry type.

If I were to select `A`, including `B`, then an attribute ambiguity error is thrown because neither of the references to `location` in the query are prefixed with the table name.

Here, I am ensuring that the table name is prefixed.


janmeier:
Thanks @jimmed  :)

janmeier:
Merged in https://github.com/sequelize/sequelize/commit/b063d617e3e9636e01102b8855cc585795af83cc

Issue #4007, opened 2015-06-25
Sequelize doesn't compare dates correctly


wmulligan:
sequelize is using the == comparison to look for changes when using instance.update().  This causes the two identical dates to show as "changed".

The line of code is here
```
if (!options.raw && ((primitives.indexOf(typeof value) === -1 && value !== null) || value !== originalValue)) {
```

This article explains the problem: http://stackoverflow.com/questions/492994/compare-two-dates-with-javascript

I suggest that we need better comparison checks for type DATE and DATEONLY

wmulligan:
Other issue is sequelize is not converting DATEONLY to a Date object, but it does convert the value read from.  So the code ends up comparing a string like 'YYYY-MM-DD' to a date object.

janmeier:
I told you dates would be an issue @mickhansen  ;) https://github.com/sequelize/sequelize/pull/3588

@wmulligan Yeah we need to implement some custom parsing logic for dateonly - currently I think its actually the db driver returning a date object

gutobortolozzo:
I implemented a small code inside ``` instance.set() ``` to prevent the method to identify date attribute has changed when the value is the same, It works as expect and nothing is added to ``` instance._changed```, but that raised a different question, when I call method ``` instance.save() ``` on one model that has nothing changed it generates one SQL updating the field ``` updatedAt ```. From my point of view, when nothing changed even with the value of ``` options.silent=false ``` no SQL should be sent to database. What do you guys think?

mickhansen:
`save` should definitely be a no-op if nothing has changed.

gutobortolozzo:
@mickhansen, Well, I make a PR to fix the date comparison and one other to prevent the ``` save ``` behavior described above. Thanks :)

gutobortolozzo:
@mickhansen, Just make the PR to fix the way dates are compared, about the ```save``` issues, I will make one other PR later.

janmeier:
Fixed by https://github.com/sequelize/sequelize/pull/4116

PR #4116, opened 2015-07-15
Sequelize doesn't compare dates correctly

gutobortolozzo:
As discussed in the issue #4007. This PR looks to fix the way that dates are compared.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -339,8 +339,13 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
           }
 
           // Convert date fields to real date objects
-          if (this.Model._hasDateAttributes && this.Model._isDateAttribute(key) && value !== null && value !== undefined && !(value instanceof Date) && !value._isSequelizeMethod) {
-            value = new Date(value);
+          if (this.Model._hasDateAttributes && this.Model._isDateAttribute(key) && !!value && !value._isSequelizeMethod) {
+            if (!(value instanceof Date)) {
+              value = new Date(value);
+            }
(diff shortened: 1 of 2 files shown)


lib/instance.js
```diff
@@ -342,6 +342,10 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
           if (this.Model._hasDateAttributes && this.Model._isDateAttribute(key) && value !== null && value !== undefined && !(value instanceof Date) && !value._isSequelizeMethod) {
             value = new Date(value);
           }
+
+          if (Utils._.isDate(value) && Utils._.isDate(originalValue) && value.getTime() === originalValue.getTime()) {
+            return;
+          }
```
mickhansen: This could probably be combined with the previous if.
So we'd have one wrapper if checking `isDateAttribute` and then two subsequent ifs making it a `Date` if it isn't and then comparing.

gutobortolozzo: @mickhansen, something like this?

```
          if (this.Model._hasDateAttributes && this.Model._isDateAttribute(key) && !!value && !value._isSequelizeMethod) {
            if (!(value instanceof Date)) {
              value = new Date(value);
            }
            if (value.getTime() === originalValue.getTime()) {
              return;
            }
          }
```

mickhansen:
Just MSSQL failing but that's having server issues atm.
Can you squash to a single commit? 3 commits with the same message does not make for nice history.

gutobortolozzo:
@mickhansen, Good thinking, Just squashed those commits :)

PR #4128, opened 2015-07-16
feat(belongs-to-many): countAssociations

mickhansen:


lib/associations/belongs-to-many.js
```diff
@@ -272,7 +272,17 @@ var BelongsToMany = function(source, target, options) {
...
      */
-    hasAll: 'has' + plural
+    hasAll: 'has' + plural,
+    /**
+     * Count everything currently associated with this, using an optional where clause
+     *
+     * @param {Object} [options]
+     * @param {Object} [options.where] An optional where clause to limit the associated models
+     * @param {String|Boolean} [options.scope] Apply a scope on the related model, or remove its default scope by passing false
+     * @return {Promise<Array<Instance>>}
```
janmeier: Int

PR #4145, opened 2015-07-18
Update mssql test credentials

sdepold:


sdepold:
:facepalm: @ commits

sdepold:
@janmeier @mickhansen could you check if `make mssql` sort of works for you locally (as in there are some integration tests which actually work). I'm not entirely sure why travis doesn't like the host :(

sdepold:
Build took 24 minutes. that's slower than before right?

janmeier:
@sdepold Yeah pretty sure its slower - but at least it doesn't fail / time out as much as before :)

sdepold:
the single tests are now in a worst case scenario at ~5 seconds. Guess I will spawn an instance in another region and check if it is faster. (We might want to ping different servers in different region for optimal results during the tests)

sdepold:
Yeah the domain is now working properly. I will squash the commits as this PR is good to go imho. I will provide another PR at some point which uses a different region (if needed)

sdepold:
rebased + squashed

janmeier:
Well, its better at least :D - only one failure out of 4. But the 0.10 build took close to 2 hours! :O

sdepold:
Build is green and for other mssql it only took 4 minutes. Not sure why the performance is so unstable. :(

janmeier:
I just re-ran the test that previously took two hours - now takes 2 minutes :D

janmeier:
No idea what happened before, but thanks for fixing it @sdepold  :D

sdepold:
You Are welcome. Please ping me when it's becoming slow again :)

PR #4149, opened 2015-07-19
doc(querying) Add documentation for the $notIn query operator.

oznu:
Documenting the $notIn query operator.


janmeier:
Thanks :)

PR #4150, opened 2015-07-19
Fix the bug in transactiona document.

Ghost141:
The previous example code is buggy because you need to get specified isolation level from Sequelize.Transaction.ISOLATION_LEVELS instead of Sequelize.Transaction.


janmeier:
Woops, thanks :)

PR #4151, opened 2015-07-20
fix(belongs-to-many): properly pair association based on the through …

mickhansen:
…argument

The code would previously pair association based on the through.model argument of the original options.
However options.through.model would generally be undefined since users usually just pass through: Model and the normalization to through.model
actually happened on association.through rather than options.through.
Normalization now happens on the options set before being assigned to association.through

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -61,6 +61,10 @@ var BelongsToMany = function(source, target, options) {
   this.doubleLinked = false;
   this.as = this.options.as;
 
+  if (!this.as && this.isSelfAssociation) {
+    throw new Error('\'as\' must be defined for many-to-many self-associations');
+  }
+
   if (this.as) {
     this.isAliased = true;
 
@@ -87,10 +91,6 @@ var BelongsToMany = function(source, target, options) {
    * If self association, this is the target association - Unless we find a pairing association
    */
   if (this.isSelfAssociation) {
-    if (!this.as) {
-      throw new Error('\'as\' must be defined for many-to-many self-associations');
-    }
-
     this.targetAssociation = this;
   }
 
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -36,11 +36,23 @@ var Utils = require('./../utils')
 var BelongsToMany = function(source, target, options) {
   Association.call(this);
 
+  options = options || {};
+
+  if (options.through === undefined || options.through === true || options.through === null) {
+    throw new Error('belongsToMany must be given a through option, either a string or a model');
+  }
+
+  if (!options.through.model) {
+    options.through = {
+      model: options.through
+    };
+  }
+
   this.associationType = 'BelongsToMany';
   this.source = source;
   this.target = target;
   this.targetAssociation = null;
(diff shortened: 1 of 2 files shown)


jagged3dge:
This fixes the issue with `SequelizeDatabaseError: ER_BAD_FIELD_ERROR` pointing to the wrong foreignKey being thown when using multiple self-referencing belongsToMany associations.

However, the *through* table gets created with only 1 reference column provided with model definition as `foreignKey`. It's missing the other reference column that references the model itself. The MySQL INSERT happens with only one value going to the `foreignKey` column in the _through_ table.

mickhansen:
@jagged3dge I'm unable to reproduce that with a unit test, the associations correctly get a foreign and other key assigned to them.

Can you try logging `Object.keys(joinModel.rawAttributes)` for your associations?

jagged3dge:
@mickhansen It's the second _joinModel_ that has this issue.

Here's the `rawAttributes` of the `UserFollowers` _joinModel_:
```
UserFollowers.rawAttributes: [ 'id',
  'createdAt',
  'updatedAt',
  'deletedAt',
  'UserId',
  'FollowerId' ]
```

Here's the `rawAttributes` for the `Invites` _joinModel_:
```
Invites.rawAttributes: [ 'createdAt', 'updatedAt', 'InviteeId' ]
```

mickhansen:
@jagged3dge You were passing Invites as a string to through correct?

mickhansen:
I think i'm able to reproduce this bug.

jagged3dge:
@mickhansen I _was_ using a string as the _through_. But when you requested the `rawAttributes` of the _joinModel_, I defined a new model and passed that to the association for retrieving the output provided above. I don't think this affects the issue.

Glad to hear you're able to reproduce this bug. Hoping for a quick and simple fix :+1:

mickhansen:
@jagged3dge `as` is required on self referential N:M relations, i'm fixing a bug where it incorrectly fails to throw an error when there's a missing `as`.

mickhansen:
The issue that both `foreignKey` and `otherKey` are defined based on the model name if not overwritten or using an `as`. Hence why we actually require `as` for this association work.

jagged3dge:
@mickhansen My test code defines both relations with `as`. Here's the code again:

```javascript
  name: {
    type: Sequelize.STRING,
    allowNull: true
  }
}, {
  paranoid: true
})

var UserFollowers = sequelize.define('UserFollowers')
var Invites = sequelize.define('Invites')

User.belongsToMany (User, {
  as: {
    singular: 'Follower',
    plural: 'Followers'
  },
  through: UserFollowers
})

User.belongsToMany (User, {
  as: {
    singular: 'Invitee',
    plural: 'Invitees'
  },
  foreignKey: 'InviteeId',
  through: Invites
})

```

janmeier:
@mickhansen Agreed :)

mickhansen:
@jagged3dge Let me just see if i can reproduce with that setup.

mickhansen:
@jagged3dge Sorry i'm unable to reproduce that. Please open a PR with a failing unit test.

jagged3dge:
@mickhansen Wierd. But, okay. I'll send in a failing unit test. Thx for your patience!

Issue #4120, opened 2015-07-15
many2many with getters through attributes


ns3777k:
Code:
```js
'use strict';

var SEQUELIZE = require('sequelize'),
	db = new SEQUELIZE('postgres', 'postgres', '123456', {
		dialect : 'postgres',
		host    : 'localhost'
	});

var ConnectionPlan = db.define('ConnectionPlan', {
	title : SEQUELIZE.TEXT,
	price : SEQUELIZE.INTEGER
}, {
	tableName : 'connection_plan',
	createdAt : false,
	updatedAt : false,
	deletedAt : false
});

var ConnectionOption = db.define('ConnectionOption', {
	title     : SEQUELIZE.TEXT,
	price     : {
		type  : SEQUELIZE.INTEGER,
		field : 'price',
		get   : function () {
			return this.getDataValue('price');
		},
		set   : function (val) {
			this.setDataValue('price', val);
		}
	}
}, {
	tableName : 'connection_option',
	createdAt : false,
	updatedAt : false,
	deletedAt : false
});

var PlanOption = db.define('PlanOption', {
	planid    : SEQUELIZE.INTEGER,
	optionid  : SEQUELIZE.INTEGER,
	recommend : SEQUELIZE.BOOLEAN,
	weight    : SEQUELIZE.INTEGER
}, {
	tableName : 'plan2option',
	createdAt : false,
	updatedAt : false,
	deletedAt : false
});

ConnectionPlan.belongsToMany(ConnectionOption, {
	through    : { model : PlanOption },
	foreignKey : 'planid',
	as         : 'compatible_options'
});

ConnectionOption.belongsToMany(ConnectionPlan, {
	through    : { model : PlanOption },
	foreignKey : 'optionid',
	as         : 'compatible_plans'
});
```

Then I create a plan and an option"

```js
db.sync({ force : true })
	.then(function () {
		return ConnectionPlan.create({
			title : 'Lite Plan',
			price : 100
		});
	})
	.then(function (plan) {
		return ConnectionOption.create({
			title : 'Additional 100GB to your storage',
			price : 10
		});
	});
```

Then I add a link between those (Strange I couldn't do that with plan.addOption) in database:

```sql
insert into plan2option (planid, optionid, recommend, weight) values (1, 1, true, 100);
```

And then I try to retrieve a plan with the options in include:

```js
ConnectionPlan.findAll({
	include : [
		{
			model      : ConnectionOption,
[...]

PR #4153, opened 2015-07-20
plain attributes without getters

ns3777k:
https://github.com/sequelize/sequelize/issues/4120 first try.

### DIFF
--- a/test/integration/instance.test.js
+++ b/test/integration/instance.test.js
@@ -1668,6 +1668,37 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
         expect(user.get({ plain: true })).to.deep.equal({ username: 'foo', id: null });
       });
     });
+
+    it('gets plain object without unrequested attributes', function() {
+      var ConnectionPlan = current.define('ConnectionPlan', {
+        title : DataTypes.TEXT,
+        price : DataTypes.INTEGER
+      }, {
+        getterMethods : {
+          printedPrice : function () {
+            return 'i_should_be_in_plain';
+          }
+        }
+      });
+
+      return ConnectionPlan.sync({force : true})
+        .then(function () {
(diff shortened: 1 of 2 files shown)


janmeier:
This means that `getterMethods` will never be called :)

```js
sequelize.define('', {

}, {
  getterMethods: {
    imNeverCalled: function ()
  }
});
```

The if should only be when _key is an actual attribute of the model.

test/integration/instance.test.js
```diff
@@ -1634,6 +1635,39 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
...
+        .then(function () {
+          return ConnectionPlan.find({
+            attributes : attrs
+          });
+        })
+        .then(function (plan) {
+          var plain = plan.get({plain: true}),
+            keys = Object.keys(plain);
+
+          expect(!!_.difference(keys, attrs).length).to.equal(false);
```
janmeier: Not sure I agreed with this test - it would seem to suggest you have to add `formattedPrice` to `attributes` in order for it to be called. But if you do so, sequelize actually tries to select a field called `formattedPrice`.

I think we should only check if the key is in attributes if it is a normal attribute

janmeier:
ping @

janmeier:
ping @ns3777k

ns3777k:
@janmeier Oh totally forgot about that :-( get it done this week

ns3777k:
@janmeier Sorry for the delay, lotta work. I updated PR, can you take a look? I'm not sure whether virtual fields and `getterMethods` should be added to the plain version of an object. I keep `getterMethods` converting to plain in PR, shall I add virtual fields as well?

janmeier:
@ns3777k Sorry for leaving this hanging.

I just tested on master, and the test passes there so I'm going to close this

PR #3780, opened 2015-05-21
Always quote identifiers after AS in join queries. Fixes #1589

idris:

PR #4156, opened 2015-07-20
Refactor join query generation

mickhansen:


mickhansen:
Hmm, so limiting a include might only be possible in postgres since postgres is the only dialect to support LATERAL JOIN's that let you reference the main table from a join.

bomattin:
Damn. I thought I was able to do this by setting `subQuery: false` but it turns out I was incorrect.

mickhansen:
@bomattin Huh?

bomattin:
In order to filter on includes AND limit on the main model, we had to call our finder with `subQuery: false`. I was thinking that this would work for the reverse and allow limiting on includes, but I did some testing on it and wasn't able to get the query I needed.

mickhansen:
@bomattin It wouldn't work because of cartesian products. If you have two products with 3 tags each you get 6 rows, and LIMIT 4 would just result in one product missing tags.

My thought was initially to limit in a subquery, but doesn't look like it's possible to reference the main table from a subquery used in a join.

bomattin:
@mickhansen Can you post a test query you're trying to generate? I think I'm coming to the same conclusion you are but just want to be certain.

lib/associations/belongs-to-many.js
```diff
@@ -323,18 +326,18 @@ BelongsToMany.prototype.injectAttributes = function() {
   if (this.primaryKeyDeleted === true) {
     targetAttribute.primaryKey = sourceAttribute.primaryKey = true;
   } else if (this.through.unique !== false) {
-    var uniqueKey = [this.through.model.tableName, this.identifier, this.foreignIdentifier, 'unique'].join('_');
```
janmeier: Can `identifier` and `foreignIdentifier` be remove completely?

lib/associations/belongs-to-many.js
```diff
@@ -356,25 +359,58 @@ BelongsToMany.prototype.injectAttributes = function() {
...
+  this.manyFromTarget = new HasMany(this.target, this.through.model, {
+    foreignKey: this.otherKey
+  });
+  this.oneFromTarget = new HasOne(this.target, this.through.model, {
+    foreignKey: this.otherKey,
+    as: this.through.model.name
+  });
+
+  if (this.paired && this.paired.otherKeyDefault) {
+    this.paired.toTarget = new BelongsTo(this.paired.through.model, this.paired.target, {
```
janmeier: Isn't this the same as `this.toSource`

lib/dialects/abstract/query-generator.js
```diff
@@ -1211,22 +1214,14 @@ var QueryGenerator = {
...
-                        source: topInclude.association.target,
-                        target: topInclude.association.source,
-                        identifier: topInclude.association.foreignIdentifier,
-                        identifierField: topInclude.association.foreignIdentifierField
-                      },
-                      required: true,
-                      include: topInclude.include,
-                      _pseudo: true
-                    }],
+                    include: Model.$validateIncludedElements({
```
janmeier: Could the validation somehow be done ahead of time, when the association is set up

mickhansen: Not really, it's the validation that's normally called as part of a find call. We have to call it here since we call selectQuery directly, the validation does all sorts of stuff like figure out if we need a subquery, what associations should be in the subquery etc.

lib/associations/belongs-to-many.js
mickhansen: That is a bit more work, but in time yes.

lib/associations/belongs-to-many.js
mickhansen: No, it's created on the other association (in case of two BTM relations on the same through model but they aren't defining otherKey only foreignKey)

test/unit/sql/select.test.js
```diff
@@ -83,7 +83,7 @@ describe(Support.getTestDialectTeaser('SQL'), function() {
       });
     });
 
-    it('include (left outer join)', function () {
+    it.skip('include (left outer join)', function () {
```
janmeier: y u skip!

test/unit/associations/belongs-to-many.test.js
```diff
@@ -110,6 +113,230 @@ describe(Support.getTestDialectTeaser('belongsToMany'), function() {
     });
   });
 
+  describe('pseudo associations', function () {
+    it('should setup belongsTo relations to source and target from join model with defined foreign/other keys', function () {
+      var Product = this.sequelize.define('Product', {
```
janmeier: might want to add a test with `field` here ?

test/unit/sql/select.test.js
mickhansen: i'm not sure who wrote those tests, but the include stubbing is horrible ;p

test/unit/associations/belongs-to-many.test.js
mickhansen: It only affects `identifierField` which is really want to move to later in the query building process - so i skipped it ;p

janmeier: Makes sense then :)

test/unit/sql/select.test.js
janmeier: preeeeeety sure you did ;)

janmeier: Or well, actually not https://github.com/sequelize/sequelize/pull/3780/files?diff=unified

One  of those pertains to quoteIdentifers, so might be worth to keep that. Not that I like quoteIdentifers but I have a feeling several people rely on it

mickhansen: Well i was dumb then! Shame on me!

janmeier:
Sweetness!

PR #4173, opened 2015-07-22
Fix documentation of fields filter for create function

elvo86:
The documentation for the .create() function was incorrect in how to specify a filter for fields being saved.


janmeier:
Thanks :-)

PR #4176, opened 2015-07-23
Update sent to database when nothing changed in instance

gutobortolozzo:
Hi,

This PR refers to cases where one instance is hydrated from database, some processing is done and no field changes, but when ```instance.save()``` is called one update with field ```updatedAt``` is sent to database. 
From my point of view even without the option silent turned on this SQL should not be sent. What you think?

### DIFF
--- a/test/integration/instance.test.js
+++ b/test/integration/instance.test.js
@@ -981,6 +981,30 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
       });
     });
 
+    describe('when nothing changed', function() {
+
+      beforeEach(function () {
+        this.clock = sinon.useFakeTimers();
+      });
+
+      afterEach(function () {
+        this.clock.restore();
+      });
+
+      it('does not update timestamps', function() {
+        var self = this;
+        return self.User.create({ username: 'John' }).then(function() {
+          return self.User.findOne({ username: 'John' }).then(function(user) {
+            var updatedAt = user.updatedAt;
+            self.clock.tick(2000);
+            return user.save().then(function(newlySavedUser) {
+              expect(newlySavedUser.updatedAt).to.equalTime(updatedAt);
+            });
+          });
+        });
+      });
+    });
+
     it('updates with function and column value', function() {
       var self = this;
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # NEXT
 - [FIXED] Fix findOrCreate regression trying to add a transaction even if there is none
 - [FEATURE] Added default validation based on attribute types. [#3472](https://github.com/sequelize/sequelize/pull/3472). The validation _cannot_ be disabled. If you really want to completely disable it, you can remove the `validate` function from the corresponding datatype, but know that this permanently disables the validation.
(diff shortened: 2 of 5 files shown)


janmeier:
Hmm, I definitely see the reasoning behind this PR, but I think some users would be confused if their `save` call sometimes did nothing ... I think this also counts as a BC break, since calling save no longer always updates `updatedAt`. What do you think @mickhansen

test/integration/instance.test.js
```diff
@@ -981,6 +981,20 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
       });
     });
 
+    it('does not update timestamps when nothing changed', function() {
+      var User = this.User;
+      return User.create({ username: 'John' }).then(function() {
+        return User.findOne({ username: 'John' }).then(function(user) {
+          var updatedAt = user.updatedAt;
+          return this.sequelize.Promise.delay(2000).then(function() {
```
janmeier: Please use fake timers http://sinonjs.org/docs/#clock to artifically advance time

mickhansen:
@janmeier we did agree a while back that no changes should be a noop - this is more of a bug i'd say.

janmeier:
Well I just forgot then :D But yeah, I guess thats the logical conclusion when we only save the fields that have changed. 

Let's make sure its properly documented in the api docs then

test/integration/instance.test.js
```diff
@@ -981,6 +981,20 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
...
+            return user.save().then(function(newlySavedUser) {
+              expect(newlySavedUser.updatedAt.getTime()).to.equal(updatedAt.getTime());
```
janmeier: `expect(newlySavedUser.updatedAt).to.equalTime(updatedAt);`

gutobortolozzo:
@janmeier @mickhansen Hi, just pushed the changes and documentation updates, but I'm not sure the changes that I made in the documents are enough. Can you take a look please? :)

janmeier:
@gutobortolozzo You might want to add a mention here as well https://github.com/sequelize/sequelize/blob/master/lib/instance.js#L467 that it will only save changed fields, and do nothing if no fields have changed.

Please squash the commits as well, then this should be ready for merge :)

gutobortolozzo:
@janmeier, changes pushed :)

janmeier:
Good work @gutobortolozzo  :)

gustavpursche:
Just a question after a quick look: This doesn't work, if the database and the model are out of sync, right? The local model can't know, whether it's dirty or not without checking the database, so I guess there should be a way to force the update.

janmeier:
@gustavpursche right, but if sequelize thinks nothing has changed it won't send anything. I think the best way to force this would be to explicitly mark all fields as changed (`instance.changed('field')`)

gustavpursche:
@janmeier alright. But isn't that confusing? I mean, even if the local instance is unchanged, in the out-of-sync scenario it should run the update. So if you want to work around that, you'd always have to set the fields to changed ...

janmeier:
I agree, it would be cumbersome - probably easier to have a flag to save that sets changed under the covers

PR #4177, opened 2015-07-23
include.seperate and include.limit support for hasMany

mickhansen:


lib/associations/has-many.js
```diff
@@ -221,247 +221,291 @@ HasMany.prototype.injectAttributes = function() {
...
+  };
 
-    return this[association.accessors.get](
-      options,
-      { raw: true }
-    ).then(function(associatedObjects) {
-      return associatedObjects.length === instances.length;
-    });
+  obj[this.accessors.remove] = obj[this.accessors.removeMultiple] = function(instances, options) {
+    return association.add(this, instances, options);
```
janmeier: ....

mickhansen: whoops :D

lib/model.js
```diff
@@ -1178,6 +1191,8 @@ Model.prototype.all = function(options) {
  * @param  {Object}                    [options.include[].where] Where clauses to apply to the child models. Note that this converts the eager load to an inner join, unless you explicitly set `required: false`
  * @param  {Array<String>}             [options.include[].attributes] A list of attributes to select from the child model
  * @param  {Boolean}                   [options.include[].required] If true, converts to an inner join, which means that the parent model will only be loaded if it has any matching children. True if `include.where` is set, false otherwise.
+ * @param  {Boolean}                   [options.include[].seperate] If true, runs a seperate query to fetch the associated instances, only supported for hasMany associations
```
janmeier: https://en.wiktionary.org/wiki/seperate

lib/model.js
```diff
@@ -1253,15 +1269,57 @@ Model.prototype.findAll = function(options) {
...
+  return Promise.map(options.include, function (include) {
+    if (!include.seperate) {
+      return Model.$findSeperate(
+        results.reduce(function (memo, result) {
+          var associations = result.get(include.association.as);
+          if (!Array.isArray(associations)) associations = [associations];
+
+          return memo.concat(associations);
+        }, []),
+        _.assign(
```
janmeier: No need to repeat this in both blocks

lib/dialects/abstract/query-generator.js
```diff
@@ -1,4 +1,4 @@
-'use strict';
+  'use strict';
```
janmeier: :fire:

test/integration/include/seperate.test.js
```diff
@@ -0,0 +1,249 @@
+'use strict';
```
janmeier: No comments here - tests look good :)

lib/model.js
mickhansen: Ah that's going to be annoying to change and remember if i constantly misspelled it :D

lib/model.js
mickhansen: The creating of options? The last argument they take is seperate in each. The omit could however be changed to only once though.

SourceBoy:
Hi guys, what does the query look like with
```
include: [{
  separate: true,
  limit: 5
}]
```
It looks like it does another query to retrieve associated records rather than retrieving them in one shot.

mickhansen:
@SourceBoy The `seperate` option tells Sequelize to pull it into a seperate query, both to support LIMIT/ORDER'ing of associations but also to counter expensive cartesian products in large joins.

Ideally we still want to support a single query limit of associations, but `seperate` will go along way :)

SourceBoy:
Cool. Thanks for the clarification.

test/support.js
```diff
@@ -29,7 +29,7 @@ Sequelize.Promise.onPossiblyUnhandledRejection(function(e, promise) {
 Sequelize.Promise.longStackTraces();
 
 // shim all Sequelize methods for testing for correct `options.logging` passing
-if (!process.env.COVERAGE) supportShim(Sequelize);
+if (!process.env.COVERAGE && false) supportShim(Sequelize);
```
mickhansen: @overlookmotel i forgot about this, i'd like to you take a look at some point, but the supportShim severely fucked up during a refactor of HasMany.

overlookmotel: H @mickhansen. Work has quietened down now, so I've got some time to get back on coding. Was planning to finally get round to writing the tests for transaction and CLS context passing.

How did the supportShim fuck up? Happy to have a look and figure it out...

overlookmotel: Did it start going haywire with this PR?

mickhansen: I don't recall exactly, but the call stack got butchered so i couldn't figure it out, but probably the stubs were no longer correct so tests were failing on the query option association.

mickhansen: Yep, because i refactored the methods in HasMany, so i reckon the stubs need to be fixed, i didn't have the energy for it when i was implementing and then i simply forgot ;p

overlookmotel: OK cool. I'll have a look when I get a chance sometime in next couple of weeks.

Issue #2042, opened 2014-07-14
Data-type UP / DOWN system


janmeier:
Supersedes #1690 

What we want is some kind of system, which allows a more or less dialect-independent way of converting data types between their javascript at DB representations. This system should also be extendable by the user

florian-rossiaud:
Hi,

Do you know for when this feature will be implement ? May I can help ?

mickhansen:
@florian-rossiaud No estimate. When we get to it is all we can say i'm afraid.

janmeier:
... But pull requests are always welcome :-). 

For this feature we'd probably need a bit of api discussion first - feel free to provide a suggestion, or more details into the use case you'd like this to support

florian-rossiaud:
I think the DataType should know how to escape his value. We could have a function inside DataType `escape` who espace the value on the query.

For now escape is like that :
      
    escape: function(value, field) {
      if (value && value._isSequelizeMethod) {
        return this.handleSequelizeMethod(value);
      }

      if (Utils._.isObject(value) && field && (field.type instanceof DataTypes.HSTORE || DataTypes.ARRAY.is(field.type, DataTypes.HSTORE))) {
        if (field.type instanceof DataTypes.HSTORE){
          return "'"  + hstore.stringify(value) + "'";
        } else if (DataTypes.ARRAY.is(field.type, DataTypes.HSTORE)) {
          return "ARRAY[" + Utils._.map(value, function(v){return "'" + hstore.stringify(v) + "'::hstore";}).join(",") + "]::HSTORE[]";
        }
      } else if(Utils._.isArray(value) && field && (field.type instanceof DataTypes.RANGE || DataTypes.ARRAY.is(field.type, DataTypes.RANGE))) {
        if(field.type instanceof DataTypes.RANGE) { // escape single value
          return "'"  + range.stringify(value) + "'";
        }
        else if (DataTypes.ARRAY.is(field.type, DataTypes.RANGE)) { // escape array of ranges
          return "ARRAY[" + Utils._.map(value, function(v){return "'" + range.stringify(v) + "'";}).join(",") + "]::" + field.type.toString();
        }
      } else if (field && (field.type instanceof DataTypes.JSON || field.type instanceof DataTypes.JSONB)) {
        value = JSON.stringify(value);
      } else if (Array.isArray(value) && field && DataTypes.ARRAY.is(field.type, DataTypes.JSON)) {
        return "ARRAY[" + value.map(function (v) {
[...]

PR #4186, opened 2015-07-24
WIP: Datatypes system refactor

janmeier:
Phew - this turned out to be a bit more work than I had anticipated - but I think its for the better!

I still need to write the docs for the user facing parts (parse / stringify), but I want to get a review on the API before doing so.

Datatypes can now have `parse` and `stringify` functions. `stringify` is called in `QueryGenerator.escape` to convert the value into a DB representation. I've moved some of the code from sql-string directly into data-types. We might be able to move some more later.
Parse is called when values are recieved from the DB. The parse part is not tied up to any model, meaning it also works for raw queries. Instead it is connected to type id (postgres and mssql) or type string (mysql and sqlite). This means that if two types use the same backing type in the DB, they cannot have two different parse functions, because we can't tell the difference when parsing. Which types this is is different from dialect to dialect - but we throw an error up front if a parse function is not supported for a particular dialect to make the error obvious to the user.

While I was messing around I also rewrote how type identification works in sqlite - it's a bit smarter now, and I've fixed a couple of bugs where dates weren't parsed properly in includes. We still issue `PRAGMA table_info` before each query though - this could probably be cached - but I don't want to go into that now :)


lib/data-types.js
```diff
@@ -39,24 +42,47 @@ var util = require('util')
...
+  deprecate(text + '\n>> Check: ' + link);
+};
+ABSTRACT.prototype.stringify = function (value, options) {
+  if (this.$stringify) {
+    return this.$stringify(value, options);
+  }
+  return value;
+};
+
+ABSTRACT.inherits = function (Constructor) {
```
janmeier: Instead of calling `util.inherits` after each datatype, I added an inherits function to the abstract datatype, which inherits both prototype and static methods. Its also a utility method which creates the constructor for the new type, in cases where the constructor doesn't do anything interesting - saves a lot of copy-pasta

lib/data-types.js
```diff
@@ -800,3 +841,14 @@ module.exports = {
   'DOUBLE PRECISION': DOUBLE,
   GEOMETRY: GEOMETRY
 };
+
+_.each(dataTypes, function (dataType) {
+  dataType.types = {};
+});
+
+dataTypes.postgres = require('./dialects/postgres/data-types')(dataTypes);
```
janmeier: All dialect-specific datatypes are now instantianted here

lib/dialects/mssql/data-types.js
```diff
@@ -1,179 +1,197 @@
...
 
-var BaseTypes = require('../../data-types')
-  , util = require('util')
-  , _ = require('lodash');
+var _ = require('lodash');
+
+module.exports = function (BaseTypes) {
+  var warn = BaseTypes.ABSTRACT.warn.bind('https://msdn.microsoft.com/en-us/library/ms187752%28v=sql.110%29.aspx');
+
+  BaseTypes.DATE.types.mssql = [42];
```
janmeier: Each dialect specifies the ids / strings we should use to map to a parse function for that type

lib/dialects/mssql/data-types.js
```diff
@@ -1,179 +1,197 @@
...
+  var DATE = BaseTypes.DATE.inherits();
 
-DATE.prototype.toSql = function() {
-  return 'DATETIME2';
-};
+  DATE.prototype.toSql = function() {
+    return 'DATETIME2';
+  };
 
-var INTEGER = function() {
```
janmeier: Nothing changed here - only indentation change by wrapping the whole file in a function

lib/dialects/mysql/query-generator.js
```diff
@@ -260,16 +257,7 @@ var QueryGenerator = {
       };
     }
 
-    var template;
-
-    if (attribute.type instanceof DataTypes.ENUM) {
```
janmeier: This happens in the datatype

lib/dialects/mysql/query-generator.js
```diff
@@ -352,21 +340,6 @@ var QueryGenerator = {
     return Utils.addTicks(identifier, '`');
   },
 
-  escape: function(value, field) {
```
janmeier: :fire:

lib/dialects/mysql/query-generator.js
```diff
@@ -391,7 +364,7 @@ var QueryGenerator = {
   },
 
   geometrySelect: function(column) {
-    return 'AsText(' + this.quoteIdentifiers(column) + ') AS ' + column;
+    return this.quoteIdentifiers(column);
```
janmeier: We select the raw wkx and parse it instead

lib/dialects/postgres/query-generator.js
```diff
@@ -869,47 +866,6 @@ var QueryGenerator = {
...
-      }
-    }
-
-    if (Utils._.isObject(value) && field && (field.type instanceof DataTypes.HSTORE || DataTypes.ARRAY.is(field.type, DataTypes.HSTORE))) {
-      if (field.type instanceof DataTypes.HSTORE){
-        return "'"  + hstore.stringify(value) + "'";
-      } else if (DataTypes.ARRAY.is(field.type, DataTypes.HSTORE)) {
-        return 'ARRAY[' + Utils._.map(value, function(v){return "'" + hstore.stringify(v) + "'::hstore";}).join(',') + ']::HSTORE[]';
-      }
-    } else if(Utils._.isArray(value) && field && (field.type instanceof DataTypes.RANGE || DataTypes.ARRAY.is(field.type, DataTypes.RANGE))) {
```
janmeier: <3 :fire:

lib/dialects/sqlite/query.js
```diff
@@ -24,6 +26,28 @@ Query.prototype.getInsertIdField = function() {
   return 'lastID';
 };
 
+Query.prototype.$collectModels = function(include, prefix) {
```
janmeier: I have a feeling this function does a lot of what `groupJoinData` does - but there's no way I'm looking into that function without being forced to :D

Basically, it creates a map from include alisesses to models. Fx `'users.projects'` -> project model

lib/instance.js
```diff
@@ -366,10 +366,6 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
           }
         }
 
-        if (this.Model._hasGeometryAttributes && this.Model._isGeometryAttribute(key) && _.isString(value)) {
```
janmeier: This happens in `GEOMETRY.parse` now

lib/model.js
```diff
@@ -87,15 +87,6 @@ var Model = function(name, attributes, options) {
       throw new Error('Unrecognized data type for field ' + name);
     }
 
-    if (attribute.type instanceof DataTypes.ENUM) {
```
janmeier: Happens in `normalizeAttribute` instead

lib/model.js
```diff
@@ -777,14 +768,6 @@ Model.prototype.refreshAttributes = function() {
     }
 
     if (definition.hasOwnProperty('defaultValue')) {
-      if (typeof definition.defaultValue === 'function' && (
```
janmeier: Happens in `normalizeAttribute` instead

lib/query-interface.js
```diff
@@ -85,26 +85,7 @@ QueryInterface.prototype.createTable = function(tableName, attributes, options)
       attribute = { type: attribute, allowNull: true };
     }
 
-    attribute.type = self.sequelize.normalizeDataType(attribute.type);
```
janmeier: Happens in `normalizeAttribute` instead

lib/query-interface.js
```diff
@@ -85,26 +85,7 @@ QueryInterface.prototype.createTable = function(tableName, attributes, options)
...
-
-    if (attribute.hasOwnProperty('defaultValue')) {
-      if (typeof attribute.defaultValue === 'function' && (
```
janmeier: Happens in `normalizeAttribute` instead

lib/sql-string.js
```diff
@@ -1,28 +1,10 @@
 'use strict';
```
janmeier: I've cleaned this file up a bit to only include the parts we need. I looked into replacing in with https://www.npmjs.com/package/sqlstring, but that's not going to happen - we have a lot of modifications, and that module only supports mysql

lib/sql-string.js
```diff
@@ -30,19 +12,7 @@ SqlString.escapeId = function(val, forbidQualified) {
   return '`' + val.replace(/`/g, '``').replace(/\./g, '`.`') + '`';
 };
 
-SqlString.escape = function(val, stringifyObjects, timeZone, dialect, field) {
-  if (arguments.length === 1 && typeof val === 'object' && val !== null) {
```
janmeier: Never used in sequelize codebase

test/integration/dialects/postgres/query-generator.test.js
```diff
@@ -548,7 +549,7 @@ if (dialect.match(/^postgres/)) {
           arguments: ['myTable', {data: new Buffer('Sequelize') }],
           expectation: "INSERT INTO \"myTable\" (\"data\") VALUES (E'\\\\x53657175656c697a65');"
         }, {
-          arguments: ['myTable', {name: 'foo', numbers: new Uint8Array([1, 2, 3])}],
```
janmeier: No idea why Uint8Array was used here..

test/integration/dialects/postgres/range.test.js
```diff
@@ -57,8 +57,9 @@ if (dialect.match(/^postgres/)) {
       });
 
       it('should handle date values', function () {
-        expect(range.stringify([new Date(Date.UTC(2000, 1, 1)),
```
janmeier: This now tests the range datatype instead of range.js directly

janmeier:
Ref https://github.com/sequelize/sequelize/issues/2042

lib/data-types.js
```diff
@@ -39,24 +42,47 @@ var util = require('util')
...
 };
 ABSTRACT.prototype.toSql = function() {
   return this.key;
 };
-ABSTRACT.prototype.warn = function(text) {
-  if (!warnings[text]) {
-    warnings[text] = true;
-    console.warn('>> WARNING:', text, '\n>> Check:', this.dialectTypes);
+ABSTRACT.warn = function(link, text) {
+  deprecate(text + '\n>> Check: ' + link);
```
BridgeAR: This prints a deprecation notice but it's not really deprecated. It's just not the right type used. And the line provided by the module is the one from the DataTypes but it should be the line where the user inserted the wrong type. So this should probably be reconsidered.

BridgeAR:
I only had a short glimpse over the changes but this is looking very promising and I really like the overall idea behind it.

lib/data-types.js
janmeier: Yeah, youre right ..

mickhansen: Agree, should be a warning more than a deprecation since we do want to normalize, but people should be aware that stuff might happen.

janmeier: But it haz suzh nize colorz!

Alright, I'll change it back to console.log...


This is a general problem with depd though - I actually thought it printed there line where the user did something wrong, but it always references a sequelize internal line number

mickhansen: I thought it did that aswell, otherwise it's quite useless.

BridgeAR: If I'm not mistaken the issue with the call stack is that the warning is on the prototype and called from a internal function. That way depd is going to look at the next upper level and find the internal function that called it and prints that as the wrong line. So the depd call would have to happen in the function itself. It might work if you do something like that (I'm not sure though):
```js
ABSTRACT.warn = function (text) {
  return text + '\n>> Check: ' + this.dialectTypes;
};
// And now somehwere where the warning is triggered:
function trigger (value) {
  if (typeof value !== 'number') {
    deprecate(this.warn(text));
  }
}
```
But it would not solve the fact that it's not really a deprecation :) (and you can't suppress that word). If you want the colors that's easy to implement without the library. Just use something like `console.log('\x1b[36mCOLORED_WORDS\x1b[0m');`

janmeier: @BridgeAR Yep, you're right.

I just did a check, and everywhere we use depd is actually called from internally (e.g. [here](https://github.com/sequelize/sequelize/blob/master/lib/utils.js#L377)), so it never works as intended :D

I was thinking of writing my own version of depd, which basically traverses up the stack until it meets a filename that's not `node_modules/sequelize/lib/**`. Dunno how hard its going to be, but right now depd is not working any better than what we had before :D

janmeier: https://www.npmjs.com/package/stack-trace looks promising, and we don't have to mess with the error object ourselves.

Anyhow, any review comments on the other million changes in this PR  ;) ?

janmeier:
@mikeringrose Just want to point your attention to this PR. Once its merged, all that needs to be done to add geometry support for sqlite and mssql is hopefully to add parse and stringify functions for the two dialects

lib/dialects/postgres/data-types.js
```diff
@@ -1,205 +1,377 @@
...
+  };
+
+  var DATE = BaseTypes.DATE.inherits();
+
+  DATE.prototype.toSql = function() {
+    return 'TIMESTAMP WITH TIME ZONE';
+  };
+
+  BaseTypes.DATE.types.postgres = {
+    oids: [1184],
```
janmeier: Add parsing for timestamp without timezone as well

janmeier:
Ping @mickhansen

janmeier:
This is going to be a bitch to rebase - looking at you @mickhansen !

janmeier:
@mickhansen This also fixes the issues mentioned by #4822, but by parsing the raw value instead of selecting it as text.

_hint hint_

lib/data-types.js
```diff
@@ -529,6 +560,20 @@ BLOB.prototype.validate = function(value) {
...
+    value = new Buffer(value);
+  }
+  var hex = value.toString('hex');
+
+  return this.$hexify(hex);
+};
+
+BLOB.prototype.$hexify = function (hex) {
+  return "X'" + hex + "'";
+};
```
mickhansen: Should `$hexify` not handle `toString('hex')`?

mickhansen:
I can't seem to locate exactly where generic types are swapped for database specific types, can you point me to that?

lib/data-types.js
janmeier: Good point

janmeier:
Here https://github.com/sequelize/sequelize/blob/datatypes/lib/sequelize.js#L1207

mickhansen:
@janmeier Ah, so that didn't change.

Not sure i fully comprehend the entire changeset but it seems well reasoned and i didn't spot any obvious mistakes.

LGTM, lets rebase and i'll merge.

janmeier:
@mickhansen Rebased 

Still a couple of loose ends to tie up with regards to documentation, and a couple more issues that couple be solved by this, but let's merge it and I'll look at the rest later

mickhansen:
@janmeier needs another one ;p

janmeier:
image

(Done)

janmeier:
<3

hendrul:
Up/Down mecanism for datatypes is not included in the docs yet?, I'm having troubles with mysql when parsing a BLOB from the database, it triggers my parse method even for TEXT fields and is not passing a Buffer as value as I expected, instead, it is passing a Field object from the mysql driver on which I have to call value.buffer() in order to get the buffer.

PR #3842, opened 2015-05-31
WIP: Add dialect version parameter

BridgeAR:
This is an initial commit for adding specific dialect version functions

I know there are some problems with old mssql versions and the data-types right now and there's probably more (So please leave a comment if you remember something else :) ).

Besides that: I think it would be nicer to use `SELECT version();` instead of using the parameter. So I think it would be best to read the version and use that. In case of the `standard_conforming_strings` this has a shortcoming, since we have to set that after we know the version.

PR #4192, opened 2015-07-26
feat(version) Automatically determine the database version when creat…

janmeier:
…ing the first connection. Re-roll of #3842

Should be good to go, just want to get a quick review on it :)

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -184,11 +184,36 @@ ConnectionManager.prototype.getConnection = function(options) {
   var self = this;
   options = options || {};
 
-  return new Promise(function (resolve, reject) {
-    self.pool.acquire(function(err, connection) {
-      if (err) return reject(err);
-      resolve(connection);
-    }, options.priority, options.type, options.useMaster);
+  var promise;
+  if (this.sequelize.options.databaseVersion === 0) {
+    if (this.versionPromise) {
+      promise = this.versionPromise;
+    } else {
+      promise = this.versionPromise = self.$connect(self.config).then(function (connection) {
+        var _options = {};
+        _options.transaction = { connection: connection }; // Cheat .query to use our private connection
+        _options.logging = function () {};
+        _options.logging.__testLoggingFn = true;
+
+        return self.sequelize.databaseVersion(_options).then(function (version) {
+          self.sequelize.options.databaseVersion = version;
+          self.versionPromise = undefined;
+
+          self.$disconnect(connection);
+        });
+      });
+    }
+  } else {
+    promise = Promise.resolve();
+  }
+
+  return promise.then(function () {
+    return new Promise(function (resolve, reject) {
+      self.pool.acquire(function(err, connection) {
+        if (err) return reject(err);
+        resolve(connection);
+      }, options.priority, options.type, options.useMaster);
+    });
   });
 };
 ConnectionManager.prototype.releaseConnection = function(connection) {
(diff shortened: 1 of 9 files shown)


lib/dialects/abstract/connection-manager.js
```diff
@@ -184,11 +184,28 @@ ConnectionManager.prototype.getConnection = function(options) {
...
+        _options.logging.__testLoggingFn = true;
+
+        return self.sequelize.databaseVersion(_options).then(function (version) {
+          self.sequelize.options.databaseVersion = version;
+
+          self.$disconnect(connection);
+        });
+      });
+    }
+  }).then(function () {
```
BridgeAR: A Promise should always [return something] (https://github.com/spion/thenlint) in the middle of a promise chain so I recommend to rewrite this to:
```js
var promise;
if (self.sequelize.options.databaseVersion === 0) {
  promise = self.$connect(self.config)...
} else {
  promise = Promise.resolve();
}
return promise.then(function () { ... });
```
That's also a bit less indentation (and you use one promise less). Some might prefere the other way to write this because of the style though. In that case it would be good to add a `return;` behind the if.

BridgeAR:
LGTM with one small comment

lib/dialects/abstract/connection-manager.js
```diff
@@ -184,11 +184,28 @@ ConnectionManager.prototype.getConnection = function(options) {
   var self = this;
   options = options || {};
 
-  return new Promise(function (resolve, reject) {
-    self.pool.acquire(function(err, connection) {
-      if (err) return reject(err);
-      resolve(connection);
-    }, options.priority, options.type, options.useMaster);
+  return Promise.resolve().then(function () {
+    if (self.sequelize.options.databaseVersion === 0) {
```
mickhansen: This will race, won't it?

lib/dialects/abstract/connection-manager.js
mickhansen: `return;` would not be returning something, i don't really see the importance in any case, but up to @janmeier

mickhansen: Although i do generally prefer short circuit code like your second example, although that example will still race.

lib/dialects/abstract/connection-manager.js
janmeier: If several concurrent calls request a transaction yes. The other alternative is to block once the first call is sent, but it would make the code slightly more complicated. I don't really think its a big problem that call getversion `maxConnections` number of times :)

mickhansen: Just if you have several concurrent calls i assume? Each will request a new connection from the pool.
Blocking would not be too tricky, save the connect + databaseversion promise in a variable, if var is set use that to resolve.

lib/dialects/abstract/connection-manager.js
BridgeAR: `return;` is sugar for `return undefined;`. So it'll return `undefined` and not nothing ;). I personally use `return void 0;` but that's just something I got used to quite a while ago. If you do not return something in the middle of a promise chain the promise thinks it's done and has to realize that there's another then chained.

lib/dialects/abstract/connection-manager.js
janmeier: Yep, nothing to do with transaction :)

janmeier:
Updated, thanks for the feedback @mickhansen and @BridgeAR  :)

Issue #4194, opened 2015-07-27
findAndCountAll crashes with a query which runs in findAll


slawo:
This example will run:
````
User.findAll({
    where: sequelize.literal('profiles.userId IS null'),
    include: [
    { model: Profile}
    ]
  })
````
This one will not:
````
User.findAndCountAll({
    where: sequelize.literal('profiles.userId IS null'),
    include: [
    { model: Profile}
    ]
  })
````

janmeier:
How will it not run?

slawo:
When running :
````
User.findAndCountAll({
    where: sequelize.literal('profiles.userId IS null'),
    include: [Profile]
  })
````
I get the following error:
````
Executing (default): SELECT count(*) AS `count` FROM `users` AS `user` WHERE profiles.userId IS null;
Unhandled rejection SequelizeDatabaseError: SQLITE_ERROR: no such column: profiles.userId
    at Query.formatError (.../node_modules/sequelize/lib/dialects/sqlite/query.js:242:14)
    at .../node_modules/sequelize/lib/dialects/sqlite/query.js:47:29
    at replacement (.../node_modules/sqlite3/lib/trace.js:20:31)
    at Statement.errBack (.../node_modules/sqlite3/lib/sqlite3.js:16:21)
````

It seems the whole include is ignored during count.

slawo:
I have just updated to 3.5
Before I had something along those lines (`sql-string.js:144 error undefined is not a function`)

slawo:
@janmeier, when running :
````
User.findAndCountAll({
    where: sequelize.literal('profiles.userId IS null'),
    include: [Profile]
  })
````
I get the following error:
````
Executing (default): SELECT count(*) AS `count` FROM `users` AS `user` WHERE profiles.userId IS null;
Unhandled rejection SequelizeDatabaseError: SQLITE_ERROR: no such column: profiles.userId
    at Query.formatError (.../node_modules/sequelize/lib/dialects/sqlite/query.js:242:14)
    at .../node_modules/sequelize/lib/dialects/sqlite/query.js:47:29
    at replacement (.../node_modules/sqlite3/lib/trace.js:20:31)
    at Statement.errBack (.../node_modules/sqlite3/lib/sqlite3.js:16:21)
````

It seems the whole include is ignored during count. As if a different query was ran.

I have just updated to 3.5
Before I had something along those lines (`sql-string.js:144 error undefined is not a function`)

janmeier:
[...]

PR #4195, opened 2015-07-27
Solves an issue with left joins being opptimised away in some queries.

slawo:
- adds the possibility to force a left join on queries,
fixes: #4194

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sequelize",
   "description": "Multi dialect ORM for Node.JS/io.js",
-  "version": "3.5.0",
+  "version": "3.5.1",
   "author": "Sascha Depold <sascha@depold.com>",
   "contributors": [
     {


janmeier:
Yeah, looks like a reasonable fix - needs api docs and regression tests though :)

janmeier:
I most admit I'm not 100% sure why we actually remove the includes, other than for performance (which is in itself a valid reason, I just don't remember if there is a deeper reason :D)

mickhansen:
I'm not totally happy with the approach here.
We remove them for performance and because they introduce complexities when you get to subquery stuff etc.

janmeier:
I was also thinking, since the only place where we do this removal of includes is findAndCount, so we might just introduce an option to not do that?

slawo:
I believe this is a reasonable fix as the includes are indeed required. And the requirement is unambiguous as to its purpose.

But I am not familiar with the insides of sequelize.

slawo:
@janmeier @mickhansen Either way, I would be happy to have an option to avoid the optimisation.

This way seemed to be the most flexible with out adding any new keywords. But I could have a new keyword (i.e. `keep`) I would be happy with that.

slawo:
@janmeier @mickhansen apparently the forced removal of includes from queries generates other bugs as well for example #4200.

Could it be possible to turn it off during some queries. While I appreciate the importance of speed optimisation I am currently working on a tool used internally by a few people and my primary objective is reliability not performance at this point.

Is there a way to switch those optimisations off entirely in the project?

janmeier:
@slawo #4200 is not directly related to this. In the case of limit we _have to_ remove the includes, because we use a subquery. We need a subquery without joins in order to limit, otherwise the limit is wrong (in that case we would be limiting the number of joined rows, instead of the number of parent rows)

slawo:
Ouch!

Why is the query not simply wrapped in a 
SELECT * FROM (query) LIMIT 2; ?

mickhansen:
@slawo Because of the reason @janmeier just explains, it won't work with cartesian products since it will limit the total amount of rows (which will be multiplied because of joins) not the amount of rows in the main table (which is the intent).

PR #3364, opened 2015-03-17
Add support for WHERE LIKE ANY queries

3ch01c:
See #3332.

PR #4203, opened 2015-07-29
Added parenthese to LIKE ANY (ARRAY...) query in Postgres

3ch01c:
Addresses issue noted in #3364

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -2113,10 +2113,6 @@ var QueryGenerator = {
           if (comparatorMap[key]) {
             comparator = comparatorMap[key];
             value = item;
-            if (_.isPlainObject(value) && value.$any){
-              comparator += ' ANY';
-              value = value.$any;
-            }
           }
         });
       }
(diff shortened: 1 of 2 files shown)


lib/dialects/abstract/query-generator.js
```diff
@@ -2127,7 +2127,11 @@ var QueryGenerator = {
         comparator = 'IS NOT';
       }
 
-      value = this.escape(value, field);
+      if (_.contains(comparator, 'ANY')){
+        value = '(' + this.escape(value, field) + ')';
+      } else {
+        value = this.escape(value, field);
+      }
```
mickhansen: Please move the fix to where we handle `$any`, don't do string matching :)

lib/dialects/abstract/query-generator.js
```diff
@@ -2127,7 +2123,13 @@ var QueryGenerator = {
...
       }
 
-      value = this.escape(value, field);
+      if (_.isPlainObject(value) && value.$any){
+        comparator += ' ANY';
+        value = value.$any;
+        value = '(' + this.escape(value) + ')';
+      } else {
+        value = this.escape(value, field);
+      }
```
mickhansen: https://github.com/3ch01c/sequelize/blob/likeany/lib/dialects/abstract/query-generator.js#L2092

3ch01c: @mickhansen I didn't put it there because I think that condition only catches ANY not LIKE ANY.  I could write a new condition to catch LIKE/NOTLIKE/ILIKE/NOTILIKE and handle it there, but it seems like it would be a lot of duplicated code from the catch-all condition at the end. Or maybe I missing something else?

sushantdhiman:
Hi @janmeier , Its already fixed in https://github.com/sequelize/sequelize/pull/5840

janmeier:
@sushantdhiman Lovely! (:

PR #4209, opened 2015-07-29
Throw the original error from authenticate

seanjh:
When the authenticate method catches errors, it wraps those errors in a new Error() call, rather than just throwing the original error. This gobbles up the original name and message into the new error's message parameter, making it more difficult to make use of the error result than necessary.

What's currently happening:
```javascript
var seqError = new sequelize.Error('YOLO');
try {
  throw new Error(seqError);
} catch (e) {
  if (e instanceof sequelize.Error) {
    console.log('Hooray, I can tell this is a Sequelize error!');
  } else {
    console.log('Bummer. I cannot easily distinguish this from a generic error.');
  }
}
```

```
Bummer. I cannot easily distinguish this from a generic error.
```

What would be preferred:

```javascript
var seqError = new sequelize.Error('YOLO');
try {
  throw seqError;
} catch (e) {
  console.log(util.inspect(e));
  if (e instanceof sequelize.Error) {
    console.log('Hooray, I can tell this is a Sequelize error!');
  } else {
    console.log('Bummer. I cannot easily distinguish this from a generic error.');
  }
}
```

```
Hooray, I can tell this is a Sequelize error!
```

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -932,7 +932,7 @@ Sequelize.prototype.drop = function(options) {
  */
 Sequelize.prototype.authenticate = function(options) {
   return this.query('SELECT 1+1 AS result', Utils._.assign({ raw: true, plain: true }, options)).return().catch(function(err) {
-    throw new Error(err);
+    throw err;
   });
 };
 
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -931,9 +931,7 @@ Sequelize.prototype.drop = function(options) {
  * @return {Promise}
  */
 Sequelize.prototype.authenticate = function(options) {
-  return this.query('SELECT 1+1 AS result', Utils._.assign({ raw: true, plain: true }, options)).return().catch(function(err) {
-    throw err;
-  });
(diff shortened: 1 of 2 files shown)


lib/sequelize.js
```diff
@@ -932,7 +932,7 @@ Sequelize.prototype.drop = function(options) {
  */
 Sequelize.prototype.authenticate = function(options) {
   return this.query('SELECT 1+1 AS result', Utils._.assign({ raw: true, plain: true }, options)).return().catch(function(err) {
```
janmeier: Not even sure we need a catch clause here

janmeier:
Yeah, makes sense :)

lib/sequelize.js
seanjh: Even better!

mickhansen: catching to wrap an error in an error, the original quote was definitely wonky.

janmeier: ping @seanjh Can you remove this cath completely, then I'll merge :)

janmeier:
Thanks! :)

PR #4220, opened 2015-07-30
Slightly enhance styling of API Docs tables

gustavpursche:
Before:
screen shot 2015-07-30 at 13 20 31

After:
screen shot 2015-07-30 at 13 21 08

This breaks the tables on very small screens, but they were already unusable before. To make the page fully responsive, could be tackled with more time.

### DIFF
--- a/docs/css/custom.css
+++ b/docs/css/custom.css
@@ -25,6 +25,10 @@ tr:nth-child(2n) {
   background-color: #f8f8f8;
 }
 
+.rst-content table.docutils td {
+  vertical-align: top;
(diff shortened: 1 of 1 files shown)


janmeier:
I like the idea, but this breaks when the default value is very long :)

image

gustavpursche:
Mh, that's true. Ok, so there is no quickfix, since simply adding a `min-width` to the second column breaks the layout. A possibility could be to move the default values to a seperate column, so each column becomes a bite more narrow.

PR #4229, opened 2015-07-31
Update JSDoc of Model.prototype.findOrInitialize()

gustavohenke:
The `@return` tag says that this method returns `Promise<Instance>`, while the
real return contains also an `initialized` boolean value.

Even though this behavior is already described in the JSDoc, it's more evident if
it's in the `@return` tag. I relied on it and my tests were breaking because I didn't
read the full docs for that method!


janmeier:
Thanks :)

PR #4239, opened 2015-07-31
String validation fix

frantzmiccoli:
Validating if a value is a string using typeof fails if the string has been created using `new String` this fixes this.

```javascript
var test1 = 'my test';
var test2 = new String(10);

console.log(typeof test1); # string
console.log(typeof test2); # object

console.log(Object.prototype.toString.call(test1)); # [object String]
console.log(Object.prototype.toString.call(test2)); # [object String]
```

I also have added a new assertion to cover that case.


janmeier:
LGTM - thanks :)

PR #4254, opened 2015-08-04
Clone instance options 

gutobortolozzo:
Hi,
The other day I ran into a code that was using the same options object to make several calls to ```instance.save(options)```. Taking a look at ```save``` and some other methods, I found out that options are being modified inside methods like this: 

```javascript
  if (!options.fields) {
    if (this.isNewRecord) {
      options.fields = Object.keys(this.Model.attributes);
    } else {
      options.fields = _.intersection(this.changed(), Object.keys(this.Model.attributes));
    }

    options.defaultFields = options.fields;
  }
```

Facing this problem, I implemented some code to clone the options object from save and some other methods trying to avoid this behavior. What you guys think about this?

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -172,11 +172,12 @@ Instance.prototype.setDataValue = function(key, value) {
  * If key is given and a field or virtual getter is present for the key it will call that getter - else it will return the value for key.
  *
  * @param {String} [key]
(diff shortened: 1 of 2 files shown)


janmeier:
Fixed by https://github.com/sequelize/sequelize/pull/4023

PR #4260, opened 2015-08-05
check result[0] is exists before read it.

pentium100:
Some times the variable results will be a 0 size array. 
when read results[0][autoIncrementFieldAlias] will cause app crash. 
so we need to check it before we read it.


janmeier:
How can you have an insert query that doen't return anything? At least an insert coming from sequelize will always have `OUTPUT INSERTED` set.

Also, we need some regression tests

pentium100:
when we save a model with options: {returning: false}, then insert query while return nothing.

janmeier:
Ah, thats true :)

Just needs a regression test then

janmeier:
Ping @pentium100

janmeier:
Closing, no activity

Issue #4274, opened 2015-08-06
If you include a model with no primaryKey in a findAll query, only one instance is eager loaded


Americas:
Example
```javascript
var CD = sequelize.define("CD", {
    id    : {
        type         : Sequelize.UUID,
        field        : "cd_a",
        defaultValue : Sequelize.UUIDV4,
        primaryKey   : true,
    },
}, {
    tableName    : "test1",
    name         : {
        singular : "parent",
        plural   : "parents",
    },
});

var CQ = sequelize.define("CQ", {
    a : {
        type         : Sequelize.UUID,
        defaultValue : Sequelize.UUIDV4,
        field        : "cq_a",
        allowNull    : false,
        unique       : "uni",
    },
    b : {
        type         : Sequelize.STRING,
        field        : "cq_b",
    },
}, {
    tableName    : "test2",
    name         : {
        singular : "child",
        plural   : "children",
    },
});
CQ.removeAttribute("id");

CD.hasMany(CQ, {foreignKey : "cq_c"});
CQ.belongsTo(CD, { foreignKey : "cq_c", unique : "uni" });

sequelize
.sync({force: true})
.then(function () {
    return CD.create() // Create one CD instance
})
.then(function(result){
    return CQ.create({b : "one", cq_c : result.id }) // Create CQ instace one related to CD
    .then(function(){
        return CQ.create({b : "two", cq_c : result.id }); // Create CQ instace two related to CD
    })
})
.then(function(){

    return CD.findAll({ include : CQ }) // Get CD instance with CQ eager loaded
    .then(function(results){
        console.log(results[0].children.length); // length = 1
    })

})
.then(function () {
    console.log("done");
    process.exit()
}, function (err) {
    console.log(err);
    process.exit()
});
```

janmeier:
I doubt this usecase will ever be supported. Join de-deduplication relies heavily on primary keys, so I'm surprised it even works this far :)

janmeier:
When you have a field thats a UUID, why not mark it as the primarykey of that table?

Americas:
Because one of the columns can be null, and primary keys can't have nulls. In my use case I do not use the "id" attribute, but since it will solve this problem, I will probably add it just to solve this.

If I use the "getAssociationns" function, it returns both children, so I was wondering why the eager load option did not work.

janmeier:
[...]

PR #4278, opened 2015-08-07
groupJoinData checks UniqueKeys if no PrimaryKey is set

Americas:
When nesting associations, if the model doesn't have a primary key, the code will now check and see if there is at least one unique key, and if there is, it will use one to create the hash that identifies each different row of results.

This change is in response to #4274

### DIFF
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -149,7 +149,21 @@ var groupJoinData = function(rows, includeOptions, options) {
       }
       return keyPrefixMemo[key];
     }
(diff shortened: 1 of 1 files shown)


janmeier:
@mickhansen has done major optimization work on this part of the code, so I'll let him review it once he's back from vacation

mickhansen:
This lowers hash calculation perf a bit.
Common case is primary key of length 1, please reintroduce that check and then only fallback to loop if length is larger than 1.

Americas:
Added back the length === 1 case.

Issue #3824, opened 2015-05-29
Querying Postgres jsonb column contaning non-text type of fields cause "operator does not exist" exception


xwk:
Suppose model Store has a column called 'jsonb' which has type jsonb. Query like

```
Store.find({
      where: {
        jsonb: {
          enabled: true
        }
      }
```

will cause SequelizeDatabaseError: operator does not exist: text = boolean exception.  

The reason for that is Sequelize will translate the query into following SQL

SELECT "id", "jsonb", "createdAt", "updatedAt" FROM "stores" AS "store" WHERE ("store"."jsonb"**#>>**'{enabled}') = true LIMIT 1.

Notice #>> operator is used, which converts the jsonb type to text. The correct SQL should be 

SELECT "id", "jsonb", "createdAt", "updatedAt" FROM "stores" AS "store" WHERE ("store"."jsonb"**#>**'{enabled}') = **'true'::jsonb** LIMIT 1;

Full sample code as below

```
var Sequelize, Store, configs, sequelize;

Sequelize = require("sequelize");

configs = require("../configs");

sequelize = new Sequelize(configs.DB_NAME, configs.DB_USERNAME, configs.DB_PASSWORD, {
  host: configs.DB_HOSTNAME,
  dialect: 'postgres'
});

Store = sequelize.define('store', {
  jsonb: Sequelize.JSONB
});

Store.sync({
  force: true
}).then(function() {
  return Store.create({
    jsonb: {
      name: 'test store',
      enabled: true,
      grade: 3
    }
  }).then(function() {
    return Store.find({
      where: {
        jsonb: {
          enabled: true
        }
      }
    });
  })
```

mickhansen:
`("store"."jsonb"#>'{enabled}')::boolean = true` might work aswell? That's the technique we use for numbers

PR #4298, opened 2015-08-11
JSON: Cast key when using the equality operator

johanneswuerbach:
Fixed `operator does not exist` exception when using queries like:
```js
{
  'data.age': 4
}
```

Fixes: https://github.com/sequelize/sequelize/issues/3824

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1959,7 +1959,8 @@ var QueryGenerator = {
             , $key
             , $cast
             , $baseKey
-            , $tmp;
+            , $tmp
+            , castKey;
 
           if (path[path.length - 1].indexOf('::') > -1) {
             $tmp = path[path.length - 1].split('::');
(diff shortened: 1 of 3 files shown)


lib/dialects/abstract/query-generator.js
```diff
@@ -1979,25 +1980,31 @@ var QueryGenerator = {
 
           $baseKey = '('+$baseKey+')';
 
+          castKey = function ($item) {
```
janmeier: I was just about to comment on the non-DRY-ness but you were quicker than me ;)

johanneswuerbach: Yep, my bad. Committed an early version of this PR and not the refactored one .... :D

janmeier:
Remember to run jshint, you have an undefined variable

johanneswuerbach:
Sorry, fixed.

janmeier:
Cool, thanks :)

janmeier:
Not sure this fixes https://github.com/sequelize/sequelize/issues/3824 though, we need to add a case for casting to boolean

PR #4299, opened 2015-08-11
consistent results from instance.update/destroy and Model.update/dest…

alekbarszczewski:
…roy (respecting returning option).

WARNING: breaks BC with 3.x

``` javascript
instance.update({}, options)
// beforeUpdate: (instance, options)
// afterUpdate: (instance, options) - instance is null if not updated, instance contains fetched data if options.returning
.then(function(instance){}) // - instance is null if not updated, instance contains fetched data if options.returning

instance.destroy(options)
// beforeDestroy: (instance, options)
// afterDestroy: (instance, options) - instance is null if not deleted, instance contains fetched (deleted) data if options.returning
.then(function(instance){}) // - instance is null if not deleted, instance contains fetched (deleted) data if options.returning

Model.update({}, options)
// beforeBulkUpdate (options)
// afterBulkUpdate (options)
.then(function([rows, instances]){}) // instances not present if result array if !options.returning

Model.destroy({}, options)
// beforeBulkDestroy (options)
// afterBulkDestroy (options)
.then(function([rows, instances]){}) // instances not present if result array if !options.returning
```

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -395,9 +395,9 @@ var QueryGenerator = {
     }
 
     if (options.limit) {
-      query = 'DELETE FROM <%= table %> WHERE <%= primaryKeys %> IN (SELECT <%= primaryKeysSelection %> FROM <%= table %><%= where %><%= limit %>)';
+      query = 'DELETE FROM <%= table %> WHERE <%= primaryKeys %> IN (SELECT <%= primaryKeysSelection %> FROM <%= table %><%= where %><%= limit %>)<%= returning %>';
     } else {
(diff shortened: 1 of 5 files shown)


janmeier:
Looks good, except for a few tests which need to be change to the new return format

However, we still haven't decided how we want to handle bc breaks - maintaining a separate branch turned out to be a PITA in the past ...

janmeier:
@alekbarszczewski so, finally master is now 4.0 - Are you up for re-applying this :) ?

janmeier:

suratovvlad:
@alekbarszczewski @janmeier 
Hi! What's about this feature for destroy?

PR #4302, opened 2015-08-11
Pass through connection mode options to sqlite

ysagal:
SQLite allows different connection modes (READONLY, READWRITE, CREATE). This PR allows passing through these options via `dialectOptions.mode` which was kind of expected in the first place. By default is sets the connection mode to sqlite's default which is READWRITE | CREATE.

Added some tests to verify the same. Did not update documentation since my expectation was that this should work via `dialectOptions` already.

Not sure how to get rid of query console logging:
```
Executing (default): CREATE TABLE foo (faz TEXT);
```

### DIFF
--- a/lib/dialects/sqlite/connection-manager.js
+++ b/lib/dialects/sqlite/connection-manager.js
@@ -4,8 +4,7 @@ var AbstractConnectionManager = require('../abstract/connection-manager')
   , ConnectionManager
   , Utils = require('../../utils')
   , Promise = require('../../promise')
-  , sequelizeErrors = require('../../errors')
-  , sqlite3 = require('sqlite3');
+  , sequelizeErrors = require('../../errors');
 
 ConnectionManager = function(dialect, sequelize) {
   this.sequelize = sequelize;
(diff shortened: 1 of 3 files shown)


ysagal:
Not sure why postgres-native is failing, but sqlite node.js v0.10 is failing because it doesn't yet have `fs.access`. Should I put in a workaround for checking file existence? E.g. `fs.open`?

lib/dialects/sqlite/connection-manager.js
```diff
@@ -30,16 +31,25 @@ ConnectionManager.prototype.getConnection = function(options) {
...
+  }
 
   return new Promise(function (resolve, reject) {
-    self.connections[options.inMemory || options.uuid] = new self.lib.Database(self.sequelize.options.storage || self.sequelize.options.host || ':memory:', function(err) {
-      if (err) {
-        if (err.code === 'SQLITE_CANTOPEN') return reject(new sequelizeErrors.ConnectionError(err));
-        return reject(new sequelizeErrors.ConnectionError(err));
+    self.connections[options.inMemory || options.uuid] = new self.lib.Database(
+      self.sequelize.options.storage || self.sequelize.options.host || ':memory:',
+      options.readWriteMode || (sqlite3.OPEN_READWRITE | sqlite3.OPEN_CREATE), // default mode
```
janmeier: You should be able to use `this.lib` here instead, so we don't have to require sqlite3 at the top of the file

The reason we do it this way is, that it allows people to specify another library, which is api compatible (for example spatialite), using  `dialectModulePath`

janmeier:
Looking good, just one small comment

janmeier:
Nice work, thanks! :)

PR #4316, opened 2015-08-13
Fix destroy should not affect previous instances

nunofgs:
This PR fixes an issue when calling destroy() multiple times on the same instance. Previously the "deletedAt" column would be updated with the newer destroy() call.


lib/instance.js
```diff
@@ -817,18 +817,23 @@ Instance.prototype.destroy = function(options) {
       return this.Model.runHooks('beforeDestroy', this, options);
     }
   }).then(function() {
-    var where;
+    var primaryKeys = this.Model.primaryKeyAttributes;
+    var where = {};
+
+    for (var i = 0; i < primaryKeys.length; i++) {
+      where[this.Model.rawAttributes[primaryKeys[i]].field] = this.get(primaryKeys[i], { raw: true });
+    }
```
mickhansen: You can use `this.where()` instead.

lib/instance.js
```diff
@@ -817,18 +817,23 @@ Instance.prototype.destroy = function(options) {
...
     if (this.Model._timestampAttributes.deletedAt && options.force === false) {
-      this.setDataValue(this.Model._timestampAttributes.deletedAt, new Date());
-      return this.save(_.extend({}, options, {hooks : false}));
+      var field = this.Model.rawAttributes[this.Model._timestampAttributes.deletedAt].field || this.Model._timestampAttributes.deletedAt;
+      var values = {};
+
+      values[field] = new Date();
+      where[field] = null;
+
+      return this.sequelize.getQueryInterface().update(this, this.Model.getTableName(options), values, where, _.defaults({ hooks: false }, options));
```
mickhansen: Should probably still `set` a local `deletedAt` value?

nunofgs: Sounds good.

lib/instance.js
nunofgs: Good idea. I had to fix a small bug in the `where` function though. Please take a look.

nunofgs:
@mickhansen Updated the PR.

lib/instance.js
```diff
@@ -127,7 +127,7 @@ Instance.prototype.where = function() {
   var where;
 
   where = this.Model.primaryKeyAttributes.reduce(function (result, attribute) {
-    result[attribute] = this.get(attribute, {raw: true});
+    result[this.Model.rawAttributes[attribute].field] = this.get(attribute, {raw: true});
```
mickhansen: I'm not too happy about that, it will confuse users who use `.where()` expecting attributes and not fields like we normally use.

You can use `Utils.mapOptionFieldNames` (https://github.com/sequelize/sequelize/blob/master/lib/utils.js#L80) instead in the destroy code.

nunofgs: Done.

lib/instance.js
```diff
@@ -897,22 +899,11 @@ Instance.prototype.restore = function(options) {
  * @return {Promise<this>}
  */
 Instance.prototype.increment = function(fields, options) {
-  var identifier = this.where()
+  var identifier = Utils.mapOptionFieldNames({ where: this.where() }, this.Model).where
```
mickhansen: I know i'm just nitpicking now, but i imagine we might have to repeat this code more than these two times, perhaps we could provide a second `fields`|boolean argument to `where()` that will apply this code?

janmeier:
Mapping of field names moved to `this.where`

https://github.com/sequelize/sequelize/commit/a15d4233f8f11a12ce5295c998fe7917222a0b05

janmeier:
Thanks @nunofgs ! :)

Issue #4167, opened 2015-07-21
Order doesn't use fields aliases


atorkhov:
In order directive for queries it doesn't take in account of field aliases.

JSteunou:
might be link to #4103 ?

janmeier:
Not directly, but the root cause is the same, yes :)

JSteunou:
Hey I saw *fields aliases*, I jumped on it :p

subjectix:
Yeah... this bug is annoying -_- 
And I can't find the origin to make a PR. can't track where order in the options end up.

faq885:
Why is this closed? i am having still this issue!

mickhansen:
It was closed via the PR fixing this @faq885
However it is a 3 year old issue, so if you are "still" having the issue then i commend your patience, if it's a recent issue maybe it's a new bug.

PR #4318, opened 2015-08-13
Respect custom field names when ordering

johanneswuerbach:
Before:
```
SELECT "User"."id", "User"."last_name" AS "lastName", "User"."createdAt", "User"."updatedAt", 
"User"."CompanyId", "Company"."id" AS "Company.id", "Company"."combined_rank" AS 
"Company.combinedRank", "Company"."createdAt" AS "Company.createdAt", "Company"."updatedAt" 
AS "Company.updatedAt" FROM "Users" AS "User" LEFT OUTER JOIN "Companies" AS "Company" 
ON "User"."CompanyId" = "Company"."id" 
ORDER BY "Company"."rank" ASC, "User"."lastName" DESC 
LIMIT 5;
```

After:
```
SELECT "User"."id", "User"."last_name" AS "lastName", "User"."createdAt", "User"."updatedAt", 
"User"."CompanyId", "Company"."id" AS "Company.id", "Company"."combined_rank" AS 
"Company.combinedRank", "Company"."createdAt" AS "Company.createdAt", "Company"."updatedAt" 
AS "Company.updatedAt" FROM "Users" AS "User" LEFT OUTER JOIN "Companies" AS "Company" 
ON "User"."CompanyId" = "Company"."id" 
ORDER BY "Company.combinedRank" ASC, "lastName" 
DESC LIMIT 5;
```

When this approach seems good to you I'll make the other tests pass.

Fixes https://github.com/sequelize/sequelize/issues/4167

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
 # Next
 - [FIXED] Show indexes query on Postgres fails to return functional indexes [#3911](https://github.com/sequelize/sequelize/issues/3911)
+- [FIXED] Custom field names when ordering [#4167](https://github.com/sequelize/sequelize/issues/4167)
 - [FIXED] Custom field names in json queries
 - [FIXED] JSON cast key using the equality operator. [#3824](https://github.com/sequelize/sequelize/issues/3824)
 - [FIXED] Map column names with `.field` in scopes with includes. [#4210](https://github.com/sequelize/sequelize/issues/4210)
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -801,7 +801,7 @@ var QueryGenerator = {
       }
 
       // add 1st string as quoted, 2nd as unquoted raw
-      var sql = (i > 0 ? this.quoteIdentifier(tableNames.join('.')) + '.' : (Utils._.isString(obj[0]) && parent ? this.quoteIdentifier(parent.name) + '.' : '')) + this.quote(obj[i], parent, force);
+      var sql = (i > 0 ? this.quoteIdentifier(tableNames.join('.') + '.'  + this.quote(obj[i], parent, force)) : this.quote(obj[i], parent, force));
       if (i < len - 1) {
         if (obj[i + 1]._isSequelizeMethod) {
           sql += this.handleSequelizeMethod(obj[i + 1]);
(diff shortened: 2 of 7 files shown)


janmeier:
`"User"."lastName" DESC ` -> `"lastName" DESC` seems wrong to me?

johanneswuerbach:
I don't think so as `"User"."last_name" AS "lastName"` is alias in the SELECT. I basically changed it to use the results of `AS` instead of underlying fields.

In our case the outer-query has actually no access to the inner table name:
```sql
SELECT "DraftAsset".*, "AssetFields"."id" AS "AssetFields.id", "AssetFields"."asset_id" AS 
"AssetFields.assetId", .... FROM (
  SELECT "DraftAsset"."id", ..., "DraftAsset"."external_id" AS "externalId", ... FROM "draft_assets" AS   
  "DraftAsset" WHERE ("DraftAsset"."deleted_at" IS NULL)
   ORDER BY "externalId" asc LIMIT 100
) AS "DraftAsset" 
LEFT OUTER JOIN "draft_asset_fields" AS "AssetFields" ON "DraftAsset"."id" = "AssetFields"."asset_id" 
ORDER BY "externalId" asc;
```

janmeier:
Ah yeah I remember we had a similar issue related to columns being aliassed in the result from subqueries.

I'll leave this one for @mickhansen to review, but LGTM

johanneswuerbach:
ok, I'll look into making everything pass now :-)

johanneswuerbach:
@mickhansen could you provide some guidance? We would need support for sorting based on alias, but the current implementation would be a BC.

Do you have any ideas using a different ordering syntax or something like that?

mickhansen:
You shouldn't need to break BC atm. A quick fix would to fix the mapping in `mapOptionFieldNames` or whatever it's called. A more deep fix would need to know if the include is inside a subquery or not.

johanneswuerbach:
Hm, ordering is not doing any mapping, just appending the tableName atm?

Currently its a BC as looking for `last_name` wouldn't work any longer when its alias to `lastName`.

mickhansen:
https://github.com/sequelize/sequelize/blob/master/lib/utils.js#L80 should be expanded to work with `options.order`

johanneswuerbach:
Sorry, I won't have time to finish this PR and we stopped using aliased fields.

PR #4321, opened 2015-08-13
Fix removeColumnQuery for mssql dialect

LeekAnarchism:
* As per the syntax tree at:

  https://msdn.microsoft.com/en-gb/library/ms190273.aspx

  the correct syntax for removing a column is:
    ALTER TABLE users DROP COLUMN name;
  instead of:
    ALTER TABLE users DROP name;
  which is instead the syntax to drop the 'name' constraint.

----
I noticed this wasn't working for our migrations against an MSSQL DB.

I haven't implemented any tests yet -- I'd like to check this change is sound in principle before going any further.

It looks like I could implement some tests at:
>test/integration/dialects/mssql/query-generator.test.js

(note that it looks like there are no mssql integration tests yet)

Are there any other places I should add tests?

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -170,7 +170,7 @@ var QueryGenerator = {
   },
 
   removeColumnQuery: function(tableName, attributeName) {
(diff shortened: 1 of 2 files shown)


janmeier:
A unit test is fine - thanks! :)

PR #4328, opened 2015-08-14
Fix Postgres showIndex not returning functional indices

sorin:


janmeier:
Good work @sorin ! :)

By the way, you don't need to close the pull request and open a new one each time you make changes. Just push to the same branch and the PR will automagically be updated and the tests run again

sorin:
Thanks for accepting PRs so fast! I'm still figuring out how to run the tests on my local so I was abusing the sequelize/master CI. But now I realized I can actually run travis on my fork, so it should be good in the future.

janmeier:
Awesome! :)

Issue #4253, opened 2015-08-04
increment and decrement should support RETURNING * in postgres


jungseob86:
```
sequelize.define('Model', { version: Sequelize.INTEGER });
var modelFixture = {version: 1};

Model.create(modelFixture)
  .then(function(model) {
    return model.increment('version');
  })
  .then(function(model) {
    console.log(model.toJSON())
    done();
  })
  .catch(function(err) {
    console.log({err:err});
    done();
  });
```

in 3.0.0 result is
```
{ version: 2 }
```

in 3.5.1 result is
```
{ version: 1 }
```

so, I try to reload instance.
```
Model.create(modelFixture)
  .then(function(model) {
    return model.increment('version');
  })
  .then(function(model) {
    return model.reload();
  })
  .then(function(model) {
    console.log(model.toJSON())
    done();
  })
  .catch(function(err) {
    console.log({err:err});
    done();
  });
```
then, result is
```
{ version: 2 }
```

Oh, thanks! :)

But my question is, why do i need to reload the instance in 3.5.1, when it worked without a reload in the previous version? I can't find any reason in changelog.

janmeier:
Are you using postgres? In that case, try passing `returning: true`, which will return the updated value. 

Otherwise, I really don't know how that worked before. It has always been mentioned in the API docs that you need to do a `reload` to get the new value

jungseob86:
@janmeier Yes, I am using postgres. So I try passing `{returning: true}` option like `instance.increment(['field'], {returning: true})`. But It wasn't return the updated value too. Is that right?

janmeier:
Hmm yeah, the query is correct, but it seems we are not acually handling the returned values and putting them into the model - tagging this as a feature request, but as always it probably gonna be fixed faster if you look into it yourself. I'll be happy to offer guidance, shouldn't be a lot of work required

jungseob86:
@janmeier Thanks! I'll try!

jungseob86:
@janmeier I saw few codes.

Bibernull:
Have been using sequelize for quite a while, and increment/decrement always updated the instances.

not sure when/if this was changed, could be something else recently changed in how it handles RAW queries.

if we look at query-interface.js increment method, it sets the query type to RAW.

https://github.com/sequelize/sequelize/blob/master/lib/query-interface.js#L684

changing it to UDPATE, successfully updates the instance.

janmeier:
[...]

PR #4343, opened 2015-08-17
use QueryTypes.UPDATE instead of RAW when incrementing a value of an …

Bibernull:
…instance #4253

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -681,7 +681,7 @@ QueryInterface.prototype.increment = function(instance, tableName, values, ident
 
   options = options || {};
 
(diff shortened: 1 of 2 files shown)


janmeier:
LGTM, but needs an integration test. `if dialect.supports.returning check that values are updated

Bibernull:
not sure if the failing test has anything to do with increment?

janmeier:
Merged, thanks! :) https://github.com/sequelize/sequelize/commit/0ed07d90abc7612053c60760ee58c64a33153cfc

janmeier:
Nope, the failure is unrelated - haven't managed to track down why that test keeps failing

PR #4339, opened 2015-08-17
Revert "Update dependencies"

janmeier:
Reverts sequelize/sequelize#4332

Appearently travis did not properly test the updated deps because of node_modules caching, so I'll have to revert this

@josephpage Didn't your experience any problems when running the tests locally?

PR #4347, opened 2015-08-18
Stop caching modules on travis

johanneswuerbach:
The cache hides test failures when dependencies are updated.

Related https://github.com/sequelize/sequelize/pull/4339

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,6 @@ node_js:
 
 sudo: false
 
-cache:
-  directories:
-    - node_modules
-
 env:
   - DIALECT=mysql
   - DIALECT=postgres


mickhansen:
Will result in an overall slowdown, not sure we'll want that.

johanneswuerbach:
Yes, it adds ~30s per job, but would ensure that sequelize tests are passing using the latest prod and dev dependencies.

janmeier:
@johanneswuerbach Unfortunately its more like 30 - 300seconds (https://travis-ci.org/sequelize/sequelize/jobs/76117438), depending on the latency between travis and the particular instance running the job. Given how many travis builds we have, I don't think its a good idea to add such a potentially large latency.

Perhaps opening an issue with travis instead would be a better idea

johanneswuerbach:
Hm, looks like you picked the worst one: https://travis-ci.org/sequelize/sequelize/builds/76117437 the others are fairly reasonable.

> opening an issue with travis

they just cache the folder, npm is doing the rest and `npm install` usually never updates to a new version when a compatible version was found, which easily leads to outdated deps and hard to debug behaviour.

josephpage:
Caching on travis works fine for node_modules. `npm install` updates packages as needed.

The problem for #4339 was a regression in validator@4.0.3 which was released only few hours after my commit.

In older to solve the problem of old packages in cache we may use `npm update` [reference](https://docs.npmjs.com/cli/update)

janmeier:
@josephpage Alright, thanks for clearing that up - Then I also better understand why the error was triggered after my merge (because it installed a newer version fo the deps than was used to test your pr)

PR #4348, opened 2015-08-18
Postgres queryInterface.createTable creates ENUM types for columns it doesnt actually create

sorin:


test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -15,6 +16,30 @@ if (dialect.match(/^postgres/)) {
...
+        return this.queryInterface.dropTable('Group').then(function () {
+          return self.queryInterface.createTable('Group', {
+            enum1: {type: DataTypes.ENUM, values: ['1', '2']}
+          });
+        });
+      });
+
+      it('should only create enum types for columns in creates', function () {
+        var self = this;
+        return self.queryInterface.createTable('Group', {
```
janmeier: Why are you calling createTable  twice  - isn't the original issue about addColumn?

sorin: So here's the breakdown of my problem again, sorry if it sounds too complicated, it really is just a way to add a new enum column to an existing model.

- create new model Group with just enum1
  - run sequelize.sync() 
    - calls Group.createTable()
      - creates enum1 type
- add enum2 to the model and write a migration to add Group.enum2
  - run sequelize.sync()
    - calls Group.createTable()
      - **creates enum2 type although it did NOT create the enum2 column**
- run sequelize-cli db:migrate
  - calls addColumn(enum2)
    - crashes while trying to create enum2 type because it's already created

The test above is what I propose fixing, a simplified version of the above with just the 2 createTable statements. I believe it's incorrect that createTable() will create an enum type but not actually create a column.

The other thing that we can fix (not addressed in this PR) is the very last bullet above - make addColumn not crash when trying to create an enum type that already exists- unfortunately Postgres does not support CREATE TYPE IF NOT EXISTS (that would have been the cleanest solution).

sorin:
@janmeier I can write the code if you think this is a good idea ...

janmeier:
@sorin Sorry for leaving you hanging on this one!

I think your approach sounds reasonable - we should only create the enum if the table does exist already. A PR for that feature would be welcome

janmeier:

PR #4350, opened 2015-08-19
Connection hook for mysql

webfacebg:
This allows to execute an action on mysql connection. For example:

var sequelize = new Sequelize(
    process.env.MYSQL_DATABASE,
    process.env.MYSQL_USERNAME,
    process.env.MYSQL_PASSWORD, {
	host: process.env.MYSQL_HOST || 'localhst',
	onConnect: function (connection) {
	    connection.query("SET NAMES utf8;");
	}
    });


lib/dialects/mysql/connection-manager.js
```diff
@@ -69,7 +69,8 @@ ConnectionManager.prototype.connect = function(config) {
 
         return;
       }
-
+	if (config.onConnect)
+	    config.onConnect(connection);
```
mickhansen: Use brackets please. We mostly use single line if's for short-circuits.
There's also an indentation issue.

webfacebg:
Brackets added :-)

janmeier:
I like the idea, but don't think this is the best place to handle this. If you add it here instead https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/connection-manager.js#L228 it will apply to all dialects.

This also needs test and some documentation

I like the simplicity of just passing an option to the constructor, but on the other hand I think it would be better if this used the same API as the rest of our hooks system. What do you think @mickhansen ?

webfacebg:
Please have a look at the last changes. I moved the "hook" to the abstract connection-manager. For some reason the onConnect parameter is not passed to the connection-manager, so I hat to put it inside the dialectParams.

lib/dialects/abstract/connection-manager.js
```diff
@@ -73,7 +73,13 @@ ConnectionManager.prototype.initPools = function () {
       name: 'sequelize-connection',
       create: function(callback) {
         self.$connect(config).nodeify(function (err, connection) {
-          callback(err, connection); // For some reason this is needed, else generic-pool things err is a connection or some shit
+	  // Hook something on connection
+	  if (config.dialectOptions && config.dialectOptions.onConnect && typeof(config.dialectOptions.onConnect) === 'function') {
+	      // Do not forget to call the callback
+	      config.dialectOptions.onConnect(connection, err, callback);
```
janmeier: I guess this should also only be called if `!err`

lib/dialects/abstract/connection-manager.js
```diff
@@ -73,7 +73,13 @@ ConnectionManager.prototype.initPools = function () {
...
+	  } else {
+              callback(err, connection); // For some reason this is needed, else generic-pool things err is a connection or some shit
```
janmeier: This comment can be removed now ;)

janmeier:
Yeah, this looks good - there's some issues with the indentation though ;)  - should be 2 spaces

If you add API docs and tests this should be good to go

mickhansen:
I agree it should probably have the same API as the other hooks, this does not appear to support promises?

janmeier:
It doesn't support promises no, and I had the same though. But I guess its okay, since you're gonna have  to play around with callbacks anyway when you work directly with the connection

mickhansen:
@janmeier Hmm, just not a huge fan of having a piece of the API be that much different, bound to give us headaches.

janmeier:
Yeah, that's true. Ideally I'd like to use proper hooks for this, but thats a bit complicated too. If we add the hook to the  constructor (`Sequelize.addHook`) it will work fine, but you won't be able to add different hooks for different instances (would that be something someone wants??)

You could also add the hooks to the instance (`sequelize.addHook`), but that won't work if `pool.min` is set, because those connections would be initialized before the hook is addded.

Perhaps we can live with option 1, a global hook for all instances?

webfacebg:
Added the error check, comments removed and indentation corrected. Could you please where/how  to add Docs and tests?

mickhansen:
@janmeier well we already have the `hooks` option on `new Sequelize` which will add hooks to all models defined after that, wouldn't the same work for connection hooks i imagine?

janmeier:
@mickhansen Yeah thats true - thats probably the best way to go :)

janmeier:
@webfacebg Docs are in `docs/docs` - the ones in `/api` are auto generated from code comments.

For this I believe hand written docs would be fine.

As for tests, I believe `test/integration/sequelize.test` or `test/integration/dialects/abstract/` would be good places

janmeier:
ping @webfacebg

webfacebg:
В 11:54 -0700 на 08.10.2015 (чт), Jan Aagaard Meier написа:
> ping @webfacebg

Sorry. Just don't have the time to do it. It is on my list.

-- 
Anton Katsarov
WebFace — Your Web Face.
anton@webface.bg
http://www.webface.bg
Smolyan, Bulgaria
tel. +359 899 108 919

janmeier:
Superceeded by https://github.com/sequelize/sequelize/pull/5688

Issue #4353, opened 2015-08-20
Reload doesn't synchronize a null include


glutentag:
Hello,
When you reload an instance after having destroyed a hasOne association instance, instance.dataValues.association is set to null, but instance[association] still has the previous value.

In instance.js set method, _setInclude is called only if there is a value
```javascript
if (this.options && this.options.include && this.options.includeNames.indexOf(key) !== -1 && value) {
    this._setInclude(key, value, options); 
```

also, in _setInclude, the value is set to null only if it is an object with null primary key
```javascript
isEmpty = value && value[primaryKeyAttribute] === null;
```

PR #4354, opened 2015-08-20
Fix instance reload doesn't set to null a destroyed include

glutentag:
Fix issue #4353. Set to null a destroyed association after an instance reload.

### DIFF
--- a/test/integration/instance.test.js
+++ b/test/integration/instance.test.js
@@ -512,6 +512,36 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
         });
       });
     });
+
+    it('should set an association to null after deletion, 1-1', function() {
+      var Shoe = this.sequelize.define('Shoe', { brand: DataTypes.STRING })
+        , Player = this.sequelize.define('Player', { name: DataTypes.STRING });
+
+      Player.hasOne(Shoe);
+      Shoe.belongsTo(Player);
+
+      return Player.sync({force: true}).then(function() {
+        return Shoe.sync({force: true}).then(function() {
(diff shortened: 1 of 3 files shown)


test/integration/instance.test.js
```diff
@@ -512,6 +512,136 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
...
     });
+
+    it('should set an association to null after deletion, 1-1', function() {
+      var Shoe = this.sequelize.define('Shoe', { brand: DataTypes.STRING })
+        , Player = this.sequelize.define('Player', { name: DataTypes.STRING });
+
+      Player.hasOne(Shoe);
+      Shoe.belongsTo(Player);
+
+      return Player.sync({force: true}).then(function() {
```
janmeier: This is pretty waterfally - could be rewritten as:

```js


return this.sequelize.sync({force: true}).then(function() {
  return Shoe.create({
    brand: 'the brand',
    Player: {
      name: 'the player'
    }
  }, {include: [Player]});
}).then(function(shoe) {
  return Player.findOne({
    where: { id: shoe.Player.id },
    include: [Shoe]
  }).then(function(lePlayer) {
    expect(lePlayer.Shoe).not.to.be.null;
    return lePlayer.Shoe.destroy().return(lePlayer);
  }).then(function(lePlayer) {
    return lePlayer.reload();
  }).then(function(lePlayer) {
    expect(lePlayer.Shoe).to.be.null;
  });
});
```

I know its just copy-pasta from some of the methods above, but we like to keep a less waterfall approach in new code :)

janmeier:
LGTM, just need a bit of refactoring in test style.

Would you mind adding a changelog entry as well?

PR #4359, opened 2015-08-20
Fix for onDuplicateKeyUpdate in mysql

rfink:
Allow onDuplicateKeyUpdate in mysql for schemas that have field names defined.

### DIFF
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -167,7 +167,7 @@ var QueryGenerator = {
     return this.insertQuery(tableName, insertValues, rawAttributes, options);
   },
 
-  bulkInsertQuery: function(tableName, attrValueHashes, options) {
+  bulkInsertQuery: function(tableName, attrValueHashes, options, rawAttributes) {
     var query = 'INSERT<%= ignoreDuplicates %> INTO <%= table %> (<%= attributes %>) VALUES <%= tuples %><%= onDuplicateKeyUpdate %>;'
       , tuples = []
       , allAttributes = []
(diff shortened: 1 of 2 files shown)


lib/dialects/mysql/query-generator.js
```diff
@@ -189,7 +189,8 @@ var QueryGenerator = {
 
     if (options && options.updateOnDuplicate) {
       onDuplicateKeyUpdate += ' ON DUPLICATE KEY UPDATE ' + options.updateOnDuplicate.map(function(attr) {
-        var key = this.quoteIdentifier(attr);
+        var field = rawAttributes && rawAttributes[attr] && rawAttributes[attr].field || attr;
```
janmeier: `rawAttributes` should always be defined, and field will always be set, so you should be able to shorten this clause

janmeier:
Looks good - could you add a test?

janmeier:
ping @rfink

rfink:
Hey guys, sorry I haven't responded.  I'll take a look tonight.  There was a reason I didn't write a test, I'll figure out why.

rfink:
So now I remember.  The only specific unit test for updateOnDuplicate works without models (it's only on the query generator side) - test/integration/dialects/mysql/query-generator.test.js.  Any recommendations on how I would go about adding one?

janmeier:
@rfink Some sql unit tests in test/unit/sql should be fine

rfink:
Okay, I'll work on something.  Thanks.

rfink:
I've added a bulkCreate test to the insert test/unit/sql/insert test, I had to skip mssql because it breaks on matching up field aliases to their hash key (specifically, when trying to access the autoIncrement property).  It seems to work as expected for MySQL, and the onDuplicateKeyUpdate gets ignored elsewhere (as it should).  Let me know if there's anything else I can do.

rfink:
Eek, small typo in postgres.  Fixing and testing now...

rfink:
The failing test was a timeout on an unrelated test.  Can you re-run?

janmeier:
Looks good @rfink - If you address my previous comment and squash + rebase, this should be good to merge

rfink:
Ah, hadn't saw that.  I will work on getting this done.  Thanks.

rfink:
I admit I'm haven't done a ton of rebasing, but let me know if this is what you are looking for.  Thanks in advance.

janmeier:
@rfink Somehow you managed to get the messages for all four commits into the squashed one :D 

No worries, I rebased and pushed in https://github.com/sequelize/sequelize/commit/da57c5b5878d079724bd81e1a6ab89d23efdc6ff

janmeier:
Thanks! :)

PR #4369, opened 2015-08-21
Fixing isolation level undefined error

khanmdv:
The isolation levels are available in the Sequelize.Transaction.ISOLATION_LEVELS map. Sequelize.Transaction.READ_COMMITTED resolves to undefined. This is w.r.t Sequelize v2.0.5.


mickhansen:
`docs/api` is based on code comments, don't edit the markdown directly.

PR #4375, opened 2015-08-22
include with separate:true fails if id is not part of the attributes list

sorin:
Should sequelize attempt to include the missing id field? Should it raise an appropriate error?

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -610,6 +610,10 @@ validateIncludedElement = function(include, tableNames, options) {
     include.duplicating = false;
   }
 
+  if (include.separate === true && options.attributes && options.attributes.length && !_.includes(options.attributes, association.source.primaryKeyAttribute)) {
+    options.attributes.push(association.source.primaryKeyAttribute);
+  }
+
   // Validate child includes
   if (include.hasOwnProperty('include')) {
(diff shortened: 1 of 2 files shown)


sorin:
@mickhansen thoughts? I can try to fix it if you think we should do something about it

mickhansen:
Hmm yes, we need to make sure that the id and needed foreign keys are added to `attributes`.
I actually thought i had already considered this, maybe i did and just postponed fixing it :)

sorin:
@mickhansen take a look please

mickhansen:
LGTM, do you mind squashing your commits?

sorin:
Done, thanks for reviewing this! BTW, any immediate plans to support separate:true for BTM? I'm tempted to give it a try.

mickhansen:
Not sure, currently a lot of my implementations are driven by the needs of my startup, so it comes and goes how much i get done :)

Hopefully soon.

sorin:
Haha sounds like our story as well, all of these bug fixes / features are coming from stuff we need. We actually have our own implementation outside Sequelize for the separate:true BTM use case, it's probably time to try and contribute it. Created feature request here: https://github.com/sequelize/sequelize/issues/4376

Issue #4191, opened 2015-07-26
findAndCountAll() and count() pollutes options.includes object


atorkhov:
findAndCountAll() and count() pollutes options.includes object that was passed to the call. This is a problem if I want to store and reuse query options sometime later.

mickhansen:
We need `options = Model.$optClone(options)` about here: https://github.com/sequelize/sequelize/blob/master/lib/model.js#L1494

harish2704:
Yes.  I can confirm that at-least Mode.count method pollutes the options .
Because of this bug, A typical pagiante function like this 
```javascript
var store = {};
return model.count(modelOpts)
.bind(store)
  .then(function(count){
    this.count = count;
    var paginationParams = {
      count: count,
      page: paginationOpts.page,
      size: paginationOpts.size,
    };
    var pagination = paginator.paginate( paginationParams );
    this.pagination = pagination;
    modelOpts.offset = pagination.start-1;
    modelOpts.limit = pagination.size;
    return model.findAll( modelOpts );
  });
```
Is not working.

PR #4378, opened 2015-08-24
Fix: Model.count pollutes options.include object

harish2704:
Please the issue https://github.com/sequelize/sequelize/issues/4191


janmeier:
LGTM - can you add a couple of unit tests? They don't need to hit the DB, so yuo can just stub the `.query` function - look at some of the other unit tests for an example

harish2704:
I shall try my best. But I need some time

PR #4379, opened 2015-08-24
Calling set with dot.separated key on a JSON/JSONB attribute will not flag the entire object as changed

pixel2:
I'm using a JSONB attribute but I'm not able to save the data using set() on my instance object. When reading the documentation is says that using a dot.separated key should change the instance and flag the entire object as changed. If I inspect the instance object I can see that the "_changed" property now contains my dot.separated keys not the entire object key as expected.

Looking at the source code for instance.js I found a problem in the way changed is called when using a det.separated key. I have attached a pull request that actually flag the entire object as changed and I also make sure only to flag if a property has changed.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -322,8 +322,11 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
           // If attribute is not in model definition, return
           if (!this._isAttribute(key)) {
             if (key.indexOf('.') > -1 && this.Model._isJsonAttribute(key.split('.')[0])) {
-              Dottie.set(this.dataValues, key, value);
-              this.changed(key, true);
+              var previousDottieValue = Dottie.get(this.dataValues, key);
+              if (!_.isEqual(previousDottieValue, value)) {
+                Dottie.set(this.dataValues, key, value);
+                this.changed(key.split('.')[0], true);
+              }
             }
             return;
           }
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# Next
+- [FIXED] Calling set with dot.separated key on a JSON/JSONB attribute will not flag the entire object as changed [#4379](https://github.com/sequelize/sequelize/pull/4379)
+
 # 3.9.0
 - [ADDED] beforeRestore/afterRestore hooks [#4371](https://github.com/sequelize/sequelize/issues/4371)
 - [ADDED] Map raw fields back to attributes names when using `mapToModel` or `returning` [#3995](https://github.com/sequelize/sequelize/pull/3995)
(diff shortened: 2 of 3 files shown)


lib/instance.js
```diff
@@ -322,8 +322,11 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
           // If attribute is not in model definition, return
           if (!this._isAttribute(key)) {
             if (key.indexOf('.') > -1 && this.Model._isJsonAttribute(key.split('.')[0])) {
-              Dottie.set(this.dataValues, key, value);
-              this.changed(key, true);
+              var previousDottieValue = Dottie.get(this.dataValues, key);
+              if (!_.isEqual(previousDottieValue, value)) {
```
mickhansen: Why `_.isEqual`?
We simply use `!==` for the other logic.

lib/instance.js
```diff
@@ -322,8 +322,11 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
           // If attribute is not in model definition, return
           if (!this._isAttribute(key)) {
             if (key.indexOf('.') > -1 && this.Model._isJsonAttribute(key.split('.')[0])) {
-              Dottie.set(this.dataValues, key, value);
-              this.changed(key, true);
```
mickhansen: Good catch, that was a dumb/un-tested piece of code by me it seems.

lib/instance.js
pixel2: I wanted to make sure to do a deep comparison if you pass a object as value, but maybe that is not a valid use case. I guess you could always use a multiple dot.separated key instead. If you prefer I can update the pull request to a ```!===``` comparison?

mickhansen: Hmm, i'm not sure, the overhead is likely a function call and two type calls (without looking at the lodash code), likely negligible, thoughts @janmeier?

pixel2: Should I update the pull request with `!==` comparison? That way it is following the convention used in the other logic?

pixel2: Any thoughts on this @mickhansen @janmeier ? I'm happy either way, so if you feel more comfortable with `!==` comparison I will change it.

janmeier: i vote for `isEqual`

janmeier: Could you add a test for setting with an object?

pixel2: Unit tests added.

janmeier:
Good work - do you mind squashing to a single, descriptive commit, preferably prefixed with `feat(model.set)` , then I'll press the big green merge button

janmeier:
Oh, and a changelog entry as well would be awesome!

pixel2:
Done, waiting for that big green click :)

mickhansen:
Thank you.

PR #4399, opened 2015-08-27
Added details about "unique" property (closes #4329)

### LINKED ISSUE #4329: Effects of “unique” property in model definition
Hi, I'm using Sequelize v3.5.1 with PostgreSQL v9.4.4 on a NodeJS server project.

In the model definition, it's not entirely clear to me what are the effects of adding the option `unique: true` to a property.

Let's take this code for example:

```js
sequelize.define('User', {
  email: {
    type: Sequelize.STRING,
    unique: true
  },
  password: {
    type: Sequelize.STRING,
    allowNull: false
  }
});
```

Does this mean that PostgreSQL will build a unique index on `email`? So, is it just a shorthand method for this?

```js
sequelize.define('User', {
  email: {
    type: Sequelize.STRING,
  },
  password: {
    type: Sequelize.STRING,
    allowNull: false
  }
}, {
  indexes: [
    {
      unique: true,
      fields: ['email']
    }
  ]
});
```

If so, will such index speed up table queries for email, or just ensure uniqueness?

Thanks!

pensierinmusica:


janmeier:
Thanks :)

Issue #4314, opened 2015-08-13
Datatypes validations are too strict!


janmeier:
Yes, we know - yes we are going to fix it soon :). However, your contributions are also very welcomed

* Number should allow number strings (i.e. `'1'` is a valid integer)
* Similarly, numbers are valid strings https://github.com/sequelize/sequelize/issues/4255
* Booleans should allow 1/0 (possibilty also as strings?
* The datatype should only be validate on insert / update, at least for some types: Examples
  * Like doesn't work on uuids (`'%foo%`) is not a valid uuid https://github.com/sequelize/sequelize/issues/4198
  * Like any doesn't work on postgres (attempts to validate an array as strings) https://github.com/sequelize/sequelize/issues/4313
* (Maybe) A flag that disables the validations completely

I've closed the issues linked here, since they all relate to the same basic problem.

_Important:_ I don't think we should roll back the validation function - it still has its merits. For affected users, either roll back to 3.4.1, or remove the validte function `DataTypes.INTEGER.validate = function () { return true };`

As mentioned, at least some validators should only be run on update / insert. My suggestion would be to add a `context` parameter to the `validate` function, so for example uuids are only validated when inserted / updated. We might also add a context like `nested`, to show the the value comes from a `$like { $any` or similar, and might be an array, object etc.

Gedzis:
For `Similarly, numbers are valid strings` validation function could be updated to
```
STRING.prototype.validate = function(value) {
  var objectType = Object.prototype.toString.call(value);
  if ( objectType !== '[object String]' && objectType !== '[object Number]') {
    if (this.options.binary) {
      if (Buffer.isBuffer(value)) {
        return true;
      }
    }
    throw new sequelizeErrors.ValidationError(util.format('%j is not a valid string', value));
  }

  return true;
};
```

janmeier:
@Gedzis however, we don't want to allow all strings - only strings that are actually numbers :)

defunctzombie:
One way to check that the string is a valid number is to just coerce it to one. IMO this is a better check than more complex ones which become more expensive as well.

```
isNaN(+'123') // false
isNaN(+'foo') // true
isNaN(+'123+e7') // false
```

Note: scientific notation is automatically handled so if the db dialect being used doesn't like that, it needs to handle that separately.

defunctzombie:
> (Maybe) A flag that disables the validations completely

[...]

PR #4402, opened 2015-08-28
Fix empty strings not being validated

fixe:
@janmeier where should I add a test for this? For what I can see we don't have any unit tests for the query-generator component.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -896,7 +896,7 @@ var QueryGenerator = {
     if (value && value._isSequelizeMethod) {
       return this.handleSequelizeMethod(value);
     } else {
-      if (field && field.type && value) {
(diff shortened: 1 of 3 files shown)


lib/dialects/abstract/query-generator.js
```diff
@@ -896,7 +896,7 @@ var QueryGenerator = {
     if (value && value._isSequelizeMethod) {
       return this.handleSequelizeMethod(value);
     } else {
-      if (field && field.type && value) {
+      if (field && field.type && !_.isUndefined(value) && !_.isNull(value)) {
```
janmeier: Isn't this already covered by `&& value`, since `!!null = false`?

fixe: This fixes the case when value is equal to `""`.

janmeier:
@fixe I'm gonna put this  on hold until https://github.com/sequelize/sequelize/issues/4314 is resolved

PR #4412, opened 2015-09-01
Implement SMALLINT and TINYINT datatypes.

dantman:
I'd like to see support for SMALLINT and TINYINT.

I 'wrote' the code for them. However I haven't tested them myself since my machine doesn't have the requirements to `npm install` Sequelize's `devDependencies`.

### DIFF
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -273,6 +273,56 @@ INTEGER.prototype.validate = function(value) {
 };
 
 /**
+ * A 16 bit integer.
+ *
+ * Available properties: `UNSIGNED`, `ZEROFILL`
+ *
+ * @property SMALLINT
+ */
+var SMALLINT = function(length) {
+  var options = typeof length === 'object' && length || {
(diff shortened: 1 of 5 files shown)


OscarGodson:
Yes, I'd really love to see TINYINT. I have a permissions level that will never be two digits, but it'll be between 0-3. I can't use BOOLEAN which means I have to fall back on INTEGER

dantman:
I've rebased and refactored the code.

However I can't test it and there is a new quirk in Sequelize.

PostgreSQL's driver now knows about the oids for the datatype. However I presume both SMALLINT and TINYINT are internally int2 since pg doesn't have a smaller int type. I don't know how this affects Sequelize.

mickhansen:
Build is completely broken, issue with syntax.

quiddle:
I'd love to see this as well!

zaggino:
Hi, still no way to use `SMALLINT` in Sequelize?

PR #4420, opened 2015-09-02
feat(model/attributes): return type and dependency fields for VIRTUAL

mickhansen:
/cc @janmeier


lib/model.js
```diff
@@ -1326,7 +1326,16 @@ Model.prototype.findAll = function(options) {
...
       options.attributes = Object.keys(this.tableAttributes);
+    } else if (this._hasVirtualAttributes) {
+      options.attributes.forEach(function (attribute) {
+        if (this._isVirtualAttribute(attribute) && this.rawAttributes[attribute].type.fields) {
+          options.attributes = options.attributes.concat(this.rawAttributes[attribute].type.fields);
+        }
+      }.bind(this));
+      options.attributes = _.without.apply(this, [options.attributes].concat(this._virtualAttributes));
+      options.attributes = _.unique(options.attributes);
     }
```
mickhansen: Maybe pull all this into mapOptionFieldNames so it will work out of the box for includes aswell?

Issue #4422, opened 2015-09-02
dropTable not schema-aware?


paulxtiseo:
So, you can `createTable` in a schema, it creates the table in the assigned schema, but seems like we are not able to `dropTable` on the down migration when reversing changes?

When using this migration:

```
return queryInterface.dropTable('ivr', {
      schema: 'raw_filemaker'
    });
```

gives
`DROP TABLE IF EXISTS "ivr"`;

This is in v3.6.0 w/postgresql.

Upon superficial scan, seems like `return Utils._.template(query)(values).trim() + ';';` in createTableQuery properly prepends the schema, whereas `return Utils._.template(query)({...});` at the end of [dropTableQuery](https://github.com/sequelize/sequelize/blob/84c9a68e55d698f7ee999217aadf80604a5ed126/lib/dialects/postgres/query-generator.js#L89) does not. Oddly, they both call `this.quoteTable(tableName)`?

andyATryonsoft:
I need this function to work with schemas as well. I also need the add and drop column calls to support schemas as well.

mickhansen:
@paulxtiseo It might not support schema via options, you can try `{tableName: 'ivr', schema: 'raw_filemaker'}` instead of `'ivr'`

alexmon:
@paulxtiseo 
@mickhansen is right, try : 

return queryInterface.dropTable(
  {
     tableName: 'ivr',
     schema: 'raw_filemaker'    
  }
);

it worked for me

victorschinaider:
I think it will be better if the options of `dropTable` are consistent with `createTable`, so you could define schemas in `dropTable`like in `createTable`. I think it's a security issue.

dkrypt:
@mickhansen is there a reason this why this is not documented anywhere on such a nice [website](https://sequelize.org/v6/index.html).

PR #4424, opened 2015-09-02
Update query-interface.js to handle schemas in postgresql

paulxtiseo:
Possible fix for https://github.com/sequelize/sequelize/issues/4422

Please note there are likely many more functions that need to be made schema-freindly, for example `QueryInterface.prototype.renameTable`, the `QueryInterface.prototype.xxxColumn` functions and the `QueryInterface.prototype.xxxIndex` ones.

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -193,8 +193,16 @@ QueryInterface.prototype.dropTable = function(tableName, options) {
   options = options || {};
   options.cascade = options.cascade || options.force || false;
 
-  var sql = this.QueryGenerator.dropTableQuery(tableName, options)
-    , self = this;
+  var self = this;
+
+  if (!tableName.schema && options.schema) {
+    tableName = self.QueryGenerator.addSchema({
+      tableName: tableName,
+      schema: options.schema
+    });
+  }
+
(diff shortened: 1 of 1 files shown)


mickhansen:
Hmm this looks slightly convuluted, but might be necessary.
Can you add a sql unit test aswell?

legomind:
Looking forward to being able to use this.

janmeier:
Ping @paulxtiseo

lib/query-interface.js
```diff
@@ -193,8 +193,16 @@ QueryInterface.prototype.dropTable = function(tableName, options) {
   options = options || {};
   options.cascade = options.cascade || options.force || false;
 
-  var sql = this.QueryGenerator.dropTableQuery(tableName, options)
-    , self = this;
+  var self = this;
+
+  if (!tableName.schema && options.schema) {
```
jonnolen: This code is duplicated all over this file. (it's in createTable twice....) probably a candidate for extraction into a method.

jonnolen:
:+1:  I just ran into this bug as well,  name of PR is misleading, this is a dialect agnostic fix.

janmeier:

PR #4428, opened 2015-09-03
Create a query option for setting schema search_path in Postgres

mliszewski:
When you have a multi-tenant application that has the same database models across multiple schemas the current sequelize model paradigm forces you to create a separate model for each schema that is identical except for the schema property of the tableName object. This change allows you to pass a searchPath option to any query that will be prepend the query with "SET search_path to <searchPath>". If not option is provided it will use DEFAULT which should be the same behavior as before. By supporting a searchPath option I can now define a common set of models and use them across tenant schemas by passing the correct searchPath.


mliszewski:
Is there a way to rerun the tests? All test pass on my fork (https://travis-ci.org/mliszewski/sequelize) and they should be the same. Plus its odd that the test failed for only one of the 4 platforms so I don't think it's a code issue.

mickhansen:
Will this add `SET search_path to X;` in front of EVERY SINGLE query run on postgres? 

Try not to commit `docs/api` next time, makes the diff very large :)

mliszewski:
It will add it in front of any query type that does not set options.supportsSearchPath = false in the query-interface for that type. Certain queries (ROLLBACK,COMMIT,CREATE INDEX, etc.). My initial attempt at this change only prepended it if it was supplied in the query options but this led to unexpected results if you didn't pass it and it has already been set on that connection so I decided to prepend "SET search_path to DEFAULT" if no value was passed. Perhaps one nice addition would be to add a dialect option "alwaysSetSearchPath" that is checked and only set a default if that is present.

Sorry about the docs.

mliszewski:
@mickhansen Can you provide some thoughts on my comments above when you get a chance? Thanks.

mliszewski:
@mickhansen can you provide some feedback on whether this looks good or what changes you would like to see made

mickhansen:
@mliszewski it will only add the statement if the user has enabled a search path value correct?

mickhansen:
Or will it do `SET search_path to DEFAULT` for all sequelize postgres users installing the new version regardless of configuring anything?

mickhansen:
>  forces you to create a separate model for each schema that is identical except for the schema property of the tableName object

should be possible to just have a base model that you call `.schema('schema')` on actually.

mliszewski:
Let me look at the .schema functionality and see if it covers what we need. I had missed this option in the code/docs. If it does not cover our use case I will elaborate and answer your questions above. Thanks.

mliszewski:
@mickhansen I have cleaned up the code and added an explanation as to why .schema does not work for our use case in a new PR https://github.com/sequelize/sequelize/pull/4534

PR #4429, opened 2015-09-03
Rename Instance private options and __options

### LINKED ISSUE #4423: Query with include returns an array with one object containing all null values (because there is a column named options in the model)
We are using sequelize 3.2.0

We have a parent table (e.g. `product`) that has associated tags (e.g. `tag` table) though a many-to-many `productTag` table).  If we query for `product` and include `tag`, but there are no tags, we get a `product` object with a `tags` array containing one `tag`-like object with all null values.  I think this is happening because the query has a left outer join on `productTag` (with inner join on `tag`), but the `tag` columns are all null, which get serialized into an associated `tag` object in the array.

Sorry if this issue already exists or is already fixed - I couldn't find anything related in my searches.

andywhite37:
- Possibly fix #4423
- Rename Instance `options` to `$options` to prevent value
  overwriting with models that have a column named `options`
- Also rename Instance `__options` to `$modelOptions` to avoid confusion
  between $options and __options.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -86,8 +86,8 @@ var Instance = function(values, options) {
   this.dataValues = {};
   this._previousDataValues = {};
   this._changed = {};
-  this.__options = this.Model.options;
-  this.options = options || {};
+  this.$modelOptions = this.Model.options;
+  this.$options = options || {};
   this.hasPrimaryKeys = this.Model.options.hasPrimaryKeys;
   this.__eagerlyLoadedAssociations = [];
   /**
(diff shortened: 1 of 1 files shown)


janmeier:
Looks good to me - good catch on the `__options` - model options is definitely a better name

This shouldn't require a major version right @mickhansen ? Do you agree that if people depend on `.options` its their own fault ;) ?

mickhansen:
It's not documented anywhere, so can hardly be considered public API. I personally probably have some code dependant on that somewhere, but yeah i'd agree that if people are using non-public stuff they can't expect SEMVER to apply there.

janmeier:
Cool, I'll merge and add it to the changelog then

Issue #4432, opened 2015-09-04
[Question] Multiple Sequelize instances


czardoz:
I'm building a database migration tool, so I'd need to connect to two different databases. The [getting started docs](http://docs.sequelizejs.com/en/latest/docs/getting-started/) mention =>

```
Sequelize will setup a connection pool on initialization so you should ideally only ever create one instance per application
```

Does this apply to me? Or does this statement mean that there should be one instance of Sequelize *per database*?

mickhansen:
One instance per database would apply in your case.

janmeier:
One per db - multiple sequelize instances should be able to coexist, we just advice people to only use one per db. 

There might be problems if migrating between two different dialects though, eg mysql to postgres etc. At least I remember there was once, but it might be fixed now :-)

czardoz:
Great thanks! I'll send a PR to update the docs to make this clearer.

PR #4433, opened 2015-09-04
make it clear that there should be one sequelize instance per database

czardoz:
So that no one asks #4432 again :)


mickhansen:
my favorite kind of issue-opener!

PR #4437, opened 2015-09-04
Update connection-manager.js to support sqlcipher

oshoemaker:
Add support for sqlcipher for use with sqlite databases

What you are doing: Adding the necessary PRAGMA commands to support sqlcipher (key and cipher)

What do you expect to happen: When the connection is started it will issue the PRAGMA commands which will allow for reading/writing to the encrypted database.

What is actually happening: What is stated above

Which dialect you are using: sqlite

Which sequelize version you are using: 3.X


mickhansen:
Looks great! If you could just squash your commits to something descriptive of what you've done, something like: "feat(sql): support sqlsipher via key and cipher options to new Sequelize"

oshoemaker:
Mick, I squashed the commits and then the CI tests failed unrelated to my changes.  Is there anything I can do to easily get the tests to run?  It was a mysql connection timeout with travis CI.

PR #4440, opened 2015-09-06
always need to join data when hasJoin is true not matter what options…

pentium100:
In handleSelectQuery , we always need to join data when hasJoin is true.
not matter the options.raw setting.


mickhansen:
Perhaps we should change the option to something more descriptive.

Issue #3752, opened 2015-05-19
Need to add prefix 'N' before Unicode string when work with MSSQL


colinyuran:
I am new to Sequelize and NodeJS, need to use it work with SQL Server 2014, and I also need to deal with Unicode string in my table columns.

In my model definition, I define a Name field with type DataTypes.STRING(255), the final real database column type is NVARCHAR(255). I can insert data with Unicode string as below sql statement:

    insert into [Test] values (N'我們')

Database can store the Chinese string correctly.

But when I use sequelize API, such as Model.create method, I will get question mark ( ???? ) in name column. After I check the NodeJS log, I find out the SQL statement generated by sequelize is as below:

    INSERT INTO [Test] ([Name]) OUTPUT INSERTED.* VALUES ('我們');

There is no prefix 'N' before the Chinese string, this is why it can't store correctly in Database.

I debug by myself but don't find a graceful way to change this behavior, is there any configuration not mentioned in the document?

I also submit a question in stackoverflow with more details (include my codes and my debug findings, also with an ugly temporary fix). Here is the link:
http://stackoverflow.com/questions/30303841/how-to-make-sequelize-add-n-before-unicode-string-in-its-sql-statement

janmeier:
Are there any reverse implications to adding the `N` - I mean, will there ever be cases where you don't want to have the N prefixed?

mbroadst:
@janmeier iirc no. I believe that putting an N before a string literal simply converts it to an NVARCHAR unicode representation. The tradeoff here is that you might not want the extra space in the on-disk representation, but we're already assuming that STRING == NVARCHAR, so if the tests pass with everything being switched to prefix with N then I'd say you're probably good to go. Maybe someone knows more than I do though :smile:

janmeier:
Yep, I'd say that's the way to go then. PR welcome @colinyuran  :=

colinyuran:
In my case, we need to store strings from many different languages in Database, so we prefer to define all string like types (char, text and so on) as unicode, so I am really glad the sequelize already translate all string type to NVARCHAR. But Microsoft SQL Server is very unique, even you define a table column as NVARCHAR which means the column is unicode, you still need to add the prefix N before the string literal value when you try to insert or update that column, this behavior is caused by some historical reason and Microsoft SQL Server need to keep compatible with old versions. So even I use Microsoft SQL Server management GUI to insert a column defined as NVARCHAR without the prefix 'N' , I still get ??? in the record.

[...]

PR #4444, opened 2015-09-07
Added escape function override to allow unicode strings

ThomasHambach:
https://github.com/sequelize/sequelize/issues/3752


sorn6541:
I think `dataTypes.TEXT` need N prefix too.

janmeier:
This seems like an awful lot of duplication just to add `'N'`. I think this will be better solved by https://github.com/sequelize/sequelize/pull/4186

janmeier:
Fixed in https://github.com/sequelize/sequelize/pull/5039

PR #4459, opened 2015-09-09
Sequelize version number as property of sequelize

overlookmotel:
This PR adds `version` property to `Sequelize` - the version number of Sequelize in use.

This is useful e.g. for plugins where API has changed between different versions of Sequelize, and the plugin wishes to support multiple versions. See https://github.com/overlookmotel/sequelize-definer/blob/master/lib/patches.js for an example.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -207,6 +207,12 @@ var Sequelize = function(database, username, password, options) {
   Sequelize.runHooks('afterInit', this);
 };
 
+/**
+ * Sequelize version number.
+ * @property version
+ */
+Sequelize.version = require('../package.json').version;
+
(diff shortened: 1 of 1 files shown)


janmeier:
Good idea :)

overlookmotel:
Thanks @janmeier!

Any chance you could push a new version to npm? I'm overhauling my plugins at the moment and would be great to be able to use `Sequelize.version`

janmeier:
Never tried that before, you're gonna have to bug @mickhansen about that

mickhansen:
`v3.8.0` released.

overlookmotel:
Awesome! Thanks @mickhansen

Issue #3951, opened 2015-06-17
ES6 in Sequelize codebase


overlookmotel:
@mickhansen recently raised the possibility of using [babel](https://babeljs.io/) to allow Sequelize to be written using ES6 features while retaining compatibility with versions of node earlier than v0.12.

@janmeier and @mbroadst were in favour, @BridgeAR has reservations.

The original discussion is here: https://github.com/sequelize/sequelize/pull/3929/files#r32521460

I thought it worthwhile to give this its own issue for discussion as it's a big topic.

overlookmotel:
My opinion is:

I'm writing node apps in ES6. Using generators, `co` and `yield` makes it *sooooo* much simpler to write async code. No more endless `.then()` promise chains...

But I'm running these apps in node v0.12 with `--harmony` so the ES6 features are native.

I'd be broadly in favour of transitioning Sequelize to ES6, because the benefits in terms of writing shorter more comprehensible code are so large.

However, what are the implications for debugging? If the source code goes through a babel compilation, does that make it hard to locate the source of a bug, if a thrown error references a line number which doesn't correspond to the right line in the source code?

overlookmotel:
I'd suggest if you wanted to go this way, to restrict ES6 features used to those available natively in node v0.12 with `--harmony`, so it can run without compiling on node v0.12. Babel compilation could be switched on only for older node versions.

mickhansen:
@overlookmotel The beautiful dynamic object keys and structural deassignments aren't in harmony though :/

mbroadst:
IMHO I think the suggestion to just use harmony is sort of dead in the water since you'd basically be telling users that sequelize no longer works with 0.10.x and if they want to use it with 0.12.x they have to run with ```--harmony```: no thanks.

I think babel is your only choice if you want to go down this path, it provides a nice platform for which to support all the things you need to support here while still looking forward.

overlookmotel:
@mbroadst My suggestion was that Sequelize would work natively on 0.12.x with `--harmony`, but that when run on 0.10.x it would be babelized. So it'd run on both platforms, but with no need for babelization if you're on node 0.12.x.

But in any case, I didn't realize that as @mickhansen said some of the best ES6 features aren't available in 0.12.x harmony. So my whole idea is dead in the water for that reason!

[...]

PR #4468, opened 2015-09-11
Support Node v4.0.0

mickhansen:


mickhansen:
Doesn't look like pg-native is updated for 4.0 yet, needs a NaN update.

mickhansen:
https://github.com/brianc/node-pg-native/issues/27

janmeier:
Hijacking this issue to extend our talk on IRC about what versions to support:

I believe we should remove 0.10 and iojs v1 and 2 and use v3 instead - Using the argument that people on io are quicker to upgrade. Of course a lot of production apps are probably still on 0.10, but i'd really love to speed up the build matrix

mickhansen:
I'd argue we could cut everything but Node v4.0 for a Sequelize 4.0 release, if we slate it for something like Q1 2016

janmeier:
I'll have to give it a bit more though, but my initial thoughts are definitely positive! :)

cc @sdepold @overlookmotel @mbroadst @BridgeAR @durango @thanpolas @cusspvz and https://github.com/sequelize/sequelize/issues/3951 (and also everyone else who wants to voice their opinon, I'm just mentioning the top comitters ;) )

One thing we might have to consider is suport for 3.0 - generally the previous version is more or less dead after a major bump - should that still be the case when official support for everything but 4.0 is removed?

mbroadst:
I say you go with node4 and node0.12, cut everything else out. Generally _for_ the adoption of node4 either way

janmeier:
Bonus question - 0.12 without or without [`--harmony`](http://stackoverflow.com/questions/28388885/ecmascript-6-features-available-in-node-js-0-12)

BridgeAR:
I'd keep 0.10 support. There was a discussion about this in iojs/Node and if I recall correct they decided to support it for quite some time further. There are probably still quite a few people using 0.10.

I would remove iojs 1 and 3 support though. The people who used iojs 1.0 are early adapters and have likely moved on to newer versions. iojs 2 did not really break a lot and most people won't have any issue using it. But iojs 3 broke a lot and nodejs 4 did aswell. If someone is relying on a native module that did not yet update NaN, they have no chance but to keep iojs 2. And Node 4 followed iojs 3 after a very short period of time and most people have likely skipped iojs 3 or moved on to Node 4.

So I'd keep:
0.10, 0.12, iojs 2 and Node 4

And I wouldn't use the harmony flag, as especially those who keep using the older versions are likely not using the harmony flags.

sdepold:
> 0.10, 0.12, iojs 2 and Node 4

I pretty much have the same opinion.

mickhansen:
Only removing iojs 1 & 3 doesn't do a whole lot for build speed or developer productivity, which would be the motivation for removing builds :)

sdepold:
> Only removing iojs 1 & 3 doesn't do a whole lot for build speed or developer productivity, which would be the motivation for removing builds :)

True. But I'm not sure if we should remove those builds just for the sake of speed :-| I wonder if there is a possibility to get a dedicated test environment on travis

sdepold:
Also build speed is still pretty good compared to other projects :D https://travis-ci.org/jruby/jruby/builds/80385715

mickhansen:
Speed and ES6 features is the major reason to remove builds, slowness of builds kills developer productivity.

johanneswuerbach:
I think there is not really a point of keeping 0.12, iojs and node 4. Shouldn't all breaking changes between 0.12 and iojs now also be included in 4?

BridgeAR:
@johanneswuerbach this might likely be the case.

overlookmotel:
Personally I vote for maintaining support for node v0.12 at least. I don't plan on moving up to node v4 until most modules that currently have problems with NaN are upgraded.

Agree with @mickhansen that the long build times are a pain. They've slowed me down considerably when putting together PRs. But actually, that's just as much about how long it takes to run all the tests for even a single db flavour on my own laptop. Slooooow.

@sdepold "I wonder if there is a possibility to get a dedicated test environment on travis" - how do you mean?

sdepold:
> @sdepold "I wonder if there is a possibility to get a dedicated test environment on travis" - how do you mean?

I asked the travis peeps if they could give us some dedicated build slots but unless someone wants to spend 500$ a month, there is no other option right now.

mickhansen:
If a move were to happen, it would not be for atleast 6 months when Node v4.0.0 is properly caught up and we've fixed as many bugs as we can on Sequelize v3.x.x.

I'd just love to move Sequelize forward aswell with ES6 classes and other features.

Converting more and more integration tests to unit tests helps the build speed quite a bit.

overlookmotel:
This is going a bit off-topic now, but on the subject of speeding up builds...

The tests spend most of their time awaiting responses from the database - it's not that the node process is running at maximum capacity. I wondered if there's any framework for running multiple tests concurrently?

In a sense, the way tests run at present is very "un-node". i.e. each test running one after another, rather than concurrent. What I'm thinking is a test runner equivalent of (gulp)[https://www.npmjs.com/package/gulp] as opposed to (grunt)[https://www.npmjs.com/package/grunt]

This would require having multiple test schemas ("schemas" in the MySQL sense of the word) and running each test file on it's own instance of sequelize bound to a separate database schema. Then multiple test files could be executed simultaneously, without their operations clashing with each other.

For example:
- Create 10 test schemas
- Test framework runs 10 test files concurrently, each connecting to a separate schema
- Test framework then un-parallelizes and interleaves the output from the tests to present it all in the same kind of serial list format as you get at present

In theory, assuming that the database itself doesn't become a bottleneck, this could increase build speed dramatically. Where the node process itself is doing a lot of work, it could also take advantage of multiple processor cores.

It seems there have been a few approaches made at this:
https://github.com/danielstjules/mocha.parallel
https://www.npmjs.com/package/grunt-mocha-parallel

Does anyone have experience with this? Or know of a complete solution which is ready to go?

horiuchi:
What about this issue?

mickhansen:
@horiuchi works fine on 4.0 (i use it), we just don't test against it yet cause there was issues with pg-native

horiuchi:
@mickhansen Thank you. I try to work sequelize on 4.2 :smile:

overlookmotel:
Hi @mickhansen. Has there been any progress with the `pg-native` issues? If Sequelize now works on Node 4.0, it'd be great to be testing against that.

mickhansen:
@overlookmotel not sure, afair it was basically resolved but was a NPM issue

overlookmotel:
OK cool. Thanks for the info.

PR #4478, opened 2015-09-12
Model.beforeSync, Model.afterSync, sequelize.beforeBulkSync, sequelize.afterBulkSync

alekbarszczewski:
* before/afterSync hooks on Model executed before/after Model.sync(...)
* before/afterBulkSync hooks on sequelize (instance) executed before/after sequelize.sync(...)
* useful for plugins (for example to create a trigger after model table has beem created etc)
* TODO: add new tests

```js
var Sequelize = require('sequelize')

var sequelize = new Sequelize('postgres://localhost/sequelize_test', {
  logging: false
})

// executed "before" `sequelize.sync(...)`
sequelize.addHook('beforeBulkSync', function (options) {
  // this = sequelize instance
  console.log('beforeBulkSync')
})

// executed "after" `sequelize.sync(...)`
sequelize.addHook('afterBulkSync', function (options) {
  // this = sequelize instance
  console.log('afterBulkSync')
})

var Model = sequelize.define('Model', {}, {
  hooks: {
    // executed "before" `Model.sync(...)`
    beforeSync: function (options) {
      // this = Model
      console.log('beforeSync')
    },
    // executed "after" `Model.sync(...)`
    afterSync: function (options) {
      // this = Model
      console.log('afterSync')
    }
  }
})

sequelize.sync()
.then(function(){
  console.log('done')
})
```

output

```
beforeBulkSync
beforeSync
afterSync
afterBulkSync
done

```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -870,6 +870,13 @@ Model.prototype.refreshAttributes = function() {
       delete definition.index;
     }
   });
+  // Create a map of field to attribute names
+  this.fieldAttributeMap = Utils._.reduce(this.fieldRawAttributesMap, function(map, value, key) {
+    if (key !== value.fieldName) {
+      map[key] = value.fieldName;
+    }
+    return map;
+  }, {});
 
   this.uniqueKeys = this.options.uniqueKeys;
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -2552,7 +2552,7 @@ Model.prototype.$getDefaultTimestamp = function(attr) {
 // Inject current scope into options. Includes should have been conformed (conformOptions) before calling this
 Model.$injectScope = function (scope, options) {
   scope = optClone(scope);
-  
+
   var filteredScope = _.omit(scope, 'include'); // Includes need special treatment
 
   _.defaults(options, filteredScope);
@@ -2572,9 +2572,6 @@ Model.$injectScope = function (scope, options) {
         }
         return sameModel;
       })) {
-        // Do not inject Model.$scope.include directly,
-        // make a shallow copy (#4470).
-        scopeInclude = _.clone(scopeInclude);
         options.include.push(scopeInclude);
       }
     });
--- a/lib/model.js
+++ b/lib/model.js
@@ -955,14 +955,15 @@ Model.prototype.removeAttribute = function(attribute) {
  * @return {Promise<this>}
  */
 Model.prototype.sync = function(options) {
-  options = Utils._.extend({}, this.options, options || {});
-  var runHooks = options ? (typeof options.hooks === 'undefined' ? true : !!options.hooks) : true;
+  options = options || {};
+  options.hooks = options.hooks === undefined ? true : options.hooks;
+  options = _.defaults(options, this.options);
 
   var self = this
     , attributes = this.tableAttributes;
 
   return Promise.try(function () {
-    if (runHooks) {
+    if (options.hooks) {
       return self.runHooks('beforeSync', options);
     }
   }).then(function () {
@@ -987,7 +988,7 @@ Model.prototype.sync = function(options) {
       return self.QueryInterface.addIndex(self.getTableName(options), _.assign({logging: options.logging}, index), self.tableName);
     });
   }).then(function () {
-    if (runHooks) {
+    if (options.hooks) {
       return self.runHooks('afterSync', options);
     }
   }).return(this);
--- a/lib/model.js
+++ b/lib/model.js
@@ -959,6 +959,10 @@ Model.prototype.sync = function(options) {
     , attributes = this.tableAttributes;
 
   return Promise.try(function () {
+    if (options.hooks) {
+      return self.runHooks('beforeSync', options);
+    }
+  }).then(function () {
     if (options.force) {
       return self.drop(options);
     }
@@ -979,6 +983,10 @@ Model.prototype.sync = function(options) {
     return Promise.map(indexes, function (index) {
       return self.QueryInterface.addIndex(self.getTableName(options), _.assign({logging: options.logging}, index), self.tableName);
     });
+  }).then(function () {
+    if (options.hooks) {
+      return self.runHooks('afterSync', options);
+    }
   }).return(this);
 };
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -954,12 +954,13 @@ Model.prototype.removeAttribute = function(attribute) {
  */
 Model.prototype.sync = function(options) {
   options = Utils._.extend({}, this.options, options || {});
+  var runHooks = options ? (typeof options.hooks === 'undefined' ? true : !!options.hooks) : true;
 
   var self = this
     , attributes = this.tableAttributes;
 
   return Promise.try(function () {
-    if (options.hooks) {
+    if (runHooks) {
       return self.runHooks('beforeSync', options);
     }
   }).then(function () {
@@ -984,7 +985,7 @@ Model.prototype.sync = function(options) {
       return self.QueryInterface.addIndex(self.getTableName(options), _.assign({logging: options.logging}, index), self.tableName);
     });
   }).then(function () {
-    if (options.hooks) {
+    if (runHooks) {
       return self.runHooks('afterSync', options);
     }
   }).return(this);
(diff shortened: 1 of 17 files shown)


overlookmotel:
Looks good to me! Of course, as you say, it needs tests.

By the way, I'm not one of the maintainers of Sequelize. I just have an interest in plugins and hooks as I've published a few plugins and did a part of implementing some of the hooks a while back.

mickhansen:
Looks good, a few unit tests (don't need integration tests) and LGTM.

lib/model.js
```diff
@@ -959,6 +959,10 @@ Model.prototype.sync = function(options) {
     , attributes = this.tableAttributes;
 
   return Promise.try(function () {
+    if (options.hooks) {
```
janmeier: Remeber to add to API docs :)

janmeier:
+1 on LG

lib/model.js
mickhansen: What's the default value of `options.hooks`? `undefined`? should probably be `true`

lib/sequelize.js
```diff
@@ -854,14 +854,16 @@ Sequelize.prototype.sync = function(options) {
...
-    when = this.drop(options);
-  } else {
-    when = Promise.resolve();
-  }
 
-  return when.then(function() {
+  return Promise.try(function () {
+    if (options.hooks) {
+      return self.runHooks('beforeBulkSync', options);
+    }
```
mickhansen: What's the default value of `options.hooks`? `undefined`? should probably be `true`

lib/model.js
alekbarszczewski: @mickhansen / @janmeier Hmm, I am not sure what do you mean with "options.hooks". I followed pattern in other hooks like here https://github.com/sequelize/sequelize/blob/master/lib/model.js#L2053 . It seems that when `sync` method is called (either on model or sequelize instance) then by by default `hooks` field is set on options object and it looks like this:

```js
hooks: { beforeSync: [ [Function] ], afterSync: [ [Function] ] },
```

alekbarszczewski:
@mickhansen / @janmeier / @overlookmotel I've added few tests in `test/integration/hooks.test.js` (it seemed to be a good place as other hook tests are there). Can you look on them? Is it enough?

Also it works like this:

```js
Model.sync() // will run beforeSync/afterSync hooks on Model
Model.sync({ hooks: false }) // will NOT run beforeSync/afterSync hooks on Model
sequelize.sync() // will run beforeBulkSync/afterBulkSync on sequelize AND beforeSync/afterSync hooks on Model
sequelize.sync({ hooks: false }) // will NOT run beforeBulkSync/afterBulkSync on sequelize AND ALSO WILL NOT run beforeSync/afterSync hooks on Model
```

I am not sure if this is what we want, what do you think?

mickhansen:
@alekbarszczewski `hooks: true` is the default for all other methods, and what plugins need, so i'd say that's the way to go.

lib/model.js
mickhansen: This is the `model.sync` call right?
You do a condition based on `options.hooks` but you have not set a default value for it as far as i can tell so it will effectively be false where most other calls have `true` as a default.

mickhansen: In the top of `sync` options is copied from the model options, you actually have to make sure that hooks is then removed and it uses the boolean value of the directly passed options instead.

alekbarszczewski:
@mickhansen At the very beginning of `Model.prototype.sync` there already was line
```
options = Utils._.extend({}, this.options, options || {});
```
and at the beginning of `Sequelize.prototype.sync` there was 

```
options = Utils._.defaults(options || {}, this.options.sync, this.options);
```

So I think it will be ok to just leave it as is, what do you think? `hooks` option will be always Model/sequelize hooks object, and if user will explicitly pass `{ hooks: whatEverThatEvaluatesToFalse }` then hooks won't be executed.

mickhansen:
@alekbarszczewski There's a big difference between the `hooks` option on Model/Sequelize and the `hooks` option passed to calls that run hooks (one is an object other is a boolean).

I think it's best to make sure it's a boolean, and if you could provide API docs for the two methods that would be great aswell :)

alekbarszczewski:
@mickhansen Do you mean something like this?:

```js
Model.prototype.sync = function(options) {
  options = Utils._.extend({}, this.options, options || {});
  // runHooks will be true if "hooks" option is not passed or it evaluates to `true`
  var runHooks = options ? (typeof options.hooks === 'undefined') ? true : !!options.hooks : true;

  var self = this
    , attributes = this.tableAttributes;

  return Promise.try(function () {
    if (runHooks) {
      return self.runHooks('beforeSync', options);
    }
  }).
  // ...
```

mickhansen:
I was thinking more like:

```js
Model.prototype.sync = function(options) {
  options = options || {};
  options.hooks = options.hooks !== undefined ? options.hooks : true;
  options = _.defaults(options, this.options);

  .....
```

alekbarszczewski:
@mickhansen Ok, I've updated code added API docs for `Sequelize.prototype.sync` method. Also updated changelog. Anything more I should do?

alekbarszczewski:
@mickhansen @janmeier Strange, CI builds fail but I think those errors are not caused by me, for example:

```
1) [POSTGRES] Sequelize Errors Constraint error Can be intercepted as UniqueConstraintError using .catch:
     AssertionError: expected spy to have been called exactly once, but it was called 0 times
      at Context.<anonymous> (test/integration/error.test.js:199:35)
  From previous event:
      at Promise.then (lib/promise.js:21:17)
      at Context.<anonymous> (test/integration/error.test.js:198:12)
```

```
5) [POSTGRES] Model upsert works with upsert on a composite key:
     AssertionError: expected true to be falsy
      at Context.<anonymous> (test/integration/model/upsert.test.js:80:38)
  From previous event:
      at Promise.then (lib/promise.js:21:17)
      at Context.<anonymous> (test/integration/model/upsert.test.js:76:12)
```

Maybe I broke something because I've recently updated my fork to have fresh changelog.md...

alekbarszczewski:
@mickhansen I've fixed failing builds - it was caused by modifying "hooks" option passed to sequelize/Model.sync(options) method. It seems that I can't actually set it to `boolean`, because other methods are expecting `hooks` option to be sequelize/Model hooks object. So now I just check "hooks" option passed by user, save it in "runHooks" variable and ignoring this option for further calls that also take "options" object as arg. Not very pretty but I couldn't figure out a better way.

mickhansen:
@alekbarszczewski Something is getting assigned incorrectly if the `hooks` options on `sync` is leaking into other stuff. It is definitely possible to have it be a boolean.

alekbarszczewski:
@mickhansen You were right :) I still don't know what was wrong, but now it works fine. So it seems to be ready. Should I update my fork to resolve conflicts on my own? Thanks.

mickhansen:
@alekbarszczewski Yes please rebase on top of master :)

alekbarszczewski:
hurray!

mickhansen:
Great @alekbarszczewski. Now we just need a squash to get a more descriptive list of (or single) commits.

alekbarszczewski:
@mickhansen Sorry, I am not so familiar with contributing to public repos - what do you mean by squash? Do I have to do something more?

mickhansen:
@alekbarszczewski You can use an interactive rebase to merge commits and update their commit messages for a nicer piece of git history, you can check out http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

alekbarszczewski:
@mickhansen I've tried to do this, but for some reason when I am trying to squash commits I get more and more conflicts... Maybe this is because I synced my branch with master few times? Now theoretically I've squashed few commits by resolving conflicts but I can't push branch to my fork because it says that it is non-fast forward. Is it normal? Sorry, I am not familiar with git "internals"...

janmeier:
When the rebase is done you need to force push. This is because you are rewriting the commit history, so you have to convince git that you're not breaking anything

alekbarszczewski:
Hmm, I've forced push, but now it says that there are conflicts to merge to sequelize/master... Why is that?

mickhansen:
Doesn't look like you've squashed either. The PR is really screwed now :/ It is generally a bad idea to merge master into a PR since it becomes quite unmanageable (better to rebase to stay up to date).

alekbarszczewski:
Damn, I will prepare another PR then :)

mickhansen:
Thank you for taking the effort / doing the work, greatly appreciated.

alekbarszczewski:
@mickhansen no problem, thanks for Sequelize guys ;). I've created another PR #4544 with single commit not I hope I wont have to rebase anything :)

janmeier:
Closing this one, since the other PR has been merged

PR #4480, opened 2015-09-12
Update spelling in line 4 of sequelize/docs/api/datatypes.md

jingpei:
Changed "The datatypes are used when definining a new model using" to "The datatypes are used when defining a new model using".

screen shot 2015-09-12 at 3 33 09 pm


janmeier:
Please update the code comments, the files in api are automatically generated. The relevant line should be in datatypes.js

PR #4487, opened 2015-09-13
Update spelling in line 11 of sequelize/lib/data-types.js

jingpei:
Changed "The datatypes are used when definining a new model using" to "The datatypes are used when defining a new model using".

screen shot 2015-09-12 at 3 33 09 pm


janmeier:
Thanks :)

Issue #4491, opened 2015-09-14
commit/rollback multiple times when not using CLS


arcana261:
Hi.

I had an issue as of late using latest stable version of Sequelize on our proprietary app. We don't use CLS (manual transaction life-cycle management on our side) and are on PostgreSQL. I would receive multiple instances of  **SET TRANSACTION ISOLATION LEVEL must be called before any query** errors. This problem was devastating because connections in connection pool were exhausted and as a result all REST APIs to our server were timed-out. I narrowed the problem down to executing **rollback** on the same transaction multiple times.

So I think that there are two problems with Sequelize. First is that it allows rollback/commit on a transaction multiple times and thus it releases (pools) the same connection multiple times. Second is that if any error should happen while initializing a new transaction (promise chain in method **prepareEnvironment**) a connection is lost and exhausted. I can not rollback such a transaction on my side because transaction context object is not given to me in the first place, if such error is to occur.

 I understood that by doing so, the connection is pooled multiple times and thus can be acquired and used in two different transactions concurrently. I'm fixing this in my code but nevertheless sequelize should not have had pooled the same connection multiple times.

In order to fix the first problem, I suggest adding a guard **if** to lib/transaction.js method **rollback** and **commit** so that any transaction is not rolleback/committed multiple times. 

In order to fix the second problem, I suggest that in method **prepareEnvironment** we should check for any error during transaction startup, if any error should happen we should rollback half-initialized transaction (to fix connection exhaust bug).

I will make a pull request soon.

janmeier:
A pull request sounds good, it's hard to reason about code changes when they're described in plain text ;-)

arcana261:
made a pull request by the same name. I cant run the tests right now but I think they would be ok, Not so many lines are changed and I've done no magick :smile:

PR #4492, opened 2015-09-14
commit/rollback multiple times when not using CLS

arcana261:
Hi.

This is a minor patch fixing issue reported by me [#4491](https://github.com/sequelize/sequelize/issues/4491). I have added a guard check in **commit** and **rollback** in **lib/transaction.js** to disallow multiple commit/rollback on a single transaction. Also to fix connection exhaust bug when an error occurs in method **prepareEnvironment** promise chain I rollback half-initialized transaction before propagating error.

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -127,6 +127,10 @@ Transaction.LOCK = Transaction.prototype.LOCK = {
 Transaction.prototype.commit = function() {
   var self = this;
 
+  if (this.finished) {
+    throw new Error('Transaction cannot be committed because it has been finished with state: ' + self.finished);
+  }
+
   this.$clearCls();
 
   return this
@@ -149,6 +153,10 @@ Transaction.prototype.commit = function() {
  */
 Transaction.prototype.rollback = function() {
   var self = this;
+
+  if (this.finished) {
+    throw new Error('Transaction cannot be rolled back because it has been finished with state: ' + self.finished);
+  }
 
   this.$clearCls();
 
@@ -180,6 +188,10 @@ Transaction.prototype.prepareEnvironment = function() {
     return self.setIsolationLevel();
   }).then(function () {
     return self.setAutocommit();
+  }).catch(function (setupErr) {
+    return self.rollback().finally(function () {
+      throw setupErr;
+    });
   }).tap(function () {
     if (self.sequelize.constructor.cls) {
       self.sequelize.constructor.cls.set('transaction', self);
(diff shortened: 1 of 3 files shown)


arcana261:
I've not run the tests though... I think they would be OK.

lib/transaction.js
```diff
@@ -150,6 +154,10 @@ Transaction.prototype.commit = function() {
 Transaction.prototype.rollback = function() {
   var self = this;
 
+  if (self.finished) {
+    throw new Error('Transaction cannot be committed because it has been finished with state: ' + self.finished);
```
mickhansen: `cannot be rolled back`

mickhansen:
This is super, since we have the same guard here: https://github.com/sequelize/sequelize/blob/master/lib/sequelize.js#L708

Just a few style comments and what-not

lib/transaction.js
```diff
@@ -127,6 +127,10 @@ Transaction.LOCK = Transaction.prototype.LOCK = {
 Transaction.prototype.commit = function() {
   var self = this;
 
+  if (self.finished) {
```
mickhansen: please use `this` both here and in rollback.

lib/transaction.js
```diff
@@ -180,6 +188,12 @@ Transaction.prototype.prepareEnvironment = function() {
     return self.setIsolationLevel();
   }).then(function () {
     return self.setAutocommit();
+  }).catch(function (err) {
+    try {
+      self.rollback();
+    } catch (e) {}
+
+    throw err;
```
mickhansen: Please refactor `self.rollback()` to be called as a promise, something like:

```js
return self.rollback().catch(function () {
  throw setupErr; // rename parent catch arg for clarity.
});
```

mickhansen:
Could you please add an integration test a long the lines of:

```js
return expect(sequelize.transaction().then(function (t) {
  return t.commit().then(function () {
    return t.commit();
  });
}).to.be.rejected;
```

lib/transaction.js
arcana261: how embarrassing :| sorry..

lib/transaction.js
arcana261: ok.. will do..

lib/transaction.js
arcana261: ok... sorry I thought they would be in-significant..

lib/transaction.js
mickhansen: no worries, that's why there' code review :)

lib/transaction.js
mickhansen: Well in case `rollback` does in fact execute, it's best that we actually wait for it to finish before proceeding :)

arcana261: how about this? reporting error about setup is more important so we are going to ignore rollback error anyway... what's important is that we should "release" the connection which we will do regardless of any error during rollback!

```javascript
  }).catch(function (setupErr) {
    return self.rollback()
      .finally(function () {
        throw setupErr;
      });
  }).tap(function () {
```

mickhansen: Ah yes, that's actually more correct since we need to rethrow on a succesfull rollback aswell :)
Please put `finally` on the same line as `rollback` though.

arcana261:
ok Sir, fixed everything as you instructed to do, added new tests and all of them passed! :smile:

mickhansen:
Looks great!
Then i have only a single final request: Squash the commits to a single descriptive commit :)

PR #4493, opened 2015-09-14
Fixed 'define' method parameter in example for building instances

jessicalc:
Edits a potentially confusing typo in a code example.

**steps to repro:**
1. go to http://docs.sequelizejs.com/en/latest/docs/instances/
2. scroll to section that says "Built instances will automatically get default values when they were defined."

**expected:**
code example should say
```
var Task = sequelize.define('Task', {
  title: Sequelize.STRING,
  rating: { type: Sequelize.STRING, defaultValue: 3 }
})
```
**actual:**
code example has **Project** instead of **Task** as the first param in the `sequelize.define` method when defining the Task class. 
```
var Task = sequelize.define('Project', {
  title: Sequelize.STRING,
  rating: { type: Sequelize.STRING, defaultValue: 3 }
})
```

See screenshot of existing problem below:
image

### DIFF
--- a/docs/docs/instances.md
+++ b/docs/docs/instances.md
@@ -19,7 +19,7 @@ Built instances will automatically get default values when they were defined&col
 
 ```js
 // first define the model
-var Task = sequelize.define('Project', {
+var Task = sequelize.define('Task', {
   title: Sequelize.STRING,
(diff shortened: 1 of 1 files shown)


janmeier:
Woops - thanks :)

PR #4497, opened 2015-09-14
Create a shallow copy of scopeInclude in Model.$injectScope

### LINKED ISSUE #4470: Memory leak & maximum call stack exceeded using hasMany with scope
Considering the following model definitions:
```js
    name: Sequelize.STRING
});

var Project = sequelize.define('Project', {
    type: Sequelize.STRING,
    title: Sequelize.STRING
});

User.addScope('defaultScope', {
    include: [Project]
}, {
    override: true
});

Project.belongsTo(User);

User.hasMany(Project, {
    scope: {
        type: 'open'
    }
});
```
We create a user and a project:
```js
let user = yield User.create({
    name: 'John'
});
yield user.createProject({
    name: 'Project 1'
});
```
Then querying the user multiple times as following: 
```js
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
yield User.findById(user.id);
```
Result in a growing array `value` in `query-generator.js:1913` in `whereItemQuery` method. Logging the array `value` around line 1923 after the recursive call to `whereItemQuery`, we obtain the following: 
```sql
'((((((((((((("Projects"."type" = \'open\' AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\') AND "Projects"."type" = \'open\')',
  '"Projects"."type" = \'open\'' 
```
The size increase at each query.

Verdier:
Closes #4470.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2547,6 +2547,9 @@ Model.$injectScope = function (scope, options) {
         }
         return sameModel;
       })) {
+        // Do not inject the Model.$scope.include[] in options directly,
+        // create a shallow copy (#4470).
+        scopeInclude = _.clone(scopeInclude);
         options.include.push(scopeInclude);
       }
     });


Verdier:
@mickhansen this pull request fix #4470 and avoid to modify `Model.$scope.include[]`.

mickhansen:
That's probably a good idea aswell.
But ideally we should track down where the object is dirtied and make sure it's cloned before that.

Verdier:
Tests fails because of a timeout that's not linked to the pull request.

mickhansen:
Can you add a test for this so we don't have a regression?

Verdier:
@mickhansen ok non problem. I've never cloned the repos...

PR #4506, opened 2015-09-14
Foreign keys with allowNull: false with new default values

### LINKED ISSUE #2831: ON DELETE constraint should default to CASCADE if foreignKey has allowNull: false
```js
var Foo = this.sequelize.define('Foo', { 
  bar_id: { type: Sequelize.INTEGER, allowNull: false }
});
var Bar = this.sequelize.define('Bar', {});

Foo.belongsTo(Bar, { foreignKey: 'bar_id'});

return Bar.sync({force: true}).then(function () {
  return Foo.sync({force: true});
});
```

Will fail on MySQL (haven't tested others so far) because the cascade defaults to SET NULL which "obviously" (took me a while to figure it out) obviously doesn't work well with a field that can't be NULL'ed.

Sequelize should check the foreignkey attribute definition, and if allowNull is false it should perhaps either use CASCADE by default or turn off constraints.

Americas:
* CASCADE on hasOne/hasMany
* NO ACTION on belongsTo

Fixes #2831

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,8 @@
 - [ADDED] Expose Association constructor as `Sequelize.Association`
 - [FIXED] Calling set with dot.separated key on a JSON/JSONB attribute will not flag the entire object as changed [#4379](https://github.com/sequelize/sequelize/pull/4379)
 - [FIXED] instances returned from `bulkCreate` now has `isNewRecord: false` and should be updateable if using `returning: true` with dialects that support it.
+- [FIXED] Find with Include with a where clause generates wrong SQL [#3940](https://github.com/sequelize/sequelize/issues/3940)
+- [FIXED] ON DELETE constraint should default to CASCADE if foreignKey has allowNull: false] [#2831](https://github.com/sequelize/sequelize/issues/2831)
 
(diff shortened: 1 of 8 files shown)


janmeier:
Changes look good - could you add a couple of tests? Unit tests that just check the constraints are set to the right thing are fine, no need for integration tests for this change

Americas:
OK, I will add them ASAP, hopefully today!

Americas:
I love this git voodoo magic.. :laughing: 

Tests were added.

mickhansen:
Some breaks for mysql/pg

mickhansen:
Errors look unrelated though, trying a rerun

Americas:
the schema part is related to the change, I'm checking it now

Americas:
I changed the behavior of the associations in a way that the first association would set the defaults, and that is what broke the tests. Changed it back to the hasOne association doing a default override when it exists.

mickhansen:
Looks like it needs a rebase @Americas

Americas:
rebased

mickhansen:
Thank you for your work @Americas

PR #4509, opened 2015-09-14
Added documentation for nested creation (see #3386).

eddieajau:


docs/docs/associations.md
```diff
@@ -635,6 +635,104 @@ Series.hasOne(Video);
...
+
+Now we can create a project with multiple tags in the following way:
+
+```js
+Product.create({
+  id: 1,
+  title: 'Chair',
+  Tags: [
+    {id: 1, name: 'Alpha'},
+    {id: 2, name: 'Beta'}
```
mickhansen: Should probably remove `id` from these examples since nested creation currently does not work with existing elements.

docs/docs/associations.md
```diff
@@ -635,6 +635,104 @@ Series.hasOne(Video);
 Trainer.hasMany(Series);
 ```
 
+## Creating with associations
 
+New elements of an association can be created in one step, provided all elements are new.
```
mickhansen: "New elements of an association" reads a little odd to me, perhaps "An instance can be created with nested association / dependencies in one step."

janmeier:
Added in https://github.com/sequelize/sequelize/commit/930e6537386c9f6802d52e0ea3daa7be661efb57 - thanks @eddieajau !

eddieajau:
Sorry I didn't get back to those minor corrections. Work has been crazy and I wish I could help more.

Issue #4376, opened 2015-08-23
Support separate:true for belongsToMany association types


sorin:

sorin:
[![Bountysource](https://www.bountysource.com/badge/issue?issue_id=26103692)](https://www.bountysource.com/issues/26103692-support-separate-true-for-belongstomany-association-types?utm_source=26103692&utm_medium=shield&utm_campaign=ISSUE_BADGE)

sorin:
To further clarify the scope of this feature request, here's how separate:true was implemented for HasMany: https://github.com/sequelize/sequelize/commit/6ba2b67493c0d82f8fee310ac966cabb9047369d

This is because performance is very bad when querying for a model with a big number of includes. Lets say you want to query for a model and include 5 associations, each of them having 10 rows. The current implementation does a single query with 5 joins returning in a total of 10^5 = 100,000 rows to be parsed.

Specifying separate:true would do 6 queries that result in a total of 1 + 10*5 = 51 rows to be parsed.

mickhansen:
Original PR for reference: https://github.com/sequelize/sequelize/pull/4177

`include.seperate` benefits both in performance because of reduces cartesian products and also in being able to express a bit more complex includes than joins sometimes support.

`include.seperate` needs to do a UNION ALL join across multiple selects, so not sure when the breakpoint for performance is.

radhomero:
This huge feature is missing on the Docs, just saying

mickhansen:
@radhomero Well since it does exist for N:M yet there's not a lot to document :) Missing for hasMany though yes.

sgehly:
Hey guys, don't mean to grave-dig old PRs, but almost a year later are there any updates on this?

frnco:
`hasMany` PR seems like a pretty good starting point and the goals are pretty straight-forward.

I am interested in contributing.

PlasmaPower:
I'll restart #4525 with a common base with the `hasMany` separate code.

PlasmaPower:
What happened to the bounty on this issue? Did it expire?

sorin:
@PlasmaPower we really needed this 2 years ago, hence the bounty. We've worked around the issue ever since

foisonocean:
Any update? Is it possible to use `separate: true` in a m:n association now ?

DaAwesomeP:
@foisonocean this is a much requested feature, but it seems that nobody has yet been able to dedicate time to complete it.

[...]

PR #4525, opened 2015-09-16
implement #4376 - support for include:separate for belongsToMany

yotamofek:
Hey,

Implements #4376 

This is my first attempt at contribution.
Please let me know if there are any bugs or anything missing.

Thanks,
Yotam.


lib/model.js
```diff
@@ -1362,13 +1363,11 @@ Model.prototype.findAll = function(options) {
 };
 
 Model.$findSeperate = function(results, options) {
-  if (!options.include || options.raw || !results) return Promise.resolve(results);
+  if (!options.include || options.raw || _.isEmpty(results)) return Promise.resolve(results);
```
mickhansen: Why the change to `_.isEmpty`?

mickhansen: Nvm, to cover `results.length`

mickhansen:
I would love it if you could get closer to the HasMany implementation in https://github.com/sequelize/sequelize/pull/4177/files in terms of testing, would love to see both unit tests for BelongsToMany.get/set/etc and for the query generation.

lib/associations/belongs-to-many.js
```diff
@@ -298,6 +298,46 @@ var BelongsToMany = function(source, target, options) {
...
+        })
+          , instanceTargets = _.at(targets, _.pluck(instanceThroughs, association.otherKey));
+
+        return [sourceId, _(instanceTargets).compact().flatten().value()];
+      })
+      .zipObject()
+      .value();
+  }));
+};
+
```
mickhansen: Please refactor `injectGetter` to use this code now as done in the other include.seperate PR, otherwise we just have duplicate code.

Ideally you should also move the other methods like done in the other PR, but not stricly necessary.

lib/associations/belongs-to-many.js
```diff
@@ -298,6 +298,46 @@ var BelongsToMany = function(source, target, options) {
...
 
+BelongsToMany.prototype.get = function(instances, options) {
+  var throughModel = this.throughModel;
+
+  var throughOptions = _.assign({
+    where: _.set({}, this.foreignKey, {$in: _.pluck(instances, this.source.primaryKeyAttribute)}),
+    logging: options.logging
+  });
+
+  var association = this;
```
mickhansen: please follow the style guide and hoist variables.

lib/associations/belongs-to-many.js
```diff
@@ -298,6 +298,46 @@ var BelongsToMany = function(source, target, options) {
...
+  return throughModel.findAll(throughOptions).then(function (throughs) {
+    var otherKey = association.otherKey
+      , targetIds = _.chain(throughs).pluck(otherKey).uniq().value();
+
+    var targetOptions = _.chain({
+      where: _.set({}, association.target.primaryKeyAttribute, {$in: targetIds})
+    })
+        .assign(association.target.$optClone(options))
+        .omit(['include'])
+        .value();
```
mickhansen: wonky identation.

lib/associations/belongs-to-many.js
```diff
@@ -298,6 +298,46 @@ var BelongsToMany = function(source, target, options) {
...
+  }).then(_.spread(function(throughs, targets) {
+    targets = _.groupBy(targets, association.target.primaryKeyAttribute);
+    
+    return _.chain(instances)
+      .pluck(association.source.primaryKeyAttribute)
+      .map(function(sourceId) {
+        var instanceThroughs = _.filter(throughs, function (through) {
+          return through.get(association.foreignKey, {raw: true}) === sourceId;
+        })
+          , instanceTargets = _.at(targets, _.pluck(instanceThroughs, association.otherKey));
```
mickhansen: wonky identation.

lib/model.js
```diff
@@ -1391,19 +1390,23 @@ Model.$findSeperate = function(results, options) {
...
+          result.set(
+            include.association.as,
+            map[result.get(include.association.source.primaryKeyAttribute)]
+          );
+        });
       });
-    });
-  }).return(original);
+    }
+  }.bind(this)).return(original);
```
mickhansen: the `bind` is not required as far as i can tell?

lib/associations/belongs-to-many.js
```diff
@@ -298,6 +298,46 @@ var BelongsToMany = function(source, target, options) {
 
 util.inherits(BelongsToMany, Association);
 
+BelongsToMany.prototype.get = function(instances, options) {
+  var throughModel = this.throughModel;
+
+  var throughOptions = _.assign({
+    where: _.set({}, this.foreignKey, {$in: _.pluck(instances, this.source.primaryKeyAttribute)}),
```
mickhansen: Please use `instance.get` with `raw: true`, object lookups are not used internally.

lib/associations/belongs-to-many.js
```diff
@@ -298,6 +298,46 @@ var BelongsToMany = function(source, target, options) {
...
+      , targetIds = _.chain(throughs).pluck(otherKey).uniq().value();
+
+    var targetOptions = _.chain({
+      where: _.set({}, association.target.primaryKeyAttribute, {$in: targetIds})
+    })
+        .assign(association.target.$optClone(options))
+        .omit(['include'])
+        .value();
+
+    return Utils.Promise.join(throughs, _.isEmpty(targetIds) ? null : association.target.findAll(targetOptions));
```
mickhansen: Would it not be possible to implement so there's only a single query needed?

test/integration/include/seperate.test.js
```diff
@@ -61,6 +61,44 @@ if (current.dialect.supports.groupedLimit) {
...
+          }).then(function (users) {
+            expect(users[0].groups).to.be.ok;
+            expect(users[0].groups.length).to.equal(1);
+            expect(users[1].groups).to.be.ok;
+            expect(users[1].groups.length).to.equal(2);
+            expect(users[1].groups).to.include(users[0].groups[0]);
+            expect(sqlSpy).to.have.been.calledThrice;
+          });
+        });
+      });
```
mickhansen: Need a test with both `seperate` and `limit`.

yotamofek:
@mickhansen - thank you very much for the review. I will fix all the styling issues and refactor as you requested.

Also- I would like to run a little `sed` to replace `seperate` with `separate`, as the former is quite wrong, and the lack of consistency is very confusing. Please let me know if this is something you support.

lib/associations/belongs-to-many.js
yotamofek: Yes, it's possible using a `JOIN` clause, but that would basically be like implementing a non-separated query on the hasMany between the Through model and the target model.
The major downside is that target objects that are referenced in multiple rows in the through table will be retrieved multiple times.

mickhansen: Sequelize doesn't currently support duplicate entries in the join model for a source and target, although that has been a feature request before with better unique: false support.

Ideally for a belongsToMany association with `unique: true` (or not `unique: false`) rather we should use an optimized single query.

It should be possible to do an INNER JOIN and a UNION ALL across a set of selects with inner joins.

/cc @janmeier

mickhansen:
Go ahead with the typo fix, it should only be in the docs, we've fixed it everywhere we've found.

test/integration/include/seperate.test.js
yotamofek: A `limit` clause on a `belongsToMany` association is a case of [greatest-n-per-group](http://stackoverflow.com/questions/tagged/greatest-n-per-group) and is not a trivial thing to implement, nor does it have one standard and efficient way of doing it.
I think really the only way to ubiquitously implement it is with a separate query the through model for each source instance.

lib/associations/belongs-to-many.js
yotamofek: Multiple duplicate records will be returned even when `unique: true` for the through model, when many sources are associated to the same target. i.e., when retrieving a list of users and their groups, a certain group will have a duplicate row for each member of that group.

test/integration/include/seperate.test.js
mickhansen: Not in the case of entries being unique, which is the common case.

lib/associations/belongs-to-many.js
mickhansen: No, that should not ever happen.

mickhansen: I don't believe so.

```
SELECT * FROM 
  (SELECT * FROM join_table INNER JOIN right_table ON ... WHERE left_id = 1 LIMIT 10) 
  UNION ALL
  (SELECT * FROM join_table INNER JOIN right_table ON ... WHERE left_id = 2 LIMIT 10) 
  .. etc
```

That's how the hasMany implementation is.

mickhansen: For join_table unique given a left_id there will be only be a single entry per right_id.

mickhansen: I could definitely be wrong though, but i'll need you to cut it out for me :D

test/integration/include/seperate.test.js
yotamofek: I'll have to disagree. The `unique: true` means, as far as I understand, that there can be no two rows containing the same source **and** target. Otherwise, the M:M definition of belongsToMany is reduced to 1:M or M:1.

Let's take the following scenario: you want to get a list of users, and for each user the 3 groups he most recently joined (i.e. `limit: 3` and ordering by the timestamp of the through model).
This is a classic case of greatest-n-per-group, and will happen with the simplest model definition, and a `unique: true` param on the through model.

lib/associations/belongs-to-many.js
yotamofek: What you said about only having a single `right_id` per `left_id` is true in the case of a `hasMany`, but will mean that the source can only be associated with a single target in a `belongsToMany`, which defeats the purpose.
BTW, the `UNION ALL` trick will work, but is essentially the same as running a query per source, which is very inefficient (but can solve the case with the `limit: ...`).

Would love a second opinion since I'm not completely certain about what I'm saying...

test/integration/include/seperate.test.js
mickhansen: Correct, `unique: true` enforces uniqueness between source and target. belongsToMany is effectively two sets of 1:M and you only need the one side since you already have all the IDs.

@yotamofek you're forgetting that in the scenario of `include.seperate` you already have the list of users preloaded, and as such you only need to load the join table and the target table, making it a 1:1 lookup for each row.

yotamofek: I haven't forgotten about it, but what I'm saying still applies-
if we do a 1:M on the through->target, a group with 100 users will be duplicated on 100 rows.
Since this is basically a `hasMany` association between the through to the target, and the user requested a `separate: true` query, then I think the `hasMany` association should also behave with `separate: true`. BTW, this gives me the interesting idea of calling `hasMany.get` from `belongsToMany.get`.

Am I missing something?

mickhansen: A group with a 100 users will have exactly a 100 rows, easy enough to limit since we're doing individual select statements for each group and then unioning. Did you check out my example query?

It's basically a hasMany with a belongsTo include yes. (and the belongsTo include is what needs to be returned).

`hasMany` supports `seperate: true` and `limit`, check the original PR.

yotamofek: Your example query is a good solution for when you have a `limit` or `orderBy` clause, but when this is not needed (i.e. all targets should be retrieved), the complexity of the query, or rather the number of queries/sub-queries, will be `O(n)`, instead of `O(1)` when issuing an intermediate query. Also, the problem that the `hasMany {separate: true}` solves will re-appear in the pseudo-`hasMany` between the through and the target.

I think `limit` and `orderBy` should not be allowed for `belongsToMany {separate: true}`, as there is no clear-cut winning implementation.

mickhansen: It should only use that query when there IS a limit, again please take a look at my original implementation for hasMany.

I strongly believe that limit should be supported, it's a large part of why we want `separate` support (other than increased performance of course).

yotamofek: Agreed. I will refactor accordingly.

BTW, I thought your original PR simply forced `separate: true` when there is a limit, now I've re-read it and understood what you meant.
I need to try it, but I think that maybe calling `hasMany.get` on the association between the intermediate model and the target is the best way of re-using your code.

Thanks for your time!

mickhansen: I do believe it forces `include.separate` on `include.limit`? It doesn't do it based on there being a general find limit though.

@yotamofek It might be possible to reuse by adding an include to the HasMany.get call and then mapping that back, since yes, it effectively becomes a 1:M lookup when we already have the source (which we do).

sorin:
great work @yotamofek and @mickhansen !!

lib/associations/belongs-to-many.js
mickhansen: `UNION ALL` is almost the same as running a query per source, except that we don't need to roundtrip and setup connections.

It will still work with multiple targets for source as far as i can tell. The combination of `right_id` and `left_id` is still unique in belongsToMany when `unique` is true (which is the common case and the default).

sorin:
@mickhansen any particular reason this PR is not merged? (except maybe it has conflicts that need to be solved)? this is a very useful feature and we'd love to award the bounty to @yotamofek

mickhansen:
@sorin Have you read the comments? There are a few issues to resolve.

yotamofek:
Was away on holiday, glad to see there's still interest. :)
Will continue working on it this week.

NHMD:
+1 any news on this one?

yotamofek:
@NHMD I don't have enough spare time for this feature right now, hopefully I can get back to it soon, but I can make no promises.

diogocdias:
@yotamofek Many thanks for your contribution! 
Looking forward for this feature. Until then I'll have to use some hacks on my project.

Djalmar:
+1, any news ?

yotamofek:
Hey,
thanks for the interest, everyone :)

I have some time to work on this PR this week, hopefully I will push some changes discussed with @mickhansen and resolve the conflict soon.

yotamofek:
Sorry, I made a bit of a mess. Tests currently fail.

jeppester:
Thank you @yotamofek!

Using you branch I've just decreased the response time a project by a HUGE margin - one of the routes has gone from 1.5s down to 80ms.

yotamofek:
That's great to hear!

I had a test failing on MSSQL, so it took a little work to be able to replicate locally (some things even docker can't solve). Turns out it was a small bug in the creation of the through table models, which only appeared on MSSQL because it is case-insensitive in object names. Now the tests pass.

Now one of the tests failed because of a timeout, and I hardly think it had to do with the code :(

@mickhansen - your review please?

lib/associations/belongs-to-many.js
```diff
@@ -302,6 +304,71 @@ var BelongsToMany = function(source, target, options) {
...
+  var throughModel = this.throughModel;
+
+  var throughOptions = _.assign({
+    where: _.set({}, this.foreignKey, {$in: _.map(instances, this.source.primaryKeyAttribute)}),
+    logging: options.logging
+  });
+
+  var association = this;
+
+  if(_.isUndefined(options.limit) && _.isUndefined(options.orderBy)) {
```
mickhansen: `if (options.limit)` should be sufficient should it not

mickhansen: also please follow styling everywhere else and add a space after if

mickhansen:
This is still using two passes, did you attempt reusing the hasMany code as we previously discussed?

lib/associations/belongs-to-many.js
```diff
@@ -302,6 +304,71 @@ var BelongsToMany = function(source, target, options) {
...
+
+      var targetOptions = _.chain({
+            where: _.set({}, association.target.primaryKeyAttribute, {$in: targetIds})
+          })
+          .assign(association.target.$optClone(options))
+          .omit('include', 'includeMap', 'includeNames')
+          .value();
+
+      return Utils.Promise.join(throughs, _.isEmpty(targetIds) ? null : association.target.findAll(targetOptions));
+    }).then(_.spread(function (throughs, targets) {
```
mickhansen: bluebird has a spread function

lib/associations/belongs-to-many.js
```diff
@@ -302,6 +304,71 @@ var BelongsToMany = function(source, target, options) {
...
+                , instanceTargets = _.at(targets, _.map(instanceThroughs, association.otherKey));
+
+            return [sourceId, _(instanceTargets).compact().flatten().value()];
+          })
+          .fromPairs()
+          .value();
+    }));
+  } else {
+    return Utils.Promise.resolve(instances)
+      .map(function (instance) {
```
mickhansen: general style calls for the method to be on the same line for first parts of promise chains

lib/associations/belongs-to-many.js
```diff
@@ -302,6 +304,71 @@ var BelongsToMany = function(source, target, options) {
...
+        var throughOptions = _.assign(
+          {
+            include: [
+              {
+                association: association.target.hasOne(throughModel),
+                where: _.set({}, association.foreignKey, instance.get(association.source.primaryKeyAttribute))
+              }
+            ]
+          },
+          _.pick(options, 'order', 'limit', 'logging')
```
mickhansen: we want users to be able to pass a top level options object that makes it all the way down regardless.
this is helpfull to plugins, performance tracing etc

lib/associations/belongs-to-many.js
```diff
@@ -302,6 +304,71 @@ var BelongsToMany = function(source, target, options) {
...
+          .value();
+    }));
+  } else {
+    return Utils.Promise.resolve(instances)
+      .map(function (instance) {
+        var throughOptions = _.assign(
+          {
+            include: [
+              {
+                association: association.target.hasOne(throughModel),
```
mickhansen: We could probably precreate this association internally in the N:M association, i believe we already are actually

yotamofek: We *are* creating an association from the source to the through model, not one for the through model to the target.
I definitely agree that this association could be created preeminently, as it makes full logical sense. Will refactor.

lib/associations/belongs-to-many.js
yotamofek: So I'll switch to blacklisting properties instead of whitelisting. :+1

lib/associations/belongs-to-many.js
yotamofek: Yeah, I guess `orderBy` can be implemented 'naively', i.e. will work naturally if the order property is propagated to the sub-query. Will test and remove constraint.

yotamofek:
@mickhansen - thanks!

:+1: for all the styling issues.

As for re-using the hasMany code, it would require making some changes to your code. Whatever you prefer.

lib/associations/belongs-to-many.js
yotamofek: Caveat: with multiple order clauses, this logic fails and the queries must be separated. (I think.)

lib/associations/belongs-to-many.js
mickhansen: Yep!

lib/associations/belongs-to-many.js
mickhansen: I think we're already creating some associations in the N:M constructor, we use one in include

lib/associations/belongs-to-many.js
mickhansen: Actually, we should have this for when the user specifically says include.order and not using the nested order syntax.

test/integration/include/seperate.test.js
```diff
@@ -65,6 +65,89 @@ if (current.dialect.supports.groupedLimit) {
...
+          return Promise.join(
+            User.bulkCreate([1, 2, 3].map(function(id) {
+              return {id: id};
+            })),
+            Group.bulkCreate([{id: 1, precedence: 5}, {id: 2, precedence: 10}])
+          ).then(_.spread(function (users, groups) {
+            return Promise.join(
+              users[0].addGroup(groups[0]),
+              users[1].addGroups(groups)
+            );
```
mickhansen: Please ammend so that users have atleast one more group than the requested limit, otherwise the test could be a false positive.

test/integration/include/seperate.test.js
```diff
@@ -65,6 +65,89 @@ if (current.dialect.supports.groupedLimit) {
...
+              return instance.get(instance.Model.primaryKeyAttribute);
+            }
+
+            expect(users[0].groups).to.be.ok;
+            expect(users[0].groups).to.have.length(1);
+            expect(users[1].groups).to.be.ok;
+            expect(users[1].groups).to.have.length(1);
+            expect(users[2].groups).to.be.empty;
+            expect(getPrimaryKey(users[1].groups[0])).to.not.equal(getPrimaryKey(users[0].groups[0]));
+            expect(sqlSpy.callCount).to.equal(4);
```
mickhansen: Would be not only be expecting 2 queries here? Or is it generating 1 + N queries?

test/integration/include/seperate.test.js
yotamofek: `users[1]` gets both groups in `groups`, and the limit is 1.

test/integration/include/seperate.test.js
yotamofek: It is generating M*1:N queries. I'll refactor to generate these as sub-queries.

lib/associations/belongs-to-many.js
yotamofek: 1) I can't find the creation of the association in the [through model definition](https://github.com/yotamofek/sequelize/blob/belongs.to.many.separate/lib/associations/belongs-to-many.js#L161)
2) Now that I think about it, the through model can be a custom pre-defined model, and then we cannot guarantee that it contains this association. Or should it be coerced somehow?

test/integration/include/seperate.test.js
mickhansen: Right you are!

test/integration/include/seperate.test.js
mickhansen: Using the hasMany code which uses UNIONS isnt much better perf wise, but it saves roundtrips and you don't take up all connections in the pools.

yotamofek: I completely agree, I think using the `hasMany.get` is the best scenario. It took me some time to realize a `groupedLimit` option generates a series of `UNION JOIN`ed subqueries.

One thing to remember is that on VERY big data-sets, a complex query with a lot of subqueries can cause a database to freeze. I've seen it happen in the past.

yotamofek:
@mickhansen -
I cannot use `HasMany.get` directly on the Through->Target association, since it requires having previously retrieved the list of targets.
What I managed, is to call a `findAll` on the target model, including and limiting by the through model. This is a separated query in itself, so it creates 3 queries overall instead of the possible 2.
IMO, this might be a beneficial outcome, since it will prevent duplicate targets from transferring from the DB to the client, which may amass to large amounts of data. if, for instance, a 100 users are all part of the same 100 groups, each group would be returned 100 times without the intermediate query.
What do you think?

Still waiting for tests to run...

PlasmaPower:
What's the status of this PR? I was hoping to work on the related issue, but I didn't notice that a PR was already open.

mickhansen:
@yotamofek It doesn't necessarily require a previous list, you can query the targets, include the join model and then filter on the join model for a single query.

yotamofek:
@mickhansen that's true. Let me refactor, and I'll put up a new version soon.

molerat619:
Cool, thanks for your support. (When) will this be merged to master?

yotamofek:
@molerat619 - there is still work to be done and I don't believe I'll have time for it in the coming weeks.

denghongcai:
@yotamofek any progress?

relwell:
Hey there @yotamofek -- still feeling the pain of this one. Any luck finding time?

aryeharmon:
still an issue guys!

Djalmar:
2 years in the waiting

yokrysty:
omg 2 years, I've just discovered recently that separate is not working with n:m :(
is sequelize or this implementation dead? should i rewrite this project with another ORM?

MMuzaffar:
Any update on this?

elharony:
Any updates on this PR? We really need it 🙏

sorin:
> Any updates on this PR? We really need it 🙏

Any century now :)

yotamofek:
Hey guys,
I'm sorry for the disappointment, I feel like more people had been following this PR than I had known and/or planned for.

This opinion is a bit harsh, but after a couple of back and forths on this PR with the library's maintainer, I lost my faith in them - both the library and the maintainer. I moved to using `knex` which served all my needs (include N:M which this PR tried to land!!) and whose code seemed a lot more maintainable IMHO. Nowadays I tried to avoid Javascript as much as I can, in general. :)

The library's author's comments about my PR have been historically documented here, and anyone who's interested can judge them for themselves, but I feel like the his insistence of providing a "good enough" solution, which he deemed would be acceptable in most cases, spoke to me of a somewhat inexperienced world-view. My trust in this library definitely eroded. Also - the fact that it's been 4 years and this feature (which is really a pretty basic feature when building an ORM for an RDBMS) has not been built into the library - should speak volumes.

PR #4526, opened 2015-09-16
added addProjects and addUser instance methods to documentation for b…

bjones6:
The documentation for belongs-to-many relationships was slightly confusing and did not clearly convey the existence of singular and plural setters ('addProject', 'addProjects', 'addUser', 'addUsers') for both sides of the relationship.

Personally ran into this problem and thought I'd make a small addendum to the docs.


janmeier:
Thanks :)

janmeier:
Actually, the two methods map to the same implementation under the hood, but its always good to mention both

Issue #4527, opened 2015-09-16
Enable .validate(options.skip) to skip specific model/custom validators


superclarkk:
This is a feature request.

It's already possible to skip field validations when calling [.validate([options])](http://sequelize.readthedocs.org/en/latest/api/instance/index.html?highlight=skip#validateoptions-promiseerrorsvalidationerrorundefined), by specifying an array of field names in the `options.skip` property.

It would be very useful to be able to skip model/custom validators too, by specifying their names/paths in the same list.

ikraloz:
I'm not sure if this is the right place to ask for this, but can you provide me with an example to use this feature? I haven't been able to find any documentation refering to this feature.

PR #4528, opened 2015-09-16
Feature #4527 "skip" option for custom validators

superclarkk:
Enable .validate(options.skip) to skip specific model/custom validators #4527
Return false to exit iteration early when using _.each and _.forIn , as per lodash docs

### DIFF
--- a/lib/instance-validator.js
+++ b/lib/instance-validator.js
@@ -133,13 +133,13 @@ InstanceValidator.prototype._builtinValidators = function() {
  * @private
  */
 InstanceValidator.prototype._customValidators = function() {
-
-
   var validators = [];
   var self = this;
-  Utils._.each(this.modelInstance.$modelOptions.validate, function(validator,
-    validatorType) {
-
+  Utils._.each(this.modelInstance.$modelOptions.validate, function(validator, validatorType) {
+    if (self.options.skip.indexOf(validatorType) >= 0) {
+      return;
+    }
+    
     var valprom = self._invokeCustomValidator(validator, validatorType)
       // errors are handled in settling, stub this
       .catch(function() {});


mickhansen:
Concerning BC: i don't think this constitutes a breaking change since i don't believe you would have model validators named the same as attributes.

/cc @janmeier

lib/instance-validator.js
```diff
@@ -108,7 +108,7 @@ InstanceValidator.prototype._builtinValidators = function() {
   var validators = [];
   Utils._.forIn(this.modelInstance.rawAttributes, function(rawAttribute, field) {
     if (self.options.skip.indexOf(field) >= 0) {
-      return;
+      return false;
```
mickhansen: Is the iteration supposed to be exited early though? `skip` might match the first field, should still validate all the other fields.

superclarkk: It was exiting early with just "return", but the lodash docs state that it should be [return false](https://lodash.com/docs#forIn). I changed this line just so that it was consistent with my main changes.

mickhansen: Are you absolutely sure about that? lodash docs says you have to explicitely return false.
If it's currently exiting the loop early with just `return` that would be a bug.

mickhansen: https://github.com/lodash/lodash/blob/master/lodash.js#L3352 it tests for explicit false, `return undefined` would not break loop.

superclarkk:
@mickhansen If by "attributes" you mean fields (table columns), then you are correct. An error is thrown if a custom validator (model validator) and an attribute have the same name.

mickhansen:
@superclarkk i did mean model/table fields, thanks for confirming :)

mickhansen:
Published in 3.9.0

PR #4530, opened 2015-09-17
static dialect requires. Fixes #4517

### LINKED ISSUE #4517: Make Dialect Requiring Static
### The bundling applications fail to include the dialects.

[The Code](https://github.com/sequelize/sequelize/blob/master/lib/sequelize.js#L179):
```js
try {
    var Dialect = require('./dialects/' + this.getDialect());
  } catch (err) {
    throw new Error('The dialect ' + this.getDialect() + ' is not supported. ('+err+')');
  }
```
This would better be handled with a switch case:
```js
try {
    var Dialect;
    swtich(this.getDialect())
    {
         case '...':
             Dialect = require('./dialects/...'); break; 
         .
         .
         .
    }
  } ...
```

This isn't so hard to fix, should I make a pull request? or do you guys have time yourselves to fix this? ^^

subjectix:
I couldn't run the tests, I would appreciate it if someone does it.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -177,7 +177,26 @@ var Sequelize = function(database, username, password, options) {
   };
 
   try {
-    var Dialect = require('./dialects/' + this.getDialect());
+    var Dialect;
+    // Requiring the dialect in a switch-case to keep the
+    // require calls static. (Browserify fix)
+    switch (this.getDialect()){
+      case 'mariadb':
+        Dialect = require('./dialects/mariadb');
+        break;
+      case 'mssql':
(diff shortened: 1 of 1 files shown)


janmeier:
Thanks :)

subjectix:
You are welcome ^^

mickhansen:
Published in 3.9.0

yonjah:
@alitaheri Wondering if this still being used.
Seem a bit like a hack that should have been solved by Browserify either by offering some kind of
config option or partly parse those dynamic properties (Webpack seem to solve this)

PR #4532, opened 2015-09-17
Fixed how an included where is joined so it generates correct SQL code.

### LINKED ISSUE #3940: Find with Include with a where clause generates wrong SQL
I am trying to do a find with a nested model with a where clause, as per the example below.

```javascript
var CD = sequelize.define("CD", {
    id    : {
        type         : Sequelize.UUID,
        field        : "a",
        defaultValue : Sequelize.UUIDV4,
        primaryKey   : true,
    },
    extra : {
        type  : Sequelize.JSONB,
        field : "ctrd_extra",
    },
}, {
    tableName    : "test1",
    name         : {
        singular : "cD",
        plural   : "cDs",
    },
});

var CQ = sequelize.define("CQ", {
    a : {
        type      : Sequelize.INTEGER,
        field     : "a",
        allowNull : false,
    },
    b     : {
        type       : Sequelize.STRING,
        field      : "b",
        primaryKey : true,
        allowNull  : false,
    },
}, {
    tableName    : "test2",
    paranoid     : false,
    name         : {
        singular : "cQ",
        plural   : "cQs",
    },
});
CQ.removeAttribute("id");

CD.hasMany(CQ, {foreignKey : "c"});
CQ.belongsTo(CD, { foreignKey : "c", primaryKey : true });

sequelize
.sync({force: true})
.then(function() {
    return CD.create({})
})
.then(function (cd) {
    return cd.createCQ({ a : 1, b : "bbb" });
})
.then(function () {
    return CD.find({
        include : [
            {
                model    : CQ,
                where    : { b : "bbb" },
                required : true,
            },
        ],
    });
})
.then(function getQuantities(result) {
    console.log(result);
});
```

This generates the following SQL for the select, which is wrong because of line 16 : `WHERE "CD"."id" = "cQs"."c" `. `"CD"."id"` does not exist at that level, it should be `"CD"."a"`.

```SQL
 SELECT "CD".*,
        "cQs"."a"          AS "cQs.a",
    	"cQs"."b"          AS "cQs.b",
        "cQs"."created_at" AS "cQs.created_at",
        "cQs"."updated_at" AS "cQs.updated_at",
        "cQs"."c"          AS "cQs.c" 
   FROM (     SELECT "CD"."a"          AS "id",
                    "CD"."ctrd_extra" AS "extra",
                    "CD"."created_at",
                    "CD"."updated_at",
                    "CD"."deleted_at" 
               FROM "test1" AS "CD" 
              WHERE "CD"."deleted_at" IS NULL 
                AND (     SELECT "c" 
                            FROM "test2" AS "cQs" 
                           WHERE "CD"."id" = "cQs"."c" 
                             AND "cQs"."b" = 'bbb' 
                           LIMIT 1 
                    ) IS NOT NULL 
              LIMIT 1
        ) AS "CD" 
  INNER 
   JOIN "test2" AS "cQs" 
     ON "CD"."id" = "cQs"."c" 
    AND "cQs"."b" = 'bbb';
```

Throws the Error: Unhandled rejection SequelizeDatabaseError: column CD.id does not exist.

I am using postgres 9.4 and sequelize 2.1.3.

Americas:
It now uses the field name instead of the model name.

Fixes #3940.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1268,7 +1268,7 @@ var QueryGenerator = {
               , subQueryWhere;
 
             associationWhere[association.identifierField] = {
-              $raw: self.quoteTable(parentTable) + '.' + self.quoteIdentifier(association.source.primaryKeyAttribute)
+              $raw: self.quoteTable(parentTable) + '.' + self.quoteIdentifier(association.source.primaryKeyField)
             };
 
             if (!options.where) options.where = {};
--- a/test/integration/include/find.test.js
+++ b/test/integration/include/find.test.js
@@ -97,6 +97,43 @@ describe(Support.getTestDialectTeaser('Include'), function() {
       });
     });
 
+    it('should include a model with a where clause when the PK field name and attribute name are different', function() {
+      var User = this.sequelize.define('User', {
+        id: {
+          type: DataTypes.UUID,
+          defaultValue: Sequelize.UUIDV4,
+          field: 'main_id',
+          primaryKey: true
+        }
+      })
+        , Task = this.sequelize.define('Task', {
+          searchString: { type: DataTypes.STRING }
+        });
+
+      User.hasMany(Task, {foreignKey: 'userId'});
+      Task.belongsTo(User, {foreignKey: 'userId'});
+
+      return this.sequelize.sync({
+        force: true
+      }).then(function() {
+        return User.create();
+      }).then(function(user) {
+        return Task.bulkCreate([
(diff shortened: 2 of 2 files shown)


janmeier:
LGTM - but needs tests

Americas:
@janmeier I added a test, but I have no clue if it is placed in the right place. Please advise :stuck_out_tongue:

mickhansen:
LGTM, re-running build (travis fuck up), test looks good.

Americas:
still one test that did not run, also because of travis fuck up

Issue #4119, opened 2015-07-15
Bug when inserting in multiple schema


Uhsac:
Hello,

I think I find a bug in sequelize today (I think it's a bug).

In my project (using express and sequelize) I want to insert value in multiple schema (I use postgres) depending on route parameter.

So I have something like these : 

```javascript
User.schema(req.params.schemaName).create({ name: req.params.schemaName })
.then(function(user) {
  res.json(user);
});
```

But when I request this route it don't always respect the schema name.

I reproduce the problem in a small project using sequelize 3.3.2 to help you to understand.

Here is the code : 

```javascript
var async = require('async');
var request = require('request');
var express = require('express');
var Sequelize = require('sequelize')

var app = express();

var sequelize = new Sequelize('test', 'test', 'test', {
  host: '192.168.99.100',
  dialect: 'postgres'
});

var User = sequelize.define('User', {
  name: Sequelize.STRING
});

sequelize.dropAllSchemas().then(function() {
  sequelize.createSchema('alpha').then(function() {
    User.sync({ force: true, schema: "alpha" });
  });
  sequelize.createSchema('beta').then(function() {
    User.sync({ force: true, schema: "beta" });
  });
});

app.post('/user/:schemaName', function(req, res) {
  User.schema(req.params.schemaName).create({ name: req.params.schemaName }).then(function(user) {
    res.json(user);
  });
});

app.listen(8000);

setTimeout(function() {
  async.parallel([
    function(cb) {
      async.times(10, function(i, next) {
        request.post("http://localhost:8000/user/alpha");
        next();
      }, function() { cb(); });
    },
    function(cb) {
      async.times(10, function(i, next) {
        request.post("http://localhost:8000/user/beta");
        next();
      }, function() { cb(); });
    }
  ]);
}, 100);
setTimeout(function() {
  User.schema("alpha").count().then(function(alphaCount) {
    console.log("There is " + alphaCount + " alpha user.");
  });
  User.schema("beta").count().then(function(betaCount) {
    console.log("There is " + betaCount + " beta user.");
  });
}, 1000);
```

And when I execute this code, this is what I get :

```console
Executing (default): DROP SCHEMA alpha CASCADE;
Executing (default): DROP SCHEMA beta CASCADE;
Executing (default): CREATE SCHEMA alpha;
Executing (default): CREATE SCHEMA beta;
Executing (default): DROP TABLE IF EXISTS "Users" CASCADE;
[...]

PR #4534, opened 2015-09-17
Create query and dialect options for prepending schema search_path in Postgres

mliszewski:
When you have a multi-tenant application that has the same database models across multiple schemas the current sequelize model paradigm forces you to create a separate model for each schema that is identical except for the schema property of the tableName object. This change adds a dialect option called prependSearchPath that will allow you to pass a searchPath option to any query that will be prepend the query with "SET search_path to ". If the prependSearchPath options is enabled and no searchPath query option is provided it will use DEFAULT which should be the same behavior as before.  By supporting a searchPath option I can now define a common set of models and use them across tenant schemas by passing the correct searchPath. If the prependSearchPath dialect option is not explicitly passed as true than the code works exactly as it did before.

On the surface this functionality would appear to be possible by using model.schema() but this does not work well when the same model needs to be used against multiple schemas concurrently in the same application as it changes the schema option on the model itself which can lead to inconsistent results. Specifically the following use cases do no work well.

1. Create an instance with model.schema('a').build, create another instance with model.schema('b').build, try and save the first instance. The expected behavior would be that it would be saved in schema 'a' but it will actually be saved in schema 'b'.

2. Since calling .schema() changes the schema option on the model itself this can lead to unexpected behavior if schema is not specified on every usage of the model. If i do model.schema('a').findAll and then later do model.findAll it will use whatever the last schema value that was set rather than a default schema. While this may be expected behavior since .schema() is essentially just a glorified setter of the schema option it is still dangerous and could lead to mixed tenant data.

3. Even when chaining .schema() directly a small window exists where if you were calling .schema() on the same model concurrently you could get unexpected results.

### DIFF
--- a/test/integration/model/searchPath.test.js
+++ b/test/integration/model/searchPath.test.js
@@ -1,0 +1,469 @@
+'use strict';
+
+/* jshint -W030 */
+/* jshint -W110 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../support')
+  , DataTypes = require(__dirname + '/../../../lib/data-types');
+var SEARCH_PATH_ONE = 'schema_one,public';
+var SEARCH_PATH_TWO = 'schema_two,public';
+
+var current = Support.createSequelizeInstance({
+  dialectOptions: {
+    prependSearchPath: true
+  }
+});
+
+var locationId;
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+  if (current.dialect.supports.searchPath) {
+
+    describe('SEARCH PATH', function() {
+      before(function() {
+        this.Restaurant = current.define('restaurant', {
+            foo: DataTypes.STRING,
+            bar: DataTypes.STRING
+          },
+          {tableName: "restaurants"});
+        this.Location = current.define('location', {
+            name: DataTypes.STRING
+          },
+          {tableName: "locations"});
+        this.Employee = current.define('employee', {
+            first_name: DataTypes.STRING,
+            last_name: DataTypes.STRING
+          },
+          {tableName: "employees"});
+        this.Restaurant.belongsTo(this.Location,
+          {
+            foreignKey: 'location_id',
+            constraints: false
+          });
+        this.Employee.belongsTo(this.Restaurant,
+          {
+            foreignKey: 'restaurant_id',
+            constraints: false
+          });
+        this.Restaurant.hasMany(this.Employee, {
+          foreignKey: 'restaurant_id',
+          constraints: false
+        });
+      });
+
+
+      beforeEach('build restaurant tables', function() {
+        var Restaurant = this.Restaurant;
+        return current.createSchema('schema_one').then(function() {
+          return current.createSchema('schema_two');
+        }).then(function() {
+          return Restaurant.sync({force: true, searchPath: SEARCH_PATH_ONE});
+        }).then(function() {
+          return Restaurant.sync({force: true, searchPath: SEARCH_PATH_TWO});
+        }).catch(function(err) {
+          expect(err).to.be.null;
+        });
+      });
+
+      afterEach('drop schemas', function() {
+        return current.dropSchema('schema_one').then(function() {
+          return current.dropSchema('schema_two');
+        });
+      });
+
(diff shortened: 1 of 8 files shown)


test/integration/model/searchPath.test.js
```diff
@@ -0,0 +1,466 @@
...
+          }).then(function(restaurant) {
+            expect(restaurant).to.not.be.null;
+            expect(restaurant.foo).to.equal('two');
+          });
+        });
+      });
+
+      describe('concurency tests', function() {
+        it('should build and persist instances to 2 schemas concurrently in any order', function() {
+          var Restaurant = this.Restaurant;
```
mliszewski: @mickhansen This use case in particular fails when trying to use .schema() with the code below:

```
it('should but wont build and persist instances to 2 schemas concurrently in any order', function() {
          var Restaurant = this.Restaurant;

          var restaurauntModelSchema1 = Restaurant.schema('schema_one').build({bar: 'one.1'});
          var restaurauntModelSchema2 = Restaurant.schema('schema_two').build({bar: 'two.1'});

          return restaurauntModelSchema1.save()
            .then(function() {
              restaurauntModelSchema1 = Restaurant.schema('schema_one').build({bar: 'one.2'});
              return restaurauntModelSchema2.save();
            }).then(function() {
              return restaurauntModelSchema1.save();
            }).then(function() {
              return Restaurant.schema('schema_one').findAll();
            }).then(function(restaurantsOne) {
              expect(restaurantsOne).to.not.be.null;
              expect(restaurantsOne.length).to.equal(2);
              restaurantsOne.forEach(function(restaurant) {
                expect(restaurant.bar).to.contain('one');
              });
              return Restaurant.schema('schema_two').findAll();
            }).then(function(restaurantsTwo) {
              expect(restaurantsTwo).to.not.be.null;
              expect(restaurantsTwo.length).to.equal(1);
              restaurantsTwo.forEach(function(restaurant) {
                expect(restaurant.bar).to.contain('two');
              });
            });
        });
      });
```

mickhansen:
`.schema()` *should* actually clone the model allowing for parallel use, if they don't there's definitely a bug/something missing about that (which thinking of it might be a known issue that was never handled)

janmeier:
https://github.com/sequelize/sequelize/issues/4119

mickhansen:
Figured there was an existing issue @janmeier

mliszewski:
I agree that this is an issue with the .schema() functionality. The issue with cloning the model to set the schema is that you introduce a large memory problem. If you have a model that needs to be used across 1000s of schemas concurrently you are having to store than object in memory 1000s of times with that solution. This does not scale and is the reason we went down the search_path test which allows you to use one copy of the model across all schemas.

janmeier:
I don't believe that would be an issue. Each new schema'd model only contains the name of the schema, and a reference to its prototype

mliszewski:
OK I missed the part about using Object.create to 'clone'. This sounds like it would solve work assuming the created objects were GC'ed properly. If they were not you would still have a slow but steady memory leak.

mickhansen:
@mliszewski i'm still open to introducing search_path but we'd need a way to make sure it's opt-in (for adding 'DEFAULT' aswell).

mliszewski:
@mickhansen Sorry I didn't call this out better. I made the functionality opt-in in this versions by only prepending if the dialectOption 'prependSearchPath' is set to true. See https://github.com/mliszewski/sequelize/blob/wrap_search_path/lib/sequelize.js#L744

mickhansen:
Oh jesus, i missed that @mliszewski, sorry about that.
Ideally we should make schemas work better and more performant, but for people who would infact have thousands of schemas additional stuff like this would be great.

mliszewski:
@mickhansen Can you rerun the tests please. They seem to have failed due to an issue with nvm.

mliszewski:
@mickhansen Can you please rerun the tests (failed due to nvm issues) and let me know if this looks good to get merge if tests pass?

mickhansen:
Merge conflict now, you will need to rebase (also an ideal time to squash your commits to a single descriptive commit)

mliszewski:
@mickhansen Rebased, squashed commits and test passed

PR #4540, opened 2015-09-18
Expose Association constructor as `Sequelize.Association`

overlookmotel:
This PR exposes the `Association` constructor as `Sequelize.Association`. The different association constructors are also exposed as `Sequelize.Association.BelongsTo`, `Sequelize.Association.HasMany` etc.

This is useful for things like:

```js
if (obj instanceof Sequelize.Association) { /* do something */ }
if (association instanceof Sequelize.Association.HasMany) { /* do something */ }
```


overlookmotel:
Thanks @mickhansen!

PR #4541, opened 2015-09-18
Fix supportShim

overlookmotel:
Attention of @mickhansen. This PR fixes the problems with `supportShim` which you ran into on https://github.com/sequelize/sequelize/pull/4177

The problem was that it wasn't shimming the new `HasMany` methods.

This PR relies on https://github.com/sequelize/sequelize/pull/4540 being merged first.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # Next
 - [FIXED] Calling set with dot.separated key on a JSON/JSONB attribute will not flag the entire object as changed [#4379](https://github.com/sequelize/sequelize/pull/4379)
 - [ADDED] Expose Association constructor as `Sequelize.Association`
+- [FIXED] Tests shim all methods to test for correct passing of `options.logging`
 
 # 3.9.0
 - [ADDED] beforeRestore/afterRestore hooks [#4371](https://github.com/sequelize/sequelize/issues/4371)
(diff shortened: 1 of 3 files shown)


overlookmotel:
Failing tests relate to checks that `options` hasn't been altered. e.g.:

```js
// test/unit/instance/decrement.test.js line 44
instance = Model.build({id: 3}, {isNewRecord: false});
var options = { by: 2 };
instance.decrement(['id'], options);
expect(options).to.deep.equal({ by: 2 });
```

`options` has changed because the shim added `options.logging`. The way the shim works is roughly:

```js
options.logging = loggingFunction;
sequelizeMethod().then(function(result) {
    delete options.logging;
    return result;
});
```

So in the above case the test fails because it doesn't wait for the promise returned by `instance.decrement()` to resolve before testing that the `options` object hasn't changed.

Hmm...

The test wouldn't fail if it was written like this:

```js
instance = Model.build({id: 3}, {isNewRecord: false});
var options = { by: 2 };
instance.decrement(['id'], options).then(function() {
    expect(options).to.deep.equal({ by: 2 });
});
```

But... maybe it should fail!

Here's my question: is it a general rule that no Sequelize method should alter the options object passed to it? If so, rather than writing individual tests for different methods, I could include this check in the test shim so it tests every method for this.

mickhansen:
Ah, as simple as including the association prototype methods, neat :)

Yes, we are trying, as a general rule to make sure that options objects can be reused on the side of the user (i.e. we clone as soon as possible).

PR #4562, opened 2015-09-23
Fixes transaction not getting cleaned up with sqlite Closes #4245

### LINKED ISSUE #4245: transaction does not clean up with sqlite
IDK if this affects other dialects but i am using sqlite3. 

Transactions are not getting removed from the connectionManager because the connection uuid is getting set to undefined before the connection is released. This might lead to long term resource  leaking.

The [old code found here](https://github.com/sequelize/sequelize/blob/f678009d7514b81a6f87e12b86360e9a597e3ca8/lib/transaction.js#L216-L225):

    Transaction.prototype.cleanup = function() {
      var cls = this.sequelize.constructor.cls;

      this.connection.uuid = undefined;

      if (cls) {
        if (cls.get('transaction') === this) {
         cls.set('transaction', null);
        }
      }

What i did to fix it (currently used as an override in my code, no PR, sorry):
    
    Transaction.prototype.cleanup = function() {
		var cls = this.sequelize.constructor.cls;
		if(cls && cls.get('transaction') === this) cls.set('transaction', null);
		
		return this.sequelize.connectionManager.releaseConnection(this.connection);
	};

codyrigney92:
This addresses https://github.com/sequelize/sequelize/issues/4245.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 - [FIXED] instances returned from `bulkCreate` now has `isNewRecord: false` and should be updateable if using `returning: true` with dialects that support it.
 - [FIXED] Find with Include with a where clause generates wrong SQL [#3940](https://github.com/sequelize/sequelize/issues/3940)
 - [FIXED] ON DELETE constraint should default to CASCADE if foreignKey has allowNull: false] [#2831](https://github.com/sequelize/sequelize/issues/2831)
+- [FIXED] sqlite file handle leak
 
 # 3.9.0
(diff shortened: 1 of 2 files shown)


codyrigney92:
I don't understand what would have caused the build to fail.  Can it be rerun?

mickhansen:
Unrelated to this PR.
LGTM, although this does expose some inconsistencies in connection releasing.

/cc @janmeier

janmeier:
LGTM too - thanks @codyrigney92 

I can't figure out a better way to fix the inconsistensies right now ...

codyrigney92:
It looks like the 3.10.0 release came out right before this was merged, so the changelog item for this is in the wrong spot.  Should that be fixed?

PR #4586, opened 2015-09-29
Oracle support for sequelize

gintsgints:
Please review.


lib/dialects/oracle/query.js
```diff
@@ -0,0 +1,142 @@
+'use strict';
+
+var Utils = require('../../utils')
+  , AbstractQuery = require('../abstract/query')
+  , _ = require('lodash')
+  , autocommit = true;
```
janmeier: Having a global option is not going to work - users might want some transactions with autocommit and some without

Couldn't you set `self.connection.autocommit` instead?

test/integration/associations/belongs-to-many.test.js
```diff
@@ -538,8 +544,8 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
     });
 
     it('should be able to set twice with custom primary keys', function() {
-      var User = this.sequelize.define('User', { uid: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true }, username: DataTypes.STRING })
-        , Task = this.sequelize.define('Task', { tid: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true }, title: DataTypes.STRING });
+      var User = this.sequelize.define('User', { myuid: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true }, username: DataTypes.STRING })
+        , Task = this.sequelize.define('Task', { mytid: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true }, title: DataTypes.STRING });
```
janmeier: Why the `my`?

lib/dialects/abstract/query-generator.js
```diff
@@ -1347,7 +1365,11 @@ var QueryGenerator = {
     if (subQuery) {
       subQueryItems.push('SELECT ' + subQueryAttributes.join(', ') + ' FROM ' + table);
       if (mainTableAs) {
-        subQueryItems.push(' AS ' + mainTableAs);
+        if (this._dialect.supports.as) {
```
janmeier: not very DRY :) 

Perhaps we could abstract it into a function , something like

```js
as: function (str) {
 return (this._dialect.supports.as ? ' AS ' : ' ') + str;
}

subQueryItems.push(this.as(mainTableAs));
```

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,125 @@
+'use strict';
+
+var BaseTypes = require('../../data-types')
+  , util = require('util')
+  , _ = require('lodash');
+
+BaseTypes.ABSTRACT.prototype.dialectTypes = 'https://msdn.microsoft.com/en-us/library/ms187752%28v=sql.110%29.aspx';
```
janmeier: Should be updated to point to oracle datatypes docs

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,125 @@
...
+};
+
+var DATE = function() {
+  if (!(this instanceof DATE)) return new DATE();
+  BaseTypes.DATE.apply(this, arguments);
+};
+util.inherits(DATE, BaseTypes.DATE);
+
+DATE.prototype.toSql = function() {
+  return 'TIMESTAMP WITH LOCAL TIME ZONE';
```
janmeier: Hmm, this is a bit different from how other dialects do it - other dialects default to UTC, except for postgres, which stores whatever we put into it

If the local timezone is different from UTC but sequelize tries to input an UTC date we might run into some problems

lib/dialects/oracle/query-generator.js
```diff
@@ -0,0 +1,611 @@
...
+    } else if (options.limit) {
+      if (options.offset) {
+      } else {
+        options.where.ROWNUM = {$lte: options.limit};
+      }
+    }
+    return options;
+  },
+
+  addLimitAndOffset: function(options){
```
janmeier: Can't this function just return `''`? As I understand it this will never work in oracle?

lib/dialects/oracle/query-generator.js
```diff
@@ -0,0 +1,611 @@
...
+  },
+
+  booleanValue: function(value){
+    return !!value ? 1 : 0;
+  },
+
+  addWhereLimitAndOffset: function(options){
+    if (options.offset && !options.limit) {
+    } else if (options.limit) {
+      if (options.offset) {
```
janmeier: No support for offset at all?

lib/dialects/oracle/query.js
```diff
@@ -0,0 +1,142 @@
...
+  var self = this;
+  this.sql = sql;
+
+  this.sequelize.log('Executing (' + (this.connection.uuid || 'default') + '): ' + this.sql, this.options);
+
+  var promise = new Utils.Promise(function(resolve, reject) {
+    // TRANSACTION SUPPORT
+    if (_.contains(self.sql, 'ROLLBACK;')) {
+      console.log('should roll back');
+      self.connection.rollback(function(err) {
```
janmeier: Can we use nodeify instead? `resolve(Promise.nodeify(self.connection.rollback)())`

If we can, we might want to save a nodeifiied copy of the function, instead of nodeifying each time

lib/sql-string.js
```diff
@@ -98,6 +107,11 @@ SqlString.escape = function(val, stringifyObjects, timeZone, dialect, field) {
       }
     });
   }
+  if (isDate) {
```
janmeier: I still think this looks ugly.

lib/sql-string.js
```diff
@@ -62,8 +69,10 @@ SqlString.escape = function(val, stringifyObjects, timeZone, dialect, field) {
     return val + '';
   }
 
+  var isDate = false;
   if (val instanceof Date) {
     val = SqlString.dateToString(val, timeZone || 'Z', dialect);
+    isDate = true;
```
janmeier: Having
```js
val = "'" + val "'";
if (dialect === 'oracle') {
      return "TIMESTAMP'" + val;
} else {
  return val;
}
```
Would be better in my opinion

test/integration/associations/belongs-to-many.test.js
```diff
@@ -1299,7 +1305,7 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
         }
       });
       this.User.belongsToMany(this.Group, {
-        through: 'group_users',
+        through: 'gu',
```
janmeier: I'm surprised if you only had to change this few table names to get around the long identifier problem.

However, I'd still like to have to solved properly instead of merging a hack, but unfortunately, noone is working on it yet

janmeier:
Looking good so far @gintsgints  - How few tests are still not passing :)?

I've set up the username and password to oracle on travis so you should be able to follow the instructions in https://github.com/cbandy/travis-oracle to get oracle express up and running

You should wrap download and install in a small script though, so they are only run for oracle jobs

lib/dialects/oracle/data-types.js
gintsgints: This is short side of oracle driver. They can only do this TIMESTAMP.

janmeier: What happens if you try to insert `2015-10-03 13:38 +00:00` - will it be converted to `2015-10-03 15:38 +02:00` in the db (assuming my local db runs CEST). That could still work I guess, as long as we get the timezone information about when we query, so its converted to a Date object correctly

lib/dialects/oracle/query.js
gintsgints: This again do not work for oracle driver. Oracle driver not yet support different autocommit options for different sessions.
I have failed feature request about that on driver.
https://github.com/oracle/node-oracledb/issues/159

test/integration/associations/belongs-to-many.test.js
gintsgints: If I remember correctly it made reserverd oracle term.

lib/dialects/abstract/query-generator.js
gintsgints: I made oneliner now
```
subQueryItems.push((this._dialect.supports.as ? ' AS ' : ' ') + mainTableAs);
```

lib/dialects/oracle/data-types.js
gintsgints: May be we have to make test for that? :)

test/integration/associations/belongs-to-many.test.js
janmeier: Wut :-D. Isn't the column name escaped? In that case I'd think you should be able to use whatever name you want..

lib/dialects/oracle/query-generator.js
gintsgints: I would say this is work in progress. It could work partly.

lib/dialects/oracle/query-generator.js
gintsgints: For now I would like to pass tests. This feature I do next.

test/integration/associations/belongs-to-many.test.js
gintsgints: Hmm I revert change and do not see that message. I put it back for now.

Creadigme:
Hello, are you still planning to finish this pull request ? Thanks you.

janmeier:
@gintsgints (and others) Is this completely dead?

lukealbao:
@gintsgints @janmeier (et. al) a simple bump here – Is this in fact dead? I have a request to support it, and as perhaps might be expected, the people who demand Oracle support are also demanding it within a very short timeframe. I'd love to not have to branch our projects and ditch Sequelize for this one client. Thanks, all!

lib/dialects/oracle/query.js
felixfbecker: imo manual wrapping is always better than using dynamic wrapper functions

janmeier:
@lukealbao I believe this is dead - No-one from the core team has experience or time to implement oracle, so this would have to be a community effort. 

I have no problem helping out with the ops side of things, making sure we have an oracle test instance etc., but someone else needs to write the code and pledge to maintain it - Which admittedly requires a large effort.

I'm going to close this, not because we don't want to support it, but because it seems dead. If anyone wants to pick it up, feel free to open a new PR

PR #4589, opened 2015-09-29
fix: Created documentation the same way we have for sequelize command.

williamgueiros:
if using 'sequelize migration:create'

Don't have (migration, DataTypes) parameters , instead of that received (queryInterface, Sequelize) parameters.


mickhansen:
Can you confirm @sdepold?

sdepold:
:eyes:

sdepold:
image

PR #4593, opened 2015-09-30
Doc association api improvements. Fixes #4591

### LINKED ISSUE #4591: API -> Associations -> Overview seems to be old.
Hello again guys. I'm working on the typescript type defs for association mixins. And for that reason I need precise api documentations i.e. I'll be bothering you with a lot these :sweat_smile: 

The overview section of the association api documentation (phew) seems to be a bit old. From what I understand from [this](http://docs.sequelizejs.com/en/latest/docs/associations/#belongs-to-many-associations) `belongsToMany` *requires* `through` it's not optional any more. And that `hasMany` cannot be used for n:m associations (right?!). 

Give me the go ahead to fix this if I'm right :sunglasses:

subjectix:
~~Don't merge this just yet. i wanna know if all is ok, then fix the Overview section. So wait until then before merging :D~~

Addressing #4591

### DIFF
--- a/lib/associations/mixin.js
+++ b/lib/associations/mixin.js
@@ -8,11 +8,12 @@ var Utils = require('./../utils')
   , BelongsTo = require('./belongs-to');
 
 /**
- * Creating assocations in sequelize is done by calling one of the belongsTo / hasOne / hasMany functions on a model (the source), and providing another model as the first argument to the function (the target).
+ * Creating assocations in sequelize is done by calling one of the belongsTo / hasOne / hasMany / belongsToMany functions on a model (the source), and providing another model as the first argument to the function (the target).
  *
- * * hasOne - adds a foreign key to target
- * * belongsTo - add a foreign key to source
- * * hasMany - adds a foreign key to target, unless you also specify that target hasMany source, in which case a junction table is created with sourceId and targetId
+ * * hasOne - adds a foreign key to the target and singular association mixins to the source.
+ * * belongsTo - add a foreign key and singular association mixins to the source.
+ * * hasMany - adds a foreign key to target and plural association mixins to the source.
+ * * belongsToMany - creates an N:M association with a join table and adds plural association mixins to the source. The junction table is created with sourceId and targetId.
  *
  * Creating an association will add a foreign key constraint to the attributes. All associations use `CASCADE` on update and `SET NULL` on delete, except for n:m, which also uses `CASCADE` on delete.
  *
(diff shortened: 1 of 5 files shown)


subjectix:
There is only one problem here. I'm not sure how the recent version of sequelize handles join tables attributes:

If you want to set several target instances, but with different attributes you have to set the attributes on the instance, using a property with the name of the through model:

>```js
>p1.UserProjects = {
 > started: true
>}
>user.setProjects([p1, p2], {started: false}) // The default value is false, but p1 overrides that.
>```
>
>Similarly, when fetching through a join table with custom attributes, these attributes will be available as an object with the name of the through model.
>
>```js
>user.getProjects().then(function (projects) {
>  var p1 = projects[0]
>  p1.UserProjects.started // Is this project started yet?
>})
>```

Is this how it's handled? Are the names correct?

janmeier:
Depends on the casing of the join table model if believe. So if the models are called capital `User` and `Project`, it will be `UserProjects` - but if underscored is true it will be `user_projects`. At least thats how I remmeber it :)

subjectix:
@janmeier :fearful: Well it doesn't matter :grin:. I'll probably find out how it really works and update the api accordingly. For now if it's ok please merge this tnx ^^

docs/api/associations/index.md
```diff
@@ -1,11 +1,12 @@
 <a name="associations"></a>
```
janmeier: Please don't update in docs/api - its autogenerated.

All your changes should be reflected in the code, otherwise they will disappear next time the docs are generated

mickhansen: They are further down.

janmeier: But don't reflect the changes here ;)

mickhansen: Oh you're right, my bad :p

subjectix:
:scream: :scream: :scream: :scream: Sorry it was a stupid move. I'll fix it.

subjectix:
Is this alright?

lib/associations/mixin.js
```diff
@@ -148,54 +149,15 @@ Mixin.hasOne = singleLinked(HasOne);
...
- * If you use a through model with custom attributes, these attributes can be set when adding / setting new associations in two ways. Consider users and projects from before with a join table that stores whether the project has been started yet:
- * ```js
- * var UserProjects = sequelize.define('userprojects', {
- *   started: Sequelize.BOOLEAN
- * })
- * User.hasMany(Project, { through: UserProjects })
- * Project.hasMany(User, { through: UserProjects })
- * ```
- * ```js
- * jan.addProject(homework, { started: false }) // The homework project is not started yet
```
janmeier: Why remove this? It might not be the best fit at this exact point, but I still think its important to have some amount of examples in the api docs

lib/associations/mixin.js
```diff
@@ -8,11 +8,12 @@ var Utils = require('./../utils')
...
- * Creating assocations in sequelize is done by calling one of the belongsTo / hasOne / hasMany functions on a model (the source), and providing another model as the first argument to the function (the target).
+ * Creating assocations in sequelize is done by calling one of the belongsTo / hasOne / hasMany / belongsToMany functions on a model (the source), and providing another model as the first argument to the function (the target).
  *
- * * hasOne - adds a foreign key to target
- * * belongsTo - add a foreign key to source
- * * hasMany - adds a foreign key to target, unless you also specify that target hasMany source, in which case a junction table is created with sourceId and targetId
+ * * hasOne - adds a foreign key to the target and singular association mixins to the source.
+ * * belongsTo - add a foreign key and singular association mixins to the source.
+ * * hasMany - adds a foreign key to target and plural association mixins to the source.
+ * * belongsToMany - creates an N:M association with a join table and adds plural association mixins to both the source and the target. The junction table is created with sourceId and targetId.
```
janmeier: A single association still only adds the mixins to the source I believe - You can use a single if you only need to access it from one side, but for both sides you need two

lib/associations/mixin.js
subjectix: Should I at least mark these as deprecated?

lib/associations/mixin.js
subjectix: I see, I'll fix this.

lib/associations/mixin.js
janmeier: Ah yeah thats true - I just meant that the examples on how to work with the through table still has merits, it should just be moved to belongstoMany instead

subjectix: `belongstoMany` does have enough examples :grin:

janmeier: True :D

janmeier:
@subjectix A rebase + squash and this should be good to merge

subjectix:
@janmeier Done

PR #4594, opened 2015-09-30
Update dependencies

BridgeAR:
Bluebird has no BC changes and they are very careful not to break anything, so I guess it's fine to get minor updates instead of just patches.
Wellknown got a new feature without BC.
Dottie has no changelog but if I read the commit messages correct version 1.0 is backwards compatible.

Coffee-script is not used in sequelize.


changelog.md
```diff
@@ -1,3 +1,12 @@
+# 3.11.0
```
mickhansen: Does changing the dependencies warrant a minor bump?
Are we as a package providing any new features by upgrading?

/cc @janmeier

BridgeAR: If you use shrinkwrap the peer dependencies have to be ok. Otherwise you can't build your shrinkwrap file. That's why I consider it as a minor bump.

mickhansen: shrinkwrap doesn't work currently with sequelize? If not i'd say this is a bug fix then, not a new feature.

BridgeAR: It does not work with sequelize if you want to use bluebird 2.10.x.

mickhansen: Because of our `~`?
Probably best to just use `Next` untill we actually cut a release.

BridgeAR: Yepp, because of the tilde. I'll rebase it in a few minutes.

BridgeAR: Done

PR #4599, opened 2015-10-01
make schema a property on the model independent of prototype

mliszewski:
This PR is to address https://github.com/sequelize/sequelize/issues/4119.

I added some unit and integration tests to cover the behavior more thoroughly.

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -202,6 +202,7 @@ var BelongsToMany = function(source, target, options) {
      * @param {Object} [options]
      * @param {Object} [options.where] An optional where clause to limit the associated models
      * @param {String|Boolean} [options.scope] Apply a scope on the related model, or remove its default scope by passing false
+     * @param {String} [options.schema] Apply a schema on the related model
      * @return {Promise<Array<Instance>>}
      * @method getAssociations
      */
@@ -464,6 +465,10 @@ BelongsToMany.prototype.injectGetter = function(obj) {
       } else {
         model = model.scope(options.scope);
       }
+    }
+
+    if (options.hasOwnProperty('schema')) {
+      model = model.schema(options.schema, options.schemaDelimiter);
     }
 
     return model.findAll(options);
--- a/lib/associations/belongs-to.js
+++ b/lib/associations/belongs-to.js
@@ -76,6 +76,7 @@ var BelongsTo = function(source, target, options) {
      *
      * @param {Object} [options]
      * @param {String|Boolean} [options.scope] Apply a scope on the related model, or remove its default scope by passing false.
+     * @param {String} [options.schema] Apply a schema on the related model
      * @return {Promise<Instance>}
      * @method getAssociation
      */
@@ -157,6 +158,10 @@ BelongsTo.prototype.injectGetter = function(instancePrototype) {
       } else {
         model = model.scope(options.scope);
       }
+    }
+
+    if (options.hasOwnProperty('schema')) {
+      model = model.schema(options.schema, options.schemaDelimiter);
     }
 
     return model.find(options);
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -20,18 +20,16 @@ var QueryGenerator = {
   options: {},
 
   addSchema: function(param) {
-    var self = this
-      , schema = (!param.options && param.schema) || (param.options && param.options.schema ? param.options.schema : undefined)
-      , schemaDelimiter = (param.options && param.options.schemaDelimiter ? param.options.schemaDelimiter : undefined);
+    var self = this;
 
-    if (!schema) return param.tableName || param;
+    if (!param.$schema) return param.tableName || param;
 
     return {
       tableName: param.tableName || param,
       table: param.tableName || param,
       name: param.name || param,
-      schema: schema,
-      delimiter: schemaDelimiter || '.',
+      schema: param.$schema,
+      delimiter: param.$schemaDelimiter || '.',
       toString: function() {
         return self.quoteTable(this);
       }
@@ -69,10 +67,8 @@ var QueryGenerator = {
     var table = this.quoteTable(
       this.addSchema({
         tableName: tableName,
-        options: {
-          schema: schema,
-          schemaDelimiter: schemaDelimiter
-        }
+        $schema: schema,
+        $schemaDelimiter: schemaDelimiter
       })
     );
 
@@ -1588,7 +1584,7 @@ var QueryGenerator = {
                    left.rawAttributes[left.primaryKeyAttribute].field
 
       /* Attributes for the right side */
-      , right = association.target
+      , right = include.model
       , asRight = include.as
       , tableRight = right.getTableName()
       , fieldRight = association instanceof BelongsTo ?
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -641,14 +641,14 @@ Instance.prototype.save = function(options) {
 
       if (self.isNewRecord) {
         query = 'insert';
-        args = [self, self.Model.getTableName(options), values, options];
+        args = [self, self.$Model.getTableName(options), values, options];
       } else {
         var where = this.where();
 
         where = Utils.mapValueFieldNames(where, Object.keys(where), this.Model);
 
         query = 'update';
-        args = [self, self.Model.getTableName(options), values, where, options];
+        args = [self, self.$Model.getTableName(options), values, where, options];
       }
 
       return self.sequelize.getQueryInterface()[query].apply(self.sequelize.getQueryInterface(), args)
@@ -832,9 +832,9 @@ Instance.prototype.destroy = function(options) {
 
       this.setDataValue(field, values[field]);
 
-      return this.sequelize.getQueryInterface().update(this, this.Model.getTableName(options), values, where, _.defaults({ hooks: false }, options));
+      return this.sequelize.getQueryInterface().update(this, this.$Model.getTableName(options), values, where, _.defaults({ hooks: false }, options));
     } else {
-      return this.sequelize.getQueryInterface().delete(this, this.Model.getTableName(options), where, _.assign({ type: QueryTypes.DELETE, limit: null }, options));
+      return this.sequelize.getQueryInterface().delete(this, this.$Model.getTableName(options), where, _.assign({ type: QueryTypes.DELETE, limit: null }, options));
     }
   }).tap(function() {
     // Run after hook
@@ -943,7 +943,7 @@ Instance.prototype.increment = function(fields, options) {
     }
   }, this);
 
-  return this.sequelize.getQueryInterface().increment(this, this.Model.getTableName(options), values, where, options).return(this);
+  return this.sequelize.getQueryInterface().increment(this, this.$Model.getTableName(options), values, where, options).return(this);
 };
 
 /**
--- a/lib/model.js
+++ b/lib/model.js
@@ -60,6 +60,9 @@ var Model = function(name, attributes, options) {
     this.tableName = this.options.tableName;
   }
 
+  this.$schema = this.options.schema;
+  this.$schemaDelimiter = this.options.schemaDelimiter;
+
   // error check options
   Utils._.each(options.validate, function(validator, validatorType) {
     if (Utils._.contains(Utils._.keys(attributes), validatorType)) {
@@ -271,8 +274,6 @@ function conformInclude(include, self) {
 
     include = { model: model, association: include, as: include.as };
   } else if (include instanceof Model) {
-    model = include;
-
     include = { model: include };
   } else if (_.isPlainObject(include)) {
     if (include.association) {
@@ -1033,19 +1034,28 @@ Model.prototype.dropSchema = function(schema) {
  * @return {this}
  */
 Model.prototype.schema = function(schema, options) { // testhint options:none
-  this.options.schema = schema;
+  var self = this;
+  var clone = Object.create(self);
+
+  clone.$schema = schema;
 
   if (!!options) {
     if (typeof options === 'string') {
-      this.options.schemaDelimiter = options;
+      clone.$schemaDelimiter = options;
     } else {
       if (!!options.schemaDelimiter) {
-        this.options.schemaDelimiter = options.schemaDelimiter;
+        clone.$schemaDelimiter = options.schemaDelimiter;
       }
     }
   }
 
-  return this;
+  clone.Instance = function() {
+    self.Instance.apply(this, arguments);
+  };
+  clone.Instance.prototype = Object.create(self.Instance.prototype);
+  clone.Instance.prototype.$Model = clone;
+
+  return clone;
 };
 
 /**
@@ -1711,7 +1721,9 @@ Model.prototype.build = function(values, options) { // testhint options:none
     return this.bulkBuild(values, options);
   }
   options = _.extend({
-    isNewRecord: true
+    isNewRecord: true,
+    $schema: this.$schema,
+    $schemaDelimiter: this.$schemaDelimiter
   }, options || {});
 
   if (options.attributes) {
@@ -2557,7 +2569,7 @@ Model.prototype.update = function(values, options) {
  * @return {Promise}
  */
 Model.prototype.describe = function(schema, options) {
-  return this.QueryInterface.describeTable(this.tableName, _.assign({schema: schema || this.options.schema || undefined}, options));
+  return this.QueryInterface.describeTable(this.tableName, _.assign({schema: schema || this.$schema || undefined}, options));
 };
 
 Model.prototype.$getDefaultTimestamp = function(attr) {
--- a/test/integration/associations/belongs-to-many.test.js
+++ b/test/integration/associations/belongs-to-many.test.js
@@ -206,7 +206,12 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
       return self.sequelize.dropAllSchemas().then(function() {
         return self.sequelize.createSchema('acme');
       }).then(function() {
-        return self.sequelize.sync({force: true});
+        return Promise.all([
+          AcmeUser.sync({force: true}),
+          AcmeProject.sync({force: true})
+        ]);
+      }).then(function() {
+        return AcmeProjectUsers.sync({force: true});
       }).bind({}).then(function() {
         return AcmeUser.create();
       }).then(function(u) {
--- a/test/integration/model/schema.test.js
+++ b/test/integration/model/schema.test.js
@@ -1,0 +1,394 @@
+'use strict';
+
+/* jshint -W030 */
+/* jshint -W110 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../support')
+  , DataTypes = require(__dirname + '/../../../lib/data-types')
+  , current = Support.sequelize
+  , Promise   = current.Promise;
+var SCHEMA_ONE = 'schema_one';
+var SCHEMA_TWO = 'schema_two';
+
+var locationId;
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+  if (current.dialect.supports.schemas) {
+
+    describe('schemas', function() {
+      before(function() {
+        this.Restaurant = current.define('restaurant', {
+            foo: DataTypes.STRING,
+            bar: DataTypes.STRING
+          },
+          {tableName: "restaurants"});
+        this.Location = current.define('location', {
+            name: DataTypes.STRING
+          },
+          {tableName: "locations"});
+        this.Employee = current.define('employee', {
+            first_name: DataTypes.STRING,
+            last_name: DataTypes.STRING
+          },
+          {tableName: "employees"});
+        this.EmployeeOne = this.Employee.schema(SCHEMA_ONE);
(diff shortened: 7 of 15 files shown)


test/integration/associations/belongs-to-many.test.js
```diff
@@ -206,7 +206,11 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
       return self.sequelize.dropAllSchemas().then(function() {
         return self.sequelize.createSchema('acme');
       }).then(function() {
-        return self.sequelize.sync({force: true});
+        return AcmeUser.sync({force: true});
```
janmeier: Why the individual sync here? Sequelize should be able to figure out the right order

test/integration/model/schema.test.js
```diff
@@ -0,0 +1,458 @@
...
+        this.RestaurantTwo = this.Restaurant.schema(SCHEMA_TWO);
+      });
+
+
+      beforeEach('build restaurant tables', function() {
+        var self = this;
+        return current.createSchema('schema_one').then(function() {
+          return current.createSchema('schema_two');
+        }).then(function() {
+          return self.RestaurantOne.sync({force: true});
```
janmeier: return self.sequelize.sync()

And you don't need to `catch` - the test will fail if the chain is rejected

test/integration/model/schema.test.js
```diff
@@ -0,0 +1,458 @@
...
+              expect(obj.foo).to.equal('one');
+            });
+        });
+
+        it('should fail to insert data into schema_two using create', function() {
+          var self = this;
+
+          return self.RestaurantTwo.create({
+            foo: 'test'
+          }).catch(function(err) {
```
janmeier: No need to `catch`

test/integration/model/schema.test.js
```diff
@@ -0,0 +1,458 @@
...
+              expect(obj.foo).to.equal('one');
+              restaurantId = obj.id;
+              return self.RestaurantOne.findById(restaurantId);
+            }).then(function(obj) {
+              expect(obj).to.not.be.null;
+              expect(obj.foo).to.equal('one');
+            });
+        });
+
+        it('should fail to insert data into schema_two using create', function() {
```
janmeier: Why should it fail?

test/integration/model/schema.test.js
```diff
@@ -0,0 +1,458 @@
...
+
+          return self.RestaurantTwo.findOne({where: {foo: 'one'}}).then(function(RestaurantObj) {
+            expect(RestaurantObj).to.be.null;
+          });
+        });
+
+        it('should fail to find schema_two object in schema_one', function() {
+          var self = this;
+
+          return self.RestaurantOne.findOne({where: {foo: 'two'}}).then(function(RestaurantObj) {
```
janmeier: The tables are empty at this point, so what are we testing exactly :D ?

test/integration/model/schema.test.js
```diff
@@ -0,0 +1,458 @@
...
+          var self = this;
+
+          var restaurauntModel = self.RestaurantOne.build({bar: 'one.1'});
+
+          return restaurauntModel.save()
+            .then(function() {
+              restaurauntModel = self.RestaurantOne.build({bar: 'one.2'});
+              return restaurauntModel.save();
+            }).then(function() {
+              restaurauntModel = self.RestaurantTwo.build({bar: 'two.1'});
```
janmeier: `return self.RestaurantTwo.create({bar: 'two.1'});`

test/integration/model/schema.test.js
```diff
@@ -0,0 +1,458 @@
...
+              return self.RestaurantOne.findAll();
+            }).then(function(restaurantsOne) {
+              expect(restaurantsOne).to.not.be.null;
+              expect(restaurantsOne.length).to.equal(2);
+              restaurantsOne.forEach(function(restaurant) {
+                expect(restaurant.bar).to.contain('one');
+              });
+              return self.RestaurantOne.findAndCountAll();
+            }).then(function(restaurantsOne) {
+              expect(restaurantsOne).to.not.be.null;
```
janmeier: This block is saying the same thing as the previous, but with `findAndCount` right?

test/integration/model/schema.test.js
```diff
@@ -0,0 +1,458 @@
...
+      });
+
+      describe('Add data via instance.save, retrieve via model.count and model.find', function() {
+        it('should be able to insert data into both schemas using instance.save count it and retrieve it via findAll with where', function() {
+          var self = this;
+
+          var restaurauntModel = self.RestaurantOne.build({bar: 'one.1'});
+
+          return restaurauntModel.save().then(function() {
+            restaurauntModel = self.RestaurantTwo.build({bar: 'two.1'});
```
janmeier: You can merge the creation into one with
```js
return Promise.all([
  self.RestaurantTwo.bulkCreate([
    { bar ... },
   { bar .... }
  ]),
  self.RestaurantOne.bulkCreate([
    { bar ... },
   { bar .... }
  ]),
]);
```

test/integration/model/schema.test.js
```diff
@@ -0,0 +1,458 @@
...
+          }).then(function(obj) {
+            expect(obj).to.not.be.null;
+            expect(obj.foo).to.equal('one');
+            expect(obj.location).to.not.be.null;
+            expect(obj.location.name).to.equal('HQ');
+          });
+        });
+
+
+        it('should be able to insert and retrieve associated data into the table in schema_two', function() {
```
janmeier: This test doesn't add anything new in regards to schemas  - the result is the same regardless of which schema you are using

test/integration/associations/belongs-to-many.test.js
mliszewski: The models that are attached to the sequelize object to do not have the schema set anymore since .schema no longer sets the option for all models but now creates a copy of the model with the $schema attribute set. sequelize.define will only addd the base model with the default schema. An alternative would be to set the schema attribute on the define but then that wouldn't be using .schema at all.

lib/associations/belongs-to-many.js
```diff
@@ -464,6 +464,10 @@ BelongsToMany.prototype.injectGetter = function(obj) {
       }
     }
 
+    if(options.hasOwnProperty('schema')){
+      model = model.schema(options.schema,options.schemaDelimiter || '');
```
janmeier: No need to add the default `|| ''` I think - An empty value should be handled further down

lib/dialects/abstract/query-generator.js
```diff
@@ -21,8 +21,10 @@ var QueryGenerator = {
 
   addSchema: function(param) {
     var self = this
-      , schema = (!param.options && param.schema) || (param.options && param.options.schema ? param.options.schema : undefined)
-      , schemaDelimiter = (param.options && param.options.schemaDelimiter ? param.options.schemaDelimiter : undefined);
+      , schema = param.$schema ||
+        (param.options && param.options.schema ? param.options.schema : undefined)
```
janmeier: Is the `param.options` still needed? Shouldn't it always be `params.$schema`?

lib/dialects/abstract/query-generator.js
```diff
@@ -1536,7 +1538,7 @@ var QueryGenerator = {
                    left.rawAttributes[left.primaryKeyAttribute].field
 
       /* Attributes for the right side */
-      , right = association.target
+      , right = include.model
```
janmeier: ?

test/integration/model/schema.test.js
mliszewski: This is purposefully done in two steps because before this PR a race condition existed wherein if you called save on a built model after calling .schema with a different schema it would save into the new schema.

lib/instance.js
```diff
@@ -642,14 +643,14 @@ Instance.prototype.save = function(options) {
...
-        args = [self, self.Model.getTableName(options), values, options];
+        args = [self, self.Model.getTableName(this.$options), values, options];
       } else {
         var where = this.where();
 
         where = Utils.mapValueFieldNames(where, Object.keys(where), this.Model);
 
         query = 'update';
-        args = [self, self.Model.getTableName(options), values, where, options];
+        args = [self, self.Model.getTableName(this.$options), values, where, options];
```
janmeier: The options object is never used by `getTableName`

lib/model.js
```diff
@@ -271,8 +271,6 @@ function conformInclude(include, self) {
 
     include = { model: model, association: include, as: include.as };
   } else if (include instanceof Model) {
-    model = include;
```
janmeier: ?

test/integration/model/schema.test.js
mliszewski: Yes. Just making sure all operations behave correctly.

test/integration/model/schema.test.js
mliszewski: Cannot use self.sequelize.sync b/c of my comment above.

lib/model.js
```diff
@@ -1051,7 +1054,16 @@ Model.prototype.schema = function(schema, options) { // testhint options:none
...
-  return this.QueryGenerator.addSchema(this);
+  if(!options) {
+    return this.QueryGenerator.addSchema(this);
+  } else {
+    return this.QueryGenerator.addSchema({
+      tableName: this.tableName,
+      name: this.name,
+      $schema: options.$schema || this.$schema,
+      $schemaDelimiter: options.$schemaDelimiter || this.$schemaDelimiter
+    });
```
janmeier:
```js
return this.QueryGenerator.addSchema(_.assign({
      tableName: this.tableName,
      name: this.name,
      $schema: this.$schema,
      $schemaDelimiter: this.$schemaDelimiter
    }, options);
```
should cover both cases

test/integration/model/schema.test.js
mliszewski: I'm purposefully doing this in multiple steps to verify that built models maintain there proper $schema setting across calls as this was not the case with the original implementation because it was changing the schema option on the shared prototype.

test/integration/model/schema.test.js
mliszewski: True.

- [ ] Remove test

test/integration/associations/belongs-to-many.test.js
janmeier: True :)

janmeier: I'd prefer an `Promise.all([sync, sync, sync` though

test/integration/model/schema.test.js
janmeier: Yeah, I realized so too after writing the comment ;)

test/integration/model/schema.test.js
janmeier: Yeah I see your point - Might want to write a comment so the structure isn't removed in an overzealous refactor then ;)

test/unit/model/schema.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+/* jshint -W030 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support   = require(__dirname + '/../support')
+  , current   = Support.sequelize;
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+  if (current.dialect.supports.schemas) {
+    var Project = current.define('project'),
```
janmeier: +1 for unit tests - nice! 

Might want to add one for delimiter as well

lib/associations/belongs-to.js
```diff
@@ -159,6 +159,10 @@ BelongsTo.prototype.injectGetter = function(instancePrototype) {
       }
     }
 
+    if(options.hasOwnProperty('schema')){
```
janmeier: Remember to add API docs

janmeier:
Overall solid and well thought through work both in the test and code department @mliszewski !

test/integration/model/schema.test.js
mliszewski: It shouldn't. 

- [ ] Delete test

test/integration/model/schema.test.js
mliszewski: These two tests should be moved to the 2 tests above them to verify I cannot get the data in the opposite schema.

- [ ] Combine opposite schema verification into create data tests

lib/associations/belongs-to-many.js
mickhansen: Note on styling, we generally have a space between `if` and `(`.
We generally also have a space after a comma in a options set, so `options.schema, options.schemaDelimiter`:

This applies to the rest of the code aswell.

lib/instance.js
mickhansen: Actually it is, we let users override it to provide user specific table names based on options passed to a call.
So it should still be `options`, not `this.$options`

mickhansen: Same goes for the rest of the codebase where call options has been changed for model/instance options (which doesn't make a lot of sense to me actually, why would instance options be used?)

janmeier: Ah yes - Maybe `this` then - since the schema is not longer stored in options?

lib/model.js
```diff
@@ -978,9 +979,9 @@ Model.prototype.sync = function(options) {
       return self.drop(options);
     }
   }).then(function () {
-    return self.QueryInterface.createTable(self.getTableName(options), attributes, options, self);
+    return self.QueryInterface.createTable(self.getTableName(), attributes, options, self);
   }).then(function () {
-    return self.QueryInterface.showIndex(self.getTableName(options), options);
+    return self.QueryInterface.showIndex(self.getTableName(), options);
```
mickhansen: Restore `options`, see other comment.

lib/model.js
```diff
@@ -1027,19 +1028,21 @@ Model.prototype.dropSchema = function(schema) {
  * @return {this}
  */
 Model.prototype.schema = function(schema, options) { // testhint options:none
-  this.options.schema = schema;
+  var self = Object.create(this);
```
mickhansen: Perhaps use `clone` rather than `self` since `self` usually is a reference to the original `this`.

test/integration/model/schema.test.js
mickhansen: Wouldn't that case happen if the two saves happened in parallel? Which they don't here

lib/dialects/abstract/query-generator.js
mliszewski: This honors a models being passed on the include. association.target is the model at the time of the association being created. If you try to do 
```javascript
 return User.findAll({
    where: {
        first_name: 'matt'
    },
    include: [{
        model: Comments.schema('new_schema'),
        as: 'comments'
    }]
}); 
```
without this code it will join the included table with the schema set on base model. Without this .schema is pretty worthless for includes because you would have to add a separate association for every schema.

lib/model.js
mliszewski: This was unused.

lib/instance.js
mickhansen: Should be `options` as before, it was not about schema it was about providing custom tables by overloading getTableName

mliszewski: options was unused in getTableName. the code looked like this 
```javascript
Model.prototype.getTableName = function(options) { // testhint options:none
  return this.QueryGenerator.addSchema(this);
}
```
The issue here is that this.Model is attached to the prototype of the model and not the model instance itself. See https://github.com/sequelize/sequelize/blob/master/lib/model.js#L718-L719. Because of this the $schema and $schemaDelimiter attributes could be different than the values on the model that find was called on. An alternative here would be to do ``` this.Model.schema(this.$options.$schema,this.$options.$schemaDelimiter) ``` for the case where ```this.Model.$schema !== this.$options.$schema || this.Model.$schemaDelimiter !== this.$options.$schemaDelimiter ```. Does that work? If so I will revert back ALL of the options changes including making getTableName do nothing with options.

test/integration/model/schema.test.js
mliszewski: They don't have to be in parallel but out of order. In other words build.schema1, build.schema2, save.schema1 would fail. I do have a test for this explicitly later though so I'll clean these up.

lib/instance.js
mickhansen: `options` is used by end-users, as i wrote in my previous comment, changing it would mean a major bump and we're not looking to do that quite yet :)

`> The issue here is that this.Model is attached to the prototype of the model and not the model instance itself

Not sure what you mean? It's attached to the instance prototype which means its attached to ALL instances. (Not sure how it's relevant either).

mliszewski:
@mickhansen @janmeier I cleaned up some items based on your feedback. Below is a summary of my changes.

- Added API docs
- Added unit test for setting schema delimiter
- Removed/consolidated redundant test cases in schema test
- Reverted changes regarding ```options```

Let me know what you think when you get chance. Thanks!

lib/associations/belongs-to-many.js
```diff
@@ -464,6 +465,10 @@ BelongsToMany.prototype.injectGetter = function(obj) {
       }
     }
 
+    if (options.hasOwnProperty('schema')) {
+      model = model.schema(options.schema, options.schemaDelimiter);
+    }
```
mickhansen: Could the user not just pass `Model.schema()` to the relation setup call instead?

lib/instance.js
```diff
@@ -640,16 +641,24 @@ Instance.prototype.save = function(options) {
         self.dataValues[createdAtAttr] = values[self.Model.rawAttributes[createdAtAttr].field || createdAtAttr] = self.Model.$getDefaultTimestamp(createdAtAttr) || now;
       }
 
+      var tableName;
+      if (self.$options.$schema !== self.Model.$schema ||
+        self.$options.$schemaDelimiter !== self.Model.$schemaDelimiter) {
+        tableName = self.Model.schema(self.$options.$schema, self.$options.$schemaDelimiter).getTableName(options);
+      } else {
```
mickhansen: Why is this code needed exactly?
Seems odd there should be a case where there's mismatch in the instance model?

lib/associations/belongs-to-many.js
mliszewski: They could but then that would require the user to setup a separate association for every schema that they need to access this association in with different names. In an ideal world I can setup one set of models and associations and use them across all schemas without having to define the models and associations again with different names for each schema. That being said it would work as you described as well if that use case make more sense for the user than you would just not pass the schema option and it would use the schema associated with the model when the association was setup.

lib/instance.js
mliszewski: This is not odd at all if I am trying to sue the same model across schemas. The model is attached to the Instance prototype when init is called on the model: https://github.com/sequelize/sequelize/blob/master/lib/model.js#L724-L725. If I then call .schema on this model it will not change $schema on the model attached to the prototype.

mliszewski:
@mickhansen Let me know if my explanations make sense and if you would like to see anything else changed or explained. Thanks.

lib/instance.js
mickhansen: Hmm alright, i was thinking of doing a full clone on schema so the instance would be cloned aswell.

lib/associations/belongs-to-many.js
mickhansen: Makes sense for dynamic schemas.

lib/instance.js
mliszewski: Yeah I think the issue with trying to do this (at least from what I could tell) is that since Model/$Model are set on the prototype of the instance then changing the schema on one instance will change it on all which causes this test case to fail: https://github.com/sequelize/sequelize/pull/4599/files#diff-7de0096efac48f1ebb1ecedda103ca05R362.

mickhansen: You'd have to first clone the model, then clone Model.Instance and then set schema properties.

mliszewski: But if you clone Model.Instance using Object.create (which is ideal from a resource perspective) it will not matter b/c the Model property of instance is set on the prototype. I guess you could detach the Model from the prototype but this will cause all kinds of memory bloat. I guess you could also create attributes to specifically hold the schema properties on the Instance object and set them but that seems like a worse solution that my current solution that will read the $schema attribute from $options which sets them based on the model the actual instance was built from (https://github.com/sequelize/sequelize/pull/4599/files#diff-a140b60b6a99a8c69f4b5d0d1f1f2bfaL1706) which seems like the expected behavior.

mickhansen: Not sure i understand, couldn't we just:

```js
var clone = Object.create(Model);
clone.instance = function() {
   Model.Instance.apply(this, arguments);
};
clone.Instance.prototype = Object.clone(Model.Instance.prototype);
clone.Instance.prototype.$Model = clone;
```

`Model.schema` should perhaps then have a cache, LRU perhaps.
I would greatly prefer it to having 3-4 places with special handling code.

/cc @janmeier

janmeier: Yeah, making a clone of the instance makes more sense to me a well

mliszewski:
@mickhansen @janmeier I changed the code to clone instance as you recommended. Tests passed for me locally but for some reason they haven't picked up in CI yet. Can you run tests and give me any additional feedback when you get a chance? Thanks

mickhansen:
Hmm looks like it didn't pick up the rebase.

mickhansen:
This looks really great now @mliszewski - Not sure why it's not running the tests though, any ideas @janmeier?

janmeier:
@mickhansen opening and closing the PR fixed it ;).

This LGTM unless you have further comments?

Issue #3193, opened 2015-02-21
Support `Model.find({ ..., include: [ { as: 'fooBar' } ] })` and `Model.find({ ..., include: [ 'fooBar' ] })`


dantman:
`include` on model queries is fairly verbose in its current state.

``` js
Model.find({ ..., include: [ { model: FooModel, as: 'fooBar' } ] });
Model.find({ ..., include: [ { association: Model.associations.fooBar } ] });
```

All of my models use `as`, because I have things like `UserEmail` as `user.emails`/`user.getEmails`, `UserPreference` as `user.preferencePairs`/`user.getPreferencePairs`, `UserAvatar` as `user.avatar`/`user.getAvatar`, etc. So using include is very verbose for me.

From what I understand models already contain all the information they need about their associations. You can't associate multiple models to the same `as` name or property (`{ model: Bar, as: 'foo' }` and `{ model: Baz, as: 'foo' }` should be impossible on one model). And using an association from a different model has no meaning and just breaks the query.

So it doesn't sound like both `modal` and `as` or a whole `association` object is needed to do an include.

So I'd like to suggest we also support these formats for doing an include:

``` js
Model.find({ ..., include: [ { as: 'fooBar' } ] });
Model.find({ ..., include: [ 'fooBar' ] });
```

I think this actually makes a lot of sense.

Implementation wise `Model` already knows that an association to the `fooBar` property exists and that the model for it is `FooModel`.

User expectation wise. The `as: 'fooBar'` association maps to the `fooBar` property. And so when you're writing `Model.find({ ..., include: [ 'fooBar' ] })` and then using `model.fooBar` it reads perfectly as `Find a Model instance and include the value of the 'fooBar' property from the associated model`.

mickhansen:
We had this feature previously but it was very error prone.
I would be very open to supporting it, but only for relations that explicitely use "as" (since we respect "as" as-is and don't transform it).

irond13:
I also think this is a good idea. I've submitted a pull request containing both of the above feature rests. @mickhansen why would this be very error prone?

mickhansen:
@irond13 we used to have quite a big of magic around naming that confused a ton of users, this has mostly been removed now so it could be worth revisiting.

PR #4601, opened 2015-10-01
Added ability to omit model reference when specifying association inc…

irond13:
…ludes #3193

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -262,6 +262,12 @@ function conformInclude(include, self) {
 
   if (include._pseudo) return include;
 
+  if (_.isString(include)) {
+    var as = include;
+    include = _.find(self.associations, 'as', include);
+    if (! include) throw new Error('Include unexpected. Unable to find Association with \'as\': ' + as);
+  }
+
   if (include instanceof Association) {
     if (self && include.target.name === self.name) {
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -289,6 +295,14 @@ function conformInclude(include, self) {
         include.as = include.association.as;
       }
     } else {
+      if (! include.model && include.as) {
+        var association = _.find(self.associations, 'as', include.as);
+        if (! association) throw new Error('Include unexpected. Unable to find Association with \'as\': ' + include.as);
+
+        include.association = association;
+        return conformInclude(include, self);
+      }
```
mickhansen: This would be in the case of `[{as: 'String', where: {}}]`?

mickhansen:
Please add a test that fails when using a string on an association without `as` specified.

lib/model.js
irond13: Yes, that's correct (see test 'should allow me refer to an association using only its alias in object form in eager loading')

irond13:
Added two negative test cases

janmeier:
LGTM - I think this is the right approach - if it's in `model.associations`, use it, otherwise fail.

We should remember to document that! :)

PR #4604, opened 2015-10-03
feat(validations) Partial rollback of datatype validation by hiding b…

janmeier:
…ehind a flag. Soften up some validation and restrict them to insert and create

Adresses https://github.com/sequelize/sequelize/issues/4314

I've hidden the validations behind a flag `validations`, which is false by default - which menas they are practically rolled back unless the user manually enables them.

Furthermore, I've restricted the validations to only run on insert, bulkinsert and update, not findAll etc. I still believe there would be some merit to validation find query args, but there are also lots of complications, as the issues mentinoed in https://github.com/sequelize/sequelize/issues/4314 show - e.g $in array and $like str%, for uuid etc.

I also refactored bulkInsert into the abstract query generator, since the code was 90% the same for mysql, sqlite and pg

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+# Next
+- [FIXED] Partial rollback of datatype validations by hiding it behind the `validation` flag.
+
+
 # 3.11.0
 - [INTERNALS] Updated dependencies [#4594](https://github.com/sequelize/sequelize/pull/4594)
     + bluebird@2.10.1
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,5 @@
 # Next
-- [FIXED] Partial rollback of datatype validations by hiding it behind the `validation` flag.
-
+- [FIXED] Partial rollback of datatype validations by hiding it behind the `typeValidation` flag.
 
 # 3.11.0
 - [INTERNALS] Updated dependencies [#4594](https://github.com/sequelize/sequelize/pull/4594)
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -84,10 +84,8 @@ STRING.prototype.toSql = function() {
 };
 STRING.prototype.validate = function(value) {
   if (Object.prototype.toString.call(value) !== '[object String]') {
-    if (this.options.binary) {
-      if (Buffer.isBuffer(value)) {
-        return true;
-      }
+    if ((this.options.binary && Buffer.isBuffer(value)) || _.isNumber(value)) {
+      return true;
     }
     throw new sequelizeErrors.ValidationError(util.format('%j is not a valid string', value));
   }
@@ -352,7 +350,7 @@ DECIMAL.prototype.toSql = function() {
   return 'DECIMAL';
 };
 DECIMAL.prototype.validate = function(value) {
-  if (!_.isNumber(value)) {
+  if (!Validator.isDecimal(value)) {
     throw new sequelizeErrors.ValidationError(util.format('%j is not a valid decimal', value));
   }
 
@@ -374,7 +372,7 @@ BOOLEAN.prototype.toSql = function() {
   return 'TINYINT(1)';
 };
 BOOLEAN.prototype.validate = function(value) {
-  if (!_.isBoolean(value)) {
+  if (!Validator.isBoolean(value)) {
     throw new sequelizeErrors.ValidationError(util.format('%j is not a valid boolean', value));
   }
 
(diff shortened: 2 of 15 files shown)


lib/data-types.js
```diff
@@ -374,7 +372,7 @@ BOOLEAN.prototype.toSql = function() {
   return 'TINYINT(1)';
 };
 BOOLEAN.prototype.validate = function(value) {
-  if (!_.isBoolean(value)) {
+  if (!Validator.isBoolean(value)) {
```
janmeier: Validator is more loose about boolean and decimal like strings

test/unit/model/validation.test.js
```diff
@@ -266,10 +269,95 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), function() {
...
+          return expect(User.create({
+            awesome: '1'
+          })).not.to.be.rejected;
+        });
+
+        it('should allow true/false string as a boolean', function () {
+          return expect(User.create({
+            awesome: 'true'
+          })).not.to.be.rejected;
+        });
```
janmeier: Are there more cases we need to test here?

mickhansen: Personally i think having `'true'` is terrible.

lib/sequelize.js
```diff
@@ -143,7 +144,8 @@ var Sequelize = function(database, username, password, options) {
     quoteIdentifiers: true,
     hooks: {},
     isolationLevel: Transaction.ISOLATION_LEVELS.REPEATABLE_READ,
-    databaseVersion: 0
+    databaseVersion: 0,
+    validation: false
```
mickhansen: perhaps `typeValidation` and perhaps allow it on model level aswell?
`validation` is perhaps slightly generic.

test/unit/model/validation.test.js
janmeier: Agreed, but that was one of the complaints regarding the validations I believe

mickhansen: Can you link the relevant comment?
I believe that if we do have validations we should definitely not allow this type. If a protocol is sending boolean values in strings the frontend should convert them appropriately.

janmeier: Nah, I must have been thinking about an older issue

janmeier: What about 0/1 strings

test/unit/model/validation.test.js
```diff
@@ -266,10 +269,95 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), function() {
...
+      describe('create', function () {
+        it('should allow number as a string', function () {
+          return expect(User.create({
+            age: '12'
+          })).not.to.be.rejected;
+        });
+
+        it('should allow decimal as a string', function () {
+          return expect(User.create({
+            number: '12.6'
```
defunctzombie: scientific notation tests? (if you want to support that)

defunctzombie: also a test where the number is clearly outside the js decimal range to make sure no funky stuff happens

test/unit/model/validation.test.js
mickhansen: 0/1 integers are ok, not strings. My personal opinion obviously but its just slightly too magic for me.
Although i honestly don't know how PG handles these cases (we might want to adjust for that).

janmeier:
```js
return Promise.join(
    User.create({ bool: 1 }),
    User.create({ bool: 0 }),
    User.create({ bool: '1' }),
    User.create({ bool: '0' }),
    User.create({ bool: 'true' }),
    User.create({ bool: 'false' })
  )
```
Postgres - all good
Mysql - `'true'` interpreted as false
Sql - If you put strings in you get strings out
Mssql - all good

I removed this line while testing https://github.com/sequelize/sequelize/blob/d8affcf2ec7ccaea5caeb4af823158fa78dde060/lib/instance.js#L363 so the actual values were tested.

In practice we already have handling for passing bools as strings, because of the linked line in `instance.set` which converts them - that was probably what I was thinking of

test/unit/model/validation.test.js
janmeier: Added below

lib/sequelize.js
```diff
@@ -143,7 +144,8 @@ var Sequelize = function(database, username, password, options) {
     quoteIdentifiers: true,
     hooks: {},
     isolationLevel: Transaction.ISOLATION_LEVELS.REPEATABLE_READ,
-    databaseVersion: 0
+    databaseVersion: 0,
+    typeValidation: false
```
janmeier: Option changed to `typeValidation`

changelog.md
```diff
@@ -1,3 +1,7 @@
+# Next
+- [FIXED] Partial rollback of datatype validations by hiding it behind the `validation` flag.
```
mickhansen: Fix changelog then LGTM (go ahead and merge yourself when changed)

damien-git:
This could be better documented, it took me a long time to find this secret `typeValidation` flag.

papb:
@damien-git can you please open a new docs issue about this, and link to this PR in your issue description? Thanks!

papb:
> 
> 
> @damien-git can you please open a new docs issue about this, and link to this PR in your issue description? Thanks!

I did it myself now - #11828

PR #4616, opened 2015-10-07
Fix: if `errDetail` is undefined it will cause exception

rwu823:


janmeier:
In which cases is the error detail empty? It would be good to add a test as well

masato25:
I had meet this kind of problem.
when I set a table has a unique field and insert the table in async way.

I will got this error exceptions:

/home/masato/netbios_lookup/node_modules/sequelize/lib/dialects/postgres/query.js:384
      match = errDetail.replace(/"/g, '').match(/Key \((.*?)\)=\((.*?)\)/);
                       ^
TypeError: Cannot read property 'replace' of undefined
    at Query.formatError (/home/masato/netbios_lookup/node_modules/sequelize/lib/dialects/postgres/query.js:384:24)
    at null.<anonymous> (/home/masato/netbios_lookup/node_modules/sequelize/lib/dialects/postgres/query.js:112:19)
    at emitOne (events.js:77:13)
    at emit (events.js:169:7)
    at Query.handleError (/home/masato/netbios_lookup/node_modules/pg/lib/query.js:108:8)
    at null.<anonymous> (/home/masato/netbios_lookup/node_modules/pg/lib/client.js:171:26)
    at emitOne (events.js:77:13)
    at emit (events.js:169:7)
    at Socket.<anonymous> (/home/masato/netbios_lookup/node_modules/pg/lib/connection.js:109:12)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)...

I had tried catch the exception by some error handling way .. but the program can not catch this exception. the only way is add if condition checking for "errDetail". After that I cat catch the error message as normal.

Masato

janmeier:
Can you create a reproducible test case that shows the problem? I feel like the error should always have a detail, and if not thats a problem with node-pg mayhaps?

Dronnikkl:
# console.log(err) in FormatError(err):
 { [error: duplicate key value violates unique constraint "car_number_index"]
  name: 'error',
  length: 122,
  severity: 'ERROR',
  code: '23505',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'nbtinsert.c',
  line: '300',
  routine: '_bt_check_unique',
  sql: 'INSERT INTO "car" ("id","reg_number","account") VALUES (DEFAULT,\'О428ОО197\',1),(DEFAULT,\'R528ОО197\',1) RETURNING *;' }

and exception thrown:
/Users/Andrew/Documents/Sites/tahocarddriver/tahocarddriver.node/node_modules/sequelize/lib/dialects/postgres/query.js:391
      match = errDetail.replace(/"/g, '').match(/Key \((.*?)\)=\((.*?)\)/);

huguesdk:
I have the same problem. I’m using PostgreSQL servers with versions 8.2 and 8.3. The error that causes the error in my case is:

    { [error: duplicate key violates unique constraint "some_table_id_key"]
      name: 'error',
      length: 139,
      severity: 'ERROR',
      code: '23505',
      detail: undefined,
      hint: undefined,
      position: undefined,
      internalPosition: undefined,
      internalQuery: undefined,
      where: undefined,
      file: 'nbtinsert.c',
      line: '277',
      routine: '_bt_check_unique',
      sql: 'INSERT INTO "some_table" ("some_column_1","some_column_2") VALUES (42,12345) RETURNING *;' }

huguesdk:
When can we hope this to be merged? Our app crashes when a database error happens because of this bug.

mickhansen:
@huguesdk Original author never really followed up on any of the comments.

janmeier:
Closing, since noone has provided at test case

PR #4623, opened 2015-10-08
Added support for passing domain option to tedious.

PenguinOfThunder:
The current mssql connection manager does not allow for passing in the "domain" to the tedious driver. This patch allows it to pass through by setting it in the dialect options.


janmeier:
Thanks :)

Issue #4103, opened 2015-07-13
bulkCreate does not understand timestamps renamed fields


JSteunou:
Migration from 2.0.x to 3.x I have some issue with `bulkCreate`

> SequelizeDatabaseError: ER_BAD_FIELD_ERROR: Unknown column 'createdAt' in 'field list'

I have this

```javascript
{
        createdAt: {
            type: DataTypes.DATE,
            field: 'created_at'
        },
        updatedAt: {
            type: DataTypes.DATE,
            field: 'updated_at'
        },
        deletedAt: {
            type: DataTypes.DATE,
            field: 'deleted_at'
        }
}
```

and `timestamp` at `true` in my model definition

JSteunou:
This is specific to timestamp fields. I test with a `fooBar` property in my model, having `foo_bar` in my table, no issue when I add `field: 'foo_bar'` in my model.

JSteunou:
More tests... it's not only bulk operation, I have the issue on `Model#create` too.

rfink:
With the "updateOnDuplicate" clause added, this also fails to recognized aliased fields.

JSteunou:
Any news on this one?

JSteunou:
@janmeier @mickhansen This issue is still present in 3.11, how can I help you to help me on this?
Is there a way around until this is fixed? This bug still prevent me to update from 2.x to 3.x.

JSteunou:
`#create` dont have the issue anymore, so I could get around by doing an `Array#map` to call create multiple time. That does the trick for test case, but I would prefer rely on `bulkCreate` for production code, depending on what SQL optimisation it does underneath

janmeier:
@JSteunou If you can create a pull request with failing tests that would maybe speed up things a bit.

Shouldn't be too much code required to actually do a fix either ;)

JSteunou:
Ok I try to follow the doc and run test against mysql but got this

```
4 errors
Makefile:67: recipe for target 'jshint' failed
make[2]: *** [jshint] Error 2
make[2]: Leaving directory '/home/jerome/Git/sequelize'
Makefile:20: recipe for target 'test' failed
make[1]: *** [test] Error 2
make[1]: Leaving directory '/home/jerome/Git/sequelize'
Makefile:74: recipe for target 'mysql' failed
make: *** [mysql] Error 2
```

JSteunou:
ok just found the `make test-unit` command to bypass jshint

janmeier:
[...]

PR #4646, opened 2015-10-13
Add failing unit tests

JSteunou:
linked to #4103

### DIFF
--- a/test/unit/model/bulkcreate.test.js
+++ b/test/unit/model/bulkcreate.test.js
@@ -19,6 +19,26 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       }
     }, { timestamps: false });
 
+    var ModelWithTS = current.define('modelwithts', {
+      accountId: {
+        type: DataTypes.INTEGER(11).UNSIGNED,
+        allowNull: false,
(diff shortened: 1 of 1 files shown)


test/unit/model/bulkcreate.test.js
```diff
@@ -45,5 +65,19 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
       });
     });
+
+    describe('renamed fields', function () {
+      it('should not fail for timestamps', function () {
+        return ModelWithTS.bulkCreate([
+          { accountId: 42 }
+        ]).bind(this).then(function () {
+          var firstCreatedModel = this.stub.getCall(0).args[1][0];
+          expect(firstCreatedModel).to.have.property('id');
```
mickhansen: `id` would not ever be present in the bulkInsert call.

JSteunou: I add it because it's present (but null) so I though it was normal

mickhansen: Ah, that might be true, but it's not something that's relevant to this test :)

JSteunou:
Updated to remove check on `id`

PR #4661, opened 2015-10-14
Force the client to be reaped if an ECONNRESET is returned on query [postgres]

robraux:
## Here's the scenario being addressed

1. Consistent SQL requests are being made through Sequelize, utilizing the postgres dialect and a pool size of 5.
2. A database failover occurs, which resets all active connections to the database server.

## What's expected

The pool members will get invalidated, and are able to reconnect when the database is accessible. 

## Without this PR

1. The Query on error will return a parsed ECONNRESET error message.
2. The client from pool is never marked invalid, so it returned on the next request for a handle.
3. The next query is requested to run on the client from the pool, however it never returns from the request. At this stage, all clients from the pool are active, but will never return, completely locking the system down.

## With this PR

1. The Query will mark the client as invalid.
2. The next query requested asks for a client from the pool. generic-pool validate function is run, reaping this client.
3. A new client is created with the proper connection credentials and database requests resume.

## Why

We faced this exact scenario in a production environment this week when an automatic database failover occurred. This PR addresses the scenario, utilizing the existing generic-pool validation methods, coupled with the ````_invalid```` variable which was introduced [here](https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/connection-manager.js#L94) to address connection issues while the database server is unreachable. 

In the scenario outlined above, the ````connection.on('error')```` is never called, only the query received the event regarding the ECONNRESET. 

## Testing

This scenario is difficult to replicate in a unit testing environment. Beyond the unit tests passing, I could not think of a test suite to test this scenario.

Here is the manual testing which was performed against this PR:

1. Run API server, which utilizes postgres and sequelize.
2. Create a steady flood of requests to the server to ensure no idle reaping of connections occurred. (e.g. [script](https://gist.github.com/robraux/ffba98e810568a5559b6))
3. For a database restart, or failover. (e.g  ````aws rds reboot-db-instance --force-failover --db-instance-identifier psql-testing````)
4. Observe the flood script start returning 500's as ECONNRESETS happen, then errors when the database is still starting up. Finally, after approximately a minute in our scenario, the API is responding to requests with a 200.

### DIFF
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -107,6 +107,12 @@ Query.prototype.run = function(sql) {
     });
 
     query.on('error', function(err) {
+
+      // set the client so that it will be reaped if the connection resets while executing
+      if(err.code === 'ECONNRESET') {
+        self.client._invalid = true;
+      }
+
       receivedError = true;
       err.sql = sql;
       reject(self.formatError(err));


robraux:
Obviously I screwed up linting here. Will address.

janmeier:
Looks good and very thorough! One question before I merge - are there other errors we should handle here that could also make the client invalid?

robraux:
@janmeier - I reviewed other scenarios and took into account the changes to mssql for destroying pools (https://github.com/sequelize/sequelize/blob/46a3cd4276d36634a4041fbdadfcb372768d7545/lib/dialects/mssql/connection-manager.js#L101) and I was unable to generate any other error codes than ECONNRESET, so I could not in good faith include any others in the reaping check.

janmeier:
Alrighty then, thanks for you work :)

robraux:
Thank you for the merge @janmeier.

As this is a rather nasty bug, is it possible to mark it as a patch and publish? 

No worries either way, but if possible, it's greatly appreciated.

janmeier:
Published in [`3.12.1`](https://github.com/sequelize/sequelize/releases/tag/v3.12.1)

contactakagrawal:
I have few clarifications on this PR:

1) once we get ECONNRESET error we are invalidating the client, but still the request will result in the error, however for the next request we may still get ECONNRESET because this is the another client out of minimum configured clients in the pool, so we will keep on getting this error for those min no. of clients.
2) why are we invalidating the client only in ECONNRESET, why not also for other kind of errors.
3) should not we just invalidate the entire pool for such errors and create a fresh pool

Thanks,
Amit

PR #4667, opened 2015-10-16
Fixed generated SQL when using find with through.where and required true

Americas:
when the primary key attribute is different from the primary key field

Kind of relates to ##3940. Using the primaryKeyAttribute generates wrong SQL if the database field and the model attributes are named differently.


Americas:
Seems like my changes don't fix the problem on mssql databases.. I can probably get my hands on one to see what the problem is

mickhansen:
> Invalid column name 'name'.
That's odd, but mssql is a strange beast.

Otherwise great work, if you feel like taming mssql we'd be happy, but otherwise i'm okay with just skipping this test for mssql untill we can get an expert on it.

Americas:
Found the problem, another primaryKeyAttribute on mssql's query-generator funtion 'addLimitAndOffset'

The failing check now doesn't have anything to do with the change.

Issue #4404, opened 2015-08-28
MSSQL query generator uses OFFSET FETCH clause only supported in SQL Server 2012 and up


treythomas123:
The query generator uses an OFFSET FETCH clause to apply the offset and limit parameters, which works fine in SQL Server 2012 and newer, but is unsupported in older versions of SQL Server.

[Link to relevant code](https://github.com/sequelize/sequelize/blob/88217aea7ec6380739697de89dba2a830ada19ea/lib/dialects/mssql/query-generator.js#L586-L599)

I found this when I tried to use `findById` against an SQL Server 2008 R2 database. The server came back with this error message:

    Invalid usage of the option NEXT in the FETCH statement.

Looking at the code, it looks like this will also happen when using `findOne`, or when the `offset` or `limit` options are used in a `findAll`, `findAndCount`, etc.

So the `offset` and `limit` functionality should either be changed to use a syntax supported by all SQL Server versions, or there should be logic to fall back to an old method when a version before 2012 is detected. Or the docs could be updated to specify that only versions newer than 2012 are supported.

janmeier:
We do have the functionality for conditionally changing syntax based on db version, see e.https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L253

Pull request with functionality for older versions would be welcome.

However, I guess the syntax is more terse for older versions?

treythomas123:
Thanks for the link @janmeier. Will work on a pull request if my team ends up using Sequelize for our project.

jacobbridges:
I found this same issue today. But so far I only get the error when doing `Model.findById`. `Model.findOne` works perfectly and does not return the `Invalid usage of the option NEXT in the FETCH statement.` error.

mickhansen:
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L1435 needs to be removed, don't need a limit when we're looking for a primary key.

romuloctba:
I'm getting `Invalid usage of the option NEXT in the FETCH statement` with this code:

```
sequelize.Clientes.findAll({limit: 30, offset: 5  }) 
	 .then(function(result) {
    res.json({result: result}); //this never happens :P
  });
```

It generates the following query sintax: `... from [Clientes] ORDER BY [cnpj] OFFSET 5 ROWS FETCH NEXT 30 ROWS ONLY`

wich is not supported by mssql 2008.
Well, I tried setting `tedious` TDS version with

```
[...]

PR #4676, opened 2015-10-18
Remove usage of "limit" in some places where it's unnecessary

treythomas123:
The queries for `Model.findById` and `Instance.reload` don't need to specify a limit of one record, since we're querying with the primary key anyway. Including the limit in the generated SQL is redundant and adds unnecessary complication to the queries.

This is also somewhat related to [issue 4404] (https://github.com/sequelize/sequelize/issues/4404) -- these changes allow `findById` and `reload` to work on MS SQL Server versions before 2012, where previously they were broken due to the `OFFSET FETCH` issue. 4404 isn't completely fixed though, because queries that actually need the `offset` and `limit` options are still broken.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1466,10 +1466,8 @@ Model.prototype.findById = function(param, options) {
     throw new Error('Argument passed to findById is invalid: '+param);
   }
 
-  // Bypass a possible overloaded findAll.
-  return Model.prototype.findAll.call(this, _.defaults(options, {
-    plain: true
-  }));
+  // Bypass a possible overloaded findOne
+  return Model.prototype.findOne.call(this, options);
 };
 Model.prototype.findByPrimary = Model.prototype.findById;
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -1440,19 +1440,19 @@ Model.$findSeparate = function(results, options) {
 };
 
 /**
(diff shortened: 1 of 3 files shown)


lib/model.js
```diff
@@ -1466,10 +1466,8 @@ Model.prototype.findById = function(param, options) {
     throw new Error('Argument passed to findById is invalid: '+param);
   }
 
-  // Bypass a possible overloaded findAll.
-  return Model.prototype.findAll.call(this, _.defaults(options, {
-    plain: true
-  }));
+  // Bypass a possible overloaded findOne
+  return Model.prototype.findOne.call(this, options);
```
mickhansen: This change wouldn't actually have done anything as far as i can tell? findAll does not add a limit, neither does findById.

treythomas123: @mickhansen you are correct, I hadn't noticed [your commit](https://github.com/treythomas123/sequelize/commit/35b9558fc09b75239b6cab955dba9546a8ec9507) that removed the limit on `findById`.  I had been using a slightly older version where it was still adding the limit, and made this change without checking to see if it was still broken on master. 

I think `findOne` still makes sense here though -- it's a little cleaner since the addition `{plain: true}` option doesn't have to be repeated (it's already being added within `findOne`).

Issue #1608, opened 2014-04-08
Prepared Statements and Security


noah-goodrich:
According to https://github.com/sequelize/sequelize/issues/998, the issue of prepared statements and parameter binding through the native mysql driver was considered and rejected previously.

I would like to reopen the issue as a security concern for queries being run against the server. 

Frameworks in other languages use bound parameters natively because it is inherently more secure than doing the escaping in the framework and passing pure strings to the mysql server to process. 

Can you explain the justification for not using prepared statements by default in Sequelize for security purposes?

calvintwr:
@mickhanson if it still interests you. 

http://security.stackexchange.com/questions/3611/sql-injection-why-isnt-escape-quotes-safe-anymore

From a theoratical point of view, prepare statements would prepare the access plans awaiting the parameters to be passed in, hence shouldn't be slow, as compared to javascript escaping the queries.

And plus, it was viewed that it's never possible to cover the myriad of possibilities to inject given all possible permutations of sql constructs. Though general escaping would reduce such possibilities to near zero.

Prepare statement the solution that attacks the fundaments and a very common sense way: That is the thing that you were intending to do has changed so much after the parameters have been passed in, the third party must have passed in need something that changed what you have intended. Thats quite rock solid.

efuquen:
@mickhansen `pg` and `sqlite3` both have support for prepared statements.  `node-mysql2` is suppose to be API compatible with `node-mysql`, meaning you should just be able to drop it in and get the benefits of prepared statements without breaking anything.  Obviously it should be tested.

Overall I'm disappointed this is not considered a priority over other issues or features.  The security benefits are well documented, personally I would think that would trump most if not all other issues outside of bugs breaking basic functionality.  And then there are the performance benefits you get as a bonus. 

I get this is an open source project and if I really needed this I could submit a PR myself (which I don't, since I'm not using sequelize).  But there is a difference between "we as maintainers don't have the time to implement new features like this" vs "we as maintainers have time to do some stuff, but we've decided this isn't important enough".  I consider something that would improve security greatly and make the web a safer place of critical importance, so if other features are being prioritized over this it really bums me out.

mickhansen:
@efuquen I understand your sentiment. Can you quantify the exact security benefits? So far the single security vulnerability in sequelize has been from an injection into the column name, which prepared statenents wouldn't have done much again. 

[...]

Issue #4494, opened 2015-09-14
Raw query with named substitution parameters gets quoted parameters


czardoz:
I'm using sequelize to perform a simple raw query:

```
    var queryStr = 'SELECT * FROM :tableName WHERE :deletionColumn > 0 LIMIT :limit;';

    return self.connection.query(queryStr, {
        replacements: {
            tableName: self.name,
            deletionColumn: self.options.deletionColumn,
            limit: self.options.limit
        }
    }).then(function (rowsToDelete) {
        _.each(rowsToDelete, function (row) {
            log.debug('Preparing to delete row: ', row);
        });
    });

```

The resulting query however, is (from the sequelize logs): 

```
Executing (default): SELECT * FROM 'test' WHERE 'deleted' > 0 LIMIT 421;
```

The generated query is invalid (table name must not be quoted).

Are there any options I can pass that'll disable the quoting? 
I'm using MySQL 5.6 here.

janmeier:
This is not possible currently

First we'd need to come up with a syntax for this - any suggestions @czardoz ?

czardoz:
Coming up with new syntax might not be necessary. 

Something like this should work too:
```
     replacements: {
            tableName: {value: self.name, quote: false}
        }
```

janmeier:
Yeah, that coudl work as well :)

PR welcome

mickhansen:
Hmm, but the value of `tableName` could be a JSON object that needs to be inserted as JSON.
We probably want to use `$` prefixed keys.

mickhansen:
This change would also not make a lot of room for introducing proper prepared queries later, perhaps a different identifier than `:` for tables/identifiers? Not sure what other libraries do for this type of stuff? Some research might be warranted.

janmeier:
I do remember we previously had support for using `??` as unescaped with unnamed replacement - but that won't work since `::` has a different meaning in postgres

czardoz:
This is a pretty uncommon use case (I think), and I am not sure if this needs to be implemented at all. I've currently worked around this using Mustache, and it works just fine. The documentation should definitely be updated to point this out though.

subjectix:
guys come on, don't we already have a familiar syntax for this: `sequelize.col('deleted')` :grin:

janmeier:
@subjectix Quoting columns is not the issue - its _not_ quoting everything else

subjectix:
[...]

PR #4688, opened 2015-10-20
implement bind-parameter for custom queries on postgres

User4martin:
with regards to issue #1608 

I did a tiny bit of work starting on this. Its postgres only. And custom query only.

I moved parameter handling into the dialect, so each dialect can then call the current quote+replace, or implement bind param.

The added code in dialect/postgres/query is copied from sqlStrings and modified a little.

----
If the general direction of this is acceptable, then this could be a first step.
I do however not know at current if I will be able to look at the next steps (i.e. make other existing queries use params)

### DIFF
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -434,32 +434,29 @@ AbstractQuery.prototype.formatBindParameters = function(sql, values, dialect, re
   }
   
   var timeZone = null;
-  if (Array.isArray(values)) {
-    sql = sql.replace(/\$(\$|(?:\d+))/g, function(match, key) {
-      if ('$' === key) {
-        return options.skipUnescape ? match : key;
-      }
+  var list = Array.isArray(values);
   
-      key = key - 1;
-      var replVal = replacementFunc(match, key, values, timeZone, dialect, options);
-      if (replVal === undefined) {
-        throw new Error('Named bind parameter "' + match + '" has no value in the given object.');
+  sql = sql.replace(/\$(\$|\w+)/g, function(match, key) {
+    if ('$' === key) {
+      return options.skipUnescape ? match : key;
+    }
+    
+    var replVal;
+    if (list) {
+      if (key.match(/^[1-9]\d*$/)) {
+        key = key - 1;
+        replVal = replacementFunc(match, key, values, timeZone, dialect, options);
       }
-      return replVal;
-    });
-  } else {
-    sql = sql.replace(/\$(\$|(?!\d)(?:\w+))/g, function(match, key) {
-      if ('$' === key) {
-        return options.skipUnescape ? match : key;
+    } else {
+      if (!key.match(/^\d*$/)) {
+        replVal = replacementFunc(match, key, values, timeZone, dialect, options);
       }
-  
-      var replVal = replacementFunc(match, key, values, timeZone, dialect, options);
-      if (replVal === undefined) {
-        throw new Error('Named bind parameter "' + match + '" has no value in the given object.');
-      }
-      return replVal;
-    });
-  }
+    }
+    if (replVal === undefined) {
+      throw new Error('Named bind parameter "' + match + '" has no value in the given object.');
+    }
+    return replVal;
+  });
   return [sql, []];
 };
 
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var Utils = require('../../utils')
+  , SqlString = require('../../sql-string')
   , Dot = require('dottie')
   , QueryTypes = require('../../query-types');
 
@@ -382,11 +383,86 @@ var groupJoinData = function(rows, includeOptions, options) {
 
 /**
  * rewrite query with parameters
+ *
+ * Examples:
+ *
+ *   query.formatBindParameters('select $1 as foo', ['fooval']);
+ *
+ *   query.formatBindParameters('select $foo as foo', { foo: 'fooval' });
+ *
+ * Options
+ *   skipUnescape: bool, skip unescaping $$
+ *   skipValueReplace: bool, do not replace (but do unescape $$). Check correct syntax and if all values are available
  */
-AbstractQuery.prototype.formatBindParameters = function(sql, parameters, dialect) {
-  sql = Utils.formatBindParameters(sql, parameters, dialect);
+AbstractQuery.prototype.formatBindParameters = function(sql, values, dialect, replacementFunc, options) {
+  if (!values) {
+    return [sql, []];
+  }
+  
+  options = options || {};
+  if (typeof replacementFunc !== 'function') {
+    options = replacementFunc || {};
+    replacementFunc = undefined;
+  }
+  
+  if (!replacementFunc) {
+    if (options.skipValueReplace) {
+      replacementFunc = function(match, key, values, timeZone, dialect, options) {
+        if (values[key] !== undefined) {
+          return match;
+        }
+        return undefined;
+      };
+    } else {
+      replacementFunc = function(match, key, values, timeZone, dialect, options) {
+        if (values[key] !== undefined) {
+          return SqlString.escape(values[key], false, timeZone, dialect);
+        }
+        return undefined;
+      };
+    }
+  } else {
+    if (options.skipValueReplace) {
+      var origReplacementFunc = replacementFunc;
+      replacementFunc = function(match, key, values, timeZone, dialect, options) {
+        if (origReplacementFunc(match, key, values, timeZone, dialect, options) !== undefined) {
+          return match;
+        }
+        return undefined;
+      };
+    }
+  }
+  
+  var timeZone = null;
+  if (Array.isArray(values)) {
+    sql = sql.replace(/\$(\$|(?:\d+))/g, function(match, key) {
+      if ('$' === key) {
+        return options.skipUnescape ? match : key;
+      }
+  
+      key = key - 1;
+      var replVal = replacementFunc(match, key, values, timeZone, dialect, options);
+      if (replVal === undefined) {
+        throw new Error('Named bind parameter "' + match + '" has no value in the given object.');
+      }
+      return replVal;
+    });
+  } else {
+    sql = sql.replace(/\$(\$|(?!\d)(?:\w+))/g, function(match, key) {
+      if ('$' === key) {
+        return options.skipUnescape ? match : key;
+      }
+  
+      var replVal = replacementFunc(match, key, values, timeZone, dialect, options);
+      if (replVal === undefined) {
+        throw new Error('Named bind parameter "' + match + '" has no value in the given object.');
+      }
+      return replVal;
+    });
+  }
   return [sql, []];
 };
+
 /**
  * Execute the passed sql query.
  *
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -381,6 +381,13 @@ var groupJoinData = function(rows, includeOptions, options) {
 };
 
 /**
+ * rewrite query with parameters
+ */
+AbstractQuery.prototype.formatBindParameters = function(sql, parameters, dialect) {
+  sql = Utils.formatBindParameters(sql, parameters, dialect);
+  return [sql, []];
+};
+/**
  * Execute the passed sql query.
  *
  * Examples:
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -394,7 +394,7 @@ var groupJoinData = function(rows, includeOptions, options) {
  *   skipUnescape: bool, skip unescaping $$
  *   skipValueReplace: bool, do not replace (but do unescape $$). Check correct syntax and if all values are available
  */
-AbstractQuery.prototype.formatBindParameters = function(sql, values, dialect, replacementFunc, options) {
+AbstractQuery.formatBindParameters = function(sql, values, dialect, replacementFunc, options) {
   if (!values) {
     return [sql, []];
   }
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -90,11 +90,11 @@ Query.prototype.parseDialectSpecificFields = parseDialectSpecificFields;
 /**
  * rewrite query with parameters
  */
-Query.prototype.formatBindParameters = function(sql, values, dialect) {
+Query.formatBindParameters = function(sql, values, dialect) {
   var bindParam = [];
   if (Array.isArray(values)) {
     bindParam = values;
-    sql = this.parent.formatBindParameters(sql, values, dialect, { skipValueReplace: true })[0];
+    sql = AbstractQuery.formatBindParameters(sql, values, dialect, { skipValueReplace: true })[0];
   } else {
     var i = 0;
     var seen = {};
@@ -110,7 +110,7 @@ Query.prototype.formatBindParameters = function(sql, values, dialect) {
       }
       return undefined;
     };
-    sql = this.parent.formatBindParameters(sql, values, dialect, replacementFunc)[0];
+    sql = AbstractQuery.formatBindParameters(sql, values, dialect, replacementFunc)[0];
   }
   return [sql, bindParam];
 };
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -87,7 +87,50 @@ Utils.inherit(Query, AbstractQuery);
 
 Query.prototype.parseDialectSpecificFields = parseDialectSpecificFields;
 
-Query.prototype.run = function(sql) {
+/**
+ * rewrite query with parameters
+ */
+Query.prototype.formatBindParameters = function(sql, values, timeZone, dialect) {
+  var bindParam = [];
+  if (Array.isArray(values)) {
+    bindParam = values;
+    sql = sql.replace(/\$(\$|(?:\d+))/g, function(value, key) {
+      if ('$' === key) {
+        return key;
+      }
+  
+      key = key - 1;
+      if (values[key] !== undefined) {
+        return value;
+      } else {
+        throw new Error('Named bind parameter "' + value + '" has no value in the given object.');
+      }
+    });
+  } else {
+    var i = 0;
+    var seen = {};
+    sql = sql.replace(/\$(\$|(?!\d)(?:\w+))/g, function(value, key) {
+      if ('$' === key) {
+        return key;
+      }
+  
+      if (seen[key] !== undefined) {
+        return seen[key];
+      }
+      if (values[key] !== undefined) {
+        i = i + 1;
+        bindParam.push(values[key]);
+        seen[key] = '$'+i;
+        return '$'+i;
+      } else {
+        throw new Error('Named bind parameter "' + value + '" has no value in the given object.');
+      }
+    });
+  }
+  return [sql, bindParam];
+};
+
+Query.prototype.run = function(sql, parameters) {
   this.sql = sql;
 
   if(!Utils._.isEmpty(this.options.searchPath)){
@@ -96,7 +139,7 @@ Query.prototype.run = function(sql) {
 
   var self = this
     , receivedError = false
-    , query = this.client.query(this.sql)
+    , query = ((parameters && parameters.length) ? this.client.query(this.sql, parameters) : this.client.query(this.sql))
     , rows = [];
 
   this.sequelize.log('Executing (' + (this.client.uuid || 'default') + '): ' + this.sql, this.options);
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -90,30 +90,15 @@ Query.prototype.parseDialectSpecificFields = parseDialectSpecificFields;
 /**
  * rewrite query with parameters
  */
-Query.prototype.formatBindParameters = function(sql, values, timeZone, dialect) {
+Query.prototype.formatBindParameters = function(sql, values, dialect) {
   var bindParam = [];
   if (Array.isArray(values)) {
     bindParam = values;
-    sql = sql.replace(/\$(\$|(?:\d+))/g, function(value, key) {
-      if ('$' === key) {
-        return key;
-      }
-  
-      key = key - 1;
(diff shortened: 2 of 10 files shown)


User4martin:
Sorry about the tests, I should have fixed it, but I have only postgres in my own app for testing.
I tried to follow the steps on https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md but the test can not find some part of docker-compose (and yes I installed it with npm, I did so as root /sudo, even tried -g)

mickhansen:
docker-compose is not an NPM thing but a docker thing to help you easily setup the entire environment needed.

User4martin:
Now I got tests for sqlite and postgres running locally.

The changes now work with one exception. My opinion is that the new failure is actually desirable and the test should be fixed/extended.

postgres needs to be able to detect the type of a param from the sql. This is fine in a case like "where column = $1" because column provides the type.
But "select $1 as foo" needs to be "select $1::int as foo"

That happens in one of the tests.

It may be possible to detect the type from the data, when sequelize creates the "$1". But this may have limits. a string may be varchar or an enum.

Also it is not desirable. If security is needed desired then dependency on data (that may come from an untrusted source) should be reduced. The sql should contain the data type.

This issue is only for costumn queries.
Queries from sequelize actions usually include columns.

However I understand, if this new restriction is to be kept, then the entire feature may need to be optional. After all existing usercode could currently have such sql and then stop working.

Adding an option will be easy. Well I may need a pointer where to add it. But once the option is present, line 721 in sequelize.js could check it, and call the old replacment code.

Please let me know your opinion. Thanks

User4martin:
I have added the option, named "useBindParameters" (bool).

Without the option all stays as it was, and no changes to tests.

With the option, the query must match the expectations of the db. (e.g. have ::type)

I have added 2 new tests for useBindParameters = true. Testing that ? and :name param are replaced.

mickhansen:
Whats the syntax for bound parameters in postgres, `$1`?
We currently use `?` or `:name` for parameters.
Could we instead just support new syntax `$1 || $name` so users can opt into the new syntax?

User4martin:
postgres uses $1 $2 ...., and the same token can be used repeated
```
select foo from bar where fooX= $1 or fooY = $1;
```
I have not researched what other db dialects uses. 

In order to allow keep user code dialect independent it makes sense to keep the current syntax.

Though of course it will not be entirely dialect free, due to restrictions where params may be used and where not. E.g. in postgres a param can not be a column or table name. it can only be data. I believe that is true for all dialects. But the need to ensure the  type of a param can be detected ($1::int) may be dialect specific.

Furthermore, if in future params are used by sequelize build sql e.g. Foo.find({where:{bar: 1}}) then the code generater needs a dialect independent way to write the sql statement.

----
However extensions could be added.
```
sequelize.query(
  'select foo from bar where foo like $1',
  { rawSql: true, replacements: ['a%'] }
);
```

Similar in addition to the global option useBindParameters, a per call option can be added
```
sequelize.query(
  'select foo from bar where foo like ?', 
  { useBindParameters: true, replacements: ['a%'] }
);
```
----

Regarding the need for a typecast in cases where the type is not given by context;
It would be possible to add something like
```
sequelize.query(
  'select foo from bar where foo like ?',
  { useBindParameters: true, replacements: [{value: 'a%', type: 'varchar'}] } // dialect specifec
);
sequelize.query(
  'select foo from bar where foo like ?',
  { useBindParameters: true, replacements: [{value: 'a%', type: sequelize.STRING}] }
);
sequelize.query(
  'select foo from bar where foo like :val',
  { useBindParameters: true, replacements: {val: {value: 'a%', type: sequelize.STRING}} }
);
```

But that can be added  separately?

mickhansen:
I prefer a syntax change rather than an options change because ideally we could then start advertising the new syntax on docs etc rather than having a hidden option some where.

/cc @janmeier thoughts?

User4martin:
I am happy to do the syntax change as an addition.
But as I said, it has several downsides, which is why I would at least offer both. (or maybe there is a 3rd option)

It seems $1 is not used by all sql dialects

Just googled / not tested:
mysql appears to use "?" https://dev.mysql.com/doc/refman/5.0/en/sql-syntax-prepared-statements.html

sqlite seems to have a whole lot of options: https://www.sqlite.org/c3ref/bind_blob.html (and $1 should be in there)

I do not know if the npm for mysql/sqlite supports this

-----
1) If 
```
 Foo.find({where:{bar: 1}}) 
```
should also use bind-params, then what sql should the query generator create?

If it creates $1 based sql then mysql-dialect handlers translate need to translate it (that would be an option).

I prefer that they generate :name based sql.

----
2) users have more work upgrading there current code.

bindReplacements (below) is easier to change than every :name into the correct $n

-----
On the other hand using a new syntax (or at least something new) means both are mixable.
```
select :foo = $1
```
That might be nice.

Here is an other idea
```
sequelize.query(
  'select :foo from bar where foo like :bar',
  { replacements: { foo: '*' },  // replaced in the old fashion
    bindReplacements: { bar: 'a%' }  // uses bind param
  }
);
```

in case of "?" it must be either  "replacements" or "bindReplacements", that is all "?" are either client side replaced (old approach) or all of them are bind param.

But with :name it can be mixed

It would also allow
```
sequelize.query(
  'select :foo from bar where foo like ?',
  { replacements: { foo: '*' },  // replaced in the old fashion
    bindReplacements: [ 'a%' ]  // uses bind param
  }
);
```

if going for "replacements" / "bindReplacements" I would *not* add support for the user writing $1 in their sql (imho not needed).

Though if you wanted it it could be done, but then it needed a different way to pass the replacements, so they would not attempt pre-processing. (Actually that would be true, even if the existing syntax was not extended for bind param)
Maybe
```
sequelize.query(
  'select  * from bar where foo like $1',
  { rawReplacements: [ 'a%' ] }
);
```
rawReplacements would be handed unmodified to the db-driver. So for postgres they must be an array, but other drivers may expect other data.

rawReplacements imho have only one usecase: they save a tiny bit of time, since they need no translation.

------
------
Personally I would go for "replacements" and "bindReplacements" both using ? / :name

And if you want add "rawReplacments" (or any better name) for replacments that need no processing in sequelize (user already wrote $1 or what the database needs)

janmeier:
Hmm, I don't really know what I think of this one - I don't have any strong opinion. The cleanest would be to introduce a new `$` syntax, and it would make an explicit distinction between whether you were using bind params or escaped strings. 

The name `replacements` might be a bit confusing since its no longer going to be replaced but bound. (which is, admittedly a very small semantic difference :D)

Honestly, I don't really know, but I'm leaning towards `$` instead of a new option

As a side-note, bind params / prepared statements are not support by node-mysql https://github.com/felixge/node-mysql/issues/274. node-mysql2 does support them though https://github.com/sidorares/node-mysql2#prepared-statements

User4martin:
Then maybe both? 
"replacements" and "bindReplacements"  and "rawReplacements"
(names can be changed)

one reason for doing :name as bound param is that it may help when rewriting internal sql.

For example: upsert, iirc it calls update and select and combines the 2 sql.

If the create paramized sql, and use $1 then the 2nd sql may have to start at $10.
That can be easily done, but the 2nd sql will use the same values again. so if both did :columnname then that cuts param by half (and sends less to the db, which is good if this includes a huge blob value)

of course we could additionally add $name..
----

The other point is that using "select $1" does still need a new name for replacement.

if "query()" finds the option "replacement" it will look for ? or :name. So it needs a new option (e.g. bindReplacments) in order to know what to look for.

If not it needs to perform additional work (and if internal queries use the feature this work has do be done always, and time adds up).
It needs to scan the query for ? : and $ and hope it finds only one. But "select ? = '$1';"  is a valid (old style) query.

----
Actually the last one is an example why adding the $ syntax is bad.
```
select $1 = '$2';
```
send to postgresql has only ONE param. (the literal is not a param).

But if node mysql starts supporting bind param, and the mysql dialect needs to translate the above to
```
select " = '$2';
```
how does it know not to translate the literal? (can be done but more parsing needed.

Again if the query comes from the user, then for mysql we could tell the user to use "?". (Which gets the user into trouble if he wants to change database)

But if the query comes from an internal generator, then it will need a dialect independent syntax.

User4martin:
And then if functions like User.find(...) should use the bind param, that would mean a global option will be needed then?

mickhansen:
The `replacements` code could be refactored to look for `$` prefixed and then apply different logic to those for dialects that support param binding. (or we could add a param `bind` that does the samething as replacements).

AFAIR bindings don't work for literal values generally, do they?

Implementing param binding in model methods is a lot more work in any case i'm afraid.

User4martin:
I am happy with using $ syntax and a new param "bind" instead of replacements.

bind can then support 
```
$1, $2  bind: [1,2]

$foo $bar bind:{foo:1, bar:2}
```

That makes total sense.

In that case should it allow the following, or give an error?
```
select :foo from User where a = $bar;
replacements:{foo: 1}
bind:{bar: 2}
```

If we can agree an the usage of bind, I am happy to work on that.

----------------------

But that still does not answer the final question.
```
User.find({where:{ a:1}})
```
If that code can generate either of the following
```
select * from User where a = 1;
select * from User where a = $1;
```
Then how will it know which one to generate?

User4martin:
I think using "bind" for $ (instead of "replacement" for both $ and :n / ?) is better to the user too.

if replacements had to detect the presence of $, the user would need to worry about that, even if he does not want to bind.

Extending replacment for $ could have broken existing code. Using "bind" as new param, will not break anything.

mickhansen:
Supporting bind params for stuff like `User.find({where:{ a:1}})` is a much larger refactor, lets keep it out of this scope for now.

`bind` will indeed not break anything, i think using both bind and replacements should throw.

Using `bind` we can chose to either adopt a new syntax and make it work cross-dialect or we can chose to use the old dialect and make that work cross dialect by rewriting it to platform specific syntax.

User4martin:
Ok so lets come up with a clear specification, so I can then start implementing

* sequelize.query will take a new argument "bind"
* "bind" and "replacement" are not allowed at the same time
 * It is an easy check, and can if needed be changed at any time. If #4494 will be solved, then using both at them may actually be desirable.
 * For now, not to allow them has the advantage that there is no worry, that "parameters" could put a $1 into the sql, which would then be seen by "bind"
* "replacements" will be unchanged and not affected (except for not being allowed if "bind" is used)
* "bind" tokens will start with a $ the dialect specific code can adapt this per database
* "bind" can take an array or object with keys.
* "bind" will do the the same error checking as "replacements" (e.g. values for token must be present)
 * therefore bind will assume every $... to be a token, even if a specific database would not see it that way
 * select a = "$2" where foo = $1; Postgres may not see $2. But "bind" takes this as dialect independent sql, and will expect a value for it
 * Optional: If the "bind" param is present, then "$$" could be an escape sequence for a $. So the above query could be  select a = "$$2" where foo = $1;
* if a dialect can not handle "bind" then it will replace it into the sql in the same way that "replacements" are done.

The 2 possible syntax will be (decided by the type of argument to "bind"
Array:
```
sequelize.query(
    'select id = $1 from bar where foo = $2 and num = $1',
    { bind: [ 1, 'fooval' ] }
);
```
Object
```
sequelize.query(
    'select id = $val from bar where foo = $foo and num = $val',
    { bind: {val: 1, foo: 'fooval' } }
);
```

----
If the above spec is acceptable, I can start on the implementation.

mickhansen:
> if a dialect can not handle "bind" then it will replace it into the sql in the same way that "replacements" are done.
And over time hopefully we'll have "bind"-support for all dialects (letting the server inject the values).

I like the spec. I don't see any breaks since it requires users to use a new parameter.
/cc @janmeier

janmeier:
Yep, I like it!  :)

User4martin:
Ok please have a look.

A couple of notes.
* when passing an array "bind: []", any none numeric $ ($foo) are ignored.
* when passing an object "bind: {}" and numeric ($1) or to be exact any starting with a number ($1a) are ignored. This is the same for :1a .  
 * as $$ serves as escape this is not strictly needed, but allowing pure numeric $1 only with arrays seems cleaner.
* implemented for postgres and sqlite
 * on postgres, I did check (with my own code) that the database log shows that bind-param were indeed received.
 * on sqlite, I dont have my own setup, so all I can do is hand it over to node-sqlite in the correct format.
* for both db it is *not* possible to use bind with multiple sql statements in one query. "select 1; select 2"
 * on sqlite, this happens if sqlite.exec is used. (currently only upsert). I added code for sqlite to fallback to replacement in that case. This is not strictly needed since such a statement should never be passed with bind param. But without this there would be an invalid API call to the node sqlite. (While in other cases it gives an error from the db)

I took all the testcases I could find for replacements, and duplicated them for bind.

mickhansen:
> when passing an array "bind: []", any none numeric $ ($foo) are ignored.
Not sure if this should actually throw for both cases. @janmeier?

User4martin:
If I read the code correctly :name also ignores (without throw) :1name
That is what I based it on. (so $1a is also ignored with bind = {})

If it should throw, then there are several options:
* $1a could be allowed. and only $1 would be disallowed.
* bind [] throws on $foo (since that can not be found in an array), but bind {} allows $1, if "1" is a key in the object.
* bind {} always throws on $1 (do distinguish between numeric and alpha mode)

----
Also maybe instead of doing replacements if sqlite.exec is used, maybe this should also throw instead. "exec" is usually (but not necessarily) used for multi query, and that fails with all other db too (all current db, so that is just postgress)

lib/dialects/postgres/query.js
```diff
@@ -87,7 +87,50 @@ Utils.inherit(Query, AbstractQuery);
...
+        bindParam.push(values[key]);
+        seen[key] = '$'+i;
+        return '$'+i;
+      } else {
+        throw new Error('Named bind parameter "' + value + '" has no value in the given object.');
+      }
+    });
+  }
+  return [sql, bindParam];
+};
```
mickhansen: This block looks extremely similar to the utils based one.

lib/dialects/sqlite/query.js
```diff
@@ -24,11 +24,61 @@ Query.prototype.getInsertIdField = function() {
...
+  
+      if (values[key] !== undefined) {
+        return value;
+      } else {
+        throw new Error('Named bind parameter "' + value + '" has no value in the given object.');
+      }
+    });
+  }
+  return [sql, bindParam];
+};
```
mickhansen: Same here, looks very similar, what are the actual differences, can they be refactored?

lib/utils.js
```diff
@@ -63,6 +63,13 @@ var Utils = module.exports = {
     var timeZone = null;
     return SqlString.formatNamedParameters(sql, parameters, timeZone, dialect);
   },
+  formatBindParameters: function(sql, parameters, dialect, options) {
+      if (parameters) {
+        var timeZone = null;
+        return SqlString.formatBindParameters(sql, parameters, timeZone, dialect, options);
+      }
+      return sql;
```
mickhansen: Too much indentation.

lib/sequelize.js
```diff
@@ -707,6 +707,14 @@ Sequelize.prototype.query = function(sql, options) {
       options.replacements = sql.values;
     }
 
+    if (sql.bind !== undefined) {
+      if (options.bind !== undefined) {
+        throw new Error('Both `sql.bind` and `options.bind` cannot be set at the same time');
+      }
+
+      options.bind= sql.bind;
```
mickhansen: Needs a space before `=`

mickhansen:
> $1a could be allowed (if present in {} bind). and only $1 would be disallowed.

How would PG handle it? I say disallow unless starting with alphanumeric tho.

> bind [] throws on $foo (since that can not be found in an array), but bind {} allows $1, if "1" is a key in the object.

Agreed on the first part.

> bind {} always throws on $1 (do distinguish between numeric and alpha mode)

Agreed.

> bind [], should $0 throw ? $n is 1 based.

Should throw

> bind [], should $01 vs $1 ?

Should throw if we can do it without false negatives.

User4martin:
similarity of formatBindParameters between sqlite and postgres

* postgres only takes $1 style.
 * if bind = [] then sql and bind can be passed unmodified. The code only checks for correctness (all values present)
 * if pind = {} then it must be translated to $1  syntax
* sqlite takes a lot of different syntax.
 * bind =[] would require "?" or "?1" in the sql. instead bind is translated to { $1: 'foo', $2: 'bar' }. Note the $ is part of the key. That differs from postgres.
 * bind = {} the difference to postgres is that bind param must include the $ in the object keys.{ $foo: 'foo'}

For bind = [] both do the param first (copy/translate) ,and then check them. The loop for checking is the same. 
If it is refactored, you prefer it in sqlStrings, or in the base class abstractQuery ?

For bind = {} they are different. If I extracted the verification part, then nothing changes for sqlite (except the current function gets split into two), Postgres will save only 2 or 3 lines, but has to run 2 loops instead of one. IMHO non a good idea.
I can still split it for future use though.

-----

Sorry about the spacing, I tried to stick to the existing style. Do you use any tool that checks formatting, and that I can run?
Will fix it with the next iteration of the patch.

----

>> $1a could be allowed (if present in {} bind). and only $1 would be disallowed.
>
> How would PG handle it? I say disallow unless starting with alphanumeric tho.

PG only accepts $1.  $1a gives an error.
However sqlite accepts $alpha. I did not see anyting in the doc, that $1a would not be allowed, and passing $1 (as alpha) is accepted (the current patch does this, and passes tests)

The question should be:
* Do we want to apply similar rules to `$foo` syntax, as the rules for `:foo` ?  I do not know why :1foo is not allowed, other than the test suggests it conflicts with `00:00`. the $ syntax offers an escaped $ as $$.
* Do we want to impose the restriction for readability.
Otherwise it is possible to allow any $alphanum for bind = {}.

>> bind [] throws on $foo (since that can not be found in an array), but bind {} allows $1, if "1" is a key in the object.
>
> Agreed on the first part.
>> bind {} always throws on $1 (do distinguish between numeric and alpha mode)
>
> Agreed.
Ok, that is current behaviour.

Leaves the above `$1a` and how to handle it when bind={} (accept or throw)

>> bind [], should $0 throw ? $n is 1 based.
>
> Should throw
Will add a test, and fix code if needed.

>> bind [], should $01 vs $1 ?
>
> Should throw if we can do it without false negatives.
Can be done.

----------
one more question

`select $` `select "$"` `select $*`  $ not followed by alphanum (match /\w/). This is currently ignored. And probably better like this. Otherwise the user must escape any such $.
This is the same for ":" :* ...
Is that ok?

janmeier:
> If it is refactored, you prefer it in sqlStrings, or in the base class abstractQuery ?

I'd prefer it in abstract query - sql-string was originally copied from node-mysql (although it's heavily modified) and I'd rather get rid of it at some point, move the logic to query + query-generator

User4martin:
I refactored the replaceBind code. Anything I had added in sqlStrings, I moved  to abstract query. I extracted any dialect specific code into a callback, and put the rest into a single function.

When using bind it will now check anything matching `/\$\w+/`

* bind = [] 
 * accepts $1 .. $n (if index is in array)
 * throws for $foo
 * throws for $0
 * throws for $01
* bind = {} 
 * accepts any key that is in object
 * except (throws even if key is in object) for $1 .. $n
 * except (throws even if key is in object) for $0  $01 ...

User4martin:
Not sure what happened to travis, it showed success. then changed to failed. But according to detail page all tests passed (except coverage, which is an allowed failure)

mickhansen:
@User4martin No clue why it fails based on the allowed failure, rerunning it.

lib/dialects/abstract/query.js
```diff
@@ -381,6 +382,85 @@ var groupJoinData = function(rows, includeOptions, options) {
...
+ *
+ *   query.formatBindParameters('select $1 as foo', ['fooval']);
+ *
+ *   query.formatBindParameters('select $foo as foo', { foo: 'fooval' });
+ *
+ * Options
+ *   skipUnescape: bool, skip unescaping $$
+ *   skipValueReplace: bool, do not replace (but do unescape $$). Check correct syntax and if all values are available
+ */
+AbstractQuery.prototype.formatBindParameters = function(sql, values, dialect, replacementFunc, options) {
```
mickhansen: `dialect` as key should not be required, it should be available with `this`.

mickhansen:
Will give this a thorough review tomorrow (i hope).
@janmeier please review aswell.

User4martin:
>dialect as key should not be required, it should be available with this.

I only seen it on the ...Dialect in dialects/foo/index.js
`PostgresDialect.prototype.name = 'postgres';`

But that is not available from inside the query.

Once we have an instance of query, dialect is available via `this.options.dialect`.
But the code runs before the instance is created, so it can not get it this way.

If we want to wait till the instance is created, we need to allocate a connection, only to throw an error. Also the behaviour would change from currently throwing, to returning a promise and rejecting it. (throw inside promise)

mickhansen:
Ah i see, you're calling it as a static method, in that case please rewrite it to be a static method rather than a prototype method.

User4martin:
I might be missing something, but will that work with inheritance (utils.inherhit)?

Currently the method is on
`AbstractQuery.prototype.formatBindParameters`
and for postgres and sqlite. 
`Query.prototype.formatBindParameters`

So on any class inheriting from any of them it can be called via
`query.prototype.formatBindParameters()`

if I declare it static
`AbstractQuery.formatBindParameters`
`Query.formatBindParameters`
then on a class for which it is declared, I can call
`query.formatBindParameters()`
But on mysql (which inherits from AbstractQuery) this will not work, because utils.inherhit put the inherited method an the prototype

----

If you want I can extend utils.inherhit to copy static methods, and also add them to parent.

mickhansen:
Or simply do a manual inheritance if need be `MysqlQuery.formatBindParameters = AbstractQuery.formatBindParameters`.

User4martin:
Not sure if my commit (for the static preference) caused any notification? Otherwise I'll wait for further feedback.

mickhansen:
@User4martin Commits don't trigger notifications so it can be good to comment that you've made changes :)

mickhansen:
LGTM.

/cc @janmeier

janmeier:
Yep, just needs a bit of documentation, then this looks good to me as well

User4martin:
Added docs. Hope it is in the correct format/place.

mickhansen:
Good work @User4martin and thanks for the documentation!

PR #4706, opened 2015-10-22
pg range handling improved

alekbarszczewski:
@janmeier @mickhansen I've improved pg ranges to handle unbound ranges, +/-infinity bounds and empty ranges:

```js
range.stringify([]) == 'empty'
range.stringify([null, 1]) == '(,1)'
range.stringify([1, null]) == '(1,)'
range.stringify([null, null]) == '(,)'
range.stringify([-Infinity, 10]) == '(-infinity,10)'
range.stringify([-Infinity, Infinity]) == '(-infinity,infinity)'

range.parse('empty', 'int4range') == [] // inclusive: []
range.parse('(,1)', 'int4range') == [null, 1] // inclusive: [false, false]
range.parse('(1,)', 'int4range') == [1, null] // inclusive: [false, false]
range.parse('(,)', 'int4range') == [null, null] // inclusive: [false, false]
range.parse('(-infinity,10)', 'int4range') == [-Infinity, 10] // inclusive: [false, false]
range.parse('(-infinity,infinity)', 'int4range') == [-Infinity, Infinity] // inclusive: [false, false]
```

I've also added/modified proper tests. However there is one thing I am not sure of - before my PR if any bound in an array was null or passed range was not an array at all, then `range.stringify()` returned empty string, which is not a valid PG range literal and caused PG error `ERROR:  malformed range literal: ""`. I am not sure if it is bug or intensional, but I partially left this behaviour:

```js
function stringify (data) {
  if (data === null) return null;
  if (!Utils._.isArray(data)) return ''; // will throw error, since '' is not a valid PG range literal
  if (!data.length) return 'empty';
  if (data.length !== 2) return ''; // will throw error, since '' is not a valid PG range literal
  ...
```
So if range is not an array OR it's not empty array (which means empty range) OR range.length is not 2 then it will throw error as I described above. It could stay as is, but maybe it would be good idea to throw some validation error here instead of leaving this for postgres?

### DIFF
--- a/test/integration/dialects/postgres/range.test.js
+++ b/test/integration/dialects/postgres/range.test.js
@@ -5,18 +5,28 @@ var chai    = require('chai')
   , Support = require(__dirname + '/../../support')
   , DataTypes = require(__dirname + '/../../../../lib/data-types')
   , dialect = Support.getTestDialect()
-  , range   = require('../../../../lib/dialects/postgres/range');
+  , range   = require('../../../../lib/dialects/postgres/range')
+  , _       = require(__dirname + '/../../support').Sequelize.Utils._;
 
 if (dialect.match(/^postgres/)) {
   describe('[POSTGRES Specific] range datatype', function () {
     describe('stringify', function () {
       it('should handle empty objects correctly', function () {
-        expect(range.stringify([])).to.equal('');
+        expect(range.stringify([])).to.equal('empty');
       });
 
-      it('should return empty string when either of boundaries is null', function () {
-        expect(range.stringify([null, 'test'])).to.equal('');
-        expect(range.stringify([123, null])).to.equal('');
+      it('should handle null as empty bound', function () {
+        expect(range.stringify([null, 1])).to.equal('(,1)');
+        expect(range.stringify([1, null])).to.equal('(1,)');
+        expect(range.stringify([null, null])).to.equal('(,)');
+      });
+
+      it('should handle Infinity/-Infinity as infinity/-infinity bounds', function () {
+        expect(range.stringify([Infinity, 1])).to.equal('(infinity,1)');
+        expect(range.stringify([1, Infinity])).to.equal('(1,infinity)');
+        expect(range.stringify([-Infinity, 1])).to.equal('(-infinity,1)');
+        expect(range.stringify([1, -Infinity])).to.equal('(1,-infinity)');
+        expect(range.stringify([-Infinity, Infinity])).to.equal('(-infinity,infinity)');
       });
 
       it('should return empty string when boundaries array of invalid size', function () {
@@ -67,8 +77,22 @@ if (dialect.match(/^postgres/)) {
         expect(range.parse(null)).to.equal(null);
       });
 
-      it('should handle empty string correctly', function () {
-        expect(range.parse('')).to.deep.equal('');
+      it('should handle empty range string correctly', function () {
+        expect(range.parse('empty')).to.deep.equal(_.extend([], { inclusive: [] }));
+      });
+
+      it('should handle empty bounds correctly', function () {
+        expect(range.parse('(1,)', 'int4range')).to.deep.equal(_.extend([1, null], { inclusive: [false, false] }));
+        expect(range.parse('(,1)', 'int4range')).to.deep.equal(_.extend([null, 1], { inclusive: [false, false] }));
+        expect(range.parse('(,)', 'int4range')).to.deep.equal(_.extend([null, null], { inclusive: [false, false] }));
+      });
+
+      it('should handle infinity/-infinity bounds correctly', function () {
+        expect(range.parse('(infinity,1)', 'int4range')).to.deep.equal(_.extend([Infinity, 1], { inclusive: [false, false] }));
+        expect(range.parse('(1,infinity)', 'int4range')).to.deep.equal(_.extend([1, Infinity], { inclusive: [false, false] }));
+        expect(range.parse('(-infinity,1)', 'int4range')).to.deep.equal(_.extend([-Infinity, 1], { inclusive: [false, false] }));
+        expect(range.parse('(1,-infinity)', 'int4range')).to.deep.equal(_.extend([1, -Infinity], { inclusive: [false, false] }));
+        expect(range.parse('(-infinity,infinity)', 'int4range')).to.deep.equal(_.extend([-Infinity, Infinity], { inclusive: [false, false] }));
       });
 
       it('should return raw value if not range is returned', function () {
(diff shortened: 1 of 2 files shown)


alekbarszczewski:
One thing I am not sure now if it is good idea to use javascript `Infinity` and `-Infinity` to describe +/-infinity bounds. Maybe just string `"infinity"`/`"-infinity"` would be better since it can be easily serialized/parsed to JSON (so "infinity" could be passed directly in JSON API HTTP request)? Or maybe both options?

janmeier:
Looks good @alekbarszczewski !

Regarding passing wrong data to ranges i think we should fail fast, and throw an error when the user passes something wrong (not a validation error, just a regular js `Error`)

Regarding inf - I know mick is going to say use javascript infinity, but since we already support booleans as strings and the like in other places we might as well support infinity literals and strings

test/integration/dialects/postgres/range.test.js
```diff
@@ -5,18 +5,28 @@ var chai    = require('chai')
   , Support = require(__dirname + '/../../support')
   , DataTypes = require(__dirname + '/../../../../lib/data-types')
   , dialect = Support.getTestDialect()
-  , range   = require('../../../../lib/dialects/postgres/range');
+  , range   = require('../../../../lib/dialects/postgres/range')
+  , _       = require(__dirname + '/../../support').Sequelize.Utils._;
```
janmeier: just `require('lodash') `

alekbarszczewski:
@janmeier Thanks for feedback. I will update code as you said. 
I am still wondering about infinity. There are three situations when it will matter:

1. `Model.set('range_field', [lower_bound, upper_bound])` - can accept both javascript `+/-Infinity` or `"+/-infinity"` string as bound(s).

2. `Model.get('range_field')` - should it return `[-Infinity, Infinity]` or `["-infinity", "infinity"]`?

3. `Model.toJSON()` - cant return javascript `+/-Infinity` here...

janmeier:
I think we should convert the value to `Infinity` on set, like we do with booleans.

`Model.toJSON()` can return infinity - since it returns a plain json object, not a json string. Calling `JSON.stringify` on that and getting `null` might be confusing for users, but that's just how the spec is ...

alekbarszczewski:
@janmeier Hmm the problem is that null value already represents unbound bound (sounds stupid :).
In pg there might be range like this `(,10)` and `(-infinity,10)` and they are not the same but they are almost the same. From pg docs:

> The lower bound of a range can be omitted, meaning that all points less than the upper bound are included in the range. Likewise, if the upper bound of the range is omitted, then all points greater than the lower bound are included in the range. If both lower and upper bounds are omitted, all values of the element type are considered to be in the range.

>This is equivalent to considering that the lower bound is "minus infinity", or the upper bound is "plus infinity", respectively. But note that these infinite values are never values of the range's element type, and can never be part of the range. (So there is no such thing as an inclusive infinite bound — if you try to write one, it will automatically be converted to an exclusive bound.)

>Also, some element types have a notion of "infinity", but that is just another value so far as the range type mechanisms are concerned. For example, in timestamp ranges, [today,] means the same thing as [today,). But [today,infinity] means something different from [today,infinity) — the latter excludes the special timestamp value infinity.

>The functions lower_inf and upper_inf test for infinite lower and upper bounds of a range, respectively.
```

So after rethinking it, maybe this is the way to go:

```
model.set()               | pg literal            | after JSON.stringify()
----------------------------------------------------------------------------
[null, null]              | (,)                   | [null, null]
[-Infinity, Infinity]     | (,)                   | [null, null]
["-infinity", "infinity"] | (-infinity, infinity) | ["-infinity","infinity"]
[]                        | empty                 | []
```

**Comment by alekbarszczewski** [comment]
@janmeier @mickhansen I think I found a bug in code that was there before my PR. I am talking about this line https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/range.js#L52.

```js
return moment(value).toDate();
```

I've checked that for example `value` returned from postgres is `"2016-01-01 08:00:00-04"`.

```js
console.log(require('moment')("2016-01-01 08:00:00-04"))
{ _isAMomentObject: true,
  _i: '2016-01-01 08:00:00-04',
  _f: 'YYYY-MM-DD HH:mm:ss',
  _isUTC: false,
  _pf: 
   { empty: false,
     unusedTokens: [],
     unusedInput: [ '-04' ],
     overflow: -1,
     charsLeftOver: 3,
     nullInput: false,
     invalidMonth: null,
     invalidFormat: false,
     userInvalidated: false,
     iso: true },
  _locale: 
   { _ordinalParse: /\d{1,2}(th|st|nd|rd)/,
     ordinal: [Function],
     _abbr: 'en',
     _ordinalParseLenient: /\d{1,2}(th|st|nd|rd)|\d{1,2}/ },
  _d: Fri Jan 01 2016 08:00:00 GMT+0100 (CET) }
```

It seems that moment ignores timezone (`unusedInput: [ '-04' ]`) and it just parses date as in GMT+00 (`Fri Jan 01 2016 08:00:00 GMT+0100 (CET)`). Since postgres returns timezone offset same as is set on server (by sequelize.options.timezone for example) I guess it should be parsed using this setting. Other way setting postgres server timezone to `-04:00` for example will cause invalid results from `moment(timestampString)`. What do you think guys?

janmeier:
The date parsing thing is definitely a bug! I'm pretty sure we can just do `new Date`, js should understand the pg date format. I think I fixed something similar to this in https://github.com/sequelize/sequelize/pull/4186 actually, but no idea when that will be merged (poke @mickhansen )

janmeier:
Why should `[-Infinity, Infinity]` turn into `(,)`, I think `(-infinity, infinity)` makes more sense. It might not make sense from the json.stringify point of view, but I think if the user passes literal infinity, we can be pretty sure thats what they want :D. Other than that, it looks good

janmeier:
@alekbarszczewski Datatype system has been refactored - no docs yet though :( -  but feel free to ask if you have any questions

alekbarszczewski:
@janmeier Thanks for info, but it seems that my tests are still passing, only MariaDB tests failed (I guess they will pass after rerunning the build). Right now it works like this:

```
model.set('range', [null, null]) -> '(,)' -> model.get('range') -> [null, null] -> stringify() -> [null, null]
model.set('range', [-Infinity, Infinity]) -> '(-infinity,infinity)' -> model.get('range') -> [-Infinity, Infinity] -> stringify() -> [null, null]
```

You think we can leave it as is? It will work on server side, but user will have to handle `Infinity -> 'infinity'` stringification and parsing on his own.

janmeier:
@alekbarszczewski Yep, that looks very good to me - If you would add a changelog entry and squash, this is ready to merge :)

alekbarszczewski:
@janmeier Sorry I was on holiday for 2 weeks without internet and mobile access :). Damn, now there are some conflicts. I will fix it asap.

PR #4718, opened 2015-10-23
Aggregation query doc improvement

yogesum:
```js
attributes: [[sequelize.fn('COUNT', sequelize.col('hats')), 'no_hats']]
```
Instead of
```js
attributes: [sequelize.fn('COUNT', sequelize.col('hats')), 'no_hats']
```
in querying docs for aggregation using sequelize.fn


mickhansen:
Thank you!

Issue #4715, opened 2015-10-23
ER_SP_DOES_NOT_EXIST: FUNCTION users.AsText does not exist


zhfish:
```
mysql> select `users`.AsText(`location`) from users;
FUNCTION users.AsText does not exist
mysql> select AsText(`location`) from users;
+--------------------+
| AsText(`location`) |
+--------------------+
| POINT(10 10)       |
| POINT(10 10)       |
| POINT(10 10)       |
+--------------------+
3 rows in set (0.05 sec)
```

could i not use `users` default ?

zhfish:
select AsText(`users`.`location`) from users;
is right

janmeier:
Can you please show the actual sequelize code that produces the issue?

zhfish:
```
        yield connect.users.update({
            location: user.location
        },{
            where:{
                ID:user.ID
            }
        });
```

janmeier:
And the model definitons, and some actual data?

zhfish:
model
```
/* jshint indent: 2 */

module.exports = function(sequelize, DataTypes) {
  return sequelize.define('users', {
    ID: {
      type: 'CHAR(36)',
      allowNull: false,
      defaultValue: '',
      primaryKey: true
    },
    userName: {
      type: DataTypes.STRING,
      allowNull: true
    },
    password: {
      type: DataTypes.STRING,
      allowNull: true
    },
    location: {
      type: DataTypes.GEOMETRY,
      allowNull: true
    }
  });
};

```
data
```
+--------------------------------------+-------------+--------------------------------------------------------------+----------------------------+
| ID                                   | username    | password                                                     | AsText(`users`.`location`) |
+--------------------------------------+-------------+--------------------------------------------------------------+----------------------------+
| 09c20a66-e4c5-4ecd-b8d7-567559847827 | 13800138000 | $2a$10$c9gmMhN4L63K1rmSF4poauzELud8X6MumK./efE/I32mxmo8npeNe | POINT(10 10)               |
| 55d2d8e5-855f-4bdd-b374-b540cd8ec288 | 13800138001 | $2a$10$Vge6nXgY6jw5D4HzeuekoOimF80wSrszkyOCH7vXY2ffgK0qWgsNq | POINT(10 10)               |
[...]

PR #4725, opened 2015-10-24
fix some bug in geometry for mysql and pqsql #4715

zhfish:
it has a problem for geometry, for example  
  #4715

models :
```
module.exports = function(sequelize, DataTypes) {
  return sequelize.define('users', {
    ID: {
      type: 'CHAR(36)',
      allowNull: false,
      defaultValue: '',
      primaryKey: true
    },
    userName: {
      type: DataTypes.STRING,
      allowNull: true
    },
    location: {
      type: DataTypes.GEOMETRY,
      allowNull: true
    },
    createBy: {
      type: DataTypes.DATE,
      allowNull: true
    },
    updateBy: {
      type: DataTypes.DATE,
      allowNull: true
    }
  });
};
```
```
module.exports = function(sequelize, DataTypes) {
  return sequelize.define('invites', {
    ID: {
      type: 'CHAR(36)',
      allowNull: false,
      defaultValue: '',
      primaryKey: true
    },
    userID: {
      type: DataTypes.STRING,
      allowNull: true
    },
    mobile: {
      type: DataTypes.STRING,
      allowNull: true
    },
    createBy: {
      type: DataTypes.DATE,
      allowNull: true
    },
    updateBy: {
      type: DataTypes.DATE,
      allowNull: true
    }
  });
};
```

actions :
```
connect.invites.findOne({
    where:{
        mobile: '13800138000'
    }
    include: [{
        model: connect.users
    }]
}).then(function(result) {
   console.log(result.get());
});
```
include user's  location can't get geoJson  ,just get a object
```
connect.users.findOne({
    where:{
        ID: '1'
    }
    include: [{
        model: connect.invites
    }]
}).then(function(result) {
   console.log(result.get());
});
```
it's error, beacuse sql \`user\`.AsText(\`location\`) AS location
```
connect.users.findOne({
    where:{
        ID: '1'
    }
}).then(function(result) {
   console.log(result.get());
});
```
it's OK, 

i try fix it, it's fine now,
i'm sorry, i can't understand how to write test case.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1111,7 +1111,11 @@ var QueryGenerator = {
           attributes = include.attributes.map(function(attr) {
             var attrAs = attr,
                 verbatim = false;
-
+            
+            if (include.model && include.model._isGeometryAttribute(attr)) {
+                attr = self.geometrySelect(attr,include.as);
+            }
+              
             if (Array.isArray(attr) && attr.length === 2) {
               if (attr[0]._isSequelizeMethod) {
                 if (attr[0] instanceof Utils.literal ||
@@ -2293,8 +2297,8 @@ var QueryGenerator = {
     return value;
   },
 
-  geometrySelect: function(column) {
-    return this.quoteIdentifiers(column);
+  geometrySelect: function(column,includeAs) {
+    return [new Utils.fn('ST_AsText',[new Utils.col((includeAs?includeAs+'.':'') + column)]),column];
   }
 };
 
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -924,8 +924,8 @@ var QueryGenerator = {
     return AbstractQueryGenerator.setAutocommitQuery.call(this, value, options);
   },
 
-  geometrySelect: function(column) {
-    return this.quoteIdentifiers(column);
+  geometrySelect: function(column,includeAs) {
+    return [new Utils.fn('ST_AsText',[new Utils.col((includeAs?includeAs+'.':'') + column)]),column];
   }
 };
 
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -352,8 +352,8 @@ var QueryGenerator = {
     return 'ALTER TABLE ' + this.quoteTable(tableName) + ' DROP FOREIGN KEY ' + this.quoteIdentifier(foreignKey) + ';';
   },
 
-  geometrySelect: function(column) {
-    return 'AsText(' + this.quoteIdentifiers(column) + ') AS ' + column;
+  geometrySelect: function(column,includeAs) {
+    return [new Utils.fn('ST_AsText',[new Utils.col((includeAs?includeAs+'.':'') + column)]),column];
   }
 };
 


mickhansen:
Several tests are broken with this change.
Please do add a new test case, no other way to guard against regressions.

Sounds like this is possible an issue with using GEOMETRY and includes though?
Would be helpfull if you included the full SQL.

In any case the build needs to be fixed for us to merge this.

zhfish:
i see the test failed reason is `SequelizeDatabaseError: ER_SP_DOES_NOT_EXIST: FUNCTION sequelize_test.ST_AsText does not exist`  
how to write test case? 

`Sounds like this is possible an issue with using GEOMETRY and includes though?  `
YES.The original is error.

### normal
```
SELECT `ID`, ST_AsText(`location`) AS `location` FROM `users` AS `user`;
```
  
### in include  
```
SELECT `invites`.`ID`, `user`.`ID`,ST_AsText(`user`.`location`) AS `user.location`,  FROM `invites` AS `invites` INNER JOIN `users` AS `user` ON `invites`.`userID` = `user`.`ID` WHERE `invites`.`ID` = `0` 
```  

### has include  
```
SELECT `users`.`ID`, `invite`.`ID`,ST_AsText(`users`.`location`) AS `location`,  FROM `users` AS `users` INNER JOIN `invites` AS `invite` ON `invite`.`userID` = `user`.`ID` WHERE `users`.`ID` = `0` 
```

zhfish:
postgres need import "postgis" functions.

janmeier:
Postgis is being installed - look at line 123 https://travis-ci.org/sequelize/sequelize/jobs/87194835

But postgres is not supposed to use the `st_astext` function as far as I remember, since that returns WKT - postgres can just natively return geojson

janmeier:
I would still like to see you add some tests to this PR, before I can comment further

zhfish:
@janmeier 
maybe .. postgres can just natively return geojson
i remove postgres support ?
i promise mysql can work, i use this now

janmeier:
Yep, postgres works fine without any modifications

lib/dialects/postgres/query-generator.js
```diff
@@ -924,8 +924,8 @@ var QueryGenerator = {
     return AbstractQueryGenerator.setAutocommitQuery.call(this, value, options);
   },
 
-  geometrySelect: function(column) {
-    return this.quoteIdentifiers(column);
+  geometrySelect: function(column,includeAs) {
+    return return this.quoteIdentifiers(column);
```
janmeier: Double return :)

lib/dialects/abstract/query-generator.js
```diff
@@ -1111,7 +1111,11 @@ var QueryGenerator = {
           attributes = include.attributes.map(function(attr) {
             var attrAs = attr,
                 verbatim = false;
-
+            
+            if (include.model && include.model._isGeometryAttribute(attr)) {
+                attr = self.geometrySelect(attr,include.as);
```
janmeier: Please add spaces between the comma and the next argument - this also applies to function definitions

janmeier:
I verified that the problem exists on mysql, and this seems to fix it - time to add a couple of integration tests @zhfish  :)

zhfish:
i revert these wrong ,and recommit it, 
maybe it needn't new test case ?

janmeier:
@zhfish Yes, it needs some test cases, since this area obviously wasn't tested before. Otherwise its sure to break again at some point

janmeier:
Ping @zhfish

PR #4732, opened 2015-10-26
Change wrong comment in `project.getTasks()` example

samuelbeek:
There was a wrong comment in the `project.getTasks()` example saying that `project.getTasks({attributes: ['title']})` would retrieve articles with an id higher than 10, which was not true, I updated the comment.


janmeier:
Merged in https://github.com/sequelize/sequelize/commit/2a02ad67c0272dec1acb2210f5135a3a43d32075

Thanks !

PR #4739, opened 2015-10-27
Added support for Model.getModel({ through : { where : { ... } }})

mcarboni-redant:
I've make the code on the answer to work :

http://stackoverflow.com/questions/33351537/use-a-column-in-the-jointable-on-the-where/33362317#33362317

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -445,6 +445,12 @@ BelongsToMany.prototype.injectGetter = function(obj) {
       throughWhere = {};
       throughWhere[association.foreignKey] = instance.get(association.source.primaryKeyAttribute);
 
+      if (options.through && options.through.where)
+        throughWhere = { $and : [
+            throughWhere,
+            options.through.where
+        ]};
+
       if (through.scope) {
         _.assign(throughWhere, through.scope);
       }
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -445,16 +445,17 @@ BelongsToMany.prototype.injectGetter = function(obj) {
       throughWhere = {};
       throughWhere[association.foreignKey] = instance.get(association.source.primaryKeyAttribute);
 
+      if (through.scope) {
+        _.assign(throughWhere, through.scope);
+      }
+
+      //If a user pass a where on the options through options, make an "and" with the current throughWhere
       if (options.through && options.through.where)
         throughWhere = { $and : [
             throughWhere,
             options.through.where
         ]};
-
-      if (through.scope) {
-        _.assign(throughWhere, through.scope);
-      }
-
+        
       options.include = options.include || [];
       options.include.push({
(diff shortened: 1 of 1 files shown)


lib/associations/belongs-to-many.js
```diff
@@ -445,6 +445,12 @@ BelongsToMany.prototype.injectGetter = function(obj) {
       throughWhere = {};
       throughWhere[association.foreignKey] = instance.get(association.source.primaryKeyAttribute);
 
+      if (options.through && options.through.where)
```
janmeier: We should probably move this after  `.scope` so through can overwrite the scope if needed

janmeier:
Code LGTM, just a small comment.

We also need an integration test

mcarboni-redant:
OK, moved after the scope and writed a little comment, for the integration test at the moment i can't write one, i can pass you the test i'm using in my project

var widgetSchema = {
        name: {
            type: DataTypes.STRING
        }
    };

var Widget = sequelize.define('widget', widgetSchema);

var userSchema = {
        name: {
            type: DataTypes.STRING
        }
    };

var User = sequelize.define('user', userSchema);

var User_widget = sequelize.define('perms', {
        read: {
            type: DataTypes.BOOLEAN,
            defaultValue: false
        },
        write: {
            type: DataTypes.BOOLEAN,
            defaultValue: false
        }
    });

Widget.belongsToMany(User, { through:User_widget, foreignKey: 'widget_id' });
User.belongsToMany(Widget, { through:User_widget, foreignKey: 'user_id' });

User.create({ id: 1, name : 'test' }).then(function(user) {
   //Make the widgets 
   //..
   widget1.perms = { read : true, write : false };
   widget2.perms = { read : false, write : false };
   widget3.perms = { read : true, write : true };
   user.getWidgets({
       through : {
            where : {
                'read' : true
            }
        }
   }).then(function(widgets) {
     expect(widgets.length).equal.to(2);
  });

janmeier:
That looks fine for a test - just add it somewhere appropriate in https://github.com/sequelize/sequelize/blob/master/test/integration/associations/belongs-to-many.test.js

lib/associations/belongs-to-many.js
```diff
@@ -449,6 +449,13 @@ BelongsToMany.prototype.injectGetter = function(obj) {
         _.assign(throughWhere, through.scope);
       }
 
+      //If a user pass a where on the options through options, make an "and" with the current throughWhere
+      if (options.through && options.through.where)
+        throughWhere = { $and : [
+            throughWhere,
+            options.through.where
+        ]};
+
```
mickhansen: Please add brackets around the if statement.
Please place `$and` on its own line in the object.

No spacing after `$and` before `:`.

mcarboni-redant:
I can't setup the tests to work in my environment, so i can't be sure the tests i push works fine

luin:
+1 for this feature

felixfbecker:
Closing in favor of #5479

PR #4749, opened 2015-10-28
Update querying documentation

Nicktho:
`order` supports both ordering by an instance's associate's fields as well as any nested associate's fields. `order` also supports aliased associates too. This update adds an example of this functionality to the documentation, so that users are both aware of and how to use this feature.


mickhansen:
Thank you!

Issue #4751, opened 2015-10-28
If the first connection attempt fails, all subsequent connection attempts will fail


simon-paris:
1. Shut down database
2. Create a Sequelize instance and attempt to connect to the database. Wait for an error to be thrown.
3. Start the database.
4. Attempt to connect again. This will fail.

Tested with MySQL, but probably affects all dialects.

PR #4752, opened 2015-10-28
Fix: If the first connection attempt fails, all subsequent connection attempts will fail

simon-paris:
#4751


lib/dialects/abstract/connection-manager.js
```diff
@@ -203,6 +204,9 @@ ConnectionManager.prototype.getConnection = function(options) {
 
           self.$disconnect(connection);
         });
+      }, function (err) {
+        self.versionPromise = null;
+        throw err;
```
mickhansen: Please use `catch` rather than the second then callback, otherwise looks good.

janmeier:
Rebased and merged in https://github.com/sequelize/sequelize/commit/940f8fb66f3241acb66289b00a3232ddf0a10e06

Thanks!

Issue #4747, opened 2015-10-28
Improve Unspported Dialect Error


alexbooker:
When instantiating Sequelize, I mistakenly specified the `dialect` value to be _"sqlite3"_ instead of _"sqlite"_ and got this error:

    Error: The dialect sqlite3 is not supported. (TypeError: Dialect is not a function)

I think this is is an easy mistake to make because the npm module for sqlite is called _"sqlite3"_.

It took me a couple of minutes to notice my error. 

I am yet to look at the code but I *imagine* the supported dialects are hard coded somewhere. 

What might have helped me realize my error quicker is a more descriptive error:

    Error: The dialect sqlite3 is not supported. Supported dialects: sqlite, mysql, etc. (TypeError: Dialect is not a function)

What do you think?

mickhansen:
Sounds good, PR welcome.

alexbooker:
I'll try my hand at this.

PR #4765, opened 2015-10-28
Made unsupported dialect error message clearer.

alexbooker:
In response to #4747.

I considered re-writing the code thus that the supported dialects were all defined in one place (e.g in an array), but ultimately decided to [_KISS_](https://en.wikipedia.org/wiki/KISS_principle).

I ran Mocha on the effected file. Something so simple shouldn't introduce a regression but let's what Travis has to say :smirk:


mickhansen:
Thank you :)

alexbooker:
Thanks for merging

I :heart: how quick you and @janmeier respond

PR #4767, opened 2015-10-28
Comments on columns postgres dialect

mwain:
Adds column commenting for createTable and addColumn for postgres dialect.

Example usage:
```javascript
queryInterface.createTable(
    'test',
    {
        id: {
            type: Sequelize.INTEGER,
            autoIncrement: true,
            primaryKey: true,
            comment: 'Unique identifier'
        }
    },
    {
        schema: 'playground'
    }
);

queryInterface.addColumn(
    {tableName: 'test', schema: 'playground'},
    'testColumn',
    {
        type: Sequelize.STRING,
        allowNull: false,
        comment: 'testing comments'
    }
);
```

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -48,18 +48,17 @@ var QueryGenerator = {
 
     var query = 'CREATE TABLE IF NOT EXISTS <%= table %> (<%= attributes%>)<%= comments %>'
       , comments = ''
-      , attrStr = []
-      , i;
+      , attrStr = [];
 
     if (options.comment && Utils._.isString(options.comment)) {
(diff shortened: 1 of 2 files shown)


mwain:
Not sure why travis-ci is failing, doesn't seem to be related to my pull.

Boelensman1:
Would really like to have this functionality. When I test it locally it seems to work.

mickhansen:
I don't recall exactly why we didn't follow up on this but looking at it now it definitely needs some refactoring so it doesn't operate on a SQL string.

Boelensman1:
I wouldn't mind doing that. Any tips on where to start?

mickhansen:
With the addColumn query generator, remember to have adequate unit testing. The problem was generally that the references code clashed with comments.

PR #4769, opened 2015-10-28
[docs] Fixed beforeBulkUpdate explanation.

adam-back:
Docs for `beforeBulkUpdate` read:
"A hook that is run **after** updating instances in bulk"

It looks like this line was a simple, accidental copy-and-paste from it's sibling, `afterBulkUpdate`.

This commit changes `beforeBulkUpdate` to read correctly:
"A hook that is run **before** updating instances in bulk"

Thanks for your work on Sequelize!

### DIFF
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -334,7 +334,7 @@ Hooks.hasHooks = Hooks.hasHook;
  */
 
 /**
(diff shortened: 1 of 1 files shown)


mickhansen:
Thank you @adam-back  :)

Issue #4788, opened 2015-10-31
Migrations "changeColumn" Documentation Formatting Is Broken


alexbooker:
Just a little bug I noticed on [this page](http://docs.sequelizejs.com/en/latest/docs/migrations/) for the `changeColumn` documentation:

![](https://i.imgur.com/1DtUYiq.png)

janmeier:
Thanks!

PR #4789, opened 2015-10-31
Possible fix for issue #4788

alexbooker:


janmeier:
Lets see if that fixes it :)

PR #4791, opened 2015-11-01
io.js has been merged into node.js so it doesn't need to be mentioned separately anymore

curiousdannii:
Also alter the Travis config: switch from io.js v1,2 to node.js v4.2 (LTS) and 'node' (currently v5)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: node_js
 
 node_js:
-  - "0.10"
+  - "node"
+  - "4.2"
   - "0.12"
-  - "iojs-v1"
-  - "iojs-v2"
+  - "0.10"
 
 sudo: false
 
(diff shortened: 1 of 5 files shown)


curiousdannii:
Travis doesn't seem to want to play nicely with node-gyp... I've never used it though so I have no idea what would be involved to fix it, but it really should be testing node 4.2 and 5 now.

mickhansen:
node-pg does not currently support 4.x - believe it has something to do with an nvm bug.
there's an existing PR here on GH about 4.0 support, theres some relevant links

curiousdannii:
The `pg` package? I have no trouble installing that on both 4.2 and 5, but that's just straight installing, not recompiling with node-gyp.

mickhansen:
@curiousdannii pg with pg-native

PR #4822, opened 2015-11-06
Handle GEOMETRY columns with include option

alxarch:
#4715

### DIFF
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -353,7 +353,11 @@ var QueryGenerator = {
   },
 
   geometrySelect: function(column) {
-    return 'AsText(' + this.quoteIdentifiers(column) + ') AS ' + column;
+    return {
+      as: column,
+      fn: 'AsText',
+      cols: [column]
+    };
   }
 };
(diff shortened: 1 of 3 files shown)


lib/dialects/mysql/query-generator.js
```diff
@@ -353,7 +353,11 @@ var QueryGenerator = {
   },
 
   geometrySelect: function(column) {
-    return 'AsText(' + this.quoteIdentifiers(column) + ') AS ' + column;
+    return {
+      as: column,
+      fn: 'AsText',
+      cols: [column]
+    };
```
mickhansen: This looks like completely new syntax?
Would `return [sequelize.fn('AsText', sequelize.col(column)), column]` not accomplish the same thing?

alxarch: Where would i get the Sequelize instance inside the abstract query generator?

alxarch:
This is a temporary fix. As it seems that the issue will be addressed in a different manner by #4186

janmeier:
Handled in https://github.com/sequelize/sequelize/issues/4834

PR #4824, opened 2015-11-07
fix the example at the docs home page

subjectix:
The previous code was calling then on the promise returned by `sync()` which obviously doesn't resolve into a `jane` instance!

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -24,10 +24,10 @@ sequelize.sync().then(function() {
   return User.create({
     username: 'janedoe',
     birthday: new Date(1980, 6, 20)
-  });
-}).then(function(jane) {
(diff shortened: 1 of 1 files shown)


janmeier:
Woops, good catch :D

subjectix:
thanks :grin:

mickhansen:
Actually the previous code worked as well. The result of the outer chain is the result of any inner calls.

subjectix:
@mickhansen Yeah you are right :+1:, since the callback returns a promise, the outer promise resolves to what the inner promise resolves to, I kinda didn't pay attention to the return. But i would guess it's clearer this way anyway: *you use what you create*.

PR #4839, opened 2015-11-11
prevent sequelize.sync from dropping tables

cglantschnig:
Currently if you add do
```javascript
sequelize.sync({
    force: undefined
});
```
all tables will get dropped. By default there should be a fallback to `false` if there is a wrong value send.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -924,7 +924,8 @@ Sequelize.prototype.sync = function(options) {
   options = options || {};
   options.hooks = options.hooks === undefined ? true : !!options.hooks;
   options.logging = options.logging === undefined ? false : options.logging;
-  options = Utils._.defaults(options, this.options.sync, this.options);
+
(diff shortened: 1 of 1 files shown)


mickhansen:
The check here https://github.com/sequelize/sequelize/blob/master/lib/model.js#L973 is truthy, `undefined` should not trigger a drop.

lib/sequelize.js
```diff
@@ -924,7 +924,7 @@ Sequelize.prototype.sync = function(options) {
   options = options || {};
   options.hooks = options.hooks === undefined ? true : !!options.hooks;
   options.logging = options.logging === undefined ? false : options.logging;
-  options = Utils._.defaults(options, this.options.sync, this.options);
+  options = Utils._.defaults(options, this.options.sync, this.options, { sync: false });
```
mickhansen: You mean `force` right?

cglantschnig:
I made a little test project to check how `sync({ force: undefined })` works and you are right, it works correctly. The issue I had was somewhere else.

Issue #4832, opened 2015-11-10
addAssociation(null) should throw


xizhao:
http://sequelize.readthedocs.org/en/latest/api/associations/belongs-to-many/#addassociationnewassociation-options-promise

To my understanding addAssociation only applies to one, and setting "null" should not clear all associations.

mickhansen:
calling `addAssociation(null)` should throw actually, not a legal call.

xizhao:
what about:

"Associate several instances with this." and newAssociation being an array?

addAssociation docs are identical to addAssociation***s***

xizhao:
Also what about the case of adding an association where the foreign key is null? For instance you may want to create an instance on the through table but not have the fk pointing to anything yet.

mickhansen:
`addAssociation/addAssociations` should both throw if the value is simply `null`.
The instance being added is not expected to have a foreign key yet, so that's fine.

PR #4841, opened 2015-11-12
Fix addAssociations() API docs: null argument does nothing

sigriston:
In the API docs for associations, the `addAssociation()` and `addAssociations()` are incorrectly documented: it is stated that passing `null` as the association argument will remove associations from the entity / object, whereas in fact it will do nothing (i.e., the call will be a no-op).

This is probably due to a cut & paste bug from the `setAssociations()` methods.

### DIFF
--- a/lib/associations/has-many.js
+++ b/lib/associations/has-many.js
@@ -115,19 +115,19 @@ var HasMany = function(source, target, options) {
     /**
      * Associate several instances with this.
      *
-     * @param {Array<Instance|String|Number>} [newAssociations] An array of instances or primary key of instances to associate with this. Pass `null` or `undefined` to remove all associations.
+     * @param {Array<Instance|String|Number>} [newAssociations] An array of instances or primary key of instances to associate with this.
      * @param {Object} [options] Options passed to `target.update`.
-     * @param {Object} [options.validate] Run validation for the join model
+     * @param {Object} [options.validate] Run validation for the join model.
      * @return {Promise}
      * @method addAssociations
      */
     addMultiple: 'add' + plural,
     /**
-     * Associate several instances with this.
+     * Associate one instance with this.
      *
-     * @param {Instance|String|Number} [newAssociation] An array of instances or primary key of instances to associate with this. Pass `null` or `undefined` to remove all associations.
+     * @param {Instance|String|Number} [newAssociation] An instance or primary key of instance to associate with this.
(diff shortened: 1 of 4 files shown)


mickhansen:
According to this: https://github.com/sequelize/sequelize/issues/4832 it will actually delete them all.
But yes it SHOULD noop.

lib/associations/has-many.js
```diff
@@ -115,19 +115,19 @@ var HasMany = function(source, target, options) {
...
      * @param {Object} [options] Options passed to `target.update`.
-     * @param {Object} [options.validate] Run validation for the join model
+     * @param {Object} [options.validate] Run validation for the join model.
      * @return {Promise}
      * @method addAssociations
      */
     addMultiple: 'add' + plural,
     /**
-     * Associate several instances with this.
+     * Associate one instance with this.
```
mickhansen: addMultiple does support several instances

sigriston: Hi @mickhansen, I believe this particular JSDoc belongs to `add`, not `addMultiple`.
Do you actually mean that the singular `add` will add multiple instances as well?

mickhansen: You are right, my bad.

sigriston:
Hi @mickhansen!

Thank you so much for taking a look and for pointing out #4832 to me.

However, it looks like @xizhao in that thread reached the understanding that `addAssociation()` and `addAssociations()` would clear all associations by simply reading the (IMHO incorrect) documentation. In other words: he didn't try it out.

I just did in the following gist: https://gist.github.com/sigriston/336e1a7878143aa74415
And indeed the calls with `null` will not remove all associations.

I had previously interpreted this from the following lines of code, but I had to try it out with tests, just to make sure:
- [`belongs-to-many.js`, line 633](https://github.com/sequelize/sequelize/blob/2b4195cb34249d179f4764288bb3373e30bf22cb/lib/associations/belongs-to-many.js#L633)
- [`has-many.js`, line 460](https://github.com/sequelize/sequelize/blob/2b4195cb34249d179f4764288bb3373e30bf22cb/lib/associations/has-many.js#L460)

Feel free to correct me if I'm wrong, or to let me know if you want the tests in the gist to be added to the Sequelize test suite.

Thanks! :-)

mickhansen:
Thanks for doing the detective work on that @sigriston :)

Issue #4843, opened 2015-11-12
Sequelize.DATE does not work with zero-date


yonjah:
mysql/mariadb dates allow the use of zero value in dates '0000-00-00 00:00:00'
Yet using this dates with Sequelize 3 causes an error when trying to update them to a proper date

``` sql
CREATE TABLE `test` (
  `key` char(255) NOT NULL,
  `ttl` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00',
  KEY `key` (`key`)
);
```

``` javascript
var Test = sequelize.define('test', {
            key: {
                type: Sequelize.STRING,
                primaryKey: true
            },
            ttl: {
                type     : Sequelize.DATE,
                defaultValue: '0000-00-00 00:00:00'
            //or defaultValue: new Date('0000-00-00 00:00:00')
            //or no default value all produce the same result
            },
        }, {
            timestamps: false,
            freezeTableName: true
        });

Test.create({key: 'KEY_' + process.pid})
    .then(function (row) {
        return Test.findById(row.key);
    }).then(function (row) {
        return row.set({'ttl': new Date()}).save();
    }).finally(sequelize.close.bind(sequelize));
```

Running this code  I get the following error -

``` javascript
Unhandled rejection TypeError: originalValue.getTime is not a function
    at Instance.set (node_modules/sequelize/lib/instance.js:348:68)
    at Instance.set (node_modules/sequelize/lib/instance.js:292:16)
    at null.<anonymous> (test.js:33:14)
    at tryCatcher (node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (node_modules/bluebird/js/main/promise.js:507:31)
    at Promise._settlePromiseAt (node_modules/bluebird/js/main/promise.js:581:18)
    at Async._drainQueue (node_modules/bluebird/js/main/async.js:128:12)
    at Async._drainQueues (node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (node_modules/bluebird/js/main/async.js:15:14)
    at processImmediate [as _immediateCallback] (timers.js:383:17)
```

Reggino:
+1

PR #4847, opened 2015-11-12
Make Instance.prototype.set safer for Date type. sequelize/sequelize#4843

vhain:
https://github.com/sequelize/sequelize/issues/4843

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -344,8 +344,11 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
           if (this.Model._hasDateAttributes && this.Model._isDateAttribute(key) && !!value && !value._isSequelizeMethod) {
             if (!(value instanceof Date)) {
               value = new Date(value);
+              if (isNaN(value.getTime())) {
+                value = '0000-00-00 00:00:00';
+              }
             }
-            if (originalValue && value.getTime() === originalValue.getTime()) {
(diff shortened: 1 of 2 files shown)


mickhansen:
Can you add some tests aswell?

Reggino:
Thanks, this fixed the issue for me

yonjah:
@vhain Thanks for this fix.
There is one thing I'm not sure about. What will happen if `originalValue` is set to some date but the new `value` is  '0000-00-00 00:00:00' ?  it seems like on line 348 the value would be set back to  '0000-00-00 00:00:00' but then on line 351 `value.getTime` will be called (which will cause an error).
I'm not sure if this is actually the case since I haven't had a chance to run the code so this is only what I could deduce from reading the commit. 
If this is actually the case I would suggest changing the next if to 
```javascript
if (originalValue === value || ((originalValue instanceof Date) && (value instanceof Date) && _.isEqual(value.getTime(), originalValue.getTime()))) {
	return;
}
```

This has a few extra changes - 
if both `originalValue` and `value` are  '0000-00-00 00:00:00' (or any identical non object value) the set will be ignored
if both `originalValue.getTime` and `value.getTime` return the same value including NaN the set will be ignored

If the original behavior (ignoring identical '0000-00-00 00:00:00' or NaN values) is needed you can instead just add the check for `value` being instance of Date to the original condition

vhain:
@yonjah thanks for your advice. I just pushed changes.

@mickhansen sorry for late reply. I just pushed test.

vhain:
similar fix was already merged https://github.com/sequelize/sequelize/pull/4945

PR #4848, opened 2015-11-12
Fix ambiguous identifier in order by

ackerdev:
This patch fixes an ambiguous column reference error in MSSQL queries with a limit.

### DIFF
--- a/test/integration/include/find.test.js
+++ b/test/integration/include/find.test.js
@@ -172,6 +172,30 @@ describe(Support.getTestDialectTeaser('Include'), function() {
         });
     });
 
+    it('should include a model when the PK field name and foreign key name match', function() {
+      var A = this.sequelize.define('a', {
+        id: {
+          type: DataTypes.INTEGER,
+          field: 'a_id',
+          primaryKey: true
+        }
+      })
+      , B = this.sequelize.define('b', {});
+
+      A.hasOne(B);
+      B.belongsTo(A);
+
+      return this.sequelize
+        .sync({force: true})
+        .then(function() {
+          return A.create({ B: {} }, { include: [B] });
+        })
+        .then(function(a) {
+          expect(a).to.not.equal(null);
+          expect(a.get('b')).to.not.equal(null);
+        });
+    });
+
 
     it('should still pull the main record when an included model is not required and has where restrictions without matches', function() {
       var A = this.sequelize.define('a', {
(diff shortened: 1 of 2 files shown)


mickhansen:
@ackerdev Any chance of a test case?

ackerdev:
I gave it a shot, but ran into a few issues I no longer have time to work through. Some dialects are, I think, complaining about the foreign key constraints on B having a null `a_id`.

This is what I had tried last.

```js
it('should include a model when the PK field name and foreign key name match', function() {
  var A = this.sequelize.define('a', {
    id: {
      type: DataTypes.INTEGER,
      field: 'a_id',
      primaryKey: true
    }
  })
  , B = this.sequelize.define('b', {});

  A.hasOne(B);
  B.belongsTo(A);

  return this.sequelize
    .sync({force: true})
    .then(function() {
      return A.create({ B: {} }, { include: [B] });
    })
    .then(function(a) {
      expect(a).to.not.equal(null);
      expect(a.get('b')).to.not.equal(null);
    });
});
```

test/integration/include/find.test.js
```diff
@@ -172,6 +172,30 @@ describe(Support.getTestDialectTeaser('Include'), function() {
...
+      })
+      , B = this.sequelize.define('b', {});
+
+      A.hasOne(B);
+      B.belongsTo(A);
+
+      return this.sequelize
+        .sync({force: true})
+        .then(function() {
+          return A.create({ B: {} }, { include: [B] });
```
janmeier: How is this related to order and limit? Nested creation doesn't use any selects as far as I know

janmeier:
Ping @ackerdev  is this still relevant? Not sure i understand the test case

ackerdev:
The test case is incomplete; I was quite swamped with work and wasn't able to run the test suite in a local environment because mssql is a pain in the butt. I wasn't able to return to it soon enough to give it another shot. It's also been quite a while since reporting this and I haven't been working directly with Sequelize for a while so forgive me if I'm a bit foggy on details.

Without being able to currently test this on the latest version of Sequelize, the lack of any changes to this logic in the `addLimitAndOffset` function of the MSSQL dialect in master lead me to believe that this must certainly still be an issue.

The issue was that when ORDER BY statements are generated in MSSQL dialect they lack a reference to the table, such that `ORDER BY [column]` where [column] can potentially exist in both tables. This was, in my case, caused by legacy tables that used `table_id` columns in both the primary and foreign keys for each table of the relationship, so `ORDER BY [table_id]` could refer to both tables, generating an ambiguous reference error.

AckerApple:
Hey Brother! I didn't see you made this pull request. I made a pull request for this as well and my fix is slightly different. I was also able to create a test case. If my pull request is accepted. We can close this pull request as well.

Here is my pull request for the same thing:
https://github.com/sequelize/sequelize/pull/6181

PR #4851, opened 2015-11-12
Postgres IS NULL bug demonstration

GabeIsman:
This is a failing unit test for the postgres dialect that demonstrates that the query:

```
{
   where: {
     field: undefined
  }
}
```

Produces the unexpected output of `field = NULL`. This is contradicts postgres's [recommendation](http://www.postgresql.org/docs/8.3/interactive/functions-comparison.html) that such queries never be used. It also breaks with intuition since `field: null`, produces the different (correct) query `field IS NULL`.

### DIFF
--- a/test/integration/dialects/postgres/query-generator.test.js
+++ b/test/integration/dialects/postgres/query-generator.test.js
@@ -444,6 +444,18 @@ if (dialect.match(/^postgres/)) {
           expectation: "SELECT * FROM \"myTable\" WHERE \"myTable\".\"field\" = E'\\\\x53657175656c697a65';",
           context: QueryGenerator
         },
+        {
+          title: 'use IS if field is null',
+          arguments: ['myTable', {where: {field: null}}],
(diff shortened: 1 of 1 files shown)


mickhansen:
Hmm, question is what should actually happen here.
I'm leaning towards the key being simple ignored if undefined, or maybe throw an error.

GabeIsman:
I would lean towards error over simply ignoring. My intuition is that many clients don't carefully distinguish between null's and undefined's, and if you want to draw a distinction in behavior, best to do so explicitly with an error.

That being said, I think my favorite option is just treating undefined the same as null. Undefined in JS means "not set", same as NULL in SQL.

GabeIsman:
I would lean towards error over simply ignoring. My intuition is that many clients don't carefully distinguish between null's and undefined's, and if you want to draw a distinction in behavior, best to do so explicitly with an error.

That being said, I think my favorite option is just treating undefined the same as null. Undefined in JS means "not set", same as NULL in SQL.

PR #4853, opened 2015-11-12
feat(datatypes) Support fractional DATETIME columns in MySQL 5.6.4+, fi…

### LINKED ISSUE #4205: Sequelize ignores milliseconds part of DATETIME in MySQL
I'm using MySQL database with some columns of type DATETIME(3) (datetime with precision 3). Also in the entity model I have defined type as DATE(3). But after I set property of instance to time with millisecond part (e.g. entity.openedAt: "2015-08-29T08:39:08.12345Z"), sequelize generates SQL query, that truncates the millisecond part (...`openedAt`='2015-08-29 08:39:08'...).  

> MySQL 5.6.4 and up expands fractional seconds support for TIME, DATETIME, and TIMESTAMP values, with up to microseconds (6 digits) precision.

sobotklp:
fixes #4205

### DIFF
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -410,7 +410,16 @@ TIME.prototype.toSql = function() {
  * A datetime column
  * @property DATE
  */
-var DATE = ABSTRACT.inherits();
+var DATE = ABSTRACT.inherits(function (length) {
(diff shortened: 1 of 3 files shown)


sobotklp:
I'll fix the build error and resubmit

PR #4860, opened 2015-11-13
Change behaviour for "$notIn empty array" (fixes #4859)

### LINKED ISSUE #4859: $notIn empty array
Hi!

`where: { id: { $notIn: [] } }` produces `WHERE id NOT IN (NULL)` which returns nothing. Shouldn't that Sequelize query logically return every row? :)

Tested with 3.9 and 3.13.

Znarkus:
Makes `where: { id: { $notIn: [] } }` return all rows instead of none.


mickhansen:
We've discussed this one before. It's definitely a breaking chance, although i'd agree that your fix is probably the right way to go.

/cc @janmeier

janmeier:
Yeah I agree, this is the right way to go

PR #4862, opened 2015-11-14
Return null to tell Bluebird non-return is intentional

faceleg:
Another fix to prevent warnings like http://bluebirdjs.com/docs/warning-explanations.html#warning-a-promise-was-created-in-a-handler-but-none-were-returned-from-it

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -97,6 +97,7 @@ ConnectionManager.prototype.initPools = function () {
       },
       destroy: function(connection) {
         self.$disconnect(connection);
+        return null;
       },
       max: config.pool.max,
       min: config.pool.min,
@@ -174,6 +175,7 @@ ConnectionManager.prototype.initPools = function () {
       },
       destroy: function(connection) {
         self.$disconnect(connection);
+        return null;
       },
       validate: config.pool.validate,
       max: config.pool.max,
@@ -191,6 +193,7 @@ ConnectionManager.prototype.initPools = function () {
       },
       destroy: function(connection) {
         self.$disconnect(connection);
+        return null;
       },
       validate: config.pool.validate,
       max: config.pool.max,
@@ -221,6 +224,7 @@ ConnectionManager.prototype.getConnection = function(options) {
           self.versionPromise = null;
 
           self.$disconnect(connection);
+          return null;
         });
       }).catch(function (err) {
         self.versionPromise = null;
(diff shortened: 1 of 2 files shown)


faceleg:
I'll keep adding commits until I stop noticing these warnings, then squash the commits together.

mickhansen:
@faceleg progress here? :)

faceleg:
No I've been working on the frontend portions of my project.

I haven't been able to nail down the "disconnect" warning, might need to ask for some bluebird devs to have a look at this one (by opening an issue on bluebird's repo).

Side note - I updated my tests to create / delete resources they use instead of running fixtures each time - MUCH FASTER.

alexturek:
Any chance of squashing + merging these for now? We just updated sequelize and now are getting a ton of incorrect warnings in our logs; this change would at least help mitigate that.

faceleg:
I'll leave the decision to @mickhansen

faceleg:
Yay! I'm stumped on the connection.disconnect warning as I said.

nennad:
Hi there,

Any chance this is going to be published on npm soon?
We're getting quite a lot of warnings after the last update.

Cheers

faceleg:
Pinging @mickhansen about a possible release.

@nennad please note this PR will not remove **all** the warnings - I'm only a mere human and can't divine the cause of some of the warnings.

geoffreak:
Has there been any more progress in removing these warnings?

faceleg:
I haven't had a chance to work on it no.

faceleg:
Double and triple check it's not your code, 9 times out of 10 I found the error was in my code. 
If it's not your code, find the offending `then` in Sequelize, return null and submit a PR.

geoffreak:
What should I be looking for?

faceleg:
Anything like:

```
somePromise
.then(function(value) {
  doSomethingWithValue(value);
  // Bluebird is warning you that you are not returning from this then block, which causes bugs, see next example
  // return null; <-- this is what bluebird wants - it wants you to explicitly tell it you know you are not returning anything from this then block
});
```

```
somePromise
.then(function(value) {
  doSomethingWithValue(value);
})
.then(function(modifiedValue) {
  console.log(modifiedValue); // outputs "null" as nothing was returned in the previous then block
});
```

geoffreak:
Hmm. Either it's not in my code or I'm missing something because I'm using `yield` on the `findAll` call to wait on the promise so I have no control over what gets returned since that's all native. Calls to `find`, `findById`, or `sequelize.query` don't seem to be throwing this warning despite all being used the same way, which is what lead me to believe there was something up with Sequelize.

faceleg:
Good investigation. I would now proceed to go through the `findAll` code to see if there are any promises not being handled they way BB wants

PR #4888, opened 2015-11-19
Fix #4881 and add tests for es6 imports

### LINKED ISSUE #4881: New babel update breaks sequelize imports
The recent babel 6 update caused models written in es6 to fail to import. The reason being is because Babel killed support for CommonJS's default export behavior (default export behavior is if the default export is the only export in a file, `module.exports = exports["default"];`). Since babel doesn't do this anymore, the default export is only available as `require('modulename').default`. 

Since sequelize does a `require` [when you import a model](https://github.com/sequelize/sequelize/blob/master/lib/sequelize.js#L640-L641), it is impossible to import a model transformed by babel without a hack (adding `module.exports = exports["default"];` to the end of your model).

Is there any way sequelize can check for `require(path).default` for the `defineCall` function? Something like `defineCall = defineCall.default || defineCall;` or `if (typeof defineCall === 'object' && defineCall.default) { defineCall = defineCall.default; }`?

alexschneider:


changelog.md
```diff
@@ -1,3 +1,6 @@
+# 3.14.1
```
mickhansen: Best to leave the version number out and just put `next` untill we actually cut it :)
Thanks you for adding changelog though!

changelog.md
```diff
@@ -1,3 +1,6 @@
...
+- [FIXED] Add support for Babel/ES6 Modules [#4881](https://github.com/sequelize/sequelize/issues/4881)
```
mickhansen: Change the wording to `Added support for Babel/ES6 imports`, people might confuse this with ES6 classes :)

test/integration/sequelize.test.js
```diff
@@ -1132,8 +1134,17 @@ describe(Support.getTestDialectTeaser('Sequelize'), function() {
...
+    });
 
+    it('imports a dao definition from a file compiled with babel', function () {
+      var es6project = babel.transformFileSync(__dirname + '/assets/es6project.es6', {
+        presets: ['es2015']
+      }).code;
+      fs.writeFileSync(__dirname + '/assets/es6project.js', es6project);
+      var Project = this.sequelize.import(__dirname + '/assets/es6project');
       expect(Project).to.exist;
+      fs.unlink(__dirname + '/assets/es6project.js');
```
mickhansen: This will be leftover on test failures i assume then? Maybe we should include a gitignore in assets for this file.

idris:
Ping @alexschneider can we get these code review issues resolved so it can be merged?

alexschneider:
Sorry it took so long. I'm not sure why the tests are failing - I didn't touch that part of the code. I merged it with master so it should merge cleanly though.

mickhansen:
We have intermittent failures once in a while, not related to you, rerunning it now.

mickhansen:
If you could just squash your commits to a single descriptive one this will be ready to merge.

alexschneider:
Done

Issue #4901, opened 2015-11-21
Support Sqlite3 Transaction Types: IMMEDIATE, EXCLUSIVE, RESERVED


EToreo:
Sqlite3 now supports 3 transaction types (IMMEDIATE, EXCLUSIVE, RESERVED): http://sqlite.org/lang_transaction.html

It seems to me that adding these to the options of a transaction would be desirable.

EToreo:
I took a guess at how the Sequelize team would like to see this organized.  Let me know if there are any adjustments you would like to see from that pull request.

sushantdhiman:
Transaction [types](https://github.com/sequelize/sequelize/blob/master/lib/transaction.js#L199) added in #4911

PR #4911, opened 2015-11-23
Added Transaction Type

EToreo:
Pull request for https://github.com/sequelize/sequelize/issues/4901

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -79,7 +79,7 @@ var url = require('url')
  * @param {Integer}  [options.pool.maxIdleTime] The maximum time, in milliseconds, that a connection can be idle before being released
  * @param {Function} [options.pool.validateConnection] A function that validates a connection. Called with client. The default function checks that client is an object, and that its state is not disconnected
  * @param {Boolean}  [options.quoteIdentifiers=true] Set to `false` to make table names and attributes case-insensitive on Postgres and skip double quoting of them.
- * @param {String}   [options.transactionType='DEFERRED'] Set the default transaction type. See `Sequelize.Transaction.TYPES` for possible options.
+ * @param {String}   [options.transactionType='DEFERRED'] Set the default transaction type. See `Sequelize.Transaction.TYPES` for possible options. Sqlite only.
  * @param {String}   [options.isolationLevel='REPEATABLE_READ'] Set the default transaction isolation level. See `Sequelize.Transaction.ISOLATION_LEVELS` for possible options.
  * @param {Boolean}  [options.typeValidation=false] Run built in type validators on insert and update, e.g. validate that arguments passed to integer fields are integer-like
 */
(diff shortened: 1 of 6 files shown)


janmeier:
LGTM, just needs some tests and jshint-fixes

mickhansen:
PG has a deferred mode aswell that we support in a different way, not sure if its comparable to SQLITE dererred transaction?

EToreo:
@janmeier I wasn't sure what to do about tests. Should I just check to make sure the correct syntax is output based on the different options? Seems like a lame test. 

@mickhansen PG has a few options in a similar vein, but they can be set inside the transaction, where as these must be set as part of the start of the transaction, that is why I felt it warranted a new set of options. MySQL and I think PG also support a read only mode, but I think that belongs in a different set of options for itself as well.

EToreo:
Added tests and corrected lint.

EToreo:
Can I get this pulled in or suggestions on improvements before a pull is granted?

mickhansen:
I'm a bit on the fence on branching out the options so much, if PG supports DEFERRED aswell we could look to provide a standardized syntax. 

PG has INITIALLY_DEFERRED/INITIALLY_IMMEDIATE, would that map to your DEFERRED/IMMEDIATE here?

EToreo:
@mickhansen I can see where you are coming from.  But here is where I was coming from... PG calls out their options as isolation levels AND allows that setting to be changed after a BEGIN statement, while Sqlite calls their setting a behavior (and can only be set with the BEGIN statement).

If all these are really isolation levels, then let's wrap them all up.  If not, I like the current implementation.

Thoughts?

mickhansen:
@EToreo No idea if they are actually equal.
thoughts @janmeier?

I suppose having a type of transaction makes sense, even for PG (we could later support types for PG aswell with the INITIALLY_DEFERRED)

EToreo:
@janmeier I know this is a busy time of year, but have you had a chance to look at this and chime in?

I'm looking to get it pulled in for a project I am working on and this is becoming gating.

janmeier:
I think there is a difference here, since the pg deferred stuff can also be limited to specific tables etc - I like your implementation here, even though its sqlite specific

EToreo:
Wonderful.  Let me know if there is anything more I can do to get this pulled in.

test/integration/transaction.test.js
```diff
@@ -239,6 +239,32 @@ describe(Support.getTestDialectTeaser('Transaction'), function() {
...
+  if (current.dialect.supports.transactionOptions.type) {
+    describe('transaction types', function() {
+      it('should support default transaction type DEFERRED', function() {
+        return this.sequelize.transaction({
+        }).bind(this).then(function (t) {
+          return t.rollback().bind(this).then(function() {
+            expect(t.options.type).to.equal('DEFERRED');
+          });
+        });
+      });
```
mickhansen: `DEFERRED` is the default for sqlite if none is defined?

EToreo: Yep. 
> Transactions can be deferred, immediate, or exclusive. The default transaction behavior is deferred.

https://www.sqlite.org/lang_transaction.html

lib/sequelize.js
```diff
@@ -1200,6 +1202,7 @@ Sequelize.where = Sequelize.condition = Sequelize.prototype.condition = Sequeliz
 
  * @param {Object}   [options={}]
  * @param {Boolean}  [options.autocommit=true]
+ * @param {String}   [options.type='DEFERRED'] See `Sequelize.Transaction.TYPES` for possible options
```
mickhansen: Might want to add a sqlite only notice here and the other place you mention it in the docs.

mickhansen:
Just a single comment then i think we're good to go

EToreo:
Good suggestion.  Notes added.

EToreo:
@mickhansen  @janmeier Thanks guys.
Can you give your thoughts and input on https://github.com/sequelize/sequelize/pull/4969 now that this perquisite has been satisfied?

PR #4919, opened 2015-11-24
[MSSQL] Fix connection manager domain usage

### LINKED ISSUE #4918: MSSQL: unable to authenticate to a domain using NTLM
Using the following Sequelize configuration, we were getting strange authentication error patterns.
```javascript
{ 
  userName: 'user-xxx',
  password: 'XXXXXX',
  server: 'MSSQL.XXXXX.LOCAL',
  port: '1433',
  database: 'db_xxxxx',
  dialectOptions: {
    domain: 'XXXXX',
    tdsVersion: '7_3_A',
    encrypt: false
  }
}
```

This is the config passed to the Tedious connection constructor by the Sequelize connectionManager:

```javascript
{ userName: 'user-xxx',
  password: 'XXXXXX',
  server: 'MSSQL.XXXXX.LOCAL',
  options:
   { port: '1433',
     database: 'db_xxxxx',
     tdsVersion: '7_3_A',
     encrypt: false },
  domain: 'XXXXX' }
```

Upon instrumenting the connection manager with Tedious debug listeners and working with the client to look at server logs, we saw that the first connection of a pool would succeed, and any further ones failed. Likewise, if we removed pooling and clustering, the first query would succeed, and subsequent ones would fail to authenticate to the domain using Windows Authentication:

```
SEQUELIZE 19275: [Mon Nov 23 2015 16:05:19 GMT-0200 (BRST)] Creating new connection with config:
{"userName":"user-xxx","password":"XXXXXX","server":"MSSQL.XXXXX.LOCAL","options":{"port":"1433","database":"db_xxxxx","tdsVersion":"7_3_A","encrypt":false},"domain":"XXXXX"}
SEQUELIZE 19275: Debug: connected to MSSQL.XXXXX.LOCAL:1433
SEQUELIZE 19275: Debug: State change: Connecting -> SentPrelogin
SEQUELIZE 19275: Debug: State change: SentPrelogin -> SentLogin7WithNTLMLogin
SEQUELIZE 19275: Debug: State change: SentLogin7WithNTLMLogin -> SentNTLMResponse
SEQUELIZE 19275: Debug: Packet size changed from 4096 to 4096
SEQUELIZE 19275: Debug: State change: SentNTLMResponse -> LoggedInSendingInitialSql
SEQUELIZE 19275: Debug: State change: LoggedInSendingInitialSql -> LoggedIn
SEQUELIZE 19275: Debug: State change: LoggedIn -> SentClientRequest
SEQUELIZE 19275: Debug: State change: SentClientRequest -> LoggedIn
SEQUELIZE 19275: Debug: State change: LoggedIn -> Final
SEQUELIZE 19275: Debug: connection to MSSQL.XXXXX.LOCAL:1433 closed
SEQUELIZE 19275: Debug: State is already Final

```
The next query begins like so. Note that `domain` is no longer present in `config.options`.
```
SEQUELIZE 19275: [Mon Nov 23 2015 16:05:23 GMT-0200 (BRST)] Creating new connection with config:
{"userName":"user-xxx","password":"XXXXXX","server":"MSSQL.XXXXX.LOCAL","options":{"port":"1433","database":"db_xxxxx","tdsVersion":"7_3_A","encrypt":false}}
SEQUELIZE 19275: Debug: connected to MSSQL.XXXXX.LOCAL:1433
SEQUELIZE 19275: Debug: State change: Connecting -> SentPrelogin
SEQUELIZE 19275: Debug: State change: SentPrelogin -> SentLogin7WithStandardLogin
SEQUELIZE 19275: Error: {"number":18456,"state":1,"class":14,"message":"Login failed for user 'user-xxx'.","serverName":"MSSQL","procName":"","lineNumber":1,"name":"ERROR","event":"errorMessage"}
SEQUELIZE 19275: Debug: State change: SentLogin7WithStandardLogin -> Final
SEQUELIZE 19275: Debug: connection to MSSQL.XXXXX.LOCAL:1433 closed
```

I will be submitting a pull request to fix this, but I'm putting this issue here for documentation.

lukealbao:
API for the Tedious driver requires that an entry for `domain`, if present,
be at the root level of the config option passed to the constructor. This
was added in commit 91fc76be10b12e0e5a2c4a9299e5bc05a2956d58.

However, that commit deleted `dialectOptions.domain` from the config object
passed into the Sequelize constructor. Since most config objects are singletons,
this mutation of the config object breaks any subsequent calls to the instance's
`connectionManager.$connect`.

This commit simply removes the call to delete the domain entry.
Fixes #4918

### DIFF
--- a/lib/dialects/mssql/connection-manager.js
+++ b/lib/dialects/mssql/connection-manager.js
@@ -53,7 +53,6 @@ ConnectionManager.prototype.connect = function(config) {
       // The 'tedious' driver needs domain property to be in the main Connection config object
       if(config.dialectOptions.domain) {
         connectionConfig.domain = config.dialectOptions.domain;
-        delete config.dialectOptions.domain;
       }
 
       Object.keys(config.dialectOptions).forEach(function(key) {
--- a/test/unit/dialects/mssql/connection-manager.js
+++ b/test/unit/dialects/mssql/connection-manager.js
@@ -1,0 +1,42 @@
+'use strict';
+
+var chai = require('chai')
+  , expect = chai.expect
+  , Sequelize = require(__dirname + '/../../../../index');
+
+
+var tedious = require('tedious')
+  , sinon = require('sinon')
+  , connectionStub = sinon.stub(tedious, 'Connection');
+
+connectionStub.returns({on: function () {}});
+
+describe('[MSSQL] Connection Manager', function () {
+  var instance
+    , config;
+  beforeEach(function () {
+    config = {
+      dialect: 'mssql',
+      database: 'none',
+      username: 'none',
+      password: 'none',
+      host: 'localhost',
+      port: 2433,
+      pool: {},
+      dialectOptions: {
+        domain: 'TEST.COM'
+      }
+    };
+    instance = new Sequelize(config.database
+                             , config.username
+                             , config.password
+                             , config);    
+  });
+  
+  it('connectionManager.$connect() Does not delete `domain` from config.dialectOptions',
+     function () {
+       expect(config.dialectOptions.domain).to.equal('TEST.COM');
+       instance.dialect.connectionManager.$connect(config);
+       expect(config.dialectOptions.domain).to.equal('TEST.COM');
+     });
+});


test/integration/dialects/mssql/connection-manager.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+      expect(config.dialectOptions.domain).to.equal('TEST.COM');
+      done();
+    })
+    .catch(function (err) {
+      // err MAY be a Chai AssertionError. Pass it to done.
+      done(err);
+    });
+  });
+});
```
mickhansen: No need to use done, mocha handles promises.

```js
return instance.dialect.connectionManager.$connect(config).then(function () {
  expect(config.dialectOptions.domain).to.equal('TEST.COM');
}).catch(SequelizeConnectionError, function (err) {
  expect(config.dialectOptions.domain).to.equal('TEST.COM');
});
```

But generally it's not a good idea to have branching in tests like this, why can connect both succeed and fail here? Will both paths always be tested?

lukealbao: No, it definitely won't always test both paths. I did that in case someone pointed this test at a DB server that could authenticate NTLM (not the mssql.sequelizejs.com). I just wanted to be sure that this test would always test whether the domain was deleted, but I suppose that with such a simple test, the chances of someone modifying this test (or mssql.sequelizejs.com authenticating TEST.COM) are pretty slim.

mickhansen: Might be better off simply using a unit test that stubs out the mssql library and just verifies that the options passed are maintained across multiple calls.

lukealbao: Thanks. It looks to me like stubbing out tedious would be, well, very tedious. How about we force the test to point at an unknown db on localhost, and perform the test on ConnectionRefusedError?

```javascript
expect(config.dialectOptions.domain).to.equal('TEST.COM');
       
return instance.dialect.connectionManager.$connect(config)
          .catch(SequelizeConnectionRefusedError, function (err) {
            expect(config.dialectOptions.domain).to.equal('TEST.COM');
          });
```
If that's acceptable, where should I place this test file? If not acceptable, any clues on where to insert a tedious stub?

mickhansen: Looks like you just need to stub out `Connection` https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/connection-manager.js#L64

The rest of the code is event emitter, so we just don't emit any events.

lukealbao: Excellent. Thanks, @mickhansen. Where should I put this test, and should I submit it as a separate PR instead?

mickhansen: Fine to update this PR, we'll likely need a new folder in https://github.com/sequelize/sequelize/tree/master/test/unit so i'm open to ideas :)

lukealbao:
@mickhansen I left it in `integration/dialects/mssql` since it kinda makes sense, and (I think?) I can't edit this PR and change the directory structure of the file. Please let me know if you'd like me to close this and create a new one, with – maybe – the test in `/unit/dialects/mssql/connection-manager.js`.

mickhansen:
@lukealbao Sorry for the long delay. Yes if you could simply move it to the `test/unit` folder equivelant that would be more than fine!

There's shouldn't be any issue in moving the file in this same PR. Just move the file, add it, and commit (it should simply make move file commit).

lukealbao:
All done. Thanks, @mickhansen!

mickhansen:
@lukealbao super, if you could then just squash your commits to a single descriptive one that would be awesome!

lukealbao:
Build broke for some unrelated reason — passes now. I think this one is good to close, @mickhansen. What do you think?

mickhansen:
@lukealbao looks like you got some weird commits into this PR somehow.

lukealbao:
Huh — strange. Don't have time to figure out how those got there, but dropped those commits. Thanks for your patience, @mickhansen.

PR #4927, opened 2015-11-26
don't overwrite options.foreignKey in associations

greenify:
I use sequelize quite often and just discovered are rare case where it doesn't exactly do what imho it is supposed to do. Let's assume we have three tables, where A & B have a string ID and C has an integer ID. Now when we try to connect B and C with A, we run into the problem that if a foreignKey is given as input, it will be *modified* by the associations. Hence this to this weird behavior:

```
var sequelize = new Sequelize('database', 'username', 'password', {
    dialect: 'sqlite',
    storage: 'foo.sqlite'
});

var A = sequelize.define('A', {
    id: {
        type: Sequelize.CHAR(20),
        primaryKey: true
    }
});
var B = sequelize.define('B', {
    id: {
        type: Sequelize.CHAR(20),
        primaryKey: true
    }
});
var C = sequelize.define('C', {});
var reqValidForeignKey = { foreignKey: { allowNull: false }};

//B.hasMany(A, reqValidForeignKey); // other associations also modify the options.foreignKey
A.belongsTo(B, reqValidForeignKey);
A.belongsTo(C, reqValidForeignKey);
var QueryGenerator = A.modelManager.sequelize.queryInterface.QueryGenerator;
console.log(QueryGenerator.attributesToSQL(A.rawAttributes, {
    context: 'createTable'
}));
//{ id: 'CHAR(20) PRIMARY KEY',
//  createdAt: 'DATETIME NOT NULL',
//  updatedAt: 'DATETIME NOT NULL',
//  BId: 'CHAR(20) NOT NULL REFERENCES `Bs` (`id`) ON DELETE CASCADE ON UPDATE CASCADE',
//  CId: 'CHAR(20) NOT NULL REFERENCES `Cs` (`id`) ON DELETE NO ACTION ON UPDATE CASCADE' }
```

Note: CId should be an integer.

This is due to the fact that we use the same `foreignKey` object which is passed into the function and modify. In my use case I discovered this because both tables are required to have foreign keys.

This PR adds a very simple fix for this problem and also a test case to check for this ;-)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# Next
+- [FIXED] Don't overwrite options.foreignKey in associations [#4927](https://github.com/sequelize/sequelize/pull/4927)
+
 # 3.14.1
 - [FIXED] Issue with transaction options leaking and certain queries running outside of the transaction connection.
 
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -323,8 +323,8 @@ BelongsToMany.prototype.injectAttributes = function() {
     , targetKey = this.target.rawAttributes[this.target.primaryKeyAttribute]
     , targetKeyType = targetKey.type
     , targetKeyField = targetKey.field || this.target.primaryKeyAttribute
-    , sourceAttribute = _.defaults(this.foreignKeyAttribute, { type: sourceKeyType })
-    , targetAttribute = _.defaults(this.otherKeyAttribute, { type: targetKeyType });
+    , sourceAttribute = _.defaults({}, this.foreignKeyAttribute, { type: sourceKeyType })
+    , targetAttribute = _.defaults({}, this.otherKeyAttribute, { type: targetKeyType });
 
   if (this.primaryKeyDeleted === true) {
     targetAttribute.primaryKey = sourceAttribute.primaryKey = true;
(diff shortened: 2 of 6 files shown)


test/integration/associations/dont-modify-options.test.js
```diff
@@ -0,0 +1,66 @@
...
+          type: Sequelize.CHAR(20),
+          primaryKey: true
+        }
+      });
+      this.C = this.sequelize.define('C', {});
+    });
+
+    it('should not be overwritten for belongsTo', function(){
+      var self = this;
+      return this.sequelize.sync({force: true}).then(function() {
```
janmeier: There's no need to call  `.sync` here - which means the test can be converted to a unit test

janmeier:
LGTM, just a single comment

If you change that, add a changelog entry and squash this should be good to merge

greenify:
> If you change that, add a changelog entry and squash this should be good to merge

Thanks for your help :)

Should I be worried about travis failing for MySQL?

janmeier:
Not as long as it's only a single job failing - we have an intermittent failure which we've not been able to track down yet

janmeier:
Woops, sorry if my comment was not clear enough, could you move the file to test/unit

greenify:
> Woops, sorry if my comment was not clear enough, could you move the file to test/unit

No worries - my fault ;-)

janmeier:
Thanks! :)

PR #4930, opened 2015-11-26
Fixes usage of promise antipatterns in the docs

### LINKED ISSUE #4909: Usage of promise anti-patterns in the docs
In the docs, you can see examples like this all over the place:
```js
Task.bulkCreate([
  {subject: 'programming', status: 'executing'},
  {subject: 'reading', status: 'executing'},
  {subject: 'programming', status: 'finished'}
]).then(function() {
  Task.destroy({
    where: {
      subject: 'programming'
    },
    truncate: true /* this will ignore where and truncate the table instead */
  }).then(function(affectedRows) {
    // affectedRows will be 2
    Task.findAll().then(function(tasks) {
      console.log(tasks) // no programming, just reading :(
    })
  })
})
```
This completely misses the best feature of promises and makes them no better than callbacks. You are not chaining the promises, you don't escape callback hell, and you don't propagate errors. It should really be like this:
```js
Task.bulkCreate([
  {subject: 'programming', status: 'executing'},
  {subject: 'reading', status: 'executing'},
  {subject: 'programming', status: 'finished'}
]).then(function() {
  return Task.destroy({
    where: {
      subject: 'programming'
    },
    truncate: true /* this will ignore where and truncate the table instead */
  })
}).then(function(affectedRows) {
  // affectedRows will be 2
  return Task.findAll()
}).then(function(tasks) {
  console.log(tasks) // no programming, just reading :(
}).catch(function(error) {
  console.error('bad error happened :/')
})
```
Flat structure, no callback hell, and you only need ONE catch at the end to handle errors. With arrow functions this would become even cleaner.

I think this is very important because it is a common pitfall for people coming from callback style to promises and then they just copy+paste examples from the internet and miss the point of promises.
Docs should teach people the best practice(s) to use their library, including showing how to handle errors.

felixfbecker:
Closes #4909
Some things I noticed while looking through the source:
 - There is a lot of trailing whitespace (which my editor trimmed - initially made a huge diff)
 - There are occurrences of Zalgo (sometimes throwing synchronously sometimes async)
 - `ModelValidor.prototype.validate()` _returns_ the error instead of throwing it. This means in _every_ call I encoutered there was a `if (err) throw err` handler attached. This looks a lot like it was just regexp-replaced from callback style to promise style. Maybe this behavior should be changed.
 - The docs recommend to be a modern developer and use the "newest" Node version 0.10 for contributing
 - The code would look _so_ much better with arrow functions. `Promise.bind`, `var self = this` everywhere. Maybe it's time to make the jump from Node 0.6 to 4 for the next major version?

### DIFF
--- a/docs/articles/getting-started.md
+++ b/docs/articles/getting-started.md
@@ -53,7 +53,8 @@ sequelize
   .authenticate()
   .then(function(err) {
     console.log('Connection has been established successfully.');
-  }, function (err) { 
+  })
+  .catch(function (err) {
     console.log('Unable to connect to the database:', err);
   });
 ```
--- a/docs/docs/associations.md
+++ b/docs/docs/associations.md
@@ -510,12 +510,12 @@ Project.create({ /* */ }).then(function(project) {
 // check if all associated objects are as expected:
 // let's assume we have already a project and two users
 project.setUsers([user1, user2]).then(function() {
-  return project.hasUsers([user1]).then(function(result) {
-    // result would be false
-    return project.hasUsers([user1, user2]).then(function(result) {
-      // result would be true
-    })
-  })
+  return project.hasUsers([user1]);
+}).then(function(result) {
+  // result would be false
+  return project.hasUsers([user1, user2]);
+}).then(function(result) {
+  // result would be true
 })
 ```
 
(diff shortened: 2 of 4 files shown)


mickhansen:
Thanks for picking up our slack.
Moving to v4.0 to only is being discussed, also whether or not to use babel.
There are still major production systems running older versions, but we could definitely start moving the docs towards ES6 and provide a ES6 model system atleast.

PR #4936, opened 2015-11-27
Documentation for custom UUID generation

oravecz:
Updated documentation to include mention of returning values from a function for primary keys.


janmeier:
docs/api is autogenerated - please move it to lib/data-types.js instead

PR #4939, opened 2015-11-28
Feature/pg schema deletequery

faceleg:
`Model.destroy({ truncate: true })` fails when one is using PG and non-standard schemas.

This PR fixes the issue using the good ol' `extractTableDetails` method added in one of my previous PR's.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,4 @@
-# 3.14.2
+# NEXT
  - [FIXED] Model.destroy({ truncate: true }) fails when using postgres and schemas
 
 # 3.14.2
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,7 @@
 # 3.14.2
+ - [FIXED] Model.destroy({ truncate: true }) fails when using postgres and schemas
+
+# 3.14.2
 - [FIXED] Model.aggregate methods now support attributes and where conditions with fields. [#4935](https://github.com/sequelize/sequelize/issues/4935)
 - [FIXED] Don't overwrite options.foreignKey in associations [#4927](https://github.com/sequelize/sequelize/pull/4927)
 - [FIXED] Support nested `$col` keys. [#4849](https://github.com/sequelize/sequelize/issues/4849)
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -331,7 +331,7 @@ var QueryGenerator = {
       , tableDetails = this.extractTableDetails(tableName, options);
 
     if (options.truncate === true) {
-      query = 'TRUNCATE ' + QueryGenerator.quoteIdentifier(tableDetails.schema) + tableDetails.delimiter + QueryGenerator.quoteIdentifier(tableDetails.tableName);
+      query = 'TRUNCATE ' + this.quoteTable(this.extractTableDetails(tableName, options));
 
       if (options.cascade) {
         query += ' CASCADE';
@@ -368,7 +368,7 @@ var QueryGenerator = {
     }
 
     var replacements = {
-      table: this.quoteIdentifiers(tableName),
+      table: this.quoteTable(this.extractTableDetails(tableName, options)),
       where: this.getWhereConditions(where),
       limit: !!options.limit ? ' LIMIT ' + this.escape(options.limit) : '',
       primaryKeys: primaryKeys[tableName].length > 1 ? '(' + pks + ')' : pks,
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -325,14 +325,13 @@ var QueryGenerator = {
   },
 
   deleteQuery: function(tableName, where, options, model) {
-    var query;
-
     options = options || {};
 
-    tableName = Utils.removeTicks(this.quoteTable(tableName), '"');
+    var query
+        , tableDetails = this.extractTableDetails(tableName, options);
 
     if (options.truncate === true) {
-      query = 'TRUNCATE ' + QueryGenerator.quoteIdentifier(tableName);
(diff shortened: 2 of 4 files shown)


lib/dialects/postgres/query-generator.js
```diff
@@ -325,14 +325,13 @@ var QueryGenerator = {
...
-
     options = options || {};
 
-    tableName = Utils.removeTicks(this.quoteTable(tableName), '"');
+    var query
+        , tableDetails = this.extractTableDetails(tableName, options);
 
     if (options.truncate === true) {
-      query = 'TRUNCATE ' + QueryGenerator.quoteIdentifier(tableName);
+      query = 'TRUNCATE ' + QueryGenerator.quoteIdentifier(tableDetails.schema) + tableDetails.delimiter + QueryGenerator.quoteIdentifier(tableDetails.tableName);
```
mickhansen: schema and delimiter are empty strings if not defined?

lib/dialects/postgres/query-generator.js
```diff
@@ -325,14 +325,13 @@ var QueryGenerator = {
   },
 
   deleteQuery: function(tableName, where, options, model) {
-    var query;
-
     options = options || {};
 
-    tableName = Utils.removeTicks(this.quoteTable(tableName), '"');
+    var query
+        , tableDetails = this.extractTableDetails(tableName, options);
```
mickhansen: identation issue

changelog.md
```diff
@@ -1,4 +1,7 @@
 # 3.14.2
+ - [FIXED] Model.destroy({ truncate: true }) fails when using postgres and schemas
```
mickhansen: put `Next` istead, we might add features before we cut a release :)

mickhansen:
A test so we don't have a regression would be neat. (a simple sql unit test will be fine).

lib/dialects/postgres/query-generator.js
faceleg: But but it passed the linting!

How would you like it listed?

lib/dialects/postgres/query-generator.js
faceleg: `public` and `.` respectively

lib/dialects/postgres/query-generator.js
mickhansen: Odd that it is. Simply:

```js
var query
  ,  tableDetails
```
prefixed commas with 2 spaces before, one after ;p

lib/dialects/postgres/query-generator.js
mickhansen: Ah right!

faceleg:
Working on the test

lib/dialects/postgres/query-generator.js
```diff
@@ -325,14 +325,13 @@ var QueryGenerator = {
...
-
     options = options || {};
 
-    tableName = Utils.removeTicks(this.quoteTable(tableName), '"');
+    var query
+      , tableDetails = this.extractTableDetails(tableName, options);
 
     if (options.truncate === true) {
-      query = 'TRUNCATE ' + QueryGenerator.quoteIdentifier(tableName);
+      query = 'TRUNCATE ' + QueryGenerator.quoteIdentifier(tableDetails.schema) + tableDetails.delimiter + QueryGenerator.quoteIdentifier(tableDetails.tableName);
```
mickhansen: Could this not be replaced by `quoteTable(this.extractTableDetails(...))`?

faceleg: Good spotting

faceleg:
@mickhansen until you have further demands, I'm done

test/unit/sql/delete.test.js
```diff
@@ -45,5 +45,80 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
...
+            options,
+            User
+          ), {
+            postgres: 'DELETE FROM "web"."test_user" WHERE "test_user_id" IN (SELECT "test_user_id" FROM "web"."test_user" WHERE "test_user_id" = 100 LIMIT 1)',
+            sqlite:   'DELETE FROM `test_user` WHERE `test_user_id` = 100',
+            mssql:    'DELETE TOP(1) FROM [test_user] WHERE [test_user_id] = 100; SELECT @@ROWCOUNT AS AFFECTEDROWS;',
+            default:  'DELETE FROM [test_user] WHERE [test_user_id] = 100 LIMIT 1'
+          }
+        );
+      });
```
mickhansen: We should actually be supporting schemas for all 4 dialects (especially mssql).
But if these tests pass i guess we don't, and that's not on you to fix that :)

mickhansen:
Looks good, if the tests pass you can go ahead and squash and i'll merge.

test/unit/sql/delete.test.js
faceleg: I thought so - but I can't/won't install MSSQL...

As I use PG I am more than happy to contribute PG related fixes, but I'm not able to port them to dialects I'm not familiar with :(

mickhansen: And you're not required to, i was simply noting that this test should actually fail :) (but it looks like support are lacking from those dialects as they were from PG)

faceleg:
Looks like I've some tests to update

test/unit/sql/delete.test.js
faceleg: Huh, looks like the tests **do** fail... I will do what I can when I can, but my time will be limited for the next few days

faceleg:
It seems I've broken MSSQL and SQLite. I will fix these as soon as I can but no promises regarding an ETA

janmeier:
Fixed in https://github.com/sequelize/sequelize/commit/35671ec63661b18fe60484a3c839ada49add1772

janmeier:
Thanks for your contribution though @faceleg !

faceleg:
My dream was that if I waited long enough someone else would fix it.

Dreams CAN come true!

PR #4945, opened 2015-11-29
convert original time to Date object

limianwang:
This should solve the follow error stack, by converting the original time to a Date Object.
```
<h1>originalValue.getTime is not a function</h1>
<h2></h2>
<pre>TypeError: originalValue.getTime is not a function
    at Instance.set (/node_modules/sequelize/lib/instance.js:348:68)
    at Instance.set (/node_modules/sequelize/lib/instance.js:292:16)
    at Instance.update (/node_modules/sequelize/lib/instance.js:791:8)
```

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -345,6 +345,9 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
             if (!(value instanceof Date)) {
               value = new Date(value);
             }
+            if (!(originalValue instanceof Date)) {
+              originalValue = new Date(originalValue);
+            }
             if (originalValue && value.getTime() === originalValue.getTime()) {
               return;
             }


mickhansen:
In what case is the original value not a date? That seems to be the underlying bug which is probably what we should fix.

Reggino:
Probably related to https://github.com/sequelize/sequelize/issues/4843

sa123j:
@limianwang  @mickhansen  When are you planning to merge this pull request for date fix ? I am having same issue "originalValue.getTime is not a function" when date field in db is "0000-00-00 00:00:00"

mickhansen:
@sa123j When comments have been adressed. Not sure under what circumstances this happens, while `new Date('0000-00-00 00:00:00')` does result in an invalid date it still passed a `instanceof` check and `getTime` is still defined in my tests

limianwang:
Sorry I got caught up with some work stuff. I do some investigation, and provide associated as soon as possible.

guisehn:
This would fix the issue #5046 .

mickhansen:
#5046 provides a decent case, although ideally in the future we fix this in another way (it really shouldn't be the job of the setter to format the original value)

Issue #4904, opened 2015-11-22
findAll raw where caluse with in and array replacements results in wrong query


umar-muneer:
I am using sequelize 3.14.1 with mysql and the following code fragment;

```javascript
var A = sequelize.define("A",{
	"id":{type:Sequelize.INTEGER,primaryKey:true,allowNull:false,autoIncrement:true},
	"name":Sequelize.STRING
});

var B = sequelize.define("B",{
	"id":{type:Sequelize.INTEGER,primaryKey:true,allowNull:false,autoIncrement:true},
	"aId":Sequelize.INTEGER,
	"detail":Sequelize.STRING
});
A.hasMany(B,{"foreignKey":'aId'});
return Promise.bind({}).then(function(){
	return sequelize.sync({
		"force":true
	});
}).then(function(){
	return A.create({
		"name":"test1"
	});
}).then(function(result){
	return B.bulkCreate([{"aId":result.id,"detail":"detail1"},{"aId":result.id,"detail":"detail2"}]);
}).then(function(){
	return A.findAll({
		"include":[{
			"model":B,
			"where":['aId not in (?)',[10,11]]
		}]
	});
}); 
```
the find All call results in the following query 

> SELECT `A`.`id`, `A`.`name`, `A`.`createdAt`, `A`.`updatedAt`, `Bs`.`id` AS `Bs.id`, `Bs`.`aId` AS `Bs.aId`, `Bs`.`detail` AS `Bs.detail`, `Bs`.`createdAt` AS `Bs.createdAt`, `Bs`.`updatedAt` AS `Bs.updatedAt` FROM `As` AS `A` INNER JOIN `Bs` AS `Bs` ON `A`.`id` = `Bs`.`aId` AND aId not in ([10,11]);

this query results in the following error
>You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '[10,11])' at line 1

However if i revert to 3.12 and run the same code fragment it runs without an error and generates the correct query

[...]

PR #4949, opened 2015-11-30
Fix for #4904

shoshomiga:


shoshomiga:
Don't merge this, I just looked at it again and it probably opens up the code to sql injection.

I don't have time to fix it right now so I'm going to close the PR for now.

PR #4951, opened 2015-11-30
Fix for #4904

shoshomiga:


test/integration/dialects/mysql/query-generator.test.js
```diff
@@ -369,6 +369,11 @@ if (Support.dialectIsMySQL()) {
           arguments: ['myTable', {where: {field: {ne: null}}}],
           expectation: 'SELECT * FROM `myTable` WHERE `myTable`.`field` IS NOT NULL;',
           context: QueryGenerator
+        }, {
```
janmeier: Please move the test to unit/sql

shoshomiga: I had no idea where this test should go, so I just searched for mysql since this bug is mysql/mariadb specific

Where do you want me to put the test exactly?

janmeier: in sql/where.test.js or sql/select.test.js - I know the bug is mysql specific, but we might as well add unit testing for all dialects

mickhansen: @janmeier can't be a where test, since it only happens for raw substitions is my guess.

test/unit/sql/select.test.js
```diff
@@ -232,4 +232,16 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
...
   });
+
+  suite('raw query', function () {
+    test('raw replacements', function () {
+      expectsql(sql.selectQuery('User', {
+        attributes: ['*'],
+        having: ['name IN (?)', [1, 'test', 3, "derp"]]
+      }), {
+        default: "SELECT * FROM [User] HAVING name IN (1,'test',3,'derp');",
+        postgres: "SELECT * FROM \"User\" HAVING name IN (ARRAY[1,'test',3,'derp']);"
```
shoshomiga: Is this even valid SQL for postgres?

http://sqlfiddle.com/#!15/18dca/2

shoshomiga:
I think this calls for a more complicated fix, and also seems to be an issue on all dialects and not only mysql

I'm going to be away from computers for a few days and then I'm going to take another stab at it, closing for now

test/unit/sql/select.test.js
mickhansen: No postgres should have the same syntax, or use ANY ARRAY, but generally we use IN for PG aswell.

PR #4956, opened 2015-12-01
Document Sequelize.TEXT('length')

Znarkus:


docs/docs/models-definition.md
```diff
@@ -85,6 +85,7 @@ Sequelize.STRING                      // VARCHAR(255)
 Sequelize.STRING(1234)                // VARCHAR(1234)
 Sequelize.STRING.BINARY               // VARCHAR BINARY
 Sequelize.TEXT                        // TEXT
+Sequelize.TEXT('length')              // TEXT
```
janmeier: I think it would be more readably if you give an example of what `'length'` could be ('tiny', 'medium', and 'long')

janmeier:
Added in https://github.com/sequelize/sequelize/commit/b9f9be647da33c3cdd45d570d985e469376ce0cd

PR #4957, opened 2015-12-01
Fix various typos/misspellings/formatting issues in the docs

treythomas123:
I noticed a misspelling when reading the docs, so I ran a spellcheck on all the doc files and found lots of typos.

I also noticed some of the doc files used character entity names (`&period;`, `&comma;`, `&lpar;`, `&rpar;` etc.) in places where they were unnecessary, so I replaced those instances with the actual characters.

The second commit in this PR is just the result of running `npm run docs`.

### DIFF
--- a/docs/articles/heroku.md
+++ b/docs/articles/heroku.md
@@ -31,7 +31,7 @@ Next step: Deploy the application to Heroku.
 
 ## Deployment to Heroku
 
-First of all, we need to add the right version of Node.JS and NPM to the `package.json`. The file should look similiar to this:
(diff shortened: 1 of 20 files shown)


mickhansen:
Can you remove the second commit? Kills the diff and we autogenerate on deploy.

treythomas123:
Would it be better to just remove the generated docs from source control? Having them there could be a little confusing to people coming in to edit them. I actually started editing some of the files directly before I noticed they had been auto-generated.

treythomas123:
Removed the commit with the generated docs. I'll write another PR to discuss removing the docs from source control.

mickhansen:
I think we need them in source for readthedocs.

PR #4965, opened 2015-12-02
Add transaction support to all migration queries

treythomas123:
I noticed one of my migrations wasn't honoring the transaction option I passed in, so I took a look at the code, and was surprised to find that the `transaction` option was being ignored for most of the migration queries. Only the `logging` option is being passed through to `sequelize.query` in most cases.

This allows the transaction option to be passed in as well.

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -45,11 +45,11 @@ QueryInterface.prototype.dropAllSchemas = function(options) {
 QueryInterface.prototype.showAllSchemas = function(options) {
   var self = this;
 
-  options = _.assign({
+  options = _.assign({}, options, {
     raw: true,
     type: this.sequelize.QueryTypes.SELECT,
     logging: false
-  }, options || {});
+  });
 
   var showSchemasSql = self.QueryGenerator.showSchemasQuery();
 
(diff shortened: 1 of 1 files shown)


lib/query-interface.js
```diff
@@ -20,13 +20,13 @@ var QueryInterface = function(sequelize) {
 QueryInterface.prototype.createSchema = function(schema, options) {
   options = options || {};
   var sql = this.QueryGenerator.createSchema(schema);
-  return this.sequelize.query(sql, {logging: options.logging});
+  return this.sequelize.query(sql, {logging: options.logging, transaction: options.transaction});
```
mickhansen: Should possible just pass along options as is, helps plugins too.

treythomas123: I actually started doing it that way, but thought there might be some reason the other options besides logging were being stripped. If it's ok to pass along all options, I can update the PR to do that.

mickhansen:
jshint error

lib/query-interface.js
```diff
@@ -95,7 +95,8 @@ QueryInterface.prototype.createTable = function(tableName, attributes, options,
       if (attributes[keys[i]].type instanceof DataTypes.ENUM) {
         sql = self.QueryGenerator.pgListEnums(tableName, attributes[keys[i]].field || keys[i], options);
         promises.push(self.sequelize.query(
-          sql, { plain: true, raw: true, type: QueryTypes.SELECT, logging: options.logging, transaction: options.transaction }
+          sql,
+          _.assign({ plain: true, raw: true, type: QueryTypes.SELECT }, options)
```
mickhansen: might want to change to `_.assign({}, options, { plain: true, raw: true, type: QueryTypes.SELECT })` for good measure

treythomas123:
Anything wrong with this PR or can it be merged?

mickhansen:
@treythomas123 Sorry about that

PR #4967, opened 2015-12-02
Failing .set test

sorin:
This test is failing - somehow setting a JSONB field mutates the model's defaultValue to whatever was last set. Couldnt identify how that is happening. @mickhansen any clues?


mickhansen:
Hmm, default values must not be cloned on initialization.
https://github.com/sequelize/sequelize/blob/master/lib/utils.js#L221 will probably need to check if something is an object and do a clone.

sorin:
@mickhansen I think I got it, can you take a look please?

PR #4969, opened 2015-12-03
Retry query when SQLITE_BUSY is thrown as error.

EToreo:
This pull request satisfies the situation where Sqlite reports back that a database is busy.  It uses https://github.com/mickhansen/retry-as-promised to ease the rate of retry as well as match the error.  The match option can be easily expanded to catch other errors that should be retried.

Before this will work:
* https://github.com/sequelize/sequelize/pull/4911 must be accepted - for blocking transaction types in Sqlite
* https://github.com/mickhansen/retry-as-promised/pull/1 must be accepted - for retry

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ We're glad to get pull request if any functionality is missing or something is b
 Still interested? Coolio! Here is how to get started:
 
 ### 1. Prepare your environment
-Here comes a little surprise: You need [Node.JS](http://nodejs.org). In order to be a productive developer, I would recommend the latest v0.10.
+Here comes a little surprise: You need [Node.JS](http://nodejs.org).
 
 ### 2. Install the dependencies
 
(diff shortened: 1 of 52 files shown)


mickhansen:
So this automatically retries queries in the case of SQLITE_BUSY yeah?
I like this idea, i'd also love to introduce it for transactions that fail on `unable to serialize on concurrenct update` for transactions (since the PG spec says you should retry the entire transaction)

EToreo:
@mickhansen It does, yes. I suspect that adding similar behavior to transactions would be easy enough and when I have some more time I would be happy to investigate that in another pull request.

mickhansen:
sqlite build is failing currently, and we'll need a rebase.

EToreo:
@mickhansen The unit tests are passing on my dev checkout of this code.  I have made sure the code is the same and my environment is clean.  Could you check out this branch and see if you are able to get the unit tests to run or if you are met with the same build failure?

EToreo:
@mickhansen I have tried this branch on a fresh checkout and install on a different box and it is passing unit tests when run from the command line.  If you could do the same I would be interested in the result.

EToreo:
Closed in favor of https://github.com/sequelize/sequelize/pull/5157#issuecomment-169914982

PR #4976, opened 2015-12-04
Merge of pull request  #4706 - pg range handling improved to HEAD

yazshel:
I've merged @alekbarszczewski's pull request #4706 with master and resolved the conflicts The changes to the parse() method are greatly simplified with the recent master changes.

### DIFF
--- a/lib/dialects/postgres/range.js
+++ b/lib/dialects/postgres/range.js
@@ -2,12 +2,34 @@
 
 var _ = require('lodash');
 
+function stringifyRangeBound (bound) {
+  if (bound === null) {
+    return '' ;
+  } else if (bound === Infinity || bound === -Infinity) {
+    return bound.toString().toLowerCase();
+  } else {
+    return JSON.stringify(bound);
+  }
+}
+
+function parseRangeBound (bound, parseType) {
+  if (!bound) {
+    return null;
+  } else if (bound === 'infinity') {
+    return Infinity;
+  } else if (bound === '-infinity') {
+    return -Infinity;
+  } else {
+    return parseType(bound);
+  }
+}
+
 function stringify (data) {
   if (data === null) return null;
 
-  if (!_.isArray(data) || data.length !== 2) return '';
-
-  if (_.any(data, _.isNull)) return '';
+  if (!_.isArray(data)) throw new Error('range must be an array');
+  if (!data.length) return 'empty';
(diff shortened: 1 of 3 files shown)


alekbarszczewski:
@yazshel wow thanks man! I had busy days and couldn't look into it.

mickhansen:
pg build is failing

lib/dialects/postgres/range.js
```diff
@@ -37,7 +67,7 @@ function parse (value, parser) {
 
   result = result
     .map(function (value) {
-      return parser(value);
+      return parseRangeBound(parser);
```
janmeier: I think you need to pass `value, parser` here

yazshel:
Doh! Didn't test my code enough! I've committed dbfee88f58780a8c5087f3cb6640bf5d41c19f66 which also passes 'value'...

yazshel:
Just to note, the conflict merge involved removing most of the date parsing changes from commit #2d02396 - these should no longer be necessary. AFAICT, the postgres dialect is now using getTypeParser() from the pg-native -> pg-types module, which handles parsing TIMESTAMP WITH TIMEZONE correctly.

mickhansen:
PG build still failing

yazshel:
I've added a commit updating the integration tests which should hopefully pass now... I haven't been able to test locally though because I'm on a Mac and the testing infrastructure is a PITA to set up.

One of the tests does require the `pg` module however, as it uses the type parser from this. I've made the test conditional on being able to load the module. Is this the correct way to do it, or should the test be removed given it's now using the `pg` / `pg-types` modules to do the stuff that we're testing?

janmeier:
It should not be necessary to check for the pg module, since we already do ` if (dialect.match(/^postgres/)) {` in the beginning of the file

janmeier:
If you just remove the try around pg (move it inside the if instead), rebase and add a changelog entry this should be good to go

yazshel:
@janmeier Rebased to `master` and changelog entry added.

janmeier:
Thanks a lot @yazshel  and @alekbarszczewski  :)  !

alekbarszczewski:
@janmeier / @yazshel Thank you!

yazshel:
The thanks should go to you @alekbarszczewski - I'd been playing with [sequelize](https://github.com/sequelize/sequelize) for a little while and was disappointed with the missing support for infinite bounds in pg ranges... until I discovered your pull request! Keep up the great work!
Thanks @janmeier for your advice and the pull.

PR #4989, opened 2015-12-07
Doc change - custom auto-generated values

oravecz:
Updated documentation to include mention of returning values from a function for primary keys.


janmeier:
Thanks! :)

Issue #4943, opened 2015-11-28
queryInterface.removeColumn fails during migration


apsavin:
Sequelize [Node: 4.2.2, CLI: 2.2.1, ORM: 3.14.0]
mssql

``` js
module.exports = {
  up: function(queryInterface) {
    return queryInterface.removeColumn('TableName', 'ColumnName'));
  },

  down: function(queryInterface, Sequelize) {
    return queryInterface
      .addColumn('TableName', 'ColumnName', {
        type: Sequelize.STRING(100),
        field: 'ColumnName',
        allowNull: false,
        defaultValue: 'some value',
      });
  },
};
```

`removeColumn` will fail because of `defaultValue` constraint.
And I [can't](https://github.com/sequelize/sequelize/issues/3116) set up a name for the constraint so it's not convenient to remove the constraint with plain sql query (query becomes more complex because I need to get the constraint's name firstly)

mickhansen:
I was not aware that default values could constrain the table in any way.
Can you post the SQL called and the full error message?

apsavin:
I was not aware, too, but now I am aware) Thanks to ms.

Sequelize error message:
`Unhandled rejection SequelizeDatabaseError: ALTER TABLE DROP COLUMN ColumnName failed because one or more objects access this column.`

```sql
sql> ALTER TABLE TableName DROP COLUMN ColumnName
[2015-11-29 15:00:28] [S0001][5074] The object 'DF_TableName_ColumnName is dependent on column 'ColumnName'.
```

Because of 

```sql
ALTER TABLE TableName
    ADD CONSTRAINT DF_TableName_ColumnName
    DEFAULT N'some value' FOR ColumnName;
```

When I use `queryInterface.addColumn` with `defaultValue` behavior is the same, just the name of constraint is not so `beautiful` because ORM auto-generates it.

See [this](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/a1ea6d8d-bef1-4f6e-8f66-a16b812de6fc/drop-column-with-default-constraint?forum=transactsql) discussion on social.msdn.microsoft.com. It's a [common](https://github.com/laravel/framework/issues/4402) problem. Unfortunetaly, I can't find any official ms docs about it.

Workaround: just use `queryInterface.sequelize.query` instead of `addColumn`/`removeColumn` methods if the column has a default value.

MichaelKearnsSoftwire:
This is also an issue my team has been working around for a while.

[...]

PR #4990, opened 2015-12-07
Mssql can remove column with default

MichaelKearnsSoftwire:
* this fixes Github issue #4943
* previously, calling removeColumn for a column with a defaultValue
  specified would cause an error -- this is because in MSSQL you have
  to drop the constraint before the column. See:
      http://stackoverflow.com/questions/8641954/how-to-drop-column-with-constraint
* also added integration tests for removeColumn function

----

I've verified that all the MSSQL tests pass, but haven't yet run the tests against other vendors -- will attempt to do that soon.

I think this is otherwise mostly ready, but there are some rough edges I have some questions about:

* I've noticed that some of the tests in query-interface.test.js don't actually assert that the correct modification to the DB has been made (e.g. addColumn doesn't ever verify that the column has been added, merely that some SQL was run and no errors occurred.) I've followed this pattern for removeColumn, but if there's no good reason for it I'm inclined to use describeTable to verify that the column has actually been removed (and likewise for existing tests)
* My mssql query-generator.test.js is copy-pasted-and-adjusted from the similar ones for other vendors; am happy to commonise the describe block (which seems to be pretty much the same across vendors) so that each test merely contains a bunch of scenarios.

This is my first non-trivial PR to Sequelize, so let me know if I've missed anything!


mickhansen:
>  I've followed this pattern for removeColumn, but if there's no good reason for it I'm inclined to use describeTable to verify that the column has actually been removed (and likewise for existing tests)

Since lazyness is not a good reason, there is none, all schema modifying calls should really have integration tests that verify correctness via a describe call.

test/integration/dialects/mssql/query-generator.test.js
```diff
@@ -0,0 +1,49 @@
...
+            QueryGenerator.options = _.assign(context.options, { timezone: '+00:00' });
+            QueryGenerator._dialect = this.sequelize.dialect;
+            var conditions = QueryGenerator[suiteTitle].apply(QueryGenerator, test.arguments);
+            expect(conditions).to.deep.equal(test.expectation);
+          });
+        });
+      });
+    });
+  });
+}
```
mickhansen: Please move tests for sql generation to `test/unit` instead.

test/integration/query-interface.test.js
```diff
@@ -454,6 +454,89 @@ describe(Support.getTestDialectTeaser('QueryInterface'), function() {
...
+          }
+        });
+      });
+
+      it('should be able to remove a column with a default value', function() {
+        return this.queryInterface.removeColumn('users', 'firstName', {
+          logging: log
+        }).then(function() {
+          expect(count).to.be.at.least(1);
+          count = 0;
```
mickhansen: Ugh this should all really be refactored to use spys, but i can't fault you for following the rest of the suite :)

lib/dialects/mssql/query-interface.js
```diff
@@ -0,0 +1,42 @@
...
+
+  var findConstraintSql = self.QueryGenerator.getDefaultConstraintQuery(tableName, attributeName);
+  return self.sequelize.query(findConstraintSql + ';', { raw: true, logging: options.logging})
+    .spread(function (results) {
+      if (!results.length) {
+          // No default constraint found -- we can cleanly remove the column
+          return;
+      }
+      var dropConstraintSql = self.QueryGenerator.dropConstraintQuery(tableName, results[0].name);
+      return self.sequelize.query(dropConstraintSql + ';', { raw: true, logging: options.logging});
```
mickhansen: Any reason the `;` can't be a part of the query generator?

mickhansen:
New tests appear to fail for postgres. https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/query-generator.js#L196 needs to be refactored to use `this.quoteTable(this.extractTableDetails(tableName))` rather than just `this.quoteTable(tableName)`.

Thank you for this contribution, the schema modification queries need a lot of love :)

MichaelKearnsSoftwire:
Thanks for the comments! I should have time to address them later in the week.

MichaelKearnsSoftwire:
As you can tell my definition of "later in the week" is somewhat liberal :)

> Since lazyness is not a good reason, there is none, all schema modifying calls should really have integration tests that verify correctness via a describe call.

@mickhansen  I'm doing this now -- am I right in thinking that I can get rid of the assertions against count for most places where I make assertions against the result of a describe call? I don't see them being particularly valuable now we're verifying that changes to the DB have actually occurred.

lib/dialects/mssql/query-interface.js
MichaelKearnsSoftwire: Not really -- initially I thought this was the standard approach, but I can see that it varies quite a lot! Have moved the semicolons to query-generator.

mickhansen:
Should be no reason to verify the amount of sql queries run since that's really an implementation detail.

MichaelKearnsSoftwire:
@mickhansen I've finally fixed the PG tests -- it looks like they were failing because the tests were not being torn down correctly.

I've added a call to dropAllSchemas in the afterEach for this test to ensure this doesn't happen -- please let me know if there's a better way.

Also, I've implemented your previous suggestion of calling this.extractTableDetails in lib/dialects/postgres/query-generator.js, but this is no longer required for the tests to pass -- I'm not very familiar with Postgres so some guidance here would be helpful! (I also had to change lib/dialects/abstract/query-generator.js for extractTableDetails to work)

test/unit/dialects/mssql/query-generator.test.js
```diff
@@ -0,0 +1,49 @@
...
+            QueryGenerator.options = _.assign(context.options, { timezone: '+00:00' });
+            QueryGenerator._dialect = this.sequelize.dialect;
+            var conditions = QueryGenerator[suiteTitle].apply(QueryGenerator, test.arguments);
+            expect(conditions).to.deep.equal(test.expectation);
+          });
+        });
+      });
+    });
+  });
+}
```
mickhansen: Could you make these in the form of: https://github.com/sequelize/sequelize/blob/master/test/unit/sql/select.test.js instead?

mickhansen:
Looks good, just a single comment and i'd also love to see the commits squashed to one or two descriptive ones instead of the current 10+

test/unit/dialects/mssql/query-generator.test.js
MichaelKearnsSoftwire: Mind if I put this in a separate PR? This is consistent with the other pre-existing query-generator tests, and I'd like to avoid making this PR much bigger :)

mickhansen: The query generator tests from the integration suite are a deprecated format, and not one we're looking to keep on using.

MichaelKearnsSoftwire: Okay -- I've done this for the mssql one as I've added the file, but I don't have time to do this for the other dialects soon I'm afraid.

MichaelKearnsSoftwire:
Thanks -- I've squashed the previous tower of commits down to 6 pretty-independent commits.

Since my previous commits, a new test has been added to query-interface.test.js which uses changeColumn to add a FK reference:
```javascript
        it('able to change column to foreign key', function() {
          return this.queryInterface.changeColumn('users', 'level_id', {
            type: DataTypes.INTEGER,
            references: {
              model: 'level',
              key:   'id'
            },
            onUpdate: 'cascade',
            onDelete: 'cascade'
          }, {logging: log}).then(function() {
            expect(count).to.be.equal(1);
            count = 0;
          });
        });
```
As far as I can tell, there's no way to verify that this FK has been added using describeTable as we now do for all the other tests -- are you aware of any other way to verify that this changeColumn has actually worked? I've left it as-is for now.

Also, do you have any comment re: my query on this.extractTableDetails above?

Thanks!

mickhansen:
quoteTable/extractTableDetails should be required on most things, if the tests arent failing it's because we have insufficient schema tests for those cases.

If it's not part of describeTable there's likely not a good way to verify this at the moment.

MichaelKearnsSoftwire:
Thanks!

mickhansen:
Released in `3.19.2`

Issue #4631, opened 2015-10-09
PostgreSQL autocommit


foxy0669:
It looks like PostgreSQL 9.4 has dropped the support for AUTOCOMMIT=OFF and PostgreSQL 9.5 has dropped support all together.

9.4
`ERROR:  SET AUTOCOMMIT TO OFF is no longer supported`

9.5
`ERROR:  unrecognized configuration parameter "autocommit"`

Is there anyway to bypass then when in a transaction?

mickhansen:
I'm not quite sure why we originally have that setting at all.
The default for all dialects is OFF i believe, having it ON would generally be weird given you're working in a transaction (maybe in CLI settings it makes sense).

/cc @janmeier

janmeier:
I've honestly never understood that setting  :D

See also https://github.com/sequelize/sequelize/issues/3568

As you say its off by default, and I can't see a setting where you'd want to enable autocommit within a transaction

mickhansen:
@janmeier We should refactor so autoCommit is only set if there's an ACTUAL value (other than a default value). Don't see it as breaking.

foxy0669:
I noticed the problem after installing 9.5. It looks to me the setting is on by default, however setting it to off would break by the look of it at least 9.1 upwards. Unless off would not issue the statement?

This is a default transaction with no options.

```
Executing (57de5f41-bd7e-494c-b1c8-53735742d7cc): START TRANSACTION;

Executing (57de5f41-bd7e-494c-b1c8-53735742d7cc): SET SESSION TRANSACTION ISOLAT
ION LEVEL REPEATABLE READ;

Executing (57de5f41-bd7e-494c-b1c8-53735742d7cc): SET autocommit = 1;
```

janmeier:
@mickhansen Agreed

DigitalSkyline:
Having the same issue in postgres 9.5.

How does one prevent "SET autocomit = 1;" ? Is there a workaround/fix?

Seems to be happening with findOrCreate, here's the error.

www-0 Executing (c04f3f10-9ac6-4271-a642-d4f19ef16300): START TRANSACTION;
www-0 Executing (c04f3f10-9ac6-4271-a642-d4f19ef16300): SET SESSION TRANSACTION ISOLATION LEVEL REPEATABLE READ;
www-0 Executing (c04f3f10-9ac6-4271-a642-d4f19ef16300): SET autocommit = 1;
[...]

PR #5010, opened 2015-12-09
Disables setting AUTOCOMMIT for Postgres >= 9.4.0

tkalfigo:
Fixes issue https://github.com/sequelize/sequelize/issues/4631
User needs to supply "databaseVersion" key in options object
when instantiating the Sequelize class. Value should be semver
compliant e.g. "9.5.0" but not "9.5.0beta2"


mickhansen:
Version is also auto detected on the first connection if not supplied.

lib/dialects/postgres/query-generator.js
```diff
@@ -873,7 +874,11 @@ var QueryGenerator = {
     }
 
     // POSTGRES does not support setting AUTOCOMMIT = OFF
-    if (!value) {
+    // Additionally it does not support AUTOCOMMIT at all starting at v9.5
+    // The assumption is that it won't be returning in future versions either
+    // The options param passed when creating the Sequelize instance should include a key "databaseVersion" 
+    // with semver compliant value e.g. "9.5.0" (without 'alpha' or 'beta' classifications)
+    if (!value || semver.gte(this.sequelize.options.databaseVersion, '9.4.0')) {
```
mickhansen: if starting with 9.5 it needs to be gte 9.5.0 or gt 9.4.0

tkalfigo:
Unfortunately I didn't get this behavior. If I don't supply databaseVersion: "9.5.0" explicitly as part of my options object when instantiating Sequelize, it thinks I'm on "9.4.0".

mickhansen:
https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/connection-manager.js#L213
This code path should check version on the first query.

mickhansen:
But if you have a beta version you might be hitting this: https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/connection-manager.js#L224 which then defaults to: https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/index.js#L51

tkalfigo:
@mickhansen You are right; that explains Sequelize's behavior with 9.5.0beta2.

Regarding the choice to go for gte(9.4.0) and not gte(9.5.0) it's because starting at 9.4.0 AUTOCOMMIT = OFF is *not* allowed; so the only choice is AUTOCOMMIT = ON.

The server itself has autocommit, so every program that uses PostgreSQL will be in autocommit mode by default unless there is an explicit BEGIN or START TRANSACTION. And if you are in a transaction, there is no point doing an AUTOCOMMIT = ON anyway.

Hope I didn't miss any case, but it seems that playing with AUTOCOMMIT through Sequelize is pointless starting at 9.4.0.

mickhansen:
autocommit as a whole is pretty dumb/pointless for transactions, not sure why it ever made it in there.
thanks for the clarification of 9.4.0, could've read the comments more closely :)

tkalfigo:
Is something missing from this to be merged?

mickhansen:
Hmm, i might have been waiting for CI and then forgot about it.
It'll need a rebase now, mind doing that?

tkalfigo:
Done.

mickhansen:
Ah looks like you merged rather than rebases, which messes up history a bit.

tkalfigo:
Will redo PR.

PR #5019, opened 2015-12-11
Fix some bluebird warnings

BridgeAR:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM iojs:1.6
+FROM node:4
```
mickhansen: Afair pg-native will fail unless npm is updated then.

test/integration/model/create.test.js
```diff
@@ -1571,13 +1571,12 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
-          expect(errors[0].errors.get('name')[0].type).to.equal('notNull Violation');
-          expect(errors[1].record.name).to.equal('bar');
-          expect(errors[1].record.code).to.equal('1');
-          expect(errors[1].errors.get('code')[0].message).to.equal('Validation len failed');
+          expect(errors.errors).to.have.length(2);
+          expect(errors.errors[0].record.code).to.equal('1234');
+          expect(errors.errors[0].errors.get('name')[0].type).to.equal('notNull Violation');
+          expect(errors.errors[1].record.name).to.equal('bar');
+          expect(errors.errors[1].record.code).to.equal('1');
+          expect(errors.errors[1].errors.get('code')[0].message).to.equal('Validation len failed');
```
mickhansen: Hmm, is this a BC breaking change?

Dockerfile
BridgeAR: pg-native works with node 5 too.

test/integration/model/create.test.js
BridgeAR: Yes, it is. You might consider it as a bugfix though, since throwing an array is not that nice. But some people will likely run into this .

mickhansen: Oh i definitely consider it a bugfix aswell :)

Dockerfile
mickhansen: Only on more recent versions on npm, there's an issue for it somewhere.

BridgeAR: This seemed to be solved: https://github.com/brianc/node-pg-native/issues/27

BridgeAR:
@mickhansen Do you want anything for me to add / change before you consider this ready to merge?

Dockerfile
mickhansen: That's the issue i'm referring to, if you read it he writes that it's only fixed in npm@2.14.4 (which was not on travis a few weeks ago when i checked, but possibly changed), not sure what version of npm ships with node:4 docker image

BridgeAR: Docker uses [4.2](https://github.com/docker-library/official-images/blob/master/library/node#L48). But this is not important, so if you want me to remove this, it's all fine to me.

mickhansen: No then it's probably fine, just wanted to make sure i could still run tests for all dialects including pg-native :)

mickhansen:
@BridgeAR Well there's the question of a major bump

lib/instance-validator.js
```diff
@@ -271,8 +271,7 @@ InstanceValidator.prototype._invokeBuiltinValidator = Promise.method(function(va
   }
 
   if (!validator[validatorType].apply(validator, [value].concat(validatorArgs))) {
-  // extract the error msg
-    throw test.msg || 'Validation ' + validatorType + ' failed';
+    throw new Error(test.msg || 'Validation ' + validatorType + ' failed');
```
benjamingr: Yes please, we do not give you meaningful debug information if you throw non-errors.

lib/model.js
```diff
@@ -2101,21 +2101,18 @@ Model.prototype.bulkCreate = function(records, options) {
...
         }
+        return instance.validate(options).then(function (err) {
+          if (err) {
+            errors.push({record: instance, errors: err});
+          }
+        });
       }).then(function() {
         delete options.skip;
         if (errors.length) {
-          return Promise.reject(errors);
```
benjamingr: This works, you may also use AggregateError which comes with bluebird.

lib/model.js
```diff
@@ -2101,21 +2101,18 @@ Model.prototype.bulkCreate = function(records, options) {
...
+          return Promise.reject(new self.sequelize.ValidationError('', errors));
```
benjamingr: This works, you may also use AggregateError which comes with bluebird.

benjamingr:
LGTM

PR #5020, opened 2015-12-11
Fix count when scope contains attributes

### LINKED ISSUE #4566: findAndCountAll with scope having attributes
When having a defaultScope defined, which contains the attributes options, the count query is messed up.

```javascript
defaultScope: {
        where: {
          active: true
        },
        attributes: ['id', 'username', 'email', 'active', 'approved', 'company_id']
      }
```
The following query is generated 
```sql
SELECT "id", "username", "email", "active", "approved", "company_id", count(*) AS "count" FROM "users" AS "users" WHERE "users"."company_id" = 1 AND "users"."active" = true;
```

2 errors within the query. The first one, in that case, the "group by" is required. The second error is that it should normally not include the "attributes" for the count query.
If I fix the first one (add the "group" option), the query runs, but "res.count" is actually returning anarray of objects from the query, with the "count" field inside.

Example : 
```javascript
{ count:
   [ { id: 1,
       username: 'Test',
       email: 'test@test,cin',
       active: true,
       approved: true,
       company_id: 1,
       count: '1' },
     { id: 2,
       username: 'Test2',
       email: 'test@test.com',
       active: true,
       approved: true,
       company_id: 1,
       count: '1' } ],
      rows:....
}
```

The error come from the file `https://github.com/sequelize/sequelize/blob/master/lib/model.js`

The findAndCountAll strips correctly the attributes options (https://github.com/sequelize/sequelize/blob/master/lib/model.js#L1604)
But then within the count query, the scope is injected with the attributes and all other options that should be omitted (https://github.com/sequelize/sequelize/blob/master/lib/model.js#L1542) 

I'd be happy to fix it, but I do not know what would be the best possible option to fix it. I see the following possibilities : 
- Add a parameter to the count function to omit the attributes/... options after the scope is injected
- Separate the count function in 2, one which will prepare the parameters for the normal count, and the second that actually does the count. This way the findAndCount function would use directly the "count" part, as the option is already ready.

Any toughs on this ?

Thanks

GabeIsman:
This fixes #4566 by specifically ignoring attributes in `count` (and therefore `findAndCount`). Attributes are mostly likely added by a scope, and when present cause the SQL error
```
SequelizeDatabaseError: column "ScopeMe.username" must appear in the GROUP BY clause or be used in an aggregate function
```
because a bad query is produced (see issue for details).

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1569,6 +1569,7 @@ Model.prototype.count = function(options) {
   options.limit = null;
   options.offset = null;
   options.order = null;
+  options.attributes = [];
 
   return this.aggregate(col, 'count', options);
 };
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # Next
 - [FIXED] Apply scopes to `aggregate` [#4764](https://github.com/sequelize/sequelize/issues/4764)
 - [ADDED/FIXED] Lower case `onDelete` option to allow the use of `onDelete: 'CASCADE', hooks: true`.
+- [FIXED] Ignore attributes in `count` [#4566](https://github.com/sequelize/sequelize/issues/4566)
 
 # 3.13.0
 - [FIXED] timestamp columns are no longer undefined for associations loaded with `separate`. [#4740](https://github.com/sequelize/sequelize/issues/4740)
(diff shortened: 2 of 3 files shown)


mickhansen:
LGTM, can you squash your commits to a single descriptive one?

GabeIsman:
Thanks for reviewing. Done!

lib/model.js
```diff
@@ -1569,6 +1569,7 @@ Model.prototype.count = function(options) {
   options.limit = null;
   options.offset = null;
   options.order = null;
+  options.attributes = [];
```
mickhansen: Maybe this should be moved to `aggregate` if we apply scope to all aggregate calls now, what do you think @janmeier?

GabeIsman:
Looking at `aggregate` it seems like maybe that was the original intention with this line:
```
  options = Utils._.extend({ attributes: [] }, options || {});
```
if the arguments to extend are re-ordered it would have the same effect. Not sure though. The comments on aggregate mention a use case with multiple attributes, so maybe I'm just not thinking of that use case?

mickhansen:
Hmm you're right, looks like it's setup to possible support multiple columns: https://github.com/sequelize/sequelize/blob/master/lib/model.js#L1519 

Better keep it like that for now.

PR #5022, opened 2015-12-12
Fixed Invalid Statement in Model API Docs

sushantdhiman:


mickhansen:
Thanks :)
For reference you don't need to change `docs/api` since its autogenerated from the code comments.

PR #5032, opened 2015-12-14
Fix for #4904

shoshomiga:


test/unit/sql/select.test.js
```diff
@@ -232,4 +232,15 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
...
 
   });
+
+  suite('raw query', function () {
+    test('raw replacements', function () {
+      expectsql(sql.selectQuery('User', {
+        attributes: ['*'],
+        having: ['name IN (?)', [1, 'test', 3, "derp"]]
+      }), {
+        default: "SELECT * FROM [User] HAVING name IN (1,'test',3,'derp');"
```
mickhansen: What about the cases where postgres users actually need an array?

shoshomiga: I thought about this too but it doesn't seem that this is a documented functionality with raw queries

mickhansen: How would bind parameters solve this exactly? uses the same replacements code

shoshomiga: I meant bind parameters in the sense of prepared queries, but actually looking at the docs it seems that arrays are not supported

You could call sequelize.escape manually instead and you would get a postgres ARRAY[], I think this is kinda strange though and maybe not the best way to do it

In any case, this PR should bring back exactly the same behavior as in pre-3.14, which means breaking people's code if they relied on post-3.14 behavior for postgres

PR #5039, opened 2015-12-14
Prefix string literals with N in mssql

### LINKED ISSUE #3752: Need to add prefix 'N' before Unicode string when work with MSSQL

zapatek:
Fixes #3752

### DIFF
--- a/test/integration/model/find.test.js
+++ b/test/integration/model/find.test.js
@@ -121,7 +121,7 @@ describe(Support.getTestDialectTeaser('Model'), function() {
           where: ['specialkey = ?', 'awesome'],
           logging: function(sql) {
             test = true;
-            expect(sql).to.contain("WHERE specialkey = 'awesome'");
+            expect(sql).to.match(/WHERE specialkey = N?'awesome'/);
           }
         }).then(function() {
           expect(test).to.be.true;
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# Future
+- [FIXED] Support Unicode strings in mssql [#3752](https://github.com/sequelize/sequelize/issues/3752)
+
 # 3.15.1
 - [FIXED] calling Model.update() modifies passed values  [#4520](https://github.com/sequelize/sequelize/issues/4520)
 - [FIXED] Instance can be chained on .set() and other methods [#4702](https://github.com/sequelize/sequelize/issues/4702)
--- a/lib/sql-string.js
+++ b/lib/sql-string.js
@@ -13,6 +13,7 @@ SqlString.escapeId = function(val, forbidQualified) {
 };
 
 SqlString.escape = function(val, timeZone, dialect, format) {
+  var prependN = false;
   if (val === undefined || val === null) {
     return 'NULL';
   }
@@ -27,6 +28,11 @@ SqlString.escape = function(val, timeZone, dialect, format) {
     return '' + !!val;
   case 'number':
     return val + '';
+  case 'string':
+    // In mssql, prepend N to all quoted vals which are originally a string (for
+    // unicode compatibility)
+    prependN = dialect === 'mssql';
+    break;
   }
 
   if (val instanceof Date) {
@@ -66,7 +72,7 @@ SqlString.escape = function(val, timeZone, dialect, format) {
       }
     });
   }
-  return "'" + val + "'";
+  return (prependN ? "N'" : "'") + val + "'";
 };
 
 SqlString.format = function(sql, values, timeZone, dialect) {
--- a/test/integration/model/create.test.js
+++ b/test/integration/model/create.test.js
@@ -712,7 +712,7 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         intVal: this.sequelize.cast('1', type)
       }, {
         logging: function(sql) {
-          expect(sql).to.match(new RegExp("CAST\\('1' AS " + type.toUpperCase() + '\\)'));
+          expect(sql).to.match(new RegExp("CAST\\(N?'1' AS " + type.toUpperCase() + '\\)'));
           match = true;
         }
       }).then(function(user) {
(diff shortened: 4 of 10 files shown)


mickhansen:
This probably fits better than data types, @janmeier?

zapatek:
@mickhansen, the enclosing quotes for string literals are put-in in the String.escape function. Since N'...' can be thought as a special variant of enclosing quotes,  I thought it would be the right function to modify.

On a separate note, this has been a showstopper for us. Not getting it fixed basically means sequelize is  not usable (with mssql) in majority of Asia. 

Given my unfamiliarity with internals of sequelize, If you could provide a more specific suggestion as to where to look at, I am willing to put in the effort and get this fixed at the soonest. On the other hand, if this is good enough then I can also fix those test cases which fail after this fix (they fail because they do not anticipate the N prefix in the built mssql queries at the moment).

mickhansen:
@janmeier will chime have to chime in here, but since it's mssql specific i think it's better suited being implemented on mssql specific data types: https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/data-types.js

zapatek:
Thanks for the pointer, I'll look at it. Meanwhile, on the defence of the current approach, the modified `SqlString.escape` function has already some dialect specific code such as:

```javascript
  if (dialect === 'postgres' || dialect === 'sqlite' || dialect === 'mssql') {
    // http://www.postgresql.org/docs/8.2/static/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS
    // http://stackoverflow.com/q/603572/130598
    val = val.replace(/'/g, "''");
  } else {
    val = val.replace(/[\0\n\r\b\t\\\'\"\x1a]/g, function(s) {
      switch (s) {
```

By the way, I have modified the test cases accordingly (anticipating the N prefix for mssql) and they all are passing now.

mickhansen:
@zapatek It indeed does, and we'd like to fix that going forward :)

test/integration/model/find.test.js
```diff
@@ -121,7 +121,7 @@ describe(Support.getTestDialectTeaser('Model'), function() {
           where: ['specialkey = ?', 'awesome'],
           logging: function(sql) {
             test = true;
-            expect(sql).to.contain("WHERE specialkey = 'awesome'");
+            expect(sql).to.match(/WHERE specialkey = N?'awesome'/);
```
mickhansen: The `?` should not be present in the final result, as evident from the existing test.

mickhansen:
Okay looking at the tests you changed i'm not sure that a change to data types will be sufficient actually, but we'll have to wait for @janmeier  to take a look

test/integration/model/find.test.js
zapatek: Ah, that question mark added in the fix is just a regular expression quantifier, meaning one or zero N characters are expected (to cover cases for mssql - one - and other dialects - zero - respectively)

mickhansen: Oh right, my bad, for a second i missed that you chained it to a regex.

janmeier:
Nah, I think this is the best way to fix it for now - either that or we have to move some more of the logic for proper escaping etc. into data types as well - So I think this is fine for now

janmeier:
Could you add a changelog entry, then this should be good to go?

zapatek:
Done.

zapatek:
Seems like just missed the 3.15.0 release. Rebased the work branch for this fix to the tip of master to resolve the resultant conflict on changelog.md

mickhansen:
I'll cut another one once i merge a few pending PRs :)

mickhansen:
@zapatek Ah damn, looks like we need another rebase. Please do ping me when it's done and tests have run so i can get it merged in time.

zapatek:
Amended the conflicting changelog.md but seems it wasn't in time for 3.15.1, so I put the entry under the 'Future' header. If you could merge now, it would surely make it into next release.. :)

mickhansen:
And needs a new rebase, how the hell does this keep happening to this PR :p

zapatek:
Haha, this time the culprit was `sql-string.js` changed by the recent commit 4f4abdc. I have just resolved the merge conflict. Could you try to merge the pr once again?

brentjanderson:
This explains why my latin-encoded columns no longer work. I have a view that has the following structure:

Column_Name | Type | Length | Nullable | Collation
-------------------- | ------ | ---------| ------------- | ------------
Column_A	| char | 7 | no | SQL_Latin1_General_CP1_CI_AS
Column_B | char | 31 | no | SQL_Latin1_General_CP1_CI_AS

When I use a pretty standard Sequelize raw query, it generates the following with this new code:

```sql
SELECT [Column_A], [Column_B] 
FROM [xyzView] AS [xyzView] 
WHERE [xyzView].[Column_A] LIKE N'Value' OR [xyzView].[Column_B] LIKE N'Value'
```

Unfortunately, that code does not return any results. Dropping the Unicde prefix `N` from the strings in the `WHERE` clause does return results. I reviewed your changes, and it seems that all strings now are forced to unicode format via Sequelize. Did I miss an override? Should I get cracking on a pull request for such an option? Otherwise I may just pin my project at the older version.

mickhansen:
@brentjanderson I don't believe any override is availab,e please do provide a PR

Shaffuan:
Why don't update this changes in sequelize latest version???

Issue #4850, opened 2015-11-12
transaction fails with already rolled back if connection fails


novacrazy:
So I recently did a clean install of my `node_modules` directory, and I guess something small upgraded automatically (semver and all that) that broke sequelize transactions entirely.

This is all I have to go on at the moment.

```
ERROR: Error: Transaction cannot be rolled back because it has been finished with state: rollback
    at Transaction.rollback (example.com/node_modules/sequelize/lib/transaction.js:158:11)
    at example.com/node_modules/sequelize/lib/sequelize.js:1216:23
From previous event:
    at Promise.catch (example.com/node_modules/sequelize/lib/promise.js:21:17)
    at transactionResolver (example.com/node_modules/sequelize/lib/sequelize.js:1211:15)
From previous event:
    at Sequelize.transaction (example.com/node_modules/sequelize/lib/sequelize.js:1227:12)
    at example.com/database/models/user/permissions/permission.js:35:19
    at next (native)
    at tryCatcher (example.com/node_modules/bluebird/js/release/util.js:11:23)
    at PromiseSpawn._promiseFulfilled (example.com/node_modules/bluebird/js/release/generators.js:90:49)
    at Promise._settlePromise (example.com/node_modules/bluebird/js/release/promise.js:550:26)
    at Promise._settlePromise0 (example.com/node_modules/bluebird/js/release/promise.js:590:10)
    at Promise._settlePromises (example.com/node_modules/bluebird/js/release/promise.js:673:18)
    at Async._drainQueue (example.com/node_modules/bluebird/js/release/async.js:125:16)
    at Async._drainQueues (example.com/node_modules/bluebird/js/release/async.js:135:10)
    at Immediate.Async.drainQueues [as _onImmediate] (example.com/node_modules/bluebird/js/release/async.js:16:14)
    at processImmediate [as _immediateCallback] (timers.js:383:17)
```

Where the code being executed in permission.js is a script to initialize permissions in the database. It uses a transaction. 

I removed that transaction entirely, and it just failed with the same message on the next transaction in the code to be executed. Same after that. Any transaction I run produces that error now.

I have a sneaking suspicion it's Babel that is messing it up, but I haven't been able to determine it yet. I use `sequelize.transaction(async (t) => {/*Use transaction*/})` in conjunction with the bluebird coroutines transformer on Babel 5, and it might be messing up that async function.

If you have any thoughts on what could lead to an error like that, please let me know. I can't do any work until I fix this.

[...]

PR #5043, opened 2015-12-15
Don't rollback twice on connection error. Fixes issue #4850.

spitimage:
The rollbacks are getting called twice when a connection error occurs during an autofunc version of `sequelize.transaction(f)` call. This simple fix just rejects with the original connection error.

### Test Case
This should fail before the error and __pass__ after the fix. I forced a connection error by specifying a bogus database.

```js
'use strict';

/* jshint -W030 */
var chai = require('chai')
  , expect = chai.expect
  , Support = require('../support')
  , current = Support.sequelize;

if (current.dialect.supports.transactions) {
  describe(Support.getTestDialectTeaser('Sequelize#transaction.autoFuncErrors'), function() {

  describe('then', function() {

    // This failure is caused by any connection error (I used a bogus database name)
    it('bypasses all promise error-handling and produces uncaught exception', function(done) {

      // None of the following paths gets hit
      try {
        return this.sequelize
        // This call fails and throws an exception:
        // Error: Transaction cannot be rolled back because it has been finished with state: rollback

          .transaction(function (trans) {
            expect(false).to.be.ok;
          })
          .catch(function (err) {
            expect(true).to.be.ok;
          })
          .finally(function () {
            expect(true).to.be.ok;
            done();
          });
      } catch(e){
        expect(false).to.be.ok;
        done();
      }
    });

  });

});

}
```

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -1239,6 +1239,9 @@ Sequelize.prototype.transaction = function(options, autoCallback) {
         // If an error occured while commiting a transaction do not try to rollback the transaction
         if (transaction.finished === 'commit') {
           reject(err);
+        } else if (transaction.finished === 'rollback'){
+          // Already in rollback state. Reject with the original error.
+          reject(err);
         } else {
           transaction.rollback().finally(function () {
             reject(err);
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -1236,11 +1236,8 @@ Sequelize.prototype.transaction = function(options, autoCallback) {
           });
         });
       }).catch(function(err) {
-        // If an error occured while commiting a transaction do not try to rollback the transaction
-        if (transaction.finished === 'commit') {
-          reject(err);
-        } else if (transaction.finished === 'rollback'){
-          // Already in rollback state. Reject with the original error.
+        // If the transaction has already finished (commit, rollback, etc), reject with the original error
+        if (transaction.finished) {
           reject(err);
         } else {
           transaction.rollback().finally(function () {


spitimage:
Linked with #4850

lib/sequelize.js
```diff
@@ -1239,6 +1239,9 @@ Sequelize.prototype.transaction = function(options, autoCallback) {
         // If an error occured while commiting a transaction do not try to rollback the transaction
         if (transaction.finished === 'commit') {
           reject(err);
+        } else if (transaction.finished === 'rollback'){
+          // Already in rollback state. Reject with the original error.
+          reject(err);
```
mickhansen: Maybe just `transaction.finished` then?

spitimage: Sounds fine to me. I didn't know if there may be other transaction states to handle. 

Should I update the PR?

mickhansen: Please do :)

mickhansen:
Hmm, when does this issue happen exactly? Double rollback shouldn't happen unless the user rollbacks inside

spitimage:
I think the transaction rollback state is set internally when `transaction.prepareEnvironment()` fails [here](https://github.com/sequelize/sequelize/blob/master/lib/sequelize.js#L1225). The rollback happens [here](https://github.com/sequelize/sequelize/blob/master/lib/transaction.js#L195) and ultimately [here](https://github.com/sequelize/sequelize/blob/master/lib/transaction.js#L171). 

So when the top-level `sequelize.transaction()` call is made, the transaction is already in rollback state.

mickhansen:
Ah, you are right, i forgot that prepareEnvironment will also attempt to rollback.

spitimage:
Added to PR

Issue #5001, opened 2015-12-08
sequlize does not support timezones from momentjs


PavelPolyakov:
It's stated here: http://docs.sequelizejs.com/en/latest/api/sequelize/, that it's possible to pass the timezone in the format which is accepted by momentjs.

However this code:
```javascript
let sequelize = new Sequelize(config.get('storage.uri'), {
    logging: debug, //false

    define: {
        timestamps: false
    },

    timezone: "Europe/Berlin",

    pool: {
        max: 20,
        min: 0,
        idle: 5000
    }
});
```

Results to the next error:
![image](https://cloud.githubusercontent.com/assets/839290/11662023/0ddce968-9dd5-11e5-82f0-0d06c22c46c5.png)

This code works:
```javascript
let sequelize = new Sequelize(config.get('storage.uri'), {
    logging: debug, //false

    define: {
        timestamps: false
    },

    timezone: "+01:00",

    pool: {
        max: 20,
        min: 0,
        idle: 5000
    }
});
```

Regards,

janmeier:
Duplicate of https://github.com/sequelize/sequelize/issues/4307

PR #5044, opened 2015-12-15
Postgres named timezone issue

### LINKED ISSUE #4307: Postgres does not support named timezones
I'm using a raw sequelize query in node to execute a postgresql function which returns a query. The problem I'm having is that the result set obtained from running the query in pgAdmin does not match the result set returned in node when using the .query() method, even though the query in pgAdmin and the query in node are both exactly the same.

here's the query I'm performing:

SELECT * FROM foo('2015-07-01','2015-07-31');
This query returns 91875 rows in pgAdmin, however, in node using sequelize, running the following only returns 87500 rows:

models.sequelize.query("SELECT * FROM foo('2015-07-01','2015-07-31');")
.spread(function (results, metadata) { 
    console.log(results.length);
});
output: 87500

The rows which are excluded correspond to the last date 2015-07-31.

I found that I can obtain all the rows in node by increasing the 2nd date parameter to go beyond the end date. For example, if my goal is to get results which include 2015-07-31, setting the end date to be 2015-08-03 works. Although, this is masking the problem and not ideal.

I thought there must be something wrong with the WHERE clause in my stored procedure function. However, if that was the case, why do I get back the desired number of rows when I run the query though pgAdmin?

sushantdhiman:
Fixes #4307 and related #5001

### DIFF
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -6,7 +6,8 @@ var AbstractConnectionManager = require('../abstract/connection-manager')
   , Promise = require('../../promise')
   , sequelizeErrors = require('../../errors')
   , semver = require('semver')
-  , dataTypes = require('../../data-types');
+  , dataTypes = require('../../data-types')
+  , moment = require('moment-timezone');
 
 ConnectionManager = function(dialect, sequelize) {
   AbstractConnectionManager.call(this, dialect, sequelize);
@@ -129,7 +130,12 @@ ConnectionManager.prototype.connect = function(config) {
     }
 
     if (!self.sequelize.config.keepDefaultTimezone) {
-      query += 'SET client_min_messages TO warning; SET TIME ZONE INTERVAL \'' + self.sequelize.options.timezone + '\' HOUR TO MINUTE;';
+      var isZone = !!moment.tz.zone(self.sequelize.options.timezone);
+      if (isZone) {
+        query += 'SET client_min_messages TO warning; SET TIME ZONE \'' + self.sequelize.options.timezone + '\'';
+      } else {
+        query += 'SET client_min_messages TO warning; SET TIME ZONE INTERVAL \'' + self.sequelize.options.timezone + '\' HOUR TO MINUTE;';
+      }
     }
 
     // oids for hstore and geometry are dynamic - so select them at connection time
--- a/test/integration/dialects/postgres/connection-manager.test.js
+++ b/test/integration/dialects/postgres/connection-manager.test.js
@@ -1,0 +1,28 @@
+'use strict';
+
+/* jshint -W030 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../../support')
+  , dialect = Support.getTestDialect()
+  , config = require(__dirname + '/../../../config/config')
+  , DataTypes = require(__dirname + '/../../../../lib/data-types')
+  , _ = require('lodash');
+
+if (dialect.match(/^postgres/)) {
+  describe('[POSTGRES Specific] Sequelize', function() {
+    it('should correctly parse the moment based timezone', function() {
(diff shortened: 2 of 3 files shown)


lib/dialects/postgres/connection-manager.js
```diff
@@ -129,7 +130,12 @@ ConnectionManager.prototype.connect = function(config) {
     }
 
     if (!self.sequelize.config.keepDefaultTimezone) {
-      query += 'SET client_min_messages TO warning; SET TIME ZONE INTERVAL \'' + self.sequelize.options.timezone + '\' HOUR TO MINUTE;';
+      var isZone = !!moment.tz.zone(self.sequelize.options.timezone);
```
mickhansen: `moment.tz.zone` validates that it's a valid timezone string ala `Europe/Berlin`?

test/integration/dialects/postgres/connection-manager.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+/* jshint -W030 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../../support')
+  , dialect = Support.getTestDialect()
+  , DataTypes = require(__dirname + '/../../../../lib/data-types')
+  , _ = require('lodash');
+
+if (dialect.match(/^postgres/)) {
```
mickhansen: `dialect === 'postgres'` should work afair

test/integration/dialects/postgres/connection-manager.test.js
```diff
@@ -0,0 +1,27 @@
...
+      var options = _.extend(this.sequelize.options, { timezone: 'Asia/Kolkata', timestamps: true });
+      var sequelize = Support.createSequelizeInstance(options);
+
+      var tzTable = sequelize.define('tz_table', { foo: DataTypes.STRING });
+      return tzTable.sync({force: true}).then(function() {
+        return tzTable.create({foo: 'test'}).then(function(row) {
+          expect(row).to.be.not.null;
+        }).catch(function(e) {
+          expect(e).to.be.null;
+        });
```
mickhansen: You can just remove the catch statement, the test will fail automatically if there's an error.

lib/dialects/postgres/connection-manager.js
sushantdhiman: Yes it does. It returns an `Object` when there is a valid zone otherwise `null`.

test/integration/dialects/postgres/connection-manager.test.js
sushantdhiman: Oh I was following conventions from https://github.com/sequelize/sequelize/blob/master/test/integration/dialects/postgres/associations.test.js#L11

sushantdhiman: Also I think its used here to check for `postgres` and `postgres-native` both. @mickhansen

mickhansen: @sushantdhiman Ah yeah that might make sense.

lib/dialects/postgres/connection-manager.js
mickhansen: Great

mickhansen:
LGTM, just want @janmeier to verify.
Could you also squash your commits please?

sushantdhiman:
Don't know how to squash them properly :( , I not much familiar with `commit --reset`.

mickhansen:
Interactive rebasing is the easiest: https://help.github.com/articles/about-git-rebase/
That lets you rebase on top of master and then select what commits to squash and what to pick and then write a new commit message afterwards.

sushantdhiman:
Squashed the commits. Any other changes?

mickhansen:
LGTM, just needs a sign off from @janmeier

janmeier:
Signed off, just needs a changelog entry,  then this is good to go

sushantdhiman:
Hi @mickhansen @janmeier , Added changelog entry.

mickhansen:
Thanks for your work again @sushantdhiman

PR #5045, opened 2015-12-15
MySQL Geometry throw unhandled exception when buffer is null

### LINKED ISSUE #4953: mysql GEOMETRY bug
https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/data-types.js#L64

`value.buffer()` maybe is null
```
node_modules/mysql/lib/protocol/Parser.js:82
        throw err;
        ^

TypeError: Cannot read property 'slice' of null
    at Object.GEOMETRY.parse.GEOMETRY.parse [as GEOMETRY] 
```
```
users:
    ID: {
      type: 'CHAR(36)',
      allowNull: false,
      defaultValue: '',
      primaryKey: true
    },
    groupID: {
      type: type: 'CHAR(36)',
      allowNull: false
    },
   
groups:
    ID: {
      type: 'CHAR(36)',
      allowNull: false,
      defaultValue: '',
      primaryKey: true
    },
    location: {
      type: DataTypes.GEOMETRY,
      allowNull: false
    }

db.users.belongsTo(db.groups,{foreignKey:'groupID',targetKey:'ID'});
```

query
```
 let userInfo = yield db.users.findAll({
    include: [db.groups]
  });
```

I temporarily fix it
```
    value = value.buffer();
    if (value == null) {
      return null;
    }

    value = value.slice(4);
```

sushantdhiman:
Fix #4953

### DIFF
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -61,8 +61,10 @@ module.exports = function (BaseTypes) {
 
   GEOMETRY.parse = GEOMETRY.prototype.parse = function(value) {
     value = value.buffer();
-    if (value === null) {
-       return null;
+
+    //MySQL doesn't support POINT EMPTY, https://dev.mysql.com/worklog/task/?id=2381
+    if(value === null){
+      return null;
     }
 
     // For some reason, discard the first 4 bytes
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -60,8 +60,13 @@ module.exports = function (BaseTypes) {
   });
 
   GEOMETRY.parse = GEOMETRY.prototype.parse = function(value) {
+    value = value.buffer();
+    if (value === null) {
+       return null;
+    }
+
     // For some reason, discard the first 4 bytes
-    value = value.buffer().slice(4);
+    value = value.slice(4);
 
     return wkx.Geometry.parse(value).toGeoJSON();
   };
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -63,7 +63,7 @@ module.exports = function (BaseTypes) {
     value = value.buffer();
 
     //MySQL doesn't support POINT EMPTY, https://dev.mysql.com/worklog/task/?id=2381
-    if(value === null){
+    if (value === null) {
       return null;
     }
 
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -273,4 +273,28 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
      // No dialects actually allow us to identify that we get an enum back..
     testFailure(Type);
   });
+
+  it('should parse an empty GEOMETRY field', function () {
+   var Type = new Sequelize.GEOMETRY();
+
+   if (current.dialect.supports.GEOMETRY) {
+     current.refreshTypes();
+
+     var User = current.define('user', { field: Type }, { timestamps: false });
+     var point = { type: "Point", coordinates: [] };
+
+     return current.sync({ force: true }).then(function () {
+       return User.create({
+          //insert a null GEOMETRY type
+          field: point
+        });
+      }).then(function () {
+        //This case throw unhandled exception
+        return User.findAll();
+      }).then(function(users){
+        //it contains the null GEOMETRY data
+        expect(users[0].field).to.be.null;
+      });
+   }
+ });
 });
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -292,11 +292,13 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
         //This case throw unhandled exception
         return User.findAll();
       }).then(function(users){
-        if (dialect === 'mysql') {
+        if (Support.dialectIsMySQL()) {
           // MySQL will return NULL, becuase they lack EMPTY geometry data support.
           expect(users[0].field).to.be.eql(null);
+        } else if (dialect === 'postgres' || dialect === 'postgres-native') {
+          //Empty Geometry data [0,0] as per https://trac.osgeo.org/postgis/ticket/1996
+          expect(users[0].field).to.be.deep.eql({ type: "Point", coordinates: [0,0] });
         } else {
-          //it contains the null GEOMETRY data
           expect(users[0].field).to.be.deep.eql(point);
         }
       });
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -292,8 +292,13 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
         //This case throw unhandled exception
         return User.findAll();
       }).then(function(users){
-        //it contains the null GEOMETRY data
-        expect(users[0].field).to.be.null;
+        if (dialect === 'mysql') {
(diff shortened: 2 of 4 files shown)


test/integration/data-types.test.js
```diff
@@ -274,3 +274,30 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
     testFailure(Type);
   });
 });
+
+describe(Support.getTestDialectTeaser('DataTypes') + " MySQL", function() {
```
mickhansen: `Support.getTestDialectTeaser` already appends mysql.
Would it be possible to simply have a general empty coordinates test for all dialects?

sushantdhiman: I am running this test for all dialects which supports `GEOMETRY`. Will wait for CI to finish to see if something goes wrong on other supported databases.

test/integration/data-types.test.js
```diff
@@ -274,3 +274,30 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
     testFailure(Type);
   });
 });
+
+describe(Support.getTestDialectTeaser('DataTypes'), function() {
```
mickhansen: Isn't this describe the same as the one on the top of the file? Might aswell just remove it.

sushantdhiman:
Hi @mickhansen , Its done. Let me know if there is anything else to improve.

test/integration/data-types.test.js
```diff
@@ -273,4 +273,29 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
...
+       return User.create({
+          field: { type: "Point", coordinates: [] }
+        });
+      }).then(function () {
+        return User.findAll();
+      }).then(function(users){
+        expect(users[0]).to.be.not.null;
+      });
+   }
+ });
```
mickhansen: This test needs to be expanded a bit.
Change the description to `should parse an empty GEOMETRY field`.

Add a comment at `field: { type: "Point", coordinates: [] }` about this generating a `NULL` insert for MySQL.
Add a verification on the fetched user that values are as expected.

sushantdhiman:
Hi @mickhansen , Changed tests and added changelog entry. I have rebased with latest `3.15` release as well. Can you please restart the CI, its not starting due to some reason.

mickhansen:
CI is queued atm it looks like

sushantdhiman:
Hmm... there are no other pull requests in queue. Others are finished. Both of my jobs blocked. Lets wait and see if they eventually timeouts.

sushantdhiman:
Hi @mickhansen , What should I return in case of null coordinates ?

`Postgres` returns ```{  type: 'Point',  coordinates: null }```
`MySQL` returns `null`

mickhansen:
Good question, might want @janmeier's opinion here.

janmeier:
I think always returning an object makes the most sense - that still encodes the type information, (Point / Polygon etc), even if the value is null

sushantdhiman:
Sorry tried to fixed it by editing files via Github , Didn't worked. I will add this change tomorrow.

sushantdhiman:
I have been working on this and I realized the `null` buffer returned by `node-mysql` is the issue behind it. `node-mysql` should return the valid `wkb` type buffer which contain information about GEOMETRY data even though its empty.

I have added an issue here https://github.com/felixge/node-mysql/issues/1310, @mickhansen If you want you can mark it blocking due to `node-mysql`. I will work on `node-mysql` for a solution.

sushantdhiman:
MySQL lacks support for `POINT EMPTY` as per this https://dev.mysql.com/worklog/task/?id=2381, This means any invalid / empty data is stored as `NULL` in the database. 

There is no way to know if that was `Point` or `LineString` or any other type of geometry data. Thus until MySQL implements `POINT EMPTY` we can't return `{ type: 'Point', coordinates: [] }`. 

If you want we can fix current issue by returning the `NULL` for now. When MySQL implements `POINT EMPTY` we can return the desired output `{ type: 'Point', coordinates: [] }`

cc @mickhansen @janmeier

mickhansen:
It should probably be null for mysql then, agree @janmeier?

sushantdhiman:
@janmeier ping

janmeier:
Yes

sushantdhiman:
Please restart the CI, two pull request are not started from last 30 min @janmeier

janmeier:
I haven't found the switch for restarting travis yet ;) - Just give it some time

sushantdhiman:
@janmeier All green, Let me know if PR need other changes.

lib/dialects/mysql/data-types.js
```diff
@@ -60,8 +60,15 @@ module.exports = function (BaseTypes) {
   });
 
   GEOMETRY.parse = GEOMETRY.prototype.parse = function(value) {
+    value = value.buffer();
+
+    //MySQL doesn't support POINT EMPTY, https://dev.mysql.com/worklog/task/?id=2381
+    if(value === null){
```
janmeier: Missing space here

(Really wish this could be handled by the linter @mickhansen , I've noticed you commenting on it more and more :)). JSHint can't do it, dunno if other linters can

test/integration/data-types.test.js
```diff
@@ -273,4 +273,35 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
...
+     return current.sync({ force: true }).then(function () {
+       return User.create({
+          //insert a null GEOMETRY type
+          field: point
+        });
+      }).then(function () {
+        //This case throw unhandled exception
+        return User.findAll();
+      }).then(function(users){
+        if (dialect === 'mysql' || dialect === 'mariadb') {
```
janmeier: `Support.dialectIsMySQL`

lib/dialects/mysql/data-types.js
mickhansen: Hmm, i think eslint does it, but that's for ES/Babel code

janmeier:
LGTM, just some tiny comments

lib/dialects/mysql/data-types.js
janmeier: But since es5 is a subset of es2015, that shouldn't be a problem :)

test/integration/data-types.test.js
sushantdhiman: will `Support.dialectIsMySQL` work for `mariadb` as well?

janmeier: Yes

mickhansen:
Just restarted the builds that failed. No reason to keep closing/opening the PR :)

sushantdhiman:
There is no restart button. Where I can find it ?

mickhansen:
Only me and Jan have access to it.

sushantdhiman:
Ok, Do you think we still need `io js v1` based tests ?  CI take too much time :disappointed:

mickhansen:
At the moment yes. When we publish 4.0 i'd like to move to only test against Node v4+

mickhansen:
A bigger problem is builds failing because of intermittent test errors.

sushantdhiman:
Yes this happens to me like everyday. Sometime CI not able to login to DB, sometime a query left running.

mickhansen:
Green now, will let @janmeier merge

sushantdhiman:
Thanks

janmeier:
Thanks @sushantdhiman :)

sushantdhiman:
Thanks @janmeier and @mickhansen

PR #5052, opened 2015-12-15
For sqlite, do not add 'DEFAULT NULL' for null default values.

### LINKED ISSUE #5051: Auto-appending "DEFAULT NULL" to sqlite queries result in incorrect table schemas
When using sqlite and Sequelize migrations to change a column (either rename, remove, or add), Sequelize backs up the table, creates a new table with the changes, then copies the data over, to get around the fact that [sqlite does not support table alterations](http://stackoverflow.com/questions/8442147/how-to-delete-or-add-column-in-sqlite).

To construct the `CREATE TABLE` query for the new table, Sequelize gets the schema of the existing table by using `PRAGMA INFO(tablename)`, then mapping those into the clauses of the `CREATE TABLE` query. As part of this, Sequelize attempts to append `DEFAULT NULL` for columns where the `defaultValue: null`.

For example: the returned object from the `PRAGMA INFO(tablename)` call:

```
...
name:
   { type: 'VARCHAR(255)',
     allowNull: true,
     defaultValue: null,
     primaryKey: false }
...
```

will be used to generate a query like:

```
'CREATE TABLE IF NOT EXISTS `rules` ( ... `name` VARCHAR(255) DEFAULT NULL ...)',
```

When this table is created, sqlite will have a 'NULL' default value for that column:

```
sqlite> PRAGMA TABLE_INFO(rules);
...
cid         name        type        notnull     dflt_value  pk
----------  ----------  ----------  ----------  ----------  ----------
3           name        VARCHAR(25  0           NULL        0
...
```

Then, subsequent `PRAGMA INFO(tablename)` calls will yield this object:

```
...
  name:
   { type: 'VARCHAR(255)',
     allowNull: true,
     defaultValue: 'NULL',
     primaryKey: false },
...
```

Note the incorrect `NULL` string literal. As you can imagine, causes all sorts of problems downstream, for example, subsequent `INSERT`s into the table populating the field value with the `NULL` string literal.

jasonku:
Fixes #5051.

### DIFF
--- a/lib/dialects/abstract/index.js
+++ b/lib/dialects/abstract/index.js
@@ -38,6 +38,9 @@ AbstractDialect.prototype.supports = {
   updateOnDuplicate: false,
   schemas: false,
   transactions: true,
+  transactionOptions: {
+    type: false
+  },
   migrations: true,
   upserts: true,
   constraints: {
--- a/lib/dialects/sqlite/index.js
+++ b/lib/dialects/sqlite/index.js
@@ -25,6 +25,9 @@ SqliteDialect.prototype.supports = _.merge(_.cloneDeep(Abstract.prototype.suppor
   index: {
     using: false
   },
+  transactionOptions: {
+    type: true
+  },
   joinTableDependent: false,
   groupedLimit: false,
   ignoreDuplicates: ' OR IGNORE'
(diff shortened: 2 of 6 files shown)


jasonku:
Replaced by https://github.com/sequelize/sequelize/pull/5061

PR #5058, opened 2015-12-16
Update() modifies the supplied values

### LINKED ISSUE #4520: 'fields' option modifies passed object
Suppose we have 
```
var obj = {a: 1, b: 2, c: 3}
```
and we want to update some model with it but pick only specific fields, and then we want to update other model with the same object, but with different fields:
```
some_model.update(obj, {fields: ['a', 'b']});

other_model.update(obj, {fields: ['c']}) // Error: column 'c' cannot be null

console.log(obj); // {a: 1, b: 2};
```

What happened? - First update function picked fields `a` and `b`, but instead of working with a clone of `obj`, it modified it.

Is it intended behaviour? The obvious workaround would be to clone `obj` manually before passing it to update function, but I think this is not intuitive.

sushantdhiman:
Fixes #4520 , `values` passed to the `update` method got modified.

As per #4520 whenever we pass the `fields` in options, the caller `values` got modified. It has been fixed. 

While writing tests I discovered `update()` method also add the `updatedAt` key to the caller `values` object. I have added tests for both cases.

I have used cloned version of `values` to fix both issues.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# Future
+- [FIXED] calling Model.update() modifies passed values  [#4520](https://github.com/sequelize/sequelize/issues/4520)
+
 # 3.15.0
 - [ADDED] Improve support for pg range type to handle unbound ranges, +/-infinity bounds and empty ranges
 - [FIXED] Postgres issue when using named timezone  [#4307](https://github.com/sequelize/sequelize/issues/4307)
--- a/lib/model.js
+++ b/lib/model.js
@@ -2403,6 +2403,9 @@ Model.prototype.update = function(values, options) {
 
   Model.$injectScope(this.$scope, options);
 
+  // Clone values so it doesn't get modified for caller scope
+  values = _.clone(values);
+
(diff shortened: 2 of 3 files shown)


mickhansen:
This might be better off as a unit test. No need to hit the database to verify whether or not something has been dirtied.

sushantdhiman:
Changed tests to unit tests. They wont hit the database now. Please do let me know if there is anything else to improve

mickhansen:
Looks great. I can't check CI though since travis is giving me an SSL error atm.

sushantdhiman:
Yeah some association related test failed on one build. Please check it whenever CI opens for you.

sushantdhiman:
CI Cleared :) You can merge now

mickhansen:
@sushantdhiman Great :) can you squash your commits to a single descriptive commit?

mickhansen:
(and please refer to issues with # so it becomes `fix(#4520): clone values passed to Model.update`)

sushantdhiman:
@mickhansen ping

PR #5061, opened 2015-12-17
5051 fix sqlite default value

jasonku:
Addresses #5051

### DIFF
--- a/test/integration/query-interface.test.js
+++ b/test/integration/query-interface.test.js
@@ -150,6 +150,10 @@ describe(Support.getTestDialectTeaser('QueryInterface'), function() {
       var self = this;
       var Users = self.sequelize.define('_Users', {
         username: DataTypes.STRING,
+        city: {
+            type: DataTypes.STRING,
+            defaultValue: null
+        },
         isAdmin: DataTypes.BOOLEAN,
         enumVals: DataTypes.ENUM('hello', 'world')
       }, { freezeTableName: true });
@@ -161,6 +165,7 @@ describe(Support.getTestDialectTeaser('QueryInterface'), function() {
 
           var id = metadata.id;
           var username = metadata.username;
+          var city = metadata.city;
           var isAdmin = metadata.isAdmin;
           var enumVals = metadata.enumVals;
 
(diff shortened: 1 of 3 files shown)


test/integration/query-interface.test.js
```diff
@@ -177,7 +182,20 @@ describe(Support.getTestDialectTeaser('QueryInterface'), function() {
           }
           expect(username.type).to.equal(assertVal);
           expect(username.allowNull).to.be.true;
-          expect(username.defaultValue).to.be.null;
+
+          switch (dialect) {
+            case 'sqlite':
+              expect(username.defaultValue).to.be.undefined;
```
mickhansen: Should it not be undefined for all dialects?

test/integration/query-interface.test.js
```diff
@@ -177,7 +182,20 @@ describe(Support.getTestDialectTeaser('QueryInterface'), function() {
...
+              break;
+            default:
+              expect(username.defaultValue).to.be.null;
+          }
+
+          switch (dialect) {
+            case 'sqlite':
+              expect(city.defaultValue).to.be.null;
```
mickhansen: Should it not be null for all dialects?

test/integration/query-interface.test.js
```diff
@@ -190,7 +208,13 @@ describe(Support.getTestDialectTeaser('QueryInterface'), function() {
           }
           expect(isAdmin.type).to.equal(assertVal);
           expect(isAdmin.allowNull).to.be.true;
-          expect(isAdmin.defaultValue).to.be.null;
+          switch (dialect) {
+            case 'sqlite':
+              expect(isAdmin.defaultValue).to.be.undefined;
```
mickhansen: Should it not be undefined for all dialects?

jasonku: Should it be? I haven't explored this behavior with other dialects and I don't have tests set up to run other dialects' tests, so I don't really feel comfortable making changes for other dialects other than sqlite.

mickhansen: I think it should, no dialects should report a null default value if there in fact is no default value, null is special.
I'll merge your sqlite fix for now in any case.

Issue #4933, opened 2015-11-27
Same included model in already included model not working


tinganho:
I have an application very similar to Facebook, You have post and underneath a post you have comments. Then there is some metadata about how to format the links in the post's status text and the comment's status text.

I discovered today that including the same model for link formatting in both the post and comment model generates the wrong SQL:

```typescript
let includeClause = [
    {
        model: Comment,
        limit: 20,
        include: [
            urlTextFormattingInclude, 
            usernameTextFormattingInclude
        ]
    },
    {
        model: PostImage,
    },
    urlTextFormattingInclude, 
    usernameTextFormattingInclude
];
let posts = await Post.findAll({ 
    where: whereClause,
    order: 'id DESC',
    limit: limit,
    include: includeClause,
    attributes: (Object.keys((Post as any).attributes) as any).concat([
        [DbContext.literal('(SELECT COUNT(*) FROM "postLikes" WHERE "postId" = post.id)'), 'likesCount'],
        [DbContext.literal('(SELECT COUNT(*) FROM comments WHERE "postId" = post.id)'), 'commentsCount']
    ])
});
```

And it generates the following SQL:
```sql
SELECT "comment".*, 
       "urltextformattings"."id"                 AS "urlTextFormattings.id", 
       "urltextformattings"."displayurl"         AS 
       "urlTextFormattings.displayUrl", 
       "urltextformattings"."rawurl"             AS "urlTextFormattings.rawUrl", 
       "urltextformattings"."start"              AS "urlTextFormattings.start", 
       "urltextformattings"."end"                AS "urlTextFormattings.end", 
       "usernametextformattings"."id"            AS "usernameTextFormattings.id" 
       , 
       "usernametextformattings"."start"         AS 
       "usernameTextFormattings.start", 
       "usernametextformattings"."end"           AS 
       "usernameTextFormattings.end", 
       "usernameTextFormattings.user"."id"       AS 
       "usernameTextFormattings.user.id", 
       "usernameTextFormattings.user"."username" AS 
       "usernameTextFormattings.user.username" 
FROM   ((SELECT "id", 
                "text", 
                "createdat", 
                "updatedat", 
                "userid", 
                "postid" 
         FROM   "comments" AS "comment" 
         WHERE  "comment"."postid" = 5 
         LIMIT  20) 
        UNION ALL 
        (SELECT "id", 
                "text", 
                "createdat", 
                "updatedat", 
                "userid", 
                "postid" 
[...]

PR #5062, opened 2015-12-17
Find with attributes and eager loading test error

elhoyos:
This PR looks forward to create evidence of an error with finding while eager loading some relations and specifying attributes.

I found the issue #4933 very likely to be referenced in the test.


mickhansen:
#4933 is about `include.separate` so not strictly related

mickhansen:
Looks like you've found a bug eitherway though, so thanks! :)

PR #5065, opened 2015-12-17
Set updatedAt and createdAt values before validation

rafis:
If model has `updatedAt` or `createdAt` fields without allowing nulls:

```javascript
        createdAt: {
            type: DataTypes.DATE,
            allowNull: false,
            field: 'created_at'
        },
        updatedAt: {
            type: DataTypes.DATE,
            allowNull: false,
            field: 'update_at'
        },
```

then I'm receiving this error:

> { [SequelizeValidationError: notNull Violation: createdAt cannot be null,
notNull Violation: updatedAt cannot be null]
  message: 'notNull Violation: createdAt cannot be null,\nnotNull Violation: updatedAt cannot be null',
  errors:
   [ { message: 'createdAt cannot be null',
       type: 'notNull Violation',
       path: 'createdAt',
       value: null },
     { message: 'updatedAt cannot be null',
       type: 'notNull Violation',
       path: 'updatedAt',
       value: null } ] }

My proposal is to set `updatedAt` and `createdAt` values before doing validation.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -526,7 +526,7 @@ Instance.prototype.save = function(options) {
     , hook = self.isNewRecord ? 'Create' : 'Update'
     , wasNewRecord = this.isNewRecord;
 
-  if (updatedAtAttr && options.fields.indexOf(updatedAtAttr) === -1) {
+  if (updatedAtAttr && options.fields.length && options.fields.indexOf(updatedAtAttr) === -1) {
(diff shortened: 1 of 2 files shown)


mickhansen:
Seems like a reasonable change, however absolutely everything all builds are failing.

rafis:
Haven't time to learn internals of `Instance.prototype.save`, so I have just made adjusments for passing tests.

PR #5069, opened 2015-12-17
Return null to tell Bluebird non-return is intentional

scragg0x:


mickhansen:
I don't believe it's needed for `new Promise` since it's not actually inside a `then` or chain otherwise.

scragg0x:
hmm, I still see a bunch of these.  Are you saying it will go away once Bluebird is updated on sequelize?

mickhansen:
@scragg0x If they are originated from the sequelize codebase it needs to be updated in Sequelize, i just don't think the line you changed has any effect since it's not a promsie chain callback, its a `new Promise` call which would ignore its return value.

mickhansen:
Hmm interesting, well can't hurt i guess.

PR #5072, opened 2015-12-18
Disables setting AUTOCOMMIT for Postgres >= 9.4.0

tkalfigo:
Fixes issue https://github.com/sequelize/sequelize/issues/4631
User needs to supply "databaseVersion" key in options object
when instantiating the Sequelize class in case version is not semver
compliant e.g. "9.5.0beta2" should be passed as "9.5.0"

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -877,8 +877,14 @@ var QueryGenerator = {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


tkalfigo:
@mickhansen I believe it should be ok now.

mickhansen:
Thanks @tkalfigo

PR #5080, opened 2015-12-21
fix query which has nested include in a parent include with separate true and has-many association

jayprakash1:
I have a query which is using including a has-many association with limit and separate attribute as true. It works fine till the above include has one level of included models. However, it does not work with more than one level of included models. 

I have added a fix and test based on my limited understanding. Please review the changes.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1630,7 +1630,7 @@ var QueryGenerator = {
       this.quoteIdentifier(fieldLeft)
     ].join('.');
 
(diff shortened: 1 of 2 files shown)


mickhansen:
The test fails without the test yes? Looks good, can you squash your commits and remove the merge commit?

jayprakash1:
Yes, the test fails without the changes above. Done other things as asked.

mickhansen:
the commit message is "some commit for now", mind changing that to be something descriptive? like the current description.

jayprakash1:
Changed the commit message. Thank you.

mickhansen:
Thank you for your patience and this fix @jayprakash1, good work :)

PR #5081, opened 2015-12-21
Fix(#4702): instance returns this on all paths

### LINKED ISSUE #4702: instance.set('id', 1) returns undefined
Seems like setting a value of a field that is a primary key returns `undefined`. Consider this snippet:

```javascript
Foo
  .findOne()
  .then( (foo) -> foo.set('id', 100).save() )
```

It throws `TypeError: Cannot read property 'save' of undefined`. Instead of this I would expect a meaningful error to be thrown when I try to save the instance.

Note: It actually doesn't matter if the field name is `id` or what it's type is. The fact that it is a primary key is relevant.

I'm using v. 3.12.0 with `sqlite` dialect.

sushantdhiman:
Fixes #4702

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
 # Future
 - [FIXED] calling Model.update() modifies passed values  [#4520](https://github.com/sequelize/sequelize/issues/4520)
+- [FIXED] Instance can be chained on .set() and other methods [#4702](https://github.com/sequelize/sequelize/issues/4702)
 
 # 3.15.0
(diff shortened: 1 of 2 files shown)


sushantdhiman:
ping @mickhansen

mickhansen:
Thanks @sushantdhiman. I won't be too responsive on github over the holidays i'm afraid :)

sushantdhiman:
:santa: :smile: No problem @mickhansen , we can work after holidays.

PR #5089, opened 2015-12-22
fix: nested create: belongsTo with foreignKey

wilzbach:
Hey,

nested creates ([#3386](https://github.com/sequelize/sequelize/pull/3386/files)) are awesome! 

I played around a bit with them and found that when using `belongsTo` relations and a foreignKey constraint (e.g. `allowNull: false`), I get a validation error as the source model is validated _before_ the target model is created.

Hence I propose to move the nested `belongsTo` before the validation -> about 60 lines up.
It shouldn't make a difference as the targeted model was already created before and with transaction there is rollback support. However we could define a default transaction that could be used for this behavior if this is wanted.

Moreover I added some tests to show my problem and check for this in the future ;-)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,5 @@
 # Future
+- [FIXED] nested belongsTo create with foreignKey constraints [#5089](https://github.com/sequelize/sequelize/pull/5089)
 - [FIXED] calling Model.update() modifies passed values  [#4520](https://github.com/sequelize/sequelize/issues/4520)
 
 # 3.15.0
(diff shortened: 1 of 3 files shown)


mickhansen:
Since we don't automatically create a transaction i think we should instead iterate over belongsTo relations and add their foreign keys to a skip list for the validator instead of creating them in advance - Less chance of leaving bad instances in the DB.

janmeier:
ping @wilzbach

janmeier:

Issue #3769, opened 2015-05-20
Raw queries: multiple values in replacement


kirill-zhirnov:
``` coffee
    db.query "
        delete from characteristic where id in (:list)
    ", null, {
        replacements :
            list : [1,2,3]
    }
```

Will execute query:

``` sql
delete from characteristic where id in (ARRAY[1,2,3])
```

Is it possible to quote multiple values?

janmeier:
What do you mean? You can add multiple replacements by adding more parameters to the query if thats wht you mean?

kirill-zhirnov:
I need to put multiple values in "IN ()" condition. I can use more parameters, e.g. "IN (?,?,?)", but in this case I need also to generate SQL query:

```coffeescript
values = ['a', 'b', 'c']

sql = []
for value in values
	sql.push '?'

db.query "
        delete from characteristic where id in (#{sql.join(',')})
    ", null, {
	replacements : values
}
```
It is not really convenient. What would be nice:

```coffeescript
    db.query "
        delete from characteristic where id in (:list)
    ", null, {
        replacements :
            list : [1,2,3]
    }
```

and in the end have SQL:
```coffeescript
delete from characteristic where id in (1,2,3)
```

mickhansen:
I was sure we had fixed this bug but it keeps popping up :p

BridgeAR:
@kirill-zhirnov What version do you use? You're using an older version.

majodev:
Just check the newest version 3.2.0 - same problem.

blakmatrix:
can confirm for `3.2.0`

ChaliceG:
Issue is still present in 3.6.0

adonomay:
+1

toddbluhm:
+1

zackexplosion:
+1

leibale:
+1

nazar:
+eleventy

ncwhale:
Still got this problem, when try to query with `IN ($1::uuid[])`

steaks:
+1

felixfbecker:
This will not work with bound parameters. You can use the ANY operator to accomplish this, it will bind the (single) parameter to the array.

Please do not +1 issues, it clutters the thread and adds no value to the discussion. Use GitHub reactions.

amarewgs:
+1

amarewgs:
this works for me

`
var roleIds = roles.map(role => role.id);
[...]

Issue #1969, opened 2014-06-26
Failure when substituting array into raw query


bblack:
Using postgres:
```
dbconn.query('select * from "User" where "userId" in :ids', null, {raw: true}, {ids: [1,2,3]}).then(function(users){ console.log(users.length); })
[object Object]
> info: Executing (default): select * from "User" where "userId" in ARRAY[1,2,3]
Possibly unhandled error: syntax error at or near "ARRAY"
```

defunctzombie:
@mickhansen any way around this?

mickhansen:
@defunctzombie I don't believe so, if it's still an issue. An argument could be made to change the default array to parameter escaping since IN is probably most likely the desired effect.

defunctzombie:
No longer using raw query for my use case so not an issue for me, but I do think that the default behavior of `IN` is likely desired.

mickhansen:
@defunctzombie I agree, did you ever test if it was still the case? I was sure i fixed this previously.

defunctzombie:
have not tested if this is the case recently since switching away from the raw query

DenisGorbachev:
I'm having the same issue

janmeier:
@DenisGorbachev which version are you experiencing this on?

DenisGorbachev:
"sequelize": "2.1.0"

juanjoLenero:
Just chiming in, I'm also having this issue in sequelize v2.1.3.

```javascript
...
if (options.trunc) {
    dates = Calendar.truncTimeFromDates(dates);
    startDateField = sequelize.fn('date_trunc', 'day', startDateField);
}

return Event.findAll({
    where: {
        $and: [
            {parentEventId: this.id},
            sequelize.where(
                startDateField,
                'IN',
                dates
            )
         ]
    },
    logging: console.log
});

// Executing (default): 
// SELECT "id", "title", "location", "timeZone", "startDate", "endDate", "notes", "rrule", "type", "isDetached", "isCanceled", "isAllDay", "createdAt", "updatedAt", "courseId", "parentEventId", "ownerId" 
// FROM "Events" AS "Event" 
// WHERE ("Event"."parentEventId" = 1 AND date_trunc('day', "Event"."startDate") = ARRAY[1420005600000]);
```

mull:
Any news on this? I just ran into it as well.

diegolmello:
+1

sushantdhiman:
@mickhansen @janmeier Duplicate of #3769

PR #5091, opened 2015-12-22
Add in escapeValues option for raw queries

toddbluhm:
Ran into an issue when trying to execute a raw query with an IN statement where arrays get converted to `Array[]` in the query and so I decided to go around this by doing `array.join("', '")` but it turns out that all `'` are auto escaped without the option to disable that. This PR adds in the option `escapeValues` for named replacements in raw queries. The default value for this option is `true` for backwards compatibility, but setting to `false` will skip all value escaping so that you can pass single quotes in around values an not have to worry about them being escaped.

Also updated docs as I added in some method comments for this option.

```js
sql.query("SELECT * FROM User WHERE name IN (:names)", {
  replacements: {
    names: "'leia', 'luke', 'skywalker'" // or `'${array.join("', '")}'` with es6
  }
}, {
  escapeValues: false
});
// = SELECT * FROM User WHERE name IN ('leia', 'luke', 'skywalker');
```

Added test cases to demonstrate the new option. This PR should provided a work around for issues: #3769 and #1969.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -685,7 +685,7 @@ Sequelize.prototype.import = function(path) {
  * @param {Boolean}         [options.supportsSearchPath] If false do not prepend the query with the search_path (Postgres only)
  * @param {Object}          [options.mapToModel=false] Map returned fields to model's fields if `options.model` or `options.instance` is present. Mapping will occur before building the model instance.
  * @param {Object}          [options.fieldMap] Map returned fields to arbitrary names for `SELECT` query type.
- *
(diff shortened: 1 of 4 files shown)


mickhansen:
Mind removing changes to `docs/api/*` from the PR? It clutters the diff and we autogenerate all that from jsdoc anyways.

toddbluhm:
Done. Just trying to follow the contrib guidelines. Let me know if you want me to update the changelog.md file.

mickhansen:
Just merged: https://github.com/sequelize/sequelize/pull/5032 which might fix this by reverting the regression

toddbluhm:
Indeed fixed by #5032

PR #5093, opened 2015-12-23
Fix #4856 : Attributes from multiple scopes doesn't merge

### LINKED ISSUE #4856: Scopes attributes are not merged
Version : **3.13.0**
Dialect : **mysql**

**Given**
 
```js
    username: {
        type: DataTypes.STRING,
        allowNull: false
    },
    about: {
        type: DataTypes.STRING
    },
    password: {
        type: DataTypes.STRING,
        allowNull: false,
        set: function (val) {
            // Some encrypting logic
        }
    }
}, {
    scopes: {
        'public': {
            attributes: [
                'id',
                'username',
                'about'
            ]
        },
        'private': {
            attributes: [
                'password'
            ]
        }
    }
});
```

**When** querying an instance with multiple scopes, scopes attributes do not get merged. 
e.g.

```js
User.scope('public','private').findOneById(1).then(function(user){
    console.log(user.about); // Exists, attribute included from public scope
    console.log(user.password); // Undefined, public and private attributes were not merged
});
```

sushantdhiman:
Fixes #4856

### DIFF
--- a/test/integration/model/scope.test.js
+++ b/test/integration/model/scope.test.js
@@ -1,0 +1,54 @@
+'use strict';
+
+/* jshint -W030 */
+/* jshint -W110 */
+var chai = require('chai')
+  , Sequelize = require('../../../index')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../support');
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+  describe('scope', function () {
+    beforeEach(function () {
+      this.ScopeMe = this.sequelize.define('ScopeMe', {
+        username: Sequelize.STRING,
+        email: Sequelize.STRING,
+        access_level: Sequelize.INTEGER,
+        other_value: Sequelize.INTEGER
+      }, {
+         scopes: {
+           lowAccess: {
+             attributes: ['other_value', 'access_level'],
+             where: {
+               access_level: {
+                 lte: 5
+               }
+             }
+           },
+           withName: {
+             attributes: ['username']
+           }
+         }
+      });
+
+      return this.sequelize.sync({force: true}).then(function() {
+        var records = [
+          {username: 'tony', email: 'tony@sequelizejs.com', access_level: 3, other_value: 7},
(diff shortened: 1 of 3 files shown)


test/integration/model/scope.test.js
```diff
@@ -0,0 +1,78 @@
...
+                expect(record.username).to.exist;
+                expect(record.access_level).to.exist;
+              });
+    });
+    it('should be able to merge attributes as object', function () {
+      return this.ScopeMe.scope('lowAccessObj','withNameObj').findOne()
+              .then(function(record){
+                expect(record.other_value).to.exist;
+                expect(record.username).to.exist;
+              });
```
mickhansen: what about `access_level` in this case?

sushantdhiman: `access_level` is removed because scope `withNameObj` have `excluded` it.

mickhansen: please include that in the test then :)

mickhansen:
Will need a rebase aswell.

sushantdhiman:
It appears to be including the `access_level` apparently due to its presence in one of the scopes. What should I do in such case? Remove it or keep it?

sushantdhiman:
Here is the implementation details @mickhansen @janmeier , Need your comments.

When dealing with ambiguous `include` and `exclude` parameters. We have two choices

* Give preference to `exclude` parameters
* Give preference to `include` parameters

Current PR is designed to give preference to `include` parameters. This means if one scope try to `include` and attribute and other scope try to `exclude` it, we will choose to `include` it in final result.

Let me know what do you think about it :)

mickhansen:
@sushantdhiman Can we start with a simple implementation that simply merges attributes when they are arrays? That should solve the immediate problem for most users, after that we can come up with a design for handling include/exclude.

sushantdhiman:
@mickhansen , it solves that problem already

mickhansen:
@sushantdhiman This PR already solves it or it's already solved? If this PR solves it just change the tests/implementation to do pure array concat implementation for now.

sushantdhiman:
@mickhansen Done

PR #5096, opened 2015-12-23
fixes #5083 by passing full config options to replication

### LINKED ISSUE #5083: Abstract Connection Manager does not pass dialectOptions on replication scenario
Hello,

When using the replication read / write feature of Sequelize, the actual dialectOptions will be lost [@line 116](https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/connection-manager.js#L116)

The dialectOptions will be passed when no replication is used, due to the inherit usage of the entire config object [@line 90](https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/connection-manager.js#L90)

This feels inconsistent, can I make a pull request to fix this ?

Best Regards,
Khez

Khez:


mickhansen:
Thanks @Khez

PR #5090, opened 2015-12-22
Fix #4986

sushantdhiman:
Fix #4986

PR #5106, opened 2015-12-28
Do not inject include twice, expand and validate include in aggregate

### LINKED ISSUE #4986: findAndCountAll() not working with scope
dialect: postgres
sequelize version: 3.14.2

```js
var Project = sequelize.define('Project', {
  title: Sequelize.STRING,
  description: Sequelize.TEXT
})

var Task = sequelize.define('Task', {
  title: Sequelize.STRING,
  description: Sequelize.TEXT,
  deadline: Sequelize.DATE
})

Task.belongsTo(Project);
Task.addScope('includeProject', { include: [{ model: Project }] });
```
running `Task.scope('includeProject').findAndCountAll()` throws this error:
```js
TypeError: Cannot read property 'association' of undefined
    at generateJoinQueries (/Users/pye/Repositories/models/node_modules/sequelize/lib/dialects/abstract/query-generator.js:1122:42)
    at Object.<anonymous> (/Users/pye/Repositories/models/node_modules/sequelize/lib/dialects/abstract/query-generator.js:1411:27)
    at Array.forEach (native)
    at Object.QueryGenerator.selectQuery (/Users/pye/Repositories/models/node_modules/sequelize/lib/dialects/abstract/query-generator.js:1410:10)
    at QueryInterface.rawSelect (/Users/pye/Repositories/models/node_modules/sequelize/lib/query-interface.js:691:33)
    at Model.aggregate (/Users/pye/Repositories/models/node_modules/sequelize/lib/model.js:1535:30)
    at Model.count (/Users/pye/Repositories/models/node_modules/sequelize/lib/model.js:1577:15)
    at Model.findAndCount (/Users/pye/Repositories/models/node_modules/sequelize/lib/model.js:1650:15)
    at repl:1:38
    at REPLServer.defaultEval (repl.js:248:27)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:412:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:210:10)
```

Update: In closer look into the code, I found the issue in `$injectScope()` function of sequelize's `model.js`. (Line https://github.com/sequelize/sequelize/blob/master/lib/model.js#L2619), where scopeInclude as adding a duplicate scopes when `findAndCountAll` is called
`return item.as === scopeInclude.as;`

So the workaround currently for me is to provide `as` in the scope like this:
`Task.addScope('includeProject', { include: [{ model: Project, as: 'Project' }] });`

### LINKED ISSUE #5121: How to find distinct rows in a table including associations
Hi,

I'm trying to query for distinct values of a column with a where clause that references associations.

For example:
```
SomeModel.aggregate('some_col', 'DISTINCT', {
  plain: false,
  include: {
    model: SomeOtherModel,
    where: { foo: true },
    include: {
      model: YetAnotherModel,
      where: { bar: false }
    }
  }
})
```

This kind of query produces the following error: `Cannot read property 'association' of undefined`

Is there another way to achieve this? Or some way to fix the above example?

Thanks!

Verdier:
Refactor `$injectScope` to better checks existence of include in `options.include` based on the related `association`. This PR also fix a bug where include are not validated/expanded in `model.aggregate`.

Fix #4986
Fix #5121
Fix `Cannot read property 'association' of undefined`

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2608,6 +2608,10 @@ Model.prototype.$expandAttributes = function (options) {
 
 // Inject current scope into options. Includes should have been conformed (conformOptions) before calling this
 Model.$injectScope = function (scope, options) {
+  // Do not inject scope more than once.
+  if (options._scopeInjected) return;
+  options._scopeInjected = true;
+
   scope = optClone(scope);
 
   var filteredScope = _.omit(scope, 'include'); // Includes need special treatment
--- a/lib/model.js
+++ b/lib/model.js
@@ -236,21 +236,23 @@ function conformOptions(options, self) {
     self.$expandAttributes(options);
   }
 
-  if (options.include) {
-    // if include is not an array, wrap in an array
-    if (!Array.isArray(options.include)) {
-      options.include = [options.include];
-    }
-
-    // convert all included elements to { model: Model } form
-    options.include = options.include.map(function(include) {
-      return conformInclude(include, self);
-    });
-
-    if (!options.include.length) {
-      delete options.include;
-    }
+  if (!options.include) {
+    return;
   }
+  // if include is not an array, wrap in an array
+  if (!Array.isArray(options.include)) {
+    options.include = [options.include];
+  } else if (!options.include.length) {
+    delete options.include;
+    return;
+  }
+
+  // convert all included elements to { model: Model } form
+  options.include = options.include.map(function(include) {
+    include = conformInclude(include, self);
+
+    return include;
+  });
 }
 
 Model.$conformOptions = conformOptions;
@@ -1510,6 +1512,11 @@ Model.prototype.aggregate = function(attribute, aggregateFunction, options) {
   options = Utils._.extend({ attributes: [] }, options || {});
   conformOptions(options, this);
   this.$injectScope(options);
+
+  if (options.include) {
+    expandIncludeAll.call(this, options);
+    validateIncludedElements.call(this, options);
+  }
 
   var attrOptions = this.rawAttributes[attribute]
     , field = attrOptions && attrOptions.field || attribute
--- a/lib/model.js
+++ b/lib/model.js
@@ -236,23 +236,21 @@ function conformOptions(options, self) {
     self.$expandAttributes(options);
   }
 
-  if (!options.include) {
-    return;
-  }
-  // if include is not an array, wrap in an array
-  if (!Array.isArray(options.include)) {
-    options.include = [options.include];
-  } else if (!options.include.length) {
-    delete options.include;
-    return;
-  }
+  if (options.include) {
+    // if include is not an array, wrap in an array
+    if (!Array.isArray(options.include)) {
+      options.include = [options.include];
+    }
 
-  // convert all included elements to { model: Model } form
-  options.include = options.include.map(function(include) {
-    include = conformInclude(include, self);
+    // convert all included elements to { model: Model } form
+    options.include = options.include.map(function(include) {
+      return conformInclude(include, self);
+    });
 
-    return include;
-  });
+    if (!options.include.length) {
+      delete options.include;
+    }
+  }
 }
 
 Model.$conformOptions = conformOptions;
@@ -601,7 +599,7 @@ validateIncludedElement = function(include, tableNames, options) {
     model = include.association.target.name === include.model.name ? include.association.target : include.association.source;
   }
 
-  Model.$injectScope(model.$scope, include);
+  model.$injectScope(include);
 
   // This check should happen after injecting the scope, since the scope may contain a .attributes
   if (!include.attributes) {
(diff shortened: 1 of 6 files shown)


Verdier:
@mickhansen could you review my pull request?

mickhansen:
Applying the same scope twice should ideally not fail.
Looks like PR might break using multiple scopes?

Verdier:
You are right @mickhansen applying scope twice should be neutral. I'll have a look.

Verdier:
@mickhansen I have found the problem, `$injectScope` badly check the presence of the included model into `options.include`: 

If `$injectScope` is called but `options.include[i].as` is defined and equal to the default value of the related association, the implementation does not detect that model are same. This is fixed by checking the related `association.as`. 

To do that, I have made `$injectScope` an instance method.

Verdier:
@mickhansen every thing is green. I think it's ready to be merged.

mickhansen:
Gonna need @janmeier to take a look on this one.

Verdier:
@janmeier @sushantdhiman this PR fix #4986 (like #5090) but is a better solution, because it check equality based on the related association. 

Compared to this one, #5090 can result in false equality. If `sameModel` is `true` and `item.as` is defined but not `scopeInclude.as` include will be considered as equal even if `item.as` does not refer to the default association `as`. In this case includes are **not** equal, and both of them must be pushed in `options.include`.

@janmeier could you review?

lib/model.js
```diff
@@ -2619,15 +2623,18 @@ Model.$injectScope = function (scope, options) {
...
+      if (scopeInclude.all || !options.include.some(function matchesModelAndAlias(item) {
+        var isSameModel = item.model && item.model.name === scopeInclude.model.name;
+        if (!isSameModel || !item.as) return isSameModel;
 
-        if (sameModel && item.as) {
+        if (scopeInclude.as) {
           return item.as === scopeInclude.as;
+        } else {
+          var association = scopeInclude.association || self.getAssociation(scopeInclude.model, scopeInclude.as);
+          return association ? item.as === association.as : false;
```
Verdier: @janmeier the new logic is here, every thing else is to have self available.

lib/model.js
```diff
@@ -236,23 +236,21 @@ function conformOptions(options, self) {
...
+    options.include = options.include.map(function(include) {
+      return conformInclude(include, self);
+    });
 
-    return include;
-  });
+    if (!options.include.length) {
+      delete options.include;
+    }
+  }
```
mickhansen: Did anything actually change here other than changing it away from short circuiting?

mickhansen:
Preferred having it as a non-prototype method but @janmeier will have to review and device.

lib/model.js
Verdier: Mmmm, yes... I think it's really more readable like this. But we could revert...

Verdier:
Ok, waiting for @janmeier then

lib/model.js
mickhansen: I just prefer to keep style changes out of PR's that fix/change functionality as much as possible, easier to focus on the matter at hand.

Verdier:
Reverted @mickhansen

Verdier:
The only job of `$injectScope` is to inject `this.$scope` into `options.include`. It's probably better that the method was on the prototype.

mickhansen:
`$injectScope`'s job was to inject an arbitrary scope into an arbitrary options set was it not?

Verdier:
Yep, but the used scope is alway set as `this.$scope` and, at the end, only `this.$scope` is injected. That should remain like this forever, otherwise it's probably an error. Setting `$injectScope` as instance method without `scope` parameter ensure that the function is not misused.

mickhansen:
It's true that the current use of the method is like that. However i'm sure that the intention was for it work for arbitrary scopes and options. But @janmeier will have to say since he wrote it.

Verdier:
I suspect that just was for testing purpose, but waiting for @janmeier.

janmeier:
I guess it was for testing purposes - I agree with @Verdier that if you ever try to inject other than `this.$scope` its probably an error

janmeier:
Just needs a couple of changelog entries, then this should be good to go

Verdier:
@mickhansen could you add the changelog entries?

mickhansen:
@Verdier You might aswell since this needs a squash aswell, just add a notice about fixing the issue at hand here and reference the PR/issues

Verdier:
No problem, I will do that tomorrow. I was thinking that's you who keep the
change log in sync.

You want me to squash the test commits?

Verdier:
Squashed and changelog added @mickhansen.

mickhansen:
Thanks @Verdier

PR #5112, opened 2015-12-29
tsrange data type is added.

ozum:
tsrange is range of timestamp without time zone for PostgreSQL. This makes range types complete as of PostgreSQL 9.5.


mickhansen:
Thanks :)

hduchesn:
`Sequelize.RANGE(Sequelize.DATENOTZ)`  failed because `Sequelize.DATENOTZ` is not defined
and `Sequelize.RANGE("datenotz")` failed because key doesn't exist for the line : 
` this._subtype = options.subtype.key;`...

How to use the datenotz?

PR #5113, opened 2015-12-29
Add support for scope in has-one association

Verdier:
This PR fix two issues:
- the association's `scope` was not injected in `where` clause when eager loaded. This is **extremely critical** because any *raw* that match without the `where` restriction is eager loaded, that could result in sending private data to the client (that's what append to us...),
- the association's `scope` was not used is the creator helper method (ie. `createInstance`)

### DIFF
--- a/lib/associations/has-one.js
+++ b/lib/associations/has-one.js
@@ -135,6 +135,10 @@ HasOne.prototype.injectGetter = function(instancePrototype) {
     var where = {};
     where[association.foreignKey] = this.get(association.sourceIdentifier);
 
+    if (association.scope) {
+      _.assign(where, association.scope);
+    }
+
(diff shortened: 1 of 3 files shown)


mickhansen:
hasOne never had scope support, soooo ;p
The two bugs you mention are actually just the fact that the implementation does not exist.

Verdier:
Haha @mickhansen that's fun... the implementation is here. Could you merge?

mickhansen:
Needs code and test for associationGetter aswell.

Verdier:
Ok I'll do that tomorrow

mickhansen:
Great, will merge then :)

Verdier:
@mickhansen done. Ready to merge.

PR #5115, opened 2015-12-30
Fix(#1150) : Test to verify findAll / find properly clone options

### LINKED ISSUE #1150: Argument passed as conditions to finder being changed
Here is a little sample:
```
var Sequelize = require( 'sequelize' );
var sequelize = new Sequelize( 'dev', 'root', '###', {
	host: '8.8.8.8',
	define: {
		timestamps: false,
		freezeTableName: true,
	    underscored: true
	}
});
var User =  sequelize.define(
        'phpfox_user',
        {
            user_id: {
                type: Sequelize.INTEGER( 10 ).UNSIGNED,
                primaryKey: true,
                autoIncrement: true
            },
            user_name: Sequelize.STRING,
        }
    );
var Tournament = sequelize.define(
        'phpfox_tournaments',
        {
            tournament_id: {
                type: Sequelize.INTEGER( 10 ).UNSIGNED,
                primaryKey: true,
                autoIncrement: true
            },
            title: Sequelize.STRING,
        }
    );
Tournament.hasMany(
	User,
	{
		as: 'AdminTournaments', /** @see https://github.com/sequelize/sequelize/issues/1131 */
		foreignKey: 'tournament_id',
		joinTableName: 'phpfox_tournaments_admin'
	}
);
User.hasMany(
	Tournament,
	{
		as: 'AdminTournaments',
		foreignKey: 'user_id',
		joinTableName: 'phpfox_tournaments_admin'
	}
);
User.hasMany( Tournament, { as: 'OwnTournaments', foreignKey: 'user_id' } );
Tournament.belongsTo( User, { as: 'Owner' });


	User.find( 13051 ).success( function( u ){
		var conditions = { where: { title: { like: '%2013%' } } };
		u.getAdminTournaments( conditions ).success( function( adminTournaments ){
			u.getOwnTournaments( conditions ).success( function( ownTournaments ){
				console.log( adminTournaments.concat( ownTournaments ) );
			});
		});
	})
```
This sample does not work because `conditions` variable has been changed on first finder launch.

This workaround works:
```
	User.find( 13051 ).success( function( u ){
		var conditions1 = { where: { title: { like: '%2013%' } } };
		var _ = require( 'underscore' ),
			conditions2 = _.clone( conditions );
		u.getAdminTournaments( conditions1 ).success( function( adminTournaments ){
			u.getOwnTournaments( conditions2 ).success( function( ownTournaments ){
				console.log( adminTournaments.concat( ownTournaments ) );
			});
		});
	})
```

To my mind, if application can leave arguments not changed – it should do it by default. There are some cases when arguments need to be changed – but in my opinion – this is not that case. Anyway even if you think this behaviour of your app is right – it should be documented at least.

BTW:
http://sequelizejs.com/docs/latest/associations
this example `project.getTasks({ where: 'id > 10' })` will not work. This will: `project.getTasks({ where: ['id > 10'] })`

sushantdhiman:
Fixes #1150

Its a really old bug that has already been fixed at some point. I am adding tests to verify that `options` are properly cloned.

### DIFF
--- a/test/unit/model/findall.test.js
+++ b/test/unit/model/findall.test.js
@@ -6,7 +6,8 @@ var chai = require('chai')
   , Support = require(__dirname + '/../support')
   , current = Support.sequelize
   , sinon = require('sinon')
-  , DataTypes = require(__dirname + '/../../../lib/data-types');
+  , DataTypes = require(__dirname + '/../../../lib/data-types')
+  , _ = require('lodash');
 
 describe(Support.getTestDialectTeaser('Model'), function() {
   describe('method findAll', function () {
@@ -87,6 +88,20 @@ describe(Support.getTestDialectTeaser('Model'), function() {
           ]);
         });
       });
+
+      it('properly clones options values', function() {
+        var options = {
+          attributes: {
+            exclude: ['name'],
+            include: ['name']
+          }
+        };
+        var optionsClones = _.clone(options);
+        return Model.findAll(options).bind(this).then(function () {
+          expect(options).to.deep.equal(optionsClones);
+        });
+      });
+
     });
   });
 });
--- a/test/unit/model/findall.test.js
+++ b/test/unit/model/findall.test.js
@@ -96,7 +96,7 @@ describe(Support.getTestDialectTeaser('Model'), function() {
             include: ['name']
           }
         };
-        var optionsClones = _.clone(options);
+        var optionsClones = _.cloneDeep(options);
         return Model.findAll(options).bind(this).then(function () {
           expect(options).to.deep.equal(optionsClones);
         });
--- a/test/unit/model/findone.test.js
+++ b/test/unit/model/findone.test.js
@@ -7,7 +7,8 @@ var chai = require('chai')
   , current = Support.sequelize
   , sinon = require('sinon')
   , DataTypes = require(__dirname + '/../../../lib/data-types')
-  , Promise = require('bluebird');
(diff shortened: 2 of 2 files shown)


test/unit/model/findall.test.js
```diff
@@ -87,6 +88,20 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
       });
+
+      it('properly clones options values', function() {
+        var options = {
+          attributes: {
+            exclude: ['name'],
+            include: ['name']
+          }
+        };
+        var optionsClones = _.clone(options);
```
Verdier: `_.cloneDeep` should be better?

test/unit/model/findone.test.js
```diff
@@ -67,5 +68,16 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         expect(this.stub.getCall(0).args[0]).to.be.an('object').to.have.property('limit');
       });
     });
+
+    it('properly clones options values', function() {
+      var options = { where: { id: { $gt: 42 }}}
+        , optionsClones = _.clone(options)
```
Verdier: same here?

test/unit/model/findall.test.js
mickhansen: Should make the test a little more robust yes.

sushantdhiman:
Added `_.deepClone` @mickhansen and thanks @Verdier 

Let me know if there are another improvements.

PR #5119, opened 2015-12-30
Show that sequelize supports geometry on documentation

oss92:
Geometry was introduced but is not on the documentation http://docs.sequelizejs.com/en/latest/api/datatypes


mickhansen:
@janmeier This is fine right? Or are the docs in the dialect specific implementations?

janmeier:
This is fine, we don't look for documentation tags in the dialect specific implementations

Issue #3759, opened 2015-05-20
How to manually set updatedAt?


stephenjwatkins:
I'm working on some tests and I need to manually set the `updatedAt` attribute. However, it always gets updated to the current date. For example, the following code does not override the `updatedAt` attribute, as I expected it to:

    instance.updateAttributes({
        updatedAt: moment.utc().subtract(31, 'minutes').toDate()
    }, {
        silent: true
    });

Am I missing something related to the proper use of `silent`?

janmeier:
Yeah, this usecase is not really supported I guess - save will always set updatedAt to the current timestamp, unless you set silent to true, in which case it will ignore it completely.


I guess it actually makes to only set updatedAt if changed returns false

mickhansen:
@stephenjwatkins as a work around you could use sinon clock for testing.

stephenjwatkins:
@mickhansen I appreciate the suggestion. I will indeed use sinon clock for testing to work around this.

ssteffl:
i wish there was a "force" flag for .save() on an instance, that just set the updatedAt flag even if nothing was changed. Im trying to use updatedAt field as a timestamp of the last time a record was known to be valid, since i'm syncing stuff in my db with a 3rd party's db. That way I can purge items from my db that are stale. 

i dont want to create a new column for this if i could just get the same functionality out of the updatedAt field

janmeier:
@ssteffl you can do `instance.changed('updatedAt', true)`

ssteffl:
ok cool. that seems to work. im just doing:

    existingRecord.changed('updatedAt', true);
    return existingRecord.update(data, {transaction: txn});

and it seems to be setting the updatedAt correctly. Thanks!

nathanmarks:
@mickhansen Any idea if it's possible to do this without using raw sql? (ie using the sequelize instance API).

My use case is that I am creating entities in an offline application (uuid primary key) and setting updatedAt in that application. The application syncs with an online database via a node API using sequelize. The "updated" `updatedAt` after syncing with the server is causing some edge case issues.

Accepting PRs for this?

nathanmarks:
Additionally, if this is something that you would be open to having implemented, do you have any opinions on what the user should pass in terms of options to the update/save/etc functions? I haven't contributed before here and not sure what the most idiomatic option is.

mickhansen:
[...]

PR #5133, opened 2016-01-03
Failing test assertion

nathanmarks:
When trying to set `updatedAt`, the instance returned from save via the promise fulfillment does not properly reflect the saved data.

Run test using:

``` bash
make test-integration-sqlite GREP="does not update timestamps when passing"
```

Refer to https://github.com/sequelize/sequelize/issues/3759#issuecomment-165292388

> `Instance.save()` trips up here but returns the instance with the set attribute: https://github.com/sequelize/sequelize/blob/master/lib/instance.js#L637

In the test, you get a successful promise fulfillment from save containing an attribute that was not committed to the database. The reason I have the 2nd assertion after re-querying the database is because that one will actually pass (if the test was able to get to it, it fails the first assertion).

### DIFF
--- a/test/integration/instance.test.js
+++ b/test/integration/instance.test.js
@@ -1071,15 +1071,25 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
     });
 
     it('does not update timestamps when passing silent=true', function() {
+      var self = this;
+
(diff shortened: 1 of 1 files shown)


janmeier:

PR #5149, opened 2016-01-06
Fixed various docs issues

### LINKED ISSUE #1220: Sequelize documentation is incorrect when expressing how you should set charset for MySQL
My system needs to be able to store 4-byte unicode characters in a MySQL database. 

Here is a sample of my database setup:

```
CREATE DATABASE `my_db` CHARACTER SET utf8 COLLATE utf8_general_ci;

CREATE TABLE tweet (
       id INT NOT NULL AUTO_INCREMENT,
       text TEXT CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL,
       PRIMARY KEY (  id ));
```

Here is the test that I was attempting to run:
```
// The utf8mb4 character doesn't want to appear in GitHub so it's located here: http://www.iemoji.com/view/emoji/26/smileys/face-with-tears-of-joy
var tweet = db.Tweet.build({
    text: 'Testing 4-byte character: [[4-BYTE CHARACTER]] #myhashtag',
});

      tweet.save().success(function(results) {
          expect(results).toBeDefined();
      }).error(function(err) {
          expect(err).toBeUndefined();
          console.log(err);
      });
```

The sequelize documentation (http://sequelizejs.com/docs/latest/usage) states that you should be able to set charset like so:

```
define: {
    underscored: false
    freezeTableName: false,
    syncOnAssociation: true,
    charset: 'utf8',
    collate: 'utf8_general_ci',
    classMethods: {method1: function() {}},
    instanceMethods: {method2: function() {}},
    timestamps: true
  },
```

So, if I set the charset to utf8mb4 and collation to utf8mb4_general_ci, I should be able to run the test successfully. 

However, using the logic above, when I run the test I receive the following:

```
 {[Error: ER_TRUNCATED_WRONG_VALUE_FOR_FIELD: Incorrect string value: '\xF0\x9F\x98\x82 #...' for column 'text' at row 1]
  code: 'ER_TRUNCATED_WRONG_VALUE_FOR_FIELD',
  index: 0,
  sql: 'INSERT INTO `tweet` (`text`) VALUES (\'Testing 4-byte character: [[4 BYTE CHARACTER]] #myhashtag\',);' }
```

I have had to set my settings to the following:

```
{
    pool: { maxConnections: 5, maxIdleTime: 30 },
    dialect: sequelizeConfig.dialect,
    host: sequelizeConfig.host,
    dialectOptions: {
      charset: 'utf8mb4'
    },
    define: {
      underscored: true,
      freezeTableName: true,
      syncOnAssociation: true
  }
```

When I set the value in dialectOptions, the test passes correctly.

### LINKED ISSUE #2912: [docs] update() documentation should reflect that it only saves values passed to it
update() has different semantics than set() + save() since update() will only save exactly the values passed to it. Hence it's a more atomic and safe operation in some cases (since you know exactly what it will save will not be dependent on earlier set operations).

#2910

### LINKED ISSUE #4648: [doc] the migration skeleton doc seems out of date
That sections reads:

```
..... The third parameter is a callback function which needs to be called once everything was executed. The first parameter of the callback function can be used to pass a possible error. In that case, the migration will be marked as failed. Here is some code:

module.exports = {
  up: function(queryInterface, Sequelize) {
    return queryInterface.dropAllTables();
  }
}
```

Firstly, the example is not relevant to `the third done() callback param`
Secondly, do we still support this callback way anymore?

### LINKED ISSUE #4674: Broken link to Utils API documentation
In the [API documentation](http://docs.sequelizejs.com/en/latest/api/sequelize/#utils), the link to `http://sequelize.readthedocs.org/en/latest/api/utils` is broken.

### LINKED ISSUE #5139: Count documentation isn't very clear
According to the documentation to use `count` in order to look for distinct entries we need to add `distinct: true`. Unfortunately it is not very clear which column it refers to. How to define the column to consider for the distinction?

sushantdhiman:
Closes #2912
Closes #1220
Closes #4648
Closes #4674
Closes #5139

### DIFF
--- a/docs/docs/migrations.md
+++ b/docs/docs/migrations.md
@@ -65,7 +65,7 @@ module.exports = {
 }
 ```
 
-The passed `queryInterface` object can be used to modify the database. The `Sequelize` object stores the available data types such as `STRING` or `INTEGER`. The third parameter is a callback function which needs to be called once everything was executed. The first parameter of the callback function can be used to pass a possible error. In that case, the migration will be marked as failed. Here is some code:
+The passed `queryInterface` object can be used to modify the database. The `Sequelize` object stores the available data types such as `STRING` or `INTEGER`. Function `up` or `down` should return a `Promise`. Here is some code:
 
 ```js
 module.exports = {
--- a/lib/model.js
+++ b/lib/model.js
@@ -1552,12 +1552,12 @@ Model.prototype.aggregate = function(attribute, aggregateFunction, options) {
  * @param {Object}        [options]
  * @param {Object}        [options.where] A hash of search attributes.
  * @param {Object}        [options.include] Include options. See `find` for details
- * @param {boolean}       [options.distinct] Apply COUNT(DISTINCT(col))
+ * @param {boolean}       [options.distinct] Apply COUNT(DISTINCT(col)) on primary key, `Model.aggregate` should be used for other columns
  * @param {Object}        [options.attributes] Used in conjunction with `group`
  * @param {Object}        [options.group] For creating complex counts. Will return multiple rows as needed.
- * @param  {Transaction}  [options.transaction] Transaction to run query under
+ * @param {Transaction}   [options.transaction] Transaction to run query under
  * @param {Function}      [options.logging=false] A function that gets executed while running the query to log the sql.
- * @param  {String}       [options.searchPath=DEFAULT] An optional parameter to specify the schema search_path (Postgres only)
+ * @param {String}        [options.searchPath=DEFAULT] An optional parameter to specify the schema search_path (Postgres only)
  *
  * @return {Promise<Integer>}
  */
--- a/docs/docs/usage.md
+++ b/docs/docs/usage.md
@@ -90,7 +90,9 @@ var sequelize = new Sequelize('database', 'username', 'password', {
     freezeTableName: false,
     syncOnAssociation: true,
     charset: 'utf8',
-    collate: 'utf8_general_ci',
+    dialectOptions: {
+      collate: 'utf8_general_ci'
+    },
     classMethods: {method1: function() {}},
     instanceMethods: {method2: function() {}},
     timestamps: true
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -770,7 +770,8 @@ Instance.prototype.hookValidate = function(options) {
 };
 
 /**
- * This is the same as calling `set` and then calling `save`.
+ * This is the same as calling `set` and then calling `save` but it only saves the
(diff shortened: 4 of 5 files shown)


lib/model.js
```diff
@@ -1552,7 +1552,7 @@ Model.prototype.aggregate = function(attribute, aggregateFunction, options) {
  * @param {Object}        [options]
  * @param {Object}        [options.where] A hash of search attributes.
  * @param {Object}        [options.include] Include options. See `find` for details
- * @param {boolean}       [options.distinct] Apply COUNT(DISTINCT(col))
+ * @param {boolean}       [options.distinct] Apply COUNT(DISTINCT(col)) on primary key, `Model.aggregate` should be used for other coloumns
```
mickhansen: columns*

changelog.md
```diff
@@ -1,4 +1,9 @@
 # Future
+- [Docs] Indicate Instance.update will update exact values passed to it [#2912](https://github.com/sequelize/sequelize/issues/2912)
+- [Docs] Collation documentation is incorrect [#1220](https://github.com/sequelize/sequelize/issues/1220)
+- [Docs] Migration docs still prefer `done` callback [#4648](https://github.com/sequelize/sequelize/issues/4648)
+- [Docs] Utils section doesn't exists [#4674](https://github.com/sequelize/sequelize/issues/4674)
+- [Docs] Unclear distinct option in Model.count() [#5139](https://github.com/sequelize/sequelize/issues/5139)
```
mickhansen: No need to apply changelog for docs :)

docs/docs/migrations.md
```diff
@@ -65,7 +65,7 @@ module.exports = {
 }
 ```
 
-The passed `queryInterface` object can be used to modify the database. The `Sequelize` object stores the available data types such as `STRING` or `INTEGER`. The third parameter is a callback function which needs to be called once everything was executed. The first parameter of the callback function can be used to pass a possible error. In that case, the migration will be marked as failed. Here is some code:
+The passed `queryInterface` object can be used to modify the database. The `Sequelize` object stores the available data types such as `STRING` or `INTEGER`. Function `up` or `down` should return a `Promise`. Third parameter is a callback function available for backwards compatibility. It needs to be called once everything has been executed. The first parameter of the callback function can be used to pass a possible error. In that case, the migration will be marked as failed. Here is some code:
```
mickhansen: Remove mention of callbacks entirely.

sushantdhiman:
Done

mickhansen:
Thanks a ton for all your work @sushantdhiman

sushantdhiman:
Thanks

Issue #488, opened 2013-03-14
Hi， is it possible to log out how long a individual sql has taken?


cmingxu:
Hi， is it possible to log out how long a individual sql has taken?

Issue #2494, opened 2014-10-28
Profiling and benchmarking


vodolaz095:
Is there any way to profile this module?
I have learned that there is a way to output the queries send to database by using
`options.logging = function(query){ ....}`
http://sequelizejs.com/docs/1.7.8/usage#block-4-line-14

Is there any way to add timestamps to this function, so we know when query was send, when it was processed by database, does it have errors or valid response and so on?

sushantdhiman:
@vodolaz095  #5116 provide an option to benchmark the queries via logs. May be it can help

sushantdhiman:
Should we close it as #5116 is now merged and there is a way to benchmark. @janmeier @mickhansen

janmeier:
@sunshinewyin At least were mostly there :)

I think for your PR to be useful for profiling, it would be nice if the logging function was passed the sql and the timing as two separate arguments, instead of having to extract the running time from a string.

In that case the `sequelize.log` function should create the `Executing...` string from its arguments

sushantdhiman:
Yes thats a good idea. I will work on it if I get some time.

janmeier:
Fixed in https://github.com/sequelize/sequelize/pull/5282

PR #5154, opened 2016-01-06
Feat #488 : Benchmarks based on the pure JS Date.now()

sushantdhiman:
Implements the `benchmark` feature using the `getTime` and return the query execution time in `milliseconds`. It doesn't depend upon the `node-microtime`

### DIFF
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -3,8 +3,7 @@
 var Utils = require('../../utils')
   , AbstractQuery = require('../abstract/query')
   , sequelizeErrors = require('../../errors.js')
-  , parserStore = require('../parserStore')('mssql')
-  , microtime = require('microtime');
+  , parserStore = require('../parserStore')('mssql');
 
 var Query = function(connection, sequelize, options) {
   this.connection = connection;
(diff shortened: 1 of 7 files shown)


lib/dialects/mssql/query.js
```diff
@@ -30,7 +30,14 @@ Query.prototype.run = function(sql, parameters) {
...
   this.sql = sql;
 
-  this.sequelize.log('Executing (' + (this.connection.uuid || 'default') + '): ' + this.sql, this.options);
+  //do we need benchmark for this query execution
+  var benchmark = this.sequelize.options.benchmark || this.options.benchmark;
+
+  if (!benchmark) {
+    this.sequelize.log('Executing (' + (this.connection.uuid || 'default') + '): ' + this.sql, this.options);
+  } else {
+    var queryBegin = (new Date()).getTime();
```
mickhansen: `Date.now()` does the same and is a bit shorter.

sushantdhiman: Ok

sushantdhiman:
Changed to `Date.now()`

sushantdhiman:
Thanks :)

sushantdhiman:
@mickhansen Can you please add this feature into changelog when releasing next minor / patch version.

tornillo:
Hi, @sushantdhiman!

I propose to pass the bench-time as the second argument to the logger function.
This will allow to do something like this:
```js
function logger (sql, time) {
	if (time => 100/*ms*/) {
		// log to file or do any other action
	}
}
```
This will be especially useful for production environment.
What do you think about this?

sushantdhiman:
Hi @tornillo 

I think thats a good idea. There is already an issue( #2494) related to feature somewhat like this. Current feature has already been merged so I cant change it. But you can send new PR targeting #2494 :)

tornillo:
@sushantdhiman I had not noticed this issue:)
Ok, I have created a new PR https://github.com/sequelize/sequelize/pull/5282 with the changes discussed.

PR #5157, opened 2016-01-07
Added retry on BUSY state for Sqlite.

EToreo:
I'm trying to clean up state from another pull request (https://github.com/sequelize/sequelize/pull/4969) that is failing on build.  Please ignore this pull request, as it is a test to see if the other branch I am using for that pull request is messed up.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# 3.17.1
+- [FIXED] Reverted benchmarking feature since it does not compile on Node v4.0
+
 # 3.17.0
 - [CRITICAL] Fixed injection vulnerability for order/limit
 - [FIXED] MySQL throws error when null GEOMETRY data results in empty buffer [#4953](https://github.com/sequelize/sequelize/issues/4953)
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,6 @@
 # 3.17.0
 - [CRITICAL] Fixed injection vulnerability for order/limit
 - [FIXED] MySQL throws error when null GEOMETRY data results in empty buffer [#4953](https://github.com/sequelize/sequelize/issues/4953)
-- [ADDED] Support for benchmarking the execution time for SQL queries [#488](https://github.com/sequelize/sequelize/issues/488)
 
 # 3.16.0
 - [ADDED] PostgreSQL tsrange (Range of timestamp without time zone) data type support.
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,4 @@
-# Future
+# 3.17.0
 - [CRITICAL] Fixed injection vulnerability for order/limit
 - [FIXED] MySQL throws error when null GEOMETRY data results in empty buffer [#4953](https://github.com/sequelize/sequelize/issues/4953)
 - [ADDED] Support for benchmarking the execution time for SQL queries [#488](https://github.com/sequelize/sequelize/issues/488)
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# 3.17.3
+- [FIXED] Regression with array values from security fix in 3.17.2
+
 # 3.17.2
 - [SECURITY] Force non-buffer blob values to string, https://github.com/nodejs/node/issues/4660
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,8 +1,11 @@
+# 3.17.2
+- [SECURITY] Force non-buffer blob values to string, https://github.com/nodejs/node/issues/4660
+
 # 3.17.1
 - [FIXED] Reverted benchmarking feature since it does not compile on Node v4.0
 
 # 3.17.0
-- [CRITICAL] Fixed injection vulnerability for order/limit
+- [SECURITY] Fixed injection vulnerability for order/limit
 - [FIXED] MySQL throws error when null GEOMETRY data results in empty buffer [#4953](https://github.com/sequelize/sequelize/issues/4953)
 
 # 3.16.0
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# [FUTURE]
+- [ADDED] Sqlite now retrys database queries that return SQL_BUSY as the status.  The default is 5 retries with a backoff function.  `retry` object can be passed to options with max: 0 to turn off this behavior.
+
 # 3.17.3
 - [FIXED] Regression with array values from security fix in 3.17.2
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
-# [FUTURE]
-- [ADDED] Sqlite now retrys database queries that return SQL_BUSY as the status.  The default is 5 retries with a backoff function.  `retry` object can be passed to options with max: 0 to turn off this behavior.
+# Future
+- [ADDED] `retry` object now part of global settings and can be overridden per call.  The default is 5 retries with a backoff function.  `retry` object can be passed to options with max: 0 to turn off this behavior.
+- [ADDED] Sqlite now retries database queries that return SQL_BUSY as the status.
 
 # 3.17.3
 - [FIXED] Regression with array values from security fix in 3.17.2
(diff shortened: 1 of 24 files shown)


EToreo:
@mickhansen, @janmeier I am really struggling here. I have tried this in every environment I can (Linux, mac, different node versions, etc.) and it never fails.  The moment Travis-CI gets to it, it fails.  Am I missing something about the test environment?  I would really like to get this wrapped up.  Any thoughts you have would be greatly appreciated.

mickhansen:
Differences in the sqlite library?
I'll see if i can find time to test it locally.

mickhansen:
@EToreo Could you link the spot in the SQLITE docs that meantions SQLITE_BUSY at the procedure?

EToreo:
@mickhansen Is this the doc you are talking about? https://www.sqlite.org/lang_transaction.html

mickhansen:
Well looking for the line of text that says SQLITE_BUSY should ALWAYS be retried.
It is, but sqlite3 builds native code, so that might differ a depending on the node.js version?

EToreo:
> An atempt to execute COMMIT might also result in an SQLITE_BUSY return code if an another thread or process has a shared lock on the database that prevented the database from being updated. When COMMIT fails in this way, the transaction remains active and the COMMIT can be retried later after the reader has had a chance to clear.

mickhansen:
Sort of, it doesn't explicitely say that is the best course of action.
I'm just wondering if this might cause weird breakage for users.

EToreo:
I'm having a hard time imagining a case where an application wouldn't want to retry at least a few times before giving up on a locked database/table/row.  That being said, the application can always pass `{retry: 0}`.

EToreo:
@mickhansen The travis-ci build didn't seem to pick up my last push.  Do you have some sort of administrative "clear cache" option and ability to force a re-build to pick up https://github.com/EToreo/sequelize/commit/9361889acb18e30fa1dab257d0f374ee6d63159d?  I have a good feeling that it fixes what I suspect is a race condition.

mickhansen:
It's blocking on a lint error atm: `test/integration/transaction.test.js: line 296, col 64, Mixed double and single quotes.`

EToreo:
@maennchen OK, thank you for pointing that out.  I think I was flustered and saw the message about the cache being used and not updated.  Anyway, the test is now passing after updating the test to force a locking collision.

The only tests failing now are failing on the project's master branch as well.  (Looks like some sort of timing issue on mysql?) Hopefully this pull request can now by accepted.  I plan to move onto function callback type transactions next.

mickhansen:
Re-running those gd intermittent errors.
Can you add a changelog and rebase.
We likely also need to mention this SQLITE default somewhere and show how to disable it.

EToreo:
@mickhansen OK, I added the changelog (I didn't know what version to call it, so I just guessed at 3.17.2).  I also updated the documentation comments to more clearly explain the retry option and how to disable it.  You might need to move along the intermittent errors again.

mickhansen:
Just put "Future", since we might get more changes before cutting a release :)

changelog.md
```diff
@@ -1,3 +1,6 @@
+# 3.17.2
+- [ADDED] Sqlite now retrys database queries that return SQL_BUSY as the status.  The default is 5 retries with a backoff function.  `retry` object can be passed to options with max: 0 to turn off this behavior.
+
```
mickhansen: Passed to options per query or globally to the sequelize constructor

mickhansen: Maybe also add a changelog about supporting `retry` in general :)

lib/sequelize.js
```diff
@@ -145,6 +149,7 @@ var Sequelize = function(database, username, password, options) {
     pool: {},
     quoteIdentifiers: true,
     hooks: {},
+    retry: {max: 5, match: ['SQLITE_BUSY: database is locked']},
```
mickhansen: So these are global settings, would they automatically also apply to transactions in the future then?

EToreo: They are global and per-call settings.  The global ones apply on every query unless overridden in the call.

mickhansen: Yes i'm aware, but in the future would we apply the global defaults to both managed transactions and queries?

EToreo: Ah, yes, that was my plan.

mickhansen: I'm just tryng to consider if there's any possible issues with that, but nothing springs to mind.

mickhansen: Well actually, if you add a global retry on stuff like `unable to serialize` then that will trigger for the COMMIT query and retry the commit where it should actually be retrying the entire transaction instead.

This could obviously be fixed by just setting this as default in `sequelize.transaction`, but how do we let users implement it if need be then?

EToreo: I don't think you can get busy or serialization errors inside a transaction, only once you try to commit.  So, it would re-try the entire transaction in that case.  Or did I misunderstand the example?

mickhansen: For postgres the serialization error with be returned when you attempt to commit. But retrying a commit will do nothing in that case, you need to rerun the sequence in a new transaction.

So if Sequelize would catch the serialization error on the COMMIT query handler and keep retrying, that would be wrong :)

EToreo: Interesting, I'm not sure in that case. Seems like the sequence will need a different retry setting.  Although, I'm not familiar with sequences. Is that the callback type of transaction in sequelize?

In any case, that seems like a job for the next PR.

mickhansen: I simply meant a sequence of queries inside a transaction :)
But yeah we'll likely need `retry` to only apply to queries and then possible have a global argument like `retryTransaction`

EToreo:
@mickhansen  Looks like everything is passing and I added those notes you wanted.  Can we get this pulled?

mickhansen:
Definitely @EToreo, we just need a squash since there appears to be duplicate commits + a merge commit (we really don't want merge commits) :)

EToreo:
Please forgive my naivety, I'm very new to git. What is a squash? Anything I can do to help sort it out, or better yet, advice on how to not cause it in the future?

mickhansen:
In the future just make sure you never merge anything, if you need to pull in changed you sync your forked master and then rebase your branch off your master.

http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

EToreo:
I did my best to rebase this after the master project moved ahead.  It's ready to be pulled again.

mickhansen:
Ah i'm sorry but now it's even worse: https://github.com/sequelize/sequelize/pull/5157/commits
How did you rebase exactly? usually you should be able to replay your commits and then possible squash them.

EToreo:
Hmmm.  One of these days git will click for me and I will understand how to make heads or tails of all this.

I needed to get the changes from the project onto my master branch.  After many searches came up with only one answer, I used the merge command to get the sequelize project's changes into my master branch.  From there I rebased my branch off my master.  Then I pushed those changes to this PR.

Is there some better way to go about that?

mickhansen:
Ah, you should have rebased your own master off the sequelize master, then rebased the branch.
It's somehwat complicated stuff at times :)

I'm not sure i have a good idea, @janmeier?

EToreo:
I tried that.  For some reason I can't remember now it did not want to rebase off of sequalize.  At any rate, advice is welcome.

mickhansen:
The diff is royally messed up now atleast, i'm not sure how to fix it, maybe @janmeier knows but you might have to resort to picking up your changes and applying them to a new branch.

janmeier:
You might be able to cherry-pick your commits onto a fresh branch https://git-scm.com/docs/git-cherry-pick

EToreo:
@mickhansen  Closed in favor of https://github.com/sequelize/sequelize/pull/5263.  (Third time a charm?)

Can we get it pulled quickly so I don't foul it up again?

PR #5162, opened 2016-01-07
Parse dates as UTC instead of local timezone

johanneswuerbach:
Looks like a regression as this wasn't the behaviour in `< 3.14`, I'll look into adding a test when you are :+1:

### DIFF
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -448,7 +448,7 @@ DATE.prototype.validate = function(value) {
 };
 
 DATE.prototype.$applyTimezone = function (date, options) {
-  date = moment(date);
+  date = moment.utc(date);
 
   if (options.timezone) {
     if (moment.tz.zone(options.timezone)) {
--- a/test/integration/model/findAll.test.js
+++ b/test/integration/model/findAll.test.js
@@ -380,6 +380,18 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         });
       });
 
+      it('should be able to find a row using equal with dates', function() {
+        return this.User.findAll({
+          where: {
+            theDate: '2013-01-10 12:00'
+          }
+        }).then(function(users) {
+          expect(users.length).to.equal(1);
+          expect(users[0].username).to.equal('boo2');
+          expect(users[0].intVal).to.equal(10);
+        });
+      });
+
       it('should be able to find a row using greater than or equal to', function() {
         return this.User.find({
           where: {


lib/data-types.js
```diff
@@ -448,7 +448,7 @@ DATE.prototype.validate = function(value) {
 };
 
 DATE.prototype.$applyTimezone = function (date, options) {
-  date = moment(date);
+  date = moment.utc(date);
```
janmeier: As I remember, the date parameter here is a JS date object, so moment should be able to parse it correctly.

Can you show a test case?

johanneswuerbach:
Sure, the issue applies when using strings instead of date objects.

Also when running the test locally I receive:
```
Executing (default): INSERT INTO "Users" ("id","username","intVal","theDate","createdAt","updatedAt","binary") VALUES (DEFAULT,'boo',5,'2013-01-01 11:00:00.000 +00:00','2016-01-07 14:56:41.369 +00:00','2016-01-07 14:56:41.369 +00:00',NULL),(DEFAULT,'boo2',10,'2013-01-10 11:00:00.000 +00:00','2016-01-07 14:56:41.369 +00:00','2016-01-07 14:56:41.369 +00:00',E'\\x01010101010101010101010101010101');
Executing (default): SELECT "id", "username", "secretValue", "data", "intVal", "theDate", "aBool", "binary", "createdAt", "updatedAt" FROM "Users" AS "User" WHERE "User"."theDate" = '2013-01-10 11:00:00.000 +00:00';
```

So it seems to be using my local timezone instead of UTC.

janmeier:
Alrighty - I would think set actually converted the string to a date somewhere..

In any case it should be used options.timezone, since users might have set something other than UTC

johanneswuerbach:
Wouldn't this be a breaking change and considered carefully? Right now all times are moved to `options.timezone` and after that everything would be parsed as `options.timezone`.

janmeier:
Not sure - we don't really have a defined behaviour for string dates - But the general rule is that everything respects options.timezone.

I see your point, just not sure which (if any) assumptions we should make about strings. I guess UTC might actually be the most predictable ... @mickhansen ?

mickhansen:
Why is the date a string in the first place rather than a date object?

johanneswuerbach:
I'm using strings as our user input are strings and it worked fine in sequelize `< 3.14` (before the DataType refactoring). If you shouldn't do that, I guess it should be added to the docs and that method should throw instead of silently parsing the string using local time.

mickhansen:
Hmm, `set` should still be converting to a data for you: https://github.com/sequelize/sequelize/blob/master/lib/instance.js#L346

mickhansen:
Hmm, `set` should still be converting to a data for you: https://github.com/sequelize/sequelize/blob/master/lib/instance.js#L346

In any case Sequelize does expect date objects for date fields.

johanneswuerbach:
Sorry, the issue is not happening during set, but during query time. 
E.g. I do:
```
this.User.findAll({
  where: {
    theDate: '2013-01-10 12:00'
  }
}
```

The date comes via an http parameter, so its a string and before I didn't had to convert it.

janmeier:
Well I guess you could argue that in that case there is no need to call stringify if the value is already a string ... This should apply to all types I guess, can't see any immediate problems

mickhansen:
Serves me right for not reading the test case, i was just assuming this had to do with inserts.

janmeier:
@mickhansen Do you agree that we should simply not call stringify if the value is already a string? I think that's better than trying to make a guess about the timezone

mickhansen:
I agree, attempting to stringify a string is probably never a reasonable thing to do.

janmeier:
@johanneswuerbach You up for that - should be somewhere around here that needs a check for isstring https://github.com/sequelize/sequelize/blob/8427412278d4557bbf0dab99d59e577ba37d1cc1/lib/dialects/abstract/query-generator.js#L968

johanneswuerbach:
I moved away from sequelize for query building, so somebody else would need to champion this.

PR #5164, opened 2016-01-07
Correct type in documentation in instance save()

samkelleher:
The type is an array of strings for the property names, indeed as described in the comment. Not an object as typed. The documentation page end up contradicting each other.


janmeier:
Fanx :)

PR #5170, opened 2016-01-08
Few more documents updates

### LINKED ISSUE #2797: [docs] Better distinguishment between hasOne and belongsTo
We need to be absolutely clear, beyond any doubt, where what relation will add a foreign key.

### LINKED ISSUE #2846: Change all examples to use lowercase model names

### LINKED ISSUE #3348: "otherKey" for belongsToMany is undocumented in associations documentation
See: http://docs.sequelizejs.com/en/latest/docs/associations/

And: https://github.com/sequelize/sequelize/blob/master/lib/associations/belongs-to-many.js#L90

Example:

```javascript
Product.belongsToMany(Category, {
  through: ProductCategory,
  foreignKey: 'product_id',
  otherKey: 'category_id'
})
```

sushantdhiman:
Closes #2797
Closes #2846
Closes #3348

### DIFF
--- a/docs/docs/associations.md
+++ b/docs/docs/associations.md
@@ -10,11 +10,13 @@ BelongsTo associations are associations where the foreign key for the one-to-one
 A simple example would be a **Player** being part of a **Team** with the foreign key on the player.
 
 ```js
-var Player = this.sequelize.define('Player', {/* attributes */})
-  , Team  = this.sequelize.define('Team', {/* attributes */});
+var Player = this.sequelize.define('player', {/* attributes */})
+  , Team  = this.sequelize.define('team', {/* attributes */});
 
 Player.belongsTo(Team); // Will add a TeamId attribute to Player to hold the primary key value for Team
 ```
+
+Observe Player is **source model** and Team is **target model**. Thus Player have information about its Team as `TeamId` making it the **source**. This relation will add a `TeamId` attribute to **source model** i.e. Player model.
 
 #### Foreign keys
 
@@ -23,13 +25,13 @@ By default the foreign key for a belongsTo relation will be generated from the t
 The default casing is `camelCase` however if the source model is configured with `underscored: true` the foreignKey will be `snake_case`.
 
 ```js
-var User = this.sequelize.define('User', {/* attributes */})
-  , Company  = this.sequelize.define('Company', {/* attributes */});
+var User = this.sequelize.define('user', {/* attributes */})
+  , Company  = this.sequelize.define('company', {/* attributes */});
 
 User.belongsTo(Company); // Will add CompanyId to user
 
-var User = this.sequelize.define('User', {/* attributes */}, {underscored: true})
-  , Company  = this.sequelize.define('Company', {
+var User = this.sequelize.define('user', {/* attributes */}, {underscored: true})
+  , Company  = this.sequelize.define('company', {
     uuid: {
       type: Sequelize.UUID,
       primaryKey: true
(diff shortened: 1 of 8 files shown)


docs/docs/associations.md
```diff
@@ -10,12 +10,14 @@ BelongsTo associations are associations where the foreign key for the one-to-one
...
-var Player = this.sequelize.define('Player', {/* attributes */})
-  , Team  = this.sequelize.define('Team', {/* attributes */});
+var Player = this.sequelize.define('player', {/* attributes */})
+  , Team  = this.sequelize.define('team', {/* attributes */});
 
 Player.belongsTo(Team); // Will add a TeamId attribute to Player to hold the primary key value for Team
 ```
 
+Observe Player is **source model** and Team is **target model**. Thus Player have information about its Team as `TeamId` making it the **source**. This relation will add a `TeamId` attribute to **source model** i.e. Player model.
+
```
mickhansen: `teamId` when using a lowercase model name actually.

docs/docs/associations.md
```diff
@@ -75,12 +77,14 @@ User.belongsTo(Company, {foreignKey: 'fk_companyname', targetKey: 'name'}); // A
...
-var Project = sequelize.define('Project', {/* ... */})
+var User = sequelize.define('user', {/* ... */})
+var Project = sequelize.define('project', {/* ... */})
  
 // One-way associations
 Project.hasOne(User)
- 
+
+Observe Project is **source model** and User is **target model**. In `hasOne` User have information about its Project as `ProjectId`, information exists on **target model**. This relation will add a `ProjectId` attribute to **target model** i.e. User model.
+
```
mickhansen: Not sure this is the most ideal explanation, we might need a section "Difference between hasOne and belongsTo"

sushantdhiman: Ok I will add extra section

sushantdhiman:
Added new section for difference

sushantdhiman:
@mickhansen Your review on changes please.

docs/docs/associations.md
```diff
@@ -42,30 +42,30 @@ User.belongsTo(Company); // Will add company_uuid to user
...
+var User = this.sequelize.define('user', {/* attributes */})
+  , Company  = this.sequelize.define('company', {/* attributes */});
 
 User.belongsTo(Company, {foreignKey: 'fk_company'}); // Adds fk_company to User
 ```
 
 #### Target keys
 
-By default the target key for a belongsTo relation will be the target primary key. To override this behavior, use the `targetKey` option.
+Its the column that will be referred from **target model**. By default the target key for a belongsTo relation will be the target model's primary key. To override this behavior, use the `targetKey` option.
```
mickhansen: Not sure this explanation makes it more clear, maybe:

```
The target key is the column on the target model that the foreign key column on the source model points to. By default the target key for a belongsTo relation will be the target models primary key. To define a custom column, use the `targetKey` option.
```

`model's` => `models` (otherwise it reads as `model is primary key`)

docs/docs/associations.md
```diff
@@ -126,19 +126,64 @@ Game.belongsTo(Team);
...
+```js
+Player.belongsTo(Team); // This will add `teamId` into Player i.e. source model
+```
+
+If **target** model has the Foreign Key , then use HasOne. For example
+
+**Team** model has the `playerId` key which makes it **target** relation. We should use the HasOne
+
+```js
+Player.hasOne(Team); // This will add `playerId` into Team model
```
mickhansen: Not sure this accurately describes it, perhaps setup a case with Team, Coach and Player where Coach and Player are all types of Person.

docs/docs/associations.md
```diff
@@ -162,6 +207,15 @@ User.belongsToMany(Project, { as: 'Tasks', through: 'worker_tasks', foreignKey:
 Project.belongsToMany(User, { as: 'Workers', through: 'worker_tasks', foreignKey: 'projectId' })
 ```
 
+`foreignKey` will allow you to set **source model** key in the **through** relation.
+`otherKey` will allow you to set **target model** key in the **through** relation.
+
+Here is some code which set both `foreignKey` and `otherKey` for **through** relation.
```
mickhansen: Remove this line, should be clear it's an example from the code block.

docs/docs/associations.md
sushantdhiman: Ok I am changing the line. 

But need clarification on `model's` part. Adding apostrophes will make it clear that target model primary key is used. I think `model's` is not read as `model is`.

docs/docs/associations.md
sushantdhiman: Sure

sushantdhiman:
Please review

PR #5175, opened 2016-01-09
Update belongs-to-many.md

fredericlefeurmou:


docs/api/associations/belongs-to-many.md
```diff
@@ -10,8 +10,8 @@ When the join table has additional attributes, these can be passed in the option
 UserProject = sequelize.define('user_project', {
   role: Sequelize.STRING
 });
-User.belongsToMany(Project, { through: UserProject });
-Project.belongsToMany(User, { through: UserProject });
+User.belongsToMany(Project, { through: 'UserProject' });
+Project.belongsToMany(User, { through: 'UserProject' });
```
mickhansen: The original was correct.

fredericlefeurmou: It doesn't work for me.

Also the `allowNull` option isn't taken into account when associated with a `name` attribute:

```
        Conversation.belongsToMany(models.User, {
          through: 'UserConversations',
          foreignKey: {
            name: 'UserId',
            allowNull: false
          }
        });
```

Without the name attribute, the foreignKey name is set to 'undefined'.

mickhansen: The change you made were incorrect in the scenario since there has been a model defined which is being references.
Please keep issues seperate, open an issue if you've found a bug with `foreignKey` in belongsToMany

PR #5184, opened 2016-01-11
Add geography support - postgis

onzag:
Add the GEOGRAPHY datatype support from postgis based on http://postgis.net/docs/using_postgis_dbmanagement.html#PostGIS_Geography which is indeed very similar to the GEOMETRY datatype but takes the roundness of the earth into account.

This is very useful for setting geographical points with a high degree of exactitude.

### DIFF
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -836,6 +836,31 @@ GEOMETRY.prototype.$stringify = function (value) {
   return 'GeomFromText(\'' + Wkt.stringify(value) + '\')';
 };
 
+/**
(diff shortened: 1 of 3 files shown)


janmeier:
Could you add an integration test along the lines of https://github.com/sequelize/sequelize/blob/master/test/integration/model/geometry.test.js#L50-L64. Since this is so similar to geometry, I think we only need a single test to ensure that data is inserted and parsed back correctly.

Then a changelog entry, and we should be good to go :)

onzag:
I was expecting that :) currently I'm a bit busy as I'm getting ready for a trip, once I'm free I'll modify it.

Issue #2084, opened 2014-07-24
Max identifier length exceeded with nested includes


seth-admittedly:
Postgres has a 63 character max for identifiers, which is pretty hard to hit in normal use, and pretty easy to hit with the current way nested includes are namespaced. Unfortunately this parameter is hardcoded into Postgres and can only be increased by changing a source file and recompiling.

mohlman3:
+1 .. a fix or workaround would be very helpful

majodev:
+1

emhagman:
+1

renou:
+1

pmettraux:
+1

SharkofMirkwood:
+1

thomaschaaf:
/12

shawnbissell:
+1000000000

josser:
Sorry for dup, but maybe somebody found workaround or quick fix? 
Quick idea: what if we will generate UUID for each 'as' which will be mapped to hash of real 'as' identifiers during query execution. 
If this is good fix I may try to create PR for this.

janmeier:
@josser As you can see in the top, we've discussed using some sort of minification scheme - it would probably be better to use some shortened version of the model name, so it's at least somewhat readable.

Unless @mickhansen has any strong objections, a PR would be welcome - It's not a pretty solution, but I don't think we can do much better - not all queries can be solved by `.seperate` i think

robraux:
:+1: here as well, just hit this again in production. Separate works for some scenario's but there are other areas it does not.

Srokap:
+1

ryanrolds:
:+1:

linpekka:
I have run into this, and was looking into if there is a way to at least get a nice error from sequelize when it happens. I noticed that setting log_min_messages = notice in postgresql.conf will generate a notice in the postgres log, but even with client_min_messages = notice I am not seeing any error or log generated by sequelize for the truncate. 

Is there some option that I missed that will allow the notice to be propagated to the sequelize log, or even better, to be able to get an error for this specific notification.

tan784147:
+1
separate works only with certain associations. would be great to have a solution cross all associations.

bockscar:
+1

DeX3:
+1 agreed, I can't even use the `separate: true` workaround because I have lots of nested `belongsTo`-relations and `separate: true` is only supported for `hasMany` :/

gossel-j:
+1
[...]

PR #5190, opened 2016-01-11
Fix issue with nested separate includes

mliszewski:
We ran into an issue where we had nested has-many includes. We were using separate=true on these to avoid https://github.com/sequelize/sequelize/issues/2084. We found that if the parent include had a limit set and the nested include did not that incorrect sql was being generated because the options.groupedLimit attribute was not being cleared from the parent. The test I added will fail with a syntax error against the current master code.


mickhansen:
LGTM, rerunning two failed builds

Issue #1626, opened 2014-04-11
afterValidate hook is not called when validation fails


alekbarszczewski:
When validation errors occur (for example during `Model.create(...)` then `afterValidate` hook is not called. It would be a good feature to be able to catch validation errors in a hook - is it possible?

If not maybe you could implement this functionality in new version?
For example like this:
```
hooks: { afterValidate: function(model, done) { console.error(model.validationErrors)... } }
```
or
```
hooks: { afterValidationFailed: function(err, model, done) { ... } }
```
or something like this.

ptusch:
*Push*

It would be nice to work with failed validation.

janmeier:
Added in https://github.com/sequelize/sequelize/issues/5196

PR #5196, opened 2016-01-12
Added validationFailed hook

skleeschulte:
See [ afterValidate hook is not called when validation fails #1626 ](https://github.com/sequelize/sequelize/issues/1626)

I added an async validationFailed hook which is called with modelInstance, options and validationError. If the hook throws an error, it replaces the original validation error.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,5 @@
+- [ADDED] validationFailed hook [#1626](https://github.com/sequelize/sequelize/issues/1626)
+
 # 3.17.1
 - [FIXED] Reverted benchmarking feature since it does not compile on Node v4.0
 
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -1,6 +1,6 @@
 <a name="hooks"></a>
 # Mixin Hooks
-[View code](https://github.com/sequelize/sequelize/blob/0de404640d4c71e2d1f1259356650dfb586a248b/lib/hooks.js#L39)
+[View code](https://github.com/sequelize/sequelize/blob/ea444d318402f5de7f13f2c8dd6512fe0b8a38d8/lib/hooks.js#L39)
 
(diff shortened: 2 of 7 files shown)


mickhansen:
Please remove docs/api/ from the PR, making it easier to review.
We don't generally have hooks for failures, what's your usecase here?

skleeschulte:
My usecase: I need to customize the error message for the internal validator for `allowNull = false`. I need to do this while having a reference to the model instance. Doing it "from the outside" would mean that I have to wrap every create and update request.

By providing a hook for validation failures with the possibility to modify the thrown error one would get all the flexibility in generating errors in the desired format, e.g. custom messages for other inbuilt validators, filtering validation errors for specific fields, ...

As you can see in [afterValidate hook is not called when validation fails #1626](https://github.com/sequelize/sequelize/issues/1626), others have requested this feature too, and your statement about this was:

> But yeah we could definitely add a validation failed hook i suppose, that wouldn't interfere with anything.

mickhansen:
Thanks for linking my previous opinion, i forgot that i wrote that, but i still do agree with myself.
I'm all for this hook, just wanted to make sure we weren't solving a problem with the wrong method :)

mickhansen:
Build failure unrelated, can you remove the docs/api/ changes, it's all autogenerated and just messes up the diff.

Should the name instead be `afterValidationFailure`? I'm not sure, it's not really an event that would have a "before". /cc @janmeier

skleeschulte:
Ok :)
I reverted the changes to `docs/api/hooks.md` (I didn't commit the changes to the other api docs files in the first place).
I'm indifferent about `validationFailed` vs `afterValidationFailure`.

janmeier:
`validationFailed` it is - I agree it doesn't make sense to have after when there is no before

PR #5203, opened 2016-01-13
	Fixes a bug with user defined attribute names overwriting builtins

rmzg:
If you define an attribute on a model with the same name as a built-in
instance method the attribute's automatically created get/set methods
will overwriting the built-in causing much hilarity.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -196,7 +196,11 @@ Instance.prototype.get = function(key, options) { // testhint options:none
         return this.dataValues[key];
       }
     }
-    return this.dataValues[key];
+		if (this.dataValues) {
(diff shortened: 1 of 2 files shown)


lib/instance.js
```diff
@@ -196,7 +196,11 @@ Instance.prototype.get = function(key, options) { // testhint options:none
         return this.dataValues[key];
       }
     }
-    return this.dataValues[key];
+		if (this.dataValues) {
```
mickhansen: `dataValues` shouldn't really ever be undefined as far as i know?

rmzg:
Trying again with better commits!

Issue #4683, opened 2015-10-19
beforeFind hook is not executed on Model.count()


alekbarszczewski:
It seems that `beforeFind` hook is not executed on `Model.count()`. I am not sure if this is intentional, but it would be cool if this could work. For example I am using `beforeFind` hook to apply custom query conditions (based on options passed to query) and when I am using `findAndCount` then those custom conditions are applied only to `findAll` call and not for `count` call. So I end up with count = 10 and 30 rows for example.

I think there are three options:

1. Pass extra option to count/findAndCount, for example `beforeCount: true` that will trigger `beforeFind`  also for `count` method. This wont break BC.

2. Just run beforeFind hook for count - this may break BC?

3. Add new hook `beforeCount`. This wont break BC.

What do you think @janmeier @mickhansen ?

janmeier:
I think a separate `beforeCount` hook is the best option

sushantdhiman:
Hi @janmeier , we can close this one

PR #5209, opened 2016-01-13
add a beforeCount hook

matthewheck:
This provides a solution to #4683. This provides a beforeCount hook to the count query, allowing for the user to change the options of count before the database is queried.

I attempted to add an afterCount hook but, because the count is not attached to the model itself, changing it after the database return would require more substantial (probably breaking) changes. Additionally, can't think of a good use case for changing the count after we have queried the database.

Long time user, first time contributor. Hoping to give back more to this great community!

### DIFF
--- a/test/integration/hooks.test.js
+++ b/test/integration/hooks.test.js
@@ -990,52 +990,6 @@ describe(Support.getTestDialectTeaser('Hooks'), function() {
     describe('on success', function() {
       it('hook runs', function() {
         var beforeHook = false;
-        this.User.beforeCount(function() {
-          beforeHook = true;
-        });
-
-        return this.User.count().then(function(count) {
-          expect(count).to.equal(3);
-          expect(beforeHook).to.be.true;
-        });
-      });
-
-      it('beforeCount hook can change options', function() {
-        this.User.beforeCount(function(options) {
-          options.where.username = 'adam';
-        });
-
-        return this.User.count({where: {username: 'joe'}}).then(function(count) {
-          expect(count).to.equal(1);
-        });
-      });
-    });
-
-    describe('on error', function() {
-      it('in beforeCount hook returns error', function() {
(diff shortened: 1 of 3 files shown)


matthewheck:
Accidentally deleted a semicolon in my initial commit, then messed up when I squashed the commits. Looks like travis-ci has now stalled out. Sorry about that!

test/integration/hooks.test.js
```diff
@@ -978,6 +978,99 @@ describe(Support.getTestDialectTeaser('Hooks'), function() {
...
+    });
+
+    describe('on error', function() {
+      it('in beforeCount hook returns error', function() {
+        this.User.beforeCount(function() {
+          throw new Error('Oops!');
+        });
+
+        return this.User.find({where: {username: 'adam'}}).catch (function(err) {
+          expect(err.message).to.equal('Oops!');
```
mickhansen: You can clean up assertions a bit by using stuff like:

```js
return expect(this.User.count({where: {username: 'joe'}})).to.eventually.equal(1);
return expect(this.User.count({where: {username: 'joe'}})).to.be.rejectedWith('Oops!');
```

matthewheck: Thanks for the tip. Clean up done!

mickhansen:
LGTM, @janmeier?

janmeier:
Yep, LGTM as well - thanks a lot @matthewheck , and sorry for leaving you waiting for so long

emersonmsantos7:
BeforeCount is not in the documentation: http://docs.sequelizejs.com/en/v3/api/hooks/

josephgrossberg:
Update: still no mention in the docs http://docs.sequelizejs.com/

This is a cool feature that merits documenting. How can I help? ping @sdepold

aat2703:
Why not just firing beforeFind hook? It seems odd

PR #5219, opened 2016-01-14
Fixes a bug with user defined attribute names overwriting builtins.

rmzg:
If you define an attribute on a model with the same name as a built-in
instance method the attribute's automatically created get/set methods
will overwriting the built-in causing much hilarity.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -74,10 +74,11 @@ var Model = function(name, attributes, options) {
     }
   });
 
-  this.attributes = this.rawAttributes = Utils._.mapValues(attributes, function(attribute, name) {
-    if (!Utils._.isPlainObject(attribute)) {
-      attribute = { type: attribute };
-    }
+ this.attributes = this.rawAttributes = Utils._.mapValues(attributes, function(attribute, name) {
+   if (!Utils._.isPlainObject(attribute)) {
+     attribute = { type: attribute };
+   }
+
 
     attribute = this.sequelize.normalizeAttribute(attribute);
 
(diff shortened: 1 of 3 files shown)


mickhansen:
A few linting errors.
Using a unit test should be sufficient in this case, simply testing that `set` still works after having an attribute named the same.

rmzg:
Updated.

test/unit/model/overwriting-builtins.test.js
```diff
@@ -0,0 +1,24 @@
...
+/* jshint -W110 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../../support')
+  , DataTypes = require(__dirname + '/../../../lib/data-types');
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+
+  describe('not breaking built-ins', function() {
+    it('test set breakage', function() {
```
mickhansen: `'it should not break instance.set by defining a model set attribute'`

lib/model.js
```diff
@@ -74,10 +74,11 @@ var Model = function(name, attributes, options) {
...
   });
 
-  this.attributes = this.rawAttributes = Utils._.mapValues(attributes, function(attribute, name) {
-    if (!Utils._.isPlainObject(attribute)) {
-      attribute = { type: attribute };
-    }
+ this.attributes = this.rawAttributes = Utils._.mapValues(attributes, function(attribute, name) {
+   if (!Utils._.isPlainObject(attribute)) {
+     attribute = { type: attribute };
+   }
```
mickhansen: didn't you remove this previously?

rmzg:
Updated the test script comment.

PR #5226, opened 2016-01-15
Fix 5092

### LINKED ISSUE #5092: Postgres JSONB 'destroy' with 'where' object fails
I'm using sequelize 3.15.0, postgres dialect on Mac OS X (El Cap). I have a column with JSONB
type. When destroying by using where object, it throws an exception.

```javascript
var Page = sequelize.define('Page', {
  content: Sequelize.JSONB
})
```

```javascript
Page.create({ name: "Test", content: { title: "Title" } })
.then(() => {
  Page.destroy({ where: { content: { title: "Title" } } })
  .then(() => {
    sequelize.close()
  })
})
```

It throws.

```
Unhandled rejection TypeError: val.replace is not a function
    at Object.SqlString.escape (/tmp/myapp/node_modules/sequelize/lib/sql-string.js:56:15)
    at Object.QueryGenerator.escape (/tmp/myapp/node_modules/sequelize/lib/dialects/abstract/query-generator.js:983:22)
    ...
```

When using netsted key `content.title`, it also throws an exception though a different one.

```javascript
Page.destroy({ where: { 'content.title': "Title" } })
```

```
Unhandled rejection SequelizeDatabaseError: column "content.title" does not exist
    at Query.formatError (/tmp/myapp/node_modules/sequelize/lib/dialects/postgres/query.js:347:14)
    at null.<anonymous> (/tmp/myapp/node_modules/sequelize/lib/dialects/postgres/query.js:81:19)
    ...
```

However, using that same 'where' object with `findOne` or `findAll` works perfectly as expected.

sushantdhiman:
Fix #5092 , where `destroy` fails on `JSONB` data with where condition.

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -375,7 +375,7 @@ var QueryGenerator = {
 
     var replacements = {
       table: this.quoteIdentifiers(tableName),
-      where: this.getWhereConditions(where),
+      where: this.getWhereConditions(where, null, model, options),
       limit: !!options.limit ? ' LIMIT ' + this.escape(options.limit) : '',
       primaryKeys: primaryKeys[tableName].length > 1 ? '(' + pks + ')' : pks,
       primaryKeysSelection: pks
--- a/test/integration/model/json.test.js
+++ b/test/integration/model/json.test.js
@@ -312,6 +312,14 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       });
 
       it('should be possible to destroy with where', function () {
+        var conditionSearch = {
+          where: {
+            data: {
+              employment : 'Hacker'
+            }
+          }
+        };
+
         return Promise.join(
           this.Event.create({
             data: {
@@ -330,16 +338,23 @@ describe(Support.getTestDialectTeaser('Model'), function() {
               },
               employment: 'Hacker'
             }
+          }),
+          this.Event.create({
+            data: {
+              name: {
+                first: ' Tyrell',
+                last: 'Wellick'
+              },
+              employment: 'CTO'
+            }
           })
         ).bind(this).then(function () {
-          return this.Event.destroy({
-            where: {
-              data: {
-                employment : 'Hacker'
-              }
-            }
+            return expect(this.Event.findAll(conditionSearch)).to.eventually.have.length(2);
+          }).then(function() {
+            return this.Event.destroy(conditionSearch);
+          }).then(function(){
+            return expect(this.Event.findAll(conditionSearch)).to.eventually.have.length(0);
           });
-        });
       });
 
     });
--- a/test/integration/model/json.test.js
+++ b/test/integration/model/json.test.js
@@ -310,6 +310,38 @@ describe(Support.getTestDialectTeaser('Model'), function() {
           });
         });
       });
+
+      it('should be possible to destroy with where', function () {
+        return Promise.join(
+          this.Event.create({
+            data: {
+              name: {
+                first: 'Elliot',
+                last: 'Alderson'
+              },
+              employment: 'Hacker'
+            }
+          }),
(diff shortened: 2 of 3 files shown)


sushantdhiman:
Ping @mickhansen , @janmeier

test/integration/model/json.test.js
```diff
@@ -310,6 +310,47 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
+              employment: 'Hacker'
+            }
+          }),
+          this.Event.create({
+            data: {
+              name: {
+                first: 'Christian',
+                last: 'Slater'
+              },
+              employment: 'Hacker'
```
janmeier: Could you give one row another employment - theoretically this test also passed if the table is truncated

lib/dialects/postgres/query-generator.js
```diff
@@ -375,7 +375,7 @@ var QueryGenerator = {
 
     var replacements = {
       table: this.quoteIdentifiers(tableName),
-      where: this.getWhereConditions(where),
+      where: this.getWhereConditions(where, null, model, options),
```
mickhansen: `getWhereConditions` is deprecated, perhaps change it so it calls the new where generator directly.

sushantdhiman: What is the new where generator?

test/integration/model/json.test.js
```diff
@@ -310,6 +310,47 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
+            }
+          })
+        ).bind(this).then(function () {
+          return this.Event.findAll(conditionSearch).bind(this).then(function(records){
+            expect(records.length).to.be.equal(2);
+            return this.Event.destroy(conditionSearch).bind(this).then(function(){
```
janmeier: Callback chain of dooooooom!

```js
).bind(this).then(function () {
	return expect(this.Event.findAll(conditionSearch)).to.eventually.have.length(2)
}).then(function() {
	return this.Event.destroy(conditionSearch);
}).then(function(){
   expect(this.Event.findAll(conditionSearch)).to.eventually.have.length(2);
});
```

test/integration/model/json.test.js
sushantdhiman: Good point

lib/dialects/postgres/query-generator.js
mickhansen: whereItemsQuery in the same file, getWhereConditions mostly proxies to that, i'm not sure if it actually does anything anymore.

sushantdhiman: `whereItemsQuery` , Ok will run tests with that

test/integration/model/json.test.js
sushantdhiman: Ok will nest them

sushantdhiman:
Done @mickhansen , @janmeier

test/integration/model/json.test.js
sushantdhiman: Add a new row of data with different employment

lib/dialects/postgres/query-generator.js
sushantdhiman: @mickhansen `whereItemsQuery` will not work as `getWhereConditions` handle the perpending of primary key from model. Otherwise it generates queries like this

`WHERE 1 LIMIT 1 ` instead of `WHERE id = 1 LIMIT 1`

mickhansen: Ah i see, thought we had killed all the syntax that simply uses where: 1.
Ok this is fine for now then.

mickhansen:
Build is getting a few jshint errors, this is fixed on master though, can you rebase?

sushantdhiman:
@mickhansen Rebased

mickhansen:
Thanks @sushantdhiman

PR #5231, opened 2016-01-15
Add `IF EXISTS` to Postgres drop schema query

nunofgs:
This PR adds `IF EXISTS` to the "drop schema" query, similar to the "create table" query found [here](https://github.com/seegno-forks/sequelize/blob/342251100f9040cd21cff4d36a0c3edeabd21a8f/lib/dialects/postgres/query-generator.js#L97).

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -23,7 +23,7 @@ var QueryGenerator = {
   },
 
   dropSchema: function(schema) {
-    var query = 'DROP SCHEMA <%= schema%> CASCADE;';
+    var query = 'DROP SCHEMA IF EXISTS <%= schema%> CASCADE;';
     return Utils._.template(query)({schema: schema});
   },
 
--- a/test/unit/sql/create-schema.test.js
+++ b/test/unit/sql/create-schema.test.js
@@ -1,0 +1,18 @@
+'use strict';
+
+var Support   = require(__dirname + '/../support')
+  , expectsql = Support.expectsql
+  , current   = Support.sequelize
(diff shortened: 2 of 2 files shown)


test/integration/dialects/postgres/query-generator.test.js
```diff
@@ -220,6 +220,13 @@ if (dialect.match(/^postgres/)) {
         }
       ],
 
+      dropSchema: [
+        {
+          arguments: ['mySchema'],
+          expectation: 'DROP SCHEMA IF EXISTS mySchema CASCADE;'
+        },
+      ],
```
mickhansen: SQL unit tests should be in test/unit/sql/*
the SQL tests in test/integration are an outdated thing i'm afraid.

nunofgs:
@mickhansen Updated.

mickhansen:
For create table IF EXISTS is only available after a certain version: https://github.com/sequelize/sequelize/blob/master/test/unit/sql/create-table.test.js#L31
Mind adding the same check for schema so we don't break anything?

nunofgs:
@mickhansen The "drop schema if exists" instruction exists since Postgres 8.2. I took a look and we're not verifying versions that old, for example: https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/query-generator.js#L97.

Do you still want me to go ahead and add a version check for >= 8.2 ?

mickhansen:
if it's that old don't worry about it, was more worried about it being a version 9 thing.

mickhansen:
(guess they added DROP IF EXISTS before CREATE IF EXISTS for most stuff)

PR #5236, opened 2016-01-17
Code fixes for jshint@2.9.1

ajfranzoia:
Minor code fixes to prevent ```[variable] was used before it was defined``` error introduced by jshint@2.9.1.
If any modification is needed, just let me know.


mickhansen:
That's some lazy evaluation by jshint since all those cases are inside function calls :p

Issue #5200, opened 2016-01-12
Model.update({ ... }, {silent: true})


astephenb:
`silent` is an option when updating an instance, but not when doing bulk updates. Is it possible to do a bulk update without updating timestamps?

```js
Model.update({ foo: 'bar' }, { silent: true });
```

mickhansen:
Not currently it looks like, other than using `fields`.

astephenb:
Thanks a lot! I will close this issue.

PR #5238, opened 2016-01-17
Support silent: true in bulk update

ajfranzoia:
Support silent: true option in bulk update as requested in [#5200](https://github.com/sequelize/sequelize/issues/5200)
If any modification is needed, just let me know.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# FUTURE
+- [ADDED] Support silent: true in bulk update [#5200](https://github.com/sequelize/sequelize/issues/5200)
+
 # 3.17.3
 - [FIXED] Regression with array values from security fix in 3.17.2
 
--- a/test/integration/instance.test.js
+++ b/test/integration/instance.test.js
@@ -1085,6 +1085,37 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
       });
     });
 
+    it('does not update timestamps when passing silent=true in a bulk update', function() {
+      var self = this
+        , updatedAtPeter
+        , updatedAtPaul
+        , data = [
+          { username: 'Paul' },
+          { username: 'Peter' }
+        ];
+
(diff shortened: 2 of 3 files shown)


changelog.md
```diff
@@ -1,3 +1,6 @@
+# 3.17.4
+- [ADDED] Support silent: true in bulk update [#5200](https://github.com/sequelize/sequelize/issues/5200)
+
```
mickhansen: Don't add your own version, just put feature.
A feature addition should be a minor bump (not a patch) in any case, but it's best to just leave it up to us depending on what and when release (also does not make people think there's a new release by looking at the changelog :))

test/integration/instance.test.js
```diff
@@ -1085,6 +1085,37 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
       });
     });
 
+    it.only('does not update timestamps when passing silent=true in a bulk update', function() {
```
mickhansen: whoops

test/integration/instance.test.js
```diff
@@ -1085,6 +1085,37 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
...
+        ];
+
+      return this.User.bulkCreate(data).then(function() {
+        return self.User.findAll();
+      }).then(function(users) {
+        updatedAtPaul = users[0].updatedAt;
+        updatedAtPeter = users[1].updatedAt;
+      })
+      .then(function() {
+        return self.sequelize.Promise.delay(2000);
```
mickhansen: A 150ms delay should be sufficient for the test.

ajfranzoia:
Changes applied. Thanks for the feedback.

mickhansen:
@ajfranzoia Still got the `.only` in there though :)

ajfranzoia:
Oops, it escaped me. Fixed.

PR #5243, opened 2016-01-17
Fix #5222: prevent race condition after transaction finished

### LINKED ISSUE #5222: Transaction `finished` race condition
There is logic in a transaction that basically flags a transaction as `finished`, which is then checked against when querying against that transaction. However, there is a race condition as both `.rollback` and `.commit` don't set this flag until after the "ROLLBACK" and "COMMIT" queries have succeeded. 

See https://github.com/sequelize/sequelize/blob/master/lib/transaction.js#L209 that the flag is set in the `finally`
See the guard at https://github.com/sequelize/sequelize/blob/2a168da8c56b9d9313d06ea6a7dcb683c5b7e891/lib/sequelize.js#L777

Basically something like below is happening:
```
transaction.rollback()
  queryInterface.rollbackTransaction()
...
some other code gets to run and does an update
...
  .finally is triggered
```
=============
The below code will reproduce

```javascript
database.transaction(function(t){
  // Forced, but original use case was something like Promise.all([...]).catch (err) -> t.rollback()
  t.rollback();
  database.query("UPDATE entities SET name='something or other'");
});
```

Observe something like 
```
Executing (5f84266c-9e62-4c96-8040-531e8d765ede): START TRANSACTION;
...
Executing (5f84266c-9e62-4c96-8040-531e8d765ede): ROLLBACK;
...
Executing (5f84266c-9e62-4c96-8040-531e8d765ede): UPDATE entities SET name`='something or other'
```

The update goes out even though the transaction was just rolled back. 
I believe the fix would be to just set the `finished` flag right after sending the ROLLBACK out, rather than in the .finally.

Minor overall, as easily avoided by just making sure all your parallel requests are done before running the rollback.

ajfranzoia:
Fixes issue #5222. 
Moves ```transaction.finished = 'rollback'``` assignment to ```QueryInterface.rollbackTransaction()``` after rollback query call.

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -211,11 +211,10 @@ Transaction.prototype.rollback = function() {
     .getQueryInterface()
     .rollbackTransaction(this, this.options)
     .finally(function() {
-      self.finished = 'rollback';
       if (!self.parent) {
         return self.cleanup();
       }
-      return null;
+      return self;
     });
 };
 
--- a/test/integration/transaction.test.js
+++ b/test/integration/transaction.test.js
@@ -129,6 +129,20 @@ describe(Support.getTestDialectTeaser('Transaction'), function() {
     ).to.eventually.be.rejected;
   });
 
+  it('does not allow queries immediatly after commit call', function() {
+    var self = this;
+    return expect(
+      this.sequelize.transaction().then(function(t) {
+        return self.sequelize.query('SELECT 1+1', {transaction: t, raw: true}).then(function() {
+          return Promise.join(
+            expect(t.commit()).to.eventually.be.fulfilled,
+            expect(self.sequelize.query('SELECT 1+1', {transaction: t, raw: true})).to.eventually.be.rejectedWith(/commit has been called on this transaction\([^)]+\), you can no longer use it/)
+          );
+        });
+      })
+    ).to.be.eventually.fulfilled;
+  });
+
   it('does not allow queries after rollback', function() {
     var self = this;
     return expect(
@@ -142,13 +156,27 @@ describe(Support.getTestDialectTeaser('Transaction'), function() {
     ).to.eventually.be.rejected;
   });
 
+  it('does not allow queries immediatly after rollback call', function() {
+    var self = this;
+    return expect(
+      this.sequelize.transaction().then(function(t) {
+        return Promise.join(
+          expect(t.rollback()).to.eventually.be.fulfilled,
+          expect(self.sequelize.query('SELECT 1+1', {transaction: t, raw: true})).to.eventually.be.rejectedWith(/rollback has been called on this transaction\([^)]+\), you can no longer use it/)
+        );
+      })
+    ).to.eventually.be.fulfilled;
+  });
+
   it('does not allow commits after commit', function () {
     var self = this;
-    return expect(self.sequelize.transaction().then(function (t) {
-      return t.commit().then(function () {
-        return t.commit();
-      });
-    })).to.be.rejectedWith('Error: Transaction cannot be committed because it has been finished with state: commit');
+    return expect(
+      self.sequelize.transaction().then(function (t) {
+        return t.commit().then(function () {
+          return t.commit();
+        });
+      })
+    ).to.be.rejectedWith('Error: Transaction cannot be committed because it has been finished with state: commit');
   });
 
   it('does not allow commits after rollback', function () {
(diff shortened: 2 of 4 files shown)


test/integration/transaction.test.js
```diff
@@ -142,6 +142,16 @@ describe(Support.getTestDialectTeaser('Transaction'), function() {
     ).to.eventually.be.rejected;
   });
 
+  it('does not allow queries immediatly after rollback call', function() {
+    var self = this;
+    return expect(
+      this.sequelize.transaction().then(function(t) {
+        t.rollback();
+        return self.sequelize.query('SELECT 1+1', {transaction: t, raw: true});
```
ajfranzoia: It's failing for sqlite because the rollback query is still running after the ```SELECT 1+1``` query call, and it expects no running queries when the test finishes. I can't think of any ideas to prevent this and at the same time return the query promise that will be rejected.

lib/transaction.js
```diff
@@ -211,7 +211,6 @@ Transaction.prototype.rollback = function() {
     .getQueryInterface()
     .rollbackTransaction(this, this.options)
     .finally(function() {
-      self.finished = 'rollback';
```
mickhansen: Same should likely be applied for commit i imagine then?

test/integration/transaction.test.js
mickhansen: Hmm yeah, that's tricky, maybe you can do:

```js
return Promise.join(
  t.rollback(),
 self.sequelize.query('SELECT 1+1', {transaction: t, raw: true})
);
```

test/integration/transaction.test.js
```diff
@@ -142,6 +142,16 @@ describe(Support.getTestDialectTeaser('Transaction'), function() {
...
+      })
+    ).to.eventually.be.rejected;
```
mickhansen: `to.eventually.be.rejectedWith(errorMessage)`, best to be specific about what kind of error we're expecting.

lib/transaction.js
ajfranzoia: Yeah, I thought the same. Do I create another PR or modify this one?

mickhansen: Same PR is good

ajfranzoia: Ok. Will modify the PR during the next days.

test/integration/transaction.test.js
```diff
@@ -129,6 +129,20 @@ describe(Support.getTestDialectTeaser('Transaction'), function() {
     ).to.eventually.be.rejected;
   });
 
+  it('does not allow queries immediatly after commit call', function() {
+    var self = this;
+    return expect(
+      this.sequelize.transaction().then(function(t) {
+        return self.sequelize.query('SELECT 1+1', {transaction: t, raw: true}).then(function() {
+          return Promise.join(
+            expect(t.commit()).to.eventually.be.fulfilled,
```
ajfranzoia: I had to use this style for assertions because not using the ```expect()``` wrapping inside ```Promise.join(...)``` causes the query to be rejected before the rollback/commit query is finished (since we're using an immediate call after). Thus, this prevents getting the 'There are still running queries' error thrown by ```sequelize.test.verifyNoRunningQueries()``` in ```afterEach()``` callback.
I utilized this approach following the suggestion on http://chaijs.com/plugins/chai-as-promised (see 'Multiple Promise Assertions' section).

test/integration/transaction.test.js
```diff
@@ -129,6 +129,20 @@ describe(Support.getTestDialectTeaser('Transaction'), function() {
...
+            expect(self.sequelize.query('SELECT 1+1', {transaction: t, raw: true})).to.eventually.be.rejectedWith(/commit has been called on this transaction\([^)]+\), you can no longer use it/)
+          );
+        });
+      })
+    ).to.be.eventually.fulfilled;
```
ajfranzoia: I had to use this style for assertions because not using the ```expect()``` wrapping inside ```Promise.join(...)``` causes the query to be rejected before the rollback/commit query is finished (since we're using an immediate call after). Thus, this prevents getting the 'There are still running queries' error thrown by ```sequelize.test.verifyNoRunningQueries()``` in ```afterEach()``` callback.
I utilized this approach following the suggestion on http://chaijs.com/plugins/chai-as-promised (see 'Multiple Promise Assertions' section).

mickhansen:
Good work :)

PR #5249, opened 2016-01-18
Updated the fix for user defined attributes overwriting builtins.

rmzg:
More appropriate fix for the merge: https://github.com/sequelize/sequelize/pull/5219
Now uses hasOwnProperty to more accurately check if the user attribute
matches a builtin.


janmeier:
Cool, thanks :)

PR #5253, opened 2016-01-19
Prevent Promise (bluebird) warnings in tests (part 2)

tornillo:
Remove Promise warnings with type:
```Warning: a promise was rejected with a non-error: [object Object]```

### DIFF
--- a/test/integration/hooks.test.js
+++ b/test/integration/hooks.test.js
@@ -1161,7 +1161,7 @@ describe(Support.getTestDialectTeaser('Hooks'), function() {
               , afterProject = false
               , beforeTask = false
               , afterTask = false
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: test/integration/hooks.test.js [hunk 1]
```diff
@@ -1161,7 +1161,7 @@ describe(Support.getTestDialectTeaser('Hooks'), function() {
               , afterProject = false
               , beforeTask = false
               , afterTask = false
-              , VeryCustomError = function() {};
+              , CustomErrorText = 'Whoops!';
```
mickhansen: The issue being that VeryCustomError does not properly inherit from Error?

tornillo: Yes.

PR #5263, opened 2016-01-20
Added retry and options to retry when queries fail with SQL_BUSY state.

EToreo:
Added retry on SQL_BUSY state.  Only works for Sqlite right now, but others can be added easily.

Replaces https://github.com/sequelize/sequelize/pull/5157/


mickhansen:
Build errors specifically for the test you just added.

EToreo:
The guard I put around the retry tests should have prevented mysql from running them (it had worked in the previous 2 attempts)...  Anyway, I upgraded the guard to explicitly call out sqlite.

After that update the newly failing tests look unrelated and probably just need to be re-run?

EToreo:
Thank you!  And my apologies for making this so hard, its been a real learning experience for me.

mickhansen:
I'm just glad you stuck with it through all this :)

EToreo:
Well thank you.  Me too.  My team has big plans for sequelize.  Rumor has it they are nearing completion on their first pull request for recursive CTEs.

Issue #4844, opened 2015-11-12
dao.get({ plain: true }) doesn't work for reloaded objects


garrett-gottlieb:
Consider the following example:
```js
var user = yield User.create({ settingsId })
yield user.reload({ include: settingsAssociation })

_.isPlainObject(user.get({ plain: true }).settings) // false
```

It works when the dao is retrieved using querying, just not when reloaded with `include`.

mickhansen:
`reload` does not update the internal options of the instance (which contains what includes to iterate while doing a get).

Issue #4452, opened 2015-09-08
Reload doesn't check if the reloaded instance is defined


glutentag:
Hello,

When we reload an instance, it doesn't check if the returned object is defined:
```
 return this.Model.findOne(options).bind(this).then(function(reload) {
    this.set(reload.dataValues, {
      raw: true,
      reset: true && !options.attributes
    });
  }).return(this);
```
Returning us a TypeError: 
`Cannot read property 'dataValues' of null`

It would be nice to have an error to catch there. Should it be a native Error or a SequelizeErrors?

mickhansen:
I suppose this would happen in cases where the object you're attempting to reload has been deleted in the meantime?

glutentag:
Exactly.

PR #5268, opened 2016-01-20
Fix Instance.reload issues

tornillo:
Refs to https://github.com/sequelize/sequelize/issues/4844 and https://github.com/sequelize/sequelize/issues/4452

1. properly updates Instance options after reload, that allows access to included association via ```.get({ plain: true}).settings``` or dot notation ```user.settings```
2. throws a error when reloaded instance is not exist in DB

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -5,6 +5,7 @@ var Utils = require('./utils')
   , BelongsToMany = require('./associations/belongs-to-many')
   , InstanceValidator = require('./instance-validator')
   , QueryTypes = require('./query-types')
+  , sequelizeErrors = require('./errors')
   , Dottie = require('dottie')
   , Promise = require('./promise')
   , _ = require('lodash')
@@ -742,7 +743,18 @@ Instance.prototype.reload = function(options) {
     include: this.$options.include || null
   });
 
-  return this.Model.findOne(options).bind(this).then(function(reload) {
+  return this.Model.findOne(options).bind(this)
+  .tap(function (reload) {
+    if (!reload) {
+      throw new sequelizeErrors.InstanceError(
+        'Instance could not be reloaded because it does not exist anymore (find call returned null)'
+      );
+    }
+  })
+  .then(function(reload) {
+    // update the internal options of the instance
+    this.$options = reload.$options;
+    // re-set instance values
     this.set(reload.dataValues, {
       raw: true,
       reset: true && !options.attributes
--- a/test/integration/instance.test.js
+++ b/test/integration/instance.test.js
@@ -513,6 +513,49 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
       });
     });
 
+    it('should update internal options of the instance', function() {
+      var Book = this.sequelize.define('Book', { title: DataTypes.STRING })
+        , Page = this.sequelize.define('Page', { content: DataTypes.TEXT });
+
+      Book.hasMany(Page);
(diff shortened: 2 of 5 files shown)


lib/instance.js
```diff
@@ -742,7 +743,16 @@ Instance.prototype.reload = function(options) {
     include: this.$options.include || null
   });
 
-  return this.Model.findOne(options).bind(this).then(function(reload) {
+  return this.Model.findOne(options).bind(this)
+  .tap(function (reload) {
+    if (!reload) {
+      throw new sequelizeErrors.InstanceError('Instance cannot be reloaded because does not exist in the database');
```
mickhansen: `Instance could not be reloaded because it does not exist anymore (find call returned null)`

tornillo:
I fixed a message... and did a wrong rebase, sorry.

mickhansen:
You'll need to fix the PR, we generally want single descriptive commits without merge history :)

tornillo:
I can not properly rebase my PR. I'm newbie in git...
So maybe I need close this PR and create a new one?

janmeier:
Have you tried 
```bash
$ git rebase -i upstream/master (where upstream = the remote pointing at the sequelize repo)
$ git push -f master
```
That should remove the merge commits, and let you squash the two into one

The force push is needed because you are rewriting commits

tornillo:
Yes, I tried it, but merged commits still exists.
And at current moment ```git push -f``` tells me ```Everything up-to-date```

janmeier:
No they don't :)

tornillo:
I figured out:)

tornillo:
How about to accept this PR?:)

test/integration/instance.test.js
```diff
@@ -513,6 +513,49 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
...
+    it('should return an error when reload fails', function() {
+      return this.User.create({ username: 'John Doe' }).then(function(user) {
+        return expect(user.reload({
+          where: { id: 2 }
+        })).to.be.rejectedWith(
+          Sequelize.InstanceError,
+          'Instance could not be reloaded because it does not exist anymore (find call returned null)'
+        );
+      });
+    });
```
mickhansen: Please change the test so it attempts reloading a destroyed instance, would mimic real life a bit more.
`reload` should really not accept a where statement (but you don't have to fix this in this PR)

tornillo: > ```reload``` should really not accept a where statement (but you don't have to fix this in this PR)

Why? Sometimes I need to reload instance with a new restricting params and check it for passing or not.
But if you think this feature unnecessary, I can remove cloning ```where``` or even prohibit accepting (throw an exception).

mickhansen: That's a discussion for another issue, leave it in for now :)

tornillo: Well, I will correct the test for destroyed instance:)

mickhansen: Did you push it? I don't see it in the diff

tornillo: **I will correct**
I need some time to do it:)

mickhansen: Oh i read it as you already had, my bad :D

tornillo:
Ok, PR is ready to be merged:)

mickhansen:
Looks good, will merge when build is done.

PR #5280, opened 2016-01-21
fix usign hstore without model instance

renatoelias:
I tried to think of a test that is this bug, but could not find an acceptable place in the current suite.

It happens when you just have to connect and make a query to a table with hstore


mickhansen:
Ah so if you're doing a raw query the returns hstore results but haven't had a defined model so the dataype hasnt been setup.

LGTM, @janmeier ?

renatoelias:
@janmeier ?

janmeier:
Thanks :)

Issue #4868, opened 2015-11-16
Really slow query with include


super-cache-money:
I've got a really tiny postgres database (tables less than 500 rows altogether). I'm running the following query, and it's taking over 5s: 

    Job.findOne({
        where: {
            id: 162
        },
        include: [
            {
                model: models.ArtisanAssignment,
                include: [{
                    model: models.Call
                }, {
                    model: models.TeamComment,
                    attributes: ['id']
                }, {
                    model: models.StageAssignment
                }]
            },
            {
                model: models.Homeowner,
                include: [{
                    model: models.TeamComment,
                    attributes: ['id']
                }]
            },
            {
                model: models.TeamComment,
                attributes: ['id']
            }, {
                model: models.Call
            }
        ]
    }

Even weirder, however, is the raw query generated, when I enable logging. While the sequelize result takes a while, it's perfectly correct. If I run the raw query directly, it returns incorrect results. I get 1000's of records with the first row repeated. Here's the raw query generated:

    SELECT
      "Job"."id",
      "Job"."service",
      "Job"."subService",
      "Job"."details",
      "Job"."materialSupplied",
      "Job"."startDate",
      "Job"."startDateIsSpecific",
      "Job"."startDateIsFlexible",
      "Job"."area",
      "Job"."createdAt",
      "Job"."updatedAt",
      "Job"."HomeownerId",
      "Job"."TeamMemberId",
      "ArtisanAssignments"."id"                                   AS "ArtisanAssignments.id",
      "ArtisanAssignments"."wonJob"                               AS "ArtisanAssignments.wonJob",
      "ArtisanAssignments"."createdAt"                            AS "ArtisanAssignments.createdAt",
      "ArtisanAssignments"."updatedAt"                            AS "ArtisanAssignments.updatedAt",
      "ArtisanAssignments"."ArtisanId"                            AS "ArtisanAssignments.ArtisanId",
      "ArtisanAssignments"."JobId"                                AS "ArtisanAssignments.JobId",
      "ArtisanAssignments"."TeamMemberId"                         AS "ArtisanAssignments.TeamMemberId",
      "ArtisanAssignments.Calls"."id"                             AS "ArtisanAssignments.Calls.id",
      "ArtisanAssignments.Calls"."answered"                       AS "ArtisanAssignments.Calls.answered",
      "ArtisanAssignments.Calls"."createdAt"                      AS "ArtisanAssignments.Calls.createdAt",
[...]

PR #5282, opened 2016-01-22
Rewriting of the benchmarking feature

tornillo:
Refs to https://github.com/sequelize/sequelize/issues/2494
I rewrote benchmark logic. Query execution timing now passed as second argument to logger function.
If default (console.log) is used, message will shown as:
```log
 SELECT "User"."id" FROM "users" AS "User" WHERE "User"."id" = 123; 1
```
In custom logger function developer can write own logic based on this timing:
```js
function logger (sql, time /*ms*/) {
    if (time > 100) {
        // log to file or do any other action
    }
}
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,5 @@
 # Future
+- [ADDED] Benchmarking feature [#2494](https://github.com/sequelize/sequelize/issues/2494)
 - [ADDED] `beforeCount` hook [#5209](https://github.com/sequelize/sequelize/pull/5209)
 - [ADDED] `validationFailed` hook [#1626](https://github.com/sequelize/sequelize/issues/1626)
 - [FIXED] Mark index as `unique: true` when `type: 'UNIQUE'`. Fixes [#5351](https://github.com/sequelize/sequelize/issues/5351)
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -484,11 +484,6 @@ AbstractQuery.prototype.checkLoggingOption = function() {
     console.log('DEPRECATION WARNING: The logging-option should be either a function or false. Default: console.log');
     this.options.logging = console.log;
   }
-
-  if (this.options.logging === console.log) {
-    // using just console.log will break in node < 0.6
-    this.options.logging = function(s) { console.log(s); };
-  }
 };
 
 /**
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -30,17 +30,17 @@ Query.prototype.run = function(sql, parameters) {
   //do we need benchmark for this query execution
   var benchmark = this.sequelize.options.benchmark || this.options.benchmark;
 
-  if (!benchmark) {
-    this.sequelize.log('Executing (' + (this.connection.uuid || 'default') + '): ' + this.sql, this.options);
-  } else {
+  if (benchmark) {
     var queryBegin = Date.now();
+  } else {
+    this.sequelize.log('Executing (' + (this.connection.uuid || 'default') + '): ' + this.sql, this.options);
   }
 
   var promise = new Utils.Promise(function(resolve, reject) {
     self.connection.query(self.sql, function(err, results) {
 
       if (benchmark) {
-        self.sequelize.log('Executed (' + (self.connection.uuid || 'default') + '): ' + self.sql + ' in ' + (Date.now() - queryBegin) + 'ms', self.options);
+        self.sequelize.log('Executed (' + (self.connection.uuid || 'default') + '): ' + self.sql, (Date.now() - queryBegin), self.options);
       }
 
       if (err) {
--- a/test/integration/sequelize.test.js
+++ b/test/integration/sequelize.test.js
@@ -223,6 +223,8 @@ describe(Support.getTestDialectTeaser('Sequelize'), function() {
   describe('query', function() {
     afterEach(function() {
       this.sequelize.options.quoteIdentifiers = true;
+
+      console.log.restore && console.log.restore();
     });
 
     beforeEach(function() {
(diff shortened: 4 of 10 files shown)


sushantdhiman:
@tornillo I think we should allow passing a function to `benchmark` option. It should be provided the time difference argument. Otherwise if its `true` we should keep on printing regular `Executed ....` statements.

Because most of people would just want to know how much time it took to execute a query. With `benchmark : function(time) { }` they will be able to do custom logs. With `benchmark: true` normal logs like `Executed SELECT * FROM "users" in 23 ms` will be available.

changelog.md
```diff
@@ -1,3 +1,6 @@
+# Future
+- [ADDED] Benchmarking feature (https://github.com/sequelize/sequelize/issues/2494)
```
sushantdhiman: Change log should use links like [#2494](https://github.com/sequelize/sequelize/issues/2494)

lib/dialects/abstract/query.js
```diff
@@ -484,11 +484,6 @@ AbstractQuery.prototype.checkLoggingOption = function() {
     console.log('DEPRECATION WARNING: The logging-option should be either a function or false. Default: console.log');
     this.options.logging = console.log;
   }
-
-  if (this.options.logging === console.log) {
-    // using just console.log will break in node < 0.6
-    this.options.logging = function(s) { console.log(s); };
-  }
```
sushantdhiman: Why removing this line? I think it is preventing build break when using `Node < 0.6`.

mickhansen: We haven't supported 0.6 for a good while

lib/dialects/mysql/query.js
```diff
@@ -29,17 +29,17 @@ Query.prototype.run = function(sql, parameters) {
...
+  } else {
+    this.sequelize.log('Executing (' + (this.connection.uuid || 'default') + '): ' + this.sql, this.options);
   }
 
   var promise = new Utils.Promise(function(resolve, reject) {
     self.connection.query(self.sql, function(err, results) {
 
       if (benchmark) {
-        self.sequelize.log('Executed (' + (self.connection.uuid || 'default') + '): ' + self.sql + ' in ' + (Date.now() - queryBegin) + 'ms', self.options);
+        self.sequelize.log('Executed (' + (self.connection.uuid || 'default') + '): ' + self.sql, (Date.now() - queryBegin), self.options);
```
mickhansen: This changes the signature so technically a breaking change

lib/dialects/abstract/query.js
tornillo: We really need to support such versions in 2016?:)
I removed it because this "fake function" accepts only one argument ```s``` and not allow to log timings (for all node versions) in default sequelize configuration.

lib/dialects/mysql/query.js
tornillo: Why? All tests passed ok.

sushantdhiman: Means it will be release in next major version :)

tornillo:
I really do not like the idea of multiple functions for logging...
At now with ```benchmark=true``` all users see in console:
```SELECT "User"."id" FROM "users" AS "User" WHERE "User"."id" = 123; 1```
when `1` - timing

File: lib/dialects/mysql/query.js (hunk 3 above)
Comment (tornillo) [inline lib/dialects/mysql/query.js]: Ok, it's not a problem for me:)

File: changelog.md (hunk 1 above)
Comment (tornillo) [inline changelog.md]: Fixed. Thanks!

File: lib/dialects/mysql/query.js (hunk 3 above)
Comment (mickhansen) [inline lib/dialects/mysql/query.js]: Apparently we don't have tests for it, but if a user is expecting options to be in the second argument you just broke his code.

4.0 is months away, i'd suggest you rather rewrite this to have the benchmark be part of the options passed to the  log call.

Comment (tornillo) [inline lib/dialects/mysql/query.js]: Please look at Sequalize.log function: https://github.com/sequelize/sequelize/blob/master/lib/sequelize.js#L1282
If options argument is received, then it is cut out (by slice) from args that are passed to the logger function. So the user will never receive this options.

Comment (mickhansen) [inline lib/dialects/mysql/query.js]: Ah, right you are.

File: test/integration/sequelize.test.js [hunk 4]
```diff
@@ -249,6 +249,18 @@ describe(Support.getTestDialectTeaser('Sequelize'), function() {
       return this.sequelize.query(this.insertQuery);
     });
 
+    it('executes a query with benchmarking option', function() {
```
janmeier: Could you add a test that tests the global option as well? You can use `support.createSequelizeInstance` to get a new instance with benchmarking enabled

janmeier:
This LGTM, not BC breaks as far as I can see - just needs an extra test

janmeier:
Regarding multiple args - I definitely like the multiple args way better, since it makes the function more usable to do actual benchmarking, i.e. getter a number of seconds. I agree that seeing the number printed in the console might not be very intuitive, but we could fix that by changing the default function for logging https://github.com/sequelize/sequelize/blob/a157de4a5f15707d450f770d54f6d1a6037fcc1a/lib/sequelize.js#L144, to something like:

```js
logging: function (sql, time) {
  var log = sql;
  if (time) {
    log += ' Elapsed time: ' + time + 'ms';
  }
  console.log(sql);
}

tornillo:
Sorry for the long wait)

I added some new tests.
Unfortunately, we can not rewrite the default options.logging function (console.log), as it will break logging tests (https://github.com/sequelize/sequelize/blob/master/test/integration/sequelize/log.test.js).

In addition, in all dialects we also have default console.log in options...
Like this (https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/query.js#L16)

So I added check of options and arguments to sequelize.log for customization of the message, when logger is equal to console.log.
https://github.com/sequelize/sequelize/pull/5282/files#diff-24f8b5cb05b7e92726ca732f7d16d5baR1294

Waiting for new comments and suggestions)

janmeier:
LGTM, thanks for the thorough description of the changes :)

janmeier:
Needs a rebase though

tornillo:
Done)

tornillo:
Previous travis tests have failed because I used Number.isInteger() from ES6. My bad)
I removed this unnecessary checking from code.
Also fixed tests which use sinon.spy of console.log

janmeier:
Added in https://github.com/sequelize/sequelize/commit/f7fa33ce0522efcbd6a730ca54565aa321ef8770, once again thanks for your work @tornillo !

tornillo:
Thanks @janmeier! I'm glad to help)

miparnisari:
Hey all! Is this feature documented somewhere? I'm interested in knowing how could I log the queries that take more than X seconds.

sushantdhiman:
http://docs.sequelizejs.com/en/latest/api/model/#findalloptions-promisearrayinstance , look for `options.benchmark`

jontelm:
What about adding benchmark to parsing also? So we can get some information about the time being spent parsing the results when this seems to be the heavy part some times when not using raw: true. 

Executing: "Select * From y"; x ms
Parsing result: y ms
Totalt time for executing & parsing: x ms

Issues that this may help: #4868, #6110

sushantdhiman:
@jontelm We have a tool for general benchmarking https://github.com/sushantdhiman/sequelize-benchmark

janmeier:
We could also add debug support to the parsing, this should give an idea of the time spent as well

jontelm:
@janmeier sounds good.

I have been trying to find a place where the raw-result from sql is looped to generate instances. What files should I look into? I will create a PR when I have something.

janmeier:
@jontelm https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query.js

PR #5283, opened 2016-01-22
geography and tests

onzag:
Add the GEOGRAPHY datatype support from postgis based on http://postgis.net/docs/using_postgis_dbmanagement.html#PostGIS_Geography which is indeed very similar to the GEOMETRY datatype but takes the roundness of the earth into account.

This is very useful for setting geographical points with a high degree of exactitude.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@
 - [ADDED] Add `IF EXIST` to postgres alter enum [#4464](https://github.com/sequelize/sequelize/pull/4464)
 - [FIXED] Postgres destroy with `where` fails on JSONB data [#5092](https://github.com/sequelize/sequelize/issues/5092)
 - [FIXED] hasMany.separate with foreign keys having `field`
+- [ADDED] Geography support for postgres
 
 # 3.17.3
 - [FIXED] Regression with array values from security fix in 3.17.2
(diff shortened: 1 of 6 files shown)


changelog.md
```diff
@@ -5,6 +5,7 @@
 - [ADDED] Add `IF EXIST` to postgres alter enum [#4464](https://github.com/sequelize/sequelize/pull/4464)
 - [FIXED] Postgres destroy with `where` fails on JSONB data [#5092](https://github.com/sequelize/sequelize/issues/5092)
 - [FIXED] hasMany.separate with foreign keys having `field`
+- [ADDED] Geography support for postgres
```
mickhansen: Don't add a changelog entry to an already published version.
You can add a `# FUTURE` version and put it there.

test/integration/model/geography.test.js
```diff
@@ -0,0 +1,189 @@
...
+          return User.findOne({where: {username: props.username}});
+        }).then(function(user) {
+          expect(user.geography).to.be.deep.eql(point2);
+        });
+      });
+    });
+
+    describe('GEOGRAPHY(LINESTRING)', function() {
+      beforeEach(function() {
+        return Support.prepareTransactionTest(this.sequelize).bind(this).then(function(sequelize) {
```
janmeier: I think you should be able to remove the `prepareTransactionTest` part - I know the other geom tests use it, but I dunno why :)

janmeier:
I've fixed the changelog, and removed `prepareTransactionTest` from the test myself

onzag:
:D

Issue #4755, opened 2015-10-28
Upsert fails when primary key attribute and field names differ


BigDataSamuli:
I'm getting unique constraint violation errors when using upsert with models that have primary keys with custom DB field names. For example:

```javascript
var UpsertTestModel = sequelize.define(
    "UpsertTest",
    {
        testKey: {
            type: Sequelize.INTEGER,
            primaryKey: true,
            field: "test_key"
        },
        value: {
            type: Sequelize.INTEGER
        }
    },
    {
        tableName: "upsert_test"
    }
).sync({ force: true });

UpsertTestModel.then(model => {
    return model.upsert({
        testKey: 1,
        value: 1
    }).then(() => {
        return model.upsert({
            testKey: 2,
            value: 1
        });
    }).then(() => {
        return model.upsert({
            testKey: 2,
            value: 2
        });
    });
}).done();
```

The third upsert will fail with SequelizeUniqueConstraintError. Removing the custom field name definition from the model will make the code work.

With a quick look, it seems like QueryInterface.prototype.upsert is trying to create a where-clause with the primary key values, but it's looking for the primary key attribute names in the values argument, which contains the values by field name instead. Thus, no where clause is created and the SQL UPDATE (in Postgres) fails.

janmeier:
Fixed in https://github.com/sequelize/sequelize/pull/5298

PR #5298, opened 2016-01-25
Fix for bug #4755

jkennedy1980:
`QueryInterface.upsert` accepts values by field name (`.field`) but picked primary key values for the where clause using attribute name. I modified `upsert` so that `.field` is taken into account when collecting data for the upsert where clause. I've also added a failing test case which now passes.

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -495,20 +495,23 @@ QueryInterface.prototype.insert = function(instance, tableName, values, options)
   });
 };
 
-QueryInterface.prototype.upsert = function(tableName, values, updateValues, model, options) {
+QueryInterface.prototype.upsert = function(tableName, valuesByField, updateValues, model, options) {
   var wheres = []
     , where
     , indexFields
     , indexes = []
-    , attributes = Object.keys(values);
+    , attributes = Object.keys(valuesByField);
 
-  where = {};
-  for (var i = 0; i < model.primaryKeyAttributes.length; i++) {
-    var key = model.primaryKeyAttributes[i];
-    if(key in values){
-      where[key] = values[key];
+  where = model.primaryKeyAttributes.reduce(function (where, key) {
+    var attribute = model.rawAttributes[key];
+
+    if (attribute.field && attribute.field in valuesByField) {
+      where[attribute.field] = valuesByField[attribute.field];
+    } else if (key in valuesByField) {
+      where[key] = valuesByField[key];
     }
-  }
+    return where;
+  }, {} );
 
   if (!Utils._.isEmpty(where)) {
     wheres.push(where);
@@ -536,7 +539,7 @@ QueryInterface.prototype.upsert = function(tableName, values, updateValues, mode
     if (Utils._.intersection(attributes, index).length === index.length) {
       where = {};
       index.forEach(function (field) {
-        where[field] = values[field];
+        where[field] = valuesByField[field];
       });
       wheres.push(where);
     }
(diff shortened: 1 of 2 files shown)


lib/query-interface.js
```diff
@@ -495,18 +495,20 @@ QueryInterface.prototype.insert = function(instance, tableName, values, options)
   });
 };
 
-QueryInterface.prototype.upsert = function(tableName, values, updateValues, model, options) {
+QueryInterface.prototype.upsert = function(tableName, valuesByField, updateValues, model, options) {
```
mickhansen: Any reason to change the variable name?

lib/query-interface.js
```diff
@@ -495,18 +495,20 @@ QueryInterface.prototype.insert = function(instance, tableName, values, options)
...
   for (var i = 0; i < model.primaryKeyAttributes.length; i++) {
     var key = model.primaryKeyAttributes[i];
-    if(key in values){
-      where[key] = values[key];
+    var attribute = model.rawAttributes[key];
+    var fieldName = attribute.field || attribute.fieldName;
+    if(fieldName in valuesByField){
+      where[fieldName] = valuesByField[fieldName];
     }
   }
```
mickhansen: Consider retaining the original code and simply mapping where through https://github.com/sequelize/sequelize/blob/master/lib/utils.js#L121 before sending it to upsertQuery

lib/query-interface.js
jkennedy1980: Mainly for clarity. It seems to me that most of the time the values are passed around by attribute name and not by field name. My intent here was to be more informative.

mickhansen: Well the values here would usually be by attribute name untill converted yeah?

jkennedy1980: Well, I'm not sure what usually happens in your code base :) What I know is that when a PK attribute has a `.field` the passed in values object will be keyed on the `.field` name rather than the attribute name. The previous code seemed to assume that the values were keyed by attribute name.

mickhansen: Right but the end user would still pass the values to upsert via attribute names.
However we already do mapping here: https://github.com/sequelize/sequelize/blob/master/lib/model.js#L2032 so not quite sure why the fix is needed in the first place.

jkennedy1980: Correct. The array of values keyed by attribute name is transformed to an array of values keyed by field name before being passed to `QueryInterface`. This is why I renamed the `values` variable to `valuesByField`. The original code then took this, iterated the model attributes and tried to use the attribute name to get data from the `valuesByField` object. This resulted in a undefined which is the whole problem.

lib/query-interface.js
```diff
@@ -495,18 +495,21 @@ QueryInterface.prototype.insert = function(instance, tableName, values, options)
...
+  for (var i = 0; i < model.primaryKeyAttributes.length; i++){
     var key = model.primaryKeyAttributes[i];
-    if(key in values){
-      where[key] = values[key];
+    var attribute = model.rawAttributes[key];
+    var fieldName = attribute.field || attribute.fieldName;
+    if(fieldName in valuesByField){
+      where[fieldName] = valuesByField[fieldName];
     }
   }
```
mickhansen: Okay so this is nitpicking, but since you're in it, this could likely be rewritten to be much prettier as:

```js
where = model.primaryKeyAttributes.reduce(function (where, key) {
  var attribute = model.rawAttributes[key];

  if (attribute.field && attribute.field in valuesByField) {
    where[attribute.field] = valuesByField[attribute.field];
  } else if (key in valuesByField) {
    where[key] = valuesByField[key];
  }
  return where;
}, {});
```

jkennedy1980: Updated ... It's your code :) Reduce seems like a strange choice to me in this case.

lib/query-interface.js
janmeier: Makes sense - thanks for tracking the problem down. 

However, we already have the same code in instance, https://github.com/sequelize/sequelize/blob/68e613050d5e1e57a91c12fdf1b68961a19d70c1/lib/instance.js#L126 - I'll merge the PR and use that instead

janmeier:
Thanks a lot @jkennedy1980  - See the change [here](https://github.com/sequelize/sequelize/commit/f778e4026fa6b678a11c07ebc93ef2f664c6a5fb) - I also replace the use of `.delay` with fake timers (not related to your change, I just noticed there was no need for them while looking at the tests)

PR #5312, opened 2016-01-27
Fixed field alias in order clause

taylorhakes:
Fixes https://github.com/sequelize/sequelize/issues/4167. Should respect parent model as well as sub model fields. If I missed something, let me know

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -115,6 +115,37 @@ var Utils = module.exports = {
       options.where = Utils.mapWhereFieldNames(options.where, Model);
     }
 
+    if (Array.isArray(options.order)) {
+      options.order.forEach(function(oGroup) {
+        var OrderModel, attr, attrOffset;
+
+        if (Array.isArray(oGroup)) {
+          OrderModel = Model;
+
+          // Check if we have ['attr', 'DESC'] or [Model, 'attr', 'DESC']
+          if (typeof oGroup[oGroup.length - 2] === 'string') {
+            attrOffset = 2;
+
+          // Assume ['attr'], [Model, 'attr'] or [seq.fn('somefn', 1), 'DESC']
+          } else {
+            attrOffset = 1;
+          }
+
+          attr = oGroup[oGroup.length - attrOffset];
+          if (oGroup.length > attrOffset) {
+            OrderModel = oGroup[oGroup.length - (attrOffset + 1)];
+            if (OrderModel.model) {
+              OrderModel = OrderModel.model;
+            }
+          }
+
+          if (OrderModel.rawAttributes && OrderModel.rawAttributes[attr] && attr !== OrderModel.rawAttributes[attr].field) {
+            oGroup[oGroup.length - attrOffset] = OrderModel.rawAttributes[attr].field;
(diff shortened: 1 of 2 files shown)


mickhansen:
Ah sweet, someone finally decided to fix this.

lib/utils.js
```diff
@@ -115,6 +115,25 @@ var Utils = module.exports = {
       options.where = Utils.mapWhereFieldNames(options.where, Model);
     }
 
+    if (Array.isArray(options.order)) {
+      options.order.forEach(function(oGroup) {
+        var OrderModel, attr;
+        if (Array.isArray(oGroup)) {
+          OrderModel = Model;
+          attr = oGroup[oGroup.length - 2];
```
mickhansen: order can be defined without a direction where this will fail.
Please add fix + test for case `[Model, field]` as opposed to `[Model, field, DESC|ASC]`

taylorhakes: Ok thanks. I will add that case. I didn't see that example in the docs here http://docs.sequelizejs.com/en/latest/docs/querying/#ordering or here http://docs.sequelizejs.com/en/latest/api/model/#findall .  I now see that is possible in the code

mickhansen: Right there's no example but ASC is usually the default for RDBMS' so it's sort of implicit :)

taylorhakes: I imagine it can be this as well `[field]` ?

mickhansen: Yep an array of of arrays with a single string describing a field is supported too

taylorhakes:
I pushed out some new code with more tests. Do you think this is a breaking change? This type of code my break

```js
sequelize.define('User', {
  firstName: {
    type: DataTypes.STRING,
    field: 'lastName'
  },
  lastName: {
    type: DataTypes.STRING,
    field: 'firstName'
  }
})
```
Might be a minor concern, but possible someone has code like this.

mickhansen:
Why would that case break? I myself probably have code like that hidden somewhere so please do test for it :)

taylorhakes:
It won't throw an error, but the response will change if you have a query like this.
```js
User.findAll({
  order: [['firstName', 'ASC']]
})
```
Previously the code would ordered by `firstName`, now because of this change it will be ordered by `lastname`

mickhansen:
Oh lol, i didnt realize the field was reversed, yea that's fine, that would be consideren an error by the user.

mickhansen:
Ok LGTM, wil merge when green (ping me if i forget :))

taylorhakes:
I put in another test for switching fields, just to make sure it's understood that behavior is correct. I am fine if you don't consider this a breaking change (needing a major version bump). Just wanted to make clear that code like above will have the output change. The user could be writing perfectly legitimate code. You needed to write code differently before this fix.

mickhansen:
Not sure i would consider

```
firstName: {
  type: DataTypes.STRING,
  field: 'lastName'
}
```

as legimitate code, especially not if field matches another attribute, that's a recipe for disaster in any case.
So while the code might not break, it's definitely not intended or documented for it to work.

PR #5321, opened 2016-01-28
fix: describe query should not uppercase the enum type

luin:
Our team relies on the result of `describeTable` method to generate migrations automatically. However, we found that enum types are uppercased by sequellize by mistake.

For example, we have a field with the type of `enum('admin', 'member', 'guest')`, the result of `describeTable` will be `ENUM('ADMIN', 'MEMBER', 'GUEST')`, which is obviously not as same as the former one.

After the changes, the result will become `ENUM('admin', 'member', 'guest')`.

### DIFF
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -91,8 +91,9 @@ Query.prototype.formatResults = function(data) {
     result = {};
 
     data.forEach(function(_result) {
+      var enumRegex = /^enum/i;
       result[_result.Field] = {
-        type: _result.Type.toUpperCase(),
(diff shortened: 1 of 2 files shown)


janmeier:
LGTM, just needs a test

luin:
Yeah. Since I'm not familiar with the code base of sequelize, would you mind to write a test for this? Just did a quick search and it seems `formatResults` function is not test-covered yet.

janmeier:
You can add an integration test, or alter one of the existing ones to include an enum https://github.com/sequelize/sequelize/blob/master/test/integration/query-interface.test.js#L148

janmeier:
pING @luin

luin:
A little busy these days. I'll write tests for it soon.

luin:
added and rebased

janmeier:
Thanks! :)

Issue #5310, opened 2016-01-26
Fields modified in bulkCreate hook are not persisted to the DB if they aren't in updateOnDuplicate


ashokfernandez:
When using the `bulkCreate` method in MySQL it's possible to "bulk save" many records at once by using the `updateOnDuplicate` option. However if there is a `beforeBulkCreate` hook on the model and a field is modified inside the hook that isn't listed in the `updateOnDuplicate` array then this modified field is not persisted to the database.

For example if you a model with a hook

```
sequelize.define('MyObject',
    {
      id: DataTypes.INTEGER,
      fieldOne: DataTypes.INTEGER,
      fieldTwo: DataTypes.INTEGER
    }, 
   {
     hooks: {
        beforeBulkCreate: function (objects, options) {
           _.forEach(objects, function (object) {
              object.set('fieldTwo', 2)
          })
     }
  })
```

If I have created objects 
```
{ id: 1, fieldOne: 1, fieldTwo: null}, {id: 2, fieldOne:2, fieldTwo: null}
```

Then I would expect that this operation
```
MyObject.bulkCreate({ id: 1, fieldOne: 2}, {id: 2, fieldOne:3}, {
   updateOnDuplicate: ['fieldOne']
})
```

Should result in

```
{ id: 1, fieldOne: 2, fieldTwo: 2}, {id: 2, fieldOne:3, fieldTwo: 2}
```

But the changes to `fieldTwo` are not persisted to the database unless the options are changed to
```
MyObject.bulkCreate({ this: 2, that:1}, {this:5, that:4}, {
   updateOnDuplicate: ['fieldOne', 'fieldTwo']
})
```

Not sure if this is intentional or not but thought I would leave this here as a reference in case someone else comes across it. Perhaps there could be a note in the hooks documentation warning of this?

ashokfernandez:
Changing the hook to 

```
hooks: {
        beforeBulkCreate: function (objects, options) {
            if (!_.includes(options.updateOnDuplicate, 'fieldTwo')) {
               options.updateOnDuplicate.push('fieldTwo)
            }

           _.forEach(objects, function (object) {
              object.set('fieldTwo', 2)
          })
     }
```

Does remove the need to the user to specify the updateOnDuplicate field in the `bulkCreate` call. Happy to make a PR to the docs if you think it will be useful?

mickhansen:
In the case where a user has defined updateOnDuplicate it would probably not be prudent to magically edit that.

[...]

PR #5328, opened 2016-01-28
Added documentation note about bulkCreate hook and updatesOnDuplicate option

ashokfernandez:
Added notes to documentation about this caveat when using updateOnDuplicate with a hook.

https://github.com/sequelize/sequelize/issues/5310

### DIFF
--- a/docs/docs/hooks.md
+++ b/docs/docs/hooks.md
@@ -193,6 +193,27 @@ Model.beforeBulkDestroy(function(whereClause, fn) {
 Model.destroy({ where: {username: 'Tom'}} /*whereClause argument*/)
 ```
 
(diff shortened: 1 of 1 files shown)


docs/docs/hooks.md
```diff
@@ -193,6 +193,27 @@ Model.beforeBulkDestroy(function(whereClause, fn) {
 Model.destroy({ where: {username: 'Tom'}} /*whereClause argument*/)
 ```
 
+If you use `Model.bulkCreate(...)` with the `updatesOnDuplicate` option, changes made in the hook to fields that aren't given in the `updatesOnDuplicate` array will not be persisted to the database. However it is possible to change the updatesOnDuplicate option inside the hook if this is what you want.
+
+```
+// Bulk updating existing users with updatesOnDuplicate option
+Users.bulkCreate({ id: 1, isMemeber: true},
```
mickhansen: typo

ashokfernandez: Whoops, good spotting. Fixed in https://github.com/ashokfernandez/sequelize/commit/12acc40cac2c8ca8e4a73b514056aa632d0f1345.

Issue #5300, opened 2016-01-25
JSON/JSONB values are incorrectly escaped when they evaluate to false.


alekbarszczewski:
```js
var Sequelize = require('sequelize')

var sequelize = new Sequelize('postgres://localhost/test')

var TestModel = sequelize.define('TestModel', {
  data: Sequelize.JSON
})

sequelize.sync({ force: true })
.then(function () {
  return TestModel.create({
    data: true // or 1
  })
})
.then(function () {
  return TestModel.create({
    data: false // or 0
  })
})

/*
Executing (default): INSERT INTO "TestModels" ("id","data","updatedAt","createdAt") VALUES (DEFAULT,'true','2016-01-25 12:18:04.032 +00:00','2016-01-25 12:18:04.032 +00:00') RETURNING *;
Executing (default): INSERT INTO "TestModels" ("id","data","updatedAt","createdAt") VALUES (DEFAULT,false,'2016-01-25 12:18:04.041 +00:00','2016-01-25 12:18:04.041 +00:00') RETURNING *;
Unhandled rejection SequelizeDatabaseError: column "data" is of type json but expression is of type boolean
*/
```

```
true/1 => 'true'/'1'
false/0 => false/0
```

Tested on sequelize@3.18.0

PR #5331, opened 2016-01-29
Properly escape number 0 in query generator

zaggino:
This fixes issue #5300


mickhansen:
#5300 deals with `false`, can you add a test for that aswell?

zaggino:
@mickhansen it should be fine now

mickhansen:
Great, now if i could simply ask you to squash your commits into a single descriptive one ala `fix #5300: false and 0 are now properly encoded when using JSON/JSONB`

zaggino:
@mickhansen done

mickhansen:
Thanks @zaggino

PR #5333, opened 2016-01-29
Fixed findAll with order on custom field, limit, and hasMany relations

### LINKED ISSUE #5332: Column does not exist for findAll with limit, order on custom field, and hasMany associations
I get a `SequelizeDatabaseError: column ${x} does not exist` for a findAll query with with limit, order on a custom field, and hasMany associations. The problem is that when there is a subquery, the subquery aliases the custom field to the attribute name in the sequelize model, but the `ORDER BY` in the main query still refers to the custom field.

This test in the associated PR (#5333) illustrates the issue:
https://github.com/sequelize/sequelize/pull/5333/files#diff-916a03a19d49cecbcd62820c088b9587R1881

The test doesn't pass without the changes in the PR.

troyastorino:
Fixed #5332


troyastorino:
I looked at the failing tests, and am skeptical that this change would have broken those tests. Would a rebuild on the [mysql build](https://travis-ci.org/sequelize/sequelize/jobs/105596675) be worth trying? And I don't really know where to start looking on the [postgres-native build](https://travis-ci.org/sequelize/sequelize/jobs/105596666).

Any help would be appreciated, thanks!

mickhansen:
We have some intermittent failures, so likely not related to your changes (also green now).

mickhansen:
Do you mind adding a unit test for the sql generation aswell? Here: https://github.com/sequelize/sequelize/blob/master/test/unit/sql/select.test.js or create a new file for order specific tests

troyastorino:
OK, I added some query generation tests, but two of the tests I added are failing for MSSQL. I tried fiddling around with them to make them work, but I'm not familiar with MSSQL, so I don't know if the tests are actually showing a problem, or (much more likely) I'm just not sure how they should be working.

mickhansen:
Linting errors, all builds are failing so i can't look at what might be failing with MSSQL.

troyastorino:
Whoops, sorry! OK, fixed the linting, and now only mssql is failing

troyastorino:
Basically, I'm not sure whether this query (https://travis-ci.org/sequelize/sequelize/jobs/106788884#L1832):
```sql
SELECT [User].*, [Posts].[id] AS [Posts.id], [Posts].[le_title] AS [Posts.leTitle]
FROM (
  SELECT [User].[first_name] AS [firstName], [User].[id] AS [id] FROM [users] AS [User] 
  ORDER BY [User].[first_name] DESC, [id] OFFSET 0 ROWS FETCH NEXT 100 ROWS ONLY
) AS [User] 
LEFT OUTER JOIN [posts] AS [Posts]
ON [User].[id] = [Posts].[user_id] ORDER BY [User].[firstName] DESC;
```
is supposed to have `, [id]` inside of it

mickhansen:
Hmmm, mssql is abit odd, we might be added the primary key by default when doing limit.
https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/query-generator.js#L622

troyastorino:
Ya, I was looking at that, but that's what was getting me confused (the conditions around that line). If that's proper behavior, I can fiddle around with generating the expected query to make sure it matches.

troyastorino:
But, since the integration tests worked, I'd assume that it's the right behavior, right?

mickhansen:
Yeah i guess that's the expected behaviour :)

Issue #5344, opened 2016-02-01
sequelize@3.19.0 breaks findAndCount with eagerly loaded model using pg


yjwong:
## How to reproduce:

Install sequelize@3.19.0 and pg@4.4.4.

```javascript
'use strict';
let Sequelize = require('sequelize');
let sequelize = new Sequelize('database', 'username', 'password', {
	dialect: 'postgres',
	host: 'localhost',
	port: 5432
});

let Company = sequelize.define('Company', {
	id: {
		type: Sequelize.INTEGER,
		primaryKey: true,
		autoIncrement: true
	}
});

let Score = sequelize.define('Score', {
	id: {
		type: Sequelize.INTEGER,
		primaryKey: true,
		autoIncrement: true
	}
});

Company.hasOne(Score);
sequelize.sync()
	.then(() => Company.findAndCountAll({ include: [{ model: Score }] })
	.then(result => console.log(result));
```

## Output

```
module.js:328
    throw err;
    ^

Error: Cannot find module 'pg-native'
    at Function.Module._resolveFilename (module.js:326:15)
    at Function.Module._load (module.js:277:25)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/tmp/test/node_modules/pg/lib/native/index.js:1:76)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at null.<anonymous> (/tmp/test/node_modules/pg/lib/index.js:78:36)
    at /tmp/test/node_modules/lodash/lodash.js:4015:32
    at baseForOwn (/tmp/test/node_modules/lodash/lodash.js:2567:24)
    at baseClone (/tmp/test/node_modules/lodash/lodash.js:2277:39)
    at /tmp/test/node_modules/lodash/lodash.js:2278:34
    at /tmp/test/node_modules/lodash/lodash.js:4015:15
    at baseForOwn (/tmp/test/node_modules/lodash/lodash.js:2567:24)
    at baseClone (/tmp/test/node_modules/lodash/lodash.js:2277:39)
    at /tmp/test/node_modules/lodash/lodash.js:2278:34
    at /tmp/test/node_modules/lodash/lodash.js:4015:15
[...]

PR #5345, opened 2016-02-01
Fixes issue #5344: Replaced _.cloneDeep with _.cloneDeepWith.

yjwong:
This fixes issue #5344, where `findAndCount` with eagerly loaded model is broken when using Sequelize with pg. In lodash 4.x, `cloneDeep` does not accept a customizer function, but a new function `cloneDeepWith` does.


janmeier:
Woops, seems I missed that!

PR #5367, opened 2016-02-04
Set updatedAt and createdAt values before validation

rafis:
If model has `updatedAt` or `createdAt` fields without allowing nulls:

```javascript
        createdAt: {
            type: DataTypes.DATE,
            allowNull: false,
            field: 'created_at'
        },
        updatedAt: {
            type: DataTypes.DATE,
            allowNull: false,
            field: 'update_at'
        },
```

then I'm receiving this error:

> { [SequelizeValidationError: notNull Violation: createdAt cannot be null,
notNull Violation: updatedAt cannot be null]
  message: 'notNull Violation: createdAt cannot be null,\nnotNull Violation: updatedAt cannot be null',
  errors:
   [ { message: 'createdAt cannot be null',
       type: 'notNull Violation',
       path: 'createdAt',
       value: null },
     { message: 'updatedAt cannot be null',
       type: 'notNull Violation',
       path: 'updatedAt',
       value: null } ] }

My proposal is to set `updatedAt` and `createdAt` values before doing validation.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -525,9 +525,10 @@ Instance.prototype.save = function(options) {
     , updatedAtAttr = this.Model._timestampAttributes.updatedAt
     , createdAtAttr = this.Model._timestampAttributes.createdAt
     , hook = self.isNewRecord ? 'Create' : 'Update'
-    , wasNewRecord = this.isNewRecord;
+    , wasNewRecord = this.isNewRecord
+    , now = Utils.now(this.sequelize.options.dialect);
 
-  if (updatedAtAttr && options.fields.indexOf(updatedAtAttr) === -1) {
+  if (updatedAtAttr && options.fields.length >= 1 && options.fields.indexOf(updatedAtAttr) === -1) {
     options.fields.push(updatedAtAttr);
   }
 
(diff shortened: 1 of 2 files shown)


lib/instance.js
```diff
@@ -525,9 +525,10 @@ Instance.prototype.save = function(options) {
     , updatedAtAttr = this.Model._timestampAttributes.updatedAt
     , createdAtAttr = this.Model._timestampAttributes.createdAt
     , hook = self.isNewRecord ? 'Create' : 'Update'
-    , wasNewRecord = this.isNewRecord;
+    , wasNewRecord = this.isNewRecord
+    , now = Utils.now(this.sequelize.options.dialect);
 
-  if (updatedAtAttr && options.fields.indexOf(updatedAtAttr) === -1) {
+  if (updatedAtAttr && options.fields.length && options.fields.indexOf(updatedAtAttr) === -1) {
```
mickhansen: Not necessary since options.fields will always be defined by then

rafis: I added this check not to check if `options.fields` are defined, but to check if we update at least one field, because if we update 0 fields then updatedAt should not be changed.

mickhansen:
LGTM i think, would like @janmeier's input.

janmeier:
Yep, looks good to me too

janmeier:
Thanks :)

PR #5368, opened 2016-02-04
NPM Script Build

### LINKED ISSUE #2727: makefile not portable - consider replacement
Hi,

trying to contribute I stumbled upon major portability issues with the makefile on Windows. Any arguments about why a native node build system shouldn't be used instead? Like grunt or gulp.

sushantdhiman:
Closes #2727

Introduce `gulp` based build system for `sequelize`. All `makefile` commands are converted to `gulp task` which provide better support for `windows` based system and easy to work / edit.


sushantdhiman:
cc @mickhansen , @janmeier

mickhansen:
Thanks for effort but we're not looking to integrate a build system. Ideally i'd love to move everything from make to just npm scripts

sushantdhiman:
With build system we have better control and management of test task. Currently we have 52 tasks in our build. Managing those tasks or changing even single command parameter will require lots of effort. Suppose having a `scripts` folder with 50 files.

Using build system is more intuitive and most of developer have exposure with them. Any specific reason to prefer scripts?

mickhansen:
gulp adds unnecessary overhead in my opinion, there's no reason to have a streaming task manager for what little we do, npm run scripts are way easier to reason about for everyone, requires no prior knowledge of gulp, just of shell basics

sushantdhiman:
Hmm, I think thats right. We don't actually use any streaming or heavy use of gulp.

Under npm scripts, we still want to support all command with dialect name like `test-unit-mysql,  test-unit-mariadb`. I think we can just use `DIALECT=mysql test-unit`. This will require far less number of scripts than we currently will have to.

sushantdhiman:
> Under npm scripts, we still want to support all command with dialect name like test-unit-mysql, test-unit-mariadb. I think we can just use DIALECT=mysql npm test-unit.

suggestions @mickhansen

mickhansen:
I use `npm run test-unit-postgres` quite a bit when testing.

sushantdhiman:
@mickhansen Changed the build to use pure `npm script`

mickhansen:
Sweet, will review it a bit later @sushantdhiman.
For general discussion of ideas you should definitely join us at: http://sequelize-slack.herokuapp.com/

package.json
```diff
@@ -87,11 +87,49 @@
...
+    "codeclimate": "npm run cover-all && npm run codeclimate-send",
+    "mysql": "DIALECT=mysql npm test",
+    "sqlite": "DIALECT=sqlite npm test",
+    "postgres": "DIALECT=postgres npm test",
+    "pgsql": "npm run postgres",
+    "postgres-native": "DIALECT=postgres-native npm test",
+    "postgresn": "npm run postgres-native",
+    "mariadb": "DIALECT=mariadb npm test",
+    "mssql": "DIALECT=mssql npm test",
+    "all": "npm run mysql && npm run sqlite && npm run postgres && npm run postgres-native && npm run mariadb && npm run mssql"
```
mickhansen: imo we should kill these in favor of more descriptive `test-[dialect]` `test-unit-[dialect]` `test-integration[dialect]` `test-all`

package.json
```diff
@@ -87,11 +87,49 @@
...
+    "test-integration-sqlite":"DIALECT=sqlite npm run test-integration",
+    "test-integration-mssql":"DIALECT=mssql npm run test-integration",
+    "test-integration-all":"npm run test-integration-mysql && npm run test-integration-postgres && npm run test-integration-postgres-native && npm run test-integration-mssql && npm run test-integration-sqlite && npm run test-integration-mariadb",
+    "cover": "rm -rf coverage && npm run teaser && COVERAGE=true ./node_modules/.bin/istanbul cover ./node_modules/.bin/_mocha --report lcovonly -- -t 30000 --ui tdd 'test/integration/**/*.test.js'",
+    "mysql-cover": "DIALECT=mysql npm run cover && mv coverage coverage-mysql",
+    "sqlite-cover": "DIALECT=sqlite npm run cover && mv coverage coverage-sqlite",
+    "postgres-cover": "DIALECT=postgres npm run cover && mv coverage coverage-postgres",
+    "postgres-native-cover": "DIALECT=postgres-native npm run cover && mv coverage coverage-postgres-native",
+    "mariadb-cover": "DIALECT=mariadb npm run cover && mv coverage coverage-mariadb",
+    "mssql-cover": "DIALECT=mssql npm run cover && mv coverage coverage-mssql",
```
mickhansen: lets invert these to `cover-mysql` so they have the same syntax as the test scripts

sushantdhiman:
@mickhansen Its ready for merge. I have fixed code coverage [issue ](https://github.com/codeclimate/javascript-test-reporter/issues/22 )that was bothering our builds.

sushantdhiman:
> For general discussion of ideas you should definitely join us at: http://sequelize-slack.herokuapp.com/

I would love to but due to some issue I never get slack invite email (may be outlook issue). I will contact slack support about it, will join as soon its fixed.

mickhansen:
Looks good, i need to test it out locally before merging though.
/cc @janmeier

mickhansen:
@sushantdhiman You can email me at maker@mhansen.io and i can try to invite you directly :)

sushantdhiman:
@mickhansen @janmeier Did you get chance to test it out locally.

janmeier:
Reviewed and tested locally - And everything looks good.

Thanks a lot @sushantdhiman !

sdepold:
> gulp adds unnecessary overhead in my opinion, there's no reason to have a streaming task manager for what little we do, npm run scripts are way easier to reason about for everyone, requires no prior knowledge of gulp, just of shell basics

I can hardly agree more :)

PR #5371, opened 2016-02-04
Fix broken links on documentation

theholla:


mickhansen:
Not sure this is correct since we're using RTD

janmeier:
The articles are not shown on RTD - the contents of getting started have been moved to https://github.com/sequelize/sequelize/blob/master/docs/docs/getting-started.md

PR #5381, opened 2016-02-08
Fix bug in SqlString.escape which caused the function to be recursive…

### LINKED ISSUE #4912: JSON type not escaping double quotes properly
Seems to be the same problem as #305, where `\"` gets changed to `\\"` and postgres throws `error: invalid input syntax for type json`. 

The breakage seems to happen when upgrading from 3.13 to 3.14

MichaelBuckley:
…ly called with the wrong parameters when passed an array of strings. This would cause the incorrect escape algorithm to be used on the strings within the array when the dialect was set to postgres, sqlite, or mssql.

Greetings,

We ran into this bug today when attempting to upgrade from 3.0.1 to 3.19.2, hoping to pick up the security fix introduced in 3.17.0.

We believe this probably fixes #4912, but the issue lacks a reproducible test case, so we can't be sure.

The crux of the issue is that a format parameter was added to SqlString.escape in commit 12057dc which added a format parameter to the function, but this function calls itself recursively when passed an array, and this recursive call was not updated with the new parameter. This resulted in the dialect parameter containing the original array passed to the function, rather than the dialect. The dialect parameter was then used to determine if the database being used was postgres, sqlite, or mssql, and if so, to use a different algorithm to escape the strings. Because the parameter was an array, this check was always returning false, even when targeting one of these three database dialects.

The end result was that strings within these arrays were getting escaped improperly for their dialect. We found the bug because it was escaping double quotes within our strings, which were used to store JSON, so when we read them back out, we got invalid JSON.

We've also made the recursive call a bit more explicit and added a test to help ensure this doesn't happen again. Thank you.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 - [ADDED] `validationFailed` hook [#1626](https://github.com/sequelize/sequelize/issues/1626)
 - [FIXED] Mark index as `unique: true` when `type: 'UNIQUE'`. Fixes [#5351](https://github.com/sequelize/sequelize/issues/5351)
 - [ADDED[ Support for IEEE floating point literals in postgres and sqlite [#5194](https://github.com/sequelize/sequelize/issues/5194)
+- [FIXED] Improper escaping of bound arrays of strings on Postgres, SQLite, and Microsoft SQL Server
 
 # 3.19.3
 - [FIXED] `updatedAt` and `createdAt` values are now set before validation [#5367](https://github.com/sequelize/sequelize/pull/5367)
--- a/lib/sql-string.js
+++ b/lib/sql-string.js
@@ -48,7 +48,7 @@ SqlString.escape = function(val, timeZone, dialect, format) {
   }
 
   if (Array.isArray(val)) {
-    var escape = _.partialRight(SqlString.escape, timeZone, dialect);
+    var escape = _.partial(SqlString.escape, _, timeZone, dialect, format);
     if (dialect === 'postgres' && !format) {
       return dataTypes.ARRAY.prototype.stringify(val, {escape: escape});
     }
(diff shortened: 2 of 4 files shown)


changelog.md
```diff
@@ -1,3 +1,6 @@
+# 3.19.3
+- [FIXED] Improper escaping of bound arrays of strings on Postgres, SQLite, and Microsoft SQL Server
+
```
mickhansen: Please just add `Future`, we add versions when we cut :)

mickhansen:
Looks like this broke mssql unicode support

changelog.md
MichaelBuckley: Thanks for letting me know. I will fix that shortly.

MichaelBuckley:
I'm not very familiar with mssql, but if I understand correctly, it looks like this change actually fixed unicode support, it's just that the failing test isn't valid for mssql.

SqlString.escape states, "In mssql, prepend N to all quoted vals which are originally a string (for unicode compatibility)", and from Googling around, that seems to be correct.

This test passes the following array: [1, 'test', 3, "derp"], and expects the following string: "IN (1,'test',3,'derp')"

For Because we want to preserve unicode support, it seems to me that, for mssql, this test should expect: "IN (1,N'test',3,N'derp')".

This test used to pass before this fix because strings inside arrays were not getting properly escaped for postgres, sqlite, and mssql, but this test should never have passed for mssql.

If this sounds correct, I will amend that test to expect the proper string for mssql.

mickhansen:
@MichaelBuckley ah yeah, looks like you're right about that actually :)

MichaelBuckley:
I don't want to come off pushy at all, but I pushed a commit a couple weeks ago to address the test failures discussed in this pull request, and I wanted to comment just in case that push didn't send a notification that this request is ready to be pulled. Thanks.

mickhansen:
Commits don't emit notifications so please do ping us @MichaelBuckley 
Looks like we need a rebase to be able to merge though.

MichaelBuckley:
Thank you. I have rebased, and all the tests have passed.

PR #5382, opened 2016-02-09
isAlpha and isAlphanumeric Validator fix

### LINKED ISSUE #5377: Validation error with IsAlphanumeric & validator
Hi, 

I've been using Sequelize for a while, and without changing my model there is a new validation error, we updated the package, but can't remember which version it was earlier.

I'm currently using Sequelize 3.19.2 (tested with 3.18 too) with PostgreSQL 9.4.5. 

There is an error when using a `IsAlphanumeric` validation for a field with `validator.js`.

Simplified model which should trigger the error. : 

```javascript 
sequelize.define('MyModel', {
		myField: {
			type: DataTypes.STRING,
			allowNull: true,
			validate: {
				isAlphanumeric: true,
			}
		}
};
```

The actual error : 
```
SequelizeValidationError: Validation error: Cannot read property 'test' of undefined
    at /.../node_modules/sequelize/lib/instance-validator.js:74:14
```

Investigating the error, it seems to trigger because when Sequelize is validating (on save), it will call the `validator.isAlphanumeric` with `true` as the second parameter, which is the locale.

```javascript
// .../validator/validator.js
 validator.isAlphanumeric = function (str, locale) {
        locale = locale || 'en-US';
        return alphanumeric[locale].test(str);
    };
```

Since `alphanumeric[true]` does not exist it calls `undefined.test(str)`.

The problem may be linked to the `validator` package update (0bdbec8).

Ask up if you need more info.

Thanks.

cbauerman:
Added extra conditions so that extra arguments will not be passed to 'isAlpha' and 'isAlphanumeric.' Fixes #5377

### DIFF
--- a/lib/instance-validator.js
+++ b/lib/instance-validator.js
@@ -261,7 +261,7 @@ InstanceValidator.prototype._invokeBuiltinValidator = Promise.method(function(va
   if (!Array.isArray(validatorArgs)) {
     if (validatorType === 'isImmutable') {
       validatorArgs = [validatorArgs, field];
-    } else if (validatorType === 'isIP') {
+    } else if (validatorType === 'isIP' || validatorType === 'isAlpha' || validatorType === 'isAlphanumeric') {
       validatorArgs = [];
     } else {
       validatorArgs = [validatorArgs];


sushantdhiman:
@cbauerman May be squash your commits. They all are same.

PR #5392, opened 2016-02-10
Fixes #5390 - dialect mariaDB missing data-types

### LINKED ISSUE #5390: dialect mariaDB throw errors when submitting raw queries
on sequelize `3.19.2` I get the following error when doing a raw query with rpelacements-
```
TypeError: Cannot read property 'DATE' of undefined
    at Object.SqlString.escape (sequelize/lib/sql-string.js:39:29)
    at sequelize/lib/sql-string.js:86:22
    at String.replace (native)
    at Object.SqlString.format (sequelize/lib/sql-string.js:81:14)
    at Object.module.exports.format (sequelize/lib/utils.js:60:22)
    at Sequelize.query (sequelize/lib/sequelize.js:760:19)
    at Object.<anonymous> (test.js:6:11)
```

Code to reproduce -
```javascript
var Sequelize = require('sequelize'),
	sequelize = new Sequelize('db', 'user', 'password', {dialect: 'mariadb'});

sequelize.query('select * from test where date=?', {replacements: [new Date()]});
```
Notice that if you change the dialect to `mysql` this will work fine 
I think this is caused by `sequelize/lib/data-types.js` not exporting a dataType for maria db
I think the fix might be simple so I'll try to fix it and send a pull request

yonjah:
#5390 dialect mariaDB throw errors when submitting raw queries

### DIFF
--- a/test/integration/sequelize.test.js
+++ b/test/integration/sequelize.test.js
@@ -407,6 +407,9 @@ describe(Support.getTestDialectTeaser('Sequelize'), function() {
           var res = result[0] || {};
           res.date = res.date && new Date(res.date);
           res.boolean = res.boolean && true;
+          if (typeof res.buffer === 'string' && res.buffer.indexOf('\\x') === 0) {
+            res.buffer = new Buffer(res.buffer.substring(2), 'hex');
+          }
           expect(res).to.deep.equal({
             number : number,
             date   : date,
--- a/test/integration/sequelize.test.js
+++ b/test/integration/sequelize.test.js
@@ -386,39 +386,37 @@ describe(Support.getTestDialectTeaser('Sequelize'), function() {
       }).to.throw(Error, 'Both `replacements` and `bind` cannot be set at the same time');
     });
 
-    if (Support.dialectIsMySQL()) {
-      it.only('properly adds and escapes replacement value', function () {
-        var logSql,
-            number  = 1,
-            date = new Date(),
-            string = 't\'e"st',
-            boolean = true,
-            buffer = new Buffer('t\'e"st');
+    it('properly adds and escapes replacement value', function () {
+      var logSql,
+          number  = 1,
+          date = new Date(),
+          string = 't\'e"st',
+          boolean = true,
+          buffer = new Buffer('t\'e"st');
 
-        date.setMilliseconds(0);
-        return this.sequelize.query({
-            query: 'select ? as number, ? as date,? as string,? as boolean,? as buffer',
-            values: [number, date, string, boolean, buffer]
-          }, {
-            type: this.sequelize.QueryTypes.SELECT,
-            logging: function(s) {
-              logSql = s;
-            }
-          }).then(function(result) {
-            var res = result[0] || {};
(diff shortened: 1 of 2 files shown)


test/integration/sequelize.test.js
```diff
@@ -386,6 +386,40 @@ describe(Support.getTestDialectTeaser('Sequelize'), function() {
       }).to.throw(Error, 'Both `replacements` and `bind` cannot be set at the same time');
     });
 
+    if (Support.dialectIsMySQL()) {
+      it.only('properly adds and escapes replacement value', function () {
```
mickhansen: whoops :)

test/integration/sequelize.test.js
```diff
@@ -386,6 +386,40 @@ describe(Support.getTestDialectTeaser('Sequelize'), function() {
...
+              number : number,
+              date   : date,
+              string : string,
+              boolean: boolean,
+              buffer : buffer
+            });
+            expect(logSql.indexOf('?')).to.equal(-1);
+        });
+      });
+    }
```
mickhansen: Any reason this should be a mysql specific test? Sounds like a good test to have for all dialects.

yonjah: I wasn't sure if the format being returned will be identical on all dialects
and since I don' t have any other db install on my machine and  so couldn't run this test with other dialects and I didn't want to add test that might fail with other dialects.

test/integration/sequelize.test.js
yonjah: oh sorry about that :-)

test/integration/sequelize.test.js
mickhansen: Try giving it a go and see what travis says

yonjah: There seem to be an issue with Buffers in Postgres where the return value
is the binsry representation of the string and not a Buffer.
```diff
 + expected - actual
       {
         "boolean": true
      -  "buffer": "\\x742765227374"
      +  "buffer": [Buffer: [
      +    116
      +    39
      +    101
      +    34
      +    115
      +    116
      +  ]]
         "date": [Date: 2016-02-10T08:29:33.000Z]
         "number": 1
         "string": "t'e\"st"
       }
```
I'm not sure if this is actual bug that need to be fixed and if this is caused because the replacement doesn't format buffer correctly or that the parsing of the result is the problem.
I can try and setup a local dev environment with postgres but it might take me some time

mickhansen:
Ah so close, looks like there's actually a bug with binary parsing in postgres raw queries.

yonjah:
I checked it farther and I'm not sure if this is an actual bug.
The thing is that Postgres returns OID 705 (unknown) for the buffer data.
I think that if you fetch data from a real table Postgres will know its type and will return a proper OID so this is only an issue if your not selecting data from a real table (which I can't see the reason to do in a real situation ).

I changed the test so it will manually create the buffer if needed

mickhansen:
Thank you @yonjah - A fix and increased test coverage to boot :)

PR #5410, opened 2016-02-12
Update instance.md

PavelPolyakov:
Model is not a method anymore


janmeier:
The API docs are auto-generated. The source and / or the generation code needs to be updated instead

PavelPolyakov:
Thanks, a question then...

Why this:
image

Was autogenerated to this:
image

The docs are autogenerated according to the latest github content?

janmeier:
Probably a bug in the code that generates the docs :-). I believe there is already an issue about that

PR #5412, opened 2016-02-12
Corrected Typos

Codertaker:
Hi, I have corrected some minor typos in some of the pages such as : "Assocation" to "Association" and "Instace" to "Instance".

### DIFF
--- a/docs/api/associations/belongs-to-many.md
+++ b/docs/api/associations/belongs-to-many.md
@@ -25,7 +25,7 @@ Project.create({ id: 11 }).then(function (project) {
 });
 ```
 
-In the API reference below, replace `Assocation(s)` with the actual name of your association, e.g. for `User.belongsToMany(Project)` the getter will be `user.getProjects()`.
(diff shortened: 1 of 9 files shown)


janmeier:
The API docs are auto-generated - please update the code as well

Codertaker:
@janmeier This is my first pull request and I am a student. Would you be able to clarify your last statement? Thanks!

janmeier:
The docs are generated automatically from code comments - which means your changes will be overwritten next time we generate the docs, unless you change the comments as well

Codertaker:
I have made the changes to the comments. Hope it helps.

janmeier:
You forgot the comments for has one and the function parameters though ;-)

Codertaker:
I'm not sure if i made the right changes but here they are! Thank you for your patience.

janmeier:
Turns out the code was already updated - only the actual docs needed to be updated. Thanks for pointing it out :)

PR #5413, opened 2016-02-12
Replaced function removed from lodash 4

cbauerme:
`where` was removed from Lodash 4 in favor of `filter`. Running `npm run docs` errors out with "where is not a function" without this fix.


janmeier:
Thanks :-)

PR #5416, opened 2016-02-13
(issue #3295) self reference tests for BelongsToMany

### LINKED ISSUE #3295: Replace 2 x hasMany with belongsToMany fails
It seems `belongsToMany` fails for a M:N relationship between one model and itself. For example, one user follows another one.

```javascript
var sequelize = new Sequelize('testa', 'testa', 'testa', { dialect: 'mysql' });

  // some properties
});

var Follow = sequelize.define('Follow', {
  // some properties
}, {
  indexes: [
    {
      unique: true,
      fields: ['UserId', 'FanId']
    }
  ]
});

User.hasMany(User, { through: Follow, as: 'User' });
User.hasMany(User, { through: Follow, as: 'Fan' });

sequelize.sync({ force: true }).complete(function(err) { console.log(err); });
```

This works but says `Using 2 x hasMany to represent N:M relations has been deprecated. Please use belongsToMany instead`. So I replaced `hasMany` with `belongsToMany`.

```javascript
User.belongsToMany(User, { through: Follow, as: 'User' });
User.belongsToMany(User, { through: Follow, as: 'Fan' });
```

This fails with the message `ER_KEY_COLUMN_DOES_NOT_EXITS: Key column 'FanId' doesn't exist in table`.

Is this a BUG? Or did I miss something?

sushantdhiman:
Added tests which confirm self reference works for same model.

Closes #3295

### DIFF
--- a/test/integration/associations/belongs-to-many.test.js
+++ b/test/integration/associations/belongs-to-many.test.js
@@ -2059,7 +2059,75 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
     });
   });
 
-  describe('selfAssociations', function () {
+  describe('selfAssociations', function() {
+    it('should work with self reference', function() {
+      var User = this.sequelize.define('User', {
+        name : Sequelize.STRING(100)
+      })
+      , Follow = this.sequelize.define('Follow')
+      , self = this;
+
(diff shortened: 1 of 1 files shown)


mickhansen:
Thanks @sushantdhiman

PR #5428, opened 2016-02-15
Log which query tried to run on a finished transaction

jharting:
The "commit has been called on this transaction" error message usually indicates a problem with chaining of query promises within a transaction block. In more complex transaction blocks it may not be immediately clear which query is improperly chained and the current error message does not help much. This change logs the query (SQL) that the application attempted to run on a finished transaction. This makes it easy to identify a problem in an application.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -789,7 +789,9 @@ Sequelize.prototype.query = function(sql, options) {
   }
 
   if (options.transaction && options.transaction.finished) {
-    return Promise.reject(new Error(options.transaction.finished+' has been called on this transaction('+options.transaction.id+'), you can no longer use it'));
+    var error = new Error(options.transaction.finished+' has been called on this transaction('+options.transaction.id+'), you can no longer use it. (The rejected query is attached as the \'sql\' property of this error)');
+    error.sql = sql;
+    return Promise.reject(error);
   }
 
   if (this.test.$trackRunningQueries) {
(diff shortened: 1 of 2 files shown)


janmeier:
Lgtm - can you add a test as well?

jharting:
Will do

mickhansen:
This potentially makes the log message really long. Perhaps we could attach the SQL to the error instead?

jharting:
I modified existing tests to expect the query in the error message. I agree that the message could potentially get long. I am personally fine with that though as you can see the problem right in the message instead of needing to tweak the error handler. 

However, if the general preference is towards short error messages I can change this PR to attach SQL to error instead. Alternatively, what about adding only first X characters of SQL to the message and attaching the entire SQL to the error?

jharting:
Any preference on "log the entire sql" vs "log a substring of sql and attache the entire query within the error"?

Also, this PR most likely failed for an intermittent test failure. Can anyone re-run CI?

mickhansen:
Rerunning intermittent error.
Hmm, i'm slightly torn, on the one hand logging the entire SQL is both convenient when trawling through logs to debug an issue (you don't have to go in and debug err.sql anywhere) - on the other hand it's also possible a security issue.

jharting:
That's a good point. Let's not use the full query in the log line then. Unless anyone objects I am going to change the code so that the sql is attached and alter the message to inform users about the fact:

> commit has been called on this transaction(123456), you can no longer use it. (See error.sql for the full query that was rejected).

mickhansen:
Sounds good

jharting:
I updated the PR

Issue #4920, opened 2015-11-24
Unique validations with custom error message being overwritten


corbanb:
Hey guys! Long time Sequelize user, love the product and hope to contribute one day. 

In the meantime I have been trying to understand a bug I have run into. Basically the scenario is you have a User model that needs unique usernames and emails being saved to the db. Obviously these should be case insensitive so 'Foo' and 'foo' are the same user and fail if a new user tries to create 'foO' or any of the other possibilities in the future a unique error is thrown. 

## Info
Sequelize Version: 3.14.1
Dialect: postgres
Github Repo: https://github.com/corbanb/sequelize-demos
Heroku App: https://sequelize-demos.herokuapp.com

## Questions

Right now the demo does know it failed on uniques but in the case of the username it sends back the emails custom error message ([see here](https://sequelize-demos.herokuapp.com/user?email=foo@bar.com&username=foo)). 

1. Why am I getting the email custom error message for username? (see below)
2. Is this the proper usage of the fields lower case for unique or is there a more 'sequelize' way of doing this? (See User Model below)

### Error response

```javascript
{
  name: "SequelizeUniqueConstraintError",
  message: "Oops. Looks like you already have an account with this email address. Please try to login.",
  errors: [
    {
      message: "Oops. Looks like you already have an account with this email address. Please try to login.",
      type: "unique violation",
      path: "username",
      value: "foo"
  },
  {
    message: "Oops. Looks like you already have an account with this email address. Please try to login.",
    type: "unique violation",
    path: "email",
    value: "foo@bar.com"
  }
  ],
  fields: {
    username: "foo",
    email: "foo@bar.com"
  }
}
```

### Example User Model

```javascript
var Instance = require('sequelize/lib/instance');

module.exports = function(sequelize, DataTypes) {
    var User = sequelize.define('User', {
        username: {
            type: DataTypes.STRING,
            unique: {
                args: true,
                message: 'Username must be unique.',
                fields: [sequelize.fn('lower', sequelize.col('username'))]
            },
            validate: {
                min: {
                    args: 3,
                    msg: 'Username must start with a letter, have no spaces, and be at least 3 characters.'
                },
                max: {
                    args: 40,
                    msg: 'Username must start with a letter, have no spaces, and be at less than 40 characters.'
                },
                is: {
                    args: /^[A-Za-z][A-Za-z0-9-]+$/i, // must start with letter and only have letters, numbers, dashes
[...]

PR #5444, opened 2016-02-17
Multiple custom unique validation messages being overwritten by the first-defined message

connor4312:
This issue was not in the dialect as https://github.com/sequelize/sequelize/issues/4920 hints, but rather in the model definition. Sequelized looks/ed for unique index names when the `unique` property is given as an object, even if it only contains a `msg`. This resulted in, when many columns contained an object like `unique: { msg: '...' }`, it grouping them under the `undefined` key and subsequently ignoring further message definitions.

This PR modifies the building so that the name is pulled out only if defined in the unique property.

### DIFF
--- a/test/integration/instance.validations.test.js
+++ b/test/integration/instance.validations.test.js
@@ -131,6 +131,45 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), function() {
           expect(err.errors[0].message).to.equal('custom unique error message');
         });
     });
+
+    it('should handle multiple unique messages correctly', function() {
+      var Model = this.sequelize.define('model', {
+        uniqueName1: {
+          type: Sequelize.STRING,
+          unique: { msg: 'custom unique error message 1' }
+        },
+        uniqueName2: {
+          type: Sequelize.STRING,
+          unique: { msg: 'custom unique error message 2' }
+        },
+      });
+      var records = [
+        { uniqueName1: 'unique name one', uniqueName2: 'unique name one' },
+        { uniqueName1: 'unique name one', uniqueName2: 'this is ok' },
+        { uniqueName1: 'this is ok', uniqueName2: 'unique name one' },
+      ];
+      return Model.sync({ force: true })
+        .then(function() {
+          return Model.create(records[0]);
+        }).then(function(instance) {
+          expect(instance).to.be.ok;
+          return expect(Model.create(records[1])).to.be.rejected;
+        }).then(function(err) {
+          expect(err).to.be.an.instanceOf(Error);
+          console.log(err.errors);
+          expect(err.errors).to.have.length(1);
+          expect(err.errors[0].path).to.include('uniqueName1');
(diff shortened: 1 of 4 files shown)


connor4312:
One check failed on the MariaDB build, but it does not appear related to the changes in this PR.

mickhansen:
Thanks for your contribution @connor4312!

test/integration/instance.validations.test.js
```diff
@@ -131,6 +131,45 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), function() {
...
+      ];
+      return Model.sync({ force: true })
+        .then(function() {
+          return Model.create(records[0]);
+        }).then(function(instance) {
+          expect(instance).to.be.ok;
+          return expect(Model.create(records[1])).to.be.rejected;
+        }).then(function(err) {
+          expect(err).to.be.an.instanceOf(Error);
+          console.log(err.errors);
```
janmeier: Woops :)

mickhansen: heh damn, i missed that, will remove on master

connor4312: derp :stuck_out_tongue:

connor4312:
No problem, thanks for the speedy merge!

matthewheck:
Unfortunately, looks like this causes a breaking change!

I was defining models as such:

```
		user_id: {
			type: DataTypes.UUID,
			allowNull: false,
			queryable: true,
			unique: {
				args: 'userAndProduct',
				msg: 'There is already a user and product matching this combination.'
			}
		},
		product_id: {
			type: DataTypes.UUID,
			allowNull: false,
			queryable: true,
			unique: {
				args: 'userAndProduct',
				msg: 'There is already a user and product matching this combination.'
			}
		},
```

Our CI passed without problems yesterday, but today started throwing a 'SequelizeUniqueConstraintError' with my message. After realizing that the only thing that had changed was running on v3.19.2 yesterday and v3.19.3 today, I found this PR. I changed my model definition to this and all works fine:
```
		user_id: {
			type: DataTypes.UUID,
			allowNull: false,
			queryable: true,
			unique: {
				name: 'userAndProduct',
				msg: 'There is already a user and product matching this combination.'
			}
		},
		product_id: {
			type: DataTypes.UUID,
			allowNull: false,
			queryable: true,
			unique: {
				name: 'userAndProduct',
				msg: 'There is already a user and product matching this combination.'
			}
		},
```

Long story short, 'args' no longer works for me and 'name' does.

This is not documented anywhere and I don't see a particular reason in the code that 'args' would work before, so maybe using 'args' before was actually a false positive. I originally followed the 'args' logic thinking that this was, essentially, like an attribute validator.

Easy change, but wanted to flag this in case it was more than just a false positive on the previous 'args'.

mickhansen:
`args` was a definitely a false positive, you want `name` in this case.

matthewheck:
Great-- thanks for the response @mickhansen!

PR #5457, opened 2016-02-19
Restore old behaviour when using SYSTEM timezone in MySQL

connor4312:
As of 3.19.0, due to the changes in https://github.com/sequelize/sequelize/pull/5442, when using the SYSTEM timezone in MySQL all dates coming out of the database were set to `null`. This caused some considerable issues on our end!

This PR restores the previous behaviour of returning the date as a string when the SYSTEM timezone is set. Though it's not ideal -- all dates should preferably be parsed to Date objects -- it is behaviour that we've come to rely on in various parts of our platform (the format of the string is different than the format when JSON encoding Dates). It'd be great to have everything be Dates, but this will be a breaking change so should be held off until 4.0.0 :wink:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+# NEXT
+
+ - [FIXED] dates being returned as `null` when using the `SYSTEM` timezone in MySQL [#5457](https://github.com/sequelize/sequelize/pull/5457)
+
 # 3.19.3
 - [FIXED] `updatedAt` and `createdAt` values are now set before validation [#5367](https://github.com/sequelize/sequelize/pull/5367)
 - [FIXED] `describeTable` maintains proper enum casing in mysql [#5321](https://github.com/sequelize/sequelize/pull/5321)
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -47,10 +47,15 @@ module.exports = function (BaseTypes) {
       return value;
     }
 
+    var zoned;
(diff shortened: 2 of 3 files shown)


mickhansen:
Just FYI #5442 was not in 3.19.0 but 3.19.3 only

connor4312:
Okay, thanks. We'll set our dependency to 3.19.2 awaiting resolution on this  :smile:

janmeier:
Not sure I agree that this is a breaking change - I'm leaning more towards that returning the value as a string in the first place is a bug, and as such something we can fix in 3.x.

Not sure how best to fix it though - or if we should even fix it. The timezone option is meant as an option where you can specify timezone offset and timezone name - that it ever worked with `SYSTEM` was just a coincidence. Thoughts @mickhansen ?

mickhansen:
I like the idea of giving the user the option of reverting our default of UTC.
However perhaps we should allow the user to actually disable timezone setting, or use SYSTEM as a constant.

connor4312:
It would make sense to eliminate the bug altogether here, though I don't think dropping support for the SYSTEM timezone altogether is a great idea. Perhaps it would be better to allow dialects a "first shot" at parsing the timezone to pick up on special cases, like MySQL's SYSTEM or Postgres' STDoffset, and fall back to this standard parsing iff we don't already have a Date. And return the string if we really can't figure out what to do.

I don't know the internals of Sequelize well enough to implement this immediately, but if this approach sounds good to you I'll take a shot at implementing it in another PR.

PR #5460, opened 2016-02-20
Parse connection options from URI query strings.

protos37:
Though passing parameters via URI query string is a common feature with database connection string, sequelize doesn't supports this. I have briefly implemented this feature, which will help to put all options in a single URI, like `postgres:///dbname?host=/var/run/postgresql` or `sqlite:///?host=/path/to/file.sqlite`.

FYI:
PostgreSQL: [31.1.1.2. Connection URIs](http://www.postgresql.org/docs/9.4/static/libpq-connect.html)
Mysql: [5.1 Driver/Datasource Class Names, URL Syntax and Configuration Properties for Connector/J](https://dev.mysql.com/doc/connector-j/en/connector-j-reference-configuration-properties.html)
SQLite: [Uniform Resource Identifiers](https://www.sqlite.org/uri.html)

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -103,27 +103,33 @@ var Sequelize = function(database, username, password, options) {
 
   if (arguments.length === 1 || (arguments.length === 2 && typeof username === 'object')) {
     options = username || {};
-    urlParts = url.parse(arguments[0]);
-    // reset username and password to null so we don't pass the options as the username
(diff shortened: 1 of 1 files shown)


janmeier:
One of the existing connection string tests are failing https://github.com/sequelize/sequelize/blob/master/test/integration/configuration.test.js#L65

PR #5479, opened 2016-02-24
Support querying the through model

luin:
The pull request  #4739 has been inactive for a while. Querying the through model is a crucial feature for us, so I'd like to move this forward.

#4739 works very well. I wrote a test for it and the original commits are still included.


mickhansen:
Looks good, could you squash commits 2 & 4?

luin:
done

PR #5487, opened 2016-02-25
Execute promises sequentially rather than parallel

mkornblum:
Promise.map allows promises to execute in parallel, which can cause
deadlocks when cascading truncates. Promise.each executes promises
sequentially, which should ensure that we don't have two or more
TRUNCATE operations getting in each others' way.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -1009,9 +1009,15 @@ Sequelize.prototype.truncate = function(options) {
     }
   }, { reverse: false });
 
-  return Promise.map(models, function(model) {
+  var truncateModel = function(model) {
     return model.truncate(options);
-  });
+  };
+
+  if (options && options.cascade) {
+    return Promise.each(models, truncateModel);
+  } else {
+    return Promise.map(models, truncateModel);
+  }
 };
 
 /**


lib/sequelize.js
```diff
@@ -1009,7 +1009,7 @@ Sequelize.prototype.truncate = function(options) {
     }
   }, { reverse: false });
 
-  return Promise.map(models, function(model) {
+  return Promise.each(models, function(model) {
```
mickhansen: Make it dependant on the cascade option, parallel is fine when not cascading.

mkornblum:
It looks like the specs failed because options are not passed. I can default options to an empty object, or check for its existence, do you have any preference? 

On a sidenote, I have been unable to get tests running locally on OSX, or I'd like to think I wouldn't be pushing busted commits :)

mickhansen:
A `options = options || {}` is fine

mkornblum:
Managed to get the tests running locally but I can't reproduce the failure from Travis. Any advice? It's not obvious to me how the failure would relate to the change in any case...

mickhansen:
It was just an intermittent error, everything looks good.
If i could just get you to squash to a single descriptive commit that would be great.

mkornblum:
OK cool, thank you for your feedback! :+1:

PR #5510, opened 2016-03-01
rejectOnEmpty mode

### LINKED ISSUE #272: flag to emit an error if select result is null
hey,

I'm using async.js waterfall method together with sequelize to prevent 'callback chains of doom'. One scenario I'm regularly confronted with is that the first call in a chain is a select query and the following parts of the chain deal with the result of this select ... most of the time this results in code like this:

```javascript
async.waterfall([...
  function(callback) {
    Client.find(:id).done(function(err, res) {
      if(err !== null) callback(err);
      else if(res !== null) callback(new Error('Sorry no DB entry could be found!'));
      else callback(null, res);
    }):
  }, 
  function(client, callback) {...
```

... it would be nice to have an extra flag in Sequelize to produce an error if there is no result available so I could just pass trough my callback to done() and handle the errors in async's final method.

### LINKED ISSUE #5480: Find returns a null instance instead of catch
So this is my section of code:
```JavaScript
  /**
   * Verifies the specified user against the specified token
   * @param {string | number} user
   * @param {string} token
   * @returns {Object} promise
   */
  static verifyEmailVerificationToken(user, token) {
    return models.User.findOne({where: {$or: [{email: user}, {id: user}], emailToken: token}}).then(function(user) {
      return user.update({emailConfirmed: true, emailToken: null}).then(function(user) {
        eventService.emit('users.verifyEmailVerificationToken', user);
        return user;
      });
    });
  }
```

The `user.update` section should not be fired am I correct? The promise's "catch" callback should be the code that is triggered, however it is not. The "then" callback is being called with a null instance.

now I know this is the intended purpose, but shouldn't it invoke the catch instead?

sushantdhiman:
Allow finder methods to throw `RecordNotFound` error when query returns null set and `kenophobic` mode is `true`. This option can be set per method basis or with model options.

Closes #5480 
Closes #272

> Kenophobic means a person with fear of emptiness or void, You can suggest any other name if want to :)

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -301,7 +301,7 @@ error.InstanceError = function (message) {
 util.inherits(error.InstanceError, error.BaseError);
 
 /**
- * Thrown when a record was not found, Usually used with kenophobic mode (see message for details)
+ * Thrown when a record was not found, Usually used with rejectOnEmpty mode (see message for details)
  * @extends BaseError
  * @constructor
  */
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -299,3 +299,15 @@ error.InstanceError = function (message) {
   this.message = message;
 };
 util.inherits(error.InstanceError, error.BaseError);
+
+/**
+ * Thrown when a record was not found, Usually used with kenophobic mode (see message for details)
+ * @extends BaseError
+ * @constructor
+ */
+error.EmptyResultError = function (message) {
+  error.BaseError.apply(this, arguments);
+  this.name = 'SequelizeEmptyResultError';
+  this.message = message;
+};
+util.inherits(error.EmptyResultError, error.BaseError);
--- a/lib/model.js
+++ b/lib/model.js
@@ -35,7 +35,7 @@ var Model = function(name, attributes, options) {
     underscored: false,
     underscoredAll: false,
     paranoid: false,
-    kenophobic: false,
+    rejectOnEmpty: false,
     whereCollection: null,
     schema: null,
     schemaDelimiter: '',
@@ -1335,7 +1335,7 @@ Model.prototype.all = function(options) {
  * @param  {Object}                    [options.having]
  * @param  {String}                    [options.searchPath=DEFAULT] An optional parameter to specify the schema search_path (Postgres only)
  * @param  {Boolean}                   [options.benchmark=false] Print query execution time in milliseconds when logging SQL.
- * @param  {Boolean|Error Instance}    [options.kenophobic=false] Throws error when no record is found
+ * @param  {Boolean|Error Instance}    [options.rejectOnEmpty=false] Throws an error when no records found
  *
  * @see    {Sequelize#query}
  * @return {Promise<Array<Instance>>}
@@ -1356,10 +1356,10 @@ Model.prototype.findAll = function(options) {
   tableNames[this.getTableName(options)] = true;
   options = optClone(options);
 
-  _.defaults(options, { hooks: true, kenophobic: false });
+  _.defaults(options, { hooks: true, rejectOnEmpty: this.options.rejectOnEmpty });
 
-  //set kenophobic option from model config
-  options.kenophobic = options.kenophobic || this.options.kenophobic;
+    //set rejectOnEmpty option from model config
+    options.rejectOnEmpty = options.rejectOnEmpty || this.options.rejectOnEmpty;
 
   return Promise.bind(this).then(function() {
     conformOptions(options, this);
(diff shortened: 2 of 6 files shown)


mickhansen:
While i like the name i think `throwOnEmpty` or `rejectOnEmpty` would be more fitting.
Could you please also extend the API to allow a custom error from the user as an argument? (Would be helpfull for people doing error handling in APIs wanting to pass a 404 erorr instance

lib/model.js
```diff
@@ -1409,6 +1415,9 @@ Model.prototype.findAll = function(options) {
 };
 
 Model.$findSeparate = function(results, options) {
+  if(_.isEmpty(results) && options.kenophobic)
+    throw new sequelizeErrors.RecordNotFoundError();
+
```
mickhansen: Lets place this in the find call. findSeperate deals with other stuff.

sushantdhiman:
Hi @mickhansen , I am handling the `kenophobic` mode from `findAll` now, not using the `$findSeparate` anymore. Also custom error instance can be passed to `kenophobic` flag and it will be thrown, added tests for that. Please review

lib/model.js
```diff
@@ -1404,6 +1410,16 @@ Model.prototype.findAll = function(options) {
       return this.runHooks('afterFind', results, options);
     }
   }).then(function (results) {
+
+    //kenophobic mode
+    if(_.isEmpty(results) && options.kenophobic) {
+      if( (typeof options.kenophobic === 'function')) {
+        throw new options.kenophobic();
```
mickhansen: It could also be an already init'e derror instance: `kenophobic: new Status404Error('No such user')`

lib/errors.js
```diff
@@ -299,3 +299,15 @@ error.InstanceError = function (message) {
...
+/**
+ * Thrown when a record was not found, Usually used with kenophobic mode (see message for details)
+ * @extends BaseError
+ * @constructor
+ */
+error.RecordNotFoundError = function (message) {
+  error.BaseError.apply(this, arguments);
+  this.name = 'SequelizeRecordNotFoundError';
+  this.message = message;
+};
```
mickhansen: `RecordNotFound` implies single result, maybe a better generic would be `EmptyResultError`

mickhansen:
We probably still need an option name change, /cc @janmeier

lib/model.js
sushantdhiman: Not sure if I have done it in right way, but this will work for both cases now. Please check

sushantdhiman:
I am good with of any of these options

* `rejectOnEmpty`
* `throwOnEmpty`
* `kenophobic`

`kenophobic` is bit inspired by `paranoid`, your comments @janmeier

lib/model.js
mickhansen: yeah looks good

lib/model.js
```diff
@@ -1404,6 +1410,18 @@ Model.prototype.findAll = function(options) {
       return this.runHooks('afterFind', results, options);
     }
   }).then(function (results) {
+
+    //kenophobic mode
+    if(_.isEmpty(results) && options.kenophobic) {
+      if( (typeof options.kenophobic === 'function')) {
```
mickhansen: space after `if` (goes for entire diff), no space after `(`, no need to have double `()` here actually.

test/integration/model/findAll.test.js
```diff
@@ -1389,4 +1389,42 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
+      return Model.sync({ force: true })
+        .then(function() {
+          return expect(Model.findAll({
+            where: {
+              username: 'some-username-that-is-not-used-anywhere-for-sure-this-time'
+            }
+          })).to.eventually.be.rejectedWith(Sequelize.ConnectionError);
+        });
+    });
+  });
```
mickhansen: We need an extra unit test so that we handle all 3 cases, none passed, class passed or instance passed

sushantdhiman:
@mickhansen I have fixed code styling issues, Added new tests to check cases of `Initialized`, `Instance` and `undefined` value for `kenophobic` flag

sushantdhiman:
@janmeier waiting for you :)

mickhansen:
Will give him another day or so and we'll just make a decision.

lib/model.js
```diff
@@ -1353,7 +1356,10 @@ Model.prototype.findAll = function(options) {
   tableNames[this.getTableName(options)] = true;
   options = optClone(options);
 
-  _.defaults(options, { hooks: true });
+  _.defaults(options, { hooks: true, kenophobic: false });
```
janmeier: Couldn't this be ` _.defaults(options, { hooks: true, kenophobic: this.options.kenophobic });`?

janmeier:
I lol'd at kenophobic, but I think `rejectOnEmpty` will be better ;)

lib/model.js
sushantdhiman: Yes it can be, it will save one line of code too

sushantdhiman:
`rejectOnEmpty` :( , Ok I will do it

lib/model.js
sushantdhiman: Just tested, I didn't knew `defaults` behavior, but it will only overwrite `undefined` values, even if value is `false` it will not assign the `true` value from model options.

mickhansen:
@sushantdhiman Hey, why'd you close this? This was about ready to merge (just forgot about it since commits don't send me a notification :()

sushantdhiman:
Hi @mickhansen , I know you are busy but please review the pull requests when you get chance, if they are merged on time I feel encouraged to work on new one.

mickhansen:
Of course @sushantdhiman, but remember that commits don't notify me so i can't know you've made changes unless you comment.

If you can simply squash to a few descriptive commits (or change the descriptions to reflect the new name) we should be ready to go.

sushantdhiman:
Hi @mickhansen , I have changed the commit messages and squashed them

mickhansen:
Thanks @sushantdhiman, great work.

sushantdhiman:
Thanks @mickhansen

PR #5516, opened 2016-03-02
Fix DELETE WHERE 1=1 when complex objects passed in options.where to destroy or update

### LINKED ISSUE #4091: Delete with 1=1
Today I ran into a big issue, my delete query was look like this:
```
HashtagSubscribe.destroy({where: params}).then(function(result){
  // Success handler ...
})
```
And the following SQL generated:
```
DELETE FROM "hashtag_subscribe" WHERE 1=1
```
Yes, it dropped everything from `hashtag_subscribe` table, fortunately the app is in development state...

I printed out the `params` value and it looks like everything is correct:
```
sails.log.debug({where: params});
// Result
// debug: { where: { hashtag: '蘋果', userId: 1 } }
```

It took me long time to find out what is the problem: I use ProtoBuffer for communication with the clients and I use the decoded proto object as `params`. It looks like the original one, but there is some additional function in their prototype and the Sequelize abstract query builder is checked the object is a plain object:
https://github.com/sequelize/sequelize/blob/3.0/lib/dialects/abstract/query-generator.js#L1986

**My suggestion**
Never replace `WHERE` parameters in `DELETE` query with `1=1` if the `smth` is none of the recognised type. Instead of this, throw an error!

sushantdhiman:
Closes #4091

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2285,6 +2285,10 @@ Model.prototype.destroy = function(options) {
     throw new Error('Missing where or truncate attribute in the options parameter of model.destroy.');
   }
 
+  if (!options.truncate && !_.isPlainObject(options.where) && !_.isArray(options.where) && options.where._isSequelizeMethod !== true) {
+    throw new Error('Expected plain object, array or sequelize method in the options.where parameter of model.destroy.');
+  }
+
   options = Utils._.extend({
     hooks: true,
     individualHooks: false,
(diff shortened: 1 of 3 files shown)


mickhansen:
This doesn't actually verify #4091, the issue with #4091 is that the `where` object was not a POJO

sushantdhiman:
Hi @mickhansen , I misunderstood the issue earlier :). I have fixed it now, Please review

mickhansen:
Thank you again for your hard work @sushantdhiman

Issue #5517, opened 2016-03-02
AfterIncludeCreate hook


Verdier:
Hi,

Is there a way to be notified after a model and all its included models have been created? We need to perform action in this situation.

``` js
Operation.addHook('afterIncludeCreate', ...);
Operation.create({ breakdowns: [...], include: [Breakdown] });
```

If not, what about adding that hook, and what name would you choose for that ?

PR #5518, opened 2016-03-02
Add afterIncludeCreate hook

Verdier:
This PR is related to #5517

### DIFF
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -42,6 +42,7 @@ var hookTypes = {
   validationFailed: {params: 3},
   beforeCreate: {params: 2},
   afterCreate: {params: 2},
+  afterIncludeCreate: {params: 2},
   beforeDestroy: {params: 2},
   afterDestroy: {params: 2},
   beforeRestore: {params: 2},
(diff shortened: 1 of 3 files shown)


Verdier:
Refactored to be only called on create (update does not support include) and some tests added. Could you review @mickhansen?

mickhansen:
I'm considering whether we shouldn't just move the afterCreate hook.

Verdier:
No, sometime we need to modify the included values, for that we need a hook before creating the include.

Verdier:
If you want, we can reverse the hooks: `afterCreateBeforeInclude` and `afterCreate` but that's a breaking change (for us for exemple, because we modify the includes in the `afterCreate` hook).

mickhansen:
You modify the overall values passed to create in an afterCreate hook to modify the included elements created?

I'd say you should do that in beforeCreate to be more pure. It's unintended that changing values in afterCreate has an effect at all.

mickhansen:
I'm not a huge fan of having a lot of hooks with long names denoting specific points in a call path. Obviously i'd love to allow plugging into a method at any point to help plugins, just not sure how we design it best.

Verdier:
Your are true about modifying the included models into the `beforeCreate` hook. Would you like me to remove the new hook and to just move the `afterCreate` hook after the creation of the included models?

Verdier:
After reflexion @mickhansen , I agree with you that it's better to only move the afterCreate hook... Could you confirm?

mickhansen:
I think that's the better solution, however i'm worried about BC if others are depending on it like you have (although that really wasn't intended, but that doesn't change that we'd like to minimize breaks).

Verdier:
Ok @mickhansen, I've locally create the hook by replacing the `save` method at runtime. So I've a solution for our problem and the merge is no more a high priority for us, we can take the time to find the best solution. But I'm certain that we have to add/move a hook to be able to achieve that.

PR #5522, opened 2016-03-02
removeColumn for column with foreign key constraint

treythomas123:
This is failing for me on MS SQL Server 2k8 R2. It might not be an issue on other database types, or newer versions of MSSQL.

Just adding failing tests for now, and will look into implementing a fix soon.

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -507,7 +507,7 @@ var QueryGenerator = {
       return '[' + identifier.replace(/[\[\]']+/g,'') + ']';
   },
 
-  getForeignKeysQuery: function(table, databaseName) {
+  getForeignKeysQuery: function(table) {
     var tableName = table.tableName || table;
     var sql = [
       'SELECT',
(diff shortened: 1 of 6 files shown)


treythomas123:
The error message coming back from SQL Server is:

```
ALTER TABLE DROP COLUMN manager failed because one or more objects access this column.
```

This seems like a similar problem to the one that was addressed in PR [4990](https://github.com/sequelize/sequelize/pull/4990), but this is for foreign key constraints rather than default constraints.

treythomas123:
Looks like I may have inadvertently discovered another issue with schemas with this same PR (I don't see a way to tell the FK column which schema the referenced model is in). Took out the schema test for now, and will look into that separately.

mickhansen:
Lots of failures currently

treythomas123:
Yeah not sure why that's happening on mysql and mariadb. Seems like
the test is causing it. I'll try to look into it later today.

treythomas123:
Should be good now. Had to add the same logic (dropping foreign key constraints before the column) for MySQL/Maria.

mickhansen:
Thanks @treythomas123

PR #5446, opened 2016-02-18
Add failling test case for findAndCountAll

yarandoo:
This Pull request adds a failing test to reflect the issues found in our production systems when upgrading Sequelize version from 3.3.0 to 3.19.0.

* FindAndCountAll retrieves inconsistent results. See Issue: #5445

While developing this test case, the issue only reproduced when including the model "Tag" into the query. The simple insertion of this model in the findOptions causes a wrongly generated query leading to an incosistent result with COUNT:2 and ROWS.length=0.

See attached queries with and without the TAG include.

[PR-correct query - without tag include.txt](https://github.com/sequelize/sequelize/files/135671/PR-correct.query.-.without.tag.include.txt)
[PR-incorrect query when Tag relation added.txt](https://github.com/sequelize/sequelize/files/135672/PR-incorrect.query.when.Tag.relation.added.txt)

PR #5534, opened 2016-03-05
Fix findAndCountAll issues

### LINKED ISSUE #3009: Invalid SQL generated in postgres using group with attributes
This might be related (or the same as) #2968. Can't really tell if it's the same cause with a different effect.

Basically, I'm getting an error when using the postgres dialect and trying to do something like "count number of comments per post".

Here's my test code:

```javascript
var Sequelize = require('sequelize'),
    extend = require('extend');

var sequelize = new Sequelize('db', 'username', 'password', {
    host: 'host',
    port: 5432,
    dialect: 'postgres',
    native: true,
    define: {
        charset: 'utf8'
    }
});

var options = {
    timestamps: false,
    freezeTableName: true,
    schema: 'schema',
    underscored: true
};

var Post = sequelize.define('Post', {
    id: { type: Sequelize.INTEGER, autoIncrement: true, primaryKey: true },
    name: { type: Sequelize.STRING, allowNull: false }
}, extend({}, options, { tableName: 'test_post' }));

var Comment = sequelize.define('Comment', {
    id: { type: Sequelize.INTEGER, autoIncrement: true, primaryKey: true }
}, extend({}, options, { tableName: 'test_comment' }));

Post.hasMany(Comment);

var query = Post.findAll({
    attributes: [ [ Sequelize.fn('COUNT', Sequelize.col('Comments.id')), 'comment_count' ] ],
    include: [
        { model: Comment, attributes: [] }
    ],
    group: [ 'Post.id' ]
});

query
    .then(function(rows) {
        console.log(require('util').inspect(rows, false, null, true));
    })
    .catch(function(err) {
        console.error(err);
    })
    .done(function() {
        process.exit();
    });

```

It's generating this SQL:

```sql
SELECT
  "Post"."id", 
  COUNT("Comments"."id") AS "comment_count", 
  "Comments"."id" AS "Comments.id" 
FROM "schema"."test_post" AS "Post" 
LEFT OUTER JOIN "schema"."test_comment" AS "Comments"
  ON "Post"."id" = "Comments"."post_id" 
GROUP BY "Post"."id";
```

The error postgres spits out is:

```javascript
 [SequelizeDatabaseError: ERROR:  column "Comments.id" must appear in the GROUP BY clause or be used in an aggregate function
LINE 1: ...."id", COUNT("Comments"."id") AS "comment_count", "Comments"...
                                                             ^
]
  name: 'SequelizeDatabaseError',
  message: 'ERROR:  column "Comments.id" must appear in the GROUP BY clause or be used in an aggregate function\nLINE 1: ...."id", COUNT("Comments"."id") AS "comment_count", "Comments"...\n                                                             ^\n',
  parent:
   { [Error: ERROR:  column "Comments.id" must appear in the GROUP BY clause or be used in an aggregate function
   LINE 1: ...."id", COUNT("Comments"."id") AS "comment_count", "Comments"...
                                                                ^
   ]
     severity: 'ERROR',
     sqlState: '42803',
     messagePrimary: 'column "Comments.id" must appear in the GROUP BY clause or be used in an aggregate function',
     statementPosition: '64',
     sourceFile: 'parse_agg.c',
     sourceLine: '911',
     sourceFunction: 'check_ungrouped_columns_walker',
     sql: 'SELECT "Post"."id", COUNT("Comments"."id") AS "comment_count", "Comments"."id" AS "Comments.id" FROM "adstudio"."test_post" AS "Post" LEFT OUTER JOIN "adstudio"."test_comment" AS "Comments" ON "Post"."id" = "Comments"."post_id" GROUP BY "Post"."id";' },
  original:
   { [Error: ERROR:  column "Comments.id" must appear in the GROUP BY clause or be used in an aggregate function
   LINE 1: ...."id", COUNT("Comments"."id") AS "comment_count", "Comments"...
                                                                ^
   ]
     severity: 'ERROR',
     sqlState: '42803',
     messagePrimary: 'column "Comments.id" must appear in the GROUP BY clause or be used in an aggregate function',
     statementPosition: '64',
     sourceFile: 'parse_agg.c',
     sourceLine: '911',
     sourceFunction: 'check_ungrouped_columns_walker',
     sql: 'SELECT "Post"."id", COUNT("Comments"."id") AS "comment_count", "Comments"."id" AS "Comments.id" FROM "schema"."test_post" AS "Post" LEFT OUTER JOIN "schema"."test_comment" AS "Comments" ON "Post"."id" = "Comments"."post_id" GROUP BY "Post"."id";' },
  sql: 'SELECT "Post"."id", COUNT("Comments"."id") AS "comment_count", "Comments"."id" AS "Comments.id" FROM "schema"."test_post" AS "Post" LEFT OUTER JOIN "schema"."test_comment" AS "Comments" ON "Post"."id" = "Comments"."post_id" GROUP BY "Post"."id";'}
```

Basically the generated SQL has the `Comments.id` column in the `SELECT` statement even though it shouldn't be. Postgres can't handle that (I think MySQL just ignores it).

Anyway, I can't tell if this is a bug or if I'm doing something wrong. At the very least, sequelize probably shouldn't generate invalid SQL.

Here are the table definitions:

```
$ \d test_comment
 Column  |  Type   |                         Modifiers
---------+---------+-----------------------------------------------------------
 id      | integer | not null default nextval('test_comment_id_seq'::regclass)
 post_id | integer | not null
Indexes:
    "test_comment_pkey" PRIMARY KEY, btree (id)
Foreign-key constraints:
    "fk_test_comment_post" FOREIGN KEY (post_id) REFERENCES test_post(id) ON UPDATE CASCADE ON DELETE CASCADE

$ \d test_post
 Column |          Type          |                       Modifiers
--------+------------------------+--------------------------------------------------------
 id     | integer                | not null default nextval('test_post_id_seq'::regclass)
 name   | character varying(255) |
Indexes:
    "test_post_pkey" PRIMARY KEY, btree (id)
Referenced by:
    TABLE "test_comment" CONSTRAINT "fk_test_comment_post" FOREIGN KEY (post_id) REFERENCES test_post(id) ON UPDATE CASCADE ON DELETE CASCADE
```

I generated this using sequelize@2.0.0-rc8.

### LINKED ISSUE #4191: findAndCountAll() and count() pollutes options.includes object

### LINKED ISSUE #5445: FindAndCount all generates wrong query
When upgrading Sequelize from 3.3.0 to 3.19.0 FindAndCountAll retrieves inconsistent results with COUNT:5 and 0 rows.

We have identified the issue as being a miss generated query. Attached are the DIFF between the queries generated in 3.3.0 and 3.19.0 for our production system.

A failing test has been added via Pull Request: #5446

We were able to reproduce this issue since Sequelize 3.5.0.

For the moment we have rolled back our production system to Sequelize 3.3.0, till we can find a solution for this issue.

DIALECT: POSTGRES
Sequelize VERSION: 3.19.0
Node: 0.10.39

[incorrectly generated query 3.19.txt](https://github.com/sequelize/sequelize/files/135663/incorrectly.generated.query.3.19.txt)
[correctly generated query 3.3.0.txt](https://github.com/sequelize/sequelize/files/135664/correctly.generated.query.3.3.0.txt)

sushantdhiman:
Fixed the `subQuery` generation which cause nested `includes` to be applied on main query. It also add other tests and fixes achieved due to this fix.

Fixes #4191, Fixes #3009, Fixes #5445
Closes #5530, Closes #5446, Closes #4378

May be it will close more issues, I will check and include tests in further PRs

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,8 @@
 # Future
 - [ADDED] rejectOnEmpty mode [#272](https://github.com/sequelize/sequelize/issues/272) [#5480](https://github.com/sequelize/sequelize/issues/5480)
+- [FIXED] `findAndCountAll` generates invalid SQL, subQuery moves to LEFT OUTER JOIN [#5445](https://github.com/sequelize/sequelize/issues/5445)
+- [FIXED] `count` methods pollute the options.includes [#4191](https://github.com/sequelize/sequelize/issues/4191)
+- [FIXED] Invalid SQL generated when using group option along with attributes [#3009](https://github.com/sequelize/sequelize/issues/3009)
 - [ADDED] `beforeCount` hook [#5209](https://github.com/sequelize/sequelize/pull/5209)
 - [ADDED] `validationFailed` hook [#1626](https://github.com/sequelize/sequelize/issues/1626)
 - [FIXED] Mark index as `unique: true` when `type: 'UNIQUE'`. Fixes [#5351](https://github.com/sequelize/sequelize/issues/5351)
--- a/lib/model.js
+++ b/lib/model.js
@@ -486,6 +486,7 @@ var validateIncludedElements = function(options, tableNames) {
         include.subQuery = false;
       } else {
         include.subQueryFilter = false;
+        include.subQuery = include.subQuery || (include.hasParentRequired && include.hasRequired);
       }
     }
 
--- a/test/integration/include/findAndCountAll.test.js
+++ b/test/integration/include/findAndCountAll.test.js
@@ -207,5 +207,76 @@ describe(Support.getTestDialectTeaser('Include'), function() {
         expect(result.rows.length).to.equal(1);
       });
     });
+
+    it('should correctly filter, limit and sort when multiple includes and types of associations are present.', function() {
+      var TaskTag = this.sequelize.define('TaskTag', {
+        id: { type: DataTypes.INTEGER, allowNull: false, primaryKey: true, autoIncrement: true },
+        name: { type: DataTypes.STRING}
+      });
+
+      var Tag = this.sequelize.define('Tag', {
+        id: { type: DataTypes.INTEGER, allowNull: false, primaryKey: true, autoIncrement: true },
+        name: { type: DataTypes.STRING}
+      });
+
+      var Task = this.sequelize.define('Task', {
+        id: { type: DataTypes.INTEGER, allowNull: false, primaryKey: true, autoIncrement: true },
+        name: { type: DataTypes.STRING}
+      });
+      var Project = this.sequelize.define('Project', {
+        id: { type: DataTypes.INTEGER, allowNull: false, primaryKey: true, autoIncrement: true },
+        m: { type: DataTypes.STRING}
+      });
+
+      var User = this.sequelize.define('User', {
+        id: { type: DataTypes.INTEGER, allowNull: false, primaryKey: true, autoIncrement: true },
+        name: { type: DataTypes.STRING }
+      });
+
+      Project.belongsTo(User);
+      Task.belongsTo(Project);
+      Task.belongsToMany(Tag, {through: TaskTag});
+      // Sync them
+      return this.sequelize.sync({ force: true }).then(function () {
+        // Create an enviroment
+        return Promise.join(
+          User.bulkCreate([
+            { name: 'user-name-1' },
+            { name: 'user-name-2' }
+          ]),
+          Project.bulkCreate([
+            { m: 'A', UserId: 1},
+            { m: 'A', UserId: 2},
+          ]),
+          Task.bulkCreate([
+            { ProjectId: 1, name: 'Just' },
+            { ProjectId: 1, name: 'for' },
+            { ProjectId: 2, name: 'testing' },
+            { ProjectId: 2, name: 'proposes' }
+          ])
+        ).then(function() {
+          // Find All Tasks with Project(m=a) and User(name=user-name-2)
+          return Task.findAndCountAll({
+            limit: 2,
+            offset: 0,
+            order: [[ 'id', 'ASC' ]],
+            include: [
+              { model: Project,
+                where: { '$and': [ { m: 'A' } ] } ,
+                include: [ {
+                    model: User,
+                    where: { '$and': [ { name: 'user-name-2' } ] }
+                  }
+                ]
+              }, {model : Tag}
+
+            ]
+          });
+        });
+      }).then(function(result) {
+          expect(result.count).to.equal(2);
+          expect(result.rows.length).to.equal(2);
+      });
+    });
   });
 });
--- a/test/integration/include/findAndCountAll.test.js
+++ b/test/integration/include/findAndCountAll.test.js
@@ -237,45 +237,47 @@ describe(Support.getTestDialectTeaser('Include'), function() {
       Task.belongsTo(Project);
       Task.belongsToMany(Tag, {through: TaskTag});
       // Sync them
-      return this.sequelize.sync({ force: true }).then(function () {
+      return this.sequelize.sync({ force: true }).then(function() {
         // Create an enviroment
-        return Promise.join(
-          User.bulkCreate([
-            { name: 'user-name-1' },
-            { name: 'user-name-2' }
-          ]),
-          Project.bulkCreate([
+        return User.bulkCreate([
+          { name: 'user-name-1' },
+          { name: 'user-name-2' }
+        ]).then(function(u){
+          return Project.bulkCreate([
             { m: 'A', UserId: 1},
             { m: 'A', UserId: 2},
-          ]),
-          Task.bulkCreate([
+          ]);
+        }).then(function(p){
+          return Task.bulkCreate([
             { ProjectId: 1, name: 'Just' },
             { ProjectId: 1, name: 'for' },
             { ProjectId: 2, name: 'testing' },
             { ProjectId: 2, name: 'proposes' }
-          ])
-        ).then(function() {
+          ]);
+        })
+        .then(function() {
           // Find All Tasks with Project(m=a) and User(name=user-name-2)
           return Task.findAndCountAll({
-            limit: 2,
+            limit: 1,
             offset: 0,
-            order: [[ 'id', 'ASC' ]],
+            order: [[ 'id', 'DESC' ]],
             include: [
-              { model: Project,
+              {
+                model: Project,
                 where: { '$and': [ { m: 'A' } ] } ,
                 include: [ {
                     model: User,
                     where: { '$and': [ { name: 'user-name-2' } ] }
                   }
                 ]
-              }, {model : Tag}
-
+              },
+              { model : Tag }
             ]
           });
         });
       }).then(function(result) {
           expect(result.count).to.equal(2);
-          expect(result.rows.length).to.equal(2);
+          expect(result.rows.length).to.equal(1);
       });
     });
   });
--- a/test/integration/model/findAll/group.test.js
+++ b/test/integration/model/findAll/group.test.js
@@ -1,0 +1,57 @@
+'use strict';
+
+/* jshint -W030 */
+var chai = require('chai')
+, expect = chai.expect
+, Support = require(__dirname + '/../../support')
+, Sequelize = Support.Sequelize
+, DataTypes = require(__dirname + '/../../../../lib/data-types')
+, current = Support.sequelize;
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+  describe('findAll', function () {
+    describe('group', function () {
+      it('should correctly group with attributes, #3009', function() {
+
+        var Post = current.define('Post', {
+          id: { type: DataTypes.INTEGER, autoIncrement: true, primaryKey: true },
+          name: { type: DataTypes.STRING, allowNull: false }
+        });
+
+        var Comment = current.define('Comment', {
+          id: { type: DataTypes.INTEGER, autoIncrement: true, primaryKey: true },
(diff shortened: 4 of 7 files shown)


sushantdhiman:
@mickhansen @janmeier  Please have a look on this one when you get chance

changelog.md
```diff
@@ -1,4 +1,7 @@
 # Future
+- [Fixed] `findAndCountAll` generates wrong SQL query [#5445](https://github.com/sequelize/sequelize/issues/5445)
+- [Fixed] `count` methods pollute the options.includes [#4191](https://github.com/sequelize/sequelize/issues/4191)
```
mickhansen: I don't see a change for this, have you just verified that it does indeed work?

lib/model.js
```diff
@@ -484,6 +484,7 @@ var validateIncludedElements = function(options, tableNames) {
         include.subQuery = false;
       } else {
         include.subQueryFilter = false;
+        include.subQuery = include.hasParentRequired && (include.hasDuplicating || include.hasRequired);
```
mickhansen: This would overwrite the line above that supports an already set `include.subQuery`.
Simply prefix the hasParentRequired check with `include.subQuery || ` and it should be ok.

changelog.md
sushantdhiman: #4191 , I have added a test for that https://github.com/sushantdhiman/sequelize/commit/3f15e064f274c2219c982f4b7fffe757956f6ac0

mickhansen: Right, more tests are awesome! Just wanted to check if it was already broken, and doesn't seem like it was

sushantdhiman: Yes its want broken because someone fixed .findAll method, But still there was no test. So its a verification only, it was already fixed

sushantdhiman:
@mickhansen I have prefixed the condition, Can you somehow start the CI build, My local test run was ok.

mickhansen:
No CI active for this PR, very odd.

sushantdhiman:
Yeah Travis didn't started any CI because last commit had `[ci skip]` flag, I have renamed the commit. CI will start now.

sushantdhiman:
@mickhansen current CI has stuck, Can you please restart

mickhansen:
Please re-open and i'll make sure it gets a thorough review.

mickhansen:
Still no CI, very odd, no idea how to fix.

test/integration/model/findAll/group.test.js
```diff
@@ -0,0 +1,57 @@
...
+            group: [ 'Post.id' ]
+          });
+        }).then(function(posts) {
+          expect(parseInt(posts[0].get('comment_count'))).to.be.equal(3);
+          expect(parseInt(posts[1].get('comment_count'))).to.be.equal(2);
+        });
+      });
+    });
+  });
+});
```
mickhansen: What are you verifying in this test?

test/integration/include/findAndCountAll.test.js
```diff
@@ -207,5 +207,78 @@ describe(Support.getTestDialectTeaser('Include'), function() {
...
+                  }
+                ]
+              },
+              { model : Tag }
+            ]
+          });
+        });
+      }).then(function(result) {
+          expect(result.count).to.equal(2);
+          expect(result.rows.length).to.equal(2);
```
mickhansen: Shouldn't this equal 1? There's only one task that matches the combination of A and user-name-2.
Ideally when testing findAndCountAll it would be great if the LIMIT was lower than rows so we can verify that both parts work correctly.

lib/model.js
```diff
@@ -484,6 +484,7 @@ var validateIncludedElements = function(options, tableNames) {
         include.subQuery = false;
       } else {
         include.subQueryFilter = false;
+        include.subQuery = include.subQuery || include.hasParentRequired && (include.hasDuplicating || include.hasRequired);
```
mickhansen: We need to add the include to the subquery if it's parent is in the subquery and itself or it's children is required.
Do we care about `hasDuplicated` here?

test/integration/model/findAll/group.test.js
sushantdhiman: Its verification for https://github.com/sequelize/sequelize/issues/3009 , standalone commit https://github.com/sequelize/sequelize/pull/5534/commits/5d0f28a38cde3ecb434963264aa05858d89509cc which add this test. While browsing through count and find all issues I came across this. Issue was already fixed so I included it in this PR

test/integration/include/findAndCountAll.test.js
sushantdhiman: The result will match to 2, because for `user-name-2` there is one project `A` with 2 tasks `testing` and `purpose`.

Yes `findAndCountAll` test will be much better with LIMIT set to `1` so we can match correct count and records. I will modify it

lib/model.js
sushantdhiman: I just need the `true` value here because its causing the nested queries to go with LEFT OUTER JOIN even if they should be part of INNER JOIN.

If we consider we just need to respect `hasParentRequired` and `hasRequired` flags only, so there is no particular role for `hasDuplicating` here. I think I saw some test failing because of that. But that time code didn't included the `hasParentRequired` which I added later on after your suggestion. I will repeat the tests to check if we really need the `hasDuplicating` now.

mickhansen: I suggest `hasParentRequired`?
I can't really see why it's needed, whether or not an include parent is in the subquery doesn't matter unless the specific include needs to be in the subquery (at which point the parent should be forced, but that happens at the parent level).

test/integration/include/findAndCountAll.test.js
mickhansen: Ah right you're counting tasks, my bad.

lib/model.js
sushantdhiman: Ah you suggested to add `include.subQuery`, before adding that when I was looking for cause of the subQuery inclusion I found that `hasDuplicating` was also needed as deciding a factor. 

Now since logic has been changed, I'll need to check if we still need the `hasDuplicating` or any factor other than `include.subQuery`

sushantdhiman:
Hi @mickhansen , changed the condition and removed `hasDuplicating` as its not required now, rebased code, [CI link here](https://travis-ci.org/sequelize/sequelize/builds/116824746)

mickhansen:
Looks great @sushantdhiman, could you change the test to use a lower limit than the count?

sushantdhiman:
Let me know if this is correct behavior @mickhansen 

SELECT  `Task`.*, 

`Project.User`.`id` AS `Project.User.id`, `Project.User`.`name` AS `Project.User.name`, `Project.User`.`createdAt` AS `Project.User.createdAt`, `Project.User`.`updatedAt` AS `Project.User.updatedAt`, 

`Tags`.`id` AS `Tags.id`, `Tags`.`name` AS `Tags.name`, `Tags`.`createdAt` AS `Tags.createdAt`, `Tags`.`updatedAt` AS `Tags.updatedAt`, 

`Tags.TaskTag`.`id` AS `Tags.TaskTag.id`, `Tags.TaskTag`.`name` AS `Tags.TaskTag.name`, `Tags.TaskTag`.`createdAt` AS `Tags.TaskTag.createdAt`, `Tags.TaskTag`.`updatedAt` AS `Tags.TaskTag.updatedAt`, `Tags.TaskTag`.`TaskId` AS `Tags.TaskTag.TaskId`,

 `Tags.TaskTag`.`TagId` AS `Tags.TaskTag.TagId` FROM (
SELECT `Task`.`id`, `Task`.`name`, `Task`.`createdAt`, `Task`.`updatedAt`, `Task`.`ProjectId`, `Project`.`id` AS `Project.id`, `Project`.`m` AS `Project.m`, `Project`.`createdAt` AS `Project.createdAt`, `Project`.`updatedAt` AS `Project.updatedAt`, `Project`.`UserId` AS `Project.UserId` FROM `Tasks` AS `Task` INNER JOIN `Projects` AS `Project` ON `Task`.`ProjectId` = `Project`.`id` AND (`Project`.`m` = 'A') ORDER BY `Task`.`id` ASC LIMIT 1) 
**`--- LIMIT APPLIED IN SUB QUERY`**

AS `Task`  

INNER JOIN `Users` AS `Project.User` ON `Project.UserId` = `Project.User`.`id` AND (`Project.User`.`name` = 'user-name-2') 

LEFT OUTER JOIN (`TaskTags` AS `Tags.TaskTag` INNER JOIN `Tags` AS `Tags` ON `Tags`.`id` = `Tags.TaskTag`.`TagId`) ON `Task`.`id` = `Tags.TaskTag`.`TaskId` 

ORDER BY `Task`.`id` ASC;  **`--- LIMIT SHOULD BE HERE ?`**

This means first we filter by LIMIT then apply JOINS on filtered IDs, Is this correct?

mickhansen:
The limit has to be on the inner query so it limits the rows we want to limit and not the cartesian product.
If a include has a where/required and it's not duplicating it can be moved to the subquery, if it's duplicating it needs to be on the outer query and add an additional subquery as filter to the subquery.

sushantdhiman:
@mickhansen If thats example is correct way then great, I have changed the `limit` to `1`.

I would like to mention that when we change order to `ASC` then test will fail, because we do filtering in sub-query first, this will get first i.e. `task 1` (because limit is `1`), then we check for `project = "A" and user = "user-name-2" and task IN (1)`.

Since we have limited task id to `1`, the main result is empty because required task ids `3, 4` (for project `A` and user `user-name-2`) were left out during sub query filter.

mickhansen:
@sushantdhiman Ordering shouldn't matter, a bug if it does.

sushantdhiman:
Thanks

Issue #4193, opened 2015-07-27
Invalid specification/document on Instance#Model


soletan:
Generated documentation states Instance#Model being a function to call. However, it turns out to be reference.

BTW, I'm unsure about how to revise jsdoc properly on this (so autogenerated doc looks as required) and thus don't provide pull request here. Obviously jsdoc implicitly considers Instance#Model to be function as the block precedes a function invocation.

BTW2: Reading the code it looks quite strange to work with (unassigned) `this.Model` in a constructor. But I'm pretty sure you guys know what you do there. ;)

mickhansen:
Can you link to where it says it's a function?
`instance.Model` refers to the Model that the instance is an instance of, we are generally moving towards `$` prefixed private properties though.

soletan:
http://sequelize.readthedocs.org/en/latest/api/instance/#model-model

reading: 

> Model() -> Model
> 
> View code
> Returns the Model the instance was created from.

sushantdhiman:
@mickhansen Not sure how `mkdocs` works but I think this one has been fixed [here](https://github.com/sequelize/sequelize/blob/master/lib/instance.js#L102). If its not fixed let me know what to change so `mkdocs` can show it as `property` rahter than `function`

janmeier:
@sushantdhiman That should be ok :)

janmeier:
It might be something with how we render the docs though - let me check up on that

soletan:
Hm, IMHO the annotations are wrong according to JSDoc. The referenced code is using `@property` which seems to be wrong in that particular case and according to JSDoc as well as my personal recent experiences in other projects. JSDoc website is stating:

> Note that property tags must be used **in doc comments for the thing that they _are properties of_**, a namespace or a class for example. This tag is intended for simple collections of static properties, it does not allow you to provide @examples or similar complex information for each property, just the type, name and description.

I'd rather vote for @name here.

The second issue is with @return ... and that's most like causing the fuzz. If there is something _returning_ it must be a function.

I'd conclude with an annotation like that instead of yours:

```
/**
   * Refers to the Model the instance was created from.
   * @see {Model}
   * @name Instance#Model
   * @type {Model}
   */
```

tremby:
The same applies to the `sequelize` property (http://sequelize.readthedocs.org/en/latest/api/instance/#sequelize-sequelize) and from having a quick glance I have a feeling there may be others with the same issue.

sushantdhiman:
[...]

PR #5545, opened 2016-03-07
Docs: Instance.{Model,sequelize} aren't functions

tremby:


tremby:
Side question: should `Instance.Model` really have a capital M?

sushantdhiman:
You will also have to update the actual comments from`/lib`, these files are auto-generated by `JS-Doc`

tremby:
Ah, wasn't aware. I'll do that.

sushantdhiman:
Also this may help, https://github.com/sequelize/sequelize/issues/4193, I was actually trying to fix this but our JS-Doc generation is a bit `custom` I think.

tremby:
Ah, this isn't something I want to learn right now, then. I'll close this and comment on the other thread.

PR #5552, opened 2016-03-08
Bugfix for postgres adapter when nested transactions are used with quoteIdentifiers disabled

MPursche:
There is a bug in the postgres adapter when nested transactions are used in conjunction with the 'quoteIdentifiers' option set to false. The savepoint identifiers generated by sequelize contain special characters which require the identifier to be always quoted. However, if the 'quoteIdentifiers' option is set to false all identifiers in postgres are unquoted.
I added a bug fix that forces the quoting of savepoint identifiers while all other identifiers are still quoted/unquoted according to the configuration.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1697,7 +1697,8 @@ var QueryGenerator = {
    */
   startTransactionQuery: function(transaction) {
     if (transaction.parent) {
-      return 'SAVEPOINT ' + this.quoteIdentifier(transaction.name) + ';';
+      // force quoting of savepoint identifiers for postgres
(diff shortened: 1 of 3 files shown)


mickhansen:
I'd rather you apply this to the abstract one rather than introduce a duplicate piece of code.
The value is completely controlled by us so unless you attack the memory holding the uuid value i think we're safe :)

MPursche:
I wasn't sure if applying the change to the abstract query generator would break other adapters, but if you think its safe I will change it.

mickhansen:
It should be safe

janmeier:
Merged in https://github.com/sequelize/sequelize/commit/328e2bb779433cf2011a652cef95c1e50474be0d

janmeier:
Thanks @MPursche  :)

PR #5574, opened 2016-03-12
Add custom validation function testcases

xdarklight:
The new test ensures that promises can be returned by custom model
validation functions (there is already a test which does the same tests
for field-based custom validation functions).

### DIFF
--- a/test/unit/model/validation.test.js
+++ b/test/unit/model/validation.test.js
@@ -405,4 +405,153 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), function() {
 
     });
   });
+
+  describe('custom validation functions', function () {
+
+    var User = current.define('user', {
+      age: {
+        type: Sequelize.INTEGER,
+        validate: {
+          customFn: function(val, next) {
+            if (val < 0) {
+              next('age must be greater or equal zero');
+            } else {
+              next();
+            }
+          }
+        }
+      },
+      name: Sequelize.STRING
+    }, {
+      validate: {
+        customFn: function() {
+          if (this.get('name') === 'error') {
+            return Promise.reject(new Error('Error from model validation promise'));
+          }
+          return Promise.resolve();
+        }
+      }
+    });
+
+    before(function () {
+      this.stub = sinon.stub(current, 'query').returns(Promise.resolve(User.build()));
+    });
+
+    after(function () {
+      this.stub.restore();
+    });
+
+    describe('should not throw', function () {
+      describe('create', function () {
+        it('custom validation functions are successful', function () {
+          return expect(User.create({
+            age: 1,
+            name: 'noerror'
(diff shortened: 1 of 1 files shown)


xdarklight:
There is a failing test which is completely unrelated to this PR:
```
  1) [POSTGRES-NATIVE] associations scope N:M on the through model should create, find and include associations with scope values:
```

sushantdhiman:
You should move this test to `unit` tests, and modify it using `sinon` so it doesn't make any database call.

xdarklight:
@sushantdhiman I added unit-tests for custom validation functions in general.

My initial commit adds an integration-test which runs a query from a custom validation function.
So I have two questions:
1. is this use-case not worth adding an integration-test?
2. are there any examples which test this promise behavior?
(in my case the custom validation function returns a promise to a function which then marks validation as failed - it's not the custom validation function which is failing directly)

mickhansen:
@xdarklight validations wouldn't ever need to hit the database, even if they use promises, why we prefer a unit test. It should still work through all the motions, just stub out the part where it hits the database.

xdarklight:
@mickhansen @sushantdhiman ah, now I see
I have updated this PR (and changed the title so it matches the patch contents)

test/unit/model/validation.test.js
```diff
@@ -405,4 +405,163 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), function() {
...
+      }
+    });
+
+    before(function () {
+      this.stub = sinon.stub(current, 'query', function () {
+        return new Promise(function (resolve) {
+          resolve(User.build({}));
+        });
+      });
+    });
```
mickhansen: Not a required change, just something you can consider for the future.

Stub could be simplified to:

`this.stub = sinon.stub(current, 'query').returns(Promise.resolve(User.build()));`
or
`this.sub = sinon.stub(current, 'query').resolves(User.build())` (assuming sinon as promised).

If model does not need to be a new instance each time.

test/unit/model/validation.test.js
```diff
@@ -405,4 +405,163 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), function() {
...
+      name: Sequelize.STRING
+    }, {
+      validate: {
+        customFn: function() {
+          var user = this;
+          return User.findAll().then(function() {
+            if (user.name === 'error') {
+              throw new Error('Error from model validation promise');
+            }
+          });
```
mickhansen: Could be simplified to:

```js
if (this.get('name') === 'error') {
  return Promise.reject(new Error('Error from model validation promise'));
}
return Promise.resolve();
```

xdarklight:
@mickhansen thanks for reviewing this!
I just updated this PR - maybe you or @sushantdhiman could have another look at it

Issue #5547, opened 2016-03-08
date.utcOffset is not a function when creating model


blove:
The following exception is being thrown when executing model.create().

**Exception**
TypeError: date.utcOffset is not a function
   at DATE.$applyTimezone (/www/local.example.com/node_modules/sequelize/lib/data-types.js:457:19)
   at DATE.$stringify (/www/local.example.com/node_modules/sequelize/lib/dialects/mysql/data-types.js:39:37)
   at ABSTRACT.stringify (/www/local.example.com/node_modules/sequelize/lib/data-types.js:76:17)
   at Object.QueryGenerator.escape (/www/local.example.com/node_modules/sequelize/lib/dialects/abstract/query-generator.js:972:32)
   at Object.QueryGenerator.insertQuery (/www/local.example.com/node_modules/sequelize/lib/dialects/abstract/query-generator.js:298:28)
   at QueryInterface.insert (/www/local.example.com/node_modules/sequelize/lib/query-interface.js:492:33)
   at null.<anonymous> (/www/local.example.com/node_modules/sequelize/lib/instance.js:665:56)
   at tryCatcher (/www/local.example.com/node_modules/bluebird/js/release/util.js:16:23)
   at Promise._settlePromiseFromHandler (/www/local.example.com/node_modules/bluebird/js/release/promise.js:497:31)
   at Promise._settlePromise (/www/local.example.com/node_modules/bluebird/js/release/promise.js:555:18)
   at Promise._settlePromise0 (/www/local.example.com/node_modules/bluebird/js/release/promise.js:600:10)
   at Promise._settlePromises (/www/local.example.com/node_modules/bluebird/js/release/promise.js:683:18)
   at Async._drainQueue (/www/local.example.com/node_modules/bluebird/js/release/async.js:125:16)
   at Async._drainQueues (/www/local.example.com/node_modules/bluebird/js/release/async.js:135:10)
   at Immediate.Async.drainQueues [as _onImmediate] (/www/local.example.com/node_modules/bluebird/js/release/async.js:16:14)
   at processImmediate [as _immediateCallback] (timers.js:374:17)

**Model**
My model is written in TypeScript (and compiled to JavaScript for execution in Node.js with Express):

```javascript
var user = sequelize.define(NAME, {
  "id": { type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true },
  "email": Sequelize.STRING,
  "first_name": Sequelize.STRING,
  "last_name": Sequelize.STRING,
[...]

PR #5588, opened 2016-03-15
utcOffset fix

WillAnd31:
Fix for #5547. The package `moment-timezone` doesn't have a `utcOffset` method while the package `moment` does. It was probably just assumed that `moment-timezone` would have the `utcOffset` method.


janmeier:
Sorry for leaving this hanging, and thanks for you contribution @WillAnd31 !

Issue #5356, opened 2016-02-03
truncate() should allow resetting sequences.


sushantdhiman:
`Model.truncate()` method should optionally allow to reset the `sequence` of the `primary key`; It can be achieved by adding  these queries

* Postgres `truncate table table_name restart identity;`

xorbis:
Can truncate reset sequences now?
The reference by kylehotchkiss does not seem related.

janmeier:
The issue is still open, so that would be a no :)

janmeier:
The reference is not present in the linked issue, so must have been a typo

xorbis:
I look forward to this feature, thanks for the quick reply, you guys are awesome.

PR #5599, opened 2016-03-16
Cast value as string to pass new Validator.js string requirement

kylehotchkiss:
Long time user, first time PR-er :)

Trying to see if #5356 can be fixed by casting all values as `String()` in instance-validator before running validator function on them. My use cases for Sequelize are sort of limited so I'm not sure how well this will work, but let's let CI tell us!

### DIFF
--- a/lib/instance-validator.js
+++ b/lib/instance-validator.js
@@ -254,12 +254,14 @@ InstanceValidator.prototype._invokeCustomValidator = Promise.method(function(val
  */
 InstanceValidator.prototype._invokeBuiltinValidator = Promise.method(function(value, test, validatorType, field) {
   var self = this;
+  // Cast value as string to pass new Validator.js string requirement
+  var valueString = String(value);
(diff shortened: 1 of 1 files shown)


mickhansen:
CI passed.

I think i'm OK with this a dirty fix, we need to consider how we remove ourselves from a specific validator library in the future.

/cc @janmeier

kylehotchkiss:
Sweet! Just as user feedback: I find those validators really helpful and hope we can keep them. Somebody mentioned Joi in that ticket which wouldn't be very backwards compatible but could maybe better validate JSON / array columns? https://github.com/hapijs/joi

janmeier:
Yep, okay with me as well :)

mickhansen:
Validators are definitely something we want to support, i just don't think we should be tied to a specific library.

Discussion here: https://github.com/sequelize/sequelize/issues/5604 about making it pluggable.

PR #5605, opened 2016-03-17
Prove that Instance.equals() still doesn't work and fix it

MrVoltz:
May need to check if it works with Mongo and Postgres, will wait for Travis CI result.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -1012,27 +1012,25 @@ Instance.prototype.decrement = function(fields, options) {
 };
 
 /**
- * Check whether all values of this and `other` Instance are the same
+ * Check whether this and `other` Instance refer to the same row
  *
  * @param {Instance} other
  * @return {Boolean}
  */
 Instance.prototype.equals = function(other) {
-  var result = true;
-
-  if (!other || !other.dataValues) {
+  if (!other || !other.Model) {
     return false;
   }
 
-  Utils._.each(this.dataValues, function(value, key) {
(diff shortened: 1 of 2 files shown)


lib/instance.js
```diff
@@ -1012,27 +1012,25 @@ Instance.prototype.decrement = function(fields, options) {
...
-  Utils._.each(this.dataValues, function(value, key) {
-    if (Utils._.isDate(value) && Utils._.isDate(other.dataValues[key])) {
-      result = result && (value.getTime() === other.dataValues[key].getTime());
-    } else {
-      result = result && (value === other.dataValues[key]);
-    }
-  });
+  if (this.get(this.Model.primaryKeyAttribute, {raw: true}) !== other.get(this.Model.primaryKeyAttribute, {raw: true})) {
+    return false;
+  }
```
mickhansen: We should probably account for multiple primary keys here.

janmeier:
I guess this is technically a breaking change - some people might use the method to check whether a row has changed (even though it would be a weird usage, it would break). Coming from other languages (java for example), you would also expect that an `equals` method checks all attributes.

Should we tag this as 4.0, or write it of as a bug?

janmeier:
Just to be clear, I really think we need this method, since it could replace a lot of internal calls checking if `x.id == y.id` - but maybe it should be called `primaryEquals` or something similar until 4.0

janmeier:
What do you think @mickhansen, bug or 4.0?

mickhansen:
Not at all sure what people usually use `equals` for.
It's rather important whether or not people use it to match rows, or use it to match values.

Issue #5592, opened 2016-03-16
queryInterface.addColumn with references not add reference in mysql


ningshen:
queryInterface.addColumn with references not work in mysql

``` javascript
queryInterface.addColumn('Contents', 'ToolId',  {
    type: Sequelize.INTEGER, 
    references: {
        model: 'Tools', 
        key: 'id'
    }
})
````

will result sql
``` sql
ALTER TABLE `Contents` ADD `ToolId` INTEGER REFERENCES `Tools` (`id`);
```

this sql statement is not work.
 [mysql alter column](http://dev.mysql.com/doc/refman/5.7/en/alter-table.html)

add column sql with reference should be:
``` sql
ALTER TABLE `Contents` ADD `ToolId` INTEGER, ADD FOREIGN KEY (`ToolId`) REFERENCES `Tools` (`id`);
```

PR #5609, opened 2016-03-18
fix `addColumn` with reference not add reference in mysql

ningshen:
- solve issue: [queryInterface.addColumn with references not add reference in mysql](https://github.com/sequelize/sequelize/issues/5592)
- all existing tests pass
- add queryInterface.addColumn test for mysql


janmeier:
Thanks for the thorough PR with changelog and tests! :)

Issue #5620, opened 2016-03-22
An incorrect code example in doc


zjma:
[This section](http://sequelize.readthedocs.org/en/latest/docs/instances/#working-in-bulk-creating-updating-and-destroying-multiple-rows-at-once) in the documentation gives the following example of how to use ``Model.update``, which is incorrect.

```js
Task.bulkCreate([
  {subject: 'programming', status: 'executing'},
  {subject: 'reading', status: 'executing'},
  {subject: 'programming', status: 'finished'}
]).then(function() {
  return Task.update(
    { status: 'inactive' }, /* set attributes' value */,
    { where: { subject: 'programming' }} /* where criteria */
  );
}).then(function(affectedRows) {
  // affectedRows will be 2
  return Task.findAll();
}).then(function(tasks) {
  console.log(tasks) // the 'programming' tasks will both have a status of 'inactive'
})
```

The [API](http://sequelize.readthedocs.org/en/latest/api/model/#updatevalues-options-promisearrayaffectedcount-affectedrows) states that the argument currently called ``affectedRows`` will actually be an array instead of a number.

New users are likely to be misled.

janmeier:
Yep, that should be changed to `spread`

PR #5621, opened 2016-03-22
Updated code example of Model.update (#5620)

zjma:
A fix of issue #5620


janmeier:
I think we should rather point the user to the `spread` function, instead of manipulating the array manually. https://github.com/sequelize/sequelize/commit/ffc59576f3501a8755fabd56a756194a008e4945 - thanks for pointing the issue out though :)

Issue #5626, opened 2016-03-22
SQLite query interface misses transations


davidrapin:
Sequelize version I use: 3.19.3.

I am using the sequelize to perform table/column updates using the query-interface like this:
```JavaScript
function test() {
  var Sequelize = require('sequelize');
  var fs = require('fs');
  var dbFile = './test_sqlite.db';
  fs.unlinkSync(dbFile);
  var sequelize = new Sequelize('test_schema', null, null, {dialect: 'sqlite', storage: dbFile});
  
  // create table
  sequelize.define('table3', {
    'colA': {
      type: Sequelize.STRING,
      allowNull: true
    }
  }, {timestamps: false});
  return sequelize.sync().then(function() {
    var q = sequelize.getQueryInterface();

    // start transaction
    return sequelize.transaction(function(transaction) {
      var options = {transaction: transaction};
      return q.renameColumn('table3s', 'colA', 'colB', options);
    });
    // end transaction
  });
}
```

This fails, here is the SQL log:
````
Executing (b7b00719-e529-4e4b-8340-a8c278c0d4a2): BEGIN DEFERRED TRANSACTION;
Executing (b7b00719-e529-4e4b-8340-a8c278c0d4a2): -- SQLite is not able to choose the isolation level REPEATABLE READ.
Executing (b7b00719-e529-4e4b-8340-a8c278c0d4a2): -- SQLite does not support SET autocommit.
Executing (b7b00719-e529-4e4b-8340-a8c278c0d4a2): PRAGMA TABLE_INFO(`table3s`);
Executing (b7b00719-e529-4e4b-8340-a8c278c0d4a2): PRAGMA TABLE_INFO(`table3s`);
Executing (default): CREATE TEMPORARY TABLE IF NOT EXISTS `table3s_backup` (`id` INTEGER PRIMARY KEY, `colB` VARCHAR(255));
Executing (default): INSERT INTO `table3s_backup` SELECT `id`, `colA` AS `colB` FROM `table3s`;
Executing (default): DROP TABLE `table3s`;
Executing (default): DROP TABLE `table3s`;
Executing (default): DROP TABLE `table3s`;
Executing (default): DROP TABLE `table3s`;
Executing (default): DROP TABLE `table3s`;
Executing (b7b00719-e529-4e4b-8340-a8c278c0d4a2): ROLLBACK;
Unhandled rejection SequelizeTimeoutError: SQLITE_BUSY: database is locked
[...]

PR #5627, opened 2016-03-22
Added missing transaction in SQLite query interface

davidrapin:
#5626 Made SQLite query-interface use the full options-object (including transactions).

### DIFF
--- a/lib/dialects/sqlite/query-interface.js
+++ b/lib/dialects/sqlite/query-interface.js
@@ -29,14 +29,14 @@ var removeColumn = function(tableName, attributeName, options) {
   var self = this;
   options = options || {};
 
-  return this.describeTable(tableName).then(function(fields) {
+  return this.describeTable(tableName, options).then(function(fields) {
     delete fields[attributeName];
 
     var sql = self.QueryGenerator.removeColumnQuery(tableName, fields)
       , subQueries = sql.split(';').filter(function(q) { return q !== ''; });
 
     return Promise.each(subQueries, function(subQuery) {
-      return self.sequelize.query(subQuery + ';', { raw: true, logging: options.logging });
+      return self.sequelize.query(subQuery + ';', Utils._.assign({raw: true}, options));
     });
   });
 };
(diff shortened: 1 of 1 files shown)


lib/dialects/sqlite/query-interface.js
```diff
@@ -29,14 +29,16 @@ var removeColumn = function(tableName, attributeName, options) {
...
     delete fields[attributeName];
 
     var sql = self.QueryGenerator.removeColumnQuery(tableName, fields)
       , subQueries = sql.split(';').filter(function(q) { return q !== ''; });
 
+    var rawOptions = Utils._.clone(options);
+    rawOptions.raw = true;
     return Promise.each(subQueries, function(subQuery) {
-      return self.sequelize.query(subQuery + ';', { raw: true, logging: options.logging });
+      return self.sequelize.query(subQuery + ';', rawOptions);
```
mickhansen: `_.assign({raw: true}, options)` is a bit cleaner.

davidrapin: Ah I was looking for something like that, thanks! As long as it does not touch the original options object :)

mickhansen: It doesn't, `assign(target, ...sources)` copies from sources to target

davidrapin:
@mickhansen Ready to merge :)

lib/dialects/sqlite/query-interface.js
```diff
@@ -100,8 +102,9 @@ var renameColumn = function(tableName, attrNameBefore, attrNameAfter, options) {
     var sql = self.QueryGenerator.renameColumnQuery(tableName, attrNameBefore, attrNameAfter, fields)
       , subQueries = sql.split(';').filter(function(q) { return q !== ''; });
 
+    var rawOptions = Utils._.assign({raw: true}, options);
```
mickhansen: No reason for a intermediate variable.

mickhansen:
Just a single comment, please squash to a single commit after.

davidrapin:
Ok, fixed and squashed.

PR #5630, opened 2016-03-23
Try to clarify adding and setting associations

gonsfx:
With a little help from @mickhansen in Slack i realized that all of the `addX` and `setY` methods added to models through associations work on *existing* rows in the database and can not be used to insert new rows, which is what the `create` association methods are for.

With this PR, i try to clarify this in the docs because it might have clicked for me when i read it like this.

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -210,7 +210,7 @@ var BelongsToMany = function(source, target, options) {
     /**
      * Set the associated models by passing an array of instances or their primary keys. Everything that it not in the passed array will be un-associated.
      *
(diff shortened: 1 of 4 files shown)


mickhansen:
`persisted` might be a better word than `existing`.
/cc @janmeier

gonsfx:
@mickhansen you're right, `persisted` is probably better. waiting for opionions from @janmeier before i update.

mickhansen:
I think `persisted` is good, go ahead and make the change and i'll merge if we can't reach @janmeier :)

janmeier:
Persisted is fine by me as well :)

janmeier:
Ping @gonsfx

gonsfx:
@janmeier Thanks :)

Issue #3914, opened 2015-06-10
Limit = 0 (zero)


felipou:
I tried using limit=0 in a findAndCountAll query, expecting to get the total but no results, but it returns all rows, as if I hadn't used the limit option.

I tried to take a look at sequelize's source code, and found this dangerous pattern being used in some places:
```
if (options.offset && !options.limit) {
```

```
      , subQuery = options.subQuery === undefined ?
                   limit && options.hasMultiAssociation :
                   options.subQuery
```

(both examples in this file: `lib/dialects/abstract/query-generator.js`)

This makes it impossible to use a zero value, which I believe should be valid.

Although I agree this is a corner case, it could be dangerous, because this pattern is also used on some delete and update queries from what I've seen.

`<rant>`
And that's why I think this pattern (negating the value to see if it is 'valid') is so dangerous and often incorrect. It should never be used for numerical values. Even if you intentionally don't want the value zero to be accepted, I think it would be best to add a explicit check for zero instead of relying on this type of language specific pattern. And as we grow too used to using this type of check, if makes it harder every time to remember that we are using a numeric value and shouldn't use it.
`</rant>`

Sorry for the rant, but I think it may spark a useful and necessary discussion here :)

mickhansen:
The offset && !limit = LIMIT = 10000 has been around since before my time, although i never really understood why, so i'd be fine with removing that case (although i suppose it becomes a major bump).

I would tend to agree that checking for negative values regarding numbers is not the best, however for the subquery calculation i'm not sure it matters.

felipou:
It doesn't matter so much, but it requires the code using sequelize to do it's own checks regarding this.

I mean, if somehow I do some calculations to get the limit, and the limit turns out to be zero, I have to check for that and skip doing the query. I think it would be better (and safer!) if sequelize already accepted zero in a safer and more convenient way.

In my specific example, this is what is happening: I have a rest api which supports the limit and offset parameters. I validate these parameters and pass then on to sequelize. But what should I do if the limit is zero? Should I let sequelize, and consequently my API, return all rows as a result? It makes little sense to me. So either I should check for a limit of zero, and in that case I would do only a count query and return the total (this API always return the total), or I could filter the output, which would solve the issue "semantically", but could still be performance a problem if the query result it too large.

So, at least in the `addLimitAndOffset` function (the first example I showed before is from there), I think sequelize should support a limit of zero.

mickhansen:
[...]

PR #5631, opened 2016-03-23
Fix limit=0 

philip1986:
This PR should fix the issue https://github.com/sequelize/sequelize/issues/3914.
When `option.limit` was set to `0` it was considered as limit isn't  set at all. This PR changes those checks to check for the existence of the limit option.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1754,10 +1754,13 @@ var QueryGenerator = {
    */
   addLimitAndOffset: function(options, model) {
     var fragment = '';
-    if (options.offset && !options.limit) {
+    var limitExist = typeof options.limit !== "undefined" && options.limit !== null;
+    var offsetExit = typeof options.offset !== "undefined" && options.offset !== null;
+
+    if (offsetExit && !limitExist) {
       fragment += ' LIMIT ' + this.escape(options.offset) + ', ' + 10000000000000;
-    } else if (options.limit) {
-      if (options.offset) {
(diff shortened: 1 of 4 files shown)


lib/dialects/abstract/query-generator.js
```diff
@@ -1754,10 +1754,13 @@ var QueryGenerator = {
...
-    if (options.offset && !options.limit) {
+    var limitExist = typeof options.limit !== "undefined" && options.limit !== null;
+    var offsetExit = typeof options.offset !== "undefined" && options.offset !== null;
+
+    if (offsetExit && !limitExist) {
       fragment += ' LIMIT ' + this.escape(options.offset) + ', ' + 10000000000000;
-    } else if (options.limit) {
-      if (options.offset) {
+    } else if (limitExist) {
+      if (offsetExit) {
```
mickhansen: You can make this all a bit cleaner by using `options.limit != null`, a oddity of javascript allows you to check both undefined and null with that.

mickhansen:
Please remove the merge commit, do a proper rebase instead.
`limit: 0` would result in no rows being returned, right?

philip1986:
Ok, I will do do. Yes thats it will return an empty result set.

lib/dialects/abstract/query-generator.js
philip1986: Unfortunately jshint settings doesn't allow `!= null`

philip1986:
Will close this PR and reopen it after rebase.

PR #5632, opened 2016-03-23
fix limit=0 issue

philip1986:
Reopen https://github.com/sequelize/sequelize/pull/5631

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -17,7 +17,8 @@
   "nonbsp": true,
   "maxdepth": 8,
   "quotmark": true, // deprecated
-
+  "-W041": false,
+  
   /* relaxing options */
   "laxbreak":true,
   "laxcomma":true,
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1754,10 +1754,12 @@ var QueryGenerator = {
    */
   addLimitAndOffset: function(options, model) {
     var fragment = '';
-    if (options.offset && !options.limit) {
+
+    /*jshint eqeqeq:false*/
+    if (options.offset != null && options.limit == null) {
       fragment += ' LIMIT ' + this.escape(options.offset) + ', ' + 10000000000000;
-    } else if (options.limit) {
-      if (options.offset) {
+    } else if (options.limit != null) {
+      if (options.offset != null) {
         fragment += ' LIMIT ' + this.escape(options.offset) + ', ' + this.escape(options.limit);
       } else {
         fragment += ' LIMIT ' + this.escape(options.limit);
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -417,8 +417,13 @@ var QueryGenerator = {
 
   addLimitAndOffset: function(options) {
     var fragment = '';
-    if (options.limit) fragment += ' LIMIT ' + this.escape(options.limit);
-    if (options.offset) fragment += ' OFFSET ' + this.escape(options.offset);
+    /*jshint eqeqeq:false*/
+    if (options.limit != null) {
+      fragment += ' LIMIT ' + this.escape(options.limit);
+    }
+    if (options.offset != null) {
+       fragment += ' OFFSET ' + this.escape(options.offset);
+    }
 
     return fragment;
   },
--- a/test/unit/dialects/postgres/query-generator.test.js
+++ b/test/unit/dialects/postgres/query-generator.test.js
@@ -291,7 +291,17 @@ if (dialect.match(/^postgres/)) {
           expectation: 'SELECT * FROM "myTable" AS "myTable" ORDER BY "myTable"."id" DESC;',
           context: QueryGenerator,
           needsSequelize: true
+        },{
+          title: 'uses limit 0',
+          arguments: ['myTable', {limit: 0}],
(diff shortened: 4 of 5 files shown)


lib/dialects/postgres/query-generator.js
```diff
@@ -417,8 +417,13 @@ var QueryGenerator = {
 
   addLimitAndOffset: function(options) {
     var fragment = '';
-    if (options.limit) fragment += ' LIMIT ' + this.escape(options.limit);
-    if (options.offset) fragment += ' OFFSET ' + this.escape(options.offset);
+
+    if (typeof options.limit !== "undefined" && options.limit !== null) {
```
sushantdhiman: What is the difference between ```(options.limit)``` and ```typeof options.limit !== "undefined" && options.limit !== null```, I think both does same thing?

May be we can use ```(!!options.limit)``` that will be better than checking for all types

lib/dialects/postgres/query-generator.js
```diff
@@ -417,8 +417,13 @@ var QueryGenerator = {
...
+      fragment += ' LIMIT ' + this.escape(options.limit);
+    }
+    if ( typeof options.limit !== "undefined" && options.limit !== null) {
```
sushantdhiman: We also need to check if ```offset``` exists or not ?

lib/dialects/postgres/query-generator.js
mickhansen: a truthy check also hits `options.limit = 0` which this PR is about.

sushantdhiman: ```!!option.limit || options.limit === 0``` will be cleaner ?

mickhansen: No need for the double negator in that case, but yeah might be slightly cleaner :)
Could also simply check `typeof options.limit === 'number'`

sushantdhiman: ```typeof options.limit === 'number'``` is perfect for this case

philip1986: The type could be also a string.
There multiple options e.g.: 

```
typeof options.limit === "number" || typeof options.limit === "string"
options.limit || options.limit === 0 
options.limit != null // will not pass jshint 
```

lib/dialects/postgres/query-generator.js
philip1986: Actually not since default offset is 0, but I did it to keep a kind of consistency.

lib/dialects/postgres/query-generator.js
mickhansen: strings actually aren't supported, they are escaped so the SQL is incorrect.

.jshintrc
```diff
@@ -17,7 +17,8 @@
   "nonbsp": true,
   "maxdepth": 8,
   "quotmark": true, // deprecated
-
+  "-W041": false,
+  
   /* relaxing options */
```
sushantdhiman: @mickhansen should we allow comments inside json structure, I know @philip1986 is adding comment because there were some comments in `.jshintrc` already, but I think we should get rid of all comments inside any `json` structure.

mickhansen: .jshintrc isn't proper JSON so i feel that comments are OK.

lib/dialects/abstract/query-generator.js
```diff
@@ -1754,10 +1754,14 @@ var QueryGenerator = {
    */
   addLimitAndOffset: function(options, model) {
     var fragment = '';
-    if (options.offset && !options.limit) {
+    /*jshint eqeqeq:false*/
+    var limitExist = options.limit != null;
+    var offsetExit = options.offset != null;
```
mickhansen: Having vars for this check kind of irks me, i'd prefer the duplicated code.

philip1986: I think that makes it more readable, but ok I'll change it.

test/unit/dialects/postgres/query-generator.test.js
```diff
@@ -291,7 +291,27 @@ if (dialect.match(/^postgres/)) {
...
+         title: 'uses offset 0',
+         arguments: ['myTable', {offset: 0}],
+         expectation: 'SELECT * FROM "myTable" OFFSET 0;',
+         context: QueryGenerator
+       }, {
+        title: "uses offset '0'",
+        arguments: ['myTable', {offset: '0'}],
+        expectation: "SELECT * FROM \"myTable\" OFFSET '0';",
+        context: QueryGenerator
+      },{
```
mickhansen: You can remove the string tests from both dialects, we don't official support limit/offset as strings anymore.

lib/dialects/abstract/query-generator.js
mickhansen: Sorry for the bikeshedding, but i actually find the other style more readable :)

philip1986: There's no accounting for taste ;)

philip1986:
@mickhansen are you going to merge it?

mickhansen:
Definitely, i didn't get notified for the later commits.
Can you squash to a single descriptive commit?

philip1986:
Nice :) 
I squashed it already.

sushantdhiman:
Please add changelog as well @philip1986

PR #5655, opened 2016-03-27
Typo in examples

adrukh:


adrukh:
My bad, the original documentation is correct.

PR #5656, opened 2016-03-27
fix include association as

ivnivnch:
Aliases for associations:

```
Model.findAll({
  include: [
    {association: Model.associations.someAssociation, as: 'name1', where: {/* where1 */}},
    {association: Model.associations.someAssociation, as: 'name2', where: {/* where2 */}}
  ]
})
```

### DIFF
--- a/test/unit/model/include.test.js
+++ b/test/unit/model/include.test.js
@@ -436,5 +436,18 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         expect(options.include[0].subQuery).to.equal(false);
       });
     });
+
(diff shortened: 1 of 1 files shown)


mickhansen:
We don't currently support `as` differing from `as` on the defined association.
We'd be open to this but we'd need a larger PR covering a lot more test cases.

janmeier:
Ping @ivnivnch

janmeier:
@ivnivnch We don't get notified of new commits, just so you know :)

We'd need both unit and integration tests

janmeier:

PR #5660, opened 2016-03-27
use IS NULL if value is undefined

vafada:
This PR fixes the issue discussed in #4851

We were bitten by this bug when we are doing 

`
Person.findAll({
  where: {
    firstName: undefined
  }
})
`

The resulting query is 

`SELECT "id", "first_name" AS "firstName", "last_name" AS "lastName" FROM "person" AS "person" WHERE "person"."first_name" = NULL;`

which is wrong according to postgres, mysql and mssql docs:

http://www.postgresql.org/docs/9.1/static/functions-comparison.html
http://dev.mysql.com/doc/refman/5.7/en/working-with-null.html
https://msdn.microsoft.com/en-us/library/ms188795.aspx

Postgres:
`Do not write expression = NULL because NULL is not "equal to" NULL. (The null value represents an unknown value, and it is not known whether two unknown values are equal.) This behavior conforms to the SQL standard.`

MySQL:
`To test for NULL, use the IS NULL and IS NOT NULL operators, as shown here:`

This fix changes the query so that passing an `undefined` or `null` in the where clause would use the `IS/IS NOT` syntax

`SELECT "id", "first_name" AS "firstName", "last_name" AS "lastName" FROM "person" AS "person" WHERE "person"."first_name" IS NULL;`

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -2219,9 +2219,9 @@ var QueryGenerator = {
         }.bind(this));
       }
 
-      if (comparator === '=' && value === null) {
+      if (comparator === '=' && (value === null || value === undefined)) {
         comparator = 'IS';
-      } else if (comparator === '!=' && value === null) {
+      } else if (comparator === '!=' && (value === null || value === undefined)) {
         comparator = 'IS NOT';
       }
 
--- a/test/unit/sql/where.test.js
+++ b/test/unit/sql/where.test.js
@@ -91,6 +91,12 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
       mssql: '[deleted] IS NULL'
     });
 
+    testsql('deleted', undefined, {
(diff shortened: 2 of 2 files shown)


mickhansen:
Not sure this is the correct course of action. If we currently do `= NULL` it will result in no rows, which is a lot better than finding null rows.

undefined is not the same as null, ideally we either throw away undefined columns completely or we thro an error.

vafada:
But `= null` doesnt make sense in sql land.

I understand undefined is not the same as null but thats in javascript land. 

I suggest throw an error. Ignoring undefined value is much worse. If i put three where condition i expect my query to have  three condition instead of two.

mickhansen:
You're not using null, you're using undefined.
Having an `undefined` value is actually just bad input since it's not ever a proper comparison value.

An error would be backwards compat breaking though, and if you're doing dynamic conditions you might be used to them not doing anything.

janmeier:

PR #5664, opened 2016-03-28
fix setter for instances with raw option

desnet:
This fix allows you to send constructed instance from raw data in setAccociation () method. This avoids the silly, unnecessary models relations.

Example:

```js
var Products = db.define('Products', {
 name: {type: DataTypes.STRING, allowNull: false},
 description: {type: DataTypes.TEXT, allowNull: true}
});

var ProductFeatures = db.define('ProductFeatures', {
 name: {type: DataTypes.STRING, allowNull: false},
 type: {type: DataTypes.STRING, allowNull: false}
});

var ProductsToFeatures = db.define('ProductsToFeatures', {
 featureId: {type: DataTypes.INTEGER, allowNull: false},
 productId: {type: DataTypes.INTEGER, allowNull: false},
 value: {type: DataTypes.STRING, allowNull: false}
});

Products.BelongsToMany(ProductFeatures, {through: ProductsToFeatures, foreignKey: 'productId', otherKey: 'featureId'});

Products.findById(1).then(function (product) {
 var features = [];
 features.push(ProductFeatures.build({
   id: 3,
   ProductsToFeatures: {
      value: 112
   }
 }, {raw: true, isNewRecord: false }));

 // Set features from raw data with additional fields
 product.setFeatures(features).then(function (features) {
 
 });
});
```

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -582,7 +582,7 @@ BelongsToMany.prototype.injectSetter = function(obj) {
         if (!newObj) {
           obsoleteAssociations.push(currentRow);
         } else {
-          var throughAttributes = newObj[association.through.model.name];
+          var throughAttributes = newObj.get(association.through.model.name);
           // Quick-fix for subtle bug when using existing objects that might have the through model attached (not as an attribute object)
           if (throughAttributes instanceof association.through.model.Instance) {
             throughAttributes = {};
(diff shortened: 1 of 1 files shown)


mickhansen:
I welcome the change ,but you've broken a few tests you'll need to fix before we can merge.

janmeier:
Ping @desnet

PR #5665, opened 2016-03-28
Tests for #3534, values modified in validate hooks are saved

### LINKED ISSUE #3534: Attributes modified in hooks are not saved
If you add a hook to (for exemple) a `User` model, and save an instance of the model with some attributes modified, such as:

``` javascript
user.email = 'toto@toto.com';
user.save();
```

And if you modify some user attributes into the (for exemple...) `afterValidate` hook:

``` javascript
function afterUserValidate(user) {
  if(user.changed('email')) {
    return u.md5(user.email).then((md5) => user.emailMd5 = md5);
  }
}
```

The `emailMd5` attribute is not saved. This is because dirty attributes are computed before the _pre-save_ hooks.

So my question: could we imagine to compute dirty attributes **after** the _pre-save_ hooks?

sushantdhiman:
Closes #3534 

Added a test which verify that values changed in the validation hooks are preserved.

### DIFF
--- a/test/integration/hooks/validate.test.js
+++ b/test/integration/hooks/validate.test.js
@@ -41,6 +41,74 @@ describe(Support.getTestDialectTeaser('Hooks'), function() {
       });
     });
 
+    describe('#3534, hooks modifications', function() {
+      it('fields modified in hooks are saved', function() {
+        var self = this;
+
+        this.User.afterValidate(function(user, options) {
+          //if username is defined and has more than 5 char
+          user.username = user.username
+                          ? (user.username.length < 5 ? null : user.username)
+                          : null;
+          user.username = user.username || 'Samorost 3';
+
+        });
+
+        this.User.beforeValidate(function(user, options) {
+          user.mood = user.mood || 'neutral';
+        });
+
+
+        return this.User.create({username: 'T', mood: 'neutral'}).then(function(user) {
+          expect(user.mood).to.equal('neutral');
+          expect(user.username).to.equal('Samorost 3');
+
+          //change attributes
+          user.mood = 'sad';
+          user.username = 'Samorost Good One';
+
+          return user.save();
+        }).then(function(uSaved) {
+          expect(uSaved.mood).to.equal('sad');
+          expect(uSaved.username).to.equal('Samorost Good One');
+
+          //change attributes, expect to be replaced by hooks
+          uSaved.username = 'One';
+
+          return uSaved.save();
+        }).then(function(uSaved) {
+          //attributes were replaced by hooks ?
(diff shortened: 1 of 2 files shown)


sushantdhiman:
@mickhansen @janmeier Please review

test/integration/hooks/validate.test.js
```diff
@@ -41,6 +41,74 @@ describe(Support.getTestDialectTeaser('Hooks'), function() {
       });
     });
 
+    describe('#3534, hooks modifications', function() {
+      it('fields modified in hooks are saved', function() {
```
mickhansen: `... in validation hooks ...`

mickhansen:
LGTM, lots of assertions though so would like another set of eyes from @janmeier

sushantdhiman:
Yes these assertions were required because I wanted to confirm if there was any issue with newly created records or old records fetched by `find` methods.

test/integration/hooks/validate.test.js
sushantdhiman: actually when test will run it will show like 

```
  [MYSQL] Hooks
      #validate
         #create
            ✓ should return the user
         #3534, hooks modifications
           ✓ fields modified in hooks are saved
```

Thats why I didn't included validation in description

mickhansen: both the issue and test deal specifically with validation hooks though. Not a major issue in any case.

mickhansen:
They are definitely required, just need an extra set of eyes so i don't miss a possible bad one :)

test/integration/hooks/validate.test.js
```diff
@@ -41,6 +41,74 @@ describe(Support.getTestDialectTeaser('Hooks'), function() {
...
+
+          //change attributes, expect to be replaced by hooks
+          uSaved.username = 'One';
+
+          return uSaved.save();
+        }).then(function(uSaved) {
+          //attributes were replaced by hooks ?
+          expect(uSaved.mood).to.equal('sad');
+          expect(uSaved.username).to.equal('Samorost 3');
+          return self.User.findById(uSaved.id);
```
janmeier: Do we need to re-find here? I get that it is to test that the values actually get saved to the DB and not just set on the object, but don't we have other tests for that

sushantdhiman: But we have no test to check this while hooks are changing the values. I am looking for any cases that may involve 

- Creating
- Finding
- Updating

because they all pass from different methods and hooks, we need to check if there is any issue in these methods when hooks modification are in effect

janmeier: True :)

janmeier:
Thanks once again for you work @sushantdhiman !

PR #5666, opened 2016-03-28
Model.destroy: add model to options

scragg0x:
This gets added in BulkCreate but not destroy. I need to figure out the model when using a global BulkDelete hook.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2320,6 +2320,7 @@ Model.prototype.destroy = function(options) {
   this.$injectScope(options);
 
   Utils.mapOptionFieldNames(options, this);
+  options.model = self;
 
   return Promise.try(function() {
     // Run before hook


scragg0x:
Sorry for the crappy commit message.  Anyhow it would be nice to have this available in the hook.

mickhansen:
Where is this added for bulkCreate?

scragg0x:
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L2242

mickhansen:
Hm, that only works for the afterBulkCreate hook then, not before.

scragg0x:
Well I'm using after hooks but I suppose it being available in before hooks could be helpful as well. I will open a PR and set the model early so it hits beforeBulkCreate and see if it passes.

 Thanks for the merge.

Issue #5339, opened 2016-01-31
Virtual fields should not be reset after reload()


reidrankin:
After setting a virtual field on an Instance, calling reload() results in the virtual field being reset. The docs say this function 

> Refresh[es] the current instance in-place, i.e. update[s] the object with current data from the DB and return[s] the same object.

Seeing as how there is no data from the DB to update virtual fields with, they should not be reset.

I suspect that issue #5289 is why they are set to `undefined`. I suspect that if this issue were fixed, they would be reset to the default values instead.

PR #5667, opened 2016-03-28
Populate virtual fields with default values on instance creation

### LINKED ISSUE #5289: Virtual fields should be assigned default values when an Instance is loaded from the database
Currently, it seems that default values are only propagated on build(). This leads to virtual fields (which, in my application, must always be set to the default `null` or a valid value) being `undefined` when loaded from the database.

krmannix:
Fixes #5289 and #5339

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -767,7 +767,7 @@ Instance.prototype.reload = function(options) {
     this.Model._virtualAttributes.forEach(function (key) {
       var virtualAttribute = this.get(key, {raw: true});
       if (virtualAttribute !== undefined) {
-        reload.set(key, virtualAttribute);
+        reload.set(key, virtualAttribute, {raw: true});
       }
     }, this);
     return reload;
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -765,8 +765,9 @@ Instance.prototype.reload = function(options) {
   .then(function(reload) {
     // persist virtual fields that have been previously set
     this.Model._virtualAttributes.forEach(function (key) {
-      if (this.dataValues[key] !== undefined) {
-        reload.dataValues[key] = this.dataValues[key];
+      var virtualAttribute = this.get(key, {raw: true});
+      if (virtualAttribute !== undefined) {
+        reload.set(key, virtualAttribute);
       }
     }, this);
     return reload;
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -61,6 +61,16 @@ var initValues = function(values, options) {
     }
   }
 
+  if (this.Model._hasVirtualAttributes) {
+    // set virtual attributes to their default values
+    this.Model._virtualAttributes.forEach(function (key) {
+      if (values[key] === undefined) {
+        this.set(key, Utils.toDefaultValue(this.Model._defaultValues[key]), defaultsOptions);
+        delete values[key];
+      }
+    }, this);
+  }
+
   this.set(values, options);
 };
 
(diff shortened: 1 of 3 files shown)


lib/instance.js
```diff
@@ -753,6 +763,15 @@ Instance.prototype.reload = function(options) {
     }
   })
   .then(function(reload) {
+    // persist virtual fields that have been previously set
+    this.Model._virtualAttributes.forEach(function (key) {
+      if (this.dataValues[key] !== undefined) {
+        reload.dataValues[key] = this.dataValues[key];
+      }
```
mickhansen: If possible i prefer you use `get(key, {raw: true})` and `set(key, {raw: true})`

test/integration/instance/update.test.js
```diff
@@ -161,13 +161,17 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
...
       }).then(function () {
         return user.reload();
       }).then(function () {
         expect(user.validateSideAffected).to.be.equal(10);
-        expect(user.validateSideEffect).not.to.be.ok;
+        expect(user.validateSideEffect).to.be.equal(5);
+        return this.User.findOne({id: user.id });
+      }).then(function (user) {
+        expect(user.validateSideAffected).to.be.equal(10);
+        expect(user.validateSideEffect).to.not.be.ok;
```
mickhansen: Why is this test changed? It should not be affected since it has no default value.

mickhansen: Ah, because we're using reload and now the value is set to what it was before the reload call.
That raises a good question though, a side effect should really be gone if using reload.
Side effects are different from default values.

test/integration/model/attributes/types.test.js
```diff
@@ -144,6 +154,54 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
+        it('should be populated to a default value on reload() if not previously set', function () {
+          return this.User.create({
+            field1: 'fizzbuzz',
+          }).then(function(user) {
+            expect(user.virtualWithDefault).to.equal('cake');
+            return user.reload();
+          }).then(function(user) {
+            expect(user.virtualWithDefault).to.equal('cake');
+            expect(user.virtualWithoutDefault).to.equal(undefined);
+            user.set('virtualWithDefault', undefined);
```
krmannix: Looking at this now, I also want to add a test case where the virtual field is not in the `attributes` option passed into reload & make sure it doesn't get set to the default value

krmannix:
This should be good to be looked at - the last commit clarified the behavior of virtual fields so that even if attributes are defined on a query or reload, if the virtual field has a default value it will still be populated. My thought on this ended up being different than what my previous line note indicated, since always populating the fields with their default value (if it exists) seems more consistent with the purpose of virtual fields

lib/instance.js
```diff
@@ -753,6 +763,16 @@ Instance.prototype.reload = function(options) {
     }
   })
   .then(function(reload) {
+    // persist virtual fields that have been previously set
+    this.Model._virtualAttributes.forEach(function (key) {
+      var virtualAttribute = this.get(key, {raw: true});
+      if (virtualAttribute !== undefined) {
+        reload.set(key, virtualAttribute);
```
mickhansen: Probably want to set with `raw: true` aswell.

b-fett:
looks like it is still an issue.

fzn0x:
> looks like it is still an issue.

Sorry, what's the issue on your side? I think this PR is not an issue but a feature

PR #5670, opened 2016-03-29
Performance improvement for #3414

### LINKED ISSUE #3414: findAll with Multiple items in array large performance degredation
Using sequelize to perform a findAll on a table that has an index on the column used in where clause saw a huge performance decrease from 1.7.x to 2.0.x when given an array with multiple values. See code below:

```javascript
Model.findALL({
    where: {
        columnName: ids
    }
});
```

*ids* can either be a single integer or array of integers.  I benchmarked this in 1.7.8, 1.7.10, 2.0.0-dev1, 2.0.0, and 2.0.4.

With 1 id, the 2.0.x branches saw ~5-10% improvement.
With 5 ids the 2.0.x branches saw ~30% decrease in performance
With 150 ids the 2.0.x branches saw ~130% decrease in performance

(using Node@0.10.36 and MySQL@5.6.17)

sushantdhiman:
After studying #3414 and its performance test suite I tracked back the performance degradation causes. It appears to be these reasons

* Generic performance hit due to Promises (we cant do much about it ?)
* `optClone` method (consumes `300+ ms` for 1501 array elements in `findAll`)
* Hitting the database (450 - 550 ms)
* Other JS execution

I was able to improve the performance by adding a small fix to `optClone` method. These are the results

| Elements / Exec. Count | Fix Applied  | Total Time (ms) | Node | Sequelize |
|---|---|---|---|---|
| 1501 / 5000 | No | 11557 | 0.12 | 1.7.0 |
| 1501 / 5000 | No | 11403 | 0.12 | 1.7.0 |
| 1501 / 5000 | No | 13173 | 0.12 | 2.0.0 |
| 1501 / 5000 | No | 13210 | 0.12 | 2.0.0 |
| 1501 / 5000 | **No** | **16447** | 0.12 | 3.20 |
| 1501 / 5000 | **No** | **17447** | 0.12 | 3.20 |
| 1501 / 5000 | **Yes** | **15707** | 0.12 | 3.20 |
| 1501 / 5000 | **Yes** | **16024** | 0.12 | 3.20 |

It improves the performance a bit. Its certainly not in line with `1.7.0` which is fastest of all these but considering Promise overhead and other new feature may be we can now close #3414

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -311,7 +311,7 @@ var optClone = Model.prototype.__optClone = Model.prototype.$optClone = function
   options = options || {};
   return Utils.cloneDeep(options, function(elem) {
     // The InstanceFactories used for include are pass by ref, so don't clone them.
-    if (elem &&
+    if (elem && typeof elem === 'object' &&
       (
         elem._isSequelizeMethod ||
         elem instanceof Model ||


sushantdhiman:
@mickhansen @janmeier Please review

mickhansen:
Okay so you sped this up by not doing instanceof checks unless they are needed.
How long is the execution time for optClone in that scenario now?

All your node versions say 0.12, i assume some are meant to be 0.10?

sushantdhiman:
optClone improvements are about `40ms+` per call for an array with 1501 elements. Since `findAll` call sit two times we have over `80ms +` improvements 

All these tests are meant for 0.12. But I have recorded tests for 0.10, 0.12 and 5.9.0, 0.12 version were  generally fastest for all sequelize version so I have included only those.

mickhansen:
So we still spend `200ms+` on a pre-database call for findAll? That's still quite horrendous.
Not calling optClone twice in findAll would probably also be great.

mickhansen:
Merging, but keep your benchmarks around, would like to do more work on this.

mickhansen:
Is optClone called recursively for each element in the $in array?

sushantdhiman:
Yes it is called recursively, its actually an modified version for `cloneDeep` so it checks each nested child element in `options`

mickhansen:
Okay, i guess we could test if the value is an array of plain values and then return that directly rather than clone for a speed boost aswell.

sushantdhiman:
Yes we can do an `isArray` type check to improve performance, also we should remove the double call for `optClone`. Will you be working on this, Can I submit PR based upon our discussion?

mickhansen:
`if (_.isArray(elem) && elem.length && _.isPrimitive(elem[0]) return elem`.
Please go ahead :)

sushantdhiman:
That check is adding overhead, I have tried to defer it after `typeof elem == 'object'` so we will check only when `object` type is under inspection, but it will still not improve performance.

It appears there is no object/array directly passed to `cloneDeep` handler method by lodash. All values I see are `numbers`, `string`. I need to look how cloneDeep actually works to improve it further. May be it pass only the leaf node values?

Issue #5639, opened 2016-03-24
matchesModelAndAlias doesn't ensure model/as definition was used


pmckinnon:
The isSameModel test in $injectScope is not ensuring that the include was defined using the model/as format instead of an association, resulting in an exception accessing scopeInclude.model.name when the include was specified using an association.

https://github.com/ojolabs/sequelize/commit/0262943798cdbb963f626f9ed27b5aabd9b015c8

hendrul:
In addition I think there should be a better query merging logic that could be used in $injectScope and in Model.scope methods. Includes should be deeply merge!

mickhansen:
@pmckinnon PR welcome.

@hendrul A top level method for merging find options would be great. We're also looking at letting the user provide a custom merge function for scopes.

PR #5672, opened 2016-03-29
model/as check on $injectScope

pmckinnon:
- Ensure scope include uses model/as format instead of association before performing same-model test

https://github.com/sequelize/sequelize/issues/5639


mickhansen:
A test broke with this change.

PR #5673, opened 2016-03-29
Minor Docs fix

krmannix:
Noticed that this was a little out of order when reading the docs. Changed the source comment and the documentation as well

### DIFF
--- a/docs/api/datatypes.md
+++ b/docs/api/datatypes.md
@@ -265,6 +265,7 @@ sequelize.define('user', {
   }
 })
 ```
+In the above code the password is stored plainly in the password field so it can be validated, but is never stored in the DB.
 
 VIRTUAL also takes a return type and dependency fields as arguments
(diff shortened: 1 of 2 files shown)


docs/api/datatypes.md
```diff
@@ -280,9 +281,6 @@ Return type is mostly useful for setups that rely on types like GraphQL.
 }
 ```
 
-In the above code the password is stored plainly in the password field so it can be validated, but is never stored in the DB.
-__Aliases:__ NONE
```
janmeier: `NONE` is actually an alias for virtual https://github.com/krmannix/sequelize/blob/datatype-doc-fix/lib/data-types.js#L969

Otherwise LGTM

krmannix: Ah gotcha, I'll throw that back in. Thanks!

krmannix:
Amended commit to include `alias: NONE`

janmeier:
Thanks :)

PR #5684, opened 2016-03-31
Update validation to return null

yads:
Clarify validate documentation as validate emits undefined if there are no errors not null.

### DIFF
--- a/lib/instance-validator.js
+++ b/lib/instance-validator.js
@@ -72,6 +72,8 @@ InstanceValidator.prototype.validate = function() {
   })).then(function() {
     if (self.errors.length) {
       return new sequelizeError.ValidationError(null, self.errors);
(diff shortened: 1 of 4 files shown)


mickhansen:
`docs/api` are generated from code, so your edit will be overwritten.
We should probably look at actually returning null :)

yads:
Oh hah, didn't know that, there was an 'edit this in github' link on the page. I'll take a look at having the code return null.

yads:
@mickhansen I've updated the PR to return null on validation success

janmeier:
Thanks @yads 

(And thanks to github for adding a squash and merge button :)! https://github.com/sequelize/sequelize/commit/ac9e3d1c3dce24c7555bfd348425d49eecd0a344

PR #5688, opened 2016-04-01
Before connection hook

### LINKED ISSUE #5687: On demand / on connection credentials

mickhansen:
Closes #5687

Adds a before connection hook as well as supporting a plain object constructor.
Enables use of dynamic credentials for databases by letting the user define empty username/password in the Sequelize options object and then injecting on a per connection basis.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -15,7 +15,8 @@ var url = require('url')
   , Promise = require('./promise')
   , Hooks = require('./hooks')
   , Instance = require('./instance')
-  , Association = require('./associations/index');
+  , Association = require('./associations/index')
+  , _ = require('lodash');
 
 /**
  * This is the main class, the entry point to sequelize. To use it, you just need to import sequelize:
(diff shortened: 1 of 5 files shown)


lib/sequelize.js
```diff
@@ -172,12 +188,16 @@ var Sequelize = function(database, username, password, options) {
 
   this.options.hooks = this.replaceHookAliases(this.options.hooks);
 
+  if ((['', null, false].indexOf(config.password) > -1) || (typeof config.password === 'undefined')) {
```
janmeier: If it's null, then set it to null?

test/unit/sequelize.test.js
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+    expect(sequelize.config.database).to.equal(database);
+    expect(sequelize.config.username).to.equal(username);
+    expect(sequelize.config.password).to.equal(password);
+    expect(sequelize.config.host).to.equal(host);
+    expect(sequelize.config.port).to.equal(port);
+  });
+
+  it('should connection uri, options', function () {
```
janmeier: support

lib/sequelize.js
mickhansen: Old code, didn't want to add cleanup as part of a feature change.

PR #5696, opened 2016-04-04
Fix defaultValues getting overwritten on build

rplopes:
### Pull Request check-list

### Description of change

Creating new model instances in sequelize has an odd behaviour whereby their default values aren't cloned and are therefore shared mutable state between all instances of a class. This means that if we change one if its attributes in one model instance, it gets changed to all other instances as well. Example:

```js
// For a `Model` model with the attribute `data` with `defaultValue` set to `{ foo: 'bar' }`:
var model1 = db.Model.build();
model1.data.foo; // 'bar'
model1.data.foo = 'biz';
model1.data.foo; // 'biz'

var model2 = db.Model.build();
model2.data.foo; // 'biz'
```

The proposed fix applies a deep clone to the objects that are defined as `defaultValue`. This also addresses the edge case of sequelize methods as default values which, due to their nature, aren't meant to be cloned.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# Future
+- [FIXED] Fix defaultValues getting overwritten on build
+
 # 3.21.0
 - [FIXED] Confirmed that values modified in validation hooks are preserved [#3534](https://github.com/sequelize/sequelize/issues/3534)
 - [FIXED] Support lower case type names in SQLite [#5482](https://github.com/sequelize/sequelize/issues/5482)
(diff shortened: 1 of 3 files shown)


mickhansen:
LGTM, @janmeier

janmeier:
Yep, LGTM as well - just needs a changelog entry then we can merge

rplopes:
Added changelog entry.

janmeier:
Thanks :)

PR #5707, opened 2016-04-06
fixed options not being passed in the DropTableQuery

eraye1:
This fixes schemas not being added as part of the drop statements when sync is called.

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -97,7 +97,7 @@ var QueryGenerator = {
     options = options || {};
     var query = 'DROP TABLE IF EXISTS <%= table %><%= cascade %>;';
     return Utils._.template(query)({
-      table: this.quoteTable(tableName, options),
+      table: this.quoteTable({tableName: tableName, schema: options.schema}),
       cascade: options.cascade ? ' CASCADE' : ''
     });
   },
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -97,7 +97,7 @@ var QueryGenerator = {
     options = options || {};
     var query = 'DROP TABLE IF EXISTS <%= table %><%= cascade %>;';
     return Utils._.template(query)({
-      table: this.quoteTable(tableName),
+      table: this.quoteTable(tableName, options),
       cascade: options.cascade ? ' CASCADE' : ''
     });
   },


janmeier:
Seems to work fine already
```js

}, {
  schema: 'foo'
})

return sequelize.sync({
  force: true,
  logging: console.log
})
```

```js
DROP TABLE IF EXISTS "foo"."users" CASCADE;
CREATE TABLE IF NOT EXISTS "foo"."users" ("id"   SERIAL , PRIMARY KEY ("id"));
```

mickhansen:
`quoteTable` automatically supports `{tableName, schema}` as its first argument. So simply pass that as the tableName arg when you need schema when running the method directly.

eraye1:
Yep, I was just passing it in the wrong place.

I thought the documentation referenced ```Model.schema(schema_name)``` as a way to set schemas as well?

Also: I still feel like it's a bit weird that I create a connection, set the schema that the connection should be using and then I call define using that connection, but the models that are defined by that connection don't have the schema propagate to them.  It seems like the models are bound to the connection, which means it should be under the namespace that it has (in this case the schema).

mickhansen:
@eraye1 Not sure what you mean, is it documented somewhere that you can set a global schema option? It likely works through the global define options.

Models are defined based on a sequelize instance, but in the future we'd like to decouple that further.

eraye1:
so if models have the property that they can have a schema set on them (this is not ideal because if you have multitenant architectures, you won't know the full set of schemas beforehand) and they are created from a sequelize instance, which also has a property that it can have a schema set on it, doesn't it make sense that when a sequelize instance has the schema set on it, that the models it creates should also have the same schema set on them?

eraye1:
@janmeier dropping the table works if you define the schema on the models but not if you define in the options for the sync.

try:
```

});

return sequelize.sync({
  force: true,
  logging: console.log,
  schema: 'foo'
})

```
DROP TABLE IF EXISTS "users" CASCADE;
CREATE TABLE IF NOT EXISTS "foo"."users" ("id"   SERIAL , PRIMARY KEY ("id"));
```

The bug is where it drops the table.

**Comment by eraye1** [comment]
Also: if we set the schema on the sync, it doesn't get added to the foreign keys correctly:

```
var Entity = sequelize.define('Entity', {
    name: DataTypes.STRING,
  }, {
    classMethods: {
      associate: function (models) {
        Entity.belongsToMany(models.User, {through: models.Permission});
      }
    }
  });

  var Permission = sequelize.define('Permission', {
    id: {
      type: DataTypes.UUID,
      allowNull: false,
      primaryKey: true
    },
    permissionType: {
      type: DataTypes.STRING,
      defaultValue: 'read',
    }
  });

    id: {
      type: DataTypes.UUID,
      allowNull: false,
      primaryKey: true
    },
    username: {
      type: DataTypes.STRING,
      allowNull: false,
      unique: true
    }
  }, {
    classMethods: {
      associate: function (models) {
        User.belongsToMany(models.Entity, {through: Permission});
      }
    }
  });

return sequelize.sync({
  force: true,
  logging: console.log,
  schema: 'foo'
})
```

The sync function will fail here because:
```
Executing (default): CREATE TABLE IF NOT EXISTS "foo"."Entities" ("name" VARCHAR(255), "id" UUID , "networkId" UUID, "identifiedNetworkId" UUID, "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL, "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL, "EntityId" UUID REFERENCES "Entities" ("id") ON DELETE SET NULL ON UPDATE CASCADE, PRIMARY KEY ("id"));
Error syncing:  { [SequelizeDatabaseError: relation "Entities" does not exist]
```

Because we only pass in the model to query-interface.attributeToSQL(attributes), so we don't add the schema, which we set in the sync options.

mickhansen:
Passing `schema` to `sync` is not a supported property at all AFAIR.
In any case the fix in this PR is not proper since it breaks quite a few postgres tests.

janmeier:
Actually it is in the API docs, but we have never tested for it - so support is spotty at best.

http://docs.sequelizejs.com/en/latest/api/sequelize/#syncoptions-promise

I think we should stick to only defining the schema on the model

PR #5712, opened 2016-04-07
Include with OR in outer WHERE

Verdier:
Hi, 

I need to review this pull request with one of you @mickhansen or @janmeier before adding documention or tests. This allow things like that:

```js
Post.findAll({ 
 where: {
  $or: {
   title: 'hello',
   '$comments.text$': 'hello'
  }
 },
 include: [{
  Model: Comment,
  joinInSubQuery: true
}]
 limit: 10,
})
```

This include the join in the Post subQuery and allow to filter on the `title` *OR* on the `comments.text` filed.

```SQL
SELECT
  ...
FROM (SELECT
        ...
      FROM "posts" AS "post" LEFT OUTER JOIN "comments" AS "comments" ON "post"."id" = "comments"."postId"
      WHERE ("post"."title" = 'hello' OR "comments"."text" = 'hello')
      LIMIT '2') AS "task"
  LEFT OUTER JOIN "comments" AS "comments" ON "post"."id" = "comments"."postId";
```

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1374,6 +1374,10 @@ var QueryGenerator = {
           joinQueries.mainQuery.push(joinQueryItem);
         }
 
+        if (include.joinInSubQuery && subQuery) {
+          joinQueries.subQuery.push(joinQueryItem);
+        }
+
         if (include.include) {
           include.include.filter(function (include) {
             return !include.separate;


janmeier:
But doesn't this break the limit - then we are again limiting the number of rows from the cartesian product, not the parent?

Might be solveable by making the subquery a `WHERE id IN (SELECT DISTINCT id subquery)`

mickhansen:
There are already options for this internally, no need to add another one.

Verdier:
@janmeier no this does not break the limit because the join is only used in the `WHERE` clause, not in the `SELECT` one.

@mickhansen what internal option are you thinking of?

mickhansen:
@Verdier it still breaks limit since you're still trying to limit a cartesian product.
`include.subQuery` and `include.duplicating` are options.

(Why is post renamed to task in this SQL? :D)

Verdier:
@mickhansen I just have a look to the `include.subQuery` and `include.duplicating` options, but none of them allow to force the join in the sub query...

(The task was a mistake when refactoring the query for the issue ^^)

mickhansen:
Why are you forcing it to the subquery? You break limit, you might aswell remove the limit or tell Sequelize that there's no cartesian product (although there is).

Verdier:
You are right @mickhansen. I propose to add a `orWhere` option to perform an `OR` with the other `where` options:

```js
Post.findAll({ 
 where: {
   title: 'hello',
 },
 include: [{
  Model: Comment,
  orWhere: {
   text: 'hello'
  }
}]
 limit: 10,
});
```

which result in the following query:

```sql
SELECT
  ...
FROM (SELECT
        ...
      FROM "posts" AS "post"
      WHERE
        "post"."title" = 'hello' OR --< HERE
        (SELECT "postId"
        FROM "comments" AS "comment"
        WHERE
          ("comment"."postId" = "post"."id" AND
           "comment"."comment" = 'hello')
        LIMIT 1) IS NOT NULL
      LIMIT 10) AS "post"
  LEFT OUTER JOIN "comments" AS "comments" ON "post"."id" = "comments"."postId"
```
Or, without limit:
```sql
SELECT
  ...
FROM "posts" AS "post"
  LEFT OUTER JOIN "comments" AS "comments" ON "post"."id" = "comments"."postId"
WHERE 
 "post"."title" = 'hello' OR --< HERE
 "comments"."comment" = 'hello';
```

That solve the general problem of searching something in the parent model `OR` in one of the included ones. What do you think of that @mickhansen and @janmeier?

tnrich:
@Verdier @mickhansen @janmeier  was this issue ever resolved? 

I too would like to search "something in the parent model `OR` in one of the included ones", but am not sure what to add to my query:

```
include: [{
  Model: Comment,
  where: {
   text: 'hello'
  }
}]
```

I do not see `include.subQuery` or `include.duplicating` mentioned anywhere in the docs. Does one of these solve the above problem?

Thanks very much!!

Issue #4198, opened 2015-07-28
validation for fields with UUID type is run against parameters for LIKE queries


nemophrost:
MyModel.findAll({ where: { id: { $like: "12345678%" } } })

results in a validation error: "12345678%" is not a valid uuid

BridgeAR:
This should probably be fixed by #4186

janmeier:
@BridgeAR, how will datatypes refactoring fix this? :-) 

I'm thinking more along the lines of restricting some validations to only run on create, and making it possible to completely disable them

janmeier:
Merged into https://github.com/sequelize/sequelize/issues/4314

PR #5713, opened 2016-04-07
Enable type validation for all queries

rplopes:
### Pull Request check-list

### Description of change

When the validation on datatypes got reverted and put behind a flag, a new rule was added to allow them only on `INSERT` and `UPDATE` queries. However, these validations are useful for the remaining queries, as well, like `SELECT`.

This PR proposes enabling these validations for all queries. This would still only be applicable to anyone who has enabled the datatypes validation flag, which is disabled by default.

This PR also removes some redundant checks on the affected conditional (`field` and `field.type` have already been checked in an outer conditional) and reduces the level of nesting used to validate the datatype.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # Future
 - [FIXED] Fix defaultValues getting overwritten on build
 - [FIXED] Queue queries against tedious connections
+- [ADDED] Enable type validation for all queries
 
 # 3.21.0
 - [FIXED] Confirmed that values modified in validation hooks are preserved [#3534](https://github.com/sequelize/sequelize/issues/3534)
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -669,8 +669,8 @@ RANGE.prototype.validate = function(value) {
 var UUID = ABSTRACT.inherits();
 
 UUID.prototype.key = UUID.key = 'UUID';
-UUID.prototype.validate = function(value) {
-  if (!Validator.isUUID(value)) {
+UUID.prototype.validate = function(value, options) {
+  if (!Validator.isUUID(value) && (!options || !options.acceptStrings || typeof value !== 'string')) {
     throw new sequelizeErrors.ValidationError(util.format('%j is not a valid uuid', value));
   }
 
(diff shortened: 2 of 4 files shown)


janmeier:
Applying the validation on select queries breaks like queries (as an example)

```js
where: {
  uuid: {
    $like: '%foo%'
  }
}
```
See https://github.com/sequelize/sequelize/issues/4198

janmeier:
See also these two tests which are now breaking - https://github.com/sequelize/sequelize/blob/9ea63e68bbcf9f7a7f6fa4603541e279ba375746/test/unit/model/validation.test.js#L351-L372, one because of the uuid like problem i mentioned, and the other because its trying to validate the array as a string.


I'm not saying I don't want validations on select statements, it just needs some more work. In particular the context (`INSERT, UPDATE, SELECT` etc.) should probably be passed to the validator, so it can take the appropriate steps - For example not validating uuids on select

rplopes:
These problems are well documented in https://github.com/sequelize/sequelize/issues/4314, my fault for not going though it and testing locally first.

I've pushed two separate commits, each addressing one of those problems. These create more moving parts, so right now they're just proposals to solve those problems.

Here's the idea behind each:

##### Fix type validation for SELECT queries for a list of values

When we query for something using the `ANY`/`ALL` comparators, we're passing a list of values. Hence, we don't want to validate that list of values against its datatype: we want to validate each element of that list.

This means that the `escape` function now has to be aware of the comparators to know if it should directly validate the value or the elements of that array value. I've chosen to do that not by passing the comparator for the `escape` function to do the necessary logic, but by doing that logic before and just passing a boolean that's relevant for `escape`. But I don't have strong views on this, so if this goes against the convention, I'm happy to adjust.

##### Fix type validation for SELECT queries with the LIKE comparator

When we use the `LIKE` comparator (or the `ILIKE` alternative) we're doing a fuzzy search. So it's kind of unfair to ask for valid UUIDs here (if we're searching using an incomplete value, then it won't be a valid UUID - if it's a valid UUID, then it's not incomplete and we could just use `=` instead).

Strings are great for fuzzy searches. So the best option I could think of was to allow values to be strings when using `LIKE`/`ILIKE`. By moving that additional validation to inside the datatype's validation, we can whitelist the datatypes that we want to have this fallback. So with this commit, it's just enabled for UUIDs (are there other datatypes where it makes sense to apply this fallback as well?).

---

Do these suggestions go in line with the goals for the project? Also, am I overlooking any other parts of the code that should be impacted as well?

janmeier:
Both fixes seem reasonable to me, and I can't see any immediate problems - Go ahead and add a couple of tests and this should be good to go

rplopes:
Great, I'll move forward with this, then.

rplopes:
I've reordered the changes and added new tests for the UUID validators. The validations for `ANY`/`ALL` and `LIKE` comparators are also tested by existing tests that failed with my initial submission.

janmeier:
Thanks :)!

There are likely to be a few more cases where validations will fail because they are trying to validate partial / search input values. But since this feature is opt-in we'll take those as they crop up

rplopes:
Great :). Happy to help if that happens, and also hoping the strategies used in this PR might be repeatable for those potential issues elsewhere.

PR #5725, opened 2016-04-10
Bug: Binding dates via bind errored

kjones1876:

### Pull Request check-list

### Description of change

Bug: Binding dates via bind errored
Issue: SqlString.escape function params had changed, the usage of this function had not
Fixed by: Updating the usage of SqlString.escape to follow the new paramater order

### DIFF
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -416,7 +416,7 @@ AbstractQuery.formatBindParameters = function(sql, values, dialect, replacementF
     } else {
       replacementFunc = function(match, key, values, timeZone, dialect, options) {
         if (values[key] !== undefined) {
-          return SqlString.escape(values[key], false, timeZone, dialect);
+          return SqlString.escape(values[key], timeZone, dialect);
         }
         return undefined;
       };


janmeier:
Please show a test case where this would precisely fail

kjones1876:
@janmeier 
```
      var startAt = moment();
      var endAt = moment();
      var question = "This is my question";
      return queryInterface.sequelize.query(`
        insert into
          questions
          (createdAt, text, startAt, endAt)
        values
          (NOW(), $question, $startAt, $endAt)
      `, {
        plain: true,
        raw: true,
        type: Sequelize.QueryTypes.INSERT,
        bind: {
          question: question,
          startAt: startTime.toDate(),
          endAt: startTime.clone().add(1, 'day').toDate()
        }
      });
```

janmeier:
`SqlString.escape` does not take a second boolean parameter https://github.com/sequelize/sequelize/blob/master/lib/sql-string.js#L15

janmeier:
Seems to work fine for me on latest master. Going to close this, please comment / reopen with a failing test case

PR #5750, opened 2016-04-14
MSSQL describeTableQuery fix

alexahn:
The old describeTableQuery sql statement generator does not work correctly with schemas. The schema restriction gets added at the end via:
```
    if (schema) {
      sql += 'AND t.TABLE_SCHEMA =' + wrapSingleQuote(schema);
    }
```
Consequently, the results for the describeTableQuery will mix data for tables that have the same name and different schemas.


janmeier:
Thanks :)

Issue #4105, opened 2015-07-13
Transactions, Tedious and Connection Pools


awolden:
I have run into an issue making concurrent model calls while in a transaction when using the MSSQL dialect (tedious).

The Error I get: 

    Unhandled rejection SequelizeDatabaseError: Requests can only be made in the LoggedIn state, not the SentClientRequest state
        at Query.formatError (/home/woldena/Desktop/testapp/node_modules/sequelize/lib/dialects/mssql/query.js:215:10)
        at Request.userCallback (/home/woldena/Desktop/testapp/node_modules/sequelize/lib/dialects/mssql/query.js:46:25)
        at Request.callback (/home/woldena/Desktop/testapp/node_modules/tedious/lib/request.js:30:27)
        at Connection.makeRequest (/home/woldena/Desktop/testapp/node_modules/tedious/lib/connection.js:1142:22)
        at Connection.commitTransaction (/home/woldena/Desktop/testapp/node_modules/tedious/lib/connection.js:1048:17)
        at /home/woldena/Desktop/testapp/node_modules/sequelize/lib/dialects/mssql/query.js:44:25
        at tryCatcher (/home/woldena/Desktop/testapp/node_modules/sequelize/node_modules/bluebird/js/main/util.js:26:23)
        at Promise._resolveFromResolver (/home/woldena/Desktop/testapp/node_modules/sequelize/node_modules/bluebird/js/main/promise.js:462:31)
        at new Promise (/home/woldena/Desktop/testapp/node_modules/sequelize/node_modules/bluebird/js/main/promise.js:69:37)
        at Query.run (/home/woldena/Desktop/testapp/node_modules/sequelize/lib/dialects/mssql/query.js:33:17)
        at /home/woldena/Desktop/testapp/node_modules/sequelize/lib/sequelize.js:713:18
        at tryCatcher (/home/woldena/Desktop/testapp/node_modules/sequelize/node_modules/bluebird/js/main/util.js:26:23)
        at Promise._settlePromiseFromHandler (/home/woldena/Desktop/testapp/node_modules/sequelize/node_modules/bluebird/js/main/promise.js:489:31)
        at Promise._settlePromiseAt (/home/woldena/Desktop/testapp/node_modules/sequelize/node_modules/bluebird/js/main/promise.js:565:18)
        at Promise._settlePromiseAtPostResolution (/home/woldena/Desktop/testapp/node_modules/sequelize/node_modules/bluebird/js/main/promise.js:244:10)
[...]

PR #5752, opened 2016-04-14
Add query queueing to MSSQL connections

kjvalencik:
### Pull Request check-list

### Description of change

Resolves https://github.com/sequelize/sequelize/issues/4105. Adds a simple queuing mechanism to tedious connections via a resource lock. This will add a small amount of overhead to all requests but will prevent the possibility of interleaving requests on a single connection, especially within a transaction.

Currently broken:

```js
sequelize.transaction(transaction => {
  return Promise.all([1, 2, 3].map(id => Model.findById(id, { transaction })));
});
```

Even though these will still *not* happen concurrently, it will no longer result in error.

### Operation

The queue leverages Bluebird promises to effectively created a linked list queue with each request containing a reference to resolve the next dependent promise. Bluebird `disposer` will ensure that the next operation will begin only *after* the current promise chain has been fulfilled.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
 # Future
 - [FIXED] Fix defaultValues getting overwritten on build
+- [FIXED] Queue queries against tedious connections
 
 # 3.21.0
 - [FIXED] Confirmed that values modified in validation hooks are preserved [#3534](https://github.com/sequelize/sequelize/issues/3534)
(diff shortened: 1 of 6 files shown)


janmeier:
Nice idea to use resolvers to build the queue!

Thanks for the well-documented PR :). If you add an integration test along the lines of `expect(your example code).not.to.be.rejected` I think this is good to go.

kjvalencik:
@janmeier Thanks for the feedback. Added an integration test and ran before / after my changes. Thanks!

janmeier:
Sweet, thanks for the good work! :)

Issue #1679, opened 2014-04-23
Need for Postgres table inheritance


alekbarszczewski:
It would be nice if Sequelize supported Postgres table inheritance.
It looks like this on SQL side (found this example in google):

```
CREATE TABLE post (id integer, title varchar(100), content text);
CREATE TABLE post_without_tags () INHERITS (post);
CREATE TABLE post_with_tags (tags text[]) INHERITS (post);
```

So it is as simple as adding `INHERITS (tableName)` after `CREATE TABLE` statement.
On Sequelize side to could look like this:

```
...
PostWithTags.inherits(Post)
PostWithoutTags.inherits(Post)
...
```

This functionality could be even emulated for other dialects/databases - `inherit` method could just copy attributes from `parent` to `child` model.

Here is documentation on Postgres table inheritance:
http://www.postgresql.org/docs/9.3/static/ddl-inherit.html

thecaddy:
I have started using inherited tables and I took some time to write a plugin to create models easily assuming you have already set up your inherited tables and you want your models to inherit.

Below is a snip-it of the syntax I came up with, you can view the plugin and example gist here:
https://gist.github.com/thecaddy/b72a062b62b835fa06a8

```
export default function(sequelize, DataTypes) {
  const ChildModel = inheritFrom(sequelize, DataTypes)('childModel', inherittedModel, {
    // define child model specific properties here
      id: Sequelize.INTEGER,
      some_model_id: Sequelize.INTEGER
  }, {
    hooks: {
      beforeDestroy(instance, options) {
        return instance.sequelize.models.someModel({
          where: {
            id: instance.some_model_id
          }
        })
    },
    classMethods: {
      associate: function (models) {
        const {someModel} = models;
        this.belongsTo(someModel)
      }
    },

    instanceMethods: {
      anInstanceMethod(){}
    },
  });
  return ChildModel;
};
```

felixfbecker:
If Instances will be instances of Models in the future it would make sense to check the prototype chain in of the Model in Model.init() for other Models and for Model.inherits() to set up prototype inheritance.

PR #5757, opened 2016-04-16
Add Postgres inheritance

fixe:
Concept only. Refs #1679.

@mickhansen, @janmeier is this something that you guys are willing to merge?

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -96,7 +96,7 @@ var QueryGenerator = {
     }
 
     if (!!options.inherits) {
-      values.parameters += ' INHERITS (' + options.inherits.tableName + ')';
+      values.parameters += ' INHERITS (' + this.quoteTable(options.inherits.tableName) + ')';
     }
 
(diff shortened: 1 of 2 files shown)


fixe:
@mickhansen, @janmeier is this something that you guys are willing to merge?

janmeier:
Yep, this sounds like a reasonable approach. We would probably need to support the `ONLY` keyword in the relevant methods to exclude child tables

mickhansen:
Application-level attribute inheritance for other dialects and real `INHERITS` for postgres.
That might make sense, but doesn't postgres inheritance actually support selecting parent table to get all children (with shared attributes)?

janmeier:
Yeah without the `ONLY` keyword it willl return rows from both parent and child tables - but since we always qualify the attributes in selects, it shouldn't be a problem attribute wise. 

Of course the behaviour would be different between postgres and other dialects then, which would not be great :(

mickhansen:
Having `only` be the default and supported a children select with joins for other dialects could work in the future.

janmeier:

PR #5758, opened 2016-04-16
fix example

lucasBertola:


docs/docs/getting-started.md
```diff
@@ -91,14 +91,14 @@ Basically a promise represents a value which will be present at some point - "I
 // DON'T DO THIS
 user = User.findOne()
 
-console.log(user.name);
+console.log(user.firstName);
```
janmeier: While we're at it, could you change the examples to use `.get('firstName')` instead ?

lucasBertola: done :)

janmeier:
Sweet, thanks :)

PR #5772, opened 2016-04-21
defaultScope can be used with complex function.

hanjukim:
### Pull Request check-list

- [Y] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [N] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [Y] Have you added new tests to prevent regressions?
- [Not needed] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [Y] Have you added an entry under `Future` in the changelog?

### Description of change

Before this PR defaultScope can be used with plain object only, while scopes can be used with complex function. If complex function parameter is optional, defaultScope as function will give comfort.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 - [FIXED] Fix defaultValues getting overwritten on build
 - [FIXED] Queue queries against tedious connections
 - [ADDED] Enable type validation for all queries
+- [ADDED] `defaultScope` can be a complex function
 
 # 3.21.0
 - [FIXED] Confirmed that values modified in validation hooks are preserved [#3534](https://github.com/sequelize/sequelize/issues/3534)
--- a/lib/model.js
+++ b/lib/model.js
@@ -719,7 +719,8 @@ Model.prototype.init = function(modelManager) {
 
   findAutoIncrementField.call(this);
 
-  this.$scope = this.options.defaultScope;
+  this.$scope = _.isFunction(this.options.defaultScope) ? this.options.defaultScope() : this.options.defaultScope;
+  this.options.scopes.defaultScope = this.options.defaultScope;
 
   if (_.isPlainObject(this.$scope)) {
     conformOptions(this.$scope, this);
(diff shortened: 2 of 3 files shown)


janmeier:
While I really appreciate the time you put into making a proper PR, I'm not sure this is something we'd want to support.

In my mind, a default scope is something that is _just there_ - if you have to invoke it and pass parameters to it, why not just add a regular scope?

```js
let scopeFn = function (name) {
  var options = {
    include: [{model: Company }]
  };

  if (name) options.where = { name: name };
  return options;
};

var User = current.define('user', {
  name: DataTypes.STRING,
}, {
  defaultScope: scopeFn(),
  scopes: {
  	foo: scopeFn
  }
});
```

hanjukim:
@janmeier Your point is right. My point is I want something to be just there with defaultScope without adding a extra scope.

hendrul:
Well the issue for me comes when I want to make a defaultScope from other scopes I need a function like this:
```javascript
//Currency.js
{ 
 //...
defaultScope: function() {
  var self = models.Currency.options.scopes;
  return {
    include: self.includeExchanges().include.concat( self.includeSystemDefaults().include )
  }
},
scopes: {
  includeExchanges: function(where) {
    return {
      include: [
        {
          as: 'exchanges',
          model: models.CurrencyExchange,
          required: false,
          where: where||{}
        }
      ]
    }
  },
  includeSystemDefaults: function() {
    return {
      include: [
        {
          as: 'systemDefaults',
          model: models.SysCurrency,
          required: true
        }
      ]
    }
  }
},
```
Where by the way I need to manually merge the includes because the scope's merging mecanism does not support that case.

hendrul:
Ok, I did the following, use functions as defaultScope and after defining all models I run it's associations and replace each defaultScope's functions with the result of calling it.

jschneier:
I know this was closed but I have a use-case I'd like to mention:

If I have something that has an expiration time like a token I would like to default to not being able to view that token which requires a scope such as `{ expiresAt: { [Op.gt]: Date.now() } }`. Since that changes based on the value of `Date.now` it must be a function.

Is there a supported way to handle this use case? It's very dangerous that code might fetch a token that has already been invalidated.

hendrul:
What about the solution I give in the previous post?

jschneier:
That is still only run once I believe, I need it to vary every time `.defaultScope` is invoked (due to the `Date.now()`) which I assume is on every query.

Issue #4711, opened 2015-10-22
How to exiting gracefully?


tdzl2003:
Before exiting, I want to wait all queries finished, then connection destroyed gracefully. Like calling 'end' method in `mysql` package.

Is there any way to do this in sequelize? I found  `destroy` `$disconnect` method in connectionManager, but I didn't find the way to call them from a `Sequelize` instance.

mickhansen:
There is `sequelize.close()`, not sure how graceful that is though.

janmeier:
It should be pretty gracefully, telling the pool to empty, which in turns tells each connection in the pool to disconnect

tdzl2003:
Currently .close didn't return a Promise, so I think it's clearly it doesn't wait queries to be finished.

```
Sequelize.prototype.close = function () {
  this.connectionManager.close();
};
```

I found `destroy` `$disconnect` method in connectionManager, but I didn't find a way to use it via sequelize instance.

janmeier:
`.close` calls `connectionmanager.close` https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/connection-manager.js, so yes you found it ;). It shuts down connections gracefully, but does not allow you to await it no. In any case, the process will not exit before connections have terminated, unless you call `process.exit`

tdzl2003:
Thanks for answering!

tedeh:
So the succint answer is `sequelize.connectionManager.close().then(() => console.log('shut down gracefully'));` for anyone else coming here.

vajahath:
I guess as of now you can use `sequelize.close().then..`

PR #5776, opened 2016-04-21
sequelize.close return a promise

llafuente:
sequelize.close return undefined. That's... not useful.

This PR aims to add a Promise to close and not just a bad promise, like ConnectionManager.close.
Also fix pgsql disconnect promise that was sync.

This PR cannot land ATM!
This must land first https://github.com/coopernurse/node-pool/pull/141 (because there is another bug there that block this fix)

After that I will update package.json

I can't find a way to do a test for this in the current tests folder. So I just attached here.

``` js
// sudo -u postgres psql -c "drop database sequelize_test;"
// sudo -u postgres psql -c "create database sequelize_test;"

var Sequelize = require('./index.js');
var sequelize = new Sequelize('sequelize_test', 'postgres', 'postgres', {
  host: 'localhost',
  //port: 9092, // vagrant port
  dialect: 'postgres',
});
// just something to sync
var datadef = sequelize.define('just_create_table', {
  created_date: Sequelize.DATE
});

sequelize.sync().then(function() {
  console.log("handles before:", process._getActiveHandles().length);
  return sequelize.close().then(function() {
    console.log("handles after:", process._getActiveHandles().length);
  })
})
.then(function() {
  console.log("handles after:", process._getActiveHandles().length);
});
```

Before this patch I see: 4 4 4
After: 4 2 2

Related issues:
https://github.com/sequelize/sequelize/issues/4711 @tdzl2003

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -64,21 +64,26 @@ ConnectionManager.prototype.refreshTypeParser = function(dataTypes) {
 
 ConnectionManager.prototype.onProcessExit = function() {
   var self = this;
-
-  if (this.pool) {
-    this.pool.drain(function() {
-      self.pool.destroyAllNow();
-    });
-  }
+  return new Promise(function (resolve, reject) {
+    if (self.pool) {
+      self.pool.drain(function() {
+        self.pool.destroyAllNow(function () {
+          resolve();
+        });
+      });
+    } else {
+      resolve();
+    }
+  });
 };
 
 ConnectionManager.prototype.close = function () {
-  this.onProcessExit();
   process.removeListener('exit', this.onProcessExit); // Remove the listener, so all references to this instance can be garbage collected.
 
   this.getConnection = function () {
     return Promise.reject(new Error('ConnectionManager.getConnection was called after the connection manager was closed!'));
   };
+  return this.onProcessExit();
 };
 
 // This cannot happen in the constructor because the user can specify a min. number of connections to have in the pool
@@ -95,8 +100,8 @@ ConnectionManager.prototype.initPools = function () {
           callback(err, connection); // For some reason this is needed, else generic-pool things err is a connection or some shit
         });
       },
-      destroy: function(connection) {
-        self.$disconnect(connection);
+      destroy: function(connection, cb) {
+        self.$disconnect(connection, cb);
         return null;
       },
       max: config.pool.max,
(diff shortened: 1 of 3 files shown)


lib/dialects/abstract/connection-manager.js
```diff
@@ -138,12 +143,21 @@ ConnectionManager.prototype.initPools = function () {
...
-    destroy: function(connection) {
-      return self.pool[connection.queryType].destroy(connection);
+    destroy: function(connection, cb) {
+      return self.pool[connection.queryType].destroy(connection, cb);
     },
-    destroyAllNow: function() {
-      self.pool.read.destroyAllNow();
-      self.pool.write.destroyAllNow();
+    destroyAllNow: function(cb) {
+      var todo = 2;
```
janmeier: `cb = _.after(2, cb);`

janmeier:
LGTM.

With regards to testing, its not easy to verify, since we likely can't hardcode the number of active handles. I'll let this one pass without a test ;)

llafuente:
I will update the PR when node-pool merge it's change.

Do you like another commit or squash?

janmeier:
Squash please :)

lirantal:
@llafuente @janmeier what's holding this PR from merging in? to squash you can do it from the GitHub interface now so surely not that.

felixfbecker:
There are merge conflicts

lirantal:
@felixfbecker I can pull locally and push a PR to fix those, is that the only thing?

felixfbecker:
what about

> This PR cannot land ATM!
> This must land first coopernurse/node-pool#141 (because there is another bug there that block  this fix)

llafuente:
@lirantal
node-pool has to be fixed first. destroyAllNow and destroy have no callback

janmeier:
@llafuente Have you tried integrating the new version of node pool into sequelize? @sushantdhiman already tried, but had some problems, see https://github.com/coopernurse/node-pool/issues/161

sushantdhiman:
@llafuente if you want a base to work with, Please checkout this branch https://github.com/sequelize/sequelize/tree/generic-pool-update 

Cases where pool resource creation fails needs to be fixed, everything else should work

lirantal:
Got it.
That node-pool open issue seems to be stale for quite a while, I wonder why and maybe we should consider some alternative?

sandfox:
@lirantal I do plan to merge it in, I just haven't much time to do so as most of my efforts are in v3 of generic-pool, and working with the old v2 codebase is a lot less fun / easy. I'm looking at it at the moment so there should be some movement or at least feedback on this today.

lirantal:
@sandfox great thank you, much appreciated.

sandfox:
v2.5.3 of  generic-pool is now released and contains the above talked about code/changes!

lirantal:
Great

PR #5781, opened 2016-04-22
adds a sscce template where code can easily be copied and run inside …

mickhansen:


package.json
```diff
@@ -129,7 +129,11 @@
     "cover-all": "npm run cover-mysql && npm run cover-postgres && npm run cover-postgres-native && npm run cover-mssql && npm run cover-sqlite && npm run cover-mariadb && npm run merge-coverage",
     "merge-coverage": "rm -rf coverage && mkdir coverage && ./node_modules/.bin/lcov-result-merger 'coverage-*/lcov.info' 'coverage/lcov.info'",
     "codeclimate-send": "npm install -g codeclimate-test-reporter && CODECLIMATE_REPO_TOKEN=ce835a510bbf423a5ab5400a9bdcc2ec2d189d840b31657c6ee7cb9916b161d6 codeclimate-test-reporter < coverage/lcov.info",
-    "codeclimate": "npm run cover-all && npm run codeclimate-send"
+    "codeclimate": "npm run cover-all && npm run codeclimate-send",
+    "sscce": "docker-compose run sequelize /bin/sh -c \"node sscce.js\"",
```
janmeier: Hmm, not sure I love the idea of people needing to have docker installed ..

mickhansen: This is as much for us, to copy paste. And we refer to docker in our general contribution guide aswell.
Without docker there's no real reason to have a template that uses test/support, because you'd need to configure sequelize everytime in any case.

janmeier: True dat!

janmeier:
Perhaps add a mention of this to the PR / issue template?

Issue #3302, opened 2015-03-07
"Error: Please install postgres package manually" after updating node and npm


jakemmarsh:
I updated my local versions of Node (now v0.12.0) and npm (now v2.7.0). I also updated all my globally-installed packages accordingly.

My project is using `sequelize` v2.0.4, `pg` v4.3.0, `pg-hstore` v2.3.1, and `pg-native` v1.8.0. None of my previously working Sequelize/connection logic has been changed. Now when starting my server I get the following error/stacktrace:

```
Starting child process with 'node server.js'
Error: Please install postgres package manually
    at new ConnectionManager (/Users/jakemarsh/Repos/monolist-api/node_modules/sequelize/lib/dialects/postgres/connection-manager.js:16:11)
    at new PostgresDialect (/Users/jakemarsh/Repos/monolist-api/node_modules/sequelize/lib/dialects/postgres/index.js:12:28)
    at new module.exports.Sequelize (/Users/jakemarsh/Repos/monolist-api/node_modules/sequelize/lib/sequelize.js:185:22)
    at Object.<anonymous> (/Users/jakemarsh/Repos/monolist-api/api/models/index.js:18:13)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
/Users/jakemarsh/Repos/monolist-api/node_modules/sequelize/lib/sequelize.js:188
      throw new Error('The dialect ' + this.getDialect() + ' is not supported.
            ^
Error: The dialect postgres is not supported. (Error: Please install postgres package manually)
    at new module.exports.Sequelize (/Users/jakemarsh/Repos/monolist-api/node_modules/sequelize/lib/sequelize.js:188:13)
    at Object.<anonymous> (/Users/jakemarsh/Repos/monolist-api/api/models/index.js:18:13)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/jakemarsh/Repos/monolist-api/api/routes/auth.js:7:16)
    at Module._compile (module.js:460:26)
Program node server.js exited with code 1
```

mickhansen:
[...]

PR #5784, opened 2016-04-22
ConnectionManager: Re-throw initialization errors

yoz:
ConnectionManager: Re-throw initialization errors unless they are MODULE_NOT_FOUND. This prevents such errors from being misleadingly overridden.

As described in https://github.com/sequelize/sequelize/issues/3302#issuecomment-199238449


mickhansen:
Thank you!

PR #5786, opened 2016-04-23
refactor(testing): Re-enable support shim, and extend to test options…

janmeier:
This build on https://github.com/sequelize/sequelize/pull/4541 by re-enabling support shim, and also checking that options are not modified. In order to check that we clone the options object before each call, which unfortunately seems to up the time it takes to run tests a bit :( But on the bright side I did find a couple of places where options were being modified / not passed down correctly.

I've also replaced `optclone` with a method that avoids cloning anything of type object thats not a plain object. This handles all types of sequelize objects (models, instances etc.) and also `Buffer`, which we had a special case for previously

Ping @overlookmotel

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -424,7 +424,7 @@ BelongsToMany.prototype.injectGetter = function(obj) {
   var association = this;
 
   obj[this.accessors.get] = function(options) {
-    options = association.target.__optClone(options) || {};
+    options = Utils.cloneDeep(options) || {};
 
     var instance = this
       , through = association.through
@@ -486,7 +486,7 @@ BelongsToMany.prototype.injectGetter = function(obj) {
     var model = association.target
       , sequelize = model.sequelize;
 
-    options = association.target.__optClone(options) || {};
+    options = Utils.cloneDeep(options);
     options.attributes = [
       [sequelize.fn('COUNT', sequelize.col([association.target.name, model.primaryKeyAttribute].join('.'))), 'count']
     ];
@@ -506,10 +506,8 @@ BelongsToMany.prototype.injectGetter = function(obj) {
       instances = [instances];
     }
 
-    options = options || {};
-    options.scope = false;
-
-    _.defaults(options, {
+    options = _.assign({}, options, {
+      scope: false,
       raw: true
     });
 
@@ -640,7 +638,7 @@ BelongsToMany.prototype.injectSetter = function(obj) {
     // If newInstances is null or undefined, no-op
     if (!newInstances) return Utils.Promise.resolve();
 
-    additionalAttributes = additionalAttributes || {};
+    additionalAttributes = _.clone(additionalAttributes) || {};
 
     var instance = this
       , defaultAttributes = _.omit(additionalAttributes, ['transaction', 'hooks', 'individualHooks', 'ignoreDuplicates', 'validate', 'fields', 'logging'])
--- a/lib/model.js
+++ b/lib/model.js
@@ -8,7 +8,6 @@ var Utils = require('./utils')
   , HasMany = require('./associations/has-many')
   , DataTypes = require('./data-types')
   , Util = require('util')
-  , Transaction = require('./transaction')
   , Promise = require('./promise')
   , QueryTypes = require('./query-types')
   , Hooks = require('./hooks')
@@ -307,27 +306,6 @@ function conformInclude(include, self) {
 
 Model.$conformInclude = conformInclude;
 
-var optClone = Model.prototype.__optClone = Model.prototype.$optClone = function(options) {
-  options = options || {};
-  return Utils.cloneDeep(options, function(elem) {
-    // The InstanceFactories used for include are pass by ref, so don't clone them.
-    if (elem && typeof elem === 'object' &&
-      (
-        elem._isSequelizeMethod ||
-        elem instanceof Model ||
-        elem instanceof Instance ||
-        elem instanceof Transaction ||
-        elem instanceof Association
-      )
-    ) {
-      return elem;
-    }
-
-    // Otherwise return undefined, meaning, 'handle this lodash'
-    return undefined;
-  });
-};
-
 var expandIncludeAllElement = function(includes, include) {
   // check 'all' attribute provided is valid
   var all = include.all;
@@ -413,7 +391,7 @@ var expandIncludeAllElement = function(includes, include) {
       used.push(parent);
 
       // include this model
-      var thisInclude = optClone(include);
+      var thisInclude = Utils.cloneDeep(include);
       thisInclude.model = model;
       if (as) {
         thisInclude.as = as;
@@ -990,9 +968,8 @@ Model.prototype.removeAttribute = function(attribute) {
  * @return {Promise<this>}
  */
 Model.prototype.sync = function(options) {
-  options = options || {};
+  options = _.extend({}, this.options, options);
   options.hooks = options.hooks === undefined ? true : !!options.hooks;
-  options = Utils._.extend({}, this.options, options);
 
   var self = this
     , attributes = this.tableAttributes;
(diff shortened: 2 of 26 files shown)


lib/utils.js
```diff
@@ -87,20 +87,23 @@ var Utils = module.exports = {
...
+    obj = obj || {};
     return _.cloneDeepWith(obj, function (elem) {
-      // Do not try to customize cloning of plain objects and strings
+      // Do not try to customize cloning of arrays or POJOs
       if (Array.isArray(elem) || _.isPlainObject(elem)) {
         return undefined;
       }
+
+      // Don't clone stuff that's an object, but not a plain one - fx example sequelize models and instances
+      if (typeof elem === 'object' && !_.isPlainObject(elem)) {
```
mickhansen: The second part of this if statement should not be needed since it should automatically be hit by the array/pojo check above.

overlookmotel:
@janmeier This is brilliant! I had started work on doing exactly the same thing, as I wanted to extend the shims to also check that CLS context is always maintained. But I'd got bogged down with trying to fix all the places where `options` was not being cloned and with the problem of cloning options without cloning Models etc etc, and hadn't got very far.

Sounds like you solved all the problems!

I haven't actually looked at your code yet, but will do as soon as I get the chance.

test/supportShim.js
```diff
@@ -14,6 +15,10 @@ module.exports = function(Sequelize) {
   shimAll(Sequelize.Model.prototype);
   shimAll(Sequelize.Instance.prototype);
   shimAll(QueryInterface.prototype);
+  shimAll(Sequelize.Association.prototype);
+   _.forIn(Sequelize.Association, function(Association) {
```
overlookmotel: An extra space has crept in at start of this line. And after all the other ones you carefully removed!

overlookmotel:
Had a quick look.

I think I may have an idea how to avoid changing the tests for logging, without needing `Sequelize.dontShim`.

Also, if you don't mind, I'd like to check through all the changes from e.g. `_.defaults(options, { ... })` to `_.assign({}, options, { ... })` to make sure the logic hasn't been changed subtlely. That isn't to question your chops - just with that many changes, it'd be a superhuman feat to not let a mistake to creep in *somewhere*!

How swiftly do you need to get this merged? I'm pretty tied up this week and it'll take time to go through it closely. Can it wait until next week?

janmeier:
No rush, it can wait - I'd rather take the time so we don't get any stupid regressions. Changes to how options objects are treated could definitely create subtle and hard to spot bugs

lib/associations/belongs-to-many.js
```diff
@@ -506,10 +506,8 @@ BelongsToMany.prototype.injectGetter = function(obj) {
       instances = [instances];
     }
 
-    options = options || {};
-    options.scope = false;
-
-    _.defaults(options, {
+    options = _.assign({}, options, {
+      scope: false,
```
overlookmotel: Think this should be: `options = _.assign({raw: true}, options, {scope: false})`. `raw: true` is a default in original, new code has it as always set as `true`

lib/model.js
```diff
@@ -990,9 +968,8 @@ Model.prototype.removeAttribute = function(attribute) {
  * @return {Promise<this>}
  */
 Model.prototype.sync = function(options) {
-  options = options || {};
+  options = _.extend({}, this.options, options);
```
overlookmotel: Is it possible that `this.options.hooks` could be defined? If so, this needs to be: `options = _.extend({}, this.options, options, {hooks: options.hooks === undefined ? true : !!options.hooks})` to maintain exact same behaviour.

lib/model.js
```diff
@@ -1617,9 +1595,8 @@ Model.prototype.aggregate = function(attribute, aggregateFunction, options) {
  * @return {Promise<Integer>}
  */
 Model.prototype.count = function(options) {
-  options = Utils._.clone(options || {});
-
-  _.defaults(options, { hooks: true });
+  options = Utils.cloneDeep(options);
+  options.hooks = true;
```
overlookmotel: This line needs to remain `_.defaults(options, { hooks: true });` to keep same behaviour. `hooks: true` is a default, not compulsory.

lib/query-interface.js
```diff
@@ -49,8 +49,7 @@ QueryInterface.prototype.showAllSchemas = function(options) {
 
   options = _.assign({}, options, {
     raw: true,
-    type: this.sequelize.QueryTypes.SELECT,
-    logging: false
+    type: this.sequelize.QueryTypes.SELECT
```
overlookmotel: Removal of `logging: false` is intentional here?

lib/sequelize.js
```diff
@@ -739,7 +739,7 @@ Sequelize.prototype.query = function(sql, options) {
 
   var self = this;
 
-  options = options || {};
+  options = _.assign({}, this.options.query, options);
```
overlookmotel: What does addition of `this.options.query` do?

lib/utils.js
overlookmotel: Agree with @mickhansen here.

lib/utils.js
```diff
@@ -87,20 +87,23 @@ var Utils = module.exports = {
...
+        return elem;
+      }
+
       // Preserve special data-types like `fn` across clones. _.get() is used for checking up the prototype chain
       if (elem && typeof elem.clone === 'function') {
         return elem.clone();
       }
-      // Unfortunately, lodash.cloneDeep doesn't preserve Buffer.isBuffer, which we have to rely on for binary data
-      if (Buffer.isBuffer(elem)) { return elem; }
```
overlookmotel: `options` never contains `Buffer`s?

overlookmotel:
Hi @janmeier.

I've checked through the code of this PR and *think* I found a couple of sneaky bugs that'd crept in in the object cloning. Have left line notes for these and a few other queries. Would it save you time if I submitted those amendments as a PR on your PR branch?

I'm looking into whether it's possible to remove `Sequelize.dontShim` by making the shim use the proper `logging` attribute of `sequelize.options` rather than it's own default. I think that'd be a cleaner solution. Will get onto that in next few days.

lib/utils.js
overlookmotel: Sorry, on 2nd thoughts, I'm not sure. The 2nd check (line 99) is to copy by reference any objects which are *not* plain objects, which is a different condition from `if (Array.isArray(elem) || _.isPlainObject(elem)) ...` which is for plain objects only.

janmeier: I don't think it should be needed - If it is a POJO it will hit the first if, so the only type of object we will see here are non-plain

lib/utils.js
janmeier: They do, but it will be caught by `typeof  == 'object'` above

lib/associations/belongs-to-many.js
janmeier: Yep

lib/query-interface.js
janmeier: Yep, I don't see why we should force it to false

lib/sequelize.js
janmeier: No, I wouldn't think so

janmeier:
Thanks for looking through this @overlookmotel - A PR would be very welcomed. And I think your approach of using  sequelize.options.logging` in the shim would work well

lib/utils.js
overlookmotel: Ah ha yes I see.

lib/utils.js
overlookmotel: Yes of course, you and @mickhansen are right!

overlookmotel:
Hi @janmeier.

I've submitted PR https://github.com/sequelize/sequelize/pull/5836 that builds on your PR. It includes the changes I suggest in line notes above, plus the shim now respects `sequelize.options.logging` so was able to remove the `Sequelize.dontShim` code. Plus some other bits and bobs...

Let's see if the tests pass!

PR #5790, opened 2016-04-25
Add context and custom options to deep creation

Verdier:
### Pull Request check-list

### Description of change

This PR adds the ability to pass custom options when creating nested records using `Model.create` and `options.instance`. It also add `parentRecord` to nested options to have a context.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -634,10 +634,15 @@ Instance.prototype.save = function(options) {
         var instance = self.get(include.as);
         if (!instance) return Promise.resolve();
 
-        return instance.save({
-          transaction: options.transaction,
-          logging: options.logging
-        }).then(function () {
+        var includeOptions =  _(self.$Model.$optClone(include))
+          .omit(['association'])
+          .defaults({
+            transaction: options.transaction,
+            logging: options.logging,
+            parentRecord: self
+          }).value();
+
+        return instance.save(includeOptions).then(function () {
           return self[include.association.accessors.set](instance, {save: false, logging: options.logging});
         });
       });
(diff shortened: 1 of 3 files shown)


lib/instance.js
```diff
@@ -634,10 +634,15 @@ Instance.prototype.save = function(options) {
...
 
-        return instance.save({
+        var includeOptions = _.defaults(include.options || {}, {
           transaction: options.transaction,
           logging: options.logging
-        }).then(function () {
+        });
+        _.assign(includeOptions, {
+          parentRecord: self
+        });
```
mickhansen: any reason to have both a defaults call and an assign call?

```
_.assign({}, include.options, {
  transaction: options.transaction,
  logging: options.logging,
  parentInstance: self
})
```

mickhansen: We could also consider having custom options simply be on `include`

mickhansen: And we should likely also inherit options from the overall create call to support plugins etc

Verdier: Yep, the `defaults` is because `transaction` and `logging` can be override, and `assign` because `parentInstance` not. Would you like to force the parent `transaction` and `logging` options?

What about inheriting options from the overall create? Which ones?

Verdier:
Moreover @mickhansen we cannot have custom options on the include declaration, otherwise we will have to maintains a `blacklist` of include's reserved options, to avoid collisions with create options...

mickhansen:
@Verdier That's the same for all option sets though.

lib/instance.js
mickhansen: All that don't conflict, the goal is that the user can send any set of options to any call and have access to those specific options anywhere in the call chain where they hook in.

Verdier:
Yes, but the consumer should care of that, he should not pass options which will collide with create options. But if we simply do:
```js
var includeOptions = this.$Model.$optClone(include);
```
there will probably be collisions now or in the future.

mickhansen:
@Verdier It's up to the user to make sure there are no collissions for options that Sequelize  expects, atleast that's how we do it in the rest of the system :)

Verdier:
Yep agree, it's why we can't use include directly. What would you propose?

mickhansen:
@Verdier Why not though? The collissions for include are the same for findAll, so the user has to take care in any case.

Verdier:
Do you prefer something like that:
```js
var includeOptions = _.defaults(this.$Model.$optClone(include), {
              transaction: options.transaction,
              logging: options.logging,
              parentRecord: self
});
```

Verdier:
Done @mickhansen.

mickhansen:
Looking good, would like @janmeier's input here though.

janmeier:
LGTM as well - Once you figure out why postgres is failing this should be good to go

janmeier:
@Verdier Needs a rebase :)

Verdier:
@janmeier done.

janmeier:
Still some tests failing for PG and mssql :(

Verdier:
Yep... I'll try to fixe them ^^

Verdier:
Everything is green @janmeier

Verdier:
Any feedback @janmeier or @mickhansen?

janmeier:
Thanks a bunch @Verdier ! :)

PR #5801, opened 2016-04-26
only check model keys for delete with limit

### LINKED ISSUE #5615: Method QueryInterface.bulkDelete no longer working when the model parameter is missing. (PostgreSQL)
When in a migration I'm trying to delete all the rows in one of my tables I get the following error
"TypeError: Cannot read property 'primaryKeys' of undefined".

My code looks like this 
```
up: function(queryInterface, Sequelize) {
	return queryInterface.bulkDelete('Foo');
}
```
I expect that all rows matching a where condition get deleted.

This happens when using sequelize 3.19.3 with PostgreSQL.

IMO the bug is caused by the changes in this PR #5443. 
More precisely, the exception occurs [here](https://github.com/sequelize/sequelize/pull/5443/files#diff-7e347c4738c97261969fbeb802a3c363R367) as `model` is undefined.

Americas:
### Pull Request check-list

### Description of change

deleteQuery assumes a model exist and tries to extract primary keys from it, even when this is an unnecessary step. This change prevents model.primaryKeys from being accessed unless options.limit is provided.

It also throws a more friendly error if model is not defined.

Fixes #5615

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -357,23 +357,28 @@ var QueryGenerator = {
       options.limit = 1;
     }
 
+    var replacements = {
+      table: tableName,
+      where: this.getWhereConditions(where, null, model, options),
+      limit: !!options.limit ? ' LIMIT ' + this.escape(options.limit) : ''
+    };
+
     if (options.limit) {
+      if (!model) {
+        throw new Error('Cannot LIMIT delete without a model.');
+      }
+
+      var pks = _.map(_.values(model.primaryKeys), function (pk) {
+        return this.quoteIdentifier((pk.field));
+      }.bind(this)).join(',');
+
+      replacements.primaryKeys = model.primaryKeyAttributes.length > 1 ? '(' + pks + ')' : pks;
+      replacements.primaryKeysSelection = pks;
+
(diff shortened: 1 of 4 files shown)


janmeier:
LGTM - mind adding a few unit tests? https://github.com/sequelize/sequelize/blob/master/test/unit/sql/delete.test.js#L46

lib/dialects/postgres/query-generator.js
```diff
@@ -357,24 +357,29 @@ var QueryGenerator = {
...
 
+    var replacements = {
+      table: tableName,
+      where: this.getWhereConditions(where, null, model, options),
+      limit: !!options.limit ? ' LIMIT ' + this.escape(options.limit) : ''
+    };
+
     if (options.limit) {
+      if (!model) {
+        throw new Error("Cannot LIMIT without a model.");
```
janmeier: `Cannot LIMIT delete` might be a bit clearer :)

Americas:
Yes, I will add tests. I just wanted to check the tests because it is giving me strange errors locally...

Americas:
test added

Americas:
@janmeier 2 builds failed, but I reckon the errors are unrelated. At least the sqlite error is, it's an I/O error.

janmeier:
Yeah, I've seen a lot of these io errors lately, not sure if its an issue with travis, or something we are doing ..

janmeier:
Thanks a lot :)

PR #5818, opened 2016-04-28
fixed a couple of commas and a semicolon on the validation docs

paolord:
### Description of change

Just added a couple of commas and a semicolon on the code sample for model definition validations. Even though it won't really make a difference when you try to run the code, I felt I should get it fixed. Hoping this is a good first PR.


janmeier:
Thanks :) - all contributions are welcomed!

PR #5823, opened 2016-04-29
Base Fix for hasMany issue  #4258

### LINKED ISSUE #4258: targetKey for hasMany and hasOne
I have two legacy tables in a 1:m association:
**Property**(id, statistical_id, name, ...), **VisitorStats**(id, statistical_id, total, ...)

The following works:

`VisitorStats.belongsTo(Property, { foreignKey: 'statistical_id', targetKey: 'statistical_id'} );`

Then, in the controller, I can query for VisitorStats and eager-load the Property for it.

However, I want to query the Property table and include the VisitorStats model. For this, I would have to do a hasMany association, like this:

`Property.hasMany(VisitorStats, { foreignKey: 'statistical_id' });`

**hasMany's doesn't support the targetKey parameter option, and automatically joins VisitorStats.statistical_id on Property.id when I query.** I expected the targetKey property to work like at a belongsTo association.

Is it possible to change the target key using hasMany? If not, how should I represent this 1:m association?

matthewhardern:
### Pull Request check-list

### Description of change

This change is to enable defining a foreign key on the source table for hasMany function, that then in turn allows using a key that is not the primary key of that table. It also falls back on primary key when no sourceKey is provided.

Closes #4258


janmeier:
Code looks good - But we need some tests as well :)

lib/associations/has-many.js
```diff
@@ -272,6 +276,8 @@ HasMany.prototype.get = function(instances, options) {
     instances = undefined;
   }
 
+  console.log(association);
```
mickhansen: ^

felixfbecker:
Thinking about this, we should then probably support both `sourceKey` and `targetKey` for `belongsToMany`. Which might get a bit confusing with `otherKey`. But there was an issue recently that would be solved by this.

janmeier:
Fixed in https://github.com/sequelize/sequelize/pull/7113

PR #5831, opened 2016-05-01
added update support for primary keys that are not autoincrement

atulhm:

### Description of change

I modified the primary key setter check to allow pks that aren't autoincrement to proceed with the set
I reviewed the test cases that were failing with the changes and updated the test cases.
THANK YOU for npm run test (it helped me validate the changes)

I believe this should update should be non-breaking.  I had originally allowed any pk to change, but that caused other problems.  After researching and by restricting autoincrement pks from being updated, the tests now pass.   

Not sure if this should be labeled enhancement or bug fix

NOTE: I have not modified the changelog yet...

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -331,8 +331,9 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
             return this;
           }
 
-          // If attempting to set primary key and primary key is already defined, return
-          if (this.Model._hasPrimaryKeys && originalValue && this.Model._isPrimaryKey(key)) {
+          // If attempting to set primary key and primary key is already defined
+          // and its autoIncrement, return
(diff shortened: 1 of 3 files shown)


janmeier:
LGTM, I think this is a bug-fix - @mickhansen ?

mickhansen:
No it's generally a security implementation, generally you don't want the user provided values to be able to set the primary key.

janmeier:
Yeah thats true actually - I misread the code and was thinking it could never hit the `raw` part for primary keys - but it does ;)

```js
user.set('id', 42, { raw: true });
user.changed('id', true);
return user.save();
```

atulhm:
LOL.  I misread it the same way which led me down a small rabbit hole and an unneeded PR.

PR #5834, opened 2016-05-02
Fixing associations doc

gtelljohann:
I'm not positive that this was what was intended originally, but as it currently is, this code snippet isn't valid javascript.  I was able to find no example of `through` taking anything other than a string, but it does seem like `attributes` and `where` can both go in `include`, so I assume that this is the right way.  This also makes sense given the context of the example.

### DIFF
--- a/docs/docs/associations.md
+++ b/docs/docs/associations.md
@@ -264,10 +264,9 @@ With Belongs-To-Many you can query based on **through** relation and select spec
 User.findAll({
   include: [{
     model: Project,
-      through: {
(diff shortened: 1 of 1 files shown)


mickhansen:
This is definitely possible. The way you changed `through` is only valid for defining attributes.

mickhansen:
Although i'll give you that the indentation is wonky as hell.
`include.through` does support where and is supposed to support attributes, but there's a bug with it currently. You can verify with the API docs

gtelljohann:
K, I'll double check with the API docs and update accordingly

PR #5835, opened 2016-05-02
refactor(testing): Re-enable support shim, and extend to test options

overlookmotel:
This PR extends @janmeier's PR https://github.com/sequelize/sequelize/pull/5786.

* The tests shim is re-enabled
* `options` objects passed into all public sequelize methods are checked to make sure they are not modified
* Various fixes to sequelize codebase to clone `options` objects before modifying them

@janmeier The changes in addition to what you'd already done:

* Includes the bugs I indicated with line notes on your PR https://github.com/sequelize/sequelize/pull/5786
* Global logging option (`sequelize.options.logging`) is maintained by the test shim, which allowed some of the workaround for that to be removed.
* Options checked for modification both synchronously (i.e. just after the call returns) and async (when the promise resolves).
* Options checked for modification on all calls to the sequelize public API, even when internally called from within another sequelize method.
* The above two revealed a few more places where options was getting modified, so these are now patched as well
* When comparing "old" and "new" options to see if modified, non-plain objects (e.g. Models) are compared by reference rather than deep compared (in line with how options is cloned). This may speed the tests up a bit.

I've left this as a whole string of commits so it can be more easily inspected, but I can squash before it gets merged.


overlookmotel:
Closed this PR and resubmitting as PR on master so Travis runs.

PR #5836, opened 2016-05-02
refactor(testing): Re-enable support shim, and extend to test options object not being modified v2

overlookmotel:
This PR extends @janmeier's PR https://github.com/sequelize/sequelize/pull/5786.

* The tests shim is re-enabled
* `options` objects passed into all public sequelize methods are checked to make sure they are not modified
* Various fixes to sequelize codebase to clone `options` objects before modifying them

@janmeier The changes in addition to what you'd already done:

* Includes the bugs I indicated with line notes on your PR https://github.com/sequelize/sequelize/pull/5786
* Global logging option (`sequelize.options.logging`) is maintained by the test shim, which allowed some of the workaround for that to be removed.
* Options checked for modification both synchronously (i.e. just after the call returns) and async (when the promise resolves).
* Options checked for modification on all calls to the sequelize public API, even when internally called from within another sequelize method.
* The above two revealed a few more places where options was getting modified, so these are now patched as well
* When comparing "old" and "new" options to see if modified, non-plain objects (e.g. Models) are compared by reference rather than deep compared (in line with how options is cloned). This may speed the tests up a bit.

I've left this as a whole string of commits so it can be more easily inspected, but I can squash before it gets merged.


overlookmotel:
Travis has ground to a halt for some reason...

janmeier:
@overlookmotel Travis is done now - seems to be a single place left where options are modified in postgres

overlookmotel:
Travis keeps failing randomly - in different places and different dialects each time, but always due to a test timing out. Is this normal? Or is the cause specific to this PR - maybe if it's slowed things down,

janmeier:
We do have periodic failures from time to time - as long as its just a single build were usually fine. But these changes have made builds take significantly longer .. :( 4-7 minutes, as opposed to 3-5 minutes per job now. I wouldn't have thought that cloning adds that amount of time, but apparently it does.

Right now all jobs are failing though - I merged a PR earlier today which used `$optClone` in a new place, so that needs to be changed to `Utils.cloneDeep`

overlookmotel:
OK, have rebased on master and replaced `$optClone()` with `Utils.cloneDeep()`. Let's see how it fares this time...

overlookmotel:
I increased the tests timeout to 30secs and hey presto Travis passes!

It does seem that the options cloning (or the re-introduction of the shim in general) is slowing things down a fair bit.

I think this needs further investigation before merging, but can we possibly conclude first whether the code I've built on top of @janmeier's PR is good in principle, aside from the concern about slowdown?

janmeier:
The code LGTM! :)

overlookmotel:
@janmeier Thanks!

But let me run a few benchmarks before merging. I want to check whether the slowdown is caused by all the options cloning in sequelize methods, or by the test shim itself. i.e. whether it's going to be slow in production, or just slow in the tests. I'd say don't merge for now until it's clear that these changes aren't massively impacting performance in production.

janmeier:
@overlookmotel Agreed

overlookmotel:
I've split this PR into 3 parts:

https://github.com/sequelize/sequelize/pull/5874 Clones options
https://github.com/sequelize/sequelize/pull/5836 (this PR) Tests for options cloning and re-enables the support shim
https://github.com/sequelize/sequelize/pull/5875 Removes tests which are now redundant

Each PR depends on the one before so would need to be merged in the above order.

I've split it for 2 reasons:
1. To make a clearer commit history
2. To see how long tests take on Travis with options cloning but without the test shims vs with the test shims

janmeier:
Good idea splitting - 2 needs a rebase now

overlookmotel:
Have rebased this onto master after merge of PR https://github.com/sequelize/sequelize/pull/5874.

Enabling the tests shim approximately doubles the run time of the tests.

On my development machine (MacBook Pro Retina), the times to run MySQL integration tests:
Without options cloning: 99 secs
With options cloning: 101 secs
With test shim: 198 secs
(the above are averages over 5 runs of each)

The above is also roughly borne out by Travis's timings.
Without options cloning: 24m 4s
With options cloning: 17m 41s
With test shim: 34m 35s
(Travis's timings seem to diverge between runs quite widely, so the above is only a rough indicator).

So... question is: is it worth it?

In my opinion, having universal tests which check for passing of `options.logging` and that `options` object are not modified is a pretty good thing. Also, I'd now like to add to the shim to test for a few other cases:

1. `options.transaction` passed correctly
2. CLS context maintained correctly
3. All Promises returned from Sequelize methods are instances of `Sequelize.Promise` (this is important for CLS)

`options.transaction` and CLS are both in my opinion critical, so I think this would increase the utility of the test shim a lot.

Also in my personal opinion, the long-term solution to decreasing build times is to move to running tests in parallel (Mocha 3 will apparently support this https://github.com/mochajs/mocha/issues/1969 or could try [mocha.parallel](https://github.com/danielstjules/mocha.parallel)). Reducing number of supported Node versions would also help. But maybe all of these solutions are a long way off.

@mickhansen I guess you'd be the one most affected by the increased test run times. What's your opinion? Is it worth it?

sushantdhiman:
> With test shim: 34m 35s

Test shims are cool but thats too much time. Ideally we should reduce (or remove) IO js builds if we want use these updates.

overlookmotel:
@mickhansen @janmeier What do you think?

janmeier:
Since we have cut down the test times considerably, I think we should add this - Its important to make sure options always work. Perhaps we could enable it for one

overlookmotel:
I've rebased this on master and adapted it for the change to ES6 classes.

However, whether we can deal with the longer build times remains an open question.

So I've disabled the shims for now by default. To run them you have to use `SHIM=true npm test`. So this  PR won't affect the build times on Travis.

Could we possibly get this merged at this point just so the shims code doesn't get behind the main codebase again? We can discuss later whether it's worth enabling the shims or not. And in the meantime anyone who wants to run them can do so locally.

How's that sound for a compromise?

janmeier:
Sounds like a good compromise to me. Perhaps we can run the shims on one node version only, like with coverage.

We could also set it to run pre-release, but that would hinder us from doing quick fix-releases

overlookmotel:
@janmeier Thanks for merging. I guess we can come back to the question of whether to enable the shims later. Obviously I'd like to see them enabled! But I can totally see that long build times hinder productivity.

PR #5840, opened 2016-05-03
Fix #5736, Invalid query generated when using LIKE with ANY

### LINKED ISSUE #4064: Postgres $any with an array gets a sytax error
If I construct a WHERE clause with $any and an array, I get an error.

```
var selectedAnimals = ['cow','pig'];

where.title = {
    $iLike: { $any: selectedAnimals}
};
```

generates this SQL:

```
SELECT "name" FROM "Animals" AS "Animal" WHERE "Animal"."type" ILIKE ANY ARRAY['cow','pig']::VARCHAR;
```

and this error:

```
syntax error at or near "ARRAY"
```

From Postgres syntax, it seems like ANY should have brackets, like this:

```
SELECT "name" FROM "Animals" AS "Animal" WHERE "Animal"."type" ILIKE ANY (ARRAY['cow','pig'])
```

### LINKED ISSUE #5736: LIKE ANY ARRAY syntax error using PostgreSQL
## What you are doing?
Using this query:
{ 
  where: { 
    commonnames: { 
      '$like': {
        $any: ["c%", "%,c%"]
      }
    } 
  } 
}

on a table called fishes, I get this query:

Executing (default): SELECT "id", "species", "genusId", "commonnames", "createdAt", "updatedAt" FROM "fishes" AS "fish" WHERE "fish"."commonnames" LIKE ANY ARRAY['%c%','%,c%'] LIMIT 1;

which produces this syntax error:

Unhandled rejection SequelizeDatabaseError: syntax error at or near "ARRAY"
    at Query.formatError (/var/node/fishambiguator/node_modules/sequelize/lib/dialects/postgres/query.js:347:14)
    at null.<anonymous> (/var/node/fishambiguator/node_modules/sequelize/lib/dialects/postgres/query.js:81:19)
    at emit (events.js:107:17)
    at Query.handleError (/var/node/fishambiguator/node_modules/pg/lib/query.js:108:8)
    at null.<anonymous> (/var/node/fishambiguator/node_modules/pg/lib/client.js:171:26)
    at emit (events.js:107:17)
    at Socket.<anonymous> (/var/node/fishambiguator/node_modules/pg/lib/connection.js:109:12)
    at Socket.emit (events.js:107:17)
    at readableAddChunk (_stream_readable.js:163:16)
    at Socket.Readable.push (_stream_readable.js:126:10)
    at TCP.onread (net.js:538:20)

using PostgreSQL.

A successful query

The above error

Dialect: postgres
Database version: PostgreSQL 9.3.10 on x86_64-unknown-linux-gnu, compiled by gcc (Ubuntu 4.8.2-19ubuntu1) 4.8.2, 64-bit
Sequelize version: "sequelize": "^3.14.1"

sushantdhiman:
### Pull Request check-list

### Description of change

Closes #5736 
Closes #4064

I have detected the presence of `LIKE` and `ANY`, fixed the query if they are used in specific order.

cc @mickhansen @janmeier

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -2240,6 +2240,11 @@ var QueryGenerator = {
 
       if (escapeValue) {
         value = this.escape(value, field, escapeOptions);
+
+        //if ANY is used with like, add parentheses to generate correct query
+        if (escapeOptions.acceptStrings && (comparator.indexOf('ANY') > comparator.indexOf('LIKE'))) {
+         value = '(' + value + ')';
+        }
       }
     }
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,5 @@
 # Future
+- [FIXED] Invalid query generated when using LIKE + ANY [#5736](https://github.com/sequelize/sequelize/issues/5736)
 - [FIXED] Method QueryInterface.bulkDelete no longer working when the model parameter is missing. (PostgreSQL) [#5615](https://github.com/sequelize/sequelize/issues/5615)
 
 # 3.22.0
--- a/test/unit/sql/where.test.js
+++ b/test/unit/sql/where.test.js
@@ -572,7 +572,7 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
               $any: ['foo', 'bar', 'baz']
             }
           }, {
-            postgres: "\"userId\" LIKE ANY ARRAY['foo','bar','baz']"
+            postgres: "\"userId\" LIKE ANY (ARRAY['foo','bar','baz'])"
           });
 
           testsql('userId', {
@@ -580,7 +580,7 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
               $any: ['foo', 'bar', 'baz']
             }
           }, {
-            postgres: "\"userId\" ILIKE ANY ARRAY['foo','bar','baz']"
+            postgres: "\"userId\" ILIKE ANY (ARRAY['foo','bar','baz'])"
           });
 
           testsql('userId', {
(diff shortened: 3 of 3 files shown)


sushantdhiman:
Hi @mickhansen , Please review :)

lib/dialects/abstract/query-generator.js
```diff
@@ -2240,6 +2240,11 @@ var QueryGenerator = {
 
       if (escapeValue) {
         value = this.escape(value, field, escapeOptions);
+
+        //if ANY is used with like, add parentheses to generate correct query
+        if (escapeOptions.acceptStrings && (comparator.indexOf('ANY') > comparator.indexOf('LIKE'))) {
+         value = '(' + value + ')';
+        }
```
mickhansen: Ideally we still want to avoid string matching, better to add a case for $like/etc with a $any child.

sushantdhiman: I can add such logic [here](https://github.com/sushantdhiman/sequelize/blob/fix-5736/lib/dialects/abstract/query-generator.js#L2220), but it will require checking if `key` is either one of `$like, $iLike, $notLike, $notiLike` , which again will be better done using `indexOf` check. Because checking for all of them by if `==` conditions will be too verbose.

While I can move logic to proposed position, it will need same check (`indexOf` match). I think code will be more simple in current way, away from `comparator` creation process. 

Please suggest what you think @mickhansen

mickhansen: indexOf check on an array is more precise than on a string though :)

We'll move forward with this fix for now.

sushantdhiman:
Thanks for merging @mickhansen , Hope it got released soon. I was facing this issue too :)

PR #5847, opened 2016-05-03
sequelize#5843

### LINKED ISSUE #5843: Different aliases for join limit subquery and include
I'm going to describe what's happening - if we need an example, I'll get to it off the clock, because I imagine it'll take me a little bit to get worked out.

When you perform an include and a limit together, the query first performs a subquery to select the top-level models, then later uses the (same?) query to perform the includes. However, when it generates the subquery, it uses the default alias (model name), instead of whatever alias is being used at the time. In my case - the example is a to-many relationship, so the alias is plural for the include query, but the model name is singular so the alias is singular in the subquery. This is fine until you try to use a literal query that references column names, because in one it'll need to be "plural"."column" and and in the other it needs to be "singular"."column", which clearly doesn't work since they're generated from the same code. To me, it seems like these should use the same alias so this can't be an issue, regardless of naming scheme.

https://github.com/ojolabs/sequelize/commit/101a7bafff36d7339c4f3a463f7e0b931bdb5a4e

Cody157:
### Pull Request check-list
- [x] Closes #5843 
### Description of change

Alias the subquery of a query with both a limit and a required join to to be the same in all places it's used.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1346,6 +1346,7 @@ var QueryGenerator = {
 
             // Creating the as-is where for the subQuery, checks that the required association exists
             $query = self.selectQuery(include.model.getTableName(), {
+              tableAs: as,
               attributes: [association.identifierField],
               where: {
                 $and: [


mickhansen:
Be great to have an integration test for this

janmeier:

PR #5848, opened 2016-05-03
Check previous date value is not null before setting

### LINKED ISSUE #5719: No null check for dates when setting previous data values
## What you are doing?

``` js
beforeValidate(model) {
  return model.timestamp && !model.previous('timestamp')
}
```

Return `true` when timestamp changes from `null` to some date.

When timestamp was previously `null`, `model.previous('timestamp')` is a `Date` set at the epoch. In other words - it is getting converted to `new Date(null)`
Hence, this can never return true.

https://github.com/ojolabs/sequelize/commit/e2b2eb7313bb0dcce924e772edf8eed28e501b5a

Cody157:
### Pull Request check-list
- [x] Closes #5719
### Description of change

Prevent setting dates in previous data values to `new Date(null)` when they should be `null`.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -346,7 +346,7 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
             if (!(value instanceof Date)) {
               value = new Date(value);
             }
-            if (!(originalValue instanceof Date)) {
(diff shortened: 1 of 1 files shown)


mickhansen:
Can you add a unit test aswell please?

janmeier:

PR #5849, opened 2016-05-03
update hooks doc

CristiScheye:
No instances in params for bulk commands; see https://github.com/sequelize/sequelize/blob/master/lib/hooks.js#L58


janmeier:
Thanks :)

PR #5853, opened 2016-05-04
Adding logging to the error handling.

manchicken:
### Pull Request check-list

- [ Yes ] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [ No issue exists to my knowledge ] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ No, this isn't an issue ] Have you added new tests to prevent regressions?
- [ No, I don't think this affects documentation ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ No ] Have you added an entry under `Future` in the changelog?

### Description of change

This is just a simple change to add logging to where error formatting is being done. This way we can get a more raw log entry written whenever an error is encountered

### DIFF
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -200,6 +200,10 @@ Query.prototype.handleShowTablesQuery = function(results) {
 
 Query.prototype.formatError = function (err) {
   var match;
+
+  // Log that the error occurred...
+  this.sequelize.log('Encountered error: ' + err.message);
+
   match = err.message.match(/Violation of UNIQUE KEY constraint '((.|\s)*)'. Cannot insert duplicate key in object '.*'. The duplicate key value is \((.*)\)./);
   match = match || err.message.match(/Cannot insert duplicate key row in object .* with unique index '(.*)'/);
   if (match && match.length > 1) {
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -122,6 +122,11 @@ Query.prototype.formatResults = function(data) {
 Query.prototype.formatError = function (err) {
   var match;
 
+  // Log that the error occurred...
+  this.sequelize.log(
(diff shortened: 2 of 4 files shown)


janmeier:
Hmm, not sure I like the idea of forcing logging on the user - For example if you do a `create` which might hit a unique index - you catch the `UniqueConstraintErrror`, because its expected and move on. You don't want those non-errors showing up in your logs.

I think it would be better to add a [rejection handler](http://bluebirdjs.com/docs/api/error-management-configuration.html#global-rejection-events) to your own code. Or perhaps we could add an `onError` hook to sequelize.

Don't you agree @mickhansen ?

mickhansen:
Isn't the original error attached to all errors, so a user can just inspect that?
I don't think we should introduce this code, rejection handlers are better since a lot of ours generally enjoy our exception handling

janmeier:
Yep, the original error is always available in `original` and `parent`

manchicken:
Don't you guys actually have this using the logging config setting? It
seems odd that you wouldn't have it log errors when the `logging` setting
is enabled.

manchicken:
This isn't forcing logging, this is using logging if it is enabled with the `logging` setting.

janmeier:
The problem is that this logs all errors - I think we should only log unhandled errors, which can be done with the bluebird settings I linked

mickhansen:
I suppose there could be an argument for logging errors after the SQL has been executed, but in that case it should be the formatted error most likely

manchicken:
The problem I'm trying to solve is one where the database has odd
transactional behavior going on - updated records which sporadically don't
seem to actually be updated, despite no error messages at all.

It seems in keeping with the spirit of the concept of having SQL logging
enabled to log the SQL error code and message in a more raw format for the
dbms being used.

The logging feature seems like a troubleshooting feature in that regard, so
anything but raw log messages seem like they could potentially obscure the
error for the user of sequelize.

PR #5870, opened 2016-05-10
add allowEmptyString in rawAttribute

lonso:
I want to add property to Model define in column.
Some time i set property allowNull is false, means i want value is not null or undefined or ''.
But in mysql '' is different null,if you set column not null ,but it can insert ''.
So i want add new property of model define,it help me easy to use instance.update() method.
That's all.

Thanks.


mickhansen:
Isn't this already handled by the string length validator?

lonso:
how can i use  string length validator to define column?

janmeier:
http://docs.sequelizejs.com/en/latest/docs/models-definition/#validations

Issue #5205, opened 2016-01-13
Only support Node v4.0.0 and up


mickhansen:
To lower CI times for increased productivity and sense of well being.

alitaheri:
I agree with this :+1:

Verdier:
Agree :+1:  and project could use ES6 features such as `class` or *even* `co`...?! :battery:

mickhansen:
@Verdier We're likely not moving to generators internally, since ideally we'd still have unofficial support for older v8 versions with less features.

We plan to add class support yes, however we still need to support the old sequelize.define :)

felixfbecker:
And dont forget how much the codebase would benefit from arrow functions (also the docs).
Maybe this would also be the time to refactor the code so that "instances" are actually instances of models? E.g. `Sequelize.prototype.define` would return an anonymous class that extends the Model class, and "instances" will actually be instances of that class.

felixfbecker:
And optionally, you can pass `define` you own class as the first argument as long as it extends `Model`, so you can actually define getters/setters as real getters/setters and customize your model with everything, even `toJson()`.

mickhansen:
@felixfbecker yeah the plan is to change it so instances are instanceas of models, we need to change a lot of model checks though.

Plan is to support both ES6 classes and define.
As for arrow functions, definitely :) Away with var self = this

felixfbecker:
Then in the very distant ES7 future instead of `define` you could use a decorator on your model class
```js
@attributes({
  ...
})
@tableName('...')
class Test extends Sequelize.Model { ... }
```

mickhansen:
@felixfbecker yeah, there's another issue for ES6 class support, you can add your help and opinons here.
For ES6 we're probably looking at something like:

```js
class Test extends Sequelize.Model {
  static tableName () => 'test'
  static attributes () => ({
    firstName: Sequelize.STRING
  })
}
```

felixfbecker:
@mickhansen No, please please please. Don't do this. Thats what I hate the most about PHP ORMs, that just because the language is so inflexible you always have to implement these methods that do nothing but return the configuration. But Javascript _is_ flexible. You can assign stuff on classes. You can build classes dynamically. Think about it, currently `define` is basically a class factory. You pass it arguments _beforehand_ that define how this new model will map to the database.

[...]

Issue #4728, opened 2015-10-25
ES6 class support


mickhansen:
Would be great to support Sequelize models as ES6 classes.

Issues:
- No static properties in ES6 classes
- Models need to be init'ed to perform normalization etc
- Models need a way to be connected to an active sequelize instance
- Would break `instanceof Sequelize.Model`, so we might need an interface check inself.

Should instances be directly initiated from a model via `new User()`? Would be more semantic.

```js
// models/user.js
class User extends Sequelize.Model {
  static attributes() {
    return [
      // ...
    ];
  }

  // Class methods
  static findByEmail(email) {
    return this.findOne();
  }

  // Instance methods
  verifyPassword(password) {

  }
}

// models/index.js
import User from './user';
import sequelize from './sequelize-instance';

// Need some way to attach the model to a sequelize instance, specifically while allowing multiple attachments. Also need to run an sequelize init script on the model to normalize attributes etc.
User = sequelize.import(User); // ?
```

Internally this need to be implemented for how `extends` is done under the hood in babel etc, so prototype inheritance.

corbanb:
:100:

novacrazy:
Would a pseudo-factory pattern work well for connecting it to the sequelize instance?

Example:
```javascript
// models/user.js
export default class User extends Sequelize.Model {
// ...
}

// models/index.js
import UserModel from './user';
import sequelize from './sequelize-instance';

export const User = sequelize.import(User);

// somefile.js
import {User} from 'models/index.js';

let newUser = new User({name: 'username', email: 'example@example.com'});
```

where `import` is something like:

```javascript
sequelize.import = function(Model) {
    class ModelFactory extends Model {
        constructor(...args) {
            super(...args); //forward all arguments to original constructor

            //_connect inherited from Sequelize.Model
            this._connect(sequelize);
        }
    }

    return ModelFactory;
}
```

That way `newUser instanceof UserModel` remains true, too.

mickhansen:
@novacrazy I was thinking something along those lines yes.

mickhansen:
Need to figure out how to test this. We won't be using babel just yet internally, so it will be vanilla JS. But id like to test it with a babel compiled file class.

kripod:
I think that this should definitely be a milestone for v4 :smile:

Verdier:
[...]

PR #5877, opened 2016-05-12
Make Instances instances of Model

felixfbecker:
This PR does some changes I suggested in https://github.com/sequelize/sequelize/issues/5205#issuecomment-171451272 and https://github.com/sequelize/sequelize/issues/4728#issuecomment-218586821. I could not get the tests to run so I open this to see what CI says and to generate some discussion.

 - This removes the `Instance` class, it essentially becomes the `Model` class.
 - All the prototype methods of the old `Model` class (`find()`, `build()`, ...) are now static methods on the new `Model` class.
 - The code of the old constructor of the `Model` class is added to the static initializer `Model.init()`.
 - `sequelize.define()` is basically now a _class factory_, that returns a new class that inherits from `Model`, initialized with the passed options.
 - `User.build() instanceof User` is now true.
 - Instead of passing `instanceMethods`, `classMethods` `getterMethods`, `setterMethods` you can now simply set those methods on the Model / the Model's prototype directly which is awesome for strict typing (TypeScript)
 - Instead of using `Model.build()` you can also just use `new Model()` to get an instance
 - To check if a variable _is_ a `Model`, you have to check `x.prototype instanceof Model`, to check if a variable is _an instance_ of a `Model` you have to check `x instanceof Model`
 - The property `instance.Model` is now exactly the same as `instance.constructor` (a reference to the class that created the object)
 - For BC the references `.Instance` are still there and set to the Model class (since the `Model` class replaces the `Instance` class)
 - `Model.name` is set to the model name in `sequelize.define()` via `Object.defineProperty`. Only in the super-old Node versions, where `Function.prototype.name` is not `configurable`, `eval()` is used to dynamically create a class with the model name.

Changes in semantic, expressed in TypeScript definitions:
```ts
abstract class Model {
  static init(attributes: ModelAttributes, options: ModelOptions, modelManager: ModelManager): void;
  static find(options: FindOptions): Promise<Model>;
  save(options?: InstanceSaveOptions): Promise<void>;
}

class Connection {
  define(modelName: string, attributes: ModelAttributes, options?: ModelOptions): typeof Model;
}
```
Model creation with `define()` as a class factory is supported just like before, but it is in theory also possible to subclass `Model` directly for type-safety and call `Model.init()` yourself:
```ts
  public username: string;
  myMethod() {
    return 'hello';
  }
}
Model.init(User.name, {username: Sequelize.STRING}, {tableName: 'users'}, sequelize.modelManager)
```
`Model.init()` is of course low-level API that is called by `define()`, but it would _finally_ allow third-party wrappers to do this in [decorators](https://github.com/sequelize/sequelize/issues/4728#issuecomment-201882392): A model would extend `Sequelize.Model` (similar to how it is today), property/method decorators would add metadata about types, hooks, associations and a class decorator would read them and call the static initializer `Model.init()` with the info.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2148,6 +2148,8 @@ Model.prototype.bulkCreate = function(records, options) {
     }
     options.updateOnDuplicate = updatableFields;
   }
+  
+  options.model = this;
 
   var self = this
     , createdAtAttr = this._timestampAttributes.createdAt
@@ -2226,8 +2228,6 @@ Model.prototype.bulkCreate = function(records, options) {
         }
       }
 
-      // Insert all records at once
-      options.model = self;
       return self.QueryInterface.bulkInsert(self.getTableName(options), records, options, attributes).then(function (results) {
         if (Array.isArray(results)) {
           results.forEach(function (result, i) {
@@ -2382,7 +2382,8 @@ Model.prototype.restore = function(options) {
   }, options || {});
 
   options.type = QueryTypes.RAW;
-
+  options.model = this;
+  
   var self = this
     , instances;
 
@@ -2498,6 +2499,8 @@ Model.prototype.update = function(values, options) {
   if (this._timestampAttributes.updatedAt && !options.silent) {
     values[this._timestampAttributes.updatedAt] = this.$getDefaultTimestamp(this._timestampAttributes.updatedAt) || Utils.now(this.sequelize.options.dialect);
   }
+  
+  options.model = this;
 
   var instances
     , valuesUse;


overlookmotel:
Hmm interesting. `new User()` rather than `User.build()` is really nice.

However, under this new scheme, is there a way to add a method universally to all instances of all models? At present you can do `Sequelize.Instance.prototype.myMethod = function() { /* blah */ }` and that works even *after* models have been defined.

How would you do that under the new scheme? Being able to add (or shim) methods on the universal Model Instance prototype is **very** useful for plugins.

Also...

> To check if a variable is a Model, you have to check x.prototype instanceof Model, to check if a variable is an instance of a Model you have to check x instanceof Model

`x.prototype instanceof Model` doesn't seem ideal to me. But I'm not sure what else could be done.

overlookmotel:
It's tricky. Depending on the circumstances, you want `user instanceof User` but also `User instanceof Model`. It's not possible to have both.

Which should take precedence is probably a matter of opinion and what feels "right" will be different users depending on what they use Sequelize for.

As someone who maintains a few Sequelize plugins, I see the schema more in the abstract so the power that `User instanceof Model` gives is more important for me, but for a more standard Sequelize consumer, all they probably care about is `user instanceof User`.

overlookmotel:
By the way, would `x instanceof Sequelize.Model` tell you whether `x` is a model instance (which model unspecified)?

felixfbecker:
@overlookmotel 
> However, under this new scheme, is there a way to add a method universally to all instances of all models?

Sure, simply do `Sequelize.Model.prototype.myMethod = function () { ... }`

> Depending on the circumstances, you want  user instanceof User  but also  User instanceof Model . It's not possible to have both.

It _is_ possible to have both, `User` will always inherit from `Model`, so `user instanceof User && user instanceof Model`. Basic OOP.

> By the way, would  x instanceof Sequelize.Model  tell you whether  x  is a model instance (which model unspecified)?

Yes.

felixfbecker:
The tests fail because I try to add the `name` property on the model inside `define()`, which is not possible in older Node versions because `name` is the name of the class and `define()` would return an anonymous class. I tested it on Node 4 and 6, where this works just fine. The only workaround I can see is to either use `new Function()` / `eval()` (bad for performance?) or set `_name` instead.

felixfbecker:
I fixed the issue with the name property, will fallback to using eval() if `Function.prototype.name` is not configurable. However, there have been commits made to master and Travis will not run a build if the conflicts are not resolved - I cannot easily merge/rebase because I add the whole Instance code manually to model.js and adapt all the properties and instanceof checks. That is a lot of work... @mickhansen Could you trigger a manual build?

overlookmotel:
> > Depending on the circumstances, you want `user instanceof User` but also `User instanceof Model`. It's not possible to have both.

> It *is* possible to have both, `User` will always inherit from `Model`, so `user instanceof User && user instanceof Model`. Basic OOP.

Sorry, I was talking about `User instanceof Model` not `user instanceof Model` (note capital U on User).

```js
var User = sequelize.define( 'User', /* blah */ );
assert( User instanceof Sequelize.Model == true );
```

This will no longer work right? After this PR, you'd have to use `User.prototype instanceof Sequelize.Model`?

So would there still be a way to add a static method to all models? Currently:

```js
var myMethod = function () { /* blah */ };
Sequelize.Model.prototype.myMethod = myMethod;
var User = sequelize.define( 'User', /* blah */ );
assert( User.myMethod == myMethod );
```

overlookmotel:
NB Isn't `Object.setPrototypeOf()` generally bad for performance?

felixfbecker:
@overlookmotel 
> This will no longer work right? After this PR, you'd have to use  User.prototype instanceof Sequelize.Model ?

Yes.

> So after this PR, to add a static method to all models, you add a method on  Sequelize.Model ?

Yes. But imo globally patching an imported class from a module is a very bad pattern. The better option, that is now possible, would be if you subclass `Sequelize.Model`, define your custom methods there, and then let your models extend your custom subclass:
```ts
abstract class MyBaseModel extends Sequelize.Model {
  static myStaticMethod() { ... }
  myMethod() { ... }
}
class User extends MyBaseModel { ... }
Model.init(User.name, ...)
```
And extending classes really is basic OOP. Of course Javascript's prototype concept is a bit hard to grasp, but imo having instances be instances of the model classes is what a user expects.

> NB Isn't  Object.setPrototypeOf()  generally bad for performance?
> In node v4 and below  util.inherits()  uses  ctor.prototype = Object.create(superCtor.prototype)  rather than  Object.setPrototypeOf() 

If you look at [my code](https://github.com/felixfbecker/sequelize/blob/574d5886b66c5eb7fbf33deefbee863bb20e9e58/lib/sequelize.js#L633) you see I do the same. But that is only so the instances inherit the prototype properties. But you also have make the new Model class inherit the static methods, like `find()`. `util.inherits()` actually doesn't care about static properties, but [Babel will transpile it that way](https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=class%20Model%20%7B%0A%20%20static%20find()%20%7B%0A%20%20%20%20%0A%20%20%7D%0A%7D%0A%0Afunction%20define()%20%7B%0A%20%20return%20class%20extends%20Model%20%7B%7D%0A%7D).
See the `_inherits` function:
```js
function _inherits(subClass, superClass) {
    if (typeof superClass !== "function" && superClass !== null) {
        throw new TypeError("Super expression must either be null or a function, not " + typeof superClass);
    }
    subClass.prototype = Object.create(superClass && superClass.prototype, {
        constructor: {
            value: subClass,
            enumerable: false,
            writable: true,
            configurable: true
        }
    });
    if (superClass) Object.setPrototypeOf ? Object.setPrototypeOf(subClass, superClass) : subClass.__proto__ = superClass;
}
```

overlookmotel:
OK, let's start again from the beginning...

I'm not against what you're doing here. I think generally speaking, having an instance of `User` be `instanceof User` rather than `instanceof User.Instance` is a good thing. I was just asking questions to ensure that no functionality gets lost in the process and that performance isn't affected.

I'd say that this should definitely be considered a breaking change - any user code like `if (item instanceof Sequelize.Model) { ... }` would break - so this should, as you say, land in 4.0.

Concerning OOP, yes `user instance User` is basic OOP which everyone should understand. It's the part about the `User` *constructor* inheriting its static methods from `Sequelize.Model` which is less standard. But anyway, let's not argue about whether something is "basic OOP" or not. It's difficult to read someone telling you that you're not understanding something "basic" as pretty equivalent to saying "you are an idiot", which isn't a conversation that's going to get us anywhere!

In an app, as you say, it'd be better to subclass `Sequelize.Model` before modifying it. However, with plugins, the user would generally expect that after applying the plugin, they can use the same objects as before but expect them to exhibit extra functionality. My rule of thumb has been that it's OK for a plugin to shim global Sequelize methods BUT the shims should not change the behaviour of the method unless an option passed to the method explicitly "activates" this new behaviour. So any other instance of Sequelize will behave exactly as before. i.e. the global has been modified but it has no effect by default.

My only real concern about this PR is whether altering the prototype of each model constructor with `Object.setPrototypeOf(model, Model)` or `model.__proto__ = Model` is going to impact performance on operations like `User.findAll()` (which is probably the most often called Sequelize method of all).

My point about `Object.setPrototypeOf()` not being used in `util.inherits()` before node v4 wasn't that you should use `util.inherits()` - that wouldn't work here - but that this may indicate that before node v4 `Object.setPrototypeOf()` isn't performant. I don't know about this, it's just a guess. Maybe it's still not performant! Certainly [MDN](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf) contains a pretty stern warning that dynamically setting prototypes is bad for performance.

Anyway, the question of performance will become clearer once Travis starts running.

I'd also suggest adding 2 static methods to `Sequelize`:

```js
Sequelize.isModel = function(item) { return item && item.prototype instanceof Sequelize.Model; };
Sequelize.isInstance = function(item) { return item && item instanceof Sequelize.Model; };
```

These could be used internally - easier to understand than `item.prototype instanceof Sequelize.Model`. Also, these methods could be added now to v3, so users can start transitioning their `item instanceof Sequelize.Model` code to `Sequelize.isModel(item)` so then it won't break with Sequelize v4.

felixfbecker:
@overlookmotel I was in no way trying to insult your programming skills. Sorry if I sounded a bit rude, just wanted to answer your questions :)

I would still argue though that a subclass inheriting not only instance methods, but also static methods, is expected behaviour - it's how it works in every other OOP language, the ES6 spec and Babel.

Altering the prototype of an object definitely has a performance impact because it destroys the "prediction" V8 can make about objects. Still, it is the only spec-compliant way to make a class inherit from another. As I said in the original post, this PR is mostly to generate discussion. I would propose to drop support for Node <4 in 4.0 and then use real classes. I'm just a bit worried that 4.0 will take a long time to release. I would rather keep the API changes small (make `define()` work like before) but clean up the semantics (`instance instanceof Model`) and make it possible to use classes and decorators for those who want it (I would really like to write a decorator library for Sequelize).

Adding `isModel()` and `isInstance()` is a great idea. Maybe `isInstance()` should be named `isModelInstance()` though? Because with the 4.0 semantics one could wonder _isInstance - instance of what?_

overlookmotel:
@felixfbecker Yes you're right `isModelInstance()` is a better name.

Concerning performance, well I guess we'll have to see what impact this has once you've rebased and Travis runs. You still can't subclass `Function` in newer Javascript versions right? But in my opinion, if this PR heavily impacts performance, that's a bit of a sticking point.

felixfbecker:
@overlookmotel well if 4.0 will only support Node >4 I would redo my PR to use classes, and then no `Object.setPrototypeOf()` or subclassing `Function` would be necessary. `define()` would just do
```js
return class extends Model {}
```

janmeier:
I think most of what needs to be said has already been said here - I'm just waiting to see the performance impact ;-).

One problem with extending (and overlookmotel might already have mentioned this) is plugins which are working on `before / after Define` hooks - these will no longer work since there is not define if using classes. But they might be called on Model.init instead ?

felixfbecker:
I will make a new PR that uses classes instead of rebasing this (that would be a pain). I originally commited on master (stupid me), it is now in https://github.com/felixfbecker/sequelize/tree/instance-instanceof-model

Issue #4639, opened 2015-10-12
Transitive Persistence (Cascading Saves, Updates and Deletes) and Offline concurrency.


hendrul:
Maybe there's a reason to this but after searching at several data mapper implementations for nodejs
I realized that none of them provides implementation of the [UnitOfWork](http://martinfowler.com/eaaCatalog/unitOfWork.html) pattern. The point of ORM is to abstract away low level storage-efficiency oriented relational model and operations, to the more conceptual and bussiness oriented Object model, but having the programmer keeping track and worring about the referential order in which they need to do things is a gap that needs to be filled by an ORM implementation.

The main points of Unit Of Work are:
1. As soon as you start doing something that may affect a database, you create a Unit of Work to keep track of the changes. Every time you create, change, or delete an object you tell the Unit of Work.
2. Check for inconsistent reads by verifying that none of the objects changed on the database during
    the business transaction. 
3. When it comes time to commit, the Unit of Work decides what to do. It opens a transaction, does any concurrency checking (using [Pessimistic Offline Lock](http://martinfowler.com/eaaCatalog/pessimisticOfflineLock.html) or [Optimistic Offline Lock](http://martinfowler.com/eaaCatalog/optimisticOfflineLock.html) ), and writes changes out to the database.  
4. Application programmers never explicitly call methods for database updates. This way they don't have to keep track of what's changed or worry about how referential integrity affects the order in which they need to do things.

hendrul:
Following with #1393. What I see as the most wanted point of UoW pattern for sequelize is the cascading of operations for a graph of related instances. The programmer should be released from this task, without being aware of the order the operations should be performed considering this information already exists on the mapping. What prevents all node's ORM projects from implementing this pattern?. Excuses by my ignorance, I know there is a great effort to improve sequelize and I really appreciate it.

hendrul:
Other important point is what's the recommended way to handle business transactions? 
how check the consistency of data edited offline before it is commited to BD?
Is it viable to include support on sequelize for this?

mickhansen:
I don't have a great deal of insight to how UoW is implemented on other platforms.
But for Node.js i'd say a major issue is how concurrency is handled, we have no way of knowing whether or not a call should go in a specific unit of work (well we might with CLS, but the user would have to wrap evertyhing in a function anyways).

As for offline locks on transaction commit, if you hold all actual calls untill you're ready to commit we could implement something i suppose.

hendrul:
[...]

PR #5885, opened 2016-05-13
Add NotAffectedRows error and adds where conditions on Instance save

dalthon:
### Pull Request check-list

### Description of change

Transitive Persistence is under discussion at #4639, and I believe that a few people need Optimistic Lock right now.

This PR, maybe, can be a baby step toward implementing Optimistic Lock on Postgres, or can provide a useful alternative before an UnitOfWork implementation.

This adds a where option on instance.save and an ensureAffectedRows flag on Query.run options that can be used to throw a NoAffectedRowsError that could be handled when desired.

Example of usage:

```
instance.save({ where: { version: 123 }, ensureAffectedRows: true})
  .catch(function(err){
    // Handle error, maybe retring...
  })
```

Ps.: I run tests with docker (on OS X but with dinghy), unfortunatelly they always break when sqlite is installed, so I removed node_modules/sqlite3 to ignore it. I believe that it will work, but I'm aware that I didn't tested properly.


lib/instance.js
```diff
@@ -482,6 +482,8 @@ Instance.prototype._setInclude = function(key, value, options) {
  *
  * @param {Object} [options]
  * @param {string[]} [options.fields] An optional array of strings, representing database columns. If fields is provided, only those columns will be validated and saved.
+ * @param {string[]} [options.where] An optional object with where conditions.
```
janmeier: Hmm, not sure I like this - `instance.save` is meant to save a single row, which should be identified by primary key - I think adding a where argument here as well might confuse new users as to what the method does. I think this feature fits better in `model.update`

janmeier: Ah, i looked at the code before reading the PR description, and I fully realize how this can help UoW. But I'm still not sure i like it

janmeier:
Do you mind adding support for other dialects as well?

dalthon:
Sure! I can add support for others dialects, just did to Postgres firstly because it's the one I needright now, and I was not sure if my PR would be accepted, is just my first PR to sequelize.

janmeier:
Ping @dalthon

dalthon:
Hi @janmeier, sorry for the delay... I'm working now on MySQL and SQLite support...

After that I'll work on MSSQL, and I'll use travis to debug it.

When I finish everything I'll ping back, and it will be sooner!

dalthon:
Don't trust my tests... They are passing, but they are wrong now...

mickhansen:
Let's keep the naming and options similar to https://github.com/sequelize/sequelize/pull/5510

lib/dialects/postgres/query.js
```diff
@@ -131,6 +132,10 @@ Query.prototype.run = function(sql, parameters) {
       }
     }
 
+    if (self.options.ensureAffectedRows && result.rowCount < 1) {
+      throw new sequelizeErrors.NoAffectedRowsError('No rows affected by: ' + sql);
+    }
```
mickhansen: Would it not be simpler to handle this at the Model.update level since all the abstractions are taking care off at that point?

dalthon:
Ok! I will change ensureAffectedRows to throwOnEmpty

lib/dialects/postgres/query.js
dalthon: My first move was to handle it on the legacy "Instance.save" (now would be Model).

But when I started doing this I noticed that each database handles it's metadata differenty, somethimes don't send nothing, names didn't match, and sometimes send it as data.

Then, the main two options that I considered were:

1. Changing Query.run signature responding something different when ensureAffectedRows is true, so I could handle at Model
2. Handing in a database-specific way

Then I choose 2.

What do you think? Another idea that I disconsidered?

mickhansen: It should be possible to handle it by looking at type and return value in `sequelize.query`, if you're seeing different results from the dialects they would need to be fixed first.

mickhansen:
@dalthon We want a different name than rejectOnEmpty, i just want the same general style (being able to use true, a class and a string). Maybe `rejectOnNonAffected`

test/integration/instance.test.js
```diff
@@ -2069,4 +2070,108 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
...
+
+    describe('when rejectOnNonAffected is set to true or not set', function() {
+      it('should not save if conditions match and rejectOnNonAffected is true', function() {
+        var user;
+        var self = this;
+        return this.User.create({ username: 'Versioned User' }).then(function(u) {
+          user = u;
+          expect(u.version).to.be.equal(0);
+          u.version++;
+          return u.save({ where: {version: 10}, rejectOnNonAffected: true });
```
janmeier: `return expect(u.save).to.be.rejectedWith(..)`

test/integration/instance.test.js
```diff
@@ -2069,4 +2070,108 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
...
+
+  describe('conditions on instance save', function() {
+    describe('when rejectOnNonAffected is set to false', function() {
+      it('should not save if conditions match', function() {
+        var self = this;
+        return this.User.create({ username: 'Versioned User' }).then(function(user) {
+          expect(user.version).to.be.equal(0);
+          user.version++;
+          return user.save({ where: {version: 10}, rejectOnNonAffected: false });
+        }).tap(function(user) {
```
janmeier:
```js
}).then(user => {
  expect(user.version).to.be.equal(1);
  return user.reload();
}).then(user => 
 expect(user.version).to.be.equal(0);
});
```

janmeier:
Looks good so far, though some dialects are failing - Looks like those that check for `isUpdateQuery`, so it might not be catching some cases. Also, its not implemented for mssql

Have a look at how `rejectOnEmpty` handle boolean/string/Error https://github.com/sequelize/sequelize/pull/5510/files#diff-a140b60b6a99a8c69f4b5d0d1f1f2bfaR1415

changelog.md
```diff
@@ -72,6 +72,10 @@
 - [FIXED] Method QueryInterface.bulkDelete no longer working when the model parameter is missing. (PostgreSQL) [#5615](https://github.com/sequelize/sequelize/issues/5615)
 - [ADDED] Context and custom options for deep creation
 - [FIXED] Dates with millisecond precision are inserted correctly in MySQL [#5855](https://github.com/sequelize/sequelize/pull/5855)
+- [ADDED] Where conditions on Instance.save
+- [ADDED] NotAffectedRows error on Postgres on update when option ensureAffectedRows is set to true
```
sushantdhiman: Just single `NotAffectedRows` change log without any dialect name will be sufficient

lib/dialects/mssql/query.js
```diff
@@ -80,6 +80,11 @@ class Query extends AbstractQuery {
               this.sequelize.log('Executed (' + (connection.uuid || 'default') + '): ' + this.sql, (Date.now() - queryBegin), this.options);
             }
 
+            if(this.options.rejectOnNonAffected && results[0] && results[0].AFFECTEDROWS < 1){
+              reject( new sequelizeErrors.NoAffectedRowsError('No rows affected by: ' + sql) );
+              return ;
+            }
```
mickhansen: Could this not still be done at a higher level than each individual dialect?

PR #5889, opened 2016-05-14
Remove what I can only assume is leftover code

JoshWillik:
### Description of change

I don't think there's any functionality change here.
As far as I can tell this line was just left over from old code?
That or the indentation is misleading.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1364,9 +1364,6 @@ Model.prototype.findAll = function(options) {
 
   _.defaults(options, { hooks: true, rejectOnEmpty: this.options.rejectOnEmpty });
 
-    //set rejectOnEmpty option from model config
-    options.rejectOnEmpty = options.rejectOnEmpty || this.options.rejectOnEmpty;
-
   return Promise.bind(this).then(function() {
     conformOptions(options, this);
     this.$injectScope(options);


sushantdhiman:
Hi @JoshWillik , This is not a leftover code. It override the `rejectOnEmpty` config from model options. Without this line even if model config set `rejectOnEmpty` true, it will not work for individual queries performed by the model.

janmeier:
Defaults show the injecting it in the line above - the problem is [here](https://github.com/JoshWillik/sequelize/blob/f5f2d244afbba141f593aeddb267fddc586e4256/lib/model.js#L1541) where findOne is setting to false, which means the global options is not used by defaults. This causes [this](https://github.com/sequelize/sequelize/blob/master/test/integration/model/find.test.js#L956) test to fail.

Currently, you can't set the reject to false on a single find call, because a false value would always be overridden by the global value.

Can you remove the false value from findOne, and add a test for global reject opt +reject false on a find call not being rejected

janmeier:
Ping @JoshWillik

sushantdhiman:
@janmeier , this piece of code should not be removed. I have added specific test for this case which in failing in all CI tests.

> Model find rejectOnEmpty mode works from model options

mickhansen:
@sushantdhiman It should not be removed, but it should be changed to check for undefined rather than falsy.

JoshWillik:
Since I'm not super familiar with the logic flow here, and this ended up being a bit more nuanced than I had first assumed.

Sorry guys :(

PR #5896, opened 2016-05-16
Ryan/master/update schema method documentation

RyanThomas73:
### Description of change

Updated the documentation for Model.schema(schema, [options]) to include a description of the intended use case and a mention of how to set a default schema for a model via the define params.


janmeier:
Thanks! :)

PR #5897, opened 2016-05-16
Fix #5880, deletedAt compare against NOW

### LINKED ISSUE #5880: deletedAt should be compare against NOW rather than IS NULL
Under current implementation sequelize always compare `deletedAt` field with `IS NULL`, while it works but it just act as a boolean field.

It should ideally compare `deletedAt` against current timestamp value like `NOW`. This will make `deletedAt` semantically correct because its a date field.

This will also allow instances to be deleted on future dates. Some features can be implemented easily which requires expiration after some time like session, tickets, oauth tokens etc...

__Dialect:__  any
__Sequelize version:__ master

sushantdhiman:
### Pull Request check-list

### Description of change

Closes #5880 , `deletedAt` field is compared against `CURRENT_TIMESTAMP` rather than `IS NULL`. It also check for `IS NULL` in case `deletedAt` is not set at all.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 - [CHANGED] Throw `bluebird.AggregateError` instead of array from `bulkCreate` when validation fails
 - [FIXED] `$notIn: []` is now converted to `NOT IN (NULL)`  [#4859](https://github.com/sequelize/sequelize/issues/4859)
 - [FIXED] Add `raw` support to `instance.get()` [#5815](https://github.com/sequelize/sequelize/issues/5815)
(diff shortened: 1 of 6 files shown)


sushantdhiman:
Hi @mickhansen @janmeier, Please have a look when you get chance. CI is clear except of one timeout error.

janmeier:
Sweet! :)

janmeier:
Notice that this PR is on the 4.0 branch - You are welcome to make a PR to v3 as well, then I'll publish a new version

sushantdhiman:
I am currently using my sequelize fork in my projects so I can wait till 4.0 is release.

If any one else need this update in current semversion let me know and I will push it to v3 as well.

PR #5902, opened 2016-05-17
Adding the ability to log warnings if they are present in MySQL.

### LINKED ISSUE #5900: MySQL Warnings are completely lost when strict mode disabled.
In environments where strict mode is disabled - common among many PHP shops - MySQL warns where it might otherwise error out.

Consider this sample program...

## Schema

Note that MySQL 

```sql
create table foo (
  id int not null auto_increment primary key,
  some_number int
);
```

## Program

```javascript
'use strict'

// connection-wrapper just instantiates Sequelize and exports an object
// with `sequelize` as the key containing that instance.
const db = require('db/connection-wrapper')
const process = require('process')

function doIt () {
  let txn

  return db.sequelize.transaction({autocommit:false})
  .then( t => {
    txn = {transaction:t}
    // Notice how we're trying to insert "horse" for an integer value
    return db.sequelize.query("insert into foo(some_number) values('horse')", txn)
  })
  .then( () => {
    return txn.transaction.commit()
  })
  .catch( err => {
    console.error( 'The update failed.', err )
    return txn.transaction.rollback()
  })
}

doIt()
.then( () => process.exit(0) )
```

## Outcomes

 - Inserted record is `id:1` and `some_number:0`
 - The warning is ignored. It contained: `Incorrect integer value: 'horse' for column 'some_number' at row 1`

I have a pull request that I will be submitting shortly. I can only test against MySQL since that's what I've got, I also don't think that this is terribly relevant to other DBMS', though I'm open to disagreement there.

manchicken:
### Pull Request check-list

 + One caveat, tests are failing on `master`. My tests fail identically to those on `master`.
- [X] Does your issue contain a link to existing issue (Closes #5900) or a description of the issue you are solving?

### Description of change

This provides the ability for MySQL logs to be reported using the existing `sequelize.log()` functionality. This should help people find issues which MySQL doesn't consider errors, but does consider worthy of warning on.

### DIFF
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -15,7 +15,8 @@ var Query = function(connection, sequelize, options) {
   this.options = Utils._.extend({
     logging: console.log,
     plain: false,
-    raw: false
+    raw: false,
+    showWarnings: false
   }, options || {});
 
   this.checkLoggingOption();
@@ -29,6 +30,7 @@ Query.prototype.run = function(sql, parameters) {
   
   //do we need benchmark for this query execution
   var benchmark = this.sequelize.options.benchmark || this.options.benchmark;
+  var showWarnings = this.sequelize.options.showWarnings || this.options.showWarnings;
 
   if (benchmark) {
     var queryBegin = Date.now();
@@ -43,19 +45,25 @@ Query.prototype.run = function(sql, parameters) {
         self.sequelize.log('Executed (' + (self.connection.uuid || 'default') + '): ' + self.sql, (Date.now() - queryBegin), self.options);
       }
 
-      // Log warnings if we've got them.
-      if (results && results.warningCount > 0) {
-        self.logWarnings();
-      }
-
       if (err) {
         err.sql = sql;
 
         reject(self.formatError(err));
       } else {
-        resolve(self.formatResults(results));
+        resolve(results);
       }
     }).setMaxListeners(100);
+  })
+  // Log warnings if we've got them.
+  .then(function(results){
+    if (showWarnings && results && results.warningCount > 0) {
+      return self.logWarnings(results);
+    }
+    return results;
+  })
+  // Return formatted results...
+  .then(function(results){
+    return self.formatResults(results);
   });
 
   return promise;
@@ -123,19 +131,27 @@ Query.prototype.formatResults = function(data) {
   return result;
 };
 
-Query.prototype.logWarnings = function () {
+Query.prototype.logWarnings = function (results) {
   var self = this;
-  self.connection.query('SHOW WARNINGS', function(err, warningResults) {
-    self.sequelize.log('MySQL Warnings (' + self.connection.uuid + '):', self.options);
-    warningResults.forEach(function(_warningResult) {
-      if (_warningResult.hasOwnProperty('Message')) {
-        self.sequelize.log('\t'+_warningResult.Message, self.options);
-      } else {
-        _warningResult.keys().forEach( function(_objectKey) {
-          self.sequelize.log('\t'+[_objectKey, _warningResult[_objectKey]].join(': '), self.options);
-        });
-      }
+  return this.run('SHOW WARNINGS').then(function(warningResults) {
+    var warningMessage = 'MySQL Warnings (' + (self.connection.uuid||'default') + '): ';
+    var messages = [];
+
+    warningResults.forEach(function(_warningRow){
+      _warningRow.forEach(function(_warningResult) {
+        if (_warningResult.hasOwnProperty('Message')) {
+          messages.push(_warningResult.Message);
+        } else {
+          _warningResult.keys().forEach( function(_objectKey) {
+            messages.push([_objectKey, _warningResult[_objectKey]].join(': '));
+          });
+        }
+      });
     });
+    
+    self.sequelize.log(warningMessage + messages.join('; '), self.options);
+
+    return results;
   });
 };
 
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -27,10 +27,6 @@ Query.prototype.run = function(sql, parameters) {
   var self = this;
   this.sql = sql;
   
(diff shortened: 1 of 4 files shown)


manchicken:
So, it looks like some PGSQL tests are failing. I don't have any diffs in that area, so I'm not sure what's up with that. Are there known issues running the Travis-CI suite against PGSQL?

test/integration/sequelize.test.js
```diff
@@ -1197,13 +1216,15 @@ describe(Support.getTestDialectTeaser('Sequelize'), function() {
 
       it('through Sequelize.sync()', function() {
         var self = this;
+        self.spy.reset();
```
janmeier: This should not be needed since a new spy is created for each test run

test/integration/model.test.js
```diff
@@ -277,7 +277,7 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         bCol: { type: Sequelize.STRING, unique: 'a_and_b' }
       });
 
-      return User.sync({ force: true, logging: _.after(2, _.once(function(sql) {
```
janmeier: What is the warning logged here - I think we should rather remove it than working around it. This breaks tests for other dialects

test/integration/sequelize.test.js
```diff
@@ -265,6 +266,24 @@ describe(Support.getTestDialectTeaser('Sequelize'), function() {
...
+          benchmark: false
+        });
+        var insertWarningQuery = 'INSERT INTO ' + qq(this.User.tableName) + ' (username, email_address, ' +
+          qq('createdAt') + ', ' + qq('updatedAt') +
+          ") VALUES ('john', 'john@gmail.com', 'HORSE', '2012-01-01 10:10:10')";
+
+        return sequelize.query(insertWarningQuery)
+        .then(function(results) {
+          expect(logger.calledTwice).to.be.true;
+          expect(logger.args[1][0]).to.be.match(/^MySQL Warnings \(default\): Data truncated for column 'createdAt' at row \d+$/m);
```
janmeier: The error message is slightly different on the mysql version on travis, so this test is failing

lib/dialects/mysql/query.js
```diff
@@ -48,9 +48,20 @@ Query.prototype.run = function(sql, parameters) {
...
       } else {
-        resolve(self.formatResults(results));
+        resolve(results);
       }
     }).setMaxListeners(100);
+  })
+  // Log warnings if we've got them.
+  .then(function(results){
+    if (results && results.warningCount > 0) {
+      return self.logWarnings(results);
```
janmeier: You could call `query.run` recursively here, setting benchmark and logging to false

janmeier:
I like the idea, once my comments are adressed this should be good to merge

janmeier:
I don't know if we would need an option to turn this of perhaps?

test/integration/model.test.js
manchicken: I'll add an option to enable warnings as a distinct option from logging.

lib/dialects/mysql/query.js
manchicken: Trying the recursive call. `Query.prototype.run()` doesn't take those arguments, but the recursive call seems to be okay using the same arguments. Since `SHOW WARNINGS` isn't really a query which could result in a warning itself, I'm not worried about infinite recursion here.

test/integration/sequelize.test.js
manchicken: I put that in because it was needed. It was breaking the tests without this.

manchicken:
@janmeier: Thanks so much for your feedback. I've updated the pull request, thanks for your help.

test/integration/sequelize.test.js
janmeier: Still breaking after you added the `showWarnings` flag?

lib/dialects/mysql/query.js
```diff
@@ -118,6 +130,30 @@ Query.prototype.formatResults = function(data) {
   return result;
 };
 
+Query.prototype.logWarnings = function (results) {
+  var self = this;
+  return this.run('SHOW WARNINGS')
+  .then( function(warningResults) {
```
janmeier: Please put the `then` on the previous line, remove space before `function` and add space after ;) (sorry for nitpicking)

janmeier:
@manchicken Needs a rebase and a couple of small style fixes

manchicken:
If we're down to just nitpicks I don't mind :)

How's everything else? It looks like there's a Travis CI build failure, but I don't know if that's "normal".

janmeier:
Travis is failing with a lint error
> lib/dialects/mysql/query.js: line 149, col 9, Missing semicolon.

manchicken:
Sorry, totally forgot to jshint with this latest push. Need to install the
sublimetext plugin so I can't forget.

janmeier:
No problem, ping me when you have fixed + rebased ;)

manchicken:
I think that's everything. Did I miss anything?

janmeier:
Branch still has conflicts - did you do a `git rebase -i upstream/master` (where upstream = the main repo)

manchicken:
I thought I had. That's weird. I'll try again.

manchicken:
I just pushed a second attempt of the rebase... it shows no diffs between `upstream/master` other than the ones in my pull request.

manchicken:
(sorry for all of the noise, totally forgot to fetch before rebasing)

manchicken:
Looks like I need to fix the regex on my test.

manchicken:
@janmeier: any thoughts on how to proceed?

janmeier:
Make the test mysql only - the other dialects are failing when trying to insert `HORSE` ;)

manchicken:
Seems so obvious when you say it. :-)

manchicken:
@janmeier: are there any next steps?

janmeier:
Nothing other than for me to press the big, green merge button - thanks for your time :) !

manchicken:
Thank you very much. Let me know if I can help the project with something
else.

PR #5905, opened 2016-05-17
BIGINT converted to strings when required (mysql)

### LINKED ISSUE #5176: BIGINT gets truncated
Accordign to the [Changelog](https://github.com/sequelize/sequelize/blob/master/changelog.md#v160), bigints are supposed to be treated as a string, which is one good way to work around Javascript's inability to store [accurate 64-bit integers](https://dev.twitter.com/overview/api/twitter-ids-json-and-snowflake). Another way would be to use a module that implements 64-bit ints with native code, like [node-int64-native](https://github.com/candu/node-int64-native) or [node-int64](https://github.com/broofa/node-int64).

However, the current implementation turns BIGINT fields into Numbers when loading them from the DB, truncating them. The program below demonstrates the problem. The example number 90071992547409921 gets truncated to 90071992547409920 (note the last digit) when represented as a Number. Saving the value into MySQL works correctly, but it is truncated into a Number on read.

```
var sequelize = new Sequelize('bigintz', 'bigintz', '');

  bid: Sequelize.BIGINT
});

sequelize.sync().then(function() {
  var u = User.build();
  u.bid = '90071992547409921'
  u.save().then(function() {
    return User.find({
      where: {
        bid: u.bid
      }
    });
  }).then(function(u2) {
    console.log('u2', u2.dataValues);
    console.log('typeof bid', typeof u2.bid);
  }).catch(function(err) {
    console.log(err);
  });
});
```

```
$ node bigint-test.js
Executing (default): INSERT INTO `Users` (`id`,`bid`,`updatedAt`,`createdAt`) VALUES (DEFAULT,'90071992547409921','2016-01-09 19:09:28','2016-01-09 19:09:28');
Executing (default): SELECT `id`, `bid`, `createdAt`, `updatedAt` FROM `Users` AS `User` WHERE `User`.`bid` = '90071992547409921' LIMIT 1;
u2 { id: 1,
  bid: 90071992547409920,
  createdAt: Sat Jan 09 2016 11:00:14 GMT-0800 (PST),
  updatedAt: Sat Jan 09 2016 11:00:14 GMT-0800 (PST) }
typeof bid number

$ mysql -u bigintz -D bigintz -e 'select bid from Users limit 1;'
+-------------------+
| bid               |
+-------------------+
| 90071992547409921 |
+-------------------+

$ mysql -u bigintz -D bigintz -e 'describe Users;'
+-----------+------------+------+-----+---------+----------------+
| Field     | Type       | Null | Key | Default | Extra          |
+-----------+------------+------+-----+---------+----------------+
| id        | int(11)    | NO   | PRI | NULL    | auto_increment |
| bid       | bigint(20) | YES  |     | NULL    |                |
| createdAt | datetime   | NO   |     | NULL    |                |
| updatedAt | datetime   | NO   |     | NULL    |                |
+-----------+------------+------+-----+---------+----------------+
```

sushantdhiman:
### Pull Request check-list

### Description of change

Fixes #5176 

By using the API of `node-mysql` now `mysql` dialect can represent `BIGINT` in string form. In Addition to that this conversion will only happen when number is too big for JS to handle.

So this change will not effect small numbers. If a number gets big enough to get truncated we will convert it to `string` to save valid number.

cc @janmeier @mickhansen

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,7 +5,8 @@
 - [CHANGED] Throw `bluebird.AggregateError` instead of array from `bulkCreate` when validation fails
 - [FIXED] `$notIn: []` is now converted to `NOT IN (NULL)`  [#4859](https://github.com/sequelize/sequelize/issues/4859)
 - [FIXED] Add `raw` support to `instance.get()` [#5815](https://github.com/sequelize/sequelize/issues/5815)
-- [ADDED] Compare deletedAt against current timestamp when using paranoid [#5880](https://github.com/sequelize/sequelize/pull/5880)
+- [ADDED] Compare `deletedAt` against current timestamp when using paranoid [#5880](https://github.com/sequelize/sequelize/pull/5880)
+- [FIXED] `BIGINT` gets truncated [#5176](https://github.com/sequelize/sequelize/issues/5176)
 
 ## BC breaks:
 - `hookValidate` removed in favor of `validate` with `hooks: true | false`. `validate` returns a promise which is rejected if validation fails
@@ -13,6 +14,7 @@
 - Remove default dialect
 - When `bulkCreate` is rejected because of validation failure it throws a `bluebird.AggregateError` instead of an array. This object is an array-like so length and index access will still work, but `instanceof` array will not
 - `$notIn: []` will now match all rows instead of none
+- (MySQL) `BIGINT` now gets converted to string when number is too big
 
 # 3.23.2
 - [FIXED] Type validation now works with non-strings due to updated validator@5.0.0 [#5861](https://github.com/sequelize/sequelize/pull/5861)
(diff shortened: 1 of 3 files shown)


changelog.md
```diff
@@ -5,6 +5,7 @@
 - [CHANGED] Throw `bluebird.AggregateError` instead of array from `bulkCreate` when validation fails
 - [FIXED] `$notIn: []` is now converted to `NOT IN (NULL)`  [#4859](https://github.com/sequelize/sequelize/issues/4859)
 - [FIXED] Add `raw` support to `instance.get()` [#5815](https://github.com/sequelize/sequelize/issues/5815)
+- [FIXED] BIGINT gets truncated [#5176](https://github.com/sequelize/sequelize/issues/5176)
```
janmeier: Mind adding a line to BC breaks below as well?

janmeier:
LGTM, just a comment to changelog, and needs a rebase

sushantdhiman:
Its ready for merge :) @janmeier

janmeier:
Sweet :)

Issue #3458, opened 2015-04-04
Field with default value not saved on create when using the "fields" option.


vanwalj:
Hello, as described in the title:

If i define a model with a default value for example in my case a UUID
```js
var Media = sequelize.define('Media', {
        uuid: {
            type: DataTypes.UUID,
            defaultValue: DataTypes.UUIDV4,
            primaryKey: true,
            allowNull: false
        },
        someValue: DataTypes.STRING
});
```
Then, if i create my entity using the fields option
```js
Media.create({someValue: 'some value'}, {
            fields: ['someValue']
});
```
The field uuid is not "inserted"
```sql
INSERT INTO `Media` (`someValue`,`updatedAt`,`createdAt`) VALUES ('some value','2015-04-04 05:41:14','2015-04-04 05:41:14');
```
But if i set my "defaultValue" field into the fields option 
```js
Media.create({someValue: 'some value'}, {
            fields: ['uuid', 'someValue']
});
```
Everything "works fine"
```sql
INSERT INTO `Media` (`uuid`,`someValue`,`updatedAt`,`createdAt`) VALUES ('711544ee-691e-4f73-a9d5-9ef11273fa65','some value','2015-04-04 05:41:14','2015-04-04 05:41:14');
```

Using Sequelize 2.0.5 with MariaDB.

mickhansen:
Appears to be a bug. Code doesn't take into account non AI PKs for this case.

lumberj:
I'm seeing this issue on 2.1.3 with sqlite

mickhansen:
@lumberj that version should have the fix, you'll need to provide more detail.

PR #5909, opened 2016-05-17
Fixed bug defaultValue was not used when attribute was set in fields …

dominik-th:
### Pull Request check-list
- [x] Does `npm run test-mariadb` pass with this change (including linting)?
### Description of change

Similar to issue [#3458](https://github.com/sequelize/sequelize/issues/3458) but when you specified the attribute even though it is not in fields array it was set to null.

```
var Media = sequelize.define('Media', {
        uuid: {
            type: DataTypes.UUID,
            primaryKey: true,
            allowNull: false
        },
        someValue: DataTypes.STRING
});

Media.create({uuid: 'some value', someValue: 'some value'}, {
            fields: ['someValue']
});
```

Would result into a `notNull Violation: uuid cannot be null`

### DIFF
--- a/test/integration/model/create.test.js
+++ b/test/integration/model/create.test.js
@@ -1203,27 +1203,6 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       });
     });
 
-    it('should use default value if attribute is excluded in fields option', function() {
-      var User = this.sequelize.define('User', {
-        username: DataTypes.STRING,
-        taskCount: {
-          type: DataTypes.INTEGER,
-          defaultValue: 2
-        }
-      });
-
-      return User.sync().then(function() {
-        return User.create({
-          username: 'John',
-          taskCount: 5
-        }, {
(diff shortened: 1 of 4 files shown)


test/integration/model/create.test.js
```diff
@@ -1203,6 +1203,27 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       });
     });
 
+    it('should use default value if attribute is excluded in fields option', function() {
+      var User = this.sequelize.define('User', {
```
janmeier: This could be written as a unit test (`User.build()`, and check that `get` returns the right value

janmeier:
Ping @dominik-th

dominik-th:
Sorry, I must have missed the previous email! :(

I noticed that the master repo is already far ahead, hopefully this is not too much of a problem, otherwise I could try to create a new pull request based on a later version.

PR #5916, opened 2016-05-18
feat(model): put model in options for hook access

scragg0x:
### Description of change

Use case for this is generic cache invalidation using the options.model.name and options.where.id or obj.id

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2148,6 +2148,8 @@ Model.prototype.bulkCreate = function(records, options) {
     }
     options.updateOnDuplicate = updatableFields;
   }
+  
+  options.model = this;
 
   var self = this
     , createdAtAttr = this._timestampAttributes.createdAt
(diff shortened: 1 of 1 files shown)


sushantdhiman:
If you want to access model inside the hook you can 

```js
.beforeCreate(function(instance){
   instance.Model.find(); 
)
```

scragg0x:
I mean for the global hooks. You don't get passed the instance.

sushantdhiman:
Can't you just `require` the model in that code. `config` object is cloned much often in our code base, It can turn out to be a pretty big performance hit.

scragg0x:
It's just including a reference to the model, don't see how it will affect
performance much.  It was explicitly included in one of the hooks already.
It was also there until recently for bulk update until a commit broke as an
unintentional side effect.

janmeier:
Thanks :)

janmeier:
@sushantdhiman I saw you posted a comment about performance earlier - model instances are not cloned, [see here](https://github.com/sequelize/sequelize/blob/3d14a4a63b2b8ecc517cb03e8d4ed59b1d177132/lib/utils.js#L98) so it won't affect performance

sushantdhiman:
:checkered_flag: great @janmeier, Thanks for contributing @scragg0x

PR #5918, opened 2016-05-18
chore(ci): Update build matrix

janmeier:
This is by no means final, I just wanted to experiment a bit :)

I've thought about moving to coveralls for some time now, since we don't use any of the extra features that codeclimate provides. And when I noticed that coveralls supports merging coverage I was sold! This means we no longer need to run a separate job for coverage, we can just enable coverage for the jobs for one of the node versions are have coveralls merge it. I've also added the unit tests to the coverage results.
The travis config is a bit uglier now (I didn't find a good way to express that we should only run coverage on node 6) but other than that, it's lovely :).

Proposals to further cut down build time:
* Remove the mariadb dialect - It's just a clone of mysql (it doesn't use node-mariasql anymore)
* Remove postgres-native (and add a couple of integration tests to check that passing `native: true` works


sushantdhiman:
Great stuff :)

- We should remove mariadb because we not using its binding. I don't think we are even hitting a real mariadb in CI.
- Yes we should only add few `native:true` integration tests and remove `postgres-native`.

Other interesting point, overall build time has decreased from [23 min](https://travis-ci.org/sequelize/sequelize/builds/131136127) to [16 min](https://travis-ci.org/sequelize/sequelize/builds/131218973), Thats a really good thing (I hate waiting for CI). But individual dialect test time has increased (by about 30+ sec), may be due to new Node versions.

Doing tests for `native:true` will take code change but we can remove `mariadb` for now to reduce build time even further.

janmeier:
Regarding individual test time - 30 seconds is nothing in travis-time - Test times can fluctutate by a minute ;)

janmeier:
Without maria we are down to 14 minutes

.travis.yml
```diff
@@ -1,24 +1,37 @@
...
-  - DIALECT=postgres-native
-  - DIALECT=sqlite
-  - DIALECT=mariadb
-  - DIALECT=mssql
+  global:
+    - CC=clang
+    - CXX=clang++
+    - npm_config_clang=1
+    - COVERALLS_REPO_TOKEN=751SiA6QhY5EH4P2ygq7JFWulqiD2ETXO
+    - COVERALS_PARALLEL=true
```
overlookmotel: Typo here. `COVERALLS` is missing an `L`.

overlookmotel:
Is it worthwhile testing against node v6 as well?

As a general point, how often (if ever) does a build pass on one version of node and fail on another? I've never run into that with Sequelize myself. If it's super-rare then maybe testing against loads of node versions is overkill.

overlookmotel:
Is this intended to go into Sequelize v4.0 or for immediate merge?

sushantdhiman:
@janmeier It looks like we are down to [11 min](https://travis-ci.org/sequelize/sequelize/builds/131233272) :clap: :clap: 

Lets merge it in master for now. `v3` still needs to be compatible with old version `0.10` and `I/O JS` so need to change there.

.travis.yml
janmeier: Hmm, this should be the variable that tells coveralls to expect multiple coverage reports - but seems to work fine without it https://coveralls.io/builds/6236971

janmeier:
> Is it worthwhile testing against node v6 as well?
As a general point, how often (if ever) does a build pass on one version of node and fail on another? I've never run into that with Sequelize myself

Very seldom indeed. Perhaps we should test again latest (6) and oldest LTS (0.12 https://github.com/nodejs/lts)

> Is this intended to go into Sequelize v4.0 or for immediate merge?

I've branched of v3 as a separate branch, and intend to make the next master release a 4.0 pre-release - There are already some minor BC breaks which have been merged https://github.com/sequelize/sequelize/blob/master/changelog.md#bc-breaks. So the answer to both is yes :)

I don't expect we'll be releasing 4.0 final before a couple of months at least

overlookmotel:
@janmeier OK cool. Let the breaking commence! :)

.travis.yml
overlookmotel: Maybe best to correct it anyway and see what happens?

mickhansen:
@overlookmotel We will likely introduce some ES6 code in 4.0, we're considering moving to testing only Node 6 (but we likely cant ignore Node 4 since it's an LTS)

mickhansen:
LGTM, needs a rebase.

overlookmotel:
Any way to get Travis to run a small set of tests on PRs and merges into master and then have a larger test suite (running against additional versions of node and perhaps with the test shims enabled too) which is only run when moving to make a release?

Maybe creating a `release` branch which you only push to when you're preparing a release, and which Travis is programmed to run a wider set of tests on...

That might be a decent compromise, cutting down test run times in general, but also ensuring that what's on NPM is fully tested. You wouldn't expect the full test suite to fail more than *really* occasionally, but still it would catch any weird edge cases before they're released.

No idea if Travis supports this though.

mickhansen:
@overlookmotel That could possibly work yeah, but i'm of the philosophy that master should always be ready for release.

It doesn't change the fact that we'd like to introduce ES6 features into the project though :)

overlookmotel:
@mickhansen I totally agree! But if that makes for build times which impede productivity, I guess there's a balance to be struck.

Given that node v0.12 ends LTS at end of this year, I guess now could be a good time to abandon it, allowing for ES6 in the codebase. Please please please can we write in co-routines now?? Those who are still on node v0.12 (including me for now, sadly) can always stick with Sequelize v3 until they update node.

mickhansen:
@overlookmotel We're looking at bluebird coroutines - But i don't have an overview of the performance degration (if any) in using them.

overlookmotel:
I don't have formal numbers on that. But a friend did a rough benchmark. He found that co-routines (using [co-bluebird](https://www.npmjs.com/package/co-bluebird) rather than bluebird's `coroutine()` method) had minimal performance impact compared to bluebird promises.

In fact, on some of his tests, co-routines actually came out faster. This is a bit bizarre, but I suppose could be explained by co-routines only running the next section of code *once the previous promise resolves* rather than constructing a big chain of promises and then resolving them all in series. So co-routines are more "lazy" in a sense.

But anyway, the gist was that performance wasn't impacted significantly. Given bluebird's emphasis on performance, bluebird co-routines are probably even faster.

There's some benchmarks here: http://bluebirdjs.com/docs/benchmarks.html

mickhansen:
Slightly slower in parallel and slightly faster in sequence, but most likely negligble in both cases.

janmeier:
With the removal of 0.12 our build time is down to 10½ minute :tada: !

Looks like we'll definitely be using coroutines in 4.0! :D

Issue #5919, opened 2016-05-18
Dollar prefixes on methods


felixfbecker:
When working with the code and also when debugging I wondered why Sequelize mixes dollar, underscore and double underscore prefixes for methods and what their meaning are. The convention seems to be to use underscore for private methods / properties, and dollar should imo be buried with jQuery. Sometimes Sequelize even assigns the same method with both a dollar and an underscore. Maybe this can be cleaned up for 4.0?

janmeier:
They are all meant to signify private methods - So this could definitely be clean up! :)

mickhansen:
I actually prefer $ for privates rather than _, but nothing i care too deeply about.
We should probably pick one thing and go with that atleast :)

felixfbecker:
@mickhansen from my experience underscore is the convention and used in all great javascript libraries, so I would vote for that.

mickhansen:
@felixfbecker You're probably right about that, don't know where i found $ originally.

PR #5924, opened 2016-05-19
Make instances instanceof Model, with ES6 classes

felixfbecker:
This PR is a reimplementation of #5877 with ES6 classes. Reasoning is performance and that 4.0 will probably require Node >=4 or even 6.


felixfbecker:
Apparently native modules do not compile, any help?

sushantdhiman:
Try adding the `addons` for `g++` and `c++`, something like

```yml
addons:
    apt:
     sources:
       - llvm-toolchain-precise-3.6
       - ubuntu-toolchain-r-test
     packages:
       - clang-3.6
       - g++-4.8
```

felixfbecker:
@sushantdhiman didnt work :/

mickhansen:
diff too large :(

mickhansen:
Can you rebase this? Might help with the tests

janmeier:
Then we promise to review it quickly! :)

felixfbecker:
I will try, hopefully there were no changes in model.js or instance.js... It's a pain to rebase this because for git it's just one huge delete and one huge insert.

mickhansen:
@felixfbecker is there a document describing your design and thoughts behind this anywhere? Syntax examples etc.

felixfbecker:
@mickhansen I outlined my thoughts in the initial PR #5877

mickhansen:
@felixfbecker Thanks i've looked through a bit of it.
Attributes, name, etc would have to be statics, so with es7 `static` and with es6 simply add it as a property after the fact (which could be done by define()).

felixfbecker:
I think I cought everything. Had to essentially reimplement every single commit on master since my PR.

@mickhansen everything that was on the prototype of `Model` is now a static. The name is set by `sequelize.define()`, and if you subclass `Model` yourself, the name will be the name you give your class.

mickhansen:
@felixfbecker We'd need a way to overwrite name, cause i use lowercase model names but i'd like pascal case class names.

We don't need to main `.Instance` for BC.

lib/model.js
```diff
@@ -713,8 +710,8 @@ class Model {
 
     this.options.indexes = this.options.indexes.map(this.$conformIndex);
 
-    this.Instance.prototype.$Model =
-    this.Instance.prototype.Model = this;
+    this.prototype.$Model =
+    this.prototype.Model = this;
```
mickhansen: This can be removed completely.

lib/model.js
```diff
@@ -40,6 +40,10 @@ var Utils = require('./utils')
  */
 class Model {
 
+  static get Instance() {
+    return this;
+  }
```
mickhansen: This can be removed completely.

mickhansen:
I assume you made no changes when merging Instance/Model other than doing a single file and doing all Model methods as static, correct?

What is the syntax for defining attributes, model name, table name, underscored etc, with `extends Sequelize.Model`?

mickhansen:
Ideally i would love to work on a doc describing how to upgrade from ES5 model to both ES6 and ES7 models (defining needed properties), this doc could be exploratory and we could work from that.

mickhansen:
But we could likely merge this for the sake of no-rebases and then move forward with the design after, as long as we arent cutting releases theres room to maneuver.

lib/model.js
felixfbecker: I added it because I didn't touch any other files and didn't know wether the tests would pass if I remove it (because it might be used in other files). I'm happy to remove it, you can always access the model class through `.constructor`

mickhansen: Leave it for now, let's make the goal to get this PR merged asap so we dont have to rebase.
And then work on removing all cruft etc.

felixfbecker:
@mickhansen 
> We'd need a way to overwrite name, cause i use lowercase model names but i'd like pascal case class names.

We could pass the model name to `Model.init()`. But this of course will override the native `name` property on classes, so `YourClass.name === 'yourClass'`. I find it a bit redundant to have to pass the name everytime, like:
```ts
class User extends Sequelize.Model {}
Model.init(User.name, {}, {}, sequelize.modelManager)
```
Then again, `Model.init()` would be more low-level API for decorator libraries and I would expect ES5/6 users to continue to use `sequelize.define()`. For `define()` you can of course always do `const User = sequelize.define('user', {...})`

> I assume you made no changes when merging Instance/Model other than doing a single file and doing all Model methods as static, correct?

Basically yes. Plus I changed `sequelize.define()` and all `instanceof Model` checks I could find.

> What is the syntax for defining attributes, model name, table name, underscored etc, with  extends Sequelize.Model ?

`Model.init()` is the static constructor of the model classes. It takes the attributes, options and the modelManager. I'm a bit unhappy with the modelManager as the method does nothing with it as set `this.modelManager = modelManager`. Is this needed? I would prefer to make the Model not know about the modelManager and add the model with something like `sequelize.addModel(Model)`.

> Ideally i would love to work on a doc describing how to upgrade from ES5 model to both ES6 and ES7 models (defining needed properties), this doc could be exploratory and we could work from that.

So you would make `Model.init()` a high-level API? I would propose a section in the docs about how to use sequelize in different languages, a guide for the old style with `define()`, a guide for ES6 with `Model.init()`, and a guide for TypeScript users with decorators etc. I expect many users to stay on the `define()` style though because atm ES6 classes don't provide _that_ much of a benefit. The potential really lays in decorators.

> But we could likely merge this for the sake of no-rebases and then move forward with the design after, as long as we arent cutting releases theres room to maneuver.

That would be nice, so other PRs can target the class implementation. So it is decided that Node 4/6 is required (for classes)? Do you want me to squash or keep it in multiple commits?

I would then go ahead and refactor other parts of the library to classes and potentially arrow functions.
Also, I would take on #5919 and remove (old?) $ methods and aliases.

mickhansen:
> YourClass.name === 'yourClass'
Should be sufficient for now.

> I would prefer to make the Model not know about the modelManager and add the model with something like sequelize.addModel(Model).
That would be the goal, addModel would then return a model aswell, since we need a version of the model that is aware of sequelize.

> So you would make Model.init() a high-level API
No, model.init is a mapper for define (imo), i'd look to a better design for ES6 and a design for ES7 we could strive to in the future.

The design could be ES7 decorators that would work for ES6 aswell (although without the decorator syntax), or it could be `Model.attributes = {}`.

I agree that most of the power will come with static props and decorators, but i think it makes sense to have an intermediate design aswell (since static props and decorators need to map to that in any case).

We've pretty much decided on Node 4.0+ atleast, and i don't think we'll go for Node 6.0+ only.

I'll squash when i merge.

felixfbecker:
@mickhansen
> That would be the goal, addModel would then return a model aswell, since we need a version of the model that is aware of sequelize.

I don't know what you mean with `addModel()` needs to return a model, I thought of something like this:
```js
  public username: string;
  myMethod() {
    return 'hello';
  }
}
User.init({username: Sequelize.STRING}, {tableName: 'users'}) // attach metadata, this can alternatively be done with decorators
sequelize.addModel(Model) // adds the model to the modelManager and sets User.modelManager
```

> No, model.init is a mapper for define (imo), i'd look to a better design for ES6 and a design for ES7 we could strive to in the future.
> 
> The design could be ES7 decorators that would work for ES6 aswell (although without the decorator syntax), or it could be  Model.attributes = {} .

The idea is that `Model.init()` is simply a static constructor and any decorator can invoke that to attach the metadata. A class decorator in ES7 simply takes the class and the attributes and options and calls `init()`. In TypeScript, property decorators can assign metadata on the class with symbols (and we can also get some type metadata automatically through this). The class decorator can then scan the class for these symbols and fill out the `attributes` object. So I would propose that `Model.init()` is that intermediate API and decorators can be implemented in third-party modules.

`define()` is a _factory_ that creates a new subclass of Model, sets a name on it, calls `Model.init()` with the attributes and options and returns it. It is nice because you can define a Model in one function call, but the con is that you don't get any type hints from your editor for that class because it is dynamically created and not declared.

felixfbecker:
How can the build fail because of style issues (mixed double/single quotes) if I only copy+pasted code?

mickhansen:
> I don't know what you mean with addModel() needs to return a model, I thought of something like this:

If we're going to be doing ES6 classes they should be decoupled.
Calling `addModel` should not affect the base instance since it should ideally be possible to use the same model for multiple connections.

> A class decorator in ES7 simply takes the class and the attributes and options and calls init()

I thought all decorators were basically just functions?

felixfbecker:
> If we're going to be doing ES6 classes they should be decoupled.
 Calling  addModel  should not affect the base instance since it should ideally be possible to use the same model for multiple connections.

If you want to support that scenario... Imo it should also be supported that I only define my class, _set_ the connection/modelManager whatever that should be used on it and then use this class directly. I do not want to use the return value of some function as my model class... Eg it should be `export class MyModel` and not `export const MyModel = sequelize.addModel(MyModelClass)`.

I think having one sequlize connection instance is the most common use case. Cloning the model class seems overkill to me to implement this feature and is tricky to do in JS (a class is just a function). I think if people want to use different connections with one model class, there should be an option to pass it to every function that needs a connection, eg `find()`, `save()`, ...
So you could either a) set one (default) sequelize connection on the model to be used or b) pass it as an option to the methods.

felixfbecker:
> I thought all decorators were basically just functions?

Yes. A sequelize Model class decorator could look like this:
```js
function ModelData(attributes, options) {
  return model => {
    model.init(attributes, options)
  }
}

@ModelData({
  username: Sequelize.STRING
}, {
  tableName: 'users'
})
  // custom methods, getter, setter etc
}
```

felixfbecker:
I rebased but Travis doesn't trigger... Can someone trigger a build manually?

mickhansen:
> I think having one sequlize connection instance is the most common use case

It is the most common, but multiple connections is not uncomming, especially in enterprise migration scenarios.

ES6 classes provide the option of decoupling - And that's definitely something we want, both for testability and other usecases.

Sequelize addModel could something like:

```
class ConnectionClass extends InputClass;
ConnectionClass.sequelize = this;
return ConnectionClass;
```

But i realize it might no inherit class properties in this case.

> Yes. A ES7 sequelize Model class decorator could look like this:

Then i would provide decorators for most common options, and just support class props for ES6 (possibly via a init function aswell).

felixfbecker:
That subclass-per-connection would get very messy imo:
```ts
class UserClass extends Sequelize.Model {
}
export const User = sequelize.addModel(UserClass)
```
`sequlize.addModel` would have to redefine `ConnectionClass.name` and I would rather like to export the actual class and only use that. What do you think of being able to (optionally) set the connection as last param in `init()` _or_ passing it to the individual model functions, similar to how you can pass a transaction? That would be a nice inversion of dependency&trade;. And you could still create subclasses that use different connections urself.

> Then i would provide decorators for most common options, and just support class props for ES6 (possibly via a init function aswell).

I am all for a simple configuration approach where you just set `attributes` on the class (or maybe a Symbol?). But the init function seems to do a whole lot more that I don't understand. Maybe you can take a look at what is actually needed in there and if we could maybe even get rid of it?

felixfbecker:
And how do I get Travis to run a build?

mickhansen:
> sequlize.addModel would have to redefine ConnectionClass.name and I would rather like to export the actual class and only use that. What do you think of being able to (optionally) set the connection as last param in init() or passing it to the individual model functions, similar to how you can pass a transaction? That would be a nice inversion of dependency™. And you could still create subclasses that use different connections urself.

If we do something like `Model.sequelize = sequelize` (which would be set by init() in some cases) then it would be trivial for users to make their own subclasses of classes and attach sequelize instances manually, that would be fine with me yeah.

>  am all for a simple configuration approach where you just set attributes on the class (or maybe a Symbol?). But the init function seems to do a whole lot more that I don't understand. Maybe you can take a look at what is actually needed in there and if we could maybe even get rid of it?

"sometime" before executing queries the model will need to be normalized, this could be via an init() function that could be called manually or by all calls.

mickhansen:
Tests have run: https://travis-ci.org/sequelize/sequelize/jobs/133009285 - Some failures in 1:M

felixfbecker:
@mickhansen I'm getting closer to the finish line (please don't commit on master in the meanwhile, I don't want to rebase again...).

Currently I have the problem that `runHooks()` supports a callback as last parameter:
```js
    if (typeof fnArgs[fnArgs.length - 1] === 'function') {
      fn = fnArgs.pop();
    }
```
however, the `afterDefine` hook it is called with the model as argument, which now is a function (class). As such, Sequelize thinks it is a callback. Should I
 - remove callback support
 - make a check for `.prototype instanceof Model` - that would create a cyclic dependency between hooks.js and model.js
   - could first export Model, then require and apply hooks
   - could require model.js late inside the the runHooks function
 - [check if the argument is a class by examining `toString()`](https://github.com/miguelmota/is-class/blob/master/is-class.js)
?

janmeier:
Don't worry, mick and I have already discussed keeping master clean until this has landed ;)

Please remove callback support :)

felixfbecker:
Stuck at this failing test, any idea?
```
1) [MYSQL] Include find should support a simple nested hasMany <-> hasMany include:

4027     SequelizeDatabaseError: ER_BAD_FIELD_ERROR: Unknown column 'Object.id' in 'order clause'

4028      at Query.formatError (lib/dialects/mysql/query.js:211:14)

4029      at Query._callback (lib/dialects/mysql/query.js:51:21)

4030      at Query.Sequence.end (node_modules/mysql/lib/protocol/sequences/Sequence.js:96:24)

4031      at Query.ErrorPacket (node_modules/mysql/lib/protocol/sequences/Query.js:94:8)

4032      at Protocol._parsePacket (node_modules/mysql/lib/protocol/Protocol.js:280:23)

4033      at Parser.write (node_modules/mysql/lib/protocol/Parser.js:73:12)

4034      at Protocol.write (node_modules/mysql/lib/protocol/Protocol.js:39:16)

4035      at Socket.<anonymous> (node_modules/mysql/lib/Connection.js:96:28)

4036      at readableAddChunk (_stream_readable.js:153:18)

4037      at Socket.Readable.push (_stream_readable.js:111:10)

4038      at TCP.onread (net.js:531:20)

4039  From previous event:

4040      at Query.run (lib/dialects/mysql/query.js:41:17)

4041      at lib/sequelize.js:837:20

4042      at node_modules/retry-as-promised/index.js:38:21

4043  From previous event:

4044      at retryAsPromised (node_modules/retry-as-promised/index.js:27:10)

4045      at lib/sequelize.js:836:12

4046  From previous event:

4047      at Promise.then (lib/promise.js:21:17)

4048      at Context.<anonymous> (test/integration/include.test.js:427:10)

```

mickhansen:
@felixfbecker Doesn't sound related to these changes. Can you post the SQL + the test?

felixfbecker:
Its the latest build for this PR. But it can be related to these changes because of the "instanceof Model" checks that are done against the include options. I just dont see where I missed one. Could you review my latest commit?

mickhansen:
I'm seing a failure for: https://github.com/felixfbecker/sequelize/blob/1681f0b39816521419683a5b49326cc4a34d459f/lib/model.js#L1784

Starts here: https://github.com/sequelize/sequelize/blob/master/lib/associations/has-many.js#L479
And build called here: https://github.com/felixfbecker/sequelize/blob/1681f0b39816521419683a5b49326cc4a34d459f/lib/model.js#L2489

Not sure if `self` is incorrect somehow.

I have no idea if `new this` is valid code, but it would be failing in a lot more cases if it was.

felixfbecker:
@mickhansen I already fixed that one in a3e573a.
Current issue is https://travis-ci.org/sequelize/sequelize/jobs/133071921

felixfbecker:
It's the nested include test that fails across all dialects:
```js
        return User.find({
          where: {
            id: 1
          },
          include: [
            {model: Product, include: [
              {model: Tag}
            ]}
          ],
          order: [
            User.rawAttributes.id,
            [Product, 'id']
          ]
        });
```

janmeier:
Its a problem in the order clause, this is the original code https://github.com/sequelize/sequelize/blob/738ea1b2058a91fe8217db4088df7f5a4e1baf77/lib/dialects/abstract/query-generator.js#L826 -

felixfbecker:
I changed it in https://github.com/sequelize/sequelize/pull/5924/commits/1681f0b39816521419683a5b49326cc4a34d459f, dont see the error though

felixfbecker:
Alright, I replaced `.Model` with `.constructor` somewhere where I thought the parameter was a Model, but apparently it's just an object with a `Model` property

felixfbecker:
All tests pass  🎉

felixfbecker:
Will you squash this all into a huge monocommit?

janmeier:
If you want I think it would be fine to split this into smaller commits, but if not, squashing is fine by me as well

janmeier:
I'm hoping to do a final review and merge over the weekend :)

test/integration/associations/belongs-to-many.test.js
```diff
@@ -632,7 +632,7 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
         this.task = task;
         return task.createUser({ username: 'foo' });
       }).then(function(createdUser) {
-        expect(createdUser.Model).to.equal(User);
+        expect(createdUser.constructor).to.equal(User);
```
janmeier: Minor thing, but couldn't this be instanceof?

felixfbecker: yes.

janmeier:
@felixfbecker I've reviewed the changes to model.js locally, so I'm ready to merge - Do you want to do the squashing dance, or do you want me to mono-squash it?

felixfbecker:
@janmeier I cannot think of a good way to squash this into more than one commit, so go ahead :)

janmeier:
Thanks for the big effort @felixfbecker  :tada: :tada: :tada:

sushantdhiman:
@felixfbecker gz

mickhansen:
Yay! Thanks a bunch @felixfbecker

nervgh:
Is there docs, how to use `extends Sequelize.Model ` in the `4.0.0.-0` version?

I have this code:
```js
// schema is sequelize instance

let Model = schema.define('news', {
  title: {
    type: Sequelize.STRING,
    defaultValue: null
  },
  content: {
    type: Sequelize.TEXT,
    allowNull: true
  }
});
```
How I could specify:
1. table' name
2. table' columns
```js
class Model extends Sequelize.Model {
}
```
? What I need to do?

Thanks.

felixfbecker:
@nervgh There is no documentation yet because the API is not finished and will probably change a lot. Currently you have to call
```js
Model.init(attributes, options, sequelize.modelManager);
sequelize.modelManager.add(Model)
```
`options` also needs to include `sequelize`.

Please note though that `sequelize.define()` still works.

nervgh:
> Please note though that sequelize.define() still works.

Yes, I know, I do it. But I want use inheritance for overriding a couple of built-in methods:

```js
class Model extends Sequelize.Model {
  save(options) {
    // my code here
    return super.save(options);
  }
}
```

@felixfbecker thank you :v: I will try your recommendations.

felixfbecker:
@nervgh You can do that with `define()` too by assigning to the prototype. `define()` returns a subclass of `Model`.

OLDIN:
@nervgh 

    // user.js 
    export default class User {

      constructor(sequelize, types) {

        return sequelize.define("users", {
          steamid: types.STRING(20),
          name: types.STRING(30),
          avatar: types.STRING(30),
          coins: types.INTEGER,
          permissions: types.INTEGER(1),
          banned: types.BOOLEAN
        });

      }

    }

    // index.js
    import userModel from './models/user';

    const User = sequelize.import('users', function(sequelize, DataTypes) {
      return new userModel(sequelize, Sequelize.DataTypes);
    })

felixfbecker:
@OLDIN that is completely wrong and has nothing to do with the new syntax. A constructor does not have a return value.

This is a very old PR. If you want to discuss anything, please move the discussion to Slack, Gitter or a new issue.

OLDIN:
@felixfbecker offer your working version. I will be glad to use.

felixfbecker:
@OLDIN I can point you to the TypeScript definitions, which contain some examples: http://typed-sequelize.surge.sh/v4/

I'm gonna lock this thread now to prevent spamming participants with notifications

PR #5925, opened 2016-05-19
Task 5248, Refactor methods which use fields and options

### LINKED ISSUE #5248: All instances of fieldsOrOptions should be refactored to only support options

sushantdhiman:
### Pull Request check-list

- [x] Is a documentation update included ?

### Description of change

Closes #5248

It appears most of methods are already using the `options` only approach. This PR refactors these methods

- [x] `increment`, `(fields, options)` to `(options)`
- [x] `decrement`, `(fields, options)` to `(options)`

Although `aggregate` methods like `max`, `min` etc also follow the signature like `(field, options)` but I have decided to not change their syntax, because its fairly common to use `User.max('id')` or `User.min('age')`. Same argument can be applied to `increment` and `decrement` as well, but for `aggregate` its more common I think.

Let me know if we need to refactor `aggregate` methods as well.

cc @janmeier

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,7 @@
 - [REMOVED] MariaDB dialect
 - [FIXED] `hasOne` now prefer aliases to construct foreign key [#5247](https://github.com/sequelize/sequelize/issues/5247)
 - [CHANGED] `instance.equals` now only checks primary keys, instead of all attributes.
+- [CHANGED] Refactor methods to use options only [#5248](https://github.com/sequelize/sequelize/issues/5248)
 
 ## BC breaks:
 - `hookValidate` removed in favor of `validate` with `hooks: true | false`. `validate` returns a promise which is rejected if validation fails
@@ -27,6 +28,7 @@
 - Removed MariaDB dialect - this was just a thin wrapper around MySQL, so using `dialect: 'mysql'` instead should work with no further changes
 - `hasOne` now prefer `as` option to generate foreign key name, otherwise it defaults to source model name
 - `instance.equals` now provides reference equality (do two instances refer to the same row, i.e. are their primary key(s) equal). Use `instance.get()` to get and compare all values.
+- Changed signature for `increment`/`decrement` to only use `options`. Use `options.fields` to pass the `fields`
 
 # 3.23.2
 - [FIXED] Type validation now works with non-strings due to updated validator@5.0.0 [#5861](https://github.com/sequelize/sequelize/pull/5861)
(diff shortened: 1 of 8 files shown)


sushantdhiman:
@janmeier Should I rebase only or you have any comments ?

mickhansen:
#5248 was actually meant for specifically fieldOrOptions.
I don't think the increment API should be changed so that we don't support increment(field) anymore.

mickhansen:
I would be fine with dropping the increment array support tho.

```
increment('field');
increment({fields: [], by: 3})
```

sushantdhiman:
This mean the `aggregate` should be re factored to 

```js
min('field') => min({ field: field });
```
And increment / decrement should still support, although drop support for `[]`

```js
increment('fields') // allowed
increment(['f1', 'f2']) // not allowed
```

sushantdhiman:
I think we should drop support for any `fields` / `field` so

```js
min('field', options)   =>   min(options) // options.field will be used
increment('fields', options)   =>   increment(options) // options.fields will be used
```

mickhansen:
This greatly reduces the usability of both methods.
The 95% usecase for min/increment is a single field with no options.

mickhansen:
The intention of the original task was to refactor methods that had `fieldsOrOptions` as a second argument (`save`, `update`, anything else).

@janmeier

sushantdhiman:
This PR can be closed then, because everything else is already `fieldsOrOptions`

mickhansen:
Everything else being?
Will keep it open for now, have to review the code, see if there's any code left that bugs me.

mickhansen:
Error here can be removed: https://github.com/sequelize/sequelize/blob/master/lib/instance.js#L498

mickhansen:
I do find the code here https://github.com/sequelize/sequelize/blob/master/lib/instance.js#L934 to be incredibly ugly.

We could reduce the case by only supporting string | options, and not array

sushantdhiman:
> Everything else being?

All major methods like `save`, `update`, `destroy` are already using `options` only approach. Only `aggregate` and `increment/decrement` are based on `field` based approach.


> We could reduce the case by only supporting string | options, and not array

Sure


> Error here can be removed: https://github.com/sequelize/sequelize/blob/master/lib/instance.js#L498

Sure

mickhansen:
@sushantdhiman Ah, dont recall why i opened that issue then.

sushantdhiman:
@mickhansen , I am going to revert my changes which removed support for `fields` in `increment/decrement`. Will work on other ends like removing support for `Array` and removing errors.

sushantdhiman:
Closing, will continue cleanup task in different PR

PR #5933, opened 2016-05-20
Model#truncate - Fixing documentation formatting

gtelljohann:
The options.cascade row of the table in Model#truncate seemed to be formatted incorrectly


janmeier:
The API docs are auto-generated from code comments, please update the corresponding comment as well

gtelljohann:
Got it, thanks @janmeier.  Looks like the code had a space in `[options.cascade = false]`, which caused it to parse incorrectly in the docs.

janmeier:
Thanks ! (:

PR #5939, opened 2016-05-21
added support for hasMany counter cache relations where association c…

validkeys:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Have you added an entry under `Future` in the changelog?

_NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open._

### Description of change

The counterCache plugin was not adhering to scoped associations for non-atomic counter operations. This update adds any scope properties defined on the association into the counter query run on non-atomic counterCache updates.

PR #5940, opened 2016-05-21
added support for hasMany counter cache relations where association c…

validkeys:
(this pull request was also made on master and tested on both branches) https://github.com/sequelize/sequelize/pull/5939

### Pull Request check-list

### Description of change

The counterCache plugin was not adhering to scoped associations for non-atomic counter operations. This update adds any scope properties defined on the association into the counter query run on non-atomic counterCache updates.

### DIFF
--- a/lib/plugins/counter-cache.js
+++ b/lib/plugins/counter-cache.js
@@ -3,7 +3,8 @@
 var Utils = require('./../utils')
   , Helpers = require('../associations/helpers')
   , DataTypes = require('../data-types')
-  , Promise = require('bluebird');
+  , Promise = require('bluebird')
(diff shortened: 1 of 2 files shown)


janmeier:
Hmm, not sure if we actually wan't to keep counter-cache around in core for 4.0... What do you think @mickhansen ?

mickhansen:
No, it should be removed.

validkeys:
@mickhansen is counter-cache being moved to an external plugin in 4.0? Or is there some other means to achieve the functionality?

mickhansen:
@validkeys It will be removed, someone else needs to make a plugin :)

PR #5953, opened 2016-05-24
Removed .gitmodules

sushantdhiman:


janmeier:
Yeah, I think we can safely remove that ;)

sushantdhiman:
It has been waiting from 1 whole year :laughing:

PR #5957, opened 2016-05-25
Task 5267, Make all association to prefer `as` their foreign key name

### LINKED ISSUE #5267: Behaviour of alias (as) to foreignKey translations in associations should be synchronized.
Right now the various associations (hasOne, belongsTo, hasMany, etc) treat aliases differently when attempting to create implicit, not defined, foreignKey names. This behaviour should be made consistent across all the different associations, either by using the alias as the foreignkey or not using it or by requiring an explicit foreignKey definition when using an alias as part of an association definition.

sushantdhiman:
### Pull Request check-list

### Description of change

Closes #5267

- [x] `hasOne` https://github.com/sequelize/sequelize/blob/master/lib/associations/has-one.js#L50
- [x] `belongsTo` https://github.com/sequelize/sequelize/blob/master/lib/associations/belongs-to.js#L50
- [x] `hasMany` https://github.com/sequelize/sequelize/blob/master/lib/associations/has-many.js#L75
- [x] `belongsToMany` https://github.com/sequelize/sequelize/blob/master/lib/associations/belongs-to-many.js#L116

cc @mickhansen , @janmeier

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -112,7 +112,7 @@ var BelongsToMany = function(source, target, options) {
     this.foreignKeyAttribute = {};
     this.foreignKey = this.options.foreignKey || Utils.camelizeIf(
       [
-        Utils.underscoredIf(this.source.options.name.singular, this.source.options.underscored),
+        Utils.underscoredIf(this.options.as || this.source.options.name.singular, this.source.options.underscored),
         this.source.primaryKeyAttribute
       ].join('_'),
       !this.source.options.underscored
--- a/lib/associations/has-many.js
+++ b/lib/associations/has-many.js
@@ -71,7 +71,7 @@ var HasMany = function(source, target, options) {
   if (!this.foreignKey) {
     this.foreignKey = Utils.camelizeIf(
       [
-        Utils.underscoredIf(this.source.options.name.singular, this.source.options.underscored),
+        Utils.underscoredIf(this.options.as || this.source.options.name.singular, this.source.options.underscored),
         this.source.primaryKeyAttribute
       ].join('_'),
       !this.source.options.underscored
--- a/changelog.md
+++ b/changelog.md
@@ -17,6 +17,7 @@
 - [CHANGED] `instance.equals` now only checks primary keys, instead of all attributes.
 - [REWRITE] Rewrite model and instance to a single class - instance instanceof Model [#5924](https://github.com/sequelize/sequelize/issues/5924)
 - [REMOVED] Counter cache plugin
+- [FIXED] All associations now prefer aliases to construct foreign key [#5267](https://github.com/sequelize/sequelize/issues/5267)
 
 ## BC breaks:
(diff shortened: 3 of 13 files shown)


lib/associations/belongs-to-many.js
```diff
@@ -113,7 +113,7 @@ var BelongsToMany = function(source, target, options) {
     this.foreignKeyAttribute = {};
     this.foreignKey = this.options.foreignKey || Utils.camelizeIf(
       [
-        Utils.underscoredIf(this.source.options.name.singular, this.source.options.underscored),
+        Utils.underscoredIf(this.options.as || this.source.options.name.singular, this.source.options.underscored),
```
mickhansen: Should this be the singular name of the alias? `WorkersId` is pretty ugly, but we could run into a lot of issues like we've done previously.

@janmeier

lib/associations/has-many.js
```diff
@@ -72,7 +72,7 @@ var HasMany = function(source, target, options) {
   if (!this.foreignKey) {
     this.foreignKey = Utils.camelizeIf(
       [
-        Utils.underscoredIf(this.source.options.name.singular, this.source.options.underscored),
+        Utils.underscoredIf(this.options.as || this.source.options.name.singular, this.source.options.underscored),
```
mickhansen: See comment for N:M

lib/associations/belongs-to-many.js
sushantdhiman: Original behavior for `as` require us to use the `as` field entirely. If we change it to `N:M`, we have to change it to `1:N` as well?

mickhansen: Yes it would be for both ?:M.
I'm not certain it's a good idea, but it basically forces users to define foreignKey every time they define as, which is probably not a bad idea but.

sushantdhiman: @janmeier ping

janmeier: Pong 

I don't think we should try being smart, lets just use `as` literally

sushantdhiman: Then we can merge it :)

janmeier:
@sushantdhiman All other dialects bug sqlite is failing

sushantdhiman:
Let me fix that :)

sushantdhiman:
@janmeier will this branch have to wait for #5924

janmeier:
#5924 is merged, so no :)

sushantdhiman:
Oh I was so close, need to rebase again I think :)

sushantdhiman:
@janmeier can you restart https://travis-ci.org/sequelize/sequelize/builds/133558506 build, its stuck

sushantdhiman:
@janmeier you broke my build :laughing: , need to rebase again

janmeier:
Yeah sorry, I forget to lint sometimes as well :(

sushantdhiman:
Finally done :hushed: , Please merge @janmeier

janmeier:
Done, thanks! :))

mickhansen:
@janmeier What ended up being the decision for ?:M?

Issue #3568, opened 2015-04-20
Transaction autocommit option


greghart:
This may be less of a bug and more of a mis-understanding. This regards the transaction `autocommit` option, as detailed here (http://sequelize.readthedocs.org/en/stable/api/sequelize/#transactionoptions-promise). 

First of all, it seems like this is unnecessary (as on MySQL at least a transaction is an implicit autocommit: false. See https://dev.mysql.com/doc/refman/5.0/en/commit.html : "To disable autocommit mode implicitly for a single series of statements, use the START TRANSACTION statement:"). Maybe this is not the case for other dialects, or I am missing the point of this field.

Regardless, code I am working on upgrading from 1.7 to 2.0 was was already using `transaction(autocommit: false)`, and claimed to need it for certain bugs (outside the scope of this issue). While working out the kinks, I noticed that applying this flag keeps the connection it's run on  with autocommit mode disabled, even after the transaction completes. My intuition is that with the flag being supplied in the transaction call, it would be reset back at the end of the transaction. If that's not the case, maybe fix here is as simple as some documentation as to this option.

Here is the code I am using to see this
```coffeescript
Sequelize = require 'sequelize'
Promise = require 'bluebird'

config = 
  host: '127.0.0.1'
  dialect: 'mysql'
  sync: {force: false}
  syncOnAssociation: false
  define:
    timestamps: false
    underscored: true
    freezeTableName: true
    syncOnAssociation: false
  maxConcurrentQueries: 250
  pool: { maxConnections: 1 } 

sequelize = new Sequelize 'test', 'user', 'pass', config

Test = sequelize.define 'test', 
  thing: Sequelize.STRING

showAutocommit = ->
  sequelize.query("show variables like 'autocommit'")
  .spread (results) ->
    console.log results[0].Value

Test.sync(force: true)
.then showAutocommit
.then ->
  sequelize.transaction(autocommit: false)
.then (t) ->
  options = transaction: t
  Test.create(
    {
      thing: "transaction"
    }, 
    options
  )
  .then ->
    t.commit()
  .catch ->
    t.rollback()
# autocommit is now OFF for this session. Should this be the case?
.then showAutocommit

```

Also I know coffeescript isn't this repo's thing, LMK if you'd like the test code in Javascript, or any other info. Thanks!

mickhansen:
1.7 had quite a few transaction bugs so probably unrelated.
[...]

PR #5960, opened 2016-05-25
Auto commit disabled by default

### LINKED ISSUE #5094: Do not set any transaction values unless explicitely defined by user or required by dialect.

sushantdhiman:
### Pull Request check-list

- [x] Have you added new tests to prevent regression ?

### Description of change

`AUTO COMMIT` query is not sent by default except for `MySQL`. Closes #5094


mickhansen:
I'd love a unit test or two for this. Simply starting a transaction and verifying it runs the expected SQL queries.

It could be the same unit tests covering transaction preperation in general, assuming we make other changes.

sushantdhiman:
Added the unit tests. From unit test I found out `sqlite` was wasting one query by sending an effective comment in place of auto commit query. Fixed it as well.

test/unit/transaction.test.js
```diff
@@ -0,0 +1,57 @@
...
+    });
+  } else {
+    it('should not run auto commit query', function() {
+      var self = this;
+      return current.transaction(function(t) {
+        expect(self.stub.calledOnce).to.be.ok;
+        expect(sqlIndex.join(' ').indexOf('SET autocommit = 1;')).to.be.eql(-1);
+        return Promise.resolve({});
+      });
+    });
```
mickhansen: Could we test on the sequence instead, test again an array of SQL calls.

sushantdhiman: I can use the latest `sql` statement from `query` stub. That will have the `auto commit`

test/unit/transaction.test.js
```diff
@@ -0,0 +1,57 @@
...
+  , dialect = Support.getTestDialect()
+  , current = Support.sequelize
+  , Promise = Sequelize.Promise;
+
+describe('Transaction', function() {
+  var sqlIndex = [];
+
+  before(function () {
+    this.stub = sinon.stub(current, 'query', function (sql) {
+      sqlIndex.push(sql);
```
mickhansen: Easier to just look at the args for the stub, no need to maintain this.

test/unit/transaction.test.js
```diff
@@ -0,0 +1,57 @@
...
+  , Sequelize = Support.Sequelize
+  , dialect = Support.getTestDialect()
+  , current = Support.sequelize
+  , Promise = Sequelize.Promise;
+
+describe('Transaction', function() {
+  var sqlIndex = [];
+
+  before(function () {
+    this.stub = sinon.stub(current, 'query', function (sql) {
```
mickhansen: sinon.stub(current, 'query').returns(Promise.resolve({}));

test/unit/transaction.test.js
sushantdhiman: Right

test/unit/transaction.test.js
sushantdhiman: Was experimenting a bit earlier, forgot to change it.

sushantdhiman:
Refactored tests as per your comments @mickhansen

test/unit/transaction.test.js
```diff
@@ -0,0 +1,53 @@
...
+    it('should not run auto commit query', function() {
+      var self = this;
+      return current.transaction(function(t) {
+        expect(self.stub.calledOnce).to.be.ok;
+        expect(self.stub.args[1]).to.be.empty;
+        expect(self.stub.args[0].join(' ').indexOf('autocommit')).to.be.eql(-1);
+        return Promise.resolve({});
+      });
+    });
+  }
```
mickhansen: Sorry for bikeshedding, but perhaps something like:

```
var expectations = {
  all: [
    'START TRANSACTION',
    'COMMIT'
  ],
  mysql: [

   ]
};
return current.transaction(function(t) {
  return Promise.resolve();
}).then(function () {
  expect(self.stub.calls.map(call => call.args[0])).to.deep.equal(expectations[dialect] || expectations.all});
});

**Comment by sushantdhiman** [comment]
I have re-factored test to check for sequence of queries. Its ready for merge

**Comment by janmeier** [comment]
LGTM, merge is awaiting the ES6 rewrite :)

File: test/unit/transaction.test.js [hunk 5]
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  it('should run auto commit query only when needed', function() {
+    var expectations = {
+      all: [
+        'START TRANSACTION;'
+      ],
+      mysql: [
+        'START TRANSACTION;',
+        'SET autocommit = 1;'
```
janmeier: @sushantdhiman Actually, I'm not sure why this is needed? As I read the mysql docs 
> with autocommit mode enabled. This means that as soon as you execute a statement that updates (modifies) a table, MySQL stores the update on disk to make it permanent. The change cannot be rolled back.

This is the normal mode, outside of transactions.


>To disable autocommit mode implicitly for a single series of statements, use the START TRANSACTION statement:

http://dev.mysql.com/doc/refman/5.7/en/commit.html

So disabling autocommit with start transaction, then immediately enabling it again feels wrong.

sushantdhiman: I'll check what happens when we disable auto commit in MySQL, may be related to #3568

sushantdhiman: Hi @janmeier @mickhansen 

I have been investigating the reason why MySQL required `autocommit`. @janmeier is correct, as per `MySQL` docs we don't need to set it for `START TRANSACTION` which implies autocommit is set to false. 

But there is a bug in sequelize as noted in #3568, connection which use the `autocommit off` once will remain `off` until user manually doesn't switch it back `on`. This should happen inside `commit()` and `rollback()` automatically?

So can we merge this PR with MySQL `autocommit` set to 1 by default, as its supposed to? Then I will work on #3568 later on and will send a PR to fix it.

janmeier: Agreed :)

sushantdhiman:
All green and ready for merge @janmeier

Issue #5708, opened 2016-04-06
Disable query field aliasing


mickhansen:
Instead of aliasing fields in the query and trying to figure out the name when coming from a subquery let's just disable aliasing completely and map it after the query has executed.

janmeier:
+1 - aliassing seemed like a good idea at the time it was implemented because we didn't have to do mapping back to field names, but it does a lot of problems for joins and subqueries.

Perhaps `instance.set` should be able to handle the aliassed keys, so query.js or similar doesn't have to worry about the aliasses?

mickhansen:
I actually thought that query.js should handle it per: https://github.com/sequelize/sequelize/issues/5685
Since it also helps people using raw queries with complex includes, and we could provide a helper to generate mappings.

janmeier:
Raw query support is a good argument for having it in query! :)

heisian:
Since this seems to be a pretty long-standing issue, I'd like to help this process along.. any suggestions from owners/contributors on where to start (besides just digging through the source of course)?

directed here from #9246

Thanks.

mickhansen:
> Why wouldn't we just store the mapping of field names and do a string conversion of the keys at the application layer?

That's the proposal, basically :)

sushantdhiman:
@heisian Issue is not closed

mickhansen:
@heisian fields are renamed so they are prefixed to avoid conflicts with other tables and to ease the building of nested structures.

heisian:
I guess I'm a bit dense, I don't see how the above helps either situation. I do see tables pluralized to their relationship, such as `"Tables"` getting `AS`'ed to `"Table"` for the ORM to populate a related model with the `Table` attribute in a `hasOne` relationship, but I don't see why that can't be handled completely in the ORM as well, without adding unnecessary SQL to the actual query.

mickhansen:
@heisian If you have models A and B with column name, and you join them, one of the fields must be renamed to allow easy identification when processing the results. Technically you could look at the order of the columns but that would probably make the code harder to read (and only just had the idea).

The plural to singular or reverse is simply from the logic being association name + field name.
The reason we use association names is again, because you could have multiple associations to the samle table and the names would clash.

>  Objectively, this ORM is 6/10th the quality of Laravel's Eloquent.

Not relevant to the discussion is it? But i would invite you to look in the differences of manpower.
What kind of query does Eloquent produce when eager loading? I'd be interested to see and you seem in a position to provide the answer.

[...]

PR #5986, opened 2016-05-28
Closes #5985 -- Fixed an issue where custom-named model fields break when offsetting, ordering, and including hasMany simultaneously.

### LINKED ISSUE #5985: Custom-named model fields break when offsetting, ordering, and including hasMany simultaneously.
## What you are doing?
Sequelize doesn't seem to understand the naming of fields in the database outside of a subquery clause when ordering, offsetting and including members from a hasMany relationship. Here is an example:

```
        var Foo = current.define('foo', {
            fooId: {
              field: 'foo_id',
              type: DataTypes.UUID,
              primaryKey: true
            },
            baseNumber: {
              field: 'base_number',
              type: DataTypes.STRING(25)
            }
          }, {
          timestamps: false
        }),
          Bar = current.define('bar', {
            barId: {
              field: 'bar_id',
              type: DataTypes.UUID,
              primaryKey: true
            },
            fooId: {
              field: 'foo_id',
              type: DataTypes.UUID,
              allowNull: false
            }
          }, {
            timestamps: false
          });

        Foo.hasMany(Bar, {foreignKey: 'foo_id', as: 'bars'})

        return current.sync({force: true})
          .then(function () {
            var options = {
              include: [
                {
                  model: Bar, as: 'bars'
                }
              ],
              offset: 0,
              limit: 50,
              order: [['base_number', 'ASC']]
            };

           // This statement throws an exception.
            return Foo.findAll(options).then(function () {
              console.log('success!')
            })
          });
```

I expect a proper data set in return.

The output throws. This appears to be because the SQL that happens is the following:

> SELECT "foo".*, "bars"."bar_id" AS "bars.barId", "bars"."foo_id" AS "bars.fooId", "bars"."foo_id" AS "bars.foo_id" FROM (SELECT "foo"."foo_id" AS "fooId", "foo"."base_number" AS "baseNumber" FROM "foos" AS "foo" ORDER BY "foo"."base_number" ASC LIMIT 50 OFFSET 0) AS "foo" LEFT OUTER JOIN "bars" AS "bars" ON "foo"."fooId" = "bars"."foo_id" ORDER BY "foo"."base_number" ASC;

This causes postgres to return the following error:

>  column foo.base_number does not exist

The proper query should be:

> SELECT "foo".*, "bars"."bar_id" AS "bars.barId", "bars"."foo_id" AS "bars.fooId", "bars"."foo_id" AS "bars.foo_id" FROM (SELECT "foo"."foo_id" AS "fooId", "foo"."base_number" AS "baseNumber" FROM "foos" AS "foo" ORDER BY "foo"."base_number" ASC LIMIT 50 OFFSET 0) AS "foo" LEFT OUTER JOIN "bars" AS "bars" ON "foo"."fooId" = "bars"."foo_id" ORDER BY "foo"."baseNumber" ASC;

__Dialect:__ any
__Database version:__ 9.5.2.0
__Sequelize version:__ 4.0.0

DavidMorton:
### Pull Request check-list

### Description of change

- [FIXED] Fixed an issue where custom-named model fields break when offsetting, ordering, and including hasMany simultaneously. [#5985] (https://github.com/sequelize/sequelize/issues/5985)


DavidMorton:
This was breaking ordering on a grid of ours, but we don't want to change our model because it'll break the API, so hence, I humbly submit this fix.

mickhansen:
As a note if you want this released soon you should target the v3 branch.

janmeier:
We might not need this fix in 4.0 at all, once https://github.com/sequelize/sequelize/issues/5708 lands

DavidMorton:
Alright. I've added #5996 to add this support in v3.

PR #5990, opened 2016-05-29
Tidy up range type

### LINKED ISSUE #5983: Range DataType's default bound inclusion inconsistent with postgres
The default bounds for range types are [exclusive, exclusive] while postgres's defaults are [inclusive, exclusive], see [here](https://www.postgresql.org/docs/9.5/static/rangetypes.html#RANGETYPES-CONSTRUCT). 

## What are you doing?

```js
Timeline.create({range: [new Date(2016, 5, 1), new Date(2016, 6, 1)]});
```

```
'["2016-06-01 00:00:00.000 +00:00","2016-07-01 00:00:00.000 +00:00")'
```

```
'("2016-06-01 00:00:00.000 +00:00","2016-07-01 00:00:00.000 +00:00")'
```

## Solution

Fixing it is just a [single line of code](https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/range.js#L38). But it's a silent and somehow dangerous breaking change. It's ok if it's added to 4.0 but before submitting a PR, fixing the code, updating docs and tests (also there really aren't any docs on the usage, I will write them with the PR) i wanted to be sure everyone agrees with the change.

__Dialect:__ postgres
__Database version:__ any
__Sequelize version:__ any

alitaheri:
This PR tends to clean up some issues with the range types

- [x] Range type default bounds should adhere to [postgres's defaults](https://www.postgresql.org/docs/9.5/static/rangetypes.html#RANGETYPES-CONSTRUCT), closes #5983 
- [x] Range types should support [element based](https://www.postgresql.org/docs/9.5/static/functions-range.html) operator (`contains element`) 
- [x] Range types should support [range operators](https://www.postgresql.org/docs/9.5/static/functions-range.html)
- [x] There should be complete documentation for range types and how they should be used.

### Pull Request check-list

In order to support `contains element` ( `@>` ) Range should be able to stringify it's subtype too.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,10 @@
+# Future
+- [CHANGED] Range type bounds now default to [postgres default](https://www.postgresql.org/docs/9.5/static/rangetypes.html#RANGETYPES-CONSTRUCT) `[)` (inclusive, exclusive) [#5990](https://github.com/sequelize/sequelize/issues/5990)
+- [ADDED] Support for range operators [#5990](https://github.com/sequelize/sequelize/issues/5990)
+
+## BC breaks:
+- Range type bounds now default to [postgres default](https://www.postgresql.org/docs/9.5/static/rangetypes.html#RANGETYPES-CONSTRUCT) `[)` (inclusive, exclusive), previously was `()` (exclusive, exclusive)
+
 # 4.0.0-0
 - [FIXED] Pass ResourceLock instead of raw connection in MSSQL disconnect handling
 - [CHANGED] Remove `hookValidate` in favor of `validate` with `hooks: true | false`.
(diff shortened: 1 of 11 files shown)


alitaheri:
@janmeier I think this is ready to be merged, No idea why mssql failed, race condition maybe? it didn't fail for my fork. O.o

janmeier:
Good work @alitaheri ! 

Do note the master = 4.0, if you need this to be release in v3 as well please provide a PR against the v3 branch

alitaheri:
Thanks :grin: This contains entangled breaking changes. There is no way to cleanly apply this on v3 without introducing severe breaking changes, I'm using the HEAD in my application anyway. :smile:

PR #6001, opened 2016-06-01
Fix 3568, MySQL transaction are not set to AutoCommit now

### LINKED ISSUE #3019: Model.count returning NaN because Mysql returns an empty string
### Problem

`Model.count` is returning `NaN` because Mysql returns an empty string when grouping.

```MySQL
SELECT count(*) AS `count` FROM `Model` AS `Model` WHERE (`Model`.`deleted_at` IS NULL AND `Model`.`a_model`='Post' AND `Model`.`a_foreign_key`=186);
# count: 0
```

```MySQL
SELECT count(*) AS `count` FROM `Model` AS `Model` WHERE (`Model`.`deleted_at` IS NULL AND `Model`.`a_model`='Post' AND `Model`.`a_foreign_key`=186) GROUP BY `Model`.`user_id`;
# count: ''
```

If there are values to group up, everything works as expected, but if not, MySQL returns an empty string instead of a `0` (zero).

### Solution

Return parsed Int, or `0` at [queryInterface](https://github.com/sequelize/sequelize/blob/master/lib/query-interface.js#L734);

### LINKED ISSUE #3568: Transaction autocommit option

sushantdhiman:
### Pull Request check-list

### Description of change

Closes #3568

`MySQL` transactions were broken because of `AUTO COMMIT`. Setting it to `0` will not work because it will remain same for whole connection. It can be solved if we don't send any `AUTO COMMIT` option at all. `MySQL` automatically keep a transaction under `0` and reset setting to `1` beyond transaction.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -20,6 +20,7 @@
 - [FIXED] All associations now prefer aliases to construct foreign key [#5267](https://github.com/sequelize/sequelize/issues/5267)
 - [REMOVED] Default transaction auto commit [#5094](https://github.com/sequelize/sequelize/issues/5094)
 - [REMOVED] Callback support for hooks [#5228](https://github.com/sequelize/sequelize/issues/5228)
+- [FIXED] Broken transactions in `MySQL` [#3568](https://github.com/sequelize/sequelize/issues/3568)
 
 ## BC breaks:
 - `hookValidate` removed in favor of `validate` with `hooks: true | false`. `validate` returns a promise which is rejected if validation fails
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1692,6 +1692,11 @@ var QueryGenerator = {
       return;
     }
 
+    // no query when value is not explicitly set
+    if (typeof value === 'undefined' || value === null) {
+      return;
+    }
+
     return 'SET autocommit = ' + (!!value ? 1 : 0) + ';';
   },
 
--- a/lib/dialects/mysql/index.js
+++ b/lib/dialects/mysql/index.js
@@ -35,10 +35,7 @@ MysqlDialect.prototype.supports = _.merge(_.cloneDeep(Abstract.prototype.support
   updateOnDuplicate: true,
   indexViaAlter: true,
   NUMERIC: true,
-  GEOMETRY: true,
-  transactionOptions: {
-    autocommit: true
-  }
+  GEOMETRY: true
 });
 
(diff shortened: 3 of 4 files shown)


sushantdhiman:
Hi @mickhansen @janmeier , Please review.

mickhansen:
LGTM, @janmeier

PR #6002, opened 2016-06-01
Doc Fixes

### LINKED ISSUE #5220: Missing documentation on options parameter across multiple methods
As seen missing [here](https://github.com/sequelize/sequelize/blob/0de404640d4c71e2d1f1259356650dfb586a248b/lib/associations/has-one.js#L71) , [here](https://github.com/sequelize/sequelize/blob/0de404640d4c71e2d1f1259356650dfb586a248b/lib/associations/has-many.js#L97) and [here](https://github.com/sequelize/sequelize/blob/0de404640d4c71e2d1f1259356650dfb586a248b/lib/associations/belongs-to-many.js#L202) there is missing documentation on the options object parameter in methods that support the options that sequelize.query() do.

Where documentation exists, as seen  [here](https://github.com/sequelize/sequelize/blob/0de404640d4c71e2d1f1259356650dfb586a248b/lib/associations/belongs-to.js#L98) it just mentions that will be passed to another method. 

As @mickhansen mentioned on the Slack channel, we should probably refer to query options from all of those methods.

### LINKED ISSUE #5633: Association getters allows "scope" and "where" only but not complete query objects
When lazy-loading an association through the corresponding getter I can only filter via where conditions, or by naming a scope, but I also want to be able to add nested inclusions or include/exclude some attributes, in that specific call; I don't want to add a scope in the target just for that particular case.
```javascript
//This is how I must proceed currently, I'm forced to create an scope on Property "includeCategory"
//which have the include.
User.getProperties({ scope: 'includeCategory', where: { configurableByUser: true }});
//What I want is to be able to specify the query directly on the call, not only a where attribute
User.getProperties({ include: [{model: Category, as: 'category'}], where: { configurableByUser: true }});
```

### LINKED ISSUE #5778: Creating elements of a "HasMany" association does not work
## What you are doing?
Hi all,

I am just using the example from the doc web site and it does not seems to work.
http://docs.sequelizejs.com/en/latest/docs/associations/#creating-elements-of-a-hasmany-or-belongstomany-association

```js
var Product = sequelize.define('product', {
    title: Sequelize.STRING
});

var Tag = sequelize.define('tag', {
    name: Sequelize.STRING
});

Product.hasMany(Tag);

    .then(function () {
        Product.create({
            id: 1,
            title: 'Chair',
            Tags: [
                { name: 'Alpha' },
                { name: 'Beta' }
            ]
        }, {
                include: [Tag]
            }).then(function(value) {

            })
    });
```

I expect tags to be created at the same time that the product.

The only output is the following:

Executing (default): INSERT INTO `products` (`id`,`title`,`createdAt`,`updatedAt`) VALUES (1,'Chair','2016-04-21 22:30:27.710 +00:00','2016-04-21 22:30:27.710 +00:00');

I am thus missing the tags creation.

__Dialect:__ sqlite
__Database version: 3.1.2
__Sequelize version: 3.21.0

sushantdhiman:
Closes #5778 , Closes #5633 , Closes #5220

### DIFF
--- a/docs/docs/associations.md
+++ b/docs/docs/associations.md
@@ -791,7 +791,7 @@ Now we can create a project with multiple tags in the following way:
 Product.create({
   id: 1,
   title: 'Chair',
-  Tags: [
+  tags: [
     { name: 'Alpha'},
     { name: 'Beta'}
   ]
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -203,6 +203,7 @@ var BelongsToMany = function(source, target, options) {
      * @param {Object} [options.where] An optional where clause to limit the associated models
      * @param {String|Boolean} [options.scope] Apply a scope on the related model, or remove its default scope by passing false
      * @param {String} [options.schema] Apply a schema on the related model
+     * @see {Model#findAll}  for a full explanation of options
      * @return {Promise<Array<Instance>>}
      * @method getAssociations
      */
--- a/lib/associations/belongs-to.js
+++ b/lib/associations/belongs-to.js
@@ -78,6 +78,7 @@ var BelongsTo = function(source, target, options) {
      * @param {Object} [options]
      * @param {String|Boolean} [options.scope] Apply a scope on the related model, or remove its default scope by passing false.
      * @param {String} [options.schema] Apply a schema on the related model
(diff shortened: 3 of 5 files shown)


sushantdhiman:
**cc** @mickhansen @janmeier

Issue #5826, opened 2016-04-30
Non-unique constraint names in MySQL


erwinhaas:
I think there is bug in the mysql dialect `addColumnQuery` query generator. 

### What I'm doing

I'm adding a column to an existing table with a migration:

```javascript
return queryInterface.addColumn('carousel', 'category_id', {
      type: Sequelize.INTEGER(11).UNSIGNED,
      defaultValue: null,
      allowNull: true,
      references: {
        model: 'categories',
        key: 'id'
      },
      onUpdate: 'CASCADE',
      onDelete: 'SET NULL'
    });
```

There is already another table that references the same `category` table which uses the field named `category_id`.

### What I expect to happen

I would expect that the column is added with an foreign key constraint in the database.

### What is actually happening?
What happens is that adding a column fails with the following error:

```
Executing (default): ALTER TABLE `carousel` ADD `category_id` INTEGER(11) UNSIGNED DEFAULT NULL, ADD CONSTRAINT `category_id_foreign_idx` FOREIGN KEY (`category_id`) REFERENCES `categories` (`id`) ON DELETE SET NULL ON UPDATE CASCADE;

Unhandled rejection SequelizeDatabaseError: ER_DUP_KEY: Can't write; duplicate key in table '#sql-d45_999'
```

### Why this is happening

I already did some digging in the code to see where the problem originates.

The naming convention that is used creates a foreign key is `foreign key + '_foreign_idx'` (see https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/query-generator.js#L270-L277)

```javascript
      if (options && options.context === 'addColumn' && options.foreignKey) {
        var attrName = options.foreignKey;

        template += Utils._.template(', ADD CONSTRAINT <%= fkName %> FOREIGN KEY (<%= attrName %>)')({
          fkName: this.quoteIdentifier(attrName + '_foreign_idx'),
          attrName: this.quoteIdentifier(attrName)
        });
      }
```

Adding multiple constraints with the same `key` name doesn't work on MySQL, as the constraint symbol value should be unique within the database.

> If the CONSTRAINT symbol clause is given, the symbol value, if used, must be unique in the database. A duplicate symbol will result in an error similar to: ERROR 1022 (2300): Can't write; duplicate key in table '#sql- 464_1'. If the clause is not given, or a symbol is not included following the CONSTRAINT keyword, a name for the constraint is created automatically.

[...]

PR #6008, opened 2016-06-01
Include tableName in MySQL fk contraint names

hagabaka:
### Pull Request check-list

### Description of change

When creating a foreign key constraint in MySQL, use the template
`{referring_table}_{attribute}_foreign_idx`, rather than just
`{attribute}_foreign_idx`, to avoid duplicate constraint names.

See GitHub #5826

### DIFF
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -103,6 +103,7 @@ var QueryGenerator = {
         key: this.quoteIdentifier(key),
         definition: this.attributeToSQL(dataType, {
           context: 'addColumn',
+          tableName: table,
           foreignKey: key
         })
       });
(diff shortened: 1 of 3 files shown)


mickhansen:
LGTM, /cc @janmeier

danielepolencic:
any chance this could be released?

mickhansen:
@danielepolencic This PR targeted master so it's slated for the next 4.0 pre release.

eliias:
@mickhansen any chance of backporting this to 3.x? As it is impossible to set the name of the constraint, only raw SQL is left for the rescue.

felixfbecker:
Isnt this a beaking change?

mickhansen:
@felixfbecker I don't believe so since we don't really depend on constraint names anywhere.
@eliias It's very unlikely that the core team will get around to that, but a PR is certainly welcome.

felixfbecker:
@mickhansen we don't, but a user might (think raw queries in migration files, etc). It's like changing the default casing of table names.

mickhansen:
@felixfbecker That is true

felixfbecker:
@eliias it will land in v4

diosney:
Sadly we have to wait until v4 for this.

In the meantime, as a workaround, I'm setting the column to be something like `ModeNameColumnName`.

If you find a nicer workaround please share it!

blackrabbit99:
Thank you for breaking changes!

jramos-br:
This problem also affects mssql dialect. Any plans to change it there?

PR #6009, opened 2016-06-01
ES6 refactor

felixfbecker:
This PR refactors every single file in the Sequelize library to use ES6 features supported in Node 4:

 - `let` / `const`
 - arrow functions
 - classes
 - property / method shorthands
 - template strings
 - `for` .. `of`

To make the commits easier to review, I did the refactor to a class in one commit and then the rest of the changes in another.

#### Variables
The current style declares variables on the top of functions like this:
```js
var one
  , two
  , three;
```
This is because `var` is hoisted and keeping the declarations on the top of the function make it easy to spot which vars are in the scope.
With `let` / `const` the same keyword cannot be applied to all declarations anymore, and the keywords have a different length. Additionally, they are not hoisted anymore, so one _can_ and _should_ declare the variable as late as possible, where they are used. To keep consistent, I changed the style to always specify the keyword:
```js
const one = 1;
const two = 2;
let three;
```
#### Export
I changed all exports to prefer `exports.something = something` over `module.exports = {something: something}` because it is closer to ES6 `export` syntax (which might already land in Node 7). I also exported the major classes both with `export default` in addition to `module.exports`, which will improve the usage of Sequelize in ES6 environments and greatly simplify the TypeScript definitions for 4.0 I am working on.

### Verification
I refactored this with the help of an `eslintrc` with rules that enforce usage of the above features. I will add that in the seperate PR.

### Whats not included
I did not touch the tests at all so I do not risk introducing bugs there. I also did not fully refactor the query generators to template strings because that would introduce major restructuring in some methods.

### Bugs
There are still (after countless fixes) bugs left and I am a bit lost from where they come from. Problems come from
 - `forEach` with a `return` that has not been changed to `continue` in `for`..`of`
 - changing anonymous functions to arrow functions where it should have been a method with dynamic `this`
 - Some lodash templates that I tried to change to template strings
 - ???

### ToDo
 - [ ] Make tests pass
 - [ ] Change arrow function style to only omit `return` if it's a one-liner

I would love to get some tips on the test failures and code review :)
I'm really thrilled to see benchmarks once the tests pass!


sushantdhiman:
```js
const one
    , two
    , three;
```

I find this style aesthetic, personally. Can we keep this as much possible ? Like all `require` on starting. Some variables once defined as `let` may need to go down on scope but we should keep the global definition style same.

alitaheri:
I personally prefer one statement per deceleration. airbnb makes a good point:

https://github.com/airbnb/javascript#variables : 

> Why? It's easier to add new variable declarations this way, and you never have to worry about swapping out a `;` for a `,` or introducing punctuation-only diffs. You can also step through each declaration with the debugger, instead of jumping through all of them at once.

felixfbecker:
@alitaheri yes, good point.
@sushantdhiman I initially tried to keep the style but it resulted in a total mess because of the _two_ declaration keywords that have different length:
```js
const one = 1
    , two = 2;
let three = one + two
  , four;
const five = 5
```
Pro/Cons:

**Comma-first**
 - aesthetic for some

**One declaration per line**
 - `let`/`const` have different length
 - add a variable at the bottom without a two-line diff
 - in line with JS Standard and Airbnb (and I think Yahoo) coding style guide
 - commas are placed at the end in enumerations in natural language
 - step through declarations with a debugger
 - it's impossible to lint comma-first because Sequelize only uses it for variables, but places commas at the end for properties. It is also not possible to lint that the indentation is aligned with the keyword. There is a rule for one-declaration-per-line though.
 - Multiple hours of work to refactor again 😃 

@mickhansen @janmeier ?

felixfbecker:
Did anyone already clone this and can help me with the tests? I have the unit tests passing and am focusing on MySQL integration tests now. I have a test failing with a custom DATE parser and don't know why.

mickhansen:
@felixfbecker Thanks for the hard work, will try to review as soon as possible (there's a ton).

docs/docs/transactions.md
```diff
@@ -97,17 +88,17 @@ You can have concurrent transactions within a sequence of queries or have some o
...
     // With CLS enable, queries here will by default use t2
     // Pass in the `transaction` option to define/alter the transaction they belong to.
-    return Promise.all([
-        User.create({ name: 'Bob' }, { transaction: null }),
-        User.create({ name: 'Mallory' }, { transaction: t1 }),
-        User.create({ name: 'John' }) // this would default to t2
-    ]);
-  });
-});
+    Promise.all([
```
mickhansen: While i understand that => implicitely returns i think it's better to have explicit return statements since you'll mess everything up if you change it to two lines later on (this goes for alle changes of this nature).

mickhansen:
Linting currently fails, be great if that could be fixed so we can analyze tests on travis.

felixfbecker:
@mickhansen Yes, I'm aware this is one huge PR to review :D
The linting currently fails because in https://github.com/felixfbecker/sequelize/blob/es6-refactor/lib/dialects/sqlite/query-generator.js#L62 there `encoding` is defined but never used. This was not caught by JSHint before the refactor because it was only set on the `replacements` object. Should I remove it?

docs/docs/transactions.md
felixfbecker: So what should be the rule? If it doesn't fit in one line, use a return? Should I only do this in the docs or also in the codebase? There are 200+ places where this would have to be changed again

mickhansen:
@felixfbecker If it's not used go ahead and remove it.

docs/docs/transactions.md
mickhansen: Both docs and codebase, single line is ok but i think multi line is too prone to mistakes in the future (one of the reasons that bluebird introduced the no promise returned errors).

felixfbecker:
@mickhansen I fixed the JSHint issues. I will adapt the arrow function style when I find the time, but getting the tests to pass is more important in the meantime.

sushantdhiman:
@mickhansen @janmeier While this PR is being reviewed can we submit more PRs or will have to wait for this one?

mickhansen:
@sushantdhiman So much work in this PR we'll likely want to merge it first so rebasing doesn't become a nightmare.

mickhansen:
@felixfbecker Hmm, odd test failures, looks like some context is lost somewhere, but can't immediately see what's wrong.

felixfbecker:
rebased

mickhansen:
All `an ES6 class` simply convert to an es6 class with no functional changes, correct?

lib/model.js
```diff
@@ -58,20 +58,20 @@ class Model {
...
       return options;
     }
 
-    var deletedAtCol = model._timestampAttributes.deletedAt
-      , deletedAtAttribute = model.rawAttributes[deletedAtCol]
-      , deletedAtObject = {}
-      , deletedAtDefaultValue = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
+    const deletedAtCol = model._timestampAttributes.deletedAt;
+    const deletedAtAttribute = model.rawAttributes[deletedAtCol];
+    const deletedAtObject = {};
```
mickhansen: This object is modified right after, shouldn't that fail with const?

mickhansen: Or does const only protect against reassignment?

mickhansen:
In the future we probably want to break this sort of PR into multiples :) Very hard to review, and rebasing becomes a constant issue.

felixfbecker:
@mickhansen yes, I did that because the class diff itself is always the whole file because it changes the indentation, and in the second commit you can then see in the diff what _actually_ changed with the rest of ES6 features.

I actually thought about splitting this with cherry-picking because I am so lost with the tests - that way we could isolate the failures.

alitaheri:
@felixfbecker I think it's better if this is broken down to smaller PRs. then we can review easier and help with the issues you're having with tests.

mickhansen:
@felixfbecker Yes, that's great - Then i can mostly ignore the ES6 class changes.

I would welcome cherry picking, one or a few files at a time (ok to take entire mysql dialect at once for instance).

lib/model.js
felixfbecker: Correct.

felixfbecker:
@mickhansen What is expected here: https://travis-ci.org/felixfbecker/sequelize/jobs/135577548#L4035 / https://github.com/sequelize/sequelize/blob/18da2d1910b6f447578f5c351b64bcc449b706f3/test/integration/instance.validations.test.js#L345 ?

felixfbecker:
Another question about queries and template strings: is it ok to refactor this
```js
'SELECT CONSTRAINT_NAME as constraint_name'
      + ' FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE'
      + ' WHERE (REFERENCED_TABLE_NAME = ' + wrapSingleQuote(tableName)
      + ' AND REFERENCED_COLUMN_NAME = ' + wrapSingleQuote(columnName)
      + ') OR (TABLE_NAME = ' + wrapSingleQuote(tableName)
      + ' AND COLUMN_NAME = ' + wrapSingleQuote(columnName)
      + ')';
```
into a multi-line template string or do you want to keep it without any line breaks because of logging?

mickhansen:
@felixfbecker I have no clue, i have no recollections of that code nor the test.

@felixfbecker Hmm, i'd love template strings but logging is an issue yes. Multi-lines might cause issues with some log parsers. The biggest issue is that the second to whatever line will have a huge amounts of spaces in front.

mickhansen:
Maybe `expect(Object.keys(err)).to.have.length(3);` should be `expect(Object.keys(err.errors)).to.have.length(3);` but i don't know for certain

janmeier:
Re templating strings, we could add a custom handler
```js
sql` SELECT Foo
and bar
    WHERE BAZ
`
```
Which would simply remove whitespace https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals

felixfbecker:
Git blame says @BridgeAR

felixfbecker:
@janmeier which would do `sql.replace(/\w+/, ' ')`?

janmeier:
Something like that, yeah - It's not fast but should work

felixfbecker:
Hm. In that case I'm more for concatting the template strings. We don't have that many multi-line queries and performance is valid concern here.

felixfbecker:
On the failing test: I tried to find the test in the logs of the latest build on master, but I cannot find it?

mickhansen:
@janmeier I actually tried something like that in my own project, for some reason `/^\w+?/` did not work as well as expected.

@felixfbecker latest is https://travis-ci.org/sequelize/sequelize/builds/135581408

felixfbecker:
@mickhansen I meant I could not find the test 😄  Did a simple search

felixfbecker:
what should I do with it?

mickhansen:
@felixfbecker Good question, no idea why it fails, i guess because it's a class now so Object.keys work differently, but you're welcome to ammend the test.

felixfbecker:
But the error does not have any properties besides `name` and `errors` as `message` is not enumerable. But the description is _produce 3 errors_, so `Object.keys()` doesn't make sense to call on the error object - I think it should be `err.errors` like you said. Rebased BTW

felixfbecker:
MySQL tests pass 🎉

mickhansen:
@felixfbecker You're right, it makes absolutely no sense :)
Did you decide to go ahead with cherry picking or are we going to keep with the monster PR?

felixfbecker:
If I get the tests to pass quickly I would keep it as is (maybe squash into chunks to your preference) or does anything speak against it? If I get stuck I will split so you can merge whats already working

lib/model.js
```diff
@@ -189,15 +187,15 @@ class Model {
     }
 
     // convert all included elements to { model: Model } form
-    options.include = options.include.map(function(include) {
+    options.include = options.include.map(include => {
```
Verdier: Is'nt better write like this : 
```js
options.include = options.include.map(include => this._conformInclude(include, self));
```
which is more readable en more concise?

mickhansen: Would be nicer, yes :)

felixfbecker:
All tests passing except SQLite. Help appreciated

mickhansen:
@felixfbecker Not exactly sure. I think it must be some checks failing because of bad references, but will take a while to check.

Perhaps we should cherry pick a few commits/files at a time and get them committed? It will help us merge chunks at a time and progress while we keep moving ahead :)

felixfbecker:
@mickhansen alright

mickhansen:
@felixfbecker Many thanks if you can make it happen, will help us get the work into the codebase faster. I realize som fixes might depend on eachother, but perhaps it would be possible to start with a few isolated files at a time and get those merged (like ES6-ification of Model.js, specfic dialects)

felixfbecker:
@mickhansen Any ideas how to split it up? I thought about data types, one for each dialect, and the rest. Any better idea?

janmeier:
You mihht want to start with docs as well, thats the only thing I can see in the diff currently :D

mickhansen:
@felixfbecker The smaller you can PR, the better, do it a merge at a time :)

felixfbecker:
@janmeier The splitup will not make the per-file diffs any smaller 😄  You have to clone my fork

mickhansen:
No we will have to live without per-file diffs on github, but smaller merges will still help.

PR #6020, opened 2016-06-03
Added support for using password for sqlite encryption plugin like sq…

asinbow:
### Pull Request check-list

- [Y] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [N] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [N] Have you added new tests to prevent regressions?
- [N] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [N] Have you added an entry under `Future` in the changelog?

### Description of change

There are some sqlite encryption plugins like [sqlcipher](https://github.com/sqlcipher/sqlcipher), with which you can encrypt your sqlite database file. The SQL syntax:
```
PRAGMA key = 'passphrase';
```
According to the [project description](https://github.com/sqlcipher/sqlcipher#encrypting-a-database), PRAGMA key or sqlite3_key should be called as the **first** operation when a database is open.

And sqlcipher is well supported by [node-sqlite3](https://github.com/mapbox/node-sqlite3).

So, what I want to do is to support it in sequelize. Not complicate,  and I did run `npm run test-sqlite`.

Thanks!

### DIFF
--- a/lib/dialects/sqlite/connection-manager.js
+++ b/lib/dialects/sqlite/connection-manager.js
@@ -64,6 +64,10 @@ ConnectionManager.prototype.getConnection = function(options) {
       }
     );
   }).tap(function (connection) {
+    if (self.sequelize.config.password) {
+      // Make it possible to define and use password for sqlite encryption plugin like sqlcipher
+      connection.run('PRAGMA key=' + self.sequelize.escape(self.sequelize.config.password));
+    }
     if (self.sequelize.options.foreignKeys !== false) {
       // Make it possible to define and use foreign key constraints unless
       // explicitly disallowed. It's still opt-in per relation


janmeier:
LGTM, do you mind rebasing, and adding a note to the docs that the password is used for sqlchiper in sqlite

McPo:
+1
Would find this useful.
Take its theres no hack to add support for this in the current sequelize version.

janmeier:
Nope, but I promise to be quicker in merging this PR as soon as it is rebased ;)

It would probably also be nice to have an afterConnect hook, which more generally allows hooking into connections

McPo:
So the quick hack is to set the npm dependency to 

"dependencies": {
    "sequelize": "git://github.com/asinbow/sequelize.git#sqlite-password-patch"
  }

However there appears to be an issue with the branch in that the following error is print to the console.

>  Parameter "url" must be a string, not object

McPo:
If I was to create a new PR with the required changes and credit asinbow, would you be willing to accept it? Would be nice to get this in.

Thanks,
Emmet

McPo:
Ive created a PR with the required changes #6366.

PR #6024, opened 2016-06-05
Docs: Update getting started with how to test connection

kevinsimper:
I thought that you should show that Sequelize does not automaticly test the connection and I simply took the example from the old docs :)


mickhansen:
Thanks @kevinsimper, generally the first query will blowup though!

kevinsimper:
Correct, however my `server.js does not make a query at start :)

mickhansen:
@kevinsimper No i guess they wouldn't normally unless you check the DB on health checks (which is generally a good idea i guess but not something i get around to either currently)

PR #6030, opened 2016-06-06
fix(Model): findOrCreate should fail explicitely, not timeout, when a…

mickhansen:
…ltering hooks before create

### DIFF
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -200,6 +200,7 @@ Query.prototype.handleShowTablesQuery = function(results) {
 
 Query.prototype.formatError = function (err) {
   var match;
+
   match = err.message.match(/Violation of UNIQUE KEY constraint '((.|\s)*)'. Cannot insert duplicate key in object '.*'.(:? The duplicate key value is \((.*)\).)?/);
   match = match || err.message.match(/Cannot insert duplicate key row in object .* with unique index '(.*)'/);
   if (match && match.length > 1) {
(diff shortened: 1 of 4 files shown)


lib/model.js
```diff
@@ -631,12 +631,8 @@ class Model {
       }
     }.bind(this));
 
-    this.attributes = this.rawAttributes = _.mapValues(attributes, function(attribute, name) {
-      if (!Utils._.isPlainObject(attribute)) {
-        attribute = { type: attribute };
-      }
-
-
+    this._fieldToAttributeMap = {};
```
janmeier: remove

test/integration/model/create.test.js
```diff
@@ -394,6 +394,40 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         });
       }
 
+      it.only('should not fail silently with concurrency higher than pool, a unique constraint and a create hook resulting in mismatched values', function() {
```
janmeier: !

test/integration/model/create.test.js
```diff
@@ -394,6 +394,40 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
+          'mick ',
+          'mick ',
+          'mick '
+        ];
+
+        return User.sync({force: true}).then(() => {
+          return Promise.all(
+            names.map(username => {
+              return User.findOrCreate({where: {username}}).catch(err => {
+                expect(err.message).to.equal(`user#findOrCreate: value used for username was not equal for both the find and the create calls, 'mick ' vs 'mick'`);
```
janmeier: Maybe add a spy here and expect it to have been called?

mickhansen:
Might need/want to hold off untill https://github.com/sequelize/sequelize/pull/6009 is merged

lib/dialects/mssql/query.js
```diff
@@ -210,12 +211,12 @@ Query.prototype.formatError = function (err) {
     if (uniqueKey && !!uniqueKey.msg) {
       message = uniqueKey.msg;
     }
-    if (!!match[2]) {
```
janmeier: Maybe the second match is for an old version ?

janmeier:
LGTM

lib/dialects/mssql/query.js
mickhansen: No idea, not sure if it's ever been correct, but it certainly isn't at the moment.

PR #6037, opened 2016-06-07
ES6 refactor: sequelize.js

felixfbecker:
see #6009

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -30,67 +30,67 @@ var url = require('url')
  */
 
  /**
- * Instantiate sequelize with name of database, username and password
- *
- * #### Example usage
- *
- * ```javascript
- * // without password and options
- * var sequelize = new Sequelize('database', 'username')
- *
- * // without options
- * var sequelize = new Sequelize('database', 'username', 'password')
- *
- * // without password / with blank password
- * var sequelize = new Sequelize('database', 'username', null, {})
- *
- * // with password and options
- * var sequelize = new Sequelize('my_database', 'john', 'doe', {})
- *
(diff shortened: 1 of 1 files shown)


janmeier:
LGTM

mickhansen:
Appears to be an issue with where/or/and etc. They only exists as statics now and not on the instance aswell.

```
-Sequelize.where = Sequelize.condition = Sequelize.prototype.condition = Sequelize.prototype.where = function(attr, comparator, logic) {
+  static where(attr, comparator, logic) {
     return new Utils.where(attr, comparator, logic);
-};
+  }
```

mickhansen:
```
- * sequelize.transaction().then(function (t) {
- *   return User.find(..., { transaction: t}).then(function (user) {
- *     return user.updateAttributes(..., { transaction: t});
- *   })
- *   .then(t.commit.bind(t))
- *   .catch(t.rollback.bind(t));
- * })
+   * sequelize.transaction().then(transaction =>
+   *   User.find(..., {transaction})
+   *     .then(user => user.updateAttributes(..., {transaction}))
+   *     .then(() => transaction.commit())
+   *     .catch(() => transaction.rollback());
+   * )
```

Should have a return statement (like we discussed).

felixfbecker:
@mickhansen changed the arrow functions. The aliases are added [here](https://github.com/felixfbecker/sequelize/blob/es6-refactor-sequelize/lib/sequelize.js#L1176-L1185)

mickhansen:
So they are, sorry about that, my diff thought it was done.

mickhansen:
LGTM!

Will merge on green

PR #6040, opened 2016-06-07
ES6 refactor: errors

felixfbecker:
See #6009

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -8,25 +8,21 @@
  * @class Errors
  */
 
-var util = require('util');
-var error = module.exports = {};
-
 /**
  * The Base Error all Sequelize Errors inherit from.
  *
  * @constructor
  * @alias Error
  */
-error.BaseError = function() {
-  var tmp = Error.apply(this, arguments);
-  tmp.name = this.name = 'SequelizeBaseError';
-
-  this.message = tmp.message;
-  if (Error.captureStackTrace)
+class BaseError extends Error {
+  constructor(message) {
+    super(message);
+    this.name = 'SequelizeBaseError';
+    this.message = message;
     Error.captureStackTrace(this, this.constructor);
-};
-util.inherits(error.BaseError, Error);
-
+  }
+}
+exports.BaseError = BaseError;
 
 /**
  * Validation Error. Thrown when the sequelize validation has failed. The error contains an `errors` property,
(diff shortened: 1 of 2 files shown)


mickhansen:
Does this maintain stack trace? I thought something like https://www.npmjs.com/package/es6-error was required

felixfbecker:
I don't know, could you write a test to find out?

mickhansen:
I'm not sure exactly how to test this, i've just been using es6-error in all my projects - I'm pretty sure it's required.

felixfbecker:
@mickhansen I tested it in Node 6
```js
function throwError() {
  throw new errors.BaseError('this is a message');
}
```
results in
```
SequelizeBaseError: this is a message
    at BaseError (C:\Users\felix\git\OpenSource\sequelize\lib\errors.js:19:5)
    at throwError (C:\Users\felix\git\OpenSource\sequelize\test\unit\errors.test.js:8:13)
```
should I add this as a unit test?

mickhansen:
This adds the base error code to the stack, i'm not sure if es6-error keeps the original stack instead?

felixfbecker:
But isnt that the expected behaviour? The stack trace is captured the moment you call `super()`. I don't really like the idea of all the stuff es6-error is doing - with `captureStackTrace()` and `stack: (new Error()).stack`. Subclassing `Error` is an explicit feature of ES6

mickhansen:
@felixfbecker No, the stack trace should never include the error class i would think.

/cc @janmeier

felixfbecker:
@mickhansen I think in Java a stack is always captured at the time the exception is thrown, but in JS and PHP it is always at the time the error is created (eg the `Error` constructor is called, for example with `super()`). You can hack around this by using
```js
    if (Error.captureStackTrace) {
      Error.captureStackTrace(this, this.constructor);
    } else {
      Object.defineProperty(this, 'stack', {
        value: (new Error()).stack
      });
    }
```
but you would have to do this not only in `BaseError`, but in all subclasses like `ValidationError`, `UniqueConstraintError` etc. Imo this is unneeded effort and code just to change Javascript's error stack semantics... It might also have a performance impact.

mickhansen:
@felixfbecker You would only need it in BaseError with es6-error as far as i can tell.
Performance impact should be negligble compared to readable stacks.

http://stackoverflow.com/questions/31089801/extending-error-in-javascript-with-es6-syntax

felixfbecker:
@mickhansen I used `captureStackTrace` since we are targeting Node 4+ and adapted the test to verify it for a subclass.

mickhansen:
@felixfbecker Great!

LGTM, /cc @janmeier

PR #6041, opened 2016-06-07
ES6 refactor: QueryInterface

felixfbecker:
See #6009

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -1,147 +1,166 @@
 'use strict';
 
-var Utils = require('./utils')
-  , _ = require('lodash')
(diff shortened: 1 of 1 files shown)


mickhansen:
LGTM.

Not sure how we'll eventually handle using bluebird.coroutine tho with the current setup (no arrow generators, they have to wrap).

So some of it might end up being refactored to .bind.

janmeier:
We might consider https://github.com/danielstjules/async-class to wrap methods in coroutnies

overlookmotel:
@janmeier https://github.com/danielstjules/async-class looks really clean.

PR #6051, opened 2016-06-07
ES6 refactor: dialects / SQLite

felixfbecker:
See #6009 

Need help on this one, don't know why tests fail. Other PRs depend on this unfortunately.

### DIFF
--- a/lib/dialects/sqlite/query.js
+++ b/lib/dialects/sqlite/query.js
@@ -1,12 +1,11 @@
 'use strict';
 
-var Utils = require('../../utils')
-  , _ = require('lodash')
-  , Promise = require('../../promise')
-  , AbstractQuery = require('../abstract/query')
-  , QueryTypes = require('../../query-types')
-  , sequelizeErrors = require('../../errors.js')
-  , parserStore = require('../parserStore')('sqlite');
+const _ = require('lodash');
+const Promise = require('../../promise');
+const AbstractQuery = require('../abstract/query');
+const QueryTypes = require('../../query-types');
+const sequelizeErrors = require('../../errors.js');
+const parserStore = require('../parserStore')('sqlite');
 
 class Query extends AbstractQuery {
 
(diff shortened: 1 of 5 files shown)


mickhansen:
Taking a look

mickhansen:
Not sure why it's only BelongsToMany failing tho, odd.

mickhansen:
`create` no longer returns a primary key value for sqlite, no idea why this isnt caught anywhere else.

lib/dialects/sqlite/query.js
```diff
@@ -1,385 +1,371 @@
...
+
+                if (benchmark) {
+                  this.sequelize.log('Executed (' + (this.database.uuid || 'default') + '): ' + this.sql, (Date.now() - queryBegin), this.options);
+                }
 
-              if (err) {
-                err.sql = self.sql;
-                reject(self.formatError(err));
-              } else {
-                var metaData = this
```
janmeier: `this` here previously referred to the statement object, which contains the last insert id. So I don't think we can use an array function here (unless you can find another way to get the statement object).

https://github.com/mapbox/node-sqlite3/wiki/API

> callback (optional): If given, it will be called when an error occurs during any step of the statement preparation or execution, and after the query was run. If an error occurred, the first (and only) parameter will be an error object containing the error message. If execution was successful, the first parameter is null. The context of the function (the this object inside the function) is the statement object. Note that it is not possible to run the statement again because it is automatically finalized after running for the first time. Any subsequent attempts to run the statement again will fail.

felixfbecker: Nice catch.

felixfbecker:
Tests pass 🎉

mickhansen:
LGTM so @janmeier can merge when green & if he agrees :>

felixfbecker:
Tests ran in Travis on my branch: https://travis-ci.org/felixfbecker/sequelize/builds/136165083

Issue #6052, opened 2016-06-07
Incorrect Column Name when include, through & where are used together


harshithkashyap:
## What you are doing?

``` js
var User = sequelize.define('User', {
    username: {
        type: Sequelize.STRING,
        primaryKey: true
    }
});

var Entity = sequelize.define('Entity', {
    entity_id: {
        type: Sequelize.INTEGER,
        autoIncrement: true,
        primaryKey: true
    },
    creator: {
        type: Sequelize.STRING,
        allowNull: false
    },
    votes: {
        type: Sequelize.INTEGER,
        allowNull: false,
        defaultValue: 0
    }
});

var Post = sequelize.define('Post', {
    post_id: {
        type: Sequelize.INTEGER,
        allowNull: false,
        primaryKey: true
    }
});

var TaggableSentient = sequelize.define('TaggableSentient', {
    nametag: {
        type: Sequelize.STRING,
        primaryKey: true
    }
});

var EntityTag = sequelize.define('EntityTag', {
    entity_id: {
        type: Sequelize.INTEGER,
        allowNull: false
    },
    tag_name: {
        type: Sequelize.STRING,
        allowNull: false
    }
});

Entity.belongsTo(User, { foreignKey: 'creator', targetKey: 'username' });
Post.belongsTo(Entity, { foreignKey: 'post_id', targetKey: 'entity_id' });

Entity.belongsToMany(TaggableSentient, { 
    as: 'tags', 
    through: { model: 'EntityTag', unique: false }, 
    foreignKey: 'entity_id', 
    otherKey: 'tag_name' 
});

TaggableSentient.belongsToMany(Entity, { 
    as: 'tags', 
    through: { model: 'EntityTag', unique: false }, 
    foreignKey: 'tag_name', 
    otherKey: 'entity_id' 
});

sequelize.sync({ force: true, logging: console.log }).then(function() {
    Post.findAll({
        include: [{
            model: Entity,
            required: true,
            include: [{
                model: User,
                required: true
            },{
                model: TaggableSentient,
                as: 'tags',
                required: true,
                through: {
                    where: {
                        tag_name: ['tag1', 'tag2']
                    }
                }
            }]
        }],
        limit: 5,
        offset: 0
    }).then(function(posts) {
        console.log(posts)
    });
}).catch(function(err) {
    console.error(err);
});
```
## What do you expect to happen?

The above findAll query is supposed to return all the posts where tags have been applied for the post and the tag_name is in specified list.
[...]

PR #6053, opened 2016-06-07
Fix for generateThroughJoin

harshithkashyap:
### Pull Request check-list
- [x] Does your issue contain a link to existing issue (Test Case for #[6052]).
### Description of change

Added Failing test case & patch for #6052

### DIFF
--- a/test/integration/include/findAll.test.js
+++ b/test/integration/include/findAll.test.js
@@ -2043,10 +2043,7 @@ describe(Support.getTestDialectTeaser('Include'), function() {
             }]
           }],
           limit: 5,
-          offset: 0,
-          order: [
-            [ { raw: '\`Entity.votes\`' }, 'DESC' ]
-          ]
+          offset: 0
         });
       }).then(function (posts) {
         expect(posts.length).to.equal(1);
--- a/test/integration/include/findAll.test.js
+++ b/test/integration/include/findAll.test.js
@@ -2015,36 +2015,46 @@ describe(Support.getTestDialectTeaser('Include'), function() {
         foreignKey: 'tag_name', 
         otherKey: 'entity_id' 
       });
-
-      return this.sequelize.sync({ force: true }).then(function () {
+      return this.sequelize.sync({ force: true }).bind(this).then(function () {
+        return User.create({ username: 'bob' });
+      }).then(function () {
+        return TaggableSentient.create({ nametag: 'bob' });
+      }).then(function () {
+        return Entity.create({ creator: 'bob' });
+      }).then(function (entity) {
+        return Promise.all([Post.create({ post_id: entity.entity_id }), entity.addTags('bob')]);
+      }).then(function () {
         return Post.findAll({
+          include: [{
+            model: Entity,
+            required: true,
             include: [{
-              model: Entity,
+              model: User,
+              required: true
+            },{
+              model: TaggableSentient,
+              as: 'tags',
               required: true,
-              include: [{
-                model: User,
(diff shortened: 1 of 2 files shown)


test/integration/include/findAll.test.js
```diff
@@ -1961,5 +1961,92 @@ describe(Support.getTestDialectTeaser('Include'), function() {
...
+                  where: {
+                    tag_name: ['tag1', 'tag2']
+                  }
+                }
+              }]
+            }],
+            limit: 5,
+            offset: 0,
+            order: [
+              [ { raw: '\`Entity.votes\`' }, 'DESC' ]
```
mickhansen: Please use proper order syntax for tests, we don't garantee `raw`

test/integration/include/findAll.test.js
```diff
@@ -1961,5 +1961,92 @@ describe(Support.getTestDialectTeaser('Include'), function() {
...
+            ]
+        }).then(function (posts) {
+          expect(posts.length).to.equal(0);
```
mickhansen: Best to verify that there is some results, and some results filtered.

mickhansen:
Might be simpler to provide this as a sql unit test focusing on where the error happens.

test/integration/include/findAll.test.js
harshithkashyap: @mickhansen I have to use raw / sequelize.literal as there is problem with when using
```js
order: [
    [ this.sequelize.col('Entity.votes'), 'DESC' ]
]
```
I'll raise a separate issue for the same.

mickhansen: @harshithkashyap And what when you use the proper nested order syntax from the docs?

harshithkashyap: @mickhansen I did. I used
```js
order: [Entity,'votes', 'DESC']
```
1. Sequelize generates ``` ORDER BY `Entity`.`votes` DESC; ``` which should be ``` `Entity.votes` DESC ```. Here,  `Entity.votes` should be a single string
2. The order is not applied to the subquery

harshithkashyap:
@mickhansen I've added insert queries & filtered expectations to the test. 
I tried looking into writing SQL unit test but it was a little confusing. This is a specific case when subQuery is thrown, there is an inner join with m:n relation using the through object in the join. Not sure how to generate all the cases. Could you guide me a little? Would be happy to submit the unit test.

harshithkashyap:
@felixfbecker @sushantdhiman Unit Tests were requested by @mickhansen for this PR. But the `generateThroughJoin` method [here](https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L1317) goes through a lot of pre-processing before being called and its very difficult to unit test the same. I've already added integration test for the same.

Incorrent Join identifier is used [here](https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L1452) and i've submitted a patch for the same. Tests pass in my local instance. Travis CI seems to be stuck because of Deadlock. Could you guys restart the same?

harshithkashyap:
@felixfbecker @sushantdhiman @janmeier @mickhansen Hey guys, could you review this please?

felixfbecker:
I am not the right person to approve/reject these changes besides code style

harshithkashyap:
Ping @janmeier

mickhansen:
We have plenty of other unit tests for SQL generation for joins, it should be possible to cover this case aswell.

PR #6059, opened 2016-06-09
ES6 refactor: utils

felixfbecker:
See #6009

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,434 +1,478 @@
 'use strict';
 
-var DataTypes = require('./data-types')
-  , SqlString = require('./sql-string')
-  , _ = require('lodash').runInContext() // Prevent anyone messing with template settings by creating a fresh copy
-  , parameterValidator = require('./utils/parameter-validator')
-  , inflection = require('inflection')
-  , uuid = require('node-uuid')
-  , primitives = ['string', 'number', 'boolean'];
+const DataTypes = require('./data-types');
+const SqlString = require('./sql-string');
+const _ = require('lodash').runInContext(); // Prevent anyone messing with template settings by creating a fresh copy
+const parameterValidator = require('./utils/parameter-validator');
+const inflection = require('inflection');
+const uuid = require('node-uuid');
+const Promise = require('./promise');
+const primitives = ['string', 'number', 'boolean'];
 
-var Utils = module.exports = {
-  inflection: inflection,
-  _: _,
-  camelizeIf: function(string, condition) {
-    var result = string;
+exports.Promise = Promise;
+exports._ = _;
+exports.inflection = inflection;
 
-    if (condition) {
-      result = Utils.camelize(string);
+function camelizeIf(string, condition) {
+  let result = string;
+
+  if (condition) {
+    result = camelize(string);
+  }
+
+  return result;
+}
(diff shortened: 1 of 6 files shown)


mickhansen:
LGTM

lib/utils.js
```diff
@@ -1,444 +1,497 @@
...
+
+function spliceStr(str, index, count, add) {
+  return str.slice(0, index) + add + str.slice(index + count);
+}
+exports.spliceStr = spliceStr;
+
+function camelize(str) {
+  return str.trim().replace(/[-_\s]+(.)?/g, (match, c) => c.toUpperCase());
+}
+exports.camelize = camelize;
```
felixfbecker: `_.camelCase`?

lib/utils.js
```diff
@@ -1,444 +1,497 @@
...
+  }
+  return now;
+}
+exports.now = now;
+
+function tick(func) {
+  const tick = (global.hasOwnProperty('setImmediate') ? global.setImmediate : process.nextTick);
+  tick(func);
+}
+exports.tick = tick;
```
felixfbecker: Can I remove this since we are targeting Node 4?

janmeier: Yep

felixfbecker: It doesn't seem to be used anywhere anyway.

lib/utils.js
janmeier: Or we could just use inflection https://github.com/dreamerslab/node.inflection#inflectioncamelize-str-low_first_letter-

felixfbecker: The only difference I see is that inflection only works on underscore => camelCase while lodash transforms from anything (kebab-case, PascalCase, CONSTANT_CASE) to camelCase. I already checked, `camelize` is used only in `camelizeIf`, so it is super easy to replace. I would vote for lodash's.

janmeier: 2nd'ed

felixfbecker:
Alright, the inflection stuff can be done later. Data types refactor depends on this ^^

felixfbecker:
Haha, seems like the old `camelize` didn't actually convert to camelCase, but kept the first capital letter:
```
2058      + expected - actual
2059
2060       [
2061         "id"
2062         "priority"
2063      -  "productId"
2064      -  "tagId"
2065      +  "ProductId"
2066      +  "TagId"
2067       ]
```

lib/utils.js
```diff
@@ -1,444 +1,491 @@
...
+  s = s.replace(/(\/\*[\n\r\s\S]*?\*\/)/mg, '');
+
+  return s;
+}
+exports.removeCommentsFromFunctionString = removeCommentsFromFunctionString;
+
+function toDefaultValue(value) {
+  if (typeof value === 'function') {
+    const tmp = value();
+    if (tmp instanceof DataTypes.ABSTRACT) {
```
SimonSchick: You can omit the `else`s here as you return in every single block, less code and easier to read imo.

felixfbecker: can't we do that in a different PR? This is really only about ES6. I have other PRs that depend on this one.

janmeier: Yep, lets hold of on style for now - You are welcome to provide a PR @SimonSchick

PR #6061, opened 2016-06-09
fixed limit <= 0 (zero) issue: dialects / PostgreSQL

vstostap:
postgres doesn't queries data when its [LIMIT 0]


mickhansen:
LIMIT 0 means don't fetch anything, i don't believe we should change this to fetch everything.

PR #6062, opened 2016-06-09
[doc] Adds list of options available on sequelize.QueryTypes object.

adam-back:
DOC CHANGE ONLY

### Currently
The current documentation for [Sequelize#QueryTypes](http://sequelize.readthedocs.io/en/latest/api/sequelize/#querytypes) is redundant and doesn't provide context for how to use it. 

screen shot 2016-06-09 at 12 00 50 pm

### Description of change
**Sequelize#QueryTypes**
- Added list of available QueryTypes to Sequelize#QueryTypes.
- Added "See" reference to Sequelize#query where there is a usage example.
- Image:
screen shot 2016-06-09 at 12 44 54 pm

**Sequelize#query**
- Added "See" reference to Sequelize#QueryTypes so users can see their available options.
- Image:
screen shot 2016-06-09 at 12 30 40 pm

### DIFF
--- a/docs/api/sequelize.md
+++ b/docs/api/sequelize.md
@@ -143,7 +143,25 @@ A handy reference to the bluebird Promise class
 ## `QueryTypes`
 [View code](https://github.com/sequelize/sequelize/blob/3e5b8772ef75169685fc96024366bca9958fee63/lib/sequelize.js#L272)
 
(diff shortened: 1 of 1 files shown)


adam-back:
Need to add doc changes to lib/sequelize.js for autogeneration.

PR #6072, opened 2016-06-10
ES6 refactor: data types

felixfbecker:
See #6009

### DIFF
--- a/lib/utils/inherits.js
+++ b/lib/utils/inherits.js
@@ -1,0 +1,14 @@
+'use strict';
+
+const util = require('util');
+const _ = require('lodash');
+
+/** like util.inherits, but also copies over static properties */
+function inherits(constructor, superConstructor) {
+  util.inherits(constructor, superConstructor); // Instance (prototype) methods
+  _.extend(constructor, superConstructor); // Static methods
+}
+
+module.exports = inherits;
+module.exports.inherits = inherits;
+module.exports.default = inherits;
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -1,14 +1,15 @@
 'use strict';
 /*jshint -W110 */
 
-var util = require('util')
-  , _ = require('lodash')
-  , Wkt = require('terraformer-wkt-parser')
-  , sequelizeErrors = require('./errors')
-  , warnings = {}
-  , Validator = require('validator')
-  , momentTz = require('moment-timezone')
-  , moment = require('moment');
+const util = require('util');
+const inherits = require('./utils/inherits');
+const _ = require('lodash');
+const Wkt = require('terraformer-wkt-parser');
+const sequelizeErrors = require('./errors');
+const warnings = {};
+const Validator = require('validator');
+const momentTz = require('moment-timezone');
+const moment = require('moment');
 
 /**
  * A convenience class holding commonly used data types. The datatypes are used when defining a new model using `Sequelize.define`, like this:
@@ -55,50 +56,28 @@ var util = require('util')
  * @class DataTypes
  */
 
-var ABSTRACT = function(options) {
-};
+function ABSTRACT() {}
 
 ABSTRACT.prototype.dialectTypes = '';
 
-ABSTRACT.prototype.toString = function(options) {
+ABSTRACT.prototype.toString = function toString(options) {
   return this.toSql(options);
 };
-ABSTRACT.prototype.toSql = function() {
+ABSTRACT.prototype.toSql = function toSql() {
   return this.key;
 };
-ABSTRACT.warn = function(link, text) {
+ABSTRACT.warn = function warn(link, text) {
   if (!warnings[text]) {
     warnings[text] = true;
     console.warn('>> WARNING:', text, '\n>> Check:', link);
   }
 };
-ABSTRACT.prototype.stringify = function (value, options) {
+ABSTRACT.prototype.stringify = function stringify(value, options) {
   if (this.$stringify) {
     return this.$stringify(value, options);
   }
   return value;
 };
-
-ABSTRACT.inherits = function (Constructor) {
-  var baseType = this;
-
-  if (!Constructor) {
-    Constructor = function () {
-      if (!(this instanceof Constructor)) {
-        var args = [null].concat(arguments);
-        var FactoryFunction = Constructor.bind.apply(Constructor, args);
-        return new FactoryFunction();
-      }
-      baseType.apply(this, arguments);
-    };
-  }
-
-  util.inherits(Constructor, baseType); // Instance (prototype) methods
-  _.extend(Constructor, this); // Static methods
-
-  return Constructor;
-};
-
 
 /**
  * A variable length string. Default length 255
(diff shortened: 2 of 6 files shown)


lib/utils/inherits.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const util = require('util');
+const _ = require('lodash');
+
+/** like util.inherits, but also copies over static properties */
+function inherits(constructor, superConstructor) {
```
janmeier: No reason to have this both in utils.js and utils/

janmeier:
LGTM

lib/utils/inherits.js
felixfbecker: Whoops I thought I removed it from utils.js. Cant use it from there because of cyclic dependency (utils -> data types -> utils.inherits).

alitaheri:
@felixfbecker any particular reason you didn't use es6 classes for the datatypes?

mickhansen:
@alitaheri Not possible if we want to maintain the current syntax for defining data types - Which we might consider changing.

felixfbecker:
Imo we should keep the syntax without `new`, it is much nicer in model definitions.

mickhansen:
@felixfbecker I would tend to agree, pre-4.0 is the time to discuss it if there's anything we'd like to change regarding the syntax or code

alitaheri:
oh, right. I think it's possible to keep the old syntax yet use classes. if each datatype function instantiates a backing class then we won't have an issue right? 
```js
class Abstract { ... }

function ABSTRACT() {
  return new Abstract();
}

class Integer extends Abstract { ... }

function INTEGER(len) {
  return new Integer(len);
}
```

and so on. This way, well keep optimization, inheritance and all the perks of classes and get to keep the old syntax which is lovely :grin:

felixfbecker:
@alitaheri Then you can't check `if (type instanceof INTEGER)` anymore.

alitaheri:
yeah, that's one breaking change :sweat_smile:. I'd say `INTEGER.is(type)` is more expressive anyway. the `is` function will then check against the `Integer` class. it's a breaking change but using classes for datatypes is worth it imo.

felixfbecker:
@alitaheri You would also have to check against the factory function because it is also possible to just pass `Sequlize.STRING` for example. And that would be messy imo, as you would have to define the `is` function for every single class... Also, if you use a factory function, the behaviour would be wrong for all people who currently use `new Sequelize.STRING()`. Imo it is best to keep data types as they are, they are not really "classes", they are more like a DSL.

alitaheri:
`type instanceof INTEGER` still doesn't work with `Sequlize.STRING`.

> as you would have to define the is function for every single class

A helper function can simply take a class and build the is and factory function for us. all abstracted away so this will be all we should do:

```
class Integer { ... }

const INTEGER = buildFactory('INTEGER', Integer);
```

>  if you use a factory function, the behaviour would be wrong for all people who currently use new Sequelize.STRING()

Not really, since all the factory function does is instantiate the class passing the input arguments, they are practically identical.

> Imo it is best to keep data types as they are, they are not really "classes", they are more like a DSL.

I'm not saying they **should** be classes, I'm just saying it's more clear to use classes, that's all. I don't really mind, just raising a discussion :grin:

felixfbecker:
@alitaheri 
> A helper function can simply take a class and build the is and factory function for us.

Things like that suck for static analysis and V8 optimization (as you probably need to use `arguments`). No editor could infer the function signature of the factories, but they can with the current structure.

> Not really, since all the factory function does is instantiate the class passing the input arguments, they are practically identical.

What I mean is if this is the factory:
```js
class StringDataType {
  constructor(length) {
    ...
  }
}
function STRING(length) {
  return new StringDataType(length);
}
exports.STRING = STRING;
```
calling `STRING` with `new` would not make any sense anymore:
```js
type: new Sequelize.STRING(3)
```
as it wouldn't return an instance of `Sequelize.STRING` but of `StringDataType`.

alitaheri:
> Things like that suck for static analysis and V8 optimization (as you probably need to use arguments)

yup, that one's true for static analysis :grin: although. using arguments the right way ( [only length and index](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#what-is-safe-arguments-usage) ) is safe and won't deopt. 

> calling STRING with new would not make any sense anymore:

returning from constructor functions overrides the usual default. it would still work. :grin:

felixfbecker:
@alitaheri Yeah, but if you want to put `is()` on the factory function,
> Be aware that adding properties to functions (e.g. fn.$inject =...) and bound functions (i.e. the result of  Function#bind ) generate hidden classes and, therefore, are not safe when using  #apply .

applies and you would have to copy `arguments` with a for loop. I think it is clear that overall it would make the whole implementation longer and more complex than simply using `if (!(this instanceof STRING)) return new STRING()` like it is now.

> returning from constructor functions overrides the usual default. it would still work. 

Yeah, it would work, but it would be semantically wrong and is definitely an antipattern (when I call `new` with a function, I expect the return to be `instanceof` that function).

alitaheri:
> applies and you would have to copy arguments with a for loop. I think it is clear that overall it would make the whole implementation longer and more complex than simply using if (!(this instanceof STRING)) return new STRING() like it is now.

yes but it's only once for the helper function. The overall cleanliness will be much higher with es6 classes. even if that function is deoptimizied, it will only impact model definition, not each query or method used across the library. Yet the optimizations that can be offered for the classes **will** impact every method and query across the lib, these datatypes are built once and used everywhere, if instances are fast, awesome if instantiation is fast, meh. 

> Yeah, it would work, but it would be semantically wrong and is definitely an antipattern (when I call new with a function, I expect the return to be instanceof that function).

We aren't advertising `new Sequelize.STRING()` but `Sequelize.STRING()`. so the `new` syntax isn't expected to work, although it will, even though it's an anti pattern. no one said `Sequelize.STRING` is a class we can instantiate. I myself found out when I looked at the code. I don't think anyone would use the `new` syntax as it wasn't documented.

Again, I mean no disrespect, I just want to reach the best possible option here, I use this lib in mission critical application run on a cluster of raspberry pies. I need it to be blazing fast :sunglasses:

mickhansen:
fyi if a user defines `STRING` we currently call `new` internally (so that instanceof works)

alitaheri:
@mickhansen The internal instantiation is mostly for accessing instance methods by other parts of the code.

Besides it's not that big a breaking change, people can do a simple find/replace to replace the `instanceof` with a call to `STRING.is(...)`. 

These datatype objects are used the most in the code. If V8 can eventually provide good optimization for them (static dispatch, etc) then it's worth the effort. I'm willing to migrate the datatypes to ES6 Classes if everyone agrees.

felixfbecker:
I don't think it is worth the effort (otherwise I would have done it in my initial PR). I don't think we will get any performance improvement out of it, but it will complicate the code a lot. You have the classes, then you have the factories, and then you have the `is` functions (that you want to patch on the factories?), and that for every dialect. And you have to export both the factories and the classes, because the dialect types need to subclass the basic types. I will also have to completely rewrite the TypeScript definitions (btw, calling the data types _without_ `new` was only very recently added to the definitions, so there are definitely people out there currently calling `type: new Sequelize.ENUM(...)`). So -1 from me.

mickhansen:
@felixfbecker `Sequelize.STRING(11)` without `new` has always worked, since the inception of Sequelize.

@alitaheri No, internal instantiation is used to so we have a normalized instance everywhere :)

felixfbecker:
@mickhansen But not with the TypeScript definitions, because the data types were typed as classes. See https://github.com/louy/typed-sequelize/pull/4 ;)
This was in response to
> We aren't advertising  new Sequelize.STRING()  but  Sequelize.STRING() . so the  new  syntax isn't expected to work, although it will, even though it's an anti pattern. no one said  Sequelize.STRING  is a class we can instantiate. I myself found out when I looked at the code. I don't think anyone would use the  new  syntax as it wasn't documented.

Issue #6069, opened 2016-06-09
setAssociation method for hasOne associations is supposed to accept a primary key as a parameter


arnaudbesnier:
... according to the documentation (see [here](http://docs.sequelizejs.com/en/latest/api/associations/has-one/#setassociationnewassociation-options-promise)) but apparently it does not.

In my code, an error is raised:
`lib/associations/has-one.js:215:79 - TypeError: associatedInstance.get is not a function`

That's because `associatedInstance` is the primary key of the record to associate and the code seems to expect it to be the record instance. I don't know if I am right...

It did not find any tests in the repository for setAssociation with a primary key as parameter.

Thanks for the help.

arnaudbesnier:
Ok I finally found integrations tests on setAssociation with a primary key as a parameter, sorry for that...

The issue is when we try to update an existing association with another object.
I wrote an integration test here: https://github.com/sequelize/sequelize/pull/6075/files

It fails but should not.

PR #6075, opened 2016-06-10
Issue #6069 - Association HasOne - Add a test on setAssociation update that should not fail

arnaudbesnier:
### Pull Request check-list

- [x] Does your issue contain a link to existing issue (Closes [#6069](https://github.com/sequelize/sequelize/issues/6069)) or a description of the issue you are solving?

### Description of change

This tests seems valid to me, but it fails.
I don't have the fix, I just wanted to be sure that the issue ([issue #6069](https://github.com/sequelize/sequelize/issues/6069)) I have in my project in production is real.
Let me know if you need help.

### DIFF
--- a/test/integration/associations/has-one.test.js
+++ b/test/integration/associations/has-one.test.js
@@ -239,6 +239,33 @@ describe(Support.getTestDialectTeaser('HasOne'), function() {
       });
     });
 
+    it('supports updating with a primary key instead of an object', function() {
+      var User = this.sequelize.define('UserXYZ', { username: Sequelize.STRING })
+        , Task = this.sequelize.define('TaskXYZ', { title: Sequelize.STRING });
+
+      User.hasOne(Task);
+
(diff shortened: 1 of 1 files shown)


test/integration/associations/has-one.test.js
```diff
@@ -239,6 +239,33 @@ describe(Support.getTestDialectTeaser('HasOne'), function() {
...
+                  return user.setTaskXYZ(task2.id).then(function() {
+                    return user.getTaskXYZ().then(function(_task) {
+                      expect(_task).not.to.be.null;
+                    });
+                  });
+                });
+              });
+            });
+          });
+        });
```
mickhansen: Holy waterfall, have you considered a bit more idiomatic promise approach?
You can also do promose assertions with chai.

mickhansen:
https://github.com/sequelize/sequelize/blob/master/lib/associations/has-one.js#L215

Confirmed bug / Missing implementation

Issue #6023, opened 2016-06-05
Singularize `as` in foreign keys in To-Many associations (amendment of PR #5957)


overlookmotel:
PR #5957 recently changed the names of foreign keys in associations where an `as` alias is specified.

So now: `Project.hasMany(User, {as: 'Workers'})` creates foreign key `WorkersId` rather than `UserId`.

This is great, but I do think the `as` should be singularized, so foreign key is `WorkerId`. This was already raised by @mickhansen in #5957, but I was too late to the party to get a comment in.

`WorkersId` (plural), in my view, is not what the user would expect. It's inconsistent with the naming when no `as` alias is provided, and with the general Sequelize convention of singularizing/pluralizing. I can imagine it tripping people up quite often because it's counter-intuitive.

Yes, using `inflection` to singularize `as` could in some few cases produce the wrong result, but in those cases the user can override it by specifying `foreignKey`. Personally, I think that's preferable to having to specify `foreignKey` in all cases in order to maintain consistent naming.

I'd be happy to make a PR for this, if there's agreement that this would be a good thing to do.

What do you think @mickhansen @janmeier @sushantdhiman?

sushantdhiman:
We should avoid magically transforming the name to something else, It may cause confusion once again. If user want a specific `foreignKey` they can always explicitly define it. But our transformation should not give any surprises.

mickhansen:
@overlookmotel I definitely agree that `WorkersId` is extremely ugly - But i also remember what a huge pain in the ass inflection has been in the past, it makes it much harder for us to garantee SEMVER when inflection can break in so many subtle ways.

The problem is that it basically forces any reasonable person to ALWAYS use `foreignKey` when using `as` (because most people deploying to production should care how there database looks).

overlookmotel:
Yes I remember being caught out with inflection in the past! But inflection is still being used for pluralization within Sequelize e.g. with one-to-many relationships `User.findAll({include: Order})` returns the orders as `Orders` (plural). So Sequelize is dependent on inflection already.

I totally agree that obeying semver is important. I'd suggest the way to guarantee that is to pin the version of inflection and only update at major version bumps of Sequelize.

@sushantdhiman:
> But our transformation should not give any surprises.

I agree. But I think getting a foreign key `WorkersId` *is* surprising!

I totally see that there are arguments on both sides and of course I go with the consensus. But I do feel that this could be the cause of users posting issues "why am I getting `WorkersId`?" which is going to be an annoyance.

[...]

PR #6077, opened 2016-06-11
Inflection

overlookmotel:
### Pull Request check-list

### Description of change

As discussed in #6023.

This PR:

* Updates [inflection](https://www.npmjs.com/package/inflection) dependency to latest and pins the version
* Exports all used [inflection](https://www.npmjs.com/package/inflection) methods as methods of `Utils`
* Updates codebase to use e.g. `Utils.underscore()` rather than `inflection.underscore()`
* Allows a later version of [inflection](https://www.npmjs.com/package/inflection) to be set by user with `Sequelize.Utils.inflection = require('inflection')` and `Utils.pluralize()` etc will use this overriden version of [inflection](https://www.npmjs.com/package/inflection).

I will add to the [test shims](https://github.com/sequelize/sequelize/blob/master/test/supportShim.js) to check that `inflection` is not accessed directly once PR #5836 is merged.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 - [ADDED] Support for range operators [#5990](https://github.com/sequelize/sequelize/issues/5990)
 - [FIXED] Broken transactions in `MySQL` [#3568](https://github.com/sequelize/sequelize/issues/3568)
 - [FIXED] `Model.count` don't include attributes [#5057](https://github.com/sequelize/sequelize/issues/5057)
+- [INTERNALS] Updated `inflection` dependency and pinned version
 
 ## BC breaks:
 - Range type bounds now default to [postgres default](https://www.postgresql.org/docs/9.5/static/rangetypes.html#RANGETYPES-CONSTRUCT) `[)` (inclusive, exclusive), previously was `()` (exclusive, exclusive)
--- a/changelog.md
+++ b/changelog.md
@@ -3,7 +3,7 @@
 - [ADDED] Support for range operators [#5990](https://github.com/sequelize/sequelize/issues/5990)
 - [FIXED] Broken transactions in `MySQL` [#3568](https://github.com/sequelize/sequelize/issues/3568)
 - [FIXED] `Model.count` don't include attributes [#5057](https://github.com/sequelize/sequelize/issues/5057)
-- [INTERNALS] Updated `inflection` dependency and pinned version
+- [INTERNALS] Updated `inflection` dependency and pinned version and expose all used `inflection` methods on `Utils`
 
 ## BC breaks:
 - Range type bounds now default to [postgres default](https://www.postgresql.org/docs/9.5/static/rangetypes.html#RANGETYPES-CONSTRUCT) `[)` (inclusive, exclusive), previously was `()` (exclusive, exclusive)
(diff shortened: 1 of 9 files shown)


felixfbecker:
What is wrong with `Utils.inflection.underscore()`?

overlookmotel:
@felixfbecker The idea was to decouple inflection a bit and make it consistent how `inflection` methods are called.

I mean what's wrong with `Utils.inflection.pluralize()` vs `Utils.pluralize()`? But it's been `Utils.pluralize()` for a long time.

felixfbecker:
Just thought always using `Utils.inflection.pluralize` would make it 100% clear that a user can replace it.

felixfbecker:
You changed it in a couple of places, so if we are gonna make it consistent, we might aswell change it the other way... That `exports.inflection` is a bit ugly and won't work with true ES6 modules (might already land in Node 7)

overlookmotel:
> Just thought always using Utils.inflection.pluralize would make it 100% clear that a user can replace it.

Yes, but that also makes it tempting to do...

```js
var Utils = require('./Utils');
var inflection = Utils.inflection;

...

function xxx() {
    ...
    inflection.pluralize() ...
}
```

Which then breaks the ability to replace `Utils.inflection` externally.

overlookmotel:
Look is this really important enough to argue about?

felixfbecker:
No 😃

overlookmotel:
@felixfbecker Sorry I was grumpy! Of course it's always worth discussing before things get folded into the codebase.

How strongly do you feel about this?

overlookmotel:
@felixfbecker Yes I agree that `exports.inflection.pluralize()` is ugly as hell. But I wasn't sure what else to do.

Usually, I'd make up a utils module with this kind of form:

```js
var Utils = {
    inflection: require('inflection'),

    pluralize: function(str) {
        return Utils.inflection.pluralize(str);
    }
};

module.exports = Utils;
```

That's less ugly to my eyes. But the whole structure of `lib/utils.js` is flat, so `exports.inflection` was best I could come up with. Do you have another suggestion?

felixfbecker:
@overlookmotel nevermind :) I don't feel strongly about this at all. Also I was the one who changed it to `exports.whatever = ` instead of `module.exports` because it is more ES6-ish. Which is also a point that should be considered: How will overriding inflection work in ES6?
```js
import * as Utils from 'sequelize/lib/utils';
import * as inflection from 'inflection';
Utils.inflection = inflection;
```
Will only update the local import AFAIK, not what sequelize is using. One option would be something like this:
```js
let inflection = require('inflection');

function useInflection(replacement) {
  inflection = replacement;
}
exports.useInflection = useInflection;

function pluralize(str) {
  return inflection.pluralize(str);
}
exports.pluralize = pluralize;
```
`useInflection` could be re-exported in the main sequelize module so users don't have to import sequelize internals. Opinions?

janmeier:
Exporting a function to explicitly override inflection might be cleaner, yeah

overlookmotel:
@felixfbecker @janmeier Exporting a function sounds good to me.

overlookmotel:
PR updated accordingly. `inflection` is still exported as `Sequelize.Utils.inflection`. Should we remove that now?

felixfbecker:
Yes, because that instance cannot be replaced. All calls should go through `Utils.pluralize` etc.

overlookmotel:
@felixfbecker Agreed. PR amended accordingly.

mickhansen:
LGTM

overlookmotel:
@mickhansen Rebased on master. Ready to merge...

overlookmotel:
Thanks @mickhansen @janmeier @felixfbecker

PR #6082, opened 2016-06-11
ES6 refactor of validator-extras.js

felixfbecker:
See #6009 

One thing I would like to discuss is if we can somehow avoid patching the global validator.

### DIFF
--- a/lib/utils/validator-extras.js
+++ b/lib/utils/validator-extras.js
@@ -1,77 +1,80 @@
 'use strict';
 
-var validator = require('validator')
-  , _ = require('lodash');
+const validator = require('validator');
+const _ = require('lodash');
 
-var extensions = {
-  extend: function(name, fn) {
+const extensions = {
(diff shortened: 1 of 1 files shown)


janmeier:
We should definitely avoid patching the global validator - Actually we'd like to remove specific validations completely from core, and only have a validate function that users can hook into - validator.js validations can then be provided as a plugin, and sequelize core won't break when validator changes their API

felixfbecker:
Like this?
```js
import {Model} from 'sequelize';
import {isEmail} from 'sequelize/lib/validations';

class MyModel extends Model {}
MyModel.init({
  email: {
    type: Sequelize.STRING,
    validate: {isEmail}
  }
})
```

mickhansen:
@felixfbecker I was actually thinking more than `validate` can hold ANY value, and plugins then validate via a hook and lookup the options internally.

mickhansen:
Merging for now.

PR #6090, opened 2016-06-11
Fix intermittently failing test on MSSQL

overlookmotel:
### Pull Request check-list

### Description of change

This PR fixes a test which intermittently fails on MSSQL on Travis. e.g.: https://travis-ci.org/sequelize/sequelize/jobs/136937830

The issue is that the order in which the `User` instances is created in this test isn't necessarily the same as the order in the code. Most of the time it is, but sometimes not and so sometime the `id`s of the Users are in an unexpected order. When this happens the test fails. Hence failure which is only intermittent.

There doesn't appear to be anything wrong with the actual behavior that the test is testing for.


mickhansen:
Thanks @overlookmotel :)

PR #6093, opened 2016-06-12
Make new Model() identical to Model.build()

felixfbecker:
This moves all logic from `Model.build()` to the constructor, so they accept the exact same parameters. The only difference is `build()` can also take an array of values. Also makes `values` optional.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1701,36 +1701,19 @@ class Model {
   }
 
   /**
-   * Builds a new model instance. Values is an object of key value pairs, must be defined but can be empty.
-
-   * @param {Object}  values
+   * Builds a new model instance.
+   *
+   * @param {Object}  [(values|values[])={}] An object of key value pairs or an array of such. If an array, the function will return an array of instances.
    * @param {Object}  [options]
    * @param {Boolean} [options.raw=false] If set to true, values will ignore field and virtual setters.
    * @param {Boolean} [options.isNewRecord=true]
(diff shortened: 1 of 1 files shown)


lib/model.js
```diff
@@ -2623,7 +2606,34 @@ class Model {
...
+   * @param {Boolean} [options.raw=false] If set to true, values will ignore field and virtual setters.
+   * @param {Boolean} [options.isNewRecord=true]
+   * @param {Array}   [options.include] an array of include options - Used to build prefetched/included model instances. See `set`
+   */
   constructor(values, options) {
+    values = values || {};
+    options = _.extend({
+      isNewRecord: true,
+      $schema: this.$schema,
+      $schemaDelimiter: this.$schemaDelimiter
```
mickhansen: `this` needs to be changed here

lib/model.js
```diff
@@ -2623,7 +2606,34 @@ class Model {
...
+    }, options || {});
+
+    if (options.attributes) {
+      options.attributes = options.attributes.map(attribute => Array.isArray(attribute) ? attribute[1] : attribute);
+    }
+
+    if (!options.includeValidated) {
+      this._conformOptions(options, this);
```
mickhansen: same here, second arg `this` should be the model

lib/model.js
felixfbecker: Silly me, I just moved the code, commited and opened PR without thinking. You are right of course.

PR #6101, opened 2016-06-14
Closes #3019, Model.count should not return NaN

### LINKED ISSUE #3019: Model.count returning NaN because Mysql returns an empty string

sushantdhiman:
### Pull Request check-list

### Description of change

Closes #3019, Issue seems to be fixed already. Test case to prevent regression.

### DIFF
--- a/test/integration/model/count.test.js
+++ b/test/integration/model/count.test.js
@@ -10,7 +10,8 @@ var chai = require('chai')
 describe(Support.getTestDialectTeaser('Model'), function() {
   beforeEach(function() {
     this.User = this.sequelize.define('User', {
-      username: DataTypes.STRING
+      username: DataTypes.STRING,
+      age: DataTypes.INTEGER
     });
     this.Project = this.sequelize.define('Project', {
       name: DataTypes.STRING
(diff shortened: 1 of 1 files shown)


janmeier:
The issue mentions a grouping that returns an empty string - Which we try to parse as an int

sushantdhiman:
Thanks @janmeier

sushantdhiman:
Should it return `Integer` or `String`, keeping in mind large counts ?

sushantdhiman:
**Cases**

1. So when we use group, its not possible to get `null` record like this `[ { count: null }, { count: 3} ]`, group by will not return it. It will result into `[ { count: 3 } ]` ( ignored `null` )
2. But its still possible to have `[ ]` when `group` doesn't return anything.
3. When a empty result condition run ( without `group`) then result is `0` `Integer`

Case **2** should keep using `[]` because user must be expecting `Array` with group.

Your comments @mickhansen @janmeier

janmeier:
Agreed with regards to 2.

janmeier:
Yay, github now sends notification emails on new pushes to PRs :tada: !

sushantdhiman:
Don't worry I'll still link you to all PRs :laughing:

PR #6102, opened 2016-06-14
Include as rewrite

LoicMahieu:
### Pull Request check-list

Not necessary I think. Documentation already refers the `options.include[].as` and `options.include[].association`.

### Description of change

It allows us to `include` the same association multiple times with a different `as`.
Example:
```js
user.findAll({
  include: [
    { association: User.Tasks, as: 'someTasks', where: { name: { $like: 'bar' } },
    { association: User.Tasks, as: 'otherTasks', where: { name: { $like: 'foo' } }
  ]
})
```

My user case: I use [graphql-sequelize](https://github.com/mickhansen/graphql-sequelize) and it generates `include` for nested object/properties. I had to apply a `where` clause on a relationship and I can't modify the options after the nested `include`.

Thanks for review.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# Future
+- [FIXED] Fixed an issue where custom-named model fields break when offsetting, ordering, and including hasMany simultaneously. [#5985] (https://github.com/sequelize/sequelize/issues/5985) 
+
 # 3.23.3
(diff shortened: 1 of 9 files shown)


LoicMahieu:
Bad branch reference. This PR is for v3. Closed for #6103

PR #6103, opened 2016-06-14
model: only rewrite `include.as` if necessary

LoicMahieu:
### Pull Request check-list

Not necessary I think. Documentation already refers the `options.include[].as` and `options.include[].association`.
### Description of change

It allows us to `include` the same association multiple times with a different `as`.
Example:

``` js
user.findAll({
  include: [
    { association: User.Tasks, as: 'someTasks', where: { name: { $like: 'bar' } },
    { association: User.Tasks, as: 'otherTasks', where: { name: { $like: 'foo' } }
  ]
})
```

My user case: I use [graphql-sequelize](https://github.com/mickhansen/graphql-sequelize) and it generates `include` for nested object/properties. I had to apply a `where` clause on a relationship and I can't modify the options after the nested `include`.
### Note

PR also applies for v4 but I am waiting review before submitting another PR.

Thanks for review.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -551,7 +551,9 @@ validateIncludedElement = function(include, tableNames, options) {
   }
 
   include.association = association;
-  include.as = association.as;
+  if (!include.as) {
+    include.as = association.as;
+  }
 
   // If through, we create a pseudo child include, to ease our parsing later on
   if (include.association.through && Object(include.association.through.model) === include.association.through.model) {
--- a/test/integration/model/find.test.js
+++ b/test/integration/model/find.test.js
@@ -569,7 +569,7 @@ describe(Support.getTestDialectTeaser('Model'), function() {
             });
           });
 
-           it('allows mulitple assocations of the same model with different alias', function() {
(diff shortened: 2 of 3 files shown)


janmeier:
This looks good to me, what do you think @mickhansen 

Passing your own `as` should only be allowed if `assocation` is set, or if source only has a single association to target though.

mickhansen:
@janmeier I'm on the fence, i think it's ok for `association` though.

mickhansen:
The problem is of course that `as` now has two behaviours, which is never a good thing

janmeier:
Only allowing it on combination with `association` might be a good idea.

Perhaps we should also change all docs examples to use `association`, then we wouldn't have to explain the `as` part as it works now - Which a lot of people has problems with. If we fix the problem with BTM being saved wrong in `model.assocations` we can simply refer to that (instead of having to save it manually)

poldridge:
What's the status on this? Is there a version of sequelize where joining twice on the same association is possible?

nadavoosh:
Does anyone have a good workaround for this type of query?

PR #6106, opened 2016-06-14
ES6 conversion for association / hooks

sushantdhiman:
### Pull Request check-list

### Description of change

Convert `integration/association`  and `integration/hooks` to ES6. Only these two section are properly grouped so can be directly converted to ES6.

Other section would need regrouping then conversion.

### DIFF
--- a/test/integration/associations/belongs-to.test.js
+++ b/test/integration/associations/belongs-to.test.js
@@ -1,20 +1,18 @@
 'use strict';
 
 /* jshint -W030 */
-var chai = require('chai')
-  , sinon = require('sinon')
-  , expect = chai.expect
-  , Support = require(__dirname + '/../support')
-  , DataTypes = require(__dirname + '/../../../lib/data-types')
-  , Sequelize = require('../../../index')
-  , Promise = Sequelize.Promise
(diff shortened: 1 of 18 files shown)


test/integration/associations/belongs-to.test.js
```diff
@@ -24,330 +22,218 @@ describe(Support.getTestDialectTeaser('BelongsTo'), function() {
...
-                          });
-                        });
-                      });
-                    });
-                  });
-                });
-              });
-            });
-          });
+          return sequelize.sync({ force: true }).then(() => User.create({ username: 'foo' }).then(user => Group.create({ name: 'bar' }).then(group => sequelize.transaction().then(t => group.setUser(user, { transaction: t }).then(() => Group.all().then(groups => groups[0].getUser().then(associatedUser => {
```
janmeier: Even though its all just setup, I'm not sure I like having it all on a single line

mickhansen: No, it's definitely close to unreadable - Consider refactoring to bluebird-coroutine instead perhaps.

sushantdhiman:
I have changed the formatting now, it will be more readable now

Issue #5898, opened 2016-05-16
4.0


felixfbecker:
I wanted to push the discussion around 4.0 a little bit, summarize what is left to decide so 4.0 can be developed / released, and make a proposal:
- Support only Node >4 #5205. Benefits: native classes, native Promises, arrow functions. Implies #5878 
- ES6 in Sequelize codebase #3951 (without Babel, classes, arrow functions)
- Make instances `instanceof Model` (with native classes) #5877 
  - Better OOP semantics
  - `define()` still works as Model factory
  - users can extend `Sequelize.Model`
    - awesome for type-safety / autocompletion
    - alternative to `classMethods` / `instanceMethods` / `getterMethods` / `setterMethods`
  - `Model.init()` static initializer attaches attribute metadata to the model
    - third-party module can define class/attribute decorators that does this for users of TypeScript / Babel - decorators are the perfect semantic for defining additional metadata for Sequelize on the attributes
  - Need some way to register a model to a `Sequelize` instance / model manager, `sequelize.add(Model)`?

Other than that, keep the API / concept changes small. Let users still use `define()` like they are used to, but _enable_ users of TS / Babel to explore new ways of defining models.

janmeier:
FYI - I've moved 3.0 to a separate branch (https://github.com/sequelize/sequelize/tree/v3), and made master 4.0.0-1. I've also started merging some PRs with minor BC breaks ([see changelog](https://github.com/sequelize/sequelize/blob/master/changelog.md#bc-breaks)) so we can hopefully get the 4.0 ball rolling.

The idea as that all new development is aimed at 4.0, but if people badly need a feature / bugfix in 3.0 they are welcome to create a backport PR. But we in the core team (Mick and I) at least will only work on 4.0.

I know some people will probably be annoyed by us "cutting" support for 3.0, but we simply don't have the resources to maintain two separate branches. We tried that with the 2.0 -> 3.0 transition and hated it :(.

With regards to node version I'm leaning mostly towards 6 only, or perhaps 6 + 0.12 with babel. I'd really like to use coroutines and generators, but that would force 0.12 users to use `--harmony`. But I think the increased readability of the code would make up for it.

We have been saving up quite a lot of issues with BC breaks that we want to land in 4.0 https://github.com/sequelize/sequelize/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+label%3Asemver%3Amajor+

mickhansen:
[...]

PR #6108, opened 2016-06-14
ES6 conversion of dialect tests

sushantdhiman:
### Pull Request check-list

### Description of change

`integration/dialect` these test are grouped properly but many tests are still scattered in some other big tests. I haven't touched them yet. This PR only convert these tests to ES6.

In next PRs I would move more dialect specific tests to these sections.

### DIFF
--- a/test/integration/dialects/abstract/connection-manager.test.js
+++ b/test/integration/dialects/abstract/connection-manager.test.js
@@ -1,99 +1,99 @@
 'use strict';
 
 /* jshint -W030 */
-var chai = require('chai')
-  , expect = chai.expect
-  , Support = require(__dirname + '/../../support')
-  , sinon = require('sinon')
-  , Config = require(__dirname + '/../../../config/config')
-  , ConnectionManager = require(__dirname + '/../../../../lib/dialects/abstract/connection-manager')
-  , Pooling = require('generic-pool')
-  , _ = require('lodash')
-  , Promise = require(__dirname + '/../../../../lib/promise');
+const chai = require('chai');
+const sinon = require('sinon');
+const _ = require('lodash');
+const Pooling = require('generic-pool');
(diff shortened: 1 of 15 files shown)


mickhansen:
We really ought to consider implement bluebird-coroutines when we do to es6, otherwise i don't see a huge value in converting the tests - It does take up review time :)

sushantdhiman:
Ok I'll start converting these to coroutines

overlookmotel:
@sushantdhiman Hold off a sec! Please see my comment here https://github.com/sequelize/sequelize/issues/5898#issuecomment-221346901 about coroutines not working with CLS, which is a documented feature of Sequelize and I think needs to be retained.

I've raided an issue #6116 about preparing for coroutines. But in my opinion, we're not ready to go for it with coroutines until the problems have been ironed out. I feel that ensuring CLS support is 100% is really important because it's used for running queries in the right transactions - so a bug threatens to undermine consistency, which I think is a pretty big deal.

This is really annoying, as coroutines are going to be a massive boost - personally I can't wait until the Sequelize codebase uses them as it'll make the code much more readable - but I feel we should hold off for now.

I'm on the case to try to fix the CLS/bluebird problems and write a full set of tests for it (see [here](https://github.com/overlookmotel/cls-bluebird-test)). And I'll sort it as soon as I can.

Of course my word isn't final! I'm just a contributor like you. But I do feel this is really important.

Please give your opinion on #6116.

sushantdhiman:
@overlookmotel It didn't make sense that test should not use coroutine because `lib` don't use them. This mean no one is able to use coroutine in production because of may be broken support of CLS. Because test are just calling the library API surface.

If there is any issue with transaction related support we can hold for that part.

Regardless I have deferred working on ES6 conversion for now as there are more pressing issues to solve.

PR #3740, opened 2015-05-17
WIP: Mysql: Fix Model.count returning NaN. Closes #3019

BridgeAR:
This will fix DOUBLE and REAL being returned as numbers instead of strings. This will also fix the issue with iojs 2.0 and the missing docs entry that .count supports paranoid. 
I also updated some internal dependencies and changed some to get patch updates instead of being fixed.

PR #6114, opened 2016-06-15
Fixes #4840, Model.count allow passing col to count on

### LINKED ISSUE #4442: Specify the column for .count
I am trying this 
<code>
return Model.count({distinct:'col_name',where:{some condition}});
</code>
It isn't working and the corresponding sql query that it produces is:
<code>
SELECT count(DISTINCT(*)) AS `count` FROM `mail_box` AS `mail_box` WHERE `mail_box`.`to_username` = 'nikhil9999';
</code>

### LINKED ISSUE #4840: Is distinct option for Model.count() actually working?
I have a Postgres database. In [latest docs](http://docs.sequelizejs.com/en/latest/api/model/#countoptions-promiseinteger), `distinct` is listed under options that can be passed to `Model.count()` function. However, as `col` is [explicitely set](https://github.com/sequelize/sequelize/blob/0de404640d4c71e2d1f1259356650dfb586a248b/lib/model.js#L1569) to `'*'` in this function implementation, resulting query looks as follows
`SELECT COUNT(DISTINCT(*)) ...`
which doesn't seem to be a valid query:
```
Executing (default): SELECT count(DISTINCT(*)) AS "count" FROM "Comments" AS "Comment";
Wed Nov 11 2015 21:23:23 GMT+0100 (CET) 'SequelizeDatabaseError' 'syntax error at or near "*"'
'SequelizeDatabaseError: syntax error at or near "*"
    at Query.formatError (/node_modules/sequelize/lib/dialects/postgres/query.js:443:14)
    at null.<anonymous> (/node_modules/sequelize/lib/dialects/postgres/query.js:118:19)
    at emitOne (events.js:77:13)
    at emit (events.js:169:7)
    at Query.handleError (/node_modules/pg/lib/query.js:108:8)
    at null.<anonymous> (/node_modules/pg/lib/client.js:171:26)
    at emitOne (events.js:77:13)
    at emit (events.js:169:7)
    at Socket.<anonymous> (/node_modules/pg/lib/connection.js:109:12)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)
    at Socket.Readable.push (_stream_readable.js:110:10)
    at TCP.onread (net.js:523:20)'
```

Sample code resulting in the error:
```JS
var Comment = sequelize.define('Comment', {
        author: Sequelize.STRING,
        text: Sequelize.TEXT
    });

var uniqueAuthors = Comment.count({
        distinct: true
    });
```

I tried also passing `where` or `attributes` hoping that this will change the resulting column, but the result is still the same. So, what is the proper way to use this option?

sushantdhiman:
### Pull Request check-list

### Description of change

Closes #4442 
Fixes #4840 
Closes #3740

`distinct` option with `Model.count` was broken because it generates a SQL query like 

`COUNT( DISTINCT(*) )` which is incorrect.

As noted in issues mentioned above, we should allow an option to pass the `col` so `distinct` can work as expected.

PR #3740 can be closed because #3019 is closed and missing `Model.count` `paranoid` documentation is added in this PR

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,8 @@
 - [INTERNALS] Updated `inflection` dependency and pinned version and expose all used `inflection` methods on `Utils`
 - [ADDED] `Sequelize.useInflection` method
 - [FIXED] `hasOne` throws error on update with a primary key [#6069](https://github.com/sequelize/sequelize/issues/6069)
+- [FIXED] `Model.count` gives SQL syntax error when using `distinct` [#4840](https://github.com/sequelize/sequelize/issues/4840)
+- [ADDED] `Model.count` now allow specifying column to count on, use `options.col` [#4442](https://github.com/sequelize/sequelize/issues/4442)
 
 ## BC breaks:
 - Range type bounds now default to [postgres default](https://www.postgresql.org/docs/9.5/static/rangetypes.html#RANGETYPES-CONSTRUCT) `[)` (inclusive, exclusive), previously was `()` (exclusive, exclusive)
--- a/lib/model.js
+++ b/lib/model.js
@@ -1530,7 +1530,9 @@ class Model {
    * @param {Object}        [options]
    * @param {Object}        [options.where] A hash of search attributes.
    * @param {Object}        [options.include] Include options. See `find` for details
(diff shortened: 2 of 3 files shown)


janmeier:
Sweet! :)

PR #6115, opened 2016-06-15
associations: has-many: fix possible ambiguous field in count()

LoicMahieu:
### Pull Request check-list
### Description of change

`has-many` `count()` method does not alias the field. In complex query, it is possible to get a `Query error with ambiguous column name in SQL` error.
Notice that `belongs-to-many` use already an alias, see https://github.com/sequelize/sequelize/blob/v3/lib/associations/belongs-to-many.js#L499

### DIFF
--- a/lib/associations/has-many.js
+++ b/lib/associations/has-many.js
@@ -342,7 +342,7 @@ HasMany.prototype.count = function(instance, options) {
 
   options = Utils.cloneDeep(options);
   options.attributes = [
-    [sequelize.fn('COUNT', sequelize.col(model.primaryKeyField)), 'count']
+    [sequelize.fn('COUNT', sequelize.col([model.name, model.primaryKeyAttribute].join('.'))), 'count']
   ];
   options.raw = true;
   options.plain = true;
--- a/test/integration/associations/has-many.test.js
+++ b/test/integration/associations/has-many.test.js
@@ -928,6 +928,46 @@ describe(Support.getTestDialectTeaser('HasMany'), function() {
 
         return expect(this.user.countActiveTasks({})).to.eventually.equal(1);
       });
+
+      it('should not fail due to ambiguous field', function () {
+        var self = this;
+        var subtasks = self.Task.hasMany(this.Task, {
+          as: 'subtasks'
+        });
+
+        return expect(
+          self.sequelize.sync({ force: true })
+            .then(function () {
+              return self.User.create({
+                username: 'John',
+                Tasks: [{
+                  title: 'Get rich', active: true
+                }]
+              }, {
+                include: [self.Task]
+              });
+            })
+            .then(function (user) {
+              return user.getTasks();
+            })
+            .then(function (tasks) {
+              return tasks[0].createSubtask({ title: 'Make a startup', active: false })
+                .then(function () {
+                  return tasks[0].createSubtask({ title: 'Engage rock stars', active: true });
+                });
+            })
+            .then(function () {
+              return self.user.countTasks({
+                include: [{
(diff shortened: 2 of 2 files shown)


janmeier:
Lint errors

lib/associations/has-many.js
```diff
@@ -342,7 +342,7 @@ HasMany.prototype.count = function(instance, options) {
 
   options = Utils.cloneDeep(options);
   options.attributes = [
-    [sequelize.fn('COUNT', sequelize.col(model.primaryKeyField)), 'count']
+    [sequelize.fn('COUNT', sequelize.col([model.target.name, model.primaryKeyAttribute].join('.')), 'count']
```
janmeier: I think this should still be `primaryKeyField

LoicMahieu:
Fixed lint.

mickhansen:
Please add a test, ideally one that would also have a differing tablename from model name

test/integration/associations/has-many.test.js
```diff
@@ -928,6 +928,46 @@ describe(Support.getTestDialectTeaser('HasMany'), function() {
...
+                include: [{
+                  association: subtasks,
+                  where: {
+                    active: true
+                  }
+                }]
+              });
+            })
+        ).to.eventually.equal(1);
+      });
```
janmeier: I code golfed a bit, and reduced the size using a couple of arrow functions and promise.join

```js
it('should not fail due to ambiguous field', function () {
  var subtasks = self.Task.hasMany(this.Task, {
    as: 'subtasks'
  });

  return this.sequelize.sync({ force: true }).then(() => {
    return this.User.create({
      username: 'John',
      Tasks: [{
        title: 'Get rich', active: true
      }]
    }, {
      include: [self.Task]
    });
  }).then(user => {
    return Promise.join(
      user.get('Tasks')[0].createSubtask({ title: 'Make a startup', active: false }),
      user.get('Tasks')[0].createSubtask({ title: 'Engage rock stars', active: true })
    ).return(user);
  }).then(user => {
    return expect(user.countTasks({
      attributes: [],
      include: [{
        association: subtasks,
        where: {
          active: true
        }
      }]
    })).to.eventually.equal(1);
  })
});
```

I also added `attributes: []` to fix `missing aggregate or group by` in postgres so, should be green now

lib/associations/has-many.js
```diff
@@ -342,7 +342,7 @@ HasMany.prototype.count = function(instance, options) {
 
   options = Utils.cloneDeep(options);
   options.attributes = [
-    [sequelize.fn('COUNT', sequelize.col(model.primaryKeyField)), 'count']
+    [sequelize.fn('COUNT', sequelize.col([model.name, model.primaryKeyAttribute].join('.'))), 'count']
```
janmeier: Pretty sure this should still be `primaryKeyField`, not attribute

janmeier:
LGTM, once you ammend the test and add a changelog entry this should be good

We also really appreciate if you forward-port this to master :)

jayprakash1:
This has been pending for a while and it does not seem fixed in the latest code. Any chance of this getting fixed?

janmeier:
@jayprakash1 You are welcome to pick up the work, rebase it and fix tests and changelog

test/integration/associations/has-many.test.js
felixfbecker: Why not just commit on his branch?

lib/associations/has-many.js
felixfbecker: Shouldn't this be `model.getTableName()` and not `model.name`?

PR #6117, opened 2016-06-16
Fix #2852, Added DEBUG support and intensive connection logging

### LINKED ISSUE #851: Intensive connection logging
Is there possible in Sequelize to log all connection events ?
I mean:
- connect
- disconnect
- reconnect after failure
- draining connection
- error

### LINKED ISSUE #2852: DEBUG support
More verbose debugging support.

Everything from hooks being called to calls.

sushantdhiman:
### Pull Request check-list

### Description of change

Closes #2852
Closes #851 

Added support for logging via DEBUG module.
Also used `depd` module to show deprecation warning. Setup a module for handling all logging tasks.

- [x] Query pre / post execution
- [x] Hooks removal / addition / running
- [x] Connection connection / disconnect / error ( all dialects)
- [x] Pool init / drain / release

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -215,7 +215,7 @@ class ConnectionManager {
     return promise.then(() => new Promise((resolve, reject) => {
       this.pool.acquire((err, connection) => {
         if (err) return reject(err);
-        debug(`connection acquire`);
+        debug(`connection acquired`);
         resolve(connection);
       }, options.priority, options.type, options.useMaster);
     }));
@@ -224,7 +224,7 @@ class ConnectionManager {
   releaseConnection(connection) {
     return new Promise((resolve, reject) => {
       this.pool.release(connection);
-      debug(`connection release`);
+      debug(`connection released`);
       resolve();
     });
   }
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -3,6 +3,7 @@
 const Pooling = require('generic-pool');
 const Promise = require('../../promise');
 const _ = require('lodash');
+const Utils = require('../../utils');
 const semver = require('semver');
 const defaultPoolingConfig = {
   max: 5,
@@ -51,6 +52,7 @@ class ConnectionManager {
   onProcessExit() {
     if (this.pool) {
       this.pool.drain(() => {
+        Utils.getLogger().debug(`connection drain via pool due to process exit`);
         this.pool.destroyAllNow();
       });
     }
@@ -75,10 +77,12 @@ class ConnectionManager {
         name: 'sequelize-connection',
         create: (callback) => {
           this.$connect(config).nodeify((err, connection) => {
+            Utils.getLogger().debug(`pool created max/min: ${config.pool.max}/${config.pool.min} with no replication`);
             callback(err, connection); // For some reason this is needed, else generic-pool things err is a connection or some shit
           });
         },
         destroy: (connection) => {
+          Utils.getLogger().debug(`connection destroy via pool`);
           this.$disconnect(connection);
           return null;
         },
@@ -122,9 +126,11 @@ class ConnectionManager {
         }
       },
       destroy: (connection) => {
+        Utils.getLogger().debug(`connection destroy via pool`);
         return this.pool[connection.queryType].destroy(connection);
       },
       destroyAllNow: () => {
+        Utils.getLogger().debug(`all connection destroy via pool`);
         this.pool.read.destroyAllNow();
         this.pool.write.destroyAllNow();
       },
@@ -208,6 +214,7 @@ class ConnectionManager {
     return promise.then(() => new Promise((resolve, reject) => {
       this.pool.acquire((err, connection) => {
         if (err) return reject(err);
+        Utils.getLogger().debug(`connection acquire via pool`);
         resolve(connection);
       }, options.priority, options.type, options.useMaster);
     }));
@@ -216,6 +223,7 @@ class ConnectionManager {
   releaseConnection(connection) {
     return new Promise((resolve, reject) => {
       this.pool.release(connection);
+      Utils.getLogger().debug(`connection release via pool`);
       resolve();
     });
   }
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -52,7 +52,7 @@ class ConnectionManager {
   onProcessExit() {
     if (this.pool) {
       this.pool.drain(() => {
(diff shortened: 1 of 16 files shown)


mickhansen:
Would the common `debug` module not be sufficient here?
Plan to add more extensive tracing as a seperate module soon in any case.

sushantdhiman:
What do you mean by common `debug` module ? Like `node debug`, I think current module will be useful as it has more option for extra logging. Like different level of loggings etc.

mickhansen:
https://www.npmjs.com/package/debug

It's almost a defacto standard for debugging module code.

sushantdhiman:
I have used same module @mickhansen

mickhansen:
So you have, got caught up in the abstraction i guess.

sushantdhiman:
So we can merge it ? Or do you want any improvements

sushantdhiman:
I added this separate module so we may be able to add different logging levels or any customizations. If you suggest , I can just add two utility methods called `deprecate` and `debug`, we can use them in entire code base. But I just wanted to keep the code separate from current utility base so it can be changed easily.

lib/dialects/abstract/connection-manager.js
```diff
@@ -52,7 +52,7 @@ class ConnectionManager {
   onProcessExit() {
     if (this.pool) {
       this.pool.drain(() => {
-        Utils.getLogger().debug(`connection drain via pool due to process exit`);
+        Utils.debug(`connection drain via pool due to process exit`);
```
mickhansen: any way to kill the Utils prefix?

sushantdhiman: We need to maintain same context of `debug` so it should be a global module. I can directly require `logger` module but we will have to access its methods ( `debug` and `deprecate` ) in similar manner.

mickhansen: debug should ideally be initiated by file so you have prefix like `sequelize:connection` or `sequelize:model`, atleast normally - however ideally our prefixed would be named for model names here.

I'd be fine with `const = require('Utils').debug()` even, i just hate seeing Utils prefixed scattered around the code :)

sushantdhiman: Yeah idea of different level of logging is good, I'll make changes to code. I'll use `const` to reduce prefix use too.

sushantdhiman:
It will now look like something this

logs

We can view all logs via `sequelize*`, if we need pool only logs `sequelize:pool`

mickhansen:
Love it! Maybe `sequelize:sql` for the logging

lib/dialects/postgres/connection-manager.js
```diff
@@ -109,18 +110,21 @@ class ConnectionManager extends AbstractConnectionManager {
           return;
         }
         responded = true;
+        debug(`connection acquire`);
```
mickhansen: Maybe we should have a debug statement for the connected being setup aswell, so you can compare timestamps.
Then statements for either disconnection, error or successfully connected.

Perhaps include a unique id so it's comparable.
Not that you necessarily need to do this, it can be added later.

lib/dialects/sqlite/connection-manager.js
```diff
@@ -60,6 +62,7 @@ class ConnectionManager extends AbstractConnectionManager {
             if (err.code === 'SQLITE_CANTOPEN') return reject(new sequelizeErrors.ConnectionError(err));
             return reject(new sequelizeErrors.ConnectionError(err));
           }
+          debug(`connection acquire ${options.uuid}`);
```
mickhansen: `acquired` past-tense

lib/dialects/postgres/connection-manager.js
sushantdhiman: connection object don't have `UUID` when I try to access it. It always comes `undefined`. I was planning to add it but I don't know which parameter should I use other than `UUID`

sushantdhiman:
I have used `sequelize:sql:<dialect>` so cross dialect calls can be traced as well

sushantdhiman:
Any other changes @mickhansen

lib/dialects/postgres/connection-manager.js
mickhansen: ah right, only for transactions, which would confuse debugging

sushantdhiman: For `postgres` even during transactions we don't have `uuid` in connection object. But query have used `this.client` object whose access we don't have in `connection-manager.js`

sushantdhiman: similar thing happen in mysql no `uuid` when using transactions. Now trying pools, may be they have uuid for connection

sushantdhiman:
Ok I find out we assign `uuid` when setup transactions and unassign it at end of transaction. So when connection connects or disconnects we don't have any `uuid`, So it can't be logged when creating / destroying connections.

Although when a query is logged `uuid` is printed along with it to preserve the context.

mickhansen:
@sushantdhiman We'll have to live without identification for now since it will confuse the current way of identifying transactions

lib/dialects/postgres/connection-manager.js
```diff
@@ -109,18 +110,21 @@ class ConnectionManager extends AbstractConnectionManager {
           return;
         }
         responded = true;
+        debug(`connection acquired`);
         resolve(connection);
       });
 
       // If we didn't ever hear from the client.connect() callback the connection timeout, node-postgres does not treat this as an error since no active query was ever emitted
       connection.on('end', () => {
+        debug(`connection disconnect`);
```
mickhansen: `connection timeout`

lib/dialects/mssql/connection-manager.js
```diff
@@ -130,6 +133,7 @@ class ConnectionManager extends AbstractConnectionManager {
     return new Promise(resolve => {
       connection.on('end', resolve);
       connection.close();
+      debug(`connection disconnect`);
```
mickhansen: `connection closed`

sushantdhiman:
Done

mickhansen:
Thanks a ton @sushantdhiman

lib/hooks.js
```diff
@@ -178,6 +181,7 @@ var Hooks = {
       return this;
     }
 
+    Utils.debug(`removing hook ${hookType}`);
```
luin: Why we use `Utils.debug` instead of `debug` here?

sushantdhiman: Yes it should be `debug`

jontelm:
Is this the right command ?

DEBUG=sequelize node server.js

sushantdhiman:
Its `DEBUG=sequelize:*` , but only available in `v4`

PR #6135, opened 2016-06-19
Use independent copy of bluebird (closes #5974)

### LINKED ISSUE #5974: Don't monkeypatch global bluebird Promise
Can we avoid monkeypatching global bluebird Promise like this https://github.com/sequelize/sequelize/blob/master/lib/promise.js#L27 ?

Ie. by making it an optional (in some kind of init(...) function) or just use sequelize's own Promise object "inherited" from bluebird?

overlookmotel:
### Pull Request check-list

### Description of change

As discussed in #5974, this PR makes `Sequelize.Promise` an independent copy of the [bluebird](https://www.npmjs.com/package/bluebird) promise library.

The implications of this are:

* The shims applied to `Sequelize.Promise` for CLS support no longer affect the global instance of bluebird (returned from `require('bluebird')`)
* `Promise` used internally in Sequelize will not be affected by any patches / global options applied by another module to global `Bluebird`
* Promises returned from all Sequelize async methods are `instanceof Sequelize.Promise` but will not be `instanceof require('bluebird')`

It also opens the door for us to use `Promise.coroutine.addYieldHandler()` to allow yielding arrays of promises within coroutines ([co](https://www.npmjs.com/package/co)-style), but without causing malfunction in other code outside of Sequelize which also uses bluebird coroutines.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,10 +11,12 @@
 - [ADDED] `DEBUG` support [#2852](https://github.com/sequelize/sequelize/issues/2852)
 - [ADDED] Intensive connection logging [#851](https://github.com/sequelize/sequelize/issues/851)
 - [FIXED] Only `belongsTo` uses `as` to construct foreign key - revert of [#5957](https://github.com/sequelize/sequelize/pull/5957) introduced in 4.0.0-0
+- [CHANGED] `Sequelize.Promise` is now an independent copy of `bluebird` library [#5974](https://github.com/sequelize/sequelize/issues/5974)
 
 ## BC breaks:
 - Range type bounds now default to [postgres default](https://www.postgresql.org/docs/9.5/static/rangetypes.html#RANGETYPES-CONSTRUCT) `[)` (inclusive, exclusive), previously was `()` (exclusive, exclusive)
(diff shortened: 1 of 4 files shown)


changelog.md
```diff
@@ -10,9 +10,11 @@
 - [ADDED] `Model.count` now allow specifying column to count on, use `options.col` [#4442](https://github.com/sequelize/sequelize/issues/4442)
 - [ADDED] `DEBUG` support [#2852](https://github.com/sequelize/sequelize/issues/2852)
 - [ADDED] Intensive connection logging [#851](https://github.com/sequelize/sequelize/issues/851)
+- [CHANGED] `Sequelize.Promise` is now an independent copy of Sequelize library [#5974](https://github.com/sequelize/sequelize/issues/5974)
```
mickhansen: Of the sequelize library you say eh?

overlookmotel:
Error spotted by @mickhansen fixed and rebased on latest master.

janmeier:
Sweet :)

overlookmotel:
Thanks @janmeier!

PR #6144, opened 2016-06-20
Do not set limit if findOne() is run querying on a unique key

jharting:
### Pull Request check-list

### Description of change

I noticed that when I am running a query using findOne() with an include and querying on a unique key, sequelize produces a MySQL query with a nested subquery. The nested subquery is not there when I change the query to use the primary key in the where clause instead of the unique key.

I extended the if clause in the findOne() method to treat both these cases equally as I believe the extra nested subquery is not needed in the former case.

I am not sure how to write a test for this as this change does not have implications on the outside - only the query is simpler internally. Please advise.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1444,10 +1444,17 @@ class Model {
     options = Utils.cloneDeep(options);
 
     if (options.limit === undefined) {
-      const pkVal = options.where && options.where[this.primaryKeyAttribute];
 
-      // Don't add limit if querying directly on the pk
-      if (!options.where || !(Utils.isPrimitive(pkVal) || Buffer.isBuffer(pkVal))) {
+      var self = this;
+      // Don't add limit if querying directly on the pk or any other unique key
+      if (!options.where || !Object.keys(this.attributes).some(function (key) {
+        var attribute = self.rawAttributes[key];
+        if (attribute.unique === true || attribute.primaryKey === true) {
+          var value = options.where[attribute.fieldName];
+          return Utils.isPrimitive(value) || Buffer.isBuffer(value);
+        }
+        return false;
+      })) {
         options.limit = 1;
       }
     }


mickhansen:
Pretty sure we already had something for this.

lib/model.js
```diff
@@ -1454,10 +1454,17 @@ class Model {
...
 
     if (options.limit === undefined) {
-      const pkVal = options.where && options.where[this.primaryKeyAttribute];
 
-      // Don't add limit if querying directly on the pk
-      if (!options.where || !(Utils.isPrimitive(pkVal) || Buffer.isBuffer(pkVal))) {
+      var self = this;
+      // Don't add limit if querying directly on the pk or any other unique key
+      if (!options.where || !Object.keys(this.attributes).some(function (key) {
+        var attribute = self.attributes[key];
```
mickhansen: There's already a cache of primary keys on the model

lib/model.js
```diff
@@ -1454,10 +1454,17 @@ class Model {
...
+        if (attribute.unique || attribute.primaryKey) {
```
mickhansen: `unique` can be a composite unique, so you'll need to check that it's a boolean value

lib/model.js
```diff
@@ -1454,10 +1454,17 @@ class Model {
     options = Utils.cloneDeep(options);
 
     if (options.limit === undefined) {
-      const pkVal = options.where && options.where[this.primaryKeyAttribute];
 
-      // Don't add limit if querying directly on the pk
-      if (!options.where || !(Utils.isPrimitive(pkVal) || Buffer.isBuffer(pkVal))) {
+      var self = this;
+      // Don't add limit if querying directly on the pk or any other unique key
+      if (!options.where || !Object.keys(this.attributes).some(function (key) {
```
mickhansen: Use `rawAttributes` since `attributes` currently refer to SQL formatted attributes (which will be refactored before version 4)

lib/model.js
mickhansen: same goes for composite primary keys, having one of them would not be sufficient.

mickhansen:
Will need some testing that covers singular unique keys, composite unique keys, composite primary keys (singular primary keys should be covered already).

In your application code, if you're looking up by primary key you should really use `Model.findById`

jharting:
I updated the pull request to incorporate the feedback on `rawAttributes` and composite unique/primary key.

Any tips on how to write tests for this? Since the effect of the change is a more optimized SQL query (no extra nested SELECT query) I am struggling with writing tests for that as the change does not have any functional effect on the outside.

janmeier:
It would probably be easiest to write these as unit tests - https://github.com/sequelize/sequelize/blob/master/test/unit/sql/select.test.js

PR #6145, opened 2016-06-20
SomeUnique example duplicate removed

cristoforo:
### Description of change

The 'someUnique' code is duplicated, as the right block is after few lines.


janmeier:
Thanks :

PR #6146, opened 2016-06-21
Add missing option description for returning and silent

frogcjn:
Add missing option description for returning and silent


janmeier:
Thanks :)

PR #6155, opened 2016-06-21
Fix #2702, before/after Save hook

### LINKED ISSUE #2702: Add `(before/after)Save` hook
I'm using 2.0.0-rc3 and finding out that there are `beforeCreate` and `beforeUpdate` but no `beforeSave`.

The reason `beforeSave` is useful is that I want to hash users' password before saving. If there's no `beforeSave`, I have to write the logic of hashing password in two place(both `beforeCreate` and `beforeUpdate`), otherwise, I can just write:

```javascript
beforeSave: function(instance, options) {
  if (instance.isModified('password')) {
    return hash(instance.password).then(function(hashedPassword) {
      instance.password = hashedPassword;
    });
  }
}
```

btw, I also wonder whether there's a `isModified` method to check if a field has been modified(or created).

Related file:
https://github.com/sequelize/sequelize/blob/master/lib/instance.js#L581

Related question:
http://stackoverflow.com/questions/21411848/pre-save-hook-and-instance-methods-in-sequelize

sushantdhiman:
### Pull Request check-list

### Description of change

Closes #2702

Added `beforeSave` and `afterSave` hooks. These are independent hooks executed after their counterparts i.e. `before/after` `create/save` hooks, Order has been updated in docs.

Also now we can remove hooks by references

### DIFF
--- a/docs/docs/hooks.md
+++ b/docs/docs/hooks.md
@@ -21,19 +21,17 @@ For a full list of hooks, see [Hooks API](/api/hooks).
   beforeCreate(instance, options, fn)
   beforeDestroy(instance, options, fn)
   beforeUpdate(instance, options, fn)
-(5)
   beforeSave(instance, options, fn)
 (-)
   create
   destroy
   update
-(6)
+(5)
   afterCreate(instance, options, fn)
   afterDestroy(instance, options, fn)
   afterUpdate(instance, options, fn)
-(7)
   afterSave(instance, options, fn)
-(8)
+(6)
   afterBulkCreate(instances, options, fn)
   afterBulkDestroy(options, fn)
   afterBulkUpdate(options, fn)
--- a/docs/docs/hooks.md
+++ b/docs/docs/hooks.md
@@ -21,15 +21,19 @@ For a full list of hooks, see [Hooks API](/api/hooks).
   beforeCreate(instance, options, fn)
   beforeDestroy(instance, options, fn)
   beforeUpdate(instance, options, fn)
+(5)
+  beforeSave(instance, options, fn)
 (-)
   create
   destroy
   update
-(5)
+(6)
   afterCreate(instance, options, fn)
   afterDestroy(instance, options, fn)
   afterUpdate(instance, options, fn)
-(6)
+(7)
+  afterSave(instance, options, fn)
+(8)
   afterBulkCreate(instances, options, fn)
   afterBulkDestroy(options, fn)
   afterBulkUpdate(options, fn)
@@ -241,8 +245,8 @@ If you use `Model.bulkCreate(...)` with the `updatesOnDuplicate` option, changes
 
 ```
 // Bulk updating existing users with updatesOnDuplicate option
-Users.bulkCreate([{ id: 1, isMemeber: true}, 
-                 { id: 2, isMember: false}], 
+Users.bulkCreate([{ id: 1, isMemeber: true},
+                 { id: 2, isMember: false}],
                  { updatesOnDuplicate: ['isMember']})
 
 User.beforeBulkCreate(function (users, options) {
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -49,6 +49,8 @@ var hookTypes = {
   afterRestore: {params: 2},
   beforeUpdate: {params: 2},
   afterUpdate: {params: 2},
+  beforeSave: {params: 2},
+  afterSave: {params: 2},
   beforeBulkCreate: {params: 2},
   afterBulkCreate: {params: 2},
   beforeBulkDestroy: {params: 1},
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -5,30 +5,18 @@ const Utils = require('./utils')
   , debug = Utils.getLogger().debugContext('hooks');
 
 /**
- * Hooks are function that are called before and after  (bulk-) creation/updating/deletion and validation. Hooks can be added to you models in three ways:
+ * Hooks are function that are called before and after  (bulk-) creation/updating/deletion and validation.
+ * Hooks can be added to you models in three ways:
  *
- * 1. By specifying them as options in `sequelize.define`
- * 2. By calling `hook()` with a string and your hook handler function
- * 3. By calling the function with the same name as the hook you want
+ * 1. By calling `hook()` with a string and your hook handler function
+ * 2. By calling the function with the same name as the hook you want
 
  * ```js
- * // Method 1
- * sequelize.define(name, { attributes }, {
- *   hooks: {
- *     beforeBulkCreate: function () {
- *       // can be a single function
- *     },
- *     beforeValidate: [
- *       function () {},
- *       function() {} // Or an array of several
- *     ]
- *   }
- * })
  *
- * // Method 2
+ * // Method 1
  * Model.hook('afterDestroy', function () {})
  *
- * // Method 3
+ * // Method 2
  * Model.afterBulkUpdate(function () {})
  * ```
  *
@@ -83,6 +71,16 @@ var hookAliases = {
   beforeConnection: 'beforeConnect'
 };
 
+/**
+ * A hook which call other hook on being called
+ */
+var hookProxies = {
+  beforeUpdate: ['beforeSave'],
+  afterUpdate: ['afterSave'],
+  beforeCreate: ['beforeSave'],
+  afterCreate: ['afterSave']
+};
+
 var Hooks = {
   replaceHookAliases: function(hooks) {
     var realHookName;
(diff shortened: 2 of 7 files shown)


sushantdhiman:
@mickhansen @janmeier any improvements ?

mickhansen:
Could before/after save not simply proxy to before/after create/update?

sushantdhiman:
They can be but I like to have separate concerns here. It occurs to me that `save` hooks should be independent of others. Just like the upsert hook I am planning to add. They will not proxy update create as well.

mickhansen:
upsert is very different from create/update (since it doesn't actually use those specific calls)

I don't see a big reason to have `before/afterSave` except to be convenience wrappers for the other hooks.

sushantdhiman:
If we consider to proxy around `before/after create/update`, how you would suggest to do that in clean way ?

mickhansen:
I was just considering a simple proxy method, but it just occurred to be that obviously won't work with bot the `addhook` and the `define({}, {hooks: {}})`  syntax

sushantdhiman:
Thats what I thought too :), So its ok to keep separate hook then?

mickhansen:
Perhaps we should consider having only a single API for adding hooks.

sushantdhiman:
I would go with `{ hooks; {} }`, but I don't think it will be very well accepted. I think both case `Model.beforeUpdate()` and `hooks: { }` are good for their own usecases. Which one you prefer?

mickhansen:
I only ever use `Model.hook()`, but i have no idea what the community generally uses.
`Model.hook()` is generally preferred in plugins i believe.

sushantdhiman:
It would be a great irony if @janmeier only used `Model.<hook>()` type. We will have three opinions among ourselves :)

Still deciding upon what to support will take some time considering community review. Can't we merge it for now, change it to proxy type if hook api is changed later on ?

janmeier:
Why would it not work to keep both? `addHook` simply adds another function to `Model.options.hooks`

(Kinda biased since I'm working on the same codebase as @mickhansen), I use `.hook()` as well.

If we only get to keep one, I think the api with methods makes more sense, especially if you're adding or removing hooks after the model has been defined - much cleaner than modifying `options.hooks`.
But I think we should keep both :)

sushantdhiman:
Keeping all three methods `addHook`, `hook` and `hooks: {}` is good option, we should not break it.

But we can't add a proxy hook for all three, Like we want the `before/after Save` hook to proxy `before/after  create/update`. This PR do it as separate hook. But I cant think of a clean way to do proxy hook in current implementation.

I think we should add `before/after save` hook as independent hook, much clean code and no bc breaks.

janmeier:
Can't we simply define in runHooks that create and update should run save as well? We already have all hookTypes defined here, maybe add `beforeSave: { aliasOf: 'beforeCreate' }`

sushantdhiman:
`beforeSave` must run for both `beforeUpdate` and `beforeCreate`. Can `aliasOf` bind to two methods?

janmeier:
Its not implemented yet - But yes ;) - Could also be the other way around `beforeCreate:  {..., alias: 'beforeSave' }` (actually I think there is already some aliassing functionality in hooks, but in this case a hook would alias to two other

sushantdhiman:
Great hint @janmeier , Let me see what can be done here :)

mickhansen:
> Keeping all three methods addHook, hook and hooks: {} is good option, we should not break it.

I don't agree, having only one method makes the API and codebase simpler and allows users to quicker understand by their only being one way to do it (when there's multiple ways of doing something you force the user to consider the differences).

sushantdhiman:
Should I kill `{ hooks: {} }` and `Model.<hookName>()`, Keep the `Model.hook('name' , fn)` only ?

mickhansen:
I think we should definitely kill `{ hooks: {} }` as for `<hook>()` vs `hook(hook)` i know that most plugins use the `<hook>()` style so that should likely be retained.

We can consider keeping both

mickhansen:
I realize i used incorrect formatting in previous comments, meant to say that i only ever use `Model.<hook>()`, not `Model.hook(name)`

sushantdhiman:
I am dropping `{ hooks: { } }` as discussed.

sushantdhiman:
What about `before/after Save`, It should be proxy or independent hook ?

sushantdhiman:
Should we consider `Event Emitter` API for hooks ? I think it will make more sense here ?

mickhansen:
@sushantdhiman Not sure what you mean? For emitting the events? The user would not be able to block/throw with that design.

I still vote for `beforeSave` as proxy

sushantdhiman:
Yeah, they can't be blocked without much effort :), Going with `proxy` around `create / update` then.

sushantdhiman:
@mickhansen @janmeier I have used a proxy implementation for hooks now. Let me know if there are any suggestions.

docs/docs/hooks.md
```diff
@@ -21,15 +21,19 @@ For a full list of hooks, see [Hooks API](/api/hooks).
   beforeCreate(instance, options, fn)
   beforeDestroy(instance, options, fn)
   beforeUpdate(instance, options, fn)
+(5)
+  beforeSave(instance, options, fn)
```
mickhansen: Wouldn't it run at the same time as beforeCreate/beforeUpdate? Not after

lib/hooks.js
```diff
@@ -119,8 +130,15 @@ var Hooks = {
     if (hookTypes[hookType] && hookTypes[hookType].sync) {
       hooks.forEach(function(hook) {
```
mickhansen: Wouldn't it be simpler to just append to the `hooks` array if for the proxy? Rather than add new logic

docs/docs/hooks.md
sushantdhiman: No it will run after their counterparts

mickhansen: It's not exactly a proxy then.

lib/hooks.js
sushantdhiman: It will break the order of execution. Proxied hooks must be called after primary hooks has been completed. Otherwise `beforeUpdate` errors will not stop the `beforeSave` hook

docs/docs/hooks.md
sushantdhiman: What will happen if there is error in `beforeUpdate` should `beforeSave` run in such cases ?

mickhansen: It should run with whatever logic hook normally does.
Which should happen automatically if we simply append the hook to the proxy targets when a hook with proxy targets is defined.

mickhansen:
This is not exactly a proxy implementation currently.
I would prefer that beforeSave defines that it proxies to beforeCreate/beforeUpdate and that you check this definition when the hook is added, and them simply add the hook to the other types.

lib/hooks.js
mickhansen: They should operate as all other hooks, see my comment in the main thread.

sushantdhiman:
I understand its not a exact proxy implementation, but what will happen if

- `beforeCreate` fails should `beforeSave` run
- `beforeCreate` is removed, should `beforeSave` be removed
- How those will work if they are just proxies over `create / update`

mickhansen:
@sushantdhiman It should follow the same logic as having 2 x beforeCreate. Definition order rules.

mickhansen:
As for hook removal, it looks like that only works if you're using named hooks? (Although it should really be refactored to work on fn reference)

sushantdhiman:
Hi @mickhansen @janmeier now we have real proxies and hooks can be removed by reference as well. Any changes ?

mickhansen:
LGTM, /cc @janmeier

janmeier:
Sweet!

sushantdhiman:
Thanks, I had dreams about hooks last night :laughing:

glebec:
@sushantdhiman & @mickhansen I am confused about Sequelize's release system. I see that this PR was merged on June 23 (same day as v3.23.4) and there have been two more GitHub / npm releases since then (3.23.5 and 3.23.6). Also, `beforeSave` is now [included in the main docs site](http://docs.sequelizejs.com/en/latest/docs/hooks/).

Yet despite all that, `beforeSave` is not actually included in the code for 3.23.6, the most recent release as of this comment. Verified by both downloading the source zip & checking an actual npm installation. When will `beforeSave` make release?

sushantdhiman:
@glebec currently `master` points to `4.x` (upcoming), similarly branch `v3` point to `3.x`(stable) . So any changes merged to master are going to be released in `4.x`.

If you need any feature ported from  `4.x` to `3.x`, you can copy code from related PR and open a PR against `v3` branch. We will release new `3.x` version with that feature

glebec:
Aaaah, that clears it up, thank you.

My only issue is that I teach students Sequelize, and they are finding `beforeSave` in the docs and confused when it silently fails (especially problematic in assessments). Is there any reason 3.x cannot throw an error if attempting to register with a nonexistent hook name?

sushantdhiman:
@glebec because docs also points to `latest` ( or `master`) by default. You can always choose the `stable` option from bottom menu.

**Image**

screenshot-docs sequelizejs com 2016-08-08 10-53-41

mickhansen:
@janmeier any way to make stable the default while we finalize 4.0?

janmeier:
Done :)

felixfbecker:
For the API, please always prefer `<hook>()` over `.hook(name)`, because you cannot get IntelliSense for string values.

PR #6168, opened 2016-06-23
Only add ORDER BY because of the OFFSET NEXT clause when no "order" i…

### LINKED ISSUE #6162: [MSSQL] Big problem sorting by primaryKey (duplicate order by column error)
## What you are doing?

I am using the "mssql" dialect and trying to make a sorted findAndCountAll query with includes that uses the primary key as sorting.

Below i give you an example and i have debugged the code and the problem comes from the file located at lib/dialects/mssql/query-generator.js in the function addLimitAndOffset.

This condition is not ok because when you try to sort by the primaryKey this fragment is also added and then SQL Server gives a "duplicated column" error:
```js
if (!options.order || (options.include && !orders.subQueryOrder.length)) {
        fragment += (options.order && !isSubQuery) ? ', ' : ' ORDER BY ';
        fragment += this.quoteIdentifier(model.primaryKeyField);
      }
```

I am using Sequelize 3 so please fix this in Sequelize 3 as well , not only in Sequelize 4.

```js

var OrderLog = sequelize.define('OrderLog',{
        id: {type : Sequelize.INTEGER(10).UNSIGNED, primaryKey:true, autoIncrement:true}
        , userid: { type: Sequelize.INTEGER(11)  }
        , state: {type: Sequelize.INTEGER(10).UNSIGNED }
        , orderid: {type: Sequelize.INTEGER(10).UNSIGNED,allowNull:false }
        , logdate: {type: Sequelize.DATE }
        , info: {type: Sequelize.TEXT }
        , visibility_level: {type: Sequelize.INTEGER(11),allowNull:false }
    },{
        tableName: 'jos_vxc_orderlog', timestamps: false
    });

var OrderState = sequelize.define('OrderState',{
        id: {type : Sequelize.INTEGER(10).UNSIGNED, primaryKey:true, autoIncrement:true}
        , nameid: {type: Sequelize.STRING(255) , allowNull:false }
        , name: {type: Sequelize.STRING(255) , allowNull:false }
    },{
        tableName : 'jos_vxc_orderstate', timestamps: false
    });

               OrderLog.findAndCountAll({
                    limit: 8, // You can put whatever number but don't leave undefined
                    offset: 0, // You can put whatever number but don't leave undefined
                    include: [{model: OrderState}], // Include at least one include
                    order:  [["id", "desc"]]  // Sort by primary key
                }).then(function(data) {
                    callback({result: "ok", data: {orderlogs: data.rows, numPages: (pageSize != -1)? Math.ceil(data.count / pageSize): 1}});
                }).catch(function(error) {
                    callback({result: "error", error: error.message});
                })

```

I wanted the function to work properly and give me results sorted as expected.

SQL Server gives the following error "A column has been specificied more than once in the order by list. Columns in the order by list must be unique.

__Dialect:__ mssql
__Database version: SQL Server 2014
__Sequelize version:__ 3.23.3

juansev:
### Pull Request check-list

- [ ] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)? NOT TESTED
- [X] Does your issue contain a link to existing issue (Closes #6162) or a description of the issue you are solving?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? No API modification

### Description of change

This modification is needed to only include the extra ORDER BY when no other sorting is defined. Before could collide when sorting by the primary Key and it gives duplicated column error.

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -685,9 +685,8 @@ var QueryGenerator = {
     }
 
     if (options.limit || options.offset) {
-      if (!options.order || (options.include && !orders.subQueryOrder.length)) {
-        fragment += (options.order && !isSubQuery) ? ', ' : ' ORDER BY ';
-        fragment += this.quoteIdentifier(model.primaryKeyField);
+      if (!options.order || (isSubQuery && !orders.subQueryOrder.length)) {
+        fragment += ' ORDER BY ' + this.quoteIdentifier(model.primaryKeyField);
       }
 
       if (options.offset || options.limit) {


mickhansen:
A test would be great!

juansev:
I have tested the code in run-time and it works ok but i haven't written any test-case of this functionality.  

I am not used to use the toolchain. I tried to run npm test-mssql and npm just throw an error because it doesn't recognize DIALECT. I am developing in a Windows environment. Are we sure that the tests pass now? Because i have checked the node-unit tests for the mssql dialect and they don't show any OFFSET NEXT clause there. 

I would write an small test to check that the old behaviour is not recovered (by sorting with the primary key with limits) but i would need an small guide on how to proper develop sequelize as i am only a user not a main developer of sequelize.

juansev:
Also, what is the proper way to include this modification in the master branch without merging? What is your standard procedure to fix bugs in old branches?

janmeier:
I think the easiest way to get a development enviroment up and running would be with docker. `docker-compose run sequelize /bin/sh -c "npm run test-integration-mssql"` should do the trick.

These are the unit tests for that specific function - You could add one which doesn't have an order sort, where it should be added, and one with order: id, to check that it isnt added twice https://github.com/sequelize/sequelize/blob/master/test/unit/sql/offset-limit.test.js

janmeier:
We don't have any strict procedure for making sure commits end up in both branches (sadly) - The easiest is for you to provide a separate PR to each branch

juansev:
Thank you very much janmeier for the information. I will try this weekend to run the tests and implement the new ones.

juansev:
I have been having problems running the tests ( i created an Ubuntu 16.04 VM and installed Docker to double-check that the error wasn't coming from Windows).

- First of all, when i try to run the command "npm run test-integration-mssql" it complains because mocha is not installed.
- Second, if i open a command-line into the container and install mocha manually and run the command, i get the following error:

npm ERR! Linux 4.4.0-21-generic
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "run" "test-integration"
npm ERR! node v4.4.6
npm ERR! npm  v2.15.5
npm ERR! code ELIFECYCLE
npm ERR! sequelize@3.23.4 test-integration: `mocha --globals setImmediate,clearImmediate --ui tdd --check-leaks --colors -t 15000 --reporter spec --grep "$GREP" 'test/integration/**/*.test.js'`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the sequelize@3.23.4 test-integration script 'mocha --globals setImmediate,clearImmediate --ui tdd --check-leaks --colors -t 15000 --reporter spec --grep "$GREP" 'test/integration/**/*.test.js''.
npm ERR! This is most likely a problem with the sequelize package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     mocha --globals setImmediate,clearImmediate --ui tdd --check-leaks --colors -t 15000 --reporter spec --grep "$GREP" 'test/integration/**/*.test.js'
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs sequelize
npm ERR! Or if that isn't available, you can get their info via:
npm ERR! 
npm ERR!     npm owner ls sequelize
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /sequelize/npm-debug.log

juansev:
BTW the command "npm run test-integration-mysql" runs just fine. The problem only appears with "npm run test-integration-mssql". So it seems that the tests themselves are not working for mssql or i am missing to install another dependency besides mocha .

janmeier:
Hmm, thats weird - there shouldn't be any extra dependencies for mssql - and everything should be installed by `npm install`

In any case, I guess you'll have to let travis run your tests.

BridgeAR:
@janmeier mssql does not run in my docker container either

sushantdhiman:
@juansev Thank you for this contribution. Issue you are facing is probably fixed by https://github.com/sequelize/sequelize/pull/6181. If that is not the case please do let me know.

PR #6170, opened 2016-06-24
Fixed behavior where a mixed required status in an include would not honor given required's

### LINKED ISSUE #5999: Grandchild include `required` option overwriting child include `required`
## What you are doing?
(see the full example here  https://github.com/cbauerme/sequelize-isolate-include-example)

I have created the following models and assocations
``` js
var User = sequelize.define('user', {
    name: { type: Sequelize.STRING, allowNull: false },
});

var Group = sequelize.define('group', {
    groupName: { type: Sequelize.STRING, allowNull: false },
});

var Budget = sequelize.define('budget', {
    balance: { type: Sequelize.FLOAT, allowNull: false },
});

User.belongsTo(Group, { as: 'group' });
Budget.belongsTo(Group, { as: 'group' });
Group.hasMany(User, { as: 'users' });
Group.hasOne(Budget, { as: 'budget' });
```
And filled them with 

```js
Group.create({
    groupName: 'Group A',
     users: [
         { name: 'Alice' },
         { name: 'Bob' }
    ],
    budget: {
        balance: 34.58
    }
});
Group.create({
    groupName: 'Group B',
    users: [
        { name: 'Hannah' },
        { name: 'Ed' }
    ],
});
```

Finally I have an unique report to write. I want to select all users. I also want the group only for the cases where the group has a budget.

```js
return User.findAll({
        include: [
            {
                model: Group,
                as: 'group',
                required: false,
                include: [
                    {
                        model: Budget,
                        as: 'budget',
                        required: true,
                    }
                ]
            }
        ]
}).then(function (users) {
    users.forEach(function (user) {
        console.log("User: " + user.name);
        if (user.group !== undefined) {
            console.log("   Group: " + user.group.groupName);
            console.log("   Budget: " + user.group.budget.balance);
        }
    });
});
```

I want the following output (in no particular order)
```
User: Alice
   Group: Group A
   Budget: 34.58
User: Bob
   Group: Group A
   Budget: 34.58
User: Hannah
User: Ed
```

```
User: Alice
   Group: Group A
   Budget: 34.58
User: Bob
   Group: Group A
   Budget: 34.58
```

__Dialect:__ sqlite (any?)
__Sequelize version:__ 3.4

cbauerme:
### Pull Request check-list

### Description of change

Changes to fix #5999.

The fundamental purpose of this change is to insulate higher includes from deeper ones when the deeper include is required and the parent or ancestors are not. To that end when there is an include required mismatch the SQL generated is changed slightly.

```sql
LEFT OUTER JOIN tableA ON tableA.parentId = parent.id INNER JOIN tableB ON tableB.tableAId = TableA.Id
```
becomes
```sql
LEFT OUTER JOIN ( tableA  INNER JOIN tableB ON tableB.tableAId = TableA.Id) ON tableA.parentId = parent.id
```

Other changes were necessary to make this possible. The most major is a change to the alias of associated models being included. Before they were aliased to a combination of the parent and itself with a period, like `tableA.tableB`. However, when surrounded by parentheses these aliases would not be found by a select. To solve this there is now an internal and external alias. An external alias is the same as before, and assigned by column as before. The internal alias replaces the period with an underscore and is used.

The abstract query-generator function `selectQuery` has been refactored. The section handling includes has been split into a separate function. This function calls to separate methods that generate either a standard join or a through join. These are returned in parts, and if the parent notices a required mismatch it is able to rearrange and add parenthesis to protect the result.

There are additional changes to a test here or there to accommodate the major changes to `selectQuery`.


lib/dialects/abstract/query-generator.js
```diff
@@ -774,7 +774,7 @@ const QueryGenerator = {
     Currently this function is only used for ordering / grouping columns and Sequelize.col(), but it could
     potentially also be used for other places where we want to be able to call SQL functions (e.g. as default values)
   */
-  quote(obj, parent, force) {
+  quote(obj, parent, force, internal) {
```
mickhansen: perhaps take a connector argument instead?

lib/dialects/abstract/query-generator.js
```diff
@@ -1493,124 +1182,273 @@ const QueryGenerator = {
...
+        attr = mainTableAs + '.' + attr;
+      }
+
+      return attr;
+
+    });
+
+  },
+
+  generateInclude(include, parentTableName, topLevelOptions, topLevelTableInfo, topLevelTableAttributes, topLevelSubQuery) {
```
mickhansen: perhaps all `topLevel*` could be grouped into a `topLevel` object, if it makes sense.

lib/dialects/abstract/query-generator.js
```diff
@@ -1493,124 +1182,273 @@ const QueryGenerator = {
...
+        }
+        return `${prefix} AS ${this.quoteIdentifier(`${includeAs.externalAs}.${attrAs}`, true)}`;
+      });
+      if (include.subQuery && topLevelSubQuery) {
+        for (let attr of attributes) {
+          topLevelTableAttributes.subQuery.push(attr);
+        }
+      } else {
+        for (let attr of attributes) {
+          topLevelTableAttributes.main.push(attr);
```
mickhansen: consider returning the attributes to the parent, make the code more functional/pure and easier to follow.

mickhansen:
Looks like great work @cbauerme - code was in dire need of a refactor.
Will take a bit to review but i promise we'll take better care of this than the CTE PR

cbauerme:
Thanks! I went ahead and changed the places commented on.

test/integration/include.test.js
```diff
@@ -858,4 +858,112 @@ describe(Support.getTestDialectTeaser('Include'), function() {
...
+        ]);
+      });
+    });
+
+  };
+
+  describe('nested includes', function () {
+
+    beforeEach(function () {
+      return createTeamsUsersAndClearences.bind(this)();
```
mickhansen: Any reason this cant simply be the code?

test/unit/sql/select.test.js
```diff
@@ -192,12 +192,12 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
           ]
         }
       }, {
-        default: 'SELECT [user].*, [POSTS].[id] AS [POSTS.id], [POSTS].[title] AS [POSTS.title], [POSTS.COMMENTS].[id] AS [POSTS.COMMENTS.id], [POSTS.COMMENTS].[title] AS [POSTS.COMMENTS.title] FROM ('+
+        default: 'SELECT [user].*, [POSTS].[id] AS [POSTS.id], [POSTS].[title] AS [POSTS.title], [POSTS_COMMENTS].[id] AS [POSTS.COMMENTS.id], [POSTS_COMMENTS].[title] AS [POSTS.COMMENTS.title] FROM ('+
```
mickhansen: Why is this changed?

test/unit/sql/generateJoin.test.js
```diff
@@ -0,0 +1,308 @@
...
+        ]
+      },
+      {
+        default: "LEFT OUTER JOIN [company] AS [Company] ON [User].[company_id] = [Company].[id]"
+      }
+    );
+
+
+    testsql(
+      "include[0]",
```
mickhansen: before the test description has some info on the parameters, maybe do something like `'include[0](where, or)`

lib/dialects/abstract/query-generator.js
```diff
@@ -1493,124 +1196,298 @@ const QueryGenerator = {
...
+      }
+
+      joinQueries.subQuery.push(subChildIncludes.join(''));
+
+    }
+
+    return {
+      mainQuery: joinQueries.mainQuery.join(''),
+      subQuery: joinQueries.subQuery.join(''),
+      newAttributes: newAttributes,
```
mickhansen: Could this be simply `attributes`?

lib/dialects/abstract/query-generator.js
```diff
@@ -1493,124 +1196,298 @@ const QueryGenerator = {
...
         }
 
-        mainQueryOrder.push(this.quote(t, model));
+        let prefix;
+        if (verbatim === true) {
+          prefix = attr;
+        } else {
+          prefix = `${this.quoteIdentifier(includeAs.internalAs)}.${this.quoteIdentifier(attr)}`;
+        }
+        return `${prefix} AS ${this.quoteIdentifier(`${includeAs.externalAs}.${attrAs}`, true)}`;
```
mickhansen: is part of this not handeable by your new escapeAttributes method?

mickhansen:
> The most major is a change to the alias of associated models being included. Before they were aliased to a combination of the parent and itself with a period, like tableA.tableB. However, when surrounded by parentheses these aliases would not be found by a select. To solve this there is now an internal and external alias.

Can you show an SQL example for this please @cbauerme

cbauerme:
@mickhansen  absolutely!

```sql
SELECT `Task`.`id`
	,`Task`.`title`
	,`Task`.`createdAt`
	,`Task`.`updatedAt`
	,`Task`.`ProjectId`
	,`Project`.`id` AS `Project.id`
	,`Project`.`title` AS `Project.title`
	,`Project`.`createdAt` AS `Project.createdAt`
	,`Project`.`updatedAt` AS `Project.updatedAt`
	,`Project`.`UserId` AS `Project.UserId`
	,`Project.User`.`id` AS `Project.User.id`
	,`Project.User`.`username` AS `Project.User.username`
	,`Project.User`.`createdAt` AS `Project.User.createdAt`
	,`Project.User`.`updatedAt` AS `Project.User.updatedAt`
FROM `Tasks` AS `Task`
LEFT OUTER JOIN `Projects` AS `Project` ON `Task`.`ProjectId` = `Project`.`id`
INNER JOIN `Users` AS `Project.User` ON `Project`.`UserId` = `Project.User`.`id`
	AND `Project.User`.`username` = 'leia';
```

was before the change, and worked perfectly fine! However, when I added the parentheses we got

```sql
SELECT `Task`.`id`
	,`Task`.`title`
	,`Task`.`createdAt`
	,`Task`.`updatedAt`
	,`Task`.`ProjectId`
	,`Project`.`id` AS `Project.id`
	,`Project`.`title` AS `Project.title`
	,`Project`.`createdAt` AS `Project.createdAt`
	,`Project`.`updatedAt` AS `Project.updatedAt`
	,`Project`.`UserId` AS `Project.UserId`
	,`Project.User`.`id` AS `Project.User.id`
	,`Project.User`.`username` AS `Project.User.username`
	,`Project.User`.`createdAt` AS `Project.User.createdAt`
	,`Project.User`.`updatedAt` AS `Project.User.updatedAt`
FROM `Tasks` AS `Task`
LEFT OUTER JOIN (
	`Projects` AS `Project` INNER JOIN `Users` AS `Project.User` ON `Project`.`UserId` = `Project.User`.`id`
		AND `Project.User`.`username` = 'leia'
	) ON `Task`.`ProjectId` = `Project`.`id`;
```

This would fail with (roughly) ````Project.User`.`id` cannot be found```

I added the underscores to produce

```sql
SELECT `Task`.`id`
	,`Task`.`title`
	,`Task`.`createdAt`
	,`Task`.`updatedAt`
	,`Task`.`ProjectId`
	,`Project`.`id` AS `Project.id`
	,`Project`.`title` AS `Project.title`
	,`Project`.`createdAt` AS `Project.createdAt`
	,`Project`.`updatedAt` AS `Project.updatedAt`
	,`Project`.`UserId` AS `Project.UserId`
	,`Project_User`.`id` AS `Project.User.id`
	,`Project_User`.`username` AS `Project.User.username`
	,`Project_User`.`createdAt` AS `Project.User.createdAt`
	,`Project_User`.`updatedAt` AS `Project.User.updatedAt`
FROM `Tasks` AS `Task`
LEFT OUTER JOIN (
	`Projects` AS `Project` INNER JOIN `Users` AS `Project_User` ON `Project`.`UserId` = `Project_User`.`id`
		AND `Project_User`.`username` = 'leia'
	) ON `Task`.`ProjectId` = `Project`.`id`;
```
which works without issue.

mickhansen:
I don't really see why the switch `.` from `_` should make any difference
Is the error from the join statement or from the SELECT?

The SELECT is also using `_`, should that not still use `.?  Why does this not break deduplication?

cbauerme:
I completely agree that it shouldn't make a difference, but that was what I saw. I just reran some of my tests to get the specific error from the failing example, which was

```sql 
no such column: Project.User.id: SELECT `Task`.`id`
```

the error is definitely with the select. If you remove the last four rows from the failing select everything works fine (minus the four rows of data we wanted).

I ended up asking this question on [stack overflow](http://stackoverflow.com/questions/37884525/sqlite-no-such-column-from-right-side-of-nested-inner-join/37884873#37884873) and one of the responses I got speculated that SQL might treat a dot differently, for whatever reason.

mickhansen:
Hmm, that's highly odd and introduces some inconsistency for query parsing that we definitely always want to be the same.

Using `.` should not be an issue as long as it's properly encased. 

Must be related to how the isolated join behaves.

Does this fail for all dialects or only mysql.

cbauerme:
If you want to see it for yourself here is the db I tested with

```sql
BEGIN TRANSACTION;
CREATE TABLE `User` (
	`id`	INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
	`userName`	TEXT
);
INSERT INTO `User` VALUES (1,'test01');
INSERT INTO `User` VALUES (2,'test02');
CREATE TABLE `Task` (
	`id`	INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
	`title`	TEXT,
	`projectId`	INTEGER
);
INSERT INTO `Task` VALUES (1,'Task A',1);
INSERT INTO `Task` VALUES (2,'Task B',1);
INSERT INTO `Task` VALUES (3,'Task C',2);
CREATE TABLE `Project` (
	`id`	INTEGER NOT NULL UNIQUE,
	`title`	TEXT,
	`userId`	INTEGER,
	PRIMARY KEY(id)
);
INSERT INTO `Project` VALUES (1,'Project 01',1);
INSERT INTO `Project` VALUES (2,'Project 02',NULL);
COMMIT;
```

This should work (normal sequelize select)
```sql
SELECT `Task`.`id`
	,`Task`.`title`
	,`Task`.`projectId`
	,`Project`.`id` AS `Project.id`
	,`Project`.`title` AS `Project.title`
	,`Project`.`userId` AS `Project.userId`
	,`Project.User`.`id` AS `Project.user.id`
	,`Project.User`.`username` AS `Project.User.username`
FROM `Task` AS `Task` 
LEFT OUTER JOIN  `Project` AS `Project` ON `Task`.`projectId` = `project`.`id`
INNER JOIN `User` AS `Project.User` ON `Project`.`userId` = `Project.User`.`id`
AND `Project.User`.`username` = 'test01';
```

this should not (my intermediate solution)
```sql
SELECT `Task`.`id`
	,`Task`.`title`
	,`Task`.`projectId`
	,`Project`.`id` AS `Project.id`
	,`Project`.`title` AS `Project.title`
	,`Project`.`userId` AS `Project.UserId`
	,`Project.User`.`id` AS `Project.User.id`
	,`Project.User`.`username` AS `Project.User.username`
FROM `Task` AS `Task`
LEFT OUTER JOIN  (`Project` AS `Project` 
INNER JOIN `User` AS `Project.User` ON `Project`.`userId` = `Project.User`.`id`
	AND `Project.User`.`username` = 'test01' ) ON `Task`.`projectId` = `project`.`id`;
``` 

and then the final statement should work again
```sql
SELECT `Task`.`id`
	,`Task`.`title`
	,`Task`.`projectId`
	,`Project`.`id` AS `Project.id`
	,`Project`.`title` AS `Project.title`
	,`Project`.`userId` AS `Project.UserId`
	,`Project_User`.`id` AS `Project.User.id`
	,`Project_User`.`username` AS `Project.User.username`
FROM `Task` AS `Task`
LEFT OUTER JOIN  (`Project` AS `Project` 
INNER JOIN `User` AS `Project_User` ON `Project`.`userId` = `Project_User`.`id`
	AND `Project_User`.`username` = 'test01' ) ON `Task`.`projectId` = `project`.`id`;
```

cbauerme:
I have confirmed failure with SQLite, and I will go ahead and check the other dialects I can. 

I agree that it is not a great solution, but I just couldn't get the dots to work. I will spend so more time with it and see what I can come up with.

mickhansen:
@cbauerme Oh right, SQLite never worked for me, we don't support paranthesis for N:M joins for sqlite either (look at the dependent join conditional).

cbauerme:
Ah, jeez. Looks like this is the same issue that makes dependent join necessary, every other dialect seems to be fine. 

So, the problem I see here is that there is not necessarily a good way to get around this specific issue like there is with the join dependent solution for through joins and SQLite. 

I can either go back to the drawing board for this fix or try to defend my underscores. Which would you prefer?

mickhansen:
@cbauerme Problem with underscores is that's something a personally would reasonably use in a table name.

We need a delimiter we can count on - I don't like sqlite having to drop too far behind like this.

mickhansen:
Although, something i had considered anyways, was generate column mapping before the query is sent, in that case we could easily take into account non delimiting `_` for mapping back - but it's a lot more work.

cbauerme:
Well, I am already part way there. `selectQuery` now maintains two names for each table, the one it will use "internally", for aliasing each table after the `FROM` and an external name, the one it returns the columns as. I could go through and improve the way internal names are created so that it would be guaranteed to never conflict. 

And, if that worked, join table dependent could porbably be removed as well.

mickhansen:
@cbauerme Right, but we want this to work with the deduplication/instance building code aswell.
And as part of that i'd like to have that code work with generic raw query results aswell.

I.e. have the user, or the code itself, provide a mapping for deduplication of some sorts.

```
{
  'User': User,
  'User_Project': User.Project or 'user.projects[]`
}
```

mickhansen:
Ah right, you're still aliasing the columns in the final select so they'll come out delimited.
Hmm, that could likely work just fine then - But i do feel it could be simplified.

cbauerme:
I certainly don't like that this is special solution for select, but I just don't have any additional good ideas that don't involve a lot of work outside of what I have currently changed, which is already pretty big.

mickhansen:
@cbauerme this breaks people targeting tables with `$$` right?

cbauerme:
@mickhansen in the case of using bind parameters with a `query()` call? Or somewhere else?

mickhansen:
@cbauerme No also the where syntax for targeting cols. `where: {'$nested.table.field$': {$col: 'thisTable.field'}}

cbauerme:
@mickhansen ah, yes that would break pretty badly. I am trying to see what we can do with sqlite to keep the names as they are, since that would make everything significantly less complicated.

cbauerme:
I have managed to confirm that the problem with a period separator inside parentheses is fundamentally unworkable with sqlite. The error comes from a function that compares the name that is being looked for against a "span" that has the form `dbName.tableName.columnName`. Each part is created by advancing down the span until the next period is found,  which means the composite names combined with a period are never actually checked in their entirety. 

So, I have a couple ideas for going forward that I could use feedback on

**1. Rework the way a separated name works so that the period problem doesn't come up.** 
My current code uses a naive version of this, but it seems like it could be enhanced to the point where is would not be likely to collide with other table names. However, this might cause issues with targeting columns as seen above.

**2.  Change the logic of the SQL fix**
Currently the SQL is rearranged when a mismatch is detected 
`A LEFT OUTER JOIN B INNER JOIN C` becomes `A LEFT OUTER JOIN (B INNER JOIN C)`
However, it may be possible to recreated something close to the above with a different combination of joins that do not involve parentheses. An example would be `TableA LEFT OUTER JOIN TableB ON TableA.d = TableB.aId AND TableB.cId IS NOT NULL LEFT OUTER JOIN TableC ...` However this doesn't really come close to replicating the first SQL I cam up with and it seems like certain combinations could break it as well.

This are the best ideas I have had, and I will keep working on them to see if I can iron out the problems. Please let me know if anyone has suggestions to fix them or even better ideas!

cbauerme:
After some more thought and experimenting I can say **2** is dead end. I think the best way forward is to work around the unfortunate problem with sqlite however we can, but keep most of what I have done already. I think replacing the connector of nested tables with `->` may work. Cases like `User->Group->Task.name` look OK, and we should be pretty safe from similar table names.

mickhansen:
Hmm, that's a major change, but one we might have to consider if we want SQLite to be supported in this aswell (already left it behind on join tables which i wasn't too happy about).

/cc @janmeier

cbauerme:
@mickhansen I am going to go ahead and revise my change so that it uses `->` internally and returns each in the expected `.` form. I will figure out how to convert any targeted columns and add some tests for them as well. I think the overall issue of `->` goes outside the scope of this change, and would prefer to get this one in and address the bigger issue separately, if that is OK.

cbauerme:
@mickhansen OK, I went ahead and handled the `$col` and `$table.association$` cases. When the `whereItemQuery` function puts them back together I sneak in the `->`

I think at this point we have a good-enough solution for solving the described bug, but not a particularly elegant one overall, because of the internal translation from `.` to `->`. In my opinion it would probably be better to get this in while I am still able to keep up with other changes, then worry about changing the overall approach to associated table aliases.

mickhansen:
I might be able to live with this. It's a major change so i'd like a second pair of eyes.

/cc @sushantdhiman @felixfbecker @janmeier

lib/dialects/abstract/query-generator.js
```diff
@@ -774,7 +774,10 @@ const QueryGenerator = {
     Currently this function is only used for ordering / grouping columns and Sequelize.col(), but it could
     potentially also be used for other places where we want to be able to call SQL functions (e.g. as default values)
   */
-  quote(obj, parent, force) {
+  quote(obj, parent, force, connector) {
+
+    connector = connector === undefined ? '.' : connector;
```
sushantdhiman: we can do  ```connector ? '.' : connector``` for compactness

sushantdhiman:
@cbauerme Its possible to get rid of the`->` completely. I have the fixed code but not sure how to get it to you

sushantdhiman:
**Before Fix** (Query with `->`)

```sql
SELECT "City"."id",
       "City"."name",
       "Houses"."id" AS "Houses.id",
       "Houses"."name" AS "Houses.name",
       "Houses"."CityId" AS "Houses.CityId",
       "Houses->Tenants"."id" AS "Houses.Tenants.id",
       "Houses->Tenants"."name" AS "Houses.Tenants.name",
       "Houses->Tenants"."HouseId" AS "Houses.Tenants.HouseId",
       "Houses->Tenants->Items"."id" AS "Houses.Tenants.Items.id",
       "Houses->Tenants->Items"."name" AS "Houses.Tenants.Items.name",
       "Houses->Tenants->Items"."TenantId" AS "Houses.Tenants.Items.TenantId"
FROM "City" AS "City"
INNER JOIN "House" AS "Houses" ON "City"."id" = "Houses"."CityId"
LEFT OUTER JOIN "Tenant" AS "Houses->Tenants" ON "Houses"."id" = "Houses->Tenants"."HouseId"
LEFT OUTER JOIN "Item" AS "Houses->Tenants->Items" ON "Houses->Tenants"."id" = "Houses->Tenants->Items"."TenantId";
```

**After Fix** (Query with `.` only)

```sql
SELECT "City"."id",
       "City"."name",
       "Houses"."id" AS "Houses.id",
       "Houses"."name" AS "Houses.name",
       "Houses"."CityId" AS "Houses.CityId",
       "Houses.Tenants"."id" AS "Houses.Tenants.id",
       "Houses.Tenants"."name" AS "Houses.Tenants.name",
       "Houses.Tenants"."HouseId" AS "Houses.Tenants.HouseId",
       "Houses.Tenants.Items"."id" AS "Houses.Tenants.Items.id",
       "Houses.Tenants.Items"."name" AS "Houses.Tenants.Items.name",
       "Houses.Tenants.Items"."TenantId" AS "Houses.Tenants.Items.TenantId"
FROM "City" AS "City"
INNER JOIN "House" AS "Houses" ON "City"."id" = "Houses"."CityId"
LEFT OUTER JOIN "Tenant" AS "Houses.Tenants" ON "Houses"."id" = "Houses.Tenants"."HouseId"
LEFT OUTER JOIN "Item" AS "Houses.Tenants.Items" ON "Houses.Tenants"."id" = "Houses.Tenants.Items"."TenantId";
```

Will this work @mickhansen  @janmeier @felixfbecker, @cbauerme should I send a PR to your branch or commit here on `sequelize` with new branch

lib/dialects/abstract/query-generator.js
felixfbecker: Why not `connector = connector || '.';`

sushantdhiman:
Ok new fix does break some tests, If I am able to fix them I will update my code

lib/dialects/abstract/query-generator.js
sushantdhiman: much better

mickhansen:
@sushantdhiman Are you sure? All testing seems to indicate `.` is not allowed in sqlite when doing inner joins in parenthesis.

sushantdhiman:
@mickhansen Sqlite tests are breaking because of this issue. Can we add a property called `JOIN_ALIAS_CONNECTOR` in Abstract query and default it to `.` for SQLite we can override it with `->` or `_`, That will solve the problem and will keep the `.` for other dialects

mickhansen:
@sushantdhiman Yes, we know, that's the entire point of the discussion of this pull request. :)

I don't see a good reason for breaking behaviour for SQlite if we don't have to, the more feature equality we have the better.

sushantdhiman:
It looks ugly in query, but the solution does work. If we want to keep it `->` or `_` for uniformity sake thats good too.

I have tested the PR at my end and gone through the code too. It looks like a good solution for now.

mickhansen:
@sushantdhiman I agree it's not pretty, and i would like to keep `.` but if we are to support SQLite we should make our best effort to do so.

The primary breakage will be raw statements targeting a nested column - We'll have to see if we can remedy those with better syntax somehow.

cbauerme:
@mickhansen I was able to fix the targeted columns issues so that periods in an associated table alias given to it will be replaced with arrows or whatever connecter ends up being used.  I have no solution for `raw`, however.

cbauerme:
So I did a little more poking around and I think I have the specific problems we are having with periods in the parenthesized joins. 

The parentheses are triggering a [subquery materialization](http://www.sqlite.org/tempfiles.html#views). That link specifically talks about sub queries that are a `select` with another `select`, but I know from my example queries we are also seeing materialization. 

The function that is used to resolve the columns we are selecting will only accept "spans" (table.field) that look like `field.name`, and it only expects the one `.`. So when we give it `Project.User.id`, it tried to find a field `User` on table `Project`. 

Basically, any SQLite query that triggers the materialization will never work with aliases with periods. I am going to look into "flattening" the query like the link describes, but I suspect even if I am able to flatten the top include, more nested includes will never be able to be flattened.

mickhansen:
@cbauerme Our best bet going forward is probably to have better APIs for targeting columns, and just be clear about raw being able to break anytime.

sushantdhiman:
@mickhansen I think we should adopt `->` for all SQLite alias connectors and keep using the `.` for others. This will allow users to target columns with `$$` with much more safety.

For other MySQL, PG and MSSQL `$$` based targeting will work as usual. For SQLite users will have to use `->` in all cases of `$$` , so there will be no confusion for selecting when to use `.` or `->`

mickhansen:
@sushantdhiman That's less than ideal since a lot of people switch between sqlite and another dialect. We should strive as much as possible to provide the exact same experience whereever feature sets allows us.

> For other MySQL, PG and MSSQL $$ based targeting will work as usual. For SQLite users will have to use -> in all cases of $$ , so there will be no confusion for selecting when to use . or ->

The syntax is not changing for `$$`, it's still a dot interface for the user but it's rewritten for the query.

lib/dialects/abstract/query-generator.js
cbauerme: After getting burned on values that resolve to false but are valid I tend to be paranoid about just saying `!var`, but I guess it is pretty safe in this case. I will go ahead and use `connector || '.'`

sushantdhiman:
Oh yes, because of aliases we can still use `$$` even if actual accessors are combined with `->` or `.`

mickhansen:
@sushantdhiman That doesn't actually matter, `$$` is an operator, we control how it's ultimately used. That the input syntax uses `.` doesnt mean that the output has to.

cbauerme:
Attempts to flatten were no good, I think what we have it is what we get. Periods will not work inside parentheses in sqlite, period.

mickhansen:
@cbauerme Yeah leave it as is. Just waiting on word from @janmeier

mickhansen:
ping @janmeier

janmeier:
Very nice work, looking good to me

sushantdhiman:
One thing I just noticed, file mode for four files has been changed from 644 to 755. Does this affect us ? @mickhansen @janmeier

janmeier:
The mode changes shouldn't matter I think :)

lib/dialects/abstract/query-generator.js
```diff
@@ -774,7 +774,10 @@ const QueryGenerator = {
     Currently this function is only used for ordering / grouping columns and Sequelize.col(), but it could
     potentially also be used for other places where we want to be able to call SQL functions (e.g. as default values)
   */
-  quote(obj, parent, force) {
+  quote(obj, parent, force, connector) {
+
+    connector = connector || '.';
```
mickhansen: extranous newlines a few places

lib/dialects/abstract/query-generator.js
mickhansen: @cbauerme

test/integration/include.test.js
mickhansen: @cbauerme

test/integration/include.test.js
```diff
@@ -858,4 +858,112 @@ describe(Support.getTestDialectTeaser('Include'), function() {
...
+                required: true,
+              }
+            ]
+          }
+        ],
+      }).then(function (teams) {
+        expect(teams).to.have.length(1);
+      });
+
+    });
```
mickhansen: nitpick but we generally dont have extra newlines at the first and last part of an it block

mickhansen:
We'll need a changelog entry describing this breaking change and how code (if anyone has any raw sql) should be changed.

cbauerme:
I am trying to figure out how someone would be affected by these changes with raw SQL. What options should of find should I look at?

test/integration/include.test.js
cbauerme: I went ahead and cleaned these up.

lib/dialects/abstract/query-generator.js
cbauerme: renamed

test/integration/include.test.js
cbauerme: I removed the extra function.

cbauerme:
Ok, I got that Postgres issue cleared up, but I have no clue to the MSSQL failure. Does anyone have any clues?

cbauerme:
I am also still trying to figure out how someone's existing code would be affected by the change.  Here is my thought process so far:

1. Find with nothing special is fine since the changes are internal and taken care of
2. Find with raw is fine since the returns are still the same as before
3. Find with table names in `$` are fine since we take care of those specifically

Am I missing some cases?

mickhansen:
@cbauerme Cases that will break will mostly relate to those that use raw queries to match column names, that's unfortunate but no way around it really.

Rerunning MySQL.

A changelog entry and i think we're good to go.

cbauerme:
@mickhansen  OK, made sure the change to the table names is called out in the BC breaks lines of the changelog.

mickhansen:
> Table names of a select query have change internally from 'originModel.associatedModel.field' to 'originModel->associatedModel.field'

Hmm, so we're mixing rather than using just one seperator.
I'm not sure if that's ideal because most cases will still work or it's confusing because you need to think about which one to use.

@janmeier thoughts?

janmeier:
I think that's the best we can get

sushantdhiman:
@mickhansen

lib/dialects/abstract/query-generator.js
```diff
@@ -1578,128 +1273,294 @@ const QueryGenerator = {
...
+
+      for (const childInclude of include.include) {
+
+        if (childInclude.separate || childInclude._pseudo) {
+          continue;
+        }
+
+        const childJoinQueries = this.generateInclude(childInclude, includeAs, topLevelInfo);
+
+        // console.log(childJoinQueries);
```
mickhansen: ^

lib/dialects/abstract/query-generator.js
```diff
@@ -1578,128 +1273,294 @@ const QueryGenerator = {
...
-    let fragment = 'SELECT ' + attributes.join(', ') + ' FROM ' + tables;
+    if (joinQuery.attributes.subQuery.length > 0) {
+      attributes.subQuery = attributes.subQuery.concat(joinQuery.attributes.subQuery);
+    }
 
-    if(mainTableAs) {
-      fragment += ' AS ' + mainTableAs;
+    if (include.include) {
+
+      for (const childInclude of include.include) {
```
mickhansen: Minor nitpick, you seem to add an extra newline after each `{` which is not how we generally do it in the rest of the codebase (goes for a large part of the diff)

cbauerme:
Sorry about that, I went through and tried to get all the left brace double newline, and any empty space that was a double newline as well.

PR #6177, opened 2016-06-25
Added the ability to catch invalid scopes

KieronWiltshire:
I literally just added the ability to catch invalid scope errors. My api allows user's to input scopes, and I need a way of catching those input errors. Adding this to the sequelize package seemed like the perfect work around for me, and it doesn't hurt to be there ;)

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -51,7 +51,7 @@ exports.SequelizeScopeError = SequelizeScopeError;
  * @constructor
  */
 class ValidationError extends BaseError {
-  constructor(message) {
+  constructor(message, errors) {
     super(message);
     this.name = 'SequelizeValidationError';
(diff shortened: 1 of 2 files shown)


KieronWiltshire:
Just stating, I'd also like to see things errors as AssociationError defined, this will make my api really easy to catch these errors and return my custom BadRequestError.

mickhansen:
LGTM, /cc @janmeier

mickhansen:
Ideally we should NEVER throw `Error`, always atleast use `BaseError` so that users can identify sequelize errors.

KieronWiltshire:
@mickhansen I'd be more than happy to go through the entire code base and do this if you like?

mickhansen:
@KieronWiltshire We'd certainly welcome

PR #6181, opened 2016-06-28
fixes msssql only error when no order by but has offset or limit present

AckerApple:
Quite the simple fix. Applies to MSSQL only.

To reproduce: Create two models and relate them. Then query one with an include of another and then ensure the query has NO order by BUT has an OFFSET or LIMIT. When the order by is automatigically prepend by mssql/query-generator.js, the table reference is not present.

I have tested in my environment but unsure how to make ya'll certain this is the right fix. It is though, I'm using it.

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -685,6 +685,11 @@ var QueryGenerator = {
     if (options.limit || options.offset) {
       if (!options.order || (options.include && !orders.subQueryOrder.length)) {
         fragment += (options.order && !isSubQuery) ? ', ' : ' ORDER BY ';
+
+        if(model.name){
+          fragment += this.quoteIdentifier(model.name) + '.';
+        }
+
         fragment += this.quoteIdentifier(model.primaryKeyField);
       }
 
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -685,12 +685,7 @@ var QueryGenerator = {
     if (options.limit || options.offset) {
       if (!options.order || (options.include && !orders.subQueryOrder.length)) {
         fragment += (options.order && !isSubQuery) ? ', ' : ' ORDER BY ';
-
-        if(model.name){
-          fragment += this.quoteIdentifier(model.name) + '.';
-        }
-
-        fragment += this.quoteIdentifier(model.primaryKeyField);
+        fragment += this.quoteTable(model.name) + '.' + this.quoteIdentifier(model.primaryKeyField);
       }
 
       if (options.offset || options.limit) {
--- a/test/unit/sql/offset-limit.test.js
+++ b/test/unit/sql/offset-limit.test.js
@@ -26,6 +26,14 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
     };
 
     testsql({
+      limit: 10,//when no order by present, one is automagically prepended, test it's existence
+      model:{primaryKeyField:'id', name:'tableRef'}
+    }, {
+      default: ' LIMIT 10',
+      mssql: ' ORDER BY [tableRef].[id] OFFSET 0 ROWS FETCH NEXT 10 ROWS ONLY'
+    });
+
+    testsql({
       limit: 10,
       order: [
         ['email', 'DESC'] // for MSSQL


janmeier:
Unit tests would be good :) https://github.com/sequelize/sequelize/blob/master/test/unit/sql/offset-limit.test.js

AckerApple:
I've tried with my brother (https://github.com/ackerdev) to do testing of Sequelize in the past and due to the enviroment setup, we burned to much time just trying to test.

I know how to unit test, help guide me better to make a test that doesn't cost me a large Sequelize test dev environment with multiple database products.

janmeier:
Though I recognized the `acker` part from before ;)

You should be able to run the unit tests without having any database setup - simply run `npm run test-unit-mssql`.

AckerApple:
Ya know, I think that just might work. I will give it my all! You lead me to the water, I can drink from here

AckerApple:
If you could reply quickly, how do I tell sequelize/test/unit/sql/offset-limit.test.js to only run "this" one test.

In Mocha that I know, it's it.only('name', test)

How can I limit 635 tests, down to one. Link to docs perhaps?

AckerApple:
I had a lucky guess with suite.only ... I got it

AckerApple:
Thanks for making that easy. I have pushed my code. Please let me know if I need to issue a new pull request.

juansev:
Beware that this doesn't solve the issue with the sorting when you are sorting by the primary key #6168 
. AckerApple, were you able to run the tests? Did you need to run anything special?

AckerApple:
Please elaborate. On when hand, I think you are telling me my fix is not passing a test. On the other hand, I think your hoping that I can fix a related issue and I bet I can. The issue you pointed out #6168, I bet would be fixed by my pull-request.

What more information can you give me? I CAN DO-IT!!!

AckerApple:
To elaborate on my own commit, juansev. I got the test to run. It passed. I pulled the code into my production server and everything runs fine.

This would solve even issues related to "sorting by the primary key"... I most certain of it. I would need an example or more info to help. I'm an MSSQL master and I can work with this Sequelize code well.

juansev:
Hello AckerApple. Maybe i didn't express myself correctly but i didn't say that your code wasn't passing the tests, just to know if you could pass the tests because i can't (i tried to use docker container but some dependencies were missing i was trying to know what was i doing wrong). 

In order to solve the other issue about sorting by the same primaryKey that produces the "duplicate column" error you have to modify the condition you can check my pull request at https://github.com/sequelize/sequelize/pull/6168/files but it is only to change the condition so the extra fragment is only added when it is really needed :
`
 if (options.limit || options.offset) {
 -      if (!options.order || (options.include && !orders.subQueryOrder.length)) {
 -        fragment += (options.order && !isSubQuery) ? ', ' : ' ORDER BY ';
 -        fragment += this.quoteIdentifier(model.primaryKeyField);
 +      if (!options.order || (isSubQuery && !orders.subQueryOrder.length)) {
 +        fragment += ' ORDER BY ' + this.quoteIdentifier(model.primaryKeyField);
        }
 `

juansev:
Final code with your modification and mine would looke like:

```
 if (options.limit || options.offset) {
        if (!options.order || (isSubQuery && !orders.subQueryOrder.length)) {
          fragment += ' ORDER BY ';

        if(model.name){
          fragment += this.quoteIdentifier(model.name) + '.';
        }

          fragment += this.quoteIdentifier(model.primaryKeyField);
        }
```

AckerApple:
OK ok ok. I'm on the right page. First, yes, I got my test to pass.

Second, your code seem to change the line FROM:
```
fragment += (options.order && !isSubQuery) ? ', ' : ' ORDER BY ';
```
INTO:
```
fragment += ' ORDER BY ';
```


I have two problems with this:

#1, I don't know whats going on with the !isSubQuery and it's not affecting me so I don't want to drag it into my pull request

#2, This is some of my first public pull requests, and not that you are messing with it BUT I don't even want to mess with it. I'm holding my gut and crossing my fingers for this pull request to be accepted and I don't want to mess with that. My fix is solid and it'll be left at that until I gain more respect points around here

juansev:
Just ignore my posts and go on with your pull request. I will handle it in my own. Good luck!

lib/dialects/mssql/query-generator.js
```diff
@@ -685,6 +685,11 @@ var QueryGenerator = {
     if (options.limit || options.offset) {
       if (!options.order || (options.include && !orders.subQueryOrder.length)) {
         fragment += (options.order && !isSubQuery) ? ', ' : ' ORDER BY ';
+
+        if(model.name){
+          fragment += this.quoteIdentifier(model.name) + '.';
+        }
```
BridgeAR: There is always a `model.name`. So you do not have to test for this. And as you want to add the table name you should use `.quoteTable(model)`. That way also schemas are supported.

AckerApple: Understood well. I will update my code in about 6 days

BridgeAR: ?

AckerApple: On vacation where I only get signal when I come down the Smokey Mountains of North Carolina. I will change my code to fit your specifications in 5 days

itsravenous:
@AckerApple eagerly awaiting your descent from the peaks as this is causing me problems too!

AckerApple:
@itsravenous, I hella appreciate the message of enthusiasm. I am sitting shotgun driving home and will update the code first thing in the AM. This change is very important to me to. I think your message helps keep this change staying in motion, so double appreciative. I will get done asap

lib/dialects/mssql/query-generator.js
AckerApple: I have made the changes. Thank you kindly for guiding me towards a proper fix. I am still learning the ropes, do let me know if further action is recommended.

AckerApple:
FYI, awaiting review. I have made the recommended changes.

sushantdhiman:
LGTM, would like @mickhansen or @janmeier to merge after their review

itsravenous:
Thanks @AckerApple @sushantdhiman @mickhansen!

Issue #6123, opened 2016-06-17
findAndCountAll with null nested field


vangoz:
I need to left join to a table and check with `is null`, so I do it with nested field. 

```js
clause.include = [{
  model: 'Table',
  required: false
}]
clause.where['$Table.x$'] = null;
```
The query works as expected with findAll(), however with findAndCountAll() it returns error
`The multi-part identifier \"Table.x\" could not be bound."`.

I think it happens because the count query does not do the actual join because of `required: false`.

janmeier:
Please create a complete [SSCCE](HTTP://SSCCE.ORG/)

Issue #5680, opened 2016-03-31
Problems using findAndCount when looking for models with non-existent associations


VividVisions:
Hi! 

I have the models Item and Category, which have a n:m relation. Now, I want to get all Items, where no Category has been assigned yet, which works when I use _findAll_ but fails when I use _findAndCount[All]_ (for pagination).

This works:
```js
models.Item.findAll({
    attributes: ['name'],
    where: sequelize.literal('"categories"."id" IS NULL'),
    include: [models.Category]
});
```

This does not:
```js
models.Item.findAndCountAll({
    attributes: ['name'],
    where: sequelize.literal('"categories"."id" IS NULL'),
    include: [models.Category]
});
```

I'm getting the following error:
`Error:  missing FROM-clause entry for table "categories"`

In a similar setup, I'm getting the error `SequelizeDatabaseError: column Item.val does not exist` as soon as I change _findAll_ to _findAndCount_.

I'm using: 
-  Sequelize v3.20.0
-  Dialect: postgres
-  PostgreSQL v9.4.4.1

janmeier:
Try
```js
where: { '$categories.id$': null }
```

VividVisions:
@janmeier Thanks, but same effect.

janmeier:
The problem is that includes are removed from the count part - And when using literal or `$$`, we can't figure out that we should actually keep the include. Try adding `required:true` to the include, see https://github.com/sequelize/sequelize/issues/2663

VividVisions:
`required:true` would return only Items with Categories in the first place.

janmeier:
Yes - unfortunately there is no good way to fix the issue right now - we need to be able to figure out that the $$ clause means we should keep the include, https://github.com/sequelize/sequelize/pull/4195 and https://github.com/sequelize/sequelize/issues/3095

VividVisions:
That's unfortunate. :( I don't really get why the query has to be altered for the count in the first place but I guess, there's a reason. Making it optional, as mentioned in the linked issues, would be great.

In the meantime, I tried a Postgres-specific workaround using `COUNT(*) OVER()`:

```javascript
models.Item.findAll({
    attributes: [
         [sequelize.literal('COUNT(*) OVER()'), 'all'],
        'name'
    ],
    where: { '$categories.id$': null },
    include: [models.Category]
});
```

This works well until I add `limit`. Then the `missing FROM-cause` error occurs again.

mickhansen:
[...]

Issue #5526, opened 2016-03-03
Eager loading counts of related models


geoffreak:
I can't seem to figure out the best way to eager load the count of a related model. For example, if an `account` has `subscriptions`, how should you load the number of subscriptions on the account without having to select all subscriptions?

The best workaround I've been able to come up with is something like this:
```coffee
@models.account.findAll
  include:    [
    model:      @models.subscription
    as:         'Subscriptions'
    attributes: ['id']
    required:   false
  ]
```
There doesn't seem to be any way to just get a subscription count without having to select at least one field to return from the query. Is there a way just to return a count?

mickhansen:
Not when using includes. Aggregates and includes/joins are generally tricker.
Usually what works best, atleast for PG, are window functions so you don't get those pesky aggregation errors.

Issue #5465, opened 2016-02-21
include.where missing from subquery filter when using scopes


jylinman:
I am having some issues joining some tables, here is the situation:

**Mysql Tables**
tickets
tickets_pools (n:m)
pools

When I do a join on the tickets_pools using the `include: [models.Pools]` it joins just fine if the `tickets_pools` table has less than 20 records of relations for a `pool` to `ticket`.  If it has more than 20, I can't do a { where: [IDS] } if the IDS are falling in the > 20 records.

The code:

``` javascript
      # scope
      pools: function(pools) {
        return {
          distinct: ['Ticket.id'],
          include: [{
            model: this.db.Pool,
            through: this.db.TicketPool,
            where: {
              id: pools
            }
          }]
        };
      }

      # query
      query = query.scope({
        method: ['pools', pools]
      });
```

The SQL generated by the simple association:

``` sql
SELECT `Ticket`.*, `Pools`.`id` AS `Pools.id`, ..... 
    FROM (SELECT `Ticket`.`id`, `Ticket`.`parent_id` AS `parentId`, `Ticket`.`creator_id` AS `creatorId`, ....  
    WHERE `Ticket`.`deleted_at` IS NULL AND ( SELECT `tickets_pools`.`ticket_id` 
    FROM `tickets_pools` AS `tickets_pools` 
INNER JOIN `pools` AS `Pool` ON `tickets_pools`.`pool_id` = `Pool`.`id` 
    WHERE (`Ticket`.`id` = `tickets_pools`.`ticket_id`) 
    LIMIT 1 ) IS NOT NULL LIMIT 20) AS `Ticket` 
INNER JOIN (`tickets_pools` AS `Pools.tickets_pools` 
INNER JOIN `pools` AS `Pools` 
    ON `Pools`.`id` = `Pools.tickets_pools`.`pool_id`) 
    ON `Ticket`.`id` = `Pools.tickets_pools`.`ticket_id` 
        AND (`Pools`.`deleted_at` IS NULL AND `Pools`.`id` IN (155, 156));
```

As you can see above, the _IS NOT NULL LIMIT 20_ is automatically added.  Is there a way to avoid this / do it better, or is it a bug?

On a second note, I can create a separate bug, but the distinct shows only in the COUNT() query and disappears from the actual content query.  Thoughts?

janmeier:
Please show model definitions and find / findAll call

mickhansen:
[...]

PR #6190, opened 2016-06-30
Fix count queries with includes

### LINKED ISSUE #1142: Custom Nested Select Attribute?
Say I need to fetch all users (with pagination -- offset + limit), along with a count of each user's "form" count --  in SQL, I'd do something like this:

```
SELECT *, (SELECT COUNT(*) FROM `Forms` WHERE UserId = Users.id) FROM `Users` LIMIT 10;
```

How can I achieve the same thing, using ```db.User.findAndCountAll({ limit: 10 })``` ?

BridgeAR:
### Pull Request check-list

### Description of change

Currently the `count` and `findAndCount` and diverge in the build queries with the same query object. This unifies those two, removes unnecessary duplicated checks and fixes a bug which quite a few people ran into by optimistically removing all includes for [performance reasons](82403cfc1d165e2a01b6b0170efae6fc0355ba85).

I want to have fast queries but primarily stability is important. I'm willing to have a look at this in more detail to improve the performance later.

Fixes #1142 #2145 #2968 #3936 #4146 #4194 #5526 #5680 #6123 (I went through to many issues to always be certain but likely all get fixed by this).
Fixes partly #2886 #3319 #4737 #5465 (people that commented had similar issues or a part of the issue is fixed with this)
Somehwat related #2713

I know there should be tests included and I'll add them as soon as you're happy with the other code.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,7 @@
 # Future
-- [FIXED] Fixed an issue where custom-named model fields break when offsetting, ordering, and including hasMany simultaneously. [#5985] (https://github.com/sequelize/sequelize/issues/5985) 
+- [FIXED] Fixed an issue where custom-named model fields break when offsetting, ordering, and including hasMany simultaneously. [#5985](https://github.com/sequelize/sequelize/issues/5985)
+- [FIXED] Don't remove includes from count queries and unify findAndCount and count queries. [#6123](https://github.com/sequelize/sequelize/issues/6123)
+- [FIXED] `Model.count` don't include attributes [#5057](https://github.com/sequelize/sequelize/issues/5057)
 
 # 3.23.3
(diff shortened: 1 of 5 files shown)


janmeier:
Yeah, this "optimization" is causing too many problems to many sense - Lets KISS ;)

BridgeAR:
I backported all the comments from the v4 branch and included the attributes fix too.

netdragonboberb:
'distinct' works now in FindAndCountAll, but it was never added to FindOptions so Typescript compiler complains.

alitaheri:
@netdragonboberb This is not a typescript project. Please open issue on https://github.com/DefinitelyTyped/DefinitelyTyped, that's where the type definitions are maintained.

felixfbecker:
@alitaheri @netdragonboberb I maintain the TypeScript typings at https://github.com/types/npm-sequelize, please open an issue there.

PR #6191, opened 2016-06-30
Fix model.count with includes and specified column

BridgeAR:
### Pull Request check-list

### Description of change

This fixes a missed case for the new col option from #6114

[edit]
__Note__: Count is still not working with multiple fields e.g. a composite primary key. And is it really the best idea to add a new option instead of overloading the distinct option to also allow strings instead of a boolean?
[/edit]


janmeier:
LGTM, needs a rebase

BridgeAR:
Done

Issue #5963, opened 2016-05-25
Switch from JSHint to ESLint


felixfbecker:
ESLint has much better support and way more rules. There is a CLI tool for converting JSHint configs to ESLint, I already tried it out on Sequelize and ESLint caught a whole more errors. For example, JSHint doesn't seem to catch indentation errors. There were a lot of errors in the code like `expected indentation of 4 characters but found 3` that break indentation-based code collapsing in editors.

Would do PR

mickhansen:
Sounds good to me.

PR #6192, opened 2016-06-30
Fix includes being removed from count queries

BridgeAR:
### Pull Request check-list

### Description of change

See #6190 for more details (same fix applied against v.3)

### DIFF
--- a/test/integration/model.test.js
+++ b/test/integration/model.test.js
@@ -1772,24 +1772,22 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       });
     });
 
-    it('supports distinct option', function() {
-      var Post = this.sequelize.define('Post', {});
-      var PostComment = this.sequelize.define('PostComment', {});
+    it('supports distinct option', function () {
+      const Post = this.sequelize.define('Post', {});
+      const PostComment = this.sequelize.define('PostComment', {});
       Post.hasMany(PostComment);
-      return Post.sync({ force: true }).then(function() {
-        return PostComment.sync({ force: true }).then(function() {
-          return Post.create({}).then(function(post) {
-            return PostComment.bulkCreate([{ PostId: post.id },{ PostId: post.id }]).then(function() {
-              return Post.count({ include: [{ model: PostComment, required: false }] }).then(function(count1) {
-                return Post.count({ distinct: true, include: [{ model: PostComment, required: false }] }).then(function(count2) {
-                  expect(count1).to.equal(2);
-                  expect(count2).to.equal(1);
-                });
-              });
-            });
-          });
-        });
-      });
+      return Post.sync({ force: true })
+        .then(() => PostComment.sync({ force: true }))
+        .then(() => Post.create({}))
+        .then((post) => PostComment.bulkCreate([{ PostId: post.id },{ PostId: post.id }]))
+        .then(() => Promise.join(
+            Post.count({ distinct: false, include: [{ model: PostComment, required: false }] }),
+            Post.count({ distinct: true, include: [{ model: PostComment, required: false }] }),
+          (count1, count2) => {
+            expect(count1).to.equal(2);
+            expect(count2).to.equal(1);
+          })
+        );
     });
 
   });
--- a/test/integration/model/count.test.js
+++ b/test/integration/model/count.test.js
@@ -133,5 +133,22 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       });
     });
 
+    it('should be able to use where clause on included models', function() {
+      const queryObject = {
+        col: 'username',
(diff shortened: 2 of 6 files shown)


test/integration/model/count.test.js
```diff
@@ -133,5 +133,47 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
+      }).then((count) => {
+        expect(parseInt(count)).to.be.eql(0);
+      });
+    });
+
+    it('should have the same count result as findAndCount', function() {
+      let result;
+      const queryObject = {
+        include: [this.Project],
+        logging: function (sql) {
```
janmeier: Whats going on with the logging here?

test/integration/model/count.test.js
```diff
@@ -133,5 +133,47 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
+      queryObject.where['$Projects.name$'] = 'project1';
+      return this.User.count(queryObject).then((count) => {
+        expect(parseInt(count)).to.be.eql(1);
+        const queryObject = {
+          col: 'username',
+          include: [this.Project],
+          where: {}
+        };
+        queryObject.where['$Projects.name$'] = 'project2';
+        return this.User.count(queryObject);
```
janmeier: No need to create the object first
```js
{
  col: 'username',
  include: [this.Project],
  where: {
    '$Projects.name$': 'project1'
  }
};
```

janmeier:
Good to have you back @BridgeAR ! :)

test/integration/model/count.test.js
BridgeAR: :+1: I'm to much used to using variables in the names ;-)

test/integration/model/count.test.js
BridgeAR: The logging is used to compare the ouput of the `findAndCount` count query and the `count` query. They should always be identical with one minor exception: If you pass attributes in a findAndCount query the attributes should be stripped and in a count they should be kept.

janmeier: Hmm, that feels a bit dirty 

You could also write this as a unit test, where you verify that the arguments to find and count are correct

test/integration/model/count.test.js
BridgeAR: Switching to eslint would be nice I guess. The rules there are much more powerful than the jshint ones. And maybe [standard](https://github.com/feross/standard) would also be a good idea to use.

janmeier: Yep, eslint would be nice :) https://github.com/sequelize/sequelize/issues/5963

BridgeAR:
Comments incorporated

test/unit/model/count.test.js
```diff
@@ -0,0 +1,67 @@
...
+  , Support = require(__dirname + '/../support')
+  , current = Support.sequelize
+  , sinon = require('sinon')
+  , DataTypes = require(__dirname + '/../../../lib/data-types')
+  , Promise = require('bluebird');
+
+describe(Support.getTestDialectTeaser('Model'), () => {
+  describe.only('method count', () => {
+    before(() => {
+      this.oldFindAll = current.Model.findAll;
```
janmeier: You can preface with `sandbox = sinon.sandbox.create(), and use `sandbox.stub`. Then you just need to call `sandbox.restore()` in  ` afterEach` to reset all the stubs

janmeier:
LGTM, would like @mickhansen or @sushantdhimans comments as well

test/integration/model.test.js
```diff
@@ -1772,22 +1772,23 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
-              return Post.count({ include: [{ model: PostComment, required: false }] }).then(function(count1) {
-                return Post.count({ distinct: true, include: [{ model: PostComment, required: false }] }).then(function(count2) {
-                  expect(count1).to.equal(2);
-                  expect(count2).to.equal(1);
-                });
-              });
-            });
-          });
+      return Post.sync({ force: true }).then(() => {
+        return PostComment.sync({ force: true });
```
sushantdhiman: Unnecessary returns in arrow functions, you can do ``` () => ( PostComment.sync({ force: true }) )```, Same applies for this whole chain

sushantdhiman:
LGTM, I agree that performance check is causing too many issues, we should remove it for now.

Later on I would report if there are any significant deviation in performance.

BridgeAR:
Done. I removed the distinct default too, as this depicts the regular database behavior.

janmeier:
[No-no-no-NONET-kill!(https://en.wikipedia.org/wiki/Nonet_(music)) Pretty sure thats a new record for most issues fixed in one PR :)

PR #6198, opened 2016-07-01
Show example of wildcards and array replacements

elliotthill:
Show example of wildcards and array replacements in raw-queries documentation. People that are using sequelize mainly for raw queries are likely to miss this information that is only on other documentation pages.


janmeier:
Thanks :)

PR #6200, opened 2016-07-01
Better validation of attribute definitions

jharting:
### Pull Request check-list

### Description of change

When writing more involved attribute definitions with functions (e.g. `attributes: [[sequelize.fn('COUNT', sequelize.fn('DISTINCT', sequelize.literal('CASE WHEN `rulez.acks`.`id` IS NULL AND...)` it happened to me couple of times already that I forgot to specify the actual alias for that attribute. In other words, the attribute was mistakenly defined as `['definition']` instead of `['definition', 'alias']`. In such cases, Sequelize fails with the following error:

```
Unhandled rejection TypeError: s.replace is not a function
    at removeTicks (/opt/insights/sequelize/lib/utils.js:445:12)
    at Object.addTicks (/opt/insights/sequelize/lib/utils.js:439:21)
    at Object.quoteIdentifier (/opt/insights/sequelize/lib/dialects/mysql/query-generator.js:293:18)
    at Object.quoteIdentifiers (/opt/insights/sequelize/lib/dialects/abstract/query-generator.js:909:19)
    at /opt/insights/sequelize/lib/dialects/abstract/query-generator.js:1052:82
```

This error message does not help at all. 

With this PR Sequelize detects when the alias param is missing a throws appropriate error.

### DIFF
--- a/test/integration/model/find.test.js
+++ b/test/integration/model/find.test.js
@@ -155,6 +155,13 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         });
       });
 
+      it('should fail with meaningful error message on invalid attributes definition', function() {
+        expect(this.User.findOne({
+          where: { id: 1 },
+          attributes: ['id', ['username']]
+        })).to.be.rejectedWith('["username"] is not a valid attribute definition. Please use the following format: [\'attribute definition\', \'alias\']');
+      });
+
       it('should not try to convert boolean values if they are not selected', function() {
         var UserWithBoolean = this.sequelize.define('UserBoolean', {
           active: Sequelize.BOOLEAN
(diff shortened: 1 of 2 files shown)


test/integration/model/find.test.js
```diff
@@ -155,6 +155,17 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         });
       });
 
+      it('should fail with meaningful error message on invalid attributes definition', function() {
+        return this.User.findOne({
```
janmeier: `expect(..).to.be.rejectedWith(message)`

jharting:
Updated

janmeier:
LGTM - If you add an entry to the changelog this should be good to go

jharting:
I added a changelong entry. The test failure (Error: SQLITE_BUSY: database is locked) looks unrelated.

janmeier:
Thanks :)

PR #6204, opened 2016-07-01
Fix changing multiple ENUM columns on Postgres

### LINKED ISSUE #6203: Postgres changeColumnQuery can generate duplicate ENUM definitions
## What you are doing?
When running the `queryGenerator.changeColumnQuery` for the Postgres dialect (I haven't checked if other dialects are affected), changing multiple attributes with the ENUM datatype. If one ENUM is provided, it must not be defined "last" in the parameters object passed in (because we are looping over object keys, which occurs in an unreliable order).
 
The ENUM definition should only be included in the final generated query exactly once.

The same ENUM definition may be included multiple times in the final generated query.

I am working on a failing test to illustrate this, and haven't yet encountered the issue in my actual projects, but @felixfbecker has also looked over this and agrees that the issue exists.

The actual problem is with this line of code:

3.x: https://github.com/sequelize/sequelize/blob/v3/lib/dialects/postgres/query-generator.js#L246
4.x: https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/query-generator.js#L217

Because it mutates the `query` template, which is reused on subsequent loop iterations, any loop iteration that occurs _after_ an enum column will incorrectly contain a duplicate enum definition.

The solution would be to mutate `attrSql` rather than `query` when encountering an enum column during a loop.

__Dialect:__ postgres
__Sequelize version:__ 3.x / 4.x

I'll work on a PR too. :)

joshrickert:
### Pull Request check-list

### Description of change

Addresses #6203 for v3 only, includes test.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
 # Future
 - [FIXED] Fixed an issue where custom-named model fields break when offsetting, ordering, and including hasMany simultaneously. [#5985] (https://github.com/sequelize/sequelize/issues/5985) 
+- [FIXED] Fixed an issue where changing multiple ENUM columns in PostgreSQL could break. [#6203] (https://github.com/sequelize/sequelize/issues/6203)
 
 # 3.23.3
 - [FIXED] Pass ResourceLock instead of raw connection in MSSQL disconnect handling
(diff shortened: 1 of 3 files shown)


sushantdhiman:
@joshrickert Its already merged to `master`, but if you still want it to be backported fell free to rebase and we will merge it to v3

PR #6206, opened 2016-07-02
support moment js date as input

sherpya:
Simple patch to support moment js date format as input by converting into a js Date object

### DIFF
--- a/lib/sql-string.js
+++ b/lib/sql-string.js
@@ -35,6 +35,11 @@ function escape(val, timeZone, dialect, format) {
     break;
   }
 
+  // moment
+  if (val.toDate) {
+    val = val.toDate();
+  }
+
   if (val instanceof Date) {
     val = dataTypes[dialect].DATE.prototype.stringify(val, { timezone: timeZone });
   }


lib/sql-string.js
```diff
@@ -35,6 +35,11 @@ function escape(val, timeZone, dialect, format) {
     break;
   }
 
+  // moment
+  if (val.toDate) {
```
sushantdhiman: Use `isMoment` [is-a-moment](http://momentjs.com/docs/#/query/is-a-moment/)

sushantdhiman:
Yes we should support `moment.js`, its very common among node developers.

lib/sql-string.js
sherpya: isMoment is specific, using toDate could support everything implements this "interface"

sushantdhiman: We need specific moment support here, I would rather avoid duck typing. You could also use `instanceof` instead of `isMoment`. But we needs to be sure its moment nothing else.

sherpya: so it would be better to use directly moment's timezone stuff and format()

sushantdhiman: Passing `format` output is already supported by sequelize but passing `moment` object isn't.

With this PR we will be able to support it. But we need to check for `isMoment` or `instanceof moment` rather than `if( val.toDate )`

sushantdhiman:
Ping @sherpya

lib/sql-string.js
felixfbecker: I generally disagree with @sushantdhiman, duck typing should always be preferred. But in this case the code should just be refactored so that the input is directly passed to `moment`, which accepts a `Date`, a string or a `moment`.

sushantdhiman: Passing anything to moment can lead to some crazy issues. Like passing `really not a date` to `moment`. Rather than trying to support every possible string out there, we should stick with well defined and valid moment object. Since @felixfbecker has raised this point. I think adding `isValid` with `isMoment` check will not hurt either.

@felixfbecker I am not against the duck typing :), But for this particular case I would prefer valid moment object.

felixfbecker: Why? Anything that has a toDate method can obviously be converted to a Date. Checking isValid can still be done.

sushantdhiman: We can't claim that any thing with `toDate` will follow our specifications, murphy's law is real :) Thus sticking with `isMoment` check with `isValid` is better solution to achieve stability

felixfbecker: So you mean we cannot be sure that a `toDate` method returns a `Date`?

sushantdhiman: Yes, we can't (if its not moment). Its all up to implementor what `toDate` might return.

felixfbecker: I highly argue against that. Interfaces in JavaScript and TypeScript are based around duck-typing, because there is no `implements` like in Java or overloading. So the method name is the contract. If you call your method `toDate`, it should return a `Date`. Event the standard JavaScript objects use loose interface like this, for example `Promise.resolve()`. If your object has a `.then()` method, it is expected to accept a promise handler and return a new promise, without any `implements` declaration. The only contract is the method name here.

felixfbecker:
I wanted to point out that this could easily be implemented in userland once #6290 is ready, but will complicate #6270 because we will not only need parse/stringify support for moments, but also for cloning and for comparison.

lib/sql-string.js
hagabaka: If we can't assume `toDate` returns a `Date`, by the same logic how can we assume the object is a moment when `isMoment` returns true?

janmeier:

PR #6213, opened 2016-07-03
Index once

### LINKED ISSUE #6134: [unique: true] creates two unique indexes on a column
For some a reason column option `unique: true` creates two unique indexes on a column. How to prevent it? Please take a look at an `email` field:

```
const mysql = require('../getConnection').mysql; // configured and connected sequelize here

const modelAttrs = {
    fname: {
        type: mysql.Sequelize.STRING(255),
        field: 'fname',
        allowNull: true,
        defaultValue: null,
    },
    email: {
        type: mysql.Sequelize.STRING(254),
        field: 'email',
        allowNull: true,
        defaultValue: null,
        unique: true,      // <<== here an email column is unique
        validate: {
            isEmail: true,
        }
    },
    time_zone: {
        type: mysql.Sequelize.STRING(100),
        field: 'time_zone',
        allowNull: true,
        defaultValue: null,
    }
};

const modelOpts = {
    freezeTableName: true,
    tableName: 'UserProfile',
    timestamps: true,
}

const UserProfile = mysql.define('UserProfile', modelAttrs, modelOpts);

mysql.sync({ force: true });
```

And here is what is actually created, one unique key on `email` is redundant:

```
CREATE TABLE `userprofile` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `fname` varchar(255) DEFAULT NULL,
  `lname` varchar(255) DEFAULT NULL,
  `full_name` varchar(500) DEFAULT NULL,
  `email` varchar(254) DEFAULT NULL,
  `accept_ts` tinyint(1) NOT NULL DEFAULT '0',
  `time_zone` varchar(100) DEFAULT NULL,
  `createdAt` datetime NOT NULL,
  `updatedAt` datetime NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `email` (`email`), //      <<==  One
  UNIQUE KEY `UserProfile_email_unique` (`email`)  //     <<==  Two ??
) ENGINE=InnoDB DEFAULT CHARSET=utf8
```

Why does it happen? How to prevent it?

BridgeAR:
### Pull Request check-list
### Description of change

Create only one unique index instead of two

Fixes #6134

### DIFF
--- a/test/integration/instance.validations.test.js
+++ b/test/integration/instance.validations.test.js
@@ -6,6 +6,7 @@ var chai = require('chai')
   , expect = chai.expect
   , Sequelize = require(__dirname + '/../../index')
   , Support = require(__dirname + '/support')
+  , dialect = Support.getTestDialect()
   , config = require(__dirname + '/../config/config');
 
 describe(Support.getTestDialectTeaser('InstanceValidator'), function() {
@@ -99,7 +100,9 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), function() {
         }).then(function(err) {
           expect(err).to.be.an.instanceOf(Error);
           expect(err.errors).to.have.length(1);
-          expect(err.errors[0].path).to.include('uniqueName');
+          if (dialect !== 'mssql') {
+            expect(err.errors[0].path).to.include('uniqueName');
+          }
           expect(err.errors[0].message).to.include('must be unique');
         });
     });


BridgeAR:
I can't properly debug mssql

test/unit/model/sync.test.js
```diff
@@ -0,0 +1,113 @@
...
+var chai = require('chai')
+  , expect = chai.expect
+  , sinon = require('sinon')
+  , Support   = require(__dirname + '/../support')
+  , DataTypes = require('../../../lib/data-types')
+  , current   = Support.sequelize
+  , Promise = current.Promise;
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+  describe.only('sync', function () {
```
mickhansen: !

BridgeAR: Oops. Fixed

janmeier:
It think we can just skip this test for mssql https://github.com/sequelize/sequelize/blob/6177a95ddcc598fe60a0cbc12273fb33371de22c/test/integration/instance.validations.test.js#L82

You should be able to debug mssql just by running `npm run test-integration-mssql` (outside of docker) - that will hit out public mssql servers.

I don't think we should try to parse which field from this: 
```js
Violation of UNIQUE KEY constraint 'UQ__models__38751322B57C6E6B'. Cannot insert duplicate key in object 'dbo.models'. The duplicate key value is (unique name one).
```

Theoretically I guess we can infer the column from the value, but noooo....

BridgeAR:
I gave up looking into this. I do not know enough about mssql.

Something does not work with mssql and findOrCreate anymore after using only the normal unique index. The only real change I see is that the returned err.fields look like: `UQ__Users__387513223698177A` instead of `uniqueName`. It seems like there is a deadlock while running the find after the unique constraint error.

mccabemj:
Went to fix this issue #6134 and saw this PR. My mssql tests pass with that change to /lib/model.js. What was the issue?

janmeier:
@mccabemj I don't remember - You are welcome to open a new PR, perhaps it was a flaky test that has since been fixed / remove

PR #6215, opened 2016-07-04
Fix upsert failing on not null checks

### LINKED ISSUE #5711: Upsert should skip notNull validations
## What you are doing?

I am trying to partial update a row, after insertion.

``` js
const test = sequelize.define('test', {
    id: {
        primaryKey: true,
        type: Sequelize.INTEGER.UNSIGNED,
        field: 'my_id'
    },
    val1: {
        type: Sequelize.STRING(36),
        allowNull: false,
        field: 'value_1'
    },
    val2: {
        type: Sequelize.STRING(36),
        allowNull: false,
        field: 'value_2'
    }
});
test.sync().then(function () {
    test.upsert({
        id: 1,
        val1: 'hello',
        val2: 'world'
    }).then(function () {

        // I was expecting this to partially update the row
        test.upsert({
            id: 1,
            val2: 'world222222'
        }).then(function () {
            console.log('Expected')
        }).catch(function (err) {
            // That's what really happens
            console.error(err);
        });
    });
});
```

I want val2 only to be updated

I am getting an exception SequelizeValidationError: notNull Violation: val1 cannot be null

Dialect: mysql
Database version: 5.6.24
Sequelize version: 3.21.0
node version: 5.7.0

BridgeAR:
### Pull Request check-list

### Description of change

Fixes #5711

### DIFF
--- a/test/unit/model/upsert.test.js
+++ b/test/unit/model/upsert.test.js
@@ -1,0 +1,40 @@
+'use strict';
+
+/* jshint -W030 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../support')
+  , current = Support.sequelize
+  , sinon = require('sinon')
+  , Promise = current.Promise
+  , DataTypes = require('../../../lib/data-types');
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+
+  if (current.dialect.supports.upserts) {
+    describe('method upsert', function () {
+      var User = current.define('User', {
+        name: DataTypes.STRING,
+        secretValue: {
+          type: DataTypes.INTEGER,
+          allowNull: false
+        }
+      });
+
+      before(function () {
+        this.query = current.query;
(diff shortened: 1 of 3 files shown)


test/unit/model/upsert.test.js
```diff
@@ -0,0 +1,39 @@
...
+      });
+
+      after(function () {
+        current.query = this.query;
+      });
+
+      it('skip validations for missing fields', function() {
+        return User.upsert({
+          name: 'Grumpy Cat'
+        });
```
janmeier: No assertion? Mayhaps stub out the validator?

BridgeAR: Well the validations are run and the test would fail without the change. And That's what I test for.

janmeier: True, its just not immediately obvious to see - maybe `expect(..).not.to.be.rejectedWith(current.ValidationError)`

BridgeAR:
Fixed & rebased

PR #6216, opened 2016-07-04
Fix #5788, passed sync options.logging to query directly

### LINKED ISSUE #5788: .sync() options.logging=console.log
The method `Sequelize.prototype.sync` says "options.logging=console.log"(Line 967 of the file "lib/sequelize.js"), however, it's default value is `false`  due to  line 978:

```js
options.logging = options.logging === undefined ? false : options.logging;
```

The method `Sequelize.prototype.query` fallbacks to `this.options.logging`,  method `truncate` and some other methods by contrast do not test `options.logging`. So which should be chosen here?

sushantdhiman:
### Pull Request check-list

### Description of change

Closes #5788, Passed `options.logging` directly to query so it can respect global `logging` options.


BridgeAR:
LGTM but it would be nice to have test for it too.

sushantdhiman:
I have just removed an obsolete code section so it matched all other surface method `findAll` , `update` and `save`. If we introduced test for `sync`, we should also test for other surface methods too.

PR #6226, opened 2016-07-05
Update jsdocs for doclets

janmeier:
I've been working on replacing our homebrewn .md file generation with an automated service - This is the result https://doclets.io/sequelize/sequelize/doclets. It still needs a bit of touching up, but personally I think it looks better than what we have currently, and I'd love to have a system where docs are automagically deployed


lib/errors.js
```diff
@@ -264,7 +242,8 @@ exports.ConnectionRefusedError = ConnectionRefusedError;
 /**
  * Thrown when a connection to a database is refused due to insufficient privileges
  * @extends ConnectionError
- * @constructor
+ * @memberof Errors
+ * @memberof Sequelize
```
BridgeAR: This is the only Error class that is also declared as `memberof Sequelize`. I thought all Error classes are memberof Sequelize?

lib/associations/belongs-to.js
```diff
@@ -78,19 +79,21 @@ class BelongsTo extends Association {
        * @param {Object} [options]
        * @param {String|Boolean} [options.scope] Apply a scope on the related model, or remove its default scope by passing false.
        * @param {String} [options.schema] Apply a schema on the related model
-       * @see {Model#findOne} for a full explanation of options
-       * @return {Promise<Instance>}
+       * @see {@link Model.findOne} for a full explanation of options
```
BridgeAR: This is the only place where you switched from # to a dot. Was that intended?

lib/errors.js
janmeier: Yeah, I still need to figure out how we document how to access the errors - I think it makes sense to have them in a separate namespace, but on the other hand I still want to document that you can access them as  `sequelize.HostNotFoundError` etc.

lib/associations/belongs-to.js
janmeier: Probably not - I think jsdoc has some convention about one being instance and the other being static members, but I'm not sure

mickhansen: Ideally i'd like something like 'require('sequelize/errors');`

felixfbecker: `Model#findOne` is the same as `Model.prototype.findOne`

lib/data-types.js
```diff
@@ -82,9 +82,8 @@ ABSTRACT.prototype.stringify = function stringify(value, options) {
 /**
  * A variable length string. Default length 255
  *
- * Available properties: `BINARY`
- *
- * @property STRING
+ * @property BINARY
```
felixfbecker: I find `@property` a bit misleading. According to JSDoc-docs, all we need is `@memberof`
http://usejsdoc.org/tags-namespace.html

lib/dialects/abstract/query-generator.js
```diff
@@ -85,90 +57,19 @@ const QueryGenerator = {
...
-      - attrNameAfter: The name of the attribute, after renaming.
-  */
-  /* istanbul ignore next */
-  renameColumnQuery(tableName, attrNameBefore, attrNameAfter) {
-    throwMethodUndefined('renameColumnQuery');
-  },
-
-  /*
     Returns an insert into command. Parameters: table name + hash of attribute-value-pairs.
+   @private
```
felixfbecker: 👍 for having more documentation of private methods, but marking them with `@private`

lib/errors.js
felixfbecker: Imo it should be enough to note that they are availbable in the Errors namespace description. Otherwise we would clutter up the properties.

lib/model.js
```diff
@@ -1429,13 +1429,14 @@ class Model {
...
+   *
    * @param  {Object}                    [options] A hash of options to describe the scope of the search
    * @param  {Transaction}               [options.transaction] Transaction to run query under
    * @param  {String}                    [options.searchPath=DEFAULT] An optional parameter to specify the schema search_path (Postgres only)
    *
-   * @see {Model#findAll}           for an explanation of options
-   * @return {Promise<Instance>}
-   * @alias find
+   * @see {@link Model#findAll}           for an explanation of options
+   * @return {Promise<Model>}
```
felixfbecker: JSDoc officially has the `@alias` tag: http://usejsdoc.org/tags-alias.html
I find it much nicer than defining it in the description. If it is not supported by doclets, maybe we can ask them to implement it?

lib/sequelize.js
```diff
@@ -28,6 +28,7 @@ const _ = require('lodash');
  * In addition to sequelize, the connection library for the dialect you want to use should also be installed in your project. You don't need to import it however, as sequelize will take care of that.
  *
  * @class Sequelize
+ * @lands Errors
```
felixfbecker: I think what you meant was `@lends` http://usejsdoc.org/tags-lends.html
But I don't think it's the right tag for what you intended, it should be `@mixes Errors` http://usejsdoc.org/tags-mixes.html

felixfbecker:
@janmeier This is such a great work. Can we get this rebased and merged?

lib/model.js
janmeier: >  treat all references to a member as if the member had a different name

Which translates to only showing the alias - In this case we want to show both

felixfbecker:
This LGTM, shall we merge?

janmeier:
We shall, thanks for the review felix :)

Now I just need to convert the rest of the docs articles, and maybe fix up the styling a bit

PR #6230, opened 2016-07-05
Added information about  operator $eq

Ginden:
It's minor update of documentation - operator `$eq` is now documented.

### DIFF
--- a/docs/docs/querying.md
+++ b/docs/docs/querying.md
@@ -123,6 +123,7 @@ $gte: 6,               // >= 6
 $lt: 10,               // < 10
 $lte: 10,              // <= 10
 $ne: 20,               // != 20
+$eq: 3,                // = 3
 $not: true,            // IS NOT TRUE
 $between: [6, 10],     // BETWEEN 6 AND 10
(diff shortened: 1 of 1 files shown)


docs/docs/querying.md
```diff
@@ -123,6 +123,7 @@ $gte: 6,               // >= 6
 $lt: 10,               // < 10
 $lte: 10,              // <= 10
 $ne: 20,               // != 20
+$eq: 3,                // = 3
```
janmeier: This might lead som users to believe that `$eq` is required for regular equals. Perhaps show a couple of examples from https://github.com/sequelize/sequelize/blob/cd2e38e48e3054ba7e9661181f74d4ffdae65fe4/test/unit/sql/where.test.js

Ginden: > This might lead som users to believe that $eq is required for regular equals. 

They are already lost, because "Basics" examples prove it isn't.

janmeier: True dat :)

janmeier:
Thanks

PR #6243, opened 2016-07-07
Fix unique indexes being added to the UPDATE's WHERE clause while doing an upsert

### LINKED ISSUE #6240: UPSERT adding `OR "unique indexes"` to update's where clause
While using `upsert` on a table with unique indexes, I found that the `UPDATE` part of the generated query is not what I expected.

Quick demonstration:
`Demo` model:

``` js
"use strict";

module.exports = function (sequelize, DataTypes) {
    let Demo = sequelize.define("Demo", {
        id      : {
            type         : DataTypes.STRING,
            field        : "id",
            primaryKey   : true,
        },
        unique1 : {
            type  : DataTypes.STRING,
            field : "unique1",
        },
        unique2 : {
            type  : DataTypes.STRING,
            field : "unique2",
        },
    }, {
        indexes : [{
            unique : true,
            fields : ["unique1", "unique2"],
            where  : {
                unique2 : {$ne : null},
            },
        }, {
            unique : true,
            fields : ["unique1"],
            where  : {
                unique2 : null,
            },
        }],
    });

    return Demo;
};
```

Do an upsert:

``` js
let Seq = require("sequelize"),
    db  = new Seq("demo", "demo", "demo", {});  // Fill with your settings

db.Demo = db.import("./Demo.js");
db.sync()
.then(() => {
    db.Demo.destroy({truncate : true});
    db.Demo.bulkCreate([{
        id      : "1",
        unique1 : "1",
        unique2 : "1",
    }, {
        id      : "2",
        unique1 : "2",
        unique2 : "2",
    }]);

    db.Demo.upsert({
        id      : "1",
        unique1 : "1",
        unique2 : "2",
    });
});
```

Queries:

``` sql
Executing (default): TRUNCATE "Demos"
Executing (default): INSERT INTO "Demos" ("id","unique1","unique2","createdAt","updatedAt") VALUES ('1','1','1','2016-07-07 10:47:32.508 +00:00','2016-07-07 10:47:32.508 +00:00'),('2','2','2','2016-07-07 10:47:32.508 +00:00','2016-07-07 10:47:32.508 +00:00');
Executing (default): CREATE OR REPLACE FUNCTION pg_temp.sequelize_upsert() RETURNS integer AS $func$ BEGIN INSERT INTO "Demos" ("id","unique1","unique2","createdAt","updatedAt") VALUES ('1','2','2','2016-07-07 10:47:32.523 +00:00','2016-07-07 10:47:32.523 +00:00'); RETURN 1; EXCEPTION WHEN unique_violation THEN UPDATE "Demos" SET "id"='1',"unique1"='2',"unique2"='2',"updatedAt"='2016-07-07 10:47:32.523 +00:00' WHERE ("id" = '1' OR ("unique1" = '2' AND "unique2" = '2') OR "unique1" = '2'); RETURN 2; END; $func$ LANGUAGE plpgsql; SELECT * FROM pg_temp.sequelize_upsert();
```

My issue is with: `UPDATE (...) WHERE ("id" = '1' OR ("unique1" = '1' AND "unique2" = '2') OR "unique1" = '1')`
The [upsert function](https://github.com/sequelize/sequelize/blob/v3.23.4/lib/query-interface.js#L508) is adding the unique key fields to the WHERE clause even though the PK is already present, which doesn't make a lot of sense to me. Even worst, it's doing it using `OR` which might cause unwanted behavior (I know at first glance using unique indexes sounds fine but unique indexes can be nullable and have where clauses, so it's not so straightforward).
Furthermore, the model's `index` field [allows a `where` clause](http://docs.sequelizejs.com/en/latest/docs/models-definition/#indexes) which is not being taken into account while adding the unique keys to the WHERE clause.

What I would expect/suggest: 
- When PK is present, use only the PK on the WHERE clause;
- I would agree/understand the use of unique indexes as PK if the PK is not available (i.e. either if the table doesn't have one or if it was not provided to the upsert);
  - If using unique indexes as PK, it should at least take into account it's own where clause.

**Dialect:** any
**Sequelize version:** 3.23.4

ke1Del:
Partially fixes #6240 : if UPSERT has PK, only filter UPDATE by PK. Else add the unique indexes logic.

I did not fix the index's where clause not being considered as it seems to me that this logic should not exist. Performing an upsert without specifying any value for the PK seems like a dangerous operation.

PS: I also fixed the Dockerfile that was running `npm install` on the wrong folder and then using the shared node_modules folder. Still `npm run test-docker` is producing a couple of MySQL errors on the v3 branch prior to my changes so I wasn't able to confirm that all tests pass.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,8 @@ FROM node:5
 
 RUN apt-get install libpq-dev
 
-WORKDIR /sequelize
-COPY package.json /sequelize/
+COPY package.json /
 RUN npm install
 
-COPY . /sequelize/
+WORKDIR /sequelize
+VOLUME /sequelize
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,8 @@ FROM node:5
 
 RUN apt-get install libpq-dev
 
-COPY package.json /
+WORKDIR /sequelize
+COPY package.json /sequelize/
 RUN npm install
 
-WORKDIR /sequelize
-VOLUME /sequelize
+COPY . /sequelize/
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,8 +3,6 @@ sequelize:
   links:
    - mysql
    - postgres
-  volumes:
-    - .:/sequelize
   environment:
     SEQ_DB: sequelize_test
     SEQ_USER: sequelize_test
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,8 @@ sequelize:
   links:
    - mysql
    - postgres
+  volumes:
+    - .:/sequelize
   environment:
     SEQ_DB: sequelize_test
     SEQ_USER: sequelize_test
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,5 @@
 # Future
+- [FIXED] Fixed unique indexes being added to the UPDATE's WHERE clause while doing an upsert. [#6240](https://github.com/sequelize/sequelize/issues/6240)
 - [FIXED] Fixed an issue where custom-named model fields break when offsetting, ordering, and including hasMany simultaneously. [#5985](https://github.com/sequelize/sequelize/issues/5985)
 - [FIXED] Don't remove includes from count queries and unify findAndCount and count queries. [#6123](https://github.com/sequelize/sequelize/issues/6123)
 - [FIXED] `Model.count` don't include attributes [#5057](https://github.com/sequelize/sequelize/issues/5057)
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -513,7 +513,7 @@ QueryInterface.prototype.upsert = function(tableName, valuesByField, updateValue
 
   options = _.clone(options);
 
-  if (!Utils._.isEmpty(where)) {
+  if (!Utils._.chain(where).values().remove(null).isEmpty().value()) {
     wheres.push(where);
   } else {
     // Lets combine uniquekeys and indexes into one
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -515,35 +515,35 @@ QueryInterface.prototype.upsert = function(tableName, valuesByField, updateValue
 
   if (!Utils._.isEmpty(where)) {
     wheres.push(where);
+  } else {
+    // Lets combine uniquekeys and indexes into one
+    indexes = Utils._.map(model.options.uniqueKeys, function (value) {
+      return value.fields;
+    });
+
+    Utils._.each(model.options.indexes, function (value) {
+      if (value.unique === true) {
+        // fields in the index may both the strings or objects with an attribute property - lets sanitize that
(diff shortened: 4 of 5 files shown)


janmeier:
Please add a test - perhaps just a unit test that stubs out `this.QueryGenerator.upsertQuery` and asserts that its called with the correct `where

ke1Del:
Tests were added.
I'm not sure why Travis failed but it doesn't seem related to the changes made.

docker-compose.yml
```diff
@@ -3,8 +3,6 @@ sequelize:
   links:
    - mysql
    - postgres
-  volumes:
```
janmeier: I think volumes is used here to mount the local files so you don't have to rebuild the container each time a file changes - But @mickhansen knows more about that

mickhansen: Should definitely not be removed.

Dockerfile
```diff
@@ -2,8 +2,8 @@ FROM node:5
...
 RUN apt-get install libpq-dev
 
-COPY package.json /
+WORKDIR /sequelize
+COPY package.json /sequelize/
 RUN npm install
 
-WORKDIR /sequelize
-VOLUME /sequelize
+COPY . /sequelize/
```
mickhansen: please don't edit dev environment files when you know for a fact that a lot of other people rely on them :)

mickhansen:
Changes to Dockerfile and docker-compose.yml need to be removed completely.

Dockerfile
ke1Del: @mickhansen Sure, I can revert the changes if you want. But I have to warn you that the `.dockerignore` file only applies to the `COPY` operation and not to the `VOLUME`.
Also, you're performing an `npm install` on the `/` (root) folder and then you mount the volume on the `/sequelize` folder, which means the `npm install` is completely useless.
Anyway, the reason I changed it is because I had some package conflicts due to my node_modules folder being compiled with node v6 (which I use on my workstation) and docker_compose using node v5 (I don't remember exactly which module did not like that and needed to be rebuilt).

janmeier: Probably sqlite.

If we can fix the issue of having to have the same node version inside and outside of the container that would be nice :). As long as we can still mount /lib and /test as volumes so we dont need a rebuild when changing files

ke1Del: You can try to create different volumes for each folder but it doesn't look very efficient (you'll probably also need the `.jshintrc` file for the lint test, I don't know about the rest of the files).
A copy operation is actually very fast so rebuilding the container only takes a few seconds. But it's really your choice, I just thought to give a hand as I had some experience with docker containers in the past and it didn't look like the dockerfile was doing what you expected.

ke1Del:
I reverted the changes to the Docker files.
Regarding the failed tests, it doesn't look to me like it's related to the changes I've made. Could you please confirm and let me know if I introduced some kind of regression? Thanks

yahao87:
I need this feature.
When is it officially merged?

PR #6251, opened 2016-07-11
Primary key should be same as unique key (doc change) 

ivesdebruycker:
If not, a primary key for taggable_id is created, so no more than one tag per taggable could be created.


mickhansen:
Actually, it should probably have a defined `id` instead.

Issue #6178, opened 2016-06-27
Foreign key being pulled from incorrect schema.


emmceemoore:
## What you are doing?

I have an existing database and I wanted to spin up a new database to try something out. I updated my sequelize config to point to a different database and I ran my migrations to get the tables created.
## What do you expect to happen?

I expected the migrations to run without issue.
## What is actually happening?

One of the migrations fails. The failure occurs because [`getForeignKeyQuery()`](https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/query-generator.js#L315) returns the name of a foreign key in a _different_ schema.

Here's the error I'm getting:

``` bash
SequelizeDatabaseError: ER_CANT_DROP_FIELD_OR_KEY: Can't DROP 'fooId_foreign_idx'; check that column/key exists
```

(The real name of the key is `foo_ibfk_1`. I'm not sure if the difference in name is because of a newer version of sequelize or a newer version of mysql.)

Regardless, it seems like it would be a good idea to additionally consider the schema when performing this query. (Similar to how [`getForeignKeysQuery()`](https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/query-generator.js#L303) does it.)

**Dialect:** mysql
**Database version:** 5.7.13
**Sequelize version:** 3.23.4

myaskevich:
Still relevant

-- 
Maxim Yaskevich

PR #6264, opened 2016-07-12
included correct database schema for mysql removeColumn query

paolord:
### Pull Request check-list
### Description of change

attempt to fix issue [#6178](https://github.com/sequelize/sequelize/issues/6178) by providing the correct database schema to getForeignKeyQuery during removeColumn queries in MySQL. Did not add new tests but changed the query-interface test to not check if mysql dialect can remove a column without a schema and with a foreign key.

### DIFF
--- a/lib/dialects/mysql/query-interface.js
+++ b/lib/dialects/mysql/query-interface.js
@@ -21,10 +21,18 @@ const _ = require('lodash');
  */
 function removeColumn(tableName, columnName, options) {
   options = options || {};
+  var tableNameSchema = tableName;
+  /* jshint validthis:true */
+  if(this.sequelize.config.database){
+    tableNameSchema = {
+      tableName: tableName,
+      schema: this.sequelize.config.database
+    };
+  }
 
   /* jshint validthis:true */
   return this.sequelize.query(
-      this.QueryGenerator.getForeignKeyQuery(tableName, columnName),
+      this.QueryGenerator.getForeignKeyQuery(tableNameSchema, columnName),
       _.assign({ raw: true }, options)
     )
     .spread(results => {
--- a/lib/dialects/mysql/query-interface.js
+++ b/lib/dialects/mysql/query-interface.js
@@ -23,7 +23,7 @@ function removeColumn(tableName, columnName, options) {
   options = options || {};
   var tableNameSchema = tableName;
   /* jshint validthis:true */
-  if(this.sequelize.config.database){
+  if (this.sequelize.config.database) {
     tableNameSchema = {
       tableName: tableName,
       schema: this.sequelize.config.database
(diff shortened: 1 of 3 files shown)


test/integration/query-interface.test.js
```diff
@@ -586,14 +586,15 @@ describe(Support.getTestDialectTeaser('QueryInterface'), function() {
           expect(table).to.not.have.property('lastName');
         });
       });
-
-      it('should be able to remove a column with a foreign key constraint', function() {
-        return this.queryInterface.removeColumn('users', 'manager').bind(this).then(function() {
-          return this.queryInterface.describeTable('users');
-        }).then(function(table) {
+      if(dialect !== 'mysql'){
```
sushantdhiman: Why we aren't running this test for `mysql` now ?

lib/dialects/mysql/query-interface.js
```diff
@@ -21,10 +21,18 @@ const _ = require('lodash');
  */
 function removeColumn(tableName, columnName, options) {
   options = options || {};
+  var tableNameSchema = tableName;
+  /* jshint validthis:true */
+  if(this.sequelize.config.database){
```
sushantdhiman: spaces ``if (this.sequelize.config.database) {```

test/integration/query-interface.test.js
paolord: Without including a schema to getForeignKeyQuery, it can mistakenly return a foreign key from another database with the same column name and table name.  This happens when removeColumn is called in MySQL dialect, since mysql can't really drop columns if it has foreign key constraints. That test is explicitly for testing removeColumn without a schema.

sushantdhiman: I think we should add a test to cover schema case, test that with describing schema foreign key is removed

paolord: Thanks for the feedback, you mean a test for removeColumn with a schema for all dialects?

lib/dialects/mysql/query-interface.js
```diff
@@ -21,10 +21,18 @@ const _ = require('lodash');
...
+    tableNameSchema = {
+      tableName: tableName,
+      schema: this.sequelize.config.database
+    };
+  }
 
   /* jshint validthis:true */
   return this.sequelize.query(
-      this.QueryGenerator.getForeignKeyQuery(tableName, columnName),
+      this.QueryGenerator.getForeignKeyQuery(tableNameSchema, columnName),
```
mickhansen: Use the `quoteTable/extractTable` logic that all other calls use.

test/integration/query-interface.test.js
sushantdhiman: Yes, lets add removeColumn test with schema in general. Other dialects may have same issue.

janmeier:

myaskevich:
😢

PR #6267, opened 2016-07-13
Bugfix/order clause in subquery 

fogine:
### Pull Request check-list

Note: because I use `postgres v9.3` few test cases (14) does not pass because of unsupported new features which were introduces in postgres v9.4 like (jsonb, type geometry)... I expect the tests to work including those which are using new features of postgres v9.4. Although I did not run tests wit postgres v9.4 installed.

### Description of change


sushantdhiman:
Bad merge, you will need to rebase :)

sushantdhiman:
Also is this intended to merge with `v3` or `master`, It contain commits from `v3` that we will not be merged back to `master`.

You need to create PR with correct `base` and `target` branches

fogine:
> Also is this intended to merge with v3 or master, It contain commits from v3 that we will not be merged back to master.

It's intended to merge with `v3`

sushantdhiman:
Looks like its being merged to `master` atm, change the `base` branch

PR #6268, opened 2016-07-13
Bugfix/order clause in subquery - for the `v3`

fogine:
### Pull Request check-list

Note: because I use `postgres v9.3` few test cases (14) does not pass because of unsupported new features which were introduces in postgres v9.4 like (jsonb, type geometry)... I expect the tests to work including those which are using new features of postgres v9.4. Although I did not run tests wit postgres v9.4 installed.

### Description of change

### DIFF
--- a/test/unit/sql/select.test.js
+++ b/test/unit/sql/select.test.js
@@ -234,6 +234,27 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
         +') AS [user] LEFT OUTER JOIN [post] AS [POSTS] ON [user].[id] = [POSTS].[user_id];'
       });
 
+      testsql({
+        table: User.getTableName(),
+        model: User,
+        include: include,
+        attributes: [
+          ['id_user', 'id'],
+          'email',
+          ['first_name', 'firstName'],
+          ['last_name', 'lastName']
+        ],
+        order: '`user`.`last_name` ASC',
+        limit: 30,
+        offset: 10,
+        subQuery: true
+      }, {
+          default: 'SELECT [user].*, [POSTS].[id] AS [POSTS.id], [POSTS].[title] AS [POSTS.title] FROM (' +
+                       'SELECT [user].[id_user] AS [id], [user].[email], [user].[first_name] AS [firstName], [user].[last_name] AS [lastName] FROM [users] AS [user] ORDER BY [user].[last_name] ASC' +
+                       sql.addLimitAndOffset({ limit: 30, offset:10, order: '`user`.`last_name` ASC' }) +
+                   ') AS [user] LEFT OUTER JOIN [post] AS [POSTS] ON [user].[id_user] = [POSTS].[user_id] ORDER BY [user].[last_name] ASC;'
+      });
+
       var nestedInclude = Model.$validateIncludedElements({
         include: [{
           attributes: ['title'],
--- a/test/unit/sql/select.test.js
+++ b/test/unit/sql/select.test.js
@@ -244,7 +244,7 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
           ['first_name', 'firstName'],
           ['last_name', 'lastName']
         ],
-        order: '`user`.`last_name` ASC',
+        order: '[user].[last_name] ASC'.replace(/\[/g, Support.sequelize.dialect.TICK_CHAR_LEFT).replace(/\]/g, Support.sequelize.dialect.TICK_CHAR_RIGHT),
         limit: 30,
         offset: 10,
         subQuery: true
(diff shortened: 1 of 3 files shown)


fogine:
Hmm, provided test case for the bugfix fails only for `DIALECT=mssql`. It fails because additional unexpected `id_user` field is added to `order` clause in subquery. Is it forced by default for `mssql` dialect so it always orders subquery by its primary key? I will try to look more into it tomorrow but for now it does not make much sense to me.

fogine:
The issue I was experiencing is somewhat related to #6162 . `mssql` dialect as the only dialect is looking for `hasMultiAssociation` option in `options` object [here](https://github.com/sequelize/sequelize/blob/v3/lib/dialects/mssql/query-generator.js#L680). if the option is not set, it will additionally order subquery by its primary key as if no specific `order` option was not set.
The `travis-ci` build should successful now.

fogine:
Hello @sushantdhiman ,
any thoughts please? Do you intend to merge the PR or does it need further investigation?

sushantdhiman:
LGTM, Can you also submit a PR to `master` branch as well. Will need review from @mickhansen @janmeier

fogine:
Sure, I will make another PR for the `master` branch.

fogine:
Just Note: I was forced to `rebase` the commits I made in this pull request because the commits were made with incorrect author's email.

felixfbecker:
@fogine the author email is still not correct (github link is broken)

fogine:
Thanks for the notification @felixfbecker , but I'm not sure what you mean?
Commits now says: `fogine committed with fogine`. The second note of fogine is link to correct account.
I don't know why github is keeping reference to the original author's name (it even the same name).
If you do `git log` on the branch I get correct author's email and username.

Could you maybe describe it more in deep?

felixfbecker:
Okay, "fogine commited with fogine" is pretty weird though..

fogine:
:) yep it got me confused as well. It's because the first `fogine` is reference to original author's name which is same. Only the email has changed.

felixfbecker:
It's because rebase is not sufficient to _change_ the author, it will keep the original author. You need to use git filter-branch.

fogine:
Thanks @felixfbecker ,
you were right. Apparently interactive rebase with `git commit --amend --author="..."` does not work as I expected.

felixfbecker:
Needs rebase

test/unit/sql/select.test.js
```diff
@@ -158,6 +158,28 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
...
+        order: '[user].[last_name] ASC'.replace(/\[/g, Support.sequelize.dialect.TICK_CHAR_LEFT).replace(/\]/g, Support.sequelize.dialect.TICK_CHAR_RIGHT),
+        limit: 30,
+        offset: 10,
+        hasMultiAssociation: true,//must be set only for mssql dialect here
+        subQuery: true
+      }, {
+          default: 'SELECT [user].*, [POSTS].[id] AS [POSTS.id], [POSTS].[title] AS [POSTS.title] FROM (' +
+                       'SELECT [user].[id_user] AS [id], [user].[email], [user].[first_name] AS [firstName], [user].[last_name] AS [lastName] FROM [users] AS [user] ORDER BY [user].[last_name] ASC' +
+                       sql.addLimitAndOffset({ limit: 30, offset:10, order: '`user`.`last_name` ASC' }) +
+                   ') AS [user] LEFT OUTER JOIN [post] AS [POSTS] ON [user].[id_user] = [POSTS].[user_id] ORDER BY [user].[last_name] ASC;'
```
felixfbecker: please indent this properly, eslint will complain otherwise

fogine:
All tests passing for me locally...

fogine:
ping @felixfbecker 
ping @sushantdhiman

felixfbecker:
We switched our MSSQL tests to AppVeyor, you probably based on an old commit. Could you rebase? Also, is there a PR for master? We need to merge master first to avoid that master lacks behind v3 in features.

fogine:
@felixfbecker , I'll try to rebase once more (I've already rebased once).
the #6313 PR for master has been already successfully merged.

fogine:
@felixfbecker , the branch is already up to date with `v3` branch.
It seems that the build exceeded some defined timeout... could you restart the build?

felixfbecker:
I don't think so - did you actually reset v3 to the upstream v3 before rebasing?

felixfbecker:
If you click "browse repo at this commit" for your first commit you will see that it does not include an appveyor.yml, and that your first commit's parent is not the latest v3 commit.

fogine:
My bad. Sorry.

PR #6279, opened 2016-07-15
Update postgres query interface

turbofoxwave:
### Pull Request check-list

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)
I have 1 failing testing on my system with the current master branch before adding my changes 
Failing test:
[POSTGRES] DataTypes should parse an empty GEOMETRY field:
     SequelizeDatabaseError: Too few ordinates in GeoJSON
All others are passing.

no

- [x] Have you added new tests to prevent regressions? yes
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?  na
- [x] Have you added an entry under `Future` in the changelog? yes

### Description of change
Changes are around the postgres query generator to address some issues I found when attempting to use dropFunction and createFunction. The root cause of the issue was due to the function expandFunctionParamList. Along the way I also found an issue with dropFunction in which the generated sql lacked a space after the keyword 'FUNCTION' ex: "FUNCTION myfunc" vs "FUNCTIONmyfunc".  I've also added tests for createFunction and dropFunction and I've stubbed out some tests for renameFunction.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+# Future
+- [FIXED] Issues with `createFunction` and `dropFunction` (PostgresSQL)
+
+
 # 4.0.0-2
 - [ADDED] include now supports string as an argument (on top of model/association), string will expand into an association matched literally from Model.associations
 - [FIXED] Accept dates as string while using `typeValidation` [#6453](https://github.com/sequelize/sequelize/issues/6453)
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -530,6 +530,7 @@ const QueryGenerator = {
   },
 
   createFunction(functionName, params, returnType, language, body, options) {
+    if (!functionName || !returnType || !language || !body) throw new Error('createFunction missing some parameters. Did you pass functionName, returnType, language and body?');
 
     const paramList = this.expandFunctionParamList(params);
     const indentedBody = body.replace('\n', '\n\t');
@@ -544,9 +545,10 @@ const QueryGenerator = {
   },
 
   dropFunction(functionName, params) {
+    if (!functionName) throw new Error('requires functionName');
     // RESTRICT is (currently, as of 9.2) default but we'll be explicit
     const paramList = this.expandFunctionParamList(params);
-    return `DROP FUNCTION${functionName}(${paramList}) RESTRICT;`;
+    return `DROP FUNCTION ${functionName}(${paramList}) RESTRICT;`;
   },
 
   renameFunction(oldFunctionName, params, newFunctionName) {
@@ -575,19 +577,25 @@ const QueryGenerator = {
       throw new Error('expandFunctionParamList: function parameters array required, including an empty one for no arguments');
     }
 
-    const paramList = Utils._.each(params, curParam => {
+    const paramList = [];
+    Utils._.each(params, curParam => {
       const paramDef = [];
       if (Utils._.has(curParam, 'type')) {
         if (Utils._.has(curParam, 'direction')) { paramDef.push(curParam.direction); }
         if (Utils._.has(curParam, 'name')) { paramDef.push(curParam.name); }
         paramDef.push(curParam.type);
       } else {
-        throw new Error('createFunction called with a parameter with no type');
+        throw new Error('function or trigger used with a parameter without any type');
       }
-      return paramDef.join(' ');
+
+      var joined = paramDef.join(' ');
+      if (joined) paramList.push(joined);
+
     });
+
     return paramList.join(', ');
   },
+  
 
   expandOptions(options) {
     return Utils._.isUndefined(options) || Utils._.isEmpty(options) ?
--- a/test/integration/dialects/postgres/query-interface.test.js
+++ b/test/integration/dialects/postgres/query-interface.test.js
@@ -115,7 +115,6 @@ if (dialect.match(/^postgres/)) {
       });
 
       it('can drop a function', function () {
-        
         //call drop function
         return expect(this.queryInterface.dropFunction('droptest',[{type:'varchar',name:'test'}])
         //now call the function we attempted to drop.. if dropFunction worked as expect it should produce an error.
@@ -128,7 +127,6 @@ if (dialect.match(/^postgres/)) {
       });
 
       it('produces an error when missing expected parameters', function () {
-       
         return Promise.all([
           expect( () => {
             return this.queryInterface.dropFunction(); 
--- a/test/integration/dialects/postgres/query-interface.test.js
+++ b/test/integration/dialects/postgres/query-interface.test.js
@@ -6,13 +6,143 @@ var chai = require('chai')
   , Support = require(__dirname + '/../../support')
   , dialect = Support.getTestDialect()
   , DataTypes = require(__dirname + '/../../../../lib/data-types')
+  , Sequelize = require('../../../../index')
+  , Promise = Sequelize.Promise
   , _ = require('lodash');
+
 
 if (dialect.match(/^postgres/)) {
   describe('[POSTGRES Specific] QueryInterface', function () {
     beforeEach(function () {
       this.sequelize.options.quoteIdenifiers = true;
       this.queryInterface = this.sequelize.getQueryInterface();
+    });
+
+    describe('createFunction', function () {
+
+      beforeEach( function () {
+        //make sure we don't have a pre-existing function called create_job
+        //this is needed to cover the edge case of afterEach not getting called because of an unexpected issue or stopage with the 
+        //test suite causing a failure of afterEach's cleanup to be called.
+        return this.queryInterface.dropFunction('create_job',[{type:'varchar',name:'test'}])
+        //suppress errors here. if create_job doesn't exist thats ok.
+        .catch( err => {});
+      });
+
+      after( function () {
+        //cleanup
+        return this.queryInterface.dropFunction('create_job',[{type:'varchar',name:'test'}])
+        //suppress errors here. if create_job doesn't exist thats ok.
+        .catch( err => {});
+      });
+
+      it('creates a stored procedure', function () {
+        var body = 'return test;';
+        var options = {};
+
+        //make our call to create a function
+        return this.queryInterface.createFunction('create_job', [{type:'varchar',name:'test'}], 'varchar', 'plpgsql', body, options)
+        //validate
+        .then( () =>  {
+          return this.sequelize.query('select create_job(\'test\');', { type: this.sequelize.QueryTypes.SELECT });
+        })
+        .then( res => {
+          return expect(res[0].create_job).to.be.eql('test');
+        });
+      });
+
+      it('treats options as optional', function () {
+        var body = 'return test;';
+
+        //run with null options parameter
+        return this.queryInterface.createFunction('create_job', [{type:'varchar',name:'test'}], 'varchar', 'plpgsql', body, null)
+        //validate
+        .then( () => { 
+          return this.sequelize.query('select create_job(\'test\');', { type: this.sequelize.QueryTypes.SELECT });
+        })
+        .then( res => {
+          return expect(res[0].create_job).to.be.eql('test');
+        });
+      });
+
+      it('produces an error when missing expected parameters', function () {
+        var body = 'return 1;';
+        var options = {};
+        
+        return Promise.all([
+          //requires functionName  
+          expect( () => {
+            return this.queryInterface.createFunction(null, [{name:'test'}], 'integer', 'plpgsql', body, options);
+          }).to.throw(/createFunction missing some parameters. Did you pass functionName, returnType, language and body/)
+        
+          //requires Parameters array
+          ,expect( () => {
+            return this.queryInterface.createFunction('create_job',null, 'integer', 'plpgsql', body, options);
+          }).to.throw(/function parameters array required/)
+          
+          //requires returnType
+          ,expect( () => {
+            return this.queryInterface.createFunction('create_job', [{type:'varchar',name:'test'}], null, 'plpgsql', body, options);
+          }).to.throw(/createFunction missing some parameters. Did you pass functionName, returnType, language and body/)
+        
+          //requires type in parameter array
+          ,expect( () => {
+            return this.queryInterface.createFunction('create_job', [{name:'test'}], 'integer', 'plpgsql', body, options);
+          }).to.throw(/function or trigger used with a parameter without any type/)
+          
+          //requires language
+          ,expect( () => {
+            return this.queryInterface.createFunction('create_job', [{type:'varchar',name:'test'}], 'varchar', null, body, options);
+          }).to.throw(/createFunction missing some parameters. Did you pass functionName, returnType, language and body/)
+
+          //requires body
+          ,expect( () => {
+            return this.queryInterface.createFunction('create_job', [{type:'varchar',name:'test'}], 'varchar', 'plpgsql', null, options);
+          }).to.throw(/createFunction missing some parameters. Did you pass functionName, returnType, language and body/)
+        ]);
+      });
+    });
+
+    describe('dropFunction',function () {
+      beforeEach( function () {
+        var body = 'return test;';
+        var options = {};
+
+        //make sure we have a droptest function in place.
+        return this.queryInterface.createFunction('droptest', [{type:'varchar',name:'test'}], 'varchar', 'plpgsql', body, options)
+        //suppress errors.. this could fail if the function is already there.. thats ok. 
+        .catch( function (err) { });
+      });
+
+      it('can drop a function', function () {
+        
+        //call drop function
+        return expect(this.queryInterface.dropFunction('droptest',[{type:'varchar',name:'test'}])
+        //now call the function we attempted to drop.. if dropFunction worked as expect it should produce an error.
+        .then( () => {
+          // call the function we attempted to drop.. if it is still there then throw an error informing that the expected behavior is not met. 
+          return this.sequelize.query('select droptest(\'test\');', { type: this.sequelize.QueryTypes.SELECT });  
+        }))
+        //test that we did get the expected error indicating that droptest was properly removed.
+        .to.be.rejectedWith(/.*function droptest.* does not exist/);
+      });
+
+      it('produces an error when missing expected parameters', function () {
+       
+        return Promise.all([
+          expect( () => {
+            return this.queryInterface.dropFunction(); 
+          }).to.throw(/.*requires functionName/)
+          ,
+          expect( () => {
+            return this.queryInterface.dropFunction('droptest'); 
+          }).to.throw(/.*function parameters array required/)
+          ,
+          expect( () => { 
+            return this.queryInterface.dropFunction('droptest', [{name:'test'}]);
+          }).to.be.throw(/.*function or trigger used with a parameter without any type/)
+        ]);
+      });
     });
 
     describe('indexes', function () {


lib/dialects/postgres/query-generator.js
```diff
@@ -524,6 +524,10 @@ const QueryGenerator = {
   },
 
   createFunction(functionName, params, returnType, language, body, options) {
+    if(!returnType) throw new Error("requires returnType");
```
sushantdhiman: Spaces around if condition something like ``if (!returnType) {``, I feel you can combine all three `returnType` , `language` and `body` checks and throw a generic error message like "createFunction missing some parameters, did you forget to pass returnType, language or body."

lib/dialects/postgres/query-generator.js
```diff
@@ -564,24 +569,32 @@ const QueryGenerator = {
...
+    const paramList = [];
+    Utils._.each(params, curParam => {
       const paramDef = [];
       if (Utils._.has(curParam, 'type')) {
         if (Utils._.has(curParam, 'direction')) { paramDef.push(curParam.direction); }
         if (Utils._.has(curParam, 'name')) { paramDef.push(curParam.name); }
         paramDef.push(curParam.type);
       } else {
-        throw new Error('createFunction called with a parameter with no type');
+        throw new Error('expandFunctionParamList: parameter missing type');
```
sushantdhiman: We don't need to change this error message, `expandFunctionParamList` is a private api that is used by `createFunction`.

package.json
```diff
@@ -113,6 +113,7 @@
     "test-integration-mysql": "DIALECT=mysql npm run test-integration",
     "test-integration-postgres": "DIALECT=postgres npm run test-integration",
     "test-integration-postgres-native": "DIALECT=postgres-native npm run test-integration",
+    "test-integration-query-interface-postgres": "DIALECT=postgres ./node_modules/.bin/mocha --globals setImmediate,clearImmediate --ui tdd --check-leaks --colors -t 15000 --reporter spec 'test/integration/dialects/postgres/query-interface.test.js'",
```
sushantdhiman: We don't need to add separate test command for `query interface`, we have integration test commands that will run those tests.

lib/dialects/postgres/query-generator.js
```diff
@@ -564,24 +569,32 @@ const QueryGenerator = {
...
       }
-      return paramDef.join(' ');
+
+      var joined = paramDef.join(' ');
+      return joined ?  paramList.push(joined): null;
```
sushantdhiman: if `joined` exists only push to `paramList`, no `return` required now

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -15,6 +15,190 @@ if (dialect.match(/^postgres/)) {
       this.queryInterface = this.sequelize.getQueryInterface();
     });
 
+
+    describe('createFunction', function(){
```
sushantdhiman: spaces in function expression, `function () {` , This applies to all other functions too

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -15,6 +15,190 @@ if (dialect.match(/^postgres/)) {
...
+
+      beforeEach(function(){
+        var self = this;
+        return self.queryInterface.dropFunction('create_job',[{type:'varchar',name:'test'}]).catch(function(err){
```
sushantdhiman: dont throw error in `beforeEach` , extract this test as separate `it` spec , Also use a better way to [check for errors](http://chaijs.com/api/bdd/#method_throw)

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -15,6 +15,190 @@ if (dialect.match(/^postgres/)) {
...
+      });
+
+      it('creates a stored procedure', function(){
+
+        var self = this;
+        var body = 'return test;';
+        var options = {};
+        return self.queryInterface.createFunction('create_job', [{type:'varchar',name:'test'}], 'varchar', 'plpgsql', body, options)
+        .then(function(){
+          return self.sequelize.query('select create_job(\'test\');', { type: self.sequelize.QueryTypes.SELECT })
```
sushantdhiman: pyramid of doom :0

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -15,6 +15,190 @@ if (dialect.match(/^postgres/)) {
...
+          });
+        });
+      });
+
+      it('requires parameters array', function(){
+        var self = this;
+        var body = 'return 1;';
+        var options = {};
+        try{
+          return self.queryInterface.createFunction('create_job',null, 'integer', 'plpgsql', body, options);
```
sushantdhiman: can be refactored to use the `.throws` helper from chaijs

lib/dialects/postgres/query-generator.js
turbofoxwave: expandFunctionParamList appears to be usedby 4 different functions in query-generator.
createTrigger, createFunction, dropFunction, renameFunction.
What do you think about reducing it to "parameter missing type" ?

package.json
turbofoxwave: understood, I'll remove this asap when I get home this evening.

lib/dialects/postgres/query-generator.js
sushantdhiman: Use this message "function or trigger used with a parameter without any type"

lib/dialects/postgres/query-generator.js
```diff
@@ -564,24 +566,32 @@ const QueryGenerator = {
...
         if (Utils._.has(curParam, 'name')) { paramDef.push(curParam.name); }
         paramDef.push(curParam.type);
       } else {
-        throw new Error('createFunction called with a parameter with no type');
+        throw new Error('expandFunctionParamList: parameter missing type');
       }
-      return paramDef.join(' ');
+
+      var joined = paramDef.join(' ');
+      paramList.push(joined);
```
sushantdhiman: `joined` can be empty, Use `if (joined) {` check

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -15,6 +17,182 @@ if (dialect.match(/^postgres/)) {
...
+        });
+      });
+
+      it('treats options as optional', function () {
+        var self = this;
+        var body = 'return test;';
+
+        //setup - ensure we don't have a create_job when we start
+        return new Promise.method( function (){
+          return self.queryInterface.dropFunction('create_job',[{type:'varchar',name:'test'}])
```
sushantdhiman: setup and cleanup both should go in `beforeEach` and `afterEach`. But you don't need to throw any errors inside `beforeEach` or `afterEach`

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -15,6 +17,182 @@ if (dialect.match(/^postgres/)) {
...
+            if (res[0].create_job !== 'test') throw new Error('expected output: test');
+          });
+        })
+        //cleanup
+        .then( function () {
+          return self.queryInterface.dropFunction('create_job',[{type:'varchar',name:'test'}]);
+        });
+      });
+
+      it('requires function name', function () {
```
sushantdhiman: all these require parameter test can be combined together with having mutiple `expect` to check for each case. Currently all these have exactly same code but with different missing argument to check

lib/dialects/postgres/query-generator.js
```diff
@@ -564,24 +566,32 @@ const QueryGenerator = {
...
         if (Utils._.has(curParam, 'name')) { paramDef.push(curParam.name); }
         paramDef.push(curParam.type);
       } else {
-        throw new Error('createFunction called with a parameter with no type');
+        throw new Error('function or trigger used with a parameter without any type');
       }
-      return paramDef.join(' ');
+
+      var joined = paramDef.join(' ');
+      if(joined) paramList.push(joined);
```
sushantdhiman: missed space with `if`

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -15,6 +17,163 @@ if (dialect.match(/^postgres/)) {
...
+      beforeEach( function () {
+        var self = this;
+        //make sure we don't have a pre-existing function called create_job
+        return self.queryInterface.dropFunction('create_job',[{type:'varchar',name:'test'}])
+        //suppress errors here. if create_job doesn't exist thats ok.
+        .catch( function (err) {  });
+      });
+
+      afterEach( function () {
+        var self = this;
```
sushantdhiman: Dont need `self`, you can just use `this`

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -15,6 +17,163 @@ if (dialect.match(/^postgres/)) {
...
+      });
+
+      it('creates a stored procedure', function () {
+        var self = this;
+        var body = 'return test;';
+        var options = {};
+
+        //make our call to create a function
+        return new Promise.method( function () {
+          return self.queryInterface.createFunction('create_job', [{type:'varchar',name:'test'}], 'varchar', 'plpgsql', body, options);
```
sushantdhiman: Why we need to wrap it in `Promise.method`, `createFunction` already returns a promise

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -15,6 +17,163 @@ if (dialect.match(/^postgres/)) {
...
+        })()
+        //validate
+        .then( function () { 
+          return self.sequelize.query('select create_job(\'test\');', { type: self.sequelize.QueryTypes.SELECT })
+          .then( function (res) {
+            if (res[0].create_job !== 'test') throw new Error('expected output: test');
```
sushantdhiman: rather than throw use `expect(res[0].create_job).to.be.eql('test')`

test/integration/dialects/postgres/query-interface.test.js
sushantdhiman: This goes for other cases too where you have wrapped `create / rename / drop Function()` with `Promise.method`, They all return Promise already, no need to wrap them

lib/dialects/postgres/query-generator.js
```diff
@@ -538,9 +539,10 @@ const QueryGenerator = {
   },
 
   dropFunction(functionName, params) {
+  if (!functionName) throw new Error("requires functionName");
```
mickhansen: indentation

mickhansen: `'` strings except for SQL

lib/dialects/postgres/query-generator.js
```diff
@@ -564,24 +566,32 @@ const QueryGenerator = {
     return this.quoteIdentifier(Utils.removeTicks(this.escape(val), "'"));
   },
 
+
+
```
mickhansen: ?

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -8,13 +8,162 @@ var chai = require('chai')
...
+
+    describe('renameFunction', function () {
+
+      it('can rename a function');
+      it('requires parameters array');
+      it('requires old function name');
+      it('requires new function name');
+      it('throws an error if called with no parameter type');
+
+    });
```
mickhansen: ?

turbofoxwave: These are test stubs for tests I felt we should have. My apologies, the last project I was in we put in stubs. I'll remove them.

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -8,13 +8,151 @@ var chai = require('chai')
...
+      it('can drop a function', function () {
+        var self = this;
+         
+        //call dropFunction
+        return self.queryInterface.dropFunction('droptest',[{type:'varchar',name:'test'}])
+        //validate
+        .then( function () {
+          return self.sequelize.query('select droptest(\'test\');', { type: self.sequelize.QueryTypes.SELECT })
+          .then( function () {
+            throw new Error('function failed to drop');
```
sushantdhiman: What is this code?  If function is dropped then `.then()` should never be triggered. it should automatically go to `catch` because we are trying to access a non existent function.

Also rather than `catch` and regex check, you can use the `expect.to.throw` here too.

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -8,13 +8,151 @@ var chai = require('chai')
...
 if (dialect.match(/^postgres/)) {
   describe('[POSTGRES Specific] QueryInterface', function () {
     beforeEach(function () {
       this.sequelize.options.quoteIdenifiers = true;
       this.queryInterface = this.sequelize.getQueryInterface();
     });
 
+    describe('createFunction', function () {
+
+      beforeEach( function () {
```
sushantdhiman: `beforeEach` and `afterEach` both are doing the cleanup work and duplicating it. In each test we are creating a function that we should remove before next test. You can just get rid of `beforeEach` , because `afterEach` is already doing the cleanup work

test/integration/dialects/postgres/query-interface.test.js
turbofoxwave: If droptest is not dropped as expected then  .then()  would be called because the droptest function will still exist.  The test will then produce an error stating the function failed to drop.

sushantdhiman: Here is my understandings so far

- `dropFunction` removes the `droptest`
- Try to call the function `droptest` (note it has been dropped in last statement), So we are calling a dropped function. This statement should be rejected
- Add `catch` handler  or expect `error` to be thrown

But you are expecting `.then()` in Step 3, then throw error yourself to catch it again why ? If `dropFunction` really fails to remove the function in Step 1, mocha should catch the failed promise.

You should do something like

```js
return self.queryInterface
.dropFunction('droptest',[{type:'varchar',name:'test'}])
.then(function () {
  return expect(self.sequelize.query('select droptest(\'test\');', {
    type: self.sequelize.QueryTypes.SELECT
  }))
  .to.eventually.rejectedWith(/function droptest.* does not exist/);
});
```

felixfbecker: Guys we can use arrow functions now 😄

test/integration/dialects/postgres/query-interface.test.js
turbofoxwave: if the test suite fails or stops unexpectedly causing afterEach's cleanup to not be called and the test suite is then restarted the cleanup in beforeEach will be needed.  What do you suggest?

sushantdhiman: Yes, `beforeEach` is definitely a good choice to prevent failures in such cases

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -6,15 +6,153 @@ var chai = require('chai')
...
+      it('creates a stored procedure', function () {
+        var self = this;
+        var body = 'return test;';
+        var options = {};
+
+        //make our call to create a function
+        return self.queryInterface.createFunction('create_job', [{type:'varchar',name:'test'}], 'varchar', 'plpgsql', body, options)
+        //validate
+        .then( function () { 
+          return self.sequelize.query('select create_job(\'test\');', { type: self.sequelize.QueryTypes.SELECT })
```
sushantdhiman: Please chain the promises properly so it doesn't become a [promise hell](https://medium.com/@pyrolistical/how-to-get-out-of-promise-hell-8c20e0ab0513). Also get rid of `self` (where possible) by using either `bind` or `arrow functions` (preferred way). Same goes for other code section too

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -6,15 +6,145 @@ var chai = require('chai')
...
+      beforeEach( function () {
+        //make sure we don't have a pre-existing function called create_job
+        //this is needed to cover the edge case of afterEach not getting called because of an unexpected issue or stopage with the 
+        //test suite causing a failure of afterEach's cleanup to be called.
+        return this.queryInterface.dropFunction('create_job',[{type:'varchar',name:'test'}])
+        //suppress errors here. if create_job doesn't exist thats ok.
+        .catch( err => {});
+      });
+
+      afterEach( function () {
```
sushantdhiman: As we discussed `beforeEach` is definitely required to cover edge cases, but we can remove `afterEach`

changelog.md
```diff
@@ -1,4 +1,6 @@
 # Future
+- [FIXED] sql syntax issue due to missing space in Postgresql query interface function "dropFunction"
```
sushantdhiman: Lets keep it simple, Like

``[FIXED] `dropFunction` syntax errors (PostgresSQL)``

changelog.md
```diff
@@ -1,4 +1,6 @@
 # Future
+- [FIXED] sql syntax issue due to missing space in Postgresql query interface function "dropFunction" 
+- [FIXED] Postgresql query interface function "createFunction" and "dropFunction" through addressing output issue with function expandFunctionParamList
```
sushantdhiman: Please remove this statement, one chagelog entry is sufficient

test/integration/dialects/postgres/query-interface.test.js
turbofoxwave: Without the afterEach the tests will not clean up after themselves. Is this ok?

sushantdhiman: Ok then we should add `after` rather than `afterEach`. With this the test will cleanup after each suite and same code will not run in each case.

sushantdhiman:
LGTM, Need @mickhansen's review before merging

turbofoxwave:
Hey Sushantdhiman, 
I know I might still have more work to do after mickhansen's review  but I wanted to take the opportunity to say thank you for your time. I appreciate your insight and direction. Its been very helpful and a great learning opportunity. 

Best Regards,
-Mike

sushantdhiman:
Thanks Mike, I was really impressed by your steadfastness and patience. Thank you for this contribution

felixfbecker:
Needs rebase

package.json
```diff
@@ -113,7 +113,6 @@
     "test-integration-mysql": "cross-env DIALECT=mysql npm run test-integration",
     "test-integration-postgres": "cross-env DIALECT=postgres npm run test-integration",
     "test-integration-postgres-native": "cross-env DIALECT=postgres-native npm run test-integration",
-    "test-integration-sqlite": "cross-env DIALECT=sqlite npm run test-integration",
```
sushantdhiman: Please dont remove the `sqlite` integration :)

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -6,15 +6,145 @@ var chai = require('chai')
...
+        var options = {};
+
+        //make sure we have a droptest function in place.
+        return this.queryInterface.createFunction('droptest', [{type:'varchar',name:'test'}], 'varchar', 'plpgsql', body, options)
+        //suppress errors.. this could fail if the function is already there.. thats ok. 
+        .catch( function (err) { });
+      });
+
+      it('can drop a function', function () {
+
```
sushantdhiman: Remove this extra line

test/integration/dialects/postgres/query-interface.test.js
```diff
@@ -6,15 +6,145 @@ var chai = require('chai')
...
+        .then( () => {
+          // call the function we attempted to drop.. if it is still there then throw an error informing that the expected behavior is not met. 
+          return this.sequelize.query('select droptest(\'test\');', { type: this.sequelize.QueryTypes.SELECT });  
+        }))
+        //test that we did get the expected error indicating that droptest was properly removed.
+        .to.be.rejectedWith(/.*function droptest.* does not exist/);
+      });
+
+      it('produces an error when missing expected parameters', function () {
+
```
sushantdhiman: Remove this extra line too

PR #6280, opened 2016-07-15
include: expand alias into association

### LINKED ISSUE #970: Include by Association Alias
As documented, you cannot eagerly load (include) an association via its alias when using a finder:

> "If an association is aliased (using the as option), you must specify this alias when including the model."

This makes it difficult to construct a finder when the query is coming from a restful client. The client would have to send the name of the model and the alias name in order for the server to perform the include in the query and the client would therefore need to know details of the data model which it should not.

I assume since the associations and there aliases are defined with the models, it should not be too difficult to perform a lookup when provided a string value to match an alias to its defined association. If that is true, it would be possible to perform a query to find an object and include its association via the association's alias.

What I would like to do is this:

``` javascript
Contact.hasMany(PhoneNumber, {as: 'phoneNumbers'});

Contact.find({
    where: { id: someId },
    include: [ 'phoneNumbers' ]
});
```

As a note, I am using the Odata v3 standard on the client side and interpreting on the server to construct the finder.

swarthy:
### Pull Request check-list

### Description of change

Expand string in `include` option into the association
(Closes #970)


sushantdhiman:
@swarthy Add a changelog entry tagged with `[ADDED]` to indicate this new feature.

LGTM, need review from @mickhansen @janmeier

mickhansen:
> Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

Be great if you also added an example or two to the proper docs.

changelog.md
```diff
@@ -1,4 +1,5 @@
 # Future
+- [ADDED] Expand string in `include` option into the association
```
mickhansen: `include now supports string as an argument (rather than model/association), string will expand into an association matched literally from Model.associations`

swarthy:
@mickhansen in which part of docs should I add an example?

mickhansen:
@swarthy https://github.com/sequelize/sequelize/blob/master/docs/docs/models-usage.md Eager loading etc

docs/docs/models-usage.md
```diff
@@ -501,6 +501,30 @@ User.findAll({ include: [{ model: Tool, as: 'Instruments' }] }).then(function(us
 })
 ```
 
+As alternative you can specify only alias.
```
mickhansen: `You can also include by alias name by specifying a string that matches the association alias:`

mickhansen:
LGTM, /cc @janmeier

felixfbecker:
LGTM, I love this feature.
@swarthy if you rebase this I will merge this

test/unit/model/include.test.js
```diff
@@ -253,6 +253,21 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       });
     });
 
+    describe('_conformInclude: string alias', function () {
+      it('should expand association from string alias', function () {
+        var options = {
```
felixfbecker: use `const`

felixfbecker:
It is not possible though to specify additional options like `attributes` that way. Maybe we can expand a string for the `association` aswell? What do you think?

swarthy:
See #6511

PR #6282, opened 2016-07-15
return a rejected promise on commit or rollback errors

connyay:
### Description of change
Returns a rejected promise on errors during inside of transaction commit & rollback.

Allows for more graceful catching of errors.

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -53,7 +53,7 @@ class Transaction {
   commit() {
 
     if (this.finished) {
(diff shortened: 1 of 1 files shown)


connyay:
[Documentation](http://docs.sequelizejs.com/en/latest/api/transaction/#commit-promise) also states that a promise will be returned which is not true in an error scenario currently.

sushantdhiman:
I think its better to return promise because that is what expected by end user.

Simple fix LGTM, Will need @mickhansen or @janmeier review.

Issue #4777, opened 2015-10-29
Problem with destroy hooks on n:m association


kiddouk:
Hi,

I am having a weird behavior despite applying what the documentation says.

I have 2 models and an n:m association via a ```through``` table. On that association, I define a hook ```afterCreate``` and it works just great when I add an association with ```post.addLikingUser(user)```

I also made sure to add the ```onDelete:'cascade', hooks: true``` options for the association. Nonetheless, the beforeDestroy/afterDestroy are never called when calling ```post.removeLikingUser(user)```

What am I missing ? 

The code (in coffeescript).

```coffee
Post = sequelize.define 'Post',
  {text:
    type: Sequelize.STRING, default: "Not set"
  ...
  },{
  getterMethods:
    createdAt: () ->
      moment(@getDataValue 'createdAt').format 'x'
    updatedAt: () ->
      moment(@getDataValue 'updatedAt').format 'x'
  }

PostLikedByUser = sequelize.define 'PostLikedByUser',
  {
    date:
      type: Sequelize.DATE, defaultValue: sequelize.NOW
  },
  {
    hooks: {
      beforeDestroy: (relation, options) ->
        console.log "beforeDestroy"
      afterDestroy: (relation, options) ->
        console.log "afterDestroy"
    }
  }


User = sequelize.define "User",
  device_id:
    type: Sequelize.STRING
  is_admin:
    type: Sequelize.BOOLEAN, defaultValue: false
  lastLoginAt:
    type: Sequelize.DATE, defaultValue: Sequelize.NOW


Post.belongsToMany User, {as: 'LikingUser', through: 'PostLikedByUser', onDelete: 'cascade', onUpdate: 'cascade', hooks: true}
User.belongsToMany Post, {as: 'LikedPost', through: 'PostLikedByUser', onDelete: 'cascade', onUpdate: 'cascade', hooks: true}

```

mickhansen:
As discussed on slack a `removeAssociation` call for `N:M` uses `Model.destroy` (bulk destroy) internally so the `individualHooks: true` option is needed to run a hook on each affected instance.

kiddouk:
While discussing with @mickhansen about that problem, it rutns out that 

1) remove call will trigger bulkDestroy
2) remove call supports ```individualHooks: true``` option in order to trigger individual after/beforeHooks.

I willl make a doc PR on that to fix that issue here.

PR #6286, opened 2016-07-16
Fix #4780, defaults to individualHooks:true for belongsToMany

### LINKED ISSUE #4780: [MAJOR] Default to `individualHooks: true` for N:M associations setup with `hooks: true`
https://github.com/sequelize/sequelize/issues/4777
- Should only apply setting if hooks for join model are detected

sushantdhiman:
### Pull Request check-list
### Description of change

Closes #4780 , Full discussion in #4777 

`BelongsToMany` will now default to `individualHooks: true` when removing the associations. This will only happen if `through` model have any `beforeDestroy` or `afterDestroy` hooks


mickhansen:
This needs to cover all associations methods that change the through table (so create and add aswell).

sushantdhiman:
Sure, I'll make those changes

sushantdhiman:
May be after other v4 issues i'll have a look on this

PR #6288, opened 2016-07-16
Add eslint

felixfbecker:
First step towards #5963 

Even though a great part of the library (tests) is not in ES6 yet and there are a whole lot of lint issues, I noticed that the continued development already introduced more style issues that did not get caught by JSHint. I decided it would be the best to add the `.eslintrc` now, so IDEs can pick it up and warn contributors about style issues while writing code.

I tried to match the current code style of the library as much as possible. Some issues like `no-extra-boolean-cast` are heavily used in the whole library, so I added them as warning for now. It is configured to especially error on not using ES6 (let/const/arrow functions etc) and I already used this file to do my refactors. One really nice thing is also that ESLint detects indentation errors while JSHint did not.

After this, I will try to fix as many style issues I find, starting with the simple ones like indentation, and when everything is resolved we can remove JSHint and add ESLint to Travis.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,47 @@
+{
+  "rules": {
+    "no-console": "off",
+    "no-extra-parens": "warn",
+    "valid-jsdoc": "off",
+    "new-cap": ["warn", {"properties": false}],
+    "comma-spacing": ["error", {"before": false, "after": true}],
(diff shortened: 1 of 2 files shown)


felixfbecker:
Anything that I should change or can this be merged?

janmeier:
Thanks for pinging @felixfbecker  ;)

PR #6289, opened 2016-07-16
Use underscore for private members instead of dollar

### LINKED ISSUE #5919: Dollar prefixes on methods

felixfbecker:
Closes #5919 

I did also touch some variables that used $ prefix for no reason.
No BC break because it's all private API.

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -514,9 +514,9 @@ class BelongsToMany extends Association {
         if (instance instanceof association.target) {
           return instance.where();
         } else {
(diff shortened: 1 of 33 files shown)


felixfbecker:
image

Issue #6131, opened 2016-06-19
PSQL queries with groupedLimit associations misquote schemas


ekosz:
## What you are doing?

```js
// Setup
var Sequelize = require('sequelize');
var sequelize = new Sequelize(process.env.DATABASE_URL, { 
  paranoid: false, timestamps: false, freezeTableName: true 
});
var Team = sequelize.define('Team', {}, { schema: 'entity', tableName: 'teams' });
var Goal = sequelize.define('Goal', {}, { schema: 'entity', tableName: 'goals' });
Team.Goals = Team.hasMany(Goal, { foreignKey: 'team_id' });

// The offending code
Team.findAll({ include: [{ model: Goal, limit: 100 }] });
```

## What do you expect to happen?
I expected to get get a list of `teams`, each one having upto 100 `goals`.

## What is actually happening?
Instead, the generated SQL throws an error: 

```
Unhandled rejection SequelizeDatabaseError: relation "entity.goals" does not exist
```

```sql
SELECT "Team"."id" FROM "entity"."teams" AS "Team";
SELECT "Goal".* FROM (
          (SELECT "id", "team_id" FROM "entity.goals" AS "Goal" WHERE "Goal"."team_id" = 1 LIMIT 
100) 
UNION ALL (SELECT "id", "team_id" FROM "entity.goals" AS "Goal" WHERE "Goal"."team_id" = 2 LIMIT 100)
) AS "Goal";
```

As you can see its trying to pull data from `"entity.goals"`. 
It should instead be correctly quoting the schema as `"entity"."goals"`

__Dialect:__ postgres
__Database version:__ psql (PostgreSQL) 9.4.4
__Sequelize version:__ 3.23.3

Related Issue: https://github.com/mickhansen/graphql-sequelize/issues/253

ekosz:
Looking through the code, it seems most likely to be a bug with [`groupedLimit`s in the `query-generator`](https://github.com/sequelize/sequelize/blob/v3.23.3/lib/dialects/abstract/query-generator.js#L1420-L1442). But I don't know the codebase well enough to pinpoint the issue past that.

ekosz:
Looking further I've traced the code to https://github.com/sequelize/sequelize/blob/v3.23.3/lib/dialects/abstract/query-generator.js#L1047 where Sequelize is calling `this.quoteTable('"entity"."goals"')` and its becoming `"entity.goals"`.

[...]

PR #6293, opened 2016-07-17
Add postgres failing test for postgres grouped limits with schemas

ekosz:
### Description of change

This adds a failing test for the issue #6131 . The error reported is:

```
SequelizeDatabaseError: relation "account.Tasks" does not exist
```

I tried to copy the surrounding tests in their style & naming, but I'm not quite sure if I got it right. Happy to update the test with feedback.

### DIFF
--- a/test/integration/include/schema.test.js
+++ b/test/integration/include/schema.test.js
@@ -707,6 +707,54 @@ describe(Support.getTestDialectTeaser('Includes with schemas'), function() {
       });
     });
 
+    it('should be possible to define a hasMany include with grouped limit', function() {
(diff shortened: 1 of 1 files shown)


mickhansen:
Hey @ekosz, thanks. Mind putting it in as unit test instead though? It's purely a SQL generation issue so no need to hit the database.

There are some for grouped limit here: https://github.com/sequelize/sequelize/blob/master/test/unit/sql/select.test.js

mickhansen:
Scratch that, lets keep the integration test.

ekosz:
@MichaelKearnsSoftwire 👌

Issue #6194, opened 2016-06-30
SQL Injection/Syntax error inserting GeoJSON documents with single quotes


contra:
## What you are doing?

Inserting inserting GeoJSON documents with single quotes in an attribute.

## What do you expect to happen?

No errors.

## What is actually happening?

Errors.

```
Executing (default): INSERT INTO "boundaries" ("id","name","type","geo","createdAt","updatedAt") VALUES ('0138248','Jacksons'' Gap','place',ST_GeomFromGeoJSON('{"type":"MultiPolygon","properties":{"NAME":"Jacksons' Gap","NAMELSAD":"Jacksons' Gap town","coordinates":[]}'),'2016-06-30 20:45:23.143 +00:00','2016-06-30 20:45:23.143 +00:00') RETURNING *;

SequelizeDatabaseError: syntax error at or near "Gap"
```

Seems like the GeoJSON string is not being escaped at all, so having a single quote in any property in the GeoJSON is causing a syntax error

__Dialect:__ postgres
__Database version:__ 9.5.3
__Sequelize version:__ 3.23.4

contra:
Just to be more clear: This is a pretty major security issue. Anyone who accepts user data with geojson types is vulnerable to a SQL injection on the insert.

contra:
Proof of concept:

```js
{
  "type": "MultiPolygon",
  "properties": {
    "exploit": "'); drop schema public cascade; -- "
  },
  "coordinates": []
}
```

POST this object to any Sequelize-backed service as the value of a GeoJSON type and it will drop the entire databse.

Query generated by Sequelize:

```sql
INSERT INTO "boundaries" ("geo") VALUES (ST_GeomFromGeoJSON('{"type":"MultiPolygon","properties":{"exploit": "'); drop schema public cascade; -- "},"coordinates":[]}')) RETURNING *;
```

PizzaBrandon:
This is now listed in NSP as a critical vulnerability and breaks all builds that use `nsp` during CI. https://nodesecurity.io/advisories/122

contra:
@PizzaBrandon 

- Vulnerability isn't in PostGIS or Postgres via ST_GeomFromGeoJSON, its solely on sequelize to do the escaping before passing the value to the DB.
- Validation won't help since this is valid GeoJSON

pdehaan:
[...]

PR #6305, opened 2016-07-19
MySQL GEOMETRY / GEOGRAPHY injection case

sushantdhiman:
### Pull Request check-list

### Description of change

After reading CVE details I realized this issue was also applicable for MySQL case but apparently by injecting wrong values to `coordinates`.  It also point to an issue in `WKT` library

Linking #6194


mickhansen:
Travis not running for some reason

PR #6313, opened 2016-07-20
Bugfix/order clause in subquery - for  the `master`

fogine:
### Pull Request check-list

### Description of change


fogine:
That's weird, I triggered travis-ci rebuild with no additional changes to code. (I fixed incorrect author's email in commits). Previous build has been successful. The current build failed with an error which does not make any sense to me.
Any ideas @sushantdhiman ?

fogine:
All tests are passing for me locally.

sushantdhiman:
I have restarted the failed build, it should be ok now

PR #6315, opened 2016-07-21
Docs: Updating getting started with Your first query

kevinsimper:
I thought the getting started guide missed Your first query, because if you go to the actual page that says "Querying", you don't get any information about actual querying, you have to go to "model usages" which is not quite logical.

So no people can start in the getting started guide like many people do and then go to the correct place to read more.

Also correct the section about promises to sound more nice.

Also `fieldName` and `freezeTableName` is nice features, but not something you need in a getting started guide. It just makes it look more complicated.

I even think that the `pool:` option in "Setting up a connection" should be removed, because who needs that anyways when they are first getting started? I could make new people think that they need it.

### DIFF
--- a/docs/docs/getting-started.md
+++ b/docs/docs/getting-started.md
@@ -59,16 +59,14 @@ Models are defined with `sequelize.define('name', {attributes}, {options})`.
 ```js
 var User = sequelize.define('user', {
   firstName: {
-    type: Sequelize.STRING,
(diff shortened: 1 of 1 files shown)


mickhansen:
I like it, thanks @kevinsimper. Querying is not complete, it was started to get all the query (where statements etc, includes) into one place, seperate from create/etc - But i only ever got so far :)

PR #6323, opened 2016-07-22
chore(hooks): remove callback support, closes #5228

### LINKED ISSUE #5228: Remove callback support for hooks

mickhansen:


lib/hooks.js
```diff
@@ -114,46 +114,43 @@ const Hooks = {
...
     }
 
     if (!Array.isArray(hooks)) {
-      hooks = hooks === undefined ? [] : [hooks];
+      hooks = [hooks];
     }
 
-    // run hooks as sync functions if flagged as sync
+    // synchronous hooks
     if (hookTypes[hookType] && hookTypes[hookType].sync) {
```
felixfbecker: Is this really needed? If the hook is sync, then `Promise.resolve()` will resolve immediately

mickhansen: The code calling this is also sync.

felixfbecker: Ah okay so the calling code relies on this returning instantly?

mickhansen: Yes, i think the only caller currently is beforeDefine

test/integration/hooks/associations.test.js
```diff
@@ -6,7 +6,8 @@ var chai = require('chai')
   , Support = require(__dirname + '/../support')
   , DataTypes = require(__dirname + '/../../../lib/data-types')
   , sinon = require('sinon')
-  , dialect = Support.getTestDialect();
+  , dialect = Support.getTestDialect()
+  , Promise = require('bluebird');
 
 describe(Support.getTestDialectTeaser('Hooks'), function() {
```
sushantdhiman: Almost all `function` expression here are missing space :laughing:, We should fix this formatting issue too

test/integration/hooks/associations.test.js
```diff
@@ -62,14 +63,14 @@ describe(Support.getTestDialectTeaser('Hooks'), function() {
           , beforeHook = false
           , afterHook = false;
 
-          this.Tasks.beforeUpdate(function(task, options, fn) {
+          this.Tasks.beforeUpdate(function(task, options) {
             beforeHook = true;
-            fn();
+            return Promise.resolve();
```
sushantdhiman: What happens if a hook doesn't return anything ? Does it hangs forever ?

mickhansen: https://github.com/sequelize/sequelize/pull/6323/files/bcd8eb11dece81b885d573e801200669f26e402a#r71967651

test/integration/hooks/associations.test.js
mickhansen: I agree, but style changes and functional changes should be kept seperate as much as possible so i'll leave it for another PR.

test/integration/hooks/associations.test.js
sushantdhiman: Oh, [Like this :wink: ](https://www.google.co.in/search?client=opera&q=recursion&sourceid=opera&ie=UTF-8&oe=UTF-8) (may have some spelling mistakes)

mickhansen: Recursion is not relevant here? We wrap whatever response in Promise.resolve() so if it returns undefined it's just an undefined result in the resolved promise.

sushantdhiman: Your link generates recursion 😂, I just given my version of that. Without `Promise.resolve` this code should work, right?

mickhansen: Oh lol, damn.

Yes without Promise.resolve it works aswell, but using `Promise.resolve()` tests the intended async nature.

sushantdhiman:
@mickhansen @janmeier LGTM, are there any other changes to make in this PR?

janmeier:
Yep, LGTM as well

Issue #6324, opened 2016-07-22
ER_SP_UNDECLARED_VAR: Undeclared variable: NaN


misgina:
We have recently upgraded from sequelize version 3.19.3 to 3.23.3. On version 3.19.3, sequelize was ignoring any false values for limit and offset (NaN, null and undefined). IN version 3.23.3 it only ignores if either null or undefined. The code links below show the changes made. We could not know if this is bug inroduced in the latest version or there is decision behind it. If it is a bug, we can submit a fix.

https://github.com/sequelize/sequelize/blob/v3.23.3/lib/dialects/abstract/query-generator.js#L1779

https://github.com/sequelize/sequelize/blob/v3.19.3/lib/dialects/abstract/query-generator.js#L1765

The error we got.
SequelizeDatabaseError: ER_SP_UNDECLARED_VAR: Undeclared variable: NaN
    at Query.formatError (/.../node_modules/sequelize/lib/dialects/mysql/query.js:175:14)
    at Query._callback (/.../node_modules/sequelize/lib/dialects/mysql/query.js:49:21)
    at Query.Sequence.end (/.../node_modules/mysql/lib/protocol/sequences/Sequence.js:85:24)
    at Query.ErrorPacket (/.../node_modules/mysql/lib/protocol/sequences/Query.js:94:8)
    at Protocol._parsePacket (/.../node_modules/mysql/lib/protocol/Protocol.js:280:23)
    at Parser.write (/.../node_modules/mysql/lib/protocol/Parser.js:74:12)
    at Protocol.write (/.../node_modules/mysql/lib/protocol/Protocol.js:39:16)
    at Socket.<anonymous> (/.../node_modules/mysql/lib/Connection.js:109:28)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:153:18)
    at Socket.Readable.push (_stream_readable.js:111:10)
    at TCP.onread (net.js:531:20)

**Dialect:** mysql 
__Database version: 5.5
__Sequelize version:3.23.3

Thanks.

sushantdhiman:
PR with fix for both `v3` and `master` would be appreciated

drunkhacker:
I have a same problem.  +1

misgina:
@drunkhacker we are working on a fix, will submit a pull request soon.

mickhansen:
The previous behaviour could also have been considered a bug.
[...]

PR #6341, opened 2016-07-27
Handle non valid offset or limit values v4

misgina:
fixes sequelize/sequelize#6324

---

This change is [Reviewable](https://reviewable.io/reviews/sequelize/sequelize/6341)

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1743,15 +1743,18 @@ const QueryGenerator = {
   addLimitAndOffset(options, model) {
     let fragment = '';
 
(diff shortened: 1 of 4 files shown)


mickhansen:
Needs a rebase

misgina:
@mickhansen updated, thanks.

sushantdhiman:
From https://github.com/sequelize/sequelize/issues/6324#issuecomment-235867273
> The previous behaviour could also have been considered a bug.
Passing a bad string should not necessarily result in loading all rows in the database.

@mickhansen Can we merge this fix or we should handle bad string case in some other way

PR #6353, opened 2016-07-29
Fix the unstable test "should work with two schema models in a hasMany association"

topikachu:
### Pull Request check-list

### Description of change

The origin  test "should work with two schema models in a hasMany association" relys on the record insert order which makes the result randomly
Add record id explicitly and query in explicitly order

### DIFF
--- a/test/integration/include/seperate.test.js
+++ b/test/integration/include/seperate.test.js
@@ -389,6 +389,7 @@ if (current.dialect.supports.groupedLimit) {
             return this.sequelize.sync({force: true}).then(() => {
               return Promise.join(
                 User.create({
+                  id:1,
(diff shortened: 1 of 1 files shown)


test/integration/include/seperate.test.js
```diff
@@ -389,6 +389,7 @@ if (current.dialect.supports.groupedLimit) {
             return this.sequelize.sync({force: true}).then(() => {
               return Promise.join(
                 User.create({
+                  id:1,
```
sushantdhiman: should be like `id: 1` (spaces)

sushantdhiman: Ok its just a test :)

PR #6354, opened 2016-07-29
Migrate to Node-MySQL2

### LINKED ISSUE #3495: Prepared statements
- [x] Review driver support
- [x] Migrate to node-mysql2
- [ ] Estimate refactor time

### LINKED ISSUE #6327: Migrate to Node-MySQL2
I have internally started migrating Sequelize code base to [node-mysql2](https://github.com/sidorares/node-mysql2). 

This is one of the step to add prepared statement support #3495 #1608 . Apart from prepared statements `node-mysql2` offers significant improvements in [performance](https://gist.github.com/sidorares/ffe9ee9c423f763e3b6b).

Are there any reasons we should stick with `node-mysql` ?

sushantdhiman:
### Pull Request check-list

### Description of change

Closes #6327 , With this all dialects driver support the ability to use the prepared statements. Its first step towards #3495 

**Notable changes**

- `MySQL` doesn't support named timezones ( out of box ) so I have used `moment-timezone` to convert named timezone to their relative offsets.
- ~~`node-mysql2` provides `createConnection()`, which directly connect with server ( `node-mysql2` offers `connect` method). It can throw `EventEmitter` based errors which can only be caught using `domains`. I am open to other ideas if we do not want to use `domains`~~ We no longer use `domains`, thanks @felixfbecker for pushing me to find alternate solution :)
- `mysql` dialect now parse `BLOB` , to return `null` in case we get an empty buffer.

**Performance**

`sequelize-benchmark` reports improvement in performance.

`MySQL Server 5.7`
`Intel® Core™ i3-3240T CPU @ 2.90GHz × 4 `
`4GB RAM` 

`node-mysql2`
mysql2

`node-mysql`
mysql

### DIFF
--- a/lib/dialects/mysql/connection-manager.js
+++ b/lib/dialects/mysql/connection-manager.js
@@ -1,11 +1,11 @@
 'use strict';
 
 const AbstractConnectionManager = require('../abstract/connection-manager');
+const SequelizeErrors = require('../../errors');
 const Utils = require('../../utils');
+const DataTypes = require('../../data-types').mysql;
+
 const debug = Utils.getLogger().debugContext('connection:mysql');
-const Promise = require('../../promise');
-const sequelizeErrors = require('../../errors');
-const dataTypes = require('../../data-types').mysql;
 const parserMap = new Map();
 
 class ConnectionManager extends AbstractConnectionManager {
@@ -16,9 +16,9 @@ class ConnectionManager extends AbstractConnectionManager {
     this.sequelize.config.port = this.sequelize.config.port || 3306;
     try {
       if (sequelize.config.dialectModulePath) {
-        this.lib = require(sequelize.config.dialectModulePath);
+        this.lib = require(sequelize.config.dialectModulePath + '/promise');
       } else {
-        this.lib = require('mysql2');
+        this.lib = require('mysql2/promise');
       }
     } catch (err) {
       if (err.code === 'MODULE_NOT_FOUND') {
@@ -27,7 +27,7 @@ class ConnectionManager extends AbstractConnectionManager {
       throw err;
     }
 
-    this.refreshTypeParser(dataTypes);
+    this.refreshTypeParser(DataTypes);
   }
 
   // Expose this as a method so that the parsing may be updated when the user has added additional, custom types
@@ -56,7 +56,6 @@ class ConnectionManager extends AbstractConnectionManager {
    * @return Promise<Connection>
    */
   connect(config) {
-    return new Promise((resolve, reject) => {
       const connectionConfig = {
         host: config.host,
         port: config.port,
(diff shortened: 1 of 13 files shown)


felixfbecker:
:-1: on domains. Isnt mysql2 supposed to be 100% API compatible with mysql? Can't we register an error handler on the EventEmitter? If not, I would rather do a PR at mysql2 than using domains, it opens a whole new bag of problems.

sushantdhiman:
@felixfbecker No more `domains`

sushantdhiman:
@janmeier @mickhansen Need your review on this please. I want to start working on prepared statements. This PR and #6170 needs to be merged for that

lib/dialects/mysql/connection-manager.js
```diff
@@ -69,82 +85,93 @@ class ConnectionManager extends AbstractConnectionManager {
...
-
-          return;
-        }
+      return new Utils.Promise((resolve, reject) => {
+        const connection = this.lib.createConnection(connectionConfig);
+        connection.once('error', (err) => {
+          reject(err);
+        });
+        connection.once('connect', () => {
+          connection.removeAllListeners('error');
```
janmeier: Might an error not still happen after we have connected? For example if the connection is lost while the connection object is in the pool, but is not being used?

janmeier:
LGTM, just a single comment about the error listener :)

lib/dialects/mysql/connection-manager.js
felixfbecker: Well in that case it makes no sense anymore though to reject the connection promise, as it is already resolved. But I don't think it is good to remove all listeners, just use `once()`.

sushantdhiman: @janmeier Connection lost error are handled in next `.then` block. https://github.com/sequelize/sequelize/pull/6354/files#diff-ed5bbd2edaf04e2052f0919e5658fe22R107

@felixfbecker If we don't remove the `error`handler in connect that we will have two erro events attached [1) reject](https://github.com/sequelize/sequelize/pull/6354/files#diff-ed5bbd2edaf04e2052f0919e5658fe22R90) and [2) lost connection handler](https://github.com/sequelize/sequelize/pull/6354/files#diff-ed5bbd2edaf04e2052f0919e5658fe22R107), where we only need (2). So its required to removed (1) in `connect` phase

mickhansen: @sushantdhiman the connection promise ought to be resolved at that point though so the reject doesn't matter.

In any case we need to remove both the connect and error listener (atleast the specific ones) on either connect or error so we don't have a leak.

sushantdhiman: @mickhansen `connect` event is attached with `once`, so it will never be called after first handshake. Will it still cause leak ?

If it does, then as you suggested I can removed both `error` and `connect` event by reference.

mickhansen: @sushantdhiman If you get an error, it will not be released. But the connection argument would then ideally be garbage collected, so i'm not actually sure.

lib/dialects/mysql/connection-manager.js
```diff
@@ -88,12 +88,16 @@ class ConnectionManager extends AbstractConnectionManager {
       return new Utils.Promise((resolve, reject) => {
         const connection = this.lib.createConnection(connectionConfig);
 
-        // clean up event
-        let errorHandler = (e) => {
+        let errorHandler, connectHandler;
```
mickhansen: i actually prefer non hoisted, and you can also use `const` here.

sushantdhiman: me too, but jslint will throw warning. Should I add ignore flag ?

mickhansen: Try const, might know throw for that :)

sushantdhiman: It does :), I am adding ignore flag

mickhansen: what rule fails?

sushantdhiman: `'connectHandler' was used before it was defined.`

sushantdhiman: I am adding `latedef:false` for this block

mickhansen: Ah right cause the methods refer to eachother.

sushantdhiman:
@mickhansen @janmeier I have added changelog entry. Let me know if there are any other changes to make in this PR. Thanks

janmeier:
Sweet :)

sidorares:
👍 
I guess it's time to push 1.0 release for mysql2
I'll push rc11  today (minor fix for pool option) and if no other critical bugs reported make 1.0

PR #6357, opened 2016-07-31
[Docs] Add include:[{model: User.scope('...')}] section

stheobald:
### Description of change
Added a missing piece regarding invoking `.scope()` on an `include`d model in a `scope` definition.


janmeier:
Thanks :)

PR #6372, opened 2016-08-03
Fix mysql LIKE where condition on date fields

rundef:
### Pull Request check-list

### Description of change

I am using the MySQL dialect and when I run this query:

```js
model.findAll({
  where: {
    startedAt: { $like: '2016-07-%' },
  }
})
```

I get something like this :
`where startedAt LIKE '2016-07-01 04:00:00'`

But what I should get is this :
`where startedAt LIKE '2016-07-%'`

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,5 @@
 # Future
+- [FIXED] MySQL LIKE condition did not work as expected with DATE fields [PR#6372](https://github.com/sequelize/sequelize/pull/6372)
 - [ADDED] SQLCipher support via the SQLite connection manager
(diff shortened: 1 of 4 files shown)


felixfbecker:
LIKE is an operator that works on strings. Have you tried `Sequelize.cast()`?

rundef:
Thanks for the help, I ended up with this : 

```js
const dateCondition = Sequelize.where(Sequelize.cast(Sequelize.col('startedAt'), 'CHAR'), { $like: '2016-07-%' });
const where = Sequelize.and(
  Sequelize.where(model.rawAttributes['deletedAt'], null),
  dateCondition
);      
model.findAll({
  where,
});
```

PR #6373, opened 2016-08-04
Fix getting true when upsert a row to its current value with mysql

### LINKED ISSUE #6310: upsert return wrong value when insert a row which db has
If table user has a row like below:

(id, name) = (1, 'daleoooo')

and run code like this

``` js
user.upsert({
  id: 1,
  name: 'daleoooo'
}).then(res => console.log(res))
```

the output will be true, which is obviously not right.
I traced the code, and found [connection.query](https://github.com/sequelize/sequelize/blob/3e5b8772ef75169685fc96024366bca9958fee63/lib/dialects/mysql/query.js#L40) returns wrong affectRows.

BTW:
I run the sql as same as the sql sequelize generated in the code above, affected row is 0. 

Thank you for your comments.

**Dialect:** mysql
**Database version:** 5.7
**Sequelize version:** 3.23.6

daleoooo:
Fixed #6310 : added flags: '-FOUND_ROWS' at createConnection in mysql dialect.

PS: I have added test to ensure my change is valid, also I have passed tests while using `DIALECT=mysql | sqlite | postgres npm run test`, but `npm run test-docker` gives me some errors so I wasn't able to confirm that all tests pass.

Thanks


janmeier:
Does this also fail with node-mysql2? (do we need this on master)

daleoooo:
I have test with node-mysql2, and it needs this flag too. 

PS: It is my bad to let you wait for such a long time(Because of my vocation).

daleoooo:
@janmeier Would you please merge my fix? This bug makes my [repo](https://github.com/Meituan-Dianping/koa-restql) unstable, thanks.

janmeier:
@daleoooo Please rebase, and provide a PR to master (should be able just to cherry-pick the commit

Issue #6363, opened 2016-08-02
Empty result when query with nested includes on postgres when quoteIdentifiers is false


topikachu:
## What you are doing?

```js
it('can select nested include', function() {
        var self = this;
        this.sequelize.options.quoteIdentifiers = false;
        this.sequelize.getQueryInterface().QueryGenerator.options.quoteIdentifiers = false;
        
        this.Professor  = this.sequelize.define('Professor', {
          fullName: DataTypes.STRING
        }, {
          quoteIdentifiers: false
        });
        this.Class = this.sequelize.define('Class', {
          name: DataTypes.STRING
        }, {
          quoteIdentifiers: false
        });
        this.Student = this.sequelize.define('Student', {
          fullName: DataTypes.STRING
        }, {
          quoteIdentifiers: false
        });
        this.ClassStudent = this.sequelize.define('ClassStudent', {
        }, {
          quoteIdentifiers: false,
	        tableName: 'class_student'
        });
        this.Professor.hasMany(this.Class);
        this.Class.belongsTo(this.Professor);
        this.Class.belongsToMany(this.Student,{through: this.ClassStudent});
        this.Student.belongsToMany(this.Class,{through: this.ClassStudent});
        return this.Professor.sync({ force: true })
          .then(function() {
            return self.Student.sync({ force: true })
          })
          .then(function() {
            return self.Class.sync({ force: true })
          })
          .then(function() {
            return self.ClassStudent.sync({ force: true })
          })
          .then(function() {
              return self.Professor.bulkCreate([
                {
                  id: 1,
                  fullName: 'Albus Dumbledore'
                },
                {
                  id: 2,
                  fullName: 'Severus Snape'
                }
              ])
            })
          .then(function() {
            return self.Class.bulkCreate([
              {
                id: 1,
                name: 'Transfiguration',
                ProfessorId: 1
              },
              {
                id: 2,
                name: 'Potions',
                ProfessorId: 2
              },
              {
                id: 3,
                name: 'Defence Against the Dark Arts',
                ProfessorId: 2
              }
            ])
          })
          .then(function(classes) {
            return self.Student.bulkCreate([
              {
                id: 1,
                fullName: 'Harry Potter',
              },
              {
                id: 2,
                fullName: 'Ron Weasley',
              },
              {
                id: 3,
                fullName: 'Ginny Weasley',
              },
              {
                id: 4,
                fullName: 'Hermione Granger',
              }
            ])
          })
          .then(function() {
[...]

PR #6375, opened 2016-08-04
Let nested include query returns results when quoteIdentifiers is false on Postgres

topikachu:
### Pull Request check-list

- [x] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving? #6363

### Description of change

The origin code change the row attributes order which makes  AbstractQuery._groupJoinData can't append the join data to the root.
Modify the code to keep the key sequence when translate the lower case key name to camelcase name

### To reviewer:
Another solution is to modify https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query.js#L481. However, I don't want to introduce the performance overhead on other database. But the order returned by Object.keys is arbitrary. So AbstractQuery._groupJoinData is builtin on a specific JS runtime behavior and may be wrong in the future.

### DIFF
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -199,7 +199,7 @@ class Query extends AbstractQuery {
               const targetAttr = attrsMap[key];
               if (typeof targetAttr === 'string' && targetAttr !== key) {
                 return targetAttr;
-              }else{
+              } else {
                 return key;
               }
             });
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -194,17 +194,17 @@ class Query extends AbstractQuery {
             m[k.toLowerCase()] = k;
             return m;
           }, {});
-          for (const row of rows) {
-            for (const key of Object.keys(row)) {
+          rows = _.map(rows,function(row) {
+            return _.mapKeys(row, function(value, key) {
               const targetAttr = attrsMap[key];
               if (typeof targetAttr === 'string' && targetAttr !== key) {
-                row[targetAttr] = row[key];
-                delete row[key];
+                return targetAttr;
+              }else{
+                return key;
               }
-            }
-          }
+            });
+          });
         }
-
         return this.handleSelectQuery(rows);
       } else if (QueryTypes.DESCRIBE === this.options.type) {
         result = {};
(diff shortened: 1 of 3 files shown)


lib/dialects/postgres/query.js
```diff
@@ -194,17 +194,17 @@ class Query extends AbstractQuery {
...
-                row[targetAttr] = row[key];
-                delete row[key];
+                return targetAttr;
+              }else{
+                return key;
               }
-            }
-          }
+            });
+          });
```
mickhansen: Why change this?

test/integration/dialects/postgres/dao.test.js
```diff
@@ -939,6 +941,144 @@ if (dialect.match(/^postgres/)) {
...
+        this.ClassStudent = this.sequelize.define('ClassStudent', {
+        }, {
+          quoteIdentifiers: false,
+	        tableName: 'class_student'
+        });
+        this.Professor.hasMany(this.Class);
+        this.Class.belongsTo(this.Professor);
+        this.Class.belongsToMany(this.Student,{through: this.ClassStudent});
+        this.Student.belongsToMany(this.Class,{through: this.ClassStudent});
+        return this.Professor.sync({ force: true })
```
sushantdhiman: Use arrow functions to minimize or eliminate `self` (where possible)

sushantdhiman:
@topikachu Test too :)

lib/dialects/postgres/query.js
```diff
@@ -194,17 +194,17 @@ class Query extends AbstractQuery {
...
-          for (const row of rows) {
-            for (const key of Object.keys(row)) {
+          rows = _.map(rows,(row)=> {
+            return _.mapKeys(row, (value, key)=> {
               const targetAttr = attrsMap[key];
               if (typeof targetAttr === 'string' && targetAttr !== key) {
-                row[targetAttr] = row[key];
-                delete row[key];
+                return targetAttr;
+              }else{
```
sushantdhiman: I forgot about this :(, need spaces around `else`

topikachu: NP. ^-^. It has conflicts now and I will refine my PR.

felixfbecker:
@topikachu your git email is not set correctly 😉

PR #6414, opened 2016-08-11
Use typeof instead of _.isPlainObject

felixfbecker:
### Pull Request check-list

### Description of change

Changes from using `_.isPlainObject` to checking with `typeof === 'object'` to support custom class instances like from `sql-template-strings`.


janmeier:
LGTM

PR #6434, opened 2016-08-14
Rename Sequelize class to Connection

felixfbecker:
Reasoning: In ES6 / TypeScript, it feels extremely weird to do
```js
import * as Sequelize from 'sequelize';
const sequelize = new Sequelize.Sequelize(process.env.DB);
```
For ES5 users, nothing changes.


felixfbecker:
Actually, changed my mind. Connection doesn't make sense really because it's a Pool.

PR #6437, opened 2016-08-14
Clean up Model.init() API

felixfbecker:
- Removes the `modelManager` argument. I checked the code and the only way `modelManager` was ever used was to get `modelManager.sequelize`, which is already available at `Model.sequelize`.
 - I moved the docs from `Sequelize.define()` to `Model.init()` since that is the primary API for the future. Added a _See ..._ note.
 - `sequelize` is passed as option in `options`. The plan is to one day make it optional, for now it is required. This is actually the way it was passed to Model.init in v3 already.
 - Moved important logic like option merging that was still in `Sequelize.define()` to `Model.init()`

One thing I don't know though is what should happen to the `beforeDefine` hook. I moved it to `Model.init`, but the name is kind of confusing. The problem is though that there is already the `beforeInit` hook and the hooks are applied as a mixin to both the Model and Sequelize. Imo this doesn't even make much sense because AFAIK some hooks can't even be applied to `Model`, like `beforeBulkSync`. My proposal is to refactor the hooks so that they are not applied to both classes, and then `beforeInit` on the model class. That issue shouldn't block this PR though.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,6 @@
 # Future
+- [CHANGED] Removed `modelManager` parameter from `Model.init()` [#6437](https://github.com/sequelize/sequelize/issues/6437)
+- [FIXED] Made `Model.init()` behave like `sequelize.define()` (hooks are called and options have proper defaults) [#6437](https://github.com/sequelize/sequelize/issues/6437)
 - [ADDED] `restartIdentity` option for truncate in postgres [#5356](https://github.com/sequelize/sequelize/issues/5356)
 - [INTERNAL] Migrated to `node-mysql2` for prepared statements [#6354](https://github.com/sequelize/sequelize/issues/6354)
 - [ADDED] SQLCipher support via the SQLite connection manager
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -57,7 +57,7 @@ class BelongsToMany extends Association {
     this.target = target;
     this.targetAssociation = null;
     this.options = options;
-    this.sequelize = source.modelManager.sequelize;
+    this.sequelize = source.sequelize;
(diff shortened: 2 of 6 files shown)


felixfbecker:
Tests pass, it's only Coveralls...
```
/home/travis/build/felixfbecker/sequelize/node_modules/coveralls/bin/coveralls.js:18
        throw err;
        ^
Bad response: 504 <!DOCTYPE html>
<html lang="en">
```

felixfbecker:
Can I get review on this please?

janmeier:
LGTM

Regarding hooks - Init could happen more than once I guess, if you want to attach the model to multiple sequelize instances, so it makes sense to have it in the model class. Good question about define hooks though - A lot of plugins depend on those hooks to manipulate attributes. 

I think we should provide init as an alias for define - (mayhaps with a warning that define is preferred)

Anyhow, not blocking this PR, so I'll just wait for travis to re-run

felixfbecker:
I would also like to provide `define` The problem is that the existing init hook has completely different smenatics, it is run before and after the constructor run: http://docs.sequelizejs.com/en/latest/api/hooks/#beforeinitname-fn

And with the current mixin-model we cannot have two hooks with the same name that behave differently for Sequelize and the Model :/

My idea was that we clearly differentiate between `Sequelize` and `Model` hooks - most of the hooks should belong to the `Model` class then and would be registered on the Model class:
```js
Sequelize.Model.beforeBulkCreate(...);
```
Which makes sense if we move towards connection-independent models.

janmeier:
Ah yes, `beforeInit` is actually before a new sequelize instance is inited, got me confused there for a sec..

janmeier:
Breakage on mssql is due to https://github.com/nodejs/node/issues/8127, merging

Issue #6270, opened 2016-07-13
Diff against a clone of attributes instead of using setters


felixfbecker:
Currently, Sequelize holds all attributes inside `dataValues`. You can use them on the instance, but it is all one big lie as on instance creation, sequelize simply defines getters and setters that proxy to `dataValues` and mark the attribute as dirty. The problem is though that setters don't catch the following changes to attributes:
- When using a JSON column, you can alter the object
- When using an ARRAY column, you can push new items, remove item etc #6197 
- When using a DATE column, you can call `setDate()` etc on the object

You have to manually call `instance.changed('myAttribute', true)` (no autocomplete for the attribute name) or reassign with `instance.myAttribute = instance.myAttribute` (wtf, should be a noop).

Now think about how a Model is defined in Sequelize 4.0 and ES6 / TS:

``` ts
class User extends Sequelize.Model {
  public username: string;
  public roles: string[];
  public createdAt: Date;
}
```

You define your class with these attributes, but in reality, they are not there - it's all a big lie. You look at the debugger and they are not there:

![image](https://cloud.githubusercontent.com/assets/10532611/16817248/3d23a2d4-4942-11e6-82de-54a0acb68ebd.png)

I think it would be a lot better if instead of `dataValues`, there was only `_previousDataValues`. When an instance is retrieved from the DB, the _actual_ instance gets all attributes set as real attributes. `_previousDataValues` gets set to a deep copy of the attributes - that means every `Date` object, every `JSON` object, every `Array` gets deeply cloned. This does not include aggregate columns, it's only about the values of the database row.

A user can then modify and use the instance as much as he likes, nothing goes through any setters or getters (except if he defined them of course).

Only upon a call to `save()` we deeply compare the instance _once_ to the `_previousDataValues`. That includes `JSON` objects and arrays. We can then generate a minimal query that only contains the exact changes that were made to the instance.

It doesn't even have to replace a whole array column, since we have an exact diff we can even use advanced array functions to only remove / insert what changed.
### Performance

`saving` would be a bit more complex because we have to compare the objects, but every userland function containing property access (get/set) will be faster because getters and setters are an optimization killer.

felixfbecker:
I just found the needed info is already there under `_previousDataValues`

mickhansen:
This will break all computes values that aren't targeted virtuals.
Virtuals don't make sense in all cases like doing a window function for a full count.

mickhansen:
```js
class User extends Sequelize.Model {
[...]

PR #6439, opened 2016-08-15
Use classes for examples in model definition docs

felixfbecker:
This is my proposal for the docs with the new style of defining models. I think it is important to get the docs out now so we can get feedback on the proposed syntax.

### DIFF
--- a/docs/docs/models-definition.md
+++ b/docs/docs/models-definition.md
@@ -1,25 +1,30 @@
 ## Definition
 
-To define mappings between a model and a table, use the `define` method. Sequelize will then automatically add the attributes `createdAt` and `updatedAt` to it. So you will be able to know when the database entry went into the db and when it was updated the last time. If you do not want timestamps on your models, only want some timestamps, or you are working with an existing database where the columns are named something else, jump straight on to [configuration ][0]to see how to do that.
+To define mappings between a model and a table in v4, you define a class that inherits from [`Sequelize.Model`](../api/model) and call [`Model.init(attributes, options)`](../api/model#init). The model name is determined by the class name. You need to pass the sequelize instance as `options.sequelize` (see example)
+Sequelize will then automatically add the attributes `createdAt` and `updatedAt` to it. So you will be able to know when the database entry went into the db and when it was updated the last time. If you do not want timestamps on your models, only want some timestamps, or you are working with an existing database where the columns are named something else, jump straight on to [configuration ][0]to see how to do that.
 
+Defining a model with [`sequelize.define()`](../api/sequelize#define) as in previous versions is still supported. Check the v3 docs for examples.
 
 ```js
-var Project = sequelize.define('project', {
+class Project extends Sequelize.Model { }
+Project.init({
   title: Sequelize.STRING,
   description: Sequelize.TEXT
-})
+}, { sequelize })
 
-var Task = sequelize.define('task', {
+class Task extends Sequelize.Model { }
+Task.init({
   title: Sequelize.STRING,
   description: Sequelize.TEXT,
   deadline: Sequelize.DATE
-})
+}, { sequelize })
 ```
 
 You can also set some options on each column:
 
 ```js
-var Foo = sequelize.define('foo', {
+class Foo extends Sequelize.Model { }
+Foo.init({
  // instantiating will automatically set the flag to true if not set
  flag: { type: Sequelize.BOOLEAN, allowNull: false, defaultValue: true},
 
@@ -70,7 +75,7 @@ var Foo = sequelize.define('foo', {
      deferrable: Sequelize.Deferrable.INITIALLY_IMMEDIATE
    }
  }
-})
+}, { sequelize })
 ```
 
 The comment option can also be used on a table, see [model configuration][0]
@@ -162,12 +167,13 @@ Usage in object notation:
 
 ```js
 // for enums:
-sequelize.define('model', {
+class MyModel extends Sequelize.Model { }
+MyModel.init({
   states: {
     type:   Sequelize.ENUM,
     values: ['active', 'pending', 'deleted']
   }
-})
+}, { sequelize })
 ```
 
 ### Range types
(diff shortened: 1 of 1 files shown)


docs/docs/models-definition.md
```diff
@@ -256,92 +262,78 @@ the rule in a transaction. See [the transaction section](transactions/#options)
...
-* retrieving an underlying property value - always use `this.getDataValue()`
+If you want to define a getter or setter with the same name as your model attribute, you can prefix it and configure the `field` column option.
 
 ```js
-/* a getter for 'title' property */
-function() {
-    return this.getDataValue('title');
+class Transaction extends Sequelize.Model {
+  get status() {
+    return this._status.toLowerCase();
```
janmeier: Having to add an underscore feels ugly to me - I'd rather we advertise `getDataValue` for getters named the same as their property

felixfbecker: Personally I dislike `getDataValue` because you cannot get any compile-time checks or IntelliSense for the attribute name, as it is a string.

But the primary motivation was that #6270 might change instances so that attributes are saved on the instance directly instead of in `dataValues` (and we only keep `previousDataValues`). There are a couple of reasons why this would be good, we wouldn't need getters and setters for the attributes anymore, which means better performance and that an instance can be treated like a plain simple instance of the model class. You can for example use the `delete` operator properly, which previously only deleted the getter/setter.

Prefixing the attribute with an underscore is the established pattern for implementing proxy accessors. Basically the actual property is private (and by convention prefixed with underscore) while the setters/getters are public.

For example http://stackoverflow.com/a/12850536/4208018 (it's TS but the same applies to ES6)

docs/docs/models-definition.md
```diff
@@ -1,25 +1,30 @@
...
+Defining a model with [`sequelize.define()`](../api/sequelize#define) as in previous versions is still supported. Check the v3 docs for examples.
 
 ```js
-var Project = sequelize.define('project', {
+class Project extends Sequelize.Model { }
+Project.init({
   title: Sequelize.STRING,
   description: Sequelize.TEXT
-})
+}, { sequelize })
```
mickhansen: Lets not change the documentation for model definition untill we actually have a proper API.
`Model.init` is not the end goal, it prefer define() to it.

felixfbecker: What is the end goal then?
@gajus just posted on Slack
> I like the `Model.init(attributes, options)` approach. It mixes well with ES6 classes.

mickhansen: I'd prefer having attributes and options defined on the class and then added to Sequelize via a call.
So close to what we have now, but attributes/options should be statically defined so as to be public and easily inspectable by users or plugins

felixfbecker: So basically something like `static options = {...}; static attributes = {...}`? We still need a `.init()` call though because there is a lot of normalization happening. And I would find it weird when I as a user declare `attributes`, but after the `init()` call they have a completely different structure because everything was normalized. I'm +1 for making `options` and `attributes` public (they are not underscore-prefixed anyway in current versions) but as long as `options` gets a somehow different value after init I prefer passing it to the `init()` function.
The desire to have the attributes and options defined _inside_ the class and not in a function call directly following the class declaration will be solved by decorators.

mickhansen: normalization can happen just in time or either on `sequelize.addModel` or `Model.init`.
As for exposing the normalization, i agree that's a tricky issue for options, less so for attributes.

I would like to see something like:

```js
class MyModel extends Sequelize.Model {
  static attributes = {};
  static tableName = '';
  static schema = '';
}

// ES6 compat

class MyModel extends Sequelize.Model {

}
MyModel.attributes = {};
MyModel.tableName = '';
MyModel.schema = '';
```

Whatever we decide on, we definitely don't have a final API so i don't think we should advertise anything other than the original .define API untill we do.

felixfbecker: > normalization can happen just in time 

I don't think this is good because it surely has a performance impact if we either don't normalize and adapt every codepiece where options are accessed or results in way more code if we call `_normalizeIfNotNormalizedYet()` everywhere.

> As for exposing the normalization, i agree that's a tricky issue for options, less so for attributes.

I don't think so - aren't the data types normalized (e.g. if you passed `Sequelize.STRING` it will be normalized to `new Sequelize.STRING()`)?

> I would like to see something like: ...

The table name is already in `options.tableName`. But is your ES6 compat so much different from a `Model.init()` call? Because if you are using Babel/TS already, you can use decorators, which work even better. This is the API I would like to see (and this supported as of 4.0.0-1):
```ts
// TypeScript

@Options({ sequelize, tableName: ''})
class User extends Model {
  @Attribute({ type: DataTypes.STRING, unique: true })
  public username: string;
}

// ES7 / Babel
@Options({ sequelize, tableName: ''})
@Attributes({
  username: {
    type: DataTypes.STRING,
    unique: true
  }
})
class User extends Model {

}

// ES6 compat
class User extends Model {

}
Model.init({
  username: {
    type: DataTypes.STRING,
    unique: true
  }
}, {
  sequelize,
  tableName: ''
})
```

> Whatever we decide on, we definitely don't have a final API so i don't think we should advertise anything other than the original .define API untill we do.

True, the API is not final, but neither is the prerelease we already have out for v4. Which means people can already use this, but they don't have any docs on it and cannot provide feedback to the new API.

mickhansen: They can use it, but i don't see a lot of value in advertising or collecting feedback on an API we haven't even finalized internally yet - It will bring more confusion when it likely changes.

I'm not a fan of your ES7 example, why not use statics rather than decorators?
Having attributes as an overall decorator also removes the options of having attribute level decorators in the future

I think there's a good case for having options be explicit rather than an object (especially for stuff like name, schema and table).

mickhansen: Do we have an issue anywhere where we're doing API design for model definitions with ES6 and ES7 in mind?

felixfbecker: > I'm not a fan of your ES7 example, why not use statics rather than decorators?

Because the decorators can normalize their input before decorating the class, while you can't if you are declaring statically for the reasons already mentioned. I admit though that from my 3 example it is definitely the most ugly one. If I wasn't using TS I would just go with `init()`.

> I think there's a good case for having options be explicit rather than an object (especially for stuff like name, schema and table).

Yes, I agree, but that's not how it is implemented right now. I just started from how it is atm because it is less code to change.

> Having attributes as an overall decorator also removes the options of having attribute level decorators in the future

That is not true - see my TypeScript example. We can supply both. Actually I am working on a decorator library I want to publish on NPM to make the first two examples possible with 4.0.0-1 and experiment with it in the projects I am involved in.

felixfbecker: https://github.com/sequelize/sequelize/issues/5898

felixfbecker: I think this deserves it's own issue though...

mickhansen: Definitely deserves it's own issue :)

> Because the decorators can normalize their input before decorating the class, while you can't if you are declaring statically for the reasons already mentioned

Good point.

> Yes, I agree, but that's not how it is implemented right now. I just started from how it is atm because it is less code to change.

Understandable, i'd love to discuss the most optimal API given future ES7 features while considering how it would look using only ES6 currently.

felixfbecker: @mickhansen I found the issue: https://github.com/sequelize/sequelize/issues/4728

sushantdhiman: I would prefer use of `define`

docs/docs/models-definition.md
sushantdhiman: `getDataValue` seems cleaner than `_status`, Once you finish implementing #6270 we can decide how to change docs to adopt to new structure.

docs/docs/models-definition.md
leebenson: > Because the decorators can normalize their input before decorating the class, while you can't if you are declaring statically for the reasons already mentioned. I admit though that from my 3 example it is definitely the most ugly one. If I wasn't using TS I would just go with init().

[Mobx](https://github.com/mobxjs/mobx) has great decorator syntax, that's ES6 / Babel compatible.

I have code snippets like this working in production:

```
import { observable } from 'mobx';

export default class Menu {
  @observable active = false;
}
```

... which is available in Babel using the [transform-decorators-legacy](https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy) plug-in.

Perhaps there's a comparable syntax? Something like:

```
@options({ tableName: 'user', sequelize }) class User extends Model {
  @attribute name = {
    type: DataTypes.STRING,
    unique: true,
    // ...
  }
}
```

BTW, I wound up here after digging around for v4 documentation and bouncing between posts referencing `Model.init()` vs `define()`... has this been resolved yet? Would really like not to go back to v3 if I can help it... too much ES6 goodness here already :-)

felixfbecker: @leebenson The plan is to keep `sequelize.define()` for BC. `Model.init()` is a new API that can be used for example by decorator libraries or by users, if they prefer it over `define()`. It can also be used by a decorator library. But as @mickhansen put it (regarding the current future syntax ideas):

> there isn't anything that just screams beautiful to me yet

I am interested in your feedback, but this shouldn't belong here.

leebenson: Apologies for dropping the comment in this commit message - until you neatly laid out the commit thread, navigating v4 was tough!

> there isn't anything that just screams beautiful to me yet

I know this isn't the place for syntax, but I'd like to point casual users that may inadvertently wind up here over to [how I'm currently writing v4-compatible syntax](there isn't anything that just screams beautiful to me yet)

IMO, the solution is very simple, and about as succinct as it gets. I'm not sure it'll 'scream beautiful' per @mickhansen's requirements, but it's exactly what I was hoping to get out of v4 and solves my use-case nicely.

tugh:
@felixfbecker Hi thanks for the example. 

I'm using it with babel and getting following error. 
```
ReferenceError: Attributes is not defined
```

I'm running my express app by following script:
```bash
babel-node --plugins transform-decorators-legacy,transform-class-properties,transform-es2015-modules-commonjs,transform-async-to-generator ./bin/www
```

Issue #6441, opened 2016-08-15
Milliseconds for createdAt & updatedAt


bsapot:
I'm wanting to use milliseconds in createdAt and updatedAt on a couple of tables in MySQL. I noticed after changing the data types to DATE(6) that the values are stored without milliseconds. 

I tracked down the issue in the code and found here that milliseconds on are removed unless the database is progress or sqlite. 

[Link to code causing issue](https://github.com/sequelize/sequelize/blob/master/lib/utils.js#L425)
## What you are doing?

This is an abbreviated version of the model. 

``` js
module.exports = function (sequelize, DataTypes) {
    var MachineEvents = sequelize.define("machine_events", {
        id: {type: DataTypes.UUID, allowNull: false, defaultValue: DataTypes.UUIDV1, unique: true, primaryKey: true},
        companyId: {type: DataTypes.UUID, allowNull: false},
        eventDate: {type: DataTypes.DATE(6), allowNull: true, defaultValue: null},
        createdAt: {type: DataTypes.DATE(6), allowNull: false},
        updatedAt: {type: DataTypes.DATE(6), allowNull: false}
    }, {
        freezeTableName: true // Model tableName will be the same as the model name
    });
```
## What do you expect to happen?

I expect the createdAt and updatedAt to include milliseconds. I'm not setting the values manually. It does work when I set eventDate manually. 

``` sql
INSERT INTO `machine_events` (`id`,`companyId`,`eventDate`,`qty`,`duration`,`tagQuality`,`tagName`,`tagValue`,`userId`,`processed`,`createdAt`,`updatedAt`) VALUES ('123f54b0-632c-11e6-96e8-4df6948f9d7d','company','2016-08-15 21:06:01.366',0,0,true,'St Louis.HAAS-90.QTY1',1,'bsapot-id',false,'2016-08-15 21:06:01.744','2016-08-15 21:06:01.744')
```
## What is actually happening?

_Output, either JSON or SQL_

``` sql
INSERT INTO `machine_events` (`id`,`companyId`,`eventDate`,`qty`,`duration`,`tagQuality`,`tagName`,`tagValue`,`userId`,`processed`,`createdAt`,`updatedAt`) VALUES ('123f54b0-632c-11e6-96e8-4df6948f9d7d','company','2016-08-15 21:06:01.366',0,0,true,'St Louis.HAAS-90.QTY1',1,'bsapot-id',false,'2016-08-15 21:06:01','2016-08-15 21:06:01')
```

[...]

PR #6442, opened 2016-08-15
Milliseconds in time stamps for createdAt, updatedAt for MySQL 

bsapot:
Closes [#6441](https://github.com/sequelize/sequelize/issues/6441)
### Pull Request check-list
### Description of change

Added mysql to the list of dialects that return milliseconds from the now function. Updated timezone test to account for milliseconds in the time being compared.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -422,7 +422,7 @@ exports.sliceArgs = sliceArgs;
 
 function now(dialect) {
   const now = new Date();
-  if (['mysql','postgres', 'sqlite'].indexOf(dialect) === -1) {
+  if (['mysql', 'postgres', 'sqlite'].indexOf(dialect) === -1) {
     now.setMilliseconds(0);
   }
   return now;
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -422,7 +422,7 @@ exports.sliceArgs = sliceArgs;
 
 function now(dialect) {
   const now = new Date();
-  if (['postgres', 'sqlite'].indexOf(dialect) === -1) {
+  if (['mysql','postgres', 'sqlite'].indexOf(dialect) === -1) {
     now.setMilliseconds(0);
   }
   return now;
--- a/test/integration/timezone.test.js
+++ b/test/integration/timezone.test.js
@@ -52,7 +52,7 @@ if (dialect !== 'sqlite') {
           // Expect 7 hours difference, in milliseconds.
           // This difference is expected since two instances, configured for each their timezone is trying to read the same timestamp
           // this test does not apply to PG, since it stores the timezone along with the timestamp.
-          expect(this.normalUser.createdAt.getTime() - timezonedUser.createdAt.getTime()).to.be.closeTo(60 * 60 * 7 * 1000, 50);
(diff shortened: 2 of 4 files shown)


test/integration/timezone.test.js
```diff
@@ -52,7 +52,7 @@ if (dialect !== 'sqlite') {
           // Expect 7 hours difference, in milliseconds.
           // This difference is expected since two instances, configured for each their timezone is trying to read the same timestamp
           // this test does not apply to PG, since it stores the timezone along with the timestamp.
-          expect(this.normalUser.createdAt.getTime() - timezonedUser.createdAt.getTime()).to.be.closeTo(60 * 60 * 7 * 1000, 50);
+          expect(this.normalUser.createdAt.getTime() - timezonedUser.createdAt.getTime()).to.be.closeTo(60 * 60 * 7 * 1000, 1000);
```
janmeier: Why does this break?

lib/utils.js
```diff
@@ -421,7 +421,7 @@ exports.sliceArgs = sliceArgs;
 
 function now(dialect) {
   const now = new Date();
-  if (['postgres', 'sqlite'].indexOf(dialect) === -1) {
+  if (['mysql','postgres', 'sqlite'].indexOf(dialect) === -1) {
```
janmeier: I guess the extra part is just ignored for regular dates in mysql? 

Does this trigger a warning?

test/integration/timezone.test.js
janmeier: Nevermind :

janmeier:
Mind adding a unit (sql generation test) for this?

lib/utils.js
bsapot: No it does not trigger a warning and yes it just drops the milliseconds. I tested on 5.7 and 5.6.

bsapot:
I'm having trouble creating a sql generation test for this. The time stamps for createdAt, updatedAt are generated in model.js [here](https://github.com/sequelize/sequelize/blob/master/lib/model.js#L3118) which is before sql generation. 

I can create a save integration test that validates the change or I can update the utils.test.js to check this. 

Do you have a prefrence?

janmeier:
Integration test is fine in that case :)

bsapot:
@janmeier do you know if there are any special settings on the mysql instance in travis? I'm getting a syntax error creating the table with DATETIME(6) columns. It does not happen on my local machine 5.7.10 or on amazon 5.6.10a. 

I looked at the travis config and did not see anything but I wanted to double check before I get too far down a rabbit hole.

janmeier:
I'm pretty sure we are running mysql 5.5. We'll probably need to wait for https://github.com/sequelize/sequelize/pull/6443 so we can get 5.6

sushantdhiman:
@bsapot need rebase

bsapot:
@sushantdhiman rebase is done, tests are passing except a postgres test on node 6. Is that an existing issue with the code I used?

sushantdhiman:
Retriggering the build

lib/utils.js
felixfbecker: Space missing after comma

felixfbecker: Space missing after comma

bsapot:
haha - good catch!

Issue #6345, opened 2016-07-28
Failing tests with MySQL 5.7


Naoto-Ida:
I am currently trying to solve #6307 and familiarize myself with the innards of the library, and have started to look into the code.

This is the first time I am dealing with tests, so would like to ask the contributors directly. I followed the instructions for contributing and ran `npm run test-mysql` in the sequelize directory.

I have the following errors without any editing done to the current code.
```
1366 passing (2m)
  2 pending
  4 failing

  1) [MYSQL] DataTypes allows me to return values from a custom parse function:
     SequelizeDatabaseError: ER_TRUNCATED_WRONG_VALUE: Incorrect datetime value: '2011-10-30 15:00:00 +00:00' for column 'dateField' at row 1
      at Query.formatError (lib/dialects/mysql/query.js:204:16)
      at Query.connection.query [as _callback] (lib/dialects/mysql/query.js:56:23)
      at Query.Sequence.end (node_modules/mysql/lib/protocol/sequences/Sequence.js:85:24)
      at Query.ErrorPacket (node_modules/mysql/lib/protocol/sequences/Query.js:94:8)
      at Protocol._parsePacket (node_modules/mysql/lib/protocol/Protocol.js:280:23)
      at Parser.write (node_modules/mysql/lib/protocol/Parser.js:74:12)
      at Protocol.write (node_modules/mysql/lib/protocol/Protocol.js:39:16)
      at Socket.<anonymous> (node_modules/mysql/lib/Connection.js:109:28)
      at readableAddChunk (_stream_readable.js:172:18)
      at Socket.Readable.push (_stream_readable.js:130:10)
      at TCP.onread (net.js:542:20)
  From previous event:
      at Query.run (lib/dialects/mysql/query.js:44:12)
      at retry (lib/sequelize.js:635:32)
      at node_modules/retry-as-promised/index.js:38:21
      at retryAsPromised (node_modules/retry-as-promised/index.js:27:10)
      at Promise.resolve.then.connection (lib/sequelize.js:635:14)
  From previous event:
      at Promise.then (lib/promise.js:21:17)
      at Context.<anonymous> (test/integration/data-types.test.js:65:42)

  2) [MYSQL] DataTypes should parse an empty GEOMETRY field:
     SequelizeDatabaseError: UNKNOWN_CODE_PLEASE_REPORT: Invalid GIS data provided to function st_geometryfromtext.
      at Query.formatError (lib/dialects/mysql/query.js:204:16)
[...]

Issue #4727, opened 2015-10-25
MySQL JSON Data Type


nodecode:
MySQL now has a JSON data type. See here: https://dev.mysql.com/doc/refman/5.7/en/json.html

An implementation in Sequelize.js (like with PostgreSQL) would be nice.

http://sequelize.readthedocs.org/en/latest/docs/models-definition/#data-types

janmeier:
Interesting definitely! :)

It doesn't seem like 5.7 is supported on travis yet (I'd rather not have to fiddle around with a script that manually installs it), so thats the first hurdle

camerondubas:
+1

mujz:
+1

golya:
+1

gilo:
+1

twilly86:
+1

miralem:
+1

tosone:
+1

seanzx85:
+1

rromanw:
+1

saurabhverma2892:
+1

ccg1415:
+1

Glavin001:
Any progress on this? I would be interested in contributing, however I have yet to dive into Sequelize internals yet. Any pointers would be greatly appreciated! Thanks!

janmeier:
@Glavin001 Perhaps you can draw some inspiration from https://github.com/sequelize/sequelize/pull/4013, which added geometry support. 

Basically, you need to add some code to mysql/data-types, something like this https://github.com/sequelize/sequelize/blob/a89f0f818657fc289c97e7f433b3499d02e661b5/lib/dialects/postgres/data-types.js#L276-337. Parse handles converting data that comes from the DB to json, while stringify converts JSON to the format the db expects.

If you have futher questions, ask them here, open the PR and we can discuss further, or join us on slack :)

janmeier:
One problem is that travis-ci doesn't support mysql 5.7 out of the box - But you could try using https://gist.github.com/BenMorel/d981f25ead0926a0cb6d to update the mysql version.

If installing a new mysql version doesn't take too long, we might as well upgrade to the 14.04 image (which doesn't have mysql installed by default) - then we wouldn't have to manually upgrade gcc and clang https://github.com/sequelize/sequelize/blob/a89f0f818657fc289c97e7f433b3499d02e661b5/.travis.yml#L28-33

ramsundhar20:
[...]

PR #6436, opened 2016-08-14
Fix all auto-fixable ESLint issues

felixfbecker:

PR #6443, opened 2016-08-16
Update Travis-CI to use Docker

toddbluhm:
Attempt to move travis-ci over to using docker to help with #4727 

My goal for this PR was to keep the same matrix style in travis-ci that already existed, leverage the existing docker/docker-compose and just get it working on travis-ci.

### Pull Request check-list

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?**Currently not all the tests pass due to: #6345 and #6279 (same error: Too few ordinates in GeoJSON)**
- [x] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving? **#4727**
**I tweaked the incorrect password integration test case because the error was different from what was expected (but consistent across all dialects). My guess is it has to do with running in the docker environment. I left the original code commented out until I know this is the way we want to go.**
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? **Updated contributor readme file with new docker instructions**
**I added an entry for it**

### Description of change

- Tweaked travis.yml to use docker for running the dbs
- Removed existing docker file as it was no longer needed
- Updated docker compose to the new v2 style
- Updated postgres docker container to a postgis container to support the new postgis queries
- Removed the ./node_modules/.bin/... from package.json scripts because they are not needed when run from npm (it by default looks in that folder)
- Updated package.json docker test scripts to include new env's for the docker db host/port settings


janmeier:
I'd prefer not having to build a sequelize docker container on CI, because it adds to the build time.

I think we should rather just expose the ports in the compose file and run sequelize via travis' built in node / nvm.

janmeier:
Otherwise, good work :)

.travis.yml
```diff
@@ -1,19 +1,14 @@
-language: node_js
```
janmeier: We might consider upgrading to trusty (https://docs.travis-ci.com/user/trusty-ci-environment/) since we don't need mysql installed anymore

toddbluhm:
@janmeier  Just updated and fixed per your comments. I also updated the contributor readme to account for the docker changes. Let me know what you think.

janmeier:
Looks good to me :)

toddbluhm:
@janmeier I created a little util called [`env-cmd`](https://github.com/toddbluhm/env-cmd) to help with passing all of the env vars to a script via an env file. I added it and I am using it with the `test-docker` scripts in package.json. This is something we could use for travisci too, instead if having all those env vars in the travis.yml (so have all the env vars in one env file). If that sounds like something you would like, I can make that change.

felixfbecker:
@toddbluhm why not https://www.npmjs.com/package/cross-env?

toddbluhm:
@felixfbecker As far as I can tell, cross-env doesn't accept a file input. I am using env-cmd because I want to pass multiple env vars without dealing with a huge list. With cross-env I think I would still need that huge list. My goal wasn't so much to make it portable as cross-env does, just make it a shorter script command...the portability was just free. Hopefully that makes sense...unless I am totally misunderstanding how cross-env works.

package.json
```diff
@@ -64,7 +64,7 @@
     "commander": "^2.6.0",
     "continuation-local-storage": "^3.1.4",
     "coveralls": "^2.11.9",
-    "dox": "~0.9.0",
```
janmeier: Please don't remove dox ;)

janmeier:
env-cmd looks good to me :)

package.json
toddbluhm: Good catch! I clearly missed that in the rebase. I will fix that right now.

toddbluhm:
Recent issues with Travis-ci and docker seem to be making random tests fail. Not much we can do about this until they get it all resolved. See issue here: https://github.com/travis-ci/travis-ci/issues/6461

janmeier:
Note to self - Once this is merged, test out https://hub.docker.com/r/rsmoorthy/mssql/

changelog.md
```diff
@@ -41,6 +41,7 @@
 - [FIXED] Nested query return correct result when quoteIdentifiers is false. (Postgres) [#6363](https://github.com/sequelize/sequelize/issues/6363)
 - [FIXED] Fixed an issue where changing multiple ENUM columns in PostgreSQL could break. [#6203] (https://github.com/sequelize/sequelize/issues/6203)
 - [FIXED] Add `parent`, `original` and `sql` properties to `UniqueConstraintError`
+- [CHANGED] Converted over to using docker containers for the dbs in Travis-ci [#4727] (https://github.com/sequelize/sequelize/issues/4727)
```
felixfbecker: Shouldn't this be `[INTERNAL]` or something like that? Not really an API change, isn't it?

package.json
```diff
@@ -97,21 +98,20 @@
...
     "test": "if [ $COVERAGE ]; then npm run coveralls; else npm run jshint && npm run teaser && npm run test-unit && npm run test-integration; fi",
-    "test-docker": "docker-compose run sequelize /bin/sh -c \"npm run test-all\"",
-    "test-docker-unit": "docker-compose run sequelize /bin/sh -c \"npm run test-unit-all\"",
-    "test-docker-integration": "docker-compose run sequelize /bin/sh -c \"npm run test-integration-all\"",
-    "build-docker": "docker-compose build",
+    "test-docker": "env-cmd ./test/config/.docker.env npm run test-all",
+    "test-docker-unit": "env-cmd ./test/config/.docker.env npm run test-unit-all",
+    "test-docker-integration": "env-cmd ./test/config/.docker.env npm run test-integration-all",
     "docs": "node docs/docs-generator.js",
-    "jshint": "./node_modules/.bin/jshint lib test",
```
felixfbecker: 👍

toddbluhm: One thing I was confused about is that there is both jshint and eslint in this project. Are we trying to convert to eslint or sticking with jshint? jshint is used in `npm test`, but eslint is used in the `npm run lint` command...just curious.

changelog.md
toddbluhm: Yeah I we can change that. Also it should be under `Future` section which is not in changelog right now, so I have to add it

package.json
felixfbecker: See #5963 #6288 #6436 
We are switching to ESLint but there are many lint issues JSHint didn't detect. So in CI JSHint is still used but we have the eslintrc for developers and their IDEs.

toddbluhm:
@janmeier took a look at that mssql container. Did not have much luck with it. I think the mssql version is just too old (mssql2000). Having said that, apparently MS is releasing SQL Server on linux sometime which will most likely make its way to docker and we can use that. Currently there is an official mssql express 2014 server for docker, but it will only run in a windows vm.

janmeier:
Roger that, thanks for looking into it (:

janmeier:
@toddbluhm Build matrix has been updated so you don't have to worry about mssql anymore. Also, https://github.com/sequelize/sequelize/pull/6479 was merged which means all tests should now pass with postgres 9.5 and mysql 5.7 

Do you mind rebasing?

toddbluhm:
@janmeier Its rebased and I also added in multi-version testing so that it now tests agains mysql 5.5-5.7 and postgres 9.4 & 9.5. There is still an error with mysql 5.6-5.7, but it looks like it might just be an error message change or something like that.

janmeier:
Cool, I really like the idea of testing against older versions! This does however add 3 extra builds, so I'd like another contributor to OK this.

[This test](https://github.com/sequelize/sequelize/blob/9f0b5d1af24f6741aadebba411a21dc1d774d2c1/test/integration/model.test.js#L2365) fails for mysql 5.7 - Appearently the error message changed.

[This one](https://github.com/sequelize/sequelize/blob/2e42171262abb734e42c248de346b392eceae513/test/integration/sequelize.test.js#L270) fails for mysql 5.6 - Probably something related to the mysql setting for when to throw an error and when to log a warning

I've re-run the postgres jobs, since it wasn't obvious to me what was wrong there - lets see it that fixes it

test/integration/data-types.test.js
```diff
@@ -285,14 +285,15 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
...
-            .then((result) => {
-              if (result[0][0] && semver.lte(result[0][0].extversion, '2.1.7')) {
-                resolve(true);
-              } else {
-                resolve();
-              }
-            }).catch(reject);
+        current.query(`SELECT PostGIS_full_version();`)
+          .then((result) => {
+            const version = result[0][0].postgis_full_version.match(/^POSTGIS="(\d*.\d*.\d*)/)[1];
```
janmeier: `PostGIS_lib_version` maybe? http://postgis.net/docs/PostGIS_Lib_Version.html

test/integration/data-types.test.js
```diff
@@ -285,14 +285,15 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
...
       if (/^postgres/.test(dialect)) {
-        current.query(`SELECT extversion FROM pg_catalog.pg_extension WHERE extname='postgis';`)
-            .then((result) => {
-              if (result[0][0] && semver.lte(result[0][0].extversion, '2.1.7')) {
-                resolve(true);
-              } else {
-                resolve();
-              }
-            }).catch(reject);
+        current.query(`SELECT PostGIS_full_version();`)
```
janmeier:
```js
current.query(`SELECT PostGIS_full_version();`, {
  type: 'SELECT'
}).spread(result => result.postgis_full_version)
```

test/integration/data-types.test.js
sushantdhiman: that didn't worked on my local setup. But current one is same thing as it reads direct tables which `PostGIS_lib_version` would have accessed :)

test/integration/data-types.test.js
sushantdhiman: Please check my other comment

test/integration/data-types.test.js
toddbluhm: The reason I changed it was that this command gives a different (correct) version number than the current one in master. For some reason they differ and I am not sure why. I just updated to use `PostGIS_lib_version` because its cleaner and maybe it will help?!

sushantdhiman:
@toddbluhm We might need to add a version check [here ](https://github.com/sequelize/sequelize/blob/9f0b5d1af24f6741aadebba411a21dc1d774d2c1/test/integration/model.test.js#L2386)too

toddbluhm:
@janmeier Looks like the only thing we have left is the weird deadlock issue with postgres. I have fixed the other mysql stuff

janmeier:
Let's see if a re-run fixes things, otherwise I'll try to clone locally and see if I can reproduce. I think I've seen something similar where the order in which indexes are created might cause a deadlock, but I thought that was fixed ...

sushantdhiman:
I think current build is pretty heavy, it takes a bit extra time to finish. Let keep it lean and only run for these matrix setup

- Node 4 , `MySQL 5.7`, `Postgres 9.5`, `Postgres 9.5(Native)` and `SQLite`
- Node 6 , `MySQL 5.7`, `Postgres 9.5`, `Postgres 9.5(Native)`

There is no need to test old database versions, if there are any specific issue related to old version maintainers can reproduce them at their local end. Lets keep CI focused on latest available database engines.

bokskar:
@sushantdhiman 
Yeah I definitely can understand that and will update it. 

One thought though is maybe setup it up to run the full suite of tests on merge into master rather than just always running them? Thoughts?

toddbluhm:
@sushantdhiman 
Yeah I definitely can understand that and will update it.

One thought though is maybe setup it up to run the full suite of tests on merge into master rather than just always running them? Thoughts?

sushantdhiman:
> One thought though is maybe setup it up to run the full suite of tests on merge into master rather than just always running them? Thoughts?

@toddbluhm If you mean running all test suite in one instance, I think thats not desirable because

- Harder to pinpoint which database dialects are failing and on which Node versions
- Waiting time will be about 20 min

toddbluhm:
No, that was not what I was meaning, but I just did some research and what I was thinking of is not easily possible with Travis-ci. Its okay, I will just push the reduced dbs shortly.

Rio517:
Any progress @toddbluhm ?

lunaroverlord:
Looking forward to getting this merged!

janmeier:
Thanks a lot @toddbluhm

PR #6469, opened 2016-08-19
Fix postgis insert empty coordinates for point test case

toddbluhm:
### Pull Request check-list

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)? **Yes**
- [x] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving? **The test case change fixes postgres errors in #6279 & #6443**
- [x] Have you added new tests to prevent regressions? **No new tests needed**
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? **No**
- [x] Have you added an entry under `Future` in the changelog? **Yes**

### Description of change

Fixed a test case for postgres/postgis that was failing do to attempting to insert a geojson object with an empty coordinates array.

`Postgis >=2.1.7` made a change to throw an error if a geojson object was created with an empty coordinates array (instead of using `[0,0]` as it was previously doing). This was not very well documented in the postgis [changelog](http://postgis.net/docs/release_notes.html#idm38454) (see malformed GeoJSON inputs), but the change can be found in this [diff](https://trac.osgeo.org/postgis/changeset/13401/) (L96-101)

This will fail under the current travis build as I am guessing travis uses a lower postgis version. But it will fix travis postgres failing test cases for #6443

### DIFF
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -274,45 +274,6 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
     testFailure(Type);
   });
 
-  it('should parse an empty GEOMETRY field', function () {
-   var Type = new Sequelize.GEOMETRY();
-
-   if (current.dialect.supports.GEOMETRY) {
-     current.refreshTypes();
-
-     var User = current.define('user', { field: Type }, { timestamps: false });
-     var point = { type: "Point", coordinates: [] };
-
-     return current.sync({ force: true }).then(function () {
-       return User.create({
-          //insert a null GEOMETRY type
-          field: point
-        });
-      }).then(function () {
-        //This case throw unhandled exception
-        return User.findAll();
-      }).then(function(users){
-        if (dialect === 'mysql') {
-          // MySQL will return NULL, becuase they lack EMPTY geometry data support.
-          expect(users[0].field).to.be.eql(null);
-        } else if (dialect === 'postgres' || dialect === 'postgres-native') {
-          // Used to return [0,0] but change made around postgis 2.1.7 to throw error instead:
-          // https://trac.osgeo.org/postgis/changeset/13401 (L96-101)
-          expect(!'Should throw error. Postgis >=2.1.7 does not support empty coordinates.');
-        } else {
-          expect(users[0].field).to.be.deep.eql(point);
-        }
-      })
-      .catch(function(e) {
-        if (dialect === 'postgres' || dialect === 'postgres-native') {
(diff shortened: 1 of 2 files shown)


janmeier:
I think we can just remove the test completely, since neither postgis (post 2.1.7) nor mysql supports empty geometry

toddbluhm:
Sounds good to me 😄

test/integration/data-types.test.js
```diff
@@ -273,36 +273,6 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
...
-          expect(users[0].field).to.be.eql(null);
-        } else if (dialect === 'postgres' || dialect === 'postgres-native') {
-          //Empty Geometry data [0,0] as per https://trac.osgeo.org/postgis/ticket/1996
-          expect(users[0].field).to.be.deep.eql({ type: "Point", coordinates: [0,0] });
-        } else {
-          expect(users[0].field).to.be.deep.eql(point);
-        }
-      });
-   }
  });
```
janmeier: This line should be removed as well ;)

toddbluhm: Fixed

sushantdhiman:
@janmeier Its fixed in https://github.com/sequelize/sequelize/tree/mysql-5.7.x-test-fail , Just need to work on MySQL Logging Warning cases

toddbluhm:
@sushantdhiman Oh nice! Did not even notice that branch. I was just fixing it so I could get the travis docker tests passing and get that merged in so we could test out newer db versions easier. I am more than happy to wait on that branch and close this out if thats best.

toddbluhm:
@sushantdhiman Actually I am wondering if maybe we should just cherry-pick that postgres/postgis fix here as its probably out of place in the mysql-5.7 branch? Then we could merge the postgis fix in and the would fix my docker pr. Then you could use the docker pr when its merged to test against mysql-5.6 & mysql 5.7 for the mysql-5.7 branch? Currently the docker pr uses mysql 5.5 so it does not suffer from the 5.6-5.7 issues. Just a thought.

sushantdhiman:
@toddbluhm I will work on the remaining fixes (just one) have them merged asap. Then you can rebase your branch with updated master :)

toddbluhm:
Okay sounds good, and no need to rush. I was just trying to help lighten your load :) I will close this out then.

sushantdhiman:
PR https://github.com/sequelize/sequelize/pull/6479

PR #6475, opened 2016-08-20
Fix issue for null id of AI index bulkCreate

frogcjn:
[mySQL]  fix issue for null id of AI index bulkCreate

### DIFF
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -92,7 +92,7 @@ class Query extends AbstractQuery {
       this.handleInsertQuery(data);
 
       if (!this.instance) {
-        // handle bulkCreate AI primiary key
+        // handle bulkCreate AI primiary key 
         if (
           data.constructor.name === 'ResultSetHeader'
           && this.model
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -92,12 +92,14 @@ class Query extends AbstractQuery {
       this.handleInsertQuery(data);
 
       if (!this.instance) {
-        if (data.constructor.name === 'ResultSetHeader'
+        // handle bulkCreate AI primiary key
+        if (
+          data.constructor.name === 'ResultSetHeader'
           && this.model
           && this.model.autoIncrementField
           && this.model.autoIncrementField === this.model.primaryKeyAttribute
           && this.model.rawAttributes[this.model.primaryKeyAttribute]
-        ){
+        ) {
           const startId = data[this.getInsertIdField()];
           result = [];
           for (let i = startId; i < startId + data.affectedRows; i++) {
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -92,8 +92,7 @@ class Query extends AbstractQuery {
       this.handleInsertQuery(data);
 
       if (!this.instance) {
-        if (
-          data.constructor.name === 'ResultSetHeader'
+        if (data.constructor.name === 'ResultSetHeader'
           && this.model
           && this.model.autoIncrementField
           && this.model.autoIncrementField === this.model.primaryKeyAttribute
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -92,11 +92,13 @@ class Query extends AbstractQuery {
       this.handleInsertQuery(data);
 
       if (!this.instance) {
-        if (data.constructor.name === 'ResultSetHeader'
+        if (
+          data.constructor.name === 'ResultSetHeader'
           && this.model
           && this.model.autoIncrementField
           && this.model.autoIncrementField === this.model.primaryKeyAttribute
-          && this.model.rawAttributes[this.model.primaryKeyAttribute]){
+          && this.model.rawAttributes[this.model.primaryKeyAttribute]
+        ){
           const startId = data[this.getInsertIdField()];
           result = [];
           for (let i = startId; i < startId + data.affectedRows; i++) {
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -6,7 +6,7 @@ const AbstractQuery = require('../abstract/query');
 const uuid = require('node-uuid');
 const sequelizeErrors = require('../../errors.js');
 const _ = require('lodash');
-
+f
 class Query extends AbstractQuery {
   constructor(connection, sequelize, options) {
     super();
@@ -92,7 +92,15 @@ class Query extends AbstractQuery {
       this.handleInsertQuery(data);
 
       if (!this.instance) {
-        result = data[this.getInsertIdField()];
+        if (data.constructor.name === "ResultSetHeader"){
+          const startId = data[this.getInsertIdField()]
+          result = []
+          for(let i = startId; i< startId + data.affectedRows; i++) {
+            result.push({ [this.model.rawAttributes[this.model.primaryKeyAttribute].field]: i })
+          }
+        } else {
+          result = data[this.getInsertIdField()];
+        }
       }
     }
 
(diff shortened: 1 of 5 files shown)


lib/dialects/mysql/query.js
```diff
@@ -92,7 +92,15 @@ class Query extends AbstractQuery {
       this.handleInsertQuery(data);
 
       if (!this.instance) {
-        result = data[this.getInsertIdField()];
+        if (data.constructor.name === "ResultSetHeader"){
+          const startId = data[this.getInsertIdField()]
+          result = []
+          for(let i = startId; i< startId + data.affectedRows; i++) {
```
felixfbecker: Missing space after `for`

lib/dialects/mysql/query.js
```diff
@@ -92,7 +92,15 @@ class Query extends AbstractQuery {
       this.handleInsertQuery(data);
 
       if (!this.instance) {
-        result = data[this.getInsertIdField()];
+        if (data.constructor.name === "ResultSetHeader"){
```
felixfbecker: Is there a reason we cannot check with `instanceof` here?

felixfbecker:
Can you please link the issue this solves? Is there a test case?

frogcjn:
@felixfbecker  which file should I add this test to?

lib/dialects/mysql/query.js
frogcjn: if I use instanceof, I should import the class.

felixfbecker:
If it is possible to assert on the generated SQL, under `unit/`. Otherwise under `integration/`. Try to put it where similar cases are tested (model / creation).

lib/model.js
```diff
@@ -2267,7 +2267,7 @@ class Model {
               instance.dataValues[this.rawAttributes[attr].field] !== undefined &&
               this.rawAttributes[attr].field !== attr
           ) {
-            instance.set(attr, instance.dataValues[this.rawAttributes[attr].field]);
+            instance.dataValues[attr] = instance.dataValues[this.rawAttributes[attr].field;
```
felixfbecker: This will bypass custom getters/setters. Is this intended?

lib/dialects/mysql/query.js
```diff
@@ -92,7 +92,19 @@ class Query extends AbstractQuery {
...
       if (!this.instance) {
-        result = data[this.getInsertIdField()];
+        if (data.constructor.name === 'ResultSetHeader'
+          && this.model
+          && this.model.autoIncrementField
+          && this.model.autoIncrementField === this.model.primaryKeyAttribute
+          && this.model.rawAttributes[this.model.primaryKeyAttribute]){
+          const startId = data[this.getInsertIdField()];
+          result = [];
+          for (let i = startId; i< startId + data.affectedRows; i++) {
```
felixfbecker: Space missing before `<`

lib/model.js
frogcjn: But if bypass, the isNewRecord will affect `createdAt` not been set.

frogcjn: Both create/bulkCreate is already set in constructor

lib/dialects/mysql/query.js
```diff
@@ -92,7 +92,20 @@ class Query extends AbstractQuery {
       this.handleInsertQuery(data);
 
       if (!this.instance) {
-        result = data[this.getInsertIdField()];
+        if (data.constructor.name === 'ResultSetHeader'
+          && this.model
+          && this.model.autoIncrementField
+          && this.model.autoIncrementField === this.model.primaryKeyAttribute
+          && this.model.rawAttributes[this.model.primaryKeyAttribute]
+        ){
```
felixfbecker: space missing here

frogcjn:
@felixfbecker I don't know why mssql is fail. Is something broken?

felixfbecker:
Yeah coverage report is broken since a week or so.

frogcjn:
@felixfbecker , fixed sqlite and mysql, but not sure why postgres-native deadlock.

frogcjn:
@felixfbecker , it passed all test except COVERAGE mssql

frogcjn:
@felixfbecker , it passed all test except COVERAGE mssql

felixfbecker:
LGTM

lib/dialects/mysql/query.js
```diff
@@ -92,7 +92,22 @@ class Query extends AbstractQuery {
...
+        if (
+          data.constructor.name === 'ResultSetHeader'
+          && this.model
+          && this.model.autoIncrementField
+          && this.model.autoIncrementField === this.model.primaryKeyAttribute
+          && this.model.rawAttributes[this.model.primaryKeyAttribute]
+        ) {
+          const startId = data[this.getInsertIdField()];
+          result = [];
+          for (let i = startId; i < startId + data.affectedRows; i++) {
```
mickhansen: Is manually calculating the ids after `last_insert_id ` the generally acceptable way of figuring out multi insert ids?
If so i'd like some internal references. This solution also fails to take into account that the increment interval can be different from 1.

lib/dialects/mysql/query.js
```diff
@@ -92,7 +92,22 @@ class Query extends AbstractQuery {
       this.handleInsertQuery(data);
 
       if (!this.instance) {
-        result = data[this.getInsertIdField()];
+        // handle bulkCreate AI primiary key 
+        if (
+          data.constructor.name === 'ResultSetHeader'
```
mickhansen: Motivation for checking the constructor name? Doesn't this tie us even more to the underlying library, does it offer any benefits?

lib/dialects/mysql/query.js
felixfbecker: There are some notes in the docs
http://dev.mysql.com/doc/refman/5.6/en/information-functions.html#function_last-insert-id

- LAST_INSERT_ID is per-connection
- It always returns the _first_ ID of a bulk insert, "to make it possible to reproduce easily the same INSERT statement against some other server"

Sorry that I overlooked this when reviewing, but I'm also on the fence whether this is a good idea after all. The answer you generally find on the internet is "not possible in MySQL".

felixfbecker: Found more reference:
http://dev.mysql.com/doc/refman/5.7/en/innodb-auto-increment-handling.html
> With innodb_autoinc_lock_mode set to 0 (“traditional”) or 1 (“consecutive”), the auto-increment values generated by any given statement will be consecutive, without gaps, because the table-level AUTO-INC lock is held until the end of the statement, and only one such statement can execute at a time.

felixfbecker: But the case where increment is not 1 remains.

lib/dialects/mysql/query.js
felixfbecker: https://github.com/sequelize/sequelize/pull/6475#discussion_r75575249
It is easier than importing the class as the library is loaded lazily and I think we would have to do a require into one of the libraries internal files to get a reference.

mickhansen: Why do we need to check the type of data at all though? Hasn't been needed previously

lib/dialects/mysql/query.js
mickhansen: Non-1 increment is rare i suppose. InnoDB is usually the storage engine of choice.
What about SQLite? And how do we test that this holds up under concurrency (i'd imagine it does though)

mickhansen:
@jaredjenkins It's already in a release, look at the commit tags: https://github.com/sequelize/sequelize/commit/b0ae95b0d58dcfeefe4df1272e87b043d8d4eeea

The maintainers are not going to be backporting features, only critical bugfixes.

felixfbecker:
@jaredjenkins v4 will be released soon.

PR #6476, opened 2016-08-20
add test for "should return autoIncrement primary key"

frogcjn:
add test for "should return autoIncrement primary key"

### DIFF
--- a/test/integration/model/create.test.js
+++ b/test/integration/model/create.test.js
@@ -1852,6 +1852,33 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         expect(m.secret).to.be.eql(M2.secret);
       });
     });
+    it('should return autoIncrement primary key', function() {
+      var Maya = this.sequelize.define('Maya', {
+        name: Sequelize.STRING,
+        id: {
+          type: Sequelize.INTEGER,
+          autoIncrement: true,
+          primaryKey: true
+        },
+        createdAt: {
+            field: 'created_at',
+            type: Sequelize.DATE
+        },
+        updatedAt: {
+            field: 'updated_at',
+            type: Sequelize.DATE
+        }
+      });
+
+      var M1 = { name: 'Prathma Maya', secret: 'You are on list #1'};
+      var M2 = { name: 'Dwitiya Maya', secret: 'You are on list #2'};
+
+      return Maya.sync({ force: true }).then(() => Maya.bulkCreate([M1, M2]))
+      .then((ms) => {
+        expect(ms[0].id).to.be.eql(1);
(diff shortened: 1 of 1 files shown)


test/integration/model/create.test.js
```diff
@@ -1852,6 +1852,33 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
+          primaryKey: true
+        },
+        createdAt: {
+            field: 'created_at',
+            type: Sequelize.DATE
+        },
+        updatedAt: {
+            field: 'updated_at',
+            type: Sequelize.DATE
+        }
```
felixfbecker: Are all those column definitions needed to reproduce the bug? Or can we omit createdAt/updatedAt and maybe even id?

test/integration/model/create.test.js
```diff
@@ -1852,6 +1852,19 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         expect(m.secret).to.be.eql(M2.secret);
       });
     });
+    it('should return autoIncrement primary key', function() {
+      var Maya = this.sequelize.define('Maya', {
+      });
```
felixfbecker: Put it in one line please :)

frogcjn:
@felixfbecker
It looks the problem not only appeared in MySQL

frogcjn:
It looks like the problem not only appeared on MySQL
@felixfbecker

felixfbecker:
Yes, SQLite has the exact same bug apparently. Can you include a fix for it aswell?

test/integration/model/create.test.js
```diff
@@ -723,6 +723,19 @@ describe(Support.getTestDialectTeaser('Model'), function() {
             });
           });
         });
+        
+        it('should make the autoincremented values available on the returned instances with bulkCreate', function () {
+          var User = this.sequelize.define('user', {});
+
+          return User.sync({force: true}).then(function () {
+            return User.bulkCreate([{}, {}], {returning: true}).then(function (users) {
+              expect(users[0].get('id')).to.be.ok;
```
felixfbecker: You can omit the `to.be.ok` if you check for equality :)

felixfbecker:
The test is not failing anymore

frogcjn:
@felixfbecker So the ok is required.

frogcjn:
@felixfbecker , the test passed because it only test current.dialect.supports.returnValues

so any dialect does not support returnValues will fail the test.

felixfbecker:
LGTM

felixfbecker:
Wasn't this test included in the PR with the fix?

felixfbecker:
Linking https://github.com/sequelize/sequelize/pull/6475

PR #6478, opened 2016-08-20
Use new SQL Server for tests

sdepold:
I need to shut down the current SQL Server instance and created a new one on the same EC2 region. Since I always forget which steps are needed to set up such a machine, I also added some instructions about that.


sdepold:
If the tests for MSSQL pass and you are +1 on it, I will just change the mssql.sequelizejs.com domain to pinpoint the new URI and change the PR one more time to only contain the instructions.

sushantdhiman:
@sdepold We should keep these steps in Wiki, away from the regular docs stuff. I can move these to a Wiki article if thats ok ?

sdepold:
Good idea!

sdepold:
Can anyone tell why the tests are failing?

janmeier:
A bug in the latest node 6 release affecting tedious

sdepold:
So it's not because of my changes, correct? Will change the dns settings then! :)

PR #6490, opened 2016-08-24
Fix port for MSSQL server

sdepold:


sdepold:
OK. Something went wrong -.- Will look into it later. Sorry for the hassles

janmeier:
Nope, only node6 failing, because of a node6 + tedious bug ;)

PR #6498, opened 2016-08-25
Critical bug fix for Model#count()

rivierasolutions:
`Model#count()` will now return a proper number when the query object has an `include` field. Previously the result was `NaN`.
This breaks our production code, please respond ASAP!

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -672,7 +672,7 @@ class QueryInterface {
         return data;
       }
 
-      let result = data ? data[attributeSelector] : null;
+      let result = data ? (data[attributeSelector] === undefined ? data.dataValues[attributeSelector] : data[attributeSelector]) : null;
 
       if (options && options.dataType) {
         const dataType = options.dataType;


felixfbecker:
I'm sorry to hear that this breaks your production code, but before this can be merged we need
 - a test that fails before the fix and passes afterwards to prevent regressions
 - a changelog entry of what was fixed

mickhansen:
You'll likely also want to target the v3 branch aswell so we can release a 3.x.x fix

rivierasolutions:
After further investigating this issue it turns out that the problem is elsewhere. 
Consider the test 
test/integration/model/count.test.js -> count -> should support include
Providing this query to `this.User.count()`
```
{
     raw: false,
     include: [{
          model: this.Project,
          where: {
            name: 'project1'
          }
     }]
}
```
yields NaN and causes the test to fail. The difference is `raw: false`. This is consistent for both `mysql` and `sqlite`.

janmeier:

PR #6500, opened 2016-08-25
SQLite JSON datatype support

### LINKED ISSUE #4873: JSON type should stringify/parse in mysql/mssql/sqlite

arcanis:
### Pull Request check-list
- [x] Does your issue contain a link to existing issue (Closes #4873)
- [x] ~~Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?~~
### Description of change

Partial fix for #4873 (cc @mickhansen) - only add JSON support to SQLite, because I don't use MySQL and MSSQL, and I feel much less confident writing tests for these environments (especially since I won't be able to see it myself if something is wrong).

I haven't been able to fix a test, because I'm not familiar enough with how `CREATE INDEX` queries should look like :disappointed: Some help on this would be great here.

I wasn't sure if JSONB should also be implemented or if there was too much differences for it to matter, so I figured I should first submit this PR and ask for your input.

Due to the previous point, I splitted some tests in order for a dialect to be able to run JSON-related tests but not JSONB-related tests nor ARRAY-related tests.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
 # Future
 - [FIXED] Accept dates as string while using `typeValidation` [#6453](https://github.com/sequelize/sequelize/issues/6453)
+- [ADDED] Sqlite JSON datatype support ([#6500](https://github.com/sequelize/sequelize/pull/6500))
 
 # 4.0.0-1
 - [CHANGED] Removed `modelManager` parameter from `Model.init()` [#6437](https://github.com/sequelize/sequelize/issues/6437)
--- a/lib/dialects/sqlite/data-types.js
+++ b/lib/dialects/sqlite/data-types.js
@@ -22,7 +22,18 @@ module.exports = BaseTypes => {
   BaseTypes.ENUM.types.sqlite = false;
   BaseTypes.REAL.types.sqlite = ['REAL'];
   BaseTypes.DOUBLE.types.sqlite = ['DOUBLE PRECISION'];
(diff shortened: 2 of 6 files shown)


lib/data-types.js
```diff
@@ -465,7 +465,7 @@ DATE.prototype._stringify = function _stringify(date, options) {
 function DATEONLY() {
   if (!(this instanceof DATEONLY)) return new DATEONLY();
 }
-util.inherits(DATEONLY, ABSTRACT);
+inherits(DATEONLY, ABSTRACT);
```
felixfbecker: ?

felixfbecker: Is this needed for the JSON to work? Could you do this in a seperate PR? The commits will get squashed

arcanis: No, it's not needed at all, I've simply noticed it hadn't been changed like other inherits calls. I've removed its commit from this PR.

wmertens:
@arcanis why did you change the tests to `JSONB`? Sqlite doesn't support that?

arcanis:
I have no use for this PR any longer, should I close it? @mickhansen

rokoroku:
@arcanis May I continue your work? 
Maybe I can add some features to query-generators and some tests  ...

arcanis:
@rokoroku For sure, go for it :) I'm not using sequelize right now, so I can't really champion this feature anymore.

sushantdhiman:
New implementation continued in https://github.com/sequelize/sequelize/pull/7094

PR #6505, opened 2016-08-27
Diff on save

felixfbecker:
The first part of #6270 
- [x] `_previousDataValues` is set to a deep clone of `dataValues`
- [x] `changed()` does deep change detection with `_.isEqual`
- [ ] `isNewRecord: false` should initialize `_previousDataValues`

Don't know why tests fail, help appreciated
### Thoughts
- When doing `new Model({what: 'ever'}, isNewRecord: false})` `_previousDataValues` should probably be set to `{what: 'ever'}`.
- Should there be a method to take a snapshot of the current state?

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3556,9 +3556,16 @@ class Model {
         values[field] = new Date();
         where[field] = attribute.hasOwnProperty('defaultValue') ? attribute.defaultValue : null;
 
-        this.setDataValue(field, values[field]);
-
-        return this.sequelize.getQueryInterface().update(this, this.constructor.getTableName(options), values, where, _.defaults({ hooks: false }, options));
+        return this.sequelize.getQueryInterface().update(
+          this,
+          this.constructor.getTableName(options),
+          values,
+          where,
+          _.defaults({ hooks: false }, options)
+        ).then(() => {
+          this.setDataValue(field, values[field]);
+          this._previousDataValues[field] = values[field];
+        });
       } else {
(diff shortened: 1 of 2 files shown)


mickhansen:
LGTM, it's a minor perf reducement, but it fixes JSON so i'll have to concede that it's preferable :)
/cc @janmeier

mickhansen:
For `new/Build` either from user provided or from database loaded values `previous` should be null`.
`snapshot` sounds good but it would be pretty simple to do in user land too.

felixfbecker:
Sure about `previous` being null? that would mean we would have to add extra checks everywhere to avoid Cannot read property x of null errors.

Or let me ask it this way: if you do `new Model({what: 'ever'}, isNewRecord: false})`, what should `model.changed()` return? Imo `false`. It should return `true` once you actually change something. So we have to initialize `_previousDataValues`.

felixfbecker:
Also, will run a benchmark tomorrow to see the impacts.
We will likely gain a bit of performance back though when we put `dataValues` on the instance directly because we don't need setters/getters anymore.

mickhansen:
No, you are right of course, when ` changed` is altered to do diffs `previous` will have to equal the data values on the initial build.

janmeier:
LGTM

felixfbecker:
@janmeier need to fix tests first, couldn't find the time yet.

SimonSchick:
Closing this due to age and merge conflicts, feel free to rebase/refresh this PR and re-open.

Issue #6504, opened 2016-08-27
MSSQL broken build, lock Node version at 6.3?


felixfbecker:
The broken build is coming from https://github.com/nodejs/node/issues/8127, introduced in 6.4.0.
This will be fixed in Node 6.5 as the fix landed in master, but in the meantime, the constant broken build in every PR is getting annoying. Should we lock the Node version at ^6.3 until 6.5 is out?

mickhansen:
Locking sounds good to me

PR #6507, opened 2016-08-27
Remove Babel devDependency

felixfbecker:
https://github.com/sequelize/sequelize/pull/6493#issuecomment-242351497

> @janmeier I don't know why the test needs a babel dep though. The issue you linked was that `sequelize.import()` did not support models exported as a `default` export instead of `module.exports`. If we need a test at all for this, it should be enough to import a simple model that is exported with `exports.default`. The change in Babel was more a fix than a change, previously it was not conforming to the ES6 spec. The issue therefor is more about ES6 support for model files and not Babel support - TypeScript uses ES6 modules aswell, and I don't think we should test integration with every possible transpiler out there.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -401,9 +401,9 @@ class Sequelize {
 
     if (!this.importCache[path]) {
       let defineCall = arguments.length > 1 ? arguments[1] : require(path);
-      if (typeof defineCall === 'object' && defineCall.__esModule) {
-        // Babel/ES6 module compatability
-        defineCall = defineCall['default'];
+      if (typeof defineCall === 'object') {
+        // ES6 module compatability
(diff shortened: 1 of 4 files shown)


mickhansen:
MSSQL tests keep failing (3-4 retries now), unsure why: https://travis-ci.org/sequelize/sequelize/jobs/155622204

felixfbecker:
@mickhansen this problem is around for a couple of weeks now. See https://github.com/sequelize/sequelize/issues/6504. Node introduced a bug in 6.4 that broke tedious. 6.3.x is working fine and the fix is already on master, but will probably land in 6.5. I suggest that we lock Travis node version on 6.3, as otherwise we cannot test MSSQL against Node 6 _at all_.

PR #6509, opened 2016-08-28
Lock Node version at 6.3

felixfbecker:
See #6504


felixfbecker:
Tests passed, but MSSQL timed out because the server seems to be super-slow recently.

PR #6512, opened 2016-08-29
include: expand alias into association (update 2)

### LINKED ISSUE #970: Include by Association Alias

swarthy:
### Pull Request check-list

### Description of change

Expand string or `{association: <string>}` in `include` option into the association
(Closes #970)

### DIFF
--- a/docs/docs/models-usage.md
+++ b/docs/docs/models-usage.md
@@ -523,6 +523,26 @@ User.findAll({ include: ['Instruments'] }).then(function(users) {
     }]
   */
 })
+
+User.findAll({ include: [{ association: 'Instruments' }] }).then(function(users) {
+  console.log(JSON.stringify(users))
+
+  /*
+    [{
+      "name": "John Doe",
+      "id": 1,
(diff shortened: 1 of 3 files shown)


felixfbecker:
LGTM

Issue #6116, opened 2016-06-15
Prepare for coroutines


overlookmotel:
Now that we're all ES6 for Sequelize v4, the door is open to be using coroutines. Personally, I think this is a great thing - it'll simplify Sequelize's codebase a _lot_.

However, I don't think we're quite ready to do that - I'd like to conclude my research on bluebird and CLS, to make absolutely sure that introducing coroutines won't break CLS support, before we take the plunge.

But when we do start using coroutines, it will be a semver major change, due to a change in the way errors are handled.
#### Error handling

At present some Sequelize async methods throw synchronously if the arguments they are passed are illegal in some way. Once those methods become coroutines, this wouldn't be the case any more - they'll return rejected promises instead.

How it is now:

``` js
Model.findAll = function(options) {
    if (!options) throw new Error('You must provide options!');
}

// Model.findAll() throws synchronously
try {
   Model.findAll();
} catch (err) {
   console.log('It throws!');
}
```

How it will be with coroutines:

``` js
Model.findAll = co.wrap( function*(options) {
    if (!options) throw new Error('You must provide options!');
} );

// Model.findAll() returns a rejected promise
Model.findAll().catch( function(err) {
    console.log('It returns a rejected promise');
} );
```
#### A proposal to "prepare the way"

I would like to see coroutines in Sequelize as soon as possible, but I don't want to hold up the release of v4 until the CLS issues are resolved.

I propose that we "prepare the way" now by converting all Sequelize methods from throwing synchronously to returning rejected promises.

This isn't hard to do using `Promise.method()`:

``` js
Model.findAll = Promise.method( function(options) {
    if (!options) throw new Error('You must provide options!');
} );

// Model.findAll() returns a rejected promise
Model.findAll().catch( function(err) {
    console.log('It returns a rejected promise');
} );
```

This way:
1. The whole API switches away from sync throwing in one go.
2. It can all be done as a single change in Sequelize v4.0.0.
3. Methods can then be refactored with coroutines as and when we feel like it, without creating any further breaking changes.
4. We won't have to wait until Sequelize v5 to start using coroutines.

What do you think?

mickhansen:
We're in 4.0.0-pre so it's the time for breaking changes!
4.0.0 is about cleaning up all the code we want so we can be more productive.

overlookmotel:
@mickhansen So that's a yes? :)

overlookmotel:
[...]

PR #6556, opened 2016-09-07
CLS support with `cls-bluebird` module

overlookmotel:
### Pull Request check-list

### Description of change

As discussed in #6116, new version of [cls-bluebird](https://www.npmjs.com/package/cls-bluebird) was recently released. The new version is very robust, with over 100,000 tests assaulting it, and also supports CLS with coroutines via a patch to `Bluebird.coroutine()`.

This PR removes the previous CLS patch to `Sequelize.Promise` and substitutes `cls-bluebird`.

In order to do this, I had to make a few other changes:

* `Sequelize.Promise` remains an independent instance of Bluebird
* No CLS patch is applied to `Sequelize.Promise` until the user instructs Sequelize to use CLS (which should improve performance for anyone who doesn't use CLS)
* CLS is activated by calling `Sequelize.useCls( ns )`
* Setting CLS namespace directly with `Sequelize.cls = ...` still works but is deprecated
* CLS namespace is stored in `Sequelize._cls`

In the process of refactoring `sequelize.transaction()`, I've removed the necessity for the `autoCallback` function to return a promise. It can now also synchronously return any value. So I've removed the "autoCallback errors when no promise chain is returned" test. Was there a reason for having this restriction?


lib/sequelize.js
```diff
@@ -974,41 +975,61 @@ class Sequelize {
...
+    // autoCallback provided
+    const wrapper = ns ? function(fn) {
+      var promise;
+      ns.run(() => promise = fn());
+      return promise;
+    } : function(fn) { return fn(); };
+
+    return wrapper(() =>
+      transaction.prepareEnvironment().then(() =>
+        autoCallback(transaction)
```
mickhansen: Please use a single line when using the short hand for return, that way one doesn't accidently screw everything up by adding a log statement thinking that's valid.

```js
transaction.prepareEnvironment().then(() => autoCallback(transaction))
```

lib/sequelize.js
```diff
@@ -974,41 +975,61 @@ class Sequelize {
...
+    const ns = Sequelize._cls;
+
+    if (!autoCallback) return transaction.prepareEnvironment().return(transaction);
+
+    // autoCallback provided
+    const wrapper = ns ? function(fn) {
+      var promise;
+      ns.run(() => promise = fn());
+      return promise;
+    } : function(fn) { return fn(); };
```
mickhansen: Could the wrapper be defined outside transaction and simply check for Sequelize._cls? We'll save a function allocation for each transaction.

lib/sequelize.js
```diff
@@ -974,41 +975,61 @@ class Sequelize {
...
+    // patch bluebird to bind all promise callbacks to CLS namespace
+    clsBluebird(ns, Promise);
+
+    // return Sequelize for chaining
+    return this;
+  }
+
+  static get cls() {
+    return this._cls;
+  }
```
mickhansen: This does add some overhead to each transaction rather than just using a static - But i suppose it's likely negligible.

felixfbecker: I don't get this. You make a private static property `_cls`, but then expose it as a public static property `cls` through a (slow) getter?

overlookmotel: Idea is that `Sequelize.cls` should not be used. But I put the getter/setter in there so as not to break compatibility. If you do `Sequelize.cls = ns` rather than `Sequelize.useCls(ns)` it works, but gives a deprecation warning.

I thought it was nice to maintain the old API for this version (but deprecated) rather than just strip it out entirely.

I could take it out, or also add a deprecation warning to the getter.

@mickhansen `Sequelize._cls` is used everywhere internally in Sequelize, so the extra overhead of the getter should never be incurred.

lib/sequelize.js
overlookmotel: Yes, good idea. I'll make this change.

lib/sequelize.js
overlookmotel: Hmm yeah. First time I've used ES6 short function notation. I normally write with backward compat to node 0.12 in mind.

It's tricky because since none of that block has any returns at all, taking that logic to its furthest end, you'd end up with the entire block being a single long line of code. But I see your point.

I've added some more return statements. They're unnecessary, but I think it preserves readability while avoiding the problem you raise.

lib/sequelize.js
mickhansen: @overlookmotel roger!

lib/sequelize.js
```diff
@@ -952,7 +953,7 @@ class Sequelize {
    * const cls = require('continuation-local-storage');
    * const ns = cls.createNamespace('....');
    * const Sequelize = require('sequelize');
-   * Sequelize.cls = ns;
+   * Sequelize.useCls(ns);
```
mickhansen: `useCLS`?

overlookmotel:
@mickhansen @felixfbecker Thanks for reviewing this. I've made all the changes you suggest, apart from the `Sequelize.cls` getter where I've added a deprecation warning.

Please have a look at the updated version when you get a chance.

lib/sequelize.js
overlookmotel: Yes. New API. Where before you would instruct Sequelize to use a CLS namespace with `Sequelize.cls = ns`, now it's `Sequelize.useCls(ns)`.

Two reasons for this:

1. Necessity: `cls-bluebird` requires the CLS namespace to be provided at the time it applies the patch. So the patch can't be applied statically. The user needs to set the CLS namespace and *then* the patch is applied.

2. Advantage: the cls-bluebird patch is only applied to `Sequelize.Promise` when the user calls `.useCls()`. So for the majority of people who don't use CLS, they avoid the overhead of the patch which should improve performance.

overlookmotel: It would be possible to keep the old API, but `Sequelize.cls` would have to become a setter, so I figured better to make it explicit.

But happy to go with what you think is best.

mickhansen: I was simply suggesting that CLS should be capitalized :)

overlookmotel:
Bloody Travis keeps failing, but each time on a different dialect and test.

lib/sequelize.js
overlookmotel: Done!

mickhansen:
LGTM, just need the original CLS man @janmeier to sign off.

janmeier:
Looks mighty fine to me! Just needs an update to https://github.com/sequelize/sequelize/blob/master/docs/docs/transactions.md then we're good to go

overlookmotel:
Hooray!

@janmeier OK I'll do the docs update. But may not get to it until tomorrow.

janmeier:
I'm glad I wasn't the one to write 10k test for cls-bluebird, but I <3 @overlookmotel for doing so!

overlookmotel:
@janmeier 100k tests, actually! Yes it was a bit of a mad mission. But I just want the damn thing to work properly!

Thanks loads for merging.

Any chance you could push to npm at this point. I'd like to start road-testing it asap.

PR #6557, opened 2016-09-07
Updated .npmignore

sushantdhiman:


felixfbecker:
I would leave the eslintrc in, sometimes you step into library code with a debugger and it prevents Ealing from putting red swiggles everywhere if your project is using a different style. Otherwise, LGTM.

PR #6560, opened 2016-09-07
Grouped limit include

janmeier:
This code is the first step towards better support for BTM associations in https://github.com/mickhansen/dataloader-sequelize/

TL;DR - Dataloader batches multiple simultaneous calls to `BTM.get` e.g. `project.getMembers()`, into a single query. This is similar to what `include.separate` does for hasMany, with some extra smarts.

I've changed the `groupedLimit.on` code to accept an an array of `[association, string]` to signify that reach row should be joined. We need this to support BTM, because we need to filter on the join table. In the project <-> users example, we already have a list of project ids, and we want to find a limited amount of matching users for each project.

This should also be the first step towards `include.separate` support for belongs to many.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1386,46 +1386,97 @@ const QueryGenerator = {
         if (!mainTableAs) {
           mainTableAs = table;
         }
+
+        const where = Object.assign({}, options.where);
+        const order = options.order;
+        let whereKey
+          , include
+          , groupedTableName = mainTableAs;
+
+        if (typeof options.groupedLimit.on === 'string') {
+          whereKey = options.groupedLimit.on;
+        } else if (options.groupedLimit.on instanceof HasMany) {
+          whereKey = options.groupedLimit.on.foreignKeyField;
+        }
+
+        if (options.groupedLimit.on instanceof BelongsToMany) {
+          // BTM includes needs to join the through table on to check ID
+          groupedTableName = options.groupedLimit.on.throughModel.tableName;
+          const groupedLimitOptions = Model._validateIncludedElements({
+            include: [{
+              association: options.groupedLimit.on.manyFromSource,
+              duplicating: false, // The UNION'ed query may contain duplicates, but each sub-query cannot
+              required: true,
+              where: {
+                '$$PLACEHOLDER$$': true
+              }
+            }],
+            model
+          });
+
+          // Make sure attributes from the join table are mapped back to models
+          options.hasJoin = true;
+          options.hasMultiAssociation = true;
+          options.includeMap = Object.assign(groupedLimitOptions.includeMap, options.includeMap);
+          options.includeNames = groupedLimitOptions.includeNames.concat(options.includeNames || []);
+          include = groupedLimitOptions.include;
+
+          if (Array.isArray(options.order)) {
+            // We need to make sure the order by attributes are available to the parent query
+            options.order.forEach((order, i) => {
+              if (Array.isArray(order)) {
+                order = order[0];
+              }
+
+              // Otherwise, alias it
+              const alias = `subquery_order_${i}`;
+              options.attributes.push([order, alias]);
+
+              if (Array.isArray(options.order[i])) {
+                options.order[i][0] = alias;
+              } else {
+                options.order[i] = alias;
+              }
+            });
+          }
+        } else {
+          // Ordering is handled by the subqueries, so ordering the UNION'ed result is not needed
+          delete options.order;
+          where['$$PLACEHOLDER$$'] = true;
+        }
+
+        // Caching the base query and splicing the where part into it is consistently > twice
+        // as fast than generating from scratch each time for values.length >= 5
+        const baseQuery = '('+this.selectQuery(
+          tableName,
+          {
+            attributes: options.attributes,
+            limit: options.groupedLimit.limit,
+            order,
+            where,
+            include,
+            model
+          },
+          model
+        ).replace(/;$/, '')+')';
+        const placeHolder = this.whereItemQuery('$$PLACEHOLDER$$', true, { model });
+        const splicePos = baseQuery.indexOf(placeHolder);
+
         mainQueryItems.push(this.selectFromTableFragment(options, model, mainAttributes, '('+
           options.groupedLimit.values.map(value => {
-            let where = options.where || {}
-              , whereKey
-              , include;
-
-            if (typeof options.groupedLimit.on === 'string') {
-              whereKey = options.groupedLimit.on;
-            } else if (options.groupedLimit.on instanceof HasMany) {
-              whereKey = options.groupedLimit.on.foreignKeyField;
-            } else if (options.groupedLimit.on instanceof BelongsToMany) {
-              include = Model._validateIncludedElements({
-                include: [{
-                  association: options.groupedLimit.on.manyFromSource,
-                  where: {
-                    [options.groupedLimit.on.otherKey]: value
-                  },
-                  duplicating: false
-                }],
-                model
-              }).include;
-            }
-
+            let groupWhere;
             if (whereKey) {
-              where = Object.assign({}, where);
-              where[whereKey] = value;
+              groupWhere = {
+                [whereKey]: value
+              };
+            }
+            if (include) {
+              groupWhere = {
+                [options.groupedLimit.on.otherKey]: value
+              };
             }
 
-            return '('+this.selectQuery(
-              tableName,
-              {
-                attributes: options.attributes,
-                limit: options.groupedLimit.limit,
-                order: options.order,
-                where,
-                include,
-                model
-              },
-              model
-            ).replace(/;$/, '')+')';
+            return Utils.spliceStr(baseQuery, splicePos, placeHolder.length, this.getWhereConditions(groupWhere, groupedTableName));
           }).join(
             this._dialect.supports['UNION ALL'] ?' UNION ALL ' : ' UNION '
           )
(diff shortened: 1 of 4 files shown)


janmeier:
Currently the code is not exposed anywhere, so users have to add the `groupedLimit` to a top level options object to use it. However, it can be used for `include.separate` for BTM, similar to [here](https://github.com/sequelize/sequelize/blob/6d90b5ba66a4ae44cd5fb81cf0a7288115ca4482/lib/associations/has-many.js#L277-L285).

In the future I'd also like to extend `groupedLimit` to use lateral joins (postgres) and cross apply (mssql)

mickhansen:
@janmeier this needs a v3 target aswell right?

lib/dialects/abstract/query-generator.js
```diff
@@ -1386,16 +1386,34 @@ const QueryGenerator = {
...
+                  where: {
+                    [options.groupedLimit.on[1]]: value
+                  }
+                }],
+                model
+              }).include;
+            } else {
+              where = _.assign(options.where);
+              where[options.groupedLimit.on] = value;
+            }
```
mickhansen: `on` is used only for defining a where statement.
Rather than having `on` as an array could we not just support having it null and then defining an arbitrary include which could have required + where?

janmeier:
> @janmeier this needs a v3 target aswell right?

It does indeed, I'll backport once we agree on the API

lib/dialects/abstract/query-generator.js
janmeier: We can't do an include on the main query (since a main include should be added to the main query :P)

We could add `groupedLimit.include`, but then we would have to do `options.groupedLimit.include[0].where[options.groupedLimit.on] = value` in the loop. Not sure I think thats prettier.

mickhansen: But there isn't really a main query with groupedLimit, or is it because how we handle it so it becomes a UNION with an include and then an outer include?

No, not sure it's prettier either, maybe have `on` be `{association, field}` rather than an array.

test/unit/sql/select.test.js
```diff
@@ -80,6 +80,49 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
...
+          values: [
+            1,
+            5
+          ]
+        }
+      }, {
+        default: 'SELECT [user].* FROM ('+
+          [
+            '(SELECT [user].[id_user] AS [id] FROM [users] AS [user] INNER JOIN [project_user] AS [projects] ON [user].[id_user] = [projects].[userId] AND [projects[.[companyId] = 1 ORDER BY [user].[last_name] ASC'+sql.addLimitAndOffset({ limit: 3, order: ['last_name', 'ASC'] })+')',
+            '(SELECT [user].[id_user] AS [id] FROM [users] AS [user] INNER JOIN [project_user] AS [projects] ON [user].[id_user] = [projects].[userId] AND [projects[.[companyId] = 5 ORDER BY [user].[last_name] ASC'+sql.addLimitAndOffset({ limit: 3, order: ['last_name', 'ASC'] })+')'
```
mickhansen: > [projects[.[companyId]

lib/dialects/abstract/query-generator.js
janmeier: We have tests for grouped limit with top level include though https://github.com/sequelize/sequelize/blob/080b3f8511027107a6dbaeeb0f3ee4a22451c101/test/unit/sql/select.test.js#L174 (Not saying its ever used in pratice, but we support it :P)

janmeier: Aggree, and object with be prettier. Not sure we even need `on` if we get an association, since we can figure out what to join on from that

mickhansen: Ah welll, we best not break that then.

mickhansen: Yeah maybe just support `on: Association`

test/unit/sql/select.test.js
janmeier: Oh, thats probably why appveyor is failing :D

test/unit/sql/select.test.js
```diff
@@ -80,6 +80,49 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
           primaryKey: true,
           autoIncrement: true,
           field: 'id_user'
+        }
+      });
+      var Project = Support.sequelize.define('project', {
```
felixfbecker: `const`

test/unit/sql/select.test.js
```diff
@@ -80,6 +80,49 @@ suite(Support.getTestDialectTeaser('SQL'), function() {
...
+          [
+            '(SELECT [user].[id_user] AS [id] FROM [users] AS [user] INNER JOIN [project_user] AS [projects] ON [user].[id_user] = [projects].[userId] AND [projects[.[companyId] = 1 ORDER BY [user].[last_name] ASC'+sql.addLimitAndOffset({ limit: 3, order: ['last_name', 'ASC'] })+')',
+            '(SELECT [user].[id_user] AS [id] FROM [users] AS [user] INNER JOIN [project_user] AS [projects] ON [user].[id_user] = [projects].[userId] AND [projects[.[companyId] = 5 ORDER BY [user].[last_name] ASC'+sql.addLimitAndOffset({ limit: 3, order: ['last_name', 'ASC'] })+')'
+          ].join(current.dialect.supports['UNION ALL'] ?' UNION ALL ' : ' UNION ')
+        +') AS [user];'
+      });
+    }());
+
+    (function () {
+      var User = Support.sequelize.define('user', {
```
felixfbecker: `const`

lib/dialects/abstract/query-generator.js
```diff
@@ -1386,16 +1386,34 @@ const QueryGenerator = {
...
+                  attributes: [],
+                  association: options.groupedLimit.on[0],
+                  where: {
+                    [options.groupedLimit.on[1]]: value
+                  }
+                }],
+                model
+              }).include;
+            } else {
+              where = _.assign(options.where);
```
felixfbecker: Isn't this a noop?

mickhansen: I think it's a typo actually

felixfbecker:
@janmeier I do not fully understand what `groupedLimit` with your changes does - could you explain it a it further?

janmeier:
@felixfbecker Sure!

Currently `groupedLimit` allows you to do a limit for each key in `groupedLimit.values`, similar to what `LATERAL JOIN / CROSS APPLY` does. 

I.e to get three users for each projects 1, 2, and 3:
```sql
SELECT * FROM (
  UNION ALL
    SELECT * FROM "user" WHERE project_id = 1 LIMIT 3,
    SELECT * FROM "user" WHERE project_id = 2 LIMIT 3,
    SELECT * FROM "user" WHERE project_id = 3 LIMIT 3
);
```

Since not every dialect supports lateral joins, this is currently implemented as a union. Currently `groupedLimit` is used by `include.separate` for hasMany includes (where we first load the main table, getting matching proejct ids, and then fetch a limited amount of users for each)

If we want to enable `include.separate` for belongsToMany, we need to join the through table 
```sql
SELECT * FROM (
  UNION ALL
    SELECT * FROM "user" INNER JOIN "user_projects" ON .. AND project_id = 1 LIMIT 3,
    SELECT * FROM "user" INNER JOIN "user_projects" ON .. AND project_id = 2 LIMIT 3,
    SELECT * FROM "user" INNER JOIN "user_projects" ON .. AND project_id = 3 LIMIT 3
);
```
Which is what this change enables.

lib/dialects/abstract/query-generator.js
janmeier: Yep, should be `where` (to clone it)

felixfbecker: Also, can use `Object.assign()`

lib/dialects/abstract/query-generator.js
```diff
@@ -1387,22 +1389,29 @@ const QueryGenerator = {
...
                 }],
                 model
               }).include;
-            } else {
-              where = _.assign(options.where);
-              where[options.groupedLimit.on] = value;
+            }
+
+            if (whereKey) {
+              where = Object.assign(where);
```
felixfbecker: this is still a noop

felixfbecker:
The `on: association` API looks nice to me for this 👍

mickhansen:
LGTM, just waiting for verification against the dataloader library from @janmeier

janmeier:
I've also added caching of the query, so we don't generate it from scratch each time, but just replace the part that changed (the where part). This makes the code run at least twice as fast for n > 5. for n = 2 caching is slightly slower (because of splicing and str index), but only by about 10%. Since the general case is for a large n, I think this is the best option

lib/dialects/abstract/query-generator.js
```diff
@@ -1551,7 +1628,7 @@ const QueryGenerator = {
     return {mainQueryOrder, subQueryOrder};
   },
 
-  selectFromTableFragment(options, model, attributes, tables, mainTableAs, whereClause) {
```
janmeier: unused

mickhansen:
LGTM, would be great to see a test with multiple order statements though

mickhansen:
And a green build :D

janmeier:
@mickhansen Good catch - I actually discovered a slight order bug by writing another test :)

janmeier:
Hold off merging for a bit - I need to make sure this supports all the cases needed when integration with dataloader :)

janmeier:
Should be good to go now

PR #6603, opened 2016-09-16
Remove Node 6.3 version pin for MSSQL

felixfbecker:
https://nodejs.org/en/blog/release/v6.6.0/


felixfbecker:
Seems like it is not available yet
```
Installing node 6.5.0 (x64)...
```

felixfbecker:
But tests passed o.o

PR #6617, opened 2016-09-21
Implement optimistic locking - v3

davidtsai:
Preliminary pull request to add support for optimistic locking directly into _Sequelize_.  Currently only implemented for _Postgres_, but will add remaining dialects and test cases if we can get general agreement on the direction of this PR.
## Why Sequelize should support optimistic locking

Optimistic locking is a widely accepted best practice to prevent data from being clobbered when multiple requests/processes/hosts attempt to write the same model instance.  Optimistic locking makes it easier to write concurrency safe application code using Sequelize without being an expert on database transactions and read isolation semantics.  In addition to that, optimistic locking is superior because it is less likely to cause deadlocks as compared to transaction based pessimistic locking.
## Why this should not be a plugin
- The practical reason for this not being a plugin is because it can't be implemented under the current plugin architecture.  The update query needs to commit the update and increment the version number only if the version matches the version stored in the model instance.  The current hooks provided does not allow this be accomplished because you cannot modify the update query directly and it also does not return how many rows were affected in the update.
- The second high level reason is that `Sequelize`should support this as a first class citizen because it is a widely used best practice that competing ORMs on different platforms support.
## What this PR does
- It adds an option to model definitions named `version` that turns on optimistic locking for the model by adding a version number attribute.
- When optimistic locking is turned on for a model, updates to a model instance will throw an error if the application does not have a current copy of the model instance.  Example:

``` javascript
var one = await Model.create();
var two = Model.findById(one.id);
one.value = 1;
await one.save(); // Succeeds
two.value = 2;
await two.save(); // Throws `OptimisticLockError` because two is a stale copy
```

To make the example above succeed you would have to reload the model:

``` javascript
var one = await Model.create();
var two = Model.findById(one.id);
one.value = 1;
await one.save(); // Succeeds
await two.reload();
two.value = 2;
await two.save(); // Succeeds
```

P.S. Thanks for reading and for creating and maintaining Sequelize!
### Pull Request check-list


felixfbecker:
Could you give some links to where this is documented in other ORMs?

felixfbecker:
It seems to me that you changed the interface of `query`. That is a breaking change and therefor it doesn't make much sense to target v3 (you should target `master` anyway and then backport to v3, we cannot merge features into v3 that don't make it into master)

davidtsai:
Sure, here are some examples:
ActiveRecord: http://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic.html
Hibernate: https://docs.jboss.org/hibernate/orm/3.3/reference/en-US/html/transactions.html#transactions-optimistic
Mongoose: http://mongoosejs.com/docs/guide.html#versionKey

davidtsai:
Yes, was considering monkeyPatching a `__rowsUpdated` field onto the instance instead of changing the interface of `query`.  But it seemed cleaner to change the interface.

Ok, will look into targeting this into master.

igrayson:
Hi @felixfbecker, re:
> That is a breaking change

Could you link me to usages of our API that would be broken by this? I'd like to understand the 'interface surface area' that we want to maintain within a given sequelize major version.

I tried to find mention of these methods in the sequelize documentation, and I only got as far as to find mention of [`QueryInterface`](http://docs.sequelizejs.com/en/v3/api/sequelize/#getqueryinterface-queryinterface), but clicking the link to see its definition gives me a 404.

It seems this PR made the assumption that `QueryInterface` was an entirely internal abstraction, and safe to augment. To correct that, and prevent it in the future, it would help to gain a better understanding of how it is vended outside the Sequelize codebase.

felixfbecker:
The QueryInterface is used in migrations, see http://docs.sequelizejs.com/en/v3/docs/migrations/#functions. We also have sequelize.query() for raw queries. I don't think how the return values are passed to this, need some other maintainer to comment / review.

Verdier:
What's the difference with a `Repeatable Read` transaction?

igrayson:
@Verdier there are a lot of good discussions comparing the characteristics pessimistic locking, and repeatable read isolation, with optimistic locking.

One of the big differences (for us) is that deadlocks are not possible when employing (only) optimistic locking.

lib/dialects/mysql/query.js
```diff
@@ -110,6 +110,8 @@ Query.prototype.formatResults = function(data) {
     result = data[0].version;
   } else if (this.isForeignKeysQuery()) {
     result = data;
+  } else if (this.isInsertQuery() || this.isUpdateQuery()) {
+    result = [result, data.affectedRows];
```
mickhansen: an insert query would have affectedRows equal to insert values or a unique constraint, would it not?

Verdier:
@igrayson I don't see how deadlock is possible when using a repeatable read transaction without lock: the transaction will fail if the model has been mutated by another access during the transaction. But the is no notion of lock nor deadlock?

davidtsai:
@Verdier With repeatable-read transactions (with Postgresql at least, not sure about Mysql) if another transaction holds a lock on a row, it will wait for that transaction to commit or rollback before proceeding.  Therefore there at least two scenarios that cause deadlocks:

1. Bad acting application code that doesn't complete transactions in a timely manner.  One common way this happens is when you run out of connections in the pool, and Sequelize is unable to grab another connection to process a request.  That then causes the request to go into a deadlocked state that cascades out until the whole app goes down and all available connections become deadlocked.

2. Consider this (postgres with default settings in a read isolation query will deadlock forever on this):
  - Transaction 1 writes to row A (postgres locks row A)
  - Transaction 2 writes to row B (postgres locks row B)
  - Transaction 1 writes to row B (row B is locked by transaction 2, Postgres waits).
  - Transaction 2 writes to row A (row A is locked by transaction 1, Postgres waits).
  - Postgres waits forever unless `lock_timeout` is specified.

lib/dialects/mysql/query.js
davidtsai: Not sure what your question is, but yes, I think it would always return 1 here because bulkInsert is handled differently.

sushantdhiman:
We need a port for this to `master`, @davidtsai you can open separate PR targeting master once we finalize everything here

lib/dialects/mysql/query.js
sushantdhiman: For `MySQL` you might want to use `changedRows` which is the actual number of rows updated, rather than `affectedRows` which is number of rows selected by query, which make difference when working on `update` statements

lib/model.js
```diff
@@ -669,6 +677,13 @@ Model.prototype.init = function(modelManager) {
       this._timestampAttributes.deletedAt = this.options.deletedAt || Utils.underscoredIf('deletedAt', this.options.underscored);
     }
   }
+  if (this.options.version) {
+    if (typeof(version) === 'string') {
```
sushantdhiman: `this.options.version` ?

lib/model.js
```diff
@@ -669,6 +677,13 @@ Model.prototype.init = function(modelManager) {
...
+      this._versionAttribute = this.options.version;
```
sushantdhiman: `Utils.underscoredIf` this case

lib/model.js
```diff
@@ -669,6 +677,13 @@ Model.prototype.init = function(modelManager) {
...
+    } else {
+      this._versionAttribute = Utils.underscoredIf('version', this.options.underscored);
```
sushantdhiman: underscoring `version` => `version` , an effective noop

mickhansen:
Algorithm wise, a `version` field is used because it's code-wise simpler than using the current values in the instance correct?

But wouldn't a value based locking system, rather than a version based one, allow concurrent updates of different fields?

lib/dialects/mysql/query.js
mickhansen: I'm just questioning whether you need to affect the isInsertQuery clause at all.

lib/instance.js
```diff
@@ -869,7 +896,19 @@ Instance.prototype.destroy = function(options) {
...
+      ).then(function(results) {
+        var rowsUpdated = results[1];
+        if (this.Model._versionAttribute && rowsUpdated < 1) {
+          throw new sequelizeErrors.OptimisticLockError({
+            model: this.Model.name,
+            values: values,
+            where: where,
+          });
+        }
+        return _.head(results);
```
mickhansen: Consider changing the implementation to use `Model.update` which already returns affected fields.

Verdier:
@davidtsai I disagree with your explanation. Postgres transaction does not lock nor the table nor the row by default, their is no wait nor deadlock. You have to explicitly add a lock.

If you use a repeatable read transaction without lock, and if you have a concurrent access, the following happen:

```
Transaction 1 begin
.. Transaction 1 mutate the model
.. Transaction 2 begin
.. Transaction 2 mutate the model
.. Transaction 1 mutate the model
.. Transaction 2 end => `success`
.. Transaction 1 mutate the model
Transaction 1 end => `throw concurrent access error`
```

Transaction 1 will be reject while the transaction 2 will success. The Postgres implementation for repeatable read transaction is extremely similar to your current implementation (it use temporary version on raws) but it's native...

igrayson:
@Verdier [13.2.2. Repeatable Read Isolation Level](https://www.postgresql.org/docs/9.1/static/transaction-iso.html)

> UPDATE, DELETE, SELECT FOR UPDATE, and SELECT FOR SHARE commands behave the same as SELECT in terms of searching for target rows: they will only find target rows that were committed as of the transaction start time. However, such a target row might have already been updated (or deleted or locked) by another concurrent transaction by the time it is found. **In this case, the repeatable read transaction will wait for the first updating transaction to commit or roll back (if it is still in progress).**

You can easily demonstrate this for yourself `psql`. In your above timeline, this statement will block until Transaction 1 is closed: `.. Transaction 2 mutate the model`.

Be sure to `SET SESSION TRANSACTION ISOLATION LEVEL REPEATABLE READ;`

davidtsai:
There is no such thing as a reaparable read transaction without lock...
locking is an internal Postgres mechanism to achieve repeatable read.

davidtsai:
@mickhansen 
> Algorithm wise, a version field is used because it's code-wise simpler than using the current values in the instance correct?
Correct.

> But wouldn't a value based locking system, rather than a version based one, allow concurrent updates of different fields?
No, in a value based approach you would have to check that _all_ values are equal because you cannot be sure that the value updated was not dependent on a potential stale value.

lib/instance.js
davidtsai: Hmm, I think that would be a much bigger change.  Looks like `Model.update` uses a `bulkInsert` query.

lib/model.js
davidtsai: Open to suggestions for a better name :)

sushantdhiman: there is no `version` in scope, is there ?

lib/instance.js
mickhansen: I'm almost completely certain that `update` does not use an insert query.

mickhansen:
> No, in a value based approach you would have to check that all values are equal because you cannot be sure that the value updated was not dependent on a potential stale value, eg X = Y + 1. You have to check that X, and Y are valid, not just X. So it's safer to always ensure that the entire model is not stale.

Except that `instance.save()` doesn't really support computed values.
It might be safer but it would also perform worse under concurrent updates (i do see your argument though).

lib/dialects/mysql/query.js
davidtsai: You do need to affect the `isInsertQuery()` because the `Instance.save()` uses Insert/Update interchangeably depending on if it's a new instance.

davidtsai:
It doesn't matter that `save()` doesn't support computed values.  The user of the model may compute them:
```javascript
const copyA = await Model.create({x:1, y:1});
const copyB = await Model.find(1);
copyA.y = 2;
await copyA.save();
copyB.x = copyB.y + 1;
await copyB.save();   // Model is now in an inconsistent state, X should equal 3, but is instead 2.
```
An optimistic locking implementing should support strong consistency semantics that prevents the user from above.

lib/dialects/mysql/query.js
mickhansen: @davidtsai I'm aware, but that should be detectable in `save`

mickhansen:
@davidtsai Agreed, it should definitely never "fail" silently - hadn't considered your case.

lib/model.js
davidtsai: ah yes, that's a mistake.  Thought you were commenting on the name.

lib/instance.js
davidtsai: Ah yes I meant `bulkUpdate`

lib/dialects/mysql/query.js
davidtsai: @mickhansen Yup, you're right, but I thought it was cleaner to maintain the same interface for both queries as it currently does.

davidtsai:
Ok, it looks like folks are ok with the general direction of this.  I'll look into incorporating the feedback and getting this prepped properly for inclusion into `master` over the next couple days.

Download:
Just getting into Sequelize and I was already looking for the optimistic locking `version` field tutorial. Great news that you are adding this!

mickhansen:
@davidtsai Have you considered a user land plugin for this? Seems like most of this could be accomplished with a plugin (not that i necessarily oppose it in core)

davidtsai:
@mickhansen 
Yes I did consider making a plugin as I wrote in the PR write up above.  But it's not possible without significant changes to hooks.  You cannot just modify the `version` count in a beforeUpdate hook.

mickhansen:
@davidtsai Why not? Should be possible to change both the values and the options.

davidtsai:
@mickhansen 
Changing the options is not enough.  There are several things that would have to be modified to properly implement optimistic locking using a plugin.
  1. The `where` clause is currently calculated after the `beforeUpdate` hook.  So that would need to be refactored such that `beforeUpdate` could modify it.
  2. `afterUpdate` currently does not have access to the rows affected.
  3.  Increment queries also need to increment the version count, and there are currently no hooks for that.

So in addition to my belief that it should belong in Sequelize core, the above reasons make implementing a plugin more cumbersome.

@sushantdhiman Pull Request for master created here:
https://github.com/sequelize/sequelize/pull/6637

janmeier:
Whats the state of this one @davidtsai ?

PR #6618, opened 2016-09-22
Include field name for built-in validation error messages

masad-frost:
`Validation len failed` --> `Validation len on name failed`

### DIFF
--- a/lib/instance-validator.js
+++ b/lib/instance-validator.js
@@ -262,7 +262,7 @@ class InstanceValidator {
       const validatorArgs = this._extractValidatorArgs(test, validatorType, field);
       if (!validator[validatorType].apply(validator, [valueString].concat(validatorArgs))) {
       // extract the error msg
-        throw new Error(test.msg || 'Validation ' + validatorType + ' failed');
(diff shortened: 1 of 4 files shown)


felixfbecker:
MSSQL failing

sushantdhiman:
Random error @felixfbecker , If you approve this change we can merge

lib/instance-validator.js
```diff
@@ -262,7 +262,7 @@ class InstanceValidator {
       const validatorArgs = this._extractValidatorArgs(test, validatorType, field);
       if (!validator[validatorType].apply(validator, [valueString].concat(validatorArgs))) {
       // extract the error msg
-        throw new Error(test.msg || 'Validation ' + validatorType + ' failed');
+        throw new Error(test.msg || 'Validation ' + validatorType + ' on ' + field + ' failed');
```
felixfbecker: A template string would be nice here

felixfbecker: A template string would be nice here

felixfbecker:
LGTM in general though. It would be nice if you could rebase (which would also trigger a new build and hopefully fix the tests).

PR #6637, opened 2016-09-26
Implement optimistic locking

davidtsai:
Add support for optimistic locking directly into _Sequelize_. Related discussion: #6617 
## Why Sequelize should support optimistic locking

Optimistic locking is a widely accepted best practice to prevent data from being clobbered when multiple requests/processes/hosts attempt to write the same model instance.  Optimistic locking makes it easier to write concurrency safe application code using Sequelize without being an expert on database transactions and read isolation semantics.  In addition to that, optimistic locking is superior because it is less likely to cause deadlocks as compared to transaction based pessimistic locking.
## Why this should not be a plugin
- The practical reason for this not being a plugin is because it can't be implemented under the current plugin architecture.  The update query needs to commit the update and increment the version number only if the version matches the version stored in the model instance.  The current hooks provided does not allow this be accomplished because you cannot modify the update query directly and it also does not return how many rows were affected in the update.
- The second high level reason is that `Sequelize`should support this as a first class citizen because it is a widely used best practice that competing ORMs on different platforms support.
## What this PR does
- It adds an option to model definitions named `version` that turns on optimistic locking for the model by adding a version number attribute.
- When optimistic locking is turned on for a model, updates to a model instance will throw an error if the application does not have a current copy of the model instance.  Example:

``` javascript
var one = await Model.create();
var two = Model.findById(one.id);
one.value = 1;
await one.save(); // Succeeds
two.value = 2;
await two.save(); // Throws `OptimisticLockError` because two is a stale copy
```

To make the example above succeed you would have to reload the model:

``` javascript
var one = await Model.create();
var two = Model.findById(one.id);
one.value = 1;
await one.save(); // Succeeds
await two.reload();
two.value = 2;
await two.save(); // Succeeds
```

P.S. Thanks for reading and for creating and maintaining Sequelize!
### Pull Request check-list


luqezman:
I'm having a lot of problems with deadlocks and row locking :'(

lib/model.js
```diff
@@ -3573,7 +3617,19 @@ class Model {
 
         this.setDataValue(field, values[field]);
 
-        return this.sequelize.getQueryInterface().update(this, this.constructor.getTableName(options), values, where, _.defaults({ hooks: false }, options));
+        return this.sequelize.getQueryInterface().update(
+          this, this.constructor.getTableName(options), values, where, _.defaults({ hooks: false }, options)
+        ).then(results => {
+          const rowsUpdated = results[1];
+          if (this.constructor._versionAttribute && rowsUpdated < 1) {
+            throw new sequelizeErrors.OptimisticLockError({
```
SimonSchick: Wouldn't it make more sense to pass the model directly?

lib/query-interface.js
```diff
@@ -537,7 +537,9 @@ class QueryInterface {
     options = _.clone(options) || {};
     options.type = QueryTypes.INSERT;
     const sql = this.QueryGenerator.bulkInsertQuery(tableName, records, options, attributes);
-    return this.sequelize.query(sql, options);
+    return this.sequelize.query(sql, options).then(results => {
+      return _.head(results);
```
SimonSchick: Can't you use `.then(_.head)`?

test/integration/model/optimistic_locking.test.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+const Support = require(__dirname + '/../support')
+  , DataTypes = require(__dirname + '/../../../lib/data-types')
+  , chai = require('chai')
+  , expect = chai.expect;
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+  describe('optimistic locking', function () {
+    var Account;
```
SimonSchick: Why the mixing of es5 and es6, you could make the code less verbose using arrows.

lib/model.js
```diff
@@ -780,12 +788,21 @@ class Model {
         this._timestampAttributes.deletedAt = this.options.deletedAt || Utils.underscoredIf('deletedAt', this.options.underscored);
       }
     }
+    if (this.options.version) {
+      if (typeof(this.options.version) === 'string') {
```
SimonSchick: From what I can tell most of the project does not use function style `typeof`.

lib/model.js
```diff
@@ -3266,6 +3288,9 @@ class Model {
     if (updatedAtAttr && options.fields.length >= 1 && options.fields.indexOf(updatedAtAttr) === -1) {
       options.fields.push(updatedAtAttr);
     }
+    if (versionAttr && options.fields.length >= 1 && options.fields.indexOf(versionAttr) === -1) {
```
SimonSchick: Use `_.includes(arr, v)` or `arr.includes(v)`

test/integration/model/optimistic_locking.test.js
```diff
@@ -0,0 +1,55 @@
...
+      }).then(function(account) {
+        expect(account.version).to.eq(1);
+      });
+    });
+
+    it('prevents stale instances from being saved', function() {
+      return expect(Account.create({number: 1}).then(function(accountA) {
+        return Account.findById(accountA.id).then(function(accountB) {
+          accountA.number += 1;
+          return accountA.save().then(function() { return accountB; });
```
SimonSchick: Since sequelize uses bluebird you can use the `.return(value)` shorthand!

lib/model.js
felixfbecker: `arr.includes()` is not available in Node 4, and we actually try to use less lodash

lib/model.js
felixfbecker: Using `this.constructor` to get the model is fine, it is done in a lot of places

lib/query-interface.js
felixfbecker: Actually, why lodash here? Wouldn't `.then(results => results[0])` be enough?

test/integration/model/optimistic_locking.test.js
felixfbecker: We use `function()` for the tests because they share a `this` context.

test/integration/model/optimistic_locking.test.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+const Support = require(__dirname + '/../support')
+  , DataTypes = require(__dirname + '/../../../lib/data-types')
+  , chai = require('chai')
+  , expect = chai.expect;
```
felixfbecker: The tests have not been refactored to ES6 yet, but the files who are refactored use one `const` statement per declaration

lib/query-interface.js
janmeier: `return this.sequelize.query(sql, options).get(0)` http://bluebirdjs.com/docs/api/get.html

test/integration/model/optimistic_locking.test.js
felixfbecker: not needed with ES6, just do `.then(() => accountB)`

lib/query-interface.js
felixfbecker: @janmeier With ES6 and arrow functions all these bluebird constructs can be expressed much more concise without having to know the bluebird API. I think where possible we should just use an arrow function expression, it is easier to understand

test/integration/model/optimistic_locking.test.js
SimonSchick: I know that sometimes `this` is used, but in this test file it's only used _once_, so everything else can be arrows 😄

lib/model.js
davidtsai: Did not pass the model directly because it doesn't serialize as cleanly.  The reasoning is to pass only the necessary information needed to debug any potential errors.

test/integration/model/optimistic_locking.test.js
davidtsai: I was following existing convention of tests using ES5.

davidtsai:
@sushantdhiman @janmeier @felixfbecker Hey guys, are we good with the general direction of this PR and any idea on timeline for when this could get merged?

igrayson:
Hi guys. Any update on this?

@sushantdhiman @janmeier @felixfbecker @mickhansen

test/integration/model/optimistic_locking.test.js
felixfbecker: Just nobody had the time yet to convert all the tests, but use ES6 for new tests.

felixfbecker:
I don't feel I have enough experience with this pattern to decide whether this should be added or not. Leaving this to the other maintainers

ckalogeropoulos:
Please merge this feature. It is extremely useful and we need it in other dialects too.

Also consider the backport to v3

davidtsai:
@ckalogeropoulos
This PR implements all supported dialects (PSQL, MySQL, SQLite, MSSql).

lib/errors.js
```diff
@@ -88,6 +88,25 @@ class ValidationError extends BaseError {
...
+ * Thrown when attempting to update a stale model instance
+ * @extends BaseError
+ * @constructor
+ */
+class OptimisticLockError extends BaseError {
+  constructor(options) {
+    options = options || {};
+    options.message = options.message || 'Attempting to update a stale model instance: ' + options.modelName;
+    super(options);
+    this.name = 'OptimisticLockError';
```
felixfbecker: Should be `SequelizeOptimisticLockError` to be consistent with the other error names

lib/model.js
```diff
@@ -3365,24 +3390,43 @@ class Model {
       if (!options.fields.length) return this;
       if (!this.changed() && !this.isNewRecord) return this;
 
+      const versionFieldName = _.get(this.constructor.rawAttributes[versionAttr], 'field') || versionAttr;
```
felixfbecker: What is the reason for using `_.get()` here? If you just pass a shallow property name, you could just access it directly

test/integration/model/optimistic_locking.test.js
felixfbecker: This is still not addressed - please use ES6 in the added tests

docs/docs/models-definition.md
```diff
@@ -502,7 +502,12 @@ var Bar = sequelize.define('bar', { /* bla */ }, {
   freezeTableName: true,
 
   // define the table's name
-  tableName: 'my_very_custom_table_name'
+  tableName: 'my_very_custom_table_name',
+
+  // Enable optimistic locking.  When enabled, sequelize will add a version count attriubte
+  // to the model and throw an OptimisticLockingError error when stale instances are saved.
+  // Set to true or a string with the attribute name you want to use to enable.
+  version: true
```
felixfbecker: Is this all documentation? Maybe at least a header "Optimistic locking" that explains what it is, what it does, why it is helpful and how to enable it

ckalogeropoulos:
One question please.

If I read the code correctly the version field is an integer. Is it possible to add row version support (timestamp) as it works on MS SQL?

lib/model.js
davidtsai: The `get` protects against accessing a property on an undefined field.  To remove the get, it would have to be:
`const versionFieldName = this.constructor.rawAttributes[versionAttr] ?  this.constructor.rawAttributes[versionAttr].field : versionAttr;`

docs/docs/models-definition.md
davidtsai: Ok, added additional documentation.

davidtsai:
@felixfbecker Thanks for the additional comments.  Updated with feedback.

janmeier:
Thanks for working so well with us on this @davidtsai. It's a pleasure to review a PR thats so well documented. Should be out when we release the final 4.0 later this month

Wsiegenthaler:
Fantastic work! One question - should the OptimisticLockError be exposed externally via the Sequelize object or is it accessible another way?

PR #6643, opened 2016-09-28
Mix in default options for insert queries.

munsellj:
This is already done for update queries but was missing for inserts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,3 @@ site
 docs/api/tmp.md
 ssce.js
 coverage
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -14,3 +14,4 @@ site
 docs/api/tmp.md
 ssce.js
 coverage
+*.sublime-*
```
felixfbecker: Out of scope for this PR, you can add this to `.git/info/exclude`

PR #6644, opened 2016-09-28
Mix in default options for insert queries.

munsellj:
This is already done for update queries but was missing for inserts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ coverage-*
 site
 docs/api/tmp.md
 ssce.js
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -13,3 +13,4 @@ coverage-*
 site
 docs/api/tmp.md
 ssce.js
+*.sublime-*
```
felixfbecker: This is out of scope for this PR, you can put this in `.git/info/exclude`.

PR #6675, opened 2016-10-10
Fixes #6674 - updates a logic error for throughAttributes in add

### LINKED ISSUE #6674: Error when updating through table via accessor.add or accessor.addMultiple
When using add single or add multiple accessor on a hasMany association and attempting to update an existing through entry. An error is thrown because instead of sending the attributes that we want to update we are sending the through instance instead. 
https://github.com/sequelize/sequelize/blob/master/lib/associations/belongs-to-many.js#L715.
Based on the [comments](https://github.com/sequelize/sequelize/blob/master/lib/associations/belongs-to-many.js#L707) It looks like this is a known issue and was attempted to be fixed but the attribures are beeing [derived](https://github.com/sequelize/sequelize/blob/master/lib/associations/belongs-to-many.js#L705) before checking the [throughAttributes](https://github.com/sequelize/sequelize/blob/master/lib/associations/belongs-to-many.js#L708).

This fix is correctly implemented in [set accessor](https://github.com/sequelize/sequelize/blob/master/lib/associations/belongs-to-many.js#L602-L607)

Jhorlin:
### Description of change

Fixing logic error for throughAttributes when adding single or multiple associations.

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -712,13 +712,14 @@ BelongsToMany.prototype.injectSetter = function(obj) {
       }
 
       changedAssociations.forEach(function(assoc) {
-        var throughAttributes = assoc[association.through.model.name]
-          , attributes = _.defaults({}, throughAttributes, defaultAttributes)
-          , where = {};
+        var throughAttributes = assoc[association.through.model.name];
         // Quick-fix for subtle bug when using existing objects that might have the through model attached (not as an attribute object)
         if (throughAttributes instanceof association.through.model.Instance) {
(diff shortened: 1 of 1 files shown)


felixfbecker:
What exactly is the "logic error"? Do you have a test case that fails before / passes after?

Jhorlin:
@felixfbecker I did not create any before and after test but I did create an issue #6674 where you can see that we are setting our attributes before we check if the throughAttributes. I also point to where this is done correctly. It seems that a known bug was attempted to be fixed with correct logic in set but incorrect in add.

janmeier:

PR #6681, opened 2016-10-11
Update hooks.js

### LINKED ISSUE #6680: Sequelize afterInit hook called only once
## What you are doing?

``` js
'use strict'

Sequelize.addHook('afterInit', (sequelize) => {
  console.log('after init 1')
})

Sequelize.addHook('afterInit', (sequelize) => {
  console.log('after init 2')
})

const sequelize = new Sequelize('postgres://localhost/postgres')

```

``` sh
after init 1
after init 2
```

``` sh
after init 1
```

**Dialect:** any
**Sequelize version:** 4.0.0-2

---

**Edit**:

Same happens with `sequelize.addHook('afterDefine', function () {})`. It seems that all hooks are called only once.

---

I think the problem is on that line: https://github.com/sequelize/sequelize/blob/master/lib/hooks.js#L109

It does `return` in a for loop.

alekbarszczewski:
### Pull Request check-list
### Description of change

Fixed error where `return` was called inside of a for loop. 
This prevented execution of more than one synchronous hook.
Closes #6680


janmeier:
LGTM, needs a test

Issue #6666, opened 2016-10-05
Composite where clause using sequelize.where()


jharting:
## What you are doing?

I am using Sequelize API `where()` method to construct agregate function params

``` js
attributes: [
        [sql.fn('SUM', sql.where(sql.col('system.code'), 'foo')), 'sum'];
    ]
```

This works as expected and generates:

```
SUM(`system`.`code` = 'foo')
```

However, when I try to use Sequelize.and() or Sequelize.or() to create a composite condition, e.g.:

``` js
[sql.fn('SUM',
    sql.and(
        sql.where(sql.col('system.code'), 'foo'),
        sql.where(sql.col('system.type'), 'bar')
    ))
    , 'sum']
```

I expect 

```
SUM(`system`.`code` = 'foo' AND `system`.`type` = 'bar')
```
## What is actually happening?

but instead I am getting:

```
TypeError: val.replace is not a function
    at Object.SqlString.escape (/opt/insights/api/node_modules/sequelize/lib/sql-string.js:63:15)
    at Object.QueryGenerator.escape (/opt/insights/api/node_modules/sequelize/lib/dialects/abstract/query-generator.js:983:22)
    at /opt/insights/api/node_modules/sequelize/lib/dialects/abstract/query-generator.js:1826:23
    at Array.map (native)
    at Object.QueryGenerator.handleSequelizeMethod (/opt/insights/api/node_modules/sequelize/lib/dialects/abstract/query-generator.js:1822:42)
    at /opt/insights/api/node_modules/sequelize/lib/dialects/abstract/query-generator.js:1083:26
    at Array.map (native)
    at Object.QueryGenerator.selectQuery (/opt/insights/api/node_modules/sequelize/lib/dialects/abstract/query-generator.js:1072:55)
    at QueryInterface.select (/opt/insights/api/node_modules/sequelize/lib/query-interface.js:651:25)
    at null.<anonymous> (/opt/insights/api/node_modules/sequelize/lib/model.js:1388:32)
    at processImmediate [as _immediateCallback] (timers.js:374:17)
```

**Dialect:** mysql
__Database version: 10.0.26-MariaDB MariaDB Server
__Sequelize version: 3.19.0

Is there some other way to get a composite where clause (for multiple attributes) out of `Sequelize.where()`?

felixfbecker:
What is this SQL supposed to do?
> ```sql
> SUM(`system`.`code` = 'foo' AND `system`.`type` = 'bar')
> ```

[...]

PR #6685, opened 2016-10-12
support condition objects in utility functions #6666

jharting:
### Pull Request check-list

  - I did not with mssql (don't have it available)
  - linting seems broken in master (13k problems)
  - not yet, want to hear team's opinion on the change first

### Description of change

https://github.com/sequelize/sequelize/issues/6666

Makes it possible to pass condition objects (e.g. `{foo: 'bar'}`) to `Sequelize.fn()`, `Sequelize.cast()`.
This makes it easy to use conditional counts in Sequelize queries, e.g.

`Sequelize.fn('sum', {name: 'Fred'})` which results in `SUM(``name``= 'Fred')` which can be used in attribute definition to count entities for which the condition matches.

I did not get a chance to test with mssql as I do not have it available. If anyone can do that for me I would appreciate that.

### DIFF
--- a/test/integration/utils.test.js
+++ b/test/integration/utils.test.js
@@ -5,7 +5,9 @@
 var chai = require('chai')
   , expect = chai.expect
   , Utils = require(__dirname + '/../../lib/utils')
-  , Support = require(__dirname + '/support');
+  , Support = require(__dirname + '/support')
+  , DataTypes = require(__dirname + '/../../lib/data-types')
+  , Sequelize = require('../../index');
 
 describe(Support.getTestDialectTeaser('Utils'), function() {
   describe('removeCommentsFromFunctionString', function() {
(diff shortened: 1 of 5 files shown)


test/unit/utils.test.js
```diff
@@ -456,4 +457,25 @@ suite(Support.getTestDialectTeaser('Utils'), function() {
...
     });
   });
+
+  suite('Sequelize.cast', function() {
+    var sql = Support.sequelize;
+    var generator = sql.queryInterface.QueryGenerator;
+    var run = generator.handleSequelizeMethod.bind(generator);
+    var expectsql = Support.expectsql
+
+    if (Support.getTestDialect() !== 'mssql') {
```
janmeier: No need to skip mssql - It doesn't need to be 100 percent correct syntax (in my mind), as long as it tests the functionality (that you can pass nested conditions)

test/unit/utils.test.js
```diff
@@ -456,4 +457,25 @@ suite(Support.getTestDialectTeaser('Utils'), function() {
...
+      test('accepts condition object (auto casting)', function fn() {
+        expectsql(run(sql.fn('SUM', sql.cast({
+          $or: {
+            foo: 'foo',
+            bar: 'bar'
+          }
+        }, 'int'))), {
+          default: 'SUM(CAST((`foo` = \'foo\' OR `bar` = \'bar\') AS INT))',
```
janmeier: Default should use `[` and `]` for escaping

test/integration/utils.test.js
```diff
@@ -226,4 +228,80 @@ describe(Support.getTestDialectTeaser('Utils'), function() {
       expect(Utils.singularize('status')).to.equal('status');
     });
   });
+
+  describe('Sequelize.fn', function() {
```
janmeier: While an integration test might not be strictly needed, I think this is good to keep as an example of how the functionality can be used

felixfbecker:
You can test MSSQL through AppVeyor ;)

jharting:
@janmeier @felixfbecker Thanks for feedback! I addressed some of the comments. I still do not have access to mssql (not even sure if it supports the syntax I am using in the tests). Let me try if I can use appveyor CI runner here to get the tests pass easily but maybe not.

jharting:
I played with mssql and it does not seem to support this syntax. The closest alternative is `SUM(CASE WHEN foo THEN 1 ELSE 0)`. Therefore, I put mssql exclusions into the integration tests.

felixfbecker:
Don't forget to mention in the documentation then that this is not supported by MSSQL

jharting:
I modified changelog and added mention to docs. From my POV this change is now finished. Let me know if you find anything missing.

janmeier:
Thanks :+1:

jharting:
@janmeier Thank you! Btw is there a chance to get this into v3 as well? Should I open a new MR against v3? Do you plan further 3.x releases?

janmeier:
Definitely - Let me see if I can cherry-pick it really quick

janmeier:
https://github.com/sequelize/sequelize/commit/75bd5256ad53ba98d1dd535c2f2a0feaa8412d5a, just going to wait for travis to do his thing before i push a new version

janmeier:
`v3.27.0

PR #6686, opened 2016-10-12
Fix unescaped field in MSSQL query generation

francescoinfante:
### Pull Request check-list

- [x] Have you added new tests to prevent regressions? **No, but I changed a previous test**

### Description of change

Prior than this fix this query was generated:
```SQL
CREATE TABLE [viz] ([right] VARCHAR(255) CHECK (right IN(...
```
`right` was not escaped and MSSQL didn't like it.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # Future
 - [FIXED] Issues with `createFunction` and `dropFunction` (PostgresSQL)
 - [FIXED] Issue with belongsTo association and foreign keys [#6400](https://github.com/sequelize/sequelize/issues/6400)
+- [FIXED] Issue with query generation in MSSQL, an identifier was not escaped [#6686]
 
 # 4.0.0-2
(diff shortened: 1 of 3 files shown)


sushantdhiman:
📡 `Incoming from C-137` Looks good for merge

felixfbecker:
What about tests, changelog entry?

francescoinfante:
I don't think additional tests are needed. Indeed, I fixed the test that was broke too.

I will edit the changelog.

changelog.md
```diff
@@ -1,6 +1,7 @@
 # Future
 - [FIXED] Issues with `createFunction` and `dropFunction` (PostgresSQL)
 - [FIXED] Issue with belongsTo association and foreign keys [#6400](https://github.com/sequelize/sequelize/issues/6400)
+- [FIXED] Issue with query generation in MSSQL, an identifier was not escaped [#6686]
```
felixfbecker: You need to link the actual issue, this is not understood by GitHub.

francescoinfante: fixed

PR #6696, opened 2016-10-13
Execute queries parallel in findAndCount - closes #6695

### LINKED ISSUE #6695: findAndCountAll is not parallel
## What you are doing?

``` js
// this takes 30 seconds!
Model.scope('safe').findAndCountAll(queryOptions)

// this takes 15 seconds!
Promise.all([
  Model.scope('safe').count({
    ...queryOptions,
    attributes: []
  }),
  Model.scope('safe').findAll(queryOptions)
])
```

Count and find should be executed at the same time.

findAndCountAll executes the count, waits for it to finish, then executes the find. This is a huge perf loss, both queries should be run at the same time. This is a simple fix that will increase performance of this operation by 100% - what a nice unexpected win!

**Dialect:** postgres
**Database version:** 9.6
**Sequelize version:** 3.24.3

contra:
### Description of change
#6695

Could have gotten fancy with Promise.race/Promise.all but I think leaving the existing flow was the best choice for readability/object allocations.


felixfbecker:
Could you please benchmark with [sequelize-benchmark](https://www.npmjs.com/package/sequelize-benchmark) before/after to prove it helps?
Also needs a rebase
LGTM otherwise

contra:
@felixfbecker Does that benchmark bench this specific function? I'll give it a run.

felixfbecker:
@contra My bad, uses only `findAll`: https://github.com/sushantdhiman/sequelize-benchmark/blob/master/lib/benchmarks/findAll.js#L76

A reason you closed the PR?

contra:
Well, something crazy happened while I tried to rebase and my branch got wiped + PR closed. No idea why I needed a rebase, I forked and sent a PR within 10 seconds - but hey, shit happens with git.

Don't really think it's worth your time for me to redo this whole process again - it'll probably be easier for you to land this on your own.

contra:
Sorry, some crazy shit went down trying to rebase and it wiped my repo. Fixed and rebased.

contra:
Any response? Can somebody review this?

sushantdhiman:
@contra I think this code was already optimized in its own way, If count was `0` we never run `findAll` query thus saving us an extra query.

I dont think there would be good performance benefits from parallelizing the queries. Your solution and current solution both have there own trade offs in their specific use cases.

That being said I dont have any particular issue with this new approach but I think you will have to show us some cases / examples where you see better (measurable) performance upgrade

felixfbecker:
@contra thanks for the ping. 
Also something to consider: If you do a COUNT and afterwards a SELECT with the exact same JOINs and WHERE the DB will hit the query cache for the SELECT. If you run them in parallel of course that can't work. It also only works when a new connection is used, because the protocol is sequential.
So all in all a lot of factors, we would need some proof/benchmark that this actually improves perf.
Maybe you can clone sequelize-benchmark and just change the `findAll` calls to `findAndCountAll` and the run before/after?

contra:
@sushantdhiman Yeah, previously it was optimizing for DB usage but sacrificing end-user response times, this update optimizes for end-user response times but will use more DB resources in cases where the count is 0 (not very common anyways).

I provided uses cases in the original issue # this closes. Any and all usage of findAndCountAll will go 2x faster with this. No specific cases needed for the optimization to take effect.

@felixfbecker In my experience this took findAndCountAll from 30s to 15s on PG 9.6 - I can mess with the benchmark if you want, or you can test my PoC code locally to save some time. Let me know.

sushantdhiman:
@contra I think your rationale about keeping response time lower is correct, if count is large the actual query  will take more time in retrieval thus user have to wait for two sequential queries.

Please rebase and we can merge it. I would also like @mickhansen and @janmeier review on this

contra:
@sushantdhiman Is it possible to rebase + merge from your web UI? https://github.com/blog/2243-rebase-and-merge-pull-requests

sushantdhiman:
Not when there is a conflict, then we will have to do manual command line based rebase @contra

felixfbecker:
No, that is only possible when there are no conflicts 😉 Also, we squash commits

contra:
Should be good to go now

PR #6706, opened 2016-10-15
Fix duplicate columns in ORDER BY statement bug

rdiaztushman:
For the v3 branch and only for MSSQL queries. Fixes bug causing duplicate columns in ORDER BY statement when order is already specified.  Error from SQL: "A column has been specified more than once in the order by list. Columns in the order by list must be unique."


felixfbecker:
What about master?

jacinto123:
Is there any chance of getting this merged in? I am having this issue?

felixfbecker:
@jacinto123 if you wanna port this to master

jacinto123:
@felixfbecker If I am understanding this correctly all that needs to do is the tests? Or is there something else?

PR #6709, opened 2016-10-16
(fix) DATEONLY parsed as Date

### LINKED ISSUE #4858: DATEONLY field is being parsed as a javascript Date with time/zone
I am using postgres.

My dateonly field is being stored as YYYY-MM-DD but when I pull that data down to the client, it is being sent as a full timestamp with time and timezone parts.

sushantdhiman:
### Pull Request check-list
### Description of change

Close #4858

### DIFF
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -24,6 +24,7 @@ module.exports = BaseTypes => {
   BaseTypes.ENUM.types.mysql = false;
   BaseTypes.REAL.types.mysql = ['DOUBLE'];
   BaseTypes.DOUBLE.types.mysql = ['DOUBLE'];
+  BaseTypes.GEOMETRY.types.mysql = ['GEOMETRY'];
 
   function BLOB(length) {
     if (!(this instanceof BLOB)) return new BLOB(length);
@@ -62,7 +63,7 @@ module.exports = BaseTypes => {
   };
 
   function DATE(length) {
-    if (!(this instanceof DATE)) return new Date(length);
+    if (!(this instanceof DATE)) return new DATE(length);
     BaseTypes.DATE.apply(this, arguments);
   }
   inherits(DATE, BaseTypes.DATE);
@@ -95,6 +96,16 @@ module.exports = BaseTypes => {
     }
 
     return value;
+  };
+
+  function DATEONLY() {
+    if (!(this instanceof DATEONLY)) return new DATEONLY();
+    BaseTypes.DATEONLY.apply(this, arguments);
+  }
+  inherits(DATEONLY, BaseTypes.DATEONLY);
+
+  DATEONLY.parse = function parse(value) {
+    return value.string();
   };
 
   function UUID() {
@@ -157,11 +168,10 @@ module.exports = BaseTypes => {
     return 'ENUM(' + _.map(this.values, value => options.escape(value)).join(', ') + ')';
   };
 
-  BaseTypes.GEOMETRY.types.mysql = ['GEOMETRY'];
-
   const exports = {
     ENUM,
     DATE,
+    DATEONLY,
     UUID,
     GEOMETRY,
     DECIMAL,
--- a/lib/dialects/postgres/data-types.js
+++ b/lib/dialects/postgres/data-types.js
@@ -24,14 +24,24 @@ module.exports = BaseTypes => {
     array_oids: [3807]
   };
 
-  BaseTypes.DATEONLY.types.postgres = {
-    oids: [1082],
-    array_oids: [1182]
-  };
-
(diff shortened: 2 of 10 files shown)


lib/dialects/mysql/data-types.js
```diff
@@ -62,7 +63,7 @@ module.exports = BaseTypes => {
   };
 
   function DATE(length) {
-    if (!(this instanceof DATE)) return new Date(length);
+    if (!(this instanceof DATE)) return new DATE(length);
```
janmeier: Good catch ;)

lib/model.js
```diff
@@ -3085,14 +3086,23 @@ class Model {
...
-                originalValue = new Date(originalValue);
-              }
-              if (originalValue && value.getTime() === originalValue.getTime()) {
-                return this;
+              // Dont parse DATEONLY to new Date, keep them as string
+              if (this.rawAttributes[key].type instanceof DataTypes.DATEONLY) {
+                if (originalValue && originalValue === value) {
+                  return this;
+                } else {
+                  value = moment(value).format('YYYY-MM-DD');
```
mickhansen: Should perhaps happen in stringify instead?

sushantdhiman: When we build instance, we need to keep `YYYY-MM-DD` there as well, it prevent build instance to have `Date`. Its a safe guard against case when user are passing `new Date` / `moment` to `DATEONLY`. This is also done in stringify, but stringify is only called when we are putting data from instance to query

.eslintrc.json
```diff
@@ -10,7 +10,7 @@
     "no-var": "error",
     "prefer-const": "error",
     "semi": ["error", "always"],
-    "space-before-function-paren": ["warn", "never"],
+    "space-before-function-paren": "off",
```
felixfbecker: Why this? I tried to align this rule with how most of the codebase is written. I know it is inconsistent in the codebase which is why I made it a warning

lib/dialects/postgres/data-types.js
```diff
@@ -24,16 +24,26 @@ module.exports = BaseTypes => {
     array_oids: [3807]
   };
 
-  BaseTypes.DATEONLY.types.postgres = {
-    oids: [1082],
-    array_oids: [1182]
-  };
-
```
felixfbecker: Why move this around?

lib/model.js
felixfbecker: Hm... I don't know, do we have other cases where values passed to `build`/`set` are converted to another data type? Might be unexpected behavior, I would think that these two are equivalent:
```js
const user = new User({birthday: new Date()})
const user = Object.assign(new User(), {birthday: new Date()});

File: test/integration/data-types.test.js [hunk 5]
```diff
@@ -438,4 +438,22 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
...
 
+  it('should return YYYY-MM-DD format string for DATEONLY', function () {
+    const Model = this.sequelize.define('user', {
+      stamp: Sequelize.DATEONLY
+    });
+    const testDate = moment().format('YYYY-MM-DD');
+
+    return Model.sync({ force: true})
+      .then(() => Model.create({ stamp: testDate }))
+      .then((record) => {
```
felixfbecker: Unneeded parens

lib/dialects/postgres/data-types.js
sushantdhiman: To keep it with DATEONLY declaration

felixfbecker: But all other oids are grouped at the top?

sushantdhiman: No all of them

felixfbecker: Okay then. Btw imo in the future we shouldn't patch on `BaseTypes`

sushantdhiman:
Can be merged @sequelize/core ?

felixfbecker:
@sushantdhiman 
https://github.com/sequelize/sequelize/pull/6709#discussion_r83595347
https://github.com/sequelize/sequelize/pull/6709#discussion_r83595511

lib/model.js
sushantdhiman: ` Object.assign(new User(), {birthday: new Date()});` Syntax wont work actually as we hide data in `dataValues`

Beside we are already converting incoming user data to `Date` instance as well. So I dont think there should be any problem converting it to `YYYY-MM-DD`, because thats what `DATEONLY` field is storing

felixfbecker: If we do it in other places as well it's okay though

sushantdhiman:
Flag is green and I think all comments are properly answers. Please merge if looks good to go @sequelize/core

janmeier:
:+1:

PR #6714, opened 2016-10-17
docs: Add documentation for [options.fields] on instance.validate().

### LINKED ISSUE #6703: Document [options.fields] on instance.validate()
The documentation for the `validate()` method on instances only covers the `skip` option, which allows you to specify which fields you want to ignore when running the validations. 

screen shot 2016-10-14 at 18 07 47

However, looking at the code, we also have the option to pass a `fields` option, which allows you to specify which exact fields you want to validate.

https://github.com/sequelize/sequelize/blob/3e5b8772ef75169685fc96024366bca9958fee63/lib/instance-validator.js#L21

Is it safe to rely on this option, or is it meant to be private for internal usage only? If it's ok to be supported, I can send a PR to update the documentation to include this option on the list, just let me know.

guisehn:
Closes #6703

### DIFF
--- a/docs/api/instance.md
+++ b/docs/api/instance.md
@@ -256,7 +256,8 @@ Emits null if and only if validation successful; otherwise an Error instance con
 | Name | Type | Description |
 | ---- | ---- | ----------- |
 | [options] | Object | Options that are passed to the validator |
-| [options.skip] | Array | An array of strings. All properties that are in this array will not be validated |
+| [options.fields] | Array | An array of strings. Only the properties that are in this array will be validated |
+| [options.skip] | Array | An array of strings. All properties that are in this array will not be validated. If this option is specified, `[options.fields]` will be ignored |
 
 
 ***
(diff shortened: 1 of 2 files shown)


docs/api/instance.md
```diff
@@ -256,7 +256,8 @@ Emits null if and only if validation successful; otherwise an Error instance con
 | Name | Type | Description |
 | ---- | ---- | ----------- |
 | [options] | Object | Options that are passed to the validator |
-| [options.skip] | Array | An array of strings. All properties that are in this array will not be validated |
+| [options.fields] | Array | An array of strings. Only the properties that are in this array will be validated |
```
janmeier: You need to add in it code comments, otherwise it will be overwritten the next time docs are regenerated

guisehn:
@janmeier makes sense. Also, should I send another pull request based on `v3`?

janmeier:
You are very welcome to

PR #6721, opened 2016-10-19
Added missing conformOptions to scope when options is an object

molotowtales:
### Pull Request check-list

### Description of change

This will fix scope options when called with parameters to enable returning an attributes object with an include list for example:
`db.Project.scope({ method: ['custom', 19]}).findAll();`

```
...
    scopes: {
      custom: (custom) => {
        return {
          attributes: {
            include: [
              [sequelize.literal('(select ' + custom + ')'), 'custom_field']
            ]
          }
        }
      }
    },
...
```

No documentation updated, as this is done as specified therein.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1180,10 +1180,12 @@ Model.prototype.scope = function(option) {
         if (Array.isArray(option.method) && !!self.options.scopes[option.method[0]]) {
           scopeName = option.method[0];
           scope = self.options.scopes[scopeName].apply(self, option.method.splice(1));
+          conformOptions(scope, self);
         }
         else if (!!self.options.scopes[option.method]) {
           scopeName = option.method;
           scope = self.options.scopes[scopeName].apply(self);
+          conformOptions(scope, self);
         }
       } else {
         scope = option;


lib/model.js
```diff
@@ -1180,10 +1180,12 @@ Model.prototype.scope = function(option) {
         if (Array.isArray(option.method) && !!self.options.scopes[option.method[0]]) {
           scopeName = option.method[0];
           scope = self.options.scopes[scopeName].apply(self, option.method.splice(1));
+          conformOptions(scope, self);
```
janmeier: Hmm, this code could be simplified a bit to avoid too much duplication. If we don't call the scope function here, but just set scope, (and possibly arguments), and then call `isFunction` and `conformOptions` after the else clause we only need the code once :)

ptshih:
Any update on this? Just ran into it today.

janmeier:
@ptshih @molotowtales Hasn't reacted to my review yet - You are welcome to pick up where he left off :)

molotowtales:
Sorry, I've been away from the project for some time now.

PR #6724, opened 2016-10-19
Added beta ribbon to docs

robdvr:

### Pull Request check-list

### Description of change

Added beta ribbon to docs
screen shot 2016-10-19 at 11 06 56 am


felixfbecker:
There is no beta version of Sequelize out and v4 will have docs hosted on doclets, what is the reason for this PR?

robdvr:
I talked to Felix about this on slack. It's particularly easy to land on latest docs especially if you come from a google search. This PR makes it hard to miss that you are reading beta docs.

robdvr:
Here's an example
screen shot 2016-10-19 at 11 16 13 am

janmeier:
Yeah ... I disabled the builds on readthedocs for `latest`, but unfortunately there is no way to remove the current builds, or prevent them from showing up in google searches.

For the final v4 release I guess it would be best if we keep both v3 and v4 on doclets to avoid confusion - So at some point I probably need to go through and clean up the API docs for v3 as well so they compile and look nice for doclets.

I think a beta ribbon is a good solution for now, since we can't really remove `latest`

felixfbecker:
I just think Beta already implies some stability. We are Alpha at best

janmeier:
Alpha (or mayhaps unstable) would be fine by me as well - As long as we can show users in some way that they are not viewing the stable version.

sushantdhiman:
`Beta / Alpha` Looks like Sequelize doesn't have any stable release at all, can we add some text in ribbon to address this issue ?

robdvr:
@janmeier @felixfbecker Do you want me to update the ribbon to say "alpha" then?

PR #6730, opened 2016-10-21
fix a bulkCreate bug when use updateOnDuplicate

gotojmp:
There's a bug when use bulkCreate with option updateOnDuplicate, bulkInsert results length not equal instances length.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2266,7 +2266,7 @@ class Model {
         return this.QueryInterface.bulkInsert(this.getTableName(options), records, options, attributes).then(results => {
           if (Array.isArray(results)) {
             results.forEach((result, i) => {
-              instances[i].set(this.primaryKeyAttribute, result[this.rawAttributes[this.primaryKeyAttribute].field], {raw: true});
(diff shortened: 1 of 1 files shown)


janmeier:
Please show some code that fails without this fix.

gotojmp:
//Test code:
var DB = new Sequelize(...);
var Page = DB.define('page', {
    id: {type: Sequelize.INTEGER.UNSIGNED, allowNull: false, primaryKey: true, autoIncrement: true},
    index: {type: Sequelize.INTEGER.UNSIGNED, allowNull: false, defaultValue: 1},
    url: {type: Sequelize.STRING(100), allowNull: false, defaultValue: ''}
}, {
    indexes: [{
        unique: true,
        fields: ['index']
    }]
});
DB.sync();
Page.bulkCreate([
    { index:1, url:'https://github.com/1' },
    { index:1, url:'https://github.com/2' }
], {
    updateOnDuplicate: ['url']
});

//line 2267 in model.js
//results outputs  [ { id: 1 }, { id: 2 }, { id: 3 } ]
//but instances's length is 2

janmeier:
Please add that as an integration test then ;)

janmeier:

PR #6735, opened 2016-10-22
Add rejectOnEmpty param to API docs

PerGullberg:
Minor docs change


janmeier:
You need to edit the inline comments in the code - the markdown files are automatically generated. 

This probably also needs to be changed on the v3 branch

PerGullberg:
@janmeier Thanks, that makes sense. And now after looking more carefully I see there is a whitespace related issue with the rejectOnEmpty comment inside model.js. That explains why the parameter i missing in the markdown files. I will rework my pull request.

PR #6774, opened 2016-10-29
Using Takis

### LINKED ISSUE #6733: Use giorgos/takis in Travis Setup
Tests are failing most of time because Postgres was not able to start in `30 sec` sleep time (ah.. hack). We need to replace this with something realistic like https://hub.docker.com/r/giorgos/takis/

sushantdhiman:
Fixes the broken builds on CI. But I still haven't used Alpine images as I didn't find enough time for building new docker images.

Please merge this PR so CI works again. We can use Alpine Images later on.

Closes #6733


janmeier:
Sweet! :)

PR #6782, opened 2016-11-01
Fixes MSSQL bulkInsertQuery when options and attributes are not passed

harshithkashyap:
### Pull Request check-list

### Description of change

When the queryInterface bulkInsert is used with MSSQL dialect in the sequelize cli seed phase,  

> Cannot read property 'name' of undefined error is thrown.

`queryInterface.bulkInsert` internally calls `QueryGenerator.bulkInsertQuery` and is a separate function in case of MSSQL. 

#### PR Changes:
- Add default options and attributes to the bulkInsertQuery method. 
- Add a check to see if the attributes object exists before reading autoIncrement property. 
- Unit Tests MSSQL bulkInsertQuery method.


harshithkashyap:
Pinging @mickhansen @janmeier

felixfbecker:
Was only MSSQL affected by this bug?

harshithkashyap:
@felixfbecker Yes, Since MSSQL has its own [bulkInsertQuery](https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/query-generator.js#L234)

lib/dialects/mssql/query-generator.js
```diff
@@ -247,14 +249,14 @@ var QueryGenerator = {
     Utils._.forEach(attrValueHashes, function(attrValueHash) {
       // special case for empty objects with primary keys
       var fields = Object.keys(attrValueHash);
-      if (fields.length === 1 && attributes[fields[0]].autoIncrement && attrValueHash[fields[0]] === null) {
+      if (fields.length === 1 && attributes[fields[0]] && attributes[fields[0]].autoIncrement && attrValueHash[fields[0]] === null) {
```
sushantdhiman: This statement has gone a bit complex now. Please extract `attributes[fields[0]]` as a variable (`firstAttr`) so we can simplify the checks. May be we can get rid of `null` check from end as well

harshithkashyap: Done!

janmeier:
:+1:

harshithkashyap:
@janmeier Thanks for the merge. I need this patch in my project. Should i submit another PR to the v3 branch for it to get released?

felixfbecker:
Yup, you can do a backport

Issue #2508, opened 2014-10-31
[proposal] Plugins connecting and `sequelize` instance sharing


cusspvz:
Since `sequelize` is relaying on `hooks` to comunicate with plugins, shouldn't be better to include a method so we can pass a `sequelize` instance to plugins?

I would propose a similar `socket.io` and  `express.js` approach...
#### Specs for `Sequelize.prototype.plugin`

1 - `.plugin` method should communicate with plugins by passing to the plugin's function: `sequelize` instance as 1st argument, and `options` object as 2nd argument.
1.1 - If first argument is a function, it should only call it with `sequelize` instance.
1.2 - For the laziest and KISS fans, if a string is provided instead as `.plugin`'s first argument, `.plugin` method should use require internally.
1.2.1 - We should be able to require relational paths
1.2.1.1 - If an wild card exists, it should search for `.js` files and require them all recursively.
1.2.2 - We should be able to require node modules
1.3 - If first argument is an array, it should call it self recursively with each value
2 - Options handling
2.1 - Second argument of `.plugin` should be optional and be passed to plugin **always** as an object
2.2 - If plugin's **function** has the key **defaultOptions** defined as an object, `.plugin` should extend options, if provided, from it.
3 - `.plugin` should throw an error if something went wrong.
4 - `.plugin` should return `sequelize` instance always for chain proposes.
#### Specs for `Sequelize.plugin`

1 - Should proxy into `Sequelize.prototype.plugin` by passing args context within `afterInit` hook
##### Core Implementation Example

```
Sequelize.plugin = function ( plugin, options ) {
  return this.addHook( 'onInit', function ( sequelize ) {
    sequelize.plugin( plugin, options );
  });
};
```
#### Plugging Examples

```
/*
 *  Plugging into Sequelize Class ( they will be applied on all future Sequelize instances )
 */
Sequelize.plugin( 'sequelize-cache-redis' );

/*
 *  Plugging into Sequelize instance (won't be applied on other Sequelize instances )
 */
var sequelize = new Sequelize( /* ... */ );

sequelize.plugin( 'sequelize-cache-redis' );

/*
 *  Plugging multiple plugins into a Sequelize Class by chaining
 */
Sequelize
  .plugin( 'sequelize-cache-redis' )
  .plugin( 'sequelize-tree' );

/*
 *  Plugging multiple plugins into a Sequelize Class by an array with names
 */
Sequelize
[...]

PR #6786, opened 2016-11-02
Add a `noImplicitTransactions` option

igrayson:
### Pull Request check-list

### Description of change

This adds an *optional* assertion which permits applications to enforce
a declarative transaction style. Specifically, queries will no longer
open/create a connection ad-hoc, and instead throw a NoTransactionError.

This overcomes a chronic pain we have with inconsistent transaction management, and gives us an assertion that our application code will be explicit with when and where to open database transactions, and what parts of the application _should_ be aware of database transactions.


igrayson:
Test failure looks unrelated to this change.

igrayson:
Can someone take a look at this?

felixfbecker:
so after enabling this option, even a find call will throw with a NoTransaction error? That seems a bit useless to me then because every GET route we have in our app would start to fail, there is no reason to run reads and also single writes in a transaction. Also not sure about the name -it implies that without this option, sequelize would use an implicit transaction for every query, which ia misleading - sequelize just sends the query to the DB

igrayson:
This option will be more familiar to postgresql users, as postgresql _does_ implicitly place naked statements in their own transaction.

https://www.postgresql.org/docs/9.5/static/tutorial-transactions.html

> PostgreSQL actually treats every SQL statement as being executed within a transaction. If you do not issue a BEGIN command, then each individual statement has an implicit BEGIN and (if successful) COMMIT wrapped around it.

I have less experience with mysql and sqlite, but I understand both also have concepts of implicit transactions, which gives reason for users to prefer to avoid them.

mickhansen:
This could be doable as a plugin instead.

felixfbecker:
@igrayson what you are describing is called "auto commit mode" and exists in the other dialects too. There are actual options on the DB/SQL level to turn this on/off, but they will not result in an error, rather in a non-commited transaction. So I see point for confusion here.

Could you answer my other question about finds?

sushantdhiman:
I agree with @mickhansen this can be done via plugin, we don't need to add this to sequelize code base

igrayson:
Great. Can you point me towards the plugin interface that can be used to achieve this?

sushantdhiman:
@igrayson We are still working on plugin system proposal #2508 , but you can see how others have implemented various plugins [Add-ons-&-Plugins](https://github.com/sequelize/sequelize/wiki/Add-ons-&-Plugins)

Please update the wiki once you have implemented `non implicit transaction` plugin so others in community can find it as well :)

PR #6794, opened 2016-11-03
fix the date in the mysql database does not match the date taken

xiaolovecai:

### Pull Request check-list

### Description of change

```js
return value.toLocaleString();
```
I config the timezone '+08:00',and the date in the mysql database does not match the date taken.I find the reason that parse date-type don‘t use toLocaleString() function.

### DIFF
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -95,7 +95,7 @@ module.exports = BaseTypes => {
       value = new Date(value + ' ' + options.timezone);
     }
 
-    return value;
+    return value.toLocaleString();
   };
 
   function DATEONLY() {


felixfbecker:
> _Please provide a description of the change here._

felixfbecker:
The function you modified is supposed to return a `Date` object, not a string.

janmeier:
With timezone `+08:00` the current UTC time (8:35) will be inserted as `'2016-11-03 16:34:28'`, and parsed with the correct timezone when selected by sequelize again.

Since mysql does not store timezone information, we have to format the time to the corresponding time zone.

Please show some concrete code (model definition, sequelize calls and SQL) that fails, and we can try to solve your issue

PR #6804, opened 2016-11-04
Tests for paranoid with only deletedAt

sushantdhiman:
Please check #6803


mickhansen:
An approach might be to allow `updatedAt: null` to disable specific timestamp fields, i see a lot of use cases where we might care about createdAt but never updatedAt.

sushantdhiman:
@mickhansen we can already disable field generation by passing `createdAt: false`, `updatedAt: false`, This PR is now just new tests which can be merged.

PR #6805, opened 2016-11-04
(fix) restore ignore field from deletedAt attr

sushantdhiman:
### Pull Request check-list

### Description of change

`restore` was ignoring the `field` attribute from `deletedAt`, trying to query to non existing column. I have fixed it by using `field` if available


sushantdhiman:
Can be merged as well @mickhansen , its v3 counterpart already released

PR #6811, opened 2016-11-06
Add missing WHERE on Merging

appsolute:
Add missing WHERE deleted = true for Merging example.


docs/docs/scopes.md
```diff
@@ -106,6 +106,7 @@ Project.scope(['deleted', 'activeUsers']).findAll();
 ```sql
 SELECT * FROM projects
 INNER JOIN users ON projects.userId = users.id
+WHERE deleted = true
```
felixfbecker: Before this condition was on the JOIN, now it is inside the WHERE. Can you confirm that this is actually the case?

PR #6812, opened 2016-11-06
Hstore & JSON fixes

### LINKED ISSUE #6752: hstore keys renamed when `options.field` is specified on a model column with the same name
## What you are doing?

``` js
const Test = sequelize.define('test', {
    testField: {
      field: 'test_field',
      type: STRING,
    }
    kv: HSTORE,
  }, {
    tableName: 'test',
    underscored: true,
    timestamps: true,
  }).schema('public')

await Test.create({
  testField: 'value1',
  kv: { testField: 'value2' },
})

const r = await Test.findAll({ where: { kv: $eq: { testField: 'value2' } } })
```

I expected the following SQL Query:

``` sql
SELECT * FROM "public"."test" WHERE "kv" = '"testField"=>"value2"'
```

Instead, I get

``` sql
SELECT * FROM "public"."test" WHERE "kv" = '"test_field"=>"value2"'
```

**Dialect:** postgres
**Sequelize version:** 3.24.3

legomind:
### Pull Request check-list

### Description of change

fixes #6752 
also fixes a small omission that prevented a few HSTORE tests from running.

### DIFF
--- a/test/integration/dialects/postgres/dao.test.js
+++ b/test/integration/dialects/postgres/dao.test.js
@@ -332,6 +332,56 @@ if (dialect.match(/^postgres/)) {
         });
       });
 
+      it('should not rename hstore fields', function() {
+        const Equipment = this.sequelize.define('Equipment', {
+          grapplingHook: {
+            type: DataTypes.STRING,
+            field: 'grappling_hook'
+          },
+          utilityBelt: {
+            type: DataTypes.HSTORE
+          }
+        });
+
+        return Equipment.sync({ force: true }).then(() => {
+          return Equipment.findAll({
+            where: {
+              utilityBelt: {
+                grapplingHook: true
+              }
+            },
+            logging(sql) {
+              expect(sql).to.equal('Executing (default): SELECT "id", "grappling_hook" AS "grapplingHook", "utilityBelt", "createdAt", "updatedAt" FROM "Equipment" AS "Equipment" WHERE "Equipment"."utilityBelt" = \'"grapplingHook"=>"true"\';');
+            }
+          });
+        });
+      });
+
+      it('should not rename json fields', function() {
+        const Equipment = this.sequelize.define('Equipment', {
+          grapplingHook: {
+            type: DataTypes.STRING,
+            field: 'grappling_hook'
+          },
+          utilityBelt: {
+            type: DataTypes.JSON
+          }
+        });
+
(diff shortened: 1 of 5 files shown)


test/integration/dialects/postgres/dao.test.js
```diff
@@ -332,6 +332,31 @@ if (dialect.match(/^postgres/)) {
         });
       });
 
+      it('should not rename hstore fields', function() {
```
janmeier: This could be a unit test instead https://github.com/legomind/sequelize/blob/8e12f8e9148540d2ad782570ec37163529f6c01f/test/unit/sql/select.test.js

legomind: No. That was where I tried put it before I knew where the issue was. Simply put, since the problem is caused by the DOA code, It cannot be a unit test.

legomind: No can do... It is the Model that transforms the where object into the stringified where conditions. Tests here will fail with 'Error: Invalid value [object Object]'

legomind:
@janmeier Should I also add a test case & fix for json datatype as well?

janmeier:
@legomind Yes please, that would be nice! :)

test/integration/dialects/postgres/dao.test.js
janmeier: Roger roger :)

test/integration/dialects/postgres/dao.test.js
```diff
@@ -332,6 +332,31 @@ if (dialect.match(/^postgres/)) {
...
+          grapplingHook: {
+            type: DataTypes.STRING,
+            field: 'grappling_hook'
+          },
+          utilityBelt: {
+            type: DataTypes.HSTORE
+          }
+        });
+
+        return Equipment.sync({ force: true }).then( () => {
```
felixfbecker: space here

legomind:
ok @janmeier @felixfbecker. everything should be addressed now.

let me know when you are ready for me to squash the commits.

test/integration/dialects/postgres/dao.test.js
legomind: sneaky spaces :-)

legomind: someone should really set the eslint rule `space-in-parens` to 'never'

legomind:
@janmeier squashed & rebased.

PR #6826, opened 2016-11-08
Apply automatic eslint fixes (--fix) to lib/

noderat:
This PR applies all possible ESLint 3.9.1 fixes to `lib/` while retaining JSHint compliance. Unfortunately `test/` isn't so easy.

### DIFF
--- a/lib/associations/base.js
+++ b/lib/associations/base.js
@@ -1,26 +1,27 @@
 'use strict';
 
-const Association = function() {};
+class Association {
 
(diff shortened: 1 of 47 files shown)


lib/associations/base.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var Association = function() {};
+const Association = function() {};
```
noderat: I'll probably refactor it into a class before I'm done

sushantdhiman:
Too big change to be accepted in single PR.

PR #6833, opened 2016-11-10
fix: groupedLimit on paranoid junction table

sushantdhiman:
### Pull Request check-list

### Description of change

When `groupedLimit` was used on a N:M relationship whoes junction/through model was paranoid, queries generated didn't include the paranoid clause. I have fixed it by applying paranoid clause when we have this case

Also `BelongsToMany` options incorrectly uses source model `options`, this is fixed as well


janmeier:
`v3.25.0`

PR #6835, opened 2016-11-10
fix: use paranoid on groupedLimit.on

sushantdhiman:
Port for #6833 to master


changelog.md
```diff
@@ -1,5 +1,7 @@
 # Future
-- [FIXED] `restore` now uses `field` from `deletedAt` 
+- [FIXED] Set `timestamps` and `paranoid` options from through model on `belongsToMany` association
+- [FIXED] Properly apply paranoid condition when `groupedLimit.on` association is `paranoid`
+- [FIXED] `restore` now uses `field` from `deletedAt`
```
janmeier: Is this actually fixed in master, or just a copy pasta?

sushantdhiman: its fixed for master as well, test will fail without said chanegs

janmeier:
Hmm, fails for mysql :(

lib/associations/belongs-to-many.js
```diff
@@ -171,6 +171,10 @@ class BelongsToMany extends Association {
       }
     }
 
+    this.options = _.assign(this.options, _.pick(this.through.model.options, [
```
felixfbecker: `Object.assign()`

test/integration/model/findAll/groupedLimit.test.js
```diff
@@ -135,6 +161,55 @@ if (current.dialect.supports['UNION ALL']) {
...
             });
           });
+
+          it('works with paranoid junction models', function() {
+            return this.User.findAll({
+              attributes: ['id'],
+              groupedLimit: {
+                limit: 3,
+                on: this.User.ParanoidProjects,
+                values: this.projects.map((item) => item.get('id'))
```
felixfbecker: Unneeded parens (a few more times further down too)

sushantdhiman:
@felixfbecker Do we really need these changes, there is no practical difference :), ping @janmeier

felixfbecker:
It's included in the ESLint. Running `eslint --fix` would remove them all - there is already enough stuff to be fixed, I would like to at least keep new code additions clean. As for the justification of the rule itself, you could add hundreds of parenthesis around every expression in the code, but it makes the code less readable.

sushantdhiman:
Hehe :smile_cat:

PR #6839, opened 2016-11-11
fix instance.set '1970-01-01' to null field

neversun:
### Pull Request check-list

### Description of change

Before this change it is not possible to use `instance.set('date', '1970-01-01')` when:
-  model has allowNull: true
- instance retrieved from database and thus date having null as value.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -346,11 +346,13 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
             if (!(value instanceof Date)) {
               value = new Date(value);
             }
-            if (!(originalValue instanceof Date)) {
-              originalValue = new Date(originalValue);
-            }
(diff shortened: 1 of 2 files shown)


felixfbecker:
We need this for master first, otherwise LGTM.

gutobortolozzo:
@neversun I would like to put two cents in your PR. I changed your code to 

```
if (originalValue && !(originalValue instanceof Date)) {
  originalValue = new Date(originalValue);
}
if (originalValue && value.getTime() === originalValue.getTime()) {
 return this;
}
```
and tests looks to work just fine. What do you think about change your code to this shorter version?

Thank you

felixfbecker:
It is not shorter in terms of code branches. You are testing the same condition twice, while the PR only tests it once.

Issue #6824, opened 2016-11-08
ROW_FORMAT support for MySQL InnoDB


noderat:
There isn't currently support for ROW_FORMAT in the Query Generator, which is incredibly handy for InnoDB on newer versions of MySQL.

InnoDB's new [Barracuda](https://dev.mysql.com/doc/refman/5.7/en/glossary.html#glos_barracuda) file format offers [2 new types of row formats](https://dev.mysql.com/doc/refman/5.5/en/innodb-row-format-dynamic.html) which offer significant improvements if working with utf8mb4 collation by supporting index key prefixes up to 3072 bytes instead of the default of 767 bytes offered with [COMPACT or REDUNDANT](https://dev.mysql.com/doc/refman/5.7/en/innodb-row-format-antelope.html) row types offered in the legacy Antelope file format. To enable the new row formats one must pass ROW_FORMAT when creating a table.

I was looking through v3 and v4, both seem pretty easy to add and it's something that's been driving me crazy because I keep running into the 767 byte limit. Is this something on the roadmap or should I work on a PR?

felixfbecker:
Not on the roadmap. Feel free to open a PR

noderat:
Okay, I've got a PR for v3 almost ready, just verifying tests pass.

felixfbecker:
We need it for master first though, you can backport it afterwards

noderat:
Okay, I'll submit a PR for master, await approval to submit PR for v3 backport.

noderat:
Submitted #6825 though I'm not sure why GitHub didn't link the issues

PR #6846, opened 2016-11-13
Fix error 'Index column size too large.'

monmon:
### Pull Request check-list

### Description of change

Setting `charset = utf8mb4` and attempting to index to a column larger than varchar (255) results in the following error.
```
SequelizeDatabaseError: Index column size too large. The maximum column
size is 767 bytes.
```
If it can be solved by setting ROW_FORMAT to DYNAMIC or COMPRESSED, it changed so that ROW_FORMAT option can be set.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,5 @@
 # Future
+- [ADDED] support for ROW_FORMAT
 - [FIXED] `restore` now uses `field` from `deletedAt` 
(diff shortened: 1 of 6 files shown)


felixfbecker:
This seems like the same as #6824

monmon:
Sorry. I did not have enough search.
Should I close this p-r? Or will you close it?

PR #6850, opened 2016-11-14
Use soft versioning

### LINKED ISSUE #6771: Don't pin version dependencies
Currently inflection, generic-pool, lodash and wkx are pinned to a fixed patch release. node-uuid is pinned to a minor.
This means that users will not get bug fixes for these libraries if they are released (unless we bump the version too and make another release, which we don't so often). It causes stuff like #6770.
It also means users will always have two versions of the libraries installed if they depend on it in another way.
There is no reason for this: In a semver-following package, nothing can break during a minor or patch release. Of course, mistakes can happen, but if the user wants to protect himself against that there is npm-shrinkwrap. It should be the user's decision whether to lock dependencies or use soft versioning.

felixfbecker:
Closes #6771 and will save us from constant greenkeeper updates and users from unneeded duplicate package installations


felixfbecker:
I just changed the node-uuid version from allowing patch versions to allowing minor versions and now every integration test related to UUIDs is failing. @broofa?

felixfbecker:
sql-string `escape()` is called with `DataTypes.UUID` and then chokes because it doesn't have an `escape` method. Not sure why a change in node-uuid version changed something here.

felixfbecker:
Okay, it is the lodash update from #6849

felixfbecker:
https://github.com/lodash/lodash/issues/2810

PR #6861, opened 2016-11-15
Bug in docs: mysql package needed, NOT mysql2?

TheNotary:
The example connection code would throw an error until I installed mysql as apposed to mysql2.

### DIFF
--- a/docs/docs/getting-started.md
+++ b/docs/docs/getting-started.md
@@ -7,7 +7,7 @@ $ npm install --save sequelize
 
 # And one of the following:
 $ npm install --save pg pg-hstore
-$ npm install --save mysql2
+$ npm install --save mysql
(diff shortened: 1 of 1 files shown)


felixfbecker:
Sequelize 4 uses mysql2, Sequelize 3 mysql. The docs on master are correct.

TheNotary:
ah, my mistake

TheNotary:
You know, now that I think about it, the `npm install sequelize` isn't a correct form either, because that defaults to sequelize v3 at present.  Maybe that bit of it should be updated instead?

sushantdhiman:
@TheNotary `npm` is correct as it installs stable version i.e. `v3`, `v4` or `master` is currently under development and not that stable. Once we think its stable enough `npm` will default to `v4` as well

PR #6862, opened 2016-11-15
Add YEAR data type

anonmily:

### Description of change

Added the YEAR data type for MySQL. However, I didn't add the new data type for other dialects. 

All tests, including linting, pass except for the tests for the GEOMETRY type. I haven't edited anything for that data type, though, so I'm not too sure why those tests are failing.

This is my first time contributing to sequelize, so please let me know if there's anything I should fix/add. Thanks! :+1:


lib/data-types.js
```diff
@@ -225,6 +225,59 @@ Object.defineProperty(NUMBER.prototype, 'ZEROFILL', {
   }
 });
 
+function NUMBER(options) {
```
sushantdhiman: Why this addition of `NUMBER`

lib/data-types.js
```diff
@@ -481,6 +534,33 @@ DATEONLY.prototype._stringify = function _stringify(date) {
 };
 
 /**
+ * A YEAR column.
+ *
+ *
```
sushantdhiman: Extra line

lib/data-types.js
```diff
@@ -481,6 +534,33 @@ DATEONLY.prototype._stringify = function _stringify(date) {
...
+}
+util.inherits(YEAR, ABSTRACT);
+
+YEAR.prototype.key = YEAR.key = 'YEAR';
+YEAR.prototype.toSql = function() {
+  return 'YEAR';
+};
+
+YEAR.prototype._stringify = function _stringify(date) {
+  return date;
```
sushantdhiman: I think we need to return year in string format like `2016` not date(if its not an string already), please add some test for this like https://github.com/anonmily/sequelize/blob/4d32f18526b49e5b586fd9ddc037588d94803918/test/integration/data-types.test.js#L441 to verify that we indeed return year in string format

lib/data-types.js
```diff
@@ -481,6 +534,33 @@ DATEONLY.prototype._stringify = function _stringify(date) {
...
+};
+
+YEAR.prototype.validate = function(value){
+  if (!Validator.isInt(String(value) || String(value).length !== 4 || String(value).length !== 2)){
```
sushantdhiman: Why `String(value).length !== 2`

lib/data-types.js
anonmily: OK, will do! 👍

lib/data-types.js
anonmily: The check for a length of 2 is because MySQL supports the two digit year format as well (e.g. 99 --> 1999, 17 --> 2017). Technically, MySQL supports one digit too, so maybe adding a check for a length of 1 would be good too?

http://dev.mysql.com/doc/refman/5.7/en/year.html

lib/data-types.js
anonmily: You're right! I think I probably left that in by accident, so I'll go ahead and take that out.

lib/data-types.js
felixfbecker: You are calling `Validator.isInt()` on the boolean expression `String(value) || String(value).length !== 4 || String(value).length !== 2`

sushantdhiman:
Oh needs rebase :(

lib/data-types.js
```diff
@@ -482,6 +482,32 @@ DATEONLY.prototype._stringify = function _stringify(date) {
...
+YEAR.prototype.key = YEAR.key = 'YEAR';
+YEAR.prototype.toSql = function() {
+  return 'YEAR';
+};
+
+YEAR.prototype._stringify = function _stringify(year) {
+  return String(year);
+};
+
+YEAR.prototype.validate = function(value){
```
felixfbecker: space missing after parens

sushantdhiman:
It appears MSSQL doesn't support `YEAR`, so will have to disable or store as `integer`

Issue #6870, opened 2016-11-17
[mssql] Sequelize does not set DB outputted value when inserting field using Sequelize.literal


renatoargh:
## What you are doing?
I am performing an insert but one of the fields are set to a `Sequelize.literal`

```js
Model.create({
  name: 'Foo',
  description: 'bar',
  order: Sequelize.literal('(SELECT MAX(order) + 1 FROM doesntMatter)') // valid subquery
})
```

## What do you expect to happen?
That I will get a full populated object, eg.:
```js
{ 
  id: 100, 
  name: 'foo', 
  description: 'bar', 
  order: 19283 // MSSQL will return the correct value for this field
}
```

## What is actually happening?

The resulting object after the insert is performed comes like this:

```js
{ 
  id: 100, 
  name: 'foo', 
  description: 'bar', 
  order: '(SELECT MAX(order) + 1 FROM doesntMatter)' // here is the problem
}
```

## My naive suggestion to fix the Problem
My **very naive** suggestion to fix this issue would be doing something like this in the `/lib/dialects/mssql/query.js` file

```js
Query.prototype.handleInsertQuery = function(results, metaData) {
  if (this.instance) {
    // add the inserted row id to the instance
    var autoIncrementField = this.model.autoIncrementField
      , autoIncrementFieldAlias = null
      , id = null;

    if (this.model.rawAttributes.hasOwnProperty(autoIncrementField) &&
        this.model.rawAttributes[autoIncrementField].field !== undefined)
      autoIncrementFieldAlias = this.model.rawAttributes[autoIncrementField].field ;

    id = id || (results && results[0][this.getInsertIdField()]);
    id = id || (metaData && metaData[this.getInsertIdField()]);
    id = id || (results && results[0][autoIncrementField]);
    id = id || (autoIncrementFieldAlias && results && results[0][autoIncrementFieldAlias]);

    // The following lines solves my problem
    const dataValues = this.instance.dataValues
    for (let key in dataValues) {
      if (dataValues.hasOwnProperty(key) && dataValues[key]) {
        if (dataValues[key].constructor.name === 'literal') {
          const field = this.model.rawAttributes[key].field
          this.instance[key] = results[0][field]
        }
      }
    }

    this.instance[autoIncrementField] = id;
  }
};
```

__Dialect:__ mssql
__Database version:__ MSSQL 12.0.2000.8
__Sequelize version:__ 3.25.0

PR #6871, opened 2016-11-17
Adding support for populating db outputted values inserted with Sequelize.literal

renatoargh:
I am opening this PR just to run the tests which I tried running locally but weren't able

### Description of change

Detailed in issue #6870 

When you run:

```js
Model.create({
  name: 'Foo',
  description: 'bar',
  order: Sequelize.literal('(SELECT MAX(order) + 1 FROM doesntMatter)') // valid subquery
})
```
 you get

```js
{ 
  id: 100, 
  name: 'foo', 
  description: 'bar', 
  order: '(SELECT MAX(order) + 1 FROM doesntMatter)' // here is the problem
}
```
 instead of
```js
{ 
  id: 100, 
  name: 'foo', 
  description: 'bar', 
  order: 19283 // MSSQL will return the correct value for this field
}
```

PS.: I believe this is a naive approach and should be peer reviewed.

### DIFF
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -315,6 +315,16 @@ class Query extends AbstractQuery {
       id = id || (results && results[0][autoIncrementField]);
       id = id || (autoIncrementFieldAlias && results && results[0][autoIncrementFieldAlias]);
 
+      const dataValues = this.instance.dataValues
+      for (let key in dataValues) {
+        if (dataValues.hasOwnProperty(key) && dataValues[key]) {
+          if (dataValues[key].constructor.name === 'literal') {
+            const field = this.model.rawAttributes[key].field
+            this.instance[key] = results[0][field]
+          }
+        }
+      }
+
       this.instance[autoIncrementField] = id;
     }
   }
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -316,14 +316,10 @@ class Query extends AbstractQuery {
       id = id || (autoIncrementFieldAlias && results && results[0][autoIncrementFieldAlias]);
 
       const dataValues = this.instance.dataValues;
-      for (let key in dataValues) {
-        if (dataValues.hasOwnProperty(key) && dataValues[key]) {
-          if (dataValues[key].constructor.name === 'literal') {
-            const field = this.model.rawAttributes[key].field;
-            this.instance[key] = results[0][field];
-          }
-        }
-      }
+      dataValues.keys().forEach(key => {
+        const field = this.model.rawAttributes[key].field
+        this.instance[key] = results[0][field]
+      })
 
       this.instance[autoIncrementField] = id;
     }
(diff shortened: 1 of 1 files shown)


felixfbecker:
I agree that Sequelize should behave like this, but I don't understand what your changes do.

lib/dialects/mssql/query.js
```diff
@@ -315,6 +315,16 @@ class Query extends AbstractQuery {
       id = id || (results && results[0][autoIncrementField]);
       id = id || (autoIncrementFieldAlias && results && results[0][autoIncrementFieldAlias]);
 
+      const dataValues = this.instance.dataValues;
+      for (let key in dataValues) {
+        if (dataValues.hasOwnProperty(key) && dataValues[key]) {
+          if (dataValues[key].constructor.name === 'literal') {
```
janmeier: I don't think this needs to be specific to `literal`, this could also apply to `fn` etc (anything that's generated serverside) 

I think we can just always map the result back into the instance.

lib/dialects/mssql/query.js
```diff
@@ -315,6 +315,16 @@ class Query extends AbstractQuery {
       id = id || (results && results[0][autoIncrementField]);
       id = id || (autoIncrementFieldAlias && results && results[0][autoIncrementFieldAlias]);
 
+      const dataValues = this.instance.dataValues;
+      for (let key in dataValues) {
+        if (dataValues.hasOwnProperty(key) && dataValues[key]) {
```
felixfbecker: You don't need to do this if you use `Object.keys()`

lib/dialects/mssql/query.js
felixfbecker: This should use `instanceof`. Otherwise if `order` comes from user input you could get SQL injection here by passing an object with `{constructor: {name: 'literal'}}`

felixfbecker: That nested `if` is not needed

renatoargh: @felixfbecker I will follow @janmeier suggestion and map the whole result back to the instance - will commit soon with the requested changes

lib/dialects/mssql/query.js
```diff
@@ -316,14 +316,10 @@ class Query extends AbstractQuery {
...
       const dataValues = this.instance.dataValues;
-      for (let key in dataValues) {
-        if (dataValues.hasOwnProperty(key) && dataValues[key]) {
-          if (dataValues[key].constructor.name === 'literal') {
-            const field = this.model.rawAttributes[key].field;
-            this.instance[key] = results[0][field];
-          }
-        }
-      }
+      dataValues.keys().forEach(key => {
```
felixfbecker: It's `for (const key of Object.keys(dataValues))`

renatoargh: Sorry, I think it is now ok; d4aea56168ca75889481e1470bdf6cd8831c971d

renatoargh:
hey guys, any chance this will be merged soon?

renatoargh:
@janmeier thanks for the feedback, will try adding some tests & checking why others are failing during the weekend!

janmeier:
Ping @renatoargh

renatoargh:
this is still an issue

On Jun 29, 2017 8:23 AM, "probot-stale[bot]" <notifications@github.com>
wrote:

Closed #6871 <https://github.com/sequelize/sequelize/pull/6871>.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/sequelize/sequelize/pull/6871#event-1144015933>, or mute
the thread

PR #6220, opened 2016-07-04
Fixed #5747, Properly parse Int4 range types

sushantdhiman:
### Pull Request check-list

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Have you added an entry under `Future` in the changelog?

### Description of change

Fixes #5747 , Incorrect `OID` was used for `Int4` in `range` type.

PR #6897, opened 2016-11-19
Fixed range integer parsing

### LINKED ISSUE #6896: Sequelize not returning range data properly
## What am i doing?
Trying to access int4range data in postgresql, model find returning no actual data, just empty arrays

Here's my model
```js
var Activity = sequelize.define('activity', {
    id: {
      type: Sequelize.UUID,
      defaultValue: Sequelize.UUIDV4,
      primaryKey: true
    },
    type: {
      type: Sequelize.ENUM('user', 'default'),
      allowNull: false
    },
    approved: {
      type: Sequelize.BOOLEAN,
      defaultValue: false,
      allowNull: false
    },
    name: {
      type: Sequelize.STRING,
      allowNull: false,
      trim: true
    },
    description: {
      type: Sequelize.STRING,
      allowNull: false,
      trim: true
    },
    content: {
      type: Sequelize.TEXT,
      trim: true
    },
    ageRange: {
      type: Sequelize.RANGE(Sequelize.INTEGER)
    },
    userRange: {
      type: Sequelize.RANGE(Sequelize.INTEGER)
    },
    timeRange: {
      type: Sequelize.RANGE(Sequelize.INTEGER)
    }
  }
);

Activity.belongsTo(User, {as: 'creator'});
```

Here's the code that's trying to access it:
```js
/**
 * List of Articles
 */
exports.list = function(req, res) {
  Activity.findAll()
  .then(function(activitys){
    console.log(activitys[0].dataValues.ageRange);
    return res.send(activitys);
  })
  .catch(function(err){
    return res.status(400).send(err);
  });
};
```

I have a single row in the db with the ageRange having this value: [8, 77)

## What do i expect to happen?
I expect the find query to return the in the response of the query.

The output for the request is this:
```json
[
  {
    "id": "2108125d-3bfb-4c8f-9f37-5fa0df793179",
    "type": "user",
    "approved": false,
    "name": "Party",
    "description": "Throw a party",
    "content": null,
    "ageRange": [
      [],
      []
    ],
    "userRange": null,
    "timeRange": null,
    "createdAt": "2016-11-19T13:20:50.451Z",
    "updatedAt": "2016-11-19T13:20:50.472Z",
    "creatorId": "a0abee24-698e-4218-a5f2-d63d6ab69e28"
  }
]
```

and the console.log is outputting this:
```
[ [], [], inclusive: [ true, false ] ]
```

__Dialect:__ postgres
__Database version:__ PostgreSQL 9.5.5 running on ubuntu
__Sequelize version:__ sequelize@3.26.0

FREEZX:
Fixes #6896

### DIFF
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -362,6 +362,32 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
         expect(user.get('decimalWithFloatParser')).to.be.eql('0.12345678');
       });
     });
+
+    it('should return Int4 range properly #5747 and #6896', function() {
+      var Model = this.sequelize.define('M', {
+        interval: {
+            type: Sequelize.RANGE(Sequelize.INTEGER),
+            allowNull: false,
+            unique: true
+        }
+      });
+
+      return Model.sync({ force: true })
+        .then(function() {
+          return Model.create({ interval: [
+            {value: 1, inclusive: true},
+            {value: 4, inclusive: false}
+          ] });
+        })
+        .then(function(){
+          return Model.findAll();
+        })
+        .spread(function(m){
+          console.log(m.interval);
+          expect(m.interval[0]).to.be.eql(1);
+          expect(m.interval[1]).to.be.eql(4);
+        });
+    });
   }
 
 });
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -383,7 +383,8 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
           return Model.findAll();
         })
         .spread(function(m){
-          console.log(m.interval);
+          expect(m.interval.inclusive[0]).to.be.eql(true);
+          expect(m.interval.inclusive[1]).to.be.eql(false);
           expect(m.interval[0]).to.be.eql(1);
           expect(m.interval[1]).to.be.eql(4);
         });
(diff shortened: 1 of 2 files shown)


onzag:
@mickhansen I see that it's a very small diff, seems that it will cause no problem is right, but I'm not sure what happens there nor why I was chosen as reviewer; nor how the oid changed and it's only noticed until now.

felixfbecker:
Needs a PR to master first

FREEZX:
@felixfbecker Seems like its fixed in master #https://github.com/sequelize/sequelize/pull/6220, and i don't intend to switch to v4 right now  .

felixfbecker:
Then please do a direct backport of that fix (including tests)

FREEZX:
Added the test from the fix, things seemed a bit messed up on my version of postgresql, so i had to explicitly set the inclusive values when creating the model instance. On the original in master, postgresql inserted [2,4) instead of (1, 4).

test/integration/data-types.test.js
```diff
@@ -362,6 +362,32 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
...
+          return Model.create({ interval: [
+            {value: 1, inclusive: true},
+            {value: 4, inclusive: false}
+          ] });
+        })
+        .then(function(){
+          return Model.findAll();
+        })
+        .spread(function(m){
+          console.log(m.interval);
```
janmeier: Remove

test/integration/data-types.test.js
```diff
@@ -362,6 +362,32 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
...
+          expect(m.interval[0]).to.be.eql(1);
+          expect(m.interval[1]).to.be.eql(4);
```
janmeier: Please add an assertion for inclusive

janmeier:
Thanks - Seems it was not properly fixed on master, I don't remember why there are two oids for each range though ...

PR #6903, opened 2016-11-21
Oracle support

ggrimbert:
Hi,

This is the first version of Oracle support done by my company. 
The minor version is Oracle 12c. 

This version contains the "base" Oracle support : 
 - Connecting to Oracle.
 - INSERT / UPDATE / SELECT / DELETE queries.
 - Managing transactions.
 - Call to stored procedure.

Unit tests work in my environment.
I tested this with our app, and found no problem.

My company needs this Oracle version for our new server, so the work will continue soon. 

TODO : 
- [ ] Connection Pooling.
- [ ] maxRows and resultSets.
- [ ] Some requests have still to be done (complete describe, list FKs ...).

Important point : 
Many things have to be installed due to Oracle (see https://github.com/oracle/node-oracledb/blob/master/INSTALL.md)


janmeier:
Any reason you are targeting v3 with this? We are considering releasing v4 soon-ish.

Seems it should be possible to create a docker container with oracle https://github.com/oracle/docker-images/tree/master/OracleDatabase

ggrimbert:
We targeted the V3 because it's stable and we're planning to release our new server on production to our clients in January, so we can't do this with a non stable branch.

For Docker, I haven't work it, can you explain a little bit what you're expecting and (maybe), how I can reach it ?

felixfbecker:
@ggrimbert We need a PR to master first. If we merge it in v3, the older version will have more features than the newer, and that must be avoided. So please first target master, and then you can backport to v3. The code changes have been minimal between v3 and v4, you should be able to cherry-pick easily.

janmeier:
@ggrimbert We just discussed this amongst the maintainers, and I can assure you that v4 wil be officially released before january :)

ggrimbert:
Ok, I will do this on V4.

By the way, I have checked @janmeier link, I think that it's just to embed an Oracle DB, so what's the purpose of using it ?

janmeier:
We are using docker to provide database engines, instead of installing them in CI - Basically we run all databases in containers, instead of directly on the host. This allows us to test against newer versions of postgres and mysql than whats installed on travis by default.

Seems there are already some prebuilt containers with oracle installed. Try adding https://hub.docker.com/r/wnameless/oracle-xe-11g/ to the docker compose file https://github.com/sequelize/sequelize/blob/master/docker-compose.yml, and starting it locally.
Basically we want everyone to be able to run tests against oracle without having to install the server themselves (having to install client libraries is fine though)

ggrimbert:
Ok, I'll try to implement this in my next version. 

Can we let this PR open to ask questions if I have some ? So I don't have to create a new issue at each time.

janmeier:
Sure, no problem - You are also very welcome to join us on [slack](https://sequelize-slack.herokuapp.com/)

ggrimbert:
Hi,

I am working on V4 and on integration tests (with a docker image).

I am in trouble, for your tests, you create a table called "users", and you alias it as "user". But user is reserve word for Oracle...

I can't change this in all tests, so I thought I may place somewhere a trick for renaming "user" into "users", but I'm not veery satisfied by this.

Do you have any idea ?

For information, this is the list of all Oracle reserved words : https://docs.oracle.com/cd/B19306_01/em.102/b40103/app_oracle_reserved_words.htm

janmeier:
Is there no way to quote reserved words in oracle? `user` is a reserved word in other dialects as well, but you can still use it if you quote it properly :)

ggrimbert:
You're right, this is the right way to do this.

By the way, I now have a new problem : Oracle needs a name for its constraints. So, I've decided to generate it by using the following pattern : PK_tableNameColumnName (same thing for all constraints).

But Oracle doesn't support names with more than 32 characters; In this case, I've decided to use at the place a timestamp. But this means that the same database on two differents servers will have differents names for those constraints. 

Is it acceptable or should I find an algorithm to generate the same name  ?

janmeier:
Hmm, we have a similar problem with postgres (see https://github.com/sequelize/sequelize/issues/2084)

Some kind of minification that still preserves some similarity with the original name would probably be best - Perhaps the first X characters of table and column name?

ggrimbert:
I might have a solution, by using CRC for the columns name.
The name would be PK_TableName + the CRC of the columns.

Doing some minification seems hard. In my company, many tables have a prefix (p_, r_ ...), so by reducing the name, I might find some duplicates.

I believe this an acceptable solution.

For example, for the table UserTasks as used in the tests, the name will be : PK_UserTasks459f822b

And, if the full name PK_TableNameColumnName isn't too long, I can keep it.

What do you think ?

janmeier:
:+1:

ggrimbert:
Now I have a problem with aliases too long...

We've thought about using the same system for aliases than the one I used for PK/FK names. 
These new aliases will be generated at each time a request is generated.

As the same problem can appear with Postgres, we think this should be done inside the abstract code (Query-generator).

What do you think ?

lib/dialects/abstract/query-generator.js
```diff
@@ -181,6 +181,9 @@ var QueryGenerator = {
     Returns an insert into command. Parameters: table name + hash of attribute-value-pairs.
   */
   insertQuery: function(table, valueHash, modelAttributes, options) {
+    console.log('///////////////////////////////////////////////////////////////////');
```
sushantdhiman: Why ? Debugging ?

lib/dialects/abstract/query-generator.js
```diff
@@ -414,7 +417,8 @@ var QueryGenerator = {
       query += ' LIMIT ' + this.escape(options.limit) + ' ';
     }
 
-    if (this._dialect.supports.returnValues) {
+    // Never return updated values in Oracle
+    if (this._dialect.supports.returnValues && this._dialect.name !== 'oracle') {
```
sushantdhiman: You should use dialect supports mechanism like https://github.com/sequelize/sequelize/blob/54556cc4f80513b34fb5ab6a5fd0866387f6c706/lib/dialects/postgres/index.js#L31

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,196 @@
...
+  this.refreshTypeParser(dataTypes);
+};
+
+/**
+ * Method for checking the config object passed and generate the full database if not fully passed
+ */
+ConnectionManager.prototype.checkConfigObject = function(config) {
+    //A connectString should be defined
+    if(config.database.length == 0) {
+      var errorToThrow = "The database cannot be blank, you must specify the database name (which correspond to the service name"
```
sushantdhiman: Use template strings here

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,196 @@
...
+      errorToThrow +=" or specify the full connectString with the format : HOST:[PORT]/SERVICE_NAME";
+      errorToThrow += "\n from tnsnames.ora : (HOST = mymachine.example.com)(PORT = 1521)(SERVICE_NAME = orcl)" 
+      throw new Error(errorToThrow);
+    }
+
+      //The connectString has a special format, we check it
+      //ConnectString format is : host:[port]/service_name
+      if(config.database.indexOf('/') == - 1) {
```
sushantdhiman: You should follow our linting rules, spaces before and after brackets in `if`

Also this is a URL parsing scheme, it should be placed here https://github.com/sequelize/sequelize/blob/54556cc4f80513b34fb5ab6a5fd0866387f6c706/lib/sequelize.js#L122

christianbundy:
Thanks for your work on this @ggrimbert, looking forward to seeing it merged!

ggrimbert:
Pull request on master https://github.com/sequelize/sequelize/pull/7123

jssuttles:
This was so close... :( I wish I was knowledgeable enough about this to contribute.

PR #6887, opened 2016-11-18
V3 mssql bulkinsert

TimMurnaghan:
Backport of #6904

### Pull Request check-list

_Please make sure to review and check all of these items:_

- [Y ] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [ Y] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ Y] Have you added new tests to prevent regressions?
- [N/A ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [Y ] Have you added an entry under `Future` in the changelog?

### Description of change
MSSQL bulkInsertQuery emits multiple semi-colon separated queries to avoid blowing up the MSSQL parser with one big query.
Closes #6866 

For testing, I believe that the new integration test is what's needed as it's run-time blowing up that's interesting.

PR #6904, opened 2016-11-21
Support Large BulkInsert for MSSQL

TimMurnaghan:
### Pull Request check-list

- [Y ] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [ Y] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ Y] Have you added new tests to prevent regressions?
- [ N/A] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ Y] Have you added an entry under `Future` in the changelog?

### Description of change

This is the PR to master corresponding to #6887 for the v3 branch (and with the review comments addressed).


TimMurnaghan:
@felixfbecker OK. Good catch on the use of const - I've put those in in the main code. The only place I could use arrow functions in these changes would be in the test file - but I was fitting in with the usage in the file which seems unconverted.
I don't want to touch a load of tests which aren't mine - but do you want the new tests added with ES6 even though it's different from the rest of the file?

felixfbecker:
Yes, please only ES6 in ýour added code. Please also use an arrow function on your `map()` call and no `var`.

lib/dialects/mssql/query-generator.js
```diff
@@ -281,25 +281,31 @@ var QueryGenerator = {
 
       allQueries.push(query);
     }
+    const commands = [];
+    let offset = 0;
+    const batch = Math.floor(250 / (allAttributes.length + 1)) + 1;
+    while (offset < Math.max(tuples.length, 1)) {
+      const replacements = {
+        table: this.quoteTable(tableName),
+        attributes: allAttributes.map(function(attr) {
```
felixfbecker: It would be nice if you could turn this in to an arrow function ;)

test/integration/associations/has-many.test.js
```diff
@@ -684,8 +685,26 @@ describe(Support.getTestDialectTeaser('HasMany'), function() {
           expect(users).to.have.length(3);
         });
       });
-    });
 
+      it('handles decent sized bulk creates', function() {
+        var User = this.sequelize.define('User', { username: DataTypes.STRING, num: DataTypes.INTEGER, status: DataTypes.STRING })
+          , Task = this.sequelize.define('Task', { title: DataTypes.STRING });
```
felixfbecker: `const` here

test/integration/associations/has-many.test.js
```diff
@@ -684,8 +685,26 @@ describe(Support.getTestDialectTeaser('HasMany'), function() {
...
+
+        Task.hasMany(User);
+
+        return this.sequelize.sync({ force: true }).then(function() {
```
felixfbecker: arrow functions here

sushantdhiman:
Re-Review Requested @felixfbecker

sushantdhiman:
Look like all ES6 changes as requested are already in done. Merging ....

PR #6918, opened 2016-11-25
Fixes MSSQL Limit in update query, closes #6636

### LINKED ISSUE #6636: MSSQL LIMIT in UPDATE
## What you are doing?

``` js
Account.update(request.body, {
    fields: [
      'nickname',
    ],
    limit: 1,
    where: {
      accountId: request.params.accountId
    }
  }).catch((error => {
    response.body = error;
    response.code = 500;
  }));
```

The generated sql should have the MSSQL equivalent of a limit, in the form of a TOP

``` sql
"UPDATE TOP(1) [Account] SET [Nickname]=N'newNickname' OUTPUT INSERTED.* WHERE [Account_Id] = N'B7068730-59E1-4664-AAAB-A09D2CA10156'"
```
### Generated SQL

``` sql
"UPDATE [Account] SET [Nickname]=N'newNickname' OUTPUT INSERTED.* WHERE [Account_Id] = N'B7068730-59E1-4664-AAAB-A09D2CA10156' LIMIT 1"
```
### Error Text

`"Incorrect syntax near 'LIMIT'."`

I expect the UPDATE query to execute and update just one row, with the assurance that the query will update ONLY 1 row. In the example above my where clause is using the primary key but in other update queries I may be using a where clause which may bring back more than one row. In those instances I want to be able to use the limit option to ensure the query doesn't update beyond the rows I'm expecting to be updated. 

From what I can see in the source, the mssql dialect's version of [query-generator.js](https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/query-generator.js) doesn't implement a method that overrides the behavior of the updateQuery method in [abstract/query-generator.js](https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L293), which is appending the LIMIT keyword to the end of the query. 

**Dialect:** mysql
**Database version:** 12.0.2000
**Sequelize version:** 4.0.0-0

harshithkashyap:
### Pull Request check-list

### Description of change

SQL Server uses TOP(n) keyword to limit the number of updated rows.
https://msdn.microsoft.com/en-us/library/ms177523.aspx

Closes #6636

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -302,6 +302,15 @@ var QueryGenerator = {
     return generatedQuery;
   },
 
+  updateQuery(tableName, attrValueHash, where, options, attributes) {
+    let sql = super.updateQuery.call(this, tableName, attrValueHash, where, options, attributes);
+    if (options.limit) {
+      const updateArgs = `UPDATE TOP(${this.escape(options.limit)})`;
+      sql = sql.replace('UPDATE', updateArgs);
+    }
+    return sql;
+  },
+
   upsertQuery(tableName, insertValues, updateValues, where, rawAttributes, options) {
     const targetTableAlias = this.quoteTable(`${tableName}_target`);
     const sourceTableAlias = this.quoteTable(`${tableName}_source`);
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -303,7 +303,7 @@ var QueryGenerator = {
   },
 
   updateQuery(tableName, attrValueHash, where, options, attributes) {
-    let sql = super.updateQuery.call(this, tableName, attrValueHash, where, options, attributes);
+    let sql = super.updateQuery(tableName, attrValueHash, where, options, attributes);
     if (options.limit) {
       const updateArgs = `UPDATE TOP(${this.escape(options.limit)})`;
       sql = sql.replace('UPDATE', updateArgs);
--- a/test/unit/dialects/mssql/query-generator.test.js
+++ b/test/unit/dialects/mssql/query-generator.test.js
@@ -45,6 +45,46 @@ if (current.dialect.name === 'mssql') {
       });
     });
 
+    test('updateQuery', function () {
+      expectsql(QueryGenerator.updateQuery('myTable', {name: 'foo', birthday: new Date(Date.UTC(2011, 2, 27, 10, 1, 55))}, { id: 2 }), {
+        mssql: "UPDATE [myTable] SET [name]=N'foo',[birthday]='2011-03-27 15:31:55.000' OUTPUT INSERTED.* WHERE [id] = 2"
+      });
+
+      expectsql(QueryGenerator.updateQuery('myTable', {bar: 2}, {name: 'foo'}), {
+        mssql: "UPDATE [myTable] SET [bar]=2 OUTPUT INSERTED.* WHERE [name] = N'foo'"
+      });
+
+      expectsql(QueryGenerator.updateQuery('myTable', {bar: 2, nullValue: null}, {name: 'foo'}), {
+        mssql: "UPDATE [myTable] SET [bar]=2,[nullValue]=NULL OUTPUT INSERTED.* WHERE [name] = N'foo'"
+      });
+
+      expectsql(QueryGenerator.updateQuery('myTable', {bar: 2, nullValue: null}, {name: 'foo'}, { omitNull: false }), {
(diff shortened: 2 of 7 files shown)


test/unit/dialects/mssql/query-generator.test.js
```diff
@@ -45,6 +45,46 @@ if (current.dialect.name === 'mssql') {
...
+      //Limit
+      expectsql(QueryGenerator.updateQuery('myTable', { foo: 'bar' }, {}, { limit: 1 }), {
+        mssql: "UPDATE TOP(1) [myTable] SET [foo]=N'bar' OUTPUT INSERTED.*"
+      });
+
+      expectsql(QueryGenerator.updateQuery('myTable', { foo: 'bar' }, { foo: 'value' }, { limit: 1 }), {
+        mssql: "UPDATE TOP(1) [myTable] SET [foo]=N'bar' OUTPUT INSERTED.* WHERE [foo] = N'value'"
+      });
+
+    });
```
mickhansen: the query-generator.test files are deprecated, please look at test/unit/sql instead

test/unit/sql/update.js
```diff
@@ -34,5 +34,28 @@ describe(Support.getTestDialectTeaser('SQL'), function() {
...
+      }, {
+        timestamps: false
+      });
+
+      expectsql(sql.updateQuery(User.tableName, { username: 'new.username' }, { username: 'username' }, { limit: 1 }),
+      {
+        mssql: "UPDATE TOP(1) [Users] SET [username]=N'new.username' OUTPUT INSERTED.* WHERE [username] = N'username'",
+        postgres: "UPDATE \"Users\" SET \"username\"=\'new.username\' WHERE \"username\" = \'username\'",
+        sqlite: "UPDATE `Users` SET `username`='new.username' WHERE `username` = 'username' LIMIT 1",
+        default: "UPDATE `Users` SET `username`=\'new.username\' WHERE `username` = \'username\' LIMIT 1"
```
mickhansen: sqlite matches the default so no reason to have a seperate case, use `[]` like the other tests do to have dialect specfic quotesi n there.

mickhansen:
Since this pretty much puts something in the front of the query, perhaps add the specific code to the mssql query generator, call the generic one from mssql and then replace `UPDATE` with `UPDATE TOP(${limit})`

test/unit/sql/update.js
harshithkashyap: Right, I'll go ahead and remove sqlite. I suppose postgres is also not required as it doesn't support limit on update and would match the default case?
Sorry, didn't get the `[]` & using dialect specific quotes part. Do the expectation string go through processing before comparison as Sql Server needs `[]` and mysql needs back ticks?

janmeier: `default: "UPDATE [Users] SET [username]='new.username' WHERE [username] = 'username' LIMIT 1",`

[ and ] are then replaced with the quotes for that specific dialect (or kept as is for mssql :) )

mickhansen: IIRC postgres supports limit on update via a subquery.

lib/dialects/mssql/query-generator.js
```diff
@@ -302,6 +302,15 @@ var QueryGenerator = {
     return generatedQuery;
   },
 
+  updateQuery(tableName, attrValueHash, where, options, attributes) {
+    let sql = AbstractQueryGenerator.updateQuery.call(this, tableName, attrValueHash, where, options, attributes);
```
felixfbecker: I think you should just be able to call `super.updateQuery()`

harshithkashyap: Yes, thanks :) Updated.

harshithkashyap:
`npm run docs` seems to be broken. Is this expected on v4?

janmeier:
Yep, docs are automagically generated by doclets for v4

harshithkashyap:
@janmeier Any other changes needed?

lib/dialects/mssql/query-generator.js
```diff
@@ -302,6 +302,15 @@ var QueryGenerator = {
     return generatedQuery;
   },
 
+  updateQuery(tableName, attrValueHash, where, options, attributes) {
+    let sql = super.updateQuery.call(this, tableName, attrValueHash, where, options, attributes);
```
felixfbecker: You can just do `super.updateQuery(tableName, ...)`

harshithkashyap:
@janmeier Could you please merge this?

janmeier:
Thanks!

PR #6921, opened 2016-11-27
Optimize ForInStatement with non local each variable

TylerBrock:
### Pull Request check-list

### Description of change

I noticed that V8 could not optimize Instance.prototype.set when doing some profiling:

profile

This fixes the problem by making the each variable for the for-in loop a local one (and renaming it to satisfy the linter). I noticed this is fixed in v4 but since it's not out I figured it would be nice to optimize this hot path in v3.

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -289,8 +289,8 @@ Instance.prototype.set = function(key, value, options) { // testhint options:non
           }
         }
       } else {
-        for (key in values) {
-          this.set(key, values[key], options);
+        for (var objKey in values) {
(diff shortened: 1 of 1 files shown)


felixfbecker:
v4 will be released soon. Besides that, this change will _redeclare_ the (function-scoped) variable _at every iteration_. I doubt that this has better performance.

TylerBrock:
I have seen that it does have an impact because I think not having the variable be local in that loop prevents the entire set function from being optimized. You might be right though. Thanks!

Issue #4597, opened 2015-10-01
Naming collision in N-M options


glutentag:
When using a N-M association accessor method, addAssociation for instance, the signature is:

```
(newInstances, additionalAttributes)
```

transformed in

```
options = additionalAttributes
```

Thus, options is used in all queries (association.through.model.findAll), but also as additional attribute when creating the association through model. It can create bugs when this attribute is a keyword, 'order' for instance.

Could it be possible to separate in the signature additionalAttributes from options?

janmeier:
Yeah, I've actually been thinking about the same thing - it would also make the code easier to maintain.

`options.through` would probably be best, in order to fit with the rest of the API

mickhansen:
I've never liked this piece of the code. But from the end user it made sense if you never used options for the setters and didnt want to pass an empty object/null.

subjectix:
I think what @janmeier proposed is the ideal solution but it introduces breaking changes (if someone has a join table with through as a column). So it will have to be fixed with the next major release.

So you might wanna consider this too:
Since every association mixin has options for the second argument, might it look ugly to modify the first argument? I mean something like this:

```js
// No through attributes.
source.addTarget(target, options);

// With through attributes.
source.addTarget({
    target: target, // Instance or PK
    through: {/* through attributes*/},
}, options);
```

I'd say this is easier to achieve without introducing breaking changes. All you have to do is take the extra attributes from options and assign the `firstArg.through` with them:
```js
// Values from the firstArg.through must override the attributes from extraAttribsFromOptions.
let throughAttributes = Object.assign({} , extraAttribsFromOptions, firstArg.though);
```
This will not introduce breaking changes, make it easy to add deprecation warnings, and keep the api consistent. As in: *"All data come from the first argument, all options come from the second argument"*.
What users will be faced with is this:

```js
// ---------------------------------------------------
// With through attributes. Without options:
// Before:
user.addProject(project, {role: 'admin'});
// After:
// OK
user.addProject({
    target: project,
    through: {role: 'admin'}
});
// OK but with deprecation warning: 
// "role is defined on the join table. This approach is deprecated please see blah blah blah..."
user.addProject(project, {role: 'admin'});


// ---------------------------------------------------
// With through attributes. With options:
// Before:
user.addProject(project, {role: 'admin', validate: true});
// After:
// OK
user.addProject({
    target: project,
    through: {role: 'admin'}
}, {validate: true});
// OK but with deprecation warning: 
[...]

PR #6931, opened 2016-11-30
Feat 4597 : Pass through values using options.through for N:M relationships

sushantdhiman:
### Pull Request check-list

### Description of change

This is one of the two changes discussed in #4597 , Now `N:M` junction table attributes can be specified by `set/add/created` calls by passing `options.through`. This will prevent polluting the `options` passed to `findAll` or `update` as well.

I will add change log after PR reviews, when its approved for merge.


mickhansen:
A much welcome change, changelog and it's good to go i reckon.

PR #6943, opened 2016-12-03
Fix: custom msg not used for notNull validations

### LINKED ISSUE #6531: Custom null validation errors in v4 don't seem to work
I can't seem to get custom null validation errors working in v4. (This may not be exclusive to v4; I haven't tried it in other versions).

I've tried several field definitions, including:

``` javascript
{
  email: {
    type: Sequelize.STRING,
    allowNull: false,
    validate: {
      notNull: {
        msg: 'Please enter your e-mail',
      },
      isEmail: {
        msg: 'Please enter a valid e-mail'
      }
    },
  },
}
```

... and variants, including setting `allowNull: true` and custom functions inside of `validate`.

**Expected:**

Using this set-up...

``` javascript
const user = new User;
user.email = args.email // null value fed in from a form
return user.save()
```

... or this...

``` javascript
return (new User(args)).save()
```

... I'm expecting to throw `Sequelize.ValidationError` with the following message:

> Please enter your e-mail

**Actual:**

This is the actual error I'm getting back:

> email cannot be null

**Workaround:**

Setting the field to any non-null value works.

``` javascript
const user = new User({ email: ''});
return user.save(); // now it'll throw the correct error
```

The issue with this approach is that I specifically need to set every value to be non-null before `validate` rules kick-in, which means adding a middleware step before processing form input.

It could also mean (not yet tested) that fields that use non-strings need to be initialised with a value other than `''` to work, which is a further workaround.

sushantdhiman:
### Pull Request check-list

- [x] Does your issue contain a link to existing issue (Closes #6531]) or a description of the issue you are solving?

### Description of change
Closes #6531 

Will add the changelog after reviews

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,5 @@
 # Future
+- [FIXED] Custom error message not used for `notNull` validation [#6531](https://github.com/sequelize/sequelize/issues/6531)
 - [FIXED] N:M `through` option naming collisions [#4597](https://github.com/sequelize/sequelize/issues/4597)
  [#6444](https://github.com/sequelize/sequelize/issues/6444)
 - [CHANGED] Updated deprecated `node-uuid` package to `uuid` [#6919](https://github.com/sequelize/sequelize/pull/6919)
--- a/lib/instance-validator.js
+++ b/lib/instance-validator.js
@@ -304,14 +304,18 @@ class InstanceValidator {
   _validateSchema(rawAttribute, field, value) {
     let error;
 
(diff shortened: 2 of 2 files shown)


thanpolas:
I see you are using a different character for quotes @sushantdhiman check the diff and pay attention to the quotes, please use the standard single quote `'`.

sushantdhiman:
@thanpolas they are template strings :stuck_out_tongue:  `${something}` wont work with normal single quotes

PR #6944, opened 2016-12-03
Fix #3899 no value in built-in/custom validation errors

### LINKED ISSUE #3899: ValidationErrorItem value property does not contain "The value that generated the error"
According to the docs, the `value` property of the [ValidationErrorItem(message, type, path, value)](http://docs.sequelizejs.com/en/latest/api/errors/#new-validationerroritemmessage-type-path-value) object should contain "The value that generated the error", but instead it contains the error message.

```
Database error {
   [SequelizeValidationError: Validation error: Password must be between 9 and 24 characters in length]
   message: 'Validation error: Password must be between 9 and 24 characters in length',
   errors: [ {
      message: 'Password must be between 9 and 24 characters in length',
      type: 'Validation error',
      path: 'password',
      value: 'Password must be between 9 and 24 characters in length',
      __raw: 'Password must be between 9 and 24 characters in length'
   } ]
}
```

sushantdhiman:
### Pull Request check-list

### Description of change

Fixes #3899 , Error object for validation error was always missing `value` field, which is important for front end validation etc. It was broken for `custom` and `builtIn` validators. Now its fixed, regression included

Will add change log after reviews

### DIFF
--- a/lib/instance-validator.js
+++ b/lib/instance-validator.js
@@ -198,7 +198,9 @@ class InstanceValidator {
       validators.push(validatorPromise.reflect());
     });
 
-    return Promise.all(validators).then(this._handleReflectedResult.bind(this, field, value));
+    return Promise
+            .all(validators)
+            .then(results => this._handleReflectedResult(field, value, results));
   }
 
   /**
@@ -239,7 +241,7 @@ class InstanceValidator {
               .catch(e => this._pushError(false, errorKey, e, optValue));
     } else {
       return Promise
-              .try(validator.bind(this.modelInstance, invokeArgs))
+              .try(() => validator.call(this.modelInstance, invokeArgs))
               .catch(e => this._pushError(false, errorKey, e, optValue));
     }
   }
--- a/lib/instance-validator.js
+++ b/lib/instance-validator.js
@@ -198,7 +198,7 @@ class InstanceValidator {
       validators.push(validatorPromise.reflect());
     });
 
-    return Promise.all(validators).then(this._handleReflectedResult.bind(this, field));
+    return Promise.all(validators).then(this._handleReflectedResult.bind(this, field, value));
   }
 
   /**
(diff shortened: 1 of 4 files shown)


lib/instance-validator.js
```diff
@@ -198,7 +198,7 @@ class InstanceValidator {
       validators.push(validatorPromise.reflect());
     });
 
-    return Promise.all(validators).then(this._handleReflectedResult.bind(this, field));
+    return Promise.all(validators).then(this._handleReflectedResult.bind(this, field, value));
```
felixfbecker: We can do this with an arrow function instead of `bind()`, better performance

lib/instance-validator.js
```diff
@@ -235,9 +235,12 @@ class InstanceValidator {
       } else {
         validatorFunction = Promise.promisify(validator.bind(this.modelInstance));
       }
-      return validatorFunction().catch(this._pushError.bind(this, false, errorKey));
+      return validatorFunction()
+              .catch(e => this._pushError(false, errorKey, e, optValue));
     } else {
-      return Promise.try(validator.bind(this.modelInstance, invokeArgs)).catch(this._pushError.bind(this, false, errorKey));
+      return Promise
+              .try(validator.bind(this.modelInstance, invokeArgs))
```
felixfbecker: same here, `try(() => validator.call(this.modelInstance, invokeArgs))` should be faster.
Btw, is `invokeArgs` an array? Because `bind()` doesn't take an array of arguments, it is variadic

sushantdhiman: > Btw, is invokeArgs an array? Because bind() doesn't take an array of arguments, it is variadic

Not its just single value from validator.args (which can be array), but its correctly passed, so no worries :)

felixfbecker: Okay, the name confused me 👍

PR #6963, opened 2016-12-08
Fix getting true when upsert a row to its current value with mysql for Sequelize v3

### LINKED ISSUE #6310: upsert return wrong value when insert a row which db has

daleoooo:
Fixed #6310 : added flags: '-FOUND_ROWS' at createConnection in mysql dialect.

This is the PR for v3 that I currently use.

Thanks

### DIFF
--- a/lib/dialects/mysql/connection-manager.js
+++ b/lib/dialects/mysql/connection-manager.js
@@ -57,6 +57,7 @@ ConnectionManager.prototype.connect = function(config) {
       host: config.host,
       port: config.port,
       user: config.username,
+      flags: '-FOUND_ROWS',
       password: config.password,
       database: config.database,
       timezone: self.sequelize.options.timezone,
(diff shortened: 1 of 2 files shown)


daleoooo:
@janmeier 
Please merge 😀

janmeier:
:+1:

Issue #5756, opened 2016-04-15
describeTable returns a wrong value for primaryKey


bumpmann:
## What you are doing?

I m using this table:

```
CREATE TABLE speaker
(
  id serial NOT NULL,
  name text NOT NULL,
  company text NOT NULL,
  "createdAt" timestamp with time zone NOT NULL,
  "updatedAt" timestamp with time zone NOT NULL,
  CONSTRAINT speaker_pkey PRIMARY KEY (id),
  CONSTRAINT speaker_name_key UNIQUE (name)
)
```

I need to get the primary key(s) name:

``` js
queryInterface.describeTable('speaker').then(function(infos) {
  console.log(JSON.stringify(info,null,' '))
})
```
## What is actually happening?

The column `name` has its field `primaryKey` set to `true` **instead of `false`**

```
{
 "id": {
  "type": "INTEGER",
  "allowNull": false,
  "defaultValue": "nextval(speaker_id_seq::regclass)",
  "special": [],
  "primaryKey": true
 },
 "name": {
  "type": "TEXT",
  "allowNull": false,
  "defaultValue": null,
  "special": [],
  "primaryKey": true
 },
 "company": {
  "type": "TEXT",
  "allowNull": false,
  "defaultValue": null,
  "special": [],
  "primaryKey": false
 },
 "createdAt": {
  "type": "TIMESTAMP WITH TIME ZONE",
  "allowNull": false,
  "defaultValue": null,
  "special": [],
  "primaryKey": false
 },
 "updatedAt": {
  "type": "TIMESTAMP WITH TIME ZONE",
  "allowNull": false,
  "defaultValue": null,
  "special": [],
  "primaryKey": false
 }
}
```

**Dialect:** postgres
**Database version:** PostgreSQL 9.4.4
**Sequelize version:** 3.21.0

sushantdhiman:
Hi @janmeier @mickhansen 

Trying to solve this issue one thing occur to me, Isn't a field with `NOT NULL` and `UNIQUE` is a `PRIMARY_KEY` ?

Because even in Postgres `information_schema` its stored as `PRIMARY_KEY`. I can apply a check for `_result.Null === 'NO'` [here](https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/query.js#L212) but it will essentially qualify as a primary key.

Your thoughts?

mickhansen:
@sushantdhiman Not necessarily, you can have multiple uniques per table but only one primary key.

[...]

PR #6966, opened 2016-12-09
Fix "describeTable returns a wrong value for primaryKey" #5756

abegines:
### Pull Request check-list

### Description of change

* Changed SQL JOINS (postgres, mssql) and `!==0` instead of `===1` (sqlite) to correct this issue:
describeTable returns a wrong value for primaryKey #5756

* Created a new test 'should correctly determine the primary key columns'
Tested on sqlite, mssql, mysql & postgres.

Saludos,
abc


changelog.md
```diff
@@ -1,4 +1,5 @@
 # Future
+- [FIXED] describeTable returns a wrong value for primaryKey [#5736] (https://github.com/sequelize/sequelize/issues/5756)
```
sushantdhiman: Hash and link doesn't match

janmeier:
Sweet, thanks @abegines

PR #6970, opened 2016-12-10
Don't lose global defaultScope `attributes` after first define()

broofa:
### Pull Request check-list

(I am currently blocked on testing due to `npm install` failing with `gyp: Call to 'pg_config --libdir' returned exit status 127 while in binding.gyp. while trying to load binding.gyp`.  Any idea what I'm doing wrong?)

### Description of change

https://gist.github.com/broofa/cee97d28e3923ad62438561438f7e12a is a standalone test case showing how sequelized.options.define.defaultScope.options.attributes is overwritten when defining a model.    It logs the state of of this property before, between, and after doing two define() calls.  The output I see (Sequelized v3.27.0) is this:

```
kieffer@Roberts-MacBook-Pro$ node defaultScope_test.js
attributes before defining ModelA: { exclude: [ 'createdAt', 'updatedAt', 'deletedAt' ] }
attributes after defining ModelA: [ 'modelAUuid', 'fooA', 'barA' ]
attributes after defining ModelB: [ 'modelAUuid', 'fooA', 'barA' ]
```

Cloning defaultScope in Model.init() as per this PR fixes this, however I'm not 100% sure this is the correct solution.  (If nothing else, doing a deep-copy would probably be more resistent to similar bugs emerging in the future).


broofa:
Oh, for fucks sake.  This is supposed to be a one-line change against `v3.27.0`.  Is there a way to diff against that tag???

broofa:
Grr. FML.

felixfbecker:
@broofa of course there is a way to diff against v3, you need to target the v3 branch, not master. But we need a master PR first before the backport

felixfbecker:
I must be based on latest v3 though.

PR #6971, opened 2016-12-10
Clone defaultScope so it doesn't get overwritten in Model#init

broofa:
### Pull Request check-list

(I have not been able to test against `master` due to  `npm install` failing with `gyp: Call to 'pg_config --libdir' returned exit status 127 while in binding.gyp.  However this is reproduceable with latest version in npm and the code seems largely unchanged, so I'm assuming this is still an issue...?)

### Description of change

https://gist.github.com/broofa/cee97d28e3923ad62438561438f7e12a is a standalone test case showing how sequelized.options.define.defaultScope.options.attributes is overwritten when defining a model.    It logs the state of of this property before, between, and after doing two define() calls.  The output I see (Sequelized v3.27.0) is this:

```
kieffer@Roberts-MacBook-Pro$ node defaultScope_test.js
attributes before defining ModelA: { exclude: [ 'createdAt', 'updatedAt', 'deletedAt' ] }
attributes after defining ModelA: [ 'modelAUuid', 'fooA', 'barA' ]
attributes after defining ModelB: [ 'modelAUuid', 'fooA', 'barA' ]
```

Cloning defaultScope in Model.init() as per this PR fixes this, however I'm not 100% sure this is the correct solution.  (If nothing else, doing a deep-copy would probably be more resistent to similar bugs emerging in the future).

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -685,9 +685,7 @@ class Model {
 
     const globalOptions = this.sequelize.options;
 
-    if (globalOptions.define) {
-      options = Utils.merge(globalOptions.define, options);
-    }
+    options = Utils.merge(_.cloneDeep(globalOptions.define), options);
 
     if (!options.modelName) {
       options.modelName = this.name;


broofa:
Verified this is still an issue in v4.0.0-2

lib/model.js
```diff
@@ -809,7 +809,7 @@ class Model {
 
     this._findAutoIncrementField();
 
-    this._scope = this.options.defaultScope;
+    this._scope = Object.assign({}, this.options.defaultScope);
```
felixfbecker: should this be a deep clone?

felixfbecker:
There are two test failures

broofa:
Hmm.  Integration check failing with `"before each" hook for "should uppercase the first letter in alias getter, but not in eager loading"`... but I see the same test failure on the `master` branch, and it seems unlikely to be related to this PR.  :-/

In other news, I stumbled across [this comment of your](https://github.com/sequelize/sequelize/blob/master/lib/utils.js#L65) on `Utils.merge`, which I'm not  sure what to make of.  Is cloning `globalOptions` like this going to run afoul of whatever that issue is?

felixfbecker:
That's not my comment, and it should not be relevant for `defaultScope`

broofa:
@felixfbecker, sorry about that.  I tracked that comment down to [this commit](https://github.com/sequelize/sequelize/commit/ecbc18394f61a8a3258b41f5780f3183eb185c0d#diff-50e3aa130a4f97a42ee2cf111c7b1d9dR45), attributed to @greenkeeperio-bot and @janmeier.

So... what's up with that failing integration check?  On this project's home page the Travis badge says "build failing", and links [here](https://travis-ci.org/sequelize/sequelize)... which calls out this PR.  Is Travis testing against all outstanding PRs or something?  My apologies if that's a noob question.

... but, as I said above, I see the same error on master (w/out my PR).

lib/model.js
```diff
@@ -683,7 +683,7 @@ class Model {
     }
     this.sequelize = options.sequelize;
 
-    const globalOptions = this.sequelize.options;
+    const globalOptions = _.cloneDeep(this.sequelize.options);
```
felixfbecker: It appears to me that `globalOptions` is not used besides in L688. if that is the case, I think we could just modify L689 to read
```js
options = Utils.merge({}, globalOptions.define, options)
```

felixfbecker:
Restarted the jobs and builds are now fixed

broofa:
Utils.merge copies-by-reference [when the target object is undefined](https://github.com/sequelize/sequelize/blob/master/lib/utils.js#L74), which turns out to not solve the original problem.  But _.cloneDeep(globalOptions.define) does the trick.

lib/model.js
```diff
@@ -676,18 +676,14 @@ class Model {
    * @return {Model}
    */
   static init(attributes, options) { // testhint options:none
-    options = options || {};
```
felixfbecker: `options` parameter must stay optional, please keep this line

lib/model.js
```diff
@@ -676,18 +676,14 @@ class Model {
...
       throw new Error('No Sequelize instance passed');
     }
     this.sequelize = options.sequelize;
 
     const globalOptions = this.sequelize.options;
 
-    if (globalOptions.define) {
-      options = Utils.merge(globalOptions.define, options);
-    }
+    options = Utils.merge(_.cloneDeep(globalOptions.define), options);
```
felixfbecker: What happens now if `globalOptions.define` is `undefined`? I think previous version was better

lib/model.js
broofa: FWIW, this is functionally the same since `options = {}`, fails the `options.sequelize` check at L681 (precommit, below), but I'll revert it.

lib/model.js
broofa: Integration test failing again for unrelated reasons?

PR #6972, opened 2016-12-10
Transactions for SQL Server

### LINKED ISSUE #3950: Transaction name too long for MSSQL
I'm attempting to persist some records within a transaction to a MSSQL database and I'm coming across the following error:

```
SequelizeDatabaseError: The identifier that starts with '1e340e20-26a3-4881-ac71-1326b01f5af4-savepoint-1' is too long. Maximum length is 32.
```

This appears to be because per the MSSQL spec (https://msdn.microsoft.com/en-us/library/ms188929.aspx):

```
transaction_name
Is the name assigned to the transaction. transaction_name must conform to the rules for identifiers, but identifiers longer than 32 characters are not allowed. Use transaction names only on the outermost pair of nested BEGIN...COMMIT or BEGIN...ROLLBACK statements. transaction_name is always case sensitive, even when the instance of SQL Server is not case sensitive.
```

The transaction name originally appears to be a generated uuid but gets the 'savepoint' text appended to it at:
https://github.com/sequelize/sequelize/blob/2.0/lib/transaction.js#L23

I'm not familiar with other character limits on transaction names in databases but the same code works as intended on Postgres.

I'm using v2.0.0 of Sequelize. Thanks in advance.

harshithkashyap:
### Pull Request check-list

### Description of change

- Based on @mbroadst [suggestion](https://github.com/sequelize/sequelize/issues/3950#issuecomment-120933650), generateTransactionId method is added which uses `crypto.randomBytes` and `uuid.v4` in case of other dialects. Closes #3950 
- Enabled tests for transactions for mssql dialect
- Added `SAVE TRANSACTION` check to call tedious.saveTransaction method
- Fix tests which are skipped


harshithkashyap:
@felixfbecker @janmeier I've skipped a lot of tests which were timing out. By default, SQL Server blocks even select calls if the row has an exclusive lock. Most of the tests in the codebase fails as transaction is never committed/rolledback.

How do you guys prefer to fix this? I'm checking if the dialect is mssql and using `Promise.resolve` to skip the select statement in some tests. There are a lot of tests based on this approach as Postgresql and Mysql only blocks writes and allows reads.

felixfbecker:
Well cheating the tests is not an option of course. Maybe just commit the transaction before the SELECT?

harshithkashyap:
But would that not cause the subsequent `then` calls to fail as it tries to use a transaction which is already committed?

harshithkashyap:
@mickhansen @sushantdhiman Suggestions?

harshithkashyap:
@felixfbecker @janmeier SQL Server allows `READ_COMMITTED_SNAPSHOT` option to allow reads during transactions. I've enabled back all the tests and all of the pass in my localhost.
`READ_COMMITTED_SNAPSHOT` must be turned on before we can use it and it cannot be enabled on master database. I've changed the database in appveyor-setup.ps1 to sequelize_test but i think it has to be manually created on your end. Could you guys please enable it and review this PR?

felixfbecker:
@harshithkashyap What do you mean with "it has to be enabled on our end"? AppVeyor is 100% configured through appveyor.yml and appveyor-setup.ps1

harshithkashyap:
I've changed the database from master to sequelize_test [here](https://github.com/harshithkashyap/sequelize/blob/b7593555a2155e2a025e255193cabfb683ad70f5/appveyor-setup.ps1#L23). I expected appveyor to create the `sequelize_test` database automatically with the credentials from appveyor-setup.ps1 file. 
Login failed error is thrown in the [build](https://ci.appveyor.com/project/sushantdhiman/sequelize/build/701/job/hbseplv72gf5bgmh#L1244).

felixfbecker:
hm no idea, im no mssql expert

harshithkashyap:
@felixfbecker I've fixed it using `appveyor-setup.ps script`

harshithkashyap:
@janmeier @sushantdhiman @felixfbecker Request review!

harshithkashyap:
@janmeier Without tedious methods, sql server throws

> Transaction count after EXECUTE indicates a mismatching number of BEGIN and COMMIT statements. Previous count = 0, current count = 1.

We use `connection.execSql(request);` [here](https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/query.js#L106). When beginTransaction and other methods are called, tedious internally calls `connection.execSqlBatch` instead of `connection.execSql`. 

Using `connection.execSqlBatch` allows us to execute sequelize generated queries. All tests seems to pass aswell. 

From the tedious [docs](http://tediousjs.github.io/tedious/api-connection.html#function_execSqlBatch), it seems that SQL Server will not reuse the execution plan if the same queries are executed again and again. This might add overhead to non transaction queries.

lib/dialects/mssql/query.js
```diff
@@ -67,16 +67,23 @@ class Query extends AbstractQuery {
           } else {
             resolve(this.formatResults());
           }
-        });
+        }, this.options.transaction.name);
+      } else if (_.includes(this.sql, 'SAVE TRANSACTION')) {
+        connection.saveTransaction(err => {
+          if (!!err) {
```
harshithkashyap: @felixfbecker `beginTransaction`, `rollbackTransaction` and `commitTransaction` also has boolean cast. Should i remove them aswell?

felixfbecker: Feel free to do so. The desired coding style is defined in eslintrc.json, it's just that the codebase does not fully comply with it yet.

test/config/config.js
```diff
@@ -29,7 +29,6 @@ module.exports = {
     host:     process.env.SEQ_MSSQL_HOST || process.env.SEQ_HOST || 'mssql.sequelizejs.com',
     port:     process.env.SEQ_MSSQL_PORT || process.env.SEQ_PORT || 1433,
     dialectOptions: {
-      instanceName:  process.env.MSSQL_INSTANCE ? process.env.MSSQL_INSTANCE : 'SQLEXPRESS',
```
felixfbecker: Why is this not needed anymore?

harshithkashyap: By default instance names are unnamed. I cannot connect to my local server even after specifying the server name and port. And it doesn't default to SQLEXPRESS

felixfbecker: Okay, was just curious. I'm not familiar with MSSQL, I just ported the thing over to AppVeyor and copied a lot from tedious PowerShell setup script.

janmeier:
Good work @harshithkashyap ! :)

PR #6996, opened 2016-12-16
Closes #6095

### LINKED ISSUE #6095: hasMany: destroyAssociations
Hello,

I'm trying to figure out if there is a way to remove associations from table instead of setting foreign key to null when i call `setAssociations(null)`.
For example, if i have Author and Post models:

``` js
const Author = Sequelize.define('author', {
  name: Sequelize.STRING
});

const Post = Sequelize.define('post', {
  title: Sequelize.STRING
});

Author.hasMany(Post);
Post.belongsTo(Author);
```

And i want to remove all posts for this author, I would like to use something like this:

``` js
const author = Author.findById(1);
author.setPosts([]).then(() => ...);
```

But the problem is that this just sets `authorId` to NULL in `posts` table. How can i literally remove it from the table?

Thanks.

afucher:
### Pull Request check-list

### Description of change

Added method to destroy all associations from through model.
Closes [#6095]

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,5 @@
 # Future
+- [FIXED] Create method to destroy all association from through model [#6095] (https://github.com/sequelize/sequelize/6095)
 - [FIXED] describeTable returns a wrong value for primaryKey [#5756] (https://github.com/sequelize/sequelize/issues/5756)
 - [FIXED] MSSQL LIMIT IN UPDATE [#6636](https://github.com/sequelize/sequelize/issues/6636)
 - [FIXED] Custom error message not used for `notNull` validation [#6531](https://github.com/sequelize/sequelize/issues/6531)
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,5 @@
 # Future
-- [FIXED] Create method to destroy all association from through model [#6095] (https://github.com/sequelize/sequelize/6095)
+- [ADDED] Create method to destroy all association from through model [#6095] (https://github.com/sequelize/sequelize/6095)
 - [FIXED] Issue with sync hooks (before/afterInit, before/afterDefine) [#6680](https://github.com/sequelize/sequelize/issues/6680)
 - [FIXED] MSSQL handle large bulk inserts [#6866](https://github.com/sequelize/sequelize/issues/6866)
 - [FIXED] describeTable returns a wrong value for primaryKey [#5756] (https://github.com/sequelize/sequelize/issues/5756)
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -287,6 +287,16 @@ class BelongsToMany extends Association {
        */
       removeMultiple: 'remove' + plural,
       /**
+       * Un-associate all instances.
+       *
+       * @param {Object} [options] Options passed to `through.destroy`
+       * @return {Promise}
+       * @method removeAssociations
+       * @memberof Associations.BelongsToMany
+       * @instance
+       */
+      destroy: 'destroy' + plural,
+      /**
        * Check if an instance is associated with this.
        *
        * @param {Model|String|Number} [instance] Can be an Instance or its primary key
@@ -449,7 +459,7 @@ class BelongsToMany extends Association {
   }
 
   mixin(obj) {
-    const methods = ['get', 'count', 'hasSingle', 'hasAll', 'set', 'add', 'addMultiple', 'remove', 'removeMultiple', 'create'];
+    const methods = ['get', 'count', 'hasSingle', 'hasAll', 'set', 'add', 'addMultiple', 'remove', 'removeMultiple', 'create', 'destroy'];
     const aliases = {
       hasSingle: 'has',
       hasAll: 'has',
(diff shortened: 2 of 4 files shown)


lib/associations/belongs-to-many.js
```diff
@@ -287,6 +287,16 @@ class BelongsToMany extends Association {
        */
       removeMultiple: 'remove' + plural,
       /**
+       * Un-associate all instances.
```
felixfbecker: Misleading comment

lib/associations/belongs-to-many.js
```diff
@@ -287,6 +287,16 @@ class BelongsToMany extends Association {
...
+       *
+       * @param {Object} [options] Options passed to `through.destroy`
+       * @return {Promise}
+       * @method removeAssociations
+       * @memberof Associations.BelongsToMany
+       * @instance
+       */
+      destroy: 'destroy' + plural,
```
felixfbecker: Should be `destroyMultiple`, and a singular one would be useful too

test/integration/associations/belongs-to-many.test.js
```diff
@@ -1715,6 +1715,32 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
           expect(tasks.length).to.equal(1);
         });
       });
+      it('should remove all entries on the through model', function() {
```
felixfbecker: ~~remove~~destroy

test/integration/associations/belongs-to-many.test.js
```diff
@@ -1715,6 +1715,32 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
...
+        }).spread(function(worker, tasks) {
+          return worker.setTasks(tasks).then(function() {
+            return worker.destroyTasks();
+          }).then(function() {
+            return worker.getTasks();
+          });
+        }).then(function(tasks) {
+          expect(tasks.length).to.equal(0);
+        });
+      });
```
felixfbecker: Please use ES6 where possible. Arrow functions, `const`

test/integration/associations/belongs-to-many.test.js
```diff
@@ -1715,6 +1715,32 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
...
+            })
+          ]);
+        }).spread(function(worker, tasks) {
+          return worker.setTasks(tasks).then(function() {
+            return worker.destroyTasks();
+          }).then(function() {
+            return worker.getTasks();
+          });
+        }).then(function(tasks) {
+          expect(tasks.length).to.equal(0);
```
felixfbecker: Iirc there is an assertion in chai for array lengths

lib/associations/belongs-to-many.js
```diff
@@ -287,6 +287,16 @@ class BelongsToMany extends Association {
        */
       removeMultiple: 'remove' + plural,
       /**
+       * Un-associate all instances.
+       *
+       * @param {Object} [options] Options passed to `through.destroy`
```
felixfbecker: The API should be like `removeMultiple`, e.g. it should accept an array of instances.
We could make it optional for `removeMultiple` as well and remove/destroy all associated if no array is passed.

changelog.md
```diff
@@ -1,4 +1,5 @@
 # Future
+- [FIXED] Create method to destroy all association from through model [#6095] (https://github.com/sequelize/sequelize/6095)
```
felixfbecker: [ADDED]

lib/associations/belongs-to-many.js
afucher: I should only modify the removeMultiple, and not create this destroyAssociation? If I do this, if someone use the removeMultiple and pass an empty array, the result would be different and will cause a lot of non expected results.
I just create another method, because the issue say that.

lib/associations/belongs-to-many.js
afucher: What do you think about: "Un-associate current from all"

lib/associations/belongs-to-many.js
felixfbecker: Since there is `removeAssociation(instance)`, I would expect a `destroyAssociation(instance)` to exist too

lib/associations/belongs-to-many.js
felixfbecker: I just said the API should be similar, accepting the same parameters. @mickhansen @janmeier?

afucher: Hmm the `destroyAssociation` accept the same parameters? I understand that you are saying that `removeMultiple` should destroy all associations if no array is passed.

felixfbecker: Yeah, but all other methods accept an array of instances and only set/remove those exact instances:
```js
setAssociations(instances)
removeAssociations(instances)
```
why does `destroyAssociations` then simply destroy _all_ instances, and not only the ones passed? That confuses me

sushantdhiman:
I think we only should add `destroy` option to `set<Multiple/Single>` rather than having a new API.

felixfbecker:
@sushantdhiman That's actually a great idea. That way you can use both of `setAssociation(s)` and `removeAssciation(s)` with destroy behavior 👍

afucher:
@felixfbecker @sushantdhiman What do you think about this:
I'll close this PR, and we discuss better in the issue's thread. When we decided what to do, I send another PR.

sushantdhiman:
@afucher Right, we should probably discuss more about `destroy` option. I am closing the PR and we can discuss API in related issue :)

kesemdavid:
@sushantdhiman is this feature currently available?

Issue #6959, opened 2016-12-07
$isIn should use $eq when only one element in the array


contra:
This should be a really easy fix, relevant code is here: https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L2020-L2055

Reduces the # of bytes in the query, makes them clearer to read, and also increases query performance (especially on postgres).

## What you are doing?

```js
SomeModel.find({
  where: {
    number: { $isIn: [ 1 ] }
  }
})

// or using the shorthand
SomeModel.find({
  where: {
    number: [ 1 ]
  }
})
```

## What do you expect to happen?

```sql
SELECT * FROM some_model WHERE "number" = 1
```

## What is actually happening?

```sql
SELECT * FROM some_model WHERE "number" IN (1)
```

__Dialect:__ any
__Sequelize version:__ 3.x / 4.x

aysark:
Submitting a PR for this one^

sushantdhiman:
How much performance gain you expect from single `IN` to `EQ` conversion ?

aysark:
@sushantdhiman good question.  So i did some testing, using EXPLAIN ANALYZE.. 

All in all, apart from the initial cache miss, it seems like there isn't any real performance gain for simple queries.  It may be different for more complex queries.

I found some more discussion on it here:
http://stackoverflow.com/questions/9926433/mysql-performance-in-clause-vs-equals-for-a-single-value

Here's my raw data:
```
sequelize_test=# EXPLAIN ANALYZE SELECT * FROM test WHERE test.number IN ('first');
                                           QUERY PLAN                                           
------------------------------------------------------------------------------------------------
 Seq Scan on test  (cost=0.00..20.75 rows=4 width=66) (actual time=0.142..0.143 rows=1 loops=1)
   Filter: ((number)::text = 'first'::text)
   Rows Removed by Filter: 2
 Planning time: 1.416 ms
 Execution time: 0.986 ms
(5 rows)

Time: 5.971 ms
sequelize_test=# EXPLAIN ANALYZE SELECT * FROM test WHERE test.number IN ('first');
                                           QUERY PLAN                                           
------------------------------------------------------------------------------------------------
 Seq Scan on test  (cost=0.00..20.75 rows=4 width=66) (actual time=0.011..0.012 rows=1 loops=1)
   Filter: ((number)::text = 'first'::text)
   Rows Removed by Filter: 2
 Planning time: 0.054 ms
[...]

PR #7003, opened 2016-12-18
IN operator should be = when only 1 element

aysark:
### Pull Request check-list

### Description of change

Fixes https://github.com/sequelize/sequelize/issues/6959

IN and NOT IN operators turn into = and != when dealing with 1 element arrays.

Doc not updated as this is an internal perf detail and i didn't think it'd need it.


sushantdhiman:
Can you show quantified performance gain before and after this conversion ?

sushantdhiman:
As discussed in github.com/sequelize/sequelize/issues/6959

PR #7005, opened 2016-12-19
Make Eager Loading Error Messages More User-Friendly

nadrane:
### Pull Request check-list

## Background
This PR aims to make error messages that appear when using the `include` keyword more user friendly. Currently, if a user specifies an association incorrectly within the `include` keyword, Sequelize throws an error stating: "Target is not associated with Source!", substituting model names where appropriate. This error message is very helpful in the instances when the target and source models are not associated; however, this is currently the only error message thrown when `include` statements are erroneously specified, even when the source and target models are associated!

## Description of Change
I've broken down incorrect `include` statement usage into 4 possible scenarios, each of which now has it's own custom error message.

# Scenario 1:

_Source and target models are not associated_

```
const User = db.define('user')
const Message = db.define('message')

db.sync({force: true})
    .then(function() {
        return Message.findAll({
            include: [User]
        });
    })
```

This scenario will preserve existing functionality, throwing an error message stating:

"User is not associated to Message!

# Scenario 2:

_Source and target models are associated **multiple** times, and an alias is not specified_

```
const User = db.define('user')
const Message = db.define('message')

Message.belongsTo(User, { as: 'toMessage' });
Message.belongsTo(User, { as: 'fromMessage' });

db.sync({force: true})
    .then(function() {
        return Message.findAll({
            include: [User]
        });
    })
```
This scenario will throw and error message stating:

"User is associated to Message multiple times. To identify the correct association, you must use the 'as' keyword to specify the alias of the association you want to include."

# Scenario 3:

_Source and target models are associated a **single** time, and an alias is not specified_

```
const User = db.define('user')
const Message = db.define('message')

Message.belongsTo(User, { as: 'message' });

db.sync({force: true})
    .then(function() {
        return Message.findAll({
            include: [User]
        });
    })
```

This scenario will throw and error message stating:

"User is associated to Message using an alias. You must use the 'as' keyword to specify the alias within your include statement."

# Scenario 4

_Source and target models are associated, but the included alias does not match the alias defined on the association._

```
const User = db.define('user')
const Message = db.define('message')

Message.belongsTo(User, { as: 'message' });

db.sync({force: true})
    .then(function() {
        return Message.findAll({
            include: [{model: User, as: 'toMessage'}]
        });
    })
```

This scenario will throw and error message stating:

"User is associated to Message using an alias. You've included an alias (toMessage), but it does not match the alias defined in your association."

## Next Steps
What do you all think of this PR? I am a teacher by day, and this PR was motivated by student confusion. My hope is that more precise and instructive error messages will help new Sequelize developers become acclimated to its API.

I want to include some more tests before merging, but I wanted to get the community's take on the PR first.

### DIFF
--- a/lib/associations/mixin.js
+++ b/lib/associations/mixin.js
@@ -210,18 +210,12 @@ const Mixin = {
   },
 
   getAssociations(target) {
-    return Object.values(this.associations).filter(association => association.target.name === target.name);
+    return _.values(this.associations).filter(association => association.target.name === target.name);
   },
 
   getAssociationForAlias(target, alias) {
     // Two associations cannot have the same alias, so we can use find instead of filter
-    const association = this.getAssociations(target)
-      .find(association => this.verifyAssociationAlias(association, alias));
-    if (!association) {
-      return null;
-    } else {
-      return association;
-    }
+    return this.getAssociations(target).find(association => this.verifyAssociationAlias(association, alias)) || null;
   },
 
   verifyAssociationAlias(association, alias) {
--- a/lib/associations/mixin.js
+++ b/lib/associations/mixin.js
@@ -209,18 +209,27 @@ const Mixin = {
     return association;
   },
 
-  getAssociation(target, alias) {
-    for (const associationName in this.associations) {
-      if (this.associations.hasOwnProperty(associationName)) {
-        const association = this.associations[associationName];
+  getAssociations(target) {
+    return Object.values(this.associations).filter(association => association.target.name === target.name);
+  },
 
-        if (association.target.name === target.name && (alias === undefined ? !association.isAliased : association.as === alias)) {
-          return association;
-        }
-      }
+  getAssociationForAlias(target, alias) {
+    // Two associations cannot have the same alias, so we can use find instead of filter
+    const association = this.getAssociations(target)
+      .find(association => this.verifyAssociationAlias(association, alias));
+    if (!association) {
+      return null;
+    } else {
+      return association;
     }
+  },
 
-    return null;
+  verifyAssociationAlias(association, alias) {
+    if (alias) {
+      return association.as === alias;
+    } else {
+      return !association.isAliased;
+    }
   }
 };
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -556,39 +556,26 @@ class Model {
     const associations = this.getAssociations(targetModel);
     let association = null;
     if (associations.length === 0) {
-      noAssociationError(this.name, targetModel.name);
+      throw new sequelizeErrors.EagerLoadingError(`${targetModel.name} is not associated to ${this.name}!`);
     } else if (associations.length === 1) {
       association = this.getAssociationForAlias(targetModel, targetAlias);
       if (!association) {
         if (targetAlias) {
-          aliasMismatchError(this.name, targetModel.name, targetAlias);
+          throw new sequelizeErrors.EagerLoadingError(`${targetModel.name} is associated to ${this.name} using an alias. ` + 
+            `You've included an alias (${targetAlias}), but it does not match the alias defined in your association.`);
         } else {
-          associationButMissingAliasError(this.name, targetModel.name);
+          throw new sequelizeErrors.EagerLoadingError(`${targetModel.name} is associated to ${this.name} using an alias. ` +
+            `You must use the 'as' keyword to specify the alias within your include statement.`);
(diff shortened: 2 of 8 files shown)


lib/associations/mixin.js
```diff
@@ -209,18 +209,27 @@ const Mixin = {
     return association;
   },
 
-  getAssociation(target, alias) {
```
felixfbecker: Is this a breaking change?

lib/associations/mixin.js
```diff
@@ -209,18 +209,27 @@ const Mixin = {
...
-    for (const associationName in this.associations) {
-      if (this.associations.hasOwnProperty(associationName)) {
-        const association = this.associations[associationName];
+  getAssociations(target) {
+    return Object.values(this.associations).filter(association => association.target.name === target.name);
```
felixfbecker: If `target` is the model class, can't we just compare by identity `===`?

lib/associations/mixin.js
```diff
@@ -209,18 +209,27 @@ const Mixin = {
...
+  },
 
-        if (association.target.name === target.name && (alias === undefined ? !association.isAliased : association.as === alias)) {
-          return association;
-        }
-      }
+  getAssociationForAlias(target, alias) {
```
felixfbecker: Could you comment a bit why you added these new methods and what they do?

lib/associations/mixin.js
```diff
@@ -209,18 +209,27 @@ const Mixin = {
...
+    // Two associations cannot have the same alias, so we can use find instead of filter
+    const association = this.getAssociations(target)
+      .find(association => this.verifyAssociationAlias(association, alias));
+    if (!association) {
+      return null;
```
felixfbecker: This is redundant, `find()` already returns `undefined` when the item was not found

lib/model.js
```diff
@@ -564,6 +552,46 @@ class Model {
...
+    } else {
+      association = this.getAssociationForAlias(targetModel, targetAlias);
+      if (!association){
+        multipleAssociationError(this.name, targetModel.name);
+      }    
+    }
+    return association;
+    
+    function noAssociationError(source, target) { 
+      throw new Error(`${target} is not associated to ${source}!`);
```
felixfbecker: It would make a more sense to define these as new `Error` subclasses

lib/associations/mixin.js
nadrane: I added `getAssociations` because we need to know how many associations two models have in order to throw the most informative errors possible.

'getAssociationForAlias' is functionally identical to what used to be `getAssociation`. I renamed it to avoid confusion between it `getAssociations`, seeing as they are functionally get quite different now. I wanted the name to convey that it checks the association's alias as well.

`verifyAssociationAlias` is just a helper method

lib/associations/mixin.js
nadrane: I assumed that `getAssociation` is a private API. I effectively replaced it and all of it's other two invocations with `getAssociationForAlias`

lib/model.js
nadrane: will fix

nadrane:
It looks like code ship failed because of Object.values. Is it best practice to avoid this function given how new it is?

lib/associations/mixin.js
felixfbecker: You could just do
```js
return this.getAssociations(target).find(association => this.verifyAssociationAlias(association, alias)) || null;
```
to keep the previous behavior

felixfbecker:
We need to support Node 4, I think that function is not available there. But lodash' `_.values()` does the same

lib/model.js
nadrane: I added `sequelizeErrors.EagerLoadingError`

lib/associations/mixin.js
janmeier: This is fine, it's now documented as part of the public API

janmeier: *its _NOT_

nadrane:
I'm teaching at [Fullstack Academy](https://www.fullstackacademy.com/) in Chicago. 

Thank you for accepting the PR! This is my first experience contributing to a major open source project, and it's been an absolute pleasure. The approval process was fast and painless, and I feel like I learned a lot.

I would love to contribute similar changes in the future to improve the usability of the library. Are there any other questions that you hear frequently on slack/github?

Issue #7010, opened 2016-12-20
Sequelize allows two associations with the same alias


nadrane:
## What you are doing?
```
const User = db.define('user')
const Message = db.define('message')

Message.belongsTo(User, { as: 'to' });
Message.belongsTo(User, { as: 'to' });
```

## What do you expect to happen?
Sequelize should throw an error. Inside `models.js`,  `_validateIncludedElements`does not support two associations with the same alias. This could produce unexpected behavior. We should warn the user preemptively.

## What is actually happening?
Sequelize silently accepts the second association.

__Sequelize version:__ master

nadrane:
this can be closed now

PR #7025, opened 2016-12-23
Prevent Duplicate Aliases in Associations

nadrane:
### Pull Request check-list

### Description of change

Closes issue #7010 

This code updates the constructor of each association to check if a given alias has been used before creating the association. An error is thrown if a particular alias already exists.


nadrane:
I definitely messed up my commit history... I hope that's not a huge deal. Otherwise I can remake the PR

lib/associations/has-one.js
```diff
@@ -18,7 +18,7 @@ const Association = require('./base');
  */
 class HasOne extends Association {
   constructor(srcModel, targetModel, options) {
```
sushantdhiman: Should name it source, target like all other associations, Please change the names

lib/model.js
```diff
@@ -2788,6 +2788,10 @@ class Model {
     return this.name;
   }
 
+  static hasAlias(alias) {
+      return _.values(this.associations).some(association => association.as === alias);
```
sushantdhiman: 2 space indentation only :)

test/unit/model/helpers.test.js
```diff
@@ -0,0 +1,24 @@
...
+const expect = chai.expect;
+const Support = require(__dirname + '/../support');
+const current = Support.sequelize;
+
+describe(Support.getTestDialectTeaser('Model'), function() {
+  describe('hasAlias', function () {
+    beforeEach(function() {
+      this.User = current.define('user');
+      this.Task = current.define('task'); 
+    });
```
sushantdhiman: Please add one extra line after `beforeEach`

test/unit/model/helpers.test.js
```diff
@@ -0,0 +1,24 @@
...
+    it('returns true if a model has an association with the specified alias', function() {
+        this.Task.belongsTo(this.User, { as: 'owner'});
+        expect(this.Task.hasAlias('owner')).to.equal(true);
+    });
```
sushantdhiman: One line here as well :)

test/unit/associations/association.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+/* jshint -W030 */
+const chai = require('chai');
+const expect = chai.expect;
+const Support = require(__dirname + '/../support');
+const current = Support.sequelize;
+const AssociationError = require(__dirname + '/../../../lib/errors').AssociationError;
+
+describe(Support.getTestDialectTeaser('belongsTo'), function() {
+  it('should throw an AssociationError when two associations have the same alias', function(){
```
sushantdhiman: Space after this `function` parentheses

test/unit/associations/association.test.js
```diff
@@ -0,0 +1,20 @@
...
+    const User = current.define('User');
+    const Task = current.define('Task');
+
+    User.belongsTo(Task, { as: 'task' });
+    const errorFunction = User.belongsTo.bind(User, Task, { as: 'task' });
+    const errorMessage = `You have used the alias task in two separate associations. Aliased associations must have unique aliases.`
```
sushantdhiman: Semi colon missing

sushantdhiman:
Travis CI just passed this PR with jshint error, need to fix that soon :)

lib/associations/base.js
```diff
@@ -1,6 +1,13 @@
 'use strict';
+const AssociationError = require('./../errors').AssociationError;
 
-var Association = function() {};
+var Association = function(source, options) {
```
felixfbecker: while you're at it, can you make this a class?

lib/associations/base.js
```diff
@@ -1,6 +1,13 @@
...
+  if (source.hasAlias(options.as)) {
+    throw new AssociationError(`You have used the alias ${options.as} in two separate associations. ` +
+    `Aliased associations must have unique aliases.`
```
felixfbecker: single quotes

lib/associations/belongs-to.js
```diff
@@ -133,7 +133,7 @@ class BelongsTo extends Association {
     this.identifierField = this.source.rawAttributes[this.foreignKey].field || this.foreignKey;
 
     this.source.refreshAttributes();
-
+
```
felixfbecker: trailing whitespace

lib/associations/base.js
felixfbecker: If we introduce a constructor here, it should take `source, target, options` like the others, and assign those so the subclasses don't have to

lib/associations/base.js
nadrane: Can we not use template strings?

test/unit/associations/association.test.js
nadrane: Is jshint supposed to catch all of these different types of errors? 
I would think it would at least catch the 2 spaces/space after function parens
Though I'm honestly more familiar with eslint

nadrane:
I'm going to add in some extra regression tests because of the changes to the association constructor

lib/associations/base.js
felixfbecker: You can use template strings when there are expressions inside it. The last string is just a simple string

test/unit/associations/association.test.js
sushantdhiman: We have rudimentary support for ESlint based rules, but its not adopted fully in our codebase. Progress on complying full codebase to ESlint rules is bit slow. In future we want to use ESlint to catch all these errors automatically :)

test/unit/associations/belongs-to-many.test.js
```diff
@@ -11,7 +11,7 @@ const BelongsTo = require(__dirname + '/../../../lib/associations/belongs-to');
 const HasMany = require(__dirname + '/../../../lib/associations/has-many');
 const HasOne = require(__dirname + '/../../../lib/associations/has-one');
 const current = Support.sequelize;
-const Promise = current.Promise;
+var Promise = current.Promise;
```
felixfbecker: ?

nadrane:
I think all of the review comments are taken care of :)

test/unit/associations/belongs-to-many.test.js
nadrane: It was originally `var`.

jshint fails with `Redefinition of 'Promise` if we use const.

nadrane: I honestly don't entirely understand why this qualifies as a redefinition... I suppose Promise exists as a builtin, but that builtin was not defined with `const`

nadrane: hah, but the test suite actually runs without error despite the jshint error. Maybe this is an issue with jshint?

nadrane: here we are: https://github.com/jshint/jshint/issues/1747#issuecomment-52574352

I added a jshint exception to the file

Issue #6947, opened 2016-12-04
benchmark not working with custom logger


OlivierCuyp:
Hi,

For now, if I add a custom logger while activating benchmark I don't get the benchmark time.

```js
function myLoggingFunction (message) {
  // Log in an instance of bunyan
  log.info(message);
}

{
  logging: myLoggingFunction,
  benchmark: true
}
```

I don't really understand why the benchmark time is not included in the message when not using `console.log`.

Apparently the issue comes from the following code :

 https://github.com/sequelize/sequelize/blob/afc25e3db8fa0a9a8bed3aa14869cafa1194f9fd/lib/sequelize.js#L1076

Should the test rather be :

```js
if ((this.options.benchmark || options.benchmark) && (typeof options.logging === 'function')) {
  ...
}
```

Still I can make it work if I define my myLogingFunction like this :

```js
function myLoggingFunction (message, benchmark) {
  if (!benchmark) {
    return log.info(message);
  }
  log.info(`${message} Elapsed time: ${benchmark} ms`);
};
```
I also don't really understand why the benchmark is pass as a second argument ...

felixfbecker:
Feel free to fix

OlivierCuyp:
Just want to understand first what's the intention when passing the benchmark as a second argument.

felixfbecker:
Probably so your logger can handle benchmark results in a custom way, differently than SQL logs?

janmeier:
This is intentional - Suppose for example that you want to trace slow queries, and only log those that take more than xx milliseconds. If the timing was embedded into a string, you would have to split and parse that string to get the number

PR #7029, opened 2016-12-23
Fix benchmark logging for custom logging function

OlivierCuyp:
Quick fix for logging benchmark when using a custom logging function in the options.

Related issue is : https://github.com/sequelize/sequelize/issues/6947

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -500,7 +500,7 @@ class Sequelize {
       if (options.replacements && options.bind) {
         throw new Error('Both `replacements` and `bind` cannot be set at the same time');
       }
-      
+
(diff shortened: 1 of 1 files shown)


abdulmoizeng:
@tornillo @janmeier 
For the commit https://github.com/sequelize/sequelize/commit/f7fa33ce0522efcbd6a730ca54565aa321ef8770#diff-24f8b5cb05b7e92726ca732f7d16d5baR1335
Can any one please explain why sql-timings is passed as second argument when benchmarking option is enabled?

janmeier:
@abdulmoizeng See my comment in the linked issue :)

PR #7040, opened 2016-12-26
Clone defaultScope so it doesn't get overwritten in Model#init v3

broofa:
As the title says. #6971 for details


broofa:
cc @felixfbecker

Issue #7030, opened 2016-12-23
Incorrect warning "Model attributes (order) found in finder"


OlivierCuyp:
The findAll method check the options attributes to see if the match any of the model attributes.
If so, it generates a warning message.
This isn't a correct behavior, it can check options attributes if they match expected ones but it shouldn't make intention guess.

See the code:

https://github.com/sequelize/sequelize/blob/7d597efbaaa2a3927b10e128e41ed7e44e1c8bb5/lib/model.js#L1451

Here is how this becomes an issue.
I have the following model :

```js
const schema = {
  order: {
    type: DataTypes.INTEGER,
    defaultValue: 0
  },
  label: {
    type: DataTypes.STRING(255)
  },
  text_fr: {
    type: DataTypes.STRING(255)
  },
  text_nl: {
    type: DataTypes.STRING(255)
  }
};

const Feedback = sequelize.define('feedback', schema, {
  underscored: true
});
```

So when I do :

```js
feedback.findAll({
  order: [ ['order', 'ASC'], ['id', 'ASC'] ]
});
```

Which produce this warning :

```
(sequelize) Warning: Model attributes (order) found in finder method options but options.where object is empty. Did you forget to use options.where?
```

I have the warning just because I do an `order` and I have a field name `order`.

So I would completely remove this check or validate that options attributes are the legal ones.

nadrane:
this is intentional behavior. Check out [#2998](https://github.com/sequelize/sequelize/issues/2998)

Generally when the options passed to findAll contain a key of the model, the user intends to filter on it. Sometimes we forget to use the `where` clause despite our best intentions. The warning you see is asking you if you are intending to filter on `order` but are accidentally using the wrong syntax.

felixfbecker:
Yeah but it should not do this if the property is a valid option like `order`.

sushantdhiman:
Fixed in https://github.com/sequelize/sequelize/pull/7047

jayrojb:
@nadrane effectively sintax error.. thanks .. my case was the option where within a finder.. thanks

PR #7047, opened 2016-12-28
Improved findall warnings

nadrane:
### Pull Request check-list

### Description of change
This PR resolves issue #7030. In addition, it adds an additional warning when invalid options are provided to findAll (and all of its callers). If a user passes an invalid option to this function, chances are his intention was to do something else. The goal of this PR is to help developers identify and correct their mistakes early.


nadrane:
hmm I just realized this is going to have a lot of misfires for functions that call findAll that have a different set of options

PR #7066, opened 2017-01-04
Consecutive function scopes

### LINKED ISSUE #7058: scope method syntax loses parameters when used multiple times
## What you are doing?
```js

// simple model with scopes that take parameters
var Location = sequelize.define('Location', {
  state: types.STRING,
  city: types.STRING
}, {
  scopes: {
    state: function(state) {
      return {where: {state: state}};
    },
    city: function(city) {
      return {where: {city: city}};
    }
  }
});

// create a scopes array to be reused
var scopes = [
  {method: ['city', 'Springfield']},
  {method: ['state', 'OH']}
];

// do some queries
Location.scope(scopes).count().then(console.log);
Location.scope(scopes).findAll().then(console.log);
```

I expect the arguments in the scopes to be passed into the queries for both queries.

Only the first query is using the arguments because the internal `method` arrays on the `scopes` array are being modified [here](https://github.com/sequelize/sequelize/blob/master/lib/model.js#L1298) when using `.splice`.

Locally, I changed that to `.slice` so the array isn't mutated and it fixed my use case. I did a little digging in the issues, but not enough to determine if this has come up before. If this is a bug, I can try to submit a PR when I get a chance to make the changes and provide unit tests.

doowb:
### Pull Request check-list

  `npm run test-sqlite` and `npm run test-postgres` pass. I don't have the other dialects setup and configured locally.
  Closes #7058 
  Add test that fails before the change.

### Description of change

This PR changes [.splice](https://github.com/sequelize/sequelize/blob/master/lib/model.js#L1306) to `.slice` to ensure that the original methods array is not modified so the scope objects may be reused in consecutive calls to `.scope`.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -34,6 +34,7 @@
 - [FIXED] Custom error messages used for incorrect eager loading [#7005](https://github.com/sequelize/sequelize/pull/7005)
 - [FIXED] Enforce unique association aliases [#7025](https://github.com/sequelize/sequelize/pull/7025)
 - [FIXED] Information warnings when findAll is given incorrect inputs [#7047](https://github.com/sequelize/sequelize/pull/7047)
+- [FIXED] scope method syntax loses parameters when used multiple times [#7058](https://github.com/sequelize/sequelize/issues/7058)
 
 ## BC breaks:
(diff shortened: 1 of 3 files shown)


janmeier:
:+1:

PR #7081, opened 2017-01-07
MSSQL removeColumn method fix for primaryKey column

harshithkashyap:
### Pull Request check-list

### Description of change

When the `removeColumn` method is called on a primaryKey column, SQL Server errors as the primaryKey column has an associated constraint which should be dropped before dropping the primaryKey column.

- Added `getPrimaryKeyConstraint` method to retrieve primary key constraint based on tableName and columnName.
- Extends #4943

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -680,6 +680,25 @@ var QueryGenerator = {
     return sql;
   },
 
+  getPrimaryKeyConstraint(table, attributeName) {
+    const tableName = wrapSingleQuote(table.tableName || table);
+    const sql = [
+      'SELECT K.TABLE_NAME AS tableName,',
+      'K.COLUMN_NAME AS columnName,',
+      'K.CONSTRAINT_NAME AS constraintName',
+      'FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS AS C',
+      'JOIN INFORMATION_SCHEMA.KEY_COLUMN_USAGE AS K',
+      'ON C.TABLE_NAME = K.TABLE_NAME',
+      'AND C.CONSTRAINT_CATALOG = K.CONSTRAINT_CATALOG',
+      'AND C.CONSTRAINT_SCHEMA = K.CONSTRAINT_SCHEMA',
+      'AND C.CONSTRAINT_NAME = K.CONSTRAINT_NAME',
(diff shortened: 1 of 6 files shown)


harshithkashyap:
@janmeier @sushantdhiman @felixfbecker Request review

lib/dialects/mssql/query-generator.js
```diff
@@ -680,6 +680,25 @@ var QueryGenerator = {
     return sql;
   },
 
+  getPrimaryKeyConstraint(table, attributeName) {
```
sushantdhiman: This method only generates SQL, lets call it `getPrimaryKeyConstraintQuery`

lib/dialects/mssql/query-generator.js
```diff
@@ -680,6 +680,25 @@ var QueryGenerator = {
...
+    const tableName = wrapSingleQuote(table.tableName || table);
+    const sql = [
```
sushantdhiman: You can directly return query without using `sql` variable :)

test/integration/query-interface.test.js
```diff
@@ -638,6 +638,20 @@ describe(Support.getTestDialectTeaser('QueryInterface'), function() {
             expect(table).to.not.have.property('manager');
         });
       });
+
+      it('Should be able to remove a column with primaryKey', function () {
```
sushantdhiman: `Should` => `should`

test/integration/query-interface.test.js
```diff
@@ -692,6 +706,20 @@ describe(Support.getTestDialectTeaser('QueryInterface'), function() {
             expect(table).to.not.have.property('lastName');
           });
       });
+
+      it('Should be able to remove a column with primaryKey', function () {
```
sushantdhiman: `Should` => `should`

lib/dialects/mssql/query-generator.js
harshithkashyap: @sushantdhiman I wanted to keep the query readable yet copy pastable when logged so went for this. Using backticks keeps the same formatting as defined when logged.

sushantdhiman: I have no issue with Array based query generation and joining it, I would mostly encourage it for query generation.

My point was you put join result in a variable and return that variable, why don't directly return output of join directly

harshithkashyap: Ok sure, I'll make the mentioned changes :)

PR #7094, opened 2017-01-10
SQLite JSON datatype support (with JSON1 extension)

rokoroku:
### Pull Request check-list

- [x] Does your issue contain a link to existing issue (part of #4873) or a description of the issue you are solving?

### Description of change

Continued from #6500 

- Add supports for [SQLite](https://sqlite.org/)'s built-in JSON datatype with [JSON1 extension](https://sqlite.org/json1.html).
  
  (JSON1 extension is enabled with ``sqlite3`` package by default. (see [node-sqlite3#538](https://github.com/mapbox/node-sqlite3/pull/538)))

- Separate postgres' JSON tests into JSON and JSONB test
  
  Since most of the existing JSON implementations (and test codes) were based on PostgreSQL, so I tried to seperate them. I think it will be better for future updates for JSON with other dialects. (for example, operators like ``->>``, ``#>>`` is only supported by postgres. It cannot be shared operator for JSON data type, IMHO)

  There seems like also inaccurate tests. What I fixed for postgres are:   
    1. $contains on JSON field only supported via JSONB type, but described in JSON spec
    2. create index on JSON field only support via JSONB type, but described in JSON spec   

* Move each dialect's JSON path extraction query into `whereItemQuery(...)` in `AbstractQueryGenerator` and add new private API.

  from this PR, dialects that supports **JSON path extraction query** should override below method

```javascript
  /**
   * Generates an SQL query that extract JSON property of given path.
   *
   * @param   {String}               column    The JSON column 
   * @param   {String|Array<String>} [path]    The path to extract (optional)
   * @returns {String}                         The generated sql query
   * @private
   */
  jsonPathExtractionQuery(column, path): String
``` 

* Move common JSON dao test suites to `abstract/dao.test.js`

* Add some injection preventions of JSON functions to each dialect's `dao.js` & `dao.test.js`.

* Change postgres' CAST syntax for JSON column (before `::`, after `CAST(... AS ...)` )

* Update related files to use ES6 syntax.

Please let me know if there is something wrong or something to fix.


lib/dialects/sqlite/query-generator.js
```diff
@@ -79,6 +81,153 @@ const QueryGenerator = {
...
+          return;
+        }
+
+        traverse(prop, item, [prop]);
+      });
+
+      const result = items.join(' AND ');
+      return items.length > 1 ? '('+result+')' : result;
+    } else {
+      return MySqlQueryGenerator.whereItemQuery.call(this, key, value, options);
```
sushantdhiman: AbstractQueryGen

lib/dialects/sqlite/query-generator.js
```diff
@@ -79,6 +81,153 @@ const QueryGenerator = {
...
+        }
+
+        if (smth.value) {
+          str += util.format(' = %s', this.escape(smth.value));
+        }
+
+        return str;
+      }
+    } else {
+      return MySqlQueryGenerator.handleSequelizeMethod.call(this, smth, tableName, factory, options, prepend);
```
sushantdhiman: AbsctractQueryGen

lib/dialects/sqlite/query-generator.js
```diff
@@ -79,6 +81,153 @@ const QueryGenerator = {
     return !!value ? 1 : 0;
   },
 
+  // A recursive parser for nested where conditions
+  // (copied from ../postgres/query-generator.js)
+  parseConditionObject(_conditions, path) {
```
sushantdhiman: If this same method being used in both dialects, we should move it to AbstractQueryGen

test/integration/dialects/sqlite/dao.test.js
```diff
@@ -3,14 +3,17 @@
 var chai = require('chai')
   , expect = chai.expect
   , Support = require(__dirname + '/../../support')
+  , dialect = Support.getTestDialect()
   , DataTypes = require(__dirname + '/../../../../lib/data-types')
-  , dialect = Support.getTestDialect();
+  , sequelize = require(__dirname + '/../../../../lib/sequelize');
```
sushantdhiman: You dont need to include it, access it as `this.sequelize`

test/integration/dialects/sqlite/dao.test.js
```diff
@@ -72,6 +75,223 @@ if (dialect === 'sqlite') {
...
+      });
+
+      it('should be able to retrieve element of array by index', function() {
+        var self = this;
+        var emergencyContact = { name: 'kate', phones: [1337, 42] };
+
+        return this.User.create({ username: 'swen', emergency_contact: emergencyContact })
+          .then(function(user) {
+            expect(user.emergency_contact).to.eql(emergencyContact);
+            return self.User.find({ where: { username: 'swen' }, attributes: [[sequelize.json('emergency_contact.phones[1]'), 'firstEmergencyNumber']] });
```
sushantdhiman: `this.sequelize`

lib/dialects/sqlite/query-generator.js
rokoroku: Sure, I'll try to move it to AbstractQueryGen.

lib/dialects/sqlite/query-generator.js
rokoroku: Well, this method is bit different from postgres one. 
in postgres, ``#>>`` syntax is used
```javascript
`${this.quoteIdentifier(_.first(path))}#>>'{${_.tail(path).join(',')}}'`
```

however, in sqlite, ``json_extract`` syntax is used
```javascript
`json_extract(${this.quoteIdentifier(_.first(path))}, '$.${_.tail(path).join('.')}')`
```

lib/dialects/sqlite/query-generator.js
```diff
@@ -79,6 +81,153 @@ const QueryGenerator = {
...
+
+        return str;
+      }
+    } else {
+      return MySqlQueryGenerator.handleSequelizeMethod.call(this, smth, tableName, factory, options, prepend);
+    }
+  },
+
+  whereItemQuery(key, value, options) {
+
```
rokoroku: I'll fix it.

lib/dialects/sqlite/query-generator.js
```diff
@@ -79,6 +81,153 @@ const QueryGenerator = {
...
+      });
+
+      const result = items.join(' AND ');
+      return items.length > 1 ? '('+result+')' : result;
+    } else {
+      return MySqlQueryGenerator.whereItemQuery.call(this, key, value, options);
+    }
+  },
+
   addColumnQuery(table, key, dataType) {
```
rokoroku: @sushantdhiman May I fix extra line here, too?

test/integration/dialects/sqlite/dao.test.js
rokoroku: OK

lib/dialects/sqlite/query-generator.js
rokoroku: I thought sqlite query-generator's ``__proto__`` is set as ``MySqlQueryGenerator``,  so I use it in inheritance manner.
I'll change it anyway.

test/integration/dialects/sqlite/dao.test.js
rokoroku: OK

lib/dialects/sqlite/query-generator.js
sushantdhiman: I dont want you to move it, just use replace `MySqlQueryGenerator => AbstractQueryGenerator`

lib/dialects/sqlite/query-generator.js
sushantdhiman: Correct but this should be fixed in future. Either method is shared using Abstract or defined in current instance. Cross dialect references couples dialect code, thus less freedom to change queries because of these references

lib/dialects/sqlite/query-generator.js
sushantdhiman: It should be possible to use Abstract Generator as MySQL doesn't define this method either ( it also depends upon Abstract)

lib/dialects/sqlite/data-types.js
```diff
@@ -23,6 +23,17 @@ module.exports = BaseTypes => {
   BaseTypes.REAL.types.sqlite = ['REAL'];
   BaseTypes.DOUBLE.types.sqlite = ['DOUBLE PRECISION'];
   BaseTypes.GEOMETRY.types.sqlite = false;
+  BaseTypes.JSON.types.sqlite = ['JSON', 'JSONB'];
+
+  function JSONTYPE() {
```
sushantdhiman: @rokoroku Also we will end up polluting internal JSON, leave it as JSONTYPE

test/integration/dialects/sqlite/dao.test.js
```diff
@@ -3,6 +3,7 @@
 var chai = require('chai')
   , expect = chai.expect
   , Support = require(__dirname + '/../../support')
+  , Sequelize = Support.Sequelize
```
rokoroku: ``json(...)`` is static method inside ``Sequelize`` class, but ``this.sequelize`` was an instance of Sequelize. (it causes test failure.)

So, I added ``Sequelize`` from ``Support.Sequelize``

lib/dialects/sqlite/data-types.js
rokoroku: @sushantdhiman  Oh, I understood you want me to erase JSONB from line 26.

So, would I just leave these code as it is?

sushantdhiman: We need `JSON.parse` so we can't rename function itself, but its okay I think. So no changes

sushantdhiman:
@rokoroku Please document that JSON supported fo SQLite as well https://github.com/sequelize/sequelize/blob/a1bf8b02a0b6f136dd2a674b8a72bff38aa8eba0/lib/data-types.js#L505

sushantdhiman:
LGTM, Its a big change, would also like reviews from other members of @sequelize/core

lib/dialects/abstract/query-generator.js
```diff
@@ -2228,6 +2228,19 @@ const QueryGenerator = {
     return result ? result : '1=1';
   },
 
+  // A recursive parser for nested where conditions
+  parseConditionObject(conditions, path) {
+    path = path || [];
+    return _.reduce(conditions, (r, v, k) => { // result, key, value
```
felixfbecker: why not name the params `result`, `key`, `value`? More common though is `previous`, `current`, `index`
(even though I realize this is copy+pasted)

lib/dialects/sqlite/query-generator.js
```diff
@@ -79,8 +82,139 @@ const QueryGenerator = {
     return !!value ? 1 : 0;
   },
 
-  addColumnQuery(table, key, dataType) {
+  handleSequelizeMethod(smth, tableName, factory, options, prepend) {
+    if (smth instanceof Utils.Json) {
+      // Parse nested object
+      if (smth.conditions) {
+        const conditions = _.map(this.parseConditionObject(smth.conditions), condition =>
```
felixfbecker: If `parseConditionObject` returns an array, you can use `array.map()` here

lib/dialects/sqlite/query-generator.js
```diff
@@ -79,8 +82,139 @@ const QueryGenerator = {
...
+          `json_extract(${this.quoteIdentifier(_.first(condition.path))}, '$.${_.tail(condition.path).join('.')}') = '${condition.value}'`
+        );
+
+        return conditions.join(' and ');
+      } else if (smth.path) {
+        let str;
+
+        // Allow specifying conditions using the sqlite json1 functions
+        // https://sqlite.org/json1.html
+        if (/(json(?:_.*)*)\(.*\)/i.test(smth.path)) {
```
felixfbecker: can you have nested parenthesis in the input, e.g using the result of a function as a parameter to the JSON function?
In that case a regexp will not suffice.
Shouldn't you pass a function call as `sequelize.fn()`?
Can this cause SQL injection if `where` comes from user input?

lib/dialects/sqlite/query-generator.js
```diff
@@ -79,8 +82,139 @@ const QueryGenerator = {
...
+        }
+
+        if (smth.value) {
+          str += util.format(' = %s', this.escape(smth.value));
+        }
+
+        return str;
+      }
+    } else {
+      return AbstractQueryGenerator.handleSequelizeMethod.call(this, smth, tableName, factory, options, prepend);
```
felixfbecker: `super.handleSequelizeMethod()`

lib/dialects/sqlite/query-generator.js
```diff
@@ -79,8 +82,139 @@ const QueryGenerator = {
...
+          return;
+        }
+
+        traverse(prop, item, [prop]);
+      });
+
+      const result = items.join(' AND ');
+      return items.length > 1 ? '('+result+')' : result;
+    } else {
+      return AbstractQueryGenerator.whereItemQuery.call(this, key, value, options);
```
felixfbecker: `super.whereItemQuery()`

test/integration/dialects/sqlite/dao.test.js
```diff
@@ -72,6 +75,223 @@ if (dialect === 'sqlite') {
       });
     });
 
+    describe('json', function() {
+      it('should tell me that a column is json', function() {
+        return this.sequelize.queryInterface.describeTable('Users')
+          .then(function(table) {
```
felixfbecker: use arrow functions.

test/integration/dialects/sqlite/dao.test.js
```diff
@@ -72,6 +75,223 @@ if (dialect === 'sqlite') {
...
+      });
+
+      it('should stringify json with insert', function() {
+        return this.User.create({
+          username: 'bob',
+          emergency_contact: { name: 'joe', phones: [1337, 42] }
+        }, {
+          fields: ['id', 'username', 'document', 'emergency_contact'],
+          logging: function(sql) {
+            var expected = '\'{"name":"joe","phones":[1337,42]}\'';
```
felixfbecker: use `const`.

test/integration/dialects/sqlite/dao.test.js
```diff
@@ -72,6 +75,223 @@ if (dialect === 'sqlite') {
...
+          return self.UserFields.create({
+            emergencyContact: { name: 'joe', phones: [1337, 42] }
+          }).then(function(user) {
+            expect(user.emergencyContact.name).to.equal('joe');
+          });
+        });
+      });
+
+      it('should update json using a custom field name', function() {
+        var self = this;
```
felixfbecker: use arrow functions.

lib/dialects/abstract/query-generator.js
rokoroku: Yes, I just copy-and-pasted existing things without modification. 
I'll change similar things anyway :)

lib/dialects/sqlite/query-generator.js
rokoroku: That's right. (It's also copied code from postgres)

lib/dialects/sqlite/query-generator.js
rokoroku: It is derived code from [lib/dialects/postgres/query-generator.js#L150](https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/query-generator.js#L150)

And it tests this spec:
```javascript
it('should be able to retrieve a row based on the values of the json document', function() {
  var self = this;

  return this.sequelize.Promise.all([
    this.User.create({ username: 'swen', emergency_contact: { name: 'kate' } }),
    this.User.create({ username: 'anna', emergency_contact: { name: 'joe' } })])
    .then(function() {
      return self.User.find({ where: Sequelize.json(`json_extract(emergency_contact, '$.name')`, 'kate'), attributes: ['username', 'emergency_contact'] });
    })
    .then(function(user) {
      expect(user.emergency_contact.name).to.equal('kate');
    });
});
```
which derived from  [test/integration/dialects/postgres/dao.test.js](https://github.com/sequelize/sequelize/blob/master/test/integration/dialects/postgres/dao.test.js#L221)

I do not understand sequelize as deeply as you, but the original code also seems to check just the query contains raw syntax or not. (means it allows user to use **raw-string query**, or query from ``Literal('...')`` maybe?)  

If this is the case, and If user want to use raw queries, SQL injection should be protected above this **raw-query layer** (before create the query) I think.

If you have any suggestions, please let me know.

test/integration/dialects/sqlite/dao.test.js
rokoroku: I agree

test/integration/dialects/sqlite/dao.test.js
rokoroku: For this, ``this`` reference is required so it cannot be replaced with arrow function..

lib/dialects/sqlite/query-generator.js
sushantdhiman: Calling `super` here will trigger `MySQLQueryGen` (as SQLITE inherits it), but `super.whereItemQuery` is not present in MySQLQueryGen (Although SQLite -> MySQL -> Abstract is true) but we should refer to AbstractQueryGen as we want to switch from MySQLQueryGen in future.

rokoroku: According to [MySQL JSON functions](https://dev.mysql.com/doc/refman/5.7/en/json-function-reference.html), they allows `->`, `->>` syntax to query JSON datatypes while SQLite does not.
And the following PR #7106 also implements ``whereItemQuery`` which returns `->>` syntax

There should be more differences in the future, so I agree that it should be switched from MySQLQueryGen someday.

test/integration/dialects/sqlite/dao.test.js
rokoroku: OK, since I cannot change all the es5-styled files, I will convert at least this file to es6.

lib/dialects/sqlite/query-generator.js
sushantdhiman: @rokoroku Please keep it `AbstractQueryGen`, I'll try to move SQLite away from MySQL dependency next weekend

test/integration/dialects/sqlite/dao.test.js
```diff
@@ -32,16 +32,16 @@ if (dialect === 'sqlite') {
...
         user.dataValues.createdAt = new Date(2011, 4, 4);
 
-        return user.save().then(function() {
-          return self.User.create({ username: 'new user' }).then(function() {
+        return user.save().then(() => {
+          return self.User.create({ username: 'new user' }).then(() => {
             return self.User.findAll({
               where: { createdAt: { $gt: new Date(2012, 1, 1) }}
-            }).then(function(users) {
+            }).then((users) => {
```
felixfbecker: please omit parenthesis if there is only one parameter

rokoroku: OK, it's not a problem

lib/dialects/sqlite/query-generator.js
sushantdhiman: @rokoroku Could you please add a case for SQL Injection attack (please check [these tests](https://github.com/sequelize/sequelize/blob/5b653701849325b298df7b27e7fdf1f8c273f1ec/test/integration/model/geometry.test.js#L181) for inspiration)

rokoroku: @sushantdhiman That seems to be a test limited to the GEOMETRY type.

Unfortunately, I really do not have any knowledge of what kind of injection is possible for JSON methods. Please let me know if there is any easy reference.

rokoroku:
While improve other tests to cover this PR, I found an issue that deleteQuery doesn't expand JSON methods in sqlite. So I'm now working on this.

rokoroku:
Add some prevention codes for possible injection I can think of.
Newly added method might looks bulky, but it effectively checks json_method brackets with quoted strings.

test/integration/data-types.test.js
```diff
@@ -269,82 +266,90 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
...
+              // MySQL will return NULL, becuase they lack EMPTY geometry data support.
+              expect(users[0].field).to.be.eql(null);
+            } else if (dialect === 'postgres' || dialect === 'postgres-native') {
+              //Empty Geometry data [0,0] as per https://trac.osgeo.org/postgis/ticket/1996
+              expect(users[0].field).to.be.deep.eql({ type: "Point", coordinates: [0,0] });
+            } else {
+              expect(users[0].field).to.be.deep.eql(point);
+            }
           });
-        }).then(function () {
```
sushantdhiman: Why this test has been removed?

test/integration/data-types.test.js
```diff
@@ -269,82 +266,90 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
...
+      const Type = new Sequelize.GEOMETRY();
 
-    // MySQL 5.7 or above doesn't support POINT EMPTY
-    if (dialect === 'mysql' && semver.gte(current.options.databaseVersion, '5.7.0')) {
-      return;
-    }
+      return testSuccess(Type, { type: "Point", coordinates: [125.6, 10.1] });
+    });
 
-    return new Sequelize.Promise((resolve, reject) => {
```
sushantdhiman: Why this test is removed?

test/integration/data-types.test.js
```diff
@@ -241,25 +248,15 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
   });
 
   it('calls parse and stringify for REAL', function () {
-    var Type = new Sequelize.REAL();
+    const Type = new Sequelize.REAL();
 
     return testSuccess(Type, 1.5);
   });
 
-  it('calls parse and stringify for GEOMETRY', function () {
```
sushantdhiman: This one too, why you have removed GEOMETRY tests?

sushantdhiman: @rokoroku Thats ok, i thought you have removed those tests for good :laughing:

test/integration/dialects/sqlite/dao.test.js
```diff
@@ -61,26 +64,242 @@ if (dialect === 'sqlite') {
...
           }).get(0);
-        }).then(function (user) {
+        }).then(user => {
           expect(user.projects[0].get('dateField')).to.be.an.instanceof(Date);
           expect(user.projects[0].get('dateField')).to.equalTime(new Date(1990, 5, 5));
         });
       });
     });
 
+    describe('json', function() {
```
sushantdhiman: We should be able to run these tests for any dialect that supports JSON, right ?

rokoroku: I'm not sure... I just copied this test suite from [postgres/dao.test.js#L97](https://github.com/sequelize/sequelize/blob/master/test/integration/dialects/postgres/dao.test.js#L97) and there would be some different test like:
-  `->>` operator in [postgres/dao.test.js#L221](https://github.com/sequelize/sequelize/blob/master/test/integration/dialects/postgres/dao.test.js#L221) 
- `json_extract(...)` method in [sqlite/dao.test.js#L232](https://github.com/sequelize/sequelize/pull/7094/files/966d159346a926739ab9f0a336c8c2b8def9775e#diff-30bcb6f77557dd6d0bb295e7815e0d1fR232)

But there are certainly common tests. So where should we move them?

sushantdhiman: Common tests should be moved to `abstract/dao.test.js`

rokoroku: Oh, there is `abstract`! OK, I'll try to move them.

lib/dialects/abstract/query-generator.js
```diff
@@ -2228,6 +2228,19 @@ const QueryGenerator = {
     return result ? result : '1=1';
   },
 
+  // A recursive parser for nested where conditions
+  parseConditionObject(conditions, path) {
+    path = path || [];
+    return _.reduce(conditions, (result, value, key) => {
+      if (_.isObject(value)) {
+        result = result.concat(this.parseConditionObject(value, path.concat(key))); // Recursively parse objects
+        } else {
```
sushantdhiman: extra space here `if` and `else` doesn't align

sushantdhiman:
@rokoroku LGTM, Just one small change with indentation.

@sequelize/core Please review

lib/dialects/abstract/query-generator.js
rokoroku: Oops, good catch.

rokoroku:
Wait, I found a better way to handle JSON path extraction method.

Actually, `whereItemQuery` method in [sqlite/query-generator.js#L169](https://github.com/sequelize/sequelize/pull/7094/files/966d159346a926739ab9f0a336c8c2b8def9775e#diff-82040f2fda08efca8a2af4e257e2464aR169) is duplicated because default path extraction method have `postgres` syntax. (see [abstract/query-generator.js#L1945](https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L1945))

But if we replace the above line (L1945) with `this.jsonPathExtractionQuery(...)`, I can remove unnecessary `whereItemQuery` override from `sqlite/query-generator.js` 

So I'd suggest a new private API for `query-generator` like below:
```javascript
  /**
   * Generates an SQL query that extract JSON property of given path.
   *
   * @param   {String}               column    The JSON column 
   * @param   {String|Array<String>} path      The path to extract
   * @returns {String}                         The generated sql query
   * @private
   */
  jsonPathExtractionQuery(column, path): String
```

Which means every dialect supporting JSON path extracting have to override above method.

How do you guys think about this suggestion?

sushantdhiman:
@rokoroku Sounds good to me

rokoroku:
@sushantdhiman Then I'm gonna change the `whereItemQuery` method of `abstract/query-generator.js` like I suggested, leaving default behavior (generating postgres syntax) to prevent breaking existing tests.

So `jsonPathExtractionQuery` method of `sqlite/query-generator.js` and `postgres/query-generator.js` will be updated as I suggested.

rokoroku:
While composing common methods to the `abstract/query-generator`, I found an inconsistency about the casting syntax of postgres.

Previously it uses `::` casting syntax when generating json query before, but I changed it using `Sequelize.Cast` (See the changes from here: [abstract/query-generator.js#dccadc-L1973](https://github.com/sequelize/sequelize/pull/7094/commits/dccadc82bea017a57b4cc8db5540ebb955445768#diff-def52474024b12c7d133bb504cdf57d1R1973)) and added a new `handleSequelizeMethod` override for `Sequelize.Cast` in [postgres/query-generator.js#dccadc-R234](https://github.com/sequelize/sequelize/pull/7094/commits/dccadc82bea017a57b4cc8db5540ebb955445768#diff-7e347c4738c97261969fbeb802a3c363R234)

Since it's difficult to make `::` casting only for json method (without overriding `whereItemsQuery`), so I changed all existing test for postgres as using `::` syntax for casting now.

Now I have are three options: 
1. Use `::` syntax for all postgres casting (current change)
2. Use `CAST()` syntax for all as well as json operations. (needs all test expectation to be updated)
3. Override `whereItemsQuery` in `postgres/query-generator` again to use the `::` syntax only for json operations.

Please leave some comment about this, I will follow the decision.

sushantdhiman:
`(1)` This means we will have different casting syntax for different dialect, right ? Thats not an option 

We can use `(2) Use CAST() syntax as well as json operations. (needs all json test to be updated)`, this means JSON operation will use CAST as well, if its not very difficult to do, its a good option.

Finally, if its necessary than you can override `whereItemsQuery` in postgres

rokoroku:
@sunshinewyin Agreed, and I think `::` casting only for JSON methods is not necessary. The existing code was written in this way, but I prefer the way to use consistent syntax `(2)`.

Then I will proceed with `(2)` :)

lib/dialects/postgres/query-generator.js
```diff
@@ -101,58 +101,128 @@ const QueryGenerator = {
     if (!schema) {
       schema = 'public';
     }
-    return 'SELECT pk.constraint_type as "Constraint", c.column_name as "Field", ' +
```
sushantdhiman: There is only indentation change here, please revert it to original

sushantdhiman:
Would like reviews from @sequelize/core

lib/dialects/postgres/query-generator.js
rokoroku: Sure, maybe my IDE formatted these lines.

rokoroku: @sushantdhiman I reverted the indentation, but trailing whitespace has been removed since `.editorconfig` overrides ``trim_trailing_whitespace = true``.

If you want revert the whitespaces as well, I'll try different editor.

sushantdhiman: No I dont want extra trailing spaces either. No change

sushantdhiman:
Now we are ready for final reviews form @sequelize/core team

@rokoroku please change description for PR to match recent changes, so its easy for others to review. 👍 great work

rokoroku:
Reviewed again, cleaned up some codes and updated description of changes :)

lib/dialects/abstract/query-generator.js
```diff
@@ -2228,6 +2230,38 @@ const QueryGenerator = {
...
+   *
+   * @param   {String}               column  The JSON column
+   * @param   {String|Array<String>} [path]  The path to extract (optional)
+   * @returns {String}                       The generated sql query
+   * @private
+   */
+  jsonPathExtractionQuery(column, path) {
+    const paths = _.toPath(path);
+    const pathStr = `{${paths.join(',')}}`;
+    const quotedColumn = this.isIdentifierQuoted(column) ? column : this.quoteIdentifier(column);
```
janmeier: In what cases is the identifier already quoted?

lib/dialects/postgres/query-generator.js
```diff
@@ -121,38 +121,108 @@ const QueryGenerator = {
...
+   * @returns {Boolean}       true if the given statement is json function
+   * @throws  {Error}         throw if the statement looks like json function but has invalid token
+   */
+  checkValidJsonStatement(stmt) {
+    if (!_.isString(stmt)) {
+      return false;
+    }
+
+    // https://www.postgresql.org/docs/current/static/functions-json.html
+    const jsonFunctionRegex = /^\s*((?:[a-z]+_){0,2}jsonb?(?:_[a-z]+){0,2})\([^)]*\)/i;
```
janmeier: These regexes can be placed outside the function, so they don't need to be compiled for each invocation

test/integration/dialects/abstract/dao.test.js
```diff
@@ -0,0 +1,240 @@
+'use strict';
```
janmeier: Please move this to `test/integration/json.test.js` or similar

test/integration/dialects/abstract/dao.test.js
```diff
@@ -0,0 +1,240 @@
...
+
+const chai = require('chai')
+  , expect = chai.expect
+  , Support = require(__dirname + '/../../support')
+  , Sequelize = Support.Sequelize
+  , current = Support.sequelize
+  , DataTypes = require(__dirname + '/../../../../lib/data-types');
+
+describe('DAO', function () {
```
janmeier: DAO = model

test/unit/dialects/postgres/query-generator.test.js
```diff
@@ -897,5 +897,53 @@ if (dialect.match(/^postgres/)) {
         });
       });
     });
+
+    describe('JSON', function () {
```
janmeier: OPTIONAL: 
If you are a completely awesome human being, please rewrite these as unit tests ,along the lines of https://github.com/sequelize/sequelize/blob/master/test/unit/sql/json.test.js

test/integration/dialects/abstract/dao.test.js
rokoroku: you mean this file?

janmeier: Yep

lib/dialects/postgres/query-generator.js
rokoroku: I didn't want to polute the query-generator class, but I agree with your suggestion. 
May I place these regexps as class properties? 
if so, should I use uppercased name? (e.g. `const JSON_FUNCTION_REGEX = /.../`)

janmeier: Either that, or set them as (static) properties in the class, so we can keep them close to the function that uses them

felixfbecker: @janmeier you sure V8 will not optimize this?
@rokoroku I would uppercase it if it's a constant (not if it's a property)

lib/dialects/abstract/query-generator.js
rokoroku: If it called via `whereItemQuery`.  [abstract/query-generator.js#1945](https://github.com/sequelize/sequelize/pull/7094/files/dd8eba5554616b4793097d6f663664c1b71c70a3#diff-def52474024b12c7d133bb504cdf57d1R1949)
But not quoted when it called via Sequelize.json(). see[postgres/query-generator.js#211](https://github.com/sequelize/sequelize/pull/7094/files/dd8eba5554616b4793097d6f663664c1b71c70a3#diff-7e347c4738c97261969fbeb802a3c363R211)

lib/dialects/postgres/query-generator.js
rokoroku: Just found this from SO: 
http://stackoverflow.com/questions/14352100/does-v8-cache-compiled-regular-expressions-automatically

So could we just leave it ?

janmeier: yES!

test/integration/dialects/abstract/dao.test.js
rokoroku: Sure :)

test/unit/dialects/postgres/query-generator.test.js
rokoroku: Ummm... I'll try to get closer to the awesome human being.

janmeier:
<3

kevingilbert100:
Can someone please provide a example use-case for finding a entry in sqllite filtering the JSONB type?

kevingilbert100:
Disregard - When will this be in a NPM build?

Issue #7075, opened 2017-01-06
bulkCreate + options.returning on a table with hasTrigger enabled does not work


iamjoyce:
## What you are doing?
```js
var Task = sequelize.define('task', {
    title: Sequelize.STRING,
    rating: { type: Sequelize.STRING, defaultValue: 3 }
}, {
    hasTrigger: true
});

Task.bulkCreate([{
    title: "Fix bug",
    rating: "5"
}, {
    title: "Write documentation",
    rating: "1"
}], {
    returning: true
});
// Error happens
```

## What do you expect to happen?
Promise should return a list of successfully inserted instances **with resulting values of autoIncrement attributes**.

## What is actually happening?
Unhandled rejection SequelizeDatabaseError: The target table '%TABLE_NAME_HERE%' of the DML statement cannot have any enabled triggers if the statement contains an OUTPUT clause without INTO clause.

__Dialect:__ mssql
__Database version:__ SQL Server 2014
__Sequelize version:__ 3.24.3

iamjoyce:
I've just realised a way to get resulting values of autoIncrement attributes without any modifications to the code.


Instead of `{returning: true}`, use `{individualHooks: true}`. Now that this works, I am not sure if this is supposed to be a bug or intended feature.

rainabba:
IndividualHooks did help me work around this, but it did so by destroying the bulkCreate behavior and causing 1 insert for each record. I'm looking into query-generator.js. It looks like bulkInsertQuery() doesn't have the tmp table logic included in insertQuery (based on hasTrigger). Can anyone confirm this before I start on a PR? There already a branch addressing this?

papb:
I will reopen this since it is still an issue.

Hi @rainabba! Thanks for being willing to help. If you can create a PR, that would be great!! Please look at PRs [7095](https://github.com/sequelize/sequelize/pull/7095/files) and [7096](https://github.com/sequelize/sequelize/pull/7096/files) before starting, perhaps you can use some of those ideas. Thanks! Let me know if you need further help.

Judahh:
No prediction when this will be solved?

Judahh:
I have a similar issue, but even with "individualHooks: true", I receive the same error, the difference is I'm using include that has an Sequelize Model that has triggers.

klondikemarlen:
I can confirm this remains reproducible with the current npm `latest` v7 packages: `@sequelize/core` and `@sequelize/mssql` `7.0.0-alpha.48`.

[...]

PR #7096, opened 2017-01-11
bulkCreate now supports returning on tables with hasTrigger enabled.

iamjoyce:
### Pull Request check-list

### Description of change

Refer to issue #7075.

I apologise for creating two pull requests. I selected the wrong base for the first pull request, which (I think) it was why the integration build failed.

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -233,17 +233,18 @@ var QueryGenerator = {
   bulkInsertQuery: function(tableName, attrValueHashes, options, attributes) {
     options = options || {};
     attributes = attributes || {};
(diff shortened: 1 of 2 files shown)


iamjoyce:
Somehow I can't get docker or npm test to work even on clean version.

Can someone verify if the following test case is valid? It is intended to be in _/test/integration/trigger.test.js_.

It is supposed to check if `Id` of the first element in the array is `null`.

- If `null` = test failed.
- If not `null` = test passed.

```js
       it('should return output rows after bulk insert', function() {
         return User.bulkCreate([{
           username: 'triggertest1'
         }, {
           username: 'triggertest2'
         }, {
           username: 'triggertest3'
         }, {
           username: 'triggertest4'
         }], {returning: true}).then(function (users) {
           // This assumes that if the first instance's Id is not null, all subsequent instances' Id is also not null.
           return expect(users[0].Id).to.not.equal(null);
         });
       });
```

PR #7107, opened 2017-01-14
Fix documentation typo for creating belongsTo associations with an alias

jharia:
The include should be a hash containing the "model" and "as" keys


sushantdhiman:
Documentation is correct here, `Creator` is a association object which itself aware about name used. Changes you made will cause code to fail because `Creator` is not a Model :)

PR #7109, opened 2017-01-14
Generic Pool Update

### LINKED ISSUE #7068: options.pool.validateConnection
Hi,
the option is mentioned here: https://github.com/sequelize/sequelize/blob/master/docs/api/sequelize.md

Searching the repository it seems to be only mentioned in docs: https://github.com/sequelize/sequelize/search?utf8=%E2%9C%93&q=validateConnection

And when I tried to do this with latest sequelize, nothing seems to happen:
```
const poolConfig = {
    min: 1,
    max: maxConnections,
    idle: 60000,
    validateConnection: function (arg1) {
      console.log('validateConnection', this, arg1);
      return false;
    }
  };
```

Is this feature actually implemented?

sushantdhiman:
### Pull Request check-list

### Description of change

- Bump to latest version of generic pool
- Fix issue where replication priorities were not used properly

Fix #7068, #4711
Close #5776

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -6,15 +6,17 @@ const _ = require('lodash');
 const Utils = require('../../utils');
 const debug = Utils.getLogger().debugContext('pool');
 const semver = require('semver');
+const timers = require('timers');
+
 const defaultPoolingConfig = {
   max: 5,
   min: 0,
   idle: 10000,
+  acquire: 10000,
   handleDisconnects: true
 };
 
 class ConnectionManager {
-
   constructor(dialect, sequelize) {
     const config = _.cloneDeep(sequelize.config);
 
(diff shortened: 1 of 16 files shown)


janmeier:
Seems like a reasonable approach @sushantdhiman , ping me when you need i final review

sushantdhiman:
@janmeier Ping

lib/dialects/abstract/connection-manager.js
```diff
@@ -51,47 +56,68 @@ class ConnectionManager {
...
-        name: 'sequelize-connection',
-        create: (callback) => {
-          this._connect(config).nodeify((err, connection) => {
-            debug(`pool created max/min: ${config.pool.max}/${config.pool.min} with no replication`);
-            callback(err, connection); // For some reason this is needed, else generic-pool things err is a connection or some shit
-          });
-        },
+      this.pool = Pooling.createPool({
+        create: () => new Promise((resolve) => {
+          this
```
janmeier: No need to wrap in a new promise, you can just return the chain

lib/dialects/abstract/connection-manager.js
```diff
@@ -118,67 +144,91 @@ class ConnectionManager {
...
+        return Promise.join(
+          this.pool.read.clear(),
+          this.pool.write.clear()
+        );
       },
-      drain: (cb) => {
-        this.pool.write.drain(() => {
-          this.pool.read.drain(cb);
+      drain: () => {
+        return this.pool.write.drain(() => {
```
janmeier: These should be able to drain in parallel

lib/dialects/postgres/index.js
```diff
@@ -12,8 +12,6 @@ class PostgresDialect extends AbstractDialect {
     super();
     this.sequelize = sequelize;
     this.connectionManager = new ConnectionManager(this, sequelize);
-    this.connectionManager.initPools();
```
janmeier: already handled in the call to `super`, right?

sushantdhiman: Yes

lib/dialects/abstract/connection-manager.js
```diff
@@ -6,40 +6,45 @@ const _ = require('lodash');
...
-      config.pool =_.defaults(config.pool || {}, defaultPoolingConfig, {
-        validate: this._validate.bind(this)
-      }) ;
-    } else {
+    if (config.pool === false) {
       throw new Error('Support for pool:false was removed in v4.0');
     }
 
+    config.pool =_.defaults(config.pool || {}, defaultPoolingConfig, {
+      validate: this._validate.bind(this)
```
sushantdhiman: Good idea, user will be able to override it as well

lib/dialects/abstract/connection-manager.js
sushantdhiman: We dont want to resolve if there is an error

PR #6652, opened 2016-09-30
V3 Fix sourceKey for hasMany, issue: #4258

Paxa:
_Thanks for wanting to fix something on Sequelize - we already love you long time! Please delete this text and fill in the template below. If unsure about something, just do as best as you're able._

_If your PR only contains changes to documentation, you may skip the template below._
### Pull Request check-list

_Please make sure to review and check all of these items:_
- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Have you added an entry under `Future` in the changelog?

_NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open._
### Description of change

_Please provide a description of the change here._

Closes #4258

PR #6651, opened 2016-09-29
V3 fix belongsto foreignkeys

eseliger:
### Pull Request check-list

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Changelog entry

### Description of change

Fixes the reversed keys issued in #6400 for a belongsTo-association join.

Closes #6400

PR #7115, opened 2017-01-17
V3 Added sourceKey support for hasOne

twolfson:
### Pull Request check-list

### Description of change

In my current repo, I have a scenario where using `belongsTo` with includes for selects isn't feasible. The scenario is:

- User has tasks
- Each of these tasks can have an active reminder as well as past inactive reminders
- We want to have a screen which lists these tasks and their reminders

The model structure is:

```
User({id})
Task({id, user_id, active_reminder_id})
Reminder({id, task_id, user_id})

User.hasMany(Task);
Task.belongsTo(User);

Task.hasOne(Reminder, {as: 'active_reminder', sourceKey: 'active_reminder_id', foreignKey: 'id'});
Reminder.belongsTo(Task);
Reminder.belongsTo(User);
```

There are alternatives to this:

- Add an `active` boolean to the `Reminder` table but then we risk 0 or 2 becoming enabled which requires prevention enforcement via triggers

But I would prefer to not go down those rabbit holes as it seems likely to cause more problems later on.

This PR is on top of #7113 and has completed the logic for `sourceKey` support in `hasOne` but I'm unable to test it due to Sequelize wanting to generate a `userId` association despite not requesting one in the tests. As a result, we are getting circular dependency issues on table drop/creation =/

Any help in that department or telling me that I'm doing it wrong would be appreciated =)

Related issues: #4258, #6652, #7113


twolfson:
Okay, I took another look at this today. It looks like `hasOne` has a lot of built-in logic for updating attributes on the target instead of the source whereas we want to update the source. This makes me think we actually want to use `belongsTo` but we can't because it doesn't set associations for `include` on the `source`. That is I can't do:

```
User({id, active_task_id})
Task({id})

Task.belongsTo(User, {as: 'active_task'});

User.findAll({include: [{model: Task, as: 'active_task'}]});
```

Thankfully this PR **seems** to work for selects which is all I really needed (I don't do nested saves as they currently don't support transactions):

```
User({id, active_task_id})
Task({id})

User.hasOne(Task, {as: 'active_task', sourceKey: 'active_task_id', foreignKey: 'id'});

User.findAll({include: [{model: Task, as: 'active_task'}]});
```

That being said, it would be great to find a solution that works and fits this repo (e.g. `belongsTo` that defines association on `source`)

twolfson:
Aaaaand after digging around some more I found out what I was doing wrong. I shouldn't be using `model/as`, I should be using `association` for include -_-;;

https://github.com/sequelize/sequelize/pull/6651/files#diff-bea10850a51c66a27b66d3cbbd7fc57fR896

```
User({id, active_task_id})
Task({id})

Task.belongsTo(User, {as: 'active_task'});

User.findAll({include: [{association: User.associations.active_task}]});
```

twolfson:
Sorry, that is:

```
User.belongsTo(Task, {as: 'active_task'});
```

and we don't need to use association:

```
User.findAll({include: [{model: Task, as: 'active_task'}]});
```

That being said, using a `belongsTo` on the parent feels very strange but I guess it works =/

mattfysh:
@twolfson - would it be possible to make this change also for V4? I would owe you one!

PR #7118, opened 2017-01-17
Fix `Instance.decrement` precision problems

### LINKED ISSUE #7112: Instance.decrement losing precision in big numbers
## What you are doing?
When calling `Instance.decrement` to decrement a specific column value, _Sequelize_ internally it uses the `Instance.increment` with the negative value. When converting the value to its negative form it loses precision right [here](https://github.com/sequelize/sequelize/blob/6f9a53bfa542f962e5df1892657d607a2c6318ab/lib/model.js#L3840) (**due to Javascript floating point problem**).

For a given big number value `'1123.123456789012345678'` in its _string_ form, I was expecting a query like this:
```sql
UPDATE x = x - '1123.123456789012345678';
```

It is using the `Instance.increment` with the Javascript negative form of the input value:
```sql
UPDATE x = x + '-1123.1234567890124';
```

Example of the JS issue:
```bash
❯ node
> const x = '1123.123456789012345678'; -x
-1123.1234567890124
```

The issue is actually in the negative form of the value. 

I'm willing to add a PR to fix this, but we have 2 choices:
- Fix the interpretation of the negative form of the value (keeping the hack)
- Implement the decrement method in the query interface to use the database subtract operator and to prevent the mutation of the input value (**IMHO** I prefer this one).

__Dialect:__  postgres
__Database version:__ 9.5.4
__Sequelize version:__ 3.23.2

pedrobranco:
Closes #7112.

### Pull Request check-list

### Description of change

As suggested by the issue #7112:
- Reuse and rename the `query-generator::incrementQuery()` to `query-generator::incrementOrDecrementQuery()`, which now receives a boolean argument to represent the intention of incrementing or decrementing.
- Rename `query-interface::increment()` to `query-interface::incrementOrDecrement()` to pass the option `options.increment`.
- Change the `model::decrement()` implementation to call `model::increment()` with the option `increment = false`.

Doubts: Tests are passing (`npm run test-docker`), but should we add any test for this internal change?

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -377,6 +377,7 @@ const QueryGenerator = {
   /*
     Returns an update query.
     Parameters:
+      - operator -> String with the arithmetic operator (e.g. '+' or '-')
       - tableName -> Name of the table
       - values -> A hash with attribute-value-pairs
       - where -> A hash with conditions (e.g. {name: 'foo'})
@@ -385,7 +386,7 @@ const QueryGenerator = {
                  If you use a string, you have to escape it on your own.
    @private
   */
-  incrementQuery(tableName, attrValueHash, where, options) {
+  arithmeticQuery(operator, tableName, attrValueHash, where, options) {
     attrValueHash = Utils.removeNullValuesFromHash(attrValueHash, this.options.omitNull);
 
     const values = [];
@@ -402,7 +403,7 @@ const QueryGenerator = {
 
     for (const key in attrValueHash) {
       const value = attrValueHash[key];
-      values.push(this.quoteIdentifier(key) + '=' + this.quoteIdentifier(key) + ' + ' + this.escape(value));
+      values.push(this.quoteIdentifier(key) + '=' + this.quoteIdentifier(key) + operator + this.escape(value));
     }
 
     options = options || {};
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -638,7 +638,17 @@ class QueryInterface {
   }
 
   increment(instance, tableName, values, identifier, options) {
-    const sql = this.QueryGenerator.incrementQuery(tableName, values, identifier, options.attributes);
+    const sql = this.QueryGenerator.arithmeticQuery('+', tableName, values, identifier, options.attributes);
+
+    options = _.clone(options) || {};
+
+    options.type = QueryTypes.UPDATE;
+    options.instance = instance;
+    return this.sequelize.query(sql, options);
+  }
+
+  decrement(instance, tableName, values, identifier, options) {
+    const sql = this.QueryGenerator.arithmeticQuery('-', tableName, values, identifier, options.attributes);
 
     options = _.clone(options) || {};
 
--- a/test/unit/dialects/mssql/query-generator.test.js
+++ b/test/unit/dialects/mssql/query-generator.test.js
@@ -171,5 +171,34 @@ if (current.dialect.name === 'mssql') {
         mssql: 'ALTER TABLE [myTable] DROP [myColumnKey]'
       });
     });
+
+    test('arithmeticQuery', () => {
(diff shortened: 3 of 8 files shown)


lib/dialects/abstract/query-generator.js
```diff
@@ -385,12 +386,17 @@ const QueryGenerator = {
                  If you use a string, you have to escape it on your own.
    @private
   */
-  incrementQuery(tableName, attrValueHash, where, options) {
+  incrementOrDecrementQuery(increment, tableName, attrValueHash, where, options) {
```
sushantdhiman: rather than `increment`, pass `operator` directly instead

lib/query-interface.js
```diff
@@ -637,8 +637,8 @@ class QueryInterface {
     );
   }
 
-  increment(instance, tableName, values, identifier, options) {
-    const sql = this.QueryGenerator.incrementQuery(tableName, values, identifier, options.attributes);
+  incrementOrDecrement(instance, tableName, values, identifier, options) {
```
sushantdhiman: We will still keep `increment / decrement` in query generator , which will call `incrementOrDecrementQuery` with different operators

sushantdhiman:
@pedrobranco 

PR needs some work

- It will be better to expose `increment`  and `decrement` methods which pass different `operator` to `incrementOrDecrementQuery`
- we need unit test around `increment`, `decrement` and `incrementORDecrementQuery`

lib/dialects/abstract/query-generator.js
pedrobranco: Sure.

lib/query-interface.js
pedrobranco: In `model.js`: 
- we keep the `decrement()` calling `this.increment()` ? or 
- we replicate the same code from `increment()` to `decrement()` ? or
- abstract the `increment()` internals to another method to be able to call `decrement()` in query-generator?

sushantdhiman: `Model.decrement` will call `Model.increment`  with `{increment: false}` as its is doing already, at last step we use `increment` to choose between `incrementQuery` or `decrementQuery`.

We will keep `incrementOrDecrementQuery` private, may be rename it to `arithmeticQuery` as well

pedrobranco:
@sushantdhiman PR addressed. Can you explain why tests are failing?

lib/query-interface.js
```diff
@@ -638,7 +638,17 @@ class QueryInterface {
   }
 
   increment(instance, tableName, values, identifier, options) {
-    const sql = this.QueryGenerator.incrementQuery(tableName, values, identifier, options.attributes);
+    const sql = this.QueryGenerator.arithmeticQuery(' + ', tableName, values, identifier, options.attributes);
```
sushantdhiman: we can run this without adding extra spaces, right.

If yes, we should not add any extra spaces. May be query generator should handle that if absolutely required, otherwise send query without spaces

pedrobranco: Right. I just kept the same format as before.

test/unit/dialects/mssql/query-generator.test.js
```diff
@@ -171,5 +171,34 @@ if (current.dialect.name === 'mssql') {
         mssql: 'ALTER TABLE [myTable] DROP [myColumnKey]'
       });
     });
+
```
pedrobranco: @sushantdhiman Can you review this queries since it does not run in tests?

sushantdhiman: looks good, beside we have integration tests for increment/decrement so its verified

sushantdhiman:
just needs a changelog entry and its good to go 👍

pedrobranco:
>just needs a changelog entry and its good to go 👍

Done.

pedrobranco:
Thanks @sushantdhiman.

PR #7123, opened 2017-01-18
Oracle Support on master

ggrimbert:
Hi,

This is the first version of Oracle support done by my company. 
The minor version is Oracle 12c. 
This pull request follows the first I've made (https://github.com/sequelize/sequelize/pull/6903). This one is on the master branch.

This version contains everything needed for using Oracle : 
 - Connecting to Oracle.
 - INSERT / UPDATE / SELECT / DELETE queries.
 - Managing transactions.
 - Call to stored procedure (with / without parameters).
 - All tests are passing
 - Docs are updated 

My company needs this Oracle version for our new server, so the work will continue soon. 

Connection Pooling is not relevant according to node-oracledb. 

TODO : 
- [ ]  resultSets
- [X] upsert
- [X] long length aliases support

ResultsSets needs some time to be implemented, for now, maxRows can be setted in the dialectOptions (default to 1000).
I can't do the upsert support as I can't understand how Oracle works with it.

Oracle doesn't support aliases / columns / tables with name's length > 31.<del> For now, nothing has been done. It will come with another pull request.</del> An aliases replacement is done to make all queries work.

A docker image has been created and is available : https://hub.docker.com/r/konnecteam/oracle12c-database/

I don't know how to integrate it with Sequelize builds, so if someone can help me.


janmeier:
I'll try to add the docker + travis setup - Could you please give me push access to your fork, if you haven't already? https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

ggrimbert:
It's already done ;)

janmeier:
@ggrimbert Is there any way you can make the image smaller? It's taking forever to pull :D. I don't know if it's ubuntu adding a lot of unneded stuff (the other images we use are based in debian:jessie)

janmeier:
Okay, [it seems](https://microbadger.com/images/konnecteam/oracle12c-database) that the bulk of the image is actually the oracle file. This makes me wonder if we should make an exception and just install the oracle binary instead. https://github.com/cbandy/travis-oracle should make it pretty easy - I've used it so far to download the client and sdk zip files that we need to be able to build the node library

janmeier:
Alright, got everything installed and up and running (I think) - Can someone help me parse this connection error :D 
>      SequelizeConnectionError: ORA-12514: TNS:listener does not currently know of service requested in connect descriptor

ggrimbert:
This can be a network error or a problem in your connections parameters.

if your behind a proxy, check that you don't have to set the domain name for the hostname (in my company, we have to add .kimdomain.local at the end of the hostname.

I know that I had problems when the service name wasn't the one by default (check https://github.com/oracle/docker-images/issues/246#issuecomment-270485255)

janmeier:
This is on travis though - The only thing I've changed from test/config/config.js is setting host to 'localhost'

ggrimbert:
If you have access to a system terminal, you can try 
```bash 
lsnrctl status
```
This should print the listener's information.

Have you tried from sqldeveloper ?

ggrimbert:
I've just checked https://github.com/cbandy/travis-oracle. You can't use it as it is for Oracle 11g, but the minor version supported is Oracle 12c.
.

janmeier:
Unfortunately I don't have ssh or external access to the server, so I can't access it from sqldeveloper. 

> You can't use it as it is for Oracle 11g, but the minor version supported is Oracle 12c.

Okay, I was worried about that :(. As far as I can see the express (free version?) only supports the previous version?


I just tried pulling the image locally - And it seems that on top of the time to pull, the database also needs several minutes to initialize??

```bash
oracle | ls: cannot access /u01/app/oracle/oradata: No such file or directory
oracle | Database not initialized. Initializing database.
oracle | Starting tnslsnr
oracle | Copying database files
oracle | 1% complete
oracle | 3% complete
oracle | 11% complete
oracle | 18% complete
oracle | 37% complete
oracle | Creating and starting Oracle instance
oracle | 40% complete
oracle | 45% complete
```

Is there anything we can do to prevent that? Having to use +5 minutes to initialize the database is not going to work - Since the other dialects can both initialize the db, and run the tests in that timespan :D. Maybe you can add some commands to the docker-file so the databse is already fully initialized once we pull it?

ggrimbert:
> Okay, I was worried about that :(. As far as I can see the express (free version?) only supports the previous version ?

You can use the standard version for tests, I had no problem with that.

The DB initialization is done only once if the image is reused, if you don't download it at each time, the problem should not appear.
Otherwise, the only solution I see to prevent this is, is to work with my docker image, so I have to push it (not really sure about this, if you're interested I can test and let you know).

For the time (it's about 10 - 15 min for me), unfortunately, it's the normal time for the first initialization of an Oracle DB. Normally it's done after the installation of the DB.

christianbundy:
Just wanted to chime in on the database initialization time: it's going to need to initialize each time it starts up with a new volume, so unless Travis is saving volumes (which I don't think happens) this will happen each time. It *may* be possible to create an Oracle image that already has the instantiation finished, but we'd be deviating pretty hard from the canonical way to spin up Oracle in a Docker container.

ggrimbert:
Any update on this ?

janmeier:
@ggrimbert Sorry, haven't had the time to look at this yet

@christianbundy Saving a completely initialized container is actually the thing I'd like to try. I know that is pretty far from regular docker usage, but we really can't have a single job taking 15-20 minutes because the database needs to initialize. On top of that I don't really understand what it's doing. Other databases don't need this long! :D

janmeier:
The docker setup is still not fast, but at least it works now :) - Ready for you to continue work @ggrimbert

janmeier:
Paging @cjbj, @brunoborges, @gvenzl - Since you are top contributors on node-oracledb / oracle-images perhaps you have some suggestions as to how we can make the database initialization faster. Any ideas (docker or otherwise) are welcome

ggrimbert:
Please, don't tell me I should make it work for node V4...

janmeier:
You definitely should! :) We plan on support node 4 at least as long as it is in the LTS phase (https://github.com/nodejs/LTS)

gvenzl:
Thanks folks for reaching out to us! Your request isn't a new one to us. Unfortunately, the way that volumes work today in Docker and the fact that Docker won't copy existing files on a host-only mounted volume means that we need to instantiate the DB on container startup.

If I understood correctly, given that Travis doesn't save volumes, you just want to have a prebuilt DB inside the container that is just started with the container.
This can be achieve by tweaking the Dockerfile a bit. There is an example on Github: https://github.com/oracle/docker-images/tree/master/OracleDatabase/samples/prebuiltdb

Essentially, the steps are:
- Remove the VOLUME instruction
- Build the image
- Startup the container
- Optional: perform administrative work, i.e. set default SYS password, create users and/or tablespaces, etc.
- Stop container
- Commit the container into a new image

Hope this helps! If you need more info or help, please let me know!

cjbj:
@janmeier I was discussing this with @gvenzl this morning;  if we've misunderstood, yell out.

@ggrimbert thanks for contributing support to sequelize.  This is pretty cool.  If you have node-oracledb questions, I'm the contact for that.  Some quick comments: the next node-oracledb release will have a `fetchAsBuffer` setting to fetch BLOBs directly as buffers (so they don't need to be streamed).  Also Oracle Database 12.2 has increased the identifier length to 128 (from 30), see http://docs.oracle.com/database/122/LNOCI/release-changes.htm#LNOCI005

ggrimbert:
'Oracle Database 12.2 has increased the identifier length to 128 (from 30)', does it work with everything (aliases, table names, transaction names) ? If yes, this is awesome !

sushantdhiman:
> 'Oracle Database 12.2 has increased the identifier length to 128 (from 30)', does it work with everything (aliases, table names, transaction names) ? If yes, this is awesome !

It should, as Postgres has same limit set to 63 and I rarely see this limit crossed except for deeply nested includes

ggrimbert:
@cjbj  Is this version already available ? And how can I get it ? On the Oracle download page, I only have Oracle 12.1

ggrimbert:
@cjbj, I havz a problem when running my tests : 

I create a user, do some work and the try to delete it. But, each time I try to delete the user (or a table), I get the following error :  ```ORA-00600 : internal error code, arguments : [12811],[402455]```.

Do you have any idea of what causes this ?

I don't have access to Oracle metalink, and Google doesn't answer me...

janmeier:
> If I understood correctly, given that Travis doesn't save volumes, you just want to have a prebuilt DB inside the container that is just started with the container.

Exactly! That was also my initial thoughts about how this should work, so I will give it another try - Thanks for the pointers!

@cjbj Actually I was just looking for pointers about oracle + docker setup, but any comments are very much appereciated, since I don't have experience with oracle - So thanks for the review! :taco:

cjbj:
@ggrimbert checking Oracle Metalink I can see there have been a few reports of ORA-600 [12811].  It seems like a DB-side issue (not node-oracledb), but the cause varied between versions.  (I'm looking at Metalink Doc ID 138384.1). Any trace files?  Email me with details such as the exact client & server versions.  See my profile for my address.

Oracle Database 12.2 is currently available on our cloud (free trials available; and at various conferences we are giving away credit).  In the last couple of weeks it also just got released to customers 'on-premise' using our engineered systems (e.g. Exadata).  General on-premise release is planned.

The fine print on the Oracle DB 12.2's longer identifiers is given in http://docs.oracle.com/database/122/SQLRF/Database-Object-Names-and-Qualifiers.htm#SQLRF51129

ggrimbert:
With my last commit, all tests pass for me.
But with travis, it seems that a request hasn't finished.
Is there a way to know which one ?

ggrimbert:
Build still fails, but works when I launch on my computer.

I don't understand what happens...

Can someone help ?

ggrimbert:
Well, after several attemps, I still have the error ```
Expected 0 running queries. 1 queries still running in [ORACLE] Model constructor allows us to customize the error message for unique constraint ``` for two tests (always the same).

And with my my last commit https://github.com/sequelize/sequelize/pull/7123/commits/b8c7e0f24a5bc70e5b147084286df507979eaf52, the error only happen with node V4.

When I launch the test alone, it doesn't happen. It really seems like a timing issue.
I don't reproduce on my own server. For now I will just skip these ones, until someone has an idea to fix it.

ggrimbert:
And now the error happens with another test...

I believe it's a timing error, but now I need some help to find the problem.

janmeier:
@cjbj Can you bring me up to speed on the licensing details? It seems that 12.2 is only available in enterprise edition currently? (at least when I try to build a docker image for it). I'm guessing we can't use that?

christianbundy:
@ggrimbert It looks like that error was also happening here: https://github.com/sequelize/sequelize/pull/7170#issuecomment-276336731

Could it be unrelated?

@janmeier I believe that's correct, the only available version is Oracle 12c, and the latest release seems to be 12.1. I don't think 12.2 EE is something we'll be able to test without a license.

ggrimbert:
Did someone had time to watch this ?

dmcghan:
@janmeier Why not just use 11.2 XE? It's free and will ensure backward compatibility to that version.

Also, it's much easier to install. :) Here's a good series on installing on CentOS in DigitalOcean:
http://ora-00001.blogspot.com/2015/06/installing-oracle-xe-ords-and-apex-on-centos-linux-part-zero.html

ggrimbert:
@dmcghan this PR is made for Oracle 12, not Oracle 11.

rpavez:
Any update on this?

ggrimbert:
No update about the merge request. But I'm planned to work for integrate with Oracle 11 too, so a new version will be done.

ggrimbert:
After some work, tests finally pass on travis.

Support for version 11 is for now delayed as it needs further investigations for dealing with offset.

ggrimbert:
New version, dealing with long alias names. Now all tests are OK, the build with node 4 fails due to errors with coverage : 
```Error: ENOENT: no such file or directory, rename 'coverage/lcov.info' -> 'coverage/integration.info'
```

janmeier:
Could you rebase this on top of latest master (there are a lot of changes to the docs folder that shouldn't be in here)

Any idea why the tests run so slow? I get that we probably can't make the database initialization much faster, but the actual tests still take _22 minutes_ to run (compared to 5-6 minutes for the other dialects)

ggrimbert:
Not really sure about how to do the rebase... just git rebase will do it ?

For tests, it takes about the same time on my computer. 
The problem comes from some functions called before each test. As the Db is destroyed / recreated, some slow queries are called and so it slows the entire process.

``` showTablesQuery (query-generator) ``` is the main problem, this request is huge, and I have no way to simplify it more than it is. This query returns all tables in the current schema ordered by number of dependencies.

```dropAllTables(override in query-interface)``` is the second one, Oracle doesn't support  dropping tables that have constraints / indexes at the same time, so I have to call each drop index / constraint by hand. (ie, normally a Promise.each (or .all) is done, but this has the effect of dropping FK and PK at the same time on different tables, and so we try to delete a PK referenced by an existing FK...).

I don't know if I'm clear, but I'm sure the slowness comes from there as in my app I have about 400 tests that run on sqlServer and Oracle and the time is the same for both.

**Comment by ggrimbert** [comment]
I've tried to rebase, but it takes each of my commit and compares it to the current version... After two hours it crashes...

Is there any other way of doing it ?

**Comment by janmeier** [comment]
* rebasing along the way to make sure you don't fall too far behind (too late now) 
* do an interactive rebase where you squash all commits into one so you don't have to apply the changes multiple times. 
* merge master 

Merging is fine in this case (we are not sadist). And I'm sorry you had to spend so long on it (on top of all the work you already did)

**Comment by ggrimbert** [comment]
The merge was done, but I think I should have done it bad previously, so I erased my doc folder by yours.

All tests keep passing for me.

With @cjbj and ConnorMcDonald we found a way to improve a little the drop tables, and we are trying to do the same with the showTablesQuery.

**Comment by ggrimbert** [comment]
I finally improved the queries so for now I'm avout 7 minutes on my computer for integration tests (about 10 on Travis).
The problem remains the mounting of the docker image (about 8-9 min).

For now my work seems valid, and we are using it internally.

**Comment by sceik** [comment]
Are there any updates?
I'm using this internally too and it works fine.
Let me know if there is any way i can help.

**Comment by ggrimbert** [comment]
any update ?

**Comment by agdolla** [comment]
Is there anything I can help to get this merged ? I can see some conflicts, can I fix it ?
We also need this.

**Comment by ggrimbert** [comment]
I've merged the last version of master.

With Oracle, I have the same problem as usual, node cannot rename the coverage file...

For MsSql, I don't understand. Can someone look at it ?

File: lib/dialects/oracle/query-generator.js [hunk 1]
```diff
@@ -0,0 +1,1268 @@
...
...SCOPE', 'SD_ALL', 'SD_INHIBIT', 'SD_SHOW', 'SEGMENT', 'SEG_BLOCK', 'SEG_FILE', 'SELECT', 'SEQUENCE', 'SERIALIZABLE', 'SESSION', 'SESSION_CACHED_CURSORS', 'SESSIONS_PER_USER', 'SET', 'SHARE', 'SHARED', 'SHARED_POOL', 'SHRINK', 'SIZE', 'SKIP', 'SKIP_UNUSABLE_INDEXES', 'SMALLINT', 'SNAPSHOT', 'SOME', 'SORT', 'SPECIFICATION', 'SPLIT', 'SQL_TRACE', 'STANDBY', 'START', 'STATEMENT_ID', 'STATISTICS', 'STOP', 'STORAGE', 'STORE', 'STRUCTURE', 'SUCCESSFUL', 'SWITCH', 'SYS_OP_ENFORCE_NOT_NULL$', 'SYS_OP_NTCIMG$', 'SYNONYM', 'SYSDATE', 'SYSDBA', 'SYSOPER', 'SYSTEM', 'TABLE', 'TABLES', 'TABLESPACE', 'TABLESPACE_NO', 'TABNO', 'TEMPORARY', 'THAN', 'THE', 'THEN', 'THREAD', 'TIMESTAMP', 'TIME', 'TO', 'TOPLEVEL', 'TRACE', 'TRACING', 'TRANSACTION', 'TRANSITIONAL', 'TRIGGER', 'TRIGGERS', 'TRUE', 'TRUNCATE', 'TX', 'TYPE', 'UB2', 'UBA', 'UID', 'UNARCHIVED', 'UNDO', 'UNION', 'UNIQUE', 'UNLIMITED', 'UNLOCK', 'UNRECOVERABLE', 'UNTIL', 'UNUSABLE', 'UNUSED', 'UPDATABLE', 'UPDATE', 'USAGE', 'USE', 'USER', 'USING', 'VALIDATE', 'VALIDATION', 'VALUE', 'VALUES', 'VARCHAR', 'VARCHAR2', 'VARYING', 'VIEW', 'WHEN', 'WHENEVER', 'WHERE', 'WITH', 'WITHOUT', 'WORK', 'WRITE', 'WRITEDOWN', 'WRITEUP', 'XID', 'YEAR', 'ZONE'];
```
eseliger: I would line-break this array, it's not very convenient to read this in github

lib/dialects/abstract/query-generator.js
```diff
@@ -101,7 +101,7 @@ const QueryGenerator = {
       emptyQuery += ' VALUES ()';
     }
 
-    if (this._dialect.supports.returnValues && options.returning) {
+    if (this._dialect.supports.returnValues && options.returning && this._dialect.name !== 'oracle') {
```
gvenzl: This looks odd to me. Why does it need another check on whether the dialect name is 'oracle'? Shouldn't `this._dialect.supports.returnValues` be false already?

Also, not fully understanding the context here but there is `RETURNING` support in Oracle too: http://docs.oracle.com/database/122/LNPLS/RETURNING-INTO-clause.htm#LNPLS01354
But that could be something else than wanted here.

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,197 @@
...
+        if (sequelize.config && 'dialectOptions' in sequelize.config) {
+          const dialectOptions = sequelize.config.dialectOptions;
+          if (dialectOptions && 'maxRows' in dialectOptions) {
+            this.lib.maxRows = sequelize.config.dialectOptions.maxRows;
+          }
+
+          if (dialectOptions && 'fetchAsString' in dialectOptions) {
+            this.lib.fetchAsString = sequelize.config.dialectOptions.fetchAsString;
+          } else {
+            this.lib.fetchAsString = [2006];
```
gvenzl: This [2006] looks like an arbitrary number. Can we specify a variable name instead or at least add a comment what this line means.

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,197 @@
...
+        });
+      }
+
+      return self.lib.getConnection(connectionConfig).then(connection => {
+        //TODO Oracle - connection pooling
+        //Not relevant, node-oracledb considers it if multiple connections are opened / closed; while testing, a few connections are created and closed.
+        //We change the session NLS_COMP and NLS_SORT to allow search case insensitive (http://stackoverflow.com/questions/5391069/case-insensitive-searching-in-oracle)
+        const alterSessionQry = "BEGIN EXECUTE IMMEDIATE 'ALTER SESSION SET NLS_COMP=LINGUISTIC'; EXECUTE IMMEDIATE 'ALTER SESSION SET NLS_SORT=BINARY_CI'; END;";
+        return connection.execute(alterSessionQry).then(() => {
+          return connection.commit().then(() => {
```
gvenzl: `ALTER SESSION` SQL statements don't require a `COMMIT` as it is not a DML operation.

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,197 @@
...
+        Object.keys(config.dialectOptions).forEach(key => {
+          connectionConfig[key] = config.dialectOptions[key];
+        });
+      }
+
+      return self.lib.getConnection(connectionConfig).then(connection => {
+        //TODO Oracle - connection pooling
+        //Not relevant, node-oracledb considers it if multiple connections are opened / closed; while testing, a few connections are created and closed.
+        //We change the session NLS_COMP and NLS_SORT to allow search case insensitive (http://stackoverflow.com/questions/5391069/case-insensitive-searching-in-oracle)
+        const alterSessionQry = "BEGIN EXECUTE IMMEDIATE 'ALTER SESSION SET NLS_COMP=LINGUISTIC'; EXECUTE IMMEDIATE 'ALTER SESSION SET NLS_SORT=BINARY_CI'; END;";
```
gvenzl: Why is the connection set to case insensitive search? Will that affect every connection? If so, why not leave it to the default which is what the developer is most likely to expect. If it is hidden in here it will only lead to confusion of the developer.

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,197 @@
...
+      connectionConfig.connectString = connectionConfig.database;
+
+      if (config.dialectOptions) {
+        Object.keys(config.dialectOptions).forEach(key => {
+          connectionConfig[key] = config.dialectOptions[key];
+        });
+      }
+
+      return self.lib.getConnection(connectionConfig).then(connection => {
+        //TODO Oracle - connection pooling
```
gvenzl: This TODO seems to be unnecessary based on the comment below.

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,197 @@
...
+          //We split to get the error number; it comes as ORA-XXXXX: 
+          let errorCode = err.message.split(':');
+          errorCode = errorCode[0];
+
+          if (errorCode) {
+            switch (errorCode) {
+              case 'ORA-28000': //Account locked
+                reject(new sequelizeErrors.ConnectionRefusedError(err));
+                break;
+              case 'ORA-01017':
```
gvenzl: Just for documentation purposes with the other error codes: `ORA-01017: invalid username/password; logon denied`

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,197 @@
...
+              case 'ORA-01017': 
+                reject(new sequelizeErrors.AccessDeniedError(err));
+                break;
+              case 'ORA-12154':
+                reject(new sequelizeErrors.HostNotReachableError(err)); //ORA-12154: TNS:could not resolve the connect identifier specified
+                break;
+              case 'ORA-12514' : // ORA-12514: TNS:listener does not currently know of service requested in connect descriptor
+              case 'ORA-12541' : //ORA-12541: TNS:No listener
+                //We can't send this kind of error, so we send 'login / password invalid'
+                err.message = 'ORA-01017 : invalid username/password; logon denied';
```
gvenzl: Why can't a `TNS: No listener` not be send back to the user? `Invalid username/password` is wrong and only leads to confusion.

lib/dialects/oracle/connection-manager.js
cjbj: It should be the constant`this.lib.CLOB`, if I'm not mistaken

docker-compose.yml
```diff
@@ -35,3 +35,11 @@ services:
     ports:
       - "127.0.0.1:8999:3306"
     container_name: mysql-57
+
+  oracle-12c:
+    image: konnecteam/oracle12c-database
```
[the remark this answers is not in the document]
ggrimbert: konnecteam/oracle-12c is a fork of sath89/oracle-12c with some modifications to make it ready for Sequelize tests.

lib/dialects/abstract/query-generator.js
```diff
@@ -291,8 +291,20 @@ const QueryGenerator = {
...
+        //This cannot be setted in where because rownum will be quoted
+        if (where && ((where.length && where.length > 0) || (Object.keys(where).length > 0))) {
+          //If we have a where clause, we add AND
+          query += ' AND ';
+        } else {
+          //No where clause, we add where
+          query += ' WHERE ';
+        }
+        query += `rownum < ${options.limit + 1}`;
+      } else {
```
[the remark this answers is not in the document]
ggrimbert: done :)

lib/dialects/abstract/query-generator.js
```diff
@@ -1642,12 +1656,25 @@ const QueryGenerator = {
             includeIgnoreAttributes: false
           }, topInclude.model);
         }
-        topLevelInfo.options.where['__' + throughAs] = this.sequelize.asIs([
+        /*topLevelInfo.options.where['__' + throughAs] = this.sequelize.asIs([
```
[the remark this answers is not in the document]
ggrimbert: Old code.. removed

.travis.yml
```diff
@@ -29,15 +29,41 @@ env:
...
     - SEQ_PG_PORT=8998
+    # oracle info
+    - NODE_ORACLEDB_HOST: localhost
+
+before_install:
+  - |
+    if [ $ORACLE_VER ]; then 
+      mkdir -p /opt/oracle && 
+      cd /opt/oracle && 
+      wget https://s3.amazonaws.com/sequelize/instantclient-basic-linux.x64-12.1.0.2.0.zip &&
```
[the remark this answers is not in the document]
ggrimbert: I have no idea what the new version adds; I can't find any changelog. But for my tests and dev, I'm with 12.1

docs/usage.md
```diff
@@ -223,6 +223,19 @@ const sequelize = new Sequelize('database', 'username', 'password', {
...
+
+The library for Oracle is`oracledb@^1.11.0` You'll just need to define the dialect:
+
+```js
+const sequelize = new Sequelize('database', 'username', 'password', {
+  dialect: 'oracle'
+})
+```
+
+As Oracle limits the length of alias names to 30 characters, some requests are rebuilt.
```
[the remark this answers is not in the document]
ggrimbert: Oracle 12.1 still has 30 char length limitations , so this part is kept for this version; but there's an if to check the version and rebuilt the request if necessary

lib/dialects/abstract/query-generator.js
ggrimbert: needs investigation

lib/dialects/oracle/connection-manager.js
ggrimbert: true, modified

lib/dialects/oracle/connection-manager.js
ggrimbert: This TODO has been removed; It stayed for review purpose, as connection poolnig is done for some dialects, it's an explanation why I haven't done it.

lib/dialects/oracle/connection-manager.js
ggrimbert: I thought it's a necessary option; in my company's case, I need this. If you think it should be removed, I'll do it and patch for my company.

lib/dialects/oracle/connection-manager.js
ggrimbert: It's for "user-friendly" purpose; a user does not have to know the entire error; 
It's a choice we have made during the development of our application; someone doesn't have to know that the server configuration isn't valid, so the error is wrapped into a bad login
If you think that all errors should be returned as the are throwed, i can modify to this.

.travis.yml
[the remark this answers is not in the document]
cjbj: Instant Client is a packing of client libraries covering a number of products, so the doc is in a number of places.  For node-oracledb one change is the improvement in "High Availability" due to session pool enhancements.  I'd recommend using 12.2 client.  It will connect to older DBs too, of course.

lib/dialects/oracle/connection-manager.js
cjbj: Not knowing the exact Oracle error (and worse - giving what seems to be a logical Oracle error about the username/password) is really going to be frustrating for people who hit this.  They will end up trying a lot of passwords and then getting frustrated with 'Oracle'.

docs/usage.md
```diff
@@ -223,6 +223,19 @@ const sequelize = new Sequelize('database', 'username', 'password', {
 })
 ```
 
+### ORACLE
+
+The library for Oracle is`oracledb@^1.11.0` You'll just need to define the dialect:
```
cjbj: It would be better to use 1.13.1

lib/dialects/oracle/connection-manager.js
gvenzl: Agreed! The correct error has to be returned otherwise the user/developer will be sent into the wrong directions which will only end up with a lot of frustration.

lib/dialects/oracle/connection-manager.js
gvenzl: Yes, this has to be removed. Oracle by default does a case sensitive search which means that this can produce wrong results when a user expects an exact match but gets multiple rows back due to case insensitivity.

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,397 @@
...
+  };
+
+  function TEXT(length) {
+    if (!(this instanceof TEXT)) return new TEXT(length);
+    BaseTypes.TEXT.apply(this, arguments);
+  }
+  inherits(TEXT, BaseTypes.TEXT);
+
+  TEXT.prototype.toSql = function toSql() {
+    //TEXT is not support by Oracle, passing through NVARCHAR
```
cjbj: NVARCHAR2 is a specialized type with a particular character set; I don't know what TEXT is but from the name wouldn't it map to VARCHAR2?

Also, there is no support for NVARCHAR2 in node-oracledb v1.  It may work in some cases where the DB charset width is big enough to hold the DB national character set.  (Node-oracledb v2 has query support for NVARCHAR2)

.travis.yml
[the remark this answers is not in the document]
ggrimbert: I don't have access to this server, can someone that has access can update this dep ?

dmcghan:
@ggrimbert Back on May 30th, you said:

> Oracle doesn't support dropping tables that have constraints / indexes at the same time

Did you try 'cascade constraints'? Here's an example:

```sql
create table p (
  id   number primary key,
  name varchar2(50)
)
/

insert into p (id, name) values (1, 'parent 1');
insert into p (id, name) values (2, 'parent 2');
insert into p (id, name) values (3, 'parent 3');

create table c (
  id   number primary key,
  p_id number not null,
  name varchar2(50),
  constraint c_p_fk foreign key (p_id) references p (id)
)
/

insert into c (id, p_id, name) values (1, 1, 'child 1');
insert into c (id, p_id, name) values (2, 1, 'child 2');
insert into c (id, p_id, name) values (3, 2, 'child 3');

-- fails
drop table p;

-- succeeds
drop table p cascade constraints;
```

docs/usage.md
ggrimbert: the ^ in front of the version number indicates version from 1.11.0 to latest. In package.json it's the same, so the last version will be downloaded

lib/dialects/oracle/data-types.js
ggrimbert: TEXT is an extended varchar type in mssql (https://docs.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql) except if I understood wrong.

That's why I've decided to map it to NVARCHAR2. But if no length is passed, I set it to CLOB

ggrimbert:
This problem has been solved. 
In fact it was the dropping used by Sequelize; multiple drops were called at the same time, and it fails sometimes due to constraints.
This has been solved by using recursivity for the "dropAllTables" method.

dmcghan:
@ggrimbert Are you unable to add the drop constraints clause? It seems the issue was related to how long it takes. Dropping tables correctly, without recursion, should be faster.

dmcghan:
@ggrimbert I can't figure out how to comment on a line you didn't change...

On line 38 of docs/getting-started.md:
```js
dialect: 'mysql'|'sqlite'|'postgres'|'mssql',
```

Could you add 'oracle' to the list of dialects?

docs/index.md
```diff
@@ -9,7 +9,7 @@
 [![npm](https://img.shields.io/npm/dm/sequelize.svg?style=flat-square)](https://npmjs.org/package/sequelize)
 [![npm](https://img.shields.io/npm/v/sequelize.svg?style=flat-square)](https://github.com/sequelize/sequelize/releases)
 
-Sequelize is a promise-based ORM for Node.js v4 and up. It supports the dialects PostgreSQL, MySQL, SQLite and MSSQL and features solid transaction support, relations, read replication and
+Sequelize is a promise-based ORM for Node.js v4 and up. It supports the dialects PostgreSQL, MySQL, SQLite MSSQL and Oracle and features solid transaction support, relations, read replication and
```
dmcghan: Need a comma after 'SQLite'.

ggrimbert:
@dmcghan  The problem for dropping table is in the asynchronous methods. Multiple drop table are called in the same time and, sometimes it fails.

I don't clearly remember the error, but I tried dropping with cascade and it didn't work.

.travis.yml
```diff
@@ -29,15 +29,41 @@ env:
...
+      wget https://s3.amazonaws.com/sequelize/instantclient-sdk-linux.x64-12.1.0.2.0.zip &&
```
sushantdhiman: who owns this `https://s3.amazonaws.com/sequelize/` ? Dont we have any official channel ?

dmcghan:
@ggrimbert 

One line 55 of docs/usage.md, could you add 'oracle' there too?
```javascript
// - currently supported: 'mysql', 'sqlite', 'postgres', 'mssql'
```

Also, if you support `dialectOptions`, please add 'oracle' to line 60 too.
```javascript
// - currently supported: 'mysql', 'postgres', 'mssql'
```

.travis.yml
```diff
@@ -29,15 +29,41 @@ env:
...
+      unzip instantclient-sdk-linux.x64-12.1.0.2.0.zip &&
+      mv instantclient_12_1 instantclient &&
+      cd instantclient && 
+      ln -s libclntsh.so.12.1 libclntsh.so &&
+      export LD_LIBRARY_PATH=/opt/oracle/instantclient:$LD_LIBRARY_PATH &&
+      cd $TRAVIS_BUILD_DIR;
+    fi
 
 before_script:
   # mount ramdisk
```
sushantdhiman: Move comment to command where actual mounting happens

changelog.md
```diff
@@ -0,0 +1,1253 @@
+# 4.0.0 (final)
```
sushantdhiman: Remove this file

docs/usage.md
```diff
@@ -223,6 +223,19 @@ const sequelize = new Sequelize('database', 'username', 'password', {
 })
 ```
 
+### ORACLE
```
dmcghan: Seems the other database names are not uppercased. `ORACLE` should be `Oracle` or `Oracle Database`.

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,168 @@
...
+      host:  config.host,
+      port : config.port,
+      database : config.database,
+      password: config.password,
+      externalAuth : config.externalAuth,
+      stmtCacheSize : 0
+    };
+
+    //We check if there dialect options
+    if ('dialectOptions' in config) {
```
sushantdhiman: just use `config.dialectOptions`, same check exists about 10 lines below, good idea to cover both of them under same if check

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,168 @@
...
+    connectionConfig.connectString = connectionConfig.database;
+
+    if (config.dialectOptions) {
+      Object.keys(config.dialectOptions).forEach(key => {
+        connectionConfig[key] = config.dialectOptions[key];
+      });
+    }
+
+    return this.lib.getConnection(connectionConfig).then(connection => {
+      return connection;
```
sushantdhiman: :). then looks kinda unnecessary

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,168 @@
...
+    }).catch(err => {
+      throw new sequelizeErrors.ConnectionError(err);
+    });
+  }
+
+  validate(connection) {
+    return connection && ['disconnected', 'protocol_error'].indexOf(connection.state) === -1;
+  }
+
+
```
sushantdhiman: Extra line

lib/dialects/abstract/query-generator.js
```diff
@@ -291,8 +291,18 @@ const QueryGenerator = {
...
+      if (this.dialect === 'oracle') {
+        //This cannot be setted in where because rownum will be quoted
+        if (where && ((where.length && where.length > 0) || (Object.keys(where).length > 0))) {
+          //If we have a where clause, we add AND
+          query += ' AND ';
+        } else {
+          //No where clause, we add where
+          query += ' WHERE ';
+        }
+        query += `rownum < ${options.limit + 1}`;
```
dmcghan: `options.limit` is not escaped, potential SQL injection vulnerability?

I recommend using `<=` over `<` to avoid the arithmetic and add self.escape (need to make sure it doesn't turn a number to a string). 

Also, the example below adds a space on the end (not sure if it's needed).

Maybe something like:

```javascript
query += `rownum <= this.escape(options.limit) `;
```

.travis.yml
christianbundy: I'm not sure who owns it, but I _do_ know that Oracle doesn't let you download any of their Instant Client utilities. The only semi-trustless ways to do this, as far as I understand:
- Add the above files to the repository (gross!).
- Use a torrent or some other content-addressable store.

lib/dialects/abstract/query-generator.js
```diff
@@ -1642,12 +1654,19 @@ const QueryGenerator = {
             includeIgnoreAttributes: false
           }, topInclude.model);
         }
-        topLevelInfo.options.where['__' + throughAs] = this.sequelize.asIs([
+
+        const queryWhere = this.sequelize.asIs([
```
dmcghan: Could some comments be added to explain why this change was needed for Oracle. I don't doubt it, it's just hard to understand at first glance.

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,168 @@
...
+    return this.lib.getConnection(connectionConfig).then(connection => {
+      return connection;
+    })
+      .catch(err => {
+        //We split to get the error number; it comes as ORA-XXXXX: 
+        let errorCode = err.message.split(':');
+        errorCode = errorCode[0];
+
+        if (errorCode) {
```
dmcghan: The `if` block seems unnecessary as the `switch` will take care of `undefined`.

.travis.yml
sushantdhiman: http://www.oracle.com/technetwork/topics/linuxx86-64soft-092277.html ?

sushantdhiman: nvm, this needs login. But I know for sure I cant let this package merged in current state. May be we can manage this an one of repo, under sequelize ? I will look into it.

christianbundy: Sorry, I should've been more clear -- Oracle doesn't let you download any of their Instant Client utilities non-interactively (i.e. with `wget` or `curl`). Clicking those links forces an interactive license agreement and redirects you to a login page. There used to be ways around it, but I wasn't able to find a workable solution to this problem without just storing the downloads in Git.

lib/dialects/oracle/connection-manager.js
```diff
@@ -0,0 +1,168 @@
...
+          }
+        } else {
+          throw new sequelizeErrors.ConnectionError(err);
+        }
+
+        throw err; //Unknown error, we throw it anyway
+      });
+  }
+
+  disconnect(connection) {
```
dmcghan: This doesn't look right. `disconnect` seems to want a promise that's resolved or rejected. The current implementation throws on the main thread. 

I suggest requiring in Utils at the top:
```javascript
const Utils = require('../../utils');
```

Then update disconnect to something like this:
```javascript
disconnect(connection) {
  return new Utils.Promise((resolve, reject) => {
    connection.release()
      .then(resolve)
      .catch(err => {
        reject(new sequelizeErrors.ConnectionError(err));
      });
  });
}
```

>

lib/dialects/oracle/connection-manager.js
dmcghan: Agree, that should be removed. 

Also, the promise used here is generated the underlying lib, which defaults to the global Promise. If Sequelize expects all promises to be based on Utils.Promise, then this should be updated.

sushantdhiman: Does `oracledb` library allow overriding what Promise instance is used ? this is essential to support cls support. Otherwise we need to manually override it everywhere we get result from library

dmcghan: @sushantdhiman Yes, just [set oracledb.Promise](https://github.com/oracle/node-oracledb/blob/master/doc/api.md#-191-custom-promise-libraries) to the promise implementation you want to use.

Not sure where exactly this should be done in Sequelize.

sushantdhiman: In connection-manager (this file's) constructor when setting `this.lib`, Also just to be safe, wrap `connect / disconnect` method with `Promise.try / new Promise` or whatever make sense.

Also wrap `connection.execute or _run` method in this Promise something like this https://github.com/sequelize/sequelize/blob/63978e184a95a8a39595cb0e04d43e663163c9bc/lib/dialects/mssql/query.js#L61

dmcghan: @sushantdhiman Shouldn't that be a new Utils.Promise?

sushantdhiman: @dmcghan It is, its imported as name `Promise`

docs/usage.md
cjbj: If the latest is always used, then you may as well update the text to 1.13.1!  Various bugs were fixed, LOB support was improved in versions after 1.11.0, and pool pinging were introduced.  I'd like to see everyone update to 1.13.1 :)

.travis.yml
ggrimbert: After the PR was created, @janmeier  created (thanks to him) the build for Oracle. So I assume he has the right to update the package.
@christianbundy The link is not an Oracle link, it's aws; this means that the file has been downloaded once and put on another server where we have access.

lib/dialects/abstract/query-generator.js
ggrimbert: This part is not Oracle specific. It's a fix I've done for all versions but as I was working on Oracle on master branch, the fix has been done here.
The fix for V3 is here https://github.com/sequelize/sequelize/pull/7436 and explains the reason this code exists.

I know it's not a good idea to fix some bugs while doing a feature, but I had no time to make a special PR for master with this.

.travis.yml
janmeier: I own that bucket :)

There is https://github.com/cbandy/travis-oracle, but as I remember, it was pretty slow and would not always work when i tested it (it uses a browser to log in and accept the terms)

.travis.yml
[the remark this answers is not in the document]
janmeier: Yes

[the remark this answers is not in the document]
janmeier: https://s3.amazonaws.com/sequelize/instantclient-basic-linux.x64-12.2.0.1.0.zip
https://s3.amazonaws.com/sequelize/instantclient-sdk-linux.x64-12.2.0.1.0.zip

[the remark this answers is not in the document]
ggrimbert: I've changed the paths, thanks :)

.travis.yml
```diff
@@ -29,15 +29,41 @@ env:
...
+  - |
+    if [ $ORACLE_VER ]; then
+    wget https://s3.amazonaws.com/sequelize/data.tar.gz
```
janmeier: I would prefer if this was not hosted on my amazon account once this is merged - Since it's 2.6 gigabytes, so it's going to rack up a bit of transfer charges.

The client and SDK are fine, but this file is a bit too big ;)

eseliger: Maybe we could simply push an image to docker hub, its free and maybe travis caches images so it chould be also faster in the future 🤔

janmeier: That would be ideal - However the normal image requires a long (!!) time to initialize the db (+1 minute) - That's what we're trying to avoid by having the DB pre-initialized.

I'm guessing that could be integrated into a docker image, I just couldn't figure out how

eseliger: Oh, didn't know that, thats long! 
If someone needs help with the docker thing, just ping me, I could try that

docker-compose.yml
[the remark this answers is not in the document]
eseliger: Ok didn't know that :)

docs/usage.md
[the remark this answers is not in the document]
eseliger: I'd just add a little mark this is just for  <= 12.1 , because thats not the behaviour from 12.2 on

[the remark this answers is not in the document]
ggrimbert: You're right, I will add it.

.travis.yml
ggrimbert: I've tried many things, but to push on docker hub a fully initialized image, but without success.
Fell free to test, any help on this is really appreciate :)
The repository is here https://github.com/konnecteam/docker-oracle-12c

eseliger: https://github.com/konnecteam/docker-oracle-12c/pull/1 there you go :)

lib/dialects/oracle/data-types.js
cjbj: What do you map NTEXT to?

.directory
```diff
@@ -0,0 +1,6 @@
+[Dolphin]
```
eseliger: what is this file for? 🙈

docker-compose.yml
```diff
@@ -39,3 +39,11 @@ services:
...
+    volumes:
+      - ./oracle-data:/u01/app/oracle
```
eseliger: this mapping can be removed when you decide to use the version I proposed to your dockerfile :)

telkomops:
What is the planned date for release? Is it in the Roadmap ? TIA

sushantdhiman:
@telkomops Please don't spam this PR with non constructive comments, it will be release when its ready

tgabi333:
@telkomops we did the following way: forked this repo and merged this branch and rebased to master to have oracle support in our product.

ggrimbert:
No space left on device... Someone could check ?

sushantdhiman:
@ggrimbert Its then ramdisk setup, remove these lines

https://github.com/sequelize/sequelize/blob/0ca7299f7ef3f849820b439c89b389747d5d3e15/docker-compose.yml#L24

https://github.com/sequelize/sequelize/blob/0ca7299f7ef3f849820b439c89b389747d5d3e15/docker-compose.yml#L38

ggrimbert:
It will not impact the other builds ?

sushantdhiman:
Slow them a bit, nothing more

ggrimbert:
Well, still failing due to missing space on storage disk...

ggrimbert:
Update to the last master.

Still having a problem of segmentation fault with node 4, don't know why.

I had to set back the sleep in Oracle container's start; without this, the tests are launched too soon and the db is mounted, but not accessible at this moment

tgabi333:
Please take a look at our branch: https://github.com/TechTeamer/sequelize/commits/master

shellthor:
It looks like coverage is failing due to Travis CI failing on postgres dialect, nodejs 4.x.  Codecov will continue to report as a failure until a Travis CI reports success, no matter what the actual code coverage delta is.

wanghaisheng:
anything update

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,392 @@
...
+  BaseTypes.DATEONLY.types.oracle = ['DATE', 'DATEONLY'];
+  BaseTypes.BOOLEAN.types.oracle = ['NUMBER'];
+  BaseTypes.BLOB.types.oracle = ['BLOB'];
+  BaseTypes.DECIMAL.types.oracle = ['DECIMAL'];
+  BaseTypes.UUID.types.oracle = false;
+  BaseTypes.ENUM.types.oracle = false;
+  BaseTypes.REAL.types.oracle = false;
+  BaseTypes.NUMERIC.types.oracle = false;
+  BaseTypes.DOUBLE.types.oracle = false;
+  BaseTypes.GEOMETRY.types.oracle = false;
```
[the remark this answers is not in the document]
cjbj: Node-oracledb v1 doesn't support mapping of object types like SDO_GEOMETRY.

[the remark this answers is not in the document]
gvenzl: Is this coming anytime soon?

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,392 @@
+'use strict';
+
+const _ = require('lodash');
+const inherits = require('../../utils/inherits');
+const moment = require('moment');
+const momentTz = require('moment-timezone');
+
+module.exports = BaseTypes => {
+  const warn = BaseTypes.ABSTRACT.warn.bind(undefined, 'http://docs.oracle.com/cd/A87860_01/doc/server.817/a76965/c10datyp.htm');
```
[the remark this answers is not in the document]
ggrimbert: thanks, I hadn't pay attention to the version...

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,392 @@
...
+
+  BaseTypes.DATE.types.oracle = ['TIMESTAMP', 'TIMESTAMP WITH LOCAL TIME ZONE'];
```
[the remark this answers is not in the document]
ggrimbert: I've chosen to use DATE as DATE only column and use timestamps as date with hours.

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,392 @@
...
+  BaseTypes.STRING.types.oracle = ['VARCHAR2', 'NVARCHAR2'];
+  BaseTypes.CHAR.types.oracle = ['CHAR', 'RAW'];
+  BaseTypes.TEXT.types.oracle = false;
```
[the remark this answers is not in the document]
ggrimbert: I don't remember; but I guess it's the 2nd solution.

I think I remember that passing this to ['CLOB'] causes some problems.

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,392 @@
...
+  BaseTypes.INTEGER.types.oracle = ['INTEGER'];
+  BaseTypes.BIGINT.types.oracle = false;
+  BaseTypes.FLOAT.types.oracle = false;
```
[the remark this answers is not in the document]
ggrimbert: It's a design problem with my company's.
We don't use FLOAT or DOUBLE but NUMBER with different precisions; so I can change this but really cannot test really.

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,392 @@
...
+  BaseTypes.TIME.types.oracle = ['TIMESTAMP WITH LOCAL TIME ZONE'];
+  BaseTypes.DATEONLY.types.oracle = ['DATE', 'DATEONLY'];
```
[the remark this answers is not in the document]
ggrimbert: it's a "trick"; this type will not be used for creating a DATEONLY column, but is needed at some pieces of code in order to work correctly.

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,392 @@
...
+  BaseTypes.BOOLEAN.types.oracle = ['NUMBER'];
+  BaseTypes.BLOB.types.oracle = ['BLOB'];
+  BaseTypes.DECIMAL.types.oracle = ['DECIMAL'];
+  BaseTypes.UUID.types.oracle = false;
+  BaseTypes.ENUM.types.oracle = false;
+  BaseTypes.REAL.types.oracle = false;
```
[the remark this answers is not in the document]
ggrimbert: As I said upward, as we don't use it in my company, I can't really test this; so I didn't implemented it.

Is it a classic type without any cast or anything ? If yes, I can implement this.

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,392 @@
...
+  BaseTypes.NUMERIC.types.oracle = false;
```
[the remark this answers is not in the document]
ggrimbert: same as upward

lib/dialects/oracle/data-types.js
```diff
@@ -0,0 +1,392 @@
...
+  BaseTypes.DOUBLE.types.oracle = false;
```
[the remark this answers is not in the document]
ggrimbert: same as upward

lib/dialects/oracle/query-generator.js
```diff
@@ -0,0 +1,1247 @@
...
+      '  SELECT COUNT(1) INTO V_COUNT FROM ALL_USERS WHERE USERNAME = ', wrapSingleQuote(schema), ';',
+      '  IF V_COUNT = 0 THEN',
+      '    EXECUTE IMMEDIATE ', wrapSingleQuote('CREATE USER ' + schema + ' IDENTIFIED BY 12345 DEFAULT TABLESPACE USERS'), ';',
+      '    EXECUTE IMMEDIATE ', wrapSingleQuote('GRANT "CONNECT" TO ' + schema), ';',
+      '    EXECUTE IMMEDIATE ', wrapSingleQuote('GRANT create table TO ' + schema), ';',
+      '    EXECUTE IMMEDIATE ', wrapSingleQuote('GRANT create view TO ' + schema), ';',
+      '    EXECUTE IMMEDIATE ', wrapSingleQuote('GRANT create any trigger TO ' + schema), ';',
+      '    EXECUTE IMMEDIATE ', wrapSingleQuote('GRANT create any procedure TO ' + schema), ';',
+      '    EXECUTE IMMEDIATE ', wrapSingleQuote('GRANT create sequence TO ' + schema), ';',
+      '    EXECUTE IMMEDIATE ', wrapSingleQuote('GRANT create synonym TO ' + schema), ';',
```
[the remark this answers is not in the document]
ggrimbert: Without the ANY keyword, I was unable to create some triggers inside my schema. 
Doesn't ```RESOURCE``` gives more rights ? I just want these ones, no more.

lib/dialects/oracle/query.js
```diff
@@ -0,0 +1,833 @@
...
+    return new sequelizeErrors.DatabaseError(err);
+  };
+
+
+  isShowIndexesQuery() {
+    return this.sql.indexOf('SELECT i.index_name,i.table_name, i.column_name, u.uniqueness') > - 1;
+  };
+
+  isSelectCountQuery() {
+    return this.sql.toUpperCase().indexOf('SELECT COUNT(*)') > - 1;
```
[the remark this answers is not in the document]
ggrimbert: Right, thanks

test/integration/associations/belongs-to-many.test.js
```diff
@@ -139,13 +139,17 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), () => {
       });
     });
 
-    it('only gets objects that fulfill options with a formatted value', function() {
-      return this.User.find({where: {username: 'John'}}).then(john => {
-        return john.getTasks({where: {active: true}});
-      }).then(tasks => {
-        expect(tasks).to.have.length(1);
+
+    //Oracle doesn't support column names non quoted by "
```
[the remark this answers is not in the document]
ggrimbert: It's an old thing; the test doesn't pass as ```active=true``` was passed as this in the query without any reformating, and so the test fails.
Now, it seems that the test is OK; so I remove the comment and the if around.

test/unit/sql/data-types.test.js
```diff
@@ -852,162 +886,188 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
...
+        oracle: 'REAL',
         mssql: 'REAL'
       });
     });
 
     suite('DOUBLE PRECISION', () => {
       testsql('DOUBLE', DataTypes.DOUBLE, {
-        default: 'DOUBLE PRECISION'
+        default: 'DOUBLE PRECISION',
+        oracle: 'NUMBER(15,5)'
```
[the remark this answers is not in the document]
ggrimbert: Related to upward, when you asked about the data-types.

test/unit/sql/data-types.test.js
```diff
@@ -852,162 +886,188 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
...
       });
 
       testsql('DOUBLE.UNSIGNED', DataTypes.DOUBLE.UNSIGNED, {
         default: 'DOUBLE PRECISION UNSIGNED',
+        oracle: 'NUMBER(15,5)',
```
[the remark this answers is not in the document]
ggrimbert: Related to upward, when you asked about the data-types.

test/unit/sql/data-types.test.js
```diff
@@ -852,162 +886,188 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
...
         postgres: 'DOUBLE PRECISION'
       });
 
       testsql('DOUBLE(11)', DataTypes.DOUBLE(11), {
         default: 'DOUBLE PRECISION(11)',
+        oracle: 'NUMBER(15,5)',
```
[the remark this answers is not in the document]
ggrimbert: Related to upward, when you asked about the data-types.

test/unit/sql/data-types.test.js
```diff
@@ -852,162 +886,188 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
...
         postgres: 'DOUBLE PRECISION'
       });
 
       testsql('DOUBLE(11).UNSIGNED', DataTypes.DOUBLE(11).UNSIGNED, {
         default: 'DOUBLE PRECISION(11) UNSIGNED',
         sqlite: 'DOUBLE PRECISION UNSIGNED(11)',
+        oracle: 'NUMBER(15,5)',
```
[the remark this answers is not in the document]
ggrimbert: Related to upward, when you asked about the data-types.

test/unit/sql/data-types.test.js
```diff
@@ -852,162 +886,188 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
...
         postgres: 'DOUBLE PRECISION'
       });
 
       testsql('DOUBLE({ length: 11 }).UNSIGNED', DataTypes.DOUBLE({ length: 11 }).UNSIGNED, {
         default: 'DOUBLE PRECISION(11) UNSIGNED',
         sqlite: 'DOUBLE PRECISION UNSIGNED(11)',
+        oracle: 'NUMBER(15,5)',
```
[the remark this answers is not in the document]
ggrimbert: Related to upward, when you asked about the data-types.

test/unit/sql/data-types.test.js
```diff
@@ -852,162 +886,188 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
...
         postgres: 'DOUBLE PRECISION'
       });
 
       testsql('DOUBLE(11).UNSIGNED.ZEROFILL', DataTypes.DOUBLE(11).UNSIGNED.ZEROFILL, {
         default: 'DOUBLE PRECISION(11) UNSIGNED ZEROFILL',
         sqlite: 'DOUBLE PRECISION UNSIGNED ZEROFILL(11)',
+        oracle: 'NUMBER(15,5)',
```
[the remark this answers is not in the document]
ggrimbert: Related to upward, when you asked about the data-types.

test/unit/sql/data-types.test.js
```diff
@@ -1129,47 +1202,57 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
 
     if (current.dialect.supports.NUMERIC) {
       testsql('NUMERIC', DataTypes.NUMERIC, {
-        default: 'DECIMAL'
+        default: 'DECIMAL',
+        oracle : 'NUMBER'
```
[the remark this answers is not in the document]
ggrimbert: Related to upward, when you asked about the data-types.

lib/dialects/oracle/data-types.js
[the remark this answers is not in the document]
gvenzl: Can you confirm for what BaseTypes.STRING is being used? Is there a different type for unlimited strings?

lib/dialects/oracle/data-types.js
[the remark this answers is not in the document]
gvenzl: Makes sense and also internally Oracle just creates a `NUMBER` datatype with the right precision. The question though is: What does it mean if this is set to false? Does that mean that Sequelize won't work with float types or does it not matter anyway as JS doesn't give you specific data types? What's the impact of having those set to `false`?

lib/dialects/oracle/data-types.js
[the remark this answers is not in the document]
gvenzl: Yes, it's mapped internally:

```
SQL> create table test (n1 float, n2 decimal, n3 numeric, n4 real);

Table created.

SQL> desc test;
 Name					   Null?    Type
 ----------------------------------------- -------- ----------------------------
 N1						    FLOAT(126)
 N2						    NUMBER(38)
 N3						    NUMBER(38)
 N4						    FLOAT(63)
```

But before you change that, let me first understand what the implications are of not having those mapped, as said above.

lib/dialects/oracle/query-generator.js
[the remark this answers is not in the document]
gvenzl: That should not be the case. `GRANT CREATE TRIGGER` allows you to create triggers within your schema. Do you perhaps create a system trigger, such as a logon trigger or similar?

The `ANY` keyword will allow the user to create a trigger and any function in any schema, which is a security violation.

The `RESOURCE` role gives a schema the grants for any resource that schema may want. The important part though is that the resources are all self-contained within the schema.

If you just add those grants, what will happen to other users that want to create other resources via Sequelize? Will they be able to?

frangarcj:
What's left for this to be merged? Can I help? Thanks

lib/dialects/oracle/data-types.js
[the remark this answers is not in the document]
ggrimbert: for string I use "NVARCHAR2", and if the parameters binary is set to true, I use RAW.

If the length specified is over 4000, I display a message saying that the MAX_STRING_USE has to be modified.

lib/dialects/oracle/data-types.js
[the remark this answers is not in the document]
ggrimbert: This doesn't mean that the type isn't supported; it is supported.

As the following test shows : 
```js
//Creation of a table with a Float column
Price = this.sequelize.define('Price', {
          value: DataTypes.FLOAT
        }),
```
which creates : 
```sql
describe prices

Nom       NULL ?   Type                              
--------- -------- --------------------------------- 
ID        NOT NULL NUMBER(38)                        
value              FLOAT(126)                        
CREATEDAT NOT NULL TIMESTAMP(6) WITH LOCAL TIME ZONE 
UPDATEDAT NOT NULL TIMESTAMP(6) WITH LOCAL TIME ZONE 
PRODUCTID          NUMBER(38)   
```

So, It seems I clearly don't remember why I set 
```js
BaseTypes.FLOAT.types.oracle = false;
```

I guess it's just a question of not overridden internally the managing of this type. Maybe @janmeier  can help ?

knoxcard:
Just throwing this out there, it was mentioned that adding oracle could possibly "slow down things a bit"?
wouldn't it be best to have something like...

npm install sequelize
then you would set your dialect: 'oracle'

--- then the optional sequelize-oracle would be required as an add on
npm install sequelize-oracle

dmcghan:
@knoxcard Adding support for any database could "slow down things a bit", no? What are the criteria for where and how the line should be drawn?

ggrimbert:
@knoxcard I guess you misunderstood the specific time and space this sentence was used.
It was about the builds done on the server; not the code itselfs. 
This is just about setting up the ramdisk for Oracle that slows down the execution of other builds; in usage condition there is no impact

gvenzl:
On that note, same question as @frangarcj: What's outstanding to get this merged? Where can we help?

knoxcard:
@dmcghan - Sequelize is a one fits all ORM solution for MariaDB, MySQL, PostgreSQL, SQLite, and hopefully soon to be Oracle...

Currently I employ MariaDB.  What would be the point of downloading all the other overhead code that is not related to MariaDB?  Unnecessary hard disk space, but that is super minor, hard disk is cheap and the extra space is most likely negligible.  Once a dialect (DB) is chosen, I am assuming Sequelize only "requires" that particular code pertaining to the dialect?

Sorry guys, I really need to just dig into the source code, which I should have done before posting.

ggrimbert:
All database dependencies are devDependencies, so you only have the dependencies you need. If you work with mariaDB, it installs mysql (as it's the same connector and the specific mariaDB code was merged into MySQL). The code for other dialects is read, but not instantiated.
And also the "dialect" folder that contains the code for each dialect is "only" 494.6 KB...

I abandoned the idea of seeing this merged. I dreamed about but the owners do not seem to be following this. Also, we saw that many issues are still open without any answer or correction, so we have decided to work on our side. [konnecteam fork](https://github.com/konnecteam/sequelize)

The develop branch is the one we keep up to date; each important commit (i.e. when we saw a bug) leads to a new tag named 'v-XXX-konnect' and is what we use in our apps. We have corrected many things about big queries with multiple joins, optimized some parts, added tests...

Feel free to use it and we also accept issues (by mail as a fork cannot have its own issues).
We are also migrating the project to typescript to improve it's readability and performance; we've planned to create a new repository when it's done (about 3-4 months).

wanghaisheng:
anything update

knoxcard:
https://github.com/adeo-proxideco/sequelize-oracle, maybe some are unaware of this repository, but it's old. :-(

ezequielzacca:
what is the current status of this?

cmcinroy:
Can anyone comment about the current state of sequelize + oracle?
Is the [dev branch](https://github.com/konnecteam/sequelize/tree/develop) of the konnecteam fork the best bet?
It seems to be more complete than [cu8sequelize](https://github.com/nhuanhoangduc/cu8-sequelize-oracle) -- which was forked from sequelize at ~v1.5 -- or any of the sundry other forks, for that matter.

I am tempted to follow @tgabi333's approach (fork-merge-rebase)... unless @ggrimbert has an alternate suggestion.

Thanks.

ggrimbert:
We are currently working on https://github.com/konnecteam/ts-sequelize/tree/develop
This is a fork of our develop branch but migrated to TypeScript.
We are actually working on merging V5 into this repository.

We work with this project for our developments and don't have any problems.
For now it's not in its final form, not still integrated to npm and some changes have to be made.

So if you need Oracle you can use https://github.com/konnecteam/ts-sequelize/tree/develop or the develop branch on https://github.com/konnecteam/sequelize/tree/develop; which is what we use in production

cmcinroy:
That is excellent info, @ggrimbert !  Thanks so much for your response.

ezequielzacca:
what happened with this feature? is not going to get implemented?

ggrimbert:
For this to be merged it needs someone to review this.
As it's opened since 2 years and no one seems to be interested in making this merged, we've decided to use our version of Sequelize.

There are two versions : 
 - [Classic version](https://github.com/konnecteam/sequelize ) which is a "better" version of Sequelize, that we use in our projects
- [Typescript version](https://github.com/konnecteam/ts-sequelize) which is our version in Typescript with some improvements; currently in test

papb:
Hello @ggrimbert, thank you so much for this. Sorry to take so long.

Hopefully we can bring this forward soon. @sushantdhiman @SimonSchick

.travis.yml
```diff
@@ -29,22 +29,51 @@ env:
...
     - SEQ_PG_PORT=8998
+    # oracle info
+    - NODE_ORACLEDB_HOST: localhost
+
+before_install:
+  - |
+    if [ $ORACLE_VER ]; then 
+      mkdir -p /opt/oracle && 
+      cd /opt/oracle && 
+      wget https://s3.amazonaws.com/sequelize/instantclient-basic-linux.x64-12.2.0.1.0.zip &&
```
cjbj: Instant Client 19.3 doesn't need a clickthrough from Oracle now, so the wget URL should use the Oracle official link, ie.  https://download.oracle.com/otn_software/linux/instantclient/193000/instantclient-basic-linux.x64-19.3.0.0.0dbru.zip

Also, recent node-oracledb versions don't need the sdk.

papb: Hi @cjbj thanks for the comment. Clearly you know much more about oracle than I do :) would you be willing to create a fork of this PR and make some updates, bringing this PR to life again?

sceik:
Hi,
it's almost two years that i used this pull request which supports oracle in production.
Can we resume the pull request? I can help but it's not clear what left to do

ggrimbert:
Hi,
Thanks for the feedback, I'm really happy that this work is useful for some people :)

Unfortunately I don't work anymore on this project or with the team that manages the fork.
Before I left, we decided that this PR is dead; as nobody in the Sequelize team seems to wants to merge this.
So we focused on our fork. You can find it [there](https://github.com/konnecteam/sequelize). It's used in a production environment and contains many bug fixes that haven't been resolved in the original project.
If you're using typescript, you can give a try at [this one](https://github.com/konnecteam/ts-sequelize). It was designed to be the future of our fork, but we hadn't had the time to completely finalize it.

If you're interested in one of these versions, feel free to contact @aegenet who is in charge of it.

papb:
Hi @sceik and @ggrimbert and everyone else :)

We aren't currently thinking of supporting more dialects, but I would say this PR is an exception. Since it is already started, with a lot of progress, I would love to see it merged :grimacing: 

It's just that we are so busy with everything (we are a bunch of people working voluntarily for the sake of open source)...

papb:
I think the next step now is to fix all the conflicts...

Is anyone interested in helping? Thanks in advance!!

sushantdhiman:
Hello Everyone,

I am closing this PR. No one in current Sequelize team has any experience with Oracle. Nor we have enough resources to maintain this new dialect.

You can use [TypeORM](https://typeorm.io) if you want to use Oracle in your projects.

Thanks for all the work you guys put into this.

Tkena (changes requested):
I like it

PR #7130, opened 2017-01-20
MSSQL - Skip isolationLevel query, enable tedious debugging

harshithkashyap:
### Pull Request check-list

### Description of change
1. Tedious accepts isolationLevel to be passed to beginTransaction query. Since we do the same [here](https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/query.js#L54), we can skip the `setIsolationLevelQuery`. The `setIsolationLevel` query is skipped by the [this](https://github.com/sequelize/sequelize/blob/master/lib/query-interface.js#L833-L837) existing snippet.
2. Enable debugging of tedious. Tedious can be run in debug mode by passing debug object with available options from [tedious api](http://tediousjs.github.io/tedious/api-connection.html) via `dialectOptions`. Added debug listener


janmeier:
:+1:

Issue #7141, opened 2017-01-24
v3: sourceKey feature for hasMany doesn't work if where clause specified in include


LoneWolfPR:
I installed the update that adds sourceKey to the hasMany feature. Initially it was working fine, however as soon as I put a where clause in the include of a find call it switched back to ignoring my specified sourceKey and is only looking at the primary key.

Here is my association defined.
```js
...
classMethods: {
  associate: (models) => {
    Contact.hasMany(models.Role, { foreignKey: 'contact__c', sourceKey: 'sfid' });
  },
},
```
Here is the find method I'm using.
```js
findByAuth0Id (auth0_id) {
  return db.User.find({
    where: { auth0_id },
    include: [{
      model: db.Contact,
      include: [
        {
          model: db.Role,
          where: { role_status__C: 'Active' },
          include: [db.Account],
        }],
    },
      db.Student,
      db.Clickwrap,
      db.Invitation,
      db.Notification,
    ],
  });
}
```
I had expected to get back a tree starting with the user that is found which would then include it's matching contact record. Then underneath the contact would be each active role with the active accounts attached to those roles.

I get `Unhandled rejection SequelizeDatabaseError: operator does not exist: integer = character varying` because the join on the contact and role tables is being done via the primary key on the contact table instead of the sourceKey I have specified in the hasMany association. If I leave everything the same, but simply remove the where clause in the include it works.

I'd be willing to work on a fix for this myself. I tried debugging the code, but I can't seem to find exactly where the final query gets written that is causing the problem. If no one else wants to take on fixing it I'd take it up if someone can point me in the right direction. I've gone through has-many.js, and the problem does not appear to be there as far as I can tell.

postgres v9.6.1
sequelize v3.30.0

LoneWolfPR:
I've actually found the problem. Currently we have this definition in `lib/dialects/abstract/query-generator.js` at line 1708
```js
, attrLeft = association instanceof BelongsTo ?
                   association.identifier :
                   left.primaryKeyAttribute
```
i believe It should be
```js
, attrLeft = association instanceof BelongsTo ?
                   association.identifier :
                   [association.sourceIdentifier || left.primaryKeyAttribute]
```

LoneWolfPR:
I have the repo pulled, and I'm ready to put in the fix, but I tried running the tests first and can't get them to complete. There's also a problem with a file disappearing after anytime tests fail. See this issue for details: https://github.com/sequelize/sequelize/issues/7143

PR #7147, opened 2017-01-25
v3: sourceKey feature for hasMany doesn't work if where clause specified in include (#7141)

LoneWolfPR:
### Pull Request check-list

### Description of change

This fixes the issue mentioned in https://github.com/sequelize/sequelize/issues/7141. Previously `sourceKey` was only respected in a find from hasMany association if the include did not contain a where clause. This fixes that issue so that even with a where clause specified the `sourceKey` supplied by the user will be respected.


LoneWolfPR:
Any chance this can get merged in sometime soon (assuming it meets all merge requirements)? We have a project that needs this pretty badly with multiple developers needing it.

lib/dialects/abstract/query-generator.js
```diff
@@ -1612,7 +1607,6 @@ var QueryGenerator = {
     }
 
     query += ';';
-
```
janmeier: Please remove the whitespace changes (it makes the changes easier to cherry-pick onto master)

test/integration/associations/has-many.test.js
```diff
@@ -1285,4 +1285,45 @@ describe(Support.getTestDialectTeaser('HasMany'), function() {
 
   });
 
+  describe('sourceKey with where clause in include', function() {
+    beforeEach(function() {
+      var User = this.sequelize.define('User',
+        { username: Sequelize.STRING, email: Sequelize.STRING },
```
janmeier: Could you change it to something that uses `field` (`email: { type: Sequelize.STRING, field: 'mail' }` or similar - So we can make sure this works (and continues to work) even when the field is called something other in the DB than in the model

janmeier: Username attribute and indexes are not needed for this test

test/integration/associations/has-many.test.js
```diff
@@ -1285,4 +1285,45 @@ describe(Support.getTestDialectTeaser('HasMany'), function() {
...
+            include: [
+              {
+                model: Task,
+                where: {taskStatus: 'Active'}
+              }
+            ],
+            where: { username: 'John' }
+          }).then(function(user) {
+            expect(user).to.be.ok;
+            expect(user.Tasks.length).to.equal(1);
```
janmeier: Please test that we get the correct task as well

LoneWolfPR:
@janmeier I've addressed all your requests. I believe it should be good to go now. Thanks for reviewing. Let me know if there are anymore issues.

janmeier:
Cherry-picked in https://github.com/sequelize/sequelize/commit/7d3172dfe4a46aaecc1492c42cc86f996bd663d2

janmeier:
Thanks for the adressing my feedback quickly @LoneWolfPR  :tada:

LoneWolfPR:
@janmeier no problem. Glad I could help. What's the timeline for this going out to the npm?

janmeier:
`+ sequelize@3.30.1`

danconnell:
Thanks @LoneWolfPR. Just found this bug myself and was very happy to see it fixed already 😄

derimagia:
I've been having an issue here where the field name is not the same as the model property name. So, if the 'field' is set and it's different the column name this isn't working.  

https://github.com/sequelize/sequelize/blob/3168d543f8d488f7c7db2d3d9d082762c288789a/lib/associations/has-many.js#L88-L98

Line 88 is using `this.target` and line 89 is using `this.source`, shouldn't they both be checking the source? 

Once `this.sourceKeyField`is fixed then `this.sourceKey` just needs to be replaced by it on lines 94-95. I'd be happy yo put a pull request if need be but wanted to check with someone who knows this more to correct me if I'm missing something.

PR #7155, opened 2017-01-26
Generic pool update v2

mkaufmaner:

### Pull Request check-list

### Description of change

This is a continuation of https://github.com/sequelize/sequelize/pull/7109

Suppresses warnings from bluebird when passed into the generic-pool module.


lib/dialects/abstract/connection-manager.js
```diff
@@ -31,10 +31,14 @@ class ConnectionManager {
...
     }
 
-    config.pool =_.defaults(config.pool || {}, defaultPoolingConfig, {
+    config.pool = _.defaults(config.pool || {}, defaultPoolingConfig, {
       validate: this._validate.bind(this),
-      Promise
-    }) ;
+      Promise: _.cloneDeep(Promise).config({
+        warnings: {
+          wForgottenReturn: false
```
sushantdhiman: User can pass `Promise` from pool config, also we allow any type of `Promise` to be used here which may or may not support this config `wForgottenReturn`

mkaufmaner: Okay, I'll close this.

Issue #6935, opened 2016-11-30
Remove order: string syntax in favor of order: Sequelize.literal()


felixfbecker:
This is an attack vector for SQL injection if people pass order from JSON user input.

mkaufmaner:
I believe this may also affect the group option.

PR #7160, opened 2017-01-27
6935 remove order string syntax

mkaufmaner:

### Pull Request check-list

### Description of change
Closes https://github.com/sequelize/sequelize/issues/6935


mkaufmaner:
Closed and reopened to trigger builds from commit that resolves change log conflict.

janmeier:
LGTM, would like one more @sequelize/core member to look at it

lib/dialects/abstract/query-generator.js
```diff
@@ -1483,24 +1483,40 @@ const QueryGenerator = {
...
       }
     };
 
+    const getOrderOption = order => {
+      const upperOrder = order.toUpperCase();
+      let index = -1;
+      let option = false;
+
+      for (const v of validOrderOptions) {
+        if(_.endsWith(upperOrder, v)){
```
felixfbecker: space after `if` and before opening brace

lib/dialects/abstract/query-generator.js
```diff
@@ -1519,8 +1535,26 @@ const QueryGenerator = {
...
+        order = _.trim(order);
+        let sql = this.quote(order, model, false, '->');
+        if(orderOption.option){
+          sql += ' ' + orderOption.option;
+        }
+        if (subQuery) {
+          subQueryOrder.push(sql);
+        }
+        mainQueryOrder.push(sql);
+      }
```
felixfbecker: Wait, how is this removing the string order option? Passing a string to `order` should throw an exception. If you want to order by a literal, use `Sequelize.literal()`. Any custom sanitisation is prone to be spoofed.

changelog.md
```diff
@@ -1,4 +1,5 @@
 # Future
+- [FIXED] Removed `Sequelize.literal()` support for order strings [#6935](https://github.com/sequelize/sequelize/issues/6935)
```
felixfbecker: This is the exact opposite of what we want. `Sequelize.literal()` must still be supported, only passing a string should be removed.

lib/dialects/abstract/query-generator.js
mkaufmaner: Fixed.

changelog.md
mkaufmaner: Fixed https://github.com/sequelize/sequelize/pull/7160/commits/33fa62050fc81f9b90b337847a78078d96443b68#diff-42ba1d994f4fa8c6ad17a7efae7936cc

lib/dialects/abstract/query-generator.js
mkaufmaner: @felixfbecker The string order option is not removed and you can still use `Sequelize.literal()`.  However, we can continue to support strings without breaking existing functionality.  Sure, custom sanitation methods are prone to being spoofed, however the order clause has a simple syntax where we can easily validate content and its arrangement.  So why loose existing functionality and when it could be extended?  If worst comes to worst, the logic being implemented here will quote a malicious or malformed order string and error out.

felixfbecker: We don't support raw strings in other places in favor of explicit Sequelize.literal(), like `attributes` etc. `order` is one of the last places where this is still supported. Sequelize codebase is already very complicated, there is also the desire to keep complexity low. Can you really guarantee that your algorithm can't be spoofed? Can you guarantee you handle every possible ORDER syntax? Just look at the possibilities for Postgres: https://www.postgresql.org/docs/9.1/static/queries-order.html, these also can vary through dialects. Since we don't test all the edge cases here there is still the potential to break someone's code and I don't want to commit to parsing every case here just to then put it inside the SQL again. Essentially this would introduce a mini DSL for `order` that is probably a subset of what's possible with SQL and it would still be an attack vector for SQL injection. I would rather target v4 and tell people to wrap in Sequelize.literal().

mkaufmaner: @felixfbecker Understood.  Will modify the tests/logic/documentation to only accept array.  Then I will do `group` option then.

felixfbecker: Awesome :)

lib/dialects/abstract/query-generator.js
felixfbecker: uhm, no?

sushantdhiman:
This is a breaking change in `order` api, as now we do not allow passing strings. Need to mention this in breaking changes section for Future release

lib/dialects/abstract/query-generator.js
```diff
@@ -1520,13 +1521,16 @@ const QueryGenerator = {
 
         mainQueryOrder.push(this.quote(t, model, false, '->'));
       }
-    } else {
-      var sql = this.quote(typeof options.order === 'string' ? new Utils.Literal(options.order) : options.order, model, false, '->');
+    } else if (typeof options.order === 'function' && options.order instanceof this.sequelize.literal) {
```
felixfbecker: If it is an `instanceof` a literal, shouldn't it be `typeof 'object'`?

test/unit/dialects/sqlite/query-generator.test.js
```diff
@@ -168,8 +168,12 @@ if (dialect === 'sqlite') {
           expectation: 'SELECT count(*) AS `count` FROM `foo`;',
           context: QueryGenerator
         }, {
+          arguments: ['myTable', {order: 'id; DROP TABLE myTable;'}],
+          expectation: 'SELECT * FROM `myTable` ORDER BY `id; DROP TABLE myTable;`;',
```
felixfbecker: This is what should throw an exception instead

lib/dialects/abstract/query-generator.js
mkaufmaner: @felixfbecker There are a bunch of changes that need to be done as well.  These are just temporary.  Can you change the status of this PR to pending?

felixfbecker: What do you mean with changing the status to pending?

mkaufmaner: Sorry, just woke up.  I mean remove the label "pr:final review"

mkaufmaner:
@felixfbecker Looks good!  Please review changes and let me know.  Take not of https://github.com/sequelize/sequelize/pull/7160/commits/3bd6aa118c4b47a8c8c99d51b3ae9fe265246908

mkaufmaner:
@sushantdhiman "Not completed yet"?

felixfbecker:
@mkaufmaner 
He just dismissed his approval.

If you want to trigger a build, please just `git commit --amend --no-edit; git push --force`, reopening sends emails to everyone in the thread

mkaufmaner:
@felixfbecker 
Gotcha and sorry about the multiple emails.  Again, not familiar with team development on github.
`git commit --amend --no-edit; git push --force` it is.

Please review and comment :)

felixfbecker:
@mkaufmaner And of course, if you just pushed a regular commit, that will trigger a build too.

changelog.md
```diff
@@ -48,6 +49,7 @@
 ## BC breaks:
 - `DATEONLY` now returns string in `YYYY-MM-DD` format rather than `Date` type
 - With `BelongsToMany` relationships `add/set/create` setters now set `through` attributes by passing them as `options.through` (previously second argument was used as `through` attributes, now its considered `options` with `through` being a sub option)
+- `options.order` Now only excepts values with type of array or sequelize method.  Support for string values (ie `{order: 'name DESC'}`) has been deprecated.
```
felixfbecker: ~~deprecated~~ removed

lib/dialects/abstract/query-generator.js
```diff
@@ -21,7 +20,7 @@ const QueryGenerator = {
     tableName = tableName || {};
     return {
       schema: tableName.schema || options.schema || 'public',
-      tableName: _.isPlainObject(tableName) ? tableName.tableName : tableName,
+      tableName: Utils._.isPlainObject(tableName) ? tableName.tableName : tableName,
```
felixfbecker: We moved away from using `Utils._`. This makes the diff unnecessary large.

lib/dialects/abstract/query-generator.js
```diff
@@ -1484,48 +1483,68 @@ const QueryGenerator = {
...
+      'ASC NULLS LAST',
+      'DESC NULLS LAST',
+      'ASC NULLS FIRST',
+      'DESC NULLS FIRST',
+      'NULLS FIRST',
+      'NULLS LAST'
+    ];
+
+    const methodOrder = order => {
+      let result = false;
```
felixfbecker: you don't need this variable

lib/dialects/abstract/query-generator.js
```diff
@@ -1484,48 +1483,68 @@ const QueryGenerator = {
...
+      'DESC',
+      'ASC NULLS LAST',
+      'DESC NULLS LAST',
+      'ASC NULLS FIRST',
+      'DESC NULLS FIRST',
+      'NULLS FIRST',
+      'NULLS LAST'
+    ];
+
+    const methodOrder = order => {
```
felixfbecker: this function should probably be in `utils` and called `isSequelizeMethod`

lib/dialects/abstract/query-generator.js
```diff
@@ -1484,48 +1483,68 @@ const QueryGenerator = {
...
             }
           } else {
-            validateOrder(_.last(t));
+            validateOrder(Utils._.last(order));
           }
         }
 
-        if (subQuery && (Array.isArray(t) && !(typeof t[0] === 'function' && t[0].prototype instanceof Model) && !(t[0] && typeof t[0].model === 'function' && t[0].model.prototype instanceof Model))) {
-          subQueryOrder.push(this.quote(t, model, false, '->'));
+        if (subQuery && (Array.isArray(order) && !(Utils._.isFunction(order[0]) && order[0].prototype instanceof Model) && !(order[0] && Utils._.isFunction(order[0].model) && order[0].model.prototype instanceof Model))) {
```
felixfbecker: Why the change to lodash typechecking functions everywhere? `typeof` is way faster than lodash's cross-browser implementation.

lib/dialects/sqlite/query-generator.js
```diff
@@ -332,7 +332,7 @@ const QueryGenerator = {
 
   quoteIdentifier(identifier) {
     if (identifier === '*') return identifier;
-    return Utils.addTicks(identifier, '`');
+    return Utils.addTicks(Utils.removeTicks(identifier, '`'), '`');
```
felixfbecker: why is this needed?

changelog.md
felixfbecker: capitalize Sequelize

lib/dialects/abstract/query-generator.js
mkaufmaner: Okay, so can we move completely away from using `Utils._.lodashMethod` and  change everything to be `_.lodashMethod`?

changelog.md
mkaufmaner: Done

lib/dialects/abstract/query-generator.js
mkaufmaner: Okay, I will change it back.

lib/dialects/abstract/query-generator.js
felixfbecker: yes. In many cases you don't even need lodash because theres a vanilla JS equivalent for it now

lib/dialects/abstract/query-generator.js
mkaufmaner: Well, I think I'll just remove this all together in favor of just checking if the `_isSequelizeMethod` property is set.

felixfbecker: No, we need the `instanceof` check. `_isSequelizeMethod` can easily be spoofed from JSON input.

lib/dialects/sqlite/query-generator.js
mkaufmaner: This is probably one of the most important changes features that prevents injection.  Similar to how values are escaped, this escapes identifiers by removing them.  See line 580 of `postgres/query-generator.js`.

lib/dialects/abstract/query-generator.js
mkaufmaner: I removed this so it is irrelevant now.

lib/dialects/sqlite/query-generator.js
mkaufmaner: @felixfbecker Maybe instead, this should escape the identifiers instead of removing them.  I think this is what `escapeId` on line 7 of `/lib/sql-string.js` was originally meant to do.

mkaufmaner: @felixfbecker After more consideration, we should not escape identifiers and just remove them.  Supporting identifiers with identifying characters in them seems silly.

janmeier:
Good work @mkaufmaner ! :tada:

PR #6325, opened 2016-07-22
Implement ON CONFLICT for postgres 9.5

inDream:
### Pull Request check-list
- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ ] Have you added new tests to prevent regressions?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Have you added an entry under `Future` in the changelog?

_NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open._
### Description of change

Implement `ON CONFLICT for postgres 9.5, Fix #4132 #3354.
It's a draft version since it's a breaking change.

I've used @ncwhale suggestion on `.upsert()`, but review is needed.
### API: Model.upsert()

``` js
/**
   * @param  {Object}          values
   * @param  {Object}          [options]
   * ...
   * @param  {Object}          [options.conflict] Actions on conflict
   * @param  {String}          [options.conflict.constraint=[primaryKey]] constraint_name
   * @param  {String}          [options.conflict.target] index_column_name OR index_expression
   * @param  {Object}          [options.conflict.where] constraint_name
   * @param  {Boolean|Array}   [options.update=[Columns exclude options.conflict.target]] "DO UPDATE SET" if provided an array of column name or not specific this option, "DO NOTHING" if false. 
   * @param  {Boolean}         [options.returning] Return updated instance if true
   *
   * @return {Promise<created>} Returns a boolean indicating whether the row was created or updated or updated instance.
   */
```
- For `options.conflict.constraint` and `options.conflict.target` please refer to https://www.postgresql.org/docs/9.5/static/sql-insert.html
- Currently if no `option.conflict` is provided, auto detection for primary key is applied.
- `bulkInsert` use the same options but under `options.updateOnConflict` namespace.
### Usage

``` js
// Use id as conflict target
Model.upsert(instance, {
  conflict: {
    target: 'id'
  }
});

// Use composite key as constraint
Model.upsert(instance, {
  conflict: {
    constraint: 'users_foo_bar_key'
  }
});

// Use complex primary key as constraint
Model.upsert(instance, {
  conflict: {
    constraint: 'users_pkey'
  }
});

// Choose which field to be updated
Model.upsert(instance, {
  conflict: {
    update: ['foo', 'bar']
  }
});

// Do nothing when conflicted
Model.upsert(instance, {
  conflict: {
[...]

PR #7174, opened 2017-01-31
Support INSERT... ON CONFLICT for upserts in v3

### LINKED ISSUE #4132: Postgres UPSERT native support
Looks like Postgres [recently added UPSERT](https://news.ycombinator.com/item?id=9509870) ([9.5alpha is it?](http://www.postgresql.org/about/news/1595/)). Seems Sequelize handles psql upsert fine presently with its workaround, so obviously no rush.

Me personally (and maybe I need re-architecting), my current project's most common SQL operation is UPSERT. That operation seems quite slow; this is the resultant SQL which _looks_ like it would be slow:

`CREATE OR REPLACE FUNCTION pg_temp.sequelize_upsert() RETURNS integer AS $func$ BEGIN INSERT INTO ... VALUES ...; RETURN 1; EXCEPTION WHEN unique_violation THEN UPDATE ... SET ... WHERE ...; RETURN 2; END; $func$ LANGUAGE plpgsql; SELECT * FROM pg_temp.sequelize_upsert();`

So I'd personally upgrade to Postgres 9.5 if native upsert becomes supported in sequelize

cuongdo:
Resolves #4132

Heavily inspired by #6325

### Pull Request check-list

### Description of change

This change adds the small extensions `options.conflict.target` and `options.conflict.constraint` to `Model.upsert` to support using PostgreSQL 9.5's `INSERT... ON CONFLICT` feature. This replaces  the stored procedure that's currently used. If Postgres >= 9.5 is used and no `options.conflict.*` options are specified, the primary key column(s) are implicitly used as the conflict target. If Sequelize is used with PostgreSQL prior to 9.5, we still use the existing stored procedure.

---
This change is [Reviewable](https://reviewable.io/reviews/sequelize/sequelize/7174)

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -216,6 +216,12 @@ var QueryGenerator = {
       emptyQuery += ' VALUES ()';
     }
 
+    if (this._dialect.supports.insertOnConflict && options.onConflict) {
+      valueQuery += ' ON CONFLICT ' + options.onConflict;
+      emptyQuery += ' ON CONFLICT ' + options.onConflict;
+      options.onConflictUsed = true;
+    }
+
     if (this._dialect.supports.returnValues && options.returning) {
       if (!!this._dialect.supports.returnValues.returning) {
         valueQuery += ' RETURNING *';
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -134,6 +134,10 @@ ConnectionManager.prototype.connect = function(config) {
     // Disable escape characters in strings, see https://github.com/sequelize/sequelize/issues/3545
     var query = '';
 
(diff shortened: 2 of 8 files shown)


cuongdo:
Could I please have some feedback?

I've also noticed that the code coverage checks are failing, and it seems that the new code isn't being triggered. Is the code coverage being measured against PostgreSQL 9.5?

cuongdo:
friendly ping

lironess:
Can I help somehow?
I really need that feature

sushantdhiman:
[v4](http://docs.sequelizejs.com/manual/changelog/changelog.html#4-0-0-final-) has been released and we have decided `v3` will only accept critical bug fixes and security updates, no more features will be added to `v3`

Although we will review and merge its master verrsion https://github.com/sequelize/sequelize/pull/6325 soon

PR #7182, opened 2017-02-02
run hooks during validation by default

### LINKED ISSUE #7178: model.validate() fails to run hooks  by default
## What you are doing?
_Post a minimal code sample that reproduces the issue, including models and associations_

```js
const sequelize = new Sequelize('sqlite://localhost:8000/nick.db');

const schema = {
  name: {
    type: Sequelize.STRING,
  }
}
const methods = {
  hooks: {
    beforeValidate: function(user) {
      console.log('running hook');
    }
  }
}
const User = sequelize.define('users', schema, methods);

  .then(() => {
    User.build({name: 'srn'}).validate();
  });
```

I expect "running hooks" to be logged to the console.

The hooks are not running.

The documentation states that the model.validate function will run hooks by default; however, when this function is invoked directly, the options object is never given any defaults, hence hooks do not run.

The reason why most users don't run into this issue is because they do not call model.validate directly. Typically, when the mode.validate function is called, it is called from another function, like Model.findAll. Model.findAll passes its options object into validate. Because findAll sets hooks to true by default, the hooks run. 

Here is the model.validate function
```
  validate(options) {
    return new InstanceValidator(this, options).validate();
  }
```

Here is my proposed change
```
  validate(options) {
    options = _.clone(options) || {};
    _.defaults(options, {hooks: true});
    return new InstanceValidator(this, options).validate();
  }
```

nadrane:
### Pull Request check-list

### Description of change

Fixes #7178


sushantdhiman:
I think we should have `hook: true` by default for `validate`, not sure why it was disabled in first place, wdyt @felixfbecker @janmeier @mickhansen ?

Although we still need tests around here https://github.com/sequelize/sequelize/blob/master/test/integration/hooks/validate.test.js#L26

felixfbecker:
Agree. Could be a breaking change, so probably best to keep it to v4.

lib/instance-validator.js
```diff
@@ -89,19 +90,27 @@ class InstanceValidator {
...
+   *   - On validation success: After Validation Model Hooks
+   *   - On validation failure: Validation Failed Model Hooks
+   *
+   * @return {Promise}
+   * @private
+   */
+  _validateAndRunHooks() {
+    const runHooks = this.modelInstance.constructor.runHooks.bind(this.modelInstance.constructor);
+    return runHooks('beforeValidate', this.modelInstance, this.options)
+      .then(() => this._validate())
```
nadrane: I was just looking at the diff and realized these are equivalent. I'm catching beforeValidate failures too now

nadrane:
@sushantdhiman What do you guys think?

test/integration/hooks/validate.test.js
```diff
@@ -108,47 +107,6 @@ describe(Support.getTestDialectTeaser('Hooks'), function() {
         });
       });
     });
-
-    describe('on error', function() {
```
sushantdhiman: These tests should work, why you have removed them

changelog.md
```diff
@@ -45,8 +45,10 @@
 - [FIXED] scope method syntax loses parameters when used multiple times [#7058](https://github.com/sequelize/sequelize/issues/7058)
 - [INTERNAL] Updated to `generic-pool@3.1.6` [#7109](https://github.com/sequelize/sequelize/issues/7109)
 - [FIXED] findAll throws error if attributes option is formatted incorrectly [#7162](https://github.com/sequelize/sequelize/issues/7163)
+- [FIXED] model validation instance method now runs hooks by default [#7182](https://github.com/sequelize/sequelize/pull/7182)
```
sushantdhiman: Link to issue #7178 , not PR

Changelog should say, "`Instance.validate` now run hooks by default", this is a breaking change, Need to add to bc breaks as well

changelog.md
```diff
@@ -45,8 +45,10 @@
...
 - [FIXED] previous gave wrong value back [#7189](https://github.com/sequelize/sequelize/pull/7189)
 
+
```
sushantdhiman: Extra line

test/integration/hooks/validate.test.js
nadrane: That's a really good question. When I initially viewed the tests, they didn't strike me as anything that necessitated being integration tests. In fact, they were written in such a way as to test the functionality of one specific function and had no intrinsic need to use the DB. I figured I would re-write them as unit tests, mostly because the integration tests are slow, and I don't see a reason to include extra ones unnecessarily.

But you're right, now we aren't calling any of the dependencies of `_validateAndRunHooks` anymore. Would you prefer I right unit tests for the rest of the methods of InstanceValidtor?

changelog.md
nadrane: Ah, I didn't know bc breaks were a thing. Will do.

test/integration/hooks/validate.test.js
nadrane: Alternatively, I can stick with the tests that are there

sushantdhiman: Keep these tests as they are, we need at least a few test running in integration, just to be sure

nadrane: Got it. Should I keep the unit tests too? Or does that seem excessive?

sushantdhiman: No keep them

nadrane:
@sushantdhiman Are we good to merge this one?

lib/instance-validator.js
```diff
@@ -79,7 +80,7 @@ class InstanceValidator {
   }
 
   /**
-   * Invoke the Validation sequence:
+   * Invoke the Validation sequence and run hooks if necessary
```
sushantdhiman: and run validation hooks if defined

test/unit/instance-validator.test.js
```diff
@@ -0,0 +1,112 @@
...
+const chai = require('chai');
+const expect = chai.expect;
+const Support = require(__dirname + '/support');
+const InstanceValidator = require('../../lib/instance-validator');
+const sinon = require('sinon');
+var Promise = Support.sequelize.Promise;
+const SequelizeValidationError = require('../../lib/errors').ValidationError;
+
+describe(Support.getTestDialectTeaser('InstanceValidator'), () => {
+
```
sushantdhiman: Please remove this extra line

test/unit/instance-validator.test.js
```diff
@@ -0,0 +1,112 @@
...
+  it('configures itself to run hooks by default', () => {
+    const instanceValidator = new InstanceValidator();
+    expect(instanceValidator.options.hooks).to.equal(true);
+  });
+
+  beforeEach(() => {
```
sushantdhiman: Good practice to keep `before/after [Each]` on top of the test, easy to see what this test suite depends upon

test/unit/instance-validator.test.js
```diff
@@ -0,0 +1,112 @@
...
+    this.User = Support.sequelize.define('user');
+  });
+
+  describe('validate', () => {
+    it('runs the validation sequence and hooks when the hooks option is true', () => {
+      const instanceValidator = new InstanceValidator(this.User.build(), {hooks: true});
+      const _validate = sinon.spy(instanceValidator, '_validate');
+      const _validateAndRunHooks = sinon.spy(instanceValidator, '_validateAndRunHooks');
+      instanceValidator.validate();
```
sushantdhiman: A new line before this line, please, and one after as well. Idea is to have something like this

```js
// define stuff

// do something

// expect things
```

Please improve test segments where this is applicable

test/unit/instance-validator.test.js
```diff
@@ -0,0 +1,112 @@
...
+    it('runs the validation sequence but skips hooks if the hooks option is false', () => {
+      const instanceValidator = new InstanceValidator(this.User.build(), {hooks: false});
+      const _validate = sinon.spy(instanceValidator, '_validate');
+      const _validateAndRunHooks = sinon.spy(instanceValidator, '_validateAndRunHooks');
+      instanceValidator.validate();
+      expect(_validate).to.have.been.calledOnce;
+      expect(_validateAndRunHooks).to.not.have.been.called;
+    });
+  });
+  describe('_validateAndRunHooks', () => {
```
sushantdhiman: A new line before this

test/unit/instance-validator.test.js
```diff
@@ -0,0 +1,112 @@
...
+
+    it('should emit an error from after hook when afterValidate fails', () => {
+      this.User.afterValidate(() => {
+        throw new Error('after validation error');
+      });
+      return expect(this.successfulInstanceValidator._validateAndRunHooks()).to.be.rejectedWith('after validation error');
+    });
+
+    describe('validatedFailed hook', () => {
+
```
sushantdhiman: Get rid of this empty line as well

changelog.md
```diff
@@ -45,9 +45,11 @@
 - [FIXED] scope method syntax loses parameters when used multiple times [#7058](https://github.com/sequelize/sequelize/issues/7058)
 - [INTERNAL] Updated to `generic-pool@3.1.6` [#7109](https://github.com/sequelize/sequelize/issues/7109)
 - [FIXED] findAll throws error if attributes option is formatted incorrectly [#7162](https://github.com/sequelize/sequelize/issues/7163)
-- [FIXED] previous gave wrong value back [#7189](https://github.com/sequelize/sequelize/pull/7189)
+- [FIXED] model validation instance method now runs hooks by default [#7182](https://github.com/sequelize/sequelize/pull/7182)
```
sushantdhiman: `Model.validate` run validation hooks by default

changelog.md
```diff
@@ -45,9 +45,11 @@
...
+- [FIXED] previous gave wrong value back [#7189](https://github.com/sequelize/sequelize/issues/7178)
 
 ## BC breaks:
+- The model validation instance method now runs hooks by default. Previously, hooks would only run when explictly passed as follows: model.validate({hooks: true})
```
sushantdhiman: `Model.validate` instance method run validation hooks by default now. Previously you need to pass `{ hooks: true }`. You can override this behavior by passing `{ hooks: false }`

nadrane:
I think I covered everything!

PR #7189, opened 2017-02-03
master:fix-previous-value

nemisj:
### Pull Request check-list

### Description of change

Previous method of the instance was working incorrectly

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2977,11 +2977,12 @@ class Model {
    * @param {any} value
    */
   setDataValue(key, value) {
-    const originalValue = this._previousDataValues[key];
+    const originalValue = this.dataValues[key];
     if (!Utils.isPrimitive(value) || value !== originalValue) {
       this.changed(key, true);
     }
 
+    this._previousDataValues[key] = originalValue;
     this.dataValues[key] = value;
   }
 
--- a/test/unit/instance/previous.test.js
+++ b/test/unit/instance/previous.test.js
@@ -1,0 +1,42 @@
+'use strict';
+
+/* jshint -W030 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support   = require(__dirname + '/../support')
+  , DataTypes = require(__dirname + '/../../../lib/data-types')
+  , current   = Support.sequelize;
+
+describe(Support.getTestDialectTeaser('Instance'), function () {
+  describe('previous', function () {
+    it('should return correct previous value', function () {
+      var Model = current.define('Model', {
+          text: {
+            type: DataTypes.STRING,
+            get: function (name) {
+              return this.getDataValue(name);
+            },
+            set: function (value, name) {
+              this.setDataValue(name, value);
+            }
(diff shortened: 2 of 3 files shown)


nemisj:
@sushantdhiman , added

test/unit/instance/previous.test.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+/* jshint -W030 */
+var chai = require('chai')
+  , expect = chai.expect
+  , Support   = require(__dirname + '/../support')
+  , DataTypes = require(__dirname + '/../../../lib/data-types')
+  , current   = Support.sequelize;
+
+describe(Support.getTestDialectTeaser('Instance'), function() {
```
sushantdhiman: Space before opening brackets `function ()`

nemisj:
done

BorisKozo:
This fix broke our application :D
Reminded me of this: https://xkcd.com/1172/

barriber:
Previous function doesn't return the correct values now,

```
return Promise.map(instances, function(instance) {
          // Record updates in instances dataValues
          Utils._.extend(instance.dataValues, values);
          // Set the changed fields on the instance
          Utils._.forIn(valuesUse, function(newValue, attr) {
            if (newValue !== instance._previousDataValues[attr]) {
              instance.setDataValue(attr, newValue);
            }
          });

```
instance.dataValues override with the new values in  Utils._.extend(instance.dataValues, values);
and  in instance.setDataValue(attr, newValue); 


```
Instance.prototype.setDataValue = function(key, value) {
  var originalValue = this.dataValues[key];
  if (!Utils.isPrimitive(value) || value !== originalValue) {
    this.changed(key, true);
  }

  this._previousDataValues[key] = originalValue;
  this.dataValues[key] = value;
};
```

the _previousDataValues overrided with same new data value

nemisj:
@barriber so actually we also have to change that code which sets dataValues directly I guess, into something like:

```
return Promise.map(instances, function(instance) {
          // Set the changed fields on the instance
          Utils._.forIn(valuesUse, function(newValue, attr) {
              instance.setDataValue(attr, newValue);
          });
```

in order to guarantee the same data flow into the instance

barriber:
@janmeier could you please comment if this is the correct solution?

nemisj:
@barriber i see that there is a note in `changelog.md` saying that from 2.0.0. version _previousDataValues must be updated after `afterUpdate`. But I don't understand this comment completely why it's like that. So i guess this solution is also not good and should be done differently.

lib/model.js
```diff
@@ -2977,11 +2977,12 @@ class Model {
    * @param {any} value
    */
   setDataValue(key, value) {
```
Verdier: This does't work if you call setDataValue multiple times... `previous` should return the original value (ie. the value in the DB when the record was fetched), am I wrong?

Verdier: Ok that was reverted... sorry ^^

Issue #7190, opened 2017-02-03
Error Insert date stringified in field DATEONLY


Fridus:
## What you are doing?

I try to set a date in a field type of `DATEONLY`  
I use the validation `isDate`

```
dob: {
      type: Sequelize.DATEONLY,
      allowNull: false,
      validate: { isDate: true },
      get: function () {
        return moment.utc(this.getDataValue('dob')).format('YYYY-MM-DD')
      }
    },
```

## What do you expect to happen?
the date pass the validation and the record was created successfully !

## What is actually happening?
When the date is strigified (ex: `2017-01-11T00:00:00.000Z`), the validation say OK, but mysql say error...
```
Error: ER_TRUNCATED_WRONG_VALUE: Incorrect date value: '2017-01-11T00:00:00.000Z' for column 'dob' at row 1
```

This bug is not present with type `DATE`

## My fix

```
dob: {
      type: Sequelize.DATEONLY,
      allowNull: false,
      validate: { isDate: true },
      get: function () {
        return moment.utc(this.getDataValue('dob')).format('YYYY-MM-DD')
      },
      set: function (val) {
        this.setDataValue('dob', new Date(val))
      }
    },
```

__Dialect:__ mysql
__Database version:__ 5.6
__Sequelize version:__ 3.30.0

sushantdhiman:
Fixed https://github.com/sequelize/sequelize/pull/6709 (will be released in v4)

Fridus:
Thx @sushantdhiman 
When the v4 will be released ?
Not possible to pick the fix in v3 ? ^^

sushantdhiman:
@Fridus If you want you can backport changes and send a PR, we will release new version in v3 with fix

PR #7192, opened 2017-02-03
data-types DATEONLY: Fix stringify

Fridus:
Issue #7190


sushantdhiman:
@Fridus I am sorry, didnt thought about this before. This would be a breaking change, we cant accept it in `v3`. Its already fixed in `v4`, please use that.

PR #7131, opened 2017-01-21
Few Improvements

sushantdhiman:
- Improved Code Coverage
- Fixed deadlock bug with Postgres https://github.com/sequelize/sequelize/issues/6843
- New unit + integration tests

PR #7196, opened 2017-02-03
Added unique indexes to rawAttributes

harshithkashyap:
## Pull Request check-list

### Description of change
`indexes` defined via the following snippet are not added to `Model.rawAttributes`.

```javascript
const User = current.define('User', {
    name: DataTypes.STRING,
    address: DataTypes.STRING
}, {
    indexes: [{
        unique: true,
        fields: ['name', 'address']
    }]
});

```

`Model.upsert` query uses `rawAttributes` to find primary/unique indexes and hence the upsert query would fail in case of the above model definition.

Fixes https://github.com/sequelize/sequelize/pull/6842#issuecomment-274774286

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -842,7 +842,19 @@ class Model {
       }
     });
 
-    this.options.indexes = this.options.indexes.map(this._conformIndex);
+    this.options.indexes = this.options.indexes.map(index => {
+      index = this._conformIndex(index);
+      //Add unique indexes to rawAttributes
+      if (index.unique && index.fields) {
+        index.fields.forEach(field => {
+          const fieldName =  (typeof field === 'string') ? field : (field.name || field.attribute);
+          if (this.rawAttributes[fieldName]) {
+            this.rawAttributes[fieldName].unique = this.attributes[fieldName].unique = index.name || index.unique;
+          }
+        });
+      }
+      return index;
+    });
 
     this.sequelize.modelManager.addModel(this);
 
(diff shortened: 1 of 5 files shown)


harshithkashyap:
Ping @sushantdhiman @felixfbecker @janmeier

lib/model.js
```diff
@@ -842,7 +842,19 @@ class Model {
       }
     });
 
-    this.options.indexes = this.options.indexes.map(this._conformIndex);
+    this.options.indexes = this.options.indexes.map(index => {
+      index = this._conformIndex(index);
+      //Add unique indexes to rawAttributes
+      if (index.unique && index.fields) {
+        index.fields.forEach(field => {
```
sushantdhiman: What will happen in case of unique index on two fields (composite unique index), will it mark both fields as `unique`? I would like to see a test which works in case of composite unique index

harshithkashyap: @sushantdhiman Composite index also works. I have added an unit test for it but I'll add an integration test also.

felixfbecker: use `for of`

lib/model.js
```diff
@@ -842,7 +842,19 @@ class Model {
...
+          const fieldName =  (typeof field === 'string') ? field : (field.name || field.attribute);
```
felixfbecker: unneeded parens and space

harshithkashyap:
@sushantdhiman Looks like the deadlocks still occur. Was this not fixed by https://github.com/sequelize/sequelize/pull/7131?

PR #7220, opened 2017-02-07
6935 remove order string syntax

mkaufmaner:

### Pull Request check-list

### Description of change

Continues https://github.com/sequelize/sequelize/pull/7160

Instead of using `SequelizeMethod._isSequelizeMethod` we use `instanceof Utils.SequelizeMethod` because _.isSequelizeMethod can be spoofed.


janmeier:
:taco:

PR #7224, opened 2017-02-08
Add quotes around column names for unique constraints in sqlite (fixes #4407)

### LINKED ISSUE #4407: Can't create UNIQUE constraints with column name of`group` in sqlite
I've created a table with a column, named `group`, and there's a `UNIQUE` constraint on it. It works on postgres, but will fail on sqlite. The SQL statement is 

``` SQL
CREATE TABLE IF NOT EXISTS `Requests` (`id` INTEGER PRIMARY KEY AUTOINCREMENT, `group` INTEGER NOT NULL REFERENCES `Groups` (`id`) ON DELETE CASCADE ON UPDATE CASCADE, `license` INTEGER NOT NULL REFERENCES `Licenses` (`id`) ON DELETE CASCADE ON UPDATE CASCADE, `createdAt` DATETIME NOT NULL, `updatedAt` DATETIME NOT NULL, UNIQUE (group, license));
```

However the correct SQL should be as below, which needs column names in `UNIQUE` to be quoted.

``` SQL
CREATE TABLE IF NOT EXISTS `Requests` (`id` INTEGER PRIMARY KEY AUTOINCREMENT, `group` INTEGER NOT NULL REFERENCES `Groups` (`id`) ON DELETE CASCADE ON UPDATE CASCADE, `license` INTEGER NOT NULL REFERENCES `Licenses` (`id`) ON DELETE CASCADE ON UPDATE CASCADE, `createdAt` DATETIME NOT NULL, `updatedAt` DATETIME NOT NULL, UNIQUE (`group`, `license`));
```

May you help to fix the issue, thank you.

Table definition: 

``` javascript
sequelize.define('Request', {
    group: {
      type: Sequelize.INTEGER,
      allowNull: false,
      unique: 'uGroupLicense'
    },
    license: {
      type: Sequelize.INTEGER,
      allowNull: false,
      unique: 'uGroupLicense'
    }
  }
```

StefanoDalpiaz:
This PR fixes a bug in the query generator for Sqlite, in which the definition of a composite unique constraint would not properly wrap in quotes the column names. This causes the statement to fail whenever a column name is the same as a reserved keyword (like `order`, `group`, `index`, ...), or when it contains certain non alpha-numerical characters (like a space).

For example the following declaration:

```js
this.sequelize.define('user', {
  name: {
    type: Sequelize.STRING,
    unique: 'uniqueConstraint'
  },
  'column with space': {
    type: Sequelize.INTEGER,
    unique: 'uniqueConstraint'
  }
});
```

would generate the following statement (new lines added for readability):

```sql
CREATE TABLE IF NOT EXISTS `users` (
  `id` INTEGER PRIMARY KEY AUTOINCREMENT,
  `name` VARCHAR(255),
  `column with space` VARCHAR(255),
  `createdAt` DATETIME NOT NULL,
  `updatedAt` DATETIME NOT NULL,
  UNIQUE (name, column with space)
);
```

The above statement contains a syntax error, as the name `column with space` inside the `UNIQUE` constraint must be wrapped in quotes.

This PR modifies the logic that builds that statement to ensure that those names are properly wrapped.

I have also provided a couple of tests to prevent regression.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -48,6 +48,7 @@
 - [INTERNAL] Updated to `generic-pool@3.1.6` [#7109](https://github.com/sequelize/sequelize/issues/7109)
 - [FIXED] findAll throws error if attributes option is formatted incorrectly [#7162](https://github.com/sequelize/sequelize/issues/7163)
 - [FIXED] previous gave wrong value back [#7189](https://github.com/sequelize/sequelize/pull/7189)
+- [FIXED] Add quotes around column names for unique constraints in sqlite [#7224](https://github.com/sequelize/sequelize/pull/7224)
 
 ## BC breaks:
 - `DATEONLY` now returns string in `YYYY-MM-DD` format rather than `Date` type
--- a/changelog.md
+++ b/changelog.md
@@ -48,7 +48,7 @@
 - [INTERNAL] Updated to `generic-pool@3.1.6` [#7109](https://github.com/sequelize/sequelize/issues/7109)
 - [FIXED] findAll throws error if attributes option is formatted incorrectly [#7162](https://github.com/sequelize/sequelize/issues/7163)
 - [FIXED] previous gave wrong value back [#7189](https://github.com/sequelize/sequelize/pull/7189)
-- [FIXED] Add quotes around column names for unique constraints in sqlite [#7224](https://github.com/sequelize/sequelize/pull/7224)
+- [FIXED] Add quotes around column names for unique constraints in sqlite [#4407](https://github.com/sequelize/sequelize/pull/4407)
 
 ## BC breaks:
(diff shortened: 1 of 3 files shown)


sushantdhiman:
Rather than integration test, we should add unit test for this change, because we just want to verify correct statement was generated

StefanoDalpiaz:
As requested, I have removed the new integration tests and added a unit test.

Also, is there any chance that this can be backported into v3? I can open another PR for that if needed.

sushantdhiman:
> Also, is there any chance that this can be backported into v3? I can open another PR for that if needed.

Please go ahead and open a PR targeting v3, we will release new version for you :)

StefanoDalpiaz:
> Please go ahead and open a PR targeting v3, we will release new version for you :)

Thanks, I just opened #7270 for this purpose.

PR #7242, opened 2017-02-13
V3 fixes callback beforeUpdate with model.update and individualHooks:…

lstone:
### Pull Request check-list

### Description of change

Fixes callback beforeUpdate with model.update and individualHooks:true not providing correct changed values in callback. Utils._.extend was overriding changed values before they were copied to _previousDataValues with Utils._.forIn in model.js.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+# 3.30.3
+- [FIXED] instance model callback beforeUpdate with model.update and individualHooks:true not providing the changed value updates in callback. Utils._.extend was overriding changed values before they were copied to _previousDataValues with Utils._.forIn in model.js.
+
 # 3.30.2
 - [FIXED] `previous` method gave wrong value back [#7189](https://github.com/sequelize/sequelize/pull/7189)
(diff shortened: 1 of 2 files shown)


overlookmotel:
Huh? Shouldn't `instance.setDataValue()` set the `.dataValues`?

I don't know this part of the codebase well, but the name `setDataValue` would suggest it sets the data value! Perhaps the fix should be in `instance.setDataValue()` instead?

lstone:
Thanks guys, @overlookmotel - setDataValues isn't in the codebase, just setDataValue. The problem is that Utils._.extend is called first and that mutates the object before saving the changed values (which were overwritten by extend. From the Lodash source code of extend(): 

     * This method is like `_.assign` except that it iterates over own and
     * inherited source properties.
     *
     * **Note:** This method mutates `object`.

@janmeier - Will add regression tests when I have a moment.

EddyBorja:
This update fixed a bug I had with a beforeUpdate hook. Any idea when it will be merged in?

sushantdhiman:
[v4](http://docs.sequelizejs.com/manual/changelog/changelog.html#4-0-0-final-) has been released and we have decided `v3` will only accept critical bug fixes and security updates, no more features will be added to `v3`

PR #7269, opened 2017-02-19
Vertica Dialect for Sequelize

yudarik:

### Pull Request check-list

### Description of change


janmeier:
This seems like close to a copy-pasta of the postgres dialect (see https://github.com/startappdev/sequelize/blob/8cd43d5ba2295ef82d8fd1f9ef81bb1a2e85dfdb/lib/dialects/vertica/query-generator.js#L45). Is there actually any difference? It also seems you can connect with the pg library (https://github.com/startappdev/sequelize/blob/8cd43d5ba2295ef82d8fd1f9ef81bb1a2e85dfdb/lib/dialects/vertica/connection-manager.js#L19), or is that just too much copy pasta :)

If it is so close to postgres there is no reason to duplicate the code.

Personally I've never heard of Vertica, and while we generally appreciate contributions, I'd be vary against merging something that only a few people need and even fewer can fix bugs on.

yudarik:
Right. The vertica dialect is very like the postgress, although there are few differences since both Vertica and Postgress where the same codebase somewhere in the history, but HP took it to the next level and created Vertica as a solution for Big Data databases.
I've made all necessary changes to make it work.
There are few unit-test that I still need to fix, but I think it's nice to have for people like me that need to work with Vertica on Node.js projects.

janmeier:
Yes, it would be nice for you, but we have to take into account that the code should also be maintainable for us :)

Is it possible to install a free version of vertica? Does it comes as a docker image?

Also, to get this merged, you would have to re-use as much as possible of the postgres dialect. You can do so by importing from the corresponding postgres files, cloning, and overwriting the methods that are different

janmeier:
We are open to _discussions_ and questions about new dialects (see the open oracle PR, which we will hopefully merge soon), but we need to see demand from a substantial amount of users to merge it

mike-seekwell:
I'm also interested in this. @yudarik Is your PR working for you? Have you been using it regularly? 

Also, has anyone tried just using the Postgres dialect with Vertica (maybe an older version like with Redshift)? I'm going to try this week, just wanted to know if anyone else has already done it and what version of Postgres you used.

Issue #7204, opened 2017-02-06
deleted records can be queried in the same second


Diluka:
## What you are doing?
querying records may contain deleted ones

```ts
// code here
// paranoid model User
await User.delete(1);
let user = await User.findById(1);
```

## What do you expect to happen?
user is null

## What is actually happening?
user maybe not null

Dialect: mysql
Database version: 5.7.16-1debian8
Sequelize version: 4.0.0-2

I've check #5880 and #5897. 
1. should use database time rather than servers'
The server time may be inconsistent with the database time.
2. should use $gt rather than $gte
For all databases, the time accuracy of the database can only be guaranteed to seconds, so deleted records may be in the same seconds to be queried. To avoid this, the first query condition can not contain the current second, followed by all the time required to the database time.

some refers: 
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L80
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L2445-L2452
https://github.com/sequelize/sequelize/blob/master/lib/utils.js#L426-L433

Celadora:
Try turning off pooling and see if you get the same results. This might be a race condition. As deleting takes slightly longer than selecting, so your results might be returned before the deleting takes place if these are run in parallel.

If this is the case you may want to try using a transaction if you don't want to turn off pooling. I can't be certain if that will fix it, but that's my hunch.

Wsiegenthaler:
This is more of a bug than the title suggests - despite my server and database both residing in the same timezone, deleted paranoid records are still returned by findAll for a full 8 hours after their deletedAt field is populated (the system is hosted on the west coast, currently 8 hours behind UTC). As op mentions in point #1, the deletedAt field should be populated with the database time for consistency.

I'll open a PR tomorrow to address this as well as issue #2 (should use $gt rather than $gte).

sushantdhiman:
Fixed https://github.com/sequelize/sequelize/commit/1f8348fbf42ff5991ee7d70296f92814b8255344

Issue #854, opened 2013-08-23
Setting timezone


nczz:
how can i use this option correct? and this option is not showing in doc.

``` js
sequelize    = new Sequelize(config.database, config.username, config.password, { timezone : "+08:00", pool:config.pool, dialect: 'mysql', logging: false, host: config.host})
```

it not working...

diosney:
@janmeier Your previous explanation saved me from a very nasty bug with moment due a confusion between local and UTC dates.

It would be nice to have it in the official docs ;) , specially the **end** paragraph, since were really helpful to me and there are a lot of people out there having issues with this too.

cherchyk:
sequelize converts time to my local zone when it reads from DB.

How to disable automated conversion?

I tried timezone : "+08:00" but values was not changed.

lovelyelfpop:
timezone : "+08:00"  only effect on writing date to database, not read from database

lovelyelfpop:
var sequelize = new Sequelize(connStr, {
    dialectOptions: {
        useUTC: false //for reading from database
    },
    timezone: '+08:00' //for writing to database
});

feiZhang:
@lovelyelfpop 
https://github.com/mysqljs/mysql
not useUTC

catalsdevelop:
a fancy solution is to custom your own cast function when a DATETIME type field found.

``` js
var sequelize = new Sequelize(connStr, {
    typeCast: function (field, next) { // for reading from database
      if (field.type === 'DATETIME') {
        return field.string();
      }
      return next();
    },
    timezone: '+08:00' // for writing to database
});
```

ernani:
Does hardcoding the timezone accounts for DST settings? Thanks!

RichAyotte:
Make sure the timezone is the same everyone.

```js
const timezone = 'UTC'
process.env.TZ = timezone

const db = {
  options: {
    dialectOptions: {
      timezone
    }
    , timezone
  }
}



```

karsens:
This is pretty shitty about Sequelize in my opinion. Why doesn't it just take over the timezone from the MySQL Date as it is? Don't change things based on false assumptions. 

I wrote this about it: https://karsens.com/handling-time-on-your-server-using-different-orms-in-node-sequelize-and-mysql2-beware/

samueldelesque:
It seems Sequelize is doing a double local timezone correction, so my dates end up being GMT-10 instead of GMT-5.

samueldelesque:
[...]

PR #7290, opened 2017-02-25
Inconsistent behavior for paranoid records since #5897 (see issue #7204)

Wsiegenthaler:
This is a fix for 2 issues introduced by #5880/#5897 and documented in #7204:

1. Paranoid records remain visible for several hours after they've been deleted. How long the records take to delete seems to depend on how far the server is from UTC. The solution is to always use database time when populating and later evaluating the deletedAt field.
2. Deleted records can be queried in the same second. The paranoid clause should use 'greater than' instead of 'greater than or equal to' when evaluating the deletedAt field.

Given that this issue fundamentally affects the ability to delete paranoid records, I'd advocate for getting this into v4.

Relevant tests have been updated and now pass, and this change has been added to the 'future' section of the changelog. Please let me know if there's anything else I can do to expedite this!

### DIFF
--- a/test/integration/instance.test.js
+++ b/test/integration/instance.test.js
@@ -1810,7 +1810,7 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
       return this.ParanoidUser.create({ username: 'fnord' }).then(function() {
         return self.ParanoidUser.findAll().then(function(users) {
           return users[0].destroy().then(function() {
-            expect(users[0].deletedAt.getMonth).to.exist;
+            expect(users[0].deletedAt.val).to.be.equal('CURRENT_TIMESTAMP');
 
             return users[0].reload({ paranoid: false }).then(function(user) {
               expect(user.deletedAt.getMonth).to.exist;
--- a/test/integration/model.test.js
+++ b/test/integration/model.test.js
@@ -1333,8 +1333,6 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       return ParanoidUser.sync({ force: true }).then(function() {
         return ParanoidUser.bulkCreate(data);
       }).bind({}).then(function() {
-        // since we save in UTC, let's format to UTC time
-        this.date = moment().utc().format('YYYY-MM-DD h:mm');
         return ParanoidUser.destroy({where: {secretValue: '42'}});
       }).then(function() {
         return ParanoidUser.findAll({order: ['id']});
@@ -1342,13 +1340,10 @@ describe(Support.getTestDialectTeaser('Model'), function() {
         expect(users.length).to.equal(1);
         expect(users[0].username).to.equal('Bob');
 
-        return self.sequelize.query('SELECT * FROM ' + qi('ParanoidUsers') + ' WHERE ' + qi('deletedAt') + ' IS NOT NULL ORDER BY ' + qi('id'));
+        return self.sequelize.query('SELECT * FROM ' + qi('ParanoidUsers') + ' WHERE ' + qi('deletedAt') + ' IS NOT NULL AND ' + qi('deletedAt') + ' <= CURRENT_TIMESTAMP ORDER BY ' + qi('id'));
       }).spread(function(users) {
         expect(users[0].username).to.equal('Peter');
         expect(users[1].username).to.equal('Paul');
-
-        expect(moment(new Date(users[0].deletedAt)).utc().format('YYYY-MM-DD h:mm')).to.equal(this.date);
-        expect(moment(new Date(users[1].deletedAt)).utc().format('YYYY-MM-DD h:mm')).to.equal(this.date);
       });
     });
 
--- a/changelog.md
+++ b/changelog.md
@@ -53,6 +53,8 @@
 - [FIXED] previous gave wrong value back [#7189](https://github.com/sequelize/sequelize/pull/7189)
 - [FIXED] Add quotes around column names for unique constraints in sqlite [#4407](https://github.com/sequelize/sequelize/pull/4407)
 - [FIXED] Connection error when fetching OIDs for unspported types in Postgres 8.2 or below [POSTGRES] [#5254](https://github.com/sequelize/sequelize/issues/5254)
+- [FIXED] Always use database time for populating and evaluating deletedAt field on paranoid records. [#7204](https://github.com/sequelize/sequelize/issues/7204)
+- [FIXED] Deleted paranoid records can be queried in the same second. [#7204](https://github.com/sequelize/sequelize/issues/7204)
 
 ## BC breaks:
 - `DATEONLY` now returns string in `YYYY-MM-DD` format rather than `Date` type
--- a/lib/model.js
+++ b/lib/model.js
@@ -77,7 +77,7 @@ class Model {
     const deletedAtObject = {};
     let deletedAtDefaultValue = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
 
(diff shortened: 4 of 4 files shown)


test/integration/model.test.js
```diff
@@ -1333,22 +1333,17 @@ describe(Support.getTestDialectTeaser('Model'), function() {
...
         expect(users.length).to.equal(1);
         expect(users[0].username).to.equal('Bob');
 
-        return self.sequelize.query('SELECT * FROM ' + qi('ParanoidUsers') + ' WHERE ' + qi('deletedAt') + ' IS NOT NULL ORDER BY ' + qi('id'));
+        return self.sequelize.query('SELECT * FROM ' + qi('ParanoidUsers') + ' WHERE ' + qi('deletedAt') + ' IS NOT NULL AND ' + qi('deletedAt') + ' <= CURRENT_TIMESTAMP ORDER BY ' + qi('id'));
       }).spread(function(users) {
         expect(users[0].username).to.equal('Peter');
         expect(users[1].username).to.equal('Paul');
-
-        expect(moment(new Date(users[0].deletedAt)).utc().format('YYYY-MM-DD h:mm')).to.equal(this.date);
```
sushantdhiman: These tests can still work, current timestamp will be returning UTC date, which can be compared with `this.date` defined above

test/integration/instance.test.js
```diff
@@ -1810,7 +1810,7 @@ describe(Support.getTestDialectTeaser('Instance'), function() {
       return this.ParanoidUser.create({ username: 'fnord' }).then(function() {
         return self.ParanoidUser.findAll().then(function(users) {
           return users[0].destroy().then(function() {
-            expect(users[0].deletedAt.getMonth).to.exist;
+            expect(users[0].deletedAt.val).to.be.equal('CURRENT_TIMESTAMP');
```
sushantdhiman: Shouldn't this be populated with actual date object, which is returned by `CURRENT_TIMESTAMP` ?

Wsiegenthaler: Since updates only return the number of affected rows, we would need to issue a subsequent SELECT to retrieve the actual timestamp assigned by the database.

test/integration/model.test.js
Wsiegenthaler: I'm not sure why, but when I run this test the locally instantiated date is set to `1970-01-01 12:00` which is nowhere near the `CURRENT_TIMESTAMP` of the database, thereby causing the test to fail. Perhaps this is part of the test suite?

Regardless, this seems to highlight the problems that can arise when database and system clocks for whatever reason aren't synced. We *could* still compare the `deletedAt` field assigned by the database with the locally instantiated one, but seeing as the server time is no longer used to populate the database I'm not sure it buys us anything.

Wsiegenthaler:
Hello Sushant,
Thanks for taking a look! I've responded to both of your comments though I'm not very pleased with the answers. It seems that treating 'deletedAt' as a true timestamp rather than a mere boolean (à la #5897) will have a caveat either way we go.

The caveat as things currently exist (caveat probably being an understatement), is that unless both your server and database are hosted in UTC, deletions may take effect several hours after expected.

On the other hand, with the introduction of this PR, destroyed instances will not contain a valid `deletedAt` timestamp but instead the `CURRENT_TIMESTAMP` literal used for the UPDATE. This doesn't feel great either, but in my opinion is far preferable to discrepancies between server, database, and UTC time causing inconsistent deletions.

Of course a third option is to revert back to treating `deletedAt` as a boolean. This obviates the need for consistent timestamps, but we lose the ability to schedule records for future deletion. Speaking of which, by using `CURRENT_TIMESTAMP` as this PR does, setting a future `deletedAt` is impossible using the standard `destroy()` call anyway. In the interest of getting 4.0 out the door it might be best to pull #5897 until all the details are thought through.

I've only been using Sequelize a short while so I'll be relying on you guys to suggest the most prudent approach. I do think we have to do *something* to address the current issue before 4.0 however - otherwise users will almost surely find their deleted records suddenly coming back from the dead.

sushantdhiman:
Thanks @Wsiegenthaler, I did some brainstorming about concerns raised by this PR. Although some of these concerns are valid but we can have timestamp backed soft delete, without much change.

First, I think change `$gte` => `$gt` is acceptable, I have no problem with that. Its the right way to do this.

Second, Sequelize consider your database server is running in UTC (Your application server can run in any timezone). Working of our timezone system as explained by @janmeier 2 years ago https://github.com/sequelize/sequelize/issues/854#issuecomment-29069241

Main concern you raise in #7204 was `The server time may be inconsistent with the database time`. You mentioned your app server is running in west coast (I assume -08:00 UTC). You just need to set `timezone` option to correct setting. Because `Utils.now` generate a timestamp which is backed by timezone mentioned in  `sequelize.config.timezone`

With correct timezone setting, the offset you were noticing should be fixed. Sequelize was considering your app was running in UTC the whole time.

As I tested with `Asia/Kolkata` (`+ 05:30`), you can see it correctly set the timezone 

```js
Executing (default): INSERT INTO "Ms" ("id","test","createdAt","updatedAt") VALUES (DEFAULT,34,'2017-03-05 11:08:39.444 +05:30','2017-03-05 11:08:39.444 +05:30') RETURNING *;
{ id: 1,
  test: 34,
  updatedAt: 2017-03-05T05:38:39.444Z,
  createdAt: 2017-03-05T05:38:39.444Z,
  deletedAt: null }
Executing (default): UPDATE "Ms" SET "deletedAt"='2017-03-05 11:08:39.466 +05:30' WHERE "id" = 1 AND "deletedAt" IS NULL
{ id: 1,
  test: 34,
  updatedAt: 2017-03-05T05:38:39.444Z,
  createdAt: 2017-03-05T05:38:39.444Z,
  deletedAt: 2017-03-05T05:38:39.466Z } // Correct UTC based timestamp
Executing (default): SELECT "id", "test", "createdAt", "updatedAt", "deletedAt" FROM "Ms" AS "M" WHERE (("M"."deletedAt" > CURRENT_TIMESTAMP OR "M"."deletedAt" IS NULL) AND "M"."id" = 1);
null
```

Wsiegenthaler:
I just ran some tests and you're absolutely right - it appears the crux is to always keep your database in UTC.  I did just this in my dev environment and haven't seen any problems, but are there any pitfalls I should be aware of when changing the timezone in production?

I'll be closing this PR and opening a new one with only the $gte change to the paranoid clause. I think the UTC requirement really ought to be in the docs front and center to prevent others from having the same issues. Let me know if I can help with that.

Thanks again for taking the time to work through this!

gabegorelick:
Apologies for coming late to this, but issues like https://github.com/sequelize/sequelize/issues/8202 have me diving into the consistency of Sequelize in this area. 

1. Reads that occur less than a millisecond after a delete (i.e. less than the timestamp resolution) will still have this issue. That's potentially what's happening in  https://github.com/sequelize/sequelize/issues/8202. The only solution I see for this is to drop the `deleted_at > CURRENT_TIMESTAMP` clause when querying and only rely on `deleted_at IS NULL`.

2. More serious, it seems to me that even if the server and DB are both operating in UTC, the server's clock could be ahead of the DB's. My understanding (and please correct me if I'm wrong) is that soft deletes use the server's current time, and not the DB's. That seems problematic if it's coupled with using the DB's time when querying since records you intended to immediately delete won't actually be deleted until the DB's clock catches up. The solution to this would be to use `CURRENT_TIMESTAMP` when deleting if no custom delete_at is passed.

Let me know if there's a better place to continue this discussion.

PR #7291, opened 2017-02-26
Expose OptimisticLockError on Sequelize object (#6637)

Wsiegenthaler:
This PR is to expose the new OptimisticLockError introduced in #6637 so users may detect and handle conflicts accordingly. Unless I'm overlooking something, optimistic locking will not be fully functional until the error is available outside the project.

Tests have been added much like those for other errors (e.g. ValidationError).


janmeier:
:+1: thanks

Wsiegenthaler:
Thanks for getting this in!

PR #7297, opened 2017-02-28
[MSSQL] Format isolation level as tedious isolation level

### LINKED ISSUE #7296: Incorrect handling of isolationLevel for mssql Transactions
## What you are doing?
I contribute to Tedious and in a conversation on an issue (https://github.com/tediousjs/tedious/pull/493), I noticed a bug in Sequelize with respect to transaction handling with Tedious.

When I do this,
```js
sequelize.transaction({
  isolationLevel: Sequelize.Transaction.ISOLATION_LEVELS.SERIALIZABLE
});
```
Sequelize is passing the string 'SERIALIZE' for isolationLevel calling connection.beginTransaction() in mssql\query.js. This is a bug. The correct type for isolationLevel is require('tedious').ISOLATION_LEVEL enum, see http://tediousjs.github.io/tedious/api-connection.html#function_beginTransaction.

What's happening currently is that the string value is ending up as 0 for isolationLevel in token stream which is defined as NO_CHANGE in the enum.

I don't know how to setup a repro but transactions for mssql are broken currently.

Pass require('tedious').ISOLATION_LEVEL as isolationLevel parameter in call to beginTransaction.

String value being passed.

harshithkashyap:
### Pull Request check-list

### Description of change

Closes #7296

### DIFF
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -51,7 +51,7 @@ class Query extends AbstractQuery {
           } else {
             resolve(this.formatResults());
           }
-        }, this.options.transaction.name, Utils.mapIsolationLevelStringToTedious(this.options.isolationLevel));
+        }, this.options.transaction.name, Utils.mapIsolationLevelStringToTedious(this.options.isolationLevel, connection.lib));
       } else if (_.startsWith(this.sql, 'COMMIT TRANSACTION')) {
         connection.commitTransaction(err => {
           if (err) {
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -51,7 +51,7 @@ class Query extends AbstractQuery {
           } else {
             resolve(this.formatResults());
           }
-        }, this.options.transaction.name, this.options.isolationLevel);
+        }, this.options.transaction.name, Utils.mapIsolationLevelStringToTedious(this.options.isolationLevel));
       } else if (_.startsWith(this.sql, 'COMMIT TRANSACTION')) {
         connection.commitTransaction(err => {
           if (err) {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -496,3 +496,20 @@ class Where extends SequelizeMethod {
 exports.Where = Where;
 
 exports.validateParameter = parameterValidator;
+
+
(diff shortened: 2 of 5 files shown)


harshithkashyap:
@janmeier @sushantdhiman I've added a test that asserts `beginTransaction` be called with correct values.

lib/dialects/mssql/query.js
```diff
@@ -51,7 +51,7 @@ class Query extends AbstractQuery {
           } else {
             resolve(this.formatResults());
           }
-        }, this.options.transaction.name, this.options.isolationLevel);
+        }, this.options.transaction.name, Utils.mapIsolationLevelStringToTedious(this.options.isolationLevel));
```
sushantdhiman: We should prefer `this.lib` over `require('tedious')` here as dialect module path can be overridden. May be pass `this.lib` to `mapIsolationLevelStringToTedious` as second argument

lib/utils.js
```diff
@@ -496,3 +496,20 @@ class Where extends SequelizeMethod {
 exports.Where = Where;
 
 exports.validateParameter = parameterValidator;
+
+
+exports.mapIsolationLevelStringToTedious = isolationLevel => {
+  const tediousIsolationLevel = require('tedious').ISOLATION_LEVEL;
```
sushantdhiman: expect `lib` in second argument, use that `tedious` instance if available, otherwise default to `require('tedious')`.

Or may be we should just enforce passing lib instance in second argument or this method should throw, wdyt ? @harshithkashyap

lib/dialects/mssql/query.js
harshithkashyap: @sushantdhiman `this.lib` will not be available here. `this.lib` is from `ConnectionManager`.

sushantdhiman: @harshithkashyap `connection.lib` will give you that reference

lib/utils.js
harshithkashyap: @sushantdhiman I've used the latter approach. Since its a private method, we are throwing error incase tedious lib instance is not passed to Utils function.

harshithkashyap:
Ping @janmeier @sushantdhiman

PR #7328, opened 2017-03-04
restore setDataValue, fixed set issue with customGetter/Setter

sushantdhiman:
Fixed incorrect changes in `previous` method working from https://github.com/sequelize/sequelize/pull/7189#issuecomment-281325021

Actual issue that PR was trying to solve was `incorrect previous value when using custom getter / setter`. So I have added few more unit tests and a fix for that issue @nemisj 

This will also conform with the fact that `_previousDataValues` should be set in last stage of hook lifecycle.

@BorisKozo @barriber initially pointed out this issue, you guys please have a look on new changes as well

### DIFF
--- a/test/integration/model/update.test.js
+++ b/test/integration/model/update.test.js
@@ -1,21 +1,17 @@
 'use strict';
 
 /* jshint -W030 */
-var Support = require(__dirname + '/../support')
-  , DataTypes = require(__dirname + '/../../../lib/data-types')
-  , chai = require('chai')
-  , expect = chai.expect
-  , Support = require(__dirname + '/../support')
-  , current = Support.sequelize
-  , _ = require('lodash');
+const Support = require(__dirname + '/../support');
+const DataTypes = require(__dirname + '/../../../lib/data-types');
+const chai = require('chai');
+const expect = chai.expect;
+const current = Support.sequelize;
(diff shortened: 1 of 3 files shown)


test/integration/model/update.test.js
```diff
@@ -1,21 +1,17 @@
...
 
-describe(Support.getTestDialectTeaser('Model'), function() {
+describe(Support.getTestDialectTeaser('Model'), function () {
   describe('update', function () {
-
-    var Account;
-
     beforeEach(function() {
-      Account = this.sequelize.define('Account', {
+      this.Account = this.sequelize.define('Account', {
```
felixfbecker: What's the benefit of using the shared test context over a scoped variable? You don't get any help from your editor for the text context

sushantdhiman: Scope is already shared among tests, its just a good practice to keep shared data passed this way. Most modern editors support hinting variables defined this way as well, but i wasn't concerned about that.

nemisj:
that's nice, thanks @sushantdhiman  :+1:

sushantdhiman:
@nemisj Please port it to v3 :)

nemisj:
@sushantdhiman sure :)

Issue #7278, opened 2017-02-21
Postgres ENUM values can not be altered by running migrations back and forth


hal0gen:
## What you are doing?
This is similar to #1871 although not exactly the same. In addition to the ENUM not being dropped when dropping the corresponding table, the new values for the ENUM are simply ignored, with no errors or warnings about what's going on.

## Steps to reproduce:
1) create a migration with an ENUM 
```js
  up: function (queryInterface, Sequelize) {
      return queryInterface.createTable('users', { 
        id: {
          type: Sequelize.INTEGER,
          primaryKey: true,
          autoIncrement: true
        },
        test: Sequelize.ENUM('a','b') 
      });
  },
  down: function (queryInterface, Sequelize) {
      return queryInterface.dropTable('users');
  }
```

2) run `sequelize db:migrate` and then `sequelize db:migrate:undo`

3) modify the migration
```js
  up: function (queryInterface, Sequelize) {
      return queryInterface.createTable('users', { 
        id: {
          type: Sequelize.INTEGER,
          primaryKey: true,
          autoIncrement: true
        },
        test: Sequelize.ENUM('c','d') 
      });
  },
  down: function (queryInterface, Sequelize) {
      return queryInterface.dropTable('users');
  }
```

4) run `sequelize db:migrate` again

## What do you expect to happen?
It should be possible to create a `users` record with `test: 'c'`

## What is actually happening?
Creating a record with `test: 'c'` fails: 
`ERROR: invalid input value for enum enum_users_test: "c"` 

## Current workaround
Run `sequelize db:migrate:undo`, execute query `DROP TYPE "enum_users_test";` on the database and then run `sequelize db:migrate` again

__Dialect:__ postgres 
__Database version:__ 9.6.1
__Sequelize version:__ 3.30.2

ajmueller:
This is still an issue and seems to be related to #2554.

Issue #2554, opened 2014-11-10
Migration on ENUM with postgres dialect: type already exists


kytwb:
Deleting previously existing ENUM column and re-creating it but with a different ENUM results in:

```
Executing (default): ALTER TABLE "Trackers" DROP COLUMN "type";
Executing (default): CREATE TYPE "enum_Trackers_type" AS ENUM('image', 'javascript', 'snippet'); ALTER TABLE "Trackers" ADD COLUMN "type" "enum_Trackers_type";
Possibly unhandled SequelizeDatabaseError: error: type "enum_Trackers_type" already exists
```

laser:
@kytwb 

> Any idea how to correctly edit enum values through migration then?

I was able to modify an ENUM by:

1. Modifying the array assigned to the `values` property of the configuration object passed as the second argument to `sequelize.define` to include a new value
2. Create a migration in which the `up` simply calls `ModelNameHere.sync()`

That generated the SQL statements required to modify the ENUM-type to match what I'd expressed in JavaScript code.

Erin

kytwb:
@laser Tried what you said, it does work when you add only one new value at a time. 

```
Executing (default): ALTER TYPE "enum_warehouses_type" ADD VALUE 'postmaster' BEFORE 'amazon'
Executing (default): ALTER TYPE "enum_warehouses_type" ADD VALUE 'amazon' AFTER 'postmaster'
Possibly unhandled SequelizeDatabaseError: "amazon" is not an existing enum label
```

`amazon` and `postmaster` being the two new values here.

Victor08:
I am facing the same issue. I was able to bypass it with using raw query
```
module.exports = {
  up: function (queryInterface, Sequelize) {
    return queryInterface.sequelize.query("CREATE TYPE \"enum_users_sellerAccountType\" AS ENUM(\'sa\', \'sm\'); ALTER TABLE \"users\" ADD COLUMN \"sellerAccountType\" \"enum_users_sellerAccountType\";");
  },

  down: function (queryInterface, Sequelize) {
    return queryInterface.sequelize.query("DROP TYPE \"enum_users_sellerAccountType\"; ALTER TABLE \"users\" DROP COLUMN \"sellerAccountType\";");
  }
};
```
Is it a bug in sequelize, that causes enum migration to fail?

delveintechnolabs:
[...]

PR #7392, opened 2017-03-17
Fix for issue#7278 and issue#2554

okolafa:
### Pull Request check-list

- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? - not sure there is any relevant documentation that needs to be changed

### Description of change

If Postgres enum type exists, it is deleted when removing a column.

This PR solves: https://github.com/sequelize/sequelize/issues/7278 and partially: https://github.com/sequelize/sequelize/issues/2554 (this PR does not implement `dropEnum` function that was mentioned in one of  the comments).

This is my first PR on github, so don't worry to tell me my approach is wrong :)

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -254,7 +254,9 @@ const QueryGenerator = {
   removeColumnQuery(tableName, attributeName) {
     const quotedTableName = this.quoteTable(this.extractTableDetails(tableName));
     const quotedAttributeName = this.quoteIdentifier(attributeName);
-    return `ALTER TABLE ${quotedTableName} DROP COLUMN ${quotedAttributeName};`;
+    let query = `ALTER TABLE ${quotedTableName} DROP COLUMN ${quotedAttributeName};`;
+    query += this.pgEnumDrop(tableName, attributeName);
+    return query;
   },
 
   changeColumnQuery(tableName, attributes) {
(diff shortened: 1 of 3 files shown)


lib/dialects/postgres/query-generator.js
```diff
@@ -254,7 +254,9 @@ const QueryGenerator = {
   removeColumnQuery(tableName, attributeName) {
     const quotedTableName = this.quoteTable(this.extractTableDetails(tableName));
     const quotedAttributeName = this.quoteIdentifier(attributeName);
-    return `ALTER TABLE ${quotedTableName} DROP COLUMN ${quotedAttributeName};`;
+    let query = `ALTER TABLE ${quotedTableName} DROP COLUMN ${quotedAttributeName};`;
+    query += this.pgEnumDrop(tableName, attributeName);
```
janmeier: I would add a `RESTRICT` here, just to be explicitly on the safe side (to prevent dropping the enum if other types depend on it) https://www.postgresql.org/docs/8.2/static/sql-droptype.html

lib/dialects/postgres/query-generator.js
```diff
@@ -789,9 +791,9 @@ const QueryGenerator = {
     return sql;
   },
 
-  pgEnumDrop(tableName, attr, enumName) {
+  pgEnumDrop(tableName, attr, enumName, cascade) {
     enumName = enumName || this.pgEnumName(tableName, attr);
-    return 'DROP TYPE IF EXISTS ' + enumName + '; ';
+    return 'DROP TYPE IF EXISTS ' + enumName + (cascade ? ' CASCADE' : ' RESTRICT') + '; ';
```
JMTyler: @janmeier It looks like they added that `RESTRICT` you were asking for, but then no new activity before stalebot closed it.  Does this look good to you now?

PR #7394, opened 2017-03-18
add class method "increment"

### LINKED ISSUE #6359: [Suggestion] Increment fields using Model.
We can increment fields by using [instance.increment(fields, [options])](http://docs.sequelizejs.com/en/latest/api/instance/#incrementfields-options-promisethis).

But I think increment with Model is also necessary. Like:

``` js
  const fields = { answer: 42, tries: 1};
  someModel.increment(fields, {
     where: {
      id: 111,
    }
  });
```

Thanks.

hieven:
### Pull Request check-list

### Description of change

I move the instance method `increment` code to be class method `increment`.
The original instance method `increment` will simply add `this.where()` to `options.where` and then call `this.constructor.increment()`.

This is the very first open source contribution I do.
Please give me any advices to improve this PR :-)

I will write document and changelog if this PR is acceptable

close #6359

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2854,6 +2854,90 @@ class Model {
     return this.associations.hasOwnProperty(alias);
   }
 
+   /**
+   * Increment the value of one or more columns. This is done in the database, which means it does not use the values currently stored on the Instance. The increment is done using a
+   * ```sql
+   * SET column = column + X
+   * ```
+   * query. To get the correct value after an increment into the Instance you should do a reload.
+   *
+   *```js
+  * Model.increment('number') // increment number by 1
+  * Model.increment(['number', 'count'], { by: 2 }) // increment number and count by 2
+  * Model.increment({ answer: 42, tries: 1}, { by: 2 }) // increment answer by 42, and tries by 1.
+  *                                                        // `by` is ignored, since each column has its own value
+  * ```
+  *
+  * @see {@link Model#reload}
+  * @param {String|Array|Object} fields If a string is provided, that column is incremented by the value of `by` given in options. If an array is provided, the same is true for each column. If and object is provided, each column is incremented by the value given.
+  * @param {Object} [options]
+  * @param {Object} [options.where]
+  * @param {Integer} [options.by=1] The number to increment by
+  * @param {Boolean} [options.silent=false] If true, the updatedAt timestamp will not be updated.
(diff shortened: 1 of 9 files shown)


hieven:
the test fails in a strange line.
If someone could help figure this out, it would be great

lib/query-interface.js
```diff
@@ -641,18 +641,18 @@ class QueryInterface {
...
+  increment(model, tableName, values, identifier, options) {
+    options = Utils.cloneDeep(options);
 
-    options = _.clone(options) || {};
+    const sql = this.QueryGenerator.arithmeticQuery('+', tableName, values, identifier, options, options.attributes);
+
+    options.model = model;
 
-    options.type = QueryTypes.UPDATE;
-    options.instance = instance;
```
janmeier: This line is what makes `returning: true` work - It shouldn't be removed :)

test/unit/increment.test.js
```diff
@@ -0,0 +1,50 @@
...
+            dataValues: {id: 3}
+          })
+        );
+      });
+
+      after(function() {
+        stub.restore();
+      });
+
+      it('should allow increments even if options are not given', function () {
```
janmeier: We should require `where` to be defined, like we do for update, so users don't accidentally update all rows in the database

lib/model.js
```diff
@@ -2837,6 +2837,90 @@ class Model {
...
+    for (const attr of Object.keys(values)) {
+      // Field name mapping
+      if (this.rawAttributes[attr] && this.rawAttributes[attr].field && this.rawAttributes[attr].field !== attr) {
+        values[this.rawAttributes[attr].field] = values[attr];
+        delete values[attr];
+      }
+    }
+
+    if (!options.increment) {
+      return this.sequelize.getQueryInterface().decrement(this, this.getTableName(options), values, where, options).then(affectedRows => {
```
janmeier: The `then` handlers are duplicated here

hieven:
Thanks for the quick help and the test finally passes!

I'd like to receive more suggestions for improving this PR since this is the very first contribution I make for open source. :-)

janmeier:
@hieven I don't have any other comments I think, code style looks good, and I like your thoroughness in testing :)

hieven:
@janmeier 
Thank you for the comment. :-)

Issue #7184, opened 2017-02-02
Regression in affected rows reporting for updates


jharting:
## What you are doing?

Imagine having a controller that handles updates to a given resource, e.g:

```
PUT /users/493824

{
    name: "Jozef",
    surname: "Hartinger",
    notes: [{
        id: "foo",
        text: "Lorem ipsum dolor sit amet..."
    }, {
        id: "bar"
        text: "Lorem ipsum dolor sit amet..."
    }]
}
```

Note that the controller is supposed to return 404 if the resource does not exist. A different method is used for creating resources. The controller could be implemented like this:

```js
sequelize.transaction(function (transaction) {
    return User.update(user, {
        where: {
            id: user_id
        },
        transaction: transaction,
        limit: 1
    }).then(function (count) {
        if (count === 0) {
            return res.status(404).end();
        } else {
            // run a query to update/create notes
            // left out for brevity
        }
    });
});
```

The "affected rows" count is used to determine if the entity exists in the database and if it does then another query is issued to update related entities.

## What do you expect to happen?
The expectation is for the "affected rows" count value to be 1 if there is an entity in the User table with the matching id or 0 if there is none.

## What is actually happening?
This works as expected with all versions of sequelize until 3.28.0

With 3.28.0 and later the count is 0 even if the User entity exists in the database as long as no data was actually changed (the data in the update query are the same as the data already in the database).

I traced this down further and found that https://github.com/sequelize/sequelize/commit/4f95f176eba4a7ad7c71108dd48955884d16eb6b is the breaking commit. The commit addresses https://github.com/sequelize/sequelize/issues/6310

In the issue @janmeier originally talked about making this the default behavior in 4.0 which sounds reasonable to me. Unfortunatelly, the behavior was also backported to 3.28.0 where it is also enabled by default without any configuration option to switch to the old behavior. This breaks our app.

__Dialect:__ mysql
__Database version:__ 10.1.20
__Sequelize version:__ 3.28.0

janmeier:
Yeah, that was a wrong merge on my part, it should definitely be hidden behind a flag on v3.

jharting:
@janmeier does `config.dialectOptions.foundRows` sound OK to you as the name of the config property?

janmeier:
Sounds good to me

sdotson:
Any updates on this? I'm still seeing this in the latest 3.x.x version.

jharting:
I am using
```
dialectOptions: {
    flags: ''
}
```

PR #7423, opened 2017-03-24
#7184 Regression in affected rows reporting for updates

jharting:
### Pull Request check-list

### Description of change

Introduces dialectOptions.foundRows configuration option that can be set to bring back
update behavior from before this commit https://github.com/sequelize/sequelize/commit/4f95f176eba4a7ad7c71108dd48955884d16eb6b

https://github.com/sequelize/sequelize/issues/7184

### DIFF
--- a/test/integration/dialects/mysql/connector-manager.test.js
+++ b/test/integration/dialects/mysql/connector-manager.test.js
@@ -102,5 +102,20 @@ if (dialect === 'mysql') {
           return cm.releaseConnection(connection);
         });
     });
+
+    it('-FOUND_ROWS can be suppressed to get back legacy behavior', () => {
+      const sequelize = Support.createSequelizeInstance({ dialectOptions: { flags: '' }});
+      const User = sequelize.define('User', { username: DataTypes.STRING });
+
+      return User.sync({force: true})
+        .then(() => User.create({ id: 1, username: 'jozef' }))
+        .then(() => User.update({ username: 'jozef'}, {
+          where: {
+            id: 1
+          }
+        }))
+        // https://github.com/sequelize/sequelize/issues/7184
+        .spread(affectedCount => affectedCount.should.equal(1));
+    });
   });
 }


lib/dialects/mysql/connection-manager.js
```diff
@@ -82,6 +81,11 @@ class ConnectionManager extends AbstractConnectionManager {
       supportBigNumbers: true
     };
 
+    // -FOUND_ROWS used by default but can be suppressed
+    if (!config.dialectOptions || config.dialectOptions.foundRows !== false) {
```
sushantdhiman: Rather than this , you can just pass `dialectOptions.flags = '-FOUND_ROWS'` ?

jharting: @sushantdhiman `-FOUND_ROWS` is the default behavior so there's never a need to pass it in this way. What I am trying to introduce is a way to suppress that flag.

If I did it your way the usecase would be for people to specify `dialectOptions.flags = ''`for the default flag to be suppressed. Both ways sound goo to me.

@janmeier we talked about the name of this config property in the past. WDYT about `config.dialectOptions.flags=""` vs `config.dialectOptions.foundRows = false`?

janmeier: Both sound good to me as well, lets go with `flags: ''`

jharting: @janmeier @sushantdhiman Apologies for the delay. I switched to `dialectOptions.flags` which effectively means this PR only adds a test. This is now ready for review again.

changelog.md
```diff
@@ -73,6 +73,7 @@
 - [ADDED] Support for JSON attributes in orders and groups. [#7564](https://github.com/sequelize/sequelize/issues/7564)
 - [REMOVED] Removes support for interpretation of raw properties and values in the where object. [#7568](https://github.com/sequelize/sequelize/issues/7568)
 - [FIXED] Upsert now updates all changed fields by default
+- [FIXED] Regression in affected rows reporting for updates [#7184](https://github.com/sequelize/sequelize/issues/7184)
```
sushantdhiman: Remove this entry then

jharting:
@janmeier @sushantdhiman Should be all fixed now

Issue #3989, opened 2015-06-23
Does sequelize support MySQL's timestamp data type?


axln:
I have existing MySQL database which has timestamp columns. How can I work with them?

I know that sequelize always stores UTC time in datatime columns, but I can't find any info on how to work with timestamp data type.

itsravenous:
I would suggest reading #3512 - there's a good example there of adding your own custom data types (I used it to add the old MSSQL DATETIME type to my app, as Sequelize uses DATETIME2 for MSSQL out of the box)

axln:
Great! Thank you very much. Will try it and reports the results.

janmeier:
@axln There has been talk about supporting timestamp before. As you have already seen, it should be pretty easy to add the datatype, and we'd welcome a PR adding a `TIMESTAMP` datatype.

For support to be complete we'd also need a way to change the `createdAt` / `deletedAt` times to use timestamp instead of date

janmeier:
https://github.com/sequelize/sequelize/issues/963

axln:
I finally implemented the basic TIMESTAMP support using the method suggested by itsravenous.

Here is my code:

lib/sequelize-mysql-timestamp.js:

```javascript
'use strict';

var BaseTypes = require('sequelize/lib/data-types');
var util = require('util');

var TIMESTAMP = function() {
  if (!(this instanceof TIMESTAMP)) {
    return new TIMESTAMP();
  }

  BaseTypes.ABSTRACT.apply(this, arguments);
};

util.inherits(TIMESTAMP, BaseTypes.ABSTRACT);

TIMESTAMP.prototype.key = TIMESTAMP.key = 'TIMESTAMP';

exports.TIMESTAMP = TIMESTAMP;
```

In models we can now use (assumed that modes are located in the models subdirectory):
```javascript
'use strict';

var customDataTypes = require('../lib/sequelize-mysql-timestamp');

module.exports = function(sequelize, DataTypes) {
  var Sample = sequelize.define('Sample', {
    title: DataTypes.STRING,
    created: customDataTypes.TIMESTAMP
  }, {
    tableName: 'sample',
    timestamps: false
  });
  return Sample;
};
```

itsravenous:
Glad it helped :)

axln:
@itsravenous, thank you very much for your help!

oaeide:
Is there any particular reason for this not being included?

mickhansen:
[...]

PR #7432, opened 2017-03-26
V3 MSSQL rowversion datatype

iamjoyce:
### Pull Request check-list

### Description of change

This PR includes support for `rowversion` column type for **MSSQL only**. New test cases only covers MSSQL.

It also supports scenarios where table's options specify `hasTrigger: true` - As `timestamp/rowversion` is an auto-generated column in MSSQL, [query-generator.js](https://github.com/sequelize/sequelize/pull/7432/files#diff-1) will not create columns for temp table if the column type is `Sequelize.VIRTUAL` or `Sequelize.ROWVERSION`.

### Points to Note:

Newer versions of MSSQL suggests `rowversion` over `timestamp`, but this PR uses `timestamp` in SQL statement instead to support older versions of MSSQL.

###  Question

~Will it be better to use `Sequelize.ROWVERSION` instead of `Sequelize.TIMESTAMP` as variable name? In case existing systems have its own custom TIMESTAMP. Just that during SQL generation, the SQL statement should still use `timestamp` instead of `rowversion`.~

Changed TIMESTAMP to ROWVERSION.

### DIFF
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -798,6 +798,15 @@ ENUM.prototype.validate = function(value) {
 };
 
 /**
+ * A rowversion column (time-based sequential number). Only available in mssql.
+ * @property ROWVERSION
+ */
+
+var ROWVERSION =  ABSTRACT.inherits();
+
+ROWVERSION.prototype.key = ROWVERSION.key = 'ROWVERSION';
+
+/**
  * An array of `type`, e.g. `DataTypes.ARRAY(DataTypes.DECIMAL)`. Only available in postgres.
  * @property ARRAY
  */
@@ -972,6 +981,7 @@ var dataTypes = {
   ARRAY: ARRAY,
   NONE: VIRTUAL,
   ENUM: ENUM,
+  ROWVERSION: ROWVERSION,
   RANGE: RANGE,
   REAL: REAL,
   DOUBLE: DOUBLE,
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -390,5 +390,23 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
         });
     });
   }
+  
+  if (dialect === 'mssql') {
+    it('should parse ROWVERSION field', function () {
+      var Type = new Sequelize.ROWVERSION();
 
+      var User = current.define('user', { name: Sequelize.STRING, version: Type }, { timestamps: false });
+
+      return current.sync({ force: true }).then(function () {
+        return User.create({
+           name: 'test'
+        });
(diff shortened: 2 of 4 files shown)


iamjoyce:
Is there anything I can do to get this PR approved?

sushantdhiman:
@iamjoyce Please get rid of autogenerated documentation changes

test/integration/data-types.test.js
```diff
@@ -145,6 +145,25 @@ describe(Support.getTestDialectTeaser('DataTypes'), function() {
       return testSuccess(Type, [1, 2]);
     });
   }
+  
+  if (dialect === 'mssql') {
+    it.only('should parse TIMESTAMP field', function () {
```
sushantdhiman: `.only` ??

lib/data-types.js
```diff
@@ -503,6 +503,22 @@ DATEONLY.prototype.toSql = function() {
 };
 
 /**
+ * A timestamp column. Only available in mssql.
```
sushantdhiman: Move this inside MSSQL datatype.js if others dont support it

iamjoyce: Will it be better to use Sequelize.ROWVERSION instead of Sequelize.TIMESTAMP? I think some projects are using TIMESTAMP as well, but it's for other dialects, and that data type can be modified unlike in MSSQL. [Example for MySQL](https://github.com/sequelize/sequelize/issues/3989)

test/integration/data-types.test.js
iamjoyce: Thank you for picking that up. I will remove this; I was doing some final debugging and missed that out.

lib/data-types.js
sushantdhiman: Yeah, It looks like this "TIMESTAMP" is not exactly a timestamp, it a time based sequential number. Its better to have different name so it doesn't confuse.

We will go with "ROWVERSION", as per your PR description we still need `timestamp` name in SQL queries, `toSQL` can handle that

Still we need to move it MSSQL code base, because other dialect doesn't support "ROWVERSION", but they do support "TIMESTAMP" which we can introduce later on

iamjoyce:
@sushantdhiman PR for master is at #7475

iamjoyce:
@sushantdhiman PR for master is at #7476

iamjoyce:
ping @sushantdhiman

sushantdhiman:
[v4](http://docs.sequelizejs.com/manual/changelog/changelog.html#4-0-0-final-) has been released and we have decide `v3` will only accept critical bug fixes and security updates, no more features will be added to `v3`

Issue #7404, opened 2017-03-21
Passing parameters to model getters


jharting:
This is a feature request. There may already be a way to achieve what I am trying to do in which case I'll be grateful for pointers.

## What you are doing?
I am implementing a versioned API. Assume the following simple example:

```
GET /v1/places

200 OK
[{
    name: "Luigi's Pizza Place",
    rating: 84
}, {
    name: "Greek Taverna",
    rating: 78
}]
```

The rating field is implemented as an INTEGER in the db and expected to have values 0-99

Now assume that for v2 of the API we need to change how we present rating so that it instead of 0-99 it returns 0-4 floating point number rating.

We can implement that easily with model a getter:

```js
getterMethods: {
    rating: function () {
        return this.getDataValue('rating') / 20;
    }
}
```

So far so good.

The problem now is how to preserve backwards compatibility in v1 but at the same time support new representation in v2. IMHO the cleanest would be to do that in the getter, e.g:

```js
getterMethods: {
    rating: function () {
        const value = this.getDataValue('rating');

        if (apiVersion === 1) {
            return value;
        }

        return value / 20;
    }
}
```

Now what I am struggling with is how to actually pass the apiVersion down to model getters. I can obviously get that value from some global constant but that can get tricky as execution of different requests may overlap.

The cleanest approach IMO would be for Sequelize to support passing in query context. Here's an example:

```js
sequelize.places.findAll({
    limit: 5
},
{ // this is a context object
    apiVersion: req.param.version
})
```

now the getter could be written as:

```js
getterMethods: {
    rating: function (param, context) {
        const value = this.getDataValue('rating');

        if (context.apiVersion === 1) {
            return value;
        }

        return value / 20;
    }
}
```

WDYT? Is there already a nice way to implement this that I am missing? If not would the aforementioned proposal make sense in Sequelize?

janmeier:
I think the options object passed to `get` will actually be passed down to each getter function http://docs.sequelizejs.com/en/v3/api/instance/#getkey-options-objectany (i.e. you would do 

```js
places.findAll().then(places => res.json(places.map(places.get({ apiVersion: 42 })))
```

If not, you could look into using cls or domains to maintain a context

jharting:
@janmeier thanks for the suggestion. It unfortunately does not seem to be the case. When I call
```
place.get({apiVersion: 3});
```

the rating getter is called with `"rating"` as a param.

Mind if I change the getter logic for this to work?

[...]

PR #7435, opened 2017-03-27
#7404 make it possible to pass parameters to getter functions

jharting:
### Pull Request check-list

### Description of change

Makes it possible for model getters to behave differently in different contexts. For example, a model definition may need to be different for different version of a RESTful API.

```js
rating: function (key, options) {
  const value = this.getDataValue('rating'); // rating is stored as integer with values 1-100

  if (context.apiVersion === 1) {
      return value / 20; // in /v1 we use 1-5 rating system
  }

  return value;  // in /v2 and newer we use 1-100 point rating system
}
```

the parameter is passed like this:

```js
product.get({plain: true, apiVersion: req.param.apiVersion}
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -60,6 +60,7 @@
 - [FIXED] Deleted paranoid records can be queried in the same second. [#7204](https://github.com/sequelize/sequelize/issues/7204)/[#7332](https://github.com/sequelize/sequelize/pull/7332)
 - [FIXED] `removeAttribute('id')` results in `undefined: null` data value [#7318](https://github.com/sequelize/sequelize/issues/7318)
 - [FIXED] `bulkCreate` now runs in O(N) time instead of O(N^2) time. [#4247](https://github.com/sequelize/sequelize/issues/4247)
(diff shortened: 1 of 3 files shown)


janmeier:
:+1:

UKNC:
Hi! Is there an entry in docs which describe this feature? Thanks

xtrimsky:
Not actual documentation but for anyone reading this:

`plain: true` seems to be required when using this.

Then the getter needs to be built this way:

```
getterMethods: {
   height: function (key, options) {
      if (options.apiVersion > 1) {
         return 185; // cm
       }

      return 6.06; // ft
   }
 }
```

And then we can call the model passing the parameter:

```
expect(product.get({plain: true, apiVersion: 1})).to.have.property('rating', 5);
```

Full example given here:
https://github.com/sequelize/sequelize/pull/7435/files/930a5b63307486f9cc8abcaeef028c0c6bb48e59#diff-057c4fdbe2adf868c990ea7adfa755dd

PR #7417, opened 2017-03-23
V3 options.where are erased when a belongsToMany is called

ggrimbert:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [X] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [ ] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [X] Have you added new tests to prevent regressions?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Have you added an entry under `Future` in the changelog?

In some findAll using belongsToMany and offset, the subquery is generated but not applied to the final request, leading to bad results.

### Description of change

The where generated for subquery in ```generateJoinQueries``` erase the options.where attribute.
This PR adds it if a where condition is already declared.

PR #7436, opened 2017-03-27
V3 options.where are erased when a belongsToMany is called

ggrimbert:
### Pull Request check-list

In some findAll using belongsToMany and offset, the subquery is generated but not applied to the final request, leading to bad results.

### Description of change

The where generated for subquery in ```generateJoinQueries``` erase the options.where attribute.
This PR adds it if a where condition is already declared.

A PR was already created (https://github.com/sequelize/sequelize/pull/7417), but some merging problems forced me to delete / recreate the repository.

Just waiting for code review to create the same PR in master.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1320,12 +1320,18 @@ var QueryGenerator = {
                   }, topInclude.model);
                 }
 
-                options.where['__' + throughAs] = self.sequelize.asIs([
+                var queryWhere = self.sequelize.asIs([
                   '(',
                     $query.replace(/\;$/, ''),
                   ')',
                   'IS NOT NULL'
                 ].join(' '));
+
+                if (_.isPlainObject(options.where)) {
+                  options.where['__' + throughAs] = queryWhere;
+                } else {
+                  options.where = { $and: [options.where, queryWhere] };
+                }
               })(include);
             }
           }
--- a/test/integration/associations/belongs-to-many.test.js
+++ b/test/integration/associations/belongs-to-many.test.js
@@ -410,6 +410,101 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
     });
   });
 
+
+  it('supports where in include with required', function () {
+      var User = this.sequelize.define('User', {
+        id: {
+          type: DataTypes.INTEGER,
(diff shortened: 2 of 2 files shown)


test/integration/associations/belongs-to-many.test.js
```diff
@@ -408,6 +408,99 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), function() {
...
+          return Promise.join(
+            user.setCompany(company),
+            company.addGroup(group),
+            company.addGroup(group2),
+            company2.addGroup(group),
+            company2.addGroup(group2)
+          );
+        }).then(function () {
+          return Promise.join(
+            Company.scope([{
```
janmeier: That's not how scopes should be used - You add a scope to the model (in the model definition or using `addScope`), and then invoke the scope by name using `scope`

http://docs.sequelizejs.com/en/v3/docs/scopes/

ggrimbert: As we define dynamic scopes, we never used the ```addScope```, we just passed an array to scopes().

Change is done, is that ok ?

ggrimbert:
Any news here ?

lib/dialects/abstract/query-generator.js
```diff
@@ -1320,12 +1320,18 @@ var QueryGenerator = {
...
 
-                options.where['__' + throughAs] = self.sequelize.asIs([
+                var queryWhere = self.sequelize.asIs([
                   '(',
                     $query.replace(/\;$/, ''),
                   ')',
                   'IS NOT NULL'
                 ].join(' '));
+
+                if (_.isPlainObject(options.where)) {
```
felixfbecker: why only for "plain" objects and not for all objects?

ggrimbert: As seen in lodash's doc, isPlainObject works for {}, isObject works for [], but I don't want arrays to be treated like an object. 
So isPlainObject, seems to be the good solution.

ggrimbert:
Unfortunately, the PR for master is inside my PR for Oracle. 
When I worked on Oracle support, I needed this fix, so I integrate it and validate the "non creating" master PR with @janmeier.

sushantdhiman:
V3 is locked so I am closing this PR

PR #7449, opened 2017-03-29
Explained in more detail what sequalize.col does

actuallymentor:
Template skipped (documentation only).

Based on [this stackoverflow](https://stackoverflow.com/questions/30452977/sequelize-query-compare-dates-in-two-columns) I added some more explanation as to what sequelize.col does.

Also, changed 'a object' to 'an object' because grammar. Couldn't resist.


janmeier:
The files in `docs/api` are automagically generated - Could you please change the code comments instead? (I know this can be confusing, and we are moving away from generated files in the repo very soon :) )

actuallymentor:
I reverted the initial commit used for this PR and moved the changes to the .js file.

janmeier:
Hmm, there are some other changes in the diff as well now - Maybe you reverted too much, or need a rebase?

actuallymentor:
I reverted to the wrong commit and went down into the rabbit hole of trying to fix it. I'll close this one and just do it over :)

PR #7454, opened 2017-03-29
7425 Association identifiers in ordering and groups

### LINKED ISSUE #7425: Ordering on Associated Models is Broken
## What you are doing?

```js
Model.findAll({
	attributes: [
		'id',
		'serialNumber'
	]
	where: {
		serialNumber: {
			'$ilike': 'B______'
		}
	},
	include: [
		{
			association: Model.associations.BelongsToModel,
			required: true,
			attributes: [
				'name'
			]
		}
	],
	limit: '1000',
	order: [
		[
			'BelongsToModel.name',
			'ASC'
		],
		[
			'BelongsToModel.name'
		],
		'BelongsToModel.name',
		[
			'Model.serialNumber',
			'ASC'
		],
		[
			'Model.serialNumber'
		],
		'Model.serialNumber',
		[
			'serialNumber',
			'ASC'
		],
		[
			'serialNumber'
		],
		'serialNumber'
	]
});
```

```js
SELECT
	"Model"."id",
	"Model"."serial_number" AS "serialNumber",
	"BelongsToModel"."id" AS "BelongsToModel.id",
	"BelongsToModel"."name" AS "BelongsToModel.name"
FROM
	"public"."device" AS "Model"
	INNER JOIN "public"."belongs_to_model" AS "BelongsToModel" ON "Model"."belongs_to_model_id" = "BelongsToModel"."id"
WHERE
	(
		("Model"."deleted_at" >= CURRENT_TIMESTAMP OR "Model"."deleted_at" IS NULL)
		AND "Model"."serial_number" ILIKE 'B______'
	)
ORDER BY
	"BelongsToModel"."name" ASC,
	"BelongsToModel"."name",
	"BelongsToModel"."name",
	"Model"."serialNumber" ASC,
	"Model"."serialNumber",
	"Model"."serialNumber",
	"serial_number" ASC,
	"serial_number",
	"serial_number"
LIMIT '1000';
```

The main table name was prefixed to the orders.

```sql
SELECT
	"Model"."id",
	"Model"."serial_number" AS "serialNumber",
	"BelongsToModel"."id" AS "BelongsToModel.id",
	"BelongsToModel"."name" AS "BelongsToModel.name"
FROM
	"public"."device" AS "Model"
	INNER JOIN "public"."belongs_to_model" AS "BelongsToModel" ON "Model"."belongs_to_model_id" = "BelongsToModel"."id"
WHERE
	(
		("Model"."deleted_at" >= CURRENT_TIMESTAMP OR "Model"."deleted_at" IS NULL)
		AND "Model"."serial_number" ILIKE 'B______'
	)
ORDER BY
	"Model"."BelongsToModel"."name" ASC,
	"Model"."BelongsToModel"."name",
	"BelongsToModel"."name",
	"Model"."Model"."serialNumber" ASC,
	"Model"."Model"."serialNumber",
	"Model"."serialNumber",
	"Model"."serial_number" ASC,
	"Model"."serial_number",
	"serialNumber"
LIMIT '1000';
```

__Dialect:__ any
__Database version:__ any
__Sequelize version:__ 4.0.0-2

mkaufmaner:
Closes #7425 

### Pull Request check-list

### Description of change
Since order strings are no longer interpreted as literals we need a way of referencing association attributes via their alias without including them in the select clause.  Here, we deprecate the utility function for mapping attributes to their aliases and add a closure to the `getQueryOrders` method in the abstract query generator.  The closure method attempts to construct a path to the association  from the by referencing either the base model's, or previous association model's, associations.


docs/docs/querying.md
```diff
@@ -287,11 +288,18 @@ something.findOne({
     // Will order by name on an associated User
     [User, 'name', 'DESC'],
 
+    // Will order by name on an associated User using the attribute's association alias
+    ['User.name', 'DESC'],
```
felixfbecker: Why do we need this syntax when we have the above syntax?

mkaufmaner: The whole idea of this is to allow a JSON object to be passed into without reference to the models themselves.  For example, before the literal was removed from ordering you could reference the column you want to sort without having to wrap the object in a Sequelize method.  This also adds the functionality of being able to pass the alias of an attribute to the order instead of the field name.  This functionality also gives the developer the ability to pass JSON straight from an HTTP request in the order property without having to do sanitation or manipulation.

An old example may be would look something like,
```javascript
order: [
  ['"Task.Project"."created_at"', 'DESC']
]
```
but this now prefixes the model name to the order clause `"Subtask"."Task.Project"."created_at" DESC`

This feature not only fixes that but brings back that same functionality in a cleaner and more effective manner.

felixfbecker: For `include` we support this by allowing to pass the `association` property. It can either be an Association object or a string with the association alias. I would prefer to use this syntax for ordering too, as an alternative to `{ model, as }`.
Using a string with its own SQL-like syntax (`'User.name'`) is what we want to move away from, it is what an ORM should abstract away.

mkaufmaner: I understand.  So ['Task', 'Project', 'createdAt', 'DESC'] where the first items of the array are either a model, association, name of an association

mkaufmaner:
I have made significant changes.  While covering most use cases, I have broken some because I deprecated the use of {raw:} as well.  Please look at https://github.com/mkaufmaner/sequelize/blob/49674543de6c74b8b0b5712a67a3a1ec8f086378/lib/dialects/abstract/query-generator.js#L623

mkaufmaner:
Guessing it needs more unit/integration tests.

mkaufmaner:
@felixfbecker Looks like all the checks have passed.  Special thanks the jan for rerunning jobs.  Awaiting review and merge. 🍻🍻🍻🍻

sushantdhiman:
This will need a rebase before merging as there are lots of unwanted commits at the moment

docs/docs/querying.md
```diff
@@ -284,22 +284,38 @@ something.findOne({
...
 
-    // Will order by name on an associated User aliased as Friend
-    [{model: User, as: 'Friend'}, 'name', 'DESC'],
+    // Will order through an associated model's created_at using the model names as the associations' names.
+    [Task, Project, 'createdAt', 'DESC'],
 
-    // Will order by name on a nested associated Company of an associated User
-    [User, Company, 'name', 'DESC'],
+    // Will order by an associated model's created_at using the name of the association.
+    ['Task', 'createdAt', 'DESC'],
```
sushantdhiman: Can we query JSON field now ? Like

```js
['Task','jsonField.level1.level2', 'DESC']
```

Would love so see a unit test around this as well

felixfbecker:
@sushantdhiman no rebase needed, we always squash & merge.

felixfbecker:
Merge conflicts

docs/docs/querying.md
mkaufmaner: @sushantdhiman I am not sure what you are asking for, can you give me an example of the query you are trying to create?

mkaufmaner: The order query would look like;
```sql
ORDER BY "Task"."jsonField.level1"."level2" DESC
```

Simply put, there is a method `quoteIdentifiers` that takes a string with a "." in it, splits it two at the ".", and then wraps each piece of the string with identifier ticks.  It may actually be a good idea to deprecate this but it will cause confusion when upgrading from 3.

mkaufmaner:
@sushantdhiman Leaving for you to press the big green button on this.

docs/docs/querying.md
mkaufmaner: @sushantdhiman I understand, you want to be able to sort by json property similar to how where clauses are constructed for next keys (http://docs.sequelizejs.com/en/latest/docs/querying/#jsonb)

sushantdhiman:
@felixfbecker @mkaufmaner So we solved inability to handle nested josn keys ?

mkaufmaner:
@sushantdhiman No.  However, I am almost done with adding that feature.  Looks like I should create a new branch.

sushantdhiman:
👍 Cool

PR #7461, opened 2017-03-30
Make Model#$expandAttributes respect scope attributes.exclude

re5et:
The treatment of attributes.excludes does not respect scopes, which is unexpected.  This also leads to situations where you might accidentally expose information that you do not want to.

This makes Model#$expandAttributes respect scope attributes.exclude

## What are you doing?
```js
var User = this.sequelize.define('user', {
  password: Sequelize.STRING,
  firstName: Sequelize.STRING,
  lastName: Sequelize.STRING,
}, {
  defaultScope: {
    attributes: {
      exclude: ['password']
    }
  }
})

User.create({
  firstName: 'John',
  lastName: 'Doe',
  password: 'abc123',
}).then((createdRecord) => {
  User.findOne({
    where: {
      id: createdRecord.id,
    },
    attributes: {
      exclude: ['lastName'],
    }
  }).then((user) => console.log(user.get({ plain: true })))
})
```

The exclude specified in scope is respected:
```
// log output
{
  firstName: 'John'
}
```

The exclude specified in scope is ignored:
```
// log output
{
  firstName: 'John',
  pasword: 'abc123'
}
```

## Which dialect you are using (postgres, mysql etc)?

any

## Which sequelize version you are using?

v3.30.4

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2647,6 +2647,9 @@ Model.prototype.$getDefaultTimestamp = function(attr) {
 Model.prototype.$expandAttributes = function (options) {
   if (_.isPlainObject(options.attributes)) {
     var attributes = Object.keys(this.rawAttributes);
+    if (Array.isArray(this.$scope.attributes)) {
+      attributes = this.$scope.attributes;
+    }
 
     if (options.attributes.exclude) {
       attributes = attributes.filter(function (elem) {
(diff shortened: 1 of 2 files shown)


overlookmotel:
Looks like should be reopened. @sushantdhiman had said "LGTM".

felixfbecker:
@overlookmotel I disagree. It targets v3, and @sushantdhiman said it needs a PR to master first. The PR to master was not opened for over 3 months.

overlookmotel:
@felixfbecker OK sure no problem. I'm not invested in this. Just got messaged by the bot for some reason so replied...

PR #7466, opened 2017-03-31
mysql2 is not compatible with mysql: it uses strings to encode large …

jaredjenkins:
This addresses this issue: https://github.com/sequelize/sequelize/issues/7465


sushantdhiman:
Closing as per reason given in https://github.com/sequelize/sequelize/issues/7465#issuecomment-290921166

Can reopen once a solution for problem I presented has been suggested

PR #7476, opened 2017-04-02
#7432 MSSQL rowversion datatype

iamjoyce:
### Pull Request check-list

### Description of change

This PR includes support for `rowversion` column type for **MSSQL only**. New test cases only covers MSSQL.

It also supports scenarios where table's options specify `hasTrigger: true` - As `timestamp/rowversion` is an auto-generated column in MSSQL, [query-generator.js](https://github.com/sequelize/sequelize/pull/7432/files#diff-1) will not create columns for temp table if the column type is `Sequelize.VIRTUAL` or `Sequelize.ROWVERSION`.

Refer to #7432 for version 3's PR. Creating a new PR because there were some problems with #7475.

### Points to Note:

Newer versions of MSSQL suggests `rowversion` over `timestamp`, but this PR uses `timestamp` in SQL statement instead to support older versions of MSSQL.

### DIFF
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -467,5 +467,26 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
         expect(record.stamp).to.be.eql(testDate);
       });
   });
+  
+  if (dialect === 'mssql') {
+    it('should parse ROWVERSION field', function () {
+      const Model = this.sequelize.define('model', {
+        name: Sequelize.STRING,
+        version: Sequelize.ROWVERSION
+      });
 
+      const sampleData = {
+        name: 'test'
+      };
+
+      return Model.sync({ force: true }).then(() => {
+        return Model.create(sampleData);
+      }).then(() => {
+        return Model.findAll().get(0);
+      }).then(model => {
+        expect(model.get('version')).to.be.ok;
+        expect(typeof model.get('version')).to.be.eql('object');
+      });
+    });
+    
 });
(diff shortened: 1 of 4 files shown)


iamjoyce:
ping @sushantdhiman

test/integration/data-types.test.js
```diff
@@ -460,5 +460,26 @@ describe(Support.getTestDialectTeaser('DataTypes'), function () {
...
+        name: 'test'
+      };
 
+      return Model.sync({ force: true }).then(() => {
+        return Model.create(sampleData);
+      }).then(() => {
+        return Model.findAll().get(0);
+      }).then(model => {
+        expect(model.get('version')).to.be.ok;
+        expect(typeof model.get('version')).to.be.eql('object');
```
sushantdhiman: Why version is an object ? I thought it would be a number (sequential based on timestamp) ?

iamjoyce: It's not really a number in MSSQL. It's an object type. If I don't remember wrongly, the value is in Binary.

>It is not possible to convert the TimeStamp colunm into a Date & Time format. Microsoft has renamed the TimeStamp data type to RowVersion to avoid the confusion.

> The timestamp/rowversion data type is really just a Database Transaction ID, and it can also be stored in a column of the Binary(8) data type.

[Source, albeit not a very credible one](https://dba.stackexchange.com/questions/98312/how-to-convert-timestamp-rowversion-to-a-date-and-time-in-sql-server)

iamjoyce: Sorry, I see why you understood it as a time-based number. I was going to fix the comments, but I realise the comments for data-types are all gone. This should be the correct comment.

```
/**
 * A rowversion column. Only available in mssql.
 * An incrementing, unique binary number that does not does not preserve a date or a time.
 * @property ROWVERSION
 */
```

PR #7490, opened 2017-04-04
Change documentation to esdoc

janmeier:
# Why?

* We don't want to keep the generated API docs in the repo
* We (I) don't want to maintain a custom script to generate the docs

Now, docs will be continously deployed to surge.sh (static hosting) (I've added redirects for the old URLs, see `docs/ROUTER`). The v3 docs will stay on RTD, with a redirect from `/en/v3` to the RTD site, so for v3 users there will be no changes. Once this is merged, I'm also going to add notices about the new docs on the v3 docs.

This will mean you can only view the docs for the most recent version. We will probably push the docs to a subdomain (https://v4.docs.sequelizejs.com etc.) once that becomes a problem.


janmeier:
Temporary domain for anyone interested jagged-leather.surge.sh

lib/associations/base.js
```diff
@@ -1,15 +1,104 @@
 'use strict';
 const AssociationError = require('./../errors').AssociationError;
 
+
+/**
```
janmeier: Taken from lib/associations/mixin

lib/associations/belongs-to-many.js
```diff
@@ -196,126 +210,15 @@ class BelongsToMany extends Association {
     const singular = Utils.uppercaseFirst(this.options.name.singular);
 
     this.accessors = {
-      /**
```
janmeier: Moved down to the actual functions (esdoc doesn't understand this)

lib/associations/mixin.js
```diff
@@ -7,104 +7,7 @@ const HasMany = require('./has-many');
...
- * In the example above we have specified that a user belongs to his profile picture. Conceptually, this might not make sense, but since we want to add the foreign key to the user model this is the way to do it.
- *
- * Note how we also specified `constraints: false` for profile picture. This is because we add a foreign key from user to picture (profilePictureId), and from picture to user (userId). If we were to add foreign keys to both, it would create a cyclic dependency, and sequelize would not know which table to create first, since user depends on picture, and picture depends on user. These kinds of problems are detected by sequelize before the models are synced to the database, and you will get an error along the lines of `Error: Cyclic dependency found. 'users' is dependent of itself`. If you encounter this, you should either disable some constraints, or rethink your associations completely.
- *
- * @namespace Associations
- * @name Associations
- */
 const Mixin = {
-  /**
-   * Creates a 1:m association between this (the source) and the provided target. The foreign key is added on the target.
```
janmeier: Moved to methods on model.js (esdoc doesn't like mixins)

lib/errors.js
```diff
@@ -1,398 +1,3 @@
 'use strict';
 
-/**
```
janmeier: moved to lib/errors/index.js, to get a more reasonable separation of the classes in the api ref (subfolder = new section)

mkaufmaner:
This would introduce a number of broken links.

janmeier:
@mkaufmaner I'm trying to figure out how to do a wildcard redirect with surge now

PR #7492, opened 2017-04-05
Fix `showIndex()` query regexp

nunofgs:
### Pull Request check-list

### Description of change

Fixes an issue when calling `showIndex()` on an index that contains a newline in its "fields". Example:

```sql
CREATE INDEX "foobar" ON "users" USING btree ((
CASE (settings->>'type')
    WHEN 'foo' THEN 'biz'
    WHEN 'bar' THEN 'baz'
    ELSE 'qux'
END)) WITH (fillfactor='70')
```

The issue is with a regexp that is matching against `.*` (which doesn't match newlines). This PR replaces that expression with `[^]*` which matches any character that is not _nothing_.

The error that was previously seen when a newline was present was:

```
TypeError: Cannot read property '1' of null
  at node_modules/sequelize/lib/dialects/postgres/query.js:139:85
  at Array.forEach (native)
  at node_modules/sequelize/lib/dialects/postgres/query.js:135:15
```


sushantdhiman:
@nunofgs Needs changelog entry as well

PR #7494, opened 2017-04-05
'clarifying the return value of findOrCreate'

strangefrond:
The documentation of findOrCreate previously had no precise indication of what to expect findOrCreate to return. I added that it returns an array with two parts in the comments that had already existed in the examples. I also fixed an inconsistency where it looked as if the first item would be returned anonymously and the second would be returned as a key/value pair of 'created' and a boolean. In fact, both are returned anonymously.

I only changed the documentation.

### Pull Request check-list

### Description of change

### DIFF
--- a/docs/docs/models-usage.md
+++ b/docs/docs/models-usage.md
@@ -43,14 +43,18 @@ User
     console.log(created)
 
     /*
-      {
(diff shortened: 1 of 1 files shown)


janmeier:
Welcome, and thanks for you contribution  @strangefrond :)

PR #7493, opened 2017-04-05
[v3] COUNT on HasMany association is not ambiguous anymore

yachaka:
### Description of change

When counting on an `HasMany` assocation, the column `id` can be ambiguous if an other model is included. This fixes it by appending the target model name.

This should also fix #6488.

PR #7496, opened 2017-04-06
[master] COUNT on HasMany association is not ambiguous anymore

### LINKED ISSUE #6488: countAssociations with "include" in where clause return "Column 'id' in field list is ambiguous"
## What you are doing?

I want to count associations by using countAssociations of Has Many association

```
Service.hasMany(models.Banner, {
                        constraints: false,
                        foreignKey: {
                            name: 'serviceId'
                        }
                    });

Banner.belongsTo(models.Position, {
                        foreignKey: {
                            name: 'positionId'
                        }
                    });
```

``` js
service.countBanners({include: [
    {
        model: models.Position,
        where: condition.Position
    }
]}).then(function(c){
        console.log(c);
        return c;
});
```

The number of associations

I got an error "Column 'id' in field list is ambiguous"
## How to fix?

In lib/associations/has-many.js:

``` js
HasMany.prototype.count = function(instance, options) {
  var association = this
    , model = association.target
    , sequelize = model.sequelize;

  options.attributes = [
    [sequelize.fn('COUNT', sequelize.col(model.primaryKeyField)), 'count']
  ];
  options.raw = true;
  options.plain = true;

  return this.get(instance, options).then(function (result) {
    return parseInt(result.count, 10);
  });
};
```

replace 

```
    [sequelize.fn('COUNT', sequelize.col(model.primaryKeyField)), 'count']
```

with

```
    [sequelize.fn('COUNT', sequelize.col([model.name, model.primaryKeyField].join('.'))), 'count']
```

**Dialect:** mysql
**Database version:** 5.0.12
**Sequelize version:** 3.24.1

yachaka:
### Description of change

When counting on an `HasMany` assocation, the column `id` can be ambiguous if an other model is included. This fixes it by appending the target model name.

This should also fix #6488.
Master port of #7493.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -65,6 +65,7 @@
 - [FIXED] Model.validate runs validation hooks by default [#7182](https://github.com/sequelize/sequelize/pull/7182)
 - [ADDED] Added support for associations aliases in orders and groups. [#7425](https://github.com/sequelize/sequelize/issues/7425)
 - [REMOVED] Removes support for `{raw: 'injection goes here'}` for order and group. [#7188](https://github.com/sequelize/sequelize/issues/7188)
+- [FIXED] `.count` for `HasMany` association now counts on `Model.primaryKey` instead of `primaryKey` [#6488](https://github.com/sequelize/sequelize/issues/6488)
 
 ## BC breaks:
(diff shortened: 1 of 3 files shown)


yachaka:
@sushantdhiman Port to master

PR #7523, opened 2017-04-14
Fix upsert behavior to update all _changed_ fields by default

louisremi:
Hello,

I recently tried to use the `upsert` method on a model that had a virtual field 'linked' to another field in the model:
```javascript
const Meta = sequelize.define('Meta', {
  virtualValue: {
    type: DataTypes.VIRTUAL,
    set(val) { return this.value = val; }
    get() { return this.value; }
  },
  value: DataTypes.STRING
});

Meta.upsert({
  virtualValue: 'Grumpy cat'
});
```

I discovered that the **`insert`** operation would correctly result in `'Grumpy cat'` being stored in the `value` field, but an **`update`** operation wouldn't update that field. This looks a lot like a bug to me.

This PR fixes that bug by simply updating *changed* fields by default, instead of only the fields passed in the `values` argument. I also included a test case, a line in the changelog, and a minor update of pg-native in packages.json, as the previous version wouldn't build on my machine.

This is my first contribution to the project. Please let me know if there's anything I need to improve :-)


janmeier:
Thanks a  thoroughly explained, documented and tested first contribution - Hope to see more from you ;)!

Issue #7506, opened 2017-04-11
Contradictions in documentation


Alexsey:
On <a href="http://docs.sequelizejs.com/en/latest/docs/models-definition/ ">models definition page</a> on definition of `underscored` property it says
```
  // don't use camelcase for automatically added attributes but underscore style
  // so updatedAt will be updated_at
```
at the same time on <a href="docs.sequelizejs.com/en/latest/api/sequelize/">API page</a> in section related to `[options.createdAt]`, `[options.updatedAt]` and `[options.deletedAt]` all have sentence `Not affected by underscored setting`

Upd.:
As far as I can see `underscored` affects `createdAt`, `updatedAt` and `deletedAt`. And I think this is how it should be

janmeier:
I see how this is not 100% clear. `Not affected by underscored setting` means that if you add `{ createdAt: 'fooBar` }` in the options object, it will not be converted to `foo_bar`, even if `underscored` is true.

You are free to provide a PR if you can come up with a better wording ;)

PR #7528, opened 2017-04-16
extend model options documentation for properties 'underscored' and 'underscoredAll'

Alexsey:
Better wording for [this](https://github.com/sequelize/sequelize/issues/7506) problem

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -670,8 +670,8 @@ class Model {
    * @param {Boolean}                 [options.omitNull] Don't persist null values. This means that all columns with null values will not be saved
    * @param {Boolean}                 [options.timestamps=true] Adds createdAt and updatedAt timestamps to the model.
    * @param {Boolean}                 [options.paranoid=false] Calling `destroy` will not delete the model, but instead set a `deletedAt` timestamp if this is true. Needs `timestamps=true` to work
-   * @param {Boolean}                 [options.underscored=false] Converts all camelCased columns to underscored if true
-   * @param {Boolean}                 [options.underscoredAll=false] Converts camelCased model names to underscored table names if true
+   * @param {Boolean}                 [options.underscored=false] Converts all camelCased columns to underscored if true. Will not affect timestamp fields named explicitly by model options and will not affect fields with explicitly set `field` option
+   * @param {Boolean}                 [options.underscoredAll=false] Converts camelCased model names to underscored table names if true. Will not change model name if freezeTableName is set to true
    * @param {Boolean}                 [options.freezeTableName=false] If freezeTableName is true, sequelize will not try to alter the model name to get the table name. Otherwise, the model name will be pluralized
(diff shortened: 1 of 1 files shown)


lib/model.js
```diff
@@ -670,8 +670,8 @@ class Model {
    * @param {Boolean}                 [options.omitNull] Don't persist null values. This means that all columns with null values will not be saved
    * @param {Boolean}                 [options.timestamps=true] Adds createdAt and updatedAt timestamps to the model.
    * @param {Boolean}                 [options.paranoid=false] Calling `destroy` will not delete the model, but instead set a `deletedAt` timestamp if this is true. Needs `timestamps=true` to work
-   * @param {Boolean}                 [options.underscored=false] Converts all camelCased columns to underscored if true
-   * @param {Boolean}                 [options.underscoredAll=false] Converts camelCased model names to underscored table names if true
+   * @param {Boolean}                 [options.underscored=false] Converts all camelCased columns to underscored if true. Will not convert column name if it set explicitly through options
+   * @param {Boolean}                 [options.underscoredAll=false] Converts camelCased model names to underscored table names if true. Will not convert names for models with option `freezeTableName` set to `true`
```
sushantdhiman: Will not change model name if `freezeTableName` is set to `true`

lib/model.js
```diff
@@ -670,8 +670,8 @@ class Model {
    * @param {Boolean}                 [options.omitNull] Don't persist null values. This means that all columns with null values will not be saved
    * @param {Boolean}                 [options.timestamps=true] Adds createdAt and updatedAt timestamps to the model.
    * @param {Boolean}                 [options.paranoid=false] Calling `destroy` will not delete the model, but instead set a `deletedAt` timestamp if this is true. Needs `timestamps=true` to work
-   * @param {Boolean}                 [options.underscored=false] Converts all camelCased columns to underscored if true
-   * @param {Boolean}                 [options.underscoredAll=false] Converts camelCased model names to underscored table names if true
+   * @param {Boolean}                 [options.underscored=false] Converts all camelCased columns to underscored if true. Will not convert column name if it set explicitly through options
```
sushantdhiman: I think you meant `field` option, proper wording to relay this message would be,

"unless column name is explicitly set with `field` options"

Alexsey: I meant options like `createdAt`. What do you mean under `field`? I have searched over [API](http://docs.sequelizejs.com/en/latest/api/sequelize/) page and didn't find `options.field` but only `options.fieldMap`

sushantdhiman: `field` is an attribute option which can override column name. Used something like this

```js
orgName: {
   type: Sequelize.STRING,
   field: 'org_name'
}
```

You are talking about timestamp fields, you can change message to something like

"Will not affect timestamp fields named explicitly by model options"

Alexsey:
If there is anything else you would like me to change in this PR to accept it - I'm open :)

Issue #7408, opened 2017-03-21
Helper method to determine whether instance has been soft deleted or not


lekhnath:
## What you are doing?
Trying to determine whether I am dealing with deleted record or not.

## What do you expect to happen?
Provide us with some helper method.

## What is actually happening?
When the model is configured to use `paranoid:true` the `deletedAt` attribute could be renamed to whatever name. But sequelize will definitely know about this column name even if custom name is provided.

So, I was expecting some helper method to identify whether the instance has be soft deleted or not. So that we don't have to hard code the `deletedAt` column name some other place in our code.

__Dialect:__ sqlite
__Database version__: v3.1.8
__Sequelize version__: v3.30.2

hd719:
Hello I am new to Open source can I work on this?

sushantdhiman:
@hd719 Thanks but Its already tackled in https://github.com/sequelize/sequelize/pull/7531, you may look at other `good-first-pr` issues https://github.com/sequelize/sequelize/labels/good-first-pr

jsatk:
@hd719 I tacked this here https://github.com/sequelize/sequelize/pull/7531, but the build is failing. It is failing at different steps each time though and seems very unrelated to my changes. If you have any feedback on my changes or have any idea why the build may be failing I'd appreciate any extra help. 😁

sushantdhiman:
Added in v4 https://github.com/sequelize/sequelize/pull/7531

PR #7531, opened 2017-04-16
Add `isSoftDeleted` helper method to model instances

jsatk:
### Pull Request check-list

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)? **I need to fix failing tests**

### Description of change

Since users can rename the `deletedAt` field when creating a model it is helpful to have a `isSoftDeleted` so other users of the DB don't have to hard code the custom `deletedAt` field name throughout their codebase.

Addresses: https://github.com/sequelize/sequelize/issues/7408

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3760,8 +3760,22 @@ class Model {
 
     const deletedAtAttribute = this.constructor.rawAttributes[this.constructor._timestampAttributes.deletedAt];
     const defaultValue = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
+    const deletedAt = this.get(this.constructor._timestampAttributes.deletedAt);
+    const isNotSet = deletedAt === defaultValue;
 
-    return this.get(this.constructor._timestampAttributes.deletedAt) !== defaultValue;
+    // No need to check the value of deletedAt if it's equal to the default
+    // value.  If so return the inverse of `isNotSet` since we are asking if
+    // the model *is* soft-deleted.
+    if (isNotSet) {
+      return !isNotSet;
+    }
+
+    const now = moment();
+    const deletedAtIsInTheFuture = moment(deletedAt).isSameOrAfter(now);
+
+    // If deletedAt is a datetime in the future or now then the model is *not* soft-deleted.
+    // Therefore, return the inverse of `deletedAtIsInTheFuture`.
+    return !deletedAtIsInTheFuture;
   }
 
   /**
--- a/lib/model.js
+++ b/lib/model.js
@@ -3747,6 +3747,24 @@ class Model {
   }
 
   /**
+   * Helper method to determine if a instance is "soft deleted".  This is
+   * particularly useful if the implementer renamed the `deletedAt` attribute
+   * to something different.  This method requires `paranoid` to be enabled.
+   *
+   * @returns {Boolean}
+   */
+  isSoftDeleted() {
+    if (!this.constructor._timestampAttributes.deletedAt) {
+      throw new Error('Model is not paranoid');
+    }
+
+    const deletedAtAttribute = this.constructor.rawAttributes[this.constructor._timestampAttributes.deletedAt];
+    const defaultValue = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
+
+    return this.get(this.constructor._timestampAttributes.deletedAt) !== defaultValue;
+  }
+
+  /**
    * Restore the row corresponding to this instance. Only available for paranoid models.
    *
    * @param {Object}      [options={}]
--- a/lib/model.js
+++ b/lib/model.js
@@ -3761,13 +3761,13 @@ class Model {
     const deletedAtAttribute = this.constructor.rawAttributes[this.constructor._timestampAttributes.deletedAt];
     const defaultValue = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
     const deletedAt = this.get(this.constructor._timestampAttributes.deletedAt);
-    const isNotSet = deletedAt === defaultValue;
+    const isSet = deletedAt !== defaultValue;
 
     // No need to check the value of deletedAt if it's equal to the default
     // value.  If so return the inverse of `isNotSet` since we are asking if
     // the model *is* soft-deleted.
-    if (isNotSet) {
-      return !isNotSet;
+    if (!isSet) {
+      return isSet;
     }
 
     const now = moment();
--- a/lib/model.js
+++ b/lib/model.js
@@ -3771,9 +3771,9 @@ class Model {
     }
 
     const now = moment();
-    const deletedAtIsInTheFuture = moment(deletedAt).isSameOrAfter(now);
+    const deletedAtIsInTheFuture = moment(deletedAt).isAfter(now);
 
-    // If deletedAt is a datetime in the future or now then the model is *not* soft-deleted.
+    // If deletedAt is a datetime in the future then the model is *not* soft-deleted.
     // Therefore, return the inverse of `deletedAtIsInTheFuture`.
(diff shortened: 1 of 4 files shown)


jsatk:
The test suite seems non-deterministic. It was passing. Adore and I pushed copy-only changes and now it's failing. I'd love to help fix if possible but a bit stumped st the moment.

test/integration/instance.test.js
```diff
@@ -2076,6 +2076,69 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
     });
   });
 
+  describe('isSoftDeleted', () => {
+    beforeEach(function() {
+      this.ParanoidUser = this.sequelize.define('ParanoidUser', {
+        username: { type: DataTypes.STRING }
+      }, { paranoid: true });
+
+      this.ParanoidUser.hasOne(this.ParanoidUser);
```
janmeier: Is this relation needed?

test/unit/instance/is-soft-deleted.test.js
```diff
@@ -0,0 +1,60 @@
...
+    it('should not throw if paranoid is set to true', function() {
+      expect(() => {
+        this.paranoidUser.isSoftDeleted();
+      }).to.not.throw();
+    });
+
+    it('should throw if paranoid is set to false', function() {
+      expect(() => {
+        this.user.isSoftDeleted();
+      }).to.throw();
```
janmeier: `to.throw('Model is not paranoid')`

test/integration/instance.test.js
jsatk: I'm not entirely sure but I believe so. https://github.com/sequelize/sequelize/pull/7531/files/354210b1992afccaae8fdb270fddcb8e4752923f#diff-2756a0ac12fc81f7843fd569393b5966R1655 has the same code and I found that I needed that logic here.

test/unit/instance/is-soft-deleted.test.js
jsatk: Done! Will amend fix to this commit.

jsatk:
@janmeier thank you! I responded to one and fixed the other. Thanks for the review. Excited to contribute. 😄

test/integration/instance.test.js
```diff
@@ -2076,6 +2076,69 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
...
+      return this.ParanoidUser.sync({ force: true });
+    });
+
+    it('returns false if user is not soft deleted', function() {
+      const self = this;
```
felixfbecker: Don't use `self`

lib/model.js
```diff
@@ -3747,6 +3747,24 @@ class Model {
   }
 
   /**
+   * Helper method to determine if a instance is "soft deleted".  This is
+   * particularly useful if the implementer renamed the `deletedAt` attribute
+   * to something different.  This method requires `paranoid` to be enabled.
+   *
+   * @returns {Boolean}
+   */
+  isSoftDeleted() {
```
felixfbecker: Yeah, shouldn't the function just check if `deletedAt` is set and less than current datetime? The default value logic looks weird to me

test/integration/instance.test.js
```diff
@@ -2076,6 +2076,69 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
     });
   });
 
+  describe('isSoftDeleted', () => {
```
jsatk: According to the linter in sequelize this is incorrect. It asks that you use arrow functions when possible over `function` and only use `function` when scope of `this` is a concern. 

`Unexpected function expression. (prefer-arrow-callback)`.

See https://github.com/sequelize/sequelize/blob/master/.eslintrc.json#L15

test/integration/instance.test.js
jsatk: Good call. It's not needed due to arrow function.

lib/model.js
jsatk: @felixfbecker The logic is meant to get the value of the property being used for soft delete (which by default is `deletedAt`. 

@sushantdhiman I didn't consider future deletes. I will update the logic to return true if the value of the soft delete property is same as or before now.

lib/model.js
```diff
@@ -3747,6 +3747,24 @@ class Model {
...
+    if (!this.constructor._timestampAttributes.deletedAt) {
+      throw new Error('Model is not paranoid');
+    }
+
+    const deletedAtAttribute = this.constructor.rawAttributes[this.constructor._timestampAttributes.deletedAt];
+    const notSoftDeleted = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
+
+    return this.getDataValue(this.constructor._timestampAttributes.deletedAt) !== notSoftDeleted;
```
jsatk: @sushantdhiman Will do! What is the difference between `this.get` and `this.getDataValue`? I see both being used in the codebase.

test/integration/instance.test.js
jsatk: I've removed it and all tests continue to pass. It seems it is not needed.

jsatk:
@felixfbecker @sushantdhiman I've addressed all comments and added in a few more unit tests. Please let me know if there's anything else. Thanks for the review. :smile:

janmeier:
@jsatk Fails for sqlite and mssql

``` 
  1) [SQLITE] Instance isSoftDeleted returns true if user is soft deleted:
     AssertionError: expected false to be true
      at null.<anonymous> (node_modules/chai/lib/chai/core/assertions.js:278:10)
      at addProperty (node_modules/chai/lib/chai/utils/addProperty.js:43:29)
      at null.<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:286:25)
      at Object.defineProperty.get (node_modules/chai/lib/chai/utils/overwriteProperty.js:50:37)
      at test/integration/instance.test.js:2100:51
  From previous event:
      at Promise.then (node_modules/cls-bluebird/lib/shimMethod.js:38:20)
  2) [SQLITE] Instance isSoftDeleted works with custom `deletedAt` field name:
     AssertionError: expected false to be true
      at null.<anonymous> (node_modules/chai/lib/chai/core/assertions.js:278:10)
      at addProperty (node_modules/chai/lib/chai/utils/addProperty.js:43:29)
      at null.<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:286:25)
      at Object.defineProperty.get (node_modules/chai/lib/chai/utils/overwriteProperty.js:50:37)
      at test/integration/instance.test.js:2127:53
  From previous event:
      at Promise.then (node_modules/cls-bluebird/lib/shimMethod.js:38:20)
```

jsatk:
Apologies.

I saw that this failed the other day and haven't had time to fix.  Will hopefully fix sometime this week.

jsatk:
@janmeier Any idea why the test suite date time is set to `1970-01-01T00:00:00.000Z`? Both my `deleted_at` and `now` values have a date time of `1970-01-01T00:00:00.000Z`. Any help? :)

lib/model.js
```diff
@@ -3747,6 +3747,38 @@ class Model {
...
+   */
+  isSoftDeleted() {
+    if (!this.constructor._timestampAttributes.deletedAt) {
+      throw new Error('Model is not paranoid');
+    }
+
+    const deletedAtAttribute = this.constructor.rawAttributes[this.constructor._timestampAttributes.deletedAt];
+    const defaultValue = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
+    const deletedAt = this.get(this.constructor._timestampAttributes.deletedAt);
+    const isNotSet = deletedAt === defaultValue;
```
felixfbecker: Avoid booleans with negatives, it's hard to read something like `!isNotSet`. Prefer `isSet`

jsatk: Done!

jsatk:
@felixfbecker @janmeier @sushantdhiman I believe this PR is good to go. Any other concerns?

PR #7557, opened 2017-04-24
feat(validation): new properties for ValidationErrorItem class

iamjochem:
### Pull Request check-list

- [x] Does your issue contain a description of the issue you are solving?

### Description of change

This PR is looking to make the `ValidationErrorItem` class more useful by adding extra/new structured data to the instances. 

The reasoning for these changes are:

1. in order to expose more useful validation-error information at the point that we want to output/log/display the errors so that we can facilitate showing/creating messages in different languages and for different types of users - in multi-lingual systems the internal error message property is unlikely to be in the current users' language and is likely to be technical in nature - we want to be able to generate localized error messages "after the fact" and in order to create useful message we need the extra information about which validator (if applicable), and it's arguments (e.g. so you can generate a message such as "lengte moet minimaal 3 zijn" (this is dutch for "length must be at least 3" :-)).

2. we desire the ability for a REST API service to respond to a [mutation] request, in the case that there are validation errors, with both the instance data and the information about the error - this is inline with [JSONAPI](http://jsonapi.org/) spec, where an `error` property has a `pointer` (akin to `ValidationErrorItem.path`) that points to some data in the response's `data` property. Attaching the DOA `instance` to the `ValidationErrorItem` instance means that we do not need to manually track which DOA instance needs to be potentially returned with which error in our service/app. 
#### new `ValidationErrorItem` instance properties

1. `validatorKey` - a string identifier for the validation error
2. `validatorName` - name of a  built-in validator function, if applicable (e.g. "in" or "len")
3. `validatorArgs` - an array of arguments that the validator was called with (only relevant to some built-in validators)
4. `instance` - a refence to the DAO instance that triggered the error, if available.
5. `origin` - contains the value of the "type" parameter (as passed in by constructor callers) mapped to a valid value in `ValidationErrorItem.ORIGINS`. 

#### new `ValidationErrorItem.getValidatorKey` instance method

this method returns a lowercased string, by default namespaced with the instance's `type` (assuming it is a valid built-in ValidationErrorItem.ORIGINS value) e.g. `"function.len"`

#### an evil `type` of property ...

the current `type` property of `ValidationErrorItem` is set to arbitrary strings making it a bad candidate for anything other that string-concatenation (with regard to generating error message), This PR suggest to deprecate `type` in favour of the new `origin` property, the fixed/valid valuesfor which are:

- **CORE** - for general [model] validation errors that sequelize [can] perform regardless (e.g. not null checks and general data-type violations)
- **DB** - for any error being triggered by the DB (e.g. constraint violations)        
- **FUNCTION** - for validation errors from explicitly defined model/attrib validation rules

At a future date when the sun sets on the `ValidationErrorItem.type` property we can remove it completely, change the `ValidationErrorItem` constructor to use it's current `type` parameter as the value for `ValidationErrorItem.origin` and update the various places in the code where `ValidationErrorItem` is instantiated to pass in a valid `origin` in place of the current `type` string (there are 4-5 places this will need to happen - this PR include comments in all these places that stipulate what the `origin` parameter will be - i.e. what the `type` parameter is currently being mapped internally by `ValidationErrorItem`)

## Caveat

1. some totally unrelated tests fail on my local machine, I am ignoring these because they have absolutely nothing to do with the changes in this PR (I checked) and because I branched from `master` which was in a v4 beta state (so not unlikely that a few tests are not currently passing), additionally I have only run tests against MySQLv5.7

2. the new `instance` and `validator` properties of the `ValidationErrorItem` class are all optional as far as the class itself is concerned, in the case of bulk processing it is possible that (at the least) a DOA instance will not be available to attach to an error - this is something I feel the user of the lib needs to be aware of and work around (if applicable). 

3. I made a couple of slight changes to existing tests related to validation errors - the tests were checking [generated] error message values in order to determine whether an error instance is valid (IMHO these tests are brittle anyway)

4. I felt that updating documentation for this was a bit much given that ValidationError and ValidationErrorItem classes are not documented at all save for the following snippet ([found here](https://github.com/sequelize/sequelize/blob/master/docs/instances.md)):
```
Tasks.bulkCreate([
  {name: 'foo', code: '123'},
  {code: '1234'},
  {name: 'bar', code: '1'}
], { validate: true }).catch(function(errors) {
  /* console.log(errors) would look like:
  [
    { record:
    ...
    errors:
      { name: 'SequelizeValidationError',
        message: 'Validation error',
        errors: [Object] } },
    { record:
      ...
      errors:
        { name: 'SequelizeValidationError',
        message: 'Validation error',
        errors: [Object] } }
  ]
  */
})
```

### DIFF
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -54,7 +54,7 @@ class ValidationError extends BaseError {
 
       // ... otherwise create a concatenated message out of existing errors.
     } else if (this.errors.length > 0 && this.errors[0].message) {
-      this.message = this.errors.map(err => err.type + ': ' + err.message).join(',\n');
+      this.message = this.errors.map(err => (err.original_type || err.type) + ': ' + err.message).join(',\n');
     }
     Error.captureStackTrace(this, this.constructor);
   }
@@ -224,21 +224,56 @@ exports.UnknownConstraintError = UnknownConstraintError;
  * Validation Error Item
  * Instances of this class are included in the `ValidationError.errors` property.
  *
- * @param {string} message An error message
- * @param {string} type The type of the validation error
- * @param {string} path The field that triggered the validation error
+ * @param {String} message  - An error message
+ * @param {String} type     - The type of the validation error
+ * @param {String} path     - The field that triggered the validation error
+ * @param {String} value    - The value that generated the error
+ * @param {Object} inst     - optional, the DAO instance that caused the validation error
+ * @param {String} fn_name  - optional, property name of the validator that caused the validation error (e.g. "in" or "len")
+ * @param {String} fn_args  - optional, parameters used with the validator (this is only relevant to built in validators)
+ *
  * @param {string} value The value that generated the error
  */
 class ValidationErrorItem {
-  constructor(message, type, path, value) {
+  constructor(message, type, path, value, inst, fn_name, fn_args) {
     this.message = message || '';
-    this.type = type || null;
+    this.type = null;
     this.path = path || null;
     this.value = value !== undefined ? value : null;
-    //This doesn't need captureStackTrace because it's not a subclass of Error
+
+    this.instance = inst || null;
+    this.validator_name = fn_name || null;
+    this.validator_args = fn_args || [];
+
+    if (type) {
+      if (ValidationErrorItem.TYPES[ type ]) {
+        this.type = type;
+      } else {
+        const lctype = (type + '').toLowerCase().trim();
+        const rtype  = ValidationErrorItem.TypeStringMap[ lctype ]; // "real type" ... but also https://en.wikipedia.org/wiki/R-Type :-D
+
+        if (rtype && ValidationErrorItem.TYPES[ rtype ]) {
+          this.type           = ValidationErrorItem.TypeStringMap[ rtype ];
+          this.original_type  = type; // expose the value of "type" as it was in previous version (just in case someone is reliant on it?)
+        }
+      }
+    }
   }
 }
-exports.ValidationErrorItem = ValidationErrorItem;
+exports.ValidationErrorItem       = ValidationErrorItem;
+
+ValidationErrorItem.TYPES         = Object.freeze({
+  CORE      : 'CORE',
+  DB        : 'DB',
+  FUNCTION  : 'FUNCTION'
+});
+
+ValidationErrorItem.TypeStringMap = {
+  'notnull violation'             : 'CORE',
+  'string violation'              : 'DB',
+  'validation error'              : '',
+  'unique violation'              : '',
+};
 
 /**
  * A base class for all connection related errors.
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -275,13 +275,18 @@ class ValidationErrorItem {
    * @throws  {Error}                 - thrown if NSSeparator is found to be invalid.
    * @return  {String}
    */
-  getValidatorKey(useTypeAsNS = true, NSSeparator = '.') {
+  getValidatorKey(useTypeAsNS/* = true */, NSSeparator/* = '.'*/) {
+    // parameter defaults: some CI instances give a syntax error on actual parameter default values
+    const
+      useTANS = typeof useTypeAsNS === 'undefined' ?  true : !!useTypeAsNS,
+      NSSep   = typeof NSSeparator === 'undefined' ? '.' : NSSeparator;
+
     const
       type  = this.type,
       key   = this.validatorKey || this.validatorName,
-      useNS = useTypeAsNS && type && ValidationErrorItem.TYPES[ type ];
+      useNS = useTANS && type && ValidationErrorItem.TYPES[ type ];
 
-    if (useNS && (typeof NSSeparator !== 'string' || !NSSeparator.length)) {
+    if (useNS && (typeof NSSep !== 'string' || !NSSep.length)) {
       throw new Error('Invalid namespace separator given, must be a non-empty string');
     }
 
@@ -289,7 +294,7 @@ class ValidationErrorItem {
       return '';
     }
 
-    return (useNS ? [type, key].join(NSSeparator) : key).trim().toLowerCase();
+    return (useNS ? [type, key].join(NSSep) : key).trim().toLowerCase();
   }
 }
 exports.ValidationErrorItem       = ValidationErrorItem;
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -296,6 +296,28 @@ class ValidationErrorItem {
 
     return (useNS ? [type, key].join(NSSep) : key).trim().toLowerCase();
   }
+
+
+  /**
+   * generate custom output for JSON.stringify(),
+   * this avoids errors related to serializing circular-refs (due to the `this.instance` property)
+   *
+   * @return  {String}
+   */
+  toJSON() {
+    const o = {
+      message       : this.message,
+      type          : this.type,
+      path          : this.path,
+      value         : this.value,
+      validatorKey  : this.validatorKey,
+      validatorName : this.validatorName,
+      validatorArgs : this.validatorArgs,
+      instance      : this.instance ? this.instance.get({ plain: true }) : null
+    };
+
+    return o;
+  }
 }
 exports.ValidationErrorItem       = ValidationErrorItem;
 
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -253,7 +253,7 @@ class ValidationErrorItem {
         const rtype  = ValidationErrorItem.TypeStringMap[ lctype ]; // "real type" ... but also https://en.wikipedia.org/wiki/R-Type :-D
 
         if (rtype && ValidationErrorItem.TYPES[ rtype ]) {
-          this.type           = ValidationErrorItem.TypeStringMap[ rtype ];
+          this.type           = rtype;
           this.original_type  = type; // expose the value of "type" as it was in previous version (just in case someone is reliant on it?)
         }
       }
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const _ = require('lodash');
+
 /**
  * Sequelize provides a host of custom error classes, to allow you to do easier debugging. All of these errors are exposed on the sequelize object and the sequelize constructor.
  * All sequelize errors inherit from the base JS error object.
@@ -224,14 +226,14 @@ exports.UnknownConstraintError = UnknownConstraintError;
  * Validation Error Item
  * Instances of this class are included in the `ValidationError.errors` property.
  *
- * @param {String} message      - An error message
- * @param {String} type         - The type/origin of the validation error
- * @param {String} path         - The field that triggered the validation error
- * @param {String} value        - The value that generated the error
- * @param {Object} inst         - optional, the DAO instance that caused the validation error
- * @param {Object} validatorKey - optional, a validation "key", used for identification
- * @param {String} fnName       - optional, property name of the BUILT-IN validator function that caused the validation error (e.g. "in" or "len"), if applicable
- * @param {String} fnArgs       - optional, parameters used with the BUILT-IN validator function, if applicable
+ * @param {String} message An error message
+ * @param {String} type The type/origin of the validation error
+ * @param {String} path The field that triggered the validation error
+ * @param {String} value The value that generated the error
+ * @param {Object} inst optional, the DAO instance that caused the validation error
+ * @param {Object} validatorKey optional, a validation "key", used for identification
+ * @param {String} fnName optional, property name of the BUILT-IN validator function that caused the validation error (e.g. "in" or "len"), if applicable
+ * @param {String} fnArgs optional, parameters used with the BUILT-IN validator function, if applicable
  *
  * @param {string} value The value that generated the error
  */
@@ -242,17 +244,17 @@ class ValidationErrorItem {
     this.path = path || null;
     this.value = value !== undefined ? value : null;
 
-    this.origin         = null;
-    this.instance       = inst || null;
-    this.validatorKey   = validatorKey || null;
-    this.validatorName  = fnName || null;
-    this.validatorArgs  = fnArgs || [];
+    this.origin = null;
+    this.instance = inst || null;
+    this.validatorKey = validatorKey || null;
+    this.validatorName = fnName || null;
+    this.validatorArgs = fnArgs || [];
 
     if (type) {
       if (ValidationErrorItem.Origins[ type ]) {
         this.origin = type;
       } else {
-        const lowercaseType = (type + '').toLowerCase().trim();
+        const lowercaseType = _.toLower(type + '').trim();
         const realType  = ValidationErrorItem.TypeStringMap[ lowercaseType ];
 
         if (realType && ValidationErrorItem.Origins[ realType ]) {
@@ -261,26 +263,28 @@ class ValidationErrorItem {
         }
       }
     }
+
+    // This doesn't need captureStackTrace because it's not a subclass of Error
   }
 
   /**
    * return a lowercase, trimmed string "key" that identifies the validator.
    *
-   * N.B. the string will be empty if the instance has neither a valid `validatorKey` property nor a valid `validatorName` property
+   * Note: the string will be empty if the instance has neither a valid `validatorKey` property nor a valid `validatorName` property
    *
-   * @param   {Boolean} [useTypeAsNS=true]    - optional, defaults to TRUE, controls whether the returned value is "namespace",
+   * @param   {Boolean} [useTypeAsNS=true]      optional, defaults to TRUE, controls whether the returned value is "namespace",
    *                                            this parameter is ignored if the validator's `type` is not one of ValidationErrorItem.Origins
-   * @param   {String}  [NSSeparator='.']     - a separator string for concatenating the namespace, must be not be empty,
+   * @param   {String}  [NSSeparator='.']       a separator string for concatenating the namespace, must be not be empty,
    *                                            defaults to "." (fullstop). only used and validated if useTypeAsNS is TRUE.
-   * @throws  {Error}                         - thrown if NSSeparator is found to be invalid.
+   * @throws  {Error}                           thrown if NSSeparator is found to be invalid.
    * @return  {String}
    */
   getValidatorKey(useTypeAsNS, NSSeparator) {
     const useTANS = typeof useTypeAsNS === 'undefined' ?  true : !!useTypeAsNS;
-    const NSSep   = typeof NSSeparator === 'undefined' ? '.' : NSSeparator;
+    const NSSep = typeof NSSeparator === 'undefined' ? '.' : NSSeparator;
 
-    const type  = this.origin;
-    const key   = this.validatorKey || this.validatorName;
+    const type = this.origin;
+    const key = this.validatorKey || this.validatorName;
     const useNS = useTANS && type && ValidationErrorItem.Origins[ type ];
 
     if (useNS && (typeof NSSep !== 'string' || !NSSep.length)) {
@@ -291,10 +295,11 @@ class ValidationErrorItem {
       return '';
     }
 
-    return (useNS ? [type, key].join(NSSep) : key).trim().toLowerCase();
+    return _.toLower(useNS ? [type, key].join(NSSep) : key).trim();
   }
 }
-exports.ValidationErrorItem       = ValidationErrorItem;
+
+exports.ValidationErrorItem = ValidationErrorItem;
 
 /**
  * An enum that defines valid ValidationErrorItem `origin` values
@@ -304,12 +309,11 @@ exports.ValidationErrorItem       = ValidationErrorItem;
  * @property DB         {String}  specifies validation errors that originate from the storage engine
  * @property FUNCTION   {String}  specifies validation errors that originate from validator functions (both built-in and custom) defined for a given attribute
  */
-ValidationErrorItem.Origins       = {
-  CORE                            : 'CORE',
-  DB                              : 'DB',
-  FUNCTION                        : 'FUNCTION'
+ValidationErrorItem.Origins = {
+  CORE : 'CORE',
+  DB : 'DB',
+  FUNCTION : 'FUNCTION'
 };
-
 
 /**
  * An object that is used internally by the `ValidationErrorItem` class
@@ -319,10 +323,10 @@ ValidationErrorItem.Origins       = {
  * @type {Object}
  */
 ValidationErrorItem.TypeStringMap = {
-  'notnull violation'             : 'CORE',
-  'string violation'              : 'CORE',
-  'unique violation'              : 'DB',
-  'validation error'              : 'FUNCTION'
+  'notnull violation' : 'CORE',
+  'string violation' : 'CORE',
+  'unique violation' : 'DB',
+  'validation error' : 'FUNCTION'
 };
 
 /**
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -54,7 +54,7 @@ class ValidationError extends BaseError {
 
       // ... otherwise create a concatenated message out of existing errors.
     } else if (this.errors.length > 0 && this.errors[0].message) {
-      this.message = this.errors.map(err => (err.original_type || err.type) + ': ' + err.message).join(',\n');
+      this.message = this.errors.map(err => (err.originalType || err.type) + ': ' + err.message).join(',\n');
     }
     Error.captureStackTrace(this, this.constructor);
   }
@@ -229,24 +229,24 @@ exports.UnknownConstraintError = UnknownConstraintError;
  * @param {String} path     - The field that triggered the validation error
  * @param {String} value    - The value that generated the error
  * @param {Object} inst     - optional, the DAO instance that caused the validation error
- * @param {String} fn_name  - optional, property name of the validator that caused the validation error (e.g. "in" or "len")
- * @param {String} fn_args  - optional, parameters used with the validator (this is only relevant to built in validators)
+ * @param {String} fnName  - optional, property name of the validator that caused the validation error (e.g. "in" or "len")
+ * @param {String} fnArgs  - optional, parameters used with the validator (this is only relevant to built in validators)
  *
  * @param {string} value The value that generated the error
  */
 class ValidationErrorItem {
-  constructor(message, type, path, value, inst, fn_name, fn_args) {
+  constructor(message, type, path, value, inst, fnName, fnArgs) {
     this.message = message || '';
     this.type = null;
     this.path = path || null;
     this.value = value !== undefined ? value : null;
 
     this.instance = inst || null;
-    this.validator_name = fn_name || null;
-    this.validator_args = fn_args || [];
+    this.validatorName = fnName || null;
+    this.validatorArgs = fnArgs || [];
 
     if (type) {
-      if (ValidationErrorItem.TYPES[ type ]) {
+      if (ValidationErrorItem.USE_OLD_TYPES || ValidationErrorItem.TYPES[ type ]) {
         this.type = type;
       } else {
         const lctype = (type + '').toLowerCase().trim();
@@ -254,7 +254,7 @@ class ValidationErrorItem {
 
         if (rtype && ValidationErrorItem.TYPES[ rtype ]) {
           this.type           = rtype;
-          this.original_type  = type; // expose the value of "type" as it was in previous version (just in case someone is reliant on it?)
+          this.originalType   = type; // expose the value of "type" as it was in previous version (just in case someone is reliant on it?)
         }
       }
     }
@@ -263,16 +263,16 @@ class ValidationErrorItem {
 exports.ValidationErrorItem       = ValidationErrorItem;
 
 ValidationErrorItem.TYPES         = Object.freeze({
-  CORE      : 'CORE',
-  DB        : 'DB',
-  FUNCTION  : 'FUNCTION'
+  CORE                            : 'CORE',
+  DB                              : 'DB',
+  FUNCTION                        : 'FUNCTION'
 });
 
 ValidationErrorItem.TypeStringMap = {
   'notnull violation'             : 'CORE',
-  'string violation'              : 'DB',
-  'validation error'              : '',
-  'unique violation'              : '',
+  'string violation'              : 'CORE',
+  'unique violation'              : 'DB',
+  'validation error'              : 'FUNCTION'
 };
 
 /**
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -249,13 +249,13 @@ class ValidationErrorItem {
     this.validatorArgs  = fnArgs || [];
 
     if (type) {
-      if (ValidationErrorItem.ORIGINS[ type ]) {
+      if (ValidationErrorItem.Origins[ type ]) {
         this.origin = type;
       } else {
         const lowercaseType = (type + '').toLowerCase().trim();
         const realType  = ValidationErrorItem.TypeStringMap[ lowercaseType ];
 
-        if (realType && ValidationErrorItem.ORIGINS[ realType ]) {
+        if (realType && ValidationErrorItem.Origins[ realType ]) {
           this.origin = realType;
           this.type = type;
         }
@@ -269,7 +269,7 @@ class ValidationErrorItem {
    * N.B. the string will be empty if the instance has neither a valid `validatorKey` property nor a valid `validatorName` property
    *
    * @param   {Boolean} [useTypeAsNS=true]    - optional, defaults to TRUE, controls whether the returned value is "namespace",
-   *                                            this parameter is ignored if the validator's `type` is not one of ValidationErrorItem.ORIGINS
+   *                                            this parameter is ignored if the validator's `type` is not one of ValidationErrorItem.Origins
    * @param   {String}  [NSSeparato='.']      - a separator string for concatenating the namespace, must be not be empty,
    *                                            defaults to "." (fullstop). only used and validated if useTypeAsNS is TRUE.
    * @throws  {Error}                         - thrown if NSSeparator is found to be invalid.
@@ -281,7 +281,7 @@ class ValidationErrorItem {
 
     const type  = this.origin;
     const key   = this.validatorKey || this.validatorName;
-    const useNS = useTANS && type && ValidationErrorItem.ORIGINS[ type ];
+    const useNS = useTANS && type && ValidationErrorItem.Origins[ type ];
 
     if (useNS && (typeof NSSep !== 'string' || !NSSep.length)) {
       throw new Error('Invalid namespace separator given, must be a non-empty string');
@@ -304,11 +304,11 @@ exports.ValidationErrorItem       = ValidationErrorItem;
  * @property DB         {String}  specifies validation errors that originate from the storage engine
  * @property FUNCTION   {String}  specifies validation errors that originate from validator functions (both built-in and custom) defined for a given attribute
  */
-ValidationErrorItem.ORIGINS         = Object.freeze({
+ValidationErrorItem.Origins       = {
   CORE                            : 'CORE',
   DB                              : 'DB',
   FUNCTION                        : 'FUNCTION'
-});
+};
 
 
 /**
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -229,21 +229,24 @@ exports.UnknownConstraintError = UnknownConstraintError;
  * @param {String} path     - The field that triggered the validation error
  * @param {String} value    - The value that generated the error
  * @param {Object} inst     - optional, the DAO instance that caused the validation error
- * @param {String} fnName  - optional, property name of the validator that caused the validation error (e.g. "in" or "len")
- * @param {String} fnArgs  - optional, parameters used with the validator (this is only relevant to built in validators)
+ * @param {Object} vKey     - optional, a validation "key", used for identification
+ * @param {String} fnName   - optional, property name of the BUILT-IN validator function that caused the validation error (e.g. "in" or "len"), if applicable
+ * @param {String} fnArgs   - optional, parameters used with the BUILT-IN validator function, if applicable
  *
  * @param {string} value The value that generated the error
  */
 class ValidationErrorItem {
-  constructor(message, type, path, value, inst, fnName, fnArgs) {
+  constructor(message, type, path, value, inst, vKey, fnName, fnArgs) {
     this.message = message || '';
     this.type = null;
     this.path = path || null;
     this.value = value !== undefined ? value : null;
 
-    this.instance = inst || null;
-    this.validatorName = fnName || null;
-    this.validatorArgs = fnArgs || [];
+    this.instance       = inst || null;
+
+    this.validatorKey   = vKey || null;
+    this.validatorName  = fnName || null;
+    this.validatorArgs  = fnArgs || [];
 
     if (type) {
       if (ValidationErrorItem.USE_OLD_TYPES || ValidationErrorItem.TYPES[ type ]) {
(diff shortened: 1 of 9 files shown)


iamjochem:
@janmeier & co - I have a working version now with working tests (I'm only missing the documentation - I figured I'd wait with that because you may have feed back that requires changes to the functionality). I also beefed up the PR description a bit. 

I'd love to know if this is good enough to go in or if needs more work.

iamjochem:
btw. the Travis build seems to only have failed on something unrelated (and only for Postgres).

iamjochem:
I just rebased my PR branch on top of sequelize/master to fix the merge conflict - seems that travis decided to flake out on one of the jobs, at least that is what it looks like to me, if there is an actual problem that I need to address please let me know!

lib/errors/index.js
```diff
@@ -217,20 +217,96 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
  *
- * @param {string} message An error message
- * @param {string} type The type of the validation error
- * @param {string} path The field that triggered the validation error
+ * @param {String} message  - An error message
+ * @param {String} type     - The type of the validation error
+ * @param {String} path     - The field that triggered the validation error
+ * @param {String} value    - The value that generated the error
+ * @param {Object} inst     - optional, the DAO instance that caused the validation error
+ * @param {Object} vKey     - optional, a validation "key", used for identification
```
felixfbecker: we don't usually use abbreviations like this, it's clearer using the full name

lib/errors/index.js
```diff
@@ -217,20 +217,96 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
 class ValidationErrorItem {
-  constructor(message, type, path, value) {
+  constructor(message, type, path, value, inst, vKey, fnName, fnArgs) {
     this.message = message || '';
-    this.type = type || null;
+    this.type = null;
     this.path = path || null;
     this.value = value !== undefined ? value : null;
+
+    this.instance       = inst || null;
```
felixfbecker: not sure if this is a good idea. This will make the error very heavy weight. It might prevent JSON stringification of ValidationErrorItems, as instances have a reference to the Model, which has a reference to Sequelize, which has a reference to all Models, so it's a cyclic reference

lib/errors/index.js
```diff
@@ -217,20 +217,96 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+    this.validatorKey   = vKey || null;
+    this.validatorName  = fnName || null;
+    this.validatorArgs  = fnArgs || [];
+
+    if (type) {
+      if (ValidationErrorItem.USE_OLD_TYPES || ValidationErrorItem.TYPES[ type ]) {
+        this.type = type;
+      } else {
+        const lctype = (type + '').toLowerCase().trim();
+        const rtype  = ValidationErrorItem.TypeStringMap[ lctype ]; // "real type" ... but also https://en.wikipedia.org/wiki/R-Type :-D
```
felixfbecker: instead of comments explaining the abbreviations, just name them `realType` and `lowerCaseType`

lib/errors/index.js
```diff
@@ -217,20 +217,96 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+
+        if (rtype && ValidationErrorItem.TYPES[ rtype ]) {
+          this.type           = rtype;
+          this.originalType   = type; // expose the value of "type" as it was in previous version (just in case someone is reliant on it?)
```
felixfbecker: why break BC? Sounds like `origin` would be more fitting than `type` anyway

lib/instance-validator.js
```diff
@@ -275,10 +275,17 @@ class InstanceValidator {
...
       const validatorArgs = this._extractValidatorArgs(test, validatorType, field);
+
       if (!validator[validatorType].apply(validator, [valueString].concat(validatorArgs))) {
-      // extract the error msg
-        throw new Error(test.msg || `Validation ${validatorType} on ${field} failed`);
+        // extract the error msg
+        const e = new Error(test.msg || `Validation ${validatorType} on ${field} failed`);
+
+        e.validatorName = validatorType;
+        e.validatorArgs = validatorArgs;
```
felixfbecker: I would use `Object.assign()` for this.

lib/instance-validator.js
```diff
@@ -351,24 +369,40 @@ class InstanceValidator {
   _handleReflectedResult(field, value, promiseInspections) {
     for (const promiseInspection of promiseInspections) {
       if (promiseInspection.isRejected()) {
-        const rejection = promiseInspection.error();
-        this._pushError(true, field, rejection, value);
+        const
+          rejection = promiseInspection.error(),
+          isBuiltIn = !!rejection.validatorName;
```
felixfbecker: please use one `const` per line

felixfbecker:
There are also merge conflicts, changelog.md doesn't exist anymore, we now use semantic-release. It would be great if you could squash your commits and follow the new commit message convention. Besides that, I'm pro adding more useful properties to ValidationErroItem

lib/instance-validator.js
iamjochem: why?

felixfbecker: much shorter:
```js
throw Object.assign(new Error(test.msg || `Validation ${validatorType} on ${field} failed`), { validatorName, validatorArgs })
```

iamjochem: ah I see, I though you were referencing just lines 285 & 286 - and that seemed a little overkill.
I actually had pretty much `Object.assign` line before but I think I decided that it was a bit too terse :-) ... I'll put this line in.

lib/errors/index.js
iamjochem: we if it's not a good idea then let's screw this PR  ... because without this particular property it is of no value to me (please read reason no. 2 in the description above).

it is easy enough to implement a custom `toJSON()` method to avoid stringification problems related to circular refs.

with regard to this making the error "heavy weight" - I don't really see this, it's adding a reference to an existing object. how do you see this as heavy-weight?

lib/errors/index.js
iamjochem: the reasoning for breaking BC is in the description of the PR - TL;DR = current property contains arbitrary strings which we should not encourage people to rely on.

if you don't agree with the reason please let me know what it is you don't agree with exactly

lib/errors/index.js
felixfbecker: Okay, maybe I just need to see some more examples of how this can be used

iamjochem: more examples? I really only have the one use case, namely generating a `pointer` attribute value for JSONAPI error data: http://jsonapi.org/format/#error-objects. I suppose that responding with error & [instance] data without having to manually track the instance is also a bonus.

lib/instance-validator.js
iamjochem: there is no rule about this in your `.eslintrc`, and no other coding style documentation exists (any more?) ... I'm a bit loath to start making CS changes based on one person's POV ... what happens when janmeier asks for the exact opposite? :-) ... I've changed it but I think it this is a CS rule it really should be in the `.eslintrc` (and if something lints correctly then I feel it should be let through, with regard to CS at least) 

I generally write my consts like this because of a vague understanding that V8 prefers it this way.

lib/errors/index.js
felixfbecker: If it is possible to implement without BC breaks, please try that. Otherwise this would have to be pushed back to v5.

lib/instance-validator.js
felixfbecker: We can't have it in eslintrc because a big amount of the codebase doesn't conform, it can't be autofixed and is a lot of work to fix manually. But new code uses this style. The old style is an artifact of this:
```js
var foo = 1
  , bar = 2
  , baz = 3;
```
now with ES6 `let`/`const` there is no aestetic to it cause we have 2 different keywords now and they don't have the same length.
Having one `const` per line minimises diffs if a variable is to be removed and keeps the type of the declaration to the declaration itself.

iamjochem:
hi Felix, 

I've never done a `git squash`, none of the sequelize maintainers has ever asked me to do this before (and it does not seem to be mentioned anywhere in the "contributing" docs) - I imagine this is a personal preference of yours, but I'm not familiar with it - if you really really want this please give me a helping hand! 

with regard to "validation errors" documentation ... there currently is none; I think it is unfair to demand that I create that part of the documentation from scratch; I have changed/expanded existing fuctionality not added something new (impling that some documentation should exist already so I can expand on it)

felixfbecker:
No worries, we can do that then when we merge. But you still need to resolve merge conflicts.

There is documentation for these: 

http://docs.sequelizejs.com/class/lib/errors/index.js~ValidationError.html
http://docs.sequelizejs.com/class/lib/errors/index.js~ValidationError.html

While it is true that not all properties are documented, whatever new properties are added should be.

lib/errors/index.js
iamjochem: doing it without a BC break means adding a property, with the old one hanging around - that would just get confusing. `ValidationErrorItem.type` is currently an abitrary string that is set by whatever instantiated the instance - I consider that brittle, unwanted. 

I'd very much rather wait for v5 than perpetuate the current functionality.

felixfbecker: You do already perpetuate the current functionality inside the `originalType` property. Even for v5, it is an avoidable breaking change with no clear benefit. If it's not useful at all, we can deprecate `type`, but I just don't see the reason to break it when the value is still kept under a different property name.

lib/errors/index.js
```diff
@@ -226,21 +226,118 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+  toJSON() {
+    const o = {
+      message       : this.message,
+      type          : this.type,
+      path          : this.path,
+      value         : this.value,
+      validatorKey  : this.validatorKey,
+      validatorName : this.validatorName,
+      validatorArgs : this.validatorArgs,
+      instance      : this.instance ? this.instance.get({ plain: true }) : null
```
felixfbecker: actually, I just remembered that `Model` has `toJSON` defined so _I think_ JSON.stringify() would already invoke that (recursively)

iamjochem:
hi Felix,

I have pushed a bunch of tweaks based on your feedback, I also added a `toJSON()` function for the `ValidationErrorItem` class to avoid serialization problems related to the referenced `instance` - please let me know what you think of that.

I thought the API documentation (http://docs.sequelizejs.com/class/lib/errors/index.js%7EValidationError.html) was being generated automatically? if that is not the case please could you point me to the place where I can update docs for the new class properties.

squashing commits & writing conforming commit message to be done once the PR is in an acceptable state - I'm guessing that if this will not land before **v5** then it will be while before that happens (probably I'll have to fix a few more conflicts by then), no?

lib/instance-validator.js
iamjochem: aha I see ... also the argument for minimizing diffs is a really good one! (and I'm glad I'm not the only OCD sufferer that is annoyed that "let" and "const" have different string lengths!)

felixfbecker:
The documentation is generated from JSDoc tags. Example: https://sourcegraph.com/github.com/sequelize/sequelize/-/blob/lib/errors/index.js?q=@property#L39-40

I still want to avoid the breaking change.

lib/errors/index.js
iamjochem: I was thinking in terms of an overlap/migration period so that "that one guy" who is depending on the old `type` values can opt-in to them quite simply (and not be forced to choose between "stick on old version" and "fix all my codes").

I don't want to deprecate `type` because `type` is exactly the right name for the property (given the new values/content) ... the old values of the `type` property are not really 'types'.

deprecating `type` means having to choose another word that is synonym for "type" and suffer the consequence of people continuing to asking about and searching about error validation item `type` properties (I'm assuming people may encounter the property in older codebases)

you offered `origin` as a alternative name ... I didn't grok that initially but now I do! 

**QUESTION**: shall I change the PR to include a new property named `origin`, remove the `originalType` property and related functionality & leave the existing `type` property alone (you can choose to mark it as deprecated as and when)

felixfbecker: yeah, that was my idea.

lib/errors/index.js
iamjochem: I knew about the `Model.toJSON()` but I was not sure as to it's behaviour regarding "includes" ... I wanted to explicitly use the `{ plain: true }` option. 

you are right about `JSON.stringify()` handling the recursion correctly.

what would you like to happen here? should I remove this `toJSON()` method?

lib/errors/index.js
iamjochem: I'll do that then!

lib/errors/index.js
iamjochem: I just noticed that "codecov" is complaining about lack of [test] coverage for this new `toJSON()` method ... I guess I need to add a test (unless you'd rather I remove the method)

felixfbecker: Yeah, just remove it. Sorry about that

iamjochem:
I've removed the breaking changes related to the `type` property and implemented a new 'origin' property - the idea is still that `origin` will supercede `type` at some point, which is one reason why I chose to leave the constructor function signature as it is. the `toJSON()` method has also been removed. 

how are we looking now?

lib/errors/index.js
```diff
@@ -226,21 +226,96 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+   *
+   * @param   {Boolean} useTypeAsNS   - optional, defaults to TRUE, controls whether the returned value is "namespace",
+   *                                    this parameter is ignored if the validator's `type` is not one of ValidationErrorItem.ORIGINS
+   * @param   {String}  NSSeparator   - a separator string for concatenating the namespace, must be not be empty,
+   *                                    defaults to "." (fullstop). only used and validated if useTypeAsNS is TRUE.
+   * @throws  {Error}                 - thrown if NSSeparator is found to be invalid.
+   * @return  {String}
+   */
+  getValidatorKey(useTypeAsNS/* = true */, NSSeparator/* = '.'*/) {
+    // parameter defaults: some CI instances give a syntax error on actual parameter default values
```
felixfbecker: it's simply that Node 4 doesn't support them. You should document the default value in the JSDoc
```js

/**
 * @param {number} [param=10] - this is default param.
 */
function myFunc(param){...}
```

lib/errors/index.js
```diff
@@ -226,21 +226,96 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+    const
```
felixfbecker: one `const` per line

lib/errors/index.js
```diff
@@ -226,21 +226,96 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+      return '';
+    }
+
+    return (useNS ? [type, key].join(NSSep) : key).trim().toLowerCase();
   }
 }
-exports.ValidationErrorItem = ValidationErrorItem;
+exports.ValidationErrorItem       = ValidationErrorItem;
+
+ValidationErrorItem.ORIGINS         = Object.freeze({
```
felixfbecker: we don't usually freeze "enum" objects, I don't think we should do here

lib/errors/index.js
```diff
@@ -226,21 +226,96 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+  // fValidation errors that orignate from checks thats Sequelize does based on 'core' attribute definitions (e.g. "notNull")
+  CORE                            : 'CORE',
+  // Validation errors that originate from the storage engine
+  DB                              : 'DB',
+  // Validation errors that originate from validator functions (both built-in and custom) defined for a given attribute
+  FUNCTION                        : 'FUNCTION'
+});
+
+ValidationErrorItem.TypeStringMap = {
```
felixfbecker: why do we need a map like this?

lib/errors/index.js
```diff
@@ -226,21 +226,96 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
 }
-exports.ValidationErrorItem = ValidationErrorItem;
+exports.ValidationErrorItem       = ValidationErrorItem;
+
+ValidationErrorItem.ORIGINS         = Object.freeze({
+  // fValidation errors that orignate from checks thats Sequelize does based on 'core' attribute definitions (e.g. "notNull")
+  CORE                            : 'CORE',
+  // Validation errors that originate from the storage engine
+  DB                              : 'DB',
+  // Validation errors that originate from validator functions (both built-in and custom) defined for a given attribute
```
felixfbecker: use docblocks so they add up in ESDoc. You can generate docs with `npm run docs`

test/integration/error.test.js
```diff
@@ -84,15 +84,102 @@ describe(Support.getTestDialectTeaser('Sequelize Errors'), () => {
...
+      const error = new Sequelize.ValidationErrorItem('error!', 'FUNCTION', 'foo', 'bar', null, 'klen', 'len', [4]);
+
+      expect(() => error.getValidatorKey(false, {}))       .to.not.throw();
+      expect(() => error.getValidatorKey(false, []))       .to.not.throw();
+      expect(() => error.getValidatorKey(false, null))     .to.not.throw();
+      expect(() => error.getValidatorKey(false, ''))       .to.not.throw();
+      expect(() => error.getValidatorKey(false, false))    .to.not.throw();
+      expect(() => error.getValidatorKey(false, true))     .to.not.throw();
+      expect(() => error.getValidatorKey(false, undefined)).to.not.throw();
+      expect(() => error.getValidatorKey(true, undefined)) .to.not.throw(); // undefined will trigger use of function parameter default
```
felixfbecker: aren't these already covered in the test above?

test/integration/model/create.test.js
```diff
@@ -1639,8 +1648,12 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         ], { validate: true }).catch(errors => {
           expect(errors).to.be.instanceof(Promise.AggregateError);
           expect(errors).to.have.length(2);
+
+          const e0name0 = errors[0].errors.get('name')[0];
+
           expect(errors[0].record.code).to.equal('1234');
-          expect(errors[0].errors.get('name')[0].type).to.equal('notNull Violation');
+          expect(e0name0.originalType || e0name0.type).to.equal('notNull Violation');
```
felixfbecker: I think this needs to be reverted

test/integration/model/create.test.js
```diff
@@ -956,7 +956,15 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       return UserNull.sync({ force: true }).then(() => {
         return UserNull.create({ username: 'foo2', smth: null }).catch(err => {
           expect(err).to.exist;
-          expect(err.get('smth')[0].path).to.equal('smth');
+
+          const smth1 = err.get('smth')[0] || {};
+
+          expect(smth1.path).to.equal('smth');
+          expect(smth1.originalType || smth1.type).to.match(/notNull Violation/);
```
felixfbecker: I think this needs to be reverted

lib/errors/index.js
iamjochem: will do. after dinner :-)

lib/errors/index.js
iamjochem: if you object strongly I'll remove it ... personally I like the idea of not being able to mutate such a thing from anywhere (i.e. outside the module)

lib/errors/index.js
iamjochem: to map current `type` string like 'nOT null VALIdation' to the new `origin` values - I have left the current instantiations as they are so that we can continue to have the original values for `type`.

test/integration/model/create.test.js
iamjochem: ah, right - the test actually works because there is no `originalType` but to be really correct it should read `smth1.type || smth1.origin` (to reflect what is going on inside the actual class)

I'll fix this too

test/integration/error.test.js
iamjochem: your question is not clear to me (above what?) but I think the answer is no (tests should not assume that a function treats all falsey/truthy values for a given parameter as equivalent).

test/integration/model/create.test.js
iamjochem: ah crap - same as the other test you mentioned, I'll fix this in the same way.

lib/errors/index.js
iamjochem: I have no idea how to do that, changing to a doc-block does not add those comments to the relevant generated esdoc file. please advise.

iamjochem:
hey Felix, 
I added a commit to address most of your feedback from the last review - I have not yet removed the `Object.freeze` call yet (waiting on feedback about that) and I failed to get esdoc to include my comments related to `ValidationErrorItem.ORIGINS` in it's output files (please help with the correct comment syntax!)

test/integration/error.test.js
felixfbecker: The test is, per description, for the error case. But `.to.not.throw` doesn't test an error case obviously. This should be covered by the test for the normal case.

lib/errors/index.js
felixfbecker: see https://sourcegraph.com/github.com/sequelize/sequelize/-/blob/lib/query-types.js#L3-22:4

lib/errors/index.js
felixfbecker: You could make that argument to almost every object in the codebase

test/integration/error.test.js
iamjochem: the test is testing the behaviour of `ValidationErrorItem.getValidatorKey()` with respect to it's namespace behaviour, the test description is:

> SequelizeValidationErrorItem.getValidatorKey() should throw if namespace separator is invalid (_only if NS is used & available_)

the `expect().to.not.throw` assertions cover the '_only_' clause in that description.

lib/errors/index.js
iamjochem: I know right :-) ... alot of stuff that could/should be immutable isn't ... IMO it ought to be JS best practice to use immutability where ever possible, and this seems like a good candidate (additionally it's a call that is done only once when the lib is required so there is no appreciable overhead in the call to `Object.freeze()` (if it was being done in a place that could affect, for instance, every DAO instance then obviously the possible performance ramifications would weigh alot more heavily)

but if it's keep you from signing off on this PR then just say the word and I'll remove the `Object.freeze()` call

lib/errors/index.js
felixfbecker: > I have not changed the constructor function signature to pass in an origin value because type will be deprecated and it's ctor parameter will be taken over by origin - at least that is the idea.

If `type` becomes deprecated then this maps just makes it harder to remove in the future because `origin` is coupled to it. I would instead add an additional constructor parameter for `origin`, that way in the future `type` just has to be removed and nothing else adapted.

lib/errors/index.js
```diff
@@ -226,21 +226,106 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+
+/**
+ * An enum that defines valid ValidationErrorItem `origin` values
+ *
+ * @type {Object}
+ * @property CORE       {String}  specifies errors that originate from the sequelize "core"
+ * @property DB         {String}  specifies validation errors that originate from the storage engine
+ * @property FUNCTION   {String}  specifies validation errors that originate from validator functions (both built-in and custom) defined for a given attribute
+ */
+ValidationErrorItem.ORIGINS         = Object.freeze({
```
felixfbecker: PascalCase `Origins` like `QueryTypes`

lib/errors/index.js
felixfbecker: I would prefer to keep consistency here with other objects in the codebase.

lib/errors/index.js
iamjochem: removing `type` in the 4/5 places it is being given the `ValidationErrorItem` constructor is slightly more work because you also have to change the constructor signature. 

the way it is now you simply replace the literal string value with a reference to something like `sequelizeError.ValidationErrorItem.ORIGINS.FUNCTION`, e.g. effectively remove the comment and everything before it on the changed lines: 

https://github.com/sequelize/sequelize/pull/7557/commits/9952b7c1284bf3924e397d3a77827ea42a57b14e

once you make that replacement, the map is no longer used (and `ValidationErrorItem.type` will no longer be set)

felixfbecker: I don't understand your argument. In both cases you need to update all call sites. But with your map, once you remove `type`, you will also want to remove the map if it's no longer used. That is way more code to change than simply removing the constructor parameter and a one-line assignment.

iamjochem: - it's not functionally required to remove the map or the code that uses it (it's effectively dormant), although you may "want to".
- your way requires changing the constructor and the call site.
- it's a quick and simple change either way (your use of the words "harder" and "way more code" in this context feel very exaggerated).

This feels like bikeshedding, I'm not going to change it.

lib/errors/index.js
iamjochem: I removed it (I'm assuming you meant "remove it" when you wrote "I would prefer").

lib/errors/index.js
iamjochem: done

lib/errors/index.js
felixfbecker: No change to the codebase should ever cause dead code, so removing the map together with removing `type` is a must. We need to avoid complexity wherever we can.

iamjochem: I said "dormant" not "dead" - there is a semantic difference. also the map can have a use even after the old `type` values are not longer used, for instance for mapping `type` values that exist in logged errors that stem from an older version of sequelize (i.e. before this changes landed).

removing the map is not a technical must, it is your opinion that it must be removed - the code functions just fine with it sitting there.

the [potential] future change being discussed is not complex at all.

iamjochem:
hi Felix & co, 

I've just rebased this PR to resolve conflict with master related to #6943 (sidenote: I found the conflict confusing because my PR already seemed contained the changes from #6943).

I was hoping you if could give an indication as to whether my PR might be merged (soon? at all?), and what if anything I would need to do to get to that stage?

rgds,
Jochem

> P.S. obviously everyone wants their own PR merged and it is also quite clear that you're pretty swapped in issues & PRs ... your time, & the challenges you face managing it all, is appreciated!

iamjochem:
ping!

pi-kei:
Interesting PR. Sometimes I'm missing info about exactly which validator failed.

iamjochem:
@felix - #begging for feedback (before this starts to bitrot)

iamjochem:
@eseliger - is there any chance you might be able to review this PR?

lib/errors/index.js
```diff
@@ -224,21 +224,108 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+  }
+
+  /**
+   * return a lowercase, trimmed string "key" that identifies the validator.
+   *
+   * N.B. the string will be empty if the instance has neither a valid `validatorKey` property nor a valid `validatorName` property
+   *
+   * @param   {Boolean} [useTypeAsNS=true]    - optional, defaults to TRUE, controls whether the returned value is "namespace",
+   *                                            this parameter is ignored if the validator's `type` is not one of ValidationErrorItem.Origins
+   * @param   {String}  [NSSeparato='.']      - a separator string for concatenating the namespace, must be not be empty,
```
eseliger: typo, Separato_r_

iamjochem: fixed.

test/integration/model/create.test.js
```diff
@@ -987,7 +987,15 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
-          expect(err.get('smth')[0].path).to.equal('smth');
+
+          const smth1 = err.get('smth')[0] || {};
+
+          expect(smth1.path).to.equal('smth');
+          expect(smth1.type || smth1.origin).to.match(/notNull Violation/);
+
+          /*
+          // TODO: code below specifies different tests for different dialects but all the test are exactly the same.
+                   can we remove this or do we need to add tests to actually support different dialects?
```
eseliger: @sequelize/orm Does anyone know why? seems like this is really unnecessary

iamjochem: my guess is that it was a consequence of the fact that the `ValidationErrorItem.type` property is an arbitrary string which can be set via the `ValidationErrorItem` constructor - something I consider bad-practice because it means you cannot rely on this property programmatically (much like error messages in general) - which is why I opted to implement `.origin` as a structured alternative to `.type`

*please note*: I intentionally left `.type` untouched (but marked as deprecated) so that this PR is BC ... meaning anyone who might, incorrectly ;-), be relying on `ValidationErrorItem.type` does not get bitten.

iamjochem:
hi there,

I added a small change to the PR to fix an existing bug as reported in #8512 (MySQL ANSI_MODE breaks FK error message parsing due to different identifier quotation char being used in MySQLs messages), I figure I'd fix that seeing as this PR already make changes to the relevant MySQL error-message parsing code - a unit test is included for this.

test/integration/model/create.test.js
janmeier: Nope, no idea :)

lib/errors/index.js
```diff
@@ -224,21 +224,108 @@ exports.UnknownConstraintError = UnknownConstraintError;
  * Validation Error Item
  * Instances of this class are included in the `ValidationError.errors` property.
  *
- * @param {string} message An error message
- * @param {string} type The type of the validation error
- * @param {string} path The field that triggered the validation error
+ * @param {String} message      - An error message
```
sushantdhiman: Not sure how esdocs reacts to this but it will be better to follow rest of code comment style and removed prefix `-`

lib/errors/index.js
```diff
@@ -224,21 +224,108 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
 class ValidationErrorItem {
-  constructor(message, type, path, value) {
+  constructor(message, type, path, value, inst, validatorKey, fnName, fnArgs) {
     this.message = message || '';
-    this.type = type || null;
+    this.type = null;
     this.path = path || null;
     this.value = value !== undefined ? value : null;
+
+    this.origin         = null;
```
sushantdhiman: No spacing please (I will add related eslint rule)

lib/errors/index.js
```diff
@@ -224,21 +224,108 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+          this.type = type;
+        }
+      }
+    }
+  }
+
+  /**
+   * return a lowercase, trimmed string "key" that identifies the validator.
+   *
+   * N.B. the string will be empty if the instance has neither a valid `validatorKey` property nor a valid `validatorName` property
```
sushantdhiman: Ditch N.B. use Note: instead

lib/errors/index.js
```diff
@@ -224,21 +224,108 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+    this.validatorKey   = validatorKey || null;
+    this.validatorName  = fnName || null;
+    this.validatorArgs  = fnArgs || [];
+
+    if (type) {
+      if (ValidationErrorItem.Origins[ type ]) {
+        this.origin = type;
+      } else {
+        const lowercaseType = (type + '').toLowerCase().trim();
+        const realType  = ValidationErrorItem.TypeStringMap[ lowercaseType ];
```
sushantdhiman: No space

lib/errors/index.js
```diff
@@ -224,21 +224,108 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+    this.instance       = inst || null;
+    this.validatorKey   = validatorKey || null;
+    this.validatorName  = fnName || null;
+    this.validatorArgs  = fnArgs || [];
+
+    if (type) {
+      if (ValidationErrorItem.Origins[ type ]) {
+        this.origin = type;
+      } else {
+        const lowercaseType = (type + '').toLowerCase().trim();
```
sushantdhiman: Use lodash `toLower`

lib/errors/index.js
```diff
@@ -224,21 +224,108 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+/**
+ * An enum that defines valid ValidationErrorItem `origin` values
+ *
+ * @type {Object}
+ * @property CORE       {String}  specifies errors that originate from the sequelize "core"
+ * @property DB         {String}  specifies validation errors that originate from the storage engine
+ * @property FUNCTION   {String}  specifies validation errors that originate from validator functions (both built-in and custom) defined for a given attribute
+ */
+ValidationErrorItem.Origins       = {
+  CORE                            : 'CORE',
```
sushantdhiman: No spaces in object construction

test/integration/error.test.js
```diff
@@ -84,15 +84,101 @@ describe(Support.getTestDialectTeaser('Sequelize Errors'), () => {
...
+      expect(error).to.have.property('validatorArgs', vargs);
+    });
+
+    it('SequelizeValidationErrorItem.getValidatorKey() should return a string', () => {
+      const error = new Sequelize.ValidationErrorItem('error!', 'FUNCTION', 'foo', 'bar', null, 'klen', 'len', [4]);
+
+      expect(error).to.have.property('getValidatorKey');
+      expect(error.getValidatorKey).to.be.a('function');
+
+      expect(error.getValidatorKey())           .to.equal('function.klen');
```
sushantdhiman: No spaces, please remove from all other sections you have used this pattern as well

test/integration/model/create.test.js
sushantdhiman: @iamjochem just remove this section, its in this state from at least 4 years

lib/errors/index.js
```diff
@@ -224,23 +226,110 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
  * Instances of this class are included in the `ValidationError.errors` property.
  *
- * @param {string} message An error message
- * @param {string} type The type of the validation error
- * @param {string} path The field that triggered the validation error
+ * @param {String} message An error message
+ * @param {String} type The type/origin of the validation error
+ * @param {String} path The field that triggered the validation error
+ * @param {String} value The value that generated the error
+ * @param {Object} inst optional, the DAO instance that caused the validation error
```
sushantdhiman: [inst] for optional

lib/errors/index.js
```diff
@@ -224,23 +226,110 @@ exports.UnknownConstraintError = UnknownConstraintError;
...
+
+    // This doesn't need captureStackTrace because it's not a subclass of Error
+  }
+
+  /**
+   * return a lowercase, trimmed string "key" that identifies the validator.
+   *
+   * Note: the string will be empty if the instance has neither a valid `validatorKey` property nor a valid `validatorName` property
+   *
+   * @param   {Boolean} [useTypeAsNS=true]      optional, defaults to TRUE, controls whether the returned value is "namespace",
```
sushantdhiman: No need to say optional, esdocs can parse `[]`, no need to specify default value as well

iamjochem:
@sushantdhiman - I *think* I've made all CS changes you requested, I also had to perform a merge conflict fix related to #8500 (I made a boo-boo by commiting merge conflict markers and fixed that in a subsequent commit) ... all checks pass so I'm really hoping that you are willing to merge now!

My colleague just pointed out that certain DataTypes contain their own `validate()` functions and that these functions are run during query building but not during model validation (unlike the built-in "notNull" check) ... I feel this is not correct ... at the very least these should also be run during model validation (and thereby generate the relevant `ValidationErrorItem` instances (that include all the structured info that this PR implements) ... what is your opinion on this?

iamjochem:
with regard to including `DataType.validate()` checks during actual model validation (as opposed to leaving it until a query is being built) I believe the `` method should be expanded to check for the existence of a relevant `DataType.validate()` function and call it if it exist (obviously generating a `ValidationErrorItem` if applicable), see here:

    https://github.com/sequelize/sequelize/blob/089504fe417310265c7823acb33ca6d2b0d4d2b6/lib/instance-validator.js#L320

iamjochem:
coming back to the idea of running DataType.validate function during Model validation I had a stab at it and I think it is a rather small addition, a possible patch for this could be this (performed against this PR's  branch):

```diff
diff --git a/lib/errors/index.js b/lib/errors/index.js
index 327cfef..62a59a1 100644
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -323,6 +323,7 @@ ValidationErrorItem.Origins = {
  * @type {Object}
  */
 ValidationErrorItem.TypeStringMap = {
+  'datatype violation' : 'CORE',
   'notnull violation' : 'CORE',
   'string violation' : 'CORE',
   'unique violation' : 'DB',
diff --git a/lib/instance-validator.js b/lib/instance-validator.js
index 9ef584f..b054546 100644
--- a/lib/instance-validator.js
+++ b/lib/instance-validator.js
@@ -319,7 +319,9 @@ class InstanceValidator {
    * @private
    */
   _validateSchema(rawAttribute, field, value) {
-    if (rawAttribute.allowNull === false && (value === null || value === undefined)) {
+    const isEmpty = (value === null || value === undefined);
+
+    if (rawAttribute.allowNull === false && isEmpty) {
       const validators = this.modelInstance.validators[field];
       const errMsg = _.get(validators, 'notNull.msg', `${this.modelInstance.constructor.name}.${field} cannot be null`);

@@ -333,6 +335,21 @@ class InstanceValidator {
       ));
     }

+    if (!isEmpty && _.isFunction(rawAttribute.type.validate)) {
+      try {
+        rawAttribute.type.validate(value);
+      } catch (e) {
+        this.errors.push(new sequelizeError.ValidationErrorItem(
+          e.message,
+          'datatype violation', // sequelizeError.ValidationErrorItem.Origins.CORE,
+          field,
+          value,
+          this.modelInstance,
+          'invalid_type'
+        ));
+      }
+    }
+
     if (rawAttribute.type === DataTypes.STRING || rawAttribute.type instanceof DataTypes.STRING || rawAttribute.type === DataTypes.TEXT || rawAttribute.type instanceof DataTypes.TEXT) {
       if (Array.isArray(value) || _.isObject(value) && !(value instanceof Utils.SequelizeMethod) && !Buffer.isBuffer(value)) {
         this.errors.push(new sequelizeError.ValidationErrorItem(
```

the problem with this particualar patch is that it changes behaviour subtly (although AFAICT the actual outcode is still the same ,namely you get a "validation" error and the relevant query is not performed), the changes in behaviour obviously break some tests related to expected error message content.

I second problem that I noticed is that the ability to pass in something like `this.sequelize.cast('1', 'signed')` is broken when an attribute defines it's own validators:

e.g. in the test file `test/integration/model/create.test.js`, there is the following test (which passes):

```js
    it('is possible to use casting when creating an instance', function() {
      const self = this,
        type = dialect === 'mysql' ? 'signed' : 'integer';
      let match = false;

      return this.User.create({
        intVal: this.sequelize.cast('1', type)
      }, {
        logging(sql) {
          expect(sql).to.match(new RegExp("CAST\\(N?'1' AS " + type.toUpperCase() + '\\)'));
          match = true;
        }
      }).then(user => {
        return self.User.findById(user.id).then(user => {
          expect(user.intVal).to.equal(1);
          expect(match).to.equal(true);
        });
      });
    });
```

base on the following model definition: 

```js
      this.User = this.sequelize.define('User', {
        username: DataTypes.STRING,
        secretValue: DataTypes.STRING,
        data: DataTypes.STRING,
        intVal: DataTypes.INTEGER,
        theDate: DataTypes.DATE,
        aBool: DataTypes.BOOLEAN,
        uniqueName: { type: DataTypes.STRING, unique: true }
      });
```

If I change the model definition to the following the test fails miserably (because the field value is an object describing the 'cast' definition):

```js
      this.User = this.sequelize.define('User', {
        username: DataTypes.STRING,
        secretValue: DataTypes.STRING,
        data: DataTypes.STRING,
        intVal: {
          type : DataTypes.INTEGER,
          validate : {
            isInt: true
          }
        },
        theDate: DataTypes.DATE,
        aBool: DataTypes.BOOLEAN,
        uniqueName: { type: DataTypes.STRING, unique: true }
      });
```

this failure occurs for any field with a validator when any kind of "query specification object" is inserted into a model (e.g. `this.sequelize.fn('upper', 'sequelize')`) ... to be honest I feel this is so broken that you seriously need to consider what/how to tackle it, it is definitely not something that should be tackled in the context of this PR.

I am not 100% sure where these "query specification objects" are allowed, It would seem that if anywhere it should only be in the context of Model methods (not instance methods); even if attribute validators were skipped when a `dataValue` item [seems to] contain an "query specification object" then you would have created an mechanism for by passing attribute validation (which seems to go against the concept of using a model in the first place). 

My advice would be to make the use of "query specification objects" as model values to be illegal (I just don't think you can reconcile model validaiton and there usage), and limit the use of "query specification objects" to `where` & `replacement` options (for `find*` methods and raw-queries respectively)

sushantdhiman:
It was a good point raised by your colleague and your patch looks good to me.

I agree these query level specification object can't be handled in validation step, their value is not realized util query has performed. So what should we do in such cases?

I suggest

1) As you mentioned, ignore any attribute whose value cant be realized without performing query. These are mostly sequelize methods I think. ( @janmeier do we have a way to check if method is sequelize method/fn ? )

2) Issue a warning when this happens, may be something like `Model.type validation was ignored, value passed to it can not be reliably checked without executing query`

Further I dont think we should drop these construct just because they cant be validated in JS, they have a usecase which cant be replaced.

iamjochem:
@sushantdhiman.

1. things generated by `sequelize.fn()` et al. are instances of `SequelizeMethod` ([see here where it is used](https://github.com/sequelize/sequelize/search?utf8=%E2%9C%93&q=SequelizeMethod&type=)), so it can be checked for with an `instanceof`

2. I undertand your point of view regarding the use-case, but I personally disagree - I feel it is counter to the concept of the model abstraction to be able to push DB "commands" through the model via it's `dataValues`, it "breaks" the model abstraction. I do think the functionality has good use in construction of `where` clauses and custom queries ... however I won't/don't personally use this functionality in this way so I'm not too bothered that you want to keep it in place :-)

I would like to take a shot at adding another PR once this one is merged to address the following:

1. skip validation when dataValue is instance of `SequelizeMethod` - seems very simple to implement.
2. perform DataType.validate calls during the "validate" stage/call of an instance (as opposed to waiting unitl query building time) - also seems relatively simple but the remifications are somewhat large because it changes what kind of errors are generated (kind of a breaking change and would also require updating quite a few tests) ... it might be smart to put this logic under an `options` flag (a bit like `invidivualHooks`) 

I hope you can get this PR merged shortly :-)

iamjochem:
joy!

PR #7593, opened 2017-05-02
documentation: update transactions.md to use `cls` property instead of `useCLS()`

pcsanwald:
### Description of change

The documentation on `transactions.md` doesn't reflect the latest sequelize: `useCLS` has been replaced with the `cls` property. This pull request updates the documentation around this.


janmeier:
Actually it's the other way around - `useCLS` is the preferred way in v4 https://github.com/sequelize/sequelize/blob/4442dfc47e16a0bcd2d480ade1583094d4d8c9ac/lib/sequelize.js#L1005

PR #7598, opened 2017-05-03
Adding description of 'fields' option for 'validate'

rodic:
Small change in documentation as discussed on slack with @janmeier.


janmeier:
:+1:

PR #7619, opened 2017-05-10
Swap rejectOnEmpty evaluation order

alexlatchford:
Allow for query-level options to override model-level ones.

### Description of change

Currently there isn't a way to override a model-level value for `rejectOnEmpty` with a query-level value. This fixes that.


sushantdhiman:
We can't accept this PR in v3 as this may be a breaking change but you can add this change to master, provided with a test

PR #7627, opened 2017-05-12
Polish docs

felixfbecker:
- Use ES6 in docs, as supported by our minimal version (Node 4). That means arrow functions and `const`.
- Make model names PascalCase. Models are classes in v4 and the name becomes the class name, PascalCase is the convention for that.
- Make indentation consistently 2 spaces.

### DIFF
--- a/docs/associations.md
+++ b/docs/associations.md
@@ -12,8 +12,8 @@ BelongsTo associations are associations where the foreign key for the one-to-one
 A simple example would be a **Player** being part of a **Team** with the foreign key on the player.
 
 ```js
-var Player = this.sequelize.define('player', {/* attributes */})
-  , Team  = this.sequelize.define('team', {/* attributes */});
+const Player = this.sequelize.define('player', {/* attributes */});
+const Team  = this.sequelize.define('team', {/* attributes */});
 
 Player.belongsTo(Team); // Will add a teamId attribute to Player to hold the primary key value for Team
 ```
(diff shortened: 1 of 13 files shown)


felixfbecker:
@sushantdhiman the docs used both PascalCase and camelCase model names, it's not consistent. I agree that attributes should always be camelCase, but _classes_ should always be PascalCase. It would be better than to always specify an alias.

sushantdhiman:
@felixfbecker In that case please update docs with proper foreign key names as well, because even though models are now Pascal cased, auto generated foreign keys aren't.

felixfbecker:
Will do in a second PR, removed the commit for now

docs/associations.md
```diff
@@ -244,10 +244,17 @@ Person.belongsToMany(Person, { as: 'Children', through: 'PersonChildren' })
...
+const User = sequelize.define('user', {})
+const Project = sequelize.define('project', {})
+const UserProjects = sequelize.define('userProjects', {
     status: DataTypes.STRING
+=======
+const User = sequelize.define('User', {})
+const Project = sequelize.define('Project', {})
+const UserProjects = sequelize.define('UserProjects', {
+  status: DataTypes.STRING
+>>>>>>> 1a74b8b3... fixup
```
sushantdhiman: Opps

felixfbecker: 👀

docs/hooks.md
```diff
@@ -245,23 +245,25 @@ Model.destroy({ where: {username: 'Tom'}} /*whereClause argument*/)
...
+```js
 // Bulk updating existing users with updatesOnDuplicate option
-Users.bulkCreate([{ id: 1, isMemeber: true},
-                 { id: 2, isMember: false}],
-                 { updatesOnDuplicate: ['isMember']})
+Users.bulkCreate([
+  { id: 1, isMemeber: true},
+  { id: 2, isMember: false},
+  { updatesOnDuplicate: ['isMember']
+});
```
sushantdhiman:
```js
Users.bulkCreate([
 { id: 1, isMemeber: true},
 { id: 2, isMember: false}
], {
 updatesOnDuplicate: ['isMember']
});

PR #7630, opened 2017-05-12
Create TOC for class methods, cleanup titles

danawoodman:
Finding content within the doc is a pain, so adding a link to their sections in the markdown should help. Also styled the class methods in the headings on the page.


felixfbecker:
Why is the sidebar not sufficient for that?

image

danawoodman:
@felixfbecker the sidebar doesn't exist when viewing on Github, which is where I noticed the lack of links. Harm in having them in both places?

felixfbecker:
I don't think we should optimize the read-on-GitHub use case when we have good hosted docs, and it is redundant on the hosted docs. The docs are generated and a manual ToC would have to be maintained and can get out of sync. It is also confusing if it is only there for some pages.

Is there any reason that prefer GH over the hosted docs?

danawoodman:
My main argument would be that this:



is far less legible and easy to navigate than this:



If there is a way to instead change how the automatically generated docs are created to make the actual method names more easily visually identified, that could be an option.

Also, the syntax highlighting of the method names in the heads would improve readability.

Thoughts?

felixfbecker:
We can definitely merge in the code highlighting, but I am against a redundant ToC. Note that these are the guides, not the API (method) reference. The API reference has pretty good ToCs imo: http://docs.sequelizejs.com/class/lib/model.js~Model.html

sushantdhiman:
I agree with @felixfbecker why have redundant ToC when you have side bar. We can accept the code highlighting part, but please remove duplicate ToC

danawoodman:
@felixfbecker done

danawoodman:
Thanks for the quick merge!

PR #7639, opened 2017-05-15
Add arrow-parens rule to ESLint

felixfbecker:
This comes up a lot in PRs. Arrow functions with a single argument do not need to specify the parenthesis, so they shouldn't. Example:
```js
[1, 2, 3].map(x => x * 2)
// looks a lot cleaner than
[1, 2, 3].map((x) => x * 2)
```
Just like you wouldn't for example wrap every assignment expression in an extra set of parenthesis:
```ts
const x = (y + 1);
```
you shouldn't do it for arrow functions if it's not needed.

Just added the rule and ran `eslint --fix`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -13,6 +13,7 @@
     "space-before-function-paren": ["warn", "never"],
     "keyword-spacing": ["warn"],
     "prefer-arrow-callback": "error",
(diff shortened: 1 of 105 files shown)


janmeier:
:+1:

PR #6297, opened 2016-07-18
Fixed attribute:[] issues when using them in include or through

sushantdhiman:
### Pull Request check-list

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Have you added an entry under `Future` in the changelog?

### Description of change

Using `through.attributes:[]` or `include.attributes: []` doesn't ignore any attributes, It throws error in most cases. Fixed it by checking if parent include exists before trying to push child include in it.

Closes #5078
Closes #4222
Closes #5958
Closes #5590
Closes #6139
Closes #4866
Closes #6242

PR #7648, opened 2017-05-17
Fix that nasty undefined $parent bug for real

louisremi:
Hi,

PR [#6297](https://github.com/sequelize/sequelize/pull/6297) fixed a bunch of bugs appearing when using `attribute: []` in `include` or `through`. For some reason, the author of the patch didn't check for the existence of `$parent` at the right place in the code, and the bug was still present for a few edge-cases.

I'm not sure exactly how my code is different from the test-cases provided in PR #6297, but this patch fixes the undefined $parent errors I encountered.

If the test case is mandatory, I can try to spend some additional time on reducing my code to see why things went wrong for me.


sushantdhiman:
> If the test case is mandatory, I can try to spend some additional time on reducing my code to see why things went wrong.

Yeah, test case will be really nice to see what exactly causing this issue.

louisremi:
Unit tests still missing. Please don't close, the bug is still here.

louisremi:
Unit tests still missing. Please don't close, the bug is still here.

sushantdhiman:
@louisremi I am going to close it, please reopen when you have added an integration test for case that fails

Issue #7641, opened 2017-05-15
Soft-deleted entities are included by default when loading associations in Sequelize v4


adamjmurray:
## What you are doing?
I am evaluating Sequelize v4 by installing via `npm install sequelize@4.0.0-2`. A lot of tests in our app's test suite started failing with this version of Sequelize. One of the problems I am facing is that soft-deleted associated entities (we are using `paranoid: true` models) in a one-to-many relationship are being returned when I include them in a query. This appears to be a regression from Sequelize v3, where soft-deleted entities are never returned by default.

Here is an example: 

```js
var Sequelize = require('sequelize');
var sequelize = new Sequelize('mysql://root@localhost/sequelize_test');

var User = sequelize.define('user', {
  name: {
    type: Sequelize.STRING
  }
});

var Role = sequelize.define('role', {
  name: {
    type: Sequelize.STRING
  }
}, { paranoid: true });

Role.belongsTo(User);
User.hasMany(Role);

var roles = [];

sequelize.sync({ force: true })
  .then(function() {
    return Role.create({ name: 'Role 1' });
  })
  .then(function(role) {
    roles.push(role);
    return Role.create({ name: 'Role 2' });
  })
  .then(function(role) {
    roles.push(role);
    return User.create({ name: 'John Doe' });
  })
  .then(function(user) {
    return user.setRoles(roles);
  })
  .then(function() {
    return roles[0].destroy();
  })
  .then(function() {
    return User.findAll({ include: [{ model: Role }] });
  })
  .then(function(users) {
    console.log(JSON.stringify(users, null, 2));
  });

```

## What do you expect to happen?
This is the output I got with Sequelize v3, and what I still expect to happen with v4:
```
[
  {
    "id": 1,
    "name": "John Doe",
    "createdAt": "2017-05-15T23:03:14.000Z",
    "updatedAt": "2017-05-15T23:03:14.000Z",
    "roles": [
      {
        "id": 2,
        "name": "Role 2",
        "createdAt": "2017-05-15T23:03:14.000Z",
        "updatedAt": "2017-05-15T23:03:14.000Z",
        "deletedAt": null,
        "userId": 1
      }
    ]
  }
]
```

## What is actually happening?
In Sequelize v4, the soft-deleted role is returned:
```
[
  {
    "id": 1,
    "name": "John Doe",
    "createdAt": "2017-05-15T23:02:36.000Z",
    "updatedAt": "2017-05-15T23:02:36.000Z",
    "roles": [
[...]

PR #7672, opened 2017-05-22
Fix documentation of include.paranoid

LaurentVB:
Hello.

It looks like the documentation for `include.paranoid` got it backwards. This fixes it.
Should probably be merged to v3 branch too.

Regards,

Laurent

### DIFF
--- a/docs/models-usage.md
+++ b/docs/models-usage.md
@@ -646,14 +646,14 @@ User.findAll({ include: [{ all: true }]});
 
 ### Including soft deleted records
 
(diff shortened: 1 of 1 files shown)


sushantdhiman:
I think its same as https://github.com/sequelize/sequelize/issues/7641

LaurentVB:
Sorry, I don't think it is.
I'm reporting a documentation issue in the docs/models-usage.md file, but the framework works as expected (when expectations are set by the API documentation).

sushantdhiman:
Hi @LaurentVB , You are correct. It should say `false` if want to load deleted records. I am sorry for dismissing it earlier :)

PR #7674, opened 2017-05-22
Default to aliases for 'createdAt', 'updatedAt', and 'deletedAt' in options

davidadas:
### Description of change

We are currently using `sequelize-auto` to generate the majority of our DB models.  Our generated models are frequently created with aliases to `createdAt`, `updatedAt`, etc in the model options field (as we have them with a different name).  This has been causing us problems.

We are, however, a dozen or so minor versions behind so please let me know if this is actually fixed somewhere else and I can begin updating our services.

### DIFF
--- a/test/integration/model/create.test.js
+++ b/test/integration/model/create.test.js
@@ -670,6 +670,42 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       });
     });
 
+    t('works with custom timestamps which alias an existing field', function() {
+      const User = this.sequelize.define('User', {
+        username: DataTypes.STRING,
+        date_of_birth: DataTypes.DATE,
+        email: DataTypes.STRING,
+        password: DataTypes.STRING,
+        createdTime: {
+          type: DataTypes.DATE,
+          allowNull: true,
+          defaultValue: DataTypes.NOW,
+          field: 'created_time'
+        },
+        updatedTime: {
+          type: DataTypes.DATE,
+          allowNull: true,
+          defaultValue: DataTypes.NOW,
+          field: 'updated_time'
+        }
+      }, {
+        createdAt: 'createdTime',
+        updatedAt: 'updatedTime',
+        tableName: 'users',
+        underscored: true,
+        freezeTableName: true,
+        force: false
+      });
+
+      return this.sequelize.sync({force: true}).then(() => {
+        return User.create({}).then(user => {
+          expect(user).to.be.ok;
+          expect(user.created_time).to.be.ok;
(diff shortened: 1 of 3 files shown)


test/integration/model/create.test.js
```diff
@@ -670,6 +670,42 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+        tableName: 'users',
+        underscored: true,
+        freezeTableName: true,
+        force: false
+      });
+
+      return this.sequelize.sync({force: true}).then(() => {
+        return User.create({}).then(user => {
+          expect(user).to.be.ok;
+          expect(user.created_time).to.be.ok;
```
sushantdhiman: This is incorrect, field should not be returned it should be attribute name which in this case is `createdTime`

sushantdhiman:
@davidadas I am not sure what your usecase is but why you want fields to be returned when creating model.

As far as I understood, I think you want to reference a field with different attribute name for timestamp fields. Which is already possible (as tested in master)

```js
'use strict';

/*
* Copy this file to ./sscce.js
* Add code from issue
* npm run sscce-{dialect}
*/

const Sequelize = require('./index');
const sequelize = require('./test/support').createSequelizeInstance();


  username: Sequelize.STRING,
  date_of_birth: Sequelize.DATE,
  email: Sequelize.STRING,
  password: Sequelize.STRING,
  createdTime: {
    type: Sequelize.DATE,
    allowNull: true,
    defaultValue: Sequelize.NOW,
    field: 'created_time'
  },
  updatedTime: {
    type: Sequelize.DATE,
    allowNull: true,
    defaultValue: Sequelize.NOW,
    field: 'updated_time'
  }
}, {
  createdAt: 'createdTime',
  updatedAt: 'updatedTime',
  tableName: 'users',
  underscored: true,
  freezeTableName: true,
  force: false
});

return sequelize.sync({force: true}).then(() => {
  return User.create({}).then(user => {
    console.log(user.toJSON());
  });
});
```

```
± |master U:2 ?:1 ✗| → DIALECT=sqlite node sscce_template.js 
{ createdTime: 2017-05-27T05:07:30.000Z,
  updatedTime: 2017-05-27T05:07:30.000Z,
  id: 1 }
```

sushantdhiman:
Closing for now as per my reasoning above https://github.com/sequelize/sequelize/pull/7674#issuecomment-304428401

Let me know if you are trying to do something else :)

davidadas:
@sushantdhiman t'was good enough reasoning for me :)

PR #7230, opened 2017-02-10
Added feature to alter tables on sync. Fixes #537

meyer9:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [X] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does your issue contain a link to existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [X] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Have you added an entry under `Future` in the changelog?

### Description of change

This patch should provide an easy way to migrate the database automatically on the fly in a development environment. This is my first PR to an open source project, so let me know if anything should be structured differently.

This patch allows for altering existing tables when syncing. This will sometimes destroy data if there is no direct conversion from the existing type to the new type. When removing a column, the data in the deleted column will be destroyed. When adding a column, the data in the added column will be the default value.

~~Also, where should I put the tests as I couldn't find `sync` being tested anywhere I could only find sync being used in the hooks test and in most tests as a prerequisite.~~

Everything is tested and ready to merge.

PR #7676, opened 2017-05-23
Fix alter option Fixes #537

### LINKED ISSUE #537: Sequelize.sync() should support ALTER statements instead of full DROP+CREATE statements
It would be nice if we could call sync() and have it compare the model schema to the database schema  and make only the changes necessary using ALTER statements. The major benefit is preservation of data. If implemented correctly, this could probably eliminate most of the need for migration files.

agramdass:
### Pull Request check-list

### Description of change

This patch improves on the changes that were done for the alter option for model sync. #537 #7230  
- It fixes the column/model name mapping (i.e. for example model = createdAt and DB = created_at).
- Removes the default dropping of database columns for any column not present in the model and adds an additional option 'dropColumn' to support it.
- Checks constraints (for postgres) to ensure valid alter of unique columns + addition of foreign keys
- Fixes showConstraintQuery for postgres (was ignoring schema as part of table name)
- Fixes alter option

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -297,20 +297,30 @@ const QueryGenerator = {
         definition += ' USING (' + this.quoteIdentifier(attributeName) + '::' + this.pgEnumName(tableName, attributeName) + ')';
       }
 
-      if (definition.match(/UNIQUE;*$/)) {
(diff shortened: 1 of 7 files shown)


Wadeck:
Thank you for the issue opening / code proposal, you saved me lots of time ! I also encounter the issue with the field attribute that is not used (using column name instead).

Issue #7680, opened 2017-05-24
Unreachable code in lib/dialects/postgres/query.js


banyudu:
__Dialect:__  postgres 

This line is unreachable.
https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/query.js#L333

PR #7681, opened 2017-05-24
refs #7680 remove unreachable break in switch

banyudu:
### Pull Request check-list

### Description of change

Remove unreachable break in switch, as described in https://github.com/sequelize/sequelize/issues/7680
This `break` is unreachable because `return` exists in both `if` and `else`.


sushantdhiman:
Thanks, but I preferred non branching fix https://github.com/sequelize/sequelize/pull/7692 :)

PR #7626, opened 2017-05-12
Closes #[7625] Ignore generic-pool release/destroy exceptions

jaredjenkins:
Cleaning up dead connections to MySQL is a bit messy: the connection can die before/after it has been returned to the pool. This prevents generic-pool from throwing errors when we attempt to clean up connections that we know to be dead, but have already cleaned up.

Closes #7625

PR #7698, opened 2017-05-27
fix: pooling fails to handle disconnection

sushantdhiman:
Continuation of ideas from https://github.com/sequelize/sequelize/pull/7626 

**Notable changes**
- No destroy from pool (Same as #7626) when handling disconnects
- No timers for watching eventual pool errors
- Wont crash when ETIMEOUT error occurs eventually

```js
'use strict';

/*
 * Copy this file to ./sscce.js
 * Add code from issue
 * npm run sscce-{dialect}
 */

const Sequelize = require('./index');
// const sequelize = require('./test/support').createSequelizeInstance();


let dbConfig = {
  username: 'sequelize_test',
  password: 'sequelize_test',
  port: '8999',
  dialect: 'mysql',
  syncOnAssociation: false,
  host: '127.0.0.1',
  seederStorage: 'sequelize',
  pool: {
    min: 1,
    max: 5,
    acquire:  10000,
  },
  dialectOptions: {
    connectTimeout: 10000
  }
}

let pool = new Sequelize(dbConfig.database, dbConfig.username, dbConfig.password, dbConfig);

setInterval(() =>
  pool.query("SELECT SLEEP(1);").catch((e) => {
    console.log('Got an exception!', e)
  })
  , 2000
)

```

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -266,8 +266,21 @@ class ConnectionManager {
     }
 
     return promise.then(() => {
-      const connectionPromise = this.pool.acquire(options.priority, options.type, options.useMaster);
-      return this._watchPoolError(connectionPromise).tap(() => { debug('connection acquired'); });
+      return Promise.race([
+        this.pool.acquire(options.priority, options.type, options.useMaster),
+        new Promise((resolve, reject) =>
+        timers.setTimeout(() => {
+          if (this.poolError) {
+            reject(this.poolError);
+          }
+        }, 0))
+      ])
+      .tap(() => { debug('connection acquired'); })
+      .catch(e => {
+        e = this.poolError || e;
+        this.poolError = null;
+        throw e;
+      });
     });
   }
 
@@ -290,31 +303,6 @@ class ConnectionManager {
   _validate(connection) {
     if (!this.dialect.connectionManager.validate) return true;
     return this.dialect.connectionManager.validate(connection);
-  }
-
-  _watchPoolError(acquirePromise) {
-    return new Promise((resolve, reject) => {
-      if (acquirePromise.isFulfilled()) {
-        this.poolError = null;
-        return resolve(acquirePromise.value());
-      }
-
-      if (acquirePromise.isRejected()) {
-        this.poolError = null;
-        return reject(acquirePromise.reason());
-      }
-
-      if (this.poolError) {
-        acquirePromise.cancel();
-        const error = this.poolError;
-        this.poolError = null;
-        return reject(error);
-      }
-
-      timers.setTimeout(() => {
-        resolve(this._watchPoolError(acquirePromise));
-      }, 0);
-    });
   }
 }
 
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -6,6 +6,7 @@ const _ = require('lodash');
 const Utils = require('../../utils');
 const debug = Utils.getLogger().debugContext('pool');
 const semver = require('semver');
+const timers = require('timers');
 
 const defaultPoolingConfig = {
   max: 5,
@@ -290,23 +291,32 @@ class ConnectionManager {
 
   _acquire(options) {
     const connectionPromise = this.pool.acquire(options.priority, options.type, options.useMaster);
+    return this._watchPoolError(connectionPromise).tap(() => { debug('connection acquired'); });
+  }
 
-    return Promise.delay(0)
-      .then(() => {
-        if (connectionPromise.isRejected()) {
-          this.poolError = null;
-          return Promise.reject(connectionPromise.reason());
-        }
+  _watchPoolError(acquirePromise) {
+    return new Promise((resolve, reject) => {
+      if (acquirePromise.isFulfilled()) {
+        this.poolError = null;
+        return resolve(acquirePromise.value());
+      }
 
-        if (this.poolError) {
-          connectionPromise.cancel();
-          const error = this.poolError;
-          this.poolError = null;
-          return Promise.reject(error);
-        }
+      if (acquirePromise.isRejected()) {
+        this.poolError = null;
+        return reject(acquirePromise.reason());
+      }
 
-        return connectionPromise;
-      });
+      if (this.poolError) {
+        acquirePromise.cancel();
+        const error = this.poolError;
+        this.poolError = null;
+        return reject(error);
+      }
+
+      timers.setTimeout(() => {
+        resolve(this._watchPoolError(acquirePromise));
+      }, 0);
+    });
   }
 }
 
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -264,16 +264,7 @@ class ConnectionManager {
       promise = Promise.resolve();
     }
 
-    return promise.then(() => Promise.race([
-      this.pool.acquire(options.priority, options.type, options.useMaster)
-        .tap(() => {
-          debug('connection acquired');
-          this.poolError = null;
-        }),
-      this._watchPoolErrors()
-    ])).catch(e => {
-      throw this.poolError || e;
-    });
+    return promise.then(() => this._acquire(options));
   }
 
   releaseConnection(connection) {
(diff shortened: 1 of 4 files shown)


sushantdhiman:
Also linking @jaredjenkins for external review

lib/dialects/mysql/connection-manager.js
```diff
@@ -92,77 +92,69 @@ class ConnectionManager extends AbstractConnectionManager {
...
+          case 'ENOTFOUND':
+            throw new SequelizeErrors.HostNotFoundError(err);
+          case 'EHOSTUNREACH':
+            throw new SequelizeErrors.HostNotReachableError(err);
+          case 'EINVAL':
+            throw new SequelizeErrors.InvalidConnectionError(err);
+          default:
+            throw new SequelizeErrors.ConnectionError(err);
+        }
+      } else {
```
felixfbecker: else is unneeded

lib/dialects/abstract/connection-manager.js
```diff
@@ -312,6 +288,36 @@ class ConnectionManager {
     if (!this.dialect.connectionManager.validate) return true;
     return this.dialect.connectionManager.validate(connection);
   }
+
+  _acquire(options) {
+    const connectionPromise = this.pool.acquire(options.priority, options.type, options.useMaster);
+    return this._watchPoolError(connectionPromise).tap(() => { debug('connection acquired'); });
+  }
+
+  _watchPoolError(acquirePromise) {
```
sushantdhiman: @felixfbecker I agree with the general ideology but this promise in some condition will never be resolved, thats why we used to watch it using timers initially, now using recursive watcher method. There is no other way around, we have to pass promise to method which will call itself with same promise

lib/dialects/abstract/connection-manager.js
```diff
@@ -312,6 +288,36 @@ class ConnectionManager {
...
+    return new Promise((resolve, reject) => {
+      if (acquirePromise.isFulfilled()) {
```
sushantdhiman: For now there isn't, but we are already tied to bluebird and we should use itsl API where required. When we will eventually migrate to `async await` we can change this segment with reset of codebase as well

lib/dialects/mysql/connection-manager.js
```diff
@@ -92,77 +92,69 @@ class ConnectionManager extends AbstractConnectionManager {
...
         connection.removeListener('connect', connectHandler);
+
+        if (config.pool.handleDisconnects) {
+          debug(`connection error ${e.code}`);
+
+          switch (e.code) {
+            case 'PROTOCOL_CONNECTION_LOST':
+              return;
+            default:
+          }
```
sushantdhiman: Yea :) there were some other cases initially, in the end just this one remained

lib/dialects/mysql/connection-manager.js
sushantdhiman: Old code segment I just changed indentation, but will remove

lib/dialects/abstract/connection-manager.js
felixfbecker: It should still be possible without. What do you mean with "watching" the promise? If the promise sometimes stays pending forevern and you just need a timeout, that's just
```ts
Promise.race([p, new Promise(resolve => setTimeout(resolve, 5000))])
```

lib/dialects/abstract/connection-manager.js
```diff
@@ -265,31 +265,23 @@ class ConnectionManager {
...
-          if (connectionPromise.isFulfilled()) {
-            resolve(connectionPromise);
-            debug('connection acquire');
-            evictTimer = true;
-          } else if (this.poolError) {
+    return promise.then(() => {
+      return Promise.race([
+        this.pool.acquire(options.priority, options.type, options.useMaster),
+        new Promise((resolve, reject) =>
+        timers.setTimeout(() => {
```
felixfbecker: what is `timers.setTimeout()`? I cannot find `timers` in our package.json and the package on npm doesn't export anything like that: https://www.npmjs.com/package/timers

lib/dialects/abstract/connection-manager.js
```diff
@@ -265,31 +265,23 @@ class ConnectionManager {
...
-            evictTimer = true;
           }
-
-          if (evictTimer) {
-            timers.clearInterval(connectionTimer);
-          }
-        }, 0);
-      })
-    );
+        }, 0))
```
felixfbecker: The logic is still hard to follow - could you add some comments? We try to acquire from the pool, but if in the next event loop tick `this.poolError` is not `null`, then we reject with that. What sets this? Is it guaranteed that that it would definitely be set on the next tick?

lib/dialects/abstract/connection-manager.js
sushantdhiman: `timers` is an internal node js utility

felixfbecker: Ah, I never saw that required explicitly. Docs say so too:

> The timer module exposes a global API for scheduling functions to be called at some future period of time. Because the timer functions are globals, there is no need to call require('timers') to use the API.

lib/dialects/abstract/connection-manager.js
sushantdhiman: When we acquire from pool, error is not simply returned by rejection, but instead its fired as an event on pool object (this happens when a connection creation fails, which is most significant case for orms)

So we try to acquire, but also keep watching this "pool error" (set in `this.poolError`), if we see this error first, we know connection creation fails and we reject with that reason.

felixfbecker: If it is emitted as an event, then this shouldn't check a property once at the next tick, but instead attach a listener to that event:
```ts
let errorListener;
Promise.race([
  aquire(...),
  new Promise((resolve, reject) => {
    errorListener = reject;
    connection.on('error', errorListener);
  })
]).finally(() => {
  connection.removeListener('error', errorListener)
})
```
Only checking the error property on the next tick, once, is brittle because if the error event happens to be emitted on a later tick then it is not caught.

sushantdhiman: This event is emitted by pooling library, not internal db layer. Also creating an event handler on each acquire call is not good. With current solution we will have only one event handler for each Sequelize constructor

Basically its same just as you suggested, but instead of creating listener every time, we have a listener per constructor which keep populating `this.poolErr`

lib/dialects/abstract/connection-manager.js
sushantdhiman: They are really useful when sinon like libraries keep overriding global versions of those methods

lib/dialects/abstract/connection-manager.js
felixfbecker: It doesn't matter where it's emitted as long as it's emitted _somewhere_.
Why do you think creating an event handler on every acquire call is not good? Attaching an event handler shouldn't have any more performance impact than e.g. attaching a `then` handler to a promise, it's marginal (especially considering we are IO bound here anyway).

All writes to `this.poolError` in this file are inside promise chains, which are specced to always call back on the next tick even if the promise was resolved sync.
What guarantees you that any possible acquire error would have been set in the next event loop tick?

lib/dialects/abstract/connection-manager.js
felixfbecker: Fair enough, so I would expect good sinon-like libraries to override this one too 😄

lib/dialects/abstract/connection-manager.js
sushantdhiman: > What guarantees you that any possible acquire error would have been set in the next event loop tick?

If there is an error in pooling lib, then it would be emitted. But yeah there is no guarantee that order of error will be preserved. Because its all up to generic-pool to emit this error, which is not emitted per acquire call level, but rather per pool level.

To properly solve this issue we have to track all acquire calls in a queue and check which one failed to acquire connection and reject that. In other words we will need our own pooling, hence removing generic-pool from play

`this.poolErr` is not safe nor does `pool.on('factoryCreateError')`

felixfbecker: I think it is safe to use the first error that is emitted between an `acquire` call and the resolving of its promise, because if its emitted on the pool, then it must affect all acquire calls equally (so it's okay to reject them all with the same error, no need to map them). But it is not safe to only use an error that may have been emitted between the aquire call and the next event loop tick, and then ignore any errors until the promise is resolved.

sushantdhiman: `factoryCreateError` only appear for acquire calls so there is no issue that acquire call might failed by `this.poolErr` for any other reason.

But its not safe to assume that error occurring in next tick was due to last acquire call, if two acquire calls are pending then and error event on one of the call will be triggered on both of those acquire call, hence rejecting them. Which is not correct either.

Only safe solution will be to have acquire call reject properly rather than emitting pool level event `factortCreateError`

felixfbecker: Can we do a PR to generic-pool to do that? I already contributed before, maintainer was pretty responsive

sushantdhiman: Related issues

- https://github.com/coopernurse/node-pool/issues/183 
- https://github.com/coopernurse/node-pool/issues/175
- https://github.com/coopernurse/node-pool/issues/161

If you can solve this issue then it will be really great (for them too :) )

felixfbecker: k, let's leave it as-is for now.

PR #7700, opened 2017-05-27
Add relative and full path support for sqlite

### LINKED ISSUE #4721: sqlite URIs do not support absolute paths
I can use a URI like `sqlite:file.sqlite` to specify a file in the current directory called `file.sqlite`. But if I specify `sqlite:/home/ohnobinki/file.sqlite`, the file is not created.

Code that works:

```
#!/usr/bin/env node
var sequelize = new Sequelize('sqlite:file.sqlite');
sequelize.define('user', { email: Sequelize.STRING});
sequelize.sync();
```

Code that doesn’t work but should:

```
#!/usr/bin/env node
var sequelize = new Sequelize('sqlite:/home/ohnobinki/file.sqlite');
sequelize.define('user', { email: Sequelize.STRING});
sequelize.sync();
```

If I’m just forming the URIs incorrectly, then the correct syntax needs to be documented.

azeemba:
### Pull Request check-list

### Description of change

If the dialect is 'sqlite', then we parse host+pathname as a file path.
sqlite documentation suggests that this is the right way to parse them:
https://sqlite.org/c3ref/open.html#urifilenameexamples

If the "storage" option is provided, then that wil take precedence.

Examples:
sqlite:subfolder/dbname.db => subfolder/dbname.db
sqlite:/home/abs/dbname.db => /home/abs/dbname.db

Closes #4721

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -73,6 +73,7 @@
 - [ADDED] Support for JSON attributes in orders and groups. [#7564](https://github.com/sequelize/sequelize/issues/7564)
 - [REMOVED] Removes support for interpretation of raw properties and values in the where object. [#7568](https://github.com/sequelize/sequelize/issues/7568)
 - [FIXED] Upsert now updates all changed fields by default
+- [ADDED] Support for paths for sqlite databases via connection strings [#4721](https://github.com/sequelize/sequelize/issues/4721)
 
 ## BC breaks:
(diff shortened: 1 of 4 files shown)


azeemba:
I can update the docs and the changelog once there is more feedback

azeemba:
Looks like the build is failing with the following error: `Error: Please install sqlite3 package manually`
So perhaps I added the test in the generic section? Instead of sqlite dialect section?

PR #7701, opened 2017-05-28
Link directly to API docs

danawoodman:
Make it clearer there is both a "Guides" and "API docs". The small "Reference" link isn't super obvious to a newcomer (like myself). I find I always scan a readme for "API docs" and this PR will make that clearer and quicker for new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,14 @@
 
 Sequelize is a promise-based Node.js ORM for Postgres, MySQL, SQLite and Microsoft SQL Server. It features solid transaction support, relations, read replication and more.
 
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Rather than this we should improve docs layout so "Reference" is more visible somehow

danawoodman:
Personally I don't see the two as mutually exclusive. Is there harm in adding it?

sushantdhiman:
@danawoodman https://github.com/sequelize/sequelize/pull/7724 this pr covers your ideas :)

danawoodman:
It doesn't actually but I don't care strongly enough to push any more

PR #7713, opened 2017-06-01
add support for federated mysql engine

gorangajic:
Are you open to adding support for federated tables in MySQL? I am moving some legacy system to Node.js that is using federated tables so support for this would be great.

Thanks


sushantdhiman:
Federated tables aren't supported by other dialects, I would prefer to keep code simple and dont add features which are rarely used.

If there is enough demand from community then we can first decide API then implement it :)

PR #7743, opened 2017-06-08
Fix bug caused by circular dependencies between utils and DataTypes

niyoko:
### Pull Request check-list

### Description of change

Fix severe bug caused by circular dependencies between `lib/utils.js` and `lib/data-types.js`. It makes `require("./data-types")` on `utils.js` to become empty object and then throws error on line 311 (`TypeError: Right-hand side of 'instanceof' is not an object`)

### DIFF
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -800,7 +800,7 @@ for (const helper of Object.keys(helpers)) {
  * }
  * ```
  */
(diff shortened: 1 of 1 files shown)


felixfbecker:
Really don't understand the benefit of _not_ exporting initially. Please post an example how to reproduce the issue this fixes.

Issue #7751, opened 2017-06-08
calling `.close` on Sequelize db instance does not properly close the connection socket


yonjah:
on v4.0.0 Sequlize is not properly closing the socket and process will not exit cleanly -
demo code -
```javascript
const Sequelize = require('sequelize');
const db  = new Sequelize('test', 'root', null, {dialect: 'mysql'});

db.query('SHOW TABLES')
	.then(() => db.close());
```

running this code process will not exit running it with `wtfnode` shows socket is still open after calling close

nrschultz:
I saw this as well, I think it has to do with the fact that the generic-pool library was upgraded and the drain call went from taking a callback to returning a promise.  I messed around in my local environment and I think I got it working by changing the callback to a `.then`.  I'll file a PR and hopefully that fixes the problem

galki:
doesnt work on windows and i have to use `process.exit()` instead of `db.close()`

**error:** `Unhandled rejection Error: pool is draining and cannot accept work`

**sequelize version:** 4.33.4

PR #7756, opened 2017-06-08
ConnectionManager.close not closing database connection completely 

nrschultz:
### Pull Request check-list

### Description of change

closes https://github.com/sequelize/sequelize/issues/7751

As mentioned in https://github.com/sequelize/sequelize/issues/7751, db.close() was not doing what it was supposed to.  It looks like during the `generic-pool` library upgrade the pool.drain() implementation was changed from taking a callback to returning a promise (https://github.com/coopernurse/node-pool/blob/master/lib/Pool.js#L481), but the sequelize code was not updated.  Because of that, the function being passed as a callback was never being called.  Adding a `.then` with the same function seems to fix the problem.

I didn't go looking for other `generic-pool` callbacks that may have this same issue.

I'm also not quite sure of the best way to test this change, some insight would be appreciated.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
 # Next
 - [ADDED] `Model.increment`, to increment multiple rows at a time [#7394](https://github.com/sequelize/sequelize/pull/7394)
+- [FIXED] calling `.close` on Sequelize db instance does not properly close the connection socket [#7751](https://github.com/sequelize/sequelize/issues/7751)
 
 # 4.0.0 (final)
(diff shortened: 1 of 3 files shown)


felixfbecker:
Thank you! This prevented any CI jobs to pass that did a `sequelize.sync()` + `sequelize.close()`

nrschultz:
@felixfbecker glad I could help out, any idea when a new version of sequelize with this fix will be out on NPM?

nrschultz:
Nevermind, I answered my own question, you cut one 2 hours ago, thanks!

Issue #7758, opened 2017-06-09
Column names in removeColumnQuery() are not properly quoted.


petey:
## What you are doing?
```js
const client = new Sequelize(...);

// ...

client.sync({ alter: true });
```

This executes some queries:
```
CREATE TABLE IF NOT EXISTS `builds` (`id` INTEGER PRIMARY KEY AUTOINCREMENT, `sha` TEXT, `commit` TEXT, `createTime` TEXT));

CREATE TEMPORARY TABLE IF NOT EXISTS `builds_backup` (`id` INTEGER PRIMARY KEY, `sha` TEXT, `commit` TEXT, `createTime` TEXT);

INSERT INTO `builds_backup` SELECT id, sha, commit, createTime FROM `builds`;
```

Note that the attribute or column names are not quoted. Due to the column name "commit" being the same as a reserved word in SQL this results in `SequelizeDatabaseError: SQLITE_ERROR: near "commit": syntax error`

## What do you expect to happen?
The INSERT INTO `builds_backup` command uses quoted column names like all the other queries.

## What is actually happening?
`SequelizeDatabaseError: SQLITE_ERROR: near "commit": syntax error`

__Dialect:__ mysql / postgres / sqlite
__Sequelize version:__ 4.x

https://github.com/sequelize/sequelize/blob/c385994028e703fd589f266d55504d305b52a8a1/lib/dialects/sqlite/query-generator.js#L391-L397

PR #7759, opened 2017-06-09
fix(7758): quote attributes in alterConstraint and renameColumn queries

petey:
### Pull Request check-list

- [x] Does your issue contain a link to existing issue (Closes #[7758]) or a description of the issue you are solving?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? N/A

### Description of change

Closes issue #7758

### DIFF
--- a/lib/dialects/sqlite/query-generator.js
+++ b/lib/dialects/sqlite/query-generator.js
@@ -386,7 +386,7 @@ const QueryGenerator = {
 
     const quotedTableName = this.quoteTable(tableName);
     const quotedBackupTableName = this.quoteTable(backupTableName);
-    const attributeNames = Object.keys(attributes).join(', ');
+    const attributeNames = Object.keys(attributes).map(attr => this.quoteIdentifier(attr)).join(', ');
 
     return this.createTableQuery(backupTableName, attributes).replace('CREATE TABLE', 'CREATE TEMPORARY TABLE')
       + `INSERT INTO ${quotedBackupTableName} SELECT ${attributeNames} FROM ${quotedTableName};`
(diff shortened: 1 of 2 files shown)


petey:
Had a lot of trouble even figuring out where to add a test for this. Eventually made the change to integration tests.

stjohnjohnson:
@janmeier do you have any idea why these tests are failing? Any hints as to how we can fix it?

test/integration/query-interface.test.js
```diff
@@ -336,7 +336,8 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
     it('rename a simple column', function() {
       const self = this;
       const Users = self.sequelize.define('_Users', {
-        username: DataTypes.STRING
+        username: DataTypes.STRING,
+        commit: DataTypes.STRING
```
felixfbecker: what does the added attribute test? does this fail before your change?

petey: @felixfbecker This is testing that the column names are properly quoted. Since `commit` is a reserved word in SQL, any query that is run using that as a column name will fail, unless that column name is quoted.

```
SequelizeDatabaseError: SQLITE_ERROR: near "commit": syntax error
```

See the bug I filed for this issue: https://github.com/sequelize/sequelize/issues/7758

felixfbecker: Could you add a comment for that? It's not obvious and might get changed by accident.
Also, wouldn't this be better caught by an SQL generation unit test?

petey:
@janmeier @felixfbecker I've made the requested changes. Can you take another look?

PR #7772, opened 2017-06-11
Minor typos

altern8tif:
### Description of change

_Documentation change only_

Should reference products instead of projects in "Creating elements of a "HasMany" or "BelongsToMany" association" section.


janmeier:
:+1:

PR #7774, opened 2017-06-11
docs(Model): Correct Model.validate() documentation

### LINKED ISSUE #7769: instance.validate() has inconsistent return
## What you are doing?
After upgrading to 4.0:

```js
const err = await model.validate();
if (err) {
  // handle error
}
```

I expect the return type of `validate()` to be `Promise<undefined|Errors.ValidationError>` as the docs state and the branch to be taken in the case of an invalid model.

The branch is always taken because I received a `Promise<Model>` instead. Taking a quick look at `InstanceValidator`, this is happening because with `hooks: true` (the new default in 4.0), the result is the model instance in the case of success instead of `null`. Relevant code:

```js
validate() {
  return this.options.hooks ? this._validateAndRunHooks() : this._validate();
}
```

```js
_validate() {
  // ...
  return Promise.all(/* ... */).then(() => {
    if (this.errors.length) {
      throw new sequelizeError.ValidationError(null, this.errors);
    }
  });
}
```

```js
_validateAndRunHooks() {
  // ...
  return runHooks('beforeValidate', this.modelInstance, this.options)
    // ...
    .return(this.modelInstance);
}

```

dpatti:
### Pull Request check-list

### Description of change

In 4.0, `model.validate()` was changed to return a promise that resolves
only on success and rejects on error. Previously, it would return a
promise that resolved to either `undefined` on success or some
`ValidationError` on error. The upgrade guide did not mention this
breaking change, and the documentation did not reflect the change. A
test was added to show the intent of the documentation.

Closes #7769

### DIFF
--- a/docs/upgrade-to-v4.md
+++ b/docs/upgrade-to-v4.md
@@ -85,6 +85,7 @@ Sequelize V4 is a major release and it introduces new features and breaking chan
 
 - `DATEONLY` now returns string in `YYYY-MM-DD` format rather than `Date` type
 - `Model.validate` instance method now runs validation hooks by default. Previously you needed to pass `{ hooks: true }`. You can override this behavior by passing `{ hooks: false }`
+- The resulting promise from the `Model.validate` instance method will be rejected when validation fails. It will fulfill when validation succeeds.
 - Raw options for where, order and group like `where: { $raw: '..', order: [{ raw: '..' }], group: [{ raw: '..' }] }` have been removed to prevent SQL injection attacks.
 
 ### New features
--- a/lib/model.js
+++ b/lib/model.js
@@ -3722,14 +3722,14 @@ class Model {
   /**
   * Validate the attributes of this instance according to validation rules set in the model definition.
   *
-  * Emits null if and only if validation successful; otherwise an Error instance containing { field name : [error msgs] } entries.
+  * The promise fulfills if and only if validation successful; otherwise it rejects an Error instance containing { field name : [error msgs] } entries.
   *
   * @param {Object} [options] Options that are passed to the validator
   * @param {Array} [options.skip] An array of strings. All properties that are in this array will not be validated
   * @param {Array} [options.fields] An array of strings. Only the properties that are in this array will be validated
   * @param {Boolean} [options.hooks=true] Run before and after validate hooks
   *
-  * @return {Promise<undefined|Errors.ValidationError>}
+  * @return {Promise<undefined>}
   */
   validate(options) {
     return new InstanceValidator(this, options).validate();
--- a/test/unit/instance-validator.test.js
+++ b/test/unit/instance-validator.test.js
@@ -10,7 +10,18 @@ const SequelizeValidationError = require('../../lib/errors').ValidationError;
 
 describe(Support.getTestDialectTeaser('InstanceValidator'), () => {
   beforeEach(() => {
-    this.User = Support.sequelize.define('user');
+    this.User = Support.sequelize.define('user', {
+      fails: {
+        type: Support.Sequelize.BOOLEAN,
+        validate: {
+          isNotTrue(value) {
+            if (value) {
+              throw Error('Manual model validation failure');
+            }
+          }
+        }
+      }
+    });
   });
 
   it('configures itself to run hooks by default', () => {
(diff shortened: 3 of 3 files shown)


dpatti:
I can add this to the changelog, but it seems like mostly a documentation update.

lib/model.js
```diff
@@ -3722,14 +3722,14 @@ class Model {
   /**
   * Validate the attributes of this instance according to validation rules set in the model definition.
   *
-  * Emits null if and only if validation successful; otherwise an Error instance containing { field name : [error msgs] } entries.
+  * The promise resolves undefined if and only if validation successful; otherwise it rejects an Error instance containing { field name : [error msgs] } entries.
```
felixfbecker: I would just say _The promise fulfills if..._

test/unit/instance-validator.test.js
```diff
@@ -40,6 +51,20 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), () => {
       expect(_validate).to.have.been.calledOnce;
       expect(_validateAndRunHooks).to.not.have.been.called;
     });
+
+    it('resolves to undefined when validation is successful', () => {
```
felixfbecker: ~~resolves to undefined~~ fulfills when

lib/instance-validator.js
```diff
@@ -112,7 +112,7 @@ class InstanceValidator {
           .then(newError => { throw newError || error; }))
       )
       .then(() => runHooks('afterValidate', this.modelInstance, this.options))
-      .return(this.modelInstance);
+      .return();
```
felixfbecker: is this method's return value returned elsewhere? This must introduce a breaking change at a different end

lib/instance-validator.js
```diff
@@ -112,7 +112,7 @@ class InstanceValidator {
           .then(newError => { throw newError || error; }))
```
felixfbecker: If `validate` fulfills `void` or rejects, can this be removed?

docs/upgrade-to-v4.md
```diff
@@ -85,6 +85,7 @@ Sequelize V4 is a major release and it introduces new features and breaking chan
 
 - `DATEONLY` now returns string in `YYYY-MM-DD` format rather than `Date` type
 - `Model.validate` instance method now runs validation hooks by default. Previously you needed to pass `{ hooks: true }`. You can override this behavior by passing `{ hooks: false }`
+- The resulting promise from the `Model.validate` instance method will be rejected when validation fails. It will resolve undefined when validation succeeds.
```
felixfbecker: ~~resolve~~ fulfill

lib/instance-validator.js
dpatti: It depends on what you mean by "breaking change". Yes, if someone depended on this behavior, then a new version would break their implementation, but resolving into the model instance is undocumented behavior.

lib/instance-validator.js
dpatti: Though it's maybe not clear from the context, this `then()` call is applied to the result of `runHooks('validationFailed')`. I don't know under what conditions `runHooks` can resolve with a value that is an error, but it semantically did not seem relevant to this pull request.

dpatti:
I changed the language as you suggested. The test coverage report brought up an interesting issue: `validate()`'s return value (in the hooks case) is being used [here](https://github.com/sequelize/sequelize/blob/ccff5f690ec85e77a054b684ee59801ef801caad/lib/model.js#L2648-L2653), but clearly there is some attempt to work around the fact that it is sometimes `undefined`. I can't say that I understand what is going on in this method, so I'm not sure how to proceed.

lib/instance-validator.js
felixfbecker: what I mean is, yes, `validate` should never fulfill with the model, but you are not changing `validate` here, but a different method that may be called in other places. It would be great if you could verify the return value is not passed to the user in other places as well.

dpatti: `_validateAndRunHooks()` is private API and only used once within the class in the `validate()` method.

dpatti: And to expand further, `InstanceValidator` is only used in one place, and that's the model's `validate()` method.

felixfbecker:
> The test coverage report brought up an interesting issue: validate()'s return value (in the hooks case) is being used here, but clearly there is some attempt to work around the fact that it is sometimes undefined. I can't say that I understand what is going on in this method, so I'm not sure how to proceed.

me neither. @sushantdhiman maybe you know?

felixfbecker:
we could also merge this in with just the docs change.

dpatti:
@felixfbecker I pushed a new commit with just the documentation change and the tests for good measure. I think that's more important, anyway.

PR #7777, opened 2017-06-12
chore: Set up semantic-release

### LINKED ISSUE #6550: Semantic release
https://github.com/semantic-release/semantic-release
- Automatic changelog
- Automatic releases

Commit message convention can be followed easily with https://github.com/commitizen/cz-cli
and validated with https://github.com/kentcdodds/validate-commit-msg

felixfbecker:
Closes #6550

This sets up semantic-release to automatically release from CI.
It will automatically tag a version, push it to the repo, publish to npm and create a GitHub release with a changelog from conventional commit messages.

The convention is described in the [AngularJS commit message conventions](https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#heading=h.em2hiij8p46d), but they are really simple. Basically every commit is prefixed with something like `feat` or ` fix`. It is validated automatically with a husky precommit hook, and you can also use commitizen to commit interactively by answering questions. Breaking changes need to be documented in the message footer with _BREAKING CHANGE:_.

As maintainers, we need to make sure that when we Squash+Merge, the convention is still followed in the squashed commit - just as we right now make sure changes are properly documented in the changelog. A nice side effect is that we have full control over the changelog entries at the moment we merge.

Overall, I think this will make our work much easier, no more will there be questions like "Any plan when this will be released" or "Hey, could you cut a new release". It fits perfectly into the new model of smaller breaking changes (where needed) and less huge diverging legacy branches with countless backports.

I removed the changelog to not confuse anyone, which is then maintained automatically in GitHub releases from now on (we were already pretty good at maintaining them). As this is git, anyone can check out the old changelog if he desires.

I extended the contribution guide to reflect these new conventions, and also adapted the GitHub PR template to include this in the checklist (instead of the changelog item).

Recommended Watching:
- https://www.youtube.com/watch?v=tc2UgG5L7WM&index=6&list=PLFZ5NyC0xHDaaTy6tY9p0C0jd_rRRl5Zm
- https://egghead.io/lessons/javascript-how-to-write-a-javascript-library-automating-releases-with-semantic-release


.travis.yml
```diff
@@ -42,6 +42,13 @@ script:
   - npm run lint
   - "if [ $COVERAGE ]; then npm run cover && bash <(curl -s https://codecov.io/bash) -f coverage/lcov.info; else npm run test; fi"
 
+after_success:
+  - npm run semantic-release
```
mickhansen: Only for master right? Or does semantic-release check this?

felixfbecker: I think they check this. This was automatically setup by their script. Otherwise this PR would have been released :D

mickhansen: Right you are :)

harshithkashyap:
@felixfbecker 
> I removed the changelog to not confuse anyone, which is then maintained automatically in GitHub releases from now on

Does this mean the we won't have a changelog.md file at the root directory and the changes are listed in each release/tag?

felixfbecker:
Exactly. Because meaningful commit messages are enforced, the change log can always be retrieved through `git log`, `git blame` and GitHub releases. I like this because it means the contents of the repo are what is under version control, and the version history is managed by the version control system.

harshithkashyap:
Is it possible to mirror the changes into changelog.md file? Before upgrading to newer version, its easier to follow a single file to see the changes before upgrading. We'll have a go through many release/tags to see the changes. How about we append the changes listed in release to changelog.md file at the end? Not sure the process used by [Angular](github.com/angular/angular) to maintain changelog, but they keep a changelog file and update it on newer tags/releases.

felixfbecker:
The process is as follows:
- Merge a commit into master with conventional commit message
- In CI, semantic release generates changelog, pushes it to GitHub, tags new version and publishes to npm

To maintain a changelog, you would have to either let semantic-release push a new commit "Update changelog.md" _for every single commit_ (so your commit log will be 50% changelog update commits), or have the contributors generate the changelog before commiting. That doesn't work, because the changelog is generated _from_ the commit message, so you don't know what to put into the changelog until _after_ you commited. And it eliminates the benefit of not having to review changelog updates in PRs, because you would have to move the automation out of CI.

In my opinion, GitHub releases read much nicer than a changelog.md file: https://github.com/sequelize/sequelize/releases
They support the same markdown features, you still have full control.
And if you want to view the changelog locally without going to GitHub, use `git log` - it's contains the exact same information.

felixfbecker:
Angular doesn't use semantic release, so they don't release on every commit and don't have to update it on every commit. They update their changelog file on every release with a commit: https://github.com/angular/angular/commits/master/CHANGELOG.md

harshithkashyap:
I believe github shows only the last 10 releases on https://github.com/sequelize/sequelize/releases and you'd go about clicking next button again and again compared to reading a single markdown file. This wouldn't affect if a user is updating sequelize frequently as the latest releases are visible. But I guess its my personal preference. It makes sense from maintenance perspective to automatically publish releases.

Issue #7773, opened 2017-06-11
Fix flaky tests


felixfbecker:
On average, I have to restart 2 jobs on every build until all jobs are green. The errors take the form of `Expected 0 queries running, still 1 queries running`

papb:
I'm surprised this problem still persists for so long :(

This might be related to sequelize/sequelize-pool#12 (also see https://github.com/sequelize/sequelize/pull/11456#issuecomment-533409558)

PR #7779, opened 2017-06-12
Fixed missing captureStackTrace for errors. (closes #7770)

### LINKED ISSUE #7770: Unit tests failing on Node.js 7+
## What you are doing?
I tried running `npm run test-unit`.

I expected the unit tests to pass since I made no modifications.

I got the following error:
```
  1) errors should maintain stack trace:

      AssertionError: expected 'SequelizeBaseError: this is a message' to equal 'SequelizeValidationError: this is a message'
      + expected - actual

      -SequelizeBaseError: this is a message
      +SequelizeValidationError: this is a message

      at Assertion.assertEqual (node_modules/chai/lib/chai/core/assertions.js:487:12)
      at Assertion.ctx.(anonymous function) (node_modules/chai/lib/chai/utils/addMethod.js:41:25)
      at Assertion.<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:255:21)
      at Assertion.ctx.(anonymous function) [as equal] (node_modules/chai/lib/chai/utils/overwriteMethod.js:49:33)
      at Context.it (test/unit/errors.test.js:19:30)
```
This happens on both Node.js 7.10.0 and 8.1.0. 
On Node.js 6.11.0, all unit tests do pass.

__Sequelize version:__ 4.0.0 (not happening on 3.x.x)
__Node.js versions:__ 7.10.0 and 8.1.0

leroydev:
### Pull Request check-list

### Description of change

This fixes stacktraces of custom errors that subclass `BaseError` not having their error name in the stacktrace (issue #7770). I've also added unit tests to all the custom errors to check if the stacktrace contains the error name.

### DIFF
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -24,6 +24,7 @@ class SequelizeScopeError extends BaseError {
   constructor(parent) {
     super(parent);
     this.name = 'SequelizeScopeError';
+    Error.captureStackTrace(this, this.constructor);
   }
 }
 exports.SequelizeScopeError = SequelizeScopeError;
@@ -56,6 +57,7 @@ class ValidationError extends BaseError {
     } else if (this.errors.length > 0 && this.errors[0].message) {
       this.message = this.errors.map(err => err.type + ': ' + err.message).join(',\n');
     }
+    Error.captureStackTrace(this, this.constructor);
   }
 
   /**
@@ -100,6 +102,7 @@ class OptimisticLockError extends BaseError {
      * @type {object}
      */
     this.where = options.where;
+    Error.captureStackTrace(this, this.constructor);
   }
 }
 exports.OptimisticLockError = OptimisticLockError;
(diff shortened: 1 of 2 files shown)


leroydev:
CI is failing because of queries that are still running. Not sure if that's because of my changes or because of flaky tests as described in https://github.com/sequelize/sequelize/issues/7773.

leroydev:
Sorry about that.. Made a bit of a mess when getting upstream changes in my branch.. It's now rebased with the upstream changes. :)

leroydev:
@janmeier Does this need a changelog entry? Also, are there other actions required from me to get this ready to be merged?

janmeier:
Nope, we use semantic release now, so the changelog is generated automatically. 

Could you rebase it on top of latest master, then I will merge

leroydev:
@janmeier done :)

janmeier:
Thanks!

BridgeAR:
This is not really a good fix because `Error.captureStackTrace` is a really heavy operation that takes quite some CPU time. Instead, a easy fix would be to write:

```js
class SequelizeBaseError extends Error {
  get name() {
    return this.constructor.name
  }
}
```

You do not even have to set the message as that's done in the super constructor. The only issue without `Error.captureStackTrace` is that in Node.js < 7 the error stack will contain it's constructor as frame. That is fixed in newer Node.js versions though.

Issue #7419, opened 2017-03-23
Soft-deletes with new MSSQL DATETIMEOFFSET field


Wsiegenthaler:
Working around another date issue I've been having, I recently updated my `DATETIME2` fields to `DATETIMEOFFSET` as will be the practice with v4 going forward (see #5403, #7201). Changing the column type fixed my other issue but seems to break soft-deletes.

To make things interesting, my server is hosted on the west coast (-7h) and the database in Hawaii (-10h). Sequelize has been configured to expect the database in "Pacific/Honolulu".

Previously, when using `DATETIME2`, a record deleted at 10:30am PDT will have a deletedAt value of "07:30" in the database. This makes sense, since 7:30 in database time (Hawaii) *is* 10:30a PDT. When compared against `CURRENT_TIMESTAMP` this field will omit the record from subsequent selections.

`DATETIMEOFFSET` now stores the timezone in addition to the timestamp, so a record deleted at 10:30a PDT will now have a deletedAt value of "07:30 -10:00". This is expected, instead of inferring the timezone from the database it's specified in the field itself. The catch is that `CURRENT_TIMESTAMP` doesn't have an offset, and so is not compared properly with the `DATETIMEOFFSET` field, resulting in "deleted" records still being selected.

It seems that our deletedAt field now needs to be compared with MSSQL's `SYSDATETIMEOFFSET()` in order to account for the offset. I've hacked this into the paranoid clause and it all begins to work as expected - regardless of whether the field is `DATETIME2` or `DATETIMEOFFSET`. `SYSDATETIMEOFFSET()` is MSSQL specific and so can't simply be substituted for `CURRENT_TIMESTAMP`. My project is only using MSSQL so for now I'll just get by with the hack but please share any thoughts or opinions on the best way forward!

iamjoyce:
What version of sequelize are you using?

Wsiegenthaler:
This issue is regarding v4 which has transitioned away from DATETIME2 in favor of DATETIMEOFFSET for SQL Server databases (see #7201).

Wsiegenthaler:
I believe this is still an issue. I've been using [this](https://github.com/Wsiegenthaler/sequelize/commit/46f8b54fb3aae788f8ce8d388c18a4fe4b6131d1) as a workaround until an official solution is found.

tony-gutierrez:
Also there is no way to do a data export and import with this field type. It cannot import dates from strings.

PR #7787, opened 2017-06-13
MSSQL dialect timestamp conflict

getglad:
If dialect is MSSQL, use SYSDATETIMEOFFSET() instead of CURRENT_TIMESTAMP as the deletedAtDefaultValue

### Pull Request check-list

**Note**: rpm run test-mssql reports 751 passing on the first step. There is a second step that wants to use the connection variables, but I am using the `msnodesqlv8` + a connection string, and I'm not sure how to pass that into the tests.

### Description of change

When `sequelize` sets the `deletedAt` time stamp in paranoid mode, it submits a time stamp in the below format:
`... SET [deletedAt]=N'2017-06-13 11:35:41.000 +00:00'`

When `sequelize` then runs `findAll` against the records, it submits 
`... WHERE ([table].[deletedAt] > CURRENT_TIMESTAMP OR [table].[deletedAt] IS NULL);`

On MS SQL, `CURRENT_TIMESTAMP` returns:
`2017-06-13 11:35:41.000`

This results in returning all records.

While `SYSDATETIMEOFFSET()` returns:
`2017-06-13 11:40:11.8163142 -00:00`

This results in returning all the undeleted records.

It's a minor change, but the only use of `CURRENT_TIMESTAMP` I can find in the project.

Closes https://github.com/sequelize/sequelize/issues/7419


getglad:
@harshithkashyap I have submitted a test, but the `paranoid` flag on the model does not seem to alter the query output when I run it. Wouldn't mind getting an extra set of eyes on this.

In my application code I can have something like:
```
sequelize.define('User', {
  username: Sequelize.STRING,
}, {
  paranoid: true,
});
```
And my queries produce:
`SELECT [id], [username], [createdAt], [updatedAt], [deletedAt] FROM [Users] AS [User] WHERE ([User].[deletedAt] > SYSDATETIMEOFFSET() OR [User].[deletedAt] IS NULL);`

 Am I missing something in the test setup?

getglad:
@harshithkashyap just thought I'd bump you - see if you had any input

harshithkashyap:
@getglad 
1. We are not using mocha's TDD interface anymore. You'd need to use `describe` and `it` blocks for your tests.
2. You could add an unit test for [this](https://github.com/sequelize/sequelize/blob/master/lib/model.js#L51-L92) function and assert the return value instead of asserting `SQL` query. `SQL` query assertions are done for methods from `queryGenerator`.
3. Since you are modifying `model.js`, add an integration test similar to [this](https://github.com/sequelize/sequelize/blob/master/test/integration/model/find.test.js#L999-L1020). 

`SYSDATETIMEOFFSET()` would return the datetime with timezone. Our MSSQL CI Appveyor is likely configured to be in UTC timezone. For the test, you can initialize sequelize to use a different timezone. Sequelize would insert `deletedAt` values with the above specified timezone. You can assert the number of records returned. Since we have specified timezone with sequelize, `deletedAt` column would have timezone embedded such as `2012-10-25 12:32:10.1474881 +10:00` and `SYSDATETIMEOFFSET()` would return `2007-04-30 13:10:02.0474381 +00:00` UTC datetimeoffset. This would perform datetime comparison considering the timezone and would return correct results.

felixfbecker:
Will review once build is green.

lib/model.js
```diff
@@ -78,7 +78,7 @@ class Model {
     const deletedAtObject = {};
     let deletedAtDefaultValue = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
 
-    deletedAtDefaultValue = deletedAtDefaultValue || { $or: { $gt: model.sequelize.literal('CURRENT_TIMESTAMP'), $eq: null } };
+    deletedAtDefaultValue = deletedAtDefaultValue || { $or: { $gt: this.sequelize.options.dialect === 'mssql' ? model.sequelize.literal('SYSDATETIMEOFFSET()') : model.sequelize.literal('CURRENT_TIMESTAMP'), $eq: null } };
```
felixfbecker: I would wrap this line

test/integration/model/find.test.js
```diff
@@ -1024,18 +1024,33 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
-        ]);
-      }).then(() => {
-        return User.destroy({ where: {username: 'Tobi'} });
-      }).delay(5000).then(() => {
-        return User.findAll({ paranoid: true });
-      }).then(users => {
-        expect(users.length).to.be.eql(1);
-      });
+
+      if (dialect === 'sqlite')
```
felixfbecker: please put braces for multiline statements.
Why is there different behaviour for SQLite?

getglad:
@harshithkashyap Can you rerun #9233.1 and #9233.2? 
They passed on my test https://travis-ci.org/getglad/sequelize/builds/246298531

lib/model.js
getglad: @felixfbecker done

test/integration/model/find.test.js
getglad: Will do on braces.

On SQLite:
The only difference is a `.delay()` after deleting a record. There appears to be some sort of race condition when deleting the record (https://travis-ci.org/sequelize/sequelize/builds/245954861) and immediately querying it back, which is what I was doing in my test. Only saw it on SQLite, but if I created 2 records, soft delete 1, and then `findAll`, I would get both records back. A brief `delay` seems to fix on SQLite, but sometimes caused an `Error: Expected 0 running queries. 1 queries still running in [POSTGRES] Model constructor should allow the user to specify indexes in options`, especially on Postgres test - thus only tested it on SQLite. It may be worth extending the delay some to get a more reliable result, but 500 seemed okay for my test.

getglad: @felixfbecker  Done (https://github.com/sequelize/sequelize/pull/7787/commits/81590619c65b3cbf89ea2abe089e70a1a72859ea) and passed on my test (https://travis-ci.org/getglad/sequelize/builds/246334145).

felixfbecker: A timeout sounds very brittle, especially if you don't know why it appears and why it only appears only for SQLite. We have some flakiness in tests in general which are solved when restarting jobs, but if it happens consistently, it indicates some bug in the code...

getglad: I'm pretty sure it is just that the write (`destory` but with `paranoid` mode on) doesn't complete before the read (`findAll`) on SQLite. It isn't a timeout issue, just an opportunity for `SQLite` to fully register the change. All I meant to say was that lengthening the wait time reduces the brittleness, which I have done, so it should be strong from here on out. It may represent a problem with the `SQLite` syntax (honestly, I think it's just a virtue of SQLite itself), but this PR is to fix the syntax for `MSSQL` on reading over `paranoid` deleted rows. The SQLite fix in my test reliably resolves the issue for me over several tests.

The failure in the last travis ci build seems to be related to this test - https://github.com/sequelize/sequelize/blob/02ea09c731e117d258e523bd0c77d10fdb55f52f/test/integration/data-types.test.js#L45 - far as I can tell unrelated to my PR.

Other failures that I have seen, but I don't believe are related to this PR:
-
 https://github.com/sequelize/sequelize/blob/99d87dfff6c494994239588d0193a30a0093e7df/test/integration/model.test.js#L362
- https://github.com/sequelize/sequelize/blob/02ea09c731e117d258e523bd0c77d10fdb55f52f/test/integration/configuration.test.js#L18
- "no such file or directory, rename 'coverage/lcov.info' -> 'coverage/integration.info'"

Thoughts?

felixfbecker: All I'm saying is there is probably a better way to fix this than a timeout, a real application wouldn't do a timeout either. The fulfilled promise of the destroy query should mean that the query finished.
@sushantdhiman @janmeier

Issue #6423, opened 2016-08-12
model underscore should produce underscored table_name and  field_name's


alexey2baranov:
It would be great if sequelize will generate PostgreSQL DDL with lower_case_underscored table name and column names. By now _underscore_ option manage foreign keys and create_at, modifed_at fields only.

In short 

`CREATE TABLE IF NOT EXISTS "Male" ("CamelCase" VARCHAR(255), "created_at" TIMESTAMP WITH TIME ZONE NOT NULL, "updated_at" TIMESTAMP WITH TIME ZONE NOT NULL);`

should become

`CREATE TABLE IF NOT EXISTS male (camel_case VARCHAR(255), "created_at" TIMESTAMP WITH TIME ZONE NOT NULL, "updated_at" TIMESTAMP WITH TIME ZONE NOT NULL);`

sushantdhiman:
Sounds like a reasonable feature to me, @janmeier ?

janmeier:
I don't think this is needed - If users want underscored names, they can simply define them as such, or use the `field` option.

The `underscored` option is for controlling auto-generated column names, but when a user tells us about the column `fooBar`, we call it fooBar ;)

This can easily be solved with a `beforeDefine` hook

sushantdhiman:
Correct, but there is lots of work involved with that :laughing: , I would prefer one single option based control.

If anyone can share their experiences from other ORMs like `ActiveRecord` (Rails) or `Eloquent` (Laravel / PHP), that will be really helpful to make this decision.

felixfbecker:
I also prefer underscored columns and tables and camelCase attributes and PascalCase model names. I think the ORM should allow to map this easily.

maxsummers:
It took me a long time to figure out why `underscored` option didn't convert my `camelCase` column names to `under_scored` ones.

I would also prefer to declare my columns as `camelCased` ones to be automatically converted later to underscored ones in SQL.

elvisgs:
+1

quocvu:
I also agree w/ @alexey2baranov.  I know we use of the `field` option, but that makes the model more verbose.  Typically, when we name a field `fooBar`, it is to keep it consistent w/ the rest of the javascript code and not that we really want `fooBar` as a column name.

The current behavior is not desirable.  Who would want mixed column naming scheme in their tables?

At the minimum, the documentation should be fixed.  It doesn't say `underscored` only applies to auto-generated fields.

```
[...]

PR #7788, opened 2017-06-14
underscore attributes

mmahalwy:
Fixes https://github.com/sequelize/sequelize/issues/6423

I think this might be a POC for allowing all fields be underscored by simply adding `underscoreAttributes` boolean to model definition options. 

I may need to write tests for get/set, please do advise!

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -17,6 +17,7 @@ const Hooks = require('./hooks');
 const associationsMixin = require('./associations/mixin');
 const defaultsOptions = { raw: true };
 const assert = require('assert');
+const inflection = require('inflection');
 
 /**
  * A Model represents a table in the database. Instances of this class represent a database row.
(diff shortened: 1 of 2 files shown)


kingjerod:
Look forward to seeing this merged. Currently spent about an hour trying to figure out why "underscored: true" didn't work properly to find out the documentation is misleading. Specifying "field" for everything is just extra work.

mmahalwy:
@kingjerod I am hopeful this will get looked at eventually, although I don't think this is the solution. Frankly, I'd like to build my own ORM shaped after ActiveRecord with all ES6 features

sushantdhiman:
Rather than introducing yet another option, I would like to merge `underscore` and `underscoreAll`. Setting `underscore: true` should convert all fields / tableName to snake case, which can be overridden by field option / tableName option

This will be a breaking change

PR #7802, opened 2017-06-15
Update model definition docs for timestamps

danawoodman:
- Clearly explain the requirement to add `createdAt` and `updatedAt` to your migration
- Move timestamp specific info to it's own section so it is easier to parse and so the initial description of model definitions is clearer
- Fix some inconsistent whitespace/quotations in examples


docs/models-definition.md
```diff
@@ -72,6 +72,37 @@ const Foo = sequelize.define('foo', {
...
+        autoIncrement: true,
+      },
+
+      // Timestamps
+      createdAt: Sequelize.DATE,
+      updatedAt: Sequelize.DATE,
+    })
+  },
+  down(queryInterface, Sequelize) {
+    return queryInterface.dropAllTables()
```
felixfbecker: why drop _all_ tables?

danawoodman: This was taken from my initial DB migration. Could change to dropping created table

sushantdhiman:
I have addressed issues pointed out by @felixfbecker , merging ⏩

Issue #7792, opened 2017-06-14
V4 hooks is not working with callbacks


albertosouza:
Hello, 
Sequelize v4 not is working with callbacks in hooks like are documented and was working on last sequelize version.

In documentation:
```
afterCreate(instances, options, fn)
```
And if do something like:
```
Book.addHook('afterCreate', 'doSomething', (book, options, cb) => {
  console.log('callback is undefined>', cb);
});
```
The callback function is undefined.
This seens to be a bug and is important to keep this callback on hooks feature to prevend big breaking changes.

If this is a bug i can help with fix ...

albertosouza:
Code to show this problem: https://github.com/albertosouza/sequelize_v4_issue_7792

deimosOmegaChan:
Hi,
The callback parameter has been removed at hooks. You need to return a Promise object.
```
Book.addHook('afterCreate', 'doSomething', (book, options) => {
  return Promise.resolve();
  // or
  return Promise.reject();
});
```

albertosouza:
Then documentation update is needed

PR #7806, opened 2017-06-17
Remove hooks callback from docs ref #7792

albertosouza:
If callback on hooks dont will be implemented, then remove it from docs:
See https://github.com/sequelize/sequelize/issues/7792 for related issue


janmeier:
:+1:

PR #7807, opened 2017-06-17
fix bulkCreate (options.updateOnDuplicate) error

Acathur:
it’s happening when target data has one and only one row

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2364,7 +2364,9 @@ class Model {
         return this.QueryInterface.bulkInsert(this.getTableName(options), records, options, attributes).then(results => {
           if (Array.isArray(results)) {
             results.forEach((result, i) => {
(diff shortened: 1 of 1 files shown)


schneefux:
This is the same as #6730. Debugging my code, I see that `records` is of length 1, `instances` is of length 1, but `results` is `[ { id: 15895 }, { id: 15896 } ]` (MariaDB 10.2, `id` is an autoincrement primary key).

ljtijhuis:
Experiencing the same bug and yes, that is the root cause @schneefux . There's no way to know how many rows actually got updated and/or inserted when using `INSERT INTO ... ON DUPLICATE ...` so just expanding the id's like is done now here (https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/query.js#L106) does not work.

sushantdhiman:
Moved https://github.com/sequelize/sequelize/pull/8259

PR #7854, opened 2017-06-28
fix(indexes): Fix composite index issue

### LINKED ISSUE #7771: [Regression] Composite unique index adds single column unique index to the columns
## What you are doing?

```js
const sequelize = new Sequelize('db', 'user', 'pass', {
  host: 'localhost',
  port: 5432,
  dialect: 'postgres',
});

const Test = sequelize.define(
  'test', {
    unique1: Sequelize.INTEGER,
    // even unique; false doesn't override the effect
    unique2: {type: Sequelize.INTEGER, unique: false},
  }, {
    indexes: [{
      unique: true,
      fields: ['unique1', 'unique2']
    }]
  },
);

Test.sync({force: true})

```

Previously ( `4.0.0-2` ) used to define **only** composite unique index. Which is the expected behavior.

But now it defines 3 unique indices:

1. `UNIQUE (unique1)`
2. `UNIQUE (unique2)`
3. `UNIQUE (unique1, unique2)`

I think the culprit is #7196.

This is obviously a regression as my queries are now broken, because `unique1` and `unique2` should be unique in combination not in their own.

_Output, either JSON or SQL_

Constraint check errors... 😱 😱 

__Dialect:__ postgres (possibly any)
__Database version:__ 9.6
__Sequelize version:__ 4.0.0

alitaheri:
### Pull Request check-list

### Description of change

Fixes regression introduced by #7196 as described in: #7771

I've also inverted the unit tests to make sure no one else breaks this by mistake.

Closes #7771

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -322,7 +322,7 @@ const QueryGenerator = {
     return sql;
   },
 
-  upsertQuery(tableName, insertValues, updateValues, where, rawAttributes) {
+  upsertQuery(tableName, insertValues, updateValues, where, model) {
     const targetTableAlias = this.quoteTable(`${tableName}_target`);
     const sourceTableAlias = this.quoteTable(`${tableName}_source`);
     const primaryKeysAttrs = [];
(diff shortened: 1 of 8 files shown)


alitaheri:
The test fails aren't my fault :/

lib/dialects/mssql/query-generator.js
```diff
@@ -333,18 +333,32 @@ const QueryGenerator = {
...
+          const fieldName = typeof field === 'string' ? field : field.name || field.attribute;
+          if (uniqueAttrs.indexOf(fieldName) === -1 && model.rawAttributes[fieldName]) {
+            uniqueAttrs.push(fieldName);
+          }
+        }
+      }
+
+      return index;
+    });
+
```
harshithkashyap: @alitaheri Since you are not modifying the `model.options.indexes` array, its better to use `for..of`. Also, no need to return the index.

harshithkashyap:
@alitaheri  Also, if you could add a test to `sequelize.sync` with the issue you are trying to solve, it'd easy to catch in future.

lib/dialects/mssql/query-generator.js
alitaheri: @harshithkashyap Good catch 👍 Thanks. I used forEach simply because I see it used everywhere. I'll change to for..of it's a lot better anyway 👍

alitaheri:
@harshithkashyap I added 2 tests to model sync. one to make sure the indices won't affect individual attributes and one to make sure they throw when violated. should cover most cases.

alitaheri:
@harshithkashyap I've done what you asked, take a look please 😁 . The test failed because of the deadlock 😞

lib/dialects/mssql/query-generator.js
```diff
@@ -322,7 +322,7 @@ const QueryGenerator = {
     return sql;
   },
 
-  upsertQuery(tableName, insertValues, updateValues, where, rawAttributes) {
+  upsertQuery(tableName, insertValues, updateValues, where, model) {
```
felixfbecker: isn't this a breaking change?

alitaheri: Is it public API?

felixfbecker: Not strictly as it's undocumented. I know I've used `QueryGenerator.selectQuery()` sometimes to generate subqueries. But I guess that would be my own risk

alitaheri: @felixfbecker I've used some internals myself. I will make a PR to expose them properly later on. But I'm always ready for it to blow up in my face. Anything that isn't documented is subject to change without notice, the first rule of open source 😎 😎. we gotta draw the line somewhere right? 😅 😅 Or either the code gets complicated or no good change can make it to the end.

OfekA:
Thanks a lot for the fix, it took me a few hours to realize this is what caused it and finally upgraded.

PR #7855, opened 2017-06-28
feat(mysql/postgres): Add $regexp operator

ScottRocha:
### Pull Request check-list

### Description of change

Adds RLIKE functionality to MySQL dialect.

There was a PR merged in v1.7.0 code-base that added this functionality, but it has not been carried over to the current version. 

This adds the $rLike where condition for the mysql dialect which allows for full RegExp to be used.

I have not created an issue for this PR as it was a fairly simple issue to solve, but I can if it is needed.


ScottRocha:
Note: The travis-ci test failed due to an unrelated and possibly unreproducable query hang.

lib/dialects/abstract/query-generator.js
```diff
@@ -2005,6 +2006,8 @@ const QueryGenerator = {
       'nlike': '$notLike',
       '$notlike': '$notLike',
       'notilike': '$notILike',
+      'rlike': '$rLike',
```
felixfbecker: this object is for maintaining BC, but `rlike` is new.

felixfbecker:
PostgreSQL supports regexp too, what about other dialects?

ScottRocha:
I can look into adding support for other dialects. The RLIKE condition is MySQL specific as far as I can tell however.

lib/dialects/abstract/query-generator.js
ScottRocha: I'm sorry. New to the sequelize codebase, what does BC mean?

mkaufmaner: BC is a "Breaking Change".

ScottRocha: Ok, I'll remove that and update the PR.

ScottRocha:
Looks like PostgreSQL has the POSIX RegExp operators (~/~* and !~/!~*). I am not sure if it is currently possible to abstract the comparatorMap out to two different operators based on the dialect.

felixfbecker:
It should be, I think we do it for a lot of operators.

> The RLIKE condition is MySQL specific as far as I can tell however.

That's why I am hesitant to add an operator called `$rLike` if that name only makes sense in MySQL, and it really is just a regexp operator, then e.g. `$regexp` might make more sense. And it should probably support regexp literals in addition to strings

ScottRocha:
Noted. Looks like RLIKE is just an alias to REGEXP. Ill change that as well.

felixfbecker:
Great, that sounds much more fitting

docs/querying.md
```diff
@@ -135,6 +135,8 @@ $like: '%hat',         // LIKE '%hat'
 $notLike: '%hat'       // NOT LIKE '%hat'
 $iLike: '%hat'         // ILIKE '%hat' (case insensitive) (PG only)
 $notILike: '%hat'      // NOT ILIKE '%hat'  (PG only)
+$regexp: '^[h|a|t]'    // REGEXP '^[h|a|t]' (MySQL only)
+$notRegexp: '^[h|a|t]' // NOT REGEXP '^[h|a|t]' (MySQL only)
```
felixfbecker: shouldn't this work with `$not: { $regexp: ... }`?

test/unit/sql/where.test.js
```diff
@@ -908,6 +908,24 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
       });
     }
 
+    if (current.dialect.supports.REGEXP) {
+      suite('$regexp', () => {
+        testsql('username', {
+          $regexp: '^sw.*r$'
```
felixfbecker: Could you add a test for using a regexp literal? We might have to call `toString()` somewhere to make it work, but I think it would be slick because you wouldn't have to double-escape backslashes. I would also change the docs then

docs/querying.md
ScottRocha: I was just following the like trend of prepending not. I can remove that if needed. PostgreSQL also allows for CI and CS regexp comparison. Should I add that in (iRegexp, regexp)?

test/unit/sql/where.test.js
ScottRocha: I will look into this as well.

docs/querying.md
ScottRocha: I remember why I did regexp and notRegexp now. PostgreSQL has ~ for regexp and !~ for negation.

felixfbecker: how about this: `$regexp` always takes a true regexp literal, and then you can check `regexp.ignoreCase`? e.g. `/abc/i.ignoreCase === true`

felixfbecker: Just thought about this more, it probably won't work if JavaScript regexp syntax is different than the SQL regexp syntax. But worth looking into

ScottRocha: Go over my most recent commit and see if this is still something you would like done.

test/unit/sql/where.test.js
ScottRocha: Other than the addition of this test, is there anything else preventing this PR from being merged?

felixfbecker: not from my side

ScottRocha: It looks like the tests for PostgreSQL are failing before they can get to the regexp code. I believe this is unrelated to my changes.

lib/dialects/abstract/query-generator.js
```diff
@@ -2016,6 +2020,8 @@ const QueryGenerator = {
       '<@': '$contained'
     };
 
+    console.log(this._dialect.name);
```
felixfbecker: you need to remove this

ScottRocha: Yeh, I noticed that when the logs where spamming. It will be removed in my next commit.

test/unit/sql/where.test.js
ScottRocha: Ok, test with \n has passed (other than the unrelated test). Is there any other specific literal you want tested?

ScottRocha:
It looks like SQLite supports something very similar (if not identical) to MySQL's implementation of REGEXP. I can add that as well if needed.

test/unit/sql/where.test.js
```diff
@@ -908,6 +908,78 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
...
+        }, {
+          mysql: "`username` REGEXP '^sw.*r$'",
+          postgres: '"username" ~ \'^sw.*r$\''
+        });
+      });
+
+      suite('$regexp', () => {
```
felixfbecker: I don't understand why this suite is here twice, if the first test worked I don't see a possibility the second one wouldn't

test/unit/sql/where.test.js
felixfbecker: I had the idea to support JavaScript RegExp literals like `$regexp: /^[h|a|t]$/i`, but thinking more about it, JavaScript regexps are based on Perl, which is very different from Posix, so the set of escape sequences differs, which could cause unwanted behaviour.

test/unit/sql/where.test.js
ScottRocha: The second one wasn't working until I added a small snippet of code to skip escaping of '\n'.

test/unit/sql/where.test.js
ScottRocha: If you would like me to add a trust for that let me know.

test/unit/sql/where.test.js
felixfbecker: maybe because it needs to be double-escaped by the user? like `\\n`.
What about `\r`?

ScottRocha: All back-slashed literals should work.

ScottRocha: If you would like me to remove this functionality let me know.

ScottRocha: This is a blocking change for my organization, so I will do whatever it takes to get it merged as quickly as possible.

ScottRocha:
How many approvals are needed before this can be merged?

ScottRocha:
@harshithkashyap Could you point me to what file these tests are in?

ScottRocha:
@harshithkashyap I made the requested changes.

Issue #7433, opened 2017-03-27
Get used scope name in hooks?


oscarr-reyes:
I am in the need of getting the name of the scope used when i am executing functions on hooks, i would find this useful for processing values through scopes, if there is any function that could be executed on scope after find i would like to know, meanwhile if this could be a new feature i'm thinking this could be used this way:

```JavaScript
sequelize.define("accounts", {. . . .}, {
	tableName: "accounts",
	scopes: {
		bash: {. . . .}
	},
	hooks: {
		afterFind: afterFind
	}
});

function afterFind(rows, options){
	if(options.scopeName == "bash"){
		for(i = 0; rows.length > i; i++){
			var row = rows[i];

			// Transform the text to be only a short chunk for preview display
			if(row.biography.length > 60){
				row.text = row.biography.slice(0, 60);
			}
		}
	}

	return rows;
}
```

sushantdhiman:
Not much but you can get current scope structure with `this._scope` inside `afterFind`, you can do manual test against scopes defined in `this.options.scopes`

Nosthertus:
I tried that but `this._scope` is undefined in all scopes. I am using Sequelize `v3.28.0` and i don't know if any further update defines that property in the instance

sushantdhiman:
Its `this.$scope` for `v3` and `this._scope` for `v4`(master)

Nosthertus:
Oh.. looks like i need to check the differences in order to see if i need to update. But still i think getting the name of scope could be a good feature

sushantdhiman:
Yes, I agree. It should be easy to do.

Just need to set new prop like `this._scopeName` in `Model.scope` and unset in `Model.unscope` + tests

You may open a PR, I am keeping this issue open as a feature request.

oscarr-reyes:
sorry for the late response, i was working on upgrading sequelize 3 to 4 and i found out there i a change with the `scope` property and the where the scopes are stored now.. since i am now including nested models in a scope it is a bit harder to compare objects of the used scoped with the list of scopes, of course this takes a bit more time to be resolved so i would like to know if there is a PR already that makes this

I will find if the code that does this is already pulled anyways, if not then i will make my own PR now

oscarr-reyes:
PR #8753

BerkSpar:
How can I do it today in v6?

PR #7857, opened 2017-06-28
FEAT get used scope name in hooks

bdsabian:
### Pull Request check-list

### Description of change

This PR exposes _scopeName on _this_ object in hooks when using named scopes as described by @sushantdhiman in #7433. 

It's unclear to me if this is considered an API addition, so not sure if/where this should be documented but happy to do so if desired.  Just let me know where would be appropriate.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1373,6 +1373,8 @@ class Model {
       }
     }
 
+    self._scopeName = scopeName;
+
     return self;
   }
 
--- a/test/integration/model/scope.test.js
+++ b/test/integration/model/scope.test.js
@@ -48,5 +48,18 @@ describe(Support.getTestDialectTeaser('Model'), () => {
                 expect(record.access_level).to.exist;
               });
     });
+
+    it('should expose _scopeName in hooks', function () {
+      let scopeName;
+
+      this.ScopeMe.afterFind(function () {
+        scopeName = this._scopeName;
+      });
+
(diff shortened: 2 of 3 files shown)


bdsabian:
@felixfbecker This PR is in response to #7433 where the poster needed to know the name of the current scope in afterFind callback. He proposed a _scopeName property to provide that since the _scope object does not provide it.

To clarify, I stumbled on this looking to make a first OSS contribution to get a sense of the process.  I read the issue, it seemed fairly straightforward and @sushantdhiman seemed to agree it was a good idea and labelled it as Good First PR so here I am :).

Having had some time to step back now, however, and think this through a bit more, I realize my error here.  The Scope object doesn't have a name because it can't, it's the result of whatever scopes have been merged in up to that point so a single scope name doesn't make sense.  My change here naively/wrongly sets the _scopeName variable to the name of the last merged scope. 

What would make more sense is an array of the names of the currently applied scopes.  That would allow the poster of the issue the information about which scope(s) he's in.

But given that he has already solved his problem and no one else is asking for this, maybe this is not worth doing?

felixfbecker:
okay, closing this then. If I understand correctly then there is already scope data saved under `_scope` though. I also wonder whether just exposing the name makes sense when other info about the scope might be useful too. Also, I think the "options" used to query an instance are exposed, and those would contain the scope attributes. That would probably be a better way to inspect the instance because it doesn't treat "manual" queries different than scopes.

PR #7864, opened 2017-06-29
docs(search): use Algolia for search

felixfbecker:
This replaces the Google Custom Search box with search powered by Algolia.
The crawler is already set up.

image

It seems like searching the API reference doesn't work yet, but I assume that is just a misconfiguration of the crawler, see https://github.com/algolia/docsearch-configs/pull/172

Still, I find this search already much more useful than Google Custom Search.

Styles were just hacked together quick, but I think it looks quite nice.

### DIFF
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -1,5 +1,38 @@
 @import url(https://fonts.googleapis.com/css?family=Titillium+Web);
 
+.algolia-autocomplete .algolia-docsearch-suggestion--category-header,
+.algolia-autocomplete .algolia-docsearch-suggestion--title,
+.algolia-autocomplete .algolia-docsearch-suggestion--subcategory-column,
+#docs-search-input {
+  font-size: 16px !important;
+  line-height: normal;
+}
+
+.algolia-autocomplete .algolia-docsearch-suggestion--wrapper {
+  padding: 0 !important;
+}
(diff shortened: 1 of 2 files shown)


oscardianno:
Algolia Search was much more efficient, but seems like it's no longer present again.
I noticed it because I was looking at a course that apparently was released in 2018, where the instructor looks for information in the documentation very effortlessly, while now some of the queries just return no results.

Can this be done again?

papb:
Hi @oscardianno, it is working for me:

image

oscardianno:
Hi @papb, yes, but it doesn't seem to be the same search engine? I'm not sure when or what changed, but for example, try searching for _freezeTableName_. No results, however in the past it yielded this: 
imagen
Or try searching for a validation property, like _isEmail_ or _isAlphanumeric_. 

I know that kinda works and there's still ways to find that information, but it seemed to be more efficient and easy in the past, and I was just wondering if it's possible to get that back.

It's also discussed here: #12050

KerimG:
Searching for freezeTableName still yields nothing for me. neither in the manual nor in the API reference. Tried it with multiple browsers. Thought it might be a browser plugin issue and started them in incognito with all plugins disabled but it still doesn't work. 

This is hella broken.

Issue #7877, opened 2017-07-03
Sequelize 4.0: Custom instance method for all models


Joanne16:
Hi, how do we create custom instance methods to be applied/inherited by all models with sequelize 4.2? In sequelize 3 we used to have an "instanceMethods: {custom1: function (){} }" in the "define" section of a main model which was extended to all other models. How do we achieve the same behaviour with sequelize 4.0?

yonjah:
This has been discussed on #7804
I've created a pull request that will allow you to do that by passing a custom class to `options.Model`

Issue #6502, opened 2016-08-26
Support inheritance for models


felixfbecker:
This needs to be supported and tested:

``` js
class Base extends Model {}
Base.init({ whatever: DataTypes.STRING }, { tableName: 'base_table' })

class Derived extends Base {}
Derived.init({ second: DataTypes.INTEGER }, { tableName: 'derived_table' })

Derived.options.__proto__ === Base.options
Derived.attributes.__proto___ === Base.attributes
```

Derived inherits all attributes and options from Base.

`Model.init()` should do something like

``` js
this.options = Object.assign(Object.create(this.options), options)
```

yonjah:
This looks a bit cumbersome.
Also I'm not sure if you actually want all the `options` to be inherited.
Think about this case -
```js
class Derived2 extends Base {}
Derived2.init({ second: DataTypes.INTEGER }, { timestamps: false })
```

If you inherit options from Base you are actually still referring the table `base_table` where with other models `tableName` would be derived from the Model.name.

Another way to approach this is to add another static method that init will call to get options and attributes. since each Model implementation can extend this method it can have ful control on what will be inherited and what will be stayed in default. you are also not required to call `init` method on parents and so not locking them to specific connection and you can have more abstract models 

code example -
```js
class Base extends Model {
	static getOptions (options) {
		const defaultBaseOptions = { timestamp: false };
		return Utils.merge(super.getOptions(defaultBaseOptions), options);
	}
}
```

The base model will need some minor changes to the `init` mehtod to support this
```js
class Model {
	static init(attributes, options) { // testhint options:none		
		options = this.getOptions(options); //replacing lines 703-726
		...
		...
	}

	static getOptions (options) { // logic from init lines 703-726
		if (!options || !options.sequelize) {
			throw new Error('No Sequelize instance passed');
		}
		const {define: globalDefineOptions, omitNull} = this.sequelize.options;
		const modelName = options.modelName || this.name;
		return Utils.merge({ 
				modelName,
				name: {
					plural: Utils.pluralize(modelName),
					singular: Utils.singularize(modelName)
				},
				indexes: [],
				omitNull 
			},  _.cloneDeep(globalDefineOptions), options);
	}

}

```

felixfbecker:
Imo the examples you gave are more cumbersome, way more complex.

`tableName` is not a problem because the `tableName` is either set explicitly or inferred from the model name. So it will always override the parent value.

[...]

PR #7881, opened 2017-07-04
feat(sequelize.define): add `Model` option

### LINKED ISSUE #7804: How to define class methods in connection in Sequelize 4?
In Sequelize 3, I could say

```javascript
new Sequelize(DBInfo.db, DBInfo.username, DBInfo.password, {
    define: {
        classMethods: {
	    ClassLevelFunction() {
                // function
            }
	}
    }
});
```

and have `ClassLevelFunction` registered as a class method on all my models. Now this does not work? What is the new way of doing this?

yonjah:
### Pull Request check-list

- [x]  Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?

### Description of change
Added options.Model to define method to allow overriding the class used when creating a model
fix #7804 #7877 as it can be set globally when creating a new Sequelize instance

### DIFF
--- a/docs/upgrade-to-v4.md
+++ b/docs/upgrade-to-v4.md
@@ -36,18 +36,37 @@ Sequelize V4 is a major release and it introduces new features and breaking chan
   New:
 
   ```js
-  const Model = sequelize.define('Model', {
-      ...
-  });
+  class ExtendedModel extends Sequelize.Model {
+    // Class Method
+    static associate (models) {
+        ...associate the models
+    }
 
-  // Class Method
-  Model.associate = function (models) {
-      ...associate the models
-  };
-
-  // Instance Method
-  Model.prototype.someMethod = function () {..}
+    // Instance Method
+    someMethod () {..}
+  }
+  
+  const Model = sequelize.define('Model', { ... }, { Model: ExtendedModel});
   ```
+
+To replace the Model class for all defined models it is possible to set a global option on the created Sequelize instance
+
+  For Example the following `ClassicModel` class will bring back the `Model` property to every model instances:
+
+  ```js
+  class ClassicModel extends Sequelize.Model {
+    static init() {
+      this.prototype.Model = this; // Add Model property on instances
+      return super.init.apply(this, arguments);
+    }
+  };
+  const sequelize = new Sequelize('DB', 'USER', 'PASS', { define: { Model: ClassicModel }});
+  const Model = sequelize.define('Model', { ... });
+  const instance = new Model();
+
+  instance.Model === Model //true
+  ```
+
 - `Model.Instance` and `instance.Model` are removed. To access the Model from an instance, simply use [`instance.constructor`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor). The Instance class (`Model.Instance`) is now the Model itself.
 - Sequelize now uses an independent copy of bluebird library.
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -665,7 +665,7 @@ class Model {
    * @param {Function}                [attributes.column.get] Provide a custom getter for this column. Use `this.getDataValue(String)` to manipulate the underlying values.
    * @param {Function}                [attributes.column.set] Provide a custom setter for this column. Use `this.setDataValue(String, Value)` to manipulate the underlying values.
    * @param {Object}                  [attributes.validate] An object of validations to execute for this column every time the model is saved. Can be either the name of a validation provided by validator.js, a validation function provided by extending validator.js (see the `DAOValidator` property for more details), or a custom validation function. Custom validation functions are called with the value of the field, and can possibly take a second callback argument, to signal that they are asynchronous. If the validator is sync, it should throw in the case of a failed validation, it it is async, the callback should be called with the error text.
-   * @param {Object}                  [options] These options are merged with the default define options provided to the Sequelize constructor
+   * @param {Object}                  [options]
    * @param {Object}                  [options.defaultScope={}] Define the default search scope to use for this model. Scopes have the same form as the options passed to find / findAll
    * @param {Object}                  [options.scopes] More scopes, defined in the same way as defaultScope above. See `Model.scope` for more information about how scopes are defined, and what you can do with them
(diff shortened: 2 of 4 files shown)


janmeier:
I don't think we need a custom option for this. You can simply do `Model.function = ` and `Model.prototype.function =`.

If you want to use classes, use the `extends Sequelize.Model` syntax

yonjah:
> I don't think we need a custom option for this. You can simply do Model.function = and Model.prototype.function =.

I'm not sure I understand what it does.

> If you want to use classes, use the extends Sequelize.Model syntax

The issue with this is that you are forced to do it for every `Model` and then you cannot use define to create your models any more. 

So now you have to do something like this -
```js
class BaseModel extends Sequelize.Model {}
const sequelize = new Sequelize( ... );

class User extends BaseModel {}
User.init(attributes, { sequelize, modelName: 'user', ...options });

class Post extends BaseModel {}
Post.init(attributes, { sequelize, modelName: 'post', ...options });
```

And with the `Model` option you can do something like this :
```js
class BaseModel extends Sequelize.Model {}
const sequelize = new Sequelize( ... , { define: { Model: BaseModel } });

const User = sequelize.define('user', attributes, options);
const Post = sequelize.define('post', attributes, options);
```
 
I think the later seems a bit cleaner but maybe it doesn't matter much.

Another way to approach it will be overwriting Sequelize.Model and having define use that instead of the privately held Model reference. but I think it has more side effects and less flexibility than having an option.

felixfbecker:
This is essentially inheritance support, but only methods are inherited, no attributes and options, which is confusing. There are open issues for discussing inheritance support, please discuss the API there first.

yonjah:
@felixfbecker  are you referring to this issue #6502 ?
If not can you put a link to the correct issue where discussion is happening ?

From the docs (and code) it seems like class inheritance is the recommended way to get Model inheritance working. Is it not the case ? Is it still under discussion ? We should probably fix the docs than.

This feature shouldn't change or affect the underlying inheritance method. 
The only change here is to have a way to tell the `define` method which Model to use as the parent class instead of the default one. The actual inheritance logic shouldn't affect it.

felixfbecker:
Yes, #6502. Inheritance is not supported yet, in any way. As described on that issue, there is more to inheritance than just inheriting methods. One would expect to also inherit attributes and options, which is not the case in this PR.

yonjah:
> Yes, #6502. Inheritance is not supported yet

We should probably change the docs -> 3rd point of new features -
http://docs.sequelizejs.com/manual/tutorial/upgrade-to-v4.html#new-features

Not sure how many people are already relying on it after reading the documentation.

I can see how inheriting  options and attributes might be useful, but the current design decided to move to use ES6 classes and use native inheritance. Since Model was decided to be a Class and attributes and options are only added to the created object I fail to see how this can be achieved without breaking the existing API.

If any this pull request detaches the inheritance logic from the API (As you can see from the code example) but since current logic is already live in V4 unfortunately we wont be able to rely on it

felixfbecker:
I don't see where in the docs it is mentioned that inheritance from other models is supported. I described in #6502 how it can be achieved

yonjah:
Maybe I'm jumping the bullet here but this code example for the new way for defining models ->
```js
class User extends Sequelize.Model {}
User.init(attributes, options)
```
For me it means I can now use inheritance for defining my models. 
And I can't see why I couldn't later do  - 
```js
class Admins extends User {}
Admin.init(attributes, options)
```
But maybe other users who see this new feature example will not jump to the same conclusion

felixfbecker:
If I may complete your above example:
```js
class User extends Sequelize.Model {}
User.init({
  firstName: DataTypes.STRING,
  lastName: DataTypes.STRING
}, {
  sequelize,
  underscore: true,
  scopes: {
    ...
  },
  indexes: ...
})

class Admins extends User {}
Admin.init({
  firstName: DataTypes.STRING,
  lastName: DataTypes.STRING
}, {
  sequelize,
  underscore: true,
  scopes: {
    ...
  },
  indexes: ...
})
```
As you can see, you don't gain anything for inheriting, you have to duplicate everything again.
To me, that is unexpected behaviour. Think about the decorator use case:
```ts
@Options({ sequelize, scopes: ..., indexes: ... })
  @Attribute()
  firstName: string;
  @Attribute()
  lastName: string;
}

class Admin extends User {}
```
Wouldn't you expect `Admin` to have all the attributes of `User`?

yonjah:
Which version of Sequelize are you aiming this for ? 
Are decorators gonna be supported any time soon with node ?

Any way I guess this can work depending on how you implement the @Attribute decorator and how the `init` method actually derives the attributes later on.

felixfbecker:
This use case is supported in v4, it was one of the motivations to switch to classes. See https://github.com/felixfbecker/sequelize-decorators for a proof-of-concept.
With the solution I proposed in #6502 it would work.

yonjah:
I think making design decisions for current version based on some early stages spec draft that doesn't even have consistent syntax between different ESNext compilers is probably not the best idea.

Any way I can't understand why it shouldn't work. does the spec does not support inheritance for decorated properties ? or is it a limitation of TypeScript ?

felixfbecker:
Currently, `User` would be initialized with the `firstName` and `lastName` properties. But `Admin` would have no attributes in the DB at all (even though the TS compiler would think the Model has).

yonjah:
I'm not using this bleeding edge features that are not in the spec yet.
And I like JS enough not to give up on it and move to other languages like TypeScript
But I think if you'll check closely you'll see the reason decorators aren't working for you is probably because current Model classes are actually Row classes and not model classes. I can see how it would confuse TypeScript.
But if you actually aware of this issue you can probably hack your decorators to work with current implementation.

felixfbecker:
The issue is not TypeScript, decorators or "bleeding edge features". This is about a fundamental OOP concept, Liskov's substitution principle. If class A extends class B, and B has a set of attributes, then A must have them too.

yonjah:
I replied to this comment on the issue #6502
Again this pull request is only adding a minor API feature and has nothing to do with the actual inheritance implementation.

I see this is going nowhere so I'm closing this pull request. if any more issues like  #7804 #7877 rises I guess some one else can add it in

PR #7887, opened 2017-07-04
Hint for async

reinerBa:
I added a hint for the await keyword wich has increasing support. But maybe useless because everyone should know this possibility for promises.

### Pull Request check-list

### Description of change

I added a hint for the usage of async/await instead of `.then` to the documentation.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -136,8 +136,11 @@ _will never work!_ This is because `user` is a promise object, not a data row fr
 User.findOne().then(user => {
   console.log(user.get('firstName'));
 });
+```
 
-When your environment or transpiler supports [async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) this will work:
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -136,6 +136,11 @@ _will never work!_ This is because `user` is a promise object, not a data row fr
 User.findOne().then(user => {
   console.log(user.get('firstName'));
 });
+
+When your environment or transpiler supports [async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) this will work:
+user = await User.findOne()
```
eseliger: I would mention that `await` can only be used inside `async` functions. This hint is clearly for people who don't know `async/await` yet.

just
```ts
async function requestHandler(req: Request, res: Response) {
    const user = await User.findOne();
    res.send(user);
}
```
or so would be perfect.

felixfbecker: I wouldn't use TypeScript

PR #7907, opened 2017-07-08
Check for null value in setDataValue

Verdier:
### Pull Request check-list

### Description of change

This prevent key to be marked as changed if `value` and `originalValue` are both `null` when calling `setDataValue`.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3081,7 +3081,7 @@ class Model {
    */
   setDataValue(key, value) {
     const originalValue = this._previousDataValues[key];
-    if (!Utils.isPrimitive(value) || value !== originalValue) {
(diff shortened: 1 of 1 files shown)


eseliger:
Could you add a test for that? :)

felixfbecker:
I don't quite understand why the attribute should be marked dirty when assigned twice a primitive like a number but not `null`

Verdier:
@felixfbecker That's to be consistent with https://github.com/sequelize/sequelize/blob/master/lib/model.js#L3309. Today, if the current `dataValue` is `null` and you call `setDataValue` with `null`, the field will be marked as changed...

Verdier:
By the way @felixfbecker that's because null is not a Primitive, so you fall in the second equality check.

Issue #7917, opened 2017-07-10
Required includes for a findAll query are being overwritten in the subquery


joshuacullenlux:
## What you are doing?

I'm trying to do a findAndCountAll on a model which has multiple 1:m associations and is using an alias for each table.

```js
const Project = sequelize.define('project', {
  title: Sequelize.STRING,
  }, {
    tableName: 'project'
  })

const Task = sequelize.define('task', {
  title: Sequelize.STRING,
  fk_project_id: {
    type: Sequelize.INTEGER,
    references: {
      model: Project,
      key: 'id'
    }
  }}, {
    tableName: 'task'
  })

const Employee = sequelize.define('employee', {
  name: Sequelize.STRING,
  fk_project_id: {
    type: Sequelize.INTEGER,
    references: {
      model: Project,
      key: 'id'
    }
  }}, {
    tableName: 'employee'
  })

Project.hasMany(Task, {as: 'tasks', foreignKey: 'fk_project_id'})
Project.hasMany(Employee, {as: 'employees', foreignKey: 'fk_project_id'})
Task.belongsTo(Project, {as: 'project', foreignKey: 'fk_project_id'})
Employee.belongsTo(Project, {as: 'project', foreignKey: 'fk_project_id'})

const query = {
  limit: 5,
  offset: 0,
  include: [
    {
      model: Employee,
      as: 'employees',
      required: true,
    },
    {
      model: Task,
      as: 'tasks',
      required: true,
    },
  ]
}

Project.findAndCountAll(query).then(function(results) {
  console.log(results.count, results.rows.length)
})
```

## What do you expect to happen?
I wanted to retrieve 5 records that match the include criteria.  
I would expect the subquery to be limited by both tasks and employees not just tasks.

Output query:
```sql
SELECT "project".*, "employees"."id" AS "employees.id", "employees"."name" AS "employees.name", "employees"."fk_project_id" AS "employees.fk_project_id", "employees"."createdAt" AS "employees.createdAt", "employees"."updatedAt" AS "employees.updatedAt", "tasks"."id" AS "tasks.id", "tasks"."title" AS "tasks.title", "tasks"."fk_project_id" AS "tasks.fk_project_id", "tasks"."createdAt" AS "tasks.createdAt", "tasks"."updatedAt" AS "tasks.updatedAt" 
FROM (
    SELECT "project"."id", "project"."title", "project"."createdAt", "project"."updatedAt" 
    FROM "project" AS "project" 
    WHERE ( 
[...]

PR #7918, opened 2017-07-10
fix(abstract/querygenerator): Bugfix for includeAs subquery clause

joshuacullenlux:
### Pull Request check-list

A fix for issue [7917](https://github.com/sequelize/sequelize/issues/7917) here.

When generating an include the includeAs paramater can be an object. When the include was being added to a subquery it was using the object as the key. Any subsequent includes would overwrite the previous as they would have the same key (object Object).
 
It will now check if includeAs is an object and use the externalAs property as the key.


joshuacullenlux:
3 tests are failing

```json
1436 passing (17m)
  2 pending
  3 failing

  1) [MYSQL] DataTypes should parse an empty GEOMETRY field:
     SequelizeDatabaseError: Invalid GIS data provided to function st_geometryfromtext.
      at Query.formatError (lib/dialects/mysql/query.js:222:16)
      at Query.connection.query [as onResult] (lib/dialects/mysql/query.js:55:23)
      at Query.Command.execute (node_modules/mysql2/lib/commands/command.js:30:12)
      at Connection.handlePacket (node_modules/mysql2/lib/connection.js:500:28)
      at PacketParser.onPacket (node_modules/mysql2/lib/connection.js:94:16)
      at PacketParser.executeStart (node_modules/mysql2/lib/packet_parser.js:77:14)
      at Socket.<anonymous> (node_modules/mysql2/lib/connection.js:102:29)
      at addChunk (_stream_readable.js:252:12)
      at readableAddChunk (_stream_readable.js:239:11)
      at Socket.Readable.push (_stream_readable.js:197:10)
      at TCP.onread (net.js:588:20)
  From previous event:
      at Promise._resolveFromExecutor (node_modules/cls-bluebird/lib/shimOnCancel.js:41:40)
      at Query.run (lib/dialects/mysql/query.js:44:12)
      at retry (lib/sequelize.js:529:32)
      at node_modules/retry-as-promised/index.js:40:21
      at retryAsPromised (node_modules/retry-as-promised/index.js:30:10)
      at Promise.try.then.connection (lib/sequelize.js:529:14)
  From previous event:
      at Promise.then (node_modules/cls-bluebird/lib/shimMethod.js:38:20)
      at Context.it (test/integration/data-types.test.js:301:10)

  2) [MYSQL] Instance default values current date should store the current date in touchedAt:

      AssertionError: expected -3595000 to equal 5000
      + expected - actual

      --3595000
      +5000
      
      at Assertion.assertEqual (node_modules/chai/lib/chai/core/assertions.js:487:12)
      at Assertion.ctx.(anonymous function) (node_modules/chai/lib/chai/utils/addMethod.js:41:25)
      at Assertion.<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:255:21)
      at Assertion.ctx.(anonymous function) [as equal] (node_modules/chai/lib/chai/utils/overwriteMethod.js:49:33)
      at Context.<anonymous> (test/integration/instance.test.js:735:39)

  3) [MYSQL] Model sql injection attacks should properly escape the single quotes in coordinates:
     SequelizeDatabaseError: Invalid GIS data provided to function st_geometryfromtext.
      at Query.formatError (lib/dialects/mysql/query.js:222:16)
      at Query.connection.query [as onResult] (lib/dialects/mysql/query.js:55:23)
      at Query.Command.execute (node_modules/mysql2/lib/commands/command.js:30:12)
      at Connection.handlePacket (node_modules/mysql2/lib/connection.js:500:28)
      at PacketParser.onPacket (node_modules/mysql2/lib/connection.js:94:16)
      at PacketParser.executeStart (node_modules/mysql2/lib/packet_parser.js:77:14)
      at Socket.<anonymous> (node_modules/mysql2/lib/connection.js:102:29)
      at addChunk (_stream_readable.js:252:12)
      at readableAddChunk (_stream_readable.js:239:11)
      at Socket.Readable.push (_stream_readable.js:197:10)
      at TCP.onread (net.js:588:20)
  From previous event:
      at Promise._resolveFromExecutor (node_modules/cls-bluebird/lib/shimOnCancel.js:41:40)
      at Query.run (lib/dialects/mysql/query.js:44:12)
      at retry (lib/sequelize.js:529:32)
      at node_modules/retry-as-promised/index.js:40:21
      at retryAsPromised (node_modules/retry-as-promised/index.js:30:10)
      at Promise.try.then.connection (lib/sequelize.js:529:14)
  From previous event:
      at Promise.then (node_modules/cls-bluebird/lib/shimMethod.js:38:20)
```

lib/dialects/abstract/query-generator.js
```diff
@@ -1342,7 +1342,9 @@ const QueryGenerator = {
         ].join(' '));
 
         if (_.isPlainObject(topLevelInfo.options.where)) {
-          topLevelInfo.options.where['__' + includeAs] = subQueryWhere;
+          //includeAs can be a string or an object
+          const key = typeof includeAs === 'object' && includeAs.externalAs ? includeAs.externalAs : includeAs;
```
felixfbecker: can `includeAs` be `null`?
instead of `:`, just use `||`, it's shorter

joshuacullenlux:
I've made the change to use || and allowed for null values.

lib/dialects/abstract/query-generator.js
```diff
@@ -1342,7 +1342,9 @@ const QueryGenerator = {
         ].join(' '));
 
         if (_.isPlainObject(topLevelInfo.options.where)) {
-          topLevelInfo.options.where['__' + includeAs] = subQueryWhere;
+          //includeAs can be a string or an object
+          const key = includeAs && includeAs.externalAs || includeAs;
```
felixfbecker: if it's a non-empty string, this will evaluate to `includeAs.externalAs`, which is `undefined`

joshuacullenlux:
Yeah brain fail. I'll switch it back to using the ternary statement so it will account for null.

I'm not sure how to write it shorter with || without ignoring the fact that includeAs could be null.

joshuacullenlux:
@yonjah

Cheers for the extra info. 

Will do, I've looked into and it seems like it will always will be defined so I'm probably being overly cautious. I'll have a read but that makes sense. Awesome, I'll get some tests in. That will help my understanding as well.

joshuacullenlux:
@yonjah @felixfbecker 

I've added a test based on your feedback. Thanks for pointing me in the right direction.

It doesn't seem possible for includeAs to be null or any other falsy value. The generateInclude function is called in the following locations:

https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L994
https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L1368

In both cases include is being pulled form the options object and passed in. If null is added as an include option the process blows up much earlier at:
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L199

I'm confident it will never be null but you guys would have a better idea.

yonjah:
@joshuacullenlux test looks great.
Small thing I would suggest is instead using `fk_project` key use just `fk` there is an eslint rule to warn if keys are not camel cased and at least with my setup it's a bit annoying to see them (though this file has anyway other code that doesn't follow this rule)

I also wonder if your test can replace the test following it. the following test is a lot less clear and it's a bit hard to understand the full thought behind it (I can't seem to understand how it might change anything to add rows and then delete them) but I think with small or almost no change at all your test cover what it checks as well.
Any way it's a bit off scope for this pull request and probably need someone else to think about it so I'll leave it as is for now unless @felixfbecker  thinks it's a good idea to do it now.

Though your fix works with no issues I think your missing one small thing in your analysis -
See line https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L1246
`includeAs` is local scoped constant and will never change (it will always be an object) it is true that since it is not frozen theoretically someone could do `delete includeAs.internalAs` but that isn't something I'll consider protecting against, and you can see all over this function that `includeAs.internalAs` is used without any checks.
(Though you'll might also notice that almost all the functions variables are assume to exist and have the correct properties without any checks. I'm not sure if this is intentional or not but with the case of local scope there's certainly no reason to be over protective  )

joshuacullenlux:
@yonjah 
I've updated the fk in the test as suggested.

Yeah you're right. That constant should make this code safe. I was being overly conservative.

With regards to replacing the test below it seems that it is has a different goal than mine. My test is ensuring that multiple required models will be taken into account within the subquery when a limit is used. The other seems to be concerned with SomeConnection being associated with multiple models. I agree with you that it's probably out of scope for this PR but I'll leave that up to you guys to decide.

Issue #7882, opened 2017-07-04
pool connection never destroy after idle milliseconds


fourcels:
## What you are doing?

```js
const Sequelize = require('sequelize');
const sequelize = new Sequelize($mysqlUri, {
  pool: {
    max: 5,
    min: 0,
    idle: 10000
  }
})

sequelize
  .authenticate()
  .then(() => {
    console.log('Connection has been established successfully.');
  })
  .catch(err => {
    console.error('Unable to connect to the database:', err);
  });
```

## What do you expect to happen?
pool connection  should be destroyed after idle(10000 milliseconds) 

## What is actually happening?
pool connection  hadn't destroyed after idle(10000 milliseconds) 

__Dialect:__ mysql 
__Database version:
__Sequelize version:4.2.1

## ps
I had found a option __evictionRunIntervalMillis__ in [generic-pool](https://github.com/coopernurse/node-pool)
__evictionRunIntervalMillis__: How often to run eviction checks. Default: 0 (does not run).
sequelize options need add this option to control idle connection destroy

RayneHwang:
Plz check if my [pullrequest](https://github.com/sequelize/sequelize/pull/7922) fixed you problem

swordfish444:
@RayneHwang Please see the comments on your PR. Once you've resolved the formatting error, lets get this tested and scheduled for release.

sushantdhiman:
#7924

PR #7922, opened 2017-07-10
fix abstract/connection-manager.js pool configuration problem

RayneHwang:
### Pull Request check-list

- [ ] Have you added new tests to prevent regressions? no test needed
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? *no changes to public APIs*

### Description of change

As described in issue [#7882](https://github.com/sequelize/sequelize/issues/7882) , current [pool confiuration](https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/connection-manager.js#L112) in `/lib/dialects/abstract/connection-manager.js` initPool method sets `autostart` to false when no replication is configured, which causes generic-pool `_scheduleEvictorRun` check routine [never gets invoked](https://github.com/coopernurse/node-pool/blob/master/lib/Pool.js#L119). As a result idle connections never get evicted even if the database server's already closed the connection.
To fix that problem, simply set `autostart` option to true and the evict check routine gets scheduled.
As documented in node-pool, the `softIdleTimeoutMillis`(default to -1 ) is checked before `idleTimeoutMillis`, so if `softIdleTimeoutMillis` is not set, the connection will be considered stale no matter what `idleTimeoutMillis` is

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -152,6 +152,10 @@ Sequelize uses a pool to manage connections to your replicas. The default option
   max: 10,
   min: 0,
   idle: 1000
+  idle: 10000,
+  acquire: 10000,
+  evict: 60000,
+  handleDisconnects: true		  
 }
 ```
 
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -13,6 +13,8 @@ const defaultPoolingConfig = {
   min: 0,
   idle: 10000,
   acquire: 10000,
+  autostart:true,
+  evict: 60000,
   handleDisconnects: true
 };
 
@@ -112,6 +114,7 @@ class ConnectionManager {
         autostart: true,
         acquireTimeoutMillis: config.pool.acquire,
         idleTimeoutMillis: config.pool.idle,
+        evictionRunIntervalMillis: config.pool.evict,
         softIdleTimeoutMillis: config.pool.idle
       });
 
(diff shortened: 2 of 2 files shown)


swordfish444:
@RayneHwang That was fast!

RayneHwang:
@swordfish444 😄

swordfish444:
@RayneHwang I've opened https://github.com/sequelize/sequelize/pull/7924 to resolve for stale connections by ensuring evictions are ran.

RayneHwang:
@swordfish444 I've tested your [PR](https://github.com/sequelize/sequelize/pull/7924) in my project, doesn't seem to successfully evict stale connections.

- `autostart` must be `true`
The default value of `autostart` is `true`, but in asbtract/connection-manager.js `autostart` is manually set to `false`.
In generic-pool [constructor](https://github.com/coopernurse/node-pool/blob/master/lib/Pool.js#L119) ,if `autostart` is not set to `true`, the [evict routine](https://github.com/coopernurse/node-pool/blob/master/lib/Pool.js#L372) won't be called for the first time, and the following check relies on `setTimeout` inside the [_scheduleEvictorRun](https://github.com/coopernurse/node-pool/blob/master/lib/Pool.js#L352) ,so no evict will be scheduled. 

- `softIdleTimeoutMillis` must be set
Though node-pool doc claims that `idleTimeoutMillis` supercedes `softIdleTimeoutMillis`, but in  the [default evictor implementation](https://github.com/coopernurse/node-pool/blob/master/lib/DefaultEvictor.js#L7) `softIdleTimeoutMillis` is checked before `idleTimeoutMillis`. If `softIdleTimeoutMillis` is not set, the default value( -1 ) always makes the connection stale, when `availableObjectsCount` is larger than `config.min`

```javascript
if (config.softIdleTimeoutMillis < idleTime && config.min < availableObjectsCount) {
     return true
}
if (config.idleTimeoutMillis < idleTime) {
    return true;
}
```
So every time `_scheduleEvictorRun` gets ran, the connection will be evicted, no matter what `idleTimeoutMillis` is. It's really a confusing option. 

- `evictionRunIntervalMillis`
Sorry I forgot to commit that option. I will push some updates.

sushantdhiman:
`autostart: false` is not acceptable, this will be fixed by https://github.com/coopernurse/node-pool/pull/201

As a workaround you can call `start` from pooling library to initialize evictor https://github.com/sequelize/sequelize/pull/7924#issuecomment-325378123

DiegoRBaquero:
The `softIdleTimeoutMillis` change was somehow reverted, and thus set to -1, which thens marks all connections as idle.

Issue #7941, opened 2017-07-12
findOrCreate fails concurrently when unique attribute contains comma in value. Postgres unique constraint error incorrectly parses error message 23505 when the value contains a ','. 


skn3:
## What you are doing?
Trying to concurrently use findOrCreate when an attribute set to unique, would contain a comma.

```javascript
'use strict';

const Sequelize = require('sequelize');

//first connect to the database
const database = new Sequelize('database', 'user', 'password', {
    dialect: 'postgres',
    host: '127.0.0.1',
    port: 5432,

    pool: {
        max: 5,
        min: 0,
        idle: 10000
    },
});

return database.authenticate()
.then(() => {
    //success connecting,
    console.log('database connected');
    return database;
})
.then(database => {
    const containerModel = database.define('TestContainer', {
        name: {
            type: Sequelize.DataTypes.STRING(32),
            allowNull: false,
            unique: 'name',
        },
    });

    return database.sync({
        alter: true,
        force: false,
    })
    .then(database => {
        return database.transaction({
            autoCommit: false,
        })
        .then(transaction => {
            return Promise.all(['hello, world', 'hello, world'].map(name => {
                console.log(`findOrCreate({name: '${name}'})`);
                return containerModel.findOrCreate({
                    where: {
                        name: name,
                    },
                    defaults: {
                        name: name,
                    },
                    transaction: transaction,
                    returning: true,
                });
            }))
            .then(() => {
                transaction.rollback();
            })
            .catch(() => {
                transaction.rollback();
            });
        });
    });
})
.catch(err => {
    //failed to connect
    console.error(err);
    process.exit(1);
});
```

## What do you expect to happen?
concurrent operations to succeed without fail!

## What is actually happening?
So lets say i set attribute of my Model attribute 'name' to unique. If i were to try and findOrCreate multiple of these models concurrently, and there is a unique validation error, then it fails!

It fails because the findOrCreate checks the fields in the err object generated by postgres/query.js. The parsing of said fields is done from the error message generated by postgres. In the postgres/query.js it is splitting the regex matched values by ',' for both the field name and value. This incorrectly truncates the field value if it contained a ','. When the findOrCreate attempts to check for concurrency, it fails because the reported value from the error and the value its expecting do not match!

Get an error of 

> TestContainer#findOrCreate: value used for name was not equal for both the find and the create calls, 'hello, world' vs 'hello'

__Dialect:__ postgres 
__Database version:__ 9.6
__Sequelize version:__ master

[...]

PR #7942, opened 2017-07-12
fix(findOrCreate): concurrent postgres unique constraint error parsing

skn3:
### Description of change

fixed an issue described here: #7941

### DIFF
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -303,7 +303,7 @@ class Query extends AbstractQuery {
         // there are multiple different formats of error messages for this error code
         // this regex should check at least two
         if (errDetail && (match = errDetail.replace(/"/g, '').match(/Key \((.*?)\)=\((.*?)\)/))) {
-          fields = _.zipObject(match[1].split(', '), match[2].split(', '));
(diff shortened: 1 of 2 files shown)


lib/dialects/postgres/query.js
```diff
@@ -303,7 +303,7 @@ class Query extends AbstractQuery {
         // there are multiple different formats of error messages for this error code
         // this regex should check at least two
         if (errDetail && (match = errDetail.replace(/"/g, '').match(/Key \((.*?)\)=\((.*?)\)/))) {
-          fields = _.zipObject(match[1].split(', '), match[2].split(', '));
+          fields = _.zipObject(match[1], match[2]);
```
felixfbecker: there must be a reason why the `split` calls were there

skn3: perhaps an old error style or copy and paste error?
I don't know enough about sequelize internals to know what to suggest?

skn3: how would someone go about find out where the need for this came from? Does sequelize have some internal postgres version variable? Do you accept version conditional code? This seems kind of an important issue for generating unique keys!

PR #7955, opened 2017-07-14
this.underscored should be  this.options.underscored

sennmac:
hi~ boss.
I think maybe this.underscored should be  this.options.underscored.
this.underscored will never be used.


sushantdhiman:
Nope, this code is correct. Its a flag indicating if Model is underscored or not

PR #7960, opened 2017-07-16
fix(queryinterface): decrement regression

### LINKED ISSUE #7810: Sequelize v4 does not update the instance when using .increment or .decrement
## What you are doing?

Hello, upgrading from v3 to v4, I had the following issue:

```js
  id: {
    type: DataTypes.UUID,
    primaryKey: true,
    defaultValue: DataTypes.UUIDV4
  },
  username: {
    type: DataTypes.STRING,
    allowNull: false
  },
  balance: {
    type: DataTypes.INTEGER,
    defaultValue: 0
  }
});

describe('.increment', ()=> {
  it('increments the property in the instance', () => {
    return User.create({username: 'alice'})
      .then(user => {
        user.increment('balance', 5)
        .then(()=> {
          expect(user.balance).toBe(5);
        });
      });
  });
});
```

In Sequelize v3, the test would pass. the balance would be incremented in the object, without the need to reload it from the DB. After the upgrade, I expected the same behaviour as demonstrated by the test.

With Sequelize v4, the record is updated in DB but the instance itself is not.
This is not documented in the upgrade notes.

__Sequelize version:__ v4.1.0

hieven:
### Pull Request check-list

### Description of change

Basically just follow what query interface `increment` does

Closes #7810

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -685,13 +685,14 @@ class QueryInterface {
     return this.sequelize.query(sql, options);
   }
 
-  decrement(instance, tableName, values, identifier, options) {
+  decrement(model, tableName, values, identifier, options) {
+    options = Utils.cloneDeep(options);
+
     const sql = this.QueryGenerator.arithmeticQuery('-', tableName, values, identifier, options, options.attributes);
 
-    options = _.clone(options) || {};
-
     options.type = QueryTypes.UPDATE;
-    options.instance = instance;
+    options.model = model;
+
     return this.sequelize.query(sql, options);
   }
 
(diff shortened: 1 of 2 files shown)


felixfbecker:
Is this a breaking change?

hieven:
Hi @felixfbecker,
No, there is no breaking change between my initial PR and this fix PR.

In my initial PR, I just moved the original code in instance `increment` method to be class `increment` method and let instance method to call the class method.
(More details, please check #7394)

What I missed is that I didn't change the way how `decrement` in `query-interface` works, so it becomes a bug.

When someone uses instance method `decrement` under any dialect supports `returning`, the original instance won't get updated.

I also add another commit for testing this circumstance. Please help review it.

Thanks!

PR #7970, opened 2017-07-17
fix #7968 - mssql bind support - add support for array in bind parame…

### LINKED ISSUE #7968: Array in bind option
I'm trying to send array in the `bind` option and getting `Conversion failed when converting the nvarchar value '1,2,3,4' to data type int` error.

Back in the days i used directly `tedious` and [i opened an issue](https://github.com/tediousjs/tedious/issues/332) about that, but @asztal said that "This is a limitation of SQL server, not Tedious".

I think that the best solution is to loop over the values and add `paramName${index}` for each one of them.

The dialect is `mssql`. The sequelize version is 4.3.2 or latest from github (both not working).

Here is a code example:
```js
database.query('SELECT * FROM Table WHERE Id IN ($ids);', {
  bind: {
    ids: [1, 2, 3, 4]
  }
});
```

leibale:
fix #7968 - mssql bind support - add support for array in bind parameters.
@renaudholcombe can you please review this commit?

### DIFF
--- a/test/integration/sequelize.test.js
+++ b/test/integration/sequelize.test.js
@@ -527,6 +527,16 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
       });
     });
 
+    it('array in `bind` property working', function() {
+      let logSql;
+      return this.sequelize.query({query: 'select $prop', bind: {prop: [1]}}, {type: this.sequelize.QueryTypes.SELECT, logging(s) { logSql = s; }}).then(result => {
+        expect(result).to.deep.equal([{'': 1}]);
+        if (dialect === 'mssql') {
+          expect(logSql.indexOf('@prop_0')).to.be.above(-1);
+        }
+      });
+    });
+
     it('dot separated attributes when doing a raw query without nest', function() {
       const tickChar = dialect === 'postgres' || dialect === 'mssql' ? '"' : '`',
         sql = 'select 1 as ' + Sequelize.Utils.addTicks('foo.bar.baz', tickChar);
--- a/test/integration/sequelize.test.js
+++ b/test/integration/sequelize.test.js
@@ -527,15 +527,15 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
       });
     });
 
-    it('array in `bind` property working', function() {
-      let logSql;
-      return this.sequelize.query({query: 'select $prop', bind: {prop: [1]}}, {type: this.sequelize.QueryTypes.SELECT, logging(s) { logSql = s; }}).then(result => {
-        expect(result).to.deep.equal([{'': 1}]);
-        if (dialect === 'mssql') {
+    if (dialect === 'mssql') {
+      it('array in `bind` property working', function () {
+        let logSql;
+        return this.sequelize.query({query: 'select $prop', bind: {prop: [1]}}, {type: this.sequelize.QueryTypes.SELECT, logging(s) {logSql = s;}}).then(result => {
+          expect(result).to.deep.equal([{'': 1}]);
           expect(logSql.indexOf('@prop_0')).to.be.above(-1);
-        }
+        });
       });
-    });
+    }
 
     it('dot separated attributes when doing a raw query without nest', function() {
       const tickChar = dialect === 'postgres' || dialect === 'mssql' ? '"' : '`',
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -151,7 +151,15 @@ class Query extends AbstractQuery {
         return seen[key];
       }
       if (values[key] !== undefined) {
-        i = i + 1;
+        i++;
+        if (Array.isArray(values[key])) {
+          return values[key].map((value, index) => {
+            const currentKey = key + '_' + index;
+            bindParam[currentKey] = value;
+            seen[currentKey] = '$' + currentKey;
+            return '@' + currentKey;
+          }).join(',');
+        }
         bindParam[key] = values[key];
         seen[key] = '$' + i;
         return '@' + key;


renaudholcombe:
@leibale, sure I'll take a look. Can you add unit/integration tests?

leibale:
@renaudholcombe I have tried to add tests, but I can't find where is the tests of the function. 
If you can help me out with that, it will be great :)

renaudholcombe:
@leibale, a couple of good ones to look at are:

- /test/integration/sequelize.test.js
- /test/unit/dialects/mssql/query-generator.test.js

I can't speak to why the travis build is failing, it seems like some sort of weird auto configuration problem with the postgres settings.

leibale:
@renaudholcombe is it ok now?

test/integration/sequelize.test.js
```diff
@@ -527,6 +527,16 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
       });
     });
 
+    if (dialect === 'mssql') {
+      it('array in `bind` property working', function () {
+        let logSql;
+        return this.sequelize.query({query: 'select $prop', bind: {prop: [1]}}, {type: this.sequelize.QueryTypes.SELECT, logging(s) {logSql = s;}}).then(result => {
+          expect(result).to.deep.equal([{'': 1}]);
+          expect(logSql.indexOf('@prop_0')).to.be.above(-1);
```
felixfbecker: Asserting on log statements is a bad practice, if you want to assert the SQL use an SQL unit test

leibale: I copy the test and change it a bit (to send an array). If you have a problem with the test you have a problem with most of the tests in the file.

leibale:
@felixfbecker without that no one will use the bind feature with mssql.

felixfbecker:
Why do you think so? This is not specific to MSSQL

test/integration/sequelize.test.js
felixfbecker: I actually do :D
they may have been written before we had SQL unit tests

leibale:
Specific with mssql no one will use it. We have to find a solution.

felixfbecker:
You just repeated what you said - why do you think noone uses `bind` support because it doesn't handle arrays differently than any other value, specifically in MSSQL?

leibale:
In my application about half of the queries has array parameter, and if i cannot send array in the bind i will use replacements, and if I have to use replacements in half of my application i probably will use it in the would application.

I don't understand what is the problem with my fix? Anyway this function changes the query.

felixfbecker:
`replacements` is a way to escape values and insert them in a query.
Query `SELECT x FROM y WHERE z = ?`, Replacements `1`
will result in the SQL 
`SELECT x FROM y WHERE z = 1` <-- Query is changed

`bind` is a way to build prepared statemetents.
**There is a 1:1 mapping between placeholder and value**.
Query `SELECT x FROM y WHERE z = ?`, Bind `1`
will result in the SQL
`SELECT x FROM y WHERE z = ?` <-- Query is not changed
with `1` passed to the DB server as value.

In Postgres, you can do what you want with
`SELECT x FROM y WHERE z ANY (?)`, Bind `[1, 2, 3]`
which will result in the SQL
`SELECT x FROM y WHERE z ANY ($1)` <-- Query is not changed (except syntax)
with `[1, 2, 3]` (as an array) passed to the DB server as a value.

If MSSQL doesn't support it, you need to use replacements your build your query with one placeholder per array item yourself. 

I don't know if MSSQL supports arrays, but if it does now or sometime in the future, Sequelize would have no way to differentiate between you wanting to insert a single array value or multiple single values. As an example, if Postgres did what you requested:
`INSERT INTO y (x) VALUES (?)` with bind `[1, 2, 3]`
would result in the totally invalid SQL
`INSERT INTO y (x) VALUES (?, ?, ?)` <-- Query would have changed
with values `1`, `2`, `3` passed individually.
I am against behaving different here depending on dialect.

renaudholcombe:
We might be going a bit in circles here, here's my two cents to try and bring things back.

Sql Server does support the use of the IN keyword in parameterized queries (it's equivalent to the IN keyword in Oracle, ANY is a little different), which does lend itself to the use of an array within the parameter value list. @leibale, what you're attempting to do is already possible by merely treating the elements of your as separate parameters rather than a single parameter, but with your change it's organizationally cleaner. I can see how it looks a little wonky with the use of the '@prop_0' variable naming syntax. It might make more sense to merely treat the parameter array as an extension of the existing naming convention so given a parameter value list like this:

` ['val1', 'val2', ['val3, val4]`

It would be generated as 
`@0, @1, @2, @3`

Rather than the current: 
`@0, @1, @key_0, @key_1`

This would make more sense from the conventions standpoint of the end sql.

renaudholcombe:
Just to clarify my previous comment, Sql Server does not support natively passing arrays into parameterized queries. The change being made is purely for making the parameter input make a little more sense from a developer standpoint. Regardless of whether the parameter value is a single value or an array, it will need to be flattened as part of the sql generation (which is what the pull request is doing).

felixfbecker:
Point is, the amount of placeholders that end up in the SQL is no longer == the amount of values in the `bind` array. That messes with the semantics of `bind` as I outlined above, where one placeholder is inserted per value in the `bind` array in the values inside the `bind` array are sent as-is to the DB server. All the other dialects behave that way for `bind` queries as far as I'm aware. 

What you desire could be easily accomplished with a oneliner in your own code
```js
'SELECT x FROM y WHERE z IN (' + values.map((v, i) => '$' + (i + 1)).join(',') + ')'
```
and you could easily put that in a helper function, for example
```js
function in(values) {
  return ' IN (' + values.map((v, i) => '$' + (i + 1)).join(',') + ')';
}
'SELECT x FROM y WHERE z ' + in(values)
// or
`SELECT x FROM y WHERE z ${in(values)}`
```

> Sql Server does not support natively passing arrays into parameterized queries.

Couldn't MSSQL support arrays (or JSON) in the future?

renaudholcombe:
@felixfbecker, it just clicked for me; you're absolutely right, I neglected the key portion of the binding (I think I was mentally using a different use case from the submitter) and this would definitely cause some problems. Your suggestions make more sense.

As far as Sql Server supporting arrays/json, it does support arrays in a fashion through the use of Table Value Parameters, but json is probably a ways down on the road map (given MS's foray into that kind of thing with XPath and XML, I suspect they're a little gun-shy on complex object parsing).

asztal:
SQL Server 2016 supports JSON.

Not useful to me until my company upgrades to it, but useful for someone maybe.

leibale:
@asztal i'm using SQL Server 2016, how can i send an array as parameter?

asztal:
Since I can't use it I haven't really looked into it. But from what I know, there isn't actually a JSON data type, it's just functions for working with JSON as nvarchar(max), which can be used in a variety of places. So you would just pass your array as a string generated with JSON.stringify.

https://blogs.technet.microsoft.com/dataplatforminsider/2016/01/05/json-in-sql-server-2016-part-1-of-4/

PR #7982, opened 2017-07-20
test(include.where): Add a failing test for deep include with where

vizath:
Hey,
Thanks for making Sequelize :tada: 
We are in the process of migrating to v4 and we ran into some undocumented breaking changes like this one.
It was easier to create a failing test than to try to explain it.
v3 was returning 1 element when adding wheres into deep includes, and v4 returns elements with empty relations.
This test is using `A belongsToMany B belongsTo C`, but is also failing on other relation types (I have another use case using `A belongsTo B hasOne C` / `C belongsTo B hasMany A`).

Tested using MySQL.

Was this intended?

v4 output:
```
[
  {
    "id": 1,
    "Roles": [
      {
        "id": 1,
        "UserRole": {
          "UserId": 1,
          "RoleId": 1
        },
        "Permissions": [
          {
            "id": 1,
            "RolePermission": {
              "RoleId": 1,
              "PermissionId": 1
            }
          }
        ]
      }
    ]
  },
  {
    "id": 2,
    "Roles": []
  }
]
```

v3 output:
```
[
  {
    "id": 1,
    "Roles": [
      {
        "id": 1,
        "UserRole": {
          "UserId": 1,
          "RoleId": 1
        },
        "Permissions": [
          {
            "id": 1,
            "RolePermission": {
              "RoleId": 1,
              "PermissionId": 1
            }
          }
        ]
      }
    ]
  }
]
```

### DIFF
--- a/test/integration/include/findAll.test.js
+++ b/test/integration/include/findAll.test.js
@@ -648,7 +648,7 @@ describe(Support.getTestDialectTeaser('Include'), () => {
               ]}
             ]
           }).then(as => {
-            expect(as.length).to.be.ok;
+            expect(as.length).to.equal(1);
 
             as.forEach(a => {
               expect(a.b.c.d.e.f.g.h).to.be.ok;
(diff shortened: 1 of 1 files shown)


test/integration/include/findAll.test.js
```diff
@@ -648,7 +648,7 @@ describe(Support.getTestDialectTeaser('Include'), () => {
               ]}
             ]
           }).then(as => {
-            expect(as.length).to.be.ok;
+            expect(as.length).to.equal(1);
```
vizath: This is still passing as expected

test/integration/include/findAll.test.js
```diff
@@ -2053,5 +2053,56 @@ describe(Support.getTestDialectTeaser('Include'), () => {
...
+                model: Permission,
+                where: {
+                  id: 1,
+                },
+              },
+            ],
+          }],
+        }))
+        .then(users => {
+          expect(users.length).to.equal(1);
```
vizath: `users.length` here is `2`

PR #7984, opened 2017-07-20
ci(deploy): use Travis' npm deploy provider

### LINKED ISSUE #7983: Investigate if we can use Travis' native npm deploy
Currently the build leader (first job) runs `npm run semantic-release`, which prepares the package.json, waits for all jobs to finish and succeed, then runs `npm publish` and cuts a release on GitHub. The problem is that it always times out waiting for the other jobs to finish. Travis also has the integrated npm deploy provider, which would handle this better. We just need to make sure we run the preparation in `before_deploy` and cut the release in `after_deploy`.

felixfbecker:
Closes #7983

I don't know if this works, the only way to find out is to merge.

There could be an issue with commits that don't trigger a release, e.g. `chore` or `docs` commits. The npm publish provider still tries to publish and fails with an error like "version already exists".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,21 +43,35 @@ script:
   - npm run lint
   - "if [ $COVERAGE ]; then npm run cover && bash <(curl -s https://codecov.io/bash) -f coverage/lcov.info; else npm run test; fi"
 
-after_success:
-  - npm run semantic-release
-
 branches:
   except:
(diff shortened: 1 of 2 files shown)


sushantdhiman:
I will done something similar in my personal projects, I will have this issue fixed in different branch

PR #7986, opened 2017-07-20
4.4.0 breaks pg connection initialization

### LINKED ISSUE #7985: Connection-manager Error with Sequelize 4.4. and pg 7.0.1 (TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined)
## What you are doing?

I just upgraded to Sequelize 4.4.0 and [node-postgres](https://github.com/brianc/node-postgres) (pg) version 7.0.1 

Then, when I call some function that connect to database (for example `findAll()` or `sync()`) the connection to Postgres database fails with error. 

It happens, for example, when I run this  example from the documentation

```js
  username: Sequelize.STRING,
  birthday: Sequelize.DATE
});
sequelize.sync()
  .then(() => User.create({
    username: 'janedoe',
  }))
  .then(jane => {
    console.log(jane.get({
      plain: true
    }));
  })
```

Sequelize should connect to Postgres and create tables, insert records and so on

Connection to postgres database fails with the following error:
```
TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined
    at Promise.then.result (C:\test\node_modules\sequelize\lib\dialects\postgres\connection-manager.js:157:31)
    at tryCatcher (C:\test\node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (C:\test\node_modules\bluebird\js\release\promise.js:512:31)
    at Promise._settlePromise (C:\test\node_modules\bluebird\js\release\promise.js:569:18)
    at Promise._settlePromise0 (C:\test\node_modules\bluebird\js\release\promise.js:614:10)
    at Promise._settlePromises (C:\test\node_modules\bluebird\js\release\promise.js:693:18)
    at Async._drainQueue (C:\test\node_modules\bluebird\js\release\async.js:133:16)
    at Async._drainQueues (C:\test\node_modules\bluebird\js\release\async.js:143:10)
    at Immediate.Async.drainQueues (C:\test\node_modules\bluebird\js\release\async.js:17:14)
    at runCallback (timers.js:651:20)
    at tryOnImmediate (timers.js:624:5)
    at processImmediate [as _immediateCallback] (timers.js:596:5)
```

When I switch back to  pg 6.4.0 the same code runs nicely. 
Interesting: With  pg 6.4.0,  it works with both Sequelize 4.4.0 and 4.3.2 
(This shows also that my database connection options are correct)

But the combination of 4.4.0 and pg 7 gives my this error.  Am I doing something wrong? 

Dialect:  postgres
Database version: 9.6
Sequelize version:  4,4.0

clncln1:
### Pull Request check-list

### Description of change

In case the initial query contains multiple statements in the [postgres connection manager](https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/connection-manager.js#L156), the result will be an array which is not being handled here. Since we are only interested in the rows from the very last statement, we can skip the rest.

This fixes #7985

### DIFF
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -153,10 +153,9 @@ class ConnectionManager extends AbstractConnectionManager {
         query += 'SELECT typname, oid, typarray FROM pg_type WHERE typtype = \'b\' AND typname IN (\'hstore\', \'geometry\', \'geography\')';
       }
 
-      return new Promise((resolve, reject) => connection.query(query, (error, result) => error ? reject(error) : resolve(result))).then(result => {
-        if (Array.isArray(result)) {
-          result = result.pop();
-        }
+      return new Promise((resolve, reject) => connection.query(query, (error, result) => error ? reject(error) : resolve(result))).then(results => {
+        const result = Array.isArray(results) ? results.pop() : results;
+        
         for (const row of result.rows) {
           let type;
           if (row.typname === 'geometry') {
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -8,6 +8,7 @@ const sequelizeErrors = require('../../errors');
 const semver = require('semver');
 const dataTypes = require('../../data-types');
 const moment = require('moment-timezone');
+const _ = require('lodash');
 
 class ConnectionManager extends AbstractConnectionManager {
   constructor(dialect, sequelize) {
@@ -154,6 +155,9 @@ class ConnectionManager extends AbstractConnectionManager {
       }
 
       return new Promise((resolve, reject) => connection.query(query, (error, result) => error ? reject(error) : resolve(result))).then(result => {
+        if (_.isArray(result)) {
+          result = result.pop();
+        }
         for (const row of result.rows) {
           let type;
           if (row.typname === 'geometry') {
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -8,7 +8,6 @@ const sequelizeErrors = require('../../errors');
 const semver = require('semver');
 const dataTypes = require('../../data-types');
 const moment = require('moment-timezone');
-const _ = require('lodash');
 
 class ConnectionManager extends AbstractConnectionManager {
   constructor(dialect, sequelize) {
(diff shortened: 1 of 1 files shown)


lib/dialects/postgres/connection-manager.js
```diff
@@ -154,6 +155,9 @@ class ConnectionManager extends AbstractConnectionManager {
       }
 
       return new Promise((resolve, reject) => connection.query(query, (error, result) => error ? reject(error) : resolve(result))).then(result => {
+        if (_.isArray(result)) {
```
felixfbecker: No need for lodash here, just use `Array.isArray()`

clncln1:
there's a couple of SETs and then you are trying to fetch the type oids which is what you are interested in

lib/dialects/postgres/connection-manager.js
```diff
@@ -154,6 +154,9 @@ class ConnectionManager extends AbstractConnectionManager {
       }
 
       return new Promise((resolve, reject) => connection.query(query, (error, result) => error ? reject(error) : resolve(result))).then(result => {
+        if (Array.isArray(result)) {
+          result = result.pop();
```
felixfbecker: this line is uncovered, can you add a test that prevents a regression?

alitaheri: @felixfbecker Not sure how a test can reach this place. The better approach would be to increment the `pg`'s version to v7 for tests, so they all fail 😅 

Also, please consider changing to:

```js
...(results => {
const result = Array.isArray(results) ? results.pop() : results;
...
})
```

Assigning to a function parameter is not very pretty, or readable for that matter.

felixfbecker: Incrementing the version to v7 would disallow us to make sure we still support v6. I am hesitant to bloat the build matrix further.

felixfbecker:
@alitaheri So you are saying this is not testable?

alitaheri:
@felixfbecker Well as far as my brain can handle, the only way to test something like this is to actually let pg v7 run the code and return the result. So either change the version for tests, or add more stuff to the build matrix, or mock pg for a single test. The last one is gonna take hours 😱 😱 

Anyway, I think we should merge this to reduce the noise for now and come up with a test case later on (if ever needed).

PR #7989, opened 2017-07-20
fix(query-generator): add parentheses to queries when using ALL with LIKE

chriskalmar:
fixed an issue with LIKE/ILIKE when used with the ALL keyword.
it already worked fine for the ANY keyword, but was missing the parentheses when I tried with ALL.
extended the tests a bit to cover all LIKE/ILIKE/NOT combinations.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -2304,8 +2304,11 @@ const QueryGenerator = {
       if (escapeValue) {
         value = this.escape(value, field, escapeOptions);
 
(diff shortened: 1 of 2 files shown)


chriskalmar:
I think I ran into the same issue as described in #7773.
One of the 8 build tests returned this error:
`Error: Expected 0 running queries. 1 queries still running`

Running the very same build tests on my fork gave me a green travis response (had to retry it once though). 
Could someone (with access) be so kind as to restart the build on travis for me?
thanks a lot

chriskalmar:
thanks @felixfbecker

PR #7995, opened 2017-07-22
docs(associations): fix syntax error

mrosata:
Just fixing some JavaScript errors in the association docs

### DIFF
--- a/docs/associations.md
+++ b/docs/associations.md
@@ -760,8 +760,8 @@ const Address = this.sequelize.define('address', {
   zip: Sequelize.STRING,
 });
 
-const Product.User = Product.belongsTo(User);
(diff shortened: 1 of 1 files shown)


docs/associations.md
```diff
@@ -760,8 +760,8 @@ const Address = this.sequelize.define('address', {
   zip: Sequelize.STRING,
 });
 
-const Product.User = Product.belongsTo(User);
-const User.Addresses = User.hasMany(Address);
+Product.User = Product.belongsTo(User);
+User.Addresses = User.hasMany(Address);
```
felixfbecker: I actually would recommend to not assign this to anything as it will be available under `Product.associations`

PR #8029, opened 2017-07-29
Fix pg v7 regressions

### LINKED ISSUE #7998: node-postgres version 7 incompatibilities
Any version of sequelize is incompatible with version 7. The main reason is that pg7 returns multiple result objects instead of one with joined results from multiple statements. Therefore, **any** multiple statement query breaks the postgre's query runner.

I have tested sequelize with pg7 and I see a dozen failed tests. We should decide how we should handle multiple statements. I see these options:

1. Always return the last result as done in #7986 
2. Add a new query type ( MULTIPLE ) and return arrays of results when chosen, the last one otherwise

Before this is fixed, there is no choice but to use pg ^6. 

__Dialect:__ postgres
__Database version:__ any
__Sequelize version:__ any

alitaheri:
This hacks around #7998.

I wrote a function which squashes the results returned by pg. except for when raw query doesn't have a type. In that case it returns whatever returned by pg (I don't think we should change this behavior).

This is a temporary fix until we come up with a solution that satisfies every use-case. Might require architectural changes. Since other dialects support multiple statements too.

Until then this fixes #7998

### DIFF
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -8,6 +8,38 @@ const Promise = require('../../promise');
 const sequelizeErrors = require('../../errors.js');
 const _ = require('lodash');
 
+function squashResults(results) {
+  if (!Array.isArray(results)) {
+    return results;
+  }
+
+  let rowCount = 0;
+  let rows = [];
+  let fields = [];
+
+  for (const result of results) {
+    if (!Number.isNaN(result.rowCount)) {
+      rowCount += result.rowCount;
+    }
+
+    if (result.fields.length > 0 && fields.length === 0) {
+      fields = result.fields;
+    }
+
+    if (result.rows.length > 0) {
+      rows = rows.concat(result.rows);
+    }
+  }
+
+  const last = results[results.length - 1];
+
+  return Object.assign({}, last, {
+    rows,
+    fields,
+    rowCount: rowCount === 0 ? NaN : rowCount,
+  });
+}
+
 class Query extends AbstractQuery {
   constructor(client, sequelize, options) {
     super();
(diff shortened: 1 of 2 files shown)


alitaheri:
@felixfbecker The fields will come from the first result that contains them. The rows are the concatenation of all rows from all results. The rowCount is the accumulation of all non `NaN` rowCounts. And everything else `command, parsers, etc.` are from the last one. Anything out of place?

Also, when the query is raw with no type, the metadata might be an array ( possible breaking change ). But I'm not sure what others would expect of this. We explicitly notify the users that the results and metadata are dialect specific. so This is expected I guess. Right?

4e1ovek:
Faced with this problem, tried to understand and find a solution. Later came across this merge request.

sequelize/lib/dialects/postgres/query.js
screenshot

Array in queryResult:
screen shot 2017-07-29 at 18 59 05

node v7.7.3, npm v5.3.0, sequelize v4.4.2, pg v7.0.2

My version of fixes:
```javascript
.then(queryResults => {
  debug(`executed(${this.client.uuid || 'default'}) : ${this.sql}`);

  if (benchmark) {
    this.sequelize.log('Executed (' + (this.client.uuid || 'default') + '): ' + this.sql, Date.now() - queryBegin, this.options);
  }

  return queryResults;
})
.then(queryResults => {
  const queryResult = Array.isArray(queryResults) ? queryResults.pop() : queryResults;
  const rows = queryResult.rows;
  const rowCount = queryResult.rowCount;
```

alitaheri:
@4e1ovek That won't always work. Some queries might have their main stuff in the second statement.

corysimmons:
I don't mean to be one of those entitled devs, but can contribs/maintainers take this one a bit more seriously and get a fix, or at least a documented workaround, merged?

Broken enums is a **huge** problem.

alitaheri:
@corysimmons downgrade to pg 6 for now. it's not just the enums that are broken. Any multi-statement query will fail.

corysimmons:
@alitaheri So `pg` @ `6.4.2`. Do I need to add `pg-native`? Do I need to downgrade `pg-hstore` to something?

alitaheri:
@corysimmons You don't need `pg-native`. Best not use it anyway, some features aren't implemented there. And `pg-hstore` is a simple serializer/deserializer with no dependencies. Just install `pg@6.4.2` and your issue will be fixed for now.

corysimmons:
Thanks Ali. Maybe we can update the README with a warning about pg7 while a fix is in the works?

gabegorelick:
Is there anything that needs to be done to get this into a mergable state (besides rebasing package.json)?

joaovitoras:
Do you intend to merge this pull request?

PR #8037, opened 2017-07-31
fix(errors): Add check when using create instead of bulkCreate

hwaterke:
### Description of change
Added a check to prevent using create instead of bulkCreate

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1960,6 +1960,11 @@ class Model {
    *
    */
   static create(values, options) {
+
+    if (Array.isArray(values)) {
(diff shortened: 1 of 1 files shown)


sushantdhiman:
I dont think most people get confused by this, `create` and `bulkCreate` names should be sufficient. If there is enough community demand, in future I will consider adding this

hwaterke:
Well, this is not a major refactoring but a simple safety net.
When passing an array to `create` you get a very strange error and it takes time to figure out that you're not supposed to do it. I'm still convinced this is worth merging...

PR #8038, opened 2017-07-31
fix: maintable.as should be quoted using  quoteIdentifier method

harish2704:
fix: maintable.as is different from tablename.
should be quoted using  quoteIdentifier method

### Pull Request check-list

### Description of change

This PR is not for solving any bug or critical issues. It is for improving quality of code base.
Sequelize has  well organized code base and it is easy to customize it for complex application.

This is PR will a fix a minor deviation from the  existing code convention

This is the summary of change: 
while generating select query, we are quoting "tablename" part and "tablename.as" part using same function ( quoteTable function ).

But, tablename can have a schema part but the "tablename.as" part can not have schema. so that part should be quoted using quoteIdentifier method

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -958,9 +958,9 @@ const QueryGenerator = {
 
     // resolve table name options
     if (options.tableAs) {
-      mainTable.as = this.quoteTable(options.tableAs);
+      mainTable.as = this.quoteIdentifier(options.tableAs);
     } else if (!Array.isArray(mainTable.name) && mainTable.model) {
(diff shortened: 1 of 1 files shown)


eseliger:
@harish2704  Could you please amend and force push your commit to retrigger appveyor?

harish2704:
@eseliger  Thanks for the response. I will be trying the following procedure.
* amend my last commit
* make some simple changes like add a space or somthing 
* commit
* force push

Also please take a took at this link. It may help you http://help.appveyor.com/discussions/questions/305-trigger-a-build

harish2704:
Hi, Is there any update regarding this PR ?

PR #8068, opened 2017-08-07
Simplify where item query

yonjah:
### Pull Request check-list

### Description of change

Abstract dialect query-generator `whereItemQuery` method was more than 400 lines and had a cyclic complexity of 85. That made it really hard to understand what's going on.
Since I have a few other issues which require some changes to this method I thought it would be a lot better if I (and other developers) have better understanding of what's going on before I do even minor logic changes.
I splited the method to smaller private methods.
I tried to converge logic and simplify logic where it was possible.
I also noticed a few bugs in the original implementation which I solved and added tests for 
(Issue #3824 which was actually fixed in #4298 but than got overwritten in #7094)


yonjah:
Working on this I noticed Sequelize eslint rules can benefit from a few extra rules.
It might be useful to extend the  "eslint:recommended" rules which protect against using duplicate keys in objects and reassigning values to const among other things (those are the two I noticed that were missing).
Also I used the complexity rule to calculate the cyclic complexity of the methods ( but I'm not sure if this is as important as the other rules)

alitaheri:
Great job 👍 👍 I've had a really hard time when I was working on range types. Thanks a lot 😍

mickhansen:
I skimmed it, and it looks like a decent way to go.
One thing i considered back when i wrote that monstrosity was to split off the logic into a seperate class/object with methods there, i see you added a few new methods to the general QueryGenerator class so i think it might make sense to split it out into a WHERE query builder.

yonjah:
@mickhansen I thought of separating the logic to a different class but didn't want to diverge too much of the original implementation.
If you think it will be better I can move it out and update the pull request.

One small thing though is that currently each dialect have their own query-generator which is being extended by the dialect and being loaded into it.
Adding a Where Query Builder will mean that each query generator will load its own extension to that as well. It's not a major issue but this design feels a bit awkward to work with.

mickhansen:
> If you think it will be better I can move it out and update the pull request.

I think it would be great if we could seperate it.

> Adding a Where Query Builder will mean that each query generator will load its own extension to that as well. It's not a major issue but this design feels a bit awkward to work with.

It definitely feels a bit off sometimes, open to suggestions.

yonjah:
> It definitely feels a bit off sometimes, open to suggestions.

I wish I had some concrete suggestion but I have nothing specific to offer at the moment

> I think it would be great if we could seperate it.

I looked into it a bit more.
`whereItemQuery` `whereItemsQuery` and all the private methods I added are using a few other methods on the query-generator like -
`escape`, `handleSequelizeMethod`, `quoteIdentifier`, `quoteTable`, `jsonPathExtractionQuery`

And untangling them out of the `query-generator` will not be simple (or necessarily make sense).
I think some major refactoring will need to be done to actually get something that feels consistent and make sense across all dialects. Not really sure if I currently have the time to do such a major change ( But I really hope I'll have it in the future I think Sequelize can benefit a lot from a simple and extendable dialect api that will allow external dialects to be loaded in).

I can still move the where logic to a separate module but there will need to be some kind of cyclic dependency between it and the query-generator which again feels awkward

yonjah:
@mickhansen do you have an insight on how to separate it or should I leave it as is ?

test/unit/sql/where.test.js
```diff
@@ -821,6 +847,18 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
           sqlite: "(json_extract(`User`.`data`, '$.name.last') = 'Simpson' AND json_extract(`User`.`data`, '$.employment') != 'None')"
         });
 
+        testsql('data', {
+          price: 5,
+          name: 'Product'
```
janmeier: Could you add unit tests for boolean and date matching as well

yonjah: Do you mean to duplicate each $between and $in tests to include all possible casting types ?
Cause we have tests for those castings in Line 934 and 949.

I added the $between and $in tests because previous logic didn't apply casting at all to them

janmeier: > Cause we have tests for those castings in Line 934 and 949.

Roger that!

PR #8088, opened 2017-08-09
fix(upsert): use correct timestamp field names when generating query

### LINKED ISSUE #6981: Upsert doesn't respects createdAt and updatedAt renamed fields
## What you are doing?
I am trying to execute a upsert with a model that renamed createdAt and updatedAt fields.

```js
var sequelize = new Sequelize(config.database, config.username, config.password, config);
const test = sequelize.define('test', {
  testId: {type: Sequelize.INTEGER, allowNull: false, field: 'test_id'},
  wserial: {type: Sequelize.BIGINT, allowNull: false},
  status: {type: Sequelize.TEXT},
  createdAt: {type: Sequelize.DATE, field: 'created_at'},
  updatedAt: {type: Sequelize.DATE, field: 'updated_at'}
});
test.sync().then(function () {
  // Create works properly, changing createdAt and updatedAt to created_at and updated_at.
  test.create({
    testId: 1,
    wserial: 123456,
    status: 'DELIVERED'
  }).then(function () {
    // I was expecting that upsert works as well
    test.upsert({
      testId: 2,
      wserial: 456789,
      status: 'REFUSED'
    }).then(function () {
      console.log('Expected')
    }).catch(function (err) {
      // but upsert not respects field option to createdAt and updatedAt like create does 
      console.error(err);
    });
  });
});
```

I was expecting upsert generate SQL with fields createdAt and updatedAt renamed like happens in create method.

I am getting the following exception: SequelizeDatabaseError: ER_BAD_FIELD_ERROR: Unknown column 'createdAt' in 'field list'.

Maybe this issue, if is really a new one, is related with #2946 issue.

#### Test case Output

Executing (default): CREATE TABLE IF NOT EXISTS `tests` (`id` INTEGER NOT NULL auto_increment , `test_id` INTEGER NOT NULL, `wserial` BIGINT NOT NULL, `status` TEXT, `created_at` DATETIME, `updated_at` DATETIME, PRIMARY KEY (`id`)) ENGINE=InnoDB;
Executing (default): SHOW INDEX FROM `tests`
Executing (default): INSERT INTO `tests` (`id`,`test_id`,`wserial`,`status`,`created_at`,`updated_at`) VALUES (DEFAULT,1,123456,'DELIVERED','2016-12-13 15:00:35','2016-12-13 15:00:35');
Executing (default): INSERT INTO `tests` (`test_id`,`wserial`,`status`,`createdAt`,`updatedAt`) VALUES (2,456789,'REFUSED','2016-12-13 15:00:35','2016-12-13 15:00:35') ON DUPLICATE KEY UPDATE `test_id`=VALUES(`test_id`), `wserial`=VALUES(`wserial`), `status`=VALUES(`status`), `updatedAt`=VALUES(`updatedAt`);
SequelizeDatabaseError: ER_BAD_FIELD_ERROR: Unknown column 'createdAt' in 'field list'

__Dialect:__ mysql
__Database version:__ 5.6.34
__Sequelize version:__ 3.27.0

Thanks in advance

chuntley:
### Pull Request check-list

### Description of change

Fixes #6981

When using a custom field name for a timestamp property in an upsert, the upsert ignores the field property and the query fails. For example:

```
  const ShareModel = sequelize.define('test_stuff', {
    myId: {
      type: DataTypes.INTEGER,
      primaryKey: true,
      allowNull: false
    },
    createdAt: { type: DataTypes.DATE, field: 'created_at' },
    updatedAt: { type: DataTypes.DATE, field: 'updated_at' },
  }, {
    freezeTableName: true,
  });
```

In the query that upsert generates, it will use the field `createdAt` and `updatedAt`, instead of the correct field names `created_at` and `updated_at`.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2190,12 +2190,12 @@ class Model {
 
       // Attach createdAt
       if (createdAtAttr && !updateValues[createdAtAttr]) {
-        const attribute = this.rawAttributes[createdAtAttr].field || createdAtAttr;
-        insertValues[attribute] = this._getDefaultTimestamp(attribute) || now;
+        const field = this.rawAttributes[createdAtAttr].field || createdAtAttr;
+        insertValues[field] = this._getDefaultTimestamp(createdAtAttr) || now;
       }
       if (updatedAtAttr && !insertValues[updatedAtAttr]) {
-        const attribute = this.rawAttributes[updatedAtAttr].field || updatedAtAttr;
-        insertValues[attribute] = updateValues[attribute] = this._getDefaultTimestamp(attribute) || now;
+        const field = this.rawAttributes[updatedAtAttr].field || updatedAtAttr;
+        insertValues[field] = updateValues[field] = this._getDefaultTimestamp(updatedAtAttr) || now;
       }
 
       // Build adds a null value for the primary key, if none was given by the user.
--- a/lib/model.js
+++ b/lib/model.js
@@ -2172,8 +2172,8 @@ class Model {
       hooks: true
     }, Utils.cloneDeep(options || {}));
 
-    const createdAtAttr = this._timestampAttributes.createdAt;
-    const updatedAtAttr = this._timestampAttributes.updatedAt;
+    const createdAtAttr = this.rawAttributes[this._timestampAttributes.createdAt].field || this._timestampAttributes.createdAt;
+    const updatedAtAttr = this.rawAttributes[this._timestampAttributes.updatedAt].field || this._timestampAttributes.updatedAt;
     const hadPrimary = this.primaryKeyField in values || this.primaryKeyAttribute in values;
     const instance = this.build(values);
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -2172,8 +2172,8 @@ class Model {
       hooks: true
     }, Utils.cloneDeep(options || {}));
 
-    const createdAtAttr = this.rawAttributes[this._timestampAttributes.createdAt].field || this._timestampAttributes.createdAt;
-    const updatedAtAttr = this.rawAttributes[this._timestampAttributes.updatedAt].field || this._timestampAttributes.updatedAt;
+    const createdAtAttr = this._timestampAttributes.createdAt;
+    const updatedAtAttr = this._timestampAttributes.updatedAt;
     const hadPrimary = this.primaryKeyField in values || this.primaryKeyAttribute in values;
     const instance = this.build(values);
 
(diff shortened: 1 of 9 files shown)


lib/model.js
```diff
@@ -2172,8 +2172,8 @@ class Model {
       hooks: true
     }, Utils.cloneDeep(options || {}));
 
-    const createdAtAttr = this._timestampAttributes.createdAt;
-    const updatedAtAttr = this._timestampAttributes.updatedAt;
+    const createdAtAttr = this.rawAttributes[this._timestampAttributes.createdAt].field || this._timestampAttributes.createdAt;
+    const updatedAtAttr = this.rawAttributes[this._timestampAttributes.updatedAt].field || this._timestampAttributes.updatedAt;
```
sushantdhiman: I suspect this could fail when timestamps are disabled, please add a test for that case

chuntley: You are correct, it did fail when timestamps were disabled. I have updated it and added a test.

lib/model.js
```diff
@@ -2190,10 +2190,12 @@ class Model {
 
       // Attach createdAt
       if (createdAtAttr && !updateValues[createdAtAttr]) {
-        insertValues[createdAtAttr] = this._getDefaultTimestamp(createdAtAttr) || now;
+        const attribute = this.rawAttributes[createdAtAttr].field || createdAtAttr;
+        insertValues[attribute] = this._getDefaultTimestamp(attribute) || now;
```
sushantdhiman: I think you still want to pass `this._getDefaultTimestamp(createdAtAttr)`

lib/model.js
```diff
@@ -2190,10 +2190,12 @@ class Model {
 
       // Attach createdAt
       if (createdAtAttr && !updateValues[createdAtAttr]) {
-        insertValues[createdAtAttr] = this._getDefaultTimestamp(createdAtAttr) || now;
+        const attribute = this.rawAttributes[createdAtAttr].field || createdAtAttr;
```
sushantdhiman: Use `field`

lib/model.js
```diff
@@ -2190,10 +2190,12 @@ class Model {
...
+        insertValues[attribute] = this._getDefaultTimestamp(attribute) || now;
       }
       if (updatedAtAttr && !insertValues[updatedAtAttr]) {
-        insertValues[updatedAtAttr] = updateValues[updatedAtAttr] = this._getDefaultTimestamp(updatedAtAttr) || now;
+        const attribute = this.rawAttributes[updatedAtAttr].field || updatedAtAttr;
+        insertValues[attribute] = updateValues[attribute] = this._getDefaultTimestamp(attribute) || now;
```
sushantdhiman: Same changes go for this case as well

lib/model.js
chuntley: Change the `const attribute =` to `const field =` ?

lib/model.js
chuntley: It is still being passed on line 2194, just with the correct date attribute.

lib/model.js
sushantdhiman: Yes

lib/model.js
sushantdhiman: If attribute has a field set, this will accidentally pass that field to `_getDefaultTimestamp` where instead we want to pass attribute in all cases

Issue #8081, opened 2017-08-08
TypeError when processing UniqueConstraintError in findOrCreate


cweckesser:
I've written some tests that attempt to prevent duplicated entities from being persisted and I'm making calls to the `findOrCreate` function in the following way:

```js
models.someEntity.findOrCreate({
    where: {
        $or: [{
            property1,
            property3
        }, {
            property2,
            property3
        }]
    },
    defaults: {
        property1: 'some value 1',
        property2: 'some value 2',
        property3: 'some value 3'
    }
})
```

The `someEntity` entity is mapped to a table that has a unique constraint on two columns (`property1` and `property3`) and when I try to insert duplicated data the insertion fails as expected, but it does due to another error that overrules the unique constraint error.

When the call to the `create` function (made from the body of the `findOrCreate` function) throws an error due to the violation of the unique constraint, the following attempt to obtain a value for the `whereFields` variable fails:

```js
...
}).catch(this.sequelize.UniqueConstraintError, err => {
    const whereFields = Object.keys(options.where).map(name => this.rawAttributes[name].field || name);
    ...
```
(snippet from the `lib/model.js` file)

and I get the following error:

```js
TypeError: Cannot read property 'field' of undefined
at Object.keys.map.name (/path/to/my/project/node_modules/sequelize/lib/model.js:2077:92)
    at Array.map (native)
    at create.then.catch.err (/path/to/my/project/node_modules/sequelize/lib/model.js:2077:56)
    at tryCatcher (/path/to/my/project/node_modules/bluebird/js/release/util.js:16:23)
    at /path/to/my/project/node_modules/bluebird/js/release/catch_filter.js:17:41
    at /path/to/my/project/node_modules/continuation-local-storage/context.js:84:17
    at tryCatcher (/path/to/my/project/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/path/to/my/project/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (/path/to/my/project/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (/path/to/my/project/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/path/to/my/project/node_modules/bluebird/js/release/promise.js:689:18)
    at Async._drainQueue (/path/to/my/project/node_modules/bluebird/js/release/async.js:133:16)
[...]

PR #8093, opened 2017-08-10
fix: Flatten where clause object and use lodash to avoid accessing un…

cweckesser:
…defined objects' properties

### Pull Request check-list

### Description of change

---
This change is [Reviewable](https://reviewable.io/reviews/sequelize/sequelize/8093)


cweckesser:
fixes #8081

cweckesser:
@sushantdhiman could you please check this one out, whenever it suits you best?

package.json
```diff
@@ -41,6 +41,8 @@
     "debug": "^3.0.0",
     "depd": "^1.1.0",
     "dottie": "^2.0.0",
+    "env-cmd": "^5.0.0",
```
sushantdhiman: `env-cmd` is a dev dependency, its already present in dev dependencies. Remove it from here

package.json
```diff
@@ -41,6 +41,8 @@
...
+    "flat": "^4.0.0",
```
sushantdhiman: I dont want to include other dependency which will be used in one function only, may be we can use lodash to get leaf keys ?

test/integration/model/create.test.js
```diff
@@ -104,6 +104,38 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+            $or: [{
+              objectId: 'asdasdasd1'
+            }, {
+              objectId: 'asdasdasd2'
+            }]
+          },
+          defaults: {
+            username: 'gottlieb'
+          }
+        })).to.eventually.be.rejectedWith(Sequelize.UniqueConstraintError);
```
sushantdhiman: assert for field as well

package.json
cweckesser: I only aded the following dependency:

`"flat": "^4.0.0",`

But I'll remove it anyways.

lib/model.js
```diff
@@ -2077,8 +2076,22 @@ class Model {
 
         return [instance, true];
       }).catch(this.sequelize.UniqueConstraintError, err => {
-        const flattenedWhere = _.map(_.keys(flat(options.where)), name => _.last(_.split(name, '.')));
-        const whereFields = flattenedWhere.map(name => _.get(this.rawAttributes, `${name}.field`, name));
+        const flattenObject = (obj, flattenedObj, subPath = undefined) => {
```
sushantdhiman: move this to `Utils`

yarn.lock
```diff
@@ -0,0 +1,4265 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
sushantdhiman: Remove this file, add `yarn.lock` to `.gitignore`

test/integration/model/create.test.js
```diff
@@ -131,7 +131,11 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           }
         }).catch(error => {
           expect(error).to.be.instanceof(Sequelize.UniqueConstraintError);
-          expect(error.fields).to.include({ username: 'gottlieb' });
+          if (dialect === 'mysql') {
```
sushantdhiman: Its failing for `sqlite`, a better solution would be to have fields return same object in `sqlite` as other dialects

cweckesser: I think that is out of the scope of the PR. The changes I made, only process the contents of the where clause of the findOrCreate function, which is supposed to be independent of the DB engine. I also saw conditional assertions based on the dialect in other tests in the same file.

sushantdhiman: May be SQLite doesnt support returning values in error, then I have never worked with SQLite in production. Please change the check to `dialect === sqlite` and assert for plain array with `username` in that case, keep current assertion for other dialects

test/integration/model/create.test.js
```diff
@@ -131,7 +131,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           }
         }).catch(error => {
           expect(error).to.be.instanceof(Sequelize.UniqueConstraintError);
-          if (dialect === 'mysql') {
+          if (_.includes(['mysql', 'postgres'], dialect)) {
```
sushantdhiman: interchange if statements and use `dialect === sqilite`, this will fail for `postgres-native` and any new dialects

cweckesser: I modified the assertion so as to check the conflicting field in the errors property (array) of the UniqueConstrainError object, since checking for the fields property would also cause the mssql tests to fail when run by AppVeyor (and in this case I couldn't get any details from the error).

lib/utils.js
```diff
@@ -417,6 +417,18 @@ function removeTicks(s, tickChar) {
 }
 exports.removeTicks = removeTicks;
 
+function flattenObject(obj, flattenedObj, subPath) {
```
sushantdhiman: rather than mutating `flattenedObj` it just simple create one internally and return it. Any specific reasoning for opting this behavior

cweckesser: I changed the function so as to instantiate the returned value and return it when the flattening process is done.

PR #8106, opened 2017-08-13
Handle passing in undefined as an assoc target

dancrumb:
### Pull Request check-list

### Description of change

When using the patterns demonstrated here: https://github.com/sequelize/express-example/blob/master/models/index.js, it's possible to accidentally pass in `undefined` as the target of an association. This results in a stack trace that isn't obvious. Instead, the lines that assume that the target is defined can check that assumption.


dancrumb:
Ooof... This was a bigger change than I expected. I'll kill this PR and put up a new one once I have time to dig into why this fails so much!

PR #8108, opened 2017-08-14
Use ramdisk for postgres tests

eseliger:
This is work in progress don't merge! 

Experimenting with ram disks for Travis tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,6 @@ services:
       POSTGRES_USER: sequelize_test
       POSTGRES_PASSWORD: sequelize_test
       POSTGRES_DB: sequelize_test
-      PGDATA: /dev/shm/pgdata/data
     ports:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -20,6 +20,7 @@ services:
       POSTGRES_USER: sequelize_test
       POSTGRES_PASSWORD: sequelize_test
       POSTGRES_DB: sequelize_test
+      PGDATA: /dev/shm/pgdata/data
```
sushantdhiman: I think you need to mount ramdisk first ? For Circle CI postgres image its pre mounted https://circleci.com/docs/2.0/postgres-config/#optimizing-postgres-images

But here we use different image as we need PostGIS support

sushantdhiman:
https://github.com/sequelize/sequelize/pull/8116

PR #8138, opened 2017-08-16
test(bug): Add failing test for issue

csreid:


sushantdhiman:
BelongsToMany scope needs to set with `through.scope` option http://docs.sequelizejs.com/manual/tutorial/associations.html#n-m

csreid:
Perfect, thank you. 

In that case, the library should provide a warning if a scoped model is used in an association, and a more useful error should be thrown long before query generation.

Issue #7354, opened 2017-03-09
Using require.main.require instead of require for dialects


grimurd:
I wanted to suggest that when sequelize is loading in dialect modules that instead of using require() like normally, you use require.main.require(). Or at least provide the option to do so.

The difference between the two is that normal require() always point to the module that is calling the require method. However require.main.require() uses the scope of the topmost module in the hierarchy which is usually the module that is being run by the user.

My use case is that I have created a small framework for my employer which is using sequelize. When using this framework in a project it is installed using npm and then required by the project. However the framework needs to include all the possible dialects in it's project.json which makes the install take significantly longer which hurts our build process and increases the size on disk.
However if sequelize was using require.main.require() wouldn't have to install all the dialects in the framework but could instead install only the dialects I need directly in the project.

This should not break anything that is using sequelize directly as require.main always points at the top level module.

I hope what I said made sense and I apologize for any grammar errors.

janmeier:
I've never had this issue myself, but it should be fine to go so - PR welcome!

BTW, why do you have to include all dialects? We use sequelize with postgres, and only have pg in our package

grimurd:
@janmeier Because it's a framework, and not all the projects using the framework are using the same database :)

I'll try to make a PR in the next few days.

PR #8141, opened 2017-08-16
feat(dialects): Changed dialectModulePath option to accept either a string or an object

grimurd:
### Pull Request check-list

### Description of change

See my previous issue: #7354

The solution i proposed in the issue didn't work as well as I expected and would have constituted a breaking change.

This new solution modifies the dialectModulePath dialect option so that it now also accepts an object if the user wants to require the module himself. This solution would allow for using require.main.require() without breaking any existing functionality.

As I couldn't find any tests for dialectModulePath and this is a fairly small change I didn't add any tests, but I'm open to it if requested.

I don't think the property name (dialectModulePath) is apt but I felt that it was better to modify it instead of adding a new property(ex. dialectModule) to avoid any possible conflicts if a user specifies both a dialectModulePath as a string and  dialectModule as an object. A alternative solution would have been to deprecate dialectModulePath in favor of dialectModule and check for both until next major version.

### DIFF
--- a/lib/dialects/mssql/connection-manager.js
+++ b/lib/dialects/mssql/connection-manager.js
@@ -18,7 +18,11 @@ class ConnectionManager extends AbstractConnectionManager {
     this.sequelize.config.port = this.sequelize.config.port || 1433;
     try {
       if (sequelize.config.dialectModulePath) {
-        this.lib = require(sequelize.config.dialectModulePath);
+        if (Utils._.isObject(sequelize.config.dialectModulePath)) {
+          this.lib = sequelize.config.dialectModulePath;
+        } else {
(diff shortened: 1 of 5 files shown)


sushantdhiman:
Cant you just pass custom path to this configuration? I am really not getting any benefit of this change.

grimurd:
@sushantdhiman That would usually work but I'm using sequelize as a part of a framework which then gets npm installed into other projects in a variety of configurations so I can't trust the path to always be correct. 

This solution would allow me to use require.main.require() which requires the dialect module through the project using the framework which is much more reliable as the project only needs to have the dialect module installed and node takes care of finding it.

PR #8143, opened 2017-08-17
feat(belongsToMany): add support for non primary key relationships

### LINKED ISSUE #4092: Specify non-primary target key in belongsToMany association
Hi & thanks for the lib. 

I am trying to configure a many to many relationship through a joining model. 
I want to use non-primary string key fields on both models referenced by the join model.

I have tried defining the foreign keys manually against each model field... these are created but ignored, the relation creates its own keys referencing the PKs.
I have tried specifying them as the `foreignKey` and `otherKey`, but this results in the fields in the join table being overwritten as integer fields when they should be strings. 
I have also tried specifying a targetKey when calling `belongsToMany()` to no avail.

Is this intended behavior? 
Is what I am trying to achieve possible?

Thanks in advance

garrypolley:
### Pull Request checklist

### Description of change

fixes #4092

This should allow people to start using `belongsToMany` with keys that are not primary keys. I've added a test to attempt to support this change.


kittencup:
When to merge to master?

garrypolley:
It seems like the limit is getting set on the incorrect part of the query when I make requests.  Thoughts on how to fix this?

test/integration/associations/belongs-to-many.test.js
```diff
@@ -321,6 +321,95 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), () => {
...
+
+      return this.sequelize.sync({force: true}).then(() => {
+        return Promise.join(
+          User.create(),
+          Group.create()
+        ).spread((user, group) => {
+          return user.addGroup(group);
+        }).then(() => {
+          return Promise.join(
+            User.findOne({
```
sushantdhiman: Create more users and group, at least two record in through table, assert the return value to match correct users

lib/dialects/abstract/query-generator.js
```diff
@@ -1513,7 +1513,16 @@ const QueryGenerator = {
     const primaryKeysTarget = association.target.primaryKeyAttributes;
     const tableTarget = includeAs.internalAs;
     const identTarget = association.foreignIdentifierField;
-    const attrTarget = association.target.rawAttributes[primaryKeysTarget[0]].field || primaryKeysTarget[0];
+
+    // Determine the attrTarget, if there is a NonPrimaryBelongsTo in our association
+    // we need to use the targetPrimaryKeyAttribute to ensure we do not default to the standard
+    // primary key.
+    let attrTarget;
```
sushantdhiman: May be both `attrTarget` and `attrSource` should be exposed by associations class

garrypolley: I don't understand this request/comment.

test/integration/associations/belongs-to-many.test.js
garrypolley: Do you mind writing that test? I'm not familiar enough with sequelize to quickly make that test. I primarily use this library with the FeathersJS web framework, and made this pull request so that I could continue to have an easy development experience.

mikefowler:
@garrypolley I'd love to help get this out. Let me know how I can help.

lib/dialects/abstract/query-generator.js
mikefowler: @sushantdhiman, would you clarify this request?

garrypolley:
@mikefowler all help is appreciated. I'm not sure how to create the tests that are requested, and do not understand the association update question/statement. 

Right now I'm working off of my branch in our production application because we needed the many-to-many custom join.

romanlab:
Really looking forward to seeing this released! Hope it gets merged soon

garrypolley:
I'm not likely going to have the time to add the desired tests to this PR. If someone could add those tests that would be appreciated. 

> defining source and target attribute should be done in association class to simplify things in query-interface

I'm not sure I fully understand this request still. Is this saying I need to define the "through" relationship in the "join" class? So instead of stating how a model gets to another in two places I need describe it in 3? 

As of right now, that isn't really supported in the internals, from what I could tell. I think this PR is a nice step towards that goal, however, I'm not going to have the time to do that full rewrite/refactor of those pieces of the code.

sushantdhiman:
I would like to clarify what my initial comment meant

> May be both attrTarget and attrSource should be exposed by associations class

In current query generator code, with this PR a change is introduced. It modify two variables `attrTarget` and `attrSource` based on various conditions. It will be preferable if this information is exposed by `association` itself. So you wont have to compute it in query generator.

This change can be made [here](https://github.com/sequelize/sequelize/blob/3f5927f5bb47db85f5e0401d5a876f652fef1cd3/lib/associations/belongs-to-many.js#L77)

You can also see this PR doesn't change [these](https://github.com/sequelize/sequelize/blob/3f5927f5bb47db85f5e0401d5a876f652fef1cd3/lib/associations/belongs-to-many.js#L250-L257) , I am sure which breaks `add / remove` helpers for belongs to many in non primary key target.

To have this change accepted you will also need to add some test around all these helpers, you can check these tests https://github.com/sequelize/sequelize/blob/master/test/integration/associations/belongs-to-many.test.js, should not be too difficult to add these test.

garrypolley:
I believe this is still an issue. I'm commenting so the stale bot does not close this PR. 

I still do not have the time to finish this PR.

brianalois:
This is needed by so many people. Would love for someone to get this out.

sushantdhiman:
Cleanup, closing this PR. If someone wants to takeover please open a new PR

PR #8151, opened 2017-08-17
fix(abstract/querygenerator): correctly name indexes that live in custom schemas

gabegorelick:
Previously, `queryInterface.addIndex({schema: 'foo', tableName: 'bar'}, ['baz'])` would create an index named `[object_object]_baz`.

### Pull Request check-list

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -424,6 +424,11 @@ const QueryGenerator = {
   },
 
   nameIndexes(indexes, rawTablename) {
+    if (typeof rawTablename === 'object') {
+      // don't include schema in the index name
+      rawTablename = rawTablename.tableName;
+    }
+
     return _.map(indexes, index => {
       if (!index.hasOwnProperty('name')) {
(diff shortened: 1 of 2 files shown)


lib/dialects/abstract/query-generator.js
```diff
@@ -423,6 +423,11 @@ const QueryGenerator = {
   },
 
   nameIndexes(indexes, rawTablename) {
+    if (typeof rawTablename === 'object') {
```
janmeier: You should be able to just call quoteTable https://github.com/gabegorelick/sequelize/blob/693bad925cf57302bda56d5d387942ea7f54fc20/lib/dialects/abstract/query-generator.js#L660

That will include the schema if defined - I guess for multi-tenancy with schemas you might want the name of the 
schema in the index name

gabegorelick: @janmeier I think you'd have to update `quoteTable` so it didn't use `.` as a separator. That makes sense when quoting table names for `FROM` clauses, but not for index names.

janmeier: Right, never mind then :-)

Issue #5320, opened 2016-01-28
Model.attributes should be Model.rawAttributes


mickhansen:

sushantdhiman:
Fixed by https://github.com/sequelize/sequelize/pull/8160

Need to removed Model.attributes in v5

superlbr:
Since attributes were loaded when model init,  what if I changed attributes defined in models and database later and keep the original js program running .I got an error because the attributes still unchanged in memory. Disappointed that I didn't take a picture...

sushantdhiman:
This change is now included in v5 branch https://github.com/sequelize/sequelize/tree/v5

PR #8160, opened 2017-08-19
fix(internal): cleanup code base and straightening internal API

sushantdhiman:
Cleanup `model` a bit

- renamed and refactored `autoIncrementField` to be `autoIncrementAttribute`, because it is an attribute not field
- `this.hasPrimaryKey` removed, not used anywhere duplicate of `this._hasPrimaryKey`
- use `this.rawAttributes` where possible, its now safe to remove `this.attributes` https://github.com/sequelize/sequelize/issues/5320 , which we can do in `v5`
- removed unused `model/attribute` file
- load errors via loop, less verbose
- Fix `UnknownConstraintError` case was not handled for `Postgres`, fixed silently failing test


felixfbecker:
FYI this should be a `refactor`, not a `fix`

[commitizen](https://github.com/commitizen/cz-cli)

PR #8162, opened 2017-08-21
fix(query): does not include primary key when grouping

### LINKED ISSUE #3417: Is there a way to exclude primary key attribute when doing COUNT + LEFT JOIN + GROUP BY + ORDER BY?
I have the following code, its in coffeescript:

``` coffeescript
Reservation.findAll({
    attributes: [
      [DB.Sequelize.fn('COUNT', 'item_id'), 'num_reservations']
    ]
    include: [
      {model: models.Item, as: 'Item', attributes: []}
    ],
    group: ['Item.id']
    order: 'num_reservations ASC'
    })
```

that generates the following SQL:

``` SQL
SELECT "reservation"."id", COUNT('item_id') AS "num_reservations", "Item"."id" AS "Item.id" 
FROM "reservation" AS "reservation" 
LEFT OUTER JOIN "item" AS "Item" ON "reservation"."item_id" = "Item"."id" AND "Item"."deleted_at" IS NULL 
WHERE "reservation"."deleted_at" IS NULL GROUP BY "Item"."id" ORDER BY num_reservations ASC
```

and the error:

```
SequelizeDatabaseError: column "res_reservation.id" must appear in the GROUP BY clause
```

The expected SQL that works would be the same as above just without "reservation"."id" from the SELECT statement:

``` SQL
SELECT COUNT('item_id') AS "num_reservations", "Item"."id" AS "Item.id" 
FROM "reservation" AS "reservation" 
LEFT OUTER JOIN "item" AS "Item" ON "reservation"."item_id" = "Item"."id" AND "Item"."deleted_at" IS NULL 
WHERE "reservation"."deleted_at" IS NULL GROUP BY "Item"."id" ORDER BY num_reservations ASC
```

Is there a way to do this? How can I tell the query builder to exclude the "id" (primary key) attribute from the Reservations table? 
I know I can solve this using raw query.

garethflowers:
When grouping a model that links through a `belongsTo` association the `primaryKey` is no longer automatically added to the attributes list.

### Pull Request check-list

### Description of change

Previously when trying to group a model through a belongsTo association the model's primaryKey would automatically be inserted into the base models attributes list. This usually led to a:
```
SequelizeDatabaseError: column "id" must appear in the GROUP BY clause
```
This PR bypasses the logic for adding the `primaryKey` only if grouping a base model that links to another using a single associaton (i.e. a `belongsTo` or `hasOne`).

Should fix #3417.


sushantdhiman:
I am not familiar with `!options.hasSingleAssociation || options.hasMultiAssociation` so this PR will have to wait for @janmeier review

test/integration/model/findAll/group.test.js
```diff
@@ -51,6 +51,49 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+        }).then(() => {
+          return Comment.findAll({
+            attributes: ['PostId', [Sequelize.fn('COUNT', Sequelize.col('Comment.id')), 'comment_count']],
+            include: [
+              { model: Post, attributes: [] }
+            ],
+            group: ['PostId']
+          });
+        }).then(posts => {
+          expect(posts[0].get().hasOwnProperty('id')).to.not.be.ok;
```
janmeier: `to.equal(false)` - I know this style is already used a lot of places in the codebase, but it's prone to uncaught typos

janmeier:
👍

PR #8166, opened 2017-08-22
fix(instance): toJSON returns internal object, instead of a copy

nemisj:
Whenever `toJSON()` or `get()` is used and returned result is modified, it breaks the internal state of the
instance

### Pull Request check-list

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3143,7 +3143,7 @@ class Model {
       }
       return values;
     }
-    return this.dataValues;
+    return Utils._.clone(this.dataValues);
   }
 
   /**
--- a/test/unit/instance/to-json.test.js
+++ b/test/unit/instance/to-json.test.js
@@ -1,0 +1,26 @@
+'use strict';
+
+const chai = require('chai'),
+  expect = chai.expect,
+  Support   = require(__dirname + '/../support'),
+  DataTypes = require(__dirname + '/../../../lib/data-types'),
+  current   = Support.sequelize;
+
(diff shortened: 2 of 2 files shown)


test/unit/instance/to-json.test.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+const chai = require('chai'),
+  sinon = require('sinon'),
```
sushantdhiman: CI fails because of unused `sinon`, please fix it

lib/model.js
```diff
@@ -3143,7 +3143,7 @@ class Model {
       }
       return values;
     }
-    return this.dataValues;
+    return Utils._.clone(this.dataValues);
```
sushantdhiman: No, this will break things, you should be able to change datavalues and hit update and expect fields to be changed.

May be clone output of `toJSON`

nemisj: This is also what I did in the first place, though I thought it is a bit odd decision to expose internal data structure through the public API which says `Convert the instance to a JSON representation`.

sushantdhiman: Right, I meant `toJSON` should return cloned data. Changing JSON representation should not change internal model stores.

nemisj: okey, got it, thanks :) will move `clone` into `toJSON`

nemisj: @sushantdhiman it's now a different PR, since github doesn't allow to reopen after force push :( 
https://github.com/sequelize/sequelize/pull/8171

PR #8174, opened 2017-08-23
fix(abstract-query): Hashing of primaryKeys works incorrectly in _groupJoinData

nemisj:
Whenever `_groupJoinData` tries to group data into nested objects and primaryKey of nested objects or
parent objects is a Buffer, hashes might collide.

### Pull Request check-list


lib/dialects/abstract/query.js
```diff
@@ -478,6 +478,7 @@ class AbstractQuery {
 
       return uniqueKeyAttributes;
     };
+    const stringify = obj => Buffer.isBuffer(obj) ? obj.toString('hex') : String(obj);
```
mickhansen: Did you benchmark this? This section of the code is highly performance sensitive.

nemisj: I haven't, but even if I would, I do not know different approach to solve this. Can you help?

mickhansen: Something might be faster than than Buffer.isBuffer and i would definitely not waste time casting everything else to strings

mickhansen: The method is likely so small it will be inlined, but it still does mean a Buffer.isBuffer call on each row (of which there can be a million easily)

nemisj: Check, i will remove casting to string everything else.  Is there anything in place for sequelize already?

nemisj: I've run simple benchmark of `isBuffer` and `instance of Buffer` locally a couple of times, using benchmark npm package:

```
Buffer.isBuffer x 53,576,349 ops/sec ±2.07% (87 runs sampled)
instanceof Buffer x 81,705,405 ops/sec ±2.34% (86 runs sampled)
```
It appears that instanceof is faster. Is this enough or do you want me to do some other benchmarking?

PR #8204, opened 2017-08-26
fix: sqlite find out deleteAt items problem

ghostpanda:
sqlite 1s 内可以查询已经删除的数据
`deleted_at` > CURRENT_TIMESTAMP 
`deleted_at` > 2017-08-26 16:36:19

`deleted_at` > now
`deleted_at` > 2017-08-26 16:36:19.836 +00:00

### Pull Request check-list

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -76,8 +76,9 @@ class Model {
     const deletedAtAttribute = model.rawAttributes[deletedAtCol];
     const deletedAtObject = {};
     let deletedAtDefaultValue = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
+    const now = Utils.now(this.sequelize.options.dialect);
 
-    deletedAtDefaultValue = deletedAtDefaultValue || { $or: { $gt: model.sequelize.literal('CURRENT_TIMESTAMP'), $eq: null } };
+    deletedAtDefaultValue = deletedAtDefaultValue || { $or: { $gt: now, $eq: null } };
 
     deletedAtObject[deletedAtAttribute.field || deletedAtCol] = deletedAtDefaultValue;
 


apricote:
There is already a testcase for this scenario at https://github.com/sequelize/sequelize/blob/master/test/integration/instance.test.js#L1685

The reason why it did not fail is the use of the Sinon fake timer. Compare the generated queries:

`sequelize:master`:
```
Executing (default): UPDATE `ParanoidUsers` SET `deletedAt`='1970-01-01 00:00:00.000 +00:00' WHERE `id` = 1 AND `deletedAt` IS NULL
Executing (default): SELECT `id`, `username`, `createdAt`, `updatedAt`, `deletedAt`, `ParanoidUserId` FROM `ParanoidUsers` AS `ParanoidUser` WHERE (`ParanoidUser`.`deletedAt` > CURRENT_TIMESTAMP OR `ParanoidUser`.`deletedAt` IS NULL);
```

`ghostpanda:master`:
```
Executing (default): UPDATE `ParanoidUsers` SET `deletedAt`='1970-01-01 00:00:00.000 +00:00' WHERE `id` = 1 AND `deletedAt` IS NULL
Executing (default): SELECT `id`, `username`, `createdAt`, `updatedAt`, `deletedAt`, `ParanoidUserId` FROM `ParanoidUsers` AS `ParanoidUser` WHERE (`ParanoidUser`.`deletedAt` > '1970-01-01 00:00:00.000 +00:00' OR `ParanoidUser`.`deletedAt` IS NULL);
```

As the Date returned by the fake timer is so far in the past, a small difference in the timing of the sqlite db and node does not matter here.

Is it possible to avoid the fake timer for this one test case?

janmeier:
@apricote Calling `this.clock.restore()` before the start of that test should work. Or if you like figure out which tests actually require fake timers and only use it in those test

PR #8207, opened 2017-08-27
feat(model.init): define index with object

dfadev:
### Pull Request check-list

### Description of change

Added support for optionally defining indexes with an Object literal.

### DIFF
--- a/docs/models-definition.md
+++ b/docs/models-definition.md
@@ -758,6 +758,44 @@ sequelize.define('user', {}, {
 })
 ```
 
+Sequelize also supports defining each index with an object literal:
+
+```js
+sequelize.define('user', {}, {
+  indexes: {
+
+    // Create a unique index on poem
+    user_poem: {
+      unique: true,
+      fields: ['poem']
+    },
+
+    // Creates a gin index on data with the jsonb_path_ops operator
+    user_data: {
+      fields: ['data'],
+      using: 'gin',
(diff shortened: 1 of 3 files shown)


sushantdhiman:
I dont think there is any need of expanding API when you can achieve this by 

```js
sequelize.define('user', {}, {
  indexes: [
    // Create a unique index on poem
    {
      unique: true,
      fields: ['poem']
    },

    // Creates a gin index on data with the jsonb_path_ops operator
    {
      fields: ['data'],
      using: 'gin',
      operator: 'jsonb_path_ops'
    },

    // By default index name will be [table]_[fields]
    // Creates a multi column partial index
    {
      name: 'public_by_author',
      fields: ['author', 'status'],
      where: {
        status: 'public'
      }
    },

    // A BTREE index with a ordered field
    {
      name: 'title_index',
      method: 'BTREE',
      fields: ['author', {attribute: 'title', collate: 'en_US', order: 'DESC', length: 5}]
    }
  ]
})
```

Whats wrong with using already available array API for indexes ?

dfadev:
I don't think there's anything wrong with the existing method and I see no pressing need to expand the API.

It tripped me up because index definitions are the only option defined using an array.

sushantdhiman:
I rather not expand API unless we really need to, and this can be achieved with existing API, so closing :)

Issue #8192, opened 2017-08-24
Infinite range bounds are not handled correctly


yazshel:
## What you are doing?

I'm wanting to filter a table on a Sequelize.Range(Sequelize.Date) with an infinite bound, but it appears that the SQL generator tries to convert the "null" range bound to a date, prior to building the range literal.

```js
const Sequelize = require('sequelize');

const sequelize = new Sequelize('testdb', '', '', { dialect: 'postgres' });

class Test extends Sequelize.Model { }
Test.init({
  range: Sequelize.DataTypes.RANGE(Sequelize.DataTypes.DATEONLY),
}, { sequelize });

Test.findAll({ where: { range: { $overlap: [new Date(), null] } } })
    .then(() => console.log('Done'));
```

## What do you expect to happen?
The infinte
```sql
SELECT "id", "range", "createdAt", "updatedAt" FROM "Tests" AS "Test" WHERE "Test"."range" && '["2017-08-25",)';
```

## What is actually happening?
A quick look at `RANGE.prototype._stringify()` in `lib/dialects/postgres/data-types.js` the `null` range bound is incorrectly passed to the subtype `stringify()` function BEFORE it's been checked for `null` or `Infinity/-Infinity`. For `Sequelize.RANGE(Sequelize.DATE)` ranges, this results in  the `null` range bound being stringified to `"Invalid Date"`, which ends up being used in the range literal in the generated SQL:
```sql
SELECT "id", "range", "createdAt", "updatedAt" FROM "Tests" AS "Test" WHERE "Test"."range" && '["2017-08-25","Invalid date")';
```

__Dialect:__ postgres
__Database version:__  9.6.4
__Sequelize version:__ 4.7.2

Issue #8, opened 2010-08-21
npm install is broken


jneen:
details here: http://gist.github.com/542626

I believe this just involves updating your package.json to provide a "main" lib.  i.e.

"main": "sequelize.js"

PR #8210, opened 2017-08-27
fix(datatypes): don't stringify range bounds when infinite (fixes: #8…

yazshel:
…192)

### Pull Request check-list

- [YES] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [YES] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [YES] Have you added new tests to prevent regressions?
- [YES] Did you follow the commit message conventions explained in CONTRIBUTING.md?

### Description of change

Fix range bound processing logic in `RANGE.prototype.stringify` to pass through range bounds of `null` or `-Infinity`/`Infinity` unchanged (ie. dont' pass these values to the range subtype's stringify function). This allows `RANGE.stringify` to correctly identify these as infinite bounds and format the range SQL output correctly.
Fixes: #8192

### DIFF
--- a/lib/dialects/postgres/data-types.js
+++ b/lib/dialects/postgres/data-types.js
@@ -422,7 +422,10 @@ module.exports = BaseTypes => {
         this.toCastType();
     }
     const valuesStringified = values.map(value => {
-      if (this.options.subtype.stringify) {
+      if (_.includes([null, -Infinity, Infinity], value)) {
+        // Pass through "unbounded" bounds unchanged
+        return value;
+      } else if (this.options.subtype.stringify) {
         return this.options.subtype.stringify(value, options);
       } else {
         return options.escape(value);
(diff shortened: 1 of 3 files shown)


lib/dialects/postgres/data-types.js
```diff
@@ -422,7 +422,10 @@ module.exports = BaseTypes => {
         this.toCastType();
     }
     const valuesStringified = values.map(value => {
-      if (this.options.subtype.stringify) {
+      if (value === null || value === -Infinity || value === Infinity) {
```
sushantdhiman: use `includes` here

```js
if ([null, -Infinity, Infinity].includes(value)) {
```

May be Node 4 doesnt supports it, if it doesn't then https://lodash.com/docs/#includes

yazshel:
I've used `_.includes()`. (According to [node.green](http://node.green/#ES2016-features-Array-prototype-includes), `Array.prototype.includes` isn't supported by Node 4)

yazshel:
I'm not sure why that AppVeyor build failed with this push - it's indicating there may be a memory leak but I don't see how this could have been introduced as I've only changed the one line:
```javascript
      if (value === null || value === -Infinity || value === Infinity) {
```
to
```javascript
       if (_.includes([null, -Infinity, Infinity], value)) {
```
? Is there an intermittent issue with another component here?

sushantdhiman:
> Is there an intermittent issue with the MSSQL driver

Yes, I have restarted the build

PR #8227, opened 2017-08-29
fix(association): use `field` value in `count`

### LINKED ISSUE #8226: `Model.countAssoc()` method doesn't take renamed fields into account
## What you are doing?

```js
class User extends Model {
    userId: {
        /* ... */
        field: 'user_id'
    }
    /* ... */
}

class Project extends Model {
    projectId: {
        /* ... */
        field: 'project_id'
    }
}

User.belongsToMany(Project);

User.findById(1)
    .then(u => u.countProjects())
    .catch(console.log);

// Logs "Column User.userId does not exist"
```

See [here](https://github.com/sequelize/sequelize/blob/master/lib/associations/belongs-to-many.js#L447) -- `primaryKeyAttribute` ignores the `field` option on the attribute

`count` should replace the `primaryKeyAttribute` on the model with whatever value is in the `field` property for the definition of that attribute

_But the output was bar!_

_Output, either JSON or SQL_

Generated SQL appears to be something like this:

```
SELECT COUNT("User"."userId") AS "count" FROM "users" AS "User" INNER JOIN "UserProject" AS "UserProject" ON "Project"."project_id" = "UserProject"."project_id" AND "user"."user_id" = 1
```

__Dialect:__ postgres
__Database version:__ 9.5
__Sequelize version:__ 4.3.1

csreid:
### Pull Request check-list

### Description of change

Closes #8226

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -442,9 +442,11 @@ class BelongsToMany extends Association {
     const model = association.target;
     const sequelize = model.sequelize;
 
+    const primaryKeyField = model.attributes[model.primaryKeyAttribute].field || model.primaryKeyAttribute;
+
     options = Utils.cloneDeep(options);
     options.attributes = [
-      [sequelize.fn('COUNT', sequelize.col([association.target.name, model.primaryKeyAttribute].join('.'))), 'count']
+      [sequelize.fn('COUNT', sequelize.col([association.target.name, primaryKeyField].join('.'))), 'count']
     ];
     options.joinTableAttributes = [];
     options.raw = true;
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -442,11 +442,9 @@ class BelongsToMany extends Association {
     const model = association.target;
     const sequelize = model.sequelize;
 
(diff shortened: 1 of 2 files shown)


test/integration/model/attributes/field.test.js
```diff
@@ -606,6 +620,14 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           });
         });
       });
+
+      it('should work with association `count`', function() {
```
sushantdhiman: belongs to many association

lib/associations/belongs-to-many.js
```diff
@@ -442,9 +442,11 @@ class BelongsToMany extends Association {
     const model = association.target;
     const sequelize = model.sequelize;
 
+    const primaryKeyField = model.attributes[model.primaryKeyAttribute].field || model.primaryKeyAttribute;
+
     options = Utils.cloneDeep(options);
     options.attributes = [
-      [sequelize.fn('COUNT', sequelize.col([association.target.name, model.primaryKeyAttribute].join('.'))), 'count']
+      [sequelize.fn('COUNT', sequelize.col([association.target.name, primaryKeyField].join('.'))), 'count']
```
sushantdhiman: use `model.primaryKeyField` instead

csreid: Not sure how I missed that property :disappointed:

PR #8240, opened 2017-09-01
Secure operators

### LINKED ISSUE #7310: Hash injection (security)
Using specially crafted requests we can trivially bypass secret_token protections on websites using sequalize. 

Many people have code like this

```
db.Token.findOne({
      where: {
        token: req.query.token
      }
);
```

But Node.js and other platforms allow nested parameters, ie token[$gt]=1 will turn into token = {"$gt":1}. When we pass such hash to sequalize it will consider it a query (greater than 1) and find the first token in DB, bypassing security of this endpoint. This behavior was copied from Mongo https://docs.mongodb.com/manual/reference/operator/query/

Using finely tuned $gt we can iterate all tokens in db and impersonate every single user. 

There are vulnerable sites in the wild.

That's how it can be exploited in Mongo http://blog.websecurify.com/2014/08/hacking-nodejs-and-mongodb.html and https://cirw.in/blog/hash-injection

My advice would be to either disable this functionality entirely (i.e. require passing token {"$eq":token} every time) or make sure the parameter isn't coming from req.query and is native hash object.

yonjah:
### Pull Request check-list

### Description of change
This PR will convert all opeators to be used internally as Symbols.
This will reduce the possibility of operators injection as described in #7310 

For backward compatibility all previous operators are available as aliases, but user are discourage from using those aliases and a deprecation warning will be logged unless user explicitly set her own aliases as described in the updated documentation

this PR also include changes from #8068 since it would have been impossible to implement without those changes

Fixes #7310


lib/sequelize.js
```diff
@@ -18,6 +18,7 @@ const Hooks = require('./hooks');
 const Association = require('./associations/index');
 const Validator = require('./utils/validator-extras').validator;
 const _ = require('lodash');
+const Operators = require('./operators');
```
sushantdhiman: Call it `Op`

lib/sequelize.js
```diff
@@ -1152,6 +1162,13 @@ Sequelize.prototype.Promise = Sequelize.Promise = Promise;
  */
 Sequelize.prototype.QueryTypes = Sequelize.QueryTypes = QueryTypes;
 
+
+/**
+ * Operators symbols to be used for querying data
+ * @see  {@link Operators}
+ */
+Sequelize.prototype.Operators = Sequelize.Operators = Operators;
```
sushantdhiman: May be expose as `Sequelize.prototype.Op` , cc @mickhansen @janmeier @eseliger

lib/sequelize.js
```diff
@@ -227,6 +231,12 @@ class Sequelize {
 
     this.dialect = new Dialect(this);
     this.dialect.QueryGenerator.typeValidation = options.typeValidation;
+    if (this.options.operatorsAliases) {
+      this.dialect.QueryGenerator.setOperatorsAliases(this.options.operatorsAliases);
+    } else {
+      Utils.deprecate('Limiting available operators aliases offers improved security and will be forced in future versions. check: http://docs.sequelizejs.com/manual/tutorial/querying.html#operators');
```
sushantdhiman: This should only be shown for FIRST time when user is actually trying to use legacy alias in code, otherwise I will get annoyed even if I am not using string based operators.

OR may be its missing `else if ( this.options.allowLegacyOperators )` ????

Message should say `String based operators are now deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/tutorial/querying.html#operators`

lib/utils.js
```diff
@@ -515,4 +515,41 @@ exports.mapIsolationLevelStringToTedious = (isolationLevel, tedious) => {
     case 'SNAPSHOT':
       return tediousIsolationLevel.SNAPSHOT;
   }
-};
\ No newline at end of file
+};
+
+
```
sushantdhiman: Extra lines

test/unit/dialects/abstract/query-generator.test.js
```diff
@@ -0,0 +1,107 @@
+'use strict';
+
+const chai = require('chai'),
+  expect = chai.expect,
+  Op = require('../../../../lib/operators'),
+  QueryGenerator = require('../../../../lib/dialects/abstract/query-generator');
+
+require(__dirname + '/../../support');
```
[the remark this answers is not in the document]
yonjah: `support` injects sequelize into the test context. it won't cause any issues if your running thw whole suit but if you only want to run this test file you'll need it to be loaded for the test to work

test/unit/dialects/abstract/query-generator.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+function quoteIdentifier(identifier) {
+  return identifier;
+}
+
+function getQueryGenerator(sequelize) {
```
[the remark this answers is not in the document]
yonjah: All the other tests for the dialects query-generator loads them manually.
I think it also makes more sense in this context of a unit test for the abstruct query generator to get a clean and base version of query-generator since the one sequelize is using changes depending on the configuration it was loaded with and the dialect

test/unit/dialects/abstract/query-generator.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+      QG.whereItemQuery(Op.and, [{test: {[Op.between]: [2, 5]}}, {test: {[Op.ne]: 3}}, {test: {[Op.not]: 4}}])
+        .should.be.equal('(test BETWEEN 2 AND 5 AND test != 3 AND test != 4)');
+
+    });
+
+    it('should not parse any strings as aliases  operators', function() {
+      const QG = getQueryGenerator(this.sequelize);
+      expect(() => QG.whereItemQuery('$or', [{test: 5}, {test: 3}]))
+        .to.throw();
```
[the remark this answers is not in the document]
yonjah: I'm getting an error for one of the internal packages expecting a string and getting an object.
I can specify the error but I'm not sure if it really matters for the test.

test/unit/sql/add-constraint.test.js
```diff
@@ -68,7 +68,10 @@ if (current.dialect.supports.constraints.addConstraint) {
               }
             }
           }), {
-            default: 'ALTER TABLE [myTable] ADD CONSTRAINT [check_mycolumn_where] CHECK (([myColumn] > 50 AND [myColumn] < 100));'
+            default: [
```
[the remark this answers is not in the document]
yonjah: I'm not really sure why it changes or if it is caused by the code changes. 99% of the time I'm getting the original result but some time they get flipped.
In general JS asks you not to rely on the order when traversing object properties. I never actually had a case where they weren't traverse in the same order but I'm not sure if I ever had a code where it would matter.

lib/dialects/abstract/query-generator.js
```diff
@@ -11,6 +11,7 @@ const Association = require('../../associations/base');
 const BelongsTo = require('../../associations/belongs-to');
```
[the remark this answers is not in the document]
yonjah: I can tried to give them the best name as I could figure out from the logic as I broke it for smaller methods. But to be honest other than understanding the actual logic I don't know a lot of the reasoning behind this code. there were a few places I had to chase down some bugs (like with the JSON traversal discrepancies  between sqlite and postgre ) so I got to find the issues and the original pull request for the changes. but for most of the code their I don't know the actual reasoning behind it other than the bare logic (which I hope is a lot easier to understand now).
maybe @mickhansen can give a bit more info where it's might be beneficial

lib/sequelize.js
```diff
@@ -163,6 +166,7 @@ class Sequelize {
       isolationLevel: null,
       databaseVersion: 0,
       typeValidation: false,
+      allowLegacyOperators: true,
```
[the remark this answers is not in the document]
yonjah: Sorry I ended up not using this option since I think it's better to just select the exact aliases you want. I'll remove it

lib/sequelize.js
yonjah: The problem is if you don't select your own aliases (or completely disable them) they are available (for backward compatibility) so most users are not even aware of the aliases available and are not using many of them (legacy ones weren't even documented) but they can still be injected even if they are not normally used.
I think showing this warning unless the user sets his own operators using the `operatorsAliases` is the best option to make sure users are aware of the issue.
I tried to make it clear in the documentation how to use `operatorsAliases` and dismiss this warning but maybe it can be made clearer

lib/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
sushantdhiman: Still try your best or ignore where you cant. A few hints are better than having scattered function with no comments.

test/unit/dialects/abstract/query-generator.test.js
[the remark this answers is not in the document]
sushantdhiman: this method could be failing for any reason, we cant be ok with that it throws something. Error should be deterministic and may be specific when trying to use String alias.

test/unit/sql/add-constraint.test.js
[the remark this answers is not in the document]
sushantdhiman: It make testing things a bit hard, may be we should sort object keys to have single defined output for same arguments

test/unit/dialects/abstract/query-generator.test.js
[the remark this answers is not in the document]
sushantdhiman: I have two suggestions here,

1) Keep this method here, but get QueryGenerator, directly with `sequelize.dialect.QueryGenerator`, no assign
2) Move this method to support lib, so others can use it as well

lib/sequelize.js
sushantdhiman: I liked the idea of default enabled flag, which will all load legacy alias, so this wont be a breaking change.

If they want better security and no deprecation warning they should just set that flag to `false`, Please remove user supplied `operatorsAliases` altogether.

test/unit/dialects/abstract/query-generator.test.js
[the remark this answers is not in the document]
yonjah: Should I change it in all the other tests for each dialect query generator ?

test/unit/dialects/abstract/query-generator.test.js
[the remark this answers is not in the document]
yonjah: I can check for the specific error I'm getting.
It is hard to detect where a string alias are being used cause in this context there aren't any string aliases (otherwise the query would work) and there might even be cases where a valid query will be generated but it will not use the operator (it will be parsed as a regular field)

test/unit/sql/add-constraint.test.js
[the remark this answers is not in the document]
yonjah: I can look into it

lib/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
yonjah: Ok I'll have another look and try my best

lib/sequelize.js
yonjah: That's how I tried to implement it the first time.
But there are a few issues with doing it that way - 

To get a fully backward supported and secure options you need two flags one for legacy aliases (which are not prefixed undocumented and probably 99% of the users we have are not using them and so hopefully would disable them) and another for the $ prefixed operators which all users are probably using and we should hope they'll switch to symbols after this is released.

since we have two flags we we'll need two warnings -
1. Deprecation warning if legacy operators are enabled (even explicitly) since they will be removed any way in future versions
2. Deprecation warning for string operators not being explicitly enabled (I assume those will still be available in future version but won't be enabled by default)

Now I'm ending up setting two flags to hide all deprecation warnings of a single feature.
Users are not really aware of what they are enabling or disabling. 
They might only '..' alias and since it would fail enable legacy operators but then they'll be forced to also enable `@>`, `gt` and all the other legacy one.

Giving the the `operatorsAliases` takes control back to the user. we don't want to set those aliases for you cause you need to be aware for them. you want '..' to equal the between operator sure you can do it. want '||' to equal the or operator yea sure. But it is up to you to define those and up to you to sanitize them. Now every user knows exactly which strings will be translated into operators cause he is the one explicitly defined them.

So if my API allows client to send requests including objects using the `$not` alias I can enable it without having to also enable the `$col` alias

test/unit/dialects/abstract/query-generator.test.js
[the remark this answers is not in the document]
sushantdhiman: No just this one is enough

docs/models-usage.md
```diff
@@ -98,11 +98,12 @@ The success handler will always receive an object with two properties:
 * `count` - an integer, total number records matching the where clause
 * `rows` - an array of objects, the records matching the where clause, within the limit and offset range
 ```js
+const {like} = Sequelize.Op;
```
sushantdhiman: May be update it as well, it doesn't work with our base version Node v4

sushantdhiman: Please update other examples where you have used destructing

docs/querying.md
```diff
@@ -160,54 +165,139 @@ as well.
...
+currently the following legacy aliases are also set but are planned to be fully removed in the near future -
+ne, not, in, notIn, gte, gt, lte, lt, like, ilike, $ilike, nlike, $notlike, notilike, .., between, !.., notbetween, nbetween, overlap, &&, @>, <@
+
+Due to the security improvements it is highly advised to only use the specific aliases you need.
+
+```js
+const Op = Sequelize.Op;
+
+//use sequelize without any operators aliases
+const connection = new Sequelize(db, user, pass, { operatorsAliases: {} });
```
sushantdhiman: We dont even need to pass `operatorAliases`, right ?

docs/querying.md
```diff
@@ -160,54 +165,139 @@ as well.
...
+
+//use sequelize with only alias for $and => Op.and
+const connection2 = new Sequelize(db, user, pass, { operatorsAliases: { $and: Op.and } });
+```
+
+Sequelize will warn you if your using the default aliases and not limiting them
+if you want to keep using all default aliases (excluding legacy ones) without the warning you can pass the following operatorsAliases option -
```
sushantdhiman: I am not ok with this, alias like `$gte` should always be available without user doing anything or passing anything.

String based alias on other hand should be available for bc, but can be turned off by flag

lib/operators.js
```diff
@@ -0,0 +1,114 @@
...
+  'notbetween': Op.notBetween,
+  'nbetween': Op.notBetween,
+  'overlap': Op.overlap,
+  '&&': Op.overlap,
+  '@>': Op.contains,
+  '<@': Op.contained
+};
+
+Op.Aliases = Aliases;
+Op.LegacyAliases = Object.assign({}, LegacyAliases, Aliases);
```
sushantdhiman: Why `LegacyAliases` are merged with `Aliases`, both arent same thing

docs/querying.md
yonjah: This specific example shows how to remove all string aliases.
so after initializing Sequelize in this way you wont be able to use even '$gt' or any other alias minimizing injection surface.
I guess I need to go over the documentation and see that this point is clear

docs/querying.md
yonjah: The thing is '$gte' is still string based.
all aliases are available by default. but if you don't actually select the ones you need you'll see a warning. but everything will work as before.

There is an easier way to set all the '$' string aliases since Op actually exposes the setting. 
But I thought it would be a better practice for the user to set them and so know which ones he is actually enabling (If they really lazy they'll probably just copy past the aliases from this docs).
Other than kinda forcing the users to be aware of the aliases it also gives a bit easier eay to introduce future upgrades, if in a future version we want to completely remove or change the default aliases users who follow the code here will still get the same aliases they had has before with no breaking changes. So I think its a bit more future proof than relying on Sequelize setting it up for them

yonjah:
@sushantdhiman all the refactor commits are available in #8068 
Unfortunately discussion there went stale and there was very high chance of introducing critical bugs when changing the operators without those changes

sushantdhiman:
Please rebase @yonjah

yonjah:
@sushantdhiman Any thing else regarding this pull request ?
Did you want me to rebase it to a single commit ?

sushantdhiman:
> Did you want me to rebase it to a single commit ?

No need, Github has `Squash and merge` button

test/unit/sql/add-constraint.test.js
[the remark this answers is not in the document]
sushantdhiman: @yonjah Did you figure out why this is an issue, it would be great if we can get some deterministic behavior here

docs/querying.md
sushantdhiman: May be just call it `operators`, setting it false or empty (can be checked with lodash) should disable legacy / string op

lib/sequelize.js
```diff
@@ -228,6 +231,12 @@ class Sequelize {
 
     this.dialect = new Dialect(this);
     this.dialect.QueryGenerator.typeValidation = options.typeValidation;
+    if (this.options.operatorsAliases) {
```
sushantdhiman: As I said above this should be renamed to just `operators`

sushantdhiman: 1. If operators is set to `true`, which will be default value, load all legacy operators
2. If operators is set to an map, set only those operators
3. if operators is set to false or empty value, disable legacy op completely

(1) Will give deprecation warning its also the default state when we release this feature
(2) and (3) will give no warning

Warning should say `String based operators are now deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/tutorial/querying.html#operators`

sushantdhiman:
@yonjah I have delete a few comments which I think were resolved to have some less clutter :)

test/unit/sql/add-constraint.test.js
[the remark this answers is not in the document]
yonjah: From what I could see I only get random results on node v4 and it is due to the fact that the alias method delete the previous keys -
https://github.com/yonjah/sequelize/blob/fb8188efb78eb21f91f87cbe0693220adbdf7ee9/lib/dialects/abstract/query-generator.js#L2107

This is how the old aliasing worked but I can change the alias logic to generate a new object instead of deleting properties on the old one (which might be a better solution any way but a bit more complex)

docs/querying.md
```diff
@@ -160,54 +165,139 @@ as well.
...
+#### Operators security
+Using Sequelize without any aliases improves security.
+Some frameworks automatically parse user input into js objects and if you fail to sanitize your input it might be possible to inject an Object with string operators to Sequelize.
+
+Not having any string aliases will make it extremely unlikely that operators could be injected but you should always properly validate and sanitize user input.
+
+For backward compatibility reasons Sequelize sets the following aliases by default -
+$eq, $ne, $gte, $gt, $lte, $lt, $not, $in, $notIn, $is, $like, $notLike, $iLike, $notILike, $regexp, $notRegexp, $iRegexp, $notIRegexp, $between, $notBetween, $overlap, $contains, $contained, $adjacent, $strictLeft, $strictRight, $noExtendRight, $noExtendLeft, $and, $or, $any, $all, $values, $col
+
+currently the following legacy aliases are also set but are planned to be fully removed in the near future -
```
sushantdhiman: Currently , capitalize `C`

docs/querying.md
```diff
@@ -160,54 +165,139 @@ as well.
...
+ne, not, in, notIn, gte, gt, lte, lt, like, ilike, $ilike, nlike, $notlike, notilike, .., between, !.., notbetween, nbetween, overlap, &&, @>, <@
+
+Due to the security improvements it is highly advised to only use the specific aliases you need.
```
sushantdhiman: For better security it is highly advised to use `Sequelize.Op` and not depend on any string alias at all. You can limit alias your application will need by setting `operators` option, remember to sanitize user input especially when you are directly passing them to Sequelize methods.

lib/sequelize.js
```diff
@@ -94,6 +95,8 @@ class Sequelize {
    * @param {Array}    [options.retry.match] Only retry a query if the error matches one of these strings.
    * @param {Integer}  [options.retry.max] How many times a failing query is automatically retried.  Set to 0 to disable retrying on SQL_BUSY error.
    * @param {Boolean}  [options.typeValidation=false] Run built in type validators on insert and update, e.g. validate that arguments passed to integer fields are integer-like.
+   * @param {Object}   [options.operators] String aliases for operators if not set all aliases are enabled by default.
```
sushantdhiman: `* @param {Object|Boolean}   [options.operators=true] String based operator alias, default value is true which will enable all operators alias. Pass object to limit set of aliased operators or false to disable completely.`

@yonjah Now I think about it, may be `operatorAlias` is better name after all :)

lib/utils.js
```diff
@@ -565,3 +565,38 @@ exports.mapIsolationLevelStringToTedious = (isolationLevel, tedious) => {
...
+/**
+ * getOperators
+ * @param  {Object} obj
+ * @return {Array<Symbol>} All operators properties of obj
+ */
+function getOperators(obj) {
+  return _.intersection(Object.getOwnPropertySymbols(obj), operatorsArray);
+}
+
```
sushantdhiman: No extra line here

lib/utils.js
```diff
@@ -565,3 +565,38 @@ exports.mapIsolationLevelStringToTedious = (isolationLevel, tedious) => {
       return tediousIsolationLevel.SNAPSHOT;
   }
 };
+
+//Collection of helper methods to make it easier to work with symbol operators
+
+/**
+ * getOperators
+ * @param  {Object} obj
+ * @return {Array<Symbol>} All operators properties of obj
```
sushantdhiman: Add `@private` so we dont accidentally expose this in our docs

lib/utils.js
```diff
@@ -565,3 +565,38 @@ exports.mapIsolationLevelStringToTedious = (isolationLevel, tedious) => {
...
+
+/**
+ * getComplexSize
+ * @param  {Object|Array} obj
+ * @return {Integer}      Length of object properties including operators if obj is array returns its length
+ */
+function getComplexSize(obj) {
+  return Array.isArray(obj) ? obj.length : getComplexKeys(obj).length;
+}
+
```
sushantdhiman: No extra line here

lib/utils.js
```diff
@@ -565,3 +565,38 @@ exports.mapIsolationLevelStringToTedious = (isolationLevel, tedious) => {
...
+/**
+ * getComplexKeys
+ * @param  {Object} obj
+ * @return {Array<String|Symbol>} All keys including operators
+ */
+function getComplexKeys(obj) {
+  return getOperators(obj).concat(_.keys(obj));
+}
+
```
sushantdhiman: No extra line here

test/support.js
```diff
@@ -153,6 +155,14 @@ const Support = {
     }
   },
 
+  getAbstructQueryGenerator(sequelize) {
```
sushantdhiman: Abstract*

test/unit/dialects/abstract/query-generator.test.js
```diff
@@ -0,0 +1,95 @@
+'use strict';
+
+const chai = require('chai'),
+  expect = chai.expect,
+  Op = require('../../../../lib/operators'),
+  getAbstructQueryGenerator = require(__dirname + '/../../support').getAbstructQueryGenerator;
+
+
```
sushantdhiman: Get rid of extra lines, just one empty line is enough

test/unit/dialects/abstract/query-generator.test.js
```diff
@@ -0,0 +1,95 @@
...
+  describe('whereItemQuery', () => {
+    it('should generate correct query for Symbol operators', function() {
+
+      const QG = getAbstructQueryGenerator(this.sequelize);
+      QG.whereItemQuery(Op.or, [{test: {[Op.gt]: 5}}, {test: {[Op.lt]: 3}}, {test: {[Op.in]: [4]}}])
+        .should.be.equal('(test > 5 OR test < 3 OR test IN (4))');
+
+      QG.whereItemQuery(Op.and, [{test: {[Op.between]: [2, 5]}}, {test: {[Op.ne]: 3}}, {test: {[Op.not]: 4}}])
+        .should.be.equal('(test BETWEEN 2 AND 5 AND test != 3 AND test != 4)');
+
```
sushantdhiman: No empty line here

test/unit/dialects/abstract/query-generator.test.js
```diff
@@ -0,0 +1,95 @@
...
+  expect = chai.expect,
+  Op = require('../../../../lib/operators'),
+  getAbstructQueryGenerator = require(__dirname + '/../../support').getAbstructQueryGenerator;
+
+
+
+describe('QueryGenerator', () => {
+  describe('whereItemQuery', () => {
+    it('should generate correct query for Symbol operators', function() {
+
```
sushantdhiman: Remove this line

test/unit/dialects/abstract/query-generator.test.js
```diff
@@ -0,0 +1,95 @@
...
+      expect(() => QG.whereItemQuery('test', {$ne: 3}))
+        .to.throw('Invalid value [object Object]');
+
+      expect(() => QG.whereItemQuery('test', {$not: 3}))
+        .to.throw('Invalid value [object Object]');
+
+      expect(() => QG.whereItemQuery('test', {$in: [4]}))
+        .to.throw('Invalid value [object Object]');
+
```
sushantdhiman: Remove this line

mkaufmaner:
I must say that I am a huge proponent of this PR.  It will also bring us one step closer to having expressive operators.

lib/sequelize.js
yonjah: @sushantdhiman singular `operatorAlias` or `operatorsAliases` ?

sushantdhiman: @yonjah  `operatorsAliases`

docs/querying.md
sushantdhiman: This example can be updated to `operatorsAliases: false`, right ?

yonjah: Yea sorry missed it when changed it back from `operators`

lib/sequelize.js
```diff
@@ -228,6 +232,12 @@ class Sequelize {
 
     this.dialect = new Dialect(this);
     this.dialect.QueryGenerator.typeValidation = options.typeValidation;
+    if (this.options.operatorsAliases === true) {
+      Utils.deprecate('String based operators are now deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/tutorial/querying.html#operators');
```
mkaufmaner: One second thought, I don't think we should remove complete support for operator aliases.

Simply put, I still want to be able to pass in a JSON object from an HTTP request into the `where` property.

eseliger: I think we should disable them by default in the future, but should not completely remove them, there are definitely use-cases for that, I think

lib/utils.js
```diff
@@ -187,12 +189,9 @@ function mapOptionFieldNames(options, Model) {
 exports.mapOptionFieldNames = mapOptionFieldNames;
 
 function mapWhereFieldNames(attributes, Model) {
-  let attribute;
-  let rawAttribute;
-
   if (attributes) {
-    for (attribute in attributes) {
-      rawAttribute = Model.rawAttributes[attribute];
+    getComplexKeys(attributes).forEach(attribute => {
```
eseliger: can't we use a `for .. of` loop here? `.foreach` is so 2012

test/unit/sql/where.test.js
```diff
@@ -267,13 +267,13 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
 
       suite('$and', () => {
         testsql('$and', {
-          shared: 1,
           $or: {
             group_id: 1,
             user_id: 2
-          }
+          },
+          shared: 1,
```
eseliger: trailing comma can be removed :)

lib/sequelize.js
yonjah: Yea I agree.
I don't think there is any thought of deprecating them but users should define the ones they want to use and not have all of them available by default.

In the future it might be useful to define aliases in the query options. So for example if you have one query that you know you gonna pass right through and you want to allow for 'like' operator you can just add an alias for it that it will only be available for that specific query

lib/utils.js
sushantdhiman: No need, JS itself is more than 20 years old :) forEach looks clean here.

sushantdhiman:
Wanted to have a review by @mickhansen  / @janmeier for deprecating API but I think its ok.

mickhansen:
@sushantdhiman I like it, only thing i would change is a bit of interface/documentation, `Op` with uppercase is a bit annoying so maybe `import {operator /* as op /*} from 'sequelize'`

PR #8247, opened 2017-09-03
fix(sequelize/query): release connection after retry

lext-7:
### Pull Request check-list

### Description of change

Connections should be release after retry rather than being released after every query.
It will cause `'Resource not currently part of this pool'` error if it's released after every query.

In v3, sequelize use `coopernurse/node-pool@2.4.8` which won't raise a error if a connection is released multiple times. See here [https://github.com/coopernurse/node-pool/blob/v2.4/lib/generic-pool.js#L420](https://github.com/coopernurse/node-pool/blob/v2.4/lib/generic-pool.js#L420)

But in v4, sequelize use new version of `coopernurse/node-pool` which will return a rejected promise if a connection is released multiple times. See here [https://github.com/coopernurse/node-pool/blob/master/lib/Pool.js#L422](https://github.com/coopernurse/node-pool/blob/master/lib/Pool.js#L422).

Because of this error, if you are using retry, you will get `'Resource not currently part of this pool'` when got every sql error. You can not get a normal sql error which shows you what is wrong.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -529,11 +529,12 @@ class Sequelize {
     }).then(connection => {
       const query = new this.dialect.Query(connection, this, options);
 
-      return retry(() => query.run(sql, bindParameters).finally(() => {
-        if (!options.transaction) {
-          return this.connectionManager.releaseConnection(connection);
-        }
-      }), Utils._.extend(this.options.retry, options.retry || {}));
+      return retry(() => query.run(sql, bindParameters), Utils._.extend(this.options.retry, options.retry || {}))
+        .finally(() => {
+          if (!options.transaction) {
+            return this.connectionManager.releaseConnection(connection);
+          }
+        });
     }).finally(() => {
       if (this.test._trackRunningQueries) {
         this.test._runningQueries--;
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -528,8 +528,9 @@ class Sequelize {
       return options.transaction ? options.transaction.connection : this.connectionManager.getConnection(options);
     }).then(connection => {
       const query = new this.dialect.Query(connection, this, options);
+      const retryOptions = Utils._.extend(this.options.retry, options.retry || {});
 
-      return retry(() => query.run(sql, bindParameters), Utils._.extend(this.options.retry, options.retry || {}))
+      return retry(() => query.run(sql, bindParameters), retryOptions)
         .finally(() => {
           if (!options.transaction) {
             return this.connectionManager.releaseConnection(connection);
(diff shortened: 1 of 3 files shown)


lib/sequelize.js
```diff
@@ -529,11 +529,12 @@ class Sequelize {
     }).then(connection => {
       const query = new this.dialect.Query(connection, this, options);
 
-      return retry(() => query.run(sql, bindParameters).finally(() => {
-        if (!options.transaction) {
-          return this.connectionManager.releaseConnection(connection);
-        }
-      }), Utils._.extend(this.options.retry, options.retry || {}));
+      return retry(() => query.run(sql, bindParameters), Utils._.extend(this.options.retry, options.retry || {}))
+        .finally(() => {
```
sushantdhiman: You can move this to `finally` block below

lext-7: But `connection` is not defined in that `finally` block below.

sushantdhiman: Right :), then just add a unit test and it should be ready to go

test/integration/sequelize.test.js
```diff
@@ -728,6 +728,23 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+          // retry for all errors
+          match: null
+        }
+      }).catch(err => {
+        // If the connection is released multiple times,
+        // than an Error(not Sequelize BaseError) will be raised by generic-pool.
+        // If the connection is released only once,
+        // then a DatabaseError which inherits Sequelize BaseError will be raised,
+        // because this is a wrong sql.
+        expect(err).to.be.an.instanceof(this.sequelize.Error);
```
sushantdhiman: Add a spy on `this.sequelize.connectionManager.releaseConnection`, assert against number of calls

lib/sequelize.js
```diff
@@ -529,11 +529,12 @@ class Sequelize {
     }).then(connection => {
       const query = new this.dialect.Query(connection, this, options);
 
-      return retry(() => query.run(sql, bindParameters).finally(() => {
-        if (!options.transaction) {
-          return this.connectionManager.releaseConnection(connection);
-        }
-      }), Utils._.extend(this.options.retry, options.retry || {}));
+      return retry(() => query.run(sql, bindParameters), Utils._.extend(this.options.retry, options.retry || {}))
```
sushantdhiman: not easy to read, add a new `const retryOptions` along side `const query`

test/integration/sequelize.test.js
```diff
@@ -728,6 +728,23 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
       });
     });
 
+    it('connections should be released only once', function() {
```
sushantdhiman: `connection should be released only once when retry fails`

PR #8254, opened 2017-09-05
feat(statement_timeout): Support for PostgreSQL statement_timeout

jkennedy1980:
I just contributed `statement_timeout` support to `node-pg` in version `7.3.0`. This PR brings `statement_timeout` support to Sequelize.

Statement timeout allows you to set a maximum milliseconds for all queries executed on a connection. This is useful when running on Heroku because they enforce a maximum HTTP timeout of 30 seconds. If you have long running queries, postgres continues to execute them even though the HTTP connection is closed and no one will receive the response.


sushantdhiman:
Added this option to whitelist here https://github.com/sequelize/sequelize/pull/8342/files

Any pg@7.3 users can now use it.

Issue #8275, opened 2017-09-09
Nested include with limit and order on column with "field" value generates broken SQL


Alexsey:
## What you are doing?
It is a minimal code snippet I could create that reproduces the bug. You could copy->paste->execute it, just set your DB credentials

```js
'use strict'

const Sequelize = require('sequelize')

const dbConfig = {
    type: 'mysql', user: 'root', password: 'root',
    database: 'db', host: 'localhost'
  }

;(async () => {
  const sequelize = new Sequelize(
    dbConfig.database, dbConfig.user, dbConfig.password, {
      host: dbConfig.host,
      dialect: dbConfig.type,
      pool: {max: 100, min: 0, idle: 10000},
      dialectOptions: {
        charset: 'utf8_general_ci',
      }
    }
  )
  await sequelize.authenticate()

  // I have removed the timestamps to simplify the generated SQL
  const A = sequelize.define('A', {
    id: {type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true},
    bId: {type: Sequelize.INTEGER, field: 'b_id'} // <- defined with field. It is important part
  }, {timestamps: false})

  const B = sequelize.define('B', {
    id: {type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true},
  }, {timestamps: false})

  const C = sequelize.define('C', {
    id: {type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true},
    bId: {type: Sequelize.INTEGER}
  }, {timestamps: false})

  A.belongsTo(B, {as: 'B', foreignKey: 'bId'})

  B.hasMany(C, {as: 'Cs', foreignKey: 'bId'})

  await sequelize.sync({force: true})

  await A.create({
    B: {
      Cs: {

      }
    }
  }, {
    include: [{
      model: B,
      as: 'B',
      include: [{
        model: C,
        as: 'Cs'
      }]
    }]
  })

  const result = await A.findAll({
    order: [['bId', 'ASC']], // <- this parameter is the source of broken SQL
    limit: 1, // <- required for broken SQL to be generated
    include: [{
      model: B,
      as: 'B',
      include: [{ // <- required nested include of 1-n relation
        model: C,
        as: 'Cs'
      }]
    }]
  })

  console.dir(result[0].toJSON(), {depth: null})

  sequelize.close()
})()
```

## What do you expect to happen?
`.findAll` should be executed with no errors and in log should be (formatted):
```js
{
  id: 1, 
[...]

PR #8278, opened 2017-09-09
fix(abstract/querygenerator): remove duplicated order expression on outer query

Alexsey:
### Pull Request check-list

### Description of change

There is a bug https://github.com/sequelize/sequelize/issues/8275 and it may be fixed by just removing part of SQL statement that looks like redundant (at list for me)

The point of change is that now `order` statement is applying to both main query and sub query. It looks like there is no need for second `order`ing. Also this would also fix a mentioned bug

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1680,8 +1680,9 @@ const QueryGenerator = {
           && !(typeof order[0] === 'string' && model && model.associations !== undefined && model.associations[order[0]])
         ) {
           subQueryOrder.push(this.quote(order, model, '->'));
+        } else {
+          mainQueryOrder.push(this.quote(order, model, '->'));
         }
-        mainQueryOrder.push(this.quote(order, model, '->'));
       }
     } else if (options.order instanceof Utils.SequelizeMethod) {
       const sql = this.quote(options.order, model, '->');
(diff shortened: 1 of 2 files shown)


Alexsey:
Tests have pass ok but I think some one else should also think about is this part is really redundant. It so - I don't think there should any tests be added as far as it doesn't change the behavior. Just remove stuff that didn't work well

Alexsey:
Keep in mind that it could have potential conflicts with deep includes with right or outer joins. As far as I can see Sequelize is not using them for now. But if it would - ordering of the main query will get some sense:
```sql
SELECT ...
FROM (... ORDER BY number) as A
RIGHT JOIN B ON ...
```
would return a result that is ordered by `A.number` mixed with `NULL`

yonjah:
I've just noticed this change actually fails the tests when run on  my local instance with MariaDB (10.1.21).

The findCount at https://github.com/sequelize/sequelize/blob/master/test/integration/associations/belongs-to.test.js#L888 generates a query that does not actually follow the order -

```sql
SELECT `entry`.*, `mail->recipients`.`id` AS `mail.recipients.id`, `mail->recipients->MailRecipients`.`mailId` AS `mail.recipients.MailRecipients.mailId`, `mail->recipients->MailRecipients`.`recipientId` AS `mail.recipients.MailRecipients.recipientId` FROM (SELECT `entry`.`id`, `entry`.`ownerId`, `entry`.`mailId`, `mail`.`id` AS `mail.id` FROM `entries` AS `entry` INNER JOIN `mails` AS `mail` ON `entry`.`mailId` = `mail`.`id` WHERE ( SELECT `mail`.`id` FROM `mails` AS `mail` INNER JOIN ( `MailRecipients` AS `recipients->MailRecipients` INNER JOIN `users` AS `recipients` ON `recipients`.`id` = `recipients->MailRecipients`.`recipientId` AND `recipients->MailRecipients`.`recipientId` = 1) ON `mail`.`id` = `recipients->MailRecipients`.`mailId` WHERE `entry`.`mailId` = `mail`.`id` LIMIT 1 ) IS NOT NULL ORDER BY `entry`.`id` DESC LIMIT 0, 10) AS `entry` INNER JOIN ( `MailRecipients` AS `mail->recipients->MailRecipients` INNER JOIN `users` AS `mail->recipients` ON `mail->recipients`.`id` = `mail->recipients->MailRecipients`.`recipientId` AND `mail->recipients->MailRecipients`.`recipientId` = 1) ON `mail.id` = `mail->recipients->MailRecipients`.`mailId`;
```

sushantdhiman:
Many users are complaining that this fix broke their queries. I am reverting it for now

keyiis:
@sushantdhiman same issues in 4.13.6

Djalmar:
same issue in 4.41.0

PR #8313, opened 2017-09-15
code updates

### LINKED ISSUE #8312: Cannot remove ENUM column with MSSQL
## What are you doing?
Trying to remove an ENUM column via `queryInteface` on MSSQL
```js
const Test = sequelize.define('testtable', {
    id : { type: Sequelize.BIGINT, primaryKey : true, autoIncrement: true },
    state : { type : Sequelize.ENUM('a','b','c'), allowNull: false, defaultValue : 'a' },
});
Test.sync({force:true}).then(() => {
    sequelize.queryInterface.removeColumn('testtables', 'state');
});
```
The previous would result in the following error
```
Unhandled rejection SequelizeDatabaseError: ALTER TABLE DROP COLUMN state failed because one or more objects access this column.
```

The column should be dropped

The column cannot be dropped

__Dialect:__ mssql
__Sequelize version:__ 4.8.4

stefanobider:
### Pull Request check-list

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)? no: on master, npm install would give an error and so would npm run test and npm run test-mssql. I did test this with a simple code (same as in bug report https://github.com/sequelize/sequelize/issues/8312)
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving? yes
- [x] Have you added new tests to prevent regressions? no
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? No need
- [x] Did you follow the commit message conventions explained in CONTRIBUTING.md? NO

### Description of change
Closes #8312 
In order to fix issue #8312 , I had to include in the `removeColumn` wrapper a step to remove any CHECK constraints on the column. CHECK constraints are added on `ENUM` columns (created as `VARCHAR`) and prevent the column to be dropped similarly to default value constraints.
I added in `lib/dialects/mssql/query-generator.js` two methods:
* `getCheckConstraintQuery` (line 728)
* `dropCheckConstraintQuery` (line 739)
and in `lib/dialects/mssql/query-interface.js` I added the necessary steps to check the constraints and drop them in the same way as the other constraints are.

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -725,6 +725,25 @@ const QueryGenerator = {
     });
   },
 
+  getCheckConstraintQuery(tableName, attributeName) {
+    const sql = 'SELECT name FROM SYS.CHECK_CONSTRAINTS ' +
+      "WHERE PARENT_OBJECT_ID = OBJECT_ID('<%= table %>', 'U') " +
+      "AND PARENT_COLUMN_ID = (SELECT column_id FROM sys.columns WHERE NAME = ('<%= column %>') " +
+      "AND object_id = OBJECT_ID('<%= table %>', 'U'));";
+    return Utils._.template(sql)({
+      table: this.quoteTable(tableName),
+      column: attributeName
+    });
+  },
+
+  dropCheckConstraintQuery(tableName, constraintName) {
+    const sql = 'ALTER TABLE <%= table %> DROP CONSTRAINT <%= constraint %>;';
+    return Utils._.template(sql)({
+      table: this.quoteTable(tableName),
+      constraint: this.quoteIdentifier(constraintName)
+    });
+  },
+
   setAutocommitQuery() {
     return '';
   },
--- a/lib/dialects/mssql/query-interface.js
+++ b/lib/dialects/mssql/query-interface.js
@@ -58,6 +58,18 @@ const removeColumn = function(tableName, attributeName, options) {
       return this.sequelize.query(dropConstraintSql, options);
     })
     .then(() => {
+      //Check if the current column has a Check constraint (ENUMS)
+      const checkConstraintSql = this.QueryGenerator.getCheckConstraintQuery(tableName, attributeName);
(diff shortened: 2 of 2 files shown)


lib/dialects/mssql/query-generator.js
```diff
@@ -725,6 +725,25 @@ const QueryGenerator = {
     });
   },
 
+  getCheckConstraintQuery(tableName, attributeName) {
+    const sql = 'SELECT name FROM SYS.CHECK_CONSTRAINTS ' +
```
sushantdhiman: Use ES6 template string substitution instead of `lodash _.template`. So where appropriate 

```js
`SOME SQL STATEMENT ${tableName}` +
`SOME SQL STATEMENT ${columnName}`
```

sushantdhiman: On another thought, you should be able to get this constraint using `showConstraint` cc @harshithkashyap

lib/dialects/mssql/query-generator.js
```diff
@@ -725,6 +725,25 @@ const QueryGenerator = {
...
+      "AND PARENT_COLUMN_ID = (SELECT column_id FROM sys.columns WHERE NAME = ('<%= column %>') " +
+      "AND object_id = OBJECT_ID('<%= table %>', 'U'));";
+    return Utils._.template(sql)({
+      table: this.quoteTable(tableName),
+      column: attributeName
+    });
+  },
+
+  dropCheckConstraintQuery(tableName, constraintName) {
+    const sql = 'ALTER TABLE <%= table %> DROP CONSTRAINT <%= constraint %>;';
```
sushantdhiman: This should be possible by `removeConstraint`

harshithkashyap:
@sushantdhiman Now that we have introduced `showContraints` queryInterface method, we can refactor MSSQL `removeColumn` method. If you look at `lib/dialect/mssql/query-interface.js`, we are running multiple queries to find and drop constraints. We can replace this with `showConstraints` call followed by a `Promise.all` query to drop all constraints.

PR #8326, opened 2017-09-18
fix(query-generator): Options.group should be quoted

iyuq:
### Pull Request check-list

### Description of change
Use aggregate function I got the following sql:
```
'SELECT "userId", sum("cost") AS "sum" FROM "creative_projects" AS "creative_projects" WHERE "creative_projects"."userId" IN ('sub') AND ("creative_projects"."createdAt" >= '2017-03-31 16:00:00.000 +00:00' AND "creative_projects"."createdAt" < '2017-09-30 16:00:00.000 +00:00') GROUP BY userId;'
```
Obviously the quote is miss in `userId`.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1138,7 +1138,7 @@ const QueryGenerator = {
 
     // Add GROUP BY to sub or main query
     if (options.group) {
-      options.group = Array.isArray(options.group) ? options.group.map(t => this.quote(t, model)).join(', ') : options.group;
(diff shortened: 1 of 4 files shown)


iyuq:
I don't have time to add test case, can somebody help me?

iyuq:
I looked at the test case that failed,
```
{
          title: 'single string argument is not quoted',
          arguments: ['myTable', {group: 'name'}],
          expectation: 'SELECT * FROM `myTable` GROUP BY name;',
          context: QueryGenerator
 }
```
It seems it intends to let the single string argument not quoted. What is the purpose of this?

mkaufmaner:
This is a vulnerability and the test case should be modified to check for quotes

iyuq:
Ok, I will commit to fix the test.

PR #8339, opened 2017-09-19
fix(DataTypes/DATE): Adds support for DATE with Infinity/-Infinity

### LINKED ISSUE #8336: Support for Infinity and -Infinity Dates in PostgreSQL
## What are you doing?
Trying to insert infinity and -infinity into a PostgreSQL column where the attribute has a type of Date.

Definition:
```js
'deletedAt': {
    type: DataTypes.DATE,
    field: 'deleted_at',
    allowNull: false,
    defaultValue: Infinity
}
```

For the record to be inserted correctly.

```sh
{ SequelizeValidationError: null is not a valid date
    at DATE.validate (node_modules/sequelize/lib/data-types.js:312:11)
    at Object.escape (node_modules/sequelize/lib/dialects/abstract/query-generator.js:898:26)
    at Object.insertQuery (node_modules/sequelize/lib/dialects/abstract/query-generator.js:187:28)
    at QueryInterface.insert (node_modules/sequelize/lib/query-interface.js:519:37)
    at Promise.try.then.then.then (node_modules/sequelize/lib/model.js:3591:56)
    at tryCatcher (node_modules/sequelize/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (node_modules/sequelize/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (node_modules/sequelize/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (node_modules/sequelize/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (node_modules/sequelize/node_modules/bluebird/js/release/promise.js:693:18)
    at Async._drainQueue (node_modules/sequelize/node_modules/bluebird/js/release/async.js:133:16)
    at Async._drainQueues (node_modules/sequelize/node_modules/bluebird/js/release/async.js:143:10)
    at Immediate.Async.drainQueues (node_modules/sequelize/node_modules/bluebird/js/release/async.js:17:14)
    at runCallback (timers.js:672:20)
    at tryOnImmediate (timers.js:645:5)
    at processImmediate [as _immediateCallback] (timers.js:617:5) name: 'SequelizeValidationError', errors: [] }
```

__Dialect:__ postgres
__Dialect version:__ Stable???
__Database version:__ 9.6 (Current)
__Sequelize version:__ 4.8.4

mkaufmaner:
### Pull Request check-list

### Description of change

Adds support for Infinity/-Infinity with for DATE data types.

Closes #8336

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3256,7 +3256,7 @@ class Model {
 
       // If there's a data type setter
       if (this.constructor._hasDataTypeSetters && this.constructor._dataTypeSetters[key]) {
-        this.constructor._dataTypeSetters[key].call(this, value, key);
+        this.dataValues[key] = this.constructor._dataTypeSetters[key].call(this, value, key);
       }
 
       // If not raw, and there's a custom setter
@@ -3276,8 +3276,10 @@ class Model {
           this._setInclude(key, value, options);
           return this;
         } else if (this.constructor._hasDataTypeComparators && this.constructor._dataTypeComparators[key] && this.constructor._dataTypeComparators[key].call(this, key, value, originalValue, options)) {
-          this._previousDataValues[key] = originalValue;
-          this.changed(key, true);
+          if (!Utils.isPrimitive(value) && value !== null || value !== originalValue) {
+            this._previousDataValues[key] = originalValue;
+            this.changed(key, true);
+          }
 
           // set data value
           this.dataValues[key] = value;
@@ -3348,12 +3350,7 @@ class Model {
             }
           }
 
-          if (
-            !options.raw && (
-              this.constructor._hasDataTypeComparators && this.constructor._dataTypeComparators[key] && this.constructor._dataTypeComparators[key].call(this, key, value, originalValue, options) ||
-              (!Utils.isPrimitive(value) && value !== null || value !== originalValue)
-            )
-          ) {
+          if (!options.raw && (!Utils.isPrimitive(value) && value !== null || value !== originalValue)) {
             this._previousDataValues[key] = originalValue;
             this.changed(key, true);
           }
--- a/lib/model.js
+++ b/lib/model.js
@@ -3281,14 +3281,14 @@ class Model {
                   value = moment(value).format('YYYY-MM-DD');
                 }
               } else { // go ahread and parse as Date if required
-                if (!(value instanceof Date) && value !== Infinity && value !== -Infinity) {
+                if (!(value instanceof Date)) {
                   value = new Date(value);
                 }
                 if (originalValue) {
-                  if (!(originalValue instanceof Date) && value !== Infinity && value !== -Infinity) {
+                  if (!(originalValue instanceof Date)) {
                     originalValue = new Date(originalValue);
                   }
-                  if (value === originalValue || value instanceof Date && originalValue instanceof Date && value.getTime() === originalValue.getTime()) {
+                  if (value.getTime() === originalValue.getTime()) {
                     return this;
                   }
                 }
--- a/lib/model.js
+++ b/lib/model.js
@@ -3281,14 +3281,14 @@ class Model {
                   value = moment(value).format('YYYY-MM-DD');
                 }
               } else { // go ahread and parse as Date if required
-                if (!(value instanceof Date)) {
+                if (!(value instanceof Date) && value !== Infinity && value !== -Infinity) {
                   value = new Date(value);
                 }
                 if (originalValue) {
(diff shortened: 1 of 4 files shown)


mkaufmaner:
Marked as WIP because I am going to be working on the test coverage.

lib/model.js
```diff
@@ -3281,14 +3281,14 @@ class Model {
                   value = moment(value).format('YYYY-MM-DD');
                 }
               } else { // go ahread and parse as Date if required
-                if (!(value instanceof Date)) {
+                if (!(value instanceof Date) && value !== Infinity && value !== -Infinity) {
```
sushantdhiman: Too much logic here which should be handled by datatype and is postgres specific. 

//cc @janmeier Should we get dialect date type and call specific method to parse `value` ? This way postgres can do its thing and other dialect can do theirs ?

janmeier: Yep, a parse method on the date type could be a good idea

mkaufmaner:
I was looking into parse methods on the data type and will make changes accordingly.  Thanks for the quick response!

lib/model.js
mkaufmaner: Maybe custom getters and setter for the DATE.prototype?

janmeier: That could work :)

mkaufmaner:
@janmeier @sushantdhiman After reviewing the code for `Model.get` and `Model.set` maybe a good idea would be to extend getters and setters into the DataType members.  The visibility of these changes would be private.  The DataType getter and setter would be called before user defined getter and setter methods ([Getters-Setters](http://docs.sequelizejs.com/manual/tutorial/models-definition.html#getters-setters)).  This will allow us to support dialect specific functionality for abstract data types.

For example, in order to support `Infinity` and `-Infinity` values for `DataType.DATE` attributes in PostgreSQL, we would add the following code to `/lib/dialects/postgres/data-types.js`;
```js
DATE.prototype.set = function getter(value, key) {
	if (!(value instanceof Date) && value !== Infinity && value !== -Infinity) {
		value = new Date(value);
	}

	this.setDataValue(key, value);
};
```

mkaufmaner:
Maybe we can go as far as abstracting the `changed` logic into the DataType as well;
```js
DATE.prototype.hasChanged = function hasChanged(value, originalValue) {
	if (originalValue) {
		if (!(originalValue instanceof Date) && value !== Infinity && value !== -Infinity) {
			originalValue = new Date(originalValue);
		}
		if (value === originalValue || value instanceof Date && originalValue instanceof Date && value.getTime() === originalValue.getTime()) {
			return true;
		}
	}
};
```

Thoughts?

sushantdhiman:
I see you are allowing users to set their own getter / setter for types, good idea but I think they should be able to call `super.get() / super.set()` perhaps.

we should lock in an API about what we really want so we can improve other types and reduce code from model file as well. Please open an issue and we can discuss over there.

First proposal could be to include a way in datatypes to set, get or compare. Let see what we can refactor

lib/model.js
```diff
@@ -952,20 +958,34 @@ class Model {
...
-        this._hstoreAttributes.push(name);
-      } else if (definition.type instanceof DataTypes.RANGE || DataTypes.ARRAY.is(definition.type, DataTypes.RANGE)) {
-        this._rangeAttributes.push(name);
-      } else if (definition.type instanceof DataTypes.JSON) {
-        this._jsonAttributes.push(name);
-      } else if (definition.type instanceof DataTypes.VIRTUAL) {
-        this._virtualAttributes.push(name);
-      } else if (definition.type instanceof DataTypes.GEOMETRY) {
-        this._geometryAttributes.push(name);
+      if (definition.type) {
```
janmeier: In which case would the definition not have a type?

test/integration/dialects/postgres/data-types.test.js
```diff
@@ -0,0 +1,89 @@
...
+          return User.create({
+            username: 'bob',
+            anotherTime: Infinity
+          }, {
+            validate: true
+          });
+        }).then(user => {
+          expect(user.username).to.equal('bob');
+          expect(user.beforeTime).to.equal(-Infinity);
+          expect(user.sometime).to.not.equal(Infinity);
```
janmeier: `to.be.closeTo(now)`

mkaufmaner:
@sushantdhiman I will work on a quick proposal with some examples as to what we should be aiming for when it comes to implementing a new API for data types.

Issue #8321, opened 2017-09-17
Nested include with limit and order generates incorrect SQL


a60814billy:
## What are you doing?

The code in line 39 (`return User.findAll( ...`) is to find top 25 users and their Posts, then order by users createdAt time

```js
'use strict'

const Sequelize = require('sequelize')

const sequelize = new Sequelize('postgres://user@localhost:5432/test', {
  logging: console.log
})

const User = sequelize.define('User', {
  id: {
    type: Sequelize.UUID,
    primaryKey: true
  },
  username: Sequelize.STRING,
  password: Sequelize.STRING,
  createdAt: Sequelize.DATE,
  updatedAt: Sequelize.DATE
})
const Post = sequelize.define('Post', {
  id: {
    type: Sequelize.UUID,
    primaryKey: true
  },
  title: Sequelize.STRING,
  content: Sequelize.STRING,
  ownerId: Sequelize.UUID,
  createdAt: Sequelize.DATE,
  updatedAt: Sequelize.DATE
})

User.hasMany(Post, {foreignKey: 'ownerId', as: 'ownedPost'})
Post.belongsTo(User, {foreignKey: 'ownerId', as: 'owner'})

sequelize.authenticate()
  .then(() => {
    return sequelize.sync()
  })
  .then(() => {
    return User.findAll({
      limit: 25,
      include: ['ownedPost'],
      order: ['createdAt']
    }).then((result) => {
      console.log(result)
      return sequelize.close()
    })
  })

```

## What do you expect to happen?

expect order both in subquery and main query.
it works correctly in sequelize <= 4.8.3

```sql
SELECT "User".*, "ownedPost"."id" AS "ownedPost.id", "ownedPost"."title" AS "ownedPost.title", "ownedPost"."content" AS "ownedPost.content", "ownedPost"."ownerId" AS "ownedPost.ownerId", "ownedPost"."createdAt" AS "ownedPost.createdAt", "ownedPost"."updatedAt" AS "ownedPost.updatedAt" 
FROM (
    SELECT "User"."id", "User"."username", "User"."password", "User"."createdAt", "User"."updatedAt" FROM "Users" AS "User" ORDER BY "User"."createdAt" LIMIT 25) AS "User" 
LEFT OUTER JOIN "Posts" AS "ownedPost" ON "User"."id" = "ownedPost"."ownerId"
ORDER BY "User"."createdAt"
```

## What is actually happening?

[...]

PR #8341, opened 2017-09-20
test(integration/model/findAll): Add test for findAll with limit and mixed association results

danconnell:
### Pull Request check-list

### Description of change

Added a test to prove that https://github.com/sequelize/sequelize/issues/8321 is a real problem. Not sure if this is the right spot to add a test like this, but this is more just for proof than anything else.

This test passes on sequelize v4.8.3 and fails on sequelize v4.8.4.

### DIFF
--- a/test/integration/model/findAll.test.js
+++ b/test/integration/model/findAll.test.js
@@ -1034,6 +1034,35 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           });
         });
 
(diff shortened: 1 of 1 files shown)


sushantdhiman:
https://github.com/sequelize/sequelize/commit/79485c3c59a62d43db70e1bad0765e2a7c7d0d50

sushantdhiman:
Reopening it as this test can be valuable for preventing future regressions around `order`

PR #8342, opened 2017-09-20
Pass-through postgres connection option 'statement_timeout'

amasad:
node-postgres recently added support for 'statement_timeout' -- a per connection query timeout.

See https://github.com/brianc/node-postgres/pull/1436

### DIFF
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -79,7 +79,9 @@ class ConnectionManager extends AbstractConnectionManager {
           'binary',
           // This should help with backends incorrectly considering idle clients to be dead and prematurely disconnecting them.
           // this feature has been added in pg module v6.0.0, check pg/CHANGELOG.md
-          'keepAlive'
+          'keepAlive',
+          // Times out queries after a set time in milliseconds. Added in pg v7.3
+          'statement_timeout'
         ]));
     }
 


amasad:
Yes, I just realized that v7 isn't supported after trying to run it and hit the #7998 bug

mkaufmaner:
> Yes, I just realized that v7 isn't supported after trying to run it and hit the #7998 bug

That being said, wouldnt we want to hold off on this PR?  Also, not a huge fan of adding dialect specific options to abstract components.

sushantdhiman:
> That being said, wouldnt we want to hold off on this PR?

Dialect specific option so there should be no problem. Sequelize is incompatible with MULTIPLE statements type, this option not related

> Also, not a huge fan of adding dialect specific options to abstract components.

Not sure I understood what you meant by abstract component, this is Postgres specific connection manager. This is a good place to set such options, we already have many pg specific options here like `keepAlive` etc and I think only way to pass these options is here. 

We are not actually adding it, we are white listing it so users can pass it as `dialectOptions`

jkennedy1980:
Model.upsert() fails with `pg@7.3.0` due to the multiple result problem. Do we have a time frame for a fix on the multiple result issue?

sushantdhiman:
@jkennedy1980 No we dont have a timeframe. Its an open source project, when someone is free enough to work on it, fix will land. Till then just wait :)

jkennedy1980:
And you guys are sure you don't want to accept PR8029: https://github.com/sequelize/sequelize/pull/8029 - It basically converts the multiple results to the same thing you get with `pg@6`? If not, can you close PR8029 so people don't think it's being worked on?

sushantdhiman:
That PR attempts to solve it but its not a good solution.

I haven't had chance to review that PR but as referred in https://github.com/sequelize/sequelize/issues/7998 we need a way to have MULTIPLE type statement define which result they need.

If someone is willing to solve that issue I can explain in details what we need

jkennedy1980:
Ok. It seems that you have ideas about how you want the API to change to support multiple results. There are many issues/PRs open where people are proposing different solutions. I think it's time that you document what you're thinking about the API so that the community can discuss it and maybe even implement it. This would prevent people wasting their time on undesired solutions. I may even choose to implement if there is a good API design in place.

PR #8350, opened 2017-09-21
feat(query): introduce alias substitution

### LINKED ISSUE #2084: Max identifier length exceeded with nested includes

clncln1:
### Pull Request check-list

### Description of change

This is an attempt to fix #2084
The feature basically introduces an ordinal-based (deep) alias substitution for column names as well as table aliases for all includes within a query.
It allows more complex use cases/queries with sequelize and postgresql by avoiding to hit the 63 character limit for identifiers that soon.

Let me know what you think about the approach. If you are okay with it, I will update the PR to:
1) make substitution option default to false (currently true in order to show full compliance with the test suite)
2) implement substitution-specific tests to cover 100% of the added lines
3) update doc

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -709,7 +709,7 @@ const QueryGenerator = {
     potentially also be used for other places where we want to be able to call SQL functions (e.g. as default values)
    @private
   */
-  quote(collection, parent, connector) {
+  quote(collection, parent, connector, queryOptions) {
     // init
     const validOrderOptions = [
       'ASC',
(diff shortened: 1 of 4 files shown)


sushantdhiman:
This solution adds too much logic to an already complex base query generator. I think a better solution will be https://github.com/sequelize/sequelize/issues/5708 

But that fix of-course will require some serious refactoring work

clncln1:
I agree that it's quite complex.
But even if you would remove column aliasing at all, you would still need to have table aliasing in place (where you also get bitten by the 63 character limit if you stick to the current aliasing scheme).
Is there somebody working on that?

yohannprigent:
This PR would really help all users affected with #2084. Any news to merge this one ?

javagaba:
+1

Relequestual:
It would be great to see this merged. We are currenly having this issue.

SimonSchick:
Closing this due to age and merge conflicts, feel free to rebase/refresh this PR and re-open.

PR #8371, opened 2017-09-25
feat(postgresql/validateSchemas)

hiradimir:
### Pull Request check-list

### Description of change

I thought production deployment cycle.
But, I am worried migration script has mistake because it is generated by manually.
I am thinking schema-validation feature and then implementing.

I think I should write test too, but I cannot all test of model. 
For example migration-scripts are correct sql, but migrated db is may not same model-schema.
ex) model string field length is 255, but schema string field length 64.

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -472,6 +472,21 @@ Sequelize has a [sister library][1] for programmatically handling execution and 
 
 Using `queryInterface` object described before you can change database schema. To see full list of public methods it supports check [QueryInterface API][2]
 
+## Schema Validation
+
+**Note:** _The `Sequelize#validateSchemas` still supports postgresql only._
+
+You can check mistake of Migration with `Sequelize#validateSchemas` .
+ValidateSchemas is search different between DB-Table and Sequelize-Models.
+
+```js
+sequlize.sync({force: process.env.NODE_ENV !== 'production'})
+  .then(()=>{
+    return sequlize.validateSchemas()
+  }).then(()=>{
+    // do something
+  })
+```
 
 [0]: https://github.com/sequelize/cli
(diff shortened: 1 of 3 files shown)


eseliger:
@hiradimir there are conflicts

hiradimir:
@eseliger fix it!

This PR is still for postgresql only. 
PR is decreased coverage because CI seems to is not checking with postgresql dialect.

eseliger:
somehow, the travis hook seems not to be triggered 🤔 try again in a few hours or so, maybe they are having difficulties right now

sushantdhiman:
Validating schema before running a migration seems like a CLI domain problem. We should not add this to ORM code base. Its supposed to be used with migration, lets move this to CLI then.

If you want you can open a PR to CLI but then you will have to add support for other database as well.

hiradimir:
@sushantdhiman 
I hope to add validate schema feature to ORM code base.
Because, I think this feature shoud be in options for sequelize#sync.
The idea is from hibernate 
ex: [hibernate.hbm2ddl.auto: validate](https://docs.jboss.org/hibernate/orm/5.0/manual/en-US/html/ch03.html)

But, I did not add sequelize#sync options, I thought it is risk.
Because, this code was imcplete, I shoud be want to add options after more brush up.

----

I want this feature running in another section on sequelize-cli.
If you not accept this PR, I will create another plugin ex) npm:seqelize-validate-schema.

sushantdhiman:
A plugin seem like best approach for now @hiradimir , we can always include it in core if there is huge demand but doing opposite is complex.

hiradimir:
@sushantdhiman Okey, Thanks for reply!

PR #8379, opened 2017-09-27
make mysql foreign key errors contain some useful structured info

iamjochem:
### Pull Request check-list

- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? (NOT APPLICABLE I believe)

### Description of change

This change fixes the extraction of structured meta-data from MySQL Foreign Key Contraint error messages so that they can be added to `options` argument of the `ForeignKeyConstraintError` constructor.

Additionally it adds a new `ForeignKeyConstraintError` option key named "reltype" which can be used to specify the "direction" of constraint (i.e. "child" or "parent") ... this option key (like all the existing ones e.g. `table`, `fields`, `value`) is completely optional (i.e. no requirement to set it).

### N.B.

I have not attempted to make the same fix for other dialects because I am not familiar with the specific content/structure their FK constraint error messages, a cursory glance at the relevant code in other dialects suggest that they also suffer from a lack of meta-data extraction ... maybe this PR can serve someone as a basis for making the changes in the remaining dialects.


lib/dialects/mysql/query.js
```diff
@@ -200,24 +202,26 @@ class Query extends AbstractQuery {
...
+        // e.g. CONSTRAINT `example_constraint_name` FOREIGN KEY (`example_id`) REFERENCES `examples` (`id`)
+        const match  = err.message.match(/CONSTRAINT `(.*)` FOREIGN KEY \(`(.*)`\) REFERENCES `(.*)` \(`(.*)`\)/);
+        const fields = match ? match[2].split(/`, *`/) : undefined;
+
+        // TODO: it is not clear whether `table` & `fields` is meant refer to the source or the referenced entity,
+        //       ideally you would want to name both, no?
 
         return new sequelizeErrors.ForeignKeyConstraintError({
-          fields: null,
+          reltype: errCode + '' === '1451' ? 'parent' : 'child',
```
sushantdhiman: Use `String(errCode)`

lib/dialects/mysql/query.js
```diff
@@ -200,24 +202,26 @@ class Query extends AbstractQuery {
...
-        });
-      }
+      case 1451:
       case 1452: {
-        const match = err.message.match(/FOREIGN KEY \(`(.*)`\) REFERENCES `(.*)` \(`(.*)`\)(.*)\)$/);
+        // e.g. CONSTRAINT `example_constraint_name` FOREIGN KEY (`example_id`) REFERENCES `examples` (`id`)
+        const match  = err.message.match(/CONSTRAINT `(.*)` FOREIGN KEY \(`(.*)`\) REFERENCES `(.*)` \(`(.*)`\)/);
+        const fields = match ? match[2].split(/`, *`/) : undefined;
+
+        // TODO: it is not clear whether `table` & `fields` is meant refer to the source or the referenced entity,
```
sushantdhiman: We are exposing all fields given by error message, is there any way to get additional fields?

iamjochem: @sushantdhiman - I do not understand your question! (how can there be "additional" things to "all" the things?)

*please note*: `fields` was an existing property, I assumed it was plural for a reason (my guess would be because of the possibility of compound foreign keys) - I chose to set `fields` as an array because of this.

lib/dialects/mysql/query.js
iamjochem: done

lib/dialects/mysql/query.js
sushantdhiman: @iamjochem I am saying we are already exposing all fields available in error message. Your comment says we might want fields from source or referenced table. So I am asking is there any way to get those fields?

If not just remove that comment.

iamjochem:
@sushantdhiman - regarding tests, I don't know what to test, please help me with the following points:

firstly, for reference, a list of `ForeignKeyConstraintError` specific  properties:
```js
    this.fields = options.fields;
    this.table = options.table;
    this.value = options.value;
    this.index = options.index;
    this.reltype = options.reltype;
```

1. I am not sure what `fields` or `table` mean, are they intended to name the source or the referenced entity? - Postgres dialect sets `table` to the source table, but I have chosen to set it to the referenced table - the discrepency is also due to the differences in dialect error messages (you can only extract what is there from the error message). 
2. No dialect currently sets `fields` or `value` (sqlite sets none of them and mssql sets only `index`)
3. There is currently no test for structure/data of a `ForeignKeyConstraintError` (the only reference to this type of exception in the tests are expectations that it is thrown) - none of the pre-existing properties `fields`, `table`, `index` & `value` have tests why is `reltype` different?
4. None of the above properties are required in the constructor, a test for these properties would be simply "can I construct a new instance & are all properties null/undefined", no?
5. should there be dialect specific tests for the structure/data of a `ForeignKeyConstraintError`? 

I think the reason there is no test for structure/data of a `ForeignKeyConstraintError` is due to the following facts:

1. it is unclear what all the properties are intended for.
2. it is not really possible to paper over the differences, between dialects, regarding information available in FK error messages.
3. in practice it does not matter that each dialect set different details, as long as a developer has enough detail to trace the source of the FK error.

lib/dialects/mysql/query.js
iamjochem: aha - I see, comment has been removed. (the comment was mostly because I wasn't sure what "fields" was intended to contain - as per my longer reply/question below)

sushantdhiman:
@iamjochem You can add a new file called `errors.test.js` [here](https://github.com/sequelize/sequelize/tree/master/test/integration/dialects/mysql). Write test where this error will be thrown (I assume writing non existing primary key id into a foreign key column). Then assert `relType` and `fields` are properly set

iamjochem:
@sushantdhiman understood, thanks for clarifying that you wanted a test specifically/only for the mysql dialect.

iamjochem:
I have added a test file for ForeignKeyConstraintError properties (as they are set by the MySQL dialect).

test/integration/dialects/mysql/errors.test.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
+
+const chai = require('chai'),
```
sushantdhiman: Use const per line

test/integration/dialects/mysql/errors.test.js
```diff
@@ -0,0 +1,81 @@
...
+  expect = chai.expect,
+  Support = require(__dirname + '/../../support'),
+  dialect = Support.getTestDialect(),
+  DataTypes = require(__dirname + '/../../../../lib/data-types')
+  ;
+
+if (dialect === 'mysql') {
+  describe.only('[MYSQL Specific] Errors', () => {
```
sushantdhiman: no `only`, I have add a rule in eslint config to prevent this in future https://github.com/sequelize/sequelize/commit/3e0c9b1acfd2538d7a155e6c3b3783938121bcd6

iamjochem: oops. sorry - my bad. I've removed it.

test/integration/dialects/mysql/errors.test.js
sushantdhiman: We run eslint on tests and we have to change all those tests before we make this change. Its not very critical, its just a styling we have been following with new code base.

iamjochem:
ah great merged! don't suppose you want to take a look at the following PR?

    https://github.com/sequelize/sequelize/pull/7557

It's about improving ValidationErrors, I feel it's alot more useful/helpful than this PR but it's being gathering dust for a while now.

PR #8357, opened 2017-09-22
8336 Adds support for postgres infinity timestamp

mkaufmaner:
### Pull Request check-list
- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you follow the commit message conventions explained in CONTRIBUTING.md?

### Description of change 
Closes #8336

See #8356 (Partial fix still requires further work)
See #8339

PR #8384, opened 2017-09-28
Fix null check for DATEONLY

eseliger:
Fixes comment mentioned [here](https://github.com/sequelize/sequelize/pull/8357/files#r141506197)

`DATEONLY` fields were sanitized wrong when being `NULL`.

### DIFF
--- a/test/integration/associations/has-many.test.js
+++ b/test/integration/associations/has-many.test.js
@@ -730,6 +730,8 @@ describe(Support.getTestDialectTeaser('HasMany'), () => {
         return this.task.getUsers();
       }).then(_users => {
         expect(_users).to.have.length(0);
+      }).finally(() => {
+        this.sequelize.options.omitNull = false;
       });
     });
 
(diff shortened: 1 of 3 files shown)


sushantdhiman:
Tests breaking

eseliger:
I know, if it's run .only this doesn't happen, do you have any idea why? Couldn't find the reason. Maybe someone of you already got into that problem before

eseliger:
Crap :/ trying to find something out

eseliger:
That took hella long to find, very bad idea to modify the global sequelize instance in a test-case :|

test/integration/associations/has-many.test.js
```diff
@@ -730,6 +730,8 @@ describe(Support.getTestDialectTeaser('HasMany'), () => {
         return this.task.getUsers();
       }).then(_users => {
         expect(_users).to.have.length(0);
+      }).finally(() => {
+        this.sequelize.options.omitNull = false;
```
janmeier: Good find :O

Issue #8385, opened 2017-09-28
Migration: removeColumn fails on unique contraints


austindebruyn:
## What are you doing?
Given a column defined like this:

```
      email: {
        type: Sequelize.STRING,
        unique: true
      },
```

I'd like to drop it like this

```
queryInterface.removeColumn('Users', 'email')
```

I have a repro case that shows the error here: https://github.com/austindebruyn/sequelize-err-on-rename-repro
I've seen this locally and on a trusty DO box.

```js
// code here
```

## What do you expect to happen?
Column to be dropped.

## What is actually happening?
A SequelizeDatabaseError error is thrown.

Before removing a column, the mysql dialect checks that there's no foreign keys attached to it with this query:
```
SELECT CONSTRAINT_NAME as constraint_name FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE (REFERENCED_TABLE_NAME = 'users' AND REFERENCED_COLUMN_NAME = 'email') OR (TABLE_NAME = 'users' AND COLUMN_NAME = 'email')
```

It looks to me like [KEY_COLUMN_USAGE](https://dev.mysql.com/doc/refman/5.7/en/key-column-usage-table.html) table stores not just foreign keys but all constraints. In my failing case, `email` is unique and present in this table, so sequelize aborts the removeColumn with the following error:

```
vagrant@vagrant-ubuntu-trusty-64:~/projects/sequelize-err-on-rename-repro$ ./node_modules/.bin/sequelize db:migrate

Sequelize CLI [Node: 8.6.0, CLI: 3.0.0, ORM: 4.11.1]

WARNING: This version of Sequelize CLI is not fully compatible with Sequelize v4. https://github.com/sequelize/cli#sequelize-support

Loaded configuration file "config/config.json".
Using environment "development".
== 20170928040207-remove-email-column: migrating =======
SELECT CONSTRAINT_NAME as constraint_name FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE (REFERENCED_TABLE_NAME = 'Users' AND REFERENCED_COLUMN_NAME = 'email') OR (TABLE_NAME = 'Users' AND COLUMN_NAME = 'email')
[ TextRow { constraint_name: 'email' },
  TextRow { constraint_name: 'email' } ]
[...]

PR #8386, opened 2017-09-28
bugfix: removing unique contraint column (#8385)

austindebruyn:
### Pull Request check-list

### Description of change

Fixes issue I opened here https://github.com/sequelize/sequelize/issues/8385 about dropping unique contraint column.

Before removing a column, the mysql dialect checks that there's no foreign keys attached to it with this query:
```
SELECT CONSTRAINT_NAME as constraint_name FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE (REFERENCED_TABLE_NAME = 'users' AND REFERENCED_COLUMN_NAME = 'email') OR (TABLE_NAME = 'users' AND COLUMN_NAME = 'email')
```

It looks to me like [KEY_COLUMN_USAGE](https://dev.mysql.com/doc/refman/5.7/en/key-column-usage-table.html) table stores not just foreign keys but all constraints. In my failing case, `email` is unique and present in this table, so sequelize aborts the removeColumn.

The fix is to add an additional condition onto the check: ` AND REFERENCED_TABLE_NAME IS NOT NULL` which also verifies that the column is a foreign key, not some other kind of constraint.

I've got a failing unit test and a passing fix.

### DIFF
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -342,6 +342,7 @@ const QueryGenerator = {
       + ' AND REFERENCED_COLUMN_NAME = ' + wrapSingleQuote(columnName)
       + ') OR (TABLE_NAME = ' + wrapSingleQuote(tableName)
       + ' AND COLUMN_NAME = ' + wrapSingleQuote(columnName)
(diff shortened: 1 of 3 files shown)


sushantdhiman:
Thanks

PR #8389, opened 2017-09-28
fix(model): do not mark a column as changed if a custom setter sets it to the same value

knibb800:
### Pull Request check-list

lint is not Windows-friendly - gives many "Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style" warnings

### Description of change

Some models can define a custom setter to modify the value that is set. For determining whether a column is changed, the original value needs to be compared with the value that is really set, not with the value that was input and modified.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3247,7 +3247,8 @@ class Model {
       // If not raw, and there's a custom setter
       if (!options.raw && this._customSetters[key]) {
         this._customSetters[key].call(this, value, key);
-        if (!Utils.isPrimitive(value) && value !== null || value !== originalValue) {
+        const newValue = this.dataValues[key];
+        if (!Utils.isPrimitive(newValue) && newValue !== null || newValue !== originalValue) {
           this._previousDataValues[key] = originalValue;
           this.changed(key, true);
         }
--- a/lib/model.js
+++ b/lib/model.js
@@ -3247,6 +3247,8 @@ class Model {
       // If not raw, and there's a custom setter
       if (!options.raw && this._customSetters[key]) {
         this._customSetters[key].call(this, value, key);
+        // custom setter should have changed value, get that changed value
+        // TODO: v5 make setters return new value instead of changing internal store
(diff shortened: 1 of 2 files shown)


test/unit/instance/set.test.js
```diff
@@ -74,5 +76,55 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
...
+      });
+
+      const User = current.define('User', {
+        phoneNumber: {
+          type: DataTypes.STRING,
+          set (val) {
+            if (typeof val === 'string') {
+              // Canonicalize phone number
+              val = val.replace(/^\+/, '00').replace(/\(0\)|[\s+\/.\-\(\)]/g, '');
+              console.log('val', val);
```
sushantdhiman: ^ console.log

lib/model.js
```diff
@@ -3247,7 +3247,8 @@ class Model {
       // If not raw, and there's a custom setter
       if (!options.raw && this._customSetters[key]) {
         this._customSetters[key].call(this, value, key);
-        if (!Utils.isPrimitive(value) && value !== null || value !== originalValue) {
+        const newValue = this.dataValues[key];
```
sushantdhiman: Add a comment saying "custom setter should have changed value, get that changed value"

Also add a comment saying "TODO: v5 make setters return new value instead of changing internal store"

test/unit/instance/set.test.js
```diff
@@ -74,5 +76,55 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
       expect(user.get('date')).to.be.an.instanceof(Date);
       expect(user.get('date')).not.to.be.NaN;
     });
+
+    describe('custom setter', () => {
+      before(function() {
+        this.stubCreate = sinon.stub(current.getQueryInterface(), 'insert', (instance) => {
+          return Promise.resolve([ instance, 1]);
```
eseliger: the space before `instance` can be removed :)

PR #8383, opened 2017-09-27
fix(model): properly serialize WHERE attributes in soft deletes

gabegorelick:
Closes #8382

### Pull Request check-list

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? **N/A**
- [x] Did you follow the commit message conventions explained in CONTRIBUTING.md?

### Description of change

Infinity date support is pretty new in Sequelize (1 day old), so it still has some kinks to work out. This PR fixes https://github.com/sequelize/sequelize/issues/8382.

PR #8397, opened 2017-09-29
fix(postgres/date): upserts with Infinity

### LINKED ISSUE #8396: Upsert error when Infinity Date is part of primary key
## What are you doing?

I want to add `deletedAt` to my composite primary keys so I can soft delete and reuse identifiers. For example, if you soft delete a user you should be able to register a new user with the same name. Since primary key attributes can't be NULL in Postgres, that means we need to use a different sentinel value like `Infinity` for `deletedAt`.

Test case added to `upsert.test.js`:
```js
if (dialect.match(/^postgres/)) {
  it('works when deletedAt is Infinity and part of primary key', function () {
      name: {
        type: DataTypes.STRING,
        primaryKey: true
      },
      address: DataTypes.STRING,
      deletedAt: {
        type: DataTypes.DATE,
        primaryKey: true,
        allowNull: false,
        defaultValue: Infinity
      }
    }, {
      paranoid: true
    });

      return User.build({name: 'user1'})
        .save()
        .then(() => {
          return User.upsert({name: 'user1', address: 'address'});
        });
    });
  });
}
```

Sequelize should generate SQL that looks something like:
```sql
CREATE OR REPLACE FUNCTION pg_temp.sequelize_upsert() RETURNS integer AS $func$ BEGIN INSERT INTO "Users" ("name","address","deletedAt","createdAt","updatedAt") VALUES ('user1','address','Infinity','1970-01-01 00:00:00.000 +00:00','1970-01-01 00:00:00.000 +00:00'); RETURN 1; EXCEPTION WHEN unique_violation THEN UPDATE "Users" SET "name"='user1',"address"='address',"updatedAt"='1970-01-01 00:00:00.000 +00:00' WHERE (("name" = 'user1' AND "deletedAt" = 'Infinity')); RETURN 2; END; $func$ LANGUAGE plpgsql; SELECT * FROM pg_temp.sequelize_upsert()
```
Note the `'Infinity'` in quotes.

Sequelize generates the following SQL:
```sql
CREATE OR REPLACE FUNCTION pg_temp.sequelize_upsert() RETURNS integer AS $func$ BEGIN INSERT INTO "Users" ("name","address","deletedAt","createdAt","updatedAt") VALUES ('user1','address','Infinity','1970-01-01 00:00:00.000 +00:00','1970-01-01 00:00:00.000 +00:00'); RETURN 1; EXCEPTION WHEN unique_violation THEN UPDATE "Users" SET "name"='user1',"address"='address',"updatedAt"='1970-01-01 00:00:00.000 +00:00' WHERE (("name" = 'user1' AND "deletedAt" = Infinity)); RETURN 2; END; $func$ LANGUAGE plpgsql; SELECT * FROM pg_temp.sequelize_upsert()
```

This causes a `SequelizeDatabaseError: column "infinity" does not exist`.

__Dialect:__ postgres
__Dialect version:__ XXX
__Database version:__ any
__Sequelize version:__ master

gabegorelick:
Closes #8396

### Pull Request check-list

### Description of change

Similar to #8383, but for upserts. Sequelize wasn't serializing attributes in the WHERE clause on the UPDATE branch.

### DIFF
--- a/test/integration/model/upsert.test.js
+++ b/test/integration/model/upsert.test.js
@@ -487,6 +487,44 @@ describe(Support.getTestDialectTeaser('Model'), () => {
             });
         });
       }
+
+      if (dialect.match(/^postgres/)) {
+        it('works when deletedAt is Infinity and part of primary key', function() {
+          const User = this.sequelize.define('User', {
+            name: {
+              type: DataTypes.STRING,
+              primaryKey: true
+            },
+            address: DataTypes.STRING,
+            deletedAt: {
+              type: DataTypes.DATE,
+              primaryKey: true,
+              allowNull: false,
+              defaultValue: Infinity
+            }
+          }, {
+            paranoid: true
+          });
+
+          return User.sync({ force: true }).then(() => {
+            return Promise.all([
+              User.create({ name: 'user1' }),
+              User.create({ name: 'user2', deletedAt: Infinity }),
+
+              // this record is soft deleted
+              User.create({ name: 'user3', deletedAt: -Infinity })
+            ]).then(() => {
+              return User.upsert({ name: 'user1', address: 'address' });
+            }).then(() => {
+              return User.findAll({
+                where: { address: null }
+              });
+            }).then(users => {
+              expect(users).to.have.lengthOf(2);
+            });
+          });
+        });
+      }
     });
   }
 });
--- a/lib/model.js
+++ b/lib/model.js
@@ -2188,6 +2188,7 @@ class Model {
     options = _.extend({
       hooks: true
     }, Utils.cloneDeep(options || {}));
+    options.model = this;
 
     const createdAtAttr = this._timestampAttributes.createdAt;
     const updatedAtAttr = this._timestampAttributes.updatedAt;


test/integration/model/upsert.test.js
```diff
@@ -487,6 +487,34 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+            }
+          }, {
+            paranoid: true
+          });
+
+          return User.sync({ force: true }).then(() => {
+            return User.build({ name: 'user1' })
+              .save()
+              .then(() => {
+                return User.upsert({ name: 'user1', address: 'address' }, {logging: console.log});
```
eseliger: Is the logging important for the test? Otherwise it just spams the mocha output

gabegorelick: Good catch, I'll remove.

lib/dialects/postgres/query-generator.js
```diff
@@ -357,7 +357,7 @@ const QueryGenerator = {
 
   upsertQuery(tableName, insertValues, updateValues, where, model, options) {
     const insert = this.insertQuery(tableName, insertValues, model.rawAttributes, options);
-    const update = this.updateQuery(tableName, updateValues, where, options, model.rawAttributes);
+    const update = this.updateQuery(tableName, updateValues, where, Object.assign({ model }, options), model.rawAttributes);
```
sushantdhiman: I think `Model.upsert` should set it, like other methods do example [destroy](https://github.com/sequelize/sequelize/blob/6adfabf1d6f3bb9986bfebba4d966228f24e553f/lib/model.js#L2463)

Please add `options.model = this` around [here](6adfabf1d6f3bb9986bfebba4d966228f24e553f) and remove this change.

test/integration/model/upsert.test.js
```diff
@@ -487,6 +487,34 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+            }
+          }, {
+            paranoid: true
+          });
+
+          return User.sync({ force: true }).then(() => {
+            return User.build({ name: 'user1' })
+              .save()
+              .then(() => {
+                return User.upsert({ name: 'user1', address: 'address' });
```
sushantdhiman: Test should be more expressive. Add three records (dont use build, just use Promise.all / create)

1) No deletedAt set
2) deletedAt set to Infinity
3) deletedAt set to -Infinity

assert 2 records are returned

gabegorelick: @sushantdhiman Updated. Let me know if that's what you had in mind.

PR #8401, opened 2017-09-30
added indexHint

felipefdl:
### Pull Request check-list

### Description of change

Adding new feature, indexHint, where we can select which index we wish use. 

The documentation is from: https://dev.mysql.com/doc/refman/5.7/en/index-hints.html

### Examples:

```javascript
{
  where: { id: 'myid' },
  indexHint: [['USE', 'id']],
}
```

```javascript
{
  where: { id: 'myid' },
  indexHint: [[['USE', 'id'], ['USE', 'created_at']]],
}
```

```javascript
{
  where: { id: 'myid' },
  indexHint: [['FORCE', 'id']],
}
```

```javascript
{
  where: { id: 'myid' },
  indexHint: [['IGNORE', 'id']],
}
```

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1120,6 +1120,26 @@ const QueryGenerator = {
       mainQueryItems.push(mainJoinQueries.join(''));
     }
 
+    // Add IndexHint to sub or main query
+    if (options.hasOwnProperty('indexHint') && Array.isArray(options.indexHint)) {
+      _.each(options.indexHint, (value) => {
+        let line = ' ';
+        if (value.hasOwnProperty('val')) {
+          line += value.val;
+        } else {
+          line += String(value[0]).toUpperCase();
+          line += ' INDEX(`' + String(value[1]) + '`)';
+          if (value[2]) line += ' ' + String(value[2]);
+        }
+
+        if (subQuery) {
+          subQueryItems.push(line);
+        } else {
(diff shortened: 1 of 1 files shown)


eseliger:
Please add some tests for this feature

sushantdhiman:
I dont think we should add this feature. Reasons 

1) I think only MySQL and MSSQL supports it.
2) This feature is not demanded by many users and will complicate query generator.
3) If you really want some query to be fine tuned for optimum performance I would recommend write that query yourself as per your database / indexes conditions. That will always beat ORM generated SQL which have to support many features and criterias.

felipefdl:
@sushantdhiman We have 19k of tables, almost 3 billions of records. The mysql system cannot choose right index for my queries, we have a lot combinations of index, dependents of the query is. So basically you saying the sequelize is not able to handle big data systems. So, I will keeping the for of sequelize with these modifications and I should keep updating the lib always when you guys release a new version, I cannot understand why sequelize doesn't  support the parameters so useful for big data softwares.

sushantdhiman:
In past I have rejected this feature because there was little to no demand, but it seems now this feature is desired by some users.

There are three PRs related to add this change #9378, #9409 & #8401

I suggest authors of these PRs or anyone else who want this feature to suggest an API #9421 and propose a single PR after API is finalized.

Looks like MySQL / MSSQL / SQLite support this feature. PR with support and tests for all these dialects will be appreciated

PR #8408, opened 2017-10-02
fix(errors): spurious Bluebird warnings

gabegorelick:
### Pull Request check-list

- [x] Have you added new tests to prevent regressions? **N/A, unless you think it's worth it**
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? **N/A shouldn't modify any API**

### Description of change

The way Sequelize instantiates some of its Error subclasses confuses Bluebird into thinking that Promises are rejected with non-Error objects, so you get warnings when running with `BLUEBIRD_DEBUG` enabled.

See https://github.com/petkaantonov/bluebird/issues/990

### DIFF
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -11,7 +11,6 @@ class BaseError extends Error {
   constructor(message) {
     super(message);
     this.name = 'SequelizeBaseError';
-    this.message = message;
     Error.captureStackTrace(this, this.constructor);
   }
 }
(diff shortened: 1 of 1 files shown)


lib/errors/index.js
```diff
@@ -11,7 +11,6 @@ class BaseError extends Error {
   constructor(message) {
     super(message);
     this.name = 'SequelizeBaseError';
-    this.message = message;
```
gabegorelick: The super class should already set `this.message = message`. The only slight API change is if clients relied on Sequelize's behavior of **not** defaulting `message` to '', which Error does.

sushantdhiman: ECMAScript specs says error messages should default to empty string where we are setting them to `undefined`.

Bluebird's error check also conforms to this, but it would be better if they also include `instance of` check as undefined error message very much is possible in everyday code.

PR #8426, opened 2017-10-04
Correct bulk update and bulk destroy examples.

dinusuresh:
bulkCreate is used in the examples of bulk update and bulk destroy. Correct the examples to reflect respective function names for bulk update and bulk destroy.

### DIFF
--- a/docs/instances.md
+++ b/docs/instances.md
@@ -155,7 +155,7 @@ User.bulkCreate([
 To update several rows at once:
 
 ```js
-Task.bulkCreate([
+Task.update([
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Examples are correct, `bulkCreate` is used to create sample data. Later in `.then` actual `update / destroy` method is used

dinusuresh:
Right. I did not notice that, forgive my ignorance.

sushantdhiman:
@dinusuresh No problem, thanks for contributing

Issue #6834, opened 2016-11-10
`Sequelize.useCLS()` breaks with native async/await


leebenson:
In node 7 with `--harmony-async-await`, transactions hang when using `Sequelize.useCLS()`.

Calling `.rollback` or `.commit` explicitly has the same effect- nothing happens.

manuelzamboni:
async and await are based on native ES6 promises but Sequelize.useCLS dependes on BlueBird promises to implement its behaviour.

felixfbecker:
Could you do `global.Promise = require('sequelize').Promise` to make it work? That way `async`/`await` would use Sequelize promises and it shouldn't cause a problem since Bluebird is compatible with native promises.

leebenson:
> Could you do global.Promise = require('sequelize').Promise to make it work? 

That should work, although requiring an overwrite of native objects to make something work feels a bit hacky.

jeremija:
It does not seem that using `global.Promise = require('sequelize').Promise` works - async functions still return native Promises:
```bash
$ node --version
v7.10.0
```

```javascript

$ node
> NativePromise = Promise;
[Function: Promise]
> global.Promise = require('sequelize').Promise
> test = async () => {}
> test() instanceof Promise
false
> test() instanceof NativePromise
true
```

PR #8430, opened 2017-10-05
fix(transactions): don't set transaction on continuation storage when…

### LINKED ISSUE #8429: Manually managed transactions set namespace value even when not executing under Sequelize._clsRun
## What are you doing?
My solution is a multi-tenant sequelize-based server, and nested transactions with CLS enabled are leaking between transactions.

Consider the following:

```js
const namespace = cls.createNamespace('tenant');
const metadataDatabase = new Sequelize(...);
const tenantDatabase = new Sequelize(...);

Sequelize.useCLS(namespace);

await tenantDatabase.transaction(async () => {
  const metadataTransaction = await metadataDatabase.transaction();
  
   await metadatabase.create({...}, { transaction: metadataTransaction });
   await metadataTransaction.commit();

   await tenantDatabase.create({...});
});
```

In this example, the call to `tenantDatabase` will attempt to use the metadata's transaction as the outer one was overridden on the namespace by the manually managed transaction.  Being that `Sequelize._clsRun` is almost never executed to wrap a manually managed transaction, the issue hasn't presented itself.

In Sequelize, if the namespace exists the transaction is unconditionally put there, and immediately lost afterwards because `Sequelize._clsRun` is not used to bind the continuation context.  In this case, the outer managed transaction is running the continuation chain, hence we get queries that executte under another connection.

I am preparing a patch which does not attempt to override the continuation storage's transaction object when constructing a manual transaction, and associated tests now.

The query for `tenantDatabase` is executed under the query for `metadataDatabase` erroneously.

__Dialect:__  any
__Dialect version:__ ???
__Database version:__ sqlite3
__Sequelize version:__ 4.13.5

tylerjwatson:
… manually managed

### Pull Request check-list

### Description of change

This patch prevents Sequelize from adding a transaction to the continuation storage namespace when the transaction is manually managed.  Traditionally `Transaction#prepareEnvironment` would unconditionally set the transaction on the continuation storage, but is sunk immediately because manual transactions are never run under `Sequelize._clsRun`.  

Has associated test.  Closes #8429.

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -99,7 +99,7 @@ class Transaction {
       });
   }
 
-  prepareEnvironment() {
+  prepareEnvironment(useCLS = true) {
     let connectionPromise;
 
     if (this.parent) {
@@ -125,7 +125,7 @@ class Transaction {
         throw setupErr;
       }))
       .tap(() => {
-        if (this.sequelize.constructor._cls) {
+        if (useCLS && this.sequelize.constructor._cls) {
           this.sequelize.constructor._cls.set('transaction', this);
         }
         return null;
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -99,8 +99,12 @@ class Transaction {
       });
   }
 
-  prepareEnvironment(useCLS = true) {
+  prepareEnvironment(useCLS) {
     let connectionPromise;
+
+    if (typeof useCLS === 'undefined') {
+      useCLS = true;
+    }
 
     if (this.parent) {
       connectionPromise = Utils.Promise.resolve(this.parent.connection);
--- a/test/integration/cls.test.js
+++ b/test/integration/cls.test.js
@@ -33,6 +33,18 @@ if (current.dialect.supports.transactions) {
     });
 
     describe('context', () => {
+      it('does not use continuation storage on manually managed transactions', function() {
+        const self = this;
+
+        return Sequelize._clsRun(() => {
+          return this.sequelize.transaction()
+          .then(transaction => {
+            console.error(self.ns.get('transaction'));
+            expect(self.ns.get('transaction')).to.be.undefined;
+          });
+        });
+      });
+
       it('supports several concurrent transactions', function() {
         let t1id, t2id;
         const self = this;
--- a/test/integration/cls.test.js
+++ b/test/integration/cls.test.js
@@ -37,9 +37,7 @@ if (current.dialect.supports.transactions) {
         const self = this;
 
         return Sequelize._clsRun(() => {
-          return this.sequelize.transaction()
-          .then(transaction => {
-            console.error(self.ns.get('transaction'));
+          return this.sequelize.transaction().then(() => {
             expect(self.ns.get('transaction')).to.be.undefined;
           });
         });
--- a/test/integration/cls.test.js
+++ b/test/integration/cls.test.js
@@ -37,8 +37,9 @@ if (current.dialect.supports.transactions) {
         const self = this;
 
         return Sequelize._clsRun(() => {
-          return this.sequelize.transaction().then(() => {
+          return this.sequelize.transaction().then(transaction => {
             expect(self.ns.get('transaction')).to.be.undefined;
+            return transaction.rollback();
           });
         });
       });
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -962,7 +962,7 @@ class Sequelize {
 
     const transaction = new Transaction(this, options);
 
-    if (!autoCallback) return transaction.prepareEnvironment().return(transaction);
+    if (!autoCallback) return transaction.prepareEnvironment(false).return(transaction);
 
     // autoCallback provided
     return Sequelize._clsRun(() => {


test/integration/cls.test.js
```diff
@@ -33,6 +33,18 @@ if (current.dialect.supports.transactions) {
     });
 
     describe('context', () => {
+      it('does not use continuation storage on manually managed transactions', function() {
+        const self = this;
+
+        return Sequelize._clsRun(() => {
```
sushantdhiman: Fix indentation, its breaking build

test/integration/cls.test.js
```diff
@@ -33,6 +33,18 @@ if (current.dialect.supports.transactions) {
...
+          return this.sequelize.transaction()
+          .then(transaction => {
+            console.error(self.ns.get('transaction'));
```
sushantdhiman: Remove this line

test/integration/cls.test.js
```diff
@@ -33,6 +33,18 @@ if (current.dialect.supports.transactions) {
     });
 
     describe('context', () => {
+      it('does not use continuation storage on manually managed transactions', function() {
+        const self = this;
+
+        return Sequelize._clsRun(() => {
+          return this.sequelize.transaction()
+          .then(transaction => {
```
sushantdhiman: transaction is not used

tylerjwatson:
Ah true, I'll fix that up shortly.

test/integration/cls.test.js
tylerjwatson: I'm confused a bit here, the indentation looks fine.  Can you please point me to where the rules are being violated?

sushantdhiman: just run `npm run lint` you will know

lib/transaction.js
```diff
@@ -99,7 +99,7 @@ class Transaction {
       });
   }
 
-  prepareEnvironment() {
+  prepareEnvironment(useCLS = true) {
```
sushantdhiman: Oh, default variable is not supported in Node 4/6

tylerjwatson:
I tested using the sqlite dialect fully and it appears its okay.  Perhaps some mssql specific stuff?

test/integration/cls.test.js
```diff
@@ -33,6 +33,16 @@ if (current.dialect.supports.transactions) {
     });
 
     describe('context', () => {
+      it('does not use continuation storage on manually managed transactions', function() {
+        const self = this;
+
+        return Sequelize._clsRun(() => {
+          return this.sequelize.transaction().then(() => {
```
janmeier: Remember to roll back the transaction, I think that's what's causing some of the other tests to hang

tylerjwatson: Oooh, good point.  I'll fix that up right away.

tylerjwatson:
I've never had to use `cls.get('transaction')`, as that would not be applicable for my use case anyway, I can't speak for everybody else.  If the `uuid` property exists publicly on a Sequelize instance, or, we make a `sequelize.clsTransaction` getter which will return the transaction from the storage with the proper key, I think nobody is going to be left out.

janmeier:
You are probably right, hopefully CLS means you should never have to worry about the transaction

lib/transaction.js
```diff
@@ -125,7 +129,7 @@ class Transaction {
         throw setupErr;
       }))
       .tap(() => {
-        if (this.sequelize.constructor._cls) {
+        if (useCLS && this.sequelize.constructor._cls) {
```
sushantdhiman: `this.sequelize.constructor._cls` should not set if you dont call `useCls`, can you verify if your test fails without this fix ?

tylerjwatson:
The patch works for my usecase and does indeed fix the immediate problem, but I do think it's worth investigating how to isolate transaction storage to sequelize instances.  The patch is still valid, but I think I can do better.  It will be some work.

tylerjwatson:
Just as a side question, would you guys ever consider just pulling in `continuation-local-stoage` internally for the transaction context passing so it's always transparent and out of the way, and doing away with such APIs like `Sequelize.useCLS`?  We can allow people to override the namespace which is set if they want to, but it just uses a `sequelize` namespace by default?

janmeier:
> Just as a side question, would you guys ever consider just pulling in continuation-local-stoage internally

Probably not:
* It's using an unofficial API (async-listener, https://github.com/node-forward/discussions/issues/28) - not that there are any better alternatives currently, but there might soon be with async-hooks https://github.com/Jeff-Lewis/cls-hooked
* It has a performance impact (async-listener, + having to patch bluebird to preserve context). So we only want to enable it if the user explicitly requests it.

tylerjwatson:
Thought as much.  I'll continue as normal.

lib/transaction.js
sushantdhiman: Ok understood, I thought we were saving env without cls

jeremija:
> It's using an unofficial API (async-listener, node-forward/discussions#28) - not that there are any better alternatives currently, but there might soon be with async-hooks https://github.com/Jeff-Lewis/cls-hooked

Hi @janmeier, could you share what's the reason against using [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked) at the moment? Is it still too early and experimental? On the first try it seems to be working better than the original CLS since `async/await` expressions just work. This is related to #6834. Thanks!

janmeier:
@jeremija I haven't experimented with it, you are welcome to do so and provide a PR :)

PR #8450, opened 2017-10-10
feat(sqlite/query-generator): sqlite update queries now support limit via rowid

netsgnut:
### Pull Request check-list

### Description of change

By default SQLite is compiled with `SQLITE_ENABLE_UPDATE_DELETE_LIMIT` off. That means the `UPDATE` statements would not support an optional `LIMIT` clause. By using `UPDATE ... WHERE rowid IN (SELECT rowid FROM ... WHERE ... LIMIT n)` instead it can provide support in environments without such feature.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2590,7 +2590,7 @@ class Model {
    * @param  {Boolean}      [options.sideEffects=true] Whether or not to update the side effects of any virtual setters.
    * @param  {Boolean}      [options.individualHooks=false] Run before / after update hooks?. If true, this will execute a SELECT followed by individual UPDATEs. A select is needed, because the row data needs to be passed to the hooks
    * @param  {Boolean}      [options.returning=false]       Return the affected rows (only for postgres)
-   * @param  {Number}       [options.limit]                 How many rows to update (only for mysql and mariadb, implemented as TOP(n) for MSSQL)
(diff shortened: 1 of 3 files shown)


netsgnut:
On a related note, it seems `DELETE` can be handled the same manner as well. Would you like that feature to be submitted as another PR, or as another commit on the same branch as this one? Thanks.

lib/model.js
```diff
@@ -2590,7 +2590,7 @@ class Model {
    * @param  {Boolean}      [options.sideEffects=true] Whether or not to update the side effects of any virtual setters.
    * @param  {Boolean}      [options.individualHooks=false] Run before / after update hooks?. If true, this will execute a SELECT followed by individual UPDATEs. A select is needed, because the row data needs to be passed to the hooks
    * @param  {Boolean}      [options.returning=false]       Return the affected rows (only for postgres)
-   * @param  {Number}       [options.limit]                 How many rows to update (only for mysql and mariadb, implemented as TOP(n) for MSSQL)
+   * @param  {Number}       [options.limit]                 How many rows to update (only for mysql and mariadb, implemented as TOP(n) for MSSQL; for sqlite it is supported only when rowid is present)
```
sushantdhiman: isnt rowId always available?

sushantdhiman:
@netsgnut A different PR for delete works

lib/model.js
netsgnut: It is enabled by default, but can be absent if created with `WITHOUT ROWID`. Ref: https://sqlite.org/withoutrowid.html

Issue #7344, opened 2017-03-08
Limit is broken for findAll and findAndCountAll with include


Alexsey:
Consider we have a two simple models "Events" and "Tags" with many-to-many relation and we want to get all events with specific tag. And we want to limit maximum number of response items.

(Consider you are using ```NodeJS``` 7.6+ with ```async/await``` or ```Babel``` with ```async/await``` plugin)
```js
'use strict'

const Sequelize = require('sequelize')

const dbConfig = {
    type: 'mysql', user: 'root', password: 'root',
    database: 'db', host: 'localhost'
}

;(async () => {
    const sequelize = new Sequelize(
        dbConfig.database, dbConfig.user, dbConfig.password, {
            host: dbConfig.host,
            dialect: dbConfig.type,
            pool: {max: 100, min: 0, idle: 10000},
            dialectOptions: {
                charset: 'utf8_general_ci',
            }
        }
    )
    await sequelize.authenticate()

    const Events = sequelize.define('events', {
        id: {type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true},
        title: {type: Sequelize.STRING}
    }, {timestamps: false})

    const Tags = sequelize.define('tags', {
        name: {type: Sequelize.STRING}
    }, {timestamps: false})

    const EventsTags = sequelize.define('events_tags', {}, {timestamps: false})

    Events.belongsToMany(Tags, {through: EventsTags})
    Tags.belongsToMany(Events, {through: EventsTags})

    await sequelize.sync()

    await Promise.all([
        Events.bulkCreate([
            {title: 'event_1'}, {title: 'event_2'},
            {title: 'event_3'}, {title: 'event_4'}
        ]),
        Tags.bulkCreate([
            {name: 'tag_1'}, {name: 'tag_2'},
            {name: 'tag_3'}, {name: 'tag_4'},
        ])
    ])
    await EventsTags.bulkCreate([
        {eventId: 1, tagId: 1}, {eventId: 2, tagId: 2},
        {eventId: 3, tagId: 3}, {eventId: 4, tagId: 1}
    ])

    const eventsWithTag1 = await Events.findAndCountAll({
        limit: 3,
        include: [{
            model: Tags,
            where: {name: 'tag_1'}
        }]
    })

    console.log('found events with tag_1:', eventsWithTag1.rows.length) // 1
    console.log('count events with tag_1:', eventsWithTag1.count) // 2

    await sequelize.close()
})()
```
This query will found only one event:
[...]

PR #8458, opened 2017-10-10
Resolve limit params being ignored on includes queries

austinksmith:
### Pull Request check-list

- [ Yes (Double Check on Your End)] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [Yes ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [No ] Have you added new tests to prevent regressions?
- [ No] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [Idk ] Did you follow the commit message conventions explained in CONTRIBUTING.md?

### Description of change

There is an outstanding issue since at least march where a query that includes other tables incorrectly forms the final sql output and sets the limit param on the subquery instead of the main query which then results in inconsistent #'s of rows being returned. 

https://github.com/sequelize/sequelize/issues/7344

# Apologize for the multiple Pr's my local git was messing up on me wanted to ensure I wasnt including extra changes beyond the minimum required

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1172,11 +1172,7 @@ const QueryGenerator = {
     // Add LIMIT, OFFSET to sub or main query
     const limitOrder = this.addLimitAndOffset(options, mainTable.model);
     if (limitOrder && !options.groupedLimit) {
-      if (subQuery) {
-        subQueryItems.push(limitOrder);
-      } else {
-        mainQueryItems.push(limitOrder);
-      }
+      mainQueryItems.push(limitOrder);
     }
 
     if (subQuery) {


austinksmith:
Unfortunately after testing this "fix" further it does not resolve the issues with counts as expected, it seems very intermittent when sequelize wants to obey the limit param and this query generator change is a small part of the overall problem I'm seeing. Includes with a limit defined is still broken....maintainers please let me know your thoughts, thanks.

sushantdhiman:
Looks like a version of https://github.com/sequelize/sequelize/pull/8278

Maybe a good start will be to submit a failing test case and then find out what we can do to fix that issue

Issue #8471, opened 2017-10-13
Postgres range inclusive property lost in JSON format


kingjerod:
## What are you doing?
Querying a Postgres table with a daterange column and returning it as JSON.

```js
const Timeline = Sequelize.define({
   id: {type: Sequelize.INTEGER, primaryKey: true},
   date_range: {type: Sequelize.RANGE(Sequelize.DATEONLY)}
});
const range = [
   {value: new Date(Date.UTC(2017, 0, 1)), inclusive: true},
   {value: new Date(Date.UTC(2017, 6, 1)), inclusive: false},
];
await Timeline.create({id: 1, date_range: range});
const row = await Timeline.findById(1);
console.log(JSON.stringify(row)); // No inclusive property
```

## What do you expect to happen?
When returning the data as JSON, the inclusive data should not be lost. 

In the docs, there is this optional format for updating, this would be better because it can be turned into JSON without losing any information:
```js
const range = [
  { value: '2017-01-01', inclusive: false },
  { value: '2017-07-01', inclusive: true },
];
```

## What is actually happening?
The inclusive property is lost. This is because the inclusive property is set on an array, and array properties are not converted into JSON (like Array.length).

```js
{id:1, date_range: ['2017-01-01', '2017-07-01']}
```

__Dialect:__ postgres
__Dialect version:__ 9.6
__Database version:__ 9.6
__Sequelize version:__ 4.4.3
__Tested with master branch:__ Yes

sushantdhiman:
Fixed by https://github.com/sequelize/sequelize/pull/9364

PR #8472, opened 2017-10-13
Add toJSON to Postgres range so inclusiveness is not lost

kingjerod:
### Pull Request check-list

I tried to run the tests but it is throwing an "undefined" error before it even gets to my new test:

conemu64_2017-10-12_22-11-38

### Description of change

This is possibly a BC break if people are relying on the JSON format to just be dates. I could alter it so the first two entries in the array are the dates for BC? As it is right now, the date range is missing the inclusiveness which is like using a time without knowing what timezone it is. 

Issue this fixes: https://github.com/sequelize/sequelize/issues/8471


sushantdhiman:
Moved https://github.com/sequelize/sequelize/pull/9364

PR #8474, opened 2017-10-13
fix(model): unhandled rejection in findAndCount

PsiXdev:
### Pull Request check-list

### Description of change

Before my changes the findAndCount method of model could throw an unhandled rejection. This happens, for example, when I refer to a non-existent column. In this case, countQuery.catch is correctly handled, but the error from the findQuery becomes unhandled (both promises start to be executed in parallel).

Sorry, but I have not written any new tests, because the tests for findAndCount just missing.

For me it seems strange the previous logic:
`if (count === 0)` - useless `if` as for me
`count || 0` - count can be `undefinde` or `null`?

But I'm not sure that it's wrong. May be I made some mistakes in my correction?

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1857,21 +1857,13 @@ class Model {
       countOptions.attributes = undefined;
     }
 
-    const countQuery = this.count(countOptions);
-    const findQuery = this.findAll(options);
-
-    return countQuery.then(count => {
-      if (count === 0) {
-        return {
-          count: 0,
-          rows: []
-        };
-      }
-      return findQuery.then(results => ({
-        count: count || 0,
-        rows: results
-      }));
-    });
+    return Promise.all([
+      this.count(countOptions),
+      this.findAll(options)
+    ]).spread((count, rows) => ({
+      count,
+      rows
+    }));
   }
 
   /**
--- a/test/unit/model/findandcount.test.js
+++ b/test/unit/model/findandcount.test.js
@@ -1,0 +1,46 @@
+'use strict';
+
+const chai = require('chai'),
+  expect = chai.expect,
+  Support = require(__dirname + '/../support'),
+  current = Support.sequelize,
+  sinon = require('sinon'),
+  DataTypes = require(__dirname + '/../../../lib/data-types'),
+  Promise = require('bluebird').getNewLibraryCopy();
+
+describe(Support.getTestDialectTeaser('Model'), () => {
+  describe('method findAndCount', () => {
+    describe('should handle promise rejection in findAndCount', () => {
+      before(() => {
+        this.oldFindAll = current.Model.findAll;
+        this.oldCount = current.Model.count;
+
+        current.Model.count = () => Promise.reject(new Error());
+        current.Model.findAll = () => Promise.reject(new Error());
+
+        this.stub = sinon.stub();
+
+        Promise.onPossiblyUnhandledRejection(() => {
+          this.stub();
+        });
+
+        this.User = current.define('User', {
+          username: DataTypes.STRING,
+          age: DataTypes.INTEGER
+        });
+      });
+
+      after(() => {
+        current.Model.findAll = this.oldFindAll;
+        current.Model.count = this.oldCount;
+      });
+
+      it('with errors in count and findAll both', () => {
+        return this.User.findAndCount({})
+          .catch(() => {
+            expect(this.stub.callCount).to.eql(0);
+          });
+      });
+    });
+  });
+});


lib/model.js
```diff
@@ -1857,21 +1857,13 @@ class Model {
       countOptions.attributes = undefined;
     }
 
-    const countQuery = this.count(countOptions);
-    const findQuery = this.findAll(options);
-
```
sushantdhiman: Both promise start executing as soon as possible, but we still want to keep `count === 0` logic

PsiXdev: You mean:
```
if (count === 0) {
	return {count, rows: []};
} else {
	return {count, rows};
}
```
? Or do you want to see findQuery.then inside countQuery.then?
I do not understand the meaning

PsiXdev:
I added a breaking test. It does not pass in the current `master` branсh and passes in code with my corrections.
Since the tests use a very old version of the sinon, I could not use it to replace `count` and `findAll` functions. `stub().returns(Promise.reject(new Error())` instantly thrown in `onPossiblyUnhandledRejection`. So I replaced them for my functions.
Hopefully this will help to understand that the current code is not working correctly.

PsiXdev:
I apologize, but your library is **broken**. Will there be any steps to fix it?

lib/model.js
```diff
@@ -1857,21 +1857,13 @@ class Model {
...
-    return countQuery.then(count => {
```
sushantdhiman: This used to be a performance improvement if count is zero then we can resolve empty array in rows and count to zero. I would like to keep that behavior

test/unit/model/findandcount.test.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
```
sushantdhiman: This test is confusing due to mocks and I am not very happy with `onPossiblyUnhandledRejection`

Why dont we add an integration test instead.

lib/model.js
```diff
@@ -1857,21 +1857,13 @@ class Model {
...
-      if (count === 0) {
-        return {
-          count: 0,
-          rows: []
-        };
-      }
-      return findQuery.then(results => ({
```
sushantdhiman: `findQuery` should reject with error promise, which user then should be able to handle in their code. I suspect it could rejecting synchronously?

PsiXdev: User calls findAndCount and he able to handle result in then/catch (he don't should be able handle countQuery and findQuery separately). And my code allows it to do. I do not understand what the problem is. could you explain in more detail?

lib/model.js
PsiXdev: First of all, it is obvious that the current solution does not work, regardless of performance. It should be fixed.
Besides, I believe that the findQuery will be executed very quickly if count is zero and there is no point in talking about reducing productivity.
Perhaps if this worries you, you have the convenient means to test performance?

iamjochem:
the following commit caused the problem (the "find" promise is unhandled in the case that count ends up being zero):

https://github.com/sequelize/sequelize/pull/6696

I don't think it should have been merged, the performance issues is secondary to correctness, besides which `findAndCountAll` is a convienience method ... if some one really needs optimal performance (i.e. parallel running queries) then all the pieces are already available to implement that in a simple custom model method ... not to mention that if sequelize performance is really the bottleneck then maybe one should be dropping down to hand-crafted raw queries.

 the OP for #6696 mentioned dropping from 30s to 15s when parallelizing the 2 queries, all well and good in principle but in practice a query that take 15 seconds is really an indication that the underlying DB schema needs to be improved (smells like a serious lack of indexes) ... any query done in response  to a user request should be way way faster, reporting queries (run in background tasks) may be really heavy/slow but that is not a inherent problem.

@sushantdhiman - I believe the `count === 0` optimization (an artifact left over from my original implementation of `findAndCountAll`) is pointless if your running the queries in parallel - the DB is already doing the work for the "find" query and nodejs is eating cycles processsing the reponse, all you gain is a slight win in promise resolution for that particular call ... but your application is still paying the underlying cost; besides which the `count === 0` optimization causes the unahdnled exception scenario (I suppose you could consider adding a "catch" handler to the find-query promise inside the `count === 0` where you log any errors.

personally I would opt for either:

1. running in parallel and forgoing the `count === 0` optimization
2. running in series and doing the `count === 0` optimization

> for consideration - running in parallel is putting extra pressure on the DB, which is likely to be something you don't want if the queries in question are really slow/heavy.

@sushantdhiman - I can't really comment on the implementation of the unit test but I think @PsiXdev's actual fix is functionally correct.

test/unit/model/findandcount.test.js
```diff
@@ -0,0 +1,46 @@
...
+  current = Support.sequelize,
+  sinon = require('sinon'),
+  DataTypes = require(__dirname + '/../../../lib/data-types'),
+  Promise = require('bluebird').getNewLibraryCopy();
+
+describe(Support.getTestDialectTeaser('Model'), () => {
+  describe('method findAndCount', () => {
+    describe('should handle promise rejection in findAndCount', () => {
+      before(() => {
+        this.oldFindAll = current.Model.findAll;
```
sushantdhiman: Use sinon stub, and reset them after each test, restore at the end

PR #8478, opened 2017-10-13
Fixed is operator for undefined value

### LINKED ISSUE #8476: Invalid syntax for undefined value
Hello. I've find small but important issue.

Please open `lib/dialects/abstract/query-generator.js`
```
return this._joinKeyValue(key, this.escape(value, field), value === null ? this.OperatorMap[Op.is] : this.OperatorMap[Op.eq], options.prefix);
```

I am doing select query:

```
.findOne({
  where: {
    a: undefined
  }
})
```

It will generate:
```
SELECT ... WHERE "table_name"."a" = NULL LIMIT 1
```

It is wrong, we should use **is** operator for `NULL`. I will provide a pull request soon.

andrew-aladev:
Closes #8476. It is a small pr that fixes `is` operator for undefined value.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -2032,7 +2032,7 @@ const QueryGenerator = {
     }
 
     if (!value) {
-      return this._joinKeyValue(key, this.escape(value, field), value === null ? this.OperatorMap[Op.is] : this.OperatorMap[Op.eq], options.prefix);
+      return this._joinKeyValue(key, this.escape(value, field), (typeof value === 'undefined' || value === null) ? this.OperatorMap[Op.is] : this.OperatorMap[Op.eq], options.prefix);
     }
 
     if (value instanceof Utils.SequelizeMethod && !(key !== undefined && value instanceof Utils.Fn)) {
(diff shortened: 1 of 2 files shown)


sushantdhiman:
Hmm I am not sure if `undefined` should be treated as `null`, I think it should not. If you want to compare against `IS NULL` do something like this `yourField = yourField || null`, rather than this implicit conversion you are suggesting.

@sequelize/orm

janmeier:
Agreed, I don't think undefined should be coerced to null.

However the current behavior is not exactly intuitive either. Not sure what should happen here. Throwing on unexpected input would require a major

eseliger:
Agree, too. This is dangerously

Alexsey:
Maybe now, when v5 is in Beta is the right moment to fix it with some breaking change? Because the generation of a request with `= NULL` just makes no sense

Alexsey:
BTW here https://github.com/sequelize/sequelize/issues/8583 people are complaining about the behavior that is proposed in this PR. But now I can see that for SQL Servier it creates `= NULL`. May it be that we have different behavior for different dialects?

PR #8485, opened 2017-10-14
fix(paranoid): use Utils.now to query for paranoid records

### LINKED ISSUE #8006: Soft deleted rows are being returned in findAll using paranoid mode
## What you are doing?
Soft deletion of row sets `deletedAt` column with ```utils.now()``` ([look here](https://github.com/sequelize/sequelize/blob/master/lib/model.js#L2473)). That means that the value would be machine's current timestamp. But when you'll try to retrieve data, the query for ignoring soft deleted rows would be using `CURRENT_TIMESTAMP` ([look here](https://github.com/sequelize/sequelize/blob/master/lib/model.js#L81)) which is the DB's machine timestamp. If these two machines (Server and DB) clocks are not synced you'll end up with soft deleted rows included in your results. 

```js
const res = model.findAll(...)
```

Shouldn't get soft deleted rows at all. The value that should be set upon deletion is `CURRENT_TIMESTAMP` as well.

I'm getting rows which I've deleted.

__Dialect: postgres 
__Database version: 9.5
__Sequelize version: 4.3.2

### LINKED ISSUE #8202: Same second query for paranoid tables returning soft deleted records
## What you are doing?

```js
sequelize.define('user', {
  //...column definitions
}, {
  paranoid: true
});

User.find({
  where: {
    id: req.params.id
  }
}).then((user) => {
  console.log(user);
  user.destroy().then(() => {
    User.find({
      where: {
        id: req.params.id
      }
    }).then((deletedUser) => {
      console.log(deletedUser);
      console.log(new Date(new Date().toUTCString()).toISOString());
    });
  });
});

//run a second later
User.find({
  where: {
    id: req.params.id
  }
}).then((user) => {
  console.log(user);
});
```

Second console.log should return null.

Second console.log is returning the object again with a deletedAt field entered that matches the third console.log's output.  Remote MySQL server runs in UTC time.  Occurs on local development stations in EST as well as servers running in UTC time.  Similar to #7204.

Running another query a second later will return null.

__Dialect:__ mysql
__Database version:__ 5.7.11
__Sequelize version:__ 4.7.5

sushantdhiman:
### Pull Request check-list

### Description of change

Fixes #8202
Fixes #8006
Closes #8204
Closes #7787

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -77,9 +77,16 @@ class Model {
     const deletedAtCol = model._timestampAttributes.deletedAt;
     const deletedAtAttribute = model.rawAttributes[deletedAtCol];
     const deletedAtObject = {};
+    const now = Utils.now(this.sequelize.options.dialect);
+
     let deletedAtDefaultValue = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
 
-    deletedAtDefaultValue = deletedAtDefaultValue || { [Op.or]: { [Op.gt]: model.sequelize.literal('CURRENT_TIMESTAMP'), [Op.eq]: null } };
+    deletedAtDefaultValue = deletedAtDefaultValue || {
+      [Op.or]: {
+        [Op.gt]: now,
+        [Op.eq]: null
+      }
+    };
 
     deletedAtObject[deletedAtAttribute.field || deletedAtCol] = deletedAtDefaultValue;
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -391,7 +391,7 @@ exports.sliceArgs = sliceArgs;
 
 function now(dialect) {
   const now = new Date();
-  if (['mysql', 'postgres', 'sqlite'].indexOf(dialect) === -1) {
+  if (['mysql', 'postgres', 'sqlite', 'mssql'].indexOf(dialect) === -1) {
     now.setMilliseconds(0);
   }
   return now;
--- a/test/integration/instance.test.js
+++ b/test/integration/instance.test.js
@@ -17,6 +17,10 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
   });
 
   afterEach(function() {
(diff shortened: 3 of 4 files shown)


sushantdhiman:
Will also close https://github.com/sequelize/sequelize/pull/7787 but we dont have proper way to test that db / app server timezone difference.

If we distill what all those threads are saying, we can conclude 

1) `destroy` use `Utils.now` to set `deletedAt`
2) `_paranoidClause` use `CURRENT_TIMESTAMP` to query for those records

If db and app server use different timezone, this will surely get incorrect records.

I think we should use application server time, hence changing `_paranoidClause` to use `Utils.now` (Although a few of these threads suggest otherwise, i.e. use database server time in `destroy`)

eseliger:
@sushantdhiman This sounds good! I'm down to using application server time for the clause.
One reason: 
If other logic might create dates in the application as well, those should match dates in the database.

gabegorelick:
Correct me if I'm wrong, but I don't think this fundamentally fixes a lot of the linked issues because `Date` in JS is not a monotonic time source. In other words, it can move backwards (NTP, leap seconds etc) which would mess queries up. 

Happy to file an issue if you want to move discussion elsewhere.

sushantdhiman:
@gabegorelick Most of linked issue are appearing because querying and setting of deletedAt are using entirely different construct (new Date vs CURRENT_TIMESTAMP), which would always fail for cross timezone interactions.

With this PR both construct aligns to use singular source i.e. new Date. This will ensure paranoid records aren't returned even if clocks aren't synced, because now we are using only app server clock for all cases.

Leap seconds like issues are a bit hard to test and honestly I don't worry about them in current context which is failing paranoid for every cross timezone case.

But you are free to discuss this among other contributors by opening an issue if you like to

gabegorelick:
Thanks @sushantdhiman. We can continue discussion in https://github.com/sequelize/sequelize/issues/8496.

PR #8493, opened 2017-10-17
fix(findOrCreate): TypeError when default attribute unknown

### LINKED ISSUE #8492: `TypeError: Cannot read property 'field' of undefined` when findOrCreate encounters unique constraint violation and defaults has nonexistent attribute
## What are you doing?
```js
it('should error correctly when defaults contain a unique key and a non-existent field', function() {
    objectId: {
      type: DataTypes.STRING,
      unique: true
    },
    username: {
      type: DataTypes.STRING,
      unique: true
    }
  });

    return User.create({
      username: 'gottlieb'
    });
}).then(() => {
    return expect(User.findOrCreate({
      where: {
        objectId: 'asdasdasd'
      },
      defaults: {
        username: 'gottlieb',
        foo: 'bar' // field that's not a defined attribute
      }
    })).to.eventually.be.rejectedWith(Sequelize.UniqueConstraintError);
});
```

Test should pass

Test fails with TypeError.

__Dialect:__ any (but tested in PG)
__Dialect version:__ XXX
__Database version:__ XXX
__Sequelize version:__ master

gabegorelick:
Closes #8492

### Pull Request check-list

### Description of change

Fix for #8492: don't crash when passed an unknown attribute in `defaults` when `findOrCreate` returns a `UniqueConstraintError`.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2102,7 +2102,9 @@ class Model {
         const flattenedWhere = Utils.flattenObjectDeep(options.where);
         const flattenedWhereKeys = _.map(_.keys(flattenedWhere), name => _.last(_.split(name, '.')));
         const whereFields = flattenedWhereKeys.map(name => _.get(this.rawAttributes, `${name}.field`, name));
-        const defaultFields = options.defaults && Object.keys(options.defaults).map(name => this.rawAttributes[name].field || name);
+        const defaultFields = options.defaults && Object.keys(options.defaults)
+          .filter(name => this.rawAttributes[name])
+          .map(name => this.rawAttributes[name].field || name);
 
         if (defaultFields) {
           if (!_.intersection(Object.keys(err.fields), whereFields).length && _.intersection(Object.keys(err.fields), defaultFields).length) {
--- a/test/integration/model/create.test.js
+++ b/test/integration/model/create.test.js
@@ -104,6 +104,35 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       });
     });
 
+    it('should error correctly when defaults contain a unique key and a non-existent field', function() {
+      const User = this.sequelize.define('user', {
+        objectId: {
+          type: DataTypes.STRING,
+          unique: true
+        },
+        username: {
+          type: DataTypes.STRING,
+          unique: true
+        }
+      });
+
+      return User.sync({force: true}).then(() => {
+        return User.create({
+          username: 'gottlieb'
+        });
+      }).then(() => {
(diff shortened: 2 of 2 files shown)


lib/model.js
```diff
@@ -2102,7 +2102,9 @@ class Model {
         const flattenedWhere = Utils.flattenObjectDeep(options.where);
         const flattenedWhereKeys = _.map(_.keys(flattenedWhere), name => _.last(_.split(name, '.')));
         const whereFields = flattenedWhereKeys.map(name => _.get(this.rawAttributes, `${name}.field`, name));
-        const defaultFields = options.defaults && Object.keys(options.defaults).map(name => this.rawAttributes[name].field || name);
+        const defaultFields = options.defaults && Object.keys(options.defaults)
```
sushantdhiman: Rather than hiding this, we should just throw an error if default doesnt contain valid attributes.

gabegorelick: I understand why that would be useful, but I think being able to pass a superset of attributes is convenient. It's also consistent with how Sequelize tends to work now, e.g. `User.foo = 1` will just ignore `foo` if it's not an attribute.

gabegorelick: `Model.create({unknownAttribute: ...})` is a better example. I think if Sequelize doesn't throw an error in this case, it shouldn't for `Model.findOrCreate({defaults: {unknownAttribute: ...}})`.

sushantdhiman:
> v5, THROW or WARN if some unknown attribute is passed to create or findOrCreate like methods

sushantdhiman:
https://github.com/sequelize/sequelize/commit/b505723927305960003dae2a8b64e1f291a5f927

Mathiasduc:
_Not sure if here is the correct place to post this (i can do a new issue and reference this one)_

Hello, whenever i use includes on `Model#findOrCreate` with nested models default, i get this warning: ``
As seen in this example, i m not sure this should be an issue, is this an anti-pattern? it seems to work just fine to do multi model creations:
``` javascript
 db.models.Event.findOrCreate({
    where: {},
    include: [
      {
        model: db.models.Product,
        as: 'Products',
        where: {
          fingerprint,
        },
        include: {
          model: db.models.Redacted,
          as: 'StubExample',
        },
      },
    ],
     defaults: {
      ...crawledEvent,
      idVenue,
      Products: {
        ...crawledProduct,
        idPlatform: platform.id,
        fingerprint,
        StubExample: {
          readacted: true,
          priority: 60,
          type: 'supply',
        },
      },
    },
   },
  });
```

PR #8494, opened 2017-10-17
feat(sqlite/query-generator): sqlite delete supporting limit via rowid

netsgnut:
### Pull Request check-list

### Description of change

By default SQLite is compiled with `SQLITE_ENABLE_UPDATE_DELETE_LIMIT` off. That means the `DELETE` statements would not support an optional `LIMIT` clause. By using `DELETE ... WHERE rowid IN (SELECT rowid FROM ... WHERE ... LIMIT n)` instead it can provide support in environments without such feature.

(This is a companion PR to #8450.)

### DIFF
--- a/lib/dialects/sqlite/query-generator.js
+++ b/lib/dialects/sqlite/query-generator.js
@@ -288,12 +288,25 @@ const QueryGenerator = {
     options = options || {};
     _.defaults(options, this.options);
 
-    let whereClause = this.getWhereConditions(where, null, model, options);
-    if (whereClause) {
-      whereClause = ' WHERE ' + whereClause;
+    if (options.truncate === true) {
+      // Truncate does not allow LIMIT and WHERE
+      return `DELETE FROM ${this.quoteTable(tableName)}`;
     }
 
-    return `DELETE FROM ${this.quoteTable(tableName)}${whereClause}`;
+    if (_.isUndefined(options.limit)) {
+      options.limit = 1;
+    }
+
+    let whereClause = this.getWhereConditions(where, null, model, options);
+    if (whereClause) {
+      whereClause = `WHERE ${whereClause}`;
+    }
+
+    if (options.limit) {
+      whereClause = `WHERE rowid IN (SELECT rowid FROM ${this.quoteTable(tableName)} ${whereClause} LIMIT ${this.escape(options.limit)})`;
+    }
+
+    return `DELETE FROM ${this.quoteTable(tableName)} ${whereClause}`;
   },
 
   attributesToSQL(attributes) {
--- a/test/unit/sql/delete.test.js
+++ b/test/unit/sql/delete.test.js
@@ -86,6 +86,7 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
           ), {
             default:  "DELETE FROM [public.test_users] WHERE `name` = 'foo'",
             postgres: 'DELETE FROM "public"."test_users" WHERE "name" = \'foo\'',
+            sqlite:   "DELETE FROM `public.test_users` WHERE `name` = 'foo'",
             mssql:    "DELETE FROM [public].[test_users] WHERE [name] = N'foo'; SELECT @@ROWCOUNT AS AFFECTEDROWS;"
           }
         );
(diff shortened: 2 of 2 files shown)


eseliger:
@netsgnut Some tests fail, but overall looks good to me

netsgnut:
@eseliger, thanks for the head-up. I think the tests should be fine now with f254ddb.

Please see if there are other cases I have missed as well. Thanks.

eseliger:
How about a integration test that verifies the functionality?

netsgnut:
> How about a integration test that verifies the functionality?
Sounds like a good idea, although frankly I am not sure how to implement a proper integration test apart from the modified test cases...

lib/dialects/sqlite/query-generator.js
```diff
@@ -288,12 +288,25 @@ const QueryGenerator = {
...
 
-    return `DELETE FROM ${this.quoteTable(tableName)}${whereClause}`;
+    if (_.isUndefined(options.limit)) {
+      options.limit = 1;
+    }
+
+    if (options.limit) {
+      return `DELETE FROM ${this.quoteTable(tableName)} WHERE rowid IN (SELECT rowid FROM ${this.quoteTable(tableName)}${whereClause} LIMIT ${this.escape(options.limit)})`;
+    } else {
+      return `DELETE FROM ${this.quoteTable(tableName)}${whereClause}`;
```
sushantdhiman: This case will never run, if limit is undefined you are setting is to 1, which will always trigger if case.

I suggest just remove `options.limit/isUndefined` part, switch query only when limit is supplied by user

kelvinwonghk01: OK. Although I am still curious on why we would need the `options.limit/isUndefined` part from [`dialect/mysql.js`](https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/query-generator.js#L182-L190), I guess we really should drop the conditional here. Updating it.

netsgnut: That said, is there a case where `limit` is `0`, though? If so that would be a case that triggers the `else` clause?

sushantdhiman: In that case it would be better to check for `isDefined` in if expression, because we want to limit DELETE even if its set to 0.

netsgnut: > In that case it would be better to check for `isDefined` in if expression, because we want to limit DELETE even if its set to 0.

Do you want that to reflect on other dialects as well? Because I am afraid this is not how the current implementation for all other dialects is. That includes [MySQL](https://github.com/sequelize/sequelize/blob/17b19d9c67a3b6619f2cce4e920bc52b2b0fc4ab/lib/dialects/mssql/query-generator.js#L464), [PostgreSQL](https://github.com/sequelize/sequelize/blob/17b19d9c67a3b6619f2cce4e920bc52b2b0fc4ab/lib/dialects/postgres/query-generator.js#L410) and [MSSQL](https://github.com/sequelize/sequelize/blob/17b19d9c67a3b6619f2cce4e920bc52b2b0fc4ab/lib/dialects/mysql/query-generator.js#L188).

netsgnut:
I have updated the PR for a little refactoring of the code (3001741), but on second thought in order to maintain the consistency with other implementations here I did not change much of the logic.

Regarding @sushantdhiman's comments:

> Dont set limit if its undefined.

This is essential for the SQLite implementation to behave like other dialects. Without setting the limit the `delete when the primary key has a different field name` test would fail.

> You need to add a unit test for both cases
> 1.  limit is set, rowId based query
> 2. limit is not set, old delete query

The cases are already covered with the current test `delete without limit` (Case 1) and `delete with limit` (Case 2), but I added the SQLite-specific line for Case 1 as well just to be explicit - it works without it as the `default` line already captures it.

---

That said, I think there should be a standard behavior defined for all the dialects facing the common cases where the `limit` is not given, `0`, as well as `null`. Those, IMHO, should go in another PR as it potentially changes/breaks current usage. If that behavior is desired, I am happy to make another PR for that including the necessary documentation change - given there is a consensus on how their behavior should be.

netsgnut:
(Not sure why the [AppVeyor build](https://ci.appveyor.com/project/sushantdhiman/sequelize/build/2496) failed though for 3001741...)

sushantdhiman:
@netsgnut Its clear that if you don't supply limit, whatever the where condition may be code will end up deleting only first row. No matter how I process it, it feels wrong.

Investigated and found [this line](https://github.com/sequelize/sequelize/blob/a07fb75982216d04cd889bddae4193265de771df/lib/query-interface.js#L1005) has been saving us till now.

I agree changing other dialect is not a part of this PR, but this is a potential mishap in waiting.

This code section is old so I dont know the reason behind this, probably implementer(s) just copied code from first dialect to others.

Since we default to limit `null` and we have a safe guard test (unit) against this issue I am ready to move forward with logic adopted from other dialects

PR #8500, opened 2017-10-17
feat(instance-validator): add model name to allowNull messages

### LINKED ISSUE #8498: allowNull validation errors don't include offending model's name
## What are you doing?
I'm currently debugging a `SequelizeValidationError` that looks like this in my logs:
> SequelizeValidationError: notNull Violation: deletedAt cannot be null

Since I have a lot of models with `deletedAt` properties, this message isn't as helpful as it could be.

Error message should include the name of the model in question to help track down the issue.

See above.

I'm scoping this issue to just model name on the allowNull validator, but I think there are similar issues with other validators.

__Dialect:__ any (tested on PG)
__Dialect version:__ any
__Database version:__ any
__Sequelize version:__ master

gabegorelick:
Closes #8498

### Pull Request check-list

### Description of change

Include `Model.name` in `allowNull` validation error messages. There are probably more things that we could include in this message, and more places we should include `Model.name`, but I figured I'd start with this.

### DIFF
--- a/lib/instance-validator.js
+++ b/lib/instance-validator.js
@@ -322,7 +322,7 @@ class InstanceValidator {
 
     if (rawAttribute.allowNull === false && (value === null || value === undefined)) {
       const validators = this.modelInstance.validators[field];
-      const errMsg = _.get(validators, 'notNull.msg', `${field} cannot be null`);
+      const errMsg = _.get(validators, 'notNull.msg', `${this.modelInstance.constructor.name}.${field} cannot be null`);
       error = new sequelizeError.ValidationErrorItem(errMsg, 'notNull Violation', field, value);
       this.errors.push(error);
     }
--- a/test/integration/instance.validations.test.js
+++ b/test/integration/instance.validations.test.js
@@ -299,7 +299,7 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), () => {
         });
       });
     });
-    
+
     describe('pass all paths when validating', () => {
       beforeEach(function() {
         const self = this;
@@ -345,7 +345,7 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), () => {
         });
       });
     });
-    
+
     describe('not null schema validation', () => {
       beforeEach(function() {
         const Project = this.sequelize.define('Project', {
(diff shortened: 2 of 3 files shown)


iamjochem:
or maybe this: #7557 .... because "fixing" errors messages so they contain something that is needed in a particular context is a lot less useful than structurally decorating ValidatorErrorItems with the instance & validator that generated the error.

gabegorelick:
@iamjochem I think both approaches are useful. More structured info is definitely good but requires custom serializers so it ends up in your logs. Better error messages are good for better out of the box debuggability.

iamjochem:
hi @gabegorelick, yeah sorry, my comment was not aimed at you or your PR as such, I was merely finding an excuse to promote my somewhat related own PR (PR for a PR? :-)). 

you are right that richer error messages can only be helpful but I would argue that they only help a developer during development ... unstructured log messages quickly become useless when lost in a sea of other messages in a log file (wihout much if any context), you are right about custom serializers (in my case I use bunyan and I have them). 

Another group of error message consumers is end users and generally error messages are useless to them, however rich the content, because the message almost always exposes underlying technical implementation which have little to do with the interface, not to mention that the user may be using a different language.

In order to expose [validation] error information to a UI (which is likely to be running in a browser where as sequelize is running on a server) you have to be able to map a given error to a specific UI element, specify what failed (i.e. which validation rule, and with which parameters) and present that information in the users own language ... so you need to be able to transport structured, language-agnostic error information from source (sequelize) to somewhere far far away.

sushantdhiman:
This could be a potential breaking case? Many users expose these messages to users.

@sequelize/orm

eseliger:
@sushantdhiman I don't think this is a breaking change 🤔 If anyone relies on error messages, I would consider this bad practice, as it's not meant to be machine readable. If people need to find out more about the validation errors, we could maybe add more properties to the error that are machine readable and help investigating, like `model` and `field` properties. 
And example why this can be bad practice: Internet explorer tends to translate the messages of its errors to the system language, but not the error name 

Let me know what you think

iamjochem:
@eseliger - you could indeed add more properties (or even set the ones that already exist - something that does not always happen) ... there is a PR for that (cover only validation errors): #7557

eseliger:
thats fine, I didn't wanna say it's needed in this PR :) 
I was just suggesting that if there is demand from people to really be able to get this information from the error, they could do a pr (like #7557) to add those fields, and I'd not mark changing error messages as breaking. Other than error name, this IS unique and distinct. message may not.

iamjochem:
@eseliger - I totally agree with not make error messages part of an API contract - but what I really want is structured validation error information so that I can store, expose, translate, map to a UI, etc ... which is why I spent quite some time making #7557 ... but nobody really seems to dare touch it (a y more) ... which sucks :-/.

gabegorelick:
Seems like there's consensus then? Can someone merge this if so?

PR #8541, opened 2017-10-25
fix(increment): support scopes

### LINKED ISSUE #8537: Model.increment throws `Missing where attribute in the options parameter` error even if used with scope
## What are you doing?

Model.increment requires a `where` option be present, even if it's empty. But it ignores `where` options populated via scopes.

```js
it('should work with scopes', function () {
    aNumber: DataTypes.INTEGER,
    name: DataTypes.STRING
  }, {
    scopes: {
      jeff: {
        where: {
          name: 'Jeff'
        }
      }
    }
  });

    return User.bulkCreate([
      {
        aNumber: 1,
        name: 'Jeff'
      },
      {
        aNumber: 3,
        name: 'Not Jeff'
      }
    ]);
  }).then(() => {
    return User.scope('jeff').increment('aNumber');
  }).then(() => {
    return User.scope('jeff').findOne();
  }).then(jeff => {
    expect(jeff.aNumber).to.equal(2);
  }).then(() => {
    return User.findOne({
      where: {
        name: 'Not Jeff'
      }
    });
  }).then(notJeff => {
    expect(notJeff.aNumber).to.equal(3);
  });
});
```

`where` from scope should be applied to `increment` operation.

`.increment` throws an Error: `Missing where attribute in the options parameter`.

__Dialect:__ any
__Dialect version:__ any
__Database version:__ any
__Sequelize version:__ master

### LINKED ISSUE #8539: Model.increment ignores scopes
## What are you doing?

Given that [you have to pass `where: {}`](https://github.com/sequelize/sequelize/issues/8537), I'd still expect scopes to work with `Model.increment`:
```js
it('should work with scopes', function () {
    aNumber: DataTypes.INTEGER,
    name: DataTypes.STRING
  }, {
    scopes: {
      jeff: {
        where: {
          name: 'Jeff'
        }
      }
    }
  });

    return User.bulkCreate([
      {
        aNumber: 1,
        name: 'Jeff'
      },
      {
        aNumber: 3,
        name: 'Not Jeff'
      }
    ]);
  }).then(() => {
    return User.scope('jeff').increment('aNumber', {
      where: {}
    });
  }).then(() => {
    return User.scope('jeff').findOne();
  }).then(jeff => {
    expect(jeff.aNumber).to.equal(2);
  }).then(() => {
    return User.findOne({
      where: {
        name: 'Not Jeff'
      }
    });
  }).then(notJeff => {
   // this fails since notJeff was incremented too
    expect(notJeff.aNumber).to.equal(3);
  });
});
```

Only `Jeff` is incremented.

All users are incremented since scope is ignored.

__Dialect:__ any
__Dialect version:__ any
__Database version:__ any
__Sequelize version:__ master

gabegorelick:
Closes #8539
Closes #8537 

### Pull Request check-list

### Description of change

Make `Model.increment` respect scopes.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2892,6 +2892,8 @@ class Model {
   * @return {Promise<this>}
   */
   static increment(fields, options) {
+    options = options || {};
+    this._injectScope(options);
     this._optionsMustContainWhere(options);
 
     const updatedAtAttr = this._timestampAttributes.updatedAt;
--- a/docs/scopes.md
+++ b/docs/scopes.md
@@ -86,7 +86,7 @@ DeletedProjects.findAll();
 DeletedProjects.findAll();
 ```
 
-Scopes apply to `.find`, `.findAll`, `.count`, `.update` and `.destroy`.
+Scopes apply to `.find`, `.findAll`, `.count`, `.update`, `.increment` and `.destroy`.
 
 Scopes which are functions can be invoked in two ways. If the scope does not take any arguments it can be invoked as normally. If the scope takes arguments, pass an object:
 
--- a/test/integration/model.test.js
+++ b/test/integration/model.test.js
@@ -2931,5 +2931,48 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         return expect(User.findById(1)).to.eventually.have.property('updatedAt').equalTime(oldDate);
       });
     });
+
+    it('should work with scopes', function() {
+      const User = this.sequelize.define('User', {
+        aNumber: DataTypes.INTEGER,
+        name: DataTypes.STRING
+      }, {
+        scopes: {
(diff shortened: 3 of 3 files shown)


lib/model.js
```diff
@@ -2892,6 +2892,7 @@ class Model {
   * @return {Promise<this>}
   */
   static increment(fields, options) {
+    this._injectScope(options);
```
gabegorelick: Having this before `_optionsMustContainWhere` makes sense to me, but if there's a reason to inject scopes after that assertion, we can move it down.

sushantdhiman:
Tests failing

lib/model.js
janmeier: I think it fits fine here

PR #8569, opened 2017-10-31
feat(cls-hooked): Add a test for cls-hooked

jeremija:
### Pull Request check-list

- [ ] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)? **please see below**

### Description of change

Hi @janmeier. Following our discussion at #8430 ([comment link][clink]), I added a test for [cls-hooked][cls-hooked]. Looks like it already works seamlessly since the API is the same as the original [CLS][cls] library.

The test is `test/integration/cls-hooked/cls-hooked.test.js`. I added it to a subfolder because I had to change ESLint's `parserOptions.ecmaVersion` to `2017` to be able to use `async / await` keywords.

### Warnings from CLS Hooked

> When running Nodejs version < 8, this module uses [AsyncWrap][async-wrap] which is an unsupported Nodejs API, so please consider the risk before using it.
>
> When running Nodejs version >= 8.2.1, this module uses the newer [async_hooks][async-hooks] API which is considered Experimental by Nodejs.

### Check for Node 8

I am not sure if this test will run in the CI script since I added a check for Node 8:

```javascript
const isNode8 = semver.satisfies(process.version, '>= 8');
if (current.dialect.supports.transactions && isNode8) {
```

My node version is `v8.7.0` and I ran the MySQL tests. My version of MySQL: `mysql  Ver 14.14 Distrib 5.7.19, for osx10.13 (x86_64) using  EditLine wrapper`.

### Failing Tests

I had 4 failing MySQL tests that seem unrelated to any changes I made, and seem related to my machine:

```bash
  1) [MYSQL] DataTypes should return YYYY-MM-DD format string for DATEONLY:
     AssertionError: expected Mon Oct 30 2017 to equal Tue Oct 31 2017

     ^ This could be because of timezone difference

  2) [MYSQL] QueryInterface removeColumn (without a schema) should be able to remove a column with primaryKey:
     SequelizeDatabaseError: Can't DROP 'books_ibfk_1'; check that column/key exists

  3) [MYSQL] QueryInterface constraints primary key should add, read & remove primary key constraint:
     SequelizeDatabaseError: Can't DROP 'books_ibfk_1'; check that column/key exists

  4) [MYSQL] QueryInterface constraints foreign key should add, read & remove foreign key constraint:
     SequelizeDatabaseError: Can't DROP 'books_ibfk_1'; check that column/key exists
```

Finally, if there is an interest in this, we should probably mention `cls-hooked` as an option in the [documentation][docs].

[async-wrap]: https://github.com/nodejs/node-eps/blob/async-wrap-ep/XXX-asyncwrap-api.md
[async-hooks]: https://github.com/nodejs/node/blob/master/doc/api/async_hooks.md
[clink]: https://github.com/sequelize/sequelize/pull/8430#issuecomment-340682663
[cls-hooked]: https://www.npmjs.com/package/cls-hooked
[cls]: https://github.com/othiym23/node-continuation-local-storage
[docs]: http://docs.sequelizejs.com/manual/tutorial/transactions.html#automatically-pass-transactions-to-all-queries

### DIFF
--- a/test/integration/cls-hooked/cls-hooked.test.js
+++ b/test/integration/cls-hooked/cls-hooked.test.js
@@ -1,0 +1,148 @@
+'use strict';
+
+const chai      = require('chai'),
+  expect    = chai.expect,
+  Support   = require('../support'),
+  Sequelize = Support.Sequelize,
+  Promise   = Sequelize.Promise,
+  cls       = require('cls-hooked'),
+  current = Support.sequelize,
+  semver = require('semver'),
+  isNode8 = semver.satisfies(process.version, '>= 8');
+
+if (current.dialect.supports.transactions && isNode8) {
+  describe(Support.getTestDialectTeaser('cls-hooked'), () => {
+    before(function() {
+      this.thenOriginal = Promise.prototype.then;
+      Sequelize.useCLS(cls.createNamespace('sequelize'));
+    });
+
+    after(() => {
+      delete Sequelize._cls;
+    });
+
+    let sequelize, ns, User;
+    beforeEach(async function() {
+      sequelize = await Support.prepareTransactionTest(this.sequelize);
+      ns = cls.getNamespace('sequelize');
+      User = sequelize.define('user', {
+        name: Sequelize.STRING
+      });
+      await sequelize.sync({ force: true });
+    });
+
+    describe('context', () => {
+      it('does not use continuation storage on manually managed transactions', () => {
+        return Sequelize._clsRun(async() => {
+          const transaction = await sequelize.transaction();
+          expect(ns.get('transaction')).to.be.undefined;
+          await transaction.rollback();
+        });
+      });
+
+      it('supports several concurrent transactions', async() => {
+        const [t1id, t2id] = await Promise.all([
+          sequelize.transaction(async() => {
+            const _t1id = ns.get('transaction').id;
+            return await Promise.resolve(_t1id);
+          }),
+          sequelize.transaction(async() => {
+            const _t2id = ns.get('transaction').id;
+            return await Promise.resolve(_t2id);
+          })
+        ]);
+        expect(t1id).to.be.ok;
+        expect(t2id).to.be.ok;
+        expect(t1id).not.to.equal(t2id);
+      });
+
+      it('supports nested promise chains', async() => {
+        await sequelize.transaction(async() => {
+          const tid = ns.get('transaction').id;
+          expect(tid).to.be.ok;
+
+          await User.findAll().then(() => {
+            expect(ns.get('transaction').id).to.equal(tid);
+          });
+          expect(ns.get('transaction').id).to.equal(tid);
+        });
+      });
+
+      it('does not leak variables to the outer scope', async() => {
+        // This is a little tricky. We want to check the values in the outer scope, when the transaction has been successfully set up, but before it has been comitted.
+        // We can't just call another function from inside that transaction, since that would transfer the context to that function - exactly what we are trying to prevent;
+
+        let transactionSetup = false,
+          transactionEnded = false;
+
+        sequelize.transaction(async() => {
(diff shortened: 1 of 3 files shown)


test/integration/cls-hooked/cls-hooked.test.js
```diff
@@ -0,0 +1,148 @@
...
+      this.thenOriginal = Promise.prototype.then;
+      Sequelize.useCLS(cls.createNamespace('sequelize'));
+    });
+
+    after(() => {
+      delete Sequelize._cls;
+    });
+
+    let sequelize, ns, User;
+    beforeEach(async function() {
```
janmeier: Please don't use es7 :). Even though you have a guard at the start of the file, older versions will still try to parse the file.

This is copy-pasta of the current cls tests right? In that case, just replace the import of `continuation-local-storage` with `cls-hooked`. As far a I can see it should work on older versions as well

jeremija: Yes it's copy-paste. I can remove that, but the main reason why I added this was to see if the `async/await` features worked correctly. I realize that you have to support multiple versions of node, but how else could I test `async/await`? Do you think using `global.Promise` in the test cases instead of patched Bluebird would make a good-enough test case?

janmeier: If all tests pass on old version of node with cls hooked, I'm happy.

Async / await support is built into async_hooks, so we don't need to specifically test for it

janmeier:
Paging @overlookmotel any thoughts on this?

overlookmotel:
Interesting! Yes `cls-hooked` is definitely the future, but I don't think it's there yet.

I suspect at this point that it will not work in sequelize now, because Bluebird does not yet support `async-hooks`, which is the underlying mechanism which makes `cls-hooked` work. Bluebird is obviously used heavily throughout Sequelize and the only reason it works at present is using the `cls-bluebird` shim. I wrote `cls-bluebird` and I'm pretty sure that it won't work if you swap in `async_hooks` for `async_listener`.

This is really tricky to write tests for. CLS will tend to work fine when there's only one thing happening at a time, as there usually is in tests. But in the real world when you have multiple "threads" of activity going on at the same time, that's when a dodgy CLS setup will start mixing up the contexts between the threads.

So what I'm saying is it's easy to write tests that pass, but then find in a real application it goes disastrously wrong.

Further:
1. There are probably other elements in sequelize which don't support `async-hooks` (like the connection pool most likely)
2. The rules `async-hooks` follows propagating context with promises is different from the rules `cls-bluebird` follows https://github.com/TimBeyer/cls-bluebird/issues/6

There's more things I could add to this list.

I'm afraid I don't have time at the moment to go through this PR. But what I'm basically saying is I would *strongly* advise not attempting this change until a lot of other groundwork has taken place.

I'm not current on the subject, so I could be completely wrong. But in my mind transactions need to be 100% rock solid as they are the foundation of data integrity, so why risk it?

janmeier:
@overlookmotel Thanks for the feedback - I had a feeling as well that everything might not work as perfectly as it looks at first sight with cls hooked.
I agree that when we provide magic to users, we should ensure that the magic actually works, and that it works all the time!

@jeremija If you have the time you are very welcome to test out cls-hooked in your own setup. However, do be aware of the caveats mentioned above. First the time being, I don't think we want to recommend it to the general sequelize userbase.

jeremija:
@janmeier @overlookmotel Thanks for the feedback - this makes perfect sense since the `async_hooks` API is still considered experimental. The good news is that `cls-hooked` can be plugged in seamlessly to Sequelize so `async/await` seems to work when using transactions, but I cannot guarantee that all cases are covered with different versions of Node. I'll try to use it in a non-critical project at some point to see if there are any caveats.

janmeier:
Excited to hear how it works @jeremija  :)

PR #8582, opened 2017-11-02
fix(query_generator.js): fixes duplicate unique index

iaptsiauri:
#6134 

### Pull Request check-list

### Description of change
Removes the unique index from the dataType script, thus not causing the duplicate indexes to be
created

### DIFF
--- a/test/unit/dialects/mysql/query-generator.test.js
+++ b/test/unit/dialects/mysql/query-generator.test.js
@@ -117,6 +117,10 @@ if (dialect === 'mysql') {
           expectation: 'CREATE TABLE IF NOT EXISTS `myTable` (`title` VARCHAR(255), `name` VARCHAR(255)) ENGINE=MyISAM;'
         },
         {
+          arguments: ['myTable', {title: 'VARCHAR(255) UNIQUE', name: 'VARCHAR(255)'}, {engine: 'MyISAM'}],
+          expectation: 'CREATE TABLE IF NOT EXISTS `myTable` (`title` VARCHAR(255), `name` VARCHAR(255)) ENGINE=MyISAM;'
+        },
+        {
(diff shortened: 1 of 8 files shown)


test/unit/dialects/mysql/query-generator.test.js
```diff
@@ -117,6 +117,10 @@ if (dialect === 'mysql') {
           expectation: 'CREATE TABLE IF NOT EXISTS `myTable` (`title` VARCHAR(255), `name` VARCHAR(255)) ENGINE=MyISAM;'
         },
         {
+          arguments: ['myTable', {title: 'VARCHAR(255) UNIQUE', name: 'VARCHAR(255)'}, {engine: 'MyISAM'}],
+          expectation: 'CREATE TABLE IF NOT EXISTS `myTable` (`title` VARCHAR(255), `name` VARCHAR(255)) ENGINE=MyISAM;'
```
sushantdhiman: Shouldn't UNIQUE be added to this query? Do we create an index now, if thats the case you need to write an integration test that will create a table and assert only one index is created per unique

iaptsiauri: Sure will add integration tests, in this case, if I understood correctly the UNIQUE here causes a duplication as it is added in 
> sequelize/lib/model.js
> Line 806 in 47b3f61
> ```if (definition.hasOwnProperty('unique') && definition.unique !== false) {  ```

sushantdhiman: I am afraid now if anyone try to use `createTable` with `unique: true` attribute it wont create any constraint. Sync handle this but `createTable` doesnt.

Can you add a test for this, create a table with `createTable` having a unique attribute and verify if unique index is still created

iaptsiauri: @sushantdhiman you are right in case of ```createTable``` it doesn't and the ```unique``` attribute, do you have any suggestions maybe? The naive one would be to check the caller, if I will come up with better one will let you know :)

Issue #8591, opened 2017-11-04
hasMany has() wrong parameters


shtse8:
Consider this Association:
```js
User.hasMany(Project, { as: 'projects' })
```
and this code:
```js
user.hasProject({ where: { id: projectId } })
```

I expected to return a boolean to represent if the user has the project with `projectId`.

But an error is thrown:
```
Invalid value { where: { id: 14 } }
```

Due is due to the wrong parameters designed in https://github.com/sequelize/sequelize/blob/master/lib/associations/has-many.js#L268
```js
has(sourceInstance, targetInstances, options) {
```

my options are passed in the second parameter. It's much different with the [get](https://github.com/sequelize/sequelize/blob/master/lib/associations/has-many.js#L165).
```js
get(instances, options) {
```

and also `count`
```js
count(instance, options) {
```

for the latest release, hasMany mixin is buggy. only `get`, `create` and `count` work properly. The others like (`has`, `add`, `remove`) work failed.

shtse8:
I have fixed this bug with https://github.com/sequelize/sequelize/pull/8592

sushantdhiman:
Where do to find docs that say 

```js
user.hasProject({ where: { id: projectId } })
```

is valid, I think you are meant to pass instance to this method not options

shtse8:
It is very complicated by passing an instance into the `has`, as we have to query the instance first. It's conflict that we don't know if the instance exists, so we need to use `has` to check, but at the end we need to `get` the instance first to check the instance if it exists. it's meaningless to use `has` which is assumed to be better performance by using `count` in `sql` to check.

sushantdhiman:
`hasAssociation` is meant to for verifying if an instance is related to another like this

`authorA.hasBook(bookA)`

and is very useful, in current form.

If you want to check for relation using a condition `get` is very good option for that

`author.get({ where: { name: 'Alpha' } }).then(books => console.log(books.length))`

What problem you have with `get` ?

shtse8:
`get` will actually fetch the data from database instead of using count in SQL resulting in poor performance.
if `hasBook` is meant to verify `authorA` if it's related, it still can be done by using `get`.
`authorA.get({ where: { id: bookA.id } }).then(books => console.log(books.length))`

PR #8592, opened 2017-11-04
hasMany mixin fix

shtse8:
This PR is to fix https://github.com/sequelize/sequelize/issues/8591


sushantdhiman:
Test fails and it contain extra commit from different PR

PR #8598, opened 2017-11-05
fix<handleSelectQuery>: obey the options.raw

shtse8:
When calling Model.find, the option raw is ignored in the latest release. When someone extends the data types or override _sanitize of datatype. options.raw always passing false so that raw data is returned and never create a proper object.

This is because AbstractQuery class override the raw option when calling BulkCreate. I think it should follow the option.

This PR is to fix this bug, please review.


sushantdhiman:
Ok I generally agree that we should allow overriding this options. Can you do a few things

1. Tests are failing, you need to fix that first
2. Write an integration test that demonstrate issue you are facing, to prevent future regressions

PR #8603, opened 2017-11-06
fix(sync): drop foreign key constraint before re-adding it to prevent…

d-fischer:
… duplicate errors

Currently, when I create two tables that have a relation, and I use `sync({alter: true})` on an existing database with the foreign key constraints already in place, the sync fails with a "duplicate key" error.

### Description of change

Before changing a column, I check whether the new column definition includes a `REFERENCES` clause. If so, just to be sure, I try removing the foreign key constraint before changing the column, ignoring failures due to the constraint actually not existing (because it might be the first time we introduce this foreign key). Now `changeColumn` can create a new foreign key constraint.

The new test fails on master and passes on my fork.

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -10,6 +10,7 @@ const Transaction = require('./transaction');
 const Promise = require('./promise');
 const QueryTypes = require('./query-types');
 const Op = require('./operators');
+const UnknownConstraintError = require('./errors').UnknownConstraintError;
 
 /**
  * The interface that Sequelize uses to talk to all databases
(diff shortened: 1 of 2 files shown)


d-fischer:
Regarding build fails: 
- MSSQL: seems random to me, not related to my changes.

- PG: why do we have different naming conventions for PG and MySQL for foreign key indexes?

- SQLite: I don't know why the table wouldn't exist, please help me out here. I checked the test with a console.log on all `query(sql)` calls. The `testSyncs` table is backed up, dropped and recreated a *lot* of times, but the last status looks like it *should* exist. Also, I checked out the error that was thrown, and it makes no sense to me:

      { Error: SQLITE_ERROR: no such table: temp.testSyncs
      errno: 1,
      code: 'SQLITE_ERROR',
      sql: 'INSERT INTO `testSyncChildren_backup` SELECT `id`, `name`, `age`, `createdAt`, `updatedAt`, `testSyncId` FROM `testSyncChildren`;' }

The table from the error is not even in the query. But I don't know the quirks of SQLite at all, so I'll need some help.

sushantdhiman:
Should be fixed by https://github.com/sequelize/sequelize/pull/8795

PR #8605, opened 2017-11-07
fix sqlstring bug: if save json string ,may cause bug

chalecao:
### Pull Request check-list

- [ ok ] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [ no ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ no ] Have you added new tests to prevent regressions?
- [ no ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ no ] Did you follow the commit message conventions explained in CONTRIBUTING.md?

### Description of change
if save json string ,sqlstring may cause bug in "typeof val" switch case which pass "val" a object, so i add a "if" change before the "typeof val switch case" .

### DIFF
--- a/lib/sql-string.js
+++ b/lib/sql-string.js
@@ -9,6 +9,9 @@ function escape(val, timeZone, dialect, format) {
   if (val === undefined || val === null) {
     return 'NULL';
   }
+  if(typeof val === "object"){
+    val = JSON.stringify(val);
+  }
   switch (typeof val) {
     case 'boolean':
     // SQLite doesn't have true/false support. MySQL aliases true/false to 1/0


sushantdhiman:
Massive tests failure and what is the use case of this change?

iamjoyce:
Just a feedback on why your test case fails.

1. The codes probably will fail the lint test as well. Leave a space between `if` and `(condition)`
1. You can't check for JSON strings this way. There are plenty of things that are also `object` type apart from JSON.

To properly check for JSON, you will need something like this:

```js
if (isJson(val)) {
    val = JSON.stringify(val);
}

function isJson(str) {
    try {
        JSON.parse(str);
    } catch (e) {
        return false;
    }
    return true;
}
```

Having said all these, I think it will be better to put `JSON.stringify` in your main codes instead of Sequelize library because I think there are other components that require the JSON object to remain as an object.

Issue #8417, opened 2017-10-03
Deprecation warning for String based operators


krishangupta:
Since I upgraded to 4.13.2 I get a deprecation warning

```
sequelize deprecated String based operators are now deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/tutorial/querying.html#operators node_modules/sequelize/lib/sequelize.js:236:13
```

I do not use any string based operators, however, or any operators of any kind. I didn't even know what an operator was actually, which is why I was confused.

I can fix this by adding ```operatorsAliases: false``` when instantiating the Sequellize object as below. Should this not be the default though? Or can we have the warning and documentation more clearly state that ```operatorsAliases: false``` must be set in order to avoid the warning?

```

const sequelize = new Sequelize(
  config.RDS_DB_NAME,
  config.RDS_USERNAME, 
  config.RDS_PASSWORD,
  {
    host: config.RDS_HOSTNAME,
    dialect: 'mysql',
    logging: false,
    freezeTableName: true,
    operatorsAliases: false
  }
)
```

veresnikov:
The same error

sushantdhiman:
This is a deprecation warning, reason behind which is clearly stated in http://docs.sequelizejs.com/manual/tutorial/querying.html#operators-security , I am sure you must have read it.

Let me distill this regardless, Most web frameworks in Node.js allow parsing a object like string to actual JS object. This become an major issue when developers are passing user input without sanitizing it to Sequelize methods.

For example, consider this sample of code

```js
db.Token.findOne({
      where: {
        token: req.query.token
      }
);
```

Now a bad actor could pass `token='{"$gt": 1}'` which will make above query to become something like this

```js
db.Token.findOne({
      where: {
        token: {
           $gt: 1
        }
      }
);
```
 
This is because `$gt` is a string based operator which can be injected as string. To mitigate this we introduced secure operators https://github.com/sequelize/sequelize/pull/8240

Secure operators are Symbols which can't be duplicated by such object conversion. If we were using above code with secure operators we get this state

```js
db.Token.findOne({
      where: {
        token: {
           $gt: 1 // invalid, as Op.gt is a operators but $gt is not. This will throw an error
        }
      }
);
```

We cant make these operators default as this would be a breaking change.

**So what should you as a developer do now**

[...]

PR #8608, opened 2017-11-08
docs(querying): Added examples for symbolic operator queries with [Op.or]

Laurensdc:
### Description of change

Adding a few basic examples for the usage of Symbolic Operators, since they were not clear to me, and the examples under 'Combinations' were quite confusing.

[Issue link](https://github.com/sequelize/sequelize/issues/8417)

### DIFF
--- a/docs/querying.md
+++ b/docs/querying.md
@@ -93,6 +93,22 @@ Post.findAll({
 });
 // SELECT * FROM post WHERE authorId = 12 AND status = 'active';
 
+Post.findAll({
+    where: {
+        [Op.or]: [{authorId: 12}, {authorId: 13}]
+    }
+});
+// SELECT * FROM post WHERE authorId = 12 OR authorId = 13;
+
+Post.findAll({
+    where: {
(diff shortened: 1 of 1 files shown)


yonjah:
@Laurensdc  Examples looks good. 

I'm just wondering if we might need a specific section for Op.or and Op.and since they are more complicated than the other operators and maybe leave the basics as is to not confuse users who only care about the basic usage

yonjah:
Also Sequelize  follows Commitizen for commit messages.
If you commit from the repository you should have `git-cz` cli  to help you set the commit message
But for this topic the commit message should be prefixed with
docs(querying):

docs/querying.md
```diff
@@ -93,6 +93,22 @@ Post.findAll({
 });
 // SELECT * FROM post WHERE authorId = 12 AND status = 'active';
 
+Post.findAll({
+    where: {
```
sushantdhiman: Other examples are using 2 space indentation, Please use that in these new examples as well

PR #8619, opened 2017-11-11
fix(model.js): fixes duplicate unique index

iaptsiauri:
### Pull Request check-list

- [] Have you added new tests to prevent regressions?

### Description of change
Removes unique definition from the rawAttributes of the model after it is added to the options thus
prevents the duplication

#6134

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1010,8 +1010,8 @@ class Model {
         idx.msg = idx.msg || definition.unique.msg || null;
         idx.name = idxName || false;
         idx.column = name;
-
         this.options.uniqueKeys[idxName] = idx;
+        delete definition.unique;
       }
 
       if (definition.hasOwnProperty('validate')) {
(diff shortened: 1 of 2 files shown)


iaptsiauri:
@sushantdhiman can you please take a look at the PR the CI checks are failing because of `dao-factory.test.js` test  
> it('handles extended attributes (unique)', function() {

but exactly because of this behavior the duplicate indexes are create, can I remove this test and add  two additional ones, one for migration and another for sync usage?

sushantdhiman:
Thanks for this PR. This is solution looks bit odd to me. You really want to keep that unique definition in `rawAttributes`, its only source of truth about attributes.

If I am not mistaken this condition occurs because `createTable` and `sync` both try to add indexes separately. I suggest just checking for presence of index on that attribute with `unique: true` using `describeTable`. So we dont end up creating two indexes with different names.

lib/model.js
```diff
@@ -1010,8 +1010,8 @@ class Model {
         idx.msg = idx.msg || definition.unique.msg || null;
         idx.name = idxName || false;
         idx.column = name;
-
         this.options.uniqueKeys[idxName] = idx;
+        delete definition.unique;
```
eseliger: using delete is imperformant, is setting it to false/undefined not enough?

sushantdhiman:
Fixed by https://github.com/sequelize/sequelize/pull/9033

Issue #8620, opened 2017-11-11
Model attributes are not protected against null or undefined values


arcdev1:
## What are you doing?

Building an adapter for Sequelize to isolate a codebase from a direct dependency on it. In doing so, I need to map a custom column definition object to a Sequelize model attribute. See code below to get the gist.

```js
const table = await createTable({
  name: 'Foo',
  columns: {
    fooId: {
      type: 'id'
    },
    name: {
      type: 'string'
    }
  }
})

async function createTable ({ name, columns, overwrite = false } = {}) {
  const attributes = toModelAttributes(columns)
  const model = sequelize.define(name, attributes, defaultModelOptions)
  await model.sync({ force: overwrite, match: /_test/ })
  ...
}

function toModelAttributes (columns) {
  const attributes = {}
  for (const columnName in columns) {
    const column = columns[columnName]
    const attribute = mapColumnToAttribute(column, {})
    attributes[columnName] = attribute
  }
  return attributes
}

function mapColumnToAttribute (column, attribute) {
  const columnDef = {
    ...attribute,
    unique: column.index,  // <- THIS LINE CAUSES THE BUG TO APPEAR!!
    allowNull: !column.mandatory,
    defaultValue: column.default,
    type: mapColumnType(column.type, column.length),
    primaryKey: column.type === 'id'
  }

  return columnDef
}
```

## What do you expect to happen?
A table named `Foos` with the specified columns should be created in my database.

## What is actually happening?
I get the following error:

```
TypeError: Cannot read property 'msg' of undefined

      at _.each (node_modules/sequelize/lib/model.js:1010:48)
      at node_modules/lodash/lodash.js:4944:15
      at baseForOwn (node_modules/lodash/lodash.js:3001:24)
      at node_modules/lodash/lodash.js:4913:18
      at Function.forEach (node_modules/lodash/lodash.js:9359:14)
      at Function.refreshAttributes (node_modules/sequelize/lib/model.js:951:7)
      ...
```

## Why is this happening?
The error happens because Sequelize uses hasOwnProperty to check for the existence of `unique` but does not defend against `null` or `undefined`. This is the problematic code:

https://github.com/sequelize/sequelize/blob/21e830d361b785bab1156288219c661bb783df4f/lib/model.js#L997-L1015

## How can we fix it?
Add a check to line `997`, like so:

```js
if (definition.hasOwnProperty('unique') && 
    definition.unique != null && // <- This will fix the problem.
    definition.unique !== false) {
...
[...]

PR #8625, opened 2017-11-13
ignore null or undefined attributes.column.unique

arcdev1:
### Pull Request check-list

### Description of change

Closes: #8620
Explicitly passing a `null` or `undefined` `attributes.column.unique` value in `sequelize.define()` was causing an ambiguous error. Updated the code to ignore `null` or `underfined` `attributes.column.unique` values.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -994,7 +994,7 @@ class Model {
         this._defaultValues[name] = _.partial(Utils.toDefaultValue, definition.defaultValue);
       }
 
-      if (definition.unique) {
+      if (definition.hasOwnProperty('unique') && definition.unique) {
         let idxName;
         if (typeof definition.unique === 'object' && definition.unique.hasOwnProperty('name')) {
           idxName = definition.unique.name;
--- a/lib/model.js
+++ b/lib/model.js
@@ -337,7 +337,7 @@ class Model {
         const model = association.target;
         const as = association.options.as;
 
-        const predicate = {model};
+        const predicate = { model };
         if (as) {
           // We only add 'as' to the predicate if it actually exists
           predicate.as = as;
@@ -517,7 +517,7 @@ class Model {
 
 
       if (through.scope) {
-        include.through.where = include.through.where ? { [Op.and]: [include.through.where, through.scope]} :  through.scope;
+        include.through.where = include.through.where ? { [Op.and]: [include.through.where, through.scope] } : through.scope;
       }
 
       include.include.push(include.through);
@@ -548,7 +548,7 @@ class Model {
     }
 
     if (include.association.scope) {
-      include.where = include.where ? { [Op.and]: [include.where, include.association.scope] }:  include.association.scope;
+      include.where = include.where ? { [Op.and]: [include.where, include.association.scope] } : include.association.scope;
     }
 
     if (include.limit && include.separate === undefined) {
@@ -744,7 +744,7 @@ class Model {
     this.sequelize.runHooks('beforeDefine', attributes, options);
 
     if (options.modelName !== this.name) {
-      Object.defineProperty(this, 'name', {value: options.modelName});
+      Object.defineProperty(this, 'name', { value: options.modelName });
     }
     delete options.modelName;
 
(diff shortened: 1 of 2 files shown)


arcdev1:
The relevant change is at line `997` of model.js.

https://github.com/arcdev1/sequelize/blob/beb262c032e23bc614ea23d72f3bcff5ff224ee2/lib/model.js#L997-L1000

lib/model.js
```diff
@@ -994,7 +994,10 @@ class Model {
         this._defaultValues[name] = _.partial(Utils.toDefaultValue, definition.defaultValue);
       }
 
-      if (definition.hasOwnProperty('unique') && definition.unique !== false) {
+      if (definition.hasOwnProperty('unique') &&
+        definition.unique !== null &&
```
sushantdhiman: May be use https://lodash.com/docs/4.17.4#isEmpty

arcdev1: My understanding of the `unique` property is that it can either be a boolean value or a string. As per: http://docs.sequelizejs.com/class/lib/model.js~Model.html#static-method-init

My understanding of `_.isEmpty` is that it tests against objects, collections, maps, or sets. The example from the [lodash docs](https://lodash.com/docs/4.17.4#isEmpty) includes:

```js
_.isEmpty(true);
// => true
```

So, I don't think that `_.isEmpty` would work in this case. 

If we're looking to shorten the code, using a truthy check against null, as in `!=null` would eliminate the need to check for both `undefined` and `null` but it breaks an `eslint` rule, so I chose to be explicit rather than have to add an `eslint` exception.

The cosmetic changes happened because my editor autoformats on save based on the `eslintrc` file, but I agree that it makes the commit noisy. Do you want me to resubmit without any formatting changes?

arcdev1: Looks like `_.isNil` may do the trick! I'll use that and resubmit.
https://lodash.com/docs/4.17.4#isNil

arcdev1:
I backed out all cosmetic changes and, in the end, it appears all we need is a simple:

```js
if (definition.unique)
```

All tests pass.

sushantdhiman:
We definitely need `definition.hasOwnProperty('unique')`

arcdev1:
What is the scenario under which `definition.unique` would be `true` and `definition.hasOwnProperty('unique')` would be false?

I should add a test to catch any regression.

sushantdhiman:
> What is the scenario under which definition.unique would be true and definition.hasOwnProperty('unique') would be false?

Mostly a sanity check, although its possible that attribute object may contain some reference to other object that contains `unique`. It all depends upon how users(or may be web frameworks) are constructing attribute definition.

If you can add a test for this it will be great :)

arcdev1:
Hmm ... I can't seem to find a way that `hasOwnProperty()` returns `false` and `.unique` returns true.

Here's what I've tried.

```js
// simple class with no property defined
class Foo {
  constructor() {
  }
}

f = new Foo()
console.log(f.hasOwnProperty('unique'), !!f.unique) // false, false

// inheritance with sub-object
class Bar extends Foo {
  constructor() {
    super()
    this.subBar = {
      unique: true
    }
  }
}

b = new Bar()
console.log(!!b.unique, b.hasOwnProperty('unique')) // false, false

// basic object with sub object
baz = {
  bop: {
    unique: true
  }
}

console.log(!!baz.unique, baz.hasOwnProperty('unique')) // false, false

// class where unique is explicitly undefined
class Bop {
  constructor() {
    this.unique = undefined
  }
}

o = new Bop()
console.log(!!o.unique, o.hasOwnProperty('unique')) // false, true <- this was the original bug!

// class where unique is explicitly set
class Bap {
    constructor() {
    this.unique = true
  }
}

p = new Bap()
console.log(!!p.unique, p.hasOwnProperty('unique')) // true, true 

// subclass of explicit unique
class Zip extends Bap {
  
}

console.log(!!Zip.unique, Zip.hasOwnProperty('unique')) // false, false 
z = new Zip()
console.log(!!z.unique, z.hasOwnProperty('unique')) // true, true 
```

https://repl.it/repls/KhakiBitterEyra

sushantdhiman:
@arcdev1 No problem, we can focus on current issue for now

Issue #1498, opened 2014-03-12
support for DataTypes.ARRAY(DataTypes.ENUM('value 1', 'value 2')) in postgres


cjroth:
it would be great if this worked:

```
module.exports = function(sequelize, DataTypes) {
  return sequelize.define('Role', {
    // ...
    'my_column': {
      type: DataTypes.ARRAY(DataTypes.ENUM('value 1', 'value 2')),
      // values: ['value 1', 'value 2'], <- and this would be awesome as well!
    },
  }, {
    tableName: 'my_table',
  });
};
```

as of 1.7.0 it causes sequelize to throw `error: type "enum[]" does not exist`.

probably low priority because it can be emulated with the isIn validator and DataTypes.STRING

wavded:
any plans for this?

Hoverbear:
+1 This is important to me as well.

tiblu:
+1 - would be nice to have.

zoltan-tamasi:
I have the same problem.
I have enum types in postgres and some table columns with arrays of those enums like:

```
CREATE TYPE enum_type AS ENUM ('value1', 'value2');
CREATE TABLE table ( enum_array_type enum_type[] );

```
And what I wanted to do is to define a sequelize model like this:
```

sequlize.define('model', {
  'enum_array_type' : 'enum_type[]'
});
```
In this case it throws an error: "field.type.toSql is not a function" when I want to insert a new row...

aliatsis:
+1

rajat246:
+1. I also need this feature.

felixfbecker:
Please don't +1 issues. It clutters the thread without adding value to the discussion and spams maintainers with notifications. Use GitHub reactions to upvote features.
![image](https://cloud.githubusercontent.com/assets/10532611/18344735/bb60c2f6-75b8-11e6-88a9-822fae298110.png)

gabybrow:
Any new about this issue ? It's sad that we still don't have an answer :(

HarleyKwyn:
I just ran into this as well. Can we get support for this?

jameskennethrobinson:
Ditto for support.

xiaomingplus:
Ditto for support.

wavded:
It would be very helpful to have this.  I started a pull request.

PR #8629, opened 2017-11-13
feat(postgres): Support for ARRAY(ENUM) datatype.

wavded:
### Pull Request check-list

### Description of change

This commit addresses the issue raised in #1498.  I welcome feedback as I'm new to this code base.  After trying a few different ways to get this to work, this seemed the least invasive to me.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -974,6 +974,12 @@ class Model {
         this._dataTypeChanges[name] = definition.type._isChanged;
       }
 
+      // Mark array enum value for Postgres.
+      if (definition.type instanceof DataTypes.ARRAY &&
+        DataTypes.ARRAY.is(definition.type, DataTypes.ENUM)) {
+        definition.type.enumName = `"${this._schema || 'public'}"${this._schemaDelimiter || '.'}"enum_${this.tableName}_${name}"`;
+      }
+
       if (definition.type instanceof DataTypes.BOOLEAN) {
         this._booleanAttributes.push(name);
       } else if (definition.type instanceof DataTypes.DATE || definition.type instanceof DataTypes.DATEONLY) {
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -189,12 +189,18 @@ class QueryInterface {
       return Promise.all(promises).then(results => {
         const promises = [];
         let enumIdx = 0;
+        let type;
 
         for (i = 0; i < keyLen; i++) {
-          if (attributes[keys[i]].type instanceof DataTypes.ENUM) {
+          type = attributes[keys[i]].type;
+          // Handle arrays of enum type.
+          if (type instanceof DataTypes.ARRAY) type = type.type;
+
+          const attr = attributes[keys[i]].field || keys[i];
+          if (type instanceof DataTypes.ENUM) {
             // If the enum type doesn't exist then create it
             if (!results[enumIdx]) {
-              sql = this.QueryGenerator.pgEnum(tableName, attributes[keys[i]].field || keys[i], attributes[keys[i]], options);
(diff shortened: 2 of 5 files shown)


lib/query-interface.js
```diff
@@ -189,12 +189,21 @@ class QueryInterface {
...
         const promises = [];
         let enumIdx = 0;
+        let type;
 
         for (i = 0; i < keyLen; i++) {
-          if (attributes[keys[i]].type instanceof DataTypes.ENUM) {
+          type = attributes[keys[i]].type;
+          // Handle arrays of enum type.
+          if (type instanceof DataTypes.ARRAY) type = type.type;
+          // } else if (definition.type instanceof DataTypes.RANGE || DataTypes.ARRAY.is(definition.type, DataTypes.RANGE)) {
```
wavded: I will remove this line.  Just an artifact.

wavded:
The tests passed the first time this PR built, then I removed a comment, and Travis failed, seems unrelated.  Seems other PRs are suffering similar issues.

lib/model.js
```diff
@@ -974,6 +974,14 @@ class Model {
         this._dataTypeChanges[name] = definition.type._isChanged;
       }
 
+      // Mark array enum value for Postgres.
+      if (definition.type instanceof DataTypes.ARRAY &&
+        DataTypes.ARRAY.is(definition.type, DataTypes.ENUM)) {
+        definition.type.enumName = '"' + (this._schema || 'public')
```
sushantdhiman: Use template strings for better readability

lib/query-interface.js
```diff
@@ -189,12 +189,20 @@ class QueryInterface {
...
 
         for (i = 0; i < keyLen; i++) {
-          if (attributes[keys[i]].type instanceof DataTypes.ENUM) {
+          type = attributes[keys[i]].type;
+          // Handle arrays of enum type.
+          if (type instanceof DataTypes.ARRAY) type = type.type;
+
+          const attr = attributes[keys[i]].field || keys[i];
+          if (type instanceof DataTypes.ENUM) {
+            type._enumName = this.QueryGenerator.pgEnumName(tableName, attr, { schema: false });
```
sushantdhiman: Why you need to set this, `_enumName` used anywhere else or it should be `enumName` ?

lib/model.js
wavded: Will do, wasn't sure if you wanted to keep older Node compatibility or not.

lib/query-interface.js
wavded: Good catch, this isn't needed to be set, will clean that up.

wavded: Was left over code from a previous approach I had tried that I forgot to remove.

sushantdhiman:
Approach looks good. I have but one concern, we are now building enum name ourselves in current PR. When enum is may be user defined(unknown to sequelize) or name is different (or changed in future) how we are going to handle that.

May be at sequelize end we can use single method, may be from in ENUM type to get its name. Also allow users to override it if they want to use some other enum.

Also we need to integration tests as I mentioned earlier.

wavded:
@sushantdhiman Good points.  In addition to the enum name, I also discovered we need the oids for the custom enum types for the PG driver to correctly parse the array).  Any recommendations on where would be a good spot to hook into the initialization process to grab these details? (agree that a single method would be nice)

sushantdhiman:
I got a chance to work on `ARRAY(ENUM)` support. It will be added by https://github.com/sequelize/sequelize/pull/8738

Closing this PR, Thanks for this initiation :)

wavded:
@sushantdhiman much appreciated!!

PR #8709, opened 2017-11-28
ci(release): update to semantic-release v11

### LINKED ISSUE #8701: Release spam
When I added semantic-release a year ago, the goal was to ensure we follow semantic versioning 100%, to release fast and to always have proper changelogs. I think the current situation takes it to far - if I look at the release page, the last 6 releases have **no changelog** (i.e. no relevant changes) at all.
- `v4.23.1`: Corrects some typos in a markdown file https://github.com/sequelize/sequelize/commit/b8cfde9d67f70b742b72113f4e0149880030ecc8
- `v4.23.0`: (a minor, i.e. **feature** release): Adds ESLint rules https://github.com/sequelize/sequelize/commit/703d4feb16849fe13500296f31411f216428d478
- `v4.22.16`: Fixes a typo in a markdown file https://github.com/sequelize/sequelize/commit/2b3193045712f8bb031dddcd9c6475f53b852db1
- `v4.22.15`: Changes some example in a markdown file https://github.com/sequelize/sequelize/commit/398c1ce7226b401ff9db1e01cd39f7446c671bdd
- `v4.22.14`: Updates markdown files https://github.com/sequelize/sequelize/commit/7a801408b48661ccc9162dfaaffd71be640e0a59
- `v4.22.13`: Updates markdown files https://github.com/sequelize/sequelize/commit/7fe6da1e42f7e48e07a2aafe88813040616aa42f

The goal of semantic-release and semantic versioning (which the readme says Sequelize follows) is 

> Given a version number MAJOR.MINOR.PATCH, increment the:
> 
> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards-compatible manner, and
> PATCH version when you make backwards-compatible bug fixes.

It is _not_ to release on _every_ commit, it is to release on every commit _that fulfils one of the above criteria_. With these criteria, none of the above examples should have triggered any kind of release.
[And configuring semantic-release to have a possibility to trigger a manual, unsemantic release](https://github.com/sequelize/sequelize/blob/b8cfde9d67f70b742b72113f4e0149880030ecc8/package.json#L154) is [intentionally disallowed by semantic-release](https://github.com/semantic-release/semantic-release#can-i-manually-trigger-the-release-of-a-specific-version):

> ### Can I manually trigger the release of a specific version?
> 
> You can trigger a release by pushing to your GitHub repository. You deliberately cannot trigger a _specific_ version release, because this is the whole point of `semantic-release`. Start your packages with `1.0.0` and semver on.

Everyone who has something like Greenkeeper, dependencies.io or Sibbel set up currently gets countless PRs or notifications for versions that contain no changes in behaviour. Libraries that depend on Sequelize without a lockfile can rarely cache the dependency.

I would like to propose to go back to the semantic-release default settings to reduce the spam and follow semver correctly.

Documentation is released on every commit, markdown should not be included in the npm package and other releases are frequent enough to get docblock updates in (and not every editor uses them anyway, Visual Studio for example uses TypeScript typings).

A refactor should never trigger a release as its definition is to not change any behavior - if it introduces a bug, there is no gain in getting that out fast. If it gets out, it will in the next release, and users can still rollback.

As I recently also worked a lot with the new semantic-release v11, if the current maintainers are fine with this, I would be happy to prepare a PR to update to that and restore the defaults.

felixfbecker:
and restore semantic-release defaults

closes #8701


.travis.yml
```diff
@@ -47,43 +47,44 @@ script:
 
 jobs:
   include:
-    - stage: tests
-      node_js: 4
+    - stage: test
```
felixfbecker: renamed this because it's the travis default

.travis.yml
```diff
@@ -47,43 +47,44 @@ script:
...
-      node_js: 6
+      node_js: '6'
       script:
-        - 'if [ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${TRAVIS_BRANCH}" = "master" ]; then npm run docs; fi'
+        - npm run docs
+    - stage: release
+      node_js: '8'
+      script:
+        - npm run semantic-release
+      before_deploy:
```
felixfbecker: put docs in the release stage because it doesn't have to spin up a new job

.travis.yml
```diff
@@ -47,43 +47,44 @@ script:
...
-    - stage: tests
-      node_js: 6
+    - stage: test
+      node_js: '6'
       env: DIALECT=sqlite
     - stage: docs
-      node_js: 6
+      node_js: '6'
       script:
-        - 'if [ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${TRAVIS_BRANCH}" = "master" ]; then npm run docs; fi'
```
felixfbecker: this is replaced by `stages` `condition` at the bottom

.travis.yml
```diff
@@ -47,43 +47,44 @@ script:
...
         project: ./esdoc/
         domain: docs.sequelizejs.com
         skip_cleanup: true
-        on:
-          branch: master
-    - stage: release
-      node_js: 8
-      env: BUILD_LEADER_ID=8
-      script:
-        - npm install --no-save --no-package-lock semantic-release@8 @bubltechnology/customizable-commit-analyzer@1.0.2-0
```
felixfbecker: moved the dependency to package.json so we can get Greenkeeper updates

.travis.yml
```diff
@@ -47,43 +47,44 @@ script:
...
       deploy:
         provider: surge
         project: ./esdoc/
         domain: docs.sequelizejs.com
         skip_cleanup: true
-        on:
-          branch: master
-    - stage: release
-      node_js: 8
-      env: BUILD_LEADER_ID=8
```
felixfbecker: not needed anymore in semantic-release v11

package.json
```diff
@@ -144,19 +146,12 @@
...
-      "feat": "minor",
-      "refactor": null,
-      "release": "minor",
-      "ci": null,
-      "tests": null
     }
   },
   "release": {
-    "analyzeCommits": "@bubltechnology/customizable-commit-analyzer"
+    "verifyConditions": [
```
felixfbecker: need to override this to exclude `@semantic-release/condition-travis` - we use build stages to ensure the release only happens on the right branch, after after all tests passed etc.

package.json
```diff
@@ -144,19 +146,12 @@
...
-    },
-    "commitTypeMap": {
-      "docs": null,
-      "fix": "patch",
-      "perf": "patch",
-      "feat": "minor",
-      "refactor": null,
-      "release": "minor",
-      "ci": null,
-      "tests": null
```
felixfbecker: restored semantic-release defaults, see #8701

.travis.yml
```diff
@@ -13,7 +13,7 @@ branches:
 
 cache:
   directories:
-    - node_modules
+    - $HOME/.npm
```
felixfbecker: Changed this because caching `node_modules` is dangerous, e.g. it doesn't work with native addons. Caching the npm cache itself is always safe.

package.json
```diff
@@ -144,19 +146,12 @@
...
     }
   },
   "release": {
-    "analyzeCommits": "@bubltechnology/customizable-commit-analyzer"
```
sushantdhiman: I used that commit analyzer to prevent that annoying `ENOCHANGE`, I sure you are aware of that issue. Can you set custom rules so every fix/feat commits are always released no matter what.

https://github.com/semantic-release/semantic-release/issues/192#issuecomment-333328071

eseliger:
has conflicts

package.json
felixfbecker: Not sure what you mean. The comment you link just describes the behaviour as I would expect, and the issue is closed. Maybe the issue was fixed? I have never had any problem with an ENOCHANGE error (I have a lot of repos with semantic release set up).

sushantdhiman: `ENOCHANGE` appears when semantic release think there are not enough changes in code to justify a new release, but actually there are. Its still there https://github.com/semantic-release/semantic-release/blob/eec4e68615c75963e06bac810b2845cc4dee02dd/index.js#L39

This is a really annoying issue but could be fixed easily by using a commit analyzer, like I used `@bubltechnology/customizable-commit-analyzer`. My goal was to have it release on `fix` / `feat` no matter what.

If you dont have any suitable suggestion than current commit analyzer, please keep current commit analyzer.

felixfbecker: Sorry, I don't understand. What is the issue? semantic-release is all about semantically determining _whether_ and _what_ release should happen on a build. With the default configuration, a `fix` triggers a patch release and a `feat` commit triggers a minor release. If there was only a `chore`, `ci`, `tests`, `docs` etc commit then there were no relevant changes, hence ENOCHANGE. That's how it's supposed to work.

felixfbecker: This is the default implementation: https://github.com/semantic-release/commit-analyzer/

sushantdhiman: > With the default configuration, a fix triggers a patch release and a feat commit triggers a minor release. 

This is not true for all cases, sometimes semantic release thinks there are not enough changes in code which could justify a new release (even if commit is `fix / feat`), hence `ENOCHANGE`.

Please keep this commit analyzer implementation (or use any official method) so all `fix / feat` commits are releases no matter what

felixfbecker: I don't see any evidence for why that would be the behaviour. If it is like that, it is a bug (that should be reported and fixed). But the issue you linked is closed, the default commit analyzer is depended on by hundreds on projects, got a lot of updates since we added semantic-release and has 100% test coverage. 

`@bubltechnology/customizable-commit-analyzer` on the other hand doesn't even have a stable release, way fewer dependents, no integration tests and has not any more customizability than the official implementation. It's also not even compatible with semantic-release v11, as v11 requires plugins to return a Promise, while that plugin still tries to call a callback.

I have high confidence that the official implementation works correctly.

sushantdhiman: Ok lets merge this. If it worked as we expected it to (release on every fix / feat commit) then great, otherwise we can figure something out.

PR #8710, opened 2017-11-28
ci(appveyor): update config

felixfbecker:
Testing whether this makes the build faster.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,19 +1,21 @@
 version: '{build}'
 
+image: Visual Studio 2017
+
 platform:
   - x64
 
 services:
   - mssql2016
 
-shallow_clone: true
+clone_depth: 1
 
 cache:
-  - node_modules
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -1,19 +1,21 @@
...
   - x64
 
 services:
   - mssql2016
 
-shallow_clone: true
+clone_depth: 1
 
 cache:
-  - node_modules
```
sushantdhiman: keep `node_modules` as well

eseliger: caching node_modules isn't safe, I would suggest caching only the npm cache

sushantdhiman: As long as we are caching modules some where I have no problem, main goal is fast build startup

felixfbecker:
Decided to not fix what ain't broke. Not worth to investigate the timezone issue.

PR #8703, opened 2017-11-28
add: support for ARRAY(ENUM) (v3)

sushantdhiman:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](../CONTRIBUTING.md)?

### Description of change

Add support for `ARRAY(ENUM)`

https://github.com/sequelize/sequelize/issues/1498

PR #8733, opened 2017-12-01
fix(postgres/connection-manager): Only fetch OIDs once

janmeier:
Fixes a bug introduced (or surfaced at least) by https://github.com/sequelize/sequelize/pull/8703

If `HSTORE` is not enabled in the DB, `SELECT typname, typtype, oid, typarray FROM pg_type WHERE typname = 'hstore'` will not return any rows, so `dataTypes.HSTORE.types.postgres.oids.length` will remain 0. This means that we will try to fetch the oids for each new connection.

After the change that also selects OIDs for enums, any enum OIDs will be pushed onto the ENUM type each time. This means the list of enum OIDs will keep growing, and so will the CPU time spent in `$refreshTypeParser`


test/integration/dialects/postgres/connection-manager.test.js
```diff
@@ -25,12 +25,5 @@ if (dialect.match(/^postgres/)) {
     it('should correctly parse the moment based timezone', function() {
       return checkTimezoneParsing(this.sequelize.options);
     });
-
-    it('should correctly parse the moment based timezone while fetching hstore oids', function() {
-      // reset oids so we need to refetch them
```
janmeier: No need to reset, this will be reset when creating a new sequelize instance, with a new connection manager

PR #8735, opened 2017-12-01
feat(model): support paranoid updates

### LINKED ISSUE #8338: Model.update does not take into account paranoid
## What are you doing?

```js
// define a paranoid model
const User = sequelize.define('user', {/* ... */}, { paranoid: true });

// run an update query on said model
User.update({/* ... */}, { where: {/* ... */} });
```

The resulting `UPDATE` query should only affect non-deleted records. In other words, I would expect a `WHERE deleted_at IS NOT NULL` clause.

At the very least, I would expect `Model.update` to accept a `paranoid` option much like the `Model.find` family of functions do.

No `deleted_at` clause is added. Adding `paranoid: true` also has no affect.

__Dialect:__ postgres
__Dialect version:__
__Database version:__ 9.6
__Sequelize version:__ 4.8.3
__Tested with master branch:__ No (but relevant code hasn't changed on master)

gabegorelick:
Closes #8338

### Pull Request check-list

### Description of change

This PR makes `Model.update` respect `paranoid`. See #8338.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2589,6 +2589,7 @@ class Model {
    * @param  {Object}       values
    * @param  {Object}       options
    * @param  {Object}       options.where                   Options to describe the scope of the search.
+   * @param  {Boolean}      [options.paranoid=true] If true, only non-deleted records will be updated. If false, both deleted and non-deleted records will be updated. Only applies if `options.paranoid` is true for the model.
    * @param  {Array}        [options.fields]                Fields to update (defaults to all fields)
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -2589,6 +2589,7 @@ class Model {
    * @param  {Object}       values
    * @param  {Object}       options
    * @param  {Object}       options.where                   Options to describe the scope of the search.
+   * @param  {Boolean}      [options.paranoid=true] If true, only non-deleted records will be updated. If false, both deleted and non-deleted records will be updated. Only applies if `options.paranoid` is true for the model.
```
gabegorelick: Wasn't sure if you want these comments aligned or not. Some are and some aren't.

PR #8740, opened 2017-12-02
fix(model): ordering by renamed sub-query attribute

### LINKED ISSUE #8739: Model.findAll with a hasMany relationship, order, and limit fails with custom field name
## What are you doing?

```js
  name: {
    type: Sequelize.STRING,
    field: 'some_other_name'
  }
});
const Project = this.sequelize.define('project', { title: Sequelize.STRING });
User.hasMany(Project);

return User.sync({ force: true })
  .then(() => Project.sync({ force: true }))
  .then(() => {
    return User.findAll({
      order: ['name'],
      limit: 1,
      include: [Project]
    });
  });
```

No error

Query generated is
```sql
SELECT "user".*, "projects"."id" AS "projects.id", "projects"."title" AS "projects.title", "projects"."createdAt" AS "projects.createdAt", "projects"."updatedAt" AS "projects.updatedAt", "projects"."userId" AS "projects.userId" 
FROM (SELECT "user"."id", "user"."some_other_name" AS "name", "user"."createdAt", "user"."updatedAt" FROM "users" AS "user" ORDER BY "user"."some_other_name" LIMIT 1) AS "user"
LEFT OUTER JOIN "projects" AS "projects" ON "user"."id" = "projects"."userId"
ORDER BY "user"."some_other_name";
```

which causes the following error since the subquery renames `user.some_other_name`, which is used by the `ORDER BY` clause, into `user.name`:

> SequelizeDatabaseError: column user.some_other_name does not exist

To work around this, you can manually project the columns Sequelize needs with `attributes: [['some_other_name', 'name'], ['some_other_name', 'some_other_name']],`. 

__Dialect:__ any
__Dialect version:__ any
__Database version:__ any
__Sequelize version:__ master

gabegorelick:
Closes #8739

### Pull Request check-list

### Description of change

Use correct column name when ordering by an attribute projected from a sub-query. See #8739.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1688,6 +1688,16 @@ const QueryGenerator = {
         ) {
           subQueryOrder.push(this.quote(order, model, '->'));
         }
+
+        if (subQuery) {
+          // Handle case where sub-query renames attribute we want to order by,
+          // see https://github.com/sequelize/sequelize/issues/8739
+          const subQueryAttribute = options.attributes.find(a => Array.isArray(a) && a[0] === order[0] && a[1]);
+          if (subQueryAttribute) {
+            order[0] = new Utils.Col(subQueryAttribute[1]);
+          }
+        }
+
         mainQueryOrder.push(this.quote(order, model, '->'));
       }
     } else if (options.order instanceof Utils.SequelizeMethod) {
--- a/test/integration/model/findAll.test.js
+++ b/test/integration/model/findAll.test.js
@@ -665,6 +665,125 @@ describe(Support.getTestDialectTeaser('Model'), () => {
             expect(workers[0].tasks[0].title).to.equal('homework');
           });
         });
+
+        // https://github.com/sequelize/sequelize/issues/8739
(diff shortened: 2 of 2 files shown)


lib/dialects/abstract/query-generator.js
```diff
@@ -1688,6 +1688,16 @@ const QueryGenerator = {
         ) {
           subQueryOrder.push(this.quote(order, model, '->'));
         }
+
+        if (subQuery) {
+          // Handle case where sub-query renames attribute we want to order by,
+          // see https://github.com/sequelize/sequelize/issues/8739
+          const subQueryAttribute = options.attributes.find(a => Array.isArray(a) && a[0] === order[0] && a[1]);
+          if (subQueryAttribute) {
+            order[0] = new Utils.Col(subQueryAttribute[1]);
```
sushantdhiman: These wont work when order contain multiple attributes being renamed by subQuery. Please add a test for two or more attributes

gabegorelick: @sushantdhiman I believe it actually will work since we loop over each `order` clause above. I added some tests which should confirm that. Let me know if I'm mistaken.

PR #8742, opened 2017-12-03
fix(query-interface): add ability to pass options from createFunction() call

### LINKED ISSUE #8741: queryInterface.createFunction() does not support sequelize.query options
## What are you doing?
```js
queryInterface.createFunction('test', [], 'integer', 'plpgsql', 'RETURN 1;', {logging: console.log})
```

Logged SQL:
```
Executing (default): CREATE FUNCTION test()
RETURNS integer AS $func$
BEGIN
        RETURN 1;
END;
$func$ language 'plpgsql';
```

```
sequelize\lib\dialects\postgres\query-generator.js:687
      '' : '\n\t' + options.join('\n\t');
                            ^
TypeError: options.join is not a function
```

__Dialect:__ postgres
__Dialect version:__ pg@6
__Database version:__ 10.0
__Sequelize version:__ 4.24.0

markablov:
Closes #8741

### Pull Request check-list

### Description of change

queryInterface.createFunction() does not support ability to pass options to sequelize.query().
I extended signature of this function similar to createTrigger() approach - separate extended SQL function description options with options that passed to sequelize.query().

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -1122,6 +1122,35 @@ class QueryInterface {
     }
   }
 
+  /**
+   * Create SQL function
+   *
+   * ```js
+   * queryInterface.createFunction(
+   *   'someFunction',
+   *   [
+   *     {type: 'integer', name: 'param', direction: 'IN'}
+   *   ],
+   *   'integer',
+   *   'plpgsql',
+   *   'RETURN param + 1;',
+   *   [
+   *     'IMMUTABLE',
+   *     'LEAKPROOF'
+   *   ]
+   * );
+   * ```
+   *
+   * @param {String} functionName Name of SQL function to create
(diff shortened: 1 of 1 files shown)


lib/query-interface.js
```diff
@@ -1122,8 +1122,8 @@ class QueryInterface {
     }
   }
 
-  createFunction(functionName, params, returnType, language, body, options) {
```
sushantdhiman: Since you are using functions, can you add proper comments for this method options and usage, something like https://github.com/sequelize/sequelize/blob/master/lib/query-interface.js#L157

markablov: done

sushantdhiman: Great :+1:

PR #8748, opened 2017-12-04
feat(postgres): returning option for arithmetic queries (#8727)

### LINKED ISSUE #8727: No option to disable returning on arithmetic queries (Postgres)
## What are you doing?
I ran into this weird edge case debugging a beforeUpdate hook that uses the increment method, and having it behave unpredictably. A minimal example looks like this:

```js
const Article = db.define('article', {
  name: Sequelize.STRING,
  version: {
    type: Sequelize.INTEGER,
    defaultValue: 0,
  }
});

Article.beforeUpdate(article => {
  article.increment('version');
  // If you return this, the name will not be updated...
});

db.sync({ force: true })
  .then(() => Article.create({ name: 'old name' }))
  .then(article => article.update({ name: 'new name' }))
  .then(updatedArticle => {
    console.log('name:', updatedArticle.name); // 'old name'
    console.log('version:', updatedArticle.version); // 1
    // the row in the database will have the correct name, though!
  });
```
I logged the queries and figured out it was appending `RETURNING *` to the increment query, so the instance values were being overwritten, including any pending updates. After poking around in the source, arithmetic queries automatically have the returning clause appended if it's supported by the dialect.

Now I know this use case is a poor example, but wouldn't it make more sense to have returning be optional? It could still default to true, but I feel like you should still be able to toggle it with an option flag, like with other queries. Either way, I think it would be worth updating the docs to reflect this behavior.

The increment method to not overwrite the instance.

The default behavior is that increment returns the db values and overwrite the instance, which is undocumented.

__Dialect:__ postgres
__Sequelize version:__ 4.23.1

weblemur:
### Pull Request check-list

### Description of change

Add handling for a `returning` option in arithmetic queries (`increment` and `decrement`). Defaults to `true` to maintain expected behavior.

Closes #8727

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -389,7 +389,7 @@ const QueryGenerator = {
   */
   arithmeticQuery(operator, tableName, attrValueHash, where, options, attributes) {
     options = options || {};
-    _.defaults(options, this.options, { returning: true });
+    _.defaults(options, { returning: true });
 
     attrValueHash = Utils.removeNullValuesFromHash(attrValueHash, this.options.omitNull);
 
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -388,13 +388,16 @@ const QueryGenerator = {
    @private
   */
   arithmeticQuery(operator, tableName, attrValueHash, where, options, attributes) {
+    options = options || {};
+    _.defaults(options, this.options, { returning: true });
+
     attrValueHash = Utils.removeNullValuesFromHash(attrValueHash, this.options.omitNull);
 
     const values = [];
     let query = 'UPDATE <%= table %> SET <%= values %><%= output %> <%= where %>';
     let outputFragment;
 
-    if (this._dialect.supports.returnValues) {
+    if (this._dialect.supports.returnValues && options.returning) {
       if (this._dialect.supports.returnValues.returning) {
         options.mapToModel = true;
         query += ' RETURNING *';
--- a/test/integration/instance.test.js
+++ b/test/integration/instance.test.js
@@ -156,6 +156,9 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
         return this.User.findById(1).then(user1 => {
           return user1.increment('aNumber', { by: 2 }).then(() => {
             expect(user1.aNumber).to.be.equal(2);
+            return user1.increment('bNumber', { by: 2, returning: false }).then(user3 => {
+              expect(user3.bNumber).to.be.equal(0);
+            });
           });
         });
       });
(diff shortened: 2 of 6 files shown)


lib/dialects/abstract/query-generator.js
```diff
@@ -388,13 +388,16 @@ const QueryGenerator = {
    @private
   */
   arithmeticQuery(operator, tableName, attrValueHash, where, options, attributes) {
+    options = options || {};
+    _.defaults(options, this.options, { returning: true });
+
```
weblemur: Not sure if this should be wrapped in conditional. I figured since the returning flag is unused by arithmetic queries other dialects, it wouldn't have a negative impact.

eseliger:
technically, this would be a breaking change. 
@sequelize/orm what do you guys think?

weblemur:
How so? The default behavior is unchanged.

lib/dialects/abstract/query-generator.js
eseliger: I think its fine, in the future, if a dialect is added that has support for it or if eg. mysql gets support for it, it will be all set.

eseliger:
Sorry, got this wrong.

test/integration/instance.test.js
```diff
@@ -323,6 +326,9 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
         return this.User.findById(1).then(user1 => {
           return user1.decrement('aNumber', { by: 2 }).then(() => {
             expect(user1.aNumber).to.be.equal(-2);
+            return user1.decrement('bNumber', { by: 2, returning: false }).then(user3 => {
+              expect(user3.bNumber).to.be.equal(0);
+            });
```
sushantdhiman: Add a test that it returns updated instance by default. `returnValues.returning` is enabled only for postgres so it should not fail for other dialect.

weblemur: The outer test accounts for the default case (`expect(user1.aNumber).to.be.equal(-2)`). If the code is unclear, I could break out the `returning: false` test case into a seperate block. I was mostly just trying to follow the existing test structure which is why I ended up nesting the tests. (I prefer chaining promises rather than nesting them anyways.)

lib/dialects/abstract/query-generator.js
sushantdhiman: Why you need `this.options` ?

test/integration/instance.test.js
sushantdhiman: No thats ok, I totally missed that

lib/dialects/abstract/query-generator.js
weblemur: Good question. It probably doesn't. I was mostly following existing patterns, and I was unable to find why other methods map defaults from `this.options`. I figured it was a gap in my own understanding and that there may be cases where options get attached to the query generator object which would need to get accounted for. I'm unaware where/how that would happen though, so I figured better safe than sorry.

sushantdhiman: Remove `this.options` then

weblemur: Done!

PR #8753, opened 2017-12-05
Added _scopeName property in instance

### LINKED ISSUE #7433: Get used scope name in hooks?

oscarr-reyes:
*The name of the used scope is now stored in a instance property "_scopeNames" which contains the list of all scope used in the query

Resolves #7433 

### Pull Request check-list

### Description of change

Property `_scopeNames` is added to the instance, this is useful when in need to check for the used scope in hooks, this holds a lists of all used scopes. If `scope()` was not called then `_scopeNames` holds `defaultScope` as the only item in the list.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -842,6 +842,7 @@ class Model {
     this._findAutoIncrementAttribute();
 
     this._scope = this.options.defaultScope;
+    this._scopeNames = ['defaultScope'];
 
     if (_.isPlainObject(this._scope)) {
       this._conformOptions(this._scope, this);
@@ -1327,6 +1328,7 @@ class Model {
     Object.defineProperty(self, 'name', {value: this.name});
 
     self._scope = {};
+    self._scopeNames = [];
     self.scoped = true;
 
     if (!option) {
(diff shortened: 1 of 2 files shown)


eseliger:
Don't you think a empty scope list would be a as good representation for the "defaultScope"? 
this technically means that the scope with the name `"defaultScope"` cannot be distinguished. While this might not happen very often, it still feels better if it would be empty then. 
@sequelize/orm what do you guys think?

sushantdhiman:
@eseliger default scoped is added as `defaultScope` name ([docs](http://docs.sequelizejs.com/manual/tutorial/scopes.html)) so I think it make sense to have it in array with such name. But lets wait for comments from other maintainers

eseliger:
Hmm you are right, forgot about that. So it cannot be distinguished anyways as its always used as the default.
I would be fine with both then.

felixfbecker:
Why would an underscore-prefixed property be considers a feature or public API?

Issue #311, opened 2012-11-04
Composite foreign keys


Zerzio:
Hi,

I found some posts about composite keys and relations but nothing is in the documentation and it doesn't seem to be supported (no relevant test cases)
Is this supported: Entity.hasMany(Other , {foreignKey: ['key1', ' key2', ' key3']});

Thanks,

Serge

kopertop:
I just ran into this problem. Is there still no plan to support multi-column indexes on relations? Is there any workaround known other then using a single key index?

Specifically, I'm trying to use the "findAll" with an "include" filter.

mickhansen:
@kopertop There's a difference between indexes and constraints, we support multi column indexes on all tables.

m4rcelofs:
+1

ivanjd:
That feature would be super helpful.

jamesaspence:
+1 I just ran into this as well, would make my job way easier. For now I guess I'm editing the table structure :/

pedrombafonso:
+1

antoineDeRengerve:
+1

nilpotence:
+1

markdboyd:
+1

mdeltito:
+1

TiagoDeus:
+1

cdupetit:
+1

ozum:
:+1:

JacopoDaeli:
+1

mickhansen:
For everyone +1'ing this, are you looking to reference tables with multiple primary keys or what are the usecases here exacttly?

cdupetit:
I want to make association like hasMany, hasOne, belongsTo between table with multiple primary keys.
Example:
```
Site.hasMany(Server, { primaryKey: [siteId] });
Server.hasMany(Share, { primaryKey: [siteId, serverId] });
Share.hasMany(Folder, { primaryKey: [siteId, serverId, shareId] });
```

ozum:
Hi,

I'm developer of pg-generator which has a built-in [sequelize template](http://www.pg-generator.com/builtin-templates/sequelize/) to generate sequelize code for PostgreSQL.

My use case is;
Sometimes, I have tables with composite primary keys (multi column PK), which I can not use surrogate keys without breaking normalization.

Below is a very simplified example (It is not easy to draw a diagram with ASCII)

```
account                                   action
-------                                   --------
(PK FK) company_id    -------------<      (PK) id
(PK FK) person_id                         (FK) company_id
[...]

PR #8763, opened 2017-12-07
RFC: Allow "tenant keys" in associations, to query on composite foreign keys

stevehollaar:
### Pull Request check-list

### Description of change

Related to https://github.com/sequelize/sequelize/issues/311

This adds a new option key to `BelongsTo`, `HasMany`, and `HasOne` associations to allow querying and joining on tables that share a composite key. The motivation here is to allow sharding by a "tenant ID" in [Citus terminology](https://docs.citusdata.com/en/v7.0/sharding/data_modeling.html#distributing-by-tenant-id), but should be generalizable for any use of composite 2-column foreign keys.

With this PR:
```js
FooModel.Bar = FooModel.belongsTo(BarModel, {
  tenantKeys: {
    foreign: 'user_id',
    target: 'user_id',
  }
})
```
Will allow:
```js
fooInstance.getBar()
```
to query:
```
SELECT * FROM bar WHERE bar.id = <foo.bar_id> AND bar.user_id = <foo.user_id>
```
and:
```js
FooModel.findAll({ include: [BarModel] })
```
to query:
```
SELECT * FROM foo LEFT OUTER JOIN bar ON (foo.bar_id = bar.id AND foo.user_id = bar.user_id)
```

### DIFF
--- a/lib/associations/belongs-to.js
+++ b/lib/associations/belongs-to.js
@@ -45,6 +45,14 @@ class BelongsTo extends Association {
       if (!this.tenantKeys.target || !this.tenantKeys.foreign) {
         throw new Error('tenantKeys for a BelongsTo association must have target and foreign keys');
       }
+
+      if (!source.rawAttributes[this.tenantKeys.foreign]) {
+        throw new Error('Tenant key: ' + this.tenantKeys.foreign + ' is not an attribute of ' + source.name);
+      }
+
+      if (!target.rawAttributes[this.tenantKeys.target]) {
+        throw new Error('Tenant key: ' + this.tenantKeys.target + ' is not an attribute of ' + target.name);
+      }
     }
 
     if (!this.foreignKey) {
@@ -89,6 +97,13 @@ class BelongsTo extends Association {
       type: this.options.keyType || this.target.rawAttributes[this.targetKey].type,
       allowNull: true
     });
+
+    if (this.tenantKeys) {
+      newAttributes[this.tenantKeys.foreign] = {
+        name: this.tenantKeys.foreign,
+        type: this.source.rawAttributes[this.tenantKeys.foreign].type,
+      };
+    }
 
     if (this.options.constraints !== false) {
       const source = this.source.rawAttributes[this.foreignKey] || newAttributes[this.foreignKey];
--- a/lib/associations/belongs-to.js
+++ b/lib/associations/belongs-to.js
@@ -22,6 +22,7 @@ class BelongsTo extends Association {
     this.associationType = 'BelongsTo';
     this.isSingleAssociation = true;
     this.foreignKeyAttribute = {};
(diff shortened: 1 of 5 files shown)


janmeier:
@sushantdhiman What do you think about this?

sushantdhiman:
What this PR try to achieve should ideally be possible with just scopes. But `scopes` are rather limited when comes to dynamic resolution, its still possible to do so by scope using `sequelize.col` and `as` but not that clean and rather limited.

As far as I understood this change, API looks limited to me

1) It only supports two column composite key
2) First source column is assumed to be primary key or `id`

After going through #311, I think @mickhansen suggested a good design https://github.com/sequelize/sequelize/issues/311#issuecomment-218080573

```js
hasMany(Target, {
  foreignKey: {
    key1: 'id1',
    key2: 'id2'
  }
});
```

Although I would like to mention https://github.com/sequelize/sequelize/issues/311#issuecomment-193822332

```js
An array seems right foreignKey: ['key1', ' key2', ' key3'] and in case of the belongsTo with targetKey -> foreignKey: ['key1', ' key2', ' key3'], targetKey: ['targetKey1ToMatchWithKey1', 'targetKey2ToMatchWithKey2', 'targetKey3ToMatchWithKey3']
```

Only because even I find my self confused with source/target notation, so some users may got confused with `{ key: id }` while array approach properly separates `foreignKey / targetKey` 

I think we can go with either of these, upto you guys @sequelize/orm 

```js
hasMany(Target, {
  foreignKeys: ['key1', 'key2', ....],
  targetKeys: ['id1', 'id2', ....]
});

OR

hasMany(Target, {
  foreignKeys: {
      'sourceKey1': 'targetKey1',
      'sourceKey2': 'targetKey2'
  }
});

OR Even 

// I like this one, I realise this is just a variant of Jan's proposal :D
hasMany(Target, {
  foreignKeys: [{
      foreignKey: '',  // can be string or attribute
      targetKey: '', // can be string or attribute
  }, {....}, .....]
});

```


> This will however clash a bit with how this works currently (conceptually at least). Right now foreignKey on specifies one side of the relation ..

Internally they are linked with primaryKey of other side so relation was already there, now we are trying to expose it

jedwards1211:
"tenant" is definitely too specific to the author's use case.  There are many other potential use cases for composite foreign keys that have nothing to do with supporting multi-tenancy.

jedwards1211:
I think that there shouldn't be a separate `extraForeignKeys` variable of any kind, and the existing case (single foreign key) should be treated as a subset of the compound foreign key case.

But, that would be a breaking change, and have to go in version 5.  What do you guys think?

papb:
@stevehollaar Are you still able to pursue this?

sushantdhiman:
Closing this PR for inactivity

PR #8786, opened 2017-12-15
 fix(mssql): dateonly value should be treated as utc when parsing

ratanakvlun:
### Pull Request check-list

### Description of change

MSSQL date only fields are being parsed to local date only strings. This is incorrect behavior because a date only string is only useful as it is. The problem is that `tedious` returns a Date object created with the date only value from the database (which it really shouldn't do). When formatting this Date object into a string using `moment`, the local time is used to determine the result:

```
let dbValue = '2017-12-15';
let tdsValue = new Date(dbValue);
// Date object as UTC time: 2017-12-15 00:00:00
// Assuming EST, Date object as local time: 2017-12-14 19:00:00 -0500
moment(tdsValue).format('YYYY-MM-DD'); // Outputs: 2017-12-14
```

We can fix this by changing the date mode:

```
moment.utc(tdsValue).format('YYYY-MM-DD'); // Outputs: 2017-12-15
```

This creates a `moment` object in UTC mode which uses the UTC time when it produces a date only string.

### DIFF
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -478,8 +478,10 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
     const Model = this.sequelize.define('user', {
       stamp: Sequelize.DATEONLY
     });
+
     const testDate = moment().format('YYYY-MM-DD');
     const newDate = new Date();
+    const dateOnly = moment(newDate).format('YYYY-MM-DD');
 
     return Model.sync({ force: true})
       .then(() => Model.create({ stamp: testDate }))
@@ -508,7 +510,7 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
         return record.reload();
       }).then(record => {
         expect(typeof record.stamp).to.be.eql('string');
-        expect(new Date(record.stamp)).to.equalDate(newDate);
+        expect(record.stamp).to.equal(dateOnly);
       });
   });
 
--- a/test/integration/instance/values.test.js
+++ b/test/integration/instance/values.test.js
@@ -122,7 +122,7 @@ describe(Support.getTestDialectTeaser('DAO'), () => {
             // so we must create a record with the right value for always_false, then reference it in an update
             let now = dialect === 'sqlite' ? self.sequelize.fn('', self.sequelize.fn('datetime', 'now')) : self.sequelize.fn('NOW');
             if (dialect === 'mssql') {
-              now = self.sequelize.fn('', self.sequelize.fn('getdate'));
+              now = self.sequelize.fn('', self.sequelize.fn('getutcdate'));
             }
             user.set({
               d: now,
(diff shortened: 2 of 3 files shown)


test/integration/data-types.test.js
```diff
@@ -478,8 +478,10 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
     const Model = this.sequelize.define('user', {
       stamp: Sequelize.DATEONLY
     });
+
     const testDate = moment().format('YYYY-MM-DD');
     const newDate = new Date();
+    const dateOnly = moment(newDate).format('YYYY-MM-DD');
```
janmeier: No need to create a new variable, you can just use `testDate`

test/integration/instance/values.test.js
```diff
@@ -122,7 +122,7 @@ describe(Support.getTestDialectTeaser('DAO'), () => {
             // so we must create a record with the right value for always_false, then reference it in an update
             let now = dialect === 'sqlite' ? self.sequelize.fn('', self.sequelize.fn('datetime', 'now')) : self.sequelize.fn('NOW');
             if (dialect === 'mssql') {
-              now = self.sequelize.fn('', self.sequelize.fn('getdate'));
```
janmeier: Is this related to DATEONLY?

test/integration/data-types.test.js
ratanakvlun: The original test seems to be two in one. I believe the original intent of the test was to not only show that the result is a 'YYYY-MM-DD' string, but also to show how a Date object is processed when passed into a DATEONLY field. `dateOnly` represents the expected string when a Date object is passed and is then retrieved from the database client. May be better to split the test?

test/integration/instance/values.test.js
ratanakvlun: Not really. I just noticed it when running tests locally and it was among the ones that failed along with DATEONLY tests due to timezone.

kareljuricka:
I'm not sure if this is related to this pull request, but I noticed where I use DATEONLY field and init it with new Date(), timezone which I set in timezone property during creating of Sequelize instance is not used. But when I use DATE field and save it, datetime is correct in database.

In example when I'm +1 UTC and time it's 00:30
```
var date = new Date();
someModel.dateonly = new Date(); => 2017-12-19
someModel.date = new Date(); => 2017-12-20 00:30
```

Date store day to 20th but dateonly to 19th.

ratanakvlun:
From what I can tell, the timezone option does not apply to DATEONLY, though it probably should...

sushantdhiman:
> From what I can tell, the timezone option does not apply to DATEONLY, though it probably should...

Yeah, I think we should apply timezone on incoming date object (possibly if its a Date object not `YYYY-MM-DD` string) and save the timezone applied value to database. While retrieving it we can just take incoming (string or date) and show as string

PR #8793, opened 2017-12-16
fix(mysql/json): properly quote JSON path to be valid ECMAScript name

### LINKED ISSUE #8718: Invalid JSON path expression on keys with hyphens/dashes
## What are you doing?
This find operation works properly if there is not a hyphen in the key name.

```
var User = sequelize.define('user', {
  type: DataTypes.JSON,
  allowNull: true
});

return sequelize.sync({
  force: true,
  logging: console.log
})
  .then(() => {
    return User.create({
      data: {
        "bad-key": ['a', 'b']
      }
    })
  })
  .then(() => {
    return User.findAll({
      where : {
        data: {
          "bad-key": {[Op.ne]: null}
        }
      }
    });
  })
  .then(console.log)
  .finally(() => sequelize.close());
```

Sequelize would return the user data with `bad-key` as a valid key.

I get `SequelizeDatabaseError: Invalid JSON path expression.`

Query: `... WHERE (`table1`.`user`->>'$.bad-key') != 'null';`

MySQL 5.7.19
Sequelize 4.23.1

sushantdhiman:
### Pull Request check-list

### Description of change

Fixes #8718 , JSON sub path if contain hyphen / dashes it will break find query. We need to quote paths properly to be a valid ECMAScript identifier

https://bugs.mysql.com/bug.php?id=81896

### DIFF
--- a/test/integration/model/json.test.js
+++ b/test/integration/model/json.test.js
@@ -423,6 +423,60 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           });
         });
 
+        it('should be possible to query for nested fields with hyphens/dashes, #8718', function() {
+          return Promise.join(
+            this.Event.create({
+              data: {
+                name: {
+                  first: 'Homer',
+                  last: 'Simpson'
+                },
+                status_report: {
+                  'red-indicator': {
+                    'level$$level': true
+                  }
+                },
+                employment: 'Nuclear Safety Inspector'
+              }
+            }),
+            this.Event.create({
+              data: {
+                name: {
+                  first: 'Marge',
+                  last: 'Simpson'
+                },
+                employment: null
+              }
+            })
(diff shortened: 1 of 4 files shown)


test/integration/model/json.test.js
```diff
@@ -423,6 +423,60 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+                employment: 'Nuclear Safety Inspector'
+              }
+            }),
+            this.Event.create({
+              data: {
+                name: {
+                  first: 'Marge',
+                  last: 'Simpson'
+                },
+                employment: null
```
eseliger: I guess this must be `'mother'`

sushantdhiman: Being a `Mother` is not a job :wink:

PR #8795, opened 2017-12-17
fix: remove constraint before alter foreign keys

lext-7:
### Pull Request check-list

### Description of change

Remove constraint before alter foreign keys. When sync with alter, it will throw an error `Can't write; duplicate key in table '#sql-xxx'`  because there is an existing constraint at the foreign key. 

Every time altering models, it should get all foreign key informations and remove duplicated foreign key constraints, recreate them.


robin-garnham:
Thanks lext-7... I'm hoping this gets merged soon... this is causing me pain daily...

n8stowell82:
I very much need this fix as I am currently blocked from making upgrades to my models.  Thank you!

lib/dialects/mysql/query-generator.js
```diff
@@ -495,15 +523,17 @@ const QueryGenerator = {
    * @private
    */
   getForeignKeyQuery(table, columnName) {
-    let tableName = table.tableName || table;
-    if (table.schema) {
-      tableName = table.schema + '.' + tableName;
-    }
-    return 'SELECT CONSTRAINT_NAME as constraint_name'
+    const tableName = table.tableName || table;
+
```
sushantdhiman: Remove this line

lib/dialects/mysql/query-interface.js
```diff
@@ -26,7 +26,10 @@ function removeColumn(tableName, columnName, options) {
   options = options || {};
 
   return this.sequelize.query(
-    this.QueryGenerator.getForeignKeyQuery(tableName, columnName),
+    this.QueryGenerator.getForeignKeyQuery(tableName.tableName ? tableName : {
+      tableName,
+      schema: this.sequelize.config.database
```
sushantdhiman: This should be `config.schema`

lib/model.js
```diff
@@ -1122,18 +1122,46 @@ class Model {
...
+                } else if (!currentAttributes.primaryKey) {
+                  // Check foreign keys. If it's a foreign key, it should remove constraint first.
+                  const references = currentAttributes.references;
+                  if (currentAttributes.references) {
+                    const database = this.sequelize.config.database;
+                    // Find existed foreign keys
+                    _.each(foreignKeyReferences, foreignKeyReference => {
+                      const constraintName = foreignKeyReference.constraintName;
+                      if (!!constraintName
+                        && foreignKeyReference.tableCatalog === database
```
sushantdhiman: This should be checking for `schema`

lib/dialects/mysql/query-generator.js
lext-7: I think it should keep the same as https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/query-generator.js#L649

lib/dialects/mysql/query-interface.js
lext-7: https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/query-generator.js#L484
https://github.com/sequelize/sequelize/blob/master/lib/query-interface.js#L697
I think in mysql, schema can be seen as database. There is no schema option, no specific schema concept in sequelize, just use database for all.

lib/dialects/mysql/query-generator.js
sushantdhiman: I meant the empty line between two const definition

lext-7: Sorry for misunderstanding

lib/dialects/mysql/query-interface.js
sushantdhiman: Got it, can you add a comment here explaining why schema is being assigned database

PR #8810, opened 2017-12-22
feat(model): allow serializing model instances into mongodb

vkarpov15:
Hi,

Admittedly this is a strange PR, but we've seen instances of people trying to store sequelize objects in MongoDB (like https://github.com/agenda/agenda/issues/562#issuecomment-353486543). It would be handy if sequelize could implement a `toBSON()` function so the [mongodb driver could automatically serialize it correctly](https://github.com/mongodb/js-bson/pull/231).

Thanks and happy holidays!

Valeri

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -4020,6 +4020,16 @@ class Model {
       })
     );
   }
+  
+  /**
+   * For serializing this object into MongoDB
+   *
+   * @see {@link Model#get}
+   * @return {object}
+   */
+  toBSON() {
+    return this.toJSON();
+  }
 
   /**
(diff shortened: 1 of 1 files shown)


sushantdhiman:
You can add a method to model instance, add `toBSON` on prototype

```js
var Model = Sequelize.define('Model', {
   // various attributes
});

Model.prototype.toBSON = function () {
   return this.toJSON();
}
```

vkarpov15:
I'm aware of that. However, I think it would be handy to have this functionality baked in to sequelize so end users don't have to think about it. Let me know, we can always just write a plugin instead.

sushantdhiman:
Yeah, its better if this is handled at application code or as a plugin.

PR #8828, opened 2018-01-02
Migration to v4 guide updated

### LINKED ISSUE #8681: Sequelize 4 no longer set required to true anymore
I have a simple query doing nested query below
const event =  await models.Event.findOne({
        where: {
      Event_Id: eventId
    },
    include: [
      {
        model: models.Scenario,
        include: [
          {
            model: models.Phase,
            where: { External_Reference_Id: externalId }
          }
        ]
      }
    ],
    transaction
  })

Previously in V3, if the where condition failed, I will get a null object which my application expects, but now it returns an object. I have to manually add required: true after model: models.Scenario so that I can get the exact same behavior as before.

Is there an easy configuration I can config so that I don't have to pass required: true everywhere in my application

__Dialect: mysql
__Sequelize version: 4.22.12

jifeon:
Today I updated one project from 3.24.8 to 4.25.2 and noted that couple of major breaking changes are not in the migrating guide, they are

- `include` always an array
- `where` does not make `include` `required`

see changed files for verbose explanation.

fixes #8681

### DIFF
--- a/docs/upgrade-to-v4.md
+++ b/docs/upgrade-to-v4.md
@@ -115,6 +115,61 @@ Sequelize V4 is a major release and it introduces new features and breaking chan
 - Raw options for where, order and group like `where: { $raw: '..', order: [{ raw: '..' }], group: [{ raw: '..' }] }` have been removed to prevent SQL injection attacks.
 - `Sequelize.Utils` is not longer part of the public API, use it at your own risk
 - `Hooks` should return Promises now. Callbacks are deprecated.
+- `include` is always an array
+
+  Previous:
+  ```js
+  User.findAll({
+    include: {
+      model: Comment,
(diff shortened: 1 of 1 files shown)


docs/upgrade-to-v4.md
```diff
@@ -115,6 +115,61 @@ Sequelize V4 is a major release and it introduces new features and breaking chan
...
+      model: Comment,
+      as: 'comments'
+    }]
+  })
+  ```
+
+- `where` clause inside `include` does not make this `include` and all its parents `required`. You can use following `beforeFind` global hook to keep previous behaviour:
+
+  ```js
+  function whereRequiredLikeInV3(modelDescriptor) {
```
sushantdhiman: Looks good, logic wise

Issue #8837, opened 2018-01-04
When a fk is defined both with a relation and a model, the onDelete is set to CASCADE instead of SET NULL if "allowsNull" is not set to true even if it is its default value


petersg83:
When a fk is defined both with a relation and a model, the onDelete is set to CASCADE instead of SET NULL if "allowsNull" is not set to true in the model definition even if it is its default value.

We can easily see it there: https://github.com/sequelize/sequelize/blob/3397b237db2c673b6e7df84c99b773f989bcea0b/lib/associations/has-many.js#L125
The default values are not taken into account if the fk is defined in the model whereas the attributes are not defined.

I don't think it is a wanted behavior so I'll propose a fix.

Please, tell me if I misunderstood something

sushantdhiman:
https://github.com/sequelize/sequelize/pull/8838#issuecomment-382740925

PR #8838, opened 2018-01-04
Fix: Take into account the default attributes values of a foreign key…

petersg83:
… if the foreign key is declared both in a relation and in the model.

### Pull Request check-list

No, but they doesn't pass with master neither. The error is: 
```
console.log('#'.repeat(process.env.DIALECT.length + 22) + '\n# Running tests for ' + process.env.DIALECT + ' #\n' + '#'.repeat(process.env.DIALECT.length + 22))

TypeError: Cannot read property 'length' of undefined
```
Yes
No
No
No

### Description of change

Issue it solves: https://github.com/sequelize/sequelize/issues/8837

While creating the foreign keys, if the foreign key is defined in a model and a relation, it will use the default attributes values if the attributes are not defined. 

It's the first time I see the source code of sequelize so maybe I missed some code to fix too.

### DIFF
--- a/lib/associations/belongs-to.js
+++ b/lib/associations/belongs-to.js
@@ -84,7 +84,7 @@ class BelongsTo extends Association {
     });
 
     if (this.options.constraints !== false) {
-      const source = this.source.rawAttributes[this.foreignKey] || newAttributes[this.foreignKey];
+      const source = { ...newAttributes[this.foreignKey], ...(this.source.rawAttributes[this.foreignKey] || {}) };
       this.options.onDelete = this.options.onDelete || (source.allowNull ? 'SET NULL' : 'NO ACTION');
       this.options.onUpdate = this.options.onUpdate || 'CASCADE';
     }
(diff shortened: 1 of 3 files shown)


lib/associations/belongs-to.js
```diff
@@ -84,7 +84,7 @@ class BelongsTo extends Association {
     });
 
     if (this.options.constraints !== false) {
-      const source = this.source.rawAttributes[this.foreignKey] || newAttributes[this.foreignKey];
+      const source = { ...newAttributes[this.foreignKey], ...(this.source.rawAttributes[this.foreignKey] || {}) };
```
sushantdhiman: No ES6, we use Node 4 as base

sushantdhiman:
This will need some tests actually for each association type we are fixing

Like this https://github.com/sequelize/sequelize/blob/11ff7764819f2812018f3a7b8fa16ba1851e8c4d/test/integration/associations/belongs-to.test.js#L410

petersg83:
Hey, I don't really have the time to dig in it. I let anyone who wants to to make a better PR

sushantdhiman:
This problem can be solved by setting `onDelete` / `onUpdate` options from `foreignKey` which are actual source of truth.

PR #8839, opened 2018-01-04
MySQL doesn't allow null values for some data types yet sequelize uses them if provided. This stops that

mccabemj:
### Pull Request check-list

### Description of change

MySQL doesn't allow null values for some data types yet sequelize uses them if provided. This stops that by making a check before the default values are added onto the SQL. Tests are also provided to check for these default values that were not there before (even though there were checks for TEXT and BLOB)

### DIFF
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -322,7 +322,7 @@ const QueryGenerator = {
     }
 
     // BLOB/TEXT/GEOMETRY/JSON cannot have a default value
-    if (attributeString !== 'BLOB' && attributeString !== 'TEXT' && attributeString !== 'GEOMETRY' && attributeString !== 'JSON' && attribute.type._binary !== true && Utils.defaultValueSchemable(attribute.defaultValue)) {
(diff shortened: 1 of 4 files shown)


lib/dialects/mysql/query-generator.js
```diff
@@ -320,8 +321,8 @@ const QueryGenerator = {
       template += ' auto_increment';
     }
 
-    // Blobs/texts cannot have a defaultValue
-    if (attribute.type !== 'TEXT' && attribute.type._binary !== true && Utils.defaultValueSchemable(attribute.defaultValue)) {
+    // BLOB/TEXT/GEOMETRY/JSON cannot have a default value
+    if (attributeString !== 'BLOB' && attributeString !== 'TEXT' && attributeString !== 'GEOMETRY' && attributeString !== 'JSON' && attribute.type._binary !== true && Utils.defaultValueSchemable(attribute.defaultValue)) {
```
sushantdhiman: `use _.includes`

mccabemj:
Changed to use _.includes

lib/dialects/mysql/query-generator.js
mccabemj: Changed to use _.includes and pushed

Issue #8830, opened 2018-01-03
Ability to do use table hints per query with MSSQL


mccabemj:
## What are you doing?
If I want only one target table to require a lock-less state, this is not possible. You can only set the isolation level to the entire transaction. I'd like to be able to set the NOLOCK table hint (or any other table hint) to limit that isolation level or suggest other options to the specific query.

## What do you expect to happen?
To be able to set NOLOCK on a specific select query, not the entire transaction

## What is actually happening?
This is not possible.

I'd like to take a stab at this, if someone thinks it would be beneficial.

__Dialect:__ mssql
__Dialect version:__ 6.4.2
__Database version:__ XXX
__Sequelize version:__ 4.28.6
__Tested with latest release:__ Yes

janmeier:
Sounds good to me, go for it!

mccabemj:
Upon coding this, I decided it would be better to allow other table hints, instead of just allowing NOLOCK. So I have changed the title and description of this issue and have done a pull request for the feature #8843

sushantdhiman:
https://github.com/sequelize/sequelize/commit/11ff7764819f2812018f3a7b8fa16ba1851e8c4d

Issue #7031, opened 2016-12-23
Support for SQL Server Table Hints & Locks


smassin:
Hi All,

This a feature request to support SQL Server table hints and locks. I did try to look through the issues list to see if it has already been discussed and code to see if its already available but didn't find anything. I apologize if this has already been discussed.

For now I've only put in a very hacky workaround into my own local version that essentially allows you to use strings to define a `SELECT ... FROM ... WITH(...)` statement that's appended onto the main query.

```
findAll({
  where: ...
  with: ["XLOCK", ... ]
})
```

In no way would that be a final implementation but I wanted to bring it up for discussion and see if there is some way we could add official support and what would be the best practice for doing so.

Thanks

sushantdhiman:
For now Sequelize will support singular `tableHints` https://github.com/sequelize/sequelize/commit/11ff7764819f2812018f3a7b8fa16ba1851e8c4d , which will be expanded later to support multiple hints if required.

PR #8843, opened 2018-01-05
Ability to do use table hints per query with MSSQL

mccabemj:
### Pull Request check-list

### Description of change

As per issue #8830 I have implemented the ability to do table hints with MSSQL queries. Simply adding the `tableHint` option with a string containing the type of table hint will add it to the SQL of the query. I have also added tests for a couple of potential table hints.

The array of potential table hints comes from the `table_hint_limited` section here 
https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table

Note I have not added the ability for multiple table hints, or the ability to use hints with index's. This could be added in the future.

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -815,6 +815,11 @@ const QueryGenerator = {
       mainFragment += ' AS ' + mainTableAs;
     }
 
+    const tableHints = ['NOLOCK', 'READUNCOMMITTED', 'UPDLOCK', 'REPEATABLEREAD', 'SERIALIZABLE', 'READCOMMITTED', 'TABLOCK', 'TABLOCKX', 'PAGLOCK', 'ROWLOCK', 'NOWAIT', 'READPAST', 'XLOCK', 'SNAPSHOT', 'NOEXPAND'];
+    if (options.tableHint && _.includes(tableHints, options.tableHint)) {
+      mainFragment += ` WITH (${options.tableHint})`;
+    }
+
     return mainFragment;
   },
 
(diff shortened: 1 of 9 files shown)


lib/dialects/mssql/query-generator.js
```diff
@@ -815,6 +815,11 @@ const QueryGenerator = {
       mainFragment += ' AS ' + mainTableAs;
     }
 
+    const tableHints = ['NOLOCK', 'READUNCOMMITTED', 'UPDLOCK', 'REPEATABLEREAD', 'SERIALIZABLE', 'READCOMMITTED', 'TABLOCK', 'TABLOCKX', 'PAGLOCK', 'ROWLOCK', 'NOWAIT', 'READPAST', 'XLOCK', 'SNAPSHOT', 'NOEXPAND'];
```
janmeier: Since we are hard coding the available values anyway, we might as well expose it as an object, similar to datatypes. That way users can also get editor completion

https://github.com/mccabemj/sequelize/blob/ca4d995f80d5ca22563061adb3ea51aed0852094/lib/sequelize.js#L1165
https://github.com/mccabemj/sequelize/blob/ca4d995f80d5ca22563061adb3ea51aed0852094/lib/query-types.js

mccabemj: Thanks @janmeier good suggestion. I have implemented that, let me know what you think

sushantdhiman:
> Note I have not added the ability for multiple table hints

Got it, lets start with small if there is demand for multiple hints we can always add that later

mccabemj:
@sushantdhiman yes, I was aiming to start small. In the future, a new option could be added for multiples, but this would be a lot more complex as it would have to allow for indexs on certain values. Plus, you are not allowed more than one hint from each group (Granularity hints and Isolation level hints) so that would have to be coded. 

Seeing as there is only my issue and #7031 I'd say this is good enough for now. Your call

PR #8872, opened 2018-01-10
fix(model/upsert): Made the validation flag work for upserts

aceew:
Added:
 - Check on options.validate before choosing to validate the instance in: lib/model.js

### Pull Request check-list

- [ x ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ x ] Have you added new tests to prevent regressions?
- [ x ] Did you follow the commit message conventions explained in [CONTRIBUTING.md](../CONTRIBUTING.md)?

### Description of change

`validate` flag did nothing on model upserts meaning validations would always be run. This PR resolves this by wrapping the model validations in a condition the same way done so for `Model.create`


test/integration/model/upsert.test.js
```diff
@@ -196,6 +196,29 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+
+        const options = { validate: false };
+
+        return User.sync({ force: true })
+          .then(() => User.upsert({ id: 1, email: 'notanemail' }, options))
+          .then(created => {
+            if (dialect === 'sqlite') {
+              expect(created).to.be.undefined;
+            } else {
+              expect(created).to.be.okay;
```
sushantdhiman: hmm, `okay` is supported too :D

PR #8876, opened 2018-01-11
feat: Add query timeout

Zhigalov:
### Pull Request check-list

- [x] Does `npm run test-postgres` pass with this change (including linting)?
- [x] Does the description below contain a description of the issue you are solving?

### Description of change

Added timeout for query processing. If query timed out in `options.queryTimeout` ms, then it would be finished with error `Query timed out`.

It allows to solve the issue of losing connections to database. If request was created while connection was lost, then it will never be completed (resolved). Connection is allocating from pool and never released. It leads to state when all connections in pool are in `allocated` status. All new queries are finished with error `ResourceRequest timed out`.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -97,6 +97,7 @@ class Sequelize {
    * @param {Integer}  [options.retry.max] How many times a failing query is automatically retried.  Set to 0 to disable retrying on SQL_BUSY error.
    * @param {Boolean}  [options.typeValidation=false] Run built in type validators on insert and update, e.g. validate that arguments passed to integer fields are integer-like.
    * @param {Object|Boolean} [options.operatorsAliases=true] String based operator alias, default value is true which will enable all operators alias. Pass object to limit set of aliased operators or false to disable completely.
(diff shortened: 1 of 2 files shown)


lib/sequelize.js
```diff
@@ -540,8 +541,19 @@ class Sequelize {
     }).then(connection => {
       const query = new this.dialect.Query(connection, this, options);
       const retryOptions = _.extend(this.options.retry, options.retry || {});
+      const queryPromises = [retry(() => query.run(sql, bindParameters), retryOptions)];
 
-      return retry(() => query.run(sql, bindParameters), retryOptions)
+      if (this.options.queryTimeout) {
```
sushantdhiman: This is already possible with "undocumented" `retry.timeout` https://github.com/mickhansen/retry-as-promised

Actually just setting `retry` config to something like this should be able to replicate what this PR is trying to achieve
```js
{
  max: 1,
  timeout: 2000,
  name: 'Query'
}
```

Zhigalov: Really it works! Thanks for the advice

xianxingg: This should definitely be documented!

PR #8896, opened 2018-01-14
fix(datatypes): Sequelize.NOW set milliseconds to zero

### LINKED ISSUE #8460: timestamp with default NOW will be created with no milliseconds
## What are you doing?

A timestamp, say `createdAt` property that has the `DataTypes.NOW` in the definition will be inserted with no milliseconds.

```js
sequelize.define('foo', {  
  time_created: {
        type: DataTypes.DATE,
        allowNull: false,
        defaultValue: DataTypes.NOW,
  }, {
      timestamps: true,
      createdAt: 'time_created',
      updatedAt: false,
      deletedAt: false,
  }
});

// Later

Foo.create({}) // Creates an insert query with a timestamp missing milliseconds
```

```sql
INSERT INTO "foo" ("time_created") VALUES ('2017-10-10 20:59:44.000 +00:00') RETURNING *;                                                                                   
```

Expect to see milliseconds

seeing 00 in place of millis.

Tested on postgres but probably applies to others.

Seemingly related to #4539

sushantdhiman:
### Pull Request check-list

### Description of change

Fix #8460
Close #8704

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -989,7 +989,7 @@ class Model {
       }
 
       if (definition.hasOwnProperty('defaultValue')) {
-        this._defaultValues[name] = _.partial(Utils.toDefaultValue, definition.defaultValue);
+        this._defaultValues[name] = _.partial(Utils.toDefaultValue, definition.defaultValue, this.sequelize.options.dialect);
       }
 
       if (definition.hasOwnProperty('unique') && definition.unique) {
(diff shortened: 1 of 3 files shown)


mickhansen:
Only for certain dialects i assume @sushantdhiman? In postgres you'd definitely want milliseconds

sushantdhiman:
Now all dialect supports milliseconds

https://github.com/sequelize/sequelize/blob/781d528856dad13cd0beae4267baaa869003220e/lib/utils.js#L394

Issue #3354, opened 2015-03-16
Sequelize Upsert() return value


hfar:
Hello,

 I need to get the id for the inserted/updated record when using .upsert() in sequelize.

right now .upsert() returns a boolean indicating whether the row was created or updated.

thanks for your help.

```
return db.VenueAddress.upsert({
            ...
        }).then(function(test){
            //test returned here as true or false how can i get the inserted id here so i can insert data in other tables using this new id?
        });
```

janmeier:
This is only supported by postgres, so to keep the API consistent across dialects this is not possible

FredKSchott:
:+1: for adding this feature, there's plenty of precedence for dialect-specific features. This seems exactly like `[options.returning=false]` on [Model.update()](http://docs.sequelizejs.com/en/latest/api/model/#updatevalues-options-promisearrayaffectedcount-affectedrows)

GabeIsman:
I also just ran into this problem. I wanted to upsert a list of records, and then do something with them (associate them with another model). That seems to be impossible the way the API is defined now. Is there a recommended workflow for this use case? I'm trying to avoid just recreating all associated records on update, which seems to be the path of least resistance given the API design (particularly that the records must already exist in the DB before they can be associated).

hfar:
+1

coderholic:
+1 for adding options.returning on upsert.

rfink:
+1, would be tremendously useful

FredKSchott:
So I did some digging into this, and it won't be easy. I wanted us to be able to take advantage of `RETURNING *` in postgres, but the problem is that the return value of insert & update are both already being used to detect which one occurred. ([link to code](https://github.com/sequelize/sequelize/blob/d3ac832c14b1fa8e0c0d17139848c8cee0103cae/lib/dialects/postgres/query-generator.js#L319))

At this point I think the best way forward is to append a simple `SELECT *` query to the end of the upsert transaction. We'll need to append the upsert response (1 or 2) in the result as well, which is where I got stuck (got `Cannot read property 'sequelize_upsert' of undefined` error when appending `; SELECT *, pg_temp.sequelize_upsert() FROM users WHERE ...`) 

I don't have any more time to work on this for a while, but if anyone wants to take over where I left off feel free!

kylehotchkiss:
+1; this would be useful for me as well.

I thought there was an update() function with { upsert: true } as an option but it's totally possible i'm getting confused with Mongoose

[...]

PR #8924, opened 2018-01-20
feat(upsert): return upserted record with options.returning=true

sushantdhiman:
### Pull Request check-list

### Description of change

Close https://github.com/sequelize/sequelize/issues/3354

Sequelize still doesnot support `ON CONFLICT`, but with using `RETURNING` its now possible for return upserted record.

```js
// current behaviour
const created = await Model.upsert(....values....);

// with options.returning = false
const created = await Model.upsert(....values...., { returning: false });

// with options.returning = true , new feature
const [record, created] = await Model.upsert(....values...., { returning: true });
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2215,22 +2215,25 @@ class Model {
    * @param  {Boolean}      [options.validate=true] Run validations before the row is inserted
    * @param  {Array}        [options.fields=Object.keys(this.attributes)] The fields to insert / update. Defaults to all changed fields
    * @param  {Boolean}      [options.hooks=true]  Run before / after upsert hooks?
+   * @param  {Boolean}      [options.returning=false] Append RETURNING * to get back auto generated values (Postgres only)
    * @param  {Transaction}  [options.transaction] Transaction to run query under
    * @param  {Function}     [options.logging=false] A function that gets executed while running the query to log the sql.
    * @param  {Boolean}      [options.benchmark=false] Pass query execution time in milliseconds as second argument to logging function (options.logging).
    * @param  {String}       [options.searchPath=DEFAULT] An optional parameter to specify the schema search_path (Postgres only)
    *
-   * @return {Promise<created>} Returns a boolean indicating whether the row was created or updated.
+   * @return {Promise<created>} Returns a boolean indicating whether the row was created or updated. For Postgres/MSSQL with (options.returning=true), it returns record and created boolean with signature `<Model, created>`.
    */
   static upsert(values, options) {
-    options = _.extend({
-      hooks: true
+    options = Object.assign({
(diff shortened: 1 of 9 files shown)


lib/model.js
```diff
@@ -2215,22 +2215,25 @@ class Model {
...
    * @param  {Array}        [options.fields=Object.keys(this.attributes)] The fields to insert / update. Defaults to all changed fields
    * @param  {Boolean}      [options.hooks=true]  Run before / after upsert hooks?
+   * @param  {Boolean}      [options.returning=false] Append RETURNING * to get back auto generated values (Postgres only)
    * @param  {Transaction}  [options.transaction] Transaction to run query under
    * @param  {Function}     [options.logging=false] A function that gets executed while running the query to log the sql.
    * @param  {Boolean}      [options.benchmark=false] Pass query execution time in milliseconds as second argument to logging function (options.logging).
    * @param  {String}       [options.searchPath=DEFAULT] An optional parameter to specify the schema search_path (Postgres only)
    *
-   * @return {Promise<created>} Returns a boolean indicating whether the row was created or updated.
+   * @return {Promise<created>} Returns a boolean indicating whether the row was created or updated. For Postgres with (options.returning=true), it returns record and created boolean with signature `<Model, created>`.
```
sushantdhiman: In v5, we can set this `Promise<record, created>` by setting `returning=true` by default. One caveat is `null` is now first argument for other dialects. May be should set this as ``Promise<created, record>``, but that doesn't match with other method's API. Any ideas?

sushantdhiman: I have added support for MSSQL as well, so now may be current order make sense.

sushantdhiman:
In issue https://github.com/sequelize/sequelize/issues/3354 there is discussion about MySQL support, but I find that `LAST_INSERTED_ID` is still limited. It wont work for any non incrementing primary keys like STRING, UUID etc. So I haven't added any support for MySQL.

Currently we support MSSQL and Postgres for this feature

aprilmintacpineda:
`options.returning` seems to not work on tables with multiple primary keys.

- sequelize-cli@4.0.0
- sequelize@4.33.4

repo: upcoming

abdabsi:
> In issue #3354 there is discussion about MySQL support, but I find that `LAST_INSERTED_ID` is still limited. It wont work for any non incrementing primary keys like STRING, UUID etc. So I haven't added any support for MySQL.
> 
> Currently we support MSSQL and Postgres for this feature

If they're using non-incrementing primary keys, they most probably won't need LAST_INSERTED_ID. Those who use auto-incremented primary keys (absolute majority) on MySql (#1 rdbms) have no use of the current upsert method. Please fix it! :)

Issue #3528, opened 2015-04-13
When selecting from users table in postgres, select statement generated tries to use reserved word 'user'


alexduan:
Driver: postgres
Created a user model using: sequelize model:create --name User --attributes email:string

When selecting from the user model with something as simple as:

models.User.findAll({}).then(function(err, users) {
    console.log(users);
});

I get the following stacktrace: 
Executing (default): SELECT id, email, createdAt, updatedAt, deletedAt FROM Users AS User WHERE User.deletedAt IS NULL;
Possibly unhandled SequelizeDatabaseError: syntax error at or near "User"

janmeier:
Do you have quoteIdentifiers set to false? Otherwise User should automatically be quoted.

Which seqeulize version are  you using?

alexduan:
Yes, I have set it to false so that table names are not created with quotes
(which is kind of weird on Postgres...).

On Tue, Apr 14, 2015 at 1:43 AM, Jan Aagaard Meier <notifications@github.com

janmeier:
Well in that case there is really nothing to do - you have told sequelize that you don't want quotes, so you're not getting any :)

`quoteIdentifiers`  should really only be used in legacy situations where the table was created without quotes. When everything is handled by sequelize, quotes is your friend ;)

mickhansen:
@alexduan includes don't work with `quoteIdentifiers: false` so i'd reenable it.

alexduan:
Ah, ok - that makes sense. I'm used to inspecting records via psql and was
weirded out by the quotes. Will re-enable and suck it up. Thanks!

On Tue, Apr 14, 2015 at 9:20 AM, Jan Aagaard Meier <notifications@github.com

PR #8935, opened 2018-01-22
feat(PostgreSQL dialect): always quote identifier if it is a reserved word

jmarton:
### Pull Request check-list

- Have you added new tests to prevent regressions? (*edit: not applicable*)
- Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? (*edit: not applicable*)

### Description of change

Having reserved words generated in the PostgreSQL queries came up a few times, e.g. in #3528, and in these situations, it was the only possible solution to re-enable quoting.

After this change, reserved words are always quoted, even if `quoteIdentifiers` is set to `false`.

This makes it possible to have case-insensitive identifiers in PostgreSQL when certain queries had table aliases generated, which turn out to be PostgreSQL reserved words.

e.g. for the model `Users`, sometimes, `User` is the generated table alias, which must be quoted to avoid naming problems.

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -12,6 +12,10 @@ const QueryGenerator = {
   options: {},
   dialect: 'postgres',
 
+  // list of reserved words in PostgreSQL 10
+  // source: https://www.postgresql.org/docs/10/static/sql-keywords-appendix.html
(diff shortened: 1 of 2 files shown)


jmarton:
@mickhansen may I ask you for some advice on how to move forward with this?

For the missing checks:
 - I didn't add new tests because some old tests requirred modification and they cover this enhancement
 - no API was changed, so I didn't include any docs

jmarton:
@sushantdhiman may I ask you for some advice on how to move forward with this?

sushantdhiman:
Seems like a good idea and actually this will allow users to use sequelize properly when identifiers are not quoted. But this needs some work, I will tackle that in my attempt to refactor query generator https://github.com/sequelize/sequelize/blob/refactor/lib/dialects/abstract/query-generator/helpers/quote.js#L46

sushantdhiman:
This fix was included to v5 https://github.com/sequelize/sequelize/pull/9236, will be released in next beta release

PR #8938, opened 2018-01-23
fix(<model>): fix findOrBuild bug

clairecliu:
Bug Fix: Modified model.js for findOrBuild, so instance is built with values and options.
address #[8919]

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2056,7 +2056,7 @@ class Model {
           values = _.defaults(values, options.where);
         }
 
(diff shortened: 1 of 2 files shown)


sushantdhiman:
This needs a integration test, https://github.com/sequelize/sequelize/tree/master/test/integration/model here create a new file called `findAndBuild.test.js`

clairecliu:
I'll work on including the findAndBuild test.

Issue #8043, opened 2017-08-01
Cannot read property '0' of undefined with pg 7


tapz:
## What you are doing?

I just upgraded to Sequelize 4.4.2 and node-postgres (pg) version 7.0.2

Then, when I call sync() to connecto the db and create tables. I get an error

It happens, for example, when I run this command:

```js
models.sequelize.sync({force: true}).
```

## What do you expect to happen?

Sequelize should connect to Postgres and the create tables.

## What is actually happening?

```js
TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined
    at Promise.then.result (C:\test\node_modules\sequelize\lib\dialects\postgres\connection-manager.js:157:31)
    at tryCatcher (C:\test\node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (C:\test\node_modules\bluebird\js\release\promise.js:512:31)
    at Promise._settlePromise (C:\test\node_modules\bluebird\js\release\promise.js:569:18)
    at Promise._settlePromise0 (C:\test\node_modules\bluebird\js\release\promise.js:614:10)
    at Promise._settlePromises (C:\test\node_modules\bluebird\js\release\promise.js:693:18)
    at Async._drainQueue (C:\test\node_modules\bluebird\js\release\async.js:133:16)
    at Async._drainQueues (C:\test\node_modules\bluebird\js\release\async.js:143:10)
    at Immediate.Async.drainQueues (C:\test\node_modules\bluebird\js\release\async.js:17:14)
    at runCallback (timers.js:651:20)
    at tryOnImmediate (timers.js:624:5)
    at processImmediate [as _immediateCallback] (timers.js:596:5)
```

__Dialect: postgres
__Database version: 9.6.2 (pg 7.0.2)
__Sequelize version: 4.4.2

chaserelock:
I get this as well, I can solve it by checking that the rows array exists in query.js:112  
  
```javascript
        // change to (rows && rows[0] && ...)
        if (rows[0] && rows[0].sequelize_caught_exception !== undefined) {
          if (rows[0].sequelize_caught_exception !== null) {
            throw this.formatError({
              code: '23505',
              detail: rows[0].sequelize_caught_exception
            });
          } else {
            for (const row of rows) {
              delete row.sequelize_caught_exception;
            }
          }
        }```

wavded:
Also confirm this is an issue for me.

fnando:
[...]

PR #8947, opened 2018-01-25
Fix pg7 - Cannot read property '0' 

Vincz:
### Pull Request check-list

### Description of change

Handle pg7 multi-queries format. See #8043.
The idea is to check for an array and re-concatenate the rows as it was in pg 6.
See the following:

There is a problem as soon as their is more than one query executed at a time.
Given the following code:

```
const { Client } = require('pg')
const query = "SELECT 'hello'; SELECT 'world';";
const client = new Client()

const f = async () => {
	await client.connect()
	const res = await client.query(query);
	console.log(res);
	await client.end()
};

f();
```

It outputs with pg 6:

```
➜  /tmp node pg6/index.js
Result {
  command: 'SELECT',
  rowCount: 1,
  oid: NaN,
  rows:
   [ anonymous { '?column?': 'hello' },
     anonymous { '?column?': 'world' } ],
  fields:
   [ Field {
       name: '?column?',
       tableID: 0,
       columnID: 0,
       dataTypeID: 705,
       dataTypeSize: -2,
       dataTypeModifier: -1,
       format: 'text' } ],
  _parsers: [ [Function: noParse] ],
  RowCtor: [Function: anonymous],
  rowAsArray: false,
  _getTypeParser: [Function: bound ] }
```

And it outputs with pg 7 :

```
➜  /tmp node pg7/index.js
[ Result {
    command: 'SELECT',
    rowCount: 1,
    oid: null,
    rows: [ [anonymous] ],
    fields: [ [Field] ],
    _parsers: [ [Function: noParse] ],
    RowCtor: [Function: anonymous],
    rowAsArray: false,
    _getTypeParser: [Function: bound ] },
  Result {
    command: 'SELECT',
    rowCount: 1,
    oid: null,
    rows: [ [anonymous] ],
    fields: [ [Field] ],
    _parsers: [ [Function: noParse] ],
    RowCtor: [Function: anonymous],
    rowAsArray: false } ]
➜  /tmp
```

This PR, check for an array as query result and turn it back into a single result (by concatenating rows). The queryResult can be use as metadata for raw queries but the rowCount was always 1 for multi-queries anyway so it doesn't really matter.

### DIFF
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -96,7 +96,7 @@ class Query extends AbstractQuery {
       })
       .then(queryResult => {
         const rows = Array.isArray(queryResult) ? queryResult.reduce((rows, r) => rows.concat(r.rows), []) : queryResult.rows;
-        const rowCount = Array.isArray(queryResult) ? queryResult[0].rowCount : queryResult.rowCount;
+        const rowCount = Array.isArray(queryResult) ? queryResult.reduce((rowCount, r) => rowCount + r.rowCount, 0) : queryResult.rowCount;
         const isTableNameQuery = sql.indexOf('SELECT table_name FROM information_schema.tables') === 0;
         const isRelNameQuery = sql.indexOf('SELECT relname FROM pg_class WHERE oid IN') === 0;
 
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -95,8 +95,8 @@ class Query extends AbstractQuery {
         return queryResult;
       })
       .then(queryResult => {
-        const rows = queryResult.rows;
-        const rowCount = queryResult.rowCount;
+        const rows = Array.isArray(queryResult) ? queryResult.reduce((rows, r) => rows.concat(r.rows), []) : queryResult.rows;
+        const rowCount = Array.isArray(queryResult) ? queryResult[0].rowCount : queryResult.rowCount;
         const isTableNameQuery = sql.indexOf('SELECT table_name FROM information_schema.tables') === 0;
         const isRelNameQuery = sql.indexOf('SELECT relname FROM pg_class WHERE oid IN') === 0;
 


lib/dialects/postgres/query.js
```diff
@@ -95,8 +95,8 @@ class Query extends AbstractQuery {
         return queryResult;
       })
       .then(queryResult => {
-        const rows = queryResult.rows;
-        const rowCount = queryResult.rowCount;
+        const rows = Array.isArray(queryResult) ? queryResult.reduce((rows, r) => rows.concat(r.rows), []) : queryResult.rows;
+        const rowCount = Array.isArray(queryResult) ? queryResult[0].rowCount : queryResult.rowCount;
```
eseliger: shouldn't this be `queryResult.map(row => row.rowCount).reduce((prev, next) => prev + next)` or so? 
In your example, both results had rowCount 1, so in sum they would be 2, which seems to be correct.

lib/dialects/postgres/query.js
```diff
@@ -95,8 +95,8 @@ class Query extends AbstractQuery {
         return queryResult;
```
sushantdhiman: A few changes

1) Please run tests against v7
2) Remove note about v7 incompatibility from https://github.com/sequelize/sequelize/blob/master/README.md#installation and https://github.com/sequelize/sequelize/blob/master/docs/usage.md#postgresql

lib/dialects/postgres/query.js
```diff
@@ -95,8 +95,8 @@ class Query extends AbstractQuery {
...
       })
       .then(queryResult => {
-        const rows = queryResult.rows;
-        const rowCount = queryResult.rowCount;
+        const rows = Array.isArray(queryResult) ? queryResult.reduce((rows, r) => rows.concat(r.rows), []) : queryResult.rows;
+        const rowCount = Array.isArray(queryResult) ? queryResult.reduce((rowCount, r) => rowCount + r.rowCount, 0) : queryResult.rowCount;
```
sushantdhiman: rowCount can be `NaN`, atleast in v6. Please handle that case (you can take hints from https://github.com/sequelize/sequelize/pull/8029/files)

PR #8022, opened 2017-07-27
fix(connection): fixing mssql reconnect bug

codyseibert:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [X] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you follow the commit message conventions explained in CONTRIBUTING.md?

### Description of change
Original Issue:
https://github.com/sequelize/sequelize/issues/8014

There is currently a bug where if you restart a mssql database,
sequelize will lock the pool resources.  Once those are locked, all
future requests to .acquire will throw a ResourceRequest timed out
error.  This commit wraps all the connect logic in a retry to prevent an
inifite loop of trying to connect, and also watchins for an 'end' event
in the mssql adapter to properly reject the promise if 'end' is sent
before 'connect' is fired.

PR #8961, opened 2018-01-29
fix(sequelize/query): closes #2113, closes #8022, closes #8014

### LINKED ISSUE #2113: Preferred way to handle reconnect
In case the db connection is lost, we need to reconnect to the database. I couldn't find a preferred way to handle this issue. Do you have a best practice? The issue can be solved on various ways:
- handle in sequelize
- handle in db module like pg
- handle this issue outside of nodejs with connection handler

Thanks in advance

### LINKED ISSUE #8014: ResourceRequest timed out - connection not reestablished after db restarted
## What you are doing?
We're running into an issue where Sequelize will not reconnect to the database after the database is stopped and started again.  If you want, you can try hosting a local mssql database using this tutorial and running the following code, and following the steps I provide below:  

https://medium.com/@reverentgeek/sql-server-running-on-a-mac-3efafda48861

```js
const options = {
  host: 'localhost',
  dialect: 'mssql',
  pool: {
    max: 1,
    min: 0,
    idle: 5000,
    acquire: 5000
  },
  dialectOptions: {
    requestTimeout: 5000
  },
  isolationLevel: 'READ UNCOMMITTED'
};

const sequelize = new Sequelize('testing', 'sa', 'P@55w0rd', options);

setInterval(() => {
  sequelize
    .query("SELECT 'connection is successful'", {
      type: Sequelize.QueryTypes.SELECT
    })
    .then((results) => {
      console.log(results);
    })
    .catch((err) => {
      console.log(err);
    })
}, 5000);
```

Steps after code is running and mssql container is running:

1. stop the container
2. wait until an interval fires and throws an error
3. start the container
4. notice that all future queries result in timeout errors

I'm pretty sure this should reject, but I'm not too sure: https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/connection-manager.js#L98

Sequelize to reconnect and have queries work as normal.

Sequelize fails to reconnect the created pool resource.

__Dialect:__ mssql, but probably all
__Database version:__ XXX
__Sequelize version:__ 4.4.2

contra:
### Pull Request check-list

### Description of change

- Wraps the entire query process in retry, instead of just the query.
- Failed queries will yield the connection back to the pool, and request another one on the next retry
  - This allows the retry functionality to work with the pool reconnect functionality 💯 

There are a few failing tests I think should be removed, but I'm opening this for discussion. The test that ensures the connection release is called only once seems wrong to me, if you're going to be retrying for 60 seconds (or longer!) shouldn't you yield the connection between tries? Otherwise the pool will get completely hosed with connections that are doing nothing but waiting for the next retry. That test was added to close a ticket that would also be solved with this change.

Anyways, let me know what you think RE tests. We're using this already on test servers and it seems to work great - I can restart the db or shut it off for a while and when I turn it back on the queries resolve just fine. This makes zero downtime multi-az failovers work 100%.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "lodash": "^4.17.1",
     "moment": "^2.13.0",
     "moment-timezone": "^0.5.4",
-    "pg-types": "^1.13.0",
     "retry-as-promised": "^2.3.1",
     "semver": "^5.5.0",
     "terraformer-wkt-parser": "^1.1.2",


sushantdhiman:
Thanks for this PR @contra , I will take care of cleaning up tests

contra:
@sushantdhiman Thanks! Let me know if you need anything else.

sushantdhiman:
Somehow cant push to contra/master forcefully, will reopen another PR

PR #8991, opened 2018-02-03
fix(sequelize/query): retry entire query

### LINKED ISSUE #2113: Preferred way to handle reconnect

### LINKED ISSUE #8014: ResourceRequest timed out - connection not reestablished after db restarted

sushantdhiman:
### Pull Request check-list

### Description of change

> 1. Wraps the entire query process in retry, instead of just the query.
> 2. Failed queries will yield the connection back to the pool, and request another one on the next retry
> This allows the retry functionality to work with the pool reconnect functionality :100: 

Continued from https://github.com/sequelize/sequelize/pull/8961

Closes #2113 
Closes #8022 
Closes #8014


test/unit/query.test.js
```diff
@@ -1,37 +0,0 @@
-'use strict';
```
sushantdhiman: As noted in original PR this test was added to ensure connection is only released once and there is no `Resource is not a part of pool` error.

Now each retry fetch new connection so this test is not required. Also that error was already handled [here](https://github.com/sequelize/sequelize/blob/d65a95c953ccd0595d5b406cd198ac149e14d955/lib/dialects/abstract/connection-manager.js#L287)

lib/sequelize.js
```diff
@@ -517,41 +525,47 @@ class Sequelize {
...
 
       if (options.transaction && options.transaction.finished) {
-        const error = new Error(options.transaction.finished+' has been called on this transaction('+options.transaction.id+'), you can no longer use it. (The rejected query is attached as the \'sql\' property of this error)');
+        const error = new Error(`${options.transaction.finished} has been called on this transaction(${options.transaction.id}), you can no longer use it. (The rejected query is attached as the \'sql\' property of this error)`);
         error.sql = sql;
         return Promise.reject(error);
       }
 
-      if (this.test._trackRunningQueries) {
+      if (isFirstTry && this.test._trackRunningQueries) {
```
sushantdhiman: Not related to this PR but a possible fix for randomly failing tests due to tracking queries //cc @janmeier

contra:
found one broken case here - retry-as-promised seems to start the timeout counter before any retries have happened, so if you try to run a really long query (lets say 120s) and you set the retry timeout for 60s it will fail with a timeout, even though 0 retries happened. seems like something that needs to be fixed in retry-as-promised.

sushantdhiman:
> found one broken case here - retry-as-promised seems to start the timeout counter before any retries have happened, so if you try to run a really long query (lets say 120s) and you set the retry timeout for 60s it will fail with a timeout, even though 0 retries happened. seems like something that needs to be fixed in retry-as-promised.

Current default timeout is `undefined` which means no timeout. But with timeout if there is such bug it should be fixed in https://github.com/mickhansen/retry-as-promised

contra:
@sushantdhiman Yep, understood where the fault lies - I think this can be merged without the fix, just wanted to mention it. I've opened the ticket there and will probably send a PR if I have time to https://github.com/mickhansen/retry-as-promised/issues/14.

lib/sequelize.js
janmeier: Could very well be!

eseliger: that would be awesome!

cladjules:
Great that is resolved :)
Do you know what version it's going to be released?

Thanks :)

reiz:
We need this fix ASAP :-) We are affected by this and looking forward to this fix!

eseliger:
https://github.com/sequelize/sequelize/releases/tag/v4.32.7 enjoy!

PR #8992, opened 2018-02-04
Add option to `sync` similar to `alter` but without `removeColumn`

timhuff:
This is a 2-line change that I thought would be useful. The idea is that it would allow the dev to write something like `User.sync({alterKeepColumns: true})` which does the same thing as `{alter: true}` only it doesn't remove old columns. This way you don't have to risk data deletion if you were to typo or something. I'm a bit crunched for time at the moment so I didn't test this or anything - I wouldn't say it's ready to be merged but was hoping to get a conversation going on the possibility of adding this to the API.

### Pull Request check-list

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1121,7 +1121,7 @@ class Model {
     })
       .then(() => this.QueryInterface.createTable(this.getTableName(options), attributes, options, this))
       .then(() => {
-        if (options.alter) {
+        if (options.alter || options.alterKeepColumns) {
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Being part of `sync` process `alter` goal is to modify current table in response to Model definition changes and come up with most up to date representation. Which means it wont keep any old stale columns.

Sorry I cant accept this patch as explained above, it fundamentally change meaning of `sync`

PR #8997, opened 2018-02-05
Add UniqueIdentifier type support for MSSQL

funkydev:
Hi,

With my team which use Sequelize we're trying to improve database performance.
Microsoft database does have it own type for UUIDs - UniqueIdentifier.

SQL Server does have special alghoritms for indexing UUIDs and it store UUIDs using less bytes.
It has an impact into time of SELECT operations.

### Pull Request check-list


EGSatelix:
Hi,
I need to use uniqueidentifier for MSSQL, it's works with type: 'uniqueidentifier' but we can't use trigger or function create/update... because "TypeError: attribute.type.toSql is not a function"

PR #9006, opened 2018-02-06
fix(escape): Remove null character in postgres

### LINKED ISSUE #6311: Security: 'unterminated quoted string' updating model
We have a case from fuzzy scanning that results in the following error:

```
unterminated quoted string at or near "'"
LINE 1: UPDATE "production"."questions" SET "text"='
```

The error is caused by a model instance being updated, like this:

```
const question = await sequelize.models.Entity.findById(questionId);
await question.update({
  text: req.body.text
});
```

The `req.body.text` is validated to be a string and within the right length. We suspect this is caused by some special character being inside req.body.text, that fools the escaping logic in Sequelize. If https://github.com/sequelize/sequelize/issues/1608 is implemented, this would be a non-issue, as you will skip any messy escaping, that is _extremely_ hard to do properly, just ask the PHP guys.

Most likely this makes Sequelize vulnerable to SQL injection for any uses that accept text from the user (which I would guess is pretty close to all uses of Sequelize).

**Dialect:** postgres
**Database version:** 9.5
**Sequelize version:** 3.23.4

holm:
### Pull Request check-list

### Description of change

This escapes the null character for Postgres, which it does not allow. 

Closes #6311

### DIFF
--- a/lib/sql-string.js
+++ b/lib/sql-string.js
@@ -55,6 +55,11 @@ function escape(val, timeZone, dialect, format) {
     // http://www.postgresql.org/docs/8.2/static/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS
     // http://stackoverflow.com/q/603572/130598
     val = val.replace(/'/g, "''");
+
+    if (dialect === 'postgres') {
+      // null character is not allowed in Postgres
+      val = val.replace(/\0/g, '\\0');
+    }
   } else {
     val = val.replace(/[\0\n\r\b\t\\\'\"\x1a]/g, s => {
       switch (s) {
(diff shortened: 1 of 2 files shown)


holm:
I don't think this requires any documentation changes. 

I am happy to take pointers on where to best add a test case for this and how to structure it. Perhaps pointing to an existing similar test will help.

As for the changelog, I can't find any. Is that not used any more?

holm:
The test failure seems unrelated to this PR.

holm:
I added a testcase and also decided that removing the null characters is probably the better option.

holm:
Yeah, it seems I missed that. Maybe that was added between v3 and v4, where I originally did my patch. I will unify the way this is handled.

holm:
I changed things a little bit to match the existing escaping and fixed the test cases.

janmeier:
:+1:

PR #9007, opened 2018-02-06
feat(paranoid): Revert paranoid to work like in v3

### LINKED ISSUE #8496: Comparing timestamps can lead to data inconsistencies in soft deleted records
I've [written about this before](https://github.com/sequelize/sequelize/pull/8485#issuecomment-337081681) a [few times](https://github.com/sequelize/sequelize/pull/7290#issuecomment-327872193) but I will summarize my concerns again here. Apologies for eschewing the issue template and for the length of this post, but I wanted to use this as a place for summarizing all the issues around Sequelize's practice of comparing timestamps for soft deletes.

## Use of client time instead of DB time

Before #8485, Sequelize would soft delete records using client's time, but query for them using database's time (by "client's time" I mean client of the database, e.g. an app server, ad hoc script, etc). This lead to inconsistencies that have been well documented; namely if client and DB clocks aren't in sync, you can potentially see a soft deleted record.  

#8485 normalized this so that both delete and query operations use the client's time. This is good in that you no longer have to worry about time zone differences or clock drift between the DB and the client. 

But the safer choice would arguably be to use the DB's time rather than the client's. This is because a typical app will have a number of clients (load balanced app instances, ad hoc scripts running on dev machines, etc), all of which need to keep their clocks in sync with each other, compared to only a handful of DB instances (and often only 1).

This problem still exists when using DB time if you have multiple database instances in a cluster, but typically the number of DB instances will be far less than the number of clients. Additionally, DB instances tend to be pretty homogeneous and under tight control, so keeping their clocks in sync would tend to be much easier than keeping client clocks in sync. 

That being said, I still think it's still dangerous to rely on DB clocks being in sync, especially when you have multiple DB instances and non-monotonic time sources.

## Non-monotonic time source

Javascript's `Date` object is not monotonic, i.e. in the presence of NTP, leap seconds, users manually adjusting the system time, etc, it can tick backwards. When that happens, your previously deleted records are now scheduled to be deleted in the future.

For single node deployments, this can potentially be fixed by using something like [`process.hrtime`](https://nodejs.org/api/process.html#process_process_hrtime_time) as a time source. But clock drift between nodes is still a problem.

## Clock drift between nodes

Even if we use a monotonic time source, keeping multiple clocks in sync is really hard. [Google resorts to atomic clocks](https://www.cockroachlabs.com/blog/living-without-atomic-clocks/) to do it! The linked article mentions that NTP, which I can imagine is more typical than atomic clocks for the majority of Sequelize's users, will typically give you an upper bound for clock offsets at "somewhere between 100ms and 250ms."

## Prior art

I haven't researched this too much, but a cursory search reveals that alternatives like ActiveRecord's ActsAsParanoid do not compare timestamps and [instead compare to a sentinel value](https://github.com/ActsAsParanoid/acts_as_paranoid/blob/6fabadbb31e89f9be206c141387a8977b7ce7781/lib/acts_as_paranoid/core.rb#L49). I'd be curious to see if they're consciously not supporting future deletes or have just never implemented it. Regardless, if Sequelize's behavior in this space really is unique, that seems like more evidence that it's not a safe practice. Especially when it's for a feature (scheduled deletes) that few may use.

## Conclusion

I think scheduled soft deletes weaken the consistency of Sequelize. It is very possible to see a soft deleted record 250ms or more after it was soft deleted. Sequelize should consider changing the default behavior to something safer.

holm:
### Pull Request check-list

### Description of change

We are currently upgrading to Sequelize v4. As many others we have run into issue #8496. I do not believe having this functionality is good in any way, and I think it was a mistake for make this the new default in v4. The problems around this is well explained in the linked issue.

This PR tries to move forward by reverting paranoid to work like in v3, with an option to have it work like v4. This will likely be a semver-major change to land in v5.

Closes #8496

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -69,7 +69,8 @@ class Model {
       }
     }
 
-    if (!model.options.timestamps || !model.options.paranoid || options.paranoid === false) {
+    const paranoid = options.paranoid !== undefined ? options.paranoid : model.options.paranoid;
+    if (!model.options.timestamps || !paranoid) {
       // This model is not paranoid, nothing to do here;
       return options;
     }
@@ -81,12 +82,20 @@ class Model {
 
     let deletedAtDefaultValue = deletedAtAttribute.hasOwnProperty('defaultValue') ? deletedAtAttribute.defaultValue : null;
 
-    deletedAtDefaultValue = deletedAtDefaultValue || {
-      [Op.or]: {
-        [Op.gt]: now,
-        [Op.eq]: null
+    if (!deletedAtDefaultValue) {
+      if (paranoid === 'legacy') {
+        deletedAtDefaultValue = {
+          [Op.or]: {
+            [Op.gt]: now,
+            [Op.eq]: null
+          }
+        };
+      } else {
+        deletedAtDefaultValue = {
+          [Op.eq]: null
+        };
       }
-    };
+    }
 
     deletedAtObject[deletedAtAttribute.field || deletedAtCol] = deletedAtDefaultValue;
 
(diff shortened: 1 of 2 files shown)


holm:
Thanks for the review. I agree that the term `simple` is not ideal, and happy to take suggestions. Not sure `legacy` is great either, as it implies it shouldn't be used.

I initially added this as a separate option, but found that I liked the shared option better. It also has a nice resemblance with the timestamps properties, where you can use a boolean or a string. Maybe the best option is to have a boolean value be the backwards compatible value, and encourage new users to use a string only, either `time-based` or `simple` (or whatever name we come up with). `true` would be equivalent to `time-based`.

holm:
Any chance for getting a review of this?

holm:
I guess we could also invert this and have the v4 logic be ‘legacy’. It would fix the issue with the simple terminology and have most users be on the default config.

holm:
I changed this PR now to restore the old functionality, with an option to use `legacy` to have it work like in v4.

I have updated the docs, but it would be great if the maintainers could take care of any other communication required, as I think that would be much more efficient.

holm:
What can I do to help move this PR forward?

blacksun1:
Any word on a version 5 release or when this code will be available to start using.

janmeier:
@sushantdhiman Is working on v5 [here](https://github.com/sequelize/sequelize/compare/v5), should be ready soon

sushantdhiman:
This fix is now included in v5 branch https://github.com/sequelize/sequelize/commit/014f2f5b28712f2304aa16c02883d19b1b43b5ab

PR #9030, opened 2018-02-10
fix  #8984  duplicate indexes when sync with alter true

### LINKED ISSUE #8984: {alter: true} creates new indexe
## What are you doing?
I use {alter: true} in dev mode to update the model.
```js
sequelize
  .sync({ alter: true })
  .then(() => {
    console.error('Connected to db!');
    return Promise.resolve();
  })
  .catch(err => {
    console.error('Error connecting to db!');
    return Promise.reject(err);
  });
```

Update model and indexes

Every time the table is altered it adds new indexe.
For instance I now got 64 unique indexe on email. It should remove the indexe or at least do nothing if the constraint already exists.

__Dialect:__ mysql 
__Dialect version:  mysql2 1.5.1
__Database version: Mysql 5.7
__Sequelize version: 4.32.2

abdoolly:
fix  #8984 

### Pull Request check-list

I actually ran the linting and fixed the lint but , I did not ran any other tests cause the change is too small to run a test for it. I think travis will run the tests.

change was too small to run a test for it.

### Description of change

I added a new function in the query interface to remove all indexes in a table and called it in the
sync function before making any changes in the table to eliminate any duplicates that would happen
while altering

This is my first pull request I hope it helps I also hope that this is the right way to fix this problem
Thanks

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -766,6 +766,25 @@ class QueryInterface {
   }
 
   /**
+   * remove indexes from a table except the primary key 
+   *
+   * @param {String} tableName          Table name to remove index from 
+   * @param {Object} [options]          Query options
+   *
+   * @return {Promise}
+   */
+  removeIndexes(tableName, options) {
+
+    return this.showIndex(tableName, options)
+      .then(indexes => {
+        const indexesToRemove = indexes.map(index => index.primary === false ?
+          this.removeIndex(tableName, index.name) : null);
+
+        return Promise.all(indexesToRemove);
+      });
+  }
+
+  /**
    * Add constraints to table
    *
    * Available constraints:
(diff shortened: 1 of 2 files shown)


lib/query-interface.js
```diff
@@ -765,6 +765,25 @@ class QueryInterface {
...
+   * @param {Object} [options]          Query options
+   *
+   * @return {Promise}
+   */
+  removeIndexes(tableName, options) {
+
+    return this.showIndex(tableName, options)
+      .then(indexes => {
+        const indexesToRemove = indexes.map(index => index.primary === false ?
+          this.removeIndex(tableName, index.name) : null);
```
eseliger: shouldn't `options` be passed here?

Issue #9050, opened 2018-02-14
Raw query replacements in MySQL INSERT query


Kirill89:
```js
db.query('INSERT INTO product (a, b) VALUES ?;', {
        replacements: [[['a', 'b'], ['c', 'd']]],
        type: Sequelize.QueryTypes.INSERT
});
```

## What do you expect to happen?

```mysql
INSERT INTO product (a, b) VALUES ('a', 'b'), ('c', 'd');
```

## What is actually happening?

```mysql
INSERT INTO product (a, b) VALUES 'a','b','c','d';
```

__Dialect:__ mysql
__Sequelize version:__ 4.32.3

So, the question is, how to insert non-determined number of items in one query?

sushantdhiman:
Have you tried

```js
db.query('INSERT INTO product (a, b) VALUES (?), (?);', {
        replacements: [['a', 'b'], ['c', 'd']],
        type: Sequelize.QueryTypes.INSERT
});
```

Might be possible, but not sure

Kirill89:
@sushantdhiman yes, it will work. But how to do that if I don't know number of items?

sushantdhiman:
Well you can repeat, `(?)` to length of your array

```js
db.query(
  `INSERT INTO product (a, b) VALUES ${data.map(a => '(?)').join(',')};`,
  {
     replacements: data,
     type: Sequelize.QueryTypes.INSERT
  }
);
```

Kirill89:
@sushantdhiman, but most of MySQL drivers can do it from the box.

I probably need to describe my case. In my project we are doing a migration from plain queries to ORM. As a first step we decide to use sequelize library to perform all SQL requests. And it was bad news for me that it's not compatible with usual placeholder syntax (like [here](https://github.com/mysqljs/mysql)).

sushantdhiman:
We are using `mysql2`, I dont think it support array replacement

sushantdhiman:
Can you test if `mysql2` can handle that sort of replacement, if it can I am willing to tag this as a feature request

Kirill89:
@sushantdhiman, in `mysql2` it works.

```javascript
const mysql = require('mysql2');
const connection = mysql.createConnection({host: 'localhost', user: 'root', database: 'test', debug: true});

connection.query(
    'INSERT INTO product (a, b) VALUES ?;',
    [[['a', 'b'], ['c', 'd']]],
    (...args) => console.log(args)
);
```

[...]

PR #9054, opened 2018-02-15
feat(query): add ability to pass arrays as replacements #9050

Kirill89:
According to the [feature request](https://github.com/sequelize/sequelize/issues/9050) added ability to pass arrays to `sequelize.query()` as replacements.

### DIFF
--- a/lib/sql-string.js
+++ b/lib/sql-string.js
@@ -4,6 +4,21 @@ const dataTypes = require('./data-types');
 const util = require('util');
 const _ = require('lodash');
 
+function arrayToList(array, timeZone, dialect, format) {
+  return array.reduce((sql, val, i) => {
+    if (i !== 0) {
+      sql += ', ';
+    }
+    if (Array.isArray(val)) {
+      sql += `(${arrayToList(val, timeZone, dialect, format)})`;
+    } else {
+      sql += escape(val, timeZone, dialect, format);
+    }
+    return sql;
+  }, '');
+}
+exports.arrayToList = arrayToList;
+
 function escape(val, timeZone, dialect, format) {
   let prependN = false;
   if (val === undefined || val === null) {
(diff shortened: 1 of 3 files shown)


lib/sql-string.js
```diff
@@ -4,6 +4,23 @@ const dataTypes = require('./data-types');
 const util = require('util');
 const _ = require('lodash');
 
+function arrayToList(array, timeZone, dialect, format) {
```
sushantdhiman: Can you cleanup this method, may be use reduce as well

lib/sql-string.js
```diff
@@ -44,7 +61,7 @@ function escape(val, timeZone, dialect, format) {
     if (dialect === 'postgres' && !format) {
       return dataTypes.ARRAY.prototype.stringify(val, {escape: partialEscape});
     }
-    return val.map(partialEscape);
```
sushantdhiman: I believe this will break other dialects, MSSQL i think. Will probably need more tests

lib/sql-string.js
Kirill89: Done

lib/sql-string.js
Kirill89: I added integration test, looks like it's okay with MSSQL.

Kirill89:
@sushantdhiman , any updates so far?

lib/sql-string.js
Vovka19: I will learn. didn't sleep 3 days. thank you

test/integration/sequelize.test.js
```diff
@@ -273,6 +273,23 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
       return this.sequelize.query(this.insertQuery);
     });
 
+    it('executes a query if a placeholder value is an array', function() {
```
eseliger: Maybe adding a unit test to check for the output of the query generator as well might be good?

Kirill89: @eseliger, done.

Kirill89:
@sushantdhiman, @eseliger, any updates? I really need this feature.

PR #9076, opened 2018-02-20
docs(Getting Started): Add `operatorsAliases: Op`

rodorgas:
### Description of change

Running the code provided on Getting Started cause a deprecation warning. This PR adds `operatorsAliases: Op` to the code on Getting Started, which is already recommended in other documents, like [Querying](http://docs.sequelizejs.com/manual/tutorial/querying.html#operators-aliases):

> For better security it is highly advised to use Sequelize.Op and not depend on any string alias at all.

This change was suggested in https://github.com/sequelize/sequelize/issues/8417#issuecomment-337884136 a few months ago and received positive response.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -33,7 +33,7 @@ const Sequelize = require('sequelize');
 const sequelize = new Sequelize('database', 'username', 'password', {
   host: 'localhost',
   dialect: 'mysql'|'sqlite'|'postgres'|'mssql',
-  operatorsAliases: Sequelize.Op,
+  operatorsAliases: false,
 
   pool: {
     max: 5,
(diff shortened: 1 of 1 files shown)


eseliger:
I would at least leave a note that this should be adjusted to the specific needs to disable as many operators as possible, @sequelize/orm what do you guys think?

janmeier:
I don't think this should be added as a default option. As a new user, this should not be the first thing I see - It's going to confuse the hell out of me :D

As a new user I shouldn't see that it is possible to use `$gt` etc. It is bad practice, so it's better if I never learn it.

I think we should rather change the example to use `operatorsAliases: false`

rodorgas:
I agree it's better to disable it completely on the Getting Started snippet.

Should we include a brief text describing what it means, and a link to [Querying - Operators Aliases](http://docs.sequelizejs.com/manual/tutorial/querying.html#operators-aliases)?

It could be something like:

> The `operatorsAliases` option is disabled for improved security. Sequelize can take string aliases for operators used in complex comparisons, but you would need to sanitize your input to avoid injection. You can [read more about it here](http://docs.sequelizejs.com/manual/tutorial/querying.html#operators-aliases).

PR #9085, opened 2018-02-21
Update model.js

leonma:
### Pull Request check-list

### Description of change
This change allow user can be able to remove the condition of deletedAt

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -92,7 +92,7 @@ class Model {
 
     if (Utils.isWhereEmpty(options.where)) {
       options.where = deletedAtObject;
-    } else {
+    } else if (!options.where.hasOwnProperty(deletedAtAttribute.field || deletedAtCol)) {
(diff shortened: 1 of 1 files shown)


eseliger:
I think what you are trying is already covered by passing `paranoid: false` to the find function. eg.

```js
const user = await User.create({})
await user.destroy()
await User.findById(user.id, { paranoid: false }) // will return user although hes "deleted"
```

leonma:
@eseliger yes, it works for find function, but can not be woking on scopes.
 eg.
```
let model = sequelize.define('user', {
  // attributes
  deletedAt: {
    field: 'deleted_at',
    type: db.types.DATE
  }
}, {
  paranoid: true,
  underscored: true,
  scopes: {
      deleted: {
        where: {
          deletedAt: {
            [sequelize.Op.not]: null
          }
        }
      }
    }
})

model.scope('deleted').findAll();
```

PR #9133, opened 2018-03-03
fix(model): Made sure errors with messages are passed to AggregateError for bulk creates.

### LINKED ISSUE #8989: Aggregate errors do not print inner error(s) correctly
## What are you doing?
The gist of the below code is this: I tried to insert into a join table. The join table has a NOT NULL column, but I tried to insert NULL by accident.

```

const sequelize = new Sequelize({ });

const STUDENTS = sequelize.define('STUDENTS', {
    STUDENT_ID: {
        type: DataTypes.CHAR(10),
        allowNull: false,
        primaryKey: true,
    }
});

const CONTACTS = sequelize.define('CONTACTS', {
    CONTACT_ID: {
        type: DataTypes.INTEGER(11),
        allowNull: false,
        primaryKey: true,
    }
});

const STUDENT_CONTACTS = sequelize.define('STUDENT_CONTACTS', {
    CONTACT_PRIORITY: {
        type: DataTypes.INTEGER(6),
        allowNull: false,
        validate: {
            min: 0,
        },
    }
});

STUDENTS.belongsToMany(CONTACTS, {
    as: 'contacts',
    foreignKey: 'STUDENT_ID',
    otherKey: 'CONTACT_ID',
    through: STUDENT_CONTACTS,
    constraints: false,
});

CONTACTS.belongsToMany(STUDENTS, {
    as: 'students',
    through: STUDENT_CONTACTS,
    foreignKey: 'CONTACT_ID',
    otherKey: 'STUDENT_ID',
    constraints: false,
});

const transaction = await sequelize.transaction({ autocommit: false });

try {
    const student = await STUDENTS.create({ }, { transaction });
    const contact = await CONTACTS.create({ }, { transaction });

    contact.STUDENT_CONTACTS = { }; // CONTACT_PRIORITY missing!
    await student.addContact(contact, { transaction });
    await transaction.commit();
} catch (ex) {
    console.error(ex.toString());
    await transaction.rollback();
}
```

An error string that correctly renders suberrors as strings. In this context, I was able to dig through `ex` while debugging and pull out the real suberror, so I would expect the full error output to be:
```
AggregateError of:
    "notNull Violation: STUDENT_CONTACTS.CONTACT_PRIORITY cannot be null"
```

Shell output:

```
AggregateError of:
    [object Object]
```

__Dialect:__ mysql2
__Dialect version:__ 1.5.1
__Database version:__ Ver 14.14 Distrib 5.7.20, for macos10.12
__Sequelize version:__ 4.32.2

aceew:
### Pull Request check-list

### Description of change
Previously objects passed to the AggregateError did not have a message property, causing them to be
rendered as [object Object] in the console log. This change fixes the issue of a missing message
property.

fixes #8989

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2374,11 +2374,7 @@ class Model {
 
         return Promise.map(instances, instance =>
           instance.validate(validateOptions).catch(err => {
-            /**
-             * Simply extending the value of `err` would cause a breaking change as the `errors`
-             * property is expected on each item in the array of errors.
-             */
-            errors.push(new sequelizeErrors.RecordLinkedError(err, instance));
+            errors.push(new sequelizeErrors.BulkRecordError(err, instance));
           })
         ).then(() => {
           delete options.skip;
--- a/lib/model.js
+++ b/lib/model.js
@@ -2374,7 +2374,11 @@ class Model {
 
         return Promise.map(instances, instance =>
           instance.validate(validateOptions).catch(err => {
-            errors.push({record: instance, errors: err});
+            /**
+             * Simply extending the value of `err` would cause a breaking change as the `errors`
+             * property is expected on each item in the array of errors.
+             */
+            errors.push(new sequelizeErrors.RecordLinkedError(err, instance));
           })
         ).then(() => {
           delete options.skip;
--- a/docs/instances.md
+++ b/docs/instances.md
@@ -167,7 +167,7 @@ Task.bulkCreate([
 }).spread((affectedCount, affectedRows) => {
   // .update returns two values in an array, therefore we use .spread
   // Notice that affectedRows will only be defined in dialects which support returning: true
-  
+
   // affectedCount will be 2
   return Task.findAll();
 }).then(tasks => {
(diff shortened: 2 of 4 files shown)


aceew:
Is it possible that the issue noted (here)[https://github.com/sequelize/sequelize/blob/master/test/integration/configuration.test.js#L30]:  is happening for other dialects than 'mssql'? I can't replicate [this test](https://travis-ci.org/sequelize/sequelize/jobs/348511120) failing locally

lib/model.js
```diff
@@ -2374,7 +2374,11 @@ class Model {
 
         return Promise.map(instances, instance =>
           instance.validate(validateOptions).catch(err => {
-            errors.push({record: instance, errors: err});
+            /**
+             * Simply extending the value of `err` would cause a breaking change as the `errors`
+             * property is expected on each item in the array of errors.
+             */
+            errors.push(new sequelizeErrors.RecordLinkedError(err, instance));
```
sushantdhiman: Why not just push validation error

aceew: It would cause a breaking change, a property of `record` and `errors` is expected in each item.

sushantdhiman: I think we can simply do this

```js
err.record = instance;
errors.push(err);
```

In v5 we can remove `record` as `instance` property is already available in ValidationErrorItem

aceew:
Hey thanks, sorry I've been afk all week

Issue #9157, opened 2018-03-08
limit/offset produces invalid result if included model has where clause


jharting:
## What are you doing?
I have two entities in many-to-many relationship:

```js
      const User = this.sequelize.define('User', {
        name: {
          type: DataTypes.STRING,
          primaryKey: true
        }
      }, {timestamps: false});

      const Project = this.sequelize.define('Project', {
        name: {
          type: DataTypes.STRING,
          primaryKey: true
        }
      }, {timestamps: false});

      User.belongsToMany(Project, {through: 'user_project'});
      Project.belongsToMany(User, {through: 'user_project'});
```

This is the test data:
User **Alice** is associated with projects: **alpha, charlie**
User **Bob** is associated with projects: **bravo**

When I query the model like this:

```js
Project.findAll({
  include: [{
    model: User,
    where: {
      name: 'Alice'
    }
  }],
  order: ['name']
```

I get the expected projects (**alpha, charlie**).

The problem occurs when I add pagination support. If I add limit and offset to the query like this:

```js
Project.findAll({
  include: [{
    model: User,
    where: {
      name: 'Alice'
    }
  }],
  order: ['name'],
  limit: 1,
  offset: 1
```

## What do you expect to happen?
I expect to get the second project associated with Alice i.e. **charlie**

## What is actually happening?
I am getting an **empty result list**.

The query that Sequelize generates is:
```
[...]

PR #9165, opened 2018-03-09
fix(query-generator): use where clause of included entity in subquery…

jharting:
… (#9157)

Closes https://github.com/sequelize/sequelize/issues/9157
Note that this does not fix LIMIT + included entities with WHERE clause completely.
This fix addresses the problem for entities included directly from the top-level entity.
The problem remains for entities that are nested more than level (as the skipped test demonstrates).

I'd like to have this tested and reviewed. If this works I plan to look into the more complex (more levels of nesting) scenarios.

### Description of change

Looks like most of the code for this was already in place. For some reason it:
* was triggering for where clauses only, ignoring includes with require=true but no where clause
* even if triggered it did not copy the where clause over to the subquery

This was pretty simple. If the fix is accepted I plan on looking further into more complex scenarios (more levels of nesting).


test/integration/include/limit.test.js
```diff
@@ -0,0 +1,320 @@
...
+  Sequelize = require('../../../index'),
+  expect = chai.expect,
+  Support = require(__dirname + '/../support'),
+  DataTypes = require(__dirname + '/../../../lib/data-types'),
+  Promise = Sequelize.Promise;
+
+describe(Support.getTestDialectTeaser('Include'), () => {
+  describe('LIMIT', () => {
+
+    let User = false;
```
sushantdhiman: Please share these models using test scopes rather than global variables 

```js
beforeEach(function() {
  this.Model = sequelize.define(/***/)
})

it('your test', function() {
   this.Model.findAll(/***/)
}); 
```

Issue #7606, opened 2017-05-04
sync() with new alter mode fails on second run when using CONSTRAINTS


skn3:
## What you are doing?

```js
'use strict';

const Sequelize = require('sequelize');

//first connect to the database
const database = new Sequelize('database', 'user', 'password', {
    dialect: 'postgres',
    host: '127.0.0.1',
    port: 5432,

    pool: {
        max: 5,
        min: 0,
        idle: 10000
    },
});

return database.authenticate()
.then(() => {
    //success connecting,
    console.log('database connected');
    return database;
})
.then(database => {
    const containerModel = database.define('TestContainer', {
        id: {
            type: Sequelize.DataTypes.UUID,
            defaultValue: Sequelize.DataTypes.UUIDV4,
            primaryKey: true,
            allowNull: false
        },
    });

    const itemModel = database.define('TestItem', {
        id: {
            type: Sequelize.DataTypes.UUID,
            defaultValue: Sequelize.DataTypes.UUIDV4,
            primaryKey: true,
            allowNull: false
        },
    });

    const userModel = database.define('TestUser', {
        id: {
            type: Sequelize.DataTypes.UUID,
            defaultValue: Sequelize.DataTypes.UUIDV4,
            primaryKey: true,
            allowNull: false
        },
    });

    containerModel.belongsTo(userModel);
    itemModel.belongsTo(userModel);
    containerModel.belongsToMany(itemModel, {through: 'TestContainerItems', timestamps: false, onDelete: 'CASCADE'});

    return database.sync({
        alter: true,
        force: false,
    });
})
.catch(err => {
    //failed to connect
    console.error(err);
    process.exit(1);
});
```

## What do you expect to happen?
Being able to run the example multiple times without fail

## What is actually happening?
On second run, an error occurs:

> SequelizeDatabaseError: constraint "TestUserId_foreign_idx" for relation "TestContainers" already exists

__Dialect:__ postgres
__Database version:__ 9.6.2
__Sequelize version:__ git master

skn3:
No idea if this is safe, but needed a quick fix for testing:

```javascript
function dropForeignKeyConstraints(database) {
    //this is a hack for dev only!
    //todo: check status of posted github issue, https://github.com/sequelize/sequelize/issues/7606
    const queryInterface = database.getQueryInterface();
    return queryInterface.showAllTables()
    .then(tableNames => {
        return Promise.all(tableNames.map(tableName => {
            return queryInterface.showConstraint(tableName)
[...]

PR #9033, opened 2018-02-10
fix(indexes): duplicate unique indexes with sync

sushantdhiman:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](../CONTRIBUTING.md)?

### Description of change

Close https://github.com/sequelize/sequelize/issues/6134
Close https://github.com/sequelize/sequelize/issues/9027
Close https://github.com/sequelize/sequelize/issues/7947
Close https://github.com/sequelize/sequelize/issues/7915
Close https://github.com/sequelize/sequelize/issues/7893

PR #9179, opened 2018-03-13
fix(model/sync): remove duplicated index after change column

lext-7:
### Pull Request check-list

### Description of change

Remove duplicated indexes after change column in sync.
https://github.com/sequelize/sequelize/issues/7606#issuecomment-364340666

As for alerting a unique column(`alert xxx unique`) will create a unique index, there will be multiple unique indexes for a column after sync with alert multiple times. So it need to find duplicated indexes after change column and remove them.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1182,6 +1182,19 @@ class Model {
         // Assign an auto-generated name to indexes which are not named by the user
         this.options.indexes = this.QueryInterface.nameIndexes(this.options.indexes, this.tableName);
 
+        const existedIndex = {};
+        const duplicatedIndexes = [];
+
+        indexes.forEach(index => {
+          const fieldNameStr = index.fields.map(field => field.attribute).join('_');
+          const key = `${index.type}_${index.primary}_${index.unique}_${fieldNameStr}`;
+          if (existedIndex[key]) {
+            duplicatedIndexes.push(index);
+          } else {
+            existedIndex[key] = true;
+          }
+        });
+
         indexes = _.filter(this.options.indexes, item1 =>
           !_.some(indexes, item2 => item1.name === item2.name)
         ).sort((index1, index2) => {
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -1182,6 +1182,22 @@ class Model {
         // Assign an auto-generated name to indexes which are not named by the user
         this.options.indexes = this.QueryInterface.nameIndexes(this.options.indexes, this.tableName);
 
+        const existedIndexes = {};
+        const duplicatedIndexes = [];
+
+        indexes.forEach(index => {
+          const fieldNameStr = index.fields.map(field => field.attribute).join('_');
+          const key = `${index.type}_${index.primary}_${index.unique}_${fieldNameStr}`;
```
sushantdhiman: `nameIndexes` method also doing something similar, to keep names consistent we should share this code

test/integration/sequelize.test.js
```diff
@@ -1018,6 +1018,27 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
         .to.be.rejectedWith('Database "cyber_bird" does not match sync match parameter "/$phoenix/"');
     });
 
+    it('won\'t create duplicated unique indexes', function() {
```
sushantdhiman: Move this test [here](https://github.com/sequelize/sequelize/blob/83dfc5ae9cf671d448690bc3c8a6d0f5ef45a681/test/integration/model/sync.test.js#L133)

Change title `should not create duplicate unique indexes with unique:true`

lext-7:
Sorry, I finally find this way cannot solve this problem perfectly. 

I think `changeColumnQuery()` should be async and return promise, Because it needs to query indexes and constraints. According to existed indexes and constraints, ignore duplicated indexes or constraint and remove those removed indexes or constraints. 

I think removing constraint or index before changing column or after changing column is a hack way for `sync()`. If you use `changeColumn` function directly, error will be raised still.

But changing `changeColumnQuery()` to be async will take a lot of work and conflict with existing api.
What's your opinion?

sushantdhiman:
I agree `changeColumnQuery` needs to take care of existing indexes, can we query for those indexes before hand and pass to `changeColumnQuery` and take care of them

There is one more issue I discovered when working on https://github.com/sequelize/sequelize/pull/9033

For MySQL if we issue `ALTER TABLE <table> CHANGE <col> UNIQUE` even if nothing is change a new index will be created. This is an inherent bug in MySQL

So whenever we try to `sync({ alter: true })` and column with unique constraint will get a new index. Proper way to handle this situation is

1. Fix that bug with MySQL :D 
2. Scan indexes to see if we are creating any duplicate indexes, Which I think is quite some work in itself

You are suggesting (2), you are right it will take some work but this wont be breaking change `changeColumnQuery` (From Query Generator) is a private API on other hand `changeColumn` (From QueryInterface) is public. So there should be no problem changing that query generator method.

Again this is not a critical issue as Sequelize recommend against using `{ alter: true }` in production

sushantdhiman:
Cleanup, closing this PR. If someone wants to takeover please open a new PR

Issue #9015, opened 2018-02-08
Support Node 6 and up


sushantdhiman:
https://github.com/nodejs/Release#release-schedule

Node 4 Maintenance LTS is ending soon (on 2018-04-30) and all cloud function providers AWS, Google, Azure and IBM supports Node v6

sushantdhiman:
Included in v5 branch

SimonSchick:
Would it be possible to skip right to v8, seeing that v8 and v10 are both LTS?

sushantdhiman:
No

SimonSchick:
> No.

Is there any particular reason why?

dijonkitchen:
Probably because version 6.x is supported as an LTS until 2019.

sushantdhiman:
Main reason is already stated

> all cloud function providers AWS, Google, Azure and IBM supports Node v6

SimonSchick:
@sushantdhiman all cloud providers you named now fully support v8.
v6 LTS ends in ~4 months, I think it would be sensible to release v5 targeting v8 instead.

sushantdhiman:
I agree @SimonSchick , I planned v5 to be a quick iteration but release has been stretched further. Now it make sense to bump directly to Node 8 at least.

`cls` is a blocker atm, will have to look into implementing https://github.com/Jeff-Lewis/cls-hooked

SimonSchick:
Can you elaborate what it is blocking exactly?

sushantdhiman:
@SimonSchick first we have to migrate our codebase to use `cls-hooked`, probably locking Node to 8.10 or above (no real issue with that), then we have to write tests for functionality using cls with native async await especially important for transactions.

Migration to cls-hooked is required as `continuation-local-storage` itself is using `AsyncListener` which wont work for Node 8 or up

This is mostly some (easy) modernization work with tedious rewrite (of tests) overall but real blocker if we want to support cls in Node 8

SimonSchick:
I assume you are going to do this?
I'm not exactly familiar with the cls stuff since I've never used it.

ayushnigamsworld:
> @SimonSchick first we have to migrate our codebase to use `cls-hooked`, probably locking Node to 8.10 or above (no real issue with that), then we have to write tests for functionality using cls with native async await especially important for transactions.
> 
> Migration to cls-hooked is required as `continuation-local-storage` itself is using `AsyncListener` which wont work for Node 8 or up
> 
[...]

PR #9212, opened 2018-03-21
v5-beta

sushantdhiman:
# v5-beta

-----------------------------------------------------------------------------------------------

### Support for Node 6 and up

Sequelize v5 will only support Node 6 and up [#9015](https://github.com/sequelize/sequelize/issues/9015)

### Model

**Attributes**

`Model.attributes` now removed, use `Model.rawAttributes`. [#5320](https://github.com/sequelize/sequelize/issues/5320)

__Note__: _Please dont confuse this with `options.attributes`, they are still valid_

**Paranoid Mode**

With v5 if `deletedAt` is set, record will be considered as deleted. So `paranoid` option will only use `deletedAt` as flag.

In v4 it used to compare current time with `deletedAt`. [#8496](https://github.com/sequelize/sequelize/issues/8496)

-----------------------------------------------------------------------------------------------

### QueryGenerator

- Common queries and quote helpers now separated
- QueryGenerator now converted to proper classes
- Refactored initialization options

### Dialect

- Converted to classes


sushantdhiman:
Going to merge this, this will be released as `5-beta` under `next` tag.

If there are any extra reviews or improvements from @sequelize/orm I will be happy to take care of that in next beta releases

PR #9248, opened 2018-03-28
fix(model): don't add LIMIT in findOne() queries on unique key

jharting:
### Pull Request check-list

### Context
When querying with `findOne()` Sequelize adds `LIMIT 1` to the query to limit the result size. In combination with duplicating joins this may result in a complicated query (subqueryFilter). An optimization exists in Sequelize where if the query's where clause contains the primary key then `LIMIT 1` is skipped.

### Description of change
This change extends the optimization to also cover unique columns.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1737,10 +1737,13 @@ class Model {
     options = Utils.cloneDeep(options);
 
     if (options.limit === undefined) {
-      const pkVal = options.where && options.where[this.primaryKeyAttribute];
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -1737,10 +1737,13 @@ class Model {
     options = Utils.cloneDeep(options);
 
     if (options.limit === undefined) {
-      const pkVal = options.where && options.where[this.primaryKeyAttribute];
+      const uniqueSingleColumns = _(this.uniqueKeys).values().filter(c => c.fields.length === 1).map('column').value();
```
janmeier: use `.chain` here to make it explicit what's happening

PR #9252, opened 2018-03-29
fix(model): apply scopes properly, closes #9087

### LINKED ISSUE #9087: Scopes' includes are not shallowly merged as the documentation claims
## What are you doing?

I am testing with a minimal example whether or not scopes' includes are shallowly merged, as the documentation claims: *`limit`, `offset`, `order`, `paranoid`, `lock` and `raw` are overwritten, while `where` and `include` are shallowly merged.* (source: http://docs.sequelizejs.com/manual/tutorial/scopes.html)

```js
const Sequelize = require("sequelize");
const sequelize = new Sequelize({ dialect: "sqlite", storage: "db.sqlite" });
const Foo = sequelize.define("foo", { name: Sequelize.STRING }, { timestamps: false });
const Bar = sequelize.define("bar", { name: Sequelize.STRING }, { timestamps: false });
Foo.hasMany(Bar, { foreignKey: "fooId" });
Foo.addScope("testScope", {
    include: [{
        model: Bar,
        where: {
            name: "bar1"
        }
    }]
});
    .then(() => Foo.create({ name: "foo1" }))
    .then(foo1 => Bar.create({ name: "bar1", fooId: foo1.id }))
    .then(() => Foo.create({ name: "foo2" }))
    .then(foo2 => Bar.create({ name: "bar2", fooId: foo2.id }))
    .then(() => Foo.scope("testScope").findAll({
        include: [{
            model: Bar,
            attributes: { exclude: ["name"] }
        }]
    }))
    .then(result => console.log(JSON.stringify(result, null, 4)));
```

I expected the scope to be merged with the findAll options, which should yield the following output:

```js
[
    {
        "id": 1,
        "name": "foo1",
        "bars": [
            {
                "id": 1,
                "fooId": 1
            }
        ]
    }
]
```

The output was:

```js
[
    {
        "id": 1,
        "name": "foo1",
        "bars": [
            {
                "id": 1,
                "fooId": 1
            }
        ]
    },
    {
        "id": 2,
        "name": "foo2",
        "bars": [
            {
                "id": 2,
                "fooId": 2
            }
        ]
    }
]
```

Clearly, the scope was completely overwritten by the findAll options.

I wanted my code to behave exactly as if I had called the following:

```js
Foo.scope("testScope").findAll({
    include: [{
        model: Bar,
        where: {
            name: "bar1"
        },
        attributes: { exclude: ["name"] }
    }]
})
```

__Dialect:__ sqlite (sqlite3 v3.1.13)
__Dialect version:__ 3.1.13
__Sequelize version:__ 4.33.4

**Note:** if somehow this is the correct behavior, honestly the docs should be clarified. The `include` field is an array of objects, and "shallowly merging" an array of objects is not perfectly clear, does it mean that the arrays are concatenated? I hope not. I understand that the behavior I'm expecting is not trivial to code, but if the documentation claims it, it should do it. My understanding of "shallowly merging" in this context would be that the objects in the include arrays are "matched" to each other based on the "model-as" pairs, shallowly merging the other fields (and hopefully, recursively, since I could have nested includes in my scopes).

Thank you very much in advance, sequelize has been very useful so far :grin:

contra:
### Pull Request check-list

### Description of change

Information: https://github.com/sequelize/sequelize/issues/9087#issuecomment-377330689

I would recommend backporting this to v4 as well since this is a security issue. This would be hard/impossible to exploit in a targeted way so I don't think it deserves a CVE or private disclosure, but this could be a bad leak of sensitive fields for some people.


sushantdhiman:
Closing this PR, Added https://github.com/sequelize/sequelize/issues/9087 to v5 milestone so this and other related scope issues will also be a focus of v5 release

contra:
@sushantdhiman What was wrong with the PR? I think it solved the issue pretty simply.

test/integration/model/scope.test.js
```diff
@@ -108,5 +108,37 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           return this.ScopeMe.scope('issue8473').findAll();
         });
     });
+
+    it('should be applied before query options for attributes exclude', function() {
```
sushantdhiman: Its worth adding a test for case when `scope` is trying to exclude some attribute but actual call is trying to include it. Which again comes to a discussion matter, I think scope should still exclude that attribute

contra: @sushantdhiman The test below this one does that (`should be applied before query options for attributes include`)

sushantdhiman: `lowAccess` is only setting attributes to `['other_value', 'access_level'],`, its not excluding `email`, so shouldn't `include` at method level should be able to inject `email` to attributes?

I think it would make sense if scope say `exclude` something then finder method should not be allowed to include that attribute, if scopes are just specifying attributes or using `include`, then finder method should be able to expand that list of attributes

contra: @sushantdhiman I think I see what you're asking for, I'll add that test but it does function like that.

contra:
@sushantdhiman Any update on this?

papb:
Hello @contra, thank you again for taking a look in my issue. Unfortunately, this PR does not solve it.

Looks like you're fixing something else with this PR though, so that's great :) Please just remove the "closes 9087" thing from the PR / commit, since it actually does not close it at all...

If you disagree and/or would like further clarifications, let me know.

sushantdhiman:
@contra I would like to wait for formal api discussion about scopes first, I don't use scopes in my own projects so I am not sure what works best here. Hopefully https://github.com/sequelize/sequelize/issues/9087 will result in an API proposal that works best for everyone, till then I am holding off any changes related to scope system

contra:
@papb Can you clarify how this doesn't fix #9087? If you could provide a test case that fails that would be ideal.

papb:
@contra Sure. [The very first comment of the issue itself](https://github.com/sequelize/sequelize/issues/9087#issue-299072157), for example, still behaves exactly as I posted.

Converting my post to a test is fairly straightforward:

```javascript
beforeEach(function() {
    this.Foo = this.sequelize.define('foo', { name: Sequelize.STRING }, { timestamps: false });
    this.Bar = this.sequelize.define('bar', { name: Sequelize.STRING }, { timestamps: false });
    this.Foo.hasMany(this.Bar, { foreignKey: 'fooId' });
    this.Foo.addScope('testScope', {
        include: [{
            model: this.Bar,
            where: {
                name: 'bar1'
            }
        }]
    });
});

it('should merge scope with findAll options correctly', function() {
    const self = this;
    return this.sequelize.sync({ force: true })
        .then(() => self.Foo.create({ name: 'foo1' }))
        .then(foo1 => this.Bar.create({ name: 'bar1', fooId: foo1.id }))
        .then(() => self.Foo.create({ name: 'foo2' }))
        .then(foo2 => this.Bar.create({ name: 'bar2', fooId: foo2.id }))
        .then(() => self.Foo.scope('testScope').findAll({
            include: [{
                model: this.Bar,
                attributes: { exclude: ['name'] }
            }]
        }))
        .then(result => {
            result = result.map(x => x.toJSON());
            expect(result).to.deep.equal([
                {
                    'id': 1,
                    'name': 'foo1',
                    'bars': [
                        {
                            'id': 1,
                            'fooId': 1
                        }
                    ]
                }
            ]);
    });
});
```

Also, [my other comment with extra examples](https://github.com/sequelize/sequelize/issues/9087#issuecomment-378440548) is also not covered so far. There I show three situations, which I call situations A, B and C, that should all have the same output (the output from A) but currently have different outputs. Another great test to be included would be a test checking those exact three situations for the same output (the one from A).

It would be great if you could add those tests to your PR and make them pass! Thank you very much once more for taking the time. If you need further clarifications, let me know.

sushantdhiman:
Fixed by https://github.com/sequelize/sequelize/issues/9087

PR #9258, opened 2018-04-01
Update querying.md with missing packages

christopher4lis:
const Sequelize = require('sequelize') is required for the statement below it to work: const Op = Sequelize.Op;

Including this will help prevent the doc viewer from having to experience an error stating the Sequelize is undefined.

### Pull Request check-list

### DIFF
--- a/docs/querying.md
+++ b/docs/querying.md
@@ -76,6 +76,7 @@ It's also possible to generate complex AND/OR conditions by nesting sets of `or`
 
 ### Basics
 ```js
+const Sequelize = require('sequelize');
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Sorry but we use this `Op = Sequelize.Op` in lots of other examples, considering `require` statement is just additional step that we have removed from many examples. 

As for docs not being properly linted, even after this statement is corrected with introduction of `require`, line with `Post` will fail for undefined variable.

I would love to cover docs with some linting but that too will require a bit restructuring

PR #9259, opened 2018-04-02
V4: can I set column's charset when define model?

ruanjiayou:
### Pull Request check-list
- model's define, set colunm's charset


sushantdhiman:
Please report bug on issue tracker

PR #9278, opened 2018-04-08
change(operators): use symbol operators

sushantdhiman:
### Pull Request check-list

### Description of change

Refactor tests and main library to use Symbol operators only


lib/sequelize.js
```diff
@@ -238,10 +238,9 @@ class Sequelize {
 
     this.dialect = new Dialect(this);
     this.dialect.QueryGenerator.typeValidation = options.typeValidation;
-    if (this.options.operatorsAliases === true) {
-      Utils.deprecate('String based operators are now deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/tutorial/querying.html#operators');
-      this.dialect.QueryGenerator.setOperatorsAliases(Op.LegacyAliases); //Op.LegacyAliases should be removed and replaced by Op.Aliases by v5.0 use
-    } else {
+
+    if (_.isPlainObject(this.options.operatorsAliases)) {
+      Utils.deprecate('String based operators are deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/tutorial/querying.html#operators');
```
yonjah: @sushantdhiman Are we sure we want to show the warning here ?
I think that if a developer decided to explicitly enable some operators there is no real security issue here.

sushantdhiman: Even if they are using `operatorsAliases` to limit operators there is still a security issue, we ideally want them to get rid of all string based operators and use `operatorsAliases` just as migration tool

yonjah: It just feels like we're not trusting the developers enough here.
It's a bit different to have operators no one is aware of than have the developer explicitly set operators he need to use. 
And if we think the aliases are not a good idea why not just remove them from this major version ?
I was thinking of maybe adding an option to set aliases per query so you can do something like 
```js
find({where: {id: {$gt: 1}, operatorsAliases: {$gt: Op.gt}});
```
Unfortunately I didn't have yet time to look at it and add a pull request (But I hope it should be relatively small change)
But if we just disable or warn on any alias enabled without allowing for any alternative, developers who actually need this aliases will have to implement their own parsing layer. And since this is only an issue with developers who fail to validate their inputs properly I doubt if they can be trusted to implement it securely

sushantdhiman: For me `operatorsAliases` is just a migration tool, I have seen people misuse them by setting `operatorsAliases: Sequelize.Op`.

If developer want to create parsing layer they can easily map their operators to Sequelize.

Providing `operatorsAliases` at different method level directly is again wrong, as it encourage use of string operators.

PR #9364, opened 2018-04-27
fix(range-bounds): observe inclusive range bounds

### LINKED ISSUE #8176: Range bounds are not applied
## What you are doing?
I compare time range with inclusive bounds.

```js
let range = [
      data.period[0],
      data.period[1]
    ]
    range.inclusive = true

const d = await model.findOne({
      where: {
        id: {
          $ne: data.id
        },
        StudentId: data.Student.id,
        TimePeriodId: data.TimePeriod.id,
        WeekDayId: data.WeekDay.id,
        level,
        period: {
          $overlap: range
        }
      }
    })
```

Query
```sql
SELECT "id", "period", "level", "studyYear", "order" FROM "AsSchedules" AS "AsSchedule" WHERE "AsSchedule"."id" != 0 AND "AsSchedule"."StudentId" = 507 AND "AsSchedule"."TimePeriodId" = 8 AND "AsSchedule"."WeekDayId" = 1 AND "AsSchedule"."level" = 1 AND "AsSchedule"."period" && '["2016-11-04 21:00:00.000 +00:00","2016-11-05 21:00:00.000 +00:00"]' LIMIT 1
```

But receive
```sql
SELECT "id", "period", "level", "studyYear", "order" FROM "AsSchedules" AS "AsSchedule" WHERE "AsSchedule"."id" != 0 AND "AsSchedule"."StudentId" = 507 AND "AsSchedule"."TimePeriodId" = 8 AND "AsSchedule"."WeekDayId" = 1 AND "AsSchedule"."level" = 1 AND "AsSchedule"."period" && '["2016-11-04 21:00:00.000 +00:00","2016-11-05 21:00:00.000 +00:00")' LIMIT 1
```

Property _inclusive_ lost after code
```js
```
in findOne() function.

__Dialect:__ postgres
__Database version: 9.6.3
__Sequelize version: 4.7.1

mjy78:
### Pull Request check-list

### Description of change

Ensure the 'inclusive' property on range bounds is not lost when querying.

Closes #8176, #8471


lib/utils.js
```diff
@@ -130,6 +130,13 @@ exports.formatNamedParameters = formatNamedParameters;
 function cloneDeep(obj) {
   obj = obj || {};
   return _.cloneDeepWith(obj, elem => {
+    // Special handling for the inclusive array property (for range types)
+    if (Array.isArray(elem) && elem.hasOwnProperty('inclusive')) {
```
sushantdhiman: This will execute for each clone operation I would like to avoid this if possible, there is also one more issue https://github.com/sequelize/sequelize/issues/8471

As you are using this type in real world, what do you think about having range type store values like

```js
[{  
  value: 10,
  inclusive: true
}, {
  value: 10,
  inclusive: true
}]
```

Rather than having non-standard array property?

mjy78: I agree that not having to worry about special handling for the non-standard "inclusive" array property would be much cleaner. We should move to a standard representation of range types for both query predicates and query results ensuring the inclusive property is still maintained when results are converted to JSON.

This will be a breaking change though, as we (and I expect many others) have a lot of code that expects range values returned in query results to be of the form

```
[ 10, 10 ]
```

Another approach might be to allow range types to have a 3rd "options" element in the array. When querying/inserting you could define your range as any of
```
[ 10, 10 ] // Defaults to inclusive [true, false]
[ 10, 10, { inclusive: true }]
[ 10, 10, { inclusive: [ true, false ] }]
```
Query results would always be returned as
```
[ 10, 10, { inclusive: [ true, false ] }]
```
This will still be a breaking change, but may lessen the impact on those queries that currently use ranges in their default form (without being concerned about inclusivity).

sushantdhiman: As we are under v5 beta process I would prefer just do one time switch to proper format, which will solve both [clone](https://github.com/sequelize/sequelize/issues/8176) and [stringify](https://github.com/sequelize/sequelize/issues/8471) issues

```js
[{  
  value: 10,
  inclusive: true
}, {
  value: 10,
  inclusive: true
}]
```

mjy78: No problem. Is it ok if I make these changes under this pull request? There'll be some documentation changes too which I'll try to cover.

sushantdhiman: Yeah no problem

docs/models-definition.md
```diff
@@ -248,13 +241,9 @@ receive:
 
 ```js
 // stored value: ("2016-01-01 00:00:00+00:00", "2016-02-01 00:00:00+00:00"]
-range // [Date, Date]
-range.inclusive // [false, true]
+range // [{ value: Date, inclusive: false }, { value: Date, inclusive: true }]
 ```
 
-Make sure you turn that into a serializable format before serialization since array
```
sushantdhiman: A note here saying that you will need to call `reload` after updating an instance with range type or use `returning: true` option

mjy78: Done.

PR #9371, opened 2018-04-29
feat(query-generator): Generate INSERTs using bind parameters

gazoakley:
### Pull Request check-list

### Description of change

This provides very basic support for using bind parameters when generating INSERT queries and partly addresses #1608. Re-do of #8923

### DIFF
--- a/test/unit/sql/insert.test.js
+++ b/test/unit/sql/insert.test.js
@@ -31,8 +31,40 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
           postgres: 'INSERT INTO "users" ("user_name") VALUES (\'triggertest\') RETURNING *;',
           default: "INSERT INTO `users` (`user_name`) VALUES ('triggertest');"
         });
+
     });
 
+    it('with bind parameters', () => {
+      const User = Support.sequelize.define('user', {
+        username: {
+          type: DataTypes.STRING,
+          field: 'user'
+        },
+        name: {
+          type: DataTypes.STRING,
+          field: 'name'
+        }
+      }, {
+        timestamps: false
+      });
+
+      const values = {
+        user: 'bindtest',
+        name: Support.sequelize.fn('UPPER', 'bindtest')
+      };
+      const options = {
+        bindParams: true
+      };
+      expectsql(sql.insertQuery(User.tableName, values, User.rawAttributes, options),
+        {
+          query: {
+            mssql: 'INSERT INTO [users] ([user],[name]) VALUES ($1,UPPER(N\'bindtest\'));',
+            postgres: 'INSERT INTO "users" ("user","name") VALUES ($1,UPPER(\'bindtest\'));',
+            default: 'INSERT INTO `users` (`user`,`name`) VALUES ($1,UPPER(\'bindtest\'));',
+          },
+          bind: ['bindtest']
+        });
+    });
   });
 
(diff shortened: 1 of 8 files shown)


test/unit/sql/insert.test.js
```diff
@@ -31,8 +31,37 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
...
+      const values = {
+        user: 'bindtest',
+        name: Support.sequelize.fn('UPPER', 'bindtest')
+      };
+      const options = {
+        bindParams: true
+      };
+      expectsql(sql.insertQuery(User.tableName, values, User.rawAttributes, options),
+        {
+          mssql: 'INSERT INTO [users] ([user],[name]) VALUES ($1,UPPER(N\'bindtest\'));',
```
janmeier: Ideally I'd like to assert on bind parameters as well. Something along the lines of 

```
expectsql(sql..., {
 query: {
 ....
  }.
  bind: [...]
})
```

gazoakley:
Thanks @janmeier . I've put in a flag out of cautiousness - it also means anyone enabling the flag will be aware of the limitations of the current implementation. It could be enabled by default in a future version.

test/unit/sql/insert.test.js
gazoakley: Good call - I'll do some tweaks shortly

janmeier:
@sequelize/orm WDYT? Should we hide this behind a flag?

janmeier:
What are the limitations exactly? Of the top of my head I can't come up with any other than functions that might cause problems for inserts and you are already handling that

gazoakley:
Whilst I've named the option/flag `bindParams` it currently only supports bind parameters for non-bulk INSERTs - I've documented this as a limitation here:

https://github.com/sequelize/sequelize/pull/9371/files#diff-4be2078a125fe364f7198675d51bc688R123

```
// Generate queries with bind parameters
// Currently only has an effect for non-bulk INSERTs
// - default: false
bindParams: true
```

Merging this won't add support for bind parameters to everywhere in the query generator, but having this behind a documented option gives users the benefit of bind parameters whilst ensuring they are aware of where it is supported. There's always a possibility to enable this option by default in the future.

I don't think there should be a need to rerun the whole integration test suite twice - there are specific tests that target this option.

janmeier:
I think what Erik is getting at is, that you remove the flag and run the full integration test suite with all its inserts again the new could we would have no problems releasing this without a flag.
If it can run against all our integration tests I'm sure 99% of cases are covered :)

gazoakley:
At a brief glance 67 unit tests need updating to use the new format across the different dialects - I haven't looked at updating the integration tests yet. I'll see if I can get some time to make the changes.

gazoakley:
@janmeier I'd come to the same conclusion. I'm thinking of passing a function as a third param to the `_stringify` methods that returns a `$n` placeholder and updates the bind params array.

janmeier:
> @janmeier I'd come to the same conclusion. I'm thinking of passing a function as a third param to the _stringify methods that returns a $n placeholder and updates the bind params array.

Sounds like a good idea! Ping me when you have something ready for review (or let me know if you need my help)

sushantdhiman:
Closing this PR as work is now continued in https://github.com/sequelize/sequelize/pull/9431

PR #9374, opened 2018-04-30
chore(*): mark aliases as deprecated

### LINKED ISSUE #9372: Suggestion: Remove all aliases in v5
Not sure if I'm too late here but over the years Sequelize has create many many aliases functions. I think it would be a good idea to remove the aliases with the next major release.

- [ ] `find` vs. `findOne`
- [ ] `findById` vs. `findByPrimary`
- [ ] `afterBulkDestroy` vs `afterBulkDelete`
- [ ] `addHook` v. `hook`
- [ ] `VIRTUAL `vs `NONE`

_**Edit**_: Formatting

SimonSchick:
### Pull Request check-list

### Description of change

Closes #9372

This deprecates all(?) aliases including references in some prototypes.

- There were some load time deprecation issues I had to work around, hence the `deprecation.js` file and `deprecate.enable()` call after the sequelize core loaded.
- Amended a `console.warn` to include the stack trace for deprecation warning (depd doesn't seem like a very good library tbh).
- Changed some references of `Util`, `errors` etc. to use static references rather then instance bound.

### DIFF
--- a/lib/associations/base.js
+++ b/lib/associations/base.js
@@ -23,7 +23,7 @@ const AssociationError = require('./../errors').AssociationError;
  * user.getPictures() // gets you all pictures
  * user.getProfilePicture() // gets you only the profile picture
  *
- * User.findAll({
+ * User.findOneAll({
  *   where: ...,
  *   include: [
  *     { model: Picture }, // load all pictures
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -65,8 +65,8 @@ const User = sequelize.define('user', {
   }
 });
 
-// Method 2 via the .hook() method (or its alias .addHook() method)
-User.hook('beforeValidate', (user, options) => {
+// Method 2 via the .addHook() method
+User.addHook('beforeValidate', (user, options) => {
   user.mood = 'happy';
 });
 
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -217,7 +217,7 @@ sequelize.transaction({
 ## Usage with other sequelize methods
 
 The `transaction` option goes with most other options, which are usually the first argument of a method.
-For methods that take values, like `.create`, `.update()`, `.updateAttributes()` etc. `transaction` should be passed to the option in the second argument.
+For methods that take values, like `.create`, `.update()`, etc. `transaction` should be passed to the option in the second argument.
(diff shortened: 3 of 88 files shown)


SimonSchick:
I will have to fix some issues with excessive logging, you are free to review this as is tho.

SimonSchick:
I think I got it now, having some trouble running all tests on my local machine.

Turns out VSCode doesn't by default index/search the test folder, had to make some adjustments here and there.

Also had to move the logger set out of the class because it seems to be instantiated multiple times?

SimonSchick:
Also there are is one issue where Sequelize is iterating over the data types and will therefor trigger a deprecation warning @sushantdhiman any ideas?

SimonSchick:
Not sure what's up with that one failing test, something about queries not completing :/

sushantdhiman:
This change is too big, there is no need to deprecate everything we are going to remove, only critical and commonly used cases should be deprecated.

First step should be to list things we are trying to remove https://github.com/sequelize/sequelize/issues/9372#issue-318736625 , then we can decide what to keep and what not, then move forward :)

SimonSchick:
> only critical and commonly used cases

This seems counter intuitive as deprecating the niche use cases has much lower deprecation impact then deprecating the `critical` ones.

I will leave it up to you to come up with a list you deem appropriate to deprecate, my aim's to cover all/most of them.

lib/associations/base.js
```diff
@@ -23,7 +23,7 @@ const AssociationError = require('./../errors').AssociationError;
  * user.getPictures() // gets you all pictures
  * user.getProfilePicture() // gets you only the profile picture
  *
- * User.findAll({
+ * User.findOneAll({
```
yacinehmito: Is that intended ?

SimonSchick: No that must be a regex gone haywire, no bueno, will fix once @sushantdhiman decided which of my changes are applicable.

sushantdhiman: Please don't wait up for me, I don't think I will have time to review 80+ files. As I suggested earlier it will be better to propose list of duplicate / removable targets, then we can discuss which one we need to actually deprecate / remove

SimonSchick:
Unfortunately I don't have time to pursue this for now.

Issue #5193, opened 2016-01-12
include on hasMany association only returning one of many results when no primary key


mholtzman:
I looked around for this issue and couldn't find it mentioned exactly as I'm seeing it, so apologies in advance if this is a duplicate. The issue is that when I use include like so:

`Project.findAll({ include: [{ model: Task }], where: { ... } })`

the query is generated correctly and the correct projects are returned, but each contains an array of Tasks that is always of length one (when the Tasks have no primary key), even though many tasks match. I can confirm this using the raw query and verifying the results. If I add a primary key to Task, it then works correctly.

Here are the related definitions, in case there are some clues there.

```
sequelize.define('Project', {
       ...
    }, {
        timestamps: true,
        underscored: true,
        createdAt: 'created_at',
        updatedAt: 'last_modified',
        tableName: 'projects',
        classMethods: {
            associate: function(models) {
                Project.hasMany(models.Task, { foreignKey: 'projectId', onDelete: 'CASCADE' });
            }
        },

sequelize.define('Task', {
        ...
    }, {
        timestamps: true,
        underscored: true,
        createdAt: 'created_at',
        updatedAt: false,
        tableName: 'tasks'
    });
```

mickhansen:
When we process associations we attempt to dedupe each row by either matching primary keys or row uniqueness.

Any chance the task rows are completely equal?

mholtzman:
Looks like this is occurring when a field in the table is named 'id'. Here's a test case:

```
const Sequelize = require('sequelize');

const connection = new Sequelize('', '', '', { dialect: 'sqlite' });

const Task = connection.define('Task', {
    taskType: { type: Sequelize.INTEGER, field: 'id' },
    taskName: { type: Sequelize.STRING(50) }
});

const Project = connection.define('Project', {
    id: { type: Sequelize.INTEGER, primaryKey: true }
});

Project.hasMany(Task);

connection.sync()
    .then(() => Project.create({ id: 1 }))
    .then(() => Task.create({ taskType: 10, taskName: 'get food', ProjectId: 1 }))
    .then(() => Task.create({ taskType: 10, taskName: 'take out trash', ProjectId: 1 }))
    .then(() => Project.findAll({ include: [Task] }))
    .then(tasks => console.log(JSON.stringify(tasks, undefined, 2)));
```

If you remove the `field: 'id'` clause, this works as expected. Is this just a mapping error on my part?

mickhansen:
[...]

PR #9384, opened 2018-05-03
WIP: Disable checkExisting to allow _duplicated_ rows (#5193)

pateketrueke:
### Pull Request check-list

### Description of change

- It potentially adds a `nested` option for skipping `checkExisting` to allow duplicate rows under certain conditions, hence the added option.
- Because this, it'll accumulate found children from `includeNames` properties.

With this, calling `findOne()` will return the same as `parent.getChildren()` does.

> I would like to avoid adding a new option, but I don't know how differentiate this kind of result without guessing too much (e.g. how `parent.getChildren()` options works?). If you know a better way to avoid all of this please let me know.


pateketrueke:
Any feedback is very welcome on this: @uncedric @trostinsky @holmberd @saigalamatya @evenfrost 🍻

sushantdhiman:
What is the motivation behind this change, can you show me the issue that you are trying to fix, or something with code.

pateketrueke:
Hi  @sushantdhiman, [I added a test](https://github.com/sequelize/sequelize/pull/9384/files#diff-04c30a7f7a4e8f01a8992bf49c67c255R564) which describes the issue I'm falling in:

- I have two models: Cart, Product
- I have CartItem model which associates them
- CartItem has a PK and its associations are not unique

Before my changes I ran the query and just one row is retrieved, I copied the generated SQL and another client gave me two rows as I expected.

After adding a some `console.log` to see what is going wrong I noticed two things:

- When [checkExisting](https://github.com/sequelize/sequelize/pull/9384/files#diff-85aaffe60f984b68bb500de75aabffe9L289) is disabled I was able to retrieve both rows as `result`.
- But now, I was not able to retrieve them through the promise. Curiously adding `raw: true` was returning both rows, but as flat keypaths. Not as _nested_ objects.

My solution was adding the block below for grouping by PK (e.g. `Cart.id`) and the test proves exactly what I wanted:

- `foundCartInstance.items` should has both rows
- `anyCartInstance.getItems()` should resolve both rows

I don't expect to be coerced to use two queries to retrieve the cart items.

weyert:
This sounds like an interesting PR :)

ben-turner:
Just for reference, I think this is a fix for #5193

fraywing:
Let's get this in!

fraywing:
@pateketrueke I think the build failed? @sushantdhiman does this look good? Experiencing the same issue in a ManyToMany `findAndCountAll` call where the includes only return a single record.

pateketrueke:
Yeah, it's failing but idk why (not sure if `checkExisting` has conflict with mysql/postgres; sqlite/mariadb are OK), also I can't re-run the tests or ssh for debug...

I'll try to run CI with some changes to understand the reasons of the failures.

pateketrueke:
Weird, now travis is OK but appveyor is not... 🦀

papb:
@pateketrueke Thanks for this. Don't worry about the failing test.

monti-tls:
Any news on this PR ? I am encountering an issue related to this, when doing a `findAll`, with `include: { all: true, nested: true }` on a model with a `belongsToMany` relation that should output duplicate rows.

PR #9431, opened 2018-05-13
feat(query-generator): Generate INSERTs using bind parameters

gazoakley:
### Pull Request check-list

### Description of change

This is a new spin of #9371 where I've removed the `bindParams` option and updated all the INSERT unit tests to match the new syntax. It's thrown up a few issues:

* Data types that don't easily match up with built in JS data types (specifically `GEOMETRY`) cause issues since drivers can't identify them before sending them as bind parameters. They need to be sent using a correct CAST and providing the bind parameter in string form e.g. `GeomFromText($1)` with a valid bind param
* Upsert needs updating - both PostgreSQL and Sqlite override the upsertQuery function in the query generator

### DIFF
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -33,6 +33,12 @@ ABSTRACT.prototype.stringify = function stringify(value, options) {
   }
   return value;
 };
+ABSTRACT.prototype.bindParam = function bindParam(value, options) {
+  if (this._bindParam) {
+    return this._bindParam(value, options);
+  }
+  return options.bindParam(this.stringify(value, options));
+};
 
 function STRING(length, binary) {
   const options = typeof length === 'object' && length || {length, binary};
@@ -274,15 +280,21 @@ DECIMAL.prototype.validate = function validate(value) {
 
 for (const floating of [FLOAT, DOUBLE, REAL]) {
   floating.prototype.escape = false;
-  floating.prototype._stringify = function _stringify(value) {
+  floating.prototype._value = function _value(value) {
     if (isNaN(value)) {
-      return "'NaN'";
+      return 'NaN';
     } else if (!isFinite(value)) {
       const sign = value < 0 ? '-' : '';
-      return "'" + sign + "Infinity'";
+      return sign + 'Infinity';
     }
 
     return value;
+  };
+  floating.prototype._stringify = function _stringify(value) {
+    return "'" + this._value(value) + "'";
+  };
+  floating.prototype._bindParam = function _bindParam(value, options) {
+    return options.bindParam(this._value(value));
   };
 }
 
@@ -529,6 +541,17 @@ BLOB.prototype._hexify = function _hexify(hex) {
   return "X'" + hex + "'";
 };
 
+BLOB.prototype._bindParam = function _bindParam(value, options) {
+  if (!Buffer.isBuffer(value)) {
+    if (Array.isArray(value)) {
+      value = Buffer.from(value);
+    } else {
+      value = Buffer.from(value.toString());
+    }
+  }
+  return options.bindParam(value);
+};
+
 function RANGE(subtype) {
   const options = _.isPlainObject(subtype) ? subtype : {subtype};
 
@@ -703,6 +726,9 @@ GEOMETRY.prototype.escape = false;
 GEOMETRY.prototype._stringify = function _stringify(value, options) {
   return 'GeomFromText(' + options.escape(Wkt.convert(value)) + ')';
 };
+GEOMETRY.prototype._bindParam = function _bindParam(value, options) {
+  return 'GeomFromText(' + options.bindParam(Wkt.convert(value)) + ')';
+};
 
 function GEOGRAPHY(type, srid) {
   const options = _.isPlainObject(type) ? type : {type, srid};
@@ -720,6 +746,9 @@ GEOGRAPHY.prototype.key = GEOGRAPHY.key = 'GEOGRAPHY';
 GEOGRAPHY.prototype.escape = false;
 GEOGRAPHY.prototype._stringify = function _stringify(value, options) {
   return 'GeomFromText(' + options.escape(Wkt.convert(value)) + ')';
+};
+GEOGRAPHY.prototype._bindParam = function _bindParam(value, options) {
+  return 'GeomFromText(' + options.bindParam(Wkt.convert(value)) + ')';
 };
 
 for (const helper of Object.keys(helpers)) {
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -106,6 +106,8 @@ class QueryGenerator {
     const modelAttributeMap = {};
     const fields = [];
     const values = [];
+    const bind = [];
+    const bindParam = this.bindParam(bind);
     let query;
     let valueQuery = '<%= tmpTable %>INSERT<%= ignoreDuplicates %> INTO <%= table %> (<%= attributes %>)<%= output %> VALUES (<%= values %>)';
     let emptyQuery = '<%= tmpTable %>INSERT<%= ignoreDuplicates %> INTO <%= table %><%= output %>';
@@ -169,7 +171,14 @@ class QueryGenerator {
       }
     }
 
+    if (_.get(this, ['sequelize', 'options', 'dialectOptions', 'prependSearchPath']) || options.searchPath) {
+      // Not currently supported with search path (requires output of multiple queries)
+      options.bindParam = false;
+    }
+
     if (this._dialect.supports.EXCEPTION && options.exception) {
+      // Not currently supported with bind parameters (requires output of multiple queries)
+      options.bindParam = false;
       // Mostly for internal use, so we expect the user to know what he's doing!
       // pg_temp functions are private per connection, so we never risk this function interfering with another one.
       if (semver.gte(this.sequelize.options.databaseVersion, '9.2.0')) {
@@ -211,7 +220,11 @@ class QueryGenerator {
             identityWrapperRequired = true;
           }
 
-          values.push(this.escape(value, modelAttributeMap && modelAttributeMap[key] || undefined, { context: 'INSERT' }));
+          if (value instanceof Utils.SequelizeMethod || options.bindParam === false) {
+            values.push(this.escape(value, modelAttributeMap && modelAttributeMap[key] || undefined, { context: 'INSERT' }));
+          } else {
+            values.push(this.format(value, modelAttributeMap && modelAttributeMap[key] || undefined, { context: 'INSERT' }, bindParam));
+          }
         }
       }
     }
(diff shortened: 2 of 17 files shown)


lib/data-types.js
```diff
@@ -725,6 +739,9 @@ GEOGRAPHY.prototype.escape = false;
 GEOGRAPHY.prototype._stringify = function _stringify(value, options) {
   return 'GeomFromText(' + options.escape(Wkt.convert(value)) + ')';
 };
+GEOMETRY.prototype.bindParam = function bindParam(value, options) {
```
sushantdhiman: `_bindParam` to match `_stringify` etc stuff

lib/dialects/abstract/query-generator.js
```diff
@@ -989,11 +989,18 @@ class QueryGenerator {
     return SqlString.escape(value, this.options.timezone, this.dialect);
   }
 
+  bindParam(bind) {
```
sushantdhiman: :+1:

test/integration/data-types.test.js
```diff
@@ -56,14 +56,18 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
     });
   });
 
-  const testSuccess = function(Type, value) {
+  const testSuccess = function(Type, value, useBindParam) {
```
sushantdhiman: Third argument should be options with arbitrary configuration possibilities ``{ useBindParam }``

gazoakley:
@sushantdhiman Hit a bit of a snag doing upserts with Postgres. The current `upsertQuery` function generates SQL along the lines of:

```sql
CREATE OR REPLACE FUNCTION pg_temp.sequelize_upsert(OUT created boolean,OUT primary_key text)  AS $func$
BEGIN
INSERT INTO "users" ("id","username","zab","createdAt","updatedAt") VALUES (1,"john","foo","1970-01-01 00:00:00.000 +00:00","1970-01-01 00:00:00.000 +00:00"); created := true;
EXCEPTION WHEN unique_violation THEN
UPDATE "users" SET "id"=42,"username"='john',"updatedAt"='1970-01-01 00:00:00.000 +00:00' WHERE ("id" = 42); created := false;
END;
$func$ LANGUAGE plpgsql;
SELECT * FROM pg_temp.sequelize_upsert();
```

That's fine when not using bind parameters/prepared statements, but when I generate something along the lines of:

```sql
CREATE OR REPLACE FUNCTION pg_temp.sequelize_upsert(arg1 <argtype>,arg2 <argtype>,arg3 <argtype>,arg4 <argtype>,arg5 <argtype>,OUT created boolean,OUT primary_key text)  AS $func$
BEGIN
INSERT INTO "users" ("id","username","zab","createdAt","updatedAt") VALUES (arg1,arg2,arg3,arg4,arg5); created := true;
EXCEPTION WHEN unique_violation THEN
UPDATE "users" SET "id"=42,"username"='john',"updatedAt"='1970-01-01 00:00:00.000 +00:00' WHERE ("id" = 42); created := false;
END;
$func$ LANGUAGE plpgsql;
SELECT * FROM pg_temp.sequelize_upsert($1,$2,$3,$4,$5);
```

I get an error back from Postgres of `cannot insert multiple commands into a prepared statement` - I can't string together multiple statements if using prepared statements. I think the result of `upsertQuery` for Postgres is going to have to return multiple statements to accomplish what its trying to do. I'm thinking of having it return along the lines of:

```json
[
    {
        "query": "CREATE OR REPLACE FUNCTION pg_temp.sequelize_upsert(arg1 <argtype>,arg2 <argtype>,arg3 <argtype>,arg4 <argtype>,arg5 <argtype>,OUT created boolean,OUT primary_key text)  AS $$func$$ BEGIN INSERT INTO \"users\" (\"id\",\"username\",\"zab\",\"createdAt\",\"updatedAt\") VALUES (arg1,arg2,arg3,arg4,arg5); created := true; EXCEPTION WHEN unique_violation THEN UPDATE \"users\" SET \"id\"=42,\"username\"='john',\"updatedAt\"='1970-01-01 00:00:00.000 +00:00' WHERE (\"id\" = 42); created := false; END; $$func$$ LANGUAGE plpgsql;",
        "bind": []
    },
    {
        "query": "SELECT * FROM pg_temp.sequelize_upsert($1,$2,$3,$4,$5);",
        "bind": [42, "john", "foo", "1970-01-01 00:00:00.000 +00:00", "1970-01-01 00:00:00.000 +00:00"]
    }
]
```

And have something further up treat an array of query elements as something that must be run as a sequence of statements that must be run one after each other. Does that seem reasonable to you?

sushantdhiman:
You are right, we can't use parameterized query with multiple statements. It seems we need to execute two queries with correct order, also keeping other `upsert` from affecting each other.

> And have something further up treat an array of query elements as something that must be run as a sequence of statements that must be run one after each other. Does that seem reasonable to you?

Yeah I agree, we need to return multiple statements with their parameters.

In old approach we used to send one single query which will create / replace function and execute it at server level, but now we need to wrap `Model.upsert` in transaction. As two different statements are going to be executed, we dont want one upsert to be replacing other upsert function and return inconsistent results.

Another potential solution is to support `ON CONFLICT` but we still need to support this for `Postgres < 9.5` so multiple statements solution still needs to be implemented.

So overall, we need to do this

1. Return array of statements from query generator
2. Execute them in order
3. Wrap `upsert` in transaction, so there is no cross talk between other `upsert` trying to replace same function

gazoakley:
The tests are now passing on everything apart from Postgres which is becoming increasingly complex. There's still some fixes needed to support RANGE correctly with bind params, and I've also spotted the Postgres specific code that is returned by the abstract insertQuery which needs the same treatment.

This PR is starting to get quite large too. So that I can split this work up better, would it be acceptable to have an option to turn off bindParams? This would let me:

- Turn it off by default if insertQuery is passed options.exception as true
- Turn it off when calling insertQuery from Postgres upsertQuery

I can then postpone having to deal with returning multiple queries and the further complexity that entails. There's also a good opportunity at that point to remove the duplicate code in insertQuery so that it calls fn/exceptionFn in postgres/query-generator.js

janmeier:
Sounds good tome :)

sushantdhiman:
Sounds good if we keep it as a private option, Use it internally so we can keep converting old queries to parameter based queries and then remove it eventually

gazoakley:
@janmeier @sushantdhiman Any chance either of you could try running the tests yourselves? I can't reproduce the failures in Travis 😕

janmeier:
You need to rebase on top of latest master (specifically this commit https://github.com/sequelize/sequelize/commit/b66f9ddc36720aef1523d677dbbfb9f257afa472 has some tests that are now failing) 

I was tipped up by the fact that my local tests said 1764 passing, while travis was 1764 passing, 4 failing so I figured my local copy was missing something

gazoakley:
@janmeier Ooops! I'll try and sort it over lunch.

gazoakley:
Tests are all green 🎉

test/unit/dialects/sqlite/query-generator.test.js
```diff
@@ -364,50 +364,92 @@ if (dialect === 'sqlite') {
...
+            query: 'INSERT INTO `myTable` (`name`,`birthday`) VALUES ($1,$2);',
+            bind: ['foo', moment('2011-03-27 10:01:55 +0000', 'YYYY-MM-DD HH:mm:ss Z').toDate()]
+          }
         }, {
           arguments: ['myTable', { name: 'foo', value: true }],
-          expectation: "INSERT INTO `myTable` (`name`,`value`) VALUES ('foo',1);"
+          expectation: {
+            query: 'INSERT INTO `myTable` (`name`,`value`) VALUES ($1,$2);',
+            // TODO: Check
+            bind: ['foo', true]
```
janmeier: Should be fine (I'm pretty sure sqlite takes whatever you throw at it)

test/unit/dialects/sqlite/query-generator.test.js
```diff
@@ -364,50 +364,92 @@ if (dialect === 'sqlite') {
...
+          expectation: {
+            query: 'INSERT INTO `myTable` (`name`,`value`) VALUES ($1,$2);',
+            bind: ['bar', null]
+          }
         }, {
           arguments: ['myTable', { name: 'bar', value: undefined }],
-          expectation: "INSERT INTO `myTable` (`name`,`value`) VALUES ('bar',NULL);"
+          expectation: {
+            query: 'INSERT INTO `myTable` (`name`,`value`) VALUES ($1,$2);',
+            // TODO: Check
```
janmeier: Hmm, not sure - There is another issue about ignoring undefined completely, but it should be fine for now

lib/dialects/abstract/query-generator.js
```diff
@@ -234,7 +247,12 @@ class QueryGenerator {
       ].join(' ');
     }
 
-    return _.template(query, this._templateSettings)(replacements);
+    query = _.template(query, this._templateSettings)(replacements);
+    // Used by Postgres upsertQuery and calls to here with options.exception set to true
+    if (options.bindParam === false) {
+      return query;
```
sushantdhiman: return an object in both cases, so all calls can just use `.query`. In case `options.bindParam: false` return no bind parameters.

lib/dialects/abstract/query-generator.js
```diff
@@ -169,7 +171,14 @@ class QueryGenerator {
       }
     }
 
+    if (_.get(this, ['sequelize', 'options', 'dialectOptions', 'prependSearchPath']) || options.searchPath) {
```
sushantdhiman: Use full path `sequelize.options.dialectOptions.prependSearchPath` in `_.get`

test/integration/dialects/postgres/dao.test.js
```diff
@@ -597,12 +597,13 @@ if (dialect.match(/^postgres/)) {
...
 
-      it('should use postgres "TIMESTAMP WITH TIME ZONE" instead of "DATETIME"', function() {
+      it('should use bind params instead of "TIMESTAMP WITH TIME ZONE"', function() {
         return this.User.create({
           dates: []
         }, {
           logging(sql) {
-            expect(sql.indexOf('TIMESTAMP WITH TIME ZONE')).to.be.greaterThan(0);
+            expect(sql).not.to.contain('TIMESTAMP WITH TIME ZONE');
+            expect(sql).not.to.contain('DATETIME');
```
sushantdhiman: assert presence of bind param?

lib/dialects/abstract/query-generator.js
gazoakley: Done

lib/dialects/abstract/query-generator.js
gazoakley: I can change it, but it's more efficient in Lodash to use the array path form. If you use the string path form (e.g. `_.get(this, 'sequelize.options.dialectOptions.prependSearchPath')`) Lodash first converts it back to the array path form which currently entails a bunch of regex operations to parse the path. For a single call like this it probably doesn't matter, but do it many times and there's a big difference in performance.

sushantdhiman: @gazoakley No need to change it then :)

gazoakley:
Now we just need to make `UPDATE`, `SELECT` and `DELETE` use bind params 🤣

kirrg001:
@gazoakley Hi there! Is there any progress on supporting bind params for select, update and delete? Thanks a lot!

PR #9441, opened 2018-05-15
Fixed empty blob

YaroslavRepeta:
### Pull Request check-list

### Description of change

SELECT query returns null instead of empty Buffer for BLOB type

### DIFF
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -43,13 +43,7 @@ module.exports = BaseTypes => {
   inherits(BLOB, BaseTypes.BLOB);
 
   BLOB.parse = function(value, options, next) {
-    const data = next();
-
-    if (Buffer.isBuffer(data) && data.length === 0) {
-      return null;
-    }
-
-    return data;
+    return next();
   };
 
   function DECIMAL(precision, scale) {
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -36,16 +36,6 @@ module.exports = BaseTypes => {
   BaseTypes.GEOMETRY.types.mysql = ['GEOMETRY'];
   BaseTypes.JSON.types.mysql = ['JSON'];
 
-  function BLOB(length) {
-    if (!(this instanceof BLOB)) return new BLOB(length);
-    BaseTypes.BLOB.apply(this, arguments);
-  }
-  inherits(BLOB, BaseTypes.BLOB);
-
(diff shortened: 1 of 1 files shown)


lib/dialects/mysql/data-types.js
```diff
@@ -43,13 +43,7 @@ module.exports = BaseTypes => {
   inherits(BLOB, BaseTypes.BLOB);
 
   BLOB.parse = function(value, options, next) {
-    const data = next();
```
sushantdhiman: I dont think so this works, can you show me a failing case

YaroslavRepeta: I'm not sure where the right place to add test case for this.

What I did here is only removed
https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/data-types.js#L48-L50
what causes returning null instead of empty buffer.

sushantdhiman: Ok show me example of what you are trying to solve, paste it here

YaroslavRepeta: Something like:
expect((await model.findOne({ some_where_clause })).empty_buffer_column).to.equal(Buffer.from(""));

In fact `mysql2` what is used returns empty buffer in select queries, when `sequelize` catches that and returns null instead.

sushantdhiman:
Ok so I have tested this change

```js
User
  .sync({ force: true })
  .then(() => {
    return User.bulkCreate([{
      username: 'anc',
      data: Buffer.from('')
    }, {
      username: 'anc',
      data: null
    }]);
  })
  .then(() => {
    return User.findAll();
  })
  .then(users => {
    console.log(users[0].data, users[1].data);
  });

### Output
<Buffer > null
```

It seems MySQL2 now handles null case correctly, so as expected binary `null` should be returned as null but empty buffer should be returned as buffer.

If you can add a test around here https://github.com/sequelize/sequelize/blob/master/test/integration/model.test.js#L2526

Some related issues https://github.com/sequelize/sequelize/pull/8661  https://github.com/sidorares/node-mysql2/issues/668

lib/dialects/mysql/data-types.js
```diff
@@ -43,13 +43,7 @@ module.exports = BaseTypes => {
   inherits(BLOB, BaseTypes.BLOB);
 
   BLOB.parse = function(value, options, next) {
```
sushantdhiman: BLOB function from this file, we will now go with mysql2 default

Issue #9376, opened 2018-05-01
Tests can fail when running with docker on Mac/Windows


mjy78:
## What are you doing?

Running integration tests using docker on a Windows or MacOS host can sometimes fail with date assertions.

An example of one such test is in ~/test/integration/postgres/data-types.test.js around line 159 `it("should be able to create and update records with Infinity/-Infinity")`, in which case the assertion `expect(new Date(user.sometime)).to.be.withinDate(date, new Date());` can sometimes fail due to the date that is returned from the db being a few hundred ms ahead of the expected date range.

This is only one example test case, there are a number of others where I've seen similar behaviour.

## What do you expect to happen?

These test cases should pass. The `date` (first param of the withinDate() assertion) is initialised before the DB insert operation, and the `new Date()`  is called after the insert. The date initialised in the DB (from calling NOW) should fall within these two values.

## What is actually happening?

These test cases fail due to time drift that occurs between the host (executing the test code) and the VM running the db (postgres/mysql) containers. 

Some test cases initialise dates in the database with `this.sequelize.fn('NOW')` when generating test rows. Due to time drift, the time in the database container can be a fraction (even just a few hundred ms) ahead of the host. This can cause some test assertions to fail where the assertion is expecting the result from the db to fall within dates created before and after the db insert statement.

__Dialect:__ postgres
__Database version:__ 9.5
__Sequelize version:__ 4.37.6
__Tested with latest release:__ Yes

The root cause of this issue is the time drift that can occur between host and docker containers on MacOS and Windows, however I believe we can avoid this issue all together if we change our documentation and test process to execute the tests inside a container rather than on the host. This way both the test execution and db containers will run using the same system time (of the docker VM).

As an example, if I run the following, the integration tests complete successfully on MacOS...

```
docker-compose run --rm -e DIALECT=postgres -e SEQ_HOST=postgres-95 sequelize npm run test-integration
```

I'm thinking a clean approach might be if the docker configuration and npm "docker" scripts were handled internally within the "sequelize" docker service. We could get to the point where contributors only need to install docker on their host (regardless of it being Windows/MacOS/Linux) and run something like:

```
docker-compose run --rm sequelize npm run test-docker
```

Thoughts?

papb:
Hello, is this still an issue with latest sequelize and docker for windows?

PR #9455, opened 2018-05-19
fix: Use node 8 on docker image

caiquecastro:
The package @semantic-release/github, that is a direct dependency for the project requires node >=8.3.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:6
+FROM node:8
 
 RUN apt-get install libpq-dev
 


sushantdhiman:
Which package requires 8.3 > ? (Apart from semantic release which is only used in CI)

caiquecastro:
It's the semantic release package. If I don't have the libraries required to run the project on my machine, I cannot run yarn install on the docker container running the node 6 version image. Only if I run with --ignore-engines. The PR was a suggestion to make the things easier for collaborators. I also think it's better to use a Active LTS version instead of a Maintenance LTS.

sushantdhiman:
> It's the semantic release package. If I don't have the libraries required to run the project on my machine, I cannot run yarn install on the docker container running the node 6 version image.

Contribution docs no where use this docker file, it was used to build service in past but not now. Until we are actually using it somewhere, there is no point changing it.

> I also think it's better to use a Active LTS version instead of a Maintenance LTS.

We can't just jump skip past Node versions for reasons like support by cloud services, usages patterns etc. I think a big group still use Node 6 that's why with v5 we have decided to go with Node 6.

So unless we start using this service to run tests (see https://github.com/sequelize/sequelize/issues/9376), I see no point changing it. Even after that I think we should keep this image on minimum supported Node version.

> Only if I run with --ignore-engines.

What is yarn's behaviour for this, it refuses to install all dependencies or just skip that one? If there is no hard error I guess everything else should work

caiquecastro:
Without the  --ignore-engines, it doesnt install any dependency with yarn. But I got your reasons. We can close it if you want.

PR #9458, opened 2018-05-20
fix: create nested with belongs not allowing null

caiquecastro:
### Pull Request check-list

This PR contains only the breaking test.


sushantdhiman:
Moved #9466

Issue #4708, opened 2015-10-22
[bug?] Creating with associations


alekbarszczewski:
I want to create Company instance along with associated Address:

``` js
var Sequelize = require('sequelize')

var sequelize = new Sequelize('postgres://localhost/db', { logging: console.log })

var Company = sequelize.define('Company', {
  AddressId: {
    type: Sequelize.INTEGER,
    allowNull: false,
    unique: true
  }
})

var Address = sequelize.define('Address', {
  street: Sequelize.TEXT
})

Company.belongsTo(Address)

sequelize.sync({ force: true })
.then(function () {
  return Company.create({
    Address: { street: 'street1' }
  }, {
    include: [
      Address
    ]
  })
})
```

Output:

```
Unhandled rejection SequelizeValidationError: notNull Violation: AddressId cannot be null
    at /Users/alek/Desktop/pros/node_modules/sequelize/lib/instance-validator.js:72:14
    at tryCatcher (/Users/alek/Desktop/pros/node_modules/sequelize/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/Users/alek/Desktop/pros/node_modules/sequelize/node_modules/bluebird/js/main/promise.js:507:31)
    at Promise._settlePromiseAt (/Users/alek/Desktop/pros/node_modules/sequelize/node_modules/bluebird/js/main/promise.js:581:18)
    at Promise._settlePromiseAtPostResolution (/Users/alek/Desktop/pros/node_modules/sequelize/node_modules/bluebird/js/main/promise.js:245:10)
    at Async._drainQueue (/Users/alek/Desktop/pros/node_modules/sequelize/node_modules/bluebird/js/main/async.js:128:12)
    at Async._drainQueues (/Users/alek/Desktop/pros/node_modules/sequelize/node_modules/bluebird/js/main/async.js:133:10)
    at Async.drainQueues (/Users/alek/Desktop/pros/node_modules/sequelize/node_modules/bluebird/js/main/async.js:15:14)
    at process._tickCallback (node.js:442:13)
```

Quick workaround:

``` js
var Sequelize = require('sequelize')

var sequelize = new Sequelize('postgres://localhost/db', { logging: console.log })

var Company = sequelize.define('Company', {
  AddressId: {
    type: Sequelize.INTEGER,
    allowNull: false,
    unique: true
  }
})

var Address = sequelize.define('Address', {
  street: Sequelize.TEXT
})

[...]

PR #9466, opened 2018-05-22
Fix: Allow foreign key to be null when association reference is set

sbonami:
### Pull Request check-list


sbonami:
@sushantdhiman @eseliger I'm not sure where the failing test is from and believe it's unrelated. How should I proceed?

lib/associations/mixin.js
```diff
@@ -68,6 +68,10 @@ const Mixin = {
     return this.getAssociations(target).find(association => this.verifyAssociationAlias(association, alias)) || null;
   },
 
+  getAssociationForForeignKey(foreignKey) {
+    return _.values(this.associations).find(association => association.foreignKey === foreignKey);
```
sushantdhiman: No need to introduce this method to mixin, filter associations in `instance-validator`

test/integration/model/create/include.test.js
```diff
@@ -60,6 +60,61 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+          return Product.create({
+            title: 'Chair',
+            User: {
+              first_name: 'Mick',
+              last_name: 'Broadstone'
+            }
+          }, {
+            include: [{
+              model: User,
+              myOption: 'option'
```
sushantdhiman: Why you are testing this?

caiquecastro: It's possible to simplify this test.

sushantdhiman: Please do if you can :)

caiquecastro:
Why does this test fails on Windows?

lib/instance-validator.js
```diff
@@ -320,17 +320,20 @@ class InstanceValidator {
...
-
-      this.errors.push(new sequelizeError.ValidationErrorItem(
-        errMsg,
-        'notNull Violation', // sequelizeError.ValidationErrorItem.Origins.CORE,
-        field,
-        value,
-        this.modelInstance,
-        'is_null'
-      ));
+      const association = _.values(rawAttribute.Model.associations).find(association => association.foreignKey === rawAttribute.fieldName);
```
sushantdhiman: You will need to filter/check if association is of `belongsTo` type

sushantdhiman:
This will also fix https://github.com/sequelize/sequelize/issues/4708

lib/instance-validator.js
sbonami: I think it's a given at this point in the call stack, right?

- This is called with each attr from the `modelInstance`'s attributes list, so we know it's a defined attribute (e.g. an attr or a foreign key)
- It's wrapped into the association query: If this wasn't a belongs-to, we won't find a related association with a matching foreign key (given another model's foreign key is a `targetKey`)
- This is a "private" function, so we shouldn't need to worry about adverse calls

I _think_ my logic is correct here, but I don't mind either admitting I've missed something or adding some bolt-tightening logic. I figured I'd suggest saving a cycle if it's pertinent.

lib/instance-validator.js
```diff
@@ -320,17 +320,20 @@ class InstanceValidator {
...
+      if (!association || !this.modelInstance.get(association.associationAccessor)) {
```
sbonami: Also, do we need to validate this works for belongsToMany? I'm not sure if those appear in the `modelInstance`'s attribute list for a null check.

lib/instance-validator.js
sushantdhiman: You can save filter cycle and check for belongsTo directly in `.find`, but I think `BelongsTo` check is important here. It should be as simple as `instanceof BelongsTo` check

sbonami:
@sushantdhiman Tweaked for BelongsTo as requested and rebased to clean-up our commits.

lib/instance-validator.js
```diff
@@ -320,17 +321,20 @@ class InstanceValidator {
...
-
-      this.errors.push(new sequelizeError.ValidationErrorItem(
-        errMsg,
-        'notNull Violation', // sequelizeError.ValidationErrorItem.Origins.CORE,
-        field,
-        value,
-        this.modelInstance,
-        'is_null'
-      ));
+      const association = _.values(rawAttribute.Model.associations).find(association => association instanceof BelongsTo && association.foreignKey === rawAttribute.fieldName);
```
sushantdhiman: `rawAttribute.Model.associations`  => `this.modelInstance.constructor.associations`, just this last change :)

sbonami:
@sushantdhiman Any chance there will be a patch release of this guy soon-ish? 😄

sushantdhiman:
@sbonami Next weekend of 16th - 17th or before, as `5.0.0-beta.8`

PR #9492, opened 2018-05-30
feat(query-generator): Generate UPDATEs using bind parameters

gazoakley:
### Pull Request check-list

### Description of change

WIP changes to allow UPDATEs to use bind parameters instead of directly included values. For example, instead of:

```UPDATE `myTable` SET `name`='foo',`birthday`='2011-03-27 10:01:55' WHERE `id` = 2```

Generate:

```UPDATE `myTable` SET `name`=$1,`birthday`=$2 WHERE `id` = $3```

(with parameters `'foo'`, `'2011-03-27 10:01:55'` and `2`)

### DIFF
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -175,9 +175,15 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
     return testSuccess(Type, 'foobar');
   });
 
-  it('calls parse and stringify for STRING', () => {
+  it('calls parse and stringify/bindParam for STRING', () => {
     const Type = new Sequelize.STRING();
 
+    // mssql has a _bindParam function that checks if STRING was created with
+    // the boolean param (if so it outputs a Buffer bind param). This override
+    // isn't needed for other dialects
+    if (dialect === 'mssql') {
+      return testSuccess(Type, 'foobar',  { useBindParam: true });
+    }
     return testSuccess(Type, 'foobar');
   });
 
--- a/test/integration/model/create.test.js
+++ b/test/integration/model/create.test.js
@@ -940,59 +940,49 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       }
     });
 
-    it('does not cast arrays for postgresql insert', function() {
-      if (dialect !== 'postgres') {
-        expect('').to.equal('');
-        return void 0;
-      }
-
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: test/integration/data-types.test.js [hunk 1]
```diff
@@ -175,9 +175,12 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
     return testSuccess(Type, 'foobar');
   });
 
-  it('calls parse and stringify for STRING', () => {
+  it('calls parse and stringify/bindParam for STRING', () => {
     const Type = new Sequelize.STRING();
 
+    if (dialect === 'mssql') {
```
sushantdhiman: Why only mssql?

test/integration/model/create.test.js
```diff
@@ -984,8 +984,8 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           return user.save({
             logging(sql) {
               test = true;
-              expect(sql).to.contain('ARRAY[]::INTEGER[]');
-              expect(sql).to.contain('ARRAY[]::VARCHAR(255)[]');
+              expect(sql).not.to.contain('ARRAY[]::INTEGER[]');
```
sushantdhiman: These tests should be moved to unit tests, but that is too much work, should be tackled in different PR. Please add some assertion which covers what is expected here, like may be just `ARRAY[]`?

test/integration/data-types.test.js
gazoakley: mssql has a `_bindParam` function that outputs either a Buffer or a normal string dependent on whether or not the `STRING` data type uses the binary param. This override isn't needed for other dialects - I've added a comment.

test/integration/model/create.test.js
gazoakley: I've tidied this up a bit:

* https://github.com/sequelize/sequelize/pull/9492/files#diff-4b224bbc09dfb867f8671d77e298bb8cR955
* https://github.com/sequelize/sequelize/pull/9492/files#diff-4b224bbc09dfb867f8671d77e298bb8cR955

gazoakley:
@sushantdhiman @janmeier Thanks 🎉

sushantdhiman:
Thanks for this contribution @gazoakley , This is an important missing feature from Sequelize

PR #9517, opened 2018-06-08
fix(constructor): set the username, the password and the database via constructor options

ducky-hong:
### Pull Request check-list

### Description of change

The [document](http://docs.sequelizejs.com/class/lib/sequelize.js~Sequelize.html#instance-constructor-constructor) says that the username, the password and the database can be passed as options with URL. (e.g. `new Sequelize('mysql://localhost:3306', { database, username, password })`)  
But the current behavior does not honour these fields in options. This PR allows to pass these fields as options.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -195,14 +195,10 @@ class Sequelize {
 
     this._setupHooks(options.hooks);
 
-    if (['', null, false].indexOf(config.password) > -1 || typeof config.password === 'undefined') {
-      config.password = null;
-    }
-
     this.config = {
-      database: config.database,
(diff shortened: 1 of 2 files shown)


lib/sequelize.js
```diff
@@ -200,9 +200,9 @@ class Sequelize {
     }
```
sushantdhiman: Line 198, which looks like a complex empty check can be removed now,  just having this should work

```js
password: config.password || this.options.password || null
```

ducky-hong: I removed the empty check of the password.

Issue #9520, opened 2018-06-09
Postgres error.test.js and other tests depend on extensions that are not installed by default


mjcmatrix:
## What are you doing?
Running the postgresql integration tests. 
```
npm run test-postgres 
```

## What do you expect to happen?
The tests run OK.

## What is actually happening?
When mocha gets to integration test `integration/dialects/postgres/error.test.js`, `beforeEach()` fails because the following SQL statement generated by error.test.js line 23 relies on a postgres extension that is not installed by default (tested on Ubuntu 16.04):
```
ALTER TABLE "Bookings" add constraint overlap_period EXCLUDE USING gist ("roomNo" WITH =, period with &&);
```
Mocha output:
```
> sequelize@5.0.0-beta.7 test-integration /home/matt/projects_git/mjcmatrix/sequelize
> mocha $npm_package_options_mocha "test/integration/**/*.test.js"
...
  [POSTGRES] Alias
    1) "before each" hook for "should uppercase the first letter in alias getter, but not in eager loading"


  0 passing (86ms)
  1 failing

  1) "before each" hook for "should uppercase the first letter in alias getter, but not in eager loading":
     SequelizeUnknownConstraintError: [object Object]
      at Query.formatError (lib/dialects/postgres/query.js:353:17)
      at query.catch.err (lib/dialects/postgres/query.js:86:18)
  From previous event:
      at Query.run (lib/dialects/postgres/query.js:79:23)
      at Promise.try.then.connection (lib/sequelize.js:555:20)
  From previous event:
      at Promise.resolve.retryParameters (lib/sequelize.js:552:8)
      at /home/matt/projects_git/mjcmatrix/sequelize/node_modules/retry-as-promised/index.js:58:21
      at new Promise (<anonymous>)
      at retryAsPromised (node_modules/retry-as-promised/index.js:48:10)
      at Sequelize.query (lib/sequelize.js:451:28)
      at self.Booking.sync.then (test/integration/dialects/postgres/error.test.js:23:31)
  From previous event:
      at Context.<anonymous> (test/integration/dialects/postgres/error.test.js:22:8)
  From previous event:
```

__Dialect:__ postgres 
__Dialect version:__ 9.4.0 ?
__Database version:__ 9.5
__Sequelize version:__ 5.0.0-beta.7
__Tested with latest release:__ Yes (5.0.0-beta.7)

## Diagnosis

The test should run the following SQL before the failing statement:
```
[...]

PR #9521, opened 2018-06-09
Improve documentation of local database instances setup for test env

mjcmatrix:
Fixed issue with an environment dependency otherwise a specific test fails.
#9520: Postgres error.test.js depends on an extension that is not installed by default: `btree_gist`


mjcmatrix:
PR changed to modify CONTRIBUTING.md only

PR #9529, opened 2018-06-12
Fix 9515: sync() with "alter:true" doesn't use field name in model definition

### LINKED ISSUE #9515: sync() with "alter:true" doesn't use field name in model definition
## What are you doing?

I have a model which is defined like
```js
sequelize.define(‘Test’, {
      test_key: {
          field: ‘testKey’,
          type: DataTypes.BIGINT,
          allowNull: false
      },
})
```
Then I tried to generate the schema by calling
```js
sequelize.sync({ alter: true, force: true })
```
I was expecting the column name to be created as "test_key", but it turns out to be "testKey"

And when I tried to call sync() without "alter:true" like
```js
```
Everything works fine.
So there might be a issue for the alter functionality, please help to look into this.

I expected sync() with "alter:true" use the "field" option to create the column name.

But sync() with "alter:true" ignores the  "field" option

__Dialect:__ postgres
__Dialect version:__ XXX
__Database version:__ XXX
__Sequelize version:__ 4.37.10
__Tested with latest release:__ 4.37.10

mjcmatrix:
Closes #9515

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1219,7 +1219,7 @@ class Model {
     options = _.extend({}, this.options, options);
     options.hooks = options.hooks === undefined ? true : !!options.hooks;
 
-    const attributes = this.tableAttributes;
+    const attributes = this.fieldRawAttributesMap;
 
     return Promise.try(() => {
       if (options.hooks) {
--- a/lib/model.js
+++ b/lib/model.js
@@ -1219,7 +1219,8 @@ class Model {
     options = _.extend({}, this.options, options);
     options.hooks = options.hooks === undefined ? true : !!options.hooks;
 
-    const attributes = this.fieldRawAttributesMap;
+    const attributes = this.tableAttributes;
+    const rawAttributes = this.fieldRawAttributesMap;
 
     return Promise.try(() => {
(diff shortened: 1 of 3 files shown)


lib/model.js
```diff
@@ -1247,11 +1248,11 @@ class Model {
 
               _.each(attributes, (columnDesc, columnName) => {
                 if (!columns[columnName]) {
-                  changes.push(() => this.QueryInterface.addColumn(this.getTableName(options), columnName, attributes[columnName]));
+                  changes.push(() => this.QueryInterface.addColumn(this.getTableName(options), attributes[columnName].field || columnName, attributes[columnName]));
                 }
               });
               _.each(columns, (columnDesc, columnName) => {
-                const currentAttributes = attributes[columnName];
+                const currentAttributes = rawAttributes[columnName];
```
sushantdhiman: rename `currentAttributes` to `currentAttribute`

lyh2668:
Why is this problem not fixed in version 4.41.2?

lyh2668:
@sushantdhiman I upgraded the version to the version 5.0.0-beta.14 and still have this problem.

Issue #8583, opened 2017-11-02
undefined fields in where options are treated as NULL


gabegorelick:
## What are you doing?
```js
User.findAll({
  where: {
    name: req.query.name
  }
});
```

## What do you expect to happen?
If `name` param wasn't passed, then this should find all users.

## What is actually happening?
`SELECT ... FROM users WHERE name IS NULL`, i.e. if a field in `where` is `undefined`, it's treated as `NULL`. 

This is especially confusing because it treats `undefined` in Javascript two different ways. If the value is present but `undefined`, then it's `NULL`. But if it's missing and `undefined`, then it's obviously not included in the WHERE clause at all. This means that consumers can't rely on `where.name !== undefined` to know whether a clause will be included. Instead you have to use the relatively obscure `in` operator.

__Dialect:__ any, but tested on Postgres
__Dialect version:__ any
__Database version:__ any
__Sequelize version:__ master
__Tested with master branch:__ Yes

sushantdhiman:
This is a point of discussion, I mostly agree `undefined != null`. This is going to be a breaking change.

We also need to find other API that may also treat `undefined` as `null`, I can think of `update` atleast

gabegorelick:
Yeah this will be a huge breaking change. It may be wise to hide it behind an option first (something like `Model.find({where: {foo: undefined}, treatUndefinedAsNull: false})` and then change the default behavior in a future major version release.

rokoroku:
I totally agree with `undefined != null` too.
We can also add a flag like `treatUndefinedAsNull` to the global configuration object. It won't cause a big change.

leahciMic:
I think this might actually be a dangerous route...

Imagine this in a destroy scenario:

```js
comments.destroy({
  where: {
    id: comment.id,
    userId: user.id,
  },
});
```

Currently, if `comment.id` is `undefined` Sequelize wont destroy every comment belonging to a particular user. I think the current working way is much safer and a much saner default. 

I can't see any good reason to make this change.

If a `key` is defined but is `undefined`, then treat it like `null`. This change could result in unintentional loss of data through application errors and the like.

@gabegorelick Could you not write/ find a function that recursively strips undefined? Here is a poor man's version:

```js
function stripUndefined(obj) {
  return JSON.parse(JSON.stringify(obj));
}
```

And use it where necessary?

gabegorelick:
> Currently, if comment.id is undefined Sequelize wont destroy every comment belonging to a particular user.

[...]

PR #9512, opened 2018-06-07
[WIP] feat(query-generator): Generate SELECTs using bind parameters

gazoakley:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [ ] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](../CONTRIBUTING.md)?

### Description of change

WIP changes to allow SELECTs to use bind parameters instead of directly included values. For example, instead of:

```sql
SELECT `email`, `first_name` AS `firstName` FROM `User` WHERE `User`.`email` = 'jon.snow@gmail.com' ORDER BY `email` DESC LIMIT 10;
```

Generate:

```sql
SELECT `email`, `first_name` AS `firstName` FROM `User` WHERE `User`.`email` = $1 ORDER BY `email` DESC LIMIT 10;
```

(with parameter 'jon.snow@gmail.com')

Issue #7056, opened 2016-12-29
update method tries to insert null when receiving undefined 


jony89:
## What you are doing?

I map some request to Sequelize model : 

```js
        var model = {
            title: req.body.title,
            description: req.body.description
        };
        models.job.update(model, {..}).then(() => {})
```

let say that req.body.title is undefined and req.body.description has value.
In the model, `title` cant be null.
For some reason sequelize tries to insert null to the title although it is undefined, resulting in getting the following error : 

> err.stack SequelizeValidationError: notNull Violation: title cannot be null

this happens probably because  model.hasOwnProperty('title') returns true, but yet, model.title is `undefined` and not `null`
## What do you expect to happen?

I think that in that case, when a field is undefined, Sequelize should not try to insert null into it (the title field) and just ignore it 

## What is actually happening?

i get an error : 

> err.stack SequelizeValidationError: notNull Violation: title cannot be null

I think we should insert null only when someone specifically specify that this field is null. 

__Dialect:__ mysql 
__Database version: 5.7.9
__Sequelize version: 3.28.0

jony89:
if i add :

```
  //remove undefined values
  Object.keys(values).forEach(function(key) {
    if (values[key] === undefined) {
      delete values[key];
    }
  }); 
```

at models.js after the comment :

> // Remove values that are not in the options.fields

it works well, I think we should insert null only when someone specifically specify that this field is null.

jony89:
any help?

jony89:
@sushantdhiman  could u help here? someone? :S

jony89:
if i add :

```js
  //remove undefined values
  Object.keys(values).forEach(function(key) {
    if (values[key] === undefined) {
      delete values[key];
    }
  }); 
```

at ./lib/models.js after the comment :

> // Remove values that are not in the options.fields

it works well, is this a possible solution ? should i create a PR?

jony89:
I've done few more testing and I found out that if the field allowed to have nulls, and i try to update with value undefined it actually skips it as desired.

This actually supports my claim.. but it gives a different approach, maybe the problem is only with the validation ? 

If it is skipped, then why the validation checks for undefined as well : 

>   if (rawAttribute.allowNull === false && ((value === null) || (value === undefined))) {
>     error = new sequelizeError.ValidationErrorItem(field + ' cannot be null', 'notNull Violation', field, value);
>     this.errors.push(error);
>   }

could be 

[...]

PR #9548, opened 2018-06-16
 fix(query-generator): ignore undefined keys in query

Alexsey:
### Pull Request check-list

### Description of change

Fixes https://github.com/sequelize/sequelize/issues/8583
Right now if a query contains an `undefined` keys they are converted to `NULL` in `= NULL` way that makes queries absolutely useless.  This PR would ignore `undefined` keys in queries, just as settled in the issue

BTW [this code](https://github.com/sequelize/sequelize/blob/master/lib/sql-string.js#L24) is treating `undefined` as `NULL` that is contradict to discussion in https://github.com/sequelize/sequelize/pull/8478 . Should it be changed some how?


sushantdhiman:
I just want to say https://github.com/sequelize/sequelize/issues/8583#issuecomment-361932059 is a good argument, especially for find / destroy operations

Alexsey:
Ok, another option that is better than current behavior is to `throw` on `undefined`. Any other acceptible ideas?

sushantdhiman:
Why don't we push `undefined` to queries, no conversion. 

I think original idea was similar to this, push `null` / `undefined` to queries, but as there is no SQL equivalent of `undefined`, it was selected to use `null` instead.

>  global parameter that would switch between throw and ignore behavior with default to throw

I don't like switching behaviours with such flags, it makes it hard to debug why two similar where options are failing on one machine but working on other machine.

https://github.com/sequelize/sequelize/issues/8583#issuecomment-361932059 argument is good, but should we follow it? if I am correct we already ignore `undefined` for updates, so it make sense we should follow common behaviour.

I guess I would like to protect those unsuspecting users which might end with `undefined` in critical queries, like DELETE queries, these `undefined` cases are hard to cover for unsuspecting developer and often appear on only production cases.

I think we can ignore `undefined` for all queries except DELETE if that make sense

Alexsey:
Ok, so I'm going to ignore `undefined` for all but `DELETE` and for throw for `DELETE`

Alexsey:
@sushantdhiman looks like we are not ignoring `undefined` on `UPDATE`:
```js
const Sequelize = require('Sequelize')

const sequelize = new Sequelize(
  'db', 'root', 'root', {
    host: 'localhost',
    dialect: 'mysql',
    pool: {max: 100, min: 0, idle: 10000},
    dialectOptions: {
      charset: 'utf8_general_ci',
    },
  }
)

;(async () => {
  await sequelize.authenticate()

    id: {type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true},
    name: {type: Sequelize.STRING, allowNull: true}
  })

  await sequelize.sync({force: true})

  await User.update({name: 'ololo'}, {where: {name: undefined}})

  await sequelize.close()
})()
```
will throw
```
Error: Named bind parameter "$3" has no value in the given object.
    at sql.replace (F:\sequelize\lib\dialects\abstract\query.js:84:15)
    at String.replace (<anonymous>)
    at Function.formatBindParameters (F:\sequelize\lib\dialects\abstract\query.js:67:15)
    at Function.formatBindParameters (F:\sequelize\lib\dialects\mysql\query.js:37:25)
    at Promise.try (F:\sequelize\lib\sequelize.js:498:44)
    at tryCatcher (F:\sequelize\node_modules\bluebird\js\release\util.js:16:23)
    at Function.Promise.attempt.Promise.try (F:\sequelize\node_modules\bluebird\js\release\method.js:39:29)
    at Promise.resolve.retryParameters (F:\sequelize\lib\sequelize.js:447:64)
    at F:\sequelize\node_modules\retry-as-promised\index.js:58:21
    at new Promise (<anonymous>)
    at retryAsPromised (F:\sequelize\node_modules\retry-as-promised\index.js:48:10)
    at Sequelize.query (F:\sequelize\lib\sequelize.js:447:28)
    at QueryInterface.bulkUpdate (F:\sequelize\lib\query-interface.js:1045:27)
    at Promise.try.then.then.then.results (F:\sequelize\lib\model.js:2963:34)
    at tryCatcher (F:\sequelize\node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (F:\sequelize\node_modules\bluebird\js\release\promise.js:512:31)
    at Promise._settlePromise (F:\sequelize\node_modules\bluebird\js\release\promise.js:569:18)
    at Promise._settlePromise0 (F:\sequelize\node_modules\bluebird\js\release\promise.js:614:10)
    at Promise._settlePromises (F:\sequelize\node_modules\bluebird\js\release\promise.js:693:18)
    at Async._drainQueue (F:\sequelize\node_modules\bluebird\js\release\async.js:133:16)
    at Async._drainQueues (F:\sequelize\node_modules\bluebird\js\release\async.js:143:10)
    at Immediate.Async.drainQueues (F:\sequelize\node_modules\bluebird\js\release\async.js:17:14)
```
and I can assume that after https://github.com/sequelize/sequelize/pull/9512 `SELECT` will also throw on the same place if any of the keys would be `undefined`

Of course, if I'll add the code I'm having now there would be no query generated for `undefined` and no throw in that place. But that would make the code that is now throwing irrelevant. And probably some more

How would do you think it should be resolved?

sushantdhiman:
@Alexsey I meant for `undefined` in update payload,

```js
user.update({
   name: undefined // will be ignored
})
```

> But that would make the code that is now throwing irrelevant. And probably some more

How so? I agree as we start ignoring undefined for update / select, we wont be generating queries for `undefined` values thus there will be no extra bind variable which will throw, but bind parameter formatting will remain relevant.

> How would do you think it should be resolved?

Ignore `undefined` for INSERT / UPDATE / SELECT, so there is no extra bind param which will throw this opaque error message

Fail fast for `undefined` keys in DELETE case, before code reaches to bind parameter formatting and stuff

Alexsey:
@sushantdhiman it's ready

lib/dialects/abstract/query-generator.js
```diff
@@ -2089,7 +2090,13 @@ class QueryGenerator {
 
   whereItemQuery(key, value, options) {
     options = options || {};
-    if (key && typeof key === 'string' && key.indexOf('.') !== -1 && options.model) {
+    if (value === undefined) {
+      if ([QueryTypes.BULKDELETE, QueryTypes.DELETE].includes(options.type)) {
+        throw Error(`WHERE parameter "${key}" of DELETE query has value of undefined`);
```
sushantdhiman: just for consistency sake

```throw new Error(`....`) ```

File: lib/dialects/abstract/query-generator.js [hunk 2]
```diff
@@ -2089,7 +2090,13 @@ class QueryGenerator {
...
+      }
+      return;
```
sushantdhiman: A comment on return 

```for other query types, ignore all where parameters with undefined value```

File: test/unit/sql/delete.test.js [hunk 3]
```diff
@@ -177,5 +185,30 @@ suite(Support.getTestDialectTeaser('SQL'), () => {
...
+      test(util.inspect(options, {depth: 2}), () => {
+        const sqlOrError = _.attempt(
+          sql.deleteQuery.bind(sql),
+          options.table,
+          options.where,
+          options,
+          User
+        );
+        return expectsql(sqlOrError, {
+          default: Error('WHERE parameter "name" of DELETE query has value of undefined'),
```
sushantdhiman: alone `default` should be enough

sushantdhiman: And use `new Error`

gabegorelick:
> Ignore undefined for INSERT / UPDATE / SELECT, so there is no extra bind param which will throw this opaque error message
> 
> Fail fast for undefined keys in DELETE case, before code reaches to bind parameter formatting and stuff

I disagree on having different behavior for DELETE. UPDATE can be just as destructive as DELETE, so having different behavior for DELETE in the name of safety doesn't seem worth it.

Alexsey:
I don't have a strong opinion on this - just want to fix current behaviour that is broken from any perspective. We can make `throw` on `undefined` in `UPDATE` and `UPSERT` as well :)

sushantdhiman:
>I disagree on having different behavior for DELETE. UPDATE can be just as destructive as DELETE, so having different behavior for DELETE in the name of safety doesn't seem worth it.

Incorrect or missing data may be, but at-least records wont be deleted. I just want to avoid gotcha moments which involves loss of data

> We can make throw on undefined in UPDATE and UPSERT as well :)

:) then half of the queries will be running in different mode than other, but actually make sense to throw for `undefined` for any `Create / Update / Delete` operation and ignore for `Select` operations. wdyt?

Would also like comments from @sequelize/orm

sushantdhiman:
Including #7056 in discussion, I propose

| Operation        | null      | undefined |
|------------------|-----------|-----------|
| Insert           | no change | ignore    |
| Update (Payload) | no change | ignore    |
| Update (Where)   | no change | throw     |
| Select           | no change | ignore    |
| Delete           | no change | throw     |

Alexsey:
I think we should be concerned only about operations that affect multiple records on write. That are `BULKDELETE` and `BULKUPDATE`. So we should `throw` for them and ignore for any other. Any `Create` is also out of the scope just because it cannot have `where` and be multiple at the same time

Alexsey:
as for #7056 I think we can include it in the scope of this PR, but it's quite different and nothing is still done for it

sushantdhiman:
> I think we should be concerned only about operations that affect multiple records on write. That are BULKDELETE and BULKUPDATE. So we should throw for them and ignore for any other.

Sounds good

Alexsey:
Because @jony89 told that he would like to finish his PR for update payload https://github.com/sequelize/sequelize/issues/7056#issuecomment-398989599 I would not try to include those changes in this PR

At this point, I have nothing more to add to code unless someone else has some comments

Alexsey:
Wait, we have forgotten one special case - for MySql for `bulkCreate` there is an [`updateOnDuplicate` option](http://docs.sequelizejs.com/class/lib/model.js~Model.html#static-method-bulkCreate) that can also update multiple records while not been `BULKUPDATE` or `BULKDELETE`. I think we should throw for it as well

Alexsey:
BTW I think this should have a **big** announcement in the migration to v5 guide, because this may silently break security:
```js
const authorizationToken = AuthorizationToken.findOne({where: {token: cookie.someSessionToken})
assert(authorizationToken, Error('Unauthorized'))
```
so this PR can break authorization check by returning arbitrary result instead of no result. Possibly some other simmilar valnaribilities

gabegorelick:
> BTW I think this should have a big announcement in the migration to v5 guide, because this may silently break security

Is there an option to change this behavior? That would make migrating a lot safer and easier for such cases.

mickhansen:
This change is definitely a bit dangerous, i don't think errors like this should fail silently.
Having an undefined value in a where object is most likely an error and should be treated as such, there are a few cases where having them be ignored is a feature, but i don't think that outweights the potential security threats here.

Alexsey:
Looking in my codebase I can tell that it's the desired behavior in at least 90% of cases, so generally, I'm for having it this way. Reverting this to the way it was is is not an option because the SQL that was generating before is close to invalid

And I don't think there is any significant risk for a new code. Only for migration

mickhansen:
@Alexsey Having invalid SQL sounds more proper to me, since that's atleast not silently ignoring an error like this change does.

While i do agree that the SQL generated was unexpected, i think the new SQL can produce even more unexpected behaviour. When something is an error i think it's better to return less results than more results, with security and ACL in mind.

As you yourself mention this could introduce vulnerabilities, which i think is generally the wrong way to go. In your example `AuthorizationToken.findOne({where: {token: cookie.someSessionToken})` i think it's more proper to generate `token=NULL` rather than no where statement at all.

Alexsey:
@mickhansen I don't think ignoring `undefined` is an error. <a href="https://github.com/sequelize/sequelize/issues/8583">Treating it as `NULL`</a> was also decided to be a bad idea

But if we want to cover all cases I can propose a new `whereRequired` that would act the same as `where` but `throw` if some key is `undefined`. Before request `where` would be merged with `whereRequired` and throw on any overlap

It's just a brainstorm - any ideas are non-final and discussible

calmdev:
Following the discussion and it does seem a bit sketchy. I tend to agree with @mickhansen and [the comment](https://github.com/sequelize/sequelize/issues/8583#issuecomment-361932059) made by @leahciMic  in the other issue. Explicitly dropping the keys from your query params prior to passing them to sequelize shows clear intent vs potentially exposing or destroying data by accident. 

The original example in #8583 could be completely avoided by ensuring that the route which is handling the request validates that name is a string before the request is accepted and route handler is executed instead of assuming the param will exist.

Placing this functionality behind a flag was also suggested. I feel that is a more more acceptable way to support this functionality as it explicitly show your intentions to automatically have sequelize drop keys from your query.

Alexsey:
@calmdev IMHO it's too minor and too specific thing to be flagged. If we will have a flag for each thing this size we will have hundreds of them. And the comment you are pointing is not relevant because `destroy` will throw if some of the keys will be `undefined`. As well as `update`. The logic here is that it is not allowed for operations that mutates multiple instances

calmdev:
Yes, true. I see in the table above. My intention referencing that comment was that it's easy enough to omit the keys with undefined values before passing them to a query. I wouldn't say it's a minor thing though, because the proposal doesn't address the potential issue of exposing more data than intended which you guys have already pointed out. All the other use cases in original issue are related to selecting data. Regardless, I'm still not convinced that this should be default behavior or even the responsibility of sequelize :man_shrugging:

jony89:
I think we should distinguish between` insert` / `update` and `select` / `delete`. while one does `insert` and `update` it seems like a correct behavior, while selecting or deleting I expect to have much more explicit code. 

IMO we should keep this PR and add this slight change. 

I understand that one can think that this way the behavior across the system is not consistent. but it does not have to be always, there is different solution for different cases.

sushantdhiman:
New version [5.0.0-beta.10](https://github.com/sequelize/sequelize/blob/master/docs/upgrade-to-v5.md#500-beta10) published with this change.

Thanks for new comments everyone, this is still a beta feature / release so we can change things. For me here are things that I don't want to support

1. A flag to switch behaviour of omitting undefined keys, just too many flags and error prone
2. Converting `undefined` to `null`, as originally intended in this PR

What do you think about converting `undefined` to `String(undefined)`, in this way it's not being modified by Sequelize to unexpected query.  Other concern related to unexpected selection set due to empty where are also addressed with this. 

Updating my original table

| Operation        | null      | undefined |
|------------------|-----------|-----------|
| Insert (Payload) | no change | ignore    |
| Update (Payload) | no change | ignore    |
| Update (Where)   | no change | **convert**     |
| Select           | no change | **convert**    |
| Delete           | no change | **convert**     |

jony89:
@sushantdhiman , I also agree with 1. + 2. 

Yet I don't think that using `String(undefined)` is the correct solution and it seems like it will pollute the code with some new constant string that in the feature will be hard to understand by just reading the code. Also, for me, I would want to get notification if somehow I got into case where I have `.delete({where: {id: undefined}})` rather than having sequelize swallow it.

I liked it better with the `throw` attitude.

Alexsey:
It will not swallow. It will throw

Alexsey:
I don't think that `undefined` string is a good idea because no one would like to see `undefined` in the db and so every time it would be used that way it would be a silent bug. Not critical, but bug. So from that perspective always `throw` on `undefined` is just better than convert

QuanticPotatoes:
I got a problem (or a feature ?). On sequelize 5, the `undefined` properties on the where clause are correctly ignored. But on the include clause, it's another story. The undefined properties seem not be ignored. It's normal ?

```js
users.findAll({
      attributes: ['idUsers', 'email'],
      where: {
         email: undefined,
      },
      include: [
        {
          model: models.roles,
          attributes: ['id_roles', 'name'],
          where: {
            name: undefined, // This line don't work on the where's include
          },
          through: {
            attributes: [],
          },
        },
      ],
    });
```

Bene-Graham:
Ignoring the parameter when it is `undefined` when doing a `SELECT` operation could introduce data leakage (possibly quite serious) in the application. 

Now I do think the developers should share in part of the onus on validating input. I think it would be better for the library to fail fast when the intention of the developer and the data given do match up. At least now in 4.x we get a query that is impossible to get results from `field = NULL`

---

Now for real life example 😄 

In my application I had an edge case where one of the field in the where clause was undefined. In sequelize v4 this was just returning zero results. With what has been purposed / merged for v5 would have caused unauthorized users see data they should not have access to too 😢

sushantdhiman:
Now Sequelize will throw for any `undefined` where parameters https://github.com/sequelize/sequelize/pull/10048

PR #9549, opened 2018-06-17
change: remove deprecation from notNull for custom messages / enforce use of allowNull with this validator 

### LINKED ISSUE #1500: customizable allowNull error messages
something like this:

```
    'email': {
      type: DataTypes.STRING,
      unique: true,
      allowNull: { args: false, msg: 'Email is required.' },
      validate: { isEmail: { msg: 'Invalid email.' } },
    },
```

sushantdhiman:
…lowNull usage with it

### Pull Request check-list

### Description of change

Fixes #1500 

Sequelize can customize `allowNull` error messages with `notNull` validator, but validator throw error when value is not `null` due to deprecation for `notNull`.

I have removed that deprecation, it was only introduced to enforce users to use `allowNull` over `notNull`, I have added some checks that will only allow use of `notNull` with `allowNull: false`.

So overall if users want to prevent null values they can as usual use `allowNull: false`, but if they want to customize that error message, they can use `notNull.msg`

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -89,7 +89,6 @@ class Model {
     this._changed = {};
     this._modelOptions = this.constructor.options;
     this._options = options || {};
-    this.__eagerlyLoadedAssociations = [];
 
     /**
      * Returns true if this instance has not yet been persisted to the database
@@ -906,7 +905,11 @@ class Model {
       attribute = this.sequelize.normalizeAttribute(attribute);
 
       if (attribute.type === undefined) {
-        throw new Error('Unrecognized data type for field ' + name);
+        throw new Error(`Unrecognized datatype for attribute "${this.name}.${name}"`);
+      }
+
+      if (attribute.allowNull !== false && _.get(attribute, 'validate.notNull')) {
+        throw new Error(`Invalid definition for "${this.name}.${name}", "notNull" validator is only allowed with "allowNull:false"`);
       }
 
       if (_.get(attribute, 'references.model.prototype') instanceof Model) {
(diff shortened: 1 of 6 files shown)


lib/utils/validator-extras.js
```diff
@@ -77,9 +77,8 @@ function extendModelValidations(modelInstance) {
 }
 exports.extendModelValidations = extendModelValidations;
 
-// Deprecate this.
-validator.notNull = function() {
-  throw new Error('Warning "notNull" validation has been deprecated in favor of Schema based "allowNull"');
+validator.notNull = function(val) {
+  return [null, undefined].indexOf(val) === -1;
```
Alexsey: Since v5 targets NodeJS 6+ it's now possible to `return [null, undefined].includes(val)`

sushantdhiman: Thanks, I just copied this code from old PR, will change

agims7:
@sushantdhiman when those changes will be online with the latest npm varsion?

lib/model.js
```diff
@@ -906,7 +905,11 @@ class Model {
       attribute = this.sequelize.normalizeAttribute(attribute);
 
       if (attribute.type === undefined) {
-        throw new Error('Unrecognized data type for field ' + name);
+        throw new Error(`Unrecognized datatype for attribute "${this.name}.${name}"`);
+      }
+
+      if (attribute.allowNull !== false && _.get(attribute, 'validate.notNull')) {
+        throw new Error(`Invalid definition for "${this.name}.${name}", "notNull" validator is only allowed with "allowNull:false"`);
```
contra: This is a great change for developer UX 👍 @sushantdhiman

Issue #8749, opened 2017-12-04
toJSON converting booleans to "t" (true) or "f" (false)


jonahlau:
## What are you doing?

I have the user model as defined below. 
```js
// User model

sequelize.define('user', {
  email: {
    type: STRING,
    allowNull: false,
    unique: true,
    validate: {
      isEmail: true,
      notEmpty: true
    }
  },
  password: {
    type: STRING,
    allowNull: false,
    visible: []
  },
  firstName: {
    type: STRING,
    allowNull: false
  },
  lastName: {
    type: STRING,
    allowNull: false
  },
  active: {
    type: BOOLEAN,
    allowNull: false,
    defaultValue: true
  },
  role: {
    type: STRING,
    allowNull: false,
    defaultValue: 'user'
  },
  isVerified: {
    type: BOOLEAN,
    defaultValue: false
  },
  resetPasswordToken: {
    type: STRING,
    allowNull: true,
    visible: []
  },
  resetPasswordExpires: {
    type: DATE,
    allowNull: true,
    visible: []
  },
  lastLogoutTime: {
    type: DATE,
    allowNull: true,
    visible: ['admin']
  }
});


```

## What do you expect to happen?
toJSON should retain booleans and not convert them to strings  

## What is actually happening?
When I call toJSON of any instance of User in my CI tests (running in as a docker container with node 8.9.1), all booleans, for example fields `active` and `isVerified`, get converted to "t" or "f". 

The problem doesn't happen when I run it on MacOS 10.12.15 running Node 8.9.1

_Output, either JSON or SQL_

__Dialect:__ postgres
__Dialect version:__ XXX
__Database version:__ 9.5
__Sequelize version:__ 4.2.0
__Tested with latest release:__ yes (If yes, specify that version)

**Note :** _Your issue may be ignored OR closed by maintainers if it's not tested against latest version OR does not follow issue template._

karn09:
I'm seeing this same issue, but with 4.26.0. @jonahlau Did you mean 4.26.0 as well? You had marked yes to tested with latest. 

Downgrading to 4.25.2 fixed the issue for me.

jonahlau:
@karn09 Ah, yes sorry for the typo. I did mean 4.26.0

Downgrading to 4.25.2 fixed it for me. So should I avoid upgrading to latest release until this is resolved?

karn09:
[...]

Issue #7318, opened 2017-03-02
`removeAttribute('id')` results in `undefined: null` data value


broofa:
Removing the `id` property results in an "undefined" property being set on the dataValues object.

```
> var Thing = db.sequelize.define('Thing')

> console.log(model.build().dataValues)
{ id: null }

> model.removeAttribute('id')

> console.log(model.build().dataValues)
{ undefined: null }

> model.build().toJSON()
{ undefined: null }
```

__Dialect:__ mysql
__Database version:__ mysql  Ver 14.14 Distrib 5.7.16, for osx10.11 (x86_64) using  EditLine wrapper
__Sequelize version:__ 3.30.2

sushantdhiman:
Its a bug, with very low priority.

**SSCCE**
```js
const Entry = sequelize.define('Entry', {
  description: {
    type: Sequelize.STRING
  },
  value: {
    type: Sequelize.DECIMAL(17, 2),
    allowNull: false
  }
});

let instance = Entry.build({ description: 'Hello', value: 23 });
console.log(instance.toJSON());

Entry.removeAttribute('value');
instance = Entry.build({ description: 'Hello', value: 23 });
console.log(instance.toJSON());

Entry.removeAttribute('description');
Entry.removeAttribute('id');
instance = Entry.build({});
console.log(instance.toJSON());
```

**OUTPUT**
```js
{ id: null, description: 'Hello', value: 23 }
{ id: null, description: 'Hello' }
{ undefined: null }
```

sushantdhiman:
Fixed in https://github.com/sequelize/sequelize/pull/7382

ratulSharker:
I'm still facing this issue. Removed `id` using `Model.removeAttribute(`id`)`, then including `hasMany` association causes `null` values to appear.

PR #9558, opened 2018-06-19
issue github

jeremie-stratumn:
I am posting this as a PR as I was not able to submit a new issue on github.. the submit button is greyd out.. This was the only way I found to get in touch :)

## What are you doing?
If `User` is a model, I do:
```js
const query = User.findAll;
...
const res = query();
```

no error

I get the following error: `Cannot read property 'warnOnInvalidOptions' of undefined`

occurs here: https://github.com/sequelize/sequelize/blob/7525ef066d9c7ef48527d15cf44bc719a69162c4/lib/model.js#L1659

if I do: `const query = User.findAll.bind(User)` then it works.

_Output, either JSON or SQL_

__Dialect:__ postgres 
__Sequelize version:__ 4.37.10
__Tested with latest release:__ No (If yes, specify that version)

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1656,6 +1656,7 @@ class Model {
       }
     }
 
+    // this is where the error occurs
     this.warnOnInvalidOptions(options, Object.keys(this.rawAttributes));
 
     const tableNames = {};


sushantdhiman:
Please use Github Issue Tracker only for reporting bugs, requesting new features or discussions. Ask questions on [Stackoverflow sequelize.js tag](https://stackoverflow.com/questions/tagged/sequelize.js) or [Slack](sequelize.slack.com). While I would like to help answer any questions but it still take too much time.

If you are reporting a bug please isolate it as SSCCE, present it nicely and follow issue template. This will help maintainers and contributors understand your issue quickly.

If you are requesting a feature spend some time to properly present your use case with some examples, current and expected outcome.

( [What is SSCCE ?](http://sscce.org) You can find template for Sequelize [here](https://github.com/sequelize/sequelize/blob/master/sscce_template.js) )

A few good example of SSCCE

1) https://github.com/sequelize/sequelize/issues/8014#issue-245907597
2) https://github.com/sequelize/sequelize/issues/7318#issuecomment-284136530
3) https://github.com/sequelize/sequelize/issues/8749#issuecomment-354734396

PR #9577, opened 2018-06-22
(fix): Offset in UNION'ed queries

bl42:
Fixed an issue when passing limit/offset inside a subquery.

### Pull Request check-list

- [ - ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ - ] Have you added new tests to prevent regressions?
- [ - ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ x ] Did you follow the commit message conventions explained in [CONTRIBUTING.md](../CONTRIBUTING.md)?

### Description of change

When executing a query like so, the offset value would not get passed to the UNION ALL Generated query.
```
models.tblLeague.findAll({
      include: [
        {
          model: models.tblPost,
          order: [['createdDate', 'DESC']],
          offset: 5,
          limit: 10,
          where: { isActive: state.active },
        },
      ],
      where: { leagueId: [1,2,3] },
    });
```


bl42:
> @bl42 @sushantdhiman @janmeier ,
> 
> Do any of you see why this wouldn't also work for Sequelize v4?

I don't see any reason why it wouldn't work for v4.

PR #9585, opened 2018-06-23
fix: skip validation for undefined values while updating V3

### LINKED ISSUE #7056: update method tries to insert null when receiving undefined

jony89:
Pull Request check-list

### Description of change

Fixes #7056, after #7167 has failed since it affects the `.create` method (and should have not).

As stated in #7056, while assigning `undefined` value into a non-nullable prop and calling `.update` a validation Error is raised, yet while assigning `undefined` to nullable prop is completely ignored and not shown in the update query, due to the `mapValueFieldNames` function which removes `undefined` props as desired.
This also aligning the behavior with #9548 .

BWC checks for non-nullable props : 

- assigning null : no implication
- assigning value : no implicaation
- assigning undefined : no implication in final query yet no exception is raised now.

after this is merged will be happy to help and migrate to v4 / master. (I'm still using v3)

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2495,6 +2495,15 @@ Model.prototype.update = function(values, options) {
 
       // We want to skip validations for all other fields
       options.skip = Utils._.difference(Object.keys(self.attributes), Object.keys(values));
+
+      // We want to skip validations for undefined values
+      // later they are removed by mapValueFieldNames and not shown in update query
+      Object.keys(values).forEach(function(key) {
+        if (values[key] === undefined && options.skip.indexOf(key) < 0) {
+          options.skip.push(key);
+        }
+      });
+
       return build.hookValidate(options).then(function(attributes) {
         options.skip = undefined;
         if (attributes && attributes.dataValues) {
--- a/test/integration/model/update.test.js
+++ b/test/integration/model/update.test.js
@@ -42,6 +42,30 @@ describe(Support.getTestDialectTeaser('Model'), function() {
       });
     });
 
+    it('should not check for notNull Violation for undefined values', function () {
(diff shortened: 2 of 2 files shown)


sushantdhiman:
Sorry but v3 wont receive any new features or fixes, only critical security fixes will be ported back to v3, I suggest you target this branch to `master`

jony89:
OK I've created another PR of the master branch #9587 

Yet I'll be happy if u will merge this as well..

sushantdhiman:
Sorry @jony89 I can't accept this PR, this is a slight breaking change and I don't want to maintain v3 along side v4 and next v5. As already noted, v3 will only receive critical security fixes

peskic93:
I am aware that it's been a year since this, but can this be merged in v4? @sushantdhiman

PR #9587, opened 2018-06-23
fix: skip validation for undefined values while updating

### LINKED ISSUE #7056: update method tries to insert null when receiving undefined

jony89:
### Pull Request check-list

### Description of change

Fixes #7056, after #7167 has failed since it affects the `.create` method (and should have not).

As stated in #7056, while assigning `undefined` value into a non-nullable prop and calling `.update` a validation Error is raised, yet while assigning `undefined` to nullable prop is completely ignored and not shown in the update query, due to the `mapValueFieldNames` function which removes `undefined` props as desired.
This also aligning the behavior with #9548 .

BWC checks for non-nullable props : 

- assigning null : no implication
- assigning value : no implicaation
- assigning undefined : no implication in final query yet no exception is raised now.


jony89:
> Why don't we just filter out undefined from values input?

@sushantdhiman since there is `valuesUse` and `values` I guessed they might be used for something else.

I can do even better and move `valuesUse` to upper scope and remove entirely the `values`, but again I think it was done for a reason. what do u think ?

sushantdhiman:
https://github.com/sequelize/sequelize/pull/9587/files#diff-a140b60b6a99a8c69f4b5d0d1f1f2bfaR2851
filter out when we initially clone values

jony89:
@sushantdhiman yeah that what I thought at start, but there is no need for that because that's exactly what `mapValueFieldNames` does later.  so we can just do it sooner ?

sushantdhiman:
`mapValueFieldNames` will convert attributes to field name which may fail in next segment as it is expected that `fields` to contain attributes (bad name :D)

So safest option is to filter out `undefined` at the top, probably use [omitBy](https://lodash.com/docs/4.17.10#omitBy) which will filter for undefined keys and create new object in one go, removing need for `clone`

test/integration/model/update.test.js
```diff
@@ -35,6 +35,53 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         }));
     });
 
+    it('should not check for notNull Violation for undefined values', function() {
+      const ownerId = 2;
+      let accountRowId;
```
sushantdhiman: Get rid of `accountRowId`, if you do `account.findOne` you will directly get our target instance, as there is `sync` in before each.

test/integration/model/update.test.js
```diff
@@ -35,6 +35,53 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+      return this.Account.create({
+        ownerId,
+        name: Math.random().toString()
+      }).then(account => {
+        accountRowId = account.get('id');
+        const accountVal = {
```
sushantdhiman: values from `accountVal` can be passed directly to `this.Account.update`, no need to keep values as constant.

test/integration/model/update.test.js
```diff
@@ -35,6 +35,53 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+          }
+        });
+      }).then(() => {
+        return this.Account.findById(accountRowId);
+      }).then(account => {
+        expect(account.ownerId).to.be.equal(ownerId);
+      });
+    });
+
+    it('should ignore undefined values', function() {
```
sushantdhiman: Looks like duplicate of test above, Please add a test here https://github.com/sequelize/sequelize/blob/master/test/integration/instance/update.test.js so we can be sure instance method update also ignores `undefined` values

test/integration/model/update.test.js
```diff
@@ -35,6 +35,53 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         }));
     });
 
+    it('should not check for notNull Violation for undefined values', function() {
```
sushantdhiman: `should ignore undefined values without throwing not null validation`

test/integration/model/update.test.js
jony89: It was one for not null props and one for the nullable props (as it might behave different) but I guess the first one is enough

jony89:
> Please add a test here https://github.com/sequelize/sequelize/blob/master/test/integration/instance/update.test.js so we can be sure instance method update also ignores undefined values

So I had to edit the `.update` non-static method as well. otherwise the test in `test/integration/instance/update.test.js` would fail, the value in the DB was not updated with `undefined` but the instance in the RAM did.
see if it's fine by you

peskic93:
@sushantdhiman Can this be merged in v4?

Issue #9113, opened 2018-02-27
add association hooks


FranckFreiburger:
I wondering if it is possible to add hooks for Associations (like `beforeDefine`, `afterDefine`, ... for model)

## What are you doing?

```js
const User = db.define('user');
const Tag = db.define('user');

User.belongsToMany(Tag, { through: UserTag }); // trigger association created hook here
Tag.belongsToMany(User, { through: UserTag }); // trigger association created hook here
```

Currently, the best I can do is:
```js

var prev = Association.BelongsTo.prototype.injectAttributes;
Association.BelongsTo.prototype.injectAttributes = function() {
  // my hook here
  return prev.apply(this, arguments);
}

// ... and same for HasOne, BelongsToMany, HasMany
```

__Sequelize version:__ latest version

lukeed:
Add a `*Create` hook on your `UserTag` model.

FranckFreiburger:
I agree that this will work for belongsToMany because of the through model,
however I also need callback for new BelongsTo, HasOne and HasMany associations.

sushantdhiman:
You want to modify association option using these hooks I assume, currently there are no such hooks available but should be easy to add them. I am tagging this as a feature request.

NicolasNarvaez:
I need this to transparently keep track of references to a model, do you know another way to do it?

sushantdhiman:
Implemented in https://github.com/sequelize/sequelize/pull/9590

`beforeAssociate / afterAssociate`

PR #9590, opened 2018-06-25
Add association hook

mirkojotic:
### Pull Request check-list

### Description of change

Added hooks for association creation as per #9113


mirkojotic:
I may be way off on this one so feel free to critique it, I did it as I understood the issue but that could be very wrong.

I've added some new integration tests but I might have missed another place where I should add them.

mirkojotic:
@sushantdhiman I've made changes to the name of the hooks and changed what gets emitted with `afterAssociate` not sure if that's what your comment meant.

lib/associations/mixin.js
```diff
@@ -90,13 +105,20 @@ function singleLinked(Type) {
...
+    }
     // the id is in the foreign table
     const association = new Type(source, target, _.extend(options, source.options));
     source.associations[association.associationAccessor] = association;
 
     association._injectAttributes();
     association.mixin(source.prototype);
 
+    if (options.useHooks) {
+      this.runHooks('afterAssociate', target, options);
```
sushantdhiman: Instead of `target`, pass `association` (that is what I meany by emit association :p)

sushantdhiman: No, we just want to know what association got created, source and target info is available on this object

mirkojotic: @sushantdhiman Ok. Here is another stab at it. I'm not passing source or target anymore since they are available within hook through `this`. We're passing association object now. I'm open to more suggestions. Tests should be passing but they might be a bit wonky so I'll accept any tips.

sushantdhiman:
I was thinking more about usecase of these hooks, I think we need to support

1. beforeHasOneAssociate /  afterHasOneAssociate
2. beforeHasManyAssociate /  afterHasManyAssociate
3. beforeBelongsToAssociate /  afterBelongsToAssociate
4. beforeBelongsToManyAssociate /  afterBelongsToManyAssociate

Each `before*` hook should provide `target / options`, where as each `after*` hook should only provide `association`

Just unit tests will be enough for these hooks, you can add them in respective association test file https://github.com/sequelize/sequelize/tree/master/test/unit/associations 

> I'm not passing source or target anymore since they are available within hook through this

I agree `source` is available as `this` but how target is available?

mirkojotic:
@sushantdhiman Ok. I've tried to change stuff up as per your direction. Tests are passing locally on all dialects but for whatever reason they've failed on Travis. I haven't touched any files that might mess them up...

mickhansen:
I'd suggest just one before/after hook for associations, let userland inspect the association for type, less surface area.

lib/associations/mixin.js
```diff
@@ -26,13 +26,21 @@ const Mixin = {
 
     options = _.extend(options, _.omit(source.options, ['hooks']));
 
+    if (options.useHooks) {
+      this.runHooks('beforeHasManyAssociate', target, options);
```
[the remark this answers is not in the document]
sushantdhiman: We can have `{source, target}` for before* hooks, ``{source, target, association}`` for after* hooks, I would like to keep `options` separate as in future we may have conflicting options

sushantdhiman:
> I'd suggest just one before/after hook for associations, let userland inspect the association for type, less surface area.

I don't see a good way to identify what kind of association is being created, especially in `before*` hooks. We can try 

1) different before hooks, single after hook
2) single before after hook, with association constructor passed as `type`

mirkojotic:
@mickhansen @sushantdhiman

I'm willing to work with either of these options, just let me know what you guys think is a better option.

mirkojotic:
Hey @sushantdhiman I gave it another shot. Let me know if further modifications are required.

PR #9604, opened 2018-06-27
fix(lib/dialects/abstract/query): Pass 'includeValidated' as false in model build so not to skip initialisation in queries with joins

### LINKED ISSUE #9602: Includes in Raw Queries with joins require explicit call to internal validation method
## What are you doing?
Because of a fairly complex query for which I don't really expect any ORM alternatives, I need to make a raw query which needs to include associations.

From what I've gathered, setting the `hasJoin` flag to true in the options object would suffice so that Sequelize became aware that the data for associations is included (although one might argue as to whether that would be redundant along with the included association descriptions, no big harm done in making it explicit). After several failed attempts with different combinations (these don't seem to be very documented even though the functionality is there if I might add), here's what I did:

- Explicitly aliased all columns in the query to match model property names;
- Wrapped column names from associations in double quotes e.g. `association.column AS "association.column"`;
- Specified which model Sequelize should parse the raw data to in `model` in the options parameter;
- Set `hasJoin` to `true` in the options parameter;
- Added association descriptions to the `include` array in the options parameter;

Example:
```js
    const rawQuery= 'SELECT ...';
    sequelize.query(rawQuery, {
      replacements: { ... },
      model: MyModel,
      hasJoin: true,
      include: [
        {
          model: AnotherModel,
          as: 'oneOccurrence'
        },
        {
          model: AnotherModel,
          as: 'anotherOccurrence'
        },
        YetAnotherAssociatedModel
      ]
    });
```

The above steps seem logic in order to get parsed results from a raw query along with associations.

The above call fails at 2 points in Sequelize's internal calls, respectively:
- `lib/dialects/abstract/query.js` at line 422 where it tries to use an uninitialised `includeMap` object
- `lib/model.js` at line 3371 where it tries to check for a key in an uninitialised `includeNames` (would-be) array

The workarounds for this would be either sending initial values for both objects in the options parameter as such:

```js
    const rawQuery= 'SELECT ...';
    sequelize.query(rawQuery, {
      replacements: { ... },
      model: MyModel,
      hasJoin: true,
      include: [
        {
          model: AnotherModel,
          as: 'oneOccurrence'
        },
        {
          model: AnotherModel,
          as: 'anotherOccurrence'
        },
        YetAnotherAssociatedModel
      ],
      includeMap: {},
      includeNames: []
    });
```

...or to make the explicit call to `_validateIncludedElements` from Sequelize's internals which initialises both before the query is issued. I've chosen this approach because it looks to me slightly more meaningful than including the 2 empty objects for no apparent reason:
```
    // Importing this also makes no sense here
    const Model = require('sequelize/lib/model');

    const includes = {
      include: [
        {
          model: AnotherModel,
          as: 'oneOccurrence'
        },
        {
          model: AnotherModel,
          as: 'anotherOccurrence'
        },
        YetAnotherAssociatedModel
      ]
    };

    Model._validateIncludedElements.call(MyModel, includes);

    const rawQuery= 'SELECT ...';
    const results = await sequelize.query(rawQuery, {
      replacements: { ... },
      model: MyModel,
      hasJoin: true,
      ...includes
    });
```

In `lib/dialects/abstract/query.js` line 292 there's a call to `this.model.bulkBuild` in the scope for queries with includes (`hasJoin === true`) which sets `includeValidated` to `true` and thus skips validating/initialisation, so my suggestion is to simply set it to `false`, or to have it so as a default as both approaches above are invasive to Sequelize's internals, are not obvious and don't really make sense in the app's code. Although I'm not sure about possible side-effects or even how to check for them I can still submit a PR if you'd like.

__Dialect: any
__Sequelize version: 4.37.10
__Tested with latest release:__ No

hkmarques:
### Pull Request check-list

**Not sure what I'm doing wrong but there are tests failing both in mysql and postgres when attempting to run tests from the provided Docker setup (macOS High Sierra)**

### Description of change

Closes #9602

### DIFF
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -271,7 +271,7 @@ class AbstractQuery {
         include: this.options.include,
         includeNames: this.options.includeNames,
         includeMap: this.options.includeMap,
-        includeValidated: true,
+        includeValidated: false,
         attributes: this.options.originalAttributes || this.options.attributes,
         raw: true
       });


sushantdhiman:
Lets discuss this first as `includeValidated: true` is definitely required

PR #9609, opened 2018-06-28
Closes sequelize/sequelize#9608. Added `dialectModule` parameter.

### LINKED ISSUE #9608: Webpack and `require(sequelize.config.dialectModulePath)`
We've enabled AWS X-Ray SDK in our project to benchmark and trace SQL calls.

It requires [this](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-nodejs-sqlclients.html):
```js
var AWSXRay = require('aws-xray-sdk');
var mysql = AWSXRay.captureMySQL(require('mysql2'));
```

So you do have `sequelize.config.dialectModulePath` for a custom (wrapped) `mysql2` instance but it is fully dynamic:

```js
if (sequelize.config.dialectModulePath) {
  require(sequelize.config.dialectModulePath)
}
```

It won't work in Webpack because Webpack won't bundle anything for such require and will only throw "Module not found: mysql2".

Now, we've worked around it in a hacky way by using `string-replace-loader` which is set up to patch the library:

```js
const sequelizeMySQLDialectStuffPath = path.resolve(basePath, 'node_modules/sequelize/lib/dialects/mysql/connection-manager.js');
if (!fs.existsSync(sequelizeMySQLDialectStuffPath)) {
  throw new Error(`[build error] Sequelize MySQL dialect file not found: ${sequelizeMySQLDialectStuffPath}`);
}
const sequelizeMySQLDialectStuff = fs.readFileSync(sequelizeMySQLDialectStuffPath, 'utf-8');
const sequelizeMySQLDialectStuffRequire = 'require(sequelize.config.dialectModulePath)';
if (sequelizeMySQLDialectStuff.indexOf(sequelizeMySQLDialectStuffRequire) < 0) {
  throw new Error(`[build error] Sequelize MySQL dialect file doesn't match.`);
}

{
  test: /\.js$/,
  include: sequelizeMySQLDialectStuffPath,
  use: [{
    loader: 'string-replace-loader',
    options: {
      search: sequelizeMySQLDialectStuffRequire,
      replace: `require("${path.resolve(basePath, '/mysql2-wrapped-with-aws-xray.js').replace(/\\/g, '\\\\')}")`,
    }
  }]
}
```

So as you can see it's hacky and a bit fragile.

You could support something like this instead:

```js
if (sequelize.config.dialectModulePath) {
  require(sequelize.config.dialectModulePath)
} else if (sequelize.config.dialectModule) {
  // use the `sequelize.config.dialectModule` directly.
}
```
And the config would be:
```js
{
  options: {
    dialectModule: require('path/to/custom/dialect.js')
  }
}
```

catamphetamine:
Added `dialectModule` parameter for Webpack.
Closes #9608

### DIFF
--- a/lib/dialects/mssql/connection-manager.js
+++ b/lib/dialects/mssql/connection-manager.js
@@ -19,6 +19,8 @@ class ConnectionManager extends AbstractConnectionManager {
     try {
       if (sequelize.config.dialectModulePath) {
         this.lib = require(sequelize.config.dialectModulePath);
+      } else if (sequelize.config.dialectModule) {
+        this.lib = sequelize.config.dialectModule;
       } else {
         this.lib = require('tedious');
       }
--- a/lib/dialects/mysql/connection-manager.js
+++ b/lib/dialects/mysql/connection-manager.js
@@ -29,6 +29,8 @@ class ConnectionManager extends AbstractConnectionManager {
     try {
       if (sequelize.config.dialectModulePath) {
         this.lib = require(sequelize.config.dialectModulePath);
+      } else if (sequelize.config.dialectModule) {
+        this.lib = sequelize.config.dialectModule;
       } else {
         this.lib = require('mysql2');
       }
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -20,6 +20,8 @@ class ConnectionManager extends AbstractConnectionManager {
       let pgLib;
       if (sequelize.config.dialectModulePath) {
         pgLib = require(sequelize.config.dialectModulePath);
+      } else if (sequelize.config.dialectModule) {
+        pgLib = sequelize.config.dialectModule;
       } else {
         pgLib = require('pg');
       }
--- a/lib/dialects/sqlite/connection-manager.js
+++ b/lib/dialects/sqlite/connection-manager.js
@@ -20,6 +20,8 @@ class ConnectionManager extends AbstractConnectionManager {
     try {
       if (sequelize.config.dialectModulePath) {
         this.lib = require(sequelize.config.dialectModulePath).verbose();
+      } else if (sequelize.config.dialectModule) {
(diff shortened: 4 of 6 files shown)


sushantdhiman:
Going to need some integration test for this, for all dialects

1. Create a new test file `dialect-module-configuration.test.js` https://github.com/sequelize/sequelize/tree/master/test/integration/
2. Test that both `dialectModule` and `dialectModulePath` works.
    1. `dialectModule` can be tested by first requiring module in your test, adding some dummy property to library module and then asserting `sequelize.connectionManager.lib` also got same property after initializing with custom module
    2. `dialectModulePath` can be tested simply by asserting if `new Sequelize` works without throwing `MODULE NOT FOUND` error

catamphetamine:
@sushantdhiman 
I agree that tests are required for such a widely used library.
Added some stubs for tests for now.
Other people can possibly join too.

sushantdhiman:
Please let me know whenever you are willing to continue on this, I will reopen this.

Also there are actually very slim chances that someone will help you write those tests, if you need this feature I am afraid you will need to work on them yourself.

yvele:
@catamphetamine what can I do to help with this?

catamphetamine:
@yvele Don't ask me, ask the author of the library: he doesn't want to merge it in.

sushantdhiman:
@catamphetamine ah, blaming me for something I never did :)

As I said earlier https://github.com/sequelize/sequelize/pull/9609#issuecomment-401034305 I need proper tests before merging this. I can't merge something untested just like that

yvele:
How can I help? @catamphetamine you need to give me access to https://github.com/catamphetamine/sequelize I think 🤔 

@sushantdhiman This issue is important to fix because I can't understand why sequelize is doing the `require`, it should be more simple and explicitly take the whole module (better for mocking, etc.)

papb:
Related: #13169

Issue #8769, opened 2017-12-09
Raw query including "sqlite_master" throws error


voltsu:
## What are you doing?
**Sample code**:
const rawSql = `INSERT INTO ${tableName} (${fieldsNames})
                              SELECT ${escapedValues.join(", ")}
                                   WHERE NOT EXISTS(
                                         SELECT * FROM ${tableName} WHERE id = '${id}'
                                    ) LIMIT 1;`
await sequelize.query(rawSql, { type: sequelize.QueryTypes.UPSERT })

ps: Since we are using fts4 table which has no primary keys, so we have to use "WHERE NOT EXISTS" to avoid duplicated data insertion. Therefore, we form a raw sql query to achieve that.

**The actual executed sql would be sth like:**
INSERT INTO `My_FTS4_TABLE` (id, content)
  SELECT 'uuid-1234', 'sqlite_master'
  WHERE NOT EXISTS(
    SELECT * FROM `My_FTS4_TABLE` WHERE id = 'uuid-1234'
  ) LIMIT 1;

If the content contains "sqlite_master", an error will be throw after the execution (but actually the data is already inserted into the database).

## What do you expect to happen?
Expected to work

## What is actually happening?
Throws an error: TypeError: Cannot read property 'map' of undefined at Database.afterExecute

afterExecute  /sequelize/lib/dialects/sqlite/query.js line #126:
if (self.sql.indexOf('sqlite_master') !== -1) {
   result = results.map(function(resultSet) { return resultSet.name; });
} else if {
  ......
} else if (self.options.type === QueryTypes.UPSERT) {
  result = undefined;
}

It goes into branch "if (self.sql.indexOf('sqlite_master') !== -1)", the result is undefined.

I currently fixed the problem by changing:
  result = results.map(function(resultSet) { return resultSet.name; });
into:
  result = results && results.map(function(resultSet) { return resultSet.name; });

BUT, I am wondering did i make any mistakes on querying raw sql like that. OR, can i simply avoid such error in a way WITHOUT modifying source code. 

__Dialect:__ sqlite
__Sequelize version:__ 3.30.4
__Tested with latest release:__ No 

**Note :** _Your issue may be ignored OR closed by maintainers if it's not tested against latest version OR does not follow issue template._

sushantdhiman:
We dont have enough resources to maintaining both v3 and v4. Please retry with v4 and let me know if this issue persists. v3 will only receive critical security fixes or important features / bugs backport.

Issue #1612, opened 2014-04-09
Query `sqlite_master`


vincentbernat:
Hi!

I would like to test if I get the same tables when running `.sync({force: true})` and when using the migration process. I am using in-memory SQLite tables for that. I thought that I would be able to query `sqlite_master` special table and compare the `sql` column. However, since this table is schema-less, it seems that there is a bit of magic behind it.

`SELECT * FROM sqlite_master` will return names. `SELECT sql FROM sqlite_master` will return `undefined` for each row. `SELECT 1 FROM sqlite_master` will also return `undefined` for each row.

How should this table be queried?

janmeier:
Closing this issue, since it is neither a bug report nor a feature request.

For general sequelize questions, please use [StackOverflow](https://stackoverflow.com/questions/tagged/sequelize.js) or [Google groups](https://groups.google.com/forum/#!forum/sequelize).

PR #9645, opened 2018-07-10
Adds support for querying sqlite_master table

tmont:
### Pull Request check-list

### Description of change
Raw queries to the `sqlite_master` table were being special-cased and as such would return inconsistent values. For example, `SELECT * FROM sqlite_master` would simply return an array populated with just the value in the `name` column. This change allows you to query the `sqlite_master` just like any other table, and removes hardcoded SQL string checks in favor of using the `QueryTypes` enum.

This is related to #1612 and #8769. Previous workarounds were to shove the `sqlite_master` table into a view so that its table name wouldn't be caught by the `indexOf` check and query the view instead.

This changeset also fixes a DataTypes test that was failing due to timezone issues.

### DIFF
--- a/lib/dialects/sqlite/query-interface.js
+++ b/lib/dialects/sqlite/query-interface.js
@@ -153,7 +153,7 @@ function addConstraint(tableName, options) {
   const describeCreateTableSql = this.QueryGenerator.describeCreateTableQuery(tableName);
   let createTableSql;
 
-  return this.sequelize.query(describeCreateTableSql, { ...options, type: QueryTypes.SELECT, raw: true })
+  return this.sequelize.query(describeCreateTableSql, _.assign({}, options, { type: QueryTypes.SELECT, raw: true }))
     .then(constraints => {
       const sql = constraints[0].sql;
       const index = sql.length - 1;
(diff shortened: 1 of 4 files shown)


lib/dialects/sqlite/query-interface.js
```diff
@@ -152,7 +153,7 @@ function addConstraint(tableName, options) {
   const describeCreateTableSql = this.QueryGenerator.describeCreateTableQuery(tableName);
   let createTableSql;
 
-  return this.sequelize.query(describeCreateTableSql, options)
+  return this.sequelize.query(describeCreateTableSql, { type: QueryTypes.SELECT, raw: true })
```
sushantdhiman: We should pass other `options`, probably `{ type: QueryTypes.SELECT, raw: true, ...options }`

tmont: That seemed to make things fail (although I never did try to figure out why). `options` was always something like
```javascript
{ type: 'unique', fields: [ 'email' ] }
```

I believe `options` is meant to be the options to create the constraint, not just general query options. Are there some specific properties I could pick out? I wasn't really sure of what the possibilities were.

tmont:
@sushantdhiman You probably already realized this, but it just occurred to me that this will be a breaking change for any query that involves the string `sqlite_master` for the sqlite dialect. So it probably can't be backported to v4.x (if that was even under consideration).

Issue #9687, opened 2018-07-18
Columns are not mapped on raw query with model


aliismayilov:
## What are you doing?

```js
const Person = sequelize.define('person', {
    firstName: {
      field: 'first_name',
      type: Sequelize.STRING,
      allowNull: true,
      defaultValue: null
    }
  });

const person = await Person.create({ firstName: 'foo' });

console.log(person.firstName);
// => foo
// CORRECT

const [mappedPerson] = await sequelize.query('SELECT * FROM persons', {
  type: sequelize.QueryTypes.SELECT,
  model: Person,
  replacements: { id: person.id }
});

console.log(mappedPerson.firstName);
// => undefined
// WRONG
console.log(mappedPerson.first_name)
// => foo
// WRONG
```

## What do you expect to happen?
`mappedPerson.firstName` prop should be properly mapped.

## What is actually happening?
Mapping of property does not happen.

_Output, either JSON or SQL_

__Dialect:__ mysql / postgres / sqlite / mssql / any
__Dialect version:__ XXX
__Database version:__ XXX
__Sequelize version:__ 4.38.0
__Tested with latest release:__ Yes

**Note :** _Your issue may be ignored OR closed by maintainers if it's not tested against latest version OR does not follow issue template._

aliismayilov:
I digged through the code to try to fix it and found out that it's expected behavior: https://github.com/sequelize/sequelize/blob/702e3659d34fb34f1eb66e152ed5903f356c603e/test/integration/sequelize.test.js#L426-L445

I'll update the docs to reflect this usage clearly, otherwise it's quite confusing to map to an instance without mapping their fields.

PR #9688, opened 2018-07-18
Update raw query docs on model mapping

aliismayilov:
I've been using sequelize for a while already. Thanks for a good work!

I've stumbled upon a confusion when using `sequelize.query()` with `model` option. Not sure what was the maintainers' decision when disabling fields mapping by default, but it is quite confusing to not have mapping when model is passed. Changing default behavior may break, that's why I've updated the docs to clearly indicate the library users that the `mapToModel` property controls whether table fields will be mapped to the model or not.

closes https://github.com/sequelize/sequelize/issues/9687

### DIFF
--- a/docs/raw-queries.md
+++ b/docs/raw-queries.md
@@ -25,9 +25,14 @@ A second option is the model. If you pass a model the returned data will be inst
 
 ```js
 // Callee is the model definition. This allows you to easily map a query to a predefined model
-sequelize.query('SELECT * FROM projects', { model: Projects }).then(projects => {
-  // Each record will now be a instance of Project
-})
+sequelize
+  .query('SELECT * FROM projects', {
+    model: Projects,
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Thanks, this was probably done because fields are already aliased by default, introduced [here](https://github.com/sequelize/sequelize/pull/3995/files)

I'll test this a bit and try to improve so we dont need this option at all, otherwise will merge this PR

sushantdhiman:
Docs point to current stable v4 version

Alfredo-Delgado:
Does that mean there's a different file that needs to be updated? I'm on 4.38.0 and would have still benefitted from accurate documentation.

I've tried looking in https://github.com/sequelize/doc/tree/rtd

Alfredo-Delgado:
@sushantdhiman derp, ok v4 is a branch in this repo 😊 I'll put in a PR. thanks.

Issue #7664, opened 2017-05-20
Problem with Model.update() and custom getter


pensierinmusica:
Hi,

When using `Model.update()` if that model has a custom getter it looks like the `get()` function doesn't have access to all the dataValues of the instance, but only to some, which causes a bug.

Let's take an example:

```js
sequelize.define('User',
  {
    id: {
      primaryKey: true,
      type: DataTypes.UUID,
      defaultValue: DataTypes.UUIDV4
    },
    verified: DataTypes.BOOLEAN(), 
    geo: {
      type: DataTypes.GEOMETRY('POINT'),
      allowNull: false,
      get () {
        return this.getDataValue('geo').coordinates;
      },
      set (coordArr) {
        this.setDataValue('geo', {
          type: 'Point',
          coordinates: coordArr
        });
      }
    }
  }
);
```

When I run this:

```js
await psql.User.update(
  {verified: true},
  {where: {id: userId}}
);
```

I get this error:

```sh
TypeError: Cannot read property 'coordinates' of undefined
```

When I try to log `this.dataValues` inside of the `geo` getter for some reason that I don't understand the `geo` property is not present on the instance.

Can you please help me out?

Thx!

RoboTeddy:
I'm seeing the same issue. Seems like a pretty major bug.

pensierinmusica:
@mickhansen @janmeier @sushantdhiman any feedback?

pensierinmusica:
@mickhansen @janmeier @sushantdhiman any updates on this issue? This is a pretyy anoying bug, any hope to see it fixed?

pensierinmusica:
@mickhansen @janmeier @sushantdhiman any updates on this issue? This is a pretyy anoying bug, any hope to see it fixed?

sushantdhiman:
@pensierinmusica I have already marked this as `bug`, if its really critical to you please send a PR, otherwise wait for someone else to fix it.

LogansUA:
Got the same bug, waiting for fix...

Yousab:
Got a workaround that may be suitable based on your structure. Before a direct return on the getter method, we may check firstly if the variable is undefined or not and then return. This can be done either if you use `getterMethods` option or you are combining the `getter method` inside the field definition itself.

So from my perspective, you can do the following:

```
updated_at: {
        type: Sequelize.DATE,
        defaultValue: Sequelize.fn('now'),
        get: function () {
          if (this.getDataValue('updated_at') instanceof Date) {
[...]

PR #9727, opened 2018-07-26
fixing a bug with Model.update() and custom getter.

u9r52sld:
### Pull Request check-list

### Description of change

fixing a bug with Model.update() and custom getter.

See issue: #7664

### DIFF
--- a/types/test/include.ts
+++ b/types/test/include.ts
@@ -1,6 +1,10 @@
-import { Model, Sequelize } from 'sequelize';
+import { Model, Sequelize, HasMany } from 'sequelize';
 
-class MyModel extends Model {}
+class MyModel extends Model {
+	public static associations: {
+		relation: HasMany
+	};
+}
 
 class AssociatedModel extends Model {}
 
@@ -20,5 +24,15 @@ MyModel.findAll({
 });
 
 MyModel.findAll({
-    include: [{ all: true }],
+	include: [{ all: true }],
+});
+
+MyModel.findAll({
+	include: [{
+		limit: 1,
+		association: 'relation',
+		order: [['id', 'DESC']],
+		separate: true,
+		where: { state: Sequelize.col('project.state') },
+	}]
 });


sushantdhiman:
Fix looks good to me, I have some performance concerns.

Now we are calling `.get(<attr>)` for all attributes, rather than one `.get` call. Other solution could be to fix `.get()` to use getters. I am tagging others for extra reviews

theoephraim:
👍 also running into this issue.

I am doing a simple update and then getting errors because the logic of a getter causes an error because something is undefined. It's not clear to me why the getter would get called in the first place.

u9r52sld:
Thank you for testing.
It's because we should get changed instance variables for subsequent DB update.

I added some updates to avoid calling `get` for all changed attributes. Please check it out.

lib/model.js
```diff
@@ -2884,7 +2884,7 @@ class Model {
         build.set(this._timestampAttributes.updatedAt, values[this._timestampAttributes.updatedAt], { raw: true });
 
         if (options.sideEffects) {
-          values = _.assign(values, _.pick(build.get(), build.changed()));
+          values = _.assign(values, _.pick(build.dataValues, build.changed()));
```
sushantdhiman: `.get()` should be fixed to return correct representation of instance data. What does it return for incorrect case?

u9r52sld: I think that what `.get()` return is not important.
`build.changed()` returns changed `.dataValues` so we should use `.dataValues` in this case.
https://github.com/sequelize/sequelize/blob/d18c2d3d1b9765e10fc804772e28d28773847da1/lib/model.js#L3533

sushantdhiman: `build.changed()` return keys using `dataValues` correct, but we still want `values` to take care of getters. 

What does `build.get()` return for incorrect case?

u9r52sld: > but we still want values to take care of getters

Could you please tell me the reason? 

> What does build.get() return for incorrect case?

Please let me know the intention of your question. I'm sorry that I cannot understand what you mean.

u9r52sld: @sushantdhiman I would appreciate your reply.

sushantdhiman: Hi @u9r52sld sorry for this late reply (as usual :smile:), so I checked original issue and code today, here is what I meant earlier

**Reference attribute definition**

```js
geo: {
  type: DataTypes.GEOMETRY('POINT'),
  allowNull: false,
  get() {
    return this.getDataValue('geo').coordinates;
  },
  set(coordArr) {
    this.setDataValue('geo', {
      type: 'Point',
      coordinates: coordArr
    });
  }
}
```

1. Error appears because `instance.get()` fails
2. `get` fails because user didn't take into account that `getDataValue('field')` can return `undefined` when using `getter` for an attribute
3. `getDataValue('field')` returns `undefined` because `Model.build` didn't set any value, it can;t as user didn't passed any value for `geo` in `update` payload

Originally I meant that `get` should be fixed, but it appears `get` implementation for this attribute is failing, I don;t think we can handle that other than properly pointing out which attribute getter call has failed.

What do you think?

Yousab:
Got a workaround that may be suitable based on your structure. Before a direct return on the getter method, we may check firstly if the variable is undefined or not and then return. This can be done either if you use `getterMethods` option or you are combining the `getter method` inside the field definition itself.

So from my perspective, you can do the following:

```
updated_at: {
        type: Sequelize.DATE,
        defaultValue: Sequelize.fn('now'),
        get: function () {
          if (this.getDataValue('updated_at') instanceof Date) {
            return new Date(this.getDataValue('updated_at')).toISOString();
          }
        }
      }
```

Change the condition to suit your code.

huangyangquang:
you can try it
`exports.updateStudent = async function (id, obj) {
	return await Student.update(obj, {
		where: {
			id
		},
		fields: Object.keys(obj)
	})
}`

from https://sequelize.org/master/class/lib/model.js~Model.html#static-method-update

huangyangquang:
You can try the following code:
`
exports.updateStudent = async function (id, obj) {
	return await Student.update(obj, {
		where: {
			id
		},
		fields: Object.keys(obj)
	})
}
`
The fields field can specify the fields we need to update, so that the get function of all fields will not be called

huangyangquang:
> ### Pull Request check-list
> _Please make sure to review and check all of these items:_
> 
> * [x]  Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
> * [x]  Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
> * [x]  Have you added new tests to prevent regressions?
> * [ ]  Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
> * [x]  Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?
> 
> ### Description of change
> fixing a bug with Model.update() and custom getter.
> 
> See issue: #7664

please try it, maybe can help fix the problem
`exports.updateStudent = async function (id, obj) {
	return await Student.update(obj, {
		where: {
			id
		},
		fields: Object.keys(obj)
	})
}`

PR #9732, opened 2018-07-28
fea(sync): add missing columns in database by default (#9731)

### LINKED ISSUE #9731: sequelize.sync() should add missing columns by default
## What are you doing?
I try to use sequelize.sync() to add missing tables and columns in a non-destructive manner to an existing database.

`sequelize.sync()` should add missing columns without the 'alter' option.

The two main reasons for databases being out of sync are missing tables and missing columns. The current default behaviour of sequelize.sync() without options is to automatically add missing tables to the database, but not missing columns.
Both can be done without the risk of loosing data - so it should be done.

`sequelize.sync({alter: true})` does add missing columns by now, but is too dangerous because of loosing content.
Sequelize-migration is a powerful tool, but in many cases it os overkill.

I will prepare a pull-request.

Missing columns are not added to tables.

__Tested with latest release:__ v4.38.0
__Tested with latest release:__ v5.0.0-beta.10

oger000:
### Pull Request check-list

### Description of change

With this PR `seqzelize.sync()` adds columns that exists in the model, but does not exists on the database without using the risky `seqzelize.sync({alter: true})` variant.

A new testcase is added to proof this behaviour. The test that are failing on master are the same as without this PR.

The documentation is not explicit about the current behaviour of sequelize.sync() in this case now, so I dont know if it need to change.

Closes #9731.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1250,24 +1250,24 @@ class Model {
     })
       .then(() => this.QueryInterface.createTable(this.getTableName(options), attributes, options, this))
       .then(() => {
-        if (options.alter) {
-          return Promise.all([
-            this.QueryInterface.describeTable(this.getTableName(options)),
-            this.QueryInterface.getForeignKeyReferencesForTable(this.getTableName(options))
-          ])
-            .then(tableInfos => {
-              const columns = tableInfos[0];
-              // Use for alter foreign keys
-              const foreignKeyReferences = tableInfos[1];
+        return Promise.all([
+          this.QueryInterface.describeTable(this.getTableName(options)),
+          this.QueryInterface.getForeignKeyReferencesForTable(this.getTableName(options))
+        ])
+          .then(tableInfos => {
+            const columns = tableInfos[0];
+            // Use for alter foreign keys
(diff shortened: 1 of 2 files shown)


sushantdhiman:
Thanks for this PR but this is not an acceptable change, `seqzelize.sync()` behavior will change and looks more like `seqzelize.sync({ alter: true })`.

If any table needs to be changed then `alter: true` is the official way to do so, I agree which is bit unsafe and should not be used in production

Floppy012:
> Thanks for this PR but this is not an acceptable change, `seqzelize.sync()` behavior will change and looks more like `seqzelize.sync({ alter: true })`.
> 
> If any table needs to be changed then `alter: true` is the official way to do so, I agree which is bit unsafe and should not be used in production

@sushantdhiman 

Would a solution with an additional option, that prevents columns from being removed, be acceptable? Something like `sequelize.sync({ alter: true, drop: false })`. It wouldn't change `sync()` or `sync({ alter: true })` behaviour but would provide a solution to make the `alter` feature a bit more safe if required.

oger000:
Yes, { drop: false } would fit may needs. Thanks.

Issue #9766, opened 2018-08-06
Instance.update() ignores previously set() values.


koblas:
## What are you doing?

If you're doing both assignment value setting and setting via the Instance update() method, no data will be written to the DB.

```js
// Init myValue to have the value == 11

const  v = MyModel.findOne({ where: { id: 7 }});

v.myValue = 77;
v.update({
   myValue: 88;
}).then(result => {
    console.log(result.myValue);
})
```

## What do you expect to happen?
I would expect 88 to be output.

## What is actually happening?
```console.log``` will log 11

_Output, either JSON or SQL_

__Dialect:__  any
__Dialect version:__ n/a
__Database version:__ n/a
__Sequelize version:__  4.38.0
__Tested with latest release:__ Yes

Details:

Upon reading the code https://github.com/sequelize/sequelize/blob/master/lib/model.js#L3936 implies that all updates of fields that have previously been set are ignored.  

1) This contract _must_ be documented if preserved
2) This behavior is unexpected and leads to bugs between modules of code using different update semantics.

**Comment by sushantdhiman** [comment]
```js
const Sequelize = require('./index');
const sequelize = require('./test/support').createSequelizeInstance();

const Model = sequelize.define('model', {
    name: Sequelize.STRING,
    value: Sequelize.INTEGER
});


(async function () {
    await Model.sync({ force: true });

    const m = await Model.create({
        name: 'weee',
        value: 23
    });

    console.log(m.value);

    m.value = 11;

    await m.update({ value: 12 });

    console.log(m.value);
})();

```

```
Executing (default): DROP TABLE IF EXISTS `models`;
Executing (default): CREATE TABLE IF NOT EXISTS `models` (`id` INTEGER PRIMARY KEY AUTOINCREMENT, `name` VARCHAR(255), `value` INTEGER, `createdAt` DATETIME NOT NULL, `updatedAt` DATETIME NOT NULL);
Executing (default): PRAGMA INDEX_LIST(`models`)
Executing (default): INSERT INTO `models` (`id`,`name`,`value`,`createdAt`,`updatedAt`) VALUES (NULL,$1,$2,$3,$4);
23
Executing (default): UPDATE `models` SET `value`=$1,`updatedAt`=$2 WHERE `id` = $3
12

```

andrewAtPaymentRails:
Nein! You need a virtual field to reproduce this -->

Add to models.js...

``` js
     it('Update for virtual fields', function() {
[...]

PR #9799, opened 2018-08-14
fix(model): Fix update ignores prev set virtual fields ( #9766)

andrewAtPaymentRails:
### Pull Request check-list

- [ X] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [X ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [X ] Have you added new tests to prevent regressions?
- [X ] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?

### Description of change

Closes #[9766]

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3945,14 +3945,14 @@ class Model {
     setOptions.attributes = options.fields;
 
     // We want to keep track of the fields changed via side effects
-    const pushed_changed = this._changed;
+    const pushedChanged = this.changed();
     this._changed = {};
     this.set(values, setOptions);
 
     // Now we need to figure out which fields were actually affected by the setter.
     const sideEffects = this.changed();
-    this._changed = _.union(Object.keys(pushed_changed), Object.keys(this._changed)).reduce((acc, key) => {
-      acc[key] = pushed_changed[key] || this._changed[key] || false;
+    this._changed = _.union(Object.keys(pushedChanged), Object.keys(this._changed)).reduce((acc, key) => {
+      acc[key] = pushedChanged[key] || this._changed[key] || false;
       return acc;
     }, {});
     const fields = _.union(Object.keys(values), sideEffects);
--- a/lib/model.js
+++ b/lib/model.js
@@ -3937,8 +3937,6 @@ class Model {
     // Clone values so it doesn't get modified for caller scope and ignore undefined values
     values = _.omitBy(values, value => value === undefined);
 
-    const changedBefore = this.changed() || [];
-
     options = options || {};
     if (Array.isArray(options)) options = {fields: options};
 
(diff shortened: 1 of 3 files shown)


lib/model.js
```diff
@@ -3937,18 +3937,24 @@ class Model {
...
-
     options = options || {};
     if (Array.isArray(options)) options = {fields: options};
 
     options = Utils.cloneDeep(options);
     const setOptions = Utils.cloneDeep(options);
     setOptions.attributes = options.fields;
+
+    // We want to keep track of the fields changed via side effects
+    const pushed_changed = this._changed;
```
sushantdhiman: 1. No camel case names

2. Why `_changed` is used here, use `changed()`?

andrewAtPaymentRails: I think you're right: I could use this.changed()

lib/model.js
```diff
@@ -3937,18 +3937,24 @@ class Model {
...
+    this._changed = {};
     this.set(values, setOptions);
 
     // Now we need to figure out which fields were actually affected by the setter.
-    const sideEffects = _.without.apply(this, [this.changed() || []].concat(changedBefore));
+    const sideEffects = this.changed();
```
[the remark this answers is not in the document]
andrewAtPaymentRails: ? I don't know what you mean by "Why don't we send new computed `_changed` to `sideEffects`?" Isn't that what I am doing?

By "Modifying `_changed` directly feels dirty" are you referring this line: "`this._changed = {};`"? I agree that's a bit dirty however I need to reset `this._changed` to an empty set. When `this.set(values, setOptions)` runs it populate the `this._changed`. The problem is `_.without.apply(this, [this.changed() || []].concat(changedBefore));` (which is is really conceptually `this.changed() - the elements in changedBefore`) does the wrong thing. If a key has been `changedBefore` but also modified by `this.set(values, setOptions)` (ie: the user's call to `someDbRecord.update(...);`) it won't get written to the DB. Essentially:

```
// someDbRecord is created with key: "Z" which is a virtual key mapping to "virtualKey"
someDbRecord.key = "A"; // this populates "changedBefore" with "virtualKey" during the update()
await someDbRecord.update({ key: "B" }); // does nothing because it's virtualKey is in changedBefore

someDbRecord.key === "B"
await reload();
someDbRecord.key === "Z" // ! argh update did nothing
```

So the fix is to figure out what changed before, push the list of changedKeys, blank/clear the list of this._changed keys, figure out what changed for the update (that's the side effect list).. The next step is to pop the list of changed keys, which we do by appending the sideEffectFields with the original changedBefore keys (making sure to deduplicate keys). 

I don't know how to do this without blanking the changed list like this: `this._changed = {}` and restoring the changed list like this: this `._changed = _.union(...);`

WikiRik:

Issue #9218, opened 2018-03-22
Unhandled rejection TypeError: Cannot read property 'query' of undefined


gabegorelick:
## What are you doing?
Still trying to reproduce, but I periodically see this error during `Resource Timeout` events:

```
Unhandled rejection TypeError: Cannot read property 'query' of undefined
    at parameters.length.Promise (/sequelize/lib/dialects/postgres/query.js:65:54)
    at Promise._execute (/sequelize/node_modules/bluebird/js/release/debuggability.js:300:9)
    at Promise._resolveFromExecutor (/sequelize/node_modules/bluebird/js/release/promise.js:483:18)
    at new Promise (/sequelize/node_modules/bluebird/js/release/promise.js:79:10)
    at Query.run (/sequelize/lib/dialects/postgres/query.js:65:9)
    at Promise.try.then.connection (/sequelize/lib/sequelize.js:558:20)
    at tryCatcher (/sequelize/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/sequelize/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (/sequelize/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromiseCtx (/sequelize/node_modules/bluebird/js/release/promise.js:606:10)
    at Async._drainQueue (/sequelize/node_modules/bluebird/js/release/async.js:138:12)
    at Async._drainQueues (/sequelize/node_modules/bluebird/js/release/async.js:143:10)
    at Immediate.Async.drainQueues (/sequelize/node_modules/bluebird/js/release/async.js:17:14)
    at runCallback (timers.js:789:20)
    at tryOnImmediate (timers.js:751:5)
    at processImmediate [as _immediateCallback] (timers.js:722:5)
```

This is similar to #8087, but since a fix for that has been merged the error must come from a different source.

## What do you expect to happen?

No error

## What is actually happening?

Error

__Dialect:__ postgres
__Dialect version:__ 9.6
__Database version:__ 9.6
__Sequelize version:__ 4.37.4
__Tested with latest release:__ Yes

gabegorelick:
Tracked the issue down to this: https://github.com/sequelize/sequelize/blob/20cac7a15d2044cac9de9f33ec9b13ab349c5567/lib/sequelize.js#L552-L554

[...]

PR #8330, opened 2017-09-18
fix(abstract/connection-manager): acquire call never finishes

sushantdhiman:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you follow the commit message conventions explained in CONTRIBUTING.md?

### Description of change

This is basically a hack but pooling library has given us no options. Pool acquire call
now resolves even if there is an error. We simply check if its an error instance or not
and reject accordingly.

Read more here https://github.com/sequelize/sequelize/issues/7884#issuecomment-330136439

`_determineConnection` method can be used to check if thing returned by pooling library is
actaully a connection or error

Closes #7884
Closes #7616
Closes #8014
Closes #8133
Closes #8322
Closes #8185
Closes #8098
Closes #8076
Closes #8014

PR #9879, opened 2018-09-05
fix(transaction): fixed unhandled rejection when connection acquiring times out (#9218)

AlexKorn:
### Pull Request check-list

### Description of change

Error described in issue #9218 happens when connection acquiring for auto-callback transaction times out: sequelize tries to automatically rollback transaction, but there is no connection to send rollback command.
Added check for inexistent connection, test and special erorr type.

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -152,11 +152,9 @@ class ConnectionManager {
       acquire: (priority, queryType, useMaster) => {
         useMaster = _.isUndefined(useMaster) ? false : useMaster;
         if (queryType === 'SELECT' && !useMaster) {
-          return this.pool.read.acquire(priority)
-            .then(mayBeConnection => this._determineConnection(mayBeConnection));
+          return this.pool.read.acquire(priority);
         } else {
-          return this.pool.write.acquire(priority)
-            .then(mayBeConnection => this._determineConnection(mayBeConnection));
+          return this.pool.write.acquire(priority);
         }
       },
       destroy: mayBeConnection => {
@@ -274,7 +272,7 @@ class ConnectionManager {
     return promise.then(() => {
       return this.pool.acquire(options.priority, options.type, options.useMaster)
         .then(mayBeConnection => this._determineConnection(mayBeConnection))
-        .catch(err => err.name === 'TimeoutError', err => { throw new errors.ConnectionAcquireTimeoutError(err); })
+        .catch({name: 'TimeoutError'}, err => { throw new errors.ConnectionAcquireTimeoutError(err); })
         .tap(() => { debug('connection acquired'); });
     });
   }
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -4,6 +4,7 @@ const Pooling = require('generic-pool');
 const _ = require('lodash');
 const semver = require('semver');
 const Promise = require('../../promise');
+const errors = require('../../errors');
 const logger = require('../../utils/logger');
 const debug = logger.getLogger().debugContext('pool');
 
@@ -273,6 +274,7 @@ class ConnectionManager {
     return promise.then(() => {
       return this.pool.acquire(options.priority, options.type, options.useMaster)
         .then(mayBeConnection => this._determineConnection(mayBeConnection))
+        .catch(/TimeoutError/, err => { throw new errors.ConnectionAcquireTimeoutError(err); })
         .tap(() => { debug('connection acquired'); });
     });
   }
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -274,7 +274,7 @@ class ConnectionManager {
     return promise.then(() => {
       return this.pool.acquire(options.priority, options.type, options.useMaster)
         .then(mayBeConnection => this._determineConnection(mayBeConnection))
-        .catch(/TimeoutError/, err => { throw new errors.ConnectionAcquireTimeoutError(err); })
+        .catch(err => err.name === 'TimeoutError', err => { throw new errors.ConnectionAcquireTimeoutError(err); })
         .tap(() => { debug('connection acquired'); });
     });
   }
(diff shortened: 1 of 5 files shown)


lib/dialects/abstract/connection-manager.js
```diff
@@ -273,6 +274,7 @@ class ConnectionManager {
     return promise.then(() => {
       return this.pool.acquire(options.priority, options.type, options.useMaster)
         .then(mayBeConnection => this._determineConnection(mayBeConnection))
+        .catch(/TimeoutError/, err => { throw new errors.ConnectionAcquireTimeoutError(err); })
```
sushantdhiman: Throw this error in `_determineConnection`, we also we want to cover cases of replication

sushantdhiman:
Dedicated error for acquire timeout has been talked about before and I think your approach is good.

Could you also work toward an integration test which demonstrate issue with transaction. Limit pool to 1 connection and try to run two transactions. That should be sufficient as a test case. Please also test if similar error which we are trying to resolve in this PR appears by that test without your fix.

lib/dialects/abstract/connection-manager.js
AlexKorn: Maybe I miss your point, but when TimeoutError happens, process doesn't fall in _determineConnection?

sushantdhiman: In case of timeout error `mayBeConnection` should be `TimeError` instance, which can be used to throw `ConnectionAcquireTimeoutError` inside `_determineConnection`

This is important because `_determineConnection` is used by both common pool and replication pool. Otherwise you will need to apply this same check to three different places (where `pool.acquire` is used)

AlexKorn: Timeout error is unfortunately exception from  case  described in #8330 - it is really rejected by pool, so _determineConnection is not called. By the way, is replication poll used directly, bypassing getConnection() method of connection-manager? Search didn't show me such cases

AlexKorn:
fixed pool integration test concerning this issue: not in pretty way but I'm stuck trying to imagine another. The problem is that this unhandled error is so tricky that it bypasses promise catches, so .eventually.rejectedWith doesn't catch it either. Now if test doesn't pass, it just crashes node process

AlexKorn:
The worst about this error that error message is misleading: first line of call stack says that it happens in promise at this line (for postgres):

    new Promise((resolve, reject) => this.client.query(this.sql, (error, result) => error ? reject(error) : resolve(result)));

but in fact it's not the main culprit. The main culprit is line 

    this.sequelize.log('Executing (' + (this.client.uuid || 'default') + '): ' + this.sql, this.options);

where uuid param of inexistent this.client is references, causing process to fail with synchronous error before promise is returned, that's why all .catch()-es doesn't work

lib/dialects/abstract/connection-manager.js
```diff
@@ -273,6 +274,7 @@ class ConnectionManager {
     return promise.then(() => {
       return this.pool.acquire(options.priority, options.type, options.useMaster)
         .then(mayBeConnection => this._determineConnection(mayBeConnection))
+        .catch(err => err.name === 'TimeoutError', err => { throw new errors.ConnectionAcquireTimeoutError(err); })
```
sushantdhiman: Short version `catch('TimeoutError', err => { ... })`

Add this check to https://github.com/sequelize/sequelize/blob/342e18fbaddb74be39d6ea1dd117cb291f1c65e0/lib/dialects/abstract/connection-manager.js#L154 and https://github.com/sequelize/sequelize/blob/342e18fbaddb74be39d6ea1dd117cb291f1c65e0/lib/dialects/abstract/connection-manager.js#L157

test/integration/pool.test.js
```diff
@@ -47,8 +47,8 @@ describe(Support.getTestDialectTeaser('Pooling'), function() {
     this.sinon.stub(this.testInstance.connectionManager, '_connect')
       .returns(new Sequelize.Promise(() => {}));
 
-    return expect(this.testInstance.transaction()
-      .then(() => this.testInstance.transaction()))
-      .to.eventually.be.rejectedWith('ResourceRequest timed out');
+    return expect(this.testInstance.transaction(() => {
+      return this.testInstance.transaction(() => {});
+    })).to.eventually.be.rejectedWith(Sequelize.ConnectionAcquireTimeoutError);
```
sushantdhiman: All ok

lib/dialects/abstract/connection-manager.js
AlexKorn: Don't know whether you've seen my question: it seems that acquire() method of replication pool proxy object is never called directly except from getConnection() method, where all necessary checks are made, so wouldn't be catch() here and moreover check for mayBeConnection redundant?

sushantdhiman: You are right, for both replication and general case, `this.pool.acquire` in `getConnection` represent all calls. Please go ahead and remove `mayBeConnection` check from replication pool proxy's `acquire` case.

Apart from that just simplify `catch('TimeoutError', err => { ... })` and this should be good to go

AlexKorn: Removed checks.
Considering catch() simplification: string didn't worked. It coul be a regex, but then it tries to match with whole stack, therefore fails test for errors in mssql/connection-manager. Found in bluebird docs the way to make it a bit shorter with object predicate.

lib/dialects/abstract/connection-manager.js
```diff
@@ -273,6 +272,7 @@ class ConnectionManager {
     return promise.then(() => {
       return this.pool.acquire(options.priority, options.type, options.useMaster)
         .then(mayBeConnection => this._determineConnection(mayBeConnection))
+        .catch({name: 'TimeoutError'}, err => { throw new errors.ConnectionAcquireTimeoutError(err); })
```
sushantdhiman: We could have used https://github.com/coopernurse/node-pool/blob/master/lib/errors.js#L18 but looks like it is not exported. Not super critical either

sushantdhiman:
Thanks @AlexKorn

lib/dialects/abstract/connection-manager.js
AlexKorn: Yeah, I also wondered whether it was exposed, but not

papandreou:
Do you think this would be backportable to v4?

papandreou:
> Do you think this would be backportable to v4?

Seems to apply cleanly. Here goes nothing: https://github.com/sequelize/sequelize/pull/9889

PR #9881, opened 2018-09-05
docs(hooks.md, model.js): add warnings about memory limitations of individualHooks

jedwards1211:
### Pull Request check-list

**Not applicable, this is a document only change**

### Description of change

I was very disappointed to realize today that if were to update or destroy a billion instances with individual hooks, Sequelize would try to fetch all billion of them into memory before running the individual hooks.  I had mistakenly assumed it would iterate through them in batches or using cursors of some kind...

I wish I had been warned about this before I decided to start using hooks, so I'm adding this warning to the docs.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -212,6 +212,8 @@ afterBulkDestroy(options)
 
 If you want to emit hooks for each individual record, along with the bulk hooks you can pass `individualHooks: true` to the call.
 
+**WARNING**: if you use individual hooks, *all instances that are updated or destroyed will get loaded into memory* before your hooks are called.  The number of instances Sequelize can handle with individual hooks is limited by available memory.
+
 ```js
 Model.destroy({ where: {accessLevel: 0}, individualHooks: true});
 // Will select all records that are about to be deleted and emit before- + after- Destroy on each instance


lib/model.js
```diff
@@ -2664,7 +2664,7 @@ class Model {
    * @param  {Object}       options                         destroy options
```
sushantdhiman: Remove all changes from this file, it is ok to have this as a documentation only warning.

jedwards1211: @sushantdhiman what do you think is the downside to also warning in this file?
The Sequelize documentation is large and this warning could be easily missed (IIRC I learned about `individualHooks` in the reference docs before seeing it in the tutorial docs.)  Having this warning in more places makes it more likely that devs will see it.

sushantdhiman: API docs should not cover such warnings or expanded explanation, they should only explain working of any give option and platform limitation if any

jedwards1211: Okay, would it be okay for the API docs to simply say "all instances will be fetched into memory before running hooks"?  That is explaining what it does...

sushantdhiman: `SELECT all records matching the where parameter...` it should be clear by that. SELECTED records will be loaded to memory and large SELECT surface will consume more memory

jedwards1211: Alright

PR #9889, opened 2018-09-07
fix: Backport #9879 to v4

papandreou:
Cherry-pick of 27e449413d9fbf4a18e2438d0a97dc56fd466ae3 with some trivial `require` conflicts sorted out.

### Pull Request check-list

### Description of change

Backport of #9879 to v4. I'm abusing CI to run the tests, as I don't have a working sequelize test setup locally.

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -6,6 +6,7 @@ const _ = require('lodash');
 const Utils = require('../../utils');
 const debug = Utils.getLogger().debugContext('pool');
 const semver = require('semver');
+const errors = require('../../errors');
 
 const defaultPoolingConfig = {
   max: 5,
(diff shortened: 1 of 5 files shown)


sushantdhiman:
This is a breaking change because error type returned in case of acquire timeout is changed, can't be backported

papandreou:
I don't mind too much as we need to run a fork anyway, but that sounds like an overly strict interpretation of semver. I wouldn't consider a different error class being thrown as enough of a breaking change to delay an important fix to the next major release.

sushantdhiman:
This is an important fix but returning different Error class entirely for some condition is definitely a breaking change. User could be catching old `TimeoutError` and retrying transactions.

There is still a way to make this change non breaking, you can just include `transaction.js` changes which prevent rollback in case of no connection. I'll happily merge that change :)

PR #9911, opened 2018-09-12
Updated the disconnect() method to wait until the connection has been…

### LINKED ISSUE #8468: Connections are not automatically closed
## What are you doing?
Assume `$DATABASE_URL` points to a real Postgres database (tested on 9.4).
```js

function main() {
  const sequelize = new Sequelize(
    process.env.DATABASE_URL,
    {dialect: 'postgres', operatorsAliases: false}
  );

  return sequelize.query(`select 1`);
}

main();
```
Full environment details and demo available at https://github.com/Vadskye/demo-sequelize.

The script should finish its execution 10 seconds after the query finishes, when the connection times out and all active connections have been closed. This was the behavior with Sequelize v3.

The script hangs *indefinitely* after the query finishes.

<details><summary>Script output with DEBUG=*</summary>
```
  sequelize:pool pool created with max/min: 5/0, no replication +0ms
  sequelize:connection:pg connection acquired +0ms
  sequelize:sql:pg executing(default) : SHOW SERVER_VERSION +0ms
  sequelize:sql:pg executed(default) : SHOW SERVER_VERSION +1ms
  sequelize:connection:pg connection timeout +20ms
  sequelize:connection:pg connection acquired +4ms
  sequelize:pool connection acquired +62ms
Executing (default): select 1
  sequelize:sql:pg executing(default) : select 1 +27ms
  sequelize:sql:pg executed(default) : select 1 +1ms
  sequelize:pool connection released +6ms
  sequelize:connection:pg connection timeout +10s
  sequelize:pool connection destroy +10s
```

__Dialect:__ postgres
__Dialect version:__ 6.4.2
__Database version:__ 9.4
__Sequelize version:__ 4.13.8
__Tested with master branch:__ No; there is one commit to master since the last release, and it does not affect this issue.

AlastairTaft:
… ended.

### Pull Request check-list

- [ ] _Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?_ **No, but the same two tests aren't passing that are on master.**

### Description of change

We're using sequelize on AWS lambda with the postgres dialect and the script execution is halted before the connection is closed which causes weird issues with database connections. This change ensures that the connection is properly closed before the execution ends.

Helps with #8468.

### DIFF
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -184,10 +184,7 @@ class ConnectionManager extends AbstractConnectionManager {
   }
 
   disconnect(connection) {
-    return new Promise(resolve => {
-      connection.end()
-      .then(() => resolve());
-    });
+    return Promise.fromCallback(callback => connection.end(callback));
   }
 
   validate(connection) {
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -185,8 +185,8 @@ class ConnectionManager extends AbstractConnectionManager {
 
   disconnect(connection) {
     return new Promise(resolve => {
-      connection.end();
-      resolve();
+      connection.end()
+      .then(() => resolve());
     });
   }
 


lib/dialects/postgres/connection-manager.js
```diff
@@ -185,8 +185,8 @@ class ConnectionManager extends AbstractConnectionManager {
 
   disconnect(connection) {
     return new Promise(resolve => {
-      connection.end();
-      resolve();
+      connection.end()
+      .then(() => resolve());
```
sushantdhiman:
```js
return Promise.fromCallback(callback => connection.end(callback));
```

sushantdhiman:
I don't think this will solve issue you are facing. I really want pool to shutdown when event loop is empty but `node-pool` implementation is such it keep process alive thus preventing clean shutdown.

AlastairTaft:
Do you know if that is on `node-pool/generic-pool`'s roadmap?

PR #9917, opened 2018-09-13
Add section for restore() to tutorial docs

### LINKED ISSUE #9885: [docs] Does creating a new instance with the same primary key after a soft delete work?
I can't tell from the docs what will happen if I soft delete instance `42` on a paranoid mode model, and then later create a new instance with that same key.  Does Sequelize overwrite the fields on the soft deleted row and set `deletedAt` to `null` in that case, or does it throw a duplicate key error?

jedwards1211:
### Pull Request check-list

### Description of change

Also added a note that creating a new instance with the same primary key as a soft-deleted instance will not work.

fix #9885

### DIFF
--- a/docs/instances.md
+++ b/docs/instances.md
@@ -128,6 +128,23 @@ If the `paranoid` options is true, the object will not be deleted, instead the `
 task.destroy({ force: true })
 ```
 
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Thanks

jedwards1211:
Thanks for all your hard work!

PR #9921, opened 2018-09-13
fix(transaction): delete autocommit when transaction begin (#9877)

### LINKED ISSUE #9877: MySQL connections have autocommit disabled after transaction with autocommit: false.
## What are you doing?

```js

async function run() {

    const s = new Sequelize('test', 'user', 'password', {
        host: 'mysql',
        port: '3606',
        dialect: 'mysql',
        pool: {
            max: 1,
            min: 1,
            acquire: 30000,
            idle: 10000,
            log: true
        }
    });

    return s.transaction({autocommit: false}).then((txn) => {
        return s.query('SELECT @@SESSION.autocommit', {transaction: txn}).spread((results, metadata) => {
            console.log(results);
        }).then(() => {
            return txn.commit();
        });
    }).then(() => {
        return s.query('SELECT @@SESSION.autocommit').spread((results, meta) => { console.log(results); });
    }).catch((err) => {
        console.log(err);
    });
}
run();
```

I expect the second SELECT statement to show autocommit=1.

autocommit=0. The 'SET AUTOCOMMIT = 0' is run inside of the transaction, but affects the session. So when the connection goes back into the pool it's in an unexpected state. Unless the application does all reads within an explicit transaction, it can lead to inconsistent reads.

Executing (b74d6416-e1ed-4e88-9a70-47f076e9bfc8): START TRANSACTION;
Executing (b74d6416-e1ed-4e88-9a70-47f076e9bfc8): SET autocommit = 0;
Executing (b74d6416-e1ed-4e88-9a70-47f076e9bfc8): SELECT @@SESSION.autocommit
[ TextRow { '@@SESSION.autocommit': 0 } ]
Executing (b74d6416-e1ed-4e88-9a70-47f076e9bfc8): COMMIT;
Executing (default): SELECT @@SESSION.autocommit
[ TextRow { '@@SESSION.autocommit': 0 } ]

__Dialect:__ mysql
__Dialect version:__ XXX
__Database version:__ 5.6.34
__Sequelize version:__ 4.38.0

tsasaki609:
### Pull Request check-list

### Description of change
fix #9877 
Statements within a transaction should either be all successful, or all fail. Autocommit arguments that initiate transactions are misleading and should not be provided. If we want to enable autocommit, we should explicitly call auto-commit. How do you think?


sushantdhiman:
MySQL transaction with AUTO COMMIT = 1 does defeat purpose of having transaction i.e. each statement is committed as soon as it is finished without any chance of rollback. This is also default behaviour of MySQL.

Two things can be done here

1. Set `autocommit = false` by default, so transactions is predictable
2. Related to fixing #9877 , we need to reset AUTO COMMIT status when transaction exits. How we will do that is not sure, may be a session variable to store current auto commit value and reset it after transaction

tsasaki609:
Since we are executing START TRANSACTION internally, is not it possible to do SET autocommit = 0 to affect the behavior until the end of the transaction?

If we explicitly use a transaction, it seems that it was behavior that autocommit had no effect until you explicitly commit or roll back.

Sorry if my understanding was wrong.

tsasaki609:
@sushantdhiman 
I am glad you understood.

In addition to MySQL, PostgreSQL, SQLite, and MS SQL Server seem to have autocommit statement.
https://www.postgresql.jp/document/9.4/html/ecpg-sql-set-autocommit.html
https://www.sqlite.org/c3ref/get_autocommit.html
https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2008-r2/ms187878(v=sql.105)

However, the essential meaning of the autocommut statement is that any product is similar,
we don't need to think about separately.


I also deleted options and query generation methods.

Thank you for your review.

sushantdhiman:
> In addition to MySQL, PostgreSQL, SQLite, and MS SQL Server seem to have autocommit statement.
https://www.postgresql.jp/document/9.4/html/ecpg-sql-set-autocommit.html
https://www.sqlite.org/c3ref/get_autocommit.html
https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2008-r2/ms187878(v=sql.105)

Yeah I meant more at Sequelize level, where we only support/implemented this for MySQL.

sushantdhiman:
Thanks @tsasaki609

PR #9953, opened 2018-09-24
fix(model): check attribute overrides when defining assoc mixins (#9952)

### LINKED ISSUE #9952: Association mixins attempt to override attributes on model, causing error
## What are you doing?

When I have a model that defines a property whose name matches one of the mixin names that would get generated when I add an association to the model, the model definition and association blows up.

```js
  const Company = sequelize.define("company", {
    hasLocations: {
      type: DataTypes.BOOLEAN
    }
  })
  Company.associate = models => {
    Company.Locations = Company.hasMany(models.location)
  }
  return Company
}

```

I don't want it to blow up and instead want to skip the `hasLocations()` method mixin definition because it conflicts with the `hasLocations` prop

I get the following error

```sh
TypeError: Cannot read property 'hasLocations' of undefined
    at Model.get (<project-root>/node_modules/sequelize/lib/model.js:3305:29)
    at Model.funcs.(anonymous function) (<project-root>/node_modules/sequelize/lib/model.js:1022:25)
    at Object.mixinMethods (<project-root>/node_modules/sequelize/lib/associations/helpers.js:58:13)
    at HasMany.mixin (<project-root>/node_modules/sequelize/lib/associations/has-many.js:155:13)
    at Function.hasMany (<project-root>/node_modules/sequelize/lib/associations/mixin.js:34:17)
```

__Sequelize version:__ 5.0.0-beta.12
__Tested with latest release:__ Yes (If yes, specify that version)

bkoltai:
### Description of change

Closes #9952

### DIFF
--- a/lib/associations/helpers.js
+++ b/lib/associations/helpers.js
@@ -57,7 +57,7 @@ function mixinMethods(association, obj, methods, aliases) {
 
   for (const method of methods) {
     // don't override custom methods
-    if (!obj[association.accessors[method]]) {
+    if (!obj.hasOwnProperty(association.accessors[method])) {
       const realMethod = aliases[method] || method;
 
       obj[association.accessors[method]] = function() {
--- a/test/unit/associations/has-many.test.js
+++ b/test/unit/associations/has-many.test.js
@@ -124,6 +124,16 @@ describe(Support.getTestDialectTeaser('hasMany'), () => {
         expect(user[method]()).to.be.a('function');
       });
     });
+
+    it('should not override attributes', () => {
+      const Project = current.define('Project', {hasTasks: DataTypes.BOOLEAN});
+
+      Project.hasMany(Task);
+
+      const company = Project.build();
+
+      expect(company.hasTasks).not.to.be.a('function');
+    });
   });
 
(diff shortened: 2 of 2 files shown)


lib/associations/helpers.js
```diff
@@ -57,7 +57,7 @@ function mixinMethods(association, obj, methods, aliases) {
 
   for (const method of methods) {
     // don't override custom methods
-    if (!obj[association.accessors[method]]) {
+    if (!obj.hasOwnProperty(association.accessors[method]) && !obj[association.accessors[method]]) {
```
sushantdhiman: Just `!obj.hasOwnProperty(association.accessors[method])` should be enough

bkoltai:
I thought so too but when I had that a bunch of other tests failed. I
believe it has something to do with the dynamic nature of some of the
method calls on models

sushantdhiman:
Just change this, I'll see if i can fix those tests

bkoltai:
Ok, I've updated that line to just use the `hasOwnProperty`

sushantdhiman:
Looks like all tests are passing, merging

PR #9984, opened 2018-10-02
fix(sum): returns 0 when is empty matching

### LINKED ISSUE #6299: .sum on an empty matching set results in NaN
## What you are doing?

I'm trying to use the SUM aggregate on an empty matching set

``` js
//imagine we have Payments table empty yet
Payment.sum('amount').then(console.log)
//logs NaN
```

I'd like the value in this case to be `0` since by definition, a sum is a number, not a NaN.
## Why this happens?

I've quickly found that this happens due to this:

``` js
//node_modules/sequelize/lib/query-interface.js:702
//result contains null as we don't have any matching records
result = parseFloat(result);
```
## What could be done?

since the dataType requested is `dataType instanceof DataTypes.DECIMAL || dataType instanceof DataTypes.FLOAT` as stated in `node_modules/sequelize/lib/query-interface.js:701`, we can additionally check whether the `result` variable  contains `null` and in this case return `0`. I can submit a PR with this fix easily though I'd first like to ensure that I'm not the only one expecting `0` in this case rather than `NaN`.

__Dialect: postgres
__Database version: 9.5.3
__Sequelize version: 3.19.3

t-bonatti:
### Pull Request check-list

### Description of change

Fix #6299

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -1117,7 +1117,9 @@ class QueryInterface {
         const dataType = options.dataType;
 
         if (dataType instanceof DataTypes.DECIMAL || dataType instanceof DataTypes.FLOAT) {
-          result = parseFloat(result);
+          if (!_.isNull(result)) {
+            result = parseFloat(result);
+          }
         } else if (dataType instanceof DataTypes.INTEGER || dataType instanceof DataTypes.BIGINT) {
           result = parseInt(result, 10);
         } else if (dataType instanceof DataTypes.DATE) {
--- a/lib/model.js
+++ b/lib/model.js
@@ -1958,7 +1958,12 @@ class Model {
     Utils.mapOptionFieldNames(options, this);
     options = this._paranoidClause(this, options);
 
-    return this.QueryInterface.rawSelect(this.getTableName(options), options, aggregateFunction, this);
+    return this.QueryInterface.rawSelect(this.getTableName(options), options, aggregateFunction, this).then( value => {
+      if (value === null) {
+        return 0;
+      }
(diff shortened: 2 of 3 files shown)


t-bonatti:
@sushantdhiman Could you take a look at this fix too?

t-bonatti:
@sushantdhiman  

I changed to return 0.
I returned `null` before because the Postgres do that.

image

lib/query-interface.js
```diff
@@ -1117,7 +1117,9 @@ class QueryInterface {
         const dataType = options.dataType;
 
         if (dataType instanceof DataTypes.DECIMAL || dataType instanceof DataTypes.FLOAT) {
-          result = parseFloat(result);
+          if (!_.isNull(result)) {
```
t-bonatti: I kept this change to not return `NaN`

sushantdhiman: :+1:

sushantdhiman:
Thanks @t-bonatti 


> I returned null before because the Postgres do that.

Correct behaviour if we want to know if record exists as per our condition, but I can't say other dialect adheres to this. We can simply return numerical values from `aggregate` for now

mariomixtega:
Is this already on 4.42.0?

holm:
I am just upgrading to Sequelize 5, and this caught me by surprise. It seems very strange that this method now returns 0 if the result is null. This seems to assume that the result of an aggregation is a number. This is not at all the case in general. It can really be any datatype, and by returning 0 as a default, this breaks the expected type. In our case we are using it to find the max of a date field.

PR #10001, opened 2018-10-06
fix(query-generator): syntax error in Sqlite with TINYINT and AUTOINCREMENT primary key

### LINKED ISSUE #9995: Sqlite dialect ERROR on AUTOINCREMENT id with datatype different then INTEGER.
## What are you doing?

The following model definition raises an error when using Sqlite dialect:

```
sequelize.define('User', {
        id: {
            type: DataTypes.SMALLINT,
            allowNull: false,
            primaryKey: true,
            autoIncrement: true
        },
        name: {
            type: DataTypes.STRING(25),
            allowNull: false,
        }, {});
```
while this is working properly: 
```
...       
    id: {
            type: DataTypes.INTEGER(4),
            allowNull: false,
            primaryKey: true,
            autoIncrement: true
        },
...
```
It should work both with **INTEGER(length)** and **SMALLINT**.

I got the following error:

    'SQLITE_ERROR: AUTOINCREMENT is only allowed on an INTEGER PRIMARY KEY'

__Dialect:__ mysql (production)/ sqlite (development)
__Dialect version:__ 4.0.2
__Sequelize version:__ 4.39.0

ugogiordano87:
### Pull Request check-list

### Description of change

Fixes #9995


ugogiordano87:
The first rule of the [column affinity rules](https://www.sqlite.org/datatype3.html#determination_of_column_affinity) specified in the SQLite documentation says:
      
> If the declared type contains the string "INT" then it is assigned INTEGER affinity.

So SQLite ignores the **UNSIGNED** keyword when converting the datatypes into affinities. 

Only the **ZEROFILL** option does't seem to be supported by SQLite.

sushantdhiman:
Is `DATATYPE UNSIGNED` actually an unsigned column? I think for SQLite it will accept any integer values you throw at it.

So `INT UNSIGNED` should accept `1` but rejected `-1`, but I think SQLite accepts both. If that is the case we can just remove UNSIGNED option for SQLite

ugogiordano87:
Well you're right, it seems that SQLite accept both. So we can remove the UNSIGNED options, maybe with a simple function like this:

```
function removeUnsupportedOptions(dataType) {
    if (dataType._zerofill || dataType._unsigned) {
      warn('SQLite does not support ' + dataType._name + '  with UNSIGNED or ZEROFILL. Plain `' + dataType._name + '` will be used instead.');
      dataType._unsigned = undefined;
      dataType._zerofill = undefined;
    }
  }
```

called inside each INT datatype.

sushantdhiman:
@ugogiordano87 Looks good

lib/dialects/postgres/data-types.js
```diff
@@ -7,6 +7,20 @@ const inherits = require('../../utils/inherits');
 module.exports = BaseTypes => {
   const warn = BaseTypes.ABSTRACT.warn.bind(undefined, 'http://www.postgresql.org/docs/9.4/static/datatype.html');
 
+  /**
+   * Removes unsupported Postgres options, i.e., LENGTH, UNSIGNED and ZEROFILL, for the integer data types.
+   * @param dataType The base integer data type.
+   */
+  function removeUnsupportedOptions(dataType) {
+    if (this._length || this.options.length || this._unsigned || this._zerofill) {
+      warn('SQLite does not support ' + dataType._name + '  with LENGTH, UNSIGNED or ZEROFILL. Plain `' + dataType._name + '` will be used instead.');
```
sushantdhiman: `PostgreSQL does not support <typename> with options. Plain '<typename>' will be used instead.`

ugogiordano87: Yes I see. What do you think if I define an util file like this:

```
'use strict';

/**
 * Removes unsupported options, i.e., LENGTH, UNSIGNED and ZEROFILL, for the integer data types.
 *
 * @param dialect The specific SQL dialect
 * @param dataType The base integer data type.
 * @param warn The dialect specific warn function.
 */
function removeUnsupportedOptions(dialect, dataType, warn) {
  let warnMessage = '';

  if (this._length || this.options.length || this._unsigned || this._zerofill) {
    this._unsigned = undefined;
    this._zerofill = undefined;

    if (dialect === 'sqlite') {
      warnMessage = 'SQLite does not support ' + dataType._name + '  with UNSIGNED or ZEROFILL. ';
    } else if (dialect === 'postgres') {
      warnMessage = 'PostgresSQL does not support ' + dataType._name + '  with LENGTH, UNSIGNED or ZEROFILL. ';
      this._length = undefined;
      this.options.length = undefined;
    }
    
    warn(warnMessage + 'Plain `' + dataType._name + '` will be used instead.');
  }
}


module.exports = removeUnsupportedOptions;
```

sushantdhiman: No that will overtime become complex to handle different dialect and their particularities. Keep these functions in `data-types.js` of those dialects, so they can handle dialect specific changes

Also rename `removeUnsupportedOptions` => `removeUnsupportedIntegerOptions`

ugogiordano87: Sorry, dataType instead of this.

ugogiordano87: Done.

sushantdhiman:
Thanks @ugogiordano87

ugogiordano87:
You're welcome @sushantdhiman

PR #10003, opened 2018-10-07
fix(syntax): correct parentheses around union (#9813)

### LINKED ISSUE #9813: Sqlite syntax error
## What are you doing?
I have a set of models/associations below that generate a query that produces a SQL error (testing with sqlite). This is as far as I could boil it down and still produce the error.

```js
'use strict';

console.log(Sequelize.version);

const sequelize = new Sequelize({ dialect: 'sqlite', storage: ':memory:', operatorsAliases: false, logging: console.log });

const Project = sequelize.define('Project', { name: Sequelize.STRING });
const LevelTwo = sequelize.define('LevelTwo', { name: Sequelize.STRING });
const LevelThree = sequelize.define('LevelThree', { type: Sequelize.INTEGER });

Project.hasMany(LevelTwo);
LevelTwo.belongsTo(Project);

LevelTwo.hasMany(LevelThree, { as: 'type_ones' });
LevelTwo.hasMany(LevelThree, { as: 'type_twos' });
LevelThree.belongsTo(LevelTwo);

function callQuery() {
    return Project.find({
        where: { id: 1 },
        include: [
            {
                model: LevelTwo,
                include: [
                    {
                        model: LevelThree,
                        as: 'type_ones',
                        where: { type: 0 },
                        separate: true,
                        limit: 1,
                        order: [['createdAt', 'DESC']]
                    },
                    {
                        model: LevelThree,
                        as: 'type_twos',
                        where: { type: 1 },
                        separate: true,
                        limit: 1,
                        order: [['createdAt', 'DESC']]
                    }
                ]
            }
        ]
    });
}

sequelize.sync()
    .then(function () {
        return Sequelize.Promise.all([
            Project.create({ name: 'testProject' }),
            LevelTwo.create({ name: 'testL2-1' }),
            LevelTwo.create({ name: 'testL2-2' })
        ]);
    })
    .bind({})
    .spread(function (project, L2_1, L2_2) {
        this.project = project;
        this.L2_1 = L2_1;
        this.L2_2 = L2_2;

        return this.project.addLevelTwo(this.L2_1);
    })
    .then(function () {
        return callQuery();
    })
    .then(function (result) {
        console.log(result.get({ plain: true }));

        return this.project.addLevelTwo(this.L2_2);
    })
    .then(function () {
        return callQuery();
    })
    .then(function (result) {
        console.log(result.get({ plain: true }));
    });
```

The above is just a base find query. The part I assume is a little more rare is I'm requesting the same table twice but the queries differ specifying the the `type` column of the table. I'm simply expecting the query to function as constructed.

The output on the second call to callQuery is the following error. If the `limit: 1` option is removed from the `LevelThree` model include object, the error goes away. I'm not sure exactly why that is, but maybe it will help (and is my current work around).
```
Unhandled rejection SequelizeDatabaseError: SQLITE_ERROR: near "UNION": syntax error
```

__Dialect:__ sqlite (only dialect tested)
__Dialect version:__ 4.0.2 (npm sqlite3 package)
__Sequelize version:__ 4.38.0
__Tested with latest release:__ Yes, 4.38.0
__Node.js version:__ 6.14.2

Thanks for the help and the great work on Sequelize.

tsasaki609:
### Pull Request check-list

### Description of change

fix #9813 

I wrapped a subquery to union.
Although it is a little tricky, no error occurs with this.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1271,7 +1271,7 @@ class QueryGenerator {
 
         // Caching the base query and splicing the where part into it is consistently > twice
         // as fast than generating from scratch each time for values.length >= 5
-        const baseQuery = '(' + this.selectQuery(
+        const baseQuery = 'SELECT * FROM (' + this.selectQuery(
           tableName,
           {
             attributes: options.attributes,
@@ -1283,7 +1283,7 @@ class QueryGenerator {
             model
           },
           model
-        ).replace(/;$/, '') + ')';
+        ).replace(/;$/, '') + ') AS sub'; // Every derived table must have its own alias
         const placeHolder = this.whereItemQuery(Op.placeholder, true, { model });
         const splicePos = baseQuery.indexOf(placeHolder);
 
--- a/test/integration/model/findAll/separate.test.js
+++ b/test/integration/model/findAll/separate.test.js
@@ -1,0 +1,103 @@
+'use strict';
+
+const chai = require('chai');
+const expect = chai.expect;
+const Support = require('../../support');
+const DataTypes = require('../../../../lib/data-types');
+const Sequelize = require('../../../../lib/sequelize');
+const current = Support.sequelize;
+
+describe(Support.getTestDialectTeaser('Model'), () => {
+  describe('findAll', () => {
+    describe('separate with limit', () => {
+      it('should not throw syntax error (union)', () => {
+        // #9813 testcase
+        const Project = current.define('Project', { name: DataTypes.STRING });
+        const LevelTwo = current.define('LevelTwo', { name: DataTypes.STRING });
+        const LevelThree = current.define('LevelThree', { type: DataTypes.INTEGER });
+
+        Project.hasMany(LevelTwo);
+        LevelTwo.belongsTo(Project);
+
+        LevelTwo.hasMany(LevelThree, { as: 'type_ones' });
+        LevelTwo.hasMany(LevelThree, { as: 'type_twos' });
+        LevelThree.belongsTo(LevelTwo);
+
+        return current.sync({ force: true }).then(() => {
+          return Sequelize.Promise.all([
+            Project.create({ name: 'testProject' }),
+            LevelTwo.create({ name: 'testL21' }),
+            LevelTwo.create({ name: 'testL22' })
+          ]);
+        }).spread((project, level21, level22) => {
+          return Sequelize.Promise.all([
+            project.addLevelTwo(level21),
+            project.addLevelTwo(level22)
+          ]);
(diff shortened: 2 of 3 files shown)


test/integration/model/findAll/separate.test.js
```diff
@@ -0,0 +1,63 @@
...
+const expect = chai.expect;
+const Support = require('../../support');
+const DataTypes = require('../../../../lib/data-types');
+const Sequelize = require('../../../../lib/sequelize');
+const current = Support.sequelize;
+
+describe(Support.getTestDialectTeaser('Model'), () => {
+  describe('findAll', () => {
+    describe('separate with limit', () => {
+      it('should not occar syntax error (union)', () => {
```
sushantdhiman: `should not throw syntax error (union)`

lib/dialects/abstract/query-generator.js
```diff
@@ -1287,7 +1287,7 @@ class QueryGenerator {
             model
           },
           model
-        ).replace(/;$/, '') + ')';
+        ).replace(/;$/, '') + ') AS sub'; // Every derived table must have its own alias
```
sushantdhiman: What if multiple includes with `separate:true` are used, wont having same `sub` alias throw an error? May be they are executed as separate statement so it will work.

Please extend your test case to demonstrate working with two includes with `separate: true` in same `findAll`

test/integration/model/findAll/separate.test.js
tsasaki609: Oh... I see.

tsasaki609:
How is it like this?

test/integration/model/findAll/separate.test.js
```diff
@@ -0,0 +1,95 @@
...
+                    where: { type: 0 },
+                    separate: true,
+                    limit: 1,
+                    order: [['createdAt', 'DESC']]
+                  }
+                ]
+              }
+            ]
+          });
+        }).then(() => {
```
sushantdhiman: Please assert on incoming result, like only one item from LevelThree is selected. Same goes for next `findAll` call

sushantdhiman:
Thanks @tsasaki609 

Not sure when you last pushed the change, I didn't get any notification for last commit

internalfx:
Is this in the latest release yet? I'm still seeing this bug.

sushantdhiman:
@internalfx `5.0.0-beta.13`

internalfx:
Any chance for a 4.x branch release?

sushantdhiman:
@internalfx cherry pick this commit and send a PR targeting v4 branch

tzellman:
Thanks @sushantdhiman! PR #10121 targets the `v4` branch.

Issue #8820, opened 2017-12-28
There are no docs for QueryInterface.bulkUpdate()


ataujenis:
## What are you doing?
I'm trying to find the docs for `QueryInterface.bulkUpdate()`.

## What do you expect to happen?
I see this is an available public method in the source code and I would expect to find docs: https://github.com/sequelize/sequelize/blob/master/lib/query-interface.js#L967

## What is actually happening?
I can't find any documentation on the `QueryInterface.bulkUpdate()` method: http://docs.sequelizejs.com/class/lib/query-interface.js~QueryInterface.html

__Tested with latest release__

sushantdhiman:
I think you meant `bulkUpdate` (your link point to `bulkInsert`) https://github.com/sequelize/sequelize/blob/bb2d0bd9c1765940024974a6de3e05d07830a388/lib/query-interface.js#L1003

Tagging as documentation

ataujenis:
@sushantdhiman That link I provided was to a specific line-number in the source file. It looks like that source file was updated 4 hours ago and changed the line positions. You are correct: the link should point to the `bulkUpdate` method.

cyrilchapon:
Any update on this ?

Even on update method ?

cyrilchapon:
related https://github.com/sequelize/sequelize/issues/8687

Songkeys:
Any update after half a year...?

ptimson:
+1

chaim1221:
💀

sushantdhiman:
Fixed by https://github.com/sequelize/sequelize/pull/10005

PR #10005, opened 2018-10-08
docs(query-interface): add inline 'bulkUpdate' docs

c-moss-talk:
Adds missing documentation on 'bulkUpdate' inline at query-interface.js. Addresses https://github.com/sequelize/sequelize/issues/8820

### DIFF
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -971,6 +971,25 @@ class QueryInterface {
     return this.sequelize.query(sql, options);
   }
 
+  /**
+   * Update multiple records of a table
+   *
+   * @example
+   * queryInterface.bulkUpdate('roles', {
+   *     label: 'admin',
+   *   }, { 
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Thanks @c-moss-talk

jeanmarcprevost:
I don't understand the documentation...
```@param {Object} identifier    A hash with conditions OR an ID as integer OR a string with conditions```

**Comment by sushantdhiman** [comment]
`where` conditions

```js
{
   name: "ABC",
   [Op.and]: [ .... ],
   ....
}
```

c-moss-talk:
Yeah, that's a hash example, but actually I tried to use and ID as integer and it resulted in unexpected behavior. I want to look into it.

jeanmarcprevost:
Oh yes thank @sushantdhiman !

PR #10024, opened 2018-10-11
Add support for postgres citext data-type

### LINKED ISSUE #6572: Sequelize omitting field when casting to array
I am attempting to use the postgres citext module with Sequelize, so far it has worked well by using literals for queries and defining the type in the model. However because Sequelize does not understand what the citext type is, it returns my citext arrays as strings. e;g "{item 1, item2}"   
To get around this I am trying to cast the citext array to a TEXT or VARCHAR array in the query, however when I do this, Sequelize does not show that field at all.    
   
If I set raw: true, the field shows fine, and as an array.

```js
Users.findAndCountAll({
        where: { },
        attributes: [
          'id',
          'createdAt',
          'updatedAt',
          db.cast(db.col('nicknames'), 'text[]')
        ]
})
```

I expect there to be a nicknames field returned from sequelize and it to be an array, instead it's as if the attribute is not there at all.

This is the generated SQL, the SQL itself works fine when I try it in a database manager, but the output Sequelize gives me omits the nickname field.

SELECT "User"."id", "User"."createdAt", "User"."updatedAt", CAST("nicknames" AS TEXT[]) FROM "Users" AS "User" WHERE "User"."deletedAt" IS NULL;

__Dialect:__ postgres
__Database version:__ 9.5
__Sequelize version:__ 3.23.4

### LINKED ISSUE #9116: Possibility of having an ARRAY datatype with CITEXT fields
## What are you doing?

I am trying to add a custom datatype for CITEXT as part of an ARRAY. I search the database with case insensitive because I don't control the search terms. Is this possible to do with sequelize?

I tried this:

```js
export default (sequelize, DataTypes) => {
  class CITEXT extends DataTypes.TEXT {
    constructor () {
      super()
      this.key = 'CITEXT'
    }
  }

  const Model = sequelize.define('artist', {
    id: {
      allowNull: false,
      autoIncrement: true,
      primaryKey: true,
      type: DataTypes.INTEGER
    },
    slug: {
      allowNull: false,
      type: 'CITEXT' // <-- here I use a custom class, this works
    },
    names: {
      allowNull: false,
      type: DataTypes.ARRAY(CITEXT) // <-- but this does not work
    }
  });

  return Model;
};
```

That I could access the ARRAY in sequelize like a normal JS array.

This is what I got in the console (which looks correct)

```sql
CREATE TABLE IF NOT EXISTS "artists" ("id" SERIAL, "slug" CITEXT NOT NULL, "names" CITEXT[] NOT NULL, "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL, "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL, PRIMARY KEY ("id"));
```

But this is what I get in the JS code

```js
sequelize.sync({ force: true }).then(() => {
  return models.artist.create(
    slug: 'cali-dubuque',
    names: ['Cali DuBuque']
  }).then(artist => {
    console.log('===> artist.names', artist.get('names'))
    // ===> artist.names {"Cali DuBuque"}
  });
});
```

`artist.get('names')` should output an array `['Cali DuBuque']` instead of a string `'{"Cali DuBuque"}'`

__Dialect:__ postgres
__Dialect version:__ 7.4.1 (node package `pg`)
__Database version:__ 10.2 (`psql --version`)
__Sequelize version:__ 4.34.1

justinkalland:
### Pull Request check-list

### Description of change

Adds support for CITEXT data-type. This is a Postgres only data-type. This also adds support for ARRAY(CITEXT).

To use CITEXT it must first be enabled in Postgres:

```sql
CREATE EXTENSION IF NOT EXISTS citext;
```

__The new tests will fail unless the module is enabled in Postgres.__ I have created as change to enable the module in the custom Docker image: sushantdhiman/sequelize-postgres#1

Closes #9116
Closes #6572

### DIFF
--- a/lib/dialects/abstract/index.js
+++ b/lib/dialects/abstract/index.js
@@ -11,6 +11,7 @@ AbstractDialect.prototype.supports = {
   'ORDER NULLS': false,
   'UNION': true,
   'UNION ALL': true,
+  'CITEXT': false,
 
   /* does the dialect support returning values for inserted/updated fields */
   returnValues: false,
--- a/lib/dialects/abstract/index.js
+++ b/lib/dialects/abstract/index.js
@@ -11,7 +11,6 @@ AbstractDialect.prototype.supports = {
   'ORDER NULLS': false,
   'UNION': true,
   'UNION ALL': true,
-  'CITEXT': false,
 
   /* does the dialect support returning values for inserted/updated fields */
   returnValues: false,
--- a/lib/dialects/abstract/index.js
+++ b/lib/dialects/abstract/index.js
@@ -57,7 +57,8 @@ AbstractDialect.prototype.supports = {
     parser: false,
     concurrently: false,
     type: false,
-    using: true
+    using: true,
+    functionBased: false
   },
   joinTableDependent: true,
   groupedLimit: true,
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -219,10 +219,10 @@ class ConnectionManager extends AbstractConnectionManager {
 
       // Reset OID mapping for dynamic type
       [
-        dataTypes.postgres.GEOMETRY,
-        dataTypes.postgres.HSTORE,
-        dataTypes.postgres.GEOGRAPHY,
-        dataTypes.postgres.ENUM,
+        dataTypes.GEOMETRY,
+        dataTypes.HSTORE,
+        dataTypes.GEOGRAPHY,
+        dataTypes.ENUM,
         dataTypes.CITEXT
       ].forEach(type => {
         type.types.postgres.oids = [];
@@ -233,15 +233,15 @@ class ConnectionManager extends AbstractConnectionManager {
         let type;
 
         if (row.typname === 'geometry') {
-          type = dataTypes.postgres.GEOMETRY;
+          type = dataTypes.GEOMETRY;
         } else if (row.typname === 'hstore') {
-          type = dataTypes.postgres.HSTORE;
+          type = dataTypes.HSTORE;
         } else if (row.typname === 'geography') {
-          type = dataTypes.postgres.GEOGRAPHY;
+          type = dataTypes.GEOGRAPHY;
         } else if (row.typname === 'citext') {
           type = dataTypes.CITEXT;
         } else if (row.typtype === 'e') {
-          type = dataTypes.postgres.ENUM;
+          type = dataTypes.ENUM;
         }
 
         type.types.postgres.oids.push(row.oid);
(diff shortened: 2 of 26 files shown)


lib/dialects/abstract/index.js
```diff
@@ -11,6 +11,7 @@ AbstractDialect.prototype.supports = {
   'ORDER NULLS': false,
   'UNION': true,
   'UNION ALL': true,
+  'CITEXT': false,
```
sushantdhiman: No need for this in `supports`, these flags are meant to switch code segments when generating queries or declaring support information that is used anywhere in `lib/`, for tests you can just compare `dialect === 'postgres'`

sushantdhiman:
There is some issue with postgres-9.5 image, I'll fix that is evening (approx 7-8 hours)

justinkalland:
Discovered that CITEXT has dynamic OIDs and changed to handle that.

lib/dialects/postgres/connection-manager.js
```diff
@@ -221,7 +222,8 @@ class ConnectionManager extends AbstractConnectionManager {
         dataTypes.postgres.GEOMETRY,
         dataTypes.postgres.HSTORE,
         dataTypes.postgres.GEOGRAPHY,
-        dataTypes.postgres.ENUM
+        dataTypes.postgres.ENUM,
+        dataTypes.CITEXT
```
sushantdhiman: `dataTypes.postgres.CITEXT` is not available, right?

As we need to dynamically update OIDs it is better to define this type in `postgres/data-types.js`

```js
  function CITEXT(options) {
    if (!(this instanceof CITEXT)) return new CITEXT(options);
    BaseTypes.CITEXT.apply(this, arguments);
  }
  inherits(CITEXT, BaseTypes.CITEXT);

  BaseTypes.CITEXT.types.postgres = {
    oids: [],
    array_oids: []
  };
```

and have it exported. So you can use `dataTypes.postgres.CITEXT` here and in code just below

justinkalland: Perhaps it is bit much to redeclare in the Postgres data-types? If you are concerned about how it looks, this would be cleaner:

```js
      // Reset OID mapping for dynamic type
      [
        dataTypes.GEOMETRY,
        dataTypes.HSTORE,
        dataTypes.GEOGRAPHY,
        dataTypes.ENUM,
        dataTypes.CITEXT
      ].forEach(type => {
        type.types.postgres.oids = [];
        type.types.postgres.array_oids = [];
      });

      for (const row of result.rows) {
        let type;

        if (row.typname === 'geometry') {
          type = dataTypes.GEOMETRY;
        } else if (row.typname === 'hstore') {
          type = dataTypes.HSTORE;
        } else if (row.typname === 'geography') {
          type = dataTypes.GEOGRAPHY;
        } else if (row.typname === 'citext') {
          type = dataTypes.CITEXT;
        } else if (row.typtype === 'e') {
          type = dataTypes.ENUM;
        }
```

sushantdhiman:
Thanks @justinkalland

mjdavidson:
when is this being released?

PR #10031, opened 2018-10-15
fix: cherry-pick memory leak perf commit (#9973)

uneess:
Issue: https://github.com/sequelize/sequelize/pull/9973

cherry-picked commit: https://github.com/sequelize/sequelize/commit/f7585ac08cf8a44a535ae3b9c42776f9fd4bc22e

### DIFF
--- a/lib/utils/validator-extras.js
+++ b/lib/utils/validator-extras.js
@@ -64,22 +64,14 @@ const extensions = {
 };
 exports.extensions = extensions;
 
-function extendModelValidations(modelInstance) {
-  const extensions = {
-    isImmutable(str, param, field) {
-      return modelInstance.isNewRecord || modelInstance.dataValues[field] === modelInstance._previousDataValues[field];
-    }
-  };
+/** Instance based validators */
+validator.isImmutable = function(value, validatorArgs, field, modelInstance) {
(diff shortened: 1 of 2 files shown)


uneess:
`validator.notNull` is not covered by the unit test, I also see that it is being deprecated, so to keep things consistent I will rollback to the old version in 4.x.x where it throws an exception:

`validator.notNull = function() {
  throw new Error('Warning "notNull" validation has been deprecated in favor of Schema based "allowNull"');
};`

and make sure to add it in the unit test so it will be covered.

What do you think @sushantdhiman ?

lib/utils/validator-extras.js
```diff
@@ -64,22 +64,14 @@ const extensions = {
...
-}
-exports.extendModelValidations = extendModelValidations;
+/** Instance based validators */
+validator.isImmutable = function(value, validatorArgs, field, modelInstance) {
+  return modelInstance.isNewRecord || modelInstance.dataValues[field] === modelInstance._previousDataValues[field];
+};
 
-// Deprecate this.
-validator.notNull = function() {
-  throw new Error('Warning "notNull" validation has been deprecated in favor of Schema based "allowNull"');
```
sushantdhiman: Remove this change

PR #10035, opened 2018-10-16
Add support for temporary tables in all dialects

justinkalland:
### Pull Request check-list

### Description of change

Added support for temporary tables. This is supported by all dialects. A temporary table is only available for the current session (connection). Therefore it is only supported if pooling is disabled (by setting max pool size to 1).

The new `temporaryTable` option can be set when defining a model or using `queryInterface.createTable()`. If pooling is enabled then an error is thrown.

One use case for this is simplifying integration tests in apps that use sequelize. Another use case is when creating migrations that move data around.

### DIFF
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -132,7 +132,7 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
           const commentMatch = dataType.match(/^(.+) (COMMENT.*)$/);
           const commentText = commentMatch[2].replace(/COMMENT/, '').trim();
           commentStr += _.template(commentTemplate, this._templateSettings)({
-            table: this.quoteIdentifier(tableName),
+            table: this.quoteIdentifier(options.temporaryTable ? `##${tableName}` : tableName),
             comment: this.escape(commentText),
             column: this.quoteIdentifier(attr)
           });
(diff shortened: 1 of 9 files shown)


justinkalland:
Having some trouble testing on the MSSQL dialect. Anyone with experience in MSSQL care to take a look?

lib/dialects/mssql/query-generator.js
```diff
@@ -132,7 +132,7 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
           const commentMatch = dataType.match(/^(.+) (COMMENT.*)$/);
           const commentText = commentMatch[2].replace(/COMMENT/, '').trim();
           commentStr += _.template(commentTemplate, this._templateSettings)({
-            table: this.quoteIdentifier(tableName),
+            table: this.quoteIdentifier(options.temporaryTable ? `##${tableName}` : tableName),
```
sushantdhiman: `##` is a global temporary table which is shared by all sessions and is available till last session using it. We should go for `#` local temporary table which acts like temporary table of other dialects i.e. limited to current session

justinkalland: Thanks, looks like I need to dig a little more to come up with the best way to handle MSSQL. Two tests failed and I am having a hard time getting the docker image for it to run.

sushantdhiman: Docker image will for anything transaction (unfortunately I don't have any real experience with MSSQL either), you can run failing tests with `DIALECT=mssql npm run test-docker-integration -- --grep test-title`

justinkalland:
@sushantdhiman I was also on the fence about submitting a pull request for this. Needing to have pooling disabled feels clunky. Perhaps I should remove the documentation from `model.js` and just leave its as a more advanced option when using the `QueryInterface` directly. I also wouldn't be hurt just to scrap the whole idea.

sushantdhiman:
@justinkalland Yeah I agree, we can rework this when we identify a good use case.

Going to close this PR, Thanks for this initiative :+1:

Issue #9079, opened 2018-02-20
[Error] You've included an alias {...} but it does not match the alias defined in your association


tim-soft:
Let's say I enter the wrong alias on a `findAll` query, I get this error

`"user is associated to comment using an alias. You've included an alias (author), but it does not match the alias defined in your association."`

This message could be a lot more helpful if it mentioned the *actual alias* given in the association

## What would be better?
`"user is associated to comment using an alias. You've included an alias (author), but it does not match the alias defined in your association (superCoolAuthor)."`

Is this possible? It's kind of nitpicky but also equally annoying to need to dig back into the models to see what the actual alias is.

JMroczka:
This would be amazingly helpful, thanks!

sushantdhiman:
There can be multiple aliases between participating models, so it would be a long list `superCoolAuthor, superCoolAuthor2, superCoolAuthor3, ....`

Not sure we want to put full list here

jaronwright:
We are experiencing the same issue has anyone found a solution/can we make this un-stale?

sushantdhiman:
Moved to #10040

sushantdhiman:
Fixed by https://github.com/sequelize/sequelize/pull/10041

PR #10041, opened 2018-10-17
fix(model): improve wrong alias error message

### LINKED ISSUE #10040: The error message for a wrong alias could be better
Consider the code below, which shows a mistake in which the wrong alias is given:

```js
const sequelize = new Sequelize({ dialect: 'sqlite', storage: 'db.sqlite' });

const Foo = sequelize.define("foo", { name: Sequelize.STRING });
const Bar = sequelize.define("bar", { name: Sequelize.STRING });
Foo.hasMany(Bar, { as: "some_alias" });

sequelize.sync()
    .then(() => Foo.create({ name: "foo" }))
    .then(() => Bar.create({ name: "bar", fooId: 1 }))
    .then(() => Foo.findAll({
        include: {
            model: Bar,
            as: "wrong_alias"
        }
    }))
    .then(result => {
        console.log(JSON.stringify(result, null, 4));
    });
```

Sequelize throws the following error:

```
SequelizeEagerLoadingError: bar is associated to foo using an alias. You've included an alias (wrong_alias), but it does not match the alias defined in your association.
```

It is already great that it includes the wrong alias in the error message, but it would be even more helpful if the correct alias(es) were also given in the error message. Something like:

```
SequelizeEagerLoadingError: bar is associated to foo using an alias. You've included an alias (wrong_alias), but it does not match the alias(es) defined in your association (some_alias).
```

papb:
### Pull Request check-list

### Description of change

Closes #10040

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -698,8 +698,9 @@ class Model {
       association = this.getAssociationForAlias(targetModel, targetAlias);
       if (!association) {
         if (targetAlias) {
+          const existingAliases = this.getAssociations(targetModel).map(association => association.as);
           throw new sequelizeErrors.EagerLoadingError(`${targetModel.name} is associated to ${this.name} using an alias. ` +
-            `You've included an alias (${targetAlias}), but it does not match the alias defined in your association.`);
+            `You've included an alias (${targetAlias}), but it does not match the alias(es) defined in your association (${existingAliases.join(', ')}).`);
         } else {
           throw new sequelizeErrors.EagerLoadingError(`${targetModel.name} is associated to ${this.name} using an alias. ` +
             'You must use the \'as\' keyword to specify the alias within your include statement.');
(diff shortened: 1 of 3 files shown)


sushantdhiman:
Thanks for this PR, this has been discussed before https://github.com/sequelize/sequelize/issues/9079. My only issue is https://github.com/sequelize/sequelize/issues/9079#issuecomment-391586268

papb:
Oops, I should have searched a bit more, sorry.

To be honest, about your comment, I don't see a problem with a potentially huge list of things there:

* One single alias is the most common situation, so this error message would be great in these cases;
* Perhaps two or three aliases could indeed happen a few times, although never happened to me, but three aliases still isn't that much. And the more aliases that are defined, the less common it is.
* And regardless of the above, at least personally, I don't mind huge error messages at all. What bugs me is short error messages that aren't useful. In fact, I am thankful that this error message is already very useful (compared to error messages from other JS libraries around), but it still could be better.

Usually this is an error that crashes the application, will be seen immediately by the developer, and I am positive that he/she wouldn't mind an error message that saves time, even if it is huge. Every time this happened to me, it took me a while to find out the correct alias...

PR #10050, opened 2018-10-20
fix(model): create instance with many-to-many association with extra column (#10034)

### LINKED ISSUE #10034: Create instance with many-to-many association with extra column
## What are you doing?

I'm trying to create an instance along with its many-to-many association which has an extra column in the junction table:

```js
const sequelize = new Sequelize({ dialect: 'sqlite', storage: 'db.sqlite' });

const Foo = sequelize.define("foo", { name: Sequelize.STRING });
const Bar = sequelize.define("bar", { name: Sequelize.STRING });
const Foo_Bar = sequelize.define("foo_bar", { baz: Sequelize.STRING });
Foo.belongsToMany(Bar, { through: Foo_Bar });
Bar.belongsToMany(Foo, { through: Foo_Bar });

sequelize.sync().then(() => {
    return Foo.create({
        name: "foo...",
        bars: [
            {
                name: "bar...",
                foo_bar: {
                    baz: "baz..."
                }
            }
        ]
    }, { include: Bar });
}).then(() => {
    return Foo.findOne({ include: Bar });
}).then(result => {
    console.log(JSON.stringify(result, null, 4));
});
```

I expected the following output (showing that everything was created successfully):

```js
{
    "id": 1,
    "name": "foo...",
    "bars": [
        {
            "id": 1,
            "name": "bar...",
            "foo_bar": {
                "baz": "baz...",
                "fooId": 1,
                "barId": 1
            }
        }
    ]
}
```

```
TypeError: Cannot read property 'indexOf' of undefined
    at model.set (temp/node_modules/sequelize/lib/model.js:3379:22)
    at Promise.map.instance (temp/node_modules/sequelize/lib/model.js:3759:26)
```

__Dialect:__ Tested with sqlite but could affect others too
__Dialect version:__ sqlite3 3.1.13 and 4.0.2
__Sequelize version:__ 4.39.0 and 5.0.0-beta.12
__Tested with latest release:__ Yes (5.0.0-beta.12)

tsasaki609:
### Pull Request check-list

### Description of change

fix #10034 
I tried to include extra attributes in the condition processing the scope.


lib/model.js
```diff
@@ -3831,8 +3832,21 @@ class Model {
...
                   Object.assign(values, include.association.through.scope);
+                  if (instance[include.association.through.model.name]) {
+                    for (const attr of Object.keys(include.association.through.model.rawAttributes)) {
+                      if (include.association.through.model.rawAttributes[attr]._autoGenerated) {
+                        continue;
+                      }
+                      if (attr === include.association.foreignKey || attr === include.association.otherKey) {
+                        continue;
+                      }
+                      values[attr] = instance[include.association.through.model.name][attr];
```
sushantdhiman: Check if `instance[include.association.through.model.name][attr]` is defined only then we should set `values[attr]`

lib/model.js
```diff
@@ -3831,8 +3832,21 @@ class Model {
...
-                  // Include values defined in the scope of the association
+                  
+                  // Include values defined in the association
                   Object.assign(values, include.association.through.scope);
+                  if (instance[include.association.through.model.name]) {
+                    for (const attr of Object.keys(include.association.through.model.rawAttributes)) {
+                      if (include.association.through.model.rawAttributes[attr]._autoGenerated) {
+                        continue;
+                      }
+                      if (attr === include.association.foreignKey || attr === include.association.otherKey) {
```
sushantdhiman: Both if statements can be folded into single check?

test/integration/associations/belongs-to-many.test.js
```diff
@@ -1786,6 +1786,40 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), () => {
...
+                  }
+                }
+              ]
+            }, {
+              include: Bar
+            });
+          }).then(() => {
+            return Foo.findOne({ include: Bar });
+          }).then(foo => {
+            expect(foo.name).to.equal('foo...');
```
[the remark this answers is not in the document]
tsasaki609: That makes sense.

sushantdhiman:
Thanks @tsasaki609 :+1:

papb:
Thank you very much @tsasaki609, excellent work! :grin:

PR #10051, opened 2018-10-20
[WIP] refactor: use sequelize-pool for pooling

sushantdhiman:
### Pull Request check-list

### Description of change

This PR will migrate Sequelize to https://github.com/sushantdhiman/sequelize-pool which is a modernized fork of `generic-pool@2.5`


lib/dialects/abstract/connection-manager.js
```diff
@@ -1,26 +1,18 @@
 'use strict';
 
-const Pooling = require('generic-pool');
+const Pool = require('sequelize-pool').Pool;
```
SimonSchick: I think you can use destructuring with node v6 eg. `const { Pool } = require('sequelize-pool')`

sushantdhiman: :+1:

SystemDisc:
Is there any way this change can be implemented in v4? We're currently experiencing occasional issues with `RequestResource timed out` being thrown in `generic-pool` when attempting to run a query like `modelInstance.find(...)`. We believe this change would resolve the issue we're having, but we cannot use beta software.

If this change cannot be implemented in v4. Is there any estimate on when v5 might come out of beta?

SimonSchick:
So that's what that is, we experienced the same issue.
@sushantdhiman I think it's fair to roll out v5 at this point, it has enough changes and fixes as is.

kelvintaywl:
that would be really great! 🙇 
we are just about to deliberate on using v5 beta for a fix.

PR #10068, opened 2018-10-23
fix OptimistickLockError message

PavelPolyakov:
`BaseError` expects to receive `message`, but the `options` object is being passed. 

It leads to this:
```
SequelizeOptimisticLockError: [object Object]
```

With the fix the error looks like:
```
SequelizeOptimisticLockError: Attempting to update a stale model instance: Customers
```

### Pull Request check-list

### DIFF
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -92,7 +92,7 @@ class OptimisticLockError extends BaseError {
   constructor(options) {
     options = options || {};
     options.message = options.message || `Attempting to update a stale model instance: ${options.modelName}`;
-    super(options);
+    super(options.message);
     this.name = 'SequelizeOptimisticLockError';
     /**
      * The name of the model on which the update was attempted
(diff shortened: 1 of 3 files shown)


sushantdhiman:
This PR needs some tests, you can add new describe section for `Optimistic locking` [here](https://github.com/sequelize/sequelize/blob/master/test/integration/error.test.js#L245) and get reference for sample tests from [here](https://github.com/sequelize/sequelize/blob/master/test/integration/model/optimistic_locking.test.js#L46), finally assert correct message is returned along with correct error type

PavelPolyakov:
That's not what you expect when you use newly released inline edit button on `github` 😅 
But I'll have a look.

PavelPolyakov:
@sushantdhiman with all respect, I don't think that integration test is necessary here. The change is very minor and it just does the thing which was originally meant by author.

So I've put additional unit test, just to be rock solid, that message it there and has correct format.

Regards,

sushantdhiman:
@PavelPolyakov integration test is important here, I'll add that test myself soon

Issue #9863, opened 2018-09-01
Support for custom range types (Automatic)


javiertury:
I want to add support for custom ranges. The only database that supports ranges is postgresql, so it is the only dialect I modify. The main challenge is that custom ranges oids are not consistent across database instances. I decided to hook postgres ranges on `_refreshDynamicOIDs`. Although this method works only on postgres versions above 8.3, it's not a problem. Support for ranges was added on 9.3.

My progress so far: https://github.com/javiertury/sequelize

- Automatically detect all base ranges, array ranges and their basetypes.
  - Hooked ranges into `_refreshDynamicOIDs`
- Use postgres global `DataTypes.postgres.RANGE.oid_map`
- Use global `DataTypes.RANGE.subtypes` for `RANGE.prototype.toSql`
- Use global `DataTypes.RANGE.castTypes` for `RANGE.prototype.toCastType`

Notice that while postgres range class creates a new property for `oid_map`, it overrides basetype range property `subtypes` and `castTypes`. The reason is that `subtypes` and `castTypes` are theoretically usable across dialects, other databases could decide to implement ranges.

After this, implementing custom ranges should be as easy as
```
function addTimerange(DataTypes) {
  DataTypes.RANGE.subtypes.custom = 'customrange';
  DataTypes.RANGE.castTypes.custom = 'custombasetype';
}
```

Doubts

- Does overriding basetype range `subtypes` and `castTypes` have any side effect? I mean, can sequelize theoretically use more than one database dialect at the same time? Then those properties would get overriden and the system would be messed up.
- The purpose of this code is to workaround different oids across database instances. Can sequelize connect to more than one instance at the same time and override `oid_map` of another instance?

sushantdhiman:
> Does overriding basetype range properties subtypes and castTypes have any side effect? I mean, can sequelize theoretically use more than one database dialect at the same time? Then those properties would get overriden and the system would be messed up.

> The purpose of this code is to workaround different oids across database instances. Can sequelize connect to more than one instance at the same time and override oid_map of another instance?

Yes, these are very valid concerns. In same application you could be using multiple sequelize instances which may connect to different databases, hence multiple difference OIDs. Check this PR https://github.com/sequelize/sequelize/pull/9026

At least for now I think Postgres OIDs won't interfere with other instances, but I still think at given instance datatypes oids information may be invalid in multiple instance environment i.e.

```js
Datatypes.postgres.CITEXT.oids // say it is set to empty at start

[...]

PR #10077, opened 2018-10-25
feat(postgres): dyanmic oids(#9863)

javiertury:
### Pull Request check-list

### Description of change

Related to issue #9863. This pull request makes postgres `DataTypes` OID agnostic, all OID information is contained in `ConnectionManager`. They are refreshed automatically.

I've added some tests but I didn't run them. I will wait for this pull request to run them on travis.

EDIT: Autodetection of Ranges dropped, reasons in #9863

### DIFF
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -18,40 +18,67 @@ class ConnectionManager extends AbstractConnectionManager {
     const pgLib = this._loadDialectModule('pg');
     this.lib = this.sequelize.config.native ? pgLib.native : pgLib;
 
+    this._clearDynamicOIDs();
     this._clearTypeParser();
     this.refreshTypeParser(dataTypes.postgres);
   }
 
   // Expose this as a method so that the parsing may be updated when the user has added additional, custom types
   _refreshTypeParser(dataType) {
-    if (dataType.types.postgres.oids) {
-      for (const oid of dataType.types.postgres.oids) {
-        this.oidMap[oid] = value => dataType.parse(value, oid, this.lib.types.getTypeParser);
+    const arrayParserBuilder = parser => {
+      return value => this.lib.types.arrayParser.create(value, parser).parse();
+    };
+    const rangeParserBuilder = parser => {
+      return value => dataType.parse(value, {parser});
+    };
+
+    // Set range parsers
+    if (dataType.key.toLowerCase() === 'range') {
+      for (const name in this.nameOidMap) {
+        const entry = this.nameOidMap[name];
+        if (! entry.rangeOid) continue;
+
+        const rangeParser = rangeParserBuilder(this.getTypeParser(entry.oid));
+        const arrayRangeParser = arrayParserBuilder(rangeParser);
+
+        this.oidParserMap.set(entry.rangeOid, rangeParser);
+        if (! entry.arrayRangeOid) continue;
+        this.oidParserMap.set(entry.arrayRangeOid, arrayRangeParser);
       }
+      return;
     }
 
-    if (dataType.types.postgres.array_oids) {
-      for (const oid of dataType.types.postgres.array_oids) {
-        this.arrayOidMap[oid] = value => {
-          return this.lib.types.arrayParser.create(value, v =>
-            dataType.parse(v, oid, this.lib.types.getTypeParser)
-          ).parse();
-        };
-      }
+    // Create parsers for normal or enum data types
+    const parser = value => dataType.parse(value);
+    const arrayParser = arrayParserBuilder(parser);
+
+    // Set enum parsers
+    if (dataType.key.toLowerCase() === 'enum') {
+      this.enumOids.oids.forEach(oid => {
+        this.oidParserMap.set(oid, parser);
+      });
+      this.enumOids.arrayOids.forEach(arrayOid => {
+        this.oidParserMap.set(arrayOid, arrayParser);
+      });
+      return;
     }
+
+    // Set parsers for normal data types
+    dataType.types.postgres.forEach(name => {
+      if (! this.nameOidMap[name]) return;
+      this.oidParserMap.set(this.nameOidMap[name].oid, parser);
+
+      if (! this.nameOidMap[name].arrayOid) return;
+      this.oidParserMap.set(this.nameOidMap[name].arrayOid, arrayParser);
+    });
   }
 
   _clearTypeParser() {
-    this.oidMap = {};
-    this.arrayOidMap = {};
+    this.oidParserMap = new Map();
   }
 
   getTypeParser(oid) {
-    if (this.oidMap[oid]) {
-      return this.oidMap[oid];
-    } else if (this.arrayOidMap[oid]) {
-      return this.arrayOidMap[oid];
-    }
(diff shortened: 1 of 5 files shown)


lib/dialects/postgres/connection-manager.js
```diff
@@ -18,40 +18,68 @@ class ConnectionManager extends AbstractConnectionManager {
...
+    for (const name of dataType.types.postgres) {
+      if (! this.nameOidMap[name]) continue;
+      this.oidParserMap.set(this.nameOidMap[name].oid, parser);
+
+      if (! this.nameOidMap[name].arrayOid) continue;
+      this.oidParserMap.set(this.nameOidMap[name].arrayOid, arrayParser);
     }
   }
 
   _clearTypeParser() {
```
sushantdhiman: `_clearTypeParser` should call `_clearDynamicOIDs`

lib/dialects/postgres/connection-manager.js
```diff
@@ -18,40 +18,68 @@ class ConnectionManager extends AbstractConnectionManager {
...
     this._clearTypeParser();
     this.refreshTypeParser(dataTypes.postgres);
   }
 
   // Expose this as a method so that the parsing may be updated when the user has added additional, custom types
   _refreshTypeParser(dataType) {
-    if (dataType.types.postgres.oids) {
-      for (const oid of dataType.types.postgres.oids) {
-        this.oidMap[oid] = value => dataType.parse(value, oid, this.lib.types.getTypeParser);
+    const self = this;
```
sushantdhiman: No self = this, you are already using arrow functions

lib/dialects/postgres/connection-manager.js
```diff
@@ -217,40 +249,39 @@ class ConnectionManager extends AbstractConnectionManager {
...
+      };
+
+      // Replace all OID mappings. Avoids temporary empty OID mappings.
+      this.nameOidMap = newNameOidMap;
+      this.enumOids = newEnumOids;
 
       this.refreshTypeParser(dataTypes.postgres);
     });
   }
+  _clearDynamicOIDs() {
```
sushantdhiman: Add a new line before this method

lib/dialects/postgres/connection-manager.js
```diff
@@ -18,40 +18,68 @@ class ConnectionManager extends AbstractConnectionManager {
...
-            dataType.parse(v, oid, this.lib.types.getTypeParser)
-          ).parse();
-        };
+    // Create parsers for normal or enum data types
+    const parser = value => dataType.parse(value);
+    const arrayParser = arrayParserBuilder(parser);
+
+    // Set enum parsers
+    if (dataType.key.toLowerCase() === 'enum') {
+      for (const oid of this.enumOids.oids || []) {
```
sushantdhiman: simple for without continue can be replaced with `forEach` for better readability

lib/dialects/postgres/data-types.js
```diff
@@ -31,40 +31,13 @@ module.exports = BaseTypes => {
...
-  BaseTypes.JSONB.types.postgres = {
-    oids: [3802],
-    array_oids: [3807]
-  };
-
-  BaseTypes.TIME.types.postgres = {
-    oids: [1083],
-    array_oids: [1183]
-  };
+  BaseTypes.UUID.types.postgres = ['uuid'];
```
sushantdhiman: This mapping doesn't serves any purpose, right? but leave it for now

javiertury: You are right, as of now it doesn't have any purpose. Originally I had two cases in mind. The first one was to allow overriding the default type parser(`this.lib.types.typeParser`) within sequelize. In this case you need the mapping to get the OID and set the parser. The second case was range autodetection, which requires base types to be mapped.

lib/dialects/postgres/data-types.js
```diff
@@ -163,15 +127,17 @@ module.exports = BaseTypes => {
...
-    oids: [],
-    array_oids: []
-  };
+  function CITEXT() {
+    if (!(this instanceof CITEXT)) return new CITEXT();
+    BaseTypes.CITEXT.apply(this, arguments);
+  }
+  inherits(CITEXT, BaseTypes.CITEXT);
+
+  CITEXT.parse = value => value;
```
sushantdhiman: Can be removed

lib/dialects/postgres/connection-manager.js
sushantdhiman: Every invocation of `_clearTypeParser` is also preceded by `_clearDynamicOIDs` (at least in connection manager) so that method should actually take care of clearing dynamic oids as well

But I understand your hesitation, they both are doing different things. You can leave this for now.

lib/dialects/postgres/data-types.js
javiertury: [Here](https://github.com/javiertury/sequelize/blob/b9f94a488014b4cac57fd49b2f23cf2582130626/lib/dialects/abstract/connection-manager.js#L51-L57), abstract connection manager requires a parse function to refresh a dataType. Without this CITEXT doesn't have an explicit parser, falls back to `node-pg-types` and fails the tests.

lib/dialects/postgres/connection-manager.js
```diff
@@ -18,40 +18,67 @@ class ConnectionManager extends AbstractConnectionManager {
...
+
+      if (! this.nameOidMap[name].arrayOid) return;
+      this.oidParserMap.set(this.nameOidMap[name].arrayOid, arrayParser);
+    });
   }
 
   _clearTypeParser() {
-    this.oidMap = {};
-    this.arrayOidMap = {};
+    this.oidParserMap = new Map();
```
sushantdhiman: You are using a `Map`, we actually have an internal support for such type mapping. I assume in a single system, all type oids are unique. Please have a look at https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/connection-manager.js#L10 parserStore implementation form other dialects. We can use that for postgres as well.

javiertury: Thanks for the suggestion. At the beginning I was excited about this, I thought this could save some code. Unfortunately, this is what I have found out after implement it.

  - Net increase of 20-30 lines of code
  - Would require sub-stores under each dialect in `parserStore`
  - Would require to map each sub-store to a connection manager
      - In my experiment I used the connection URL to map them
  - Would require to create a new postgres-specific method or the ability to handle postgres specificities
      - The builtin `.set()` method of `Map` is enough

In postgres, each `connectionManager` needs a separate store. The reason is that OIDs can be mapped to different data types across postgres instances. `parserStore` is a centralized database. To make `parserStore` compatible, it should allow multiple sub-stores for each dialect. These sub-stores need to be mapped to each connection manager, which increases complexity.

The `.refresh(`) method used by parserStore is not the one that postgres needs. `.refresh()` assumes that `dataType.parser` is the parser and that the names `dataType.types.postgres` are the keys to index them. Postgres is different, it uses OIDs instead of names to find the parser. On top of that, composite dataTypes(arrays and ranges) need special parsers. I firmly believe that the `.set(<key/oid>, <value/parser>)` method is the most appropiate.

sushantdhiman:
Thanks @javiertury :+1:

PR #10085, opened 2018-10-27
fix(model): validate virtual attribute (#9947)

### LINKED ISSUE #9947: bulkCreate ignores virtual attribute validation
## What are you doing?

```js
  await User.bulkCreate([
      {
        email: 'user1@example.com',
        password: '123',
        createdAt: new Date(),
        updatedAt: new Date(),
      },
    ],{ validate: true})

// In model attribute:
      password: {
        type: dt.VIRTUAL,
        set: function (val) { this.setDataValue('password', val)  },
        validate: { 
          len: [10,200],
          virtualFieldValidator: function() {
            console.log('---validator---') // never called
            return false;
          },
        },
      },
```

Expected Validation Error in promise

Add a new option of bulkCreate to explicitly enable virtual field validators.
```js
{ 
  validate: true,
  validateVirtual: true,
}
```

Or

implicitly enable virtual field validators whenever 'validate' is set to true

User records are successfully created since validators of virtual field are ignored

Sequelize version: "4.38.1"

tsasaki609:
### Pull Request check-list

### Description of change

fix #9947 
In the case of processing that is not bulk, we use rawAttribute, so we should do the same in this case.
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L3627

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2500,7 +2500,7 @@ class Model {
       ignoreDuplicates: false
     }, options);
 
-    options.fields = options.fields || Object.keys(this.tableAttributes);
+    options.fields = options.fields || Object.keys(this.rawAttributes);
 
     const dialect = this.sequelize.options.dialect;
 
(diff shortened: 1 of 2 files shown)


sushantdhiman:
Thanks @tsasaki609

PR #10086, opened 2018-10-27
Update sequelize.js

nbouvrette:
This is a proposal which would support dynamic model options. For example, if you would like to re-use a model but change the table name for database performance reasons. it would be easy to inject this using an option while importing the model. I'm not sure exactly how the argument length block was used considering the JsDoc - definitely something to look at before merging.

### Pull Request check-list

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -385,10 +385,11 @@ class Sequelize {
    * @tutorial https://github.com/sequelize/express-example
    *
    * @param {string} importPath The path to the file that holds the model you want to import. If the part is relative, it will be resolved relatively to the calling file
+   * @param {Object} [options] Model options
    *
    * @returns {Model} Imported model, returned from cache if was already imported
    */
-  import(importPath) {
+  import(importPath, options) {
     // is it a relative path?
     if (path.normalize(importPath) !== path.resolve(importPath)) {
       // make path relative to the caller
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Hmm, you can acheive this at application level, by defining same model with different options through some type of high order function. Model file name should usally match with table they are referring. Entire process will look something like this

```js
# hoc model factory
```

```js
model1 = hoc({ tableName: 'model1', ...otherOptions })
```

```js
model2 = hoc({ tableName: 'model2', ...otherOptions })
```

nbouvrette:
I understand. I guess your approach works as well. In my case, I have a class which can dynamically create tables depending on the parameters used. I added a wrapper class in front of Sequelize which can import models with options like this:

```js
    loadModel(name, options) {
        const path = `${__dirname}/models/${name}`;
        const cacheKey = path + JSON.stringify(options);

        if (!this.modelCache[cacheKey]) {
            this.modelCache[cacheKey] = require(path)(this.sequelize, Sequelize.DataTypes, options);
        }

        return this.modelCache[cacheKey];
    }
```

And this is an example of a model file which can load with dynamic table names:

```js
module.exports = (sequelize, DataTypes, options) => {

    if (!options.tableName) {
        throw 'Undefined table name'
    }

    return sequelize.define(options.tableName,
        {
            id: {
                type: DataTypes.BIGINT(19).UNSIGNED,
                primaryKey: true,
                autoIncrement: true
            },
            someField: {
                type: DataTypes.STRING(760),
                allowNull: false
            }
        }, {
            underscored: true
        }
    );
};
```

I thought it could be an interesting feature to add direction into Sequelize but maybe I just have a non-typical use case!

sushantdhiman:
Yeah I try to avoid expanding API if problem can be solved cleanly at userland code. Thanks @nbouvrette

Issue #10093, opened 2018-10-29
SQlite3 dialect promise rejection returning undefined promise


jordantogether:
## What are you doing?
Trying to create a SQlite3 instance, when using an incorrect file path. 
```js
var Sequelize = require('sequelize');

var sequelize = new Sequelize('mainDB', null, null, {
    dialect: "sqlite",
    storage: './nonexistent.path.sqlite',
});

const isWorking= await sequelize.authenticate(); // Should be error, isn't.

// OR
var sequelize = new Sequelize('mainDB', null, null, {
    dialect: "sqlite",
    storage: './nonexistent.path.sqlite',
});

  sequelize.authenticate()
  .then(function(err) {
    console.log('Connection has been established successfully.');
  }, function (err) {
    console.log('Unable to connect to the database:', err);
  });

// Errors - sequelize is undefined.

```

## What do you expect to happen?
Get an error saying can't find the file. 

## What is actually happening?
Get an error saying sequelize authenticate is undefined (i.e. doesn't return promise).

_Output, either JSON or SQL_

## Pull request
I've added a pull request to resolve the issue, creating the bug to resolve. 👍 

__Dialect:__ sqlite
__Dialect version:__ Latest
__Database version:__ Latest
__Sequelize version:__ Latest
__Tested with latest release:__ Yes (If yes, specify that version) : __v5.0.0-beta.13__

## More info
* __NodeJs__: nodeJs version 10.0.0
* __NPM__: npm version 6.4.1
* __OS__: Windows 10 Pro, 10.0.15063 Build 15063 on VMWare Virtual Machine, Intel Xeon E5-2699A v4 @ 2.40GHz, 2394 Mhz, 2 Core(s), 2 Logical Processor(s)
* __SQLite Version__: "sqlite3": "^4.0.2"

**Note :** _Your issue may be ignored OR closed by maintainers if it's not tested against latest version OR does not follow issue template._

sushantdhiman:
> Sequelize version: Latest
> Tested with latest release: Yes (If yes, specify that version)

You failed to specify most important detail :)

jordantogether:
Updated, my apologies. Also, I'm doing the pull request and it looks like there are existing errors in master, do I need to fix those before I do a pull request? There's over 25 errors currently.

sushantdhiman:
latest master build looks green to me https://travis-ci.org/sequelize/sequelize/builds/447361698?utm_source=github_status&utm_medium=notification

[...]

PR #10094, opened 2018-10-29
fix: SQLite3 promise rejection bug

jordantogether:
### Pull Request check-list

- [x] Have you added new tests to prevent regressions? (___N/A__)

### Description of change
Bugfix for https://github.com/sequelize/sequelize/issues/10093

### DIFF
--- a/lib/dialects/sqlite/connection-manager.js
+++ b/lib/dialects/sqlite/connection-manager.js
@@ -59,8 +59,7 @@ class ConnectionManager extends AbstractConnectionManager {
         options.readWriteMode || this.lib.OPEN_READWRITE | this.lib.OPEN_CREATE, // default mode
         err => {
           if (err) {
-            if (err.code === 'SQLITE_CANTOPEN') return reject(new sequelizeErrors.ConnectionError(err));
-            return reject(new sequelizeErrors.ConnectionError(err));
+            reject(new sequelizeErrors.ConnectionError(err));
           }
           debug(`connection acquired ${options.uuid}`);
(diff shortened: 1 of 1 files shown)


lib/dialects/sqlite/connection-manager.js
```diff
@@ -59,8 +59,7 @@ class ConnectionManager extends AbstractConnectionManager {
         options.readWriteMode || this.lib.OPEN_READWRITE | this.lib.OPEN_CREATE, // default mode
         err => {
           if (err) {
-            if (err.code === 'SQLITE_CANTOPEN') return reject(new sequelizeErrors.ConnectionError(err));
-            return reject(new sequelizeErrors.ConnectionError(err));
+            reject(new sequelizeErrors.ConnectionError(err));
```
SimonSchick: Missing return? This would still trigger the resolver (basically a no-op).

SimonSchick:
Also no test case.

lib/dialects/sqlite/connection-manager.js
jordantogether: Right, but when it triggers the resolver, lines 62 and 63 return undefined, and with the return, it returns "undefined" instead of the promise, which results in a `cannot read property 'then' of undefined'. 

I'll update the test cases (even with master I'm still getting errors about the DB being locked on some tests) but I don't know why there are two conditions in the case which return exactly the same case - if the error code is can't open return x, if it isn't, return x.

sushantdhiman:
This is invalid issue, bug is likely userland https://github.com/sequelize/sequelize/issues/10093

Issue #9541, opened 2018-06-16
Connection pool doesnt reconnect with MySQL after server dies and comes back


stringbeans:
## What are you doing?
my repro steps using mariadb 10.2 and sequelize v4.37.10:

1) start my api server
2) issue a query (works fine)
3) stop the database
4) issue a query (fails due to db being down)
5) start the database 
6) issue a query (continues to fail)

My config looks like so...

```js
{
    dialect: 'mysql',
    dialectOptions: {
      multipleStatements: true,
      decimalNumbers: true
    },
    replication: {
      write: {
        host: process.env.MYSQL_MASTER_HOST,
        port: process.env.MYSQL_MASTER_PORT,
        username: process.env.MYSQL_MASTER_USERNAME,
        password: process.env.MYSQL_MASTER_PASSWORD
      },
      read: [{
        host: process.env.MYSQL_SLAVE_HOST,
        port: process.env.MYSQL_SLAVE_PORT,
        username: process.env.MYSQL_SLAVE_USERNAME,
        password: process.env.MYSQL_SLAVE_PASSWORD
      }]
    },
    pool: {
      max: process.env.MYSQL_MASTER_CONNECTION_LIMIT,
      min: 1,
      idle: 10000
    },
    define: {
      timestamps: false
    }
}
```

## What do you expect to happen?
After the database is restarted, I should not have to restart the api server to issue queries. I would expect the connection pool to self-heal and begin issuing queries again.

## What is actually happening?
All subsequent queries after the database comes back up will time out.

__Dialect:__ mysql
__Dialect version:__ mysql2@1.5.3
__Database version:__ mariadb 10.2.13
__Sequelize version:__ 4.37.10
__Tested with latest release:__ Yes (4.37.10)

stringbeans:
On closer inspection, it looks as though the pool reconnection WORKS if I don't specify the replication object in my configuration and instead just use a single connection. Also, maybe an important detail is that locally my master and my slave are both pointing to the same database.

knoxcard:
Close ticket?

stringbeans:
@knoxcard no its still an issue except it only seems to affect configurations with replication...

gsogol:
I get the same

PR #10103, opened 2018-10-31
refactor: use sequelize-pool for pooling

mikeringrose:
### Pull Request check-list

### Description of change

This fixes an issue with generic-pool not allowing the node process to exit, referenced in #8468. This PR pool over the work done by sushantdhiman in #10051.

> This PR will migrate Sequelize to https://github.com/sushantdhiman/sequelize-pool which is a modernized fork of generic-pool@2.5
>
>In theory this should solve issues like
>
> #9541
> #8468


sushantdhiman:
@mikeringrose I can't accept this PR for v4. It swaps pooling library this can have unintended side effects on someone's production, lots of pooling code base has been refactored which again could have introduced breaking change.

mikeringrose:
That’s unfortunate as a fair number of people running sequelize in lambdas are experiencing timeouts.

sushantdhiman:
@mikeringrose So you expect me to break things for others, I've to follow semver. If you want you can use this branch in your code or upgrade to v5

mikeringrose:
@sushantdhiman that's assuming others would be broken by this, which is why there are unit and integration tests. with that said, i'd prefer to fix this issue that people are experiencing while maintaining the current code base for others.

sushantdhiman:
@mikeringrose I migrated that old code base from `generic-pool` and even I can't guarantee if it will be fully compatible (actually it can't be as generic-pool@2.5 and 3 are breaking).

Passing unit tests and integrations tests is fine but I don't want try swapping semver breaking libraries in same major version, when used for such critical component. I am certain there will be some hidden case which we both can't consider right now, but users will later complain, which I will have to hear.

I am just trying to keep v4 non breaking, it's not like I don't want this patch to be ported over to v4. Historically most v4 ports has been accepted and released pretty fast.

PR #10106, opened 2018-10-31
refactor(cls): lazy load cls and make it entirely optional

SimonSchick:
### Pull Request check-list

### Description of change

I noticed that `cls-bluebird` is always installed even when cls is entirely optional (same as `continuation-local-storage`).

This change makes both dependencies peer dependencies and asserts them lazily when calling `.useCLS`.

Pro:
 - Less packages shipped by default.
 - Lower load time
Contra:
 - Peer dependencies generate warnings, we can just remove them however.
 - 1 More manual install

### DIFF
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -54,6 +54,8 @@ return sequelize.transaction(t => {
 
 ### Automatically pass transactions to all queries
 
+In order to use CLS you must first `npm install cls-bluebird`.
+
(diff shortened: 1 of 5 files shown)


SimonSchick:
Added a test too, not sure if it's too hacky tho.

sushantdhiman:
It is ok for now, seems hacky. Probably better to extract this as plugin or remove CLS support altogether or have support with cls-hooked

PR #10118, opened 2018-11-04
fix(query): correct-quote-identifiers(#9964)

### LINKED ISSUE #9964: Sequelize does not generate correct SQL query when using the findAll query with a schema that contains fields with '.' characters.
## What you are doing?
I have a schema with fields such as 'foo.bar.baz'
Sequelize will generate the correct 'Create Table if not exist' query but will not generate a correct 
'SELECT' query with this column name. The SELECT that gets generated from this is " SELECT `foo.bar`.`baz` from <table> where <condition>. The SELECT query should be "SELECT `foo.bar.baz`. As a result the db responds with an error, (Responding with error. SequelizeBaseError: ER_BAD_FIELD_ERROR: Unknown column 'foo.bar.baz' in 'field list') which is also misleading.

The following model will reproduce with mySQL db
{
        id: {
            type: sequelize.UUID,
            primaryKey: true
        },
        'foo.bar.baz': sequelize.TEXT
}

         model.findAll({ where: {id: 123} }).then(
                items => {},
                error => {}
            );

SELECT `foo.bar.baz`, `id` FROM <modelname> WHERE `id` = 123

SELECT `foo.bar`.`baz`, `id` FROM <modelname> WHERE `id` = 123

_Output, either JSON or SQL_

__Dialect:__ mysql
__Database version:__ AWS Aurora
__Sequelize version:__ 3.30

tsasaki609:
### Pull Request check-list

### Description of change

fix #9964

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -883,6 +883,14 @@ class QueryGenerator {
     return this.quoteIdentifier(identifiers);
   }
 
+  quoteAttribute(attribute, model) {
+    if (model && attribute in model.rawAttributes) {
+      return this.quoteIdentifier(attribute);
+    } else {
+      return this.quoteIdentifiers(attribute);
+    }
+  }
+
   /**
    * Quote table name with optional alias and schema attribution
    *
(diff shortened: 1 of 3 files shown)


sushantdhiman:
Thanks @tsasaki609

mbecker16:
Thanks @tsasaki609 for fixing this

PR #10091, opened 2018-10-28
Chore/optimize

SimonSchick:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [ ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ ] Have you added new tests to prevent regressions?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?

### Description of change

This PR does a couple of things (please advise if I should split this into separate PRs) that are all somewhat related.

1)

Replace various operations with faster equivalents (contains some minor `else if` simplification as well.
Main component is the replacement of `_.memoize` which was used to cache array lookups with shiny `Set`s. This also removes some dead code.

2)

While debugging and issue I got really frustrated with the amount of conditional nesting is going on
I reduced the amount of `return } else if (...)` by roughly 50% (mostly using return instead).
In `query.js` ~I also wrapped some promises that were doubled wrapped and nested for some reason, this code is drastically simpler now~ This broke stuff, I still moved the post query logic to a method to make it easier to read.

3)

NOTE: I switched the test reporter to `dot` as that is slightly faster and doesn't create so much useless output, please mention if this is undesirable.
Replaces all occurences of `new Buffer` and `sinon.sandbox.create`  and stubbed out (most) data-type related warnings that were spamming logs.
~Also refactored `ABSTRACT` prototype class to class since it's never directly instantiated.~ Nvm see https://github.com/nodejs/node/issues/23960

4)

Previously forgot to port this :)

PR #10122, opened 2018-11-05
refactor: optimize memoize use, misc cases

SimonSchick:
### Pull Request check-list

### Description of change

Followup of #10091, re-implements the previous changes that were dropped for being too out of scope.

This PR also optimizes/streamlines type checks eg. `_.isType` to `typeof ... === 'type'`, lodash is generally slower for those as it uses alternative methods for type checking that are compatible with legacy browsers and the likes.


lib/utils.js
```diff
@@ -639,3 +635,17 @@ function nameIndex(index, tableName) {
...
+
+/**
+ * Fast(er) set to array than spreading.
+ * @param {Array} set
+ * @param {Set} set
+ * @returns Array
+ * @private
+ */
+function addSetToArray(array, set) {
+  for (const v of set) {
```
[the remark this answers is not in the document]
SimonSchick:
```
'use strict';

function addSetToArray(array, set) {
  for (const v of set) {
    array.push(v);
  }
}

const set = new Set([1, 2, 3, 4, 5, 6, 7, 8, 9]);

console.time('addSetToArray');
for(let i = 0;i<0xffffff;++i) {
  addSetToArray([], set);
}
console.timeEnd('addSetToArray');

console.time('Array.from');
for(let i = 0;i<0xffffff;++i) {
  [].concat(Array.from(set));
}
console.timeEnd('Array.from');
```

```
addSetToArray: 736.759ms
Array.from: 20121.014ms
```

On my machine using `.concat(Array.from(set))` is about **25x** slower.

The intention here is to make the code faster using sets, converting sets to arrays is a rather expensive operation, `addSetToArray` is the fastest way to do it pretty much, using `Array.from` would nullify all performance gained 😛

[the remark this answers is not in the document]
sushantdhiman: I personally verified and `Array.from` is indeed really slow. Bummer, it was really concise. Will go with `for + push` approach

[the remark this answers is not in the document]
SimonSchick: See also https://stackoverflow.com/questions/53146637/array-resize-performance-setting-length-property-vs-repeated-pushing

SimonSchick:
Uhh, wait, why did you merge master into my branch? 😛

SimonSchick:
@sushantdhiman mergy merge?

lib/model.js
```diff
@@ -3397,9 +3368,9 @@ class Model {
       } else {
         // Loop and call set
         if (options.attributes) {
-          let keys = options.attributes;
+          let keys = options.attributes.slice();
           if (this.constructor._hasVirtualAttributes) {
-            keys = keys.concat(this.constructor._virtualAttributes);
+            Utils.addSetToArray(keys, this.constructor._virtualAttributes);
```
sushantdhiman: I didn't merged because this still doesn't looks right to me, I think we should keep using array for `_virtualAttributes`

SimonSchick: The point of this PR was to primarily to get rid of `memoize + includes` and replace with sets since that's slows all reads. Also keep in mind the set is only converted for iteration.

I could change the code in this location to not create arrays at all.

sushantdhiman: > I could change the code in this location to not create arrays at all.

Hmm, give it a shot

SimonSchick: Done, waiting for build, I'm in the process of moving atm so I might take a while to reply.

sushantdhiman:
Thanks @SimonSchick :+1:

PR #10125, opened 2018-11-05
fix(query-generator): fix mssql issue with ON DELETE foreign key constraints

bminer:
### Pull Request check-list

### Description of change

Fixed issue with MSSQL query generator.  Before, ON DELETE clauses were only added to CREATE TABLE queries for the first foreign key constraint.  This behavior has been updated to only add a foreign key constraint for a given model once.

Note that the `.toString()` method previously used was not overwritten; it returned `[object Object]`.

Please let me know if you have any thoughts or suggestions.  I'd love to help get this merged ASAP.  Thanks!


sushantdhiman:
Add a new test [here](https://github.com/sequelize/sequelize/blob/b896ba4dbbc90811c53fc2dce4351d0daaed8455/test/unit/sql/create-table.test.js#L30) which try to add two references and assert correct query for all dialects

mirkojotic:
@bminer let me know if you don't have bandwidth to finish this and I'll add the required test.

bminer:
@mirkojotic - thank you for reaching out.  Unfortunately, I do not have the bandwidth at this time, and I'm not very familiar with how to run the tests.  If you could help, I would greatly appreciate it!

bminer:
@sushantdhiman @mirkojotic - Just responding back here.  This issue still exists and prevents Sequelize from handling foreign key constraints properly in MSSQL.  Using `.toString()` as before returns `[Object object]`, and `JSON.stringify` is the correct approach, I think.  I'm sorry that I can't make a test without a little more knowledge.

bminer:
I should also mention that this is a tiny change, so I would really appreciate someone with some testing knowledge to take a stab at writing the test.

papb:
@mirkojotic Are you available to help?

bminer:
After doing some additional testing, this might be working on the latest version of Sequelize...

I will write back to confirm

PR #10149, opened 2018-11-11
fix(query-generator): adding offset in unioned queries

schie:
fixed an issue when passing limit/offset inside of a subquery

### Pull Request check-list

- [ - ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ - ] Have you added new tests to prevent regressions?
- [ - ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ x ] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?

### Description of change

When executing a query like the following, the offset value would not get passed to the UNION ALL Generated query.

```js
models.tblLeague.findAll({
      include: [
        {
          model: models.tblPost,
          order: [['createdDate', 'DESC']],
          offset: 5,
          limit: 10,
          where: { isActive: state.active },
        },
      ],
      where: { leagueId: [1,2,3] },
    });
```

In a nutshell, adds same code (addition of offset) from #9577  from v5 to v4

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1101,6 +1101,7 @@ const QueryGenerator = {
           {
             attributes: options.attributes,
             limit: options.groupedLimit.limit,
+            offset: options.offset,
             order: groupedLimitOrder,
             where,
             include,


lib/dialects/abstract/query-generator.js
```diff
@@ -1101,6 +1101,7 @@ const QueryGenerator = {
           {
             attributes: options.attributes,
             limit: options.groupedLimit.limit,
+            offset: options.offset,
```
mickhansen: `options.groupedLimit.offset`?

PR #10173, opened 2018-11-17
fix(association): enable eager load with include all(#9928)

### LINKED ISSUE #9928: as: {singular, plural} not working with include: {all: true}
## What are you doing?

I am specifying the singular and plural names of an alias. I then later try to include with the `all: true` option, but that results in a `SequelizeDatabaseError` of `{ name: 'SequelizeEagerLoadingError' }`.

When I specify an `as` without explicit singular/plurals this issue does not arise.

Note: I think the issue is similar to what @raxod502 commented [here](https://github.com/sequelize/sequelize/issues/6007#issuecomment-315431681).

```js
let name = 'theAliasName'

// Below works with include: { all: true }
through: db.define( name, {
		amount: Sequelize.FLOAT,
		unit: Sequelize.STRING,
		dataSource: Sequelize.STRING
	} ),
	as: name + 's'
} )

// Below DOES NOT work
through: db.define( name, {
		amount: Sequelize.FLOAT,
		unit: Sequelize.STRING,
		dataSource: Sequelize.STRING
	} ),
	as: {
		singular: name,
		plural: name + 's'
	}
} )

```

Later these associations are called with:

```js
Model.findOne( {
    where: { id: relevant_id, userId: user_id },
    include: [ { all: true } ]
} ) )
```

I expect:
- The alias table to be populated when a relationship is set (happens in both cases, no issue)
- The `include: { all: true }` to load the associated data

Sequelize throws `{ name: 'SequelizeEagerLoadingError' }`

__Dialect:__ postgres
__Database version:__ 10.5
__Sequelize version:__ 4.38.1

Edit 1: added code used to findOne
Edit 2: added error name

tsasaki609:
### Pull Request check-list

### Description of change

fix #9928 
Since conditions are different for each relation, I tried using polymorphism.

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -755,6 +755,18 @@ class BelongsToMany extends Association {
       sourceInstance[association.accessors.add](newAssociatedObject, _.omit(options, ['fields'])).return(newAssociatedObject)
     );
   }
+
+  verifyAssociationAlias(alias) {
+    if (typeof alias === 'string') {
+      return this.as === alias;
+    }
+
+    if (alias && alias.plural) {
+      return this.as === alias.plural;
+    }
+
+    return !this.isAliased;
+  }
 }
 
 module.exports = BelongsToMany;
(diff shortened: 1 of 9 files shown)


sushantdhiman:
Thanks @tsasaki609 :)

PR #10180, opened 2018-11-20
fix(transaction): prevent nested savepoints from having the same name

Verdier:
### Pull Request check-list

### Description of change

Fix a bug where nested savepoints have the name. This result in a critical bug where calling `rollback` on a preceding savepoint only `rollback` to the latest one.

```js
const transaction = await sails.sequelize.transaction();

const event = await Event.create({
    type: 'test',
    data: {}
}, {
    transaction
});

const savepoint1 = await sails.sequelize.transaction({ transaction });

event.type = 'test2';
await event.save({ transaction: savepoint1 });

const savepoint2 = await sails.sequelize.transaction({ transaction: savepoint1 });

event.type = 'test3';
await event.save({ transaction: savepoint2 });

await savepoint1.rollback();
await event.reload({ transaction });

// before the fix: event.type === 'test2'
// after the fix: event.type === 'test'
```

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -35,7 +35,7 @@ class Transaction {
     }, options || {});
 
     this.parent = this.options.transaction;
-    this.id = this.parent ? this.parent.id : generateTransactionId();
+    if (this.parent && this.parent.parent) this.parent = this.parent.parent;
 
     if (this.parent) {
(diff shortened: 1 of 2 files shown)


SimonSchick:
Closing due to age and failing tests.

PR #10201, opened 2018-11-26
docs(migrations): Simplify CLI Call

negezor:
### Description of change

Instead of directly calling the CLI from node_modules, you should use NPX that comes with NPM from v5.2.


sushantdhiman:
Thanks @negezor

Issue #1830, opened 2014-05-29
How to parse model and associations from raw query?


alekbarszczewski:
From time to time I have to write very complex query that cannot be handled by Sequelize default methods. So I write raw query by hand and I execute it using sequelize instance and passing also sequelize model:

```
Post.hasMany(PostTranslation);

q = '
SELECT 
   "Posts".*, 
   "PostTranslations"."id" AS "PostTranslations.id", 
   "PostTranslations"."language" AS "PostTranslations.language",
   "PostTranslations"."PostId" AS "PostTranslations.PostId"
FROM "Posts" 
INNER JOIN "PostTranslations" AS "PostTranslations"
ON ("Posts"."id" = "PostTranslations"."PostId")"
';

sequelize.query(q,Post).done(function(err,posts){
    console.error(posts[0])
})
```

That partialy works as I get array of post models however Sequelize does not detect that there are any associated models included... Is there any way to "inform" Sequelize how to parse selected rows to build models with their associations?

For now `posts[0].dataValues` looks like this:

```
{      id: 1,
       type: 'video',
       isPublished: true,
       publishedAt: Wed Jan 01 2014 01:00:00 GMT+0100 (CET),
       thumbnailUrl: 'http://placehold.it/230x150&text=post1',
       isFeatured: false,
       isDailyContent: false,
       lovesCount: 0,
       commentsCount: 0,
       CategoryId: 1,
       SeriesId: 1,
       UserId: 4,
       AuthorId: null,
       createdAt: Wed May 28 2014 18:07:28 GMT+0200 (CEST),
       updatedAt: Wed May 28 2014 18:07:28 GMT+0200 (CEST),
       'PostTranslations.id': 2,
       'PostTranslations.language': 'zh-cn',
       'PostTranslations.PostId': 6 }
```

As you can see all columns are treated as Post model properties instead of PostTranslation properties. Am I doing something wrong?
Using Sequelize 1.7.5

juliocanares:
what is the final solution for this ? ,

var options = {
  include: etc
};
Sequelize.Model.validateIncludedElements(options);

Not work for me , I'm using -> sequelize ^2.0.2

mickhansen:
@juliocanares There is none, it's on the roadmap.

juliocanares:
@mickhansen  then I need format manually the data ?

we4sz:
How is this going?

[...]

PR #10213, opened 2018-11-30
feat(raw query): add support for include on select

javiertury:
### Pull Request check-list

### Description of change

Related to #1830. Handle options.include for sequelize.query() for SELECT queries.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -531,6 +531,11 @@ class Sequelize {
         }
       }
 
+      if (options.model && options.type === QueryTypes.SELECT && options.include) {
+        options.hasJoin = true;
+        options.model._validateIncludedElements(options);
+      }
+
       if (options.transaction && options.transaction.finished) {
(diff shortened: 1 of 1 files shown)


javiertury:
I realized that `sequelize.query()` is not an interface to do custom queries. It is the internal function that executes every pre-processed query. I should not touch it.

prionkor:
@javiertury if not `sequelize.query()` do we need a new interface for custom queries?

javiertury:
@prionkor that could be feasible. `sequelize.query()` could be split into two. Already pre-processed queries (internal ones) would call `sequelize._query()` directly. Meanwhile, custom queries could use `query()`, which is a wrapper around `sequelize._query()` that applies some processing before execution.

On the other hand, maybe there is a way for `sequelize.query` to detect if the options have already been pre-processed.

Right now I am working on a big issue(binded parameters), so I won't be able to help.

PR #10217, opened 2018-12-02
fix(model): prevent version number from being interpreted as string

pixelspark:
### Pull Request check-list

- [X] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? --> not applicable

### Description of change

While testing with CockroachDB as backend database, I encountered a strange situation where the version number used by Sequelize for optimistic locking is (correctly) stored as INT in the table, but is incremented as string, and ends up becoming "1111...". Eventually the version number becomes too large to fit an integer. The exact same code does not show the issue when using MySQL or SQLite as backend store.

This fix ensures that the version number is incremented as a number. It merely makes the existing desired behaviour explicit and should not change the behaviour for backends where it is already working correctly.

I am not sure about how I would go about adding tests for this - more research would be needed to see whether this is CockroachDB or maybe also PostgreSQL-specific. Adding a test suite for Cockroach would be a good idea but out of my reach for now (I barely know the Sequelize source code at this point).

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3731,7 +3731,7 @@ class Model {
         where = this.where(true);
         where = Utils.mapValueFieldNames(where, Object.keys(where), this.constructor);
         if (versionAttr) {
-          values[versionFieldName] += 1;
+          values[versionFieldName] = parseInt(values[versionFieldName], 10) + 1;
         }
         query = 'update';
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Thanks @pixelspark

PR #10237, opened 2018-12-06
docs(Model.init): set correct path for `validate` option.

farisn:
docs(Model.init): set correct path for `validate` option.


sushantdhiman:
Are you sure about this! I dont remember `attribute.column.validate` to be a thing, but `attribute.validate` is pretty important? You might have got this idea from other `attributes.column.*` thingies?

farisn:
In documentation `attributes` is first argument of `Model.init` method.  `attributes.column` is property of `attributes` argument, where `column` is a attribute name.  Finally, the `attributes.column.validate` is the object for the validation of the attribute.
What's wrong with it?

PR #10276, opened 2018-12-17
chore: cleanup

### LINKED ISSUE #10275: Update documentation to NOT use promise lib specific functions such as `.spread`
The comment at the end of this code snippet says:
>In the example above, the "spread" on line 39 divides the array into its 2 parts and passes them as arguments to the callback function defined beginning at line 39, which treats them as "user" and "created" in this case. (So "user" will be the object from index 0 of the returned array and "created" will equal "true".)

There is no mention that this is a bluebird (and maybe others) specific promise implementation function. 

Two things should be addressed IMO
1. Remove promise specific implementation logic
2. Update documentation which mentions things like `line 39`. There is no line numbers so it's pointless and confusing

http://docs.sequelizejs.com/manual/tutorial/models-usage.html#-findorcreate-search-for-a-specific-element-or-create-it-if-not-available

image

SimonSchick:
### Pull Request check-list

### Description of change

Fixes #10275 I think

This is another small scale cleanup PR.
- Replace some loops with `Object.assign`
- Replace all usage of `spread` with native array spread
- Use object literals in various places
- Avoid aliasing of `this`

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -375,13 +375,12 @@ class BelongsToMany extends Association {
   get(instance, options) {
     options = Utils.cloneDeep(options) || {};
 
-    const association = this;
-    const through = association.through;
+    const through = this.through;
     let scopeWhere;
     let throughWhere;
 
-    if (association.scope) {
-      scopeWhere = _.clone(association.scope);
+    if (this.scope) {
+      scopeWhere = _.clone(this.scope);
     }
 
     options.where = {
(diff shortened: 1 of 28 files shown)


lib/associations/belongs-to-many.js
```diff
@@ -545,30 +541,38 @@ class BelongsToMany extends Association {
...
           if (Object.keys(attributes).length) {
-            promises.push(association.through.model.update(attributes, Object.assign(options, { where })));
+            promises.push(
+              this.through.model.update(attributes, Object.assign(options, {
+                where: {
+                  [identifier]: sourceInstance.get(sourceKey),
+                  [foreignIdentifier]: newObj.get(targetKey)
+                }
+              }
+              ))
```
sushantdhiman: odd indentation

sushantdhiman:
Thanks @SimonSchick :+1:

PR #10284, opened 2018-12-19
feat(query): bind parameters

javiertury:
### Pull Request check-list

### Description of change

Issue #10285. All user-facing queries use binded parameters in all dialects, except select queries on mysql dialect.

Changes

* Select queries on mysql are not binded because of `node-mysql2`
  * execute API: Not possible to use custom parsers, and no hints about data type.
  *  [node-mysql2 execute typecast issue](https://github.com/sidorares/node-mysql2/issues/717) needs to be solved.
    * No way to convert geometries to geoJSON
    * [No way to apply timezone](https://github.com/sidorares/node-mysql2/issues/888)
  * `mariadb-connector-nodejs` can be used as an alternative
    * Limitations of mariadb driver would apply to mysql, e.g., DECIMAL and BIGINT parsed as Numbers
* All tests of user-facing queries use binded parameters
  * Commitment to bind everything in the future.
  * Only integration tests cover escaping values in select queries for mysql
* New `Composition` object that holds strings (sql commands), `Slot` objects(binded parameters) and placeholders. This object is converted to a query in the very end.
  * QueryGenerators returns `Composition` objects
  * `sequelize.query()` accepts a `Composition` object as argument. `Composition` objects provide guarantees that make `formatBindParameters()` unnecessary. No need to reparse query with regex.
  * Tests must use dialect specific bind parameters, except for `default` dialect
* Many helper functions are converted to return `Composition` objects
   * Binding is delayed
   * Can choose between binding with `QB.composeQuery()` and or escaping with `QB.composeString()`
* Removes support for old SQL Server 2008 (Microsoft will discontinue extended support in 2019)
  * Offset hack is very difficult to deal with
* Fix version detection using semver, coherce non-semver compliant strings
* Logs binded parameters(useful for testing)
* Postgres `searchPath` query is run separately if there are values to bind
* Corner case: in postgres concurrent findOrCreate queries sharing the same transaction use `ignoreDuplicates` instead of exception handling
  * Now everything can be in 1 query
  * Exception handling was expensive
  * Requires postgres 9.5+ to handle this corner case with a shared transaction, user is warned
* Upsert query is run using two separate queries in sqlite
  * sqlite cannot run multiple statements in a single query if values are binded
  * Minimal delay, both queries are run in parallel and database is local
  * Both queries are idempotent, tests are satisfied
* Placeholder is now a Class, not an operator
  * More secure. Previously operator substitution searched for a string and replaced it, which can be faked by user. Users cannot fake a class instance.
  * Works better with Composition class
* Some DBs have limits on number of binded parameters
  * Previously, sqlite could insert up to 999 rows per query with default server settings. Now the limit is 999 binded parameters per query, which translates to only (999/#binds per row) rows per query with default server settings. To raise this number, tweak sqlite's config SQLITE_LIMIT_VARIABLE_NUMBER.
  * SQL server has a fixed limit of 2100 binded parameters, which translates to (2100/#binds per row) rows
* Methods `sequelize.slot()` and `sequelize.placeholder()` can be combined with `sequelize.composition()` to bind values and create placeholders.
  * Documentation not yet public. This is a proposed API, but perhaps there is a better way.
* For json data types, both the value and the json path are binded
  * In mysql the `->>` operator didn't allow to bind the json path. Changed to `json_unquote(json_extract(...))`


sushantdhiman:
Thanks @javiertury , I was out for Christmas and New year. Will try to give this thorough review this weekend :+1:

mkaufmaner:
Currently reviewing the PR and have to say I am rather impressed.  This really brings Sequelize into the future.  Good work @javiertury!!!

lib/dialects/abstract/query-generator.js
```diff
@@ -1917,21 +2117,21 @@ class QueryGenerator {
    * @private
    */
   addLimitAndOffset(options) {
-    let fragment = '';
+    const composition = new Composition();
 
     /* eslint-disable */
```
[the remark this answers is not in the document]
mkaufmaner: I see now, maybe instead remove the eslint-disable code and use the following instead?

```js
if (options.offset !== undefined && options.offset !== null && (options.limit === null || options.limit === undefined)) {
  // @TODO: Throw an exception or remove this code to circumvent potential DOS vector
  composition.add('LIMIT ', new Slot(options.offset), ', ' + 10000000000000);
} else if (options.limit !== undefined && options.limit !== null) {
  if (options.offset !== undefined && options.offset !== null) {
    composition.add('LIMIT ', new Slot(options.offset), ', ', new Slot(options.limit));
  } else {
    composition.add('LIMIT ', new Slot(options.limit));
  }
}
```

Side note: I think Sequelize should throw an exception when `offset` is passed but not `limit`.

lib/dialects/abstract/query-generator/placeholder.js
```diff
@@ -0,0 +1,14 @@
+
```
mkaufmaner: Extra space.

lib/dialects/mssql/query-generator.js
```diff
@@ -782,101 +805,38 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
     return 'ROLLBACK TRANSACTION;';
   }
 
-  selectFromTableFragment(options, model, attributes, tables, mainTableAs, where) {
-    let topFragment = '';
-    let mainFragment = `SELECT ${attributes.join(', ')} FROM ${tables}`;
-
-    // Handle SQL Server 2008 with TOP instead of LIMIT
-    if (semver.valid(this.sequelize.options.databaseVersion) && semver.lt(this.sequelize.options.databaseVersion, '11.0.0')) {
```
mkaufmaner: Removal of MSSQL <11.0.0 support should probably be well documented in the affected methods.

lib/dialects/mssql/query-generator.js
```diff
@@ -782,101 +805,38 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
...
+      composition.add(' FETCH NEXT ', new Slot(options.limit), ' ROWS ONLY');
     }
 
-    return mainFragment;
+    return composition;
   }
 
-  addLimitAndOffset(options, model) {
-    // Skip handling of limit and offset as postfixes for older SQL Server versions
-    if (semver.valid(this.sequelize.options.databaseVersion) && semver.lt(this.sequelize.options.databaseVersion, '11.0.0')) {
```
mkaufmaner: Removal of MSSQL <11.0.0 support should probably be well documented in the affected methods.

lib/dialects/postgres/index.js
```diff
@@ -42,7 +42,7 @@ PostgresDialect.prototype.supports = _.merge(_.cloneDeep(AbstractDialect.prototy
     functionBased: true
   },
   inserts: {
-    onConflictDoNothing: ' ON CONFLICT DO NOTHING'
+    onConflictDoNothing: 'ON CONFLICT DO NOTHING'
```
mkaufmaner: I don't believe the supports `EXCEPTION` is still in use with these changes. Maybe remove it entirely?

lib/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
javiertury: I have switched to strict null/undefined comparison.

I agree, Sequelize should at least warn if offset is used without limit. Furthermore, Sequelize should warn if either offset or limit are used without order by. This will produce inconsistent pagination [as warned by postgres](https://www.postgresql.org/docs/current/queries-limit.html) and other DBs.

lib/dialects/postgres/index.js
javiertury: Agree, exception handling requires temporary functions which are incompatible with binding parameters.

lib/dialects/mssql/query-generator.js
javiertury: Sequelize will throw an error before connecting to MSSQL < 11. The error is thrown by `versionSupport` in `lib/dialects/mssql/connection-manager.js`.

lib/dialects/abstract/query-generator.js
```diff
@@ -99,153 +105,149 @@ class QueryGenerator {
...
-        } LANGUAGE plpgsql; SELECT (testfunc.response).*, testfunc.sequelize_caught_exception FROM pg_temp.testfunc(); DROP FUNCTION IF EXISTS pg_temp.testfunc()`;
-      } else {
-        options.exception = 'WHEN unique_violation THEN NULL;';
-        valueQuery = `CREATE OR REPLACE FUNCTION pg_temp.testfunc() RETURNS SETOF ${quotedTable} AS $body$ BEGIN RETURN QUERY ${valueQuery}; EXCEPTION ${options.exception} END; $body$ LANGUAGE plpgsql; SELECT * FROM pg_temp.testfunc(); DROP FUNCTION IF EXISTS pg_temp.testfunc();`;
+    if (options.ignoreDuplicates) {
+      if (this._dialect.supports.inserts.onConflictDoNothing) {
+        if (insertProto.onConflict.length) insertProto.onConflict.add(' ');
+        insertProto.onConflict.add(this._dialect.supports.inserts.onConflictDoNothing);
+      }
+      if (this._dialect.supports.inserts.ignoreDuplicates) {
```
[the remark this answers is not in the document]
javiertury: The documentation of `upsert` and `findOrCreate` is now updated. I think you meant `findOrCreate` instead of `bulkCreate`.

lib/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
javiertury: I think a warning is more adequate than an error.

There are some cases in which offset without limit is ok. For instance, let's say your app manages collections of elements such as folders and files, song lists or photo albums. The main page shows a sample of elements from each collection using `SELECT values FROM mytable LIMIT 5`. The user can press the *show all* button to load the rest of the collection without limit using `SELECT values FROM mytable OFFSET 5`.

Perhaps this is anecdote, but sometimes LIMIT clauses are not used for pagination and instead are used for defensive programming. In this case ORDER BY is not required. Let's say your app stores the number of siblings, cars, houses or languages that a person has. A person will typically have around 1 or 2 siblings, cars, houses, languages... They will seldom have 10 or 20. And you know they will never have more than 100. The limit clause protects you from poorly constructed queries(like `id = NULL`) that bring thousands/millions of rows and put the DB to its knees.

javiertury:
I took the freedom to replace `if (options.limit !== null && options.limit !== undefined)` with `if (options.limit)`. I think this simplifies comparison and unifies behavior across dialects. `LIMIT 0` has different meanings across dialects, in mysql it means no limit while in postgres it always returns 0 rows.

lib/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
mkaufmaner: That I did ;)

lib/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
mkaufmaner: See https://github.com/sequelize/sequelize/issues/10331

mkaufmaner:
@javiertury Can you change `if (options.limit)` back to `if (options.limit !== null && options.limit !== undefined)`?  A new issue has been created here https://github.com/sequelize/sequelize/issues/10331

mkaufmaner:
Also, I think this PR needs more coverage with the tests.  Once the tests pass, take a look at whats missing in the coverage report.

javiertury:
The functions that stringify values into sql statements became untested since all user-facing queries bind values. Values are only stringified for administrative queries like creating/altering column default values, indexes and constraints.

I added new tests that cover column default values for all `dataTypes`.

lib/dialects/abstract/query-generator/composition.js
```diff
@@ -0,0 +1,61 @@
...
+        item instanceof Placeholder) {
+        a.push(item);
+        return a;
+      }
+      if (item instanceof this.constructor) {
+        Array.prototype.push.apply(a, item.items);
+        return a;
+      }
+
+      throw new Error(`Invalid query item: ${JSON.stringify(item)}`);
```
sushantdhiman: `inspect` ?

lib/dialects/mariadb/query.js
```diff
@@ -59,11 +59,11 @@ class Query extends AbstractQuery {
       queryBegin = Date.now();
     } else {
       this.sequelize.log(
-        `Executing (${this.connection.uuid || 'default'}): ${this.sql}`,
+        `Executing (${this.connection.uuid || 'default'}): ${this.sql} with parameters ${JSON.stringify(parameters)}`,
```
sushantdhiman: A few places where I have seen this now, we should probably use `Logger` class to inspect or probably use something central/configurable like `sequelize.inspect` IIRC it should already be there. My reasoning is, sometimes `parameters` are too big to log so users can override `sequelize.inspect` or other central inspect method, thus can make sense out of big block of text

sushantdhiman:
> I find it impossible to use a different data type parser on node-mysql2 for BinaryRow

Correct, one time I tried bind everywhere, was blocked by this. At that time there was no mariadb connector :) So we only need `mariadb` connector now for both MySQL/MariaDB, right?

mkaufmaner:
@sushantdhiman From what I have gathered, mariadb connector replaces the mysql connector.

javiertury:
@sushantdhiman yes `mariadb-connector-nodejs` is used for both mysql and mariadb.

### Proposed `inspect` solution

I can reach a middleground. In a logging/debug context, values will be inspected using `sequelize.options.inspector`. This inspector can be configured by the user. However, error throwing will use a hard-coded inspector.

Will this solution be acceptable? If so, which default and hard-coded value inspector do you(@sushantdhiman and @mkaufmaner) think is better ?

* `value => require('util').inspect(value, false, 3)`
* `JSON.stringify`
* Other

### Arguments

Files like `dialect/connection-manager.js` or `dialect/query.js` have access to `sequelize.options` and inspect values for logging/debug purposes.

In contrast, auxiliary files like `lib/sql-string.js` and `lib/dialects/abstract/query-generator/composition.js` do not have access to the sequelize instance. If the sequelize instance is required, it will run into circular dependencies. The alternatives are:

1. Wrap these files in a function and pass the sequelize instance as an argument. Now these files are imported in this way `require('path/sql-string')(sequelize)`. Their dependents would also need to be passed the sequelize instance, e.g., `require('path/utils')(sequelize)`.
2. Attach these files into the sequelize instance, for instance as `sequelize.sqlString`. Their dependents would need to be attached as well, e.g. `sequelize.utils`.

I think these alternatives add more complexity than functionality.

sushantdhiman:
@javiertury 

1. Add `inspect` method to https://github.com/sequelize/sequelize/blob/master/lib/utils/logger.js#L14, default depth to 3
2. Export that method in same file, like others at bottom
3. Use like `logger` used here https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L9

javiertury:
@sushantdhiman values are now inspected using `logger.inspect` from `lib/utils/logger.js`.

sushantdhiman:
Sorry I still need to review this, if everything went ok this will be the last major PR before v5 is released. I'll be free next Sunday so I'll try to review it then :)

lib/dialects/abstract/query-generator/composition-group.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Useful for grouping and manipulating anonymous compositions. It
+ * holds compositions or single composition items.
+ *
+ * Concatenable methods.
+ */
+
+class CompositionGroup {
+  constructor(...args) {
+    this.compositions = args.slice(0);
```
SimonSchick: The slice isn't needed, as `args` itself is not a reference.

lib/dialects/abstract/query-generator/composition-group.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+  get length() {
+    return this.compositions.length;
+  }
+  add(...compositions) {
+    Array.prototype.push.apply(this.compositions, compositions);
```
SimonSchick: I think `this.compisitions = this.compisitions.concat(compositions);` or `this.composititions.push(...compositions)` would be cleaner than this prototype hackery.

lib/dialects/abstract/query-generator/composition-group.js
```diff
@@ -0,0 +1,52 @@
...
+
+    return this;
+  }
+  merge(group) {
+    this.compositions = this.compositions.concat(group.compositions);
+
+    return this;
+  }
+  slice(...args) {
+    const newObj =  new this.constructor();
```
SimonSchick: Spacing.

lib/dialects/abstract/query-generator/composition-group.js
```diff
@@ -0,0 +1,52 @@
...
+    }, [this.compositions[0]]);
+
+    return this;
+  }
+  merge(group) {
+    this.compositions = this.compositions.concat(group.compositions);
+
+    return this;
+  }
+  slice(...args) {
```
SimonSchick: I think it would be better to make these more explicit.

lib/dialects/abstract/query-generator/composition.js
```diff
@@ -0,0 +1,62 @@
...
+  constructor(...items) {
+    // Should clone *items* array. Should not copy *items* array by reference.
+    this.items = this._processArgs(items).slice(0);
+  }
+  get length() {
+    return this.items.length;
+  }
+  _processArgs(items) {
+    // Should return new array
+    return items.reduce((a, item) => {
```
SimonSchick: Shouldn't this be using `.map`?

lib/dialects/abstract/query-generator/composition.js
```diff
@@ -0,0 +1,62 @@
...
+  Should be simple, secure and composable. It can hold three items: (sql) strings,
+  value *slots* or placeholders. Keep this class simple, for higher abstractions
+  build something else. This should not depend on dialect or db version.
+
+  @private
+ */
+class Composition {
+  constructor(...items) {
+    // Should clone *items* array. Should not copy *items* array by reference.
+    this.items = this._processArgs(items).slice(0);
```
SimonSchick: But `processArgs` returns a new array, there is no need to clone it?

lib/dialects/abstract/query-generator/composition.js
```diff
@@ -0,0 +1,62 @@
...
+      if (item instanceof this.constructor) {
+        Array.prototype.push.apply(a, item.items);
+        return a;
+      }
+
+      throw new Error(`Invalid query item: ${logger.inspect(item)}`);
+    }, []);
+  }
+  add(...items) {
+    Array.prototype.push.apply(this.items, this._processArgs(items));
```
SimonSchick: Prefer `this.items.push(...this._processArgs(items))`.

lib/dialects/abstract/query-generator/composition-group.js
```diff
@@ -0,0 +1,52 @@
...
+    const newObj =  new this.constructor();
+    newObj.compositions = this.compositions.slice(...args);
+    return newObj;
+  }
+  toComposition() {
+    return new Composition(...this.compositions);
+  }
+}
+
+module.exports = CompositionGroup;
```
SimonSchick: Would it be possible you export these via `module.exports.CompositionGroup = CompositionGroup`, makes it easier to write typings without adding any extra work for you :)

lib/dialects/abstract/query-generator/composition.js
```diff
@@ -0,0 +1,62 @@
...
+    return this;
+  }
+  clone() {
+    const clone = new this.constructor();
+    clone.items = this._processArgs(this.items);
+    return clone;
+  }
+}
+
+module.exports = Composition;
```
SimonSchick: See above.

lib/dialects/abstract/query-generator/insert-proto.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+const Composition = require('./composition');
+const QueryProto = require('./query-proto');
+
+class InsertProto extends QueryProto {
+  toComposition() {
+    if (! (this.table && this.table.length)) {
```
SimonSchick: Spacing.

lib/dialects/abstract/query-generator/query-proto.js
```diff
@@ -0,0 +1,49 @@
...
+      return this[key] = new Composition();
+    });
+
+    return this;
+  }
+  clone() {
+    // Should clone Composition, NOT copy shallow references
+    const clone = new this.constructor();
+
+    this.constructor.partNames.forEach(key => {
```
SimonSchick: Prefer `for .. of` in general

lib/dialects/abstract/query-generator/select-proto.js
```diff
@@ -0,0 +1,40 @@
...
+    composition.add('SELECT');
+
+    if (this.attributes.length) {
+      composition.add(' ', this.attributes);
+    } else {
+      composition.add(' *');
+    }
+
+    if (! (this.from && this.from.length)) {
+      throw new Error('Missing FROM clause in select proto query');
```
SimonSchick: Might be good to have a `CompositionError` class.

lib/dialects/mssql/query-generator.js
```diff
@@ -339,37 +340,38 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
...
-        tuples.push(`(${
-          allAttributes.map(key =>
-            this.escape(attrValueHash[key])).join(',')
-        })`);
+        const values = allAttributes.map(key => {
+          if (attrValueHash[key] instanceof Utils.SequelizeMethod) {
+            return this.handleSequelizeMethod(attrValueHash[key]);
+          }
+
+          if (attrValueHash[key] === null || attrValueHash[key] === undefined) return 'NULL';
```
SimonSchick: @sushantdhiman do you have any preference on single line returns, imo forcing them to always be blocks is easier to read.

lib/dialects/mssql/query-generator.js
```diff
@@ -409,8 +411,15 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
     const updateKeys = Object.keys(updateValues);
     const insertKeys = Object.keys(insertValues);
     const insertKeysQuoted = insertKeys.map(key => this.quoteIdentifier(key)).join(', ');
-    const insertValuesEscaped = insertKeys.map(key => this.escape(insertValues[key])).join(', ');
-    const sourceTableQuery = `VALUES(${insertValuesEscaped})`; //Virtual Table
+    const insertValuesEscaped = new CompositionGroup(...insertKeys.map(key => {
```
SimonSchick: Maybe you should allow passing in an array rather than spreading it all the time.

lib/dialects/mssql/query-generator.js
```diff
@@ -782,101 +806,40 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
...
-                            ` FROM ${tables} AS ${tmpTable}${whereFragment})` +
-                          ` AS ${tmpTable} WHERE row_num > ${offset})` +
-                        ` AS ${tmpTable}`;
-        return fragment;
-      }
-      mainFragment = `SELECT ${topFragment}${attributes.join(', ')} FROM ${tables}`;
+    if (options.offset || isLimit) {
+      const offset = options.offset ? new Slot(options.offset) : '0';
+      composition.add('OFFSET ', offset, ' ROWS');
+      if (!isLimit) logger.warn('options.offset was used without options.limit. Most likely this was unintended.');
```
SimonSchick: Imo this should throw.

lib/model.js
```diff
@@ -2284,35 +2291,38 @@ class Model {
...
-        }
 
-        if (_.intersection(Object.keys(err.fields), whereFields).length) {
-          _.each(err.fields, (value, key) => {
-            const name = this.fieldRawAttributesMap[key].fieldName;
-            if (value.toString() !== options.where[name].toString()) {
-              throw new Error(`${this.name}#findOrCreate: value used for ${name} was not equal for both the find and the create calls, '${options.where[name]}' vs '${value}'`);
+        if (err.fields) { 
+          const flattenedWhere = Utils.flattenObjectDeep(options.where);
+          const flattenedWhereKeys = Object.keys(flattenedWhere).map(name => _.last(name.split('.')));
```
SimonSchick: I think you could instead write a quick util to find the last segment after <char> in a string rather than creating an array and throwing it away right away.

lib/model.js
```diff
@@ -2284,35 +2291,38 @@ class Model {
...
+          const whereFields = flattenedWhereKeys.map(name => _.get(this.rawAttributes, `${name}.field`, name));
+          const defaultFields = options.defaults && Object.keys(options.defaults)
+            .filter(name => this.rawAttributes[name])
+            .map(name => this.rawAttributes[name].field || name);
+
+          if (defaultFields) {
+            if (!_.intersection(Object.keys(err.fields), whereFields).length && _.intersection(Object.keys(err.fields), defaultFields).length) {
```
SimonSchick: Should cache `Object.keys(err.fields)`

lib/model.js
```diff
@@ -2284,35 +2291,38 @@ class Model {
...
 
         // Someone must have created a matching instance inside the same transaction since we last did a find. Let's find it!
         return this.findOne(Utils.defaults({
           transaction: internalTransaction ? null : transaction
         }, options)).then(instance => {
-          // Sanity check, ideally we caught this at the defaultFeilds/err.fields check
+          // Sanity check, ideally we caught this at the defaultFields/err.fields check
           // But if we didn't and instance is null, we will throw
-          if (instance === null) throw err;
+          if (instance === null) return Promise.reject(err);
```
SimonSchick: You are already in a promise context, no need to wrap it, you can just throw.

lib/query-interface.js
```diff
@@ -930,6 +930,26 @@ class QueryInterface {
...
+      return Promise.all([
+        this.sequelize.query(update, Object.assign({}, options, { type: QueryTypes.UPDATE })),
+        this.sequelize.query(insert, Object.assign({}, options, { type: QueryTypes.INSERT }))
+      ]).then(([updateRes, insertRes]) => {
+
+        if (insertRes && insertRes[1] || updateRes && updateRes[1]) {
+          return [!!insertRes[1], undefined];
+        }
+
+        return Promise.reject(new Error('Upsert query did not update or insert anything'));
```
SimonSchick: `throw`

lib/utils.js
```diff
@@ -88,6 +91,26 @@ function spliceStr(str, index, count, add) {
 }
 exports.spliceStr = spliceStr;
 
+function replacePlaceholder(composition, replacement, name) {
+  const newObj = new QueryComposition();
+
+  replacement = replacement instanceof QueryComposition ?
+    replacement.items : [replacement];
+
+  newObj.items = composition.items.reduce((a, item) => {
```
SimonSchick: `.map`

lib/utils.js
```diff
@@ -88,6 +91,26 @@ function spliceStr(str, index, count, add) {
...
+    if (item instanceof Placeholder &&
+      (!item.name && !name || item.name === name)) {
+      Array.prototype.push.apply(a, replacement);
```
SimonSchick: See above, use spreading

test/integration/data-types.defaults.test.js
```diff
@@ -0,0 +1,383 @@
...
+  Support = require('./support'),
+  DataTypes = require('../../lib/data-types'),
+  sequelize = Support.sequelize,
+  dialect = Support.getTestDialect();
+
+describe(Support.getTestDialectTeaser('DataTypes'), () => {
+
+  function getInstance(Model) {
+    return sequelize.sync({ force: true }).then(() => {
+      return Model.create();
```
SimonSchick: Prefer arrow return.

test/integration/data-types.defaults.test.js
```diff
@@ -0,0 +1,383 @@
...
+    }).then(record => {
+      return Model.findOne({ where: { id: record.id } });
+    }).then(instance => {
```
SimonSchick: Redundant.

test/integration/data-types.defaults.test.js
```diff
@@ -0,0 +1,383 @@
...
+    it('date', () => {
+      const Model = sequelize.define('Defaults', {
+        date: {
+          type: DataTypes.DATE,
+          defaultValue: new Date(Date.UTC(2000, 0, 1))
+        }
+      });
+
+      return getInstance(Model).then(instance => {
+        expect(JSON.stringify(instance.date)).to.equal(JSON.stringify(new Date(Date.UTC(2000, 0, 1))));
```
SimonSchick: Why not compare `getTime()`?

test/integration/model/findOne.test.js
```diff
@@ -924,9 +930,13 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       });
 
       it('throws error when record not found by findByPk', function() {
+        // Postgres complains about invalid integer
+        const rejectArgs = dialect === 'postgres' ? 
+          [Sequelize.DatabaseError, 'invalid input syntax for integer: "4.732322332323333e+30"'] : [Sequelize.EmptyResultError];
```
SimonSchick: This is kind of unusual, usually you want `.catch` always be called with an error object, not a tuple and instead store error metadata on the error object itself.

test/unit/dialects/mssql/query-generator.test.js
```diff
@@ -59,89 +58,24 @@ if (current.dialect.name === 'mssql') {
     it('bulkInsertQuery', function() {
       //normal cases
       expectsql(this.queryGenerator.bulkInsertQuery('myTable', [{ name: 'foo' }, { name: 'bar' }]), {
-        mssql: "INSERT INTO [myTable] ([name]) VALUES (N'foo'),(N'bar');"
+        query: 'INSERT INTO [myTable] ([name]) VALUES (@0),(@1);',
+        bind: { '0': 'foo', '1': 'bar' }
```
SimonSchick: `'0'` is redundant.

test/unit/model/validation.test.js
```diff
@@ -273,10 +273,13 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), () => {
     });
 
     before(function() {
-      this.stub = sinon.stub(current, 'query').callsFake(() => {
-        return new Promise(resolve => {
-          resolve([User.build({}), 1]);
-        });
+      this.stub = sinon.stub(current, 'query').callsFake(sql => {
+        // Force to call composeQuery, which calls validation
+        return Promise.resolve(current.dialect.QueryGenerator.composeQuery(sql)).then(() =>
```
SimonSchick: I think you are fine just calling this fn and then `return Promise.resolve([...])`?

lib/dialects/abstract/query-generator/select-proto.js
javiertury: Done

SimonSchick:
*Please* don't force push, it makes it hard to view changes inside of a PR.

test/integration/model/findOne.test.js
javiertury: It was there before. Perhaps I moved it around, and git attributes this change to me.

lib/model.js
javiertury: It was there before. But I cached it anyways.

lib/model.js
javiertury: It was there before. I took out `flattenedWhereKeys` out of the equation, it was redundant.

lib/dialects/abstract/query-generator/composition-group.js
javiertury: I moved them to the same file `composition.js` and they are exported as `module.exports.Class = Class` for easier typings.

The classes `Slot`, `Placeholder`, `Composition` and `CompositionGroup` are very connected to each other. The only reason they were in separate files was to use `module.exports = Class`, but it seems to be inconvenient for typing.

I also did put the classes `QueryProto`, `SelectProto`, `InsertProto` and `UpdateProto` together in the file `query-proto.js` and are exported as `module.exports.Class = Class`.

javiertury:
This pull request is large, but I think this is the most convenient way. After a couple of statement generators use binding through `Composition`, every dependent function must use composition elements. Also, git exaggerates the changes I have done. Many lines that show on the diffs have remained equal.

In this last update, I have tried to lower the use of the spread(...) operator. I created the `CompositionGroup.from(array)` static method which should be more performant.

On a side note, both the `Composition` and the `CompositionGroup` classes are nothing but array wrappers. At the beginning they were subclasses of `Array`. Then I realized subclassing *built-in*s is not supported on all platforms.

SimonSchick:
Interesting, according to https://node.green/ you can subclass `Array` just fine on node v6+

sushantdhiman:
@papb I am currently reviewing this PR, first main concern I have is divergence of MySQL and MariaDB APIs.

> Now mariadb-connector-nodejs is used for MariaDB dialect instead of node-mysql2
> - I find it impossible to use a different data type parser on node-mysql2 for BinaryRow
> mysql2.execute() (for prepared statements) API lacks many features in comparison to mysql2.query() API. Not possible to use custom parsers, and no hints about data type.
> - The developers of mariadb-connector-nodejs claim higher performance

I am aware of `mysql2.execute` issues, I propose we can just use `placeholders` from `mysql2` i.e. keep using `mysql2.query`, it is bound to provide same level of security we have currently by escaping queries  at dialect library

lib/dialects/mssql/query-generator.js
sushantdhiman: Before v5 is released let's setup `prettier` so everything looks good

lib/dialects/abstract/query-generator/composition.js
```diff
@@ -0,0 +1,153 @@
...
+ * It can be converted to an escaped sql string or to a binded param. It also
+ * must be self contained, the escaper or binder should be part of the slot. It
+ * is identified for being an instace of this class, otherwise reject (reflects
+ * unsafe handling)
+ *
+ */
+class Slot {
+  constructor(value, field, options) {
+    this.value = value;
+    this.field = field;
```
sushantdhiman: Why `field` is required, what does `options` contain

lib/dialects/abstract/query-generator/query-proto.js
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    return clone;
+  }
+  toComposition() {
+    throw new Error(`Class ${this.constructor.name} has not implemented toComposition method`);
+  }
+}
+
+QueryProto.partNames = [];
```
sushantdhiman: You can define it as a static getter in class itself, for clarity

lib/dialects/mariadb/query.js
```diff
@@ -355,4 +355,6 @@ class Query extends AbstractQuery {
   }
 }
 
+Query.outputName = 'MariaDB';
```
sushantdhiman: This can be removed if we decided to keep mysql2 for MySQL servers

lib/dialects/mssql/query-generator.js
```diff
@@ -782,101 +804,40 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
...
+  bindParam(bind) {
+    return value => {
+      bind.push(value);
+      return `@${bind.length -1}`;
+    };
   }
 
-  booleanValue(value) {
-    return value ? 1 : 0;
+  outputBind(array) {
```
sushantdhiman: An explanation as comment why MSSQL require this treatment, I know it is some format difference but please specify that difference so this method can be safely changed in future

lib/dialects/postgres/index.js
```diff
@@ -21,7 +21,6 @@ class PostgresDialect extends AbstractDialect {
 
 PostgresDialect.prototype.supports = _.merge(_.cloneDeep(AbstractDialect.prototype.supports), {
   'DEFAULT VALUES': true,
-  'EXCEPTION': true,
```
sushantdhiman: Why this was removed?

lib/model.js
```diff
@@ -2216,8 +2216,12 @@ class Model {
...
    *
    * If no transaction is passed in the `options` object, a new transaction will be created internally, to prevent the race condition where a matching row is created by another connection after the find but before the insert call.
-   * However, it is not always possible to handle this case in SQLite, specifically if one transaction inserts and another tries to select before the first one has committed. In this case, an instance of sequelize. TimeoutError will be thrown instead.
-   * If a transaction is created, a savepoint will be created instead, and any unique constraint violation will be handled internally.
+   * If a transaction is passed, a savepoint will be created instead, and any unique constraint violation will be handled internally.
+   *
+   * **Dialect specificities:**
+   *
+   * * SQLite: A TimeoutError will be thrown if one transaction inserts and another tries to select before the first one has committed.
+   * * PostgreSQL: requires PostgreSQL 9.5+ to correctly handle multiple concurrent findOrCreate over the same row inside the same transaction. Insert makes use of `ON CONFLICT DO NOTHING` not to abort the transaction.
```
sushantdhiman: This is a new limitation?

lib/dialects/abstract/query-generator/query-proto.js
javiertury: Good idea

lib/dialects/abstract/query-generator/composition.js
javiertury: A `slot` doesn't know if it will be binded or stringified. Depending on whether the slot is binded or stringified, the final value for some dataTypes will be different. For instance `N'abc'` if stringified or just `abc` if binded.

The field argument provides the `stringify` and `bindParam` methods. The `options` argument is used to as `field.stringify(value, options)`.

This helps dealing with very complex dataTypes. For instance, take a look at `Array.prototype._stringify` inside `lib/dialects/postgres/data-types.js`. This function if stringified requires to know the database field name, which is passed through options.

javiertury:
@sushantdhiman, I see your point. Changes in the client/server protocol could create problems between `mariadb-connector-nodejs` and a mysql server.

Luckily, the content of SQL statements won't cause incompatibilities since they are not processed by the client(except for comments).

`node-mysql2` placeholders don't provide the same level of security as binded parameters. Placeholders are equivalent to stringifying values, except it's `node-mysql2` the one that stringifies them, not Sequelize.

lib/model.js
sushantdhiman: I thought about this, looks like a major disadvantage to me. How would you recommend communicating this to developers?

sushantdhiman:
> `node-mysql2` placeholders don't provide the same level of security as binded parameters. Placeholders are equivalent to stringifying values, except it's `node-mysql2` the one that stringifies them, not Sequelize.

I agree. We will remain at same level of protection against SQL injections.

> Perhaps we can let users choose between `node-mysql2` and `mariadb-connector-nodejs`. `node-mysql2` with placeholders could be the default, so that users don't blame Sequelize if `mariadb-connector-nodejs` breaks compatibility with mysql. Just like the libraries `pg` and `pg-native` for postgres.

Otherwise users can run MySQL database against our `dialect: maridb`. So we can keep `mysql` untouched

javiertury:
[@sushantdhiman](/sushantdhiman)

The connector `node-mysql` has been restored for dialect mysql. It binds all queries except `QueryInterface.select()`, which needs `typeCast`. These select queries have their parameters formatted into the sql string. Hopefully `node-mysql` will eventualy add `typeCast` support on the `.execute()` interface.

Regarding postgres version support, the following warning is displayed every time `findOrCreate` uses a shared transaction on postgres 9.4 or lower.

```
logger.warn('findOrCreate may abort shared transactions unexpectedly on postgres 9.4 or lower');
```

javiertury:
After adding some tests, coverage has increased.

lib/model.js
javiertury: The next review displays a warning every time findOrCreate uses a shared transaction on postgres 9.4 or lower.

lib/dialects/sqlite/query.js
```diff
@@ -66,10 +64,10 @@ class Query extends AbstractQuery {
   }
 
   _handleQueryResponse(metaData, columnTypes, benchmark, queryBegin, err, results) {
-    debug(`executed(${this.connection.uuid || 'default'}) : ${this.sql}`);
+    debug(`executed(${this.connection.uuid || 'default'}) : ${this.sql} with parameters ${logger.inspect(this.options.bind)}`);
```
SimonSchick: `util.inspect` is rather expensive function to call, we should really not call it unless we really need to in production code.
I think a better approach would be to redirect all `debug` calls to the logger itself and then check if the debug message should be emitted before doing the expensive calls.
This would look like `logger.debug({ action: string/*executed etc.*/, sql, bound?, connId?, ... })`
This would allow us to dedupe a fair chunk of logging code too!

If you want I could take a stab at this.

javiertury: I see the problem. `logger.inspect` is always called even if debug is not enabled, which drags performance. I like the solution of using `logger.debug`. And yes, it'd be great if you could help me with this.

lib/dialects/mssql/query-generator.js
javiertury: What I am afraid is breaking people's programs. Perhaps we can issue warnings on the current sequelize version and throw errors on the next one, to developers time to catch up.

SimonSchick: Isn't this meant to go into v6 anyways?

lib/dialects/sqlite/query.js
javiertury: II have been thinking about the logging issues. The concerns raised at #10630 are difficult to solve without debug.

I didn't know that `debug` had [formatters](https://www.npmjs.com/package/debug#formatters) and more importantly, that they are [configurable](https://www.npmjs.com/package/debug#environment-variables).

For instance:

```js
debug('executed(%s) : %s with parameters %O', this.connection.uuid || 'default', this.sql, this.options.bind);
```

I think they can satisfy the requirements of @sushantdhiman (configurability), @SimonSchick (performance) and this pull request (verbose query params). What do you think?

lib/dialects/mssql/query-generator.js
javiertury: If you don't mind, I prefer to implement this in a separate pull request. I have opened a separate issue #10649 to implement this feature.

SimonSchick: That's fair.

lib/dialects/sqlite/query.js
javiertury: Thanks for your patch, I've merged it with this pull request.

SimonSchick:
Still a little concerned with the amount and complexity of the involved changes, is there any way to split it into smaller chunks?

javiertury:
I think it will be quite some extra work, but I will give it a try. Each commit should be able to pass tests. However some commits will have a temporary negative impact on test coverage.

The reason is, the commits will provide latent functionality that is only enabled and tested once other commits are ready. Also, some commits may temporarily introduce "duplicated" functionality

mschipperheyn:
This amazing. Cant wait to see this merged.

mschipperheyn:
Looks like this PR has stalled because of the request to split into smaller chunks (which sounds like a humongous task to me). This sucks. Sequelize needs this.

papb:
@sushantdhiman

papb:
@javiertury This PR has conflicts, can you please rebase?

javiertury:
@papb this feature is very time consuming. I want to get it merged, rather than to stall completely. But before investing more hours, we should first have a clear plan.

Last time it was decided to merge this little by little into a new branch, [sequelize:feat/compositions](https://github.com/sequelize/sequelize/tree/feat/compositions). However it is stalling.

What if instead, I update and merge everything at once into a new branch(e.g. `sequelize:feat/bind-params`)?

The idea is to allow everyone to work simultaneously on the *full implementation* at the same time. If enough people commit to this plan, we can have multiple reviewers and multiple coders working simultaneously and autonomously.

After it becomes sufficiently polished, we can merge it into `master`.

papb:
@javiertury Makes sense to me - but this kind of thing is up to @sushantdhiman :grimacing:

mschipperheyn:
Anyone, anyone, Buehler, Buehler?

mschipperheyn:
@sushantdhiman as this PR flounders, it's getting more and more conflicts with the master 👎 Really disappointed in the way this crucial change is being ignored. I remember the day when the CTO from the client saw the string interpolation on sequelize. He practically exploded and wanted to throw sequelize out then and there.

papb:
@mschipperheyn I'm sorry... Our time is really short. Hopefully we will be able to pick up on this soon...

Issue #10265, opened 2018-12-13
Binary UUID : PR Wanted? 


bradennapier:
#9272

I wrote out a skeleton for how we can support UUID Binary type as recommended by MySQL team and many other articles to store uuid as binary(16).  

I’m happy to look into implementing it as a PR for official support as a data type if it would be welcomed by the team as an addition.  

My time is somewhat limited in coming months so don’t want to do unless it’s a desired feature :) 

Specifically take a look at the last post where I implement the uuid stuff client side.

sushantdhiman:
Yes

bradennapier:
@sushantdhiman So I put that PR together but realized there are probably quite a few places that I would be missing potentially important pieces.  I extracted the core values into two packages and published them to npm and tested.

---

Specifically around how to handle the new `BINARY` type (I played with using `UUIDV1.BINARY` as the default value for this which seems to work then having a `BINARY` type as the module below defines.

I wasn't sure how to best handle the `virtual` field in this case in a way that works similar to precedent.  It's fairly important IMO to be able to still get the human-readable values.  Seems some kind of transform hook would be good to always provide to the user as a string but always save to db as the binary.  This may already be available that I may be missing.

---

https://github.com/odo-network/binary-uuid

https://github.com/odo-network/sequelize-binary-uuid

```javascript
import Sequelize from "sequelize";
import { withBinaryUUID } from "sequelize-binary-uuid";

const User = sequelize.define(
  "User",
  withBinaryUUID(
    {
      // any other fields here
      someKey: Sequelize.DataTypes.TEXT
    },
    {
      primaryID: "id", // default
      virtualID: "uuid", // default
      field: {
        primaryKey: true
      }
    }
  )
);
```

Issue #9271, opened 2018-04-05
Suggestion: Please support DataTypes.UUID.BINARY


omidkrad:
From the [documentation](http://docs.sequelizejs.com/variable/index.html#static-variable-DataTypes)

> The `CHAR` and `STRING` types expose the `BINARY` property

I think this is also needed for `UUID` data type so we can store it as `BINARy(16)` is MySQL.

for example:

```javascript
sequelize.define('model',` {
  uuid: {
    type: DataTypes.UUID.BINARY,
    defaultValue: DataTypes.UUIDV4,
    primaryKey: true
  }
})
```

heisian:
@omidkrad A UUID is 36 characters long. Maybe try that?

bradennapier:
I agree this should be added!  mysql even recommends it.  

@omidkrad this appears to work at first glance but im still have some issues overall so not sure im 100% there yet!

```javascript
    id: {
        type: 'binary(16)',
        primaryKey: true,
        defaultValue: Sequelize.fn(
          '',
          Sequelize.fn('UUID_TO_BIN', Sequelize.literal('UUID()'), true),
        ),
      },
```

bradennapier:
**Update:** 

So I fixed the issues and it runs and works as needed now except for the fact that I have to turn constraints off on all my associations.  Now quite sure why right now as not needed otherwise. 

Definitely Sequelize should support binary types by default.  It would also be important to run `BIN_TO_UUID()` on the result when selecting the result:

```
SELECT BIN_TO_UUID(id) FROM t;
```

bradennapier:
So after tinkering all day on this I got it working well but there are some tradeoffs.  I personally didn't want to implement the UUID stuff myself which is the better choice in the long-run but some of this would still be needed anyway.

I may be missing some important pieces still but its working with my db well thus far.

```javascript
db.sql.define(
    MODEL_NAME,
    {
      id: {
        type: 'binary(16)',
        allowNull: false,
        primaryKey: true,
      },
      uuid: {
        type: new Sequelize.VIRTUAL(Sequelize.STRING, ['id']),
      },
...
```

then each model using it needs to have some hooks defined:

```javascript
export async function afterCreate(instance) {
  if (!instance.get) return;
  const originalID = instance.get('id');
  const [value] = await this.sequelize.query('SELECT BIN_TO_UUID(:binary, true) as uuidString', {
    type: this.sequelize.QueryTypes.SELECT,
    replacements: { binary: originalID },
[...]

Issue #4997, opened 2015-12-08
MySQL UUIDs stored as CHAR(36)


nimish:
UUIDs are 128 bit integers, which means they should fit into BINARY(16), but sequelize generates CHAR(36) strings instead-- a rather huge increase in size.

Is there a way to get binary(16) support in?

nathanmarks:
For reference:

http://iops.io/blog/storing-billions-uuid-fields-mysql-innodb/
http://mysql.rjweb.org/doc.php/uuid

mickhansen:
Assuming this is not a breaking change it shouldn't be too tricky to change: https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/data-types.js#L45

rafaelalmeidatk:
I began working on this issue and the database is already using the `BINARY(16)` format, but it shows the binary format to the user. How can I apply a function to the field before returning it?

Rudedog9d:
@rafaelalmeidatk Could you leverage `BIN_TO_UUID` to convert the bytes to UUID?

Check out this article for more: https://mysqlserverteam.com/mysql-8-0-uuid-support/

bradennapier:
@Rudedog9d see that #10290 already implements that natively without the need.  In addition you can utilize it without the need for MySQL / MySQL 8 by utilizing the libraries i built for them:

https://github.com/odo-network/sequelize-binary-uuid

https://github.com/odo-network/binary-uuid

which implement an identical effect.

omidkrad:
@bradennapier thanks, this is great. I wished I could use [sequelize-binary-uuid](https://github.com/odo-network/sequelize-binary-uuid) with [sequelize-typescript](https://github.com/RobinBuschmann/sequelize-typescript). If this gets added to sequelize, there's a good chance that they also add it to typescript-sequelize.

bradennapier:
@omidkrad well the good news is my company is switching to 100% typescript so... I’d imagine it will be supported soon :)

spencerwasden:
@bradennapier 
> @omidkrad well the good news is my company is switching to 100% typescript so... I’d imagine it will be supported soon :)
> 
> They just decided to wait to see if there’s enough demand for binary uuid in this way so make sure you indicate your support for it officially as well :)

Any developments on using this with sequelize-typescript?

bradennapier:
> @spencerwasden 
> Any developments on using this with sequelize-typescript?

[...]

PR #10290, opened 2018-12-21
Feature/binary UUID

bradennapier:
### Pull Request check-list

### Description of change

#10265 #9271 #4997 

This **starts** to implement a binary uuid type.  It is meant to open the general discussion around what may be missing.  I have implemented a working package as a module here to give a better idea: https://github.com/odo-network/sequelize-binary-uuid

The benefit is fairly massive (@see: https://www.percona.com/blog/2014/12/19/store-uuid-optimized-way/)

This PR implements the `UUIDV1` shuffling as recommended by above as well as mysql team directly to greatly improve performance while indexing UUIDs.

### Example

```javascript
const User = sequelize.define(
  'TableName',
  {
    id: {
      primaryKey: true,
      type: Sequelize.DataTypes.BINARYUUID,
      defaultValue: Sequelize.DataTypes.UUIDV1.BINARY
    }
  }
);
```

image

---

### `DataTypes.UUID.BINARY`

I could not make it work when trying to use `UUID.BINARY` as it was breaking when going to dialects and not passing options on, among other things... probably makes the most sense there but way too many side effects so I changed to using `BINARYUUID` type.

### Virtual / Transformed Field

In https://github.com/odo-network/sequelize-binary-uuid it also provides a virtual representation of the field in a fairly straight-forward way.  I wasn't sure the best way to do something similar here.

https://github.com/odo-network/sequelize-binary-uuid#using-withbinaryuuid-preset-recommended
https://github.com/odo-network/sequelize-binary-uuid#creating-a-uuid-virtual-field

Specifically since the `uuidv1` is transformed to binary - there should be a simple and easy way to view the field as its non-binary form.  

### v5 breaks `X'${hex}'` handling

It would appear that v5 breaks the handling of anything using `X'...` - I haven't tested but as far as I can tell this would break a few areas such as `BLOB` but I may be wrong.  In the end it seems that stringify just doesn't make sense to use.

I was hoping it would work though as it makes the LOGGING much more helpful than just putting up question marks like it does now.


bradennapier:
I am not really sure why MACADDR is being parsed as binary(16) - they have nothing to do with each other at all... so will need some clarification as-to why it thinks that.

lib/data-types.js
```diff
@@ -789,14 +790,38 @@ UUID.prototype.validate = function validate(value, options) {
   return true;
 };
 
+/**
+ * A column storing a modified uuidv1 as binary(16).
+ * @see https://www.percona.com/blog/2014/12/19/store-uuid-optimized-way/
+ *
+ * @namespace DataTypes.BINARYUUID
+ */
+function BINARYUUID() {
```
SimonSchick: @sushantdhiman I was wondering if you want to enforce these to be classes (and be called with `new`) in v5?

lib/data-types.js
```diff
@@ -789,14 +790,38 @@ UUID.prototype.validate = function validate(value, options) {
...
+
 /**
  * A default unique universal identifier generated following the UUID v1 standard
  *
+ * @param {boolean} [binary=false] Whether UUID should be transformed to binary(16)
  * @namespace DataTypes.UUIDV1
  */
-function UUIDV1() {
-  if (!(this instanceof UUIDV1)) return new UUIDV1();
+function UUIDV1(binary) {
```
SimonSchick: I suggest you make this parameter an object instead, avoid future breaking changes if we wish to add more optional parameters.

lib/utils.js
```diff
@@ -8,7 +8,7 @@ const uuidv4 = require('uuid/v4');
 const Promise = require('./promise');
 const operators  = require('./operators');
 const operatorsArray = _.values(operators);
-
+const toBinaryUUID = require('binary-uuid').toBinaryUUID;
```
SimonSchick: Is it really worth it adding yet another dependency for a very niche data type like this?
Wouldn't it make more sense to split special datatypes and dialects into smaller packages?

lib/data-types.js
bradennapier: I'd prefer it!  Was just following what I saw previous!

lib/utils.js
bradennapier: The dependency was written for this PR specifically - just left it as its own package as it was small and could be used in other places easily.

sushantdhiman:
For now let's keep this as a separate package @bradennapier , if there is more demand we will add its support to Sequelize

omidkrad:
@sushantdhiman Yes, there is demand. Thanks!

NickCoad-NewsPerform:
I'm surprised this was called a niche use case. This package implements UUIDs in the way they're supposed to be implemented when used for primary keys, and UUIDs as primary keys are certainly not niche or uncommon.

Another +1 here, this would be invaluable for us.

bradennapier:
In the meantime, the package does work well :) - but agreed it should be built in.

Seth10001:
+1 for me too, I would like to see this included!

PR #10293, opened 2018-12-24
flattenObject function throws uncaught error when it recursively flat…

skellertor:
…tens objects and a value is null. Added a check to make sure the value is not null before recursively calling the flattenObject function again.

### Pull Request check-list

### Description of change

Added an additional check for null when flattening objects in the util.js file. Since null is a `typeof` object, the function would call itself and try to use the `Object.keys` function on null, which would then throw an error that was covering a validation error.


SimonSchick:
Can you add a test for this?

skellertor:
@SimonSchick Tests written

sushantdhiman:
Thanks @skellertor

PR #10310, opened 2019-01-02
fix(instance-validator): don't skip custom validators on null (#9143)

### LINKED ISSUE #9143: Do not skip custom validators if value is null and allowNull is true
In 2013, the pull request #601 was created to resolve some issues such as #270.

@tremby said ([source](https://github.com/sequelize/sequelize/issues/270#issuecomment-14104551))

> I'd like to see this implemented, but I don't think it should even be an option. If allowNull is true, clearly you want to allow null. If it's not null, only then should the other validations happen.

Five years later, I would like to respectfully disagree. It should be an (opt-out, of course) option. Below I will present a situation where the mandatory aspect of this change is preventing me from doing what I want.

My use case is that I want to allow null only in some cases. Therefore, neither "always" (accomplished with `allowNull: true`) nor "never" (accomplished by `allowNull: false`) suit me, and since PR #601 made this a definitive check, any attempt of providing custom validators don't work.

```js
myField: {
  validate: {
    custom(value) {
      if (value === null && !foobar(this.otherField1, this.otherField2)) {
        throw new Error("myField can't be null when otherField1 and otherField2 satisfy property X");
      }
    }
  }
}
```

When I attempt to create the instance giving null to that field, my custom validator above does not execute at all, no matter what value `allowNull` is set to.

Therefore this should be an option instead of a forced behaviour.

papb:
### Pull Request check-list

### Description of change

Fixes #9143. See [this comment](https://github.com/sequelize/sequelize/issues/9143#issuecomment-450890321) where the changes created by this PR are explained.

This is technically a breaking change but I believe the amount of affected people will be minimal. Regardless, migration instructions were added to docs/upgrade-to-v5.md and the documentation was properly updated, of course. Also, I since v5 is in beta, I think a minor breaking change is fine at the moment :grin:


test/unit/model/validation.test.js
```diff
@@ -604,4 +604,148 @@ describe(Support.getTestDialectTeaser('InstanceValidator'), () => {
...
+
+    const customValidator = sinon.fake(function(value) {
+      if (value === null && this.age !== 10) {
+        throw new Error("name can't be null unless age is 10");
+      }
+    });
+
+    describe('with allowNull set to true', () => {
+
+      const User = current.define('user', {
```
sushantdhiman: Move this to `before`, use as `this.User` in assertions

sushantdhiman: Same goes for other tests

sushantdhiman:
Thanks @papb

PR #10312, opened 2019-01-03
Docs and more tests for #9087

papb:
### Pull Request check-list

### Description of change

The PR #9735 solved the issue #9087 very nicely. I was thinking that it deserved more tests, to help preventing regressions, and also deserved one section of the tutorial docs to be better explained.

This way, in this PR I have added a strong test to prevent regressions, and also some extra documentation.

### DIFF
--- a/docs/scopes.md
+++ b/docs/scopes.md
@@ -121,7 +121,7 @@ Project.scope('defaultScope', 'deleted').findAll();
 SELECT * FROM projects WHERE active = true AND deleted = true
 ```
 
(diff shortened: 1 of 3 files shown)


papb:
Note: for the tests I used a library called [js-combinatorics](https://github.com/dankogai/js-combinatorics), and for that I added it to package.json. According to [packagephobia](https://packagephobia.now.sh/result?p=js-combinatorics@0.5.4) it is only 62 kB, so I think it should be fine. It is very useful to quickly assert that the order in which the merge is applied does not matter (which is something I wanted my test to check).

If adding extra dev-dependencies like this is not good, I can hard-code all the permutations directly, but that would be a bit ugly and unreadable (in my opinion).

sushantdhiman:
Thanks @papb :+1:

Issue #6725, opened 2016-10-19
TypeError: Cannot read property '2' of null


abartolo:
## What you are doing?

Creating a record with the same value in a column that is Unique:true. I'm using FeathersJS with uses Sequelize. I successfully inserted new record via restful api using Postman.  However if I send the same request again( create a user with the same email which the column is unique) it crashes the server with error "TypeError: Cannot read property '2' of null".

Below is the model I am using

``` js
// code here
const Sequelize = require('sequelize');

module.exports = function(sequelize) {
  const user = sequelize.define('users', {
    email: {
      type: Sequelize.STRING,
      allowNull: false,
      unique: true
    },
    password: {
      type: Sequelize.STRING,
      allowNull: false
    }
  }, {
    freezeTableName: true
  });
```
## What do you expect to happen?

I expected to get a response back with an error stating its a duplicate or at least a clean error.
## What is actually happening?

Crashed server since error wasn't caught.

_Output, either JSON or SQL_

![image](https://cloud.githubusercontent.com/assets/5394228/19537833/7c84e15c-961f-11e6-9f0e-ec5aaa9a72f3.png)

**Dialect:** mysql
**Database version:** 5.7.13
**Sequelize version:** 3.24.4

felixfbecker:
That stack trace does not mention sequelize anywhere. If you still believe this is a sequelize bug, create an [SSCCE](http://www.sscce.org/). You can find a template [here](https://github.com/sequelize/sequelize/blob/master/sscce_template.js)

janmeier:
@felixfbecker Actually it does ;), [here](https://github.com/sequelize/sequelize/blob/v3/lib/dialects/mysql/query.js#L132)

@abartolo Which language is your mysql installation? Currently we only support parsing messages in english?

In any case, we need a guard before accessing the match, checking if anything was actually matched

felixfbecker:
Sorry, overlooked that on the screenshot.

abartolo:
@janmeier Mysql is installed in english.

So last night I tried to replicate the issue in my home environment since this was at work, and it worked! Sooo.... I assumed perhaps the installation of node_modules went wrong since my workplace has a lot of firewalls and strict security. But after reinstalled dependencies I still got the issue.

The DB at my home env is 5.6.x and I doubt Mysql version is the issue so I looked at my node version found I'm using v4.2.1 @ work. This is a red flag since FeatherJS documentation is stating nodeV 5+ and if anything is below is required to use babel.

[...]

PR #10320, opened 2019-01-05
repaire a bug; when Duplicate with \r ,that makes match() return null

zhangshichun:
### Pull Request check-list

- [ yes ] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [ yes ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ no ] Have you added new tests to prevent regressions?
- [ no ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ no ] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?

### Description of change
as this issue
https://github.com/sequelize/sequelize/issues/6725
when get a \r in the err.message，match() will return null


SimonSchick:
Where in the error message would this newline be, when are newlines contained in error messages, please provide more info, this seems like a larger issue.

zhangshichun:
> Where in the error message would this newline be, when are newlines contained in error messages, please provide more info, this seems like a larger issue.

with using mysql，this may case the error:
`user.create({
name: 'john\r'
});
user.create({
name: 'john\r'
});
`
the error code is 1062，and the match() just return null.

SimonSchick:
In this case your fix is not sufficient.
There are many occurrences where `err.message` is matched via regex and `.replace(string, string)` only replaces a single occurrences, my suggestion is you add a `util` method called `removeNewlines(str) { return str.replace(/[\r\n]/g, '');` and apply it to all of them.

zhangshichun:
> In this case your fix is not sufficient.
> There are many occurrences where `err.message` is matched via regex and `.replace(string, string)` only replaces a single occurrences, my suggestion is you add a `util` method called `removeNewlines(str) { return str.replace(/[\r\n]/g, '');` and apply it to all of them.

do you think that is a good idea to change
`(.*)`
to
`([\s\S]*)`
when match err.message?

SimonSchick:
That could work too, depending on @sushantdhiman stance on v5 node version we could also use the `/s` flag (dotall) if we were to target v10.

sushantdhiman:
Hehe, at this point I just want to release v5 asap. Please just use simple Node 6 constructs for now

SimonSchick:
Some tests randomly fail due to timing issues, you can ignore these.

knoxcard:
v5 really is the way to go!

PR #10355, opened 2019-01-16
refactor(errors): one file per error

papb:
### Pull Request check-list

### Description of change

All errors were defined in a single huge file. This PR refactors it so that every error has its file, and uses a folder structure for the nested error types.

I encountered this when I was about to contribute with a new error, but I felt like refactoring this first.


SimonSchick:
I don't really see the advantage of this, the error logic is far from complicated and all this does is increase load time.

papb:
@SimonSchick - I just tested on my machine here, using [time-require](https://www.npmjs.com/package/time-require):

* It took 345.7 ms on average (out of 10 samples) without this PR
* It took 354.5 ms on average (out of 10 samples) with this PR

I didn't compute the standard deviation for each case explicitly but the values ranged from ~330 to ~370 on both cases. To be honest I was surprised, I thought the load time would increase a lot more! Regardless, I believe this is a very small price to pay (the difference is eaten by the standard deviation!) for better readability and maintainability.

@sushantdhiman Nice catch! You're right. By removing it, esdoc parses correctly the inheritance chain by default! Nice! I just did it.

SimonSchick:
@papb that is weird, when I did the mass cleanups before it increased load times by almost 10% which is why I omitted it from my PR, I guess this depends on the system it's running.

My previous test was using `console.time` on node v10 btw.

You should add trailing newlines btw.

Also @sushantdhiman what's the deal with the manual `captureStackTrace`, is it just to omit a stack frame?

papb:
@SimonSchick - I see... I am on windows on Node 6 (old, I know, I'll do something about it later).

I've just added the trailing newlines as requested :grin:

sushantdhiman:
> what's the deal with the manual captureStackTrace, is it just to omit a stack frame?

@SimonSchick Looks like that

SimonSchick:
@sushantdhiman as a future note, this can be removed once support for v6 is dropped, iirc it was a workaround for prototype shenanigans pre v4.

PR #10359, opened 2019-01-17
chore(package): update generic-pool to 3.5.0

DiegoRBaquero:
Includes some bug fixes, specifically in evicting. Meant for v4

### Pull Request check-list

### Description of change

Update to latest `generic-pool` in v4

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "debug": "^3.1.0",
     "depd": "^1.1.0",
     "dottie": "^2.0.0",
-    "generic-pool": "^3.4.0",
+    "generic-pool": "^3.5.0",
     "inflection": "1.12.0",
     "lodash": "^4.17.1",
     "moment": "^2.20.0",


karthikv:
@sushantdhiman This bug in generic-pool is very significant, and can have a large effect on performance. Prior to @DiegoRBaquero's fixes in generic-pool (see https://github.com/coopernurse/node-pool/pull/242), *every* eviction cycle would cause *every* idle connection to be evicted [1], regardless of the specified idle time (except those needed to meet the minimum size).

The default eviction duration is 10 seconds in v4, so every idle connection would be evicted every 10 seconds. This is further exacerbated by the default minimum pool size of 0, so this would cause constant, periodic evictions leading to many unnecessary reconnects.

Given this information, I think it'd be wise to publish a new version that contains this fix sooner than later, and perhaps mention this issue explicitly in the changelog or release history so users know. Does this sound reasonable to you?

[1] Given that `softIdleTimeoutMillis` is `-1`, which is the default value that's implicitly used by sequelize. See https://github.com/coopernurse/node-pool/issues/192 for details.

DiegoRBaquero:
I'm going to add that, because of https://github.com/coopernurse/node-pool/pull/243, previously, any **1** (idle) connection would be evicted every 10 seconds, but now, up to **3** *really idle*  connections are evicted every 10 seconds. By default.

karthikv:
Thanks for the clarification, @DiegoRBaquero! My comment about every idle connection being evicted is incorrect due to the `numTestsPerEvictionRun` option, and @DiegoRBaquero's explanation above is accurate. Still, having 1 connection evicted every 10 seconds without the fixes in generic-pool 3.5.0 can cause significant performance penalties.

PR #10369, opened 2019-01-21
fix(model/update): propagate paranoid to individualHooks query

miroslavLalev:

### Description of change

I was testing some scenarios and noticed that when:
 1. Have a model with `paranoid: true` set
 2. Have `beforeUpdate` hook defined in the model
 3. Use the static `<Model>.update` method with `individualHooks: true`

The hook does not get called and therefore update statement is not executed. After a bit of investigation, I found the root cause - the `paranoid` option was not propagated to the select statement.

As it was quite straightforward change, I decided to give it a go and created PR directly :)

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2907,7 +2907,13 @@ class Model {
 
       // Get instances and run beforeUpdate hook on each record individually
       if (options.individualHooks) {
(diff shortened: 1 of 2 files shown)


miroslavLalev:
It seems that the AppVeyor CI test failed with something sporadic as I couldn't reproduce it. Is there any way of retriggering it?

sushantdhiman:
Thanks @miroslavLalev

PR #10374, opened 2019-01-22
fix(model): respect 'underscored' and 'underscoredAll' model init options

### LINKED ISSUE #10373: Model init options underscored and underscoredAll ignored for field names
## What are you doing?

When using the `underscored` option, camelCase field names are not underscored into snake_case. __This is tested against v4.42.0 of Sequelize using Postgres 10.4.__ A full repo demonstrating the issue is here: https://github.com/brentjanderson/sequelize-underscore-not-working (

```js

const sequelize = new Sequelize('postgres', 'postgres', 'postgres', {
  host: 'localhost',
  dialect: 'postgres',
});

  userName: Sequelize.STRING,
  birthDay: Sequelize.DATE
}, { underscored: true });

  .then(() => User.create({
    userName: 'janedoe',
    birthDay: new Date(1980, 6, 20)
  }))
  .then(() => {
    return sequelize.query('SELECT * FROM users');
  }).then(([resultSet]) => {
    console.log(resultSet);
    process.exit(0);
  });

```

When defining a model, the `underscored` option will translate model attributes in camelCase to snake_case. Using postgres 10.4.

```sql
-- The underlying table should look like this:

  Column    |           Type
======================================
 id         | integer
 user_name  | character varying(255) -- Note the snake case
 birth_day  | timestamp with time zone -- Note the snake case
 created_at | timestamp with time zone
 updated_at | timestamp with time zone

```

```sql
-- The underlying actually looks like this

  Column    |           Type
======================================
 id         | integer
 userName   | character varying(255)
 birthDay   | timestamp with time zone
 created_at | timestamp with time zone
 updated_at | timestamp with time zone

```

## Additional notes

[v4.42.0 includes the `underscoredIf` function](https://github.com/sequelize/sequelize/blob/v4.42.0/lib/utils.js#L39-L48), but this function is not called when setting each attribute's `field` property.

brentjanderson:
### Pull Request check-list

### Description of change

Fix a regression where `underscored` and `underscoredAll` did not affect the name of the field defined in the database for sequelize model attributes.

fix #10373

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -774,7 +774,7 @@ class Model {
     this.associations = {};
     this._setupHooks(options.hooks);
 
-    this.underscored = this.underscored || this.underscoredAll;
+    this.underscored = this.options.underscored || this.options.underscoredAll;
 
     if (!this.options.tableName) {
       this.tableName = this.options.freezeTableName ? this.name : Utils.underscoredIf(Utils.pluralize(this.name), this.options.underscoredAll);
@@ -955,7 +955,7 @@ class Model {
       definition._modelAttribute = true;
 
       if (definition.field === undefined) {
-        definition.field = name;
+        definition.field = Utils.underscoredIf(name, this.underscored);
       }
 
       if (definition.primaryKey === true) {
--- a/test/unit/model/define.test.js
+++ b/test/unit/model/define.test.js
@@ -26,6 +26,24 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       expect(Model.rawAttributes).not.to.have.property('updated_at');
     });
 
+    it('should underscore camelCased attributes with underscored: true', () => {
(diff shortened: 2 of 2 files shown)


sushantdhiman:
As per docs

> // don't use camelcase for automatically added attributes but underscore style
  // so updatedAt will be updated_at

> Converts all camelCased columns to underscored if true. Will not affect timestamp fields named explicitly by model options and will not affect fields with explicitly set field option

Confusing :laughing: That means this PR will be a breaking change as user supplied attributes are now underscored with `underscore: true`, These options are [fixed](https://github.com/sequelize/sequelize/blob/master/docs/upgrade-to-v5.md#model) in v5 

As the behavior for these options is not properly defined in v4 context I would rather not merge this PR to prevent any breaking changes.

Let me know your thoughts

PR #10378, opened 2019-01-23
feat(model): unbound method call error

papb:
### Pull Request check-list

### Description of change

Sometimes I forget that some sequelize methods rely internally on the `this` object reference, and end up calling them unbounded and getting bad error messages. Some examples:

```js
someInitialSetupAsync().then(sequelize.sync) // TypeError: Cannot read property 'options' of undefined
doSomethingAsync().then(Foo.findAll) // TypeError: Cannot read property 'warnOnInvalidOptions' of undefined
Bar.findAll().then(foo.addBars) // TypeError: Cannot read property 'get' of undefined
```

Now, with experience, as soon as I see such errors I immediately think that I probably forgot to call [bind](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind). But at the first time, it took me a long time.

This PR adds a new error type in sequelize, `UnboundMethodCallError`, and the new behavior is as follows:

```js
someInitialSetupAsync().then(sequelize.sync) // UnboundMethodCallError: Sequelize#sync was called unbounded
doSomethingAsync().then(Foo.findAll) // UnboundMethodCallError: Model.findAll was called unbounded
Bar.findAll().then(foo.addBars) // UnboundMethodCallError: Model#addBars was called unbounded
```

This works by adding a one-line check (`if (!this)`) to each method. I didn't add this check on all methods, because that would be too much work, so I did it only for the public documented methods which can be called without parameters (i.e. have no required parameter) - the rationale being that if a method requires one or more parameters it is unlikely that it will be called unbounded. I also did it for the association mixins, because I made the mistake with them more than once (see the example above), so I thought it might happen to others too.

I also made some performance checks and I am confident to say that one single extra `if (!this)` is super negligible.

### DIFF
--- a/lib/associations/helpers.js
+++ b/lib/associations/helpers.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const sequelizeErrors = require('./../errors');
+
 function checkNamingCollision(association) {
   if (association.source.rawAttributes.hasOwnProperty(association.as)) {
     throw new Error(
@@ -61,6 +63,7 @@ function mixinMethods(association, obj, methods, aliases) {
       const realMethod = aliases[method] || method;
 
       obj[association.accessors[method]] = function() {
+        if (!this) throw new sequelizeErrors.UnboundMethodCallError(`Model#${association.accessors[method]}`);
         return association[realMethod].apply(association, [this, ...Array.from(arguments)]);
       };
     }
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -12,6 +12,7 @@ exports.InstanceError = require('./instance-error');
 exports.OptimisticLockError = require('./optimistic-lock-error');
 exports.QueryError = require('./query-error');
 exports.SequelizeScopeError = require('./sequelize-scope-error');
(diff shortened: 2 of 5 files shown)


sushantdhiman:
I don't think we should handle this. What if you bind different `this` object, should it check for correct `this` type etc.

Do you have an example if someone else is using this sort of error in their library.

papb:
@sushantdhiman thanks for the feedback. If someone binds a different `this` object, they are messing up on purpose, so this PR is not for them. My idea here is to save time from people that actually make a honest mistake (like I did a few times myself already).

I think my examples above are very plausible mistakes (putting the function directly in a `then`

papb:
As for an example on another library, I don't have that, I came up with this myself.

I agree that the benefit is not so big, but the cost is zero (both for performance and code readability), so I think it's worth it.

sushantdhiman:
Sorry but this new error type and additional `this` check doesn't look good to me. Binding with incorrect `this` or forgot to bind context should be very easy to catch at development phase.

PR #10426, opened 2019-02-08
Display message when passing include's model as null/undefined

serge1peshcoff:
### Pull Request check-list

- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? - no API changes here

### Description of change

For now, if you use `include` and pass the `null`/`undefined` as a parameter (for example, in case of a typo, happened to me actually) it would crash with `Cannot read property '_pseudo of undefined` error, which can be not obvious. I've changed it so it'd through the error if the passed object is not truthy and also provided a test for it.

Here's the example of the issue: https://stackoverflow.com/questions/41790321/sequelize-error-unhandled-rejection-typeerror-cannot-read-property-pseudo-of

Example:

```js
const User = sequelize.define('User', {}); // declaring the model
User.findAll({ include: [{
    model: null
]} }); // would throw an error
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -326,6 +326,10 @@ class Model {
   static _conformInclude(include, self) {
     let model;
 
+    if (!include) {
+      throw new Error('The model object passed to "include" is undefined or null.');
+    }
+
     if (include._pseudo) return include;
 
     include = this._transformStringAssociation(include, self);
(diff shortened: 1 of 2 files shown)


SimonSchick:
This seems like an anti pattern, the developer should ensure the data structure is correct, if sequelize would validate every incorrectly passed property the code base would be 10x the size it is now.

Issue #4900, opened 2015-11-20
Support for time INTERVAL data type in  postgres


mark-lester:
I have times > 24 hours, which just broke upon moving to postrgres, so I think I need to implement an INTERVAL data type in data-types.js.

mark-lester:
I added the obvious
```
/**
 * A time interval column
 * @property INTERVAL
 */

var INTERVAL = function() {
  if (!(this instanceof INTERVAL)) return new INTERVAL();
  ABSTRACT.apply(this, arguments);
};
util.inherits(INTERVAL, ABSTRACT);

INTERVAL.prototype.key = INTERVAL.key = 'INTERVAL';
INTERVAL.prototype.toSql = function() {
  return 'INTERVAL';
};
```
and the entry in module.exports.
Unless I get told otherwise I'll submit that shortly. I cant fully test yet as I have a weird bug on constraints and block inserts (it goes away when I bring the block size down so I need to have a proper look at what is being said to postrgres)

mickhansen:
The newer version uses `ABSTRACT.inherts(cb)` afaik, but @janmeier is more knowledgeable here.

janmeier:
Have a look at https://github.com/sequelize/sequelize/blob/master/lib/dialects/postgres/data-types.js for an example of how to extend the abstract datatype. 

You'll need to add an entry in https://github.com/sequelize/sequelize/blob/master/lib/data-types.js as well, like with hstore and json, otherwise the type won't be accessible

janmeier:
ping @mark-lester  :)

mark-lester:
Hey. where did I get to ?, I think I gave up, yes, I aborted and went for strings, because ...,
https://groups.google.com/forum/#!topic/sequelize/IBoT8wf72tg
I am just in the middle of critical phase in today's cooking, let me remember why I bottled out, ..., brb

janmeier:
While other dialects use strings to identify their types (e.g. INTEGER, GEOMETRY etc), postgres uses oids, which are numbers. The oid for interval is 1186, and interval[] is 1187

(You can check for an oid using `select typname, oid, typarray from pg_type where typtype = 'b')

mark-lester:
ok, thanks,
 I am about to return to the server and so will have another punt, I need to to implement a sequence manager, something that does this kind of thing (all totally untested). 

```
sequenceManage:function(offset){
	var self=this;
	var query = "UPDATE "+this.tableName+
[...]

PR #10441, opened 2019-02-13
feat(postgres): add INTERVAL support

gabegorelick:
### Pull Request check-list

### Description of change

Add `INTERVAL` type for PostgreSQL.

See https://github.com/sequelize/sequelize/issues/4900


lib/data-types.js
```diff
@@ -603,6 +603,16 @@ DATEONLY.prototype._isChanged = function _isChanged(value, originalValue) {
...
 };
 
+function INTERVAL() {
+  if (!(this instanceof INTERVAL)) return new INTERVAL();
+}
+util.inherits(INTERVAL, ABSTRACT);
+
+INTERVAL.prototype.key = INTERVAL.key = 'INTERVAL';
+INTERVAL.prototype.validate = function validate() {
+  return true;
```
gabegorelick: I've not found a good way to validate intervals. 

See https://stackoverflow.com/questions/38692972/how-to-validate-a-moment-js-duration for an example of this not working.

lib/dialects/postgres/connection-manager.js
```diff
@@ -186,6 +186,10 @@ class ConnectionManager extends AbstractConnectionManager {
         }
       }
 
+      if (this.sequelize.config.intervalStyle) {
+        query += `SET intervalstyle TO ${this.sequelize.config.intervalStyle};`;
```
gabegorelick: Postgres returns intervals as strings. But we need ISO format to be able to accurately parse them.
See https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-OUTPUT

lib/dialects/postgres/data-types.js
```diff
@@ -87,6 +88,47 @@ module.exports = BaseTypes => {
 
   BaseTypes.DATEONLY.types.postgres = ['date'];
 
+  function INTERVAL(...args) { // TODO precision
```
gabegorelick: Need to figure out precision, but can probably punt on it for first pass?

lib/dialects/postgres/data-types.js
```diff
@@ -87,6 +88,47 @@ module.exports = BaseTypes => {
...
+    if (moment.isDuration(value)) {
+      return value.toISOString();
+    }
+
+    return value;
+  };
+
+  INTERVAL.prototype._sanitize = function _sanitize(value) {
+    if (typeof value === 'string') {
+      return moment.duration(value);
```
gabegorelick: `moment.duration` seemed like the best thing to sanitize to. Alternatively, we could normalize to strings, but then we'd potentially need to decide on a format (ISO, SQL, PG). Or we could normalize to number of milliseconds (or seconds).

docs/data-types.md
```diff
@@ -32,6 +32,8 @@ Sequelize.DECIMAL(10, 2)              // DECIMAL(10,2)
 Sequelize.DATE                        // DATETIME for mysql / sqlite, TIMESTAMP WITH TIME ZONE for postgres
 Sequelize.DATE(6)                     // DATETIME(6) for mysql 5.6.4+. Fractional seconds support with up to 6 digits of precision
 Sequelize.DATEONLY                    // DATE without time.
+Sequelize.INTERVAL                    // Duration of time. PostgreSQL only.
```
gabegorelick: Other DBs have a notion of an interval or duration type. E.g. Oracle has `INTERVAL YEAR [(year_precision)] TO MONTH` and `INTERVAL DAY [(day_precision)] TO SECOND [(fractional_seconds)]`.  This should be generalizable to that. 

Note that many DBs, including PG, take a `fields` option that modifies the interval. For example, `INTERVAL YEAR TO MONTH`. I think supporting that should be out of scope.

lib/data-types.js
gabegorelick: Actually, is `validate` called by anything? I was halfway through updating this to validate ISO and postgres interval strings when I realized nothing seems to call these functions. 

Is that the case?

lib/dialects/postgres/data-types.js
```diff
@@ -87,6 +88,33 @@ module.exports = BaseTypes => {
...
 
+  function INTERVAL(...args) {
+    if (!(this instanceof INTERVAL)) return new INTERVAL(...args);
+    BaseTypes.INTERVAL.apply(this, args);
+  }
+
+  inherits(INTERVAL, BaseTypes.INTERVAL);
+
+  INTERVAL.parse = function parse(value) {
+    return moment.duration(value);
```
steelbrain, line 99: Curious, why not use `parse` from `postgres-interval`? https://github.com/bendrucker/postgres-interval/blob/f58ec4bb28d67fe1f9c9a1ad4ab8066c26de24bc/index.js#L113

PR #3041, opened 2015-02-01
Add support for postgres on io.js

ruimarinho:
Adds support for postgres on io.js.

* Upgrade pg@4.2.0
* Upgrade pg-native@1.8.0

Two important notes:

* There is a new `SET` that increases the message level to `warning` (up from `notice`) since the upgraded pg client is now capable of bubbling these messages.
* There is something going on with timers that I had to workaround with promises. The tests where sinon's fake clock are used hang for some reason. This doesn't seem related to io.js since the lib tests used by sinon to fake timers are passing on the latest io.js version. If anyone is able to figure that out, I'd appreciate it.

PR #10448, opened 2019-02-14
feat(postgres): allow customizing client_min_messages

gabegorelick:
### Pull Request check-list

### Description of change

For some reason, Sequelize sets `client_min_messages` to `warning` when connecting to Postgres. This was introduced in https://github.com/sequelize/sequelize/pull/3041 with the only justification being "There is a new `SET` that increases the message level to `warning` (up from `notice`) since the upgraded pg client is now capable of bubbling these messages." 

`pg` does indeed allow you to [register an event when the server returns a message](https://node-postgres.com/api/client#client-on-39-notice-39-notice-string-gt-void-gt-void), but it doesn't follow that Sequelize needs to override the default `client_min_messages`. 

Furthermore, this behavior accidentally became coupled to the `keepDefaultTimezone` setting in https://github.com/sequelize/sequelize/commit/95a5c0d86dbdd32133a3dcb64659e98f12691b02. When `keepDefaultTimezone` is `true`, `client_min_messages` is not overridden. 

This behavior has been in place so long that it may be safest to keep `client_min_messages` set to `warning` by default. But we can at least allow users to customize it and decouple it from `keepDefaultTimezone`. 

If we want to make a riskier change, we can set `client_min_messages` to `null` by default and have Sequelize default to *not* updating the DB's setting. I have not done that in this PR, but it's an easy change if we want to make it.

### DIFF
--- a/lib/dialects/postgres/connection-manager.js
+++ b/lib/dialects/postgres/connection-manager.js
@@ -177,12 +177,16 @@ class ConnectionManager extends AbstractConnectionManager {
         query += 'SET standard_conforming_strings=on;';
       }
 
+      if (this.sequelize.options.clientMinMessages !== false) {
+        query += `SET client_min_messages TO ${this.sequelize.options.clientMinMessages};`;
+      }
+
       if (!this.sequelize.config.keepDefaultTimezone) {
         const isZone = !!moment.tz.zone(this.sequelize.options.timezone);
         if (isZone) {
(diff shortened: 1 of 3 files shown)


gabegorelick:
Failing test is due to a MySQL deadlock. Almost certainly a flaky test and not caused by these changes in the Postgres connection manager.

PR #10473, opened 2019-02-23
fixes msssql only error when no order by but has offset or limit present

jcvaldes:
encontré un bug en la librería sql cuando se usa limit duplica en la clausula order by el campo y da error
"error when no order by but has offset or limit present" 
con el cambio que puse se solucionó


lib/dialects/mssql/query-generator.js
```diff
@@ -665,7 +665,7 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
    * @returns {string}
    */
   _getForeignKeysQueryPrefix(catalogName) {
-    return `${'SELECT ' +
+      return `${'SELECT ' +
```
SimonSchick: I understand most of the formatting changes you made, this one doesn't make sense tho.

PR #10492, opened 2019-03-02
feat(types): handle Numbers passed as Objects

### LINKED ISSUE #10491: User defined number types not serialized properly
## What are you doing?
```js
  age: Sequelize.BIGINT
});

const age = BigInt(Number.MAX_SAFE_INTEGER).add(Number.MAX_SAFE_INTEGER);

  return User.create({age});
}).then(user => {
  expect(BigInt(user.age).toString()).to.equal(age.toString());
  return User.findAll({
    where: {age}
  });
}).then(users => {
  expect(users).to.have.lengthOf(1);
  expect(BigInt(users[0].age).toString()).to.equal(age.toString());
});
```

Sequelize serializes the [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) as a string and generates valid SQL.

`SequelizeDatabaseError: invalid input syntax for integer: ""18014398509481982""`

__Dialect:__ any (but tested on PG)
__Dialect version:__ any
__Database version:__ any
__Sequelize version:__ master

gabegorelick:
This adds support for Number types that aren't represented as actual Javascript Number instances. For example, [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) (new in Node 10), [Decimal.js](https://github.com/MikeMcl/decimal.js/), etc. Anything with a `.toString` should work.

Fixes #10491

### Pull Request check-list

### Description of change

Fixes #10491

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "@types/bluebird": "^3.5.26",
     "@types/node": "^10.12.27",
     "@types/validator": "^10.9.0",
+    "big-integer": "^1.6.42",
     "chai": "^4.x",
     "chai-as-promised": "^7.x",
     "chai-datetime": "^1.x",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -32,6 +32,7 @@
     "new-cap": [
       "error",
       {
+        "capIsNewExceptionPattern": "^BigInt",
         "properties": false
       }
     ],
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -214,6 +214,17 @@ NUMBER.prototype.validate = function(value) {
 
   return true;
 };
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -53,6 +53,7 @@
     "@types/bluebird": "^3.5.26",
     "@types/node": "^10.12.27",
     "@types/validator": "^10.9.0",
+    "big-integer": "^1.6.42",
```
gabegorelick: Shim for BigInt, which is new in Node 10. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt.

PR #10495, opened 2019-03-02
chore(data-types): move to classes

SimonSchick:
### Pull Request check-list

### Description of change

- Moves all datatypes to be classes while retaining 100% backwards compatibility.
 - No more prototype madness
 - Removed lots of duplicate code
 - Code is *much* cleaner.

Remaining:
 - There are 4 questions:
   - Why does postgres alter base types?
   - Should there be an intermediate class for floating point data types to avoid more prototype patching?
   - The compatibility layer uses a `Proxy` which isn't necessarily very performant but should have no real impact on performance.
   - Should `inferNew` be exposed (as in the new docs)?


SimonSchick:
Arg, latest change caused a circular dependency, fixing..

SimonSchick:
@sushantdhiman I cannot repro these errors about `toString` not being generic, no idea what's going on here...

SimonSchick:
@sushantdhiman works now

SimonSchick:
Please note there is a single potentially 'breaking' change:
`DataType.toString() // static call` will now yield the name of the DataType rather than the class definition, this seems to be a technical limitation of `Proxy`, I hope that's ok.

SimonSchick:
boop

SimonSchick:
Hey @sushantdhiman since you are around, I'd love to get some feedback on this :)

lib/dialects/postgres/data-types.js
```diff
@@ -538,13 +451,7 @@ module.exports = BaseTypes => {
...
   };
 
-  RANGE.prototype.toSql = function toSql() {
-    return BaseTypes.RANGE.types.postgres.subtypes[this._subtype.toLowerCase()];
-  };
-  RANGE.prototype.toCastType = function toCastType() {
-    return BaseTypes.RANGE.types.postgres.castTypes[this._subtype.toLowerCase()];
-  };
-
+  // TODO: Why are base types being manipulated??
```
janmeier: I think it kind of derives from the fact that basetypes.ARRAY is also the postgres.ARRAY type (since only postgres uses the array type) (yes, you could make the same argument about range and hstore, which are doing it the right way :D)

There really should be a specialized postgres.ARRAY type, so it works the same as it does for every other type

SimonSchick: Fixed that!

lib/data-types.js
```diff
@@ -393,30 +352,25 @@ DOUBLE.prototype.key = DOUBLE.key = 'DOUBLE PRECISION';
...
+      throw new sequelizeErrors.ValidationError(util.format('%j is not a valid decimal', value));
+    }
+    return true;
+  }
+}
 
-  return true;
-};
-
+// TODO: Create intermediate class
```
janmeier: :+1:

SimonSchick:
@janmeier Not going to do the `ARRAY` and number base class thing for now as that's a rabbit hole I am *not* willing to dive into any further at this moment.

If you are interested in poking it see my branch `chore/array-to-class` in my fork.

Please approve this PR as is.

lib/utils.js
```diff
@@ -11,6 +11,8 @@ const operatorsArray = _.values(operators);
 
 let inflection = require('inflection');
 
+exports.implicitNew = require('./utils/implicitNew');
```
sushantdhiman: There is a method `classToInvokable` in these utils, which is same as `implicitNew`. Either use that old method or get rid of  `classToInvokable` so we only use one common method

sushantdhiman:
There are some unrelated changes in this PR, please rebase with master

PR #10512, opened 2019-03-06
fix(types): allow specifying additional options for db.query and add missing retry

### LINKED ISSUE #10506: sequelize.query typings are not allowing model property in typings (v5)
## What are you doing?

```js
sequelize.query(/* my query here */, {
  type: QueryTypes.SELECT,
  model: ResolvedTag,
  transaction: tx
})
```

Should permit specifying the model that should be returned (populated) for rows selected.

`error TS2345: Argument of type '{ type: QueryTypes; model: typeof ResolvedTag; transaction: Transaction; }' is not assignable to parameter of type 'QueryOptions \| QueryOptionsWithType \| u ndefined'.`

__Dialect:__ mysql
__Dialect version:__ 1.6.5
__Database version:__ `mariadb:10.2` - `907f5f6c749d`
__Sequelize version:__ 5.0.0-beta.16

SimonSchick:
### Pull Request check-list

### Description of change

See title
fixes #10506

@sushantdhiman can you clarify if you can specify BOTH the query `type` AND `model`?
It seems that when model is specified it implicitly uses `SELECT` but what if `type` is specified?

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -1688,7 +1688,7 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
    *   where: ...,
    *   limit: 12,
    *   offset: 12
-   * }).then(function (result) {
+   * }).then(result => {
    *   ...
    * })
    * ```
@@ -1776,7 +1776,7 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
 
   /**
    * Find a row that matches the query, or build (but don't save) the row if none is found.
-   * The successfull result of the promise will be (instance, initialized) - Make sure to use .spread()
+   * The successfull result of the promise will be (instance, initialized) - Make sure to use `.then(([...]))`
    */
   public static findOrBuild<M extends Model>(
     this: { new (): M } & typeof Model,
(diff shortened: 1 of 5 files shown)


sushantdhiman:
https://github.com/sequelize/sequelize/blob/45648dd034450f754500bfdf27b2f1d7b7a22d98/lib/sequelize.js#L482-L488

Both `type` and `model` can be specified, when there is no `type` we use `options.model` to set it to `SELECT` type.

sushantdhiman:
Let me know if this is ready

SimonSchick:
@sushantdhiman what exactly is the return type if, for example `model` and `type: INSERT` is specified?

sushantdhiman:
`type: INSERT`, as `model` won't modify `type` if it's already set

SimonSchick:
So does `setting` model have any effect for any type except `SELECT`? Sorry the code is really hard to understand..

sushantdhiman:
Setting model has no effect on type, if type is already set

SimonSchick:
So, in typings we should probably not support setting both at the same time then as it's pointless and probably an error.

In this case this PR is complete.

PR #10522, opened 2019-03-08
feat(sqlite/query-generator): support restart identity in SQLite

SavageCore:
### Pull Request check-list

### Description of change

Follow-up to #5356 adding support for SQLite.

`npm run test-sqlite` passes and I also tested the generated SQL against my DB

I believe current docs suffice?

### DIFF
--- a/lib/dialects/sqlite/query-generator.js
+++ b/lib/dialects/sqlite/query-generator.js
@@ -260,7 +260,7 @@ class SQLiteQueryGenerator extends MySqlQueryGenerator {
   truncateTableQuery(tableName, options = {}) {
     return [
       `DELETE FROM ${this.quoteTable(tableName)}`,
-      options.restartIdentity ? `; DELETE FROM \`sqlite_sequence\` WHERE \`name\` = '${tableName.schema}.${tableName.table}';` : ''
+      options.restartIdentity ? `; DELETE FROM ${this.quoteTable('sqlite_sequence')} WHERE ${this.quoteIdentifier('name')} = '${tableName.schema}.${tableName.table}';` : ''
     ].join('');
   }
 
--- a/lib/dialects/sqlite/query-generator.js
+++ b/lib/dialects/sqlite/query-generator.js
@@ -257,8 +257,11 @@ class SQLiteQueryGenerator extends MySqlQueryGenerator {
     return { query, bind };
   }
 
-  truncateTableQuery(tableName) {
-    return `DELETE FROM ${this.quoteTable(tableName)}`;
+  truncateTableQuery(tableName, options = {}) {
+    return [
+      `DELETE FROM ${this.quoteTable(tableName)}`,
+      options.restartIdentity ? `; DELETE FROM \`sqlite_sequence\` WHERE \`name\` = '${tableName.schema}.${tableName.table}';` : ''
+    ].join('');
   }
 
   deleteQuery(tableName, where, options = {}, model) {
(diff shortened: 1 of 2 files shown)


lib/dialects/sqlite/query-generator.js
```diff
@@ -257,8 +257,11 @@ class SQLiteQueryGenerator extends MySqlQueryGenerator {
     return { query, bind };
   }
 
-  truncateTableQuery(tableName) {
-    return `DELETE FROM ${this.quoteTable(tableName)}`;
+  truncateTableQuery(tableName, options = {}) {
+    return [
+      `DELETE FROM ${this.quoteTable(tableName)}`,
+      options.restartIdentity ? `; DELETE FROM \`sqlite_sequence\` WHERE \`name\` = '${tableName.schema}.${tableName.table}';` : ''
```
sushantdhiman: `this.quoteTable` for `sqlite_sequence`, `this.quoteIdentifier` for `name`

sushantdhiman: `'${tableName.schema}.${tableName.table}'` should be replaced with `this.quoteTable(tableName)`

SavageCore: I tried with quoteTable and it used backticks which wasn't valid, must be single inverted comma?

sushantdhiman: Isn't delete from statement using same `quoteTable`, that works.

SavageCore: From my testing it seemed that 

```
DELETE FROM `table`; DELETE FROM `sqlite_sequence` WHERE `name` = 'table'
```

Works but

```
DELETE FROM `table`; DELETE FROM `sqlite_sequence` WHERE `name` = `table`
```

Returns 

```
-- At line 1:
DELETE FROM `sqlite_sequence` WHERE `name` = `table`
-- Result: no such column: table
```

sushantdhiman: Yeah that wont work as we are comparing against string `name`, you can try `'this.escape(tableName.table)'`. It is possible `tableName` is already a string, so you need to handle that as well

Do you think we need to prefix `schema` with table name? How is `sqlite_sequence` is storing table names, prefixed with schema?

SavageCore: > 
> 
> Yeah that wont work as we are comparing against string `name`, you can try `'this.escape(tableName.table)'`. It is possible `tableName` is already a string, so you need to handle that as well
> 
> Do you think we need to prefix `schema` with table name? How is `sqlite_sequence` is storing table names, prefixed with schema?

Yes, `sqlite_sequence` name is `public.test_users` in the test examples.

lib/dialects/sqlite/query-generator.js
```diff
@@ -257,8 +257,11 @@ class SQLiteQueryGenerator extends MySqlQueryGenerator {
     return { query, bind };
   }
 
-  truncateTableQuery(tableName) {
-    return `DELETE FROM ${this.quoteTable(tableName)}`;
+  truncateTableQuery(tableName, options = {}) {
+    return [
+      `DELETE FROM ${this.quoteTable(tableName)}`,
+      options.restartIdentity ? `; DELETE FROM ${this.quoteTable('sqlite_sequence')} WHERE ${this.quoteIdentifier('name')} = '${tableName.schema}.${tableName.table}';` : ''
```
sushantdhiman:
```suggestion
      options.restartIdentity ? `; DELETE FROM ${this.quoteTable('sqlite_sequence')} WHERE ${this.quoteIdentifier('name')} = '${this.escape(tableName.schema)}.${this.escape(tableName.table)}';` : ''
```

SavageCore: Oh hang on, in both my original and this suggestion we don't handle cases without schema. We need the logic from quoteTable.

sushantdhiman: You can use ```Utils.addTicks(Utils.removeTicks(this.quoteTable(table), '`'), "'");```

PR #10524, opened 2019-03-09
feat(types): add generics types for associations

Kilimangaro:
### Description of change

Added generics types (default to "typeof Model) to all associations class definitions:

- HasOne
- HasMany
- BelongsTo
- BelongsToMany

These associations can now be defined as follow

```typescript
// ...
public static associations: {
  group: BelongsTo<User, UserGroup>;
};
// ...

/**
 * userType is now a "typeof User" instead of "typeof Model" 
 */
const userType: typeof User = User.associations.group.source;
```
Also, associations methods now returns the typed associations

```typescript
/**
 * Group is now a "BelongsTo<User, UserGroup>" instead of "BelongsTo" 
 */
export const Group = User.belongsTo(UserGroup, { as: 'group', foreignKey: 'groupId' });
```


sushantdhiman:
LGTM, @SimonSchick please review

types/lib/associations/base.d.ts
```diff
@@ -1,9 +1,9 @@
 import { ColumnOptions, Model } from '../model';
 
-export abstract class Association {
+export abstract class Association<S = typeof Model, T = typeof Model> {
```
SimonSchick: Can you please add an export to `model.d.ts` that's called `ModelType = typeof Model` and use that instead?

types/lib/model.d.ts
```diff
@@ -2235,7 +2235,10 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
    * @param target The model that will be associated with hasOne relationship
    * @param options Options for the association
    */
-  public static hasOne(target: typeof Model, options?: HasOneOptions): HasOne;
+  public static hasOne<M, T extends typeof Model>(
+    this: M,
```
SimonSchick: Should probably be `{ new (): M } & typeof Model` but feel free to create a generic such as `ModelCtor<M> = { new (): M } & typeof Mode` to clean this up a bit.

types/lib/model.d.ts
```diff
@@ -2300,7 +2306,10 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
    * @param target The model that will be associated with hasOne relationship
    * @param options Options for the association
    */
-  public static hasMany(target: typeof Model, options?: HasManyOptions): HasMany;
+  public static hasMany<M, T extends typeof Model>(
+    this: M, 
+    target: T, options?: HasManyOptions
```
SimonSchick: Nit: Please be consistent about newlines

types/lib/associations/base.d.ts
Kilimangaro: done

types/lib/model.d.ts
Kilimangaro: Added  export type "ModelCtor"

types/lib/model.d.ts
Kilimangaro: done.

It could be nice to have some prettier config

types/test/models/User.ts
```diff
@@ -68,7 +71,11 @@ User.afterFind((users, options) => {
 User.addHook('beforeFind', 'test', (options: FindOptions) => {
   return undefined;
 });
+
 // associate
 // it is important to import _after_ the model above is already exported so the circular reference works.
 import { UserGroup } from './UserGroup';
 export const Group = User.belongsTo(UserGroup, { as: 'group', foreignKey: 'groupId' });
+
+// const userType: ModelCtor<User> = User.associations.group.source;
```
sushantdhiman: Why are these lines commented?

Kilimangaro: my mistake... fixed that in last commit

Issue #8868, opened 2018-01-10
On renameColumn throw error if defaultValue CURRENT_TIMESTAMP


usebaz:
## What are you doing?
I run migration with code:

```js
await queryInterface.renameColumn(tableName, 'timestamp', 'updated_timestamp');
```

## What do you expect to happen?
Migration successfully applied

## What is actually happening?
Throw error:
```
{
  Error: Invalid default value for 'updated_timestamp'
  code: 'ER_INVALID_DEFAULT',
  errno: 1067,
  sqlState: '42000',
  sqlMessage: 'Invalid default value for \'updated_timestamp\'',
  sql: 'ALTER TABLE `billing_apple_callback` CHANGE `timestamp` `updated_timestamp` DATETIME NOT NULL DEFAULT \'CURRENT_TIMESTAMP\';'
}
```

SQL: ``ALTER TABLE `billing_apple_callback` CHANGE `timestamp` `updated_timestamp` DATETIME NOT NULL DEFAULT \'CURRENT_TIMESTAMP\';``

__Dialect:__ mysql
__Dialect version:__ musql2@1.5.1
__Database version:__ 5.7.13
__Sequelize version:__ 4.31.0
__Tested with latest release:__ No

If change sql to ``ALTER TABLE `billing_apple_callback` CHANGE `timestamp` `updated_timestamp` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP;`` all be fine. Error in `\'`around `CURRENT_TIMESTAMP`.

sushantdhiman:
I think you have set `defaultValue` for this attribute to `CURRENT_TIMESTAMP`, remove that and let Sequelize handle it

usebaz:
@sushantdhiman where I set `defaultValue`?

sushantdhiman:
For `timestamp`, how its defined?

usebaz:
In DB it have `defaultValue: CURRENT_TIMESTAMP` but in migration I not set this value.

sushantdhiman:
Shouldn't matter, its an issue with improper escaping, right? Tagging as bug

usebaz:
Yes it about escapin.
Sequelize must generate sql:
``ALTER TABLE `billing_apple_callback` CHANGE `timestamp` `updated_timestamp` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP;``

[...]

PR #10525, opened 2019-03-09
fix(mysql): issue with MySQL, TIMESTAMPS, and CURRENT_TIMESTAMP

garrinmf:
### Pull Request check-list

- [ x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ x] Have you added new tests to prevent regressions?
- [ x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?

### Description of change

Do not quote CURRENT_TIMESTAMP when it's being used with MySQL.  I think it's safe to never do this.  If I knew where to scope this to 'defaultValue' I'd feel safer?

See #8868, attempting to fix this issue.

### DIFF
--- a/lib/sql-string.js
+++ b/lib/sql-string.js
@@ -66,6 +66,10 @@ function escape(val, timeZone, dialect, format) {
     throw new Error(`Invalid value ${util.inspect(val)}`);
   }
 
+  if (dialect === 'mysql' && val === 'CURRENT_TIMESTAMP') {
+    return val;
+  }
+
   if (dialect === 'postgres' || dialect === 'sqlite' || dialect === 'mssql') {
     // http://www.postgresql.org/docs/8.2/static/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS
     // http://stackoverflow.com/q/603572/130598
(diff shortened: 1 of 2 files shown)


lib/sql-string.js
```diff
@@ -66,6 +66,10 @@ function escape(val, timeZone, dialect, format) {
     throw new Error(`Invalid value ${util.inspect(val)}`);
   }
 
+  if (dialect === 'mysql' && val === 'CURRENT_TIMESTAMP') {
```
SimonSchick: Hardcoding specific values seems kind of hacky, we should probably handle this at a higher level or at least have a specific list.
I'm not an expect in this regard but aren't there other 'magic' default values too?

SimonSchick: See https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html

SimonSchick:
Also have you tried using `literal` instead?

SimonSchick:
Tested locally, you should be using `literal('CURRENT_TIMESTAMP')` here.

garrinmf:
@SimonSchick, I realize that the ticket I linked to is long and varying but below is the specific case I'm testing against in my development:

* **Existing Column**: in users defined as `created_when      TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP`
* **migration**: `queryInterface.renameColumn('users', 'created_when', 'created_at')` 
* **output logs**: `Executing (default): ALTER TABLE `users` CHANGE `created_when` `created_at` TIMESTAMP NOT NULL DEFAULT 'CURRENT_TIMESTAMP';`
* **error**: `ERROR: Invalid default value for 'created_at'`

I've used `literal` successfully in other cases but in this case something gets lost in the translation of existing column metadata?

readysetawesome:
I'm having the same problem as @garrinmf and it is a huge pain in the ass for such a totally reasonable use case.

PR #10532, opened 2019-03-10
chore: enforce stricter linting

SimonSchick:
### Pull Request check-list

### Description of change

- See enabled lint rules for rough overview.
  - `no-invalid-this` required some refactors but made the code overall much more understable in many places.
- Deduped `Query` class constructors
- Removed `_.partial` in favour of arrow functions (partial was hella confusing)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -102,11 +102,22 @@
     "one-var-declaration-per-line": "error",
     "comma-dangle": "error",
     "no-shadow": "warn",
-    "camelcase": "warn",
+    "camelcase": "error",
     "prefer-template": "error",
     "no-else-return": ["error", { "allowElseIf": false }],
     "no-lonely-if": "error",
-    "no-this-before-super": "error"
+    "no-this-before-super": "error",
+    "no-throw-literal": "error",
+    "prefer-promise-reject-errors": "error",
+    "no-invalid-this": "error",
+    "radix": "error",
+    "no-with": "error",
+    "no-useless-concat": "error",
+    "no-useless-catch": "error",
+    "no-useless-call": "error",
+    "no-unused-expressions": "error",
+    "no-sequences": "error",
+    "no-self-compare": "error"
   },
   "parserOptions": {
     "ecmaVersion": 6,
@@ -133,5 +144,5 @@
   //   "Model": true,
   //   "QueryInterface": true,
   //   "QueryType": true
-  // }
+  // },
 }
--- a/lib/dialects/sqlite/query-interface.js
+++ b/lib/dialects/sqlite/query-interface.js
@@ -18,6 +18,7 @@ const QueryTypes = require('../../query-types');
   It will create a backup of the table, drop the table afterwards and create a
   new table with the same name but without the obsolete column.
 
+  @param  {QueryInterface} qi
   @param  {string} tableName     The name of the table.
   @param  {string} attributeName The name of the attribute that we want to remove.
   @param  {Object} options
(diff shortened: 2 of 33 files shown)


.eslintrc.json
```diff
@@ -121,17 +132,4 @@
     "mocha": true,
     "es6": true
   }
-  // "globals": {
```
sushantdhiman: keep these, they are used for jsdoc linting

SimonSchick: Ah, I was wondering what these were for, but why are they disabled?

sushantdhiman: jsdoc linting is not enable for now

SimonSchick:
Also I noticed a few tests are failing where I removed `parseInt`, this isn't unexpected.
I can add those back in those particular cases I expected sequelize to actually return a number, not a string, maybe these should be fixed?

lib/data-types.js
```diff
@@ -358,7 +358,7 @@ class DECIMAL extends NUMBER {
   }
   toSql() {
     if (this._precision || this._scale) {
-      return `DECIMAL(${[this._precision, this._scale].filter(_.identity).join(',')})`;
```
sushantdhiman: `identity` is clear

SimonSchick: Actually, what if `_scale` is specified but `_precision` isn't, wouldn't that generate a bad result?

.eslintrc.json
SimonSchick: Sounds like something we should fix :>

lib/data-types.js
sushantdhiman: Yes, it will. I think many database allow just precision, so I think that is what being handled here. In worst case you can specify just the `scale`.

But I think those errors should be caught in development phase. Proper validation for this can be implemented when construing DECIMAL though

SimonSchick:
@sushantdhiman so there's a total of 5 failing tests across different DBs atm, all related to the fact that in certain cases sequelize seems to return strings where the result should be a number, all are related to aggregation.
Do you want to look into this error before we merge this or should I revert the changes in those places and open an issue?

sushantdhiman:
Looks like Postgres, MySQL and SQLite are returning string for `count`, odd. We can do `parseInt` simply but I would like to know why string is being returned by those dialects. Revert those tests for now

lib/dialects/sqlite/query.js
```diff
@@ -234,6 +218,7 @@ class Query extends AbstractQuery {
   }
 
   run(sql, parameters) {
+    const conn = this.connection;
```
sushantdhiman: call this `connection`

.eslintrc.json
```diff
@@ -102,11 +102,35 @@
...
+    "radix": "error",
+    "no-with": "error",
+    "no-useless-concat": "error",
+    "no-useless-catch": "error",
+    "no-useless-call": "error",
+    "no-unused-expressions": "error",
+    "no-sequences": "error",
+    "no-self-compare": "error"
+  },
+  "globals": {
```
sushantdhiman: Are you fixing jsdocs as well, if not just remove them and plugin dependency for now

SimonSchick: Oh, I was going to create another PR right after this one if that's ok 😛

SimonSchick:
All fixed and squashed.

lib/dialects/sqlite/query-interface.js
```diff
@@ -130,50 +136,53 @@ function removeConstraint(tableName, constraintName, options) {
...
     });
 }
 exports.removeConstraint = removeConstraint;
 
-function addConstraint(tableName, options) {
-  const constraintSnippet = this.QueryGenerator.getConstraintSnippet(tableName, options);
-  const describeCreateTableSql = this.QueryGenerator.describeCreateTableQuery(tableName);
+/**
+ * @param {QueryInterface} qi
+ */
```
sushantdhiman:
```suggestion
 * @private
 */
```

sushantdhiman: For other methods as well, or just remove the new jsdoc headers from these dialect based query interface methods

SimonSchick:
@sushantdhiman Tbh, the current state of jsdoc inside of Sequelize is pretty bad, type integrity is all over the place, which makes sense given the amount of commits, complexity and such.

I'm not sure if pursuing JSDoc makes sense anymore as the time you'd need to invest to fix all of the issues and create new jsdoclets would probably exceed the time required to just move everything to TypeScript.

I prepared another PR that I will open after this one was merged which will at least re-enables linting and fixes most existing JSDoc issues.

PR #10535, opened 2019-03-10
chore: even more linting

SimonSchick:
### Pull Request check-list

### Description of change

- Enabled eslint default rules
  - Most of the changes here are related to redundant escape sequences
- Fix JSDoc stuff

I think we're legit done after this one 😛


.eslintrc.json
```diff
@@ -117,7 +109,23 @@
     "no-useless-call": "error",
     "no-unused-expressions": "error",
     "no-sequences": "error",
-    "no-self-compare": "error"
+    "no-self-compare": "error",
+    "no-case-declarations": "off"
+  },
+  "globals": {
+    "Association": true,
```
sushantdhiman: This is going to be an issue with linting, now eslint will treat these as global and wont fail if someone is using `Model` without importing it first.

We can have a second linting phase for just jsdocs, with its own jsdocs related config in separate file.

SimonSchick: Or I just change those to typedefs.

sushantdhiman: So isn't this thread still valid. `npm run lint` wont catch misuse of these global?

SimonSchick: I forgot to push latest changes, unfortunately type references are broken: https://github.com/gajus/eslint-plugin-jsdoc/issues/109

SimonSchick:
@sushantdhiman apparently `test/integration/dialects/postgres/query-interface.test.js` is failing now since I've fixed the postgres version regex, should I change the regex to only match X.Y instead of X.Y.Z?

lib/dialects/sqlite/query-interface.js
```diff
@@ -147,7 +151,8 @@ exports.removeConstraint = removeConstraint;
 
 /**
  * @param {QueryInterface} qi
- * @private
```
sushantdhiman: Keep `@private`

sushantdhiman:
> should I change the regex to only match X.Y instead of X.Y.Z?

Yes

Issue #10511, opened 2019-03-06
'subQuery' option for find is not documented


serge1peshcoff:
I was recently trying to find the solution for this issue: https://github.com/sequelize/sequelize/issues/9869, where the following example was throwing the `missing FROM-clause` error:

```js
Event.findAll({
    where: {
        deleted: false,
        '$applications.user_id$': req.user.id
    },
    include: [{
        model: Application,
        required: true
    }]
})
```

I've found one question on StackOverflow that suggested to use `subQuery: false`, so I've updated my example and now it's working, here's the working solution:

```js
Event.findAll({
    where: {
        deleted: false,
        '$applications.user_id$': req.user.id
    },
    subQuery: false,  // <----------- this line
    include: [{
        model: Application,
        required: true
    }]
})
```

But as I've found out the `subQuery` option for `findAll()` (and I suppose that for `findOne()` it's the same) is not documented or at least I couldn't find it.

Is it possible to document it so other people would be aware of it?
Thanks in advance ;)

knoxcard:
@serge1peshcoff - great recommendation, you should submit a pull request.

serge1peshcoff:
done :D

Issue #9869, opened 2018-09-04
Deep eager load chains does not produce valid sql, regression from sequelize 3


linpekka:
## What are you doing?
When upgrading from sequelize 3, i noticed that some lookups does no longer produce valid sql.

```js
return db.Task.findOne({
	where: {
		behavior_id: Integer
	},
	include:[
		{model: db.Behavior, include: [db.Node]},
		{
			model: db.TaskSession,
			where: {
				session_id:Integer
			},
			include: [{
				model: db.TaskReport,
				required: true,
				include:[
					{model: db.Attachment}, 
					{
						model: db.TaskParticipation,
						required: true,
						include:[{
							model:db.Participation, 
							required: true,
							include:[
								{
									model:db.OrganisationParticipation,
									where: {user_id:Integer},
									include:[
										db.User
									]
								}	
							]
						}]
					}
				]
			}]
		}
	],
	transaction:t
})
```

## What do you expect to happen?
I want the findOne to generate valid sql

## What is actually happening?
SequelizeDatabaseError: missing FROM-clause entry for table "TaskSessions->TaskReports->TaskParticipations->Participation"

_Output, either JSON or SQL_
[...]

PR #10543, opened 2019-03-13
docs: added subQuery documentation for findAll

serge1peshcoff:
Fixes: https://github.com/sequelize/sequelize/issues/10511
Related: https://github.com/sequelize/sequelize/issues/9869

I'm not sure if there can be a better description, since I'm not quite aware of how subQuery works internally. But this should make it more clear. If there's a better way to document it, I can update it.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1647,6 +1647,7 @@ class Model {
    * @param  {Object}                                                    [options.having] Having options
    * @param  {string}                                                    [options.searchPath=DEFAULT] An optional parameter to specify the schema search_path (Postgres only)
    * @param  {boolean|Error}                                             [options.rejectOnEmpty=false] Throws an error when no records found
+   * @param  {boolean}                                                   [options.subQuery] If true, uses sub-query for querying associates.
    *
    * @see {@link Sequelize#query}
    *


lib/model.js
```diff
@@ -1647,6 +1647,7 @@ class Model {
    * @param  {Object}                                                    [options.having] Having options
    * @param  {string}                                                    [options.searchPath=DEFAULT] An optional parameter to specify the schema search_path (Postgres only)
    * @param  {boolean|Error}                                             [options.rejectOnEmpty=false] Throws an error when no records found
+   * @param  {boolean}                                                   [options.subQuery] If true, uses sub-query for querying associates.
```
sushantdhiman: This option should not be used, instead you should be using `duplicating: false` etc (Which are not documented either :D)

serge1peshcoff: @sushantdhiman then I propose the following: document both of these features, but mention that subQuery should not be used.
Can you do that? (I'm not that familiar with internals of Sequelize unfortunately and don't know how either of this options works and why one should use duplicating instead of subQuery)

SimonSchick: Please also add it to the typings.

PR #10551, opened 2019-03-14
Migration documentation

rimiti:
Hello,

- Small mistakes fixed into migration page. 🚀 

Regards,

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -17,7 +17,7 @@ $ npm install --save sequelize-cli
 To create an empty project you will need to execute `init` command
 
 ```bash
-$ npx sequelize init
+$ npx sequelize-cli init
 ```
 
 This will create following folders
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -71,7 +71,7 @@ We will use `model:generate` command. This command requires two options
 Let's create a model named `User`.
 
 ```bash
-$ npx sequelize model:generate --name User --attributes firstName:string,lastName:string,email:string
+$ npx sequelize-cli model:generate --name User --attributes firstName:string,lastName:string,email:string
 ```
 
 This will do following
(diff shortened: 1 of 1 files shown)


rimiti:
Pipeline broken after a documentation update? 🧐

papb:
Hello @rimiti, don't worry about the pipeline - as you said, it doesn't make sense for it to break after a documentation change. This is a known problem, some tests have intermittent behavior. The maintainers are well aware of it :)

About the change you're making in this PR, are you sure you've got it right? I haven't used the CLI for a long time, but apparently the binary exported is still just `sequelize` and not `sequelize-cli`: [see here](https://github.com/sequelize/cli/blob/v5.4.0/package.json#L6).

Have you tested it? Are you sure `npx sequelize-cli` is correct? :)

rimiti:
> Hello @rimiti, don't worry about the pipeline - as you said, it doesn't make sense for it to break after a documentation change. This is a known problem, some tests have intermittent behavior. The maintainers are well aware of it :)

Thank you for the details. If you need some help to maintain the repo, I can help you.

> About the change you're making in this PR, are you sure you've got it right? I haven't used the CLI for a long time, but apparently the binary exported is still just `sequelize` and not `sequelize-cli`: [see here](https://github.com/sequelize/cli/blob/v5.4.0/package.json#L6).
> 
> Have you tested it? Are you sure `npx sequelize-cli` is correct? :)

Of course ;) 
For your information **npx** uses the package name and **npm** uses bin key from package.json file: https://github.com/sequelize/cli/blob/0828c1f5f9234f226d332a61c8fe34ea2e9e8f37/package.json#L6

papb:
:open_mouth: I didn't know that. But... This is strange...

I have just tested it and `npx sequelize --version` works fine. Also, `npx sequelize-cli --version` behavior is very odd. Instead of just working immediately, the npm spinner shows up, "fetch metadata", "extract", etc, and then `npx: installed 106 in 11.55s`, and then it also works. But looks like it re-downloads it every time. At least from my local tests, looks like `npx sequelize` works and is even better than `npx sequelize-cli`.

Perhaps this is a OS issue? I'm on windows.

sushantdhiman:
Yes `sequelize-cli` exposes binary as `sequelize` so docs are already correct

rimiti:
It's really weird... @sunshinewyin is right.
I've tried again (after a nvm update), and it works... 🧐

PR #10553, opened 2019-03-14
WIP: chore: improve performance for iteration

SimonSchick:
### Pull Request check-list

### Description of change

1)
Replaced `_.forEach`, `_.each` and `_.forOwn` with a faster custom function.
There are some small performance gains in this but this is mostly done to further reduce dependencies on lodash.

2) Minor optimisations on repetitive calls of `Object.keys()`

3) Minor optimization for `unrecognizedOptions` 

4) Minor optimization for `Utils.getOperators`


frlinw:
@SimonSchick according to my lastest tests https://jsperf.com/object-keys-vs-hasownproperty/64
double loop
```js
Object.keys(obj).forEach(...)
```
is slower than a simple loop + condition
```js
for (const key in obj) {
   if (obj.hasOwnProperty(key) {
      ....
   }
}
```

Additional note: forEach usage is not consistent: most of the time `forEAch`, sometimes `Utils.forEach`

SimonSchick:
I did my own benchmarks as well and the Object.keys variant was always faster.

```
/* eslint-disable no-console */
/* eslint-disable no-loop-func */
'use strict';

const { forEach, forOwn, forIn } = require('lodash');

function oKeys(obj, fn) {
  Object.keys(obj).forEach(k => {
    fn(obj[k], k);
  });
}

function inKeys(obj, fn) {
  for (const k in obj) {
    if (obj.hasOwnProperty(k)) {
      fn(obj[k], k);
    }
  }
}

const obj = {};
for (let i = 0;i<0xf;++i) {
  obj[i] = i;
}
forEach(obj, console.log);
forOwn(obj, console.log);
forIn(obj, console.log);
oKeys(obj, console.log);
inKeys(obj, console.log);

console.time('forIn');
for (let i = 0;i<0xFFFFF;++i) {
  forIn(obj, g => {
    k = g;
  });
}
console.timeEnd('forIn');

console.time('forEach');
for (let i = 0;i<0xFFFFF;++i) {
  forEach(obj, g => {
    k = g;
  });
}
console.timeEnd('forEach');

console.time('forOwn');
for (let i = 0;i<0xFFFFF;++i) {
  forOwn(obj, g => {
    k = g;
  });
}
console.timeEnd('forOwn');

console.time('Object.keys');
for (let i = 0;i<0xFFFFF;++i) {
  oKeys(obj, g => {
    k = g;
  });
}
console.timeEnd('Object.keys');

console.time('Object.entries');
for (let i = 0;i<0xFFFFF;++i) {
  for (const [, g] of Object.entries(obj)) {
    k = g;
  }
}
console.timeEnd('Object.entries');

console.time('for in');
for (let i = 0;i<0xFFFFF;++i) {
  inKeys(obj, g => {
    k = g;
  });
}
console.timeEnd('for in');
```

frlinw:
@SimonSchick You are right. But I suspect the browser to cache the result of `entries` if there are multiple calls on the same reference. To make it a bit more real, try it with an array of different objects.

sushantdhiman:
I don't think small performance gains are worth it for the change purposed.

Firstly lodash methods are optimized for various use-case and stable performance. There are always very quick bug fixes in lodash for any issues with given method, their community base is very large. If there are some issues with our implementation they may never get caught.

Secondly I don't like our dependency upon `Utils.*` or `lodash` either but lodash is the lesser evil here with good benefits.

lib/model.js
```diff
@@ -19,6 +19,14 @@ const Hooks = require('./hooks');
 const associationsMixin = require('./associations/mixin');
 const Op = require('./operators');
 
+// This list will quickly become dated, but failing to maintain this list just means
+// we won't throw a warning when we should. At least most common cases will forever be covered
+// so we stop throwing erroneous warnings when we shouldn't.
+const validQueryKeywords = new Set(['where', 'attributes', 'paranoid', 'include', 'order', 'limit', 'offset',
```
sushantdhiman: This optimization is acceptable, we should not be creating this every time we check for invalid options

SimonSchick:
@sushantdhiman the main motivation here is to reduce lodash usage.
As we (probably) move on to v6 ~80% of lodash's functionality will be available natively and lodash is quite a heavy library.
My intermediate goal is to eliminate lodash as much as possible and then only import lodash submodules for remaining operations.
Also: object iteration isn't very hard or complicated and is extremely unlikely to break, I did benchmark this code in various scenarios and lodash was always a tad slower.

HOWEVER: I just read lodash's V5 roadmap https://github.com/lodash/lodash/wiki/Roadmap#v500-2019, I'm going to cherry-pick the optimizations I did in this PR and PR them separately, there might be hope yet!

PR #10569, opened 2019-03-17
fix(query-generator): convert numbers to date types

### LINKED ISSUE #10563: Op.between fails to convert arguments to Date strings
## What are you doing?

Querying a DataType.DATE column using Op.between.

```js
const Op = sequelize.Sequelize.Op;
const Event = sequelize.define('Event', {
  rowid: {type: DataTypes.INTEGER, primaryKey: true},
  time: {type: DataTypes.DATE, allowNull: false},
}, {
  tableName: 'events',
  name: {singular: 'event', plural: 'events'},
  timestamps: false,
});

// LTE + GTE query
Event.findAll({where: {time: {[Op.and]: [{[Op.gte]: 0}, {[Op.lte]: 10}]}}});

// BETWEEN query
Event.findAll({where: {time: {[Op.between]: [0, 10]}}});
```

As BETWEEN is simply shorthand for `val >= min AND val <= max`, I would expect conversion to column types to be handled similarly.  That is, the two queries above should generate functionally equivalent SQL like this:
```
SELECT `rowid`, `time` FROM `events` AS `Foo` WHERE (`Foo`.`time` >= '1970-01-01 00:00:00.000 +00:00' AND `Foo`.`time` <= '1970-01-01 00:00:00.010 +00:00');
SELECT `rowid`, `time` FROM `events` AS `Foo` WHERE `Foo`.`time` BETWEEN '1970-01-01 00:00:00.000 +00:00' AND '1970-01-01 00:00:00.010 +00:00';
```

The BETWEEN query fails to convert the min/max values to Date strings, instead injecting them as raw values:
```
SELECT `rowid`, `time` FROM `events` AS `Foo` WHERE (`Foo`.`time` >= '1970-01-01 00:00:00.000 +00:00' AND `Foo`.`time` <= '1970-01-01 00:00:00.010 +00:00');
SELECT `rowid`, `time` FROM `events` AS `Foo` WHERE `Foo`.`time` BETWEEN 0 AND 10;
```

__Dialect:__ sqlite
__Dialect version:__ XXX
__Database version:__ 3.19.3 2017-06-27 16:48:08 2b0954060fe10d6de6d479287dd88890f1bef6cc1beca11bc6cdb79f72e2377b
__Sequelize version:__ sequelize@5.1.0
__Tested with latest release:__ Yes (5.1.0)

SimonSchick:
### Pull Request check-list

### Description of change

Fixes #10563

Also renamed some tests as using mocha's `-f GREB` doesn't work well with `$`.


SimonSchick:
Seems like it doesn't like the timezone being included for some reason, it works fine locally..

SimonSchick:
@sushantdhiman I'll be pretty busy with work for a while, can you look into this?

package.json
```diff
@@ -121,15 +121,15 @@
     "test-docker-integration": "env-cmd $npm_package_options_env_cmd npm run test-integration",
     "docs": "esdoc && cp docs/ROUTER esdoc/ROUTER && cp docs/favicon.ico esdoc/favicon.ico",
     "teaser": "node scripts/teaser",
-    "test-unit": "mocha --require scripts/mocha-bootload --globals setImmediate,clearImmediate --exit --check-leaks --colors -t 30000 --reporter spec \"test/unit/**/*.js\"",
```
eseliger: is this related to the issue? or is it just not needed anymore?

SimonSchick: Oh, I didn't really mean to commit this but no, this isn't needed anymore.

eseliger: maybe we can put this in a separate pr then so if it would cause trouble at some end it's easier to pin it down using bisect

Issue #10123, opened 2018-11-05
[5.0.0-beta.14][Regression] mixed where clauses in include


frlinw:
## What are you doing?
Nested where conditions are not related to their association.

```js
const Sequelize = require('./')

const sequelize = new Sequelize({
  dialect: 'sqlite',
  file: ':memory:'
})

const Clientfile = sequelize.define('clientfile', {
  id: {
    type: Sequelize.UUID,
    defaultValue: Sequelize.UUIDV4,
    primaryKey: true
  }
})

const Mission = sequelize.define('mission', {
  id: {
    type: Sequelize.UUID,
    defaultValue: Sequelize.UUIDV4,
    primaryKey: true
  },
  startDate: {
    type: Sequelize.DATE
  }
})

const Building = sequelize.define('building', {
  id: {
    type: Sequelize.UUID,
    defaultValue: Sequelize.UUIDV4,
    primaryKey: true
  }
})

Clientfile.hasOne(Mission)
Clientfile.hasOne(Building)

const sync = async () => {
  await Clientfile.sync()
  await Mission.sync()
  await Building.sync()

  await Clientfile.findAll({
    attributes: [],
    include: [
      {
        association: 'mission',
        attributes: ['id'],
        where: {
          startDate: {
            [Sequelize.Op.gte]: new Date(),
            [Sequelize.Op.lte]: new Date()
          }
        }
      },
      {
        association: 'building',
        attributes: ['id']
      }
    ]
  })
}

sync()
```

## What do you expect to happen?
The nested where condition is on `mission` association and should affect this association only

## What is actually happening?
`column building.startDate does not exist`
The where condition of `mission` association seems to be applied to the `building` association.

Dialect: postgres
Sequelize version: 5.0.0-beta.14
Tested with latest release: Yes, 5.0.0-beta.15

sushantdhiman:
Please follow the issue template

frlinw:
Still an issue in 5.0.0-beta.15

frlinw:
The example should be executable

sushantdhiman:
If you execute your own (edited) code you will see error very clearly :)

```js
Error: Unrecognized datatype for attribute "mission.startDate"
    at rawAttributes._.mapValues (/var/www/github/sequelize/lib/model.js:976:15)
    at /var/www/github/sequelize/node_modules/lodash/lodash.js:13402:38
    at /var/www/github/sequelize/node_modules/lodash/lodash.js:4911:15
[...]

PR #10580, opened 2019-03-20
fix(uniq-includes-with-association): modify uniqIncludes

u9r52sld:
### Pull Request check-list

### Description of change

fixing `_uniqIncludes` function with `include.association`.

See issue: #10123

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -782,7 +782,10 @@ class Model {
     if (!options.include) return;
 
     options.include = _(options.include)
-      .groupBy(include => `${include.model && include.model.name}-${include.as}`)
+      .groupBy(include => {
+        const modelName =  include.model ? include.model.name : include.association;
+        return `${modelName}-${include.as}`;
+      })
       .map(includes => this._assignOptions(...includes))
       .value();
   }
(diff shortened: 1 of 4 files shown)


test/integration/model/scope/associations.test.js
```diff
@@ -418,6 +418,31 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+              return Clientfile.findAll({
+                include: [
+                  {
+                    association: 'mission',
+                    where: {
+                      secret: 'foo'
+                    }
+                  },
+                  {
+                    association: 'building'
```
sushantdhiman: What if you are passing association object, then `uniqueInclude` can fail due to `[object]` stringification

lib/model.js
```diff
@@ -782,7 +782,10 @@ class Model {
     if (!options.include) return;
 
     options.include = _(options.include)
-      .groupBy(include => `${include.model && include.model.name}-${include.as}`)
+      .groupBy(include => {
+        const modelName =  include.model ? include.model.name : include.association;
+        return `${modelName}-${include.as}`;
+      })
```
frlinw: `baseMerge` call `_conformOptions` which mutates options.include and `_uniqInclude` is called after which process the mutated options.include so at this moment, include.model is always defined and include.association is an instance of Association, or I missed something?

Issue #10576, opened 2019-03-19
[Typescript] Many weird behaviors happens with Sequelize included type definitions


TitaneBoy:
## What are you doing?
Since I've updated Sequelize to v5 and using native provided type definition, I can't use models properly to do some request.

For example

```js
import debug = require('debug');
import { join } from 'path';
import { modelInstance, SequelizeSingleton } from './core/singletons/sequelizeSingleton';

const mySequelize = SequelizeSingleton.getInstance().sequelizer;
const debugConsole = debug('Ruby');

const session = mySequelize.import(join(process.cwd(), 'build/models/Sessions.js'));

/**
 * Asynchronously verify if a user is authenticated by verifying if a session ID exists in the Sessions database table.
 * If it is, the Promise will return the row associated with the session ID in the Session database table.
 *
 * @export
 * @param {(string | number)} sid The session ID to check in the Sessions database table
 * @returns {void}
 */
export function verifyAuthentication(sid: string | number) {
  return new Promise<modelInstance>((resolve, reject) => {
    mySequelize.authenticate().then(() => {
      return mySequelize.sync().then(() => {
        return session.findByPk(sid, {attributes: [ 'data' ]}).then((sessionData) => {
          if (!sessionData) {
            resolve();
          } else {
            resolve(sessionData);
          }
        }).catch((error) => {
          reject(error);
        });
      });
    }).catch((err) => {
      debugConsole('Unable to authenticate to the database:', err);
      reject(err);
    });
  });
}
```
This should not, normally, report any errors but I got this in my IDE:

![image](https://user-images.githubusercontent.com/15614466/54618707-d80f8d80-4a39-11e9-8990-4334430c1ed1.png)

![image](https://user-images.githubusercontent.com/15614466/54618756-e8c00380-4a39-11e9-925c-df313fd345cf.png)

![image](https://user-images.githubusercontent.com/15614466/54618774-ef4e7b00-4a39-11e9-9b3e-02af24a904f1.png)

![image](https://user-images.githubusercontent.com/15614466/54618782-f4132f00-4a39-11e9-9ab6-abf65d34609e.png)

As you can see, i'm trying to import my "Sessions" model and to make a "findByPk" request,  but Typescript is saying the error you can see in my previous screenshots. And when I mouse hover on the "then" word, it shows me an "any" return type, instead of a Promise type.

Also, When trying to see the findByPk with my cursor, I can see this:

[...]

PR #10584, opened 2019-03-20
docs(define): Update docs to use classes instead of `define`

SimonSchick:
### Pull Request check-list

### Description of change

First step in dealing with issues like

#10576 

by moving docs towards the more es6-ish and static analysable `Model.init` calls.


sushantdhiman:
`init` is not compatible with define for two reasons

1) Classes does not support getters with arguments
2) Class declaration requires `modelName` in camel-cased format. So `Model.init` call needs to pass `modelName: 'model'` to work exactly like `define`
3) [Just a preference for some] `define` is single step, `init` is double step

Due to (1) & (2) (& probably (3)) we can't move away from `define` yet. Should suggest this only for Typescript setups

PR #10596, opened 2019-03-22
Infer model attributes datatypes to basic types

Kilimangaro:

### Description of change

I've added some utility conditional types to extract/get  attributes "basic types" (string, number, boolean, Date) from a ModelAttributes

So : 
```typescript
const productAttributes = {
  sku: DataTypes.STRING,
  qty: {
     type: DataTypes.INTEGER,
     allowNull: true
  },
  instock: DataTypes.BOOLEAN,
  label: {
     type: DataTypes.STRING(100),
     allowNull: true
};
type TProduct = AttributesTypes<typeof productAttributes>;
// {
//   sku: string;
//   qty: number;
//   instock: boolean;
//   label: string;
// }
```

`sequelize.define` now return a model typed with all the attributes used, without the need to declare an interface

```typescript
const Product = sequelize.define('Product', 
    { 
        sku: DataTypes.STRING(3),
        availableOn: DataTypes.DATE,
        price: DataTypes.DECIMAL(7,2),
        instock: DataTypes.BOOLEAN
    }, 
    { tableName: 'users' }
);

async function test2() {
    const p = await Product.findOne();
    if (p) {
        p.instock = true;
        p.availableOn = new Date();
        p.price = 10;
        p.sku = 'P1000';
        await p.save();
    }
}
```

Some "creative" use of these new types

```typescript
export function defineModel<A extends ModelAttributes = ModelAttributes>(
    attributes: A,
    modelOptions: Partial<InitOptions> = {}
): { new (...args: any[]): Model & AttributesTypes<A> } & typeof Model & { connect: (sequelize: Sequelize) => void } {
    return class extends Model {
        static readonly associations: {
            [key: string]: Association;
        };
        public static connect(sequelize: Sequelize) {
            this.init(attributes, {
                ...modelOptions,
                sequelize
            });
        }
    } as any;
}

class ProductModel extends defineModel({
    productId: DataTypes.INTEGER,
    qty: {
        type: DataTypes.INTEGER,
        allowNull: true
    },
    sku: {
        type: DataTypes.STRING,
        allowNull: true
    },
    description: DataTypes.BLOB,
    instock: DataTypes.BOOLEAN
}) {
    categoryId: number;

    isAvailable() {
        return this.instock && this.qty > 0;
    }

    static associate() {
        return {
            Category: this.belongsTo(CategoryModel, {
                as: "Category",
                foreignKey: "categoryId"
            })
        };
    }
    static associations: ReturnType<typeof ProductModel.associate>;
}

class CategoryModel extends defineModel({
    categoryId: DataTypes.INTEGER,
    name: DataTypes.STRING
}) {
    static associate() {
        return {
            Products: this.hasMany(ProductModel, {
                as: "Products",
                foreignKey: "categoryId"
            })
        };
    }

    static associations: ReturnType<typeof CategoryModel.associate>;
}

export function connect(sequelize: Sequelize) {
    // Connect all models
    ProductModel.connect(sequelize);
    CategoryModel.connect(sequelize);

    // Create associations
    ProductModel.associate();
    CategoryModel.associate();
}

//  ... ... 

ProductModel.findOne().then(p => {
    if (p && p.isAvailable()) {
        return p;
    }
});

```

### DIFF
--- a/types/test/define.ts
+++ b/types/test/define.ts
@@ -30,3 +30,27 @@ async function test() {
 
     await user2.save();
 }
+
+
+// Instance types are infered from the model attributes
+
+const Product = sequelize.define('Product', 
+    { 
+        sku: DataTypes.STRING(3),
+        availableOn: DataTypes.DATE,
+        price: DataTypes.DECIMAL(7,2),
+        instock: DataTypes.BOOLEAN
+    }, 
+    { tableName: 'users' }
+);
+
+async function test2() {
+    const p = await Product.findOne();
+    if (p) {
+        p.instock = true;
+        p.availableOn = new Date();
+        p.price = 10;
+        p.sku = 'P1000';
+        await p.save();
+    }
+}
--- a/types/test/define.ts
+++ b/types/test/define.ts
@@ -1,4 +1,4 @@
-import { DataTypes, Model } from 'sequelize';
+import { DataTypes, Model, AttributesTypes, IntanceAttributesTypes } from 'sequelize';
 import { sequelize } from './connection';
 
 // I really wouldn't recommend this, but if you want you can still use define() and interfaces
@@ -32,11 +32,22 @@ async function test() {
 }
 
 
-// Instance types are infered from the model attributes
+// Types are infered from the model attributes
 
 const Product = sequelize.define('Product', 
-    { 
-        sku: DataTypes.STRING(3),
+    {
+        id: {
+            type: DataTypes.INTEGER,
+            primaryKey: true
+        }, 
+        sku: {
+            type: DataTypes.STRING(3),
+            /**
+             * "as false" is needed for typescript v < 3.4
+             * typescript 3.4 will allow to declare object "as const"
+             */
+            allowNull: false as false 
+        },
         availableOn: DataTypes.DATE,
         price: DataTypes.DECIMAL(7,2),
         instock: DataTypes.BOOLEAN
@@ -45,12 +56,29 @@ const Product = sequelize.define('Product',
 );
 
 async function test2() {
-    const p = await Product.findOne();
-    if (p) {
+   const p = await Product.findOne();
+
+    
+    if (p !== null) {
         p.instock = true;
-        p.availableOn = new Date();
+        
         p.price = 10;
+        p.price = undefined; // ok
+
         p.sku = 'P1000';
+        // p.sku = undefined; // error, sku does not allow null
+
+        // let availableOn: Date = p.availableOn;
+        let sku: string = p.get("sku");
+        let instock: boolean = p.get("instock");
+
         await p.save();
     }
+
+
+    const p2 = new Product();
+    // p2.id = 'NOPE'; // error
+    p2.price = 2.67;
+    await p2.save();
+    
 }
--- a/types/lib/data-types.d.ts
+++ b/types/lib/data-types.d.ts
@@ -310,7 +310,9 @@ export interface DecimalDataTypeOptions {
   scale?: number;
 }
 
-export interface BooleanDataType extends AbstractDataType {}
+export interface BooleanDataType extends AbstractDataType {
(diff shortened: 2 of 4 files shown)


SimonSchick:
This is both really interesting and very complicated and even took me a while to understand.

Some problems:
 - Will this capture `allowNull` and append the `| null` to type unions?
 - Will this work with enums(assuming type unions) type values?
 - Will this work with setters/getters?
 - There's probably more.

It's definitely a creative approach to auto typing models but I'm not sure if we want to pursuing given its complexity.

That said it might be worth looking into creating a script that traces calls on `Model.init` and association functions and auto generates class definitions based on those instead.

PS:
This really reminds of an attempt to auto generate a type from a [joi](https://github.com/hapijs/joi) schema, which ultimately didn't work out due to complexity issues (compile time got out of control and code was absolutely unreadable).

Kilimangaro:
@SimonSchick Yes, the goal is to automatically types models based on the attributes and the options given at the definition.

The task is not simple, but, i believe that we can gradually improve the types inferences by adding some smart conditional types.

I'm aware that we cant achieve a full types inferences of a model without a strict form of definition structure.

The `sequelize.define` "form"  will be limited to what is given to `define`. I'm confident that we can extract all the models attributes including getter and setter (in initOptions), but, there will be no types for associations, specific class and instances method.

The `class MyModel extends Model` form is much more open. There is many ways a model can be defined . One can call `Model.init` inside or outside the class, models associations can be done  inside or outside the class. So, i believe, that the only way to obtain a fully typed model is to follow some "creative" and "opinionated" form like the one in my example above. 

Maybe we can use the generic type of Model (`Model<T = any, T2 = any>`) and use these to upgrade methods  that return "intances" of model. (btw, why we have 2 generic here?)

```typescript
findAll<M extends Model, TInstance = M & T>(this: { new (): M } & typeof Model, options?: FindOptions): Promise<TInstance[]>;
```

> Will this capture allowNull and append the | null to type unions?

Actual PR does not handle "allowNull", but this can be "easilly" achieved since allowNull is an attribute of ModelAttributeColumnOptions. Something like this

```typescript
type ExtractAttributeAllowNull<T extends DataType | ModelAttributeColumnOptions> = 
  T extends DataType
  ? true :
  T extends ModelAttributeColumnOptions
  ? T["allowNull"] extends true ? true : false : false;
```
> Will this work with enums(assuming type unions) type values?

Yes, it work

```typescript
const enumDataType = DataTypes.ENUM("a", "b", "c");
type enumType = CastDataType<typeof enumDataType>;
let data1: enumType = "c"; // ok
let data2: enumType = "d"; // error
```

> Will this work with setters/getters?

Actually no, but this acn be achieved with another smart types

> There's probably more.

Sure ! There is much work to do :)

SimonSchick:
There are 2 generics to retain compatibility with the DT typings (at little to no cost).

Kilimangaro:
@SimonSchick Attributes types inference now return "partial" type for nullable attributes.

I've also made some use for **T** of Model (Model<**T** = any, T2 = any>). 

This new type:

```typescript
type ModelInstance<M> = M extends Model<infer I, any>? M & I : M;
```
Now allow Model.find* methods to return Model and **T**

types/test/define.ts
```diff
@@ -30,3 +30,55 @@ async function test() {
...
+            type: DataTypes.INTEGER,
+            primaryKey: true
+        }, 
+        sku: {
+            type: DataTypes.STRING(3),
+            /**
+             * "as false" is needed for typescript v < 3.4
+             * typescript 3.4 will allow to declare object "as const"
+             */
+            allowNull: false as false
```
SimonSchick: Eh, this seems like a hack and will be very hard to communicate to consumers.

sushantdhiman:
> I have think this will cause a lot of confusion without a lot of documentation that explains this and all it's caveats.

Closing this PR for now

PR #10630, opened 2019-03-26
chore(logger): streamline logging

SimonSchick:
### Pull Request check-list

### Description of change

Replaces `debug` package with `debuglog` native and acts as a foundation for https://github.com/sequelize/sequelize/pull/10284#discussion_r269296280

cc @javiertury 

~~#Fixes 10661~~

### DIFF
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -5,7 +5,7 @@ const SqlString = require('../../sql-string');
 const QueryTypes = require('../../query-types');
 const Dot = require('dottie');
 const deprecations = require('../../utils/deprecations');
-const logger = require('../../utils/logger').Logger.getLogger();
+const { logger } = require('../../utils/logger');
 
 class AbstractQuery {
 
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -330,7 +330,10 @@ class AbstractQuery {
     const { connection, options } = this;
     const benchmark = this.sequelize.options.benchmark || options.benchmark;
     const startTime = Date.now();
-    const fmt = `(${connection.uuid || 'default'}): ${sql}`;
+    let fmt = `(${connection.uuid || 'default'}): ${sql}`;
+    if (options.bind) {
+      fmt += `\n  BOUND:\n  ${options.bind.join('\n  ')}`;
+    }
     const msg = `Executing ${fmt}`;
     debugContext(msg);
     if (!benchmark) {
--- a/lib/dialects/abstract/query.js
+++ b/lib/dialects/abstract/query.js
@@ -5,6 +5,7 @@ const SqlString = require('../../sql-string');
 const QueryTypes = require('../../query-types');
 const Dot = require('dottie');
 const deprecations = require('../../utils/deprecations');
+const logger = require('../../utils/logger').Logger.getLogger();
 
 class AbstractQuery {
 
@@ -14,12 +15,9 @@ class AbstractQuery {
     this.model = options.model;
     this.sequelize = sequelize;
     this.options = Object.assign({
-      // eslint-disable-next-line no-console
-      logging: console.log,
       plain: false,
       raw: false
     }, options || {});
-
     this.checkLoggingOption();
   }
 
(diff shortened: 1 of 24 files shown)


eseliger:
Looking good so far, but the build is failing

SimonSchick:
hnnng

SimonSchick:
Yes, see https://nodejs.org/api/util.html#util_util_debuglog_section

SimonSchick:
I actually went ahead and streamlined all query logging into a single method `sequelize.logQuery` that handles both debug and regular logging mechanisms.
I also managed to purge a good amount of duplication while doing so.

Please note that it's relatively _easy_ to bring back the `debug` package if that's highly desired.

eseliger:
to me this looks very clean and like the way we want to be headed to. So if you guys feel like this is a breaking change I'd still be up for pushing this for a v6 release instead 🤔

SimonSchick:
I don't think this is breaking btw, in no place sequelize states that it uses `debug` or that you can use `DEBUG=sequelize` as env do enable debugging, it's not part of any public interfaces either.

eseliger:
I'm totally up for merging this, but the build is broken again 😔

SimonSchick:
Heck.

lib/dialects/postgres/query.js
```diff
@@ -43,28 +43,18 @@ class Query extends AbstractQuery {
   }
 
   run(sql, parameters) {
-    const connection = this.connection;
-    this.sql = sql;
+    const { connection } = this;
 
     if (!_.isEmpty(this.options.searchPath)) {
-      this.sql = this.sequelize.getQueryInterface().QueryGenerator.setSearchPath(this.options.searchPath) + sql;
+      sql = this.sequelize.getQueryInterface().QueryGenerator.setSearchPath(this.options.searchPath) + sql;
```
eseliger: hmm I think the idea was before to not change the input string by using the "copy" above. I don't know if there is any assumptions made it is immutable or not 🤔 but if the tests are passing, should be ok 
Also, same with the above

sushantdhiman:
This is a breaking change and can't be merged. Even if docs are lacking that we support `debug` I am sure many are using this feature since we added support for `debug`. Also the environment variable for logs will be changed from `DEBUG` to `NODE_DEBUG`

SimonSchick:
Imo the only _breaking_ change is the change in any *public* documentation or otherwise explicitly exposed format.

How would anyone know you can use `debug` if it's not in the docs?
They'd have to inspect the actual code to figure that out, and add that point 99% of developers would probably just add `console.log` at that point or use the `logging` option instead.

I have reverted it but please consider documenting the things you want developers to use/know about.

SimonSchick:
Boiled it down a little, the `DebugContext` class was a remnant of me trying make logging more lazy, ended up not using it.

SimonSchick:
@sushantdhiman snug in a fix for #10661

lib/utils/logger.js
```diff
@@ -11,38 +11,34 @@
 const debug = require('debug');
```
sushantdhiman: Not too happy with changes in this class.

1) You are caching `this.logger` on `Logger` class (that may confusing as `this` is used in static method). Make a const declaration after class with is an instance of class as `logger = new Logger()`.

2) We can probably get rid of both old and new `getLogger` method, just export instance `logger` of `Logger` class created in step (1), require and use that everywhere

lib/utils/logger.js
```diff
@@ -11,38 +11,34 @@
...
   }
 
   warn(message) {
     // eslint-disable-next-line no-console
     console.warn(`(${this.config.context}) Warning: ${message}`);
   }
 
-  debugContext(childContext) {
-    if (!childContext) {
-      throw new Error('No context supplied to debug');
```
sushantdhiman: Can probably let it slide, but this empty check may prevent logging to `sequelize:undefined` etc

lib/dialects/abstract/query.js
```diff
@@ -320,6 +320,41 @@ class AbstractQuery {
...
+    };
+  }
+
+  /**
+   * Gets a debug context for sub classes.
+   *
+   * @param {string} dialect
+   * @private
+   */
+  static makeDebugContext(dialect) {
```
sushantdhiman: Get rid of this method, let dialect query class create debug context by old `logger.debugContext(...)` method

lib/dialects/mariadb/query.js
```diff
@@ -1,21 +1,21 @@
...
 const ER_DUP_ENTRY = 1062;
 const ER_ROW_IS_REFERENCED = 1451;
 const ER_NO_REFERENCED_ROW = 1452;
 
+const debug = AbstractQuery.makeDebugContext('mariadb');
+
 class Query extends AbstractQuery {
   constructor(connection, sequelize, options) {
-    super(connection, sequelize, Object.assign({ showWarnings: false }, options));
+    super(connection, sequelize, debug, Object.assign({ showWarnings: false }, options));
```
sushantdhiman: Why debug is passed to super.constructor, `AbstractQuery` is not using debug in constructor, third argument should be `options`.

SimonSchick: Uh oh.

lib/utils/logger.js
```diff
@@ -11,38 +11,28 @@
 const debug = require('debug');
 
 class Logger {
+  static warn(message) {
```
sushantdhiman: static `warn` can be removed, just use `warn` from instance exported

lib/dialects/abstract/query.js
sushantdhiman: Resolved? `AbstractQuery` shouldn't be responsible for this, let's got with simple approach?

SimonSchick: Oh, final push didn't go through 😅

PR #10631, opened 2019-03-27
test: fix running query count in test suite

eseliger:
### Pull Request check-list

### Description of change

This should (hopefully) fix all the errors in the test-suite that were thrown when runningQueries was not 0. This was bc the counter was increased every try (retries included), but only decreased once for each query call.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -571,14 +571,11 @@ class Sequelize {
 
         return query.run(sql, bindParameters)
           .finally(() => {
+            if (this.test._trackRunningQueries) this.test._runningQueries--;
             if (!options.transaction) {
               return this.connectionManager.releaseConnection(connection);
             }
           });
-      }).finally(() => {
-        if (this.test._trackRunningQueries) {
-          this.test._runningQueries--;
-        }
       }), retryOptions));
     });
   }
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -266,7 +266,7 @@ class Sequelize {
         this._trackRunningQueries = true;
       },
       verifyNoRunningQueries() {
-        if (this._runningQueries > 0) throw new Error(`Expected 0 running queries. ${this._runningQueries} queries still running`);
+        if (this._runningQueries !== 0) throw new Error(`Expected 0 running queries. ${this._runningQueries} queries still running`);
       }
     };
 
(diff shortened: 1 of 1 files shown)


eseliger:
Seems like, it works like a charm for local testing now...

eseliger:
maybe there is an actual leak though, there was still one failure in the pg-native tests 
will look into it

lib/sequelize.js
```diff
@@ -577,11 +577,11 @@ class Sequelize {
               return this.connectionManager.releaseConnection(connection);
             }
           });
-      }), retryOptions)).finally(() => {
+      }).finally(() => {
         if (this.test._trackRunningQueries) {
           this.test._runningQueries--;
         }
-      });
+      }), retryOptions));
```
sushantdhiman: If query is retried for any reason (could be `options.retry`) the `_runningQueries` count will become negative. I suspect this is why it works now.

Please verify if `_runningQueries` count is ever negative in your local tests

sushantdhiman:
> This was bc the counter was increased every try (retries included), but only decreased once for each query call.

Doesn't the counter only increase for first try https://github.com/sequelize/sequelize/blob/ab7a95487575f6d30aff5bb08cb14331d863ac0b/lib/sequelize.js#L554

lib/sequelize.js
eseliger: oof. you are right. I verified locally and you were right, my bad. 
I went ahead and played around with it a little more and figured out that it makes sense though to increase and decrease the counter on every try because there is the edge-case of looking at hooks (especially when I think of the pre and post query hooks we talked about earlier) and therefore, it would still be marked as running although it's about to attempt the next try. What do you think?

lib/sequelize.js
```diff
@@ -577,11 +575,11 @@ class Sequelize {
               return this.connectionManager.releaseConnection(connection);
             }
           });
-      }), retryOptions)).finally(() => {
+      }).finally(() => {
```
sushantdhiman: this block can be removed then, merge it with `finally` block just above it which is releasing connection. decrement then release connection

papb:
Ohh, it would be great to see these intermittent travis errors stop once and for all :grin:

eseliger:
closing in favor of a new one

papb:
@eseliger - what's the new one?

papb:
Nevermind, I think I found it, seems to be #10891 - good work!

eseliger:
whoops! yes you are right, that's the one :)

PR #10644, opened 2019-03-28
fix(mysql): boolean TINYINT support

### LINKED ISSUE #10600: v4.43.0 broke delete flag handling
## What are you doing?

After upgrading to `4.43.0` (from `4.42.0`), we can see the following SQL being produced:

```sql
UPDATE `Response` SET `deleted`='true',`updatedBy`=24751,`updatedAt`=NOW() WHERE `id` = 416695 AND `deleted` = 'false'
```

Note that `deleted` is being correctly to a string. This leads to the following error:

```
Error Code: 1292. Truncated incorrect DOUBLE value: 'false'
```

This code being used is:

```js
await Response.update(
  {
    deleted: true,
  },
  {
    where: {id: existingResponse.id},
    ...context,
  }
);
```

The column is defined in the model as:

```js
deleted: {
  type: DataTypes.TINYINT,
},
```

Reverting to 4.42.0 fixes the issue.

The latest version of Sequelize 4 should not convert my deleted booleans to strings.

The latest version of Sequelize 4 is converting my deleted booleans to strings.

__Dialect:__ mysql
__Database version:__ 5.7.23
__Sequelize version:__ 4.43.0
__Tested with latest release:__ Yes (of Sequelize 4)

### LINKED ISSUE #10633: Boolean values not parsed properly to integer fields for MySQL
## What are you doing?
```js
isPublic: {type: DataTypes.BOOLEAN, allowNull: false, defaultValue: false}

// ---------
// Later on...

Model.create({key1: 'value1', key2: 'value2'})
//or...
Model.create({key1: 'value1', key2: 'value2', isPublic: true})
```

_Record being created into the database with a proper converted default value to __0 or 1__, since MySQL uses `TINYINT(1)` as the column type for boolean fields._

`SequelizeDatabaseError: Incorrect integer value: 'false' for column 'isPublic' at row 1`
`SequelizeDatabaseError: Incorrect integer value: 'true' for column 'isPublic' at row 1`

__Dialect:__ mysql
__Dialect version:__ 2.6.0
__Database version:__ 8.0.15
__Sequelize version:__ 5.2.1

gabegorelick:
This regressed in #10492

Fixes #10600
Fixes #10633

### Pull Request check-list

### Description of change

Bring back support for passing booleans into `TINYINT` fields.

### DIFF
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -248,6 +248,43 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
     });
   });
 
+  if (dialect === 'mysql') {
+    it('should handle TINYINT booleans', function() {
+      const User = this.sequelize.define('user', {
+        id: { type: Sequelize.TINYINT, primaryKey: true },
+        isRegistered: Sequelize.TINYINT
+      });
+
+      return User.sync({ force: true }).then(() => {
+        return User.create({ id: 1, isRegistered: true });
+      }).then(registeredUser => {
+        expect(registeredUser.isRegistered).to.equal(true);
+        return User.findOne({
+          where: {
+            id: 1,
+            isRegistered: true
+          }
+        });
+      }).then(registeredUser => {
+        expect(registeredUser).to.be.ok;
+        expect(registeredUser.isRegistered).to.equal(1);
+
+        return User.create({ id: 2, isRegistered: false });
+      }).then(unregisteredUser => {
+        expect(unregisteredUser.isRegistered).to.equal(false);
+        return User.findOne({
+          where: {
+            id: 2,
+            isRegistered: false
+          }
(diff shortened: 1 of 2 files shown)


test/integration/data-types.test.js
```diff
@@ -248,6 +248,43 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
     });
   });
 
+  if (dialect === 'mysql') {
```
SimonSchick: I think we could make this a simple unit test, not mysql specific?

gabegorelick: This behavior is specific to MySQL. On other platforms, where `TRUE` and `FALSE` are not aliases for `1` and `0` respectively, passing a boolean literal where a number is expected will result in an error.

PR #10660, opened 2019-03-30
fix(mysql): backport boolean TINYINT support to v4

### LINKED ISSUE #10600: v4.43.0 broke delete flag handling

gabegorelick:
### Pull Request check-list

### Description of change

Backports #10644 to v4. Since the original feature that broke this was backported to v4, it's only fair to backport the fix too. 

Fixes #10600

### DIFF
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -147,7 +147,7 @@ NUMBER.prototype.validate = function(value) {
   return true;
 };
 NUMBER.prototype._stringify = function _stringify(number) {
-  if (typeof number === 'number' || number === null || number === undefined) {
+  if (typeof number === 'number' || typeof number === 'boolean' || number === null || number === undefined) {
     return number;
   }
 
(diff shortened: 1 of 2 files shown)


gabegorelick:
Thanks as always @sushantdhiman!

PR #10662, opened 2019-03-30
fix(model): make aggregate merge attributes rather than override

SimonSchick:
### Pull Request check-list

### Description of change

I cba to create an issue for this for now.

This was broken by:
https://github.com/sequelize/sequelize/commit/88a340dacec0b1b27d4f67866e2b416482ce7b85#diff-a140b60b6a99a8c69f4b5d0d1f1f2bfaR1964

So cc @u9r52sld


test/integration/associations/has-many.test.js
```diff
@@ -1494,7 +1494,7 @@ describe(Support.getTestDialectTeaser('HasMany'), () => {
             include: [
               { model: Task, separate: true, include: [{ model: Job, separate: true }] }
             ],
-            group: [['userId']]
+            group: ['userId']
```
sushantdhiman: This just became a breaking change :D

SimonSchick: Not really, the `group` here never worked right since the `intersect` was added which would've removed it.

eseliger: Sounds reasonable to me, this feels like it was broken before.. So IMO we could also go ahead and and push this to v5

eseliger:
@SimonSchick @sushantdhiman how would we want to proceed? Can we convince you it's just a bug-fix or would we want to push this to a v6 branch?

SimonSchick:
Imo this should be patchsince it's a bug fix, a nested array was never supported (and never worked right).

test/integration/associations/has-many.test.js
sushantdhiman: @SimonSchick 

```sql
Executing (default): SELECT count("user"."userId") AS "count" FROM "users" AS "user" GROUP BY "user"."userId";
Executing (default): SELECT "user"."userId" FROM "users" AS "user" GROUP BY "user"."userId";
Executing (default): SELECT "task"."taskId", "task"."title", "task"."createdAt", "task"."updatedAt", "task"."userId" FROM "tasks" AS "task" WHERE "task"."userId" IN (1);
Executing (default): SELECT "jobId", "title", "createdAt", "updatedAt", "taskId" FROM "jobs" AS "job" WHERE "job"."taskId" IN (1, 2);
```

Seems like `group` is working properly here. You can see `GROUP BY userId` in first two queries.

We can make this change non-breaking by taking care of arrays like `[[item]]`

SimonSchick: What if you remove the `group` in this case entirely?
What happened to me is that the same query was generated eg. `the double nested `group` had no effect.

SimonSchick:
@sushantdhiman if you want you can flatten whatever is passed to `group` if you think that's best.
Feel free to push to this branch, I'd like to get this fixed asap.
Please leave a deprecation notice tho.

SimonSchick:
@sushantdhiman ping

sushantdhiman:
@SimonSchick This can't be merged in this state, it needs to support `[[attr]]` format or go to v6. If you don't have time for that change we can close this one for now

SimonSchick:
If that's what it takes to fix an actual regression...
I will add a deprecation warning tho.

SimonSchick:
Done.

lib/utils/deprecations.js
```diff
@@ -8,3 +8,4 @@ exports.noRawAttributes = deprecate(noop, 'Use sequelize.fn / sequelize.literal
 exports.noTrueLogging = deprecate(noop, 'The logging-option should be either a function or false. Default: console.log', 'SEQUELIZE0002');
 exports.noStringOperators = deprecate(noop, 'String based operators are deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/querying.html#operators', 'SEQUELIZE0003');
 exports.noBoolOperatorAliases = deprecate(noop, 'A boolean value was passed to options.operatorsAliases. This is a no-op with v5 and should be removed.', 'SEQUELIZE0004');
+exports.noDoubleNestedGroup = deprecate(noop, 'Passing a double nested nested array to `group` is unsupported and will be removed in v6.', 'SEQUELIZE0005');
```
sushantdhiman: This pattern is supported everywhere, do you plan to remove this from all other segments, for now this looks like it is only deprecated from `aggregate` methods

SimonSchick: Does it also work in `findAll` etc?

PR #10674, opened 2019-03-31
docs: tutorial improvements

papb:
### Description of change

Improves a bit the tutorials:

* Quick start script now works (changed `Model` to `Sequelize.Model`)
* A few other clarifications, link updates and minor formatting changes

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -90,7 +90,7 @@ Sequelize will keep the connection open by default, and use the same connection 
 
 ## Modeling a table
 
-A model is a class that extends `Sequelize.Model`. Models are defined with `Sequelize.Model.init(attributes, options)`:
+A model is a class that extends `Sequelize.Model`. Models can be defined in two equivalent ways. The first, with `Sequelize.Model.init(attributes, options)`:
 
 ```js
 const Model = Sequelize.Model;
@@ -111,7 +111,7 @@ User.init({
 });
 ```
 
-Alternatively (legacy declaration, using `sequelize.define`):
+Alternatively, using `sequelize.define`:
 
 ```js
 const User = sequelize.define('User', {
@@ -128,6 +128,8 @@ const User = sequelize.define('User', {
   // options
 });
 ```
+
+Internally, `sequelize.define` calls `Model.init`.
 
 The above code tells Sequelize to expect a table named `users` in the database with the fields `firstName` and `lastName`. The table name is automatically pluralized by default (a library called [inflection](https://www.npmjs.com/package/inflection) is used under the hood to do this). This behavior can be stopped for a specific model by using the `freezeTableName: true` option, or for all models by using the `define` option from the [Sequelize constructor](http://docs.sequelizejs.com/class/lib/sequelize.js~Sequelize.html#instance-constructor-constructor).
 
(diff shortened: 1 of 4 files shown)


docs/index.md
```diff
@@ -33,7 +33,7 @@ You are currently looking at the **Tutorials and Guides** for Sequelize. You mig
 const Sequelize = require('sequelize');
```
SimonSchick: Can we change this to `const { Sequelize, Model }`, just demonstrates the exports a little bit better.

papb: Good point, I think that form of importing is nice as well. Since in part this is a matter of style, and I also think that it would be good to make it clear that `Model` can be found directly in `Sequelize` (by `Sequelize.Model`), I decided to leave "both methods" in the example, one of them commented. What do you think?

I think this way the example conveys a lot of "information per line of code example", so to speak :)

papb:
Looks like codecov has gone crazy. And it's not the first time I see it happening, this is curious...

docs/index.md
SimonSchick: Keep in mind that `module.exports = Sequelize` and `Sequelize.Sequelize === Sequelize`, explaining that to consumers is kinda tricky.
Note: I seriously wish this wasn't done, it makes imports/exports with Sequelize a PITA to model in TS.
I think we should try to restrict documentation to only use one way to import from sequelize for consistency, but it's fair to mention that it's 'aliased'.

papb: I see, `Sequelize.Sequelize` is a bit odd indeed but personally I can accept it, I guess. I could also accept it not being done though. You said it's *kinda tricky* but maybe it's not that complicated. I think that if we find a quick way to pass that information, it's fine. Consumers deserve to know (unless it would get too confusing). Perhaps this doesn't have to be in the first page of the docs, though. I will make some more changes soon.

papb: @SimonSchick - On a second thought, would you agree with leaving this problem of "how to import sequelize" for another PR? I agree with you that the way it was done isn't so good, and that at least the documentation should take a stance on a *standard* way to do things, and this is important, but I don't want this to block the merge of my PR, since it does some other things as well.

So, I'd like to ask, would you agree with using my original version for now? At least it stays consistent with other tutorials. On the other PR, which I could help with later, we shall update the whole documentation on this matter. What do you say? I will revert my previous commit in advance. Thanks a lot.

SimonSchick: Yea, this would require cleanup on a bigger scale anyway, just changing it here would be more of a more of a drop in the pond.

docs/getting-started.md
```diff
@@ -113,7 +111,7 @@ User.init({
 });
 ```
 
-Alternatively (legacy):
+Alternatively (legacy declaration, using `sequelize.define`):
```
sushantdhiman: `sequelize.define` was marked legacy? when? I thought that was only for TypeScript. `init` requires two step where `define` is single step. It should be ok to use it on non-typescript projects.

Please remove `legacy` :)

papb: Oh, very well. Looks like it was a misunderstanding. Sorry about that. It wasn't me who added the `legacy` terminology for it, I just saw it being done and assumed it happened. In fact I did think it was a bit strange, so it is good from hear to you that it's not really legacy. Great! Will fix it later.

SimonSchick: Yea my bad, shouldn't have labeled legacy, I think you can call this `alternative (non es6)` or something along those lines.

papb: @sushantdhiman - Done! Let me know if you want more changes.

PR #10677, opened 2019-03-31
docs(typescript): improve typescript docs and add peer deps

### LINKED ISSUE #10668: model type can not be infer because of the lack of @types/bluebird
## What are you doing?

I am using `sequelize` with typescript. Below is my demo file
```js
// code here
import {Sequelize, Model, STRING, DATE} from 'sequelize'
const sequelize = new Sequelize('postgres://user:pass@example.com:5432/dbname');

class User extends Model {}
User.init({
  username: STRING,
  birthday: DATE
}, { sequelize });

User.create({
  username: 'janedoe',
}).then(vvvv => {

})
```

`vvvv` should be infer to type `User`, but it doesn't.

`vvvv` is infer to `any`.

when I run `npm i @types/bluebird`, the infer works.

__Dialect:__ any
__Dialect version:__ any
__Database version:__ any
__Sequelize version:__ 5.2.7
__Tested with latest release:__ Yes. 5.2.7

SimonSchick:
### Pull Request check-list

### Description of change

Fixes #10668

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,11 +85,6 @@
     "typescript": "^3.3.3333",
     "semantic-release": "^15.13.3"
   },
-  "peerDependencies": {
-    "@types/node": ">=6.14.3",
-    "@types/validator": "^10.11.0",
-    "@types/bluebird": "^3.5.26"
-  },
   "keywords": [
     "mysql",
     "mariadb",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "@commitlint/config-angular": "^7.3.1",
     "@types/bluebird": "^3.5.26",
     "@types/node": "^10.12.27",
-    "@types/validator": "^10.9.0",
+    "@types/validator": "^10.11.0",
     "big-integer": "^1.6.42",
     "chai": "^4.x",
     "chai-as-promised": "^7.x",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -85,6 +85,11 @@
     "typescript": "^3.3.3333",
     "semantic-release": "^15.13.3"
   },
+  "peerDependencies": {
```
sushantdhiman: Hmm, what about non-typescript users. They shouldn't be forced to add these dependencies to their project.

SimonSchick: That was my thought too, that's why I didn't add them initially, but we've had a few users (see issue) complain about 'weird' errors due to these packages not being installed.

`peerDependencies` don't automatically install but give you a small warning which you can ignore, might be a fair trade off.

papb: @SimonSchick @sushantdhiman I am starting to think that `peerDependencies` is a bad idea too. I wouldn't like to see these as a non-TS user. What if we add a script to run once sequelize is loaded, to auto detect if typescript is being used, check if the dependencies are installed and print a warning message automatically on the console if they're missing?

papb: Also, about "some users complaining", I think that once the rest of this PR is merged (without peerDependencies), the documentation will be much clearer on TypeScript usage and these mistakes will be less frequent. How about just removing the `peerDependencies` thing from this PR and waiting a bit to see if the confusion remains?

papb: Someone else just had the same problem. I think this PR should be merged ASAP - but without the `peerDependencies` part. If the confusion persists, then we can think about it again.

(this is just my opinion, of course)

sushantdhiman: @SimonSchick remove peer dependencies and merge it

PR #10685, opened 2019-04-02
docs: lint the markdown docs by markdownlint

ALiangLiang:
### Pull Request check-list

### Description of change

I think our docs is hard to read. And too many unnecessary continuously blank lines 

like this below. It's hard to know the scope of cope block.

~~~~
```
code1 here
```
content here
```
code2 here
```
~~~~

So I follow markdownlint, add a blank line between them and fill in the type on the code block.

~~~~
```js
code1 here
```

content here

```text
code2 here
```
~~~~

---

BTW. I want to discuss should we introduce [markdownlint](https://github.com/DavidAnson/markdownlint) like `eslint`?

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -6,7 +6,7 @@ For a full list of hooks, see [Hooks file](https://github.com/sequelize/sequeliz
 
 ## Order of Operations
 
-```
+```text
 (1)
   beforeBulkCreate(instances, options)
   beforeBulkDestroy(options)
@@ -42,6 +42,7 @@ For a full list of hooks, see [Hooks file](https://github.com/sequelize/sequeliz
 ```
 
 ## Declaring Hooks
+
 Arguments to hooks are passed by reference. This means, that you can change the values, and this will be reflected in the insert / update statement. A hook may contain async actions - in this case the hook function should return a promise.
 
 There are currently three ways to programmatically add hooks:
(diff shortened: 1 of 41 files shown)


docs/hooks.md
```diff
@@ -244,7 +248,7 @@ User.create({username: 'Boss', accessLevel: 20}).then(user => {
 
 Sometimes you'll be editing more than one record at a time by utilizing the `bulkCreate, update, destroy` methods on the model. The following will emit whenever you're using one of those methods:
 
-```
+```js
```
SimonSchick: This really isn't valid JS.

docs/whos-using.md
```diff
@@ -1,4 +1,4 @@
-# Who's using sequelize?
```
papb: I personally didn't like the removal of this question mark... Was this automatically suggested by `markdownlint`?

papb:
As for your question, I personally think it would be nice to have `markdownlint` configured. But not in this PR, I think it would be better to have a different PR for it, where we can discuss precisely which options would be enabled/disabled.

docs/whos-using.md
ALiangLiang: Yes. It's a [rule](https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md026) `Trailing punctuation in heading`.

ALiangLiang: But I prefer to follow the rule. No question mark is more like a title.

papb: I see... I would personally suggest changing the default config of markdownlint to allow question marks. It's even an example in the docs you linked to. But if you really disagree, at least I think the whole title should be changed then. "Who's using sequelize" without a question mark seems odd to me. Would you suggest a different title?

ALiangLiang: Um..... I was persuaded. Add the question mark back.

ALiangLiang:
@papb Good idea. Later I will create a PR about introducing markdownlint.

sushantdhiman:
@ALiangLiang yeah markdownlint is a good idea

ALiangLiang:
@papb 
I was created the PR. #10692

PR #10686, opened 2019-04-02
docs(data-types): move data-types.js to data-types/index.js

ALiangLiang:
### Pull Request check-list

### Description of change

The `esdoc` group the classes by path. So I move `data-types.js` to `data-types/index.js` to separate to be a group.

image

to

image


SimonSchick:
You will probably have to fix a bunch of include paths.

sushantdhiman:
I don't think we should be moving file just for documentation stake. I would rather go for some different solution if available like `esdoc` config, CSS or a plugin may be

papb:
@sushantdhiman Hi, any chance you would reconsider this? My PR #10898 improved things a bit but the [API Reference main page](https://sequelize.org/master/identifiers.html) is still cluttered. I could do something similar of #10898, but I think this would start getting way too complicated. With a separated folder we could also link directly to it, improving the datatypes manual, etc.

papb:
Of course, to merge this my PR #10898 should be reverted, which I don't mind.

PR #10692, opened 2019-04-03
Introduce markdownlint

ALiangLiang:
### Pull Request check-list

### Description of change

Now we write manuals through markdown, which are irregular and make it difficult to read. So I suggest introduce [markdownlint](https://github.com/igorshubovych/markdownlint-cli) to rule out markdown style. And I try to pick loose rules.

If everybody has questions or opinions, please feel free to comment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,6 +116,7 @@
   },
   "scripts": {
     "lint": "eslint lib test --quiet",
+    "lint-docs": "markdownlint docs",
     "test": "npm run teaser && npm run test-unit && npm run test-integration",
     "test-docker": "npm run test-docker-unit && npm run test-docker-integration",
     "test-docker-unit": "npm run test-unit",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -71,19 +71,20 @@
     "lcov-result-merger": "^3.0.0",
     "lint-staged": "^8.1.5",
     "mariadb": "^2.0.3",
+    "markdownlint-cli": "^0.14.1",
```
SimonSchick: Should we add yet another tooling dependency?

I agree with linting pass you did in general but integrating it into the CI seems a little overkill 😛

sushantdhiman: Well it gives us more tooling :smile: I was also thinking about a linting stage in CI so others build wont have to run lint everytime

SimonSchick: Yea that would seem reasonable, you can also squash TS and all of that in there.
The downside is that it will still try to run tests even if lint failed due to a coding issue (undefined variabled etc.).

sushantdhiman: If one stage fails in travis all next stages are cancelled, so that should not be a problem

SimonSchick: Oh I misunderstood here, I thought you wanted to extract all lint tasks into a separate environment var gated pipeline. 😛

PR #10694, opened 2019-04-03
fix(types): add basic support for `bind` in Sequelize.query() method

devoto13:
### Pull Request check-list

### Description of change

The `bind` property described in the [docs](http://docs.sequelizejs.com/manual/raw-queries.html#bind-parameter) was missing in the type definitions.

### DIFF
--- a/types/lib/query-interface.d.ts
+++ b/types/lib/query-interface.d.ts
@@ -5,6 +5,8 @@ import QueryTypes = require('./query-types');
 import { Sequelize, RetryOptions } from './sequelize';
 import { Transaction } from './transaction';
 
+type BindOrReplacements = { [key: string]: unknown } | unknown[];
+
 /**
  * Interface for query options
  */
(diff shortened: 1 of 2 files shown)


types/lib/query-interface.d.ts
```diff
@@ -44,6 +44,14 @@ export interface QueryOptions extends Logging, Transactionable {
     [key: string]: string | number | string[] | number[];
   } | string[];
 
+  /**
+   * Either an object of named parameter replacements in the format `:param` or an array of unnamed
+   * replacements to replace `?` in your SQL.
+   */
+  bind?: {
```
SimonSchick: Can you create a type here to avoid duplication?

devoto13: Sure, done.

devoto13:
While upgrading our codebase from 4 to 5 I've noticed that some queries also bind `null` and arbitrary JSON (for PostgreSQL JSON field). Probably `bind` should have more relaxed type than  `replacements`. Something like `{ [key: string]: any } | any[]`. What do you think?

SimonSchick:
Not `any` no, we can define a simple type which represents anything json serializable.

But for now `unknown` would be ok

Issue #10285, opened 2018-12-19
Add support for bind parameters


javiertury:
I would like sequelize to support bind parameters. All the work I've done so far is in pull request #10284.

All user-facing queries bind parameters. Performance is also taken into consideration to avoid round trips. If you have any issue, comment or question, please discuss.

javiertury:
Replaced by #11586

PR #10703, opened 2019-04-05
feat(log): configurable inspector

javiertury:
Provides a configurable `logger.inspector` that will format non-string JS elements in logs and errors. This will be useful for configuring how binded parameters are logged, once that functionality is merged into sequelize.

### DIFF
--- a/lib/utils/logger.js
+++ b/lib/utils/logger.js
@@ -9,6 +9,7 @@
  */
 
 const debug = require('debug');
+const util = require('util');
 
 class Logger {
   constructor(config) {
@@ -22,6 +23,10 @@ class Logger {
   warn(message) {
     // eslint-disable-next-line no-console
     console.warn(`(${this.config.context}) Warning: ${message}`);
+  }
+
+  inspect(value) {
+    return util.inspect(value, false, 3);
   }
 
   debugContext(name) {
--- a/lib/sql-string.js
+++ b/lib/sql-string.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const dataTypes = require('./data-types');
-const util = require('util');
+const { logger } = require('./utils/logger');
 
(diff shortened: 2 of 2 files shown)


lib/utils/logger.js
```diff
@@ -15,7 +17,8 @@ class Logger {
 
     this.config = Object.assign({
       context: 'sequelize',
-      debug: true
+      debug: true,
+      inspector: value => util.inspect(value, false, 3)
```
sushantdhiman: No need to pass this in config I think, just use this as default in the new `inspect` method

javiertury: Agree

lib/utils/logger.js
```diff
@@ -1,5 +1,11 @@
 'use strict';
 
+const util = require('util');
```
sushantdhiman: Move this line to the line after `require('debug');`

lib/utils/logger.js
```diff
@@ -1,5 +1,11 @@
...
+
+function defaultInspector(value) {
```
sushantdhiman: Get rid of this, just use `util.inspect(value, false, 3);` directly in `inspect` method

lib/utils/logger.js
```diff
@@ -24,6 +25,13 @@ class Logger {
     console.warn(`(${this.config.context}) Warning: ${message}`);
   }
 
+  inspect(value) {
+    if (this.config.inspector) {
```
sushantdhiman: There is no need for this config, as there is currently no way sequelize let users pass this config. So they can't override this.

Proper way to override will be to just replace `logger.inspect` method that is exported by this module.

sushantdhiman:
Remove the submodule

javiertury:
Done

PR #10704, opened 2019-04-05
feat(connection): deprecate sql server 2008

javiertury:
Deprecate sql server 2008 which will become completely unsupported by Microsoft in July 9, 2019. This will helps us clean sequelize code by removing some mssql specific hacks in the future.

The `versionSupport` method can be easily extended to other dialects. Additionally, use `semver.coherce` for robustness. Some DBs that report incomplete non-conforming semver versions (e.g. 9.5 instead of 9.5.0).

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -6,6 +6,7 @@ const semver = require('semver');
 const Promise = require('../../promise');
 const errors = require('../../errors');
 const { logger } = require('../../utils/logger');
+const { noUnsupportedVersion } = require('../../utils/deprecations');
 const debug = logger.debugContext('pool');
 
 /**
@@ -40,6 +41,11 @@ class ConnectionManager {
     });
 
     this.initPools();
+
+    // Check version support, if set through options
+    if (this.versionSupport) {
+      this.versionSupport();
+    }
   }
 
   refreshTypeParser(dataTypes) {
@@ -241,10 +247,8 @@ class ConnectionManager {
 
     let promise;
     if (this.sequelize.options.databaseVersion === 0) {
-      if (this.versionPromise) {
-        promise = this.versionPromise;
-      } else {
-        promise = this.versionPromise = this._connect(this.config.replication.write || this.config)
+      if (!this.versionPromise) {
+        this.versionPromise = this._connect(this.config.replication.write || this.config)
           .then(connection => {
             const _options = {};
 
@@ -267,11 +271,17 @@ class ConnectionManager {
 
             this.versionPromise = null;
             return this._disconnect(connection);
+          }).then(() => {
+            // Check version support, if autodetected
+            if (this.versionSupport) {
+              return this.versionSupport();
+            }
           }).catch(err => {
             this.versionPromise = null;
             throw err;
           });
       }
+      promise = this.versionPromise;
     } else {
       promise = Promise.resolve();
     }
@@ -334,6 +344,15 @@ class ConnectionManager {
     }
 
     return this.dialect.connectionManager.validate(connection);
+  }
+
+  versionSupport() {
+    const version = _.get(this.sequelize, 'options.databaseVersion', 0);
+    const minVer = this.dialect.supports.minimumSupportedVersion;
+
+    if (minVer && semver.valid(version) && semver.lt(version, minVer)) {
+      noUnsupportedVersion(this.dialectName, version, minVer);
+    }
   }
 }
 
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -5,6 +5,7 @@ const path = require('path');
 const retry = require('retry-as-promised');
 const clsBluebird = require('cls-bluebird');
 const _ = require('lodash');
+const semver = require('semver');
 
 const Utils = require('./utils');
 const Model = require('./model');
@@ -180,6 +181,11 @@ class Sequelize {
       typeValidation: false,
       benchmark: false
     }, options || {});
+
+    // Coherce databaseVersion to valid semver format
+    if (this.options.databaseVersion) {
+      this.options.databaseVersion = semver.valid(semver.coerce(this.options.databaseVersion)) || 0;
+    }
 
     if (!this.options.dialect) {
       throw new Error('Dialect needs to be explicitly supplied as of v4.0.0');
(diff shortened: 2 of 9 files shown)


javiertury:
@sushantdhiman, now only a warning is shown

lib/dialects/mssql/connection-manager.js
```diff
@@ -166,6 +169,15 @@ class ConnectionManager extends AbstractConnectionManager {
 
     return connection && connection.loggedIn;
   }
+
+  versionSupport() {
+    const version = _.get(this.sequelize, 'options.databaseVersion', 0);
+
+    if (semver.valid(version) &&
+      semver.lt(this.sequelize.options.databaseVersion, '11.0.0')) {
+      logger.warn('SQL Server 2008 and lower versions are not supported by Sequelize');
```
SimonSchick: I think we should be using the deprecation functions instead 😛

javiertury: I've tried to use deprecations. There is a catch, deprecations cannot be spied, even if they are wrapped in another function. I can only spy on methods(like those of Logger) with sinon.

Another issue are deprecation codes, should we use a new code for each  dialect and required version?

SimonSchick: What I did is just wrap noops in `util.deprecate`, see `deprecations.js`

Example:
```
exports.noBoolOperatorAliases = deprecate(noop, 'A boolean value was passed to options.operatorsAliases. This is a no-op with v5 and should be removed.', 'SEQUELIZE0004');
```

Just use an incremental number for the code.

javiertury: Yes, I tried something similar. The problem is that `require('.../deprecations')` exports a node module composed of functions, which cannot be spied on. Yesterday I tried to no avail

```js
// deprecations.js
exports.noUnsupportedDB = (dialect, minVer, version) => deprecate(noop, `(${dialect}) Database version ${version} is unsupported. Only ${minVer} and higher are supported.`, 'SEQUELIZE0005')();

// connection-manager.test.js
const depreSpy = spy(deprecations, 'noUnsupportedDBVersion')
```

But the spy didn't work. It seems spy can only spy on an instance method.

SimonSchick: Oh you can't spy on them because they are referenced directly 🤔

sushantdhiman:
This needs some cleanup

1. Define a new key `minimumSupportedVersion: undefined` [here](https://github.com/sequelize/sequelize/blob/c44c72ae23b6add02f2dfd7390431491a934aeec/lib/dialects/abstract/index.js#L6)
2. For MSSQL set that to `11.0.0` [here](https://github.com/sequelize/sequelize/blob/c44c72ae23b6add02f2dfd7390431491a934aeec/lib/dialects/mssql/index.js#L23)
3. After version is acquired you can check if it is less than `minimumSupportedVersion` [here](https://github.com/sequelize/sequelize/blob/c44c72ae23b6add02f2dfd7390431491a934aeec/lib/dialects/abstract/connection-manager.js#L270), given that `minimumSupportedVersion` is not empty
4. For now a deprecation warning from [here](https://github.com/sequelize/sequelize/blob/master/lib/utils/deprecations.js) will be used, in next version we can throw an error

Once you update with master you will notice there is some duplication with this certain code [here](https://github.com/sequelize/sequelize/blob/c44c72ae23b6add02f2dfd7390431491a934aeec/lib/dialects/abstract/connection-manager.js#L259-L262) and [here](https://github.com/sequelize/sequelize/blob/c44c72ae23b6add02f2dfd7390431491a934aeec/lib/dialects/postgres/connection-manager.js#L130-L133). We should move this is utils perhaps and add some more unit tests against this. With that there is no need to parse version [here](https://github.com/sequelize/sequelize/blob/c44c72ae23b6add02f2dfd7390431491a934aeec/lib/sequelize.js#L829)

javiertury:
Done. Finally I discarded the test about the warning message. The reason is that I can't spy on deprecation warning functions using sinon. If later we use errors for unsupported DB versions, I would be able to make tests again.

lib/utils/deprecations.js
```diff
@@ -9,3 +9,4 @@ exports.noTrueLogging = deprecate(noop, 'The logging-option should be either a f
 exports.noStringOperators = deprecate(noop, 'String based operators are deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/querying.html#operators', 'SEQUELIZE0003');
 exports.noBoolOperatorAliases = deprecate(noop, 'A boolean value was passed to options.operatorsAliases. This is a no-op with v5 and should be removed.', 'SEQUELIZE0004');
 exports.noDoubleNestedGroup = deprecate(noop, 'Passing a double nested nested array to `group` is unsupported and will be removed in v6.', 'SEQUELIZE0005');
+exports.noUnsupportedVersion = (dialect, version, minVer) => deprecate(noop, `(${dialect}) Database version ${version} is unsupported. Only ${minVer} and higher are supported.`, 'SEQUELIZE0006')();
```
SimonSchick: We can't use a single deprecation code fo for multiple versions/dialects.

lib/dialects/abstract/connection-manager.js
```diff
@@ -335,6 +345,15 @@ class ConnectionManager {
 
     return this.dialect.connectionManager.validate(connection);
   }
+
+  versionSupport() {
+    const version = _.get(this.sequelize, 'options.databaseVersion', 0);
+    const minVer = this.dialect.minimumSupportedVersion;
```
sushantdhiman: `this.dialect.supports.minimumSupportedVersion`?

lib/sequelize.js
```diff
@@ -181,6 +182,12 @@ class Sequelize {
       benchmark: false
     }, options || {});
 
+    // Coherce databaseVersion to valid semver format
+    if (this.options.databaseVersion) {
+      this.options.databaseVersion = semver.valid(semver.coerce(this.options.databaseVersion)) || 0;
+
```
sushantdhiman: Extra new line

lib/dialects/abstract/connection-manager.js
javiertury: My bad

javiertury:
I'm in a deadlock. On the one hand @SimonSchick, you are pushing for dialect/version specific warnings. On the other hand @sushantdhiman, you are pushing for an abstracted solution using `minimumSupportedVersion`.

I can do either

```
// lib/dialects/mssql/connection-manager.js
const { noMssqlBelow2012 } = require('../../utils/deprecations');

versionSupport() {
    const version = _.get(this.sequelize, 'options.databaseVersion', 0);

    if (semver.valid(version) && semver.lt(version, '11.0.0')) {
      noMssqlBelow2012();
    }
}
```

Or

```
// lib/dialects/abstract/connection-manager.js
const { noUnsupportedVersion } = require('../../utils/deprecations');

  versionSupport() {
    const version = _.get(this.sequelize, 'options.databaseVersion', 0);
    const minVer = this.dialect.supports.minimumSupportedVersion;

    if (minVer && semver.valid(version) && semver.lt(version, minVer)) {
      noUnsupportedVersion(this.dialectName, version, minVer);
    }
  }
```

Also, the more I look at `util.deprecate`, the more I think it's not the appropriate tool for this job. I think the purpose of `util.deprecate` and therefore `lib/utils/deprecations.js` is to deprecate API function calls[0]. Using `util.deprecate` to print warnings conditional on a remote response(DB version) is stretching too much.

[0]: Docs for [util.deprecate](https://nodejs.org/api/util.html#util_util_deprecate_fn_msg_code)

sushantdhiman:
Closing this PR in favor of https://github.com/sequelize/sequelize/pull/12218, Thanks for this work :)

PR #10705, opened 2019-04-05
fix(mariadb): named timezone connection bug

javiertury:
Mariadb doesn't support named timezones. Previous workaround missed `connectionConfig.timezone`. It didn't show on online tests, because they use UTC.

### DIFF
--- a/lib/dialects/mariadb/connection-manager.js
+++ b/lib/dialects/mariadb/connection-manager.js
@@ -54,13 +54,18 @@ class ConnectionManager extends AbstractConnectionManager {
    * @private
    */
   connect(config) {
+    // Named timezone is not supported in mariadb, convert to offset
+    let tzOffset = this.sequelize.options.timezone;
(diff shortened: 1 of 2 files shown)


rusher:
hmm, that PR doesn't really solve the issue, but create more problems. 
Offset is initialized with some offset, but for timezone with daylight saving time (DST) for example, that will result in having new offset when changing to summer/winter time for example. 

Last driver release now support [IANA](https://www.iana.org/time-zones) timezone. 

Better to revert that change

javiertury:
@rusher I think your points are valid. However this pull request didn't decide to convert named timezones to hour offsets. That was probably decided earlier in another commit. This pull request addresses how they are converted.

Previously the connection was initialized using raw `options.timezone`, and the timezone was immediately changed to an hour offset using `SET time_zone = '${tzOffset}';`. This pull request initializes the connection with an hour offset directly, so the connection doesn't crash. Before and after this commit, the final timezone has always been an hour offset. The difference is that the driver doesn't throw any errors.

PR #10707, opened 2019-04-05
docs(typescript): correct misspelling of `associations`

abernix:
### Description of change

This changes two occurrences of `associtations` (note the extra `t`).

I figure if this isn't corrected, this typo will slowly seep into the
various implementations (like end-users like myself) who are copying and
pasting examples from the documentation into their own projects. 😄

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -43,7 +43,7 @@ class User extends Model {
   // actively include a relation.
   public readonly projects?: Project[];
 
-  public static associtations: {
+  public static associations: {
     projects: Association<User, Project>;
   };
 }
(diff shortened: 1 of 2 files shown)


arnavb:
Oh wow! Looks like we both found the same error and opened a PR (mine is #10713). Since yours is more complete, I will close mine.

sushantdhiman:
Typing tests are breaking

abernix:
@sushantdhiman I've added a follow-up commit which I believe to be the correct adjustment in `docs-example.ts` (and proper fix for a previously incorrect reference which is now surfaced because of the typing which this PR corrects!).

Tests are passing now!

PR #10708, opened 2019-04-05
chore: more and better issue templates

eseliger:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,51 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+<!--
```
SimonSchick: Can we have something like `<!-- DO NOT IGNORE THIS ISSUE TEMPLATE OR YOUR ISSUE MAY/WILL BE CLOSED -->`.

Applies to other templates too.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,51 @@
...
+For general questions, please use StackOverflow or Slack.
+-->
+
+## What are you doing?
+<!-- Post a minimal, self-contained code sample that reproduces the issue, including models and associations -->
+
+```js
+// code here
+```
+**To Reproduce**
```
sushantdhiman: Invalid for sequelize context, there is no clicking, scrolling etc

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,51 @@
...
+_Output, either JSON or SQL_
+
+## Environment
+Dialect:
+- [ ] mysql
+- [ ] postgres
+- [ ] sqlite
+- [ ] mssql
+- [ ] any
+Dialect version: XXX
```
sushantdhiman: `Dialect library version`, many are often confused with this

.github/ISSUE_TEMPLATE/bug_report.md
sushantdhiman: There is already a note like that, it will always be ignored by some. To effectively enforce this we need to close any issue that is not following issues template (manually)

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,47 @@
...
+## What do you expect to happen?
+_I wanted Foo!_
+
+## What is actually happening?
+_But the output was bar!_
+
+_Output, either JSON or SQL_
+
+## Environment
+Dialect:
```
SimonSchick: Please also add `OS` and Node version.
Also in case of typescript errors we should ask for the typescript version as well.

PR #10709, opened 2019-04-05
feat(composition): add base composition support

javiertury:
This patch provides the base for building binded parameters. It defined the composition elements, their sequelize interface and the basic rules about processing them.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1052,7 +1052,7 @@ class QueryGenerator {
       }
 
       if (typeof item !== 'string') {
-        throw new sequelizeError.CompositionError(`Query item is not a slot or a string:\n${util.inspect(item)}`);
+        throw new sequelizeError.CompositionError('Item to be composed is not a slot or a string', item);
       }
 
       return item;
@@ -1077,7 +1077,7 @@ class QueryGenerator {
       }
 
       if (typeof item !== 'string') {
-        throw new sequelizeError.CompositionError(`Query item is not a slot or a string:\n${util.inspect(item)}`);
+        throw new sequelizeError.CompositionError('Query item to be composed is not a slot or a string', item);
       }
 
       return item;
@@ -2027,7 +2027,7 @@ class QueryGenerator {
 
       return value === 'NULL' ? `${key} IS NULL` : [key, value].join(` ${smth.comparator} `);
     }
-    if (smth instanceof Utils.Literal || smth instanceof Utils.Composition) {
+    if (smth instanceof Utils.Literal || smth instanceof Utils.CompositionMethod) {
       return smth.val;
     }
     if (smth instanceof Utils.Cast) {
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1041,7 +1041,6 @@ class QueryGenerator {
       composition = new Composition(composition);
     }
 
-    const result = {};
     const bind = [];
     const bindParam = this.bindParam(bind);
 
@@ -1059,10 +1058,8 @@ class QueryGenerator {
     });
 
     queryItems.push(';');
-    result.query = queryItems.join('');
 
-    result.bind = this.outputBind(bind);
-    return result;
+    return { query: queryItems.join(''), bind: this.outputBind(bind) };
   }
 
   composeString(composition) {
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -2024,8 +2024,11 @@ class QueryGenerator {
 
       return value === 'NULL' ? `${key} IS NULL` : [key, value].join(` ${smth.comparator} `);
     }
-    if (smth instanceof Utils.Literal || smth instanceof Utils.CompositionMethod) {
+    if (smth instanceof Utils.Literal) {
       return smth.val;
+    }
+    if (smth instanceof Utils.CompositionMethod) {
+      return Utils.convertComposable(smth, Composition);
     }
     if (smth instanceof Utils.Cast) {
       if (smth.val instanceof Utils.SequelizeMethod) {
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -19,6 +19,8 @@ const sequelizeError = require('../../errors');
 
 const QuoteHelper = require('./query-generator/helpers/quote');
 
+const { Slot, Composition } = require('./query-generator/composition');
+
 /**
  * Abstract Query Generator
  *
@@ -971,6 +973,16 @@ class QueryGenerator {
   }
 
   /*
+   * Converts an array of binded parameters to the format required by the dialect
+   *
+   * @param {array} array
+   * @private
+  */
+  outputBind(bind) {
+    return bind;
+  }
+
+  /*
     Returns a bind parameter representation of a value (e.g. a string, number or date)
     @private
   */
@@ -1022,6 +1034,54 @@ class QueryGenerator {
         throw error;
       }
     }
+  }
+
+  composeQuery(composition) {
+    if (!(composition instanceof Composition)) {
+      composition = new Composition(composition);
+    }
+
+    const result = {};
+    const bind = [];
+    const bindParam = this.bindParam(bind);
+
+    const queryItems = composition.items.map(item => {
+      if (item instanceof Slot) {
+        // Really binds, returns only acompaning sql string with bind placeholder
+        return this.format(item.value, item.field, item.options, bindParam);
+      }
+
+      if (typeof item !== 'string') {
+        throw new sequelizeError.CompositionError(`Query item is not a slot or a string:\n${util.inspect(item)}`);
+      }
+
+      return item;
+    });
+
+    queryItems.push(';');
+    result.query = queryItems.join('');
+
+    result.bind = this.outputBind(bind);
+    return result;
+  }
+
+  composeString(composition) {
+    if (!(composition instanceof Composition)) {
+      composition = new Composition(composition);
+    }
+
+    // For each query item, get sql parts and join them
+    return composition.items.map(item => {
+      if (item instanceof Slot) {
+        return this.escape(item.value, item.field, item.options);
+      }
+
+      if (typeof item !== 'string') {
+        throw new sequelizeError.CompositionError(`Query item is not a slot or a string:\n${util.inspect(item)}`);
+      }
+
+      return item;
+    }).join('');
   }
 
   isIdentifierQuoted(identifier) {
@@ -1967,7 +2027,7 @@ class QueryGenerator {
 
       return value === 'NULL' ? `${key} IS NULL` : [key, value].join(` ${smth.comparator} `);
     }
-    if (smth instanceof Utils.Literal) {
+    if (smth instanceof Utils.Literal || smth instanceof Utils.Composition) {
       return smth.val;
     }
     if (smth instanceof Utils.Cast) {
--- a/lib/dialects/abstract/query-generator/composition.js
+++ b/lib/dialects/abstract/query-generator/composition.js
@@ -1,0 +1,153 @@
+'use strict';
+
+const util = require('util');
+const { CompositionError } = require('../../../errors');
+
+/*
+ * Self contained value representation
+ *
+ * It can be converted to an escaped sql string or to a binded param. It also
+ * must be self contained, the escaper or binder should be part of the slot. It
+ * is identified for being an instace of this class, otherwise reject (reflects
+ * unsafe handling)
+ *
+ */
+class Slot {
+  constructor(value, field, options) {
+    this.value = value;
+    this.field = field;
+    this.options = options || {};
+  }
+}
+
+module.exports.Slot = Slot;
+
+/*
+ * It is a placeholder in a QueryItems instance
+ *
+ */
+class Placeholder {
+  constructor(name) {
+    this.name = name;
+  }
+}
+
+module.exports.Placeholder = Placeholder;
+
+/*
+  Buildind block for queries of any dialect
+
+  Should be simple, secure and composable. It can hold three items: (sql) strings,
+  value *slots* or placeholders. Keep this class simple, for higher abstractions
+  build something else. This should not depend on dialect or db version.
+
+  @private
+ */
+class Composition {
+  constructor(...items) {
+    // Should clone *items* array. Should not copy *items* array by reference.
+    this.items = this.constructor._processArgs(items);
+  }
+  get length() {
+    return this.items.length;
+  }
+  static _processArgs(items) {
+    // Should return new array
+    return items.reduce((a, item) => {
+      if (typeof item === 'string' || item instanceof Slot ||
+        item instanceof Placeholder) {
+        a.push(item);
+        return a;
+      }
+      if (item instanceof this) {
+        a.push(...item.items);
+        return a;
+      }
+
+      throw new CompositionError(`Invalid query item: ${util.inspect(item)}`);
+    }, []);
+  }
+  add(...items) {
+    this.items.push(...this.constructor._processArgs(items));
+
+    return this;
+  }
+  prepend(...items) {
+    this.items.unshift(...this.constructor._processArgs(items));
+
+    return this;
+  }
+  set(...items) {
+    this.items = this.constructor._processArgs(items);
+
+    return this;
+  }
+  clone() {
+    const clone = new this.constructor();
+    clone.items = this.items.slice(0);
+    return clone;
+  }
+  static from(array) {
+    const composition = new this();
+    composition.items = this._processArgs(array);
+
+    return composition;
+  }
+}
+
+module.exports.Composition = Composition;
+
+/*
+ * Holds multiple compositions or objects cohercible to compositions
+ *
+ * Useful for grouping and manipulating anonymous compositions. It
+ * holds compositions or single composition items.
+ *
+ * Concatenable methods.
+ */
+
+class CompositionGroup {
+  constructor(...args) {
+    this.compositions = args;
+  }
+  get length() {
+    return this.compositions.length;
+  }
+  add(...compositions) {
+    this.compositions.push(...compositions);
+
+    return this;
+  }
+  // Inserts a new composition(*spacer*) between existing compositions
+  space(spacer) {
+    if (this.compositions.length === 0) return this;
+
+    this.compositions = this.compositions.slice(1).reduce((a, group) => {
+      a.push(spacer, group);
+      return a;
+    }, [this.compositions[0]]);
+
+    return this;
+  }
+  merge(group) {
+    this.compositions = this.compositions.concat(group.compositions);
+
+    return this;
+  }
+  slice() {
+    const compositions = this.compositions.slice.apply(this.compositions, arguments);
+
+    return this.constructor.from(compositions);
+  }
+  toComposition() {
+    return Composition.from(this.compositions);
+  }
+  static from(array) {
+    const group = new this();
+    group.compositions = array.slice(0);
+
+    return group;
+  }
+}
+
+module.exports.CompositionGroup = CompositionGroup;
--- a/lib/dialects/abstract/query-generator/composition.js
+++ b/lib/dialects/abstract/query-generator/composition.js
@@ -1,7 +1,7 @@
 'use strict';
 
-const util = require('util');
 const { CompositionError } = require('../../../errors');
+const Utils = require('../../../utils');
 
 /*
  * Self contained value representation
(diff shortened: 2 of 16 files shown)


lib/sequelize.js
```diff
@@ -911,6 +919,53 @@ class Sequelize {
...
+   * that will be binded or escaped. Strings and slots will be joined without
+   * spaces.
+   *
+   * @param {...string|Object} items String or slot to be part of the query
+   * @memberof Sequelize
+   *
+   * @returns {Sequelize.composition}
+   */
+  static composition(...items) {
+    return new Utils.Composition(...items.map(item => {
```
SimonSchick: I've noticed you are performing a lot of array/argument destructuring, how performant is this?
Wouldn't it be better to pass arrays around?

SimonSchick: Also aren't you directly importing `Composition` already? Might as well use that instead of the `Utils` alias (assuming they are the same).

lib/sequelize.js
```diff
@@ -1192,6 +1247,8 @@ Sequelize.prototype.fn = Sequelize.fn;
 Sequelize.prototype.col = Sequelize.col;
 Sequelize.prototype.cast = Sequelize.cast;
 Sequelize.prototype.literal = Sequelize.literal;
+Sequelize.prototype.composition = Sequelize.composition;
```
SimonSchick: @sushantdhiman do we want to add even more of the aliases?
IIrc the plan was to make people import them directly via `const { literal } = require('sequelize')` in the future since they are effectively pure functions with no side effects.

lib/utils.js
```diff
@@ -452,6 +456,15 @@ class Literal extends SequelizeMethod {
 }
 exports.Literal = Literal;
 
+class Composition extends SequelizeMethod {
```
SimonSchick: This feels like a workaround, can we not have the base `Composition` class inherit from a sequelize method?

test/integration/raw-queries.test.js
```diff
@@ -0,0 +1,266 @@
+'use strict';
+
+const chai = require('chai'),
```
SimonSchick: Can you please make that 1 const declaration per line?
Also please destructure `chai` so we don't have single use/proxy imports :)

test/integration/raw-queries.test.js
```diff
@@ -0,0 +1,266 @@
...
+        return sequelize.query(query, {
+          type: sequelize.QueryTypes.INSERT
+        });
+      }).then(() => {
+        return sequelize.query(sequelize.composition('SELECT * FROM dummies WHERE id = ', sequelize.slot(20), ' AND name = \'abc$\';'), {
+          model: Model,
+          type: sequelize.QueryTypes.SELECT
+        });
+      }).then(instances => {
+        expect(instances.length).to.equal(1);
```
SimonSchick: Please use `expect(...).to.have.lengthOf(...)`, this show show a nicer diff on failure :)

test/unit/composition.test.js
```diff
@@ -0,0 +1,383 @@
+'use strict';
+
+const Support = require('../support'),
```
SimonSchick: See above.

test/unit/composition.test.js
```diff
@@ -0,0 +1,383 @@
...
+
+  it('should have *add* method', () => {
+    const items = ['foo ', new Slot(5), ' bar'];
+
+    const compositionEmpty = new Composition();
+    expect(compositionEmpty.add(...items)).to.deep.equal(new Composition(...items));
+
+    // *add* puts items towards the end
+    const composition = new Composition('start ');
+    expect(composition.add(...items).items).to.deep.equal(['start '].concat(items));
```
SimonSchick: Nit: `['start ', ...items]`

test/unit/composition.test.js
```diff
@@ -0,0 +1,383 @@
...
+    expect(group.space(', ').compositions).to.deep.equal(['foo ', ', ',
+      new Slot(5), ', ', ' bar', ', ', new Composition()]);
+
+    // *space* modifies instace without need to reassign
+    expect(group.compositions).to.deep.equal(['foo ', ', ',
+      new Slot(5), ', ', ' bar', ', ', new Composition()]);
+
+    // *space* returns same object
+    const group2 = CompositionGroup.from(compositions);
+    expect(group2 === group2.space(', ')).to.be.ok;
```
SimonSchick: Can we avoid using `.to.be.ok` assertions please?
They tend to hide diff's when they fail.

lib/dialects/abstract/query-generator.js
```diff
@@ -1024,6 +1036,54 @@ class QueryGenerator {
...
+    const bindParam = this.bindParam(bind);
+
+    const queryItems = composition.items.map(item => {
+      if (item instanceof Slot) {
+        // Really binds, returns only acompaning sql string with bind placeholder
+        return this.format(item.value, item.field, item.options, bindParam);
+      }
+
+      if (typeof item !== 'string') {
+        throw new sequelizeError.CompositionError(`Query item is not a slot or a string:\n${util.inspect(item)}`);
```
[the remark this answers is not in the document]
javiertury: This is a side effect of spliting the patch into smaller pull requests. Later, when `logger.support` is merged, I will do another commit.

lib/dialects/abstract/query-generator/query-proto.js
```diff
@@ -0,0 +1,160 @@
...
+  toComposition() {
+    throw new Error(`Class ${this.constructor.name} has not implemented toComposition method`);
+  }
+}
+
+module.exports.QueryProto = QueryProto;
+
+class SelectProto extends QueryProto {
+  static get partNames() {
+    return ['attributes', 'from', 'join', 'where', 'group',
```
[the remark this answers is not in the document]
javiertury: > readonly

I think that is only possible on typescript. A close approximation in vanilla javascript is this:

```
  static get partNames() {
    return Object.freeze(['attributes', ...]);
  }
```

test/unit/composition.test.js
javiertury: Done

test/integration/raw-queries.test.js
javiertury: Yes, totally missed that. Now it's solved.

test/integration/raw-queries.test.js
javiertury: Done

lib/dialects/abstract/query-generator/composition.js
```diff
@@ -0,0 +1,153 @@
...
+
+    return this;
+  }
+  merge(group) {
+    this.compositions = this.compositions.concat(group.compositions);
+
+    return this;
+  }
+  slice() {
+    const compositions = this.compositions.slice.apply(this.compositions, arguments);
```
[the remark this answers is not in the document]
javiertury: I did it for performance, I think passing through `arguments` is more efficient.

lib/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
javiertury: I agree. The `item` property in `CompositionError` was there, but I had forgot to use ti.

lib/dialects/abstract/query-generator/composition.js
```diff
@@ -0,0 +1,153 @@
...
+ */
+class Slot {
+  constructor(value, field, options) {
+    this.value = value;
+    this.field = field;
+    this.options = options || {};
+  }
+}
+
+module.exports.Slot = Slot;
```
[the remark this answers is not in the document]
javiertury: Done

lib/dialects/abstract/query-generator/composition.js
```diff
@@ -0,0 +1,153 @@
...
+  add(...compositions) {
+    this.compositions.push(...compositions);
+
+    return this;
+  }
+  // Inserts a new composition(*spacer*) between existing compositions
+  space(spacer) {
+    if (this.compositions.length === 0) return this;
+
+    this.compositions = this.compositions.slice(1).reduce((a, group) => {
```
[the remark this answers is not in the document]
javiertury: Done. I modified slightly your code, previously the for loop wrote `arr.length * 2` elements.

lib/sequelize.js
javiertury: The performance impact only affects `sequelize.query` calls, which behaves similar to `sequelize.literal`. It is not used internally, so any performance penalty is only paid by users that customize queries.

```
// This is the only example that came to my mind
// I believe it can be used for more powerfull things
const field = Model.rawAttributes.field;
Model.findAll({
  where: {
    field: {
      [Op.overlap]: sequelize.composition('int4range(', sequelize.slot(unsafeStart, field), sequelize.slot(unsafeEnd, field), ')')
    }
  },
});
```

lib/sequelize.js
javiertury: The goal of `sequelize.composition`, `sequelize.slot` is to give users functionality similar to `sequelize.literal`, but with support for binded parameters. The goal of `sequelize.placeholder` is to replace `Op.placeholder` in a later patch, which I think it is unsafe and doesn't work with compositions. So all in all, I think these aliases are necessary.

```
// This is the only example that came to my mind
// I believe it can be used for more powerfull things, like what people used literal for
const field = Model.rawAttributes.field;
Model.findAll({
  where: {
    field: {
      [Op.overlap]: sequelize.composition('int4range(', sequelize.slot(unsafeStart, field), sequelize.slot(unsafeEnd, field), ')')
    }
  },
});
```

lib/utils.js
javiertury: I've rename this to `CompositionMethod` to make the code clearer..

Regarding this question, I don't think `Composition` can inherit from `SequelizeMethod`. `SequelizeMethod` is an intermediate object that requires processing(`handleSequelizeMethod`) to be converted to a part of the query, which is a Composition.

javiertury:
I want to propose something. Before using `Composition` and a separate `Utils.Composition` (which I'v just renamed to `Utils.CompositionMethod` for clarity), I thought about another possibility.

I initially conceived `sequelize.composition` to return a plain `Composition` object. This required that many comparisons to `instanceof Utils.SequelizeMethod` needed to also compare against `instanceof Composition`. I discarded because I understood that `instanceof` was a quite expensive, it checks the whole prototype chain.

SimonSchick:
Just a headsup, my notebook died, won't be able to do any reviewing till tomorrow evening (gmt).

javiertury:
> Just a headsup, my notebook died, won't be able to do any reviewing till tomorrow evening (gmt).

I hope you don't lose any data. Don't worry about the pull request, I am still thinking about solving some of the issues you have risen.

lib/utils.js
```diff
@@ -658,6 +658,10 @@ exports.intersects = intersects;
  * @private
  */
 function interleave(arr, delimiter) {
+  if (!arr.length) {
```
SimonSchick: Oh, I should probably fix that in my code base 😅 
Also you should probably return `[]`.

test/unit/utils.test.js
```diff
@@ -250,6 +250,27 @@ describe(Support.getTestDialectTeaser('Utils'), () => {
     });
   });
 
+  describe('Interleave', () => {
+    it('interleaves',  () => {
+      expect(Utils.interleave(['a', 'b', 'c'], ',')).to.deep.equal(['a', ',', 'b', ',', 'c']);
+    });
+
+    it('Returns new array', () => {
```
SimonSchick: Nit: Since these strings are effectively used for building full test names you generally want to keep it lowercased, not capitalised.

test/unit/utils.test.js
```diff
@@ -250,6 +250,27 @@ describe(Support.getTestDialectTeaser('Utils'), () => {
...
+      const arr = ['a', 'b', 'c'];
+      expect(Utils.interleave(['a', 'b', 'c'], ',')).not.to.equal(arr);
```
SimonSchick: I think you meant to pass `arr` here.

lib/dialects/abstract/query-generator.js
```diff
@@ -1024,6 +1036,54 @@ class QueryGenerator {
...
+
+      if (typeof item !== 'string') {
+        throw new sequelizeError.CompositionError('Item to be composed is not a slot or a string', item);
+      }
+
+      return item;
+    });
+
+    queryItems.push(';');
+    result.query = queryItems.join('');
```
SimonSchick: Nit: you can do `return { query: ..., bind: ... }` and remove the declaration of `result`.

lib/dialects/abstract/query-generator/composition.js
```diff
@@ -0,0 +1,147 @@
...
+ */
+class Composition {
+  constructor(...items) {
+    // Should clone *items* array. Should not copy *items* array by reference.
+    this.items = this.constructor._processArgs(items);
+  }
+  get length() {
+    return this.items.length;
+  }
+  static _processArgs(items) {
```
SimonSchick: Note: This probably should be not static since you rely on inheritance here, eg. you use `this.constructor._processArgs` below.
There isn't much of a benefit to making it static.

lib/dialects/abstract/query-generator/query-proto.js
```diff
@@ -0,0 +1,157 @@
...
+    return clone;
+  }
+  toComposition() {
+    throw new Error(`Class ${this.constructor.name} has not implemented toComposition method`);
+  }
+}
+
+class SelectProto extends QueryProto {
+  static get partNames() {
+    return Object.freeze(['attributes', 'from', 'join', 'where', 'group',
```
SimonSchick: Not what I meant, my bad.

I was trying to suggest create the array outside of the class and pass a reference instead.

lib/dialects/abstract/query-generator/query-proto.js
```diff
@@ -0,0 +1,157 @@
...
+const { Composition } = require('./composition');
+
+/*
+ * Named collections of Composition elements.
+ *
+ * Useful for organizing and manipulating the different parts of a query.
+ * Allow disordered query building and easy substitution of query parts. Flexible,
+ * non-nested and non-exahustive organization.
+ */
+
```
SimonSchick: Nit: RM this line.

lib/dialects/abstract/query-generator/query-proto.js
```diff
@@ -0,0 +1,157 @@
...
+    const composition = new Composition();
+    composition.add('SELECT');
+
+    if (this.attributes.length) {
+      composition.add(' ', this.attributes);
+    } else {
+      composition.add(' *');
+    }
+
+    if (! (this.from && this.from.length)) {
```
SimonSchick: Nit: Spacing, I should do another lint thingy 🤔

lib/dialects/abstract/query-generator/query-proto.js
```diff
@@ -0,0 +1,157 @@
...
+    if (this.page.length) composition.add(' ', this.page);
+    if (this.lock.length) composition.add(' ', this.lock);
+
+    return composition;
+  }
+}
+
+class InsertProto extends QueryProto {
+  static get partNames() {
+    return Object.freeze(['preQuery', 'flags', 'table', 'ignoreDups', 'attributes',
```
SimonSchick: See above.

lib/dialects/abstract/query-generator/query-proto.js
```diff
@@ -0,0 +1,157 @@
...
+      'output', 'values', 'onConflict', 'return', 'postQuery']);
+  }
+  toComposition() {
+    if (!this.table || !this.table.length) {
+      throw new Error('Missing table in INSERT query');
```
SimonSchick: Shouldn't this throw the `CompositionError` you created?

lib/errors/composition-error.js
```diff
@@ -0,0 +1,27 @@
...
+
+const BaseError = require('./base-error');
+
+/**
+ * Thrown when a Composition encounters invalid elements.
+ *
+ * @param {Error}  error   Error description
+ * @param {*}      item    Invalid item
+ *
+ * @extends BaseError
```
SimonSchick: `@extends` is redundant with es6.

test/support.js
```diff
@@ -247,15 +248,22 @@ const Support = {
...
     } else {
       expect(query.query || query).to.equal(expectation);
-    }
 
-    if (assertions.bind) {
-      const bind = assertions.bind[Support.sequelize.dialect.name] || assertions.bind['default'] || assertions.bind;
-      expect(query.bind).to.deep.equal(bind);
+      if (assertions.bind) {
+        let bind = assertions.bind[Support.sequelize.dialect.name] || assertions.bind['default'] || assertions.bind;
+        if (Support.sequelize.dialect.name === 'mssql' && Array.isArray(bind)) {
```
SimonSchick: Note: I know this is a common (anti)pattern, but shouldn't we be checking `dialect.supports.*` instead?

test/unit/composition.test.js
```diff
@@ -0,0 +1,385 @@
...
+    const date = new Date(Date.UTC(2000, 0, 1));
+    expect(() => new Composition('Only allows strings, slots and placholders', date))
+      .to.throw('Invalid query item').with.property('item', date);
+    const object = {};
+    expect(() => new Composition('Otherwise, throw', object)).to.throw('Invalid query item').with.property('item', object);
+    const array = [];
+    expect(() => new Composition('bad', array)).to.throw('Invalid query item').with.property('item', array);
+  });
+
+  it('should have *set* method', () => {
```
SimonSchick: Nit: No need to highlighting method names.

lib/dialects/abstract/query-generator/query-proto.js
javiertury: I did that originally, but in a previous review I was told to use `static get fn() {}` instead

lib/dialects/abstract/query-generator/composition.js
javiertury: I usually make static all functions that don't expect an instance. ES6 classes inherit static methods, so inheritance should not be a problem.

lib/utils.js
javiertury: My bad

test/unit/utils.test.js
javiertury: You are right

test/unit/utils.test.js
javiertury: Yes, I forgot

lib/dialects/abstract/query-generator.js
javiertury: Done

lib/dialects/abstract/query-generator/query-proto.js
javiertury: Or perhaps you mean to define the array outside?
```
const selectProtoPartNames = Object.freeze([...]);

class SelectProto extends QueryProto {
  static get partNames() {
    return selectProtoPartNames;
  }
}
```

I thought that JS engines were allowed to cache static getters, but it can also be done this way.

lib/dialects/abstract/query-generator/query-proto.js
javiertury: Yes, now it uses it

test/unit/composition.test.js
javiertury: Asterisks removed

test/support.js
javiertury: I changed my mind, for simplicity I will call `outputBind` for every dialect, which should translate regular bind arrays to specific dialect formats(if any).

lib/sequelize.js
javiertury: Take a look at the new review with `composableMixin`, it should be more performant. Now, the classes are `Composition = Object  + composableMixin` and `CompositionMethod = SequelizeMethod + composableMixin`.

I've implemented a little helper `Utils.converComposable` which converts one composable class into another composable class with performance in mind.

Now `sequelize.composition` looks like this:

```
static composition(...items) {
    return Utils.Composition.from(items);
}
```

I could increase performance using `arguments` instead. I'm curious, which is the reason for not using `arguments`?

lib/dialects/abstract/query-generator/composition.js
SimonSchick: The point is that you are effectively using a static method like an instance method, it doesn't really matter that there is no `this` context (since `this` would just be the class itself, I'd just ask you to make it non static :)

lib/dialects/abstract/query-generator/query-proto.js
SimonSchick: I don't think getters can be cached at all, since it needs to return dynamic values, I'd just do `SelectProto.partNames = Object.freeze(...)`

SimonSchick:
Is any of this intended to be public API, if yes, could you please add corresponding typings?

javiertury:
@SimonSchick, I've tried my best. (I have zero experience with typescript)

eseliger:
On first view looks good, but will compare with the js in depth again. Thanks for the effort!

PR #10716, opened 2019-04-07
fix(bind): add escsape of null character for postgres bind parameters

holm:
### Pull Request check-list

### Description of change

Postgres doesn't support null characters in strings. This has generally been handled via https://github.com/sequelize/sequelize/blob/master/lib/sql-string.js#L75. However with Sequelize 5 using bind parameters for insert and update, the replacement of null characters no longer happens. This PR adds replacement of null characters in case of bind parameters being used.

I have added tests, but have not been able to run them across the other database. I think it should pass though, otherwise I would appreciate help getting them fixed.

### DIFF
--- a/lib/dialects/postgres/query.js
+++ b/lib/dialects/postgres/query.js
@@ -20,11 +20,14 @@ class Query extends AbstractQuery {
    * @private
    */
   static formatBindParameters(sql, values, dialect) {
-    let bindParam = [];
+    const stringReplaceFunc = value => typeof value === 'string' ? value.replace(/\0/g, '\\0') : value;
+
+    let bindParam;
     if (Array.isArray(values)) {
-      bindParam = values;
+      bindParam = values.map(stringReplaceFunc);
       sql = AbstractQuery.formatBindParameters(sql, values, dialect, { skipValueReplace: true })[0];
     } else {
+      bindParam = [];
       let i = 0;
       const seen = {};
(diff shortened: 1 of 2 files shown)


lib/dialects/postgres/query.js
```diff
@@ -20,11 +20,20 @@ class Query extends AbstractQuery {
    * @private
    */
   static formatBindParameters(sql, values, dialect) {
-    let bindParam = [];
+    const stringReplaceFunc = value => {
+      if (_.isString(value)) {
```
SimonSchick: Nit: You can make this `value => typeof value === 'string' ? value.replace(...) : value`

holm: I thought that wasn’t a complete check due to the issues explained at https://stackoverflow.com/questions/4059147/check-if-a-variable-is-a-string-in-javascript

SimonSchick: Most of the codebase already uses typeof checks, https://stackoverflow.com/questions/5750656/whats-the-point-of-new-stringx-in-javascript explains the debacle pretty well.

In the end, sequelize doesn't really support these boxed types and I don't think it should as it would increase complexity in hundreds of places.

holm:
I addressed the comment now. The build failure doesn't seem related to my changes.

holm:
Thanks for the quick merge and release

Issue #10399, opened 2019-01-29
TypeError: include.attributes.map is not a function


aaronfulkerson:
## What are you doing?

https://gist.github.com/aaronfulkerson/ea67f28369c13fbb4267e365b0f879fa

## What do you expect to happen?
In the Gist, my ticket model has a scope on it called 'show' which I use when I display individual tickets. It works fine except when I begin trying to get the users associated with a discussion.
I should be able to get the users of a discussion since I have created a belongsToMany association through UserDiscussions. The code in the Gist works fine on version 4.42.0!

## What is actually happening?
 I can create a ticket and associate users to a discussion but when I try to retrieve them along with a post, I get the error in the title: "TypeError: include.attributes.map is not a function"

__Dialect:__ postgres
__Dialect version:__ XXX
__Database version:__ 11
__Sequelize version:__ 5.0.0-beta.16
__Tested with latest release:__ No (If yes, specify that version)

**Note :** _Your issue may be ignored OR closed by maintainers if it's not tested against latest version OR does not follow issue template._

SimonSchick:
Could you also post the stacktrace for good measure?

contra:
Can confirm this is happening on the latest anywhere a scope contains `excludes` and `includes` keys instead of a plain array.

aaronfulkerson:
Yeah, thanks to @contra I figured out what's going on. On my User model the default scope excludes the password digest and if I remove that it "works" fine.

contra:
@aaronfulkerson You can use the internal function to pre-prepare the attributes as a temporary workaround.

```js
const model = Model.scope(scope)
model._expandAttributes(query) // make the attribute option valid
```

(`_expandAttributes` mutates `query.attributes`)

aaronfulkerson:
@contra fantastic tip. Thank you very much.

yoann217:
same probleme here on 5.1.0 with mysql

rlindgren:
Same issue as @yoann217 with 5.1.0 where `attributes.exclude` is defined on the `defaultScope`

weyert:
Yes, I have the same issue with Postgres

ALiangLiang:
Same issue on MySQL.....

SimonSchick:
Can one of you please post a full repro, I tried to repro this locally just now but could not cause that error.

SimonSchick:
fyi @sushantdhiman #9735 broke this (verified by bisect).

SimonSchick:
[...]

PR #10718, opened 2019-04-07
fix(query-generator): handle field names in scope exclude…

amitaymolko:
### Pull Request check-list

### Description of change

Additional fix to issue: https://github.com/sequelize/sequelize/issues/10399
Adds fix for using attribute field names

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1426,6 +1426,8 @@ class QueryGenerator {
     // includeIgnoreAttributes is used by aggregate functions
     if (topLevelInfo.options.includeIgnoreAttributes !== false) {
       include.model._expandAttributes(include);
+      Utils.mapOptionFieldNames(include, include.model);
+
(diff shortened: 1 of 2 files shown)


eseliger:
could you explain what didn't work before with the field names?

amitaymolko:
If you had default scope with includes that exclude attributes with a cutom field name it didn't work.
The fix provided here: https://github.com/sequelize/sequelize/commit/e07a7beae2be62fb4ac586f844974dbd0a343a3d
Didn't handle the case for custom attribute names.

You can see the test changes in the commit.
```
this.DefaultScopeExclude = this.sequelize.define('DefaultScopeExclude', {
        name: Sequelize.STRING,
        other_value: {
          type: Sequelize.STRING,
          field: 'otherValue'
        }
      }
```
if the model was defined with a `field` name like `other_value` in the test, it would try to access the column as `other_value` instead of `otherValue` and would throw an error

Cheers

amitaymolko:
Thanks for the quick release :)

PR #10722, opened 2019-04-08
fix(model): don't alter original scopes when combining them

### LINKED ISSUE #10696: defaultScope mutates after applying .scope(['defaultScope', 'otherScope'])
## What are you doing?

With the code below, the default scope of the model seems to be overridden after doing `Model.scope(['defaultScope', 'withBar'])`, leading to unexpected results.

```js
const config = require('./config/config.json')
const sequelize = new Sequelize(config['development'])

    name: Sequelize.STRING,
    foo: Sequelize.BOOLEAN,
    bar: Sequelize.BOOLEAN,
    baz: Sequelize.BOOLEAN
  }, {});

  User.addScope('defaultScope', {
    where: { foo: true }
  }, { override: true })

  User.addScope('withBar', {
    where: { bar: true }
  })

  User.addScope('noBaz', {
    where: { baz: false }
  })

const users = [
  {
    name: 'Alice',
    foo: true,
    bar: true,
    baz: true
  },
  {
    name: 'Bob',
    foo: true,
    bar: true,
    baz: false,
  },
  {
    name: 'Charlie',
    foo: true,
    bar: false,
    baz: false,
  },
  {
    name: 'Danika',
    foo: false,
    bar: false,
    baz: false
  }
]

async function doStuff() {
  await sequelize.sync({force: true})
  await User.bulkCreate(users)

  const defaultscope = await User
    .findAll()
    .then(result => result.map(u => u.name))

  const withBar = await User
    .scope('withBar')
    .findAll()
    .then(result => result.map(u => u.name))

  const withBarAndDefault = await User
    .scope(['defaultScope', 'withBar'])
    .findAll()
    .then(result => result.map(u => u.name))

  const defaultscope2 = await User
    .findAll()
    .then(result => result.map(u => u.name))

  const noBaz = await User
    .scope('noBaz')
    .findAll()
    .then(result => result.map(u => u.name))

  const unscoped = await User
    .unscoped()
    .findAll()
    .then(result => result.map(u => u.name))

  const defaultscope3 = await User
    .findAll()
    .then(result => result.map(u => u.name))

  console.log('defaultscope: ', defaultscope)
  console.log('withBar: ', withBar)
  console.log('withBarAndDefault: ', withBarAndDefault)
  console.log('defaultscope2: ', defaultscope2)
  console.log('noBaz: ', noBaz)
  console.log('unscoped: ', unscoped)
  console.log('defaultscope3: ', defaultscope3)

  process.exit(0)
}

doStuff()
```

According to the docs `.scope()` should return a new model with the scope applied, so multiple invocations of `User.scope().findAll()` should have no bearing on each other.

Expected results:
```
defaultscope:  [ 'Alice', 'Bob', 'Charlie' ]
withBar:  [ 'Alice', 'Bob' ]
withBarAndDefault:  [ 'Alice', 'Bob' ]
defaultscope2:  [ 'Alice', 'Bob', 'Charlie' ]
noBaz:  [ 'Bob', 'Charlie', 'Danika' ]
unscoped:  [ 'Alice', 'Bob', 'Charlie', 'Danika' ]
defaultscope3:  [ 'Alice', 'Bob', 'Charlie' ]
```

Instead, later invocations of `User.findAll()` _were_ affected.

Actual results:
```
// correct
defaultscope:  [ 'Alice', 'Bob', 'Charlie' ] 

// correct
withBar:  [ 'Alice', 'Bob' ]

// correct
withBarAndDefault:  [ 'Alice', 'Bob' ] 

// should include 'Charlie'
defaultscope2:  [ 'Alice', 'Bob' ]

 // correct
noBaz:  [ 'Bob', 'Charlie', 'Danika' ]

// correct
unscoped:  [ 'Alice', 'Bob', 'Charlie', 'Danika' ]

// should include 'Charlie'
defaultscope3:  [ 'Alice', 'Bob' ] 
```

__Dialect:__ mysql
__Dialect version:__ XXX  (where do I find this?)
__Database version:__ mysql  Ver 15.1 Distrib 10.3.13-MariaDB, for Linux (x86_64) using readline 5.1
__Sequelize version:__ 5.2.11
__Tested with latest release:__ Yes, 5.2.11 (updated today)

jorgelf:
### Pull Request check-list

### Description of change

Model was previously altering the original scopes when merging them for queries.
This happened because [Model._mergeFunction](https://github.com/sequelize/sequelize/blob/b6c9117f279d3df19565797ae3fc1be0a1a5eda9/lib/model.js#L793) was in some cases returning references to attributes of the original scope objects instead of cloning them (see [line 811](https://github.com/sequelize/sequelize/blob/b6c9117f279d3df19565797ae3fc1be0a1a5eda9/lib/model.js#L811)).

Also added a new test to prevent this problem from happening again.

Closes #10696

### Additional notes

Tried to clone the object with both [Utils.cloneDeep](https://github.com/sequelize/sequelize/blob/b6c9117f279d3df19565797ae3fc1be0a1a5eda9/lib/utils.js#L126) and Lodash's `_.cloneDeep`, but both result in _JavaScript heap out of memory_ errors when running the tests. My guess is there are some cases where the scopes can include references to complex objects that should not be cloned (maybe references to _Sequelize_ itself?). Because of this, it's cloned with a call to Lodash's `_.cloneDeepWith` specifying a function that results in cloning only plain objects. If this function for deep cloning only plain objects can be useful in other places, maybe it could be moved to [Utils](https://github.com/sequelize/sequelize/tree/master/lib/utils.js).

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -808,7 +808,12 @@ class Model {
         }
       });
     }
-    return srcValue === undefined ? objValue : srcValue;
+    return srcValue === undefined ? objValue : _.cloneDeepWith(srcValue, elem => {
+      // Only clone plain objects (cloning everything can lead to memory heap problems)
+      if (typeof elem !== 'object' || !_.isPlainObject(elem)) {
+        return elem;
+      }
+    });
   }
 
   static _assignOptions(...args) {
--- a/lib/model.js
+++ b/lib/model.js
@@ -808,12 +808,13 @@ class Model {
         }
       });
     }
-    return srcValue === undefined ? objValue : _.cloneDeepWith(srcValue, elem => {
-      // Only clone plain objects (cloning everything can lead to memory heap problems)
-      if (typeof elem !== 'object' || !_.isPlainObject(elem)) {
-        return elem;
-      }
-    });
+    // If we have a possible object/array to clone, we try it.
+    // Otherwise, we return the original value when it's not undefined,
+    // or the resulting object in that case.
+    if (srcValue) {
+      return Utils.cloneDeep(srcValue, true);
+    }
+    return srcValue === undefined ? objValue : srcValue;
   }
 
   static _assignOptions(...args) {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -123,7 +123,7 @@ function formatNamedParameters(sql, parameters, dialect) {
 }
 exports.formatNamedParameters = formatNamedParameters;
 
-function cloneDeep(obj) {
+function cloneDeep(obj, onlyPlain) {
   obj = obj || {};
   return _.cloneDeepWith(obj, elem => {
     // Do not try to customize cloning of arrays or POJOs
@@ -131,8 +131,9 @@ function cloneDeep(obj) {
       return undefined;
     }
 
-    // Don't clone stuff that's an object, but not a plain one - fx example sequelize models and instances
-    if (typeof elem === 'object') {
+    // If we specified to clone only plain objects & arrays, we ignore everyhing else
+    // In any case, don't clone stuff that's an object, but not a plain one - fx example sequelize models and instances
+    if (onlyPlain || typeof elem === 'object') {
       return elem;
     }
 
--- a/test/unit/model/scope.test.js
+++ b/test/unit/model/scope.test.js
@@ -112,6 +112,10 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       it('should unite attributes with array', () => {
         expect(User.scope('aScope', 'defaultScope')._scope.attributes).to.deep.equal({ exclude: ['value', 'password'] });
       });
+
+      it('should not modify the original scopes when merging them', () => {
+        expect(User.scope('defaultScope', 'aScope').options.defaultScope.attributes).to.deep.equal({ exclude: ['password'] });
+      });
     });
 
     it('defaultScope should be an empty object if not overridden', () => {


papb:
@jorgelf - Nice. Thanks also for digging into the `clone` stuff. Looks like we have three different types of cloning around Sequelize now, though. I've seen direct usage of `_.clone` and also [Utils.cloneDeep](https://github.com/sequelize/sequelize/blob/b6c9117f279d3df19565797ae3fc1be0a1a5eda9/lib/utils.js#L126), and now we have this "third" way of cloning which you introduced. I'm not saying that you shouldn't have done that, on the contrary, you've investigated and chosen what seemed to be the best solution. Nice work. What I'd like to ask, since you've already digged into this, is: could you investigate the other usages of `_.clone` and `Utils.cloneDeep` around Sequelize to see if you could refactor those to have a standard way to clone things, making the code easier to understand? This would be work for another PR, of course, this PR is great as it is :grimacing:

jorgelf:
You are probably right, too many flavours of cloning. Maybe this new cloning method can be merged with the already existing `Utils.cloneDeep` by adding a new optional parameter?
```javascript
function cloneDeep(obj, onlyPlain) {
  obj = obj || {};
  return _.cloneDeepWith(obj, elem => {
    // Do not try to customize cloning of arrays or POJOs
    if (Array.isArray(elem) || _.isPlainObject(elem)) {
      return undefined;
    }

    // If we specified to clone only plain objects & arrays, we ignore everyhing else
    // In any case, don't clone stuff that's an object, but not a plain one - fx example sequelize models and instances
    if (onlyPlain || typeof elem === 'object') {
      return elem;
    }

    if (elem && typeof elem.clone === 'function') {
      return elem.clone();
    }
  });
}
```
Tests are still passing after this change, so nothing seems to break with it. If it looks fine to you, I can just commit those changes to this branch, or make a new PR if you prefer it that way.

I probably wouldn't do any changes regarding usage of Lodash's `_.clone`. It's a simple, straightforward way of making shallow copies of any element. Merging it with the deep cloning method may lead to confusion. Maybe it could have a wrapper inside `Utils`, but I'm not sure it's actually worth it.

And about the Travis build fail, those tests don't give me any trouble in my machine, so not sure why they failed.

papb:
@jorgelf - Sorry to take long. I fully agree with you. Merging with the already existing `cloneDeep` is excellent, and leaving `_.clone` untouched is good too.

And don't worry about the travis build fail, it happens.

jorgelf:
Great, thanks for the comment, @papb.

A new commit has been added doing that small refactoring. If something doesn't look good to you, just tell me.

PR #10739, opened 2019-04-09
Some v6 suggestions

SimonSchick:
### Pull Request check-list

### Description of change

This is a small selection of non breaking changes that will are related to upgrading to node v6 which is going EOL end of this month.

The ONLY breaking change here should be that we are dropping node v6.

I will probably create another PR for more breaking changes in v6 @eseliger and I are looking into potentially dropping bluebird.

### DIFF
--- a/docs/upgrade-to-v6.md
+++ b/docs/upgrade-to-v6.md
@@ -1,0 +1,9 @@
+# Upgrade to v5
+
+Sequelize v5 is the next major release after v4
+
+## Breaking Changes
+
+### Support for Node 8 and up
+
+Sequelize v6 will only support Node 8 and up
--- a/docs/upgrade-to-v6.md
+++ b/docs/upgrade-to-v6.md
@@ -1,6 +1,6 @@
-# Upgrade to v5
+# Upgrade to v6
 
-Sequelize v5 is the next major release after v4
+Sequelize v6 is the next major release after v4
 
 ## Breaking Changes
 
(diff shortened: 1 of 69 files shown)


lib/dialects/abstract/query-generator.js
```diff
@@ -489,9 +483,7 @@ class QueryGenerator {
       - rawTablename, the name of the table, without schema. Used to create the name of the index
    @private
   */
-  addIndexQuery(tableName, attributes, options, rawTablename) {
-    options = options || {};
-
+  addIndexQuery(tableName, attributes, options = {}, rawTablename) {
```
frlinw: Just a suggestion, it's better if optionals parameters are explicits: `rawTablename = null`

lib/model.js
```diff
@@ -269,7 +269,7 @@ class Model {
       };
     }
 
-    const existingAttributes = _.clone(this.rawAttributes);
+    const existingAttributes = { ...this.rawAttributes };
```
frlinw: Is a clone really required here ?

SimonSchick: Idk, trying not to change logic.

SimonSchick:
Revert function defaults for now as it has too many implicit breaking changes.

SimonSchick:
@frlinw @eseliger Unfortunately I cannot test postgres locally atm since my dev env died and docker doesn't want to work on my current one, if one of you could figure out what exactly broke here that'd be great!

SimonSchick:
@sushantdhiman this now builds, I would really appreciate some feedback on this so we can move forward for v6 🙂

docs/upgrade-to-v6.md
```diff
@@ -0,0 +1,9 @@
+# Upgrade to v5
```
eseliger: v6

SimonSchick: aaaaaa

PR #10753, opened 2019-04-11
fix(querying): swap startWith and endsWith operators

### LINKED ISSUE #10743: startsWith and endsWith operators are reversed
## What are you doing?

https://github.com/sequelize/sequelize/pull/9999 introduced the `startsWith` and `endsWith` operators, but they seem to be reversed:

image

`[Op.startsWith]: 'hat'` should be defined as `hat%`, i.e. the string begins with `hat` and then has anything afterwards.

`[Op.endsWith]: 'hat'` should be defined as `%hat`, i.e. the string begins with anything but ends with `hat`.

`startsWith` and `endWith` are the reverse of what I expect. The [unit tests for this feature](https://github.com/sequelize/sequelize/blob/3dbb318ed6a603bd87284c8edfceabe760e9baf8/test/unit/sql/where.test.js#L437-L453) exhibit the same behavior.

__Dialect:__  any
__Dialect version:__ any
__Database version:__ any
__Sequelize version:__ master

SimonSchick:
### Pull Request check-list

### Description of change

Fixes #10743

### DIFF
--- a/docs/querying.md
+++ b/docs/querying.md
@@ -162,8 +162,8 @@ const Op = Sequelize.Op
 [Op.notLike]: '%hat'       // NOT LIKE '%hat'
 [Op.iLike]: '%hat'         // ILIKE '%hat' (case insensitive) (PG only)
 [Op.notILike]: '%hat'      // NOT ILIKE '%hat'  (PG only)
-[Op.startsWith]: 'hat'     // LIKE '%hat'
-[Op.endsWith]: 'hat'       // LIKE 'hat%'
+[Op.startsWith]: 'hat'     // LIKE 'hat%'
+[Op.endsWith]: 'hat'       // LIKE '%hat'
 [Op.substring]: 'hat'      // LIKE '%hat%'
 [Op.regexp]: '^[h|a|t]'    // REGEXP/~ '^[h|a|t]' (MySQL/PG only)
 [Op.notRegexp]: '^[h|a|t]' // NOT REGEXP/!~ '^[h|a|t]' (MySQL/PG only)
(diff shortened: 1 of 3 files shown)


gabegorelick:
I agree that the current behavior of `startsWith` and `endsWith` is very wrong, but either no one is using it, or people are using it and working around that they do the opposite of what they should. If it's the latter, this change will break consumers of Sequelize. If it's the former, then arguably these aliases should just be removed.

SimonSchick:
@gabegorelick I disagree here, it's obviously unintended behavior eg. a bug, most people that use this functionality probably haven't even noticed this isn't working as intended.
By the same logic *any* bug that has unintended side effects should be considered a feature.

I will leave final judgement to @sushantdhiman

gabegorelick:
> most people that use this functionality probably haven't even noticed this isn't working as intended.

I guess I hadn't considered this, because that's a very scary thought. 

> By the same logic any bug that has unintended side effects should be considered a feature.

I don't wish to rehash [common debates](https://xkcd.com/1172/) around [semantic versioning](https://gist.github.com/jashkenas/cbd2b088e20279ae2c8e) here, so let me just give an example.  

Let's say that 0.00001% of the time, `Model.findOne` actually returned an array of model instances. I would agree that this is a bug that could be fixed in a minor version since. But what happens if that rate was 100%? Clearly, `Model.findOne` would still be *wrong* in such a case. But undeniably, users would have lots of code that depended on the output. Is it OK to break them? Probably not.

It's not cut and dry, so I don't object to merging this, but we should at least understand the implications and be prepared for people to complain when their code stops working. Again, that probably won't happen in this case since no one's complained about `startsWith` being incorrect. So that's more reason to merge this.

gabegorelick:
Sequelize also tends to "move fast and breaks things", which is arguably a good stance to take when most consumers are applications that will be pinned to a specific version anyway. 

Anyway, sorry to turn this into a larger thing. I'm fine merging this.

Issue #9421, opened 2018-05-09
Index hints


sushantdhiman:

jwt1143:
Okay, I think a good starting point might be to go over SQL capabilities, and how/if we want to implement all of these.

Valid MySQL queries:
1. `SELECT * FROM t1 USE INDEX ();` Without using indices
2. `SELECT * FROM t1 USE INDEX (i1);` With single index
3. `SELECT * FROM t1 USE INDEX (i1,i2);` With multiple indices
4. `SELECT * FROM t1 USE INDEX (i1) USE INDEX (i2);` With multiple USE INDEX

I think 1, 2, and 3 are all reasonable functionality. I don't see any reason why we would need to implement 4. As far as I know, 4 is simply a syntactically different way to represent 3, without adding any new functionality (I could be wrong).

**Note:** It is perfectly valid to use the same index more than once in an index hint.
For example:
`SELECT * FROM t1 USE INDEX (i1) USE INDEX (i1);`
and
`SELECT * FROM t1 USE INDEX (i1,i1);`
are both valid statement so we don't need to enforce index uniqueness.

jwt1143:
Also, it looks like all 3 Pull Requests referenced in the original post are for MySQL. I don't there is any actual demand for this feature outside of the MySQL dialect; at least nothing that I have been able to find online.

henryarbolaez:
Agree with you @jwt1143 on my implementation I was able to pass a collection of indexes and be like option `1`, `2` and `3`

`SELECT * FROM Order USE INDEX(idx_status_id) WHERE status_id = 1`

henryarbolaez:
Also, I like to support the ability to use `FORCE INDEX` sometimes is useful but rare.

ezequielzacca:
is not rare at all for very large databases

sushantdhiman:
@jwt1143 Your proposal looks good, some comments

`type` should be an constant like `query hints`

```js
indexHints: [
    { type: IndexHints.USE || FORCE || IGNORE, values: ["i1","i2"] },
]
```

I also do perfer object based approach

```js
indexHints: [
    { type: IndexHints.USE, values: ["i1","i2"] },
]
```

As for implementation

**MSSQL**

We are only going to introduce this

```sql
WITH (TableHints, INDEX (<index>))
```

[...]

PR #10756, opened 2019-04-12
feat(query-generator): ability to use indexHints option for mysql

jonmagic:
### Pull Request check-list

### Description of change

Per https://github.com/sequelize/sequelize/issues/9421 this takes the proposed `indexHints` API and implements it for mysql.

```js
Project.findAll({
  indexHints: [
    { type: IndexHints.USE, values: ['index_project_on_name'] }
  ],
  where: {
    id: {
      [Op.gt]: 623
    },
    name: {
      [Op.like]: 'Foo %'
    }
  }
})
```

Will generate a mysql query that looks like this:

```sql
SELECT * FROM Project USE INDEX (index_project_on_name) WHERE name LIKE 'FOO %' AND id > 623;
```


lib/index-hints.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+/**
+ * An enum of index hints to be used in mysql for querying with index hints
+ *
+ * @property USE
+ * @property FORCE
+ * @property IGNORE
+ */
+const IndexHints = module.exports = { // eslint-disable-line
```
SimonSchick: Please use `exports.IndexHints = { ... }`

lib/dialects/mysql/query-generator.js
```diff
@@ -434,6 +435,24 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
     return result;
   }
 
+  selectFromTableFragment(options, model, attributes, tables, mainTableAs) {
+    let fragment = `SELECT ${attributes.join(', ')} FROM ${tables}`;
```
SimonSchick: Just to sanity check, do we need to escape anything here?

jonmagic: This is taken directly from abstract/query-generator.js so I don't believe so?

lib/index-hints.js
jonmagic: Updated in 37772bcb0c0a7bc9eb1982466ce0f9a4dbe341f0

lib/dialects/mysql/query-generator.js
```diff
@@ -434,6 +435,24 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
     return result;
   }
 
+  selectFromTableFragment(options, model, attributes, tables, mainTableAs) {
```
sushantdhiman: No need to override this you can implement this with `dialect.supports`.

1. First start with MySQL, define `indexHints: true` someone where [here](https://github.com/sequelize/sequelize/blob/2c18e084509469c7e8b9fdbabcb4c713312c216d/lib/dialects/mysql/index.js#L43)
2. Along with `options.indexHints` you will now also check for `this._dialect.supports.indexHints`

jonmagic: Done in bbd3c2d5dd5aced953bb6b50ca00f2ec8191ba7c

lib/dialects/abstract/query-generator.js
```diff
@@ -1906,6 +1907,14 @@ class QueryGenerator {
       fragment += ` AS ${mainTableAs}`;
     }
 
+    if (options.indexHints && this._dialect.supports.indexHints) {
+      for (const hint of options.indexHints) {
+        if (IndexHints[hint.type]) {
+          fragment += ` ${IndexHints[hint.type]} INDEX (${hint.values.join(',')})`;
```
sushantdhiman: You will need to quote index names from `values`

```js
hint.values.map(indexName => this.quoteIdentifiers(indexName)).join(',')
```

jonmagic: Updated in 65c91b178d1f5427489ac458298cd2557d2c135e

PR #10778, opened 2019-04-16
change count type from number to string

### LINKED ISSUE #10757: model.count doesn't always return a Promise<number>
https://github.com/sequelize/sequelize/blob/728707cda2de842439130a6e7a20be0e36c8a91b/types/lib/model.d.ts#L1751

if you use group/attributes it will return an array of objects

mjdavidson:
fixes #10757

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -239,36 +239,36 @@ export interface WhereOperators {
 
   /**
    * MySQL/PG only
-   * 
+   *
    * Matches regular expression, case sensitive
-   * 
+   *
    * Example: `[Op.regexp]: '^[h|a|t]'` becomes `REGEXP/~ '^[h|a|t]'`
    */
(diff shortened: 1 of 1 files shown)


SimonSchick:
I am not sure what do here exactly, it should *never* return a string, this is a bug and the typings shouldn't really reflect that. @sushantdhiman any ideas on how to fix this?

mjdavidson:
@SimonSchick so is this a deeper routed problem

SimonSchick:
Correct, for now we can make this `number | string` so people will have to ensure it's actually a number they will process, so this will be non-breaking once it's fixed.

eseliger:
I checked and could not find a case where this happens
Could you provide a repro?

papb:
Apparently this is [no longer an issue](https://github.com/sequelize/sequelize/issues/10757#issuecomment-515265638).

Issue #7930, opened 2017-07-11
Conversion failed when converting date and/or time from character string.


epozsh:
on sequelize version 4.3.0, i get this error 
cause when i give new Date() sequelize pass this format

> 2014-04-11 14:38:44.000 +00:00 

on sequelize version 3.27.0 i do not get error
cause it pass it like this 
> 2014-04-11 14:38:44.000

How i remove +00:00? so i can use newer version?

cladjules:
@Shadowman4205 that may look ugly but, in sequelize/lib/data-types.js, you can replace the method:

`DATE.prototype._stringify = function _stringify(date, options) {
  date = this._applyTimezone(date, options);

  // Z here means current timezone, _not_ UTC
  // return date.format('YYYY-MM-DD HH:mm:ss.SSS Z');
  return date.format('YYYY-MM-DD HH:mm:ss.SSS');
};`

Shadowman4205:
@cladjules  Thank you for this solution for now.

carlosscheffer:
I'm facing the same problem, however it's within the ambit of comparison:
` WHERE [PedidoVenda].[DATA_ALTERACAO] > N'2017-11-01 11:14:03.267 +00:00'`

thats returns me 

> DatabaseError: Conversion failed when converting date and/or time from character string.

jnreynoso:
+1

andersao:
+1

balazsnemeth:
+1 -> Same issue with dateformat "2018-02-08T05:00:00.000Z" on version 4.32.6

spyter:
+1

The date value below is a result of setting processed = new Date();

UPDATE [dbo].[Alert] SET [Processed]=N'2018-02-12 20:31:19.470 +00:00' OUTPUT INSERTED.* WHERE [Id] = N'27961F66-E5A5-47FE-8B89-0DA140D2C1EC'

RicardoDMelo:
+1

Vovka19:
THANK YOU! I JUST TRY, but too many keys dosent work at macbook. i will
check later didnt slept 2 days.

вс, 18 февр. 2018 г. в 20:24, Ricardo Melo <notifications@github.com>:

balazsnemeth:
Any news on this?

HenryJW:
The workaround I'm currently using is to use strings instead of dates for my models and then passing the ISO string instead of the date object.

```
// Doesn't work because new Date() is an object, not a string
MyModel.create({
[...]

PR #10793, opened 2019-04-17
Pass useUTC dialect option to mssql/Tedious

molaux:
### Pull Request check-list

### Description of change

Hello,
first of all, many thanks for your great project that I am discovering. 

I struggled for hours to understand why sequelize was not returning dates from mssql according to my timezone : here is my fix.
Note that I have no local mssql to run tests, so even if first part of tests seems to pass, the second part is trying to connect to a local server that I don't have. The server I use is a production one that I use read only  and I have no admin access.
In the same subject, I commented an issue about datetimeoffset serialization and mssql there : https://github.com/sequelize/sequelize/issues/7930#issuecomment-484094201

Hope it can help

### DIFF
--- a/lib/dialects/mssql/connection-manager.js
+++ b/lib/dialects/mssql/connection-manager.js
@@ -56,6 +56,10 @@ class ConnectionManager extends AbstractConnectionManager {
         Object.assign(connectionConfig.authentication, config.dialectOptions.authentication);
       }
 
+      if (config.dialectOptions.useUTC !== undefined) {
+        connectionConfig.options.useUTC = config.dialectOptions.useUTC;
+      }
+
       Object.assign(connectionConfig.options, config.dialectOptions.options);
     }
 


mirkojotic:
I'm glad you've found the solution @molaux and thanks for the contribution! A side note Sequelize comes with docker container definitions with all the databases you might need for integration testing. I don't have mssql locally either but fortunately docker works well. The CI failures are unrelated to your change, I believe.

On to the change, this might need to a documentation change and maybe a default value (also documented). @sushantdhiman

sushantdhiman:
This doesn't require any special treatment, you can pass this option with `dialectOptions` already.

```js
dialectOptions: {
   options: {
      useUtc: true
   }
}
```

molaux:
Thanks @sushantdhiman,  
This way it was consitent with mysql  `useUTC` dialect option. But maybe it is not required.

sushantdhiman:
@molaux I am not aware of `useUTC` option for mysql, can you point to documentation about that. Any thing under `dialectOptions` is dialect client library api, so there is no need for us to document those options.

molaux:
You're right,
sorry for the noise

PR #10800, opened 2019-04-18
feat(types): adding upsert with returning: true

Chocobozzz:
### Pull Request check-list

- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving? **No**
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? **No**

### Description of change

Added correct return type to `upsert` when `{ returning: true }`

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -239,36 +239,36 @@ export interface WhereOperators {
 
   /**
    * MySQL/PG only
-   * 
+   *
    * Matches regular expression, case sensitive
-   * 
+   *
    * Example: `[Op.regexp]: '^[h|a|t]'` becomes `REGEXP/~ '^[h|a|t]'`
    */
   [Op.regexp]?: string;
 
   /**
    * MySQL/PG only
-   * 
+   *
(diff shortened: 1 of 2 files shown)


types/lib/model.d.ts
```diff
@@ -1909,9 +1909,15 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
   public static upsert<M extends Model>(
     this: { new (): M } & typeof Model,
     values: object,
-    options?: UpsertOptions
+    options?: UpsertOptions & { returning?: false | undefined }
```
eseliger: Just a thought:
what might also work: 

```ts
public static upsert<M extends Model, T extends boolean>(this: { new (): M } & typeof Model, values: object, options?: UpsertOptions): Promise<T extends true ? [M, boolean] : boolean>
```

would reduce duplication..

SimonSchick: @eseliger That'd be preferable indeed, most of these typings were written when this functionality wasn't around :)

PR #10815, opened 2019-04-20
feat(types): add index hints

SimonSchick:
### Pull Request check-list

### Description of change

This should have been added with the original PR


types/lib/index-hints.d.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Available index hints to be used for querying data in mysql for index hints.
+ */
+declare enum IndexHints {
+  USE = 'USE',
+  FORCE = 'FORCE',
+  IGNORE = 'IGNORE'
+}
+
+export = IndexHints;
```
SimonSchick: @sushantdhiman this is why I asked you to avoid writing `module.exports` directly, this forces us to use `require` and `export =` in the TS files.

eseliger: What you could do is adding the same stuff in the JS to an exported property and just add this to the type definitions 

```
Module.exports.IndexHints = module.exports
```

SimonSchick: That just makes it even more complicated, we should only be using named exports, no `default` exports either.

PR #10816, opened 2019-04-20
fix(redshift): allow standard_conforming_strings option

aheuermann:
### Pull Request check-list

### Description of change

We use sequelize with Redshift using the sequelize postgres dialect. It doesn't support the `standard_conforming_strings` option since it is based on an older version of postgres.  Since https://github.com/sequelize/sequelize/commit/850c7fd04669e0fef9238b6dc4f8d6ee93ed71e9 was introduced,  sequelize will try to set this option and fail, so I added a config option.


types/lib/sequelize.d.ts
```diff
@@ -305,6 +305,15 @@ export interface Options extends Logging {
    */
   operatorsAliases?: OperatorsAliases | false;
 
+
+  /**
+   * The PostgreSQL `standard_conforming_strings` session parameter. Set to `false` to not set the option.
+   * WARNING: Setting this to false may expose vulnerabilities and is not recommended!
+   *
+   * @default all aliases
```
SimonSchick: Nit: Should be `true`.

types/lib/sequelize.d.ts
```diff
@@ -305,6 +305,15 @@ export interface Options extends Logging {
    */
   operatorsAliases?: OperatorsAliases | false;
```
SimonSchick: Optional: While we are here, can you remove `| false` here pretty please? :)

aheuermann:
Fixed, thanks for taking a look!

I removed the tests as they weren't really testing anything since the  default is `on` and the query to set the option runs on connect and is per-session. I couldn't find a good way to spy on the query itself. If you have any suggestions let me know, and I can add a test.

sushantdhiman:
This should be support by different `dialect = redshift` or version check, it can leave users open to SQL injection if redshift updated their postgres version

PR #10822, opened 2019-04-22
fix(types): add join and values operators

pascal-giguere:
### Pull Request check-list

### Description of change

Add TypeScript typings for the `Op.join` and `Op.values` operators.

Typings currently don't reflect the [source code](https://github.com/sequelize/sequelize/blob/master/lib/operators.js) and [documentation](https://github.com/sequelize/sequelize/blob/master/docs/querying.md).

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -128,6 +128,11 @@ export interface AllOperator {
   [Op.all]: (string | number)[];
 }
 
+/** Undocumented? */
+export interface ValuesOperator {
+  [Op.values]: (string | number)[];
+}
+
 /**
  * Operators that can be used in WhereOptions
  *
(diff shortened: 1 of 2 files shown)


types/lib/operators.d.ts
```diff
@@ -18,6 +18,7 @@ declare const Op: {
   readonly in: unique symbol;
   readonly iRegexp: unique symbol;
   readonly is: unique symbol;
+  readonly join: unique symbol;
```
SimonSchick: Op.join only appears to be referenced a single time, I'm not sure if this is internal only.

types/lib/operators.d.ts
```diff
@@ -39,5 +40,6 @@ declare const Op: {
   readonly strictLeft: unique symbol;
   readonly strictRight: unique symbol;
   readonly substring: unique symbol;
+  readonly values: unique symbol;
```
SimonSchick: You will have to add this to `model.d.ts` too, probably with a signature of `(number | string)[]`.

SimonSchick:
Please omit `join` for now.

types/lib/operators.d.ts
pascal-giguere: To be frank I'm not super comfortable editing `model.d.ts` with my limited knowledge of Sequelize, especially with `values` only being partially documented. I tried something, let me know if it makes sense.

types/lib/model.d.ts
```diff
@@ -128,6 +128,11 @@ export interface AllOperator {
   [Op.all]: (string | number)[];
 }
 
+/** Undocumented? */
```
SimonSchick: If you know how this works please add it to the documentation as well!

types/lib/model.d.ts
```diff
@@ -170,14 +175,14 @@ export interface WhereOperators {
    *  - `[Op.like]: '%hat',` becomes `LIKE '%hat'`
    *  - `[Op.like]: { [Op.any]: ['cat', 'hat']}` becomes `LIKE ANY ARRAY['cat', 'hat']`
    */
-  [Op.like]?: string | Literal | AnyOperator | AllOperator;
+  [Op.like]?: string | Literal | AnyOperator | AllOperator | ValuesOperator;
```
SimonSchick: Is this actually correct?
The only reference in the code I can see is within an `Op.all` context.

pascal-giguere:
Unfortunately the change is more complex than I anticipated, and it's more than I feel comfortable doing, as I'm not familiar enough with Sequelize to write proper typings, documentation and tests.

I created an issue instead: https://github.com/sequelize/sequelize/issues/10838 so that someone who's more familiar than me can pick it up.

Issue #8663, opened 2017-11-20
Issue with using the new [Op.gt] symbols across library boundaries 


mansona:
I know this may seem like a bit of an edge case, but I thought it was prudent to bring it up as a different issue even though it is related to the discussion here: https://github.com/sequelize/sequelize/issues/8417

I am developing a "common" library that is to be used between different node applications using the same Sequelize models and connections. I am in the process of splitting out the Sequelize code but I couldn't understand why something was not working. I'll try my best to lay out the scenario below: 

The API application has the following dependencies in package.json:
```
    "@company/sequelize-common": "^1.18.7",
    "sequelize": "^4.22.8",
```

and the sequelize-common libarary has the following dependency:
```
    "sequelize": "^4.22.8",
```

The API code has the following (cut down) code to check Access Tokens: 
```
const { Op } = require('sequelize');
const { db } = require('@company/sequelize-common');

module.exports = (req, res, next) => {
    db.token.findOne({
      where: {
        token: req.query.token,
        expires: {
          [Op.gt]: new Date(),
        },
      },
    }).then((token) => {
      // do stuff with token
    });
};
```

Now this code was failing *every time*, no matter what I did it would never find the right token that would expire in the future even though I confirmed with the database multiple times.

It's only when I saw the debug log of the SQL that I saw the problem: 
```
SELECT "token"."token", "token"."expires" FROM "token_table" AS "token" WHERE "token"."token" = 'big_long_token' AND "token"."expires" = '2017-11-20 14:54:26.100 +00:00';
```

Even though I was querying:
```
        expires: {
          [Op.gt]: new Date(),
        },
```
it was using `=` as the operator!?

And this is where it hit me that the above `sequelize.Op.gt` is *not the same symbol* as the internal Sequelize *inside* the `@company/sequelize-common` module. Because of the context switch it didn't understand the Op symbol and *defaulted to eq*.

Now I had a bad feeling about this change, and I understand your reasoning for the security of what you are doing but I *strongly recommend* that you reconsider your choice to a) make this a deprecation or b) any plans to make this a required default behaviour in the future. 

A lot of developers have already tripped up on this issue as it says here: https://github.com/sequelize/sequelize/issues/8417 and would imagine that there are more cases like mine that will cause *actual real bugs in production code* based on slight misunderstandings of the subtleties of this change.

sushantdhiman:
Your concern is very real and this could potentially break many tools and libraries.

[...]

PR #10830, opened 2019-04-23
refactor(operators): use namespaced symbol names for operators

SimonSchick:
### Pull Request check-list

### Description of change

Using `Symbol.for` adds them to a global registry of symbols, this seems redundant in the case of sequelize, an alternative would be to namespace them (as nodejs does).


sushantdhiman:
We can namespace them like `sequelize.op.<>`.  We use `Symbol.for` to prevent issues like https://github.com/sequelize/sequelize/issues/8663

This is probably a major change, there is no real benefit from this change either

SimonSchick:
I read the issue you linked and I don't understand what is going on, did they install different versions of sequelize in the same project? As long as you always use `Op` consistently the references should always be the same.

I think renaming them is fine to (also helps when debugging), but I don't think the symbol names should be considered public API 😛

sushantdhiman:
Yes this is useful when a project is using two version of sequelize, depends upon how they are included by `require`. If we use `Symbol`, we end up creating a new symbol which wont match other version of sequelize.

This is a breaking change because you can no longer mix symbols from various versions of Sequelize, not sure if someone's setup depends on that.

SimonSchick:
Tbf, if someone loads 2 different versions of sequelize they probably have significantly worse problems than the symbols 😛 (instanceof checks etc.).
I will go ahead and namespace the symbols and target the PR against v6.

SimonSchick:
Going to move this into v6.

Issue #9440, opened 2018-05-15
Named parameter has no value in the given object.


ben-eb:
## What are you doing?

I'd like to use named parameters - query below is contrived, I would expect to send more parameters than this when doing a real query. Unfortunately the regex for finding replacements does not take into account values inside string literals.

```js
const Sequelize = require('sequelize');

const sequelize = new Sequelize(process.env.DATABASE_URI, {
  dialect: 'postgres',
});

const query = `
  SELECT * FROM "_Join:roles:_Role" WHERE "owningId" = :owningId
`;

sequelize.query(query, {
  type: sequelize.QueryTypes.SELECT,
  replacements: {
    owningId: 'TQdJ6RuES6'
  }
}).then(console.warn)
```

Relevant line is here:

https://github.com/sequelize/sequelize/blob/v4/lib/sql-string.js#L112

## What do you expect to happen?

Sequelize should not try to find a named parameter inside a string literal.

## What is actually happening?

Sequelize is trying to find a `:roles` parameter.

```
Unhandled rejection Error: Named parameter ":roles" has no value in the given object.
    at sql.replace (/app/node_modules/sequelize/lib/sql-string.js:120:13)
    at String.replace (<anonymous>)
    at Object.formatNamedParameters (/app/node_modules/sequelize/lib/sql-string.js:112:14)
    at Object.formatNamedParameters (/app/node_modules/sequelize/lib/utils.js:126:20)
    at Promise.try (/app/node_modules/sequelize/lib/sequelize.js:500:23)
    at tryCatcher (/app/node_modules/bluebird/js/release/util.js:16:23)
    at Function.Promise.attempt.Promise.try (/app/node_modules/bluebird/js/release/method.js:39:29)
    at retryParameters (/app/node_modules/sequelize/lib/sequelize.js:454:48)
    at /app/node_modules/retry-as-promised/index.js:39:21
    at Promise._execute (/app/node_modules/bluebird/js/release/debuggability.js:303:9)
    at Promise._resolveFromExecutor (/app/node_modules/bluebird/js/release/promise.js:483:18)
    at new Promise (/app/node_modules/bluebird/js/release/promise.js:79:10)
    at retryAsPromised (/app/node_modules/retry-as-promised/index.js:29:10)
    at Sequelize.query (/app/node_modules/sequelize/lib/sequelize.js:454:12)
    at Object.<anonymous> (/app/test.js:11:11)
    at Module._compile (module.js:652:30)
[...]

PR #10841, opened 2019-04-25
feat(query): add liberalReplace switch to query

k4ny:
adds possibility to switch if missing replacement result in error or not

### Pull Request check-list

### Description of change

PR solves [issue](https://github.com/sequelize/sequelize/issues/9440) when query contains ':' character as a part of query data. It's solved by introducing new option called `liberalReplace`. If `liberalReplace` set true, replacement method ignores missing replacement value and pastes back origin value to the query.

### DIFF
--- a/docs/raw-queries.md
+++ b/docs/raw-queries.md
@@ -76,6 +76,18 @@ sequelize.query('SELECT * FROM users WHERE name LIKE :search_name ',
 })
 ```
 
+### Liberal replacement
+
+There can be used character `:` as a valid part of the query data. Default replacement behaviour throws an Error for missing replacement key. In this case there is a possibility to ignore missing replace key in replacements object via query option `liberalReplace` that returns back original value to the query.
+
+```js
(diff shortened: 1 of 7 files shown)


SimonSchick:
This seems more like a workaround, can we instead add an escape character for `:`?

k4ny:
@SimonSchick no problem, but how it should look like? `::` is used in postgres (dialect)[https://github.com/sequelize/sequelize/blob/8d1972f6dee6a861b825fbc33e3b3078e8563d58/lib/sql-string.js#L114] Can we use `\:`?

SimonSchick:
An alternative would be allowing to specify the templating prefix, thoughts @sushantdhiman @eseliger ?

eseliger:
I think some tools rely on the colon and don't allow to pass custom properties 🤔 eg for template strings I think there are some.. but have to research further

luanxuechao:
Hi, Is there any other solutions?

PR #10852, opened 2019-04-27
refactor(model): remove build in favor or constructor usage

eseliger:
### Pull Request check-list

### Description of change

Removes the build function in favor of the constructor usage as it's really just a proxy and it's supposed to make it much clearer that there is nothing wrong about using the constructor directly.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2143,24 +2143,15 @@ class Model {
   }
 
   /**
-   * Builds a new model instance.
+   * Builds multiple models in one operation.
    *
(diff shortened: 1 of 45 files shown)


lib/model.js
```diff
@@ -2126,50 +2126,12 @@ class Model {
...
-   */
-  static build(values, options) {
-    if (Array.isArray(values)) {
-      return this.bulkBuild(values, options);
-    }
-
-    return new this(values, options);
-  }
-
-  static bulkBuild(valueSets, options) {
```
SimonSchick: Are we sure we want to remove this function? After all this has some distinct funtionality.

PR #10881, opened 2019-05-01
Docs: fixed typo in the model definition example

toxuin:
### Description of change

Copy-pasting an example into your code would make it crash. This is obvious typo mistake.


eseliger:
good catch :)

PR #10885, opened 2019-05-02
fix(typings): use correct set of allowed options for addIndex

eseliger:
### Pull Request check-list

### Description of change
I looked up all the values that are used and valid and cleaned up the index types a bit. Also merged with the query interface to have a consistent interface (as they both call addIndexQuery anyways)

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -16,7 +16,7 @@ import { ValidationOptions } from './instance-validator';
 import { ModelManager } from './model-manager';
 import Op = require('./operators');
 import { Promise } from './promise';
-import { QueryOptions } from './query-interface';
+import { QueryOptions, IndexesOptions } from './query-interface';
 import { Config, Options, Sequelize, SyncOptions } from './sequelize';
 import { Transaction } from './transaction';
 import { Col, Fn, Literal, Where } from './utils';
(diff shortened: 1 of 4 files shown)


types/lib/model.d.ts
```diff
@@ -1140,16 +1140,13 @@ export interface ModelIndexesOptions {
...
+  /** For FULLTEXT columns set your parser */
+  parser?: string | null;
 
   /**
-   * The method to create the index by (`USING` statement in SQL). BTREE and HASH are supported by mysql and
-   * postgres, and postgres additionally supports GIST and GIN.
+   * Index type. Only used by mysql. One of `UNIQUE`, `FULLTEXT` and `SPATIAL`
    */
-  method?: string;
+  type?: 'UNIQUE' | 'FULLTEXT' | 'SPATIAL';
```
SimonSchick: Please move these to separate types, some may already exist.

types/lib/query-interface.d.ts
```diff
@@ -123,18 +123,7 @@ export interface QueryInterfaceDropAllTablesOptions extends QueryInterfaceOption
...
-  /** The name of the index. Default is __ */
-  name?: string;
-
-  /** For FULLTEXT columns set your parser */
-  parser?: string;
-
-  /** Set a type for the index, e.g. BTREE. See the documentation of the used dialect */
-  using?: string;
-}
+export interface QueryInterfaceIndexOptions extends ModelIndexesOptions, QueryInterfaceOptions {}
```
SimonSchick: Imo the model options should inherit from query interface options.

PR #10895, opened 2019-05-05
fix: cleanup parser-store

eseliger:
### Pull Request check-list


lib/dialects/abstract/parser-store.js
```diff
@@ -0,0 +1,14 @@
+class ParserStore extends Map {
```
SimonSchick: Do we really want to subclass map?

lib/dialects/abstract/parser-store.js
```diff
@@ -0,0 +1,14 @@
...
+  }
+
+  refresh(dataType) {
+    for (const type of dataType.types[this.dialectName]) {
+      this.set(type, dataType.parse);
+    }
+  }
+}
+
+module.exports = { ParserStore };
```
SimonSchick: Prefer `exports.name = name`

lib/dialects/abstract/parser-store.js
eseliger: wouldn't we? It really is a map, and was before it just proxied very weirdly to add this one extra function 🤔 also, it's really easy to type it well

PR #10899, opened 2019-05-05
docs: separate usage.md and remove duplicates

papb:
The ["Basic Usage"](https://github.com/sequelize/sequelize/blob/5f9d5907425ae1d86b402133718d007a52f2b39e/docs/usage.md) tutorial contained duplicate explanations from other places and mentioned some things that I wouldn't call "basic usage". This PR moves everything relevant from that file to other places, where they fit better.


SimonSchick:
Gotta fix md lint first tho :)

papb:
@SimonSchick Oops, fixed :grimacing:

polarathene:
The [dialects doc](http://docs.sequelizejs.com/manual/dialects.html) links to the old usage doc for `options`. Just hit it and found out why it's 404.

I am not sure if it exists anymore in the official docs, at least it was not clear where. 

While at it, when I first came across the docs and went to `Getting Started` I was a bit confused as I thought maybe that was the top-level doc that I had landed on(index) originally but it wasn't, that requires you to click the logo or manually adjust the url. Minor UX thing.

As a new user, I also came across [these older docs](https://sequelize.readthedocs.io/) last updated in 2016. It's not immediately clear that they're out of date either, if that's something under this projects control, perhaps it should be pulled down or given some notice to the maintained docs.

PR #10918, opened 2019-05-08
docs: add index operator option

emmanuelgautier:
### Description of change

Add index operator option for API Documentation.

Operator option is used here : https://github.com/sequelize/sequelize/blob/5f9d5907425ae1d86b402133718d007a52f2b39e/lib/dialects/abstract/query-generator.js#L577

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -477,6 +477,8 @@ class QueryGenerator {
                   - length: An integer. Optional
                   - order: 'ASC' or 'DESC'. Optional
         - parser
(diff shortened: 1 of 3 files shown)


SimonSchick:
Just to sanity check, is this present in the typings?

emmanuelgautier:
@SimonSchick Yes, this option is present in typing : https://github.com/sequelize/sequelize/blob/master/types/lib/query-interface.d.ts#L174.

For older versions, I made a PR in DefinitelyTyped project https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35284

PR #10939, opened 2019-05-12
Undocumented alter option in SyncOptions

wouterds:
Thought it was removed, but it's still there. Just not documented.

Workaround until this is merged `await sequelize.sync({ alter: true } as SyncOptions);`


wouterds:
I see #10948 also fixes this issue

wouterds:
🎉

PR #10990, opened 2019-05-22
fix(types): add literal to possible where options

sHooKDT:
I faced a problem with typing during usage of raw query for where.
Since WhereOptions didn't accept Literal, compiler complained. But with since it works ok with ts-ignore, I decided to fix typings here.


SimonSchick:
Needs test.

sHooKDT:
@SimonSchick test for behaviour with literal or test for typings?

SimonSchick:
Just for tests.

sHooKDT:
Finally I found some time for it.
So, check this out. I'm not sure what CodeCov is talking about, because my changes couldn't affect coverage in any way.

Issue #10925, opened 2019-05-09
findByPk options where does't work 


YinghaoFu:
"sequelize": "^5.6.0"  
in docment:
"See:
Model.findAll for a full explanation of options"
I try add 'where' to findByPk's options,but it's did't work.

seromenho:
There is no `where` on the options to [findByPk.
Documentation here](http://docs.sequelizejs.com/class/lib/model.js~Model.html#static-method-findByPk)

YinghaoFu:
> There is no `where` on the options to `findByPk`
> [Documentation here](http://docs.sequelizejs.com/class/lib/model.js~Model.html#static-method-findByPk)
Documentation here

> Model.findAll for a full explanation of options 
It's does't say no `where` on on the options to `findByPk`.

smuralimohan:
Hi, I am testing Sequelize using Postgres dialect. For some reason, **findByPk(1)** (with id = 1) does not work at all. And it is not just findByPk(1) that does not work, but also **findOne({where: {id:1}}), findAll({ where: {id:1} })** and **findAll({ where: { id: { [Op.eq] : 1}  }})** don't work.

However, for ids > 1 it is working fine. I tested using in-memory SQLite also and the problem exists there as well. 
My sequelize and node versions are as below.

"pg": "^8.7.1",
 "pg-hstore": "^2.3.4",
 "sequelize": "^6.12.0",

Node: v16.13.0

My database table have the following records.
id      name                notes
1	"Daily Dose"	"Diabetes, BP"
2	"Daily Dose"	"Diabetes, BP"
3	"Daily Dose"	"Sugar, BP"

PR #11005, opened 2019-05-27
docs(findbypk): fix document and type definition

dy93:
### Description of change
Fix issue #10925, correct the documentation and type definition of Model.findByPk()

findByPk takes the same option as findAll, except the `where` option in the implementation, so we should correct the document and type definition

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -489,6 +489,8 @@ export interface IndexHintable {
   indexHints?: IndexHint[];
 }
 
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
+
 /**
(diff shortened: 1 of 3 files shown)


types/lib/model.d.ts
```diff
@@ -489,6 +489,8 @@ export interface IndexHintable {
   indexHints?: IndexHint[];
 }
 
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
```
SimonSchick: This is in conflict with the new `Omit` definition that will ship with the next version of TS, so this line might be removed in the future.

SimonSchick: This is in conflict with the new `Omit` definition that will ship with the next version of TS, so this line might be removed in the future.

PR #11009, opened 2019-05-28
fix(types): add support for `model.associate` in typescript

joeytepp:
### Pull Request check-list

### Description of change

A common pattern for adding associations between models is defining a `model.associate` function (see [this example](https://github.com/sequelize/express-example/blob/master/models/task.js) here). However, when using Typescript, this does not work as there is no associate function defined in the type definitions. This pull request provides the necessary type definitions so that this pattern can be used in Typescript as well.

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -2496,6 +2496,19 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
    */
   constructor(values?: object, options?: BuildOptions);
 
+   /**
+   * A function where asociations between different models are defined.
+   * For example:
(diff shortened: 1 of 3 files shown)


sushantdhiman:
I don't think this should be added as Sequelize itself doesn't support this method. It is a common pattern, but not a part of API.

Can't typescript handle this without adding this method to our base typings @SimonSchick?

SimonSchick:
You are correct, this should not be added, if you want to use this method you must declare it on every model individually, you are also not required to perform associations in these methods, I keep them in a central file for instance.

Issue #2701, opened 2014-12-09
Is it possible to add foreign key to a table with schema in migrations?


H256:
I'm trying to add a foreign-key constraint within a table on postgres using migrations.

Here's what i do - Sequelize doesn't attach the schema to the tables for the constraint, thus the statement fails:

``` javascript
migration.createTable("Groups", {
            id: {
               allowNull: false,
               autoIncrement: true,
                primaryKey: true,
               type: DataTypes.INTEGER
            },
            Name: {
                type: DataTypes.STRING(64)
            }, {schema: "userstuff"});

migration.createTable("User", {
            id: {
               allowNull: false,
               autoIncrement: true,
                primaryKey: true,
                type: DataTypes.INTEGER
            },        
           GroupId: {
                type: DataTypes.INTEGER,
               references: "Groups",
               referenceKey: "id",
               onUpdate: "NO ACTION",
               onDelete: "NO ACTION"
          },    
          Name: {
               type: DataTypes.STRING(64)
           }, {schema: "userstuff"});
```

I get a database error 'Relation "Group" does not exist.'

I've also tried prefixing the 'references'-tablename with the schema to no avail :( Is there any workaround for this?

sushantdhiman:
Can be closed @mickhansen @janmeier ?

mickhansen:
@sushantdhiman Stil lacks testing and proper documentation.

iamjoyce:
This issue can be closed, it's just that it will be useful to have a code snippet sample:

```js
queryInterface.createTable("User", {
    RoleId: {
        type: Sequelize.INTEGER,
        references: {
            model: {
                tableName: "Role",
                schema: "Authentication"
            },
            key: "Id"
        },
        allowNull: false
    }
});
```

stonedraider:
iamjoyce you saved my life!

furqanaziz:
How to apply `onDelete: 'cascade'` on reference? I tried following but it did not worked,
```
references: {
  onDelete: 'cascade',
  model: 'my_refernced_table',
},
```

Vitor-Felipe:
@iamjoyce , it is possible to set the name of my foreign key?

sushantdhiman:
https://github.com/sequelize/sequelize/pull/11097

PR #11064, opened 2019-06-14
docs: update v5 docs (#2701)

VictorKunzler:
Added docs for migrations with foreign key in another Schema

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -248,6 +248,40 @@ module.exports = {
         })
     }
 };
+
+```
+
+The next is an example of a migration that has a foreign key. You can use the `references` property to specify the table that has being referenced.
+
+```js
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('Person', {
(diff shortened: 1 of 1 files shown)


docs/migrations.md
```diff
@@ -249,6 +249,36 @@ module.exports = {
     }
 };
 ```
+The next is an example of a migration that has a foreign key. You can use the `references` property to specify the table that has being referenced.
```
paulodiovani: You can use `references` to specify a foreign key:

```js
      userId: {
        type: Sequelize.INTEGER,
        references: {
          model: {
            tableName: 'users',
            schema: 'otherSchema'
          }
          key: 'id'
        },
        allowNull: false
      },
```

docs/migrations.md
```diff
@@ -249,6 +249,36 @@ module.exports = {
     }
 };
 ```
```
paulodiovani: According to `npm run lint-docs` you must have a blank line before and after any code block.
It is missing one here.

PR #11068, opened 2019-06-14
fix(types): add missing findCreateFind type

levibostian:


levibostian:
Oh, sorry about that. Good catch. I added a test to the typing. If it doesn't satisfy the needs of the project, I am happy to make a change.

Issue #8352, opened 2017-09-21
Unexpected behaviour of Retry feature using some RDBMSs versions


rubens21:
## What are you doing?
The Retry feature fails to identify a rolled back transaction caused by a deadlock (or there is a lack of documentation).

The default behaviour of some MySQL and MariaDb versions is rolling back a transaction if a dead lock is detected (https://dev.mysql.com/doc/refman/5.7/en/innodb-deadlock-detection.html). 
Although it may be changed by configuration option (MySQL 
https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_deadlock_detect, and MariaDb >= 10.2.26 https://mariadb.com/kb/en/library/xtradbinnodb-server-system-variables/#innodb_deadlock_detect), the default action is rolling back the whole transaction, not only the last query.

Due this behaviour, the Retry option may hide an error because the Transaction Promise chain is not broken even after the transaction in the database has been rolled back.

In summary, the Sequelize Transaction promise chain may finish successfully even when all data has been lost.

```js
const Promise = require('bluebird');
const Sequelize = require('sequelize');

let sequelize = new Sequelize('maydb', 'myuser', 'a-secure-password',
  {
    host: 'localhost',
    dialect: 'mysql',
    retry: {
      match: /Deadlock/i,
      max: 5
    },
    logging: console.log
  });

/*
    CREATE TABLE foo (col1 TEXT);
    CREATE TABLE tableA (col1 INT);
    INSERT INTO tableA VALUES (0);
    CREATE TABLE tableB (col1 INT);
    INSERT INTO  tableB VALUES (0);
     */

let update = (t, table, value, delay) => {
  return new Promise((resolve, reject) => {
    setTimeout(() => {
      console.log(`${table} -> ${value};`);
      sequelize
      .query(`UPDATE ${table} SET col1 = ${value};`, {transaction: t})
      .then(resolve)
      .catch(reject);
    }, delay);
  });
};

let process1 = () => {
  return sequelize.transaction((t1) => {
    return writeInFoo("Process1 has started", t1).then(()=> {
      return update(t1, 'tableA', 1, 0);//first update: timer to keep the right order to force a dead lock
    }).then(() => {
      return writeInFoo("Process1 changed tableA", t1);
    }).then(() => {
      return update(t1, 'tableB', 1, 10);//fourth update: timer to keep the right order to force a dead lock
    }).then(() => {
[...]

PR #11074, opened 2019-06-17
fix(mysql): deadlocks indicate a rollback

schmod:
### Pull Request check-list

### Description of change

When MySQL or MariaDB encounter a deadlock, [the default behavior](https://dev.mysql.com/doc/refman/5.7/en/innodb-deadlock-detection.html) is to roll back the active transaction (having roughly the same effect as manually issuing the `ROLLBACK;` command). 

Currently, Sequelize does not apply any special handling to these errors.  If a query yields a deadlock error, all future queries against that "transaction" will actually be run against the default context.  This can have potentially destructive implications, as those subsequent queries cannot be rolled back.

This PR adds some code to mark any deadlocked transactions as though they've been rolled-back.  This prevents Sequelize from running any further queries against these transactions (utilizing the mechanisms that already exist for that purpose).

Mitigates #8352.

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -614,17 +614,22 @@ class Sequelize {
         [sql, bindParameters] = this.dialect.Query.formatBindParameters(sql, options.bind, this.options.dialect);
       }
 
+      const checkTransaction = () => {
+        if (options.transaction && options.transaction.finished && !options.completesTransaction) {
+          const error = new Error(`${options.transaction.finished} has been called on this transaction(${options.transaction.id}), you can no longer use it. (The rejected query is attached as the 'sql' property of this error)`);
+          error.sql = sql;
+          throw error;
+        }
+      };
+
       const retryOptions = Object.assign({}, this.options.retry, options.retry || {});
 
       return Promise.resolve(retry(() => Promise.try(() => {
         if (options.transaction === undefined && Sequelize._cls) {
           options.transaction = Sequelize._cls.get('transaction');
         }
-        if (options.transaction && options.transaction.finished) {
-          const error = new Error(`${options.transaction.finished} has been called on this transaction(${options.transaction.id}), you can no longer use it. (The rejected query is attached as the 'sql' property of this error)`);
-          error.sql = sql;
-          throw error;
-        }
+
+        checkTransaction();
 
         return options.transaction
           ? options.transaction.connection
@@ -632,6 +637,7 @@ class Sequelize {
       }).then(connection => {
         const query = new this.dialect.Query(connection, this, options);
         return this.runHooks('beforeQuery', options, query)
+          .then(() => checkTransaction())
           .then(() => query.run(sql, bindParameters))
           .finally(() => this.runHooks('afterQuery', options, query))
           .finally(() => {
--- a/test/integration/dialects/mysql/errors.test.js
+++ b/test/integration/dialects/mysql/errors.test.js
@@ -70,7 +70,6 @@ if (dialect === 'mysql') {
             reltype: 'child'
           }));
       });
-
     });
   });
 }
--- a/lib/dialects/mariadb/query.js
+++ b/lib/dialects/mariadb/query.js
@@ -34,10 +34,10 @@ class Query extends AbstractQuery {
 
   run(sql, parameters) {
     this.sql = sql;
-    const { connection } = this;
+    const { connection, options } = this;
 
     const showWarnings = this.sequelize.options.showWarnings
-      || this.options.showWarnings;
+      || options.showWarnings;
 
     const complete = this._logQuery(sql, debug);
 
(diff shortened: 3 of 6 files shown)


schmod:
I don't think the test failures are related to this change.  They pass locally for me.

lib/sequelize.js
```diff
@@ -614,24 +614,30 @@ class Sequelize {
         [sql, bindParameters] = this.dialect.Query.formatBindParameters(sql, options.bind, this.options.dialect);
       }
 
+      const checkTransaction = () => {
+        if (options.transaction && options.transaction.finished && !options.completesTransaction) {
```
sushantdhiman: Why `completesTransaction` is added to this check? Is it to let rollback and commit run on already finished transaction?

lib/sequelize.js
```diff
@@ -614,24 +614,30 @@ class Sequelize {
...
+
+        checkTransaction();
 
         return options.transaction
           ? options.transaction.connection
           : this.connectionManager.getConnection(options);
       }).then(connection => {
         const query = new this.dialect.Query(connection, this, options);
         return this.runHooks('beforeQuery', options, query)
+          .tap(() => checkTransaction())
```
sushantdhiman: Why `checkTransaction` is called here? To check deadlocks caused by `beforeQuery`?

Also just use `then` here, we aren't using return value in promise chain

test/integration/dialects/mysql/errors.test.js
```diff
@@ -70,7 +70,52 @@ if (dialect === 'mysql') {
             reltype: 'child'
           }));
       });
+    });
+
+    describe('Deadlock in transaction', () => {
```
sushantdhiman: Use this testcase, modify as per your requirements. Add this test-case to https://github.com/sequelize/sequelize/blob/master/test/integration/transaction.test.js#L368

```js
    if (dialect === 'mysql' || dialect === 'mariadb') {
      describe('deadlock handling', () => {
        it('should treat deadlocked transaction as rollback', function () {
          const Task = this.sequelize.define('task', {
            id: {
              type: Sequelize.INTEGER,
              primaryKey: true
            }
          });

          // This gets called twice simultaneously, and we expect at least one of the calls to encounter a
          // deadlock (which effectively rolls back the active transaction).
          // We only expect createTask() to insert rows if a transaction is active.  If deadlocks are handled
          // properly, it should only execute a query if we're actually inside a real transaction.  If it does
          // execute a query, we expect the newly-created rows to be destroyed when we forcibly rollback by
          // throwing an error.
          // tl;dr; This test is designed to ensure that this function never inserts and commits a new row.
          const update = (from, to) => this.sequelize.transaction(transaction => {
            return Task.findAll({
              where: {
                id: {
                  [Sequelize.Op.eq]: from
                }
              },
              lock: transaction.LOCK.UPDATE,
              transaction
            })
            .then(() => Promise.delay(10))
            .then(() => {
              return Task.update({ id: to }, {
                where: {
                  id: {
                    [Sequelize.Op.ne]: to
                  }
                },
                lock: transaction.LOCK.UPDATE,
                transaction
              });
            })
            .catch(e => { console.log(e.message); })
            .then(() => Task.create({ id: 2 }, { transaction }))
            .catch(e => { console.log(e.message); })
            .then(() => { throw new Error('Rollback!'); });
          }).catch(() => {});

          return this.sequelize.sync({ force: true })
            .then(() => Task.create({ id: 0 }))
            .then(() => Task.create({ id: 1 }))
            .then(() => Promise.all([
              update(0, 1),
              update(1, 0)
            ]))
            .then(() => {
              return Task.count().then(count => {
                // If we were actually inside a transaction when we called `Task.create({ id: 2 })`, no new rows should be added.
                expect(count).to.equal(2, 'transactions were fully rolled-back, and no new rows were added');
              });
            });
        });
      });
    }
```

lib/sequelize.js
schmod: It's because we don't want to actually run this check against a `COMMIT` or `ROLLBACK` query, because `transaction.finished` actually gets set just before the query actually fires off.

lib/sequelize.js
schmod: Correct.  I wanted to move this check as close to the actual execution of the query as possible.  

If a commit/rollback/deadlock happens _during_ `getConnection()` or `beforeQuery`, I don't want to mistakenly run a query outside of a transaction.  

Because `beforeQuery` and `getConnection()` can both be asynchronous, I'm actually more concerned about something in the background finishing the transaction, rather than `beforeQuery` committing the transaction itself.

I'll change this to a `then`.

test/integration/dialects/mysql/errors.test.js
schmod: Are you suggesting that I should remove the test from this file as well?

PR #11095, opened 2019-06-21
feat(postgres): minify aliases option

mrrinot:
### Pull Request check-list

### Description of change
Add alias minifier to fix the following issue: https://github.com/sequelize/sequelize/issues/2084  

All aliases are replaced with an auto increment alias (_1, _2, _3 ...) before querying and immediately replaced after the query has been run.

This does pose a colliding problem if any column is named accordingly to the alias and I could see the case for making it customizable or just more complex.  

I didn't found the right place in the doc to warn users about that but I would be happy to write it down somewhere appropriate.

(Also thanks to @mattp94 for helping me do this)


shawnbissell:
@mrrinot I've found an issue related to running raw queries ... I get the following error:
TypeError: Cannot read property 'get' of undefined
        at Object.entries.reduce (/.../web/node_modules/sequelize/lib/dialects/postgres/query.js:80:49)
Here is my query:
models.sequelize.query(`select exists(select * from pg_proc where proname = 'ssl_is_used');`,
    { type: models.sequelize.QueryTypes.SELECT });

mrrinot:
@eseliger I'm currently looking through the tests to find where to add my tests.  
Do you have a specific place in mind ?

mrrinot:
@eseliger Tests have been added

mtryba:
@eseliger I checked it and generally looks very good, but when I use `belongsToMany` so query didn't use aliases. Can you confirm it?

mrrinot:
@mtryba I'm sorry I didn't understand your message. Do you mean it doesn't work with `belongsToMany` ?

mtryba:
@mrrinot It means that when I have relations one to one everything is ok, aliases work correct but I have one relation many to many and aliases in the relation are without changes. It looks like
```
"deliveryMaterialInspections->inspectionRejectReasons"."updatedAt" AS "_34",
"deliveryMaterialInspections->inspectionRejectReasons->deliveryMaterialInspectionRejectReason"."id" AS "deliveryMaterialInspections.inspectionRejectReasons.deliveryMaterialInspectionRejectReason.id",
```

the second line is the without alias

mrrinot:
Ho I see, I'll look into that and add a test case.

mrrinot:
At first I thought that table name aliasing wasn't affected by the Postgres limit of 64 but it seems otherwise. I need to fix that too and it complexify the problem quite a bit.

mtryba:
@mrrinot I checked it and I think everything is ok ❤️ Probably you safe my life 😄

mtryba:
@eseliger Can we merge it?

shawnbissell:
I've been testing this with a large codebase (hundreds of different queries and many deeply nested) without problems. How do we get this reviewed and merged to master? @sushantdhiman @mickhansen @papb

mrrinot:
Is there a way to restart the CI build manually without pushing something ?

I don't know if it is specific to this branch but the CI seems rather unstable and seems to pass, or not, based on luck, most of the time. 😕

eseliger:
```
git commit --amend --no-edit && git push -f
```
is a kinda easy way 😀 but I restarted it already

cyrilchapon:
Merge this for Christ sake 😍

This is fixing a 5 years old, pretty critical and inconsistency creating issue => https://github.com/sequelize/sequelize/issues/2084

antoine-pous:
This PR is an important fix and should be merged ASAP, it took only few minutes to review it. Actually i've to use crappy workaround which can break my app logic...

papb:
@cyrilchapon @antoine-pous Hello, we are all eager for improvements on Sequelize, but sometimes things aren't so fast in FOS software.

But if you are in a big hurry, this is what you could do at your own risk:

1. Clone mrrinot's github repo locally
2. Rebase his changes to the official sequelize master branch, as follows:

```
git remote add upstream https://github.com/sequelize/sequelize.git
git fetch upstream
git rebase upstream/master
```

3. Instead of using the official sequelize in your project, use your local repo instead, until this is merged

mrrinot:
@sushantdhiman I do not know how to do that and would happily do it.

 Does that mean that the test suites I added are not needed if I add an entire build ? Or I just let them anyway ?

sushantdhiman:
- Set `minifyAliases` against an env variable (`SEQ_PG_MINIFY_ALIASES`) [here](https://github.com/sequelize/sequelize/blob/master/test/config/config.js#L79) 
- Create a new pg 10 build with [travis.yml](https://github.com/sequelize/sequelize/blob/master/.travis.yml#L63) and set `SEQ_PG_MINIFY_ALIASES=1` for that build
 

> Does that mean that the test suites I added are not needed if I add an entire build ? Or I just let them anyway ?

Keep those tests

sushantdhiman:
@mrrinot https://github.com/sequelize/sequelize/blob/master/test/support.js#L65 Please add `minifyAliases` here so build can actually use it

cyrilchapon:
@papb
> @cyrilchapon @antoine-pous Hello, we are all eager for improvements on Sequelize, but sometimes things aren't so fast in FOS software.

Thanks for the piece of information. Though 5 years a bugfix seems a pretty decent wait in my mind. But I might be wrong. We already forked and we already bound our repos to that branch; which reintroduced vulnerabilities because the rebase was 2 months ago. Making a dangerous installation on a production repository is not OK for us; especially **in an ORM dependency**. Doing it so without any guarantee the feature is merged on the next 5 years especially.

"Things are slow", "you can bind to a develop branch without any guarantee to resolve" are not the right answer to a critical bugfix in an ORM, sorry.

Anyway, this is fixed, this is going to be merged; love Sequelize ❤️

sushantdhiman:
`40+` tests are failing with minify aliases option. Once these are resolved we can merge this PR. All participants are welcome to tackle failing tests.

mrrinot:
I started looking into the failing tests a bit.
The first one [here](https://github.com/sequelize/sequelize/blob/master/test/integration/associations/belongs-to-many.test.js#L275) fails on the first `findOne` with the following query:
```sql
SELECT "User".*,
       "Groups"."group_id"                  AS "_1",
       "Groups"."createdAt"                 AS "_2",
       "Groups"."updatedAt"                 AS "_3",
       "Groups->User_has_Group"."createdAt" AS "_4",
       "Groups->User_has_Group"."updatedAt" AS "_5",
       "Groups->User_has_Group"."UserId"    AS "_6",
       "Groups->User_has_Group"."GroupId"   AS "_7"
FROM (SELECT "User"."user_id" AS "_0", "User"."createdAt", "User"."updatedAt"
      FROM "tbl_user" AS "User"
      LIMIT 1) AS "User"
       LEFT OUTER JOIN ("tbl_user_has_group" AS "Groups->User_has_Group" INNER JOIN "tbl_group" AS "Groups" ON
  "Groups"."group_id" = "Groups->User_has_Group"."GroupId") ON "User"."id" = "Groups->User_has_Group"."UserId";
```
The problem seems to come from the `ON "User"."id" ` that doesn't match the alias given to the column in the previous nested JOIN (`"User"."user_id" AS "_0"`).

In order to solve this kind of problem I need to store more information on each minifying I do and try to find every place where aliases are referenced and replace these occurrences as well.

This might take some time 😕

papb:
@mrrinot Thanks for looking! I would help but I don't have time at the moment. Good luck!

cyrilchapon:
@mrrinot I can help, but I dunno where to start.
Any chance you can give me the direction ?

mrrinot:
I thought about it some more and found a potential solution:

If I store more information about every minifying in order to store both the replaced alias and the name of the field that is being aliased like so
```js
    const mappings = [
        {
            _0: {
                aliased: '"id"',
                fullAlias: '"User"."id"'
            },
        },
    ];
```

I thought I could just String.replace() every `fullAlias` found in the sql query with their respective `_XX`. (`aliased` is there to do the permutation after the results come back)
I can't find any problem with this approach so I'll probably try it out this week-end if I got the time.

I think this is way easier than going about in all the places aliased fields could be referenced and changing them manually.

mrrinot:
@sushantdhiman I have a question regarding the `field` attribute used in the failing test that I am debugging at the moment.

The `User` table used in the test is defined like so: 
```js
        id: {
          type: DataTypes.UUID,
          allowNull: false,
          primaryKey: true,
          field: 'user_id'
        }
      }, {
        tableName: 'tbl_user'
      });
```

If I understand the documentation correctly, this means that the actual column name in the database is `user_id` and the `field` property sets up an alias to `id` for this field.

If that is correct, why is the join on the `User` table using `ON "User"."id"` ?

I think this behaviour is weird and while running the failing test suite without `minfiyAliases` I found an inconsistency.  
The test suite runs both a `findOne` and a `findAll` :
```js
  User.findOne({
              where: {},
              include: [Group]
            }),
  User.findAll({
              include: [Group]
            })
```

Those two queries should look identical but there is a differences on the `ON`mentioned above:
the `findOne` generates this: 
```sql 
 ON "User"."id" = "Groups->User_has_Group"."UserId";
```
whereas the `findAll` generates this: 
```sql
 ON "User"."user_id" = "Groups->User_has_Group"."UserId";
```

If I comment the `findOne` the query has no problem finding `"User"."user_id"` and runs with minified aliases.

Can you confirm that this is a bug unrelated to my PR ?

mrrinot:
**Update:** I tried various solutions based on the "find and replace occurrences of the alias in the query" but it ended up messing up at various points, mostly due to the fact that `WHERE` and other parts of a query do not use aliases but the original field name.

In trying to fix the `ON` from the comment above I ended up doing unwanted replacement.

One of the failing test present this case:
```js
        id3: {
          field: 'id',
          type: Sequelize.INTEGER,
          primaryKey: true
        },
        id: {
          field: 'id2',
          type: Sequelize.INTEGER,
          allowNull: false
        },
        id2: {
          field: 'id3',
          type: Sequelize.INTEGER,
          allowNull: false
        }
      });
```

Here my `replace` went berserk and didn't behave like I would have hoped.

Unless someone comes up with a variant of the `replace` solution, I am going to try the more painful approach and go into the code where aliases are being used (think `JOIN`, `GROUP BY` etc) and doing the permutation there instead of a lazy `replace`

I have no idea how long (if possible) this is going to take me. Luckily I have some time ahead of me but still...

papb:
Hi @mrrinot, thank you very much for taking the time. I can't look at it right now... I have one question though, since nobody answered your last question:

> Can you confirm that this is a bug unrelated to my PR ?

Are you assuming it is related or unrelated?

mrrinot:
I don't think it is related to my PR as the behaviour is present without the `minifyAlias` option.

I also do not think it is a bug, just a weird behaviour that didn't seem normal to me (the use/non-use of the alias in what seemed like two identical queries).

I didn't screen the whole query for the `findOne/findAll` but the `findOne` is done with a subquery and that subquery makes it so that the alias of the primaryKey is used instead of the `field` prop.

Hence my question to know if this is intended or not.

mrrinot:
I have a question concerning two [failing tests](https://github.com/sequelize/sequelize/blob/master/test/integration/dialects/postgres/dao.test.js#L206):
```
      -SELECT "id", "grappling_hook" AS "_0", "utilityBelt", "createdAt", "updatedAt" FROM "Equipment" AS "Equipment" WHERE CAST(("Equipment"."utilityBelt"#>>'{grapplingHook}') AS BOOLEAN) = true;
      +SELECT "id", "grappling_hook" AS "grapplingHook", "utilityBelt", "createdAt", "updatedAt" FROM "Equipment" AS "Equipment" WHERE CAST(("Equipment"."utilityBelt"#>>'{grapplingHook}') AS BOOLEAN) = true;
```

These tests fail due to being:
```js
expect(sql).to.equal();
```

How do I manage these tests ? Ignore them when `minifyAliases` is set ? Does mocha have snapshots ?

papb:
> I have a question concerning two [failing tests](https://github.com/sequelize/sequelize/blob/master/test/integration/dialects/postgres/dao.test.js#L206):
> 
> ```
>       -SELECT "id", "grappling_hook" AS "_0", "utilityBelt", "createdAt", "updatedAt" FROM "Equipment" AS "Equipment" WHERE CAST(("Equipment"."utilityBelt"#>>'{grapplingHook}') AS BOOLEAN) = true;
>       +SELECT "id", "grappling_hook" AS "grapplingHook", "utilityBelt", "createdAt", "updatedAt" FROM "Equipment" AS "Equipment" WHERE CAST(("Equipment"."utilityBelt"#>>'{grapplingHook}') AS BOOLEAN) = true;
> ```
> 
> These tests fail due to being:
> 
> ```js
> expect(sql).to.equal();
> ```
> 
> How do I manage these tests ? Ignore them when `minifyAliases` is set ? Does mocha have snapshots ?

@mrrinot I suggest replacing `expect(sql).to.equal` with `expect(sql).to.contain` and testing only for the end of the sql, which is the relevant part for that test. I am not sure @sushantdhiman would agree though.

sushantdhiman:
> **Update:** I tried various solutions based on the "find and replace occurrences of the alias in the query" but it ended up messing up at various points, mostly due to the fact that `WHERE` and other parts of a query do not use aliases but the original field name.
> 
> In trying to fix the `ON` from the comment above I ended up doing unwanted replacement.
> 
> One of the failing test present this case:
> 
> ```js
>  const User = this.sequelize.define('User', {
>         id3: {
>           field: 'id',
>           type: Sequelize.INTEGER,
>           primaryKey: true
>         },
>         id: {
>           field: 'id2',
>           type: Sequelize.INTEGER,
>           allowNull: false
>         },
>         id2: {
>           field: 'id3',
>           type: Sequelize.INTEGER,
>           allowNull: false
>         }
>       });
> ```
> 
> Here my `replace` went berserk and didn't behave like I would have hoped.
> 
> Unless someone comes up with a variant of the `replace` solution, I am going to try the more painful approach and go into the code where aliases are being used (think `JOIN`, `GROUP BY` etc) and doing the permutation there instead of a lazy `replace`
> 
> I have no idea how long (if possible) this is going to take me. Luckily I have some time ahead of me but still...

Both behavior are correct. `findAll` use inner join and `findOne` use outer join (for limit 1). They both require different logic to handle attribute naming collisions

mrrinot:
I decided to push my work so far because it reached a point where I need feedback on my choices so here goes:

*Problem:* aliased field where being mentioned elsewhere in queries and needed to be replaced with their respective alias (_0, _1 etc).

*Solution:* I first started to try to `.replace()` my way around this but this prompted problems described above so I went for the other option.

The other option was to go find all the places where aliased field where being used and retrieve the alias instead.

In order to do so easily I added another object to map the alias in the other way (and also add the information of the field's table)
This is a simple object like so: 
```js
  options.aliasesByTable[`${tableName}${alias}`] = minifiedAlias;
```
At first it was a double object `aliasesByTable[table][field]`, but this didn't provide any bonus other a simple object.

---

I then had to use this object at various places and chose to do so with a custom "private" function returning the alias for a given field or `null` if not found/not minifying.

This function looks very odd for a few reasons and most of them are due to `ORDER BY` being sophisticated.

The current version uses a huge chunk of code belonging to [this.quote()](https://github.com/sequelize/sequelize/blob/master/lib/dialects/abstract/query-generator.js#L724).

I encountered this [test case](https://github.com/sequelize/sequelize/blob/master/test/integration/dialects/postgres/dao.test.js#L1209) during debugging:
```js
  order: [
                ['id'],
                [this.Class, 'id'],
                [this.Class, this.Student, 'id']
              ]
```
The [code](https://github.com/sequelize/sequelize/pull/11095/files#diff-def52474024b12c7d133bb504cdf57d1R1629) copied from `this.quote()` is here to handle this case but no test case uses this kind of `ORDER BY` **and** aliasing the key being ordered. 

Do I need to add such a test ?

I feel like this is a rare edge case that could be fixed later on if need be.
I don't know if I should let this chunk of code stay or remove it all.

mrrinot:
the problem with this order: 
```js
  order: [
                ['id'],
                [this.Class, 'id'],
                [this.Class, this.Student, 'id']
              ]
```
is that my map table looks like this: 
```js
{
  'ClassesClasses.id': '_0',
  'ClassesClasses.name': '_1',
  'ClassesClasses.createdAt': '_2',
  'ClassesClasses.updatedAt': '_3',
  'ClassesClasses.ProfessorId': '_4',
  'Classes->StudentsClasses.Students.id': '_5',
  'Classes->StudentsClasses.Students.fullName': '_6',
  'Classes->StudentsClasses.Students.createdAt': '_7',
  'Classes->StudentsClasses.Students.updatedAt': '_8',
  'Classes->Students->ClassStudentClasses.Students.ClassStudent.createdAt': '_9',
  'Classes->Students->ClassStudentClasses.Students.ClassStudent.updatedAt': '_10',
  'Classes->Students->ClassStudentClasses.Students.ClassStudent.ClassId': '_11',
  'Classes->Students->ClassStudentClasses.Students.ClassStudent.StudentId': '_12'
}
```
The second part is being constructed by `this.quote()`: `Classes.Students.id`
But the tricky part is the first one, the `model` passed to my `getAliasForField` function is `Professor` and it will try to append both together and not find anything that way.

I would have to modify the copied code in order to also modify the `model` variable to be `Classes->Students` and I don't really know how to do that without copying another wall of code.

mrrinot:
I didn't end up doing the whole *order aliasing like `this.quote()`* because it isn't really a thing, you don't usually order on an aliased column if I understand correctly.

The two cases that needed aliasing were when ordering was done on a renamed field. Since I re-alias the alias, I needed to check for that but there is a specific `if` in the code for that, so that was clean.

The other case is the groupedLimit case where the order is done on another inner-alias: `subquery_order_X`.  
I tried to handle this case in the `getQueryOrders()` function but that ended up very tricky.

So I added a special case in the aliasing function to simply ignore any field matching the pattern 
```js
alias.match(/subquery_order_[0-9]/);
```
And that took care of all the aliasing problems on `ORDER BY`

mrrinot:
In the latest commit, I fixed a failing [test](https://github.com/sequelize/sequelize/blob/master/test/integration/sequelize.test.js#L1316) with a *harcoded* solution because the query is being constructed and run in a second time. The aliases are stored in the query's options so they aren't available in the *running* part and cannot revert `_0` back to `cnt`.

If you want me to do this another way, like ignore the test with the option or something else I didn't think of, tell me @sushantdhiman.

mrrinot:
Ok, so now all integration tests are good, there is a lot of unit tests failing.
Is there a way to run only the integration tests ? Do I create new test that check the aliasing system and run them separately ?

sushantdhiman:
@mrrinot Just integration tests should be enough, once build is green review this once again.

mrrinot:
@sushantdhiman I suppose I need to modify the travis.yml file but I don't know what to change.

.travis.yml
```diff
@@ -61,6 +61,10 @@ jobs:
       node_js: '6'
       sudo: required
       env: POSTGRES_VER=postgres-10 SEQ_PG_PORT=8991 DIALECT=postgres
+    - stage: test
+      node_js: '6'
+      sudo: required
+      env: POSTGRES_VER=postgres-10 SEQ_PG_PORT=8991 SEQ_PG_MINIFY_ALIASES=1 DIALECT=postgres
```
sushantdhiman:
```suggestion
      env: POSTGRES_VER=postgres-10 SEQ_PG_PORT=8991 SEQ_PG_MINIFY_ALIASES=1 DIALECT=postgres
      script:
        - npm run test-integration
```

mrrinot:
Tests appears OK @sushantdhiman, I don't understand the codecov report failing though.

lib/dialects/abstract/query-generator.js
```diff
@@ -1413,7 +1428,13 @@ class QueryGenerator {
...
           deprecations.noRawAttributes();
         }
-        attr = [attr[0], this.quoteIdentifier(attr[1])].join(' AS ');
+        let alias = attr[1];
+
+        if (this.options.minifyAliases) {
+          alias = this._getMinifiedAlias(alias, mainTableAs, options);
+        }
+
+        attr = [attr[0], this.quoteIdentifier(alias)].join(' AS ');
```
papb: Why is it necessary to minify something inside a `escapeAttributes` function?

mrrinot: This function is only used for the attributes of the main query ([here](https://github.com/sequelize/sequelize/pull/11095/files#diff-def52474024b12c7d133bb504cdf57d1R1168)).
I don't really know why this is a different function than the one used for includes or subQueries.

Telokis:
Any idea how much work is left before we get this feature?

sushantdhiman:
Coverage is down a bit but can ignore that

mjy78:
Can the minifyAliases option be passed to individual queries? eg. `Model.findAll({ where {...}, include: {...}, minifyAliases: true });`

mjy78:
> Can the minifyAliases option be passed to individual queries? eg. `Model.findAll({ where {...}, include: {...}, minifyAliases: true });`

Tested this and confirmed you cannot pass minifyAliases: false in for individual queries.

mrrinot:
No you cannot, I thought it would be more of a global options than a individual one.

dhruvinparikh:
If I pass `minifyAliases : false` , it will minify the output and does the vice-versa.

obo20:
This doesn't appear to be working for me. I'm on 0.5.18.

henrywoody:
I'm running into this issue but for the table names, rather than table + column names. Meaning that the names of the joined tables are already too long for Postgres, so I get an error saying the _table_ name is specified more than once.

Can we add handling for table names in addition to column names?

papb:
Hi @henrywoody, please open a new issue about this so we can track progress properly, thanks

Nick-Riggs:
Bug opened for minifyAliases not applying to the FROM clause aliases: https://github.com/sequelize/sequelize/issues/11937

Issue #4324, opened 2015-08-14
bulkCreate can't do "upsert" logic with postgres, even though normal "upsert" is supported by postgres?


GeoffreyPlitt:
How come bulkCreate can't use "upsert" logic with postgres? The normal "upsert" command supports Postgres.

janmeier:
Do you mean upsert as in `ON DUPLICATE KEY UPDATE `? That feature only exists in postgres 9.5, which isn't officially released yet. 

Right now single row upsert is handled using a function which catches the unique constraint error. It's not pretty :-). We're definitely going to switch to the new syntax once it's released officially, but not before we can test it on Travis

GeoffreyPlitt:
Here's what I'm currently doing:

```javascript
// True if the model has a non-autogenerated primary key
const has_actual_primary_key = (model) => {
  const attrs = Object.keys(model.attributes)
  const pk_fields = attrs.filter( (x) => {
    return model.attributes[x].primaryKey && !model.attributes[x]._autoGenerated
  })
  return pk_fields.length>0
}

// Uses true "upsert" if the model has a primary key, otherwise findOrCreate().
const bulk_upsert = (model, rows) => {
  return Promise.all(rows.map( (row) => {
    if(has_actual_primary_key(model)) {
      return model.upsert(row)
    } else {
      return model.findOrCreate({
        where: row,
        defaults: row
      })
    }
  }))
}

module.exports = bulk_upsert
```

janmeier:
Yeah, something along those lines is probably your best bet for now

janmeier:
Closing, works as intended until postgres 9.5 is out

diosney:
It seems that Postgres 9.5.1 is already out, maybe you can reopen the issue now?

mickhansen:
@diosney Is it on travis yet?

diosney:
It seems that not yet (latest version they have in the docs are 9.4.1) :(

I will leave here the links for future reference:

https://docs.travis-ci.com/user/using-postgresql/
https://github.com/travis-ci/travis-ci/issues/4264

prasadbhosale:
Postgres 9.5 is already out . You can reopen this issue now ?

janmeier:
but its not yet easily available on travis https://github.com/travis-ci/travis-ci/issues/4264

bbnnt:
It is now available; isn't it ?

andrewgbliss:
+1

charlesdeuter:
+1

[...]

PR #11163, opened 2019-07-08
feat(postgres): enabling the updateOnDuplicate BulkCreate option

thaddeusmt:
### Pull Request check-list

### Description of change

I have attempted to enable the `BulkCreate()` `updateOnDuplicate` option for the Postgres dialect using the `INSERT ... ON CONFLICT DO NOTHING/UPDATE` syntax added in Postgres 9.5. This essentially enables "upsert" functionality on "BulkCreate" operations (as is currently provided for the MySql and Mariadb dialects with the `ON DUPLICATE KEY UPDATE` functionality).

NOTE: The Postgres `upsert()` functionality still uses the `INSERT EXCEPTION WHEN` method, I did not refactor that to use `ON CONFLICT`.

The Postgres `ON CONFLICT` command requires either listing out the columns to key conflicts on, or using an actual unique constraint by name. This easiest way I could see to support both primary key and "unique" constraints in the model definitions was to add some logic in the `bulkInsertQuery()` method that looks up those columns, and then add to the `ON CONFLICT` command directly. By default it uses the primary key columns, but if there is a separate unique constraint in the model definition it will use that instead.

I added an integration test called `should support the updateOnDuplicate option with primary keys` to test the primary key case, since the existing `should support the updateOnDuplicate option` test only checked unique constraints. The unit and integration tests are passing for Postgres.

I am only set up to run the tests with the Postgres dialect right now, so hopefully I didn't break any other dialects... 

I think this might fix issues #4132 and #6325 and #3354 and #4324.

It appears this outstanding PR #6325 was attempting to do this as well.

This is my first attempt at a PR on a project like this, please advise on how I can improve my submission, thank you.


lib/dialects/mariadb/index.js
```diff
@@ -28,7 +28,8 @@ MariadbDialect.prototype.supports = _.merge(
     schemas: true,
     inserts: {
       ignoreDuplicates: ' IGNORE',
-      updateOnDuplicate: true
+      updateOnDuplicate: ' ON DUPLICATE KEY UPDATE'
+      //updateOnDuplicate: true // now using a string instead of boolean, like ignoreDuplicates
```
sushantdhiman: Remove these comments

lib/query-interface.js
```diff
@@ -983,17 +983,26 @@ class QueryInterface {
...
+   * @param {Model}  model       Model class
    * @param {string} tableName   Table name to insert record to
    * @param {Array}  records     List of records to insert
    * @param {Object} options     Various options, please see Model.bulkCreate options
    * @param {Object} attributes  Various attributes mapped by field name
    *
    * @returns {Promise}
    */
-  bulkInsert(tableName, records, options, attributes) {
+  bulkInsert(model, tableName, records, options, attributes) {
```
sushantdhiman: This changes the signature of this method which is part of public api. Caller should pass `options.upsertKeys` directly, add this new option `options.upsertKeys` to api docs and typings

lib/dialects/abstract/query-generator.js
```diff
@@ -300,10 +300,20 @@ class QueryGenerator {
...
 
     if (this._dialect.supports.inserts.updateOnDuplicate && options.updateOnDuplicate) {
-      onDuplicateKeyUpdate = ` ON DUPLICATE KEY UPDATE ${options.updateOnDuplicate.map(attr => {
-        const key = this.quoteIdentifier(attr);
-        return `${key}=VALUES(${key})`;
-      }).join(',')}`;
+      if (this._dialect.supports.inserts.updateOnDuplicate == ' ON CONFLICT DO UPDATE SET') { // postgres
+        // If no conflict target columns were specified, use the primary key names from options.upsertKeys
+        onDuplicateKeyUpdate = ` ON CONFLICT (${options.upsertKeys.map(attr => {
+          return `"${attr}"`;
```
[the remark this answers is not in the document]
sushantdhiman: Also use `this.quoteIdentifier(<attr>)` instead of `"<attr>"`

lib/dialects/mariadb/index.js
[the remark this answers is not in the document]
thaddeusmt: I removed these in commit 8381105. Any guidance on how to update the changelog? Is there a manually generated changelog in the docs somewhere? Thanks

thaddeusmt: I removed these in commit 8381105.

lib/query-interface.js
thaddeusmt: I think I figured out a way to do this without changing the method signature in commit 8381105. I moved the logic that I needed the `model` for in to `Model.bulkCreate()` instead of `bulkInsert()`.

lib/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
thaddeusmt: In commit 8381105 I updated with `quoteIdentifier()` (oops). 

I am not sure how best to inline this though? Especially since I added the `quoteIdentifier()` call? My JS-fu is weak, there is probably some classic syntax I am not familiar with. I was following the existing code example which looks like this:

```
onDuplicateKeyUpdate = `${this._dialect.supports.inserts.updateOnDuplicate} ${options.updateOnDuplicate.map(attr => {
  const key = this.quoteIdentifier(attr);
  return `${key}=VALUES(${key})`;
}).join(',')}`;
```

thaddeusmt:
What's the next step? Do I just need to update the Documentation? Or is there anything else I should improve? Thanks

lib/dialects/abstract/query-generator.js
```diff
@@ -300,10 +300,21 @@ class QueryGenerator {
     }
 
     if (this._dialect.supports.inserts.updateOnDuplicate && options.updateOnDuplicate) {
-      onDuplicateKeyUpdate = ` ON DUPLICATE KEY UPDATE ${options.updateOnDuplicate.map(attr => {
-        const key = this.quoteIdentifier(attr);
-        return `${key}=VALUES(${key})`;
-      }).join(',')}`;
+      if (this._dialect.supports.inserts.updateOnDuplicate == ' ON CONFLICT DO UPDATE SET') { // postgres
```
sushantdhiman: Replace this with (modify if anything missing)

```js
if (this._dialect.supports.inserts.updateOnDuplicate == ' ON CONFLICT DO UPDATE SET') { // postgres
  onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) DO UPDATE SET ${updateKeys.join(',')}`;
} else { // mysql / maria
  const valueKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=VALUES(${this.quoteIdentifier(attr)})`);
  onDuplicateKeyUpdate = `${this._dialect.supports.inserts.updateOnDuplicate} ${valueKeys.join(',')}`;
}
```

lib/dialects/postgres/index.js
```diff
@@ -42,7 +42,9 @@ PostgresDialect.prototype.supports = _.merge(_.cloneDeep(AbstractDialect.prototy
     functionBased: true
   },
   inserts: {
-    onConflictDoNothing: ' ON CONFLICT DO NOTHING'
+    onConflictDoNothing: ' ON CONFLICT DO NOTHING',
+    updateOnDuplicate: ' ON CONFLICT DO UPDATE SET'
+    // updateOnDuplicate: true // this is why mysql does
```
sushantdhiman: Remove this

papb:
@thaddeusmt Hello! Thanks for the PR! Please update the docs: [here](https://github.com/sequelize/sequelize/blob/715a82fd0278672d665386a34d0684cdc7dbeb68/lib/model.js#L2520) and [here](https://github.com/sequelize/sequelize/blob/715a82fd0278672d665386a34d0684cdc7dbeb68/types/lib/model.d.ts#L741).

bmeriaux:
the same refactor on the upsert method and it will be a major improvement for postgres insert performance in sequelize, nice work!

lib/dialects/abstract/query-generator.js
thaddeusmt: Much cleaner, thanks for the coding tips :+1:

lib/dialects/postgres/index.js
thaddeusmt: Removed the comment

thaddeusmt:
> @thaddeusmt Hello! Thanks for the PR! Please update the docs: [here](https://github.com/sequelize/sequelize/blob/715a82fd0278672d665386a34d0684cdc7dbeb68/lib/model.js#L2520) and [here](https://github.com/sequelize/sequelize/blob/715a82fd0278672d665386a34d0684cdc7dbeb68/types/lib/model.d.ts#L741).

I missed those JSDoc comments, thanks. I updated them to add that Postgres 9.5 is supported.

thaddeusmt:
> the same refactor on the upsert method and it will be a major improvement for postgres insert performance in sequelize, nice work!

I have thought about replacing the Postgres `upsert` implementation to use `ON CONFLICT` instead of `INSERT EXCEPTION WHEN` - if I do that I think I'll open a new PR, so if it takes me a long time to motivate this one isn't blocked.

papb:
@thaddeusmt Hi, thanks for the changes. I have one more request. I am bothered by the specific, explicit mention of Postgres **9.5**. This makes it seem that no other version whatsoever is supported. Is this true? Is there a version of Postgres that is not supported?

sushantdhiman:
`ON CONFLICT` is not supported by versions below `9.5`

papb:
@sushantdhiman Ah, so by 9.5 it is meant "9.5 and above"? I think that deserves clarification

sushantdhiman:
> Ah, so by 9.5 it is meant "9.5 and above"? I think that deserves clarification

Docs updated

prachikhadke:
Been trying to use this but keep getting an error `Unhandled rejection SequelizeDatabaseError: syntax error at or near ";"` ` DO UPDATE SET ;`. Any idea what could be the issue?

papb:
@prachikhadke You should open a new issue with a SSCCE/MCVE/reprex so we can help you :+1:

thaddeusmt:
@prachikhadke Link me to (or tag me in) an Issue with some details for reproducing and I'll take a look. I'm curious:

- What does the Sequelize model init/creation look like (how are you defining Primary and Unique keys)
- What does your `BulkCreate()` call look like?

Thanks

prachikhadke:
`ObjectModel.bulkCreate(objectsToInsertOrUpdate, { transaction, returning: false })`

Where `ObjectModel` defines the object model. There is only one primary key field - an id. No other unique keys.

thaddeusmt:
@prachikhadke The first thing that jumps out is that you are not setting `updateOnDuplicate: true` on the `bulkCreate()` options? That enables the feature in this pull request. But perhaps this bug is that I broke the default non-update functionality?

It might still be useful to see how the model is created. Is the primary key `id` automatically defined, or is it defined with `primaryKey: true` in the column definition?  There could be some case I missed where primary key is defined in a way that it doesn't appear when I query `model.primaryKeys` (and `model.uniqueKeys`) to build the `ON CONFLICT` SQL clause.

prachikhadke:
@thaddeusmt I did try setting the `updateOnDuplicate: true` option after upgrading to the latest sequelize version. That's when I got the error. The id field does have the `primaryKey: true`. The primary key is not being set using a sequence but is set as part of the record.

papb:
@prachikhadke Please open a new bug report for this, instead of discussing here.

Issue #10523, opened 2019-03-08
Allow to specify returning fields instead of 'RETURNING *' for PostgreSQL


Mikhus:
## What are you doing?
I'm trying to update or insert a record, but I want to return only specified fields. For example:
```js
// somewhere in my create user function
// (this case could be valuable for update as well)
const user = new User(createData);
// that's what I want to have
user.save({ returning: ['id', 'name', 'email'] });
// instead of 
user.save({ returning: true });
```

## What do you expect to happen?
I want my created/updated model to have fields defined only those which being specified within returning option of the save() operation. Actually, PostgreSQL allows me to do that, like
```sql
INSERT INTO "User" ("id", "name", "lastName", "email", "password")
VALUES (DEFAULT, 'John', 'Doe', 'john@doe.com', 'xxxxxx')
RETURNING id, name, email;
```

## What is actually happening?
Database will return whole set of fields for the user which are populated to the model, so after I serialize and send the model data through the network it will contain even those fields which I don't need.

__Dialect:__ postgres 
__Dialect version:__ any
__Database version:__ 9.6
__Sequelize version:__ 4.42.0
__Tested with latest release:__ No

**Note :** _Your issue may be ignored OR closed by maintainers if it's not tested against latest version OR does not follow issue template._

sushantdhiman:
Those extra fields aren't defined in model?

Mikhus:
Defined. But when I use 
```js
User.findByPk(42, { attributes: ['id', 'name', 'email'] });
```
I get exactly what I need. when serialize it it will contain only those fields which was requested to have. It could be used as a workaround (but will cause an extra query execution) or i would need to use 'returning *' and then handle extra fields removal manually after serialization... I think specifying the fields I need within returning statement is a proper improvement for orm, as far, as datavase itself works exactly in this way, but orm does not reflect such behavior correctly

thurt:
This would be a nice feature to have. I feel like scopes should be utilized in this problem as well. For example, if I have a `defaultScope` in my model then the `returning: true` option should respect that by returning only the fields specified in the `defaultScope`

Mikhus:
Working through `sequelize-typescript` package I've been able to achieve this overriding `Sequrlize.query()` method in a following way:

```typescript
import { Sequelize as SequelizeOrigin, Model } from 'sequelize-typescript';

export class Sequelize extends SequelizeOrigin {
    public query(
        sql: string | { query: string, values: any[] },
[...]

PR #11170, opened 2019-07-09
feat: specify fields to return for bulkCreate

mikesadowski12:
### Pull Request check-list

### Description of change
Allow to specify fields for the returning clause on bulkCreate()
https://github.com/sequelize/sequelize/issues/10523


mikesadowski12:
> Needs typings and docs.

Updated the docs to explain the new behaviour.

What is meant by 'typings?'

SimonSchick:
You need to update the jsdoc tags and the typescript definitions, both should be relatively trivial.

mikesadowski12:
> You need to update the jsdoc tags and the typescript definitions, both should be relatively trivial.

Updated the JSDocs, still not quite sure what the `typescript definitions` are.

papb:
@mikesadowski12 You have to update the relevant TypeScript files present in the typings folder of the project (`sequelize/types/lib`).

types/lib/model.d.ts
```diff
@@ -743,7 +743,7 @@ export interface BulkCreateOptions extends Logging, Transactionable {
   /**
    * Return the affected rows (only for postgres)
    */
-  returning?: boolean;
+  returning?: boolean | Array<string>;
```
SimonSchick: Use `string[]` instead.

lib/model.js
```diff
@@ -2521,7 +2521,7 @@ class Model {
    * @param  {Transaction}  [options.transaction]            Transaction to run query under
    * @param  {Function}     [options.logging=false]          A function that gets executed while running the query to log the sql.
    * @param  {boolean}      [options.benchmark=false]        Pass query execution time in milliseconds as second argument to logging function (options.logging).
-   * @param  {boolean}      [options.returning=false]        Append RETURNING * to get back auto generated values (Postgres only)
+   * @param  {boolean|Array} [options.returning=false]       If true, append RETURNING * to get back all values; if an array of column names, append RETURNING <columns> to get back specific columns (Postgres only)
```
sushantdhiman: This array should be mapped to their field names. This will allow users to pass attributes like `createdAt` and we will query correct column `created_at` etc. Something like this https://github.com/sequelize/sequelize/blob/1cffab70382c5367fbb6ff611d11056f6410867c/lib/model.js#L2646

lib/dialects/abstract/query-generator.js
```diff
@@ -308,8 +308,15 @@ class QueryGenerator {
 
     const ignoreDuplicates = options.ignoreDuplicates ? this._dialect.supports.inserts.ignoreDuplicates : '';
     const attributes = allAttributes.map(attr => this.quoteIdentifier(attr)).join(',');
-    const returning = this._dialect.supports.returnValues && options.returning ? ' RETURNING *' : '';
     const onConflictDoNothing = options.ignoreDuplicates ? this._dialect.supports.inserts.onConflictDoNothing : '';
+    let returning = '';
+
+    if (Array.isArray(options.returning)) {
+      const fields = options.returning.map(field => this.quoteIdentifier(field)).join(',');
+      returning += ` RETURNING ${fields}`;
```
papb: What if the dialect does not support this, i.e., what if `this._dialect.supports.returnValues` is false? Wouldn't this generate a broken query?

You probably want to add a check for `this._dialect.supports.returnValues` here as well, right?

docs/instances.md
```diff
@@ -170,6 +170,32 @@ User.bulkCreate([
...
+User.bulkCreate([
+  { username: 'barfooz', isAdmin: true },
+  { username: 'foo', isAdmin: true },
+  { username: 'bar', isAdmin: false }
+], { returning: ['username'] }) // will return only the specified columns for each row inserted
+.then((result) => {
+  console.log(result);
+});
+```
+
```
papb: Excellent addition, please just clarify that this is available for postgres only

lib/model.js
papb: @mikesadowski12 Please address sushantdhiman's comment as well :+1:

mikesadowski12: just addressed it, although i am not entirely sure i did what he meant. let me know if i need to do anything else.

docs/instances.md
```diff
@@ -170,6 +170,32 @@ User.bulkCreate([
 })
 ```
 
+Insert several rows and return all columns:
```
papb: This part is also Postgres only, isn't it?

lib/model.js
papb: @mikesadowski12 You did right :+1:

Just add a test to ensure this is followed, i.e., a test in which you pass `createdAt` and the generated query contains `created_at` (and not `createdAt`, which is what it would have if you didn't make this last change) (of course, you need the model to have `field: "created_at"` or at least `underscored: true`)

Let me know if you have any doubts

docs/returning.md
```diff
@@ -0,0 +1,12 @@
+# Returning
```
sushantdhiman: IMO we don't need these docs at all

lib/model.js
sushantdhiman: PR looks good to go in current state

docs/returning.md
papb: I agree, also this file as is doesn't appear in the docs anyway, just remove this file

lib/model.js
mikesadowski12: should i add these tests then or is it ready to go with the removal of that file?

papb: @mikesadowski12 Ready to go, I believe :+1:

papb:
LGTM

mikesadowski12:
awesome, thanks guys :)

PR #11172, opened 2019-07-09
Add option for forcing function creation (fixes #11165)

### LINKED ISSUE #11165: `queryInterface.createFunction` should default to `CREATE OR REPLACE` syntax.
## What are you doing?

```js
    queryInterface.createFunction(
            'is_timezone',
            [{ type: 'TEXT', name: 'tz', direction: 'IN' }],
            'BOOLEAN',
            'plpgsql',
            `
                PERFORM now() AT TIME ZONE tz;
                RETURN TRUE;
                EXCEPTION WHEN invalid_parameter_value THEN
                    RETURN FALSE;
            `,
            ['IMMUTABLE', 'LEAKPROOF'],
        );
```
**To Reproduce**
1. Call the above function twice, and you'll get an error

Expecting following sql to run:
```sql
CREATE OR REPLACE FUNCTION is_timezone(IN tz TEXT) RETURNS BOOLEAN AS $func$ BEGIN 
  PERFORM now() AT TIME ZONE tz;
  RETURN TRUE;
  EXCEPTION WHEN invalid_parameter_value THEN
    RETURN FALSE;
END; $func$ language 'plpgsql'IMMUTABLE LEAKPROOF;
```

```sql
CREATE FUNCTION is_timezone(IN tz TEXT) RETURNS BOOLEAN AS $func$ BEGIN 
  PERFORM now() AT TIME ZONE tz;
  RETURN TRUE;
  EXCEPTION WHEN invalid_parameter_value THEN
    RETURN FALSE;
END; $func$ language 'plpgsql'IMMUTABLE LEAKPROOF;
```
(Note the lack of `OR REPLACE` syntax)

## Environment
Dialect:
- [ ] mysql
- [x] postgres
- [ ] sqlite
- [ ] mssql
- [ ] any
Dialect **library** version: ^7.9.0
Database version: 9.6
Sequelize version: ^5.3.5
Node Version: 8.x
OS: Linux
Tested with latest release:
- [x] No
- [ ] Yes, specify that version:

mannol:
### Pull Request check-list

### Description of change
This adds an option to enable function replacement as described in #11165.

### DIFF
--- a/types/lib/query-interface.d.ts
+++ b/types/lib/query-interface.d.ts
@@ -83,6 +83,10 @@ export interface QueryOptionsWithType<T extends QueryTypes> extends QueryOptions
   type: T;
 }
 
+export interface QueryOptionsWithForce extends QueryOptions {
+  force?: boolean;
+}
+
 /**
 * Most of the methods accept options and use only the logger property of the options. That's why the most used
 * interface type for options in a method is separated here as another interface.
@@ -553,7 +557,8 @@ export class QueryInterface {
     returnType: string,
     language: string,
     body: string,
-    options?: QueryOptions
+    optionsArray?: string[],
+    options?: QueryOptionsWithForce
   ): Promise<void>;
 
   /**
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -600,13 +600,15 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
     return `ALTER TRIGGER ${this.quoteIdentifier(oldTriggerName)} ON ${this.quoteTable(tableName)} RENAME TO ${this.quoteIdentifier(newTriggerName)};`;
   }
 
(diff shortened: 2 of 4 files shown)


mannol:
Some unrelated test failed ^. Can anyone retrigger?

papb:
Nice work! Don't worry about the flaky test.

types/lib/query-interface.d.ts
```diff
@@ -553,7 +557,8 @@ export class QueryInterface {
     returnType: string,
     language: string,
     body: string,
-    options?: QueryOptions
+    optionsArray?: string[],
+    options?: QueryOptionsWithForce
```
papb: Looks like you also found (and fixed!) a bug in typings, right? Where it was previously `QueryOptions` you fixed it to `string[]`. Nice work.

mannol: Yeah :+1:

PR #11232, opened 2019-07-26
chore(.github): ignored issue template reply

papb:
My idea is to standardize the reply we shall use for issues that don't follow the template. Just to make it easy for us to copy-paste a standard response instead of writing it down every time.

I based it on the latest one used by @sushantdhiman, with minor changes.

What do you think? @sushantdhiman @SimonSchick @eseliger 

I also plan on updating the Issue Template to clarify the need for SSCCE/MCVEs.


sushantdhiman:
You can save this template in saved replies

image

I don't think we should include this in repository itself.

papb:
Ohh, I didn't know about this feature! Thanks. But apparently we can't have a "shared" saved reply for all of us, can we? I think it would be nice if we all used the same reply... And could suggest improvements if needed.

For instance, some links on your saved reply are broken :grimacing:

sushantdhiman:
Thanks I will update links in my saved reply.

There is no need for this to be standardized, I think it is better to personalize this as per one's taste :)

sushantdhiman:
Main idea behind this is to quickly click button, choose a reply and close the issue. If we want to follow a standard reply we will have to open that page in repo, copy text and paste stuff and then close the issue. Saved reply feels quicker to me

papb:
Understood :+1:

PR #11234, opened 2019-07-26
fix(#11233): lodash ^4.17.15 to prevent potential vulnerability

### LINKED ISSUE #11233: [Security] Recommended upgrade of lodash
- Per https://github.com/lodash/lodash/issues/4348, an upgrade is required to address a recent security vulnerability CVE-2019-10744
- The modules specifically reported as vulnerable are used in this repo, example at https://github.com/sequelize/sequelize/blob/f2b0aec14c98691c425c59a005c2595419da6e4f/lib/utils.js#L53
- https://github.com/sequelize/sequelize/blob/master/package.json#L38 locks lodash at `^4.17.11`. While it is possible for projects listing this as a dependent to locally upgrade the version to a more secure version, a fresh install may still pick up the vulnerable version.

I'll go on and submit a PR for this, if you decide you don't want to apply it that is fine but this upgrade is recommended for the reasons above.

quetzaluz:
### Description of change
- Fixes #11233, please see description there

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "debug": "^4.1.1",
     "dottie": "^2.0.0",
     "inflection": "1.12.0",
-    "lodash": "^4.17.11",
+    "lodash": "^4.17.15",
     "moment": "^2.24.0",
     "moment-timezone": "^0.5.21",
     "retry-as-promised": "^3.1.0",


papb:
This is not necessary, [see my comment on the issue](https://github.com/sequelize/sequelize/issues/11233#issuecomment-515581549).

Issue #11151, opened 2019-07-05
Bugfix invalid query log not bind parameters in sequelize@5.9.0


ryuki0628:
version: sequelize@5.9.0

Invalid query log not bind parameters in sequelize@5.9.0

## What are you doing?
use query logging.
execute update method.

## What do you expect to happen?
output sql log
```
Executing (default): UPDATE users SET email = 'test@test.com', updated_at = '2019-01-01 10:00:00' WHERE id = 1;
```

## What is actually happening?
query log not bind parameters

output sql log
```
Executing (default): UPDATE users SET email = ?, updated_at = ? WHERE id = ?;
```

## Environment
Dialect:
- [x] mysql
- [ ] postgres
- [ ] sqlite
- [ ] mssql
- [x] any
Database version: 5.7
Sequelize version: 5.9.0

If TypeScript related: TypeScript version: XXX
Tested with latest release:
- [ ] No
- [x] Yes, specify that version: 

## How to modify
target file path: lib/dialects/mysql/query.js
line: 37
```js
    const complete = this._logQuery(parameters ? connection.format(sql, parameters) : sql, debug);
```

sushantdhiman:
Fixed by https://github.com/sequelize/sequelize/pull/11267

soufian044:
I'm still seeing the same issue with the "?" in the logs for the UPDATE 
using the latest version 5.19.0

![image](https://user-images.githubusercontent.com/10144036/65225178-f0ce2e00-dac4-11e9-8fa5-b7ec489b33ee.png)

sushantdhiman:
@soufian044 You will need to pass `logQueryParameters: true` to your Sequelize constructor options to enable parameter logging

soufian044:
@sushantdhiman  thank you, i would suggest to update the [API Reference for the Sequelize constructor](https://sequelize.org/master/class/lib/sequelize.js~Sequelize.html#instance-constructor-constructor).

sushantdhiman:
@soufian044 It will automatically update in 24 hours

Nogias9x:
Hi @sushantdhiman ,

Can we somehow have it 
```sql
UPDATE `users` SET `age`=10;
```
instead of
```sql
UPDATE `users` SET `name`=?; 10
```

I've having trailing params while using MySQL

Issue #10067, opened 2018-10-23
Invalid logging of single INSERT and UPDATE SQL queries


ezze:
## What are you doing?

Here is a [repo reproducing the issue](https://github.com/ezze/sequelize-log-sql-issue).

In Sequelize `5.0.0-beta.13`:

```js
const connection = new Sequelize(database, username, password, {
    dialect,
    host,
    define: {
        freezeTableName: true,
        underscored: false,
        createdAt: 'createdAt',
        updatedAt: 'updatedAt'
    },
    logging: message => {
        console.log(message);
    }
});

const Task = connection.define('task', {
    id: {
        type: Sequelize.INTEGER,
        primaryKey: true,
        autoIncrement: true
    },
    createdAt: {
        field: 'created_at',
        type: Sequelize.DATE,
        allowNull: false
    },
    updatedAt: {
        field: 'updated_at',
        type: Sequelize.DATE,
        allowNull: false
    },
    name: {
        type: Sequelize.TEXT,
        allowNull: false
    },
    status: {
        type: Sequelize.INTEGER,
        allowNull: false,
        defaultValue: 0
    }
});

(async() => {
    ...
    await Task.create({ name: 'First task', status: 1 }, { transaction });
    await Task.update({ status: 2 }, { where: { status: 1 }, transaction });
})();
```

## What do you expect to happen?

Output valid SQL queries with inserted binded parameters' values in console:

```sql
INSERT INTO "task" ("id","created_at","updated_at","name","status") VALUES (DEFAULT,'2018-10-23 13:58:05.380 +00:00','2018-10-23 13:58:05.380 +00:00','First task',1) RETURNING *
UPDATE "task" SET "status"=2,"updated_at"='2018-10-23 13:58:05.398 +00:00' WHERE "status" = 1
```

## What is actually happening?

I get SQL queries with tokens starting with `$` instead of parameters' values:

```sql
INSERT INTO "task" ("id","created_at","updated_at","name","status") VALUES (DEFAULT,$1,$2,$3,$4) RETURNING *;
UPDATE "task" SET "status"=$1,"updated_at"=$2 WHERE "status" = $3
```

Note that `bulkCreate` method works fine.

__Dialect:__ postgres
__Dialect version:__ 7.5.0
__Database version:__ 9.5.14
__Sequelize version:__ 5.0.0-beta.13
__Tested with latest release:__ Yes (latest beta version 5.0.0-beta.13)

sushantdhiman:
[...]

PR #11238, opened 2019-07-28
fix(log): show bind parameters in logs instead of placeholders

supperchong:
Fixes https://github.com/sequelize/sequelize/issues/11151
Fixes https://github.com/sequelize/sequelize/issues/10067#

### Pull Request check-list


papb:
Thanks! Some tests are failing though, can you take a look?

supperchong:
@papb  Yeah!Thanks for your review, I will  fix it.

papb:
Tests are still failing, but you probably know that. Just saying in case you missed it. (Take your time! No rush!)

supperchong:
@papb Thanks for your caring!:smile:

papb:
We now have two equivalent PRs, what to do? Perhaps you two can join forces instead?

#11159 #11238

lib/dialects/abstract/query.js
```diff
@@ -86,7 +86,10 @@ class AbstractQuery {
 
     sql = sql.replace(/\$(\$|\w+)/g, (match, key) => {
       if ('$' === key) {
-        return options.skipUnescape ? match : key;
+        if (!['postgres', 'sqlite'].includes(dialect)) {
+          return options.skipUnescape ? match : key;
+        } 
+        return match;
```
papb: Hello, just out of curiosity, why was this change necessary? (I'm a maintainer but I'm sure you understand this part much better than I do)

test/integration/model.test.js
```diff
@@ -934,7 +934,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
               if (dialect === 'mssql') {
                 expect(sql).to.not.contain('createdAt');
               } else {
-                expect(sql).to.match(/UPDATE\s+[`"]+User1s[`"]+\s+SET\s+[`"]+secretValue[`"]=(\$1|\?),[`"]+updatedAt[`"]+=(\$2|\?)\s+WHERE [`"]+id[`"]+\s=\s(\$3|\?)/);
+                expect(sql).to.match(/Executing \(default\):\sUPDATE\s+[`"]+User1s[`"]+\s+SET\s+[`"]+secretValue[`"]='?(\d+)'?,[`"]+updatedAt[`"]+=('?\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}:\d{2}(\.\d{3})?(\s\+\d{2}:\d{2})?'?)\s+WHERE [`"]+id[`"]+\s=\s(\d+)/);
```
papb: I believe you could remove this part `Executing \(default\):\s` in the beginning and the regex would still match, right? If yes, please do it.

test/integration/model/create.test.js
```diff
@@ -931,7 +931,26 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+          myvals: { type: Sequelize.ARRAY(Sequelize.INTEGER) },
+          mystr: { type: Sequelize.ARRAY(Sequelize.STRING) }
+        });
+
+        let test = false;
+        return User.sync({ force: true }).then(() => {
+          return User.create({ myvals: [1, 2, 3], mystr: ['One', 'Two'] }, {
+            logging(sql) {
+              test = true;
+              expect(sql).to.contain('INSERT INTO "UserWithArrays" ("id","myvals","mystr","createdAt","updatedAt") VALUES (DEFAULT,ARRAY[1,2,3],ARRAY[\'One\',\'Two\']');
```
papb: Here you are checking an incomplete SQL (abruptly ended at `ARRAY[\'One\',\'Two\']`), I think this could cause confusion. Please assert on the complete SQL. I suggest updating the `User` model you create in this test to not have timestamps (`{ timestamps: false }`) so that you don't have to go over the hassle of asserting the `createdAt`/`updatedAt` fields.

test/integration/model/create.test.js
```diff
@@ -942,15 +961,15 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           return User.create({ myvals: [], mystr: [] }, {
             logging(sql) {
               test = true;
-              expect(sql).to.contain('INSERT INTO "UserWithArrays" ("id","myvals","mystr","createdAt","updatedAt") VALUES (DEFAULT,$1,$2,$3,$4)');
+              expect(sql).to.contain('INSERT INTO "UserWithArrays" ("id","myvals","mystr","createdAt","updatedAt") VALUES (DEFAULT,\'{}\',\'{}\'');
```
papb: Same thing as above, please assert on the complete SQL, avoid timestamps with `{ timestamps: false }` in the model.

test/integration/model/create.test.js
```diff
@@ -964,7 +983,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
             return user.save({
               logging(sql) {
                 test = true;
-                expect(sql).to.contain('UPDATE "UserWithArrays" SET "myvals"=$1,"mystr"=$2,"updatedAt"=$3 WHERE "id" = $4');
+                expect(sql).to.contain('UPDATE "UserWithArrays" SET "myvals"=\'{}\',"mystr"=\'{}\'');
```
papb: Same thing as above, please assert on the complete SQL, avoid timestamps with `{ timestamps: false }` in the model.

sushantdhiman:
Do we require all this, I think we can just append parameters after queries. I have seen this format used with bind queries before

```sql
Executing(default): SELECT * FROM "table" WHERE "name" = ?; test;
```

lib/dialects/abstract/query.js
supperchong: Because $ is used to escape a literal $ sign. If we unescape `$$1`  with `$1` , then later we replace logSql with bind Parameter, the  `$1` will be `${Parameter}`, but actual expect `$1`.
 For example we have a query   
`sequelize.query('SELECT *, "text with literal $$1 and literal $$status" as t FROM projects WHERE status = $1',
  { bind: ['active'], type: sequelize.QueryTypes.SELECT }
)`  
the log sql will be  
`SELECT *, "text with literal 'active' and literal $status" as t FROM projects WHERE status = 'active'`
but expect   
`SELECT *, "text with literal $1 and literal $status" as t FROM projects WHERE status = 'active'`.

test/integration/model.test.js
supperchong: Yeah, great.

test/integration/model/create.test.js
supperchong: Good idea!

supperchong:
@sushantdhiman  Yeah.this would be more easier and uniform. Is there a plan to do this?   
Moreover there is  a little trouble that people should know how module(like [pg](https://github.com/brianc/node-postgres)) hanlde parameters occasionally.

papb:
This PR looks ready to merge for me! :grimacing: Great work, @supperchong!

@sushantdhiman That makes sense, I dunno why we didn't think of that before. Regardless, I think this PR is great as is awyway. I think we should merge it, and perhaps on another occasion we can change back to your suggestion, in another PR. What do you think?

sushantdhiman:
@papb This PR is a good effort but I think we should not merge this because 1) There is new dependency on `sql-string` now 2) There is lots of extra code just to display query logs.

@supperchong You should pass [`_logQuery`](https://github.com/sequelize/sequelize/blob/f3b5f99b342cb5d47865daa49a2aa6316064507c/lib/dialects/abstract/query.js#L330) a third argument `parameters`. Then in that function just append parameters to query if they are not empty. You may continue in this PR or start a new one.

papb:
> 1) There is new dependency on sql-string

This is not true, is it? This is true for the other PR (#11159), not this one.

papb:
Regardless, I think I am convinced. Your idea is simpler and more uniform. Let's just find a good way to show the bound parameters, how would named parameters be handled?

sushantdhiman:
Ah, sorry @papb but still we can't merge this extra code which we really don't need.

papb:
@supperchong Are you willing to rework the PR to follow the new suggestion from sushantdhiman instead of your original idea?

supperchong:
@papb  Yeah, I give another [PR](https://github.com/sequelize/sequelize/pull/11267).  Thanks very much!

papb:
Closing in favor of #11267.

PR #11246, opened 2019-07-29
feat(postgres): native updateOnConflict for upsert

### LINKED ISSUE #4132: Postgres UPSERT native support

bmeriaux:
### Pull Request check-list

### Description of change

replace the function to catch exception on existing element for upsert method with Postgres by the native method ON CONFLICT () UPDATE SET, it is supported by postgres >= 9.5, is this blocking ?

Closes #4132

### DIFF
--- a/lib/dialects/abstract/index.js
+++ b/lib/dialects/abstract/index.js
@@ -7,7 +7,6 @@ AbstractDialect.prototype.supports = {
   'DEFAULT VALUES': false,
   'VALUES ()': false,
   'LIMIT ON UPDATE': false,
-  'ON DUPLICATE KEY': true,
   'ORDER NULLS': false,
   'UNION': true,
   'UNION ALL': true,
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -3,7 +3,6 @@
 const util = require('util');
 const _ = require('lodash');
 const uuidv4 = require('uuid/v4');
-const semver = require('semver');
 
 const Utils = require('../../utils');
 const deprecations = require('../../utils/deprecations');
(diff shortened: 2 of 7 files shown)


papb:
> postgres >= 9.5, is this blocking ?

Without your PR, what is the minimum required postgres version? (I probably should know this, but I don't...)

bmeriaux:
if we look at the previous default version, it was 9.4
but a previous PR  #11163 on bulkInsert with the same ON CONFLICT syntax already implicit set required version to 9.5

papb:
I see. Thanks! I think it is better for @sushantdhiman to take a look.

sushantdhiman:
Tests are failing

up73k:
@bmeriaux, will you fix tests in nearest future?

papb:
Also, there are conflicts with the master branch, @bmeriaux please rebase your branch.

bmeriaux:
I will try this week.
But I find out that with this native upsert, we lose the information if it is an insert or an update,
And it may not work on case with 2 distinct unique constraints.
So i do not know if we must detect in which case we are to keep existing temp function for some edge case, and use native upsert for others ?
What did you think ?



Le 13 août 2019 à 14:37:41, Pedro Augusto de Paula Barbosa (notifications@github.com<mailto:notifications@github.com>) a écrit:

Also, there are conflicts with the master branch, @bmeriaux<https://github.com/bmeriaux> please rebase your branch.

papb:
> I will try this week. But I find out that with this native upsert, we lose the information if it is an insert or an update, And it may not work on case with 2 distinct unique constraints. So i do not know if we must detect in which case we are to keep existing temp function for some edge case, and use native upsert for others ? What did you think ?

Can you show a SSCCE/MVCE/reprex for the situation that is covered by Sequelize's workaround that is not covered by the native implementation?

PR #11249, opened 2019-07-30
Fix: postgresql's ensureEnums (#11053)

### LINKED ISSUE #11053: Adding multiple new enum values may cause model sync errors (ensureEnums, postgres)
## What are you doing?
I'm trying to add several new enum values to already existing type in postgres

**To Reproduce**

1. Have any model with enum column like this:

```
{
  ...
  myEnumColumn: {
    type: Sequelize.ENUM([
      'type1',
      'type2'
    ])
  }
  ...
}
```

2. Sync this model and then apply changes:
```
{
  ...
  myEnumColumn: {
    type: Sequelize.ENUM([
      'type-1', // new
      'type0', // new
      'type1',
      'type2',
      'type3', // new
      'type4', // new
      'type5' // new
    ])
  }
  ...
}
```
3. Sync again and this should fail with errors like:
`[SequelizeDatabaseError]: "type0" is not an existing enum label`
or
`[SequelizeDatabaseError]: "type4" is not an existing enum label`

I wanted values to be added to enum type based on already existing types

It looks like ensureEnum function uses not-added-yet types as reference for BEFORE or AFTER, making queries like these:

```
ALTER TYPE "public"."enum_db_myenumcolumn" ADD VALUE IF NOT EXISTS 'type-1' BEFORE 'type0'
```
or
```
ALTER TYPE "public"."enum_db_myenumcolumn" ADD VALUE IF NOT EXISTS 'type3' BEFORE 'type4'
```

## Environment
Dialect:
- [ ] mysql
- [x] postgres
- [ ] sqlite
- [ ] mssql
- [ ] any
Database version: 11.3
Sequelize version: 5.8.8
Node Version: 12

Equals182:
### Pull Request check-list

*I am sorry, I wasn't able to run `npm run test` or `npm run test-DIALECT` on my local machine, nor did I added any new tests or even followed 'commit message conventions'. My bad.*

*I tested this solution locally on two arrays with various values, all got me right results. I also tested it on my local postgres database on real project, which also worked just fine.*

**Please, feel free to help me make this Pull Request look good and be merged**

---

### Description of change

Fixes #11053

The following is an attempt to explain how my solution works:

General idea is to use already existing values of enum type as a reference for new values. Previous (and currently active) version has some flaws, like referring non-existent values and some sort of 'race condition' from using Promise.all. You may check issue #11053 for more details.

Current implementation intent is to avoid those flaws by:
1. Going through already existing values
2. Using waterfall-like promise resolving

**Going through already existing values:**  

Let us have an existing enum array `[1,2]` and we want to make it look like `[4,3,2,6,1,0,5]`.  
Assuming we cannot change order of already existing values, ideal result should look like this: `[6,1,4,3,2,0,5]`  
To achieve this result, we must go through values `[1,2]` and find every new value that prepends it. 

We find position in new array for each old value, which is idx `4` for `1` and `2` for `2`. Then we get as many values on the left side as we can and going through them in reverse to find only *new* values that prepends current.

For value `1` its `[4,3,2,6]`, reversed `[6,2,3,4]` and new value is `[6]`
For value `2` its `[4,3]` which gives us `[3,4]` as result

Next trick is to add SQL query for each new value in right order, so we use `promises.splice(capturedPromiseLength, 0, asyncFunction)` to 'reverse-push' values.  
That way we added all 'left-side' values using BEFORE and referring already existing values.

For the rest of new values on the right side, we just use `slice(indexOfDextralOldValueInNewArray)` and then going through resulting array in reverse order we push values in `promise` array. The reverse order is needed because we use AFTER, so appending `[1,2]` to `0` needs to be a sequence of `ADD 2 AFTER 0` and `ADD 1 AFTER 0`.

**Using waterfall-like promise resolving**

Since `Promise.all` does not guarantee execution order and creates a 'race condition' for ENUM altering, we switch to 'waterfall' system, where we attach a function that returns Promise (basically, an async function) to previous Promise, using `then` method.  
As a result we're getting something like this:
`Promise.resolve().then(() => sequelize.query()).then(() => sequelize.query()) ... .tap(...)`

### DIFF
--- a/lib/dialects/postgres/query-interface.js
+++ b/lib/dialects/postgres/query-interface.js
@@ -74,6 +74,20 @@ function ensureEnums(qi, tableName, attributes, options, model) {
           const enumVals = qi.QueryGenerator.fromArray(results[enumIdx].enum_value);
           const vals = enumType.values;
 
+          console.log('already exists', enumVals)
+          console.log('new values', vals)
+
+          enumVals.forEach((value, idx) => {
+            const index = vals.indexOf(value);
+            if (idx === 0) {
+              // first value in existing enum array
+              if (index > 0) {
+                // this value isn't first anymore, we need to append values before this one in reverse order
+              }
+            } else {
+            }
+          })
+
           vals.forEach((value, idx) => {
             // reset out after/before options since it's for every enum value
             const valueOptions = _.clone(options);
--- a/lib/dialects/postgres/query-interface.js
+++ b/lib/dialects/postgres/query-interface.js
@@ -63,11 +63,11 @@ function ensureEnums(qi, tableName, attributes, options, model) {
       switch (position) {
         case 'after':
           valueOptions.after = relativeValue;
-          break
+          break;
         case 'before':
         default:
           valueOptions.before = relativeValue;
-          break
+          break;
       }
 
       promises.splice(spliceStart, 0, () => {
@@ -75,7 +75,7 @@ function ensureEnums(qi, tableName, attributes, options, model) {
           tableName, field, value, valueOptions
         ), valueOptions);
       });
-    }
+    };
 
     for (i = 0; i < keyLen; i++) {
       const attribute = attributes[keys[i]];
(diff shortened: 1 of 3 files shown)


papb:
Hello! :grimacing: 

> *I am sorry, I wasn't able to run npm run test or npm run test-DIALECT on my local machine, nor did I added any new tests or even followed 'commit message conventions'. My bad.*

It's ok, you are a first-time contributor, thank you very much for all the work so far, later I can show you how to write and run tests!

I can't take a deeper look right now, sorry, perhaps someone else will be able to. Otherwise you will have to wait a bit... Thanks for understanding!

PR #11255, opened 2019-07-30
fix(model): destroying paranoid models with custom deletedAt

### LINKED ISSUE #11253: Soft-deleting paranoid model with custom deletedAt DB column name sets wrong attribute on instance
## What are you doing?

```js
const ParanoidUser = this.sequelize.define('ParanoidUser', {
  username: Support.Sequelize.STRING,

  // Custom DB column name
  deletedAt: {type: Support.Sequelize.DATE, field: 'deleted_at'}
}, {paranoid: true});

return ParanoidUser.sync({force: true}).then(() => {
  return ParanoidUser.create({
    username: 'username'
  });
}).then(user => {
  return user.destroy();
}).then(user => {
  // Both these assertions fail. Sequelize sets dataValues.deleted_at instead of deletedAt
  expect(user.dataValues.deletedAt).to.be.ok;
  expect(user.dataValues.deleted_at).to.not.be.ok;
});
```

**To Reproduce**
1. Define paranoid model with `deletedAt` that declares custom `field`
2. Soft delete model instance
3. Check resulting model instance

`deletedAt` should be set.

`deleted_at` is set.

## Environment
Dialect:
- [ ] mysql
- [ ] postgres
- [ ] sqlite
- [ ] mssql
- [x] any
Dialect **library** version: any
Database version: any
Sequelize version: master
Node Version: XXX
OS: XXX
Tested with latest release:
- [ ] No
- [x] Yes, specify that version: master

gabegorelick:
Fixes #11253

### Pull Request check-list

- [x] Did you update the typescript typings accordingly (if applicable)? **N/A**

### Description of change

Fixes #11253

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -4052,29 +4052,19 @@ class Model {
       const where = this.where(true);
 
       if (this.constructor._timestampAttributes.deletedAt && options.force === false) {
-        const attribute = this.constructor.rawAttributes[this.constructor._timestampAttributes.deletedAt];
-        const field = attribute.field || this.constructor._timestampAttributes.deletedAt;
-        const values = Utils.mapValueFieldNames(this.dataValues, this.changed() || [], this.constructor);
-
-        values[field] = new Date();
-        where[field] = Object.prototype.hasOwnProperty.call(attribute, 'defaultValue')
+        const attributeName = this.constructor._timestampAttributes.deletedAt;
+        const attribute = this.constructor.rawAttributes[attributeName];
+        const defaultValue = Object.prototype.hasOwnProperty.call(attribute, 'defaultValue')
           ? attribute.defaultValue
           : null;
+        const currentValue = this.getDataValue(attributeName);
+        const undefinedOrNull = currentValue == null && defaultValue == null;
+        if (undefinedOrNull || _.isEqual(currentValue, defaultValue)) {
+          // only update timestamp if it wasn't already set
+          this.setDataValue(attributeName, new Date());
+        }
 
-        this.setDataValue(field, values[field]);
-
-        return this.constructor.QueryInterface.update(
-          this, this.constructor.getTableName(options), values, where, _.defaults({ hooks: false, model: this.constructor }, options)
-        ).then(([results, rowsUpdated]) => {
-          if (this.constructor._versionAttribute && rowsUpdated < 1) {
-            throw new sequelizeErrors.OptimisticLockError({
-              modelName: this.constructor.name,
-              values,
-              where
-            });
-          }
-          return results;
-        });
+        return this.save(_.defaults({ hooks: false }, options));
       }
       return this.constructor.QueryInterface.delete(this, this.constructor.getTableName(options), where, Object.assign({ type: QueryTypes.DELETE, limit: null }, options));
     }).tap(() => {
--- a/test/integration/instance.test.js
+++ b/test/integration/instance.test.js
@@ -685,5 +685,109 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
         expect(user.username).to.equal('Peter');
       });
     });
+
+    it('supports custom deletedAt field', function() {
+      const ParanoidUser = this.sequelize.define('ParanoidUser', {
+        username: DataTypes.STRING,
+        destroyTime: DataTypes.DATE
+      }, { paranoid: true, deletedAt: 'destroyTime' });
+
+      return ParanoidUser.sync({ force: true }).then(() => {
+        return ParanoidUser.create({
+          username: 'username'
+        });
+      }).then(user => {
+        return user.destroy();
+      }).then(user => {
+        expect(user.destroyTime).to.be.ok;
+        expect(user.deletedAt).to.not.be.ok;
+        return user.restore();
+      }).then(user => {
+        expect(user.destroyTime).to.not.be.ok;
+        return ParanoidUser.findOne({ where: { username: 'username' } });
+      }).then(user => {
+        expect(user).to.be.ok;
+        expect(user.destroyTime).to.not.be.ok;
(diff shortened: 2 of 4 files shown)


lib/model.js
```diff
@@ -4052,29 +4052,17 @@ class Model {
...
-          if (this.constructor._versionAttribute && rowsUpdated < 1) {
-            throw new sequelizeErrors.OptimisticLockError({
-              modelName: this.constructor.name,
-              values,
-              where
-            });
-          }
-          return results;
-        });
+        return this.save(_.defaults({ hooks: false }, options));
```
gabegorelick: Using `save` lets us reuse a lot of code.

test/integration/instance.test.js
```diff
@@ -685,5 +685,104 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
         expect(user.username).to.equal('Peter');
       });
     });
+
+    it('supports custom deletedAt field', function() {
```
gabegorelick: `restore` tests are still in here. Happy to move them to their own file if you want.

test/integration/instance/destroy.test.js
```diff
@@ -76,24 +102,254 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
...
       });
 
+      const logging = sinon.spy();
+
       return UserDelete.sync({ force: true }).then(() => {
         return UserDelete.create({ name: 'hallo', bio: 'welt' }).then(u => {
           return UserDelete.findAll().then(users => {
             expect(users.length).to.equal(1);
-            return u.destroy({
-              logging(sql) {
```
gabegorelick: This didn't verify that `logging` was actually called.

papb:
Looks great to me! Superb PR.

test/integration/instance/destroy.test.js
```diff
@@ -56,6 +56,32 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
       });
     });
 
+    it('does not update deletedAt with custom default in subsequent destroys', function() {
```
gabegorelick: Pretty sure this is what's intended, although this test fails on `master`.

lib/model.js
```diff
@@ -4052,29 +4052,18 @@ class Model {
...
-
-        values[field] = new Date();
-        where[field] = Object.prototype.hasOwnProperty.call(attribute, 'defaultValue')
+        const attributeName = this.constructor._timestampAttributes.deletedAt;
+        const attribute = this.constructor.rawAttributes[attributeName];
+        const defaultValue = Object.prototype.hasOwnProperty.call(attribute, 'defaultValue')
           ? attribute.defaultValue
           : null;
+        const currentValue = this.getDataValue(attributeName);
+        if (currentValue == defaultValue || _.isEqual(currentValue, defaultValue)) {
```
gabegorelick: Needed to add `==` check for non-Postgres DBs. For some reason they can return `undefined` for `currentValue`, which is not `_.isEqual` to `null`.

gabegorelick:
@papb Thanks. Tests are passing now. Most of the issues were in the tests not accounting for non-PG quirks (not supporting sub-second dates or Infinity).

test/integration/instance/destroy.test.js
papb: Great catch! This reminded me of something I saw recently. I like how you fixed this, using sinon. Would you be willing to do something similar to refactor several other tests with a similar problem?

If you search for `expect(test).to.be.true` you will find 9 occurrences in tests, they all could be refactored to use sinon, would you like to do that?

(And perhaps, if the code gets too big, you could even create a support method to be reused around tests)

test/integration/instance.test.js
```diff
@@ -685,5 +685,104 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
...
+      }, { paranoid: true });
+
+      return ParanoidUser.sync({ force: true }).then(() => {
+        return ParanoidUser.create({
+          username: 'username'
+        });
+      }).then(user => {
+        return user.destroy();
+      }).then(user => {
+        return user.restore();
```
sushantdhiman:
```js
expect(user.dataValues.deletedAt).to.be.ok;
expect(user.dataValues.deleted_at).to.not.be.ok;
return user.restore();
```

test/integration/instance.test.js
```diff
@@ -685,5 +685,104 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
...
+      }, { paranoid: true, deletedAt: 'destroyTime' });
+
+      return ParanoidUser.sync({ force: true }).then(() => {
+        return ParanoidUser.create({
+          username: 'username'
+        });
+      }).then(user => {
+        return user.destroy();
+      }).then(user => {
+        return user.restore();
```
sushantdhiman:
```js
expect(user.dataValues.destroyTime).to.be.ok;
expect(user.dataValues.deletedAt).to.not.be.ok;
expect(user.dataValues.destroy_time).to.not.be.ok;
return user.restore();
```

lib/model.js
gabegorelick: Linting didn't like the extra parens around `currentValue == null && defaultValue == null` so I had to move it out into a variable:
```js
const undefinedOrNull = currentValue == null && defaultValue == null;
if (undefinedOrNull || _.isEqual(currentValue, defaultValue))
```

test/integration/instance.test.js
gabegorelick: Done

test/integration/instance.test.js
gabegorelick: Done

test/integration/instance/destroy.test.js
papb: (of course, this could be in a separate PR, if you're willing to do it - has nothing to do with this PR)

Issue #10622, opened 2019-03-25
Best Practice documentation for import


chrisgo:
Awesome job getting to v5!

Possibly add a small example "best practice" of how to completely separate the model files in the documentation 

http://docs.sequelizejs.com/manual/models-definition.html > Import

is very close but it does not address how to best put the associations into the separate model files

The examples here assume that you have imported all the models into this one file and then code all the associations after all the models have been imported

http://docs.sequelizejs.com/manual/associations.html

I believe most are using this an example prior to v5
https://codewithhugo.com/using-es6-classes-for-sequelize-4-models/

trappar:
👍 This has me stumped too

chrisgo:
There is an example here
https://github.com/sequelize/express-example/blob/master/models/index.js

brenc:
Also:

http://docs.sequelizejs.com/manual/models-definition.html > Import

```javascript
const Project = sequelize.import(__dirname + "/path/to/models/project")

module.exports = (sequelize, DataTypes) => {
  class Project extends sequelize.Model { }
  Project.init({
    name: DataTypes.STRING,
    description: DataTypes.TEXT
  }, { sequelize });
  return Project;
}
```
This for me crashes with:

```TypeError: Class extends value undefined is not a constructor or null```

The `sequelize` instance doesn't appear to have a reference to `Model`. What works for me is this:

```javascript
const Project = sequelize.import(__dirname + "/path/to/models/project")

// In the mode file:
const Sequelize = require('sequelize')

module.exports = (sequelize, DataTypes) => {
  class Project extends Sequelize.Model { }
  Project.init({
    name: DataTypes.STRING,
    description: DataTypes.TEXT
  }, { sequelize });
  return Project;
}
```

jackie1120:
> Also:
> 
> http://docs.sequelizejs.com/manual/models-definition.html > Import
> 
> ```js
> const Project = sequelize.import(__dirname + "/path/to/models/project")
> 
> module.exports = (sequelize, DataTypes) => {
>   class Project extends sequelize.Model { }
>   Project.init({
>     name: DataTypes.STRING,
>     description: DataTypes.TEXT
>   }, { sequelize });
>   return Project;
> }
> ```
> 
> This for me crashes with:
> 
> `TypeError: Class extends value undefined is not a constructor or null`
> 
> The `sequelize` instance doesn't appear to have a reference to `Model`. What works for me is this:
> 
> ```js
[...]

PR #11260, opened 2019-07-31
fix(import): add missing Model in Sequelize prototype

irekg:
### Pull Request check-list

### Description of change

According to discussion on #10622 current example for importing modules from separate files don't work. It's because instance od `Sequelize` don't have `Model` class defined in prototype. [Example of Error](https://github.com/sequelize/sequelize/issues/10622#issuecomment-500492926)

To fix this issue `Model` class was introduced in "Aliases" section of main `sequelize.js` file. It was tested manually and with mocha tests using `DIALECT=mysql npm run test-docker-integration`.

What have been done:
- Extend Sequelize Aliases with Model.
- Modify project.js asset used to test import feature.  It should detect issue if it will appear in again in future.  Also it now reflects example from [documentation](https://sequelize.org/master/manual/models-definition.html#import).

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -1261,6 +1261,7 @@ Sequelize.prototype.or = Sequelize.or;
 Sequelize.prototype.json = Sequelize.json;
 Sequelize.prototype.where = Sequelize.where;
 Sequelize.prototype.validate = Sequelize.prototype.authenticate;
+Sequelize.prototype.Model = Model;
 
 /**
  * Sequelize version number.
--- a/test/integration/assets/project.js
+++ b/test/integration/assets/project.js
@@ -1,7 +1,12 @@
 'use strict';
 
-module.exports = function(sequelize, DataTypes) {
-  return sequelize.define(`Project${parseInt(Math.random() * 9999999999999999, 10)}`, {
-    name: DataTypes.STRING
-  });
+module.exports = (sequelize, DataTypes) => {
+  class Project extends sequelize.Model { }
+
+  Project.init({
+    name: DataTypes.STRING,
+    description: DataTypes.TEXT
+  }, { sequelize, modelName: `Project${parseInt(Math.random() * 9999999999999999, 10)}` });
+
+  return Project;
 };


lib/sequelize.js
```diff
@@ -1261,6 +1261,7 @@ Sequelize.prototype.or = Sequelize.or;
 Sequelize.prototype.json = Sequelize.json;
 Sequelize.prototype.where = Sequelize.where;
 Sequelize.prototype.validate = Sequelize.prototype.authenticate;
+Sequelize.prototype.Model = Model;
```
papb: Please add a typescript typing for this. Observe the above lines:

https://github.com/sequelize/sequelize/blob/50c8a42045d3ea75337d7523ecbc327eb4cffb9d/lib/sequelize.js#L1254-L1262

They all have typings:

https://github.com/sequelize/sequelize/blob/d041e77e275c3e950cedcdb3a625d891ea8b056b/types/lib/sequelize.d.ts#L369-L451

So please add a typing for the newly added `Sequelize.prototype.Model = Model` in the same file. I suggest you do it [after line 451](https://github.com/sequelize/sequelize/blob/d041e77e275c3e950cedcdb3a625d891ea8b056b/types/lib/sequelize.d.ts#L451) to preserve the same positioning.

irekg: Will do. I just have a question about this 
https://github.com/sequelize/sequelize/blob/50c8a42045d3ea75337d7523ecbc327eb4cffb9d/lib/sequelize.js#L1326

There is static implementation of `Model`. But it's left with no documentation in both files. Maybe it should be documented along with public method or something?

papb: I think you can leave it without a JSDoc documentation then. Just add the typing.

papb:
Actually, [as you said yourself](https://github.com/sequelize/sequelize/issues/10622#issuecomment-516944789), this can be accessed by simply doing `sequelize.Sequelize.Model`, so I think this is not necessary.

Let me know if you think there's anything else to be done.

irekg:
Yes It's no longer needed. Maybe only this:
https://github.com/sequelize/sequelize/pull/11260/commits/50c8a42045d3ea75337d7523ecbc327eb4cffb9d#diff-6b9b824eadb37f6f05b301f55c072107

But with `class Project extends sequelize.Sequelize.Model { }`
To avoid similar errors in future. But I won't insist ;)

papb:
@irekg I think it's better to leave it this way. Regardless, thanks for being so kind to take some time to help Sequelize! I hope to see more PRs from you in the future :+1:

SimonSchick:
This was removed in https://github.com/sequelize/sequelize/releases/tag/v5.1.0 actually.

PR #11262, opened 2019-07-31
feat(associations): add support for default referential action

### LINKED ISSUE #10824: Change default referential action
**Problem**
Associations (foreign keys) in Sequelize define a rather opinionated default referential action for `ON UPDATE` / `ON DELETE` clauses.

For `ON UPDATE`, Sequelize defaults to `CASCADE`.
For `ON DELETE`, Sequelize defaults to `SET NULL` if the source field allows null values; otherwise, `CASCADE` is the default.

It would be convenient to set `RESTRICT` or `NO ACTION` as the default (database-wide) in some circumstances.  In MySQL, for example, the default referential action is `RESTRICT` when not explicitly specified; Sequelizes seems to override these defaults.

**Describe the solution you'd like**
It would be nice to override the "default" referential action for `ON UPDATE` and `ON DELETE` when associations are created.  It can get annoying to override this behavior for every single association in the database if `RESTRICT` is the general rule / policy.  This could be done in the `options` Object passed to the new Sequelize instance (see usage example below).

**Why should this be in Sequelize**
This is a small change (probably ~10 lines of code), but it gives extra flexibility to the developer without requiring the developer to be verbose with association definitions.

**Describe alternatives you've considered**
Another possible solution is creating some sort of hook that allows one to change the default `onDelete` and `onUpdate` options of an association.  I don't know of any such hook at this time.

**Usage example**
```js
const db = new Sequelize(config.database, config.username, config.password, {
	host: config.host,
	// New `associations` key
	associations: {
		onDelete: 'RESTRICT',
		onUpdate: 'CASCADE',
	}
})
```

bminer:
### Pull Request check-list

Closes #10824

### Description of change

It would be nice to override the "default" referential action for ON UPDATE and ON DELETE when associations are created. It can get annoying to override this behavior for every single association in the database if RESTRICT is the general rule / policy. This could be done in the options Object passed to the new Sequelize instance (see usage example below).

```js
const db = new Sequelize(config.database, config.username, config.password, {
	host: config.host,
	// New `associations` key
	associations: {
		onDelete: 'RESTRICT',
		onUpdate: 'CASCADE',
	}
})
```

### DIFF
--- a/lib/associations/mixin.js
+++ b/lib/associations/mixin.js
@@ -23,6 +23,11 @@ const Mixin = {
     // Since this is a mixin, we'll need a unique letiable name for hooks (since Model will override our hooks option)
     options.hooks = options.hooks === undefined ? false : Boolean(options.hooks);
     options.useHooks = options.hooks;
+    const assocOptions = this.sequelize.options.associations;
+    if(assocOptions) {
+      options.onDelete = options.onDelete || assocOptions.onDelete;
+      options.onUpdate = options.onUpdate || assocOptions.onUpdate;
+    }
 
     options = Object.assign(options, _.omit(source.options, ['hooks']));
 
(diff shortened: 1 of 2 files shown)


papb:
Hello, thanks for the PR! This is a nice feature. Thanks for spending your time to help Sequelize! I hope to see more PRs from you in the future :grimacing:

sushantdhiman:
You can achieve this with `beforeAssociate / afterAssociate` hooks, right?

sushantdhiman:
I think this can be achieved with hooks, if that is not possible let me know

bminer:
@sushantdhiman - Can you provide an example on how to use the `beforeAssociate` hooks?  I think I've tried doing that in the past.

sushantdhiman:
https://github.com/sequelize/sequelize/search?q=beforeAssociate&unscoped_q=beforeAssociate

PR #11267, opened 2019-08-01
fix(log): show bind parameters in logs instead of placeholders

supperchong:
Closes #[11151](https://github.com/sequelize/sequelize/issues/11151)
Closes #[10067](https://github.com/sequelize/sequelize/issues/10067)
### Pull Request check-list


papb:
Hi @supperchong, thank you very much once again. I see you are still adding commits, so when you think this is ready please comment here to make it clear, so maintainers can look.

supperchong:
@papb soga~It's ok.

lib/dialects/abstract/query.js
```diff
@@ -324,14 +324,22 @@ class AbstractQuery {
...
     const { connection, options } = this;
     const benchmark = this.sequelize.options.benchmark || options.benchmark;
     const startTime = Date.now();
-    const fmt = `(${connection.uuid || 'default'}): ${sql}`;
+    let logParameter='';
+    if (!sql.endsWith(';')) {
+      logParameter=';';
+    }
+    if (parameters) {
+      logParameter+=JSON.stringify(parameters);
```
sushantdhiman: As I said in comments below, we want one standard style for appending `parameters` after SQL. I think comma separated values works best.

Also please use space before/after `=` :)

supperchong: Thanks a lot. The parameters are object for mssql and sqlite. Should remain them in log ?
like
```js
INSERT INTO `Users` (`id`,`username`,`emailAddress`) VALUES (NULL,$1,$2);{"$1":"john","$2":"john@gmail.com"}
```
and 
```js
INSERT INTO `Users` (`id`,`username`,`emailAddress`) VALUES (NULL,$one,$two);{"$one":"john","$two":"john@gmail.com"}
```

sushantdhiman: No they should be appended as array to final sql. Like `SQL SQL SQL; value1, value2, value3`. You will have to convert object to array in `_logQuery` function

supperchong: @sushantdhiman So the append values have the same order and  repeat with placeholders ?

sushantdhiman: Hmm object property ordering could be an issue, when it is an array append it, when it is an object stringify it.

I think we might want to hide this behind extra flag like `sequelize.options.loggingParameters= false`

supperchong: I have a question. Single quotes in PostgreSQL are used to create a text string, so should we convert string to single quotes for postgreSQL?
eg. 
```js
INSERT INTO "users" ("id","name") VALUES (DEFAULT,$1) RETURNING *; 'john'
```
In my opinion, sometimes the parameters will handle again by Individual module . For example postgres use [pg](https://github.com/brianc/node-postgres) .
So we remain what we send to this module. Just  as
```js
   //_logQuery.js
    if (!sql.endsWith(';')) {
      logParameter=';';
    }
    if (parameters) {
      if (Array.isArray(parameters)) {
        logParameter = logParameter+' '+parameters.map(p=>JSON.stringify(p)).join(', ')
      } else {
        logParameter += JSON.stringify(parameters);
      }
    }
```
Is this ok?

sushantdhiman: These queries are only being logged and there no way you can actually execute what we log `SQL; parameters` is just for our logging purposes. There is no need to quote parameter strings in any particular style when logging then

supperchong: I agree with you .But 
```js
sequelize.query(`select $1 as foo,$2 as bar`, { bind: ['foo, bar','bar'])
```
If we don't quote strings ,It would be 
```
select $1 as foo,$2 as bar; foo, bar, bar
```

sushantdhiman: Right, this is getting complicated (i thought we could just append them). There are our issues so far

1. We don't have same format for all dialects, due to object vs array

Possible Sol: I think bindParam is just an array, we should be able to get array of parameters for all cases. Perhaps check `options.bindParams` if that is available in context.

2. String should be properly quoted, so they don't lose their semantics in relation for bounded variables.

Possible Sol: JSON.stringify seems like a good idea for that

3. There are buffers types, do we want to print them all out. That result in significant noise in logs

Possible Sol: filter object or array for buffer representation or hexified data and omit that before json stringify

Don't need to worry too much about (3) now, but we will tackle that at the end of this PR. Let's start with (1) and (2) first

lib/dialects/sqlite/query.js
```diff
@@ -221,13 +221,16 @@ class Query extends AbstractQuery {
     const conn = this.connection;
```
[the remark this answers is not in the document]
supperchong: These changes is necessary. When method is 'exec', the sql will be formated and does't contain `$ | ?` , but parameters is not undefined.

lib/dialects/abstract/query.js
supperchong: 1 and 2 is ok. Please check it. Furthermore, although `options.bindParams` is an array, but sql may contain `$one` etc.

lib/dialects/sqlite/query.js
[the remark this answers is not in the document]
sushantdhiman: Can't we handle undefined parameters at `_logQuery`?

lib/dialects/abstract/query.js
sushantdhiman: No problem, at-least all bounded variables will be logged in same format by using parameters from `options.bindParams`

lib/dialects/sqlite/query.js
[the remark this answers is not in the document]
supperchong: emmm ,for example
```js
run(sql, parameters) {
    const conn = this.connection;
    this.sql = sql;
    const method = this.getDatabaseMethod();
    let complete;
    if (method === 'exec') { 
      console.log('prevSql', sql); 
//INSERT OR IGNORE INTO `Users` (`username`,`mood`) VALUES ($1,$2);
      // exec does not support bind parameter
      sql = AbstractQuery.formatBindParameters(sql, this.options.bind, this.options.dialect || 'sqlite', { skipUnescape: true })[0];
      this.sql = sql;
      console.log('parameters', parameters); 
//{ '$1': 'Toni', '$2': 'happy' }  
      console.log('sql', sql); 
//INSERT OR IGNORE INTO `Users` (`username`,`mood`,`createdAt`,`updatedAt`) VALUES ('Toni','happy'); )
      complete = this._logQuery(sql, debug);
    } else {
      complete = this._logQuery(sql, debug, parameters);
    }
   ```

File: lib/dialects/abstract/query.js (hunk 1 above)
Comment (supperchong) [reply lib/dialects/abstract/query.js]: Yeah, i know, i will modify it.

File: lib/dialects/sqlite/query.js (hunk 2 above)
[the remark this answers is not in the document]
Comment (sushantdhiman) [reply lib/dialects/sqlite/query.js]: Ah understood, keep these changes then

File: lib/dialects/abstract/query.js (hunk 1 above)
Comment (supperchong) [reply lib/dialects/abstract/query.js]: `options.bind`?

Comment (supperchong) [reply lib/dialects/abstract/query.js]: There is not `options.bindParams` , `options.bind` and `options.replacements` may be object.

**Comment by papb** [comment]
Hi @supperchong, is this ready in your opinion? Or are you still working on this?

**Comment by supperchong** [comment]
@papb Hi, I hope to resolve this problem. I'm waiting for  @sushantdhiman to reply.

**Comment by sushantdhiman** [comment]
@supperchong Hide this feature behind a config flag, like `sequelize.config.logQueryParameters`, otherwise LGTM

**Comment by supperchong** [comment]
@sushantdhiman @papb Thanks for your advise. I add logQueryParameters options, is this ok?

**Comment by superlbr** [comment]
not convenient to rerun sql...

**Comment by papb** [comment]
@superlbr What do you mean? Can you be more specific?

**Comment by superlbr** [comment]
@papb  sql in error is different
before:
```
original: {
  "code":"ER_DATA_TOO_LONG",
  "sql":"INSERT INTO `crm_member` (`id`,`road_name`) VALUES (DEFAULT,'xxx');"
}
```
after:
```
original: {
  "code":"ER_DATA_TOO_LONG",
  "sql":"INSERT INTO `crm_member` (`id`,`road_name`) VALUES (DEFAULT,?);"
  "parameters": [DEFAULT, 'xxx']
}
```

papb:
Hmm, I'd say the new way is better though :grimacing: you can even rerun the query with bind parameters now, it has much better security!

PR #11274, opened 2019-08-02
fix(postgres): schema for index and relations

### LINKED ISSUE #11276: Postgres: Schema does not default to current schema for references when model is set only on parent model
## What are you doing?

I need to switch schemas dynamically for multi tenancy in an application and when I change the schema for a model, I would like sequalize to default to that schema also for references and indexes. This is not the case yet and I would consider this a bug as it's the only sensible default behaviour.
The main point where this happens is when I call `sync({ schema: 'archive' })` on a model,

Sequalize to default to the schema of the model I'm using.

If I sync a model into a schema for example (I have used `sync({ schema: 'archive' })` on the models before), for references the current schema is not applied but no schema prefix at all.

## Environment
Dialect:
- [ ] mysql
- [x] postgres
- [ ] sqlite
- [ ] mssql
- [ ] any
Database version: XXX
Sequelize version: XXX
Node Version: XXX
OS: XXX
Tested with latest release:
- [ ] No

vanthome:
Relations and Indexes created from models that have set a schema are not properly prefixed with the schema. This patch solves this.

### Pull Request check-list

### Description of change

In several cases, the schema prefix was not applied for generated indexes and related tables. This PR fixes this.

Fixes #11276.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -569,6 +569,13 @@ class QueryGenerator {
       options.where = this.whereQuery(options.where);
     }
 
+    if (options.schema) {
+      tableName = {
+        tableName,
+        schema: options.schema
+      };
+    }
+
     if (typeof tableName === 'string') {
       tableName = this.quoteIdentifiers(tableName);
     } else {
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -514,7 +514,19 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
     }
 
     if (attribute.references) {
-      const referencesTable = this.quoteTable(attribute.references.model);
+      let referencesTable = this.quoteTable(attribute.references.model);
+      const tableDetails = this.extractTableDetails(referencesTable, options);
+      let schema;
+      if (options.schema) {
+        schema = options.schema;
+      } else if ((!attribute.references.model || typeof attribute.references.model == 'string')
+        && options.table && options.table.schema) {
+        schema = options.table.schema;
+      }
+
+      if (schema) {
+        referencesTable = schema + tableDetails.delimiter + referencesTable;
+      }
       let referencesKey;
 
       if (attribute.references.key) {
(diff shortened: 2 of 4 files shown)


vanthome:
From my perspective this is complete now, please merge it.

papb:
@vanthome I see you did not add any test... Please add at least one simple test that wouldn't pass before your PR and passes after your PR. Thanks.

vanthome:
Added a test. For whatever reason now the tests fail although they run locally?

vanthome:
Seems like MySQL tests fail due to some code that I integrated through rebasing. So will you merge this?

papb:
@vanthome The test you added is running for all dialects, and failing for every dialect which is not postgres. You have to add your test in a way that it will run only for postgres. Let me know if you need further help.

papb:
@vanthome Just add your test [here](https://github.com/sequelize/sequelize/blob/master/test/integration/dialects/postgres/associations.test.js) instead and it should work.

vanthome:
Strange because previously the tests passed I think. Anyways, I have moved it there.

vanthome:
Now the PG native test seems to fail which has nothing to do with my change. What to do now?

papb:
@vanthome Nothing else to do! That error in postgres-native is an intermittent error that has been happening in Sequelize for a while... We can all ignore it.

Thank you very much for the PR! Looks ready to merge.

Let's wait for @sushantdhiman

vanthome:
Great, looking forward.

papb:
@vanthome There are conflicts with master, can you please resolve them?

vanthome:
Oh crap, I just pushed another update to my local branch, I did not want to add this to this PR. Do you know how I can revert this?

papb:
@vanthome Ah, that explains the last commit :grimacing:

You can just revert the commit normally (`git revert`), don't worry about the "ugly git history" because everything will be squashed on merge anyway.

vanthome:
I have undone the malicious change. When will this be merged now?

papb:
Hi! There are still conflicts...

vanthome:
Well this was w/o conflicts. why has it not been merged then?
I mean why should I put more effort into this, is it does not get merged anyway?

papb:
> Well this was w/o conflicts.

Indeed! But as time passes, conflicts may appear, it's normal.

> why has it not been merged then?

Because maintainers are busy

> I mean why should I put more effort into this, is it does not get merged anyway?

You don't have to do anything. In open source no one is forced to anything... I, like you, would like to see every PR merged as well.

papb:
I fixed the conflict for you :+1:

vanthome:
Great, looking forward for a merge.

PR #11287, opened 2019-08-06
Fix: PK datatype was not used

### LINKED ISSUE #10941: model: On model save QueryGenerator cannot get PK attribute type
## What are you doing?

```js
const { Sequelize, DataTypes, Model } = require('sequelize');
const { bufferToGuid, guidToBuffer, guid } = require('@up24/guid');
const { inherits } = require('util');

const sequelize = new Sequelize({
  database: 'model_issue',
  username: 'root',
  password: 'qwe123456',
  host: '127.0.0.1',
  dialect: 'mysql',
  define: { timestamps: false }
});

//#region GUID
function GUID () {
  if (!(this instanceof GUID)) { return new GUID(); }
}
inherits(GUID, DataTypes.ABSTRACT);
Object.assign(GUID, DataTypes.ABSTRACT);

GUID.prototype.key = GUID.key = 'GUID'
GUID.prototype.toSql = function toSql () { return 'BINARY(16)'; }
GUID.prototype._stringify = function _stringify (guid) {
  if (!guid) {
    return null;
  } else {
    return guidToBuffer(guid);
  }
}
GUID.prototype._sanitize = function _sanitize (value) {
  if (!value) {
    return null;
  } else {
    if (typeof value === 'string') {
      return value;
    } else if (Buffer.isBuffer(value)) {
      return bufferToGuid(value);
    }
  }
}
//#endregion

class Foo extends Model {}
Foo.init({
  id: { type: GUID, primaryKey: true, defaultValue: guid },
  text: { type: DataTypes.STRING }
}, {
  sequelize,
  tableName: 'foo'
})

async function main() {
  await Foo.truncate();

  const actual = 'a';
  const expected = 'b';

  const foo = await Foo.create({ text: actual });

  // Fails
  foo.text = expected;
  await foo.save();
  await foo.reload();
  console.log('Updated: ', foo.text === expected); // false

  // Success
  foo.text = expected;
  // HACK options: Abstract QueryGenerator::whereItemQuery calls _findField that get primary key field from model
  await foo.save({ model: Foo });
  await foo.reload();
  console.log('Updated: ', foo.text === expected); // true
}

main()
```
**To Reproduce**
1. git clone https://github.com/SnosMe/sequelize-model-save.git
2. npm i
3. node ./index.js

Sequelize uses `DataType._stringify` method.

It cannot because `Model.save()` dont set `model` prop on `options`.

## Possible fix
This worked for me, all `npm run test-mysql` tests passed.
```patch
--- ./model.js	2019-05-12 21:34:15.429035300 +0300
+++ ./model.js	2019-05-12 21:34:23.165273100 +0300
@@ -3655,7 +3655,8 @@
     options = _.defaults(options, {
       hooks: true,
-      validate: true
+      validate: true,
+      model: this.constructor
     });
 
     if (!options.fields) {
```

## Environment
Dialect:
- [x] mysql
- [ ] postgres
- [ ] sqlite
- [ ] mssql
- [ ] any
Dialect **mysql2** version: 1.6.5
Database version: 8.0.14
Sequelize version: 5.8.6
Node Version: v10.15.0
OS: Win10

SnosMe:
### Pull Request check-list

### Description of change

Closes #10941

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -470,9 +470,7 @@ class BelongsToMany extends Association {
       instances = [instances];
     }
 
-    options = Object.assign({
-      raw: true
-    }, options, {
+    options = Object.assign({}, options, {
       scope: false,
       attributes: [this.target.primaryKeyAttribute],
       joinTableAttributes: []
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -470,7 +470,9 @@ class BelongsToMany extends Association {
       instances = [instances];
     }
 
-    options = Object.assign({}, options, {
+    options = Object.assign({
+      raw: true
+    }, options, {
       scope: false,
       attributes: [this.target.primaryKeyAttribute],
       joinTableAttributes: []
--- a/test/integration/associations/belongs-to-many.test.js
+++ b/test/integration/associations/belongs-to-many.test.js
@@ -557,6 +557,28 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), () => {
       });
     });
 
+    it('compares primary keys using attribute type info', function() {
+      return Promise.all([
+        this.Article.create({ title: 'Article' }),
+        this.Label.create({ text: 'Awesomeness' }),
+        this.Label.create({ text: 'Epicness' })
+      ]).then(([article, label1, label2]) => {
+        return article.setLabels([label1, label2]).then(() => {
+          DataTypes.INTEGER.prototype._sanitize = function _sanitize(value) { return value; };
+          sinon.spy(DataTypes.INTEGER.prototype, '_sanitize');
+          this.Label.refreshAttributes();
+          return article.hasLabels([
+            label1[this.Label.primaryKeyAttribute],
+            label2[this.Label.primaryKeyAttribute]
+          ]).then(() => {
+            expect(DataTypes.INTEGER.prototype._sanitize).calledTwice;
+          }).finally(() => {
+            delete DataTypes.INTEGER.prototype._sanitize;
+          });
+        });
+      });
+    });
+
     it('answers true for labels that have been assigned multitple times', function() {
       this.ArticleLabel = this.sequelize.define('ArticleLabel', {
         id: {
--- a/test/integration/associations/belongs-to-many.test.js
+++ b/test/integration/associations/belongs-to-many.test.js
@@ -557,23 +557,23 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), () => {
       });
     });
 
-    it('compares primary keys using attribute type info', function() {
+    it('options.raw default value can be overwritten', function() {
       return Promise.all([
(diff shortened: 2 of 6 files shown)


lib/associations/belongs-to-many.js
```diff
@@ -470,9 +470,7 @@ class BelongsToMany extends Association {
       instances = [instances];
     }
 
-    options = Object.assign({
-      raw: true
-    }, options, {
+    options = Object.assign({}, options, {
```
papb: Why was `raw: true` removed?

test/integration/associations/belongs-to-many.test.js
```diff
@@ -557,6 +557,28 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), () => {
...
     });
 
+    it('compares primary keys using attribute type info', function() {
+      return Promise.all([
+        this.Article.create({ title: 'Article' }),
+        this.Label.create({ text: 'Awesomeness' }),
+        this.Label.create({ text: 'Epicness' })
+      ]).then(([article, label1, label2]) => {
+        return article.setLabels([label1, label2]).then(() => {
+          DataTypes.INTEGER.prototype._sanitize = function _sanitize(value) { return value; };
```
papb: If in the future a `_sanitize` function gets added to the `INTEGER` datatype for some reason, you will be overwriting it and deleting it at the end of the test, potentially messing up other tests. Please alter this to make it work safely if a `_sanitize` function exists on the `INTEGER` prototype.

papb:
Also, to be honest, I didn't understand how the test you added ensures your fix works. Can you explain what's the connection between your test and [your original problem](https://github.com/sequelize/sequelize/issues/10941#issue-443134345)?

test/integration/associations/belongs-to-many.test.js
SnosMe: fixed

lib/associations/belongs-to-many.js
SnosMe: I believe our way of using a custom data type is almost a hack.

image

Since multiple DataTypes use a common type over DB Binary Protocol, we cannot use `.parse` and do the trick in `_sanitize`. 

Now I wonder why typecasting uses the `column type` of Binary Protocol and not the DataType  defined by Model (`column name` = `Model schema attribute`). Is it because of support for raw queries, aliases, sql functions, etc ?

papb:
I think I finally understand what your PR is doing. You want the custom `._sanitize()` function created on a custom datatype to be called correctly when `.save()` is called, right?

test/integration/model.test.js
```diff
@@ -844,6 +844,32 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         .then(() => User.findByPk(94))
         .then(user => expect(user.id2).to.equal(8877));
     });
+
+    it('should use PK attribute DataType', function() {
```
papb: Rename this to something more clear, such as:

```suggestion
    it('should call stringify from PK datatype correctly upon save', function() {
```

WikiRik:

PR #11293, opened 2019-08-07
fix(types): return a usable type when using the sequelize.models lookup

schmod:
### Pull Request check-list

### Description of change

Currently, looking up model classes via the `sequelize.models` interface does not yield an object with a usable type signature (for reasons that I admittedly do not entirely understand).

Static functions such as `findAll` and `belongsTo` do not work on the values pulled off of `sequelize.models`, presumably because these methods use the `ModelCtor` "hack" to infer the correct return type.

```ts
class Model1 extends Model {}

const M1 = sequelize.models.model1;
m1.findAll({});  // <-- ERROR

/**
The 'this' context of type 'typeof Model' is not assignable to method's 'this' of type 'ModelCtor<Model<unknown, unknown>>'.
  Type 'typeof Model' is not assignable to type 'new () => Model<unknown, unknown>'.
    Cannot assign an abstract constructor type to a non-abstract constructor type.ts(2684)
/**
```

It appears that these methods _can_ be called if `sequelize.models` is redefined to contain `ModelCtor`s instead of `typeof Model`.

### DIFF
--- a/types/test/sequelize.ts
+++ b/types/test/sequelize.ts
@@ -49,4 +49,8 @@ Sequelize.afterConnect(() => {
 
 const rnd: Fn = sequelize.random();
 
-const myModel: typeof Model = sequelize.models.asd;
+class Model1 extends Model{}
+class Model2 extends Model{}
+const myModel: typeof Model1 = sequelize.models.asd;
+myModel.hasOne(Model2)
+myModel.findAll();
--- a/types/lib/sequelize.d.ts
+++ b/types/lib/sequelize.d.ts
@@ -19,6 +19,7 @@ import {
   UpdateOptions,
   WhereAttributeHash,
   WhereOperators,
+  ModelCtor,
 } from './model';
 import { ModelManager } from './model-manager';
 import * as Op from './operators';
@@ -767,7 +768,7 @@ export class Sequelize extends Hooks {
    * Dictionary of all models linked with this instance.
    */
   public readonly models: {
-    [key: string]: typeof Model;
+    [key: string]: ModelCtor<Model>;
   };
 
   /**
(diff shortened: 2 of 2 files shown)


types/test/sequelize.ts
```diff
@@ -49,4 +49,8 @@ Sequelize.afterConnect(() => {
 
 const rnd: Fn = sequelize.random();
 
-const myModel: typeof Model = sequelize.models.asd;
```
papb: I think you should keep this, just rename it so it does not conflict with your new lines below. Or this no longer works?

schmod: Nah, it still works.  I can re-add if you like, although I was assuming that the line below would adequately cover the same use-case:

```ts
const myModel: typeof Model1 = sequelize.models.asd;
```

schmod: That being said, it's not really a valid use-pattern, because you'd end up running into the issues that we're trying to address, because we're casting to a less-flexible type. 

```ts
const myModel: typeof Model = sequelize.models.asd;
myModel.findAll(); // <-- TS error
```

papb: Ah, ok :+1: so keep it removed. Thanks for the explanation!

schmod:
Yeah, FWIW, I wouldn't mind if we added some more guidance on how to best use Sequelize in a typesafe manner, particularly with regard to model definitions.  

A bunch of the tricks that folks normally use to manage associations without introducing cyclical dependencies don't quite work in the TS world.

papb:
@schmod What kind of tricks are you referring to? Can you show an example? (No rush, and not relevant to this PR)

PR #11294, opened 2019-08-07
QueryInterface.removeColumn(): fix parameter desc

waldyrious:
"Columns" was improperly pluralized.


sushantdhiman:
@papb You can freely merge PR's like this :)

Issue #11223, opened 2019-07-25
Updating using bulkCreate does not returning the pimary key for new record


VidhiyaElango:
## What are you doing?

```js
let data = [{id:1, name:'name1', value:'value1'},{id:2, name:'name2', value},{id:'', name:'name3',value:'value3'}];
Model.bulkCreate(data, {updateOnDuplicate:['name', 'value']}).then((response)=>{console.log(response)})
```
**To Reproduce**
model:
 {
        id:  {
                type: Sequelize.INTEGER,
                primaryKey: true,
                autoIncrement: true,
                unique : true
            },
            name: Sequelize.TEXT,
            value :  Sequelize.TEXT
    }
## What do you expect to happen?
To return the primary key for all objects

## What is actually happening?
I have inserted the first 2 data earlier and while updating have added the 3rd object to the array.
while getting response i am getting primary key's for first 2 data and null for newly added object.
response :
 [ {id:1, name:'name1', value:'value1'},{id:2, name:'name2', value},{id:**null**, name:'name3',value:'value3'} ]

## Environment
Dialect: -  mysql
Dialect **library** version: ^2.16.0
Sequelize version: ^5.2.0
Node Version: 10.15.3
OS: linux

papb:
Can you check with another dialect, such as SQLite for instance?

VidhiyaElango:
Hi papb,
     I have tried with SQLite too and the same problem exits.
     But i have found a solution by settting the value of the id in the model

Model :
{
    id: {
        type: Sequelize.INTEGER,
        primaryKey: true,
        autoIncrement: true,
        unique: true
         set() {
                    let value = this.getDataValue('id')
                    this.setDataValue('id', value);
          },
    },
    name: Sequelize.TEXT,
    value: Sequelize.TEXT
}


Thank you for your response.

VidhiyaElango:
But if i set the value in the model it is duplicating all the records and not updating

papb:
I see...

You posted only a code snippet, can you provide a [SSCCE](http://sscce.org/) (also known as [MCVE/reprex](https://stackoverflow.com/help/minimal-reproducible-example)) to easily reproduce it all by copy-pasting?

sushantdhiman:
MySQL only returns ids for records which got inserted with `updateOnDuplicate`. I don't think we can reliably support this unless MySQL adds support for `RETURNING *` equivalent

```js
const Sequelize = require('./index');
[...]

PR #11307, opened 2019-08-10
feat: support include option in bulkInsert

### LINKED ISSUE #3277: Is there any option for create bulk asociations?
I need to create bulk instances with associated options (n:m relation) and I want to insert this options in bulk too.

There's any option to avoid single insert?
There's any option to query directly the join table?

Thanks.

javiertury:
# Pull Request check-list

### Description of change

Closes #3277.

This PR makes `bulkInsert` capable of handling associations. It uses the `include` option, just like other model methods.

It requires only 1 insert statement for each association model. E.g. 100 instances with 100 children and 100 grandchildren are created in 3 insert statements.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2530,163 +2530,313 @@ class Model {
       return Promise.resolve([]);
     }
 
-    options = Object.assign({
-      validate: false,
-      hooks: true,
-      individualHooks: false,
-      ignoreDuplicates: false
-    }, options);
-
-    options.fields = options.fields || Object.keys(this.rawAttributes);
-
     const dialect = this.sequelize.options.dialect;
-
-    if (options.ignoreDuplicates && ['mssql'].includes(dialect)) {
-      return Promise.reject(new Error(`${dialect} does not support the ignoreDuplicates option.`));
-    }
-    if (options.updateOnDuplicate && (dialect !== 'mysql' && dialect !== 'mariadb' && dialect !== 'postgres')) {
-      return Promise.reject(new Error(`${dialect} does not support the updateOnDuplicate option.`));
-    }
-
-    if (options.updateOnDuplicate !== undefined) {
-      if (Array.isArray(options.updateOnDuplicate) && options.updateOnDuplicate.length) {
-        options.updateOnDuplicate = _.intersection(
-          _.without(Object.keys(this.tableAttributes), this._timestampAttributes.createdAt),
-          options.updateOnDuplicate
-        );
-      } else {
-        return Promise.reject(new Error('updateOnDuplicate option only supports non-empty array.'));
-      }
-    }
+    const now = Utils.now(this.sequelize.options.dialect);
 
     options.model = this;
 
-    const createdAtAttr = this._timestampAttributes.createdAt;
-    const updatedAtAttr = this._timestampAttributes.updatedAt;
-    const now = Utils.now(this.sequelize.options.dialect);
-
-    let instances = records.map(values => this.build(values, { isNewRecord: true }));
-
-    return Promise.try(() => {
-      // Run before hook
-      if (options.hooks) {
-        return this.runHooks('beforeBulkCreate', instances, options);
+    if (!options.includeValidated) {
+      this._conformIncludes(options, this);
+      if (options.include) {
+        this._expandIncludeAll(options);
+        this._validateIncludedElements(options);
       }
-    }).then(() => {
-      // Validate
-      if (options.validate) {
-        const errors = new Promise.AggregateError();
-        const validateOptions = _.clone(options);
-        validateOptions.hooks = options.individualHooks;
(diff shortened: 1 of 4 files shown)


papb:
Hey, thanks for the PR. Can you show a code snippet showing the usage as an example?

javiertury:
The same as regular `create` method.

```
        const Product = this.sequelize.define('Product', {
          title: Sequelize.STRING
        });
        const Tag = this.sequelize.define('Tag', {
          name: Sequelize.STRING
        });

        Product.hasMany(Tag);

        return this.sequelize.sync({ force: true }).then(() => {
          return Product.bulkCreate([{
            id: 1,
            title: 'Chair',
            Tags: [
              { id: 1, name: 'Alpha' },
              { id: 2, name: 'Beta' }
            ]
          }, {
            id: 2,
            title: 'Table',
            Tags: [
              { id: 3, name: 'Gamma' },
              { id: 4, name: 'Delta' }
            ]
          }], {
            include: [{
              model: Tag,
              myOption: 'option'
            }]
          }).then(savedProducts => {
            // ...
          });
```

sushantdhiman:
Looks good, can you update typings https://github.com/sequelize/sequelize/blob/master/types/lib/model.d.ts#L707

lib/model.js
```diff
@@ -2530,163 +2530,313 @@ class Model {
...
+          return model.QueryInterface.bulkInsert(model.getTableName(options), records, options, fieldMappedAttributes).then(results => {
+            if (Array.isArray(results)) {
+              results.forEach((result, i) => {
+                const instance = instances[i];
+
+                for (const key in result) {
+                  if (!instance || key === model.primaryKeyAttribute &&
+                    instance.get(model.primaryKeyAttribute) &&
+                    ['mysql', 'mariadb', 'sqlite'].includes(dialect)) {
+                    // The query.js for these DBs is blind, it autoincrements the
```
sushantdhiman: For MySQL I have seen this behaviour with updateOnDuplicate https://github.com/sequelize/sequelize/issues/11223#issuecomment-515669115

javiertury:
@sushantdhiman typings added

lib/model.js
javiertury: > I don't think we can reliably support this unless MySQL adds support for RETURNING * equivalent

That is exactly the root of the problem. PostgreSQL and SQL Server both support `RETURNING` and therefore don't have any issue.

papb:
> I haven't given new code a deep look but new tests looks good and old one passed :)

Same from me! :+1:

simonbrent:
Thanks a lot for implementing this feature, it will be very helpful.

I have one observation regarding it though - the options passed to recursiveBulkCreate are as follows:
```            
const includeOptions =  _(Utils.cloneDeep(include))
              .omit(['association'])
              .defaults({
                transaction: options.transaction,
                logging: options.logging
              }).value();
```
but this means that if I want to validate at every stage of creation, I cannot just set `validate: true` as a top level option - instead I must set `validate: true` both at the top level and in every include.

Would you be willing to either allow the includeOptions to default validate to options.validate, or maybe if that is undesirable, support a validateAll option, which is used to set options.validate to true at the start of recursiveBulkCreate?

javiertury:
I see your point about `validation`. I'm now wondering why `validation` is disabled by default. Is `bulkCreate` assumed to not receive user input? Is it about performance?

@simonbrent, are you using `bulkCreate` with user input?

simonbrent:
@javiertury yes I am. 
I've got a workaround for now in a beforeBulkCreate hook, which modifies options.include to set `validate: options.validate` for each entry in the include array.

I assume validation is disabled by default because it can be expensive.

marinalohova:
Hi, I'm getting
"instance[include.association.accessors.set] is not a function"

My code is as following:
await this.models.products.bulkCreate(products, {
        updateOnDuplicate: ['name', 'description', 'subscriptionId', 'creditId'],
        returning: true,
        include: {
          model: this.models.subscriptions,
          as: 'subscription',
          updateOnDuplicate: ['subscriptionFee', 'subscriptionReoccurrence'],
          returning: true,
        },
      });
What could be the issue? Thanks!

jeffersonveloso:
> Hi, I'm getting
> "instance[include.association.accessors.set] is not a function"
> 
> My code is as following:
> 
> ```
> await this.models.products.bulkCreate(products, {
>         updateOnDuplicate: ['name', 'description', 'subscriptionId', 'creditId'],
>         returning: true,
>         include: {
>           model: this.models.subscriptions,
>           as: 'subscription',
>           updateOnDuplicate: ['subscriptionFee', 'subscriptionReoccurrence'],
>           returning: true,
>         },
>       });
> ```
> 
> What could be the issue? Thanks!

Have any solution for this problem ? @srslyimthebest

PR #11311, opened 2019-08-12
Feature: source/target keys in belongsToMany associations

### LINKED ISSUE #4092: Specify non-primary target key in belongsToMany association

swarthy:
### Pull Request check-list

### Description of change

Closes #4092

### DIFF
--- a/docs/associations.md
+++ b/docs/associations.md
@@ -531,6 +531,70 @@ Person.belongsToMany(Person, { as: 'Children', through: 'PersonChildren' })
 
 ```
 
+#### Source and target keys
+
+If you want to create a belongs to many relationship that does not use the default primary key some setup work is required.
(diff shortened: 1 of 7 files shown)


swarthy:
Need some help with test coverage, what I'm doing wrong :smile: ?

papb:
> 
> 
> Need some help with test coverage, what I'm doing wrong smile ?

I don't see anything wrong with the coverage :+1:

Issue #11090, opened 2019-06-20
`bulkCreate`'s `updateOnDuplicate` option default does not work


ozhi:
## What are you doing?
I am creating entities in bulk with `bulkCreate`. In case of a conflict on the primary key, I want that entity to be updated with the new values I have given.

```js
await Asset.destroy({ where: {} });

await Asset.create({
    id: 1,
    uuid: 'uuid-1',
    name: 'asset-1',
});

await Asset.bulkCreate([
    {
        id: 2,
        uuid: 'uuid-2',
        name: 'asset-2',
    },
    {
        id: 1,
        uuid: 'uuid-updated',
        name: 'asset-updated',
    },
],{
    // updateOnDuplicate: ['name'],
});
```
**To Reproduce**
1. Define model `Asset`:
```js
Asset.init(
	{
		id: {
			type: Sequelize.INTEGER,
			primaryKey: true,
		},
		uuid: {
			type: Sequelize.UUID,
		},
		name: {
			type: Sequelize.STRING,
			allowNull: false,
		},
	},
	{
		sequelize,
	}
);
```
2. Run the code above
3. See a `SequelizeUniqueConstraintError` 

## What do you expect to happen?
I wanted 
```js
const assets = await Asset.findAll({ where: {} });
console.log(assets.map(a => a.toJSON()));
```
to produce:
```
[ { id: 1,
    uuid: 'uuid-updated',
    name: 'asset-updated',
    createdAt: 2019-06-20T13:04:53.000Z,
    updatedAt: 2019-06-20T13:04:53.000Z },
  { id: 2,
    uuid: 'uuid-2',
    name: 'asset-2',
    createdAt: 2019-06-20T13:04:53.000Z,
    updatedAt: 2019-06-20T13:04:53.000Z } ]
```

## What is actually happening?
The mentioned code produces
```
SequelizeUniqueConstraintError: Validation error
      at Query.formatError (node_modules/sequelize/lib/dialects/mysql/query.js:218:16)
      at Query.handler [as onResult] (node_modules/sequelize/lib/dialects/mysql/query.js:46:23)
      at Query.execute (node_modules/mysql2/lib/commands/command.js:30:14)
      at Connection.handlePacket (node_modules/mysql2/lib/connection.js:449:32)
      at PacketParser.Connection.packetParser.p [as onPacket] (node_modules/mysql2/lib/connection.js:72:12)
[...]

PR #11321, opened 2019-08-15
Fix documentation to represent true operation and workaround for desi…

JamesJansson:
…red operation

Addresses issue raised in https://github.com/sequelize/sequelize/issues/11090

### Pull Request check-list

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2518,7 +2518,7 @@ class Model {
    * @param  {boolean}      [options.hooks=true]             Run before / after bulk create hooks?
    * @param  {boolean}      [options.individualHooks=false]  Run before / after create hooks for each individual Instance? BulkCreate hooks will still be run if options.hooks is true.
    * @param  {boolean}      [options.ignoreDuplicates=false] Ignore duplicate values for primary keys? (not supported by MSSQL or Postgres < 9.5)
-   * @param  {Array}        [options.updateOnDuplicate]      Fields to update if row key already exists (on duplicate key update)? (only supported by MySQL, MariaDB & Postgres >= 9.5). By default, all fields are updated.
+   * @param  {Array}        [options.updateOnDuplicate]      Fields to update if row key already exists (on update of record with existing primary key). Only supported by MySQL, MariaDB & Postgres >= 9.5). Is undefined by default, which results in a SequelizeUniqueConstraintError. To update all fields, use `updateOnDuplicate=Object.keys(Model.attributes)` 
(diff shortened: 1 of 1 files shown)


lib/model.js
```diff
@@ -2518,7 +2518,7 @@ class Model {
    * @param  {boolean}      [options.hooks=true]             Run before / after bulk create hooks?
    * @param  {boolean}      [options.individualHooks=false]  Run before / after create hooks for each individual Instance? BulkCreate hooks will still be run if options.hooks is true.
    * @param  {boolean}      [options.ignoreDuplicates=false] Ignore duplicate values for primary keys? (not supported by MSSQL or Postgres < 9.5)
-   * @param  {Array}        [options.updateOnDuplicate]      Fields to update if row key already exists (on duplicate key update)? (only supported by MySQL, MariaDB & Postgres >= 9.5). By default, all fields are updated.
+   * @param  {Array}        [options.updateOnDuplicate]      Fields to update if row key already exists (on update of record with existing primary key). Only supported by MySQL, MariaDB & Postgres >= 9.5). Is undefined by default, which results in a SequelizeUniqueConstraintError. To update all fields, use `updateOnDuplicate=Object.keys(Model.attributes)` or `updateOnDuplicate=Object.keys(Model.rawAttributes).filter(attr => attr !== 'createdAt' && attr !== 'updatedAt' && attr !== 'deletedAt')`
```
sushantdhiman: > Is undefined by default

No need for this

> which results in a SequelizeUniqueConstraintError. To update all fields, use `updateOnDuplicate=Object.keys(Model.attributes)` or `updateOnDuplicate=Object.keys(Model.rawAttributes).filter(attr => attr !== 'createdAt' && attr !== 'updatedAt' && attr !== 'deletedAt')`
 
What if their are virtual attributes?

Please use this instead

```
By default no field is updated, which can result in unique constraint error.
```

JamesJansson: There is no need, except that this issue has already been raised once before and it has been made pretty clear that there would be no in-built support something like updateOnDuplicate='*'. I agree virtual attributes could be a problem, but it would be better to leave a comment about being careful about virtual attributes rather than leaving people high and dry on what is a pretty fundamental use of DBs (upserting many records).

sushantdhiman:
We can support `updateOnDuplicate=*` with `updateOnDuplicate = true` option, which can be as simple as ```updateOnDuplicate = Object.keys(this.tableAttributes)```

Please open a new PR if you want to make this change :)

Issue #11316, opened 2019-08-13
Virtual column ends up in SQL query when associated model has default scope with attributes configuration


bparan:
## What are you doing?

```js
const Sequelize = require('sequelize');

const sequelize = new Sequelize('sqlite::memory:');

const Game = sequelize.define('Game', {
    name: Sequelize.TEXT,
    version: {
        type: Sequelize.VIRTUAL,
        get() {
            return 1;
        }
    }
});

const User = sequelize.define('User', {
    login: Sequelize.TEXT,
    session: {
        type: Sequelize.VIRTUAL,
        get() {
            return 'New';
        }
    }
}, {
    defaultScope: {
        attributes: {
            exclude: ['login']
            /*include: ['login']*/
        }
    }
});

Game.hasMany(User);
await sequelize.sync({force: true});

const games = await Game.findAll({
    include: [{
        model: User
    }]
});

console.log(games.length);
```
**To Reproduce**
1. Define model to be associated with defaultScope that includes `attributes` configuration
2. Define a parent model
3. Associate the parent model with child model using `hasMany`.
4. Query the parent records including the association.

## What do you expect to happen?
_On empty tables I'd expect `0` to be printed to the console_

## What is actually happening?
_I got SequelizeDatabaseError_

_SequelizeDatabaseError: SQLITE_ERROR: no such column: Users.session
      at Query.formatError (C:\Proj\node_modules\sequelize\lib\dialects\sqlite\query.js:419:16)
      at Query._handleQueryResponse (C:\Proj\node_modules\sequelize\lib\dialects\sqlite\query.js:73:18)
      at afterExecute (C:\Proj\node_modules\sequelize\lib\dialects\sqlite\query.js:247:31)
      at replacement (C:\Proj\node_modules\sqlite3\lib\trace.js:19:31)
      at Statement.errBack (C:\Proj\node_modules\sqlite3\lib\sqlite3.js:16:21)_

## Environment
Dialect:
- [ ] mysql
- [x] postgres
- [x] sqlite
- [x] mssql
- [ ] any
Dialect **library** version: pg@7.0.0, pg-hstore@2.3.3, sqlite3@4.0.9, tedious@6.0.0
Database version: sqlite3, PostgreSQL 11.4, MSSQL 2017
Sequelize version: 5.12.1
Node Version: 10.13.0
OS: Windows 10
If TypeScript related: TypeScript version: XXX
Tested with latest release:
- [ ] No
[...]

PR #11327, opened 2019-08-17
fix(query-generator): handle virtual column on associations with scopes

bparan:
### Pull Request check-list

### Description of change

Resolves issue described in #11316. When compiling list of attributes to be included as part of SQL query the initial list is cleaned up from virtual column and then updated with column aliases. Before this change only column aliases worked correctly. Pull request doesn't include new test but there is a change to test/integration/model/scope/find.test.js file that adds a virtual column to `DefaultScopeExclude` model that has scope with `attributes` config already defined.
I think the test case `should work when included with default scope` works to prevent regression here.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1449,7 +1449,7 @@ class QueryGenerator {
     // includeIgnoreAttributes is used by aggregate functions
     if (topLevelInfo.options.includeIgnoreAttributes !== false) {
       include.model._expandAttributes(include);
-      Utils.mapOptionFieldNames(include, include.model);
+      Utils.mapFinderOptions(include, include.model);
 
       const includeAttributes = include.attributes.map(attr => {
         let attrAs = attr;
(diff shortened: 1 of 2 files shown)


papb:
Thanks a lot for the PR!

> Pull request doesn't include new test but there is a change to test/integration/model/scope/find.test.js file that adds a virtual column to `DefaultScopeExclude` model that has scope with `attributes` config already defined.
I think the test case `should work when included with default scope` works to prevent regression here.

You may be right (I didn't really check deeply), but regardless I think a dedicated test is deserved. To an outsider, the extra virtual field you added there might look like unused code (or at least something strange), and someone might remove it in a refactor or just get confused unnecessarily.

I think you should add a test replicating your MCVE from your issue and asserting that `games.length` is equal to 0.

bparan:
@papb Thanks for your comment. I added a separate regression test. You are right the initial approach test approach doesn't not protect from accidental refactoring. The separate test makes it easier to catch the issue in future.

papb:
LGTM

Issue #7879, opened 2017-07-03
Sequelize 4.0: Remove timezone on database options


Joanne16:
Hi, how do we remove/bypass the timezone on sequelize options (dialect mysql) ? With version 3.0 setting timezone: '' did the trick and the dates where not modified. With sequelize 4.0, setting timezone to blank string throws an error... What option can be used so that the server date is used without any timezone applied ?

Thanks

Shadowman4205:
@Joanne16  Did you find solution for this?

Joanne16:
Hi @Shadowman4205 , the solution I found is to put a timezone in the options, else the wrong time is given.

Shadowman4205:
@Joanne16  That's a pity, i will stuck for previous version for a while cause i do not want timezone at all. Cause of that [issue:7930](https://github.com/sequelize/sequelize/issues/7930)

Joanne16:
@Shadowman4205 Yep. I hope there will be another way of disabling the timezone option. As putting the value as '+00:00' (the default) still modifies the dates returned from queries.

cladjules:
that may look ugly but, in sequelize/lib/data-types.js, you can replace the method:

```
DATE.prototype._stringify = function _stringify(date, options) {
  date = this._applyTimezone(date, options);

  // Z here means current timezone, _not_ UTC
  // return date.format('YYYY-MM-DD HH:mm:ss.SSS Z');
  return date.format('YYYY-MM-DD HH:mm:ss.SSS');
};
```

Shadowman4205:
So i have to go into node-modules, data-types.js, and modify it or fork this lib? Yes this is messy for simple option. :/

cladjules:
I copied the file locally, applied the change and I did a postinstall script in package.json which copies the file into node_modules, that will run automatically after every npm install.

In package.json:
```
  "scripts": {
    "postinstall": "./fixme.sh",
```

fixme.sh
```
#!/bin/sh
cp data-types.js ./node_modules/sequelize/lib/data-types.js
```

dlredden:
I'm so thankful to @cladjules for his solution to this problem. I thought I'd share how I solved this in my singleton file. I like this a little better because the change is still ubiquitous in my app and I didn't have to modify a node_modules file.

In db.js:
```javascript
import Sequelize from 'sequelize';
import settings from '../settings';

// Override timezone formatting
Sequelize.DATE.prototype._stringify = function _stringify(date, options) {
[...]

Issue #11177, opened 2019-07-10
Custom datatype not working


Bene-Graham:
## What are you doing?

```ts
// Custom Type File
import Big from "big.js";

export default function sequelizeAdditions(Sequelize) {

    const DataTypes = Sequelize.DataTypes;

    /*
     * Create new types
     */
    class BIGJS extends DataTypes.DECIMAL {
        public static parse(value) {
            return new Big(value);
        }

        private _precision: number;
        private _scale: number;

        constructor(precision: number, scale: number) {
            super();

            this._precision = precision;
            this._scale = scale;
        }

        public toSql() {
            if (this._precision || this._scale) {
                return `DECIMAL(${[this._precision, this._scale].filter(this.identity).join(",")})`;
            }

            return "DECIMAL";
        }

        // Optional, validator function
        public validate(value, options) {
            if (value instanceof Big) {
                return true;
            } else {
                return false;
            }
        }
    }

    DataTypes.BIGJS = BIGJS;

    // Mandatory, set key
    DataTypes.BIGJS.prototype.key = DataTypes.BIGJS.key = "BIGJS";

    // For convenience
    // `classToInvokable` allows you to use the datatype without `new`
    Sequelize.BIGJS = Sequelize.Utils.classToInvokable(DataTypes.BIGJS);
}

// Connection File
import { Sequelize } from "sequelize";
import BigJsCol from "./types/BigJsColumn";

BigJsCol(Sequelize);

const sequelize = new Sequelize(...);

// Model File
MyModel.init({
    id: {
        type: DataTypes.INTEGER.UNSIGNED,
        autoIncrement: true,
        primaryKey: true,
    },
    amount: new DataTypes.BIGJS(13, 4)
}, {
    sequelize: Connection,
    tableName: "MyModel",
});

// Simple Query
await MyModel.findAll();
```
**To Reproduce**
1. Defined models, and get everything wired up
2. Run simple query
3. Result for the field amount is not wrapped in a BigJs object

## What do you expect to happen?
I would like the result coming back from the database to be wrapped

## What is actually happening?
It is just returning a number :(

## Environment
Dialect:
- [x] mysql
- [ ] postgres
- [ ] sqlite
- [ ] mssql
- [ ] any
Dialect mysql2 version: 1.6.5
Database version: 5.7.26
Sequelize version: 5.9.3
Node Version: 12.x
OS: Mac OS 10.14.5
If TypeScript related: TypeScript version: 3.5.3
[...]

PR #11344, opened 2019-08-21
Add SMALLDATETIME type for MSSQL

moberegger:
### Pull Request check-list

### Description of change

Adds a `SMALLDATETIME` type to be used by MSSQL. 

If the MSSQL database is setup to use `SMALLDATETIME`, the Sequelize `DATE` type doesn't work as expected because it includes a timezone, which gets dropped by the database. The only way `SMALLDATETIME` differs from `DATE` is that the former forces UTC on write.

Solution inspired from https://github.com/sequelize/sequelize/issues/7879#issuecomment-329971730 and a previous attempt that was done in https://github.com/sequelize/sequelize/pull/10167. The solution proposed in this PR differs slightly, though; this PR uses [`toISOString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) to ensure UTC, rather than specifying a specific format.

TODO:
[ ] Add additional test(s) to ensure proper `stringify`ing

### DIFF
--- a/docs/data-types.md
+++ b/docs/data-types.md
@@ -32,6 +32,7 @@ Sequelize.DECIMAL(10, 2)              // DECIMAL(10,2)
 Sequelize.DATE                        // DATETIME for mysql / sqlite, TIMESTAMP WITH TIME ZONE for postgres
 Sequelize.DATE(6)                     // DATETIME(6) for mysql 5.6.4+. Fractional seconds support with up to 6 digits of precision
 Sequelize.DATEONLY                    // DATE without time.
+Sequelize.SMALLDATETIME               // SMALLDATETIME datatype for MSSQL
 Sequelize.BOOLEAN                     // TINYINT(1)
 
 Sequelize.ENUM('value 1', 'value 2')  // An ENUM with allowed values 'value 1' and 'value 2'
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -497,6 +497,21 @@ class DATEONLY extends ABSTRACT {
 }
 
 /**
+ * A small date time column.
+ *
+ * Only available for MSSQL
+ */
+class SMALLDATETIME extends DATE {
+  toSql() {
+    return 'SMALLDATETIME';
+  }
+  _stringify(date, options) {
+    const timeZoneDate = this._applyTimezone(date, options);
+    return timeZoneDate.toISOString();
+  }
+}
+
+/**
  * A key / value store column. Only available in Postgres.
  */
 class HSTORE extends ABSTRACT {
@@ -993,6 +1008,7 @@ const DataTypes = module.exports = {
   TIME,
   DATE,
   DATEONLY,
+  SMALLDATETIME,
   BOOLEAN,
   NOW,
   BLOB,
--- a/lib/dialects/mssql/data-types.js
+++ b/lib/dialects/mssql/data-types.js
@@ -27,6 +27,7 @@ module.exports = BaseTypes => {
    */
 
   BaseTypes.DATE.types.mssql = [43];
+  BaseTypes.SMALLDATETIME.types.mssql = [58];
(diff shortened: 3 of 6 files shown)


moberegger:
I have added some preliminary tests already, but I'd really like to add something to ensure a proper write to the DB. I looked around the existing tests around the DATE type and got quite lost.

Any recommendations of where a good place to write this test would be? Or existing test(s) to use as inspiration?

moberegger:
I have no idea how to get the new test to pass:

```
1616 passing (4m)
  1 pending
  1 failing
  1) [MSSQL] DataTypes
       calls parse and stringify for SMALLDATETIME:
     AssertionError: expected spy to have been called at least once, but it was never called
      at null.<anonymous> (node_modules\sinon-chai\lib\sinon-chai.js:83:18)
      at propertyGetter (node_modules\chai\lib\chai\utils\addProperty.js:62:29)
      at Object.proxyGetter [as get] (node_modules\chai\lib\chai\utils\proxify.js:98:22)
      at current.sync.then.then.then (test\integration\data-types.test.js:92:33)
      at C:\projects\sequelize\node_modules\continuation-local-storage\context.js:84:17
  From previous event:
      at Promise.then (node_modules\cls-bluebird\lib\shimMethod.js:38:20)
      at testSuccess (test\integration\data-types.test.js:91:8)
      at Context.it (test\integration\data-types.test.js:165:14)
```

I haven't defined a `parse` function anywhere, and I don't see the other Date types having a `parse` either (although those ones pass).

Is there date-specific logic somewhere that checks for certain data types that I need to include `SMALLDATETIME` to?

papb:

> this PR uses [`toISOString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) to ensure UTC, rather than specifying a specific format.

Seems a good idea!

> TODO:
> [ ] Add additional test(s) to ensure proper `stringify`ing

Can you clarify what you meant by this?

> I have added some preliminary tests already, but I'd really like to add something to ensure a proper write to the DB. I looked around the existing tests around the DATE type and got quite lost.
> 
> Any recommendations of where a good place to write this test would be? Or existing test(s) to use as inspiration?

Don't worry about the place for now, you can place it anywhere and later I will suggest a location (I don't have much time right now, sorry). Don't worry about "git history" being weird because everything will be squashed upon merging anyway.

As for how to do it, I am not sure I understood your question correctly, but wouldn't simply defining a basic model, calling `create`, then `findOne` and checking if the field is properly filled suffice?

papb:
As for [your last comment](https://github.com/sequelize/sequelize/pull/11344#issuecomment-523635024), I don't have the time to investigate right now, sorry... It would be great if you could dig into it further for me!

sushantdhiman:
Have you considered a custom data type for this? https://sequelize.org/master/manual/data-types.html#extending-datatypes

papb:
@sushantdhiman he did look at it in #11177

moberegger:
Closing this for a few reasons.

>If the MSSQL database is setup to use SMALLDATETIME, the Sequelize DATE type doesn't work as expected because it includes a timezone, which gets dropped by the database.

As far as I can tell, that is no longer true. This looks like it was an issue at some point, but this no longer seems to happen. Perhaps the `tedious` driver takes care of this for us now.

The MSSQL documentation also [recommends](https://docs.microsoft.com/en-us/sql/t-sql/data-types/smalldatetime-transact-sql?view=sql-server-ver15) to use types other than SMALLDATETIME:
> Use the time, date, datetime2 and datetimeoffset data types for new work. These types align with the SQL Standard. They are more portable. time, datetime2 and datetimeoffset provide more seconds precision. datetimeoffset provides time zone support for globally deployed applications. 

I don't think it makes sense to include support for a data type that shouldn't be used. This feels like something more appropriately handled by a custom data type.

Issue #11183, opened 2019-07-11
Updating only updatedAt on a model does not cause an update, even with silent: true


jamelling:
## What are you doing?
```js
person = Person.create({ name: 'Original Name'})
await Person.update({ updatedAt: new Date('02-26-1991')}, {where: { id: person.id }, silent: true})
// [ 0 ]

await person.reload()
await person.updatedAt
// 2019-04-12T20:13:52.923Z

await Person.update({ updatedAt: new Date('02-26-1991'), name: 'A New Name'}, {where: { id: person.id }, silent: true})
Executing (default): UPDATE "Person" SET "updatedAt"=$1,"name"=$2 WHERE "id" = $3
// [ 1 ]

await person.reload()
person.updatedAt
// 1991-02-26T05:00:00.000Z
```
**To Reproduce**
1. Create a model `Person` and an instance `person`
2. Use `Person.update` to update only the updatedAt field on person
3. Reload `person` to see that updatedAt has not been updated

## What do you expect to happen?
I expected updatedAt to be updated,

## What is actually happening?
updatedAt was not updated

## Environment
Dialect:
- [ ] mysql
- [ ] postgres
- [ ] sqlite
- [ ] mssql
- [x] any
Dialect **library** version: XXX
Database version: PostgreSQL 9.6
Sequelize version: 5.10.0
Node Version:  10.15.3
OS: Ubuntu 18.04.1 on Windows 10
Tested with latest release:
- [ ] No
- [x] Yes, specify that version: 5.10.0

papb:
What is the use case for updating only the updatedAt field?

jamelling:
I only use it while doing tests for some filtering of data by updatedAt times.

Ladvace:
I have the same problem

wangzhilongh:
I have the same problem

papb:
Duplicate of #3759

PR #11348, opened 2019-08-23
feat(findbypks): add method and some initials tests

### LINKED ISSUE #11297: Add findByPks convenience method
A common thing we do is searching for model that match a given id set. This is not that hard with the following clause: `Model.findAll({where: {id: {[Op.in]: [1, 2, 3]}}})`, however it might be easier if a `findByPks` method was added so that one would do `Model.findByPks([1,2,3])` and/or `Model.findByPks(1,2,3)`.

Rails has this: https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find

TypeORM has this: https://typeorm.io/#/entity-manager-api Search for findByIds

I think if this will be added, we should investigate what are convenience methods could be added. For example, there is an open issue about adding `exists`: https://github.com/sequelize/sequelize/issues/10187

fboechats:
### Pull Request check-list

### Description of change

This is a initial idea to create a new method called findByPks that will receive an array of ids and get all models using those ids.

Fixes #11297

### DIFF
--- a/types/lib/sequelize.d.ts
+++ b/types/lib/sequelize.d.ts
@@ -97,6 +97,11 @@ export interface PoolOptions {
   acquire?: number;
 
   /**
+   * The time interval, in milliseconds, after which sequelize-pool will remove idle connections.
+   */
+  evict?: number;
+
+  /**
    * A function that validates a connection. Called with client. The default function checks that client is an
    * object, and that its state is not disconnected
    */
--- a/types/test/sequelize.ts
+++ b/types/test/sequelize.ts
@@ -15,6 +15,9 @@ export const sequelize = new Sequelize({
     match: ['hurr'],
   },
   dialectModule: {},
+  pool: {
+    evict: 1000,
+  }
 });
 
 const conn = sequelize.connectionManager;


lib/model.js
```diff
@@ -1885,6 +1885,35 @@ class Model {
...
+ *
+ * @returns {Promise<Model>}
+ */
+  static findByPks(param) {
+  // return Promise resolved with null if no arguments are passed
+    if ([null, undefined].includes(param)) {
+      return Promise.resolve(null);
+    }
+
+    if (typeof param === 'number' || typeof param === 'number') {
```
marclerodrigues: Why are you comparing this twice?

marclerodrigues:
Missing documentation on how to use this method.

lib/model.js
```diff
@@ -1885,6 +1885,35 @@ class Model {
...
+      throw new Error('The argument passed to findByPks must be an array of ids'); 
+    }
+
+    if (param.length >= 2 && Array.isArray(param) || Array.isArray(param) && param.length === 0) {
+      return this.findAll({ where: { id: param } });
```
marclerodrigues: In the original issue the code given as example is `Model.findAll({where: {id: {[Op.in]: [1, 2, 3]}}})` are you sure this line behaves in the same way?

papb:
Hi @fboechats, I agree with the review by @marclerodrigues above (thanks!)

lib/model.js
```diff
@@ -1885,6 +1885,35 @@ class Model {
...
+    } if (param.length == 1) {
```
marclerodrigues: I don't see how this is useful, if I want to use `findByPks([1])` I should be able to. IMHO.

lib/model.js
```diff
@@ -1885,6 +1885,35 @@ class Model {
...
+ * {@link Model.findAll}           for a full explanation of options, Note that options.where is not supported.
+ *
+ * @returns {Promise<Model>}
+ */
+  static findByPks(param) {
+  // return Promise resolved with null if no arguments are passed
+    if ([null, undefined].includes(param)) {
+      return Promise.resolve(null);
+    }
+
```
marclerodrigues: As far as I can see, the code below this line could be refactored to something like:

```javascript
if (!Array.isArray(param)) {
  throw new Error('The argument passed to findByPks must be an array of ids');
}

 return this.findAll({ where: { id: { [Op.in]: param } } });
```

marclerodrigues: @papb please let me know your thoughts on this.

papb: Completely agree!

lib/model.js
```diff
@@ -1885,6 +1885,29 @@ class Model {
...
+  /**
+ * Search for a many instances by its primary key._
+ *
+ * @param  {Array}      param The value of the desired instance's primary key.
+ *
+ * @see
+ * {@link Model.findAll}           for a full explanation of options, Note that options.where is not supported.
+ *
+ * @returns {Promise<Model>}
+ */
```
papb: Please fix indentation here (two extra spaces per line needed) and remove these big blank spaces like here:

```
@param  {Array}      param
```

Also the method returns `Promise<Array<Model>>`

lib/model.js
```diff
@@ -1885,6 +1885,29 @@ class Model {
...
+  static findByPks(param) {
+  // return Promise resolved with null if no arguments are passed
+    if ([null, undefined].includes(param)) {
+      return Promise.resolve(null);
+    }
```
papb: Please remove this `if`, I think it's better to throw the error from the line below instead of silently resolving with null.

test/integration/model/findByPks.test.js
```diff
@@ -0,0 +1,82 @@
...
+      theDate: DataTypes.DATE,
+      aBool: DataTypes.BOOLEAN,
+      binary: DataTypes.STRING(16, true)
+    });
+
+    return this.User.sync({ force: true });
+  });
+
+  describe('findByPks', () => {
+    if (current.dialect.supports.transactions) {
```
papb: The tests for `findByPks` should not involve transactions, please rewrite the tests removing transaction stuff

test/integration/model/findByPks.test.js
```diff
@@ -0,0 +1,82 @@
...
+  beforeEach(function() {
+    this.User = this.sequelize.define('User', {
+      username: DataTypes.STRING,
+      secretValue: DataTypes.STRING,
+      data: DataTypes.STRING,
+      intVal: DataTypes.INTEGER,
+      theDate: DataTypes.DATE,
+      aBool: DataTypes.BOOLEAN,
+      binary: DataTypes.STRING(16, true)
+    });
```
papb: I think you should use a simpler model, something like `this.User = this.sequelize.define('User', { username: DataTypes.STRING })` should suffice!

test/integration/model/findByPks.test.js
```diff
@@ -0,0 +1,82 @@
...
+    if (current.dialect.supports.transactions) {
+      it('supports transactions', function() {
+        return Support.prepareTransactionTest(this.sequelize).then(sequelize => {
+          const User = sequelize.define('User', { 
+            id: {
+              type: Sequelize.INTEGER,
+              primaryKey: true
+            },
+            username: Sequelize.STRING 
+          });
```
papb: This definition of `User` is not necessary, you can just use the `this.User` you defined above

test/integration/model/findByPks.test.js
```diff
@@ -0,0 +1,82 @@
...
+              });
+            });
+          });
+        });
+      });
+    }
+
+    it('should not crash on an empty where array', function() {
+      return this.User.findByPks([]);
+    });
```
papb: This test is OK, but I think you can improve it by checking that the result is an empty array (instead of just checking if it doesn't throw)

test/integration/model/findByPks.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+    it('should be able to get info from a user', function() {
```
papb: You can remove this test, it is redundant comparing to the first test you wrote

sushantdhiman:
Thanks for this work but I don't think I want to add this feature to Sequelize directly. It is a simple additional method that should be supported as a plugin.

You can publish this as a plugin and we can that to our [plugins directory](https://github.com/sequelize/sequelize/blob/master/docs/resources.md#addons--plugins)

sushantdhiman:
Documentation changes included in this PR are valuable, you can remove `findByPks` related stuff so we can merge them

fboechats:
@papb Do you guys have any template to follow?? I've never created a plugin before.

papb:
@fboechats Unfortunately no... I was planning to create one indeed.

In the meantime, I suggest you take a look in [sequelize-hierarchy](https://github.com/overlookmotel/sequelize-hierarchy/) by overlookmotel. Feel free to further ask me questions in this PR if you need help.

Thanks again for your time to help Sequelize! It is much appreciated.

fboechats:
@papb Hello, I'm really not sure about this, but when you got some time can you take a look in this initial idea?? https://github.com/fboechats/sequelize-findbypks

papb:
@fboechats It's a good start, but I have suggestions, can I open PRs in your repository or would you prefer me to simply outline them here? Also take a look into enabling Travis CI (or another CI if you prefer) in your repository so that it runs tests automatically (like Sequelize itself does).

fboechats:
You can do as you wish.

papb:
@fboechats - Ok! I will open a PR there later.

In the meantime, as sushantdhiman said, can you please remove the `findByPks` specific stuff from this PR, converting it into a general documentation improvement, we would like to merge the documentation changes by you

docs/models-usage.md
```diff
@@ -2,32 +2,36 @@
...
-// search for known ids
+// search for known id
 Project.findByPk(123).then(project => {
   // project will be an instance of Project and stores the content of the table entry
   // with id 123. if such an entry is not defined you will get null
-})
+});
+
+// search for known array of ids
+Project.findByPk([1, 2, 3]).then(project => {
```
papb: Remove this part, since it doesn't work

docs/models-usage.md
```diff
@@ -2,32 +2,36 @@
...
-// search for known ids
+// search for known id
 Project.findByPk(123).then(project => {
   // project will be an instance of Project and stores the content of the table entry
   // with id 123. if such an entry is not defined you will get null
-})
+});
+
+// search for known array of ids
+Project.findByPks([1, 2, 3]).then(project => {
```
papb: This will only work if the user installs your plugin, so this should not be present in the sequelize main documentation, please remove this, this PR will become a docs-only change

papb:
@fboechats What happened?

fboechats:
@papb You told me to remove, don't you??

papb:
Not the PR! I wanted you to remove only the line 19 from docs/models-usage.md...

fboechats:
But that's the only modification I did in that file, added `findByPks` usage.

papb:
Really? Look [at this](https://github.com/sequelize/sequelize/commit/1e3c1317675aeffebab102fb1c62e450e44e06bf)

fboechats:
That's problably eslint or something like this, if you want I can keep it then.

papb:
Hmm, that makes sense. I think we don't need it then I guess, I am also working at the docs right now so it would just make things more messy.

By the way, thanks again for your time! I will try to look at your plugin later. I am very busy though, feel free to ping me (@papb) here if it feels like I forgot.

fboechats:
@papb Hello, should we try to implement the plugin??

papb:
Hello! Sorry for disappearing :( I [am assigned to almost 100 issues](https://github.com/sequelize/sequelize/issues/assigned/papb) already, unfortunately I won't be able to help you in the near future. I am really sorry, I wish I had more time to dedicate to Sequelize :grimacing: 

By the way, when you came forward to help with this, were you specifically interested in this feature or just chose it as something to help? If it's the latter, I might be able to point you to other things that I could use some help :)

fboechats:
No problem at all. Just choose to help and also learn, if you have the time to send me other things it would be great.

papb:
Nice. Thanks. How about creating an [SSCCE](https://github.com/papb/sequelize-sscce) for issue #11183?

fboechats:
Sure, woking on it.

Issue #9035, opened 2018-02-11
belongsTo+DefaultScope while findAndCountAll got Unknown column error


wddd:
## What are you doing?

```js
const Sequelize = require('sequelize');
const sequelize = require('../db');
const Enterprise = require('./enterprise');

const User = sequelize.define('pm_users',
    {
        id: {
            type: Sequelize.INTEGER,
            primaryKey: true,
            autoIncrement: true
        },
        name: {
            type: Sequelize.STRING(50),
        },
        account: {
            type: Sequelize.STRING(50),
            unique: 'account'
        },
        email: {
            type: Sequelize.STRING(50),
            unique: 'email'
        },
        role: {
            type: Sequelize.INTEGER
        },
        phone: {
            type: Sequelize.STRING(50)
        },
        password: {
            type: Sequelize.STRING(255)
        }
    }, {
        freezeTableName: true,
        defaultScope: {
           attributes: { exclude: ['password', 'createdAt', 'updatedAt'] }
        },
        scopes: {
            userPassword: {
                attributes: ['id', 'email', 'role', 'password']
            }
        }
    });

// User-Enterprise
Enterprise.hasMany(User, { as: 'users', foreignKey: 'enterpriseId' });
User.belongsTo(Enterprise, { as: 'enterprise'});
```
**Execute code belows**
```js
User.findAndCountAll({
                    where: where,
                    order: [['id', 'ASC']],
                    include: [
                        {
                            model: Enterprise, as: 'enterprise'
                        }
                    ],
                    limit: limit,
                    offset: offset,
                    distinct: true
                })
```
## What is actually happening?
**Got this Error**
`"SequelizeDatabaseError: Unknown column 'pm_users.enterpriseId' in 'on clause'"`

check the generated sql, i found enterpriseId filed is not appear in select statement.

- if remove the defaultScope , the query will success.
- if append a filed on user model will also resolve this problem

```js
 enterpriseId:{
     type: Sequelize.INTEGER,
 },
```
I don't know maybe this is a feature or a Bug, but it really confuse me for a while
Apologize for my bad english.😸

- Dialect: mysql
- Database version: 5.7.0
- Sequelize version: 4.31.2
- Tested with latest release: yes

sushantdhiman:
How you are creating this table, as you have referenced `Enterprise` with `User` by `enterpriseId`, you need to define this in your table as well

wddd:
@sushantdhiman 
[...]

PR #11385, opened 2019-09-04
fix(include): fix rare bug involving subqueries on combination of multiple includes

wertlex:
### Pull Request check-list

- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones) (no updates)?
- [x] Did you update the typescript typings accordingly (if applicable)? (not applicable)

### Description of change
Fixes https://github.com/sequelize/sequelize/issues/9035
Fixes abstract query generator, so all supported database engines are affected.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1174,9 +1174,11 @@ class QueryGenerator {
       // appending fields which are required by 'BelongsTo' associations when main attributes filter is in use
       if (options.include) {
         for (const include of options.include) {
-          const { associationType, foreignKey } = include.association;
-          if (associationType === 'BelongsTo' && attributes.main.indexOf(foreignKey) === -1) {
-            attributes.main.push(foreignKey);
+          if (include.association && include.association.associationType && include.association.foreignKey) {
+            const { associationType, foreignKey } = include.association;
+            if (associationType === 'BelongsTo' && attributes.main.indexOf(foreignKey) === -1) {
+              attributes.main.push(foreignKey);
+            }
           }
         }
       }
@@ -1700,7 +1702,7 @@ class QueryGenerator {
         const joinSource = `${asLeft.replace(/->/g, '.')}.${attrLeft}`;
 
         // Check for potential aliased JOIN condition
-        joinOn = this._getAliasForField(asLeft, joinSource, topLevelInfo.options) || this.quoteIdentifier(joinSource);        
+        joinOn = this._getAliasForField(asLeft, joinSource, topLevelInfo.options) || this.quoteIdentifier(joinSource);
       }
     }
 
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1170,6 +1170,16 @@ class QueryGenerator {
           attributes.main.push(mainTable.model.rawAttributes[keyAtt].field ? [keyAtt, mainTable.model.rawAttributes[keyAtt].field] : keyAtt);
         }
       }
+
+      // appending fields which are required by 'BelongsTo' associations when main attributes filter is in use
+      if (options.include) {
+        for (const include of options.include) {
+          const { associationType, foreignKey } = include.association;
+          if (associationType === 'BelongsTo' && attributes.main.indexOf(foreignKey) === -1) {
+            attributes.main.push(foreignKey);
+          }
+        }
+      }
     }
 
     attributes.main = this.escapeAttributes(attributes.main, options, mainTable.as);
@@ -1965,7 +1975,7 @@ class QueryGenerator {
 
     if (Array.isArray(options.order)) {
       for (let order of options.order) {
-        
+
         // wrap if not array
         if (!Array.isArray(order)) {
           order = [order];
@@ -1989,7 +1999,7 @@ class QueryGenerator {
           const subQueryAttribute = options.attributes.find(a => Array.isArray(a) && a[0] === order[0] && a[1]);
           if (subQueryAttribute) {
             const modelName = this.quoteIdentifier(model.name);
-            
+
             order[0] = new Utils.Col(this._getAliasForField(modelName, subQueryAttribute[1], options) || subQueryAttribute[1]);
           }
         }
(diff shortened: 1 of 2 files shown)


papb:

test/integration/include/findAll.test.js
```diff
@@ -2091,5 +2091,23 @@ describe(Support.getTestDialectTeaser('Include'), () => {
         });
       });
     });
+
+    it('should be able to generate a correct request for entity with 1:n and m:1 associations and limit', function() {
+      return this.fixtureA().then(() => {
+        return this.models.Product.findAll({
+          where: {},
```
papb: Is this `where: {}` necessary?

test/integration/include/findAll.test.js
```diff
@@ -2091,5 +2091,23 @@ describe(Support.getTestDialectTeaser('Include'), () => {
...
+          attributes: ['title'],
+          include: [
+            { model: this.models.User },
+            { model: this.models.Price }
+          ],
+          limit: 10
+        }).then( products => {
+          // checking that internally added fields used to handle 'BelongsTo' associations are not leaked to result
+          expect(products[0].UserId).to.be.equal(undefined);
+          expect(products[0].title).to.be.a('string');
```
papb: Can you add more `expect`s to guarantee the expected query result? Just to make the test a bit more robust. Checking only the type of `title` of the first result seems weak

lib/dialects/abstract/query-generator.js
```diff
@@ -1170,6 +1170,16 @@ class QueryGenerator {
           attributes.main.push(mainTable.model.rawAttributes[keyAtt].field ? [keyAtt, mainTable.model.rawAttributes[keyAtt].field] : keyAtt);
         }
       }
+
+      // appending fields which are required by 'BelongsTo' associations when main attributes filter is in use
+      if (options.include) {
+        for (const include of options.include) {
+          const { associationType, foreignKey } = include.association;
```
papb: I am not very familiar with this part of the code, how can we be 100% sure that `include.association` will always be defined here (and be an object)? Recall that `const { x, y } = undefined;` gives `TypeError: Cannot destructure property `x` of 'undefined' or 'null'`. I would be more confident if this was wrapped in an `if (include.association)`... What do you think?

test/integration/include/findAll.test.js
wertlex: Yeah, there was no need in it.

test/integration/include/findAll.test.js
wertlex: I've added a bit more `expect`s to check `products` shape more precisely.

lib/dialects/abstract/query-generator.js
wertlex: Good point. Updated it to _paranoid_ mode

papb: @wertlex Excellent!

papb:
Nice! Another question: you claim to be fixing #9035 (and I believe you), but it is not immediately clear to me from the PR modifications themselves. The OP from #9035 posted something about `defaultScope` being relevant to the bug, and I don't see anything on your PR about this. Can you explain to me a bit further what exactly is going on?

wertlex:
Sure!
It looks like #9035 's topic starter problem is not in `scopes` itself, but in `attributes` value set in scopes. Note **Got this Error** part: "SequelizeDatabaseError: Unknown column 'pm_users.enterpriseId' in 'on clause'". Error is related not to scope, but to missed column.
So, skipping scopes and getting down to queries itself.

Let's consider Sequelize and corresponding SQL queries from test: https://github.com/sequelize/sequelize/pull/11385/files#diff-916a03a19d49cecbcd62820c088b9587R2095

Will start from something a bit simpler. Given Sequelize query: 
```
 return this.models.Product.findAll({
   attributes: ['title'],
   include: [
     { model: this.models.User },
     { model: this.models.Price }
   ]
})
```

will produce following SQL:
```
SELECT 
  "Product"."id", "Product"."title", "User"."id" AS "User.id", "User"."createdAt" AS "User.createdAt", "User"."updatedAt" AS "User.updatedAt", "Prices"."id" AS "Prices.id", "Prices"."value" AS "Prices.value", "Prices"."createdAt" AS "Prices.createdAt", "Prices"."updatedAt" AS "Prices.updatedAt", "Prices"."ProductId" AS "Prices.ProductId" 
  FROM "Products" AS "Product" 
    LEFT OUTER JOIN "Users" AS "User" ON "Product"."UserId" = "User"."id" 
    LEFT OUTER JOIN "Prices" AS "Prices" ON "Product"."id" = "Prices"."ProductId";
```

Important thing here is that it does not use subquery - it is just quite straight `SELECT * FROM` query with couple of joins.

But obviously it cannot handle `limit` well. So when Sequelize query stuffed with `limit`:
```
 return this.models.Product.findAll({
   attributes: ['title'],
   include: [
     { model: this.models.User },
     { model: this.models.Price }
   ],
   limit: 10
})
```
SQL query turned into a little bit different thing:
```
SELECT 
  "Product".*, "User"."id" AS "User.id", "User"."createdAt" AS "User.createdAt", "User"."updatedAt" AS "User.updatedAt", "Prices"."id" AS "Prices.id", "Prices"."value" AS "Prices.value", "Prices"."createdAt" AS "Prices.createdAt", "Prices"."updatedAt" AS "Prices.updatedAt", "Prices"."ProductId" AS "Prices.ProductId" 
  FROM (SELECT "Product"."id", "Product"."title" FROM "Products" AS "Product" LIMIT 10) AS "Product" 
    LEFT OUTER JOIN "Users" AS "User" ON "Product"."UserId" = "User"."id" 
    LEFT OUTER JOIN "Prices" AS "Prices" ON "Product"."id" = "Prices"."ProductId";
```
Note that subquery at `FROM (SELECT ...)` - it have `LIMIT 10` set as expected, but also it enumerates attributes passed in `attributes` property of Sequelize query. But next line (`LEFT OUTER JOIN "Users" AS "User" ON "Product"."UserId" = "User"."id"`) requires `"Product"."UserId"` field to perform join. And this field obviously is not in fields obtained for `Products`, which is effectively gives broken query.

This pr detects foreign keys for such relationships and pushes them into collection of attributes of main query, similar to primary key (which is also not passed in `attributes`, still appears in SQL query but not in result)

sushantdhiman:
Thanks for this PR, I have investigated this issue and applied a different fix https://github.com/sequelize/sequelize/pull/11404

Please let me know if that works for you!

wertlex:
Yeah, thank you. 
It works now :)

P.S. sorry for late reply

PR #11402, opened 2019-09-08
fix(enum): escape values

warkenji:
Enum doesn't escape values on "Create Table"


papb:

PR #11406, opened 2019-09-09
feat: added support for multiple primary key attributes

### LINKED ISSUE #11392: Add types for primaryKeyAttributes into Model class
Please add `primaryKeyAttributes` to Model class to allow access to primaryKeyAttributes from TypeScript without errors.

sequelize/types/lib/model.d.ts
```typescript
export abstract class Model<T = any, T2 = any> extends Hooks {
  /** The name of the database table */
  public static readonly tableName: string;

  /**
   * The name of the primary key attribute
   */
  public static readonly primaryKeyAttribute: string;

  /**
   * The name of the primary key attributes
   */
  public static readonly primaryKeyAttributes: string[];
```

nainkunal933:
added support for multiple primary key attributes in sequelize/types/lib/model.d.ts

closes #11392

### Pull Request check-list

### DIFF
--- a/types/test/model.ts
+++ b/types/test/model.ts
@@ -67,19 +67,8 @@ UserModel.findCreateFind({
   }
 })
 
-class ModelCompositePrimaryKey extends Model {}
-
-ModelCompositePrimaryKey.init({
-  uid: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
-  primaryKeyOne: { type: DataTypes.STRING, primaryKey: true, allowNull: false },
-  title: DataTypes.STRING,
-  description: DataTypes.TEXT,
-}, { sequelize, modelName: 'ModelCompositePrimaryKey' });
-
-ModelCompositePrimaryKey
-  .create({ primaryKeyOne: "Type 1", title: 'Test Record', description: 'Record created to test composite primary key' })
-  .then(modelCompositePrimaryKey => {
-    console.log(modelCompositePrimaryKey.get('title'));
-    console.log(modelCompositePrimaryKey.get('description'));
-    console.log(ModelCompositePrimaryKey.primaryKeyAttributes);
-  })
+/**
+ * Test for primaryKeyAttributes.
+ */
+class TestModel extends Model {};
+TestModel.primaryKeyAttributes;
--- a/types/test/model.ts
+++ b/types/test/model.ts
@@ -67,3 +67,19 @@ UserModel.findCreateFind({
   }
 })
 
+class ModelCompositePrimaryKey extends Model {}
+
+ModelCompositePrimaryKey.init({
+  uid: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
+  primaryKeyOne: { type: DataTypes.STRING, primaryKey: true, allowNull: false },
+  title: DataTypes.STRING,
+  description: DataTypes.TEXT,
+}, { sequelize, modelName: 'ModelCompositePrimaryKey' });
+
+ModelCompositePrimaryKey
+  .create({ primaryKeyOne: "Type 1", title: 'Test Record', description: 'Record created to test composite primary key' })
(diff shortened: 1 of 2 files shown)


papb:
Hello! Thank you for the PR!

Please add a test for the typing you added. If you don't know how to do it, please take a look on other recently merged PRs with the `type: typescript` label. Feel free to ping me if you have questions.

nainkunal933:
Hi @papb, I was going through the test cases in the previous PR. I am planning to add the test case in ```types/test/model.ts```. The current model has only one index 'foo'. How do I add another index in the array? I do not know the other fields of the table.
Current Model: 
Snip20190909_10
 I am trying to do the following:
```typescript
indexes: [
    {
      fields: ['foo'],
      using: 'gin',
      operator: 'jsonb_path_ops',
    },
   {
      fields: ['COLUMN_NAME_HERE'],
      using: 'gin',
      operator: 'jsonb_path_ops',
    }
  ],
```
If this works, then it will test whether the DB will accept composite primary keys with multiple columns. Please let me know if I am on the right track. Thank you.

papb:
It's simpler than this, you just have to add some code that compiles with your PR and wouldn't compile without it.

nainkunal933:
Hi @papb, I have added the test case in mode.ts. Please let me know if it is what you expected. 

Sorry for my ignorance. I am new to open source. Thank you for helping me.

nainkunal933:
Also, in terms of documentation, the [model definition page](https://sequelize.org/master/manual/models-definition.html#indexes) has documentation for multiple indexes. So, for now, I will consider the documentation update to be done.
Snip20190909_11

nainkunal933:
Hi @papb, any updates on this PR.

papb:
Hi, thanks for the heads up, somehow your PR got lost among my notifications.

I think you are a bit confused on the intent your own PR. Recall that this is merely an issue with the typescript typings provided by sequelize, which means that the feature itself already exist and works, the only issue is that when someone tries to do `MyModel.primaryKeyAttributes`, the typescript compiler will complain, since it doesn't know that models have a `primaryKeyAttributes` field. The word 'test' might have confused you here (and surely the contributing guidelines deserve an update, I am working on that already by the way!), but a 'test' in the context of typescript typings is just a code that compiles. What it does when executed is irrelevant.

So, in summary, your test is too complicated. Here is a super mini test that would do the job:

```ts
class TestModel extends Model {};
TestModel.primaryKeyAttributes;
```

The above code compiles with your PR and does not compile without it, and it is very short, which is great. Do you see that your test was unnecessarily complicated? Please update your PR with this simpler code.

papb:
> Also, in terms of documentation, the model definition page has documentation for multiple indexes. So, for now, I will consider the documentation update to be done.

Although it is true that this PR does not need any documentation update (since it is just a typings update), again I see that you are a bit lost (don't worry!) - what you've posted about indexes is not related to the task at hand. You're confusing indexes with foreign keys; indexes are another thing.

types/test/model.ts
```diff
@@ -67,3 +67,19 @@ UserModel.findCreateFind({
...
+  description: DataTypes.TEXT,
+}, { sequelize, modelName: 'ModelCompositePrimaryKey' });
+
+ModelCompositePrimaryKey
+  .create({ primaryKeyOne: "Type 1", title: 'Test Record', description: 'Record created to test composite primary key' })
+  .then(modelCompositePrimaryKey => {
+    console.log(modelCompositePrimaryKey.get('title'));
+    console.log(modelCompositePrimaryKey.get('description'));
+    console.log(ModelCompositePrimaryKey.primaryKeyAttributes);
+  })
```
papb: As I said above, please simplify all this to

```ts
class TestModel extends Model {};
TestModel.primaryKeyAttributes;
```

nainkunal933: Hi @papb, I understand my test was too complicated. Thank you for the thorough explanation. I have made the changes. Please let me know if the changes are ok.

PR #11409, opened 2019-09-09
fix(mssql): removed the extended DATEONLY class from mssql dialect

### LINKED ISSUE #10982: DATEONLY in MSSQL converted to localtime
## What are you doing?
Using DATEONLY, when querying the database directly the date is '2019-06-01', but Sequelize returns '2019-05-31'. I'm in the US Central timezone, so I assume it is still treating it as a UTC date and converting it to my timezone.

It is only returning the string in 'YYYY-MM-DD' format, which is good, but it is converting it based on my local time, which is bad.

Sequelize:  5.8.5
Tedious: 6.1.1
MS SQL Server 2017

Return the date string as stored without converting based on local timezone.

The date string is converted to local time, which displays as the previous day. '2019-06-01' is stored in the database, but '2019-05-31' is returned by Sequelize.

I rolled back to Sequelize 4.41.2 and Tedious 3.0.1 and the same date is returned as '2019-06-01T05:00:00.000Z'. So, it seems that DATEONLY has since been fixed to only return a date string without time and timezone, but it now assumes that the stored date is in UTC and incorrectly converts it to local time.

## Summary:
SQL Server DATEONLY Field Value: '2019-06-01'
Sequelize 5.8.5 returns: '2019-05-31'
Sequelize 4.41.2 returns: '2019-06-01T05:00:00.000Z'
My timezone: US Central Time (currently -5:00 with Daylight Saving Time)

## Environment
Dialect:
- [ ] mysql
- [ ] postgres
- [ ] sqlite
- [x] mssql
- [ ] any
Dialect **library** version: Tedious 6.1.1
Database version: MS SQL Server 2017
Sequelize version: 5.8.5
Node Version: 8.11.0
OS: Windows 10
Tested with latest release:
- [ ] No
- [x] Yes, specify that version: 5.8.6

dead:
### Pull Request check-list

### Description of change

Fix #10982 - Removed the DATAONLY extended class from MSSQL dialect.


dead:
@papb 
Two tests stopped working in the appveyor build.
Not sure how to proceed... I'm unable to run `npm run test-mssql` on my machine, for some reason it is getting timeout in the `getAssociations` test.

```
  1) [MSSQL] BelongsToMany
       getAssociations
         supports transactions:
     Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\Users\Gustavo\Desktop\sequelize\test\integration\associations\belongs-to-many.test.js)


  2) "after each" hook for "supports transactions":
     Error: Expected 0 running queries. 1 queries still running in [MSSQL] BelongsToMany getAssociations supports transactions
      at Context.<anonymous> (test\integration\support.js:24:9)
```

papb:
@dead Looks like the errors you posted aren't the same from the [PR build](https://ci.appveyor.com/project/sushantdhiman/sequelize/builds/27282371)?

krazi3:
@papb @dead This UTC issue with `DATEONLY` is also occurring for mysql test cases

emmiranda:
@papb 
@dead 

This commit solved my problem, in my case. It makes sense to convert without UTC to an attribute that is only date.
```
class DATEONLY extends BaseTypes.DATEONLY {
    static parse(value) {
      return moment(value).utc(false).format('YYYY-MM-DD');
    }
  }
```

Using DATEONLY, the date was recorded correctly, for example '2020-5-1'. However, when consulting the object, it resulted in '2020-4-1' (always one day less)
Something important is that I am using timezone setting "-03: 00" (I'm in Brazil).
`timezone: '-03:00'`

For now I had to change the original file inside node_modules.(\node_modules\sequelize\lib\dialects\mssql\data-types.js )

Could this pull request be approved to have this correction in a next version?

xjpro:
👍 This fixes an issue for me. How can we get it merged in?

bradwallen:
I second @xjpro, this fixes a problem for my team as well.

requius:
This fix all my problems with mssql and Sequelize.DATEONLY.

xjpro:
@dead Is this bit of code needed to make it pass the string test?

https://github.com/sequelize/sequelize/compare/master...xjpro:master#diff-a43588f59195933c926bea81581bf6aeR130

tonyguinta:
I have also confirmed that this change fixes the [original issue I reported](https://github.com/sequelize/sequelize/issues/10982), but has yet to be implemented in Sequelize 6.3.5:

```
class DATEONLY extends BaseTypes.DATEONLY {
    static parse(value) {
      return moment(value).utc(false).format('YYYY-MM-DD');
    }
  }
```
What needs to be done to have this included in the next release?

tonyguinta (approved):

tonyguinta (changes requested):
I haven't tried this code, but the code snippet from @emmiranda does work:
```
class DATEONLY extends BaseTypes.DATEONLY {
    static parse(value) {
      return moment(value).utc(false).format('YYYY-MM-DD');
    }
  }
```

tonyguinta (approved):
I have confirmed that this code solves [issue 10982](https://github.com/sequelize/sequelize/issues/10982).

spearmootz:
I have also tested this independently

spearmootz:
Meanwhile i have done the following and this will overwrite the parsing function to work correctly.

const moment = require('moment');
const sequelize = require('sequelize');

sequelize.DataTypes.mssql.DATEONLY.parse = function (value) {
  return moment.utc(value).format('YYYY-MM-DD');
};

quex46:
> Meanwhile i have done the following and this will overwrite the parsing function to work correctly.
> 
> const moment = require('moment');
> const sequelize = require('sequelize');
> 
> sequelize.DataTypes.mssql.DATEONLY.parse = function (value) {
> return moment.utc(value).format('YYYY-MM-DD');
> };

Yeah, it works well when we have proper dates in DB and just need to display it properly, but it also might need to patch `_stringify` and `_sanitize` methods as well. Otherwise, we will have (date-1d) set in DB in case of negative timezone. For example, if you set date to '2020-11-17' with America/New_York timezone (-05:00) it will be '2020-11-16' in DB.

So complete solution for me is:

```js
const moment = require('moment');
const sequelize = require('sequelize');

sequelize.DataTypes.mssql.DATEONLY.parse = function (value) {
  return moment.utc(value).format('YYYY-MM-DD');
};

sequelize.DataTypes.mssql.DATEONLY.prototype._stringify= function (value) {
  return moment.utc(value).format('YYYY-MM-DD');
};

sequelize.DataTypes.mssql.DATEONLY.prototype._sanitize= function (value) {
  if ((!options || (options && !options.raw)) && !!value) {
     return moment.utc(value).format('YYYY-MM-DD');
  }
  return value
};
```

An alternative approach - set default timezone for moment module used by sequelize:

```js
require("sequelize/node_modules/moment").tz.setDefault('UTC");
```

so, we won't need to patch any existing methods: internal sequelize `moment(...)` calls will work same as `moment.utc(...)`

Both approaches work for sequelize **6.3.5**

PR #11429, opened 2019-09-14
fix associations link

kdsonit:


papb:
Hello, thanks but your changes shouldn't apply, those links work as they are right now. You can see them in action [here](https://sequelize.org/master/).

kdsonit:
Please read and check on this page.
https://github.com/sequelize/sequelize/blob/master/docs/associations.md#scopes

Click on Scopes. link
it will redirect to https://github.com/sequelize/sequelize/blob/master/docs/scopes.html

404 Github page.

papb:
That's right, but the sequelize docs were not made to be navigated using github. You should navigate through the docs using the official docs website, http://sequelize.org, in which those links work correctly.

PR #11433, opened 2019-09-15
fix(postgres-bulkcreate): updateOnDuplicate uses correct PK column names

PeledYuval:
### Pull Request check-list

- [ ] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ ] Have you added new tests to prevent regressions?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? - no need
- [x] Did you update the typescript typings accordingly (if applicable)? - not applicable
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?

### Description of change

This pull request fixes the following issue:
https://github.com/sequelize/sequelize/issues/11432

When performing `bulkCreate` with `updateOnDuplicate` using Postgres dialect, the generated query has bad column names. The query has the Javascript model's field names instead of the column names.

Issue #11432, opened 2019-09-15
Postgres - bulkCreate with updateOnDuplicate fails when one of the model's primary key fields' name is different from its column name


PeledYuval:
## Issue Description
bulkCreate with updateOnDuplicate throws error when one of the model's primary key fields has a different name from its DB column name.

bulkCreate generates the field names in the: `ON CONFLICT (column1...)` part of the query by looking at: `this.primaryKeys.fieldName` which is the model's field name, instead of looking at `this.primaryKeys.field` which is the DB's column name.

The error received by Postgres and thrown by Sequelize is, expectedly: `Unhandled rejection SequelizeDatabaseError: column "fieldName" does not exist`

### What are you doing?
Minimal self contained reproducible example:
```js
const Sequelize = require('sequelize');

const sequelize = new Sequelize('postgres://postgres:postgres@localhost:5432/postgres');

class User extends Sequelize.Model {
}
User.init({
        userName: {
            type: Sequelize.STRING,
            primaryKey: true,
            field: 'user_name',
        },
        email: {
            type: Sequelize.STRING,
        }
    },
    { sequelize }
);

const user1 = {
    userName: 'John Smith',
};

User.sync({ force: true }).then(() =>
    User.bulkCreate([user1], {updateOnDuplicate: ['email']})
    // throws: Unhandled rejection SequelizeDatabaseError: column "userName" does not exist
);
```

### What do you expect to happen?
The code above should run without errors.

bulkCreate with updateOnDuplicate should work even when primary key fields are named differently in the DB.

### What is actually happening?
Sequelize throws: `Unhandled rejection SequelizeDatabaseError: column "userName" does not exist`

### Additional context
The code above works if the field name is changed to `user_name`.

The bug is in this line:
https://github.com/sequelize/sequelize/blob/d5e4a10f4e8dc1e37fe46678f04bbfa66b1299cc/lib/model.js#L2703

`fieldName` should be `field`.

A PR to fix this: https://github.com/sequelize/sequelize/pull/11434

### Environment

- Sequelize version: 5.8.14
- Node.js version: 8.15.0
- Operating System: macOS Mojave 10.14.6

### How does this problem relate to dialects?
- [ ] I think this problem happens regardless of the dialect.
- [x] I think this problem happens only for the following dialect(s): Postgres
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

[...]

PR #11434, opened 2019-09-15
fix(bulkcreate): use correct PK column names in updateOnDuplicate

PeledYuval:
Reopened accidentally closed PR: https://github.com/sequelize/sequelize/pull/11433

### Pull Request check-list

- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? - no need
- [x] Did you update the typescript typings accordingly (if applicable)? - not applicable

### Description of change

This pull request fixes the following issue:
https://github.com/sequelize/sequelize/issues/11432

When performing `bulkCreate` with `updateOnDuplicate` using Postgres dialect, the generated query has bad column names. The query has the Javascript model's field names instead of the column names.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2700,7 +2700,7 @@ class Model {
           if (options.updateOnDuplicate) {
             options.updateOnDuplicate = options.updateOnDuplicate.map(attr => model.rawAttributes[attr].field || attr);
             // Get primary keys for postgres to enable updateOnDuplicate
-            options.upsertKeys = _.chain(model.primaryKeys).values().map('fieldName').value();
+            options.upsertKeys = _.chain(model.primaryKeys).values().map('field').value();
             if (Object.keys(model.uniqueKeys).length > 0) {
               options.upsertKeys = _.chain(model.uniqueKeys).values().filter(c => c.fields.length === 1).map('column').value();
             }
--- a/test/integration/model/bulk-create.test.js
+++ b/test/integration/model/bulk-create.test.js
@@ -519,7 +519,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
 
           it('when the primary key column names and model field names are different', function() {
             const data = [
-              { plateNumber: 'abc', color: 'Black' },
+              { plateNumber: 'abc', color: 'Grey' },
               { plateNumber: 'def', color: 'White' }
             ];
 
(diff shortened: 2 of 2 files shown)


PeledYuval:
I'm debating where to add tests for this issue, since it relates to `model.js` and Postgres dialect:

- `test/unit/model/bulkCreate.test.js`? (but this is dialect-agnostic and my fix is PG centric)
- `test/unit/dialects/postgres/model.test.js`? (which would be a new file)
- `test/integration/model/bulkCreate.test.js`? (dialect-agnostic)
- `test/integration/dialects/postgres/model.test.js?` (which would be a new file)

Please advise and I will add appropriate tests.

papb:
Hello! I see you are a first-time contributor, thank you for taking the time to help Sequelize! I hope to see more PRs from you in the future! :grimacing: 

As for the test, I think you should do the third option. It makes sense to be dialect agnostic to me because updateOnDuplicate should work on all dialects it is supported. So although it works on MySQL and MariaDB already, it should work on every supported dialect (i.e. every dialect except MSSQL).

Therefore I think the best option is your third one (`test/integration/model/bulkCreate.test.js`). Thanks!!

papb:
I have tagged it as SQLite as well because I reproduced the same problem on SQLite. I would guess your suggested fix already fixes SQLite as well, but would you be willing to double check? And fix both in the same PR? Thank you very much!!

PeledYuval:
Sure, thanks for the quick response. I'm on it

PeledYuval:
Hi @papb,

I've added a test for this fix.
Since this is my first contribution, the idea was to keep the current tests unchanged. Since the current tests didn't a have a Model fit for this test (one that has a primary key with a different field name and column name) I created a new Model.

I also grouped two relevant tests under a single new `describe` - one preexisting test which I left unchanged (other than nesting in a `describe`) and my added test.

I ran the tests with Postgres, Sqlite and Mysql and they pass.
Happily, the test fails when my fix is reverted :)

Is there anything you'd like me to add/change before this can be merged?

PeledYuval:
I see the Travis build is failing but after a short investigation I don't see how it has anything to do with this PR.

I'm continuing to investigate, but on the chance that this is caused by a flaky test- could I trigger a rebuild? I couldn't find it in the Travis UI.

Thanks!

PeledYuval:
Alright, I pushed an inconsequential commit to trigger a rebuild (I was a little impatient...) and it does seem like a flaky test, the build passes now.

PR #11439, opened 2019-09-16
fix(mssql): filter tables only in mssql+mysql showTablesQuery

### LINKED ISSUE #11438: MS SQL showTablesQuery() returns VIEWs
## Issue Description
In the MS SQL dialect, `showTablesQuery()` returns a query that selects both tables and views.
This affects among others, sequelize-auto which uses showTablesQuery() to retrieve the list of tables for which to generate models. And Sequelize does not like when models exist for VIEWs.

### What are you doing?

```js
const db = new Sequelize(...)

db.query(`CREATE VIEW V_Fail AS SELECT 1 Id`)
  .then(() => db.getQueryInterface().showAllTables())
  .then(r => {
    if (r.filter(e => e.tableName == "V_Fail").length) {
      console.log("FAIL: V_Fail returned from showAllTables()")
      return 1
    } else {
      console.log("PASS")
      return 0
    }
  }, err => {
      console.log(err)
      return 1
  })
  .then(status => db.query(`DROP VIEW V_Fail`).then(r => status, e => status))
  .then(status => process.exit(status))
```

PASS in testcase. No VIEWs returned. This matches the behavior of the other dialects.

```
`FAIL: V_Fail returned from showAllTables()`
```

### Additional context
Culprit
https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/query-generator.js#L225

### Environment

- Sequelize version: master
- Node.js version: 12.10

mssql (Microsoft SQL Server)

joaoe:
Closes #11438

### Pull Request check-list

### Description of change

Filter tables by 'BASE TABLE' in `MSSQLQueryGenerator.showTablesQuery`.
Added tests.

### DIFF
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -149,7 +149,7 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
   }
 
   showTablesQuery() {
-    return 'SHOW TABLES;';
+    return "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = 'BASE TABLE';";
   }
 
   addColumnQuery(table, key, dataType) {
--- a/test/integration/query-interface.test.js
+++ b/test/integration/query-interface.test.js
@@ -44,6 +44,10 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
   describe('showAllTables', () => {
     it('should not contain views', function() {
       const cleanup = () => {
+        // NOTE: The syntax "DROP VIEW [IF EXISTS]"" is not part of the standard
+        // and might not be available on all RDBMSs. Therefore "DROP VIEW" is
+        // the compatible option, which can throw an error in case the VIEW does
+        // not exist. In case of error, it is ignored by reflect()+tap().
         return this.sequelize.query('DROP VIEW V_Fail').reflect();
       };
       return this.queryInterface
--- a/test/integration/query-interface.test.js
+++ b/test/integration/query-interface.test.js
@@ -41,6 +41,28 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
     });
   });
 
+  describe('showAllTables', () => {
+    it('should contain only tables', function() {
+      const cleanup = r => {
+        return this.sequelize.query('DROP VIEW V_Fail').then(() => r, () => r);
+      };
+      return this.queryInterface
+        .createTable('my_test_table', { name: DataTypes.STRING })
+        .then(cleanup)
+        .then(() => this.sequelize.query('CREATE VIEW V_Fail AS SELECT 1 Id'))
+        .then(() => this.queryInterface.showAllTables())
+        .then(cleanup)
+        .then(tableNames => {
+          if (tableNames[0] && tableNames[0].tableName) {
+            tableNames = tableNames.map(v => v.tableName);
+          }
+          expect(tableNames).to.have.length(1);
+          expect(tableNames).to.contain('my_test_table');
+          expect(tableNames).to.not.contain('V_Fail');
+        });
+    });
+  });
+
   describe('renameTable', () => {
     it('should rename table', function() {
       return this.queryInterface
--- a/test/integration/query-interface.test.js
+++ b/test/integration/query-interface.test.js
@@ -42,23 +42,21 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
   });
 
   describe('showAllTables', () => {
-    it('should contain only tables', function() {
-      const cleanup = r => {
-        return this.sequelize.query('DROP VIEW V_Fail').then(() => r, () => r);
+    it('should not contain views', function() {
+      const cleanup = () => {
+        return this.sequelize.query('DROP VIEW V_Fail').reflect();
       };
       return this.queryInterface
         .createTable('my_test_table', { name: DataTypes.STRING })
-        .then(cleanup)
+        .tap(cleanup)
         .then(() => this.sequelize.query('CREATE VIEW V_Fail AS SELECT 1 Id'))
         .then(() => this.queryInterface.showAllTables())
-        .then(cleanup)
+        .tap(cleanup)
         .then(tableNames => {
           if (tableNames[0] && tableNames[0].tableName) {
             tableNames = tableNames.map(v => v.tableName);
           }
-          expect(tableNames).to.have.length(1);
-          expect(tableNames).to.contain('my_test_table');
-          expect(tableNames).to.not.contain('V_Fail');
+          expect(tableNames).to.deep.equal(['my_test_table']);
         });
     });
   });
--- a/lib/dialects/mssql/query-generator.js
+++ b/lib/dialects/mssql/query-generator.js
@@ -222,7 +222,7 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
   }
 
   showTablesQuery() {
-    return 'SELECT TABLE_NAME, TABLE_SCHEMA FROM INFORMATION_SCHEMA.TABLES;';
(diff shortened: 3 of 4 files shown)


papb:

I will now review your PR.

test/integration/query-interface.test.js
```diff
@@ -41,6 +41,28 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
     });
   });
 
+  describe('showAllTables', () => {
+    it('should contain only tables', function() {
```
papb: Please change this to `should not contain views`, I think it is more descriptive

```suggestion
    it('should not contain views', function() {
```

test/integration/query-interface.test.js
```diff
@@ -41,6 +41,28 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+      const cleanup = r => {
+        return this.sequelize.query('DROP VIEW V_Fail').then(() => r, () => r);
+      };
```
papb: Wouldn't using `DROP VIEW IF EXISTS` simplify this catch-and-ignore pattern here? If yes, please change this to:

```js
const cleanup = () => this.sequelize.query('DROP VIEW IF EXISTS V_Fail');
```

If not, please tell me why and regardless refactor this to:

```js
const cleanup = () => this.sequelize.query('DROP VIEW V_Fail').reflect();
```

Which is more readable.

test/integration/query-interface.test.js
```diff
@@ -41,6 +41,28 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+      return this.queryInterface
+        .createTable('my_test_table', { name: DataTypes.STRING })
+        .then(cleanup)
```
papb: Change this to `.tap(cleanup)` to improve readability (and make my suggestions above work!)

test/integration/query-interface.test.js
```diff
@@ -41,6 +41,28 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+        .then(() => this.sequelize.query('CREATE VIEW V_Fail AS SELECT 1 Id'))
+        .then(() => this.queryInterface.showAllTables())
+        .then(cleanup)
```
papb: Also change this to `.tap(cleanup)`

test/integration/query-interface.test.js
```diff
@@ -41,6 +41,28 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+        .then(tableNames => {
+          if (tableNames[0] && tableNames[0].tableName) {
+            tableNames = tableNames.map(v => v.tableName);
+          }
+          expect(tableNames).to.have.length(1);
+          expect(tableNames).to.contain('my_test_table');
+          expect(tableNames).to.not.contain('V_Fail');
```
papb: Since you are asserting that the array has length 1 and asserting that it contains `'my_test_table'`, automatically it won't contain `'V_Fail`, right? I think you should refactor these lines to a simpler assert:

```js
expect(tableNames).to.deep.equal(['my_test_table']);
```

papb:
Also, by the way, don't worry about the code coverage.

test/integration/query-interface.test.js
joaoe: The `IF EXISTS` only became supported in MS SQL 2016. So I'm a bit afraid that adding `IF EXISTS` would break tests in some old databases. Unless the CI pipeline has the oldest versions of each RDBMS that is officially supported.

joaoe:
Hello. I fixed everything, but had to push a rebased branch because the first one had the wrong committer info.
Thanks.

test/integration/query-interface.test.js
papb: Makes sense. Please use my second suggestion then, and add a comment above it explaining what you just said.

test/integration/query-interface.test.js
```diff
@@ -41,6 +41,26 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+        .createTable('my_test_table', { name: DataTypes.STRING })
+        .tap(cleanup)
+        .then(() => this.sequelize.query('CREATE VIEW V_Fail AS SELECT 1 Id'))
+        .then(() => this.queryInterface.showAllTables())
+        .tap(cleanup)
+        .then(tableNames => {
+          if (tableNames[0] && tableNames[0].tableName) {
+            tableNames = tableNames.map(v => v.tableName);
+          }
+          expect(tableNames).to.deep.equal(['my_test_table']);
```
papb: Actually, I take back what I said before for this part of the code. The fact that you named the variable `tableNames` in the `.then` confused me. Here is the new request:

* Rename the variable from `tableNames` to `tables`, since it is not an array of strings, as the name would suggest
* Remove the if
* Add the following expects

```js
expect(tables).to.be.an('array');
expect(tables).to.have.length(1);
expect(tables[0].tableName).to.equal('my_test_table');
```

joaoe: > Remove the if

I can't remove the if. The `showAllTables()` returns either an array of strings of an array of object with name+schema in different dialects. I do think the API should be normalized to always return names+schemas but that's not in the scope of this task.

papb: Ah! I didn't know that, sorry. Thanks for the explanation. I thought objects were always returned. Do you know which dialect(s) return only names?

joaoe: Just checked the code in master. mysql, postgres and sqlit return only table names without schemas. If you check the test for `renameTable` you'll see the same if.
Changing the behavior of showAllTables() should be considered a change that breaks the API.

joaoe: Just checked the code in master. mysql, postgres and sqlit return only table names without schemas. If you check the test for renameTable you'll see the same if.
Changing the behavior of showAllTables() should be considered a change that breaks the API.

papb: @joaoe Thanks! Yeah, breaks API, so it will have to wait for the next major version. I have opened a new issue to track this (#11451 11451).

Your PR is good as is! Thanks!

lib/dialects/mysql/query-generator.js
```diff
@@ -149,7 +149,7 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
   }
 
   showTablesQuery() {
-    return 'SHOW TABLES;';
+    return "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = 'BASE TABLE';";
```
UziTech: This introduced another bug. It shows tables for all schemas not just the current schema.

joaoe: Well, not a bug since it is how the API works in all other dialects. But an API breaking change then.

joaoe: Btw, you mean the default schema for that user ?

UziTech: in mysql databases are schemas so when you select a database in sequelize you are actually selecting a schema but this query returns tables for all databases/schemas not just the selected database/schema.

UziTech: The correct query would probably be `show full tables where Table_Type = 'BASE TABLE'`

joaoe: How, so `SHOW TABLES` actually works like a SELECT statement ? I do not have a mysql db locally to test so I relied on the CI jobs. You can make a small PR just to revert that change. 
But when #11451 it taken care of, showAllTables will be normalized across all dialects and that means that all schemas are returned. But that will be done during a major release so API breakage is expected.

papb: Hmmm I see now that it's a bit more complicated than I thought, since as UziTech says in MySQL databases are schemas, hmmm, this brings some complications... As @joaoe said, please send a PR if you can, @UziTech, and thanks for the report!

UziTech: I added #11456 to try to fix this without just reverting it.

PR #11454, opened 2019-09-19
fix: sequelize.where col not null (#11447)

### LINKED ISSUE #11447: Sequelize.where col not null
## Issue Description

Can't query col is not NULL with Sequelize.where function.

### What are you doing?
Write a select query with the following where state.

```
where: [{}, 
    sequelize.where(sequelize.col('id'), Op.ne, null),
    {'id': {
        [Op.ne]: null
    }}]
```

The first one is for not getting `literal replacements` error but it's a different topic.
Second and the third object should generate the same `id IS NOT NULL`.

The second (`sequelize.where(sequelize.col('id'), Op.ne, null)`) generate `id IS NULL`.

```
id IS NULL AND id IS NOT NULL
```

### Environment

- Sequelize version: 5.18.4 
- Node.js version: v12.0.0
- Operating System: OSX -> Docker node:12.0 container
- MySql version: 5.7.23

- [x] I think this problem happens only for the following dialect(s): MySQL

### LINKED ISSUE #11452: Error: Support for literal replacements in the `where` object has been removed.
### What are you doing?
Just copy and past it to the `getWhereConditions` unit test in `test/unit/sql/where.test.js` file and run `test-unit-mysql`:

```js
    testsql([current.where(current.col('name'), null), current.where(current.fn('SUM', current.col('hours')), '>', 0)], {
      default: '([name] IS NULL AND SUM([hours]) > 0)'
    });
```

Correct me if I'm wrong but it should pass.

Get an error:
`Error: Support for literal replacements in the `where` object has been removed.`

```
Output here
```

### Additional context

If you put an empty object to the where array, then it will pass and generate the correct output.
```js
   testsql([{}, current.where(current.col('name'), null), current.where(current.fn('SUM', current.col('hours')), '>', 0)], {
      default: '([name] IS NULL AND SUM([hours]) > 0)'
    });
```

### Environment

 - Sequelize version: 5.18.4
 - Node.js version: v12.0.0
 - Operating System: OSX -> Docker node:12.0 container
 - MySql version: 5.7.23

- [x] I think this problem happens only for the following dialect(s): MySQL

- [x] No, I don't have the time, although I believe I could do it if I had the time... but I'll try to fix when I'll have time

dodderjs:
### Pull Request check-list

### Description of change

Closes #11447
Closes #11452


lib/dialects/abstract/query-generator.js
```diff
@@ -2070,7 +2070,16 @@ class QueryGenerator {
       if (value && value instanceof Utils.SequelizeMethod) {
         value = this.getWhereConditions(value, tableName, factory, options, prepend);
 
-        return value === 'NULL' ? `${key} IS NULL` : [key, value].join(` ${smth.comparator} `);
+        if (value === 'NULL') {
```
sushantdhiman: Coverage is down for these lines, Can you add additional unit test for this case?

dodderjs: I just added some to test query-generator.test.js and some for utils.test.js.

Issue #11451, opened 2019-09-19
Make public, document and standardize the API of the `queryInterface.showAllTables` method


papb:
Currently the `queryInterface.showAllTables` method is marked as `@private` and therefore does not appear in the auto-generated docs. The jsdoc placed on it deserves improvement as well.

https://github.com/sequelize/sequelize/blob/d5e4a10f4e8dc1e37fe46678f04bbfa66b1299cc/lib/query-interface.js#L407-L421

Its API is also inconsistent across dialects:
* Some dialects return an array of strings (table names)
* Some dialects return an array of objects of the form `{ tableName, schema }`

I think this should be changed to always return `{ tableName, schema? }`. Of course this would be a breaking change though.

joaoe:
All RDBMSs support schemas. Sqlite does not. So for sqlite perhaps: a) keep old behavior to break less existing code using sqlite databases currently b) return `{schema:null}` to normalize the API. I don't have any strong opinions.

joaoe:
Feature request: extend `showAllTables()` to accept a `schemaName` as argument and list only tables for that schema.

papb:
@joaoe Thanks!! I prefer something like option `b`, but instead of returning `{tableName:...,schema:null}` I think simply returning `{tableName: 'abc'}` is better (no schema at all)

misos1:
Maybe something similar should be also done for showAllSchemas. It is public in documentation where it returns just Promise\<Array\> but content differ per dialect.

mysql (lists only actual database with tables):
``` js
[
  { Tables_in_database1: 'table_1' },
  { Tables_in_database1: 'table_2' }
]
```

mariadb (lists all databases):
``` js
[ 'database1', 'database2' ]
```

papb:
@misos1 Indeed! Thanks. Can you open a new issue to track this? And link to this one? Thanks

nodejh:
@papb Hi, what properties does the `schema` have? 

Sometimes I want to get some information (e.g.:  `charset` `comment`) passed to `queryInterface.createTable` when I use `queryInterface.showAllTables` method. 

```js
  /**
   * Create a table with given set of attributes
   *
   * ```js
   * queryInterface.createTable(
   *   'nameOfTheNewTable',
   *   {
   *     // ...
   *   },
   *   {
   *     engine: 'MYISAM',    // default: 'InnoDB'
   *     charset: 'latin1',   // default: null
   *     schema: 'public',    // default: public, PostgreSQL only.
[...]

PR #10192, opened 2018-11-22
feat(mariadb): implementation for MariaDB

rusher:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Have you added new tests to prevent regressions?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?

### Description of change

Implementation of MariaDB. 
This differs from MySQL mostly for 3 things:
- the driver is different ([mariadb](https://github.com/MariaDB/mariadb-connector-nodejs)
- use of schema
- MariaDB JSON implementation differ from MySQL

Breaking change is the use of schemas. MySQL/MariaDB have schemas, but no "default" public schema. 
Current implementation for MySQL will create a table with name `mySchema_myTable`, using schema, that is `mySchema`.`myTable`.
I haven't corrected that for MySQL since it will break every existing application, but for new MariaDB implementation, it seems appropriate to use a better base with real schemas. 

Tests have been updated accordingly (defaut method that drop all schemas is replace to keep default schema that is needed for MariaDB)

PR #11456, opened 2019-09-19
Fix mysql showAllTables listing all tables for all databases

### LINKED ISSUE #10629: mariadb: showAllTables lists tables from other databases
## What are you doing?
With mariadb, `showAllTables` lists all tables on the server (except system tables), rather than only those tables in the current database (which seems to be the behavior with other dialects, in particular mysql and postgres).

```js
  const sequelize = new Sequelize(config.dbUrl);
  console.log(await sequelize.getQueryInterface().showAllTables());
```

`[]`

Tables from other databases on the server.

Notably, this means that `dropAllTables` will attempt to clear out other databases, too.

The query seems to be using `information_schema.tables`, but (unlike Postgres) information_schema is global in MySQL/MariaDB. (Switching dialect to mysql, which uses `SHOW TABLES`, works.)

__Dialect:__ mariadb
__Dialect version:__ `mariadb@2.0.3`
__Database version:__ 10.3
__Sequelize version:__ master

UziTech:
### Pull Request check-list

### Description of change

Fixes https://github.com/sequelize/sequelize/pull/11439#pullrequestreview-290558915
Fixes #10629

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,6 +56,8 @@ services:
       MYSQL_DATABASE: sequelize_test
       MYSQL_USER: sequelize_test
       MYSQL_PASSWORD: sequelize_test
+    volumes:
+    - $MYSQLDB_ENTRYPOINT:/docker-entrypoint-initdb.d
     ports:
       - "8980:3306"
     container_name: mysql-57
@@ -69,4 +71,3 @@ services:
     ports:
       - "8970:1433"
     container_name: mssql
-
--- a/lib/dialects/mariadb/query-generator.js
+++ b/lib/dialects/mariadb/query-generator.js
@@ -27,7 +27,7 @@ class MariaDBQueryGenerator extends MySQLQueryGenerator {
   showTablesQuery(database) {
     let query = 'SELECT TABLE_NAME, TABLE_SCHEMA FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = \'BASE TABLE\'';
     if (database) {
-      query += ` AND TABLE_SCHEMA = '${database}'`;
+      query += ` AND TABLE_SCHEMA = ${this.escape(database)}`;
     } else {
       query += ' AND TABLE_SCHEMA NOT IN (\'MYSQL\', \'INFORMATION_SCHEMA\', \'PERFORMANCE_SCHEMA\')';
     }
--- a/lib/dialects/mariadb/query-generator.js
+++ b/lib/dialects/mariadb/query-generator.js
@@ -24,8 +24,14 @@ class MariaDBQueryGenerator extends MySQLQueryGenerator {
     return `SELECT SCHEMA_NAME as schema_name FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME NOT IN ('MYSQL', 'INFORMATION_SCHEMA', 'PERFORMANCE_SCHEMA'${skip ? skip.reduce( (sql, schemaName) => sql += `,${this.escape(schemaName)}`, '') : ''});`;
   }
 
-  showTablesQuery() {
-    return 'SELECT TABLE_NAME, TABLE_SCHEMA FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA NOT IN (\'MYSQL\', \'INFORMATION_SCHEMA\', \'PERFORMANCE_SCHEMA\') AND TABLE_TYPE = \'BASE TABLE\'';
+  showTablesQuery(database) {
+    let query = 'SELECT TABLE_NAME, TABLE_SCHEMA FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = \'BASE TABLE\'';
+    if (database) {
+      query += ` AND TABLE_SCHEMA = '${database}'`;
+    } else {
+      query += ' AND TABLE_SCHEMA NOT IN (\'MYSQL\', \'INFORMATION_SCHEMA\', \'PERFORMANCE_SCHEMA\')';
+    }
+    return `${query};`;
   }
 }
 
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -148,8 +148,14 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
     return `SHOW FULL COLUMNS FROM ${table};`;
   }
 
-  showTablesQuery() {
-    return "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = 'BASE TABLE';";
+  showTablesQuery(database) {
+    let query = 'SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = \'BASE TABLE\'';
+    if (database) {
+      query += ` AND TABLE_SCHEMA = '${database}'`;
+    } else {
+      query += ' AND TABLE_SCHEMA NOT IN (\'MYSQL\', \'INFORMATION_SCHEMA\', \'PERFORMANCE_SCHEMA\', \'SYS\')';
+    }
+    return `${query};`;
   }
 
   addColumnQuery(table, key, dataType) {
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -151,7 +151,7 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
   showTablesQuery(database) {
     let query = 'SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = \'BASE TABLE\'';
     if (database) {
-      query += ` AND TABLE_SCHEMA = '${database}'`;
+      query += ` AND TABLE_SCHEMA = ${this.escape(database)}`;
     } else {
       query += ' AND TABLE_SCHEMA NOT IN (\'MYSQL\', \'INFORMATION_SCHEMA\', \'PERFORMANCE_SCHEMA\', \'SYS\')';
     }
--- a/test/config/mysql/mysql.sql
+++ b/test/config/mysql/mysql.sql
@@ -1,0 +1,3 @@
+GRANT ALL ON *.* TO 'sequelize_test'@'%' with grant option;
+
+FLUSH PRIVILEGES;
--- a/test/integration/query-interface.test.js
+++ b/test/integration/query-interface.test.js
@@ -63,6 +63,25 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
           expect(tableNames).to.deep.equal(['my_test_table']);
         });
     });
+
+    if (dialect !== 'sqlite' && dialect !== 'postgres') {
+      // NOTE: sqlite doesn't allow querying between databases and
+      // postgres requires creating a new connection to create a new table.
+      it('should not show tables in other databases', function() {
+        return this.queryInterface
+          .createTable('my_test_table1', { name: DataTypes.STRING })
+          .then(() => this.sequelize.query('CREATE DATABASE my_test_db'))
+          .then(() => this.sequelize.query(`CREATE TABLE my_test_db${dialect === 'mssql' ? '.dbo' : ''}.my_test_table2 (id INT)`))
+          .then(() => this.queryInterface.showAllTables())
+          .tap(() => this.sequelize.query('DROP DATABASE my_test_db'))
+          .then(tableNames => {
+            if (tableNames[0] && tableNames[0].tableName) {
+              tableNames = tableNames.map(v => v.tableName);
+            }
+            expect(tableNames).to.deep.equal(['my_test_table1']);
+          });
+      });
+    }
   });
 
   describe('renameTable', () => {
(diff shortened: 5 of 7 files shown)


test/config/mysql/mysql.sql
```diff
@@ -0,0 +1,3 @@
+GRANT ALL ON *.* TO 'sequelize_test'@'%' with grant option;
```
UziTech: MySql is failing because the user `sequalize_test` doesn't have permissions to use the created database.

I noticed that MariaDB has a file like this to grant it privileges but I don't know how to do the same in mysql.

@papb any ideas?

papb: Hmm... Note that [this line in .travis.yml](https://github.com/sequelize/sequelize/blob/8b53f721a07e4ea5a42d74199b319f813f2c109c/.travis.yml#L31) tells MariaDB that it should use that file. I am not sure if something similar exists for MySQL, but if it does, you'll have to update the .travis.yml to do something similar.

Another options that comes to my mind is trying the following at the beginning of your test:

```js
sequelize.query(`GRANT ALL ON *.* TO 'sequelize_test'@'%' with grant option;`)
```

UziTech: I tried your second option but it just threw an error saying `sequalize_test` didn't have permission lol

papb: Hmm... Good luck :P

lib/dialects/mariadb/query-generator.js
```diff
@@ -24,8 +24,12 @@ class MariaDBQueryGenerator extends MySQLQueryGenerator {
     return `SELECT SCHEMA_NAME as schema_name FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME NOT IN ('MYSQL', 'INFORMATION_SCHEMA', 'PERFORMANCE_SCHEMA'${skip ? skip.reduce( (sql, schemaName) => sql +=  `,${this.escape(schemaName)}`, '') : ''});`;
   }
 
-  showTablesQuery() {
-    return 'SELECT TABLE_NAME, TABLE_SCHEMA FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA NOT IN (\'MYSQL\', \'INFORMATION_SCHEMA\', \'PERFORMANCE_SCHEMA\') AND TABLE_TYPE = \'BASE TABLE\'';
+  showTablesQuery(database) {
+    let query = 'SELECT TABLE_NAME, TABLE_SCHEMA FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA NOT IN (\'MYSQL\', \'INFORMATION_SCHEMA\', \'PERFORMANCE_SCHEMA\') AND TABLE_TYPE = \'BASE TABLE\'';
```
UziTech: I would like to get feedback from @rusher since he created the PR for MariaDB (#10192).

I have a couple questions about your initial PR:

1. Was it intentional to return all tables in all databases?
2. If MariaDB has schemas separate from databases why is `TABLE_SCHEMA` the same as the database in `INFORMATION_SCHEMA.TABLES`?

UziTech:
I'm not sure why MariaDB is failing on travis but it seems unrelated to this PR. @papb can you rerun that job on travis?

papb:
@UziTech We have been experiencing these flaky tests in Sequelize for quite a while now, they always have the form of `Error: Expected 0 running queries. 1 queries still running`. Don't worry about them, me and sushantdhiman know that we can ignore those for now, so your PR is green for our eyes :D

Also whenever you want to trigger a rerun you can just:

```
git commit --allow-empty --no-verify -m "rerun CI" && git push
```

(Side note: I don't have permissions in Travis to trigger a rerun...)

docker-compose.yml
```diff
@@ -56,6 +56,8 @@ services:
       MYSQL_DATABASE: sequelize_test
       MYSQL_USER: sequelize_test
       MYSQL_PASSWORD: sequelize_test
+    volumes:
+    - $MYSQLDB_ENTRYPOINT:/docker-entrypoint-initdb.d
```
papb: I see you figured it out, excellent!!

UziTech:
> `git commit --allow-empty --no-verify -m "rerun CI" && git push`

Thanks that's a neat hack.

I wonder if the flaky tests are related to https://github.com/sequelize/sequelize-pool/pull/12. I originally found that issue when I noticed tests that relied on sequalize made jest linger after tests were completed.

test/integration/query-interface.test.js
```diff
@@ -63,6 +63,23 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
           expect(tableNames).to.deep.equal(['my_test_table']);
         });
     });
+
+    if (dialect !== 'sqlite' && dialect !== 'postgres') {
```
joaoe: Postgre supports `CREATE DATABASE`.
The odd one out is SQLite. For that one you can use `ATTACH DATABASE 'blah.db' as my_db` although the test is somewhat unnecessary because the `sqlite_master` view is limited to each database.

lib/dialects/mysql/query-generator.js
```diff
@@ -148,8 +148,12 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
     return `SHOW FULL COLUMNS FROM ${table};`;
   }
 
-  showTablesQuery() {
-    return "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = 'BASE TABLE';";
+  showTablesQuery(database) {
+    let query = 'SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA NOT IN (\'MYSQL\', \'INFORMATION_SCHEMA\', \'PERFORMANCE_SCHEMA\') AND TABLE_TYPE = \'BASE TABLE\'';
```
joaoe: What if I want to query the `MYSQL` schema ?
Perhaps this check would be a bit better
```
if (database) {
  query += ' AND TABLE_SCHEMA = `${database}`'
} else {
  query += ' AND TABLE_SCHEMA NOT IN ...'
}
```

lib/dialects/mariadb/query-generator.js
joaoe: https://dev.mysql.com/doc/refman/8.0/en/create-database.html
`CREATE SCHEMA is a synonym for CREATE DATABASE.`

test/integration/query-interface.test.js
UziTech: Postgres doesn't allow creating a table in another database

```sql
CREATE TABLE my_test_db.my_test_table2 (id INT)
```

joaoe: Googling the issue shows that one needs a new connection to the other DB. Seems a bit overkill so not worthwhile testing for that. But at least you should write a comment why these dialects are being excluded.

lib/dialects/mariadb/query-generator.js
UziTech: It seems to be the same for MariaDB

https://mariadb.com/kb/en/library/create-database/

I've never used MariaDB and I'm starting to wonder if it actually supports schemas the same way PostgreSQL and MSSQL do.

UziTech:
This PR seems ready but I would still like to get feedback from @rusher on (#10192).

I have a couple questions about his initial PR:

1. Was it intentional to return all tables in all databases?
2. Does MariaDB actually have support for schemas like PostgreSQL and MSSQL?
   `CREATE SCHEMA` seems to just be a synonym of `CREATE DATABASE`
   https://mariadb.com/kb/en/library/create-database

test/config/mysql/mysql.sql
```diff
@@ -0,0 +1,3 @@
...
+
+FLUSH PRIVILEGES;
```
UziTech: I'm not sure if it would be better to add this file or just use the one for MariaDB since they are exactly the same.

papb: I think it's better to keep separate files.

rusher:
@UziTech 

See the same discussion here : https://github.com/sequelize/sequelize/issues/10629

>Does MariaDB actually have support for schemas like PostgreSQL and MSSQL?
>CREATE SCHEMA seems to just be a synonym of CREATE DATABASE
Not really. Actually. MariaDB schema is synonym of database.
MariaDB doesn't have databases but schemas. In fact, to try to be clear, MariaDB handle like there is only one database named 'def'.

See extract from information_schema.TABLES: 
image

Having multiple databases might come in a very near future, mostly because of compatibility. 

When creating a table with a schema, it will be rendered like a mariadb user would expect it, in the designated schema.  
Using Mysql implementation will create a table named 'schema_table'.

UziTech:
Ok so if I'm getting this correct in MariaDB a "Database" is the same as a "Schema" like MySQL but MariaDB **might** support "Catalogs" in the future (right now it is always "def") which is like a "Database" in MSSQL and PostgreSQL.

I've never used MariaDB but if I ever did I would assume a "Database" is a "Schema" not a "Catalog". Therefore to make `showAllTables` intuitive `returns the name of all existing tables in the database.` means it returns all tables in the "Schema" not the "Catalog"

UziTech:
@papb I think this PR is ready to be merged.

test/integration/query-interface.test.js
```diff
@@ -63,6 +63,45 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+      // NOTE: sqlite doesn't allow querying between databases and
+      // postgres requires creating a new connection to create a new table.
+      it('should not show tables in other databases', function() {
+        return this.queryInterface
+          .createTable('my_test_table1', { name: DataTypes.STRING })
+          .then(() => this.sequelize.query('CREATE DATABASE my_test_db'))
+          .then(() => this.sequelize.query(`CREATE TABLE my_test_db${dialect === 'mssql' ? '.dbo' : ''}.my_test_table2 (id INT)`))
```
rusher: having a test `dialect === 'mssql'`  inside a test `if (dialect !== 'sqlite' && dialect !== 'postgres') {` ?

UziTech: because `mssql` requires listing the schema (`dbo`) when creating the table but `mysql` and `mariadb` don't.

test/integration/query-interface.test.js
```diff
@@ -63,6 +63,45 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+            expect(tableNames).to.deep.equal(['my_test_table1']);
+          });
+      });
+
+      if (dialect === 'mysql' || dialect === 'mariadb') {
+        it('should show all tables in all databases', function() {
+          return this.queryInterface
+            .createTable('my_test_table1', { name: DataTypes.STRING })
+            .then(() => this.sequelize.query('CREATE DATABASE my_test_db'))
+            .then(() => this.sequelize.query(`CREATE TABLE my_test_db${dialect === 'mssql' ? '.dbo' : ''}.my_test_table2 (id INT)`))
```
rusher: having a test dialect === 'mssql' inside a test if (dialect !== 'sqlite' && dialect !== 'postgres') { ? (sorry for previous comment, place was here :p )

UziTech: good catch

UziTech:
Looks like all tests are passing but travis hasn't reported back. I rebased as well so the commits actually make sense for semantic-release.

lib/dialects/mariadb/query-generator.js
```diff
@@ -24,8 +24,14 @@ class MariaDBQueryGenerator extends MySQLQueryGenerator {
     return `SELECT SCHEMA_NAME as schema_name FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME NOT IN ('MYSQL', 'INFORMATION_SCHEMA', 'PERFORMANCE_SCHEMA'${skip ? skip.reduce( (sql, schemaName) => sql += `,${this.escape(schemaName)}`, '') : ''});`;
   }
 
-  showTablesQuery() {
-    return 'SELECT TABLE_NAME, TABLE_SCHEMA FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA NOT IN (\'MYSQL\', \'INFORMATION_SCHEMA\', \'PERFORMANCE_SCHEMA\') AND TABLE_TYPE = \'BASE TABLE\'';
+  showTablesQuery(database) {
+    let query = 'SELECT TABLE_NAME, TABLE_SCHEMA FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = \'BASE TABLE\'';
+    if (database) {
+      query += ` AND TABLE_SCHEMA = '${database}'`;
```
sushantdhiman: `TABLE_SCHEMA = ${this.escape(database)}`

lib/dialects/mysql/query-generator.js
```diff
@@ -148,8 +148,14 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
     return `SHOW FULL COLUMNS FROM ${table};`;
   }
 
-  showTablesQuery() {
-    return "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = 'BASE TABLE';";
+  showTablesQuery(database) {
+    let query = 'SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = \'BASE TABLE\'';
+    if (database) {
+      query += ` AND TABLE_SCHEMA = '${database}'`;
```
sushantdhiman: ```TABLE_SCHEMA = ${this.escape(database)}```

misos1:
Seems showAllTables is missing in documentation.

https://sequelize.org/master/class/lib/query-interface.js~QueryInterface.html

papb:
@misos1 Correct, see #11451

PR #11457, opened 2019-09-19
Allow HasManyCreateAssociationMixin Values to be Configurable

todd:
### Pull Request check-list

### Description of change

The Typescript signature for `values` in `HasManyCreateAssociationMixin` doesn't allow for great type safety and assumes that your values have an index signature instead of being explicitly typed. This change adds support for configuring `values` with a new `TValues` generic that defaults to the current index signature type.

One could also make the argument that the default type here should be `Partial<TModel>`, but we should discuss. Whichever route we decide on, we should also implement for the other association create mixins.


papb:

Can you please show an example of usage that is not possible with the current typings and will become possible after your PR is merged? Thanks!

todd:
@papb Yes, of course!

Consider a model such as:

```typescript
class ModelA extends Model {
  public model_b_id!: number;
  public column_one!: string;
  public column_two!: number;
}
```

And another model:

```typescript
class ModelB extends Model {
  public createModelA!: HasManyCreateAssociationMixin<ModelA>;
}
```

Because of the loose `values` types in `HasManyCreateAssociation`, it's possible to write code where invalid attributes are passed to the mixin method without TypeScript raising a fuss.

```typescript
const instance = await ModelB.findOne();

// This code will not result in a compilation failure despite two glaring
// problems with it:
//   1) column_three does not exist on ModelA
//   2) column_two values should be numbers, but we've provided a string here
await instance.createModelA({
  column_one: 'foo',
  column_two: 'bar',
  column_three: 'bar'
});
```

Optionally, users should be able to override the default `values` types to get better type safety here. That's what the PR I've opened supports. However, an argument could be made that the `TValues` type I've introduced should really default to `Partial<TModel>` since we don't really want to support anything that's not defined on the model.

Let me know if you need more clarification.

todd:
@papb Bump

todd:
@papb What's the deal here? Some feedback would be great.

papb:
Hi @todd, sorry for taking long. I've been very busy last month, and notifications were piling up, since I didn't have time for everything I decided to prioritize the oldest notifications :sweat_smile: since you bumped this PR, it went up (github overrides the notifications from the same place) so I kept prioritizing other posts :sweat_smile: bad luck. Sorry about that.

I will reply faster now, thanks for the examples, can you please expand on your example showing how to use your new functionality so that the code no longer compiles (as it shouldn't)?

todd:
@papb Sorry, now it's my turn to apologize for letting this languish 😅

To use what I'm proposing, you'd add an additional type to the mixin invocation that would whitelist the allowed attributes:

```typescript
interface MyValues {
  column_one: string;
  column_two: number;
}

class ModelB extends Model {
  public createModelA!: HasManyCreateAssociationMixin<ModelA, Partial<MyValues>>;
}
```

Using this would result in [my example](https://github.com/sequelize/sequelize/pull/11457#issuecomment-537325579) breaking the build.

As I expanded upon in that missive, it might be appropriate to set the default for `TValues` to `Partial<TModel>` - this would, by default, prevent assignment of attributes not explicitly defined on `TModel`. The downside of this approach is that additional members of `TModel` would be pulled in for that (think instance methods and the like) that you wouldn't want exposed in a creation method.

Having been using the built-in TypeScript defs for a couple of months now, it appears that this type-safety issue isn't just limited to these mixin methods, but essentially every create/update type method that takes an object of attributes. Perhaps it's worth adding an optional generic type on `Model` that would allow end-users to optionally specify what the global attributes that can be used in those methods is, similar to what I've proposed with the mixin method. Something along the lines of:

```typescript
interface ModelAAttributes {
  column_one: string;
  column_two: number;
}

class ModelA extends Model<ModelAAttributes> {
  public model_b_id!: number;
  public column_one!: string;
  public column_two!: number;
}
```

We could then modify the types to lock down create/update methods to only allow attributes whitelisted by that type, while defaulting to the existing scheme (`{ [attribute: string]: unknown }`) so as to not break things for existing users or folks who don't care about or need the added type-safety this would provide. But that's likely a much larger discussion.

papb:
Hi @todd, thanks!



> it appears that this type-safety issue isn't just limited to these mixin methods, but essentially every create/update type method that takes an object of attributes

Indeed... I think the PR #11820, by the creator of [sequelize-typescript](https://github.com/RobinBuschmann/sequelize-typescript), is addressing this. Can you take a look on what he is doing there and how this relates to your ideas? Let's not make duplicate work :)

Also clearly the two of you know more typescript than I do, so it would be great if you two could join forces :) Let me know, thanks!

todd:
@papb Ah, yes. I was unaware of that PR. While it doesn't appear to cover the work I've done here, I believe that's easily remedied with the work that's being done there. I'll close this PR and jump on over to that PR.

Issue #10975, opened 2019-05-20
Sequelize does not properly use mysql sessions


misos1:
## What are you doing?

When I start transaction with `isolationLevel: Sequelize.Transaction.ISOLATION_LEVELS.READ_UNCOMMITTED` :

1. It runs `SET SESSION TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;` after `START TRANSACTION;` so it does not affect the current ongoing transaction as is stated in mysql manual https://dev.mysql.com/doc/refman/8.0/en/set-transaction.html#set-transaction-isolation-level

2. But it affects another transactions that runs on same (reused) connection from pool.

When I start new transaction it always appears in logs with new "session" identifier like `9101e35d-e571-491b-8fc3-c03c6fed1811`. But this is different from mysql session which is defined by sequelize connection and sequelize reuses connections in pool. So when I (or sequelize) change session isolation level for some transaction with sequelize session like `9101e35d-e571-491b-8fc3-c03c6fed1811` it can affect future transactions in different sequelize sessions like `494a41f6-4fe9-4710-9c52-ab016d4d6378` which unhappily used same connection from sequelize connection pool. Once is set session transaction isolation level it will be here permanently for particular sequelize connection not only for current sequelize "session" (or transaction) because sequelize uses connection pool and reuses connections for its "sessions".

```node.js
await sequelize.transaction(async function(transaction)
{
	// Executing (9101e35d-e571-491b-8fc3-c03c6fed1811): START TRANSACTION;
	console.log((await sequelize.query("SELECT @@TX_ISOLATION;", { type: sequelize.QueryTypes.SELECT, transaction }))[0]["@@TX_ISOLATION"]);
	// Executing (9101e35d-e571-491b-8fc3-c03c6fed1811): SELECT @@TX_ISOLATION;
	// REPEATABLE-READ
	await sequelize.query("SET SESSION TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;", { transaction });
	// Executing (9101e35d-e571-491b-8fc3-c03c6fed1811): SET SESSION TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;
	console.log((await sequelize.query("SELECT @@TX_ISOLATION;", { type: sequelize.QueryTypes.SELECT, transaction }))[0]["@@TX_ISOLATION"]);
	// Executing (9101e35d-e571-491b-8fc3-c03c6fed1811): SELECT @@TX_ISOLATION;
[...]

PR #11475, opened 2019-09-25
Fix setting isolation level in MySQL and MariaDB

poldridge:
### Pull Request check-list

Fixes Issue #10975 : Setting isolation level does not work in MySQL or MariaDB. For both those dialects the setIsolationLevel query must be executed before the startTransaction query. For postgres the reverse is required, but it must be set before the first query in the transaction. For other dialects, it can be set at any time before or during a transaction.

This drops `SESSION` from the `SET TRANSACTION` queries, in order to set the isolation for a just a single transaction when future transactions in the session don't specify one, and updates MySQL and MariaDB to issue the `SET TRANSACTION` statement before the `START TRANSACTION`.

Three tests are added to confirm that changes isolation level works in all dialects. All added tests fail in mysql and mariadb without this fix.

### DIFF
--- a/lib/dialects/abstract/index.js
+++ b/lib/dialects/abstract/index.js
@@ -30,6 +30,7 @@ AbstractDialect.prototype.supports = {
   bulkDefault: false,
   schemas: false,
   transactions: true,
+  settingIsolationLevelDuringTransaction: true,
   transactionOptions: {
     type: false
   },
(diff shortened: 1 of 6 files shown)


poldridge:
I think I broke the code coverage report by force pushing so I am remaking this PR.

PR #11495, opened 2019-10-01
fix(mssql): Virtual column checking at mssql @tmp table at v4

weslly:
### Pull Request check-list

### Description of change

When using the `hasTrigger: true` option at mssql tables, sequelize creates a temporary table instead of using `OUTPUT INSERTED`, but this temp table includes the column names from relationships twice because `attribute.type` doesn't seem to have the expected `VIRTUAL` datatype. 

This fix uses the `references` attribute to check for relationship columns, since it is only set on those relationship columns that are already defined earlier without it.


sushantdhiman:
Sorry can't accept this PR, `v4 / v3` will only receive security patches or critical bug fixes

PR #11517, opened 2019-10-07
fix(createtable): pass table field name in attr

### LINKED ISSUE #11285: queryInterface.createTable() does not work with ARRAY(ENUM) with defaultValue
## What are you doing?

In postgres:

```js
await sequelize.queryInterface.createTable('test', {
    test: {
        type: Sequelize.ARRAY(Sequelize.ENUM(['foo', 'bar', 'baz', 'qux'])),
        allowNull: false,
        defaultValue: ['foo']
    }
});
```

**To Reproduce:** Just run the snippet above, no models or other setup needed
2. See error: `TypeError: Cannot read property 'getTableName' of undefined`

It should work

`TypeError: Cannot read property 'getTableName' of undefined`

Thrown at line:

https://github.com/sequelize/sequelize/blob/748f69080fdf86b8968e14c02e330b2a1cf41ebc/lib/dialects/postgres/data-types.js#L481

## Environment
Dialect:
- [x] postgres

Tested with latest release:
- [x] Yes, specify that version: 5.12.3

## Extra info

If the `defaultValue` is removed, it works:

```js
await sequelize.queryInterface.createTable('test', {
    test: {
        type: Sequelize.ARRAY(Sequelize.ENUM(['foo', 'bar', 'baz', 'qux'])),
        allowNull: false
    }
});
```

krazi3:
### Pull Request check-list

### Description of change
Closes #11285

Set `tableName` and `fieldName` on attribute with type ARRAY(ENUM)

### DIFF
--- a/lib/dialects/postgres/data-types.js
+++ b/lib/dialects/postgres/data-types.js
@@ -478,7 +478,7 @@ module.exports = BaseTypes => {
 
       if (this.type instanceof BaseTypes.ENUM) {
         castKey = `${Utils.addTicks(
-          Utils.generateEnumName(options.field.Model.getTableName(), options.field.fieldName),
+          Utils.generateEnumName(options.field.tableName || options.field.Model.getTableName(), options.field.fieldName),
           '"'
         ) }[]`;
       }
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -476,6 +476,12 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
 
         if (attribute.type instanceof DataTypes.ARRAY) {
           type += '[]';
+
+          // Add table and field name if not exists on attributes with ARRAY(ENUM) type
+          if (!attribute.fieldName) {
+            attribute.fieldName = options.key;
+            attribute.tableName = options.table.tableName || options.table;
+          }
         }
 
       } else {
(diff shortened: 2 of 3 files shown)


lib/dialects/postgres/query-generator.js
```diff
@@ -476,6 +476,12 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
 
         if (attribute.type instanceof DataTypes.ARRAY) {
           type += '[]';
+
+          // Add table and field name if not exists on attributes with ARRAY(ENUM) type
+          if (!attribute.fieldName) {
+            attribute.fieldName = options.key;
+            attribute.tableName = options.table.tableName || options.table;
```
sushantdhiman: `attribute.tableName` is not used anywhere else, I think.

lib/dialects/postgres/data-types.js
```diff
@@ -478,7 +478,7 @@ module.exports = BaseTypes => {
 
       if (this.type instanceof BaseTypes.ENUM) {
         castKey = `${Utils.addTicks(
-          Utils.generateEnumName(options.field.Model.getTableName(), options.field.fieldName),
+          Utils.generateEnumName(options.field.tableName || options.field.Model.getTableName(), options.field.fieldName),
```
sushantdhiman: Isn't `options.table.tableName` available here? It seems odd for attribute to keep `tableName`

krazi3: I don't think i get the `table` object in options ... i will check once again

lib/dialects/postgres/query-generator.js
krazi3: Yes this looked a bit weird to me as well ... but i had to pass table info at
https://github.com/sequelize/sequelize/pull/11517/files#diff-7e347c4738c97261969fbeb802a3c363R511

WikiRik:

Issue #5561, opened 2016-03-09
Defining a default ON UPDATE behavior for models, such as ON UPDATE CURRENT_TIMESTAMP


brennanMKE:
I want tables created with the definitions below but I do not see how I can do that with Sequelize when these columns are automatically defined. This is also being done with a MySQL database.

The reason I want it defined in the database is because multiple apps use this database and not all are built with Sequelize. And I want to enforce these default values. Is there a way to have Sequelize create tables with the schema shown below? Or would an update after the tables are created be necessary to older these columns?

```
`createdAt` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
`updatedAt` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```

I am looking through the docs and I do not see how these columns would be created as I intend.

brennanMKE:
I created a solution which is run after `sync` so that all tables are checked for the `createdAt` and `updatedAt` columns having missing defaults and altering those tables as needed. If the queries return 1 for the count it indicates the defaults are not set. Then the alter statements are used to set them.

Now other apps which use the same database without using Sequelize will benefit from these defaults.

I do not know if this kind of change would be accepted with the Sequelize code base. I found a couple of related issues in the past couple of years.

### Queries

```
SELECT COUNT(*) AS `count`
FROM `INFORMATION_SCHEMA`.`COLUMNS` 
WHERE `TABLE_SCHEMA`='experiment' 
AND `COLUMN_NAME` = 'createdAt' 
AND `COLUMN_DEFAULT` IS NULL
AND `TABLE_NAME`='users';

SELECT COUNT(*) AS `count`
FROM `INFORMATION_SCHEMA`.`COLUMNS` 
WHERE `TABLE_SCHEMA`='experiment' 
AND `COLUMN_NAME` = 'updatedAt' 
AND `COLUMN_DEFAULT` IS NULL
AND `EXTRA` = ''
AND `TABLE_NAME`='users';
```

### Alter Table

```
ALTER TABLE `users` MODIFY COLUMN `createdAt` DATETIME NOT NULL DEFAULT NOW();
ALTER TABLE `users` MODIFY COLUMN `updatedAt` DATETIME NOT NULL DEFAULT NOW() ON UPDATE NOW();
```

mickhansen:
If `defaultValue` does not currently support `sequelize.literal()` it should probably be updated to do so.

[...]

PR #11536, opened 2019-10-10
fix(upsert): createdAt config ignored

ptrwtts:
### Pull Request check-list

### Description of change

We are using milliseconds for our `createdAt` and `updatedAt` columns, by setting `defaultValue: Sequelize.fn('NOW',3)`. It works great, except when we `Upsert`, in which case only `updatedAt` has milliseconds.

This because when setting `createdAt`, it incorrectly checks for an existing value in the `updateValues` instead of the `insertValues`.

This PR changes that, and fixes the issue.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2465,7 +2465,7 @@ class Model {
       const now = Utils.now(this.sequelize.options.dialect);
 
       // Attach createdAt
-      if (createdAtAttr && !updateValues[createdAtAttr]) {
+      if (createdAtAttr && !insertValues[createdAtAttr]) {
         const field = this.rawAttributes[createdAtAttr].field || createdAtAttr;
         insertValues[field] = this._getDefaultTimestamp(createdAtAttr) || now;
       }


papb:
Hello! I see you are a first-time contributor, thank you for taking the time to help Sequelize!

> We are using milliseconds for our `createdAt` and `updatedAt` columns, by setting `defaultValue: Sequelize.fn('NOW',3)`. It works great, except when we `Upsert`, in which case only `updatedAt` has milliseconds.

Can you please create a [quick SSCCE](https://github.com/papb/sequelize-sscce) to show this?

ptrwtts:
While making the SSCCE, I seemed to find a workaround, that fixed the problem for us. We were defining our model like so:

```
        name: DataTypes.TEXT,	        name: DataTypes.TEXT,
        pass: DataTypes.TEXT	        createdAt: {
          field: 'date_created',
          type: Sequelize.DATE(3),
          allowNull: true,
          defaultValue: Sequelize.fn('NOW', 3)
        },
        updatedAt: {
          field: 'date_modified',
          type: Sequelize.DATE(3),
          allowNull: true,
          defaultValue: Sequelize.fn('NOW', 3)
        }
    },{
      createdAt: 'date_created',
      updatedAt: 'date_modified',
    });	    });
``` 

But changing to this made it work:

```
        name: DataTypes.TEXT,	        name: DataTypes.TEXT,
        pass: DataTypes.TEXT	        createdAt: {
          field: 'date_created',
          type: Sequelize.DATE(3),
          allowNull: true,
          defaultValue: Sequelize.fn('NOW', 3)
        },
        updatedAt: {
          field: 'date_modified',
          type: Sequelize.DATE(3),
          allowNull: true,
          defaultValue: Sequelize.fn('NOW', 3)
        }
    });	    });
``` 

That said, I still think there may be a bug. With the first snippet, it was creating the columns as DATETIME instead DATETIME(3) for some reason that I could not figure out. So my SSCCE did not work as I hoped. But if you look at the query (in MySQL test), you can see how it is strange:

```
INSERT INTO `Users` (`name`,`date_created`,`date_modified`) VALUES (?,?,NOW(3)) ON DUPLICATE KEY UPDATE `name`=VALUES(`name`);
```

Why does the modified column use NOW(3) but not the created? This is the issue I encountered. But while trying to get the SSCCE to work, I found a workaround. So I'm not sure if there really is a bug, or we were just doing something wrong. Hope that helps.

papb:
Hmmm... Thanks for the SSCCE and the extra information :) The behavior is strange indeed, apparently. I don't have time to take a deeper look now though. Glad that you found a workaround for it.

By the way, usually this `defaultValue: Sequelize.fn('NOW', 3)` that you are applying to `createdAt` and `updatedAt` shouldn't be necessary in Sequelize, probably could even be causing part of the issues; instead Sequelize populates those fields automatically under the hood. This reminded me of #5561...

sushantdhiman:
Issue is with this config

```js
createdAt: 'date_created',
updatedAt: 'date_modified',
```

These are attribute names not column names, so sequelize either looks for attribute with name `date_created` if not it creates them. With that config your model got two createdAt attributes i.e. `createdAt` and `date_created`, both pointing to same column `date_created`.

PR #11538, opened 2019-10-11
fix(postgres): update upsert regex to match the last `RETURNING *`

### LINKED ISSUE #11523: Unable to upsert on postgres text containing "RETURNING *" substring
## Issue Description

When using upsert on postgres, the string `RETURNING *` anywhere inside a text or json column value is being replaced with `RETURNING \"id\" INTO primary_key`. Besides changing the data to be upserted, this also makes it throw an error when inserting or updating into a JSON column since its not valid JSON.

### What are you doing?

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/9

It should be possible to upsert a text or JSON value that happens to contain the substring `RETURNING *`.

The `RETURNING *` substring in the input is being replaced with `RETURNING \"id\" INTO primary_key`

### Environment

- Sequelize version: 5.19.4 and 4.44.3
- Node.js version: 10.16.3
- Operating System: Ubuntu 14.04.5 LTS

I think this problem happens only for the following dialect(s): postgres

No, I don't have the time and I wouldn't even know how to start.

Americas:
### Pull Request check-list

### Description of change

Fixes #11523

Postgres' upsert query generator replaces the first `RETURNING *` it finds on the insert and update query, so if the data being added has that string, it replaces it (so it replaces the wrong `RETURNING *`) changing the input data and messing with the the final query.

This PR updates the replace method to use a regex with a negative lookahead so it only chooses the last `RETURNING *` in the query.

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -347,8 +347,12 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
     const insert = this.insertQuery(tableName, insertValues, model.rawAttributes, upsertOptions);
     const update = this.updateQuery(tableName, updateValues, where, upsertOptions, model.rawAttributes);
 
-    insert.query = insert.query.replace('RETURNING *', `RETURNING ${primaryField} INTO primary_key`);
-    update.query = update.query.replace('RETURNING *', `RETURNING ${primaryField} INTO primary_key`);
+    if (options.returning) {
+      const returningRegex = /RETURNING \*(?![\s\S]*RETURNING \*)/;
+
+      insert.query = insert.query.replace(returningRegex, `RETURNING ${primaryField} INTO primary_key`);
+      update.query = update.query.replace(returningRegex, `RETURNING ${primaryField} INTO primary_key`);
+    }
 
     return this.exceptionFn(
       'sequelize_upsert',
--- a/test/unit/dialects/postgres/query-generator.test.js
+++ b/test/unit/dialects/postgres/query-generator.test.js
@@ -44,12 +44,14 @@ if (dialect.startsWith('postgres')) {
           expectation: 'CREATE DATABASE "myDatabase" ENCODING = \'UTF8\' LC_COLLATE = \'en_US.UTF-8\' LC_CTYPE = \'zh_TW.UTF-8\' TEMPLATE = \'template0\';'
         }
       ],
+
       dropDatabaseQuery: [
         {
           arguments: ['myDatabase'],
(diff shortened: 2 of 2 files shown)


papb:
I have edited

> Fixes issue #11523.

to the following

> Fixes #11523

so that GitHub detects the fix automatically.

papb:
By the way, I have a question, was [the SSCCE (papb/sequelize-sscce#9)](https://github.com/papb/sequelize-sscce/pull/9) useful to you? Or do you think you could have fixed the issue as easily without it?

Americas:
> By the way, I have a question, was [the SSCCE (papb/sequelize-sscce#9)](https://github.com/papb/sequelize-sscce/pull/9) useful to you? Or do you think you could have fixed the issue as easily without it?

I used it yes, just changed the log to an error if the result was different from the expected. It helps to have the problem easily available, otherwise there is this great setup cost to start analyzing the issue :smile:

papb:
@Americas Nice! I'm happy that my idea of an SSCCE repo was useful then :)

papb:
I was thinking... Would you be willing to add an explicit test for the [JSONB case](https://github.com/papb/sequelize-sscce/blob/93794ffcbf245fa279b5f693730ff8ffde4eed10/src/sscce.js#L23) as well?

lib/dialects/postgres/query-generator.js
```diff
@@ -347,8 +347,10 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
     const insert = this.insertQuery(tableName, insertValues, model.rawAttributes, upsertOptions);
     const update = this.updateQuery(tableName, updateValues, where, upsertOptions, model.rawAttributes);
 
-    insert.query = insert.query.replace('RETURNING *', `RETURNING ${primaryField} INTO primary_key`);
-    update.query = update.query.replace('RETURNING *', `RETURNING ${primaryField} INTO primary_key`);
+    if (options.returning) {
+      insert.query = insert.query.replace(/RETURNING \*(?![\s\S]*RETURNING \*)/, `RETURNING ${primaryField} INTO primary_key`);
```
SimonSchick: Nit: Regex should be moved to a constant.

Americas: Done.

Americas:
> I was thinking... Would you be willing to add an explicit test for the [JSONB case](https://github.com/papb/sequelize-sscce/blob/93794ffcbf245fa279b5f693730ff8ffde4eed10/src/sscce.js#L23) as well?

Added JSON example from SSCCE and a normal example.

PR #11542, opened 2019-10-12
refactor(lib): simplify some method

VictorKmin:
change if(){condition} else{} to condition ? true : false

love you

### Pull Request check-list


lib/associations/has-many.js
```diff
@@ -212,11 +212,9 @@ class HasMany extends Association {
       where;
 
     if (Object.prototype.hasOwnProperty.call(options, 'scope')) {
-      if (!options.scope) {
-        Model = Model.unscoped();
-      } else {
-        Model = Model.scope(options.scope);
-      }
+      Model = !options.scope ?
```
Americas: WDYT about switching the statements and removing the negation?

lib/sequelize.js
```diff
@@ -1232,8 +1233,8 @@ class Sequelize {
     if (Object.prototype.hasOwnProperty.call(attribute, 'defaultValue')) {
       if (typeof attribute.defaultValue === 'function' && (
         attribute.defaultValue === DataTypes.NOW ||
-          attribute.defaultValue === DataTypes.UUIDV1 ||
-          attribute.defaultValue === DataTypes.UUIDV4
+        attribute.defaultValue === DataTypes.UUIDV1 ||
+        attribute.defaultValue === DataTypes.UUIDV4
```
Americas:
```diff
-        attribute.defaultValue === DataTypes.NOW ||
-        attribute.defaultValue === DataTypes.UUIDV1 ||
-        attribute.defaultValue === DataTypes.UUIDV4
+        [DataTypes.NOW, DataTypes.UUIDV1, DataTypes.UUIDV4].includes(attribute.defaultValue)
```

papb:

I will review your PR now

lib/data-types.js
```diff
@@ -441,16 +441,19 @@ class DATE extends ABSTRACT {
     return value;
   }
   _isChanged(value, originalValue) {
```
papb: I like that you wanted to refactor this `_isChanged` method, however I think an even better refactor can be done:

```js
_isChanged(value, originalValue) {
  if (value instanceof Date && originalValue instanceof Date) {
    return value.getTime() !== originalValue.getTime();
  }
  return value !== originalValue;
}
```

Please convince yourself that my code above is equivalent to the current code and use it instead :)

lib/data-types.js
```diff
@@ -488,11 +491,8 @@ class DATEONLY extends ABSTRACT {
     if (originalValue && !!value && originalValue === value) {
       return false;
     }
-    // not changed when set to same empty value
-    if (!originalValue && !value && originalValue === value) {
-      return false;
-    }
-    return true;
+
+    return !(!originalValue && !value && originalValue === value);
```
papb: Same as above, please use my improved refactor for the whole function.

lib/query-interface.js
```diff
@@ -214,16 +213,12 @@ class QueryInterface {
...
-    }
+    const promise = this.sequelize.options.dialect === 'postgres' ?
+      PostgresQueryInterface.ensureEnums(this, tableName, attributes, options, model) :
+      Promise.resolve();
 
-    if (
-      !tableName.schema &&
-      (options.schema || !!model && model._schema)
-    ) {
+    const canAddSchema = !tableName.schema && (options.schema || !!model && model._schema);
```
papb: The `!!` boolean cast is unnecessary, also I know it's debatable but I think adding parentheses around the `&&` improves readability; not everyone knows that `&&` has higher precedence than `||` in JS.

```suggestion
    const canAddSchema = !tableName.schema && (options.schema || (model && model._schema));
```

lib/sequelize.js
```diff
@@ -1109,7 +1109,8 @@ class Sequelize {
           // Rollback transaction if not already finished (commit, rollback, etc)
           // and reject with original error (ignore any error in rollback)
           return Promise.try(() => {
-            if (!transaction.finished) return transaction.rollback().catch(() => {});
+            if (!transaction.finished) return transaction.rollback().catch(() => {
+            });
```
papb: I think the previous format was better here, please revert

lib/sequelize.js
papb: I agree with @Americas here as well :+1:

lib/sequelize.js
VictorKmin: My IDE do this automatically. I will revert. Sorry about that.

WikiRik:
Sorry for not merging this earlier, but I will close this since much has changed since October 2019.

Issue #11533, opened 2019-10-10
Sequelize.fn doesn't work with bind replacements if value has `$` in it 


andrew8088:
### What are you doing?

I'm inserting a record and using a database function (like `upper`) to format a field with a `$` in it.

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/11

```js
// more details in the link above

const User = sequelize.define('User', {
	name: DataTypes.TEXT
});

await User.create({
	name: sequelize.fn('upper', '$user3')
});
```

### What do you expect to happen?

I expect to successfully insert the record.

### What is actually happening?

I get the error: 

```
Named bind parameter "$user3" has no value in the given object.
```

### Environment

- Sequelize version: 5.16.0
- Node.js version: 10.15.0
- Operating System: MacOS X, Linux

### How does this problem relate to dialects?

- [x] I think this problem happens regardless of the dialect.

### Would you be willing to resolve this issue by submitting a Pull Request?

- [x] Yes, I have the time but I don't know how to start, I would need guidance.

papb:
Thanks for the report!!

> ### Would you be willing to resolve this issue by submitting a Pull Request?
> * [x]  Yes, I have the time but I don't know how to start, I would need guidance.

Fantastic. What kind of guidance do you want? The solution to this problem is probably to detect that the string given by the user has a `$` in it and escape it properly. If you don't know how to start, I suggest you start by creating a failing test. Let me know if you need further help.

zacharysierakowski:
I spent some time looking into this as well. Should this actually be escaped? As I followed the code path to where I think the escape should happen, I got worried it could effect other instances of `Utils.SequelizeMethod`.

If we escape it, won't that mean that there is no support for variables? Maybe the solution here would be to document that if you need to do this, make sure to escape the dollar signs on your own. If you are aware of this and know to escape it when using `sequalize.fn`, it works as expected. 

For example:
```
sequelize.fn('upper', '$$user3')
```
will result in `$USER3` right now.

Heads up - this is the first time I've looked at Sequelize code, so there might be some history or context here that I'm missing. I'd be happy to put a PR up if it's decided that `$` should definitely be escaped.

zacharysierakowski:
I created the PR above mainly as an example for now, but it also couldn't hurt to leave it in the codebase depending on the outcome of discussion in this issue.

papb:
@zacharysierakowski Thanks for taking the time to help us!!

[...]

PR #11551, opened 2019-10-15
Added unit test as an example and to cover dollarsign in sequelize.fn

zacharysierakowski:
### Description of change

Added unit test that covers dollar sign in `sequelize.fn`. This serves mainly as an example for now, but might also be merged depending on conversation in the related issue. 

Related to #11533.


papb:
Great!! Thanks. Let's proceed the discussion on #11533 and later we can make some adjustments if needed and merge this!

PR #11553, opened 2019-10-15
Document .sequelizerc configuration options

samuliasmala:
The .sequelizerc has several configuration options which are defined and described in the source code (cli/src/core/yargs.js) but not included in the documentation.


sushantdhiman:
Linting fails

PR #11578, opened 2019-10-19
fix(associations): allow binary key for belongs-to-many

### LINKED ISSUE #11545: "has" method of BelongsToMany not working when data type of key is binary
## Issue Description
I tried to use belongsToMany association with binary key,
but it's not working fine, always return false.

### What are you doing?

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/13 

return ture when already have association

Please see SSCCE link.

### Environment

- Sequelize version: 5.19.6 
- Node.js version: 10.16.3 
- Operating System: Mac

tsasaki609:
### Pull Request check-list

### Description of change

fix #11545

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -532,7 +532,8 @@ class BelongsToMany extends Association {
     };
 
     return this.get(sourceInstance, options).then(associatedObjects =>
-      _.differenceBy(instancePrimaryKeys, associatedObjects, this.targetKey).length === 0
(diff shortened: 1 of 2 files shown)


tsasaki609:
Oops CI failed...
>SequelizeDatabaseError: BLOB/TEXT column 'id' used in key specification without a key length

MySQL requires a length specification for the BLOB data type.
I'll try to fix.

sushantdhiman:
@tsasaki609 merged to master which is currently pointing to `v6-beta`, if you need a `v5` release please submit a new PR for `v5` branch as well

Issue #11027, opened 2019-06-03
Missing Typings for QueryInterface


a88zach:
Many of the QueryInterface methods accept `string | { tableName: string, schema: string }` as the first argument.  For instance, the following is valid and will run successfully.

```
queryInterface.bulkDelete({ tableName: 'foo', schema: 'bar' } as any, {}, {});
```

However, the typings only support a `string` as the first argument.

**To Reproduce**
Create a migration with the above code using typescript

## What do you expect to happen?
The code should compile without error

## What is actually happening?
TSC will not compile without casting to any

## Environment
TypeScript version: XXX
Tested with latest release: 3.4.5
- [ ] No
- [X] Yes, specify that version:

papb:
Would you be willing to submit a PR?

ephys:
I believe this has since been fixed, as these methods now accept `TableName` which is equal to

```ts
export type TableName = string | TableNameWithSchema;
```

PR #11582, opened 2019-10-19
fix: fix type of methods in query-interface

2kindsofcs:
### Pull Request check-list

(npm run test-typings, actaully)

### Description of change
as described in #11027, many of the QueryInterface methods accept something like `string | { tableName: string, schema: string }` as the first argument. however, they are accepting tableName with string type only. So I added interface and type so that they could get arguments other than string.

### DIFF
--- a/types/lib/query-interface.d.ts
+++ b/types/lib/query-interface.d.ts
@@ -120,14 +120,14 @@ export interface QueryInterfaceDropAllTablesOptions extends QueryInterfaceOption
   skip?: string[];
 }
 
-export interface TablenameWithSchema {
+export interface TableNameWithSchema {
   tableName: string;
   schema?: string;
   delimiter?: string;
   as?: string;
   name?: string;
 }
-export type Tablename = string | TablenameWithSchema;
+export type TableName = string | TableNameWithSchema;
 
 export type IndexType = 'UNIQUE' | 'FULLTEXT' | 'SPATIAL';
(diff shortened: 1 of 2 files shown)


papb:
Hi, thanks for the PR, can you please add more tests?

types/lib/query-interface.d.ts
```diff
@@ -120,6 +120,15 @@ export interface QueryInterfaceDropAllTablesOptions extends QueryInterfaceOption
   skip?: string[];
 }
 
+export interface TablenameWithSchema {
```
SimonSchick: Should probably be `TableName`.

2kindsofcs:
@papb  hi. I added some tests. I was wondering if you think there should be more tests.

types/lib/query-interface.d.ts
2kindsofcs: Do you mean it should be TableName or TableNameWithSchema?

papb: He probably means `TableNameWithSchema`

types/lib/query-interface.d.ts
```diff
@@ -120,6 +120,15 @@ export interface QueryInterfaceDropAllTablesOptions extends QueryInterfaceOption
...
+  tableName: string;
+  schema?: string;
+  delimiter?: string;
+  as?: string;
+  name?: string;
+}
+export type Tablename = string | TablenameWithSchema;
```
papb:
```suggestion
export type TableName = string | TableNameWithSchema;
```

papb:
@2kindsofcs Thanks, I think those tests are enough now.

PR #11584, opened 2019-10-20
feat: support cls-hooked / tests

sushantdhiman:
### Pull Request check-list

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -3,7 +3,6 @@
 const url = require('url');
 const path = require('path');
 const retry = require('retry-as-promised');
-const clsBluebird = require('cls-bluebird');
 const _ = require('lodash');
 
 const Utils = require('./utils');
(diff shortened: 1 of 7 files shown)


lib/sequelize.js
```diff
@@ -1130,8 +1129,9 @@ class Sequelize {
     // save namespace as `Sequelize._cls`
     this._cls = ns;
 
-    // patch bluebird to bind all promise callbacks to CLS namespace
-    clsBluebird(ns, Promise);
+    Promise.config({
+      asyncHooks: true
```
jedwards1211, line 1133: @sushantdhiman builtlin Promises automatically support async hooks right?  Want to know if I need to replace this with something when we get rid of bluebird.

papb: Good question @jedwards1211, I don't know the answer, hopefully @sushantdhiman does. However, I'd guess that you could just try removing it, in theory we should have tests that fail if this stops working...

sushantdhiman: @jedwards1211 No, you need to manually enable them, https://nodejs.com/api/async_hooks.html#async_hooks_promise_execution_tracking. But we don't need to enable anything. `cls-hooked` enable those api when namespace is created.

So no change will be required at sequelize level

jedwards1211: okay great

Issue #11378, opened 2019-09-03
typings for PoolOptions.evict is not defined


wata-lb:
## Issue Description
The typings for PoolOptions.evict is not defined.

### What are you doing?
Upgrading to Sequelize 5 doing something like this will no longer compile in TypeScript`

```js
export const sequelize = new Sequelize(process.env.DATABASE_URL, {
        pool: {
            evict: 20000,
          },
    });
```

### What do you expect to happen?

PoolOptions in sequelize.d.ts, define evict:number

### What is actually happening?
```
error TS2345: Argument of type '{ pool: { evict: number; }; }' is not assignable to parameter of type 'SequelizeOptions'.
  Types of property 'pool' are incompatible.
    Type '{ evict: number; }' is not assignable to type 'PoolOptions'.
      Object literal may only specify known properties, and 'evict' does not exist in type 'PoolOptions'.

77             evict: 20000,
```

### Environment

- Sequelize version: 5.18.0 
- Node.js version: v8.16.0 
- Operating System: Mac OS 10.14.6
- If TypeScript related: TypeScript version: 3.4.5

### How does this problem relate to dialects?

- [x] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?

- [x] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

PR #11596, opened 2019-10-22
Support `{ plain: true }` in typescript typings for sequelize.query

### LINKED ISSUE #11595: Typescript bindings incomplete when using `sequelize.query()`
###  Issue Description
It seems that the Typescript bindings are incomplete when using `.query()` with the `plain: true`.
It always returns an array of objects and doesn't take the `plain` option into account which should return a single row.

Adding the following typing resolves the issue but not entirely sure if it's the way to go.

```
public query<T extends object>(sql: string | { query: string; values: unknown[] }, options: QueryOptionsWithType<QueryTypes.SELECT> & { plain: true }): Promise<T>;
```

### What are you doing?

```js
const { count } = await sequelize.query(
    `
    SELECT COUNT(1) as count
    FROM foo f
    WHERE
      f.status = 'OK'
  `,
    {
      plain: true,
    },
  );
```

The above query should return a single row and successfully destructure `count` from the single row that was returned using the `plain` option.

Instead Typescript complains about the following:
`Property 'count' does not exist on type 'object[]'.`

### Environment

- Sequelize version: `5.21.1`
- Node.js version: `10.15.3`
- Operating System: `macOS Mojave`
- If TypeScript related: TypeScript version: `3.6.2`

segersniels:
### Pull Request check-list

### Description of change

Closes #11595

### Related PR's
- https://github.com/sequelize/sequelize/pull/11929

### DIFF
--- a/types/test/sequelize.ts
+++ b/types/test/sequelize.ts
@@ -67,11 +67,18 @@ sequelize.query('SELECT * FROM `user`', { type: QueryTypes.RAW }).then(result =>
 });
 
 sequelize
-  .query("SELECT COUNT(1) as count FROM `user`", {
+  .query<{ count: number }>("SELECT COUNT(1) as count FROM `user`", {
     type: QueryTypes.SELECT,
     plain: true
   })
   .then(result => {
-    const oneRowOnly = (a: any) => a;
-    oneRowOnly(result);
+    result.count.toExponential(); // is a number!
+  });
+
+sequelize
+  .query("SELECT COUNT(1) as count FROM `user`", {
+    plain: true
+  })
+  .then(result => {
+    console.log(result.count);
   });
--- a/types/test/sequelize.ts
+++ b/types/test/sequelize.ts
@@ -65,3 +65,13 @@ sequelize.query('SELECT * FROM `user`', { type: QueryTypes.RAW }).then(result =>
   const arraysOnly = (a: any[]) => a;
   arraysOnly(data);
 });
+
+sequelize
+  .query("SELECT COUNT(1) as count FROM `user`", {
+    type: QueryTypes.SELECT,
+    plain: true
+  })
+  .then(result => {
+    const oneRowOnly = (a: any) => a;
+    oneRowOnly(result);
+  });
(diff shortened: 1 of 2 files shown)


papb:

Please add at least one test to make sure your implementation works as intented and to prevent regressions in the future. Take a look into recent merged PRs to see how others created tests. If you have trouble, let me know.

Testing the TypeScript typings consists simply on checking if some code can be compiled. In other words, a typings test is just some TS code that should compile. What it does when executed is irrelevant, the test is simply to compile it without errors, as can be seen in our CI configuration (using `npm run test-typings`) [here](https://github.com/sequelize/sequelize/blob/bd59b8729b8d898d2f3ce8deb764777bab7e57d5/.travis.yml#L49) and [here](https://github.com/sequelize/sequelize/blob/bd59b8729b8d898d2f3ce8deb764777bab7e57d5/package.json#L152). Since it's just a compilation, getting no output from it means that it passed. Getting a compilation error is a failure. Note that if you directly throw an error from your typings test code, the test will still pass because throwing an error is something that compiles just fine.

For good examples on how to write typescript typings, check the following PRs: #11368 #11378 #11520

segersniels:
@papb Added a test 👌

types/test/sequelize.ts
```diff
@@ -65,3 +65,13 @@ sequelize.query('SELECT * FROM `user`', { type: QueryTypes.RAW }).then(result =>
...
 });
+
+sequelize
+  .query("SELECT COUNT(1) as count FROM `user`", {
+    type: QueryTypes.SELECT,
+    plain: true
+  })
+  .then(result => {
+    const oneRowOnly = (a: any) => a;
+    oneRowOnly(result);
```
papb: I don't think this is really testing anything, because it is passing `result` into an parameter that expects `any`, so it will certainly work anyway.

papb:
Hello! I ended up having direct improvement ideas for your PR while I was reviewing it. I decided to commit my ideas directly in your PR to save time for both of us :) Please let me know if you disagree with any of the changes I made.

types/test/sequelize.ts
segersniels: Yeah, was a bit confused by the way the other tests were written 😅

segersniels:
@papb Agree with the changes you made, should be sufficient!

papb:
Thank you!!

PR #11606, opened 2019-10-24
fix: escape dollarsign in sequelize.fn related to issue 11533

schopr9:
### Pull Request check-list

### Description of change
https://github.com/sequelize/sequelize/issues/11533
- changed the abstract query generator handleSequelizeMethod to escape dollar ($) in argument

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -943,7 +943,6 @@ class QueryGenerator {
         }
       }
     }
-
     return SqlString.escape(value, this.options.timezone, this.dialect);
   }
 
@@ -2134,7 +2133,7 @@ class QueryGenerator {
         if (_.isPlainObject(arg)) {
           return this.whereItemsQuery(arg);
         }
-        return this.escape(arg);
+        return this.escape(arg.replace('$', '$$$'));
       }).join(', ')})`;
     }
     if (smth instanceof Utils.Col) {
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -2133,7 +2133,7 @@ class QueryGenerator {
         if (_.isPlainObject(arg)) {
           return this.whereItemsQuery(arg);
         }
-        return this.escape(arg.replace('$', '$$$'));
+        return this.escape(typeof arg === 'string' ? arg.replace('$', '$$$') : arg);
       }).join(', ')})`;
     }
     if (smth instanceof Utils.Col) {
--- a/test/integration/model/create.test.js
+++ b/test/integration/model/create.test.js
@@ -330,6 +330,16 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       });
     });
 
+    it('returns instance if username is a sequelize fn with $ in value', function() {
+      const data = {
+        username: Sequelize.fn('upper', '$username')
+      };
+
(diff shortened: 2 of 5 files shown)


papb:

Tests are failing though, please take a look :)

Also your branch was outdated, I just brought it up to date for you (don't worry about this):

image

Let me know if you need help. Thanks!!

test/integration/model/create.test.js
```diff
@@ -330,6 +330,16 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       });
     });
 
+    it('returns instance if username is a sequelize fn with $ in value', function() {
+      const data = {
+        username: Sequelize.fn('upper', '$username')
+      };
+
+      return this.User.create(data).then(user => {
+        expect(user.username).to.equal('$USERNAME');
```
schopr9: @papb this test is failing for mysql and mariadb, is this some other bug or those dialect doesn't support functions

lib/dialects/abstract/query-generator.js
```diff
@@ -2134,7 +2133,7 @@ class QueryGenerator {
         if (_.isPlainObject(arg)) {
           return this.whereItemsQuery(arg);
         }
-        return this.escape(arg);
+        return this.escape(typeof arg === 'string' ? arg.replace('$', '$$$') : arg);
```
papb: Just one question, out of curiosity, why `$$$`? I would have guessed `$$`.

test/unit/dialects/abstract/query-generator.test.js
```diff
@@ -98,6 +98,12 @@ describe('QueryGenerator', () => {
       QG.handleSequelizeMethod(this.sequelize.where(this.sequelize.col('foo'), Op.not, null))
         .should.be.equal('foo IS NOT NULL');
     });
+
+    it('should correctly escape $ in sequelize.fn arguments', function() {
+      const QG = getAbstractQueryGenerator(this.sequelize);
+      QG.handleSequelizeMethod(this.sequelize.fn('upper', '$user'))
+        .should.include('$$user');
```
papb: And why this test checks for only `$$`? Couldn't it check for `$$$`? Asking out of curiosity, I approve this PR regardless :)

lib/dialects/abstract/query-generator.js
schopr9: The replacement string with special replacement patterns
`$$ | Inserts a "$".`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace
`> a = '$sequelize'
'$sequelize'
> a.replace
[Function: replace]
> a.replace('$', '$$')
'$sequelize'
> a.replace('$', '$$$')
'$$sequelize'`

papb: Ahhh, of course, `$` has a special meaning within the JS escape function, so you had to escape the escape so that the escape escapes correctly :)

kristijanhusak:
I'm having this problem:

```js
const theCode = '10$off';
PromoCodes.findAll({
  where: {
    code: sequelize.where(sequelize.fn('lower', sequelize.col('code')), sequelize.fn('lower', theCode))
  }
});
```

This never finds it because it does `lower("code") = lower('10$$off')`. Solution was to do `theCode.toLowerCase()`, but this is still a bug.

Issue #11569, opened 2019-10-17
BulkCreate with composite unique constraint does not work


cjancsar:
### What are you doing?
I am trying to use `bulkCreate` to `updateOnDuplicate` when a duplicate is detected. Duplicate should be detected by a "multi column unique index".

```js
"use strict";

const Sequelize = require("./index");
const sequelize = require("./test/support").createSequelizeInstance();
const chai = require("chai"),
  expect = chai.expect;

class Bar extends Sequelize.Model {}
Bar.init(
  {
    value: Sequelize.DataTypes.STRING,
    timestamp: {
      type: Sequelize.DataTypes.DATE,
      allowNull: false
    }
  },
  {
    indexes: [
      {
        name: "unique_timestamp_per_foo",
        unique: true,
        fields: ["timestamp", "FooId"]
      }
    ],
    sequelize
  }
);

class Foo extends Sequelize.Model {}
Foo.init({}, { sequelize });

Bar.belongsTo(Foo, {
  foreignKey: { allowNull: false },
  onDelete: "CASCADE"
});
Foo.hasMany(Bar, { foreignKey: { allowNull: false }, onDelete: "CASCADE" });

(async () => {
  await sequelize.sync({ force: true });

  // Create a _parent_ `Foo` model
  const foo = await Foo.create();

  // Create a `Bar` object
  const originalBar = await Bar.create({
    value: "original",
    timestamp: new Date(Date.UTC(2016, 0, 1)),
    FooId: foo.id
  });

  // We obviously expect a `Bar` object to exist with the unique constraints of `timestamp` and `FooId`
  expect(
    await Bar.count({
      where: { FooId: foo.id, timestamp: new Date(Date.UTC(2016, 0, 1)) }
    })
  ).to.equal(1);

  // We create a modified `Bar` object that has an overlapping `timestamp` and `FooId`
  // This should be detected as duplicate by the `uniqueTimestampPerFoo` unique constraint/index,
  // during the next `bulkCreate` operation with `updateOnDuplicate` set.
  const modifiedBar = Bar.build({
    value: "modified",
    timestamp: new Date(Date.UTC(2016, 0, 1)),
    FooId: foo.id
  });

  // "Bulk" create to "upsert" a modified `existingBar` when conflict detected.
  // We only want `value` and `updatedAt` to be updated on conflict.
  await expect(
    Bar.bulkCreate([modifiedBar.toJSON()], {
      updateOnDuplicate: ["value", "updatedAt"],
[...]

PR #11611, opened 2019-10-25
feat(postgres): Add `upsertIndex` and `upsertKeys` options to `BulkCreateOptions`

### LINKED ISSUE #11534: Sqlite: bulkCreate with updateOnDuplicate fails when the model contains a composite key.
Related issue: #11432

## Issue Description
The `bulkCreate` fails if the model contains a composite key and doesn't explicitly define a primary key.

The `upsertKeys` array is emptied when the model contains a composite key since the length of the fields array of the `compositeKey` object will be of size > 1 (which causes this filter to skip it):

https://github.com/sequelize/sequelize/blob/3a6006930eda19e608342f4eee1bf3f1de4e8609/lib/model.js#L2705

My model looks like this:
```javascript
"use strict";
  const User = sequelize.define(
    "User",
    {
      username: { type: DataTypes.STRING, unique: "compositeKey" },
      discordId: { type: DataTypes.STRING, unique: "compositeKey" }
    },
    {}
  );
  User.associate = function(models) {
    // associations can be defined here
  };
  return User;
};
```
In this case, the `model.uniqueKeys` object only contains a single `compositeKey` object.
Shouldn't the `model.uniqueKeys` object _also_ contain the automatically generated `id` primary key for this model?

Changing the model to this works:
```javascript
"use strict";
  const User = sequelize.define(
    "User",
    {
      username: { type: DataTypes.STRING, unique: true },
      discordId: { type: DataTypes.STRING, unique: true }
    },
    {}
  );
  User.associate = function(models) {
    // associations can be defined here
  };
  return User;
};
```

### What are you doing?
Upserting users for testing:
```js
models.sequelize
  .sync()
  .then(async () => {
    if (process.env.UPSERT_TEST_USERS === "true") {
      const numTestUsers = parseInt(process.env.NUM_TEST_USERS || 50);
      const testUsers = new Array(numTestUsers).fill(null).map((user, i) => ({
        id: i,
        username: `Test${i}`,
        discordId: `discordId${i}`
      }));

      await models.User.bulkCreate(testUsers, {
        updateOnDuplicate: ["username", "discordId"]
      });
    }
  })
  .then(() => {
    debug("Database synchronization succeeded.");

    /**
     * Listen on provided port, on all network interfaces.
     */
    server.listen(port);
    server.on("error", onError);
    server.on("listening", onListening);
  })
  .catch(err => {
    debug(`Database synchronization failed: ${err}`);
  });
```

I expect the `bulkCreate` method to succeed and upsert the test users to the database.

Doing this with this model results in the following error:
```
SequelizeDatabaseError: SQLITE_ERROR: near ")": syntax error
```
The error is referring to the `ON CONFLICT()`  since it is an error to have an empty `ON CONFLICT` clause. This should be:
```sql
ON CONFLICT('username', 'discordId')
```

### Environment

- Sequelize version: 5.19.3 
- Node.js version: v10.16.3 

- [x] I think this problem happens only for the following dialect(s): Postgres, Sqlite

fnimick:
This allows the use of a unique index for a `bulkCreate` with
`updateOnDuplicate` set, in postgres. The unique index may be
a partial index.

### Pull Request check-list

### Description of change

This change adds the ability to specify `upsertIndex` or `upsertKeys` in `BulkCreateOptions`, so that a `ON CONFLICT` clause can be generated using columns other than the table's primary key.

Note: I was not able to get the tests running on my machine, and would appreciate help writing a new test for this functionality.

Closes #11534, #11569

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -288,9 +288,16 @@ class QueryGenerator {
     if (this._dialect.supports.inserts.updateOnDuplicate && options.updateOnDuplicate) {
       if (this._dialect.supports.inserts.updateOnDuplicate == ' ON CONFLICT DO UPDATE SET') { // postgres / sqlite
         // If no conflict target columns were specified, use the primary key names from options.upsertKeys
-        const conflictKeys = options.upsertKeys.map(attr => this.quoteIdentifier(attr));
         const updateKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=EXCLUDED.${this.quoteIdentifier(attr)}`);
-        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) DO UPDATE SET ${updateKeys.join(',')}`;
+        if (Array.isArray(options.upsertKeys)) {
+          const conflictKeys = options.upsertKeys.map(attr => this.quoteIdentifier(attr));
+          onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) DO UPDATE SET ${updateKeys.join(',')}`;
+        } else {
+          // If upsertKeys is an object, generate an ON CONFLICT WHERE
+          const conflictKeys = options.upsertKeys.fields.map(attr => this.quoteIdentifier(attr));
+          const where = this.whereQuery(options.upsertKeys.where);
+          onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) ${where} DO UPDATE SET ${updateKeys.join(',')}`;
+        }
       } else { // mysql / maria
         const valueKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=VALUES(${this.quoteIdentifier(attr)})`);
         onDuplicateKeyUpdate = `${this._dialect.supports.inserts.updateOnDuplicate} ${valueKeys.join(',')}`;
--- a/lib/model.js
+++ b/lib/model.js
@@ -2668,10 +2668,22 @@ class Model {
         // Map updateOnDuplicate attributes to fields
         if (options.updateOnDuplicate) {
           options.updateOnDuplicate = options.updateOnDuplicate.map(attr => model.rawAttributes[attr].field || attr);
-          // Get primary keys for postgres to enable updateOnDuplicate
-          options.upsertKeys = _.chain(model.primaryKeys).values().map('field').value();
-          if (Object.keys(model.uniqueKeys).length > 0) {
-            options.upsertKeys = _.chain(model.uniqueKeys).values().filter(c => c.fields.length >= 1).map(c => c.fields).reduce(c => c[0]).value();
+          if (options.upsertIndex !== undefined) {
+            const upsertIndex = model._indexes.find(index => index.name === options.upsertIndex);
+            if (upsertIndex === undefined) {
+              throw new Error(`upsertIndex '${options.upsertIndex}' not defined.`);
+            }
+            if (!upsertIndex.unique) {
+              throw new Error(`upsertIndex '${options.upsertIndex}' not a unique index.`);
+            }
+            options.upsertKeys = { fields: upsertIndex.fields, where: upsertIndex.where };
+          }
+          if (options.upsertKeys == null) {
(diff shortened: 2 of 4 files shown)


fnimick:
Note: if anyone wants to use this while we go through the PR process, I have a `patch-package` patch available here: https://gist.github.com/fnimick/5dd3c6855cea6e000c86d1bc38876958

papb:

To help you write tests, can you please first show usage examples of your feature?

fnimick:
A usage example for `upsertIndex`:
```
model.bulkCreate(data, {
  updateOnDuplicate: ['updateField'],
  upsertIndex: 'unique_index',
});
```
NOTE: the index in question here must be defined on the model as well.

A usage example for `upsertKeys` with a partial index:
```
model.bulkCreate(data, {
  updateOnDuplicate: ['updateField'],
  upsertKeys: {
    fields: ['unique_field_one', 'unique_field_two'],
    where: { deleted_at: { [Op.is]: null } },
  },
});
```

A usage example for `upsertKeys`:
```
model.bulkCreate(data, {
  updateOnDuplicate: ['updateField'],
  upsertKeys: ['unique_field_one', 'unique_field_two'],
});
```

fnimick:
@papb just updated this to resolve conflicts with #11666 . Still not sure how to write tests but can investigate if necessary.

papb:
Hello :grimacing: thank you very much!!

papb:
If you are having trouble to write tests, I suggest you try playing around with [sequelize-sscce](https://github.com/papb/sequelize-sscce), if you can make short pieces of code that verify what you want, converting them to real tests will be almost trivial :) Let me know!

molinajulian:
Hi everyone! :wave:  Good job in this PR :tada:.
I recently have a problem with `bulkInsert` because of this bug
I think that it's a breaking change for version 5.x, it would be nice if we don't change the interface and we use the unique keys in the model for know if exist a composite unique keys or keep this change for versión 6.x and release a fix in another PR for 5.x.
@fnimick if you need help for tests, you can tell me

fnimick:
Closed because a better approach has been merged in - see https://github.com/sequelize/sequelize/pull/11984

fnimick:
@thaddeusmt went to implement my use case for this PR with your updated updateOnDuplicate and realized there's a case that my branch handles, but yours (I don't think) does - a partial unique index. In our case, I have a unique index on two fields only when a third field is null, and I generate an `ON CONFLICT` clause based on that index. Do you see an easier way to add support for that without needing to add the new functionality that I added here?

fnimick:
@papb I've now added postgres specific tests for this. Please let me know if my style of writing them is OK.

fnimick:
I've now also added sqlite tests, since this feature is supported by sqlite as well, and have improved the documentation on the new options.

fnimick:
@papb is there anything I can do to move this forward? Thanks!

malammar:
Would love to help out to get this done. It's critical for bulkCreate usage on Postgres.

ramezmagdy:
This PR is urgent to get bulk upsert work correctly  on postgres

CharlesEgan:
Agreed this is urgently needed.  Happy to help resolving the conflicts so this can move forward.

fnimick:
Unfortunately this PR was not reviewed before the rewrite in Typescript was started, so not only are there many conflicts, some of the files I have changed no longer exist.

I asked in https://github.com/sequelize/sequelize/issues/12429 about the best strategy, but got no clear answer.

I would recommend doing a raw sql query for now - or using sequelize 5.21.1 with my patch, if you absolutely must have the ORM interface which matches this PR. There's no guarantee that this interface will be merged in though, so I would recommend raw queries instead. @CharlesEgan 

https://gist.github.com/fnimick/313091df3ecc6c5e5f17be1e0950c33f

ramezmagdy:
any update about this pull request

sfalihi:
i think your solution is not working for composite unique indexes. 
the problem is you are just separating fields of composite index.
the true solution for this problem ( composite index ) is using this kind of query:
"ON CONFLICT ON CONSTRAINT constraint_name"
for example  if composite index name is "user_message_x_y_z", you should not separate them but should do this:
"ON CONFLICT ON CONSTRAINT user_message_x_y_z"

* i think "user_message_x_y_z" uniqueness is not the same as [user, message, x, y, z]'s

WikiRik:

PR #11612, opened 2019-10-25
fix(typings): Add Date type to WhereValue 

### LINKED ISSUE #11381: Typescript issue with date in where
## Issue Description
Date types are not accepted as an option in a findAll where.

### What are you doing?

```js
import { Sequelize, Model, DataTypes, BuildOptions } from 'sequelize';

const sequelize = new Sequelize('mysql://root:asd123@localhost:3306/mydb');

// We need to declare an interface for our model that is basically what our class would be
interface MyModel extends Model {
    readonly id: number;
    dt: Date;
  }

  // Need to declare the static model so `findOne` etc. use correct types.
  type MyModelStatic = typeof Model & {
    new (values?: object, options?: BuildOptions): MyModel;
  }

  // TS can't derive a proper class definition from a `.define` call, therefor we need to cast here.
  const MyDefineModel = <MyModelStatic>sequelize.define('MyDefineModel', {
    id: {
      primaryKey: true,
      type: DataTypes.INTEGER.UNSIGNED,
    },
    dt: { type: DataTypes.DATE, allowNull: true },
  });

  function stuffThree() {
    MyDefineModel.findAll({ where: { dt: new Date() } });
  }
```

tsc should compile the file without issue.

tsc complains that a Date is not assignable to `WhereAttributeHash`
I feel like `WhereValue` should contain Date, but I assume it is not there for a reason?

```
test.ts:26:29 - error TS2322: Type '{ dt: Date; }' is not assignable to type 'WhereOptions'.
  Type '{ dt: Date; }' is not assignable to type 'WhereAttributeHash'.
    Property 'dt' is incompatible with index signature.
      Type 'Date' is not assignable to type 'string | number | boolean | WhereAttributeHash | AndOperator | OrOperator | Literal | Where | Fn | Col | WhereOperators | WhereGeometryOptions | (string | number | WhereAttributeHash)[]'.
        Type 'Date' is not assignable to type 'WhereAttributeHash'.
          Index signature is missing in type 'Date'.

26     MyDefineModel.findAll({ where: { dt: new Date() } });
                               ~~~~~

  node_modules/sequelize/types/lib/model.d.ts:64:3
    64   where?: WhereOptions;
         ~~~~~
    The expected type comes from property 'where' which is declared here on type 'FindOptions'


Found 1 error.
```

### Environment

- Sequelize version: 5.17.2
- Node.js version: v10.15.3
- Operating System: Ubuntu 18.04.3 LTS
- If TypeScript related: TypeScript version: 3.4.5

ibarrae:
### Pull Request check-list

### Description of change

Add `Date` type in `WhereValue`. Closes #11381

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -341,6 +341,7 @@ export type WhereValue =
   | string // literal value
   | number // literal value
   | boolean // literal value
+  | Date // literal value
   | Buffer // literal value
   | null
   | WhereOperators
--- a/types/test/where.ts
+++ b/types/test/where.ts
@@ -22,6 +22,7 @@ where = {
   buffer: Buffer.alloc(0),
   buffers: [Buffer.alloc(0)],
   null: null,
+  date: new Date()
 };
 
 // Operators


nabilzhafri:
I'm using version 5.21.5, but apparently this fix seems to be missing. This hasnt been deployed yet?

djejaquino:
> I'm using version 5.21.5, but apparently this fix seems to be missing. This hasnt been deployed yet?

[Looks like so.](https://github.com/sequelize/sequelize/blob/3d2df28e9fcfc6a640dbad21b0435ed8804602d4/types/lib/model.d.ts#L344)

PR #11618, opened 2019-10-28
More precise beforeDestroy type

thesame:
This is just an update to typescript model declaration that fixes "beforeDestroy" callback type.


papb:
Thanks!!

Issue #11178, opened 2019-07-10
Discrepancy in docs vs typescript types for locks


michaelfarrell76:
## What are you doing?

I am attempting to lock a table when I perform a find query. The documentation seems to have a discrepancy in how these locks can be created, however the typescript types do not sync up

The following will produce a type error:

```ts
await MigrationLock.findAll({
  where: { isLocked: false },
  transaction,
  lock: true,
});
```

See below for all of the typescript warnings that also seem to be valid code as stated in the docs [here](http://docs.sequelizejs.com/manual/transactions.html#locks) and [here](http://docs.sequelizejs.com/class/lib/transaction.js~Transaction.html#static-get-LOCK)

![Screen Shot 2019-07-10 at 1 02 16 PM](https://user-images.githubusercontent.com/10264973/61000950-65415f80-a313-11e9-8c74-4823ed624f07.png)
![Screen Shot 2019-07-10 at 1 02 37 PM](https://user-images.githubusercontent.com/10264973/61000951-65d9f600-a313-11e9-9723-8bc74b12ef36.png)
![Screen Shot 2019-07-10 at 1 02 41 PM](https://user-images.githubusercontent.com/10264973/61000953-65d9f600-a313-11e9-8778-6b22e32bb733.png)

**To Reproduce**
1. Define model `MigrationLock` using typescript
2. Try to provide `lock: true` in `findAll`
3. See typescript error

## What do you expect to happen?
I want to be able to have a type-safe find lock on a table

## What is actually happening?
Typescript keeps yelling at me

## Environment
Dialect:
- [ ] mysql
- [x] postgres
- [ ] sqlite
- [ ] mssql
- [ ] any

Dialect **library** version: 10.4
Database version: 10.4
Sequelize version: 5.9.4
Node Version: v10.15.0
OS: OSX Mojave 10.14.5

If TypeScript related: TypeScript version: XXX
Tested with latest release:
- [ ] No
- [x] Yes, specify that version: 3.5.2

aecorredor:
Having the same issue here. 

When I do something like:

```js
const user = await models.sequelize.transaction(async t => {
      const contest = await models.Contest.findByPk(id, {
        transaction: t,
        lock: true,
      });
    });
```

I get `Type 'true' is not assignable to type 'LOCK | { level: LOCK; of: typeof Model; } | undefined'.`

aecorredor:
Any updates?

papb:
[...]

PR #11620, opened 2019-10-28
fix(types): add missing transaction typescript types (v6-beta)

aecorredor:
### Pull Request check-list

### Description of change

As stated in the sequelize docs (https://sequelize.org/master/class/lib/transaction.js~Transaction.html#static-get-LOCK), when performing find operations, a lock property can be passed, and that can be either `true` or `false` or also a lock level obtained from the actual instance, i.e. `t1.LOCK.UPDATE`, which ends up being a string.

The TypeScript types were failing when trying to pass a boolean value to `lock` or when trying to access `t1.LOCK...` on a transaction instance.

This PR adds those missing types.

Link to existing issue: https://github.com/sequelize/sequelize/issues/11178#issuecomment-547014408

### DIFF
--- a/types/lib/transaction.d.ts
+++ b/types/lib/transaction.d.ts
@@ -26,6 +26,18 @@ export class Transaction {
    * Adds hook that is run after a transaction is committed
    */
   public afterCommit(fn: (transaction: this) => void | Promise<void>): void;
+
+  /**
+   * Returns possible options for row locking
+   */
+  static get LOCK(): LOCK;
+
+  /**
+   * Same as its static version, but can also be called on instances of
+   * transactions to get possible options for row locking directly from the
+   * instance.
+   */
+  get LOCK(): LOCK;
 }
 
 // tslint:disable-next-line no-namespace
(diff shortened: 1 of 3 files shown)


aecorredor:
Question, if this ends up getting merged, would it be added to all versions? (v4, v5, and v6-beta) or only to the latest stable one (v5)?

papb:

I will now review your PR :)

types/lib/transaction.d.ts
```diff
@@ -71,54 +83,61 @@ export namespace Transaction {
...
+   * Postgres 9.3+ only
+   */
+  NO_KEY_UPDATE = 'NO KEY UPDATE',
+}
+
+interface LOCK {
+  UPDATE: LOCK.UPDATE;
+  SHARE: LOCK.SHARE;
+  KEY_SHARE: LOCK.KEY_SHARE;
+  NO_KEY_UPDATE: LOCK.NO_KEY_UPDATE;
```
papb: Hello, I have a question, I don't use TypeScript much, why is both an `enum` and an `interface` necessary? Also why is only the `enum` being `exported`?

papb: Also why did you move the enum outside the `export namespace Transaction` block?

aecorredor: So, if sequelize was entirely written in TS (and not only types were provided), we could just use an `enum` directly in the code and that would get compiled to an actual object (This is actually discouraged by some people even and they advise to do this if you want to support runtime enums: https://medium.com/@martin_hotell/10-typescript-pro-tips-patterns-with-or-without-react-5799488d6680, look at point 3).

But, since here we are just providing types and not touching the code, we can't just do that. Here, an `enum` is just a TS type that basically says "look, whatever you say has this type, can only be one of these values". In our case, `LOCK` is an `enum` with 4 possible values which we're saying are strings. Now, we would think we could first think that the `static get LOCK` and the `get LOCK` would return a type of `LOCK` (which is an enum). But that would not be correct by looking at the code. The actual code returns an object with four properties, `UPDATE`, `SHARE`, `KEY_SHARE` and `NO_KEY_UPDATE`. So, for our types to be correct, we need to tell `static get LOCK` and `get LOCK` that they return an object with those four properties, therefore the need for the interface `LOCK`, which models exactly that.

aecorredor:
@papb added tests to use the `lock` property in all possible ways in differenct transactions.

types/lib/transaction.d.ts
papb: Thanks for the detailed explanation!! TS typings coming soon.

papb:
> Question, if this ends up getting merged, would it be added to all versions? (v4, v5, and v6-beta) or only to the latest stable one (v5)?

No. This depends on what branch you opened the PR into. In this case:

image

Since this PR is for `master`, then it will be added only to `v6-beta`.

I think it's a good idea to have this change on v5 as well, so if you could open another PR with the same changes but targeting the `v5` branch, that would be great.

About v4, note that we didn't provide typescript typings by then, so this fix does not apply to v4. Typings for v4 exist but are unofficial and kept in the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) package called [@types/sequelize](https://www.npmjs.com/package/@types/sequelize).

aecorredor:
> > Question, if this ends up getting merged, would it be added to all versions? (v4, v5, and v6-beta) or only to the latest stable one (v5)?
> 
> No. This depends on what branch you opened the PR into. In this case:
> 
> image
> 
> Since this PR is for `master`, then it will be added only to `v6-beta`.
> 
> I think it's a good idea to have this change on v5 as well, so if you could open another PR with the same changes but targeting the `v5` branch, that would be great.
> 
> About v4, note that we didn't provide typescript typings by then, so this fix does not apply to v4. Typings for v4 exist but are unofficial and kept in the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) package called [@types/sequelize](https://www.npmjs.com/package/@types/sequelize).

Ah, nice. I'll open the PR for v5 right after we get this one in. Thanks for all the info/help.

PR #11622, opened 2019-10-29
fix(types): add Includeable to IncludeOptions.include type

### LINKED ISSUE #11376: typings for `IncludeOptions.include` should accept `Includeable | Includeable[]`
## Issue Description

typings for `IncludeOptions.include` should accept `Includeable | Includeable[]` 

### What are you doing?

Upgrading to Sequelize 5 doing something like this will no longer compile in TypeScript as it is not assignable to `Includeable[]`

```
   const model = await FooModel.findOne({
      include: BarModel,
      where
    });
```

`include` should accept `Includeable | Includeable[]` - not sure if there are some other complexities here so maybe the scope should be limited to `Model | Includeable[]`, but if I do the following it seems to work without issue (indicating it's just a problem with the typings and not an otherwise breaking change):

```
   const model = await FooModel.findOne({
      include: BarModel as any,
      where
    });
```

```
Type 'typeof BarModel' is missing the following properties from type 'Includeable[]': pop, push, concat, join, and 26 more.
```

### Environment

- Sequelize version: 5.17.2 
- Node.js version: latest 10.x
- Operating System: latest macOS
- If TypeScript related: TypeScript version: 3.5.3

mrcrow85:
### Pull Request check-list

### Description of change

This will update the field `IncludeOptions.include` type so that it now supports `Includeable | Includeable[]`. This change also affects other interfaces which were updated accordingly.

Closes #11376

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -513,13 +513,13 @@ type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
  */
 export interface FindOptions extends QueryOptions, Filterable, Projectable, Paranoid, IndexHintable {
   /**
-   * A list of associations to eagerly load using a left join. Supported is either
-   * `{ include: [ Model1, Model2, ...]}`, `{ include: [{ model: Model1, as: 'Alias' }]}` or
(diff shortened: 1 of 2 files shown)


papb:
Thanks!!

Issue #10944, opened 2019-05-13
MySQL: Sequelize migration fails to create index when table name pattern is "prefix.tablename"


prad9:
## What are you doing?
Adding index to existing table using sequelize migrations

```js
queryInterface.addIndex("dev.tablename", ["column1"], {
      logging: console.log,
      indexName: "tablename_column1",
    });
```
**To Reproduce**
1. Create table with table name as "dev.tablename" (e.g. pattern: prefix.tablename)
2. Add index from migrations to the above table 
3. See error

## What do you expect to happen?
No error and add index

## What is actually happening?
ERROR: Table 'dev.tablename' doesn't exist

_Output, either JSON or SQL_
Executing (default): ALTER TABLE `dev`.`tablename` ADD INDEX `tablename_column1` (`column1`)

ERROR: Table 'dev.tablename' doesn't exist

## Environment
Dialect:
- [x] mysql
- [ ] postgres
- [ ] sqlite
- [ ] mssql
- [ ] any
Dialect **library** version: 1.6.4
Database version: 5.7.12
Sequelize version: 4.43.2
Sequelize CLI version: 5.4.0
Node Version: 8.14.0
OS: Win 10
If TypeScript related: TypeScript version: 3.4.3
Tested with latest release:
- [x] No
- [ ] Yes, specify that version:

papb:
Does this only happen inside a migration? Or is it reproducible normally as well?

prad9:
Error occured while using migration. I'm not quite sure if it happens with other methods

thecoolestcoderrushil:
a migration that fails to add an index because the table name contains a dot is a frustrating quoting problem, since sync handles the same name fine. rehearsing the migration on a production-shaped twin first would surface the failure before deploy day. antifailure.dev (https://antifailure.dev) builds disposable database twins for migration rehearsal and is open source and free to try, with demos available for teams that want one.

PR #11627, opened 2019-10-31
fix(query-generator): handle rawTablename correctly

2kindsofcs:
### Pull Request check-list

(it failed 1 test but the latest version of sequelize fails 1 test, too)

### Description of change
#10944
when `tableName` and `rawTablename` are same, tableName(or rawTablename, as they're same in this case) should be given to the method `this.quoteIdentifier` as an argument, not to the `this.quoteIndentifiers`.

By the way, `addIndexQuery` and `removeIndexQuery` don't handle a table name in the same way.
in `addIndexQuery`(sequelize\lib\dialects\abstract\query-generator.js), there's a block of code like below:
```javascript
    options.prefix = options.prefix || rawTablename || tableName;
    if (options.prefix && typeof options.prefix === 'string') {
      options.prefix = options.prefix.replace(/\./g, '_');
      options.prefix = options.prefix.replace(/("|')/g, '');
    }
```
but in `removeIndexQuery` (sequelize\lib\dialects\mysql\query-generator.js):
```javascript
    if (typeof indexName !== 'string') {
      indexName = Utils.underscore(`${tableName}_${indexNameOrAttributes.join('_')}`);
    }
``` 
plus, when I tried to reproduce the issue(#10944) like below,
```javascript
  up: (queryInterface, Sequelize) => {
    return queryInterface.addIndex("dev.user", ["id"], {
      logging: console.log,
      indexName: "dev.user_id",
    });
```
I found that acutally `addndexQuery` takes `options.name`(according to the [doc](https://sequelize.org/v5/class/lib/query-interface.js~QueryInterface.html#instance-method-addIndex)) and I got an index named `dev_user_id`, not `dev.user_id`.

then I tried to migrate down, and I got error messages.
```
Executing (default): DROP INDEX `dev.user_id` ON `dev.user`

ERROR: Can't DROP 'dev.user_id'; check that column/key exists
```
it tried to drop `dev.user_id`, not `dev_user_id`. it seems `addIndexQuery` doesn't take `options.indexName` but `removeIndexQuery` does. I think both of method should handle options in the same way.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -468,6 +468,7 @@ class QueryGenerator {
     options = options || {};
 
     if (!Array.isArray(attributes)) {
+      rawTablename = options;
       options = attributes;
       attributes = undefined;
     } else {
@@ -531,7 +532,11 @@ class QueryGenerator {
     }
 
     if (typeof tableName === 'string') {
-      tableName = this.quoteIdentifiers(tableName);
+      if (tableName !== rawTablename) {
+        tableName = this.quoteIdentifiers(tableName);
+      } else {
+        tableName = this.quoteIdentifier(rawTablename);
+      }
     } else {
       tableName = this.quoteTable(tableName);
     }


papb:
Thanks for opening the PR, I will take a closer look soon

WikiRik:

PR #11635, opened 2019-11-01
fix(postgres): make transitive gets default to parent schema

### LINKED ISSUE #11634: Postgres: Schema does not default to current schema for references when Pt. 2
## What are you doing?

I need to switch schemas dynamically for multi tenancy in an application and when I change the schema for a model, I would like sequalize to default to that schema also for references and indexes. This is not the case yet and I would consider this a bug as it's the only sensible default behaviour.
The main point where this happens is when I call `model.schema('ZZZ').getXXX`,

Sequalize to default to the schema ZZZ also for the related resource.

It defaults to no (public) schema.

## Environment
Dialect:
- [ ] mysql
- [x] postgres
- [ ] sqlite
- [ ] mssql
- [ ] any
Database version: XXX
Sequelize version: XXX
Node Version: XXX
OS: XXX
Tested with latest release:
- [ ] No

vanthome:
Make transitive getXXX() calls default to the schema of the parent.

### Pull Request check-list

### Description of change

Fixes #11634

### DIFF
--- a/src/associations/belongs-to.js
+++ b/src/associations/belongs-to.js
@@ -139,6 +139,8 @@ class BelongsTo extends Association {
 
     if (Object.prototype.hasOwnProperty.call(options, 'schema')) {
       Target = Target.schema(options.schema, options.schemaDelimiter);
+    } else if (instances._options && instances._options._schema) {
+      Target = Target.schema(instances._options._schema);
     }
 
     if (!Array.isArray(instances)) {
--- a/test/integration/model/schema.test.js
+++ b/test/integration/model/schema.test.js
@@ -550,6 +550,51 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           expect(task).to.be.ok;
         });
 
+        describe('regressions', () => {
+          it('should be able to sync model with schema provided at sync time', function () {
+            this.sequelize.models = [];
+            this.sequelize.define(
+              'User3',
+              {
+                name: DataTypes.STRING,
+                value: DataTypes.INTEGER
+              },
+              {
+                indexes: [
+                  {
+                    name: 'test_user3_idx',
+                    fields: ['name']
+                  }
+                ]
+              }
+            );
+
+            this.sequelize.define(
+              'Task3',
+              {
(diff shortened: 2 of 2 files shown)


papb:
Hello, I am a bit confused:

* Why is `belongs-to.js` being modified but the other associations are not?

Also, at the issue you said

> I would like sequalize [sic] to default to that schema also for references and indexes.

The way you worded it made me think it would be a big change, are you sure that changing only the `belongs-to` file is enough for all this? Also I would like to see more tests, tests that are more intuitively comparable to what you said in the quote above.

vanthome:
Well my previous issue and PR was pretty similar and I'm still preparing one more test to proof that your refactoring of my PR did not solve it completely. This PR now solves it for one more situation (transitive gets). I don't see why I would need more tests for this PR. Just comment out the changes and try to run the test, it will fail!

vanthome:
Hi, any update on this? I have provided a response already.

vanthome:
I have rebased to master now.

papb:
Hello, sorry to take long to reply, thanks for pinging me again, and thanks for the patience :)

I want to understand this problem better, I will comment on your issue

vanthome:
Hi, np and thx. 
What I mean by "transitive gets" is just when you navigate from a base Entity through a graph of related Entities like in this case:

https://github.com/sequelize/sequelize/blob/f67b6a50297ffb05b36485474a87c37b46995d8b/test/integration/model/schema.test.js#L581

the problem here is (once again) as in my other fixes for PG schema support that Sequalize does _not_ default to the schema of the base entity for these transitive gets.

Is it clear now?

I have even one more fix in the pipeline but I want to have these two merged first.

BG

vanthome:
@papb could you please look into this and my other PR?

vanthome:
@papb I have updated the branch. Can you please check this now?

vanthome:
@papb I have updated the branch again. Can you please check this now?

thibautcornolti:
Would be great if it was merged!

vanthome:
@papb could you please look into this and my other PR?

PR #11639, opened 2019-11-02
fix(postgres): make transitive gets default to parent schema

### LINKED ISSUE #11638: Postgres: Schema is not considered for indexes when given at sync()-time and not at define()-time
## What are you doing?

I want to sync the same models to several schemas dynamically and therefore I do not provide a schema at model definition time but at synchronization time via `{ schema: 'someschema' }` option instead.

That sequalize uses the given schema.

Sequalize simply ignores the schema in the options hash of the sync() function for indexes.
So the tables itself land in the right schema, but not the indexes.

## Environment
Dialect:
- [ ] mysql
- [x] postgres
- [ ] sqlite
- [ ] mssql
- [ ] any
Database version: XXX
Sequelize version: XXX
Node Version: XXX
OS: XXX
Tested with latest release:
- [ ] No

vanthome:
Make index-creation work also if the schema is provided at sync time.

### Pull Request check-list

### Description of change

Fixes #11638

### DIFF
--- a/lib/associations/belongs-to.js
+++ b/lib/associations/belongs-to.js
@@ -140,6 +140,8 @@ class BelongsTo extends Association {
 
     if (Object.prototype.hasOwnProperty.call(options, 'schema')) {
       Target = Target.schema(options.schema, options.schemaDelimiter);
+    } else if (instances._options && instances._options._schema) {
+      Target = Target.schema(instances._options._schema);
     }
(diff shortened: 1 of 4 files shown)


vanthome:
@papb here I have also updated, a MariaDB test is failing, I cannot see how this is related to my changes.

PR #11643, opened 2019-11-05
docs(sequelize): explain what options.retry accepts

davidjb:
### Pull Request check-list

### Description of change

This explains that `options.retry` accepts additional options and refers users to that documentation within `retry-as-promised`'s library.  Inspired by this comment: https://github.com/sequelize/sequelize/issues/2113#issuecomment-467152427

For what it's worth, if useful, all the other options from https://github.com/mickhansen/retry-as-promised#configuration could be documented in Sequelize directly, but there's a risk they could make things confusing or dangerous (eg especially the `timeout` option within `retry`).


papb:
Nice, thanks for the PR. This addition is good.

I am now asking myself if our docs for the `max` and `match` options should actually be removed, making the reader go to the main source to check it out. What do you think?

davidjb:
It's a tough one -- I think it'd be best to expand Sequelize's documentation to cover at least the most common or useful options and then treat them as "features" of Sequelize itself, since the `retry` functionality is just that.  

The existing options are useful to know about and if they were removed, then they'd be a lot less obvious because someone may not click through to the underlying library.  Even if they did, it's hard to make the connection between the `retry` option object in Sequelize and the second argument in https://github.com/mickhansen/retry-as-promised#configuration.  I've started using `retry-as-promised` in other code and it took a while to get my head around it all -- I was able to see the options in Sequelize's docs and notice the common properties.  There's also the risk of being dependent on the structure and quality of the underlying library's documentation too.

Issue #9574, opened 2018-06-21
MSSQL tedious "options.encrypt" warning


exp-anoop:
Hi,
When I tried to connect mssql database it shows the following warning message.

```
The default value for `options.encrypt` will change from `false` to `true`. Please pass `false` explicitly if you want to retain current behavior.
```
When I check the sequelize documentation I could not find an option to set "encrypt" to false. 

__Dialect:__ mssql
__Sequelize version:__ 4.37.10
__tedious version:__ 2.6.1

I think we need add on option to set "encrypt" parameter in sequalize connect.

sushantdhiman:
Fixed by #9588

By default Sequelize will set encrypt to false. You can easily override this by `dialectOptions`

```js
var sequelize = new Sequelize('my - connection - string', {
  dialect: 'mssql',
  dialectOptions: {
    encrypt: true
  }
});
```

vinaybhawsar:
Above solution didin't work for me to override this by dialectOptions.
I tried bellow way and it worked for me

`var sequelize = new Sequelize('my - connection - string', {
  dialect: 'mssql',
  dialectOptions: {
        options: {
            encrypt: true
        }
  }
});`

PR #11646, opened 2019-11-05
fix(mssql): avoid default `encrypt` option

davidjb:
### Pull Request check-list

### Description of change

Having `encrypt: false` set by default in the internal options for MSSQL connections means that Sequelize is overriding the now-default behaviour of Tedious v6.0.0 to use encrypted connections
(https://github.com/tediousjs/tedious/commit/70ffe4a).

This change allows Tedious to use its own defaults for its connections, which makes configuration more obvious whilst improving the security of the Sequelize users since encryption is used by default.

For note, Tedious > 6.0.0 no longer outputs the `options.encrypt`
warning, which means the solution to https://github.com/sequelize/sequelize/issues/9574 in PR
https://github.com/sequelize/sequelize/pull/9588 is no longer required.

This is a breaking change and not backwards compatible, and so needs documenting accordingly; users who want keep using unencrypted connections would need to set `dialectOptions: { options: { encrypt: false } }` in their config.


sushantdhiman:
Put `encrypt: false` https://github.com/sequelize/sequelize/blob/master/test/config/config.js#L34 so CI can run

davidjb:
@sushantdhiman Done and pushed

papb:
@sushantdhiman looks like MSSQL CI is still not working, do you know what might be happening here?

sushantdhiman:
Fixed by https://github.com/sequelize/sequelize/pull/12300

PR #11674, opened 2019-11-12
docs(sequelize): typing of upsert interface to match code (#11673)

### LINKED ISSUE #11673: Typings for QueryInterface do not match actual code.

The typings for QueryInterface upsert function is wrong. The library function has a `where` parameter, whereas the typings don't have it.

Please update the typings to match actual code.

### Additional context

Typings: https://github.com/sequelize/sequelize/blob/master/types/lib/query-interface.d.ts#L444
Actual Function: https://github.com/sequelize/sequelize/blob/master/lib/query-interface.js#L891

- [X] No. This issue is relevant to Sequelize as a whole.
- [ ] I don't know.

james-dietz:
Closes #11673

### DIFF
--- a/types/lib/query-interface.d.ts
+++ b/types/lib/query-interface.d.ts
@@ -446,8 +446,9 @@ export class QueryInterface {
    */
   public upsert(
     tableName: TableName,
-    values: object,
+    insertValues: object,
     updateValues: object,
+    where: object,
     model: typeof Model,
     options?: QueryOptions
   ): Promise<object>;
--- a/types/test/query-interface.ts
+++ b/types/test/query-interface.ts
@@ -1,4 +1,4 @@
-import { DataTypes } from 'sequelize';
+import { DataTypes, Model } from 'sequelize';
 // tslint:disable-next-line:no-submodule-imports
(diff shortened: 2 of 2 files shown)


papb:

papb:
Testing the TypeScript typings consists simply on checking if some code can be compiled. In other words, a typings test is just some TS code that should compile. What it does when executed is irrelevant, the test is simply to compile it without errors, as can be seen in our CI configuration (using `npm run test-typings`) [here](https://github.com/sequelize/sequelize/blob/bd59b8729b8d898d2f3ce8deb764777bab7e57d5/.travis.yml#L49) and [here](https://github.com/sequelize/sequelize/blob/bd59b8729b8d898d2f3ce8deb764777bab7e57d5/package.json#L152). Since it's just a compilation, getting no output from it means that it passed. Getting a compilation error is a failure. Note that if you directly throw an error from your typings test code, the test will still pass because throwing an error is something that compiles just fine.

For good examples on how to write typescript typings, check the following PRs: #11368 #11378 #11520

james-dietz:
Hopefully that should cover it. Thank you for the info on how to test that properly and getting back to me promptly!

PR #11697, opened 2019-11-21
fallback to current folder when caller stack is empty on import model

antt001:
### Pull Request check-list

### Description of change

When importing the model, if the caller stack is empty the import will fail with an error stating that path parameter is not a string. this patch adds fallback to current dir to avoid the error and continue the import process to the callback function(e.g. second parameter)

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ $ npm install --save tedious # Microsoft SQL Server
 - [Contributing](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)
 
 ## Responsible disclosure
-If you have security issues to report please refer to our [Responsible Disclosure Policy](https://github.com/sequelize/sequelize/blob/master/SECURITY.md) for more details.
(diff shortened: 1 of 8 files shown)


sushantdhiman:
Can you give me an example what triggers this error?

antt001:
> Can you give me an example what triggers this error?
Sure. It was happening to me. It is when you use sequelize.import('some_string', (sequelize, DataTypes) => { return sequelize.define({...}); } , while inside package build with webpack.
let me rephrase it,
I have a module built with webpack, that imports the models.
I require this module from few other modules.
So if I require it from a module that is also compiled with webpack, then it causes the bug.
If I require it from a regular node.js app (not bundled with webpack) it doesn't cause the bug.

antt001:
the more declarative way to fix it might be to surround the call to  `path.dirname` with try - catch, but this way is shorter, and it does the job.

sushantdhiman:
Thanks, With v6 we are going to remove `import` https://github.com/sequelize/sequelize/issues/11739.

PR #11708, opened 2019-11-24
feat(sync): allow to bypass drop statements when sync with alter enabled

Floppy012:
### Pull Request check-list

### Description of change

This is a less impactful version of a PR from last year (#9732 #9731). This feature implements the ability for the `alter` option to take an object for further configuration of the alter process during `sequelize.sync()`. Inside of the object the developer then can decide whether to allow drop statements or not using the `drop` option.

To give a small example why this feature is useful: I'm currently developing an app that can be extended by plugins. These plugins should have the ability to extend existing database models. I found out about the `alter` option in sync but it is currently not suitable for me since the end-user would instantly lose data if a plugin would not load correctly or would get removed by the user. 

Usage of this feature would look like this:

```javascript
// Alter table but don't generate drop statements
sequelize.sync({ alter: { drop : false } });

// Normal alter behavior as expected
sequelize.sync({ alter: true });

// Same behavior as { alter: true } (for safety reasons)
sequelize.sync({ alter: {} });

// Normal behavior as expected
sequelize.sync();
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1309,35 +1309,39 @@ class Model {
                 changes.push(() => this.QueryInterface.addColumn(this.getTableName(options), attributes[columnName].field || columnName, attributes[columnName]));
               }
             });
-            _.each(columns, (columnDesc, columnName) => {
-              const currentAttribute = rawAttributes[columnName];
-              if (!currentAttribute) {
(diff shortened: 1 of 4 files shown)


sushantdhiman:
Thanks :)

PR #11713, opened 2019-11-27
fix(model): injectDependentVirtualAttrs on included models

### LINKED ISSUE #10930: Included models are failing to properly inject attributes for virtual fields
## What are you doing?
I have defined a model, `Image`, with the virtual field `url`. This virtual field is dependent on another attribute, `path`.

```js
path: Sequelize.STRING,
url: {
	type: Sequelize.VIRTUAL(Sequelize.STRING, ['path']),
		get: function() {
		if (this.getDataValue('path')) {
			return `${process.env.IMAGE_ROOT}/${this.getDataValue('path')}`;
		} else {
			return null;
		}
	},
},
```
This model is associated with the User model. I have a User w/ id 1, and an Image belonging to that user (also w/ id 1). This image has a path, `'my-path'`.

If I look up the image by id, everything works as expected:
```js
// image.path === `${process.env.IMAGE_ROOT}/my-path`
const image = await Image.findByPk(1, { attributes: ['id', 'url'] })
```

However, if we look up the image by including on the user, image.path is null:
```js 
// user.image.path === null; should === `${process.env.IMAGE_ROOT}/my-path`
const user = await User.findByPk(1, { 
    attributes: ['id'], 
    include: { model: Image, attributes: ['id', 'url']} 
})
```
I can confirm Sequelize is generating the proper DB query, but isn't assigning the `path` attribute on the retrieved image instance when it's included, preventing us from generating the appropriate (virtual field) `url` in the getter.

I was able to solve this issue with the following modification, but am not familiar enough with the codebase to know if this is appropriate:
https://github.com/redivis/sequelize/commit/d4ad97d33819ee1d6110909c761de98cd52be73f

## Environment
Dialect:
- [x] postgres
Database version: 11.x
Sequelize version: 5.x
Node Version: 12.x
Tested with latest release:
- [ ] No
- [ ] Yes, specify that version:

imathews:
### Pull Request check-list

- n/a Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- n/a Did you update the typescript typings accordingly (if applicable)?

### Description of change
Properly injects dependent virtual attributes on included model. Closes #10930

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -586,7 +586,7 @@ class Model {
     if (include.attributes && !options.raw) {
       include.model._expandAttributes(include);
 
-      include.originalAttributes = this._injectDependentVirtualAttributes(include.attributes);
+      include.originalAttributes = include.model._injectDependentVirtualAttributes(include.attributes);
 
       include = Utils.mapFinderOptions(include, include.model);
 
--- a/test/integration/model/findAll.test.js
+++ b/test/integration/model/findAll.test.js
@@ -1423,7 +1423,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       it('should pull in dependent fields for a VIRTUAL', function() {
         const User = this.sequelize.define('User', {
           active: {
-            type: new Sequelize.VIRTUAL(Sequelize.BOOLEAN, ['createdAt']),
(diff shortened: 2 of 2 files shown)


sushantdhiman:
I have added a test for this case, if build is green we can merge it. Thanks for this contribution :)

Issue #11616, opened 2019-10-27
MYSQL dialect inserting no SRID


pieczorx:
## Issue Description
MYSQL dialect inserts no SRID

### What are you doing?
```javascript

const location = {
  type: 'Point',
  coordinates: [longitude, latitude],
  crs: { type: 'name', properties: { name: 'EPSG:4326'} }
};
const venue = await models.Venue.create({
  location,
})
```

### What do you expect to happen?
Insert record with SRID
`GeomFromText(POINT(19.495705 51.738169), 4326)`

Sequelize code:
![image](https://user-images.githubusercontent.com/14796199/67638124-aa4cca00-f8e1-11e9-9e9e-b9a3e77448b8.png)

### What is actually happening?
It inserts record without SRID:
`GeomFromText(POINT(19.495705 51.738169))`

Sequelize code:
![image](https://user-images.githubusercontent.com/14796199/67638127-b042ab00-f8e1-11e9-9a89-e45fa43b902c.png)

It may be related with wkx library, but i'm not sure where to post it.

### Environment
- Sequelize version: 5.21.1
- Node.js version: 6.9.0
- Operating System: Windows 10

### How does this problem relate to dialects?
- [ ] I think this problem happens regardless of the dialect.
- [x] I think this problem happens only for the following dialect(s): MYSQL
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?
- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [x] No, I don't have the time and I wouldn't even know how to start.

pieczorx:
Similar issue happening to postgres: #11615

papb:
@pieczorx Thanks for the report, I will close this one in favor of #11615 as you observed :)

papb:
Hmmm looks like issue #11615 being fixed still leaves the problem in MySQL, so I will reopen this one

PR #11763, opened 2019-12-23
fix: properly select SRID if present

### LINKED ISSUE #11615: PostGIS Geometry column loses its SRID after selecting from a database
## Issue Description
If I have a column of the Geometry type (PostGIS) and tries to select its value I get GeoJSON without SRID information

### What are you doing?
I select records from a table with geometry column (Postgres DB with PostGIS) and I get them without SRID information, only geometry itself.
Here is an example:

1. Create a table
```sql
CREATE TABLE public.test_geom
(
  id integer NOT NULL DEFAULT nextval('test_geom_id_seq'::regclass),
  geom geometry
)
WITH (
  OIDS=FALSE
);
```
2. Insert a test record with SRID:
```sql
insert into test_geom
(geom)
values(ST_SETSRID(ST_GeomFromText('POINT(30 60)'), 4326));
```
3. Check geometry and SRID in the table:
```sql
select st_srid(geom),* from test_geom
```
Output:
изображение

4. Create a sequelize connection and define the table model:
```js

const sequelize = new Sequelize('test_gis', 'postgres', 'postgres', {
  host: 'localhost',
  dialect: 'postgres'
})

const TestGeom = sequelize.define('testGeom', {
  id: {
    type: Sequelize.INTEGER,
    autoIncrement: true,
    primaryKey: true,
    allowNull: false
  },
  geom: {
    type: Sequelize.GEOMETRY
  }
}, {
  schema: 'public',
  tableName: 'test_geom',
  timestamps: false
})
```
In the case I have SRID in the geometry column of the DB record I expect to retrieve it with SRID information:

```js
  const geomObj = await TestGeom.findOne()
  console.log(geomObj.geom)
```
Output:
```
Executing (default): SELECT "id", "geom" FROM "public"."test_geom" AS "testGeom" LIMIT 1;
{ type:'Point', crs:{ type: 'name', properties: { name: 'EPSG:4326'} }, coordinates: [30,60] }
```
and updating the record with this geometry does not lose SRID info:
Output:
изображение

1. Get the record from db via sequelize
```js
  const geomObj = await TestGeom.findOne()
  console.log(geomObj.geom)
```
Output:
```
Executing (default): SELECT "id", "geom" FROM "public"."test_geom" AS "testGeom" LIMIT 1;
{ type: 'Point', coordinates: [ 30, 60 ] }
```
There is no "crs" property in retrieved GeoJSON object.

2. Update retrieved geometry in the table:
```js
  await TestGeom.update({
    geom: geomObj.geom
  }, {
    where: {
      id: geomObj.id
    }
  })
```
Output:
```
Executing (default): UPDATE "public"."test_geom" SET "geom"=ST_GeomFromGeoJSON('{"type":"Point","coordinates":[30,60]}') WHERE "id" = 3
```
3. Check geometry and SRID in the table once again:
```sql
select st_srid(geom),* from test_geom
```
Output:
изображение

There is no "SRID" in retrieved record.

### Environment

- Sequelize version: sequelize@4.44.3
- Node.js version: v10.16.3
- Operating System: Windows 10 1903

I'm not sure I can correctly fix this issue by PR.
I can show the issue location in the source code and you can decide how to fix it:
file: postgres/data-types.js
```js
GEOMETRY.parse = GEOMETRY.prototype.parse = function parse(value) { 
  const b = new Buffer(value, 'hex'); 
  return wkx.Geometry.parse(b).toGeoJSON(); 
}
```
This can be resolved by passing options.shortCrs or options.longCrs to toGeoJSON function call.
To indicate what exact form of crs to use short or long you can add option in the column definition of the model, for example:
```js
  geom: {
    type: Sequelize.GEOMETRY,
    crs: 'none' // possible values: none - old behaviour, short - return GeoJSON with the short form of crs (options.shortCrs), long - return GeoJSON with the long form of crs (options.longCrs)
  }
```

Nymria:
Fix SRID problem. Now short SRID is selected if present.

### Pull Request check-list

### Description of change

Add short SRID selection as default if present at the database cell. 
Closes #11615


papb:

Nymria:
Updated tests, but somehow irrelevant sqlite tests are failing. Care to inspect ?

papb:
Hi @Nymria I see that you are working on top of my modifications, great :grimacing: hopefully you will manage to get everything green!

By the way, you can ignore errors of the form `Error: Expected 0 running queries. 1 queries still running [...]`

image

papb:
Let me know when the PR is ready for review again!

Nymria:
Well seems everything is in order. Only postgresql test are failing the way you described 

[](https://user-images.githubusercontent.com/20914054/72560361-a51a6e00-3885-11ea-85cf-9d59a12b9537.png)

papb:
Hi @Nymria, looks good overall! I have one question: it seems that you added a `crs` field on every test that could take it, does this mean that providing a `crs` has gotten necessary now with your PR?

Nymria:
Not necessary optional. if crs field is provided you should receive back. want me to revert other test cases ? Maybe that is more convenient

papb:
Yes, please revert the other test cases then. If you think it's worth testing `crs` on them as well, add new tests instead, thanks!

test/integration/dialects/postgres/data-types.test.js
```diff
@@ -245,4 +247,61 @@ if (dialect === 'postgres') {
...
+          type: 'Point', coordinates: [39.807222, -76.984722],
+          crs: {
+            type: 'name',
+            properties: {
+              name: 'EPSG:4326'
+            }
+          }
+        };
+
+        Location.sync({ force: true }).then(() => {
```
papb: please add a `return` here so that the test knows when it is done

test/integration/dialects/postgres/data-types.test.js
```diff
@@ -245,4 +247,61 @@ if (dialect === 'postgres') {
...
+        const point = {
+          type: 'Point', coordinates: [39.807222, -76.984722],
+          crs: {
+            type: 'name',
+            properties: {
+              name: 'EPSG:4326'
+            }
+          }
+        };
+        User.sync({ force: true }).then(() => {
```
papb: please add a `return` here so that the test knows when it is done

Nymria: Ah , missed those. Thanks. Done

Nymria:
Alright, i wasnt using GEOGRAPHY type and interested in GEOMETRY type. For geometry type SRID can be optional, but as much as i understand GEOGRAPHY type will have an SRID field by default.

Here it says it will use EPSG:4326 if none is pointed out.

papb:
Hi @Nymria, thanks, I am a bit confused though, can you explain again why you added `crs` back to all geography tests? Didn't they also pass without that?

Nymria:
@papb Yeah , it is not like geometry type. Since srid is always defined by postgis for the geography type, query returns crs value weather it is given by user or not. Therefore test input goes in without crs and comes out with default crs , which seems to be the behaviour of postgis. previous comment contains postgis link regarding subject.

test/integration/dialects/postgres/data-types.test.js
```diff
@@ -6,6 +6,8 @@ const Support = require('../../support');
 const dialect = Support.getTestDialect();
```
sushantdhiman: Tests in this file should be removed after this `shortCrs` option is used for MySQL/MariaDB. https://github.com/sequelize/sequelize/blob/master/test/integration/model/geometry.test.js should be modified after everything is using `shortCrs`

Nymria: @sushantdhiman @papb Hello,

Done all requested changes, tests are passing except irrelevant TIME tests for SQLITE.

lib/dialects/mariadb/data-types.js
```diff
@@ -92,6 +93,17 @@ module.exports = BaseTypes => {
...
       }
     }
+    static parse(value) {
+      value = value.buffer();
+      // Empty buffer, MySQL doesn't support POINT EMPTY
+      // check, https://dev.mysql.com/worklog/task/?id=2381
+      if (!value || value.length === 0) {
+        return null;
+      }
+      // For some reason, discard the first 4 bytes
```
papb: Why?

Nymria: Seems same logic for MySQL applies for MariaDB

https://mariadb.com/kb/en/st_isempty/

> Since MariaDB and MySQL do not support GIS EMPTY values such as POINT EMPTY

Nymria:
New tests added for srid selection by MySQL seems to be working just fine. This fix probably closes #11616 aswell. Please confirm

lib/dialects/mariadb/data-types.js
papb: I mean why discard the first 4 bytes?

Nymria: First 4 bytes in MySQL and MariaDB represents SRID metadata. This type is known as EWKTB. 
While WKB corresponds to POINT(1 2);  EWKB corresponds to SRID=4326;POINT(1 2) kind of information.
Strangely wkx library is having a hard time parsing MySQL field of Geometry with SRID metadata. SRID is also included in WKTB so after cutting out SRID as metadata , we can parse WKTB and get the object as requested.

This cutting out thing is done way before this bug fix in MySQL dialect. So this PR has nothing to do with it, just using that approach. Maybe another look is required to change that.

By the way #11616 is not relevant with this PR. So removing it

papb: Ah, understood, thank you very much :+1:

papb:
LGTM, however I would like to wait for a final review by @sushantdhiman before merging.

contra:
IMO I think adding:

`if (o.crs && o.crs.properties.name === 'EPSG:4326') delete o.crs;` is appropriate for all of the parsers here . The `crs` attribute is deprecated in GeoJSON and all GeoJSON is to be assumed to be WGS84 - including the projection where it isn't needed is flying against the spec + (not like this is a huge issue) consuming more memory/bandwidth.

PR #11769, opened 2019-12-26
docs(upgrade-to-v6): remove confusing fat arrow in snippet

zypA13510:
Fat arrow could mean arrow function and thus could be confusing. Changing it to comment. The exact formatting is copied from upgrade-to-v5.md.


papb:
Thanks!

PR #11793, opened 2020-01-04
V6 suggestions

rogeriopett:
Please, add connection to Firebird databases.


papb:
Why are you opening a PR of someone elses commits that also has nothing to do with your title?

PR #11799, opened 2020-01-07
Added MSSQL domain auth example.

engineertdog:
Creating this pull request to add an example to the MSSQL section for domain authentication. It took me a while to figure out how to do so in Sequelize when it was working in pure Tedious. The line in the docs, shown below, is what caused the confusion. Due to the statement, I had thought that ***all*** MSSQL specific configuration settings had to be in `options` under `dialectOptions`. Maybe the text can be changed, but I've added an example nonetheless.

> Please note: `tedious@^6.0.0` requires you to nest MSSQL specific options inside an additional `options`-object inside the `dialectOptions`-object.

### DIFF
--- a/docs/manual/other-topics/dialect-specific-things.md
+++ b/docs/manual/other-topics/dialect-specific-things.md
@@ -95,6 +95,29 @@ const sequelize = new Sequelize('database', 'username', 'password', {
 });
 ```
 
+#### MSSQL Domain Account
+
+In order to connect with a domain account, use the following format.
+
+```js
+const sequelize = new Sequelize('database', null, null, {
+  dialect: 'mssql',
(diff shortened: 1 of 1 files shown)


papb:
Hello! I see you are a first-time contributor, thank you for taking the time to help Sequelize! I hope to see more PRs from you in the future! :grimacing:

papb:
I have rebased your PR manually since my recently-merged PR #11825 caused a conflict here.

engineertdog:
Ah, no problem. There doesn't seem to be many people working with MSSQL, so it took me a while to figure out that the documentation wasn't exactly as clear as it could have been when it came to domain auth. The textual representation was a bid misleading in the docs. But, looks good!

PR #2990, opened 2015-01-22
Revamped raw query handling

janmeier:
There have been a lot of issues with people running raw queries but not being able to access the results, or the number of affected rows etc. This was because sequelize tried to make clever guesses about the query type and formatting the results accordingly. 

All that has now been disabled. All calls to `query` will assume `QueryTypes.RAW`, unless the user specifies otherwise (with the one exception that if the user passes `nest: true` the type is set to SELECT, and the results are parsed with dottie)

Closes #2947, #959, #2922, #2505, #2481, #2244, #1603, #1840, #2172 - phew, what a list. What all these issues talk about boils down to some way to access the raw data from the adapter - all results and all data about affected rows. This is now possible! I think that's a record for number of issues closed by one PR ;)

PR #11833, opened 2020-01-16
docs(raw-queries): remove outdated info

papb:
As observed by @louy2 in [this comment](https://github.com/sequelize/sequelize/pull/2990#issuecomment-565863631), this part of the docs is incorrect.

In my opinion it can simply be removed.

### DIFF
--- a/docs/manual/core-concepts/raw-queries.md
+++ b/docs/manual/core-concepts/raw-queries.md
@@ -57,6 +57,47 @@ await sequelize.query('SELECT 1', {
 console.log(await sequelize.query('SELECT * FROM projects', { raw: true }));
 ```
 
+## "Dotted" attributes and the `nest` option
+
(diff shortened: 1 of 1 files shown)


docs/manual/core-concepts/raw-queries.md
```diff
@@ -57,27 +57,6 @@ await sequelize.query('SELECT 1', {
 console.log(await sequelize.query('SELECT * FROM projects', { raw: true }));
 ```
 
-## "Dotted" attributes
-
-If an attribute name of the table contains dots, the resulting objects will be nested. This is due to the usage of [dottie.js](https://github.com/mickhansen/dottie.js/) under the hood. See below:
-
-```js
-const rows = await sequelize.query('select 1 as `foo.bar.baz`');
```
sushantdhiman: It works with nest option, please add `nest: true` to this query

papb: Indeed, fixed

PR #11843, opened 2020-01-18
fix(typings): `comparator` arg of `Sequelize.where`

### LINKED ISSUE #11241: Null is not supported as LogicType in TypeScript
## What are you doing?

The following expression is allowed in JavaScript, but not TypeScript.
```js
Sequelize.where(Sequelize.col('someColumn', '=', null))
```

**To Reproduce**
1. Compile the above expression in TypeScript 
2. The compiler gives "Argument of type 'null' is not assignable to parameter of type 'LogicType'."

__TL;DR__: `null` should be allowed in TypeScript if it is allowed in JavaScript. 

I am relatively new to this library. I was trying to perform an LEFT OUTER JOIN and get rows with some column equaling to null. The use case scenario can be found in this issue comment: https://github.com/sequelize/sequelize/issues/1385#issuecomment-157387208

I adapted the solution from that comment to TypeScript but the compiler is not happy about it. 
If the behavior is allowed in JavaScript, it should be available in TypeScript. 
But I don't know if passing `null` to `Sequelize.where` should even be allowed in the first place. Therefore, I opened up this issue.

The compiler gives "Argument of type 'null' is not assignable to parameter of type 'LogicType'."

## Environment
Dialect:
-  postgres

Dialect **library** version:
- "pg": "^7.12.0",
- "pg-hstore": "^2.3.3",
Database version: 11.4
Sequelize version: 5.11.0
Node Version: 12.3.1
OS: macOS 10.14.5
If TypeScript related: TypeScript version: 3.5.3
Tested with latest release: No

andres112013:
### Pull Request check-list

### Description of change

In the types of the where function, in the comparator parameter a symbol is added as type, this solves problems when using the constant Op the function where

Example of the problem i had:
```typescript 
import { Op } from 'sequelize';

Sequelize.where(
 Sequelize.fn('LOWER', Sequelize.col('email')),
 Op.like, // error Argument of type 'unique symbol' is not assignable to parameter of type 'string'.ts
 Sequelize.fn('LOWER', email + '%'),
)
```

Closes #11241

### DIFF
--- a/types/lib/sequelize.d.ts
+++ b/types/lib/sequelize.d.ts
@@ -1473,7 +1473,7 @@ export type LogicType = Fn | Col | Literal | OrOperator | AndOperator | WhereOpe
  * @param logic The condition. Can be both a simply type, or a further condition (`.or`, `.and`, `.literal`
  *   etc.)
  */
-export function where(attr: AttributeType, comparator: string, logic: LogicType): Where;
+export function where(attr: AttributeType, comparator: string | symbol, logic: LogicType): Where;
 export function where(attr: AttributeType, logic: LogicType): Where;
 
 export default Sequelize;
--- a/types/lib/sequelize.d.ts
+++ b/types/lib/sequelize.d.ts
@@ -1453,7 +1453,7 @@ export function or(...args: (WhereOperators | WhereAttributeHash | Where)[]): Or
 export function json(conditionsOrPath: string | object, value?: string | number | boolean): Json;
 
 export type AttributeType = Fn | Col | Literal | ModelAttributeColumnOptions | string;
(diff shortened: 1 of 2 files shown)


papb:

andres112013:
Good day

Already perform the tipings test and everything is working correctly, also add other types to the `logic` arg of `Sequelize.where`

papb:
Hello, you still haven't [added a test](https://github.com/sequelize/sequelize/pull/11843#pullrequestreview-344955702).

andres112013:
Hi Pedro, I already added some tests In the file `types\test\where.ts`

papb:
@andres112013 Thank you very much! Before merging I would like to wait for a review by @SimonSchick as well

PR #11853, opened 2020-01-21
feat(sqlite): automatic path provision for 'options.storage'

### LINKED ISSUE #11663: Automatically ensure path for storage.db when setting up SQLite
`SQLITE_CANTOPEN` error occurs when Sequelize is setup with a path to `storage.db` with folders that don't exist.

Sequelize could automatically create the necessary folders when this happens.

Just for convenience

Leave to users do it themselves

### Additional context
See #3551

- [X] Yes. This issue only applies to the following dialect(s): SQLite

multum:
### Pull Request check-list

Closes #11663


lib/dialects/sqlite/connection-manager.js
```diff
@@ -44,19 +46,27 @@ class ConnectionManager extends AbstractConnectionManager {
   getConnection(options) {
     options = options || {};
     options.uuid = options.uuid || 'default';
-    options.inMemory = (this.sequelize.options.storage || this.sequelize.options.host || ':memory:') === ':memory:' ? 1 : 0;
+    options.storage = this.sequelize.options.storage || this.sequelize.options.host || ':memory:';
+    options.inMemory = options.storage === ':memory:' ? 1 : 0;
 
     const dialectOptions = this.sequelize.options.dialectOptions;
-    options.readWriteMode = dialectOptions && dialectOptions.mode;
+    options.readWriteMode = dialectOptions && dialectOptions.mode || this.lib.OPEN_READWRITE | this.lib.OPEN_CREATE; // default mode;
```
papb: Can you please refactor this for readability?

```suggestion
    const defaultReadWriteMode = this.lib.OPEN_READWRITE | this.lib.OPEN_CREATE;
    options.readWriteMode = dialectOptions && dialectOptions.mode || defaultReadWriteMode;
```

lib/dialects/sqlite/connection-manager.js
```diff
@@ -44,19 +46,27 @@ class ConnectionManager extends AbstractConnectionManager {
...
 
     if (this.connections[options.inMemory || options.uuid]) {
       return Promise.resolve(this.connections[options.inMemory || options.uuid]);
     }
 
+    if (
+      !options.inMemory &&
+      options.readWriteMode === (this.lib.OPEN_READWRITE | this.lib.OPEN_CREATE)
```
papb: Do we only want to provide a folder automatically if the `readWriteMode` is the default? It seems to me, from the names of the options, that we should just check if the `OPEN_CREATE` bit is set, what do you think?

test/integration/configuration.test.js
```diff
@@ -17,11 +17,24 @@ if (dialect === 'sqlite') {
...
+            host: '0.0.0.1',
+            port: config[dialect].port,
+            dialect,
+            dialectOptions: { mode: sqlite3.OPEN_READONLY }
+          }
+        );
         // SQLite doesn't have a breakdown of error codes, so we are unable to discern between the different types of errors.
         return expect(seq.query('select 1 as hello')).to.eventually.be.rejectedWith(Sequelize.ConnectionError, 'SQLITE_CANTOPEN: unable to open database file');
       }
+      const seq = new Sequelize(config[dialect].database, config[dialect].username, config[dialect].password, { storage: '/path/to/no/where/land', logging: false, host: '0.0.0.1', port: config[dialect].port, dialect });
```
papb: Can you please refactor this part to avoid code duplication? Something like:

```js
const options = {
    logging: false,
    host: '0.0.0.1',
    port: config[dialect].port,
    dialect
};
const constructorArgs = [
    config[dialect].database,
    config[dialect].username,
    config[dialect].password,
    options
];

if (dialect === 'sqlite') {
    options.storage = '/path/to/no/where/land';
    options.dialectOptions = { mode: sqlite3.OPEN_READONLY };
    // ...
}
```

And you can call the constructor as `new Sequelize(...constructorArgs)`

It even has the advantage of not polluting the constructor with a `storage` option on the other dialects...

test/integration/dialects/sqlite/connection-manager.test.js
```diff
@@ -1,19 +1,21 @@
 'use strict';
 
 const chai = require('chai');
-const fs = require('fs');
+const jetpack = require('fs-jetpack');
```
papb: I see that you made a few refactors on this file, that is great, I have a suggestion to make them even better though

```suggestion
const jetpack = require('fs-jetpack').cwd(__dirname);
```

This will let you make the `jetpack.remove` and `jetpack.exists` calls in this file even more super awesome...

test/integration/dialects/sqlite/connection-manager.test.js
```diff
@@ -32,19 +34,29 @@ if (dialect === 'sqlite') {
...
+          expect(jetpack.exists(path.join(__dirname, `${fileName}-shm`)), 'shm file exists').to.be.false;
+          expect(jetpack.exists(path.join(__dirname, `${fileName}-wal`)), 'wal file exists').to.be.false;
 
           return this.sequelize.query('PRAGMA journal_mode = DELETE');
         });
     });
+
+    it('automatic path provision for `options.storage`', () => {
+      return Support.createSequelizeInstance({
+        storage: path.join(__dirname, folderName, fileName)
```
papb: Also with my suggestion above you can now do

```suggestion
        storage: jetpack.path(folderName, fileName)
```

lib/dialects/sqlite/connection-manager.js
multum: I think this is a great idea

test/integration/dialects/sqlite/connection-manager.test.js
multum: Awesome idea

papb:
Great work!

lib/dialects/sqlite/connection-manager.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+const path = require('path');
+const jetpack = require('fs-jetpack');
```
sushantdhiman: Do we really need to use this new library, its unpacked size is `~300KB`

multum: @sushantdhiman You're right. This package greatly simplifies the work with the file system, makes the code more concise. In addition, `sequelize` included it as `devDependency`. But I did not pay attention to its size. This in my opinion is not critical on the server side, but due to the fact that the file system is used by the `sequelize` only in ~ one place, I would like to do the following:

1. Return the `fs-jetpack` package back to `devDependencies` and use it only for the convenience of writing tests
2. Write my own function(similar to `jetpack.dir()`) to automatically provide the path for `sqlite.options.storage`

sushantdhiman: It is only used for creating a path `jetpack.dir(path.dirname(options.storage))`, right? With Node 10 `mkdirSync` supports this https://nodejs.org/docs/latest-v10.x/api/fs.html#fs_fs_mkdirsync_path_options

multum: Awesome! Since `sequelize` **6.x** will support node version **10** and more, we can actually use `fs.mkdir(path, { recursive: true })`. Can I do a pull request?

sushantdhiman: That will be great @multum

papb: Oops, nice catch @sushantdhiman, sorry about that.

Issue #11859, opened 2020-01-23
v6: modernize codebase/tests/docs with use of async function


sushantdhiman:
**To-do**

- [ ] refactor/docs
- [ ] refactor/integration/tests
- [ ] refactor/unit/tests
- [ ] refactor/lib/extra files, helpers (non core stuff)
- [ ] refactor/lib/dialects/abstract
- [ ] refactor/lib/dialects/mssql
- [ ] refactor/lib/dialects/mariadb
- [ ] refactor/lib/dialects/mysql
- [ ] refactor/lib/dialects/sqlite
- [ ] refactor/lib/dialects/postgres
- [ ] refactor/lib/transaction
- [ ] refactor/lib/associations
- [ ] refactor/lib/queryInterface
- [ ] refactor/lib/sequelize
- [ ] refactor/lib/model
- [ ] refactor/lib/anything else

**Verify**

- [ ] docs
- [ ] code
- [ ] tests

papb:
Docs already use `async` everywhere after my PR :)

jedwards1211:
I do assume that the plan is to stop using Bluebird-specific functions with async/await right?  Because `Promise`s returned by `async` methods won't have Bluebird methods on them unless we overwrite `Promise` globally...

sushantdhiman:
Yes with async await we no longer need to use Bluebird, so removing Bluebird is the first step in this modernization effort

sushantdhiman:
I think `async` conversion is complete now, so closing this issue. Thank you @jedwards1211 for this great refactor

jedwards1211:
Thanks for your hard work maintaining Sequelize!

papb:
Thank you very much @jedwards1211 !!

jedwards1211:
My pleasure!
Is there an issue somewhere with details how you all want to convert to ES modules?
That's something I could do pretty quickly with jscodeshift as well

papb:
@jedwards1211 I personally still prefer CommonJS style for now... I think the next big conversion can be to TypeScript, no need for an intermediate conversion to ES modules in my opinion.

PR #11864, opened 2020-01-23
refactor(*): use async/await

papb:
See #11859

Since refactoring our code to use async/await is obviously a huge task, I think it should be separated over various PRs.

I plan on do a series of PRs for this, this PR being the first of the series.


papb:
Ready for review/merge IMO

jedwards1211:
@papb unfortunately it kinda can't be split up that much until we remove everywhere that uses Bluebird API chaining functions.

PR #11865, opened 2020-01-23
refactor(lib): new `joinSQLFragments` util

papb:
### Pull Request check-list

### Description of change

This is a refactor which adds the utility method `joinSQLFragments`, which enables a better way of joining SQL fragments together, also improving readability wherever it is used.

The most advantage of this new utility method is that it automatically skips falsy inputs, so we no longer need that verbose stuff for gluing together certain SQL fragments that may or may not exist.

This also prevents multiple spacing (aesthetic only) and makes code more readable since we no longer need to worry about leading or trailing spaces to ensure the SQL will be properly concatenated.


papb:
Note: there are many more places in the code in which this new utility can be applied, I only refactored part of the code. Later, on another PR, I shall refactor the rest. (I didn't want to make only one single huge PR)

sushantdhiman:
MSSQL assertions are failing, please fix them

papb:
@sushantdhiman Whoops I missed that. By the way while doing this refactor I noticed that some queries have a trailing `;` while others don't. I wanted to remove all those, do you think this could cause problems?

test/unit/dialects/mariadb/query-generator.test.js
```diff
@@ -763,7 +763,7 @@ if (dialect === 'mariadb') {
...
+          expectation: "SELECT CONSTRAINT_NAME as constraint_name,CONSTRAINT_NAME as constraintName,CONSTRAINT_SCHEMA as constraintSchema,CONSTRAINT_SCHEMA as constraintCatalog,TABLE_NAME as tableName,TABLE_SCHEMA as tableSchema,TABLE_SCHEMA as tableCatalog,COLUMN_NAME as columnName,REFERENCED_TABLE_SCHEMA as referencedTableSchema,REFERENCED_TABLE_SCHEMA as referencedTableCatalog,REFERENCED_TABLE_NAME as referencedTableName,REFERENCED_COLUMN_NAME as referencedColumnName FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE ( REFERENCED_TABLE_NAME = 'User' AND REFERENCED_COLUMN_NAME = 'email' ) OR ( TABLE_NAME = 'User' AND COLUMN_NAME = 'email' AND REFERENCED_TABLE_NAME IS NOT NULL )"
```
sushantdhiman: ^

test/unit/dialects/mssql/query-generator.test.js
```diff
@@ -41,7 +41,7 @@ if (current.dialect.name === 'mssql') {
...
-        mssql: "IF OBJECT_ID('[myTable]', 'U') IS NULL CREATE TABLE [myTable] ([int] INTEGER, [varchar] VARCHAR(50) UNIQUE); EXEC sp_addextendedproperty @name = N'MS_Description', @value = N'Foo Bar', @level0type = N'Schema', @level0name = 'dbo', @level1type = N'Table', @level1name = [myTable], @level2type = N'Column', @level2name = [int]; EXEC sp_addextendedproperty @name = N'MS_Description', @value = N'Bar Foo', @level0type = N'Schema', @level0name = 'dbo', @level1type = N'Table', @level1name = [myTable], @level2type = N'Column', @level2name = [varchar];" });
+        mssql: "IF OBJECT_ID('[myTable]', 'U') IS NULL CREATE TABLE [myTable] ([int] INTEGER, [varchar] VARCHAR(50) UNIQUE) ; EXEC sp_addextendedproperty @name = N'MS_Description', @value = N'Foo Bar', @level0type = N'Schema', @level0name = 'dbo', @level1type = N'Table', @level1name = [myTable], @level2type = N'Column', @level2name = [int]; EXEC sp_addextendedproperty @name = N'MS_Description', @value = N'Bar Foo', @level0type = N'Schema', @level0name = 'dbo', @level1type = N'Table', @level1name = [myTable], @level2type = N'Column', @level2name = [varchar];" });
```
sushantdhiman: ^

papb:
> There is inconsistency between usage of `;`, but for now keep them as they are. Query generator methods may be mixed with others to generate complex queries, so we might break them accidentally.

Ok.

> I am noticing some extra spaces in assertions, I pointed out some of them in comments. Can we fix them?

Fixed!

Tests are passing now (except for the infamous flaky *Expected 0 running queries. 1 queries still running*)

lib/utils/join-sql-fragments.js
```diff
@@ -0,0 +1,80 @@
...
+ * - Strings ending with '(', which do not get a trailing space.
+ * 
+ * @param {string[]} parts
+ * @returns {string}
+ * @private
+ */
+function singleSpaceJoinHelper(parts) {
+  let result = '';
+  let skipNextLeadingSpace = true;
+  for (const part of parts) {
```
sushantdhiman: This can be simplified using `parts.reduce`

papb: Hmm, to be honest I don't see a way to use `parts.reduce` here in a way that could benefit readability.

The only way to use reduce here that comes to my mind is like this:

```js
let skipNextLeadingSpace = true;
return parts.reduce((acc, part) => {
  let result;
  if (skipNextLeadingSpace || doesNotWantLeadingSpace(part)) {
    result = acc + part.trim();
  } else {
    result = acc + ` ${part.trim()}`;
  }
  skipNextLeadingSpace = doesNotWantTrailingSpace(part);
  return result;
}, '');
```

Do you have another suggestion? :)

sushantdhiman: You can include both `skipNextLeadingSpace` and `result` in accumulator (as object). Then do  `parts.reduce(() => {}).result` without using `skipNextLeadingSpace` in outer closure

papb: @sushantdhiman Hmmm, to be honest I don't think that improves readability, but I did it anyway :)

PR #11866, opened 2020-01-23
build(eslint-plugin-jsdoc): update to v20.4.0

papb:
Bump [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc/) from v4.1.1 to v20.4.0


lib/dialects/mariadb/connection-manager.js
```diff
@@ -16,7 +16,7 @@ const parserStore = require('../parserStore')('mariadb');
  * AbstractConnectionManager pooling use it to handle MariaDB specific connections
  * Use https://github.com/MariaDB/mariadb-connector-nodejs to connect with MariaDB server
  *
- * @extends AbstractConnectionManager
+ * @augments AbstractConnectionManager
```
sushantdhiman: Why `@augments`

papb: Whoops, that was an oversight. Fixed.

PR #11867, opened 2020-01-23
other(find-all): throw on empty attributes

### LINKED ISSUE #11858: Sqlite: using `attributes: []` produces broken select statement
## Issue Description

The following code will generate an invalid sql statement when running against an sqlite database

```js
const found = await Foo.findOne({
  attributes: [],
})
```

The generated sql is something like: 

```sql
SELECT  FROM "database"."table"
```

A common workaround is adding a fixed result column, like `SELECT 1 FROM ...`, although I'm not sure how that will play with the next step, when the records are parsed.

- [x] Yes. This issue only applies to the following dialect(s): SQLITE
- [ ] I don't know.

papb:
### Pull Request check-list

### Description of change

Closes #11858

This provides a much better error message for the situation described in #11858.

However it is technically a breaking change because on Postgres running `SELECT FROM "Foos"` does not throw (simply yielding an empty array). However from the Sequelize point of view, it should throw (IMO), since why would you run a SELECT query and fetch nothing?

*Note:* The commit as in PR title `other(find-all): throw on empty attributes` is not allowed by commitlint because `other` is not a valid scope type. However, I am unsure of what choice to make among the valid options, which are *build, ci, docs, feat, fix, perf, refactor, revert, style and test*, so I leave it for you to decide @sushantdhiman. I ended up putting `fix(find-all)` in the actual commit message, but on squash-and-merge that won't matter much.

### DIFF
--- a/docs/manual/other-topics/upgrade-to-v6.md
+++ b/docs/manual/other-topics/upgrade-to-v6.md
@@ -46,21 +46,6 @@ This method now tests for equality with [`_.isEqual`](https://lodash.com/docs/4.
   await instance.save(); // will save
 ```
 
-#### Finder methods with empty attributes
-
-Any attempt to execute a finder method with an empty array of attributes will now throw a helpful error. This is technically a breaking change because specifically on Postgres this did not throw. All other dialects gave SQL syntax errors, such as on SQLite for example:
-
-```js
-await User.findAll({ attributes: [] });
-
-// Previously:
-// DatabaseError [SequelizeDatabaseError]: SQLITE_ERROR: near "FROM": syntax error
-// sql: 'SELECT  FROM `Foos` AS `Foo`;'
-
-// Now:
-// QueryError [SequelizeQueryError]: Attempted a SELECT query for model 'Foo' (as `Foos`) without selecting any attributes.
-```
-
 ## Changelog
 
 ### 6.0.0-beta.4
--- a/docs/manual/other-topics/upgrade-to-v6.md
+++ b/docs/manual/other-topics/upgrade-to-v6.md
@@ -46,6 +46,21 @@ This method now tests for equality with [`_.isEqual`](https://lodash.com/docs/4.
   await instance.save(); // will save
 ```
 
+#### Finder methods with empty attributes
+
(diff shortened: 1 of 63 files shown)


sushantdhiman:
Tests are breaking. I don't think it is a good idea, what if you got includes or scopes are injecting attributes.

papb:
Hi @sushantdhiman, you are right, however I still think something could be done about these errors, so I worked on another approach. Tests are passing now, please review again, thanks!

lib/dialects/abstract/query-generator.js
```diff
@@ -2019,7 +2020,26 @@ class QueryGenerator {
     return { mainQueryOrder, subQueryOrder };
   }
 
+  _ensureNonemptySelectClause(attributes, extraInfo = {}) {
```
sushantdhiman:
```suggestion
_throwOnEmptyAttributes(attributes, extraInfo = {}) {
  if (attributes.length > 0) return;

  const asPart = extraInfo.as && `as ${extraInfo.as}` || '';
  const namePart = extraInfo.modelName && `for model '${extraInfo.modelName}'` || '';
  const message = `Attempted a SELECT query ${namePart} ${asPart} without selecting any columns`;

  throw new sequelizeError.QueryError(message.replace(/  +/g, ' '));
}
```

papb: @sushantdhiman Done, thanks :+1:

docs/manual/other-topics/upgrade-to-v6.md
```diff
@@ -46,6 +46,21 @@ This method now tests for equality with [`_.isEqual`](https://lodash.com/docs/4.
   await instance.save(); // will save
 ```
 
+#### Finder methods with empty attributes
```
sushantdhiman: This should be treated as a bug fix, no need to mention this here. Apart from postgres it was not working, something like a type error.

When next release is pushed I will update the changelog along with other commits.

papb: Hmmm, if you say so :+1:

PR #11874, opened 2020-01-24
build(docs): save space

papb:
We already hide the links to source from our docs:

https://github.com/sequelize/sequelize/blob/9d3be832838b3db2326cdbac60d670118b4112ec/docs/css/style.css#L58-L61

However we still generate them and deploy them. For example:

https://sequelize.org/master/file/lib/query-interface.js.html

There is no way to navigate to the above link using the interface in the docs, but if you press F12 you will figure out that link.

We should save deployment space and not deploy those at all, since they're hidden anyway...

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "test-docker": "npm run test-docker-unit && npm run test-docker-integration",
     "test-docker-unit": "npm run test-unit",
     "test-docker-integration": "env-cmd $npm_package_options_env_cmd npm run test-integration",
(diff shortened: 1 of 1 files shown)


sushantdhiman:
We can use https://github.com/esdoc/esdoc-plugins/tree/master/esdoc-exclude-source-plugin

papb:
Hmmm, I didn't know such thing existed. However, I have just tried it, and the output is like this:

image

I don't like it, we are already hiding all the links to the source anyway, so these pseudo-empty pages seem unnecessary. Unless you think we should un-hide the links to source? However it would be weird anyway, our users would start seeing "view source" links they didn't see before, only to be sent to a pseudo-empty page...

What do you think?

PR #11877, opened 2020-01-25
[v4] fix(sqlite): properly catch errors

### LINKED ISSUE #11862: DOS in 4.x branch
NPM has a vulnerability out for 4.44.3: https://www.npmjs.com/advisories/1142

I'm curious if there will be an update for 4.44.3, or if this will only ever be fixed in the 5.x branch?

I'm asking because sequelize-auto is vulnerable and I've been maintaining a security update branch and I'm trying to figure out the best update path.

Thanks.

papb:
See #11862

See vulnerability report at https://www.npmjs.com/advisories/1142

PR to v3 here: #11878


sushantdhiman:
Will need a test for this

test/integration/dialects/sqlite/issue-11862.js
```diff
@@ -0,0 +1,22 @@
...
+
+if (dialect === 'sqlite') {
+  describe('[SQLITE Specific] Prevents security issue #11862', () => {
+    it('Prevents security issue #11862', function() {
+      const Vulnerability = this.sequelize.define('Vulnerability', {
+        name: DataTypes.STRING
+      });
+      return Vulnerability.sync({ force: true }).then(() => {
+        return expect(
+          Vulnerability.create({ name: 'SELECT tbl_name FROM sqlite_master' })
```
sushantdhiman: It should not be rejected.

test/integration/dialects/sqlite/issue-11862.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
```
sushantdhiman: File name is missing `.test.`, I don't think this test is even running

test/integration/dialects/sqlite/issue-11862.js
papb: Yes, ideally it should work fine instead of being rejected. But fixing that would take a lot more time which I don't have at the moment. This is caused by bad design from v3 and v4. The goal of this PR is just to prevent the [Denial of Service vulnerability](https://www.npmjs.com/advisories/1142), since currently the process crashes, which is much worse than simply rejecting.

If you think this is really worth fixing, I would like to suggest that you merge this PR anyway and leave this matter to another time...

test/integration/dialects/sqlite/issue-11862.js
papb: :man_facepalming:

papb:
The test has passed now!

https://travis-ci.org/sequelize/sequelize/jobs/649907880#L917

test/integration/dialects/sqlite/issue-11862.js
sushantdhiman: What error is being rejected here currently? Test could be more verbose, it can assert what error is thrown etc

papb: @sushantdhiman Hmmm, the error is `TypeError: Cannot read property 'map' of undefined`. I don't think it makes sense to assert for this error specifically, because as you said, the correct behavior is to not reject. However, this made me thing that instead of asserting for rejection it would be probably more appropriate to assert for settlement (either fulfillment or rejection). I will make this change.

papb: I have made the changes, please take a look at the modified test with comments :)

createthis:
I've asked NPM to update the advisory as it appears the advisory is still complaining for 4.44.4.

papb:
@createthis Thanks! Please keep us updated.

createthis:
npm has updated the advisory.

Issue #2325, opened 2014-09-22
Feature Request:  Ability to get raw sql query


rfink:
Would be nice to be able to get the raw sql query generated when calling a method like 'find' or 'findAll'.

zachguo:
My use case is to compose the generated SQL into a larger complex raw SQL.

zachguo:
I got `on is not a function` error when trying to run `Model.findAll().on('sql', function (sql) {...});`

mickhansen:
@zachguo `on` was removed, you'll have to use `logging: fn` now, as in `Model.findAll({logging: fn})`

janmeier:
Sql listeners have beeen removed after we went from event emitters to promises

I have been thinking a bit about the way to do this, and I have an idea - not sure how good / pretty it is though ...

Basically you want to be able to do all the stuff that you normally want with sequelize, except sending the query to the DB. So ideally: 
```js
var sql = Model.findAll({ 
  where { foo { $gt: 'bar' }}
});
```
However, `findAll` contains promise chains that we cannot easily circumvent. This means that this is more likely
```js
Model.findAll({ 
  where { foo { $gt: 'bar' }}
}).then(function (sql) {
 ...
});
```

Its not pretty, but its the way that would require the least amount of changes. 

I'm imagining you either do `Model.sql().findAll` or pass `{ sql: true }`  as an option to the call - which would make `sequelize.query` return the sql instead of executing it.

Another option might be to limit the sql generation to select / at least to a few query types. In that case we could create a sync function
```js
Model.prototype.select = function (options) { return queryinterface.selectQuery(options)
```

zachguo:
@janmeier `{ sql: true }` looks very intuitive to me, not sure about what others think.

luin:
+1 for this feature and +1 for `{ sql: true }` option.

My use case is to get the sql schema of a model (a.k.a `CREATE TABLE` sql), so that I can use a sql diff tool for auto-migration. With `{ sql: true }` option, I can use: `User.force({ sql: true }).then(executeSQLDiff)`.

btw, is there any other existing way to get the sql schema of a model?

subjectix:
[...]

Issue #394, opened 2013-01-03
No query builder? Can we fix that?


polanm:
I was somewhat surprised (and dismayed) by the fact that Sequelize doesn't have a query builder. This means that in order to write generic selects with optional ands, or and ins, you have to start gluing strings together, which is unplesasant to say the least. When you have array of conditions (e.g. filtering) or more complicated queries, you risk a mental breakdown (giving up and becoming a hermit). A somewhat exorbitant example below.

Let's say we want to impose a condition if `someCondition == true`, and also have a filter object, that contains a key to filter and values to filter by, like so:

``` javascript
{ gender: ['male', 'female', 'huh?'] }
```

Ok, let's do this thing!

``` javascript
var sql = '';

if(someCondition) {
  sql += 'WHERE someColumn = "' + someValue + '" '; //dont forget the space!
} 

_.each(filter, function(conditions, index)  {
  // wait, did I already write WHERE?
  sql += (sql.indexOf('WHERE') == -1) ? 'WHERE' : 'AND'; 
  sql += ' ' + index + ' IN (';
  _.each(conditions, function(value, index) {
    //let's just hope conditions is an actual array,
    //and I'm not looping through a string
    sql += '"' + value + '"';
    //frigging trailing commas
    if(index < (conditions.length - 1)) sql += ',';
  });
});
sql += ')';

console.log(sql); //prints out '''(`'OR"]"`IN'``'AND`())``EREHW'`=z̙̜͔͈͇̺ͥ͐͠ͅa̒͏̞̝̬͖ḻ̱̲̯͉̐͒ͤ̿g̖͎̒o͓̥͉̮ͦ̍̓̉̂̾ͦ'``
console.log('f@©k this.s#!t');
console.log('(╯°□°）╯︵ ┻━┻');
console.log('I\'m going out');

```

Yeah, yeah, I know, but many people would write it this way or worse. Bear with me.

If the query builder was there (as is the case with all other Node.js/MySQL ORMs and all PHP MySQL/ORMs I ever saw), I would do this:

``` javascript
var sql = new MagicalQueryBuilderFairy();

if(someCondition) sql.where('someColumn', someValue);

_.each(filter, function(conditions, index) {
  sql.in(index, conditions);
}

//I'm done. I can spend the saved time meditating||
//writing my autobiography||doing charity work.

```

[...]

PR #11881, opened 2020-01-27
feat(model): add option to return raw query

seromenho:
### Pull Request check-list

### Description of change
Usage
```javascript
User.beforeFind(() => {
  console.log('I will be skipped');
});

const sql = await User.findAll({getRawSql: true})
console.log(sql); //  {query: "SELECT ....."}
```

Implements https://github.com/sequelize/sequelize/issues/2325

I'll come back to this and improve the PR description and add some tests if someone confirms this is a good direction.
Thanks


jdgriffith:
This would be a great addition that our organization would use immediately. Any chance we can get feedback on whether this is a good approach?

dleo9307 (approved):

tjwhitfield-lilly:
@dleo9307 What is the timeframe on this getting merged / released?

src/dialects/abstract/query-interface.js
```diff
@@ -975,8 +975,13 @@ class QueryInterface {
 
   async select(model, tableName, optionsArg) {
     const options = { ...optionsArg, type: QueryTypes.SELECT, model };
+    const query = this.QueryGenerator.selectQuery(tableName, options, model);
```
baweinbe, line 978: looks like this should be `const query = this.queryGenerator.selectQuery(tableName, options, model);`

viraldudhat (approved):

jdgriffith:
Do I dare hope this is about to make it into sequelize?! 😍

sdepold:
if it is WIP: Can we maybe convert it into a draft PR?

sdepold:
fwiw: I just rebased. main is now including some bits and pieces about typescript. it shouldn't make a huge diff

lib/hooks.js
```diff
@@ -90,7 +90,10 @@ const Hooks = {
 
   async runHooks(hooks, ...hookArgs) {
     if (!hooks) throw new Error('runHooks requires at least 1 argument');
-
+    // Skip hooks if we are only getting raw SQL
+    if (hookArgs[0] && typeof hookArgs[0] === 'object' && hookArgs[0].getRawSql) {
```
Keimeno, line 94: Hooks can also modify the resulting query, such as `beforeFind`. I think it's best if we extend the `hookTypes` object, and add a new option to each hook that we want to run, when getting the raw SQL query

lib/dialects/abstract/query-interface.js
```diff
@@ -949,11 +949,12 @@ class QueryInterface {
 
   async select(model, tableName, optionsArg) {
     const options = { ...optionsArg, type: QueryTypes.SELECT, model };
+    const query = this.queryGenerator.selectQuery(tableName, options, model);
+    if (options.getRawSql) {
```
Keimeno, line 953: I'm unsure if `getRawSql` is the best name for this option, I would probably go with something like `queryOnly`, `getQuery`, `query` or something else.

Keimeno (changes requested):
Thanks for your contribution @seromenho, and sorry for the late response!

I've marked two things that I would change about this PR.

There's also a problem inside our `lib/model.js`, where we make use of the `select` method. https://github.com/sequelize/sequelize/blob/407137822a62897f7366980acd7eeceb443601b9/lib/model.js#L1754-L1770

We have a problem here, because our result is processed further, we will have to check for these scenarios as well.

Also, I'm unsure if we should limit this option for `findAndCountAll`, `findAll` and `findOne` queries, but it might as well make sense to separate this into multiple PRs

ephys:
One note I have about this PR is that we should avoid having different return types for our functions. It makes typing them and using them with TypeScript much more complicated.

Instead, let's add a new method that does this. Maybe `Model.getFindQuery`, `Model.findAll.getQuery()`, or let's keep it as a method on the Query Builder? #394

piotr-pawlowski:
When do you plan to merge this PR? We really need that feature.

cr0mbly:
I know there was talks about making the `QueryGenerator` public, but it's been a few years. If this isn't going to be the case we can institute the above PR. Would love to use this natively.

jdgriffith:
A part of me is sad to look back and see my first comment on this thread was four years ago. 

I think it's worth merging this PR, assuming it's ready or can be modified easily, as a stop gap until new features or v7 arrives.

Nick-Riggs:
I've needed this for years for different projects, and I've always monkey-patched it. It may not cover every case, but something to the effect of:

```
    const realQuery = sequelize.query.bind(sequelize);

    sequelize.query = async function (sql: string, options: QueryOptions) {
        if (options?.returnSqlOnly) {
            return sql;
        }

        return realQuery(sql, options);
    }
```

I hope to see it in the framework proper someday as well.

jdgriffith:
@Nick-Riggs Our team has had to do something similar because we have a dynamic query layer based on the sequelize ORM.

cr0mbly:
@ephys this seems like a pretty big want/need from the community, are we able to look at this again?

I Just note again most ORMs I've dealt with have offered easy exposure of the SQL they're building within the ORM builder itself, it'd be nice if sequlize kept to that convention:

```
### Rails
Model.all.to_sql
>>> "SELECT `models`.* FROM `models`" 

### Django
Model.objects.all().query()
>>> "SELECT `models`.* FROM `models`"
```

PR #11885, opened 2020-01-28
feat(postgres): add mustExist option to removeColumn

### LINKED ISSUE #11774: queryInterface: Support DROP COLUMN IF EXISTS

When doing some migrations, I would like to delete a column created in a previous migration.
However, this is not possible with the current code for some scenarios.

An example : 
**migration v3**
```js
'use strict';

let opts = {tableName: 'Exercises'};

module.exports = {
    up: (queryInterface, Sequelize) => {
        if (queryInterface.sequelize.options.schema) {
            opts.schema = queryInterface.sequelize.options.schema;
        }
        return Promise.all([
            queryInterface.addColumn(opts, "isValidated", {
                type: Sequelize.BOOLEAN,
                allowNull: false,
                defaultValue: false
            })
        ]);
    },

    down: (queryInterface, Sequelize) => {
        if (queryInterface.sequelize.options.schema) {
            opts.schema = queryInterface.sequelize.options.schema;
        }
        return Promise.all([
            queryInterface.removeColumn(opts, "isValidated")
        ]);
    }
};
```

**migration v4**
```js
'use strict';

let opts = {tableName: 'Exercises'};

module.exports = {
    up: (queryInterface, Sequelize) => {
        if (queryInterface.sequelize.options.schema) {
            opts.schema = queryInterface.sequelize.options.schema;
        }
        return Promise.all([
            queryInterface.removeColumn(opts, "isValidated"),
            queryInterface.addColumn(opts, "state", {
                type: Sequelize.ENUM("42", "42-42"),
                default: "42"
            })
        ]);
    },

    down: (queryInterface, Sequelize) => {
        if (queryInterface.sequelize.options.schema) {
            opts.schema = queryInterface.sequelize.options.schema;
        }
        return Promise.all([
            queryInterface.removeColumn(opts, "state")
        ]);
    }
};
```

image

Just add an option in the [options of removeColumn](https://sequelize.org/v5/class/lib/query-interface.js~QueryInterface.html#instance-method-removeColumn)

If I take the [test in the codebase](https://github.com/sequelize/sequelize/blob/master/test/unit/sql/remove-column.test.js#L21) as example : 

```js
      it('schema', () => {
        expectsql(sql.removeColumnQuery({
          schema: 'archive',
          tableName: 'user'
        }, 'email', { mustExist: false } ), {
          mssql: 'ALTER TABLE [archive].[user] DROP COLUMN [email];',
          mariadb: 'ALTER TABLE `archive`.`user` DROP `email`;',
          mysql: 'ALTER TABLE `archive.user` DROP `email`;',
          postgres: 'ALTER TABLE IF EXISTS "archive"."user" DROP COLUMN "email";'
        });
      });
```

Just read what I said above ;)

None as truncate removeColumn promise to deal with this case is horrible :( 

- [x] Yes. This issue only applies to the following dialect(s): PostgreSQL

nassimmoussa:
### Pull Request check-list

### Description of change
added "mustExist" option to queryInterface.dropColumn, the option must be boolean and if marked as false, then the database checks if column exist before trying to drop, and doesn't throw an error when trying to remove non-existing column in postgresql.

Closes #11774

### DIFF
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -259,10 +259,10 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
     return query;
   }
 
-  removeColumnQuery(tableName, attributeName) {
+  removeColumnQuery(tableName, attributeName, options) {
     const quotedTableName = this.quoteTable(this.extractTableDetails(tableName));
     const quotedAttributeName = this.quoteIdentifier(attributeName);
-    return `ALTER TABLE ${quotedTableName} DROP COLUMN ${quotedAttributeName};`;
+    return `ALTER TABLE ${quotedTableName} DROP COLUMN${options && !options.mustExist ? ' IF EXISTS' : ''} ${quotedAttributeName};`;
   }
 
   changeColumnQuery(tableName, attributes) {
--- a/lib/dialects/postgres/query-interface.js
+++ b/lib/dialects/postgres/query-interface.js
@@ -154,4 +154,21 @@ function ensureEnums(qi, tableName, attributes, options, model) {
 }
 
 
+/**
+ A wrapper that fixes MySQL's inability to cleanly remove columns from existing tables if they have a foreign key constraint.
+
+ @param  {QueryInterface} qi
+ @param  {string} tableName     The name of the table.
+ @param  {string} columnName    The name of the attribute that we want to remove.
+ @param  {object} options
+
+ @private
+ */
+function removeColumn(qi, tableName, columnName, options) {
+  options = options || {};
+  return qi.sequelize.query(qi.QueryGenerator.removeColumnQuery(tableName, columnName, options), options);
+}
+
+
 exports.ensureEnums = ensureEnums;
+exports.removeColumn = removeColumn;
--- a/lib/query-interface.js
+++ b/lib/query-interface.js
@@ -536,6 +536,8 @@ class QueryInterface {
       case 'mariadb':
         // mysql/mariadb need special treatment as it cannot drop a column with a foreign key constraint
         return MySQLQueryInterface.removeColumn(this, tableName, attributeName, options);
+      case 'postgres':
+        return PostgresQueryInterface.removeColumn(this, tableName, attributeName, options);
       default:
         return this.sequelize.query(this.QueryGenerator.removeColumnQuery(tableName, attributeName), options);
     }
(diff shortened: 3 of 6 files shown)


lib/dialects/postgres/query-interface.js
```diff
@@ -154,4 +154,21 @@ function ensureEnums(qi, tableName, attributes, options, model) {
...
+
+ @param  {QueryInterface} qi
+ @param  {string} tableName     The name of the table.
+ @param  {string} columnName    The name of the attribute that we want to remove.
+ @param  {object} options
+
+ @private
+ */
+function removeColumn(qi, tableName, columnName, options) {
+  options = options || {};
```
negreirosleo: you can use a default parameter here:
```js
function removeColumn(qi, tableName, columnName, options = {}) {
```

lib/dialects/postgres/query-generator.js
```diff
@@ -259,10 +259,10 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
     return query;
   }
 
-  removeColumnQuery(tableName, attributeName) {
+  removeColumnQuery(tableName, attributeName, options) {
     const quotedTableName = this.quoteTable(this.extractTableDetails(tableName));
     const quotedAttributeName = this.quoteIdentifier(attributeName);
-    return `ALTER TABLE ${quotedTableName} DROP COLUMN ${quotedAttributeName};`;
+    return `ALTER TABLE ${quotedTableName} DROP COLUMN${options && !options.mustExist ? ' IF EXISTS' : ''} ${quotedAttributeName};`;
```
hjJunior: What do you think about moving this logic of `options.mustExist` to a const, and late it, just use it on your return

lib/dialects/postgres/query-interface.js
hjJunior: This snip is good, but there is a small problem, if executed passing null, for example, imagine this following function invocation:

```js
removeColumn(qi, tableName, columnName, null)
```

The options passed will be null instead of an empty object... Actually I'm not too sure if this could trigger an error, but, if it is required to assert it, it'll be necessary to do something close to what he did.

But also it isn't good practice to modify function attributes, I think you could create a new variable instead of modifying the `options`

lib/dialects/postgres/query-interface.js
```diff
@@ -154,4 +154,21 @@ function ensureEnums(qi, tableName, attributes, options, model) {
 }
 
 
+/**
+ A wrapper that fixes MySQL's inability to cleanly remove columns from existing tables if they have a foreign key constraint.
```
sushantdhiman: Not true for `postgres`

lib/dialects/postgres/query-interface.js
sushantdhiman: Please use the default parameter

papb:
I agree with the review comments by sushantdhiman

PR #11929, opened 2020-02-12
Support `{ plain: true }` in typescript typings for sequelize.query

### LINKED ISSUE #11595: Typescript bindings incomplete when using `sequelize.query()`

segersniels:
### Pull Request check-list

### Description of change

Closes #11595

### Related PR's
- https://github.com/sequelize/sequelize/pull/11596

### DIFF
--- a/types/lib/sequelize.d.ts
+++ b/types/lib/sequelize.d.ts
@@ -1206,6 +1206,7 @@ export class Sequelize extends Hooks {
   ): Promise<M[]>;
   public query<T extends object>(sql: string | { query: string; values: unknown[] }, options: QueryOptionsWithType<QueryTypes.SELECT> & { plain: true }): Promise<T>;
   public query<T extends object>(sql: string | { query: string; values: unknown[] }, options: QueryOptionsWithType<QueryTypes.SELECT>): Promise<T[]>;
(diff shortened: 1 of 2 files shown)


papb:
It looks like I am not able to add commits on top of your branch. This never happened to me before, maybe you did not check the box 'allow edits from maintainers'? Or do you have any other idea what could be causing this?

Please apply the changes from the other PR yourself here then :) thanks!

segersniels:
@papb I cherry-picked the required changes 🚀

papb:
Thank you!! :tada:

PR #11931, opened 2020-02-15
fix(query-interface): allow passing null for query interface insert

SimonSchick:
### Pull Request check-list

### Description of change

Add `null` as allowed parameter to `insert` on `queryInterface` and adjust return type of `bulkInsert` to include `number` as it can be one.

### DIFF
--- a/types/lib/query-interface.d.ts
+++ b/types/lib/query-interface.d.ts
@@ -451,7 +451,7 @@ export class QueryInterface {
   /**
    * Inserts a new record
    */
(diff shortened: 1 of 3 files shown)


types/lib/query-interface.d.ts
```diff
@@ -473,7 +473,7 @@ export class QueryInterface {
     records: object[],
     options?: QueryOptions,
     attributes?: string[] | string
-  ): Promise<object>;
+  ): Promise<object | number>;
```
papb: When does it return an object and when does it return a number? And by the way what is this object/number? I couldn't find in the docs :sweat_smile:

PR #11933, opened 2020-02-18
test(mssql): re-enable config test

papb:
Do not skip this MSSQL test


papb:
Actually it didn't work :sweat_smile:

PR #11943, opened 2020-02-19
Handle affectedRows being null after an update

framp:
`affectedRows` is null for some reason after an update with zero affected rows.

This prevents Sequelize from throwing an error, but you may want to patch bulkUpdate not to return null and returns a empty [].

This happened while using the postgres driver.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3225,7 +3225,7 @@ class Model {
       return this.QueryInterface.bulkUpdate(this.getTableName(options), valuesUse, options.where, options, this.tableAttributes).then(affectedRows => {
         if (options.returning) {
           instances = affectedRows;
-          return [affectedRows.length, affectedRows];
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Can you demonstrate this issue? What version of postgres, sequelize etc

framp:
Hi @sushantdhiman, thanks for having a look!

You can reproduce it with latest node packages for sequelize, postgres.
I used postgres 9.6.16

Small gist here that showcase the problem, with instructions:
https://gist.github.com/framp/e09e4d91638decd8c5ab988cfb67733b

sushantdhiman:
I can confirm this issue (with `plain: true`). Please add a test [here](https://github.com/sequelize/sequelize/blob/master/test/integration/model/bulk-create.test.js) and this should be good to go

WikiRik:

Issue #11894, opened 2020-02-02
coun() and Distinct = syntax error


AuHau:
### What are you doing?

Using `count()` and `distinct: true` option. 
Not directly but through [feathers-sequalize](https://github.com/feathersjs-ecosystem/feathers-sequelize/blob/master/lib/index.js#L131). 

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/43

### What do you expect to happen?

Count returned

### What is actually happening?

SQL generaged:
```
SELECT count(DISTINCT(*)) AS `count` FROM `StorageOffers` AS `StorageOffer`;
```

Error returned:
```
SQLITE_ERROR: near \"*\": syntax error
```

### Environment

- Sequelize version: sequelize@5.21.3 
- Node.js version: v12.9.0 
- Operating System: macos
- If TypeScript related: TypeScript version: 3.7.5

### How does this problem relate to dialects?

- [ ] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [x] I don't know, I was using SQLite, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [X] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

TimMensch:
Good timing! I just submitted a PR for this: #11893

I'm using Feathers-Sequelize and ran into the same problem. 

It also fails in PostgreSQL, and looking at other (closed and ignored?!) bug reports, in MySQL as well.

Sequelize just shouldn't generate invalid SQL given reasonable parameters. There _is_ a workaround, but the real fix is absolutely brain-dead trivial, so in my opinion should be an easy merge decision. 

The workaround, for what it's worth, is to set `sequelize.col = Model.primaryKeyAttributes` in a before hook. Here's what my Feathers-Sequelize code looks like (though with TypeScript):

```typescript
export function fixCount(ThisModel: typeof Model) {
    return (context: HookContext) => {
        const sequelize = context.params.sequelize || {};
        context.params.sequelize = sequelize;

        sequelize.col = ThisModel.primaryKeyAttributes;
    };
}
```

And you'd use this like so:

```typescript
const hooks = {
    before: {
        find: [fixCount(MyModel)],
    }
[...]

PR #11893, opened 2020-02-02
fix(model): count() code generation for distinct rows

TimMensch:
### Pull Request check-list

- [X] (see below) Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [X] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [X] Have you added new tests to prevent regressions? 
- [X] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? (n/a)
- [X] Did you update the typescript typings accordingly (if applicable)? (n/a)
- [X] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?

### Description of change

Add code to `count()` to cause it to create valid SQL when `distinct=true` and `col` is unspecified or `"*"`. Otherwise the SQL generated is `COUNT(DISTINCT(*))`, which is invalid. Issues referenced below were already closed, but the issue hadn't actually been fixed, and it _keeps coming up_.

Made this change against v5, because I need it ASAP, but the identical change should work against v6. I can submit another PR against `master`

#### Notes:

Tested against Postgres 11 only. Three time-based tests failed; it seems like there's a time zone issue? Some other Model/create test failed, also likely due to database version mismatch (permission denied to create extension "uuid-ossp"). Sorry, I _don't_ have time to install the custom Docker Postgres; the related tests passed.

Related tickets: #2713, #6404, #6418, #7344 -- all were "closed", but none were previously _actually fixed_. Workarounds were listed, but those workarounds don't work _at all_ if you're using Sequelize with Feathers-Sequelize and Sequelize-TypeScript. Besides, Sequelize shouldn't generate bad code when the options aren't perfect.

Issue #6418, opened 2016-08-11
result.count for findAndCountAll is incorrect with include


andreabianco:
## What you are doing?
Hello, I'm trying to use findAndCountAll methods for list view pagination.
I have different models but I can post one example.

```js
sequelize.define('Model1');
sequelize.define('SubModel1');
sequelize.define('SubModel2');
sequelize.define('SubModel3');
```
Model1 is associated with SubModule1, SubModule2 and SubModules with hasMany relation.

Now in the database I have 15 Model1 items.

## What do you expect to happen?
If I perform the findAndCountAll I expect to see the following response:
```js
{ count: 15,
  rows: [...] }
```
## What is actually happening?
But the real result is:
{ count: 33,
  rows: [...] }

__Dialect:__ mysql
__Database version:__ 5.6
__Sequelize version:__ 3.23.2

janmeier:
Please show the actual code for each model, their associations, and the call to find and count - we can't help you if we can't replicate the issue...

andreabianco:
Sure!

```js
var Module1 = sequelize.define('Module1', {
    name: DataTypes.STRING
  }, {
  associate: function(models) {
     Module1.hasMany(models.SubModel1);
     Module1.hasMany(models.SubModel2);
    Module1.addScope('all', {
        include: [{
          model: models.SubModel1
        }, {
          model: models.SubModel2
        }]
      });
  }
});

var SubModel1 = sequelize.define('SubModel1', {
    name: DataTypes.STRING
  });

var SubModel2 = sequelize.define('SubModel2', {
    name: DataTypes.STRING
  });

return Module1
    .scope('all')
    .findAndCountAll({ where: {}, limit: 10, offset: 0 })

```

Sorry for bad formatting

leohihimax:
+1, we met the same issue.

sushantdhiman:
@leohihimax and @andreabianco Have you tried passing `distinct: true` ( with `col`, if required).

andreabianco:
@sushantdhiman Can you explain better how to use distinct property?

leohihimax:
@sushantdhiman thx, distinct option solved this issue.
@andreabianco [distinct usage](http://sequelize.readthedocs.io/en/v3/api/model/#countoptions-promiseinteger).  Actually, the API `findAndCountAll` use `count` API to count.

andreabianco:
@leohihimax Is it right?

```js
return Module1
    .scope('all')
[...]

PR #11946, opened 2020-02-20
fix(model): count() code generation for distinct rows

TimMensch:
Add code to count() to cause it to create valid SQL when distinct=true and col is unspecified or `"*"`. Otherwise the SQL generated is `COUNT(DISTINCT(*))`, which is invalid. Issues referenced below were already closed, but the issue hadn't actually been fixed, and it _keeps coming up_.

Closes: #11894, #2713, #6404, #6418, #7344

### Pull Request check-list

- [X] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)? **See Below**
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? **N/A**
- [ ] Did you update the typescript typings accordingly (if applicable)? **N/A**

### Description of change

Add code to count() to cause it to create valid SQL when distinct=true and col is unspecified or "*". Otherwise the SQL generated is COUNT(DISTINCT(*)), which is invalid. Issues referenced below were already closed, but the issue hadn't actually been fixed, and it keeps coming up.

#11893 was this PR against v5. This PR is against master.

Notes:
**Tests:** Tested against Postgres 11 only. There are failures, but they seem to be failures in `master`, not in anything that I changed. The only change I made is in the `count()` function, and none of the failures hits that function. Sorry, I don't have time to install the custom Docker Postgres; the related tests passed.

Related tickets: #2713, #6404, #6418, #7344 -- all were "closed", but none were previously actually fixed. Workarounds were listed, but those workarounds don't work at all if you're using Sequelize with Feathers-Sequelize and Sequelize-TypeScript. Besides, Sequelize shouldn't generate bad code when the options aren't perfect.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2033,7 +2033,9 @@ class Model {
       if (options.include) {
         col = `${this.name}.${options.col || this.primaryKeyField}`;
       }
-
+      if (options.distinct && col === '*') {
+        col = this.primaryKeyField;
+      }
       options.plain = !options.group;
       options.dataType = new DataTypes.INTEGER();
       options.includeIgnoreAttributes = false;
--- a/test/integration/model/count.test.js
+++ b/test/integration/model/count.test.js
@@ -109,6 +109,21 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         });
     });
 
+    it('should be able to specify NO column for COUNT() with DISTINCT', function() {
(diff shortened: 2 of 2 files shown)


papb:
Nice, thank you very much! 🎉

aommyindy:
thank you very much!.

Issue #11665, opened 2019-11-11
bulkCreate generates incorrect `ON CONFLICT ... DO UPDATE` clause when model has > 1 primary key field and column names differ from field names


huned:
## Issue Description

bulkCreate generates incorrect `ON CONFLICT ... DO UPDATE` clause when model has > 1 primary key field and column names differ from field names

### What are you doing?

Here is a test case that illustrates the problem:

```js
it('when the primary key column names and model field names are different and have unique constraints', function() {
  const Person = this.sequelize.define('Person', {
    emailAddress: {
      type: DataTypes.STRING,
      allowNull: false,
      primaryKey: true,
      unique: true,
      field: 'email_address'
    },
    name: {
      type: DataTypes.STRING,
      allowNull: false,
      field: 'name'
    }
  }, {});

  return Person.sync({ force: true })
    .then(() => {
      const inserts = [
        { emailAddress: 'a@example.com', name: 'Alice' }
      ];
      return Person.bulkCreate(inserts);
    })
    .then(people => {
      expect(people.length).to.equal(1);
      expect(people[0].emailAddress).to.equal('a@example.com');
      expect(people[0].name).to.equal('Alice');

      const updates = [
        { emailAddress: 'a@example.com', name: 'CHANGED NAME' },
        { emailAddress: 'b@example.com', name: 'Bob' }
      ];

      return Person.bulkCreate(updates, { updateOnDuplicate: ['emailAddress', 'name'] });
    })
    .then(people => {
      expect(people.length).to.equal(2);
      expect(people[0].emailAddress).to.equal('a@example.com');
      expect(people[0].name).to.equal('CHANGED NAME');
      expect(people[1].emailAddress).to.equal('b@example.com');
      expect(people[1].name).to.equal('Bob');
    });
});
```

The test fails because model is generating the `ON CONFLICT ... DO UPDATE` clause using the name of the sequelize attribute (`emailAddress`), not the table's column name (`email_address`).

```
  1) [POSTGRES] Model
       bulkCreate
         updateOnDuplicate
           should support the updateOnDuplicate option with primary keys
             RUNME when the primary key column names and model field names are different and have unique constraints:
     SequelizeDatabaseError: column "emailAddress" does not exist
      at Query.formatError (lib/dialects/postgres/query.js:366:16)
      at /Users/huned/data/work/forks/sequelize-non-fork/lib/dialects/postgres/query.js:72:18
  From previous event:
[...]

PR #11984, opened 2020-03-08
fix(model): updateOnDuplicate handles composite keys now

thaddeusmt:
### Pull Request check-list

### Description of change

I noticed that my original PR #11163 to support `updateOnDuplicate` for Postgres (and subsequent community updates since then) isn't working with composite unique keys. This fixes that, I think.

This bug appears to have been raised in issues #11569 and #11534, and maybe #11665?

I think PR #11611 is supposed to fix this? But it looks like it also adds additional functionality (`upsertIndex` and `upsertKeys` options on `bulkCreate()`), and has not been merged yet.

I added 2 tests as well: one for composite `primaryKey` bulkCreates, and one for composite `unique` bulkCreates. All Postgres tests appear to be passing.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -858,7 +858,7 @@ class Model {
    *
    * @see
    * <a href="/master/manual/model-basics.html">Model Basics</a> guide
-   * 
+   *
    * @see
    * <a href="/master/manual/model-basics.html">Hooks</a> guide
    *
@@ -2708,7 +2708,7 @@ class Model {
             // Get primary keys for postgres to enable updateOnDuplicate
             options.upsertKeys = _.chain(model.primaryKeys).values().map('field').value();
             if (Object.keys(model.uniqueKeys).length > 0) {
-              options.upsertKeys = _.chain(model.uniqueKeys).values().filter(c => c.fields.length === 1).map(c => c.fields[0]).value();
+              options.upsertKeys = _.chain(model.uniqueKeys).values().filter(c => c.fields.length >= 1).map(c => c.fields).reduce(c => c[0]).value();
             }
           }
 
@@ -3842,13 +3842,13 @@ class Model {
 
   /**
    * Validates this instance, and if the validation passes, persists it to the database.
-   * 
+   *
    * Returns a Promise that resolves to the saved instance (or rejects with a `Sequelize.ValidationError`, which will have a property for each of the fields for which the validation failed, with the error message for that field).
-   * 
+   *
    * This method is optimized to perform an UPDATE only into the fields that changed. If nothing has changed, no SQL query will be performed.
-   * 
+   *
    * This method is not aware of eager loaded associations. In other words, if some other model instance (child) was eager loaded with this instance (parent), and you change something in the child, calling `save()` will simply ignore the change that happened on the child.
-   * 
+   *
    * @param {object}      [options] save options
    * @param {string[]}    [options.fields] An optional array of strings, representing database columns. If fields is provided, only those columns will be validated and saved.
    * @param {boolean}     [options.silent=false] If true, the updatedAt timestamp will not be updated.
--- a/test/integration/model/bulk-create.test.js
+++ b/test/integration/model/bulk-create.test.js
@@ -586,6 +586,117 @@ describe(Support.getTestDialectTeaser('Model'), () => {
                 expect(people[1].name).to.equal('Bob');
               });
           });
+
+          it('when the composite primary key column names and model field names are different', function() {
+            const Person = this.sequelize.define('Person', {
+              systemId: {
+                type: DataTypes.INTEGER,
+                allowNull: false,
+                primaryKey: true,
+                field: 'system_id'
+              },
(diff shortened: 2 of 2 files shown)


cxcorp:
Will this fix be released under v5?

JPeer264:
@cxcorp I am not sure if it will get fixed in v5 as well but there are very little [breaking changes](https://sequelize.org/master/manual/upgrade-to-v6.html) in v6 (yet) so the migration from v5 to v6 should be nearly seamless.

ericjansen1988:
Is this fix released already? I wanted to bulkCreate something with a composite primary key, but it keeps failing me. I am using postgres. Have tried both v5 and v6 of sequelize.
bulkCreate(data) --> returns SequelizeUniqueConstraintError: Validation error
bulkCreate(data, updateOnDuplicate: ['column1', column2]) --> returns SequelizeDatabaseError: syntax error at or near ")"

JPeer264:
@ericjansen1988 no, not yet it is just on master right now.

There are two possibilities. But those possibilities should just be done if you know what you are doing and which effects it could have for the future:

1. You can install the version of the master, you could also specify the commit SHA (this is what we do currently):

```sh
$ npm i https://github.com/sequelize/sequelize
```

But I just recommend this if you are aware of the risks pulling from master directly (use a package-lock too just in case). This version is still in Beta, there might be other bugs appearing - which I don't think so, but the risk is still there. Remember, if you might update to stable v6 the official API could be changed.

--- 

2. Create a fork from a stable v5 version and cherry-pick this commit into your forked codebase -> Then install your own fork, also via Github to not spam the npm registry with forks

This could be leading to merging conflicts, depending on the codebase.

lib/model.js
```diff
@@ -2708,7 +2708,7 @@ class Model {
             // Get primary keys for postgres to enable updateOnDuplicate
             options.upsertKeys = _.chain(model.primaryKeys).values().map('field').value();
             if (Object.keys(model.uniqueKeys).length > 0) {
-              options.upsertKeys = _.chain(model.uniqueKeys).values().filter(c => c.fields.length === 1).map(c => c.fields[0]).value();
+              options.upsertKeys = _.chain(model.uniqueKeys).values().filter(c => c.fields.length >= 1).map(c => c.fields).reduce(c => c[0]).value();
```
SimonSchick, line 2711: I'm confused about the `reduce` here, would this not effectively end up returning `undefined` all the time?

cc @eseliger @sushantdhiman

eseliger: I think you are correct. The initializer for the reduce is `undefined` (not specified), and `c` is the `previous` argument, so it would effectively always return the undefined, leaving the actual values untouched.

ridakk:
when will this fix be released? any plan or a date?

klausXR:
Is there any workaround or a temporary patch for v5?
I think this is a pretty useful feature for a lot of people, I have to do this in multiple parts of my project, I've been keeping an eye on this issue, its fixed in v6, but thats in beta, I can't just deploy that to production.

brianschardt:
This is a much needed feature when can we get this released?

brianschardt:
Can we get this change in version 5?

klausXR:
The issue, as I understand it, is that only conflicts on the primary key, as of now, will trigger the update on duplicate part, so if you have a primary key and a unique constraint, only the primary key will be taken into account, because it only gets the first one, so having more than one is no good. 
In my case, my tables have two columns with together form uniqueness, so I used composite primary keys to solve the problem - I ended up with a single primary key, but it was a combination of two columns. What this solves is the need for having PK + Unique constraint.

As a simple example, if you have the following table

players
  name teamId

and for the sake of simplicity we assume that a team cannot have two players with the same name, instead of adding the unique constraint, you can add a composite primary key like so

`ALTER TABLE "players" ADD CONSTRAINT "pk_name_goes_here" PRIMARY KEY ("name", "teamId")`

Jplus2:
Can we get this on V5 please

brianschardt:
what can i do to get this in v5?

aecorredor:
@sushantdhiman is there a chance we get this into v5 since v6 is still in beta? It seems like a pretty important feature since a lot of people want to do bulk updates. Is there any reason for not getting it into v5?

brianschardt:
Just tried this on v6 and this will fail if the composite is a unique index. as the this.uniqueKeys does not pickup composit indexes. So it looks like there is still some more work to do on this front.

aecorredor:
@brianschardt I can confirm it's still an issue for me as well for your same use case. Any other workarounds to achieve the same goal?

brianschardt:
This PR I just created should fix this issue: https://github.com/sequelize/sequelize/pull/12516
Lets encourage the sequelize team to get this merged.

lib/model.js
VineetJD3: Due to the `.reduce` updateOnDuplicate does not work for Composite unique keys or multiple unique keys. Can you recheck for the same.

PR #12007, opened 2020-03-16
Added static Model.decrement for TypeScript

Duplicated:
### Pull Request check-list

### Description of change

I actually came across this issue while I was at work. During the migration of my company's API project from JS to TS, I noticed that the `static decrement` type was missing. Dug around a bit and found out that the previous contributor only added the `increment` one.

The code is essentially just a reword, since all overloads are the same as the `increment` counterpart. I've run it against `test-typings` with zero errors. Please let me know if I'm missing anything else.


types/lib/model.d.ts
```diff
@@ -2077,6 +2077,33 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
...
   ): Promise<M>;
 
+  /**
+   * Decrements a single field.
+   */
+  public static decrement<M extends Model, K extends keyof M>(
+    this: { new(): M },
+    field: K,
+    options: IncrementDecrementOptionsWithBy
+  ): Promise<M>;
```
SimonSchick: I actually believe the docs (and typings on increment) are wrong.

The static return signature appears to be
```return promise.then(affectedRows => {
      if (options.returning) {
        return [affectedRows, affectedRows.length];
      }

      return [affectedRows];
    });
```

so `[number] | [unknown[], number]`

Duplicated: @SimonSchick I haven't looked into the actual `static increment` and `static decrement` source code yet. I'll look into it later this week then.

Duplicated: @SimonSchick I have re-committed an update to my PR. Feel free to review it.

Duplicated:
> It actually seems increment/decrement are even more messed up:
> 
> From a personal project:
> 
> ```
> > void(p.container.models.user.increment({ verified: 1 }, { where: {} }).then(data => console.log(data)))
> undefined
> > [ [ undefined, 1 ] ]
> ```
> 
> So the return type appears to be something nested for some reason...
> 
> Sorry I didn't have the time to dig into this before.

Given that the project's about to go forward with v6, should I change the typing to match what QueryInterface currently returns (so, `Promise<Array<M[]>> | Promise<Array<M[], number>>`)? Or does the team has plans to continue maintaining v5 still?

SimonSchick:
I think aligning the return types would be a good thing yes.

lib/model.js
```diff
@@ -3385,10 +3385,10 @@ class Model {
 
     return promise.then(affectedRows => {
       if (options.returning) {
-        return [affectedRows, affectedRows.length];
+        return [affectedRows[0], affectedRows[1]];
```
SimonSchick: You could just return the tuple directly 😛

Duplicated:
@SimonSchick Please see the latest commit.

Something I have discovered from  https://github.com/sequelize/sequelize/blob/v5/lib/model.js#L3386:
```js
    return promise.then(affectedRows => {
      if (options.returning) {
        return [affectedRows, affectedRows.length];
      }

      return [affectedRows];
    });
});
```
Here, `affectedRows` (as returned from `QueryInterface.increment` and `QueryInterface.decrement`) always return both actual affected rows **and** their count.

However, there's something else that I'm not quite sure about: both the docs and the code suggest that `static increment` should return the number of rows affected if `options.returning` is set to `true`. However, this option doesn't even exist in the `IncrementDecrementOptionsWithBy` type. So maybe there are some confusions here (especially since `affectedRows` returns row count by default)?

I could open a separate PR to add another type with that `returning` flag to the `IncrementDecrementOptionsWithBy` extending list though.

SimonSchick:
Unsure, @papb @sushantdhiman any ideas?

lib/model.js
Duplicated: Done :smile:

Duplicated:
@SimonSchick Still haven't received any response from either maintainers. Am I missing anything in the PR?

Duplicated:
> I don't think we are missing anything?

Thanks again @SimonSchick 

One other thing: I've just run `npm run test-integration-all` again on my local machine to try to replicate the error (as reported by CI). However, it went through that specific error without any trouble. Does this mean that CI should be able to rebuild my PR correctly now?

lib/model.js
```diff
@@ -3383,12 +3383,12 @@ class Model {
       promise = this.QueryInterface.increment(this, this.getTableName(options), values, where, options);
     }
 
-    return promise.then(affectedRows => {
+    return promise.then(([actualRows, rowCount]) => {
```
sushantdhiman, line 3386: I need to check this for v5

Duplicated:
> This change should be applied to master first

Do I need to do anything else from my end? Should I create a new PR from master (which I assume is v6-based) instead?

sushantdhiman:
@Duplicated Yes, please open this PR for master as well

Duplicated:
@sushantdhiman I couldn't replicate the same behavior in `master`. Maybe it's something that pertains to v5 only?

Regardless, since this PR is quite out of date, maybe someone has already fixed it. I'll start a new one if I run into the issue again down the road.

Issue #10946, opened 2019-05-13
Field with type Sequelize.JSON returns string when MySQL dialect is used for a MariaDB database


kishmiryan-karlen:
## What are you doing?
When querying a model, which has a field with type `sequelize.JSON`, it returns JSON string instead of parsed JS object for that field.

```js
// dbService.js
// Connecting to DB, params are coming from config file
export const databaseService = new Sequelize(database, user, password, {
  host,
  port,
  dialect: 'mysql',
  logging,
  operatorsAliases: Sequelize.Op,
  define: {
    freezeTableName: true,
    charset,
    collate,
  },
});

// models/index.js
// Creating the models
import fs from 'fs';
import path from 'path';
import sequelize from 'sequelize';
import { databaseService } from 'utils/database/dbService';

const basename = path.basename(__filename);
const db = {};

fs.readdirSync(__dirname)
  .filter((file) => {
    return !file.startsWith('.') && file !== basename && file !== 'index.js' && file.slice(-3) === '.js';
  })
  .forEach((file) => {
    const fileName = path.join(__dirname, file);
    const model = require(fileName).default(databaseService, sequelize);
    db[model.name] = model;
  });

Object.keys(db).forEach((modelName) => {
  if (db[modelName].associate) {
    db[modelName].associate(db);
  }
});

db.sequelize = databaseService;
export default db;

// models/Test.js
// Creating a Test model
export default (dbService, sequelize) => {
  const Test = dbService.define('Test', {
    id: {
      type: sequelize.UUID,
      defaultValue: sequelize.UUIDV1,
      primaryKey: true,
    },

    data: {
      type: sequelize.JSON,
    },
  });

  return Test;
};

// test.js
// Usage in test script
import './models';
import { databaseService } from './dbService';

const newTest = await db.models.Test.create({
  data: { x: 1 }
});
console.log('check 1', typeof newTest.data);  // check1 object

const test = await db.models.Test.findOne({ id: newTest.id });
console.log('check 2', typeof test.data);  // check 2 string
console.log('check 3', typeof test.get('data')); // check 3 string
```

**To Reproduce**
1. As you can see above, I'm connecting to my DB and creating model `Test`, which has two fields: `id` and `data`. The latter one is of type `sequelize.JSON`.
2. Then in my test script, I create an instance of `Test` model, providing JS object for `data` field. On the next line when I print type of `data`, it prints correctly `object`.
[...]

PR #12008, opened 2020-03-16
fix(mariadb): correct json columns when connecting to mysql

SimonSchick:
### Pull Request check-list

- [-] Have you added new tests to prevent regressions?

### Description of change

This fixes reading JSON columns when using the mariadb connector for mysql databases (which should work just fine according to mariadb connector spec).

Testing for this is obviously hard since it would require an e2e suite that would connect the mariadb driver with a mysql db so some guidance would be appreciated.

On a note: It might be worth trying to merge the mariadb and mysql adapters at some point.


lib/dialects/mariadb/query.js
```diff
@@ -179,19 +179,21 @@ class Query extends AbstractQuery {
...
-          row[modelField.fieldName] = row[modelField.fieldName] ? JSON.parse(
-            row[modelField.fieldName]) : null;
+        for (const r of rows) {
+          if (!wasParsed) {
+            r[fieldName] = r[fieldName] ? JSON.parse(r[fieldName]) : null;
+          }
           if (DataTypes.JSON.parse) {
-            return DataTypes.JSON.parse(modelField, this.sequelize.options,
-              row[modelField.fieldName]);
+            r[fieldName] = DataTypes.JSON.parse(modelField, this.sequelize.options, r[fieldName]);
```
papb: This does not seem equivalent to the previous implementation, is this intentional? Previously the array element itself was replaced by the result of the `DataTypes.JSON.parse` call. Now the `r[fieldName]` is replaced, right?

papb:
> Testing for this is obviously hard since it would require an e2e suite that would connect the mariadb driver with a mysql db so some guidance would be appreciated.

Hmm indeed... But do we officially support such usage? I've seen this around, but I no longer remember, why would someone want to use the mariadb driver on a mysql db (or vice-versa)?

> On a note: It might be worth trying to merge the mariadb and mysql adapters at some point.

Agreed... Not an easy task though :sweat_smile:

lib/dialects/mariadb/query.js
SimonSchick: Seems that way, misinterpreted the code due to lack of docs, will revert that assignment.

lib/dialects/mariadb/query.js
```diff
@@ -179,19 +179,21 @@ class Query extends AbstractQuery {
...
-          row[modelField.fieldName] = row[modelField.fieldName] ? JSON.parse(
-            row[modelField.fieldName]) : null;
+        for (const r of rows) {
+          if (!wasParsed) {
+            r[fieldName] = r[fieldName] ? JSON.parse(r[fieldName]) : null;
+          }
           if (DataTypes.JSON.parse) {
-            return DataTypes.JSON.parse(modelField, this.sequelize.options,
-              row[modelField.fieldName]);
+            DataTypes.JSON.parse(modelField, this.sequelize.options, r[fieldName]);
```
papb: I'm afraid this is still not equivalent to the original code. Now you're computing something but not saving it anywhere. It seems to me that using `rows = rows.map(/* ... */)` is probably the best

SimonSchick: Let me re-read this entire segment...

SimonSchick: This doesn't make sense since `rows` is never returned only new arrays are spawned of it and then iterated.
Why should a full row be replaced with the result of a single column being parsed? This logic does not make any sense to me and seems to be in error.

lib/dialects/mariadb/query.js
SimonSchick: Unless the purpose of `DataTypes.JSON.parse` is to mutate new data by not returning, this doesn't really work anyways since `rows` is discarded.

SimonSchick:
@papb the mariadb driver states it's fully compatible with mysql and from tests run locally connecting the driver to mysql and a percona cluster this appears to be the case.
The mariadb driver is also more performant than mysql.

lib/dialects/mariadb/query.js
papb: @SimonSchick Ah, you're very right, I was judging your code assuming the current code was somewhat correct, but as you said the `rows` variable is reassigned but unused after that point!

Sorry about that. Well, since the current code is incorrect, I wonder what it should be doing instead...

papb: Do you think this weird part of the code (in which the element is reassigned to no use) can be the cause for the issue you're trying to solve? And if yes, why would it happen only when "mixing" mysql and mariadb? Or did we just stumble on a different issue?

papb:
> @papb the mariadb driver states it's fully compatible with mysql and from tests run locally connecting the driver to mysql and a percona cluster this appears to be the case.
> The mariadb driver is also more performant than mysql.

I see... Perhaps we should have all our tests be executed as well in this "mix" of mysql and mariadb then? As a new job in our CI?

lib/dialects/mariadb/query.js
SimonSchick: No, the issue is that the data is already parsed and the JSON.parse ends up throwing :)

sushantdhiman:
Is this supposed to fix https://github.com/sequelize/sequelize/issues/10946 ?

SimonSchick:
Don't have time atm, sorry.

PR #12022, opened 2020-03-22
fix(model): do not lose the value if the attribute alias matches the include alias (#12021)

### LINKED ISSUE #12021: The value of the aliased attribute is lost if alias matches the include alias
### What are you doing?

```js
User.findOne({
  attributes : [
    'id',
    [
      db.sequelize.fn('COUNT', db.sequelize.col('projects.id')), 
      'projects' // <= matches
    ]
  ],
  include : [
    {
      model : Project,
      as : 'projects',  // <= matches
      attributes : []
    }
  ],
  group : ['user.id']
})
```

User projects count in the `projects` property of the result.

The `projects` property is missing.

### Environment

- Sequelize version: 5.21.5 
- Node.js version: 12.14.0 
- Operating System: Mac OS X 10.15.3

aristov:
### Pull Request check-list

### Description of change

Closes #12021

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1815,9 +1815,11 @@ class Model {
     if (!results.length) return original;
 
     // Filter attribute aliases with the same name
-    const includes = options.include.filter(include => {
-      return !options.attributes.some(attr => Array.isArray(attr) && attr[1] === include.as);
-    })
+    const includes = options.attributes?
+      options.include.filter(include => {
+        return !options.attributes.some(attr => Array.isArray(attr) && attr[1] === include.as);
+      }) :
+      options.include
     return Promise.map(includes, include => {
       if (!include.separate) {
         return Model._findSeparate(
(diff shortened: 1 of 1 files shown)


aristov:
Let me know If anybody cares. Then I complete the PR by tests.

lib/model.js
```diff
@@ -1814,7 +1814,13 @@ class Model {
 
     if (!results.length) return original;
 
-    return Promise.map(options.include, include => {
+    // Filter attribute aliases with the same name
+    const includes = options.attributes?
```
sushantdhiman, line 1818: This adds additional complexity for something that can be fixed by changing alias name for attributes by user, I don't think this should be fixed. Your thoughts

aristov: Since the DBMS supports this naming, I expect sequelize to support it as well. The current behavior is unexpected and makes it take time to understand why the ORM is not working as expected.

facuparedes:
Hi @aristov!
I have this issue, do u know a workaround?

Greetings!

aristov:
@facuparedes 
In my project I'm still using my [fork from v5](https://github.com/aristov/sequelize/tree/v5) with this and other fixes I need, but it's not a very good solution.
If you are using v6, you can migrate the changes from this PR to v6 and redefine the corrected method at your project level.

PR #12069, opened 2020-04-05
Convert to async/await, don't use bluebird API [WIP]

jedwards1211:
#11859

@sushantdhiman @papb Here's my combined work so far, FWIW.  I saw that you want to make small
PRs one by one, but after seeing how much the codebase is littered with Bluebird API
usage, it can be difficult to make small changes without breaking tests.  Unless
we remove all Bluebird API usage first, but I dunno, removing some of those methods,
like `.tap`, would leave the code in a pretty hard to read intermediate state...

I was hoping I could automate most of the conversion to `async`/`await` with `jscodeshift`/https://github.com/sgilroy/async-await-codemod, but that project cautions that its `await-promise-chain` it doesn't preserve behavior in all cases.  It's tempting to try to improve on it/write my own tool, but I decided it's probably better to just carefully transform stuff by hand.

If nothing else, the fully converted files checked below should be good to cherry-pick into separate PRs.

- [ ] lib/associations/base.js
- [ ] lib/associations/belongs-to-many.js
- [ ] lib/associations/belongs-to.js
- [ ] lib/associations/has-many.js
- [ ] lib/associations/has-one.js
- [ ] lib/associations/base.js
- [ ] lib/dialects/connection-manager.js
- [ ] lib/dialects/mariadb/connection-manager.js
- [ ] lib/dialects/mariadb/query.js
- [ ] lib/dialects/mssql/connection-manager.js
- [ ] lib/dialects/mssql/query-interface.js
- [ ] lib/dialects/mysql/connection-manager.js
- [ ] lib/dialects/mysql/query-interface.js
- [ ] lib/dialects/postgres/connection-manager.js
- [ ] lib/dialects/postgres/query-interface.js
- [ ] lib/dialects/sqlite/connection-manager.js
- [ ] lib/dialects/sqlite/query-interface.js
- [ ] lib/hooks.js
- [ ] lib/instance-validator.js
- [ ] lib/model.js
- [x] lib/query-interface.js
- [x] lib/sequelize.js
- [ ] lib/transaction.js

### Bluebird API usage to remove
- [x] `Promise.map(items, mapper)` => `Promise.all(items.map(mapper))`
- [x] `.map(mapper)` => `.then(value => Promise.all(value.map(mapper)))`
- [x] `Promise.map(items, mapper, {concurrency})` => `require('p-map')(items, mapper, {concurrency})` (just one case of this in the tests AFAIK)
- [x] `.return(value)` => `.then(() => value)`
- [x] `.tap(() => return something())` => `.then(value => Promise.resolve(something()).then(() => value))`
- [x] `Promise.try(fn)` => `Promise.resolve().then(fn)`
- [ ] `Promise.each(items, itertee)` => async `for..of` loop
- [ ] `.throw(err)` => `.then(() => { throw err })`
- [x] `Promise.join`
- [x] `.reflect` - `instance-validator.js` no longer relies on this
- [x] `.get`
- [ ] `.finally`
- [ ] `Promise.promisify`
- [ ] `Promise.using`
- ...probably more I'm not aware of


jedwards1211:
There are a lot of existing methods that return promises but can sometimes throw synchronously instead of returning a rejected promise, which is not really kosher.  I fixed all these methods in `QueryInterface`, and currently a number of tests that expect them to synchronously throw are broken.

sushantdhiman:
@jedwards1211 If possible we should convert bluebird based API so standard promise API in one PR, then remove bluebird with async/await in next PR. So difference is not too big

jedwards1211:
@sushantdhiman 
is this an okay conversion for `.tap`?  It's pretty ugly...but it's the only foolproof automated conversion besides using a wrapper function like `.then(wrapper(() => ...))`, which would be less amenable to automated async/await conversion.

```js
.tap(() => {
  if (x) return y
  return z
})
```
```js
.then(value => {
  if (x) return Promise.resolve(y).then(() => value)
  return Promise.resolve(z).then(() => value)
})
```

jedwards1211:
How about `.each(iterator)`, is this okay?

```js
.then(async iterable => {
  for (const element of iterable) {
    await iterator(element)
  }
})
```

papb:
> How about `.each(iterator)`, is this okay?
> 
> .then(async iterable => {
>   for (const element of iterable) {
>     await iterator(element)
>   }
> })

@jedwards1211 Actually this has to be

```js
.then(async iterable => {
  const elements = [];
  for (const _element of iterable) {
    const element = await _element;
    elements.push(element);
    await iterator(element);
  }
  return elements;
})
```

papb:
In fact I would like to suggest using [all the Sindre Sorhus amazing promise modules](https://github.com/sindresorhus/promise-fun) when applicable, since even with `async / await` they are 100% valid. Sindre wouldn't make a module whose behavior can be easily achieved with a simple `async / await` anyway :grimacing:

jedwards1211:
@papb actually in most cases the code wasn't using the finer details of `.each` like `index` argument and resolution value, so in my PR I just convered all those functions using `.each` to `async` and used `for..of` loops to iterate over the arrays

jedwards1211:
Also I'm still thinking I might be able to engineer a `jscodeshift` transform to batch convert all promise chains to `async`/`await`, and library functions would be more opaque to something like that than `for..of` loops

papb:
@jedwards1211 Ah, ok, if you took care about the other details of `.each`, then that's great!

> Also I'm still thinking I might be able to engineer a `jscodeshift` transform to batch convert all promise chains to `async`/`await`, and library functions would be more opaque to something like that than `for..of` loops

Well, if you say so! I never used `jscodeshift` so I cannot help much on that... If you think that is feasible, that is great.

And thank you very much for taking the time to help with this, Sequelize really needs this desperately.

By the way, I have just modernized [sequelize/umzug](github.com/sequelize/umzug) entirely as well... Hopefully in the near future the whole Sequelize codebase will be much better... :grimacing: Thanks!!

jedwards1211:
Ah cool!
Yeah in going through some of the old code, I noticed a lot of nested promise chains, which kind of defeats the purpose of what promises were created for...at first I thought whoever created the promise chains didn't get the point of promises, then I looked back in the history and saw that they had just converted old callback pyramids of doom to promise pyramids of doom 😅

jedwards1211:
Here was my `jscodeshift` transform for converting `.map` for example.  To get started you can just paste code into https://astexplorer.net/ to see what the AST looks like for writing a transform.  The TypeScript types are really good and help a lot with writing the transforms.

```ts
import { ASTPath, FileInfo, API, CallExpression } from 'jscodeshift'
import isInPromiseChain from './util/isInPromiseChain'

module.exports = function removeMap(
  fileInfo: FileInfo,
  api: API
): string | null | undefined | void {
  const j = api.jscodeshift

  const root = j(fileInfo.source)

  root
    .find(j.CallExpression, {
      callee: {
        type: 'MemberExpression',
        object: { type: 'Identifier', name: 'Promise' },
        property: { type: 'Identifier', name: 'map' },
      },
    })
    .replaceWith(
      (path: ASTPath<CallExpression>): CallExpression => {
        if (path.node.arguments.length !== 2) return path.node
        const [iterable, mapper] = path.node.arguments
        return j.template.expression`Promise.all(${iterable}.map(${mapper}))`
      }
    )

  root
    .find(j.CallExpression, {
      callee: {
        type: 'MemberExpression',
        property: { type: 'Identifier', name: 'map' },
      },
    })
    .replaceWith(
      (path: ASTPath<CallExpression>): CallExpression => {
        const {
          callee,
          arguments: [mapper],
        } = path.node
        if (
          !isInPromiseChain(path) ||
          callee.type !== 'MemberExpression' ||
          path.node.arguments.length !== 1
        )
          return path.node
        const { object } = callee
        if (object.type === 'Identifier' && object.name === 'Promise')
          return path.node
        return j.template
          .expression`${object}.then(value => Promise.all(value.map(${mapper})))`
      }
    )

  return root.toSource()
}
```

papb:
@jedwards1211 Awesome, thanks for sharing! I don't have time to look into it now, but will do in the future!

jedwards1211:
I just checked and there are are least 5000 `.then` and `.catch` calls in master, so I think it will definitely be worth trying to write a systematic codemod to process them

sushantdhiman:
Also use https://github.com/lebab/lebab

jedwards1211:
WIP on my asyncify transform here: https://github.com/codemodsquad/asyncify

PR #12088, opened 2020-04-09
feat(belongs-to-many): allow creation of paranoid join tables

### LINKED ISSUE #11991: Paranoid as option when creating many-to-many association
## Issue Description
I want to be able to create a belongs-to-many association that is paranoid. The association should create a join table that has a deletedAt column. 

I found this in `lib/associations/belongs-to-many.js`. Why does it say that a paranoid join table doesn't make sense?
```js
    if (typeof this.through.model === 'string') {
      if (!this.sequelize.isDefined(this.through.model)) {
        this.through.model = this.sequelize.define(this.through.model, {}, Object.assign(this.options, {
          tableName: this.through.model,
          indexes: [], //we don't want indexes here (as referenced in #2416)
          paranoid: false, // A paranoid join table does not make sense
          validate: {} // Don't propagate model-level validations
        }));
      } else {
        this.through.model = this.sequelize.model(this.through.model);
      }
    }
```

I was thinking something along the lines of a `paranoid` key in the options object and you would create the association like: 
```js
Customer.belongsToMany(PhoneNumber, {
  through: 'CustomerPhoneNumbers',
  as: 'SecondaryPhoneNumbers',
  paranoid: true
})
```

A workaround is to define the join table as a separate model and have the model be paranoid. 
But this requires potentially many models that are only required because we want to add a deletedAt column.

- [X] No. This issue is relevant to Sequelize as a whole.

Leithal3:
### Pull Request check-list

### Description of change

add check of through options to allow creation of paranoid join tables
 - default is still false

no breaking changes

Closes #11991


Leithal3:
Open to feedback. Happy to discuss reasoning.

Unsure how to fix the continuous-integration/appveyor/pr build - the log seems to indicate that a script file can't be found.

types/lib/associations/belongs-to-many.d.ts
```diff
@@ -22,7 +22,13 @@ export interface ThroughOptions {
   /**
```
sushantdhiman, line 22: Please add tests for these typing changes

Leithal3: I have added a test, please let me know if the test I have added is inadequate or in the wrong place.

sushantdhiman:
Please remove changes from lock file

PR #12089, opened 2020-04-09
fix(types): set and update function corrections

aecorredor:
### Pull Request check-list
 
### Description of change

I'm updating the types for two functions on `Model`: `set` and `update`. When the user calls those functions by passing in an object first (which is supposed to be the new data), the keys of those objects should be restricted to the model attributes and not include the public properties of the actual `Model` class. This is done by using `Partial<Omit<this, keyof Model>>` which just says: the keys of the object can be anything from `this` which is the specific model defined by the user, minus the properties defined on the actual `Model` class. All this simply translates to the model attributes.

- Also added some line breaks for readability (better for people that prefer 80 chars per line)

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -2601,7 +2601,7 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
    * @param options.reset Clear all previously set data values
    */
   public set<K extends keyof this>(key: K, value: this[K], options?: SetOptions): this;
-  public set(keys: Partial<this>, options?: SetOptions): this;
+  public set(keys: Partial<Omit<this, keyof Model>>, options?: SetOptions): this;
   public setAttributes<K extends keyof this>(key: K, value: this[K], options?: SetOptions): this;
   public setAttributes(keys: object, options?: SetOptions): this;
 
@@ -2654,8 +2654,15 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
   /**
    * This is the same as calling `set` and then calling `save`.
    */
-  public update<K extends keyof this>(key: K, value: this[K], options?: InstanceUpdateOptions): Promise<this>;
-  public update(keys: object, options?: InstanceUpdateOptions): Promise<this>;
+  public update<K extends keyof this>(
+    key: K, 
+    value: this[K], 
+    options?: InstanceUpdateOptions
+  ): Promise<this>;
+  public update(
+    keys: Partial<Omit<this, keyof Model>>, 
+    options?: InstanceUpdateOptions
+  ): Promise<this>;
 
   /**
    * Destroy the row corresponding to this instance. Depending on your setting for paranoid, the row will
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -2601,7 +2601,7 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
    * @param options.reset Clear all previously set data values
    */
   public set<K extends keyof this>(key: K, value: this[K], options?: SetOptions): this;
-  public set(keys: Partial<Omit<this, keyof Model>>, options?: SetOptions): this;
+  public set(keys: Partial<this>, options?: SetOptions): this;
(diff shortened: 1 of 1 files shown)


types/lib/model.d.ts
```diff
@@ -2654,8 +2654,15 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
   /**
    * This is the same as calling `set` and then calling `save`.
    */
-  public update<K extends keyof this>(key: K, value: this[K], options?: InstanceUpdateOptions): Promise<this>;
-  public update(keys: object, options?: InstanceUpdateOptions): Promise<this>;
+  public update<K extends keyof this>(
```
sushantdhiman, line 2657: What if you want to update an attribute which also a valid property of model? This will need reviews from @SimonSchick / @papb as I am a typescript noob

aecorredor: Shouldn't update just be for updating actual model data defined by the programmer?

sushantdhiman: These are valid keys from Model class

```js
[ 'getAssociations',
  'getAssociationForAlias',
  'Mixin',
  'default',
  '_setupHooks',
  'runHooks',
  'addHook',
  'removeHook',
  'hasHook',
  'hasHooks',
  'beforeValidate',
  'afterValidate',
  'validationFailed',
  'beforeCreate',
  'afterCreate',
  'beforeDestroy',
  'afterDestroy',
  'beforeRestore',
  'afterRestore',
  'beforeUpdate',
  'afterUpdate',
  'beforeSave',
  'afterSave',
  'beforeUpsert',
  'afterUpsert',
  'beforeBulkCreate',
  'afterBulkCreate',
  'beforeBulkDestroy',
  'afterBulkDestroy',
  'beforeBulkRestore',
  'afterBulkRestore',
  'beforeBulkUpdate',
  'afterBulkUpdate',
  'beforeFind',
  'beforeFindAfterExpandIncludeAll',
  'beforeFindAfterOptions',
  'afterFind',
  'beforeCount',
  'beforeAssociate',
  'afterAssociate',
  'beforeSync',
  'afterSync',
  'beforeBulkSync',
  'afterBulkSync',
  'beforeQuery',
  'afterQuery' ]
```

This PR will prevent users from updating an model with any attribute from list above. For e.g. now you can't have a Model with attribute `default` or `hasHooks`

aecorredor: Yeah I see what you are saying. Originally I thought this could actually be what we wanted since why would users be adding those attribute names to their models since they're really sequelize model functions. But, then again, they should be able to if they wanted to. So, in that sense you are right, I'll update this PR so it just limits the update keys type to `keys: Partial<this>`

aecorredor:
What do we think about this guys?

types/lib/model.d.ts
SimonSchick: You could probably use `Exclude<keyof this, keyof Model>`.

jedwards1211: Associations add dynamically named methods too, you would need a way to exclude those

jedwards1211: Associations add dynamically named methods too, you would need a way to exclude those

aecorredor: so from what @sushantdhiman mentioned, he doesn't really want to add that restriction. Basically he wants to let the user have whatever attribute names he wishes. So all I'm doing in this PR is updating the `update` signature to match what the `set` signature already had set up, which was like this:

`public set<K extends keyof this>(key: K, value: this[K], options?: SetOptions): this;`

It may not be perfect but it provides more type safety than what the current `update` signature has.

jedwards1211: I didn't realize that about `set`, that seems pretty bad, I mean Sequelize is probably not defined with static typing in mind, but the old DefinitelyTyped defs (don't know if they still exist, but I converted them to FlowTyped defs) actually used the `Model` type parameters to determine what attributes were available.  The current approach of the defs in this package seems destined for trouble...

PR #12122, opened 2020-04-19
refactor(model): asyncify methods

jedwards1211:

### DIFF
--- a/test/integration/model.test.js
+++ b/test/integration/model.test.js
@@ -1555,7 +1555,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
   });
 
   describe('restore', () => {
-    it('synchronously throws an error if the model is not paranoid', async function() {
+    it('rejects with an error if the model is not paranoid', async function() {
       await expect(this.User.restore({ where: { secretValue: '42' } })).to.be.rejectedWith(Error, 'Model is not paranoid');
     });
 
--- a/test/integration/model.test.js
+++ b/test/integration/model.test.js
@@ -1556,10 +1556,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
 
   describe('restore', () => {
     it('synchronously throws an error if the model is not paranoid', async function() {
-      expect(() => {
-        this.User.restore({ where: { secretValue: '42' } });
-        throw new Error('Did not throw synchronously');
-      }).to.throw(Error, 'Model is not paranoid');
+      await expect(this.User.restore({ where: { secretValue: '42' } })).to.be.rejectedWith(Error, 'Model is not paranoid');
     });
 
     it('restores a previously deleted model', async function() {
--- a/test/unit/model/update.test.js
+++ b/test/unit/model/update.test.js
@@ -46,12 +46,10 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       });
     });
 
-    it('can detect complexe objects', function() {
+    it('can detect complexe objects', async function() {
       const Where = function() { this.secretValue = '1'; };
 
-      expect(() => {
-        this.User.update(this.updates, { where: new Where() });
-      }).to.throw();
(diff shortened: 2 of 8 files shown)


sushantdhiman:
@jedwards1211 Is this automated by codemod?

jedwards1211:
@sushantdhiman I ran my codemod first and then cleaned up cases where the output was kind of messy

test/integration/model.test.js
```diff
@@ -1556,10 +1556,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
 
   describe('restore', () => {
     it('synchronously throws an error if the model is not paranoid', async function() {
```
sushantdhiman, line 1558: Remove `synchronously`

sushantdhiman: Can say `reject with an error if the model is not paranoid`

test/unit/model/update.test.js
```diff
@@ -46,12 +46,10 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
     });
 
-    it('can detect complexe objects', function() {
+    it('can detect complexe objects', async function() {
       const Where = function() { this.secretValue = '1'; };
 
-      expect(() => {
-        this.User.update(this.updates, { where: new Where() });
-      }).to.throw();
+      await expect(this.User.update(this.updates, { where: new Where() })).to.be.rejected;
```
papb, line 52: :open_mouth: I didn't know this worked! I thought it should be `to.be.eventually.rejected` or some other construction. Do you know where this fact is documented, @jedwards1211?

papb: On second thought I'm even more confused now. Wasn't the previous test detecting a synchronous throw? If so, shouldn't the new test throw instead of passing?

papb: (note to self: this is also done [here](https://github.com/sequelize/sequelize/pull/12125/files?diff=unified&w=1#diff-23797c4f52e452f7c44f18458cc223ed))

jedwards1211: @papb some methods in the API that normally return a Promise were designed to synchronously throw in some cases, which is bad, since you normally expect an async method to always return a Promise.  My refactoring got rid of the synchronous throws and replaced them with returning a rejected promise.

I don't know were `.to.be.rejected` is documented, but there's no way to synchronously inspect whether a builtin Promise is rejected anyway, so it wouldn't make sense for `eventually` to be needed in that case.  `eventually` is mainly intended for chaining assertions that are normally synchronous like `.to.eventually.deep.equal` etc.

The `chai-as-promised` docs say

> There are also a few promise-specific extensions (with the usual expect equivalents also available):

```
return promise.should.be.fulfilled;
return promise.should.eventually.deep.equal("foo");
return promise.should.become("foo"); // same as `.eventually.deep.equal`
return promise.should.be.rejected;
return promise.should.be.rejectedWith(Error); // other variants of Chai's `throw` assertion work too.
```

I've noticed that `.to.be.rejectedWith(...).and.have.deep.property('code').that.equal(...)` doesn't work though, for some reason it needs `eventually` to work.

papb: @jedwards1211 Ah, of course, I looked at the test refactor as if it was isolated, but of course you refactored the source code itself to always return a promise.

And thanks for the explanations!! Keep up the good work!

jedwards1211: Btw I think doing `.to.be.rejectedWith(...).and.X` performs assertion `X` on the returned promise, whereas `.to.be.eventually.rejectedWith(...).and.X` performs assertion `X` on the resolved `value` or rejection `reason`.

jedwards1211: In my tests for the new async queue I have a few
```js
.to.be.eventually.rejectedWith(ConnectionError, 'the connection was closed before this query could finish executing')
          .and.have.property('parent').that.instanceOf(AsyncQueueError);
```

PR #12128, opened 2020-04-19
refactor(dialects/abstract): asyncify methods

jedwards1211:

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -110,9 +110,7 @@ class ConnectionManager {
   async close() {
     // Mark close of pool
     this.getConnection = async function getConnection() {
-      throw new Error(
-        new Error('ConnectionManager.getConnection was called after the connection manager was closed!')
-      );
+      throw new Error('ConnectionManager.getConnection was called after the connection manager was closed!');
     };
 
     return await this._onProcessExit();
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -91,15 +91,15 @@ class ConnectionManager {
    * @private
    * @returns {Promise}
    */
-  _onProcessExit() {
+  async _onProcessExit() {
     if (!this.pool) {
-      return Promise.resolve();
+      return;
     }
 
-    return this.pool.drain().then(() => {
-      debug('connection drain due to process exit');
-      return this.pool.destroyAllNow();
-    });
+    await this.pool.drain();
+    debug('connection drain due to process exit');
+
+    return await this.pool.destroyAllNow();
   }
 
   /**
@@ -107,13 +107,15 @@ class ConnectionManager {
    *
    * @returns {Promise}
    */
-  close() {
+  async close() {
     // Mark close of pool
-    this.getConnection = function getConnection() {
-      return Promise.reject(new Error('ConnectionManager.getConnection was called after the connection manager was closed!'));
+    this.getConnection = async function getConnection() {
+      throw new Error(
+        new Error('ConnectionManager.getConnection was called after the connection manager was closed!')
+      );
     };
 
-    return this._onProcessExit();
+    return await this._onProcessExit();
   }
 
   /**
@@ -127,11 +129,10 @@ class ConnectionManager {
       this.pool = new Pool({
         name: 'sequelize',
         create: () => this._connect(config),
-        destroy: connection => {
-          return this._disconnect(connection)
-            .then(result => {
-              debug('connection destroy');return result;
-            });
+        destroy: async connection => {
+          const result = await this._disconnect(connection);
+          debug('connection destroy');
+          return result;
         },
         validate: config.pool.validate,
         max: config.pool.max,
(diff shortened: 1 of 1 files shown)


lib/dialects/abstract/connection-manager.js
```diff
@@ -91,29 +91,31 @@ class ConnectionManager {
...
    * @returns {Promise}
    */
-  close() {
+  async close() {
     // Mark close of pool
-    this.getConnection = function getConnection() {
-      return Promise.reject(new Error('ConnectionManager.getConnection was called after the connection manager was closed!'));
+    this.getConnection = async function getConnection() {
+      throw new Error(
+        new Error('ConnectionManager.getConnection was called after the connection manager was closed!')
```
sushantdhiman, line 114: Double `new Error` statement, remove one please

jedwards1211: this was a bug in my automated conversion of `return Promise.reject`, tomorrow I'll double-check my other PRs to make sure no instances of this mistake escaped testing

SimonSchick:
I wonder if this is wise, before doing this we should probably drop bluebird as the promise implementation to avoid mixing bluebird and native promises.

I did some work on this on another branch but never got around to finish it.

jedwards1211:
I've tracked down and replaced all of those cases I could find so far, though I guess it's true that it would be good to remove Bluebird entirely

SimonSchick:
@jedwards1211 if you are bored https://github.com/sequelize/sequelize/tree/chore/remove-bluebird

jedwards1211:
@SimonSchick @sushantdhiman 
The only remaining Bluebird API usage in `lib` on master that I found is:
```
lib/dialects/mssql/query.js Position { line: 117, column: 17, token: 714 } Promise.using
lib/sequelize.js Position { line: 1144, column: 4, token: 3317 } Promise.config
```
do you know what exactly the purpose of the `Promise.using` stuff for MSSQL is?  (`lib/dialects/mssql/resource-lock.js`)
I'm a bit confused what that code is doing.

The remaining Bluebird API usage in `test` is a bunch of `join()` calls, one case of `Promise.onPossiblyUnhandledRejection`, and a few `Promise.using` calls for MSSQL.

SimonSchick:
`.using` is similar to `using` in c#, it will call disposer function when exiting the scope, which is meant to 'cleanup' resources eg. close connections/file handles etc.
See https://docs.microsoft.com/en-us/dotnet/api/system.idisposable?view=netframework-4.8 and
http://bluebirdjs.com/docs/api/promise.using.html

jedwards1211:
@SimonSchick I sort of get that but is it just an overcomplicated method of ensuring two queries aren't run at once on the same connection?  It doesn't actually seem like it's being used to dispose of resources.  The `disposer` is just used to resolve the previous promise:

```js
  lock() {
    const lock = this.previous;
    let resolve;
    this.previous = new Promise(r => {
      resolve = r;
    });
    return lock.disposer(resolve);
  }
```

jedwards1211:
Seems like kind of a mess because this `ResourceLock` object which isn't actually a connection but a wrapper gets passed to `this.pool.destroy`, and goes on to `ConnectionManager._disconnect`, which passes it to `beforeDisconnect` and `afterDisconnect` hooks.

jedwards1211:
Seems to me like the cleanest replacement would be to wrap the `connection` using `Object.create` to add a `waitUntilIdle` method.

jedwards1211:
We could use [`p-limit`](https://github.com/sindresorhus/p-limit) to instead to prevent more than one query from running at once on a connection

jedwards1211:
all of these PRs should be ready to go

sushantdhiman:
@jedwards1211 / @SimonSchick we discussed a bit here https://github.com/sequelize/sequelize/issues/11859

Ideally we can remove Bluebird first and then convert to async methods, but for some cases intermediate code may become messy. So I think removing Bluebird API usage from core code and async conversion is faster.

We will remove bluebird after everything is converted, so there will be no promise mixing

SimonSchick:
@sushantdhiman should I bring my bluebird removal branch up to date?
Once that is done we can utility automated tools to perform async/await conversion in many places!

sushantdhiman:
@SimonSchick that branch contains other changes as well, also too big to review. It is better to handle this in small PRs.

papb:
Hi @SimonSchick I forgot you had that branch. Please see with @jedwards1211 if it is feasible to use part of your work in the work he is currently doing.

SimonSchick:
A lot of that is probably still usable.
The branch is kind of a mush (got very carried away).
You can definitely pick and choose individual files and do a quick sanity check and them use them, I wouldn't blindly copy things however.

The branch also still contains stuff like the hooks refactor (which I'd love to land in v6 as it DOES do a lot of streamlining).

papb:
> The branch also still contains stuff like the hooks refactor (which I'd love to land in v6 as it DOES do a lot of streamlining).

Awesome, I agree the sequelize codebase desperately needs lots of refactoring, so that's great. Let's wait for @jedwards1211 to finish his super work on async/await and then we can look at these other refactors

jedwards1211:
@SimonSchick it's not worth it I'm afraid, as a lot of PRs I've made that have been merged have already eliminated nearly all of the bluebird usage.

jedwards1211:
Oh I didn't see what you said about hooks refactoring though.  How heavily does it depend on the changes to remove bluebird in your branch?

PR #12136, opened 2020-04-20
refactor(mssql/connection-manager): eliminate Promise.using/promise.disposer calls

jedwards1211:
BREAKING CHANGE: when using MSSQL dialect the connection is once again a `tedious` `Connection` instance instead of a wrapper.  `tedious`' `Connection` is extended to add an `enqueue` method for performing queries one at a time.


jedwards1211:
@sushantdhiman why does the precommit hook stash my changes if linting fails?  It's annoying, I wish it would just fail the commit and not blow away my local changes.  (Unstashing is a bit of a hassle because I have some jscodeshift scripts in the dir that I'm not committing)

sushantdhiman:
@jedwards1211 Just a convenience in some cases, You may add those extra folders to `.gitignore` which we can cleanup later on

SimonSchick:
Can't we just store a single promise chain instead of using a dependency module?

Something along the lines of:

```
class Something {
   constructor() { this.queue = Promise.resolve(); }
   async someMethod() {
      await this.queue.then(() => ourLogic());
  }
}
```

sushantdhiman:
> Can't we just store a single promise chain instead of using a dependency module?
> 
> Something along the lines of:
> 
> ```
> class Something {
>    constructor() { this.queue = Promise.resolve(); }
>    async someMethod() {
>       await this.queue = this.queue.then(() => ourLogic());
>   }
> }
> ```

I think we should go with `p-limit`, It is well tested module with just one dependency

sushantdhiman:
@jedwards1211 Please update your npm version `npm i -g npm@latest` and do npm install once again to refresh `package-lock.json`.

sushantdhiman:
Travis is down atm, https://travis-ci.org/. Will take a look at this PR in my evening time. And thanks for all this refactor work @jedwards1211

papb:
> BREAKING CHANGE: when using MSSQL dialect the connection is once again a `tedious` `Connection` instance instead of a wrapper. `tedious`' `Connection` is extended to add an `enqueue` method for performing queries one at a time.

Hi @jedwards1211 can you please clarify how end users are impacted by this? Since this is a breaking change, what steps should they take to upgrade?

jedwards1211:
@papb well, I didn't see anywhere in the docs talk about how the `connection` object was different for MSSQL than other dialects, so if anyone's using it they would have had to figure that out for themself.  I assume it's not common for people to use `connection` from the sequelize pool manually, but if they are this would catch them off guard if we don't mention it.

jedwards1211:
@SimonSchick I was tempted to do something like that, but I wonder what it would do to stack traces/heap usage.  I assume modern VMs probably support endless promise chains like this somehow, but I haven't researched it.

jedwards1211:
Hmmm maybe we should think about whether to use `p-limit` for a bit: https://github.com/sindresorhus/p-limit/issues/22

Looking at its code I'm also confused about how we would get promise rejections out of `p-limit`.

jedwards1211:
Okay I confirmed that getting rejections is not a problem, I forgot that passing a rejected promise to `resolve` works: 
```
> new Promise(resolve => { resolve(Promise.reject(new Error('test'))) }).then(console.log, console.error)
Promise { <pending> }
> Error: test
    at repl:1:49
    at new Promise (<anonymous>)
    at repl:1:1
```

jedwards1211:
@sushantdhiman argh, no idea why a SQLite test is failing

SimonSchick:
I still disagree with adding a dependency for just a single use-case like that.
Pro:
 - Code clarify, maybe?

Contra:
 - It add effectively adds 2 dependencies, package bloat is a thing and we should aim to have a minimal footprint.
- We aren't even taking advantage of the actual limiting/concurrency regulation ability and taking a relatively big hit in overhead.
- A promise chain is a well known usage of promises that doesn't require any more code than our *usage* `p-limit` does (line count will be less or equal).
- Test coverage of `p-limit` is irrelevant here as all affected code tested by us already.

I might be biased here since I'm usually more inclined to create my own utilities (and test them) vs. taking a third party dependency and potential liability (security, maintenance etc.).

jedwards1211:
Do we know for sure that an endless promise chain doesn't cause stack trace or memory usage problems?  I'm open to it, I would just want to know what happens over time

jedwards1211:
I assume this issue is probably fixed but in the early days of the spec it was apparently an unsafe pattern, so I need to find evidence that it's no longer a problem: https://alexn.org/blog/2017/10/11/javascript-promise-leaks-memory.html

jedwards1211:
Okay AFAICT from look at comments about this the danger is only if you have a recursive call in a handler that prevents the chain from ever resolving, unlike the latter you suggested.

jedwards1211:
at minimum we would have to use
```js
  async enqueue(cb) {
    const result = this.queue.then(cb);
    this.queue = result.catch(() => {});
    return result;
  }
```

SimonSchick:
Are you sure you don't want to handle errors in callbacks?
Seems dangerous and should at least emit a warning of some sort.

jedwards1211:
`Query.run` has to return a rejected promise if something goes wrong.

SimonSchick:
We should still signal that an error occured?

jedwards1211:
@SimonSchick why don't you believe me?

At the end of `sequelize.query`:
```js
      try {
        await this.runHooks('beforeQuery', options, query);
        await checkTransaction();
        return await query.run(sql, bindParameters);
      } finally {
        await this.runHooks('afterQuery', options, query);
        if (!options.transaction) {
          await this.connectionManager.releaseConnection(connection);
        }
      }
```

The `query.run` call has to do
```js
  run(sql, parameters) {
    return this.connection.enqueue(() => this._run(this.connection, sql, parameters));
  }
```

Hence the queuing mechanism has to return rejection errors to the caller.

SimonSchick:
Alright then, sounds good.
I've become very weary of empty catch blocks, I've wasted days of my life debugging swallowed errors 😞

papb:
Hi, I didn't look deeply into the discussion, but would using `p-limit` prevent this confusion involving an empty catch block?

jedwards1211:
Actually I wasn't thinking anyway, as the catch block in my example *would* undesirably prevent unhandled rejections.  I'm guessing `p-limit` does require the caller to handle rejections, will confirm.  If we were gonna do it our own way it would probably need to be something like

```js
  async enqueue(cb) {
    return new Promise((resolve, reject) => {
      this.queue = this.queue.finally(() => cb().then(resolve, reject));
    });
  }
```

jedwards1211:
@papb actually I'm starting to agree with @SimonSchick that we should use our own implementation, because for one, I'm guessing that we need to fail any queued operations whenever the connection is closed, but `p-limit`'s `clearQueue` just empties the queue, and never resolves or rejects any of the promises that were pending.

jedwards1211:
@SimonSchick 
> I've wasted days of my life debugging swallowed errors
Yeah I totally agree that we don't want that.  As long as we ensure that an error will cause an unhandled rejection if the user doesn't catch it, we won't have that problem

jedwards1211:
Okay this works:
```js
const queue1 = {
  queue: Promise.resolve(),
  enqueue(cb) {
    return new Promise((resolve, reject) => {
      this.queue = this.queue.then(
        () => cb().then(resolve, reject)
      )
    }) 
  },
}

queue1.enqueue(async () => { console.log('resolve 1') })
queue1.enqueue(async () => { throw new Error('reject 2') }).catch(e => console.error(e.stack))
queue1.enqueue(async () => { throw new Error('reject 3') })
queue1.enqueue(async () => { console.log('resolve 4') })
```
Output
```
queue1
resolve 1
Error: reject 2
    at cb (/Users/andy/queue.js:14:36)
    at /Users/andy/queue.js:6:15
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
resolve 4
(node:13865) UnhandledPromiseRejectionWarning: Error: reject 3
    at cb (/Users/andy/queue.js:15:36)
    at /Users/andy/queue.js:6:15
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(node:13865) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 3)
(node:13865) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

SimonSchick:
Isn't the `new Promise` redundant?
This could be:
```js
const queue1 = {
  queue: Promise.resolve(),
  enqueue(cb) {
      return this.queue = this.queue.then(cb);
  },
}
```

Getting the same output when running your test cases locally.

jedwards1211:
No it's absolutely not redundant, there's no way you're getting the same output.

For one `this.queue = this.queue.then(cb)` hangs the queue forever if any callback rejects. For another it generates a lot more unhandled promise rejections.

I tried it:

```js
const queue1 = {
  queue: Promise.resolve(),
  enqueue(cb) {
    return new Promise((resolve, reject) => {
      this.queue = this.queue.then(() => cb().then(resolve, reject))
    })
  },
}

queue1.enqueue(async () => { console.log('A: resolve 1') })
queue1.enqueue(async () => { throw new Error('A: reject 2') }).catch(e => console.error(e.stack))
queue1.enqueue(async () => { throw new Error('A: reject 3') })
queue1.enqueue(async () => { console.log('A: resolve 4') })
  .then(() => {
    console.log()

    const queue2 = {
      queue: Promise.resolve(),
      enqueue(cb) {
        return (this.queue = this.queue.then(cb))
      },
    }

    queue2.enqueue(async () => { console.log('B: resolve 1') })
    queue2.enqueue(async () => { throw new Error('B: reject 2') }).catch(e => console.error(e.stack))
    queue2.enqueue(async () => { throw new Error('B: reject 3') })
    queue2.enqueue(async () => { console.log('B: resolve 4') })
  })
```
```
$ node queue.js 
A: resolve 1
Error: A: reject 2
    at /Users/andy/queue.js:11:36
    at /Users/andy/queue.js:5:42
A: resolve 4

B: resolve 1
Error: B: reject 2
    at /Users/andy/queue.js:25:40
(node:44356) UnhandledPromiseRejectionWarning: Error: A: reject 3
    at /Users/andy/queue.js:12:36
    at /Users/andy/queue.js:5:42
(node:44356) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 3)
(node:44356) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:44356) UnhandledPromiseRejectionWarning: Error: B: reject 2
    at /Users/andy/queue.js:25:40
(node:44356) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 5)
```

SimonSchick:
Seems like I managed to use use the `diff` command wrong when comparing the outputs, that is my bad, not having a good day 😅

sushantdhiman:
> I still disagree with adding a dependency for just a single use-case like that.
> .........

I agree. Least dependencies we have the best. I just assumed `p-limit` was good enough. Ss @jedwards1211 noted `p-limit` is resolving / rejecting errors when queue is cleared, that module is useless for our use-case

> I'm guessing that we need to fail any queued operations whenever the connection is closed, but p-limit's clearQueue just empties the queue, and never resolves or rejects any of the promises that were pending.

Hmm, that is not good. I suppose https://github.com/sindresorhus/p-limit/issues/25 should fix this.

sushantdhiman:
> Okay this works and doesn't require an empty `catch` block:
> 
> ```js
> const queue1 = {
>   queue: Promise.resolve(),
>   enqueue(cb) {
>     return new Promise((resolve, reject) => {
>       this.queue = this.queue.then(
>         () => cb().then(resolve, reject)
>       )
>     }) 
>   },
> }
> 
> queue1.enqueue(async () => { console.log('resolve 1') })
> queue1.enqueue(async () => { throw new Error('reject 2') }).catch(e => console.error(e.stack))
> queue1.enqueue(async () => { throw new Error('reject 3') })
> queue1.enqueue(async () => { console.log('resolve 4') })
> ```
> 
> Output
> 
> ```
> queue1
> resolve 1
> Error: reject 2
>     at cb (/Users/andy/queue.js:14:36)
>     at /Users/andy/queue.js:6:15
>     at processTicksAndRejections (internal/process/task_queues.js:93:5)
> resolve 4
> (node:13865) UnhandledPromiseRejectionWarning: Error: reject 3
>     at cb (/Users/andy/queue.js:15:36)
>     at /Users/andy/queue.js:6:15
>     at processTicksAndRejections (internal/process/task_queues.js:93:5)
> (node:13865) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 3)
> (node:13865) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
> ```

@jedwards1211 We can use this queue, looks simple enough. Just add some sanity test for various cases. Like clearing queue case should rejects (or resolve?) pending operations

jedwards1211:
@SimonSchick ah yeah that kinda day, sorry I was kinda stern in my response.  You've definitely encouraged me to think more carefully about how this needs to work :)

Here's the full implementation I have now, including rejecting pending promises when the connection is closed.  Will see if I can figure out how to test it tomorrow.
```js
class AsyncQueue {
  constructor() {
    this.previous = Promise.resolve();
    this.closed = false;
    this.rejectCurrent = () => {};
  }
  close() {
    this.closed = true;
    this.rejectCurrent(new Error('the connection was closed before this query could finish executing'));
  }
  enqueue(cb) {
    return new Promise((resolve, reject) => {
      this.previous = this.previous.then(
        () => {
          this.rejectCurrent = reject;
          if (this.closed) {
            return reject(new Error('the connection was closed before this query could be executed'));
          }
          return cb().then(resolve, reject);
        }
      );
    });
  }
}

module.exports = AsyncQueue;
```

jedwards1211:
Okay I'm done writing the new queue and tests for it

lib/dialects/mssql/AsyncQueue.js
```diff
@@ -0,0 +1,31 @@
...
+
+class AsyncQueue {
+  constructor() {
+    this.previous = Promise.resolve();
+    this.closed = false;
+    this.rejectCurrent = () => {};
+  }
+  close() {
+    this.closed = true;
+    this.rejectCurrent(new Error('the connection was closed before this query could finish executing'));
```
sushantdhiman, line 11: This should be a connection error, add a new error [here](https://github.com/sequelize/sequelize/tree/master/lib/errors/connection). Name suggestion `connection-queue-error`.

lib/dialects/mssql/AsyncQueue.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
```
sushantdhiman, line 1: Filename `AsyncQueue.js` => `async-queue.js`

lib/dialects/mssql/AsyncQueue.js
```diff
@@ -0,0 +1,31 @@
...
+    // However, this ensures that this.previous will never be a rejected promise so the queue will
+    // always keep going, while still communicating rejection from callback to the user.
+    return new Promise((resolve, reject) => {
+      this.previous = this.previous.then(
+        () => {
+          this.rejectCurrent = reject;
+          if (this.closed) {
+            return reject(new Error('the connection was closed before this query could be executed'));
+          }
+          return callback().then(resolve, reject);
```
sushantdhiman, line 24: This assumes `callback` returns a promise, rename `callback` to `promiseCallback` other other suitable name

test/integration/dialects/mssql/query-queue.test.js
```diff
@@ -33,5 +34,93 @@ if (dialect.match(/^mssql/)) {
...
+
+    it('requests that reject should not affect future requests', async function() {
+      const User = this.User;
+
+      await expect(this.sequelize.transaction(async t => {
+        await Promise.all([
+          expect(User.create({
+            username: new Date()
+          })).to.be.rejected,
+          expect(User.findOne({
```
sushantdhiman, line 46: Add delay here so we are really sure this promise is executed after the first one

jedwards1211: Actually in this case I could just await the queries in series

lib/dialects/mssql/AsyncQueue.js
jedwards1211: would `connection-closed-error` be fine? What happens in other dialects when the connection is closed while a query is in flight?

jedwards1211: this is awkward: `ConnectionError` expects a `parent` error that was thrown by the dialect-specific connection, but in this case, we're throwing the error ourselves so there's no `parent` to wrap.  At the moment I made `ConnectionClosedError` inherit from `BaseError` instead

jedwards1211: In `connection-error.js`:

lib/errors/connection/connection-closed-error.js
```diff
@@ -0,0 +1,16 @@
...
+const BaseError = require('../base-error');
+
+/**
+ * Thrown when a connection to a database is closed while an operation is in progress
+ */
+class ConnectionClosedError extends BaseError {
+  constructor(message) {
+    super(message);
+    this.name = 'SequelizeConnectionClosedError';
+    Error.captureStackTrace(this, this.constructor);
```
SimonSchick, line 12: Again, this is redundant with node 10+

jedwards1211: It's in the other error classes so maybe we should just wait for another PR to remove them all?

SimonSchick: All I'm saying there isn't really a need to add more, feel free you create a PR for the others though :)

lib/dialects/mssql/AsyncQueue.js
sushantdhiman: In that case we can have an `async-queue-error`, define that error in this file. Use that as parent error in generic connection error thrown by dialect

lib/errors/connection/connection-closed-error.js
jedwards1211: cool, I left it out when refactoring the error to what Sushant requested

PR #12140, opened 2020-04-20
remove bluebird

jedwards1211:


jedwards1211:
Sigh I don't even understand what the random test failures are about
I really wish I could go back in time and somehow prevent Petka Antonov from creating Bluebird

jedwards1211:
@sushantdhiman @papb @SimonSchick if y'all have any hunches why some of the tests are failing here let me know...gonna be tricky to debug

lib/dialects/mssql/connection-manager.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 const AbstractConnectionManager = require('../abstract/connection-manager');
-const ResourceLock = require('./resource-lock');
-const Promise = require('../../promise');
+const pLimit = require('p-limit');
```
SimonSchick, line 4: See dicsussion here: https://github.com/sequelize/sequelize/pull/12136
This will lead to conflicts.

lib/errors/aggregate-error.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Validation Error. Thrown when the sequelize validation has failed. The error contains an `errors` property,
+ * which is an array with 1 or more ValidationErrorItems, one for each validation that failed.
+ *
+ * @param {string} message Error message
+ * @param {Array} [errors] Array of ValidationErrorItem objects describing the validation errors
+ *
+ * @property errors {ValidationErrorItems[]}
+ */
+class AggregateError extends BaseError {
```
SimonSchick, line 16: We should probably make a breaking change here and not extend from error AND array and instead pass a readonly array of errors into the constructor and make them publicly available to the consumer eg.

This ensures only lazy initialisation which is usually better for performance, it also makes the API easier to model in TS.

```js
class AggregateError extends BaseError {
  constructor(message, errors) { super(message); this.errors = errors; } ... }
```

lib/errors/aggregate-error.js
```diff
@@ -0,0 +1,57 @@
...
+  constructor(message) {
+    super(message);
+    this.name = 'AggregateError';
+    Error.captureStackTrace(this, this.constructor);
+  }
+
+  toString() {
+    let indent = Array(level * 4 + 1).join(' ');
```
SimonSchick, line 24: Modern JS can use `string.repeat`

lib/errors/aggregate-error.js
```diff
@@ -0,0 +1,57 @@
...
+      ret += `${str }\n`;
+    }
+    level--;
+    return ret;
+  }
+}
+
+module.exports = AggregateError;
+
+const methods = 'join pop push shift unshift slice filter forEach some every map indexOf lastIndexOf reduce reduceRight sort reverse'.split(' ');
```
SimonSchick, line 44: As per above comment, we should probably not done this :)

lib/errors/aggregate-error.js
```diff
@@ -0,0 +1,57 @@
...
+      }
+      str = lines.join('\n');
+      ret += `${str }\n`;
+    }
+    level--;
+    return ret;
+  }
+}
+
+module.exports = AggregateError;
```
SimonSchick, line 42: Can we make all internal exports use `exports.NAME = NAME`? This is easier to model in es modules.

SimonSchick:
I'm not sure about the tests, we had a bunch of issues with flaky tests in the past, in fact I managed to fix a bunch of those.

sushantdhiman:
@SimonSchick I think this PR should just cover bluebird removal, other changes can come in different PRs

lib/errors/aggregate-error.js
jedwards1211: all true, I basically just copied over and lightly adapted Bluebird's AggregateError

lib/dialects/mssql/connection-manager.js
jedwards1211: yeah I had started this branch from the remove using branch, I'll have to rebase it

SimonSchick:
@jedwards1211 bluebird was a good library for the time it was created, it was powerful and effective, things become less so over time :)

lib/errors/aggregate-error.js
papb: Hi, I didn't read the discussion in-depth, but since you're talking about `AggregateError`, what if we just use [`sindresorhus/aggregate-error`](https://github.com/sindresorhus/aggregate-error)?

Yes, I know, I am a fan of his repositories...

SimonSchick: That comes with at least additional 2 extra dependencies and cannot inherit from the sequelize error base.

lib/errors/aggregate-error.js
jedwards1211: @SimonSchick I got rid of this anyway, but it's not really hard with `jscodeshift` to convert all `module.exports = ` statements to `export default` statements.

jedwards1211: Oh wait, wrong PR, I haven't gotten rid of this error

package.json
```diff
@@ -30,13 +30,13 @@
...
   "license": "MIT",
   "dependencies": {
-    "bluebird": "^3.7.1",
     "debug": "^4.1.1",
     "dottie": "^2.0.0",
     "inflection": "1.12.0",
     "lodash": "^4.17.15",
     "moment": "^2.24.0",
     "moment-timezone": "^0.5.21",
+    "p-limit": "^2.3.0",
```
sushantdhiman, line 39: Remove

lib/errors/aggregate-error.js
SimonSchick: @jedwards1211 Please avoid default exports, see https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/ as an example why.

SimonSchick:
If you are not comfortable with touching the typings I can take care of them after this was merged, so feel free to omit that.

package.json
jedwards1211: whoops

jedwards1211:
@SimonSchick actually, I don't think any typing changes are needed after all?  Unless @sushantdhiman do you want me to re-export `AsyncQueueError` in `lib/errors.js`?

SimonSchick:
I think it makes sense to export it from there, you want users to be able to catch them properly :)

lib/errors/aggregate-error.js
jedwards1211: sigh, that would only complicate the conversion of everything to `export default`

jedwards1211: Or are y'all against `import default` and `export default` as well?

SimonSchick: export/import should generally be avoided, it leads to incosistent naming across modules, see https://google.github.io/styleguide/jsguide.html#named-vs-default-exports

jedwards1211: # 🙄

jedwards1211: I hope you're at least not going to make users convert
```js
import Sequelize from 'sequelize'
// or
```
to
```js
import { Sequelize } from 'sequelize'
// 
const { Sequelize } = require('sequelize')
```

jedwards1211: And again, it we keep the code more consistent now, it's easier to automatically convert to whatever new form we want later

SimonSchick: > The only nitpicky rules I respect are ones that can be enforced through scripts with an automated fix, because that doesn't waste developers' time. 

Point taken, I will around if there are eslint rules for this in the future.

jedwards1211: apparently this does exist already, can't tell if it has an autofix or not?
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-named-as-default.md

SimonSchick: Let's deal with this in another PR, there are opportunities to streamline exports/imports across the codebase.

jedwards1211: Yeah, I don't mind helping convert them automatically to either default or named.  (with the exception of the main export, I think we should leave that as is to minimize disruption for users)

jedwards1211:
@sushantdhiman I think there's a race condition somewhere in the SQLite dialect... I can't repro the only error that's failing CI now, yet I get different errors locally

lib/errors/aggregate-error.js
jedwards1211: Man I feel bad for being such a grouch about this, I see you weren't having a great day with trying to debug the test failures, I imagine I made it worse 😕. Let me know what other stuff I can help with.

SimonSchick: Nah, you have a fair point with me being nit-picky about things that aren't in the lint configs trying to micro-manage these things in PRs isn't particularly productive and can be frustrating.

SimonSchick:
@jedwards1211 do you want me to take a look at your sqlite changes?
What test(s) are failing or are flaky?

jedwards1211:
Well, let's see if the fixes you merged today fix the issue

SimonSchick:
@jedwards1211 it does not address fix the `close connection and remove journal and wal files` task, I pulled your branch and rebased it with on master.

This is a shot in the dark but it looks like it's a race condition caused by sqlite, adding as much as a 5ms delay to the test fixes it, we might've never noticed it because bluebird was slow 🤔

SimonSchick:
I re-wrote the entire test file to ensure no funny business was going on with `jetpack-fs` (why was this even added?) as well code is here:

```js
'use strict';

const fs = require('fs');
const chai = require('chai');
const expect = chai.expect;
const Support = require('../../support');
const DataTypes = require('../../../../lib/data-types');
const { join } = require('path');
const delay = require('delay');
const rimraf = require('rimraf');

const directoryName = `${Math.random()}_test_directory`;
const nestedFileName = join(__dirname, directoryName, 'subdirectory', 'test.sqlite');

if (dialect === 'sqlite') {
  describe('[SQLITE Specific] Connection Manager', () => {
    after(() => {
      rimraf.sync(directoryName);
    });

    it('close connection and remove journal and wal files', async function() {
      const sequelize = Support.createSequelizeInstance({
        storage: nestedFileName
      });
      const User = sequelize.define('User', { username: DataTypes.STRING });

      await sequelize.query('PRAGMA journal_mode = WAL');
      await User.create({ username: 'user1' });
      await sequelize.transaction(transaction => User.create({ username: 'user2' }, { transaction }));
      expect(fs.existsSync(nestedFileName)).to.be.equal(true);
      expect(fs.existsSync(`${nestedFileName}-shm`), 'shm file should exists').to.be.equal(true);
      expect(fs.existsSync(`${nestedFileName}-wal`), 'wal file should exists').to.be.equal(true);

      await sequelize.close();

      expect(fs.existsSync(nestedFileName)).to.be.equal(true);
      await delay(5);
      expect(fs.existsSync(`${nestedFileName}-shm`), 'shm file exists').to.be.false;
      expect(fs.existsSync(`${nestedFileName}-wal`), 'wal file exists').to.be.false;

      await this.sequelize.query('PRAGMA journal_mode = DELETE');
    });

    it('automatic path provision for `options.storage`', () => {
      return Support.createSequelizeInstance({ storage: nestedFileName })
        .define('User', { username: DataTypes.STRING })
        .sync({ force: true }).then(() => {
          expect(fs.existsSync(nestedFileName)).to.be.equal(true);
        });
    });
  });
}
```

SimonSchick:
@sushantdhiman do you know why 

`await this.sequelize.query('PRAGMA journal_mode = DELETE');` is called at all?
`this.sequelize` is never otherwise used in this context since the test creates it's own stuff.

sushantdhiman:
`await this.sequelize.query('PRAGMA journal_mode = DELETE');` statement is there so it must be executing :)

I think this has something to do with changing journal mode, which is globally applied. So this is a reset. Can move this to `afterEach`

SimonSchick:
@sushantdhiman The test uses an entirely different database than `this.sequelize` how would this be applied globally?
I'm no expert in sqlite but this line seems redundant.

SimonSchick:
Did some digging into node-sqlite, it seems the calls to sqlite3_close are performed correctly, I have no idea how these files would still exists, but I can replicate the behaviour on my local machine which is worrying.

sushantdhiman:
@jedwards1211 / @SimonSchick test was failing because SQLITE wasn't check-pointed, so on connection close it still need to commit changes from WAL file to main database. I guess there was a delay related to that, which was not caught earlier due to slow Bluebird I suppose.

Also removed `verbose()` which should [improve SQLITE performance  ](https://github.com/mapbox/node-sqlite3/wiki/Debugging)

jedwards1211:
Oh wow I didn't expect this to be merged yet, thank you guys for figuring those issues out.  It probably would have been hard for me to figure out.
I'll make PRs to asyncify the sqlite dialect and tests soon.  (I think that's all that remains?)

SimonSchick:
@jedwards1211 

image

Most (not all) should probably be updated as well, some are doc strings :)

jedwards1211:
there may still be some cases in test like

```js
Promise.all([
  User.update({...}).then(() => User.findAll()),
  Post.update({...}).then(() => Post.findAll()),
]);
```
Which we should leave because the alternative is worse:

```js
Promise.all([
  (async () => {
    await User.update({...});
    return User.findAll();
  })(),
  (async () => {
    await Post.update({...});
    return Post.findAll();
  })(),
]);
```

SimonSchick:
1 is subjectively better and easier to read.

jedwards1211:
agreed

PR #12153, opened 2020-04-22
test: debug transaction leak

SimonSchick:
Dummy PR to trigger CI, please ignore

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -644,11 +644,11 @@ class Sequelize {
       const query = new this.dialect.Query(connection, this, options);
 
       try {
-        await this.runHooks('beforeQuery', options, query);
-        await checkTransaction();
+        await this.runHooks('beforeQuery', options, query, sql);
+        checkTransaction();
         return await query.run(sql, bindParameters);
       } finally {
-        await this.runHooks('afterQuery', options, query);
+        await this.runHooks('afterQuery', options, query, sql);
         if (!options.transaction) {
           await this.connectionManager.releaseConnection(connection);
         }
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -645,7 +645,7 @@ class Sequelize {
 
       try {
         await this.runHooks('beforeQuery', options, query, sql);
-        await checkTransaction();
+        checkTransaction();
         return await query.run(sql, bindParameters);
       } finally {
         await this.runHooks('afterQuery', options, query, sql);
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -1100,6 +1100,7 @@ class Sequelize {
     }
 
     const transaction = new Transaction(this, options);
+    this.runHooks('transactionCreated', transaction);
 
     if (!autoCallback) {
       await transaction.prepareEnvironment(false);
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -645,7 +645,7 @@ class Sequelize {
 
       try {
         await this.runHooks('beforeQuery', options, query, sql);
-        checkTransaction();
+        await checkTransaction();
         return await query.run(sql, bindParameters);
       } finally {
         await this.runHooks('afterQuery', options, query, sql);
--- a/test/integration/support.js
+++ b/test/integration/support.js
@@ -34,7 +34,7 @@ afterEach(function() {
     return;
   }
   let msg = `Expected 0 running queries. ${runningQueries.size} queries still running in ${this.currentTest.fullTitle()}\n`;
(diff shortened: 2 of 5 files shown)


SimonSchick:
The single time I want things to fail, they don't....

SimonSchick:
@sushantdhiman I ran this about 20 times now, no failures, any ideas?

SimonSchick:
@sushantdhiman any objections of pulling these changes into master for a bit?
I'm having 0 luck reproducing errors at this time, the 'api' I added should probably not to public hence not docs.

jedwards1211:
@SimonSchick this is related to the CI test failures in my remove bluebird branch right?  I'm kinda mystified by the queries still running errors.

SimonSchick:
@jedwards1211 correct, so far it seems that that some (empty) transaction gets committed outside of a test, but when trying to trace it (see new Error().stack), it works fine.

I'm trying to narrow down what is going on here by enforcing builds to obey package-lock.json and updating dev dependencies (just in case).

Starting to think the way I try to trace things might cause the error not to occur due to timing changes 😞

jedwards1211:
Yeah such are heisenbugs 🙁
I was looking through the code, especially SQLite last night, to see if I would notice anything being unawaited, but so far there hasn't been any smoking gun.

lib/sequelize.js
```diff
@@ -644,11 +644,11 @@ class Sequelize {
       const query = new this.dialect.Query(connection, this, options);
 
       try {
-        await this.runHooks('beforeQuery', options, query);
-        await checkTransaction();
+        await this.runHooks('beforeQuery', options, query, sql);
+        checkTransaction();
```
eseliger, line 648:
```suggestion
        await checkTransaction();
```

just trying to see if the extra tick in between will cause the error to come back ...

test/integration/support.js
```diff
@@ -2,14 +2,29 @@
...
+      if (queryList) {
+        queryList.push(sql);
+      } else {
+        runningTransactions.set(options.transaction.id, [sql]);
+      }
+    }
+  });
+  this.sequelize.addHook('afterQuery', (options, query, sql) => {
     runningQueries.delete(query);
+    if (options.transaction && sql.includes('COMMIT')) {
```
eseliger, line 25: should this better be:?

```suggestion
    if (options.transaction && ['COMMIT', 'ROLLBACK'].includes(sql)) {
```

lib/sequelize.js
```diff
@@ -644,11 +644,11 @@ class Sequelize {
       const query = new this.dialect.Query(connection, this, options);
 
       try {
-        await this.runHooks('beforeQuery', options, query);
+        await this.runHooks('beforeQuery', options, query, sql);
         await checkTransaction();
```
eseliger, line 648:
```suggestion
       checkTransaction();
```

jedwards1211:
I was looking at `retry-as-promised`:

```js
    Promise.resolve(callback({ current: options.$current }))
      .then(resolve)
      .then(function() {
        if (timeout) clearTimeout(timeout);
        if (backoffTimeout) clearTimeout(backoffTimeout);
      })
      .catch(function(err) {
        if (timeout) clearTimeout(timeout);
        if (backoffTimeout) clearTimeout(backoffTimeout);
```
This is probably a dumb idea but I don't guess there's a way the timeout could fire again between `.then(resolve)` and what comes after it?

SimonSchick:
@jedwards1211 I'm unsure, might be worthwhile to introduce a test case directly against that library.
I will merge this now (and revert it later) to see if we can repro this mess in other branches.

sushantdhiman:
Why was this merged, it just pollute commit history. If something can't be replicated in this branch how merging in master would replicate that issue

jedwards1211:
@SimonSchick mentioned wanting to try to repro in other branches, so I guess the idea is to pull this into those branches.  I think there are ways to pull in changes from this branch since master into another branch, but I forget how to do it...

sushantdhiman:
If this branch can't replicate that failure even once how it will help in other branches

SimonSchick:
Me and @eseliger were both pretty baffled why we couldn't replicate the error with those changes as functionality nothing was changed.

We came up with the idea to have this check run on more builds and hopefully come across one where this condition occurs rather than mindlessly hammering travis over and over and wasting a LOT of time in the process.

It's not 100% not ideal but doesn't really affect anything outside of CI.

sushantdhiman:
@SimonSchick This PR worked by the way. It seems https://github.com/sequelize/sequelize/blob/master/test/integration/cls.test.js#L67 is culprit

SimonSchick:
Cool, will file a PR.

PR #12154, opened 2020-04-23
ci(install): use package lock for install

SimonSchick:
### Pull Request check-list

### Description of change

Use package-lock for installs.
I also ran the `dedupe command` prior to this.

This will:
 - Speed up installs.
 - Allow reproducable builds.

Side effects:
Apparently `Out-File` emits BOM with utf8 encoding which breaks `npm ci` so I removed it, `pg-native` isn't installed by default so that was a noop and skipping `sqlite3` doesn't bring much of a benefit either.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,11 +36,11 @@ Here comes a little surprise: You need [Node.JS](http://nodejs.org).
 
 ### 2. Install the dependencies
 
-Just "cd" into sequelize directory and run `npm install`, see an example below:
+Just "cd" into sequelize directory and run `npm ci`, see an example below:
 
 ```sh
 $ cd path/to/sequelize
-$ npm install
+$ npm ci
 ```
 
 ### 3. Database
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ branches:
 cache: npm
 
 install:
-  - npm install
+  - npm ci
   - |-
     if [ "$DIALECT" = "postgres-native" ]; then npm install pg-native; fi
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,12 +14,7 @@ environment:
 
 install:
   - ps: Install-Product node $env:NODE_VERSION x64
-  - ps: |
-      $pkg = ConvertFrom-Json (Get-Content -Raw package.json)
-      $pkg.devDependencies.PSObject.Properties.Remove('sqlite3')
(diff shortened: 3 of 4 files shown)


sushantdhiman:
Need lock file update

CONTRIBUTING.md
```diff
@@ -36,11 +36,11 @@ Here comes a little surprise: You need [Node.JS](http://nodejs.org).
 
 ### 2. Install the dependencies
 
-Just "cd" into sequelize directory and run `npm install`, see an example below:
+Just "cd" into sequelize directory and run `npm ci`, see an example below:
```
sushantdhiman, line 39: `npm ci` is meant to be used in CI environments, right? I think we should not ask contributors to use this on their machine. Remove changes from this file

SimonSchick: No, https://docs.npmjs.com/cli/ci.html
`ci` ensures it installs exactly what is in the package-lock.json file, `install` will only look at the package.json

SimonSchick: That being said, while not *required* for developers it's always a good idea to ensure the dep tree is the same as in the CI to avoid confusion.

sushantdhiman: Not sure if I still prefer this but no problem with this change

sushantdhiman:
@SimonSchick You are using latest `npm`, right?

SimonSchick:
Currently 6.13.7
If you are worried, v10 latest ships with 6.14
https://nodejs.org/en/download/releases/

SimonSchick:
Build failure unrelated.. again.

sushantdhiman:
Update to latest npm and do install to update lock file

SimonSchick:
Is there an issue with the lockfile I'm not aware of?

sushantdhiman:
Yes, it contains some changes which should not be there with latest npm

SimonSchick:
I can do that but could you be more specific? I looked at the npm-cli changelog and I don't see anything that indicates issue that were resolved between my version and latest 🤔

sushantdhiman:
There are no issues with either version, just lock file changes will be very minimal with latest npm. Old npm version shift some package version in lockfile etc, resulting unnecessary changes to lockfile

SimonSchick:
As mentioned in my PR, I ran `npm dedupe` which optimized the minimal dependency tree.

SimonSchick:
@sushantdhiman can we merge this or do you want me to revert the changes to the lockfile?
I really want to get to the bottom of this query leak 😭

PR #12156, opened 2020-04-23
refactor(transaction): asyncify methods

jedwards1211:

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -64,7 +64,7 @@ class Transaction {
     } finally {
       this.finished = 'commit';
       if (!this.parent) {
-        await this.cleanup();
+        this.cleanup();
       }
       for (const hook of this._afterCommitHooks) {
         await hook.apply(this, [this]);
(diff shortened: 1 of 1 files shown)


jedwards1211:
@sushantdhiman when I asyncify the tests, are you going to want me to do those in small PRs or is it okay if I just do one big PR for them?

sushantdhiman:
@jedwards1211 Tests can be converted in one big PR

lib/transaction.js
```diff
@@ -117,47 +117,49 @@ class Transaction {
...
+    }
+
+    return result;
   }
 
-  setDeferrable() {
+  async setDeferrable() {
     if (this.options.deferrable) {
-      return this
+      return await this
```
SimonSchick, line 145: Nit: await seems redundant here.

jedwards1211: Oh weird, that was definitely not intentional, I had intended for my codemod to never await this 😅

jedwards1211: Oh I see there's stuff after it.  Yeah I went back and forth on whether to `return await` promises or just return them.  In the end I decided to `return await` in most cases to make it clear the method I was calling is async, and in case someone makes changes where they remove the `return` so they can add more statements after it, but forget to add `await`.

However, my codemod doesn't try to determine whether a method call is sync or async so there may be a few unnecessary `await`s, which are harmless but not ideal.  (I tried to look through the changes and get rid of `await` on any known sync calls).

In any case, removing `await` from return statements (caution: as long as they aren't in a try block) would be easy to do with a codemod, do you want me to make that change everywhere?

SimonSchick: I think it's best to keep `return await` the exception for try blocks.
I'd defer fixing existing code until the async/await conversion is done :)

PR #11727, opened 2019-12-03
Fixes #11726 - correct the overloaded method order

dancrumb:

PR #10880, opened 2019-05-01
refactor(hooks): remove method aliases for hooks

SimonSchick:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [ ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?

### Description of change

Another suggested change to streamline our API, this should reduce confusion for new users and drastically reduce churn for us to maintain the hook typings

PR #12160, opened 2020-04-23
refactor(hooks): refactor hooks to use composition (#10880)

SimonSchick:
#11727 # Pull Request check-list

### Description of change

Salvaged from previously suggested changes:

BREAKING CHANGE:

In order to streamline API:

- All method style add hook functions have been removed in favor of a composition based approach.
- Hook names have been removed, you can add and remove them by function reference instead which was supported before (this in line with browser and nodejs EventEmitter APIs)
- `this` inside of hooks no longer refers to the the the hook subject, it should not be used.

This affects `Model`, `Sequelize` and `Transaction`.

Before: `MyModel.beforeCreate(...)`
After: `MyModel.hooks.add('beforeCreate', ...)`

Before: `MyModel.addHook('beforeCreate', ...)`
After: `MyModel.hooks.add('beforeCreate', ...)`

Before: `MyModel.removeHook('beforeCreate', ...)`
After: `MyModel.hooks.remove('beforeCreate', ...)`

Before: `transaction.afterCommit(...)`
After: `transaction.hooks.add('afterCommit', ...)`

Before:

```js
MyModel.addHook('beforeCreate', 'named', fn);
MyModel.removeHook('beforeCreate', 'named');
```

After:

```js
MyModel.hooks.add('beforeCreate', fn);
MyModel.hooks.remove('beforeCreate', fn);
```

Before: `MyModel.addHook('beforeCreate', function() { this.someMethod(); });`
After: `MyModel.hooks.add('beforeCreate', () => { MyModel.someMethod(); });`


types/test/models/User.ts
```diff
@@ -71,22 +71,26 @@ User.init(
...
-User.afterFind((users, options) => {
+User.hooks.add('afterFind', (users: User[], options: FindOptions) => {
   console.log('found');
 });
 
+User.hooks.add('afterBulkCreate', (users: User[]) => {
+
+})
+
 // TODO: VSCode shows the typing being correctly narrowed but doesn't do it correctly
```
eseliger, line 92: is this still valid?

lib/hooks.js
```diff
@@ -66,39 +67,40 @@ const getProxiedHooks = hookType =>
...
+   * @param {Hooks} parent
    * @private
-   * @memberof Sequelize
-   * @memberof Sequelize.Model
    */
-  _setupHooks(hooks) {
-    this.options.hooks = {};
-    _.map(hooks || {}, (hooksArray, hookName) => {
+  constructor(hooks, parent) {
+    this.hooks = {};
```
eseliger, line 82: This seems like would be a good use-case for a `Map` 😁 But it's not really the scope of this PR so it's not blocking

types/test/models/User.ts
SimonSchick: Nah, but I will keep em as type assertions.

lib/hooks.js
```diff
@@ -181,416 +164,27 @@ const Hooks = {
...
   }
 }
-exports.applyTo = applyTo;
-
-/**
- * A hook that is run before validation
- *
- * @param {string}   name
- * @param {Function} fn   A callback function that is called with instance, options
- * @name beforeValidate
```
sushantdhiman, line 234: We should preserve / update this documentation. It can be placed alongside respective `hooksType` entry

docs/manual/other-topics/upgrade-to-v6.md
```diff
@@ -46,6 +46,51 @@ This method now tests for equality with [`_.isEqual`](https://lodash.com/docs/4.
   await instance.save(); // will save
 ```
 
+### Refactored hooks
+
+In order to streamline API:
+
+- All method style add hook functions have been removed in favor of a composition based approach.
+- Hook names have been removed, you can add and remove them by function reference instead which was supported before.
+- Another notable change that `this` inside of hooks no longer refers to the the the hook subject, it should not be used.
```
sushantdhiman, line 55: This is a pretty big breaking change imo, this is very useful writing common hook functions.

lib/hooks.js
SimonSchick: Can do, while I'm there, do you mind if we remove the `params: <NUMBER>` bit?
I just noticed this data is not used at all.

docs/manual/other-topics/upgrade-to-v6.md
SimonSchick: The primary motivation is to avoid having `this` change in context, you know where that usually leads to.
I wonder if it would be better to explicitly pass the model (class) as the first parameter for every hook.

sushantdhiman: > The primary motivation is to avoid having this change in context, you know where that usually leads to.

`this` is not bad, it is useful depends on how it is used, just like every other feature. I would like to keep it.

>  I wonder if it would be better to explicitly pass the model (class) as the first parameter for every hook.

That introduces another breaking change, I am ok with cleaning API by moving hooks to `something.hooks` but other things must be kept same

lib/hooks.js
SimonSchick: Ideally maps would be used *everywhere* instead of objects, prototype pollution is becoming more and more of an issue.

eseliger: Absolutely. Also, it has a performance benefit to it.

SimonSchick: Maps are faster than objects?
I'll need some numbers 😛

eseliger: https://jsperf.com/map-vs-object-sequelize/1 :)

lib/hooks.js
sushantdhiman: `param` indicate how many arguments are passed to hook, so you can deal with those based on that data.

sushantdhiman: We can remove `param` if documentation is kept with hook definition, thus clearly stating what arguments are being passed

SimonSchick: Will do, however expressing that 'clearly' is rather difficult with the limitation of jsdoc.
The typings will provide the most accurate API documentation here.

sushantdhiman: Just state what kind of object is being passed, jsdoc can handle that very well. Old docs from this file can be a starting point. Not everyone is using Typescript so we can't treat them as documentation

lib/hooks.js
SergeyCherman: @eseliger I was interested and gave it a test. Maps have a faster delete, but if just doing a set, then objects are faster.
https://jsperf.com/sequelize-map-vs-obj-2/1

eseliger: Interesting :)
I also added reading the values in again, seems to still be faster (for those who are interested)
https://jsperf.com/sequelize-map-vs-obj-2/3

SimonSchick: Interesting, in my yesterday benchmarks Maps were universally faster.

eseliger: What operation were you looking at?

lib/hooks.js
SimonSchick: @sushantdhiman I did recover the documentation, probably not like you envisioned it though.

Feel free to give me an example because I really don't use jsdoc and have no idea how to model this accordingly.

docs/manual/other-topics/upgrade-to-v6.md
```diff
@@ -52,6 +52,50 @@ This method now tests for equality with [`_.isEqual`](https://lodash.com/docs/4.
...
+### Refactored hooks
+
+In order to streamline API:
+
+- All method style add hook functions have been removed in favor of a composition based approach.
+- Hook names have been removed, you can add and remove them by function reference instead which was supported before.
+
+This affects `Model`, `Sequelize` and `Transaction`.
+
+#### Composition
```
sushantdhiman, line 64: Why are you calling these composition? :)

lib/hooks.js
```diff
@@ -5,50 +5,96 @@ const { logger } = require('./utils/logger');
 const debug = logger.debugContext('hooks');
 
 const hookTypes = {
-  beforeValidate: { params: 2 },
```
sushantdhiman, line 8: We can also consider `Model.hooks.beforeValidate(() => {])` in addition to `Model.hooks.add('beforeValidate', () => {})`, thus adding these methods directly in `Hooks` class. Syntax is bit shorter

Then we can document each method properly.

lib/hooks.js
sushantdhiman:
```js
/**
 * A hook that is run after validation
 *
 * @param {string}   name
 * @param {Function} fn   A callback function that is called with instance, options
 * @name afterValidate
 * @memberof Sequelize.Model
 */
```

This is want I was thinking

lib/transaction.js
```diff
@@ -19,9 +21,9 @@ class Transaction {
    * @param {string} [options.deferrable] Sets the constraints to be deferred or immediately checked. PostgreSQL only
    */
   constructor(sequelize, options) {
+    this.hooks = new Hooks();
```
sushantdhiman, line 24: Wont this means, now you can do `transaction.hook.add('beforeCreate')` only to find it won't work, as only `afterCommit` hooks is invoked in this module.

We should find a way to throw with `.add` method.

As I mentioned in another comment, if we go with `<object>.hooks.<hookName>` approach this can be avoided

docs/manual/other-topics/upgrade-to-v6.md
SimonSchick: Lack of a better word since the hooks the hooks are now a composition rather than a mix in, probably should call this `Examples` instead.

lib/hooks.js
SimonSchick: That'd kinda defeat the point of this tho, hooks don't know what they merely act as a promise event emitter.

Hooks that might be present in one context might not be in another.

lib/transaction.js
SimonSchick: That is a good point, I tried to avoid this scenario with strict typings but not everyone uses that 😞 , however my comment https://github.com/sequelize/sequelize/pull/12160#discussion_r415230504 still applies, adding methods to hooks for all possible hooks would be misleading and using inheritance would lead to a lot of bloaty code.

Example:

```
class ModelHooks extends Hooks {
  beforeCreate(a, b, c): Promise<void> {
    return this.run(a, b, c);
  // A lot of code
}

class SequelizeHooks extends ModelHooks {
  // ...
}
```

I wanted to avoid making hooks` specific to their consumers but I guess this would work, I'd make the `run` method private tho to avoid too much API surface and potential misuse.

SimonSchick:
@eseliger any chance you'd be interested in picking this up, kinda bust atm.

eseliger:
I thought it's been merged way ago haha. I can see if I find some time, but no promise

PR #12163, opened 2020-04-24
fix(model): fix unchained promise in association logic in bulkCreate

SimonSchick:
### Pull Request check-list

### Description of change

Found this while messing around in code base.
Wrote no test due to high complexity of function, plus nothing seems to be failing as a result...

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2642,7 +2642,7 @@ class Model {
               const associationInstance = createdAssociationInstances[idx];
               const instance = associationInstanceIndexToInstanceMap[idx];
 
-              instance[include.association.accessors.set](associationInstance, { save: false, logging: options.logging });
+              await include.association.set(instance, associationInstance, { save: false, logging: options.logging });
             }
           }));
         }


lib/model.js
```diff
@@ -2642,7 +2642,7 @@ class Model {
               const associationInstance = createdAssociationInstances[idx];
               const instance = associationInstanceIndexToInstanceMap[idx];
 
-              instance[include.association.accessors.set](associationInstance, { save: false, logging: options.logging });
+              await include.association.set(instance, associationInstance, { save: false, logging: options.logging });
```
sushantdhiman, line 2645: `save: false`, means this will not return promise. So await is not useful here. Why `accessors` is being removed?

SimonSchick: All of the `set` function were converted to async so they *always* return a promise now.
Using `accessors.set` was only used to look up the instance `set` method which is effectively just a proxy (see helpers.js:64) seemed redundant to me 🤷‍♂️

sushantdhiman: Ah yes, it is an async method now. That make sense

Issue #11739, opened 2019-12-12
change: remove import helper


sushantdhiman:
Tracking https://github.com/sequelize/sequelize/pull/10892

PR #12175, opened 2020-04-26
refactor: remove sequelize.import helper

sushantdhiman:
### Pull Request check-list

### Description of change
Closes https://github.com/sequelize/sequelize/issues/11739

### DIFF
--- a/docs/manual/other-topics/upgrade-to-v6.md
+++ b/docs/manual/other-topics/upgrade-to-v6.md
@@ -20,9 +20,11 @@ You should now use [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked) packag
   Sequelize.useCLS(namespace);
 ```
 
-### Promise
+### Sequelize
 
-Bluebird has been removed. Public API now returns native promises. `Sequelize.Promise` is no longer available.
+- Bluebird has been removed. Public API now returns native promises.
+- `Sequelize.Promise` is no longer available.
+- `sequelize.import` method has been removed.
 
 ### Model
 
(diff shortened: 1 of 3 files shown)


MichielDeMey:
I noticed that `.import` has been deprecated in `beta6`, is there an upgrade path or best practices on how to migrate existing code?

The upgrade guide only mentions that the function has been deprecated, but offers no other directions. 😅

papb:
Hi @MichielDeMey, thanks for the comment, we should improve that part of the docs. Can you tell me more about your situation? Can you give an example of code snippet that you have using `import`?

Does your case stem from [this issue in the CLI](https://github.com/sequelize/cli/issues/895) or is it something else?

MichielDeMey:
Yep, looks like the code was originally generated by the CLI.
We can follow up on that issue regarding the deprecated `import` method. 👍

roychri:
For those coming here and tryign that old issue with CLI which is now gone, here's how you can migrate.

From:

```
sequelize.import(path.join(__dirname, '', filename));
```

To:

```
require(path.join(__dirname, file))(sequelize, Sequelize.DataTypes);
```

mdodge-ecgrow:
I switched over to using the `require` line and am now getting this error (using Node 14 and Sequelize 6):
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: C:\.....\models\Permissions.js
require() of ES modules is not supported.
require() of C:\.....\models\Permissions.js from C:\.....\models\index.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename Permissions.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from C:\.....\package.json.
```

I don't want to rename all my models to .cjs and I don't want to remove `"type":"module"`, so I need to somehow figure out that third option.

machine-gurning:
@mdodge-ecgrow did you find a solution? I have the same problem

ephys:
you need to use `import` instead of `require`

PR #12182, opened 2020-04-27
Fix tedious connect deprecation

### LINKED ISSUE #12181: tedious 8.3.0 deprecation
## Issue Description

`tedious` will require calling `.connect()` on the connection in v9 it currently reports a deprecation in v8.3.0

### What are you doing?

using `tedious` v8.3.0 to connect to MS SQL Server

No deprecation notice

tedious is showing deprecation notice:

```
In the next major version of `tedious`, creating a new `Connection`
instance will no longer establish a connection to the server automatically. Please use
the new `connect` helper function or call the `.connect` method on the newly created
`Connection` object to silence this message.
```

### Additional context

The `tedious` PR with this deprecation: https://github.com/tediousjs/tedious/pull/1069

### Environment

- Sequelize version: 5.21.7
- Node.js version: 12.6.1

- [x] I think this problem happens only for the following dialect(s): MSSQL

UziTech:
### Pull Request check-list

### Description of change

Call `connect` after creating `tedious` connection if required.

Fixes #12181

### DIFF
--- a/lib/dialects/mssql/connection-manager.js
+++ b/lib/dialects/mssql/connection-manager.js
@@ -61,6 +61,9 @@ class ConnectionManager extends AbstractConnectionManager {
     try {
       return await new Promise((resolve, reject) => {
         const connection = new this.lib.Connection(connectionConfig);
+        if (connection.state === connection.STATE.INITIALIZED) {
+          connection.connect();
+        }
         connection.queue = new AsyncQueue();
         connection.lib = this.lib;
 
--- a/test/unit/dialects/mssql/connection-manager.test.js
+++ b/test/unit/dialects/mssql/connection-manager.test.js
@@ -79,5 +79,29 @@ if (dialect === 'mssql') {
           expect(err.parent.message).to.equal('Connection was closed by remote server');
         });
     });
+
+    it('connectionManager._connect() should call connect if state is initialized', function() {
+      const connectStub = sinon.stub();
+      const INITIALIZED = { name: 'INITIALIZED' };
+      this.connectionStub.returns({
+        STATE: { INITIALIZED },
+        state: INITIALIZED,
+        connect: connectStub,
+        once(event, cb) {
+          if (event === 'connect') {
+            setTimeout(() => {
+              cb();
+            }, 500);
+          }
+        },
+        removeListener: () => {},
+        on: () => {}
+      });
+
+      return this.instance.dialect.connectionManager._connect(this.config)
+        .then(() => {
+          expect(connectStub.called).to.equal(true);
+        });
+    });
   });
 }
--- a/test/unit/dialects/mssql/connection-manager.test.js
+++ b/test/unit/dialects/mssql/connection-manager.test.js
@@ -39,6 +39,8 @@ if (dialect === 'mssql') {
 
     it('connectionManager._connect() does not delete `domain` from config.dialectOptions', function() {
       this.connectionStub.returns({
+        STATE: {},
+        state: '',
         once(event, cb) {
           if (event === 'connect') {
             setTimeout(() => {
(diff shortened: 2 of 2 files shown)


UziTech:
I'm not sure how to test this since it would require installing different versions of tedious.

RodrigoHamuy:
Thanks for this fix. When can we expect it to be on v5? Thanks.

sushantdhiman:
If someone opens a PR for v5, it will get auto released on merge

UziTech:
this fix is available in v5.21.10

RodrigoHamuy:
Thanks!

lib/dialects/mssql/connection-manager.js
```diff
@@ -61,6 +61,9 @@ class ConnectionManager extends AbstractConnectionManager {
     try {
       return await new Promise((resolve, reject) => {
         const connection = new this.lib.Connection(connectionConfig);
+        if (connection.state === connection.STATE.INITIALIZED) {
+          connection.connect();
+        }
```
arthurschreiber, line 66: @UziTech I don't think you should be checking against `state` here. Just calling `.connect` should be just fine. 🤔

UziTech: Unless the user is using an old version of tedious where calling connect would throw an error.

killerfurbel: But this Code leads to an unhandled exception if used with msnodesqlv8 instead of tedious.

connection.state is undefined as well as connection.STATE, so it will throw a `cannot read property INITIALIZED of undefined` Error.

Could you please check if connection.STATE is defined before accessing INITIALIZED? Checking if `.connect` exists would not be suitable because msnodesqlv8 does not require the `.connect()` call at all, this is done in the Constructor already.

arthurschreiber: @killerfurbel This code is not supposed to work with `msnodesqlv8` in the first place. It has a completely different API than `tedious`.

killerfurbel: Really? See https://github.com/TimelordUK/node-sqlserver-v8#use-with-sequelize ... and why does sequelize support the `dialectModule` property in the constructor, if tedious is the only choice?
I agree, tedious is built in, therefore it is possibly the first choice for most users. But it is possible to use others and this code breaks it... I don't think it has a completely different API - it has a specific sequelize connector (https://github.com/TimelordUK/node-sqlserver-v8/blob/master/lib/sequelize/connection.js).

arthurschreiber: So, the `msnodesqlv8` integration for `sequelize` works by re-implementing the `tedious` API, which to me as the `tedious` maintainer, seems extremely questionable.

arthurschreiber: I think the main idea behind `dialectModule` is so you can e.g. use e.g. a forked version of `tedious` or any other connector used by `sequelize`.

The way that `msnodesqlv8` build compatibility for `sequelize` "works", but I'm not sure that is a robust approach. Shouldn't there be a separate dialect for `msnodesqlv8` if that is supposed to be a fully supported option?

sushantdhiman: Sequelize is not designed to work with `msnodesqlv8`

UziTech: > and this code breaks it... I don't think it has a completely different API

If it wants to keep working with sequelize it could add a `STATE` property.

killerfurbel: I created an issue in the msnodesqlv8 repo: https://github.com/TimelordUK/node-sqlserver-v8/issues/159

However, since you also backported this behavior also to v5, I would strongly vote for a check within sequelize or revert the commit in the v5 branch. I think "removing" a depreciation warning by implementing a breaking change is not reasonable and should only happen in a major release (--> 6.x).

UziTech: You could create a PR to check for `STATE`. That would probably be the best chance of getting it changed.

PR #12213, opened 2020-05-02
refactor: use object spread instead of Object.assign for new objects

SimonSchick:
### Pull Request check-list

### Description of change

This replaces most `Object.assign` usage with the more performant (about 40%) (**1**) object spread.

- All cases of `Object.assign({}, ...)` have been auto refactored by eslint.
- The remaining ones where objects where composed over multiple lines were manually adjusted.
- The `|| {}` pattern is not needed as `{ ...undefined }` yields `{}`.
- `val = Object.assign(val, ...)` has been replaced with `Object.assign(val, ...)` to avoid confusion and redundant assigns.

Notes:
- (**1**) Performance may vary by node version, in v10+ spread is generally faster.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -94,7 +94,8 @@
     "no-unused-expressions": "error",
     "no-sequences": "error",
     "no-self-compare": "error",
-    "no-case-declarations": "off"
+    "no-case-declarations": "off",
+    "prefer-object-spread": "error"
   },
   "settings": {
     "jsdoc": {
--- a/lib/associations/has-one.js
+++ b/lib/associations/has-one.js
@@ -190,9 +190,7 @@ class HasOne extends Association {
    * @returns {Promise}
    */
   async set(sourceInstance, associatedInstance, options) {
-    options = Object.assign({}, options, {
-      scope: false
-    });
+    options = { ...options, scope: false };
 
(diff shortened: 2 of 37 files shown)


.eslintrc.json
```diff
@@ -94,7 +94,8 @@
     "no-unused-expressions": "error",
     "no-sequences": "error",
     "no-self-compare": "error",
-    "no-case-declarations": "off"
+    "no-case-declarations": "off",
+    "prefer-object-spread": "error"
```
eseliger, line 98: +1 for thinking of this

lib/associations/has-one.js
```diff
@@ -190,9 +190,7 @@ class HasOne extends Association {
    * @returns {Promise}
    */
   async set(sourceInstance, associatedInstance, options) {
-    options = Object.assign({}, options, {
-      scope: false
-    });
+    options = { ...options, scope: false };
```
eseliger, line 193: ideally parameters should be read-only, but that's another problem

SimonSchick: Correct, I've seen numerous fixes creeping in over time that end up cloning/deep cloning inputs, but this almost impossible to fix on a greater scale without static analysis.

eseliger: Totally agree

PR #12214, opened 2020-05-02
refactor: replace lodash clone with object spread

SimonSchick:
### Pull Request check-list

### Description of change

This replaces (almost) all `_.clone` calls with object spreading (there was 1 array clone).

Turns out `_.clone` is around least an order of magnitude slower compared to native object spreading!

Consider this (rather naive) performance test:

```js
'use strict';

const { clone } = require('lodash');

const data = {
  a: 1,
  b: 2,
  c: 3,
  d: 4,
  e: 5
};


let out;
console.time('clone');
for (let i = 0; i < 0xffffff; i++) {
  data[Math.floor(Math.random() * 10)] = Math.random(); //avoid opt
  out = clone(data);
}
console.timeEnd('clone');

console.time('spread');
for (let i = 0; i < 0xffffff; i++) {
  data[Math.floor(Math.random() * 10)] = Math.random(); //avoid opt
  out = { ...data };
}
console.timeEnd('spread');
console.log(out); // avoid opt
```

The results (on my machine):
```
$ node speed.js
clone: 23.944s
spread: 840.681ms
```

See also https://www.measurethat.net/Benchmarks/Show/4542/0/object-cloning-with-lodash-clone-vs-clonedeep-vs-merge

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -122,7 +122,7 @@ class Model {
     let defaults;
     let key;
 
-    values = values && _.clone(values) || {};
+    values = { ...values };
 
     if (options.isNewRecord) {
       defaults = {};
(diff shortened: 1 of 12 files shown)


lib/model.js
```diff
@@ -2598,12 +2593,14 @@ class Model {
       }
       if (options.individualHooks) {
         await Promise.all(instances.map(async instance => {
-          const individualOptions = _.clone(options);
+          const individualOptions = {
+            ...options,
+            validate: false,
+            hooks: true
+          };
           delete individualOptions.fields;
```
eseliger, line 2601: maybe this can leverage a `omit` on the options, bc delete is slow

lib/model.js
```diff
@@ -3140,10 +3137,12 @@ class Model {
           }
         } else {
           instances = await Promise.all(instances.map(async instance => {
-            const individualOptions = _.clone(options);
+            const individualOptions = {
+              ...options,
+              hooks: false,
+              validate: false
+            };
             delete individualOptions.individualHooks;
```
eseliger, line 3145: same here

lib/model.js
SimonSchick: `omit` creates a new object entirely which is MUCH slower.

SimonSchick: I'd agree if this was more than 2-3 properties but for a single one cloning the entire object just isn't worth it.

SimonSchick:
More deadlocking 🎉

PR #12217, opened 2020-05-03
refactor: move all dialect conditional logic into subclass

SimonSchick:
### Pull Request check-list

#### Note:

The idea of this PR is to de-couple the dialects from the core query interface.
This reduces overhead as not all dialects required certain pre-computing that was performed in some places.

Some of these `breaking changes` are not really breaking due to not being in the official API but better safe than sorry.

Many of the postgres specific query interface APIs were never documented, should they?

Some of the changes (such as the rename of `QueryInterface -> queryInterface` and the removal of `stack` aren't strictly needed but do streamline the API.

BREAKING CHANGE:

All instances of `QueryInterface` and `QueryGenerator` have been renamed to their lowerCamelCase variants eg. `queryInterface` and `queryGenerator` when used as property names on Model and Dialect, the class names remain the same.

`Utils.stack` has been removed.

`QueryInterface` is now an abstract class.

`query-interface.js` was moved to `lib/dialects/abstract/query-interface` and the exports have been reduced to `QueryInterface`.
`default` and `module.exports` no longer refer to this class.

`QueryInterface.addConstraint`s now only takes 2 parameters, `tableName` and `options`, previously the second parameter _could_ be a list of column names to apply the constraint to, this list must now be passed in `options` as the `fields` property.
The fourth `rawTableName` parameter was entirely removed as it was effectively unused.


SimonSchick:
Mostly looking for feedback on the 'breaking' changes here, the build should be fixed by tomorrow.

sushantdhiman:
Breaking changes you have listed are fine for me

SimonSchick:
That was.. easier to fix than I thought.

SimonSchick:
OH, there are some unrelated changes in `support.js` which made sqlite integration tests MOSTLY pass on windows, file locking was super fucky.

lib/dialects/abstract/query-interface.js
```diff
@@ -2,25 +2,24 @@
...
- * @class QueryInterface
+ * @abstract
  */
 class QueryInterface {
-  constructor(sequelize) {
+  constructor(sequelize, queryGenerator) {
     this.sequelize = sequelize;
-    this.QueryGenerator = this.sequelize.dialect.QueryGenerator;
+    if (!queryGenerator) {
+      throw new Error('wut?');
```
sushantdhiman, line 20: ??? Remove

lib/dialects/mariadb/index.js
```diff
@@ -52,7 +54,7 @@ MariadbDialect.prototype.supports = _.merge(
 
 ConnectionManager.prototype.defaultVersion = '5.5.3';
 MariadbDialect.prototype.Query = Query;
-MariadbDialect.prototype.QueryGenerator = QueryGenerator;
```
sushantdhiman, line 55: This should be kept as `QueryGenerator`, to indicate a class is being exported

lib/dialects/abstract/query-interface.js
```diff
@@ -2,25 +2,24 @@
...
+const DataTypes = require('../../data-types');
+const Transaction = require('../../transaction');
+const QueryTypes = require('../../query-types');
+const Op = require('../../operators');
 
 /**
  * The interface that Sequelize uses to talk to all databases
  *
- * @class QueryInterface
+ * @abstract
```
sushantdhiman, line 14: This should be marked as a private class, all inherited query interface classes from dialect should be marked as 'public' by removing `@private` directive if present in their definition

lib/dialects/abstract/query-interface.js
SimonSchick: Yup, debug code.

lib/dialects/abstract/query-interface.js
SimonSchick: The base class isn't really private though, it's currently exposed in the docs.
If anything it should be treated like an interface but idk how to express that in jsdoc.

lib/dialects/mssql/query-interface.js
```diff
@@ -1,53 +1,55 @@
...
-  @param  {QueryInterface} qi
-  @param  {string} tableName     The name of the table.
-  @param  {string} attributeName The name of the attribute that we want to remove.
-  @param  {object} options
-  @param  {boolean|Function} [options.logging] A function that logs the sql queries, or false for explicitly not logging these queries
-
-  @private
+ * Returns an object that treats MSSQL's inabilities to do certain queries.
+ *
+ * @private
```
sushantdhiman, line 8: As previously noted in other comment, remove `@private` directive. Comment can be improved to indicate this is a MSSQL Query Interface

lib/dialects/mysql/index.js
```diff
@@ -50,7 +52,7 @@ MysqlDialect.prototype.supports = _.merge(_.cloneDeep(AbstractDialect.prototype.
 
 ConnectionManager.prototype.defaultVersion = '5.6.0';
 MysqlDialect.prototype.Query = Query;
-MysqlDialect.prototype.QueryGenerator = QueryGenerator;
```
sushantdhiman, line 53: Remove

lib/dialects/mysql/query-interface.js
```diff
@@ -1,89 +1,81 @@
...
 /**
- A wrapper that fixes MySQL's inability to cleanly remove columns from existing tables if they have a foreign key constraint.
-
- @param  {QueryInterface} qi
- @param  {string} tableName     The name of the table.
- @param  {string} columnName    The name of the attribute that we want to remove.
- @param  {object} options
-
- @private
+ * Returns an object that treats MySQL's inabilities to do certain queries.
```
sushantdhiman, line 7: Update documentation here, say this is the MySQL query interface. This comment will be exposed in documentation

SimonSchick:
@sushantdhiman I'm off for tonight, feel free to make any amends to this PR as you see fit directly rather than through comment, I will look at any changes tomorrow.

lib/dialects/postgres/query-interface.js
```diff
@@ -2,154 +2,252 @@
...
-  return result;
-}
 
+  /**
+   * Since postgres has a special case for enums, we should drop the related
+   * enum type within the table and attribute
+   *
+   * @override
+   */
+  async dropTable(tableName, options) {
```
sushantdhiman, line 219: Nice refactor

lib/dialects/sqlite/query-interface.js
```diff
@@ -2,192 +2,169 @@
 
 const sequelizeErrors = require('../../errors');
 const QueryTypes = require('../../query-types');
+const { QueryInterface } = require('../abstract/query-interface');
+const { cloneDeep } = require('../../utils');
+
+class SQLiteQueryInterface extends QueryInterface {
+  /**
+   * A wrapper that fixes SQLite's inability to remove columns from existing tables.
```
sushantdhiman, line 10: This comment needs improvement as well

lib/dialects/postgres/query-interface.js
```diff
@@ -2,154 +2,252 @@
...
+    await super.dropTable(tableName, options);
+    const promises = [];
+    const instanceTable = this.sequelize.modelManager.getModel(tableName, { attribute: 'tableName' });
+
+    if (!instanceTable) {
+      // throw error??
```
SimonSchick, line 225: Oh yea, a comment here would be appreciated.

sushantdhiman: Not sure, this case is for Migrations where models may not be available. So can't throw here. Ideally low level API's should not be aware of models.

papb:
Fantastic work, @SimonSchick !!

lib/dialects/postgres/query-interface.js
```diff
@@ -2,154 +2,252 @@
...
+
+    const getTableName = (!options || !options.schema || options.schema === 'public' ? '' : `${options.schema}_`) + tableName;
+
+    const keys = Object.keys(instanceTable.rawAttributes);
+    const keyLen = keys.length;
+
+    for (let i = 0; i < keyLen; i++) {
+      if (instanceTable.rawAttributes[keys[i]].type instanceof DataTypes.ENUM) {
+        const sql = this.queryGenerator.pgEnumDrop(getTableName, keys[i]);
+        options.supportsSearchPath = false;
```
jifeon, line 237: @SimonSchick @sushantdhiman 

Looks like `options` are optional here, so if you invoke Model.drop without options, there will be an error here.

```
TypeError: Cannot set property 'supportsSearchPath' of undefined
    at PostgresQueryInterface.dropTable (node_modules/sequelize/lib/dialects/postgres/query-interface.js:237:36)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

PR #12220, opened 2020-05-04
refactor(abstract-query): improve _groupJoinData perf and readability

SimonSchick:
### Pull Request check-list

### Description of change

This cleans up the probably worst function in the entire code base and manages to also make it slightly faster in the process.

Worth mentioning:
 - All code related to `getUniqueKeyAttributes` was removed as it was effectively broken (the does something that makes absolutely no sense and won't ever have an effect), I want to fix this but I don't understand what it was suppose to a achieve at all.
 - Function is about 10% faster.
 -

### DIFF
--- a/src/dialects/abstract/query.js
+++ b/src/dialects/abstract/query.js
@@ -457,10 +457,6 @@ class AbstractQuery {
      * Previous set won't necessarily be parent set (one parent could have two children, one child would then be previous set for the other)
      */
 
-    /*
-     * Author (MH) comment: This code is an unreadable mess, but it's performant.
-     * groupJoinData is a performance critical function so we prioritize perf over readability.
-     */
     if (!rows.length) {
       return [];
     }
@@ -469,14 +465,32 @@ class AbstractQuery {
 
     const memoKeyPrefix = makeDoublePrefixMemoizer();
 
+    const uniqueKeyMap = {};
+
+    const getUniqueKeyAttributes = model => {
+      if (model.uniqueKeys.length === 0) {
+        return [];
+      }
+      if (Object.prototype.hasOwnProperty.call(uniqueKeyMap, model)) {
+        return uniqueKeyMap[model];
+      }
+      const uniqueKeys = [];
+      for (const { fields: uniqueFields } of Object.values(model.uniqueKeys)) {
+        for (const attribute of Object.values(model.rawAttributes)) {
+          if (uniqueFields.includes(attribute.field)) {
+            uniqueKeys.push(attribute.fieldName);
+          }
+        }
+      }
+      uniqueKeyMap[model] = uniqueKeys;
+
+      return uniqueKeys;
+    };
+
     /**
      * @type {string}
      */
     let prevKey;
-    /**
-     * @type {object}
-     */
-    let values;
     /**
      * @type {boolean}
      */
@@ -489,7 +503,7 @@ class AbstractQuery {
 
     const resultMap = {};
 
-    const computeHashKeyForInstance = (row, prevKeyPrefix, topHash) => {
+    const computeHashKeyForInstance = (row, prevKeyPrefix, topHash, values) => {
       const length = prevKeyPrefix.length;
       /**
        * @type {string}
@@ -504,6 +518,7 @@ class AbstractQuery {
        */
       let itemHash = topHash;
 
+      let isEmpty = false;
       for (let i = 0; i < length; i++) {
         /**
          * @type {string}
@@ -513,6 +528,15 @@ class AbstractQuery {
         for (const primKeyAttr of includeMap[prefix].model.primaryKeyAttributes) {
           itemHash += stringify(row[`${prefix}.${primKeyAttr}`]);
         }
+        if (includeMap[prefix].model.primaryKeyAttributes.length === 0) {
+          const uniqueKeyAttributes = getUniqueKeyAttributes(includeOptions.model);
+          for (const uniqAttr of uniqueKeyAttributes) {
+            if (row[`${prefix}.${uniqAttr}`] !== null) {
+              itemHash += stringify(row[`${prefix}.${uniqAttr}`]);
+            }
+          }
+        }
+        isEmpty = itemHash === prefix;
         if (!parentHash) {
           parentHash = topHash;
         }
@@ -549,13 +573,16 @@ class AbstractQuery {
       if (!parent[lastKeyPrefix]) {
         parent[lastKeyPrefix] = [];
       }
-      parent[lastKeyPrefix].push((resultMap[itemHash] = values));
+      if (!isEmpty) {
+        resultMap[itemHash] = values;
+        parent[lastKeyPrefix].push(resultMap[itemHash]);
+      }
     };
 
     const rowsLength = rows.length;
     const results = checkExisting ? [] : new Array(rowsLength);
-    // Keys are the same for all rows, so only need to compute them on the first row
 
+    // Keys are the same for all rows, so only need to compute them on the first row
     const keys = rowsLength > 0 ? Object.keys(rows[0]) : undefined;
     const keyPrefixes = keys ? new Array(keys.length) : undefined;
     // precompute prefixes and include maps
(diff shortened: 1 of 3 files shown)


lib/dialects/abstract/query.js
```diff
@@ -7,6 +7,46 @@ const Dot = require('dottie');
...
 const uuid = require('uuid/v4');
 
+const stringify = obj => obj instanceof Buffer ? obj.toString('hex') : obj;
+
+// Calculates the array prefix of a key (['User', 'Results'] for 'User.Results.id')
+const makeDoublePrefixMemoizer = () => {
+  const keyPrefixMemo = {};
+  return str => {
+    // We use a double memo and keyPrefixString so that different keys with the same prefix will receive the same array instead of differnet arrays with equal values
+    if (!Object.prototype.hasOwnProperty.call(keyPrefixMemo, str)) {
```
sushantdhiman, line 17: For better readability

```Reflect.apply(Object.hasOwnProperty, keyPrefixMemo, [str]```

**Comment by sushantdhiman** [comment]
This code segment does indeed need this refactor but it is performance critical.

> Function is about 10% faster.

Any benchmark script to support this?

File: lib/dialects/abstract/query.js [hunk 2]
```diff
@@ -7,6 +7,46 @@ const Dot = require('dottie');
...
+  };
+};
+
+/**
+ * Calculate the last item in the array prefix ('Results' for 'User.Results.id')
+ *
+ * @param {string} str
+ */
+const secondSuffix = str => {
+  const lastDelim = str.lastIndexOf('.');
```
sushantdhiman, line 34: Will this work?

```js
  if (!str.includes('.')) return '';

  const array = str.split('.');
  return array[array.length - 1];
```

SimonSchick: I tried this approach and it was significantly slower.

lib/dialects/abstract/query.js
SimonSchick: This feels rather subjective, if anything I'd like to generate all usage of `Object.prototype.hasOwn...` to a util called `Utils.hasOwnProperty(target, str)` but this is out of scope of this PR.
The temporary array is also likely to incur another performance hit.

I'll do some benchmarking around this, I never really worked with the `Reflect` library, it feels like a clone of `Object`.

SimonSchick:
@sushantdhiman I grabbed the the unit tests and threw them in a simple test suite `using console.time/timeEnd`, the drop was around 10% each time, if you want I can attach the benchmark here tomorrow.

sushantdhiman:
@SimonSchick Got it, I am more concerned about unique attributes thingi. I will try to test how it was used etc

SimonSchick:
Thanks, I was counting on you to look into that, me and @eseliger couldn't figure out what the original author was trying to achieve, the code just didn't make any sense what so ever.

lib/dialects/abstract/query.js
SimonSchick: Naive benchmark:
```js
'use strict';

const obj = {
};

function randIndex() {
  return Math.floor(Math.random() * 100);
}

for (let i = 0; i < 100;++i) {
  obj[randIndex()] = Math.random();
}

let out;
console.time('Reflect.apply');
for (let i = 0; i < 0xffffff; i++) {
  out = Reflect.apply(Object.hasOwnProperty, obj, [randIndex()]);
}
console.timeEnd('Reflect.apply');

console.time('Object.prototype.hasOwnProperty.call');
for (let i = 0; i < 0xffffff; i++) {
  out = Object.prototype.hasOwnProperty.call(obj, randIndex());
}
console.timeEnd('Object.prototype.hasOwnProperty.call');

const bound = Object.prototype.hasOwnProperty;
function hasOwnProp(obj, prop) {
  return bound.call(obj, prop);
}

console.time('hasOwnProp');
for (let i = 0; i < 0xffffff; i++) {
  out = hasOwnProp(obj, randIndex());
}
console.timeEnd('hasOwnProp');

console.log(out);
```

```
Reflect.apply: 547.062ms
Object.prototype.hasOwnProperty.call: 353.256ms
hasOwnProp: 343.112ms

Reflect.apply: 544.96ms
Object.prototype.hasOwnProperty.call: 353.499ms
hasOwnProp: 357.777ms

Reflect.apply: 547.306ms
Object.prototype.hasOwnProperty.call: 354.269ms
hasOwnProp: 350.131ms
```

Reflect is quite a bit slower than the Object variant, the wrapped (tidy) version is sometimes faster, sometimes slower but pretty much the same speed on average, I will likely file a PR to replace those later.

SimonSchick:
@sushantdhiman do we want to move forward with this, did you find anything related to the unique checks?

sushantdhiman:
@SimonSchick yes I want to merge this PR, will look over unique attribute case tomorrow (weekend)

SimonSchick:
Also speed:

```
Before:
Suite 1: 4.774s
Suite 2: 4.326s
Suite 3: 4.814s
Suite 4: 4.142s
Suite 5: 3.155s
Suite 6: 3.756s
Suite 7: 2.079s

After:
Suite 1: 4.468s
Suite 2: 4.076s
Suite 3: 4.370s
Suite 4: 3.943s
Suite 5: 2.921s
Suite 6: 3.377s
Suite 7: 1.640s
```

https://gist.github.com/SimonSchick/aea921ec33879833b19b6cb198d2f375
All the suites are based on the existing test cases

eseliger:
That's awesome!

sushantdhiman:
@SimonSchick This method is used to populate results when included model got no PK. But it is currently broken. It relies on `model.attributes` incorrectly, it should be using `model.rawAttributes`

```diff
const getUniqueKeyAttributes = model => {
  let uniqueKeyAttributes = _.chain(model.uniqueKeys);
  uniqueKeyAttributes = uniqueKeyAttributes
    .result(`${uniqueKeyAttributes.findKey()}.fields`)
-   .map(field => _.findKey(model.attributes, chr => chr.field === field))
+   .map(field => _.findKey(model.rawAttributes, chr => chr.field === field))
    .value();

  return uniqueKeyAttributes;
};
```

I am including a test case below with correct results. It is broken on

1) `master`, because `model.attributes` is used to find unique fields
2) `this branch`, because it completely removed handling of unique non-PK based models 

```js
(async () => {
  const Foo = sequelize.define("Foo", {
    name: {
      type: Sequelize.STRING,
      unique: true,
    },
  });

  // Foo.removeAttribute("id");

  const Bar = sequelize.define("Bar", {
    name: {
      type: Sequelize.STRING,
      unique: true,
    },
  });

  Bar.removeAttribute("id");

  Foo.hasMany(Bar, { foreignKey: "fooName", sourceKey: "name" });
  Bar.belongsTo(Foo, { foreignKey: "fooName", sourceKey: "name" });

  await Foo.bulkCreate([{ name: "foo-a" }, { name: "foo-b" }]);
  await Bar.bulkCreate([{ name: "bar-a", fooName: "foo-a" }]);
  await Bar.bulkCreate([{ name: "bar-b", fooName: "foo-a" }]);

  console.log(
    JSON.stringify(await Foo.findAll({ include: { all: true } }), null, 2)
  );
})();
```

```sql
[
  {
    "id": 1,
    "name": "foo-a",
    "createdAt": "2020-05-09T06:41:22.289Z",
    "updatedAt": "2020-05-09T06:41:22.289Z",
    "Bars": [
      {
        "name": "bar-a",
        "createdAt": "2020-05-09T06:41:22.294Z",
        "updatedAt": "2020-05-09T06:41:22.294Z",
        "fooName": "foo-a"
      },
      {
        "name": "bar-b",
        "createdAt": "2020-05-09T06:41:22.295Z",
        "updatedAt": "2020-05-09T06:41:22.295Z",
        "fooName": "foo-a"
      }
    ]
  },
  {
    "id": 2,
    "name": "foo-b",
    "createdAt": "2020-05-09T06:41:22.289Z",
    "updatedAt": "2020-05-09T06:41:22.289Z",
    "Bars": [
      {
        "name": null,
        "createdAt": null,
        "updatedAt": null,
        "fooName": null
      }
    ]
  }
]
```

When issue pointed in this comment is addressed, https://github.com/sequelize/sequelize/issues/5193 will be fixed as well :tada:

SimonSchick:
This still doesn't make sense with `.result(`${uniqueKeyAttributes.findKey()}.fields`}` as this only returns the first item of the `uniqueKeyAttributes`?

sushantdhiman:
> This still doesn't make sense with .result(${uniqueKeyAttributes.findKey()}.fields} as this only returns the first item of the uniqueKeyAttributes?

AFAIU, in case of multiple unique keys it work with the first available unique key. It make sense as records will be unique for any of `uniqueKeys` selection, we can use any of those keys to find unique record.

SimonSchick:
Alright, I will re-add this logic, thanks for looking into it.

SimonSchick:
Will this not fail for compound unique keys though? See
`_.findKey(model.attributes, chr => chr.field === field)` compares fields by reference which might be different arrays?

sushantdhiman:
> _.findKey(model.attributes, chr => chr.field === field)

I think this logic is trying to convert field names to their attribute name, `uniqueKey.fields` are usually column names, which may not match with their respective attribute name in model

eseliger:
@SimonSchick Addressed Sushant's feedback. So this closes #5193.

PR #12221, opened 2020-05-04
refactor(model-manager): simplify model manager

eseliger:
### Pull Request check-list

### Description of change

Refactored the model manager to not depend on the sequelize instance. Rather, sequelize exposes the model manager instances now

### DIFF
--- a/lib/model-manager.js
+++ b/lib/model-manager.js
@@ -4,26 +4,21 @@ const Toposort = require('toposort-class');
 const _ = require('lodash');
 
 class ModelManager {
-  constructor(sequelize) {
+  constructor() {
     this.models = [];
-    this.sequelize = sequelize;
   }
 
   addModel(model) {
     this.models.push(model);
-    this.sequelize.models[model.name] = model;
-
     return model;
   }
 
   removeModel(modelToRemove) {
     this.models = this.models.filter(model => model.name !== modelToRemove.name);
-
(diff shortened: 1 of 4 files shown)


lib/model-manager.js
```diff
@@ -4,26 +4,21 @@ const Toposort = require('toposort-class');
...
 
   removeModel(modelToRemove) {
     this.models = this.models.filter(model => model.name !== modelToRemove.name);
-
-    delete this.sequelize.models[modelToRemove.name];
   }
 
-  getModel(against, options) {
-    options = _.defaults(options || {}, {
+  getModel(against, options = {}) {
```
SimonSchick, line 20: fyi there is a difference, if you passed anything `falsy` as `options` it would default to empty object, now it will only if the value is not specified.
This change is _probably_ fine though.

lib/sequelize.js
```diff
@@ -342,8 +342,12 @@ class Sequelize {
     /**
      * Models are stored here under the name given to `sequelize.define`
      */
-    this.models = {};
-    this.modelManager = new ModelManager(this);
+    this.modelManager = new ModelManager();
+    this.models = new Proxy(this.modelManager, {
```
SimonSchick, line 346: While I was the first one to introduce proxies to sequelize, I'd rather we don't do this.
Proxies are _slow_ and shouldn't really be used at runtime (startup time is usually fine).

Issue #12102, opened 2020-04-13
MySQL Syntax Error: Functions in DEFAULT Constraint must be wrapped in parentheses


DRKSLV:
## Issue Description
Built-In Function Calls within MySQL Default constraints must be enclosed in parantheses, sequelize omits these and causes a SyntaxError.

### What are you doing?
I am trying to use a built-in db function within a default constraint. (MySQL Inno-DB)

```js
const foo = sequelize.define("foo", {
    uuid: {
        type: "BINARY(16)",
        defaultValue: Sequelize.fn("UUID_TO_BIN", Sequelize.fn("UUID")),
        primaryKey: true
    }
}, { 
    timestamps: false //to simpify the query generated
});
```

### What do you expect to happen?
As described in [this answer](https://stackoverflow.com/a/52769071) SQL-Functions within the Default Constraint must be enclosed by parentheses. If they are added to the output below, the query completes successfully.
(Running `INSERT INTO foos () VALUES ();` shows the constraint is working)

### What is actually happening?
The Output misses the required Parentheses enclosing the Function Calls, as seen below:

```
Executing (default): 
CREATE TABLE IF NOT EXISTS `foos` 
(
`uuid` BINARY(16) 
DEFAULT UUID_TO_BIN(UUID()) , 
PRIMARY KEY (`uuid`)
) ENGINE=InnoDB;
```

### Environment

- Sequelize version: 5.21.6
- Node.js version: 12.14.0
- Operating System: Windows 10 Home (1909)

### How does this problem relate to dialects?

- [ ] I don't know, I was using MySQL (InnoDB), with connector library version 2.1.0 and database version 8.0.17 (MySQL Community Server -GPL)

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] No, I don't have the time and I wouldn't even know how to start.

DRKSLV:
changing line 364 in sequelize\lib\dialects\mysql\query-generator.js to:
```template += ` DEFAULT (${this.escape(attribute.defaultValue)})`;```
seemes to have fixed this issue for me.

i dont know if this issue e xists in other dialects or if this "fix" creates other problems.
unfortunately i dont have the means to investigate this further.

FawzyMokhtar:
@DRKSLV , I came out with a temporary solution that will work for you until this bug is fixed.

You can use the `Sequelize.literal` function instead of `Sequelize.fn` function since your are not passing any external parameters to the `UUID_TO_BIN` function.
[...]

PR #12231, opened 2020-05-05
fix(mysql): Add parentheses to default value

JuarezLustosa:
### Pull Request check-list

### Description of change

- This close issue #12102. However, I did tests on MySQL version 5.7 and 8. 

####  On 5.7 Do not accept default with parentheses.

```
CREATE TABLE IF NOT EXISTS `foo2` (`uuid` CHAR(16) DEFAULT (0) );
```
We have this error 
```
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(0) )' at line 1
```

#### - On Version 8 accept 
This same query creates the table, so it works as well.

I don't know how the project works with old versions. It up to you guys.

### DIFF
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -388,7 +388,12 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
     if (!typeWithoutDefault.has(attributeString)
       && attribute.type._binary !== true
       && Utils.defaultValueSchemable(attribute.defaultValue)) {
-      template += ` DEFAULT (${this.escape(attribute.defaultValue)})`;
+
+      if (attribute.defaultValue instanceof Utils.Fn) {
+        template += ` DEFAULT (${this.escape(attribute.defaultValue)})`;
+      } else {
+        template += ` DEFAULT ${this.escape(attribute.defaultValue)}`;
+      }
     }
 
     if (attribute.unique === true) {
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -388,7 +388,7 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
     if (!typeWithoutDefault.has(attributeString)
       && attribute.type._binary !== true
       && Utils.defaultValueSchemable(attribute.defaultValue)) {
-      template += ` DEFAULT ${this.escape(attribute.defaultValue)}`;
+      template += ` DEFAULT (${this.escape(attribute.defaultValue)})`;
     }
 
     if (attribute.unique === true) {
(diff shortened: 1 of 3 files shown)


lib/dialects/mysql/query-generator.js
```diff
@@ -388,7 +388,7 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
     if (!typeWithoutDefault.has(attributeString)
       && attribute.type._binary !== true
       && Utils.defaultValueSchemable(attribute.defaultValue)) {
-      template += ` DEFAULT ${this.escape(attribute.defaultValue)}`;
+      template += ` DEFAULT (${this.escape(attribute.defaultValue)})`;
```
sushantdhiman, line 391:
```js
If (attribute.defaultValue instanceof Utils.Fn) {
   // add () to DEFAULT
} else {
  // keep old code
}
```

sushantdhiman: This should fix issue for both MySQL 5.7 & 8

JuarezLustosa: Thank You @sushantdhiman. I'll make some tests for this.

sushantdhiman:
Add a test in this file https://github.com/sequelize/sequelize/blob/master/test/integration/data-types.test.js

Create a model with `defaultValue` using function. Create a row and update it

test/integration/data-types.test.js
```diff
@@ -378,12 +378,31 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
 
     it('should parse an empty GEOMETRY field', () => {
       const Type = new Sequelize.GEOMETRY();
-
       // MySQL 5.7 or above doesn't support POINT EMPTY
       if (dialect === 'mysql' && semver.gte(current.options.databaseVersion, '5.7.0')) {
         return;
       }
 
+      it('should return parenteses when default values is a function', function() {
```
sushantdhiman, line 386: Some changes

1. This `it` block is placed inside another `it` block so it won't execute. Please move it outside of `    it('should parse an empty GEOMETRY field', () => {` block
2. You should use async/await

test/integration/data-types.test.js
```diff
@@ -378,12 +378,31 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
...
+        });
+
+        return Model.sync({ force: true }).then(() => {
+          return Model.create({
+            id: 1
+          });
+        }).then(() => {
+          return Model.findOne({ where: { id: 1 } });
+        }).then(user => {
+          expect(user.get('double')).to.eq(Infinity);
```
sushantdhiman, line 402: Model doesn't have a `double` attribute

JuarezLustosa: Hey @sushantdhiman . Thank you for your support.

I've been troubling whit this test. When I put the function on default, it is trying to create on the database, and on the test, the version of MySQL is 5.7, so it can't happen cuz this version does not support this parenthesis. What do you do in those cases?

Thank you again.
Cheers

WikiRik:

PR #12241, opened 2020-05-10
fix(sqlite): multiple primary keys results in syntax error

### LINKED ISSUE #12194: Implementing weak entity with sqlite dialect results in syntax error
### What are you doing?

```js

const weakEntity = (sequelize, dataTypes) => {
    const model = sequelize.define('weak', {});

    model.associate = (models) => {
        model.belongsTo(models.parent, {foreignKey: {primaryKey: true}});
    };

    return model;
};

const parentEntity = (sequelize, dataTypes) => {
    const model = sequelize.define('parent', {
        name: {
            type: dataTypes.STRING,
            primaryKey: true,
        },
    });

    model.associate = (models) => {
        model.hasOne(models.weak);
    };

    return model;
};

const models = {
    weak: weakEntity,
    parent: parentEntity,
};

const sequelize = new Sequelize({ dialect: 'sqlite', logging: message => console.log(message) });
Object.keys(models).forEach(modelName => {
    const modelDef = models[modelName];
    const model = modelDef(sequelize, Sequelize);
    models[modelName] = model;
});
Object.keys(models).forEach(modelName => {
    const model = models[modelName];
    if (model.associate) {
        model.associate(models);
    }
});
sequelize.authenticate()
.then(() => sequelize.sync())
.then(() => {
    console.log('Done initializing database');
})
.catch(err => {
    console.log(err);
});
```

Created table called "parent" with column called "name"
Created table called "weak" with column called "parentName" that is primary key

Error when attempting to create table called "weak" which has an autoincremented "id" column, in addition to the "parentName" expected column.

```
CREATE TABLE IF NOT EXISTS `weak` (`id` INTEGER NOT NULL NOT NULL AUTOINCREMENT, `createdAt` DATETIME NOT NULL, `updatedAt` DATETIME NOT NULL, `parentName` VARCHAR(255) NOT NULL REFERENCES `parent` (`name`) ON DELETE SET NULL ON UPDATE CASCADE, PRIMARY KEY (`id`, `parentName`));
```

Note the repeating `NOT NULL NOT NULL` for the `id` column. This results in a syntax error:

> DatabaseError [SequelizeDatabaseError]: SQLITE_ERROR: near "AUTOINCREMENT": syntax error
>     at Query.formatError (C:\Users\soryy\Desktop\mcve\node_modules\sequelize\lib\dialects\sqlite\query.js:422:16)

### Additional context

N/A

### Environment

- Sequelize version: 5.21.7
- Node.js version: 12.16.2
- Operating System: Windows 10 Home x64

- [X] I don't know, I was using sqlite, with connector library version 4.2.0

soryy708:
### Pull Request check-list

### Description of change

Closes #12194
Same as https://github.com/sequelize/sequelize/pull/12237

### DIFF
--- a/lib/dialects/sqlite/query-generator.js
+++ b/lib/dialects/sqlite/query-generator.js
@@ -44,7 +44,11 @@ class SQLiteQueryGenerator extends MySqlQueryGenerator {
 
           if (needsMultiplePrimaryKeys) {
             primaryKeys.push(attr);
-            dataTypeString = dataType.replace('PRIMARY KEY', 'NOT NULL');
+            if (dataType.includes('NOT NULL')) {
+              dataTypeString = dataType.replace(' PRIMARY KEY', '');
+            } else {
+              dataTypeString = dataType.replace('PRIMARY KEY', 'NOT NULL');
+            }
           }
         }
         attrArray.push(`${this.quoteIdentifier(attr)} ${dataTypeString}`);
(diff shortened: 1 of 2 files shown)


soryy708:
The failure is with the MariaDB dialect. I think this is irrelevant.

PR #12248, opened 2020-05-12
fix: typo in JSDoc for PG array operators

segeda:
In JSDoc are PG array operators with bad closing bracket

### DIFF
--- a/docs/manual/core-concepts/model-querying-basics.md
+++ b/docs/manual/core-concepts/model-querying-basics.md
@@ -427,14 +427,14 @@ Keep in mind, the provided range value can [define the bound inclusion/exclusion
 
 ```js
 [Op.contains]: 2,            // @> '2'::integer  (PG range contains element operator)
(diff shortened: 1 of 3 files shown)


docs/manual/core-concepts/model-querying-basics.md
```diff
@@ -427,14 +427,14 @@ Keep in mind, the provided range value can [define the bound inclusion/exclusion
 
 ```js
 [Op.contains]: 2,            // @> '2'::integer  (PG range contains element operator)
-[Op.contains]: [1, 2],       // @> [1, 2)        (PG range contains range operator)
```
sushantdhiman, line 430: These comments are correct, `[1,2)` signifies inclusive/exclusive bounds. Lower bound is inclusive and upper bound is exclusive by default

https://www.postgresql.org/docs/9.5/rangetypes.html

segeda: Thank you for clarification. I close PR

Issue #12266, opened 2020-05-17
Make sequelize correctly append table name to "having" clause


weagle08:
**Issue**
When creating a "having" clause sequelize does not currently split on dot notation or append the correct table name to the query.

*Example No.1:*

```
query.having = {
    amount: {
        [Op.gt]: fn('sum', col('payments.amount'))
    }
};

// this gives an "ambigous reference to column amount" error
```

*Example No.2*

```
query.having = {
    ['LeaseAgreement.amount']: {
        [Op.gt]: fn('sum', col('payments.amount'))
    }
};

// this gives a "column LeaseAgreement.amount does not exist" error
```

**Fix description**

The proposed fix simply treats the having clause as just part of the where appending the "HAVING" statement so that Example No.1 functions as expected.

PR #12267, opened 2020-05-17
fix "HAVING" statement to correctly prepend table name

weagle08:
Addresses issue #12266


weagle08:
needs to be done on v5

PR #12275, opened 2020-05-19
Fix tedious connect deprecation v5

UziTech:
#12182 backport to v5


UziTech:
I don't believe the failing tests have anything to do with my changes since it should only affect mssql and the failing test is on mariadb

PR #12281, opened 2020-05-20
fix(model destroy): Return 0 when options is truncate. 

### LINKED ISSUE #12164: Model.destroy returns NaN as destroyed count
### What are you doing?

Calling `destroy()` on `Model` in order to truncate table. Based on documentation this call should return `Promise<number>` with count of destroyed entries.

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/72 

Count of destroyed entries

`NaN is returned instead`

### Environment

- Sequelize version: 5.21.7
- Node.js version: v12 
- Operating System: macos
- If TypeScript related: TypeScript version: 3.8

- [x] I think this problem happens only for the following dialect(s): postgres, MySQL, mariadb

- [x] No, I don't have the time and I wouldn't even know how to start.

JuarezLustosa:
### Pull Request check-list

### Description of change

This close #12164.
 
I was debugging and realize, the [SSCE](https://github.com/papb/sequelize-sscce/pull/72) examples is passing truncate as an option. Therefore truncate on the database does not show affected rows so we can't return the value, thus I found the Postgres NaN in these cases.
 
- I changed the doc adding info about it when use truncate will no show rows affected only on Sqlite do this.
- Fixed Postgres rowCount when queryResult.rowCount isn't returning 
- Fixed MSsql  when data is empty return 0 on isBulkDeleteQuery

### DIFF
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -214,7 +214,7 @@ class Query extends AbstractQuery {
       return data.length;
     }
     if (this.isBulkDeleteQuery()) {
-      return data[0] && data[0].AFFECTEDROWS;
+      return data[0] && data[0].AFFECTEDROWS || 0;
     }
     if (this.isVersionQuery()) {
       return data[0].version;
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -214,7 +214,7 @@ class Query extends AbstractQuery {
       return data.length;
     }
     if (this.isBulkDeleteQuery()) {
-      return data[0] && data[0].AFFECTEDROWS || 0;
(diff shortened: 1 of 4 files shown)


lib/dialects/mssql/query.js
```diff
@@ -214,7 +214,7 @@ class Query extends AbstractQuery {
       return data.length;
     }
     if (this.isBulkDeleteQuery()) {
-      return data[0] && data[0].AFFECTEDROWS;
+      return data[0] && data[0].AFFECTEDROWS || 0;
```
sobrinho, line 217: Probably:

```suggestion
      return data[0] ? data[0].AFFECTEDROWS : 0;
```

JuarezLustosa: Nice one!

Issue #12317, opened 2020-05-26
Mysql logging can lose CLS context when benchmark flag is true


Mikxail:
## Issue Description
Mysql logging can lose CLS context when benchmark flag is true

### What are you doing?

```js
const Sequelize = require('sequelize');
const cls = require('cls-hooked');
const ns = cls.createNamespace('aabb');

const dbName = '';
const dbUser = '';
const dbPassword = '';
const dbHost = '';

Sequelize.useCLS(ns);

const s = new Sequelize(
    dbName,
    dbUser,
    dbPassword,
    {
        dialect: 'mysql',
        host: dbHost,
        benchmark: true,
        logging: (msg, duration, attrs) => {
            const match = msg.match(/SELECT\s(\d+)/);
            if (match[1] !== String(ns.get('logger'))) {
                console.error('CLS is losed. Expected: ' + ns.get('logger') + '. Actual: ' + match[1]);
            }
        }
    },
);

const execQuery = async (v = 0) => {
    ns.set('logger', v);
    const resp = await s.query(`SELECT ${v}`);
    return resp;
};

for (let idx=0; idx<10; idx++) {
    ((v) => {
        ns.runPromise(() => {
            return Promise.all([
                execQuery(v),
            ])
        });
    })(idx);
}
```

### What do you expect to happen?
Noting in error log;

### What is actually happening?
Error log contains: `CLS is losed. Expected: 2. Actual: 9`

### Environment

- Sequelize version: 5.21.4
- Mysql2 version: 1.7.0 / 2.1.0
- Node.js version: v12.16.3
- Operating System: Linux

### How does this problem relate to dialects?

- [x] I think this problem happens only for the following dialect(s): mysql

### Possible fix:
file: `lib/dialects/mysql/query.js`
```js
    return new Utils.Promise((resolve, reject) => {
        const handler = (err, results) => {
-        complete(); // remove complete function
// ........
    // Log warnings if we've got them.
      .then(results => {
+        complete(); // add complete function
        if (showWarnings && results && results.warningStatus > 0) {
          return this.logWarnings(results);
        }
        return results;
      })
```

sushantdhiman:
Fixed by https://github.com/sequelize/sequelize/pull/12328

PR #12328, opened 2020-05-30
fix(logger): use promise.then for processing query results(mysql)

Mikxail:
### Pull Request check-list

### Description of change

Using promise.then instead of custom handler function for processing run query results for mysql dialect.
It change fix losing CLS context for mysql logging custom function in case of `benchmark: true` #12317
MR for v5: #12318

### DIFF
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -50,40 +50,23 @@ class Query extends AbstractQuery {
 
     const complete = this._logQuery(sql, debug, parameters);
 
-    const result = await new Promise((resolve, reject) => {
-      const handleTransaction = err => {
-        if (err) {
-          err.sql = sql;
-          err.parameters = parameters;
-          reject(this.formatError(err));
-          return;
-        }
-        resolve(this.formatResults());
-      };
+    const query = new Promise((resolve, reject) => {
       // TRANSACTION SUPPORT
       if (sql.startsWith('BEGIN TRANSACTION')) {
-        return connection.beginTransaction(handleTransaction, options.transaction.name, connection.lib.ISOLATION_LEVEL[options.isolationLevel]);
+        return connection.beginTransaction(error => error ? reject(error) : resolve([]), options.transaction.name, connection.lib.ISOLATION_LEVEL[options.isolationLevel]);
       }
       if (sql.startsWith('COMMIT TRANSACTION')) {
-        return connection.commitTransaction(handleTransaction);
+        return connection.commitTransaction(error => error ? reject(error) : resolve([]));
       }
       if (sql.startsWith('ROLLBACK TRANSACTION')) {
-        return connection.rollbackTransaction(handleTransaction, options.transaction.name);
+        return connection.rollbackTransaction(error => error ? reject(error) : resolve([]), options.transaction.name);
       }
       if (sql.startsWith('SAVE TRANSACTION')) {
-        return connection.saveTransaction(handleTransaction, options.transaction.name);
+        return connection.saveTransaction(error => error ? reject(error) : resolve([]), options.transaction.name);
       }
-      const results = [];
-      const request = new connection.lib.Request(sql, (err, rowCount) => {
 
-        if (err) {
-          err.sql = sql;
-          err.parameters = parameters;
-          reject(this.formatError(err));
-        } else {
-          resolve(this.formatResults(results, rowCount));
-        }
-      });
+      const rows = [];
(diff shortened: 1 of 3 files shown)


papb:

Please add at least one test to make sure your implementation works as intended and to prevent regressions in the future. You can write a test based on the example code you wrote in the issue. Take a look into recent merged PRs to see how others created tests. If you have trouble, let me know.

lib/dialects/mysql/query.js
```diff
@@ -33,32 +33,25 @@ class Query extends AbstractQuery {
...
+      : new Promise((resolve, reject) => connection.query({ sql }, (error, result) => error ? reject(error) : resolve(result)).setMaxListeners(100));
 
-        if (err) {
-          // MySQL automatically rolls-back transactions in the event of a deadlock
-          if (options.transaction && err.errno === 1213) {
-            options.transaction.finished = 'rollback';
-          }
-          err.sql = sql;
-          err.parameters = parameters;
+    const complete = this._logQuery(sql, debug, parameters);
```
papb, line 40: Please move this line above the query calls (as it was before your change).

lib/dialects/mysql/query.js
```diff
@@ -33,32 +33,25 @@ class Query extends AbstractQuery {
...
+      if (options.transaction && err.errno === 1213) {
+        options.transaction.finished = 'rollback';
       }
+      err.sql = sql;
+      err.parameters = parameters;
+
+      throw this.formatError(err);
     });
+
+    complete();
```
papb, line 53: Nice, I think it makes sense for `complete()` to be called only if no error happens. Good change :+1:

lib/dialects/mysql/query.js
```diff
@@ -33,32 +33,25 @@ class Query extends AbstractQuery {
...
-        } else {
-          resolve(results);
-        }
-      };
-      if (parameters) {
-        debug('parameters(%j)', parameters);
-        connection.execute(sql, parameters, handler).setMaxListeners(100);
-      } else {
-        connection.query({ sql }, handler).setMaxListeners(100);
+    const results = await query.catch(err => {
```
papb, line 42: Can you please refactor this `.catch` usage into a try-catch block?

sushantdhiman:
Reminder this needs test before it can be reviewed :)

This issue must be affecting other dialects as well, having a test case will ensure that

lib/dialects/mysql/query.js
Mikxail: Done

lib/dialects/mysql/query.js
Mikxail: Sure. Done.

Mikxail:
@papb Thank you for review. I made the requested changes. 
Also, I made changes in dialect/mssql for fixing same issue, but I hasn't idea why mssql test case brokes. Locally all tests passed.

lib/dialects/mssql/query.js
```diff
@@ -112,6 +110,10 @@ class Query extends AbstractQuery {
 
       connection.execSql(request);
     });
+
+    complete();
```
sushantdhiman, line 114: This means, benchmark will now include Sequelize record parsing overhead.

Earlier `benchmark = time to execute query by dialect`
Now `benchmark = time to execute query by dialect + time taken by sequelize to format result`

This deviation can be significant when parsing large number of records.

Main question is why placing `complete()` here works, do you have any ideas @Mikxail

Mikxail: Actually, I don't know how `cls-hooked` works under the hood. But I can assume that the problem in `connection.lib.Request` that saves `callback` function in variable, and then call `callback` from `execSql`. Probably that can breaks `cls-hooked` algorithms.
A similar situation was in the `run` function (dialects/mysql/query.js).

I can refactor `_run` function (dialects/mssql/query.js), but it require a lot of changes. Can I do it withing this MR or I should create separate MR?

sushantdhiman: I think issues listed below are variant of this issue
- https://github.com/nodejs/node/issues/21078
- https://github.com/Jeff-Lewis/cls-hooked/issues/36

At Sequelize level, we can extract record/error formatting away from callback. Something like this should work

1. Get raw result/error from callback where `triggerAsyncId` is incorrect (new Promise wrapper can be used)
2. Call `complete`
3. Parse error or raw results and return final result.

Mikxail: I agree with it.
I did refactoring `_run` functions in dialects/mssql.

PR #12337, opened 2020-06-03
add support for optional values in "where" clauses

### LINKED ISSUE #11564: Typing of findAll seems to be missing the support of undefined values in where
## Issue Description
When using sequelize with typescript, we run into the issue of not being able to use objects that have potentially undefined attributes for the `where` attribute of a `findAll` call. 

### What are you doing?

```js

class class user extends Sequelize.Model {
/* define your table here, it shouldn't matter for types */
}
const queryObj:{id:string, optionalValue?:string} = { id: 'someID'};

user.findAll({ where: queryObj }); // This is where the issue appears
```

According to my understanding of the discussion in [issue #8583](https://github.com/sequelize/sequelize/issues/8583) the function should support undefined values as input. (If as null, or ignoring it completely, is irrelevant for this issue)

We get type errors, not allowing undefined values inside the `where`object.

### Additional context

none

### Environment

- Sequelize version: 5.19.6
- Node.js version: ^10, should not be relevant in any way
- Operating System: OS X and WIN 10, should also make no difference
- If TypeScript related: TypeScript version: 3.6.2

- [x] Yes, I have the time and I know how to start. (if it is simply adjusting the types)

T0TProduction:
### Pull Request check-list

- [x] Have you added new tests to prevent regressions? _not needed_
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? _potentially?_

### Description of change

fixes #11564 
as discussed in the issue, this might not be the wanted solution. I'm opening a PR though because the issue has been dead for a while and as long as there is no real reason against this change, it will fix the issue.


sushantdhiman:
Please add a test here https://github.com/sequelize/sequelize/blob/master/types/test/where.ts

T0TProduction:
@sushantdhiman done, I hope I understood how the test is supposed to work 😁

Issue #12325, opened 2020-05-29
findOrCreate should accept 'fields' to save


sonirico:
### Is your feature request related to a problem? Please describe.

`CreateOptions` has a optional property named 'fields' which describes which fields are writable, in order to prevent additional and/or malicious fields sent on payloads persisted in databases. However, `FindOrCreateOptions` does not provide such configuration to be set. Is that intentional?

### Describe the solution you'd like
Some configuration flag to enable writes applicable to certain fields, as `CreateOptions` does.

### Is this issue dialect-specific?

- [x] No. This issue is relevant to Sequelize as a whole.
- [ ] Yes. This issue only applies to the following dialect(s): XXX, YYY, ZZZ

### Would you be willing to resolve this issue by submitting a Pull Request?

- [x] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

sushantdhiman:
`findOrCreate` does pass options to `create` method. I think typings can be improved. Please send a PR if you can spare some time

sonirico:
Ok, I will spare some time to do it this week.

sushantdhiman:
Fixed by 12338, Thanks @sonirico

PR #12338, opened 2020-06-03
fix: add missing fields to 'FindOrCreateType'

sonirico:
### Pull Request check-list

### Description of change

This PR includes some typings needed to provide clarification in regards to the usage of `findOrCreate` functionality, thus the `FindOrCreateOptions` interface has been altered to include the `fields` string array to whitelist which fields should be persisted as explained in https://github.com/sequelize/sequelize/issues/12325 . Tests are not passing, as well as in master upstream.

### DIFF
--- a/types/test/model.ts
+++ b/types/test/model.ts
@@ -99,6 +99,20 @@ UserModel.findCreateFind({
 })
 
 /**
+ * Tests for findCreateFind() type.
+ */
+
+ UserModel.findOrCreate({
+  fields: [ "jane.doe" ],
+  where: {
+    username: "jane.doe"
+  },
+  defaults: {
+    username: "jane.doe"
+  }
+})
+
+/**
  * Test for primaryKeyAttributes.
  */
 class TestModel extends Model {};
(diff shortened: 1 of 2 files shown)


sushantdhiman:
Please add a test here https://github.com/sequelize/sequelize/blob/master/types/test/model.ts

sonirico:
@sushantdhiman I've added a handful of tests 👍

types/test/model.ts
```diff
@@ -98,6 +98,86 @@ UserModel.findCreateFind({
...
+      fields: [ 'beta_user', 'username' ],
+      where: {
+        username: "test.user",
+      },
+      defaults: {
+        username: "test.user"
+      }
+    })
+
+    if (! created) {
```
sushantdhiman, line 116: These tests are supposed to only testing syntax, not actually doing any db operations. I don't think these conditions or checks should be included in this code.

Just add some calls to `<Model>.findOrCreate` to assert correct typings or move these to `e2e` if you prefer that

sonirico: Ok, I have added only the typings because I have seen enough tests for `findOrCreate` in the integrations folder. Thank you for your guidance!

types/test/model.ts
```diff
@@ -98,6 +98,20 @@ UserModel.findCreateFind({
   }
 })
 
+/**
+ * Tests for findCreateFind() type.
```
sushantdhiman, line 102:
```suggestion
 * Tests for findOrCreate() type.
```

PR #12360, opened 2020-06-09
Better typing support for Model.init()

OwenCalvin:
By converting the `ModelAttribute` interface into an advanced type, you are able to add an autocompletion on the fields while using `Model.init()`.

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -1338,12 +1338,9 @@ export interface ModelAttributeColumnOptions<M extends Model = Model> extends Co
 /**
  * Interface for Attributes provided for a column
  */
-export interface ModelAttributes<M extends Model = Model> {
-  /**
-   * The description of a database column
-   */
-  [name: string]: DataType | ModelAttributeColumnOptions<M>;
-}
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Tests are failing, you should be able to run those test locally.

Additionally I don't think we support class member for attribute definition in native types, perhaps you are using `sequelize-typescript`. This also means if there a method defined as class member, it will also be eligible name for attribute.

I don't think I can accept this change

OwenCalvin:
Now it works like exaclty like the current Sequelize behaviour expect that you have the autocompletion. Yes the methods appears but I think that's better than showing nothing, I think that there will be less confusion in the code

sushantdhiman:
@SimonSchick Source implementation keep attributes as `Model.rawAttributes`, not sure how we can implement decorators with just typings in this repo. I guess that will require source to move attributes to class properties, I don't think that is going to happen in v6 at-least.

This PR feels like a solution for `sequelize-typescript` specific decorator based model.

Can't this be patched by `sequelize-typescript`? Natively in main implementation, model attributes are not stored as class properties.

Issue #4361, opened 2015-08-21
i18n on validation and database errors


GTTo:
Hello!
When  I create a model and I define a validation over an attribute, I can define my own messages:
````
sequelize.define('Foo', {
  foo: {
    type: Sequelize.STRING,
    validate: {
      isInt: {
         msg: "Must be an integer number of pennies"
       }
    }
  }
}
````
In this case, you can translate the message as you want.

When an sql validation error is thrown, the messages is only in English
lib/data-types.js
````
INTEGER.prototype.validate = function(value) {
  if (!Validator.isInt(value)) {
    throw new sequelizeErrors.ValidationError(util.format('%j is not a valid integer', value));
  }
  return true;
};
````
How can I translate this messages? Why It doesn't have an i18n file to translate it?

Thank you!

GTTo:
there is anyone out there?

janmeier:
@GTTo , yes, but we all have jobs, school, family, etc. on the side - so response is not always instantaneous even though we try to be fast. :)

I'm not really sure this is a complexity we wan't to handle in sequelize. When you're at the point were your app needs i18n I think it should be handled somewhere else than directly in the ORM. 

I do realize that the particular error you're refering to is hard to translate - it's a built in validation, and has nothing to do with the `isInt` validator - see https://github.com/sequelize/sequelize/pull/3472. So the only way to find out that the validationerror comes from this validator is to do string matching, which is ugly.

We should probably add a `type` or similar to the error / use errorItem so users can check which validation actually failed

cc @mickhansen and possibly @fixe

fixe:
Since string matching doesn't work in this case because the value is dynamic I'm +1 for providing a `type` or `code` associated with the validation error.

billxinli:
I am facing this issue as well, application is I18N, however, I don't really want to handle translation in the model/ORM layer. I also don't want to do string matching of if err === "this message" then translate with this key

Since Sequelize can be used in conjunction with other libraries, I am wondering would it make sense to have a possible to have an abstract callback layer to handle the translations? Users now can have a method to provide a set of override to handle translations, if one isn't provided, sequelize will fallback to the default hardcoded strings?

misterdai:
[...]

PR #12380, opened 2020-06-17
Contextable: eject context to model's _options

oney:
### Pull Request check-list

## Description

This PR adds the ability to eject a context value to model's _options when finding, creating and building models.
With ejecting context to models, implementing something like i18n will be much easier. So this PR can solve #4361.

### Usage
```js
const User = sequelize.define("user", {
  age: {
    type: DataTypes.INTEGER,
    validate: {
      isEven(value) {
        if (parseInt(value) % 2 !== 0) {
          throw new Error(this._options.context.i18n("isEvenError"));
        }
      }
    }
  },
});

const i18n = function() {
  return "localized message";
}
const b = await User.create({ age: 21 }, {
  context: { i18n }
});
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1765,7 +1765,7 @@ class Model {
           if (result._options && result._options instanceof Object) {
             result._options.context = options.context;
           }
-        })
+        });
       }
       if (options.hooks) {
         return this.runHooks('afterFind', results, options);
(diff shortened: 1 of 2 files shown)


sushantdhiman:
IMO `i18n` is usually user context / session based, this means you ideally don't want to specify `i18n` or `language` option for each query.

This solution only implements ( sort of ) translations for custom validations, which is already possible, as error messages are dev defined for custom validations.

Closing this PR as this not an acceptable solution for `i18n`

PR #12405, opened 2020-06-22
feat(types): added optional stricter typing for Model attributes

hsource:
### Pull Request check-list

- [X] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)? **Also run `npx tsc --project types/test/` and `npx tsc --project types/`

### Description of change

This code allows users to have typechecking for model attributes in all relevant functions in the `Model` class, as well as some functions outside of the `Model` class. This was inspired by the [Sequelize v4 DefinitelyTyped typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sequelize/index.d.ts), but is made to be almost fully backwards compatible for users of the current typings.

Specific things I tried to ensure:

- Works with `sequelize.define` as well as `new Model` formats of building models
- Use the attributes as much as possible
- Allows for omitting the types - everything still works fine without them

Here are the things I changed:

1. Added 2 dummy variables (`_attributes` and `_creationAttributes`) to the Model class so that the static functions can be typed correctly
   - Used the dummy types to get the attributes type in static functions in `model.d.ts`
   - Before this, I'd tried doing the following, but Typescript fails to infer the `TAttributes` automatically
```ts
  public static findOne<M extends Model<TAttributes>, TAttributes>(
    this: { new(): M },
    options: NonNullFindOptions<TAttributes>
  ): Promise<M>;
```
   - Used the dummy types to get the attributes type in static functions in `model.d.ts`
2. Retyped some functions outside of `model.d.ts` to use the new typings, and made others just fallback to just using `any` for the generic arguments
3. Added tests and updated existing tests in `types/tests`

#### Testing

In addition to adding a few tests and modifying the tests for `User.ts` and `define.ts` to use the new typings, I also have this running on a codebase with ~20k LOC for https://wanderlog.com - everything typechecks correctly, and it's an improvement from our v4 typings.

#### Questions for reviewers

- Currently, some of the `Option` interfaces like `FindOptions` now has a required generic parameter. Should I make this parameter also optional to make it fully backwards-compatible?
  - This means users who had their own function like `doFancyFind(option: FindOptions)` won't need to change it to `doFancyFind(option: FindOptions<any>)` with this change
- Is it OK to just bundle an `Optional` type in the typings in `index.d.ts`? It's very useful for constructing the optional types and can help


types/test/errors.ts
```diff
@@ -5,7 +5,7 @@ import { OptimisticLockError } from '../lib/errors';
 
 async function test() {
   try {
-    await User.create({ username: 'john_doe' });
```
hsource, line 8: I made `firstName` required in the attributes I defined, so I updated this code to match

types/lib/model.d.ts
```diff
@@ -685,28 +683,28 @@ export interface Hookable {
 /**
  * Options for Model.findOrCreate method
  */
-export interface FindOrCreateOptions extends Filterable, Logging, Transactionable {
+export interface FindOrCreateOptions<TAttributes, TCreationAttributes> extends FindOptions<TAttributes> {
```
hsource, line 686: I made this code just extend FindOptions as other options (like `attributes`) actually are accepted too, per the code inside `model.js` which just directly does a `findOne` with the options.

hsource:
EDIT: I'm still having trouble getting the typings to work well for both users of `db.define` and `class MyModel extends Model`

If I type the functions like this in `model.d.ts`:

```ts
  public static findOne<M extends Model>(
    this: ModelCtor<M>,
    options: NonNullFindOptions<M['_attributes']>
  ): Promise<M>;
  
// ...
export type ModelCtor<M extends Model> = { new(): M } & typeof Model;
```

I get errors on the `new Model` path like this:

```
Error:(49, 1) TS2684: The 'this' context of type 'typeof User' is not assignable to method's 'this' of type 'ModelCtor<User>'.
  Type 'typeof User' is not assignable to type 'typeof Model'.
    Construct signature return types 'User' and 'Model<TModelAttributes, TCreationAttributes>' are incompatible.
      The types of '_attributes' are incompatible between these types.
        Type 'UserAttributes' is not assignable to type 'TModelAttributes'.
          'UserAttributes' is assignable to the constraint of type 'TModelAttributes', but 'TModelAttributes' could be instantiated with a different subtype of constraint '{}'.
```

However, if I type the functions like this, I get errors on the define path.

```ts
  public static findOne<M extends Model>(
    this: ModelStatic<M>,
    options: NonNullFindOptions<M['_attributes']>
  ): Promise<M>;
  
// ...
export type ModelStatic<M extends Model> = { new(): M };
```

hsource:
I made some fixes to the documentation and this is now ready for review!

sushantdhiman:
> Currently, some of the Option interfaces like FindOptions now has a required generic parameter. Should I make this parameter also optional to make it fully backwards-compatible?
>   - This means users who had their own function like doFancyFind(option: FindOptions) won't need to change it to doFancyFind(option: `FindOptions<any>`) with this change

Yes make those generic parameters optional, we should avoid breaking change if we can.

> Is it OK to just bundle an Optional type in the typings in index.d.ts? It's very useful for constructing the optional types and can help

No problem with exporting this helper type

types/lib/model.d.ts
```diff
@@ -1859,77 +1912,81 @@ export abstract class Model<T = any, T2 = any> extends Hooks {
...
 
   /**
    * Builds a new model instance and calls save on it.
    */
   public static create<M extends Model>(
-    this: { new(): M } & typeof Model,
-    values?: object,
-    options?: CreateOptions
+    this: ModelStatic<M>,
+    values?: M['_creationAttributes'],
```
sushantdhiman, line 1969: It is possible to create record with `id` i.e. primary key. Distinction between `CreationAttributes` and `Attributes` from user point of view make sense, but internally all attributes are accepted.

We can also think about accepting `TAttributes` generic which contains all attributes, then create a dynamic type for "CreationAttributes" by making primary key attribute optional in that type

hsource:
> > Currently, some of the Option interfaces like FindOptions now has a required generic parameter. Should I make this parameter also optional to make it fully backwards-compatible?
> > 
> > * This means users who had their own function like doFancyFind(option: FindOptions) won't need to change it to doFancyFind(option: `FindOptions<any>`) with this change
> 
> Yes make those generic parameters optional, we should avoid breaking change if we can.

I just pushed a new change that made all types optional. I reverted some of the changes adding explicit `<any>` types to the `types/test` dir to test it. Everything seems to typecheck correctly!

types/lib/model.d.ts
hsource: This is an interesting idea! The problem is - how do we figure out what's the primary key? We'd probably need users to provide it in a new generic (maybe `TPrimaryKey`) but then that's just one more thing to provide. And then, there are still fields with `defaultValue` which users can also omit, and other cases we'd need to handle.

I think `TCreationAttributes` is a pretty flexible solution that solves for both primary keys and other columns with default values that can be omitted. It's also been used for the previous sequelize v4 typings, so would be familiar to those users.

sushantdhiman: Right, there is no good way to infer primary key in current Typescript schema. Actual attributes are passed to `init` or `define`, which contains this info.

hsource: Yeah, I think defining it in `init` is especially problematic, so I recommend just leaving it as-is.

zaggino:
Just a question, would it be possible somehow to drop repeating of this?

image

hsource:
> Just a question, would it be possible somehow to drop repeating of this?
> 
> image

Ideally, we'd be able to extend multiple classes and be able to do something like this below. But Typescript does not have multiple inheritance:

```ts
class UserAttributes {
  public id!: number;
  public username!: string;
  public firstName!: string;
  public lastName!: string;
  public groupId!: number;
}

export class User extends Model<UserAttributes, UserCreationAttributes>, UserAttributes {
  public static associations: {
    group: BelongsTo<User, UserGroup>;
  };

  public createdAt!: Date;
  public updatedAt!: Date;

  // mixins for association (optional)
  public group?: UserGroup;
  public getGroup!: BelongsToGetAssociationMixin<UserGroup>;
  public setGroup!: BelongsToSetAssociationMixin<UserGroup, number>;
  public createGroup!: BelongsToCreateAssociationMixin<UserGroup>;
}
```

This isn't possible because Typescript doesn't support multiple inheritance, but you might be able to achieve something similar with Mixins - https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#putting-all-of-the-above-rules-together-in-an-example

knoid:
I can't find a way to make this work when doing a create with include. e.g.:
```js
class User extends Model<UserAttributes, UserCreationAttributes> {...}
class Project extends Model<ProjectAttributes, ProjectCreationAttributes> {...}
Project.hasMany(User);
Project.create({
  users: [{ name: 'project owner' }],
}, { include: [User] });
```

I get the error: "Object literal may only specify known properties, and 'users' does not exist in type `ProjectCreationAttributes`". I'd add `users` into `ProjectCreationAttributes` but that's just a `Optional<ProjectAttributes, ...>` and even if I did add it. What exactly should I add? `UserAttributes[]` or `UserCreationAttributes[]`? I hope we can fix this problem, I love this new TS approach.

hsource:
> I can't find a way to make this work when doing a create with include. e.g.:
> 
> ```js
> class User extends Model<UserAttributes, UserCreationAttributes> {...}
> class Project extends Model<ProjectAttributes, ProjectCreationAttributes> {...}
> Project.hasMany(User);
> Project.create({
>   users: [{ name: 'project owner' }],
> }, { include: [User] });
> ```
> 
> I get the error: "Object literal may only specify known properties, and 'users' does not exist in type `ProjectCreationAttributes`".
> 
> I'd add `users` into `ProjectCreationAttributes` but that's just a `Optional<ProjectAttributes, ...>` and even if I did add it. What exactly should I add? `UserAttributes[]` or `UserCreationAttributes[]`? I hope we can fix this problem, I love this new TS approach.

For includes, you should just manually add it via a cast. The old typings didn't really handle it either, and it'd be complicated to near-impossible to make this work automatically.

Here's what we do in our code

```ts
interface ProjectAttributesWithUser extends ProjectAttributes {
  User: UserAttributes;
}

const project = await Project.findAll({ ...someOptions, include: [User] }) as ProjectAttributesWithUser[];
```

PR #12410, opened 2020-06-23
fix (mssql): bulkUpdate returning values

ShaharHD:
### Pull Request check-list

### Description of change

Fix MSSQL returning values for bulkUpdate

The issue creates a major performance hit for v5 - as bulkUpdates right now does not work correctly and requires to run another query to get the model data - which is returning already in the update itself.

- bulkUpdate will now return the updates if options.returning is true.
- updateQuery removed redundant setting check
- Fix documentation for model.update

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -393,7 +393,7 @@ class QueryGenerator {
 
           suffix += selectFromTmp;
         }
-      } else if (this._dialect.supports.returnValues && options.returning) {
+      } else if (options.returning) {
         // ensure that the return output is properly mapped to model fields.
         options.mapToModel = true;
         suffix += ' RETURNING *';
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -393,7 +393,7 @@ class QueryGenerator {
 
           suffix += selectFromTmp;
         }
-      } else if (options.returning) {
+      } else if (this._dialect.supports.returnValues && options.returning) {
         // ensure that the return output is properly mapped to model fields.
         options.mapToModel = true;
         suffix += ' RETURNING *';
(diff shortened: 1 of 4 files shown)


ShaharHD:
I'm failing on returning values with `returning:false`, I'll work on this tomorrow and add a commit to fix this test as well

ShaharHD:
The CI fails on an `npm install` error for some reason? (unrelated to anything that I touched)

sushantdhiman:
You need to submit a fix to master branch before this can be merged

ShaharHD:
> You need to submit a fix to master branch before this can be merged

#12413 was created as requested.
But the issue has a major preformance hit for v5 - as bulkUpdates right now does not work correctly and requires in the code to run another query to get the model data - which is returning already in the update itself.

lib/dialects/abstract/query-generator.js
```diff
@@ -393,7 +393,7 @@ class QueryGenerator {
 
           suffix += selectFromTmp;
         }
-      } else if (this._dialect.supports.returnValues && options.returning) {
```
sushantdhiman, line 396: Keep this check, first check test if dialect supports returnings like MySQL which doesn't support it. Second if user is requesting to return records

sushantdhiman:
> The issue creates a major performance hit for v5

I didn't understand this, are you saying this change will improve performance?

lib/dialects/abstract/query-generator.js
ShaharHD: Look at line 367 - you won't even get into this if statement if `this._dialect.supports.returnValues` was false. why recheck for it?

ShaharHD:
> > The issue creates a major performance hit for v5
> 
> I didn't understand this, are you saying this change will improve performance?

Yes.

Consider this scenario:

```js
const [rowsUpdate, [updatedBook]] = Book.update(
   {price: 100},
   {returning: true, where: {price: 0} }
 )
```

Without this PR - `updatedBook` will be `undefined` (and not populated with the updated records returned by MSSQL `OUTPUT INSERTED.*`) which means I'll have to issue another query to pull the data (and in some cases its not possible as the data might have already changed)

ShaharHD:
@sunshinewyin Added the new test to this branch as well as `master`

Look for `should output the updated record` in the tests.

PR #12413, opened 2020-06-23
fix(mssql): returning data for bulkUpdate

ShaharHD:
### Pull Request check-list

### Description of change

Fix for the MSSQL dialect to return changed records when using bulkUpdate with `returning: true`

- bulkUpdate will now return the updates if options.returning is true.
- Fix documentation for model.update

This is in relation to #12410 as requested

### DIFF
--- a/lib/dialects/mssql/query.js
+++ b/lib/dialects/mssql/query.js
@@ -197,6 +197,10 @@ class Query extends AbstractQuery {
       return data[0];
     }
     if (this.isBulkUpdateQuery()) {
+      if (this.options.returning) {
+        return this.handleSelectQuery(data);
+      }
+
       return rowCount;
     }
     if (this.isBulkDeleteQuery()) {
--- a/test/integration/model/update.test.js
+++ b/test/integration/model/update.test.js
@@ -134,6 +134,25 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       });
     }
 
+    if (_.get(current.dialect.supports, 'returnValues.output')) {
+      it('should output the updated record', async function() {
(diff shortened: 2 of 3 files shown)


sushantdhiman:
No tests for this change? Is this already covered by previous tests

ShaharHD:
> No tests for this change? Is this already covered by previous tests

You do have tests for `returning: true` I can modify those tests (in the mssql) to verify the updated data is returned.

I can see something similar on the postgres tests - but the postgres tests seems way more complex than what you have to MSSQL.
I can add a simple test which shows with `returning: true` to replicate the one test from postgres on bulkUpdate

sushantdhiman:
Here is the relevant test file

https://github.com/sequelize/sequelize/blob/77b5fe6f4a08eae4c1cc7393e525174f7009edde/test/integration/model/update.test.js#L120

You need to modify `if` check to this

```js
if (current.dialect.supports.returnValues) {
```

ShaharHD:
> You need to modify `if` check to this
> 
> ```js
> if (current.dialect.supports.returnValues) {
> ```

Shouldn't it be this?
`if (_.get(current.dialect.supports, 'returnValues.output')) {`

ShaharHD:
> You need to modify `if` check to this
> 
> ```js
> if (current.dialect.supports.returnValues) {
> ```

@sunshinewyin I've kept the same syntax as the old test, so I used:
```js
if (_.get(current.dialect.supports, 'returnValues.output')) {
```

Look for `should output the updated record` in the tests.
Note the `firstAcc.reload()` as MSSQL only returns updated fields, so I've added this to still complete the test as the postgres one does.

PR #12445, opened 2020-06-30
feat(types): Add ModelDefined type as syntactic sugar

Keimeno:
### Pull Request check-list

### Description of change

The current way to define a module in a functional way with Typescript works like this:

```ts
const Note: ModelCtor<
  Model<
    NoteAttributes, 
    NoteCreationAttributes
  >
> = sequelize.define(
  ...
);
```

However this approach is a bit much to write, and with the `ModelDefined` type it could be reduced to:

```ts
const Note: ModelDefined<
  NoteAttributes, 
  NoteCreationAttributes
> = sequelize.define(
  ...
);
```

It's just added syntactic sugar, that simplifies things.
Also there is no reference to `ModelCtor` nor a functional way with Typescript in the docs.

So another change is, I've added an example for defining models in a functional way in Typescript.

### DIFF
--- a/docs/manual/other-topics/typescript.md
+++ b/docs/manual/other-topics/typescript.md
@@ -15,7 +15,9 @@ In order to avoid installation bloat for non TS users, you must install the foll
 
 Example of a minimal TypeScript project with strict type-checking for attributes.
 
-**NOTE:** Keep the following code in sync with `typescriptDocs/ModelInit.ts` to ensure it typechecks correctly.
+<!-- 
+**NOTE:** Keep the following code in sync with `/types/test/typescriptDocs/ModelInit.ts` to ensure it typechecks correctly.
+-->
 
 ```ts
 import {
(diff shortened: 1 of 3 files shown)


sushantdhiman:
> **NOTE:** Keep the following code in sync with typescriptDocs/ModelInit.ts to ensure it typechecks correctly.

Please take care of that

Keimeno:
Is synchronized now. Also commented out the Notice about synchronization

docs/manual/other-topics/typescript.md
```diff
@@ -15,12 +15,15 @@ In order to avoid installation bloat for non TS users, you must install the foll
 
 Example of a minimal TypeScript project with strict type-checking for attributes.
 
-**NOTE:** Keep the following code in sync with `typescriptDocs/ModelInit.ts` to ensure it typechecks correctly.
+<!--
```
sushantdhiman, line 18: Comment block [breaks esdoc](https://github.com/esdoc/esdoc/issues/388), so please remove comments markup

Keimeno: Removed them now.

PR #12447, opened 2020-06-30
fix(upsert): sets isNewRecord to false when upserting

knoid:
### Pull Request check-list

### Description of change
I'm having problems on upsert returned instances. Doing a `.save()` on an instance right after an upsert returns validation error because it tries to insert instead of updating. I think I tracked this error to having `isNewRecord` as `true` on that returned instance. I tried fixing it by setting it to false on `upsert` method but failed on other tests.

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -2586,6 +2586,8 @@ class Model {
       options
     );
 
+    instance.isNewRecord = false;
+
     if (options.hooks) {
       await this.runHooks('afterUpsert', result, options);
       return result;
--- a/src/model.js
+++ b/src/model.js
@@ -2586,7 +2586,8 @@ class Model {
       options
     );
 
-    instance.isNewRecord = false;
+    const [record] = result;
+    record.isNewRecord = false;
 
     if (options.hooks) {
       await this.runHooks('afterUpsert', result, options);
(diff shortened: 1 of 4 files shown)


sushantdhiman:
No travis build on this PR, can you try pushing an empty commit?

knoid:
done! check Travis, it fails on MARIADB and I don't know how to fix that.

knoid:
managed to fix all related errors! now it fails on mysql something totally unrelated. Am I in the right path? I needed to set `isNewRecord` to false manually after the insert because if not, it would not populate default values.

src/model.js
```diff
@@ -2586,6 +2586,8 @@ class Model {
       options
     );
 
+    instance.isNewRecord = false;
```
sushantdhiman, line 2589: Although here `instance === result`, we should change naming here. May be rename `result` to `record` and then set `record.isNewRecord = false`

knoid: Result is actually `[instance, boolean]`. Should I do this?
```suggestion
    const record = [result];
    record.isNewRecord = false;
```

sushantdhiman: Yes, that works

sushantdhiman: Looks like tests are failing with this commit

jingsam:
I encountered same problem. waiting to get this PR to be merged.

sushantdhiman:
Please open a PR for `v6` branch to get a npm release

Issue #12288, opened 2020-05-21
`id` column is autogenerated despite having a primaryKey defined in an association


soryy708:
### What are you doing?

```js
const Sequelize = require('sequelize');
const Model = Sequelize.Model;

const sequelize = new Sequelize({
    dialect: 'sqlite',
    logging: message => console.log(message),
});

const weakEntity = (sequelize, dataTypes) => {
    const model = sequelize.define('weak', {});

    model.associate = (models) => {
        model.belongsTo(models.parent, {foreignKey: {primaryKey: true}});
    };

    return model;
};

const parentEntity = (sequelize, dataTypes) => {
    const model = sequelize.define('parent', {});

    model.associate = (models) => {
        model.hasOne(models.weak);
    };

    return model;
};

const models = {
    weak: weakEntity,
    parent: parentEntity,
};

(async () => {
    Object.keys(models).forEach(modelName => {
        const modelDef = models[modelName];
        const model = modelDef(sequelize, Sequelize);
        models[modelName] = model;
    });
    Object.keys(models).forEach(modelName => {
        const model = models[modelName];
        if (model.associate) {
            model.associate(models);
        }
    });

    await sequelize.authenticate();
    await sequelize.sync();
})();
```

### What do you expect to happen?

`weak` table to have only 1 column: `parentId`

### What is actually happening?

`weak` table has 2 columns: `parentId` and `id`

### Environment

- Sequelize version: 5.21.8
- Node.js version: 12.16.2
- Operating System: Windows 10 Home x64

### How does this problem relate to dialects?

- [X] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?

- [X] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

sushantdhiman:
Yes, because `model` was created before relationship was invoked. Thus it created a default `id` primary key.

You can either predefine your `primaryKey` attribute in model, so `id` is not created. Or call `removeAttribute('id')`

Issue #12242, opened 2020-05-10
SQLite dialect: composite key with autogenerated `id` column results in syntax error


soryy708:
## Issue Description

This is related to https://github.com/sequelize/sequelize/issues/12194

### What are you doing?

```js
const Sequelize = require('sequelize');

const weakEntity = (sequelize, dataTypes) => {
    const model = sequelize.define('weak', {});

    model.associate = (models) => {
        model.belongsTo(models.parent, {foreignKey: {primaryKey: true}});
    };

    return model;
};

const parentEntity = (sequelize, dataTypes) => {
    const model = sequelize.define('parent', {
        name: {
            type: dataTypes.STRING,
            primaryKey: true,
        },
    });

    model.associate = (models) => {
        model.hasOne(models.weak);
    };

    return model;
};

const models = {
    weak: weakEntity,
    parent: parentEntity,
};

const sequelize = new Sequelize({ dialect: 'sqlite', logging: message => console.log(message) });
Object.keys(models).forEach(modelName => {
    const modelDef = models[modelName];
    const model = modelDef(sequelize, Sequelize);
    models[modelName] = model;
});
Object.keys(models).forEach(modelName => {
    const model = models[modelName];
    if (model.associate) {
        model.associate(models);
    }
});
sequelize.authenticate()
.then(() => sequelize.sync())
.then(() => {
    console.log('Done initializing database');
})
.catch(err => {
    console.log(err);
});
```

### What do you expect to happen?

Created table called "parent" with column called "name"
Created table called "weak" with column called "parentName" that is primary key

### What is actually happening?

Error when attempting to create table called "weak" which has an autoincremented "id" column, in addition to the "parentName" expected column.

```
CREATE TABLE IF NOT EXISTS `weak` (`id` INTEGER NOT NULL AUTOINCREMENT, `createdAt` DATETIME NOT NULL, `updatedAt` DATETIME NOT NULL, `parentName` VARCHAR(255) NOT NULL REFERENCES `parent` (`name`) ON DELETE SET NULL ON UPDATE CASCADE, PRIMARY KEY (`id`, `parentName`));
```

Note the `AUTOINCREMENT` for the `id` column. According to [sqlite documentation](https://www.sqlite.org/autoinc.html):

[...]

Issue #12096, opened 2020-04-10
Non-primary key column "id" would always override the value of the actual primary key column


briantan1979:
## Issue Description
If a non-primary key column is named "id" in the model, the value of this field would always override the value of the actual primary key in the data returned from Model.create() function

### What are you doing?

```js
const TestTable = sequelize.define("TestTable",{
    actualId: {
        type: Sequelize.DataTypes.BIGINT,
        autoIncrement: true,
        primaryKey: true
    },
    id: {
        type: Sequelize.DataTypes.UUID,
        unique: true,
        defaultValue: Sequelize.DataTypes.UUIDV4
    }
});

TestTable.create({},{})
            .then(data => {
                // data.actualId is the same as data.id which is wrong!!!
                console.log(data);
            })
            .catch(err => {
                console.log(err);
            })
```

### What do you expect to happen?

The field actualId should return the auto-incrementing integer value.

### What is actually happening?

The value of actualId is actually overwritten by the value from the id column.

### Environment

- Sequelize version: 5.X
- Node.js version: 10.X
- Operating System: All
- Dialect: mssql

### How does this problem relate to dialects?
- [x] I think this problem happens only for the following dialect(s): mssql

### Would you be willing to resolve this issue by submitting a Pull Request?
- [x] Yes, I have the time but I don't know how to start, I would need guidance.

Found the cause of the bug here:
```js
\\ inside lib\dialects\mssql\query.js
...
class Query extends AbstractQuery {
  getInsertIdField() {
    // The id column name is hard-coded as 'id' here
    //return 'id';
    // Suggest to refactor to the following line to retrieve the primary key column name from the model
    return Object.keys(this.model.primaryKeys)[0];
  }
...

```

Issue #12092, opened 2020-04-09
Error in creating new record with association when an "id" column is not primary key


briantan1979:
## Issue Description
When an "id" column is defined in the model but is not set as the primary key, when creating a new record with association, the record insertion ends up in error. This error only occurs in mssql.

### What are you doing?
A simplified reproduction documented below:

```javascript
//'./schemas/MainTable.js'

var modelName = 'MainTable';
var tableName = 'main_table';

function model (sequelize, DataTypes)
{
    const model = sequelize.define(modelName,{
        counter: {
            type: DataTypes.BIGINT,
            autoIncrement: true,
            primaryKey: true
        },
        id: {
            type: DataTypes.UUID,
            unique: true,
            defaultValue: DataTypes.UUIDV4
        },
        documentNumber: {
            type: DataTypes.STRING,
            unique: true,
            allowNull: true
        }
    },
    {
        paranoid: true,
        underscored: false,
        tableName: tableName
    });
    return model;
}

module.exports = model;
```

```javascript
//'./schemas/AssociatedTable.js'

var modelName = 'AssociatedTable';
var tableName = 'associated_table';

function model (sequelize, DataTypes)
{
    const model = sequelize.define(modelName,{
        someText: {
            type: DataTypes.STRING,
            allowNull: true
        }
    },
    {
        paranoid: true,
        underscored: false,
        tableName: tableName
    });
    return model;
}

module.exports = model;
```
```javascript
// db.js
const Sequelize = require('sequelize');
const sequelize = new Sequelize(
    'dbname',
    'dbuser',
    'dbpassword',
    {
        host: 'localhost',
        port: 1433,
        dialect: 'mssql',
        define: {
            underscored: true
        }
    }
);

const db = {};

db.Sequelize = Sequelize;
db.sequelize = sequelize;

db.MainTable = require('./schemas/MainTable')(sequelize, Sequelize.DataTypes);
db.AssociatedTable = require('./schemas/AssociatedTable')(sequelize, Sequelize.DataTypes);

db.MainTable.hasOne(db.AssociatedTable, {as: 'associatedTable'});

```
```javascript
// test.js
var db = require('./db);
var UUID4 = require('uuid/v4');
var newRecord = {
    id: UUID4()
    associatedTable: {
       someText: 'Testing'
    }
};

db.sequelize.transaction({isolationLevel: db.Sequelize.Transaction.ISOLATION_LEVELS.REPEATABLE_READ})
     .then(function (tran) {
[...]

PR #12458, opened 2020-07-03
feat(model): Adding Option to Toggle Automatic Id Column Creation

### LINKED ISSUE #11830: Stop providing a default `id` attribute for models
The default `id` that Sequelize provides for models is often source of confusion.

```js
const Foo = sequelize.define('foo', {
  name: DataTypes.STRING
});
```

Is actually equivalent to

```js
const Foo = sequelize.define('foo', {
  id: {
    type: DataTypes.INTEGER,
    allowNull: false,
    primaryKey: true,
    autoIncrement: true
  },
  name: DataTypes.STRING
});
```

It has the obvious advantage of less boilerplate code. However, it is often source of confusion and weird errors, such as in issue #11723 (as clarified in [this comment](https://github.com/sequelize/sequelize/issues/11723#issuecomment-560881383)) and #11714

I suggest a breaking change of no longer providing this default attribute. Let's make users define them. This way this kind of confusion will never happen.

Keimeno:
### Pull Request check-list

### Description of change

Resolves #11830 

This adds the option to disable the automatic creation of an id column which is the primary key, if no primary key was set in the model initialization. 

Related Issues: #7800 #12096 #12092 #12242 #12288

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -228,6 +228,12 @@ class Model {
         throw new Error(`A column called 'id' was added to the attributes of '${this.tableName}' but not marked with 'primaryKey: true'`);
       }
 
+      if (!this.options.generatePrimaryKey) {
+        // don't do anything further or it will
+        // implement an id column that we don't want
+        return;
+      }
+
       head = {
         id: {
           type: new DataTypes.INTEGER(),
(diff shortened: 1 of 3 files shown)


sushantdhiman:
Thanks for this PR, for now I wont accept it. Sequelize assume `id` field / attribute with defined model, as long as that is not changed internally, I don't think it will be safe to provide this feature.

lib/model.js
```diff
@@ -228,6 +228,12 @@ class Model {
         throw new Error(`A column called 'id' was added to the attributes of '${this.tableName}' but not marked with 'primaryKey: true'`);
       }
 
+      if (!this.options.generatePrimaryKey) {
+        // don't do anything further or it will
+        // implement an id column that we don't want
+        return;
```
sushantdhiman, line 234: This also means timestamp, version and paranoid attribute will be disabled when you turn on this flag

PR #12484, opened 2020-07-10
fix: mark database drivers as optional peer dependencies

DRoet:
### Pull Request check-list

### Description of change

When using yarn 2 and PnP, it will throw an error because sequelize is trying to access the database drivers without listing them as direct dependencies/peerDependencies in package.json

```sh
(node:13140) [MODULE_NOT_FOUND] Error: sequelize tried to access mysql2, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

Since we don't want to list them as direct dependencies, we can fix this error by listing all the database drivers as optional peer dependencies using the new `peerDependenciesMeta` field.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,26 @@
     "tedious": "8.3.0",
     "typescript": "^3.9.3"
   },
+  "peerDependenciesMeta": {
+    "pg": {
+      "optional": true
+    },
+    "pg-hstore": {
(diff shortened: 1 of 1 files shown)


DRoet:
@sushantdhiman see https://github.com/sequelize/sequelize/pull/12494

PR #12497, opened 2020-07-11
feat(ts): add bundled types for toposort-class

eseliger:
# Pull Request check-list

### Description of change

As this dependency has no types included, I bundled them here, so we can use them when migrating to TS.

### DIFF
--- a/src/custom-types/toposort-class/index.d.ts
+++ b/src/custom-types/toposort-class/index.d.ts
@@ -1,0 +1,9 @@
+declare module 'toposort-class' {
+  class Toposort {
+    add(item: string, deps: string | string[]): Toposort;
+    sort(): string[];
+    clear(): Toposort;
+    Toposort: typeof Toposort;
+  }
+  export = Toposort;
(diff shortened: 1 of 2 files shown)


src/custom-types/toposort-class/index.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'toposort-class' {
+  class Toposort {
+    add(item: string, deps: string | string[]): Toposort;
+    sort(): string[];
+    clear(): Toposort;
+    Toposort: typeof Toposort;
+  }
+  export = Toposort;
```
eseliger, line 8:
```suggestion
  export class Toposort {
    add(item: string, deps: string | string[]): Toposort;
    sort(): string[];
    clear(): Toposort;
    Toposort: typeof Toposort;
  }
```

eseliger: @SimonSchick as discussed on a call, PTAL

PR #12498, opened 2020-07-11
Migrate ModelManager to TypeScript

eseliger:
### Pull Request check-list


src/model-manager.ts
```diff
@@ -1,33 +1,29 @@
...
-  constructor(sequelize) {
-    this.models = [];
-    this.sequelize = sequelize;
-  }
+export class ModelManager {
+  private models: typeof Model[] = [];
+  constructor(private sequelize: any) {}
 
-  addModel(model) {
+  public addModel(model: typeof Model): typeof Model {
```
SimonSchick, line 10: Should probably be generic.

eseliger: We don't have generic models yet so I figured we'd revisit each use of Model anyways. Do you want me to change it to: ?

```suggestion
  public addModel<T extends typeof Model>(model: T): T {
```

SimonSchick: Basically yes.

PR #12500, opened 2020-07-12
feat(ts): convert sql-string to ts

eseliger:
### Pull Request check-list

### Description of change

Needed to make the logger a module as well, so I can import it.

### DIFF
--- a/src/sql-string.ts
+++ b/src/sql-string.ts
@@ -26,7 +26,7 @@ export function escape(val: unknown, timeZone: string | null, dialect: string, f
       // for us. Postgres actually has a boolean type with true/false literals,
       // but sequelize doesn't use it yet.
       if (dialect === 'sqlite' || dialect === 'mssql') {
-        return String(+!!val);
+        return val ? '1' : '0';
       }
       return (!!val).toString();
     case 'number':
@@ -49,7 +49,7 @@ export function escape(val: unknown, timeZone: string | null, dialect: string, f
       return (dataTypes as any)[dialect].BLOB.prototype.stringify(val);
     }
 
-    return dataTypes.BLOB.prototype.stringify(val, undefined);
+    return dataTypes.BLOB.prototype.stringify(val, undefined); // TODO: Remove unknown
   }
 
   if (Array.isArray(val)) {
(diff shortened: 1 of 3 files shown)


src/sql-string.ts
```diff
@@ -27,7 +26,7 @@ function escape(val, timeZone, dialect, format) {
       // for us. Postgres actually has a boolean type with true/false literals,
       // but sequelize doesn't use it yet.
       if (dialect === 'sqlite' || dialect === 'mssql') {
-        return +!!val;
+        return String(+!!val);
```
SimonSchick: I think this could be simplified to `return val ? '1' : '0'`

src/sql-string.ts
```diff
@@ -40,21 +39,21 @@ function escape(val, timeZone, dialect, format) {
...
 
   if (Buffer.isBuffer(val)) {
-    if (dataTypes[dialect].BLOB) {
-      return dataTypes[dialect].BLOB.prototype.stringify(val);
+    if ((dataTypes as any)[dialect].BLOB) {
+      return (dataTypes as any)[dialect].BLOB.prototype.stringify(val);
     }
 
-    return dataTypes.BLOB.prototype.stringify(val);
+    return dataTypes.BLOB.prototype.stringify(val, undefined);
```
SimonSchick: We shouldn't have to pass this.

src/sql-string.ts
```diff
@@ -99,10 +99,9 @@ function escape(val, timeZone, dialect, format) {
   }
   return `${(prependN ? "N'" : "'") + val}'`;
 }
-exports.escape = escape;
 
-function format(sql, values, timeZone, dialect) {
-  values = [].concat(values);
+export function format(sql: string, values: unknown[], dialect: string): string {
+  values = ([] as unknown[]).concat(values);
```
SimonSchick: Should probasbly just use `.slice()` for cloning.

src/sql-string.ts
```diff
@@ -113,21 +112,19 @@ function format(sql, values, timeZone, dialect) {
...
+    return escape(values.shift(), null, dialect, true);
   });
 }
-exports.format = format;
 
-function formatNamedParameters(sql, values, timeZone, dialect) {
+export function formatNamedParameters(sql: string, values: Record<string, string>, dialect: string): string {
   return sql.replace(/:+(?!\d)(\w+)/g, (value, key) => {
-    if ('postgres' === dialect && '::' === value.slice(0, 2)) {
+    if (dialect === 'postgres' && value.slice(0, 2) === '::') {
```
SimonSchick: Use `startsWith`

PR #12526, opened 2020-07-18
fix(model): clone options object instead of modifying

### LINKED ISSUE #12406: model.bulkCreate modifies the "options" by assigning "model" as an attribute to the object
## Issue Description
While using model.bulkCreate, as we pass the options, an object which contains transaction, it comes modified after the execution

### What are you doing?

We have called bulkCreate on a model object. The options object which is prepared as an argument to the function, if we check it after execution, it contains the reference to the model object.

**Here is the link to the SSCCE for this issue:** LINK-HERE 

```js
// MINIMAL, SELF-CONTAINED code here (SSCCE/MCVE/reprex)
```

_The options object should not have been modified._

_The options object now have a reference to "model" object._

```
Output here
```

### Additional context
When we checked the code in sequelize -> model.js, it reflected the assignment on the passed in argument, instead of using the cloned copy of "options"

image

### Environment

- Sequelize version: 5.21.10
- Node.js version: 12.16.3
- Operating System: MacOS Catalina, Ubuntu 18.04
- If TypeScript related: TypeScript version: 4.28.9

auipc:
### Pull Request check-list

### Description of change

closes #12406

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -2629,6 +2629,8 @@ class Model {
     const dialect = this.sequelize.options.dialect;
     const now = Utils.now(this.sequelize.options.dialect);
 
+    options = Utils.cloneDeep(options);
+
     options.model = this;
 
     if (!options.includeValidated) {


sergiopereira:
Anything blocking this from getting an approval and being merged?

PR #12568, opened 2020-07-29
fix(describe-table-query): add condition to prevent subquery error

nachocab:
### Pull Request check-list

### Description of change

If there is more than one schema with the same table names (e. g.: `public` and `backup`), the `pgd.description `subquery in `describeTableQuery` will return more than one row and will result in the error `SequelizeDatabaseError: more than one row returned by a subquery used as an expression`. 

The added condition prevents this from happening.

### DIFF
--- a/src/dialects/postgres/query-generator.js
+++ b/src/dialects/postgres/query-generator.js
@@ -133,7 +133,7 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
       'c.is_nullable as "Null", ' +
       "(CASE WHEN c.udt_name = 'hstore' THEN c.udt_name ELSE c.data_type END) || (CASE WHEN c.character_maximum_length IS NOT NULL THEN '(' || c.character_maximum_length || ')' ELSE '' END) as \"Type\", " +
       '(SELECT array_agg(e.enumlabel) FROM pg_catalog.pg_type t JOIN pg_catalog.pg_enum e ON t.oid=e.enumtypid WHERE t.typname=c.udt_name) AS "special", ' +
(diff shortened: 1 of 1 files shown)


nachocab:
Any updates on this. It seems like a pretty simple merge

dmercertaylor:
It looks like one of the key maintainers on this project just left, so it might take the remaining team a minute to get to these. It looks like the PRs have been piling up for the last little bit.

nachocab:
It would be great if someone could review this 😅

jiqiang90:
Hi guys we also facing the same issue, any update on this?

WikiRik:
Feel free to open a new PR to include this, we would need additional tests. At least one for the normal behaviour that already works on `main` and one that doesn't work on main (so multiple schemas with the same table name) and is fixed with this change.
What you probably would want to test is if the `describeTable` function of the query interface is correct. We have integration tests on that [here](https://github.com/sequelize/sequelize/blob/main/test/integration/query-interface/describeTable.test.js) but seeing as this change is only for postgres it might be worth putting them [here](https://github.com/sequelize/sequelize/blob/main/test/integration/dialects/postgres/query-interface.test.js).
Although it might be nicer if you could make tests that are done on all dialects that support multiple schemas.

@ephys you're working on schemas as part of #14999 so maybe you can give additional pointers or take it up yourself

dorlanpabon (commented):
This change is appropriate and necessary to improve the accuracy and reliability of queries generated by Sequelize.

Prevents subquery errors: By adding the condition AND c.table_schema = st.schemaname, this ensures that the subquery returns results only from the correct schema, preventing the "more than one row returned by a subquery used as an expression" error, which occurs when there are tables with the same name in different schemas.

Clarity and accuracy: This change makes the query more precise by ensuring that the information is fetched from the correct schema, which is crucial in databases with multiple schemas or complex configurations.

Avoids unexpected behavior: Without this condition, the query could produce unpredictable results in environments with multiple schemas. Adding this check ensures more robust and reliable behavior

PR #12573, opened 2020-07-30
fix(typings): allow Col in ProjectionAlias

### LINKED ISSUE #12572: Typescript does not allow col() as part of ProjectionAlias
(Opening a Docs issue assuming typescript typings count as docs)

## Issue Description

When writing a sequelize query using the col() function in the `attributes` array, Typescript throws an error `Type 'Col' is not assignable to type 'string | Fn | Literal'.`

This is caused by the typing here: https://github.com/sequelize/sequelize/blob/v6/types/lib/model.d.ts#L482
> `export type ProjectionAlias = [string | Literal | Fn, string];`

As a bare minimum example, the goal is to write a query like so:

```
import { col, Model } from 'sequelize';

class MyModel extends Model {}
class MyIncludedModel extends Model {}

MyModel.findAll({
    attributes: [
        'id',
        [col('MyIncludedModel.id'), 'includedId'],
    ],
    include: [{
        model: MyIncludedModel,
        attributes: [],
    }],
    raw: true,
});
```

(obviously some more setup is required to _actually_ work, but this is enough to trigger the Typescript error).

Using the `col()` method in attributes is totally supported in code.

The definition for ProjectionAlias is missing Col
e.g. export type ProjectionAlias = [string | Literal | Fn` | Col`, string];

Add Col to ProjectionAlias

### Additional context

I hit this issue right away, but it feels like it is explicitly left out. If there is a good reason not to use `Col` and to leave it out, I'd appreciate hearing the justification.

I'll open a PR updating typings, but the reasoning for the issue is to actually discuss why _not_ to include it (if such reason exists, and is not just oversight)

- [ ] I don't know.

JacobLey:
Closes #12572 

 # Pull Request check-list

### Description of change

Allow Col to be used in ProjectionAlias. See issue https://github.com/sequelize/sequelize/issues/12572 for more discussion


JacobLey:
Looks like https://github.com/sequelize/sequelize/pull/12573 was chosen instead

papb:
Hi @JacobLey, indeed, sorry about that! Thank you for opening that issue and this PR. I merged the other one before even seeing yours, and for that I apologize.

Issue #12633, opened 2020-08-20
Model.findAndCountAll() uses two database connections instead of one


marcogrcr:
## Issue Description

Using `Model.findAndCountAll()` uses two database connections, when I expect only one to be used.

### What are you doing?

```js
'use strict';

const { Sequelize, Model, DataTypes } = require('sequelize');

async function test() {
    const sequelize = new Sequelize({
        dialect: 'mysql',
        // (...database connection settings...),
        pool: {
          max: 2
        }
    });

    const Foo = sequelize.define('Foo', { name: DataTypes.TEXT });
    await sequelize.sync();

    for (let i = 0; i < 100; ++i) {
      await Foo.create({ name: 'foo' });
    }

    await Foo.findAndCountAll();
};

test().then(() => console.log('done')).catch(e => console.error(e));
```

### What do you expect to happen?

Only one connection to the database is opened and used when `Foo.findAndCountAll()` is invoked.

### What is actually happening?

Two connections to the database are opened and used when `Foo.findAndCountAll()` is invoked.

### Additional context

This is caused by `Foo.findAndCountAll()`'s usage of `Promise.all()` when executing:

**v6.3.4**

https://github.com/sequelize/sequelize/blob/490db410e33241158f7d8a7cdeffbf3766af731c/lib/model.js#L2085-L2088

**v5.22.3**

https://github.com/sequelize/sequelize/blob/3d2df28e9fcfc6a640dbad21b0435ed8804602d4/lib/model.js#L2100-L2107

This is especially problematic when running `sequelize` on AWS Lambda where the assumption that AWS Lambda containers only processes one request at a time has lead to multiple issues and articles recommending developers to use `{ pool: { max: 1 } }`.

This `{ max: 1 }` configuration can be problematic with `Model.findAndCountAll()` when `this.count()` takes longer than `{ pool: { acquire } }`. The `this.findAll()` query that is waiting for the connection used by `this.count()` will time out, resulting in a `SequelizeConnectionAcquireTimeoutError: Operation timeout`.

Now, this may be considered "by design", but I would consider adding an `options` field to let developers choose whether they want to issue both queries concurrently (resulting in two connections used), or whether they want to serialize them (resulting in one connection used).

Otherwise, I would consider documenting this behavior so developers know what to expect. Especially when configuring pools with `{ max: 1 }`. I would document this in:

- https://sequelize.org/master/manual/model-querying-finders.html#-code-findandcountall--code-
[...]

Issue #12364, opened 2020-06-11
Intermittent connection hanging with no errors using .authenticate()


nickr-davey:
## Intermittent connection hanging with no errors using .authenticate()

I am using v5 with lambda functions and RDS postgres 10.6, node 12

Using both v4 and v5 and pg 7+ and 8+ I get an issue where sometimes the connection is never established I think. It seems to just hang in the authenticate() function where it is not returning an error or resolving. 
I have gone through the other issues where people were running into this issue on v4 so I updated to v5 but I still see the same problem. When it happens once it wont get out of this state until and new instance of the function is created. It is really hard to debug when there are no errors happening and my function will just timeout.

It makes no difference using context.callbackWaitsForEmptyEventLoop = false; or increasing or decreasing the idle time.

### How I am setting up my connection
```javascript
  let sequelize: Sequelize = null;
export const db = () => {
  if (sequelize == null) {
    console.log("new db connection");
    return createInstanceDB();
  } else {
    console.log("old db connection");
    return sequelize
      .authenticate()
      .then(() => sequelize)
      .catch((err) => {
        console.log("SEQ_ERROR", err);
        return err
      });
  }
};

export const createInstanceDB = async () => {
  const { database, username, password, host } = JSON.parse(process.env.DB);

  // Open database connection
  sequelize = new Sequelize(database, username, password, {
    host: host,
    dialect: "postgres",
    pool: { max: 1, min: 0, idle: 2000},
    logging: process.env.DB_DEBUG === "true" ? console.log : false,
  });

  // Initialize each model in the database
 
// setting up associations here

  
  try {
    await sequelize.authenticate();
    console.log("Connection has been established successfully.");
    return sequelize;
  } catch (error) {
    console.error("Unable to connect to the database:", error);
    throw error;
  }
};
```

### Environment
Node 12.x
PG 8.21.x
Sequelize 5.21.10
AWS Lambda
AWS RDS postgres

### Is this issue dialect-specific?

- [ ] No. This issue is relevant to Sequelize as a whole.
- [ ] Yes. This issue only applies to the following dialect(s): XXX, YYY, ZZZ
- [X] I don't know.

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [X] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
[...]

Issue #12363, opened 2020-06-10
intermittent connection hanging with no errors using .authenticate()


nickr-davey:
## Intermittent connection hanging with no errors using .authenticate()

I am using lambda functions and RDS postgres 10.6

Using both v4 and v5 and pg 7+ and 8+ I get an issue where sometimes the connection is never established I think. It seeems to just hang in the authenticate() function where it is not returning an error or resolving. 
I have gone through the other issues where people were running into this issue on v4 so I updated to v5 but I still see the same problem. When it happens once it wont get out of this state until and new instance of the function is created. It is really hard to debug when there are no errors happening and my function will just timeout.

It makes no difference using context.callbackWaitsForEmptyEventLoop = false; or increasing or decreasing the idle time.

#### I never see - executing (default): SELECT 1+1 AS result. when this happens so I am not sure what is happening is it isn't trying to see if it is connected.

`
  sequelize = new Sequelize(database, username, password, {
    host: host,
    dialect: "postgres",
    pool: { max: 1, min: 0, idle: 1000 },
    logging: console.log,
  
  });

.........

  try {
    await sequelize.authenticate();
      console.log("Connection established");
    }
    return sequelize;
  } catch (error) {
    console.error("--------------  database error:", error);
    throw error;
  }
`

### Is this issue dialect-specific?
- [x] I don't know.

### Would you be willing to resolve this issue by submitting a Pull Request?

- [x] Yes, I have the time but I don't know how to start, I would need guidance.

sushantdhiman:
Duplicate of https://github.com/sequelize/sequelize/issues/12158

Follow the template next time

nickr-davey:
Not a duplicate as we are using node 12.x also same thing happens with node 10.x , 
and filled out the template as well.

sushantdhiman:
@nickr-davey Then reopen issue by properly following template, you missed versions, dialect and reproduction code etc. Spend some time to properly present your use case, see examples from other open issues

marcogrcr:
For people experiencing issues in AWS Lambda, I have created a pull request that documents how to properly configure `sequelize` for Lambda. Hopefully you'll be able to see it in <https://sequelize.org/master/manual/aws-lambda.html> once it gets merged. In the meantime, please check #12642 out and provide feedback if you find any inaccuracies or space for improvement.

Issue #12245, opened 2020-05-11
[Question] Have anyone tried using AWS Lambda Proxy with RDS


muriloSanto:
Have anyone tried using AWS Lambda Proxy with RDS to reuse connections ?

marcogrcr:
For people experiencing issues in AWS Lambda, I have created a pull request that documents how to properly configure `sequelize` for Lambda. Hopefully you'll be able to see it in <https://sequelize.org/master/manual/aws-lambda.html> once it gets merged. In the meantime, please check #12642 out and provide feedback if you find any inaccuracies or space for improvement.

PR #12642, opened 2020-08-25
Create aws-lambda.md

marcogrcr:
This document aims to guide developers on how to use `sequelize` in AWS Lambda. Properly configuring `sequelize` for AWS Lambda is a common source of pain for developers.


marcogrcr:
Here are some of the issues that I found related to AWS Lambda and connection pools: #3895, #4938, #5441, #6026, #8087, #8468, #9242, #10439, #11021, #11394, #12245, #12363, #12364, #12633

marcogrcr:
I have updated the guide with a simpler solution and more in-depth details for people that may be interested in diving deeper.

marcogrcr:
I have updated the guide fixing some typos and other minor clarifications ([diff](https://github.com/sequelize/sequelize/compare/cdd340da3434430794951251e13e83773c7a385b..925fd74cf45b2bd2189e6da2de8baec98d8cba93)).

catamphetamine:
Man, you really dug deep into this...

Respect

amitaymolko:
Thanks for this document it's very useful.  
I didn't understand what connection pool settings you recommend in the end?  
Also if you are using RDS proxy are any of the pool settings relevant?

marcogrcr:
Hi @amitaymolko:

> I didn't understand what connection pool settings you recommend in the end?
> 
> Also if you are using RDS proxy are any of the pool settings relevant?

Since connection pooling is not used across invocations, the settings are not as important as before. For most cases the default settings should be OK:

```js
{
  "max": 5,
  "min": 0,
  "idle": 10000,
  "acquire": 60000,
  "evict": 1000,
}
```

However you can optimize on a few things:

- `max`: as far as I know, all `sequelize` operations issue at most `2` concurrent queries. So you should probably not want to use a value lower than 2 if you don't want your queries to be serialized. If you explicitly issue queries concurrently, same principle applies. That being said be careful about not exceeding your underlying server's capacity.
- `idle`: this value is still important because your Lambda may timeout before you close the connections on the pool. So you want those connections to be closed when the function is invoked again. Use `1ms` when you configure `evict` as recommended.
- `acquire`: since Lambda containers handle a single request at a time, you can use a more aggressive value here. If you have a `max` value for all the Lambda's current queries, then you can put a value that represents the maximum time you're willing to wait for a connection to be established (e.g. `3000`).
- `evict`: this basically is the argument for a `setTimeout()` used by the pool to clean idle connections. Since you close them manually at the end of the request but your Lambda function may timeout, the ideal setting is to use the same value as your Lambda function's timeout.

I'll probably update the pull request with these recommendations.

johnwquarles:
@marcogrcr Really appreciate the time & thought put into this writeup!

I ran into the issue today when setting up a serverless framework API, and noted that the `TL;DR` workaround seemed to produce errors on warm lambda invocations (cold starts were fine):

```
Error: ConnectionManager.getConnection was called after the connection manager was closed!
```

I believe it's because calling `connectionManager.close()` rewrites `connectionManager.getConnection` to throw the error I was seeing on those warm invocations:

https://github.com/sequelize/sequelize/blob/042cd693635ffba83ff7a2079974692af6f710a7/src/dialects/abstract/connection-manager.js#L110

I got around this by writing a function to do the same things, but without rewriting `getConnection`,

```javascript
  const closeConnection = async () => {
    await sequelize.connectionManager.pool.drain()
    return sequelize.connectionManager.pool.destroyAllNow()
  }
```

and calling it at the end of each handler in lieu of `sequelize.connectionManager.close()`

My understanding is that this setup closes & then reinstantiates the `pool` on each subsequent lambda invocation, which is all that we need-- and that reusing the `connectionManager` itself poses no problem given that setup.

amitaymolko:
@johnwquarles I can confirm your code works for me, thanks!  
Sequelize: 5.21.2

marcogrcr:
@johnwquarles you're right about that. As a disclaimer, I have not had the opportunity run the **TL;DR** code on my project yet. When I analyzed that code, I got confused for a moment and thought the `sequelize.connectionManager.pool.acquire()` method was the one being overwritten instead of `sequelize.connectionManager.getConnection()`. My initial thought was that by using `sequelize.connectionManager.initPools()` the method overwrite would be reverted since a new instance of the pool would be created and assigned to `sequelize.connectionManager.pool`.

The code that you're proposing should do the trick since it's the equivalent of `sequelize.connectionManager.close()` except for the method override:

**v6.3.5**

https://github.com/sequelize/sequelize/blob/933b3f62640e218587d34ec141029b2416ff7845/lib/dialects/abstract/connection-manager.js#L110-L117

https://github.com/sequelize/sequelize/blob/933b3f62640e218587d34ec141029b2416ff7845/lib/dialects/abstract/connection-manager.js#L94-L103

**v5.22.3**

https://github.com/sequelize/sequelize/blob/3d2df28e9fcfc6a640dbad21b0435ed8804602d4/lib/dialects/abstract/connection-manager.js#L117-L124

https://github.com/sequelize/sequelize/blob/3d2df28e9fcfc6a640dbad21b0435ed8804602d4/lib/dialects/abstract/connection-manager.js#L101-L110

---

The only downside I see with this code is that `sequelize.connectionManager.pool` is not part of the public API:

**v6.3.5**

https://github.com/sequelize/sequelize/blob/933b3f62640e218587d34ec141029b2416ff7845/types/lib/connection-manager.d.ts#L14-L34

**v5.22.3**

https://github.com/sequelize/sequelize/blob/3d2df28e9fcfc6a640dbad21b0435ed8804602d4/types/lib/connection-manager.d.ts#L16-L36

This means that:

- TypeScript users will have to break type-safety.
- The code may break on newer versions of `sequelize`.

That being said, that's the best alternative I see that does not involve the overhead of creating a new instance of `Sequelize` on every invocation.

---

I guess the ideal solution would be to have `sequelize.connectionManager.initPools()` restores the overwritten method so there's no need to access class internals and avoid creating a new instance of `Sequelize` on every invocation. I guess it's a matter of proposing it to the `sequelize` team in a pull request to see if they accept it.

marcogrcr:
@amitaymolko / @johnwquarles : Thanks to your feedback, I have created a new revision (see diff [here](https://github.com/sequelize/sequelize/compare/925fd74cf45b2bd2189e6da2de8baec98d8cba93..ee3f724753f92d0b17ee2d38cb985196a06bc820)).

papb:
Hello everyone!! Thank you so much @marcogrcr for the detailed explanation! I will not merge yet because I want to take a closer look first, and I haven't had much time, please wait a little more and thank you very much!! :rocket:

RobertFischer (approved):

cathykc:
This is so great @marcogr! We've been struggling with the proper sequelize <> serverless config. Would love to see this merged in and be part of the official documentation. Anything I can do to help?

marcogrcr:
Thank you @cathykc! I've rebased the commit so that the CI build passes.

I don't think there's more than I can do other than wait for @papb to merge it.

docs/manual/other-topics/aws-lambda.md
```diff
@@ -0,0 +1,730 @@
...
+  // re-use the sequelize instance across invocations to improve performance
+  if (!sequelize) {
+    sequelize = await loadSequelize();
+  } else {
+    // restart connection pool to ensure connections are not re-used across invocations
+    sequelize.connectionManager.initPools();
+
+    // restore `getConnection()` if it has been overwritten by `close()`
+    if (sequelize.connectionManager.hasOwnProperty("getConnection")) {
+      delete sequelize.connectionManager.getConnection;
```
prestonlimlianjie, line 75: Thanks for the awesome and detailed writeup @marcogrcr! 

I was testing out the `TL;DR` example code and the Typescript compiler was giving an `The operand of a 'delete' operator must be optional.` error in response to the following line:
```
delete sequelize.connectionManager.getConnection;
```

I know this is out of the scope of this PR, but maybe a better solution to the problem is to modify the [`getConnection()` function](https://github.com/sequelize/sequelize/blob/9634338d8f9a07ff151f70a91793b37547e2c23e/lib/dialects/abstract/connection-manager.js#L243) to check if `!this.pool` and throw an error. That way, we wouldn't need to get the [`close()` function](https://github.com/sequelize/sequelize/blob/9634338d8f9a07ff151f70a91793b37547e2c23e/lib/dialects/abstract/connection-manager.js#L110) to overwrite `getConnection()`. 

I would gladly make a PR if you deem fit @papb

marcogrcr: Hey @prestonlimlianjie:

Thanks for reviewing and the kind words!

I like your solution better. It's simple and it doesn't require changes to the TypeScript definitions. Happy to update this pull request if #13197 gets merged before this one.

BTW, I have returned you the favor reviewed #13197 with a couple of minor comments.

marcogrcr:
Hello @papb

Just checking if you've had a chance to take a look at this pull request? It seems like various people are taking notice of it and many more would benefit from this being published in https://sequelize.org/.

Would you be able to share an ETA of when do you think someone from the `sequelize` project may be able to review?

docs/manual/other-topics/aws-lambda.md
prestonlimlianjie: Thanks for the PR review @marcogrcr - btw it seems like you're not on the Sequelize Slack org, you might want to join it especially since you're very passionate about Sequelize/they're looking for maintainers on the project! https://github.com/sequelize/sequelize#note-looking-for-maintainers

marcogrcr:
Hey @sdepold:

Can we re-open this pull-request? I rebased the branch on top of `main` as requested.

sdepold:
Haven't tried it but you have my trust :)

agustinparamio:
Hello @marcogrcr 

"Since connection pooling is not used across invocations, the settings are not as important as before. For most cases the default settings should be OK"

Here (https://github.com/sequelize/sequelize/issues/4938#issuecomment-677933370) you also wrote:  "Keep in mind that AWS has released a product for AWS Lambda called [Amazon RDS Proxy](https://aws.amazon.com/rds/proxy/) where you don't have to worry about connection pooling and you can create a new connection for every request."

How do you suggest we build our "lambda.handler"? 

Followed the guide here (https://sequelize.org/docs/v6/other-topics/aws-lambda/) but we continue to experience "SequelizeConnectionAcquireTimeoutError" from time to time.  I even removed the close() and initPool() because seems unnecessary by your comments.

And RDS connections to the DB seems to be stacking to the limit of out current allowance.


Thank you for the time you put doing this.

patricktyndall:
Hi all, thanks for all the work here.

I'm still having an issue with sequelize (^6.19.0) and lambda (nodejs12.x) w/ RDS Aurora Postgres.

We've implemented the code as described in the lambda doc, and are now seeing new error when a lambda function is invoked after "re-connecting" using the same sequelize instance (e.g. after calling the `sequelize.connectionManager.initPools()`).

The symptoms of the error:
- We're getting an error `SequelizeDatabaseError` with message `Error: Connection terminated`.
  - Here's a copy/paste of stack trace:
  - ```{ name: 'SequelizeDatabaseError', parent: Error: Connection terminated at Connection.<anonymous> (/var/task/node_modules/pg/lib/client.js:131:36) at Object.onceWrapper (events.js:420:28) at Connection.emit (events.js:326:22) at Connection.EventEmitter.emit (domain.js:483:12) at Socket.<anonymous> (/var/task/node_modules/pg/lib/connection.js:62:12) at Socket.emit (events.js:314:20) at Socket.EventEmitter.emit (domain.js:483:12) at TCP.<anonymous> (net.js:675:12)```

- This often happens in the middle of a request, after other successful DB calls. It seems that some connections in the pool work just fine, but others do not.
- We don't see this error in requests that have to initialize a brand new sequelize instance
- We also didn't see this error when we were closing and reconnecting the pool on each request (but clearly don't want to do this)

For more context, our entire API is implemented using a single lambda function, so we're reusing containers at a pretty high frequency.

Also, here's our sequelize config. I have played with many configurations (including using the config in the doc), and have not found any solutions, so I currently have it stripped down. Anything here that could be the problem?
```
	sequelize = new Sequelize(
		db_name,
		db_user,
		db_pass,
		{
			host: db_host,
			dialect: 'postgres',
			port: db_port,
			benchmark: true,
			minifyAliases: true,
		},
	);
```

TIA!

PR #12660, opened 2020-08-31
fix(truncate): fix missing await in truncate all model with cascade

bmeriaux:
### Pull Request check-list

### Description of change

Fix a missing `await` in `sequelize.truncate` with `cascade` option. Without it, all models try to truncate in parallel which triggers a deadlock when 2 models truncate the same child model.

### DIFF
--- a/src/sequelize.js
+++ b/src/sequelize.js
@@ -817,17 +817,13 @@ class Sequelize {
    */
   async truncate(options) {
     const models = [];
-    this.modelManager.forEachModel(
-      model => {
-        if (model) {
-          models.push(model);
-        }
-      },
-      { reverse: false }
-    );
+
+    this.modelManager.forEachModel(model => models.push(model), { reverse: false });
 
     if (options && options.cascade) {
-      for (const model of models) await model.truncate(options);
+      for (const model of models) {
+        await model.truncate(options);
+      }
     } else {
       await Promise.all(models.map(model => model.truncate(options)));
     }
--- a/test/integration/sequelize.test.js
+++ b/test/integration/sequelize.test.js
@@ -384,6 +384,47 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
   });
 
   describe('truncate', () => {
+    it('truncates all models sequentially in cascade', async function () {
+      const Project = this.sequelize.define(`project-${config.rand()}`, { title: DataTypes.STRING });
+      const Project2 = this.sequelize.define(`project2-${config.rand()}`, { title: DataTypes.STRING });
+      const TopProject = this.sequelize.define(`top-project-${config.rand()}`, { title: DataTypes.STRING });
+      Project.belongsTo(TopProject);
+      Project2.belongsTo(TopProject);
+      TopProject.hasMany(Project);
+      TopProject.hasMany(Project2);
+      await this.sequelize.sync({ force: true });
+      const project = await Project.create(
+        { title: 'bla', topProject: { title: 'topProject1' } },
+        { include: TopProject }
+      );
+      expect(project).to.exist;
+      expect(project.title).to.equal('bla');
+      expect(project.id).to.equal(1);
+      const project2 = await Project2.create(
+        { title: 'bla2', topProject: { title: 'topProject2' } },
+        { include: TopProject }
+      );
+      expect(project2).to.exist;
+      expect(project2.title).to.equal('bla2');
+      expect(project2.id).to.equal(1);
+      try {
+        await this.sequelize.truncate({ cascade: true });
+
+        expect(await Project.count()).to.equal(0);
+        expect(await Project2.count()).to.equal(0);
+        expect(await TopProject.count()).to.equal(0);
+      } catch (e) {
+        if (this.sequelize.dialect.supports.truncateCascade) {
+          expect.fail(
+            e,
+            undefined,
+            `truncate with cascade for dialect ${this.sequelize.dialect.name} failed with ${e.message}`
+          );
+        } else {
+          expect(e).to.be.instanceOf(Error, 'truncate with cascade is not supported for the current dialect');
+        }
+      }
+    });
     it('truncates all models', async function () {
       const Project = this.sequelize.define(`project${config.rand()}`, {
         id: {
--- a/src/dialects/abstract/index.js
+++ b/src/dialects/abstract/index.js
@@ -63,7 +63,8 @@ AbstractDialect.prototype.supports = {
   groupedLimit: true,
   indexViaAlter: false,
   JSON: false,
-  deferrableConstraints: false
+  deferrableConstraints: false,
+  truncateCascade: false
 };
 
 module.exports = AbstractDialect;
--- a/src/dialects/abstract/query-interface.js
+++ b/src/dialects/abstract/query-interface.js
@@ -965,6 +965,9 @@ class QueryInterface {
     options = _.defaults(options, { limit: null });
 
     if (options.truncate === true) {
+      if (options.cascade && !this.sequelize.dialect.supports.truncateCascade) {
+        throw new Error('truncate with cascade is not supported for the current dialect');
+      }
       return this.sequelize.query(this.queryGenerator.truncateTableQuery(tableName, options), options);
     }
 
--- a/src/dialects/postgres/index.js
+++ b/src/dialects/postgres/index.js
@@ -56,7 +56,8 @@ PostgresDialect.prototype.supports = _.merge(_.cloneDeep(AbstractDialect.prototy
   JSONB: true,
   HSTORE: true,
   deferrableConstraints: true,
-  searchPath: true
+  searchPath: true,
+  truncateCascade: true
 });
 
 PostgresDialect.prototype.defaultVersion = '9.5.0';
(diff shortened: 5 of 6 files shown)


test/integration/sequelize.test.js
```diff
@@ -384,6 +384,54 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+          autoIncrement: true
+        },
+        title: DataTypes.STRING
+      });
+      Project.belongsTo(TopProject);
+      Project2.belongsTo(TopProject);
+      TopProject.hasMany(Project);
+      TopProject.hasMany(Project2);
+      await this.sequelize.sync({ force: true });
+      const project = await Project.create({ title: 'bla', topProject: { title: 'topProject1' } });
```
papb, line 417:
```suggestion
      const project = await Project.create({ title: 'bla', topProject: { title: 'topProject1' } }, { include: TopProject });
```

papb:
Hello! Thanks for catching this! Please fix the tests and I will merge, thanks!

bmeriaux:
MariaDb and MySQL tests are failing because of the foreign key, which require to truncate in cascade, so it seems to not have the cascade set properly ? is it. another issue because i didn't touch anything specific to an implementation

papb:
@bmeriaux Hmmm... Probably the fact that it was not being awaited caused the error to not be caught by CI. I agree this does not seem your fault.

Do you have any idea how to solve it, though?

bmeriaux:
> MariaDb and MySQL tests are failing because of the foreign key, which require to truncate in cascade, so it seems to not have the cascade set properly ? is it. another issue because i didn't touch anything specific to an implementation

After reading the MySQL doc, it does not support truncate with cascade on foreign key, so how to proceed ? add a if on dialect for the option to do a bulkDelete for maria and mysql ?

bmeriaux:
for. what i see here: https://dev.mysql.com/doc/refman/8.0/en/truncate-table.html
there is no possibility to use truncate on a table with foreign key, I see 3 choices:
- disable my test for MySQL and Maria, to keep the actual behavior which work when no FK for maria and mysql but wil throw an explicit error 
- try to detect foreign key ini query-interface.truncateTableQuery() for mysql and mariadb to do a delete instead, but it will be very slow on big table or disable contrainst but it needs to be only  when truncate  all tables
- disable truncate with cascade on mysql and mariadb dialect

papb:
Let me ping @jedwards1211, maybe he will have an idea of why this `await` was missing in the first place :)

bmeriaux:
> Let me ping @jedwards1211, maybe he will have an idea of why this `await` was missing in the first place :)

without it, i have a lot of deadlock error with postgres when truncatatin models with foreign key in cascade, because multiple truncate try to cascade on the same reference table at the same time

papb:
Thanks for the investigation @bmeriaux, intuitively I think your first bullet is the best option, but can you please do an extra investigation: does the sequelize documentation claim anything about truncate working or not working in mysql/mariadb/mssql?

We must give users whatever our docs says (or, maybe, change the docs)

papb:
> > Let me ping @jedwards1211, maybe he will have an idea of why this `await` was missing in the first place :)
> 
> without it, i have a lot of deadlock error with postgres when truncating models with foreign key in cascade, because multiple truncate try to cascade on the same reference table at the same time

Yes, I am sure the `await` is necessary, I decided to ping him because he was the one who made the transition in our codebase from promises to async/await, maybe when seeing this he might think of other places a similar mistake may have happened

bmeriaux:
i added an option in dialect support `truncateCascade`, and use it when there are model with at least an association (FK) to. use DELETE instead of truncate

jedwards1211:
Do you know where the original code went?  I don't see any missing awaits in the diff...seems totally plausible that one was missing though because there were a few heisenbugs in the refactoring

papb:
> explicit DB error will still be thrown

@bmeriaux Yeah, that's fine, if the underlying database does not support it, it is ok to have an error in Sequelize as well.

papb:
@jedwards1211 It's [this line](https://github.com/sequelize/sequelize/blob/e1446837196c07b8ff0c23359b958d68af40fd6d/src/sequelize.js#L831) in `src/sequelize.js` (former `lib/sequelize.js`)

bmeriaux:
finally, i removed the "best effort" to support truncate with cascade option set to true for dialect which does not support it, 
it is more explicit and closest to the reality

src/sequelize.js
```diff
@@ -818,17 +818,16 @@ class Sequelize {
...
-    this.modelManager.forEachModel(
-      model => {
-        if (model) {
-          models.push(model);
-        }
-      },
-      { reverse: false }
-    );
+    this.modelManager.forEachModel(model => {
+      if (model) {
```
papb, line 822: I know this is also not your fault, but while we're at it, do you know why is this `if` necessary? It seems really odd for `.forEachModel` to return a falsy model :sweat_smile:

test/integration/sequelize.test.js
```diff
@@ -384,6 +384,67 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+        title: DataTypes.STRING
+      });
+      const TopProject = this.sequelize.define(`top-project${config.rand()}`, {
+        id: {
+          type: DataTypes.INTEGER,
+          primaryKey: true,
+          autoIncrement: true
+        },
+        title: DataTypes.STRING
+      });
```
papb, line 411: I am 99% sure the following simplification should still work, can you please check?

```suggestion
      const Project = this.sequelize.define(`project-${config.rand()}`, { title: DataTypes.STRING });
      const Project2 = this.sequelize.define(`project2-${config.rand()}`, { title: DataTypes.STRING });
      const TopProject = this.sequelize.define(`top-project-${config.rand()}`, { title: DataTypes.STRING });
```

test/integration/sequelize.test.js
```diff
@@ -384,6 +384,67 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+      expect(project2.id).to.equal(1);
+      try {
+        await this.sequelize.truncate({ cascade: true });
+
+        const projectsCount = await Project.count();
+        expect(projectsCount).to.equal(0);
+        const projects2Count = await Project2.count();
+        expect(projects2Count).to.equal(0);
+        const topProjectsCount = await TopProject.count();
+        expect(topProjectsCount).to.equal(0);
```
papb, line 439:
```suggestion
        expect(await Project.count()).to.equal(0);
        expect(await Project2.count()).to.equal(0);
        expect(await TopProject.count()).to.equal(0);
```

test/integration/sequelize.test.js
```diff
@@ -384,6 +384,67 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+      } catch (e) {
+        if (this.sequelize.dialect.supports.truncateCascade) {
+          expect.fail();
```
papb, line 442: Can you please give a meaningful error message here, on why the test failed?

test/integration/sequelize.test.js
```diff
@@ -384,6 +384,67 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+        } else {
+          expect(e).to.instanceOf(Error, 'truncate with cascade is not supported for the current dialect');
```
papb, line 444:
```suggestion
          expect(e).to.be.instanceOf(Error, 'truncate with cascade is not supported for the current dialect');
```

jedwards1211:
okay, yup I don't remember if I automated that case of refactoring or not but if so the issue seems to have been that the iteratee was declared as a separate variable instead of provided as an inline function.  I looked over the rest of that commit for good measure and didn't manage to find any other mistakes like this at least

papb:
@jedwards1211 Hmm, I see. Can you give me the link to that commit? I'd like to take a look too

jedwards1211:
sure: https://github.com/sequelize/sequelize/pull/12074/files

papb:
@bmeriaux Thanks again for the great work!!

I want to make sure nothing here is a breaking change, I want to make more tests and take a closer look at the code (not only your changes but look at how things worked before), so I will not merge this PR yet.

However, since it is very obvious that this missing `await` is a fatal mistake, I would like to ask you a favor:

* Please open a new PR in which you do just a single line change, adding the missing `await`, no need to create any test
* Please do the same as the above bullet but targeting the `v6` branch instead of the `master` branch

I will merge those two as long as they don't break any existing tests, since it is very obvious that the `await` is necessary.

This way your issue is fixed asap while I don't risk making breaking changes unexpectedly.

bmeriaux:
> @bmeriaux Thanks again for the great work!!
> 
> I want to make sure nothing here is a breaking change, I want to make more tests and take a closer look at the code (not only your changes but look at how things worked before), so I will not merge this PR yet.
> 
> However, since it is very obvious that this missing `await` is a fatal mistake, I would like to ask you a favor:
> 
>     * Please open a new PR in which you do just a single line change, adding the missing `await`, no need to create any test
> 
>     * Please do the same as the above bullet but targeting the `v6` branch instead of the `master` branch
> 
> 
> I will merge those two as long as they don't break any existing tests, since it is very obvious that the `await` is necessary.
> 
> This way your issue is fixed asap while I don't risk making breaking changes unexpectedly.

it is done with #12663 and #12664

bmeriaux:
I rebased this one on master

PR #12690, opened 2020-09-11
fix(upsert): do not overwrite an explcit created_at during upsert

slickmb:
### Pull Request check-list

### Description of change

We found that when doing an upsert with model data that already included a `createdAt` timestampe, our explcit `createdAt` was being overwritten with the current time any time we also utilized the `fields` option.

e.g.

```
const instance = await MyModel.upsert({ id: 1, myfield: 'blah', createdAt: new Date('2010-01-01T12:00:00.000Z') },  { fields: [ 'myfield' ], returning: true });
console.log(instance.createdAt); // expected 2010-01-01T12:00:00.000Z, but got a now()-ish timestamp.
```

Issue appears to be that the check for a provided `createdAt` was being checked against the `updateValues` instead of the `insertValues`.  Most of the time, this is inconsequential because the `insertValues` and `updateValues` both contain the same fields.  But, when using the `fields` feature, the `createdAt` field is stripped away from the `updateValues`, so sequelize happily overwrites the `insertValues.createAt` value.

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -2571,7 +2571,7 @@ class Model {
     const now = Utils.now(this.sequelize.options.dialect);
 
     // Attach createdAt
-    if (createdAtAttr && !updateValues[createdAtAttr]) {
+    if (createdAtAttr && !insertValues[createdAtAttr]) {
       const field = this.rawAttributes[createdAtAttr].field || createdAtAttr;
       insertValues[field] = this._getDefaultTimestamp(createdAtAttr) || now;
     }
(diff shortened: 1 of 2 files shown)


slickmb:
I rebased my fix on the merge-base for master and v6, and then opened two pulls (against master and v6).  I did that based on the comments here: https://github.com/sequelize/sequelize#release

Please let me know if I'm not following the process correctly.

slickmb:
something seems off with that failing codecov report.

PR #12738, opened 2020-10-05
Add missing 'through' option in documentation example

lorainekv:
### Description of change

Documentation change only. 

In order to get the described output, I needed to add the `through` option to the query.


fncolon:
Look like the changes already made.

PR #12761, opened 2020-10-15
fix(transaction): nested transactions and transaction cls

Nafaya:
### Pull Request check-list

Issue:
https://github.com/sequelize/sequelize/issues/12762

### DIFF
--- a/src/dialects/abstract/query-interface.js
+++ b/src/dialects/abstract/query-interface.js
@@ -1247,7 +1247,6 @@ class QueryInterface {
     }
 
     options = { ...options, transaction: transaction.parent || transaction };
-    options.transaction.name = transaction.parent ? transaction.name : undefined;
     const sql = this.queryGenerator.startTransactionQuery(transaction);
 
     return await this.sequelize.query(sql, options);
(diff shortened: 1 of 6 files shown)


papb:

PR looks great, thank you! :+1: Before merging, can you please teach me a little - what is the difference between a transaction name and a transaction ID?

Nafaya:
Hi, @papb , thanks for your interest.
I think transaction `id` is unique id of a transaction, while `id` is inherited subtransactions(aka savepoints). As I understand `id` is used for tracking and logging sql queries. `id` is not sent to database.
While `name` is unique at all(i.e. for any `Transaction` instance). `name` is used as savepoint name (if we speak in sequelize terms - when you start subtransaction). `name` is send to database with `SAVEPOINT 'name'`, `ROLLBACK TO 'name'`

Also, I want to add here few words about savepoints and subtransactions. There terms are perceived slighty differently.
When you use subtransactions you would expect these to be completly undependent of each other, while savepoints mean just saved states with the corresponding identifiers and depent on their rollback functionality deponds on order in which savepoints were created. Unfortunately completely isolated subtransactions is what MySQL ot PostreSQL would allow to do.

This is the reason why I changed slighly architecture of parent-child relations of transactions.

PR #12791, opened 2020-10-26
Update query.js: Change assignment of "result" in "isInsertQuery" function.

florianmarkusse:
Hello,

### Pull Request check-list

- [ ] Have you added new tests to prevent regressions? _Not applicable_
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? _Not applicable_
- [ ] Did you update the typescript typings accordingly (if applicable)? _Not applicable_

### Description of change

In the `isInsertQuery` function, variable `result` was declared and reassigned in the following way:

`let result = true;`
`result = result && this.sql.toLowerCase().startsWith('insert into');`

Setting `result` to `true` first and then reassigning it with a logical AND-operation will always result in the variable being assigned to the second operand of the AND-operation. Therefore, these 2 lines can be combined into a single variable declaration:

`let result = this.sql.toLowerCase().startsWith('insert into');`

I hope this is clear enough and look forward to your comments.

Kind regards,
Florian

### DIFF
--- a/src/dialects/abstract/query.js
+++ b/src/dialects/abstract/query.js
@@ -167,14 +167,12 @@ class AbstractQuery {
   }
 
   isInsertQuery(results, metaData) {
-    let result = true;
-
     if (this.options.type === QueryTypes.INSERT) {
       return true;
     }
 
     // is insert query if sql contains insert into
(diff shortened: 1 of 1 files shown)


florianmarkusse:
Less code coverage because I compacted 2 lines into a single one.

PR #12838, opened 2020-11-15
Fix function definition for await keyword usage

fateslayer:
**await** keyword doesn't work inside a function if the function is not defined using **async** keyword

### Description of change

Added **async** keyword to the anonymous function


papb:

I made a mini-change to your edit before merging. Thank you for catching this!

fateslayer:
Thanks for fixing my mistake. I am glad I could contribute something.

papb:
@fateslayer Thank you very much!!

PR #12841, opened 2020-11-16
fix(mysql, mariadb): release connection on deadlocks

### LINKED ISSUE #11571: mariadb transaction deadlock does not release connection
### What are you doing?

Link to SSCCE PR: https://github.com/papb/sequelize-sscce/pull/15

```js
"use strict";

const CONFIG = require("../config");
const Sequelize = require("sequelize");

async function main() {
  const sequelize = new Sequelize(CONFIG.dbUrl, {
    pool: { max: 2, acquire: 1000 },
    keepDefaultTimezone: true, timezone: "" /* Suppress timezone warnings */
  });

  await sequelize.query("DROP TABLE IF EXISTS test1");
  await sequelize.query("CREATE TABLE test1 (id INT NOT NULL PRIMARY KEY)");
  await sequelize.query("DROP TABLE IF EXISTS test2");
  await sequelize.query("CREATE TABLE test2 (id INT NOT NULL PRIMARY KEY)");
  await sequelize.query("INSERT INTO test1 VALUES (1)");
  await sequelize.query("INSERT INTO test2 VALUES (1)");

  let t1 = await sequelize.transaction();
  let t2 = await sequelize.transaction();
  await sequelize.query("UPDATE test1 SET id = id WHERE id = 1", { transaction: t1 });
  await sequelize.query("UPDATE test2 SET id = id WHERE id = 1", { transaction: t2 });

  /* This query will wait on t2, don't block on it */
  sequelize.query("UPDATE test2 SET id = id WHERE id = 1", { transaction: t1 });
  try {
    await sequelize.query("UPDATE test1 SET id = id WHERE id = 1", { transaction: t2 });
  } catch (e) {
    console.log(e.message);
    console.log("t2 status is " + t2.finished);
    /* Cannot `t2.rollback()` here because it is marked as finished */
  }
  try {
    let t3 = await sequelize.transaction();
  } catch (e) {
    console.log(e.stack);
  }
}
main().then(() => process.exit(0));
```

The second query in `t1` will wait for `t2`, and the second query of `t2` will result in a deadlock. After this, `t2` should be rolled back, so we should have one available connection to start `t3`.

```
[...initial setup...]
Executing (68180761-6e97-443c-a070-9ca00810e563): START TRANSACTION;
Executing (50539931-2dd5-4b45-be57-6747c2af47e9): START TRANSACTION;
Executing (68180761-6e97-443c-a070-9ca00810e563): UPDATE test1 SET id = id WHERE id = 1
Executing (50539931-2dd5-4b45-be57-6747c2af47e9): UPDATE test2 SET id = id WHERE id = 1
Executing (68180761-6e97-443c-a070-9ca00810e563): UPDATE test2 SET id = id WHERE id = 1
Executing (50539931-2dd5-4b45-be57-6747c2af47e9): UPDATE test1 SET id = id WHERE id = 1
(conn=5493, no: 1213, SQLState: 40001) Deadlock found when trying to get lock; try restarting transaction
sql: UPDATE test1 SET id = id WHERE id = 1 - parameters:[]
t2 status is rollback
SequelizeConnectionAcquireTimeoutError: Operation timeout
    at [project]/node_modules/sequelize/lib/dialects/abstract/connection-manager.js:282:52
```

After one second of waiting, the timeout error is thrown for `t3`.

### Additional context

The problem seems to be these lines in `mariadb/query.js`:

```js
 59            // MariaDB automatically rolls-back transactions in the event of a deadlock
 60            if (options.transaction && err.errno === 1213) {
 61              options.transaction.finished = 'rollback';
 62            }
```

These mark the transaction as rolled back, but do not call `cleanup` to release the connection. (Adding `t2.cleanup()` in the catch seems to solve the problem.)

### Environment

- Sequelize version: `sequelize@5.19.8`
- Node.js version: `v12.11.1`
- Operating System: Ubuntu

- [x] I think this problem happens only for the following dialect(s): mariadb and mysql

Tested with mariadb, but the same lines appear in `mysql/query.js`.

hsource:
### Pull Request check-list

### Description of change

Closes #11571

This change makes deadlocked queries release the MariaDB/MySQL connection and does a redundant rollback. These ensure that subsequent queries work properly.

#### Debugging

While running sequelize in production for a collaborative document editor, we found that deadlocks on MariaDB would sometimes cause other queries to stall. After debugging, we realized that this is because MySQL/MariaDB has custom deadlock handling in Sequelize.

1. Because InnoDB (MariaDB/MySQL engine) automatically rolls back one query in the case of deadlocks, sequelize marks the `transaction.finished = 'rollback'` in the `run` function of [query.js](https://github.com/sequelize/sequelize/blob/4097aaa/src/dialects/mariadb/query.js#L54)
2. In `transaction()` of [sequelize.js](https://github.com/sequelize/sequelize/blob/4097aaa/src/sequelize.js#L1105), `transaction.rollback()` isn't called if `transaction.finished` is set
3. As such, `transaction.cleanup()`, which releases the connection is never called for this connection, and can cause hangs in this case

#### Fix

The fix was to ensure we call `transaction.cleanup()`, even if we don't rollback the transaction. I added a test for this case too.

Interestingly, on MariaDB, we still need to do an explicit rollback query. Without it, a follow-up READ COMMITTED transaction wouldn't behave as expected. I added this to the testcase too.

### DIFF
--- a/test/integration/transaction.test.js
+++ b/test/integration/transaction.test.js
@@ -424,13 +424,60 @@ if (current.dialect.supports.transactions) {
 
     if (dialect === 'mysql' || dialect === 'mariadb') {
       describe('deadlock handling', () => {
-        it('should treat deadlocked transaction as rollback', async function() {
-          const Task = this.sequelize.define('task', {
+        // Create the `Task` table and ensure it's initialized with 2 rows
+        const getAndInitializeTaskModel = async sequelize => {
+          const Task = sequelize.define('task', {
             id: {
               type: Sequelize.INTEGER,
               primaryKey: true
             }
           });
+
+          await sequelize.sync({ force: true });
+          await Task.create({ id: 0 });
+          await Task.create({ id: 1 });
+          return Task;
+        };
+
+        // Lock the row with id of `from`, and then try to update the row
+        // with id of `to`
+        const update = async (sequelize, Task, from, to) => {
+          await sequelize
+            .transaction(async transaction => {
+              try {
+                try {
+                  await Task.findAll({
+                    where: { id: { [Sequelize.Op.eq]: from } },
+                    lock: transaction.LOCK.UPDATE,
+                    transaction
+                  });
+
+                  await delay(10);
+
+                  await Task.update(
+                    { id: to },
+                    {
+                      where: { id: { [Sequelize.Op.ne]: to } },
+                      lock: transaction.LOCK.UPDATE,
+                      transaction
+                    }
+                  );
+                } catch (e) {
+                  console.log(e.message);
+                }
+
+                await Task.create({ id: 2 }, { transaction });
+              } catch (e) {
+                console.log(e.message);
+              }
+
+              throw new Error('Rollback!');
+            })
+            .catch(() => {});
+        };
+
+        it('should treat deadlocked transaction as rollback', async function() {
+          const Task = await getAndInitializeTaskModel(this.sequelize);
 
           // This gets called twice simultaneously, and we expect at least one of the calls to encounter a
           // deadlock (which effectively rolls back the active transaction).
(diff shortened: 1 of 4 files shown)


test/integration/transaction.test.js
```diff
@@ -414,73 +414,105 @@ if (current.dialect.supports.transactions) {
 
     if (dialect === 'mysql' || dialect === 'mariadb') {
       describe('deadlock handling', () => {
-        it('should treat deadlocked transaction as rollback', async function () {
-          const Task = this.sequelize.define('task', {
+        // Create the `Task` table and ensure it's initialized with 2 rows
+        const getAndInitializeTaskModel = async sequelize => {
```
papb, line 418: Nice refactor here on the tests to avoid duplicating code. Thank you!

test/integration/transaction.test.js
```diff
@@ -414,73 +414,105 @@ if (current.dialect.supports.transactions) {
...
+          await Promise.all([update(this.sequelize, Task, 1, 0), update(this.sequelize, Task, 0, 1)]);
+
+          // Verify that both of the connections were released
+          expect(this.sequelize.connectionManager.releaseConnection.callCount).to.equal(2);
+
+          // Verify that a follow-up READ_COMMITTED works as expected.
+          // For unknown reasons, we need to explicitly rollback on MariaDB,
+          // even though the transaction should've automatically been rolled
+          // back.
+          // Otherwise, this READ_COMMITTED doesn't work as expected.
```
papb, line 500: I think these comments belong on the code itself (not the test), do you agree?

test/integration/transaction.test.js
```diff
@@ -414,73 +414,105 @@ if (current.dialect.supports.transactions) {
...
+          });
+          await this.sequelize.sync({ force: true });
+          await this.sequelize.transaction(
+            { isolationLevel: Transaction.ISOLATION_LEVELS.READ_COMMITTED },
+            async transaction => {
+              const users0 = await User.findAll({ transaction });
+              await expect(users0).to.have.lengthOf(0);
+              await User.create({ username: 'jan' }); // Create a User outside of the transaction
+              const users = await User.findAll({ transaction });
+              return expect(users).to.have.lengthOf(1); // We SHOULD see the created user inside the transaction
```
papb, line 512:
```suggestion
              await expect(users).to.have.lengthOf(1); // We SHOULD see the created user inside the transaction
```

src/transaction.js
```diff
@@ -154,6 +146,11 @@ class Transaction {
   }
 
   cleanup() {
+    // Don't release the connection if there's a parent transaction or
+    // if we've already cleaned up
+    if (this.parent || this.connection.uuid === undefined) return;
+
+    this._clearCls();
```
papb, line 153: I see you removed two calls to `this._clearCls();` above and added one here, why?

papb (changes requested):
Hi! Thanks for the PR!

test/integration/transaction.test.js
hsource: I agree - I added a longer comment to the code itself to reflect this!

```ts
    } catch (err) {
      // MariaDB automatically rolls-back transactions in the event of a
      // deadlock.
      //
      // Even though we shouldn't need to do this, we initiate a manual
      // rollback. Without the rollback, the connection ends up in a limbo
      // state where future transactions may not behave properly.
      //
      // For example (in our tests), a follow-up READ_COMMITTED transaction
      // doesn't work as expected unless we explicitly rollback the
      // transaction: it would fail to read a value inserted outside of that
      // transaction.
      if (options.transaction && err.errno === 1213) {
        try {
          await options.transaction.rollback();
        } catch (err) {
          // Ignore errors - since MariaDB automatically rolled back, we're
          // not that worried about this redundant rollback failing.
        }
        options.transaction.finished = 'rollback';
      }
```

test/integration/transaction.test.js
hsource: I actually removed the `await`s since `expect` in Chai is synchronous, so `expect` doesn't need to be returned or awaited. It makes the assertions a bit more readable too!

src/transaction.js
hsource: I think it makes more sense to put it here, because it ensures that we do the cleanup even if neither `rollback` nor `commit` are called. This is needed for the MySQL cases.

1. User calls `sequelize.transaction`
2. MySQL throws for a deadlock. It marks `transaction.finished = 'rollback'`
3. In sequelize.js, since the transaction is already finished, we call `transaction.cleanup` without calling `transaction.rollback`

If we only called `_clearCls()` in `rollback` or `commit`, it wouldn't get called in this case, and the CLS (Continuation Local Storage) wouldn't get cleared.

It also makes the `cleanup` function more closely mirror the `prepareEnvironment` function. I added a comment mentioning that too!

hsource:
Thanks for the prompt review! I improved the comments and pushed some new changes. @papb - feel free to take another look

hsource:
@papb - quick ping on this - can you take a look? We're currently using our own branch, and would love to get this checked in so we can get back to the normal releases.

hsource:
Quick follow-up ping! @papb - thanks again!

ckeboss:
Looks like this is more robust than https://github.com/sequelize/sequelize/pull/12588

test/integration/transaction.test.js
papb: @hsource Of course! Thank you!

papb:
Hello!! Thank you very much for this super high quality PR. Great work! Sorry to take so long!

papb (approved):
Great work!!

arbieo:
Thanks for this. I believe the scope of this fix includes postgres as well for the same reasons as why this effects mysql. We were seeing unreleased connections after deadlocks caused by non-transaction calls within transactions.

papb:
Hi @arbieo, if I understand correctly you are saying this PR also fixed a similar postgres issue? Great news then :)

arbieo:
@papb Correct. I made a tentative fix in my local for the issue we had and then found this online afterwards that is largely the same (I probably should have checked here first).

PR #12852, opened 2020-11-20
fix(query-generator): add column casting in nested select (#11749)

### LINKED ISSUE #11749: Postgresql : Bug in Sequelize query generator in findAndCountAll
### What are you doing?

> User Model:
```

  const User = sequelize.define("User",
    {
      id: {
        type: DataTypes.INTEGER,
        autoIncrement: true
      },
 uid: {
        type: DataTypes.STRING(20),
        unique: true,
        primaryKey: true,
        unique: {
          args: true,
          msg: "Uid already exists"
        },
        field: "uid"
      },
roles: {
        type: DataTypes.ARRAY(DataTypes.STRING),
        field: "roles",
        validate: {
          validateRole(value) {
            if (!Array.isArray(value) || (Array.isArray(value) && value.length == 0)) {
              throw new Error('Role / Type is required.');
            }
          }
        },
        get: function () {
          var obj = [];

          if (this.getDataValue("roleList") == null) {
            obj = [];
          } else {
            obj = this.getDataValue("roleList");
          }

          return obj;
        },
        set: function (value) {
          let roleUids = _.pluck(value, "uid");

          this.setDataValue("roles", roleUids);
        }
      }, token: {
        type: DataTypes.STRING,
        field: "token"
      },
    },
    { tableName: "users", timestamps: true, paranoid: true }
  );

  User.associate = function (models) {
User.hasOne(models.SchoolStudents, { as: "schoolstudent", foreignKey: "user_uid", sourceKey: "uid" });
};
```

> schoolstudent Model:
```
  const SchoolStudents = sequelize.define(
    "SchoolStudents",
    {
      id: {
        type: DataTypes.INTEGER,
        autoIncrement: true
      },
      userId: {
        type: DataTypes.STRING(50),
        primaryKey: true,
        field: "user_uid"
      },specialties: {
        type: DataTypes.ARRAY(DataTypes.STRING),
        field: "specialties",
        get: function () {
          var obj = [];

          if (this.getDataValue("specialtyList") == null) {
            obj = [];
          } else {
            obj = this.getDataValue("specialtyList");
          }

          return obj;
        },
        set: function (value) {
          let specialties = [];

          if (value) {
            specialties = _.pluck(value, "uid");
          }

          this.setDataValue("specialties", specialties);
        }
      }},
    {
      freezeTableName: true,
      paranoid: true,
      tableName: "school_students",
      timestamps: true
    }
  );

  SchoolStudents.associate = function (models) {
    SchoolStudents.belongsTo(models.User, { foreignKey: "user_uid", targetKey: "uid" });
    SchoolStudents.hasMany(models.StudentSpecialties, { as: "specialtyList", foreignKey: "stud_user_uid", sourceKey: "user_uid" });
  };

```

I want filter based studentmodel specialties (character varying[] datatype)
> Query:
The below query , I am passing to findAndCountAll method
`"{"subQuery":false,"distinct":true,"limit":10,"order":[["uid","asc"]],"attributes":["firstName","lastName","email","program","phd","uid"],"offset":0,"where":{"$and":[{"roles":{"$contains":["RL-UkwtNA=="]}},{"$schoolstudent.specialties$":{"$contained":["STS-U1RTLTE=","STS-U1RTLTI=","STS-U1RTLTM="]}}]},"include":[{"as":"schoolstudent","attributes":[]}]}"`

The is querying generating like this ,

Output here
image

The sequelize query generator is generating for main table(parent table) is fine, 
for child table not generating properly, see the above screen shot.

please let me know the solution it is very urgent.

### Environment

- Sequelize version: 5.x
- Node.js version: 10.x 
- Operating System: windows 

eriksvedin2:
Pass includeMap to whereItemsQuery in order to get model properties from nested fields.

Closes #11749

### Pull Request check-list

### Description of change

Pass along includeMap in order to get field properties for nested fields.

See issue: (https://github.com/sequelize/sequelize/issues/11749) where for the nested array column it didnt add the varchar property due to it could'nt find out anything about the actual field in the _findField function.

### DIFF
--- a/src/dialects/abstract/query-generator.js
+++ b/src/dialects/abstract/query-generator.js
@@ -2542,6 +2542,16 @@ class QueryGenerator {
     if (options.model && options.model.fieldRawAttributesMap && options.model.fieldRawAttributesMap[key]) {
       return options.model.fieldRawAttributesMap[key];
     }
+
+    if (typeof key === 'string' && key.startsWith('$') && key.endsWith('$') && options.includeMap) {
+      const trimmed = key.slice(1, -1);
+      const fieldsArray = trimmed.split('.');
+      const modelName = fieldsArray[Math.max(fieldsArray.length - 2, 0)];
+      const keyName = fieldsArray[Math.max(fieldsArray.length - 1, 0)];
+      const { model } = options.includeMap[modelName];
+
+      if (model) return this._findField(keyName, { model });
+    }
   }
 
   // OR/AND/NOT grouping logic
@@ -2883,7 +2893,8 @@ class QueryGenerator {
       return this.whereItemsQuery(smth, {
         model: factory,
         prefix: prepend && tableName,
-        type: options.type
+        type: options.type,
+        includeMap: options.includeMap
       });
     }
     if (typeof smth === 'number') {
@@ -2900,13 +2911,15 @@ class QueryGenerator {
 
       return this.whereItemsQuery(where, {
         model: factory,
-        prefix: prepend && tableName
+        prefix: prepend && tableName,
+        includeMap: options.includeMap
       });
     }
     if (typeof smth === 'string') {
       return this.whereItemsQuery(smth, {
         model: factory,
-        prefix: prepend && tableName
+        prefix: prepend && tableName,
+        includeMap: options.includeMap
       });
     }
     if (Buffer.isBuffer(smth)) {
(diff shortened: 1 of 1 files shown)


eriksvedin2:
@papb ci seems to fail on nvm install, dont know if that is a problem related to this pr or something weird with the ci itself. Any hopes on looking at this?

PR #12921, opened 2020-12-19
fix(typings): raw=true queries return custom attributes

### LINKED ISSUE #12171: Queries with raw results cannot be typed properly
### What are you doing?

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/73

I would expect to be able to indicate the type of the query using generics; for example like this:

```ts
const myFoo = await Foo.findOne<{name: string}>({raw: true, attributes: ['name']})
```

Unfortunately this isn't accepted, because the typing demands that whatever is passed as a type parameter (in the above case, `{name: string}`) extends Sequelize's `Model`, which it obviously does not.

As per the example, `myFoo` gets type of `Foo`, which is incorrect -- it does not inherit all the trappings of Sequelize's `Model`.

### Environment

- Sequelize version: sequelize@5.21.3
- Node.js version: v12.14.1 
- Operating System: MacOS 10.15.4
- If TypeScript related: TypeScript version: v3.7.5

### LINKED ISSUE #12575: Typescript queries with `raw: true` should not return model instances
(Assuming Typings count as documentation)

## Issue Description

Currently when running a query with a model in Typescript, the result(s) are instances of the model.
e.g.
```
import { Model } from 'sequelize';

class Foo extends Model {};

async () => {
    // Foo.findAll() returns Promise<Foo[]>
    const foos: Foo[] = await Foo.findAll();
};
```

This is expected and ideal behavior.
However when running a query with `raw: true` (https://sequelize.org/master/manual/raw-queries.html) the response is very much not an instance of the model (_what_ it is is up for discussion in this issue).

The typing for queries that support raw should support returning non-model responses.

```
async () => {
    const foos: = await Foo.findAll({
        attributes: [fn('RAND'), 'random'],
        raw: true,
    });
    
    foos.forEach(foo => {
        // Typescript fails on `random` is not on `Foo`
        console.log(foo.random);
    });
};
```

Presently, all Model query methods that support `{raw: true}` still declare they return the model instance, when raw queries are explicitly _not_ instances.

So regular model attributes like `get` are not even available (which are used as a workaround for custom attributes on non-raw results).

Allow type parameters on methods that support `raw: true` to return a custom type, defaulting to the models raw attributes.

As an example, below is the existing definition for findAll: https://github.com/sequelize/sequelize/blob/v6/types/lib/model.d.ts#L1826
```
  public static findAll<M extends Model>(
    this: ModelStatic<M>,
    options?: FindOptions<M['_attributes']>): Promise<M[]>;
```

If we added another method definition that explicitly declared when `raw: true` to return a type parameter, it could correctly handle typing raw queries:
```
public static findAll<M extends Model>(
    this: ModelStatic<M>,
    options?: FindOptions<M['_attributes']> & { raw?: false }): Promise<M[]>;
  public static findAll<M extends Model, R extends any = M['_attributes']>(
    this: ModelStatic<M>,
    options?: FindOptions<M['_attributes']> & { raw: true }): Promise<R[]>;
```

The example query above could now be supported as:
```
Foo.findAll<Foo, { random: number }>({
   attributes: [fn('RAND'), 'random'],
   raw: true,
})
```

### Additional context

If there is another known method (or suggestions) for typing raw queries I'd appreciate the advice.

I know it is possible to just run the query as `raw: false` and use getters to load the custom attributes, but I believe raw queries serve their purpose when the extra noise of model instances is unnecessary. Not to mention the current typing is objectively wrong in this case.

- [ ] I don't know.

Happy to make an actual PR with my suggested `findAll()` as well as a couple others (possibly `findOne` + `findByPk`?) if the solution sounds correct

JacobLey:
Closes #12575
Closes #12171

 # Pull Request check-list

### Description of change

raw=true queries return custom attributes, not model instances. See issue https://github.com/sequelize/sequelize/issues/12575 for more discussion


mkadirtan:
When will this be merged?

JacobLey:
@papb I know some PRs have gotten hidden in the mess, wanted to bring attention to it for feedback

papb:
Hi @JacobLey, thank you! This looks good, but I need two things:

* Change the PR so it points to the `main` branch instead of `v6`. You can do this via the *edit* button next to the PR title:
    image
* Please add at least one test to make sure your implementation works as intended and to prevent regressions in the future. For good examples on how to test typescript typings, check the following PRs: #11368 #11379 #11520. You can also use the [`expect-type`](https://www.npmjs.com/package/expect-type#documentation) package to make complex typing tests if needed.

types/index.d.ts
```diff
@@ -7,6 +7,7 @@ import IndexHints = require('./lib/index-hints');
 import Utils = require('./lib/utils');
 
 export * from './lib/sequelize';
+export { default } from './lib/sequelize';
```
JacobLey, line 10: I wanted to call this out. Idk if it should have its own issue or whatnot, but I didn't expect much pushback so I'm throwing it in.

When using Sequelize with native ESM (`"type":"module"`) the **only** valid way to import sequelize is:
```
```
With that object being the one defined here: https://github.com/sequelize/sequelize/blob/main/lib/sequelize.js#L1369

Test for yourself with
```
import('sequelize').then(m => sequelize = m);
sequelize.default === require('sequelize') // true
sequelize.where() // throws error
```

Any "named exports" are _only_ good for Typescript typings, which is still really valuable, but should be treated as types _only_.
e.g. something like this: https://github.com/sequelize/sequelize/blob/main/types/test/where.ts#L2 _will fail_ when actually executed

Supporting named exports would be great, but that is not actually the case yet. I didn't want to get into a refactor of a bunch of other types/tests, but worth noting that the current advertised usage is invalid for ESM (it _may_ work with Babel-ified code, i haven't tested)

Typescript was already grabbing the default export if [allowSyntheticDefaultImports=true](https://www.typescriptlang.org/tsconfig/#allowSyntheticDefaultImports) was enabled, but it is nice to do it explicitly.

My suggestion would be to convert all the non-default export statements in the `index.d.ts` file to `export type ...` to enforce this behavior. Then fix up tests to match.

That _may_ be a breaking change for some people (again idk if it currently works with Babel).

On second thought I probably will make another issue for this (not a blocker at all to this PR btw)

JacobLey: Opened https://github.com/sequelize/sequelize/issues/13132

Resolving this "conversation" because its noisy and unrelated to the PR

JacobLey:
Updated with type tests and targeting main

mkadirtan (dismissed):
Great job!

mkadirtan:
Please review & approve this, so it can be merged. I reviewed it and @JacobLey did a great job. However, someone with write access is required.

Keimeno (dismissed):

types/lib/model.d.ts
```diff
@@ -1825,25 +1828,43 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
   public static findByPk<M extends Model>(
     this: ModelStatic<M>,
     identifier: Identifier,
-    options: Omit<NonNullFindOptions<M['_attributes']>, 'where'>
+    options: Omit<NonNullFindOptions<M['_attributes']>, 'where'> & { raw?: false }
   ): Promise<M>;
+  public static findByPk<M extends Model, R extends any = M['_attributes']>(
```
JacobLey, line 1833: Continuing conversation from https://github.com/sequelize/sequelize/pull/13219#discussion_r642108698

@AllAwesome497 this is the PR for raw addition.

See attached issue, but the presence of `raw: true` does more than expose attributes, but allows overriding that type for whatever the response should be.

AllAwesome497: yea, my comment on the other PR should also allow that but without the overloading, if TS is smart enough to infer raw's type as its value, not the type of the object (eg, as having the property `raw` as true instead of `raw?: boolean`)

types/test/findAll.ts
```diff
@@ -0,0 +1,24 @@
...
+            ['bar', 'foo'],
+            'ignored',
+            [Sequelize.col('table.id'), 'xyz'],
+        ],
+        raw: true,
+    });
+    expectTypeOf(rawUsers).toEqualTypeOf<RawUser[]>();
+
+    // @ts-expect-error
+    rawUsers[0].id = 123; // not an instance
```
AllAwesome497, line 23: can you also include tests to make sure that TS infers these generics?

`User.findAll({ raw: true })` should at least infer that `M` is `typeof User`, but I'm not sure about the second generic.

AllAwesome497 (commented):
Looks pretty handy, though a few things:


- The additional generic you added (`R`) seems redundant.  I can't see any use case where someone should use that, assuming that the model is typed. 
- This typing isn't really accurate.  You can have includes with `raw: true`, and this will cause a typing error when additional fields are being used.
- I think it'd be handy to have a generic that sets raw in options, instead of having `& { raw: true }` and such everywhere.  Eg: 

```ts
type WithRaw<
  TOptions extends { raw?: boolean },
  Raw extends boolean
> = Omit<TOptions, 'raw'> & Raw extends true 
  ? { raw: true }
  : { raw?: false };

// in usage
function <M extends Model>findAll(
  this: M,
  options: WithRaw<
    FindOptions<M['_attributes']>,
    true
  >
): M['_attributes']
```

types/test/findAll.ts
JacobLey: What would those tests look like?

In the `findAll` definition, `M` is actually `User` not `typeof User`. The usage of the _instance_ as the type does seem a bit weird to me, but that is existing Sequelize behavior that I would prefer to leave unchanged, so that it is not a breaking change

JacobLey:
> The additional generic you added (R) seems redundant. I can't see any use case where someone should use that, assuming that the model is typed

I personally use it in many cases. It is useful when you want to use Sequelize for the actual query generation, but the response should not be an instance itself. My use cases are ensuring the response only has a single column (e.g. _only_ the `id` column, or combining multiple aggregate methods (e.g. output of MAX and MIN in one query).
The default behavior is the raw attributes because that would be the result of a default `Model.findAll({ raw: true })` but the custom generic allows much more powerful overloading

> This typing isn't really accurate. You can have includes with raw: true, and this will cause a typing error when additional fields are being used.

Can you elaborate on this? I'm not entirely clear what you mean. `raw` is an option on the top-level, not in the includes. The typing of `include` is unaffected by that option.
Or do you mean the response type? Including attributes from an include would result in returning the unparsed SQL results (e.g. `{ 'User->Project.id': 123 }`. If you want to have a parsed include, you probably should not be using `raw: true` option.

> * I think it'd be handy to have a generic that sets raw in options, instead of having `& { raw: true }` and such everywhere.  

Agreed the `& { raw: true }` is not beautiful

types/test/findAll.ts
AllAwesome497: I mean to just check that it works without specifying generics.  It can be a bit confusing for ppl just starting to use TS to get strange behavior with generics.

AllAwesome497:
> I personally use it in many cases. It is useful when you want to use Sequelize for the actual query generation, but the response should not be an instance itself. My use cases are ensuring the response only has a single column (e.g. only the id column, or combining multiple aggregate methods (e.g. output of MAX and MIN in one query).

Understandable. I think this is something that can be solved from sequelize at some point, but I'm not sure if TS has evolved to the point where we can tell ts that the fields in attributes are the fields of the return type (esp considering how attributes can also be an object and other types.)

Personally I think its more readable to name it then, eg, `TReturn` instead of just R. Since its pretty   uniform here from what I see though its not gonna be a huge difference either way.

 > Can you elaborate on this? I'm not entirely clear what you mean. raw is an option on the top-level, not in the includes. The typing of include is unaffected by that option.

Yea I mean the return type.  Typing probably isn't advanced enough yet (at least to infer this), but eventually with ts we should be able to type these things via template literal types and such.  

Yes I mean `'User->Project.id'`.  I'd agree that its probably best not to use raw: true, but that doesn't mean its not possible. 

Probably not a big deal though if you have `R` there anyways since people can just override that generic. 

> At first glance though it does help a viewer determine why one typescript definition was used over the other

It's not this, but that's a fair point.  I think for me, if I hover over the generic with my cursor it'll expand it to be `options & {...`.

It's mostly an issue since this is also something that could be used elsewhere (upserts, inserts, updating, destroying, etc) (with returning: true which should also be typed!).

socramjunio2 (approved):

Darkle:
Any movement on this? Would love to have it.

richherbs:
What is happening with this as I am  running into this bug still?

WikiRik:
I'll reopen it since it was accidentally closed by the bot

types/test/findAll.ts
maxwrlr: @JacobLey I guess he means like so:
```suggestion
    const users = await User.findAll({ raw: false });
    expectTypeOf(users).toEqualTypeOf<User[]>()

    const users = await User.findAll({ raw: true });
    expectTypeOf(users).toEqualTypeOf<User['_attributes'][]>()

    interface RawUser {
        foo: 'bar';
    }
    const rawUsers = await User.findAll<User, RawUser>({
        attributes: [
            ['bar', 'foo'],
            'ignored',
            [Sequelize.col('table.id'), 'xyz'],
        ],
        raw: true,
    });
    expectTypeOf(rawUsers).toEqualTypeOf<RawUser[]>();

    // @ts-expect-error
    rawUsers[0].id = 123; // not an instance
```

src/model.d.ts
```diff
@@ -2207,7 +2207,10 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
    */
   public static findAll<M extends Model>(
     this: ModelStatic<M>,
-    options?: FindOptions<Attributes<M>>): Promise<M[]>;
+    options?: FindOptions<Attributes<M>> & { raw?: false }): Promise<M[]>;
+  public static findAll<M extends Model, R extends any = Attributes<M>>(
```
ephys, line 2211:
```suggestion
  public static findAll<M extends Model, R = Attributes<M>>(
```

src/model.d.ts
```diff
@@ -2216,25 +2219,43 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
   public static findByPk<M extends Model>(
     this: ModelStatic<M>,
     identifier: Identifier,
-    options: Omit<NonNullFindOptions<Attributes<M>>, 'where'>
+    options: Omit<NonNullFindOptions<Attributes<M>>, 'where'> & { raw?: false }
   ): Promise<M>;
+  public static findByPk<M extends Model, R extends any = Attributes<M>>(
```
ephys, line 2224:
```suggestion
  public static findByPk<M extends Model, R = Attributes<M>>(
```

src/model.d.ts
```diff
@@ -2216,25 +2219,43 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
...
+    identifier: Identifier,
+    options: Omit<NonNullFindOptions<Attributes<M>>, 'where'> & { raw: true }
+  ): Promise<R>;
   public static findByPk<M extends Model>(
     this: ModelStatic<M>,
     identifier?: Identifier,
-    options?: Omit<FindOptions<Attributes<M>>, 'where'>
+    options?: Omit<FindOptions<Attributes<M>>, 'where'> & { raw?: false }
   ): Promise<M | null>;
+  public static findByPk<M extends Model, R extends any = Attributes<M>>(
```
ephys, line 2234:
```suggestion
  public static findByPk<M extends Model, R = Attributes<M>>(
```

src/model.d.ts
```diff
@@ -2216,25 +2219,43 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
...
 
   /**
    * Search for a single instance. Returns the first instance found, or null if none can be found.
    */
   public static findOne<M extends Model>(
     this: ModelStatic<M>,
-    options: NonNullFindOptions<Attributes<M>>
+    options: NonNullFindOptions<Attributes<M>> & { raw?: false }
   ): Promise<M>;
+  public static findOne<M extends Model, R extends any = Attributes<M>>(
```
ephys, line 2247:
```suggestion
  public static findOne<M extends Model, R = Attributes<M>>(
```

src/model.d.ts
```diff
@@ -2216,25 +2219,43 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
...
+    this: ModelStatic<M>,
+    options: NonNullFindOptions<Attributes<M>> & { raw: true }
+  ): Promise<R>;
   public static findOne<M extends Model>(
     this: ModelStatic<M>,
-    options?: FindOptions<Attributes<M>>
+    options?: FindOptions<Attributes<M>> & { raw?: false }
   ): Promise<M | null>;
+  public static findOne<M extends Model, R extends any = Attributes<M>>(
```
ephys, line 2255:
```suggestion
  public static findOne<M extends Model, R = Attributes<M>>(
```

test/types/findAll.ts
```diff
@@ -0,0 +1,27 @@
...
+            'ignored',
+            [col('table.id'), 'xyz'],
+        ],
+        raw: true,
+    });
+    expectTypeOf(customUsers).toEqualTypeOf<CustomUser[]>();
+
+    // @ts-expect-error
+    customUsers[0].id = 123; // not an instance
+};
```
ephys, line 27:
```suggestion
(async () => {
    const users = await User.findAll({ raw: false });
    expectTypeOf(users).toEqualTypeOf<User[]>();

    const rawUsers = await User.findAll({ raw: true });
    expectTypeOf(rawUsers).toEqualTypeOf<Attributes<User>[]>();

    interface CustomUser {
        foo: 'bar';
    }
    const customUsers = await User.findAll<User, CustomUser>({
        attributes: [
            ['bar', 'foo'],
            'ignored',
            [col('table.id'), 'xyz'],
        ],
        raw: true,
    });
    expectTypeOf(customUsers).toEqualTypeOf<CustomUser[]>();

    // @ts-expect-error
    customUsers[0].id = 123; // not an instance
})();
```

to prevent some linters flagging this as unused

test/types/findByPk.ts
```diff
@@ -1,3 +1,20 @@
+import { expectTypeOf } from 'expect-type'
+import { Attributes } from '@sequelize/core';
 import { User } from './models/User';
 
-User.findByPk(Buffer.from('asdf'));
+async () => {
+    const user = await User.findByPk(Buffer.from('asdf'));
+    expectTypeOf(user).toEqualTypeOf<User | null>()
+
+    const rawUser = await User.findByPk(123, { raw: true });
```
ephys, line 9: can you add a test for `raw:true & rejectOnEmpty:true`, as well as `raw: false & rejectOnEmpty:true` to make sure they still work properly?

ephys (changes requested):
Only a few small changes, otherwise LGTM

test/types/findByPk.ts
```diff
@@ -1,3 +1,34 @@
+import { expectTypeOf } from 'expect-type'
+import { Attributes } from '@sequelize/core';
 import { User } from './models/User';
 
-User.findByPk(Buffer.from('asdf'));
+async () => {
```
ephys, line 5: same note about making this a IIFE so linters don't error on this when we enable them on these files :)

As well as on `findOne.ts`

ephys (changes requested):

ephys (dismissed):

ephys:
Sorry for the delay, this one was buried in my github notifications

ephys:
Arg, looks like https://github.com/sequelize/sequelize/pull/14280 broke this PR.
I'll see about updating this, as I should have merged this PR first

WikiRik:

ephys:
I've updated the PR. I believe the test suite now tests all possible scenarios. I had to update the implementation as the new test suite unearthed a couple more bugs

test/types/findByPk.ts
```diff
@@ -19,4 +22,69 @@ import { expectTypeOf } from 'expect-type';
...
+
+  expectTypeOf(await User.findByPk(Buffer.from('asdf'), {
+    raw: false,
+  })).toEqualTypeOf<User | null>();
+
+  expectTypeOf(await User.findByPk(Buffer.from('asdf'), {
+    raw: undefined,
+  })).toEqualTypeOf<User | null>();
+
+  // combination
```
WikiRik, line 40: Do we want to combine the undefined as well?

WikiRik: Same for findOne

src/model.d.ts
```diff
@@ -2462,6 +2487,7 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
     options?: FindOptions<Attributes<M>>
   ): Promise<M | null>;
 
+
```
WikiRik, line 2490:
```suggestion
```

No need for this extra new line

WikiRik (commented):
Two small remarks; one about a possible missing test and a styling one

src/model.d.ts
ephys: Removed

test/types/findByPk.ts
ephys: Done

WikiRik (approved):

PR #12941, opened 2021-01-04
Feature/ts data types

carlosingles:
### Pull Request check-list

### Description of change

Hello 👋

I thought I would try to help with the typescript migration that is being performed for v7. I hope this will prove useful, I only did one file since I was unsure if my changes would be accepted or not. Please let me know if this work is something that you would like to see more of and I will continue to try and help with the migration.

I have a few questions regarding my PR. I currently do not have a full understanding of the `data-types` column definitions, so I did some guess work and in places where I was unsure I used `unknown` instead of `any`.

- I cannot find a reference of the static `parse` method being used, would anyone have any ideas on what this function is used for?
- The `_sanitize` function seems to take the primitive value and attempt to make it adhere to the column data type. However, in cases where it does not pass, it returns the raw value instead. Is it okay for this function to remain typed as `unknown` and in cases where we need to use it, we can define the correct types? (see line 439's `BOOLEAN.prototype._sanitize` method)
- I have made individual interfaces for the constructor options inline, is there a preferred method of where these types of interfaces should be placed?

Please do not hesitate to reach out to me, I am happy to hear any feedback that you may have :)

### DIFF
--- a/lib/data-types.ts
+++ b/lib/data-types.ts
@@ -14,6 +14,7 @@ import mysqlDataTypes from './dialects/mysql/data-types';
 import mariadbDataTypes from './dialects/mariadb/data-types';
 import sqliteDataTypes from './dialects/sqlite/data-types';
 import mssqlDataTypes from './dialects/mssql/data-types';
+import db2DataTypes from './dialects/db2/data-types';
 import snowflakeDataTypes from './dialects/snowflake/data-types';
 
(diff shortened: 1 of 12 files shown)


WikiRik:
@carlosingles sorry for the late response, but we are currently working on the migration to TypeScript as a next release of v6. So feel free to continue working on this but make sure that all changes are backwards compatible.

PR #12968, opened 2021-01-21
fix(dialect:mysql): support more GEOMETRY subtypes

OtaK:
More things got into this PR as the basic functionality was requiring more changes over the code base. I've kept the changes as minimal as possible, though.

Tell me if you want this squashed or the split commits are easier to reason with.

### Pull Request check-list

	* Yes
	* No
	* Yes
	* No
	* N/A

### Description of change

* Fixed mysql GEOMETRY type to allow more subtypes that were allowed even back in 5.7 (cf: https://dev.mysql.com/doc/refman/8.0/en/spatial-type-overview.html). The supported types now look like this:
	* GEOMETRY
	* POINT
	* LINESTRING
	* POLYGON
	* MULTIPOINT
	* MULTILINESTRING
	* MULTIPOLYGON
	* GEOMETRYCOLLECTION
* Fixed an issue in the MySQL query builer (when creating tables) that was incorrectly not wrapping default values in `()` as recommended in the standard to support functions as default values. This now behaves as per-spec
* Fixed an issue in the MySQL `SHOW INDEX` query results parser, which didn't support `DESC` indexes at all, even for MySQL 8.0. This now behaves correctly
* Fixed the `should allow the user to specify indexes in options` test in `test/integration/model.test.js` using `MyISAM` engine where it should use `InnoDB` as MyISAM does *not* support `DESC` indexes and thus this test was doomed to fail every single time. cf. https://dev.mysql.com/doc/refman/8.0/en/descending-indexes.html `Descending indexes are supported only for the InnoDB storage engine`
* `DEFAULT` now supports DB functions with the help of `Sequelize.fn`.

### DIFF
--- a/lib/data-types.js
+++ b/lib/data-types.js
@@ -844,10 +844,10 @@ class GEOMETRY extends ABSTRACT {
     this.srid = options.srid;
   }
   _stringify(value, options) {
-    return `GeomFromText(${options.escape(wkx.Geometry.parseGeoJSON(value).toWkt())})`;
+    return `ST_GeomFromText(${options.escape(wkx.Geometry.parseGeoJSON(value).toWkt())})`;
   }
   _bindParam(value, options) {
-    return `GeomFromText(${options.bindParam(wkx.Geometry.parseGeoJSON(value).toWkt())})`;
+    return `ST_GeomFromText(${options.bindParam(wkx.Geometry.parseGeoJSON(value).toWkt())})`;
   }
 }
 
@@ -887,10 +887,10 @@ class GEOGRAPHY extends ABSTRACT {
     this.srid = options.srid;
   }
   _stringify(value, options) {
-    return `GeomFromText(${options.escape(wkx.Geometry.parseGeoJSON(value).toWkt())})`;
+    return `ST_GeomFromText(${options.escape(wkx.Geometry.parseGeoJSON(value).toWkt())})`;
   }
   _bindParam(value, options) {
-    return `GeomFromText(${options.bindParam(wkx.Geometry.parseGeoJSON(value).toWkt())})`;
+    return `ST_GeomFromText(${options.bindParam(wkx.Geometry.parseGeoJSON(value).toWkt())})`;
   }
 }
 
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -94,7 +94,7 @@ module.exports = BaseTypes => {
     'MULTIPOINT',
     'MULTILINESTRING',
     'MULTIPOLYGON',
-    'GEOMETRYCOLLECTION',
+    'GEOMETRYCOLLECTION'
   ];
 
   class GEOMETRY extends BaseTypes.GEOMETRY {
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -86,7 +86,16 @@ module.exports = BaseTypes => {
     }
   }
 
-  const SUPPORTED_GEOMETRY_TYPES = ['POINT', 'LINESTRING', 'POLYGON'];
+  const SUPPORTED_GEOMETRY_TYPES = [
+    'GEOMETRY',
+    'POINT',
+    'LINESTRING',
+    'POLYGON',
+    'MULTIPOINT',
+    'MULTILINESTRING',
+    'MULTIPOLYGON',
+    'GEOMETRYCOLLECTION',
+  ];
 
   class GEOMETRY extends BaseTypes.GEOMETRY {
     constructor(type, srid) {
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -376,7 +376,11 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
     if (!typeWithoutDefault.has(attributeString)
       && attribute.type._binary !== true
       && Utils.defaultValueSchemable(attribute.defaultValue)) {
-      template += ` DEFAULT (${this.escape(attribute.defaultValue)})`;
+      if (attribute.defaultValue instanceof Utils.Fn) {
+        template += ` DEFAULT (${this.escape(attribute.defaultValue)})`;
+      } else {
+        template += ` DEFAULT ${this.escape(attribute.defaultValue)}`;
+      }
     }
 
     if (attribute.unique === true) {
--- a/lib/dialects/mysql/query-generator.js
+++ b/lib/dialects/mysql/query-generator.js
@@ -376,7 +376,7 @@ class MySQLQueryGenerator extends AbstractQueryGenerator {
     if (!typeWithoutDefault.has(attributeString)
       && attribute.type._binary !== true
       && Utils.defaultValueSchemable(attribute.defaultValue)) {
-      template += ` DEFAULT ${this.escape(attribute.defaultValue)}`;
+      template += ` DEFAULT (${this.escape(attribute.defaultValue)})`;
     }
 
     if (attribute.unique === true) {
--- a/lib/dialects/mysql/query.js
+++ b/lib/dialects/mysql/query.js
@@ -274,10 +274,17 @@ class Query extends AbstractQuery {
         item.fields = [];
       }
 
+      let order = undefined;
+      if (item.Collation === 'A') {
+        order = 'ASC';
+      } else if (item.Collation === 'D') {
+        order = 'DESC';
+      }
+
       acc[item.Key_name].fields[item.Seq_in_index - 1] = {
         attribute: item.Column_name,
         length: item.Sub_part || undefined,
-        order: item.Collation === 'A' ? 'ASC' : undefined
+        order
       };
       delete item.column_name;
 
--- a/test/integration/dialects/mysql/dao-factory.test.js
+++ b/test/integration/dialects/mysql/dao-factory.test.js
@@ -22,7 +22,7 @@ if (dialect === 'mysql') {
         const User = this.sequelize.define(`User${config.rand()}`, {
           username: { type: DataTypes.STRING, defaultValue: 'foo' }
         }, { timestamps: false });
-        expect(this.sequelize.getQueryInterface().queryGenerator.attributesToSQL(User.rawAttributes)).to.deep.equal({ username: "VARCHAR(255) DEFAULT ('foo')", id: 'INTEGER NOT NULL auto_increment PRIMARY KEY' });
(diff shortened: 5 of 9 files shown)


lib/data-types.js
```diff
@@ -844,10 +844,10 @@ class GEOMETRY extends ABSTRACT {
     this.srid = options.srid;
   }
   _stringify(value, options) {
-    return `GeomFromText(${options.escape(wkx.Geometry.parseGeoJSON(value).toWkt())})`;
+    return `ST_GeomFromText(${options.escape(wkx.Geometry.parseGeoJSON(value).toWkt())})`;
```
papb, line 847: Can you please update [this line](https://github.com/sequelize/sequelize/blob/6d03c09175b2b91e19b68dd200c38c085beedc2e/lib/data-types.js#L795) accordingly?

Also can you quickly explain what is the difference between these two functions? Why does mysql provide two similar-looking functions like this while one of them doesn't work for some types (which is what I am inferring)? (I never used this, could be obvious to you, sorry)

lib/dialects/mysql/data-types.js
```diff
@@ -86,7 +86,16 @@ module.exports = BaseTypes => {
...
+  const SUPPORTED_GEOMETRY_TYPES = [
+    'GEOMETRY',
+    'POINT',
+    'LINESTRING',
+    'POLYGON',
+    'MULTIPOINT',
+    'MULTILINESTRING',
+    'MULTIPOLYGON',
+    'GEOMETRYCOLLECTION'
+  ];
```
papb, line 98: Can you please add [at least one more example here](https://github.com/sequelize/sequelize/blob/6d03c09175b2b91e19b68dd200c38c085beedc2e/lib/data-types.js#L799-L829) to showcase one of these new types?

papb (changes requested):

----

> More things got into this PR as the basic functionality was requiring more changes over the code base. I've kept the changes as minimal as possible, though.

It's okay! Thank you!

> Tell me if you want this squashed or the split commits are easier to reason with.

Split commits are better. I will squash everything at the moment of merging.

> * Fixed mysql GEOMETRY type to allow more subtypes that were allowed even back in 5.7 (cf: https://dev.mysql.com/doc/refman/8.0/en/spatial-type-overview.html). The supported types now look like this:
>   
>   * GEOMETRY
>   * POINT
>   * LINESTRING
>   * POLYGON
>   * MULTIPOINT
>   * MULTILINESTRING
>   * MULTIPOLYGON
>   * GEOMETRYCOLLECTION

Can you please create an integration test in [geometry.test.js](https://github.com/sequelize/sequelize/blob/main/test/integration/model/geometry.test.js) for each of the new values? (`GEOMETRY`, `MULTIPOINT`, `MULTILINESTRING`, `MULTIPOLYGON`, `GEOMETRYCOLLECTION`). Just declare a model with a column of this new type, call `TheModel.create()` passing a value of this type for the column, and then read it back with `TheModel.findByPk` and assert that it was retrieved in the same way you wrote it.

> * Fixed an issue in the MySQL query builer (when creating tables) that was incorrectly not wrapping default values in `()` as recommended in the standard to support functions as default values. This now behaves as per-spec

Cool, thanks!

> * Fixed an issue in the MySQL `SHOW INDEX` query results parser, which didn't support `DESC` indexes at all, even for MySQL 8.0. This now behaves correctly
> 
> * Fixed the `should allow the user to specify indexes in options` test in `test/integration/model.test.js` using `MyISAM` engine where it should use `InnoDB` as MyISAM does _not_ support `DESC` indexes and thus this test was doomed to fail every single time. cf. https://dev.mysql.com/doc/refman/8.0/en/descending-indexes.html `Descending indexes are supported only for the InnoDB storage engine`

I apologize I didn't really understand this part. You said *this test was doomed to fail every single time* but clearly it is passing on the main branch. Did your change cause this test to start failing? Also, as the test is now, would it fail if your changes in the parser were undone?

> * `DEFAULT` now supports DB functions with the help of `Sequelize.fn`.

This is very cool if I understood correctly. Can you add a new integration test for this? If you're unsure where to put it, feel free to add it in [model.test.js](https://github.com/sequelize/sequelize/blob/main/test/integration/model.test.js).

papb:
Also, how does all of this relate to mariadb, for the best of your knowledge? Could the exact same changes be done for mariadb, maybe?

lib/data-types.js
gileze33: Looks like they are synonyms, and the recommendation from MySQL is to use the ST_ prefixed versions

```
ST_GeomFromText(), ST_GeometryFromText(), GeomFromText(), and GeometryFromText() are synonyms. For more information, see the description of ST_GeomFromText().

GeomFromText() and GeometryFromText() are deprecated; expect them to be removed in a future MySQL release. Use ST_GeomFromText() and ST_GeometryFromText() instead.
```

From https://dev.mysql.com/doc/refman/5.7/en/gis-wkt-functions.html

gileze33:
I'm in need of these exact changes, and was planning to prep a PR but saw this one - @OtaK do you know if you'll have a chance to apply the small changes suggested to the comment sections? If not I could instead make the changes if you're not going to get a chance

lib/data-types.js
OtaK: > Can you please update [this line](https://github.com/sequelize/sequelize/blob/6d03c09175b2b91e19b68dd200c38c085beedc2e/lib/data-types.js#L795) accordingly?

Will do!

> Looks like they are synonyms, and the recommendation from MySQL is to use the ST_ prefixed versions
> 
> ```
> ST_GeomFromText(), ST_GeometryFromText(), GeomFromText(), and GeometryFromText() are synonyms. For more information, see the description of ST_GeomFromText().
> 
> GeomFromText() and GeometryFromText() are deprecated; expect them to be removed in a future MySQL release. Use ST_GeomFromText() and ST_GeometryFromText() instead.
> ```
> 
> From [dev.mysql.com/doc/refman/5.7/en/gis-wkt-functions.html](https://dev.mysql.com/doc/refman/5.7/en/gis-wkt-functions.html)

Correct, the non-`ST_` versions have been deprecated for a few years now.

lib/dialects/mysql/data-types.js
OtaK: Will do as well, I used this branch extensively (in production) for the past year and as such I have quite a bit of code.

OtaK:
> Also, how does all of this relate to mariadb, for the best of your knowledge? Could the exact same changes be done for mariadb, maybe?

@papb: MariaDB, having taken a way more forward approach, doesn't have those issues with cross-major version compatibility & dealing with legacy cruft that the MySQL has to deal with. Basically MariaDB is unaffected as everything worked properly from the get go.


> I'm in need of these exact changes, and was planning to prep a PR but saw this one - @OtaK do you know if you'll have a chance to apply the small changes suggested to the comment sections? If not I could instead make the changes if you're not going to get a chance

I haven't had much time to touch up the changes in the past months, but I'll try to find an hour or two soon to do that.



> Hello! I see you are a first-time contributor, thank you for taking the time to help Sequelize! I hope to see more PRs from you in the future!
> 

This is actually my second contribution; I implemented the HAVING clause support back in 2014.


> > Tell me if you want this squashed or the split commits are easier to reason with.
> 
> Split commits are better. I will squash everything at the moment of merging.

Gotcha!

> > * Fixed mysql GEOMETRY type to allow more subtypes that were allowed even back in 5.7 (cf: [dev.mysql.com/doc/refman/8.0/en/spatial-type-overview.html](https://dev.mysql.com/doc/refman/8.0/en/spatial-type-overview.html)). The supported types now look like this:
> >   
> >   * GEOMETRY
> >   * POINT
> >   * LINESTRING
> >   * POLYGON
> >   * MULTIPOINT
> >   * MULTILINESTRING
> >   * MULTIPOLYGON
> >   * GEOMETRYCOLLECTION
> 
> Can you please create an integration test in [geometry.test.js](https://github.com/sequelize/sequelize/blob/main/test/integration/model/geometry.test.js) for each of the new values? (`GEOMETRY`, `MULTIPOINT`, `MULTILINESTRING`, `MULTIPOLYGON`, `GEOMETRYCOLLECTION`). Just declare a model with a column of this new type, call `TheModel.create()` passing a value of this type for the column, and then read it back with `TheModel.findByPk` and assert that it was retrieved in the same way you wrote it.

Will do!


> > * Fixed an issue in the MySQL `SHOW INDEX` query results parser, which didn't support `DESC` indexes at all, even for MySQL 8.0. This now behaves correctly
> > * Fixed the `should allow the user to specify indexes in options` test in `test/integration/model.test.js` using `MyISAM` engine where it should use `InnoDB` as MyISAM does _not_ support `DESC` indexes and thus this test was doomed to fail every single time. cf. [dev.mysql.com/doc/refman/8.0/en/descending-indexes.html](https://dev.mysql.com/doc/refman/8.0/en/descending-indexes.html) `Descending indexes are supported only for the InnoDB storage engine`
> 
> I apologize I didn't really understand this part. You said _this test was doomed to fail every single time_ but clearly it is passing on the main branch. Did your change cause this test to start failing? Also, as the test is now, would it fail if your changes in the parser were undone?

Yes sorry, me adding support for the MySQL 8.0 & MariaDB `DESC` indexes caused the test to start failing indeed on MyISAM since DESC indexes aren't supported there. Basically without support for DESC indexes, the test passes indeed because ASC indexes are supported in every single version of MySQL & MariaDB, including all of their engines.
The fix was needed when it comes to DESC indexes as there are conditionals (MySQL >= 8.0 w/ InnoDB) to how the index ends up being represented in the database. For instance, if you try to create a DESC index with a MyISAM table engine, you'd silently end up with an ASC index.

> > * `DEFAULT` now supports DB functions with the help of `Sequelize.fn`.
> 
> This is very cool if I understood correctly. Can you add a new integration test for this? If you're unsure where to put it, feel free to add it in [model.test.js](https://github.com/sequelize/sequelize/blob/main/test/integration/model.test.js).

Will do!

OtaK:
## Tracking PR touchup progress

- [ ] Update https://github.com/sequelize/sequelize/blob/6d03c09175b2b91e19b68dd200c38c085beedc2e/lib/data-types.js#L795 with mentions of `ST_` functions
- [ ] "Can you please create an integration test in geometry.test.js for each of the new values? (GEOMETRY, MULTIPOINT, MULTILINESTRING, MULTIPOLYGON, GEOMETRYCOLLECTION). Just declare a model with a column of this new type, call TheModel.create() passing a value of this type for the column, and then read it back with TheModel.findByPk and assert that it was retrieved in the same way you wrote it."
- [ ] Examples for the new data types here https://github.com/sequelize/sequelize/blob/6d03c09175b2b91e19b68dd200c38c085beedc2e/lib/data-types.js#L799-L829
- [ ] New integration tests (https://github.com/sequelize/sequelize/blob/main/test/integration/model.test.js?rgh-link-date=2021-09-10T17%3A46%3A52Z) for `DEFAULT` with `Sequelize.fn()`

PR #12970, opened 2021-01-23
fix(types): allow changing values on before hooks

### LINKED ISSUE #12959: Readonly arrays – why?
### What are you doing?

I want to make sure for some Models that regardless of the selected `attributes` a certain attribute is always included. Up to version 6.3.5 this code worked:

```js
Foo.init(
  {
    // ...
  },
  {
    // ...
    hooks: {
      beforeFind(options: FindOptions<SequenceStepAttributes>): void {
        if (options.attributes) {
          if (Array.isArray(options.attributes)) {
            options.attributes.push('type');
          } else {
            if (options.attributes.exclude) {
              let index = options.attributes.exclude.indexOf('type');
              while (index >= 0) {
                options.attributes.exclude.splice(index, 1);
                index = options.attributes.exclude.indexOf('type');
              }
            }
          }
        }
      }
    }
  }
);
```

The attribute `type` should always be queried.

The code does not compile, because 287607a changed all types within [`FindAttributeOptions`](https://github.com/sequelize/sequelize/blob/fbf3a4ccb3e3ed3eb9631da01d3109c240ab3d26/types/lib/model.d.ts#L484)  into `readonly`.

### Additional context

Is there some other way to achieve my goal?

### Environment

- Sequelize version: 6.4.0
- Node.js version: v14.15.3
- Operating System: osx
- TypeScript version: 4.1.3

papb:
### Pull Request check-list

- [X] (N/A) Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description of change

Closes #12959

### DIFF
--- a/types/lib/hooks.d.ts
+++ b/types/lib/hooks.d.ts
@@ -40,7 +40,7 @@ export interface ModelHooks<M extends Model = Model, TAttributes = any> {
     instance: M,
     options: InstanceUpdateOptions<TAttributes> | CreateOptions<TAttributes>
   ): HookReturn;
-  beforeBulkCreate(instances: readonly M[], options: BulkCreateOptions<TAttributes>): HookReturn;
+  beforeBulkCreate(instances: M[], options: BulkCreateOptions<TAttributes>): HookReturn;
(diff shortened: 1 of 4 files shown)


papb:
The failed checks are caused to a flaky mysql test that I will fix soon

PR #12991, opened 2021-02-01
fix(sqlite): update primary key on sqlite upsert (#12683)

### LINKED ISSUE #12683: Model upsert does not return updated model for PK when autoincrement and SQLite dialect
## Issue Description
When using `sqlite` as dialect and using model's `upsert` method, the returned model does not have the primary key field updated if it's defined as `autoincrement`.

### What are you doing?
Upserting a model where its PK is defined as auto increment.

I could not use the repo for SSCCE because only supports v5 of Sequelize and this behaviour is with v6.
I have created a public repo in my github that reproduces the error:
**Here is the link to the SSCCE for this issue:** https://github.com/gborobio73/sqlite-upsert 

Here's the code (as a test) that reproduces the issue.
```js
// file upsert.test.js
const { Sequelize, Model, DataTypes } = require('sequelize');

const sequelize = new Sequelize('database', 'username', 'password', {
  dialect: 'sqlite',
  storage: 'database/database.sqlite', // or ':memory:'
  dialectOptions: {
    // Your sqlite3 options here
  }
});

class User extends Model {}
User.init({
  id: {
    type: DataTypes.INTEGER,
    primaryKey: true,
    autoIncrement: true,
  },
  ssn: {
    type: DataTypes.STRING,
    unique: true,
  },
  username: DataTypes.STRING,
  birthday: DataTypes.DATE
}, { sequelize, modelName: 'user' });

afterAll(async () => {
  await sequelize.close();
});

test('should return model with id', async () => {

  const [jane] = await User.upsert({
    ssn: '123456-XXX',
    username: 'janedoe',
  });

  console.log('janedoe upsert', jane.toJSON());
  expect(jane.id).not.toBeNull(); // fails because the id is null.
});
```

_I expect that the PK of the model returned by `upsert` is updated to the correct value._

_The PK of the model returned by `upsert` has `null` as value_

Here's the test output
```
  ● should return model with id

    expect(received).not.toBeNull()

    Received: null

      39 |
      40 |   console.log('janedoe upsert', jane.toJSON());
    > 41 |   expect(jane.id).not.toBeNull(); // fails because the id is null.
         |                       ^
      42 | });
      43 |

      at Object.<anonymous> (test/upsert.test.js:41:23)
```
Console output:
```
console.log
    janedoe upsert {
      id: null,
      ssn: '123456-XXX',
      username: 'janedoe',
      birthday: 1980-07-19T22:00:00.000Z
    }

      at Object.<anonymous> (test/upsert.test.js:40:11)
```

### Environment

- Sequelize version: 6.3.5 
- Node.js version: 12.18.2 
- Operating System: macOS Catalina 10.15.6 (19G2021)

- [X] I think this problem happens only for the following dialect(s): SQLite 

- [X] Yes, I have the time but I don't know how to start, I would need guidance.

HelloWorld017:
### Pull Request check-list

  - N/A
  - N/A

### Description of change

- Make `getDatabaseMethod()` return 'run', instead of 'exec', to
  retrieve the Statement metadata.
- Update primary key from Statement metadata on upsert.

Closes #12683


papb:

papb:
I've basically undid what you did in 2d31ce0, in order to merge this asap. However, I am interested in knowing why we should always run those three tests, instead of only running if `current.dialect.supports.returnValues`.

papb:
Also, next time, whenever you open a PR, please pretend you're the reviewer and go to the ["files changed" tab](https://github.com/sequelize/sequelize/pull/12991/files) and take a look. The diff on `test/integration/model/upsert.test.js` was very confusing. It took me over 20 minutes to realize that you just moved the tests around while not really changing their content at all. I invite you to look at the diff yourself again: 2d31ce0

Ideally you would have added to the PR description something like:

> Hello! The diff on `test/integration/model/upsert.test.js` looks crazy but I just moved three tests outside the if statement and put them into a new `describe` block.

This would have made my review much easier.

papb (approved):

HelloWorld017:
Hello,

> Also, next time, whenever you open a PR, please pretend you're the reviewer and go to the "files changed" tab and take a look. The diff on test/integration/model/upsert.test.js was very confusing.

I'm very sorry about that. Next time when I open the PR, I'll add descriptions for diffs hard to recognize.

> However, I am interested in knowing why we should always run those three tests, instead of only running if current.dialect.supports.returnValues.

I have originally thought that at least PKs should be updated even the dialect doesn't support returnValues, and to make sure that, I have made those three tests run regardless the option.

But now, I think those test should be run only if returnValues are supported, as it is basically testing about returning option. And if a new dialect, without ability to get PK of updated row, is implemented, it could be problemsome. So I think it is good to revert the commit 2d31ce0baf844f1841250785d6be244c443adbbf.

Sorry for inconvenience and thank you for reviewing this PR.

papb:
@HelloWorld017 Awesome, thank you very much!!

PR #13014, opened 2021-02-08
fix(query): JSON value already an object

harpsdev:
### Pull Request check-list

### Description of change

Adding a type check before parsing JSON value (ensure that value is a string).
Was crashing as already an object in my context.

Behavior do not change if value is a string.

### DIFF
--- a/lib/dialects/mariadb/query.js
+++ b/lib/dialects/mariadb/query.js
@@ -197,8 +197,9 @@ class Query extends AbstractQuery {
       if (modelField.type instanceof DataTypes.JSON) {
         // Value is returned as String, not JSON
         rows = rows.map(row => {
-          row[modelField.fieldName] = row[modelField.fieldName] ? JSON.parse(
-            row[modelField.fieldName]) : null;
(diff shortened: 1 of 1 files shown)


papb:

harpsdev:
> 
> 
> Hello! I see you are a first-time contributor, thank you for taking the time to help Sequelize! I hope to see more PRs from you in the future!
> 
> Please add at least one test to make sure your implementation works as intended and to prevent regressions in the future. Take a look into recent merged PRs to see how others created tests. If you have trouble, let me know.

I do (have trouble), I am not sure where to add my test. Could you point me the the adequate path/file?
Knowing that my fix is about retrieving a value from a JSON column in a MariaDb context.

Thank you for your help.

papb:
You can add a test here: https://github.com/sequelize/sequelize/blob/main/test/integration/json.test.js

Also, although you found and fixed this for MariaDB, perhaps this problem also happens in other dialects, so once you write your test in this file, perhaps it will fail on another dialect (which is good - so we know about that, and you could also make the similar fix in the other dialects' codes).

Issue #13003, opened 2021-02-04
Support for custom dialect adapters


talyssonoc:
### Is your feature request related to a problem? Please describe.

Based on the code in the main branch right now I see we're not able to use externally-developed dialect adapters unless it's in the form of some patching to some of the existing ones (https://github.com/sequelize/sequelize/blob/main/lib/sequelize.js#L320-L338). We're working on the adapter for the CockroachDB dialect, which currently does this patch, but it would be awesome to provide a way to support external dialects.

### Describe the solution you'd like

Ideally, it would be good to be able to pass both a string (to select one of the Sequelize native dialects) or a function/class that would be used when the dialect is instantiated (here https://github.com/sequelize/sequelize/blob/main/lib/sequelize.js#L340). Something like this:

```js
    let Dialect;

    const dialectOption = this.getDialect();

    if(typeof dialectOption === 'string') {
      // Requiring the dialect in a switch-case to keep the
      // require calls static. (Browserify fix)
      switch (dialectOption) {
        case 'mariadb':
          Dialect = require('./dialects/mariadb');
          break;
        case 'mssql':
          Dialect = require('./dialects/mssql');
          break;
        case 'mysql':
          Dialect = require('./dialects/mysql');
          break;
        case 'postgres':
          Dialect = require('./dialects/postgres');
          break;
        case 'sqlite':
          Dialect = require('./dialects/sqlite');
          break;
        default:
          throw new Error(`The dialect ${this.getDialect()} is not supported. Supported dialects: mssql, mariadb, mysql, postgres, and sqlite.`);
      }
    } else {
      Dialect = dialectOption
    }

    this.dialect = new Dialect(this);
```

### Why should this be in Sequelize

Right now we don't have a defacto officially supported way to develop custom dialects, that would be a good improvement.

### Describe alternatives/workarounds you've considered

It's possible to have some kind of workaround for it patching some of the native ones (e.g. https://github.com/cockroachdb/sequelize-cockroachdb/blob/master/index.js#L42) but it comes with the drawback that it makes me no longer able to use both PostgreSQL and CockroachDB in the same app with different Sequelize instances.

### Is this feature dialect-specific?

- [x] No. This issue is relevant to Sequelize as a whole.
- [ ] Yes. This issue only applies to the following dialect(s): XXX, YYY, ZZZ

### Would you be willing to implement this feature by submitting a Pull Request?

- [x] Yes, I have the time and I know how to start.
[...]

PR #13016, opened 2021-02-09
Support custom dialect adapters

douglasselias:
### Pull Request check-list

### Description of change

Add support for custom dialect by passing an object in the constructor instead of a string.

Closes [#13003](https://github.com/sequelize/sequelize/issues/13003)

### DIFF
--- a/docs/manual/core-concepts/getting-started.md
+++ b/docs/manual/core-concepts/getting-started.md
@@ -41,7 +41,7 @@ const sequelize = new Sequelize({
 // Option 2: Passing parameters separately (other dialects)
 const sequelize = new Sequelize('database', 'username', 'password', {
   host: 'localhost',
-  dialect: /* one of 'mysql' | 'mariadb' | 'postgres' | 'mssql' or a custom dialect object */
+  dialect: /* one of 'mysql' | 'mariadb' | 'postgres' | 'mssql' or a custom dialect class */
 });
 ```
 
--- a/docs/manual/core-concepts/getting-started.md
+++ b/docs/manual/core-concepts/getting-started.md
@@ -41,7 +41,7 @@ const sequelize = new Sequelize({
 // Option 2: Passing parameters separately (other dialects)
 const sequelize = new Sequelize('database', 'username', 'password', {
   host: 'localhost',
-  dialect: /* one of 'mysql' | 'mariadb' | 'postgres' | 'mssql' */
+  dialect: /* one of 'mysql' | 'mariadb' | 'postgres' | 'mssql' or a custom dialect object */
 });
 ```
 
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -24,7 +24,7 @@ class ConnectionManager {
     this.config = config;
     this.dialect = dialect;
     this.versionPromise = null;
-    this.dialectName = typeof this.sequelize.options.dialect === 'string' ? this.sequelize.options.dialect : this.sequelize.options.dialect.prototype.name;
+    this.dialectName = typeof this.sequelize.options.dialect === 'string' ? this.sequelize.options.dialect : this.sequelize.options.dialect.name;
 
     if (config.pool === false) {
       throw new Error('Support for pool:false was removed in v4.0');
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -24,7 +24,7 @@ class ConnectionManager {
     this.config = config;
     this.dialect = dialect;
     this.versionPromise = null;
-    this.dialectName = typeof this.sequelize.options.dialect === 'string' ? this.sequelize.options.dialect : this.sequelize.options.dialect.name;
+    this.dialectName = typeof this.sequelize.options.dialect === 'string' ? this.sequelize.options.dialect : this.sequelize.options.dialect.prototype.name;
 
     if (config.pool === false) {
       throw new Error('Support for pool:false was removed in v4.0');
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -24,7 +24,7 @@ class ConnectionManager {
     this.config = config;
     this.dialect = dialect;
     this.versionPromise = null;
-    this.dialectName = this.sequelize.options.dialect;
+    this.dialectName = typeof this.sequelize.options.dialect === 'string' ? this.sequelize.options.dialect : this.sequelize.options.dialect.prototype.name;
 
     if (config.pool === false) {
       throw new Error('Support for pool:false was removed in v4.0');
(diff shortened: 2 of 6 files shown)


lib/sequelize.js
```diff
@@ -130,7 +130,7 @@ class Sequelize {
    * @param {string}   [options.username=null] The username which is used to authenticate against the database.
    * @param {string}   [options.password=null] The password which is used to authenticate against the database.
    * @param {string}   [options.database=null] The name of the database
-   * @param {string}   [options.dialect] The dialect of the database you are connecting to. One of mysql, postgres, sqlite and mssql.
+   * @param {string|object} [options.dialect] The dialect of the database you are connecting to. One of mysql, postgres, sqlite, mariadb and mssql, or a custom dialect object.
```
talyssonoc, line 133:
```suggestion
   * @param {string|object} [options.dialect] The dialect of the database you are connecting to. One of mysql, postgres, sqlite, mariadb and mssql, or a custom dialect class.
```

lib/sequelize.js
```diff
@@ -48,7 +48,7 @@ class Sequelize {
    * // option examples
    * const sequelize = new Sequelize('database', 'username', 'password', {
    *   // the sql dialect of the database
-   *   // currently supported: 'mysql', 'sqlite', 'postgres', 'mssql'
+   *   // currently supported: 'mysql', 'sqlite', 'postgres', 'mssql' or a custom dialect object
```
talyssonoc, line 51:
```suggestion
   *   // currently supported: 'mysql', 'sqlite', 'postgres', 'mssql' or a custom dialect class
```

lib/dialects/abstract/connection-manager.js
```diff
@@ -24,7 +24,7 @@ class ConnectionManager {
     this.config = config;
     this.dialect = dialect;
     this.versionPromise = null;
-    this.dialectName = this.sequelize.options.dialect;
+    this.dialectName = typeof this.sequelize.options.dialect === 'string' ? this.sequelize.options.dialect : this.sequelize.options.dialect.prototype.name;
```
talyssonoc, line 27: Do we want to get the class name here? If so it's not gonna be in the prototype but in the class itself:

```suggestion
    this.dialectName = typeof this.sequelize.options.dialect === 'string' ? this.sequelize.options.dialect : this.sequelize.options.dialect.name;
```

lib/sequelize.js
```diff
@@ -371,7 +378,7 @@ class Sequelize {
   /**
    * Returns the specified dialect.
    *
-   * @returns {string} The specified dialect.
+   * @returns {string|object} The specified dialect.
```
talyssonoc, line 381:
```suggestion
   * @returns {string|class} The specified dialect.
```

types/lib/sequelize.d.ts
```diff
@@ -166,7 +166,7 @@ export interface Config {
   };
 }
 
-export type Dialect = 'mysql' | 'postgres' | 'sqlite' | 'mariadb' | 'mssql';
+export type Dialect = 'mysql' | 'postgres' | 'sqlite' | 'mariadb' | 'mssql' | Object;
```
talyssonoc, line 169: Maybe we should create a type to represent an abstract dialect (https://github.com/sequelize/sequelize/blob/main/lib/dialects/abstract/index.js)

What do you think?

```suggestion
interface AbstractDialect {
  new (...args: any[]): any
}

export type Dialect = 'mysql' | 'postgres' | 'sqlite' | 'mariadb' | 'mssql' | AbstractDialect;
```

docs/manual/core-concepts/getting-started.md
```diff
@@ -41,7 +41,7 @@ const sequelize = new Sequelize({
 // Option 2: Passing parameters separately (other dialects)
 const sequelize = new Sequelize('database', 'username', 'password', {
   host: 'localhost',
-  dialect: /* one of 'mysql' | 'mariadb' | 'postgres' | 'mssql' */
+  dialect: /* one of 'mysql' | 'mariadb' | 'postgres' | 'mssql' or a custom dialect object */
```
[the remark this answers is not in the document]
douglasselias: ok, makes sense.

types/lib/sequelize.d.ts
douglasselias: good idea, I will create one.

lib/dialects/abstract/connection-manager.js
douglasselias: That would be the expected behavior, but for some reason it doesn't work without prototype

talyssonoc: What do we get when we call `.prototype.name`?

test/integration/sequelize-custom.test.js
```diff
@@ -0,0 +1,54 @@
...
+
+describe('Custom Dialect', () => {
+  describe('authenticate', () => {
+    it('connects to database successfully', async () => {
+      try {
+        await this.sequelize.authenticate();
+        expect(true).to.equal(true);
+      } catch (error) {
+        console.error(error);
+      }
```
talyssonoc, line 31: This is not a safe way to test promises. If this promise `rejects`, the test suite will not fail, just log. The safe way to do it is:

```suggestion
      return expect(this.sequelize.authenticate()).resolves.to.equal(true);
```

types/lib/sequelize.d.ts
```diff
@@ -166,7 +167,7 @@ export interface Config {
   };
 }
 
-export type Dialect = 'mysql' | 'postgres' | 'sqlite' | 'mariadb' | 'mssql';
+export type Dialect = 'mysql' | 'postgres' | 'sqlite' | 'mariadb' | 'mssql' | AbstractDialect;
```
talyssonoc, line 170: What we are gonna receive here is _not_ an _instance_ of `AbstractDialect`, that's why we can't pass it like this, because what we expect here is the class itself. If we wanna use the `AbstractDialect` type for that we can do it like this:

```suggestion
interface AbstractDialectConstructor {
  new(...args: any[]): AbstractDialect
}

export type Dialect = 'mysql' | 'postgres' | 'sqlite' | 'mariadb' | 'mssql' | AbstractDialectConstructor;
```

lib/dialects/abstract/connection-manager.js
talyssonoc: I see the other dialects define directly in the prototype (https://github.com/sequelize/sequelize/blob/main/lib/dialects/mysql/index.js#L57). If we wanna enforce it for custom dialects as well (which should be fine, I guess?) we gotta document it somewhere.

test/integration/sequelize-custom.test.js
talyssonoc, line 31: Using try/catch in tests is not safe. If this promise rejects the test suite will not fail. I see we have [chai-as-promised](https://www.chaijs.com/plugins/chai-as-promised/) installed in the package, so we can do:

```suggestion
       return  expect(this.sequelize.authenticate()).to.eventually.equal(true);
```

Issue #13026, opened 2021-02-11
Eager Loading improvement opportunity


thesame:
I found a place where eager loading efficiency could be improved a lot with
LATERAL JOIN syntax. It's applicable at least to PostgreSQL and MySQL.  It's
always easier to show with an example. I'll use these models:

```javascript
  class A extends Model {};
  A.init ({name: {type: DataTypes.STRING}}, {sequelize, timestamps: false});

  class B extends Model {};
  B.init({name: {type: DataTypes.STRING}}, {sequelize, timestamps:false});

  A.hasMany (B);
```

and this data:

```javascript
  await A.bulkCreate(
    ['A1', 'A2', 'A3'].map ((a) => ({
      name: a,
      Bs: ['B1','B2','B3'].map ((b) => ({
        name: a+b
      }))
    })),
    {
      include: B
    }
  );
```

in other words:

```
temp=# select * from "As";
 id | name
----+------
  1 | A1
  2 | A2
  3 | A3
(3 rows)

temp=# select * from "Bs";
 id | name | AId
----+------+-----
  1 | A1B1 |   1
  2 | A1B2 |   1
  3 | A1B3 |   1
  4 | A2B1 |   2
  5 | A2B2 |   2
  6 | A2B3 |   2
  7 | A3B1 |   3
  8 | A3B2 |   3
  9 | A3B3 |   3
(9 rows)
```

Now I'd like to fetch all A's with their B's but not all B's but with
LIMIT+OFFSET windowing. This is how I express it in Sequelize:

```javascript
  const res = await A.findAll ({
    include: {
      model: B,
      limit: 1,
      offset: 2,
      order: ['name']
    }
  });
```

This request produces correct result, but uses this couple of SQL queries for it:

```sql
SELECT "A"."id",
       "A"."name"
FROM "As" AS "A";


SELECT "B".*
FROM
  (SELECT *
   FROM
     (SELECT "id",
             "name",
             "AId"
      FROM "Bs" AS "B"
      WHERE "B"."AId" = 1
      ORDER BY "B"."name"
      LIMIT 1
      OFFSET 2) AS sub
   UNION ALL SELECT *
   FROM
     (SELECT "id",
             "name",
             "AId"
      FROM "Bs" AS "B"
      WHERE "B"."AId" = 2
      ORDER BY "B"."name"
      LIMIT 1
      OFFSET 2) AS sub
   UNION ALL SELECT *
   FROM
     (SELECT "id",
             "name",
[...]

PR #13029, opened 2021-02-12
LATERAL joins for complex Eager Loading

### LINKED ISSUE #13048: SQL generator is producing incorrect queries for non-trivial includes
**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/148 

This is the simplest case I've come up with:
```ts
    const user = sequelize.define ('user', {}, {timestamps: false});
    const team = sequelize.define ('team', {}, {timestamps: false});
    const project = sequelize.define ('project', {}, {timestamps: false});

    project.hasMany (user);
    user.belongsTo (team);

    await sequelize.sync ({force: true});

    await project.bulkCreate ([{}, {}]);

    await project.findAll ({
        include: [{
            model: user,
            "limit": 1,
            "include": [
                {
                    model: team
                }
            ],
            "attributes": ["id"],
        }]
    });
```

I expect the query to be correct and produce the result.

The query raises an error (the exact wording depends on the DB engine): ```column user.teamId does not exist```

I think this problem happens regardless of the dialect.

I hope the issue could be resolved by this PR: https://github.com/sequelize/sequelize/pull/13029 but it still needs some effor to be done.

thesame:
### Pull Request check-list

### Description of change

This commit addresses the issue with excessive queries on eager loading. Details with examples are in #13026. This PR also fixes #13048.

### DIFF
--- a/lib/dialects/abstract/index.js
+++ b/lib/dialects/abstract/index.js
@@ -65,7 +65,8 @@ AbstractDialect.prototype.supports = {
   groupedLimit: true,
   indexViaAlter: false,
   JSON: false,
-  deferrableConstraints: false
+  deferrableConstraints: false,
+  LATERAL: false
 };
 
 module.exports = AbstractDialect;
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -19,6 +19,17 @@ const IndexHints = require('../../index-hints');
 
 const QuoteHelper = require('./query-generator/helpers/quote');
 
+class PrebuiltWhere
+{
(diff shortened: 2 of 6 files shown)


thesame:
Ok, now the tests are passing. (Except one that fails randomly even on current "main" branch). Now my concern is function `quote` in `lib/dialects/abstract/query-generator.js`. This function is not aware of possible table aliases so I had to cheat in `getQueryOrders` by substituting actual model with object `{ name: options.tableAs }`. Maybe somebody could hint me with the proper way to respect table alias in `getQueryOrders` function.

PR #13042, opened 2021-02-17
fix(types): adds missing function signature to Model.prototype.previous() types

danielschwartz:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [N/A] Have you added new tests to prevent regressions?
- [N/A] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you update the typescript typings accordingly (if applicable)?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)?

### Description of change

In the type definition for `previous()` on a model instance, the only specified signature is for when you pass in an argument to get a specific properties previous value. However, Sequelize will let you call `previous()` with no arguments and get an object back that maps all changed properties to their previous values. This PR adds that additional function signature to the type definitions so you can use `previous()` that way as well.

PR #13044, opened 2021-02-17
fix(types): Update model.d.ts to give more correct return type for `.toJSON()`

JonathanWolfe:
### Description of change

Change the return type of a model's `.toJSON()` method from generic `object` to the more specifc and helpful `TModelAttributes`.

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -2820,7 +2820,7 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
    * Convert the instance to a JSON representation. Proxies to calling `get` with no keys. This means get all
    * values gotten from the DB, and apply all custom getters.
    */
-  public toJSON(): object;
+  public toJSON(): TModelAttributes;
 
   /**
    * Helper method to determine if a instance is "soft deleted". This is
--- a/types/test/model.ts
+++ b/types/test/model.ts
@@ -16,6 +16,8 @@ class OtherModel extends Model {}
 
 const Instance: MyModel = new MyModel({ int: 10 });
 
+const gotten = Instance.get();
+expectTypeOf(Instance.toJSON()).toEqualTypeOf<typeof gotten>();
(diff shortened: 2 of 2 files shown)


JonathanWolfe:
CI seems to have failed from a deadlock, but that has to be entirely unrelated to my changes because I only edited the type defs 😂

papb:

Please add at least one test to make sure your implementation works as intended and to prevent regressions in the future. For good examples on how to test typescript typings, check the following PRs: [11368](https://github.com/sequelize/sequelize/pull/11368/files), [11379](https://github.com/sequelize/sequelize/pull/11379/files), [11520](https://github.com/sequelize/sequelize/pull/11520/files), [13042](https://github.com/sequelize/sequelize/pull/13042/files). You can also use [expect-type](https://www.npmjs.com/package/expect-type#documentation) if you need more complex assertions for your test.

papb:
> 
> 
> CI seems to have failed from a deadlock, but that has to be entirely unrelated to my changes because I only edited the type defs 😂

Oh that deadlock was happening randomly with MySQL but I've fixed that now, won't bother again haha

JonathanWolfe:
@papb I added the tests like you requested. Give it another look over

types/lib/model.d.ts
```diff
@@ -2820,7 +2820,7 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
    * Convert the instance to a JSON representation. Proxies to calling `get` with no keys. This means get all
    * values gotten from the DB, and apply all custom getters.
    */
-  public toJSON(): object;
+  public toJSON(): TModelAttributes;
```
Nkmol, line 2823: Might be outside of the scope of this PR, but we could possibly even extend this type to use any generic specialization.

```
public toJSON<T extends TModelAttributes>(): T
```

This makes it possible to have a bit more control over the output when you want a very specific (derived) type. For example the case where certain properties are non-null or `Required`.

liSong5713:
waiting for this pr. it`s important for know  definitions

PR #13045, opened 2021-02-18
fix(postgres): reuse `searchPath` in `extractTableDetails` 

burgalon:
### Pull Request check-list

### Description of change

Reuse `searchPath` in `extractTableDetails` to make enums + addColumn/updateColumn queries match the default schema

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -41,8 +41,9 @@ class QueryGenerator {
   extractTableDetails(tableName, options) {
     options = options || {};
     tableName = tableName || {};
+    const schema = tableName.schema || options.schema || options.schema !== false && this.sequelize.options.searchPath || 'public';
     return {
-      schema: tableName.schema || options.schema || 'public',
+      schema,
       tableName: _.isPlainObject(tableName) ? tableName.tableName : tableName,
       delimiter: tableName.delimiter || options.delimiter || '.'
     };
--- a/test/unit/sql/enum.test.js
+++ b/test/unit/sql/enum.test.js
@@ -17,6 +17,11 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
         schema: 'foo'
       });
 
+      const FooUser2 = current.define('user', {
+        mood: DataTypes.ENUM('happy', 'sad')
+      }, {
+      });
+
       const PublicUser = current.define('user', {
         mood: {
           type: DataTypes.ENUM('happy', 'sad'),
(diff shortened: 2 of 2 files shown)


danAlmog13 (approved):

lib/dialects/abstract/query-generator.js
```diff
@@ -41,8 +41,9 @@ class QueryGenerator {
   extractTableDetails(tableName, options) {
     options = options || {};
     tableName = tableName || {};
+    const schema = tableName.schema || options.schema || options.schema !== false && this.sequelize.options.searchPath || 'public';
```
PanopticaRising, line 44: Is there a reason that `this.sequelize.options.searchPath` is used instead of `this.sequelize.options.schema`?

Also, thanks! We just spent an hour looking at this and came up with basically the same solution.

burgalon: it seems like `schema` and `searchPath` are somewhat similar but then used for different aspects of the Sequelize.... not sure exactly why. If I remember correctly, specifying `schema` only was not sufficient

burgalon:
@PanopticaRising i wonder how can we promote this PR to merge? is there a way to ping the authors?

burgalon:
can we please reopen? this is still relevant

eliasmistler:
+1 on reopening this. We're having issues with it as well

WikiRik:
Reopened, is there a reason this is for Sequelize 5 instead of the latest version? And is there something else that needs to be done or can a maintainer look at this?

burgalon:
@WikiRik I guess we need this both in v5 + v6? maybe a separate cherry-pick / PR to pick this to v6 as well?

burgalon:
@WikiRik can we please try to promote this?

WikiRik:
@burgalon We're only doing security updates for v5 but this feature still needs to be included you can update the PR so it's against main and then we can cherry-pick it to be included for v6

lib/dialects/abstract/query-generator.js
mike-usa: I agree with @PanopticaRising -- this function is utility in that it takes names and adds in the scope (schema) primarily for queries.  It takes a best guess at what that scope is and sets that as default.  In my testing, I've discovered it shouldn't default to `'public'`, which only applies to one specific database -- I have a PR that updates this line (https://github.com/sequelize/sequelize/pull/14578/files#r889673513)

The `searchPath` in postgres can be more than one schema. It is like a `$PATH` environment variable, that will check the locations for the bare resource you're requesting.  Which means if you have `SELECT * FROM users;` and 20 different tables called `users`, it's going to walk through each schema listed on `search_path` until it finds it.

This function is "dumb" in that it doesn't hit the database at all.  It doesn't know which schema the unqualified tableName would belong to.  If your search_path is `etl, test, qa, $USER, public`, how would Sequelize know which one to choose?

The answer: whatever is calling the function needs to be smarter if their table is that ambiguous.

ephys:
We're very sorry for rejecting this PR but unfortunately `searchPath` should not be used as the default schema. Instead the solution should be based on the global `schema` option. (`this.sequelize.options.schema`). `searchPath` can list multiple schemas and using it by default is going to lead to bugs.

Sequelize 6 [takes the global `schema` option into account](https://github.com/sequelize/sequelize/blob/v6/src/dialects/abstract/query-generator.js#L46) (something Sequelize 5 does not do). I recommend migrating to Sequelize 6+ to resolve this issue, as we consider Sequelize 5 to have reached its end of life.

ephys:
That depends on what the query is, but I recommend posting this in a "Help & Question" discussion here: https://github.com/sequelize/sequelize/discussions :)

PR #13067, opened 2021-03-03
This PR is a rework of #12937

shivamklr:
The following changes update the link of an obsolete resource that is the `GeoJSON` link in `data-types.js` file. The old resource is from 2008 and the new one is from 2016. @papb please leave a review.


papb:
Hi @shivamklr, this is great! Thank you very much!

papb (approved):

PR #13068, opened 2021-03-03
Update lodash version to 4.17.21

ivangrosseADSK:
In this version, a vulnerability is fixed.

Info: https://nvd.nist.gov/vuln/detail/CVE-2021-23337

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "debug": "^4.1.1",
     "dottie": "^2.0.0",
     "inflection": "1.12.0",
-    "lodash": "^4.17.20",
+    "lodash": "^4.17.21",
     "moment": "^2.26.0",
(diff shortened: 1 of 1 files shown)


f1729 (approved):
LGTM

papb:

This is not necessary though. The caret `^` already present in `^4.17.20` in our `package.json` is specifying that any version from `4.17.20` and above (below `5` though) is acceptable ([learn more](http://semver.org/)). This way, `npm` will automatically install `4.17.21` for anyone that installs/updates sequelize. You can see it for yourself by running `npm i sequelize` and then `node -p "require('lodash/package.json').version"`.

papb:
I will close for now, but let's discuss more if needed.

PR #13089, opened 2021-03-11
chore(development) make docker-compose file compatible with recent co…

vanthome:
…mpose versions

Recent versions don't like how the volume is specified, see the discussion and fix here:

https://stackoverflow.com/questions/59904878/docker-compose-volume-name-is-too-short-names-should-be-at-least-two-alphanume


papb (approved):

papb:
Thank you!!

PR #13093, opened 2021-03-11
fix(types): Allow transaction to be null in Transactionable interface (#13092).

### LINKED ISSUE #13092: TypeScript: The Transactionable interface should also accept null
### What are you doing?

When using TypeScript with `strictNullChecks` turned on, and with the following snippet:

```ts
await sequelize.query("SELECT 1", { transaction: null });
```

Should not produce a compilation error. Passing `null` to `transaction` allows a query to be run outside the context of a transaction when auto-transactions are enabled.

I think the correct definition of the `Transactionable` interface should be:

```ts
export interface Transactionable {
  /**
   * Transaction to run query under
   */
  transaction?: Transaction | null;
}
```

Compilation error:

```
Type 'null' is not assignable to type 'Transaction | undefined'.
```

### Environment

- Sequelize version: 5.22.3 
- Node.js version: v12.21.0 
- Operating System: Ubuntu 20.04 LTS
- If TypeScript related: TypeScript version: 4.2.3

yjwong:
### Pull Request check-list

- [x] Have you added new tests to prevent regressions? (typing issue)

### Description of change

When using TypeScript with `strictNullChecks` enabled, we cannot pass `null` to `transaction`, but this is actually allowed and used to execute in a non-transaction context when in an auto-transaction block.

Closes #13092.

I wish for this change to also be backported to v5. How do I go about doing that?

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -35,7 +35,7 @@ export interface Transactionable {
   /**
    * Transaction to run query under
    */
-  transaction?: Transaction;
+  transaction?: Transaction | null;
 }
 
 export interface SearchPathable {
--- a/types/test/transaction.ts
+++ b/types/test/transaction.ts
@@ -81,6 +81,6 @@ async function nestedTransact() {
 
 async function excludeFromTransaction() {
   await sequelize.transaction(async t =>
-    sequelize.query('SELECT 1', { transaction: null })
+    await sequelize.query('SELECT 1', { transaction: null })
   );
 }
--- a/types/test/transaction.ts
+++ b/types/test/transaction.ts
@@ -78,3 +78,9 @@ async function nestedTransact() {
   });
   await tr.commit();
 }
+
+async function excludeFromTransaction() {
+  await sequelize.transaction(async t =>
+    sequelize.query('SELECT 1', { transaction: null })
+  );
+}


papb (approved):

papb:
Hello! Thanks!

papb (changes requested):
Actually, can you please add a typings test [here](https://github.com/sequelize/sequelize/blob/main/types/test/transaction.ts) for this?

papb:
> I wish for this change to also be backported to v5. How do I go about doing that?

You create a new branch off the `v5` branch, do the same changes you did here (including a test, please), and open a new pull request selecting `v5` as the target branch (instead of `main`). And write something like this in its description:

> *This PR is a backport of #13093 into v5*

yjwong:
I tried running `npm run test-typings` locally but got this output:

```
> sequelize@0.0.0-development test-typings /home/yjwong/Documents/sequelize
> tsc -b types/tsconfig.json && tsc -b types/test/tsconfig.json

types/test/define.ts:30:43 - error TS2511: Cannot create an instance of an abstract class.

30   expectTypeOf<UserModel>().toMatchTypeOf(new User());
                                             ~~~~~~~~~~

types/test/define.ts:63:50 - error TS2511: Cannot create an instance of an abstract class.

63   expectTypeOf<UntypedUserModel>().toMatchTypeOf(new UntypedUser());
                                                    ~~~~~~~~~~~~~~~~~


Found 2 errors.
```

These errors doesn't seem to be related to the tests I added.

papb (approved):

papb:
@yjwong Hmm, that's weird, but since tests are passing on GitHub Actions, don't worry about it. Just in case, can you confirm what is your TS local version?

yjwong:
@papb I found out why, I manually installed TypeScript 4.1.x using `npm install --save-dev typescript@~4.1` and it ran without error. Seems like it's more of an issue with TypeScript 4.2.x.

PR #13096, opened 2021-03-12
feat(add-constraint): add defferable option

ThePooN:
### Pull Request check-list

### Description of change

Needed an `UNIQUE` and deferrable constraint that was bound to multiple fields, but I found out it was only implemented in model definition.

Also tested these changes in a project where I needed them.


papb:

papb (approved):
Great work! I really like it.

papb:
I have a few questions:

1. Would you be willing to open another PR just improving [this jsdoc comment](https://github.com/sequelize/sequelize/blob/6388507ebd916efe6adf3e6a94a4c5ce235dee2a/lib/deferrable.js#L90-L94)? (especially the last two lines, that currently have no description whatsoever)

2. I noticed the typings you added allow for `string | Deferrable`, and I know that [our typings is already like this](https://github.com/sequelize/sequelize/blob/6388507ebd916efe6adf3e6a94a4c5ce235dee2a/types/lib/transaction.d.ts#L148), but I was wondering, can't we stop allowing `string`? Do you, for example, happen to use raw strings for this in your codebase?

ThePooN:
> I have a few questions:
> 
>     1. Would you be willing to open another PR just improving [this jsdoc comment](https://github.com/sequelize/sequelize/blob/6388507ebd916efe6adf3e6a94a4c5ce235dee2a/lib/deferrable.js#L90-L94)? (especially the last two lines, that currently have no description whatsoever)

Will do - but I also wonder if we should not rather split these first 3 and the last 2. First 3 are used in constraints definition, while last 2 are used in transactions AFAIK.

>     2. I noticed the typings you added allow for `string | Deferrable`, and I know that [our typings is already like this](https://github.com/sequelize/sequelize/blob/6388507ebd916efe6adf3e6a94a4c5ce235dee2a/types/lib/transaction.d.ts#L148), but I was wondering, can't we stop allowing `string`? Do you, for example, happen to use raw strings for this in your codebase?

We don't use raw strings.  
AFAIK, all possible deferrable options are implemented, so I think we could stop allowing `string` in typings. I will submit a separate PR for that, based on that one.

papb:
Thank you very much! On e7decdd I disallowed strings on the part you just added (which is not a breaking change since the feature didn't exist). For the ones that already allow strings, we'll see in #13121. Thanks!!

PR #13102, opened 2021-03-14
fix(mysql): release connection on deadlocks

papb:
This PR is a follow-up on #12841, which didn't fix the problem entirely for MySQL.

* Commit e1e27b9 adds a [failing test](https://github.com/sequelize/sequelize/runs/2105186637)
* Commit cb3574c [fixes it](https://github.com/sequelize/sequelize/runs/2105204721)
* Commit 63d5508 is just a refactor to make the source code for MySQL and MariaDB more similar


papb:
Ping @hsource, this might interest you

hsource:
I took a quick post-change review and I really liked your `expectedExecutionOrder` testing method! It's a lot more rigorous. Looks good!

madshell:
Is this fixed/merged in 5.x ? it doesn't seem so and we have production apps running MySQL 5.6 which produces deadlock and hangs for no reason, I think it's related to this. I think this should be considered a security issue because of potential possible DDOS.

PR #13105, opened 2021-03-14
fix(types): add type Col to type ProjectionAlias

### LINKED ISSUE #13104: Typescript:  Cannot use method sequelize.col for attributes option
### What are you doing?

I'm using sequelize with typescript.
Trying to alias an attribute of included model with sequlize.col in attributes option:
```ts
const boards = await user.$get('boards', {
      attributes: [
        'id',
        'title',
        [sequelize.col('tasks.title'), 'aliasedName'],  // TypeError
      ],
      include: [
        {
          model: taskModel,
          as: 'tasks',
        }
      ]
}
```

There should be no error.

TypeError Occurs.

```
Type 'Col' is not assignable to type 'string | Literal | Fn'.
Property 'clone' is missing in type 'Col' but required in type 'Fn'.
```

### Environment

- Sequelize version: 6.5.0 
- Node.js version: 14.4.0 
- Operating System: ubuntu
- TypeScript version: 4.1.5

shkilo:
### Pull Request check-list

### Description of change

Added type Col to type ProjectionAlias so that we can use method sequelize.col in attributes option.

Closes #13104

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -479,7 +479,7 @@ export type Order = string | Fn | Col | Literal | OrderItem[];
  * Please note if this is used the aliased property will not be available on the model instance
  * as a property but only via `instance.get('alias')`.
  */
-export type ProjectionAlias = readonly [string | Literal | Fn, string];
+export type ProjectionAlias = readonly [string | Literal | Fn | Col, string];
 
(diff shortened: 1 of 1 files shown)


papb:

papb (approved):

PR #13106, opened 2021-03-14
fix(types): Model.prototype.previous() parameters and return value now properly typed

Keimeno:
### Pull Request check-list

- N/A Have you added new tests to prevent regressions?
- N/A Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description of change

Let's look at the following example code:
```js
// Instrument is a model
const instrument = Instrument.build({
	orchestraId: 1,
	purchaseDate: new Date(),
	type: "violin",
});

instrument.set("orchestraId", 3);

instrument.previous("orchestraId"); // => 1
```

In JavaScript, this works just fine, in TypeScript however, we get the following error: `Argument of type '"orchestraId"' is not assignable to parameter of type 'keyof Model<InstrumentAttributes, InstrumentCreationAttributes>'`, since currently the used type `keyof this` refers to the functions and properties of the Model instance, not the creation attributes. This is fixed with this pull request.

Also, if we call `instrument.previous('type')` it would return `undefined`, since it does not have a previous value. This is also taken care of with the new return type.

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -2708,7 +2708,7 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
   /**
    * Returns the previous value for key from `_previousDataValues`.
    */
-  public previous<K extends keyof this>(key: K): this[K];
+  public previous<K extends keyof TCreationAttributes>(key: K): TCreationAttributes[K];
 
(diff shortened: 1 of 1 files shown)


papb (approved):
Awesome, thanks!!

PR #13120, opened 2021-03-18
docs(deferrable): add doc for deferrable transactions

ThePooN:
### Description of change

As mentioned in #13096: improve documentation for `Deferrable` objects.

As also mentioned in my original PR, the 3 first objects (`INITIALLY_DEFERRED`, `INITIALLY_IMMEDIATE`, `NOT`) are meant to be used in constraints declaration (in model or constraints definition), while the 2 latter (`SET_DEFERRED`, `SET_IMMEDIATE`) are used within transactions to enable or disable deferring eligible constraints.  
The context of use of these properties is different, so I find it confusing that they are not more explicitly distinguished. I think we should split that class in two or something.  
But as that could be a breaking change, I'm rather leaving this call to maintainers and just improving documentation as previously mentioned.


papb:
> As also mentioned in my original PR, the 3 first objects (`INITIALLY_DEFERRED`, `INITIALLY_IMMEDIATE`, `NOT`) are meant to be used in constraints declaration (in model or constraints definition), while the 2 latter (`SET_DEFERRED`, `SET_IMMEDIATE`) are used within transactions to enable or disable deferring eligible constraints.

Ahh, so that's it! Thanks!

> The context of use of these properties is different, so I find it confusing that they are not more explicitly distinguished. I think we should split that class in two or something.

I find it confusing too.

> But as that could be a breaking change, I'm rather leaving this call to maintainers and just improving documentation as previously mentioned.

I am in favor of this change, can you please do a PR? But it is indeed a breaking change, so I will only merge it when v7 comes. It would be great if you could do the PR in advance, so it is ready when the time comes. No rush though.

papb (approved):

ThePooN:
After more consideration, I'm more skeptical about this.

For example, in `operators.js`, we have all kind of SQL operators, that could go in many different places in an SQL query. These aren't split at all.

I wonder if rather than splitting the class, we should not just name these differently, to make it more obvious in which context they belong?

papb:
I think that any attempt to use `Deferred.SET_DEFERRED` or `Deferred.SET_IMMEDIATE` when declaring a constraint should give an error and the typings should disallow.

Conversely, any attempt to use `Deferred.INITIALLY_IMMEDIATE`, `Deferred.INITIALLY_DEFERRED` or `Deferred.NOT` when starting a transaction should give an error and the typings should disallow.

Do you agree?

Whether this behavior is going to be implemented by splitting a class in two or in another way, I don't have a strong opinion right now.

ThePooN:
After more consideration, I'm more skeptical about this.

For example, in `operators.js`, we have all kind of SQL operators, that could go in many different places in an SQL query. These aren't split at all.

I wonder if rather than splitting the class, we should not just name these differently, to make it more obvious in which context they belong?

ThePooN:
I agree, maybe we should split them in two classes internally (so we can then throw when appropriate, and better typings), but still expose them the same way.

papb:
Great! Would you like to do a PR? The way you propose has the advantage of not being a breaking change.

PR #13121, opened 2021-03-18
Deferrable typings

ThePooN:
**This PR is based on #13096, and should be merged after.**

*******

### Description of change

As mentioned in #13096: disallow the use of `string` in deferrable properties, for TypeScript users.

All Deferrable options should be implemented in the `Deferrable` class already, and should be considered for addition if not.


papb:
Hello, I will not merge this yet as this technically a breaking change. I will keep this PR in mind when the time arrives to do breaking changes.

PR #13124, opened 2021-03-19
fix: update uuid

### LINKED ISSUE #13123: Package subpath './v1' is not defined by "exports"
## Issue Description

Sequelize v5 throws the error `Package subpath './v1' is not defined by "exports"` seems like this is an issue with a [deprecation in uuid](https://github.com/uuidjs/uuid#deep-requires-no-longer-supported).

I don't really understand why it happens since sequelize v5 uses uuid@v3 which should work but for some reason it doesn't.

### Additional context

Anywhere in the codebase `require("uuid/v1")` should be changed to `require("uuid").v1` like in Sequelize v6

### Environment

- Sequelize version: 5.22.4
- Node.js version: 14.15.4

UziTech:
### Pull Request check-list

### Description of change

Update uuid in sequelize v5 to prevent [deprecation error](https://github.com/uuidjs/uuid#deep-requires-no-longer-supported).

fixes #13123

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -2,7 +2,7 @@
 
 const util = require('util');
 const _ = require('lodash');
-const uuidv4 = require('uuid/v4');
+const uuidv4 = require('uuid').v4;
 const semver = require('semver');
 
 const Utils = require('../../utils');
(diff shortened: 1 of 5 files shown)


UziTech:
It seems the only [breaking changes](https://github.com/uuidjs/uuid/blob/master/CHANGELOG.md) in uuid are:
- use of RNG in browsers (Does Not Apply)
- remove v3 and v5 uuids (Does Not Apply)
- removed support for deep requiring (Fixed in this PR)

PR #13127, opened 2021-03-21
feat(types): defining models, offers stricter type safety by default

Keimeno:
### Pull Request check-list

### Description of change

***This PR currently introduces a breaking change***

#### Better API design for strictly defining models

The best way to offer type safety for a Sequelize model that is defined with the `define` method is to create a `ModelAttributes` and a `CreationAttributes` interface like the following:

```ts
interface InstrumentAttributes {
	id: number;
	type: string;
	purchaseDate: Date;
}

interface InstrumentCreationAttributes
	extends Optional<InstrumentAttributes, "id"> {}

const Instrument: ModelDefined<
  InstrumentAttributes, 
  InstrumentCreationAttributes
> = sequelize.define("instrument", {
	type: DataTypes.STRING,
	purchaseDate: DataTypes.DATE,
});
```

This generally works, but it has one flaw. We must always import the `ModelDefined` type. A better solution for this is to instead write:

```ts
const Instrument = sequelize.define<InstrumentAttributes, InstrumentCreationAttributes>("instrument", {
	type: DataTypes.STRING,
	purchaseDate: DataTypes.DATE,
});
```

This PR introduces this new API design.

#### Stricter type safety if not strictly passing `ModelAttributes` & `CreationAttributes`

Currently, when defining a model, without passing the `ModelAttributes` and `CreationAttributes`, we run into the following Problem; The return type is: `ModelCtor<Model>`, which does not offer type safety for attributes at all, meaning if we write this: `Instrument.create({ type: 'foo' })`, it won't give any type suggestions for the attribute `type`.

With this PR and out above example, but with the generics `InstrumentAttributes` and `InstrumentCreationAttributes` removed, `sequelize.define` would return the following type: 

```ts
ModelDefined<
    | Record<string, unknown>
    | ({
        type: unknown; 
        purchaseDate: unknown;
      } & Record<string, unknown>),
    {
      type: unknown; 
      purchaseDate: unknown;
    } & Record<string, unknown>
  >
```

With this addition, it is known that the attributes `type` and `purchaseDate` exist, but are of type `unknown`. It is also possible to add `attributes`, that were not explicitly specified during the model definition, such as `id`.

#### Summing up

This PR offers a more accessible API design, at the cost of introducing a breaking change. The breaking change is, that you can't strictly define the `ModelAttributes` and `CreationAttributes`, as you could before. I'm currently investigating this, to see if there's any way we can offer this backwards compatibility, and I'm open to ideas.

### DIFF
--- a/docs/manual/other-topics/typescript.md
+++ b/docs/manual/other-topics/typescript.md
@@ -21,7 +21,6 @@ Example of a minimal TypeScript project with strict type-checking for attributes
 import {
   Sequelize,
   Model,
-  ModelDefined,
   DataTypes,
   HasManyGetAssociationsMixin,
   HasManyAddAssociationMixin,
@@ -176,10 +175,10 @@ Address.init(
 );
 
 // And with a functional approach defining a module looks like this
-const Note: ModelDefined<
-  NoteAttributes,
-  NoteCreationAttributes
-> = sequelize.define(
+const Note = sequelize.define<
+    NoteAttributes,
+    NoteCreationAttributes
+  >(
   'Note',
   {
     id: {
--- a/types/lib/sequelize.d.ts
+++ b/types/lib/sequelize.d.ts
@@ -1,4 +1,3 @@
-import * as DataTypes from './data-types';
 import { HookReturn, Hooks, SequelizeHooks } from './hooks';
 import { ValidationOptions } from './instance-validator';
 import {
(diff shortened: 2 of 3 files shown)


Keimeno:
There may be more incompatibilities introduced by this than I initially thought. Closing this now, but this might get relevant for v7

PR #13128, opened 2021-03-21
types: use TModelAttributes for model.update keys parameter

rdsedmundo:


papb:

You must add an explanation on why this change is important. What is it fixing?

Please add at least one test to make sure your implementation works as intended and to prevent regressions in the future. For good examples on how to test typescript typings, check the following PRs: #11368 #11379 #11520

EmilioCristalli:
i was going to open a similar PR to fix this issue.

IMO this is important because right now, when you define a "typed" model (like `class User extends Model<UserAttributes, UserCreationAttributes>`), `user.update({ columnd1: value1 })` allows any object to be used for the update.
With the change in this PR it should only allow the attributes defined in the model with their appropriate value types (`UserAttributes`).

For the tests i would suggest adding something like this to `types/test/update.ts` (you might need to rebase with `main` to get this test file):

```ts
TestModel.update({ foo: 'bar' }, { where: {} });  // to validate models that were defined without types still allow any attribute/values

User.update({ username: 'name_123' }, { where: {} });
// @ts-expect-error
User.update({ foo: 'bar' }, { where: {} }); // because User doesn't have a `foo` attribute
```

let me know if you need help with this, i can try to update your branch, or create a new one reusing your branch

rdsedmundo:
Just updated the PR with the test @EmilioCristalli provided—thanks!

He has also already explained very well the reasoning for the change.

EmilioCristalli:
hi @papb !
do you think this can be merged? or anything else i can help with to unblock it?

emilio-notable:
@rdsedmundo @papb this PR might not be needed anymore now that https://github.com/sequelize/sequelize/pull/13216 has been merged, right?

EmilioCristalli:
@rdsedmundo @papb this PR might not be needed anymore now that https://github.com/sequelize/sequelize/pull/13216 has been merged, right?

sdepold:
@rdsedmundo @EmilioCristalli @Keimeno @AllAwesome497 Should we close this one?

PR #13139, opened 2021-03-24
feat: pass through keepAliveInitialDelayMillis to postgres

FauxFaux:
### Description of change

node-postgres accepts a `keepAliveInitialDelayMillis` parameter, in addition to the `keepAlive` parameter we already accept:
https://github.com/brianc/node-postgres/blob/2ef55503738eb2cbb6326744381a92c0bc0439ab/packages/pg/test/unit/client/set-keepalives-tests.js#L25

Allow this value to be passed through.

This can be useful, because the default on many OSes is two hours, which isn't ideal as a failover time for a production application.

Or, we could do what is documented?

> An object of additional options, which are passed directly to the connection library

Directly, as in, without deleting options?


Keimeno (approved):

Keimeno:
> Directly, as in, without deleting options?

@papb this was introduced in #2039, but I don't see a valid reason not to merge the objects directly, without limiting its keys, especially as this is also done in the mariadb dialect

https://github.com/sequelize/sequelize/blob/main/lib/dialects/mariadb/connection-manager.js#L70

PR #13145, opened 2021-03-25
fix(mssql): support instance.update on models with date as primary key

### LINKED ISSUE #12717: Model instance.get('dateAttr') returns js date instead of raw sql attribute
### What are you doing?

I have a model with a date as primary key.

I do this

```js
let item = await Model.findByPk(myDate);
await item.update({
  value: "new value",
});
await item.reload();
console.log(item.value); // logs old value
```

The value should have been updated

The value is not updated

### Additional context

`model.where` calls `this.get("value", { raw: true })` which returns a js Date

The update query is built with bound a param that is an actual date object, which seems to be the problem.

### Environment

- Sequelize version: 6.3.5 
- Node.js version: 10.13.0 
- Operating System: OsX

- [x] I don't know, I was using mysql

armellarcier:
### Pull Request check-list

### Description of change

Fix #12717


armellarcier:
Trial to fix #12717

armellarcier:
@sdepold I cannot believe this nonsense

armellarcier:
Can we please re open this?

PR #13155, opened 2021-03-29
feat(on-conflict-where): add on confluct do update where fixes #11656

### LINKED ISSUE #11656: updateOnDuplicate with where clause
## Issue Description
I'd like to be able to include a where clause in the a postgres upsert `INSERT ON CONFLICT DO UPDATE` statement. Something like:
```sql
INSERT INTO 
  person (name, last_action_tsp) 
  values ('steve', '2019-11-06T15:41:47.249Z') 
ON CONFLICT DO UPDATE 
  SET name=EXCLUDED.name, last_action_tsp=EXCLUDED.last_action_tsp 
WHERE 
   EXCLUDED.last_action_tsp > last_action_tsp
```
We only want to upsert if the timestamp (`last_action_tsp`) is greater than the value currently in the database.

I think an api like this could work but open to suggestions.
```js
Model.bulkCreate(data, {
  updateOnDuplicate: {
    columns: ['name', 'last_action_tsp'],
    where: {
      lastActionTsp: {
        [Op.lt]: literal('EXCLUDED.last_action_tsp'),
      },
    },
  },
});
```

This should be in sequelize since it is a supported by postgres but it looks like it may not be supported by mysql.

The only workaround I can think of is to run an insert followed by an update with the condition but this results in two queries when ideally this could be done in one.

- [x] Yes. This issue only applies to the following dialect(s): postgres

I implemented this based on a pervious PR https://github.com/inDream/sequelize/pull/4
I know the merged implementation is different than the one proposed, but I am comfortable diving in and implementing this if you are open to it.

aheuermann:
Closed #11656

**Wanting some feedback on the API and can adjust.** Mysql doesn't support a where clause for updates.

### Pull Request check-list

### Description of change
I'd like to be able to include a where clause in the a postgres upsert `INSERT ON CONFLICT DO UPDATE` statement.  This is not currently possibly with sequelize. Something like:
```sql
INSERT INTO 
  person (name, last_action_tsp) 
  values ('steve', '2019-11-06T15:41:47.249Z') 
ON CONFLICT DO UPDATE 
  SET name=EXCLUDED.name, last_action_tsp=EXCLUDED.last_action_tsp 
WHERE 
   EXCLUDED.last_action_tsp > last_action_tsp
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2559,6 +2559,9 @@ class Model {
       if (options.updateOnDuplicate && (dialect !== 'mysql' && dialect !== 'mariadb' && dialect !== 'sqlite' && dialect !== 'postgres')) {
         throw new Error(`${dialect} does not support the updateOnDuplicate option.`);
       }
+      if (options.updateOnDuplicate && options.where && dialect !== 'postgres') {
+        throw new Error(`${dialect} does not support the where clause option for bulkCreate.`);
+      }
 
       const model = options.model;
 
--- a/lib/dialects/abstract/index.js
+++ b/lib/dialects/abstract/index.js
@@ -39,7 +39,8 @@ AbstractDialect.prototype.supports = {
   inserts: {
     ignoreDuplicates: '', /* dialect specific words for INSERT IGNORE or DO NOTHING */
     updateOnDuplicate: false, /* whether dialect supports ON DUPLICATE KEY UPDATE */
-    onConflictDoNothing: '' /* dialect specific words for ON CONFLICT DO NOTHING */
+    onConflictDoNothing: '', /* dialect specific words for ON CONFLICT DO NOTHING */
+    onConflictDoUpdateWhere: false /* where dialect supports ON CONFLICT UPDATE with WHERE clause */
   },
   constraints: {
     restrict: true,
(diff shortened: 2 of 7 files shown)


revlucio (approved):

lib/model.js
```diff
@@ -2559,6 +2559,9 @@ class Model {
       if (options.updateOnDuplicate && (dialect !== 'mysql' && dialect !== 'mariadb' && dialect !== 'sqlite' && dialect !== 'postgres')) {
         throw new Error(`${dialect} does not support the updateOnDuplicate option.`);
       }
+      if (options.updateOnDuplicate && options.where && dialect !== 'postgres') {
+        throw new Error(`${dialect} does not support the where clause option for bulkCreate.`);
+      }
```
AllAwesome497, line 2564: Both sqlite and postgres support this.

Also;

There's 2 clauses which might show up for both dialects in an `ON CONFLICT DO UPDATE` statement: 
- `ON CONFLICT WHERE` - This is used for indices, only references to the target table are allowed. E.g. `ON CONFLICT WHERE "username"` vs `ON CONFLICT WHERE "users"."username"`.
- `ON CONFLICT DO UPDATE WHERE` - This doesn't help with indices - however it allows for more advanced clause for when to update.

AllAwesome497 (changes requested):

lib/model.js
AllAwesome497: Oh nvm, its actually `onConflictDoUpdateWhere` not `where`.  Might be worth simplifying this term, I'm using `conflictWhere` for the first `WHERE`.

clarencedesmarais:
Can we reopen this by any chance ? This is a very useful feature.

PR #13183, opened 2021-04-09
Use npm script to run commands with docker env vars

Uko:
### Pull Request check-list

### Description of change

NPM package.json options are not widely known and may be confusing. Moreover, in NPM v7 the `options` dictionary is renamed to `config` and breaks the current functionality https://docs.npmjs.com/cli/v7/configuring-npm/package-json#config.
By re-using an npm script to run commands with docker-specific environment variables we make the configuration more understandable and compatible with the current and the upcoming node versions

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,9 +123,6 @@
     "nodejs",
     "object relational mapper"
   ],
-  "options": {
-    "env_cmd": "-f ./test/config/.docker.env"
-  },
   "commitlint": {
     "extends": [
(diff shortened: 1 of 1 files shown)


papb:
Hello! Thanks for this, but I think its no longer necessary. Same thing I said [here](https://github.com/sequelize/sequelize/pull/13189#issuecomment-868947968). Sorry for not looking earlier. Let me know if you disagree or have anything else you want to say, thanks :)

PR #13191, opened 2021-04-13
Add schema to table name in references attribute for sqlite

### LINKED ISSUE #13190: Table name not prefixed with schema for references when using sqlite
## Issue Description

Hey guys not sure this is an issue or not because it's not really documented anywhere. It seems that when I add column to my model that has a `references` field, I also have to add the schema name with the dot notation.
```
Series = sequelize.define('series', {
  trainer_id: {
    type: DataTypes.INTEGER,
    references: {
      model: "public.trainers",
      key: "id"
    }
  }
});
```

Not sure if this is intended behavior or not. I would expect to be able use `model: 'trainers'` without adding the schema name, as described in v3 docs [here](https://sequelize.org/v3/docs/associations/#enforcing-a-foreign-key-reference-without-constraints).

I used sqlite in dialect.

A possible solution for this would be changing [attributesToSQL](https://github.com/sequelize/sequelize/blob/main/lib/dialects/sqlite/query-generator.js#L247) function something like
```
attributesToSQL(attributes, { table }) {
  ...
  sql += ` REFERENCES ${table.schema}.${referencesTable} (${referencesKey})`;
  ...
}
```

I noticed that for postgres it takes the schema into account: https://github.com/sequelize/sequelize/blob/main/lib/dialects/postgres/query-generator.js#L507-L509

### Environment

- Sequelize version: 6.3.5
- Node.js version: v12.8.0
- Operating System: Debian GNU/Linux 9.9 (stretch)

- [x] I think this problem happens only for the following dialect(s): sqlite

mihaildu:
fixes #13190


mihaildu:
Branched off from master, oops

PR #13193, opened 2021-04-13
Eliminate random MSSQL build failures

Uko:
### Pull Request check-list

### Description of change

Looks like the mssql docker image has a weird behaviour when you can't connect to the DB after stratup. The solution in the PR is to run the creation of the MSSQL test DB inside the health check command.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,18 +152,17 @@ jobs:
           SA_PASSWORD: Password12!
         ports:
           - 1433:1433
-        options: >-
-          --health-cmd="/opt/mssql-tools/bin/sqlcmd -S localhost -U SA -P "Password12!" -l 30 -Q \"SELECT 1\" || exit 1"
-          --health-start-period 10s
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 10
+        options: >- # we have to run the db creation inside of a health check or it has a high chance to fail…
+          --health-cmd="/opt/mssql-tools/bin/sqlcmd -S localhost -U SA -P \"Password12!\" -l 30 -Q \"CREATE DATABASE sequelize_test; ALTER DATABASE sequelize_test SET READ_COMMITTED_SNAPSHOT ON;\" || exit 1"
+          --health-start-period=10s
+          --health-interval=10s
+          --health-timeout=5s
+          --health-retries=10
     env:
       DIALECT: mssql
       SEQ_USER: SA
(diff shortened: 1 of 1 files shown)


Keimeno:
Possibly related: https://github.com/Microsoft/mssql-docker/issues/203

Uko:
@Keimeno thanks a lot!!!

Keimeno:
This issue first occurred [here](https://github.com/sequelize/sequelize/runs/2262841241?check_suite_focus=true) 9 days ago, from a commit that changed GitHub actions.

My guess would be that this might be a caching problem, or the commit triggered an update of `mssql` onto a newer version.

Uko:
I'm not an expert in containers, but from my experiments, it seems that if the DB creation script fails once, it will also fail on the following retries. On the other hand, the jobs never failed when I created the DB in the health-check script.

One person mentioned that SQL commands should be done from the same layer where you start the server https://github.com/microsoft/mssql-docker/issues/203#issuecomment-623470495. Maybe we need a custom image that creates the DB on startup?

Uko:
@Keimeno I don't see how the task from the commit that you referenced can impact the mssql builds. I also tried to rollback those changes, and that didn't help

Keimeno:
I don't think that the changes itself changed something about the pipeline, I think the changes triggered something within GitHub pipelines to either clear the cache everywhere and/or update the mssql servers

Keimeno (approved):

papb:

papb:
Thanks for looking into this... Indeed this is super crazy. Isn't it a little dangerous to do this in the healthcheck script, though? The health check is something that I'd expect to execute many times, to keep checking the health, so it seems bad to try to put code there that was supposed to only execute once. Thoughts?

I noticed in the issue linked by @Keimeno (thank you!) that some people seem to have gotten away with it with a sleep, maybe we could wrap that `/opt/mssql-tools/bin/sqlcmd` call retrying it a few times if it fails, with `sleep 2` in between or something?

papb:
Also, there was a mistake with the healthcheck command that I fixed after this PR was opened: https://github.com/sequelize/sequelize/commit/830909244e02ff7e0ddaa31166462a30d1f09ec6#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR158

It's probably the cause of the current conflict. Maybe it was just that? To be honest I have no idea how it could be even working before I made the above fix.

PR #13197, opened 2021-04-15
refactor(conn-mgr): remove overwrite of this.getConnection() in close()

prestonlimlianjie:
### Pull Request check-list

### Description of change

This PR modifies the `connectionManager` such that:
- the `close()` method does not overwrite the `getConnection()` method; instead, it sets `this.pool = null;`
- the `getConnection()` method checks `this.pool` and throws an error if it does not exist (either because `initPools()` has not been called, or `close()` has been called).

Closes https://github.com/sequelize/sequelize/issues/13198

### DIFF
--- a/lib/dialects/abstract/connection-manager.js
+++ b/lib/dialects/abstract/connection-manager.js
@@ -25,6 +25,7 @@ class ConnectionManager {
     this.dialect = dialect;
     this.versionPromise = null;
     this.dialectName = this.sequelize.options.dialect;
+    this.pool = null;
 
     if (config.pool === false) {
       throw new Error('Support for pool:false was removed in v4.0');
@@ -108,12 +109,8 @@ class ConnectionManager {
    * @returns {Promise}
    */
   async close() {
-    // Mark close of pool
-    this.getConnection = async function getConnection() {
-      throw new Error('ConnectionManager.getConnection was called after the connection manager was closed!');
-    };
-
-    return await this._onProcessExit();
+    await this._onProcessExit();
+    this.pool = null;
   }
 
   /**
@@ -242,6 +239,12 @@ class ConnectionManager {
    */
   async getConnection(options) {
     options = options || {};
+
+    if (!this.pool) {
+      throw new Error(`
+        ConnectionManager.getConnection was called either after the pool was closed or before the pool was initialized`
+      );
+    }
 
     if (this.sequelize.options.databaseVersion === 0) {
       if (!this.versionPromise) {


lib/dialects/abstract/connection-manager.js
```diff
@@ -108,12 +109,8 @@ class ConnectionManager {
...
   async close() {
-    // Mark close of pool
-    this.getConnection = async function getConnection() {
-      throw new Error('ConnectionManager.getConnection was called after the connection manager was closed!');
-    };
-
-    return await this._onProcessExit();
+    await this._onProcessExit();
+    this.pool = null;
   }
```
marcogrcr, line 114: **Nitpick:** You can short circuit this with:

```js
async close() {
  if (!this.pool) {
    return;
  }

  // (...)
}
```

Though some may consider it redundant given that `_onProcessExit()` already has this.

lib/dialects/abstract/connection-manager.js
```diff
@@ -243,6 +240,12 @@ class ConnectionManager {
   async getConnection(options) {
     options = options || {};
 
+    if (!this.pool) {
+      throw new Error(`
+        ConnectionManager.getConnection was called either after the pool was closed or before the pool was initialized`
+      );
+    }
```
marcogrcr, line 247: **Question:** Does it make sense to have a check on `initPools()` as follows?

```js
if (this.pool) {
  throw new Error("Cannot override initialized connection pool(s). Make sure the existing pool(s) are closed by first by invoking ConnectionManager.close");
}
```

This will prevent resource leakage by misusing the class.

prestonlimlianjie: This is a good idea - let's see if @papb agrees and I'll add it.

I was also thinking of adding the following tests:
- Client should not be able to run `initPools()` if there already is an existing pool
- Client should be able to reuse same connectionManager to init pool, get connection, and close pool multiple times
- Client should not be able to call `getConnection` if pool has not been initialized
- Client should not be able to call `getConnection` if pool has been closed

lib/dialects/abstract/connection-manager.js
prestonlimlianjie: I had that consideration - about it being redundant code - so I didn't add it

lib/dialects/abstract/connection-manager.js
papb: Hi @prestonlimlianjie and @marcogrcr, yes I agree!

@prestonlimlianjie please do it, and also add those tests, these are all great ideas. Thank you all very much. Sorry to take so long to give you the deserved attention.

PR #13210, opened 2021-04-20
fix(13204): use proper field for ARRAY(ENUM) type

### LINKED ISSUE #13204: Custom types on arrays in posgres generates the wrong name when using underscores in columns
## Issue Description
When using underscored column names in pg enum array types, bulkUpdate queries fails (Actually this will happen for any field uses the "field" property to rename the column, as the data-type file in the postgres dialect is not using the correct "overwritten" propery)

### What are you doing?

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/158

I want that when using a model with underscored columns on the table in a postgres database, custom types in array are casted to the proper type when doing queries (in particular bulkCreate)

For a simple model like:

```js
        BookDetails.init({
            uniqueName: {
                type: Sequelize.DataTypes.STRING(100),
                unique: true,
                allowNull: false,
                primaryKey: true
            },
            originalCategories: {
                type: Sequelize.DataTypes.ARRAY(Sequelize.DataTypes.ENUM({
                  values: ['drama', 'comedy'],
                })),
              },
        }, {
            underscored: true,
            modelName: 'BookDetails',
            sequelize
        });
```

when executing the bulkCreate I'm expecting that the data is inserted, but instead fails

```js
        await BookDetails.bulkCreate([{
            uniqueName: 'test',
            originalCategories: ['drama']
        }]);
```

the error output is clear

```sql
INSERT INTO "book_details" ("unique_name","original_categories","created_at","updated_at") VALUES ('test',ARRAY['drama']::"enum_book_details_originalCategories"[],'2021-04-16 19:38:40.298 +00:00','2021-04-16 19:38:40.298 +00:00') RETURNING "unique_name","original_categories","created_at","updated_at";
```

`enum_book_details_originalCategories` is not the proper type in the database, it should be `enum_book_details_original_categories` respecting the underscoring.

### Environment

- Sequelize version: 6.6.2
- Node.js version: v12.22.1
- Operating System: Ubuntu 20.04

### Dev notes ###
I locally patched my code with a simple sed command:
`sed -i 's/options\.field\.fieldName/options\.field\.field/g' ./node_modules/sequelize/lib/dialects/postgres/data-types.js`

And from what I followed in the code is just a change on line 483 at `lib/dialects/postgres/data-types.js` instead of 
`Utils.generateEnumName(options.field.Model.getTableName(), options.field.fieldName),` the code should read `          Utils.generateEnumName(options.field.Model.getTableName(), options.field.field),` (I debuged it to confirm and the sed patch works locally).

So the change is realively easy, my only concern is extra testing that may be needed.

nahog:
### Pull Request check-list

- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? - Not aplicable
- [x] Did you update the typescript typings accordingly (if applicable)? - Not aplicable

### Description of change

Closes #13204


papb (approved):

This made me very happy. This is a very high quality PR. Fixing a very high quality issue. With a very high quality SSCCE. Congratulations. Excellent work. Thank you so much :rocket::rocket:

PR #13215, opened 2021-04-22
fix(typings): returning can specify column names

JacobLey:
### Pull Request check-list

### Description of change

Update returning to match other methods like bulkCreate: https://github.com/sequelize/sequelize/blob/main/types/lib/model.d.ts#L772

Test was updated with attributes to enforce valid values, rather than `any`


papb:
Thank you!!

PR #13219, opened 2021-04-23
fix(typings): allow where in order + specify order options

JacobLey:
Based on https://github.com/sequelize/sequelize/pull/12921/files to get `findAll` test, but that should probably still be merged/reviewed separately

### Pull Request check-list

### Description of change

Allow `where()` functions in ORDER (useful for sorting by boolean).

Also explicitly declare order options for better type inference/autocomplete: https://github.com/sequelize/sequelize/blob/main/lib/dialects/abstract/query-generator.js#L735

Note the `OrderItemColumn` type still includes `string` so ending an array with any string is still valid

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -1818,7 +1818,10 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
    */
   public static findAll<M extends Model>(
     this: ModelStatic<M>,
-    options?: FindOptions<M['_attributes']>): Promise<M[]>;
+    options?: FindOptions<M['_attributes']> & { raw?: false }): Promise<M[]>;
+  public static findAll<M extends Model, R extends any = M['_attributes']>(
+    this: ModelStatic<M>,
+    options?: FindOptions<M['_attributes']> & { raw: true }): Promise<R[]>;
 
   /**
    * Search for a single instance by its primary key. This applies LIMIT 1, so the listener will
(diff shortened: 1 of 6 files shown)


types/test/findOne.ts
```diff
@@ -1,6 +1,17 @@
...
 import { User } from "./models/User";
 
-User.findOne({ where: { firstName: 'John' } });
+async () => {
+    const user = await User.findOne({ where: { firstName: 'John' } });
+    expectTypeOf(user).toEqualTypeOf<User | null>()
 
-// @ts-expect-error
-User.findOne({ where: { blah: 'blah2' } });
+    // @ts-expect-error
```
AllAwesome497, line 8:
```suggestion
    // @ts-expect-error blah is not an attribute of User
```

types/test/findAll.ts
```diff
@@ -0,0 +1,42 @@
+import { expectTypeOf } from 'expect-type'
+import { Sequelize, Op } from 'sequelize';
+import { User } from './models/User';
+
+async () => {
+    const users = await User.findAll({ raw: false });
+    expectTypeOf(users).toEqualTypeOf<User[]>()
```
AllAwesome497, line 7: haven't seen `expectTypeOf` before. Not sure how this is supposed work, given what I know of ts (and how asserts work, since iirc block scoped type assertions don't currently work). 

```suggestion
    const users: User[] = await User.findAll({ raw: false });
```

types/lib/model.d.ts
```diff
@@ -1825,25 +1830,43 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
...
     this: ModelStatic<M>,
     identifier?: Identifier,
-    options?: Omit<FindOptions<M['_attributes']>, 'where'>
+    options?: Omit<FindOptions<M['_attributes']>, 'where'> & { raw?: false }
   ): Promise<M | null>;
+  public static findByPk<M extends Model, R extends any = M['_attributes']>(
+    this: ModelStatic<M>,
+    identifier?: Identifier,
+    options?: Omit<FindOptions<M['_attributes']>, 'where'> & { raw: true }
+  ): Promise<R | null>;
```
AllAwesome497, line 1849: `plain: true` isn't being handled and I think you can do this in a neater way.

eg:
```ts
type ModelAttributes<TModel extends Model> = TModel['_attributes'];

  public static findByPk<
    M extends Model,
    TOptions extends Omit<
      FindOptions<ModelAttributes<M>>, 'where'
  >>(
    this: ModelStatic<M>,
    identifier?: Identifier,
    options?: TOptions
  ): Promise<
    null
    | TOptions extends { raw: true } | { plain: true } 
      ? ModelAttributes<M>
      : M
    > {
```

TS might not be smart enough to handle this though, not sure.

JacobLey: This conversation should occur on the PR this is based on https://github.com/sequelize/sequelize/pull/12921/files

But for the sake of response...
To my knowledge Sequelize doesn't advertise the `plain` option for Model lookups... so I wouldn't expect special type handling for that use case.

The only documentation I can find is for raw SQL queries+custom getters, both of which have custom typing for this use case.

Also that suggested change removes the custom typing for `raw` which is the point of that PR

types/test/findAll.ts
JacobLey: `expectTypeOf` actually performs stricter type comparison than Typescript in some cases, and is the general test pattern of sequelize.

The testing only happens in Typescript, which is fine since these files are type-only (never actually executed javascript)

types/test/findOne.ts
JacobLey: Added to base PR https://github.com/sequelize/sequelize/pull/12921/commits/9abadb1781610db10e16f308aac6131e4672a6c5

types/lib/model.d.ts
AllAwesome497: Nope, it still has the raw part, though I'm not sure if it'll work: depends on how smart ts is when it infers the typing from TOptions. It'll either be have `raw?: boolean` or the actual type of the value (might be able to use `as const` to get ts to infer it better, but I haven't looked into that.

The `TOptions extends { raw: true } | { plain: true } ? ModelAttributes<M> : M` will give the model attributes if `TOptions` has `raw: true` or `{ plain: true }` and fallback on M. 

Mostly just a matter of preference.

types/test/findAll.ts
AllAwesome497: :shrug: I'll take your word for it.

papb: @AllAwesome497 `expectTypeOf` uses a <s>hack</s> great idea involving conditional types and rest parameters, it really works. By the way it was created by our Umzug maintainer, mmkal :)

PR #13220, opened 2021-04-23
fix(typings): use correct ignoreDuplicates option, add create test

JacobLey:
### Pull Request check-list

### Description of change

`onDuplicate` is not a real option. Should be `ignoreDuplicates` https://github.com/sequelize/sequelize/blob/main/lib/dialects/abstract/query-generator.js#L196

Added tests and extended `returning` option


types/lib/model.d.ts
```diff
@@ -749,7 +754,7 @@ export interface BulkCreateOptions<TAttributes = any> extends Logging, Transacti
   individualHooks?: boolean;
 
   /**
-   * Ignore duplicate values for primary keys? (not supported by postgres)
+   * Ignore duplicate values for primary keys?
```
JacobLey, line 757: While the `ignoreDuplicates` constant is undefined for postgres, the option maps directly to `onConflictDoNothing` https://github.com/sequelize/sequelize/blob/main/lib/dialects/abstract/query-generator.js#L197 which is defined for postgres https://github.com/sequelize/sequelize/blob/main/lib/dialects/postgres/index.js#L48

types/lib/model.d.ts
```diff
@@ -1975,7 +1980,7 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
   public static create<M extends Model>(
     this: ModelStatic<M>,
     values: M['_creationAttributes'],
-    options: CreateOptions<M['_attributes']> & { returning: false }
+    options: Omit<CreateOptions<M['_attributes']>, 'returning'> & { returning: false }
```
JacobLey, line 1983: Should we give `ignoreDuplicates` similar treatment?

A conflict with "do nothing" could potentially end up in the instance not being saved to the database. Might be better to invalidate the response (`void`) and force a reload of the instance if necessary

JacobLey:
Pretty sure test failure is unrelated...

types/lib/model.d.ts
JacobLey: Actually I like that, let's do it. Better to be correct on the safe side.

papb (approved):
Great work! Thank you very much!

PR #13226, opened 2021-04-25
fix(utils): clone attributes before mutating them

### LINKED ISSUE #12868: Options parameter gets mutated between uses
## Issue Description
First of all, I'd like to thank the contributors and maintainers of this library. We are using Sequelize heavily, talking with a PostgreSQL database, across 5 different Node.JS services, each with its own schema.

Starting with v5.13.0, and as recently as v6.3.5, we are having issues with scopes defined in associations. Prior to that version (tested and verified with v5.12.3), when using, for example, Foo's addBars method, everything will come out ok. Since v5.13.0 (and tested and verified with versions between v5.13.10 and v6.3.5, inclusive), the updated records are missing the field set in the scope, inside the association definition.

### What are you doing?
The following description has the names of models and fields changed, for simplicity.

We have a Foo model, with a hasMany association to a Bar model. When defining this association, we also add a scope, to set a field, called 'someField', to 'foo', on Bars added to a Foo. This field is in camelCase in our code, but snake_case in the db

One of our services, on request, finds all Bars in the DB that match a select query, duplicates them, omitting the scoped field, and inserts the new records to the DB. Then, the service creates a new Foo record, and adds the newly created Bars to that Foo, using `foo.addBars(bars)`.

**Here is the link to the SSCCE for this issue:** LINK-HERE 
(I haven't been able to create a simple reproduction of this just yet, but I am still working on it. Will update this issue when I manage that).

The update query for the Bars should be `UPDATE "bars" SET "some_field"=$1,"updated_at"=$2 WHERE id IN (...)`.
The rows in the DB should have some_field set to the value defined in the scope inside the association.

The update query for the Bars comes out as `UPDATE "bars" SET "updated_at"=$1 WHERE id IN (...)`.
The rows in the DB should have some_field set to NULL.

### Additional context
I tried to pinpoint the change, between v5.12.3 and v5.13.0, that led to the issue I am describing, and I think I found the specific change that causes it.

In #11281, the library switched some places that used it's custom Utils.cloneDeep method to simpler `Object.assign({}, ...)` calls. One such change is in the select method in the query-interface.js file (was in lib, moved to src/dialects/abstract, and the line has changed since then to `{ ...options, ... }`, but the issue remains). When I revert that specific change, so it uses Utils.cloneDeep again, everything works perfectly, both in v6.3.5 and in v5.13.0.

We had similar issues in our repositories, where using `Object.assign({}, ...)` or using `immutability-helper` left some of the nested fields as references to the original, and updating the result of the assign altered the original as well. It seems to be a similar case in this issue.

I couldn't understand why that would cause the issue, and on further investigation, I saw that inside the update method, defined on the Model class, the `build.validate` function was passed an options parameter. This parameter changes between v5.12.3 and v5.13.0, specifically the `fields` field on the options object.

As you can see below (with the original fields and models), the fields passed to the validate function contain one field written in snake_case, and the rest in camelCase. This causes the resulting attributes.dataValues to not contain the scoped field (some_field in the simplification, asset_type in our actual code), meaning that line 3079 in model.js overwrites values, dropping the field we wanted to add.

I'm not sure where in the library is that fields array accessed or changed, prior to getting to the update method, and I couldn't find any obvious places where it might happen...

```
{
  transaction: Transaction {
    sequelize: Sequelize {
      options: [Object],
      config: [Object],
      dialect: [PostgresDialect],
      queryInterface: [PostgresQueryInterface],
      models: [Object],
      modelManager: [ModelManager],
      connectionManager: [ConnectionManager]
    },
    savepoints: [],
    _afterCommitHooks: [],
    options: { type: 'DEFERRED', isolationLevel: null, readOnly: false },
    parent: undefined,
    name: undefined,
    id: '0c535f38-33e0-42e7-98e2-8d7b557f6054',
    connection: Client {
      ...
    }
  },
  where: { id: [ 2274 ] },
  validate: true,
  hooks: true,
  individualHooks: false,
  returning: false,
  force: false,
  sideEffects: true,
  type: 'BULKUPDATE',
  fields: [ 'assetId', 'updatedAt', 'asset_type' ],
  model: Permission,
  skip: [ 'id', 'flag', 'assetType', 'targetType', 'targetId', 'createdAt' ]
}
```

A simple (and verified) fix would be to revert the changes in query-interface.js.

```
    const options = Utils.cloneDeep(optionsArg);
    options.type = QueryTypes.SELECT;
    options.model = model;
    // const options = { ...optionsArg, type: QueryTypes.SELECT, model }; // This is the current lint
```

A more complicated to implement, but what might be a safer solution, would be to update the entire library, to not modify the parameters passed to it, as is often the case (with options and values parameters, for example).

If there is any more information needed, I'd be glad to help. I'm working on writing a test for us, that will verify the sequelize version we are using won't be affected by this issue. Once done, hopefully by early next week, I'll be more than happy to port it over to this library.

### Environment

- Sequelize version: >=5.13.0 
- Node.js version: 13.14.0 
- Operating System: MacOS
- If TypeScript related: TypeScript version: 4.1.2

roikoren755:
### Pull Request check-list

### Description of change

Closes #12868.


joni7777:
👍

papb (approved):

This is a very high quality PR! Congratulations. Excellent work. Thank you so much 🚀 🚀

Issue #13066, opened 2021-03-02
BulkCreate: updateOnDuplicate option is not resolve the ON CONFLICT unique key correctly


wl1711:
### What are you doing?

I am trying to create entities in bulk with bulkCreate. In case of a conflict on the primary key, I want that entity to be updated with the new values I have given.

Model:
```
module.exports = (sequelize, DataTypes) => {
  const Capability = sequelize.define(
    'Capability', {
      id: {
        type: DataTypes.INTEGER,
        field: 'id',
        allowNull: false,
        primaryKey: true,
        autoIncrement: true
      },
      name: {
        type: DataTypes.STRING(255),
        field: 'name',
        allowNull: false
      },
      description: {
        type: DataTypes.TEXT,
        field: 'description',
        allowNull: true
      },
      categoryId: {
        type: DataTypes.INTEGER,
        field: 'categoryId',
        allowNull: false,
        defaultValue: 1,
        references: {
          model: 'Category',
          key: 'id'
        },
        onUpdate: 'NO ACTION',
        onDelete: 'NO ACTION'
      },
      organizationId: {
        type: DataTypes.INTEGER,
        field: 'organizationId',
        allowNull: false,
        defaultValue: 18,
        references: {
          model: 'Organization',
          key: 'id'
        },
        onUpdate: 'NO ACTION',
        onDelete: 'NO ACTION'
      },
      deleted: {
        type: DataTypes.BOOLEAN,
        field: 'deleted',
        allowNull: false,
        defaultValue: false
      },
      updatedAt: {
        type: DataTypes.DATE,
        field: 'updatedAt',
        allowNull: false
      },
      updatedById: {
        type: DataTypes.INTEGER,
        field: 'updatedById',
        allowNull: false,
        references: {
          model: 'User',
          key: 'id'
        },
        onUpdate: 'NO ACTION',
        onDelete: 'NO ACTION'
      },
      createdAt: {
        type: DataTypes.DATE,
        field: 'createdAt',
        allowNull: false
      },
      createdById: {
        type: DataTypes.INTEGER,
        field: 'createdById',
        allowNull: false,
        references: {
          model: 'User',
          key: 'id'
        },
        onUpdate: 'NO ACTION',
        onDelete: 'NO ACTION'
      },
      capabilityStatusId: {
        type: DataTypes.INTEGER,
        field: 'capabilityStatusId',
        allowNull: false,
        defaultValue: 1,
        references: {
          model: 'CapabilityStatus',
[...]

Issue #13031, opened 2021-02-12
Upsert where condition matching


IlyaDonskikh:
## Issue Description
When i create an index with where condition it does not match ON CONFLICT statement on upsert.

### What are you doing?

PSQL Index: 
```sql
CREATE UNIQUE INDEX ON "lists" ("sample", "example") WHERE ("deletedAt" IS NULL);
```

Model Index:
```js
indexes: [
  {
    unique: true,
    fields: ['sample', 'example'],
    where: {
      deletedAt: null,
    },
  },
]
```

Request:
```js
 await List.upsert(attributes, { transaction });
```

### What do you expect to happen?

It have to INSERT new record or UPDATE existing one based on uniq index condition.

```sql
INSERT INTO "lists" ("sample","example","createdAt","updatedAt") VALUES ($1,$2,$3,$4) 
  ON CONFLICT ("sample","example") WHERE deletedAt is NULL DO...
```

### What is actually happening?

Error:
> original: error: there is no unique or exclusion constraint matching the ON CONFLICT specification

```
INSERT INTO "lists" ("sample","example","createdAt","updatedAt") VALUES ($1,$2,$3,$4) 
  ON CONFLICT ("sample","example") DO UPDATE SET...
```

### Additional context

Seems like `ON CONFLICT WHERE` case does not covered. At least according to following PR [code](https://github.com/sequelize/sequelize/pull/12301/files#diff-185469cea2079250f6cac5b57384909522b66a45fbd6c1180169f89b5ac0befaR188).

So, instead of mapping where condition into ON CONFLICT statement, the builder just ignore it.

### Environment

- Sequelize version: 6.3.5 
- Node.js version: v12.16.1
- Operating System: MacOs 10.15.7

### How does this problem relate to dialects?

- [ ] I think this problem happens regardless of the dialect.
- [x] I think this problem happens only for the following dialect(s): Psql
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [x] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

deggertsen:
[...]

Issue #12774, opened 2020-10-20
bulkCreate with unique partial index in postgres


ramezhanna:
## Issue Description
 model bulkCreate function with updateOnDuplicate don't support partial unique index for postgres: 
`[SequelizeDatabaseError]: there is no unique or exclusion constraint matching the ON CONFLICT specification`

### What are you doing?
While trying to use bulkCreate function with unique partial index in postgres i could manage to use the model and create the query to bulkCreate 

**postgres partial unique index**
```sql
create unique index table_partial_idx
    on table (key1, key2, key3)
    where (key3 not in (3,4));
```

**Model Defination**

```js
const Table = sequelize.define('Table', {
    id: {
      allowNull: false,
      type: DataTypes.STRING,
      primaryKey: true,
      defaultValue: DataTypes.UUIDV4,
    },
    key1: {
      type: DataTypes.STRING,
      unique: 'table_partial_idx',
    },
    key2: {
      type: DataTypes.STRING,
      unique: 'table_partial_idx',
    },
    key3: {
      type: DataTypes.INTEGER,
      unique: 'table_partial_idx',
    },
    key4: {
      type: DataTypes.DATE,
      defaultValue: DataTypes.NOW,
    },
    key5: {
      type: DataTypes.STRING,
    },
  }, {
    indexes: [{
      name: 'table_partial_idx',
      unique: true,
      fields: ['key1', 'key2', 'key3'],
      where: {
        status: {
          [Op.notin]: [3,4],
        },
      },
    }],
  });
```

**Model usage**
```js
await sequelize[entityName].bulkCreate([...data] {
        updateOnDuplicate: ['key4', 'key5'],
        returning: true,
      }
```

### What do you expect to happen?
Expected to generate a insert query with `On Conflict` key and a where condition
**Expected query** 
```sql
insert
	into
	"alarms" ("id", "key1", "key2", "key3", "key4", "key5")
values (uuid_generate_v4(), 'text', 'text', 1, '2020-10-20 12:01:32.143 +00:00', 'test')
 ON CONFLICT (key1, key2, key3)  where (key3 not in (3,4))
DO UPDATE SET key4 = EXCLUDED.key4 , key5 = EXCLUDED.key5;
```
### What is actually happening?
generate a insert query with `On Conflict` key but without a where condition that leaded to 

` DatabaseError [SequelizeDatabaseError]: there is no unique or exclusion constraint matching the ON CONFLICT specification`

**output query**
```sql
insert
	into
[...]

Issue #12742, opened 2020-10-07
bulkCreate updateOnDuplicate ignores unique keys


oxyii:
## Issue Description

(ref #11569)

bulkCreate `updateOnDuplicate` bug if unique composite index declared in model options instead of field options.

### What are you doing?

```js
const dbInstance = require('./db');

const Model1 = dbInstance.define('Model1', {
  field1: DataTypes.STRING,
  field2: DataTypes.STRING,
}, {
  indexes: [
    {
      name: 'compositeUnique',
      fields: ['field1', 'field2'],
      unique: true,
    },
  ],
})

const Model2 = dbInstance.define('Model2', {
  field1: {
    type: DataTypes.STRING,
    unique: 'compositeUnique',
  },
  field2: {
    type: DataTypes.STRING,
    unique: 'compositeUnique',
  },
})
```

### What do you expect to happen?

The same behavior in both cases

### What is actually happening?

```js
// get Model1.uniqueKeys
{}
// get Model2.uniqueKeys
{
  "compositeUnique": { /* fields, name.... */ },
}
```

because of this `Insert` request with `ON CONFLICT` fragment use `primaryKey` only in first case (Model1)

### Additional context

nope

### Environment

- Sequelize version: **6.3.5**
- Node.js version: **12.18.4**
- Operating System: tested on **Ubuntu 18.04, 20.04** and **MacOS 10.15**
- If TypeScript related: TypeScript version: **irrelevant**, tried on TS 4.0.3 and pure js

### How does this problem relate to dialects?

- [x] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [x] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

kreinholt:
Some relevant pull requests that might solve this are: https://github.com/sequelize/sequelize/pull/12516 and https://github.com/sequelize/sequelize/pull/12547

PR #13270, opened 2021-05-11
feat(postgres, sqlite): add support for where clauses with ON CONFLICT queries

### LINKED ISSUE #13240: Upsert does not always find/handle unique index
### What are you doing?

```js
class User extends Model {
}
User.init({
    name: DataTypes.STRING,
    someVal: DataTypes.INTEGER,
}, {
   indexes: [
       fields: [fn('LOWER', 'name')],
       unique: true,
       where: { someVal: 123 },
       name: '<some-name>',
   ],
})

const [{ id }] = await User.upsert({
    name: 'abc',
    someVal: 123,
}, {
    fields: ['name'],
    returning: ['id'],
});
const [{ id: id2 }] = await User.upsert({
    name: 'abc',
    someVal: 123,
}, {
    fields: ['name'],
    returning: ['id'],
});
console.log(id === id2) // true
```

In the above code, I would expect the User model to be upserted based on the partial unique index.

Sequelize fails to detect the unique index and upserts based on primary key. So Postgres rejects the insert due to unique constraint without explicit handling.

### Additional context

To my understanding, here are the general causes to this failure:

The first is that the index detection only looks for string literals for the field name https://github.com/sequelize/sequelize/blob/main/lib/dialects/abstract/query-interface.js#L789

So when an index uses functions, it cannot find the values (because `fn() !== fn()`)

The second is that even if the unique-index selection logic found util methods, it would still fail to stringify the values (currently it only expects string literals).
https://github.com/sequelize/sequelize/blob/main/lib/dialects/abstract/query-generator.js#L186
(Will open PR for this fix)

And third is that _only_ the fields are used in declaring the conflict. An index with a `where` clause should also have that included.
See `index_predicate` here https://www.postgresql.org/docs/13/sql-insert.html

### Environment

- Sequelize version: 6.6.2
- Node.js version: 14.16.1
- Operating System: MacOS

- [x] I think this problem happens only for the following dialect(s): Postgres (which supports partial index with functions). May impact other languages with similar functionality

I will open a PR for the second part, but the first and third are a bit trickier.

For the first, I think the best solution is to provide the name of the index in the options, and map that to the found index. Note the generated SQL should still use `unique index inference` for the reason stated in Postgres docs, but those issues don't apply to Sequelize.

For the third, I think the insert query options needs to take a `where` option as well. And when present, stringify it and attach to the ON CONFLICT statement.

AllAwesome497:
Add support for where clauses used in the ON CONFLICT part of the queries generated by model.bulkCreate and model.upsert for usage >

### Pull Request check-list

### Description of change

This is similar to #11611 but a bit different.  Unique indexes partial on the records not being soft deleted (eg, `CREATE UNIQUE INDEX 'abc' ON 'table1' ("col1") WHERE time_deleted IS NULL`) are picked up automatically so long as:
1. All of the fields in the index are being specified in `conflictFields` (for upsert, `upsertFields` for bulkCreate, happy to change naming if needed) 
2. The index is in the model's definition.

`Model.upsert` has 2 more options:
- `conflictFields` - This is for specifying the keys which are added in `ON CONFLICT()`. (eg, `ON CONFLICT(${conflictFields})`).
- `conflictWhere` - This is used to specify a where clause for the `ON CONFLICT` part of the query. 

`Model.bulkCreate` has the same options but with different names (might be worth changing them to match):
- `upsertFields` - Same thing as `conflictFields`.
- `upsertWhere` - Same as `conflictWhere`. 
Note: bulkCreate will ignore these options if `updateOnDuplicate` isn't specified.

An example case of where this is handy:
```js
// Create a table for Users
// Users need unique usernames, but only if they haven't been deleted.
const Users = sequelize.define('users', {
    username: { type: DataTypes.CITEXT },
    email: { type: DataTypes.CITEXT },
    time_deleted: { type: DataTypes.DATE }, 
}, {
    paranoid: true,
    deletedAt: 'time_deleted',
    indexes: [{
        fields: ['username'],
        unique: true,
        where: { time_deleted: null },
    }],
});

await Users.upsert({
    username: 'User123',
    email: 'newEmail@abc.com' 
}, {
    conflictFields: ['username'],
});
```

Closes #13240, #13066, #13031, #12774, #12742, #12595, #11656

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -187,7 +187,7 @@ class QueryGenerator {
         const conflictKeys = options.upsertKeys.map(attr => this.quoteIdentifier(attr));
         const updateKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=EXCLUDED.${this.quoteIdentifier(attr)}`);
         const maybeWhere = this.whereQuery(conflictWhere);
-        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) ${maybeWhere || `${maybeWhere} `}DO UPDATE SET ${updateKeys.join(',')}`;
+        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) ${maybeWhere ? `${maybeWhere} ` : ''}DO UPDATE SET ${updateKeys.join(',')}`;
       } else {
         const valueKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=VALUES(${this.quoteIdentifier(attr)})`);
         onDuplicateKeyUpdate += `${this._dialect.supports.inserts.updateOnDuplicate} ${valueKeys.join(',')}`;
@@ -295,7 +295,7 @@ class QueryGenerator {
         const conflictKeys = options.upsertKeys.map(attr => this.quoteIdentifier(attr));
         const updateKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=EXCLUDED.${this.quoteIdentifier(attr)}`);
         const maybeWhere = this.whereQuery(conflictWhere);
-        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) ${maybeWhere || `${maybeWhere} `}DO UPDATE SET ${updateKeys.join(',')}`;
+        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) ${maybeWhere ? `${maybeWhere} ` : ''}DO UPDATE SET ${updateKeys.join(',')}`;
       } else { // mysql / maria
         const valueKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=VALUES(${this.quoteIdentifier(attr)})`);
         onDuplicateKeyUpdate = `${this._dialect.supports.inserts.updateOnDuplicate} ${valueKeys.join(',')}`;
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -187,7 +187,7 @@ class QueryGenerator {
         const conflictKeys = options.upsertKeys.map(attr => this.quoteIdentifier(attr));
         const updateKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=EXCLUDED.${this.quoteIdentifier(attr)}`);
         const maybeWhere = this.whereQuery(conflictWhere);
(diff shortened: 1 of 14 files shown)


revlucio:
This feature will help us out, we currently have a custom upsert function that does this that this would replace

mohanagy (dismissed):

lib/dialects/abstract/query-generator.js
```diff
@@ -181,11 +182,12 @@ class QueryGenerator {
...
 
     if (this._dialect.supports.inserts.updateOnDuplicate && options.updateOnDuplicate) {
-      if (this._dialect.supports.inserts.updateOnDuplicate == ' ON CONFLICT DO UPDATE SET') { // postgres / sqlite
+      if (this._dialect.supports.inserts.updateOnDuplicate === ' ON CONFLICT DO UPDATE SET') { // postgres / sqlite
         // If no conflict target columns were specified, use the primary key names from options.upsertKeys
         const conflictKeys = options.upsertKeys.map(attr => this.quoteIdentifier(attr));
         const updateKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=EXCLUDED.${this.quoteIdentifier(attr)}`);
-        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) DO UPDATE SET ${updateKeys.join(',')}`;
+        const maybeWhere = this.whereQuery(conflictWhere);
+        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) ${maybeWhere ? `${maybeWhere} ` : ''}DO UPDATE SET ${updateKeys.join(',')}`;
```
Keimeno, line 190: Just a small thing, but `${maybeWhere ? `${maybeWhere} ` : ''}` in L190 - L298 can be optimized to `${maybeWhere || ''}`.

Additionally, it might make even more sense to abstract the above lines into a function, since they share the same functionality as L295 - L298 (but this might already be out of scope of this PR)

AllAwesome497: Opted not to do this since I couldn't think of a good name for an abstract function + it doesn't make sense (imo) to make a function specifically for this case (and since this only used twice + isn't much logic).

papb:
Hello! Just to have it written here, as we discussed earlier on Slack: please split this PR in four parts, considering #13346 to be the first of them. The other three parts should be:

* Add a new `conflictWhere` option to `upsert`, similarly to #13346
* Add a new `conflictFields` option to `bulk-create`
* Add a new `conflictFields` option to `upsert`

Note: maybe you can join the last two in a single PR if you want.

AllAwesome497:
Sounds good, going to close this now and break it up. I'll follow up and mention the other PRs once I make them.

AllAwesome497:
> This feature will help us out, we currently have a custom upsert function that does this that this would replace

#13412 should take care of that.  I'm currently working on a similar pr for `bulkCreate`.

Issue #12583, opened 2020-08-02
Sequelize crash - mariaDB 10.5.4 & JSON field


usb248:
## Issue Description
![plantage](https://user-images.githubusercontent.com/7659927/89119973-90fd2d80-d4b2-11ea-9268-ab3b58f8ba40.png)
mariaDB 10.4 OK, but mariaDB 10.5.X crashes my app
![image](https://user-images.githubusercontent.com/7659927/89120135-1af9c600-d4b4-11ea-849d-37751384b383.png)

### What are you doing?

```js
    data: {
      type: db.Sequelize.JSON,
      allowNull: false
    },
```

```js
Store.findOne({ 
			where: { ... },
			limit: 1,
			attributes: [ 'data'], // is the json field
			include: [{
		        model: model2,
		        attributes: [...],
		        required: true,
		        include: [{
			     	model: model3,
			     	attributes: [...],
			     	required: true
			    }]
		    }]
		})
```

### What do you expect to happen?

no crash ?

### What is actually happening?

trying to `JSON.parse` on something wrong
https://github.com/sequelize/sequelize/blob/master/src/dialects/mariadb/query.js#L180-L185

### Environment

- Sequelize version: sequelize@6.3.3
- Node.js version: v12.18.3
- Operating System: debian 10
- If TypeScript related: TypeScript version: no ts

### How does this problem relate to dialects?

- [ ] I think this problem happens regardless of the dialect.
- [x] I think this problem happens only for the following dialect(s): mariadb
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

EDIT:
After debugging, i noticed that `row[modelField.fieldName]` is already an object, so obviously, if sequelize tries to json.parse an object it throws this error.... please fix this bug.

miladmeidanshahi:
same error
```
server error SyntaxError: Unexpected token o in JSON at position 1
    at JSON.parse (<anonymous>)
    at /opt/bazzino-backend/node_modules/sequelize/lib/dialects/mariadb/query.js:183:72
    at Array.map (<anonymous>)
    at Query.handleJsonSelectQuery (/opt/bazzino-backend/node_modules/sequelize/lib/dialects/mariadb/query.js:182:21)
    at Query.formatResults (/opt/bazzino-backend/node_modules/sequelize/lib/dialects/mariadb/query.js:127:12)
[...]

PR #13273, opened 2021-05-13
fix(mariadb): Process JSON fields only if actually requested (#13027 & #12583

### LINKED ISSUE #13027: Nothing returns when used attributes with JSON field in table on MariaDB dialect
### What are you doing?
Here is the link with the problem https://github.com/roychri/sequelize_debug_2021-02-11

Trying to query a table with JSON type field with attribute projection, but if I don't state JSON field in attribute it will return nothing.
Wanted to retrieve data with attribute projection.

Returned nothing

### Additional context
As described on git page I linked, if you remove the JSON field from model, it will works normally.
### Environment

- Sequelize version: 5.21.7 // Doesn't really matter
- Node.js version: 14.4.0
- Operating System: Windows

- [X] I think this problem happens only for the following dialect(s): MariaDB

- [X] No, I don't have the time and I wouldn't even know how to start.

das-peter:
Adding JSON fields to rows regardless if requested or not breaks AbstractQuery::_groupJoinData() for example.

Fixes #13027 & #12583

### DIFF
--- a/lib/dialects/mariadb/query.js
+++ b/lib/dialects/mariadb/query.js
@@ -183,11 +183,12 @@ class Query extends AbstractQuery {
       if (modelField.type instanceof DataTypes.JSON) {
         // Value is returned as String, not JSON
         rows = rows.map(row => {
-          row[modelField.fieldName] = row[modelField.fieldName] ? JSON.parse(
-            row[modelField.fieldName]) : null;
-          if (DataTypes.JSON.parse) {
-            return DataTypes.JSON.parse(modelField, this.sequelize.options,
-              row[modelField.fieldName]);
+          if(row[modelField.fieldName] !== undefined) {
+            row[modelField.fieldName] = JSON.parse(row[modelField.fieldName]);
+            if (DataTypes.JSON.parse) {
+              return DataTypes.JSON.parse(modelField, this.sequelize.options,
+                row[modelField.fieldName]);
+            }
           }
           return row;
         });


das-peter:
Cross linking issue that sent me on this debugging journey https://github.com/ForestAdmin/forest-express-sequelize/issues/723

AllAwesome497 (changes requested):
Please update this pull request to match the template and / or match [contributing guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md#pull-requests). 

I'm not sure what the issue is.

das-peter:
@AllAwesome497 Thanks for the feedback. Updated the description. 
Management summary:
MariaDB JSON field processing always adds JSON fields to result-set, even if fields weren't requested.
This breaks stuff that's working based on result-set structure.
Solution: Only process JSON field if it was requested  / part of the result-set.

AllAwesome497:
> @AllAwesome497 Thanks for the feedback. Updated the description.
> Management summary:
> MariaDB JSON field processing always adds JSON fields to result-set, even if fields weren't requested.
> This breaks stuff that's working based on result-set structure.
> Solution: Only process JSON field if it was requested / part of the result-set.

I'm still not sure what the issue is, can you give more info like:
- When / why this would be an issue
- An example of the query that you see this issue in
- What your result was
- What the result you expected is
- An example model where this issue comes up in.

das-peter:
I sincerely am not sure if this now all about satisfying bureaucracy :|

However, here's a minimal example based on pseudo code:
Schema:
```
  const attributes = {
    id: {
      type: DataTypes.INTEGER,
      primaryKey: true,
      autoIncrement: true
    },
    whatever: {
      type: DataTypes.STRING,
      allowNull: false
    },
    evilJSONField: {
      type: DataTypes.JSON,
      allowNull: false,
      parse: true
    }
  };
  sequelize.define('Demos', attributes, { timestamps: true });
```

Query:
```
  await db.demo.sync({ alter: true });
  await db.demo.truncate({ logging: console.log });
  await db.demo.create({ whatever: 'foo', evilJSONField: JSON.stringify(['random', 'data']) });
  const records = await db.demo.findAll({ attributes: ['id', 'whatever'], raw: true });
  console.log(records);
```
This would trigger following select query:  ` SELECT id, whatever FROM Demos AS Demos;`

And it will return:
```
[ { id: 1, whatever: 'foo', evilJSONField: null } ]
```

See that `evilJSONField`? That's bad. Because it was not requested, it shouldn't be there. 
And it's presence doesn't just violate the expectation you'd have from the generated query or the promise of `raw: true` to show only what the query returns, but that unexpected field will break the processing in  `AbstractQuery::_groupJoinData()` and probably other locations.

AllAwesome497 (approved):
I edited your comment to use [syntax highlighting](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting).  Thanks for adding code samples + the clarification.


Looks fine once linting is fixed.  Can you also add a test for this to prevent regression?

lib/dialects/mariadb/query.js
```diff
@@ -183,11 +183,12 @@ class Query extends AbstractQuery {
       if (modelField.type instanceof DataTypes.JSON) {
         // Value is returned as String, not JSON
         rows = rows.map(row => {
-          row[modelField.fieldName] = row[modelField.fieldName] ? JSON.parse(
-            row[modelField.fieldName]) : null;
-          if (DataTypes.JSON.parse) {
-            return DataTypes.JSON.parse(modelField, this.sequelize.options,
-              row[modelField.fieldName]);
+          if(row[modelField.fieldName] !== undefined) {
```
AllAwesome497, line 186: oh also: it looks like this doesn't fix #13027.

for that you should probably do something like:
```suggestion
          if(typeof row[modelField.fieldName] === 'string') {
```

Issue #13276, opened 2021-05-18
DataType.TIME doesn't insert into Postgres correctly


sebastiankade:
### What are you doing?
Upgrading to Sequelize 6 from 4.x.

We use DataType.TIME for TIMESTAMP postgres fields.

```
@AutoColumn({ type: DataType.TIME })
  practiceTime: Maybe<Date>;
```

However when this is getting inserted into sql it is being inserted as a date:
```invalid input syntax for type time: "2021-05-18T08:00:00.000+00:00"```

Time syntax should be: `HH:MM:SS`

### What do you expect to happen?

DataType.TIME should insert fields in `HH:MM:SS` format

### What is actually happening?

Trying to insert as full date strings.

### Environment

Sequelize Versions:
"sequelize": "^6.6.2",
"sequelize-cli": "^6.2.0",
"sequelize-typescript": "^2.1.0",

- Node.js version: v12.16.1 
- Operating System: Mac
- If TypeScript related: TypeScript version: 3.7.5

### How does this problem relate to dialects?

- [ ] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [x] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [x] No, I don't have the time and I wouldn't even know how to start.

PR #13281, opened 2021-05-20
feat(time): add Date data type support to time field  When a date is provided as input for time field convert it to time string 

jemshid123:
### Pull Request check-list

### Description of change

This PR will enable passing Date objects to fields with 'TIME' datatype. Earlier passing Date object to time field would have generated an error, now the time will automatically be extracted from Date object.

```js
const user = sequelize.define('users',{
    name:{
        type: Sequelize.DataTypes.STRING,
        allowNull: true
    },
    test_time:{
        type: Sequelize.DataTypes.TIME,
        allowNull: true
    }
})

sequelize.sync();

user.create({
    name:"Test user",
    test_time:new Date('2020-05-18T09:00:00.000+00:00')
})
```

Now if you have a date object then it can be passed directly to a time field.

This PR also fixes a problem similar to following issue [issue#13276](https://github.com/sequelize/sequelize/issues/13276)


lib/data-types.js
```diff
@@ -417,6 +417,21 @@ class TIME extends ABSTRACT {
...
+    return date;
+  }
+  _sanitize(value, options) {
+    if ((!options || options && !options.raw) && value instanceof Date ) {
+      return moment(value).format('HH:mm:ss');
+    }
+    return value;
+  }
+  _isChanged(value, originalValue) {
+    return value !== originalValue;
```
AllAwesome497, line 433: this is going to break if `value` and `originalValue` are both instances of the JS `Date` class, since this compares the references not the values.

eg, `new Date() !== new Date()`. 

`_isChanged` in `DATE` has logic to take care of this, you could probably copy and paste that.

test/unit/sql/insert.test.js
```diff
@@ -124,6 +124,32 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
     });
   });
 
+  describe('time', () => {
+    it('formats time datatype correctly when inserting', () => {
```
AllAwesome497, line 128: 2 issues here:
1. it doesn't make sense to have `describe` for a single test (/ `it`).
2. The change doesn't belong to `insert`, so this shouldn't be a unit test there.  Instead it should be a unit test on data-types (file: [here](https://github.com/sequelize/sequelize/blob/main/test/unit/sql/data-types.test.js)) and (optionally) in integration tests for the places that implement it.

AllAwesome497 (changes requested):
I edited your pr to use [syntax highlighting](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting). (edit: also unchecked the commit message conventions box)

probably too late to fix this now, but if / when you contribute next, please don't skip verification and format the commit message as per our [contributing guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md#5-commit).

edit: Also, please change the name of the pr to match this.

lib/data-types.js
jemshid123: I have changed the PR  name to match the convention, also has updated `__isChanged` function with code from `DATE` class. Hope the rest looks good.

test/unit/sql/insert.test.js
jemshid123: I have removed the test from `insert`  and added test in `data-types.test.json ` .  Hope this is okay.

test/unit/sql/data-types.test.js
```diff
@@ -203,6 +203,22 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
...
+            type.validate('foobar');
+          }).to.throw(Sequelize.ValidationError, '"foobar" is not a valid date');
+        });
+
+        it('should return `true` if `value` is a date', () => {
+          const type = DataTypes.DATE();
+
+          expect(type.validate(new Date())).to.equal(true);
+        });
+      });
```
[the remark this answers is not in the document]
jemshid123: A separate `describe`  block has been added for `TIME` data type.

sebastiankade:
@AllAwesome497 just bumping incase you didn't see, I think this needs an approval from a previous contributor...

test/unit/sql/data-types.test.js
```diff
@@ -205,6 +205,40 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
...
+    describe('TIME', () => {
+      testsql('TIME', DataTypes.TIME, {
+        postgres: 'TIME',
+        mssql: 'TIME',
+        mariadb: 'TIME',
+        mysql: 'TIME',
+        sqlite: 'TIME'
+      });
+
+      describe('validate time', () => {
```
AllAwesome497, line 217:
```suggestion
      describe('validate', () => {
```

jemshid123: The change has been made.

lib/data-types.js
```diff
@@ -417,6 +417,36 @@ class TIME extends ABSTRACT {
...
+    return true;
+  }
+  _stringify(date) {
+    if (date instanceof Date) {
+      return moment(date).format('HH:mm:ss');
+    } 
+    return date;
+  }
+  _sanitize(value, options) {
+    if ((!options || options && !options.raw) && value instanceof Date ) {
```
AllAwesome497, line 433:
```suggestion
    if ((!options || !options.raw) && value instanceof Date ) {
```

lib/data-types.js
```diff
@@ -417,6 +417,36 @@ class TIME extends ABSTRACT {
...
+      return moment(value).format('HH:mm:ss');
+    }
+    return value;
+  }
+  _isChanged(value, originalValue) {
+    if (originalValue && !!value &&
+      (value === originalValue ||
+        value instanceof Date && originalValue instanceof Date && value.getTime() === originalValue.getTime())) {
```
AllAwesome497, line 441: a note here - if the date is truncated from the original value this could be inaccurate.

Eg, if the original value was `5:23:53.234` the value from the db would be `5:23:53` so it'd be different. 

Not sure if this is a big issue though.

AllAwesome497 (dismissed):
a few minor nits but looks fine.

lib/data-types.js
jemshid123: Going through docs for the supported databases only mssql supports the millisecond all other time formats only allow up to the second so doesn't seem like a big issue.

lib/data-types.js
jemshid123: Has removed this unnecessary check.

lib/data-types.js
AllAwesome497: dunno, still think its worth checking just in case. Should just be `Math.floor(value.getTime() / 1000) === Math.floor(otherValue.getTime() / 1000)`

jemshid123: I have added the checks. So now the milliseconds will be ignored while comparing.

sebastiankade: @AllAwesome497 soft bump :)

AllAwesome497 (approved):
Looks good to me.  Would be nice to have a test for ms truncation but it's not too big of a deal.

sebastiankade:
@AllAwesome497 looks like all comments have been addressed besides the non-mandatory ms-truncation test. Can it be merged in?

AllAwesome497:
> @AllAwesome497 looks like all comments have been addressed besides the non-mandatory ms-truncation test. Can it be merged in?

I can't merge, only approve.  CC @papb

sebastiankade:
@papb @AllAwesome497 @jemshid123 Did this end up getting in or just going stale?

WikiRik:
@sebastiankade thanks for the message! It seems that this was just going stale and was accidentally closed in the few days that the bot also closed stale PRs. I will reopen this and see what needs to be done to merge it in the end

lib/data-types.js
```diff
@@ -460,6 +460,42 @@ class TIME extends ABSTRACT {
   toSql() {
     return 'TIME';
   }
+  validate(value) {
+    if (!Validator.isDate(String(value)) && !Validator.isDate(`01-01-1970 ${String(value)}`)) {
```
ephys, line 464:
```typescript
    if (value instanceof Date) {
      if (Number.isNaN(value.getTime())) {
        throw new Error()
      }
    }
    
    if (typeof value !== 'string) {
      throw new Error()
    }
    
    if (!/-?\d+:[0-5][0-9]:[0-5][0-9](?:\.\d+)?/.test(value)) {
      throw new Error()
    }

    return true;
```

While I'm fine with supporting `Date`, passing a date as a string should not be valid.

We should only either pass a `Date` object or a string that matches the format 
`-?\d+:[0-5][0-9]:[0-5][0-9](?:\.\d+)?`.

``Validator.isDate(`01-01-1970 ${String(value)}`)`` is also not the right way to check if a value is TIME as TIME can be greater than `23:59:59` and less than `00:00:00` when used as a duration.

lib/data-types.js
```diff
@@ -460,6 +460,42 @@ class TIME extends ABSTRACT {
...
+      throw new sequelizeErrors.ValidationError(util.format('%j is not a valid date', value));
```
ephys, line 465:
```suggestion
      throw new sequelizeErrors.ValidationError(util.format('%j is not a valid time', value));
```

ephys (changes requested):
I would personally recommend against using Date to store just the time. But it's the individual developer's responsibility to be careful when using Date so that's no reason to block it.

lib/data-types.js
```diff
@@ -460,6 +460,42 @@ class TIME extends ABSTRACT {
...
+    }
+    return true;
+  }
+  _stringify(date) {
+    if (date instanceof Date) {
+      return moment(date).format('HH:mm:ss');
```
ephys, line 471: We should include milliseconds in this

lib/data-types.js
```diff
@@ -460,6 +460,42 @@ class TIME extends ABSTRACT {
...
+    } 
+    return date;
+  }
+  _sanitize(value, options) {
+    if ((!options || !options.raw) && value instanceof Date ) {
+      return moment(value).format('HH:mm:ss');
```
ephys, line 477: We should include milliseconds in this

lib/data-types.js
```diff
@@ -460,6 +460,42 @@ class TIME extends ABSTRACT {
...
+    return value;
+  }
+  _isChanged(value, originalValue) {
+    if (
+      originalValue && 
+      !!value &&
+        (
+          value === originalValue ||
+          value instanceof Date && originalValue instanceof Date && 
+          Math.floor(value.getTime() / 1000) === Math.floor(originalValue.getTime() / 1000)
```
ephys, line 488: This should not exclude milliseconds, TIME supports milliseconds

ephys (changes requested):

lib/data-types.js
WikiRik: I think milliseconds support is dependent on the dialect, but I will take a look into that

ephys: That's a good point, thank you for double-checking :)

lib/data-types.js
jemshid123: These changes have been made and the validator now accepts only dates object and time strings.

lib/data-types.js
jemshid123: Done

lib/data-types.js
jemshid123: Done

lib/data-types.js
jemshid123: I researched this issue and it seems like Mysql, Mariadb, SQLite, Postgres and MSSQL support HH:MM:SS.sss formatted times. So I have changed the code to save milliseconds also in case a date object is passed as the argument.

jemshid123:
Looks like 1 test corresponding to DB2 is failing. But it doesn't look like it's related to my changes. @WikiRik  @ephys

ephys:
It's not :) You can ignore it

lib/data-types.js
```diff
@@ -460,6 +460,59 @@ class TIME extends ABSTRACT {
...
+  _sanitize(value, options) {
+    if ((!options || !options.raw) && value instanceof Date) {
+      return moment(value).format('HH:mm:ss.SSS');
+    }
+
+    return value;
+  }
+
+  _isChanged(value, originalValue) {
+    if (
```
ephys, line 497: For this one, I think we should first normalize `value` & `originalValue` to strings if they're Date objects. This way `_isChanged(new Date('2022-01-15T12:00:00Z'), '12:00:00.000')` will return `false`

ephys (changes requested):

WikiRik:
Thanks for the update @jemshid123
I was checking docs yesterday as well for this. I found that some dialects only support from 00:00:00 to 23:59:59 (or 24:00:00) and others go from -859:59:59.9999 to 859:59:59.9999 so I think we'll have to do some additional things there as well. 

The list below are the docs I found for the various dialects we're supporting in our CI;
- MariaDB 10.3, 10.5; https://mariadb.com/kb/en/time/
- MySQL 5.7; https://dev.mysql.com/doc/refman/5.7/en/time.html
- MySQL 8.0; https://dev.mysql.com/doc/refman/8.0/en/time.html (only difference is numbering of two referenced sections)
- Postgres 9.5; https://www.postgresql.org/docs/9.5/datatype-datetime.html
- Postgres 10; https://www.postgresql.org/docs/10/datatype-datetime.html (seems to have the same contents)
- MSSQL 2017, 2019; https://docs.microsoft.com/en-us/sql/t-sql/data-types/time-transact-sql?view=sql-server-ver15
- Snowflake; https://docs.snowflake.com/en/sql-reference/data-types-datetime.html#time
- SQLite; https://www.sqlite.org/lang_datefunc.html
- DB2 11.5; https://www.ibm.com/docs/en/db2/11.5?topic=functions-time

ephys:
If I sum up, I think we end up with this support table:

| dbms         | supported | microsecond | configurable ms precision | allow negative | allow > 24:00:00 | supports timezone | throws if input precision too high |
|--------------|-----------|-------------|---------------------------|----------------|------------------|-------------------|---------------------------------|
| MariaDB 10.5 | ✅         | ✅           | 0-6                       | ✅              | ✅                | ❌                 | ?                               |
| MySQL 8.0    | ✅         | ✅           | 0-6                       | ✅              | ✅                | ❌                 | ?                               |
| Postgres 10  | ✅         | ✅           | 0-6                       | ❌              | ❌                | ✅                 | ?                               |
| MSSQL 2019   | ✅         | ✅           | 0-7                       | ❌              | ❌                | ❌                 | ?                               |
| Snowflake    | ✅         | ✅           | 0-9                       | ❌              | ❌                | ❌                 | ?                               |
| SQLite       | ❌         | ❌           | ❌                         | ❌              | ❌                | ❌                 | ?                               |
| DB2          | ✅         | ❌           | ❌                         | ❌              | ❌                | ❌                 | ?                               |

lib/data-types.js
jemshid123: So we should convert the date object to `HH:mm:ss.SSS` format string and then compare them right?

ephys: Yes, I think we can call stringify on the inputs :)
We should also make sure `12:00:00` and `12:00:00.000` are equal

WikiRik:
@jemshid123 could you look into the remarks by @ephys?

jemshid123:
> @jemshid123 could you look into the remarks by @ephys?

Sure, will test out those scenarios and update here.  Have been rather busy recently, didn't get any time to check these out.

ephys:
The more I think about this PR the more I think we should not accept it. `Date` is not a good way to represent a plain time, and we should not encourage doing this by supporting it

sebastiankade:
@ephys how do you suppose you use Sequelize with postgres `time` fields then? 

https://www.postgresql.org/docs/current/datatype-datetime.html

ephys:
Yes, that's already how we represent time.
When temporal comes out, we'll add support for Temporal.PlainTime

PR #13303, opened 2021-06-07
Fix: Incorrect typescript typings for findAndCountAll function

### LINKED ISSUE #13302: Incorrect typescript typings for findAndCountAll function
### What are you doing?

I am trying to strongly type my application using Typescript and Sequelize. I have noticed the following behaviour of the `findAndCountAll` function. 
1. When no group by is supplied it returns `Promise<{ rows: M[]; count: number }>`.
2. When group by is supplied it returns `Promise<{ rows: M[]; count: { count: number }[]>>`

The current type of Sequelize `findAndCountAll` is 
```js
  public static findAndCountAll<M extends Model>(
    this: ModelStatic<M>,
    options?: FindAndCountOptions<M['_attributes']>
  ): Promise<{ rows: M[]; count: number }>;
```

Since `findAndCountAll` returns two different results the type should ideally be 

```js
  public static findAndCountAll<M extends Model>(
    this: ModelStatic<M>,
    options?: FindAndCountOptions<M['_attributes']>
  ): Promise<{ rows: M[]; count: number | { count: number }[]>;
```
I have noticed other attributes also being returned inside the count object so just a `count :  { count: number }` might not work.

As discussed in [this issue](https://github.com/sequelize/sequelize/issues/6148) to return the proper count I need to take a count of the returned count array. But since the type says count is a number I am unable to do so.

```js
    const allUsersRaw = await this.User.findAndCountAll({
        attributes,
        include,
        where,
        order,
        limit,
        offset,
        group: ["Order.id"],
        subQuery: false
    });
    const count = allUsersRaw.count.length;
```

Following error is thrown

```
Property 'length' does not exist on type 'number'. ts(2339)
 ```

### Environment

- Sequelize version: 6.6.2 
- Node.js version: v16.3.0 
- Operating System: MacOs BigSur 11.4
- If TypeScript related: TypeScript version: 4.3.2

### PR DESCRIPTION
**Pull request by Patil2099**
Fixes #13302

### CODE REVIEW

**Comment by Keimeno** [comment]
Thanks for your PR!

Can you please add additional tests? These will make sure that the new type works as expected, even in edge cases.
If you require more complex type assertions, you can use [expectTypeOf](https://www.npmjs.com/package/expect-type)

File: types/lib/model.d.ts [hunk 1]
```diff
@@ -1914,10 +1914,14 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
    * without
    * profiles will be counted
    */
-  public static findAndCountAll<M extends Model>(
+   public static findAndCountAll<M extends Model>(
```
Keimeno, line 1917: Please remove the added space here

Keimeno:
Closed in favour of #13477

PR #13334, opened 2021-06-21
fix(typings): makes Transactionable compatible with TransactionOptions

knoid:
### Pull Request check-list

### Description of change

This change makes `Transactionable` compatible with `TransactionOptions`. Having both compatible allows for much cleaner code like this:

```javascript
async afterDestroy(instance: Model, options: InstanceDestroyOptions) {
  return sequelize.transaction(options, async (transaction) => {
```

This piece of code was working before but broke on #13093. We can either merge this change or revert that PR to make these interfaces compatible but I think allowing `null` (merging this PR) is the way to go because it better represents the reality of this library.

### DIFF
--- a/types/lib/transaction.d.ts
+++ b/types/lib/transaction.d.ts
@@ -149,7 +149,7 @@ export interface TransactionOptions extends Logging {
   /**
    * Parent transaction.
    */
-  transaction?: Transaction;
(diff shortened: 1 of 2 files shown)


papb (approved):
Thanks! Good work. I made a small change to the test to make it a little more clear (from my point of view).

PR #12516, opened 2020-07-16
fix(model.js): updateOnDuplicate for unique index

brianschardt:
Currently updateOnDuplicate for unique index and indexes do not work. This PR solves this issue.

PR #13345, opened 2021-06-24
fix(bulk-create): `ON CONFLICT` with unique index

papb:
Supersedes #12516

Also addresses https://github.com/sequelize/sequelize/pull/11984/files#r414863957 via refactor

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2687,7 +2687,7 @@ class Model {
           const upsertKeys = [];
 
           for (const i of model._indexes) {
-            if (i.unique) {
+            if (i.unique && !i.where) { // Don't infer partial indexes
               upsertKeys.push(...i.fields);
             }
           }
--- a/lib/model.js
+++ b/lib/model.js
@@ -2683,11 +2683,22 @@ class Model {
         // Map updateOnDuplicate attributes to fields
         if (options.updateOnDuplicate) {
           options.updateOnDuplicate = options.updateOnDuplicate.map(attr => model.rawAttributes[attr].field || attr);
-          // Get primary keys for postgres to enable updateOnDuplicate
-          options.upsertKeys = _.chain(model.primaryKeys).values().map('field').value();
-          if (Object.keys(model.uniqueKeys).length > 0) {
-            options.upsertKeys = _.chain(model.uniqueKeys).values().filter(c => c.fields.length >= 1).map(c => c.fields).reduce(c => c[0]).value();
+
+          const upsertKeys = [];
+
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -2683,11 +2683,22 @@ class Model {
...
           options.updateOnDuplicate = options.updateOnDuplicate.map(attr => model.rawAttributes[attr].field || attr);
-          // Get primary keys for postgres to enable updateOnDuplicate
-          options.upsertKeys = _.chain(model.primaryKeys).values().map('field').value();
-          if (Object.keys(model.uniqueKeys).length > 0) {
-            options.upsertKeys = _.chain(model.uniqueKeys).values().filter(c => c.fields.length >= 1).map(c => c.fields).reduce(c => c[0]).value();
+
+          const upsertKeys = [];
+
+          for (const i of model._indexes) {
+            if (i.unique) {
```
AllAwesome497, line 2690: probably worth making sure that the index isn't a partial index - if its a partial index then it won't apply unless that where clause is specified, which currently we don't do.

lib/model.js
```diff
@@ -2683,11 +2683,22 @@ class Model {
...
+            }
           }
+
+          const firstUniqueKey = Object.values(model.uniqueKeys).find(c => c.fields.length >= 1);
+
+          if (firstUniqueKey && firstUniqueKey.fields) {
+            upsertKeys.push(...firstUniqueKey.fields);
+          }
+
+          options.upsertKeys = upsertKeys.length > 0 ? upsertKeys : Object.values(model.primaryKeys).map(x => x.field);
```
AllAwesome497, line 2701: small nit:
```suggestion
          options.upsertKeys = upsertKeys.length > 0
            ? upsertKeys
            : Object.values(model.primaryKeys).map(x => x.field);
```

lib/model.js
papb: Done at 9bc372a

lib/model.js
```diff
@@ -2683,11 +2683,24 @@ class Model {
...
+
+          const upsertKeys = [];
+
+          for (const i of model._indexes) {
+            if (i.unique && !i.where) { // Don't infer partial indexes
+              upsertKeys.push(...i.fields);
+            }
           }
+
+          const firstUniqueKey = Object.values(model.uniqueKeys).find(c => c.fields.length >= 1);
```
AllAwesome497, line 2695: personally I prefer `c.fields.length > 0` but not a big deal.

AllAwesome497 (approved):
Lgtm :+1:

PR #13346, opened 2021-06-24
feat(bulk-create): add `conflictWhere` option

papb:
Rework of #13155 adding documentation and TypeScript tests (and with less unrelated changes)

### DIFF
--- a/docs/manual/other-topics/upgrade-to-v6.md
+++ b/docs/manual/other-topics/upgrade-to-v6.md
@@ -74,7 +74,7 @@ Signature for this method has been changed to `Promise<Model,boolean | null>`. F
 - SQLite - Implemented with ON CONFLICT DO UPDATE
 - MSSQL - Implemented with MERGE statement
 
-_<ins>Note for Postgres users:</ins>_ If upsert payload contains PK field, then PK will be used as the conflict target. Otherwise first unique constraint will be selected as the conflict key.
+_<ins>Note for Postgres and SQLite users:</ins>_ If the upsert payload contains a primary key field, then it will be used as the conflict target. Otherwise the first unique constraint will be selected as the conflict target. Also, since v6.7.0, you can provide a where condition with the new `conflictWhere` option.
 
 ### QueryInterface
 
--- a/types/test/bulk-create.ts
+++ b/types/test/bulk-create.ts
@@ -1,0 +1,40 @@
+import { DataTypes, Sequelize } from 'sequelize';
+
+const sequelize = new Sequelize('mysql://user:user@localhost:3306/mydb');
+
+const User = sequelize.define('User', {
+  username: DataTypes.STRING,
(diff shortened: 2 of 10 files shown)


types/test/bulk-create.ts
```diff
@@ -0,0 +1,40 @@
+import { DataTypes, Sequelize } from 'sequelize';
```
AllAwesome497, line 1: small nit here: can you add some tests for invalid typing and use `// @ts-expect-error comment explaining why the typing is invalid`?

types/test/bulk-create.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const data = [
+  { uniqueName: 'Peter', secretValue: '12', intVal: 12 },
+  { uniqueName: 'Paul', secretValue: '23', intVal: 23 }
+];
+
+User.bulkCreate(data, {
+  fields: ['uniqueName', 'secretValue', 'intVal'],
+  updateOnDuplicate: ['secretValue', 'intVal']
+});
```
AllAwesome497, line 26: nit: this should probably have some sort check for the return type of this.

docs/manual/other-topics/upgrade-to-v6.md
```diff
@@ -74,7 +74,7 @@ Signature for this method has been changed to `Promise<Model,boolean | null>`. F
 - SQLite - Implemented with ON CONFLICT DO UPDATE
 - MSSQL - Implemented with MERGE statement
 
-_<ins>Note for Postgres users:</ins>_ If upsert payload contains PK field, then PK will be used as the conflict target. Otherwise first unique constraint will be selected as the conflict key.
+_<ins>Note for Postgres and SQLite users:</ins>_ If the upsert payload contains a primary key field, then it will be used as the conflict target. Otherwise the first unique constraint will be selected as the conflict target. Also, since v6.7.0, you can provide a where condition with the new `conflictWhere` option.
```
AllAwesome497, line 77: make sure that this is still accurate when you merge it.

AllAwesome497 (approved):
Looks good.  A few long lines that imo should be split into multiple but given as there isn't any enforcement of that e.g. prettier, not a big deal.

AllAwesome497 (dismissed):
Actually it looks like this PR also adds this functionality to `upsert`, there just isn't any tests or docs on it.  Can you remove that or add in tests? I'll throw up another PR for this if its removed.

AllAwesome497 (changes requested):
Well, just learned something! Apparently there's a difference in `ON CONFLICT WHERE (...) DO UPDATE ...` and `ON CONFLICT DO UPDATE ... WHERE`.

The first one will use indices, and can only target the current rows, e.g. `"isUnique" = true` not `"table"."isUnique" = true`, and the second one won't affect how indices are handled.

PR #13376, opened 2021-07-16
fix(sqlite): fix wrongly overwriting storage if empty string

### LINKED ISSUE #13375: Sequelize doesn't respect storage parameter for SQLite3 anonymous disk-based database
## Issue Creation Checklist

[ ] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description

It is impossible to create an Anonymous disk-based database using Sequelize.

### SSCCE

```
new Sequelize({dialect: 'sqlite', storage: ''})
```

The empty string in the `storage` parameter should have been passed to sqlite3, which [sqlite3 documentation](https://github.com/mapbox/node-sqlite3/wiki/API#new-sqlite3databasefilename-mode-callback) says ought to create an
> anonymous disk-based database
which will be automatically deleted when the handle is closed.

An in-memory SQLite database is created

### Additional context

Looking in the source code, the `storage` parameter is part of a chain of `||` (logic or operator)s. This is a common JS shorthand for picking the rightmost value that isn't falsy. Unfortunately an `''` (empty string) is also falsy, which causes Sequelize to keep going right and picking something else that isn't falsy.
For example, in [/lib/dialects/sqlite/connection-manager:48](https://github.com/sequelize/sequelize/blob/main/lib/dialects/sqlite/connection-manager.js#L48) we have `this.sequelize.options.storage || this.sequelize.options.host || ':memory:'`. Here if `storage` is empty string it will pick either `host` or `':memory:'`.

### Environment

- Sequelize version: 6.5.0 
- Node.js version: v14.15.0 

soryy708:
Empty string in SQLite3 means anonymous disk-based db, but sequelize overwrited it with :memory:

Closes #13375

### Pull Request check-list

### Description of change

Closes https://github.com/sequelize/sequelize/issues/13375

### DIFF
--- a/lib/dialects/sqlite/connection-manager.js
+++ b/lib/dialects/sqlite/connection-manager.js
@@ -45,7 +45,7 @@ class ConnectionManager extends AbstractConnectionManager {
   async getConnection(options) {
     options = options || {};
     options.uuid = options.uuid || 'default';
-    options.storage = this.sequelize.options.storage || this.sequelize.options.host || ':memory:';
+    options.storage = this.sequelize.options.storage ?? (this.sequelize.options.host || ':memory:');
     options.inMemory = options.storage === ':memory:' ? 1 : 0;
 
     const dialectOptions = this.sequelize.options.dialectOptions;
--- a/test/support.js
+++ b/test/support.js
@@ -113,7 +113,7 @@ const Support = {
       sequelizeOptions.native = true;
     }
 
-    if (config.storage) {
+    if (config.storage || config.storage === '') {
       sequelizeOptions.storage = config.storage;
     }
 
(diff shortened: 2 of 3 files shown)


soryy708:
I get 3 failing tests, but they happen when running `test-sqlite` on main without my changes.

sdepold:
So it's still a temporary database that gets deleted after stopping the app, correct? With the only difference that it is either memory or hard drive based.

sdepold:
Seems that all tests are green except for linking. Would you like to have a look and fix it?

soryy708:
> 
> 
> So it's still a temporary database that gets deleted after stopping the app, correct? With the only difference that it is either memory or hard drive based.

The way I understand it, yes. The use case for a temporary on-disk database is temporary workloads on a memory-constrained machine, like a CI server.

> Seems that all tests are green except for linking. Would you like to have a look and fix it?

I'm looking in to it. That's odd, because when I ran the linter locally it all passed.

soryy708:
@sdepold All checks passed now. Except 1 skipped check?

PR #13389, opened 2021-07-20
fix(types): add missing fields to FindOr{Create,Build}Options

### LINKED ISSUE #13388: FindOrCreateOptions does not extend CreateOptions
## Issue Creation Checklist

- [X]  I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

### SSCCE

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/172 

The `findOrCreate` method should accept `hooks: boolean` as an option.

While the `hooks` option does affect the functionality (`hooks: false` prevents the hooks from being called), the typescript interface `FindOrCreateOptions` does not include the `hooks: boolean` property, neither does it extend an interface that does.

### Additional context

I believe the `FindOrCreateOptions` interface should extend the `CreateOptions` interface just how it's extending the `FindOptions` interface.

### Environment

- Sequelize version: 6.5.0 
- Node.js version: 14.16.1 
- If TypeScript related: TypeScript version: 4.1.3 

devagrawal09:
Fixes #13388


FreTimmerman:
shouldnt it also extend `FindOptions` ?

see https://github.com/sequelize/sequelize/pull/11924

AllAwesome497 (dismissed):
Looks good, but can you add a test to `test/types` to prevent regressions?  Thanks.

ephys:
This PR has been untouched for a while, I've taken the liberty of adding the necessary test.

I've also split `FindOrCreateOptions` into `FindOrCreateOptions` and `FindOrBuildOptions` as `findOrBuild` used the same type even though it accepted different inputs.

WikiRik (approved):

PR #13411, opened 2021-08-03
feat(postgres, sqlite): add conflictWhere option to upsert

### LINKED ISSUE #13240: Upsert does not always find/handle unique index

### LINKED ISSUE #13412: ON CONFLICT WHERE for upserts.
## Issue Creation Checklist

## Feature Description

Currently sequelize doesn't support upserts with unique partial indexes.

There's many cases for unique partial indexes - such as a unique constraint on rows that haven't been soft deleted.

For example, a generic membership table with permissions:
```js
const Memberships = sequelize.define(
  'memberships',
  {
    user_id: { type: DataTypes.INTEGER, allowNull: false },
    foreign_key: { type: DataTypes.INTEGER, allowNull: false },
    permissions: { type: DataTypes.ENUM('admin', 'member', 'guest'), allowNull: false }
  },
  {
    deletedAt: 'time_deleted',
    indexes: [
      {
        fields: ['user_id', 'foreign_key'],
        unique: true,
        where: { time_deleted: null },
      },
    ],
  }
);

// Adds a membership for a user as an admin (updates previous membership instead if there is one)
await Memberships.upsert(
  {
    user_id: userId,
    foreign_id: foreignId,
    permissions: 'admin',
  },
  {
    conflictWhere: { time_deleted: null },
  }
);

```

There's several cases where this comes up:
- Unique usernames, except for deleted users.
- Membership models - something can't belong to the same item multiple times, however it can be fine for the association to be deleted then created later.
 

See the code sample above.

### Additional context

Part of #13240.

I also have a PR up at #13411, which is a piece of #13270 (closed)

- [ ] No. This feature is relevant to Sequelize as a whole.
- [x] Yes. This feature only applies to the following dialect(s): Postgres, Sqlite

#13411

AllAwesome497:
### Pull Request check-list

### Description of change

- Added the option `conflictWhere` to `QueryGenerator.prototype.insertQuery`.  This is used to generate the `ON CONFLICT WHERE` part of the query (note: not `ON CONFLICT DO UPDATE WHERE`).
- Added an option to `Model.upsert`, `conflictWhere`, This is passed to the query generator.
- Updated types
- Added tests  

Example use case:

```js
const Memberships = sequelize.define(
  'memberships',
  {
    user_id: { type: DataTypes.INTEGER, allowNull: false },
    foreign_key: { type: DataTypes.INTEGER, allowNull: false },
    permissions: { type: DataTypes.ENUM('admin', 'member', 'guest'), allowNull: false }
  },
  {
    deletedAt: 'time_deleted',
    indexes: [
      {
        fields: ['user_id', 'foreign_key'],
        unique: true,
        where: { time_deleted: null },
      },
    ],
  }
);

// Adds a membership for a user as an admin (updates previous membership instead if there is one)
await Memberships.upsert(
  {
    user_id: userId,
    foreign_id: foreignId,
    permissions: 'admin',
  },
  {
    conflictWhere: { time_deleted: null },
  }
);

```

Closes #13412.

### TODO:

Add some tests which validate `conflictWhere` & `conflictFields` work together properly.


test/integration/model/upsert.test.js
```diff
@@ -598,6 +598,168 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+                {
+                  unique: true,
+                  fields: ['name'],
+                  where: { isUnique: true }
+                }
+              ]
+            }
+          );
+
+          beforeEach(async () => Users.sync({ force: true }));
```
Keimeno, line 623: Just a small thing, but there's no need for the `async` keyword, if you're not awaiting the `Users.sync` call

lib/model.js
```diff
@@ -2426,6 +2426,7 @@ class Model {
    * @param  {Function}     [options.logging=false]                       A function that gets executed while running the query to log the sql.
    * @param  {boolean}      [options.benchmark=false]                     Pass query execution time in milliseconds as second argument to logging function (options.logging).
    * @param  {string}       [options.searchPath=DEFAULT]                  An optional parameter to specify the schema search_path (Postgres only)
+   * @param  {object}       [options.conflictWhere]                       An optional parameter that specifies a where clause for the `ON CONFLICT` part of the query (in the same format as the where option for a finder such as `findByPk`). Postgres/SQLite only. (note: updateOnDuplicate must be present for this to be used.)
```
Keimeno, line 2429: We should consider logging a warning if using `conflictWhere` without `updateOnDuplicate`, since this might be overseen by some users, even though this note is added.

Keimeno (dismissed):

test/integration/model/upsert.test.js
AllAwesome497: :man_facepalming: oops, forgot to remove - I originally did await it as I had originally defined the model in the beforeEach.

lib/model.js
AllAwesome497: :+1: worth considering.  It's worth noting that in the future we might use other methods after `ON CONFLICT` (such as `DO NOTHING`) which would still use this.

AllAwesome497: oh wait this is specifically for Model.upsert - nevermind. `updateOnDuplicate` is for `bulkCreate` upserts.

Keimeno (dismissed):

sdepold:
@AllAwesome497 What shall we do with this one?

AllAwesome497:
> @AllAwesome497 What shall we do with this one?

I'll need to resolve merge conflicts, but it should be fine. 

Now that #13723 is merged I think I'll also add a few more tests which test the combined behavi for some more specific use cases.

sdepold:
@AllAwesome497 any eta on when you resolve the conflicts?

AllAwesome497:
> @AllAwesome497 any eta on when you resolve the conflicts?

I'll try to tal today if I can finish with typeifying postgres `data-types`.  If not today, should be by the end of this week.

ephys:
I'm wondering if it wouldn't be worth it to group the different options for `ON CONFLICT` together:

```typescript
User.upsert(
  { name: 'John', isUnique: true }, 
  {
    onConflict: {
      // ON CONFLICT (column)
      attributes: [],
      // ON CONFLICT ... WHERE predicate
      where: { isUnique: true },
      // ON CONFLICT ON CONSTRAINT
      constraint: '',
    },
  },
);
```

src/dialects/abstract/query-generator.js
```diff
@@ -200,9 +200,30 @@ class QueryGenerator {
         const conflictKeys = options.upsertKeys.map(attr => this.quoteIdentifier(attr));
         const updateKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=EXCLUDED.${this.quoteIdentifier(attr)}`);
         onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')})`;
```
ephys, line 202: This value is never used, it's overwritten on line 226

test/unit/sql/insert.test.js
```diff
@@ -66,6 +66,68 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
...
+            },
+            upsertKeys,
+          },
+        ),
+        {
+          default:
+            'INSERT INTO `users` (`user_name`,`pass_word`) VALUES ($1,$2);',
+          postgres:
+            'INSERT INTO "users" ("user_name","pass_word") VALUES ($1,$2) ON CONFLICT ("user_name") WHERE "user_name" = \'test where value\' DO UPDATE SET "user_name"=EXCLUDED."user_name","pass_word"=EXCLUDED."pass_word","updated_at"=EXCLUDED."updated_at";',
+          mssql: 'INSERT INTO [users] ([user_name],[pass_word]) VALUES ($1,$2);',
```
ephys, line 119: IMO if `conflictWhere` is not supported by the dialect, the query generator should throw an error.

It doesn't match how things were done before (i.e. 'updateOnDuplicate' is ignored if unsupported) but I think we should slowly make Sequelize safe by default (not generating part of a query is unsafe as devs may think their query works).

Discussable of course :)

ephys (dismissed):
Just a few small notes

AllAwesome497:
> I'm wondering if it wouldn't be worth it to group the different options for `ON CONFLICT` together:
> 
> ```ts
> User.upsert(
>   { name: 'John', isUnique: true }, 
>   {
>     onConflict: {
>       // ON CONFLICT (column)
>       attributes: [],
>       // ON CONFLICT ... WHERE predicate
>       where: { isUnique: true },
>       // ON CONFLICT ON CONSTRAINT
>       constraint: '',
>     },
>   },
> );
> ```
> 
> It would be a breaking change though

I think this would make a lot of sense!

Would want something like: 
```ts
interface OnConflictOptions {
  // What to do if there's a conflict.
  resolution: 'update' | 'ignore', // could probably a union.
  // Not sure if this is the best key, but with ON CONFLICT UPDATE there's 2 where clauses.
  // The first one is for index clauses (to use partial unique indexes) and the second one is more flexible 
  // and used to determine whether each field should be updated.
  // This also has a bunch of restrictions to which fields can be accessed, e.g. you can't reference other tables. 
  where: WhereOptions<TAttributes>, 
  // The attributes which we're checking for conflicts - ie `conflictFields`
  attributes: (keyof TAttributes)[],
  // (only if resolution is `'update'`)
  // time_updated should maybe be here automatically, even if this is an empty array.   
  update: (keyof TAttributes)[],
  //  (again, only if resolution is 'update' - OnConflictOptions would probably be an intersecting union)
  // The clause evaluated on each conflicting row if an update should occur.
  // (this is not yet supported by sequelize)
  updateWhere: WhereOptions<TAttributes>
}
```
```

**Comment by ephys** [comment]
Designing these is always a headache. There are a lot of possibilities

The good news is that going with `onConflict` opens up the ability to support literals (in a future PR), e.g. `onConflict: literal('ON CONSTRAINT constraint_name DO NOTHING')`.
That, and the plan of improving literals, we only need to support the subset of syntax that benefits from using the pojo syntax.

A potential solution to the issue with `WHERE` being specifiable twice:

Naming inspired by https://www.postgresql.org/docs/current/sql-insert.html

```typescript
interface OnConflictOptions {
  // I think strings are fine here, there aren't many options
  resolution: 'update' | 'ignore',
  // previously: attributes & where
  target: Array<keyof TAttributes> | ConflictTarget<TAttributes>,
  // previously: update & updateWhere
  update: Array<keyof TAttributes> | UpdateActionOptions<TAttributes>,
}

interface ConflictTarget<TAttributes> {
  attributes: Array<keyof TAttributes>,
  where: WhereOptions<TAttributes>
}

interface UpdateActionOptions<TAttributes> {
  attributes: Array<keyof TAttributes>,
  where: WhereOptions<TAttributes>
}
```

Alternatively, `ConflictTarget` could be split into `target` and `targetWhere`, and `update` could be split into `update` & `updateWhere`

AllAwesome497:
> ```ts
> interface OnConflictOptions {
>   // I think strings are fine here, there aren't many options
>   resolution: 'update' | 'ignore',
>   // previously: attributes & where
>   target: Array<keyof TAttributes> | ConflictTarget<TAttributes>,
>   // previously: update & updateWhere
>   update: Array<keyof TAttributes> | UpdateActionOptions<TAttributes>,
> }
> 
> interface ConflictTarget<TAttributes> {
>   attributes: Array<keyof TAttributes>,
>   where: WhereOptions<TAttributes>
> }
> 
> interface UpdateActionOptions<TAttributes> {
>   attributes: Array<keyof TAttributes>,
>   where: WhereOptions<TAttributes>
> }
> ```

I think this is a good idea (though ideally `OnConflictOptions` is an intersecting union as it doesn't make sense to have `update` when `resolution` is set to `ignore`; assuming this is what you mean but I wanted to mention it anyways).

This is a pretty major change though, any thoughts on backwards compatibility?  We'd have to support both the new and old way at every layer (even `query-generator`) to prevent this from being a breaking change.

ephys:
> though ideally OnConflictOptions is an intersecting union

I agree

> This is a pretty major change though, any thoughts on backwards compatibility?

If we're landing this in v7, the easiest solution would be to throw if one of the old options is used, with a message describing what the new option is

But I think you need this in v6, and normalizing the options is going to be a bit of a pain. We could add "conflictFields", document the plan for `onConflict`, and implement it in a later major

AllAwesome497:
> If we're landing this in v7, the easiest solution would be to throw if one of the old options is used, with a message describing what the new option is
> 
> But I think you need this in v6, and normalizing the options is going to be a bit of a pain. We could add "conflictFields", document the plan for `onConflict`, and implement it in a later major

I think the "best" way to approach this is to:

1. Add `onConflict` to v6 in a feature update.  Translate the current options to `onConflict` w/ a warning.
2. In v7 or v8, throw an error if the field is used at all. 

This isn't super convenient on my end but IMO it's better than adding a field which we know will be removed; better to take the time to do things right the first time then spend more time fixing it later (especially from users of sequelize).

AllAwesome497:
Another thought on this - we could drop `upsert` altogether and add `onConflict` to `Model.create`.   I think this is the most elegant solution since (I believe) we'd also be able to use this for associations.

E.g. `Project.createMembership({ user_id: userId }, { onConflict: 'ignore' })` (it'd also be nice if we could infer attributes and allow `onConflict: 'ignore' | 'update'`)

mohanagy:
Any updates on this pr?

AllAwesome497:
> If we're landing this in v7, the easiest solution would be to throw if one of the old options is used, with a message describing what the new option is
> 
> But I think you need this in v6, and normalizing the options is going to be a bit of a pain. We could add "conflictFields", document the plan for `onConflict`, and implement it in a later major

Let's go with `conflictFields` and `conflictWhere` for v6 and fancy new api for v7.  This change is long overdue; imo it's best to get it out now and polish later.

clarencedesmarais:
This change will be very valuable to our team  🎉 Excited about having it merged.

motzyball:
Bumping this -- what's the status?

AllAwesome497:
> Bumping this -- what's the status?

Waiting for reviewers still; just bumped it in the sequelize slack.

ephys (dismissed):
I still think we need to do a thorough clean up of the different options upsert accepts but this is good

AllAwesome497:
@ephys re-requesting review - had to patch for compatibility with main which dismissed your approval.

WikiRik (dismissed):

packages/core/test/unit/sql/insert.test.js
```diff
@@ -66,6 +66,70 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
...
+        }
+
+        expectsql(result, {
+          default: new Error(
+            'missing dialect support for conflictWhere option',
+          ),
+          postgres:
+            'INSERT INTO "users" ("user_name","pass_word") VALUES ($sequelize_1,$sequelize_2) ON CONFLICT ("user_name") WHERE "user_name" = \'test where value\' DO UPDATE SET "user_name"=EXCLUDED."user_name","pass_word"=EXCLUDED."pass_word","updated_at"=EXCLUDED."updated_at";',
+          sqlite:
+            'INSERT INTO `users` (`user_name`,`pass_word`) VALUES ($sequelize_1,$sequelize_2) ON CONFLICT (`user_name`) WHERE `user_name` = \'test where value\' DO UPDATE SET `user_name`=EXCLUDED.`user_name`,`pass_word`=EXCLUDED.`pass_word`,`updated_at`=EXCLUDED.`updated_at`;',
```
WikiRik, line 129:
```suggestion
          'postgres sqlite':
            `INSERT INTO [users] ([user_name],[pass_word]) VALUES ($sequelize_1,$sequelize_2) ON CONFLICT ([user_name]) WHERE [user_name] = 'test where value' DO UPDATE SET [user_name]=EXCLUDED.[user_name],[pass_word]=EXCLUDED.[pass_word],[updated_at]=EXCLUDED.[updated_at];`,
```
A new change, but this is now possible

AllAwesome497:
@WikiRik can you approve this again?  the rebase dismissed it.

WikiRik (approved):

himanshugogroup:
conflictWhere not working if I use model column name instead of table column name.
eg; Model variable name is userName and table column name is user_name

PR #13419, opened 2021-08-05
feat(postgres, sqlite): allow override of conflict keys in Model.bulkCreate

AllAwesome497:
### Pull Request check-list

### Description of change

Allows for the override of the `upsertKeys` value passed to the query generator via `options.conflictFields` (for `Model.bulkCreate`).

Note that the different name is intentional - `upsertKeys` isn't as clear as `conflictFields` IMO.  (its unclear whether an `upsertKey` is referring to a key that will be updated or a key that should cause the existing row to be updated if it matches)

This is a part of #13270 and it is needed for `conflictWhere` to work (properly at least) in some cases.  Though it is worth noting that there should be better logic for determining defaults, I think this is still handy for special edge cases.   

Example use case (using a many:many joiner table)
```ts
const Memberships: typeof Model = sequelize.define(
  'my_joiner_table',
  {
    id: {
      primaryKey: true,
      autoIncrement: true,
      type: DataTypes.INTEGER,
      allowNull: false,
    },
    group_id: {
      type: DataTypes.INTEGER,
      primaryKey: true,
      allowNull: false,
      // Note that the ideal way to make this relation is to have a unqiue partial index on `deletedAt IS NULL`
      // that way the unique constraint will ignore soft-deleted rows.
      // Once #13420 lands, that will be the ideal way to make this type of relation.
      unique: 'my_constraint',
    },
    user_id: {
      type: DataTypes.INTEGER,
      primaryKey: true,
      allowNull: false,
      unique: 'my_constraint',
    },
  },
  { timestamps: true }
);

// In usage
await Memberships.bulkCreate(
  [
    {
      group_id: 5,
      team_id: 3,
    },
    {
      group_id: 5,
      team_id: 4,
    },
  ],
  {
    conflictFields: ['group_id', 'user_id'],
  }
); // `ON CONFLICT ("group_id", "user_id")` instead of `ON CONFLICT ("id", "group_id", "user_id")` 

```


ThePletch:
Wanted to note here that this feature doesn't seem to properly support the `underscored: [true/false]` model option. If you call `init` for a model and list fields using camel case, then pass `underscored: true` in the table options, most other sequelize options will internally translate back to the underlying snake-case field name, but this change passes the fields directly into the query, and thus expects them to already be in snake case.

e.g.
```javascript
Floob.init(
  {
    someField: {
      type: DataTypes.TEXT,
    },
  },
  {
    sequelize,
    underscored: true,
  }
);

Floob.bulkCreate(
  // ...
  {
    conflictFields: ['someField'],
  }
  // ...
);
// -> SequelizeDatabaseError: column "someField" does not exist
```

AllAwesome497:
> Wanted to note here that this feature doesn't seem to properly support the `underscored: [true/false]` model option. If you call `init` for a model and list fields using camel case, then pass `underscored: true` in the table options, most other sequelize options will internally translate back to the underlying snake-case field name, but this change passes the fields directly into the query, and thus expects them to already be in snake case.
> 
> e.g.
> 
> ```js
> Floob.init(
>   {
>     someField: {
>       type: DataTypes.TEXT,
>     },
>   },
>   {
>     sequelize,
>     underscored: true,
>   }
> );
> 
> Floob.bulkCreate(
>   // ...
>   {
>     conflictFields: ['someField'],
>   }
>   // ...
> );
> // -> SequelizeDatabaseError: column "someField" does not exist
> ```
> 
> (also, you get a type error trying to list the snake case version of the fields in the new attribute, since they're not properties of the model)

This should be fixed now.

ephys (dismissed):
Looks good. It has a failing unit test that tries to access the database

This comment is also relevant https://github.com/sequelize/sequelize/pull/13411#issuecomment-1066833203 :)

test/unit/model/bulk-create.test.js
```diff
@@ -11,36 +11,77 @@ const current = Support.sequelize;
...
+      { timestamps: false },
+    );
+
+    const stub = sinon
+      .stub(current.getQueryInterface(), 'bulkInsert')
+      .resolves([]);
 
-      this.stub = sinon.stub(current.getQueryInterface(), 'bulkInsert').resolves([]);
+    beforeEach(async () => {
+      await Model.sync({ force: true });
```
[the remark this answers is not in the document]
AllAwesome497: oops!  dunno what I was thinking.

test/integration/model/bulk-create.test.js
```diff
@@ -745,6 +745,104 @@ describe(Support.getTestDialectTeaser('Model'), () => {
             this.User.bulkCreate(data, { updateOnDuplicate: [] }),
           ).to.be.rejectedWith('updateOnDuplicate option only supports non-empty array.');
         });
+
+        if (current.dialect.supports.inserts.conflictFields) {
+          it('should respect the conflictFields option', async function () {
+            const Model = this.sequelize.define(
```
ephys, line 751: Because I've seen tests modify the base Model class, renaming this class will help me quickly determine if there is a red flag in the test in the future :)

```suggestion
            const Permission = this.sequelize.define(
```

test/types/bulkCreate.ts
```diff
@@ -0,0 +1,76 @@
...
+      testEnum: 'eafe',
+    },
+    {
+      // @ts-expect-error testString should be a string or null
+      testString: 324,
+    },
+  ];
+
+  const newItems: Array<
+    MakeNullishOptional<InferCreationAttributes<TestModel>>
```
ephys, line 39: `Infer*` should only be used during model definition for portability, but the following type is equivalent to what you're doing:

```suggestion
    CreationAttributes<TestModel>
```

test/types/bulkCreate.ts
```diff
@@ -0,0 +1,76 @@
...
+  > = [
+    {
+      // @ts-expect-error this isn't an enum value
+      testEnum: 'eafe',
+    },
+    {
+      // @ts-expect-error testString should be a string or null
+      testString: 324,
+    },
+  ];
```
ephys, line 36: These (+ line 19) are already tested in `test/types/infer-attributes.ts` :)

test/types/bulkCreate.ts
```diff
@@ -0,0 +1,76 @@
...
+import { sequelize } from './connection';
+import type { MakeNullishOptional } from '@sequelize/core/types/utils';
+
+class TestModel extends Model<
+  InferAttributes<TestModel>,
+  InferCreationAttributes<TestModel>
+> {
+  declare id: CreationOptional<number>;
+  declare testString: CreationOptional<string | null>;
+  declare testEnum: CreationOptional<'d' | 'e' | 'f' | null>;
```
ephys, line 16: somewhat recent change: nullable fields are always optional

```suggestion
  declare testString: string | null;
  declare testEnum: 'd' | 'e' | 'f' | null;
```

ephys (dismissed):
The API design in https://github.com/sequelize/sequelize/pull/13411 would be really nice to include here (don't hesitate to ping me in case I don't notice the notification so we can resolve the API design fast enough to be non-blocking for you), otherwise only a few small notes

WikiRik:
@ephys ping

Phillip-Kemper:
is anything blocking to merge this in? :)

src/model.d.ts
```diff
@@ -1158,6 +1158,12 @@ export interface BulkCreateOptions<TAttributes = any> extends Logging, Transacti
    * Return all columns or only the specified columns for the affected rows (only for postgres)
    */
   returning?: boolean | Array<keyof TAttributes | Literal | Col>;
+
+  /**
+   * Optional override for the conflict fields in the ON CONFLICT part of the query.
+   * Only supported in Postgres >= 9.5 and SQLite >= 3.24.0
+  */
```
ephys, line 1165:
```suggestion
   */
```

src/model.d.ts
```diff
@@ -1158,6 +1158,12 @@ export interface BulkCreateOptions<TAttributes = any> extends Logging, Transacti
...
+   conflictFields?: Array<keyof TAttributes>;
```
ephys, line 1166: We're not using `fields` anymore in the project, it's slowly being replaced by either `attributes` (if the user must specify the JS name) or `columns` (if it's the DB/sql name)

Based on the typing, I'd propose `conflictAttributes`

src/model.js
```diff
@@ -2825,17 +2825,25 @@ Specify a different name for either index to resolve this issue.`);
...
           options.updateOnDuplicate = options.updateOnDuplicate.map(attr => model.rawAttributes[attr].field || attr);
 
-          const upsertKeys = [];
-
-          for (const i of model.getIndexes()) {
-            if (i.unique && !i.where) { // Don't infer partial indexes
-              upsertKeys.push(...i.fields);
+          if (options.conflictFields) {
+            options.upsertKeys = options.conflictFields.map(
+              attr => (Object.prototype.hasOwnProperty.call(this.rawAttributes, attr)
```
ephys, line 2830: This is a pattern we use all the time, we should have a util method `getModelColumnName(model: ModelStatic, attributeName: string): string`

ephys (changes requested):

src/model.d.ts
AllAwesome497: My only concern there is that we already have this name for `Model.upsert` so that could be confusing.

src/model.js
AllAwesome497: Agreed though I think it should be a `protected` attribute of `Model`.  Probably something to be added in a followup pr.

nguyenphuocnhatthanh:
When this PR will be approved ?

AllAwesome497:
> is anything blocking to merge this in? :)

> When this PR will be approved ? @AllAwesome497 can you continue to resolve these comments ? This PR's helpful.

Missed these @nguyenphuocnhatthanh @Phillip-Kemper.

I was a bit busy & stopped checking in on these for a while.  Now that #13411 has been merged I'll be prioritizing this more then #13420.

ephys (approved):
Looks good :)

PR #13422, opened 2021-08-07
fix: expect result is 0 but got NaN

caolvchong:
in `rawSelect (lib/dialects/abstract/query-interface.js)`, if `dataType` is `DataTypes.INTEGER` or `DataTypes.BIGINT`，`result` will call parseInt:(https://github.com/sequelize/sequelize/blob/56bb1d6ef9827f604d7bcef945abb7e213f2322d/lib/dialects/abstract/query-interface.js#L1013-L1015):

if result is `null`, result will be convert to `NaN`.

in `aggregate (lib/model.js)` call `rawSelect`, judge the return value is null will return `0`，but the value is `NaN`, `NaN` will return:(https://github.com/sequelize/sequelize/blob/56bb1d6ef9827f604d7bcef945abb7e213f2322d/lib/model.js#L1986-L1990
)

expect `0` but got `NaN`


caolvchong:
@papb @SimonSchick have a look pls

sdepold:
@caolvchong Could you add a test for this?

sdepold:
Nice one. I think the tests could be simplified but I guess it suffices :) 

Thanks a bunch!

PR #13465, opened 2021-08-28
fix(model): columns with getter sort problem

bulog:
### Pull Request check-list

### Description of change

The column(s) with getter always show at the beginning of the column(s) without getter

Please check the example code bellow of the problem is
1. define a model with the getter for column `age`.
```javascript
const Model = db.define('model', {
      name: DataTypes.STRING,
      age: {
        type: DataTypes.INTEGER,
        get() {
          const rawValue = this.getDataValue('age');
          return rawValue + 1;
        }
      }
}, { timestamps: false });
```
use `findAll` and `findOne`, then return the result to JSON

we think we should get

```
{
  "id": 0,
  "name": "",
  "age": 1
}
```
but we got

```
{
  "age": 1,
  "id": 0,
  "name": ""
}
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3499,27 +3499,26 @@ class Model {
       const values = {};
       let _key;
 
-      if (this._hasCustomGetters) {
-        for (_key in this._customGetters) {
-          if (
-            this._options.attributes
-            && !this._options.attributes.includes(_key)
-          ) {
-            continue;
+      for (_key in this.dataValues) {
+        if (Object.prototype.hasOwnProperty.call(this.dataValues, _key)) {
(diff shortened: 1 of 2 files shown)


Keimeno:
If I'm understanding this correctly, this PR is about the sort order of keys within an object? That should be irrelevant to the result. Also, `console.log` automatically sorts keys alphabetically, which might be another reason that `age` the topmost key.

bulog:
Yes, I agree with "That should be irrelevant to the result." I think I should close this pull request

If I add another column `zip` with getter after the `age`
Then the json I got：
```
{
  "age": 1,
  "zip": "",
  "id": 0,
  "name": ""
}
```

PR #13467, opened 2021-08-28
fix(types): change attributes type to 'object | undefined'

### LINKED ISSUE #13466: Typing for QueryInterface's bulkInsert & bulkUpdate attributes parameter is incorrect
## Issue Creation Checklist

### SSCCE

```ts
const { createSequelizeInstance } = require('./dev/sscce-helpers');
const { Model, DataTypes } = require('.');
import { QueryInterface } from 'sequelize';

class User extends Model {}
User.init({
  username: DataTypes.STRING,
  birthday: DataTypes.DATE
}, { sequelize, modelName: 'user' });

(async () => {

   const queryInterface: QueryInterface = sequelize.getQueryInterface();

   // Gives typying error because attributes type is string[] | string, but that is not correct.  
  await queryInterface.bulkUpdate("user", {}, {}, {}, attributes);
 await queryInterface.bulkInsert("user", {}, {}, {}, attributes);
  
  await sequelize.close();
})();
```

#### The type of `attributes` parameter is incorrect for [`bulkInsert`](https://github.com/sequelize/sequelize/blob/56bb1d6ef9827f604d7bcef945abb7e213f2322d/types/lib/query-interface.d.ts#L501) & [`bulkUpdate`](https://github.com/sequelize/sequelize/blob/56bb1d6ef9827f604d7bcef945abb7e213f2322d/types/lib/query-interface.d.ts#L523).

 * Currently, the type of `attributes` is `string[] | string | undefined` but it should be of type `object`.
 * Query interface API [docs](https://sequelize.org/master/class/lib/dialects/abstract/query-interface.js~QueryInterface.html)
Screenshot 2021-08-29 at 12 46 37 AM

Screenshot 2021-08-29 at 12 59 46 AM

### Environment

- Sequelize version: v6.6.5
- Node.js version: 14.16.1
- If TypeScript related: TypeScript version: 3.9.6

aryem:
### Pull Request check-list

### Description of change

 change attributes type to 'object':
 - attributes param in QueryInterface.bulkInsert() change to 'object | undefined' from 'string[] | string | undefined'.
 - attributes param in QueryInterface.bulkUpdate() change to 'object | undefined' from 'string[] | string | undefined'.
 - add '{}' as last param in bulkInsert() & bulkUpdate() tests

Closes #13466


types/lib/query-interface.d.ts
```diff
@@ -498,7 +498,7 @@ export class QueryInterface {
     tableName: TableName,
     records: object[],
     options?: QueryOptions,
-    attributes?: string[] | string
+    attributes?: object
```
AllAwesome497, line 501: This isn't right  either actually.

I don't know if there's already a type for this or not.  At the top of my head, it should be moreso like:


```ts

type Attribute = Literal | string | [Literal | Fn | Col , string]

type Attributes = Array<Attribute> | { include?: Array<Attribute>, exclude: Array<Attribute> }

```

note: this is mostly psuedocode, I don't remember the exact typings for col, fn, and literals.

AllAwesome497 (commented):
huh, why are you using the query interface directly? Is this correct on the model definition but not queryInterface?

types/lib/query-interface.d.ts
aryem: @AllAwesome497 You are right, It isn't correct either. I think `ModelAttributes` from Model.d.ts seems to be appropriate here. What do you say?.
https://github.com/sequelize/sequelize/blob/23aa67e9abe440027aa07091beb35834e584e42e/types/lib/model.d.ts#L1358

AllAwesome497: Such as this?

```ts
type Attribute<ModelAttr extends string> = Literal | Col | ModelAttr | [Literal | Fn | Col, ModelAttr];

type Attributes<> = Array<ModelAttr extends string = string> | { include?: Array<Attribute<ModelAttr>>, exclude: Array<Attribute<ModelAttr>> }
```

In usage:
```
attributes?: Attributes<keyof ModelAttributes>
```

If so, then I think that would be the ideal solution.

ephys:
We merged https://github.com/sequelize/sequelize/pull/13945 a little while ago which updated bulkInsert. I'm pretty sure the new typings for bulkInsert are correct and should be an object instead of an array based on their usage in query-generator.js

bulkUpdate hasn't been touched by that other PR however based on their usage in `updateQuery` in query-generator, it looks like their type should be an object too, not an array

If you're still willing to update this pretty old PR (sorry about that), I can take a look into merging this

ephys:
Looks like this has since been fixed by another. Thank you for opening this one nonetheless. Sorry we never got around to merging it :)

PR #13477, opened 2021-09-04
Issue 13302

### LINKED ISSUE #13302: Incorrect typescript typings for findAndCountAll function

aryem:
### Pull Request check-list

### Description of change

add overload type for findAndCountAll:
 - add overload type for findAndCountAll with the group.
 - add test for checking return type of findAndCountAll with and without group
 - update return type of findAndCountAll in model.js docs comment

Closes #13302

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2069,7 +2069,7 @@ class Model {
    * @see
    * {@link Model.count} for a specification of count options
    *
-   * @returns {Promise<{count: number, rows: Model[]}>}
+   * @returns {Promise<{count: number | number[], rows: Model[]}>}
    */
   static async findAndCountAll(options) {
     if (options !== undefined && !_.isPlainObject(options)) {
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -1923,6 +1923,10 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
     this: ModelStatic<M>,
     options?: FindAndCountOptions<M['_attributes']>
   ): Promise<{ rows: M[]; count: number }>;
+  public static findAndCountAll<M extends Model>(
+    this: ModelStatic<M>,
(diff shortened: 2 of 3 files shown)


Keimeno (approved):

Keimeno:
Hi @aryem, 
thanks for your PR and for addressing the issues in #13303.
I will close #13303 in favour of this PR.

PR #12955, opened 2021-01-18
feat(tsvector): TSVECTOR support

papb:
This is PR #12631 rebased on top of the new `main` branch.

PR #13481, opened 2021-09-06
Resolved the typescript issue with the Op.match (https://github.com/sequelize/sequelize/pull/12955)

vikasg603:
resolved the typescript issue for Op.match (https://github.com/sequelize/sequelize/pull/12955)

### DIFF
--- a/types/lib/operators.d.ts
+++ b/types/lib/operators.d.ts
@@ -244,6 +244,18 @@ declare const Op: {
    */
   readonly lte: unique symbol;
   /**
+   * Operator @@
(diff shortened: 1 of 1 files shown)


Keimeno:
Hi @vikasg603, thanks for your PR.

Can you please add the Operator to the `WhereOperators` interface in the `model.d.ts` file as well, and also typing tests to ensure, that `Op.match` works as expected? 

You can look at the following PR as reference: https://github.com/sequelize/sequelize/pull/13499/files

PR #13482, opened 2021-09-07
fix(postgres): fix `findCreateFind` to work with postgres transactions

Americas:
### Description of change

This PR fixes the `findCreateFind` method so it can work with postgres transactions without the risk of failing the transaction and rendering it unusable, even when the code logic allows things to continue.

To do so, when we run the create step in the method, we add the `ignoreDuplicates` option if the dialect is postgres and a transaction object is passed. Then, when dealing with the query response, if the response returns no rows (and is an insert query and is for a single instance), the process throws an error (`EmptyResultError`) instead of assuming that something was created, when in fact nothing was.


diogotorres97:
@papb @janmeier @sdepold any updates on this?

PR #13490, opened 2021-09-10
fix(select): do not force set `subQuery` to `false`

floydspace:
This occurs on validating included elements function in spite of a user sets it to `true` in `include` options

### Pull Request check-list

### Description of change

Here it is the SSCCE:
```javascript
const { createSequelizeInstance } = require('./dev/sscce-helpers');
const { Model, DataTypes } = require('.');

class User extends Model {}
User.init({
  scopeId: DataTypes.INTEGER
}, { sequelize, modelName: 'User' });

class Company extends Model {}
Company.init({
  name: DataTypes.STRING,
  public: DataTypes.BOOLEAN,
  scopeId: DataTypes.INTEGER
}, { sequelize, modelName: 'Company' });

class Profession extends Model {}
Profession.init({
  name: DataTypes.STRING,
  scopeId: DataTypes.INTEGER
}, { sequelize, modelName: 'Profession' });

User.belongsTo(Company, { foreignKey: 'companyId' });
User.belongsTo(Profession, { foreignKey: 'professionId' });
Company.hasMany(User, { foreignKey: 'companyId' });
Profession.hasMany(User, { foreignKey: 'professionId' });

(async () => {

  const companies = await Company.findAll({
    where: { '$Users.Profession.name$': 'test' },
    include: [
      {
        model: User,
        attributes: [],
        include: [
          {
            model: Profession,
            attributes: [],
            required: true
          }
        ],
        required: true,
        subQuery: true
      }
    ],
    limit: 5,
    offset: 0,
    subQuery: true
  });

  await sequelize.close();
})();

which generate an invalid SQL query

```

I expect that query would not use not existing alias

But it generated query like this:
```sql
SELECT `Company`.* FROM (
    SELECT `Company`.`id`, `Company`.`name`, `Company`.`public`, `Company`.`scopeId`, `Company`.`createdAt`, `Company`.`updatedAt` FROM `Companies` AS `Company` 
    INNER JOIN `Professions` AS `Users->Profession` ON `Users`.`professionId` = `Users->Profession`.`id` 
    WHERE `Users->Profession`.`name` = 'test' AND ( 
        SELECT `Users`.`companyId` FROM `Users` AS `Users` 
        INNER JOIN `Professions` AS `Profession` ON `Users`.`professionId` = `Profession`.`id` 
        WHERE (`Users`.`companyId` = `Company`.`id`) LIMIT 1 
    ) IS NOT NULL LIMIT 0, 5
) AS `Company` 
INNER JOIN `Users` AS `Users` ON `Company`.`id` = `Users`.`companyId`;
```
as you can see in line 3 it joins to the `Users` table but the `Users` alias is not exist in current sub select scope, it has been concatenated at the end of query. So DB returns error: `Unknown column 'Users.professionId' in 'on clause'`.

The following PR fixes the issue, for me at least. And now generates the following SQL:
```sql
SELECT `Company`.* FROM (
    SELECT `Company`.`id`, `Company`.`name`, `Company`.`public`, `Company`.`scopeId`, `Company`.`createdAt`, `Company`.`updatedAt` FROM `Companies` AS `Company` 
    INNER JOIN `Users` AS `Users` ON `Company`.`id` = `Users`.`companyId`
    INNER JOIN `Professions` AS `Users->Profession` ON `Users`.`professionId` = `Users->Profession`.`id` 
    WHERE `Users->Profession`.`name` = 'test' AND ( 
        SELECT `Users`.`companyId` FROM `Users` AS `Users` 
        INNER JOIN `Professions` AS `Profession` ON `Users`.`professionId` = `Profession`.`id` 
        WHERE (`Users`.`companyId` = `Company`.`id`) LIMIT 1 
    ) IS NOT NULL LIMIT 0, 5
) AS `Company`;
```
All tests look fine. 

Please have a look on it.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -528,7 +528,7 @@ class Model {
 
       if (include.subQuery !== false && options.hasDuplicating && options.topLimit) {
         if (include.duplicating) {
-          include.subQuery = false;
+          include.subQuery = include.subQuery || false;
           include.subQueryFilter = include.hasRequired;
         } else {
           include.subQuery = include.hasRequired;
@@ -538,7 +538,6 @@ class Model {
         include.subQuery = include.subQuery || false;
         if (include.duplicating) {
           include.subQueryFilter = include.subQuery;
-          include.subQuery = false;
         } else {
           include.subQueryFilter = false;
           include.subQuery = include.subQuery || include.hasParentRequired && include.hasRequired && !include.separate;
--- a/test/unit/sql/select.test.js
+++ b/test/unit/sql/select.test.js
@@ -449,7 +449,7 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
       });
     });
 
-    it('include (subQuery alias)', () => {
+    describe('include (subQuery alias)', () => {
       const User = Support.sequelize.define('User', {
         name: DataTypes.STRING,
         age: DataTypes.INTEGER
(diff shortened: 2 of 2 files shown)


floydspace:
Let me know if I'm using incorrect  find options, I also tried using `duplicating` flag which says nothing to me from documentation, but it didn't fix the query. 
I need nested filtering and pagination at the same query.
Thank you

sdepold:
Hey @floydspace, thanks for your work. Do you think we have to update any documentation or are we just fixing it as such as you would have expected to begin with.

I have to admit that I only understand half of what is going on here, but I can tell that we are generating a broken SQL query and it's better now :D

PR #13491, opened 2021-09-10
fix(types): add hooks to InstanceDestroyOptions type

apazzolini:
### Description of change

In the model `destroy` method, the code allows for the user to specify `hooks`, which was missing from the `InstanceDestroyOptions` type definition.

See: https://github.com/sequelize/sequelize/blob/56bb1d6ef9827f604d7bcef945abb7e213f2322d/lib/model.js#L4150-L4197


ephys:
Looks good to me but we should add a typing test to prevent a regression. It's been a while since you've opened this PR, I'll add it for you

ephys (approved):

WikiRik (approved):

PR #13493, opened 2021-09-11
fix: findAndCountAll return type (groupBy scenario)

nikli2009:
### Description of change

* To improve `findAndCountAll` return type. (when user uses `groupBy` in query, the `count` field is in fact an array of `{ count: number }`)


nikli2009:
> Hi @nikli2009,
> thanks for your contribution to sequelize. Unfortunately, I will have to close this PR, since there already is an ongoing PR (#13477) regarding this issue (#13302), which also adds tests and provides a more solid fix using function overloading.

No worries, didn't check before creating PR, happy to see there is other PR already covers this.

PR #13530, opened 2021-10-02
Update collaboration related scripts + docs

sdepold:
Update of docker usage documentation


fncolon:
can you update the sequelize hooks docs as well? @sdepold

PR #13531, opened 2021-10-02
Add tests and fix for usage of count with grouping

sdepold:
Addendum to #13477

### DIFF
--- a/test/integration/model.test.js
+++ b/test/integration/model.test.js
@@ -1720,7 +1720,10 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         group: ['data']
       });
       expect(count).to.have.lengthOf(2);
-      expect(count).to.deep.equal([{ data: 'B', count: 1 }, { data: 'A', count: 2 }]);
+
+      // The order of count varies across dialects; Hence find element by identified first.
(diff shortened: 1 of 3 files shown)


Keimeno:
Looks like we have a bug with the Postgres dialect, it seems to return a string as count, and not a number: https://github.com/sequelize/sequelize/pull/13531/checks?check_run_id=3776129422

sdepold:
Seems that there is actually a missing conversion in pg where we receive a string instead of a number:



```
      AssertionError: expected [ { userId: 1, count: '1' } ] to deeply equal [ { userId: 1, count: 1 } ]
      + expected - actual

       [
         {
      -    "count": "1"
      +    "count": 1
           "userId": 1
         }
       ]
```

test/integration/model.test.js
```diff
@@ -1720,6 +1720,10 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         group: ['data']
       });
       expect(count).to.have.lengthOf(2);
+
+      // The order of count varies across dialects; Hence find element by identified first.
```
sdepold, line 1724: We could fix this by ordering the to be grouped properties by value but that seems a bit extensive.

PR #13552, opened 2021-10-10
fix(model): error saving with custom field names and no primary key

### LINKED ISSUE #13421: Sequelize Error: no such column with no primary key and underscores
## Issue Creation Checklist

## Bug Description

When a model with no primary key and underscored column names is retrieved using findAll and where option, and then updated, and error occurs. When a model is saved, the generated sql for the update tries to use the camelCase attribute names instead of the underscored column names.

### SSCCE

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/179

The db record should be updated using the correct underscored column names.

The generated sql uses the camelCase attribute names in the where clause, and an error occurs

```

DatabaseError [SequelizeDatabaseError]: SQLITE_ERROR: no such column: userId

```

### Additional context

This issue was introduced in sequelize@6.6.4 in https://github.com/sequelize/sequelize/commit/1a16b915ff45ac621a0517aea6c2d86557149500 . You can see further discussion of that fix and this issue in https://github.com/sequelize/sequelize/issues/12868

### Environment

- Sequelize version: 6.6.5
- Node.js version: v14.16.1

roikoren755:
### Pull Request check-list

### Description of change

If no primary key exists in a model, `Model#save` uses the `options.where` passed to its constructor as the where clause, but this might use renamed columns.
This PR fixes that issue by using `Utils.mapWhereFieldNames` to get the right field names :)
Closes #13421.


roikoren755:
@AllAwesome497 thanks for the review!

I updated the test, I think it should be fine now, though my experience is more with `jest`, not `mocha`.

The MSSQL tests should pass now, I think I may have stumbled upon a different bug - when the dialect is MSSQL, if you remove the primary key attribute (in this case the automatically generated `id` field), the `Model#primaryKeyField` doesn't update, and queries with limit or offset (like `Model.findOne`) will be generated with reference to that attribute, and fail to execute. I changed the test added in this PR to use `Model.findAll`, to avoid running into that bug. If it is indeed a bug, I'll open an issue on it :).

Regarding the postgres-native tests, I had nothing to pull from main, unfortunately. It looks like the beforeEach hook is taking more than 30 seconds to complete, leading to a timeout and fail test. I'll try looking at it next, though I can't seem to get `pg-native` to install on my machine :(.

ephys:
We have an update regarding this PR:

We plan on removing the ability to call instance methods `save`, `update`, `reload`, `destroy`, `restore`, `decrement`, and `increment` on models that do not have a primary key.

See https://github.com/sequelize/sequelize/pull/14173#issuecomment-1065388719

ephys:
I'll close this PR as we do not plan on supporting these instance methods on models that do not include a PK, due to how unsafe it would be, per https://github.com/sequelize/sequelize/issues/14402

PR #13561, opened 2021-10-13
fix(types): fix incorrect type for Op.overlap argument

werkshy:
The overlap operator `&&` is true when any element from one array is
present in the other array. For this reason the type of the
right-hand-side of the Op.overlap operator should be an array, not a
Rangeable.

This problem was exposed with the upgrade to Typescript 4.4.

### Pull Request check-list


sdepold:
@werkshy just seeing now that there is no test case. do you have an example for this change?

sdepold:
@AllAwesome497 Would you like to create a replacement for this PR?

werkshy:
Sorry folks, I clearly don't have time to work on this. I would have added tests but I found the tests didn't run to completion in a reasonable time on my laptop and it wasn't clear how to run a subset.

ephys:
You can use `npm run test-typings` (i think) to only run typing tests

I'm still interested in seeing this merged, I'll keep it open and tag it as need maintainer until we have a replacement :)

ephys:
I'm taking over this PR and including it in https://github.com/sequelize/sequelize/pull/14022

PR #13568, opened 2021-10-17
refactor(connection-manager): change nullish coalescence implementation

sdepold:
Follow-up to #13376


sdepold:
@soryy708 FYI :)

soryy708:
@sdepold ?

Issue #4048, opened 2015-07-02
Problem with enum when using with schema


Uhsac:
Hi,

I have found a bug today, if i'm using enum datatypes and tried to sync in a specified schema it don't work.

Example :

```javascript
  var InnovationDetail = sequelize.define('InnovationDetail', {
    language: dataTypes.ENUM(
                  'fr',
                  'en'),
    name: dataTypes.STRING,
    description: dataTypes.STRING
  }, {
    classMethods: {
      associate: function(models) {
      }
    }
  });
```

If I try to sync this in a specific schema like this :

```javascript
InnovationDetail.sync({
  schema: 'public',
  force: true
});
```

I have this error :

```console
Unhandled rejection SequelizeDatabaseError: type "enum_public.InnovationDetails_language" does not exist
```

It prefix the enum with the schema and I think it don't have to do that.

If I remove the schema it work.

Thanks.

janmeier:
Duplicate of https://github.com/sequelize/sequelize/issues/3171

PR #13592, opened 2021-10-25
fix(postgres): Support enum array

codemers:
### Pull Request check-list

### Description of change

Currently when dealing with an array of enum using the Postgres dialect, Sequelize generates an invalid enumeration cast key while inserting data. 

Related issue (I think): https://github.com/sequelize/sequelize/issues/4048 

DDL of the table:

```
create table "Table"
(
    id uuid not null constraint "Table_pkey" primary key,
    "createdAt"      timestamp with time zone         not null,
    "updatedAt"      timestamp with time zone         not null,
    "deletedAt"      timestamp with time zone,
    column           "enum_Table_column"[] not null
);

```
DDL of the type 

```
create type "enum_Table_column" as enum ('Value1', 'Value2', 'Value3');
```

Then while inserting data here's the cast key generated by sequelize: 

```
array['Value1']::"enum_public.Table_Column"[]
```

The correct cast key should be:

```
array['Value1']::"enum_Table_Column"[]
```

This pr fixes this issue.


sdepold:
@codemers Can you look into the tests and rebase your branch?

sdepold:
Is this by any chance the same as https://github.com/sequelize/sequelize/pull/13807 ?

sdepold:
cc @nahog

nahog:
Yes, this is the same issue, but the PR here only works for public as is not passing the schema if its redefined, the #13807 fix handle both scenarios, when is using the public/empty schema, that both:
"enum_my_enum"[] and "public"."enum_my_enum"[] works
and the scenario where schemas are needed like:
"my_schema"."enum_my_enum"[]

@codemers try the current `main` branch (merged today) code to confirm the issue is fixed for you, but I think it is.

nahog:
@codemers 6.12.2 was just released with a fix for this issue. Please try that version and if it works this bug can be probably closed.

PR #13594, opened 2021-10-25
fix(upsert): fall back to DO NOTHING if no update key values provided

slickmb:
We found that if we did an upsert on a record with the `fields` option, if the data values being passed into the upsert did not include any of those fields, that we'd get a postgres syntax error when the query was executed.

example:
```js
   // for this example, MyModel has attributes id, a, b, x, and y.
   const [ instance ] = await MyModel.upsert({ id: 1, x: 1, y, 2 }, { fields: [ 'a', 'b' ], returning: true });
   // this would result in the following query:
   //    INSERT INTO "MyModels" ("id", "x", "y")
   //      VALUES ($1, $2, $3)
   //       ON CONFLICT ("id") DO UPDATE SET RETURNING "id", "a", "b", "x", "y";
   //                                      ^^^ notice the missing clause here.
```

If there are no update values included, and we call `upsert`, I would expect the query to still attempt the insert and just not update anything if there's a conflict.

To fix, we added code into the abstract query generator that will recognize that none of the update values were actually provided, and revert to use a `DO NOTHING` instead of a `DO UPDATE`.

For the `ON DUPLICATE KEY UPDATE` sql languages, we try fall back to just updating the identity column that triggered the duplicate constraint (per https://stackoverflow.com/questions/4596390/insert-on-duplicate-key-do-nothing).  If there are no identity columns nor update values, then we throw an error.

I'm also fixing an error in the mssql upsertQuery, where it was looking at the `updateValues` instead of the `insertValues` to determine if the `IDENTITY_INSERT` needed to be turned on.  This was preventing my unit test for the above fix from passing.


AllAwesome497:
This seems like a breaking change, and IMO doesn't make much sense.  `ON CONFLICT DO NOTHING` isn't an upsert, its moreso a "try to insert."

From the postgres documentation on insert:

> `ON CONFLICT DO UPDATE` guarantees an atomic `INSERT` or `UPDATE` outcome; provided there is no independent error, one of those two outcomes is guaranteed, even under high concurrency. This is also known as `UPSERT` — “`UPDATE` or `INSERT`”.

IMO it makes more sense to do one of the following:

Add an `onConflict` field to `Model.create`
```js
Model.create(
  { /* ... */ },
  {
    // assume ConflictResolution is an enum. Naming could probably be better.
    onConflict: Sequelize.ConflictResolution.DO_NOTHING,
  }
);
```

Add a `tryCreate` to `Model`.
```js
Model.tryCreate({ /* ... */ });
```

At the moment, I believe this should work as a workaround:

```js
Model.bulkCreate(
  [{ /* ... */ }],
  {
    ignoreDuplicates: true,
  },
);
```

slickmb:
Thank you for the feedback.  You make a good point, and I see where you're coming from.  Before we make a decision, please consider the following:

In the use-case outline in the PR, the current behavior of sequelize is that sequelize generates mal-formed SQL and the database throws when sequelize attempts to execute that mal-formed SQL.  The PR is not breaking any behavior that is currently working and, even if the PR is not accepted, sequelize should do something more graceful than generating bad SQL.  That said, the correct solution may be to detect this scenario and throw a more descriptive error instead of generating bad SQL or generating an `INSERT ON CONFLICT DO NOTHING` as this PR is doing.

The suggested fix in this PR operates under the assumption that when executing `upsert` the caller means "insert the row and/or update these fields on the row where possible".  In the above use-case, the data has insert-able fields, but none of the updateable fields are present.  The insert is still possible, and, since there are no updatable fields, the update is a noop.  

Phrased another way, by calling
```
await MyModel.upsert({ id: 1, x: 1, a: 1 }, { fields: [ 'a', 'b' ] }); // There ARE updatable fields present
``` 
the caller is asking us (per the current, working sequelize behavior) to insert with fields `[ 'id', 'x', 'a' ]`, or to update with fields `[ 'a' ]`.  Notice how sequelize is *not* throwing even though the `'b'` field is not present as an updatable field.  The `upsert` call works.

What is the expected behavior if the caller does the following?
```
await MyModel.upsert({ id: 1, x: 1 }, { fields: [ 'a', 'b' ] }) // There are NOT any updatable fields present
``` 
I believe the caller is asking `insert_fields: ['id', 'x'], update_fields: []`.  Everything hinges on whether `update_fields: []` should be an error condition.  In the previous example, sequelize did not throw when `'b'` was missing from the data.  Why should it throw if both `'a'` and `'b'` are missing?  My opinion is that sequelize should *not* throw, and that the update portion of the `upsert` should be a noop.... thus, why I convert it to an `ON CONFLICT DO NOTHING`.

There is precedence for this noop style behavior in sequelize.

Example of existing noop behavior:
```
   const r = MyModel.build({ id: 1 }, { isNewRecord: false });
   await r.save(); // executes a noop update statement of `id = 1 where id = 1`
   // no error
```

An example of existing noop behavior using the `fields` options in the same manner as the example in the PR.
```
   const r = MyModel.build({ id: 1}, { isNewRecord: false} );
   await r.update({ a: 2 }, { fields: ['b'] }); // sends `""` (literally an empty string) to `sequelize.query`, but the call succeeds.
   // no error
```

AllAwesome497:
Personally, I think I'd rather it throw or do nothing if none of the fields I want to update have changed.   

If no fields are specified (eg: `{ fields: [] }`) in `upsert` we should throw IMO, however if some are specified, however aren't provided in the row's attributes (eg: `{}, { fields: ['a'] }`) we should do one of:
- return the falsy value (I don't recall what the return type of this is)
- fetch the row and return with its value (in the same format that `upsert` returns)

slickmb:
>  or do nothing if none of the fields I want to update have changed

We cannot do "nothing" because the record might not exist yet.  There is an implicit guarantee that after an `upsert`, the record in question will exist.  We either need to throw an error, or attempt the insert.

Before making a final decision, consider two more points.  In the end, I'm aiming for consistency in behavior in sequelize.  I've already demonstrated that my proposed PR is consistent with update and save.  Now, consider the following:

```
            async function nonAtomicUpsert(data, { fields } = {}) {
                await sequelize.transaction(async transaction => {
                    try {
                        sequelize.query(`SAVEPOINT x;`, { transaction });
                        await MyModel.create(data, { transaction });
                    } catch (err) {
                        if (err.name !== "SequelizeUniqueConstraintError") { throw err; }
                        sequelize.query('ROLLBACK TO SAVEPOINT x;', { transaction });
                        await MyModel.update(data, { where: { id: data.id }, fields, transaction});
                    }
                });
            }

            // this first call results in an insert
            await nonAtomicUpsert({ id: 1, a: 2 }, { fields: [ 'b' ]});
            
            // this second call results in a noop.
            await nonAtomicUpsert({ id: 1 a: 2 }, { fields: [ 'b' ]});
```

While the above `nonAtomicUpsert` is non-atomic and is subject to race conditions, it's meant to demonstrate pseudo-logic of an upsert and, at the same time, demonstrate sequelize's current behavior when performing these operations.

In the function, we first ask sequelize to insert the record.  Then, if we hit a duplicate key, we ask sequelize to perform an update.

With fields set to `[ 'b' ]`, and `b` not provided in the data, the first call to this function results in an insert, and the second call to this function results in a noop (the insert fails, and the update is a noop because there were no fields to update).

My PR updates sequelize's `upsert` function to behave exactly how sequelize's insert and update statements behave in the same situation.  It does so by making use of the `ON CONFLICT DO NOTHING`, but in the end, it amounts to the same things... the insert works, the update is a noop because there were no fields to update.

*More importantly*: prior to sequelize 6.0.0, the unit test I'm introducing in this PR *PASSES* (I tested this against commit `b2bccb8`, which is the commit just prior to the introduction of native upserts).  The original behavior of sequelize prior to version 6 is exactly what I'm describing above, and it was only after the introduction of native upserts (`4d9165b`) that my unit test starts failing.  In other words, the introduction of native upserts inadvertently broke the original behavior of sequelize's `upsert`, and my PR is merely restoring that original behavior while still using the native DB constructs for atomic upserts.

As I pointed out above, we cannot just "do nothing", as that ignores the insert operation that is implicit to an upsert.  If we decide that we should just throw an error instead of doing a noop, I fine with that.  I still contend, however, that it would be inconsistent with how sequelize handles this exact same scenario in other places, and also inconsistent with sequelize's own upsert logic prior to version 6.

slickmb:
Actually, I might have to eat my words here....    I was attempt to run a `git bisect` to figure out exactly when my unit test starts failing... and it's currently passing on main **without** my PR   🤦 .   Either my unit test is not working, or the issue I'm attempting to fix has already been fixed on main...

slickmb:
Found the issue.  My unit test is invalid because the `timestamps` option is turned on for the model I'm testing, and thus there are always fields to update.  Setting `timestamps: false` on the model reproduces the error I'm attempting to fix.  I'll fix this in my unit test and re-run my test on `b2bccb8` to see if my assertion that the was sequelize's original upsert behavior still holds true.

slickmb:
Interesting development: After fixing my unit test to properly reproduce the error, and apply that change to my branch... the unit test still fails: this time with an `EmptyResultError`.

https://github.com/sequelize/sequelize/blob/84421d7d738176ee6d0de705c493b145b9488532/lib/dialects/postgres/query.js#L270

The `isInsertQuery` logic looks for any query that contains `INSERT INTO` and a few other criteria, which the native upsert statements now match.

I ran the updated test on sequelize on the commit prior to native upsert change, and... it failed with an `TypeError: Cannot read property 'replace' of undefined` on this line:

https://github.com/sequelize/sequelize/blob/18dcc7ee3c00dbb1337684495dc9f03980301fbe/lib/dialects/postgres/query-generator.js#L355

(`updates.query` is undefined, presumably because the update is a noop).

slickmb:
Maybe failing with an `EmptyResultError` is the correct behavior?  It does ensure that the insert will happen, if it insert is possible, and throws if there were no fields to update, which was one of the potential solutions mentioned in our discussions above.

Still, throwing in the case of an upsert with no updatable fields seems inconsistent with how sequelize handles this scenario everywhere else (e.g. `Model.update`), and I'd argue that falling into the `isInsertQuery` for an upsert query statement may be unintentional in this case.  https://github.com/sequelize/sequelize/blob/84421d7d738176ee6d0de705c493b145b9488532/lib/dialects/postgres/query.js#L269-L271

I retested all of my above assertions (with `Model.save`, `Model.update`, and `Model.create`), and those assertions hold true even with `timestamps: false`.  Sequelize ignores noop updates everywhere else (does not throw), so I'm unsure why a noop `upsert` should behave differently.

I'm going to push up the latest unit test in order to further this discussion, but it will fail at this point due to the `Empty

slickmb:
To keep the discussion going, I pushed up a new commit that

1. updates my unit test to properly reproduce the error I'm tryng to fix (adds `{ timestamps: false }` to my test model)
2. update the `if` block https://github.com/sequelize/sequelize/blob/97f5c5e036acd906e6ae3f6ce2271e138b342f08/lib/dialects/postgres/query.js#L267-L271 so that it excludes upsert statements.  Upsert statements by their very nature ensure that a record exists in an atomic fashion.  Based on the comments above the if block, I think upserts fall outside the intended concern this `if` block is testing for.

My argument is still that because Sequelize treats updates with no updatable fields as a noop (without throwing) everywhere else, it should treat the update portion of an upserts without updatable fields as a noop (without throwing) in the `upsert` function as well.  Or, if we decide to throw in `upsert`, we should also throw for the same scenario when running `update`.

slickmb:
If we do decide that throwing is the right course... I actually like the way it fails in that `if` block (prior to my changing the `if` block, I mean).  It allows the insert portion of the upsert to happen if possible (without failing), but fails (and throws) only if the update portion of the upsert is not possible.  But, again, I'm still feel that's inconsistent with the way Sequelize handles this scenario in the update command.

sdepold:
Hey all,

first of all thanks a bunch for your contribution. It's thoroughly investigated and provided with the mindset of consistency. Well done :)
To everyone who gets to read this in the future, please make sure to read through https://github.com/sequelize/sequelize/pull/13594#issuecomment-956297889. I think it is spot on and I would agree that upset should mimic the pseudo code provided.
@AllAwesome497 do you still object to the overall idea? 

I'll go through the code changes just now however :)

lib/dialects/abstract/query-generator.js
```diff
@@ -185,9 +185,25 @@ class QueryGenerator {
         // If no conflict target columns were specified, use the primary key names from options.upsertKeys
         const conflictKeys = options.upsertKeys.map(attr => this.quoteIdentifier(attr));
         const updateKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=EXCLUDED.${this.quoteIdentifier(attr)}`);
-        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) DO UPDATE SET ${updateKeys.join(',')}`;
```
sdepold, line 188: Funny. I didn't even know you could do that in SQL :D

lib/dialects/abstract/query-generator.js
```diff
@@ -185,9 +185,25 @@ class QueryGenerator {
...
+        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')})`;
+        // if update keys are provided, then apply them here.  if there are no updateKeys provided, then do not try to
+        // do an update.  Instead, fall back to DO NOTHING.
+        onDuplicateKeyUpdate += updateKeys.length ? ` DO UPDATE SET ${updateKeys.join(',')}` : ' DO NOTHING';
```
sdepold, line 191: Could you please use [isEmpty](https://www.geeksforgeeks.org/lodash-_-isempty-method/) here?

lib/dialects/abstract/query-generator.js
```diff
@@ -185,9 +185,25 @@ class QueryGenerator {
...
       } else {
         const valueKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=VALUES(${this.quoteIdentifier(attr)})`);
+        // the rough equivalent to ON CONFLICT DO NOTHING in mysql, etc is ON DUPLICATE KEY UPDATE id = id
+        // So, if no update values were provided, fall back to the identifier columns provided in the upsertKeys array.
+        // This will be the primary key in most cases, but it could be some other constraint.
+        if (!valueKeys.length && options.upsertKeys) {
```
sdepold, line 197: Please use [isEmpty](https://www.geeksforgeeks.org/lodash-_-isempty-method/)

lib/dialects/abstract/query-generator.js
```diff
@@ -185,9 +185,25 @@ class QueryGenerator {
...
+          valueKeys.push(...options.upsertKeys.map(attr => `${this.quoteIdentifier(attr)}=${this.quoteIdentifier(attr)}`));
+        }
+
+        // edge case... but if for some reason there were no valueKeys, and there were also no upsertKeys... then we
+        // can no longer build the requested query without a syntax error.  Let's throw something more graceful here
+        // so the devs know what the problem is.
+        if (!valueKeys.length) {
```
sdepold, line 204: isEmpty

lib/dialects/abstract/query-generator.js
```diff
@@ -185,9 +185,25 @@ class QueryGenerator {
         // If no conflict target columns were specified, use the primary key names from options.upsertKeys
         const conflictKeys = options.upsertKeys.map(attr => this.quoteIdentifier(attr));
         const updateKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=EXCLUDED.${this.quoteIdentifier(attr)}`);
-        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')}) DO UPDATE SET ${updateKeys.join(',')}`;
+        onDuplicateKeyUpdate = ` ON CONFLICT (${conflictKeys.join(',')})`;
+        // if update keys are provided, then apply them here.  if there are no updateKeys provided, then do not try to
+        // do an update.  Instead, fall back to DO NOTHING.
+        onDuplicateKeyUpdate += updateKeys.length ? ` DO UPDATE SET ${updateKeys.join(',')}` : ' DO NOTHING';
       } else {
         const valueKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=VALUES(${this.quoteIdentifier(attr)})`);
```
sdepold, line 193: I have a little bit of a hard time understanding what updateOnDuplicate is and what it is used for.

lib/dialects/mssql/query.js
```diff
@@ -369,6 +374,7 @@ class Query extends AbstractQuery {
   }
 
   handleInsertQuery(results, metaData) {
+
```
sdepold, line 377: Can we remove this empty line?

test/integration/model/upsert.test.js
```diff
@@ -311,6 +311,31 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         clock.restore();
       });
 
+      it('falls back to a noop if no update values are found in the upsert data', async function() {
+
```
sdepold, line 315: empty line--

test/integration/model/upsert.test.js
```diff
@@ -311,6 +311,31 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+        const User = this.sequelize.define('user', {
+          username: DataTypes.STRING,
+          baz: {
+            type: DataTypes.STRING,
+            field: 'zab',
```
sdepold, line 320: this is melting my brain. 🙈 Can you maybe introduce some real-world property names and values?

sdepold (changes requested):
I think this is quite a nice change. Let's see what @AllAwesome497 says. Thanks :)

lib/dialects/abstract/query-generator.js
slickmb: It contains the list of fields (just the field names, no values) that will be updated when an insert fails with a duplicate error.

So if a developer called something like this:

```
const [ instance ] = await MyModel.upsert({ id: 1, email: 'xxx@yyy.zzz', phone: '123456' }, { fields: [ 'phone' ], returning: true });
```

The `upsert` query-interface function get's called with something like this:

```
await upsert('MyModel', { id: 1, email: 'xxx@yyy.zzz', phone: '123456' }, { phone: '123456' }, { where: { id: 1 } }, options);
```

It gets set by the abstract query interface here: https://github.com/sequelize/sequelize/blob/199b632b021830f9d09210fd7430045710638631/lib/dialects/abstract/query-interface.js#L775

The `upsert` function overridden by the mysql interface here: https://github.com/sequelize/sequelize/blob/7006eb4ebbfd38cc055873f56c6593f633f28bad/lib/dialects/mysql/query-interface.js#L48

So by the time it gets to this code, the in the `query-generator`, the `updateOnDuplicate` option is set to `[ 'phone' ]`.

I'm adding a comment in the code as I respond to the rest of your code review comments.

test/integration/model/upsert.test.js
slickmb: switched `baz` to `email`.

lib/dialects/abstract/query-generator.js
AllAwesome497: The terminology (and logic for that matter) for upserts (being `Model.bulkCreate` and `Model.upsert`) in general isn't very consistent I've found.

`updateOnDuplicate` for `Model.bulkCreate` serves the same purpose as `firelds` for `Model.upsert`.

There's additionally an option for `Model.bulkCreate`, `ignoreDuplicates` which translates to `ON CONFLICT DO NOTHING`, similar to what this pr will change `upsert` to do implicitly if none of the specified fields are provided.

AllAwesome497 (changes requested):
I think I misunderstood this change before, looking at it now the change makes a lot of sense.  My only request is to specify in the docs for `upsert` that if none of `fields` are in the object provided, it'll implicitly ignore duplicates.

slickmb:
Updated the inline documentation for `Model.upsert`.  Is there anywhere else I should update?

PR #13600, opened 2021-10-26
fix: queryinterface does not provide existing fkeys when schema is used

denizkanmaz:
### Pull Request check-list

### Description of change

`queryinterface.getForeignKeyReferencesForTable` (Postgres) does not provide existing foreign keys (when schema is used) therefore the ORM creates the same foreign key for every `sync` execution. Eventually, this causes to have multiple foreign keys for the same relation.

An example code block reproduces the issue can be found [here](https://stackoverflow.com/questions/64806845/sequelize-sync-duplicates-existing-foreign-key-associations).

This issue reproduced with the followings:
Postgres: 10.5
Node: 14.5
Dependencies:
pg: 8.5
pg-hstore: 2.3.3

The original PR: #12833


sdepold:
Why are we doing this? What's the problem with passing the actual table name to the function instead of an object containing the table name?

sdepold:
Never mind. I just found the typescript definition

sdepold (approved):

liv1n9:
```
sequelize: 6.35.2
pg: 8.11.3
```
I still get this error
image

liv1n9:
> ```
> sequelize: 6.35.2
> pg: 8.11.3
> ```
> 
> I still get this error after syncing database image
> 
> I'm using NestJS + sequelize-typescript. My config: image

I find the cause of it. It is due to the `schema` property, which is `"public"` in my case. After I remove it, the duplicate foreign keys and constraints are disapeared

PR #13609, opened 2021-10-28
fix(dialect-mysql): custom msg bug fix for duplicate entry on MySQL 8.0

Puvipavan:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ ] Have you added new tests to prevent regressions?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Did you update the typescript typings accordingly (if applicable)?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)?

### Description of change

This is a fix for the issue #13601 

In MySQL Version 8.0, index name is displayed after appending the schema name. 
Sample Error on MySQL 5.7
`#1062 - Duplicate entry '1234567890' for key 'users_phone_unique'`

Sample Error on MySQL 8.0
`#1062 - Duplicate entry '1234567890' for key 'users.users_phone_unique'`

Therefore I added a small change which doesn't hurt anything regardless of MySQL Version.

Issue #13601, opened 2021-10-27
Unique Field DB error custom message feature broken.


Puvipavan:
## Issue Creation Checklist

- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description
One of the field in my model which has custom error message for unique field.

```js
phone: {
  type: DataTypes.STRING(13),
    allowNull: true,
    unique: {
      args: true,
      msg: "Phone already in use."
  },
  validate: {
    len: {
      args: [10, 12],
      msg: "Phone number can contain 10-12 numbers only"
    },
    isNumeric: {
      args: true,
      msg: "Phone number can contain numbers only"
    },
    checkPhoneNull(value) {
      if (value == null && this.email == null) {
        throw new Error('Please Enter valid Email or Phone Number');
      }
    }
  }
}
```

### What do you expect to happen?
It should throw the error message: "Phone already in use."

### What is actually happening?
It throws the error message: "users.users_phone_unique must be unique"

### Additional context
I started developing my project after several months(Maybe 4 or 5). Previously it worked fine but in the middle I think I have upgraded or changed my Database from MariaDB to MySQL. I tried updating Sequelize but it didn't work. This is the error message I got for the duplicate entry.
```
#1062 - Duplicate entry '0000000000' for key 'users.users_phone_unique'
```
The bad regex at following line: https://github.com/sequelize/sequelize/blob/eeb6a8fbeb6549be038f2dbb0eefb414c7450653/lib/dialects/mysql/query.js#L215
yields to the wrong `fieldKey ` at following line: https://github.com/sequelize/sequelize/blob/eeb6a8fbeb6549be038f2dbb0eefb414c7450653/lib/dialects/mysql/query.js#L219 which sets to unique key to false at line no 221. Also it's passing same `fieldKey` to the following line(): https://github.com/sequelize/sequelize/blob/eeb6a8fbeb6549be038f2dbb0eefb414c7450653/lib/dialects/mysql/query.js#L233

So the expected `fieldKey` should be `users_phone_unique` but the actual `filedKey` is `users.users_phone_unique`

I'm not expert at this, But I will try to give a PR by tomorrow night. Going to sleep, It's already morning 6:24 AM :(

### Environment
- Sequelize version: 6.8.0
- Node.js version: 10.19.0
- MySQL Server version(Community Server): 8.0.25

### How does this problem relate to dialects?

[...]

PR #13610, opened 2021-10-29
test(mysql): add test-support for MySQL 8

sdepold:
### Pull Request check-list

### Description of change

This PR implementing the changes for #13601 and #13609 by introducing a separate build using MySQL 8.0. It also fixes all relevant details that have been introduced in v8 such as migrating from GeomFromText to ST_GeomFromText and rejection of descending indices in MyISAM databases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,6 +103,14 @@ jobs:
             image: mysql:5.7
             dialect: mysql
             node-version: 12
+          - name: MySQL 8.0
+            image: mysql:8.0
+            dialect: mysql
+            node-version: 10
+          - name: MySQL 8.0
+            image: mysql:8.0
+            dialect: mysql
+            node-version: 12
           - name: MariaDB 10.3
             image: mariadb:10.3
             dialect: mariadb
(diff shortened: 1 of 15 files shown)


sdepold:
@Puvipavan I enabled a test workflow for MySQL 8 and there are a couple of failing tests. As you can see, I have copied your fix to this PR already. I'll have a look into the other issues in the next couple of days.

Puvipavan:
Thank you @sdepold. Main reason for the most of the error is using deprecated function `GeomFromText` for MySQL 8.0(https://dev.mysql.com/doc/refman/8.0/en/mysql-nutshell.html). We have to use `ST_GeomFromText` instead of `GeomFromText`. Fortunately `ST_GeomFromText` is supported by 5.6/5.7 too(https://dev.mysql.com/doc/refman/5.6/en/populating-spatial-columns.html). So I had to replace `GeomFromText` with `ST_GeomFromText` on the file `lib/data-types.js`. Then I ran the integration test for MySQL 8.0 and errors were reduced from 19 to 3. I tested with 5.7 too and all the tests were passed. I will investigate the remaining errors when I get some time.

Updated my PR: https://github.com/sequelize/sequelize/pull/13609/

PS: I tested both versions by editing the docker image version on the docker compose file at `dev/mysql/5.7/docker-compose.yml`

sdepold:
> Fortunately `ST_GeomFromText` is supported by 5.6/5.7

This is exactly what I just looked into and pushed :)

sdepold:
Woop. Green. @Puvipavan I have copied all your lovely commits into this one now

Puvipavan:
Wow. Thank you @sdepold

Keimeno (approved):

PR #13632, opened 2021-11-06
meta: persist lockfile

Keimeno:
### Pull Request Checklist

### Description Of Change

Persists a yarn lock-file. 
This allows us to create releases that are more transparent about the dependencies that were installed. It also reduces the risks we could get from dependencies with breaking changes.


WikiRik:
Why move to yarn for this? npm also supports a lock file, but it was removed in https://github.com/sequelize/sequelize/commit/22d35953da85869cf742d223cd49498c99027d0c

sdepold:
Because it's much faster I guess. Don't we have to ignore the file somehow when building the npm package?

sdepold:
Or do we plan to ship the lockfile?

Keimeno:
I don't think there's much of a reason to ship the file as well.

Also, yarn has resolutions, which allows us to easily overwrite sub-dependency versions

sdepold (approved):

PR #13633, opened 2021-11-07
Fix MariaDB 10.5 JSON

### LINKED ISSUE #12583: Sequelize crash - mariaDB 10.5.4 & JSON field

rickbergfalk:
### Description Of Change

Fixes JSON field handling for MariaDB 10.5.

No new tests added, but does add MariaDB 10.5 to test matrix

Closes #12583
(Also relevant: Issue #13027. PR #13273)


rickbergfalk:
🙌 Thanks @sdepold for merging and maintaining!

cwoliveira:
ref com https://github.com/fitinsur/fitinsur-sompo/issues/2041

PR #13637, opened 2021-11-08
fix: expect result is null but got zero

r253hmdryou:
fix: expect result is 0 but got NaN #13422

This is wrong
Should return Null instead of Zero or NaN

```
MariaDB [test]> SELECT MAX(users.id) FROM users;
+---------------+
| MAX(users.id) |
+---------------+
|          NULL |
+---------------+
1 row in set (0.010 sec)

MariaDB [test]> INSERT INTO users(id, name) VALUE(0, "name");
Query OK, 1 row affected (0.014 sec)

MariaDB [test]> SELECT MAX(users.id) FROM users;
+---------------+
| MAX(users.id) |
+---------------+
|             0 |
+---------------+
1 row in set (0.010 sec)
```

### DIFF
--- a/test/integration/model/notExist.test.js
+++ b/test/integration/model/notExist.test.js
@@ -28,7 +28,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
   });
 
   describe('max', () => {
-    it('should type exist', async function() {
+    it('type A to C should exist', async function() {
       await expect(this.Order.sum('sequence', { where: { type: 'A' } })).to.eventually.be.equal(10);
       await expect(this.Order.max('sequence', { where: { type: 'A' } })).to.eventually.be.equal(4);
       await expect(this.Order.min('sequence', { where: { type: 'A' } })).to.eventually.be.equal(1);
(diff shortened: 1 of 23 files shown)


test/integration/model/notExist.test.js
```diff
@@ -41,18 +42,22 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
       await expect(this.Order.max('sequence', { where: { type: 'C' } })).to.eventually.be.equal(0);
       await expect(this.Order.min('sequence', { where: { type: 'C' } })).to.eventually.be.equal(0);
-
-      // DataTypes.DECIMAL or DataTypes.FLOAT: previous and PR#13422 both use `to.eventually.be.equal(0)`
       await expect(this.Order.sum('amount', { where: { type: 'C' } })).to.eventually.be.equal(0);
       await expect(this.Order.max('amount', { where: { type: 'C' } })).to.eventually.be.equal(0);
       await expect(this.Order.min('amount', { where: { type: 'C' } })).to.eventually.be.equal(0);
     });
+
+    it('should type not exist', async function() {
```
sdepold, line 54: Could you please give that a more sensible title? Must have slipped through earlier :)

sdepold (approved):
Nice. Thank you!

sdepold (changes requested):
Hm. Seems to be broken everywhere somehow

test/integration/model/notExist.test.js
r253hmdryou: I'm sorry, I'm non-native so I can't think of proper English...

r253hmdryou:
@sdepold 
I was misunderstanding
This time it should move firmly

sdepold (approved):

PR #13640, opened 2021-11-08
fix(types): consistent create association values. Fixes #12463

### LINKED ISSUE #12463: Inconsistent typing of Create functions
## Issue Description

`BelongsToCreateAssociationMixin` typing does not seem consistent with standard `Create` types.

https://github.com/sequelize/sequelize/blob/9a45509e3f4b23b957a195190dd989ea33cae8d8/types/lib/associations/belongs-to.d.ts#L119 Seems to expect an object with an implicit key signature of string while https://github.com/sequelize/sequelize/blob/9a45509e3f4b23b957a195190dd989ea33cae8d8/types/lib/model.d.ts#L1979 expects a standard JavaScript object. 

If we had:
```ts
interface UserData {
  id: number;
  name: string;
}

interface ProjectData {
  id: number;
}

class Project extends Model {
  id!: number;
  
  public addUser: BelongsToCreateAssociationMixin<User>;
}

// Project.init etc...

const newProjectData : ProjectData = { id: 1};
const newProject = Project.create(newProjectData); // Works

const newUser: UserData = { id: 1, name: "John"};
newProject.addUser(newUser); // Throws type error
```

Since this behavior works for the standard create function, we would expect no errors from the TSC. 

`Project.addUser(newUser); ` throws the following error: `Argument of type 'UserData' is not assignable to parameter of type '{ [attribute: string]: unknown; }'.
  Index signature is missing in type 'UserData '.`

### Additional context

This error makes sense when looking at the type definitions files. A simple change of https://github.com/sequelize/sequelize/blob/9a45509e3f4b23b957a195190dd989ea33cae8d8/types/lib/associations/belongs-to.d.ts#L120 from ` values?: { [attribute: string]: unknown },` to ` values?: object,` (like it is on the standard `create()` would fix this. For consistency all create functions should use the same type signatures. 

### Environment

- Sequelize version: 5.xx  BUT THE CODE I LINKED IS FROM V6
- If TypeScript related: TypeScript version: 3.9

- [] Yes, I have the time and I know how to start.

WikiRik:
### Pull Request Checklist

- [] Did you update the typescript typings accordingly (if applicable)?

### Description Of Change

Change association create type signatures like proposed in #12463. Closes #12463 

### Todos

- [ ] Add regression tests (if needed)
- [ ] Add new tests (if needed)
- [ ] Check documentation if changes are needed
- [ ] Hear back from issue submitter

### DIFF
--- a/types/lib/associations/belongs-to-many.d.ts
+++ b/types/lib/associations/belongs-to-many.d.ts
@@ -304,7 +304,7 @@ export interface BelongsToManyCreateAssociationMixinOptions extends CreateOption
  * @see Instance
  */
 export type BelongsToManyCreateAssociationMixin<TModel> = (
-  values?: { [attribute: string]: unknown },
+  values?: object,
   options?: BelongsToManyCreateAssociationMixinOptions
 ) => Promise<TModel>;
 
--- a/types/lib/associations/belongs-to.d.ts
+++ b/types/lib/associations/belongs-to.d.ts
@@ -117,7 +117,7 @@ export interface BelongsToCreateAssociationMixinOptions
  * @see Instance
  */
 export type BelongsToCreateAssociationMixin<TModel> = (
-  values?: { [attribute: string]: unknown },
+  values?: object,
   options?: BelongsToCreateAssociationMixinOptions
 ) => Promise<TModel>;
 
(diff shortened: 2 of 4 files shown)


sdepold:
What's missing here?

WikiRik:
I think https://github.com/sequelize/sequelize/pull/13678 and https://github.com/sequelize/sequelize/pull/13685 already fixed this issue, so I will close this

PR #13647, opened 2021-11-10
fix(sqlite): fix individual column unique with unique multi-col index

### LINKED ISSUE #12488: SQLite: multi-col unique constraints becoming individual column unique constraints
## Issue Description

Using SQLite and Sequelize 6.3.1:

If you create a table with a multi-column unique constraint, then remove an unrelated column, you are left with multiple single-column unique constraints.

I'm not entirely sure, but I think it might have been introduced in this change: https://github.com/sequelize/sequelize/pull/12420

### What are you doing?

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/92

I would expect the original constraint to remain, or prefer the old behavior of constraints being removed and needing to be re-added. New behavior is a breaking change.

Unique constraints are being added that are not intended.

### Environment

- Sequelize version: 6.3.1
- Node.js version: 12.16.1

- [x] I think this problem happens only for the following dialect(s): SQLite

- [x] Maybe, I don't know what the preferred solution is

rickbergfalk:
### Pull Request Checklist

### Description Of Change

This rolls back some of the changes originally introduced in #12420.

Multi-column unique indexes in SQLite are being converted to single column unique indexes when the table is rebuilt. This is problematic, as it alters the underlying data model.

Closes #12488


sdepold:
I'm going to take this PR to our upcoming monthly sync. Feel free to join and talk about it as well: https://www.meetup.com/eBay-Europe-Technology/events/282125063/

sdepold:
Does it mean you cannot have a unique index across multiple columns in sqlite?

rickbergfalk:
@sdepold sort of. You can create a table with a unique index with multiple columns, but you cannot perform any alteration of it that causes a table rebuild in SQLite due to SQLites limited ALTER support. 

For example, let's say you create a table with columns A, B, and C, and you put an index, UNIQUE (A, B). If you drop column C, a new table would be created, using `describeTable`'s output, which incorrectly identifies column A being unique, and column B being unique.

The definition of the table is altered, moving from 

Columns: A, B, C
Indexes: 
* UNIQUE (A, B)

to

Columns: A, B
Indexes:
* UNIQUE (A)
* UNIQUE (B)

If someone has data in this table, it is likely that column A and/or column B are not unique on their own, causing this operation to fail.

Prior to #12420, the constraints would be lost entirely, requiring the sequelize consumer to re-add constraints as they see fit. While not convenient, it is functional. 

A reason **NOT** to accept this PR is that it technically is a breaking change for folks that **only** use single-column indexes and rely on today's behavior. The original addition did add some value, but also broke things.

rickbergfalk:
I'm not certain but this issue looks like it could be related #13653

sdepold:
Thanks a bunch for that explanation! Very helpful. I'll ponder a bit more about this and get back to you :)

test/integration/model/sync.test.js
```diff
@@ -215,16 +215,12 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           const results = await this.sequelize.getQueryInterface().showIndex(User.getTableName());
           if (dialect === 'sqlite') {
             // SQLite doesn't treat primary key as index
-            // However it does create an extra "autoindex", except primary == false
```
sdepold, line 218: Why does that change?

test/integration/query-interface/changeColumn.test.js
```diff
@@ -235,56 +235,7 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
-            type: DataTypes.STRING
-          },
-          email: {
-            allowNull: false,
-            unique: true,
-            type: DataTypes.STRING
-          }
-        });
-
-        await this.queryInterface.addColumn('Foos', 'phone', {
```
sdepold, line 260: so before adding a column everything was previously just fine. but after the column, it would have lost the unique index somehow?

test/integration/query-interface/changeColumn.test.js
```diff
@@ -235,56 +235,7 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
     }
 
     if (dialect === 'sqlite') {
-      it('should not remove unique constraints when adding or modifying columns', async function() {
```
sdepold, line 238: What am I missing. This test case doesn'T seem to create a multi column unique index?

sdepold (commented):
My prime question is: How can a change in describeTable have any impact on manipulating the table structure. Shouldn't we keep that particular logic and rather look into the modifying methods such as addColumn, removeColumn, etc.?

sdepold:
What do you think about actually rendering a warning when you try to change a table that has a multi column unique constraint?

test/integration/model/sync.test.js
rickbergfalk: The PR that introduced this bug added this to the test to make it pass https://github.com/sequelize/sequelize/pull/12420/files#diff-81b952d2bbfce09ba3fc0712ea8248f597d3d3421d819a1c69eb06f68fcf318eR218

test/integration/query-interface/changeColumn.test.js
rickbergfalk: Previously this test ensured single-column unique indexes persisted with a table modification. This PR rolls back the changes that make this test work.

rickbergfalk:
@sdepold to maybe clear up some confusion - this PR reverts some of the changes and tests added in https://github.com/sequelize/sequelize/pull/12420. Prior to #12420, index structure was not kept when modifying tables in SQLite because there are limitations in SQLite that prevent persisting constraints. #12420 then took a shot at persisting indexes, but does so incorrectly when they are for multi-column indexes.

As for why we need to alter describe table in this PR - it is not accurately describing the table. I'm not too familiar with the nuances of Sequelize code, but I don't think the multi-column unique indexes are supposed to equate to a unique boolean on individual columns when the table is described.

> What do you think about actually rendering a warning when you try to change a table that has a multi column unique constraint?

Would this warning be in place of rolling back the problematic bits of #12420? 

I think a warning would be helpful if the original behavior was restored where indexes were lost. A warning while keeping the incorrect behavior would be frustrating otherwise.

## A potential alternate solution

I don't know if this would be too hacky, but something we could try in order to keep indexes in SQLite is to obtain the current index SQL statements for the table being altered prior to the alteration with the following query:

```sql
SELECT sql
FROM sqlite_master
WHERE 
  type = 'index'
  AND tbl_name = 'target_table_name'
```

After the table is altered (and all indexes are lost), we could iterate over the index SQL statements, and apply them, ignoring errors. 

The reason we'd need to ignore errors is that a column could be removed that was part of a multi-column index, and that particular index query would fail.

lib/dialects/sqlite/query-interface.js
```diff
@@ -210,7 +210,7 @@ class SQLiteQueryInterface extends QueryInterface {
       }
       for (const index of indexes) {
         for (const field of index.fields) {
-          if (index.unique !== undefined) {
+          if (index.unique !== undefined && index.fields.length === 1) {
```
rickbergfalk, line 213: Updated this condition to only set a column to unique if the index contains 1 column. This allows existing behavior to stay in place for single column indexes.

Multi-column indexes are NOT persisted when tables are rebuilt. I'm not sure how this is preferred in Sequelize code base.

test/integration/model/sync.test.js
rickbergfalk: This test still needs to change with my fix. I don't think it should have been changed in #12420.

test/integration/query-interface/changeColumn.test.js
```diff
@@ -284,6 +284,62 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
         expect(table.name.unique).to.equal(true, '(2) name column should remain unique');
       });
 
+      it('should not add single column unique constraints from multi-column constraint when adding or modifying columns', async function() {
```
rickbergfalk, line 287: Added a new test to ensure multi-column indexes to not turn into single column indexes.

rickbergfalk:
@sdepold This is ready for another review. There are still quirks with SQLite implementation, but this at least addresses #12488.

WikiRik:
@sdepold ping. This coment probably explains the questions you had earlier; https://github.com/sequelize/sequelize/pull/13647#issuecomment-985953590

Issue #13638, opened 2021-11-08
offset doesn't work for sqlserver 2008 when using conditions on includes


sschwenker:
When trying to FindAll items with an include statement that has conditions on it, the query generator neglects to write the join statements as part of the inner query.  I've figured out where to fix it and I'm now creating this bug and will be submitting a PR shortly for a fix.

When using sqlserver 2008, when you call findAll while using the include and offset functionality, the query generator generates a query with a sub query to select the offset objects.  On the inner query the generator doesn't include the joined tables, but it includes the where conditions to query against the joined tables.  I have modified the inner query to include the joined tables so that it can appropriately filter the rows to be included in the offset list.

WikiRik:
Since the associated PR is merged, I will close this issue. For others reading it in the future, it will be part of the release after 6.9.0

PR #13651, opened 2021-11-10
Bugfix 13638

sschwenker:
### Pull Request Checklist

- [NA] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [NA] Did you update the typescript typings accordingly (if applicable)?

### Description Of Change

This PR is an implementation to fix issue #13638 . I updated the code, fixed affected unit test and added a new unit test.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -12,9 +12,9 @@ jobs:
         id: stale
         with:
           stale-issue-label: "stale"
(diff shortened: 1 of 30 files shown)


sschwenker:
Hey.  Are we able to re-run the build/tests?  I didn't touch any of the mysql code.

WikiRik:
Yeah, it seems that this test fails on main as well sometimes. Seems to be related to MySQL 8 and not Node.

sschwenker:
However.  I'm going to make an adjustment to the code to only generate the extended query if there is an include.  If there is no includes, then it's not needed and it can slow down the query.  So it will continue to generate the older query.

sschwenker:
Ok.  Done.

PR #13670, opened 2021-11-16
feat: option for attributes having dotnotation

SutharMukesh:
### Pull Request Checklist

### Description Of Change

The select query raises ambiguity errors when I join two tables that have the same attribute/column names which contain a dot(".") in them.

This PR fixes it by provides a new option; 
`options.dotnotation = true`

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1452,7 +1452,7 @@ class QueryGenerator {
           ? this.quoteAttribute(attr, options.model)
           : this.escape(attr);
       }
-      if (!_.isEmpty(options.include) && !attr.includes('.') && addTable) {
+      if (!_.isEmpty(options.include) && (!attr.includes('.') || options.dotnotation) && addTable) {
         attr = `${mainTableAs}.${attr}`;
       }
 
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1452,7 +1452,7 @@ class QueryGenerator {
           ? this.quoteAttribute(attr, options.model)
           : this.escape(attr);
       }
-      if (!_.isEmpty(options.include) && (!attr.includes('.') || options.dotnotation) && addTable) {
+      if (!_.isEmpty(options.include) && (!attr.includes('.') || options.dotNotation) && addTable) {
         attr = `${mainTableAs}.${attr}`;
       }
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -1672,6 +1672,7 @@ class Model {
    * @param  {object}                                                    [options.having] Having options
    * @param  {string}                                                    [options.searchPath=DEFAULT] An optional parameter to specify the schema search_path (Postgres only)
    * @param  {boolean|Error}                                             [options.rejectOnEmpty=false] Throws an error when no records found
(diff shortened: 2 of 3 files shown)


WikiRik:
Hi, thanks for your PR. Unfortunately I do not see any tests or changes to the documentation. Could you add those?

SutharMukesh:
oops, will do it and update here

SutharMukesh:
Hey @WikiRik, the test cases are ready.
But I'm not sure where to document this change, can you please point me in a direction.
Thanks :)

WikiRik:
That's a good question. I also wouldn't really know off the top of my head but I expect that other users will want to know about this feature as well. Maybe it's good if you look around the documentation and see where you would expect this

SutharMukesh:
I've documented my changes.
@WikiRik I must say, the repo is well documented and maintained 🙌

lib/dialects/abstract/query-generator.js
```diff
@@ -1452,7 +1452,7 @@ class QueryGenerator {
           ? this.quoteAttribute(attr, options.model)
           : this.escape(attr);
       }
-      if (!_.isEmpty(options.include) && !attr.includes('.') && addTable) {
+      if (!_.isEmpty(options.include) && (!attr.includes('.') || options.dotnotation) && addTable) {
```
sdepold, line 1455: Can you please rename it to `dotNotation`?

lib/model.js
```diff
@@ -1672,6 +1672,7 @@ class Model {
    * @param  {object}                                                    [options.having] Having options
    * @param  {string}                                                    [options.searchPath=DEFAULT] An optional parameter to specify the schema search_path (Postgres only)
    * @param  {boolean|Error}                                             [options.rejectOnEmpty=false] Throws an error when no records found
+   * @param  {boolean}                                                   [options.dotnotation] Allows including tables having the same attribute/column names - which have a dot in them.
```
sdepold, line 1675: I'm confused. Are we allowing dots in column names or that multiple tables have the same column name?

sdepold: I don't think we are talking about conflicting / overlapping properties but allowing dot notation in columns?

sdepold (changes requested):

sdepold:
Are people really using dot notation in column names?

lib/model.js
SutharMukesh: > I'm confused. Are we allowing dots in column names or that multiple tables have the same column name?

I just realized the commit message is confusing, my bad

Sequelize does allow dot notation in columns 
but it raises errors when two tables have the same column name which includes dot in them, 
let's say 
`table1 (column1, "column2.label")` and `table2 (column1, "column2.label")`

then sequelize is generating a query like 
```sql
SELECT table1.column1,
       'column2.label',
       table2.column1,
       'column2.label'
FROM table1 AS "table1"
         LEFT OUTER JOIN table2 AS "table2" ON ....         
```
which raises ambiguity errors 
so having this `dotNotation` option will allow sequelize to concat the table_name even if a column includes a dot.

lib/dialects/abstract/query-generator.js
SutharMukesh: Sure

sdepold (approved):

Issue #13672, opened 2021-11-16
Postgres dialect removes extra options from connection string and makes SSL hard to configure


rafiss:
## Issue Creation Checklist

- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

### Is your feature request related to a problem? Please describe.

I want to connect to my database using a connection string like this:
```
postgresql://root:password@localhost:5432/my_db?sslmode=verify-full&sslrootcert=%2Ftmp%2Fcerts%2Fca.crt&options=--search_path%3dmy_schema
```

However, it doesn't work, and I get a message from the server saying that I did not connect with the proper SSL configuration. Note that the same connection string works for me if I connect with https://github.com/brianc/node-postgres directly.

If I disable SSL in the server, and just try connecting with
```
postgresql://root@localhost:5432/my_db?options=--search_path%3dmy_schema
```
then I see that the `search_path` option does not get passed correctly.

### Describe the solution you'd like

I'd like to be able to use my connection string as-is.

node-postgres already includes a package that can handle the connection string and get all the SSL options from it. See https://github.com/brianc/node-postgres/blob/947ccee346f0d598e135548e1e4936a9a008fc6f/packages/pg-connection-string/README.md

However, currently Sequelize doesn't support this because it has different URL parsing here: https://github.com/sequelize/sequelize/blob/47c4494968422585bf265063925d1662ffcd4173/lib/sequelize.js#L185

Also, this part of the postgresql dialect only allows a subset of options to be passed in `dialectOptions`: https://github.com/sequelize/sequelize/blob/3ca085db318201fa59422a2ce191bcf76e5f37dc/lib/dialects/postgres/connection-manager.js#L96

I'd like to change that to allow the `options` parameter to be passed in.

### Why should this be in Sequelize

It would simplify the process of connecting to Postgres in two ways:
1. easier to pass in additional session options in the URL
2. easier to configure SSL

### Describe alternatives/workarounds you've considered

The workaround for the SSL configuration is to pass it in the options like this:
```
{
  dialect: "postgres",
  username: "root",
  host: "localhost",
  port: 5432,
  database: "my_db"
  dialectOptions: {
    ssl: {
      ca: fs.readFileSync("/tmp/certs/ca.crt").toString())
    },
  },
  logging: false,
}
```

However, as far as I can see, there is *no* workaround available for passing in the session variable `options` at connection time.

### Additional context

[...]

PR #13673, opened 2021-11-16
feat(postgresql): easier SSL config and options param support

rafiss:
### Pull Request Checklist

I don't think docs are needed

### Description Of Change

This commit uses the pg_connection_string package to parse the
connection string if the dialect is postgresql. This is helpful because
it automatically handles reading SSL certs that are specified in the
connection string.

As part of this, support was added for the `options` URL parameter,
which allows arbitrary session variables to be configured in the
connection string.

fixes https://github.com/sequelize/sequelize/issues/13672

### Todos

None

### DIFF
--- a/lib/sequelize.js
+++ b/lib/sequelize.js
@@ -2,6 +2,7 @@
 
 const url = require('url');
 const path = require('path');
+const pgConnectionString = require('pg-connection-string');
 const retry = require('retry-as-promised');
 const _ = require('lodash');
 
@@ -231,6 +232,12 @@ class Sequelize {
             }
           }
         }
+      }
+
+      // For postgres, we can use this helper to load certs directly from the
+      // connection string.
+      if (options.dialect === 'postgres' || options.dialect === 'postgresql') {
+        Object.assign(options.dialectOptions, pgConnectionString.parse(arguments[0]));
       }
     } else {
       // new Sequelize(database, username, password, { ... options })
--- a/test/integration/dialects/postgres/connection-manager.test.js
+++ b/test/integration/dialects/postgres/connection-manager.test.js
@@ -47,6 +47,13 @@ if (dialect.match(/^postgres/)) {
       const error = await sequelize.query('select pg_sleep(2)').catch(e => e);
       expect(error.message).to.equal('Query read timeout');
     });
+
+    it('should allow overriding session variables through the `options` param', async () => {
+      const sequelize = Support.createSequelizeInstance({ dialectOptions: { options: '-csearch_path=abc' } });
+      const result = await sequelize.query('SHOW search_path');
+      expect(result[0].search_path).to.equal('abc');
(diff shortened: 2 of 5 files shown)


lib/sequelize.js
```diff
@@ -232,6 +233,12 @@ class Sequelize {
           }
         }
       }
+
+      // For postgres, we can use this helper to load certs directly from the
+      // connection string.
+      if (options.dialect === 'postgres' || options.dialect === 'postgresql') {
+        Object.assign(options.dialectOptions, pgConnectionString.parse(arguments[0]));
```
sdepold, line 240: Just a thought really: isn't args[0] === database?

test/integration/dialects/postgres/connection-manager.test.js
```diff
@@ -47,6 +47,13 @@ if (dialect.match(/^postgres/)) {
       const error = await sequelize.query('select pg_sleep(2)').catch(e => e);
       expect(error.message).to.equal('Query read timeout');
     });
+
+    it('should allow overriding session variables through the `options` param', async () => {
```
sdepold, line 51: I somehow expected a test that would use a database url as you proposed in the description

rafiss: I would like to add a test like that - but I wasn't sure where before. i just found it in `test/integration/sequelize.test.js` now. I'll add a test for `options` there, but I'm having trouble finding any sort of SSL testing anywhere.

lib/sequelize.js
rafiss: I don't think so - as per https://github.com/sequelize/sequelize/blob/47c4494968422585bf265063925d1662ffcd4173/lib/sequelize.js#L185

it is treated as a connection string in this conditional branch.

sdepold: Yeah I found that line confusing as well :) anyway. I don't mind either way

sdepold (approved):

rafiss:
Thanks for the review! It looks like the test failures are in MSSQL and MySQL, but they seem unrelated to my change.

rafiss:
Thanks for releasing this! here's a little example that shows how much easier it is to use the connection string without needing to parse it: https://github.com/cockroachdb/examples-orms/pull/144/files#diff-0d2dbddd194015ba8186248f2006b9102c626cb8d27f60fdeafec56dd5d7ccb0

sdepold:
Nice one!

lib/sequelize.js
WikiRik: Can you check if this still occurs on `main`/the latest v7 alpha? If so, can you create an issue for it?

ankon: https://github.com/sequelize/sequelize/issues/15823

Took a while. Issue creation has a lot of requirements :)

PR #13684, opened 2021-11-19
fix(types): add instance member declaration

joe223:
### Pull Request Checklist

### Todos

https://github.com/sequelize/sequelize/blob/51bc05ae806ee11d36764bdafb4e4645d00e93be/lib/sequelize.js#L1257-L1264
These methods are not only static member of Sequelize. So I made this PR and trying to keep the definition and behavior consistent.

### DIFF
--- a/types/lib/sequelize.d.ts
+++ b/types/lib/sequelize.d.ts
@@ -426,6 +426,7 @@ export class Sequelize extends Hooks {
    * @param args All further arguments will be passed as arguments to the function
    */
   public static fn: typeof fn;
+  public fn: typeof fn;
 
   /**
    * Creates a object representing a column in the DB. This is often useful in conjunction with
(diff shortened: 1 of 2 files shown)


sdepold (changes requested):
Could you please add a test to the types area of the repo? 
I think this would be a good place: https://github.com/sequelize/sequelize/blob/main/types/test/sequelize.ts

Prior to your change you should not have been able to do `sequelize.cast(1, 'string')` for example.

joe223:
> Could you please add a test to the types area of the repo? I think this would be a good place: https://github.com/sequelize/sequelize/blob/main/types/test/sequelize.ts
> 
> Prior to your change you should not have been able to do `sequelize.cast(1, 'string')` for example.

I'll add tests for `sequelize.cast` `sequelize.fn` and so on, to ensure these instance members do exist after types changed.

PR #13685, opened 2021-11-19
fix: wrong interface used within mixin

mmahalwy:
### Pull Request Checklist

### Description Of Change
Fixes a typo from here: https://github.com/sequelize/sequelize/pull/13678

### Todos

- [ ] 
- [ ] 
- [ ]

### DIFF
--- a/types/test/typescriptDocs/ModelInit.ts
+++ b/types/test/typescriptDocs/ModelInit.ts
@@ -51,6 +51,7 @@ interface ProjectAttributes {
   id: number;
   ownerId: number;
   name: string;
+  description?: string;
 }
 
 interface ProjectCreationAttributes extends Optional<ProjectAttributes, "id"> {}
@@ -105,6 +106,10 @@ Project.init(
     name: {
       type: new DataTypes.STRING(128),
       allowNull: false,
+    },
+    description: {
+      type: new DataTypes.STRING(128),
+      allowNull: true,
     },
   },
   {
--- a/types/test/typescriptDocs/ModelInit.ts
+++ b/types/test/typescriptDocs/ModelInit.ts
@@ -2,17 +2,9 @@
  * Keep this file in sync with the code in the "Usage" section in typescript.md
  */
 import {
-  Sequelize,
-  Model,
-  ModelDefined,
-  DataTypes,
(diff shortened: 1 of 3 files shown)


mmahalwy:
@sdepold my bad

sdepold:
> @sdepold my bad 

Seems the build is broken :)

mmahalwy:
@sdepold the funny thing is, this is actually _now_ doing the right thing. Typings failed because of missing create attributes :)

types/test/typescriptDocs/ModelInit.ts
```diff
@@ -204,6 +201,7 @@ async function doStuffWithUser() {
 
   const project = await newUser.createProject({
     name: "first!",
+    ownerId: 123,
```
EmilioCristalli, line 204: we are investigating some issues in our app after upgrading to Sequelize 6.12.0 that might be related to this PR.

Since user-> project association is defined like this:

```ts
User.hasMany(Project, { sourceKey: "id", foreignKey: "ownerId", as: "projects" });
```

Shouldn't `ownerId` be automatically set when you do `await newUser.createProject({ name: "first!" })`?

Looking at the code in [lib/associations/has-many.js](https://github.com/sequelize/sequelize/blob/main/lib/associations/has-many.js#L481), it seems like the `ownerId` will even be overwritten with `newUser.get('id')`.

I will try to open a PR with an idea to fix it, but wanted to first confirm this is not expected behavior

mmahalwy: @EmilioCristalli issue being that you need to specify the id otherwise it fails TS?

EmilioCristalli: yes.
Since `ownerId` is going to be automatically set, i think it shouldn't be required by TS. Or even not allowed, since it will be overwritten

mmahalwy: yeah, that's a good catch

EmilioCristalli: i just opened this PR https://github.com/sequelize/sequelize/pull/13892 with an idea to fix it

PR #13686, opened 2021-11-19
ci: add other ts versions to ci

mmahalwy:
### Pull Request Checklist

### Description Of Change
We are now on TS version 4.5! But our tests were only up to 4.1 😅 . This adds additional versions. My recommendation is to consider dropping version 3.9? 

### Todos

- [ ] 
- [ ] 
- [ ]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ts-version: ["3.9", "4.0", "4.1"]
+        ts-version: ["3.9", "4.0", "4.1", "4.2", "4.3", "4.4", "4.5"]
     name: TS Typings (${{ matrix.ts-version }})
     runs-on: ubuntu-latest
     steps:


WikiRik:
I'm not sure how TypeScript versions work with support, but DefinitelyTyped supports 3.8 and up so I think there's no need to drop 3.9 support. Unless there are features in 4.0 that we really want to use

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/docs/support-window.svg

mmahalwy:
@WikiRik sounds good

sdepold (approved):

sdepold:
> I'm not sure how TypeScript versions work with support, but DefinitelyTyped supports 3.8 and up so I think there's no need to drop 3.9 support. Unless there are features in 4.0 that we really want to use
> 
> https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/docs/support-window.svg

Did this change drop any coverage?

sdepold:
Ah never mind. I found the reference to your response

Issue #13132, opened 2021-03-22
Declare types to match ESM usage


JacobLey:
## Issue Description

Current typings are emitted such that something like:
```
import { where } from 'sequelize';
where()
```
_appears_ to be entirely valid code.

While that is the case with Transpiled code (e.g. the above becomes something like:
```
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
const sequelize_1 = require("sequelize");
sequelize_1.where();
```
in a commonjs module)

it is not valid when using native ESM (`"type":"module"`), which will result in the following error:
```
import { where } from 'sequelize';
         ^^^^^
SyntaxError: Named export 'where' not found. The requested module 'sequelize' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'sequelize';
const { where } = pkg;
```

### What was unclear/insufficient/not covered in the documentation

Current typings encourage the usage of named exports from Sequelize https://sequelize.org/master/manual/typescript.html and similarly internally https://github.com/sequelize/sequelize/blob/main/types/test/where.ts

This documentation infers a commonjs environment, and is not compatible with ESM.

### If possible: Provide some suggestion on how we can enhance the docs

There are 3 general approaches I can think of:

1) Change all named exports in [index.d.ts](https://github.com/sequelize/sequelize/blob/main/types/index.d.ts) to `export type` to encourage usage of the default export. This would most likely be a breaking change for people that use commonjs typescript (probably 95%) but would result in the most "correct" typings
2) Add footnote to documentation saying something like "If you use ESM, only use the default export", which would be enough to "support it" but I think it would be confusing to technically have two different usages
3) Actually export named values. While this would actually be the best, there are a _lot_ of exports and would require actual JS changes, which overall would require a lot more work.

### Additional context

Originally brought up here: https://github.com/sequelize/sequelize/pull/12921#discussion_r598869811

I use Sequelize with ESM/Typescript and have run into this issue before. 

using something like
```
import Sequelize, { Transaction } from 'sequelize';
const { Model }

class Foo extends Model {}...

export const async withTransaction(transaction?: Transaction): Promise<Foo> => Foo.findOne({ transaction })
```
has worked for me (e.g. use default export for Javascript values, named exports for types)

### Is this issue dialect-specific?

- [X] No. This issue is relevant to Sequelize as a whole.
[...]

PR #13689, opened 2021-11-22
Add support for native ESM named exports

ephys:
### Pull Request Checklist

### Description Of Change

This PR adds support for non-transpiled ESM named imports (Closes https://github.com/sequelize/sequelize/issues/13132)

Instead of converting the codebase to ESM and transpiling to CJS, this PR takes the simpler approach of adding a separate entry point (index.mjs) for ESM that imports the CJS and re-exports it as the default export, and its properties as named exports.

Note: while testing this I noticed that the TypeScript typing declares an export "validator", but the actual export is called "Validator".

`exports` has been set in package.json to make CommonJS require `index.js` & ESM import `index.mjs`


ephys:
Does the CI give any detail as to why the lint check failed? Replicating the steps locally gives me no errors 
image

WikiRik:
I think it was a temporary issue with GitHub Actions, @sdepold can you re-run the checks?

sdepold:
Assuming we eventually have full support for TS, will this still be needed? I'm somehow reluctant to merge since we would have to keep both index files in sync, right?!

ephys:
Once you have a full TS implem that is ESM compatible (the only impact is I think that you need to specify the file extension in imports, and not use directory imports), then you should be able to transpile to both CommonJS and ESM.

Both index files would have to be kept in sync yes, this is more of a temporary solution that works today and can be removed once the TS migration is complete

sdepold:
@Keimeno @AllAwesome497 --^ 

Opinion?

sdepold:
I think I would rather try to avoid merging this.

Keimeno:
Will take a deeper look into this in the evening

ephys:
I understand that having to maintain two exports file is annoying. If this does not get merged, do you have an estimate regarding how long it would take to fully migrate to an ESM compatible implementation?

More and more libraries are moving to native-esm only, forcing apps to migrate too and while we can import commonjs, every import must be changed from this 

image

to this

image

(as well as having to split uses between the type and the value, making it incompatible with TS's `emitDecoratorMetadata`).

This temporary patch would remove the need to make the migration until the TS migration is complete.

AllAwesome497:
> I think I would rather try to avoid merging this.

I'd also opt to avoid merging this, I think its better to wait until we can convert this to TS.

If we want to get this working for mjs ASAP I think something like this should work:
```js
// lib/sequelize.js

module.exports = Sequelize;
// We're changing the `Sequelize` object here, its just more readable as 
// module.exports + that matches what we already have in `lib/sequelize.js`
module.exports.Sequelize = Sequelize;
module.exports.default = Sequelize;
module.exports.__esModule = true; // tell ES modules that this is an esmodule.
```

ephys:
But that is not how native ES modules work. Only bundlers respect the `__esModule` property. 
When a module imports a commonjs script, [only the `default` import is available (and equal to `module.exports`)](https://nodejs.org/api/esm.html#import-statements).

There are only 3 ways to support ES Modules:

## Option 1: Convert the whole codebase to ES Modules and drop CommonJS support

*cons:*

- This is a huge breaking change, a lot of apps are written using CommonJS and CommonJS can only import ES Modules using Dynamic Import (`import(x): Promise`)
- It's a lot of work: You need to make sure every import is compatible with ESM. That means always specifying the file extension, and being careful when importing third-party libraries as CommonJS modules *do not support* named exports (you can only import `module.exports` as the default export)

## Option 2: Convert the whole codebase to ES Modules and transpile to CommonJS

*cons*:

- Similarly, a lot of work
- doing `import('sequelize')` and `require('sequelize')` will return two different instances of Sequelize as `import()` which *will* cause issues.

## Option 3: Keep the codebase as CommonJS and include a module that re-exports the CommonJS module using named exports

*cons*:

- You need to maintain an export file (but that's what unit tests are for)

---

Closing note: Node does have some static analysis for exposing named exports from CommonJS modules but it [should not be relied upon](https://simonplend.com/node-js-now-supports-named-imports-from-commonjs-modules-but-what-does-that-mean/), it's also not available for node 12.

AllAwesome497:
Alternatively, once we finish the migration to TypeScript we should be able to configure the package to allow us to build for both nodejs and cjs, as [esbuild supports this](https://esbuild.github.io/api/#format).

We'd probably want to implement a clone of the index file though, since `lib/seqeulize.js` currently relies on `module.exports` (the export is the class `Sequelize`, but `Transaction`, `DataTypes`, etc are also exported).

ephys:
Transpiling to both CommonJS and ESM is exactly the [option 2 I described](https://github.com/sequelize/sequelize/pull/13689#issuecomment-982437910) and is going to cause huge issues to CommonJS users:

> doing import('sequelize') and require('sequelize') will return two different instances of Sequelize as import() will import the module version and require the script. This will cause issues. Especially since import() is available to CommonJS

You *can* include both `index.js` and `index.mjs`, that's exactly what this PR is doing.

ephys:
Sorry for the ping but any update on this @Keimeno @sdepold? I think https://github.com/sequelize/sequelize/pull/13689#issuecomment-982437910 explains why this PR is the only way to support ESM named exports and CommonJS at the same time without causing weird bugs

AllAwesome497:
> doing import('sequelize') and require('sequelize') will return two different instances of Sequelize as import() will import the module version and require the script. This will cause issues. 

What kinds of issues?  I can't see why a module would be partially CJS and partially MJS at runtime.

I think its also better to have a full ESM module for tree shaking purposes rather than having a mixed source.  

> Especially since import() is available to CommonJS

In NodeJS v16.7.0 I get `SyntaxError: Cannot use import statement outside a module`.

And similarly, in NodeJS v16.7.0 I can't use `require` from a `mjs` module without using node's module api to create a `require` function, ie:

```js
import Module from 'module';

const require = Module.createRequire('/full/path/to/file');
```

AllAwesome497:
I would be fine with supporting esm in the short term with this, I'd rather have something than nothing for anyone who wants to use ESM.

However I think that in the long term it'd be best to have both esm and cjs compile targets, as the source will be esm and I'd like to keep the transpiled code as close to its native runtime as possible.

@ephys this change doesn't account for esbuild, can you modify the `build.js` script to account for this?  I believe a simple `cp index.mjs dist/` (translated to use node's `fs` builtin if possible) should suffice.

AllAwesome497 (changes requested):
Didn't think to request changes in my previous comment, but:

This change doesn't account for `dist/`, it'll need to be modified to copy the `index.mjs` file to `dist/` (I don't think any esbuild tanspiling is necessary here).

I believe `package.json` will also need to be updated a bit for this.

ephys:
> In NodeJS v16.7.0 I get SyntaxError: Cannot use import statement outside a module.

You're mistaking the static import with the dynamic import. [As specified here](https://nodejs.org/api/esm.html#import-expressions), dynamic imports are available to CommonJS. Static imports are not.

image

Plus, as you said yourself, ESM could also access `require` through `module.createRequire` and load two different versions of Sequelize without realizing that they are different.

---

> What kinds of issues? I can't see why a module would be partially CJS and partially MJS at runtime.

Well for starter using any library written in CommonJS that depends Sequelize. They will all break for ESM users.

Say your write your app in full native esm.  
Your app uses a library written in CommonJS, which itself depends on Sequelize (say https://github.com/ephys/sequelize-cursor-pagination).

Your app is therefore going to do this:
```
esm app → `import 'sequelize'` (imports the ESM version of Sequelize)
        → `import '@ephys/sequelize-find-by-cursor'` → `require('sequelize')` (imports the CommonJS version of Sequelize)
```

`@ephys/sequelize-find-by-cursor` has a method `findByCursor` which takes a Sequelize Model as a parameter.

For type safety, this method checks whether the parameter it received is indeed a sequelize model like so:
image

Using the solution of providing two completely separate transpilation outputs, that is going to return false as there will be two different `Model` classes. Whereas Solution 3 exposes the same `Model` class to both CJS and ESM, so this type safety check will pass.

<details>
  <summary>Extra scenario</summary>

Alternative scenario: Many people are currently using the ESM import syntax but then transpiling and running it as CJS.  
Given *Dynamic ESM imports* are available to CommonJS module, any transpiler could decide that if the target is node, then *static imports* are transpiled to CommonJS, but *dynamic imports* are not.

So the following code

```typescript
import SequelizePkg from 'sequelize';

(async () => {
  console.log((await import('sequelize')).default === SequelizePkg);
})();
```

could be transpiled to this:

```javascript
const SequelizePkg = require('sequelize');

(async () => {
  console.log((await import('sequelize')).default === SequelizePkg);
})();
```

which would output `false` using your solution, `true` using Solution 3.

So as long as you support CommonJS, Solution 3 is your bug free version. 
Any other version is going to cause issues. 
Once you drop CommonJS, sure, drop this file too.
</details>

---

> I think its also better to have a full ESM module for performance & security rather than having a mixed source.

What extra security are you speaking of?  
As for performance, the tiny improvement to startup time this would bring is definitely not worth the cost of introducing weird bugs like Sequelize from `import()` not being equal to Sequelize from `require()`.

---

By the way, please note that the code that's currently being written for the TypeScript migration is not valid ESM.  
To support ESM, the source code needs to specify the file extension when it imports a file. The TypeScript compiler is not going to add them: https://nodejs.org/api/esm.html#mandatory-file-extensions

---

> This change doesn't account for dist/, it'll need to be modified to copy the index.mjs file to dist/ (I don't think any esbuild tanspiling is necessary here).

Will do

ephys:
I've copied the file to /dist and I've added a test that ensures CommonJS & ESM have export parity.

The tests will fail because the typings Sequelize declares do not actually match reality. Can I fix that in this PR?

The issues that need to be fixed:

- The typings declare exporting `validator` but the reality is that it exports `Validator`. I'd like to change the typing.
- The typings declare exporitng classes `BelongsTo`, `HasOne`, `HasMany`, `BelongsToMany` & `QueryInterface` but they aren't. Can I add add these exports or should I make the typing say that they are type only?

sdepold:
I have added a v7 branch now and we could start removing support for node 10 in that branch since we planned to do that anyway. Should we target that branch instead of main maybe?

ephys:
I think we can land this in v6, node 10 users won't be able to import this file anyway so I can just skip the test for node 10.

ephys:
The test has been disabled on node 10, only thing left to resolve is decide whether we handle the typing issues these tests unearthed in this PR (https://github.com/sequelize/sequelize/pull/13689#issuecomment-987812367)

WikiRik:
Let's do that in a separate PR if you don't mind and we will merge this PR after that one

ephys:
Here we go :) https://github.com/sequelize/sequelize/pull/13751

sdepold:
looks promising

sdepold:
I assume some update is needed now?

ephys:
Yes, just need to tell the test that Sequelize.db2 is not an export. Should be good now :)

sdepold (approved):

sdepold:
Woop

ephys:
🥳
Thank you for merging :)

sdepold:
I'll release this as part of our beta releases in a moment :)

WikiRik:
Nice work on this! @ephys 
I just thought it might be nice to include this to the docs somewhere, could you maybe take a look into that soon?

PR #13703, opened 2021-11-24
ci(node): use Node 16 instead of 12

WikiRik:
### Pull Request Checklist

- [X] Have you added new tests to prevent regressions? _Not needed_
- [X] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)? _No, but unrelated to this PR_
- [X] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? _Not needed_
- [X] Did you update the typescript typings accordingly (if applicable)? _Not needed_

### Description Of Change
Use Node 16 instead of Node 12 for tests, linting, test typing, and releasing.


WikiRik:
@sdepold I think this one can be merged even with the failing tests in main

Keimeno:
I might even go the other way here personally. I doubt that running the tests on multiple node versions adds any significant value. Personally, I would only run the tests on the last supported node version v10.

This is also due to the problem that we have many tests randomly failing, this will only increase with this PR. This would require a maintainer to look through every failing workflow to check if it's a false negative, or if the failing workflow is due to the changed code

WikiRik:
> I might even go the other way here personally. I doubt that running the tests on multiple node versions adds any significant value. Personally, I would only run the tests on the last supported node version v10.
> 
> This is also due to the problem that we have many tests randomly failing, this will only increase with this PR. This would require a maintainer to look through every failing workflow to check if it's a false negative, or if the failing workflow is due to the changed code

I see your point, but I'm just afraid of what will happen if something would be removed/changed between the last supported version and the current LTS. What we could do is only test on Node 10 and Node 16 and just hope they don't have changes reverted between those versions.

WikiRik:
Marked it as draft again until we have reached an agreement on which Node versions we want to test on

Keimeno:
> I see your point, but I'm just afraid of what will happen if something would be removed/changed between the last supported version and the current LTS. What we could do is only test on Node 10 and Node 16 and just hope they don't have changes reverted between those versions.

This in my opinion is a good compromise. Especially, as I don't see them reverting breaking changes. We would still have numerous workflows running, but at least not over 60

sdepold:
> > I see your point, but I'm just afraid of what will happen if something would be removed/changed between the last supported version and the current LTS. What we could do is only test on Node 10 and Node 16 and just hope they don't have changes reverted between those versions.
> 
> This in my opinion is a good compromise. Especially, as I don't see them reverting breaking changes. We would still have numerous workflows running, but at least not over 60

sounds good to me

WikiRik:
Okay, then I will update this PR accordingly

sdepold (approved):

WikiRik:
@sdepold can you update the required checks?

sdepold:
> @sdepold can you update the required checks?

I think we talked about that the other day and i have to admit that I forgot the settings for that again. It's not sufficient as it is?

WikiRik:
> > @sdepold can you update the required checks?
> 
> I think we talked about that the other day and i have to admit that I forgot the settings for that again. It's not sufficient as it is?

It's part of the [Branch protection rules](https://github.com/sequelize/sequelize/settings/branches). It wants the results of the Node 12 tests as well right now, but it won't get them (as you can see it is still expecting 13 checks for this PR)

Keimeno (approved):

PR #13706, opened 2021-11-24
refactor: use includes instead of or operators

WikiRik:
As mentioned in https://github.com/sequelize/sequelize/pull/13374


sdepold:
Automated or manually fixed?

test/unit/model/define.test.js
```diff
@@ -117,7 +117,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           }
         });
 
-        if (dialect === 'postgres' || dialect === 'sqlite' || dialect === 'mssql') {
+        if (['postgres', 'mariadb', 'mssql'].includes(dialect)) {
```
sdepold, line 120: should be pg, sqlite, mssql

sdepold: since no test failed, I think we are lacking some

WikiRik:
Mostly manually, only one file was done automatically since it had a lot of the same. I checked everything again before committing but it would be good if you could look at it as well

sdepold (changes requested):

test/unit/model/define.test.js
WikiRik: Yup, will add some

WikiRik:
Uhmm... I think we have a bigger issue; https://github.com/sequelize/sequelize/runs/4315490677?check_suite_focus=true

WikiRik:
Draft until the tests are fixed on main

fncolon (approved):

test/unit/model/define.test.js
sdepold: Did you add a test?

sdepold: > Yup, will add some

Actually ... it was a test already?!

PR #13708, opened 2021-11-24
test: fix failing stack trace test

WikiRik:
The remaining checks that fail for postgres and sqlite are due to https://github.com/sequelize/sequelize/pull/13594 
The one test that failed on MySQL 8 can be ignored, that one happens quite often and we are not yet sure why

@slickmb we had an issue that the tests in the checks were not actually running. This is fixed now and we noticed some failing tests for your PR on postgres and sqlite. Any idea on how to solve them?


slickmb:
👀

slickmb:
I see the issue. Opening PR to fix now.

PR #13710, opened 2021-11-24
feat(types): transition lib/errors

Keimeno:
### Pull Request Checklist

### Description Of Change

Transitions the `lib/errors` folder to TypeScript


lib/errors/database-error.ts
```diff
@@ -0,0 +1,40 @@
+import BaseError, { ErrorOptions } from './base-error';
+
+interface IDatabaseError extends Error {
```
Keimeno, line 3: Unsure about the name of this interface, but might reuse and rename it elsewhere when necessary

Keimeno: Just seen the `CommonErrorProperties` interface in `types/lib/errors.d.ts`, will change it then

lib/errors/bulk-record-error.ts
```diff
@@ -10,12 +9,15 @@ const BaseError = require('./base-error');
  * @param {object} record  DAO instance that error belongs to
  */
 class BulkRecordError extends BaseError {
-  constructor(error, record) {
+  errors: Error;
```
Keimeno, line 12: The parameter name here is misleading. It should be singular, it's just a small thing, but changing this could result in a breaking change.

@sdepold can we make a milestone for major releases (aka. v7), where we can reference this?

Keimeno:
A few things I've noticed while migrating to typescript here. Most types are inconsistent with the JS code. Changes that were done here won't be breaking changes for JS users, but might be breaking changes for TS users.

lib/errors/base-error.ts
```diff
@@ -8,10 +21,11 @@
...
-  constructor(message) {
+  constructor(message: string) {
     super(message);
     this.name = 'SequelizeBaseError';
   }
 }
 
-module.exports = BaseError;
+export default BaseError;
+export { ErrorOptions, CommonErrorProperties };
```
Keimeno, line 31: I'm unsure if we want to keep named exports at the bottom of the file, or if we should directly export when we create a class, function, variable etc (in the same line), as was done in `lib/utils/logger.ts`

test/integration/model/bulk-create.test.js
```diff
@@ -310,7 +310,6 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         const expectedValidationError = 'Validation len on code failed';
         const expectedNotNullError = 'notNull Violation: Task.name cannot be null';
 
-        expect(error).to.be.instanceof(AggregateError);
```
Keimeno, line 313: `instanceof` does not work with transpiled code. See https://www.chaijs.com/api/bdd/#method_instanceof and https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work for more details

lib/dialects/mssql/async-queue.ts
```diff
@@ -0,0 +1,60 @@
...
+            )
+          );
+        }
+        return asyncFunction().then(resolve, reject);
+      });
+    });
+  }
+}
+
+export default AsyncQueue;
```
sdepold, line 59: Any particular reason to not export the functions during the definition already? export function bla?

lib/errors/base-error.ts
sdepold: Aha! I'd vote for inlining it with the definition instead of at the bottom of the file

lib/errors/bulk-record-error.ts
sdepold: Don't you have the permissions?

WikiRik: We already have a milestone for v7: https://github.com/sequelize/sequelize/milestone/20

lib/errors/base-error.ts
Keimeno: Using inline syntax for named exports now. Though, I would keep default exports at the bottom of the file. Furthermore, I would keep the default exported object at the very bottom of the file, as seen in `validation-error.ts`

lib/dialects/mssql/async-queue.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Thrown when a connection to a database is closed while an operation is in progress
+ */
+export class AsyncQueueError extends BaseError {
+  constructor(message: string) {
+    super(message);
+    this.name = 'SequelizeAsyncQueueError';
+  }
+}
+
+class AsyncQueue {
```
sdepold, line 13: export default class AsyncQueue?

lib/errors/base-error.ts
Keimeno: As discussed internally, we will use inline named exports and keep default exports at the bottom of the file.

lib/errors/base-error.ts
```diff
@@ -8,10 +21,10 @@
  * The Base Error all Sequelize Errors inherit from.
  */
 class BaseError extends Error {
-  constructor(message) {
+  constructor(message: string) {
```
sdepold, line 24: Given that we were previously calling `super()`, I wonder if this should be `message?: string`?

lib/dialects/mssql/async-queue.ts
```diff
@@ -0,0 +1,59 @@
...
+    // However, this ensures that this.previous will never be a rejected promise so the queue will
+    // always keep going, while still communicating rejection from asyncFunction to the user.
+    return new Promise((resolve, reject) => {
+      this.previous = this.previous.then(() => {
+        this.rejectCurrent = reject;
+        if (this.closed) {
+          return reject(
+            new ConnectionError(
+              new AsyncQueueError(
+                'the connection was closed before this query could be executed'
```
sdepold, line 48: I considered not noting this but since I asked something anyway, I decided to do at least ask :)

Should we maybe store the connectionError instance as a singleton somewhere and use it here and also in the close function?

lib/errors/aggregate-error.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param {Error[]} [errors] Array of errors
+ *
+ * @property errors {Error[]}
+ */
+class AggregateError extends BaseError {
+  errors: Array<AggregateError | Error>;
+
+  constructor(errors: Array<AggregateError | Error>) {
+    super('');
```
sdepold, line 14: If we were to declare the BaseError message optional, then this could be an empty super call

lib/errors/bulk-record-error.ts
sdepold: Should we add a ticket to the v7 project?

test/integration/error.test.js
```diff
@@ -153,7 +153,7 @@ describe(Support.getTestDialectTeaser('Sequelize Errors'), () => {
     });
 
     it('SequelizeValidationErrorItem.Origins is valid', () => {
-      const ORIGINS = Sequelize.ValidationErrorItem.Origins;
+      const ORIGINS = Sequelize.ValidationErrorItemOrigin;
```
sdepold, line 156: Does this exists?

lib/errors/validation-error.ts
```diff
@@ -1,64 +1,85 @@
-'use strict';
-
-const BaseError = require('./base-error');
+import BaseError, { ErrorOptions } from './base-error';
+import type { Model } from '../../types/lib/model';
 
 /**
```
sdepold, line 4: I have to admit that this change is a bit too much for my current state of mind, but you have my trust.

test/integration/error.test.js
sdepold: I found it further up. I tend to think that this might be breaking change and that we should add an alias + remove it in v7?

sdepold (changes requested):

sdepold:
Some minor things really. Over all good stuff :)

lib/dialects/mssql/async-queue.ts
```diff
@@ -0,0 +1,59 @@
...
+  constructor(message: string) {
+    super(message);
+    this.name = 'SequelizeAsyncQueueError';
+  }
+}
+
+class AsyncQueue {
+  previous: Promise<unknown>;
+  closed: boolean;
+  rejectCurrent: (reason?: any) => void;
```
AllAwesome497, line 16: bit of a nit: can you specify `private` / `protected` as needed here (and everywhere else)?

AllAwesome497: I personally also prefer explicitly setting `public` but I don't feel super strongly on that.

lib/errors/base-error.ts
```diff
@@ -8,10 +21,10 @@
  * The Base Error all Sequelize Errors inherit from.
  */
 class BaseError extends Error {
```
AllAwesome497, line 23: Can this be an [abstract](https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-classes-and-members) class?  (might not be available in our lowest supported TS version, if not can you add a todo?)

lib/errors/optimistic-lock-error.ts
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * The name of the model on which the update was attempted
+     *
+     * @type {string}
+     */
+    this.modelName = options.modelName;
+    /**
+     * The values of the attempted update
+     *
+     * @type {object}
```
AllAwesome497, line 39:
```suggestion
```

lib/errors/optimistic-lock-error.ts
```diff
@@ -0,0 +1,46 @@
...
+    options.message =
+      options.message ||
+      `Attempting to update a stale model instance: ${options.modelName}`;
+
+    super(options.message);
+    this.name = 'SequelizeOptimisticLockError';
+    /**
+     * The name of the model on which the update was attempted
+     *
+     * @type {string}
```
AllAwesome497, line 33:
```suggestion
```

lib/errors/optimistic-lock-error.ts
```diff
@@ -0,0 +1,46 @@
...
+  constructor(options: OptimisticLockErrorOptions) {
+    options = options || {};
+    options.message =
+      options.message ||
+      `Attempting to update a stale model instance: ${options.modelName}`;
+
+    super(options.message);
+    this.name = 'SequelizeOptimisticLockError';
+    /**
+     * The name of the model on which the update was attempted
```
AllAwesome497, line 31: I think this jsdoc needs to be where this property's type is declared, not where its assigned.

lib/errors/validation-error.ts
```diff
@@ -161,13 +192,13 @@ class ValidationErrorItem {
    *
    * @private
    */
-  getValidatorKey(useTypeAsNS, NSSeparator) {
+  getValidatorKey(useTypeAsNS: boolean, NSSeparator: string): string {
```
AllAwesome497, line 195: please rm the `@private` in the jsdoc and specify `private` so TS will pick it up.

```suggestion
  private getValidatorKey(useTypeAsNS: boolean, NSSeparator: string): string {
```

lib/errors/validation-error.ts
```diff
@@ -177,37 +208,71 @@ class ValidationErrorItem {
...
-
-/**
- * An object that is used internally by the `ValidationErrorItem` class
- * that maps current `type` strings (as given to ValidationErrorItem.constructor()) to
- * our new `origin` values.
+ * @param {string} message Error message
+ * @param {Array} [errors] Array of ValidationErrorItem objects describing the validation errors
  *
- * @type {object}
+ * @property errors {ValidationErrorItem[]}
```
AllAwesome497, line 222: Shouldn't need to specify the type here iirc.

lib/errors/validation-error.ts
```diff
@@ -177,37 +208,71 @@ class ValidationErrorItem {
...
-  DB: 'DB',
-  FUNCTION: 'FUNCTION'
-};
-
-/**
- * An object that is used internally by the `ValidationErrorItem` class
- * that maps current `type` strings (as given to ValidationErrorItem.constructor()) to
- * our new `origin` values.
+ * @param {string} message Error message
+ * @param {Array} [errors] Array of ValidationErrorItem objects describing the validation errors
```
AllAwesome497, line 220:
```suggestion
 * @param message Error message
 * @param [errors] Array of ValidationErrorItem objects describing the validation errors
```

lib/errors/validation-error.ts
```diff
@@ -177,37 +208,71 @@ class ValidationErrorItem {
...
+      this.stack = options.stack;
+    }
+  }
+
+  /**
+   * Gets all validation error items for the path / field specified.
+   *
+   * @param {string} path The path to be checked for error items
+   *
+   * @returns {Array<ValidationErrorItem>} Validation error items for the specified path
```
AllAwesome497, line 266:
```suggestion
   * @param path The path to be checked for error items
   *
   * @returns Validation error items for the specified path
```

AllAwesome497 (commented):
Looks mostly fine, just some nits.

Can you remove the types from the comments?  We should infer from TS types.

Also personal preference: can you explicitly specify `public`, `protected`, `private`, etc?  IMO its better to do so since it forces you to think "should this be public?" (no strong opinions on this though)

sdepold:
> Looks mostly fine, just some nits.
> 
> 
> 
> Can you remove the types from the comments?  We should infer from TS types.
> 
> 
> 
> Also personal preference: can you explicitly specify `public`, `protected`, `private`, etc?  IMO its better to do so since it forces you to think "should this be public?" (no strong opinions on this though)

I think the types in the comments are for the docs right? I was wondering the same however?

Regarding visibility: Let's have a poll on that :)

sdepold:
Seems we are leaning towards omitting public modifier and only go with private and protected.

lib/errors/validation-error.ts
Keimeno: This actually is a public function, I believe the `@private` flag was only set because it should be used internally within sequelize and not outside.

I will remove the `@private` flag and keep the function public, I believe this to be the best way. Otherwise, we would have to call `validationErrorItem['getValidatorKey'](...)` instead of `validationErrorItem.getValidatorKey(...)` internally to access the private property

test/integration/error.test.js
Keimeno: Setting it as static property of `ValidationErrorItem` now, added the old test back, next to this one and added the deprecation flag

lib/dialects/mssql/async-queue.ts
Keimeno: I would keep these properties as is, since they weren't prefixed with an underscore. I think most properties in `lib/errors`, if not all, are meant to be public

lib/dialects/mssql/async-queue.ts
Keimeno: I'm unsure if we get a benefit from doing so 🤔, maybe I'm missing something out

sdepold: Oh sorry. I'm seeing it now. I thought it was the same error message

sdepold (approved):

Issue #13372, opened 2021-07-11
Chore: migrate away from moment.js


soryy708:
## Issue Creation Checklist

- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Issue Description

Moment.JS is in maintenance mode, + it's a pretty old package which doesn't play well with modern build pipelines.
I propose we migrate away from MomentJS to something else, like [date-fns](https://date-fns.org/) or something else.

### Additional context

[Moment.JS docs](https://momentjs.com/docs/) say the following:

> We now generally consider Moment to be a legacy project in maintenance mode
> In most cases, you should not choose Moment for new projects
> There are several great options to consider using instead of Moment.

### Is this issue dialect-specific?

- [x] No. This issue is relevant to Sequelize as a whole.
- [ ] Yes. This issue only applies to the following dialect(s): XXX, YYY, ZZZ
- [ ] I don't know.

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [x] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

imwh0im:
i agree this issue. migrate from momentjs to dayjs

soryy708:
> 
> 
> i agree this issue. migrate from momentjs to dayjs

Cool. Why dayjs specifically?

Moment recommends the following options:

* Luxon - 1.3m weekly dls on npm
* Day.js - 4.6m weekly dls on npm
* date-fns - 8.1m weekly dls on npm
* js-Joda - 0.2m weekly dls on npm

imwh0im:
dayjs was almost the same as momentjs in usage.
I recommend to look at the this.
https://day.js.org/

> 2kb: low resource
> simple: If you use Moment.js, you already know how to use Day.js.
> immutable: this is helps prevent bugs and avoid long debugging sessions.
> I18n: Day.js has great support for internationalization

it's not a coercion. just a suggestion. :)

soryy708:
https://date-fns.org/

> **modular**: works well with bundlers and supports tree shaking, which helps it be very light
> **immutable & pure**: helps to prevent bugs and avoid long debugging sessions
> **I18n**: date-fns has dozens of locales. Only the ones that you use will be included in your project
[...]

PR #13712, opened 2021-11-25
fix(data-types): unnecessary warning when getting data with DATE dataTypes

fncolon:
### Pull Request Checklist

### Description Of Change

unnecessary warning when getting data with DATE dataTypes attributes.

Capture

### Todos

- [x] check if value is date

### DIFF
--- a/lib/dialects/mariadb/data-types.js
+++ b/lib/dialects/mariadb/data-types.js
@@ -54,8 +54,12 @@ module.exports = BaseTypes => {
       return this._length ? `DATETIME(${this._length})` : 'DATETIME';
     }
     _stringify(date, options) {
-      date = this._applyTimezone(date, options);
-      return date.format('YYYY-MM-DD HH:mm:ss.SSS');
+      if(_.isDate(date)){
+        date = this._applyTimezone(date, options);
+        return date.format('YYYY-MM-DD HH:mm:ss.SSS');
+      }
+
+      return date;
     }
     static parse(value, options) {
       value = value.string();
--- a/lib/dialects/mysql/data-types.js
+++ b/lib/dialects/mysql/data-types.js
@@ -53,12 +53,16 @@ module.exports = BaseTypes => {
       return this._length ? `DATETIME(${this._length})` : 'DATETIME';
     }
     _stringify(date, options) {
-      date = this._applyTimezone(date, options);
-      // Fractional DATETIMEs only supported on MySQL 5.6.4+
-      if (this._length) {
-        return date.format('YYYY-MM-DD HH:mm:ss.SSS');
+      if(_.isDate(date)){
+        date = this._applyTimezone(date, options);
+        // Fractional DATETIMEs only supported on MySQL 5.6.4+
+        if (this._length) {
+          return date.format('YYYY-MM-DD HH:mm:ss.SSS');
+        }
+        return date.format('YYYY-MM-DD HH:mm:ss');
       }
-      return date.format('YYYY-MM-DD HH:mm:ss');
+
+      return date;
     }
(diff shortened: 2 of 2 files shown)


sdepold:
Can we have a test that ensures that the situation has improved?

fncolon:
> Can we have a test that ensures that the situation has improved?

Look like checking this warning has resolve or not cannot be checked with unit test, only manual test. Correct me if I'm wrong.

fncolon:
You can check by manual test with 

```
  const crowdFundings = await CrowdFunding.findAndCountAll({
    distinct: true, // don't count the include part https://github.com/sequelize/sequelize/pull/4016#issuecomment-116294996
    attributes: [
     'date_attributes' // <- will cause the warning from momentjs
    ],
    include: [{ all: true }],
    where: {
      ...data,
      [Op.or]: searchUtil.search(crowdFundingAttributes, query.q || ""),
    },
    ...pagination.getPaginationOptionsData(),
  });
```

I take this codebase from my project haha

sdepold:
Is the result of the function different if it is not a date?

fncolon:
> Is the result of the function different if it is not a date?

No, it's just produce momentjs warning which may some developers don't like, so I improve it to just apply timezone when the value in the correct format.

WikiRik:
I don't think we need to merge this, especially if https://github.com/sequelize/sequelize/issues/13372 will be done soon

fncolon:
> I don't think we need to merge this, especially if #13372 will be done soon

It might be implemented too but we don't have an updated PR for it. So merge this now wil lstill be needed so someone not confusing with this kind of error just because we apply timezone for non-Date values.

fncolon:
Also we need to check the value first if it's date or not, whether it is using momentjs or others

WikiRik:
Sure, then this looks good

moonjoungyoung:
@fncolon @sdepold 

Hello.
After this PR, using moment objects are throwing error.
Is this an intended modification?

fncolon:
> @fncolon @sdepold
> 
> Hello. After this PR, using moment objects are throwing error. Is this an intended modification?

Hello, can you reproduce your example in an issue, technicallly it's just check a raw date value.

cmelendez:
> > @fncolon @sdepold
> > Hello. After this PR, using moment objects are throwing error. Is this an intended modification?
> 
> Hello, can you reproduce your example in an issue? technicallly it's just check a raw date value.

Hi @fncolon.

We upgraded from 6.9.0 to 6.12.1 today and our app started throwing errors when using a Moment object to do equality checks. I'm not sure if this PR is the cause but it seems to be similar to the problem described by @moonjoungyoung.

Here's a piece of code that is throwing a `Invalid value Moment<2021-12-21T19:52:23+00:00>` error.

```
phone_validation_until: {
   [Op.gte]: moment()
  },
```

Thanks.

cmelendez:
Some extra info:

* When setting any date value with `moment`, the stored value is `0000-00-00 00:00:00`.
* We are using `mysql`.

fncolon:
> @fncolon @sdepold
> 
> Hello. After this PR, using moment objects are throwing error. Is this an intended modification?

Just find out the problem, it was returning date instead of formatted date, also change isDate from `lodash` with isMoment from `moment`.

fncolon:
> Some extra info:
> 
> * When setting any date value with `moment`, the stored value is `0000-00-00 00:00:00`.
> * We are using `mysql`.
> * Moment objects works perfectly up to 6.11.0. Issues starts with >=6.12.0.
> 
> edit: pinging @sdepold

Just created the pull request in #13818, thanks!

PR #13715, opened 2021-11-26
docs(contributing): add section on adding/updating deps

### LINKED ISSUE #13714: Yarn add/remove doesn't work, "Found incompatible module"
## Issue Creation Checklist

## Issue Description

When trying to `yarn add` or `yarn remove`, `yarn.lock` isn't changed and an error is written to the console:

```
error @typescript-eslint/eslint-plugin@5.3.0: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0". Got "14.15.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

- [ ] I don't know.

WikiRik:
Fixes #13714

@soryy708 Would something like this be good?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,7 +94,7 @@ Most operating systems provide all the needed tools (including Windows, Linux an
 
 - Mandatory:
 
-  - [Node.js](http://nodejs.org) (^12.22.0 || ^14.17.0 || >=16.0.0, or >10 with using the `--ignore-engines` on Yarn)
+  - [Node.js](http://nodejs.org)
   - [Git](https://git-scm.com/)
 
 - Optional (recommended):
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -94,7 +94,7 @@ Most operating systems provide all the needed tools (including Windows, Linux an
 
 - Mandatory:
 
-  - [Node.js](http://nodejs.org)
+  - [Node.js](http://nodejs.org) (^12.22.0 || ^14.17.0 || >=16.0.0, or >10 with using the `--ignore-engines` on Yarn)
```
Keimeno, line 97: I am not convinced that this really explains the problem correctly. The engine specification mentioned here comes from the package @soryy708 tried to install: `@typescript-eslint/eslint-plugin`.

We have the same issue, but with different engine limitations coming from `@npmcli/map-workspaces`, one of our sub-dependencies (`"^12.13.0 || ^14.15.0 || >=16"`).

I would probably add a point 2.1 titled "Adding and updating dependencies", where we give a hint to the contributor, that they must use `yarn` and they might run into a `Found incompatible module` error

Keimeno (changes requested):

CONTRIBUTING.md
WikiRik: Agree, thanks for the suggestion!

Keimeno (approved):

Issue #13671, opened 2021-11-16
The coordinates appear flipped in the examples, should be lng,lat.


reyemtm:
https://github.com/sequelize/sequelize/blob/47c4494968422585bf265063925d1662ffcd4173/lib/data-types.js#L805

https://github.com/sequelize/sequelize/blob/47c4494968422585bf265063925d1662ffcd4173/lib/data-types.js#L825

In GeoJSON coordinates are coded as `[lng,lat]` while the coordinates in the examples appear to be in `[lat,lng]`.

reyemtm:
Sorry I get so used to seeing -80*, 3* I didnt even take into account that
the reverse is also valid. Thanks for understanding and not reprimanding
me. 👍

sdepold:
:D I did some investigation and noticed that the given coordinates are somewhere in the Antarctica and if you swap them you end up on the US coast somewhere. So it seemed like a reasonable change indeed :)

PR #13718, opened 2021-11-28
fix(example): fix coordinates format as per GeoJson

madguy02:
### Description Of Change

A minor fix in the examples for GeoJson format

closes: #13671


sdepold:
This might sound like a stupid question and maybe it even is one but .... why is the current example not valid?

sdepold:
ok. nevermind. I think people rather expect a particular location somewhere in the us as opposed to some spot in the Antarctica :)

PR #13721, opened 2021-11-29
chore: add package support info

fncolon:
### Pull Request Checklist

### Description Of Change

It's not about which version we support by specifying numbers but follows Node.js conventions will be better. Nodejs have been working how packages are supported.

https://github.com/nodejs/package-maintenance/blob/main/docs/PACKAGE-SUPPORT.md

We can choose between all, lts, etc, check here https://github.com/nodejs/package-maintenance/blob/main/docs/PACKAGE-SUPPORT.md, also we need to match all of these supported versions in unit test.

This also included newer versions of Node.js instead of LTS.

Close https://github.com/sequelize/meetings/issues/5

### DIFF
--- a/package-support.json
+++ b/package-support.json
@@ -1,0 +1,16 @@
+{
+  "versions": [
+    {
+      "version": "*",
+      "target": {
+        "node": "supported"
+      },
+      "response": {
+        "type": "time-permitting"
+      },
+      "backing": {
+        "npm-funding": true
+      }
+    }
+  ]
+}


sdepold:
Should we add a ticket to the v7 project to change support to its versions only?

package-support.json
```diff
@@ -0,0 +1,16 @@
...
+    {
+      "version": "*",
+      "target": {
+        "node": "supported"
+      },
+      "response": {
+        "type": "time-permitting"
+      },
+      "backing": {
+        "npm-funding": true
```
sdepold, line 12: Should this be changed to 

```
donations: ["https://opencollective.com/sequelize"]
```

?

Alternatively we should probably add a funding property to the package.json

fncolon:
> Should we add a ticket to the v7 project to change support to its versions only?

Look like we should with providing the evidence of the supported versions by tests

package-support.json
fncolon: Yes, it's should! will work on it today. 😃

fncolon: Changed, also some changes for following the conventions from https://docs.npmjs.com/cli/v6/configuring-npm/package-json

package.json
```diff
@@ -1,9 +1,12 @@
 {
   "name": "sequelize",
-  "description": "Multi dialect ORM for Node.JS",
-  "version": "0.0.0-development",
-  "maintainers": [
-    "Pedro Augusto de Paula Barbosa <papb1996@gmail.com>"
+  "description": "An easy-to-use and promise-based multi SQL dialects ORM tool for Node.js",
+  "version": "6.12.0-beta.1",
```
sdepold, line 4: Please revert this line

package.json
```diff
@@ -1,9 +1,12 @@
 {
   "name": "sequelize",
-  "description": "Multi dialect ORM for Node.JS",
-  "version": "0.0.0-development",
-  "maintainers": [
-    "Pedro Augusto de Paula Barbosa <papb1996@gmail.com>"
+  "description": "An easy-to-use and promise-based multi SQL dialects ORM tool for Node.js",
```
sdepold, line 3: I think I would prefer this:

```
Sequelize is a promise-based Node.js ORM tool for Postgres, MySQL, MariaDB, SQLite, Microsoft SQL Server, Amazon Redshift and Snowflake’s Data Cloud. It features solid transaction support, relations, eager and lazy loading, read replication and more.
```

package.json
sdepold: Context: https://github.com/semantic-release/cli#what-it-does (see number 4)

sdepold (changes requested):

package.json
fncolon: More comprehensive, i liked it too

package.json
fncolon: Changed ✔️

package.json
fncolon: Changed ✔️

sdepold (approved):

PR #13735, opened 2021-12-02
test(integration): mark and fix flaky test

jesse23:
### Pull Request Checklist

### Description Of Change

- Mark all flaky tests with `Flaky test` in comments.
- Refactor some flaky test assertions to print more clear details.
- Fix one flaky test.

### DIFF
--- a/test/integration/associations/scope.test.js
+++ b/test/integration/associations/scope.test.js
@@ -321,7 +321,7 @@ describe(Support.getTestDialectTeaser('associations'), () => {
             this.Post.belongsToMany(this.Tag, { as: 'tags', through: this.PostTag, scope: { type: 'tag' } });
           });
 
-          it('should create, find and include associations with scope values', async function() {
+          it('[Flaky] should create, find and include associations with scope values', async function() {
             await Promise.all([this.Post.sync({ force: true }), this.Tag.sync({ force: true })]);
             await this.PostTag.sync({ force: true });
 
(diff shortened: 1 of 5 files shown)


test/integration/associations/scope.test.js
```diff
@@ -357,19 +357,25 @@ describe(Support.getTestDialectTeaser('associations'), () => {
...
+            // Flaky test: randomly one of the value on B will be 0 sometimes, for
+            // now no solution. Not reproducible at local or cloud with logging enabled
+            expect([
+              postACategories.length,
+              postATags.length,
+              postBCategories.length,
+              postBTags.length,
+              postCCategories.length,
+              postCTags.length
+            ]).to.eql([1, 1, 1, 1, 1, 1]);
```
sdepold, line 369: Shouldn't this need a deep equality comparison?

sdepold (changes requested):
Could you add flaky to the test case text so that we see it immediately in the actions?

test/integration/associations/scope.test.js
jesse23: `eql` is same as deep equal in mocha...
https://stackoverflow.com/questions/36798993/what-is-the-difference-between-equal-and-eql-in-chai-library

Yes I also feel use `.deep.equal` is better for next step ( switch to jest), let me fix it very quick.

jesse23: Forget about that `is better` above - since anyway we need to change the `.to.` later. This should be good to go.

sdepold: Interesting. I didn't know that and always used `to.deep.equal`:)

sdepold (dismissed):

jesse23:
> Could you add flaky to the test case text so that we see it immediately in the actions?

Good idea..

PR #13736, opened 2021-12-03
fix(model.d): fix findAndCountAll.count type

zorji:
### Pull Request Checklist

### Description Of Change

`findAndCountAll` should only return `count` as an array when the `group` option is provided.

This means in that case, `options` is no longer optional because when `options` is not provided, it's never a group count.
I also rearrange `findAndCountAll` order to ensure the single count overload is applied first.

### Todos

- [ ]


sdepold:
Do you plan to continue the work on this PR or why is it in draft state?

zorji:
Hi @sdepold 

Yes, I do want to work on in, probably over the next weekend. But if you are interested in it feel free to contribute. I think it's only the test cases are incomplete.

zorji:
> Do you plan to continue the work on this PR or why is it in draft state?

The fix is done, could you help review too?

jesse23 (approved):

jesse23 (changes requested):
Could u help to update the doc too if possible? Thx...

```
docs/manual/core-concepts/model-querying-finders.md
```

also it it posible to update the comments(JsDoc/TsDoc) for 2nd impl?

zorji:
> Could u help to update the doc too if possible? Thx...

Hi @jesse23 I was trying to update the docs and I found the documentation and type for `Model.count()` is very wrong. I think it's better to fix the docs in a separated PR because the `findAndCountAll()` is based on the implementation of `count()`.

i.e. look at this implementation here https://github.com/sequelize/sequelize/blob/5657a349a4f17997cc48979d9c1e49674c9322eb/lib/model.js#L2052

It's returning `Array<Record<string, any> & { count: number }>` but the doc and type says it returns `number[]`.

I'll prefer to keep this PR as it is and address that doc later.

sdepold (approved):

sdepold:
Sounds good. Looking forward to another PR

zorji:
@sdepold @jesse23 started draft PR here to fix count type https://github.com/sequelize/sequelize/pull/13786

PR #13748, opened 2021-12-07
ci(node): drop support for node 10

sdepold:
This PR is dropping support for Node 10. Should we go straight to the latest LTS version of Node 12 or is any version >= 12 fine?

### DIFF
--- a/build.js
+++ b/build.js
@@ -58,7 +58,7 @@ async function main() {
       // Adds source mapping
       sourcemap: true,
       // The compiled code should be usable in node v12.22
-      target: 'node12',
+      target: 'node12.22',
       // The source code's format is commonjs.
       format: 'cjs',
 
--- a/build.js
+++ b/build.js
@@ -58,7 +58,7 @@ async function main() {
       // Adds source mapping
       sourcemap: true,
       // The compiled code should be usable in node v10
-      target: 'node10',
+      target: 'node12',
       // The source code's format is commonjs.
       format: 'cjs',
 
--- a/build.js
+++ b/build.js
@@ -57,7 +57,7 @@ async function main() {
     build({
       // Adds source mapping
       sourcemap: true,
-      // The compiled code should be usable in node v10
+      // The compiled code should be usable in node v12.22
       target: 'node12',
       // The source code's format is commonjs.
       format: 'cjs',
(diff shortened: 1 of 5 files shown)


sdepold:
We might have to tweak the required builds for the project with this PR

package.json
```diff
@@ -19,7 +19,7 @@
   "main": "./dist/index.js",
   "types": "./dist",
   "engines": {
-    "node": ">=10.0.0"
+    "node": ">=12.0.0"
```
WikiRik, line 22:
```suggestion
    "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
```

As per the package.json of eslint

build.js
```diff
@@ -58,7 +58,7 @@ async function main() {
       // Adds source mapping
       sourcemap: true,
       // The compiled code should be usable in node v10
-      target: 'node10',
+      target: 'node12',
```
WikiRik, line 61:
```suggestion
      target: 'node12.22',
```

This should be tested, but seems that it would work. Otherwise we can do es2019 which is the last version on which all features are supported by Node 12.22.0 (as per https://node.green )

build.js
```diff
@@ -58,7 +58,7 @@ async function main() {
       // Adds source mapping
       sourcemap: true,
       // The compiled code should be usable in node v10
```
WikiRik, line 60:
```suggestion
      // The compiled code should be usable in node v12.22
```

WikiRik (changes requested):
Made some suggestions, after that it should be good to go. The Node versions in the GitHub Actions don't need to be specified as they will use the latest Node 12 version (12.22.7 as of now)

markdirish (commented):
I get that this is a `ci` PR, but since "public" files like `package.json` are touches, might be a good idea to change `README.md` as well:

https://github.com/sequelize/sequelize/blob/v7/README.md?plain=1#L15

sdepold:
> I get that this is a `ci` PR, but since "public" files like `package.json` are touches, might be a good idea to change `README.md` as well:
> 
> https://github.com/sequelize/sequelize/blob/v7/README.md?plain=1#L15

fixed :)

sdepold:
🤦 Merged this guy by accident on the command line.

sdepold:
Please adjust as you see fit

PR #13751, opened 2021-12-08
Fix typing & export mismatch

ephys:
### Pull Request Checklist

### Description Of Change

Fixes the mismatch between typings & exports that is described here: https://github.com/sequelize/sequelize/pull/13689#issuecomment-987812367

Notes:

- error.d.ts includes a copy of the real error.ts file. If we want to avoid duplication we'll need to merge the types folder with the lib folder. I think it's fine to just duplicate and delete the types folder once the migration is complete.
- I've opted for exporting the missing classes instead of marking their export as type only (sadly there is no "export type * from x" and we would have had to export each entry individually).


types/index.d.ts
```diff
@@ -16,7 +16,7 @@ export * from './lib/errors';
 export { BaseError as Error } from './lib/errors';
 export { useInflection } from './lib/utils';
 export { Utils, QueryTypes, Op, TableHints, IndexHints, DataTypes, Deferrable };
-export { Validator as validator } from './lib/utils/validator-extras';
+export { Validator } from './lib/utils/validator-extras';
```
WikiRik, line 19: Could this be seen as a breaking change for some people?

WikiRik: Also, shouldn't there be a test which should be updated for this?

WikiRik (commented):
Look good in general, shouldn't we include some additional tests for this?

types/index.d.ts
ephys: Oh yes, good point. I'll include the old name as well, but deprecated.

As for tests, I'm not sure how we can test that the typings match the JS file. It should not be an issue once the typescript migration is complete though.

WikiRik: Ah, if it's undefined now it should not be a problem.

For the tests; I also wouldn't really know to be honest. Not very experienced on that front. Should also indeed be fixed with the TypeScript migration so not a huge priority.

ephys: https://github.com/sequelize/sequelize/pull/13689 includes a test that ensures no declared export has undefined as a value. I think that + the typescript migration is the best we can

sdepold (approved):

PR #13754, opened 2021-12-09
ci(yarn): ignore engines during yarn installation

sdepold:
Tiny adjustments of #13374


WikiRik:
I removed them here because they run on Node 16 and then we don't need it.

sdepold:
Even better. Means we don't need this change?

WikiRik:
Yup

sdepold:
Perfect!

PR #13755, opened 2021-12-09
ci(release): require snowflake and db2 runs before release

sdepold:
Require snowflake and db2 runs to be successful before releases


WikiRik:
I thought we left this out on purpose since the support is still experimental and we do not have integration tests for Snowflake. But including it can't hurt

sdepold:
Actually looking at how brittle db2 seems to be and the fact that it is elemental, I really agree.

Issue #13722, opened 2021-11-29
Extract dialects out of the core


sdepold:
In order to support additional dialects more easily in the future, we want to pull the dialect specific files out of the core and move them into dedicated npm modules.

Relates to https://github.com/sequelize/meetings/issues/2

```[tasklist]
### Current Progress
- [x] Decide on the API
- [x] Can we have the same set of tests for all built-in dialects?
- [x] https://github.com/sequelize/sequelize/pull/17198
- [ ] https://github.com/sequelize/sequelize/pull/17206
- [x] https://github.com/sequelize/sequelize/pull/17202
- [x] https://github.com/sequelize/sequelize/pull/17190
- [x] https://github.com/sequelize/sequelize/pull/17205
- [ ] https://github.com/sequelize/sequelize/pull/17207
- [ ] https://github.com/sequelize/sequelize/pull/17209
- [x] https://github.com/sequelize/sequelize/pull/17197
- [ ] https://github.com/sequelize/sequelize/pull/17216
- [ ] https://github.com/sequelize/sequelize/pull/17219
- [ ] https://github.com/sequelize/sequelize/pull/17222
- [ ] https://github.com/sequelize/sequelize/pull/17252
- [ ] https://github.com/sequelize/website/pull/742
- [ ] https://github.com/sequelize/sequelize/pull/17269
- [ ] Require the dialect class to be provided instead of a dialect name
- [ ] Add a README.md to all packages
```

ephys:
We're finally working on this, but we still need to iterate on parts of the API

The high level plan is to have individual packages for each dialect, such as `@sequelize/postgres`, `@sequelize/mariadb`, `@sequelize/oracle`, etc

Each package will provide all the dialect-specific code. This means that it will be much easier for the community to create third-party dialect packages, even though they would currently still need to use many unstable internal APIs. This second part is something we'll improve in the future by providing a clean public dialect API, such as `@sequelize/core/dialect-toolbox`

Each package will also include the relevant connector library (pg, mariadb, tedious, etc). Users won't have to figure out which version should be installed anymore.

When initializing Sequelize, users will now need to provide the dialect package instead of a dialect name, but the exact API is still up for debate.

The main design issue surrounds credentials. Credentials are a dialect-specific option, but replication is a Sequelize option that is a list of dialect-specific credentials.

# Option 1: All options are provided to Sequelize itself

```ts
import { PostgresDialect } from '@sequelize/postgres';

new Sequelize({
[...]

PR #13756, opened 2021-12-10
Introduce scope to package name

sdepold:
### Pull Request Checklist

### Description Of Change

With version 7 of Sequelize, we want to split the dialect specific files from the core library. This PR introduces a project scope (@sequelize) and renames the lib to @sequelize/core. The dialects will go into a yet to be defined project but something like @sequelize/postgres-dialect or @sequelize/dialects-postgres seems likely.

Relates to https://github.com/sequelize/sequelize/issues/13722


sdepold:
I'm a little bit concerned if this change is done too prematurely since it will probably cause some trouble during the v6 work. As an alternative, we could do this as late as possible and start with the scoping only for the newly created modules for the moment

PR #13763, opened 2021-12-11
meta(typescript): convert utils to typescript

sdepold:
### Pull Request Checklist

### Description Of Change

Convert utils to typescript.


lib/utils/class-to-invokable.ts
```diff
@@ -1,14 +1,14 @@
...
  * Wraps a constructor to not need the `new` keyword using a proxy.
  * Only used for data types.
  *
- * @param {Function} Class The class instance to wrap as invocable.
+ * @param {ProxyConstructor} Class The class instance to wrap as invocable.
  * @returns {Proxy} Wrapped class instance.
  * @private
  */
-function classToInvokable(Class) {
+export function classToInvokable(Class: any): Proxify {
```
sdepold, line 11: This is a bit too hardcore for my understanding. Theoretically, `Class` should be of type `ProxyConstructor` but that causes problems with the spread operator: 

```A spread argument must either have a tuple type or be passed to a rest parameter.```

**Comment by sdepold** [comment]
Hmmmm. I'm trying to fix the stub but it somehow doesn't work

File: lib/utils/class-to-invokable.ts (hunk 1 above)
Comment (AllAwesome497) [reply lib/utils/class-to-invokable.ts:11]: Something like this should work:

```ts
export function classToInvokable<
  T extends (new () => any) // the () might not be necessary, can't remember. 
>(Class: T): T extends new (...infer Args) => infer Instance 
  ? (...args: Args) => Instance & (new (...args: Args) => Instance) & (Args['length'] extends 0 ? Instance : {})
   : never // this won't ever if  T is a constructor, its just here since we need an `extends` to use `infer`.
 { /* function body */}
```

(haven't verified, may need a bit of tweaking).

lib/utils/join-sql-fragments.ts
```diff
@@ -1,54 +1,55 @@
...
  * Otherwise, if a non-string, non-falsy value is present, a TypeError will be thrown.
- * 
+ *
  * @returns {string} The joined string.
- * 
+ *
  * @private
  */
-function joinSQLFragments(array) {
+export function joinSQLFragments(array: any[]): string {
```
AllAwesome497, line 52: This should be valid (unsure if ts v3.9 will support it):
```ts
type Falsy = 0 | '' | null | undefined | false;
type SQLFragment = string | Falsy | SQLFragment[];
```


w/ the function signature being:
```suggestion
export function joinSQLFragments(array: SQLFragment[]): string {
```

Though this requires a bit of tweaking to get the function working properly (iirc the ts assertions will break a lil - I also converted this file to TS locally and I mostly rewrote the function + optimized it a bit).

lib/utils/join-sql-fragments.ts
```diff
@@ -80,4 +85,8 @@ function joinSQLFragments(array) {
 
   return singleSpaceJoinHelper(array);
 }
-exports.joinSQLFragments = joinSQLFragments;
+
+type JoinSQLFragmentsError = TypeError & {
+  args?: any[];
+  fragment?: any;
+};
```
AllAwesome497, line 92: nit: can you make this extend `Error` for a standard JS error?

ie:
```ts
class JoinSQLFragmentsError extends TypeError {
  args: SQLFragment[];
  fragment: any; // iirc this error is only used when we get an invalid fragment.

  constructor(args: SQLFragment[], fragment: any, message: string) {
    super(message);
    this.args = args;
    this.fragment = fragment;
    // This is good to do explicitly for some bundlers / minnification. 
    this.name = 'JoinSQLFragmentsError';
  }
}
```

lib/utils/class-to-invokable.ts
sdepold: what.. the... hell 😂

test/integration/dialects/abstract/connection-manager.test.js
```diff
@@ -125,10 +142,15 @@ describe(Support.getTestDialectTeaser('Connection Manager'), () => {
     };
 
     await connectionManager.getConnection(queryOptions);
-    chai.expect(deprecationStub).to.have.been.calledOnce;
+    chai.expect(stub).to.have.been.calledOnce;
+    chai
+      .expect(stub.getCalls()[0].args[0])
+      .to.contain(
+        'This database engine version is not supported, please update your database server.'
```
sdepold, line 149: I changed this to this format now. The rest of the file is just formatting.

Instead of checking that the deprecation function was called, I'm checking one level further down where deprecate is calling process.emitWarning.

lib/generic/falsy.ts
```diff
@@ -0,0 +1 @@
+export type Falsy = 0 | '' | null | undefined | false;
```
WikiRik, line 1: Do we want to include -0, 0n and NaN as well? As per https://developer.mozilla.org/en-US/docs/Glossary/Falsy

test/integration/dialects/abstract/connection-manager.test.js
```diff
@@ -1,14 +1,12 @@
...
-  sinon = require('sinon'),
-  Config = require('../../../config/config'),
-  ConnectionManager = require('sequelize/lib/dialects/abstract/connection-manager'),
-  Pool = require('sequelize-pool').Pool;
-
+const chai = require('chai');
+const { unsupportedEngine } = require('../../../../lib/utils/deprecations');
+const Support = require('../../support');
+const sinon = require('sinon');
+const ConnectionManager = require('../../../../lib/dialects/abstract/connection-manager');
```
WikiRik, line 5: Why was it an absolute path before and why did you change it to a relative path? (not knowing why it was this in the first place is a valid answer)

WikiRik (approved):
Don't know about the specific TypeScript questions you posted, but it looks good enough to be merged for me. I do have two questions, but they are non-blocking

lib/generic/falsy.ts
sdepold: Seems that NaN is actually no type (I think it is a number actually)

test/integration/dialects/abstract/connection-manager.test.js
sdepold: fixed

WikiRik (approved):
Looks good enough to be merged, restricting the Class and fragment is non-blocking and something for further improvements

lib/generic/falsy.ts
AllAwesome497: Yea `NaN` doesn't have a constant type in TS and is instead declared as a `number`, despite there being a float32 value specifically for NaN (unsure if JS actually implements this) :/.

PR #13779, opened 2021-12-16
build(sqlite): update node-sqlite3 without tar vulnerability

WikiRik:
As per https://github.com/sequelize/umzug/pull/529 we move to vscode's version of sqlite3 which does not have the node-gyp version with a vulnerable tar dependency; https://github.com/mapbox/node-sqlite3/issues/1493#issuecomment-978912655


WikiRik:
Okay, tests fail because they changed things in their errors. I will fix this, marked it as draft for now

WikiRik:
Changes made are the minimal ones in order to keep supporting the original node-sqlite3 for if we move back

PR #13780, opened 2021-12-16
fix(abstract): patch jsonb operator for pg if value is json

Drethic:
### Pull Request Checklist

### Description Of Change
Issue: When querying a JSONB column in PG and the data stored in the column is an Object PG uses `#>>` when it should be `#>`

Updated abstract/query-generator to check the operator and value type while running _traverseJSON.  If the value is a string and the operator is Op.contains it attempts to parse the value as JSON.  If it doesn't fail it sends true to jsonPathExtractionQuery new third parameter isJson.  This is only used in postgres so it knows to use `#>` instead of `#>>` to join the query string.

### DIFF
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1062,12 +1062,13 @@ class QueryGenerator {
   /**
    * Generates an SQL query that extract JSON property of given path.
    *
-   * @param   {string}               column  The JSON column
-   * @param   {string|Array<string>} [path]  The path to extract (optional)
-   * @returns {string}                       The generated sql query
+   * @param   {string}               column   The JSON column
+   * @param   {string|Array<string>} [path]   The path to extract (optional)
+   * @param   {boolean}              [isJson] The value is JSON use alt symbols (optional)
+   * @returns {string}                        The generated sql query
    * @private
    */
-  jsonPathExtractionQuery(column, path) {
+  jsonPathExtractionQuery(column, path, isJson) {
     let paths = _.toPath(path);
     let pathStr;
     const quotedColumn = this.isIdentifierQuoted(column)
@@ -1102,8 +1103,9 @@ class QueryGenerator {
         return `json_unquote(json_extract(${quotedColumn},${pathStr}))`;
 
       case 'postgres':
+        const join = isJson ? '#>' : '#>>';
         pathStr = this.escape(`{${paths.join(',')}}`);
-        return `(${quotedColumn}#>>${pathStr})`;
+        return `(${quotedColumn}${join}${pathStr})`;
 
       default:
         throw new Error(`Unsupported ${this.dialect} for JSON operations`);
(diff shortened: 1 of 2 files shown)


lib/dialects/abstract/query-generator.js
```diff
@@ -1095,12 +1095,13 @@ class QueryGenerator {
...
-   * @param   {string|Array<string>} [path]  The path to extract (optional)
-   * @returns {string}                       The generated sql query
+   * @param   {string}               column   The JSON column
+   * @param   {string|Array<string>} [path]   The path to extract (optional)
+   * @param   {boolean}              [isJson] The value is JSON use alt symbols (optional)
+   * @returns {string}                        The generated sql query
    * @private
    */
-  jsonPathExtractionQuery(column, path) {
+  jsonPathExtractionQuery(column, path, isJson) {
```
sdepold, line 1104: I think we should rather have a third Paramus which allows to override the joint fragment. What do you think?

lib/dialects/abstract/query-generator.js
```diff
@@ -1135,8 +1136,9 @@ class QueryGenerator {
         return `json_unquote(json_extract(${quotedColumn},${pathStr}))`;
 
       case 'postgres':
+        const join = isJson ? '#>' : '#>>';
```
sdepold, line 1139: So this would become 

const join = options.jointFragment || "#>>";

Drethic: My worry with this option is it won't solve the issue because the fragment is based on data type and you can have a where clause that tries to use both and it will fail because you can only use the default fragment or the option.  Example

Data Structure
```js
const userData = {
  id: DataTypes.STRING,
  name: DataTypes.STRING,
  profile: DataTypes.JSONB,
};
```

Demo Profile
```js
const profile = {
  lastLogin: 2021-12-19T21:26:58.489Z,
  failedLoginAttempts: 2,
  locked: false,
  settings: [
    { name: 'allowEmails', value: true },
    { name: 'allowPasswordReset', value: true },
  ],
};
```

If we want to find out when a user last logged in and if they can reset their password you would need a query similar to this (I'm writing this off the cuff so syntax may not be perfect):

`SELECT * FROM userData WHERE id = '1a2b3c4d' AND profile #>> lastLogin > ${now() + 1 hour} AND profile #> settings @> '[{ name: 'allowPasswordReset', value: true }]';`

If we passed in options.joinFragment = '#>' that query would be:
`SELECT * FROM userData WHERE id = '1a2b3c4d' AND profile #> lastLogin > ${now() + 1 hour} AND profile #> settings @> '[{ name: 'allowPasswordReset', value: true }]';`
because there is no way the query can tell which of the two needs to be replaced.

sdepold: Agreed

sdepold:
Could you add a test that verifies the new query?

Drethic:
> Could you add a test that verifies the new query?

Tests have been added and catches put in place for `mssql` and `snowflake` to ensure CI passes.

sdepold (approved):

sdepold:
:+1:

TrurlMcByte:
in this case https://github.com/sequelize/sequelize/blob/27cd0a4e96c293deb64125edb0c3ae70f9d3d858/lib/dialects/abstract/query-generator.js#L863 also must be like
```item = this.jsonPathExtractionQuery(identifier, path, true);```
because "order by json" is better as "order by json converted to text, including integer and float"

sdepold:
Wanna file a pr?

PR #13782, opened 2021-12-17
refactor: convert utils to typescript

sdepold:
### Pull Request Checklist

### Description Of Change

Conversion of lib/utils to typescript and splitting into smaller chunks.


lib/utils/sequelize-method.ts
```diff
@@ -0,0 +1,101 @@
...
+    this.val = val;
+    this.type = (type || '').trim();
+    this.json = json || false;
+  }
+}
+
+export class Literal extends SequelizeMethod {
+  private val: any;
+
+  constructor(val: any) {
```
ephys, line 53: can `val` be anything other than `string`?

lib/utils/sequelize-method.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+}
+
+export class Where extends SequelizeMethod {
+  private attribute: any;
+  private comparator: any;
+  private logic: any;
+
+  constructor(attribute: any, comparator: any, logic: any) {
```
ephys, line 89:
```suggestion
  constructor(attribute: AttributeType, comparator: string | symbol, logic: LogicType);
  constructor(attribute: AttributeType, logic: LogicType);
  constructor(attribute: any, comparatorOrLogic: string | symbol | LogicType, logic: LogicType | undefined) {
```

we can make use of overloading here
Should we bring over `AttributeType` & `LogicType` from `sequelize.d.ts`?

lib/utils/object.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/**
+ * Returns an new Object which keys are camelized
+ *
+ * @param {object} obj
+ * @returns {string}
+ * @private
+ */
+export function camelizeObjectKeys(obj: { [key: string]: any }) {
```
ephys, line 164: I got carried away and started work on typing that changes the keys to make them camelcased.

It's not finished but all that's missing is listing all the whitespace characters matched by \s & trim.

I can continue working on this later

```typescript
type CamelCaseImpl<S extends string, Separator extends string> =
  S extends `${infer P1}${Separator}${infer P2}${infer P3}`
    ? CamelCaseImpl<`${P1}${Uppercase<P2>}${P3}`, Separator>
  : S;

// must be nested, cannot be CamelCaseImpl<S, '-' | '_' | ' '>
//  or strings like 'first-second_third' will not parse properly.
type Camelize<S extends string> = CamelCaseImpl<CamelCaseImpl<CamelCaseImpl<Trim<S>, '-'>, '_'>, ' '>;

// according to https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.trim
//                <SPACE> <TAB>   <ZWNBSP>
type WhiteSpace = ' ' | '\t' | '\u2060';
// <VT>
// <FF>
// <USP>; // https://www.compart.com/en/unicode/category/Zs

//                    <LF>   <CR>
type LineTerminator = '\n' | '\r';
// <LS>
// <PS>

type Trimable = WhiteSpace | LineTerminator;

type TrimRight<T extends string> = T extends `${infer R}${Trimable}` ? TrimRight<R> : T;
type TrimLeft<T extends string> = T extends `${Trimable}${infer R}` ? TrimLeft<R> : T;
type Trim<T extends string> = TrimRight<TrimLeft<T>>;

const a = {
  test_test: 5,
  '  first--second_third FOURTH   --  test  \t': 10,
};

function camelizeObjectKeys<In extends object>(input: In): {
  [key in keyof In as key extends string ? Camelize<key> : key]: In[key]
} {
  return input;
}

const b = camelizeObjectKeys(a);
// b is {testTest: number, firstSecondThirdFOURTHTest: number}
```

ephys (changes requested):
The majority of my review is about making the types stricker or add hints for TypeScript.

Some of them are based on my understanding of the methods, so a double review is welcome

ephys:
I've resolved the issue you had with this PR being dependant on the DataTypes PR.

Solution was the same as my own ts conversion PR:
- Never import anything from /types. If you need something from /types, import from index.js instead. [Example](https://github.com/sequelize/sequelize/blob/main/lib/transaction.ts#L2)
- Same logic if something in /types needs to import from /lib. [Example](https://github.com/sequelize/sequelize/blob/main/types/lib/model.d.ts#L9)
- I set `allowJs` to false. If you need to import from a JS file, use `require` instead of `import`

If you're ok with it I can push my changes to this branch. They include a resolution for the merge conflict + these changes. It compiles now :)

package.json
```diff
@@ -58,6 +58,10 @@
     "wkx": "^0.5.0"
   },
   "devDependencies": {
+    "@types/inflection": "^1.13.0",
+    "@types/lodash": "^4.14.178",
+    "@types/uuid": "^8.3.3",
+    "js-combinatorics": "^0.6.1",
```
ephys, line 64: The return of `js-combinatorics` is my bad. Merge error.
Pointing it out so we don't forget to remove it.

lib/utils/check.ts
```diff
@@ -0,0 +1,59 @@
+import _ from 'lodash';
+import { getOperators } from './format';
+import { Where } from '../../types/lib/utils';
+
+const DataTypes = require('../../lib/data-types');
+
+export function isPrimitive(val: any) {
```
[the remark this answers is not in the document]
sdepold: May I ask what this is doing? 

How is it different from

```
export function isPrimitive(val: string|number|boolean)
```

yarn.lock
```diff
@@ -1046,6 +1046,11 @@
   resolved "https://registry.yarnpkg.com/@types/geojson/-/geojson-7946.0.8.tgz#30744afdb385e2945e22f3b033f897f76b1f12ca"
   integrity sha512-1rkryxURpr6aWP7R786/UQOkJ3PcpQiWkAXBmdWc7ryFWqN6a4xfK7BtjXvFBKO9LjQ+MWQSWxYeZX1OApnArA==
 
+"@types/inflection@^1.13.0":
```
WikiRik, line 1049: @ephys did you regenerate this file or were these changes done by merging?

package.json
WikiRik: And pin the types so we don't get a renovate PR for it

yarn.lock
ephys: yarn handled the merge operation for that file but added inflection because I forgot to remove it from package.json

lib/utils/check.ts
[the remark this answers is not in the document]
ephys: in this version, you're only able to pass string, number, or boolean to `isPrimitive`
```
export function isPrimitive(val: string|number|boolean);

isPrimitive(new User()); // typing error!
```

in this version, can pass anything to `isPrimitive`, and if the result is `true`, typescript knows the value has to be either string, number, or boolean
```
export function isPrimitive(val): string|number|boolean;

isPrimitive(new User()); // not a typing error
```

Demo of second scenario: https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAEzgZTgWwKZQBYxgDmA6jPgLICeAagIYA2I2AFJrY8wFyICCATvzpUAPHTBUAfIgA+iMCEwAjbPwCUiAN4AoRHsQxgiFjADOAOUUr+bDk2xqNO-S8QB6N4ihUADtlMQ-DA+UIgA1mBwAO6miOz09gaxdPJWqrquevGc2AB0UHAAYjAAHtjILGoA3O6eIgC0XkWl5YjYJWZQsQipyukuAL5tDKbYWhmZHl6+-oHBoRHRsdmJZnyCwhOuK8y5AFZwhJU1Uw2IB4RtHaZdiD0CQlRbA9ov2qCQsD1mln02AG6MHjiKhqHiAhhJXrWcZ6fi4ED8JDePxwIwQxAAXmxiAA5Ao-riqq9tNogA

ephys:
Working a bit more on this currently

I'm changing part of `mapWhereFieldNames` because [this piece of code](https://github.com/sequelize/sequelize/blob/main/lib/utils.js#L219-L222) is going to cause subtle bugs for a model that looks like this:

```typescript
sequelize.define('User', { 
  firstName: {
    type: DataTypes.STRING,
    field: 'name' // this is also the name of another attribute but it should work because User.name uses "name2" as a column name!
  },
  name: {
    type: DataTypes.STRING,
    field: 'name2'
  }
});
```

ephys:
Damnit. I broke TypeScript < 4.4 because they finally allowed `symbol` to be used as keys in 4.4

ephys:
@sdepold can't request a review from you as you're the author but I think it's ready for a review :)

lib/utils/format.ts
```diff
@@ -0,0 +1,256 @@
...
+}
+
+export function mapWhereFieldNames(where: Record<string | symbol, any>, Model: ModelStatic<Model>): object {
+  if (!where) {
+    return where;
+  }
+
+  const newAttributes: Record<string | symbol, any> = Object.create(null);
+  // note on 'as any[]': TypeScript < 4.4 does not support using Symbol for keys.
+  // Cast can be removed in sept. 2022 when we drop support for < 4.4
```
WikiRik, line 110: Maybe we can set an expiring TODO or similar for this? I was looking at [the expiring-todo-comments rule from eslint-plugin-unicorn](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md) earlier, maybe we could implement that?

lib/utils/check.ts
```diff
@@ -0,0 +1,64 @@
+import isEmpty from 'lodash/isEmpty';
+import isPlainObject from 'lodash/isPlainObject';
+import type { DataType } from '../..';
+import { getOperators } from './format';
+// eslint-disable-next-line import/order
```
WikiRik, line 5: Is there a reason that we don't want the requested order?

lib/utils/check.ts
```diff
@@ -0,0 +1,64 @@
...
+import { Where } from './sequelize-method';
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
```
WikiRik, line 8: I assume this will be removed after we have migrated data-types to TS?

lib/utils/format.ts
ephys: We can! Are you ok with me setting it to 'warn' instead of 'error'? I don't want unrelated PRs to start failing because the date expired

lib/utils/check.ts
WikiRik: Will not duplicate this comment over the other files since I think the reason will be the same in each case

lib/utils/check.ts
WikiRik: Will not duplicate this comment over the other files since I think the reason will be the same in each case

lib/utils/dialect.ts
```diff
@@ -0,0 +1,69 @@
+import isPlainObject from 'lodash/isPlainObject';
+// eslint-disable-next-line import/order
+import { v1 as uuidv1, v4 as uuidv4 } from 'uuid';
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const DataTypes = require('../data-types');
+
+const dialects = new Set([
```
WikiRik, line 8: Snowflake not needed for this one?

lib/utils/dialect.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  if (isPlainObject(value)) {
+    return { ...(value as object) };
+  }
+
+  return value;
+}
+
+// Note: Use the `quoteIdentifier()` and `escape()` methods on the
```
WikiRik, line 59: For who is this note?

lib/utils/check.ts
WikiRik: Might add a TODO before this line to mention this? Especially considering my single comment from before about unicorn/expiring-todo-comments

lib/utils/format.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  return newAttributes;
+}
+
+/**
+ * getComplexKeys
+ *
+ * @param  {object} obj
+ * @returns {Array<string|symbol>} All keys including operators
+ * @private
```
WikiRik, line 160: Not really specific to this line, but what was the consensus reached on marking functions as private? Or is this just for the docs generation?

lib/utils/format.ts
```diff
@@ -0,0 +1,256 @@
...
+}
+
+/**
+ * Used to map field names in values
+ *
+ * @param dataValues
+ * @param attributeNames
+ * @param ModelClass
+ */
+export function mapValueFieldNames( // TODO: rename to mapAttributesToColumNames?
```
WikiRik, line 193: Is this something to be addressed in this PR?

lib/utils/format.ts
```diff
@@ -0,0 +1,256 @@
...
+  dataValues: Record<string, any>,
+  attributeNames: string[], // TODO: can attributes be symbols too?
```
WikiRik, line 195: Is this something to be addressed in this PR?

lib/utils/format.ts
```diff
@@ -0,0 +1,256 @@
...
+      const columnName = ModelClass.rawAttributes[attributeName]?.field ?? attributeName;
+
+      values[columnName] = dataValues[attributeName];
+    }
+  }
+
+  return values;
+}
+
+// TODO: rename to removeNullishValues as it also removes undefined.
```
WikiRik, line 212: Is this something to be addressed in this PR?

lib/utils/check.ts
WikiRik: Ah, later on you mention; "-- caused by temporarily mixing require with import". Maybe copy that line over to all files?

lib/utils/check.ts
WikiRik: Later you mention "-- .js files must be imported using require" which solves my answer. Still think the TODO might be nice

lib/utils/object.ts
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * @param a
+ * @param b
+ */
+export function mergeDefaults<T>(a: T, b: Partial<T>): T {
+  return mergeWith(a, b, (objectValue, sourceValue) => {
+    // If it's an object, let _ handle it this time, we will be called again for each property
+    if (!isPlainObject(objectValue) && objectValue !== undefined) {
+      // _.isNative includes a check for core-js and throws an error if present.
+      // Depending on _baseIsNative bypasses the core-js check.
```
WikiRik, line 29: Do you mean _.baseIsNative here (with the period)?

lib/utils/object.ts
```diff
@@ -0,0 +1,215 @@
...
+      } else {
+        result[key] = value;
+      }
+    });
+  }
+
+  return result;
+}
+
+/* eslint-disable consistent-return -- lodash actually wants us to return `undefined` to fallback to the default customizer. */
```
WikiRik, line 76: I think I prefer the usage of eslint-disable-next-line over eslint-disable. Makes us more aware of what we are disabling (and also makes us not forget to enable it again later)

WikiRik: And I really like the explanation afterwards of why it is disabled. We should incorporate this throughout the codebase as a best practice in my opinion.

(And if we decide to do that as a best practice we should also write that down somewhere, but that's for a different PR)

lib/utils/object.ts
WikiRik: Guess this comment can be resolved right?

lib/utils/sequelize-method.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Do not use me directly. Use {@link Sequelize.col}
+ */
+export class Col extends SequelizeMethod {
+  private readonly col: string[] | string;
+
+  constructor(col: string[] | string, ...args: string[]) {
+    super();
+    // TODO(ephys): this does not look right. First parameter is ignored if a second parameter is provided.
```
WikiRik, line 38: Is this something that you want to address in this PR?

lib/utils/sequelize-method.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    if (logic === undefined) {
+      logic = comparatorOrLogic;
+      comparator = '=';
+    } else {
+      comparator = comparatorOrLogic;
+    }
+
+    this.attribute = attribute;
+    // @ts-expect-error
```
WikiRik, line 125: Can we include a comment here as well explaining why?

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "esnext",
     "module": "commonjs",
     "moduleResolution": "node",
-    "allowJs": true,
+    "allowJs": false,
```
WikiRik, line 6: I assume that we want to remove this after the TS migration is done, right?

types/lib/query-interface.d.ts
```diff
@@ -8,14 +8,17 @@ import {
   WhereOptions,
   Filterable,
   Poolable,
-  ModelCtor, ModelStatic, ModelType
```
WikiRik, line 11: ModelCtor is removed here because of an earlier PR, right?

WikiRik (commented):
Started the review before the last commit adding the expiring TODOs but I think most if not all comments are still valid. Some TypeScript concepts are still unclear to me most of the time so I didn't take the best look at the implementation details.

lib/utils/check.ts
ephys: Will add the comment everywhere & add TODO :)

lib/utils/format.ts
WikiRik: Yes, that's a good idea

lib/utils/dialect.ts
ephys: Looks like an oversight from the Snowflake PR, because their DataTypes support milliseconds.

Although really this should be a flag in Dialect.supports.

I'll note it down and do a follow-up PR with it just in case there is something fishy going on.

lib/utils/object.ts
ephys: I haven't added it yet because it's not quite right but I'll back it up to a gist and add it later if it ends up being useful

lib/utils/format.ts
ephys: I think it's for docs generation

lib/utils/format.ts
ephys: Yes, I put the TODO because I wanted external input :)

It's related to https://github.com/sequelize/meetings/issues/17
Or I can just not include it in this PR and list changes in the above issue

lib/utils/object.ts
ephys: It's an internal method from lodash (lodash/_baseIsNative).

lib/utils/format.ts
ephys: I think that one I'm just going to rename without external input. It would match the name of the nullish coalescing operator, so there's precedent

lib/utils/check.ts
WikiRik: Does the plugin still work though? Or do we maybe want to fork it? Was thinking of forking eslint-plugin-small-import since installing your config will give a warning about the peerDependency not matching. Maybe we can make a special plugin with rules unmaintained elsewhere that we use internally?

lib/utils/format.ts
WikiRik: I think this would fit better in that issue then. Or add the RFC issue to the TODO

lib/utils/format.ts
WikiRik: Yeah, I think the updated name is better

lib/utils/object.ts
WikiRik: Ah, gotcha

lib/utils/sequelize-method.ts
ephys: I'd rather do a follow-up PR, to make sure I'm not breaking something deliberate

tsconfig.json
ephys: Correct, it's a temporary solution to resolve issues with types being partially in /types and partially in /lib

types/lib/query-interface.d.ts
ephys: It's the same as ModelStatic now but in this specific case it just was not used, it was added (but not used) in https://github.com/sequelize/sequelize/pull/12405

lib/utils/sequelize-method.ts
ephys: Done :)

lib/utils/check.ts
ephys: `eslint-plugin-small-import` would be an easy fork too, we just need to update its peerDependencies declaration

https://github.com/mysticatea/eslint-plugin-eslint-comments is probably more work but also much more useful for sequelize

lib/utils/format.ts
ephys: I have my answer for this one.
Being able to use Symbols in attributes would require being able to use Symbols in `Model.init` which is not the case, as `Model.init` uses Object.keys to retrieve the list of attributes.
Removed!

lib/utils/object.ts
ephys: Regarding eslint-disable vs eslint-disable-next-line: Me too, but `consistent-return` is a rule that often changes where the error is reported when refactoring. It's easier to disable the whole function

If we fork eslint-plugin-eslint-comments, we could enforce these directives to go in pairs (https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/disable-enable-pair.html)

lib/utils/format.ts
ephys: Done :)

lib/utils/check.ts
ephys: Fixed. Well, documented really :)

lib/utils/dialect.ts
ephys: I can't say. I think it's just a warning to maintainers that may try escaping using that variable?

lib/utils/object.ts
ephys: Oh and we can enforce descriptions too https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/require-description.html

The fact that this page talks about v7 and later makes me think the plugin might still work with eslint 8, I'll give it a try.

lib/utils/dialect.ts
WikiRik: Ah, this was already present in the JS file?

ephys: Yep!

lib/utils/format.ts
sdepold: I was wondering in general if we can remove the bigger parts of those comments since input and output types should be retrievable from the function signature?

ephys: I expected eslint to at least remove types automatically but I forgot to turn `jsdoc/no-types` on (https://github.com/sequelize/sequelize/pull/13930#issuecomment-1013703583)

sdepold: 👌

ephys:
I'll try to resolve the last few issues today

lib/utils/object.ts
ephys: Given that these rules have been enabled, I think I can resolve this conversation now :)

ephys:
Ok, I think this PR is ready now

ephys:
I'd like to get https://github.com/sequelize/sequelize/pull/13990 merged first as this PR impacts the old branded type logic - will make it easier to merge that other one in v6

ephys:
Wooh tests are passing again!
PR ready & up to date :)

src/utils/check.ts
```diff
@@ -0,0 +1,63 @@
...
+export function canTreatArrayAsAnd(arr: unknown[]): arr is Array<object | Where> {
+  return arr.some(arg => isPlainObject(arg) || arg instanceof Where);
+}
+
+/**
+ * Determine if the default value provided exists and can be described
+ * in a db schema using the DEFAULT directive.
+ *
+ * @param value Any default value.
+ * @returns yes / no.
```
sdepold, line 33: Should we go with Boolean here?

src/utils/check.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  return typeof value !== 'function';
+}
+
+/**
+ * Returns true if a where clause is empty, even with Symbols
+ *
+ * @param obj
+ * @returns
```
sdepold, line 58: Boolean too

sdepold:
Lgtm :)

src/utils/check.ts
ephys: We're not specifying types in the jsdoc anymore since TypeDocs will extract the typing information from typescript, but I think we could go with

```
@returns true if a where clause is empty, even with Symbols ; false otherwise
```

src/utils/check.ts
ephys: Same as other comments, I think we can just completely remove this `@returns` since it doesn't provide any information and the return type is extract from ts types

src/utils/check.ts
sdepold: I don't really mind either way tbh. You choose

ephys: Based on what I see, I think it's better to remove the `@returns` altogether if it doesn't add anything that's not already in the description

image

ephys:
There is something fishy going on in this PR. Some of our tests are randomly failing. I've seen sqlite fail, then postgres in a re-run.

ephys:
For future me: https://github.com/sequelize/sequelize/pull/13992/files#diff-3274f1a37032fb0ae4e2823def0007c634e869ae0dfc304ff6a12c36513c3a52R271 this is the code that I need to include in the new utils when resolving the new conflict

sdepold:
> There is something fishy going on in this PR. Some of our tests are randomly failing. I've seen sqlite fail, then postgres in a re-run.

In also some failing tests With mariadb and than something else on my other pr

ephys:
So basically every dialect, but at different times. Very weird.

I'll keep a list of failing suites here so we can investigate whether they have something in common:
https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true
https://github.com/sequelize/sequelize/runs/5152023145?check_suite_focus=true
https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true

WikiRik:
> I'll keep a list of failing suites here so we can investigate whether they have something in common: https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true https://github.com/sequelize/sequelize/runs/5152023145?check_suite_focus=true https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true

The logs are only available for a limited period (I thought 1 week but haven't checked) so download them before they're gone (or fix it before ofc)

ephys:
I've copy-pasted the relevant CI logs in my previous comment to make sure we don't lose them

I've also resolved the merge conflict

@sdepold if you have time, do you mind if I let you take over the remaining of this PR now that the TS part is done?

sdepold:
```
1) [SQLITE] Transaction
[2387](https://github.com/sequelize/sequelize/runs/5152023145?check_suite_focus=true#step:6:2387)
[2388](https://github.com/sequelize/sequelize/runs/5152023145?check_suite_focus=true#step:6:2388)
     AssertionError: expected promise to be rejected with an error including 'SQLITE_BUSY: database is locked' but it was fulfilled with [ undefined, undefined ]
[2389](https://github.com/sequelize/sequelize/runs/5152023145?check_suite_focus=true#step:6:2389)
      at assertIfNotNegated (node_modules/chai-as-promised/lib/chai-as-promised.js:68:19)
[2390](https://github.com/sequelize/sequelize/runs/5152023145?check_suite_focus=true#step:6:2390)
      at /home/runner/work/sequelize/sequelize/node_modules/chai-as-promised/lib/chai-as-promised.js:183:17
[2391](https://github.com/sequelize/sequelize/runs/5152023145?check_suite_focus=true#step:6:2391)
      at async Context.<anonymous> (test/integration/transaction.test.js:739:9)
[2392](https://github.com/sequelize/sequelize/runs/5152023145?check_suite_focus=true#step:6:2392)
```

```
1) [POSTGRES Specific] DAO
[2865](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2865)
       enums
[2866](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2866)
         ARRAY(ENUM)
[2867](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2867)
           should be able to insert a new record even with an array of enums in a schema:
[2868](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2868)
     schema "special_schema" does not exist
[2869](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2869)
  Error
[2870](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2870)
      at Query.run (lib/dialects/postgres/query.js:55:25)
[2871](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2871)
      at /home/runner/work/sequelize/sequelize/lib/sequelize.js:315:28
[2872](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2872)
      at runMicrotasks (<anonymous>)
[2873](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2873)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
[2874](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2874)
      at async /home/runner/work/sequelize/sequelize/lib/dialects/postgres/query-interface.js:104:76
[2875](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2875)
      at async /home/runner/work/sequelize/sequelize/lib/dialects/postgres/query-interface.js:104:96
[2876](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2876)
      at async PostgresQueryInterface.ensureEnums (lib/dialects/postgres/query-interface.js:104:20)
[2877](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2877)
      at async PostgresQueryInterface.createTable (lib/dialects/abstract/query-interface.js:85:5)
[2878](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2878)
      at async Function.sync (lib/model.js:938:5)
[2879](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2879)
      at async Context.<anonymous> (test/integration/dialects/postgres/dao.test.js:490:11)
[2880](https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true#step:9:2880)
```

```
1) [MARIADB] Model
[2585](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2585)
       scope
[2586](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2586)
         complex merge
[2587](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2587)
           should merge complex scopes with findAll options correctly regardless of their order:
[2588](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2588)

[2589](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2589)
      AssertionError: expected { id: 1, name: 'foo1', bars: [ …(2) ] } to deeply equal { id: 1, name: 'foo1', bars: [ …(2) ] }
[2590](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2590)
      + expected - actual
[2591](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2591)

[2592](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2592)
           {
[2593](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2593)
             "bazs": [
[2594](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2594)
               {
[2595](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2595)
                 "barId": 1
[2596](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2596)
      -          "id": 2
[2597](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2597)
      +          "id": 1
[2598](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2598)
                 "quxes": [
[2599](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2599)
                   {
[2600](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2600)
      -              "bazId": 2
[2601](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2601)
      -              "id": 1
[2602](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2602)
      -              "name": "qux3"
[2603](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2603)
      +              "bazId": 1
[2604](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2604)
      +              "id": 3
[2605](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2605)
      +              "name": "qux1"
[2606](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2606)
                   }
[2607](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2607)
                   {
[2608](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2608)
      -              "bazId": 2
[2609](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2609)
      -              "id": 2
[2610](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2610)
      -              "name": "qux4"
[2611](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2611)
      +              "bazId": 1
[2612](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2612)
      +              "id": 4
[2613](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2613)
      +              "name": "qux2"
[2614](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2614)
                   }
[2615](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2615)
                 ]
[2616](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2616)
               }
[2617](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2617)
               {
[2618](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2618)
                 "barId": 1
[2619](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2619)
      -          "id": 1
[2620](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2620)
      +          "id": 2
[2621](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2621)
                 "quxes": [
[2622](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2622)
                   {
[2623](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2623)
      -              "bazId": 1
[2624](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2624)
      -              "id": 3
[2625](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2625)
      -              "name": "qux1"
[2626](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2626)
      +              "bazId": 2
[2627](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2627)
      +              "id": 1
[2628](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2628)
      +              "name": "qux3"
[2629](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2629)
                   }
[2630](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2630)
                   {
[2631](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2631)
      -              "bazId": 1
[2632](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2632)
      -              "id": 4
[2633](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2633)
      -              "name": "qux2"
[2634](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2634)
      +              "bazId": 2
[2635](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2635)
      +              "id": 2
[2636](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2636)
      +              "name": "qux4"
[2637](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2637)
                   }
[2638](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2638)
                 ]
[2639](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2639)
               }
[2640](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2640)
             ]
[2641](https://github.com/sequelize/sequelize/runs/5156282992?check_suite_focus=true#step:8:2641)
```

ephys:
The MariaDB test has failed in another PR https://github.com/sequelize/sequelize/runs/5320675964?check_suite_focus=true, which means it's also unrelated to this PR

Only the postgres one remains

WikiRik:
I have seen the Postgres one as well I believe. At least the 'schema does not exist' in DAO tests. Not sure about this specific test

ephys:
Then I think we can merge this PR. I'll resolve the merge conflict

src/utils/sequelize-method.ts
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Do not use me directly. Use {@link Sequelize.col}
+ */
+export class Col extends SequelizeMethod {
+  private readonly col: string[] | string;
+
+  constructor(col: string[] | string, ...args: string[]) {
+    super();
+    // TODO(ephys): this does not look right. First parameter is ignored if a second parameter is provided.
```
WikiRik, line 38: This is for a future PR, right?

src/utils/sequelize-method.ts
```diff
@@ -0,0 +1,135 @@
...
+      this.conditions = conditionsOrPath;
+    }
+  }
+}
+
+/**
+ * Do not use me directly. Use {@link Sequelize.where}
+ */
+export class Where<Operator extends keyof WhereOperators = typeof Op.eq> extends SequelizeMethod {
+  // TODO [=7]: rename to leftOperand after typescript migration
```
WikiRik, line 106: Isn't this after the TypeScript migration? Or when the TypeScript migration is completely done?

WikiRik (commented):
Few small remarks/questions before we merge this

src/utils/sequelize-method.ts
ephys: I think when I first wrote this the idea was to wait until the TS migration was completely done, so typescript can tell us which parts of the code need to be updated.
But any follow-up PR should be fine, out existing unit tests should be able to cover this change

src/utils/sequelize-method.ts
ephys: Yep! It's in my backlog but something I want to take a look at in a different PR

src/utils/dialect.ts
```diff
@@ -0,0 +1,69 @@
+import isPlainObject from 'lodash/isPlainObject';
+// eslint-disable-next-line import/order -- caused by temporarily mixing require with import
+import { v1 as uuidv1, v4 as uuidv4 } from 'uuid';
+
+const DataTypes = require('../data-types');
+
+const dialects = new Set([
```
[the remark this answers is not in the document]
ephys: Not sure. This variable name is terrible because based on usage, it looks like it's used to mean "dialects that support milliseconds in their datatypes"

package.json
```diff
@@ -56,14 +56,17 @@
     "wkx": "^0.5.0"
   },
   "devDependencies": {
-    "@rushstack/eslint-patch": "1.1.0",
     "@commitlint/cli": "16.2.1",
     "@commitlint/config-angular": "16.2.1",
     "@ephys/eslint-config-typescript": "14.1.1",
+    "@rushstack/eslint-patch": "1.1.0",
     "@types/chai": "4.3.0",
+    "@types/inflection": "1.13.0",
```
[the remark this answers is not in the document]
ephys: Only if they're used by the generated `types` folder
Checked right now and we need to include the types for these 3 packages:
image

src/utils/dialect.ts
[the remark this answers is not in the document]
WikiRik: Yes, let's update this here. Or at least add a comment to this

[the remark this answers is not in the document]
ephys: Done! I've also added a comment on how to rework this function when we work on splitting dialects from core

src/utils/format.ts
```diff
@@ -0,0 +1,269 @@
...
+        return [columnName, attributeName];
+      }
+
+      return attributeName;
+    });
+  }
+
+  if (options.where != null && isPlainObject(options.where)) {
+    // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error -- this fails in TS 4.4 and up, but not before
+    // @ts-ignore the work necessary to type the return type of mapWhereFieldNames is not worth it
```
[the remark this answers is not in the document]
ephys: I think it's something to evaluate as we migrate parts of the codebase that use `mapWhereFieldNames`. If it's useful to these parts, let's migrate. Otherwise the complex typings it would require aren't worth it

WikiRik (approved):
Looks good!

Issue #13787, opened 2021-12-18
ERR_PACKAGE_PATH_NOT_EXPORTED error on startup with 6.12.0 on Node 12 or 14


EverlastEngineering:
## Issue Creation Checklist

[x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

### What do you expect to happen?

_I want Node 12 or Node 14 to continue to start when using 6.12.0._

### What is actually happening?

_We receive this error on startup after an npm update today:_

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in /Users/xxx/Source/xxx-service/WebContent/node_modules/sequelize/package.json
    at throwExportsNotFound (internal/modules/esm/resolve.js:285:9)
    at packageExportsResolve (internal/modules/esm/resolve.js:508:3)
    at resolveExports (internal/modules/cjs/loader.js:450:36)
    at Function.Module._findPath (internal/modules/cjs/loader.js:490:31)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:888:27)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:92:18)
    at Object.<anonymous> (/Users/xxx/Source/xxx-service/WebContent/node_modules/sequelize-values/lib/patches.js:9:32)
    at Module._compile (internal/modules/cjs/loader.js:1085:14) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
```

### Additional context

Manually specifying 6.11.0 works fine. I have not updated node yet since I am mirroring a deployed instance on AWS for support purposes. 

### Environment

- Sequelize version: sequelize@6.12.0  
- Node.js version: v14.17.1 or 12.22.7 (untested on others) 

### How does this problem relate to dialects?

- [x] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
[...]

PR #13795, opened 2021-12-20
fix: allow deep imports

ephys:
### Pull Request Checklist

- [N/A] Did you update the typescript typings accordingly (if applicable)?

### Description Of Change

Fixes https://github.com/sequelize/sequelize/issues/13787

The issue was that once `exports` is defined in package.json, node uses it as the source of truth for all valid exports of a package.

I've configured `exports` to replicate how the imports worked in v6.  
I also had to remove the alias in `registerEsbuild.js` as it was redundant with `exports`.  
It also matches reality better as the tests use the same export definition as the published build. :)

I recommend changing the export declaration for V7, see below

---

## Breaking changes recommendations for V7

In v7, we could change the export definition to

```json
{
  "exports": {
    ".": {
      "import": "./dist/index.mjs",
      "require": "./dist/index.js"
    },
    "./package.json": "./package.json",
    "./*": "./dist/lib/*.js"
  },
}
```

This way, instead of importing `sequelize/lib/model.js`, users could import `sequelize/model`.

If we still want users to specify `/lib` in v7, we can use this one instead

```json
{
  "exports": {
    ".": {
      "import": "./dist/index.mjs",
      "require": "./dist/index.js"
    },
    "./lib/*": "./dist/lib/*.js",
    "./package.json": "./package.json",
  },
}
```

**[recommended]** If we want to prevent users from importing `/lib` which could result in unexpected breaking changes, we could instead use this one which only allows importing `sequelize` & `sequelize/package.json`

```json
{
  "exports": {
    ".": {
      "import": "./dist/index.mjs",
      "require": "./dist/index.js"
    },
    "./package.json": "./package.json",
  },
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,8 @@
       "import": "./dist/index.mjs",
       "require": "./dist/index.js"
     },
+    "./lib/*": "./dist/lib/*.js",
+    "./lib/errors": "./dist/lib/errors/index.js",
     "./*": "./*"
   },
   "engines": {
--- a/package.json
+++ b/package.json
@@ -20,8 +20,11 @@
   "types": "./dist",
   "type": "commonjs",
   "exports": {
-    "import": "./dist/index.mjs",
-    "require": "./dist/index.js"
+    ".": {
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.js"
+    },
+    "./*": "./*"
   },
   "engines": {
     "node": ">=10.0.0"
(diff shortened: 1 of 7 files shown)


sdepold:
Completely agreed on the fact that people would better not import anything from the package other than the main entry point.

ephys:
I'll PR v7 to completely forbid it once this is merged :)

sdepold:
Thanks a bunch. I think I'll just release a new version straight away or was there anything else we need to fix?

ephys:
@sdepold Yep!

Issue #13791, opened 2021-12-18
Typescript 4.5.4 does not like sequelize-ESM imports in v6.12


dilyanpalauzov:
My understanding of https://github.com/sequelize/sequelize/pull/13689, included in v6.12 is, that sequelize is not any more exported just in CommonJS, but can be used also as ESM-imports.  Thus, the code in v6.11
```JavaScript
import sequelize_ from 'sequelize'
const {DataTypes, Model, Op} = sequelize_ //this is CommonJS hack
```
can be changed in 6.12 to `import {DataTypes, Model, Op} from 'sequelize'`.

Typescript 4.5.4 fails on the above code with
```
$ node_modules/.bin/tsc
node_modules/sequelize/lib/operators.ts:524:1 - error TS1203: Export assignment cannot be used when targeting ECMAScript modules. Consider using 'export default' or another module format instead.

524 export = Op;
    ~~~~~~~~~~~~

Found 1 error.
```
My understanding is, that Typescript is only supposed to verify the types in sequelize by parsing the code in the `node_modules/sequelize/types` directory, and not loose time to verify each file in `node_modules/sequelize/lib/`.  But `types/index.d.ts` contains `import * as Op from "../lib/operators";` so my assumption is not correct.  Likewise for `node_modules/sequelize/types/lib/model.d.ts:import * as Op from '../../lib/operators';`

### How does this problem relate to dialects?

- [x] I think this problem happens regardless of the dialect.

### Would you be willing to resolve this issue by submitting a Pull Request?
- [x] No, I don't have the time and I wouldn't even know how to start.

WikiRik:
@ephys can you take a look at this?

shoeyn:
I'm having this issue also since the 6.12 release

Changing `export = Op;` to `export default Op;`
Then changing cjs imports from `require('./operators')` to `require('./operators').default`
and esm imports from `import * as Op` to `import Op`
seems to pass typescript tests, but I don't know if there's any other consequences and whether this sort of change conforms with your code style.

ephys:
@WikiRik Yep! I'll look into this :)

ephys:
Release 6.12.1 should fix this as we've reverted to the previous import style.

If it does not, my recommendation would be to use the [skipLibCheck](https://www.typescriptlang.org/tsconfig/skipLibCheck.html) option until we migrate to esm import/exports in v7

shoeyn:
[...]

PR #13796, opened 2021-12-20
fix: make operator.ts export style match import style

ephys:
### Pull Request Checklist

- [?] Have you added new tests to prevent regressions?

### Description Of Change

Attempt to fix https://github.com/sequelize/sequelize/issues/13791

There was an issue in PR https://github.com/sequelize/sequelize/pull/13731 where `Op` was exported using `export =` but was imported using `import Op from './operator'` which is not compatible (source https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require).

This is a simple fix that changes the export to be `export default Op` instead.

---

### Potential Breaking change

The change of export style from `export = ` to `export default` can be considered a breaking change as users that did `const Op = require('sequelize/lib/operators')` are going to have to change to `const Op = require('sequelize/lib/operators').default` (or better: `const { Op } = require('sequelize')`). 

I don't know if anyone does this, but if we want to be sure to avoid a breaking change here (also the case for the rest of the TypeScript migration), we should avoid changing the export style from `export =` to `export default`.

This also means sticking with `import Op = require('./lib/operators')` too, for now.

We can revisit this in v7.

### eslint not linting

I have also reverted a change in `index.d.ts` where the quotes were changed from single to double quotes which goes against the eslint config of the project. 

That's because eslint is not configured to run on .d.ts. We should fix that by changing `lint` from `"lint": "eslint lib test --quiet --fix",` to `"lint": "eslint lib test --quiet --fix --ext .js,.mjs,.cjs,.ts,.mts,.cts,.d.ts",` (yes file extensions for JS are getting wild).

### Two operators.ts definitions

- `index.d.ts` imports `'../lib/operators'` which points to `sequelize/lib/operators.ts` instead of `sequelize/dist/lib/operators.d.ts`. This could be a problem. I don't have a solution other than "finish the typescript migration"

### DIFF
--- a/lib/associations/belongs-to-many.js
+++ b/lib/associations/belongs-to-many.js
@@ -9,7 +9,7 @@ const HasMany = require('./has-many');
 const HasOne = require('./has-one');
 const AssociationError = require('../errors').AssociationError;
 const EmptyResultError = require('../errors').EmptyResultError;
-const Op = require('../operators');
+const Op = require('../operators').default;
 
 /**
  * Many-to-many association with a join table.
--- a/lib/associations/belongs-to.js
+++ b/lib/associations/belongs-to.js
@@ -4,7 +4,7 @@ const Utils = require('./../utils');
 const Helpers = require('./helpers');
 const _ = require('lodash');
 const Association = require('./base');
-const Op = require('../operators');
+const Op = require('../operators').default;
 
 /**
  * One-to-one association
--- a/lib/associations/has-many.js
+++ b/lib/associations/has-many.js
@@ -4,7 +4,7 @@ const Utils = require('./../utils');
 const Helpers = require('./helpers');
 const _ = require('lodash');
 const Association = require('./base');
-const Op = require('../operators');
+const Op = require('../operators').default;
 
 /**
  * One-to-many association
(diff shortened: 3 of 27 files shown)


WikiRik:
About your first concern, that won't really work. `index.d.ts` is in the types folder which is ignored by eslint; per https://github.com/sequelize/sequelize/blob/b97f93469bee6c3429d8cab7d0cb6418cbf05892/.eslintrc.json#L132

After much of the TypeScript conversion has been done we'll look at the eslint config; https://github.com/sequelize/sequelize/projects/3#card-73929575

I'm also not sure about your second concern. Maybe our TypeScript people have answers, I've requested their reviews

Something I'm concerned about is if this is not a breaking change. As I remembered the syntax for a default import is different, so would people not be affected by this?

ephys:
> Something I'm concerned about is if this is not a breaking change. As I remembered the syntax for a default import is different, so would people not be affected by this?

@WikiRik I was just updating the list of concerns as you wrote this :)

Yes I believe this is a breaking change. If we want to avoid introducing breaking changes with the TypeScript migration we need to stick to `export = ` & `import x = require()`.

I want to avoid that so I think I'll close this PR and open a new one that simply reverts the import style from #13731 back to `ìmport Op = require()`.

I propose we update the import style starting with v7?

WikiRik:
Yes, sounds good to start updating the import style in v7. Looking forward to the new PR

PR #13797, opened 2021-12-20
fix: fix invalid ts import style of lib/operators

ephys:
### Pull Request Checklist

### Description Of Change

Replacement PR of https://github.com/sequelize/sequelize/pull/13796

Attempt to fix https://github.com/sequelize/sequelize/issues/13791

There was an issue in PR https://github.com/sequelize/sequelize/pull/13731 where `Op` was exported using `export =` but was imported using `import Op from './operator'` which is not compatible (source https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require).

This is a simple fix that changes the imports back to `import Op = require('./lib/operators')` syntax.

This one does not include any lint correction either as types/* is not linted and will be removed anyway (https://github.com/sequelize/sequelize/pull/13796#issuecomment-998023312)

---

## Concerns

Things I noticed while working on this PR. Putting them here to decide between fixing them here or opening an issue.

### Avoiding breaking changes

As talked about in https://github.com/sequelize/sequelize/pull/13796, we should stick to [this import/export syntax](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) in v6 to avoid breaking changes.

We can revisit this in v7.

### Two operators.ts definitions

`index.d.ts` imports `'../lib/operators'` which points to `sequelize/lib/operators.ts` instead of `sequelize/dist/lib/operators.d.ts`. 

This could be a problem.

A potential solution to import /lib from /types would be to change the import from `../lib/operators` to `sequelize/lib/operators`.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,6 +1,6 @@
 import DataTypes = require("./lib/data-types");
 import Deferrable = require("./lib/deferrable");
-import Op = require('../lib/operators');
+import Op = require("../lib/operators");
 import QueryTypes = require("./lib/query-types");
 import TableHints = require("./lib/table-hints");
 import IndexHints = require("./lib/index-hints");
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,6 +1,6 @@
 import DataTypes = require("./lib/data-types");
 import Deferrable = require("./lib/deferrable");
-import * as Op from "../lib/operators";
+import Op = require('../lib/operators');
 import QueryTypes = require("./lib/query-types");
 import TableHints = require("./lib/table-hints");
 import IndexHints = require("./lib/index-hints");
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -1,6 +1,6 @@
 import DataTypes = require("./lib/data-types");
 import Deferrable = require("./lib/deferrable");
-import * as Op from "../lib/operators";
+import Op = require('../lib/operators');
```
WikiRik, line 3:
```suggestion
import Op = require("../lib/operators");
```

WikiRik: Just to make it consistent with the rest of the requires, but we'll fix it later with the updated eslint config

WikiRik (dismissed):
Looks good, I did make a suggested change but it's inconsistent throughout the files either way si just see what you do with it

ephys:
I don't think the failed test is related to this PR. Very strange. Can we try relaunching that test?

types/index.d.ts
ephys: Changed :)

WikiRik (approved):

WikiRik:
@sdepold this together with https://github.com/sequelize/sequelize/pull/13795 might be good for 6.12.1, agree?

sdepold:
v6.12.1 is on it's way in this very moment :)

ephys:
🤞

PR #13798, opened 2021-12-21
refactor(dialect): remove all switch dialect in core

jesse23:
### Pull Request Checklist

### Description Of Change

Refactor all `switch(dialect)` in core

### Todos ( by follow up PR )

- `sequelize.random` needs more refactor
- `data-types` needs to be refactored to decouple from its own arch, into sequelize.js

### Approach
- when create a dialect, do `registerDialectImpl('myDialect', MyDialctClass)` in ur dialect.js.
- when customer use the new dialect, they just need to import the new dialect, then call `new Sequelize()` as today.
- For existing dialect, after we refactor them to sub package in monorepo, we can either remain the default registration, as impl in this PR, or we can let user run the additional import at their side ( as breaking change).

### DIFF
--- a/lib/dialects/abstract/registry.js
+++ b/lib/dialects/abstract/registry.js
@@ -1,0 +1,40 @@
+
+const _registries = Object.create(null);
+
+/**
+ * Register dialect implementation.
+ *
+ * @param {AbstractDialect} [DialectClass] dialect implemetation as subclass of AbstractDialect.
+ */
+function registerDialect( DialectClass ) {
+  const dialectName = DialectClass.getDialectName();
+  if (_registries[dialectName] != null && _registries[dialectName] !== DialectClass) {
+    throw new Error(`Dialect '${dialectName}' has been registered!`);
+  }
+  _registries[dialectName] = DialectClass;
+}
+
+/**
+ * Get dialect implementation by name.
+ *
+ * @param {string}  [name] the name of dialect.
+ * @returns {AbstractDialect} dialect implemetation as subclass of AbstractDialect if implementation exists
+ */
+function getDialect( name ) {
+  return _registries[name];
+}
+
+/**
+ * Get all available dialect names.
+ *
+ * @returns {Array<string>} available dialect names.
+ */
+function getRegisteredDialects() {
+  return Object.keys(_registries);
+}
+
+module.exports = {
+  getDialect,
+  getRegisteredDialects,
+  registerDialect
+};
--- a/lib/dialects/db2/index.js
+++ b/lib/dialects/db2/index.js
@@ -9,6 +9,10 @@ const DataTypes = require('../../data-types').db2;
 const { Db2QueryInterface } = require('./query-interface');
 
 class Db2Dialect extends AbstractDialect {
+  static getDialectName() {
+    return 'db2';
+  }
+
   constructor(sequelize) {
     super();
     this.sequelize = sequelize;
--- a/test/unit/dialect-module-configuration.test.js
+++ b/test/unit/dialect-module-configuration.test.js
@@ -4,7 +4,9 @@ const chai = require('chai'),
   expect = chai.expect,
   path = require('path'),
   Support = require(`${__dirname}/support`),
+  DataTypes = require('sequelize/lib/data-types'),
   Sequelize = Support.Sequelize,
+  registerDialect = Support.Sequelize.registerDialect,
   dialect = Support.getTestDialect();
 
 describe(Support.getTestDialectTeaser('Sequelize'), () => {
@@ -52,6 +54,32 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
           dialectModulePath: '/foo/bar/baz'
         });
       }).to.throw('Unable to find dialect at /foo/bar/baz');
+    });
+
+    it('initialization fails for unsupported dalect', () => {
+      registerDialect( {
+        getDialectName: () => 'supportedDialect'
+      });
+      DataTypes.supportedDialect = DataTypes.mysql;
+
+      expect(() => {
+        new Sequelize('dbname', 'root', 'pass', {
+          port: 999,
+          dialect: 'unsupportedDialect'
+        });
+      }).to.throw('The dialect unsupportedDialect is not supported. Supported dialects: mysql, db2, mssql, sqlite, mariadb, postgres, snowflake, supportedDialect.');    
+    });
+
+    // NOTE: disable this test before we refactor data-types
+    xit('initialization success for supported dalect', () => {
+      registerDialect( {
+        getDialectName: () => 'supportedDialect'
+      });
+
+      new Sequelize('dbname', 'root', 'pass', {
+        port: 999,
+        dialect: 'supportedDialect'
+      });
     });
   });
 });
--- a/index.mjs
+++ b/index.mjs
@@ -4,6 +4,8 @@ export default Pkg;
 
 // export * from './lib/sequelize';
 export const Sequelize = Pkg.Sequelize;
+export const registerDialect = Pkg.registerDialect;
+
 export const fn = Pkg.fn;
 export const col = Pkg.col;
 export const cast = Pkg.cast;
--- a/lib/dialects/abstract/query-generator.js
+++ b/lib/dialects/abstract/query-generator.js
@@ -1109,47 +1109,8 @@ class QueryGenerator {
    * @private
    */
   jsonPathExtractionQuery(column, path, isJson) {
-    let paths = _.toPath(path);
-    let pathStr;
-    const quotedColumn = this.isIdentifierQuoted(column)
-      ? column
-      : this.quoteIdentifier(column);
 
-    switch (this.dialect) {
-      case 'mysql':
-      case 'mariadb':
-      case 'sqlite':
-        /**
-         * Non digit sub paths need to be quoted as ECMAScript identifiers
-         * https://bugs.mysql.com/bug.php?id=81896
-         */
-        if (this.dialect === 'mysql') {
-          paths = paths.map(subPath => {
-            return /\D/.test(subPath)
-              ? Utils.addTicks(subPath, '"')
-              : subPath;
-          });
-        }
-
-        pathStr = this.escape(['$']
-          .concat(paths)
-          .join('.')
-          .replace(/\.(\d+)(?:(?=\.)|$)/g, (__, digit) => `[${digit}]`));
-
-        if (this.dialect === 'sqlite') {
-          return `json_extract(${quotedColumn},${pathStr})`;
-        }
-
-        return `json_unquote(json_extract(${quotedColumn},${pathStr}))`;
-
-      case 'postgres':
-        const join = isJson ? '#>' : '#>>';
-        pathStr = this.escape(`{${paths.join(',')}}`);
-        return `(${quotedColumn}${join}${pathStr})`;
-
-      default:
-        throw new Error(`Unsupported ${this.dialect} for JSON operations`);
-    }
+    throw new Error(`Unsupported ${this.dialect} for JSON operations`);
   }
 
   /*
--- a/lib/dialects/db2/query-generator.js
+++ b/lib/dialects/db2/query-generator.js
@@ -884,7 +884,6 @@ class Db2QueryGenerator extends AbstractQueryGenerator {
   quoteIdentifier(identifier, force) {
     return Utils.addTicks(Utils.removeTicks(identifier, '"'), '"');
   }
-
 }
 
 // private methods
--- a/lib/dialects/index.js
+++ b/lib/dialects/index.js
@@ -1,0 +1,13 @@
+/**
+ * Helpers to load OOTB dialect by default. It could be delete and control by customer when we 
+ * refactor all the dialect as separate lib like '@sequelize/db2'
+ */
+const { registerDialect } = require('./abstract/registry');
+
+registerDialect(require('./mysql'));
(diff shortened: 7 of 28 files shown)


lib/dialects/db2/index.js
```diff
@@ -62,4 +63,7 @@ Db2Dialect.prototype.TICK_CHAR_LEFT = '"';
 Db2Dialect.prototype.TICK_CHAR_RIGHT = '"';
 Db2Dialect.prototype.DataTypes = DataTypes;
 
+// register dialect implementation
+registerDialectImpl('db2', Db2Dialect);
```
ephys, line 67: I am not a fan of imports with side effects.

I think it would be better if users did

```typescript
import MyCustomDialect from 'my-custom-dialect-package';

Sequelize.registerDialect('dialectName', MyCustomDialect);
```

instead of

```typescript
import 'my-custom-dialect-package';
```

It's slightly more verbose but it's explicit that something is happening.

The `sequelize.js` file can still register the existing dialect for backward compatibility of course.

lib/dialects/abstract/utils.js
```diff
@@ -0,0 +1,37 @@
+
+const _dialects = {};
+
+/**
+ * Register dialect implementation.
+ *
+ * @param {string}  [name] the name of dialect.
+ * @param {AbstractDialect} [dialectClass] dialect implemetation as subclass of AbstractDialect.
+ */
+function registerDialectImpl( name, dialectClass ) {
```
ephys, line 10: I would recommend removing the `name` parameter to, and instead add it as a static property of `AbstractDialect`:

```typescript
class Db2Dialect extends AbstractDialect {
  static getDialectName() {
    return 'db2';
  }
}
```

this way users can simply do

```
Sequelize.registerDialect(MyCustomDialect);
```

without specifying the name themselves (this is related to the comment regarding side effects in imports)

ephys (dismissed):
Some of my comments are opinionated, lmk what you think :)

jesse23:
> Some of my comments are opinionated, lmk what you think :)

Thx...that is helpful. Addressed

lib/dialects/abstract/utils.js
```diff
@@ -0,0 +1,37 @@
+
+const _dialects = {};
+
```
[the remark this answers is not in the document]
ephys: Note that I don't mean all modified files, just newly created ones :)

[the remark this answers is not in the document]
jesse23: Let me do that, as follow up, after this....I feel it needs to be changed with other file in /abstract together...

lib/dialects/abstract/utils.js
```diff
@@ -0,0 +1,36 @@
...
```
sdepold, line 1: I wonder if we want to rename this file to registry.js?

lib/dialects/abstract/utils.js
```diff
@@ -0,0 +1,36 @@
...
+const _dialects = {};
```
sdepold, line 2:
```
const registry = {}
```

?

lib/dialects/abstract/utils.js
```diff
@@ -0,0 +1,36 @@
...
+function getDialect( name ) {
+  return _dialects[name];
+}
+
+/**
+ * Get all available dialect names.
+ *
+ * @returns {Array<string>} available dialect names.
+ */
+function getAvailableDialects() {
```
sdepold, line 28: `getRegisteredDialects` ?

lib/dialects/abstract/utils.js
```diff
@@ -0,0 +1,36 @@
+
+const _dialects = {};
+
+/**
+ * Register dialect implementation.
+ *
+ * @param {AbstractDialect} [DialectClass] dialect implemetation as subclass of AbstractDialect.
+ */
+function registerDialect( DialectClass ) {
+  _dialects[DialectClass.getDialectName()] = DialectClass;
```
sdepold, line 10: I think we should extend AbstractDialect to have a getDialectName function that needs to be overridden?

test/unit/dialect-module-configuration.test.js
```diff
@@ -53,5 +55,31 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+    it('initialization fails for unsupported dalect', () => {
+      registerDialect( {
+        getDialectName: () => 'supportedDialect'
+      });
+      DataTypes.supportedDialect = DataTypes.mysql;
+
+      expect(() => {
+        new Sequelize('dbname', 'root', 'pass', {
+          port: 999,
+          dialect: 'unSupportedDialect'
```
sdepold, line 68: `unsupportedDialect` :)

sdepold (commented):
Terrific work!

sdepold:
Do I understand it correct that the main entry file for my dialect would call 

```
import {registerDialect} from 'sequelize'

registerDialect(MyDialectClass)
```

?

joaoe:
There is a bunch of ifdefery in tests, like https://github.com/jesse23/sequelize/blob/723418bd4705ddf03a4afd08fbd09be3ab23f49f/test/integration/sequelize/query.test.js#L237
Is this something you wish to change as part of this PR ?

lib/dialects/abstract/utils.js
jesse23: This is a static method which could not be inherit...use class method is a bit strange.

There may be an approach in TS that `class AbstractDialect implements InterfaceWithGetName`. That might work.

jesse23:
> https://github.com/jesse23/sequelize/blob/723418bd4705ddf03a4afd08fbd09be3ab23f49f/test/integration/sequelize/query.test.js#L237

Yes this is a good point and I considered it when I make this change. We can do it step by step for decoupling those.

jesse23:
> Do I understand it correct that the main entry file for my dialect would call
> 
> ```
> import {registerDialect} from 'sequelize'
> 
> registerDialect(MyDialectClass)
> ```
> 
> ?

My 1st approach is the dialect owner call it as u said.

But @ephys suggest we leave that responsibility to the user. So for user they might need to do it in the place they use that dialect:
```javascript
import myDialect from 'my-dialect';

//...
Sequelize.registerDialect(myDialect); 
//...
```

I am fine for either of them.

lib/dialects/abstract/utils.js
sdepold: Ah we are still in js here. My bad :)

lib/dialects/abstract/registry.js
```diff
@@ -0,0 +1,36 @@
+
+const _registries = {};
```
ephys, line 2:
```suggestion
const _registries = Object.create(null);
```

so getDialect('toString') returns null and avoids prototype injection

lib/dialects/abstract/registry.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Register dialect implementation.
+ *
+ * @param {AbstractDialect} [DialectClass] dialect implemetation as subclass of AbstractDialect.
+ */
+function registerDialect( DialectClass ) {
+  _registries[DialectClass.getDialectName()] = DialectClass;
```
ephys, line 10: We can throw if `(_registries[dialectName] != null && _registries[dialectName] !== DialectClass)` to ensure a user isn't accidentally registering the same dialect with two different implementations.

ephys (changes requested):

lib/dialects/abstract/utils.js
ephys: @jesse23 you called it registries instead of registry, it should be singular as it's a single registry :)

lib/dialects/abstract/registry.js
sdepold: Can we call it registry? It's a single registry after all :)

lib/dialects/abstract/registry.js
sdepold: > > Do I understand it correct that the main entry file for my dialect would call
> > ```
> > import {registerDialect} from 'sequelize'
> > 
> > registerDialect(MyDialectClass)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> My 1st approach is the dialect owner call it as u said.
> 
> But @ephys suggest we leave that responsibility to the user. So for user they might need to do it in the place they use that dialect:
> 
> ```js
> import Sequelize from 'sequelize';
> import myDialect from 'my-dialect';
> 
> //...
> Sequelize.registerDialect(myDialect); 
> //...
> ```
> 
> I am fine for either of them.

I think I would like to have something like this:

```
import Postgres from '@sequelize/postgres';
import Sequelize from '@sequelize/core';

const sequelize = new Sequelize(/* whatever */, { dialect: Postgres })
```

And now that I write it, I wonder what we need the registry for?!

ephys: Sequelize supports database URIs (`new Sequelize('mysql://localhost:3306/database')`), it would need to use the registry to select the corresponding Dialect class.

It also keeps backward compatibility with specifying the dialect as string (given this is targeting v6) (`new Sequelize(/* whatever */, { dialect: 'pg' })`)

But we should also support passing the Dialect class as a parameter, bypassing the registry. (`new Sequelize(/* whatever */, { dialect: Postgres })`)

sdepold: Oh is this supposed to work in v6? I thought it's one of the early things for v7?

ephys: I assumed it was v6 because of the backward compatibility but i'm not sure whether this was opened before the branch rename.

If this is targetting v7:

I'd vote to not break accepting URIs even but we could change the API to something that looks like this:

```typescript
new Sequelize(/* options */, { dialect: PostgresDialect });

new Sequelize('mysql://localhost:3306/database', {
   // array of dialects, one of them must support `mysql:`
  dialect: [MysqlDialect, PostgresDialect],
});
```

sdepold: But what is your use case to add more than one dialect? Do we currently support something like this where you connect with one sequelize instance to different databases?

sdepold: -->

```
new Sequelize(/* options */, { dialect: PostgresDialect });
new Sequelize('mysql://localhost:3306/database', {
   // array of dialects, one of them must support `mysql:`
  dialect: MysqlDialect,
});
```

ephys: I honestly do not have one. I went with an array to mimic the current behavior where the protocol defines the dialect. The only use-case I could see would be to have the ability to let the config define the protocol. It's not a good use case and it's achievable if the user wants it anyway.

In sumary, this sounds good to me:

```typescript
new Sequelize(database, username, password, { dialect: PostgresDialect });
new Sequelize('mysql://localhost:3306/database', { dialect: MysqlDialect });
```

sdepold: 💯 :)

sdepold:
Does that mean there is no need for the registration part to begin with?

lib/dialects/abstract/registry.js
joaoe: > new Sequelize(/* options */, { dialect: PostgresDialect });

That does not help much with the use case of specifying the dialect as an environment variable for tests. Unless ofc, the proper dialect is loaded in `test/supports.js`

ephys: we can pass the right dialect based on the environment variable :)

ephys:
@sdepold correct

ephys:
@jesse23 hey sorry for the last minute rework but have you seen this proposed simplified api change? https://github.com/sequelize/sequelize/pull/13798#discussion_r776396622

In summary, instead of a registry, you'd pass the dialect class to the Sequelize constructor:

```typescript
new Sequelize(database, username, password, { dialect: PostgresDialect });
new Sequelize('mysql://localhost:3306/database', { dialect: MysqlDialect });
```

which would mean no registry required at all, but dialects would have to be importable from `sequelize` (until they're moved to different packages)

jesse23:
> @jesse23 hey sorry for the last minute rework but have you seen this proposed simplified api change? [#13798 (comment)](https://github.com/sequelize/sequelize/pull/13798#discussion_r776396622)
> 
> In summary, instead of a registry, you'd pass the dialect class to the Sequelize constructor:
> 
> ```ts
> new Sequelize(database, username, password, { dialect: PostgresDialect });
> new Sequelize('mysql://localhost:3306/database', { dialect: MysqlDialect });
> ```
> 
> which would mean no registry required at all, but dialects would have to be importable from `sequelize` (until they're moved to different packages).
> 
> No need to worry about backward compatibility as this is targeting v7

That is not possible at this moment - we need to refactor the `DataTypes.js` to not rely on dialect name....will be working on that after this PR.

jesse23:
@ephys Please help me to review again when u have time...thx

ephys:
@jesse23 I think we can merge as part one yes, however one last (i hope) note:

If we agree that the end goal of this API is to look like this:

```typescript
new Sequelize(database, username, password, { dialect: PostgresDialect });
new Sequelize('mysql://localhost:3306/database', { dialect: MysqlDialect });
```

Then we should not expose `registerDialect` at the top level. I don't want to introduce a future breaking change.

sdepold:
What Zoé said :) 💯

jesse23:
> What Zoé said :) 💯

Yes we can do refactor and clean up in next PR as one piece in v7 ( since it is a breaking change)

jesse23:
Will close this and create a new one

PR #13799, opened 2021-12-21
refactor(data-types): convert data-types to TS

AllAwesome497:
### Pull Request Checklist

### Description Of Change

converts data-types to TS.

I've typed this in a way that the types of columns can be extracted from `sequelize.define` (in theroy - haven't gotten far enough to test that yet).  I also added some generic types to assist in getting types of a DataType.

For any reviewers, I've labled 3 (ts) types which are relevant to each DataType:
- "Acceptable" type (ie what it'll accept as user input, e.g. for dates this is numbers, Dates, or strings)
- "Sane" type - What sequelize will normally return (e.g. for dates, Date)
- "Raw" type - Whatever we need to deserealize for a type.  (e.g. for a boolean this is a number, buffer, or string)

image

### Child PRs

- #13970
- #13981
- #14081
- #14220
- #14238
- #14239
- #14240

### The approach of this PR

Due to the size of this change, I've opted to split up this PR into many to make it more reviewable.

Here's the steps I'll be taking:
1. Create a "master" pr containing the changes which other data-types prs will be based on (this pr)
2. For each dialect, create a PR based on this branch containing the needed changes.
3. Once this PR is approved, merge in the child prs as they are approved.
4. Once all of the dialect specific code is in (meaning: all  lib/ code is approved and data-types should be working), start converting the tests to TS (in distinct prs, again for reviewability).
5. Once all test prs are approved & merged, merge this pr (assuming tests are green)


WikiRik:
@AllAwesome497 checkout #12941 which was updated this week

AllAwesome497:
> @AllAwesome497 checkout #12941 which was updated this week

Oh I completely missed that.

My change covers a few things that aren't handled in that pr such as named exports, inline static keys (e.g. `key` is then declared in the class itself rather than being exposed later), and more specific typing for types which vary from their raw values (in particular: "sane" vs "accepted" vs "raw" as I've called them).

sdepold:
@jesse23 @Keimeno Up for a review?

sdepold:
I think my TS skills don't really suffice to review this beast. I have an opinion the wording however:

"Acceptable" type (ie what it'll accept as user input, e.g. for dates this is numbers, Dates, or strings)
--> InputType

"Sane" type - What sequelize will normally return (e.g. for dates, Date)
--> SanitizedType

"Raw" type - Whatever we need to deserealize for a type. (e.g. for a boolean this is a number, buffer, or string)
--> OutputType

sdepold:
Adding @ephys as well :)

jesse23:
@AllAwesome497 2 questions in general...

- It is possible enable the test 1stly, then try to do pair converting with the test?  For example in this case, no need to convert all impacted test but at least the major `test/unit/sql/data-types.test.js`.

- Shall we merge with `data-types.d.ts` ( and eliminate `data-types.d.ts` eventually )?

ephys (commented):
Looks good to me, this is my only note

lib/utils/class-to-invokable.ts
```diff
@@ -1,28 +1,23 @@
...
-): Invokeable<Args, Instance> {
-  return new Proxy<Invokeable<Args, Instance>>(Class as any, {
-    apply(_target, _thisArg, args: Args) {
-      return new Class(...args);
-    },
-    construct(_target, args: Args) {
-      return new Class(...args);
-    },
+export function classToInvokable<
+  Class extends new (...args: Array<never>) => any
```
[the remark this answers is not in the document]
AllAwesome497: This could also be `Array<any>` (and that's what I'd normally put).  I just did this since eslint recommended it and I was unsure if it'd work.  Will probably change it back to `Array<any>` since that's more intuitive IMO.

AllAwesome497:
> I think my TS skills don't really suffice to review this beast. I have an opinion the wording however:
> 
> "Acceptable" type (ie what it'll accept as user input, e.g. for dates this is numbers, Dates, or strings) --> InputType
> 
> "Sane" type - What sequelize will normally return (e.g. for dates, Date) --> SanitizedType
> 
> "Raw" type - Whatever we need to deserealize for a type. (e.g. for a boolean this is a number, buffer, or string) --> OutputType

Sounds good, though I think that `RawType` is a bit clearer than `OutputType` since its more of an "internal" output type,

`Boolean` is one of the types where this is more relevant, since a "raw" boolean can be a buffer, string, or number imo.

sdepold:
How are we doing on this front? I just wanted to point out one more time, that we are now targeting v7 and could hence drop support for any weird hack that was introduced earlier.

AllAwesome497:
> How are we doing on this front? I just wanted to point out one more time, that we are now targeting v7 and could hence drop support for any weird hack that was introduced earlier.

Have been pretty busy recently, back on it for now.  

Looks like we'll need to re-model how `DataTypes` works but I don't think it'll be too bad (ie most standard usage should be fine iiuc).

ephys:
I'm researching how we could support customising what type Sequelize must output for a given DataType. I'm waiting on this PR to be merged before properly experimenting but since you've knee-deep in DataTypes I'd like to know if my current draft would impact your decisions in this PR, like for `SaneTypeOf` https://gist.github.com/ephys/50a6a05be7322c64645be716fea6186a#support-for-alternative-js-types

Also I'd still like to bikeshed the names of the 3 types based on how I understood them:

- "Acceptable" -> UserInputTypeOf
- "Sane" SequelizeOutputTypeOf
- "Raw" -> SqlInputTypeOf

AllAwesome497:
> I'm researching how we could support customising what type Sequelize must output for a given DataType. I'm waiting on this PR to be merged before properly experimenting but since you've knee-deep in DataTypes I'd like to know if my current draft would impact your decisions in this PR, like for `SaneTypeOf` https://gist.github.com/ephys/50a6a05be7322c64645be716fea6186a#support-for-alternative-js-types
> 
> Also I'd still like to bikeshed the names of the 3 types based on how I understood them:
> 
>     * "Acceptable" -> UserInputTypeOf
> 
>     * "Sane" SequelizeOutputTypeOf
> 
>     * "Raw" -> SqlInputTypeOf

Nope, I think your current draft is fine - though it will be a pain to migrate typing to this.

As I'm finding, there's a lot of undocumented dialect-specific logic around typing :/. 

As for renaming:  I will do this, I just haven't yet since there hasn't been a clear choice in terms of names yet (and I don't see this PR getting merged soon, given its scale so I figure its worth waiting).

ephys:
> though it will be a pain to migrate typing to this.

Yup! Not looking forward to it but I want the feature enough to be motivated at least

AllAwesome497:
> * 1stly

:thinking: I thought I responded to this but I don't see any response.

> It is possible enable the test 1stly, then try to do pair converting with the test? For example in this case, no need to convert all impacted test but at least the major test/unit/sql/data-types.test.js.

I've opted to convert everything to TS first and add tests later (however: this will be before this PR is merged).  I added some more details to the PR description.

> Shall we merge with data-types.d.ts ( and eliminate data-types.d.ts eventually )?

Yep!  That should be done by the time I merge this if I haven't already removed the file.

AllAwesome497:
For renaming, thoughts on:

`AcceptedType` -> `InputType`
`SaneType` -> `OutputType`
`RawType` -> `ParsedType` 

CC @sdepold @ephys

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+   */
+  constructor(options: number | Readonly<NumberOptions> = {}) {
+    super();
+    if (typeof options === 'number') {
+      options = {
+        length: options,
+      };
+    }
+
+    this.options = options;
```
ephys, line 503: I think it would be a good opportunity to normalize `options`. We should either have `options`, or `_length` & co. But not both, otherwise we need to keep both in sync and it's error prone.

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+
+  /**
+   * @param {string|number} [precision] defines precision
+   * @param {string|number} [scale] defines scale
+   */
+  constructor(options?: DecimalOptions);
+  constructor(precision: number, scale?: number);
+  constructor(...args: DecimalConstructorParams);
+  constructor(precision?: number | DecimalOptions, scale?: number) {
+    if (typeof precision === 'object') {
```
ephys, line 756: use `isObject` from lodash instead

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+      // Sadly TS isn't smart enough to infer that this will be defined by the constructor of NUMBER.
+      // This should never do anything becuase this.options should already be defined.
+      this.options ??= {};
+
+      this.options.precision = precision;
+      this.options.scale = scale;
+    }
+
+    this._precision = this.options.precision;
+    this._scale = this.options.scale;
```
ephys, line 770: same comment about deduplicating `options` & instance fields

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+   */
+  constructor(options: NumberOptions);
+  constructor(length: number, decimals?: number);
+  constructor(...args: FloatConstructorParams);
+  constructor(length?: number | NumberOptions, decimals?: number) {
+    super(typeof length === 'object' ? length : { length, decimals });
+  }
+
+  public validate(value: AcceptedNumber): asserts value is AcceptedNumber;
+  public validate(value: AcceptedNumber): true {
```
ephys, line 667: can't this be simplified to:

```suggestion
  public validate(value: AcceptedNumber): asserts value is AcceptedNumber {
```

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+    if (this._precision || this._scale) {
+      return `DECIMAL(${[this._precision, this._scale]
+        .filter(num => num != null)
+        .join(',')})`;
+    }
+
+    return 'DECIMAL';
+  }
+
+  public validate(value: AcceptedNumber): asserts value is AcceptedNumber;
```
ephys, line 783: `AcceptedNumber` doesn't include string. But `DECIMAL` should accept strings to avoid loss of precision

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+    }
+
+    return true;
+  }
+}
+
+/**
+ * Decimal type, variable precision, take length as specified by user
+ */
+export type DECIMAL = _DECIMAL;
```
ephys, line 799: I know decorators are not standard yet but all this boilerplate would be much cleaner by using a `@newOptional` decorator.

This type of decorators is also possible with the updated decorator proposal so I think it's safe to use it internally, we'll just replace it when the time comes.

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+  public static readonly key = 'BOOLEAN';
+
+  public toSql() {
+    // Note: This may vary depending on the dialect.
+    return 'TINYINT(1)';
+  }
+
+  public validate(value: boolean | Falsy): asserts value is boolean;
+  public validate(value: boolean | Falsy): true {
+    if (!Validator.isBoolean(String(value))) {
```
ephys, line 819: value says it only accepts booleans (and falsy) but this tries to also accept things like `'true` and `'false'` and whatever object stringifies to these two strings.

I think we should restrict this one to only `true` and `false` and check using `typeof value === 'boolean'`

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+      );
+    }
+
+    return true;
+  }
+
+  protected _sanitize(
+    value: string | number | Buffer | boolean,
+  ): boolean | null {
+    if (value !== null && value !== undefined) {
```
ephys, line 832:
```suggestion
    if (value != null) {
```

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+  /**
+   * @param [length] precision to allow storing milliseconds
+   */
+  constructor(length?: number | DateOptions) {
+    super();
+    const options: DateOptions
+      = typeof length === 'object' ? length : { length };
+    this.options = options;
+    // TODO: Check if this is still used and remove if it isn't.
+    this._length = options.length || '';
```
ephys, line 903: same comment about deduplicating `options` & instance fields

'_length' is probably used for the precision parameter most dialects support.

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+export type DATEONLY = _DATEONLY;
+export const DATEONLY = classToInvokable(_DATEONLY);
+
+class _HSTORE extends ABSTRACT<Record<string, unknown>> {
+  public static readonly key = 'HSTORE';
+
+  public validate(value: Record<string, unknown>) {
+    const proto = Object.getPrototypeOf(value);
+
+    if (proto && proto !== Object.prototype) {
```
ephys, line 1036: use `isPlainObject` from lodash instead

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+  /**
+   * @param [length=''] could be tiny, medium, long.
+   */
+  constructor(length?: BlobLength | BlobOptions) {
+    super();
+    const options: BlobOptions
+      = typeof length === 'object' ? length : { length };
+
+    this.options = options;
+    this._length = options.length?.toLowerCase();
```
ephys, line 1120: same comment about deduplicating

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+  }
+
+  toSql() {
+    switch (this._length) {
+      case BlobLength.TINY:
+        return 'TINY_BLOB';
+      case BlobLength.MEDIUM:
+        return 'MEDIUM_BLOB';
+      case BlobLength.LONG:
+        return 'LONG_BLOB';
```
ephys, line 1130: I don't think any dialect has `_` in their blob name https://new.sequelizejs.com/docs/v7/other-topics/other-data-types#blobs

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+
+/**
+ * A default value of the current timestamp.  Not a valid type.
+ */
+export type NOW = _NOW;
+export const NOW = classToInvokable(_NOW);
+
+type AcceptedBlob = Buffer | string;
+
+enum BlobLength {
```
ephys, line 1095: BlobLength should be exported

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+
+    if (typeof options.subtype === 'function') {
+      options.subtype
+        = typeof options.subtype === 'function'
+          ? new options.subtype()
+          : options.subtype;
+    } else {
+      // @ts-expect-error This errors since INTEGER doesn't always apply to T, but we'll assume that if T is provided,
+      // the user will provvide the type themselves.
+      options.subtype ??= new INTEGER();
```
ephys, line 1200: we should avoid mutating options without making a copy first

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+ * A default unique universal identifier generated following the UUID v4 standard
+ */
+export type UUIDV4 = _UUIDV4;
+export const UUIDV4 = classToInvokable(_UUIDV4);
+
+class _VIRTUAL<T> extends ABSTRACT<T> {
+  public static readonly key = 'VIRTUAL';
+
+  returnType?: ABSTRACT<T>;
+  fields?: string[];
```
ephys, line 1326:
```suggestion
  returnType: ABSTRACT<T> | undefined;
  fields: string[] | undefined;
```

these fields are always created, but they can be undefiend

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+    } else {
+      options = {
+        // this rule is irrelevant - this is using `concat` for cases such as `new Union('a', ['b', 'c]])`
+        // eslint-disable-next-line unicorn/prefer-spread
+        values: ([] as Member[]).concat(...(args as Array<Member | Member[]>)),
+      };
+    }
+
+    this.values = options.values;
+    this.options = options;
```
ephys, line 1427: name note about deduplicating options

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+
+    const [arg0] = args;
+
+    if (typeof arg0 === 'object' && !Array.isArray(arg0)) {
+      options = arg0;
+    } else if (args.length === 1 && Array.isArray(arg0)) {
+      options = { values: arg0 };
+    } else {
+      options = {
+        // this rule is irrelevant - this is using `concat` for cases such as `new Union('a', ['b', 'c]])`
```
ephys, line 1420: `DataTypes.ENUM('a', ['b', 'c'])` should throw. We should only support

```
DataTypes.ENUM('value', 'another value')
DataTypes.ENUM(['value', 'another value'])
DataTypes.ENUM({
  values: ['value', 'another value']
})
```

to avoid future accidental breaking changes because a refactor broke something someone accidentally used

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+  public static readonly key = 'ENUM';
+  public values: Member[];
+  protected options: EnumOptions<Member>;
+
+  /**
+   * @param {...any|{ values: any[] }|any[]} args either array of values or options object with values array. It also supports variadic values
+   */
+  constructor(options: EnumOptions<Member>);
+  constructor(members: Member[]);
+  constructor(...members: Array<Member | Member[]>);
```
ephys, line 1406: `DataTypes.ENUM('a', ['b', 'c'])` shouldn't be accepted

```suggestion
  constructor(...members: Member[]);
```

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+  constructor(...args: EnumParams<Member>); // This is the overload TS will chose when using ConstructorParams and is needed.
```
ephys, line 1407: this line doesn't need to be specified, unless you want TypeScript to allow users to use this overload. But I don't think they should.

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+   */
+  constructor(type: DataType<T> | { type: DataType<T> }) {
+    super();
+    const opts = ABSTRACT.isType(type) ? { type } : type;
+
+    this.options = {
+      type: typeof opts.type === 'function' ? new opts.type() : opts.type,
+    };
+
+    this.type = this.options.type;
```
ephys, line 1486: same comment about deduplicating

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+   * @param {string} [type] Type of geometry data
+   * @param {string} [srid] SRID of type
+   */
+  constructor(type: Type, srid?: number);
+  constructor(options: GeometryOptions<Type>);
+  constructor(...args: GeometryParams<Type>);
+  constructor(type?: Type | GeometryOptions<Type>, srid?: number) {
+    super();
+    const options = typeof type === 'object' ? type : { type, srid };
+    this.options = options;
```
ephys, line 1549: same note about deduplicating

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+  protected type?: Type;
+  protected srid?: number;
+
+  /**
+   * @param {string} [type] Type of geometry data
+   * @param {string} [srid] SRID of type
+   */
+  constructor(type: Type, srid?: number);
+  constructor(options: GeometryOptions<Type>);
+  constructor(...args: GeometryParams<Type>);
```
ephys, line 1545: using a rest parameter here, when you don't actually accept rest params, is misleading. Plus this seems to be already covered by the previous 2 overloads

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+  DOUBLE,
+  GEOMETRY,
+  GEOGRAPHY,
+  CIDR,
+  INET,
+  MACADDR,
+  CITEXT,
+  TSVECTOR,
+};
+export type DataTypes = typeof DataTypes;
```
ephys, line 1826: is this line necessary?

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+class _ABSTRACT<
+  /** The type of value we'll accept - ie for a column of this type, we'll accept this value as user input. */
+  AcceptedType,
+  /** The "sane" type of this column - ie the type of a value of a field if it was sanitized via sequelize. */
+  SaneType extends AcceptedType = AcceptedType,
+  /** The type of value retrieved from. */
+  RawType = AcceptedType,
+> {
+  public static readonly key: string = 'ABSTRACT';
+  // @internal
```
ephys, line 75: doesn't `stripInternal` only apply to jsdoc? (either way this should be jsdoc)

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+
+const warnings: Record<string, boolean> = {};
+
+// If T is a constructor, returns the type of what `new T()` would return-
+// otherwise, returns T
+export type Constructed<T> = T extends abstract new () => infer Instance
+  ? Instance
+  : T;
+
+export type AcceptableTypeOf<T extends DataType> =
```
ephys, line 23: the purpose of these different types should be documented

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+import { classToInvokable } from '../../utils/class-to-invokable';
+import { joinSQLFragments } from '../../utils/join-sql-fragments';
+import { logger } from '../../utils/logger';
+import { validator as Validator } from '../../utils/validator-extras';
+
+const warnings: Record<string, boolean> = {};
+
+// If T is a constructor, returns the type of what `new T()` would return-
+// otherwise, returns T
+export type Constructed<T> = T extends abstract new () => infer Instance
```
ephys, line 19: isn't this the same as the built-in type `InstanceType`?

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+import wkx from 'wkx';
+import { kSetDialectNames } from '../../dialect-toolbox';
+import { ValidationError } from '../../errors';
+import type { Falsy } from '../../generic/falsy';
+import { classToInvokable } from '../../utils/class-to-invokable';
+import { joinSQLFragments } from '../../utils/join-sql-fragments';
+import { logger } from '../../utils/logger';
+import { validator as Validator } from '../../utils/validator-extras';
+
+const warnings: Record<string, boolean> = {};
```
ephys, line 15:
```suggestion
const warnings: Record<string, boolean> = Object.create(null);
```

src/data-types.ts
```diff
@@ -0,0 +1,9 @@
+export * from './dialects/abstract/data-types';
+export * as db2 from './dialects/db2/data-types';
```
ephys, line 2: what is the reason for exposing these? Shouldn't users only use `DataTypes.x` and not `DataTypes.postgres.x`?

ephys (changes requested):

ephys:
Some comments may be unrelated to this PR

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+    if (!Object.prototype.hasOwnProperty.call(this, 'types')) {
+      const prop = {
+        value: {},
+        writable: false,
+        enumerable: false,
+        configurable: false,
+      };
+
+      Reflect.defineProperty(this, 'types', prop);
+      Reflect.defineProperty(this.prototype, 'types', prop);
```
ephys, line 97: Is there a reason to add it to the prototype too?

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+        && (value === 0 || value === 1)
+      ) {
+        return Boolean(value);
+      }
+    }
+
+    return null;
+  }
+
+  static readonly parse = this.prototype._sanitize;
```
ephys, line 859: I'd prefer to have the `_sanitize` instance method call the static `parse` method instead of getting something from the prototype

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+      this.options.scale = scale;
+    }
+
+    this._precision = this.options.precision;
+    this._scale = this.options.scale;
+  }
+
+  public toSql() {
+    if (this._precision || this._scale) {
+      return `DECIMAL(${[this._precision, this._scale]
```
ephys, line 775: if scale is specified but precision is not, we should throw with a clear error message

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+  toSql() {
+    switch (this.options.length) {
+      case 'tiny':
+        return 'TINY_TEXT';
+      case 'medium':
+        return 'MEDIUM_TEXT';
+      case 'long':
+        return 'LONG_TEXT';
+      default:
+        return this.key;
```
ephys, line 415: Why were these changed, they should be `TINYTEXT` & co

src/dialects/abstract/data-types.ts
```diff
@@ -0,0 +1,1826 @@
...
+  validate(value: string): asserts value is string;
+  validate(value: string): true {
+    if (typeof value !== 'string') {
+      throw new ValidationError(
+        util.format('%j is not a valid string', value),
+        [],
+      );
+    }
+
+    return true;
```
ephys, line 428: We should remove these `return true` from validate.

ephys (changes requested):

AllAwesome497:
Closing in favor of https://github.com/sequelize/sequelize/pull/14505.

PR #13805, opened 2021-12-21
Fix TS support for operators.ts

### LINKED ISSUE #13791: Typescript 4.5.4 does not like sequelize-ESM imports in v6.12

ephys:
### Pull Request Checklist

### Description Of Change

Closes #13791

Fix turned out more complicated than previously thought (https://github.com/sequelize/sequelize/pull/13797)

Current solution is definitely a hack to please both TypeScript and Node.

I think the TypeScript migration is going to keep breaking things due to how `esbuild` transpiles ESM. I've opened an issue regarding this as the decision will impact this PR https://github.com/sequelize/sequelize/issues/13806

### DIFF
--- a/lib/operators.ts
+++ b/lib/operators.ts
@@ -521,4 +521,9 @@ const Op: OpTypes = {
   match: Symbol.for('match')
 } as OpTypes;
 
-export = Op;
+export default Op;
+
+// https://github.com/sequelize/sequelize/issues/13791
+// remove me in v7: kept for backward compatibility as `export default Op` is
+// transpiled to `module.exports.default` instead of `module.exports`
+module.exports = Op;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,6 +1,6 @@
 import DataTypes = require("./lib/data-types");
 import Deferrable = require("./lib/deferrable");
-import Op = require("../lib/operators");
+import Op from "../lib/operators";
 import QueryTypes = require("./lib/query-types");
 import TableHints = require("./lib/table-hints");
 import IndexHints = require("./lib/index-hints");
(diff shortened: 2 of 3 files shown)


WikiRik:
Yeah, I'm starting to think the TypeScript migration might be better for v7. There are already various things that we've found from the conversion that we want to change and it gives a better overview to do them directly then add comments in the code that we need to refactor later. But let's discuss that at the next meeting in January; can you add things you've noticed about the migration to https://github.com/sequelize/meetings/issues/6 ?

ephys:
Oh thanks! I was looking for a place to put such a discussion and didn't think to use the meetings repo. Will do :)

WikiRik:
@sdepold time for a 6.12.2?

PR #13807, opened 2021-12-21
fix(13804): use schema for ARRAY(ENUM) type name

### LINKED ISSUE #13804: Enums data types does not honor schema names in postgres
## Issue Creation Checklist

## Bug Description

For postgres array of enum types.

When sequelize generates the enum name for bulk create queries for an ARRAY<ENUM> datat type is not considering that enums can be inside a schema, and tries to use the full name of the object (schema).(table) and adds enum_ at the begining.

This is not the logic used for creating the initial value of the enum and the final enum name is invalid. If there is a schema present it should only add enum_ to the table name and use the schema for reference at the begining of the type.

### SSCCE

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/204

I want that when using a model with a property that is an array of enums on the table in a postgres database, custom types in array are casted to the proper type when doing queries (in particular bulkCreate).

For a simple model like: (look that the schema is not the default and is redefined to: 'test')
```js
  BookDetails.init({
    uniqueName: {
        type: Sequelize.DataTypes.STRING(100),
        unique: true,
        allowNull: false,
        primaryKey: true
    },
    originalCategories: {
        type: Sequelize.DataTypes.ARRAY(Sequelize.DataTypes.ENUM({
          values: ['drama', 'comedy'],
        })),
      },
  }, {
      underscored: true,
      modelName: 'BookDetails',
      schema: 'test',
      sequelize
  });
```
when executing the bulkCreate I'm expecting that the data is inserted, but instead fails
```js
  await BookDetails.bulkCreate([{
      uniqueName: 'test',
      originalCategories: ['drama']
  }]);
```
the error output is clear:
```sql
INSERT INTO "test"."book_details" ("unique_name","original_categories") VALUES ('test',ARRAY['drama']::"enum_test.book_details_original_categories"[]) RETURNING "unique_name","original_categories";`,
    parameters: undefined
  },
  original: error: type "enum_test.book_details_original_categories[]" does not exist
enum_test.book_details_original_categories is not the proper type in the database, it should be test.enum_book_details_original_categories respecting the underscoring.
```

`enum_test.book_details_original_categories` is not the proper type in the database, it should be `test.enum_book_details_original_categories` respecting the schema name.

### Environment

- Sequelize version: 6.12.1 
- Node.js version: 14.18.2 
- Operating System: Ubuntu 20.04

nahog:
### Pull Request Checklist

- [x] Does the description below contain a link to an existing issue (Closes  #[issue]) or a description of the issue you are solving?

### Description Of Change

Closes #13804

This change is also realated with #13204 PR #13210. It not caused by it but the source of both bugs is the same, improper handling of ARRAY[ENUM] on postgres. In this scenario is only when we are using schemas.

### DIFF
--- a/test/integration/dialects/postgres/dao.test.js
+++ b/test/integration/dialects/postgres/dao.test.js
@@ -466,8 +466,9 @@ if (dialect.match(/^postgres/)) {
         });
 
         it('should be able to insert a new record even with an array of enums in a schema', async function() {
+          const schema = 'special_schema';
+          this.sequelize.createSchema(schema);
           const User = this.sequelize.define('UserEnums', {
-            schema: 'specialSchema',
             name: DataTypes.STRING,
             type: DataTypes.ENUM('A', 'B', 'C'),
             owners: DataTypes.ARRAY(DataTypes.STRING),
@@ -480,6 +481,8 @@ if (dialect.match(/^postgres/)) {
               ])),
               field: 'special_permissions'
             }
+          }, {
+            schema
           });
 
           await User.sync({ force: true });
--- a/test/integration/dialects/postgres/dao.test.js
+++ b/test/integration/dialects/postgres/dao.test.js
@@ -465,6 +465,35 @@ if (dialect.match(/^postgres/)) {
           expect(user.length).to.equal(1);
         });
 
+        it('should be able to insert a new record even with an array of enums in a schema', async function() {
+          const User = this.sequelize.define('UserEnums', {
+            schema: 'specialSchema',
+            name: DataTypes.STRING,
+            type: DataTypes.ENUM('A', 'B', 'C'),
+            owners: DataTypes.ARRAY(DataTypes.STRING),
+            specialPermissions: {
+              type: DataTypes.ARRAY(DataTypes.ENUM([
+                'access',
+                'write',
+                'check',
+                'delete'
+              ])),
+              field: 'special_permissions'
+            }
+          });
+
+          await User.sync({ force: true });
+
+          const user = await User.bulkCreate([{
+            name: 'file.exe',
+            type: 'C',
+            owners: ['userA', 'userB'],
+            specialPermissions: ['access', 'write']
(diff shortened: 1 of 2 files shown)


test/integration/dialects/postgres/dao.test.js
```diff
@@ -465,6 +465,35 @@ if (dialect.match(/^postgres/)) {
           expect(user.length).to.equal(1);
         });
 
+        it('should be able to insert a new record even with an array of enums in a schema', async function() {
+          const User = this.sequelize.define('UserEnums', {
+            schema: 'specialSchema',
```
WikiRik, line 470: schema is part of the options right? So it needs to be in the second array. And possibly the schema needs to be defined before as well, but not sure about that.

WikiRik (changes requested):
I believe your new test is not correct, can you check?

test/integration/dialects/postgres/dao.test.js
nahog: Yes, you are right, and the schema is needed, I already have it on my local db, but I will add it on the test.

nahog: The updated test is now passing. A [Flaky] test for mysql8/node16 failed but this fix is postgres specific, I don't think is realted, I checked the test and is using enums or anything related from what I see.

WikiRik: Yeah, the flaky means that it fails sometimes for no real reason. I will rerun the checks for you

nahog: I saw I had a typo on my last commit, fix(31804) instead of fix(13804), if the merges are squash and merge, is probably fine, if not, let me know and I can ammed the commit and force push to this branch with the new message

WikiRik (approved):

sdepold (approved):

sdepold:
thx!

PR #13818, opened 2021-12-23
fix(data-types): moment object throwing error

### LINKED ISSUE #13816: Sequelize v6.12.0+ is throwing an Error when using a moment.js object in date fields
## Bug Description

Sequelize is throwing an `Invalid value` error when using a `moment.js` object in `DATETIME` fields to do date comparisons. Additionally, when using a Moment object to set a date, the date value is set to `0000-00-00 00:00:00`

This is happening since v6.12.0.

```js
  User.findAll({
    where: {
      birthday: {
       [Op.gte]: moment()
      }
   }
  });
```

Moment objects should be accepted for date comparison, creation, etc. as usual.

An `Invalid value` error is thrown, crashing the app.

```
Invalid value Moment<2021-12-21T19:52:23+00:00>
```

### Additional context

* Using `mysql` as dialect
* Could be related to #13712 

### Environment

- Sequelize version: 6.12.0
- Node.js version: 16.13.0

fncolon:
### Pull Request Checklist

### Description Of Change

Fix #13816 , wrong if else and replace date checking with isMoment.

### Todos

- [x] fix(data-types): moment object throwing error


WikiRik:
Can we add regression tests for this? So one inputting a JS Date and one with a Moment object to check that this works without errors?

sdepold:
I think we should have a test that verifies that you can pass a moment instance.

fncolon:
Hello guys already pinged you about the regressed tests in slack, kindly checked it ;)

sdepold:
I have added a test case but it is bailing since the code change does not seem to fix the problem

WikiRik:
@sdepold might be good to revert https://github.com/sequelize/sequelize/pull/13712 in the v6 branch for now since it was just a minor change to get rid of a deprecation warning (and with moment being deprecated itself not to worry that it will be removed). This second change looks too big to me and with timezone things you always need to be careful

And then later we can see if @fncolon can find a better fix that works with different types of Date-related objects

sdepold:
I agree

fncolon:
> I have added a test case but it is blowing up since the code change does not seem to fix the problem

I tried to find the blowing up parts, can you give me the specific result?

fncolon:
> @sdepold might be good to revert #13712 in the v6 branch for now since it was just a minor change to get rid of a deprecation warning (and with moment being deprecated itself not to worry that it will be removed). This second change looks too big to me and with timezone things you always need to be careful
> 
> And then later we can see if @fncolon can find a better fix that works with different types of Date-related objects

Agree! 😃

sdepold:
Actually I'm confused now myself. Let me have a look.

sdepold:
Local:

```  1) [MYSQL] Model
       findAll
         special where conditions/smartWhere object
           should be able to find a row using greater than or equal to logic with moment dates:
     Error: Invalid value Moment<2013-01-09T00:00:00+01:00>
      at Object.escape (lib/sql-string.js:65:11)
      at MySQLQueryGenerator.escape (lib/dialects/abstract/query-generator.js:1034:22)
      at MySQLQueryGenerator._whereParseSingleValueObject (lib/dialects/abstract/query-generator.js:2679:41)
      at MySQLQueryGenerator.whereItemQuery (lib/dialects/abstract/query-generator.js:2389:21)
      at /Users/sdepold/Projects/sequelize/lib/dialects/abstract/query-generator.js:2290:25
      at Array.forEach (<anonymous>)
      at MySQLQueryGenerator.whereItemsQuery (lib/dialects/abstract/query-generator.js:2288:35)
      at MySQLQueryGenerator.getWhereConditions (lib/dialects/abstract/query-generator.js:2706:19)
      at MySQLQueryGenerator.selectQuery (lib/dialects/abstract/query-generator.js:1365:28)
      at MySQLQueryInterface.select (lib/dialects/abstract/query-interface.js:979:27)
      at Function.findAll (lib/model.js:1764:47)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at Context.<anonymous> (test/integration/model/findAll.test.js:391:23)```

sdepold:
oh I forgot to build the project 👌

sdepold:
Alright. So, it works just fine locally (as in the tests are green). Should we merge the change or rather revert to the old solution. I'm fine with both really. Since this might be one of the last versions of sequelize 6 it might be worth fixing the warning. However, if we feel it's too unsafe, reverting is also ok.

ephys:
So I noticed this change when reviewing https://github.com/sequelize/sequelize/pull/14400, and this PR + https://github.com/sequelize/sequelize/pull/13712 sound very strange to me. Why would we only apply the `timezone` option only to non-moment things?

Isn't that a breaking change? Before this change, the [`timezone`](https://sequelize.org/api/v7/interfaces/options#timezone) option was applied to all dates indiscriminately of which library created the date.

With time zones it's very likely that this breaking change would have gone unnoticed.

fzn0x:
> So I noticed this change when reviewing #14400, and this PR + #13712 sound very strange to me. Why would we only apply the `timezone` option only to non-moment things?
> 
> Isn't that a breaking change? Before this change, the [`timezone`](https://sequelize.org/api/v7/interfaces/options#timezone) option was applied to all dates indiscriminately of which library created the date.
> 
> With time zones it's very likely that this breaking change would have gone unnoticed.

It is actually throwing error before this changes for some cases 

What I did was to follow the default of how we handle timezone and date and then I found out we can't applied to all dates if the filter was using momentjs, so that's why this error did happen. Before this change, the [timezone](https://sequelize.org/api/v7/interfaces/options#timezone) option was applied to all dates indiscriminately of which library created the date (https://github.com/sequelize/sequelize/issues/13816).

fzn0x:
But I agree with you, this should be breaking change also we didn't applied all dates indiscrimantely before too, so we need much further changes to fix this issue #13819 without breaking changes, or accept breaking change (since no one complain about if there's any usage turns out to be breaking)

ephys:
> It is actually throwing error before this changes for some cases

Because of https://github.com/sequelize/sequelize/pull/13712, or before that PR too? Because that PR introduced the breaking change too.

> since no one complain about if

It's only been a couple of months since this was published and a subtle change like this would not cause an error to be thrown. You'd have to look at the data to realize something changed.

fzn0x:
> > It is actually throwing error before this changes for some cases
> 
> Because of #13712, or before that PR too? Because that PR introduced the breaking change too.
> 
> > since no one complain about if
> 
> It's only been a couple of months since this was published and a subtle change like this would not cause an error to be thrown. You'd have to look at the data to realize something changed.
> 
> Either way, I think both of these PRs should be reverted. If all they did was fixed the warning from moment, we should look into fixing the warning in a non-breaking way.

before that PR too

ephys:
Do you have more info about the error that was being thrown prior to https://github.com/sequelize/sequelize/pull/13712?

fzn0x:
> Do you have more info about the error that was being thrown prior to https://github.com/sequelize/sequelize/pull/13712?

I don't have any but that's actually not correctly way of applyTimezone and one of the issue report is the one above

ephys:
This one https://github.com/sequelize/sequelize/issues/13816?

But that issue *was* caused by https://github.com/sequelize/sequelize/pull/13712, so it's not prior to it.

> but that's actually not correctly way of applyTimezone 

Why?

fzn0x:
> This one https://github.com/sequelize/sequelize/issues/13816?
> 
> But that issue *was* caused by https://github.com/sequelize/sequelize/pull/13712, so it's not prior to it.
> 
> > but that's actually not correctly way of applyTimezone 
> 
> Why?

applyTimezone throws warning when the input is not supported by moment.

fzn0x:
> This one https://github.com/sequelize/sequelize/issues/13816?
> 
> But that issue *was* caused by https://github.com/sequelize/sequelize/pull/13712, so it's not prior to it.
> 
> > but that's actually not correctly way of applyTimezone 
> 
> Why?

The reason is we need to use proper timezone on applyTimezone

ephys:
> The reason is we need to use proper timezone on applyTimezone

I'm sorry I don't understand what you mean, what was incorrect with `applyTimezone` before?

> applyTimezone throws warning when the input is not supported by moment.

A warning, not an error. So it must be resolved in a backward compatible way.

The current fix that skips applyTimezone if it's already a moment object is not backward compatible as we end up with different values being stored:

image

Do you remember in which scenarios moment logged that warning?

fzn0x:
> > The reason is we need to use proper timezone on applyTimezone
> 
> I'm sorry I don't understand what you mean, what was incorrect with `applyTimezone` before?
> 
> > applyTimezone throws warning when the input is not supported by moment.
> 
> A warning, not an error. So it must be resolved in a backward compatible way.
> 
> The current fix that skips applyTimezone if it's already a moment object is not backward compatible as we end up with different values being stored:
> 
> image
> 
> Do you remember in which scenarios moment logged that warning?

It's incorrect because we use applyTimezone without filtering if it was moment compatible date or not so it will throwing a warning that might we didn't want to see and don't know how to hide it, but unfortunately like you said it is introducing breaking change (but I didn't see anyone notice if it was a breaking change).

ephys:
> It's incorrect because we use applyTimezone without filtering if it was moment compatible date or not so it will throwing a warning that might we didn't want to see and don't know how to hide it

As I said, the warning should be fixed, but in a backward compatible way.
Which is why I asked if you remembered in which scenarios the warning was logged. I can't manage to replicate it by just passing a moment object to the old function.

> I didn't see anyone notice if it was a breaking change

As I said earlier, it doesn't mean it did not introduce bugs that have gone unnoticed, which again is especially easy when it comes to time zones. Or that it won't introduce a bug for someone that has not updated to Sequelize > 6.12 (even if they already were on Sequelize 6), so this is really not a valid point.

fzn0x:
I actually didn't input anything when the warning was thrown, it was thrown when I was getting data with the DATE type attribute.

fzn0x:
This is the workaround 
```js
  const crowdFundings = await CrowdFunding.findAndCountAll({
    distinct: true, // don't count the include part https://github.com/sequelize/sequelize/pull/4016#issuecomment-116294996
    attributes: [
     'date_attributes' // <- will cause the warning from momentjs
    ],
    include: [{ all: true }],
    where: {
      ...data,
      [Op.or]: searchUtil.search(crowdFundingAttributes, query.q || ""),
    },
    ...pagination.getPaginationOptionsData(),
  });
 ```

ephys:
Thanks, I'll see if I can reproduce the error!

WikiRik:
@ephys did you reproduce it? Do we need to revert it in v6 or shall we just keep it this way until we get complaints?

ephys:
I haven't taken the time to reproduce the initial warning yet, but I did test the behavior before & after this change and it does produce different datetime strings.

It produces the same UTC time, but using different UTC offsets, e.g.
`2022-04-18 23:05:05.617 +09:00` and `2022-04-18 16:05:05.617 +02:00`

This may be an issue in itself for some applications. It requires the user to be using both moment and the timeZone option in the sequelize constructor, which is probably going to be a small percentage of users.

fzn0x:
> I haven't taken the time to reproduce the initial warning yet, but I did test the behavior before & after this change and it does produce different datetime strings.
> 
> It produces the same UTC time, but using different UTC offsets, e.g. `2022-04-18 23:05:05.617 +09:00` and `2022-04-18 16:05:05.617 +02:00`
> 
> This may be an issue in itself for some applications. It requires the user to be using both moment and the timeZone option in the sequelize constructor, which is probably going to be a small percentage of users.

Thanks for the research

PR #13824, opened 2021-12-24
docs(indexes): match type definitions

benb116:
Change docs describing index creation for models to match existing type definitions. Fields passed as objects should have a "name" property, not an "attribute."

### Pull Request Checklist

### Description Of Change

In the documentation of indexes, an index can be added by passing an object with the property `fields`, which itself is an array of either strings or objects. According to the type definition for IndexOptions [/types/lib/query-interface.d.ts](../blob/main/types/lib/query-interface.d.ts#L187), objects in the `fields` array must have a `name` property.

However, the [docs file for indexes](../blob/main/docs/manual/other-topics/indexes.md?plain=1#L38) and the [lib file for model initialization](https://sequelize.org/master/class/lib/model.js~Model.html#static-method-init) both specify that the property should be called `attribute.` It seems like sequelize will copy the `attribute` property to the `name` property [here](../blob/main/lib/dialects/abstract/query-generator.js#L544), so this may not cause functional issues, but changing the documentation will help prevent TypeScript compilation errors for valid code.

I changed both the markdown file describing indexes and the parameter documentation for model initialization to use `name` instead of `attribute`.

This is my first ever PR, so please let me know if I should change anything. Thanks!


WikiRik (approved):
Seems indeed that name is the preferred term here. Thanks for the PR!

sdepold:
just fyi: there is also utils.js:

```function nameIndex(index, tableName) {
  if (tableName.tableName) tableName = tableName.tableName;

  if (!Object.prototype.hasOwnProperty.call(index, 'name')) {
    const fields = index.fields.map(
      field => typeof field === 'string' ? field : field.name || field.attribute
    );
    index.name = underscore(`${tableName}_${fields.join('_')}`);
  }

  return index;
}
```

sdepold:
@WikiRik feel free to merge :)

sdepold (approved):

WikiRik:
Hi @benb116 
Thanks for the PR! If I'm correct it will show up in the docs tomorrow

PR #13010, opened 2021-02-06
fix(types): models with attributes cannot be used in some functions

KapitanOczywisty:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] ~Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?~
- [x] Did you update the typescript typings accordingly (if applicable)?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)?

### Description of change

Models with attributes aren't accepted in some functions, this PR fixes these cases. Example:

```ts
interface UserAttributes {
    id: number;
    name: string;
}

class User extends Model<UserAttributes> implements UserAttributes {
    id!: number;
    name!: string;
}

interface ProjectAttributes {
    id: number;
    name: string;
}

interface UserProjectAttributes {
    id: number;
    customField: string;
}

class Project extends Model<ProjectAttributes> implements ProjectAttributes {
    id!: number;
    name!: string;
}

class UserProject extends Model<UserProjectAttributes> implements UserProjectAttributes {
    id!: number;
    customField!: string;
}

User.belongsToMany(Project, { through: { model: UserProject } });
Project.belongsToMany(User, { through: { model: UserProject } });

User.findAll({ include: [Project] });
```
image

Fixes https://github.com/sequelize/sequelize/issues/12842 https://github.com/RobinBuschmann/sequelize-typescript/issues/882

Related PR: https://github.com/RobinBuschmann/sequelize-typescript/pull/900 Original change: https://github.com/sequelize/sequelize/pull/12405

PR #13834, opened 2021-12-27
fix(types): allow use of `sequelize.define` models in types

hsource:
### Pull Request Checklist

### Motivation

#13010 had fixed an issue where class-based Models with explicit types couldn't be used in includes and other places that expect static models, but it had inadvertently broken the same for Models created with `sequelize.define()`, which were the standard before it.

This is because it created a new `ModelType` type, that required explicit properties. `sequelize.define` outputs a `ModelCtor` type that doesn't conform to that:

**Code with error:**

```ts
  const commonQuestionOptions: FindOptions<QuestionAttributes> = {
    include: [
      { model: Answer },
    ],
  };
```

**Error:** TS2322: Type 'ModelCtor<AnswerInstance>' is not assignable to type 'ModelType<any, any> | undefined'.

### Fix

I wrote the original Typescript typings for v6 (#12405), but didn't document it really well. This change better documents how the types should be used and swaps out most uses of `ModelType` with `ModelStatic`, which is compatible with both class and `sequelize.define` Models.

Here's what I did:

1. Document when to use `ModelCtor` and `ModelStatic` in the typings file
2. Switch all uses of `ModelType` to use `ModelStatic`, which works with both class and `sequelize.define` models
3. Add Typescript tests to ensure that using both class and function components type-check in `User.ts`

### Todos

None


ephys (approved):
Looks good to me, thank you for the extra jsdoc too

types/lib/model.d.ts
```diff
@@ -2911,15 +2911,50 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
...
 
-// Do not switch the order of `typeof Model` and `{ new(): M }`. For
-// instances created by `sequelize.define` to typecheck well, `typeof Model`
-// must come first for unknown reasons.
+/**
+ * The output of `sequelize.define` and other functions that output a model.
+ *
+ * Do not switch the order of `typeof Model` and `{ new(): M }`. For
+ * instances created by `sequelize.define` to typecheck well, `typeof Model`
+ * must come first for unknown reasons.
```
AllAwesome497, line 2933: This is probably because `typeof Model` is a constructor and as such, its `new` function overrides the constructor of `{ new(): M }`.

ephys:
Please don't merge this yet. This issue is related to a problem I've been trying to solve for the past year (ModelCtor not acting just right) and I think I've *just* solved it by changing `ModelCtor` to this definition

```typescript
export type ModelCtor<M extends Model> = Omit<typeof Model, 'new'> & { new(): M };
```

Checking if tests pass, brb

ephys:
Tests have passed on https://github.com/sequelize/sequelize/pull/13890

@hsource would you mind taking a look at it too?

hsource:
@ephys and @AllAwesome the fix in #13890 looks better. Ship it!

WikiRik:
I guess this one can be closed then, right?

ephys:
Thanks @hsource :) I'll close this one then

PR #13836, opened 2021-12-28
meta(deps): update (dev)deps

WikiRik:
Since renovate does not seem to work yet. For dependencies with versions 1 or above I updated minor and patch versions, for dependencies still on 0.x I updated only patch versions.

Can be merged to v6 as well if needed


package.json
```diff
@@ -76,26 +76,26 @@
     "copyfiles": "^2.4.1",
     "cross-env": "^7.0.3",
     "delay": "^5.0.0",
-    "esbuild": "^0.14.1",
+    "esbuild": "^0.14.8",
```
sdepold, line 79: I think we want to be careful about this one because of https://github.com/sequelize/sequelize/issues/13791#issuecomment-1001246172

Please hardcode 0.14.1 instead (without the ^)

WikiRik: Yeah, I was using this PR to check if the tests would fail. Also cc @ephys in this

sdepold (commented):
See above

package.json
sdepold: 👍

sdepold (changes requested):
See above

sdepold:
I have somehow royally screwed up the comments now :D

WikiRik:
Unit tests are indeed failing on esbuild 0.14.8, will try various versions locally to pin-point what esbuild changes caused this so we can fix this here/in a future PR

WikiRik:
@ephys changes from esbuild 0.14.3 - 0.14.4 caused the issue (but use 0.14.5 because of a wrong published version); https://github.com/evanw/esbuild/releases
Can you check it out and make a new PR to fix it? Should not be difficult I think but I can't really wrap my head around this stuff yet

sdepold (approved):

ephys:
(on phone): we can't update esbuild for V6 because they fix the workaround used in operators.ts

If this is targeting V7, we can resolve the workaround and update esbuild :)

PR #13837, opened 2021-12-28
ci(release): dispatch custom event to trigger notifications

sdepold:
### Pull Request Checklist

### Description Of Change

The notification flow can be triggered by sending a dispatch event to API including the release's ID. This change will automatically send such and event at the end of the release process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -273,3 +273,9 @@ jobs:
           node-version: 16.x
       - run: yarn install --frozen-lockfile
       - run: npx semantic-release
+      - id: sequelize
+        uses: sdepold/github-action-get-latest-release@master
+        with:
+          repository: sequelize/sequelize
+      - run: |
+          curl -XPOST -u "sdepold:${{ secrets.GH_TOKEN }}" -H "Accept: application/vnd.github.everest-preview+json" -H "Content-Type: application/json" https://api.github.com/repos/sequelize/sequelize/dispatches --data '{"event_type":"Release notifier","client_payload":{"release-id": ${{ steps.sequelize.outputs.id }}}}'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -278,4 +278,4 @@ jobs:
         with:
           repository: sequelize/sequelize
       - run: |
-          curl -XPOST -u "sdepold:${{ secrets.GH_TOKEN }}" -H "Accept: application/vnd.github.everest-preview+json" -H "Content-Type: application/json" https://api.github.com/repos/sequelize/sequelize/dispatches --data '{"event_type":"Release notifier","client_payload":{"release-id": ${{ steps.sequelize.outputs.id }}}}'
+          curl -XPOST -u "sdepold:${{ secrets.GH_TOKEN }}" -H "Accept: application/vnd.github.v3+json" -H "Content-Type: application/json" https://api.github.com/repos/sequelize/sequelize/dispatches --data '{"event_type":"Release notifier","client_payload":{"release-id": ${{ steps.sequelize.outputs.id }}}}'


.github/workflows/ci.yml
```diff
@@ -273,3 +273,9 @@ jobs:
           node-version: 16.x
       - run: yarn install --frozen-lockfile
       - run: npx semantic-release
+      - id: sequelize
+        uses: sdepold/github-action-get-latest-release@master
```
WikiRik, line 277: Let's update it to https://github.com/pozetroninc/github-action-get-latest-release once your PR is merged

.github/workflows/ci.yml
```diff
@@ -273,3 +273,9 @@ jobs:
...
+        with:
+          repository: sequelize/sequelize
+      - run: |
+          curl -XPOST -u "sdepold:${{ secrets.GH_TOKEN }}" -H "Accept: application/vnd.github.everest-preview+json" -H "Content-Type: application/json" https://api.github.com/repos/sequelize/sequelize/dispatches --data '{"event_type":"Release notifier","client_payload":{"release-id": ${{ steps.sequelize.outputs.id }}}}'
```
WikiRik, line 281: Why do you want the release ID? Isn't the tag enough?

WikiRik, line 281:
```suggestion
          curl -XPOST -u "sdepold:${{ secrets.GH_TOKEN }}" -H "Accept: application/vnd.github.v3+json" -H "Content-Type: application/json" https://api.github.com/repos/sequelize/sequelize/dispatches --data '{"event_type":"Release notifier","client_payload":{"release-id": ${{ steps.sequelize.outputs.id }}}}'
```

As per https://docs.github.com/en/rest/reference/repos#create-a-repository-dispatch-event--parameters

WikiRik (commented):
I have some questions about this, but I also think I'm missing some context. Why would we need this release notifier?

sdepold:
> I have some questions about this, but I also think I'm missing some context. Why would we need this release notifier?

This PR will cause releases to trigger the notify.yml GitHub workflow which is announcing new releases on Twitter and Open Collective.

sdepold:
https://opencollective.com/sequelize/updates

sdepold:
https://twitter.com/SequelizeJS/status/1475798293166833666

.github/workflows/ci.yml
sdepold: The notification script (https://github.com/sequelize/sequelize/blob/main/.github/workflows/notify.yml) will read the necessary meta information based on the release id.

.github/workflows/ci.yml
sdepold: If you don't mind, I would rather not wait since GitHub Actions are essentially just repositories that are cloned during runtime. There is no real point in waiting for the PR to be merged

sdepold:
To explain in more depths, where this PR is coming from:

Originally, we were posting releases on Twitter in the ci.yml after the release stage finished. 
I wanted to separate the notifications from the CI flow, since it feels kinda mixed concerns.

Hence there is now a notify.yml flow. Yesterday, I tried to have that flow to react to releases which were made by our semantic-release script. However, it turned out that one flow cannot possibly trigger another flow unless you make use of the dispatch functionality --> https://github.community/t/github-action-trigger-on-release-not-working-if-releases-was-created-by-automation/16559/6

So, that new notify.yml is now reacting on:

```
name: Notify release channels
on: repository_dispatch
```

In addition to posting on twitter (example: https://twitter.com/SequelizeJS/status/1475798293166833666), the new notify.yml will also call a new GitHub action (https://github.com/sequelize/proxy-release-to-open-collective) that will take the release and post a corresponding update on Open Collective (example: https://opencollective.com/sequelize/updates/release-of-sequelize-v7-0-0-alpha-3).

The somewhat nice part about the dispatch approach is the fact, that you can trigger the flow by sending a certain curl request. See here: https://github.com/sequelize/sequelize/blob/main/.github/workflows/notify.yml#L1-L5

sdepold:
btw. if that is working out find, we could change the docs generation to trigger only after a release.

.github/workflows/ci.yml
WikiRik: Agree, merge this now but in a future PR put it back to the original when your PR is merged by pozetroninc

WikiRik (approved):

WikiRik:
> btw. if that is working out find, we could change the docs generation to trigger only after a release.

That would be nice

PR #13839, opened 2021-12-29
ci(docs): trigger new website build through dispatch event

sdepold:
Upon a successful release, we will send a dispatch event to the sequelize.org repo which will pick it up and build a new version of the docs.

This is a sample build that was triggered using the same curl command as in this PR https://github.com/sequelize/sequelize.org/actions/runs/1633977126


WikiRik (approved):
Guess we need to change something on the website repo as well right?

EDIT: Nevermind, I see you've done that already

sdepold:
> Guess we need to change something on the website repo as well right?

done already :)

PR #13840, opened 2021-12-29
ci(deps): enable basic renovate

sdepold:
Enable renovate to get automatic dependency update


sdepold:
Interesting. I created this PR and suddenly we have #13841 .... I guess we can delete this one in that case?!

WikiRik:
Yeah, let's delete this one

Issue #12603, opened 2020-08-10
Typescript unusable & Typescript tutorial completely broken `"sequelize": "^6.3.4",`


stevemarksd:
- Is missing `await sequelize.sync()` before `async function doStuffWithUser() {`
 - Is missing to invoke `doStuffWithUser` function.
 - I also get a warning: `(sequelize) Warning: PostgresSQL does not support 'INTEGER' with LENGTH, UNSIGNED or ZEROFILL. Plain 'INTEGER' will be used instead`
 - Nothing is saved to the database: I get this error `ValidationError [SequelizeValidationError]: notNull Violation: User.name cannot be null`

After a long time figuring out what I had wrong in my code, I figured out why attributes are not saved to the db:

Let's say I copy the UserModel from the guide: (simplified)

```

  class User extends Model<UserAttributes, UserCreationAttributes>
    implements UserAttributes {
    public id!: number; // Note that the `null assertion` `!` is required in strict mode.
    public name!: string;
    public preferredName!: string | null; // for nullable fields

    // timestamps!
    public readonly createdAt!: Date;
    public readonly updatedAt!: Date;
  }

```

This will not save any attributes in the db. If I comment out the `public name!: string;` I get: `INSERT INTO "users" ("name","createdAt","updatedAt") VALUES ($1,$2,$3) RETURNING "id","name","preferredName","createdAt","updatedAt";` (note that there is no `preferredName`)

![image](https://user-images.githubusercontent.com/58653359/89747800-ec529f80-daf2-11ea-9c84-520b0455278b.png)

Notice how preferredName is null. That's because is in the model!

Now If I comment both attributes: `INSERT INTO "users" ("name","preferredName","createdAt","updatedAt") VALUES ($1,$2,$3,$4) RETURNING "id","name","preferredName","createdAt","updatedAt";` 

```

  class User extends Model<UserAttributes, UserCreationAttributes>
    implements UserAttributes {
    public id!: number; // Note that the `null assertion` `!` is required in strict mode.
   // public name!: string;
  //  public preferredName!: string | null; // for nullable fields

    // timestamps!
    public readonly createdAt!: Date;
    public readonly updatedAt!: Date;
  }

```

I get all keys!

![image](https://user-images.githubusercontent.com/58653359/89747823-09876e00-daf3-11ea-8b86-7818b7eee685.png)

As you can see this is unusable and I'm not even sure how to use it now with typescript.

[...]

Issue #11675, opened 2019-11-12
undefined attributes following typescript example


nicgirault:
### What are you doing?

```ts
// this is a copy of https://sequelize.org/master/manual/typescript.html
import { Sequelize, Model, DataTypes, BuildOptions } from 'sequelize';
import { HasManyGetAssociationsMixin, HasManyAddAssociationMixin, HasManyHasAssociationMixin, Association, HasManyCountAssociationsMixin, HasManyCreateAssociationMixin } from 'sequelize';

class User extends Model {
  public id!: number; // Note that the `null assertion` `!` is required in strict mode.
  public name!: string;
  public preferredName!: string | null; // for nullable fields

  // timestamps!
  public readonly createdAt!: Date;
  public readonly updatedAt!: Date;

  // Since TS cannot determine model association at compile time
  // we have to declare them here purely virtually
  // these will not exist until `Model.init` was called.

  public getProjects!: HasManyGetAssociationsMixin<Project>; // Note the null assertions!
  public addProject!: HasManyAddAssociationMixin<Project, number>;
  public hasProject!: HasManyHasAssociationMixin<Project, number>;
  public countProjects!: HasManyCountAssociationsMixin;
  public createProject!: HasManyCreateAssociationMixin<Project>;

  // You can also pre-declare possible inclusions, these will only be populated if you
  // actively include a relation.
  public readonly projects?: Project[]; // Note this is optional since it's only populated when explicitly requested in code

  public static associations: {
    projects: Association<User, Project>;
  };
}

const sequelize = new Sequelize('mysql://root:asd123@localhost:3306/mydb');

class Project extends Model {
  public id!: number;
  public ownerId!: number;
  public name!: string;

  public readonly createdAt!: Date;
  public readonly updatedAt!: Date;
}

class Address extends Model {
  public userId!: number;
  public address!: string;

  public readonly createdAt!: Date;
  public readonly updatedAt!: Date;
}

Project.init({
  id: {
    type: DataTypes.INTEGER.UNSIGNED, // you can omit the `new` but this is discouraged
    autoIncrement: true,
    primaryKey: true,
  },
  ownerId: {
    type: DataTypes.INTEGER.UNSIGNED,
    allowNull: false,
  },
  name: {
    type: new DataTypes.STRING(128),
    allowNull: false,
  }
}, {
  sequelize,
  tableName: 'projects',
});

User.init({
  id: {
    type: DataTypes.INTEGER.UNSIGNED,
[...]

Issue #11494, opened 2019-10-01
Mention known issues involving Babel + Sequelize in docs


papb:
### What was unclear/insufficient/not covered in the documentation

[Sequelize and Babel don't get along well sometimes...](https://github.com/sequelize/sequelize/issues/11326#issuecomment-522424850)

For example, @levibostian made [a thorough analysis of one of these issues](https://github.com/sequelize/sequelize/issues/11032#issuecomment-533661568), involving specifically the babel plugin called [`@babel/plugin-proposal-class-properties`](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties).

### If possible: Provide some suggestion on how we can enhance the docs

We could have a section on the docs about this.

### Additional context
None

### Is this issue dialect-specific?

- [X] No. This issue is relevant to Sequelize as a whole.
- [ ] Yes. This issue only applies to the following dialect(s): XXX, YYY, ZZZ
- [ ] I don't know.

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [X] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

ephys:
Done in https://github.com/sequelize/sequelize/pull/13877

Issue #11032, opened 2019-06-04
Can't insert records when using Typescript.


fenos:
I'm struggling for more than 4 hours now in one of the most basic operations with Sequelize using the `Typescript` example from the doc. I'm simply trying to store a simple record in SQL Server. I came to the conclusion that it's either a gigantic bug sitting somewhere OR the doc is omitting something crucial. Here is the code. 

## What are you doing?

```js
// connection.ts
export const sequelize = new Sequelize('mssql://sa:test@test:1433/Test', {
    dialectOptions: {
        instanceName: "SQLEXPRESS",
    }
});

// model.ts
class MemberInfo extends Model {
    public memb_guid!: number;
    public memb___id!: string;
    public memb__pwd!: string;
    public memb_name!: string;
    public sno__numb!: string;
    public bloc_code!: string;
    public ctl1_code!: string;
    public post_code!: string  | null;
    public addr_info!: string  | null;
    public addr_deta!: string  | null;
    public mail_addr!: string  | null;
    public fpas_ques!: string  | null;
    public fpas_answ!: string  | null;
    public mail_chek!: boolean | null;
}

MemberInfo.init({
    memb_guid: {
        type: DataTypes.INTEGER,
        autoIncrement: true,
        allowNull: false
    },
    memb___id: {
        type: DataTypes.STRING(10),
        primaryKey: true,
        allowNull: false
    },
    memb__pwd: {
        type: DataTypes.STRING,
        allowNull: false
    },
    memb_name: {
        type: DataTypes.STRING,
        allowNull: false
    },
    sno__numb: {
        type: DataTypes.STRING,
        allowNull: false
    },
    post_code: {
        type: DataTypes.STRING,
        allowNull: true
    },
    addr_info: {
        type: DataTypes.STRING,
        allowNull: true
    },
    addr_deta: {
        type: DataTypes.STRING,
        allowNull: true
    },
    tel__numb: {
        type: DataTypes.STRING,
        allowNull: true
    },
    phon_numb: {
        type: DataTypes.STRING,
        allowNull: true
    },
    mail_addr: {
        type: DataTypes.STRING,
        allowNull: true
    },
    fpas_ques: {
        type: DataTypes.STRING,
        allowNull: true
    },
    fpas_answ: {
        type: DataTypes.STRING,
        allowNull: true
    },
    job__code: {
        type: DataTypes.STRING,
        allowNull: true
    },
    appl_days: {
        type: 'SMALLDATETIME',
[...]

PR #13861, opened 2021-12-29
Update documentation for TypeScript Model Definition & Class Fields

ephys:
### Pull Request Checklist

### Description Of Change

There is an incompatibility between Sequelize & adding public class fields for attributes.

This PR documents how this incompatibility & explains how to add typing without introducing actual class fields that will cause bugs.

It also detects uses of public class fields that shadow a model's attributes, and warns the user. Doing this:

image

Will log the following to the console:

image

Closes https://github.com/sequelize/sequelize/issues/10579
Closes https://github.com/sequelize/sequelize/issues/12603
Closes https://github.com/sequelize/sequelize/issues/11675
Closes https://github.com/sequelize/sequelize/issues/11032
Closes https://github.com/sequelize/sequelize/issues/11494

---

Note: this documentation update is a temporary fix as things are going to break again once decorators are stage 4. I'll open an rfc with possible solutions that resolve this caveat for good for V7.

### DIFF
--- a/docs/manual/core-concepts/model-basics.md
+++ b/docs/manual/core-concepts/model-basics.md
@@ -101,6 +101,7 @@ User.init({
 const user = new User({ id: 1 });
 user.id; // undefined
 ```
+
 ```typescript
 // Valid
 class User extends Model {
--- a/docs/manual/core-concepts/model-basics.md
+++ b/docs/manual/core-concepts/model-basics.md
@@ -77,6 +77,68 @@ console.log(User === sequelize.models.User); // true
 
 Internally, `sequelize.define` calls `Model.init`, so both approaches are essentially equivalent.
 
+#### Caveat with Public Class Fields
+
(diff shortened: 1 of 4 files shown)


docs/manual/core-concepts/model-basics.md
```diff
@@ -77,6 +77,69 @@ console.log(User === sequelize.models.User); // true
...
+}, { sequelize });
+
+const user = new User({ id: 1 });
+user.id; // 1
+```
+
+In TypeScript, you can add typing information without adding an actual public class field by using the `declare` keyword:
+
+```typescript
+// Invalid
```
sdepold, line 126: is it invalid or valid? :D

sdepold (approved):

sdepold:
do we want to have that in v7 as well and eventually remove it?

docs/manual/core-concepts/model-basics.md
ephys: Good catch! Changed, it is valid :)

ephys:
Yes let's add it to v7 too until we've resolved https://github.com/sequelize/meetings/issues/16 :)

PR #13862, opened 2021-12-29
fix: semantic-release should not push beta or alpha to default channel

ChrisCrewdson:
### Pull Request Checklist

### Description Of Change

v6 and v7 have been publishing updated versions to the default `latest` tag and this should correct that.
Documentation: https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#channel

Evidence:
```
$ npm dist-tag ls sequelize
latest: 6.12.4
v4: 4.44.4
v5: 5.22.5
```

This means that anyone installing with `npm i sequelize@latest` will get alpha and beta versions (e.g., `7.0.0-alpha.3`) when they just happen to be the latest published versions.

### Todos

- [ ] Get this fix into v6
- [ ] Get this fix into main: https://github.com/sequelize/sequelize/pull/13863

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,11 +199,13 @@
       "v6",
       {
         "name": "v6-beta",
-        "prerelease": "beta"
+        "channel": "beta",
+        "prerelease": true
       },
       {
         "name": "v7",
-        "prerelease": "alpha"
+        "channel": "alpha",
+        "prerelease": true
       }
     ]
   },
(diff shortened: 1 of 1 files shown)


WikiRik:
We're merging main into v7 whenever we want to do a new release, so this PR is not needed. Right?

ChrisCrewdson:
> We're merging main into v7 whenever we want to do a new release, so this PR is not needed. Right?

Sounds good. I didn't know what the process would be like.

PR #13863, opened 2021-12-29
fix: semantic-release should not push beta or alpha to default channel

ChrisCrewdson:
### Pull Request Checklist

### Description Of Change

v6 and v7 have been publishing updated versions to the default `latest` tag and this should correct that.
Documentation: https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#channel

Evidence:
```
$ npm dist-tag ls sequelize
latest: 6.12.4
v4: 4.44.4
v5: 5.22.5
```

This means that anyone installing with `npm i sequelize@latest` will get alpha and beta versions (e.g., `7.0.0-alpha.3`) when they just happen to be the latest published versions.

### Todos

- [ ] Get this fix into v6
- [x] Get this fix into v7 (consider https://github.com/sequelize/sequelize/pull/13862)

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,10 +199,12 @@
       "v6",
       {
         "name": "v6-beta",
+        "channel": "beta",
         "prerelease": "beta"
       },
       {
         "name": "v7",
+        "channel": "alpha",
         "prerelease": "alpha"
       }
     ]


package.json
```diff
@@ -199,10 +199,12 @@
       "v6",
       {
         "name": "v6-beta",
+        "channel": "beta",
         "prerelease": "beta"
```
WikiRik, line 203: Is this still needed?

package.json
```diff
@@ -199,10 +199,12 @@
...
       },
       {
         "name": "v7",
+        "channel": "alpha",
         "prerelease": "alpha"
```
WikiRik, line 208: Same comment as above

WikiRik (approved):
Looks good, thanks for the PR! Do have one comment though

EDIT: Comment resolved, all good to go from my side

package.json
ChrisCrewdson: I believe it is. "prerelease" will be used as the suffix by `semantic-release` and generate releases like `6.0.0-beta`. Otherwise, this will make releases from `v6-beta` without the suffix.

Documentation: https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#prerelease

package.json
ChrisCrewdson: Same answer as above. It is needed to make releases like `7.0.0-alpha` instead of `7.0.0` from the v7 branch. Once v7 is a stable release channel, it should be a bare "v7" like the "v6" branch reference (or just remove the "channel" and "prerelease" properties, leaving only "name").

package.json
WikiRik: Ah yes, that makes sense. Thanks!

sdepold: V6 Beta can be removed completely

sdepold:
Tbh maybe it's me not understanding something but it sounds somehow relatable that you will get the alpha version when you ask for latest, no?

sdepold:
Going to revert this since I believe it was actually behaving correctly.

sdepold:
I would have kept it, but the change seems to have broken the release process somehow and now semantic-release can no longer find the right base version?!

sdepold:
https://github.com/sequelize/sequelize/actions/runs/1643568186

ChrisCrewdson:
And it looks like this is back to the old bad behavior where latest is an alpha:
```
$ npm dist-tag ls sequelize
latest: 7.0.0-alpha.4
v4: 4.44.4
v5: 5.22.5
```

ChrisCrewdson:
I think the v7 branch needs a commit to bump the patch version to `7.0.1-alpha` because the release error was:
```
fatal: tag 'v7.0.0-alpha.1' already exists
```

sdepold:
Yes. A) I'm not sure if this is even a bug to begin with and b) semantic release fails finding the right tags now. I'm happy to merge a working pr if you can find out what needs to be done to get the release working again. See link above for the problem

ChrisCrewdson:
> Tbh maybe it's me not understanding something but it sounds somehow relatable that you will get the alpha version when you ask for latest, no?

No, the `latest` published to npm should be a stable version. You can still request the alpha with something like `npm i sequelize@alpha`, but `npm i sequelize` is the same as `npm i sequelize@latest`.
The way it currently is, a new user could `npm i sequelize` at the wrong time and get the alpha.

sdepold:
> I think the v7 branch needs a commit to bump the patch version to `7.0.1-alpha` because the release error was:
> 
> ```
> 
> fatal: tag 'v7.0.0-alpha.1' already exists
> 
> ```

How do I bump to alpha.4 so that the next release would be alpha.5?

PR #13869, opened 2021-12-30
build(ci): add extra stages to ci checks

WikiRik:
Introduces extra stages to the GitHub Actions CI workflow. Should limit the time needed to check on PRs that are obviously failing (like we had with the esbuild)
chrome_1tbn3LGuFJ
One thing that's not clear from the picture is that the test-mysql-mariadb is not dependent on docs, docs goes directly to release as you can see here; https://github.com/sequelize/sequelize/actions/runs/1637499004

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
   docs:
     name: Generate docs
     runs-on: ubuntu-latest
+    needs: lint
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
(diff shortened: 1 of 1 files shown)


sdepold:
I wonder if we want to to have the following flow:

Linting + docs + typing
-----
Unit tests for all dialects
-----
Integration test for all dialects
-----
Release

WikiRik:
Don't think that would be efficient. Unit tests only take a few seconds and if we would split the unit tests from the integration tests we would have to spin up the same containers twice which would take longer

PR #13870, opened 2021-12-30
test(mysql8): disable flaky mysql8 test

WikiRik:
The test itself is not a problem, but for some reason it just fails for MySQL 8. But it's not really reproducible so I think it's better to just disable it for now

### DIFF
--- a/test/integration/associations/scope.test.js
+++ b/test/integration/associations/scope.test.js
@@ -5,7 +5,9 @@ const chai = require('chai'),
   Support = require('../support'),
   DataTypes = require('sequelize/lib/data-types'),
   Sequelize = require('sequelize'),
-  Op = Sequelize.Op;
+  current = Support.sequelize,
+  Op = Sequelize.Op,
+  semver = require('semver');
 
 describe(Support.getTestDialectTeaser('associations'), () => {
   describe('scope', () => {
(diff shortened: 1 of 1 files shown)


sdepold:
Isn't that also frequently failing for MySQL?

jesse23:
> Isn't that also frequently failing for MySQL?

Yes...we need to see:
- Is there a bug inside our code that we are not maintain the query sequence correctly in this kind of ref use case?
- If not, is it as design that the return value from SQL is not order guaranteed?
   - If yes, shall we sort them in code to make them follow a default order?

WikiRik:
> Isn't that also frequently failing for MySQL?

You mean MySQL 5.7? Or MS SQL with the timeouts when starting?

WikiRik:
> > Isn't that also frequently failing for MySQL?
> 
> Yes...
> 
> We need to see:
> 
> * Is there a bug inside our code that we are not maintain the query order correctly in this kind of ref use case?
> * If not, is it as design that the return value from SQL is not order guaranteed?
>   
>   * If yes, shall we sort them in code to make them follow a default order?

Yeah, we should make a separate issue to debug and fix this test (or a possible bug in the code) but for now it does not seem to affect end users but it does affect our development experience

fncolon:
In my opinion, it's better to find out the source of the problem, why is it flaky first and also implement the fix to all dialects flaky tests

fncolon:
It could be also problem with async/await i guess

```js
await Promise.all([this.Post.sync({ force: true }), this.Tag.sync({ force: true })]);
await this.PostTag.sync({ force: true });
```

i don't know why `await this.PostTag.sync({ force: true });` not included also inside the Promise.all.

It might be added in the database, but got sync forced by the `await this.PostTag.sync({ force: true });` and the data got deleted, so the result is 0.

fncolon:
> It could be also problem with async/await i guess
> 
> ```js
> await Promise.all([this.Post.sync({ force: true }), this.Tag.sync({ force: true })]);
> await this.PostTag.sync({ force: true });
> ```
> 
> i don't know why `await this.PostTag.sync({ force: true });` not included also inside the Promise.all.
> 
> It might be added in the database, but got sync forced by the `await this.PostTag.sync({ force: true });` and the data got deleted, so the result is 0.

Actually what i mean we can gather many possibilities as possible why we got flaky tests.

sdepold:
Posttag must not be synced before post and Tag is created hence it is in the second line. Otherwise it will not be able to reference the other tables

fncolon:
> Posttag must not be synced before post and Tag is created hence it is in the second line. Otherwise it will not be able to reference the other tables

async/await do not run synchronously, so it might be done after postA.getCategory(), or any line in the code. It could be one of many possibilities too.

jesse23:
I feel `[flaky]` with enough print is good for now so that:
- We still get the test running for real regression.
- We have enough print to identify the failure reason ( because we know the flaky print alread)

@sdepold please set approve if u are OK with disabling it.

ephys:
Given that this test blocks other PRs, I think we should disable it. At least until we figure out why it's broken.

ephys:
> async/await do not run synchronously, so it might be done after postA.getCategory(), or any line in the code. It could be one of many possibilities too.

@fncolon `await` will "suspend" the execution of that function until the promise returned by `this.PostTag.sync({ force: true });` so unless there is a major issue in `sync()` (such as the promise resolving too soon, but unlikely), it's impossible for `postA.getCategory()` to execute before `PostTag.sync`

WikiRik:
> Given that this test blocks other PRs, I think we should disable it. At least until we figure out why it's broken.

Yeah, it's a single test for one version of a dialect we're supporting that will be disabled temporarily. This will save quite some time in merging PRs since this (together with MSSQL instances not booting properly) is the main reason we need to run several attempts to merge many PRs

We should indeed find why it fails for MySQL 8.0 but for now, its impact on maintainers is bigger than on users.

ephys:
@WikiRik it looks like the test still ran in https://github.com/sequelize/sequelize/pull/13899 on MySQL 8.0

WikiRik:
It seems to have run on this PR as well so apparently I need to fix something. Will do later today

PR #12621, opened 2020-08-18
Ability to set custom connection timeout in postgres client.

gabrieloczkowski:
What's new:
- Ability to set `connectionTimeoutMillis` for postgres client using `config.dialectOptions`.

PR #12597, opened 2020-08-06
docs(dialect-specific): add details about dialectOptions

dmercertaylor:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Have you added new tests to prevent regressions?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you update the typescript typings accordingly (if applicable)?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/master/CONTRIBUTING.md)?

This is just docs so most of these don't apply.

### Description of change

Closes #10992. Added links to docs with dialect options for each library, and expanded upon what exactly could be passed into Postgres and SQLite since those are filtered to an extent.

PR #13878, opened 2022-01-01
docs(dialect-specific): add details about dialectOptions

fncolon:
Copy of #12597 

---

Possibility to continue the work of #12636 #12621

Lets see if we have dialectOptions which is still not exposed

cc active maintainers @jesse23 @sdepold @WikiRik @ephys


jesse23 (approved):

jesse23:
Looks good to me. Thx!

PR #13881, opened 2022-01-02
feat: add options.rawErrors to `Sequelize#query` method 

### LINKED ISSUE #13838: Clarify supported Microsoft SQL Server versions
## Issue Creation Checklist

## Issue Description

In `.github/workflows/ci.yml:200` there is the line
```
mssql-version: [2017, 2019]
```

In `lib/dialects/mssql/index.js:63` there is the line
```
MssqlDialect.prototype.defaultVersion = '12.0.2000'; // SQL Server 2014 Express, minimum supported version
```
In the documentation, there isn't a referenced minimum version
https://sequelize.org/master/manual/dialect-specific-things.html#mssql

If you need to refresh which versions of MS SQL Server exist, there's wikipedia, which also includes their end-of-support dates.
https://en.wikipedia.org/wiki/SQL_Server_Express#Version_history

### Possible Solution

The issue is that the three sources above don't match, and hence it is not entirely clear which is the minimum version supported.
If 2014 (version 12) shall be supported, there should at the very least be a CI job testing against that version. Microsoft is quite good with backwards compatibility so testing for that version should ensure compatibility with all newer versions.

Else, `mssql/index.js`  should be updated to mention version 14 and the version should be added to the docs.

- [x] Yes. This issue only applies to the following dialect(s): MS SQL
- [ ] I don't know.

joaoe:
### Pull Request Checklist

### Description Of Change

**feat(query): added options.rawErrors to query() method**

This option causes all errors from the underlying connection/database library
to be propagated unformatted and unmodified. Use this for lesser common errors
sequelize can't parse or in case you prefer to see the full raw errors.

Use case: in my project, we have a central database where different apps connect, like some scientific libraries and a REST API implemented using Node.JS. All errors in all apps are logged centrally. So, if there are DB errors, it is mostly productive to get the original error messages from the database so all developers understand them, and not just those that develop with node.js+sequelize. Also, Sequelize unfortunatelly cannot parse all errors, like at some point there was a cryptic error, because sequelize could not insert in a table with a trigger, something that was easily visible once we saw the original db error.

So, the rawErrors option is a nice to have for those that are not so interested in Sequelize's own error parsing. Also, it saves some CPU cyles. It can be set per query or in the global `sequelize.options.query`.

Also bonus: stack handling for db2.

### Todos

- [ ] Wait for all pipelines to run green


package.json
```diff
@@ -207,7 +207,7 @@
   },
   "scripts": {
     "----------------------------------------- static analysis -----------------------------------------": "",
-    "lint": "eslint lib test --quiet --fix",
+    "lint": "eslint lib test *.[tj]s --quiet --fix",
```
WikiRik, line 210: This would still limit eslint to the lib and test folder right? Don't think this actually changes anything

.github/workflows/ci.yml
```diff
@@ -209,7 +209,14 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [12, 16]
-        mssql-version: [2017, 2019]
+        # 2014 (version 12) is the lowest support MS SQL version.
```
WikiRik, line 212: Could we have all these comments in the dialects index.js file? I think it would fit better there and there is no real need to have all this information twice.

WikiRik (commented):
Thanks for the PR! I only looked at the mssql docs changes and the eslint changes for now and have some comments on that

WikiRik:
And it seems that Microsoft currently only offers 2017 & 2019 Docker images so we need to figure out a different way to test on 2014 if we really want to; https://hub.docker.com/_/microsoft-mssql-server

.github/workflows/ci.yml
joaoe: Sure

package.json
joaoe: Yes, it limits to the test and lib folders plus all ts+js files in the root, like index.js and build.js.

joaoe:
> And it seems that Microsoft currently only offers 2017 & 2019 Docker images so we need to figure out a different way to test on 2014 if we really want to; https://hub.docker.com/_/microsoft-mssql-server

Bummer.
Perhaps phase out ms sql 2014 and 2016 support for Sequelize v7 due to lack of support from MS ?

package.json
ephys: While we're changing this, can you include `m` & `c` prefix for mjs, cjs, mts, & cts?

lib/dialects/abstract/query.js
```diff
@@ -23,6 +23,10 @@ class AbstractQuery {
       ...options
     };
     this.checkLoggingOption();
+
+    if (options.rawErrors) {
```
ephys, line 27: I don't like changing methods at runtime but I get that it's easier to bypass method overrides this way.

For future us, can you document this line?

build.js
```diff
@@ -35,16 +35,58 @@ async function rmDistDir() {
...
+      const destStat = fs.statSync(dest, { throwIfNoEntry: false });
+      const ok = sourceStat && (!destStat || destStat.mtimeMs < sourceStat.mtimeMs);
+      resolve(ok ? source : null);
+    })))).filter(Boolean);
+}
+
+async function listFilesToGenerate() {
+  let toCompile, rmPromise;
+
+  const toCompilePromise = glob('./lib/**/*.[tj]s', { onlyFiles: true, absolute: false });
```
ephys, line 56: same as eslint, while we're changing this can you add c & m optional prefixes?

ephys (changes requested):

lib/dialects/abstract/query.js
joaoe: Hi. I agree it is not the prettiest of things, but the alternative would be to check the flag in every single dialect. And newer dialects could easily miss that. So at least this solution is very simple.

I did see elsewhere in `lib/dialect/abstract/query-generator.js` the method `quoteIdentifier` being wrapped, in case you consider modifying methods is something you don't like or should not have,

What would you like to be documented there ?

ephys: Simply explain what this line does (prevent descending classes from overriding this method) & why (so calling `.formatError` on the dialect returns the database error untranformed).

test/support.js
```diff
@@ -101,14 +101,13 @@ const Support = {
 
     const config = Config[options.dialect];
 
-    const sequelizeOptions = _.defaults(options, {
-      host: options.host || config.host,
+    const sequelizeOptions = _.defaultsDeep(options, {
```
sdepold, line 104: Since _.defaultsDeep (as well as _.default) overrides the input object, I wonder if we should rather change it to:

```
const sequelizeOptions = _.merge({
      host: config.host,
      logging: process.env.SEQ_LOG ? console.log : false,
      port: process.env.SEQ_PORT || config.port,
      pool: config.pool,
      dialectOptions: config.dialectOptions || {},
      minifyAliases: config.minifyAliases
    }, options)

```

sdepold: Not 100% sure if it is producing the same object, but it projects the idea :)

build.js
```diff
@@ -35,16 +35,58 @@ async function rmDistDir() {
...
+
+  if (process.argv.indexOf('--check-mtime') >= 0) {
```
sdepold, line 58: Can you add an example on the top of the file on how to use it?

build.js
```diff
@@ -35,16 +35,58 @@ async function rmDistDir() {
...
+  return path.resolve(outdir, source).replace(/\.ts$/, '.js');
+}
+
+async function ageCheckFiles(paths) {
+  return (await Promise.all(
+    paths.map(source => new Promise(resolve => {
+      const dest = sourcePathToDestPath(source);
+      const sourceStat = fs.statSync(source, { throwIfNoEntry: false });
+      const destStat = fs.statSync(dest, { throwIfNoEntry: false });
+      const ok = sourceStat && (!destStat || destStat.mtimeMs < sourceStat.mtimeMs);
```
sdepold, line 48: Do I understand it correctly, that we will only update the files that have changed?

test/integration/sequelize/query.test.js
```diff
@@ -342,6 +342,20 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
 
+        it('emits raw errors if requested', async function() {
+          const sql = 'SELECT 1 FROM NotFoundTable';
+          try {
+            await this.sequelize.query(sql, { rawErrors: false });
+          } catch (err) {
+            expect(err).to.be.instanceOf(DatabaseError);
+          }
+          try {
+            await this.sequelize.query(sql, { rawErrors: true });
```
sdepold, line 353: Should we make sure that we actually hopped into the catch block? expect.fail in the next line?

test/integration/sequelize/query.test.js
```diff
@@ -342,6 +342,20 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+          } catch (err) {
+            expect(err).to.not.be.instanceOf(DatabaseError);
```
sdepold, line 355: Can you please assert against the actual error?

test/integration/sequelize/query.test.js
```diff
@@ -342,6 +342,20 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
           await this.UserVisit.sync({ force: true });
         });
 
+        it('emits raw errors if requested', async function() {
+          const sql = 'SELECT 1 FROM NotFoundTable';
+          try {
+            await this.sequelize.query(sql, { rawErrors: false });
+          } catch (err) {
+            expect(err).to.be.instanceOf(DatabaseError);
```
sdepold, line 350: Can you check that it is NOT the actual error class?

test/support.js
joaoe: I don't think it's a problem to change the config, but sure.

build.js
joaoe: Right, I was going to ask the question: how should I document this so people learn to use the feature ? Perhaps add also a new entry in package.json ? like build-fast or some such ?

test/integration/sequelize/query.test.js
joaoe: The actual error is something thrown deep within the connection/database library, which is dialect specific I think.

test/integration/sequelize/query.test.js
joaoe: > the actual error class?
Same problem as before, the "actual error class" is dialect specific, and always != sequelize's own DatabaseError.

build.js
sdepold: `build-dev` maybe?

test/integration/sequelize/query.test.js
sdepold: alright!

build.js
joaoe: `build-dev` sounds like a dev build, something that would typically contain debug symbols and so forth.
Something like `build-fast` or `build-newest` would be more appropriate.

test/integration/sequelize/query.test.js
joaoe: fixed

build.js
ephys: `build-incremental` like tsc's flag? I'm also ok with the other names

ephys (dismissed):
Good on my end. Just need to resolve other reviewer's requests & fix tests :)

build.js
joaoe: ok, incremental it is

joaoe:
Howdy.

Life's busy, but I've just updated the PR with a rebased branch.
The MS SQL commit I shall do another PR.
The commit to pass the root js files to the linter is now not needed since someone else changed that in another PR.

build.js
joaoe: Is it good enough with the entry in package.json ?
Else, it would be a bit awkward to write a single comment, when there should actually be a header there a bit more verbose that explains what the script does.

joaoe:
@ephys Hi. The pipeline is acting weird. A lot of MS SQL and MySQL fails which are completely unrelated with my work.

ephys:
Updating the branch to see if it still happens. It's not happening in the other PRs so if it still happens here it must be a subtle issue that was introduced :/

ephys:
Unfortunately they are still failing. Could be related to the changes made in `test/support.js`?

joaoe:
> Unfortunately they are still failing. Could be related to the changes made in `test/support.js`?

That would be quite fascinating given the randomness of the fails. Oh well, tomorrow I might try to push the branch one commit at a time.

build.js
joaoe: done

build.js
```diff
@@ -32,16 +42,60 @@ async function rmDistDir() {
   }
 }
 
-async function main() {
-  console.info('Compiling sequelize...');
-  const [declarationFiles, filesToCompile] = await Promise.all([
+function sourcePathToDestPath(source) {
+  return path.resolve(outdir, source).replace(/\.ts$/, '.js');
```
ephys, line 46:
```suggestion
  return path.resolve(outdir, source).replace(/\.([cm])?ts$/, '.$1js');
```

to support cts/mts again :)

ephys (changes requested):
Last note on my end I think, the rest looks good to me

joaoe:
@ephys I'm very confused. Something happened recently with the project now the partial build takes as long at the whole build... bah ! Since there is no benefit here, I'll just skip this commit as well... :/

WikiRik:
> @ephys I'm very confused. Something happened recently with the project now the partial build takes as long at the whole build... bah ! Since there is no benefit here, I'll just skip this commit as well... :/

Strange. Do you happen to know around what time this change happened? For example a release/commit in which it still worked. So we can try to figure out what happened

joaoe:
> > @ephys I'm very confused. Something happened recently with the project now the partial build takes as long at the whole build... bah ! Since there is no benefit here, I'll just skip this commit as well... :/
> 
> Strange. Do you happen to know around what time this change happened? For example a release/commit in which it still worked. So we can try to figure out what happened

The past two weeks or so. There was a big commit about making eslint becoming more stringent and doing autoformatting. Perhaps that changed some settings. I could try to bisect this but then can leave it for another PR.

joaoe:
> > @ephys I'm very confused. Something happened recently with the project now the partial build takes as long at the whole build... bah ! Since there is no benefit here, I'll just skip this commit as well... :/
> 
> Strange. Do you happen to know around what time this change happened? For example a release/commit in which it still worked. So we can try to figure out what happened

I might be hallucinating. I tried a month old commit from early december. It doens't work.
I think I might be confusing build.js not doing anything when I just edited test files.

WikiRik (dismissed):
Looks good to me

build.js
joaoe: obsolete

build.js
joaoe: obsolete

joaoe:
I see a fail in DB2 + Node 16 but that is also happening in the main branch.

WikiRik:
Yeah we got a flaky test there, don't worry about that

ephys:
I think we can safely add this to v6 too

joaoe:
ok, who can merge ?

ephys:
Thank you for the PR :)

PR #13883, opened 2022-01-02
docs: sync documentation to latest changes

fncolon:
### Pull Request Checklist

### Description Of Change

- Since we have version picker, we don't need documentation section anymore, and some additional changes to keep README.md simpler as much as possible.

- Adding repository discussions as Sequelize resource

- Remove dollar sign so copy is much easy


ephys:
I agree with most changes but you should keep the documentation section even if it's just a small paragraph redirecting to https://sequelize.org/

fncolon:
> I agree with most changes but you should keep the documentation section even if it's just a small paragraph redirecting to https://sequelize.org/

It's avoid future changes for chaining version inside documentation section and different latest version for https://sequelize.org/master

How about

```md
## Documentation

- [Latest Version Documentation](https://sequelize.org/master)
- [Homepage](https://sequelize.org)
- [Contributing](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)
```

ephys:
I think we use `/main` instead of `/master` for the documentation too now.

If we include the latest documentation version, we should specify for which version it is for (v7) to reduce risk of confusion.

And I propose "All documentation versions" instead of "Homepage" so users not looking for v7 have a easier time finding it :)

fncolon:
> I think we use `/main` instead of `/master` for the documentation too now.
> 
> If we include the latest documentation version, we should specify for which version it is for (v7) to reduce risk of confusion.
> 
> And I propose "All documentation versions" instead of "Homepage" so users not looking for v7 have a easier time finding it :)

Noted

fncolon:
Another solution to reduce risk of confusion, maybe to provide `/latest` which is will automatic redirecting to latest version of Sequelize.

README.md
```diff
@@ -33,33 +33,28 @@ We have recently created an [OpenCollective based money pool](https://opencollec
...
 
 If you have security issues to report, please refer to our [Responsible Disclosure Policy](https://github.com/sequelize/sequelize/blob/main/SECURITY.md) for more details.
 
 ## Resources
 
 - [Changelog](https://github.com/sequelize/sequelize/releases)
+- [Discussions](https://github.com/sequelize/sequelize/discussions)
 - [Slack Inviter](http://sequelize-slack.herokuapp.com/)
-- [Stack Overflow](https://stackoverflow.com/questions/tagged/sequelize.js)
+- [StackOverflow](https://stackoverflow.com/questions/tagged/sequelize.js)
```
WikiRik, line 57:
```suggestion
- [Stack Overflow](https://stackoverflow.com/questions/tagged/sequelize.js)
```

Stack Overflow uses its name with a space

ephys:
> If we include the latest documentation version, we should specify for which version it is for (v7) to reduce risk of confusion.

Sorry, what I meant was add "v7" to the link name, especially since it's a pre-release and everyone will stay on v6 until it's ready :)

fncolon:
> > If we include the latest documentation version, we should specify for which version it is for (v7) to reduce risk of confusion.
> 
> Sorry, what I meant was add "v7" to the link name, especially since it's a pre-release and everyone will stay on v6 until it's ready :)

can you elaborate about link name?

ephys:
Simply "Latest Version Documentation (v7)" instead of "Latest Version Documentation"

sdepold:
Why do we remove the changelog and and links? I think they are really useful

fncolon:
> Why do we remove the changelog and and links? I think they are really useful

Sorry, but I didn't remove the changelog.
Alreayd revert the links too.

PR #13884, opened 2022-01-02
fix(model): fix count with grouping typing

### LINKED ISSUE #13871: Typings for the `Model.count()` method are inaccurate
## Issue Creation Checklist

## Bug Description

The TypeScript typings indicate that whenever the `group` option is provided to the `Model.count()` function, then an object will be returned with variable string keys containing numbers. However, what is actually returned is an array of objects, which contain the values for the given attributes supplied to the `group` option and a `count` key containing the actual number of results.

### SSCCE

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/208

See above.

See above. Another view of output: 

```js
> await AutoUser.count({
...       group: ["hour", "minute"]
...     })
[ { hour: 5, minute: 40, count: 1 } ]
> await AutoUser.count({
... group: ["hour"]
... });
[ { hour: 5, count: 1 } ]
```

### Additional context

N/A

### Environment

- Sequelize version: sequelize@6.12.4 
- Node.js version: v17.3.0 
- If TypeScript related: TypeScript version: typescript@4.5.4

ephys:
### Pull Request Checklist

### Description Of Change

Closes #13871

Current typings state that count with grouping returns an object `{ [columnName: count] }` when in reality something like this:

image

returns this:

image

An array where each entry is an object with column -> column value of the group, plus a `count` property.

`count.test.js` backs this, so it's only a typing issue.

---

I also noticed `Model.count({ group: ['count'] })` is going to return a broken result. I've added it to my [future improvements list](https://gist.github.com/ephys/50a6a05be7322c64645be716fea6186a).

### DIFF
--- a/types/lib/model.d.ts
+++ b/types/lib/model.d.ts
@@ -1907,7 +1907,7 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
   public static count<M extends Model>(
     this: ModelStatic<M>,
     options: CountWithOptions<M['_attributes']>
-  ): Promise<{ [key: string]: number }>;
+  ): Promise<Array<{ [groupKey: string]: unknown, count: number }>>;
 
   /**
    * Count the number of records matching the provided where clause.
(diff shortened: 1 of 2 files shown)


fncolon (approved):

fncolon:
Looks good to me

Unknown : i don't know
Any : i don't care

😄

PR #13890, opened 2022-01-03
fix: improve ModelCtor / ModelStatic typing

ephys:
### Pull Request Checklist

### Description Of Change

Alternative PR for https://github.com/sequelize/sequelize/pull/13834

I'm hoping I finally figured out why ModelCtor / ModelStatic are so tricky to use. I tried fixing ModelStatic & Merging both types together instead of having two different ones.

Crossing my fingers for tests to pass.

Edit: this should land in v6


hsource:
I tested this against our very-large Typescript codebase and it works well! LGTM

types/test/type-helpers/deep-writable.ts
```diff
@@ -6,10 +6,34 @@
...
+  Sequelize,
+  ModelCtor,
+  ModelDefined,
+  ModelStatic,
+} from 'sequelize';
 
-type Builtin = string | number | boolean | bigint | symbol | undefined | null | Function | Date | Error | RegExp;
-type SequelizeBasic = Builtin | Sequelize | Model | ModelCtor<Model> | ModelType | ModelDefined<any, any> | ModelStatic<Model>;
+type Builtin =
+  | string
```
sdepold, line 18: Is the pipe in the beginning on purpose?

sdepold (commented):
I have to admit that I'm still a bit confused by all those types.

Is this correct?:

const User: ModelStatic = sequelize.define('user');
const user:Model = user.findOne();

types/test/type-helpers/deep-writable.ts
ephys: Yes, it's so all lines look the same no matter whether it's the first element or the last. Good practice for future diffs :)

ephys:
> I have to admit that I'm still a bit confused by all those types.

Your code is correct. ModelStatic represents the class itself, Model represents an instance of the class

`ModelStatic` lets us request the model itself in functions, which can be used to do things like this:

image

image

We used to have 3 different types to represent the class itself (ModelType, ModelCtor, ModelStatic) but this PR makes `ModelStatic` universal and able to replace the other 2 types

sdepold:
>We used to have 3 different types to represent the class itself (ModelType, ModelCtor, ModelStatic) but this PR makes ModelStatic universal and able to replace the other 2 types

Exactly and it drove me half crazy the other day. :praisethesun:

sdepold:
Nice editor theme

fncolon:
What theme is it? @ephys haha

ephys:
@fncolon If I remember right the base was Material Palenight (From Material Theme UI for IntelliJ)

fncolon:
> t

Thank you!

PR #13892, opened 2022-01-03
fix(types): omit FK and scope keys in HasManyCreateAssociationMixin

EmilioCristalli:
### Pull Request Checklist

### Description Of Change

When using `HasManyCreateAssociationMixin `, and similar methods, Sequelize automatically sets the FK value using the source key from the source model, as well as attributes used in the scope, so the `create*` mtheods shouldn't allow to specify these values.

I couldn't think of an automatic way to do this so i added new generic options, but i made them optionals to keep it backwards compatible (though i'm not super convinced the FK generic should be optional).

Didn't do it for all the association types because i wanted to get some feedback first, to make sure this approach makes sense.

Related to https://github.com/sequelize/sequelize/pull/13685/files#r777743536

### Todos

- [ ] Update the rest of the association mixins
- [ ] 
- [ ]

### DIFF
--- a/types/lib/associations/has-many.d.ts
+++ b/types/lib/associations/has-many.d.ts
@@ -209,8 +209,12 @@ export interface HasManyCreateAssociationMixinOptions extends CreateOptions<any>
  * @see https://sequelize.org/master/class/lib/associations/has-many.js~HasMany.html
  * @see Instance
  */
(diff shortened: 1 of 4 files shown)


fncolon (approved):
Looks good to me

mmahalwy (commented):
lgtm

Issue #11847, opened 2020-01-19
Explain `joinTableAttributes` versus `include: { attributes }` in the manuals


papb:
## Issue Description

Currently the `joinTableAttributes` option is barely mentioned in the manuals ([ctrl+f here](https://sequelize.org/master/manual/advanced-many-to-many.html)).

### What was unclear/insufficient/not covered in the documentation

What does it do and how it relates to the more known `include: { attributes: [ /* ... */ ] }` option

### If possible: Provide some suggestion on how we can enhance the docs

Just briefly mention it in the manuals and explain its difference, no need for complicated things...

### Is this issue dialect-specific?

- [X] No. This issue is relevant to Sequelize as a whole.
- [ ] Yes. This issue only applies to the following dialect(s): XXX, YYY, ZZZ
- [ ] I don't know.

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [X] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

msallent:
THANK YOU!

I spent hours trying to find how to omit the through table's attributes entirely while using a mixin and this was exactly what I was looking for.

Adding to this, using joinTableAttributes is not possible while using TypeScript as the type BelongsToManyGetAssociationsMixin doesn't include it so it might be a good idea to update all the mixin types as well.

papb:
@msallent

> THANK YOU!
> 
> I spent hours trying to find how to omit the through table's attributes entirely while using a mixin and this was exactly what I was looking for.

:grimacing:

> Adding to this, using joinTableAttributes is not possible while using TypeScript as the type BelongsToManyGetAssociationsMixin doesn't include it so it might be a good idea to update all the mixin types as well.

Actually this was already fixed earlier this week at #11818, however there was no release yet.

PR #13902, opened 2022-01-04
docs: `joinTableAttributes` in `belongsToMany` mixin

cincodenada:
This address #11847 partially at least - I'm not sure where else it should be documented.

### DIFF
--- a/docs/manual/core-concepts/assocs.md
+++ b/docs/manual/core-concepts/assocs.md
@@ -611,7 +611,7 @@ The same ones from `Foo.hasMany(Bar)`:
 * `fooInstance.removeBars()`
 * `fooInstance.createBar()`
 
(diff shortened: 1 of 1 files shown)


cincodenada:
I'm still not _quite_ sure why this doesn't work - I see that we're manually adding an entry to the include array in [belongs-to-many.js](https://github.com/sequelize/sequelize/blob/4443d2af14c78b21ff2a70f4aeb69bd9d3f8c2e2/lib/associations/belongs-to-many.js#L458)

I'm guessing this is so that we have the join model attributes by default, and because we're not controlling the join, it would be awkward to try to add it. A naive attempt at suppressing the join table would look something like:
```
Foo.getBars({
    include: [{
      model: Bar,
      through: { attributes: [] }
    }]
})
```
but that would actually attempt to join the returned `Bar`s in turn to `Bar`, which won't work (or at least won't have the intended effect, and I'm not sure it's even possible for the end-user to replicate the join that's generated reasonably.

I don't know the best way to explain that succinctly - I've done my best here.

fncolon (approved):

yisibl:
Thank you very much for adding the documentation.

Issue #13901, opened 2022-01-04
Support dropping index CONCURRENTLY in Postgres


shayonj:
## Issue Creation Checklist

- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Feature Description

Support drop index [`CONCURRENTLY`](https://www.postgresql.org/docs/9.1/sql-createindex.html) in postgres. Postgres has the ability to add or drop an index `CONCURRENTLY`. This operation tends to be safer and more reliable in production (dependingon the workloads). Since there is already an option to pass `concurrently: true` to `addIndex`, it'd be great to support the same for `removeIndex`.

Example: 
```javascript
  await queryInterface.removeIndex({ schema: '<schema>', tableName: 'Person' }, 'SuperDuperIndex', { concurrently: true });
```

### Is your feature request related to a problem? Please describe.

No
### Describe the solution you'd like

Since there is already an option to pass `concurrently: true` to `addIndex`, it'd be great to support the same for `removeIndex`.

Example: 

```javascript
  await queryInterface.removeIndex({ schema: '<schema>', tableName: 'Person' }, 'SuperDuperIndex', { concurrently: true });
```

### Why should this be in Sequelize

To keep this in parity with `addIndex`, also providing an easy way for developers to write safer SQL via ORM w/o having to write raw sql. Additionally, due to the coupled nature and the query being generated from sequelize, it perhaps makes sense for this to be in sequelize package. 

### Describe alternatives/workarounds you've considered

Writing a raw sql statement over using the DSL.

### Additional context

Mentioned Above

### Is this feature dialect-specific?

- [ ] No. This feature is relevant to Sequelize as a whole.
- [x]  Yes. This feature only applies to the following dialect(s): Postgres

### Would you be willing to implement this feature by submitting a Pull Request?

- [x]  Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

I know how to start and will share a PR soon. Would really appreciate feedback and guidance in the PR.

PR #13903, opened 2022-01-04
feat(postgres): drop indices concurrently in Postgres

shayonj:
### Pull Request Checklist

### Description Of Change

This is a follow up to the [feature request](https://github.com/sequelize/sequelize/issues/13901).
On `removeIndex` it now extends the `options` object
to take in a new attribute, `concurrently`, which
will ensure that the the SQL statement being run
includes `DROP INDEX CONCURRENTLY`...

Having this is nice so developers can use the ORM settings
when dropping an index, safely, instead of having to write
raw SQL.

`concurrently` is a supported attribute on addIndex

Closes https://github.com/sequelize/sequelize/issues/13901

### DIFF
--- a/lib/dialects/mssql/connection-manager.js
+++ b/lib/dialects/mssql/connection-manager.js
@@ -163,7 +163,7 @@ class ConnectionManager extends AbstractConnectionManager {
   }
 
   validate(connection) {
-    return connection && (connection.loggedIn || (connection.state.name === "LoggedIn"));
+    return connection && (connection.loggedIn || connection.state.name === 'LoggedIn');
   }
 }
 
--- a/lib/dialects/postgres/query-generator.js
+++ b/lib/dialects/postgres/query-generator.js
@@ -412,14 +412,18 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
     ].join(' ');
   }
 
-  removeIndexQuery(tableName, indexNameOrAttributes) {
+  removeIndexQuery(tableName, indexNameOrAttributes, options) {
     let indexName = indexNameOrAttributes;
 
     if (typeof indexName !== 'string') {
       indexName = Utils.underscore(`${tableName}_${indexNameOrAttributes.join('_')}`);
     }
 
-    return `DROP INDEX IF EXISTS ${this.quoteIdentifiers(indexName)}`;
+    return [
+      'DROP INDEX',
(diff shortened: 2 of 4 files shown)


shayonj:
Rebased and pushed again.

lib/dialects/mssql/connection-manager.js
```diff
@@ -163,7 +163,7 @@ class ConnectionManager extends AbstractConnectionManager {
   }
 
   validate(connection) {
-    return connection && (connection.loggedIn || (connection.state.name === "LoggedIn"));
+    return connection && (connection.loggedIn || connection.state.name === 'LoggedIn');
```
shayonj, line 166: this is from running the linter, happy to rule this change out. Lmk.

lib/dialects/postgres/query-generator.js
```diff
@@ -412,14 +412,15 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
...
+  removeIndexQuery(tableName, indexNameOrAttributes, options) {
+    const concurrently = options && options.concurrently ? ' CONCURRENTLY ' : ' ';
     let indexName = indexNameOrAttributes;
 
     if (typeof indexName !== 'string') {
       indexName = Utils.underscore(`${tableName}_${indexNameOrAttributes.join('_')}`);
     }
 
-    return `DROP INDEX IF EXISTS ${this.quoteIdentifiers(indexName)}`;
+    return `DROP INDEX${concurrently}IF EXISTS ${this.quoteIdentifiers(indexName)}`;
```
sdepold, line 423: is this more readible?

```
return [
  "DROP INDEX",
  options && options.concurrently && "CONCURRENTLY",
  `IF EXISTS ${this.quoteIdentifiers(indexName)}`
].filter(Boolean).join(' ');
```

sdepold:
looks good to me :)

ephys:
And looks like typings for `removeIndex` already had a `concurrently` option so we're all good on that front too.

LGTM

lib/dialects/postgres/query-generator.js
shayonj: Definitely! Makes sense, i noticed this pattern in other places too. Updated and also rebased.

fncolon (approved):
Overall.. looks good to me!

fncolon:
I think maintainers done with their review.

Issue #13905, opened 2022-01-04
Support adding a NOT VALID to foreign key constraint in Postgres


shayonj:
## Issue Creation Checklist

[x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Feature Description

Postgres has the ability to add `NOT VALID` to a foreign key constraint. When added, it skips the long check on a table. In certain cases this can be a desirable and more reliable way of adding constraints on large production databases. 

It'd be great if there was an easy and nice way to support this using the ORM DSL, instead of having to write raw SQL to accomplish the task.

### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

It'd be great if there was an easy and nice way to support this using the ORM DSL, instead of having to write raw SQL to accomplish the task.

Example: Introducing a `notValid: true` key in the options object. Default being `false`, to preserve existing behavior. 

```js
queryInterface.addConstraint('Posts', {
  fields: ['username'],
  type: 'foreign key',
  name: 'custom_fkey_constraint_name',
  references: { //Required field
    table: 'target_table_name',
    field: 'target_column_name'
  },
  notValid: true,
});```

### Why should this be in Sequelize

Since the query being generated is from sequelize query interface, it perhaps makes sense for this to be in sequelize package.

### Describe alternatives/workarounds you've considered

Writing raw SQL statements. 

### Additional context

This feature is geared more towards giving developers simpler options to write more reliable SQL statements when working on large production databases (albeit with tradeoffs). In similar vein as https://github.com/sequelize/sequelize/issues/13901. Having this would mean that developers can use the DSL over raw SQL, making it a good dev UX.

## Feature Request Checklist

<!-- Please answer the questions below. If you don't, your issue may be closed. -->

### Is this feature dialect-specific?

- [ ] No. This feature is relevant to Sequelize as a whole.
- [x] Yes. This feature only applies to the following dialect(s): Postgres

### Would you be willing to implement this feature by submitting a Pull Request?

<!-- Remember that first contributors are welcome! -->

- [x] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

I am more than happy to work on this and create PR if/when the feature is approved. Thank you!

WikiRik:
[...]

PR #13910, opened 2022-01-05
feat(postgres): support for adding foreign keys with NOT VALID

shayonj:
### Pull Request Checklist

### Description Of Change

This extends `addConstraint` and `changeColumn` to support
adding foreign keys (via `ALTER` statements) with a `NOT VALID`
option. It does it by introducing a new attribute `notValid: true`
on the references object (`ModelAttributeColumnReferencesOptions` type).

This is specific to postgres dialect only. Added unit and integration specs,
and update type definition.

NOTE: This doesn't apply to `addColumn` or `createTable`
since PG doesn't support `NOT VALID` on `REFERENCES`,
which is the same syntax being shared by the engine itsefl. Ex: 
```sql
ALTER TABLE "public"."books" ADD COLUMN "usernamee" varchar(30) REFERENCES "users" ("name") ON DELETE SET NULL ON UPDATE CASCADE;
```

Closes https://github.com/sequelize/sequelize/issues/13905

### DIFF
--- a/lib/dialects/abstract/query-interface.js
+++ b/lib/dialects/abstract/query-interface.js
@@ -466,10 +466,11 @@ class QueryInterface {
   /**
    * Change a column definition
    *
-   * @param {string} tableName          Table name to change from
-   * @param {string} attributeName      Column name
-   * @param {object} dataTypeOrOptions  Attribute definition for new column
-   * @param {object} [options]          Query options
+   * @param {string} tableName                                Table name to change from
+   * @param {string} attributeName                            Column name
+   * @param {object} dataTypeOrOptions                        Attribute definition for new column
+   * @param {object} [dataTypeOrOptions.references.notValid]  Mark the foreign key being added as NOT VALID to skip validation check. PostgreSQL Only.
+   * @param {object} [options]                                Query options
    */
   async changeColumn(tableName, attributeName, dataTypeOrOptions, options) {
     options = options || {};
@@ -480,7 +481,7 @@ class QueryInterface {
       context: 'changeColumn',
       table: tableName,
     });
-    const sql = this.queryGenerator.changeColumnQuery(tableName, query);
+    const sql = this.queryGenerator.changeColumnQuery(tableName, query, dataTypeOrOptions);
 
     return this.sequelize.query(sql, options);
   }
@@ -730,18 +731,19 @@ class QueryInterface {
    *   onUpdate: 'cascade'
    * });
    *
-   * @param {string} tableName                   Table name where you want to add a constraint
-   * @param {object} options                     An object to define the constraint name, type etc
-   * @param {string} options.type                Type of constraint. One of the values in available constraints(case insensitive)
-   * @param {Array}  options.fields              Array of column names to apply the constraint over
-   * @param {string} [options.name]              Name of the constraint. If not specified, sequelize automatically creates a named constraint based on constraint type, table & column names
-   * @param {string} [options.defaultValue]      The value for the default constraint
-   * @param {object} [options.where]             Where clause/expression for the CHECK constraint
-   * @param {object} [options.references]        Object specifying target table, column name to create foreign key constraint
-   * @param {string} [options.references.table]  Target table name
-   * @param {string} [options.references.field]  Target column name
-   * @param {string} [options.references.fields] Target column names for a composite primary key. Must match the order of fields in options.fields.
-   * @param {string} [options.deferrable]        Sets the constraint to be deferred or immediately checked. See Sequelize.Deferrable. PostgreSQL Only
+   * @param {string} tableName                      Table name where you want to add a constraint
+   * @param {object} options                        An object to define the constraint name, type etc
+   * @param {string} options.type                   Type of constraint. One of the values in available constraints(case insensitive)
+   * @param {Array}  options.fields                 Array of column names to apply the constraint over
+   * @param {string} [options.name]                 Name of the constraint. If not specified, sequelize automatically creates a named constraint based on constraint type, table & column names
+   * @param {string} [options.defaultValue]         The value for the default constraint
+   * @param {object} [options.where]                Where clause/expression for the CHECK constraint
+   * @param {object} [options.references]           Object specifying target table, column name to create foreign key constraint
+   * @param {string} [options.references.table]     Target table name
+   * @param {string} [options.references.field]     Target column name
+   * @param {string} [options.references.fields]    Target column names for a composite primary key. Must match the order of fields in options.fields.
+   * @param {object} [options.references.notValid]  Mark the foreign key being added as NOT VALID to skip validation check. PostgreSQL Only.
+   * @param {string} [options.deferrable]           Sets the constraint to be deferred or immediately checked. See Sequelize.Deferrable. PostgreSQL Only
    *
    * @returns {Promise}
    */
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -49,6 +49,7 @@ class PostgresDialect extends AbstractDialect {
     TSVECTOR: true,
     deferrableConstraints: true,
     searchPath: true,
+    notValid: true,
   });
 
   constructor(sequelize) {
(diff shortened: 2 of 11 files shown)


test/integration/query-interface.test.js
```diff
@@ -636,6 +636,56 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+          type: 'foreign key'
+        });
+        let constraints = await this.queryInterface.showConstraint('posts');
+        constraints = constraints.map(constraint => constraint.constraintName);
+
+        if (dialect === 'postgres') {
+          const foreignKeys = await this.sequelize.query(
+            this.queryInterface.queryGenerator.getForeignKeysQuery('posts', this.sequelize.config.database),
+            { type: this.sequelize.QueryTypes.FOREIGNKEYS }
+          );
```
shayonj, line 684: I am using the `condef` (`pg_catalog.pg_get_constraintdef(r.oid, true)`) to ensure that the statement that was run has the `NOT VALID` in it. I'd be curious if there another function I could use, happy to update accordingly.  Let me know.

I noticed `getForeignKeysQuery` is not supported under `getForeignKeyReferencesQuery` so didn't want to update an unrelated function in this change. 

https://github.com/sequelize/sequelize/blob/996e273ceee8ab634d99cf9576e82b76188b6725/lib/dialects/postgres/query-generator.js#L879

test/unit/dialects/postgres/query-generator.test.js
```diff
@@ -303,6 +303,27 @@ if (dialect.startsWith('postgres')) {
             col_2: "ENUM('value 3', 'value 4') NOT NULL"
           }],
           expectation: 'ALTER TABLE "myTable" ALTER COLUMN "col_1" SET NOT NULL;ALTER TABLE "myTable" ALTER COLUMN "col_1" DROP DEFAULT;CREATE TYPE "public"."enum_myTable_col_1" AS ENUM(\'value 1\', \'value 2\');ALTER TABLE "myTable" ALTER COLUMN "col_1" TYPE "public"."enum_myTable_col_1" USING ("col_1"::"public"."enum_myTable_col_1");ALTER TABLE "myTable" ALTER COLUMN "col_2" SET NOT NULL;ALTER TABLE "myTable" ALTER COLUMN "col_2" DROP DEFAULT;CREATE TYPE "public"."enum_myTable_col_2" AS ENUM(\'value 3\', \'value 4\');ALTER TABLE "myTable" ALTER COLUMN "col_2" TYPE "public"."enum_myTable_col_2" USING ("col_2"::"public"."enum_myTable_col_2");'
+        },
+        {
+          arguments: ['myTable', {
+            col_1: "ENUM('value 1', 'value 2') NOT NULL",
+            col_2: "ENUM('value 3', 'value 4') NOT NULL"
+          }, { references: { notValid: true, model: 'level', key: 'id'  } }],
```
shayonj, line 311: this is more of a safety check to ensure that if no `REFERENCES` are mentioned in the definition then `notValid: true` has no impact whatsoever (no `NOT VALID` in the end)

shayonj:
note: rebased, and looks like the failed spec is from `main` (https://github.com/sequelize/sequelize/runs/4822024637?check_suite_focus=true)

ephys:
Yes but don't worry, we're aware of the issue with the DB2 test, will merge anyway :)

lib/dialects/postgres/index.js
```diff
@@ -49,6 +49,7 @@ class PostgresDialect extends AbstractDialect {
     TSVECTOR: true,
     deferrableConstraints: true,
     searchPath: true,
+    notValid: true,
```
ephys, line 52: this needs to be declared in abstract/index.ts, along with a default value.

I think it would be better to group this flag with the other constraint flags (`constraints.notValid`), along with a small comment about its purpose (e.g. supports the "NOT VALID" option on constraints)

lib/dialects/postgres/query-generator.js
```diff
@@ -281,7 +281,7 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
     return `ALTER TABLE ${quotedTableName} DROP COLUMN ${quotedAttributeName};`;
   }
 
-  changeColumnQuery(tableName, attributes) {
+  changeColumnQuery(tableName, attributes, dataTypeOrOptions) {
```
ephys, line 284: this method has no reason to accept a dataType, this should just be `options`

```suggestion
  changeColumnQuery(tableName, attributes, options) {
```

lib/dialects/postgres/query-generator.js
```diff
@@ -317,6 +317,10 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
 
       if (definition.includes('REFERENCES')) {
         definition = definition.replace(/.+?(?=REFERENCES)/, '');
+        if (dataTypeOrOptions.references && dataTypeOrOptions.references.notValid) {
```
ephys, line 320:
```suggestion
        if (dataTypeOrOptions?.references?.notValid) {
```

lib/dialects/postgres/query-generator.js
```diff
@@ -317,6 +317,10 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
...
+          definition = definition.concat(' NOT VALID');
```
ephys, line 321:
```suggestion
          definition += ' NOT VALID';
```

lib/dialects/abstract/query-interface.js
```diff
@@ -480,7 +481,7 @@ class QueryInterface {
       context: 'changeColumn',
       table: tableName,
     });
-    const sql = this.queryGenerator.changeColumnQuery(tableName, query);
```
ephys, line 483: `dataTypeOrOptions` should not be passed if it's a DataType instead of `ModelAttributeColumnOptions`. Checking if it's a plain object (lodash has a isPlainObject function) should be enough

test/integration/query-interface.test.js
```diff
@@ -649,6 +649,56 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+        expect(constraints).to.include('posts_username_users_fk');
+
+        if (dialect === 'postgres') {
+          const foreignKeys = await this.sequelize.query(
+            this.queryInterface.queryGenerator.getForeignKeysQuery('posts', this.sequelize.config.database),
+            { type: this.sequelize.QueryTypes.FOREIGNKEYS },
+          );
+          expect(Object.keys(foreignKeys[0])).to.have.length(8);
+          expect(foreignKeys[0].condef).to.equal('FOREIGN KEY (username) REFERENCES users(username) ON UPDATE CASCADE ON DELETE CASCADE');
+        }
```
ephys, line 661: I don't think this change is necessary. The unit test already checks that the SQL is generated without `NOT VALID`

test/integration/query-interface/changeColumn.test.js
```diff
@@ -181,6 +181,47 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
           expect(newForeignKeys[0].columnName).to.be.equal('level_id');
+
+          if (dialect === 'postgres') {
+            const updatedForeignKeys = await this.sequelize.query(
+              this.queryInterface.queryGenerator.getForeignKeysQuery('users', this.sequelize.config.database),
+              { type: this.sequelize.QueryTypes.FOREIGNKEYS },
+            );
+            expect(Object.keys(updatedForeignKeys[0])).to.have.length(8);
+            expect(updatedForeignKeys[0].condef).to.equal('FOREIGN KEY (level_id) REFERENCES level(id) ON UPDATE CASCADE ON DELETE CASCADE');
+          }
```
ephys, line 192: same idea as previous comment here.

I think a more interesting series of test for `changeColumn` would be:

- column has `NOT VALID`, and `changeColumn` has `notValid` set to false (removes it)
- column does not have `NOT VALID`, and `changeColumn` has `notValid` set to true (adds it)
- column has `NOT VALID`, and `changeColumn` has `notValid` not set or undefined (leaves it unchanged, not removed)

ephys (changes requested):
Thank you for the PR. I'm going through our PR list to see if there's anything that hasn't been reviewed and it looks like yours was missed. very sorry about that.

I have a few comments but overall this PR looks really great! :)

PR #13917, opened 2022-01-07
test(mysql8): actually disable flaky mysql8 test

WikiRik:
So the `current.options.databaseVersion` is not filled until the beforeEach so can only be found inside a test. So this would be a proper fix to #13870

### DIFF
--- a/test/integration/associations/scope.test.js
+++ b/test/integration/associations/scope.test.js
@@ -264,11 +264,11 @@ describe(Support.getTestDialectTeaser('associations'), () => {
           include: [{ model: this.Comment, as: 'comments' }]
         });
         this.post = post;
-        for (const comment of  post.comments) {
+        for (const comment of post.comments) {
           expect(comment.get('commentable')).to.equal('post');
         }
         post = await this.Post.scope('withComments').findByPk(this.post.id);
-        for (const comment of  post.comments) {
+        for (const comment of post.comments) {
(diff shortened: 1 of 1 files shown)


test/integration/associations/scope.test.js
```diff
@@ -323,105 +323,110 @@ describe(Support.getTestDialectTeaser('associations'), () => {
...
-              expect(postB.get('categories')[0].get('type')).to.equal('category');
-              expect(postB.get('tags')[0].get('type')).to.equal('tag');
-              expect(postC.get('categories')[0].get('type')).to.equal('category');
-              expect(postC.get('tags')[0].get('type')).to.equal('tag');
-            });
-          }
-        });
+          it('should create, find and include associations with scope values', async function() {
+            // We don't know the databaseVersion outside of the tests
+            const isMySQL8 = Support.getTestDialect() === 'mysql' && semver.satisfies(current.options.databaseVersion, '>=8.0.0');
```
ephys, line 328: my only request would be to document why this test is disabled, for future maintainers/us :)

ephys (approved):

test/integration/associations/scope.test.js
WikiRik: That was mentioned later in the test I believe. I'll make an issue to debug it and re-enable the test later

ephys: Oh yes you're right I remember seeing that last time

PR #13922, opened 2022-01-08
refactor: migrate Transaction, Enums, & AbstractDialect

ephys:
### Description Of Change

*Don't merge this in v6. Includes breaking changes*

Migrated `transaction.ts`, `table-hints.ts`, `query-types.ts`, `index-hints.ts`, & `dialects/abstract/index.ts`.

I also removed the hack in `operators.ts`

Note: I opted not to use any default exports in my refactor. This way, code that has not been migrated yet can be changed to
`const { Op } = require('./operators');` instead of the more verbose `const { default: Op } = require('./operators');`  
It's also easier to do `export * from` when we don't use default exports :)

Lots of changed files due to the export style change.

More important note: The typings for `ConnectionManager#getConnection` declared accepting `{ type: 'read' | 'write' }` as a parameter. It in fact accepted 'SELECT' instead of 'read'. I've changed it to 'read' to match typings. This is a breaking change for people using ConnectionManager directly.

### Todos

- [x] See which dialect fails


WikiRik:
Could you document the breaking changes in the migration docs as well? https://github.com/sequelize/sequelize/blob/main/docs/manual/other-topics/upgrade-to-v7.md

Preferably some examples of the old behaviour and new behaviour so people see how to adopt it.

ephys:
Done! :)

ephys:
Before we merge this I'm still a bit confused when it comes to which branch needs to be targetted.

This should not land on v6, only on v7. Does that mean I should target the v7 branch or continue with `main`?

lib/dialects/abstract/index.ts
```diff
@@ -0,0 +1,142 @@
...
+export class AbstractDialect {
+  // the actual field is on the prototype,
+  //  hence the `declare` keyword
+  declare supports: DialectSupports;
+
+  // TODO: Replace with QueryGenerator class once its typings are complete.
+  declare queryGenerator: unknown;
+}
+
+AbstractDialect.prototype.supports = {
```
sdepold, line 77: Any reason why we don't inline this?

```
export class AbstractDialect {
  const supports: DialectSupports = { /* stuff here */ } 
}
```

lib/index-hints.ts
```diff
@@ -1,14 +1,13 @@
...
  * @property USE
  * @property FORCE
  * @property IGNORE
  */
-const IndexHints = module.exports = { // eslint-disable-line
-  USE: 'USE',
-  FORCE: 'FORCE',
-  IGNORE: 'IGNORE'
-};
+// eslint-disable-next-line no-shadow
```
sdepold, line 8: What are we shadowing?

sdepold:
(I hope) We have agreed on:

- `main` to be the primary development branch for v7
- Any change to be introduced to v6 is either pointing to the master branch and cherry-picked to the v6 branch manually. Tracking happens by adding those changes to the v6 release project
- Any problem that only exists in the v6 code base because we have diverged too much at some point, can be fixed via a change pointing at v6 directly (which will immediately trigger a new release).

sdepold (dismissed):
Amazing stuff!

lib/index-hints.ts
ephys: I have no idea! I think it's no-shadow being broken with typescript. Should be fixed once we install `@typescript/no-shadow` instead

lib/dialects/abstract/index.ts
ephys: I copied this part over but my guess is the initial idea was to be able to access this object without instantiating the Dialect class.

sdepold: What makes more sense from your pov?

lib/index-hints.ts
sdepold: Maybe just remove it?

lib/dialects/abstract/index.ts
ephys: I think adding a field to a prototype is prone to errors. I'd prefer assigning a frozen object to the constructor & the instance.

Let's find out if anything breaks if we do this.

lib/index-hints.ts
ephys: I'd prefer to keep it for now. I'll try to see about writing a better eslint config & fix all issues in that PR :)

Speaking of which [I have a preset for eslint](https://github.com/ephys/eslint-config) that I've developed over the years initially based on airbnb. Could be a good starting point?

lib/dialects/abstract/index.ts
ephys: Done. `AbstractDialect.prototype.supports` still exists but it's now a getter that returns the static field

lib/index-hints.ts
sdepold: I'm fine with any config really :) as long as my editor understands how to format the file for me.

sdepold: I think we should format the entire code base as a one off thing once we settled on a config. Wanna file a pr?

ephys: Sure :)

WikiRik: I believe we wanted to wait until the TypeScript migration was done so we wouldn't have to keep any specific rules for vanilla JS and we could do a proper config at once. But we can try out the preset of @ephys if it doesn't give many issues with the current JS codebase

lib/dialects/abstract/index.ts
```diff
@@ -0,0 +1,143 @@
...
+  'DEFAULT VALUES': boolean,
+  'VALUES ()': boolean,
+  'LIMIT ON UPDATE': boolean,
+  'ON DUPLICATE KEY': boolean,
+  'ORDER NULLS': boolean,
+  'UNION': boolean,
+  'UNION ALL': boolean,
+  'RIGHT JOIN': boolean,
+
+  /* does the dialect support returning values for inserted/updated fields */
```
WikiRik, line 12: Do we need the comments on both parts? I would only keep these in the implementations

lib/dialects/db2/index.js
```diff
@@ -1,14 +1,47 @@
...
-const AbstractDialect = require('../abstract');
+const { AbstractDialect } = require('../abstract');
 const ConnectionManager = require('./connection-manager');
 const Query = require('./query');
 const QueryGenerator = require('./query-generator');
 const DataTypes = require('../../data-types').db2;
 const { Db2QueryInterface } = require('./query-interface');
 
 class Db2Dialect extends AbstractDialect {
+  static supports = _.merge(_.cloneDeep(AbstractDialect.supports), {
```
WikiRik, line 12: Can we only set the ones that are different from AbstractDialect here?

WikiRik: Also counts for the other dialects of course

lib/dialects/mysql/index.js
```diff
@@ -1,16 +1,50 @@
...
+      GEOMETRY: true,
+      JSON: true,
+      REGEXP: true,
+      _test_: 2
+    }
+  );
+
   constructor(sequelize) {
     super();
+
```
WikiRik, line 47: I don't think this empty line is consistent with the other dialects

lib/index-hints.ts
ephys: Well, only 12290 errors!

WikiRik (changes requested):
Few small remarks, but overall looks good. Nice work!

lib/dialects/abstract/index.ts
WikiRik: I was about to say, we should refactor the usages of protoype out as much as possible in later PRs.

lib/index-hints.ts
WikiRik: After running it with the --fix flag or before?

ephys: There are some rules I'll disable for the JS codebase, but keep enabled for TS. This should ease the transition.

ephys: Down to 900 errors. There are a lot of autofixed changes, which will be annoying for existing PRs. It's designed to minimise diff noise in the future so I think it's still worth it.

ephys: 578 errors left but these are actual code smells that should be properly fixed (like using `==`, `new Array`, etc)

sdepold: Hope you'll do that in another pr 😅

ephys: Yeah :)

lib/dialects/mysql/index.js
ephys: Removed!

lib/dialects/db2/index.js
ephys: Done. And updated the typings accordingly

lib/dialects/abstract/index.ts
ephys: I kept them in the typing only so they're shared by all implementations

lib/dialects/mssql/index.js
```diff
@@ -1,32 +1,15 @@
...
-      sequelize,
-      this.queryGenerator
-    );
-  }
-}
-
-MssqlDialect.prototype.supports = _.merge(
-  _.cloneDeep(AbstractDialect.prototype.supports),
-  {
+  static supports2 = _.merge(_.cloneDeep(AbstractDialect.supports), {
```
WikiRik, line 12: I think this can be removed right?

ephys: 🤦‍♀️

I added that to ensure both versions were deep equal, removing

WikiRik (dismissed):

ephys:
I'd like to see about merging https://github.com/sequelize/sequelize/pull/13930 first so I can lint these new files too

ephys:
The new linter detected an issue in Transaction.ts: `cleanup` was not awaited in `commit` or `rollback`.

I don't know if this was voluntary but it could lead to an unhandled promise rejection error.

Also after commit hooks are awaited sequentially (https://github.com/sequelize/sequelize/pull/13922/files#diff-08c4998b82afd148218df2f2e7ce65983d0068469af7236efe0d15126f942269R83). I don't know if this is on purpose but I opted to not change it.

lib/index-hints.ts
ephys: Shadowing resolved!

WikiRik (dismissed):
One small suggestion, but the changes because of the updated linting still look good

WikiRik (approved):

PR #13935, opened 2022-01-10
build(deps): update yarn.lock file

WikiRik:
~We should be able to remove this within a week or so. Not entirely sure why there are so many changes to yarn.lock but I removed the version in main and the node_modules folder and reinstalled everything~

~https://snyk.io/blog/open-source-npm-packages-colors-faker/~

EDIT: it seems that by updating the yarn.lock file we already got the versions of dependencies that pin colors so that's the only thing needed. We might want to look into why there are so many changes for the file even with renovate. Maybe we can find a way to automatically update the entire yarn.lock file every week or so since we use it in a locked state for our CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,9 +141,6 @@
       "optional": true
     }
   },
-  "resolution": {
-    "colors": "1.4.0"
-  },
   "keywords": [
     "mysql",
     "mariadb",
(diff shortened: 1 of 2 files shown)


ephys (approved):
Just updating was enough because npm removed the colors release. Looks like tests are passing

PR #13936, opened 2022-01-10
build: allow db2 tests to fail

WikiRik:
I noticed that DB2 tests were failing but CI still let them pass, this should solve that


ephys (approved):
That's weird. Is there a reason why db2 was allowed to fail?

sdepold:
I guess we did that because it was considered experimental. It seems semi stable however?

WikiRik:
> I guess we did that because it was considered experimental. It seems semi stable however?

They are two flaky tests for DB2 (especially for Node 12) but in general it seems to be running fine. At least we can see that some tests have failed. Maintainers can still merge even if DB2 tests fail because they are not required so there is no big harm in this change. But I'll leave it up to you @sdepold to merge

sdepold:
Aaaand it is optional for the release step right?

fncolon (approved):
I thought it's just worked like p-retry when we scrape something and got cloudflare in front of us but it turns out it's not.

LGTM.

ephys:
I think I get why the tests were allowed to fail now 😅

sdepold:
> I think I get why the tests were allowed to fail now 😅

😅

Issue #8310, opened 2017-09-15
Invalid JSON if using bulkInsert for JSON columns


borislemke:
## What are you doing?
In my `customer-seeder.js`:
```js
export async function up (queryInterface) {
  const customers = [
    {
      first_name: "Boris",
      enabled_payments: ["credit_card", "bank_transfer"] // this is a JSON column
    }
  ]

  await queryInterface.bulkInsert('customers', customers, {})
}
```

## What do you expect to happen?
Should parse the object to JSON

## What is actually happening?
Throws
```
Seed file failed with error: Invalid JSON text
```

A workaround is to manually `stringify` the value to be inserted:
```
...
  const customers = [
    {
      first_name: "Boris",
      enabled_payments: JSON.stringify(["credit_card", "bank_transfer"])
    }
  ]
...
```

This only happens when seeding, when using the Model's instance, it works perfectly:
```
// CustomerModel is an instance of sequelize.Model
CustomerModel.create({
  first_name: "Boris",
  enabled_payments: ["credit_card", "bank_transfer"] // this will be parsed as JSON
})
```

_Output, either JSON or SQL_

__Dialect:__ mysql
__Dialect version:__ 1.4.2
__Database version:__ 5.7.19
__Sequelize version:__ 4.8.0
__Tested with master branch:__ No

**Note :** _Your issue may be ignored by maintainers if its not tested against `master` branch OR does not follow issue template._

Julyyq:
+1

andrew-aladev:
Same issue for me. I wanted to have array of JSON objects. I had to make an ugly workaround.

```
const proto = Sequelize.DataTypes.ARRAY.prototype;
const stringifyOrig = proto.stringify;

proto.stringify = function (values) {
  if (values._arrayWithJSON) {
    values = values.map((value) => {
      return `'${JSON.stringify(value)}'::json`;
    });
  }

  return `array[${values.join(",")}]`;
};

datas.forEach((data) => {
  if (data.smth) {
    data.smth._isArrayWithJSON = true;
  }
});

return queryInterface.bulkInsert("datas", datas)
  .finally(() => {
    proto.stringify = stringifyOrig;
  });
```

andrew-aladev:
`lib/sql-string.js`:

```
if (Array.isArray(val)) {
  const partialEscape = _.partial(escape, _, timeZone, dialect, format);
  if (dialect === 'postgres' && !format) {
[...]

PR #13945, opened 2022-01-13
fix(types): fix QueryInterface#bulkInsert attribute arg type

ChristopherChudzicki:
### Pull Request Checklist

- [x] (N/A) Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description Of Change

This is a very small PR to improve the typings to `bulkInsert` and the `JSON`/`JSONB` data types.

**Motivation:** I have a typescript seeders file that is seeding some JSONB columns in a postgres database. The seed file uses `queryInterface.bulkInsert` to insert a bunch of rows read from a file. But since `queryInterface` does not know the model column types, it does know to convert a few specific columns to JSONB. It was suggested in https://github.com/sequelize/sequelize/issues/8310#issuecomment-430287656 to pass `attributes` as a 4th arg to `bulkInsert`. But the existing types for `attributes` seems to be incorrect. (Additionally I discovered `new Sequelize.JSONB()` was throwing a "not constructable" type error.)

*PS: I'm pretty new to ORMs. Sequelize has been great so far. That's for it!*

### DIFF
--- a/types/test/data-types.ts
+++ b/types/test/data-types.ts
@@ -1,7 +1,7 @@
 import { expectTypeOf } from 'expect-type';
 import { DataTypes } from 'sequelize';
 
-const { TINYINT, SMALLINT, MEDIUMINT, BIGINT, INTEGER } = DataTypes;
+const { TINYINT, SMALLINT, MEDIUMINT, BIGINT, INTEGER, JSON, JSONB } = DataTypes;
 
 // TINYINT
 expectTypeOf(TINYINT()).toEqualTypeOf<DataTypes.TinyIntegerDataType>();
@@ -32,3 +32,11 @@ expectTypeOf(INTEGER()).toEqualTypeOf<DataTypes.IntegerDataType>();
 expectTypeOf(new INTEGER()).toEqualTypeOf<DataTypes.IntegerDataType>();
 expectTypeOf(INTEGER.UNSIGNED.ZEROFILL()).toEqualTypeOf<DataTypes.IntegerDataType>();
 expectTypeOf(new INTEGER.UNSIGNED.ZEROFILL()).toEqualTypeOf<DataTypes.IntegerDataType>();
+
+// JSON
+expectTypeOf(new JSON()).toEqualTypeOf<DataTypes.AbstractDataType>();
+expectTypeOf(JSON()).toEqualTypeOf<DataTypes.AbstractDataType>();
+
+// JSONB
+expectTypeOf(new JSONB()).toEqualTypeOf<DataTypes.AbstractDataType>();
+expectTypeOf(JSONB()).toEqualTypeOf<DataTypes.AbstractDataType>();
--- a/types/test/data-types.ts
+++ b/types/test/data-types.ts
@@ -1,7 +1,7 @@
 import { expectTypeOf } from 'expect-type';
 import { DataTypes } from 'sequelize';
 
-const { TINYINT, SMALLINT, MEDIUMINT, BIGINT, INTEGER, JSON, JSONB } = DataTypes;
+const { TINYINT, SMALLINT, MEDIUMINT, BIGINT, INTEGER, JSON, JSONB, CITEXT, MACADDR, TSVECTOR, CIDR, INET } = DataTypes;
 
 // TINYINT
 expectTypeOf(TINYINT()).toEqualTypeOf<DataTypes.TinyIntegerDataType>();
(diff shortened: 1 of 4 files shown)


ChristopherChudzicki:
Both DB2 CI tests are failing with 
```
A communication error has been detected. Communication protocol being used: "TCP/IP".  Communication API being used: "SOCKETS".  Location where the error was detected: "127.0.0.1".  Communication function detecting the error: "selectForConnectTimeout".  Protocol specific error code(s): "115", "*", "*".  SQLSTATE=08001
```
https://github.com/sequelize/sequelize/runs/4798649046?check_suite_focus=true#step:7:2122 

unsure what that is :/

ephys:
Don't worry about the DB2 tests, we'll merge anyway :)

ephys (changes requested):
It's looking good. All that's missing is a type-test in the `types/test` folder (basically just use the method with the right parameters).

Also, is your issue with `AbstractDataTypeConstructor` resolved if we change its definition from

```typescript
interface AbstractDataTypeConstructor {
  key: string;
  warn(link: string, text: string): void;
}
```

to 

```typescript
interface AbstractDataTypeConstructor extends Function {
  key: string;
  warn(link: string, text: string): void;
}
```

instead of adding the constructor? Because that's a change I was planning on doing

ChristopherChudzicki:
@ephys Thanks for the quick review! I'll take a look at the `types/test`.

I don't think that making `interface AbstractDataTypeConstructor extends Function {...}` is sufficient. (See example: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgIICMDOYqLAETjDgBUBPABwgGEB7EbKAVwTFqmQgA9IQATTMgBiTEK2D1kAbwBQyZAGsIZAFzJGoAOYBuOcgDucKCAAUAG1AK1GkJoA0ySD2s4tASjUA3WsD66AvjIyCPTY6nAgwGBk1AAWEAgKyAC8yCYAyrQAthCExORUahiMeHmklDShOCxsUG4pAHzSegD0LY6YelAQYEzGyCAQ+siZOWUFECZuAdpAA ). My understanding is that `interface X extends Y {..}` means that `X` has methods/properties of *instances* of `Y`. Since `Function` instances don't have a `new` method, the `AbstractDataTypeConstructor` won't either.

### Adding `new` to `AbstractDataTypeConstructor`
This might be a better option than what I did in the PR:
```ts
interface AbstractDataTypeConstructor {
  key: string;
  warn(link: string, text: string): void;
  new (): AbstractDataType;
}
```
then:
1.  these would all become constructable:
    ```
    export const JSON: AbstractDataTypeConstructor;
    export const JSONB: AbstractDataTypeConstructor;
    export const CIDR: AbstractDataTypeConstructor;
    export const INET: AbstractDataTypeConstructor;
    export const MACADDR: AbstractDataTypeConstructor;
    export const CITEXT: AbstractDataTypeConstructor;
    export const TSVECTOR: AbstractDataTypeConstructor;
    ```
    This seems desirable?
2. `DateDataType` would become constructable, too:
    ```
    interface DateDataTypeConstructor extends AbstractDataTypeConstructor { /* omitted from this comment* /}
    export interface DateDataType extends AbstractDataTypeConstructor {
      options: DateDataTypeOptions;
    }
    ```
    But it seems like maybe there's a typo here and `DateDataType` should be extending `AbstractDataType`?

I was trying to keep my changes a little more isolated, but if you have any thoughts on the above approach, it might be better.

Looking around issues/PRs, I see sequelize is moving to convert more of the repo to Typescript, so hopefully these manual `d.ts` typings won't be permanent...

ephys:
Thank you for the tests!

> I don't think that making `interface AbstractDataTypeConstructor extends Function {...}`

You're right, let's continue with your solution

> Adding new to AbstractDataTypeConstructor

That means `new DataTypes.JSON` will return `AbstractDataType` instead of `JSONDataType` but given it's all typing already anyway (you can't use instanceof on AbstractDataType), it doesn't really matter. 
Go for it!

> But it seems like maybe there's a typo here and DateDataType should be extending AbstractDataType?

Definitely looks like a mistake, would you mind adding it to your changes?

> I was trying to keep my changes a little more isolated, but if you have any thoughts on the above approach, it might be better.

I'm fine with including a small change like this one

> Looking around issues/PRs, I see sequelize is moving to convert more of the repo to Typescript, so hopefully these manual d.ts typings won't be permanent...

Yes! :) Hopefully we'll be done with this soon enough
There's already some work in progress to migrate DataTypes https://github.com/sequelize/sequelize/pull/13799

> PS: Also, should I rebase or squash the commits or anything?

No need, unless you've carefully crafted your commits & request to not squash, we squash and merge PRs :)

types/test/data-types.ts
```diff
@@ -32,3 +32,11 @@ expectTypeOf(INTEGER()).toEqualTypeOf<DataTypes.IntegerDataType>();
...
 expectTypeOf(INTEGER.UNSIGNED.ZEROFILL()).toEqualTypeOf<DataTypes.IntegerDataType>();
 expectTypeOf(new INTEGER.UNSIGNED.ZEROFILL()).toEqualTypeOf<DataTypes.IntegerDataType>();
+
+// JSON
+expectTypeOf(new JSON()).toEqualTypeOf<DataTypes.AbstractDataType>();
+expectTypeOf(JSON()).toEqualTypeOf<DataTypes.AbstractDataType>();
+
+// JSONB
+expectTypeOf(new JSONB()).toEqualTypeOf<DataTypes.AbstractDataType>();
+expectTypeOf(JSONB()).toEqualTypeOf<DataTypes.AbstractDataType>();
```
WikiRik, line 42: Can you make sure that there is an empty line at the end of this file?

ChristopherChudzicki: Oops, apparently I failed at the new-line even after review. Looking into it... I have the newline locally, but it seems not to be showing up on github, which is confusing me. Looking into it....

ephys (approved):
Looks good to me :)

WikiRik (approved):
Yes, just needs a better title and then we can merge. @ephys can you take care of that and add this to the v6 project?

ephys:
Thank you @ChristopherChudzicki :)

PR #13951, opened 2022-01-15
fix: don't call overloaded versions of find functions internally

cincodenada:
### Pull Request Checklist

- [N/A] Did you update the typescript typings accordingly (if applicable)?
- [N/A] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?

### Description Of Change

Comments in `model.js` indicated that we should avoid calling overloaded versions of `findAll` and `findOne` internally, but we were in fact doing just that. It appears this behavior was lost in the transition to ES6 classes in #5924. This fixes the code behavior to match the comment (and previous behavior); if this is no longer desired the comments could just be deleted instead.


ephys (approved):
Looks good! I'm not opposed to letting people override Model static methods but I think it's safer not to.

Thanks for the test file name refactor, we're slowly refactoring to a more consistent codebase :)

cincodenada:
> Looks good! I'm not opposed to letting people override Model static methods but I think it's safer not to.
> 
> Thanks for the test file name refactor, we're slowly refactoring to a more consistent codebase :)

Thanks! And yeah, little bits here and there add up over time, figured I'd do my part :slightly_smiling_face: 

And my understanding is that folks can still override the static methods if they want (I don't think we can really prevent it) but this makes sure if they do, we at least use the internal versions for internal calls where the callers depend on them behaving in a predictable way. Which should make it harder to break things by overriding them, at least! :smile: 

Thanks for looking this over and getting it merged (and fixing up my PR title)

fncolon:
I was wrong about the commit type, i thought it was all test related, but I think it's better with `fix`, thanks @ephys

Issue #13952, opened 2022-01-15
Husky throws away my commit message if it doesn't like it


cincodenada:
## Issue Creation Checklist

[x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description

I'm all for good commit messages, and in fact consider myself a champion of them. But if I write a commit message that `commitlint` doesn't like, Husky gives me an error and then _throws away my commit message_, with no guidance about how to proceed, other than the validation errors. This is extremely frustrating!

As someone who writes detailed, useful commit messages using an editor, it's a very offputting experience (and very much against the spirit of git) to write a nice useful paragraph in the commit body, and then have it vanish into the ether because I didn't prefix my subject correctly or some such. This is doubly so because this is most likely to happen to new contributors who are not deeply familiar with git and don't know where their commit message went!

This may better be filed as an upstream bug with Husky (which I also plan to do also), but if a project is implementing aggressive commit linting around particular commit formatting (which, properly implemented, is fine!), I think it's incumbent on them to do it in a way that isn't destructive. I'm all for enforcing a correct style, just not for throwing away someone's commit message because they made a typo.

After some research I found that the commit message isn't actually gone, and if you don't make another commit you can recover it [with some cumbersome commands](https://salferrarello.com/recover-failed-git-commit-message/), but I doubt most git users are aware it is even a possibility, much less how to do it. This project should offer a better way.

### What do you expect to happen?

I wanted to be able to edit my paragraph-long commit message to conform to the guidelines

### What is actually happening?

My commit message was discarded with no apparent way to recover it

### Would you be willing to resolve this issue by submitting a Pull Request?

Did it, #13953 - it ain't pretty, but it's better than what we've got.

ephys:
Thank you :) Let us know if husky ever implements it upstream too

cincodenada:
> Thank you :) Let us know if husky ever implements it upstream too

Will do! I still need to figure out where best some mitigation might fit upstream - it may be better placed with commitlint - but I do plan to see what I can get in place up there. Thanks for the quick merge on this end in the meantime!

PR #13953, opened 2022-01-15
fix: don't discard commit message on lint fail

cincodenada:
### Pull Request Checklist

### Description Of Change

Per #13952, If we fail a commit message, we basically throw it away, which is
frustrating if the commit message was more than trivial, which good
commit messages often are.

Instead, offer the developer a command to recover and re-edit their
commit message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -268,6 +268,7 @@
     "test-mssql": "cross-env DIALECT=mssql npm test",
     "test-db2": "cross-env DIALECT=db2 npm test",
     "----------------------------------------- development ---------------------------------------------": "",
+    "fix-commit": "export TMPFILE=$(mktemp) && grep -v '^#' $(git rev-parse --git-dir)/COMMIT_EDITMSG > $TMPFILE && git commit -e -F $TMPFILE",
     "sscce": "node sscce.js",
     "sscce-mariadb": "cross-env DIALECT=mariadb node sscce.js",
     "sscce-mysql": "cross-env DIALECT=mysql node sscce.js",
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,20 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn commitlint --edit $1
+BOLD_RED="\033[1;31m"
+BOLD_WHITE="\033[1;37m"
+YELLOW="\033[33m"
+WHITE="\033[37m"
+RESET="\033[0m"
+
+exitCode=0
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -268,6 +268,7 @@
     "test-mssql": "cross-env DIALECT=mssql npm test",
     "test-db2": "cross-env DIALECT=db2 npm test",
     "----------------------------------------- development ---------------------------------------------": "",
+    "fix-commit": "export TMPFILE=$(mktemp) && grep -v '^#' $(git rev-parse --git-dir)/COMMIT_EDITMSG > $TMPFILE && git commit -e -F $TMPFILE",
```
cincodenada, line 271: The `grep -v` is because otherwise you end up with two commit templates one after the other, which is not the end of the world, but also not great. Better to avoid it.

The tempfile was necessary because not all editors support just piping STDIN into them without any other flags, and some (e.g. vim) horribly break the terminal output when you do, at least on my machine.

fncolon:
This is much help for development experience, well done!

fncolon (approved):

WikiRik:
Thanks for the PR! Although it was already merged I was wondering if you also got a chance to test this on various OS like Windows, Linux and Mac?

ephys:
Note that on windows, it's fine by me if it only works in WSL

fncolon:
> Thanks for the PR! Although it was already merged I was wondering if you also got a chance to test this on various OS like Windows, Linux and Mac?

I think Windows should be created for different PRs, we don't actually know whether the author using Windows or not. So let Windows developer help to reach about it.

fncolon:
> Note that on windows, it's fine by me if it only works in WSL

Yup currently modern Windows support WSL, I think we will let Windows user using WSL for now, I will think if I can help for adding Windows support.

cincodenada:
I tried to keep things to a pretty common denominator - I've tested on both Linux and Mac OS and it works as expected. I don't have ready access to a WSL machine, but I don't see any reason it woudn't work on that, or even on Cygwin.

The only requirements are `grep -v`, which is POSIX, and `mktemp`, which isn't technically POSIX but might as well be, it's pretty ubiquitous. And the additions to the commit hook itself are bog-standard shell script that should run anywhere - Husky runs the commit hooks under `sh` in the first place anyway.

WikiRik:
Yeah, WSL should just work and since husky executes it I think we should be good there as well. Was just wondering where you tested it on. Thanks again for your work!

ephys:
True, should work fine. Thanks! :)

PR #13909, opened 2022-01-05
feat(types): add `InferAttributes` utility type

ephys:
### Pull Request Checklist

- [x] Have you added new tests to prevent regressions?
- [x] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you update the typescript typings accordingly (if applicable)?
- [ ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)?

### Description Of Change

This PR introduces a few new utility types:
- `InferAttributes<M, opts>`: Used to drastically reduce the boilerplate needed to define Model Attributes in the class-based approach.
- `InferCreationAttributes<M, opts>`: Like `InferAttributes` but fields tagged with `CreationOptional<x>` are optional.
- `CreationOptional<x>`: A branded type used to mark an attribute as optional in Creation Attributes.
- `NonAttribute<x>`: A branded type used to omit a property from the attribute list.
- `Attributes<M>`: An alias for `M['_attributes']` for public use. Also forward compatible with planned v7 changes.
- `CreationAttributes<M>`: Same as `Attributes` but types that accept `undefined` are also `optional`.

To reduce risk of collision between `Model` and its subclasses, `Model#_attributes` and `Model#_creationAttributes` have been deprecated (I'd like to replace them with symbols in v7). Use `Attributes<Model>` and `CreationAttributes<Model>` instead.

In v7, we should replace these two with non-exported unique symbols.

PR #13954, opened 2022-01-15
feat(types): drop TypeScript < 4.1

ephys:
### Pull Request Checklist

- [no, the opposite actually] Have you added new tests to prevent regressions?

Follow-up to https://github.com/sequelize/meetings/issues/6#issuecomment-1013081721
This PR drops support for TypeScript < 4.1 and documents our TS support policy.

This release should cause a SemVer MINOR bump.

### DIFF
--- a/docs/manual/other-topics/typescript.md
+++ b/docs/manual/other-topics/typescript.md
@@ -1,6 +1,9 @@
 # TypeScript
 
-Since v5, Sequelize provides its own TypeScript definitions. Please note that only TS >= 3.1 is supported.
+Sequelize provides its own TypeScript definitions.
+
+Please note that only **TypeScript >= 4.1** is supported.
+Our TypeScript support does not follow SemVer. We will support TypeScript releases for one year, after which they may be dropped in a SemVer MINOR release.
 
 As Sequelize heavily relies on runtime property assignments, TypeScript won't be very useful out of the box. A decent amount of manual type declarations are needed to make models workable.
 
--- a/docs/manual/other-topics/typescript.md
+++ b/docs/manual/other-topics/typescript.md
@@ -3,7 +3,7 @@
 Sequelize provides its own TypeScript definitions.
 
 Please note that only **TypeScript >= 4.1** is supported.
-Our TypeScript support does not follow SemVer. We will support TypeScript releases for one year, after which they may be dropped in a SemVer MINOR release.
+Our TypeScript support does not follow SemVer. We will support TypeScript releases for at least one year, after which they may be dropped in a SemVer MINOR release.
 
 As Sequelize heavily relies on runtime property assignments, TypeScript won't be very useful out of the box. A decent amount of manual type declarations are needed to make models workable.
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ts-version: ["3.9", "4.0", "4.1", "4.2", "4.3", "4.4", "4.5"]
+        ts-version: ["4.1", "4.2", "4.3", "4.4", "4.5"]
     name: TS Typings (${{ matrix.ts-version }})
     runs-on: ubuntu-latest
     needs: lint


docs/manual/other-topics/typescript.md
```diff
@@ -1,6 +1,9 @@
 # TypeScript
 
-Since v5, Sequelize provides its own TypeScript definitions. Please note that only TS >= 3.1 is supported.
+Sequelize provides its own TypeScript definitions.
+
+Please note that only **TypeScript >= 4.1** is supported.
+Our TypeScript support does not follow SemVer. We will support TypeScript releases for one year, after which they may be dropped in a SemVer MINOR release.
```
WikiRik, line 6:
```suggestion
Our TypeScript support does not follow SemVer. We will support TypeScript releases for at least one year, after which they may be dropped in a SemVer MINOR release.
```

WikiRik (commented):
Small suggestion for additional clarity

WikiRik:
Also, do we really want this in v6?

ephys:
This in v6 would allow us to merge https://github.com/sequelize/sequelize/pull/13909 in v6 which would provide an easier migration path to the model changes I'd like to make in v7 (simplying model definition to `class User extends Model<User> {}`).

But I am also of two minds about it

fncolon:
I think it will be backward incompatible for some users right? how about adding it to v7 instead?

ephys:
It will be incompatible for users that haven't updated TypeScript since Nov. 19 2020 yes. Which is why I'm of two minds.

On one hand, adding this to v6 would allow us to land https://github.com/sequelize/sequelize/pull/13909 in v6 which would be a great improvement for TS users & make migration to v7 easier. And 4.1 is more than one year old.
I also believe we should not make our TS changes SemVer major (hence the documentation update to reflect our policy).

On the other hand, `tsc` will report an error for users that are still on < 4.1.

But honestly, I have typings errors all the time after upgrades. Even Sequelize adding proper types for `Model#toJSON` completely broke my typings, so we're (and others) are already causing typing breaking changes all the time in minor versions, or even patch. It's just too difficult to avoid. TypeScript itself doesn't follow SemVer (https://github.com/sequelize/meetings/issues/6#issuecomment-1013649742)

Honestly my only problem with this change is that we didn't document it earlier. But I think given it's been more than a year since 4.1, it's fine. Users that haven't updated TS since then are not going to update Sequelize either.

WikiRik:
Yeah, it's more about that we have a policy now relatively close to a new major release. But I agree that with the changes we made to the typings many people that have updated to the latest version of sequelize also have a relatively new version of TS as well.

If migrating to v7 is made easier with this change I think it's fine if we do this in v6

WikiRik (approved):

fncolon:
If this makes our work easier for migrate typings, approve!

ephys:
It won't make our job easier but it will allow users to massively reduce the boilerplate they need to write for TS support in v6

Current way: https://github.com/sequelize/sequelize/blob/main/docs/manual/other-topics/typescript.md#usage
Possible new way if we merge this in v6: https://github.com/sequelize/sequelize/blob/120e2447dc166d256f030dbc07928776b24aa649/docs/manual/other-topics/typescript.md#usage

Which would also make migration easier as that second way is what I'd like to make the default in v7

ephys:
I'll merge this but we can still debate whether we land it in v6 or not :)

sdepold:
Power to the people!

ephys:
Thanks @sdepold :')

WikiRik:
> I'll merge this because there is consensus that we should support >= 4.1 in v7 ; but we can still debate whether we land it in v6 or not :)

Seeing the difference in usage I think it's good to include this in v6 as well. Seeing that we have updated various typings in the recent releases of v6 already it should be fine.

> Edit: If I understand correctly we first need to mark tests `TS Typings (3.9)` & `TS Typings (4.0)` as not required in the repository settings, which requires sdepold?

I updated the required checks so it requires 4.1 - 4.5 already for the main branch

sdepold:
What's the benefit of having this in v6?

WikiRik:
> It won't make our job easier but it will allow users to massively reduce the boilerplate they need to write for TS support in v6
> 
> Current way: https://github.com/sequelize/sequelize/blob/main/docs/manual/other-topics/typescript.md#usage Possible new way if we merge this in v6: https://github.com/sequelize/sequelize/blob/120e2447dc166d256f030dbc07928776b24aa649/docs/manual/other-topics/typescript.md#usage
> 
> Which would also make migration easier as that second way is what I'd like to make the default in v7

@sdepold this is the benefit of having it in v6

fncolon:
> It won't make our job easier but it will allow users to massively reduce the boilerplate they need to write for TS support in v6
> 
> Current way: https://github.com/sequelize/sequelize/blob/main/docs/manual/other-topics/typescript.md#usage Possible new way if we merge this in v6: https://github.com/sequelize/sequelize/blob/120e2447dc166d256f030dbc07928776b24aa649/docs/manual/other-topics/typescript.md#usage
> 
> Which would also make migration easier as that second way is what I'd like to make the default in v7

Cool!

sdepold:
Ah sorry. My Mumbo jumbo brain somehow thought that this thread is about the utils conversion.

I think we should change the docs part in the commit message to something else. Maybe Feature for the lack of a better idea since docs is likely to result in a patch version

ephys:
`feat` works for me. None of the others would cause a minor bump and we're not tagging it as BREAKING CHANGE

PR #13956, opened 2022-01-15
build: improve eslint config (part 2)

ephys:
### Pull Request Checklist

### Description Of Change

This PR:
- Removes `--fix` from the `lint-no-fix` script.
- Adds rules related to eslint
- Adds unicorn/expiring-todo-comments
- Errors on types used in JSDoc in TypeScript files (TypeDocs will extract the typescript definition instead)
- Makes eslint run on cjs, mjs, cts, mts files too
- Fixes the unresolved issues from PR https://github.com/sequelize/sequelize/pull/13930


lib/dialects/abstract/query.js
```diff
@@ -19,8 +19,7 @@ class AbstractQuery {
     this.options = {
       plain: false,
       raw: false,
-      // eslint-disable-next-line no-console
-      logging: console.log,
+      logging: console.debug,
```
WikiRik, line 22: What would this change mean effectively?

WikiRik, line 22: What would this change mean effectively for the end users?

WikiRik: Also holds for the other instances

lib/sequelize.js
```diff
@@ -1200,8 +1197,8 @@ class Sequelize {
       options = last;
 
       // remove options from set of logged arguments if options.logging is equal to console.log
```
WikiRik, line 1199: If we allow console.debug we should also mention that here

package.json
```diff
@@ -185,7 +185,7 @@
     }
   },
   "lint-staged": {
-    "*!(d).[tj]s": "eslint --fix --report-unused-disable-directives"
+    "*.{js,mjs,cjs,ts,mts,cts}": "eslint --fix --report-unused-disable-directives"
```
WikiRik, line 188: Didn't know mts and cts files also existed. Does this still ignore .d.ts files?

test/integration/dialects/sqlite/dao-factory.test.js
```diff
@@ -108,7 +108,7 @@ if (dialect === 'sqlite') {
             expect(user.name).to.equal('user');
           });
 
-          it.skip('should make aliased attributes available', async function () { // eslint-disable-line mocha/no-skipped-tests
+          it('should make aliased attributes available', async function () {
```
WikiRik, line 111: Any idea why we skipped this test before?

WikiRik (dismissed):
Not sure what the impact is of the console.log to console.debug change for the end-users. But apart from that, I don't see any weird changes. Thanks for the PR! 

EDIT: The impact is at least some failing tests

lib/dialects/abstract/query.js
ephys: `console.debug` is an alias for `console.log`

I'm not sure where to document that but I've configured eslint to forbid `console.log` while still allowing every other console methods.
The idea behind it is that we often use `console.log` to debug something. With this rule, we'll be warned that we forgot to remove throwaway logging.
`console.debug` is used to say "I am logging this deliberately."

For the same reason `debugger` is also forbidden by eslint.

People can still use them during development (encouraged even), but once they commit it should be removed :)

package.json
ephys: I've un-ignored .d.ts files because the linter is now able to process them. the `types` folder is still ignored, as we're migrating it anyway

test/integration/dialects/sqlite/dao-factory.test.js
ephys: Not sure, I re-enabled it to see which dialects (if any) fails. Currently they all fail, I'll look into it

lib/dialects/abstract/query.js
WikiRik: I thought that console.debug would only show if you changed some configuration so I'm not sure if all end users have that set up so that they will see these outputs. So not sure what to do with that to be honest

test/integration/dialects/sqlite/dao-factory.test.js
WikiRik: Yeah but it doesn't fail because of this so think its good that you re-enabled it then

lib/sequelize.js
ephys: Done!

lib/dialects/abstract/query.js
ephys: If you use another logger but not the base console.debug, it's just an alias :) https://nodejs.org/api/console.html#consoledebugdata-args

ephys: Fun fact, console.info is also just an alias https://nodejs.org/api/console.html#consoleinfodata-args

WikiRik: Ah, that's good to know. Reasonable change then

test/integration/dialects/sqlite/dao-factory.test.js
WikiRik: Nevermind, it failed this time

WikiRik: Apparently the test was disabled in https://github.com/sequelize/sequelize/commit/e10114fc58f4faaebc712d4711ecb8abf887a0b2 back in January 2014. Not sure why, but I think we can remove this test if we've been skipping this for this long

ephys: Confirmed that it still fails

If you're still on a computer, would you mind doing it?

WikiRik: Done

ephys: Thanks!

WikiRik (approved):

PR #13981, opened 2022-01-21
refactor(data-types): convert mysql data-types to TS

AllAwesome497:
 

### Description Of Change
converts mysql-specific data-types to TS.


WikiRik:
The lodash types are added in #13782 which I think we'll merge first and then you can merge/rebase main in the various branches

package.json
```diff
@@ -58,14 +58,15 @@
...
   "devDependencies": {
-    "@rushstack/eslint-patch": "1.1.0",
-    "@commitlint/cli": "16.0.2",
+    "@commitlint/cli": "16.1.0",
     "@commitlint/config-angular": "16.0.0",
     "@ephys/eslint-config-typescript": "14.1.1",
+    "@rushstack/eslint-patch": "1.1.0",
     "@types/chai": "4.3.0",
     "@types/mocha": "9.0.0",
+    "@types/mysql": "^2.15.20",
```
WikiRik, line 67: This one needs to be pinned, otherwise renovate will do that

package.json
```diff
@@ -58,14 +58,15 @@
     "wkx": "^0.5.0"
   },
   "devDependencies": {
-    "@rushstack/eslint-patch": "1.1.0",
```
WikiRik, line 61: I can remember other PRs that also change this and ibm_db so we don't have to do this here. I'll check which one and if that one is not looking like it will be merged soon I will make a separate PR

WikiRik (commented):
Two remarks on the package.json. Also would like to point out that we check for expiring TODOs (as a linter warning); https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md
Not sure if they really fit in this PR but just so you know about them

AllAwesome497:
Closing in favor of #14505.

Issue #13288, opened 2021-05-26
TypeScript complains about $nested.column$ syntax in query where clause.


MarkMurphy:
### What was unclear/insufficient/not covered in the documentation

TypeScript complains about `$nested.column$` syntax in query where clause. Because the where clause is typed as needing to be attributes of the model being queried, this syntax triggers an error since these object keys are not attributes of the model.

### If possible: Provide some suggestion on how we can enhance the docs

Maybe this is a bug if it's intended to work but otherwise the documentation seems to be lacking how to resolve this.

### Additional context

<img width="1380" alt="Screen Shot 2021-05-26 at 10 11 41 AM" src="https://user-images.githubusercontent.com/873474/119665455-ca6ca400-be0a-11eb-9589-f24ca025ff2d.png">

### Is this issue dialect-specific?

- [x] No. This issue is relevant to Sequelize as a whole.
- [ ] Yes. This issue only applies to the following dialect(s): XXX, YYY, ZZZ
- [ ] I don't know.

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [x] No, I don't have the time, although I believe I could do it if I had the time...
- [x] No, I don't have the time and I wouldn't even know how to start.

belisarenata:
Same problem here. TS indicates No overload match, but works just fine when a ts ignore flag is set to the nested clause as a workaround:   
![Screenshot from 2021-05-27 14-08-44](https://user-images.githubusercontent.com/40744589/119868161-172ca980-bef5-11eb-8389-e378fa3a5959.png)


Also, at Sequelize V.5 this does not raise any errors.

aryem:
@MarkMurphy Can you specify sequelize version.
In v6.6.5 `WhereOptions` have `Literal` that should support nested column literal.
https://github.com/sequelize/sequelize/blob/56bb1d6ef9827f604d7bcef945abb7e213f2322d/types/lib/model.d.ts#L112

aryem:
I have added additional tests to ensure nested column in where option in #13478

ephys:
I have a PR in progress for this https://github.com/sequelize/sequelize/pull/13983

PR #11825, opened 2020-01-14
docs(manuals): extensive rewrite

papb:
# docs(manuals): extensive rewrite

Finally! This is the PR I have been working on for a few months.

I have basically rewritten the whole manuals. This will solve several issues at once.

## Summary

* The `docs` folder was reorganized;
* Almost every manual was rewritten;
* Several new manuals were added;
* A lot more examples were provided;
* And **I took care not to break any existing links!**

  * Two of the manuals were renamed:

    * `manual/dialects.html` changed to `manual/dialect-specific-things.html`;
    * `manual/instances.html` changed to `manual/model-instances.html`.

    To make sure the old links for these two still work, I have set up a permanent redirect for the old URLs to the new ones. See `docs/redirects` and `docs/redirects.json`.

  * The following five manuals were split into several pages, with a lot more details:

    * `associations.md`;
    * `data-types.md`;
    * `models-definition.md`;
    * `models-usage.md`;
    * `querying.md`

   Again, to make sure every link that is around the internet is still valid and useful, each of these pages became a mini-page linking for the new guides. See for example `docs/manual/moved/associations.md`.

## Consequences

This PR will not only make me a lot more comfortable with the Sequelize guides in general, but will also close several issues!

* Closes #2640
* Closes #4553
* Closes #5471
* Closes #6541
* Closes #8534
* Closes #9103
* Closes #9405
* Closes #10503
* Closes #10828
* Closes #10860
* Closes #10969
* Closes #10972
* Closes #10974
* Closes #10989
* Closes #11280
* Closes #11389
* Closes #11424
* Closes #11687
* Closes #11731
* Closes #11745
* Closes #11751
* Closes #11787
* Closes #11796

PR #13983, opened 2022-01-21
fix(types): accept `$nested.syntax$` in WhereAttributeHash

ephys:
### Pull Request Checklist

### Description Of Change

Tiny change that has become possible since we dropped TS < 4.1: 
`User.findOne({ where: { '$nested.column$': 'blah2' } })` is not reported as an error by TypeScript anymore.

Closes https://github.com/sequelize/sequelize/issues/13288

### TODO

- [x] Check whether `$name$` is valid
- [x] Check whether `$include.include.name$` is valid
- [x] Add tests for nested nested include


ephys:
TypeScript < 4.4 does not support using Template Literal Types as keys so I had to ts-ignore both the test & the implementation - but it doesn't cause issues in ts < 4.4 and works in ts >= 4.4

We can enable it again on 2022-05-26 when we drop support for TS < 4.4

ephys:
I think with some more black magic we should be able to typecheck `include.where` too, but I'll keep it for a follow-up PR

```typescript
User.findOne({
  include: [{
    association: User.associations.group,
    where: {
      // these are not currently type-checked
      test: 'test',
    }
  }],
});
```

ephys:
This PR should also include support for the JSON nested key syntax https://sequelize.org/v4/manual/tutorial/querying.html#nested-key

I can't find the documentation on that syntax for v6. Is it still a thing?

WikiRik:
> This PR should also include support for the JSON nested key syntax https://sequelize.org/v4/manual/tutorial/querying.html#nested-key
> 
> I can't find the documentation on that syntax for v6. Is it still a thing?

It seems that it got lost in https://github.com/sequelize/sequelize/pull/11825

ephys:
I'll open a follow-up PR for the `json.syntax` as the typing I have in mind are a bit complicated

types/lib/model.d.ts
```diff
@@ -372,8 +372,19 @@ export type WhereAttributeHash<TAttributes = any> = {
...
+} & {
+  /**
+   * Makes $nested.syntax$ valid, but does not type-check the name of the include nor the name of the include's attribute.
    */
-  [field in keyof TAttributes]?: WhereValue<TAttributes> | WhereOptions<TAttributes>;
+  // TODO [2022-05-26]: Remove this ts-ignore once we drop support for TS < 4.4
+  // TypeScript < 4.4 does not support using a Template Literal Type as a key.
+  //  note: this *must* be a ts-ignore, as it works in ts >= 4.4
+  // @ts-ignore
+  [attribute: `$${string}.${string}$`]: WhereValue<TAttributes> | WhereOptions<TAttributes>;
```
AllAwesome497, line 387: won't this still be broken in ts <4.4?

It might be better to do something like `& { [attribute: string]: WhereValue<any>` (iirc it'll keep the other keys typed but anything else will be accepted, can't recall for certain though)

ephys: It will still be broken in TS < 4.4, but not more than before (it's just going to ignore this addition)

Strict typings in `where` is so important IMO that it's not worth it to dumb it down for old TS versions.
I've caught a handful of broken `where` in my own code after properly typing my attributes.

If someone wants to fix this in TS < 4.4, I'd say open a follow-up PR that uses [`typesVersion`](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions) to provide TS < 4.4 specific code.

ephys:
As the only issue is that the false-negative is still present in TS < 4.4, I think I'll merge this. Then we can fix https://github.com/sequelize/sequelize/issues/14038, and then open a follow-up PR  that uses `typesVersion` to provide an alternative typing for TS < 4.4

WikiRik:
Sounds good! We should also publish a new alpha release of v7 after we merged all these typing PRs

ephys:
@WikiRik @AllAwesome497 if you're ok with the above plan, can I get one of your review approvals? :)

WikiRik (approved):

Issue #9605, opened 2018-06-28
`limit` with `subQuery: false` return wrong result


husnulhamidiah:
## What are you doing?

Let's say I have `Transaction` model which have many-to-many relation with `Good` model. And also belongs to `User` model. Here's my model and relation definition.

```js
module.exports = (sequelize, DataTypes) => {
  const Transaction = sequelize.define('Transaction', {
    id: {
      type: DataTypes.UUID,
      primaryKey: true,
      defaultValue: DataTypes.UUIDV4,
      allowNull: false,
      autoIncrement: false,
    },
    userId: {
      type: DataTypes.UUID,
      allowNull: false,
      foreignKey: true,
      references: {
        model: 'users',
        key: 'id',
      },
    },
    status: DataTypes.ENUM('verified', 'waiting', 'empty', 'rejected'),
    channel: {
      type: DataTypes.INTEGER,
      allowNull: false,
    },
  }, {});

  Transaction.associate = (models) => {
    Transaction.belongsTo(models.User, { as: 'user', foreignKey: 'userId' });
  };

  return Transaction;
};
```

```js
module.exports = (sequelize, DataTypes) => {
  const Good = sequelize.define('Good', {
    id: {
      type: DataTypes.UUID,
      primaryKey: true,
      defaultValue: DataTypes.UUIDV4,
      allowNull: false,
      autoIncrement: false,
    },
    rphAset: {
      type: DataTypes.BIGINT,
      allowNull: false,
      defaultValue: 0,
    },
    urUpb: {
      type: DataTypes.STRING,
      allowNull: false,
    },
  }, {});

  return Good;
};
```

```js
db.Good.belongsToMany(db.Transaction, { as: 'transactions', through: db.TransactionGood, foreignKey: 'goodId' });
db.Transaction.belongsToMany(db.Good, { as: 'goods', through: db.TransactionGood, foreignKey: 'transactionId' });
```

I already seed my database with data so user with id `30e1c140-d4d0-4edb-980a-3d7b9a47a958` has 7 transaction. I want to get transaction for that user with all goods in it. Here's my code.

```js
const x = await models.Transaction.findAll({
    subQuery: false,
    distinct: true,
    where: {
      userId: '30e1c140-d4d0-4edb-980a-3d7b9a47a958',
    },
    include: [
      {
        model: models.User,
        as: 'user',
        attributes: ['id','name','createdAt','updatedAt'],
      },
      {
        model: models.Good,
        as: 'goods',
      },
[...]

PR #13985, opened 2022-01-21
docs: clarify how the `limit` option works

ephys:
### Pull Request Checklist

### Description Of Change

I'm attempting to clarify what `FindOptions#limit` does and its interaction with `FindOptions#subQuery`

Related to https://github.com/sequelize/sequelize/issues/9605


WikiRik (dismissed):
Looks good. Just to check; do we have tests for these three examples to ensure that's the actual behavior?

ephys:
We test for `limit + include` here: https://github.com/sequelize/sequelize/blob/main/test/unit/sql/select.test.js#L281
However we never test `limit + include + subQuery: false`. But we do it: https://github.com/sequelize/sequelize/blob/main/lib/dialects/abstract/query-generator.js#L1164

There is also an undocumented `groupedLimit` option, I'm not sure of what it does.

ephys:
I've removed the third example I added as it's an unrelated limit option with a different behavior.

WikiRik:
> However we never test `limit + include + subQuery: false`. But we do it: https://github.com/sequelize/sequelize/blob/main/lib/dialects/abstract/query-generator.js#L1164

Could you add tests for this?

ephys:
Sounds good, I'll look into it :)

ephys:
Test added

WikiRik (dismissed):

WikiRik (commented):
Woops, was too early with the approval. Do fix the mssql assertion first

ephys:
That's weird. Why does mssql produce 
`ORDER BY [user].[last_name] ASC, [user].[id_user]` 
instead of 
`ORDER BY [user].[last_name] ASC`

it doesn't seem to be doing that in other queries or when `subQuery` is true

ephys:
Ok it should be fixed

ephys:
Not in this PR but we'll eventually get to it as we migrate to TS

PR #13987, opened 2022-01-21
feat: throw if include.limit set, but not separate

ephys:
### Pull Request Checklist

### Description Of Change

v7 only.

Specifying `limit` on an eagerly loaded association will silently turn the `separate` option on.

I'm making a small change where the user *must* enable `separate` manually instead. This is an opinionated change, but I believe running two queries sequentially has too much of a performance impact to not notify the user.

It also throws if `separate` is explicitly set to false even though `limit` is specified (whereas currently it will silently ignore `limit`).

---

Still working on this because if `separate` is false, it could generate this kind of query instead:

```sql
SELECT * FROM (SELECT * FROM users LIMIT 2) users 
LEFT JOIN (SELECT * FROM projects LIMIT 2) projects 
  ON projects.uid = users.id;
```


WikiRik:
Do you want the branch to stay until that later date?

PR #13988, opened 2022-01-22
Feature/v6 release

sdepold:
Tiny script to merge stuff into the v6 branch based on the `to be released` column of https://github.com/sequelize/sequelize/projects/4


sdepold:
I'm going to close this since it's in v6 already and there is not really any value in main for it :)

PR #13992, opened 2022-01-22
feat: forbid using attr names that are ambiguous with $nested.syntax$

ephys:
### Pull Request Checklist

### Description Of Change

v7 only

It's currently possible to name an attribute `$name$` or `$na.me$`. This is ambiguous with the `$nested.syntax$` that can be used in queries.

This PR makes that syntax reserved and rejects any attribute named in a way that would be ambiguous with the nested attribute reference syntax.

It's opinionated as I found a couple places in the code where Sequelize *tries* to make it work, but there are many, many other places where it just silently breaks.

This is a breaking change, and should be marked as such when squashing the commits from this PR.


ephys:
I think we should forbid using `.` in attribute names too, as that would be ambiguous with the `json.reference.syntax` and the `$nested.attribute.reference.syntax$`

docs/manual/other-topics/upgrade-to-v7.md
```diff
@@ -16,6 +16,18 @@ As a result, the manual typings that were formerly best-effort guesses on top of
 have been removed and all typings are now directly retrieved from the actual TypeScript code.
 You'll likely find many tiny differences which however should be easy to fix.
 
+## Attributes cannot start and end with `$`, or include `.`
```
WikiRik, line 19: Maybe good to add an example of how people can change the names of the attributes they are using?

ephys: Done! :)

WikiRik (approved):

sdepold:
Can we add this to the list of breaking changes in the migration docs?

ephys:
It is :)

sdepold:
Ah true :) Guess I didn't see it on my phone. Let's get this into main then!

PR #14004, opened 2022-01-24
fix: rollback PR #13951 in v6

### LINKED ISSUE #14003: Recent changes to not call overloaded model functions are a breaking change
This change: https://github.com/sequelize/sequelize/commit/b253d8ed63c91bc2c7143f07806554b5a5ac67eb avoids calling overloaded functions internally. I understand why it was undertaken, but this effectively breaks the way some users (and even other libraries) use Sequelize with 6.14.x.

Specifically, I believe [sequelize-typescript](https://github.com/RobinBuschmann/sequelize-typescript) to be broken with this change. I believe this package overrides the default model functions: https://github.com/RobinBuschmann/sequelize-typescript/blob/b60c011be2e971e56cb783d4ade994965faab916/src/model/model/model.ts#L209. Our codebase is a complex one, so I'm having difficulty concluding 100% that this override is the cause...

In any case, I have confirmed that if the lines to avoid the overrides are reverted, 6.14.x works fine by manually reverting them in the sequelize dist in my `node_modules`. I'm not a contributor to to either of these repos, but IMHO, a minor version shouldn't cause significant breakages like this. I imagine that over the next week, many devs will have to track down why their code is suddenly not working with a minor sequelize update.

ephys:
### Pull Request Checklist

### Description Of Change

Rollbacking changes from #13951 in v6. The change is still present in v7.
I'm opening a follow-up PR to document the breaking change in v7.

Closes #14003


WikiRik (approved):

WikiRik:
@ephys I'll let you merge this yourself, but do note that it will trigger an automatic release

ephys:
Sounds good
The only failing test is the old flaky mysql one. Let's publish that fix :)

PR #14006, opened 2022-01-24
test(mssql): remove unhandled rejection tests

WikiRik:
Quote from @ephys ;
"I think it's weird that we test that not handling the promise rejection causes an unhandled promise rejection. That's just the standard promise behavior. It's also the only place where we test unhandled promise rejection.

IMO we can just drop the unhandled rejections suite - or just check that the promise rejects."

This allows us to update mocha to the latest version since the current test does not work with that.


ephys (approved):
`test/unit/model/validation.test.js` already has tests to ensure `create` rejects if the input is invalid so this is good to go

PR #14009, opened 2022-01-25
feat(dialects): add support for DB2 for IBM i

markdirish:
### Pull Request Checklist

### Description Of Change

This PR adds support for Db2 for i, the database that lives at the heart of [IBM i](https://www.ibm.com/it-infrastructure/power/os/ibm-i). Every effort was made to limit the amount of changes made outside of the dialect, and there should only be a few small `dialect` checks in the base code. The vast majority of the PR is inside the dialect code itself, and the test suite to 

This is a multi-year effort that is finally of enough quality to merge. The dialect currently passes 1528 integration tests, *and fails only 6*.

### Todos

- [x] Ensure unit tests pass
- [ ] Document `odbcConnectionString` usage when creating a sequelize object


markdirish:
There are 43 failing unit tests, most all of them just need me to go in and change the expectation for the test. Working on that now.

WikiRik:
Thanks for your work on this!
While you fix the remaining unit tests, I have some additional TODOs before we merge this;
- Not sure what the conclusion was last meeting, but if there is some way we can run the unit tests that would be great. You can add that [here](https://github.com/sequelize/sequelize/blob/main/.github/workflows/ci.yml)
- There is documentation that needs to be updated for this, mainly in the [docs index](https://github.com/sequelize/sequelize/blob/main/docs/index.md), [getting started](https://github.com/sequelize/sequelize/blob/main/docs/manual/core-concepts/getting-started.md), and the [dialect specific page](https://github.com/sequelize/sequelize/blob/main/docs/manual/other-topics/dialect-specific-things.md)
- Something that we from the maintainer side need to do is update the description of the GitHub repository (including tags, for both this and Snowflake) and create a new label for issues/PRs called `dialect: ibmi` (and for Snowflake and DB2 as well since we haven't done that yet)

Feel free to add additional ones if you come up with it (cc @ephys @sdepold )

markdirish:
Unit tests now all pass:
```
  919 passing (1s)
```

ephys:
Congrats!
The only failing test is the ESM one
You can fix it by adding 'ibmi' to the list of ignored exports in `esm-named-exports.test.js`

ephys:
I'm icky about not having a way to run automated tests for this. Without them we are inevitably going to break this and not realise.

markdirish:
> Congrats! The only failing test is the ESM one You can fix it by adding 'ibmi' to the list of ignored exports in `esm-named-exports.test.js`

Oops, forgot to push the changes. Should be in now

> I'm icky about not having a way to run automated tests for this. Without them we are inevitably going to break this and not realize.

I think we talked about picking 10-20 tests that we can run against a (slowwww) public system. The tests will take much longer than the default timeouts to run. We will have to set `SEQ_TEST_CLEANUP_TIMEOUT` to something very large, and make sure that `.mocharc.jsonc` has `timeout` also set to a large number.

I will also try my hardest to run the integration tests as frequently as possible.

markdirish:
@WikiRik ,

> *  Not sure what the conclusion was last meeting, but if there is some way we can run the unit tests that would be great. You can add that here
> * There is documentation that needs to be updated for this, mainly in the docs index, getting started, and the dialect specific page

Working today on these. I think I will try to isolate 10-20 of the more complex integration tests that can be run for a build. I will also work on creating an action that does all of the steps needed to get this running:
1. Create a Linux container
2. Download and install unix-odbc
3. Download and install nodejs (12, 14, 16?)
4. Download our ODBC driver
5. Run the tests

WikiRik:
Sounds like a good approach. We use Node 12 and 16 in the CI but it would be nice if we could easily change that in a matrix like we do with the others

markdirish:
Documentation should now be updated.

I also did not know about `dialectOptions` in the main sequelize.options. Have moved my `odbcConnectionString` option to there.

Working on GitHub Actions now.

.github/workflows/ci.yml
```diff
@@ -78,6 +78,41 @@ jobs:
...
+      - name: Install unixodbc
+        run: sudo apt-get install -y unixodbc unixodbc-dev
+      - name: Download IBM i ODBC driver
+        run: |
+          mkdir installer
+          wget --no-verbose https://public.dhe.ibm.com/software/ibmi/products/odbc/ibm-iaccess-1.1.0.15-1.0.amd64.deb -O installer/ibm-iaccess-1.1.0.15-1.0.amd64.deb
+      - name: Install IBM i ODBC driver
+        run: sudo dpkg -i installer/ibm-iaccess-1.1.0.15-1.0.amd64.deb
+      - name: Install `odbc` package
+        run: yarn add odbc
```
WikiRik, line 111: I see that you're still testing things out, but I think this should be added to the dev dependencies and marked as optional in peerDependenciesMeta. Do note that because of the frozen lockfile you need to commit the updates to yarn.lock as well before CI picks it up correctly

markdirish: Good tip. Not sure the relationship between the yarn.lock and stuff like dev dependencies, so I will look into it.

markdirish:
So... I have the CI workflow working on my system. But because the smoke tests require credentials to an outside system, I thought it best to store them in a secret that can then be pulled in. Unfortunately, PR commits don't have access to secrets for security reasons. So we may have to come up with some solution for that. But here is how it builds on my fork (notice that 9 tests took over 20 minutes...):

image

WikiRik:
Yeah, secrets are not passed to this PR since your branch is from a fork. Not sure what a good solution would be, but my initial idea is to only run the IBMi smoke tests after something has been pushed to main (like release but then on main) or run on a schedule on the main branch. This way we will notice that IBMi smoke tests fail before we make a new release and it does not impact PRs. But probably there is a nicer way of solving this

.github/workflows/ci.yml
markdirish: I think I have the package.json/yarn.lock stuff fixed, and I have removed the offending line from the ci.yml

markdirish:
Most recent build starting running tests for Node 12 and Node 16, but some failed tests. Why? They are simultaneously testing against the same resource. May need to either run in series, or create separate usernames for each node version.

test/integration/associations/scope.test.js
```diff
@@ -11,6 +11,10 @@ const current = Support.sequelize;
 const Op = Sequelize.Op;
 const semver = require('semver');
 
+const {
```
WikiRik, line 14: Why did you add this here? It caused problems with some of the other dialects and does not seem to be used in this file

markdirish: The only thing I can think of is that it was some sort of debug tool that didn't get cleaned up. Removing now

markdirish:
Re-built and reran. Everything passes except db2 (?) and IBM i, but IBM i works on my own repo with access to the secret credentials.

WikiRik:
Yeah, don't worry about DB2. We need to fix that separately.
I do think we need to figure out a way to run the IBM i tests sometimes on this repo as well before we merge this. 
But for now, I think we can go ahead and do a code review. I'll probably take a good look at this on Friday

test/unit/sql/insert.test.js
```diff
@@ -99,7 +103,7 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
 
     it('formats date correctly when sub-second precision is explicitly specified', () => {
       const timezoneSequelize = Support.createSequelizeInstance({
-        timezone: Support.getTestDialect() === 'sqlite' ? '+00:00' : 'CET',
+        timezone: Support.getTestDialect() === 'sqlite' || Support.getTestDialect() === 'ibmi' ? '+00:00' : 'CET',
```
WikiRik, line 106:
```suggestion
        timezone: ['sqlite','ibmi'].includes(Support.getTestDialect()) ? '+00:00' : 'CET',
```

test/unit/sql/insert.test.js
```diff
@@ -66,7 +68,7 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
   describe('dates', () => {
     it('formats the date correctly when inserting', () => {
       const timezoneSequelize = Support.createSequelizeInstance({
-        timezone: Support.getTestDialect() === 'sqlite' ? '+00:00' : 'CET',
+        timezone: Support.getTestDialect() === 'sqlite' || Support.getTestDialect() === 'ibmi' ? '+00:00' : 'CET',
```
WikiRik, line 71:
```suggestion
        timezone: ['sqlite','ibmi'].includes(Support.getTestDialect()) ? '+00:00' : 'CET',
```

test/unit/sql/index.test.js
```diff
@@ -68,6 +69,11 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
         sqlite: 'CREATE UNIQUE INDEX `a_b_uniq` ON `User` (`fieldB`, `fieldA` COLLATE `en_US` DESC)',
         mssql: 'CREATE UNIQUE INDEX [a_b_uniq] ON [User] ([fieldB], [fieldA] DESC)',
         db2: 'CREATE UNIQUE INDEX "a_b_uniq" ON "User" ("fieldB", "fieldA" DESC)',
+        ibmi: `BEGIN
```
WikiRik, line 72: Any reason that this is not on 1 line?

test/unit/sql/create-table.test.js
```diff
@@ -30,6 +30,11 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
           mariadb: 'CREATE TABLE IF NOT EXISTS `foo`.`users` (`id` INTEGER NOT NULL auto_increment , `mood` ENUM(\'happy\', \'sad\'), PRIMARY KEY (`id`)) ENGINE=InnoDB;',
           mysql: 'CREATE TABLE IF NOT EXISTS `foo.users` (`id` INTEGER NOT NULL auto_increment , `mood` ENUM(\'happy\', \'sad\'), PRIMARY KEY (`id`)) ENGINE=InnoDB;',
           mssql: 'IF OBJECT_ID(\'[foo].[users]\', \'U\') IS NULL CREATE TABLE [foo].[users] ([id] INTEGER NOT NULL IDENTITY(1,1) , [mood] VARCHAR(255) CHECK ([mood] IN(N\'happy\', N\'sad\')), PRIMARY KEY ([id]));',
+          ibmi: `BEGIN
```
WikiRik, line 33: Any reason the changes in these files are on separate lines?

test/support.js
```diff
@@ -24,7 +24,7 @@ chai.should();
 // Make sure errors get thrown when testing
 process.on('uncaughtException', e => {
   console.error('An unhandled exception occurred:');
-  throw e;
+  // throw e;
```
WikiRik, line 27:
```suggestion
  throw e;
```

test/integration/query-interface.test.js
```diff
@@ -476,8 +476,6 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
         expect(Object.keys(foreignKeys[0])).to.have.length(8);
       } else if (['mysql', 'mariadb', 'mssql'].includes(dialect)) {
         expect(Object.keys(foreignKeys[0])).to.have.length(12);
-      } else {
-        throw new Error(`This test doesn't support ${dialect}`);
```
WikiRik, line 480: Why is this removed?

test/integration/query-interface.test.js
```diff
@@ -666,7 +664,10 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
           throw new Error('Error not thrown...');
         } catch (error) {
           expect(error).to.be.instanceOf(Sequelize.UnknownConstraintError);
-          expect(error.table).to.equal('users');
+          if (dialect !== 'ibmi') {
+            expect(error.table).to.equal('users');
```
WikiRik, line 668: Anything we can expect for ibmi?

lib/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,848 @@
+'use strict';
+
+const Utils = require('../../utils');
+const util = require('util');
+// const Transaction = require('../../transaction');
```
WikiRik, line 5: Why is this here? I think it can be removed right?

lib/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,848 @@
...
+    let schema;
+    if (typeof options.schema === 'string') {
+      schema = this.quoteIdentifiers(options.schema);
+    }
+
+    // Although the function is 'addIndex', and the values are passed through
+    // the 'indexes' key of a table, Db2 for i doesn't allow REFERENCES to
+    // work against a UNIQUE INDEX, only a UNIQUE constraint.
+    if (options.unique) {
+      // return `ALTER TABLE ${tableName} ADD CONSTRAINT ${this.quoteIdentifiers(options.name)} UNIQUE (${fieldsSql.join(', ')}${options.operator ? ` ${options.operator}` : ''})`;
```
WikiRik, line 405: This is not needed anymore right?

lib/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,848 @@
...
+    const usingClause = `USING (
+      SELECT * FROM (${this.quoteTable(tableName)}
+      VALUES(42)
+      ) AS ${aliasTable}("id") ON (${aliasTable}."id" = ${this.quoteTable(tableName)}."id")`;
+
+    query += usingClause;
+    query += ` WHEN MATCHED THEN ${this.updateQuery(tableName, tableName, where, options, updateValues)}
+    WHEN NOT MATCHED THEN ${this.insertQuery(tableName, insertValues, model, options).sql}`;
+
+    // where = this.getWhereConditions(where, null, model, options);
```
WikiRik, line 450: Is this still needed? (same as for lines 451-454)

lib/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,848 @@
...
+    if (offset) {
+      if (typeof offset === 'number' && Number.isInteger(offset)) {
+        fragment += ` OFFSET ${offset} ROWS`;
+      } else {
+        console.warn('"offset" must be an integer. Offset is not added');
+      }
+    }
+
+    if (limit) {
+      // if (typeof limit === 'number' && Number.isInteger(limit)) {
```
WikiRik, line 533: This can also be removed, just like lines 535-537

lib/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,848 @@
...
+
+    if (attribute.unique === true && !attribute.primaryKey) {
+      template += ' UNIQUE';
+    }
+
+    if (attribute.primaryKey) {
+      template += ' PRIMARY KEY';
+    }
+
+    // Db2 for i comments are a mess
```
WikiRik, line 699: Do you want to keep this?

test/integration/sequelize/query.test.js
```diff
@@ -130,16 +130,17 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
           benchmark: true,
         });
 
-        await sequelize.query('select 1;');
-        expect(logger.calledOnce).to.be.true;
+        await sequelize.query(`select 1${dialect === 'ibmi' ? ' FROM SYSIBM.SYSDUMMY1' : ''};`);
+        console.log(logger);
```
WikiRik, line 134: This can be removed

test/integration/sequelize/query.test.js
```diff
@@ -130,16 +130,17 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+        expect(logger.called).to.be.true;
```
WikiRik, line 135:
```suggestion
        expect(logger.calledOnce).to.be.true;
```
Or did you change this on purpose?

test/integration/sequelize/query.test.js
```diff
@@ -195,18 +196,21 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
             },
           });
 
+          console.log(createSql);
```
WikiRik, line 199: We can remove this, and the line after this

test/integration/sequelize/query.test.js
```diff
@@ -195,18 +196,21 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+          console.log(updateSql);
           expect(createSql).to.match(/; ("john", "john@gmail.com"|{"(\$1|0)":"john","(\$2|1)":"john@gmail.com"})/);
           expect(updateSql).to.match(/; ("li", 1|{"(\$1|0)":"li","(\$2|1)":1})/);
         });
 
+        // TODO: fix this
```
WikiRik, line 205: Do you still want to do this?

test/integration/sequelize/query.test.js
```diff
@@ -658,58 +669,58 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
       } else if (dialect === 'mssql') {
         expect(logSql).to.include('@0');
         expect(logSql).to.include('@1');
-      } else if (dialect === 'mysql') {
-        expect(logSql.match(/\?/g).length).to.equal(2);
+      } else if (dialect === 'mysql' || dialect === 'ibmi') {
+        // expect(logSql.match(/\?/g).length).to.equal(2);
```
WikiRik, line 673: Why is this commented out?

WikiRik (commented):
Many small comments but overall looks good. Thanks for your work!

markdirish:
Taking a look at review now.

.github/workflows/ci.yml
```diff
@@ -78,6 +79,39 @@ jobs:
         run: yarn test-unit
       - name: Integration Tests
         run: yarn test-integration
+  test-ibmi:
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [12]
```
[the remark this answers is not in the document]
markdirish: The issue I was running into is that both actions were running against the same user/schema, so they were creating/deleting each other's tables (the pitfalls of testing against an external DB). I think that I have multiple users set up now, just need to adjust the action so that they use their own. Still could be an issue when they use an explicit schema.

docs/index.md
```diff
@@ -15,7 +15,7 @@
...
+Sequelize is an easy-to-use and promise-based [Node.js](https://nodejs.org/en/about/) [ORM tool](https://en.wikipedia.org/wiki/Object-relational_mapping) for [Postgres](https://en.wikipedia.org/wiki/PostgreSQL), [MySQL](https://en.wikipedia.org/wiki/MySQL), [MariaDB](https://en.wikipedia.org/wiki/MariaDB), [SQLite](https://en.wikipedia.org/wiki/SQLite), [DB2](https://en.wikipedia.org/wiki/IBM_Db2_Family), [Microsoft SQL Server](https://en.wikipedia.org/wiki/Microsoft_SQL_Server), and [IBM i](https://www.ibm.com/support/pages/db2-ibm-i). It features solid transaction support, relations, eager and lazy loading, read replication and more.
```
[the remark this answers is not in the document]
markdirish: What looks like happened is I assumed that this file was in-sync with `README.md`, since it is mostly the same. But it isn't, as `README.md` lacks Redshift and Snowflake but includes Db2. Copy-paste error.

lib/model.js
```diff
@@ -2719,11 +2722,11 @@ class Model {
         }
       }
 
-      if (options.ignoreDuplicates && ['mssql', 'db2'].includes(dialect)) {
+      if (options.ignoreDuplicates && ['mssql', 'db2', 'ibmi'].includes(dialect)) {
         throw new Error(`${dialect} does not support the ignoreDuplicates option.`);
       }
 
-      if (options.updateOnDuplicate && (dialect !== 'mysql' && dialect !== 'mariadb' && dialect !== 'sqlite' && dialect !== 'postgres')) {
+      if (options.updateOnDuplicate && (dialect !== 'mysql' && dialect !== 'mariadb' && dialect !== 'sqlite' && dialect !== 'postgres' && dialect !== 'ibmi')) {
```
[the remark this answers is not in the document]
markdirish: I tried to hit these where I could, but I definitely missed a few. Most of the `db2` checks were changed to `.includes` since they went from 1 to 2 . I might skip going through and touching these for now, and then refactor in separate PR

package.json
```diff
@@ -88,16 +88,17 @@
...
     "lint-staged": "12.3.2",
     "mariadb": "2.5.5",
     "markdownlint-cli": "0.30.0",
     "mocha": "9.2.0",
     "module-alias": "2.2.2",
-    "ibm_db": "2.8.1",
     "mysql2": "2.3.3",
     "node-hook": "1.0.0",
     "nyc": "15.1.0",
+    "odbc": "^2.4.1",
```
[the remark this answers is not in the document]
markdirish: And we have a new `odbc` version, so bumping to `2.4.2`

test/integration/associations/belongs-to.test.js
```diff
@@ -628,7 +628,8 @@ describe(Support.getTestDialectTeaser('BelongsTo'), () => {
 
     // NOTE: mssql does not support changing an autoincrement primary key
     if (Support.getTestDialect() !== 'mssql'
```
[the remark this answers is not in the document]
markdirish: Furthermore, looks like the file already stores `getTestDialect()`, so might as well use the stored value.

test/integration/associations/has-one.test.js
```diff
@@ -458,7 +458,8 @@ describe(Support.getTestDialectTeaser('HasOne'), () => {
 
     // NOTE: mssql does not support changing an autoincrement primary key
     if (Support.getTestDialect() !== 'mssql'
```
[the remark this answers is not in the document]
markdirish: And again, looks like dialect is saved at the top of the file. Not sure if that was added later, but definitely better than calling `getTestDialect` over and over.

test/integration/model/attributes/field.test.js
```diff
@@ -247,6 +247,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       it('increment should work', async function () {
         await this.Comment.destroy({ truncate: true });
         const comment1 = await this.Comment.create({ note: 'oh boy, here I go again', likes: 23 });
+        console.log(comment1);
```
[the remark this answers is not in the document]
markdirish: Sorry about these, I thought I searched for `console.` and removed all of my debug lines. It's possible that I excluded the test files in my search.

test/unit/sql/create-table.test.js
markdirish: The query generator itself generates these queries as multi-line queries. I did this for readability, as I find:

```
BEGIN
  DECLARE CONTINUE HANDLER FOR SQLSTATE VALUE '42891'
  BEGIN END;
  ALTER TABLE "User" ADD CONSTRAINT "a_b_uniq" UNIQUE ("fieldB", "fieldA" DESC);
END
```
a lot easier to conceptualize than
```
BEGIN DECLARE CONTINUE HANDLER FOR SQLSTATE VALUE '42891' BEGIN END; ALTER TABLE "User" ADD CONSTRAINT "a_b_uniq" UNIQUE ("fieldB", "fieldA" DESC); END
```

test/integration/query-interface.test.js
markdirish: Unfortunately, IBM i doesn't give much good information in its error in this case. Instead, it will simply indicate that there was an unknown constraint of `unknown__constraint__name`.

lib/dialects/abstract/query-generator.js
```diff
@@ -2670,7 +2670,7 @@ class QueryGenerator {
         }
 
         if (value.length > 0) {
-          return this._joinKeyValue(key, `(${value.map(item => this.escape(item, field)).join(', ')})`, comparator, options.prefix);
+          return this._joinKeyValue(key, `(${value.map(item => this.escape(item, field, { where: true })).join(', ')})`, comparator, options.prefix);
```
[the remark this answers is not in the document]
markdirish: If I remember right, there was some syntax required for IBM i to build this join that is triggered by the `where` option. I _thought_ about moving this to the IBM i query generator, but it felt like overkill for a single change to this large function. Trying to track down the test now.

markdirish:
Rerunning the entire test suite now with changes and merge from `main` (will take a while). Tomorrow I will look at any failing tests, and fix them up.

@WikiRik , I will let you resolve the remainder of the conversations at your discretion, I tried to answer the questions you had.

.github/workflows/ci.yml
```diff
@@ -78,6 +79,39 @@ jobs:
...
+      - name: Download IBM i ODBC driver
+        run: |
+          mkdir installer
+          wget --no-verbose https://public.dhe.ibm.com/software/ibmi/products/odbc/ibm-iaccess-1.1.0.15-1.0.amd64.deb -O installer/ibm-iaccess-1.1.0.15-1.0.amd64.deb
+      - name: Install IBM i ODBC driver
+        run: sudo dpkg -i installer/ibm-iaccess-1.1.0.15-1.0.amd64.deb
+      - name: Unit Tests
+        run: yarn test-unit
+      - name: Smoke Tests
+        run: yarn test-smoke
```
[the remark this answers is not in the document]
sdepold: Given how kind those tests run and how often we merge to master I wonder if a time based approach is better. Should we run tests once a day maybe?

lib/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
markdirish: Tracked it down. There are queries where Db2 for i will not accept `NULL`, and instead required `cast(NULL as int)`. I have asked the developers of the database, and they were surprised by this as well, so I _think_ its a database bug. But anyways, I need a way to differentiate between `NULL` in these two types of queries:

```
SELECT "id", "createdAt", "updatedAt" FROM "Users" AS "User" WHERE "User"."id" IN (1, 2, cast(NULL as int))
```
and
```
SELECT "createdAt", "updatedAt", "deletedAt", "UserId", "ProjectId" FROM "UserProjects" AS "UserProjects" WHERE ("UserProjects"."deletedAt" IS NULL AND ("UserProjects"."UserId" = 2 AND "UserProjects"."ProjectId" IN (2, 1, 3))
```
I looked at the values passed to escape in those 2 calls to `escape`, and couldn't find anything that I could use to differentiate them. So I passed `where: true` as a flag to indicate that NULL should be cast in this case.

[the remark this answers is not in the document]
WikiRik: Makes sense then. Is the issue with not accepting `NULL` documented somewhere public so we know when the casting can be removed?

test/integration/model/create.test.js
```diff
@@ -1094,7 +1096,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         try {
           await User.sync({ force: true });
           const tableName = User.getTableName();
-          await this.sequelize.query(`CREATE UNIQUE INDEX lower_case_username ON "${tableName}" ((lower(username)))`);
+          await this.sequelize.query(`CREATE UNIQUE INDEX lower_case_username ON "${tableName}" ((lower("username")))`);
```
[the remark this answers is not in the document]
WikiRik: Yeah, it seems that the quotes do not change the outcome so that's nice

lib/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
markdirish: I don't believe it is really documented behavior just a bug. I also think it won't be fixed until a future release of IBM i, which will remain "in support" for 4-8 years. So I think it will have to remain for the foreseeable future, unless we can figure a different way to indicate this particular SQL structure.

.github/workflows/ci.yml
```diff
@@ -78,6 +78,36 @@ jobs:
...
+    needs: test-typings
+    env:
+      DIALECT: ibmi
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: yarn install --frozen-lockfile --ignore-engines
+      - name: Install unixodbc
```
WikiRik, line 97: I believe this can be removed, we should be able to do the same as with [test-snowflake](https://github.com/sequelize/sequelize/blob/65a6fe904da3aea896734c41f6d61b9b7d668497/.github/workflows/ci.yml#L274-L291) iirc

package.json
```diff
@@ -247,6 +252,7 @@
     "test-unit-sqlite": "cross-env DIALECT=sqlite yarn test-unit",
     "test-unit-mssql": "cross-env DIALECT=mssql yarn test-unit",
     "test-unit-db2": "cross-env DIALECT=db2 yarn test-unit",
+    "test-unit-ibmi": "cross-env DIALECT=ibmi yarn test-unit",
     "test-unit-snowflake": "cross-env DIALECT=snowflake yarn test-unit",
     "test-unit-all": "yarn test-unit-mariadb && yarn test-unit-mysql && yarn test-unit-postgres && yarn test-unit-postgres-native && yarn test-unit-mssql && yarn test-unit-sqlite && yarn test-unit-snowflake && yarn test-unit-db2",
```
WikiRik, line 257:
```suggestion
    "test-unit-all": "yarn test-unit-mariadb && yarn test-unit-mysql && yarn test-unit-postgres && yarn test-unit-postgres-native && yarn test-unit-mssql && yarn test-unit-sqlite && yarn test-unit-snowflake && yarn test-unit-db2 && test-unit-ibmi",
```

src/dialects/ibmi/data-types.js
```diff
@@ -0,0 +1,331 @@
...
+      super(length, decimals);
+      if (this._decimals) {
+        warn('Db2 for i does not support FLOAT with decimal scale. FLOAT will be used instead. If you want to specify scale, use DECIMAL or NUMERIC data types.');
+        this._length = undefined;
+        this._decimals = undefined;
+      }
+    }
+  }
+
+  class DOUBLE extends BaseTypes.DOUBLE {}
```
WikiRik, line 239: Haven't checked with the other dialects, but can this just be empty?

test/integration/associations/belongs-to.test.js
```diff
@@ -627,9 +627,8 @@ describe(Support.getTestDialectTeaser('BelongsTo'), () => {
     }
 
     // NOTE: mssql does not support changing an autoincrement primary key
-    if (Support.getTestDialect() !== 'mssql'
-        && Support.getTestDialect() !== 'db2') {
-      it('can cascade updates', async function () {
+    if (['mssql', 'db2', 'ibmi'].includes(dialect)) {
+      it.skip('can cascade updates', async function () {
```
WikiRik, line 631: This should still run on the other dialects

src/model.js
```diff
@@ -2725,11 +2726,11 @@ class Model {
         }
       }
 
-      if (options.ignoreDuplicates && ['mssql', 'db2'].includes(dialect)) {
+      if (options.ignoreDuplicates && ['mssql', 'db2', 'ibmi'].includes(dialect)) {
         throw new Error(`${dialect} does not support the ignoreDuplicates option.`);
       }
 
-      if (options.updateOnDuplicate && (dialect !== 'mysql' && dialect !== 'mariadb' && dialect !== 'sqlite' && dialect !== 'postgres')) {
+      if (options.updateOnDuplicate && (dialect !== 'mysql' && dialect !== 'mariadb' && dialect !== 'sqlite' && dialect !== 'postgres' && dialect !== 'ibmi')) {
```
WikiRik, line 2733: Can we make this an includes as well?

WikiRik (approved):
Many changes are looking good, just have a few last remarks

WikiRik (changes requested):
Woops, accidentally approved this.

Many changes are looking good, just have a few last remarks and then we can merge

markdirish:
Thanks @WikiRik . I noticed that some changes I made to the `formatBindParameters` function has caused some unit tests to fail, so I am running some final integration tests to confirm that the unit test expectations need to be changed. Then I will fix the things from your comments and hopefully get this thing landed!

src/dialects/ibmi/data-types.js
```diff
@@ -0,0 +1,331 @@
...
+      return value.toString();
+    }
+
+  }
+
+  class TEXT extends BaseTypes.TEXT {
+    toSql() {
+      let len = 0;
+      if (this._length) {
+        if (isNaN(this._length)) {
```
ephys, line 138: using `isNaN` is banned by our linter (but not run on js files, only ts, as we resolve these issues one by one) because it's a code-smell.

If you want to detect whether something is NaN, use Number.isNaN
If you're trying to detect whether this._length is a number, use `Number.isSafeInteger`
If it's not a safe integer, and it's not one of the 3 accepted strings, and it's not undefined (for default value), I think this should throw.

src/dialects/ibmi/data-types.js
ephys: It will inherit everything, constructors included.
However, is it something we need to do? Or can this line just be removed?

src/dialects/ibmi/index.js
```diff
@@ -0,0 +1,71 @@
...
+const { QueryGenerator } = require('./query-generator');
+const { IBMiQueryInterface } = require('./query-interface');
+const DataTypes = require('../../data-types').ibmi;
+
+class IBMiDialect extends AbstractDialect {
+
+  static supports = _.merge(
+    _.cloneDeep(AbstractDialect.supports),
+    {
+      DEFAULT: true,
```
ephys, line 16: no need to override an option if it's already the same in AbstractDialect (we actually went ahead and cleaned up the other dialects)

src/dialects/ibmi/index.js
```diff
@@ -0,0 +1,71 @@
...
+  constructor(sequelize) {
+    super();
+    this.sequelize = sequelize;
+    this.connectionManager = new ConnectionManager(this, sequelize);
+    this.queryGenerator = new QueryGenerator({
+      _dialect: this,
+      sequelize,
+    });
+    this.queryInterface = new IBMiQueryInterface(this.sequelize, this.queryGenerator);
+    this.supports.autoIncrement.defaultValue = true;
```
ephys, line 55: is there a reason why these settings are overwritten here instead of in `static supports` ?

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+class IBMiQueryGenerator extends AbstractQueryGenerator {
+
+  // Version queries
+  versionQuery() {
+    return 'SELECT CONCAT(OS_VERSION, CONCAT(\'.\', OS_RELEASE)) AS VERSION FROM SYSIBMADM.ENV_SYS_INFO';
+  }
+
+  // Schema queries
+  createSchema(schema) {
+    return `CREATE SCHEMA "${schema}"`;
```
ephys, line 22: This may fail if the schema includes a `"` character

and/or this may let users do things like `createSchema('db"."public')` which would result in `CREATE SCHEMA "db"."public"`

I think we should either escape ", or throw if it's present in the schema string.

This applied to other places where a schema name is inlined too

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+      }
+    }
+
+    return `SELECT DISTINCT SCHEMA_NAME AS "schema_name" FROM QSYS2.SYSSCHEMAAUTH WHERE GRANTEE = CURRENT USER${skippedSchemas}`;
+  }
+
+  // Table queries
+  createTableQuery(tableName, attributes, options) {
+    const primaryKeys = [];
+    const foreignKeys = {};
```
ephys, line 43:
```suggestion
    const foreignKeys = Object.create(null);
```

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+    }
+
+    let attributesClause = attrStr.join(', ');
+    const pkString = primaryKeys.map(pk => this.quoteIdentifier(pk)).join(', ');
+
+    if (options.uniqueKeys) {
+      _.each(options.uniqueKeys, (columns, indexName) => {
+        // if primary keys === unique keys, then skip adding new constraint
+        const uniqueIsPrimary = columns.fields.length === primaryKeys.length && columns.fields.sort().every((value, index) => {
+          return value === primaryKeys.sort()[index];
```
ephys, line 68: `primaryKeys.sort()` is going to be called many times


I think this piece of code would be better extracted to a `shallowEqual` function:
```typescript
columns.fields.sort();
primaryKeys.sort();

shallowEqual(columns.fields, primaryKeys);
```

---

`sort` is going to mutate columns.fields. I am generally opposed to mutating inputs as it often leads to unexpected bugs. (this is something sequelize does a lot but that I'd like to resolve)

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+      }
+    }
+
+    let tableObject = {};
+    if (typeof tableName === 'string') {
+      tableObject.table = tableName;
+    } else {
+      tableObject = tableName;
+    }
+
```
ephys, line 100:
```suggestion
    let tableObject;
    if (typeof tableName === 'string') {
      tableObject = { table: tableName };
    } else {
      tableObject = tableName;
    }

```

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+   @private
+  */
+  addIndexQuery(tableName, attributes, options, rawTablename) {
+    options = options || {};
+
+    if (!Array.isArray(attributes)) {
+      options = attributes;
+      attributes = undefined;
+    } else {
+      options.fields = attributes;
```
ephys, line 332: this mutates the input

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+  }
+
+  addConstraintQuery(tableName, options) {
+    const query = super.addConstraintQuery(tableName, options);
+
+    return query.replace(/;$/, '');
+  }
+
+  _toJSONValue(value) {
+    // true/false are stored as strings in mysql
```
ephys, line 421:
```suggestion
    // true/false are stored as strings in ibmi
```

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+    if (typeof value === 'boolean') {
+      return value.toString();
+    }
+
+    // null is stored as a string in mysql
```
ephys, line 426:
```suggestion
    // null is stored as a string in ibmi
```

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+   * @returns {string}         The generated sql query.
+   * @private
+   */
+  addLimitAndOffset(options) {
+    let fragment = '';
+    const offset = options.offset;
+    const limit = options.limit;
+
+    if (offset) {
+      if (typeof offset === 'number' && Number.isInteger(offset)) {
```
ephys, line 517: I'd opt for `Number.isSafeInteger` instead, as out-of-representable-range integers can behave in weird ways:

image

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+        fragment += ` OFFSET ${offset} ROWS`;
+      } else {
+        console.warn('"offset" must be an integer. Offset is not added');
```
ephys, line 520: I would throw here instead

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+      'CONSTRAINT_SCHEMA AS "constraintSchema",',
+      'CONSTRAINT_TYPE AS "constraintType",',
+      'TABLE_NAME AS "tableName",',
+      'TABLE_SCHEMA AS "tableSchema"',
+      'from QSYS2.SYSCST',
+      `WHERE table_name='${tableName}'`,
+    ].join(' ');
+
+    if (constraintName) {
+      sql += ` AND CONSTRAINT_NAME = '${constraintName}'`;
```
ephys, line 600: most things here are not escaped, is that intended?

src/dialects/ibmi/query-generator.js
```diff
@@ -0,0 +1,836 @@
...
+      if (attribute.onUpdate && attribute.onUpdate.toUpperCase() !== 'CASCADE') {
+        template += ` ON UPDATE ${attribute.onUpdate.toUpperCase()}`;
+      }
+    }
+
+    return template;
+  }
+
+  attributesToSQL(attributes, options) {
+    const result = {};
```
ephys, line 730:
```suggestion
    const result = Object.create(null);
```

src/dialects/ibmi/query-interface.js
```diff
@@ -0,0 +1,65 @@
...
+
+ @class QueryInterface
+ @static
+ @private
+ */
+
+class IBMiQueryInterface extends QueryInterface {
+
+  startTransaction(transaction, options) {
+    if (!transaction || !(transaction instanceof Transaction)) {
```
ephys, line 18: could be simplified to just

```suggestion
    if (!(transaction instanceof Transaction)) {
```

src/dialects/ibmi/query-interface.js
```diff
@@ -0,0 +1,65 @@
...
+    return transaction.connection.beginTransaction();
+  }
+
+  commitTransaction(transaction) {
+    if (!transaction || !(transaction instanceof Transaction)) {
+      throw new Error('Unable to commit a transaction without transaction object!');
+    }
+
+    if (transaction.parent) {
+      // Savepoints cannot be committed
```
ephys, line 35: I think it would be better to throw when someone tries to create a savepoint instead of silently ignoring them. This is the kind of think I think users should be aware of

src/dialects/ibmi/query.js
```diff
@@ -0,0 +1,291 @@
...
+      id = id || results && results[autoIncrementAttribute];
+      id = id || metaData && metaData[autoIncrementAttribute];
+
+      this.instance[this.model.autoIncrementAttribute] = id;
+    }
+  }
+
+  handleShowIndexesQuery(data) {
+
+    const indexes = {};
```
ephys, line 200:
```suggestion
    const indexes = Object.create(null);
```

src/model.js
```diff
@@ -2725,11 +2726,11 @@ class Model {
         }
       }
 
-      if (options.ignoreDuplicates && ['mssql', 'db2'].includes(dialect)) {
+      if (options.ignoreDuplicates && ['mssql', 'db2', 'ibmi'].includes(dialect)) {
```
ephys, line 2729: maybe out of scope for this PR but I'd prefer if we had a flag on Dialect.supports instead of hardcoding dialects here

src/model.js
ephys, line 2733: maybe out of scope for this PR but I'd prefer if we had a flag on Dialect.supports instead of hardcoding dialects here

src/sequelize.js
```diff
@@ -300,6 +300,10 @@ class Sequelize {
       throw new Error('Setting a custom timezone is not supported by SQLite, dates are always returned as UTC. Please remove the custom timezone parameter.');
     }
 
+    if (this.options.dialect === 'ibmi' && this.options.timezone !== '+00:00') {
```
ephys, line 303: maybe out of scope for this PR but I'd prefer if we had a flag on Dialect.supports instead of hardcoding dialects here

src/sequelize.js
```diff
@@ -900,7 +907,7 @@ class Sequelize {
       ...options,
     };
 
-    await this.query('SELECT 1+1 AS result', options);
+    await this.query(`SELECT 1+1 AS result${this.options.dialect === 'ibmi' ? ' FROM SYSIBM.SYSDUMMY1' : ''}`, options);
```
ephys, line 910: can we add a method on queryInterface (queryGenerator?) to generate this query from the code that's controlled by the dialect instead?

ephys (changes requested):
I haven't reviewed the tests yet but here's the first part of my review

in general, I'd like to minimise as much as possible having dialect-specific code outside of the dialect folder as we're trying to split the dialects from core

It's not really possible everywhere yet but I noted a few places where I think it could be done

src/dialects/ibmi/query-generator.js
markdirish: Will cause https://github.com/sequelize/sequelize/blob/main/test/unit/sql/offset-limit.test.js#L76 to fail. Every other dialect (excepting `db2`) seems to just accept the input, rather than even do a type check.

src/dialects/ibmi/query-generator.js
markdirish: Not sure the best way to handle escaping. The only escaping that takes place is the any internal single-quote is matched without another single quote, so that `'this ' is my value'` becomes `'this '' is my value'`. I'm not sure whether these things should be left to the user, or if `sequelize` should take over.

src/dialects/ibmi/query-generator.js
markdirish: I think on IBM i, schema names might only be able to be alphanumeric. Will look into this further.

markdirish:
Pushed a commit with all of the easy victories, still some work to be done.

src/dialects/ibmi/connection-manager.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+
+const AbstractConnectionManager = require('../abstract/connection-manager');
```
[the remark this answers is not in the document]
markdirish: I could try, I'm not sure how quick it would be done though and I'm afraid of it delaying this PR for another few weeks.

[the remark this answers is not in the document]
ephys: That's definitely a valid concern and I won't block this PR for this
Would you be willing to work on it in a follow-up PR?

src/dialects/ibmi/query-generator.js
ephys: Thanks :)
I think that's something we should look into changing in v7 but we can do that in another PR

src/dialects/ibmi/query-generator.js
ephys: In general users shouldn't have to think about escaping (except when using literals), sequelize should handle it in its own literals

We've had cases of users exploiting parts of queries not being escaped, and having their code break in other places where their input was escaped

I won't block the PR for this but if you want to do a round of escaping in a follow-up PR, it'd be welcomed :)

src/dialects/ibmi/connection-manager.js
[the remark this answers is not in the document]
markdirish: Absolutely!

src/dialects/ibmi/query-generator.js
markdirish: I tried to create a schema called `SCHE'MA"MAN` by running:
```
CREATE SCHEMA "SCHE'MA""MAN"
```
I received the following error:
```
Name "SCHE'MA""MAN" not allowed. Cause . . . . . :   "SCHE'MA""MAN" contains a character that is not allowed or does not contain any characters. An ordinary identifier must begin with (A-Z, $, #, or @) and be followed by zero or more (A-Z, 0-9, $, #, @, or _). A delimited identifier is a string of characters within SQL escape characters.  The characters allowed in delimited names depends on the type of name. -- The characters between the escape characters for system table names, schema names, package names and other system object names can be any characters except Hex 00-3F, Hex 40 (space), Hex 5C (*), Hex 6F (?), Hex 7D ('), Hex 7F ("), and Hex FF
```

So it appears that a valid schema should never include a `"` character, and if it does, I would expect it to crash. Do you think its best to let the DBMS handle the error, or try and throw our own error? My only fear is at some later date, the characters allowed in a valid schema might change, and in that case we'd have to fix it again.

src/dialects/ibmi/query-generator.js
markdirish: I think I have this fixed now. Creates a copy of `primaryKeys` to sort (one time per table), and creates a copy of `uniqueKeys.fields` to sort (on time per unique key), then does its comparison on those arrays.

markdirish:
Ok... I think I have all of the things fixed that are required for this PR. @WikiRik  and @ephys  if you could point me to any unresolved issues that you need me to fix, I will get them fixed up. Remaining things to follow in a separate PR:
 
 * Look through the tests, and instead of multiple dialect checks use arrays with `.includes`
 * Convert the ibm dialect `.js` files to `.ts` files
 * Add additional `Dialect.supports` keywords to further excise dialect-specific code from the core
 * Escape user-entered values inserted into the database for IBM i query generator

package.json
```diff
@@ -247,15 +252,17 @@
     "test-unit-sqlite": "cross-env DIALECT=sqlite yarn test-unit",
     "test-unit-mssql": "cross-env DIALECT=mssql yarn test-unit",
     "test-unit-db2": "cross-env DIALECT=db2 yarn test-unit",
+    "test-unit-ibmi": "cross-env DIALECT=ibmi yarn test-unit",
     "test-unit-snowflake": "cross-env DIALECT=snowflake yarn test-unit",
-    "test-unit-all": "yarn test-unit-mariadb && yarn test-unit-mysql && yarn test-unit-postgres && yarn test-unit-postgres-native && yarn test-unit-mssql && yarn test-unit-sqlite && yarn test-unit-snowflake && yarn test-unit-db2",
+    "test-unit-all": "yarn test-unit-mariadb && yarn test-unit-mysql && yarn test-unit-postgres && yarn test-unit-postgres-native && yarn test-unit-mssql && yarn test-unit-sqlite && yarn test-unit-snowflake && yarn test-unit-db2 && && test-unit-ibmi",
```
WikiRik, line 257:
```suggestion
    "test-unit-all": "yarn test-unit-mariadb && yarn test-unit-mysql && yarn test-unit-postgres && yarn test-unit-postgres-native && yarn test-unit-mssql && yarn test-unit-sqlite && yarn test-unit-snowflake && yarn test-unit-db2 && test-unit-ibmi",
```

WikiRik (commented):
Haven't taken a very in-depth look to the src/dialects/ibmi files this time but I expect that this should be good for now if the tests pass. We'll take a look again at those when we slowly convert them to .ts

Just the final change in package.json and then I can approve. No need to re-run your tests for that change so no need to postpone this for much longer

package.json
markdirish: Ha, copy and paste error

WikiRik (dismissed):

ephys (approved):
My last comment for this PR:

Some tests are disabled because the dialect doesn't support the necessary feature, I think we should keep them enabled and test that Sequelize throws an Error that warns the user the feature isn't supported in the dialect.

This and my other remaining comments can be sorted out in a follow-up PR :)

PR #14017, opened 2022-01-26
feat(types): make `Model.getAttributes` stricter

ephys:
### Pull Request Checklist

### Description Of Change

Follow up to https://github.com/sequelize/sequelize/pull/13728

The typings for `Model.getAttributes` now declare returning an object where the keys are the attributes declared on the model.

```typescript
class User extends Model<InferAttributes<User>> {
  id: number;
  name: string;
}

// before, this would not have been reported as an error:
User.rawAttributes.iDontExist;

// Now, TS reports an error on this:
User.getAttributes().iDontExist;
// but not on this
User.getAttributes().name;
```

This PR deprecates `Model.rawAttributes` & recommends using `Model.getAttributes` instead. I've done this because we cannot strictly type `rawAttributes`.


WikiRik:
Do you want to deprecate this in v7 and remove in v8 or deprecate in v6 and remove in v7?

ephys:
I don't think we'll remove it in v7 because I think it's widely used. Maybe v8

ephys:
I'm really happy with all the TS changes we've been making. It's nice to have an easily strongly typed project

image

ephys:
I'm really happy with all the TS changes we've been making. It's nice to have an easily strongly typed project

image

AllAwesome497:
Hm, a bit biased but the script I made (which I've yet to adapt for general sequelize usage) uses `rawAttributes` as a source of truth to generate typing since its almost always going to be more accurate than TS typing (unless we can get TS to infer everything which might be ideal, though it could have some performance implications with large codebases).

I'd also be hesitant to touch `rawAttributes` or `getAttributes` without first typing the code in `dialects/` since that logic can be extremely hard to grep + varies greatly.

ephys:
Note that `ModelAttributeColumnOptions` is untouched by this PR.
The only thing this changes is restrict the list of attributes to those the user declares. Is there a scenario where that list of keys is not accurately represented by the list of keys of `TAttributes`? (Except the user not properly declaring their attributes obv)

> unless we can get TS to infer everything which might be ideal

I don't think we'll be able to make TS infer everything. I mean we could, using even more Branded Types, but it'd be verbose. I think this is a nice middle ground, especially since typing for `rawAttributes`/`getAttributes` can still be overridden by your script.

sdepold:
I think I like it :)

PR #14018, opened 2022-01-26
fix(types): correct typing definitions for `Sequelize.where`

ephys:
### Pull Request Checklist

### Description Of Change

The typings for `Sequelize.where(leftOperand, operator, rightOperand)` were very incorrect:

- It stated that the left operand accepted strings, but that is not the case. Using a string as the left operand will crash.
- The typings for `operator` accepted strings. It works but users should use Op symbols. I can revert this change but the typing would not be great.
- The typings for `rightOperand` were all over the place. I've now delegated them to `WhereOperators` so `where` accepts the same values as the POJO version.
  i.e. in `where(col('first_name'), Op.iLike, name)`, `name` will be accept the exact same values as `{ firstName: { [Op.iLike]: name } }`.

The signature used to be `where(attribute, comparator, logic)` but I've renamed it to `where(leftOperand, operator, rightOperand)` which is much closer to reality as the operands can be much more than just an attribute.

I've also added tests for things the JSDoc says are supported but were not tested.

I have not updated the typings for `lib/utils/Where` as it's internal and the changes I want to make would impact a lot more files - but I'd like to refactor it for v7 in https://github.com/sequelize/sequelize/pull/13782


sdepold (dismissed):
Tbh... I didn't even know that you can pass anything other than an object to where 🙈 so you have my trust

ephys:
It's for those couple use-cases where you need to use `fn` on your column :)

Before merging, I'll re-add `string` as a valid operator. I don't want to add a breaking change in v6.
We'll see in https://github.com/sequelize/sequelize/tree/ephys/new-where-builder in the future if we want to remove it

AllAwesome497 (approved):
I haven't used `where` as a function personally, but the types look fine and I trust your judgement so lgtm.

ephys:
Thanks :) Here we go then

Issue #13753, opened 2021-12-09
Problem on sequelize.where function with correctly placing % operators.


amitrana-dev:
## Issue Creation Checklist

[x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description
Sequalize.where isn't placing % operators on startsWith, endsWith & substring operations.

### SSCCE

**Here is the link to the SSCCE for this issue:** [#202](https://github.com/papb/sequelize-sscce/pull/202) 

### JS Code
```
const Foo = sequelize.define('Foo', { name: DataTypes.TEXT });
await Foo.create({ name: 'foo' })
Foo.count({where: { [Op.and]: [ Sequelize.where(Sequelize.col('name'),Op.startsWith,'fo') ] } })
```
### What do you expect to happen?

Expected query should be:

`SELECT count(*) AS count FROM Foos AS Foo WHERE (name LIKE 'fo%');`

### What is actually happening?

Actual query generated is:

`SELECT count(*) AS count FROM Foos AS Foo WHERE (name LIKE 'fo');`

### Environment

- Sequelize version: 6.11.0 
- Node.js version: v14.17.0 

### How does this problem relate to dialects?

- [ ] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [x] I don't know, I was using mysql, with connector library version mysql2@2.3.3 and database version 5.7.29

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [x] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

amitrana-dev:
Up

Issue #13602, opened 2021-10-27
`Op.not` operator is ignored when used with `Sequelize.where()`


ephys:
## Issue Creation Checklist

- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description

the `NOT` operator is ignored when used with `Sequelize.where` like in [this exemple](https://github.com/ephys/sequelize-sscce/blob/master/src/sscce.js#L34)

### SSCCE

**Here is the link to the SSCCE for this issue:** https://github.com/papb/sequelize-sscce/pull/194

### What do you expect to happen?

`{ [Op.not]: Sequelize.where(col, op, value) }` generates as as `NOT(col op value)`

### What is actually happening?

The generated query is lacking a NOT operator: `{ [Op.not]: Sequelize.where(col, op, value) }` generates as as `col op value`

### Additional context

Add any other context and details here.

### Environment

- Sequelize version: 6.8.0
- Node.js version: 14.18.1

### How does this problem relate to dialects?

- [x] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?

- [x] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

ephys:
Still an issue

ephys:
not stale :)

madguy02:
@WikiRik can i start looking into this issue? If you haven't taken it up already?

WikiRik:
It would be great if you could look into this issue, thanks!

Issue #13773, opened 2021-12-14
Add `Sequelize.attr`


ephys:
I've completely dropped the previous RFC. See edits if you're interested in its history.

Here is the new draft:

## `Sequelize.attr`

Like `Sequelize.col` but where `Sequelize.col` is used to reference a column name, `Sequelize.attr` is used to reference an attribute name.

Because sometimes you need to be able to reference another attribute, and having to specify that other attribute's column name instead is inconsistent as column names are abstracted away almost everywhere else.

```typescript
// instead of having to write
User.findOne({
  where: {
    birthDate: { [Op.gt]: Sequelize.col('register_date') },
  },
})

// you'd write

User.findOne({
  where: {
    birthDate: { [Op.gt]: Sequelize.attr('registerDate') },
  },
})
```

`Sequelize.attr` would follow the same syntax as plain attribute references. Meaning it would support:

- `Sequelize.attr('$association.value$')`
- `Sequelize.attr('json.path')`

ephys:
> I fear most users don't know there is a difference between the columns & attributes.

Thinking about this more I think that's actually an argument for using `.attr` instead of `.col`. Would help devs understand the difference between the two.

So the methods would be:

- `Sequelize.rawCol(col)` or `Sequelize.col(col)`
- `Sequelize.attr(Model, attr)
- `Model.attr(attr)`

fncolon:
> > I fear most users don't know there is a difference between the columns & attributes.
> 
> Thinking about this more I think that's actually an argument for using `.attr` instead of `.col`. Would help devs understand the difference between the two.
> 
> So the methods would be:
> 
> * `Sequelize.rawCol(col)` or `Sequelize.col(col)`
> * `Sequelize.attr(Model, attr)`
> * `Model.attr(attr)`
> 
> If we go with `.col` instead of `.rawCol`, I can add this to v6 instead of v7

```
We add attributes to entities; we add columns to tables.
```

IMHO, I prefer using `col` instead of `attr`

ephys:
> We add attributes to entities; we add columns to tables.

Yep!

> IMHO, I prefer using col instead of attr

I find it inconsistent that half of my query is using the attribute name and the other half is using the column name. But being able to use the column name is important too so I'm not planning on removing `col` either :)

fncolon:
> > We add attributes to entities; we add columns to tables.
> 
> Yep!
> 
> > IMHO, I prefer using col instead of attr
> 
[...]

PR #14020, opened 2022-01-26
feat: rewrite the part of QueryGenerator responsible for `WHERE`

ephys:
Large WIP PR to completely rewrite how `WHERE` is generated.

This PR is dependent on https://github.com/sequelize/sequelize/pull/14022, as unit tests & types must be improved first to ensure this PR doesn't break anything.

It's definitely not ready yet. A lot of tests are still failing.

The bulk of changes are in https://github.com/sequelize/sequelize/blob/ephys/new-where-builder/lib/dialects/abstract/where-sql-builder.ts

The new system is much more type-safe, with each operator being typed individually.

With this new system, it's also easier for dialects to override how an operator is built. They can override the method specific to that operator without needing to modify the other.

Closes https://github.com/sequelize/sequelize/issues/13602
Closes https://github.com/sequelize/sequelize/issues/13753

Goals of this PR:

- Migrate the implementation of `where` to TypeScript
- Remove all differences between the implementations of `Sequelize.where()` and pojo where objects (i.e. `where: { firstName: 'zoé' }`). Fixes https://github.com/sequelize/sequelize/issues/13602
- Remove behavior differences between `{ attribute: value }` and `{ attribute: { [Op.eq]: value } }`
- Prepare the groundwork for a follow-up PR that implements [`Sequelize.attr`](https://github.com/sequelize/sequelize/issues/13773) as a attributeName-aware alternative to `Sequelize.col`
- Prepare the groundwork for a follow-up PR that implements bind parameters in `Sequelize.literal` (eg. `Sequelize.literal('name = ?name', { name: 'my name' })`)
- Accept all iterables in operators that do not explicitely need an array (eg. `{ attribute: { [Op.in]: new Set([1,2,3]) } }`)
- Be more strict: An invalid `where` object should throw with a message describing what is invalid, instead of crashing with an obscure message.
- Support using `or` inside another operator: `{ id: { [Op.eq]: or(1, 2) } }` should function

TODO

- Investigate supporting any value on both sides of `where`: `Sequelize.where([], Op.eq, [])`
- Ensure all queries in `types/where.ts` are actually tested
- Ensure `Op.any` and `Op.all` are supported in all/most operators
- Check if https://github.com/sequelize/sequelize/issues/14114 is fixed
- Support `literal` in `Op.match`: https://github.com/sequelize/sequelize/issues/14051
- fix https://github.com/sequelize/sequelize/discussions/15290#discussioncomment-4163339


ephys:
If there are other open issues related to `WHERE` being incorrectly generated, don't hesitate to post them here so I can include them in the new implementation

WikiRik:
Looking at the goals of this PR I don't really see any breaking changes. Is that correct? Since I do not think we'll land this in the initial release of v7 but postponing it to v8 would also be a waste

ephys:
There is one breaking change: Writing the following is no longer supported

```ts
import { where } from '@sequelize/core';

User.findAll({
  where: where(col("a"), '=', col("b"))
})
```

instead users need to write the following:

```ts
import { where } from '@sequelize/core';

User.findAll({
  where: where(col("a"), Op.eq, col("b"))
});
```

This is because `where` needs to know which operator is used to build the query properly. The left and right operands are not always serialized in exactly the same way, depending on the operator.

If the user needs to write a raw query, they can do this instead:

```ts
import { sql } from '@sequelize/core';

User.findAll({
  where: sql`${col("a")} = ${col("b")}}`
});
```

Well... once I get to it https://github.com/sequelize/sequelize/issues/14299

WikiRik:
No worries about working on it for v8, v7 already has some nice new and improved features that this can wait a bit. Thanks for the information!

ephys:
Replaced by https://github.com/sequelize/sequelize/pull/15598

PR #14036, opened 2022-02-01
build: replace npm with yarn

WikiRik:
### Pull Request Checklist

- [X] Have you added new tests to prevent regressions? _Not applicable_
- [X] Did you update the typescript typings accordingly (if applicable)? _Not applicable_

### Description Of Change

This PR consists of two parts; one is just updating the npm examples and adding yarn equivalents to them. The second is replacing the npm scripts with yarn. I've tested most of these locally and they all keep working.


sdepold (dismissed):
Fine with me. However, there is a chance that the dev environment will not have yarn, right?

WikiRik:
That chance is indeed present, but small. We are not requiring the use of yarn berry, just the classic one. Many people should automatically have that installed and we are using it in the pre-commit hook already.

ephys:
I think it's fine to require the use of `yarn` in the development environment.
Reviewing

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -85,9 +85,9 @@ Add any other context and details here.
 
 ### Environment
 
-- Sequelize version: XXX <!-- run `npm list sequelize` to obtain this -->
+- Sequelize version: XXX <!-- run `npm list sequelize` (or `yarn list --pattern sequelize`) to obtain this -->
 - Node.js version: XXX <!-- run `node -v` to obtain this -->
-- If TypeScript related: TypeScript version: XXX <!-- run `npm list typescript` to obtain this -->
+- If TypeScript related: TypeScript version: XXX <!-- run `npm list typescript` (or `yarn list --pattern typescript`) to obtain this -->
```
ephys, line 90: What about `npm list typescript` / `yarn list --pattern typescript` (slash instead of or + parenthesis) to put them on equal levels

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,7 +11,7 @@ If your PR only contains changes to documentation, you may skip the template bel
 _Please make sure to review and check all of these items:_
 
 - [ ] Have you added new tests to prevent regressions?
-- [ ] Does `npm run test` or `npm run test-DIALECT` pass with this change (including linting)?
+- [ ] Does `npm run test` / `yarn test` or `npm run test-DIALECT` / `yarn test-DIALECT` pass with this change (including linting)?
```
ephys, line 14: `npm run test` uses `yarn` internally so the user *has* to have it. We could drop `npm run test` here and only list the yarn version.

I would propose:
- All commands aimed at maintainers only list yarn
- Commands aimed at the users of the library list both yarn & npm, separated by a slash

docs/manual/core-concepts/getting-started.md
```diff
@@ -7,18 +7,29 @@ In this tutorial you will learn to make a simple setup of Sequelize.
...
 
 ```sh
-# One of the following:
-$ npm install --save pg pg-hstore # Postgres
-$ npm install --save mysql2
-$ npm install --save mariadb
-$ npm install --save sqlite3
-$ npm install --save tedious # Microsoft SQL Server
+# using npm
+$ npm i pg pg-hstore # PostgreSQL
```
ephys, line 20: while we're changing these lines, can we remove the `$` character, like in the readme? (makes copy-paste harder)

ephys (approved):

docs/manual/other-topics/migrations.md
```diff
@@ -20,7 +23,7 @@ For details see the [CLI GitHub repository](https://github.com/sequelize/cli).
 
 To create an empty project you will need to execute `init` command
 
-```text
+```sh
 npx sequelize-cli init
```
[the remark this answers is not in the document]
WikiRik: If sequelize-cli is not installed this won't work, but I can check if it works if sequelize-cli is installed.

docs/manual/core-concepts/getting-started.md
WikiRik: I wasn't sure why they were here so I left them. But I will remove these

.github/PULL_REQUEST_TEMPLATE.md
WikiRik: Oh right, that's a good one. I will update the PR accordingly

docs/manual/other-topics/migrations.md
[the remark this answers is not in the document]
ephys: Ah yes I forgot to write that part of my comment
It's only valid because we tell the user to run `yarn add sequelize-cli --dev` a couple lines above
Still it's good to double check :)

WikiRik:
Marked as draft until I implement the requested changes. Will probably get to that on Friday

ephys (approved):
Looks good to me

bl0cknumber:
This is nice addition!

mariusa:
Hi, just curios for the motivation to drop npm, which is bundled with node?

JustFly1984:
Yarn also bundled with latest node

WikiRik:
@mariusa we're only dropping yarn for developer commands. Users of sequelize are still able to use npm.
yarn is quicker overall and the lockfile is nicer.

PR #14059, opened 2022-02-06
build(ci): don't run tests twice on v6 PRs

ephys:
CI PR to stop the CI from testing PRs targetting v6 twice.

This is going to trigger a release. I don't know if the release will be aborted since the only change is tagged a build.


WikiRik (approved):
We can always cancel the checks before it gets to the release step

ephys:
I've done just that :) https://github.com/sequelize/sequelize/actions/runs/1802435371

PR #14063, opened 2022-02-07
feat: gen /lib & /types from /src & drop /dist (v6)

### LINKED ISSUE #14038: sequelize types import files outside of 'dist' dir, causing downstream tsc issues
## Issue Creation Checklist

## Bug Description

Typescript builds are failing on projects using sequelize when the typescript project has certain strict typechecking options enabled in its `tsconfig.json` file.

The root cause is the changes to the `types/index.d.ts` file introduced in PR #14002 (though there are other similar errors in the index.d.ts file that likely need to be addressed as well).

Additional info:
https://github.com/sequelize/sequelize/commit/3f7588014b93b550acd17fc1163c60661b417524
https://github.com/sequelize/sequelize/commit/5e932434ff14eb9b1fd83f344106900e4aabe852

This change uses `../` imports to reference files for their types, but since the index.d.ts file ends up top-level in the `dist` folder during a build, this effectively causes the build output to refer to `<module root>/dist/../lib/errors` (for example), which resolves to `<module root>/lib/errors`. This causes unbuilt `ts` files to be referenced, ultimately resulting in downstream projects failing their `tsc` builds when settings such as the following are set in their `tsconfig.json`:

```jsonc
"noImplicitOverride": true,
"strictNullChecks": true,
"exactOptionalPropertyTypes": true,
```

When all type checking guards available in the tsconfig.json file are turned on, here is the exhaustive list of errors that occur after importing the module into a downstream project:

```
node_modules/sequelize/lib/errors/aggregate-error.ts:18:3 - error TS4114: This member must have an 'override' modifier because it overrides a member in the base class 'BaseError'.

18   toString(): string {
     ~~~~~~~~

node_modules/sequelize/lib/errors/database/exclusion-constraint-error.ts:12:7 - error TS2420: Class 'ExclusionConstraintError' incorrectly implements interface 'ExclusionConstraintErrorOptions'.
  Types of property 'constraint' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.

12 class ExclusionConstraintError extends DatabaseError implements ExclusionConstraintErrorOptions {
         ~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/sequelize/lib/errors/database/exclusion-constraint-error.ts:23:27 - error TS2379: Argument of type '{ stack: string | undefined; }' is not assignable to parameter of type 'ErrorOptions' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Types of property 'stack' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.

23     super(options.parent, { stack: options.stack });
                             ~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/sequelize/lib/errors/database/foreign-key-constraint-error.ts:32:27 - error TS2379: Argument of type '{ stack: string | undefined; }' is not assignable to parameter of type 'ErrorOptions' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Types of property 'stack' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.

32     super(options.parent, { stack: options.stack });
                             ~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/sequelize/lib/errors/database/unknown-constraint-error.ts:12:7 - error TS2420: Class 'UnknownConstraintError' incorrectly implements interface 'UnknownConstraintErrorOptions'.
  Types of property 'constraint' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.

12 class UnknownConstraintError extends DatabaseError implements UnknownConstraintErrorOptions {
         ~~~~~~~~~~~~~~~~~~~~~~

node_modules/sequelize/lib/errors/database/unknown-constraint-error.ts:23:27 - error TS2379: Argument of type '{ stack: string | undefined; }' is not assignable to parameter of type 'ErrorOptions' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Types of property 'stack' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.

23     super(options.parent, { stack: options.stack });
                             ~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/sequelize/lib/errors/validation-error.ts:224:42 - error TS2532: Object is possibly 'undefined'.

224     } else if (this.errors.length > 0 && this.errors[0].message) {
                                             ~~~~~~~~~~~~~~

node_modules/sequelize/lib/errors/validation/unique-constraint-error.ts:31:44 - error TS2379: Argument of type '{ stack: string | undefined; }' is not assignable to parameter of type 'ErrorOptions' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Types of property 'stack' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.

31     super(options.message, options.errors, { stack: options.stack });
                                              ~~~~~~~~~~~~~~~~~~~~~~~~


Found 8 errors.
```

I suspect the type definitions need to be updated to not refer to any files outside of the `types` source folder.

I can confirm locally that by adjusting these `../` entries to instead be `./` in the dist folder resolves the issues, but I am not knowledgable to know if this translates well to the source files themselves. It looks like it wouldn't work without some additional type files in the `types` directory.

### To recreate yourself:

1. create a new node project
2. install typescript
3. run `npx tsc --init`
4. modify the tsconfig.json file to include stricter type checking as per the above
5. create a basic index.ts file with the following content:
    ```ts
    import { Sequelize } from 'sequelize'
    console.log(Sequelize);
    ```
6. run `npx tsc && node src/index.js`

### Environment

- Sequelize version: 6.15.0
- Node.js version: 16.13.2
- If TypeScript related: 4.5.5

- [x] Yes, but only if guidance for the ideal solution were provided and if I would not have to go through the extensive setup instructions for running unit tests in the contribution guide, since this issue is not related to actual functionality and is instead a build/packaging issue.

ephys:
Closes #14038

This is a pretty big PR that changes how we build sequelize.
Hopefully this is the last post TS migration fix PR.

During the initial TS migration, `/lib` was changed from a pure JS folder to a mix of JS and TS. `/dist/lib` became the "new" `/lib`.
This already broke things so we told node to map `/lib` to `/dist/lib` using 'exports' in package.json.

We still had issues where users would sometimes manage to import their typings from `/lib` instead of `/types` so we deduplicated typings to ensure users would always import the right typing file.

This lead to a subsequent issue where, because files in `/lib` were typescript files and not typescript *declaration* files, typescript would complain if the code did not match how the user configured their `tsconfig.json`

So now, I'm opening a pretty drastic PR: The goal is to allow .d.ts to reference our new .ts files without having to publish any .ts file.

- `/lib`, `/types` have been merged together in `/src`
- `/lib` and `/types` are generated using `yarn build`. They contain respectively only pure JS and the ts declarations. This matches as closely as I could what we had before the TS migration.
- `/dist` is gone ~(which should be fine because it was already impossible to import directly due to `exports` in package.json. It had to be imported as "sequelize/lib")~
- Only `/lib`, and `/types` folders are published to npm
- `exports` in package.json reflects the changes so imports should not be broken.

---

This PR targets v6.

Unfortunately, this is going to break many/all PRs once added to main. :(
I'll open a PR to do the same in main once this one is merged.


ephys:
I accidentally ran eslint on .d.ts files it looks like. I can revert those changes if you want.

src/utils/set-required.d.ts
```diff
@@ -1,16 +1,16 @@
...
-	// If `InferredType` extends the previous, then for each key, use the inferred type key.
-	infer InferredType
-		? {[KeyType in keyof InferredType]: InferredType[KeyType]}
+/**
+ * Full credits to sindresorhus/type-fest
+ * 
+ * https://github.com/sindresorhus/type-fest/blob/v0.8.1/source/set-required.d.ts
+ * 
+ * Thank you!
+ */
```
ephys, line 7: This file was using the wrong line returns

.gitignore
```diff
@@ -14,4 +14,5 @@ test/binary/tmp/*
 .vscode/
 esdoc
 node_modules
-dist
+/lib
```
WikiRik, line 17: Why `/lib`? `lib` should also be enough right?

src/utils/set-required.d.ts
```diff
@@ -1,16 +1,16 @@
...
+export type SetRequired<BaseType, Keys extends keyof BaseType = keyof BaseType> =
+	// Pick just the keys that are not required from the base type.
+	Pick<BaseType, Exclude<keyof BaseType, Keys>> &
+	// Pick the keys that should be required from the base type and make them required.
+	Required<Pick<BaseType, Keys>> extends
+	// If `InferredType` extends the previous, then for each key, use the inferred type key.
+	infer InferredType
+		? {[KeyType in keyof InferredType]: InferredType[KeyType]}
 		: never;
```
WikiRik, line 16: If we're already updating this file it would be nice to end on an empty line

WikiRik (approved):
Two small remarks but looks good. Too bad we can't really do a v6-beta release to let people test this out, so we'll just have to make a new release with this. Should we rename the PR to be a feature?

.gitignore
WikiRik: Ah, that make sense. Didn't know that was an option

sdepold:
Since I prefer src over lib anyway I'm a big fan if this change 😅

sdepold:
Shouldn't we have the same in v7?

ephys:
I plan on opening a follow up PR to cherry pick in main (which I assume we can easily merge in v7 too? I don't know the extend to which main and v7 differ). Starting with v6 felt safer for some reason

src/utils/set-required.d.ts
ephys: Here you go :)

sdepold:
Oh v7 is just a copy of main. So if we target main, we can eventually just merge it into v7

ephys:
Here we go. I hope this is the end of this painful situation :/

PR #14068, opened 2022-02-08
feat: gen /lib & /types from /src & drop /dist (main)

ephys:
This PR cherry-picks PR #14063 into the main branch


ephys:
Given most changes are simple file renames, I'm hoping to see a minimum amount of conflict in existing PRs

src/query.d.ts
```diff
@@ -0,0 +1,3 @@
+// TODO [>=7]: This is a legacy file to avoid introducing a breaking change in v6. Remove me in v7.
```
WikiRik, line 1: Let's remove this file

WikiRik (approved):
Not sure of the legacy file can be removed without other changes so that might be for a next PR. Can you also make a PR soon for linting the additional folders?

src/query.d.ts
ephys: It can be, removed

ephys:
Didn't notice the linter was failing. It's fixed now

.eslintrc.js
```diff
@@ -140,6 +140,13 @@ module.exports = {
       'babel/no-invalid-this': 'off',
       'func-names': 'off',
       'import/order': 'off',
+
```
ephys, line 143: This was in `test/.eslintrc.json`, I moved it here to have better control over it

ephys:
I'll cherry pick https://github.com/sequelize/sequelize/pull/14073 in this PR once it's merged

WikiRik (commented):
One small suggestion, but otherwise we are good

.eslintrc.js
```diff
@@ -140,6 +140,13 @@ module.exports = {
...
+      'no-invalid-this': 'off',
+      'no-unused-expressions': 'off',
+      camelcase: 'off',
```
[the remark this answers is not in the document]
ephys: I can't, `quote-props` is set to `as-needed`
We can change it to [`consistent-as-needed`](https://eslint.org/docs/rules/quote-props#options) but it also means adding a key that needs quotes would cause a larger diff in PRs due to needing to update all the other keys in that object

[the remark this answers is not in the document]
WikiRik: Nevermind then, no need to add additional rules that take more time

WikiRik (dismissed):

WikiRik (approved):

ephys:
And I'll need to dismiss it again, turns our that `Sequelize.version` was untested *and* untyped :o

image

Issue #12741, opened 2020-10-07
removeColumn on sqlite removes foreign key constraints from table


martineso:
## Issue Description
When using Sequelize along with an sqlite db to remove a column via `queryInterface.removeColumn` the foreign key constraints from the table are removed. 

### What are you doing?
I am doing a migration and want to remove a column from a table. When doing so, the result table does not have any foreign key constraints which were put on the original table.

Given the schema below
```sql
CREATE TABLE IF NOT EXISTS `User` (
  `id` UUID PRIMARY KEY, 
  `name`  VARCHAR(255) NOT NULL
);

CREATE TABLE IF NOT EXISTS `Actor` (
  `id` UUID PRIMARY KEY, 
  `name` VARCHAR(255) NOT NULL, 
  `userId` UUID NOT NULL REFERENCES "User" ("id") ON DELETE RESTRICT ON UPDATE CASCADE
);
```
Say I am doing a migration to remove the `name` column from **Actor** schema as it is redundant

```js

 function up(queryInterface, Sequelize) {
    return queryInterface.removeColumn('Actor', 'name');
}

```

### What do you expect to happen?

It should keep the foreign key constraint on the **Actor** table.

### What is actually happening?

The foreign key constraints are removed from the **Actor** table and there is no reference to the **User** table after this query is done.

### Additional context

I have looked through the source code of Sequelize for the sqlite driver and mainly at the `query-generator.js` [(line)](https://github.com/sequelize/sequelize/blob/febc083adee2cd3e6f24b18a556acdc4c4f50f96/src/dialects/sqlite/query-generator.js#L376) where it seems that the approach taken is to create a backup table and then recreate the original one without the aforementioned column. This is fine however there is one thing missing and that is recreating the original table does not keep any foreign key constraints as mentioned in the sqlite docs [here](https://sqlite.org/lang_altertable.html). This is a major problem but can be fixed without too much hassle. There is another big issue with the current approach - since foreign key constraints are not disabled before starting the process for removing a column, when the table is dropped, all foreign key constraints with ON DELETE CASCADE will delete records in the referenced table. This is a major issue which I am facing right now.

### Environment

- Sequelize version: 5.19.1 
- Node.js version: v12.18.2 
- Operating System: Ubuntu 20.04.1 LTS
### How does this problem relate to dialects?

- [ ] I think this problem happens regardless of the dialect.
- [ x ] I think this problem happens only for the following dialect(s): sqlite
[...]

PR #14070, opened 2022-02-08
fix(sqlite): Fix removeColumn and changeColumn foreign key issues for the sqlite dialect

martineso:
### Pull Request Checklist

- [?] Did you update the typescript typings accordingly (if applicable)?
  There is missing type definition for the return type of `queryInterface.tableDescription`. I have not updated it though since it was missing other type information before I have made any changes and I did not want to mess up any tests.

### Description Of Change

This PR addresses the following issue #12741. It also includes a fix for `queryInterface.changeColumn` of the **sqlite** dialect.

### Todos

- [ ] Maybe add TS definitions for `queryInterface.tableDescription` for **sqlite** only?


martineso:
Can someone take a look at the failing integration tests and if they are due to a change I have made? I have made sure to only touch the tests related to **sqlite** however I may have caused a problem with other dialects unknowingly and I need a second pair of eyes to check. Thanks!

WikiRik:
They're failing because they were only intended for sqlite. If you move the check for sqlite up by two tests you should be in the green again (except for DB2 probably but we have very flaky tests in those)

martineso:
I have resolved the issue with ```changeColumn.test.js``` for sqlite and I merged the latest **main**. 
There is however an issue with the ```transaction.test.js``` where for some reason the sqlite dialect fails for 
```
1) [SQLITE] Transaction
```
I noticed this test is marked as flaky in this issue #14161.
I am certain that the changes I've made are not related to this issue since the test is failing in **main** branch as well.

WikiRik:
Yes, that failing test is not due to your PR. We'll review this soon

src/dialects/sqlite/query-interface.js
```diff
@@ -23,11 +23,41 @@ class SQLiteQueryInterface extends QueryInterface {
     const fields = await this.describeTable(tableName, options);
     delete fields[attributeName];
 
-    const sql = this.queryGenerator.removeColumnQuery(tableName, fields);
-    const subQueries = sql.split(';').filter(q => q !== '');
+    const savepointName = 'remove_column_savepoint';
```
ephys, line 26: we should prefix our internal savepoints to be somewhat safe.
Or use a uuid, or a prefixed uuid. e.g. `sequelize_3958bbe4-284a-4d2f-9c5c-b136adc95774`

src/dialects/sqlite/query-interface.js
```diff
@@ -23,11 +23,41 @@ class SQLiteQueryInterface extends QueryInterface {
...
 
-    for (const subQuery of subQueries) {
-      await this.sequelize.query(`${subQuery};`, { raw: true, ...options });
+    // Disable foreign key constraints for the next sequence of operations
+    await this.sequelize.query('PRAGMA foreign_keys = OFF;', options);
+    // Use SAVEPOINT to allow for nested transactions
+    await this.sequelize.query(`SAVEPOINT ${savepointName};`, options);
```
ephys, line 31: if this fails for any reason (such as network issue). `PRAGMA foreign_keys` will not be reverted.

There are a few places where `PRAGMA foreign_keys` is used. It would be better extracted to a helper function:

```typescript
async withForeignKeysOff(options, cb) {
  await this.sequelize.query('PRAGMA foreign_keys = OFF', options);
  
  try {
    return await cb();
  } finally {
    await this.sequelize.query('PRAGMA foreign_keys = ON', options);
  }
}

// ...

this.withForeignKeysOff(options, () => {
  await this.sequelize.query(`SAVEPOINT ${savepointName};`, options);
  // ...
});
```

src/dialects/sqlite/query-interface.js
```diff
@@ -23,11 +23,41 @@ class SQLiteQueryInterface extends QueryInterface {
...
+
+    try {
+      const sql = this.queryGenerator.removeColumnQuery(tableName, fields);
+      const subQueries = sql.split(';').filter(q => q !== '');
+
+      for (const subQuery of subQueries) {
+        await this.sequelize.query(`${subQuery};`, { raw: true, ...options });
```
ephys, line 38: can these be executed in parallel?

src/dialects/sqlite/query-interface.js
```diff
@@ -44,11 +74,41 @@ class SQLiteQueryInterface extends QueryInterface {
...
 
-    for (const subQuery of subQueries) {
-      await this.sequelize.query(`${subQuery};`, { raw: true, ...options });
+    // Disable foreign key constraints for the next sequence of operations
+    await this.sequelize.query('PRAGMA foreign_keys = OFF;', options);
+    // Use SAVEPOINT to allow for nested transactions
+    await this.sequelize.query(`SAVEPOINT ${savepointName};`, options);
+
+    try {
+      const sql = this.queryGenerator.removeColumnQuery(tableName, fields);
```
ephys, line 85: A lot of code here is shared with `removeColumn`, can they be deduplicated?

ephys (changes requested):
Looking good, I have a few notes :)

src/dialects/sqlite/query-interface.js
martineso: These should be executed sequentially since the order of execution matters.

src/dialects/sqlite/query-interface.js
martineso: I was thinking the same thing while writing the code and I had the following dilemma. 
Extracting the logic to a common function will reduce the overall code however it will increase the cognitive load due to the added abstraction. I may however rethink the way we can extract this logic to a separate function using your suggestion for the ```withForeignKeysOff``` function

src/dialects/sqlite/query-interface.js
martineso: Looks good, I will redo the logic inside ```changeColumn``` and ```removeColum``` using the above abstraction.

src/dialects/sqlite/query-interface.js
martineso: Right, makes sense. 
Is there a helper function in the repo for generating uuids?

ephys: the uuid package is available :)

martineso:
Is there any progress on this PR?

test/integration/query-interface/changeColumn.test.js
```diff
@@ -134,115 +134,116 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
 
+          const describedTable2 = await this.queryInterface.describeTable({ tableName: 'users' });
+          expect(describedTable2.level_id.comment).to.be.equal('FooBar');
+        });
+      }
+    });
+
+    // sqlite has limited ALTER TABLE capapibilites which requires a workaround involving recreating tables.
+    // This leads to issues with losing data or losing foreign key references.
+    // The tests below address these problems
```
ephys, line 246: I think we can run these tests on all dialects (and only disable the dialects that fail)

test/integration/query-interface/changeColumn.test.js
```diff
@@ -361,6 +362,131 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+
+        await this.queryInterface.changeColumn('users', 'name', {
+          type: DataTypes.CHAR,
+          allowNull: false,
+        });
+
+        const constraintsQuery = this.queryInterface.queryGenerator.showConstraintsQuery('users');
+        const [{ sql: usersSql }] = await this.queryInterface.sequelize.query(constraintsQuery, {
+          type: 'SELECT',
+        });
```
ephys, line 409: Why not this?

```suggestion
        const constraintsQuery = this.queryInterface.showConstraint('users');
```

test/integration/query-interface/changeColumn.test.js
```diff
@@ -361,6 +362,131 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+            primaryKey: true,
+            autoIncrement: true,
+          },
+          name: {
+            type: DataTypes.CHAR,
+            allowNull: false,
+          },
+        });
+
+        await this.queryInterface.changeColumn('users', 'name', {
```
ephys, line 401: Can you also (or instead) apply a change to the `level_id` column? (such as update its nullability and/or its name) to make sure it does not lose its reference?

test/integration/query-interface/removeColumn.test.js
```diff
@@ -87,6 +87,129 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
           expect(table).to.not.have.property('email');
         });
       }
+
+      // sqlite has limited ALTER TABLE capapibilites which requires a workaround involving recreating tables.
+      // This leads to issues with losing data or losing foreign key references.
+      // The tests below address these problems
+      if (dialect === 'sqlite') {
```
ephys, line 94: we may as well run this on all dialects, to be sure.

ephys (changes requested):
This PR got lost in limbo for so long because I lost its notification. I'm very sorry :(

Overall this is a really high quality PR, it's great work!
I have a few final questions but this other than that it's ready to go

test/integration/query-interface/changeColumn.test.js
martineso: I have tried the ```this.queryInterface.showConstraint('users');``` approach but it does not work for sqlite. It seems the problem lies within `src/dialects/abstract/query-interface.js#showConstraint` function. The `QueryTypes.SHOWCONSTRAINTS` type returns empty result for the sqlite dialect. This poses a problem however if we are to support all dialects within these new tests. 
I have problems with porting these test cases for MariaDB and MySQL (possibly others) due to using the workaround for sqlite (the one used in the test case before the suggested change). If you have a workaround or a suggestion I would gladly apply it.

test/integration/query-interface/changeColumn.test.js
martineso: I have problems with porting these test cases for other dialects. I have described the problem in another comment below (as a reply to another suggestion of yours). I am not well versed within the sequelize code base so I would love if you can provide a way to port the tests for other dialects.

test/integration/query-interface/changeColumn.test.js
martineso: Would you prefer (1) a different test case for each dialect or (2) a branching logic within the same test depending on the dialect type? 

1. 
```js

if (dialect === 'sqlite) {
 it('sqlite specific version of this test', ...)
} else if (dialect === 'mysql') {
 it('mysql version of the test', ...)
}
...
```
2.
```js
it('some generic test', async () => {
  if (dialect === 'sqlite') {
    expect().some.condition()
  } else {
    expect().some.other.condition()
  }
}

```

martineso: Also there is a 3rd solution, to fix the `showConstraint` functionality for sqlite dialect altough this would be out of the scope of this PR.

ephys: Definitely option 2. However I would also prefer to fix `showConstraint` first, maybe in a parallel PR we can merge first? (if you're willing to help with that too)

martineso: Yup, sure. I will open a new PR with a fix for `showConstraint`. Can you point me to a type definition for a valid return type for this function?

ephys: I ran this locally and I concur that fixing sqlite's showConstraint is going to be very complicated and I don't want to block this PR until that minor issue is fixed

Can you add the following comment above the call to `showConstraintsQuery`?

`// TODO: replace with queryInterface.showConstraint once it lists foreign keys properly for sqlite`

test/integration/query-interface/changeColumn.test.js
```diff
@@ -361,6 +362,131 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+            references: {
+              key: 'id',
+              model: 'level',
+            },
+            onDelete: 'CASCADE',
+            onUpdate: 'CASCADE',
+          },
+        });
+
+        await this.queryInterface.createTable('level', {
```
ephys, line 438: `level` should be created first, it is needed by `users`

ephys (changes requested):

test/integration/query-interface/changeColumn.test.js
ephys: I'm also currently working on `changeColumn` to fix many more bugs, I'll take a look at making these tests run in all dialects in my PR because I first need to fix a bunch more bugs in `showConstraint`, `describeTable` etc

Can you add a `// TODO: run in all dialects` above this condition?

test/integration/query-interface/changeColumn.test.js
martineso: Done!

test/integration/query-interface/changeColumn.test.js
martineso: Added a change for level_id -> changed the nullability.

test/integration/query-interface/changeColumn.test.js
martineso: Alright. I have added a comment as per your suggestion.

ephys (approved):
Good work, thank you for this :)

PR #14081, opened 2022-02-10
refactor(data-types): convert snowflake data-types to TS 

AllAwesome497:
 

### Description Of Change
converts snowflake-specific data-types to TS.


lib/dialects/snowflake/data-types.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import moment from 'moment';
+import { kSetDialectNames } from '../../dialect-toolbox';
+import * as BaseTypes from '../abstract/data-types';
+import type {
+  AcceptableTypeOf,
+  StringifyOptions,
+} from '../abstract/data-types';
+
+// TODO [>6]: Why are we changing abstract data-types???
```
sdepold, line 11: Do we all agree that this will only ever land in v7 and onwards?

sdepold (approved):
Can this be merged? Or do we wait for something?

lib/dialects/snowflake/data-types.ts
WikiRik: Yeah, this PR will not land in v6. But I think the idea of this TODO is that it will be fixed in a next PR but before the release of v7

WikiRik:
> ### Note: This is a child PR of #13799 - it should only be merged into that branch when **both** this pr and that pr have been approved.

@sdepold I think this is what we're waiting on. Splitting the data-types of the dialects in separate PRs will make #13799 easier to review. Further steps to the process are also explained in that PR

sdepold:
I noticed that after writing the comment as well 😊🙈

AllAwesome497:
Closing in favor of #14505.

PR #14091, opened 2022-02-11
feat(types): simplify Model definitions

ephys:
### Pull Request Checklist

### Description Of Change

Pretty large PR, it's a typing breaking change that makes `InferAttributes` the default way of handling attribute typing in v7.

So instead of writing:

```typescript
class User extends Model<InferAttributes<User>, InferCreationAttributes<User>> {}
```

You now write:

```typescript
class User extends Model<User> {}
```

---

Instead of writing:

```typescript
interface UserAttributes {
  id: number;
  name: string;
}

interface UserCreationAttributes extends Optional<UserAttributes, 'id'> {}

interface UserInstance
  extends Model<UserAttributes, UserCreationAttributes>,
    UserAttributes {}

const UserModel = sequelize.define<UserInstance>('User', {
  id: {
    primaryKey: true,
    type: DataTypes.INTEGER.UNSIGNED,
  },
  name: {
    type: DataTypes.STRING,
  }
});
```

You now write:

```typescript
interface IUserModel extends Model<IUserModel> {
  id: CreationOptional<number>;
  name: string;
}

const UserModel = sequelize.define<IUserModel>('User', {
  id: {
    primaryKey: true,
    type: DataTypes.INTEGER.UNSIGNED,
  },
  name: {
    type: DataTypes.STRING,
  },
});
```

---

Extra changes:

- `Sequelize.define` now uses `InferAttributes` too.
- `InferAttributes` & co are purely internal in v7, they are not exported anymore.
- `Model['_attributes']` has been replaced with `Model[AttributeSymbol]` to prevent key collision in child models.
- `Attributes<Model>` now supports ModelStatic too! Usefull when used with `Sequelize.define`

The documentation has been updated to reflect this (as a result it's also much, much shorter)
The breaking change has been documented


ephys:
So this is interesting! The same code that was breaking in https://github.com/sequelize/sequelize/pull/14032 is now breaking here

I updated expect-type to version 0.13.0 and the issue is gone
I guess I found the faulty code after all. I just have no idea which one it was.

WikiRik:
Is this part of one of the RFCs you proposed?

ephys:
It was in my personal wishlist but I never opened an actual RFC for this breaking change
If there is any opposition or concerns, don't hesitate to talk about it in this PR :)

WikiRik:
Ah, I read something about `InferAttributes` somewhere and was just curious. Will read into this more later but shorter in most cases is better so I'm happy about this

ephys:
`InferAttributes` is the type I introduced in v6 a few weeks ago to reduce model typing boilerplate without causing a breaking change (https://github.com/sequelize/sequelize/pull/13909)

Since it looks like it's stable (I can't say how many people have used it yet, but it worked in my different projects), I'm going one step further and implementing that missing breaking change :)

sdepold:
Simpler is always better, hm? :)

```
class User extends Model<User> {}
```

What is this User referring to? Another type? Or is the User class?

docs/manual/other-topics/upgrade-to-v7.md
```diff
@@ -16,6 +25,92 @@ As a result, the manual typings that were formerly best-effort guesses on top of
...
+  declare firstName: string;
+}
+```
+
+You instead need to write it like this:
+
+```typescript
+import { Model, CreationOptional } from 'sequelize';
+
+class User extends Model<User> {
```
sdepold, line 53: Since I have never used this syntax before, I have no clue but I think one would have to call the Model's init function afterwards somehow?

Can you add that to the example?

ephys:
> What is this User referring to? Another type? Or is the User class?

Itself! It generates its attribute typings from its own class

sdepold:
> > What is this User referring to? Another type? Or is the User class?
> 
> Itself! It generates its attribute typings from its own class

Pure magic! I love it!

ephys:
I will draft this until https://github.com/sequelize/sequelize/issues/14125 has been resolved

ephys:
This PR has been in limbo for 2 months but it's blocked by https://github.com/sequelize/sequelize/issues/14302, which itself is blocked by https://github.com/sequelize/sequelize/pull/14280

In the meantime I've replicated its non-blocked changes in https://github.com/sequelize/sequelize/pull/14333, to reduce the scope of this PR

ephys:
I'll revisit this change when the rest of the project is ready for it

Issue #14083, opened 2022-02-10
API reference section link broken for v6 & v7


vidurb:
## Issue Creation Checklist

[x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

### What was unclear/insufficient/not covered in the documentation

The `API Reference` section of the docs is breaking for v7 and v6. v5 and prior work fine.

PR #14094, opened 2022-02-11
build: generate jsdoc from src instead of lib

ephys:
### Pull Request Checklist

Fixes https://github.com/sequelize/sequelize/issues/14083


WikiRik (approved):
Too bad about the line endings, but looks good. We need this for v6 as well right?

ephys:
Ah yep, the line ending change is unfortunate. Looks like we still have a couple of files that are using windows-style line endings

I'll open a second PR for v6 with these changes cherry-picked

PR #14095, opened 2022-02-11
test: cover utils/check with tests

sdepold:
Adding some more tests for the utils' check file. Inspired by #13992 .

### DIFF
--- a/test/unit/utils/check.test.ts
+++ b/test/unit/utils/check.test.ts
@@ -1,5 +1,13 @@
-const { expect } = require('chai');
-const { isColString } = require('sequelize/lib/utils/check');
+import { expect } from 'chai';
+import Sequelize, { Utils } from 'sequelize';
+
+const {
+  isPrimitive,
+  isColString,
+  canTreatArrayAsAnd,
+  defaultValueSchemable,
+  isWhereEmpty,
+} = Utils;
 
 describe('utils / check', () => {
   describe('isColString', () => {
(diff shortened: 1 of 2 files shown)


test/unit/utils/check.test.ts
```diff
@@ -0,0 +1,22 @@
+const { expect } = require('chai');
+const { isColString } = require('sequelize/lib/utils/check');
```
ephys, line 2: We could use `import` here since it's a typescript file

ephys (commented):
I think we could add a test for `table.col` which should result in false
Otherwise, LGTM!

test/unit/utils/check.test.ts
sdepold: I tried and it didn't work. let me check

sdepold:
I'm also adding more tests for the rest of the file right now

test/unit/utils/check.test.ts
sdepold: like this?

```import { expect } from "chai";
import { Utils } from "sequelize";

const { isPrimitive, isColString, canTreatArrayAsAnd, defaultValueSchemable } =
  Utils;
```

ephys: Should work, but I expected this to work too:
```
import { isColString } from 'sequelize/lib/utils/check';
```

sdepold: done

sdepold: aha! somehow didn't work before but here we go

PR #14096, opened 2022-02-11
build: generate jsdoc from src instead of lib (v6)

ephys:
Cherry picks https://github.com/sequelize/sequelize/pull/14094 in v6

Fixes https://github.com/sequelize/sequelize/issues/14083 too


sdepold (approved):

sdepold:
Cool. Will do it tonight :)

PR #14099, opened 2022-02-11
feat(perf): use `->>` for single paths and strings

### LINKED ISSUE #14066: Allow for `->>` operator on single paths for JSONB (postgres)
## Issue Creation Checklist

## Feature Description

When using postgres, allow for a JSONB query to executed using the `->>` operator, rather than `#>>` when there is a single path.

In our use case, we are typically querying by a single path in a JSONB field.  For instance:

```
SELECT foo FROM bar WHERE value->>'field' = 'baz'
```

where sequelize would write the query as:

```
SELECT foo FROM bar WHERE value#>>'{field}' = 'baz'
```

which is a single path entry, but ends up being 8% slower on our production and test databases when using `#>>` versus `->>`.

In addition, it means that we must keep copies of indexes that for each type of query, `#>>` and `->>`.  This means that there is about 100GB of duplicated index data.  Those indexes exist only for sequelize, and about 6x the number of executed queries are not using sequelize, and thus using the `->>` specific indexes.

Given the 8% performance drop (with our database and data) when using `#>>'{field}'`, it does not make sense for us to move those non-sequelize queries to use `#>>'{field}'`.

In addition, we use multiple languages, and would like to have a single solution across all languages when using a single path.

My preferred solution would to have the ability to drop to using `->>'field'` when there is a single path.  I understand that this change might disrupt others who are only using sequelize and have built indexes using `#>>'{field}'` for those use cases, so it may make sense to add a feature flag or configuration and not make the change be the default for all users.

A simple case would be to make a change on `query-generator.js`:

```js
      case 'postgres':
        if (paths.length > 1) {
          pathStr = this.escape(`{${paths.join(',')}}`);
          return `(${quotedColumn}#>>${pathStr})`;
        } else {
          pathStr = this.escape(`${paths[0]}`);
          return `(${quotedColumn}->>${pathStr})`;
        }
```

instead of:

```js
      case 'postgres':
        pathStr = this.escape(`{${paths.join(',')}}`);
        return `(${quotedColumn}#>>${pathStr})`;
```

For environments where multiple languages are used, or a combination of sequelize and non-sequelize, it would be very helpful to be able to use only a single set of indexes in this case, as well as use the optimized query path.  In our case, indexes are 9GB-14GB each, and when duplicated use a lot of additional space, memory, and cpu time across all fields that are indexed.  When under load, this becomes a much bigger issue, especially with the 8% penalty.

We would like to optimize for the fastest query, `->>'field'`, and having this option available in sequelize would help significantly.

1. switch all queries and other ORMs to use `#>>'{field}'`, but this gives us a performance penalty of 8% (on our database/data), and requires us to maintain forks of other ORMs for other languages that would implement `->>'field'`
2. maintain our own fork sequelize for this single feature
3. keep the status quo, which gives us a performance penalty when not under load, and a much larger performance penalty due to the size of the indexes when under load

- [ ] No. This feature is relevant to Sequelize as a whole.
- [x] Yes. This feature only applies to the following dialect(s): Postgres

- [x] Yes, I have the time and I know how to start. (but if done as a configuration or enabling a feature might cause questions to be asked while creating the PR)

JerrySievert:
Breaking change to switch to `->>` instead of `#>>` in the case of a
single element array or string

### Pull Request Checklist

### Description Of Change

Breaking change, closes #14066.  When a single element or a string are passed, use the single element JSONB operator `->>` instead of the pathing version `#>>`.

This is a change only for postgres.

### Todos

- [ ] 
- [ ] 
- [ ]

### DIFF
--- a/src/dialects/postgres/query-generator.js
+++ b/src/dialects/postgres/query-generator.js
@@ -985,8 +985,21 @@ class PostgresQueryGenerator extends AbstractQueryGenerator {
       ? column
       : this.quoteIdentifier(column);
 
-    const join = isJson ? '#>' : '#>>';
-    const pathStr = this.escape(`{${_.toPath(path).join(',')}}`);
+    const useSingular = !isJson && typeof path === 'string' || (Array.isArray(path) && path.length === 1);
+
+    const join = isJson
+      ? '#>'
+      : (useSingular
+        ? '->>'
+        : '#>>'
+      );
+
+    const pathStr = useSingular
+      ? (Array.isArray(path)
+        ? this.escape(path[0])
+        : this.escape(path)
+      )
+      : this.escape(`{${_.toPath(path).join(',')}}`);
 
     return `(${quotedColumn}${join}${pathStr})`;
   }
--- a/test/integration/utils.test.js
+++ b/test/integration/utils.test.js
@@ -104,7 +104,7 @@ describe(Support.getTestDialectTeaser('Utils'), () => {
           },
           another_json_field: { x: 1 },
         };
-        const expected = '("metadata"#>>\'{language}\') = \'icelandic\' AND ("metadata"#>>\'{pg_rating,dk}\') = \'G\' AND ("another_json_field"#>>\'{x}\') = \'1\'';
+        const expected = '("metadata"->>\'language\') = \'icelandic\' AND ("metadata"#>>\'{pg_rating,dk}\') = \'G\' AND ("another_json_field"->>\'x\') = \'1\'';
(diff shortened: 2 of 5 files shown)


WikiRik:
Thanks for the PR! It seems that two tests are still failing, can you take a look into that?

JerrySievert:
I converted to a draft until this is complete.

WikiRik:
I've looked at this again, and after resolving the conflicts and updating tests there is only one test that requires more research/fixing;
https://github.com/sequelize/sequelize/blob/fcbdb5165d056c81bed1fb7756415ba24e080ea4/test/integration/json.test.js#L191-L201

Issue #11588, opened 2019-10-20
Sequelize should combine duplicated properties in the where clauses of multiple scopes by ANDing them instead of replacing them


loban:
### Is your feature request related to a problem? Please describe.
Sequelize's *"combine where clause when using multiple scopes"* feature is severely limited because it replaces duplicated properties instead of **AND**ing them.

For example combining
```js
where: { id: 1 }
```
with
```js
where: { company_id: 2 }
```
results in the expected
```js
where: { id: 1, company_id: 2}
```
This is the equivalent of combining
```sql
WHERE id = 1
```
with
```
WHERE company_id = 2
```
to produce
```sql
WHERE id = 1 AND company_id = 2
```
However, if I wanna combine, e.g.
```sql
WHERE id = 1 OR status = 'P'
```
with
```sql
WHERE company_id = 2 OR company_status = 'P'
```
to produce
```sql
WHERE (id = 1 OR status = 'P') AND (company_id = 3 OR company_status = 'P')
```
I **CANNOT** do this with combined scopes. Because if I combine
```js
where: { [Op.or]: { id: 1, status: 'P' } }
```
with
```js
where: { [Op.or]: { id: company_id, company_status: 'P' } }
```
I get
```js
where: { [Op.or]: { id: company_id, company_status: 'P' } }
```
Meaning the latter is just overwriting the former. Another example that does **NOT** work is combining
```js
where: { id: { [Op.gt]: 5 } }
```
with
```js
where: { id: { [Op.lt]: 50 } }
```
which would have been super useful.

### Describe the solution you'd like
The trouble is that when Sequelize finds a duplicate key in the where clauses of multiple scopes, whether it is a column name, or an `Op.*` operator, it just replaces them! 

**This also means that `Model.scope([a, b])` is not guaranteed to be `Model.scope([a, b])`.**

My solution is to first realize that
```js
where: { id: 1, company_id: 2}
```
is really a shorthand for
```js
where: {
  [Op.and]: [ { id: 1 }, { company_id: 2} ]
}
```
So Sequelize should combine duplicated properties in multiple scopes by **AND**ing them together.

So combining
```js
where: { id: 1 }
```
with
```js
where: { company_id: 2 }
```
makes
```js
where: { 
  [Op.and]: [
    { id: 1 }
    { company_id: 2 }
  ]
}
```
which is the same as
```js
where: { id: 1, company_id: 2}
```
And now the other combinations work too. E.g.
```js
[...]

Issue #10873, opened 2019-04-30
Merge strategy option for multiple scopes


PatrickGeyer:
**Is your feature request related to a problem? Please describe.**
I think it would be useful to be able to define a merge strategy when querying using scopes.

E.g. the examples show the following:
```
{
  scope1: {
    where: {
      firstName: 'bob',
      age: {
        [Op.gt]: 20
      }
    },
    limit: 2
  },
  scope2: {
    where: {
      age: {
        [Op.gt]: 30
      }
    },
    limit: 10
  }
}
```
Calling .scope('scope1', 'scope2') will yield the following query
```
WHERE firstName = 'bob' AND age > 30 LIMIT 10
```
I would like to merge the two ```age``` where indexes using Sequelize.and.
For example if I wanted to get everyone with an age above 30 AND their age being in a list. I would call it like this:

```
User.scope2().findAll({
   where: {
        age: {
            [Sequelize.Op.in]: [20, 23, 40, 41, 50, 55]
        }
   }
})
```
However, now the age parameter from the scope is overwritten. 
**Describe the solution you'd like**
Be able to specify the merge strategy of the where parameters, in the scope definition and in the FindParams options. See below.

**Describe alternatives you've considered**
Using [Sequelize.and] in the findAll parameters after calling the scope so that age and [Sequelize.and] are NOT merged. I.e.

```
User.scope2().findAll({
   where: {
[Sequelize.Op.and]: [{
        age: {
            [Sequelize.Op.in]: [20, 23, 40, 41, 50, 55]
        }
}]
   }
})
```
However, this only lets me do it once, as the next time I use the Sequelize.and symbol, it will overwrite the previous one.

**Usage example**
Ideal scenario:
```
User.scope2().findAll({
    whereMergeStrategy: 'AND'
})
```

And
```
{
     scope2: () => ({
         whereMergeStrategy: 'AND',
         where: {
             age: {
            [Op.gt]: 30
            }
         },
    })
}

```

PatrickGeyer:
@SimonSchick @knoxcard @sushantdhiman if I created a PR for this would you consider merging?

SimonSchick:
Can't hurt.

ephys:
This should be resolved thanks to:

https://github.com/sequelize/sequelize/pull/14110
https://github.com/sequelize/sequelize/pull/14152

PR #14110, opened 2022-02-14
feat: merge scope where conditions using the and operator

EtienneTurc:
### Pull Request Checklist

### Description Of Change

Issues related:
- https://github.com/sequelize/sequelize/issues/11588
- https://github.com/sequelize/sequelize/issues/10873

Pull request that tried to solve these issues:
- https://github.com/sequelize/sequelize/pull/11805

This pr adds the following elements:
- A way to merge scopes with where clauses using the `AND` operator instead of overwriting it with the last value. It supports merging attributes, `[Op.and]` and `[Op.or]`. 
- The flag `mergeWhereScopesWithAndOperator` to enable this behavior for a model or all the models.

Examples:
```js
// Attributes merging

// Scope 1
whereAttributeIs1: {
  where: {
    field: 1,
  },
},
// Scope 2
whereAttributeIs2: {
  where: {
    field: 2,
  },
}

// Current behavior (without mergeWhereScopesWithAndOperator)
myModel.scope('whereAttributeIs1', 'whereAttributeIs2') ._scope ===
{
  where: {
    field: 2,
  }
}

// Expected behavior (with mergeWhereScopesWithAndOperator)
myModel.scope('whereAttributeIs1', 'whereAttributeIs2') ._scope ===
{
  where: {
    field: {
      [Op.and]: [1, 2]
    }
  }
}
```

```js
// Op.and merging

// Scope 1
whereOpAnd1: {
  where: {
    [Op.and]: [{ field: 1 }],
  },
},
// Scope 2
whereOpAnd2: {
  where: {
    [Op.and]: [{ field: 2 }],
  },
},

// Current behavior (without mergeWhereScopesWithAndOperator)
myModel.scope('whereOpAnd1', 'whereOpAnd2') ._scope ===
{
  where: {
    [Op.and]: [{ field: 2 }],
  },
}

// Expected behavior (with mergeWhereScopesWithAndOperator)
myModel.scope('whereOpAnd1', 'whereOpAnd2') ._scope ===
{
  where: {
    [Op.and]: [{ field: 1 }, { field: 2 }],
  },
}
```

```js
// Op.or merging

// Scope 1
whereOpOr1: {
  where: {
    [Op.or]: [{ field: 1 }],
  },
},
// Scope 2
whereOpOr2: {
  where: {
    [Op.or]: [{ field: 2 }],
  },
},

// Current behavior (without mergeWhereScopesWithAndOperator)
myModel.scope('whereOpOr1', 'whereOpOr2') ._scope ===
{
  where: {
    [Op.or]: [{ field: 2 }]
  },
}

// Expected behavior (with mergeWhereScopesWithAndOperator)
myModel.scope('whereOpAnd1', 'whereOpAnd2') ._scope ===
{
  where: {
    [Op.and]: [
      { [Op.or]: [{ field: 1 }] }, 
      { [Op.or]: [{ field: 2 }] }
    ],
  },
}
```

Why should this be in Sequelize ?

The current behavior for merging different attributes already acts as an `AND` operator. Thus it is natural to expect the  
same behavior for different conditions on the same attribute.

### DIFF
--- a/test/unit/model/include.test.js
+++ b/test/unit/model/include.test.js
@@ -1,14 +1,23 @@
 'use strict';
 
 const chai = require('chai');
+const util = require('util');
 
 const expect = chai.expect;
 const Support   = require('../support');
 const Sequelize = require('@sequelize/core');
 
+const Op = Sequelize.Op;
+
 const current   = Support.sequelize;
 
 describe(Support.getTestDialectTeaser('Model'), () => {
+  // Using util.inspect to correctly assert objects with symbols
+  // Because expect.deep.equal does not test non iterator keys such as symbols (https://github.com/chaijs/chai/issues/1054)
+  chai.Assertion.addMethod('deepEquals', function (expected, depth = 5) {
+    expect(util.inspect(this._obj, { depth })).to.deep.equal(util.inspect(expected, { depth }));
+  });
+
   describe('all', () => {
     const Referral = current.define('referal');
 
@@ -179,7 +188,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           include: [{ model: this.Project }],
         });
 
-        expect(options.include[0]).to.have.property('where').which.deep.equals({ active: true });
+        expect(options.include[0]).to.have.property('where').which.deepEquals({ [Op.and]: [{ active: true }] });
       });
 
       it('adds the where from a scoped model', function () {
@@ -188,7 +197,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           include: [{ model: this.Project.scope('that') }],
         });
 
-        expect(options.include[0]).to.have.property('where').which.deep.equals({ that: false });
+        expect(options.include[0]).to.have.property('where').which.deepEquals({ [Op.and]: [{ that: false }] });
         expect(options.include[0]).to.have.property('limit').which.equals(12);
       });
 
@@ -198,7 +207,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           include: [{ model: this.Project.scope('attr') }],
         });
 
-        expect(options.include[0]).to.have.property('attributes').which.deep.equals(['baz']);
+        expect(options.include[0]).to.have.property('attributes').which.deepEquals(['baz']);
       });
 
       it('merges where with the where from a scoped model', function () {
@@ -207,7 +216,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           include: [{ where: { active: false }, model: this.Project.scope('that') }],
         });
 
-        expect(options.include[0]).to.have.property('where').which.deep.equals({ active: false, that: false });
+        expect(options.include[0]).to.have.property('where').which.deepEquals({ [Op.and]: [{ that: false }, { active: false }] });
       });
 
       it('add the where from a scoped associated model', function () {
@@ -216,7 +225,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           include: [{ model: this.Project, as: 'thisProject' }],
         });
 
-        expect(options.include[0]).to.have.property('where').which.deep.equals({ this: true });
+        expect(options.include[0]).to.have.property('where').which.deepEquals({ [Op.and]: [{ this: true }] });
       });
 
       it('handles a scope with an aliased column (.field)', function () {
@@ -225,7 +234,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           include: [{ model: this.Project.scope('foobar') }],
         });
 
-        expect(options.include[0]).to.have.property('where').which.deep.equals({ foo: 42 });
+        expect(options.include[0]).to.have.property('where').which.deepEquals({ [Op.and]: [{ foo: 42 }] });
       });
     });
 
--- a/test/unit/model/include.test.js
+++ b/test/unit/model/include.test.js
@@ -1,23 +1,14 @@
 'use strict';
 
 const chai = require('chai');
-const util = require('util');
 
 const expect = chai.expect;
 const Support   = require('../support');
-const Sequelize = require('@sequelize/core');
-
-const Op = Sequelize.Op;
+const { Sequelize, Op, Utils } = require('@sequelize/core');
 
 const current   = Support.sequelize;
 
 describe(Support.getTestDialectTeaser('Model'), () => {
-  // Using util.inspect to correctly assert objects with symbols
-  // Because expect.deep.equal does not test non iterator keys such as symbols (https://github.com/chaijs/chai/issues/1054)
-  chai.Assertion.addMethod('deepEquals', function (expected, depth = 5) {
-    expect(util.inspect(this._obj, { depth })).to.deep.equal(util.inspect(expected, { depth }));
-  });
-
   describe('all', () => {
     const Referral = current.define('referal');
 
@@ -188,7 +179,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           include: [{ model: this.Project }],
         });
 
-        expect(options.include[0]).to.have.property('where').which.deepEquals({ [Op.and]: [{ active: true }] });
+        expect(options.include[0]).to.have.property('where').which.deep.equals({ active: true });
       });
 
       it('adds the where from a scoped model', function () {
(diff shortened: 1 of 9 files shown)


ephys:
Wouldn't it be easier and less error prone to simply combine where conditions with `Op.and`?

So:

```typescript
// Scope 1
whereAttributeIs1: {
  where: { field: 1 },
},
// Scope 2
whereAttributeIs2: {
  where: { field: 2 },
}

myModel.scope('whereAttributeIs1', 'whereAttributeIs2') ._scope === 
{
  where: {
    [Op.and]: [
      { field: 1 },
      { field: 2 },
    ]
  }
}
```

EtienneTurc:
> bl0cknumber

Thanks for the reply and the suggestion. Indeed, it would be nice to update the manual documentation with this new feature. I assume it would be best to add a new sub section in the merging section in https://sequelize.org/v6/manual/scopes.html (https://github.com/EtienneTurc/sequelize/blob/add-option-to-merge-where-scope-with-and-operator/docs/manual/other-topics/scopes.md).

> ephys

Thanks for the reply and the suggestion. Indeed, it looks easier and thus the previous examples will look like these:

```js
// Attributes merging

// Scope 1
whereAttributeIs1: {
  where: {
    field: 1,
  },
},
// Scope 2
whereAttributeIs2: {
  where: {
    field: 2,
  },
}

// New expected behavior
myModel.scope('whereAttributeIs1', 'whereAttributeIs2')._scope ===
{
  where: {
    [Op.and]: [
      {field: 1},
      {field: 2}
   ]
  }
}
```

```js
// Op.and merging

// Scope 1
whereOpAnd1: {
  where: {
    [Op.and]: [{ field: 1 }],
  },
},
// Scope 2
whereOpAnd2: {
  where: {
    [Op.and]: [{ field: 2 }],
  },
},

// New expected behavior
myModel.scope('whereOpAnd1', 'whereOpAnd2')._scope ===
{
  where: {
    [Op.and]: [
      { field: 1 },
      { field: 2 }
    ],
  },
}
```

```js
// Op.or merging

// Scope 1
whereOpOr1: {
  where: {
    [Op.or]: [{ field: 1 }],
  },
},
// Scope 2
whereOpOr2: {
  where: {
    [Op.or]: [{ field: 2 }],
  },
},

// New expected behavior
myModel.scope('whereOpAnd1', 'whereOpAnd2')._scope ===
{
  where: {
    [Op.and]: [
      { [Op.or]: [{ field: 1 }] },
      { [Op.or]: [{ field: 2 }] }
    ],
  },
}
```

But let's also note that this implementation will also impact those cases (even if this should not be an issue):
```js
// Scope 1
whereAttributeIs1: {
  where: {
    field1: 1,
  },
},
// Scope 2
whereAttributeIs2: {
  where: {
    field2: 2,
  },
}

// Current behavior
myModel.scope('whereAttributeIs1', 'whereAttributeIs2')._scope ===
{
  where: {
    field1: 1,
    field2: 2
  },
}

// New expected behavior
myModel.scope('whereAttributeIs1', 'whereAttributeIs2')._scope ===
{
  where: {
    [Op.and]: [
      {field1: 1},
      {field2: 2}
   ]
  }
}
```

(Currently making the modifications)

EtienneTurc:
I've made the requested changes. Note that it slightly changes the behavior of merging `Op.and`:

```js
// Op.and merging

// Scope 1
whereOpAnd1: {
  where: {
    [Op.and]: [{ field: 1 }],
  },
},
// Scope 2
whereOpAnd2: {
  where: {
    [Op.and]: [{ field: 2 }],
  },
},

// New behavior
myModel.scope('whereOpAnd1', 'whereOpAnd2')._scope ===
{
  where: {
    [Op.and]: [
      { [Op.and]: [{ field: 1  }] },
      { [Op.and]: [{ field: 2  }] }
    ],
  },
}
```

It drastically simplifies the code and makes debugging easier because where clauses from different scopes are not mixed together.

I am waiting for you to approve these changes before adding a new sub-section in the manual documentation :)

ephys (changes requested):
From what I can see, `mergeWhereScopesWithAndOperator` is in the `Model`'s options. I think that's fine because IMO the current merge behavior is just plain wrong and should be replaced with the new one.

---

This is targeting `main` (v7 alpha). I think it would be better to only introduce this flag in v6 to avoid a breaking change, and use the new behavior as the default behavior in v7.

That would mean:
- Open a new PR that targets the v6 branch and copy-paste the current code there.
- Change this PR to make merging with AND the only behavior in v7, without a flag

---

Can you add a few tests that verify that `Model.findAll` merges the scopes using the right strategy?

EtienneTurc:
Thanks for the review. I also believe that this behavior should be the default one, so I am glad that you are willing to use it for the v7.

I will make the two PRs in the following days :)

EtienneTurc:
I've made the requested modifications. So now this pr (for the v7) changes the merging behavior from the `overwrite` one to the `and` one. Thus this change is breaking and tests have been modified/added to take this into consideration.

For the v6, I've opened this pr:
- https://github.com/sequelize/sequelize/pull/14152

NB: I took the liberty to name the current behavior `overwrite` instead of `assign` as suggested; I find it more explicit.

ephys:
Thanks! I'll try to not take too long before reviewing both PRs.

I'm not the most knowledgeable users of scopes (never used them, actually) so I'd like another member of @sequelize/maintainers to review this too, but based on my understanding of scopes, these changes look good.

In-depth review as soon as I can

EtienneTurc:
I've made the requested changes (on both prs). 

I was reluctant at first to unpack `Op.and` because in some cases it mixes multiples scopes together making it a bit harder to debug (but still functional). However, in most cases, this operand is just unnecessary, so I guess it makes sense to remove it.

test/unit/model/include.test.js
```diff
@@ -179,7 +180,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           include: [{ model: this.Project }],
         });
 
-        expect(options.include[0]).to.have.property('where').which.deep.equals({ active: true });
+        expect(options.include[0]).to.have.property('where').which.deep.equal({ active: true  });
```
ephys, line 183: You reverted your usage of deepEqual, is that intended? Didn't you add it to be able to compare `symbol`?

I saw a commit fail because it was trying to compare 'object with null prototype' to 'object', I assume that's why you reverted?

https://github.com/chaijs/chai/issues/1054 has had some movement but I don't think they'll release the change until chai 5

---

also if `deep.equals` and `deep.equal` are the same, can you revert to `deep.equals` to reduce risk of conflict?

test/unit/model/include.test.js
```diff
@@ -207,7 +208,9 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           include: [{ where: { active: false }, model: this.Project.scope('that') }],
         });
 
-        expect(options.include[0]).to.have.property('where').which.deep.equals({ active: false, that: false });
+        expect(options.include[0]).to.have.property('where');
+        expect(Utils.getComplexKeys(options.include[0].where)).to.deep.equal([Op.and]);
```
ephys, line 212: Aah I see you test it manually, then you can ignore the first part of my previous comment :)

Can you add `// TODO [chai>5]: simplify once '.deep.equal' includes support for symbols (https://github.com/chaijs/chai/issues/1054)`

ephys (changes requested):
Looking good :)
I agree with your concern, if we didn't use _.merge we could have enough information to only add [Op.and] when necessary without unpacking existing [Op.and] but I think this is good enough

Only 3 nits and it's good on my end
The concept was also approved by other team members

EtienneTurc:
> Looking good :) 

Great ! I have done the few changes.

test/unit/model/include.test.js
```diff
@@ -179,7 +179,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           include: [{ model: this.Project }],
         });
 
-        expect(options.include[0]).to.have.property('where').which.deep.equals({ active: true });
+        expect(options.include[0]).to.have.property('where').which.deep.equals({ active: true  });
```
ephys, line 182:
```suggestion
        expect(options.include[0]).to.have.property('where').which.deep.equals({ active: true });
```

not sure why the linter didn't catch this
will merge as soon as this suggestion is committed :)

ephys (approved):

test/unit/model/include.test.js
EtienneTurc: Don't know either 🤔

EtienneTurc: I've removed the extra space, so it is good for the merge ;)

ephys (approved):

ephys:
And here we go! 🎊

PR #14119, opened 2022-02-15
feat(postgres): support connectionTimeoutMillis dialectOption

mmfKupl:
### Pull Request Checklist

### Description Of Change

Add support of "connectionTimeoutMillis" option for PostgreSQL dialect
https://node-postgres.com/api/client


sdepold:
I think this should target main and we cherry pick it over. How confident are we that the camel case notation is correct where all the other props seem to prefer underscores.

To be investigated:

Do we have tests for that?
Do we document the supported options somewhere?
Is there actually any reason to not just forward all passes option to the driver in question? Why would sequelize do a sanity check in the first place?
Can we make use of the drivers typescript annotations to prevent people passing rubbish?

sdepold:
Thanks for your pr :)

ephys:
I agree with the sanity check being redundant. I think dialectOptions should be passed to the dialect as-is

sdepold:
> I agree with the sanity check being redundant. I think dialectOptions should be passed to the dialect as-is

@mmfKupl Interested in doing that change?

mmfKupl:
> > I agree with the sanity check being redundant. I think dialectOptions should be passed to the dialect as-is
> 
> @mmfKupl Interested in doing that change?

Yes. I will change my pr with this comment

ephys:
`clientMinMessages`, `odbcConnectionString` `authentication`, `debug`, and `prependSearchPath` are the 5 dialect options I could find that are handled by Sequelize

I think we'll leave things as they are right now and we'll revisit what dialectOptions is meant to be in Sequelize 8 when we split the dialects to their own packages. It should allow us to strictly type `dialectOptions` based on the dialect.

ephys (approved):

PR #14120, opened 2022-02-15
fix: maintains previousDataValues for update when returning or individual hooks are true

### LINKED ISSUE #8803: Using previous() in `afterUpdate`-hook not working
## What are you doing?
I am trying to use a `afterUpdate`-hook to check status changes on records so I can inform interested parties For this I am using the model for the hook listed below:

```js
{
      orderDate: {
        type: Sequelize.DATEONLY,
        field: 'date',
        allowNull: false,
      },
      partySize: {
        type: Sequelize.INTEGER,
        field: 'party_size',
        allowNull: false,
      },
      bookingReference: {
        type: Sequelize.STRING,
        field: 'booking_reference',
        allowNull: false,
      },
      status: {
        type: Sequelize.ENUM(
          StatusType.CONFIRMED,
          StatusType.CANCELLED,
          StatusType.NO_SHOW,
          StatusType.ARRIVED,
          StatusType.SEATED,
          StatusType.FINISHED
        ),
        allowNull: false,
      }
    }
```

```js
          afterUpdate: (order, options, fn) => {
            const oldOrderStatus = order.previous('status')
            const orderStatus = order.status
            app.log.debug('Order.afterUpdate() oldOrderStatus: %s orderStatus: %s', oldOrderStatus, orderStatus)
            fn(null, order)
          },
```

I am issuing the following `update`-call to update the record only 

```js
        this.app.orm.Order.update(sanitizedOrderData, {
          where: { id },
          individualHooks: true,
          returning: true,
          transaction: t,
        })
```

I am expecting that I am getting the updated record data and that I can access previous value before I updated the order. Together, I am also receiving the new record data as part of the `resolve`-promise in the `Order.update`-call were the record data use field name such as `bookingReference` instead of `book_reference`.

Instead of receiving the data I am getting a runtime error:

```
Error:  TypeError: order.previous is not a function
    at Model.afterUpdate (/Users/xxx/Development/Projects/batch/backend/api/models/Order.js:19:54)
    at Model.tryCatcher (/Users/xxx/Development/Projects/batch/backend/node_modules/bluebird/js/release/util.js:16:23)
```

If I am disable `returning` it works as expected. The `afterUpdate` hook is being called and I can use the `previous`-function.

The generated SQL code:
```
xecuting (31ac8a93-ae42-47aa-b68e-fa38de33879e): START TRANSACTION;
Executing (31ac8a93-ae42-47aa-b68e-fa38de33879e): SET SESSION TRANSACTION ISOLATION LEVEL REPEATABLE READ;
Executing (31ac8a93-ae42-47aa-b68e-fa38de33879e): SELECT "id", "date" AS "orderDate", "party_size" AS "partySize", "booking_reference" AS "bookingReference", "status", "created_at", "updated_at" FROM "order" AS "Order" WHERE "Order"."id" = '1';
Executing (31ac8a93-ae42-47aa-b68e-fa38de33879e): UPDATE "order" SET "date"='2017-12-06',"updated_at"='2017-12-19 21:52:14.775 +00:00' WHERE "id" = '1' RETURNING *
E
```

__Dialect:__ `postgres`
__Dialect version:__ `pg 6.4.2` 
__Database version:__ `PostgreSQL 9.6.3 on x86_64-pc-linux-gnu, compiled by gcc (Debian 4.9.2-10) 4.9.2, 64-bit`
__Sequelize version:__ `4.28.6`
__Tested with latest release:__ Yes,  (If yes, specify that version)

ChadHelbling:
### Pull Request Checklist

### Description Of Change

Closes #8803

There is some inconsistency in instance properties on update when `returning: true` and `individualHooks: true`. When returning = true the pre-read instances that get built are overwritten with the returned models from the update (https://github.com/sequelize/sequelize/blob/a729c4df41fa3a58fbecaf879265d2fb73d80e5f/src/model.js#L3389). This causes all the instance properties (e.g. _previousDataValues) to be reset. This is specifically an issue when you have afterUpdate hooks that require the previous data to make a decision. 

I took the approach of expanding pre-read to both individualHooks = true and returning = true. There may be a better approach but not being super familiar with the codebase, this seemed the most obvious. I'm definitely open to suggestions/feedback.

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -3302,7 +3302,8 @@ class Model {
     // Get instances and run beforeUpdate hook on each record individually
     let instances;
     let updateDoneRowByRow = false;
-    if (options.individualHooks) {
+
+    if (options.individualHooks || options.returning) {
       instances = await this.findAll({
         where: options.where,
         transaction: options.transaction,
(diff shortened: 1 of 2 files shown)


charlie-s:
Any way to get movement on this? This was almost 1 year ago and the logs for the failed tests aren't available, but shouldn't the Postgres checks be all passing before this will get reviewed? I have time to commit to this issue.

PR #14123, opened 2022-02-16
docs(op): fix queries example for [Op.any]

quappo:
### Description Of Change

Fixed example queries for [Op.any]


WikiRik:
Thanks for the PR! We're happy that Sequelize is your first contribution. You're right that the example was not correct and this indeed fixes it as per the [unit tests](https://github.com/sequelize/sequelize/blob/main/test/unit/sql/where.test.js#L602-L617)

However, there are a few other places where these examples are given which are still incorrect. Like in src/model.d.ts (line 121 and 143), src/operators.ts (line 48 and 58) and test/types/where.ts (line 73 and 81-84). And there might be more.
Can you also update those examples?

WikiRik:
Thanks again for bringing this to our attention!

I have made two PRs to further fix this issue; https://github.com/sequelize/website/pull/66 and https://github.com/sequelize/sequelize/pull/14321

PR #14126, opened 2022-02-16
fix(types): update 'replication' option property

collierrgbsitisfise:
### Description Of Change

Type adjustments according **'replication'** option.

I am trying do setup *replication* option asynchronous using for it hook [beforeConnect](https://sequelize.org/master/manual/hooks.html). According types definition hook receive as parameter [Config](https://github.com/sequelize/sequelize/blob/main/src/sequelize.d.ts#L690). As You can see there is no *replication* defined for [Config Interface](https://github.com/sequelize/sequelize/blob/main/src/sequelize.d.ts#L148), which is wrong.

As result in typescript project, when we try define replication  asynchronous in hook

```javascript
const myReplicationConfig = {...};
new Sequelize({
  ...defautlOptions
  hooks: {
    beforeConnect: (config: DeepWriteable<Config>) => {
       // ts error here >>>
       // Type 'false | ReplicationOptions' is not assignable to type 'DeepWriteable<boolean>'.
       // Type 'ReplicationOptions' is not assignable to type 'DeepWriteable<boolean>'
       config.replication = myReplicationConfig;
    },
  },
});
```

### DIFF
--- a/src/sequelize.d.ts
+++ b/src/sequelize.d.ts
@@ -161,7 +161,7 @@ export interface Config {
   readonly protocol: 'tcp';
   readonly native: boolean;
   readonly ssl: boolean;
-  readonly replication?: ReplicationOptions | false;
+  readonly replication: ReplicationOptions | false;
   readonly dialectModulePath: null | string;
   readonly keepDefaultTimezone?: boolean;
   readonly dialectOptions?: {
--- a/src/sequelize.d.ts
+++ b/src/sequelize.d.ts
@@ -161,7 +161,7 @@ export interface Config {
   readonly protocol: 'tcp';
   readonly native: boolean;
   readonly ssl: boolean;
-  readonly replication: boolean;
+  readonly replication?: ReplicationOptions | false;
   readonly dialectModulePath: null | string;
   readonly keepDefaultTimezone?: boolean;
   readonly dialectOptions?: {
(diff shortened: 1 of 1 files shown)


src/sequelize.d.ts
```diff
@@ -308,7 +308,7 @@ export interface Options extends Logging {
    *
    * @default false
    */
-  replication?: ReplicationOptions;
+  replication?: ReplicationOptions | false;
```
collierrgbsitisfise, line 311: more fair type, cause `false` is by default

bl0cknumber (dismissed):

src/sequelize.d.ts
```diff
@@ -161,7 +161,7 @@ export interface Config {
   readonly protocol: 'tcp';
   readonly native: boolean;
   readonly ssl: boolean;
-  readonly replication: boolean;
+  readonly replication?: ReplicationOptions | false;
```
ephys, line 164:
```suggestion
  readonly replication: ReplicationOptions | false;
```

`replication` is optional in `Options` but not in `Config` as Sequelize will assign a default value

ephys (changes requested):
Looks good :)

src/sequelize.d.ts
collierrgbsitisfise: @ephys fixed

ephys (approved):

ephys:
Thank you for the PR :)

PR #14132, opened 2022-02-17
meta: update comment in bug report template

WikiRik:
The inline comment () does not like it when there are two dashes which caused the comment to be rendered as normal text. This fixes that but unfortunately does space out the three points as separate paragraphs. So it's the lesser of two evils in my view

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -85,12 +85,14 @@ Add any other context and details here.
 
 ### Environment
 
+<!--- To find the version numbers for the three systems below use the following commands:
+- `npm list sequelize` / `yarn list --pattern sequelize`
+- `node -v`
+- `npm list typescript` / `yarn list --pattern typescript` --->
+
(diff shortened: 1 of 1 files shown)


sdepold (approved):
Lgtm

ephys:
Since the list isn't very long, maybe we could move all instructions before the list? It would solve the list being split into 3 lists

WikiRik:
Sure I can do that

bl0cknumber:
Or 

- Sequelize version: XXX
```sh
# run `npm list sequelize` or `yarn list --pattern sequelize` to obtain sequelize version
npm list sequelize # or
yarn list --pattern sequelize
```

- Node.js version: XXX
```sh
# run `node -v` to obtain Node.js version
node -v
```

```sh
# run `npm list typescript` / `yarn list --pattern typescript` to obtain TypeScript version
npm list typescript # or
yarn list --pattern typescript
```

ephys:
That would add a lot of noise to issues, and they're already noisy. I'd prefer this remained a comment

ephys (approved):

PR #14137, opened 2022-02-18
fix(types): missing snowflake and db2 dialects

bl0cknumber:
Missing snowflake and db2 dialects types, causing casting with any

### DIFF
--- a/src/sequelize.d.ts
+++ b/src/sequelize.d.ts
@@ -170,7 +170,7 @@ export interface Config {
   };
 }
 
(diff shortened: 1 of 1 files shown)


WikiRik:
Is there some test we can add to check that this list is complete?

ephys:
I don't think so. It would have to be tested individually for each dialect and someone forgetting to add the dialect name to the typings will also forget to add its corresponding test.

It won't be an issue anymore once the dialect option is a Dialect class instead of a string

bl0cknumber:
Cool. thanks guys, I will merge this for v6.

PR #14145, opened 2022-02-20
fix: return type of static increment and decrement functions

Qlio:
Static model methods `increment` and `decrement` returns array. But typescript is implemented differently.

### DIFF
--- a/src/model.d.ts
+++ b/src/model.d.ts
@@ -2216,7 +2216,7 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
     this: ModelStatic<M>,
     field: keyof Attributes<M>,
     options: IncrementDecrementOptionsWithBy<Attributes<M>>
-  ): Promise<M>;
+  ): Promise<[M[], number?]>;
 
   /**
    * Increments multiple fields by the same value.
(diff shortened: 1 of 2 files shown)


ephys:
This is concerning but I can't find any test for the return type of `Model.increment` / `Model.decrement`.

I think the return type is `Promise<[affectedModels: M[], affectedRowCount?: number]>` (affectedModels is an array. I've also added tuple labels as they're supported since TS 4.0)

src/model.d.ts
```diff
@@ -2243,7 +2243,7 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
     this: ModelStatic<M>,
     field: keyof Attributes<M>,
     options: IncrementDecrementOptionsWithBy<Attributes<M>>
-  ): Promise<M>;
+  ): Promise<[M[], number?]>;
```
ephys, line 2246: last request but could you label these tuples ? :)

ephys, line 2246:
```suggestion
  ): Promise<[affectedRows: M[], affectedCount?: number]>;
```

last request but could you label these tuples ? :)

ephys (changes requested):
thanks for adding tests!

ephys:
Actually quick update: As the return type is dependant on `option.returning` I think we could handle it like `Model.count` and have two method signatures based on option.returning:

- one with option.returning set to true that returns [affectedCount: number, affectedRows: M[]]
- one without option.returning (or set to false) that returns [affectedCount: number]

PS you don't have to keep your commit history clean as all PRs are squashed before merging :)

PR #14155, opened 2022-02-23
fix(types): update return type of `Model.update`

guyellis:
### Pull Request Checklist

### Description Of Change

As per the docs in the type I've updated the type.

A few lines above the change is a description where the update() method can return 2 types. I added the 2nd type.


src/model.d.ts
```diff
@@ -2207,7 +2207,7 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
         [key in keyof Attributes<M>]?: Attributes<M>[key] | Fn | Col | Literal;
     },
     options: UpdateOptions<Attributes<M>>
-  ): Promise<[number, M[]]>;
+  ): Promise<[number] | [number, M[]]>;
```
guyellis, line 2210: A few lines above this the docs for this method say:

> The promise returns an array with one or two elements...

guyellis: And I can confirm that the functionality is accurate based on the docs and that the types didn't match the return type which is why I'm submitting this PR.

guyellis:
Failing test looks like a random issue unrelated to this PR:

```
  1 failing

  1) [SQLITE] Transaction
     AssertionError: expected promise to be rejected with an error including 'SQLITE_BUSY: database is locked' but it was fulfilled with [ undefined, undefined ]
      at assertIfNotNegated (node_modules/chai-as-promised/lib/chai-as-promised.js:68:19)
      at /home/runner/work/sequelize/sequelize/node_modules/chai-as-promised/lib/chai-as-promised.js:183:17
      at async Context.<anonymous> (test/integration/transaction.test.js:739:9)
```

src/model.d.ts
ephys, line 2210: Suggestion to follow the format that's used in https://github.com/sequelize/sequelize/pull/14145

```suggestion
  ): Promise<[affectedCount: number, affectedRows?: M[]]>;
```

ephys (changes requested):
I think we could handle this like `Model.count`: have two method signatures:
- one with option.returning set to true that returns `[affectedCount: number, affectedRows: M[]]`
- one without option.returning (or set to false) that returns `[affectedCount: number]`

ephys:
Also you're right, the sqlite issue is something that started happening randomly recently but unrelated to your PR

guyellis:
> I think we could handle this like `Model.count`: have two method signatures:
> 
> * one with option.returning set to true that returns `[affectedCount: number, affectedRows: M[]]`
> * one without option.returning (or set to false) that returns `[affectedCount: number]`

That was a great idea. Thank you! I've made the changes and pushed them. LMK what you think. Thanks.

src/model.d.ts
```diff
@@ -2202,12 +2202,25 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
    * affected rows (only supported in postgres and mssql with `options.returning` true.)
    */
   public static update<M extends Model>(
+    this: ModelStatic<M>,
+    values: {
+        [key in keyof Attributes<M>]?: Attributes<M>[key] | Fn | Col | Literal;
+    },
+    options: UpdateOptions<Attributes<M>> & { returning: true }
+  ): Promise<[affectedCount: number, affectedRows?: M[]]>;
```
ephys, line 2210: I think as you split the two signatures, the new signature should be:

```suggestion
  ): Promise<[affectedCount: number, affectedRows: M[]]>;
```

ephys (approved):
Thanks, I think we're good to go :)
Merging another PR first then I'll update & merge this one

guyellis:
> Thanks, I think we're good to go :) Merging [another PR](https://github.com/sequelize/sequelize/pull/14151) first then I'll update & merge this one

Thank you for doing this! You are awesome! :trophy:

Kessa713:
There's an issue with this in reference to the documentation and current functionality for `returning: (keyof TAttributes)[]` as listed [here](https://sequelize.org/v6/class/src/model.js~Model.html#static-method-update)
This breaks existing type checks when updating to v6.17.0

ruiaraujolink:
Hi, this PR create a problem, when I pass an array of attributes this happens
Screenshot 2022-03-04 at 13 58 40

WikiRik:
> Hi, this PR create a problem, when I pass an array of attributes this happens
> Screenshot 2022-03-04 at 13 58 40
>  

Hi! This should be fixed with #14184 right? That PR will be reviewed soon and then we will release a new version

ruiaraujolink:
> > Hi, this PR create a problem, when I pass an array of attributes this happens
> > Screenshot 2022-03-04 at 13 58 40
> 
> Hi! This should be fixed with #14184 right? That PR will be reviewed soon and then we will release a new version

I think so, thanks.

PR #14158, opened 2022-02-24
meta: set access to public for publishing

WikiRik:
This should fix the publishing error we got on v7


sdepold:
From the docs and the reason for this change:
> Default: 'restricted' for scoped packages, 'public' for unscoped packages

sdepold (approved):

PR #14159, opened 2022-02-24
build(deps): move @types/validator to prod deps

WikiRik:
### Pull Request Checklist

### Description Of Change

I've moved @types/validator to the prod dependencies since their respective dependency is also a prod dependency and we need the types available for the TS migration.
I've also regenerated the yarn.lock file. This cleaned up unnecessary versions of some packages. But probably dependabot or renovate will not agree with it fully so some of these changes will be undone in their next PR


ephys:
OK for `@types/validator` but `@types/node` should remain in devDeps as its version is controlled by the user

WikiRik:
> OK for `@types/validator` but `@types/node` should remain in devDeps as its version is controlled by the user

I believe that it also mentions in the typedocs what version things were added but to be safe I put that back

ephys (approved):
Can you update the typescript.md documentation to remove the mention of `@types/validator`?

WikiRik:
Ah yes, I forgot about the docs. Done

ephys (approved):
good to go :)

PR #14160, opened 2022-02-24
ci: disable CI on main

ephys:
Proposal PR to disable running the CI on main.

As all changes on `main` are done through PRs, we already know the the main branch passes the CI. 
Nothing is released from the main branch either.

Disabling the CI on the main branch would reduce how many actions are being run concurrently and allow PRs to run faster.

I may be missing something so feel free to reject this PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,6 @@ name: CI
 on:
   push:
     branches:
-      - main
       - v6
       - v7
   pull_request:


WikiRik (approved):
PRs do need to be up to date with the main branch so it might make sense. It sounds scary though so we should watch this and evaluate in a while to see if we think this will be a good change

sdepold:
Somehow I don't know any longer how to approve something on mobile GitHub 😅 anyway, I had the same thoughts a couple of times recently hence 👍

WikiRik:
> Somehow I don't know any longer how to approve something on mobile GitHub 😅 anyway, I had the same thoughts a couple of times recently hence 👍

You need to click on the file changes (where it also says 'Review changes' in blue) to get to the changes and then at the bottom of your screen you can review. This is on Android at least

sdepold:
Aaah there it is :) just as on web 🙈😅

sdepold (approved):

PR #14163, opened 2022-02-24
ci: skip 'rm awaiting response' if label is absent

ephys:
This PR aborts the 'remove awaiting response' action before its environment is set up if the issue/pr doesn't have the label the CI tries to remove.

I'm not sure how to test this without merging it


WikiRik (approved):
Best way to test it is to merge it indeed

ephys:
The abort functioned as expected https://github.com/sequelize/sequelize/actions/runs/1895397031
Let's see if the label is still removed

ephys:
😊

PR #14168, opened 2022-02-25
build: add caching to CI workflow

WikiRik:
### Pull Request Checklist

### Description Of Change

This will hopefully speed up rerunning checks. Something that @ephys and I have been discussing is that it might be nice if we can store the installed and built artifacts in the first job of this workflow and that it can be reused for the other jobs but that's for a future PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 16.x
+          cache: yarn
       - run: yarn install --frozen-lockfile
       - run: yarn lint-no-fix
       - run: yarn lint-docs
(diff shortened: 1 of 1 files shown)


ephys:
I've compared a few different runs and I think on average the install time is shorter
It's also sometimes slower (by a few seconds) but it may still be positive on average.

WikiRik:
Oh yeah, this will only have an impact when you re-run the checks I believe. It's not reusing artifacts within the various checks. That's something I want to take a look at over the next days to really speed up the CI

ephys:
It should still overall have a positive impact across PRs as long as the lockfile is untouched
Do we want to merge this, or do we want to enable it in install-and-build in https://github.com/sequelize/sequelize/pull/14171? (it may be redundant)

WikiRik:
I'll add it to #14171 to see if it helps, otherwise we'll not do it at all. I'll close this PR

PR #14171, opened 2022-02-25
[WIP] ci(artifact): add build artifacts

WikiRik:
### Pull Request Checklist

### Description Of Change

This PR will hopefully speed up CI a lot by introducing build artifacts that are made once and used for all other jobs, minimizing duplication throughout the CI.

This is in draft because I'm still running tests on it.

### Todos

- [ ] Testing


ephys:
TypeScript: it looks like `yarn add` causes a full rebuild of all packages with a build step (like ibm_db), *and* runs sequelize's build script (`npm run build && husky install`)

It's a bit hacky but we could do `npm i --no-save typescript@4.1 --no-audit` instead. It doesn't seem to be rebuilding ibm_db, and it doesn't run our `build` script. `no-audit` reduces the install time drastically too,

WikiRik:
> TypeScript: it looks like `yarn add` causes a full rebuild of all packages with a build step (like ibm_db), _and_ runs sequelize's build script (`npm run build && husky install`)

Yes, the build script is not a huge thing since not letting it run is probably going to cost more than it will benefit in CI but the full rebuild is what we should avoid.

> It's a bit hacky but we could do `npm i --no-save typescript@4.1 --no-audit` instead. It doesn't seem to be rebuilding ibm_db, and it doesn't run our `build` script. `no-audit` reduces the install time drastically too.

But that's using npm. I quickly searched for a no-save alternative for yarn v1 but didn't find anything yet. I'll still change it for now but preferably I want to remain using yarn here

WikiRik:
I still want to change the TypeScript install back to yarn in a future PR but this --no-save does not impact the other dependencies so it's fine for now. I will improve that in a future PR. For now I'm happy that this reduces every check by about 2 minutes

ephys:
Using npm for that specific command shouldn't have a negative impact as npm seems to only be touching the typescript package (and eventual dependencies) the same way yarn would have. And in both scenarios we would have been bypassing the lockfile

ephys:
We should consider doing the same for pg-native

WikiRik:
> Using npm for that specific command shouldn't have a negative impact as npm seems to only be touching the typescript package (and eventual dependencies) the same way yarn would have. And in both scenarios we would have been bypassing the lockfile

Yes indeed. That's why it's fine for now. I just like the consistency for using yarn everywhere

ephys:
Looks like while doing it for TypeScript worked fine, doing it for pg-native doesn't. npm doesn't update the module tree correctly :(
Definitely a hack, let's keep it for TypeScript for now as long as it's documented

WikiRik:
Only on Node 12 though so I'll just return to `yarn add` for Node 12 and keep `npm install --no-save` for Node 16

ephys:
Really appreciating the performance boost 🥰

WikiRik:
Okay, I think it's done now. Next steps for further performance boost is related to DB2 but that's for a later moment

.github/workflows/ci.yml
```diff
@@ -146,12 +199,18 @@ jobs:
...
+      - uses: actions/download-artifact@v2
+        with:
+          name: install-build-artifact-node-${{ matrix.node-version }}
+      - name: Extract artifact
+        run: tar -xf install-build-node-${{ matrix.node-version }}.tar
+      - name: Install pg-native (Node 12)
+        run: yarn add pg-native --ignore-engines
+        if: matrix.native && matrix.node-version == 12
+      - name: Install pg-native (Node 16)
+        run: npm install --no-save --no-audit pg-native
```
ephys, line 212: we should document this npm-based hack to be safe

ephys (dismissed):
Other than that, good to go

.github/workflows/ci.yml
```diff
@@ -12,17 +12,70 @@ env:
...
       - run: yarn lint-no-fix
       - run: yarn lint-docs
+  unit-test:
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [12, 16]
+    name: Unit test all dialects (Node ${{ matrix.node-version }})
+    runs-on: ubuntu-latest
+    needs: [ install-and-build, lint ]
```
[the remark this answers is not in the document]
WikiRik: Oh yes, that's back from when there were separate jobs for Node 12 and 16. Good catch

.github/workflows/ci.yml
WikiRik: I've included `# This step uses npm instead of yarn to minimize the time needed. See #14171`. Is that enough?

ephys: Yup!

PR #14173, opened 2022-02-25
docs(legacy): clarify implications on models without primary keys

dilyanpalauzov:
### Pull Request Checklist

- [x] Documentation update is included (if this change modifies existing APIs, or introduces new ones)?

### DIFF
--- a/docs/manual/other-topics/legacy.md
+++ b/docs/manual/other-topics/legacy.md
@@ -52,7 +52,7 @@ Collection.init({
 }, { sequelize });
 ```
 
-And if your model has no primary key at all you can use `Model.removeAttribute('id');`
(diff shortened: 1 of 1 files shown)


WikiRik:
Thanks for the PR! We'll take a look at it soon.
Don't worry about the failing tests. The MariaDB tests have failed since this evening suddenly and the DB2 tests are very flaky

ephys:
We should consider resolving https://github.com/sequelize/sequelize/pull/13819 first. I've also found it weird that it stored the query used to select the model on the Model instead of the instance.
My opinion of the solution is that using instance methods that rely on the primary key when there are no primary keys is dangerous and prone to introducing bugs. I think we forbid using methods that are PK-dependent if there are no primary keys, and only allowing to use static methods.

dilyanpalauzov:
I opened this case, as I assumed that https://github.com/sequelize/sequelize/pull/13819 will not be resolved any time soon.  Once the other case is resolved, this documentation change shall not be necessary.

dilyanpalauzov:
Integrating the proposed change in the documentation will save one day struggling with undocumented bugs.  At least it costed me one day and by updating the documentation I am contributing to prevent others from having the same troubles.

Are there compelling reasons against integrating the proposed change?

ephys:
I'll close this PR since https://github.com/sequelize/sequelize/pull/13819 is its replacement

PR #14181, opened 2022-02-27
feat: reserve `::`, `->` syntax in attribute names

ephys:
### Pull Request Checklist

### Description Of Change

A PR similar to https://github.com/sequelize/sequelize/pull/13992

This PR forbids using `::` in attribute names (not column names!), because `::` can be used in some places to cast an attribute.

Example:

```typescript
User.findOne({
  'id::STRING': '1'
});

// SELECT * FROM users WHERE CAST(id AS STRING) = '1'
```

Note: the support for this syntax is spotty. I plan on expanding it through https://github.com/sequelize/sequelize/pull/14022 & https://github.com/sequelize/sequelize/pull/14020

---

This PR also reserves `->` as it is used for nested associations in the SQL we generate

Finally, this PR tweaks the change introduced in https://github.com/sequelize/sequelize/pull/13992: Instead of forbidding attribute names that start & end with `$`, we now forbid attribute names that start *or* end with `$`


WikiRik (approved):
Looks good. Any other special syntax that we want to reserve in v7?

ephys:
I actually have a last one:

Right now we're forbidding attributes starting *and* ending with `$`.

If we have a JSON object that starts with `$`, then `$a.b$` could both be interpreted as `"a"."b"` (accessing 'b' on include 'a'), or `"$a" #>> {b}` (accessing key 'b' on json attribute '$a')

ephys:
Some queries generate the character sequence `->` which may clash with attribute names

```typescript
      testSql('$offer.organization.id$', {
        [Op.col]: 'offer.user.organizationId',
      }, {
        default: '[offer->organization].[id] = [offer->user].[organizationId]',
      });
```

We may want to forbid that one too

Not sure why it generates `[offer->organization].[id]` instead of `[offer.organization].[id]`

ephys:
Updated

test/unit/model/define.test.js
```diff
@@ -49,9 +49,15 @@ describe(Support.getTestDialectTeaser('Model'), () => {
     it('should throw when the attribute name is ambiguous with $nested.attribute$ syntax', () => {
       expect(() => {
         current.define('foo', {
-          $id$: DataTypes.INTEGER,
```
WikiRik, line 52: Maybe we still want to test $id$?

WikiRik (approved):
Looks good. I have one suggestion for a possible extra test but I'm fine with it not being included

test/unit/model/define.test.js
ephys: The code would have to be a bit convulter to have $id and id$ throw but not $id$ but I can add it if you prefer :)

WikiRik: No, it's fine like this. Let's just merge it

PR #14194, opened 2022-03-03
fix(mssql): stop to automatically add `order by` when there is already one

### LINKED ISSUE #14185: MSSQL - strange behavior with limit and already specified order
## Issue Creation Checklist

- [X] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Issue Description

Hi guys 👋, I just want to confirm with you if I found a bug or if I just do something wrong.

Let's say I have 2 models defined like this:
```js
const City = sequelize.define(
  'city',
  {
    id: {
      type: DataTypes.INTEGER,
      primaryKey: true,
      autoIncrement: true,
      allowNull: false,
    },
    name: {
      type: DataTypes.STRING,
      allowNull: false,
    },
  },
  {
    tableName: 'city',
    timestamps: false,
  },
);

const Address = sequelize.define(
  `address`,
  {
    id: {
      type: DataTypes.INTEGER,
      autoIncrement: true,
      primaryKey: true,
      allowNull: false,
    },
    postalCode: {
      type: DataTypes.STRING,
    },
  },
  {
    tableName: 'address',
    timestamps: false,
  },
);

City.hasMany(Address);
Address.belongsTo(City);
```

I have this findAll query 👇, the goal is pretty simple, I just want to make a leaderboard by the most used `city` in `addresses`. 
```js
const query = {
  attributes: [
    [col('city.name'), 'city__grouped__'],
    [fn('count', '*'), '__aggregate__'],
  ],
  group: ['city.name'],
  include: [
    {
      association: 'addresses',
      attributes: [],
      include: [],
    },
  ],
  limit: 5,
  order: [[fn('count', '*'), 'DESC']],
  subQuery: false,
  raw: true,
};
```

This is where you can help me determine if it's a bug or if I did something wrong.
The call to the findAll function result to this query 👇
```sql
SELECT
  [city].[name] AS [city__grouped__],
  COUNT(*) AS [__aggregate__]
FROM
  [city] AS [city]
  LEFT OUTER JOIN [address] AS [addresses] ON [city].[id] = [ddresses].[cityId]
GROUP BY
  [city].[name]
ORDER BY
  COUNT(*) DESC,
  [city].[id] OFFSET 0 ROWS FETCH NEXT 5 ROWS ONLY;
```
In fact this query result to a SQL error that say:
`Column "city.city_id" is invalid in the ORDER BY clause because it is not contained in either an aggregate function or the GROUP BY clause.`

I deep dive in the code and I found something strange here https://github.com/sequelize/sequelize/blob/abdb8967ef1fa458e1e36cd641ea140874409e7d/src/dialects/mssql/query-generator.js#L993-L1006

Sorry about the long message, but my question is: **Why [city].[id] is added if I specify an order in my query?** 

let me know if it not clearer enough.
Thank's for your time 🙏 have a great day.

- [X] Yes. This issue only applies to the following dialect(s): MSSQL
- [ ] I don't know.

arnaud-moncel:
### Pull Request Checklist

### Description Of Change

Should fix #14185 

Should not auto adding `order by` clause when a `group` option is provided. Should avoid sql error
`[column] is invalid in the ORDER BY clause because it is not contained in either an aggregate function or the GROUP BY clause.`

### DIFF
--- a/src/dialects/mssql/query-generator.js
+++ b/src/dialects/mssql/query-generator.js
@@ -976,6 +976,10 @@ export class MsSqlQueryGenerator extends AbstractQueryGenerator {
       return '';
     }
 
+    if (!options.limit && !options.offset) {
+      return '';
+    }
+
     const offset = options.offset || 0;
     const isSubQuery = options.subQuery === undefined
       ? options.hasIncludeWhere || options.hasIncludeRequired || options.hasMultiAssociation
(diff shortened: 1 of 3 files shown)


ephys:
Thank you for the PR!

Instead of the unit test that specifically tests `addLimitAndOffset`, I'd rather have a test in `test/unit/sql/select.test.js` that tests a minimal reproducible findAll similar to what you have in https://github.com/sequelize/sequelize/issues/14185

arnaud-moncel:
I don't understand why some tests are failing 🤷,
I think the modifications I made should not impact them, I'm right?

ephys:
You're targetting the v6 branch which has some failing tests with the newer docker images, but I don't think it's related to your PR

arnaud-moncel:
Hi 👋 I do it like this at the beginning. And one integration tests fail. 👇

https://github.com/sequelize/sequelize/blob/e974e202ca755a008f450c88123fc166a5497bb2/test/integration/model/findAll/groupedLimit.test.js#L118-L138

I deep dive in the test and, I make test on my end to understand what happens. And the result is pretty strange 😅.

I extracted the subquery created by the queryGenerator and I executed it manually.
The first attempt without limit / offset:
image

We are agree about, if i add limit 3 the result should be `id: [3, 4, 5]`.
But in reality the result is:
image

IMO that’s why there is a "Flaky" test and that’s why the code is like this.

I don’t know if it clear.

arnaud-moncel:
So what do we do? 😅

ephys:
I would be consider the test to be at fault here. 
The ORDER BY provided in that test is not stable, multiple people have the same age.

I think it's the responsibility of the developer to provide a stable ORDER BY. Trying to fix their ORDER BY ourselves is going to be incompatible with aggregation functions and since users can use literals, it's not possible for us to determine whether an aggregation function is used to skip 'fixing' the order by.

I think the most we should do is provide a default ORDER BY if one is needed and none was provided. The rest is up to the developer :)

I would change the test to

```typescript
order: [ 
  Sequelize.fn('ABS', Sequelize.col('age')),
  'id',
], 
```

ping @sequelize/maintainers if someone wants to weigh in on this issue

ephys:
Oh no I've just noticed you opened this PR against v6.
This is very likely to be a breaking change, we can't include it in v6. Could you add it to `main` instead?

ephys:
> I must re do in another PR targeting v6 the first version of my fix?

I think it may be best. Let's have a temporary workaround in v6 to resolve your issue, and a proper fix in v7

WikiRik:
@arnaud-moncel can you resolve the conflicts? Then we'll take another look at this

ephys (changes requested):
The unit tests are failing, can you re-request a review once the tests are green?

evanrittenhouse (dismissed):

Issue #14196, opened 2022-03-03
convert multiple dialect checks to use Array.includes


markdirish:
Per comments in #14009, when tests check whether a dialect is or isn't being used, it should use `[].includes` when more than one dialect is checked instead of chaining `===`.

PR #14197, opened 2022-03-03
refactor: convert multiple dialect checks to use Array.includes

markdirish:
Signed-off-by: Mark Irish <mirish@ibm.com>

### Pull Request Checklist

### Description Of Change

Fixes https://github.com/sequelize/sequelize/issues/14196

I wasn't sure if this should only apply to current instances where multiple dialects are checked, or if single dialects should be checked as well (more function overhead, but future-proof changes?)

### DIFF
--- a/test/integration/data-types.test.js
+++ b/test/integration/data-types.test.js
@@ -183,7 +183,7 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
     // mssql has a _bindParam function that checks if STRING was created with
     // the boolean param (if so it outputs a Buffer bind param). This override
     // isn't needed for other dialects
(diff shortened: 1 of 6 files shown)


WikiRik:
> I wasn't sure if this should only apply to current instances where multiple dialects are checked, or if single dialects should be checked as well (more function overhead, but future-proof changes?)

To be honest I don't think it really matters. We'll be splitting dialects from core and need to find a different solution for this later.

WikiRik (approved):
Haven't checked if there are unchanged left but the changes look good

markdirish:
I searched in `test` directory for every instance where `dialect ===` was used, then looked to see if there were multiple dialects checked. Then I checked for just `===` and saw if there were any remaining. And didn't see any more.

PR #14199, opened 2022-03-05
fix(postgres, sqlite): map conflictFields to column names in Model.upsert

### LINKED ISSUE #14150: SQL field name used instead of property name in conflictFields
## Issue Creation Checklist

### SSCCE

```typescript
const { createSequelizeInstance } = require("./dev/sscce-helpers");
const { Model, DataTypes } = require(".");

class User extends Model {}
User.init(
  {
    userName: {
        type: DataTypes.STRING,
        field: 'user_name',
        unique: 'user'
    }
  },
  { sequelize, modelName: "user" }
);

(async () => {

  await User.upsert({
    userName: "janedoe"
  }, {
    conflictFields: ["userName"]
  });

  await sequelize.close();
})();
```

I expect request like `INSERT INTO "user" ("user_name") VALUES ($1) ON CONFLICT ("user_name") DO UPDATE SET "user_name" = EXCLUDED."user_name" RETURNING "user_name";`

This one generates request like `ON CONFLICT ("userName") ...`.

### Additional context

If I use `conflictFields: ["user_name"]`, TypeScript won't compile because it expects `Array<Attributes<User>>`, and `user_name` is not a valid attribute name.

### Environment

- Sequelize version: 6.16.1
- Node.js version: 12.22.10
- If TypeScript related: TypeScript version: 4.5.5

- [x] I think this problem happens only for the following dialect(s): PostgreSQL

AllAwesome497:
### Pull Request Checklist

### Description Of Change
fixes an issue where the attributes provided in conflictFields wouldn't be mapped to column names.

Closes #14150


test/integration/model/upsert.test.js
```diff
@@ -768,6 +775,41 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+            const [updatedMembership] = await Memberships.upsert(
+              {
+                user_id: 1,
+                group_id: 6,
+                otherID: originalMembership.otherID,
+                permissions: 'admin',
+              },
+              {
+                fields: ['user_id', 'group_id'],
+                conflictFields: ['other_id'],
```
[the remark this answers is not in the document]
AllAwesome497: oops, that's a typo.  Fixed!

src/model.js
```diff
@@ -2613,6 +2613,11 @@ class Model {
     const updateValues = Utils.mapValueFieldNames(updatedDataValues, options.fields, this);
     const now = Utils.now(this.sequelize.options.dialect);
 
+    options.conflictFields = options.conflictFields
+      && options.conflictFields.map(
+        attr => (this.rawAttributes[attr] && this.rawAttributes[attr].field) || attr,
+      );
+
```
ephys, line 2620: just a small code style change

```suggestion
    options.conflictFields = options.conflictFields?.map(
      attr => (this.rawAttributes[attr]?.field) || attr,
    );

```

This piece of code is used a lot in sequelize. We could add a `Model.getColumnName(attributeName: string): string` util which does this

ephys (dismissed):
Only change request :)

src/model.js
```diff
@@ -2544,6 +2544,10 @@ Specify a different name for either index to resolve this issue.`);
     const updateValues = Utils.mapValueFieldNames(updatedDataValues, options.fields, this);
     const now = Utils.now(this.sequelize.dialect);
 
+    options.conflictFields = options.conflictFields?.map(
+      attr => (this.rawAttributes[attr]?.field) || attr,
```
ephys, line 2548: Since this is a breaking change, should we rename `conflictFields` to `conflictAttributes` in this PR? (see comment in https://github.com/sequelize/sequelize/pull/13419)

ephys (changes requested):

WikiRik:
@muraliprajapati Some further updates might need to be done in order to merge this since it's been a year since the last update. We've done quite some work internally since then. Feel free to open a new PR that adds this feature and new tests to the current codebase

PR #14212, opened 2022-03-08
fix(types): Instance can be null for update

maxwrlr:
### Pull Request Checklist

### Description Of Change

Just like insert and delete, the update function of the query-interface shouldn't require an instance.


ephys:
Thank you for the PR but I don't think this is a direction we should take. This sounds like a less safe `bulkUpdate` method that has no way of specifying the attributes.

What is your use case for this that can't be done using bulkUpdate?

ephys:
Good news :)

Issue #10831, opened 2019-04-23
Sequelize v5 renames `addIndex(... { indicesType: ...})` to `addIndex(... { type: ...})`


frankthelen:
Sequelize v4 used to support `addIndex(... { indicesType: ...})`,
which has been renamed to `addIndex(... { type: ...})` in Sequelize v5.

You don't get an error during `addIndex` -- in fact, the index is created but, e.g., the UNIQUE constraint is missing. Which is tricky to find sometimes.

## What was unclear/insufficient/not covered in the documentation

Neither was this marked deprecated in v4 (at least not with console warning as far as I can tell).
Nor is this stated in the "Upgrade to v5" section in the docs.
Either would have saved me some time when migrating from v4 to v5.

### If possible: Provide some suggestion how we can enhance the docs

Put a hint into "Upgrade to v5" section would be great.
This is clearly a breaking interface change between v4 and v5.

Thank you.

SimonSchick:
See also https://github.com/sequelize/sequelize/pull/10844

PR #14216, opened 2022-03-09
docs: update renamed field

gabrielmlinassi:
Update field to **type** instead of **indicesType** since it was renamed since v5  as it says here on issue _[#10831](https://github.com/sequelize/sequelize/issues/10831) Sequelize v5 renames addIndex(... { indicesType: ...}) to addIndex(... { type: ...})_


ephys (dismissed):
Thanks :)

bl0cknumber (approved):

PR #14220, opened 2022-03-10
refactor(data-types): convert db2 data-types to TS

AllAwesome497:
### Description Of Change
converts db2-specific data-types and tests to TS.


src/dialects/db2/data-types.ts
```diff
@@ -0,0 +1,380 @@
...
+    return this._hexify(hex);
+  }
+
+  protected _hexify(hex: string) {
+    return `x'${hex}'`;
+  }
+}
+
+export class STRING extends BaseTypes.STRING {
+  static readonly escape: false;
```
AllAwesome497, line 124: This is super spooky imo.

IIRC I determined this was safe, though I don't remember exactly why.

I'm not entirely certain that `_stringify` is always going to be called when formatting a user's input, but assuming it is this is safe?

src/dialects/db2/data-types.ts
```diff
@@ -0,0 +1,380 @@
...
+
+    return typeof value === 'boolean' ? value : null;
+  }
+
+  static readonly parse = this.prototype._sanitize;
+}
+
+export class UUID extends BaseTypes.UUID {
+  toSql() {
+    return 'CHAR(36) FOR BIT DATA';
```
AllAwesome497, line 206: Not strictly DB2 related: This seems to be what we do in other dialects without proper UUID support too, but ideally we'd store it as a 16 byte buffer / binary instead.

This would reduce the size of uuids by >50% and in theory speed things up a lil' too.

src/dialects/db2/data-types.ts
```diff
@@ -0,0 +1,380 @@
...
+      return date.format(`YYYY-MM-DD HH:mm:ss${msec}`);
+    }
+
+    return date.format('YYYY-MM-DD HH:mm:ss');
+  }
+
+  static parse(value: RawTypeOf<BaseTypes.DATE>) {
+    if (typeof value !== 'string') {
+      // @ts-expect-error what type is value?
+      value = value.string();
```
AllAwesome497, line 254: I have no idea what type this is.  In mysql this was the type the underlying lib returned and parse was called via mysql's `connection-manager` but it looked like db2 didn't use this (at least not in the same way mysql does) when I looked.

src/dialects/db2/data-types.ts
```diff
@@ -0,0 +1,380 @@
...
+      this._zerofill = undefined;
+    }
+  }
+}
+
+export class ENUM<Member extends string> extends BaseTypes.ENUM<Member> {
+  toSql() {
+    return 'VARCHAR(255)';
+  }
+}
```
AllAwesome497, line 354: Should we drop support for this entirely?  It seems misleading to support it as a string.

ephys: I agree, its constructor should throw & say to use DataTypes.STRING + the `isIn` column validator instead

src/dialects/db2/data-types.ts
ephys: Sounds good to me :)

src/dialects/db2/data-types.ts
AllAwesome497: My only concern with this is that it'll be pretty annoying to change w/ an upgrade.

Thoughts on leaving this for now and then in a different change add an  `ENUM_VIRTUAL` datatype + deprecate fake enum usage?

AllAwesome497:
Closing in favor of #14505.

PR #14221, opened 2022-03-10
Update model.d.ts

arniu:
### Pull Request Checklist

### Todos

- [ ] 
- [ ] 
- [ ]


src/model.d.ts
```diff
@@ -1373,13 +1373,13 @@ export interface ModelAttributeColumnOptions<M extends Model = Model> extends Co
    * What should happen when the referenced key is updated. One of CASCADE, RESTRICT, SET DEFAULT, SET NULL or
    * NO ACTION
    */
-  onUpdate?: string;
+  onUpdate?: 'CASCADE' | 'RESTRICT' | 'SET DEFAULT' | 'SET NULL' | 'NO ACTION';
```
ephys, line 1376: I've made the same change in my [documentation update](https://github.com/sequelize/sequelize/blob/feature/typedoc/src/associations/base.d.ts#L42) PR

I'm happy to see this merged first, but could you extract it to a type alias so the new documentation can reference it? :)

Their lowercase variants are supposed to be valid to (i.e. 'cascade'), but I'm fine with normalizing to uppercase for every user, but we need to update its usage in `test/types/query-interface.ts` otherwise typing tests will faill

ephys (changes requested):

PR #14233, opened 2022-03-11
docs: clarify implications on models without primary keys

dilyanpalauzov:
Puts warnings in the documentation for sequelize-usage when the tables have no primary key.

Targetting v6 branch, related to https://github.com/sequelize/sequelize/pull/14173#issuecomment-1065388719

### DIFF
--- a/docs/manual/other-topics/legacy.md
+++ b/docs/manual/other-topics/legacy.md
@@ -52,7 +52,7 @@ Collection.init({
 }, { sequelize });
 ```
 
-And if your model has no primary key at all you can use `Model.removeAttribute('id');`
+And if your model has no primary key at all you can use `Model.removeAttribute('id');`.  Instances without primary keys, retrieved using `findOne()`, store the exact query criteria in the model class: once many instances are queried simultaneously, doing instance.save() on any of the instances, will always update the last retrieved instance.  The mitigation is to use `instance.update()` and provide there the search criteria.  [#13819](https://github.com/sequelize/sequelize/pull/13819) proposes a fix, storing the query per instance.
 
 ## Foreign keys
 


dilyanpalauzov:
This is not a duplicate, as the target branches are different.

docs/manual/other-topics/legacy.md
```diff
@@ -52,7 +52,7 @@ Collection.init({
 }, { sequelize });
 ```
 
-And if your model has no primary key at all you can use `Model.removeAttribute('id');`
+And if your model has no primary key at all you can use `Model.removeAttribute('id');`.  Instances without primary keys, retrieved using `findOne()`, store the exact query criteria in the model class: once many instances are queried simultaneously, doing instance.save() on any of the instances, will always update the last retrieved instance.  The mitigation is to use `instance.update()` and provide there the search criteria.  [#13819](https://github.com/sequelize/sequelize/pull/13819) proposes a fix, storing the query per instance.
```
ephys, line 55: I'd prefer to include a disclaimer instead, as the ability to call these instance methods on models that don't have a primary keys will be removed in a future release.

```suggestion
And if your model has no primary key at all you can use `Model.removeAttribute('id');`.

Instances without primary keys can still be retrieved using `Model.findOne` and `Model.findAll`.\
While it's currently possible to use their instance methods (`instance.save`, `instance.update`, etc…), doing this is discouraged, deprecated, and will lead to subtle bugs.

If your model has no primary keys, you need to use the static equivalent of the following instance methods, and provide your own `where` parameter:
- `instance.save`: `Model.update`
- `instance.update`: `Model.update`
- `instance.reload`: `Model.findOne`
- `instance.destroy`: `Model.destroy`
- `instance.restore`: `Model.restore`
- `instance.decrement`: `Model.decrement`
- `instance.increment`: `Model.increment`
```

ephys (changes requested):

docs/manual/other-topics/legacy.md
dilyanpalauzov: You can adjust the text as you want. The proposed changes are a matter of personal style.  I do not see a need to make further changes myself.

ephys: I'll close this PR in favor of https://github.com/sequelize/website/pull/5 then

PR #14239, opened 2022-03-12
refactor(data-types): convert sqlite data-types to TS

AllAwesome497:
 

### Description Of Change
converts sqlite-specific data-types to TS.


WikiRik (commented):
In #14240 you have
```ts
export { JSONTYPE as JSON };
```
Don't we need to export `JSONTYPE as JSON` and `DOUBLE as 'DOUBLE PRECISION'` here?

PR #14264, opened 2022-03-19
Docs: await without async

Bishwodahal:
updated docs 
await without async in [getting-started](https://sequelize.org/v6/manual/getting-started.html)


ephys:
I don't think we necessarily need to always wrap our await calls in an async function for two reasons:

- they's small code snippets, they don't need to show the whole picture
- top level await is available nowadays, so the example is valid (in esm)

Bishwodahal:
Ok 😀
I saw a post regarding this on other school club so I thought it would help newbies

PR #14277, opened 2022-03-27
feat!: make `set` association method delete old associated entity for non-null FKs 

evanrittenhouse:
### Pull Request Checklist

- [x ] Have you added new tests to prevent regressions?
- [ x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)?

### Description Of Change

Added a check to set() in has-many to update or destroy rows based on if the foreign key can be null or not.

### Todos

- [ ] Need to work with someone to prevent the regression error mentioned in commit 3d21cdc12c31995d79f3c32fe9784285d438c942


src/associations/has-many.js
```diff
@@ -346,16 +346,25 @@ class HasMany extends Association {
...
-        {
-          ...options,
-          where: updateWhere,
-        },
-      ));
+      // update or delete obsolete associations depending on foreign key
+      let udMethod;
+      const udArgs = [{ ...options, where: updateWhere }];
+      const foreignKeyIsNullable
+        = this.target.rawAttributes[this.identifierField].allowNull;
```
ephys, line 358: `identifierField` is a column name, `rawAttributes` is indexed by attribute name. I think this should be

```suggestion
        = this.target.rawAttributes[this.foreignKey].allowNull;
```

I don't think `allowNull` is normalized, so it could be undefined too (which defaults to true). The solution should be:

```suggestion
        = this.target.rawAttributes[this.foreignKey].allowNull !== false;
```

src/associations/has-many.js
```diff
@@ -346,16 +346,25 @@ class HasMany extends Association {
...
+
+      if (foreignKeyIsNullable) {
+        udMethod = this.target.update.bind(this.target.unscoped());
+        udArgs.unshift(update);
+      } else {
+        udMethod = this.target.destroy.bind(this.target.unscoped());
+      }
```
ephys, line 365: This kind of code will be impossible to type during the TS migration, prefer this instead:

```suggestion
      const target = this.target.unscoped();
      if (foreignKeyIsNullable) {
        promises.push(target.update({
          [this.foreignKey]: null,
        }, {
          ...options,
          where: updateWhere,
        }));
      } else {
        promises.push(target.destroy({
          ...options,
          where: updateWhere,
        }))
      }
```

I think it's more readable too

test/unit/associations/has-many.test.js
```diff
@@ -212,6 +212,85 @@ describe(Support.getTestDialectTeaser('hasMany'), () => {
...
+      NullableAddress.belongsTo(NullableUser);
+      NullableUser.hasMany(NullableAddress);
+      await current.sync();
+
+      const jane = await NullableUser.create({
+        username: 'jane',
+      });
+
+      const addr1 = await jane.createNullableAddress({ street: 'st' });
+      let addresses = await jane.getNullableAddresses();
```
ephys, line 277: Here you get the addresses a first time, but never do anything with them. This line can be removed

test/unit/associations/has-many.test.js
```diff
@@ -212,6 +212,85 @@ describe(Support.getTestDialectTeaser('hasMany'), () => {
...
+        username: 'jane',
+        UserId: 1,
+      });
+
+      const addr1 = await jane.createAddress({ street: 'st' });
+      const addr2 = await Address.build({
+        street: 'st2',
+        UserId: jane.id,
+      });
+      await addr2.save();
```
ephys, line 248: build isn't async, `create` is (create = build+save)

```suggestion
      const addr2 = await Address.create({
        street: 'st2',
        UserId: jane.id,
      });
```

Can you also put these inside a Promise.all to do them in parallel?

```typescript
      const [addr1, addr2] = await Promise.all([
        jane.createAddress({ street: 'st' }),
        Address.create({
          street: 'st2',
          UserId: jane.id,
        }),
      ]);
```

test/unit/associations/has-many.test.js
```diff
@@ -212,6 +212,85 @@ describe(Support.getTestDialectTeaser('hasMany'), () => {
...
+      const Address = current.define('Address', {
+        street: DataTypes.STRING,
+      });
+
+      Address.belongsTo(User, {
+        foreignKey: {
+          allowNull: false,
+        },
+      });
+      User.hasMany(Address);
```
ephys, line 235: This belongsTo association isn't necessary for this test. I think this should be sufficient:

```suggestion
      User.hasMany(Address, {
        foreignKey: {
          allowNull: false,
        },
      });
```

test/unit/associations/has-many.test.js
```diff
@@ -212,6 +212,85 @@ describe(Support.getTestDialectTeaser('hasMany'), () => {
...
+        userId: jane.id,
+      });
+      await addr2.save();
+
+      await jane.setNullableAddresses([addr2]);
+
+      addresses = await jane.getNullableAddresses();
+
+      const newAddresses = await NullableAddress.findAll();
+      expect(newAddresses.length).to.equal(2);
```
ephys, line 290: this test should check that addr1's UserId has been set to null :)

ephys (changes requested):
Thank you for the PR :)

Most of my comments in the first test are also valid for the second test.

The same issue also occurs in `HasMany.remove`. I think `HasMany.set` could use `HasMany.remove` to disassociate the obsolete associations.

That's actually what I did in my TS migration :)
image

The same issue is also present in `HasOne.set` and `HasOne.create`, could you look into these too?

src/associations/has-many.js
evanrittenhouse: Definitely agree on readability! Also really like the dynamic property change. Didn't think of that, but super clean.

evanrittenhouse:
@ephys Incorporated your comments (really like the dynamic typing change in has-many.js). 

A little unclear on this part - mind elaborating? I notice that you keep the `target.destroy()` call in has-many.js, so not sure if we want to use that or `HasMany.remove`. Also, what is the same issue that you refer to?

> The same issue also occurs in HasMany.remove. I think HasMany.set could use HasMany.remove to disassociate the 
> obsolete associations.

Once we decide on a solution here, I can definitely incorporate into `HasOne.set()` and `HasOne.create()`.

ephys:
Thanks, looking good :)

What I mean is that `HasMany.remove` will also fail if the foreign key is non-nullable.
We should add this new behavior to that method too, to be symetrical

If we look at the source code, the part of [`HasMany.set`](https://github.com/sequelize/sequelize/blob/main/src/associations/has-many.js#L344-L359) that handles removing obsolete associations is the same code as [`HasMany.remove`](https://github.com/sequelize/sequelize/blob/main/src/associations/has-many.js#L424-L439). Instead of duplicating your changes in both methods, I think `HasMany.set` should call `HasMany.remove` with the obsolete associations, and `HasMany.remove` should handle choosing between deleting or setting the foreign key to null

evanrittenhouse:
@ephys Got it - have refactored the logic to be in `HasMany.remove()`. If that looks good, I can move on to replicating this in `HasOne`. I'll have to look at how the removal works in that class

evanrittenhouse:
@ephys Sounds good. One question on design - how should I pass the obsolete associations? I'm unfamiliar with the `HasMany.remove()` call, and don't want to change the default interface for fear of breaking stuff. We can calculate the obsolete observations in `HasMany.remove()` instead?

ephys:
No, `set` needs to determine which associations are obsolete then pass them to remove :)

Take a look at the refactor I did in the migration branch:
[set](https://github.com/sequelize/sequelize/blob/ephys/ts-associations/src/associations/has-many.ts#L377) calls [remove](https://github.com/sequelize/sequelize/blob/ephys/ts-associations/src/associations/has-many.ts#L444-L468)

evanrittenhouse:
@ephys I'm running into an interesting problem here, with the `destroy()` call. Wondering if you can help out. Essentially, `Utils.mapWhereFieldNames()`, when being called in `destroy()`, formats the object incorrectly. It yields this: 
```
[Object: null prototype] {
  UserId: 1,
  id: [
    [Object: null prototype] {
      id: 1,
      street: 'st',
      createdAt: '2022-03-30 14:01:23.317 +00:00',
      updatedAt: '2022-03-30 14:01:23.317 +00:00',
      UserId: 1
    }
  ]
}
```
It seems like the `[Object: null prototype]` in the root id field throws an error in `dialects/abstract/query-interface.js`, when `bulkDelete()` is called. I've been investigating this for the past couple days but am coming up empty. It appears that this is causing   The actual error text is 

```
  1) [SQLITE] HasMany
       set
         should unassociate (delete) addr1 given a non-null constraint on userId:
     Error: Invalid value [Object: null prototype] {
  id: 1,
  street: 'st',
  createdAt: '2022-03-30 14:16:39.642 +00:00',
  updatedAt: '2022-03-30 14:16:39.642 +00:00',
  UserId: 1
}
      at Object.escape (lib/sql-string.js:58:11)
      at SQLiteQueryGenerator.escape (lib/dialects/abstract/query-generator.js:715:22)
      at /Users/evanrittenhouse/Desktop/projects/open_source/sequelize/lib/dialects/abstract/query-generator.js:1888:71
      at Array.map (<anonymous>)
      at SQLiteQueryGenerator._whereParseSingleValueObject (lib/dialects/abstract/query-generator.js:1888:52)
      at SQLiteQueryGenerator.whereItemQuery (lib/dialects/abstract/query-generator.js:1705:19)
      at SQLiteQueryGenerator.whereItemsQuery (lib/dialects/abstract/query-generator.js:1630:25)
      at SQLiteQueryGenerator.getWhereConditions (lib/dialects/abstract/query-generator.js:1973:19)
      at SQLiteQueryGenerator.deleteQuery (lib/dialects/sqlite/query-generator.js:195:28)
      at SQLiteQueryInterface.bulkDelete (lib/dialects/abstract/query-interface.js:401:59)
      at Function.destroy (lib/model.js:1816:42)
      at async HasMany.remove (lib/associations/has-many.js:282:7)
      at async Promise.all (index 0)
      at async HasMany.set (lib/associations/has-many.js:248:5)
      at async Context.<anonymous> (test/integration/associations/has-many.test.js:1601:7)



error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Have you seen this before and/or do you have any ideas on how to solve it?

evanrittenhouse:
@ephys mind checking these commits over? There is a regression failure that I commented out (`tests/integration/associations/has-many.test.js`, line 749) - this is more to draw attention to it to discuss why this should destroy rather than update the rows. Good news is that other than this line, the rest of the tests pass. I've also added a `destroy` field to `remove()`'s options, so we can destroy at any point in our choosing going forward.

test/integration/associations/has-many.test.js
```diff
@@ -745,8 +745,10 @@ describe(Support.getTestDialectTeaser('HasMany'), () => {
         await article.addLabel(label1.id);
         await article.setLabels([label2.id]);
         const labels = await article.getLabels();
-        expect(labels).to.have.length(1);
-        expect(labels[0].text).to.equal('label two');
+        expect(labels).to.have.length(2);
```
ephys, line 748: `article.setLabels([label2.id])` means that only label2 will be associated to article, so the new length of labels should be one. The original test is correct

src/associations/has-many.js
```diff
@@ -330,32 +330,20 @@ class HasMany extends Association {
...
       targetInstances = [];
+      options = options ? { ...options, destroy: true } : { destroy: true };
     } else {
       targetInstances = this.toInstanceArray(targetInstances);
     }
 
     const oldAssociations = await this.get(sourceInstance, { ...options, scope: false, raw: true });
     const promises = [];
-    const obsoleteAssociations = oldAssociations.filter(old => !targetInstances.some(obj => obj[this.target.primaryKeyAttribute] === old[this.target.primaryKeyAttribute]));
+    const obsoleteAssociations = oldAssociations.filter(old => !targetInstances.some(obj => obj[this.target.primaryKeyAttribute] === old[this.target.primaryKeyAttribute]) || old === null);
```
ephys, line 340:
```suggestion
    const obsoleteAssociations = oldAssociations.filter(old => !targetInstances.some(obj => obj[this.target.primaryKeyAttribute] === old[this.target.primaryKeyAttribute]));
```

old can't be null

src/associations/has-many.js
```diff
@@ -330,32 +330,20 @@ class HasMany extends Association {
...
-      };
-
-      promises.push(this.target.unscoped().update(
-        update,
-        {
-          ...options,
-          where: updateWhere,
-        },
-      ));
+    if (obsoleteAssociations.length > 0 || targetInstances == null) {
```
ephys, line 345:
```suggestion
    if (obsoleteAssociations.length > 0) {
```

targetInstances can't be null

src/associations/has-many.js
```diff
@@ -416,24 +404,35 @@ class HasMany extends Association {
...
    *
    * @returns {Promise}
    */
-  async remove(sourceInstance, targetInstances, options = {}) {
-    const update = {
-      [this.foreignKey]: null,
-    };
-
-    targetInstances = this.toInstanceArray(targetInstances);
+  async remove(sourceInstance, rawTargetInstances, options = {}) {
```
ephys, line 412: this should work both with raw instances and non-raw instances, because this method isn't just an internal method, it's the actual `x.removeAssociations(y)` method that get mixed in the model's prototype

```suggestion
  async remove(sourceInstance, targetInstances, options = {}) {
```

src/associations/has-many.js
```diff
@@ -330,32 +330,20 @@ class HasMany extends Association {
   async set(sourceInstance, targetInstances, options) {
     if (targetInstances === null) {
       targetInstances = [];
+      options = options ? { ...options, destroy: true } : { destroy: true };
```
ephys, line 333: why set the `destroy` option here?

If you want to let the user decide whether to delete the disassociated entity, the option should be set by the user
If you want to decide automatically, `remove` should determine that itself by checking if the foreign key is nullable

src/associations/has-many.js
```diff
@@ -416,24 +404,35 @@ class HasMany extends Association {
...
-    await this.target.unscoped().update(update, { ...options, where });
+    // update or delete based on the target's foreign key
+    const foreignKeyIsNullable = this.target.rawAttributes[this.foreignKey].allowNull !== false;
+    const target = this.target.unscoped();
+    if (foreignKeyIsNullable && !options.destroy) {
+      const update = {
+        [this.foreignKey]: null,
+      };
+
+      where[this.foreignKey] = sourceInstance.get(this.sourceKey);
```
ephys, line 429: the same condition should be used for both update and destroy, this should be moved back inside `where` for both branches like it was before

src/associations/has-many.js
```diff
@@ -416,24 +404,35 @@ class HasMany extends Association {
...
+
+      await target.update(update, { ...options, where });
+    } else {
+      const destroyWhere = where[this.target.primaryKeyAttribute];
+      await target.destroy({ ...options, where: destroyWhere });
```
ephys, line 434: both update & destroy should use the same where, they're modifying the same rows

```suggestion
      await target.destroy({ ...options, where });
```

ephys (changes requested):

src/associations/has-many.js
evanrittenhouse: @ephys Using this code, I get the error described [here](https://github.com/sequelize/sequelize/pull/14277#issuecomment-1083211443). This has to do with the `Utils.mapWhereFieldName()` call used in `destroy()` but not in `update()` - I was trying to work around that. I looked through the integration tests but couldn't find an example of destroying. Is there an example you could point me to on the correct usage? Sorry about that - I'll keep looking in the meantime.

evanrittenhouse: @ephys this passes the entire `where` object - in this test, for example:
```
    [Object: null prototype] {
      id: 1,
      street: 'st1',
      createdAt: '2022-04-09 16:14:22.563 +00:00',
      updatedAt: '2022-04-09 16:14:22.563 +00:00',
      UserId: 1
    }
``` 
to `queryInterface.bulkDelete`.
This throws an invalid value error. 

`queryInterface.bulkUpdate`'s where value takes the form 
```
[Object: null prototype] { id: [ 2 ] }
```
so I think that we need to reformat the `where` object before calling `destroy()`. Do you have any suggestions on how to do this?

ephys: Nope, I can confirm that the `where` that's used should be exactly the same in both update & destroy branches

If your object looks like this:
```
[Object: null prototype] {
  UserId: 1,
  id: [
    [Object: null prototype] {
      id: 1,
      street: 'st',
      createdAt: '2022-03-30 14:01:23.317 +00:00',
      updatedAt: '2022-03-30 14:01:23.317 +00:00',
      UserId: 1
    }
  ]
}
```

then you have something fishy going on with targetInstances, that's where the issue lies

evanrittenhouse: @ephys Gotcha. I've done some more digging - it appears that the object gets formed like that if `this.toInstanceArray(targetInstances)` in `remove()` gets called when the association is defined as such:
```
      User.hasMany(Address, {
        foreignKey: {
          allowNull: false,
        },
      });
```
with `allowNull: false`. Do you know why this would be? I can't find how that'd affect `toInstanceArray`.

ephys: I'm experiencing the same issue in my Association refactor PR #14280
The issue is that `set()` calls `remove()` with an array of raw entities
That array is then passed to `toInstanceArray`

`toInstanceArray` is supposed to convert these to models, but does it incorrectly because the way it checks if it's already a model is by doing `instanceof this.target` - raw entities will fail this check.

I'm fixing that in my refactor PR

evanrittenhouse: Gotcha - okay, mind letting me know when that PR is merged? I don't think I can proceed with tests passing as expected without that call working properly.

ephys: Sure! I'll ping you when it's ready. I'm going through the last remaining failing unit tests

evanrittenhouse: Hey @ephys just wondering if you've gotten a chance to get these to pass?

WikiRik: #14280 is ready, but it will take a while before we properly reviewed it. If you want you can temporarily base your branch on the one from that PR to check if things will work once we merge that

ephys: I don't know if git will be able to handle that gracefully given we're going to squash-merge #14280

But I can take care of copying your changes over once #14280 is merged, I think your PR is basically ready

evanrittenhouse: Oh sweet! Yeah that'd be great, thank you. Will be pretty busy for the next few days, but will look to pick up a new issue afterwards.

ephys: https://github.com/sequelize/sequelize/pull/14280 has been merged!
I'll see about updating this PR when I have a minute :)

WikiRik: @ephys I'm clearing up my notifications and trying to ping you in some of those so they're back on your radar

ephys:
@evanrittenhouse just checking before I do it: You don't mind that I take over this branch to update the PR?

evanrittenhouse:
@ephys not at all, go ahead.

klausXR:
@ephys is there any movement on this?

ephys:
@klausXR Yep! https://github.com/sequelize/sequelize/pull/15840

PR #14278, opened 2022-03-28
docs on JSON updated

Bishwodahal:
[here](https://sequelize.org/v6/manual/other-data-types.html) it is said that JSON is supported by other database but on data-types it is not mentioned.

I have a doubt on snowflake. It is said this about JSON. Does it mean it is supported by snowflake?
```
types: {
    postgres: [ 'json' ],
    mysql: [ 'JSON' ],
    mariadb: [ 'JSON' ],
    sqlite: [ 'JSON', 'JSONB' ],
    snowflake: [ 'JSON' ]
  },
```

### DIFF
--- a/src/data-types.js
+++ b/src/data-types.js
@@ -573,7 +573,7 @@ class DATEONLY extends ABSTRACT {
 }
 
 /**
- * A key / value store column. Only available in Postgres.
+ * A key / value store column. Only available in SQLite, MySQL, MariaDB and PostgreSQL.
  */
 class HSTORE extends ABSTRACT {
   validate(value) {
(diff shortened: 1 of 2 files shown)


src/data-types.js
```diff
@@ -573,7 +573,7 @@ class DATEONLY extends ABSTRACT {
 }
 
 /**
- * A key / value store column. Only available in Postgres.
+ * A key / value store column. Only available in SQLite, MySQL, MariaDB and PostgreSQL.
```
ephys, line 576: From my recent research, hstore is only available in Postgres. Do you have a link to the documentation of sqlite/mysql/mariadb for hstore? https://new.sequelizejs.com/docs/v7/other-topics/other-data-types#miscellaneous-datatypes

WikiRik:
I think the other changes are still valid, right?

ephys:
I think it's not the case either but to be verified.
According to the tests, `DataTypes.JSON` always generates the SQL type `JSON`, which is only supported in [postgres, mariadb and mysql](https://sequelize.org/docs/v7/other-topics/other-data-types/#json--jsonb)

JSONB is for sure postgres-only.

However I would like to expand the support for DataTypes.JSON because most dialects have at least built-in functions to treat text as json

Issue #13429, opened 2021-08-10
Documentation mismatch? No type property on ForeignKeyOptions


xaragen:
Working on a project I ran into the documentation for customizing the Foreign Key.
The docs read:

```javascript
// https://sequelize.org/master/manual/assocs.html
// For example, to use UUID as the foreign key data type instead of the default (INTEGER), you can simply do:

const { DataTypes } = require("Sequelize");

Foo.hasOne(Bar, {
  foreignKey: {
    // name: 'myFooId'
    type: DataTypes.UUID
  }
});
Bar.belongsTo(Foo);
```

Good enough, but when I went to work on it, it threw an error :
```
 Object literal may only specify known properties, and 'type' does not exist in type 'ForeignKeyOptions'
```
Looking at the files, it seems true. Is this correct behavior, missing type or incorrect documentation?

https://github.com/sequelize/sequelize/blob/56bb1d6ef9827f604d7bcef945abb7e213f2322d/types/lib/associations/base.d.ts#L37

**EDIT**:

After working on this, looks like it does work but the option is not part of the ForeignKey Options as described in the docs.
```js
Foo.belongsTo(Bar, {
    foreignKey: {
        name: 'myFooId',
    },
    as: 'bars',
    keyType: sequelize.DataTypes.STRING,
});
```

ephys:
This would be a good point to start an investigation: https://github.com/sequelize/sequelize/blob/main/src/associations/base.d.ts#L37
`ForeignKeyOptions` extends `ColumnOptions`, which doesn't include `type`
The only option bag that accepts `type` is `ModelAttributeColumnOptions`

The question is:
- can type be moved from `ModelAttributeColumnOptions` to `ColumnOptions` (what other methods use `ColumnOptions` and do they accept `type`)
- can foreign key extend `ModelAttributeColumnOptions`? Does it support all of these options
- or do we need to create an intermediary type

bera5186:
Hey @ephys ,

`ModelAttributeColumnOptions` extends `ColumnOptions` , so all the properties of `ColumnOptions` should be accessible via extending `ModelAttributeColumnOptions`.

Looks like extending `ModelAttributeColumnOptions` in `ForeignKeyOptions` would work, but I think creating intermediary type is a better option.

Correct me if I am wrong. Waiting for your inputs on this !

ephys:
https://github.com/sequelize/sequelize/issues/13429 fixes this issue

Issue #12782, opened 2020-10-22
BelongsToManyGetAssociationsMixinOptions incomplete


fabrykowski:
### What are you doing?

I am trying to query in TypeScript models linked via `belongsToMany` with conditions to the through model.

**Here is the link to the SSCCE for this issue:** papb/sequelize-sscce#123 

### What do you expect to happen?

The code to compile.

### What is actually happening?

The compilation fails

```
src/sscce.ts:161:5 - error TS2345: Argument of type '{ through: { where: { isManager: boolean; }; }; }' is not assignable to parameter of type 'BelongsToManyGetAssociationsMixinOptions'.
  Object literal may only specify known properties, and 'through' does not exist in type 'BelongsToManyGetAssociationsMixinOptions'.
```

### Additional context

n.a.

### Environment

- Sequelize version: 5.22.3 in SSCCE; 6.3.5 locally
- Node.js version: v14.13.0
- Operating System: OS X
- TypeScript version: 4.0.3

### How does this problem relate to dialects?

- [x] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [x] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

fabrykowski:
Still an issue

fabrykowski:
Bump

WikiRik:
Hi, sorry about needing to bump a few times. We have done some edits to this since 6.3.5. Does this still occur on the latest release? (6.11.0 at the time of writing)

ephys:
Fixed in https://github.com/sequelize/sequelize/pull/14280

Issue #11837, opened 2020-01-16
A more descriptive error when a model is being used without being initialized.


bobspace:
### Is your feature request related to a problem? Please describe.
Associating models via `hasOne`, `belongsTo`, etc. fails if the dependent model has not been initialized. This can happen if models are initialized in the wrong order, or a set has not been entirely initialized when an association method is called (as was the case in #11691).

### Describe the solution you'd like
Perhaps a quick check to a flag set when a model is initialized, and it throwing a descriptive error when method is called that requires a model to be initialized first.

### Why should this be in Sequelize
Noticing that various properties are undefined does not point the way to the `init` function, unless you're familiar with the lifecycle of a Sequelize model. A more descriptive error would save time and ultimately improve stability.

### Describe alternatives/workarounds you've considered
The alternative is suffering and wasting @papb's time until you happen to notice that your setup code is running out of order.

### Additional context
¯\_(ツ)_/¯

### Is this issue dialect-specific?

- [X] No. This issue is relevant to Sequelize as a whole.
- [ ] Yes. This issue only applies to the following dialect(s): XXX, YYY, ZZZ

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [X] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

papb:
> The alternative is suffering and wasting @papb's time until you happen to notice that your setup code is running out of order.

:laughing:

Issue #5144, opened 2016-01-04
Automatic creation of inverse side of bidirectional relationships


slavafomin:
Hello!

Thank you for your great work!

However, I'm new to Sequelize and right now it's association configuration looks somewhat counter-intuitive to me. I've used Doctrine before that for PHP. The corresponding documentation is also pretty difficult to read.

I'm trying to setup a bidirectional association between two models. Let's call them `User` and `Order`. After some time figuring out the difference between `hasOne`, `belongsTo` and `hasMany` associations types I've came up with the following config:

``` javascript

User.hasMany(Order, {
  foreignKey: 'userId',
  onDelete: 'CASCADE',
  hooks: true
});

Order.belongsTo(User, {
  foreignKey: 'userId',
  onDelete: 'CASCADE',
  hooks: true
});

```

With such configuration I'm getting the correct database structure, foreign keys, accessor methods and such, however I have to duplicate options, i.e.: `foreignKey`, `onDelete` and `hooks` in both models. In other case I'm getting two foreign keys with different names or cascade is not set correctly.

Could you tell me if I'm doing something wrong? Or is it an actual issue with Sequelize?

In Doctrine there is a concept of _owning side_ of the association and all options are defined only on one side of it, not on both.

Thank you!

mickhansen:
It's a requirement at the time since associations don't know about eachother.
Something i'd love to change but not quite sure how since users would still like do define aliases and stuff.

Perhaps a way to define that a relation in the inverse of another.

```js
User.Orders = User.hasMany(Order, {
  foreignKey: 'userId',
  onDelete: 'CASCADE',
  hooks: true
});

Order.belongsTo(User, {
  inverse: User.Orders,
  as: 'users'
});
```

Verdier:
+1 looks good ! Every time I create a relation, I ask my self about that... It's hard to predict what will take precedence (mostly for `CASCADE`).

slavafomin:
@mickhansen that looks good to me too. Doctrine solves this problem in the similar fashion.

As an alternative idea, maybe we could use just a single definition instead of two:

```javascript
User.hasMany(Order, {
  foreignKey: 'userId',
  onDelete: 'CASCADE',
  hooks: true,
  inverse: true // Will add singular mixins on target
});
```

However, I think that two definitions are more explicit, which is good in my opinion.

mickhansen:
@slavafomin That would work aswell in cases where no alias is defined for the relation.

[...]

PR #14280, opened 2022-03-28
refactor: migrate associations to TypeScript

ephys:
This PR migrates the 'association' folder to TypeScript.

This PR brings makes very important changes to how associations work in Sequelize:

# List of changes

- Association names have been normalized: All associations are named, and their name always corresponds to their value in `Model.associations.x`.
- Associations are unique: It's not possible to replace an association once defined (breaking change).
- Associations are immutable: Once defined, they cannot be changed (breaking change).
- The way associations are determined when using `include` has changed (breaking change).
- Includes are now strict.

## Changes to how foreign keys are created

Instead of duplicating code across all associations, `BelongsTo` is the only association capable of adding Foreign Keys. All other associations delegate to BelongsTo:

- `HasMany` creates its inverse `BelongsTo` association
- `HasOne` creates its inverse `BelongsTo` association
- `BelongsToMany` creates its inverse `BelongsToMany` association, both of which create a `HasMany` association between Through and Source, which creates a `BelongsTo`

## Association names have been normalized

Sequelize 6 would consider these two associations to be different:

```typescript
Project.belongsTo(User);
Project.belongsTo(User, { as: 'User' });
```

This worked most of the time, but it was inconsistent internally. Both associations shared the same name and `Project.associations.User` be equal to the last association to have been defined. 
Even more inconsistently, `project.getUser` and other mixin methods would belong to the first association to have been defined.

After this PR, the above code will throw an error saying that the alias is already in use and to use another one.

## Associations are unique

The following code will now throw, because you can't replace an association that has already been defined anymore (your codebase would have been *very* fragile if you relied on this anyway, it's a very unsafe thing to do):

```javascript
Project.belongsTo(User);
Project.belongsTo(User);
```

## The way associations are determined when using `include` has changed

When specifying an `include` using a model+as, Sequelize 6 would search for an association that was defined with that model & as.

So these failed (on purpose):

```javascript
Project.belongsTo(User);

Project.findAll({
  include: { model: User, as: 'User' },
});

// ----

Project.belongsTo(User, { as: 'User' });

Project.findAll({
  include: { model: User },
});
```

these worked:

```javascript
Project.belongsTo(User);

Project.findAll({
  include: { model: User },
});

// ----

Project.belongsTo(User, { as: 'User' });

Project.findAll({
  include: { model: User, as: 'User' },
});
```

---

In Sequelize 7, the way it works is that if only one association has been defined with that model as the target, it will use that one, otherwise it will disambiguate using `as`.

**If you're using sequelize-typescript, that is already how it works as they override our system to implement this**

That means using `as` is exactly the same as using `association:` and I think we should update our documentation to promote the use of `association:` over the use of `model:`. i.e. instead of doing this:

```javascript
Project.findAll({
  include: { model: User, as: 'User' },
});
```

do this:

```javascript
Project.findAll({
  include: 'User',
});
```

## Associations are immutable

Before this PR, associations were a bit fuzzy: Defining an association on both sides of the association would attempt to merge and reconcile their options.

A good idea in theory, but these associations are applied as soon as the method is called. This means reconciliation may need to revert changes that were already done on other objects, like renaming a previously added foreign key on a model.

Determining which association was paired with which was also tricky.

With this PR, it's still possible to define association on both sides, but their options must be perfectly compatible. For instance this is not valid:

```typescript
User.belongsToMany(Countries, { foreignKey: 'user_id' });
Countries.belongsToMany(User);
```

but this is:

```typescript
User.belongsToMany(Countries, { foreignKey: 'user_id' });
Countries.belongsToMany(User, { otherKey: 'user_id' });
```

To make things easier, I've introduced the long requested `"inverse"` option which lets a user define the association on both sides in one call.

Instead of doing this:

```typescript
User.belongsToMany(Countries, { as: 'countries' });
User.belongsToMany(Countries, { as: 'citizen' });
```

users should now do this:

```typescript
User.belongsToMany(Countries, { as: 'countries', inverse: { as: 'citizen' } });
```

Closes https://github.com/sequelize/sequelize/issues/5144

## Includes are now strict

This PR normalizes Includes, which resolves a series of inconsistencies and should make them more stable.

For instance, `Includes` are always validated. If no association is found matching that include, an error will be thrown (including nested includes).

## ~PENDING QUESTION: Found an issue with `BelongsToMany`.~

Edit: resolved, see comments

Given a `BelongsToMany` association from `User` to `Project` through `User_Project`,
and using `groupedLimit` on the `BelongsToMany` association,

- [This unit test](https://github.com/sequelize/sequelize/blob/main/test/unit/sql/select.test.js#L110-L134) expects the through table to be aliased to `User_Project`.
- [This integration test](https://github.com/sequelize/sequelize/blob/main/test/integration/model/findAll/groupedLimit.test.js#L76-L93) expects the through table to be aliased to `Project`.

This needs to be unified, although I'm not sure which choice is best. Right now I've picked the second behavior.

## Other breaking changes

- The options 'foreignKeyConstraint' and 'constraints' have been merged in 'foreignKeyConstraints' as they did the same thing.
- The 'timestamps' option in `BelongsToManyOptions` has been moved to `ThroughOptions.timestamps`
- The 'uniqueKey' option in `BelongsToManyOptions` has been merged with `ThroughOptions.unique`
- HasMany / HasOne don't use 'as' as the foreign key anymore (it wouldn't make sense. You ended up with the class `Project.hasMany(User)` meant the model `User` received a foreign key called `UserId`.
- The `foreignKey` and `otherKey` options are unified with `ColumnOptions`:
   - The 'keyType' option is now 'foreignKey.type'
   - The 'onDelete' option is not 'foreignKey.onDelete'
   - The 'onUpdate' option is not 'foreignKey.onUpdate'
- The default value for `onDelete` for non-null FKs is unified. It's now always `CASCADE`. Previously, it was `NO ACTION` if defined first through `belongsTo`, and `CASCADE` if defined first through `hasOne` or `hasMany`, leading to unexpected behaviors depending on which method was called first.

Closes https://github.com/sequelize/sequelize/issues/13429
Closes https://github.com/sequelize/sequelize/issues/12782
Closes https://github.com/sequelize/sequelize/issues/11837

# What's left to be done

- [ ] Merge https://github.com/sequelize/sequelize/pull/14333
- [x] Merge https://github.com/sequelize/sequelize/pull/14359
- [x] Merge https://github.com/sequelize/sequelize/pull/14360
- [x] Add more typing tests
- [x] Check 'inverse' supports all options
- [x] add configuration for `fromThroughToSource` and `fromSourceToThrough`
- [x] Add support for `inverse` in `HasMany`, `HasOne`, `BelongsToMany`
- [x] Add support for `inverse` in `BelongsTo`
- [x] Determine the best option for the issue 'Found an issue with `BelongsToMany`.'

# What's next (in follow-up PRs)?

- [ ] Consider renaming `otherKey` to `inverse.foreignKey`
- [ ] Forbid defining an association if the primary key is composite. As this is not supported and breaks silently. (we can add support for this but users should be aware that it will not work).
- [ ] Document the breaking changes in the upgrade guide
- [ ] Update the documentation to reflect new recommended behaviors (`includes`, super many-to-many, `inverse`)
- [ ] Document the best practices to avoid errors with associations: Use 'inverse' instead of defining the association twice, beware of creating the same association twice without naming the inverse (`this.User.belongsToMany(this.Task) +  this.User.belongsToMany(this.Task, {as:'association2'})`)
- [ ] Improve [`ORDER BY`](https://github.com/sequelize/sequelize/tree/ephys/order-typing) thanks to the normalization of associations


test/unit/esm-named-exports.test.js
```diff
@@ -124,11 +124,12 @@ Either add these exports to "index.mjs" (and "types/index.d.ts"), or mark them a
 
     for (const key of esmKeys) {
       expect(sequelizeEsm[key]).not.to.eq(undefined, `esm is exporting undefined under key ${JSON.stringify(key)}`);
-      expect(cjsKeys).to.include(key, `esm entry point is declaring export ${JSON.stringify(key)} that is missing from CJS`);
 
-      // exported values need to be the same instances
-      //  if we want to avoid major bugs:
-      //  https://github.com/sequelize/sequelize/pull/13689#issuecomment-987412233
```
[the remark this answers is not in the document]
ephys: I removed it by accident 😅

ephys:
New breaking change to continue with working on making associations safe by default:

Calling `Project.belongsToMany(User)` and `User.belongsToMany(Project)` with different options in each call is not allowed anymore. It lead to unpredictable behavior.

This was done because `belongsToMany` is the only association that is automatically defined on both the source & target models.

Doing so results in the following error being thrown:

```
AssertionError [ERR_ASSERTION]: Project.belongsToMany(User) was called with different options than User.belongsToMany(Project). This is not allowed.
Note that 'belongsToMany' associations are automatically created on the target model as well, so you only need to call this method on one side.
```

ephys:
Crazy that this is allowed, and even used in tests:

image

This defined two different associations with the same name. `User.associations.tasks` was equal to the most recent call to `belongsToMany`

After this PR, this just throws an error explaining that the association name is already in use for that model, and to use `as` to rename it. The updated code looks like this:

image

WikiRik:
About `BelongsToMany`; is there any reason why you chose for the second? In general I would say that `User_Project` is clearer naming so that would have my preference but I haven't looked into it

ephys:
Honestly the more I look at that, the less I know what the right answer is

The query generator transforms `groupedLimit` on belongsToMany associations into a `groupedLimit` on a `hasOne` association between the model and the belongsToMany's through table: https://github.com/sequelize/sequelize/blob/ephys/ts-associations/src/dialects/abstract/query-generator.js#L1282

We could:
- lie and pretend it's the belongsToMany association (name the join `project`)
- continue with the hasOne association (name it `userProjects`)
- use the same 'hack' other parts of the code use that explicitely rename that hasOne association to the name of the through model.

BelongsToMany are weird but mostly it's for good reasons. I think we should go with the third option after all and name it `User_Project` (the name of the through model)

ephys:
My latest commit changes it back to `User_Project`, which makes it aligned with every other place the through model is automatically included

sdepold:
Oh ... my ... god ... 🙈😳👌

ephys:
If you're looking for pointers to review this huge PR, here is what I would recommend:

- Go through the list of changes described in the original post to see if you agree with them. It's not because something is implemented that it can't be reversed
- Go through the tests first: see if the changes still make sense to you
- Then the actual source code 😅

ephys:
Let me know when you're ready to review, I'll resolve the conflicts then

WikiRik:
I took the time a few days back to look at the changes and they look good to me. It will cause some time to upgrade projects so we need to be mindful with the examples we give in the upgrade guide (and in general how we document the new best practices in using associations in the various pages)

Next step for me would indeed be reviewing the tests but that will take some time. Would be nice if you could resolve the conflicts here once #14333 has been merged

ephys:
> so we need to be mindful with the examples we give in the upgrade guide 

yup! Ideally I'd like to do that & rewrite the documentation on association before we merge, but I don't know when I'll be able to

> Would be nice if you could resolve the conflicts here once https://github.com/sequelize/sequelize/pull/14333 has been merged

will do

ephys:
Conflicts resolved

src/associations/has-one.ts
```diff
@@ -0,0 +1,462 @@
...
+      && !source.getAttributes()[options.sourceKey]
+    ) {
+      throw new Error(`Unknown attribute "${options.sourceKey}" passed as sourceKey, define this attribute on model "${source.name}" first`);
+    }
+
+    if ('keyType' in options) {
+      throw new TypeError('Option "keyType" has been removed from the BelongsTo\'s options. Set "foreignKey.type" instead.');
+    }
+
+    // TODO: throw is source model has a composite primary key.
```
WikiRik, line 103: When should we work on this, before the release of v7 or after?

src/associations/has-one.ts
```diff
@@ -0,0 +1,462 @@
...
+ *  declare createRole: HasOneCreateAssociationMixin<Role>;
+ * }
+ *
+ * User.hasOne(Role);
+ * ```
+ *
+ * @see Model.hasOne
+ */
+export type HasOneCreateAssociationMixin<T extends Model> = (
+  // TODO: omit the foreign key
```
WikiRik, line 459: Is this for before the release of v7 or after?

src/associations/has-many.ts
```diff
@@ -0,0 +1,834 @@
...
+    if (!Array.isArray(instances)) {
+      isManyMode = false;
+      instances = [instances];
+    }
+
+    const findOptions: FindOptions = { ...options };
+
+    const where = Object.create(null);
+
+    // FIXME: scopes should be combined using AND instance of overwriting.
```
WikiRik, line 199:
```suggestion
    // TODO: scopes should be combined using AND instance of overwriting.
```
Is this for this PR, before the release of v7 or after?

src/associations/has-many.ts
```diff
@@ -0,0 +1,834 @@
...
+        };
+      }),
+    };
+
+    const findOptions: HasManyGetAssociationsMixinOptions<T> = {
+      ...options,
+      scope: false,
+      attributes: [this.target.primaryKeyAttribute],
+      raw: true,
+      // TODO: current WhereOptions typings do not allow having 'WhereOptions' inside another 'WhereOptions'
```
WikiRik, line 325: Will this be addressed in #14020 ?

src/associations/has-many.ts
```diff
@@ -0,0 +1,834 @@
...
+  ): Promise<T> {
+    if (Array.isArray(options)) {
+      options = {
+        fields: options,
+      };
+    }
+
+    if (this.scope) {
+      for (const attribute of Object.keys(this.scope)) {
+        // TODO: fix the typing of {@link AssociationScope}
```
WikiRik, line 505: When do you think we should take care of this? Before or after the release of v7?

src/associations/belongs-to-many.ts
```diff
@@ -0,0 +1,1316 @@
...
+  }
+
+  /**
+   * The name of the Foreign Key attribute, located on the through table, that points to the Target model.
+   *
+   * Not to be confused with @link {BelongsToMany.foreignKey}, which points to the Source model instead.
+   *
+   * @type {string}
+   */
+  // '!' added because this is initialized in constructor through a function call
```
WikiRik, line 131: Where is that '!' added?

src/associations/belongs-to-many.ts
```diff
@@ -0,0 +1,1316 @@
...
+   *
+   * @param instance instance
+   * @param options find options
+   */
+  async get(instance: SourceModel, options?: BelongsToManyGetAssociationsMixinOptions<TargetModel>): Promise<TargetModel[]> {
+    const through = this.through;
+
+    const findOptions: FindOptions<Attributes<TargetModel>> = {
+      ...options,
+      // TODO: current WhereOptions typings do not allow having 'WhereOptions' inside another 'WhereOptions'
```
WikiRik, line 446: Just for completeness;  I made a comment about this in one of the other files as well. That same comment applies here

src/associations/belongs-to-many.ts
```diff
@@ -0,0 +1,1316 @@
...
+      return instance as TargetModel[TargetKey];
+    });
+
+    const associatedObjects: TargetModel[] = await this.get(sourceInstance, {
+      ...options,
+      raw: true,
+      scope: false,
+      attributes: [this.targetKey],
+      joinTableAttributes: [],
+      // TODO: current WhereOptions typings do not allow having 'WhereOptions' inside another 'WhereOptions'
```
WikiRik, line 548: Just for completeness; I made a comment about this in one of the other files as well. That same comment applies here

src/associations/belongs-to-many.ts
```diff
@@ -0,0 +1,1316 @@
...
+      });
+
+      promises.push(this.through.model.bulkCreate(bulk, { validate: true, ...options }));
+    }
+
+    for (const changedTarget of changedTargets) {
+      // @ts-expect-error -- gets the content of the "through" table for this association that is set on the model
+      let throughAttributes = changedTarget[this.through.model.name];
+      const attributes = { ...defaultAttributes, ...throughAttributes };
+      // Quick-fix for subtle bug when using existing objects that might have the through model attached (not as an attribute object)
```
WikiRik, line 737: Do we want to make this a TODO for a future PR if somebody wants to fix this properly?

src/associations/belongs-to-many.ts
```diff
@@ -0,0 +1,1316 @@
...
+    throw new TypeError('The "timestamps" option in belongsToMany has been renamed to through.timestamps');
+  }
+
+  if ('uniqueKey' in options) {
+    throw new TypeError('The "uniqueKey" option in belongsToMany has been renamed to through.unique');
+  }
+
+  const sequelize = target.sequelize!;
+
+  // TODO: normalize otherKey using the same logic as in normalizeBaseAssociationOptions
```
WikiRik, line 880: When should we take care of this?

ephys: Oh, it's done. That's an old TODO. Removed

test/integration/hooks/associations.test.js
```diff
@@ -91,79 +90,78 @@ describe(Support.getTestDialectTeaser('Hooks'), () => {
...
-            await project.setTask(task);
-            await expect(project.destroy()).to.eventually.be.rejectedWith(CustomErrorText);
-            expect(beforeProject).to.be.true;
-            expect(afterProject).to.be.true;
-            expect(beforeTask).to.be.true;
-            expect(afterTask).to.be.false;
-          });
-        });
-      });
+      // this test makes no sense: task was destroyed by the database through CASCADE.
```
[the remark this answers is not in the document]
ephys: I didn't feel comfortable removing a test but this is one last good opportunity to check that the test was broken and that we can delete this

src/model.d.ts
```diff
@@ -2208,17 +2230,28 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
    * @param {string} alias
    * @return {boolean}
    */
+  // TODO: rename to hasAssociation
```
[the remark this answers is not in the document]
ephys: I would suggest doing it during the migration of model.js to TypeScript. Doesn't really matter if we do it before or after the release of v7, we'll probably deprecate `hasAlias` and add `hasAssociation` to avoid another breaking change

src/associations/belongs-to.ts
```diff
@@ -0,0 +1,488 @@
...
+  inverse: Association | undefined;
+
+  constructor(
+    secret: symbol,
+    source: ModelStatic<S>,
+    target: ModelStatic<T>,
+    options: NormalizedBelongsToOptions<SourceKey, TargetKey>,
+    parent?: Association,
+  ) {
+    // TODO: throw is source model has a composite primary key.
```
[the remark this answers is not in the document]
ephys: It's not urgent. Could be for v7, could be for v8. We could also add support for composite foreign keys

src/associations/belongs-to.ts
```diff
@@ -0,0 +1,488 @@
...
+      Target = Target.schema(options.schema, options.schemaDelimiter);
+    }
+
+    let isManyMode = true;
+    if (!Array.isArray(instances)) {
+      isManyMode = false;
+      instances = [instances];
+    }
+
+    // FIXME: the scope is ignored
```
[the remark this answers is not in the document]
ephys: It was already like this in v6. We should investigate whether it's intended or not but either way it's not a priority

src/associations/belongs-to-many.ts
ephys: Stale comment, I've removed it

src/associations/has-many.ts
ephys: Either is fine. It wouldn't be a breaking change if done properly

test/support.ts
```diff
@@ -304,14 +345,24 @@ export function isDeepEqualToOneOf(actual: unknown, expectedOptions: unknown[]):
 export function minifySql(sql: string): string {
   // replace all consecutive whitespaces with a single plain space character
   return sql.replace(/\s+/g, ' ')
-  // remove space before comma
+    // remove space before coma
```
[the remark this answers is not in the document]
ephys: merged

WikiRik (approved):
I still have some questions open, but nothing that's blocking this PR. Good work!

WikiRik (changes requested):
Woops, it seemed that merging main broke some unit tests. Should have looked at that as well, before approving 😅

ephys:
> I mainly have questions to when TODOs will have to be taken care of

Would it be enough if I mentioned whether the change was a breaking change? As much as I'd like to have some of them fixed before v7, I can't guarantee it with the time I can allocate to this project :/

WikiRik (approved):

bminer:
Imagine you have a `Task` belongs to `Project` association. What if a Task had a `similarProjectID` field? How would this association work in Sequelize v7?

bminer:
@WikiRik @ephys - Any ideas on my prior question? If table `tasks` has a `projectID` and a `similarProjectID`, how can you create both associations in Sequelize v7?

bminer:
Please ignore my prior comment. I misunderstood this change, sorry! I'm going to hide it.

PR #14304, opened 2022-03-31
feat(snowflake): add support of 'configure' properties for Snowflake dialect

ReAleR09:
### Pull Request Checklist

- [ ] Have you added new tests to prevent regressions? (change does not influence default behavior)
- [ ] Does `yarn test` or `yarn test-DIALECT` pass with this change (including linting)? (no, but don't think it's required for this change)
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? (no)
- [ ] Did you update the typescript typings accordingly (if applicable)? (no typings required)

### Description Of Change

Adding ability to pass additional 'configure' options for 'snowflake-sdk' driver. This is necessary to pass inside 'logLevel' (self-explained) and 'insecureConnection' (ignore OCSP checks) options.

### DIFF
--- a/src/dialects/snowflake/connection-manager.js
+++ b/src/dialects/snowflake/connection-manager.js
@@ -69,6 +69,9 @@ export class SnowflakeConnectionManager extends ConnectionManager {
     try {
 
       const connection = await new Promise((resolve, reject) => {
+        if (connectionConfig.configure) {
+          this.lib.configure({ ...connectionConfig.configure });
+        }
         this.lib.createConnection(connectionConfig).connect((err, conn) => {
           if (err) {
             console.error(err);


ephys:
The options passed to the Sequelize constructor are instance options. These options modifying the global options of something else would break that expectation, two instantiations of Sequelize with different values in the 'configure' field will impact each-other.

Users should do this instead:

```typescript
const Snowflake = require('snowflake-sdk');
Snowflake.configure(/* your options */);

const sequelize = new Sequelize({ dialect: 'snowflake', /* remaining options */ })
```

ReAleR09:
> The options passed to the Sequelize constructor are instance options. The options passed to `Snowflake.configure` are global options shared by all instances of Snowflake. Making a Sequelize instance configure that global which would break the important expectation that they're self contained instances.
> 
> Users should do this instead:
> 
> ```ts
> const Snowflake = require('snowflake-sdk');
> Snowflake.configure(/* your options */);
> 
> const sequelize = new Sequelize({ dialect: 'snowflake', /* remaining options */ })
> ```

Makes sense, thanks!

Issue #13978, opened 2022-01-20
When SQL errors occur, add 'message 'to DEBUG log or DatabaseError


mprevdelta:
## Issue Creation Checklist

- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Feature Description

When SQL errors occur, make the 'message' property describing the error available.

### Is your feature request related to a problem? Please describe.

When invoking Sequelize methods that create errors, the 'message' attribute of the error often gets buried. This message often describes precisely where the error occurs and is very useful for debugging problems with Models.

I propose this be logged using 'debug', or perhaps added to the created Error. 

### Describe the solution you'd like

Either a 'message' property added to DatabaseError, or for each driver to invoke debug() with the message when an error occurs. For the debug solution, user could use the DEBUG env var and get messages like this:

  sequelize:sql:mssql The INSERT statement conflicted with the FOREIGN KEY constraint "fk_bla_bla_bla". The conflict occurred in database "yourdb", table "yourtable", column 'id'

### Additional context

This issue may be specific to the mssql dialect.

### Is this feature dialect-specific?

- [ ] No. This feature is relevant to Sequelize as a whole. It looks like 
- [ ] Yes. This feature only applies to the following dialect(s): XXX, YYY, ZZZ
- [x] Possibly, depending on the solution used. Some other dialects may handle errors better than the mssql dialect is.

### Would you be willing to implement this feature by submitting a Pull Request?

- [x] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

I'd be willing to submit a simple pull adding the debug() call in mssql query.js. If the solution involved reworking DatabaseError I wouldn't feel comfortable making the change because it'd affect all dialects.

ephys:
We could use the new [`Error#cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error) option, that'd improve the error output for users on node >= 16.9.0

PR #14318, opened 2022-04-04
feat: use native `Error#cause` in Sequelize errors

ephys:
### Pull Request Checklist

### Description Of Change

This PR migrates our errors to use the native `Error#cause` property.

This was previously done by the `parent` property. It is still present but logs a deprecation warning when used.

This should lead to clearer error messages.

This is only available in node 16+. A workaround has been added for node 14.

image

Closes https://github.com/sequelize/sequelize/issues/13978

### DIFF
--- a/src/errors/base-error.ts
+++ b/src/errors/base-error.ts
@@ -13,7 +13,7 @@ export interface CommonErrorProperties {
 //  Remove me in Sequelize 8, where this is added natively by TypeScript (>= 4.6):
 //  This is a breaking change and must be done in a MAJOR release.
 export type ErrorOptions = {
-  cause?: unknown,
+  cause?: Error,
 };
 
 const supportsErrorCause = (() => {
@@ -33,7 +33,7 @@ const supportsErrorCause = (() => {
  * This means that errors can be accessed using `Sequelize.ValidationError`
  */
 abstract class BaseError extends Error {
-  declare cause: unknown;
+  declare cause: Error | undefined;
 
   get parent(): this['cause'] {
     useErrorCause();
--- a/src/errors/base-error.ts
+++ b/src/errors/base-error.ts
@@ -1,14 +1,10 @@
-export interface ErrorOptions {
+import { useErrorCause } from '../utils/deprecations.js';
+
+export interface SequelizeErrorOptions {
   stack?: string;
 }
 
 export interface CommonErrorProperties {
-  /** The database specific error which triggered this one */
-  readonly parent: Error;
-
-  /** The database specific error which triggered this one */
-  readonly original: Error;
-
   /** The SQL that triggered the error */
   readonly sql: string;
 }
@@ -22,8 +18,20 @@ export interface CommonErrorProperties {
  * This means that errors can be accessed using `Sequelize.ValidationError`
  */
 abstract class BaseError extends Error {
-  constructor(message?: string) {
-    super(message);
+  get parent(): this['cause'] {
+    useErrorCause();
+
+    return this.cause;
+  }
+
+  get original(): this['cause'] {
+    useErrorCause();
+
+    return this.cause;
+  }
+
+  constructor(message?: string, options?: ErrorOptions) {
+    super(message, options);
     this.name = 'SequelizeBaseError';
   }
 }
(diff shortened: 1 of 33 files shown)


fzn0x (approved):

src/errors/base-error.ts
```diff
@@ -22,8 +18,20 @@ export interface CommonErrorProperties {
  * This means that errors can be accessed using `Sequelize.ValidationError`
  */
 abstract class BaseError extends Error {
-  constructor(message?: string) {
-    super(message);
+  get parent(): this['cause'] {
```
fzn0x, line 38: Lookup types seems good

WikiRik:
Just to check; this uses https://github.com/tc39/proposal-error-cause right? That would mean that we should put Node 16.9.0 as our minimal supported version when we remove the workaround. We don't need to do anything about that now, but something to keep in mind. I have written it down in my personal notes. I want to do something about it in our CI later as well but will write more about that later

src/errors/base-error.ts
```diff
@@ -22,9 +33,33 @@ export interface CommonErrorProperties {
...
+  constructor(message?: string, options?: ErrorOptions) {
+    // TODO [>=2023-04-30]: remove this ts-ignore (Sequelize 8)
+    // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error -- Supported in TS 4.6, not before
+    // @ts-ignore
+    super(message, options);
     this.name = 'SequelizeBaseError';
+
+    if (!supportsErrorCause) {
+      // TODO [>=2023-04-30]:
+      //  Once all supported node versions have support for Error.cause (added in node 16), delete this line:
```
WikiRik, line 59:
```suggestion
      //  Once all supported node versions have support for Error.cause (added in Node 16.9.0), delete this line:
```
As per my earlier comment

src/errors/database-error.ts
```diff
@@ -1,14 +1,17 @@
-import type { CommonErrorProperties, ErrorOptions } from './base-error';
+import type { CommonErrorProperties, SequelizeErrorOptions } from './base-error';
```
WikiRik, line 1: Shouldn't this be ErrorOptions? Since that has `cause`, SequelizeErrorOptions has `stack`. Or am I reading it wrong?

src/utils/deprecations.ts
```diff
@@ -8,3 +8,4 @@ export const noStringOperators = deprecate(noop, 'String based operators are dep
 export const noBoolOperatorAliases = deprecate(noop, 'A boolean value was passed to options.operatorsAliases. This is a no-op with v5 and should be removed.', 'SEQUELIZE0004');
 export const noDoubleNestedGroup = deprecate(noop, 'Passing a double nested nested array to `group` is unsupported and will be removed in v6.', 'SEQUELIZE0005');
 export const unsupportedEngine = deprecate(noop, 'This database engine version is not supported, please update your database server. More information https://github.com/sequelize/sequelize/blob/main/ENGINE.md', 'SEQUELIZE0006');
+export const useErrorCause = deprecate(noop, 'The "parent" and "original" properties in Sequelize errors have been replaced with the native "cause" property. Use that one instead.', 'SEQUELIZE0007');
```
WikiRik, line 11: I see that 'parent' still works, but 'original' is removed. Is that on purpose?

WikiRik (commented):
Overall looks good, just have a few remarks

src/errors/database-error.ts
ephys: `DatabaseError` doesn't accept a `cause` option because its first parameter is the cause:

image

src/utils/deprecations.ts
ephys: I'm not quite sure what the state of `original` is.
Prior to this PR, it was already not possible to pass an option called `original`, its value was always equal to `parent`

The deprecation warning mentions it because accessing the `original` property will log the warning too: https://github.com/sequelize/sequelize/pull/14318/files#diff-e86925093ea986c091226dc2f7e5af9b4c604ee4fa5c65f0795b7a4876eb16fdR44

ephys:
> That would mean that we should put Node 16.9.0 as our minimal supported version when we remove the workaround.

Correct :) Ideally Sequelize 8 will support Node >= 16.9.0 (or whichever minor version we decide to support on release)

WikiRik (approved):

PR #14320, opened 2022-04-04
chore(generic): add more falsy values

fzn0x:
## Description Of Change

falsy/falsey means internal toBoolean returns false for the values

0.0 | -0.0 : any zeroish number return false
0x0 | -0x0 : represent hex number of 0
NaN : Not a Number
---
**new falsy values on ES2020**
0x0n | -0x0n : the bigint of zero and negative zero, alongside 0n and -0n
---


fzn0x:
Let me fix the lint 👀

fzn0x:
Just deleted the zero fraction.

there is no differentiation of integers and floats. All numbers are inherently floating.

WikiRik:
Any reason not to follow https://developer.mozilla.org/en-US/docs/Glossary/Falsy ? Apart from `document.all`

fzn0x:
I prefer using branch for cleaner commit details, including squash.

fzn0x:
I prefer using rebase for cleaner commit details, including squash.

fzn0x:
> Any reason not to follow https://developer.mozilla.org/en-US/docs/Glossary/Falsy ? Apart from `document.all`

You can take a look at the description, it was giving some alternative values for falsy/falsey too.

I wish we can add more falsy values to prevent non-truthy values being passed.

fzn0x:
> Any reason not to follow https://developer.mozilla.org/en-US/docs/Glossary/Falsy ? Apart from `document.all`

You can take a look at the description, it was giving some alternative values for falsy/falsey too.

I wish we can add more falsy values to prevent non-truthy values being not passed by typescript.

fzn0x:
Updated the PR description for the changes goals.

src/generic/falsy.ts
```diff
@@ -1 +1 @@
-export type Falsy = false | 0 | -0 | 0n | '' | null | undefined;
+export type Falsy = false | 0 | -0 | 0n | '' | null | undefined | 0x0n | 0x0 | -0x0;
```
ephys, line 1: This addition isn't necessary. The `0x` notation is just a different notation, the value is the same, as you can see in this playground https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAQwBQEoBciAMAPHRAbwFgAoRRAJwFMoQqkcBucgX3PNElgUQCMM2QqQrU6DJgVZk2QA

ephys (changes requested):

WikiRik:
Apparently there was a reason to not add NaN here; https://github.com/sequelize/sequelize/pull/13763#discussion_r767351644

fzn0x:
> Apparently there was a reason to not add NaN here; [#13763 (comment)](https://github.com/sequelize/sequelize/pull/13763#discussion_r767351644)

It was actually discussed in here too https://github.com/Microsoft/TypeScript/issues/28682 but it was still feasible to add it yet, there is some mathematical operation are returning `NaN`

ephys:
Since there are no `NaN` type in TypeScript yet, I think we should not change anything. Trying to get the type of NaN will at best be number, which will match any number

fzn0x:
> Reference in

Closed this PR, this PR should be future notes also

PR #14323, opened 2022-04-05
docs: restore Model api reference & make fail on error

ephys:
### Pull Request Checklist

### Description Of Change

This PR modifies model.js to use syntax compatible with esdoc. This brings the Model api reference page back.

It also ensures the `docs` script fails if it's unable to parse a file again

Related to https://github.com/sequelize/website/issues/43

### DIFF
--- a/docs.sh
+++ b/docs.sh
@@ -1,0 +1,23 @@
+set -e
+
+rimraf esdoc
+
+exec 5>&1
+OUT=$(esdoc -c docs/esdoc-config.js|tee /dev/fd/5)
+
+cp docs/favicon.ico esdoc/favicon.ico
+cp docs/ROUTER.txt esdoc/ROUTER
+
+node docs/run-docs-transforms.js
+node docs/redirects/create-redirects.js
+
+rimraf esdoc/file esdoc/source.html
+
+set +e
+GREP_RESULT=$(echo "$OUT" | grep -c 'could not parse the following code')
+set -e
+
+if [ "$GREP_RESULT" -ge 1 ]; then
+  echo "esdoc generation encountered an error. See above logging for details."
+  exit 1
+fi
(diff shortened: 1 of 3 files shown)


src/model.js
```diff
@@ -1034,8 +1034,8 @@ class Model {
       }
     });
 
-    if (!_.includes(['and', 'overwrite'], this.options?.whereMergeStrategy)) {
-      throw new Error(`Invalid value ${this.options?.whereMergeStrategy} for whereMergeStrategy. Allowed values are 'and' and 'overwrite'.`);
+    if (!_.includes(['and', 'overwrite'], this.options && this.options.whereMergeStrategy)) {
```
WikiRik, line 1037: This has tests somewhere right? Because this doesn't seem right to me, && just return true and false right? So it would never include 'and' or 'overwrite'

docs.sh
```diff
@@ -0,0 +1,23 @@
...
+node docs/run-docs-transforms.js
+node docs/redirects/create-redirects.js
+
+rimraf esdoc/file esdoc/source.html
+
+set +e
+GREP_RESULT=$(echo "$OUT" | grep -c 'could not parse the following code')
+set -e
+
+if [ "$GREP_RESULT" -ge 1 ]; then
```
WikiRik, line 20: Did you test this on GitHub Actions already? Just to check if it fails the same way as locally

src/model.js
ephys: && returns the first operand if it is false, and the second otherwise value, `1 && 2 === 2`
A feature very specific to JavaScript

docs.sh
ephys: I haven't but I'm running the same os and these are base tools, and it succeeded in the ci, it should fail properly if there is a syntax error

src/model.js
WikiRik: Aah, today I learned

PR #14337, opened 2022-04-07
feat: add Model.{getInitialModel,withoutScope}, rename schema+(un)?scope

ephys:
### Pull Request Checklist

### Description Of Change

Another building block for https://github.com/sequelize/sequelize/pull/14280

Context:
`Model.scope`, `Model.unscope` and `Model.schema` return new classes.
This is a problem for the above PR because that PR needs to be able to check whether two models classes are the same.

This PR does a few things:
- Deduplicate models generated by these 3 methods. Which means that `Model.schema('public') === Model.schema('public')`
- Adds `Model.withInitialScope`: `unscope()` is a misnomer, it doesn't restore the model to its state before the call to `scope()`. Instead it returns a model with no scope at all (including the default scope). `withInitialScope()` does return a model with the default scope.
- Adds `Model.getInitialModel` which returns the model, like it was before any call to `scope`, `unscope` or `schema`
- Renames `unscope` to `withoutScope`, `scope` to `withScope` and `schema` to `withSchema`.
  This change was done to signal that these methods will not mutate the Model and, instead, return a new Model. These names follow conventions for similar methods that we can find in other libraries.
  The old names still exist but will log a deprecation notice. There are no breaking changes in this PR.


WikiRik (approved):
Looks good! I'm assuming the documentation of the renames will follow later?

ephys:
In our upgrade guide? Yep :)

PR #14339, opened 2022-04-07
feat!: drop support for node 12

ephys:
Note 12 is going EOL in 23 days, Sequelize 7 won't be released by then. Let's drop support for Node 12 in the upcoming major


ephys:
This will require removing the required checks for node 12, and adding the ones for node 14

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           cache: yarn
       - name: Install dependencies and build sequelize
         run: yarn install --frozen-lockfile --ignore-engines
-        if: matrix.node-version == 12
+        if: matrix.node-version == 14
```
WikiRik, line 30: Is this still needed?

build.js
```diff
@@ -79,7 +79,7 @@ async function main() {
       // Adds source mapping
       sourcemap: true,
       // The compiled code should be usable in node v12.22
```
WikiRik, line 81:
```suggestion
      // The compiled code should be usable in node v14.17
```

WikiRik (changes requested):

.github/workflows/ci.yml
ephys: Let's find out!

ephys: Looks like it isn't needed anymore

.github/workflows/ci.yml
```diff
@@ -219,9 +215,9 @@ jobs:
           name: install-build-artifact-node-${{ matrix.node-version }}
       - name: Extract artifact
         run: tar -xf install-build-node-${{ matrix.node-version }}.tar
-      - name: Install pg-native (Node 12)
+      - name: Install pg-native (Node 14)
         run: yarn add pg-native --ignore-engines
-        if: matrix.native && matrix.node-version == 12
+        if: matrix.native && matrix.node-version == 14
       # This step uses npm instead of yarn to minimize the time needed. See #14171
```
WikiRik, line 221: Can you check if we need this still?

WikiRik (commented):
One more comment, but good to see that we can remove some workarounds now

.github/workflows/ci.yml
ephys: Looks like it is

ephys:
Ready :) We just need to update the required checks for this branch

WikiRik (approved):

Issue #14343, opened 2022-04-08
Add support for YugabyteDB


priyanshi-yb:
## Issue Description

Add support for [YugabyteDB](https://www.yugabyte.com/), a cloud-native Distributed SQL database.

## Issue Creation Checklist

- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

### Is your feature request related to a problem? Please describe.

The `findOrCreate()` API with YugabyteDB fails while parsing undefined error fields for YugabyteDB.

Refer to this [issue](https://github.com/yugabyte/yugabyte-db/issues/9294) for more information.

### Describe the solution you'd like

Implement a new dialect for the YugabyteDB.

### Why should this be in Sequelize

YugabyteDB is a high-performance, cloud-native distributed SQL database that is [growing rapidly](https://blog.yugabyte.com/yugabyte-raises-188-million-series-c-to-make-distributed-sql-ubiquitous/) It'd help the increasing user community of YugabyteDB who intend to use Sequelize.

### Describe alternatives/workarounds you've considered

The inbuilt dialect of PostgresQL works fine with YugabyteDB but due to the below mentioned reasons we need to have a separate dialect:

- Issue using `findOrCreate()` API with YugabyteDB.
- It would make it easier to add YugabyteDB-specific enhancements in future.

We currently have our own package[ sequelize-yugabytedb](https://github.com/yugabyte/sequelize-yugabytedb) .
### Additional context

We had a [conversation](https://github.com/sequelize/sequelize/pull/13431) with @sdepold via emails and community slack for adding the support for YugabyteDB.

### Is this feature dialect-specific?

- [x] No. This feature is relevant to Sequelize as a whole.
- [ ] Yes. This feature only applies to the following dialect(s): XXX, YYY, ZZZ

### Would you be willing to implement this feature by submitting a Pull Request?

- [x] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

PR #14345, opened 2022-04-08
feat(yugabyte): add support for yugabytedb dialect

priyanshi-yb:
### Pull Request Checklist

### Description Of Change

Added a dialect for YugabyteDB.

### How to run the integrations test

Start the docker container of YugabyteDB using `yarn start-yugabytedb` and then run the integrations test.

### Link with existing issue

#14343


test/integration/include/schema.test.js
```diff
@@ -893,7 +893,8 @@ describe(Support.getTestDialectTeaser('Includes with schemas'), () => {
       expect(products[0].Tags.length).to.equal(1);
     });
 
-    it('should be possible to extend the on clause with a where option on nested includes', async function () {
+    // Assertion Error for yugabytedb due to the expection of rows to be selected in the same way it is inserted in the database can't be possible
```
ephys, line 896: can these tests be fixed instead? What is the issue?

test/integration/model.test.js
```diff
@@ -2587,6 +2592,12 @@ describe(Support.getTestDialectTeaser('Model'), () => {
             break;
           }
 
+          case 'yugabytedb': {
```
ephys, line 2595: Add this `case 'yugabytedb':` before `case 'postgres':` to avoid duplicating their contents

test/integration/model/findAll.test.js
```diff
@@ -501,7 +501,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         expect(users[1].intVal).to.equal(10);
       });
 
-      if (['postgres', 'sqlite'].includes(dialect)) {
+      if (['postgres', 'sqlite'].includes(dialect)) { // Yugabyte not support CITEXT yet
```
ephys, line 504: Let's add `AbstractDialect.supports.CITEXT` instead to determine whether the type is supported

test/integration/model/findAll.test.js
```diff
@@ -1383,7 +1383,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         expect(users[0].id).to.be.above(users[2].id);
       });
 
-      it('handles offset and limit', async function () {
+      (dialect !== 'yugabytedb' ? it : it.skip)('handles offset and limit', async function () { // Assertion Error in yugabytedb due to order.
```
ephys, line 1386: add an order by to the test instead of disabling

test/integration/model/findAll.test.js
```diff
@@ -1160,7 +1160,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           }));
         });
 
-        it('sorts by 1st degree association', async function () {
+        (dialect !== 'yugabytedb' ? it : it.skip)('sorts by 1st degree association', async function () {  // Operation Expired or aborted by a conflict in yugabytedb
```
ephys, line 1163: This is a major test to disable. What is the issue? Is the test broken? I would assume yugabyte supports joins

test/integration/model/geography.test.js
```diff
@@ -8,8 +8,13 @@ const DataTypes = require('@sequelize/core/lib/data-types');
 
 const current = Support.sequelize;
 
-describe(Support.getTestDialectTeaser('Model'), () => {
+(current.dialect.name !== 'yugabytedb' ? describe : describe.skip)(Support.getTestDialectTeaser('Model'), () => {
   if (current.dialect.supports.GEOGRAPHY) {
+    /**
+     * Yugabyte has some issues with postGIS set up for now and due to which tests will fail, so need to skip them for now.
```
ephys, line 14: does that mean yugabyte does not support geography? In which case it should be disabled in the dialect's flags

test/integration/pool.test.js
```diff
@@ -16,7 +16,9 @@ function assertSameConnection(newConnection, oldConnection) {
     case 'postgres':
       expect(oldConnection.processID).to.be.equal(newConnection.processID).and.to.be.ok;
       break;
-
+    case 'yugabytedb':
```
ephys, line 19: you can deduplicate code by doing it the same way as mariadb&mysql instead:

image

test/integration/pool.test.js
```diff
@@ -41,7 +43,9 @@ function assertNewConnection(newConnection, oldConnection) {
     case 'postgres':
       expect(oldConnection.processID).to.not.be.equal(newConnection.processID);
       break;
-
+    case 'yugabytedb':
```
ephys, line 46: same comment about deduplicating code

test/integration/pool.test.js
```diff
@@ -133,6 +137,12 @@ describe(Support.getTestDialectTeaser('Pooling'), () => {
             break;
           }
 
+          case 'yugabytedb': {
+            connection.end();
```
ephys, line 141: same comment about deduplicating code

test/integration/query-interface.test.js
```diff
@@ -604,8 +604,9 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
     }
 
     describe('primary key', () => {
-      it('should add, read & remove primary key constraint', async function () {
-        await this.queryInterface.removeColumn('users', 'id');
+      (dialect !== 'yugabytedb' ? it : it.skip)('should add, read & remove primary key constraint', async function () {
+
+        await this.queryInterface.removeColumn('users', 'id'); // Removing a PRIMARY KEY column is not yet supported in Yugabyte
```
ephys, line 609: let's split the 'remove' part in another test then, 'read' & 'add' should still be tested (unless they're also unsupported)

test/integration/query-interface/createTable.test.js
```diff
@@ -123,7 +124,7 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
         });
 
         const table = await this.queryInterface.describeTable('SomeTable');
-        if (dialect.includes('postgres')) {
+        if (dialect.includes('postgres') || dialect.includes('yugabytedb')) {
```
ephys, line 127: the reason for `includes` is that it could be `postgres` or `postgres-native`, you should still use `===`

```suggestion
        if (dialect.includes('postgres') || dialect === 'yugabytedb') {
```

test/integration/query-interface/describeTable.test.js
```diff
@@ -78,6 +78,9 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
         case 'postgres':
           assertVal = 'CHARACTER VARYING(255)';
           break;
+        case 'yugabytedb':
+          assertVal = 'CHARACTER VARYING(255)';
```
ephys, line 82: same comment about deduplicating switch cases

test/integration/query-interface/describeTable.test.js
```diff
@@ -128,14 +132,14 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
           expect(isAdmin.defaultValue).to.be.null;
       }
 
-      if (dialect.startsWith('postgres')) {
+      if (dialect.startsWith('postgres') || dialect.startsWith('yugabytedb')) {
```
ephys, line 135: same comment about using `===` because you don't have to take `postgres-native` into account

test/integration/sequelize/deferrable.test.js
```diff
@@ -6,7 +6,7 @@ const expect = chai.expect;
 const Support = require('../support');
 const Sequelize = require('@sequelize/core');
 
-if (!Support.sequelize.dialect.supports.deferrableConstraints) {
+if (!Support.sequelize.dialect.supports.deferrableConstraints || Support.getTestDialect() === 'yugabytedb') { // Doesn't support Deferrable Constraints properly, refer- https://github.com/yugabyte/yugabyte-db/issues/9288
```
ephys, line 9: Then you should disable it in your dialect's flag instead

test/integration/sequelize/query.test.js
```diff
@@ -95,7 +95,7 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
     });
 
     describe('retry',  () => {
-      it('properly bind parameters on extra retries', async function () {
+      (dialect !== 'yugabytedb' ? it : it.skip)('properly bind parameters on extra retries', async function () { // Assertion Error in yugabytedb
```
ephys, line 98: what is the error? Tests should only be disabled with a good explanation of why they were disabled.
If the test is at fault, it can also be fixed

test/unit/sql/change-column.test.js
```diff
@@ -33,23 +33,24 @@ if (current.dialect.name !== 'sqlite') {
...
-        }).then(sql => {
-          expectsql(sql, {
-            ibmi: 'ALTER TABLE "users" ALTER COLUMN "level_id" SET DATA TYPE FLOAT NOT NULL',
-            mssql: 'ALTER TABLE [users] ALTER COLUMN [level_id] FLOAT NOT NULL;',
-            db2: 'ALTER TABLE "users" ALTER COLUMN "level_id" SET DATA TYPE FLOAT ALTER COLUMN "level_id" SET NOT NULL;',
-            mariadb: 'ALTER TABLE `users` CHANGE `level_id` `level_id` FLOAT NOT NULL;',
-            mysql: 'ALTER TABLE `users` CHANGE `level_id` `level_id` FLOAT NOT NULL;',
-            postgres: 'ALTER TABLE "users" ALTER COLUMN "level_id" SET NOT NULL;ALTER TABLE "users" ALTER COLUMN "level_id" DROP DEFAULT;ALTER TABLE "users" ALTER COLUMN "level_id" TYPE FLOAT;',
-            snowflake: 'ALTER TABLE "users" ALTER COLUMN "level_id" SET NOT NULL;ALTER TABLE "users" ALTER COLUMN "level_id" DROP DEFAULT;ALTER TABLE "users" ALTER COLUMN "level_id" TYPE FLOAT;',
+      if (current.dialect.name !== 'yugabytedb'){ // Alter Column type not supported yet in yugabytedb
```
ephys, line 36: (code style) Prefer adding your comments before the if statement

```suggestion
      // Alter Column type not supported yet in yugabytedb
      if (current.dialect.name !== 'yugabytedb'){
```

test/unit/sql/index.test.js
```diff
@@ -99,7 +100,7 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
       });
     });
 
-    if (current.dialect.supports.index.using === 2) {
+    if (current.dialect.name !== 'yugabytedb' && current.dialect.supports.index.using === 2) { // yugabytedb uses ybgin in place of gin
```
ephys, line 103: You can resolve that by changing the line

```
using: 'gin',
```

with 

```js
using: current.dialect.name === 'yugabytedb' ? 'ybgin' : 'gin',
```

instead of disabling it, so it's still tested :)

src/dialects/yugabytedb/index.js
```diff
@@ -0,0 +1,153 @@
...
+    this.queryGenerator = new QueryGenerator({
+      _dialect: this,
+      sequelize,
+    });
+    this.queryInterface = new YugabyteQueryInterface(
+      sequelize,
+      this.queryGenerator,
+    );
+
+    Model.findOrCreate = async function findOrCreate(options) {
```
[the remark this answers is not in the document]
priyanshi-yb: The reason behind overwriting `findOrCreate()`  API is that this API is failing with YugabyteDB as it is parsing error Fields which cannot be done with YugabyteDB as it is undefined for us. So, we have to handle that case before parsing error fields. you can more on this issue [here](https://github.com/yugabyte/yugabyte-db/issues/9294).

src/sequelize.d.ts
```diff
@@ -168,7 +168,7 @@ export interface Config {
   };
 }
 
-export type Dialect = 'mysql' | 'postgres' | 'sqlite' | 'mariadb' | 'mssql' | 'db2' | 'snowflake';
+export type Dialect = 'mysql' | 'postgres' | 'sqlite' | 'mariadb' | 'mssql' | 'db2' | 'snowflake' | 'yugabytedb';
```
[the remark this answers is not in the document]
priyanshi-yb: We want to have the dialect name as  `yugabytedb` .

dev/yugabytedb/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+services:
+  yugabyte:
+    container_name: sequelize-yugabyte
+    image: yugabytedb/yugabyte:latest
```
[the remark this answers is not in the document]
priyanshi-yb: Yes, we have a release page on our docs for mentioning the versions, please find it [here](https://docs.yugabyte.com/preview/releases/).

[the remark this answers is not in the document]
ephys: Thanks :)
I think we should test against both 2.6 and 2.12 then

test/integration/include/schema.test.js
priyanshi-yb: Issue is the same which is order not included while using findAll, will fix such tests as previously we thought we can't edit the test cases so skipped these for yugabytedb.

test/integration/query-interface.test.js
priyanshi-yb: So, can we add the new splitted test there itself which will run for yugabytedb only or should we write it separately for dialect specific?

test/integration/dialects/yugabytedb/associations.test.js
```diff
@@ -0,0 +1,121 @@
...
+const Support = require('../../support');
+
+const dialect = Support.getTestDialect();
+const DataTypes = require('@sequelize/core/lib/data-types');
+
+if (dialect.startsWith('yugabytedb')) {
+  describe('[YUGABYTEDB Specific] associations', () => {
+    describe('many-to-many', () => {
+      describe('where tables have the same prefix', () => {
+        it('should create a table wp_table1wp_table2s', function () {
```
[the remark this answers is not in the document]
ephys: > shall we make the postgres's test run for yugabytedb as well there in postgres itself?

I think so :)
It's what we should have done for MariaDB / MySQL to avoid duplicating tests. We can separate tests once they become specific to YugabyteDB

test/integration/query-interface.test.js
ephys: I think what would be best would be to split this test in 2:
- one for "is able to add & read primary key constraint"
- one for "is able to remove primary key constraint"

And only disable the second one if the dialect is `yugabytedb`

test/integration/sequelize/deferrable.test.js
priyanshi-yb: Actually we support deferrable constraints, but there are cases mentioned [here](https://github.com/yugabyte/yugabyte-db/issues/9288) due to which we want to skip some test here, will skip those only instead of whole `deferrable.test.js`.

test/integration/query-interface.test.js
priyanshi-yb: Got that, but there is one issue like in the starting of the test we are removing the `PRIMARY KEY` column which cannot be possible in YugabyteDB so, need to modify the test for that as well. 
So, shall we need to modify the test in general or just add another modified test for `yugabytedb`?

src/dialects/yugabytedb/index.js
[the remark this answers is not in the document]
priyanshi-yb: `errFieldKeys`  is storing the keys which `error.fields` is returning like -
for postgres: 
  `error.fields: { username: 'gottlieb' }`
for yugabytedb: 
 ` error.fields: undefined`

Hence, this is a kind of handling these `undefined` fields For YugabyteDB only.

test/integration/sequelize/deferrable.test.js
ephys: I see, thanks :)

src/dialects/yugabytedb/index.js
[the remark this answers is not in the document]
priyanshi-yb: @ephys  shall I move ahead for making this change in `model.js`?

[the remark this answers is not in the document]
ephys: Go ahead :)

I do understand what `errFieldKeys` is but what I need to understand is what it's used for and will setting it to an empty array risk returning bad data to the user

I know it's used to throw an error, setting it to an empty array may silence an important problem the dev should be aware of

package.json
```diff
@@ -268,6 +278,7 @@
     "test-integration-db2": "cross-env DIALECT=db2 yarn test-integration",
     "test-integration-ibmi": "cross-env DIALECT=ibmi yarn test-integration",
     "test-integration-snowflake": "cross-env DIALECT=snowflake yarn test-integration",
+    "test-integration-yugabytedb": "export DIALECT=yugabytedb SEQ_TEST_CLEANUP_TIMEOUT=100000  && yarn mocha \"test/integration/**/*.test.[tj]s\" --timeout 600000",
```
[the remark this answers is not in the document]
priyanshi-yb: Yes, few tests take a bit more time to get executed for YugabyteDB so had to increase the timeout.

test/integration/model/findAll.test.js
priyanshi-yb: @ephys I have addressed all the other comments and only these ones which are related to adding the flag in `Abstract Dialect` are left, I wanted to know whether should I go with that as it will require to add value for that particular support in each dialect as well?

priyanshi-yb:
@ephys  Thanks for initial comments!
we have added the test for `yugabytedb` dialect in CI and wanted to know whether the green pipeline is required for this PR to be merged as I can see one other database is also failing?

test/integration/model/findAll.test.js
WikiRik: We'll make a PR to refactor these in the next few days so you don't have to do that in this PR. We will link the PR here after it is opened

priyanshi-yb:
Hi @ephys @WikiRik , Sorry for the late reply.
Due to the distributed nature of YugabyteDB, we expect some latency in executing DDLs as compared to PostgreSQL. 

We have analysed the test runs and couldn’t find a good way to cut down on the time for executing the tests in the CI. One way one could do that is by skipping some tests, which is obviously not desired.

If you have any other suggestions (from the test framework side), please do let us know!
Thanks!

ephys:
If the issue comes from the very frequent dropping and creation of tables before and after each test, we could look into optimizing that.

This would be a huge task, we would have to go through each integration test suite and determine whether it needs:

- to only create tables & insert test data once for the whole suite (e.g. select tests)
- to only create tables once, but truncate & reset the test data for each test (e.g. insert / update tests)
- to drop & recreate the whole database for each test (`sync` tests).

We could drastically improve the performance of our test suite by doing this I think.

priyanshi-yb:
That's a great suggestion @ephys, thanks!
Yes, it will be a good exercise that'll benefit the entire test suite run, including future dialect additions and not just YugabyteDB. Do let us know if we could be of any help here.

Can we then move forward on this PR while the work to optimise the test runs is addressed separately?

ephys:
If you want to give it a try, removing [https://github.com/sequelize/sequelize/blob/main/test/integration/support.js#L23-L25](this forced reset of the database), and placing it in each test suite instead would be a good start.
That would be a good first PR, and once it's done we can split the actual optimization work to get this done faster.

This is going to be a lot of work. There is no easy way to say this, but we won't be able to merge this PR before that optimization work is done. We cannot afford to have every PR take more than 2 hours to complete, this would make maintenance very painful. 🙁

I'm hoping we can get this work done quickly, so we can get back to merging this PR.

PR #14352, opened 2022-04-09
feat!: block access to `/lib`

ephys:
### Pull Request Checklist

### Description Of Change

This PR prevents users from importing anything other than `@sequelize/core` and `@sequelize/core/package.json`.

With this change, it is safe to rework our internal source code without risking to introduce a breaking change.

If users need to access our internals, they should open a feature request so we can expose a properly tested solution.

Breaking change documented: https://github.com/sequelize/website/pull/78/ 

As a last resort, if a user *really* needs to access them, I added `@sequelize/core/_non-semver-use-at-your-own-risk_` as an alias for `@sequelize/core/lib`. Hopefully it is clear enough that these imports should only be used as a last resort and are not officially supported, and could break in any release, including patch releases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,9 +24,8 @@
       "import": "./lib/index.mjs",
       "require": "./lib/index.js"
     },
-    "./lib/*": "./lib/*.js",
-    "./lib/errors": "./lib/errors/index.js",
-    "./*": "./*"
+    "./_non-semver-use-at-your-own-risk_/*": "./lib/*",
+    "./package.json": "./package.json"
   },
   "engines": {
     "node": "^14.17.0 || >=16.0.0"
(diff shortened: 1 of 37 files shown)


WikiRik:
Why would a user want to import our package.json?

ephys:
Typically to access our version number (even though it's exposed as `Sequelize.vesrion`), like in this issue https://github.com/sequelize/sequelize/issues/13787. [Other packages](https://github.com/eslint/eslint/blob/main/package.json#L11) also expose their `package.json` even though their other modules are restricted, so there may be reasons I don't know about

WikiRik (approved):
Can't hurt to export package.json then, so should be good

PR #14359, opened 2022-04-11
feat: simplify withSchema API

ephys:
### Pull Request Checklist

### Description Of Change

As a follow up to https://github.com/sequelize/sequelize/pull/14337. While we're renaming `Model.schema` to `Model.withSchema`, we might as well simplify its API. I changed it from:

```typescript
MyModel.schema('public');
MyModel.schema('public', /* schema delimiter */ '.');
MyModel.schema('public', { schemaDelimiter: '.' });
```

to

```typescript
MyModel.withSchema('public');
MyModel.withSchema({ schema: 'public', schemaDelimiter: '.' });
```

Not a breaking change because `.schema` keeps its old API and `.withSchema` hasn't been released yet

### DIFF
--- a/test/support.d.ts
+++ b/test/support.d.ts
@@ -1,0 +1,5 @@
+declare namespace Chai {
+  interface Assertion {
+    deepEqual(expected: any): Assertion;
+  }
+}
--- a/src/model.d.ts
+++ b/src/model.d.ts
@@ -86,6 +86,8 @@ export interface DropOptions extends Logging {
  * Schema Options provided for applying a schema to a model
  */
 export interface SchemaOptions extends Logging {
+  schema: string;
+
   /**
    * The character(s) that separates the schema name from the table name
    */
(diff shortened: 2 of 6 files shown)


test/support.d.ts
```diff
@@ -0,0 +1,5 @@
+declare namespace Chai {
+  interface Assertion {
+    deepEqual(expected: any): Assertion;
```
WikiRik, line 3: Where would we need this for?

WikiRik (approved):
I have one question, but it's non-blocking

test/support.d.ts
ephys: `support.js` declares a custom chai method:

```
chai.Assertion.addMethod('deepEqual', function (expected, depth = 5) {
  expect(inspect(this._obj, { depth })).to.deep.equal(inspect(expected, { depth }));
});
```

which works like `deep.equal` but includes symbols

This declares it to TypeScript so `expect(y).to.deepEqual(x)` is valid

WikiRik: Ah, that makes sense

PR #14360, opened 2022-04-11
feat: add `isModelStatic`, `isSameInitialModel`

ephys:
### Pull Request Checklist

### Description Of Change

This PR adds two new utility methods:

- `isModelStatic`: returns true if the value is a subclass (not instance) of model
- `isSameInitialModel`: returns true if two models are variants of the same model

Both were initially created for https://github.com/sequelize/sequelize/pull/14280, but this PR also refactors the existing codebase to use them where it makes sense

### DIFF
--- a/src/utils/model-utils.ts
+++ b/src/utils/model-utils.ts
@@ -1,0 +1,27 @@
+import type { ModelStatic } from '../model';
+import { Model } from '../model';
+
+/**
+ * Returns true if the value is a model subclass.
+ *
+ * @param val The value whose type will be checked
+ */
+export function isModelStatic<M extends Model>(val: any): val is ModelStatic<M> {
+  return typeof val === 'function' && val.prototype instanceof Model;
+}
+
+/**
+ * Returns true if a & b are the same model.
+ * The difference with doing `a === b` is that this method will also
+ * return true if one of the models is scoped, or a variant with a different schema.
+ *
+ * @example
+ * isSameModel(a, a.scope('myScope')) // true;
+ *
+ * @param a
+ * @param b
+ */
+export function isSameInitialModel(a: ModelStatic<any>, b: ModelStatic<any>): boolean {
+  return isModelStatic(a) && isModelStatic(b)
+    && (a.getInitialModel() === b.getInitialModel());
+}
--- a/src/utils/model-utils.ts
+++ b/src/utils/model-utils.ts
@@ -11,12 +11,13 @@ export function isModelStatic<M extends Model>(val: any): val is ModelStatic<M> 
 }
 
 /**
- * Returns true if a & b are the same model.
+ * Returns true if a & b are the same initial model, ignoring variants created by {@link Model.withSchema}, {@link Model.withScope}, and the like.
+ *
  * The difference with doing `a === b` is that this method will also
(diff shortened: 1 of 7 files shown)


src/utils/model-utils.ts
```diff
@@ -0,0 +1,27 @@
...
+  return typeof val === 'function' && val.prototype instanceof Model;
+}
+
+/**
+ * Returns true if a & b are the same model.
+ * The difference with doing `a === b` is that this method will also
+ * return true if one of the models is scoped, or a variant with a different schema.
+ *
+ * @example
+ * isSameModel(a, a.scope('myScope')) // true;
```
WikiRik, line 19:
```suggestion
 * isSameInitialModel(a, a.scope('myScope')) // true;
```

src/utils/model-utils.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Returns true if the value is a model subclass.
+ *
+ * @param val The value whose type will be checked
+ */
+export function isModelStatic<M extends Model>(val: any): val is ModelStatic<M> {
+  return typeof val === 'function' && val.prototype instanceof Model;
+}
+
+/**
+ * Returns true if a & b are the same model.
```
WikiRik, line 14: Do you want to add some extra info here since we are checking for the same initial model?

test/unit/utils/model-utils.test.ts
```diff
@@ -0,0 +1,54 @@
...
+const sequelize: Sequelize = require('../../support').sequelize;
+
+describe('isModelStatic', () => {
+  it('returns true for model subclasses', () => {
+    const MyModel = sequelize.define('myModel', {});
+
+    expect(isModelStatic(MyModel)).to.be.true;
+  });
+
+  it('returns true for model instances', () => {
```
WikiRik, line 15:
```suggestion
  it('returns false for model instances', () => {
```

Or is the test incorrect?

WikiRik (commented):
Nice utility function to add. Do have a few remarks

test/unit/utils/model-utils.test.ts
ephys: The title was wrong :)

src/utils/model-utils.ts
ephys: See if the updated description is better

test/unit/utils/model-utils.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  it('returns false for the Model class', () => {
+    expect(isModelStatic(Model)).to.be.false;
+  });
+
+  it('returns false for the anything else', () => {
+    expect(isModelStatic(Date)).to.be.false;
+  });
+});
+
+describe('areSameInitialModel', () => {
```
WikiRik, line 30:
```suggestion
describe('isSameInitialModel', () => {
```

Since that's the name of the function we are testing

WikiRik (changes requested):
Final remark and then we're good to go

test/unit/utils/model-utils.test.ts
ephys: I thought I reverted that :x

WikiRik (approved):

PR #14368, opened 2022-04-12
fix(types): make `WhereOptions` more accurate

ephys:
Backport of PR #14022 to v6


WikiRik (approved):
Looking through it quickly doesn't show any major wrongs with backporting so should be good

nholmes3:
hey @ephys and @WikiRik 

I am upgrading from version 6.17.0 to version 6.21.1 (to include my fix for Postgres schemas) and the changes included in version 6.19.0  break my code in a couple places in a way that I can't quite wrap my head around...

My typescript version is 4.6.2

Break 1:
id is of type string
```
const instance = await this.scope(null).findOne({ where: { id } });

// this leads to the following typescript output:

error TS2769: No overload matches this call.
  Overload 1 of 2, '(this: ModelStatic<T>, options: NonNullFindOptions<Attributes<T>>): Promise<T>', gave the following error.
    Type '{ id: string; }' is not assignable to type 'WhereOptions<Attributes<T>>'.
      Types of property 'id' are incompatible.
        Type 'string' is not assignable to type 'WhereAttributeHashValue<Attributes<T>[string]>'.
  Overload 2 of 2, '(this: ModelStatic<T>, options?: FindOptions<Attributes<T>>): Promise<T>', gave the following error.
    Type '{ id: string; }' is not assignable to type 'WhereOptions<Attributes<T>>'.
      Types of property 'id' are incompatible.
        Type 'string' is not assignable to type 'WhereAttributeHashValue<Attributes<T>[string]>'.
```

Break 2:

```
await this.update(values, {
    where: { id: { [Op.in]: queue.splice(0, max) } },
});


// this leads to the following output:

error TS2769: No overload matches this call.
  Overload 1 of 2, '(this: ModelStatic<...>, values: { [key in keyof Attributes<T>]?: Fn | Col | Literal | Attributes<T>[key]; }, options: Omit<UpdateOptions<Attributes<T>>, "returning"> & { ...; }): Promise<...>', gave the following error.
    Type '{ id: { [OpTypes.in]: string[]; }; }' is not assignable to type 'WhereOptions<Attributes<T>>'.
      Types of property 'id' are incompatible.
        Type '{ [OpTypes.in]: string[]; }' is not assignable to type 'WhereAttributeHashValue<Attributes<T>[string]>'.
          Types of property '[Op.in]' are incompatible.
            Type 'string[]' is not assignable to type 'undefined'.
  Overload 2 of 2, '(this: ModelStatic<...>, values: { [key in keyof Attributes<T>]?: Fn | Col | Literal | Attributes<T>[key]; }, options: UpdateOptions<Attributes<T>>): Promise<...>', gave the following error.
    Type '{ id: { [OpTypes.in]: string[]; }; }' is not assignable to type 'WhereOptions<Attributes<T>>'.
      Types of property 'id' are incompatible.
        Type '{ [OpTypes.in]: string[]; }' is not assignable to type 'WhereAttributeHashValue<Attributes<T>[string]>'.
          Types of property '[Op.in]' are incompatible.
            Type 'string[]' is not assignable to type 'undefined'.
```


I would greatly appreciate any guidance you can offer on how to successfully upgrade past version 6.19.0

Thanks!

WikiRik:
@nholmes3 
I'm not the most comfortable with the typings, but these are my thoughts on it.

Break 1:
In the [`WhereAttribueHash` type](https://sequelize.org/api/v7/index.html#WhereAttributeHash) (type is the same for v6) I do see in the examples that you need a key-value pair. So instead of looking for one that has an `id`, you should look for one field where the `id` equals a specific value. I'm not so sure how well this worked in the past, but something like `{ id: { [Op.not]: null } }` should work as well.

Break 2:
Looking at the `Type 'string[]' is not assignable to type 'undefined'.` I have the idea that `queue` and/or `queue.splice(0, max)` is able to return `undefined` which we do not allow. So you might have to move that out of the statement and make a variable of it so it's only in the update call if it's not `undefined`

nholmes3:
@WikiRik thanks for your response!

For break 1 I have tried: 

```
const instance = await this.scope(null).findOne({ id: { [Op.eq]: id } });
```
which does not work 

For break 2 I have tried to separate out the variable for `queue.splice` with `as string[]` which still does not work

@ephys are you able to provide any feedback here?

I appreciate your time

PR #14376, opened 2022-04-13
feat: Add `Op.notStartsWith`, `Op.notEndsWith` and `Op.notSubstring`

### LINKED ISSUE #14375: [Feature Request] new operators: notStartsWith, notEndsWith and notSubstring
## Issue Creation Checklist

No, no bugs, no problem, but a solution. This is a new feature request for 3 new operators. The new requested ones are negations of other 3 already featured.

The current available operators `startsWith`, `endsWith` and `substring` are facilitators of the operator `like` (`LIKE` in SQL) that allow developers to use the `like` operator without having to transform the value. Ex: (with `startsWith`)
```js
Model.findAll({
    where: {
        field: {[Op.startsWith]: value}
    }
});
// is the same of the one below, but without the need to add a % to the value
Model.findAll({
    where: {
        field: {[Op.like]: `${value}%`}
    }
});
```
The idea of this is simply to add 3 new operators, just like the ones mentioned, but with the negation using the SQL `NOT LIKE` operator instead of `LIKE`. This would help developers the same way these other _LIKE-helpers_ did :)

```js
const {Op} = Sequelize;

function expressLikeMiddleware(req, res, next) {
    const {value1, value2, value3} = sanitizeSomehow(req.query);
    const posts = await Post.findAll({
        where: {
            title: {
                [Op.and]: [
                    {[Op.startsWith]: value1},
                    {[Op.notEndsWith]: value2} // new Op
                ]
            },
            subTitle: {[Op.notSubstring]: value3} // new Op
        }
    });
   //...
}
```

We already have `startsWith`, `endsWith` and `substring`. This should only add their negation variants, namely `notStartsWith`, `notEndsWith` and `notSubstring`, to the list of operators.

I already tried:
```js
Model.findAll({
    where: {
        name: {
            [Op.not]: {[Op.startsWith]: value}
        }
    }
});
Model.findAll({
    where: {
        name: {
            [Op.startsWith]: {[Op.not]: value}
        }
    }
});
```
None of them works

### Additional context
You might be asking why not to use `Op.notLike` and add the extra `%`. Consider that in some complex querying systems, the value needs to be immutable or can not change (like when you have a complex set of conditions in a function to change the operator but no access to the value). So that's where it comes very handy.
Ex:
```js
getSequelizeOperator(op) {
    switch(op) {
        case "eq": return Op.eq;
        case "gt": return Op.gt;
        //...
        case "sw": return Op.startsWith;
        case "nsw": return Op.notStartsWith; // new Op
        case "in": return Op.substring;
        case "nin": return Op.notSubstring; // new Op
        //...
    }
}
```

- [ ] Yes. This feature only applies to the following dialect(s): XXX, YYY, ZZZ

ReDemoNBR:
### Pull Request Checklist

### Description Of Change
Issue (feature request by myself): https://github.com/sequelize/sequelize/issues/14375

Closes #14375

### Todos

- [x] Add operators (notStartsWith, notEndsWith and notSubstring)
- [x] Add tests and ensure no regressions
- [x] Add documentations


ephys (approved):
Edited your message to mark #14375 as closed by this PR

I would typically be against adding notX versions of operators as you can do `{ Op.not: Op.startsWith: ... }`

But we already have notLike, notILike, and `a NOT LIKE b` is an actual operator different from  `NOT (a LIKE b)`.

I do wish we implemented the new behavior for `substring`, `startsWith`, `endsWith` before extending them. I don't know if we'll have time to do that breaking change before the release of v7 though.

Note that v7 is very close to a stable release, so v6 is considered feature frozen as much as possible. This will only release with v7

All that being said, this looks good to me. Thank you :)

PR #14380, opened 2022-04-13
docs: redirect engines.md to our website

ephys:
This PR removes the Supported Database table from ENGINE.md, redirecting the user to https://sequelize.org/releases/ instead, to deduplicate the information that we have to maintain :)

### DIFF
--- a/ENGINE.md
+++ b/ENGINE.md
@@ -1,3 +1,0 @@
-# Database Engine Support
-
-You can find the information about which database each version of Sequelize supports on our [Versioning Page](https://sequelize.org/releases/)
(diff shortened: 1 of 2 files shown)


WikiRik:
Do we need this file? It was added in https://github.com/sequelize/sequelize/pull/12168 without a real reason imo.

It might be better to refer to the Releases page from the README and then we should also be fine

ephys:
I kept it in case someone referenced to it, but it's not worth cluttering our main repo to avoid breaking a link.
I'll remove it and update the readme

WikiRik (approved):

Issue #14295, opened 2022-03-30
[RFC] Consider deprecating & replacing `DataTypes.DATE`


ephys:
`DataTypes.DATE` is a bit of a mess. It has very different meanings depending on the dialect. Some implementations include a timezone, others do not.
It doesn't provide a way to choose whether the type should include a timezone.
Finally, `DATE` means `DATEONLY` in SQL, not `DATETIME`.

I propose we deprecate that DataType and replace it with:

- `DataTypes.DATETIME`: Always means "Date without timezone".
- `DataTypes.DATETIME.ZONED`: Always means "Date with timezone", throws if the dialect doesn't support this.

mytharcher:
My proposal for the naming:

* `DateTypes.DATETIME`: Always means "Date **with** timezone". Because it will work fine in most cases by default.
* `DataTypes.DATETIME.PLAIN`: Always means "Date **without** timezone". It could only accepts `Temporal.PlainDateTime`.

Many developers arguing in other issues for demanding of date/time without timezone, because they didn't find out the best practice of usage of time.

In my opinion, when we want to use a date/time field in database, it should always be with timezone. And physically, time is always with timezone. The cases when people are not mentioning timezone, mostly they only consider the local time. But use time with timezone in these cases are also acceptable and no bad effects.

The base reason why should we use date/time type of fields in database, is Just because we need them to be calculate with other values (comparing, manipulating, etc.). If people only need time for literally expressing, the timezone will not be strictly necessary. In this case, just use plain time or even string will be ok. But if we use time for calculating, we need the exact value (with timezone).

Another thing is, always use UTC formatted in any transferring is the best practice. Also for storing to database. So in PG it works great by default. It will be the clients' stuff to transform stored time to local timezone in view. And by default `Date` in browser parsing an UTC time string works right.

Think about it, if we use time with timezone in a project, whether if it is a local one or global, it always work fine. But if use time without timezone, it only works for local projects.

crazyscience:
> But use time with timezone in these cases are also acceptable and no bad effects.

Super not true. When you have a table with 30 billion rows and 4 datetime columns in each row, the addition of useless data (in this case, timezone info) definitely has an impact. Maybe not noticeable when someone's building yet another todo app. But ORMs should be built for performance application and production loads, not hobbyists building a blog.

[...]

PR #14400, opened 2022-04-15
refactor: ♻️ migrate `moment.js` to `dayjs`

DraftProducts:
### Pull Request Checklist

### Description Of Change

Migrate `moment.js` to `dayjs`.
Closes: #13372


DraftProducts:
I have 4 failing tests.
It looks like it's not coming from this PR : *(please note that the results are same with sqlite and postgres tests)*
image

WikiRik:
> I have 4 failing tests.
> Looks like it's not from this PR : *(please note that the results are same with sqlite and postgres tests)*
> image
> 

Yes, these stem from #14390. No need to fix them in this PR

test/integration/data-types.test.js
```diff
@@ -46,15 +46,15 @@ describe(Support.getTestDialectTeaser('DataTypes'), () => {
     await current.sync({ force: true });
 
     await User.create({
-      dateField: moment('2011 10 31', 'YYYY MM DD'),
+      dateField: dayjs('2011 10 31', 'YYYY MM DD'),
```
ephys, line 49: We should not accept custom dayjs objects here, to avoid introducing a breaking change when we remove the dayjs dependency
This should throw instead

src/sequelize.js
```diff
@@ -150,7 +151,7 @@ export class Sequelize {
...
    * @param {object}   [options.sync={}] Default options for sequelize.sync
-   * @param {string}   [options.timezone='+00:00'] The timezone used when converting a date from the database into a JavaScript date. The timezone is also used to SET TIMEZONE when connecting to the server, to ensure that the result of NOW, CURRENT_TIMESTAMP and other time related functions have in the right timezone. For best cross platform performance use the format +/-HH:MM. Will also accept string versions of timezones used by moment.js (e.g. 'America/Los_Angeles'); this is useful to capture daylight savings time changes.
+   * @param {string}   [options.timezone='+00:00'] The timezone used when converting a date from the database into a JavaScript date. The timezone is also used to SET TIMEZONE when connecting to the server, to ensure that the result of NOW, CURRENT_TIMESTAMP and other time related functions have in the right timezone. For best cross platform performance use the format +/-HH:MM. Will also accept string versions of timezones used by dayjs (e.g. 'America/Los_Angeles'); this is useful to capture daylight savings time changes.
```
ephys, line 154:
```suggestion
   * @param {string}   [options.timezone='+00:00'] The timezone used when converting a date from the database into a JavaScript date. The timezone is also used to SET TIMEZONE when connecting to the server, to ensure that the result of NOW, CURRENT_TIMESTAMP and other time related functions have in the right timezone. For best cross platform performance use the format +/-HH:MM. Will also accept string versions of timezones supported by Intl (e.g. 'America/Los_Angeles'); this is useful to capture daylight savings time changes.
```

src/utils/dayjs.ts
```diff
@@ -0,0 +1,16 @@
+import dayjs from 'dayjs';
+import timezone from 'dayjs/plugin/timezone';
+import utc from 'dayjs/plugin/utc';
+
+dayjs.extend(utc);
+dayjs.extend(timezone);
+
+export function isValidTimeZone(tz: string) {
```
ephys, line 8: can we memoize these results?

src/sequelize.d.ts
```diff
@@ -278,7 +278,7 @@ export interface Options extends Logging {
    * used to SET TIMEZONE when connecting to the server, to ensure that the result of NOW, CURRENT_TIMESTAMP
    * and other time related functions have in the right timezone. For best cross platform performance use the
    * format
-   * +/-HH:MM. Will also accept string versions of timezones used by moment.js (e.g. 'America/Los_Angeles');
+   * +/-HH:MM. Will also accept string versions of timezones used by dayjs (e.g. 'America/Los_Angeles');
```
ephys, line 281:
```suggestion
   * +/-HH:MM. Will also accept string versions of timezones supported by Intl (e.g. 'America/Los_Angeles');
```

ephys (changes requested):
Thank you for the PR :)
Overall, looks good, my main comment is that we should not include a built-in support for dayjs objects (i.e. they should not be accepted as values by DataTypes.DATE), to avoid needing to introduce another breaking change when Temporal becomes stable

src/utils/dayjs.ts
DraftProducts: Why should we memorize it ?
I'm not sure I understand what you mean 🤔

test/integration/data-types.test.js
DraftProducts: Okay, I would change that 👍
To be sure, we throw an error if it's another type or instance than : 
 - Date instance
 - string value

src/utils/dayjs.ts
ephys: So subsequent calls don't need to create a DateTimeFormat & Error object to know the result

test/integration/data-types.test.js
ephys: I think `number` (for timestamp) is supported too, maybe `bigint` but it should be double checked

DraftProducts: Hi, @ephys after multiple tests, I don't think throwing an error if the time is not a Date, string, number is a good idea.
We should accept any value which can be stringified to a value readable by a Date object.
```js
import dayjs from "dayjs";

console.log(new Date(dayjs().toString()));
console.log(new Date(dayjs()));
// Fri Apr 15 2022 19:40:06 GMT+0200 (heure d’été d’Europe centrale)
```

As we accept custom dayjs objects, we also accept any other custom date object that can be stringified.

```js
import dayjs from "dayjs";
import moment from "moment";

console.log(dayjs(moment()).toDate());
// Fri Apr 15 2022 19:40:06 GMT+0200 (heure d’été d’Europe centrale)
```

We should let the validator validate the type and then let Dayjs parse the string.

I can remove the `dayjs.isDayjs` method if you think it's needed, but it's only an instance checker to avoid changing the timezone if the developer has done the job in his codebase.

src/utils/dayjs.ts
DraftProducts: Sure 👍

test/integration/data-types.test.js
ephys: Yes, actually that sounds fine.

My concern was avoiding a breaking change once we migrate to `Temporal`, but I think the way forward to support temporal will be to introduced [new DataTypes](https://github.com/sequelize/sequelize/issues/14295) instead, so `DataTypes.DATE` will keep using dayjs or other `Date` based libraries.

How does the timeZone option behave if a moment object is passed to the method?

DraftProducts:
Sorry for the review request, your answer wasn't loaded

test/integration/data-types.test.js
DraftProducts: Nice 👍 

The timezone is not concerned since it's a config option and used as the second argument of dayjs function.
The moment object is not receive as a moment object, only as a stringified version of the date.

ephys:
Looks like there is a failing test when the timeZoneOffset is `+00:00`. Can you ping me once the tests pass? Or if you need help :)

src/sequelize.d.ts
fzn0x: > 

I think it's should be Intl.Locale

src/data-types.js
```diff
@@ -518,18 +518,18 @@ class DATE extends ABSTRACT {
...
+      return dayjs(date).utcOffset(options.timezone);
     }
 
-    return momentTz(date);
+    return dayjs(date);
   }
 
   _stringify(date, options) {
-    if (!moment.isMoment(date)) {
+    if (!dayjs.isDayjs(date)) {
```
ephys, line 532: This is the only line that worries me.
Doesn't this mean that, if `options.timezone` is set, then all dates will have their timeZone changed, except if it's an instance of dayjs

I think we should either always overwrite the timeZone (including dayjs)

Edit: it looks like that was the initial intention, as [a recent PR](https://github.com/sequelize/sequelize/issues/13816) introduced this change. I'll look into it more

DraftProducts: Okay 👍

DraftProducts:
Have you an idea of where is coming from the last CI fail ? (CI / DB2 (Node 14))
I had a look, but I don't understand where is the link with our changes.

DraftProducts:
Hi @ephys Have you an idea of where is coming from the last CI fail ? (CI / DB2 (Node 14))
I had a look, but I don't understand where is the link with our changes.

ephys:
That DB2 test is randomly failing, you can ignore it. Sorry for having it waste your time

src/data-types.js
fzn0x: > This is the only line that worries me. Doesn't this mean that, if `options.timezone` is set, then all dates will have their timeZone changed, except if it's an instance of dayjs
> 
> I think we should either always overwrite the timeZone (including dayjs)
> 
> Edit: it looks like that was the initial intention, as [a recent PR](https://github.com/sequelize/sequelize/pull/13818) introduced this change. I'll look into it more, see if that change should not be reverted.

I can help you to explain the change, let's checkout to that PR.

ephys: After the discussion in the other thread, I think it's safe to remove the condition, and call _applyTimezone to all inputs

DraftProducts: Okay 👍

sdepold:
What's the status of this one?

DraftProducts:
Okay, last changes requested are done.

DraftProducts:
No conflicts anymore 👍

sdepold (commented):
looks fine overall!

yarn.lock
```diff
@@ -5882,7 +5887,7 @@ moment-timezone@^0.5.15, moment-timezone@^0.5.34:
   dependencies:
     moment ">= 2.9.0"
 
-"moment@>= 2.9.0", moment@^2.29.1, moment@^2.29.3:
+"moment@>= 2.9.0", moment@^2.29.3:
```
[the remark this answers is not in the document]
DraftProducts: Yes, it's still needed, moment is a dependency of `snowflake-sdk@1.6.9` and `mariadb@2.5.6`.

You might wonder what the point of this PR would be if we keep the `moment` dependency.
The advantage will be that it will no longer be an internal dependency of the project.
It will only be needed if the developer needs the `mariadb` peer dependency or if the developer is a sequelize contributor because `snowflake-sdk` is a development dependency.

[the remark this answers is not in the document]
sdepold: Makes sense :)

sdepold (approved):

Issue #14358, opened 2022-04-11
sequelize.query() handling of replacement params for valid SQL operators ?& , ?|


mariusa:
## Issue Creation Checklist

- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description

Trying to run these 2 raw queries with `sequelize.query` and passing the array as a replacement, so it would handle escaping?
```
SELECT * FROM product where tags ?& ARRAY['tag1', 'tag2']; 
SELECT * FROM product where tags ?| ARRAY['tag1', 'tag2']; 
```
`product.tags` is a JSONB array in Postgres.

I tried
```
let [recs, meta] = await sequelize.query(`SELECT * FROM product WHERE tags ?| ?`, {
        replacements: [['tag1', 'tag2']]
    })
```

but the `?` in `?|` operator is parsed as a replacement.
Thanks!

### What do you expect to happen?

`?|`, `?&` should be left intact, these are valid Postgres operators

### What is actually happening?

the `?` in `?|` operator is parsed as a replacement.

### Environment

- Sequelize version: 6.18.0
- Node.js version: 16.x
- Database & Version: 'PostgreSQL 14'
- Connector library & Version: pg 8

## Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [x] No, I don't have the time and I wouldn't even know how to start.

ephys:
I believe this is the part of the code at fault: https://github.com/sequelize/sequelize/blob/main/src/sql-string.js#L110-L124

It searches for any `?` character, without checking if it's part of something else.

We need to improve the regex, but at the same time other things should keep working, list casting `?::integer`

fzn0x:
> I believe this is the part of the code at fault: https://github.com/sequelize/sequelize/blob/main/src/sql-string.js#L110-L124
> 
> It searches for any `?` character, without checking if it's part of something else.
> 
> We need to improve the regex, but at the same time other things should keep working, list casting `?::integer`

I think we should add unit tests for all the regex that is in the Sequelize internal, seems like a lot of similar issues in other functions.

ephys:
PR opened: https://github.com/sequelize/sequelize/pull/14404

PR #14404, opened 2022-04-16
fix: dont treat `?&` or `?|` as replacements

ephys:
Special cases the `?&` and `?|` postgres operators so their question mark is not treated as a replacement. 

See https://github.com/sequelize/sequelize/issues/14358 for details.

Closes https://github.com/sequelize/sequelize/issues/14358


fzn0x (approved):

WikiRik:
What is the intended behaviour for things like `?|?`, `??|` or  ?||`?

ephys:
Very good question :) Let's add tests, especially since `||` is an actual operator

ephys:
`?|?` would be treated as `?|<replacement>`
`??|` would be treated as `<replacement>?|`
`?||` would be treated as `?||`

Should we make all `?` preceded or succeded by `?`, `&` or `|` as not being replacements?

WikiRik:
No, I think this should be fine. But I think we should add these operators to sequelize directly so they are more robust. That's for another PR though

WikiRik:
Maybe the `?||` should become `<replacement>||` though, but not sure

ephys:
> should add these operators to sequelize directly so they are more robust

Unfortunately, because of when bind & replacements are parsed, they would still be interpreted as replacements (cf the recent conversation we had in core), until https://github.com/sequelize/sequelize/tree/ephys/fix-13817 is implemented that is

> Maybe the `?||` should become `<replacement>||` though, but not sure

I would actually say the opposite: always consider that ? to not be a replacement so people learn to surround their replacement with a space, to avoid clashes with potential future operators

WikiRik:
> > should add these operators to sequelize directly so they are more robust
> 
> Unfortunately, because of when bind & replacements are parsed, they would still be interpreted as replacements (cf the recent conversation we had in core), until https://github.com/sequelize/sequelize/tree/ephys/fix-13817 is implemented that is

Ah yes

> > Maybe the `?||` should become `<replacement>||` though, but not sure
> 
> I would actually say the opposite: always consider that ? to not be a replacement so people learn to surround their replacement with a space, to avoid clashes with potential future operators

Hmm... I did not really consider that before. That might not be too bad. I think we should just merge this one for now and think about this again later (but before we release v7)

ephys:
Sounds good to me

PR #14407, opened 2022-04-17
refactor: migrate test support to ts

ephys:
### Pull Request Checklist

### Description Of Change

This PR migrates `test/supports.js` and its dependencies to TS. Some of these changes were initially done in https://github.com/sequelize/sequelize/pull/14280, but they're often needed in new tests (such as the ones being written in https://github.com/sequelize/sequelize/tree/ephys/fix-13817), so this PR extracts them out of that PR


test/support.ts
```diff
@@ -0,0 +1,350 @@
...
+
+  if (onNext || unhandledRejections) {
+    return;
+  }
+
+  console.error('An unhandled rejection occurred:');
+  throw e;
+});
+
+afterEach(() => {
```
WikiRik, line 62: Why did you remove the check for global.afterEach?

test/support.ts
```diff
@@ -0,0 +1,350 @@
...
+ * during this test (instead of failing the test)
+ */
+export async function nextUnhandledRejection() {
+  return new Promise((resolve, reject) => {
+    onNextUnhandledRejection = reject;
+  });
+}
+
+/**
+   * Pushes all unhandled rejections that occur during this test onto destArray
```
WikiRik, line 78: I think this indentation is off

test/support.ts
```diff
@@ -0,0 +1,350 @@
...
+
+  return sequelize;
+}
+
+export function createSequelizeInstance(options: Options = {}) {
+  options.dialect = getTestDialect();
+
+  const config = Config[options.dialect];
+
+  const sequelizeOptions = defaults(options, {
```
WikiRik, line 123: Why is `odbcConnectionString` not mentioned here anymore?

test/support.ts
```diff
@@ -0,0 +1,350 @@
...
+      return x;
+    }
+  }
+
+  // @ts-expect-error
+  return new ModdedQueryGenerator({ sequelize, _dialect: sequelize.dialect });
+}
+
+export function getTestDialect(): Dialect {
+  let envDialect = process.env.DIALECT || 'mysql';
```
WikiRik, line 212: Why would we default to mysql here? I know it's in the original file, but seems odd to me

test/support.ts
```diff
@@ -0,0 +1,350 @@
...
+
+export function expectsql(
+  query: { query: string, bind: unknown } | Error,
+  assertions: { query: PartialRecord<ExpectationKey, string | Error>, bind: PartialRecord<ExpectationKey, unknown> },
+): void;
+export function expectsql(
+  query: string | Error,
+  assertions: PartialRecord<ExpectationKey, string | Error>,
+): void;
+export function expectsql(
```
WikiRik, line 250: Why are we exporting this multiple times?

test/support.ts
```diff
@@ -0,0 +1,350 @@
...
+export function rand() {
+  return Math.floor(Math.random() * 10e5);
+}
+
+export function isDeepEqualToOneOf(actual: unknown, expectedOptions: unknown[]): boolean {
+  return expectedOptions.some(expected => isDeepStrictEqual(actual, expected));
+}
+
+/**
+   * Reduces insignificant whitespace from SQL string.
```
WikiRik, line 299: Indentation is off

test/support.ts
```diff
@@ -0,0 +1,350 @@
...
+   *
+   * @param sql the SQL string
+   * @returns the SQL string with insignificant whitespace removed.
+   */
+export function minifySql(sql: string): string {
+  // replace all consecutive whitespaces with a single plain space character
+  return sql.replace(/\s+/g, ' ')
+  // remove space before coma
```
WikiRik, line 307:
```suggestion
  // remove space before comma
```

test/support.ts
```diff
@@ -0,0 +1,350 @@
...
+    .replace(/ ,/g, ',')
+  // remove whitespace at start & end
+    .trim();
+}
+
+export const sequelize = createSequelizeInstance();
+
+before(function onBefore() {
```
WikiRik, line 315: Why is the check for global.beforeEach removed here?

WikiRik (commented):
Few remarks/questions

test/support.ts
ephys: It wasn't correctly set (according to the typings, and its usage in ibmi's connection manager). The `obdcConnectionString` isn't a valid option, it's named `dialectOptions.obdcConnectionString`. Which means it was unused

test/support.ts
ephys: Sounds weird to me too. I think we should throw if DIALECT is not set

test/support.ts
ephys: It's overloading (https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads)

WikiRik: Ah yes, I was looking for the term of this already. But the thing I am missing here is why we use this here, because we have 2 arguments in all three options. And the third just looks like a merger of the first two. But probably I'm missing something here in my thought process

ephys: The implementation is always a merger of all previous overloads, but it's 'internal' of sorts (even if exported), TypeScript won't include it as part of the possible overloads when you use the method

test/support.ts
ephys: All of our tests are run by mocha, so these are always defined

WikiRik (approved):

PR #14415, opened 2022-04-20
ci: test v7 against node 18

ephys:
### Description Of Change

This PR replaces node 16 with node 18 in our CI, so that we test against both our minimum and maximum supported node versions


WikiRik:
Can we wait until 2022-10-25 with this so Node 18 has had its first LTS release?

ephys:
We're going to need to wait a little bit more, until `eslint-plugin-jsdoc` supports node 18

ephys:
> Can we wait until 2022-10-25 with this so Node 18 has had its first LTS release?

Should we? I can understand waiting until it reaches LTS before upgrading an app, but I think a library should be ready before then

WikiRik:
We skipped Node 17 as well and currently Node 18 is in the same state. I agree that we should be ready to support it when the first LTS release is out, but don't think replacing Node 16 with this is the way to go at this moment. Maybe we can add the latest version as a third version to test?
Or we wait a bit until I finally have written down my ideas for a more comprehensive test suite that runs once a week (up for debate) and focusses on more combinations of Node versions, database connector versions and database versions.

ephys:
We skipped node 17 because we didn't plan on supporting it at all, as the release is only maintained for 6 months.
There is no real difference in stability between node 18 current and node 18 LTS, if we want to support it we should start today instead of when it gets promoted to LTS

> Maybe we can add the latest version as a third version to test?

We can, but this would increase our CI time to ~22 minutes, from ~15 minutes currently.

WikiRik:
> We're going to need to wait a little bit more, until `eslint-plugin-jsdoc` supports node 18

I have a feeling we'll get a new release from that later today.

WikiRik:
Okay, after the update of `eslint-plugin-jsdoc` we still have `@azure/msal-node` left (through `tedious` -> `@azure/identity`). PR is open for that, afterwards we should be able to install dependencies without adding `--ignore-engines`

Just for now, you could just add `--ignore-engines` so we can at least make sure that tests pass.

WikiRik:
Another update, it looks like `@azure/msal-node` will get a regular update in about 2 weeks which will most likely include the support for Node 18.

WikiRik:
Linting fails, see https://github.com/lo1tuma/eslint-plugin-mocha/pull/325 for context

WikiRik:
Let's wait until https://github.com/TypeStrong/typedoc/issues/1935 is solved and then we should be good to merge this

WikiRik (approved):
Tests pass, just need to update it in the repo settings.

ephys:
Wooh 🥳

PR #14437, opened 2022-04-23
fix(mysql/mariadb): upsert on duplicate key allow calculations by using values instead of VALUES()

### LINKED ISSUE #12834: Increment or Decrement value using Sequelize upsert() does not work when updating
## Issue Description

`upsert()` method in `sequelize` does not increment value when it updates using `sequelize.literal()`, it only modifies it.

### What are you doing?
In Sequelize with a MySQL Database I am using upsert() to insert a new row if it doesn't exist or increment one of its field values by some other value if it exists.

Inserting works fine but when the value gets updated it is not getting incremented but completely updated

    let x = 5;

    OrderDetails.upsert({ 
       menuItemId: 1234, 
       orderId: 4321, 
       quantity: sequelize.literal((`quantity + ${x}`)
    })

This results in this query:

    INSERT INTO `OrderDetails` (`menuItemId`,`orderId`,`quantity`,`createdAt`,`updatedAt`) VALUES (?,?,quantity + 5,?,?) 
    ON DUPLICATE KEY UPDATE `menuItemId`=VALUES(`menuItemId`),`orderId`=VALUES(`orderId`),`quantity`=VALUES(`quantity`),`updatedAt`=VALUES(`updatedAt`);

So if for example the previous value was `10` then after the query the updated value will be `5` instead of `15`. 

My guess is that there is a kind of bug in Sequelize when using `upsert()` with `sequelize.literal()`. Apparently, it is only recognized in the insertion and not in the update.

**_NOTE:_** This obviously **works** if I use `update()` with `sequelize.literal()`

The specified field should be incremented instead of completely modified and replaced with the value

If the previous value in my table was `10` then after executing the query the updated value will be `5` instead of `15`. 

### Additional context

None.

### Environment

- Sequelize version: `6.3.5`
- Node.js version: `12.18.1`
- Operating System: `Linux Ubuntu 18.04.1`

- [X] No, I don't have the time and **I wouldn't even know how to start.**

elijahb:
### Pull Request Checklist

- [ x] Have you added new tests to prevent regressions?
- [ x] Does `yarn test` or `yarn test-DIALECT` pass with this change (including linting)?
- [ x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ x] Did you update the typescript typings accordingly (if applicable)?
- [ x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)?

### Description Of Change
Doing an upsert in MySQL/MariaDB sequelize uses the `VALUES()` function in the `ON DUPLICATE` expression which presents the problem that if you want to calculate the value, it won't work if it needs to do an update. This closes #12834.

Example of what currently doesn't work and this PR fixes:
```js
// this works
const [insertUser] =await User.upsert({
  counter: 10
})
// this doesn't work
const [user] = await User.upsert({
  id: insertUser.id,
  counter: this.sequelize.literal('`counter` + 1')
})
await user.reload()
// user.counter should be 11, but it's null
```

_Also, MySQL at version 8 shows a deprecation warning if using that function in the ON DUPLICATE. I just changed it here for the `insert` and not the `insertBulk`, but probably should be removed from there too in another PR._

### DIFF
--- a/src/dialects/abstract/query-generator.js
+++ b/src/dialects/abstract/query-generator.js
@@ -118,7 +118,7 @@ export class AbstractQueryGenerator {
     const bind = [];
     const fields = [];
     const returningModelAttributes = [];
-    const values = [];
+    const values = {};
     const quotedTable = this.quoteTable(table);
     const bindParam = options.bindParam === undefined ? this.bindParam(bind) : options.bindParam;
     let query;
@@ -174,9 +174,9 @@ export class AbstractQueryGenerator {
           if (!this._dialect.supports.autoIncrement.defaultValue) {
             fields.splice(-1, 1);
           } else if (this._dialect.supports.DEFAULT) {
-            values.push('DEFAULT');
+            values[key] = 'DEFAULT';
           } else {
-            values.push(this.escape(null));
+            values[key] = this.escape(null);
           }
         } else {
           if (modelAttributeMap && modelAttributeMap[key] && modelAttributeMap[key].autoIncrement === true) {
@@ -184,9 +184,9 @@ export class AbstractQueryGenerator {
           }
 
           if (value instanceof Utils.SequelizeMethod || options.bindParam === false) {
-            values.push(this.escape(value, modelAttributeMap && modelAttributeMap[key] || undefined, { context: 'INSERT' }));
+            values[key] = this.escape(value, modelAttributeMap && modelAttributeMap[key] || undefined, { context: 'INSERT' });
           } else {
-            values.push(this.format(value, modelAttributeMap && modelAttributeMap[key] || undefined, { context: 'INSERT' }, bindParam));
+            values[key] = this.format(value, modelAttributeMap && modelAttributeMap[key] || undefined, { context: 'INSERT' }, bindParam);
           }
         }
       }
@@ -207,7 +207,7 @@ export class AbstractQueryGenerator {
         // do an update.  Instead, fall back to DO NOTHING.
         onDuplicateKeyUpdate += _.isEmpty(updateKeys) ? ' DO NOTHING ' : ` DO UPDATE SET ${updateKeys.join(',')}`;
       } else {
-        const valueKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=VALUES(${this.quoteIdentifier(attr)})`);
+        const valueKeys = options.updateOnDuplicate.map(attr => `${this.quoteIdentifier(attr)}=${values[attr]}`);
         // the rough equivalent to ON CONFLICT DO NOTHING in mysql, etc is ON DUPLICATE KEY UPDATE id = id
         // So, if no update values were provided, fall back to the identifier columns provided in the upsertKeys array.
         // This will be the primary key in most cases, but it could be some other constraint.
@@ -231,7 +231,7 @@ export class AbstractQueryGenerator {
       onConflictDoNothing: options.ignoreDuplicates ? this._dialect.supports.inserts.onConflictDoNothing : '',
       attributes: fields.join(','),
       output: outputFragment,
-      values: values.join(','),
+      values: Object.values(values).join(','),
       tmpTable,
     };
 
--- a/test/integration/model/upsert.test.js
+++ b/test/integration/model/upsert.test.js
@@ -570,6 +570,20 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         });
       }
 
+      if (dialect === 'mysql' || dialect === 'mariadb') {
+        it('should allow to use calculated values on duplicate', async function () {
+          const [insertUser] = await this.User.upsert({
+            bar: 10,
+          });
+          const [user] = await this.User.upsert({
+            id: insertUser.id,
+            bar: this.sequelize.literal('`bar` + 1'),
+          });
+          await user.reload();
+          expect(user.bar).to.equal(11);
+        });
+      }
+
       if (current.dialect.supports.returnValues) {
         describe('returns values', () => {
           it('works with upsert on id', async function () {
--- a/test/integration/model/upsert.test.js
+++ b/test/integration/model/upsert.test.js
@@ -572,7 +572,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
 
       if (dialect === 'mysql' || dialect === 'mariadb') {
         it('should allow to use calculated values on duplicate', async function () {
-          const [insertUser] = await this.User.upsert({
+          const insertUser = await this.User.create({
             bar: 10,
           });
           const [user] = await this.User.upsert({
--- a/test/unit/dialects/mysql/query-generator.test.js
+++ b/test/unit/dialects/mysql/query-generator.test.js
@@ -571,6 +571,13 @@ if (dialect === 'mysql') {
           },
           needsSequelize: true,
         },
+        {
+          arguments: ['myTable', { name: 'foo' }, undefined, { updateOnDuplicate: ['name'] }],
+          expectation: {
+            query: 'INSERT INTO `myTable` (`name`) VALUES ($1) ON DUPLICATE KEY UPDATE `name`=$1;',
+            bind: ['foo'],
+          },
+        },
       ],
 
       bulkInsertQuery: [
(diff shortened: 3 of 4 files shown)


src/dialects/abstract/query-generator.js
```diff
@@ -175,19 +175,19 @@ export class AbstractQueryGenerator {
...
+            values[key] = this.escape(null);
           }
         } else {
           if (modelAttributeMap && modelAttributeMap[key] && modelAttributeMap[key].autoIncrement === true) {
             identityWrapperRequired = true;
           }
 
           if (value instanceof Utils.SequelizeMethod || options.bindParam === false) {
-            values.push(this.escape(value, modelAttributeMap && modelAttributeMap[key] || undefined, { context: 'INSERT', replacements: options.replacements }));
+            values[key] = this.escape(value, modelAttributeMap && modelAttributeMap[key] || undefined, { context: 'INSERT' replacements: options.replacements });
```
ephys, line 188: Hey! Thank you for this PR.

You are missing a comma here:

```suggestion
            values[key] = this.escape(value, modelAttributeMap && modelAttributeMap[key] || undefined, { context: 'INSERT', replacements: options.replacements });
```

Ping me when your tests pass (or if only db2 is failing) so I can review and merge this :)

elijahb:
@ephys I've fixed the merge mistake. I also had to do another fix, there was a new function `mapBindParameters` that didn't replace the parameter correctly if it was followed by a semi-colon (for example: `SELECT * FROM users WHERE id = $id;`, because of that last semi-colon the $id parameter wouldn't be replaced).

test/unit/query-interface/upsert.test.ts
```diff
@@ -36,8 +36,8 @@ describe('QueryInterface#upsert', () => {
     const firstCall = stub.getCall(0);
     expectsql(firstCall.args[0] as string, {
       default: 'INSERT INTO [Users] ([firstName]) VALUES ($sequelize_1) ON CONFLICT ([id]) DO UPDATE SET [firstName]=EXCLUDED.[firstName];',
-      mariadb: 'INSERT INTO `Users` (`firstName`) VALUES ($sequelize_1) ON DUPLICATE KEY UPDATE `firstName`=VALUES(`firstName`);',
-      mysql: 'INSERT INTO `Users` (`firstName`) VALUES ($sequelize_1) ON DUPLICATE KEY UPDATE `firstName`=VALUES(`firstName`);',
+      mariadb: 'INSERT INTO `Users` (`firstName`) VALUES ($sequelize_1) ON DUPLICATE KEY UPDATE `firstName`=$sequelize_1;',
+      mysql: 'INSERT INTO `Users` (`firstName`) VALUES ($sequelize_1) ON DUPLICATE KEY UPDATE `firstName`=$sequelize_1;',
```
ephys, line 40: It's likely that we'll eventually add a `Model.bulkUpsert` method, which would generate something like this:

```sql
INSERT INTO `Users` (`firstName`) VALUES ($sequelize_1),($sequelize_2) ON DUPLICATE KEY UPDATE `firstName`=VALUES(`firstName`);
```

In that scenario, using `VALUES(firstName)` is important as it will return a different value for each row.

I think a way to reconcile both would be to have the upsert query be generated using `VALUES(...)` by default, but use the value itself if the value if an instanceof `SequelizeMethod` (for literals).

Basically,

```js
const [insertUser] = await User.upsert({
  counter: 10
});
```

would use `ON DUPLICATE KEY UPDATE counter = VALUES(counter)`,

but 

```js
const [insertUser] = await User.upsert({
  counter: literal('counter + 1')
});
```

would use `ON DUPLICATE KEY UPDATE counter = counter + 1`,

test/unit/dialects/mysql/query-generator.test.js
```diff
@@ -571,6 +571,15 @@ if (dialect === 'mysql') {
           },
           needsSequelize: true,
         },
+        {
+          arguments: ['myTable', { name: 'foo' }, undefined, { updateOnDuplicate: ['name'] }],
```
ephys, line 575: Could you use a `literal()` as the value for `name` in this test?

Also, would you mind moving this test over to [this test file](https://github.com/sequelize/sequelize/blob/main/test/unit/query-interface/upsert.test.ts) ? It would be interesting to test how each dialect handles this (even if the solution for the other dialects is just to throw an error).

This way we can also remove the test duplication between mysql & mariadb

test/integration/model/upsert.test.js
```diff
@@ -570,6 +570,20 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
       }
 
+      if (dialect === 'mysql' || dialect === 'mariadb') {
+        it('should allow to use calculated values on duplicate', async function () {
+          const insertUser = await this.User.create({
+            bar: 10,
+          });
+          const [user] = await this.User.upsert({
+            id: insertUser.id,
+            bar: this.sequelize.literal('`bar` + 1'),
```
ephys, line 580: It would be interesting to test whether this works when there was no duplicate (if there isn't already a test for that).

src/dialects/abstract/query-generator.js
```diff
@@ -119,7 +119,7 @@ export class AbstractQueryGenerator {
     const bind = Object.create(null);
     const fields = [];
     const returningModelAttributes = [];
-    const values = [];
+    const values = {};
```
ephys, line 122:
```suggestion
    const values = Object.create(null);
```

ephys (changes requested):

ephys:
I can confirm that you can ignore db2, thank you for your hard work on this. I've finally found some time to review it!

test/unit/utils/sql.test.ts
```diff
@@ -73,14 +73,15 @@ describe('mapBindParameters', () => {
     expect(parameterSet).to.deep.eq(new Set(['a']));
   });
 
-  it(`does not consider the token to be a bind parameter if it does not follow '(', ',', '=' or whitespace`, () => {
```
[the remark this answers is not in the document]
elijahb: I think that regardless of how it's described, that's the test and the change that I did was necessary. I see the new test, but it doesn't test the same. Still there's a bug if the same token is used multiple times. The last time that the token is used, if it's followed by a `;` it doesn't get replaced. Try it by adding a `;` on that test as I did.

test/unit/dialects/mysql/query-generator.test.js
elijahb: Moved

WikiRik:

test/unit/utils/sql.test.ts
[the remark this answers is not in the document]
ephys: Yes but it's fixed now (https://github.com/sequelize/sequelize/blob/main/test/unit/utils/sql.test.ts#L57-L67), proof being that your PR doesn't modify the source code anymore but the modified test still passes

test/unit/query-interface/upsert.test.ts
```diff
@@ -204,4 +204,47 @@ describe('QueryInterface#upsert', () => {
...
+        counter: literal('`counter` + 1'),
+      },
+      {
+        counter: literal('`counter` + 1'),
+      },
+      // TODO: weird mssql/db2 specific behavior that should be unified
+      dialectName === 'mssql' || dialectName === 'db2' ? { id: 1 } : {},
+      {
+        model: User,
+        bind: ['John'],
```
ephys, line 224: this bind option is unused

ephys (approved):
That's my last comment, other than that this PR is good to go!

soknifedev:
is there any update on this?

WikiRik:
@ephys I've reverted the changes to the test so I think we should be good to merge this now. Could you please check?

batflarrow:
When will this be merged and released?

ephys:
Conflict but then I think we're ready to merge

WikiRik:
@ephys I've resolved the conflict

ephys (approved):
Thanks!

Issue #13817, opened 2021-12-23
Replacements break with unrelated user provided data


pedroassis:
## Issue Creation Checklist

[x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)
https://github.com/HandleCollections/sequelize-sscce
## Bug Description

When using replacements with user provided data we get errors when users input the colon symbol.

This happens because the user string is injected into the sql which later is parsed by sql-string.js file's `formatNamedParameters` function.

The `formatNamedParameters` does not have a way to know where the colon symbols comes from.

### SSCCE

**Here is the link to the SSCCE for this issue:** 
https://github.com/HandleCollections/sequelize-sscce

```js
  // lib/sql-string.js formatNamedParameters gets the sql with `WHERE name = "some:foo"`
  // it will try to replace the ":foo", even tho it is not supposed to get replaced
  const query = {
    where: {
      name: 'some:foo', // this is a user provided string I expect to be sanitized
    },
    replacements: {
      test: 'this is used somewhere', // not used in example for the sake of simplicity
    },
  };
```

### What do you expect to happen?

I want to be able to search for strings containing a colon symbol and use replacements at the same time.

I also expect user provided data to be sanitized and never be treated as anything other than text. 

### What is actually happening?

sequelize tries to parse user provided data and throws an error.

I think it might be an avenue for sql injection since sequelize is treating an user provided string as parseable text.

```
Error: Named parameter ":foo" has no value in the given object.
    at /Users/pedro/projects/sequelize-sscce/node_modules/sequelize/dist/lib/sql-string.js:103:11
```

### Environment

- Sequelize version: 6.12.2 and 4.4.4 
- Node.js version: v14.16.0 

### How does this problem relate to dialects?

- [x] I think this problem happens regardless of the dialect.

### Would you be willing to resolve this issue by submitting a Pull Request?

It seems to be a very difficult problem to solve correctly, meaning only replacing actual replacements, not just as string with a colon.

I also thought about an ignore replacement error, where it would not replace a string that doesn't match a replacement, but this would still allow for a soft "injection" if the user matched a existing replacement.

- [ ] Yes, I have the time and I know how to start.
- [x] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

pedroassis:
[...]

PR #14447, opened 2022-04-24
fix: fix `$bind` and `:repl` being replaced in user input

ephys:
This PR is still WIP but I'm opening it to describe the changes as I go.

Fixes in v7: #13817 

This (v7) PR fixes how bind & replacements are handled in query interface & findAll methods.

### Before this PR

Let's take the following query:

```typescript
User.findAll({
  where: and(
    literal('soundex("firstName") = soundex(:firstName)'),
    { lastName: lastName },
  ),
  replacements: { firstName },
})
```

Before this PR, Sequelize would first generate the SQl for the `where` parameter:

```sql
SELECT * FROM users WHERE soundex("firstName") = soundex(:firstName) AND "lastName" = 'Doe'
```

Then pass that query to `Sequelize#query`, which would process the `replacements` option (in this case, the `:firstName` parameter):

```sql
SELECT * FROM users WHERE soundex("firstName") = soundex('John') AND "lastName" = 'Doe'
```

---

Because of this two-step processing, a malicious input can result in SQL injection. For instance with these two user inputs:

```js
const lastName = ':firstName'; // <- notice how it includes the syntax for a replacement parameter
const firstName = '; DROP TABLE users;'
```

With the above input, `QueryGenerator` will first generate this:

```sql
SELECT * FROM users WHERE soundex("firstName") = soundex(:firstName) AND "lastName" = ':firstName'
```

Then `Sequelize#query` will process `:firstName`, resulting in this:

```sql
SELECT * FROM users WHERE soundex("firstName") = soundex('; DROP TABLE users;') AND "lastName" = ''; DROP TABLE users;''
```

Resulting in the injection of `DROP TABLE users;`, interpreted as actual SQL.

### After this PR

Replacements are now handled by `QueryGenerator` instead. This allows `QueryGenerator` to *only* process replacements in `literal()` and nowhere else.

`QueryGenerator` will generate this:

```sql
SELECT * FROM users WHERE soundex("firstName") = soundex('; DROP TABLE users;') AND "lastName" = ':firstName'
```

Then the query will be passed to a new method, `Sequelize#queryRaw`, which does *not* process replacements at all. So the query is executed as-is.

`Sequelize#query` will still process replacements, but as long as you don't put replacements inside strings there, and don't inline user input yourself, you'll be fine.

### Other changes

- The `bind` & `replacement` options can now be used at the same time! This is because handling of bind parameters is done by a new parser that knows whether `$name` is a bind parameter, or just text inside a string.
- It's now possible to use `bind` parameters in `update`, and `insert` methods. It was not previously possible because sequelize generates its own bind parameters in these methods, which would overwrite the ones provided by the user. They are now safely merged.

### TODO

- Test the new `bind` replacer against comments


ephys:
Crossing my fingers for the last test suite but i think this may be ready

ephys:
Breaking changes documented https://github.com/sequelize/website/pull/135

test/unit/query-generator/select-query.test.ts
```diff
@@ -0,0 +1,287 @@
...
+      ibmi: 'SELECT "id" FROM "Users" AS "User" OFFSET 1 ROWS',
+      mssql: `SELECT [id] FROM [Users] AS [User] ORDER BY [User].[id] OFFSET 1 ROWS;`,
+    });
+  });
+
+  describe('replacements', () => {
+    it('parses named replacements in literals', async () => {
+      // The goal of this test is to test that :replacements are parsed in literals in as many places as possible
+
+      const sql = queryGenerator.selectQuery(User.tableName, {
```
WikiRik, line 54: What a test, good work!

test/unit/query-generator/select-query.test.ts
```diff
@@ -0,0 +1,287 @@
...
+            INNER JOIN "Users" AS "contributors"
+            ON "contributors"."id" = "contributors->ProjectContributor"."UserId"
+            AND 'where'
+          )
+          ON "Project"."id" = "contributors->ProjectContributor"."ProjectId"
+        `,
+      });
+    });
+
+    // TODO: test subQueries
```
WikiRik, line 270: This is for a future PR, right?

test/unit/query-interface/decrement.test.ts
```diff
@@ -0,0 +1,48 @@
...
+          id: 2,
+          data: 3,
+        },
+      },
+    );
+
+    expect(stub.callCount).to.eq(1);
+    const firstCall = stub.getCall(0);
+    expectsql(firstCall.args[0] as string, {
+      postgres: `UPDATE "Users" SET "age"="age"- ':age',"name"=':name' WHERE "id" = ':id' RETURNING ":data"`,
```
WikiRik, line 42: Not sure how I can get a comment on multiple lines, but can you swap `postgres` here with `default`?

test/unit/query-interface/increment.test.ts
```diff
@@ -0,0 +1,48 @@
...
+          id: 2,
+          data: 3,
+        },
+      },
+    );
+
+    expect(stub.callCount).to.eq(1);
+    const firstCall = stub.getCall(0);
+    expectsql(firstCall.args[0] as string, {
+      postgres: `UPDATE "Users" SET "age"="age"+ ':age',"name"=':name' WHERE "id" = ':id' RETURNING ":data"`,
```
WikiRik, line 42: Same comment as above, can you swap `postgres` and `default`?

test/unit/query-interface/insert.test.ts
```diff
@@ -0,0 +1,103 @@
...
+      mssql: `INSERT INTO [Users] ([firstName]) OUTPUT INSERTED.[:data] VALUES ($sequelize_1);`,
+      db2: `SELECT * FROM FINAL TABLE (INSERT INTO "Users" ("firstName") VALUES ($sequelize_1));`,
+      ibmi: `SELECT * FROM FINAL TABLE (INSERT INTO "Users" ("firstName") VALUES ($sequelize_1))`,
+    });
+    expect(firstCall.args[1]?.bind).to.deep.eq({
+      sequelize_1: 'Zoe',
+    });
+  });
+
+  it('throws if a bind parameter name starts with the reserved "sequelize_" prefix', async () => {
```
WikiRik, line 42: Now that I think about this, why is this test only present in some files?

test/integration/sequelize/query.test.js
```diff
@@ -807,29 +733,33 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
-        const expected = ['db2', 'ibmi'].includes(dialect) ? [{ FOO: 1, BAR: '$ / $1' }] : [{ foo: 1, bar: '$ / $1' }];
+        const expected = ['db2', 'ibmi'].includes(dialect) ? [{ FOO: 1, BAR: '$$ / $$1' }] : [{ foo: 1, bar: '$$ / $$1' }];
         expect(result[0]).to.deep.equal(expected);
         if (['postgres', 'sqlite', 'db2', 'ibmi'].includes(dialect)) {
           expect(logSql).to.include('$1');
         }
       });
 
-      it('binds named parameters object handles escaped $$', async function () {
+      // this was a legacy band aid that has since been removed, because the underlying issue (transforming bind params in strings) has been fixed.
```
WikiRik, line 754: Like above, do we need this comment? I get that it's useful for the review but doesn't need to stay here right?

test/integration/sequelize/query.test.js
```diff
@@ -807,29 +733,33 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
         expect(result[0]).to.deep.equal(expected);
       });
 
-      it('binds named parameters array handles escaped $$', async function () {
+      // this was a legacy band aid that has since been removed, because the underlying issue (transforming bind params in strings) has been fixed.
```
WikiRik, line 736: Do we need this comment?

WikiRik (commented):
Finished looking at the tests, do have a few comments but looks great! Would be nice if we could clear up most of the TODOs before releasing v7

test/integration/model.test.js
```diff
@@ -1054,11 +1056,18 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
-          if (['mssql', 'ibmi'].includes(dialect)) {
-            expect(sql).to.not.contain('createdAt');
-          } else {
-            expect(sql).to.match(/UPDATE\s+["`]+User1s["`]+\s+SET\s+["`]+secretValue["`]=(\$1|\?),["`]+updatedAt["`]+=(\$2|\?)\s+WHERE ["`]+id["`]+\s=\s(\$3|\?)/);
-          }
+
+          expect(sql).to.match(/^Executing \(default\): /);
+          sql = sql.slice(21);
+
+          expectsql(sql, {
```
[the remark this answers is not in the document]
ephys: It's probably similar to db2, but I don't really know for sure as I can't run this code against ibmi

test/integration/sequelize/query.test.js
ephys: It's there to explain why a test that does seemingly nothing exists as not transforming the contents of strings seems like the logical thing to do (at least to me) and someone may be tempted to remove it, but I kept the test because we used to have a weird behavior there

test/unit/query-interface/insert.test.ts
ephys: They're the only methods reserving some bind parameter names, because they're the only methods that add extra bind parameters

test/unit/query-interface/decrement.test.ts
ephys: click on one line, then shift+click on another one to comment on multiple lines :)

test/unit/query-generator/select-query.test.ts
ephys: Ideally no, I should add these here

test/unit/query-generator/bulk-insert-query.test.ts
```diff
@@ -0,0 +1,27 @@
+import { DataTypes, literal } from '@sequelize/core';
+import { expectsql, sequelize } from '../../support';
+
+describe('QueryGenerator#bulkInsertQuery', () => {
+  const queryGenerator = sequelize.getQueryInterface().queryGenerator;
+
+  const User = sequelize.define('User', {
+    firstName: DataTypes.STRING,
+  }, { timestamps: false });
+
```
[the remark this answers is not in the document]
ephys: Yep, would be good to add these in each query-interface file if the query is supposed to support replacements

test/unit/query-generator/bulk-insert-query.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    }], {
+      replacements: {
+        injection: 'a string',
+      },
+    });
+
+    expectsql(sql, {
+      default: `INSERT INTO [Users] ([firstName]) VALUES ('a string');`,
+      mssql: `INSERT INTO [Users] ([firstName]) VALUES (N'a string');`,
+      // TODO: db2 should use the same system as ibmi
```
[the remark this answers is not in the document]
ephys: I may just remove the comment. I think `ibmi` does something different 'incorrectly'. I think its query is written like this to emulate 'RETURNING', but the option is not enabled, so `ibmi` may be wrong here

src/dialects/abstract/query-generator.js
```diff
@@ -2254,7 +2268,20 @@ export class AbstractQueryGenerator {
     }
 
     if (smth instanceof Utils.Literal) {
-      return smth.val;
+      if (includesPositionalReplacements(smth.val)) {
+        throw new TypeError(`The following literal includes positional replacements (?). Only named replacements (:name) are allowed in literal() because we cannot guarantee the order in which they will be evaluated:
```
WikiRik, line 2272: Can we format this so that it's more readable?

WikiRik: As in; that it doesn't take up the entire screen

src/dialects/abstract/query-interface.js
```diff
@@ -1066,10 +1119,14 @@ export class QueryInterface {
 
     const dataType = options.dataType;
 
+    // FIXME: DECIMAL is not safely representable as a float!
```
WikiRik, line 1122: This is for a next PR right? And why did you use FIXME here instead of TODO?

WikiRik (commented):
Few additional comments, but again; it looks great. Thanks for your hard work on this!

test/integration/model.test.js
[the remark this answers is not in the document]
WikiRik: So we should add a TODO to check this, right?

test/integration/sequelize/query.test.js
WikiRik: Ah yes, that's a good reason

test/unit/utils/utils.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      it('can take a value to compare against', () => {
+        const path = 'metadata.pg_rating.is';
+        const value = 'U';
+        expect(queryGenerator.handleSequelizeMethod(new Utils.Json(path, value))).to.equal('("metadata"#>>\'{pg_rating,is}\') = \'U\'');
+      });
+    });
+  }
+
+  describe('inflection', () => {
+    it('works better than lingo ;)', () => {
```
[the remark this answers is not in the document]
fzn0x: Haha this one has a funny mesage

src/dialects/mssql/index.js
```diff
@@ -49,6 +51,11 @@ export class MssqlDialect extends AbstractDialect {
       this.queryGenerator,
     );
   }
+
+  createBindCollector() {
+    // unlike postgres, sqlite supports $named bind parameters, but it wasn't deemed worth it to implement it.
```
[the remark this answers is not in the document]
ephys: Not only is it mssql, but I ended up implementing a collector that handles named parameters for mssql & sqlite anyway

src/dialects/mssql/query-interface.js
```diff
@@ -84,6 +90,8 @@ export class MsSqlQueryInterface extends QueryInterface {
 
     const sql = this.queryGenerator.upsertQuery(tableName, insertValues, updateValues, where, model, options);
 
-    return await this.sequelize.query(sql, options);
+    delete options.replacements;
```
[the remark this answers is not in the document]
ephys: `queryRaw` will throw if `options.replacements` is specified in case someone tries to use it like `query`
I delete it before calling queryRaw to prevent it from throwing. I thought I documented it each time but it's missing here

src/sequelize.d.ts
```diff
@@ -1265,6 +1366,33 @@ export class Sequelize extends Hooks {
   public query(sql: string | { query: string; values: unknown[] }, options: (QueryOptions | QueryOptionsWithType<QueryTypes.RAW>) & { plain: true }): Promise<{ [key: string]: unknown } | null>;
   public query(sql: string | { query: string; values: unknown[] }, options?: QueryOptions | QueryOptionsWithType<QueryTypes.RAW>): Promise<[unknown[], unknown]>;
 
+  /**
+   * Works like {@link Sequelize#query}, but does not inline replacements. Only bind parameters are supported.
+   *
+   * @param sql The SQL to execute
+   * @param options The options for the query. See {@link QueryRawOptions} for details.
+   */
+  public queryRaw(sql: string | { query: string; values: unknown[] }, options: QueryRawOptionsWithType<QueryTypes.UPDATE>): Promise<[undefined, number]>;
```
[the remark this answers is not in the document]
ephys: They're basically the same as `query()`, but they do not accept the `replacements` option

src/sequelize.js
```diff
@@ -534,10 +536,70 @@ export class Sequelize {
...
+  }
+
+  async queryRaw(sql, options) {
+    if (typeof sql !== 'string') {
+      throw new TypeError('Sequelize#rawQuery requires a string as the first parameter.');
+    }
+
+    if (options != null && 'replacements' in options) {
+      throw new TypeError(`Sequelize#rawQuery does not accept the "replacements" options.
+Only bind parameters can be provided, in the dialect-specific syntax.
```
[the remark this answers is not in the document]
ephys: It's on purpose, if I indent the string, it'll end up having a lot of whitespace. This is something https://github.com/tc39/proposal-string-dedent will improve :D

src/dialects/abstract/query-interface.js
ephys: For another PR yep, it's just a thing I noticed while I was working on that file
I'll change it to TODO for consistency. I sometimes use `FIXME` when it's a bug but it's probably a bad idea, `TODO` is universally recognized, `FIXME` isn't

src/sequelize.js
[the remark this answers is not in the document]
WikiRik: Ah yes

test/integration/model.test.js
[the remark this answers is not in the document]
ephys: Yep!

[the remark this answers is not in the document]
ephys: its output is likely to be `UPDATE "users1" SET "secretValue"=?,"updatedAt"=? WHERE "id" = ?;`, I'll add that

test/unit/query-generator/bulk-insert-query.test.ts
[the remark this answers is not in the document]
ephys: Oh, this is already a query-generator test actually

test/unit/utils/utils.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      expect(Utils.pluralize('status')).to.equal('statuses');
+
+      expect(Utils.singularize('status')).to.equal('status');
+    });
+  });
+
+  describe('flattenObjectDeep', () => {
+    it('should return the value if it is not an object', () => {
+      const value = 'non-object';
+      // @ts-expect-error
```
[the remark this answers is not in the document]
ephys: I've changed the signature of `flattenObjectDeep` to accept any input since that's what it does

test/unit/dialects/abstract/query-generator.test.js
```diff
@@ -122,10 +123,15 @@ describe('QueryGenerator', () => {
         .should.be.equal('foo IS NOT NULL');
     });
 
-    it('should correctly escape $ in sequelize.fn arguments', function () {
+    // this was a band-aid over a deeper problem ('$bind' being considered to be a bind parameter when it's a string), which has been fixed
```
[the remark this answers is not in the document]
ephys: I prefer to keep it to explain why there is a test that checks that something is not done

test/unit/query-interface/decrement.test.ts
ephys: click one line, then shift+click another to comment on multiple lines :)

test/unit/query-interface/insert.test.ts
ephys: Only insert, update, and bulkUpdate add extra bind parameters in their SQL, so they're the only 3 methods that reserve bind parameters

ephys:
Changes implemented. I just need to add this test now:

> // TODO: test subQueries

src/dialects/abstract/query-generator.js
```diff
@@ -469,9 +471,13 @@ export class AbstractQueryGenerator {
    * @private
    */
   arithmeticQuery(operator, tableName, where, incrementAmountsByField, extraAttributesToBeUpdated, options) {
+    // TODO: this method should delegate to `updateQuery`
```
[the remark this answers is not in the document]
WikiRik: No worries, we can leave it like this

WikiRik (dismissed):
New test looks good and other comments have also been addressed so I think this is good to go

ephys:
There is one last fix I want to add before merging.
Now that we have a string-aware SQL "parser", we can prevent replacements from being transformed in literals too `literal('= \':value\'')` should not consider `:value` to be a replacement

test/unit/utils/sql.test.ts
```diff
@@ -264,3 +264,327 @@ describe('mapBindParameters', () => {
...
+    }).to.throw('Named replacement ":toString" has no entry in the replacement map.');
+
+  });
+
+  it('parses named replacements followed by cast syntax', () => {
+    const sql = injectReplacements(`SELECT * FROM users WHERE id = :id::string`, dialect, {
+      id: 1,
+    });
+
+    // this is postgres-only syntax. It doesn't work in other dialects.
```
WikiRik, line 293: Do we want to run this test on all dialects then?

test/unit/utils/sql.test.ts
```diff
@@ -264,3 +264,327 @@ describe('mapBindParameters', () => {
...
 });
+
+describe('injectReplacements (named replacements)', () => {
+  it('parses named replacements', () => {
+    const sql = injectReplacements(`SELECT ${dialect.TICK_CHAR_LEFT}:id${dialect.TICK_CHAR_RIGHT} FROM users WHERE id = ':id' OR id = :id OR id = ''':id'''`, dialect, {
+      id: 1,
+    });
+
+    expectsql(sql, {
+      default: `SELECT [:id] FROM users WHERE id = ':id' OR id = 1 OR id = ''':id'''`,
```
WikiRik, line 275: Why are `dialect.TICK_CHAR_LEFT` and `dialect.TICK_CHAR_RIGHT` not needed in the result?

test/unit/utils/sql.test.ts
```diff
@@ -264,3 +264,327 @@ describe('mapBindParameters', () => {
...
+      a: 1,
+    });
+
+    expectsql(sql, {
+      default: `SELECT * FROM users WHERE id = 1`,
+    });
+  });
+
+  it(`does not consider the token to be a replacement if it does not follow '(', ',', '=' or whitespace`, () => {
+    const sql = injectReplacements(`SELECT * FROM users WHERE id = fn(:id) OR id = fn('a',:id) OR id=:id`, dialect, {
```
WikiRik, line 310: Do we want to add `OR id = :id` here because the test name mentions whitespace?

WikiRik (commented):
In general looks good, I have three comments about the tests. Same comments also apply to their respective equivalent for positional replacements

test/unit/utils/sql.test.ts
```diff
@@ -0,0 +1,590 @@
...
+        SELECT * FROM users /*
+        WHERE id = ?
+        */
+        WHERE id = 1
+      `,
+    });
+  });
+
+  // https://github.com/sequelize/sequelize/issues/14358
+  it('does not parse ?& and ?| operators as replacements (#14358)', async () => {
```
WikiRik, line 551: Do we want to test this one on all dialects even though it's postgres exclusive?

test/unit/utils/sql.test.ts
ephys: `default` automatically converts `[` and `]` into the right tick for the dialect being tested (except in the case of `ARRAY[x]`)

test/unit/utils/sql.test.ts
ephys: Ah yes

test/unit/utils/sql.test.ts
ephys: Yep! I'd prefer it to be consistent across all dialects, it could be added to other dialects in the future if they start supporting arrays (last I heard arrays were added to the SQL spec)

test/unit/utils/sql.test.ts
WikiRik: Ah, that makes sense. Thanks!

test/unit/utils/sql.test.ts
ephys: Done

test/unit/utils/sql.test.ts
ephys: Turns out more dialects support it. It's not unlikely that it may get adopted by even more some day, might as well make sure we parse it properly

WikiRik (approved):
Thanks again for your thorough work on this!

ephys:
Now onto v6

PR #14453, opened 2022-04-26
fix: resolve various db2 issues

ephys:
This PR (hopefully) fixes the underlying issue that caused DB2 tests to be flaky: `Db2Query` was silencing errors emitted by DB2 in an attempt to fix them automatically.

This PR removes that. It appears the flakiness is gone, and this PR fixes the issues that were hidden by that behavior.

List of fixed issues:
- DB2's `Db2Query` muted errors. This has been removed.
- DB2's `createTable` used `CREATE TABLE` instead of `CREATE TABLE IF NOT EXISTS` (like the other dialects did)
- DB2's `dropTable` used `DROP TABLE` instead of `DROP TABLE IF EXISTS` (like the other dialects did)

The rest of the changes are adapting the tests to make them work properly with db2.

---

There is one change we may want to preserve:
> in db2, an "ALTER TABLE" query will cause the table to be put in "reorg pending state"
> and subsequent "ALTER TABLE" queries on said table will fail with code 7.
> The user either needs to wait for the reorg to happen, or `CALL SYSPROC.ADMIN_CMD('REORG TABLE tablename')` must be executed.

This `REORG TABLE` operation was done in `Db2Query`. Do we want to preserve it?

---

- [x] Research what to do with `REORG TABLE`
- [x] Research what to do with `ERRORSCHEMA.ERRORTABLE`
- [x] (test suite) only run `CREATE TABLESPACE SYSTOOLSPACE` if the tablespace does not exist


ephys:
It takes much, much longer to run locally actually.
Looks like only running a few tests doesn't fail, locally or on the CI (to double check).
I'm running the whole suite locally to see if anything fails, but it'll take a while.

Something I want to check is: does it ever fail if only one job is using the db2 database (including the other PRs)

WikiRik:
So you think that maybe the two jobs (Node 14 and 16) are still linked to each other? That would be weird, but we could try to define different database names and credentials to see if we won't have failing tests that way

Hopefully we can find something

ephys:
That would be really weird. I'm hoping I'll get it to fail locally so I can rule that out

ephys:
Well, it did not fail locally and I've been running it a few times on the CI without it failing there either :/

ephys:
Db2 database names cannot be longer than 8 characters 😨

ephys:
I noticed DB2 wasn't deleting the schemas, but saying it succeeded. Turns out [DB2 mutes a lot of error responses](https://github.com/sequelize/sequelize/blob/main/src/dialects/db2/query.js#L204).

ephys:
I'm not seeing the flaky tests anymore after removing the error muting and fixing some of the errors it was hiding. More errors to fix but this may actually be good news.

ephys:
I have constant 28 failing tests for every rerun. I'm hopeful that if I can manage to fix them cleanly, we won't have flaky tests anymore 🤞🏼

mike-usa:
> I have the same 28 tests failing consistently in every rerun. I'm hopeful that if I can manage to fix them cleanly, we won't have flaky tests anymore (because the flaky tests are not listed in the 28 tests) 🤞🏼

Annnd you can use failing tests as a vetting process 😈

I don't know if you implemented this, but I think I have all my tests passing locally now (Node v14) -- though I don't test against Snowflake or DBI.

I'm updating the existing pull request (#14578), though this is also a big commit. It includes dev updates and surely there's going to be opinions about something I did (or stupidity in something I was ignorant to) that will need to be changed.

ephys:
> though I don't test against `Snowflake` or `ibmi`.

Neither do we, to be fully honest, as we don't have a database we can test against :/

ephys:
I think the resolution of DB2's flakiness is complete (with a pending question, see top post). I've cleaned up this test PR into a proper PR.

test/integration/sequelize/query.test.js
```diff
@@ -266,22 +266,28 @@ describe(Support.getTestDialectTeaser('Sequelize'), () => {
...
+
+        await sequelize.query(this.insertQuery);
+
+        try {
+          await sequelize.query('DROP PROCEDURE foo');
+        } catch (error) {
+          // DB2 does not support DROP PROCEDURE IF EXISTS
+          // -204 means "FOO" does not exist
+          // https://www.ibm.com/docs/en/db2-for-zos/11?topic=sec-204
+          if (error.cause.sqlcode !== -204) {
```
sdepold, line 280: Should we do assert(error.cause.sqlcode===204) ?

test/integration/model.test.js
```diff
@@ -2172,7 +2172,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       const expectedLengths = {
         mssql: 2,
         postgres: 2,
-        db2: 10,
+        db2: 2,
```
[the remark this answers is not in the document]
ephys: `dropTestSchemas` was broken so previous test schemas were not being purged. You'd get 8 extra schemas created by previous tests that were never deleted.

After this PR, only `[ 'special ', 'schema_test' ]` are returned because all other (non technical) schemas are properly deleted (`special` and `schema_test` are created by this test's `beforeEach`)

test/integration/sequelize/query.test.js
ephys: No because the goal is to mute sql error -204 to emulate `DROP IF EXISTS` (as it means there is nothing to drop), but other errors should still surface and asserting here would shadow them

src/dialects/db2/query-interface.js
```diff
@@ -84,6 +84,13 @@ export class Db2QueryInterface extends QueryInterface {
     return [result, undefined];
   }
 
+  async dropSchema(schema, options) {
+    // error from dropSchema will be stored in this table.
+    await this.dropTable({ tableName: 'ERRORTABLE', schema: 'ERRORSCHEMA' });
```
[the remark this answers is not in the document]
ephys: Yes, it sounds like a table we might make more use of in the future

src/dialects/db2/query-generator.js
```diff
@@ -185,15 +185,6 @@ export class Db2QueryGenerator extends AbstractQueryGenerator {
     return 'SELECT TABNAME AS "tableName", TRIM(TABSCHEMA) AS "tableSchema" FROM SYSCAT.TABLES WHERE TABSCHEMA = USER AND TYPE = \'T\' ORDER BY TABSCHEMA, TABNAME';
   }
 
-  dropTableQuery(tableName) {
```
WikiRik, line 188: Why do we remove this here? Is the Abstract function for this the same now?

test/integration/query-interface/changeColumn.test.js
```diff
@@ -9,12 +9,9 @@ const { DataTypes } = require('@sequelize/core');
 const dialect = Support.getTestDialect();
 
 describe(Support.getTestDialectTeaser('QueryInterface'), () => {
-  beforeEach(function () {
+  beforeEach(async function () {
     this.sequelize.options.quoteIdenifiers = true;
     this.queryInterface = this.sequelize.getQueryInterface();
-  });
-
-  afterEach(async function () {
```
WikiRik, line 17: What happens after the last test of this suite? Will the test schemas still be dropped? Or could this cause a flaky test as well?

test/support.ts
```diff
@@ -228,7 +228,7 @@ export async function dropTestSchemas(sequelize: Sequelize) {
     }
   }
 
-  await Promise.all(schemasPromise.map(async p => p.catch((error: unknown) => error)));
```
WikiRik, line 231: Is this change releated to DB2?

WikiRik:
I believe this is also related to https://github.com/sequelize/cli/issues/1039

src/dialects/db2/query-generator.js
ephys: Yep! It was overwritten to remove 'IF EXISTS' but our minimum supported db2 version does support `IF EXISTS` here

test/integration/query-interface/changeColumn.test.js
ephys: `beforeEach` from `integration/support.js` drops all schemas

test/support.ts
ephys: Strictly speaking no, but it was muting errors and contributing to the flakiness

test/integration/query-interface/changeColumn.test.js
WikiRik: Ah yes

src/dialects/db2/query-interface.js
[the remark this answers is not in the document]
WikiRik: Do you want to move that in this PR?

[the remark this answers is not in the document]
ephys: Yep!

ephys:
> I believe this is also related to https://github.com/sequelize/cli/issues/1039

It very well could, it would be interesting to start a branch of the cli that works for v7 I think

ephys:
Before merging this PR, we still need to decide on what to do with `ERRORTABLE` and `REORG TABLE`:

- Do we execute `REORG TABLE` automatically when an `ALTER` query fails because the table is in `REORG PENDING STATE`, or do we leave it to the user?
- Is it safe to drop `ERRORSCHEMA.ERRORTABLE` automatically before all `DROP SCHEMA` queries?

WikiRik:
> * Do we execute `REORG TABLE` automatically when an `ALTER` query fails because the table is in `REORG PENDING STATE` (then re-execute the `ALTER`), or do we leave it to the user?

I think it would be better to do this automatically since that will be more unified with the other dialects. What we might want to do in a later PR is making this optional through the config so users can decide if they want sequelize to do it for them or that they want to do it themselves.

> * Is it safe to drop `ERRORSCHEMA.ERRORTABLE` automatically before all `DROP SCHEMA` queries? What does DB2 recommend here (it's such a weird way of emitting errors)

This I would not know. I think it should be safe for now and we should look into it later

ephys:
I've changed how deleting `ERRORSCHEMA.ERRORTABLE` is handled. It's handled by `dropSchema` if db2 signals that it has been created. It also uses a serial number in the table name to prevent to concurrent calls to `dropSchema` from using the same error table.

src/dialects/db2/query-interface.js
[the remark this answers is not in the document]
ephys: No need to move it after all, it's not hardcoded anymore (it's now dynamic to reduce the risk of concurrency issues)

WikiRik:
@ephys can you check out the failed tests? I think they are due to this PR

ephys:
Hmm that's annoying. `ADMIN_DROP_SCHEMA` is failing on that test with error

> -911 THE CURRENT UNIT OF WORK HAS BEEN ROLLED BACK DUE TO DEADLOCK OR TIMEOUT. REASON reason-code, TYPE OF RESOURCE resource-type, AND RESOURCE NAME resource-name

https://www.ibm.com/docs/en/db2-for-zos/11?topic=codes-911

src/dialects/db2/query-interface.js
```diff
@@ -84,6 +85,61 @@ export class Db2QueryInterface extends QueryInterface {
...
+    const outParams = new Map();
+
+    // DROP SCHEMA works in a weird way in DB2:
+    // Its query uses ADMIN_DROP_SCHEMA, which stores the error message in a table
+    // specified by two IN-OUT parameters.
+    // If the returned values for these parameters is not null, then an error occurred.
+    const response = await super.dropSchema(schema, {
+      ...options,
+      // db2 supports out parameters. We don't have a proper API for it yet
+      // so this temporary API will have to do.
```
WikiRik, line 98: Should we make this a TODO?

src/dialects/db2/query-interface.js
```diff
@@ -122,4 +178,33 @@ export class Db2QueryInterface extends QueryInterface {
...
+
+  /**
+   * DB2 can put tables in the "reorg pending" state after a structure change (e.g. ALTER)
+   * Other changes cannot be done to these tables until the reorg has been completed.
+   *
+   * This method forces a reorg to happen now.
+   */
+  async executeTableReorg() {
+    // https://www.ibm.com/support/pages/sql0668n-operating-not-allowed-reason-code-7-seen-when-querying-or-viewing-table-db2-warehouse-cloud-and-db2-cloud
+    return await this.sequelize.query(`CALL SYSPROC.ADMIN_CMD('REORG TABLE "users"')`);
```
WikiRik, line 208: Do we only want to do this on the users table? Might be nice if we could pass the schema and table to this function to use it more often

src/dialects/db2/query-interface.js
```diff
@@ -122,4 +178,35 @@ export class Db2QueryInterface extends QueryInterface {
...
+   * DB2 can put tables in the "reorg pending" state after a structure change (e.g. ALTER)
+   * Other changes cannot be done to these tables until the reorg has been completed.
+   *
+   * This method forces a reorg to happen now.
+   *
+   * @param {TableName} tableName - The name of the table to reorg
+   */
+  async executeTableReorg(tableName) {
+    // https://www.ibm.com/support/pages/sql0668n-operating-not-allowed-reason-code-7-seen-when-querying-or-viewing-table-db2-warehouse-cloud-and-db2-cloud
+    return await this.sequelize.query(`CALL SYSPROC.ADMIN_CMD('REORG TABLE ${this.queryGenerator.quoteTable(tableName)}')`);
```
WikiRik, line 210: Reorg also supports schemas. Do we want to support that as well or is that for a later PR?

WikiRik (approved):

src/dialects/db2/query-interface.js
ephys: `quoteTable` handles the schema too, `tableName` can be `{ "tableName": "users", "schema": "public" }`

ephys: Update: `addConstraint` crashes when a schema is specified. It tries to execute

```sql
LTER TABLE "archive"."users" ADD CONSTRAINT "[object Object]_id_pk" PRIMARY KEY ("id")
```

This bug is unrelated to this PR. I've added a disabled test we can sort out in a separate PR

WikiRik (approved):

PR #14466, opened 2022-04-28
build(deps): remove acorn

WikiRik:
Closes #14452


ephys:
Looking into the commit history, it looks like it was (or is) a peer dependency. But peer dependencies are installed automatically nowadays iirc

WikiRik:
I think it was a peer dependency since they should not be automatically installed

Issue #14174, opened 2022-02-25
MariaDB 10.5.15 breaks a test


WikiRik:
## Issue Creation Checklist

[X] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Issue Description

MariaDB 10.5.15 breaks [one test](https://github.com/sequelize/sequelize/blob/main/test/integration/transaction.test.js#L541-L660)

```
[MARIADB] Transaction
       deadlock handling
         should release the connection for a deadlocked transaction (2/2):

      AssertionError: expected false to be true
      + expected - actual

      -false
      +true
```

### Additional context

- [Dockerhub for MariaDB](https://hub.docker.com/_/mariadb)
- [Code for the Docker image](https://github.com/MariaDB/mariadb-docker/tree/master/10.5)

### Is this issue dialect-specific?

- [ ] No. This issue is relevant to Sequelize as a whole.
- [X] Yes. This issue only applies to the following dialect(s): MariaDB 10.5.15
- [ ] I don't know.

### Would you be willing to resolve this issue by submitting a Pull Request?

- [X] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

WikiRik:
Fix for the short term is included in #14171 by pinning the patch version of MariaDB. Renovate will do automatic updates on this so it is not an issue.

WikiRik:
@ephys I only updated it for the main branch. This is also the failing test on v6. Renovate does not do automatic updates on v6 so we can either disable the test for now or fix it. But I'm not sure when I can get to fixing it

ephys:
I made a branch to fix it a little while ago but changed my mind and opted to fix the test instead
Now that a v6 release is blocked by it I should prioritize it 😅

WikiRik:
@ephys would be nice if you could take another look, otherwise I can check it out next week

ephys:
The test fails because it tries to cause a deadlock but fails. MariaDB may have changed their heuristics to reduce the number of deadlocks:

Here is the query execution order: (ONE & TWO are the connection IDs of each transaction)

```
Executing (ONE): START TRANSACTION;
[...]

PR #14472, opened 2022-04-29
fix: do not replace `:replacements` inside of strings

### LINKED ISSUE #13817: Replacements break with unrelated user provided data

ephys:
### Pull Request Checklist

### Description Of Change

This PR backports the non-breaking changes from https://github.com/sequelize/sequelize/pull/14447 in v6

Only replacements have been backported, as unfortunately an escape hack was put in place to handle the bind parameter scenario, and reverting it would be a breaking change.

Closes #13817


fzn0x (approved):
We are good to go

WikiRik (approved):
Looks like nothing weird got backported

ephys:
I just hope this doesn't introduce a breaking change, but in theory it will only be a breaking change if the feature was used 'incorrectly' (used a replacement in a place where a replacement should not be treated as one, like inside of a string).

ephys:
I will skip the mysql 8 tests: It would seem we are experiencing the same issue as we did with mariadb https://github.com/sequelize/sequelize/issues/14174

WikiRik:
This is the cause of #14509 
Not sure why `Generate docs` check passes while there is clearly a SyntaxError for it. @ephys would you be able to check it out?

src/utils.js
```diff
@@ -114,13 +114,26 @@ function pluralize(str) {
 }
 exports.pluralize = pluralize;
 
+/**
+ * @deprecated use {@link injectReplacements} instead. This method has been removed in v7.
+ *
+ * @param {[string, ...unknown[]]} arr - first item is the SQL, following items are the positional replacements.
```
WikiRik, line 120: @ephys I assume the issue is with this line
```
SyntaxError: Invalid regular expression: /[~][string, ..$/: Unterminated character class
```

ephys: That's most likely it, and the docs check did not catch it

ephys: PR ready https://github.com/sequelize/sequelize/pull/14513

evangow:
This just broke my deployment, so it's a breaking change for me at the very least.

"in theory it will only be a breaking change if the feature was used 'incorrectly' (used a replacement in a place where a replacement should not be treated as one, like inside of a string)"

Not sure what you mean by replacements being used incorrectly inside a string.

The docs do this:

```
await sequelize.query(
  'SELECT * FROM projects WHERE status = :status',
  {
    replacements: { status: 'active' },
    type: QueryTypes.SELECT
  }
);
```
https://sequelize.org/docs/v6/core-concepts/raw-queries/

peterbaricic:
Please, if you break things in a minor release update then please provide an example how to use replacements with raw queries. Why is it incorrect usage?

ephys:
By 'using a replacement inside a string' i meant a sql string, something like this;

```js
await sequelize.query(
  `SELECT * FROM projects WHERE status = ':status'`,
  {
    replacements: { status: 'active' },
    type: QueryTypes.SELECT
  }
);
```

Note the quotes around `:status`. Injecting a replacement there would have resulted in sql injection. That's the thing we changed.

Your usage is correct and should not have broken. We'll look into what happened there.

peterbaricic:
this is my usage, I use it to call sql functions, I don't have any strings there

```ts
const res: bulkItem[] = await database.query(`SELECT "portfolioInvestmentId", value FROM ${metric}(:fundId, array[:pciIds]::uuid[], :untilDate, :excludeEventId)`, {
      type: QueryTypes.SELECT,
      replacements: {
        fundId,
        pciIds,
        untilDate: dateToStr(untilDate),
        excludeEventId,
      },
```

ephys:
This results in 

```sql
SELECT "portfolioInvestmentId", value FROM table('abc', array[:pciIds]::uuid[], 'ijk', 'lmn')
```

Looks like the culprit is `array[:pciIds]::uuid[]`. Technically it's not a valid spot for a SQL value, but v6 doesn't have a way to specify the type of a replacement and this is breaking a legitimate workaround to that so we'll fix the parser in v6 to allow it.

ephys:
PR: https://github.com/sequelize/sequelize/pull/14518

evangow:
I had 2 cases where using replacements broke. The first was with using it with LIMIT doing something like the below

```javascript
await sequelize.query(
  'SELECT * FROM projects WHERE status = :status LIMIT :first',
  {
    replacements: { status: 'active', first: 20 },
    type: QueryTypes.SELECT
  }
);
```

The 2nd case was with using it with an array as well, such as `WHERE "tags" @> ARRAY[:tags]::TEXT[]`. Seems like the PR you submitted should fix this 2nd case but not the first

ephys:
I just ran it against our latest v6 and it resulted in `SELECT * FROM projects WHERE status = 'active' LIMIT 20`

There may be something more going on. If you'd rather avoid posting the real query in public, you can send it to me privately on our [slack](http://sequelize-slack.herokuapp.com/) (if you're comfortable with that), and I'll look into it.

nabilzhafri:
@ephys i see, if thats the concern that make sense to patch such thing. In my case, it's all constructed internally, so it wasn't really a problem.

And thanks for the solution! That'll work.
In my case I have to go with second solution because my queries are all in .sql files

Caerbannog:
Here are two more cases where the replacement is not detected, with PostgreSQL.

1. JSONB indexing
```sql
select ('{"abc": 1}'::jsonb)->>:key;
```

2. Inside a `do` block. I assume that `$$` triggers the "inside of strings" limitation.
```sql
do $$
  begin
    if :value then
     refresh materialized view abc;
   end if;
  end
$$;
select 1;
```

We worked around these cases by concatenating strings, and taking care of sanitization ourselves.

WikiRik:

src/sql-string.d.ts
```diff
@@ -1,5 +1,5 @@
 export type Escapable = undefined | null | boolean | number | string | Date;
 export function escapeId(val: string, forbidQualified?: boolean): string;
-export function escape(val: Escapable | Escapable[], timeZone?: string, dialect?: string, format?: string): string;
+export function escape(val: Escapable | Escapable[], timeZone?: string, dialect?: string, format?: boolean): string;
 export function format(sql: string, values: unknown[], timeZone?: string, dialect?: string): string;
 export function formatNamedParameters(sql: string, values: unknown[], timeZone?: string, dialect?: string): string;
```
benasher44, line 5: It looks like this wasn't updated, so existing TS users won't see the new injectReplacements function, nor will they see the deprecation.

ephys:
I haven't been able to resolve the `DO` and `EXECUTE` use cases.

The problem is that I cannot guarantee that the code injected inside of the string is safe. It requires a much more complex escape function.

In your example, if I set `value` to `'$$'`, the result will be

```sql
do $$
  begin
    if '$$' then
      refresh materialized view abc;
    end if;
  end
$$;
```

Which is a syntax error because postgres considers `$$` to be the end of the string passed to the do block

This limitation wasn't introduced by this PR, so that code was vulnerable before this PR, but I can't enable it until we've figured out a way to safely escape the value that needs to be injected inside of the string.

vivekkairi:
Hi @ephys 

Screenshot 2022-09-26 at 6 34 18 PM

With earlier sequelize versions, passing `undefined` as positional parameter was consider as `null` but now an error is been thrown resulting in a breaking change.

src/utils/sql.ts
```diff
@@ -0,0 +1,221 @@
...
+      const previousChar = sqlString[i - 1];
+      // we want to be conservative with what we consider to be a replacement to avoid risk of conflict with potential operators
+      // users need to add a space before the bind parameter (except after '(', ',', and '=')
+      if (previousChar !== undefined && !/[\s(,=]/.test(previousChar)) {
+        continue;
+      }
+
+      const remainingString = sqlString.slice(i, sqlString.length);
+
+      const match = remainingString.match(/^:(?<name>[a-z_][0-9a-z_]*)(?:\)|,|$|\s|::)/i);
```
hbrysiewicz, line 144: This regex does not take into account an ending `;` on the statement, so I corrected locally to `remainingString.match(/^:(?<name>[a-z_][0-9a-z_]*)(?:\)|,|\;|$|\s|::)/i);` and then also discovered it does not like when the sql statement has JSON in it such as: `UPDATE table SET value = value || '{"column::form": :formTemplateId}' where key = 'formTemplates.defaultForms'` (with replacements for `:formTemplateId` passed in)

ephys: That's not a bug, your `:formTemplateId` is inside of a SQL string so we won't replace it because that will lead to sql injection as well as the risk of mangling data

The `;` thing may be a bug, I'd need an example of a failing query to make sure

denodaeus:
> Hi @ephys
> 
> Screenshot 2022-09-26 at 6 34 18 PM
> 
> With earlier sequelize versions, passing `undefined` as positional parameter was consider as `null` but now an error is been thrown resulting in a breaking change.

This is also how it broke us.  It is often not uncommon in javascript to pass undefined in as positional arguments.  It's worse that this slipped in as not a major breaking change - this is a pretty major, breaking API change.

PR #14473, opened 2022-04-29
ci: cancel old run on new push in PRs

ephys:
This PR makes sure PRs only run one workflow at a time by cancelling the previous workflow if a new commit is added.

This doesn't impact doing multiple pushes to the v7 branch, only PRs are impacted.

I also removed the mention of the v6 branch in the workflow triggers, as the configuration present in the v6 branch is used instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,6 @@ name: CI
 on:
   push:
     branches:
-      - v6
       - v7
   pull_request:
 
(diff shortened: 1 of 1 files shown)


ephys:
Looks like it works. The previous workflow was cancelled, freeing resources for the new workflow and other PRs.

fzn0x (approved):
Don't know if this option exist

sdepold:
Oh this is really cool 😊

Issue #14296, opened 2022-03-30
[RFC] Support for Alternative JS Types


ephys:
With the arrival of new built-in JavaScript types like `BigInt`, and `Temporal`, we need to provide a way to tell Sequelize to use these types instead of the currently used `string` and `Date`.

Such a mechanism would allow users to easily migrate to these new types without creating a very large breaking change.

## Proposed solution

My idea is to add a new option to the different `DataTypes`: `jsType` (e.g. `jsType: 'string'`, `jsType: 'bigint'`). This option would tell Sequelize which built-in JS type should be used when retrieving an attribute from the database.

In the following example, `id` is declared as bigint in the database table, but will be represented as a string in JS.

```typescript
class User extends Model {
  id: string;
}

User.init({
  id: DataTypes.BIGINT,
});
```

By setting the `jsType` option, we call make it be represented as a `bigint` in JS:

```typescript
class User extends Model {
  id: bigint;
}

User.init({
  id: DataTypes.BIGINT({ jsType: `bigint` }),
});
```

*side note: `jsType: 'bigint'` is likely to become the default setting for this data type, in a future major release.*

This only impacts how Sequelize casts the values that come from the database. You can still set the value to a string yourself, this will continue functioning.

### List of valid `jsType` values per Data type

*if the data type is not listed, it doesn't accept this option.*

- [`DataTypes.BINARY`](https://github.com/sequelize/sequelize/issues/14259): `ArrayBuffer`, `Buffer`
- `DataTypes.BIGINT`: `string` (default), `bigint`
- `DataTypes.DECIMAL`: `string` (default), `unsafe-number` (because decimal is an exact decimal, and number is an inexact decimal)
- `DataTypes.DATE`: `Date` (default), `string`
- `DataTypes.DATEONLY`: `string` (default), `Temporal` (= Temporal.PlainDate)
- [`DataTypes.DATETIME`](https://github.com/sequelize/sequelize/issues/14295): `string` (default), `Temporal` (= Temporal.PlainDateTime)
- [`DataTypes.DATETIME.ZONED`](https://github.com/sequelize/sequelize/issues/14295): `string` (default), `Temporal` (= Temporal.ZonedDateTime)
- `DataTypes.TIME`: `string` (default), `Temporal.PlainTime`, `Temporal.Duration`
- `DataTypes.BLOB`: `ArrayBuffer`, `Buffer`

QuentinFarizon:
[...]

PR #14496, opened 2022-05-11
fix(typescript): type-checking fields  in WhereOptions

dattp:
### Pull Request Checklist

#### Before PR!

i have a `User` is model with `name` field is `string`

```
   const user = await User.findOne({
        where: {
            name: 123456789 // pass a number => typescript not error 
        }
   })
```

field `name` is `string` in model, but when i use `findOne` and pass a `number` type, typescript not have error and my program run still.

#### After PR!

show error as `No overload matches this call.` and  my program throw exception.

### DIFF
--- a/src/model.d.ts
+++ b/src/model.d.ts
@@ -207,7 +207,7 @@ export type WhereOptions<TAttributes = any> = AllowNotOrAndWithImplicitAndArrayR
   | Fn
   | Where
   | Json
->;
+> & TAttributes;
 
 // number is always allowed because -Infinity & +Infinity are valid
 export type Rangable<T> = readonly [


ephys:
That isn't the right way to handle this. The part of the typing that tries to determine which types are acceptable is [`StaticValues`](https://github.com/sequelize/sequelize/blob/main/src/model.d.ts#L261-L270).

Right now, it restricts `null`, pg arrays, and pg ranges. Other values are currently all comparable with `boolean | string | number | Buffer | Date` because we can't determine what DataType the TypeScript type maps to. For instance, an attributes typed as `string` could be `DataTypes.STRING` (which only accepts strings) or `DataTypes.DATE` (which also accepts `Date`).

Right now, it was decided that they would be loosely typed voluntarily. We may revisit this once https://github.com/sequelize/sequelize/issues/14296 has been implemented.

dattp:
> Right now, it restricts `null`, pg arrays, and pg ranges. Other values are currently all comparable with `boolean | string | number | Buffer | Date` because we can't determine what DataType the TypeScript type maps to. For instance, an attributes typed as `string` could be `DataTypes.STRING` (which only accepts strings) or `DataTypes.DATE` (which also accepts `Date`).

I think it(type-checking in Where) should be clearly defined, as it can affect program performance.
For example, if the `name` in the DB is `varchar`, but when the query I pass in is a `number`, the index is not used.

Looking forward to the next version!

ephys:
> I think it(type-checking in Where) should be clearly defined

For sure, it would catch more errors, but we can't do that currently. It will need to wait for a future major release (after v7) where Sequelize provides a way to have stable data types.

I'll close for now, when can revisit once the rest of Sequelize is ready for it

PR #14501, opened 2022-05-12
feat(replication): support connection strings

beninsydney:
### Pull Request Checklist

_Please make sure to review and check all of these items:_

- [x] Have you added new tests to prevent regressions?
- [x] Does `yarn test` or `yarn test-DIALECT` pass with this change (including linting)?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you update the typescript typings accordingly (if applicable)?
- [ ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)?

### Description Of Change

Added support for connection strings to options.replication so the write object and the read array can be initialized with strings which are then converted into the object format.

```
{
    replication: {
        write: 'mysql://....',
        read: [ 'mysql://....' ]
    }
}
```

PR #14525, opened 2022-05-20
ci: enable sqlite unit tests

ephys:
SQLite's unit tests were missing from the CI.

This PR enables these unit tests & fixes a (unreleased) regression introduced in https://github.com/sequelize/sequelize/pull/14501


WikiRik (approved):
Not sure how we missed that, but good it's back

Issue #10858, opened 2019-04-28
SequelizeConnectionAcquireTimeoutError randomly DB connections pool error


idangozlan:
I'm running a very simple app that using sequelize basics, it was working for the recent year without any issues on sequelize v4.
Recently, the app is crashing every day once a day, with the following error:

```
TimeoutError: ResourceRequest timed out
    at ResourceRequest._fireTimeout (/usr/src/app/node_modules/generic-pool/lib/ResourceRequest.js:62:17)
    at Timeout.bound (/usr/src/app/node_modules/generic-pool/lib/ResourceRequest.js:8:15)
    at ontimeout (timers.js:498:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:290:5)
```

It was happening with sequelize 4.42.0, mysql2 v1.6.5, node carbon alpine.
After reading a lot about this issue, I figured out that sequelize v5 might handle the connections pool better, so I've tried sequelize 5.7.6, mariadb v2.0.3, but nothing, now the error that I'm getting is:

```
SequelizeConnectionAcquireTimeoutError: Operation timeout
    at pool.acquire.catch.err (/usr/src/app/node_modules/sequelize/lib/dialects/abstract/connection-manager.js:281:53)
    at tryCatcher (/usr/src/app/node_modules/bluebird/js/release/util.js:16:23)
    at /usr/src/app/node_modules/bluebird/js/release/catch_filter.js:17:41
    at tryCatcher (/usr/src/app/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/usr/src/app/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (/usr/src/app/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (/usr/src/app/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/usr/src/app/node_modules/bluebird/js/release/promise.js:690:18)
    at _drainQueueStep (/usr/src/app/node_modules/bluebird/js/release/async.js:138:12)
    at _drainQueue (/usr/src/app/node_modules/bluebird/js/release/async.js:131:9)
    at Async._drainQueues (/usr/src/app/node_modules/bluebird/js/release/async.js:147:5)
    at Immediate.Async.drainQueues (/usr/src/app/node_modules/bluebird/js/release/async.js:17:14)
[...]

PR #14535, opened 2022-05-24
fix: kill connection on commit/rollback error

ephys:
### Pull Request Checklist

### Description Of Change

This is an attempt at fixing https://github.com/sequelize/sequelize/issues/10858. I don't know if it'll completely fix it, so in doubt I'm not closing that issue.

#### What was going on?

If an error occurred during a transaction commit or rollback (such as a network error), Sequelize would call `transaction.cleanup()`, which released the connection to the pool.

The problem is that if an error occurred, we cannot guarantee that the transaction did actually commit or roll back. It may still be active. Releasing the connection to the pool does not kill the connection, and it will end up used by other queries - unknowingly running in another transaction.

#### The fix

This PR changes that by taking the drastic approach of killing the connection if an error occurred during `COMMIT TRANSACTION` or `ROLLBACK TRANSACTION`.

### DIFF
--- a/src/dialects/abstract/connection-manager.js
+++ b/src/dialects/abstract/connection-manager.js
@@ -298,8 +298,6 @@ class ConnectionManager {
    * Release a pooled connection so it can be utilized by other connection requests
    *
    * @param {Connection} connection
-   *
-   * @returns {Promise}
    */
   releaseConnection(connection) {
     this.pool.release(connection);
--- a/src/dialects/abstract/connection-manager.js
+++ b/src/dialects/abstract/connection-manager.js
@@ -301,9 +301,19 @@ class ConnectionManager {
    *
    * @returns {Promise}
    */
-  async releaseConnection(connection) {
+  releaseConnection(connection) {
     this.pool.release(connection);
     debug('connection released');
+  }
+
+  /**
+   * Destroys a pooled connection and removes it from the pool.
+   *
+   * @param {Connection} connection
+   */
+  async destroyConnection(connection) {
+    await this.pool.destroy(connection);
+    debug(`connection ${connection.uuid} destroyed`);
   }
 
   /**
--- a/test/integration/sequelize.transaction.test.js
+++ b/test/integration/sequelize.transaction.test.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const sinon = require('sinon');
 const chai = require('chai'),
   expect = chai.expect,
   Support = require('./support'),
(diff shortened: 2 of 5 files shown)


src/dialects/abstract/connection-manager.js
```diff
@@ -301,11 +301,21 @@ class ConnectionManager {
    *
    * @returns {Promise}
```
sdepold, line 302: I guess this is wrong

src/dialects/abstract/connection-manager.js
```diff
@@ -301,11 +301,21 @@ class ConnectionManager {
...
    */
-  async releaseConnection(connection) {
+  releaseConnection(connection) {
     this.pool.release(connection);
```
sdepold, line 305: This is sync?

ephys: It did not await nor return anything, and looking into sequelize-pool, it's doing anything async, just adding the connection back to the available connection set.

sdepold: Strange hm? But I agree looking at async fun without return or await certainly doesn't make sense :)

ephys:
Yes, I think that's how I'll do that :)

test/integration/sequelize.transaction.test.js
```diff
@@ -1,141 +1,181 @@
...
-      it('gets triggered once a transaction has been successfully committed', async function() {
-        let called = false;
+  const stubs = [];
+
+  afterEach(() => {
+    for (const stub of stubs) {
+      stub.restore();
+    }
+
+    stubs.length = 0;
```
sdepold, line 26: Uhm. Well. Hmm... `stubs = []` maybe? 🙈

test/integration/sequelize.transaction.test.js
```diff
@@ -1,141 +1,181 @@
...
           .sequelize
           .transaction();
 
-        await t.commit();
-        called = 1;
-        expect(called).to.be.ok;
+        await expect(t.commit()).to.be.rejectedWith('Oh no, an error!');
+
+        // connection should have been destroyed
+        expect(this.sequelize.connectionManager.pool.size).to.eq(Math.max(0, initialPoolSize - 1));
```
sdepold, line 53: I don't even understand my confusion and why this line bothers me but doesn't this somehow allow for initialPoolSize to be 0? Is that a realistic scenario?

test/integration/sequelize.transaction.test.js
ephys: I wanted to avoid using `let` but it's not very important so I can change it if you like :)

In the future I want to use `sinon.restore()` after all tests, which would make this redundant, but right now it breaks a bunch of things

sdepold: I think my future me would appreciate that change :)

test/integration/sequelize.transaction.test.js
sdepold: Hmm... Okay then? Maybe? 😂

sdepold (approved):

sdepold:
I approved it. Don't know how strong I feel about the .length change. You decide

test/integration/sequelize.transaction.test.js
ephys: Alright then, I'll need re-approval though :)

sdepold (approved):

PR #14547, opened 2022-05-25
docs: update Contributing.md (v6)

mike-usa:
### Description Of Change

fix spacing between listed databases in Docker's optional (recommended) requirements

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,7 +101,7 @@ Most operating systems provide all the needed tools (including Windows, Linux an
 
   - [Docker](https://docs.docker.com/get-docker/)
     - It is not mandatory because you can easily locally run tests against SQLite without it.
-    - It is practically mandatory if you want to locally run tests against any other database engine (MySQL, MariaDB, Postgres,Db2 and MSSQL), unless you happen to have the engine installed and is willing to make some manual configuration.
+    - It is practically mandatory if you want to locally run tests against any other database engine (MySQL, MariaDB, PostgreSQL, Db2 and MSSQL), unless you happen to have the engine installed and is willing to make some manual configuration.
   - [Visual Studio Code](https://code.visualstudio.com/)
     - [EditorConfig extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)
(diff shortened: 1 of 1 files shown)


fzn0x (approved):

WikiRik:
Not sure if we need this on the v6 branch, especially since most people will read this file in the main branch. Could you rebase this on main?

ephys:
I was about to say the same thing :)
We're considering v6 to be in maintenance, changes to contributing docs like this should target `main`, at least at first.

Other than that the contents are good

mike-usa:
This may have been a way of me experimenting with the change-control process for a minor issue.  I'm comfortable with closing this PR entirely.

I have yet to understand (haven't looked into yet) how NPM gets updated from this repo.  I thought `v6` branch corresponds with sequelize@6 in NPM and `main` corresponds with sequelize@7 and since v7 is still alpha/beta then patches should be applied to v6 so that users performing `npm install` will receive the updates.

WikiRik:
Thanks for your involvement and the points you mention, we could definitely use those in improving the CONTRIBUTING files and our documentation.

> I have yet to understand (haven't looked into yet) how NPM gets updated from this repo.  I thought `v6` branch corresponds with sequelize@6 in NPM and `main` corresponds with sequelize@7 and since v7 is still alpha/beta then patches should be applied to v6 so that users performing `npm install` will receive the updates. 

This is mostly correct, we do have a `v7` that corresponds with the v7 releases under `@sequelize/core`. New commits to `main` do not correspond to a new release. We do support patches in v6, but we try to refrain from new features since we want to focus our efforts on releasing v7 soon.

The CONTIRBUTING file, however, is only shown on GitHub (or linked through other pages) which has `main` as the branch you see first so 90% of the views are on there. So doing this small change on the v6 branch is not worth it in my opinion.

ephys:
> So doing this small change on the v6 branch is not worth it in my opinion.

This is the main reason imo :)

The other reason we first update the main branch, then replicate the change in other branches, is that it reduces the risk of 'losing' work.
If you make a change in main, then forget to backport to v6, it's 'just' a feature that was not backported. Future versions will continue to have it.
If a change is made in v6, but not in main, then the change is missing when users upgrade, which is an undocumented breaking change. Future versions will not have it.

Issue #14548, opened 2022-05-25
`findOne` logic for adding `limit: 1` led to severe performance degredation.


rharriso:
## Issue Creation Checklist

- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description

Misuse of `@Unique` annotation in `seqsequelize-typescript` triggered a condition in `findOne` that lead to ALL rows being queried, rather than halting after a single row was retrieved. The invalid "uniqueness" of the column caused the `findOne` method to omit the `LIMIT 1` clause causing ALL of the matching rows to be queried.

Foreign keys should be used to ensure data integrity, and should not impact query performance. It's unclear how omitting `LIMIT 1` benefits any query that should return a single result.

### SSCCE

```ts
export class APITransaction extends Model<IAPITransactionAttributes> {
  @ForeignKey(() => APIUser)
  @Unique // should not be here, foreign key is not actually unique
  @AllowNull(false)
  @NotEmpty
  @Column({
    type: DataType.STRING,
  })
  apiUserId: string;
}

// selects and sorts 50k results just to get a single value.
const apiTx = await APITransaction.findOne({
    where: { apiUserId: apiUser.slug },
    order: [['createdAt', 'DESC']]
});
```

### What do you expect to happen?

`findOne` should **always** select a single row from the database

### What is actually happening?

In the case outlined above, `findOne` would query for **every** row matching the where clause, then return the first.
This led to rapidly degrading query performance and cascading problems as application servers failed over. The databases because saturated with abandoned connections. Application servers thrashed. Database and Application servers failed over under modest load.

### Additional context

This issue was eventually resolved after reviewing the slow query log, and replacing all uses of `findOne` with `findAll({ limit: 1})`. It was not detectable until the row count that matched the where clause reached > 10k. 

We also resolved the misuse of `@Unique` in our own code base.

### Environment

- Sequelize version: 6.8
- Node.js version: 16
- If TypeScript related: TypeScript version: 
- Database & Version: PSQL 13
- Connector library & Version: 

## Would you be willing to resolve this issue by submitting a Pull Request?

Yes, coming shortly.

WikiRik:
No worries, let us know in the PR if there's something we can help you with

WikiRik:
Yeah, we updated some files in the dev folder. Merging main into your branch should work

rharriso:
🥳

WikiRik:
[...]

PR #14549, opened 2022-05-25
feat: Always add LIMIT 1 to `findOne` queries

rharriso:
### Pull Request Checklist

- [x] Have you added new tests to prevent regressions? **No, tests remove**
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)? No this removes an undocumented behavior

### Description Of Change

Remove conditional logic for adding `LIMIT 1` to `findOne` queries. Instead, always add it.

* Foreign keys should be used to ensure data integrity, and should not impact query performance.
* It's unclear how omitting `LIMIT 1` benefits any query that should return a single result.
* Models are imperfect representations of schema, making unnecessarily assumptions about their correctness can lead to severe issues.
  * see #14548

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -1875,18 +1875,8 @@ ${associationOwner._getAssociationDebugList()}`);
     }
 
     options = Utils.cloneDeep(options);
-
-    if (options.limit === undefined) {
-      const uniqueSingleColumns = _.chain(this.uniqueKeys).values().filter(c => c.fields.length === 1)
-        .map('column')
-        .value();
-
-      // Don't add limit if querying directly on the pk or a unique column
-      if (!options.where || !_.some(options.where, (value, key) => (key === this.primaryKeyAttribute || uniqueSingleColumns.includes(key))
-          && (Utils.isPrimitive(value) || Buffer.isBuffer(value)))) {
-        options.limit = 1;
-      }
-    }
+    // findOne only ever needs one result
+    options.limit = 1;
 
     // Bypass a possible overloaded findAll.
     return await Model.findAll.call(this, (_.defaults(options, {
--- a/src/model.js
+++ b/src/model.js
@@ -1876,7 +1876,11 @@ ${associationOwner._getAssociationDebugList()}`);
 
     options = Utils.cloneDeep(options);
     // findOne only ever needs one result
-    options.limit = 1;
+    // conditional temporarily fixes 14618
+    // https://github.com/sequelize/sequelize/issues/14618
+    if (options.limit === undefined) {
+      options.limit = 1;
+    }
 
     // Bypass a possible overloaded findAll.
     return await Model.findAll.call(this, (_.defaults(options, {
--- a/test/unit/model/find-one.test.js
+++ b/test/unit/model/find-one.test.js
@@ -22,72 +22,11 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       this.stub = Sequelize.Model.findAll = sinon.stub().resolves();
     });
 
-    describe('should not add limit when querying on a primary key', () => {
-      it('with id primary key', async function () {
-        const Model = current.define('model');
-
-        await Model.findOne({ where: { id: 42 } });
-        expect(this.stub.getCall(0).args[0]).to.be.an('object').not.to.have.property('limit');
-      });
-
-      it('with custom primary key', async function () {
-        const Model = current.define('model', {
-          uid: {
-            type: DataTypes.INTEGER,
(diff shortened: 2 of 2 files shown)


kpeluso:
+1

what is being corrected here is an "anti-feature"

ephys:
I agree, not adding `LIMIT 1` if we're comparing the primary key sounds weird to me. I'd expect `Model.findOne()` to be equivalent to `(await Model.findAll({ limit: 1 }))[0]`.

Strangely enough, some mssql tests are failing with this change

rharriso:
@ephys , sorry to say I have been able to identify a commit where these tests started failing. I'm wondering if that has come up in other PRs?

To Reiterate, the MSSQL tests fail locally for me on `main`

ephys:
are you using `start-mssql` to start the server?

rharriso:
@ephys, I'm trying to make sure I'm properly running the test locally. 
I'll get back to you once I'm certain that I'm getting the same resuts as CI.

Thanks for your patience.

src/model.js
```diff
@@ -1875,18 +1875,8 @@ ${associationOwner._getAssociationDebugList()}`);
     }
 
     options = Utils.cloneDeep(options);
-
-    if (options.limit === undefined) {
```
WikiRik, line 1879: I'm not sure why but keeping this if statement fixes the failing tests.
```js
  options.limit = 1;
}
```
seems to work just fine

rharriso: One of the failing tests is below.
For some reason the MSSQL query is including the `User.id` column in the query even after it is deleted on line 666 🤘.

Logging the sql string in `abstract/query-interface.js#select`:

```sql
SELECT [id], [title], [createdAt], [updatedAt], [user_name] FROM [Tasks] AS [Task] WHERE [Task].[title] = N'some task' ORDER BY [Task].[id] OFFSET 0 ROWS FETCH NEXT 1 ROWS ONLY;
SELECT [username], [createdAt], [updatedAt] FROM [Users] AS [User] WHERE [User].[username] = N'bob' ORDER BY [User].[id] OFFSET 0 ROWS FETCH NEXT 1 ROWS ONLY;
```

Notice `[User].[id]` in the order statement.

https://github.com/sequelize/sequelize/blob/428a85b1376b09373a362d42abedfae0790b8293/test/integration/associations/belongs-to.test.js#L662-L682

rharriso: Dug a bit deeper. And I think I found the source of the issue. 

`src/dialects/mssql/query-generator.js#addLimitAndOffeset`.

`primaryKeyField` is used to order the subqueries. You'll notice in my comment above that the primary key would have been deleted.

See the following console output.

```
console.log('shouldn't be', Object.keys(model.rawAttributes).join(' 🤝 '));
shouldn't be username 🤝 createdAt 🤝 updatedAt
console.log('still here','${this.quoteTable(options.tableAs || model.name)}.${this.quoteIdentifier(model.primaryKeyField)}'
still here [User].[id]
'😭 ${model.primaryKeyField}'
😭 id
```

Not quite sure what the resolution is here. If the PK has been deleted.....

rharriso:
I've added the conditional pointed out by @WikiRik, and opened up issue #14618

ephys:
I found the reason why LIMIT was not included https://github.com/sequelize/sequelize/pull/9248. I'm documenting it here for future PRs but I disagree with the premise of that PR and I agree LIMIT should be specified.

ephys (approved):

WikiRik (approved):

WikiRik:
I'll run the tests again after GitHub solves their incident; https://www.githubstatus.com/incidents/6r4sms3hcj3x

vanthome:
Hello, I have my doubts that this is correct. Are we sure that `findOne` queries should really only ever return one row? We are using `findOne` in conjunction with includes and then, because of outer joins happening, we do get and expect multiple rows. Please correct me if I'm wrong. Thx

ephys:
Hi :)

In `find{One,All}`, `limit` doesn't necessarily mean only one row in the database is going to be fetched. We want `limit` to only apply to the level it's on, in the redesign. So if you wrote the following:

```ts
User.findOne({
  include: [User.associations.projects],
});
```

It would return a single user, but include all of the user's projects

We know that currently `limit` behaves in very inconsistent ways, but it's something [I'm working on addressing](https://github.com/sequelize/website/pull/298/files#diff-d8aa946ffed5a9f3e0abc36699e64393e20f1ad6ca5b0997aa165325f3fa651cR324)

PR #14539, opened 2022-05-25
docs: add CONTRIBUTING.POSTGRESQL.CHEATSHEET.md to v6

mike-usa:
### Description Of Change

Cheatsheet is an extension of CONTRIBUTING.md

It covers:
    - test prerequisites
    - using docker and connecting containers
    - configuring pgadmin
    - debugging expectations and discussion
    
It's intended as a quick guide targeting those new to _docker_ and _Sequelize_ development.  This commit applies some focus to PostgreSQL and configuring pgadmin.

PR #14554, opened 2022-05-26
docs(contributing-cheatsheet): add new cheatsheet docs

mike-usa:
Cheatsheet is an extenstion to CONTRIBUTING.md. This flavor focuses on the PostgreSQL dialect.

It covers:
- Sequelize tests
  - prequisite steps
  - running various tests and expectations
  - debugging expectations
- using docker (basic commands)
  - creating docker networks
  - inspecting docker containers
  - connecting containers to docker networks
- setting up pgadmin4
  - running as a container
  - connecting it with database container
  - configuring pgadmin settings

This is not intended as a replacement to CONTRIBUTING.md, which contains must-read information. This document is intended as a quick guide targeting those new to Docker and Sequelize development. There is great opportunity for stylistic and content improvements.

### DIFF
--- a/CONTRIBUTING.CHEATSHEET.POSTGRESQL.md
+++ b/CONTRIBUTING.CHEATSHEET.POSTGRESQL.md
@@ -1,0 +1,176 @@
+# Contributing: Cheatsheet (PostgreSQL)
+
+The [CONTRIBUTING](CONTRIBUTING.md) doc provides an extensive amount of information. Below is a summary of that info to get up-and-running quickly, with a few additional explanations.
+
+## Pre-Requisites
+
+1. Have docker installed
+1. Have `yarn` installed (see _Helpful Notes_ section below)
+1. Have the database installed and running
+
+   Test Type | Requires Running Database
+   ---         | --- |
+   Unit        | No  |
+   Integration | Yes |
+   Other       | Yes | 
+
+   Note:
+   - tests DO NOT automatically start the database they're testing against  
+   (see _Helpful Notes_ section below)
+   - if the database is not running, you may encounter an error message similar to:
+      ```bash
+      SequelizeConnectionRefusedError: connect ECONNREFUSED 127.0.0.1:23010
+      ```
+
+---
+
+## Running Tests
+This example will use _PostgreSQL_ as the database being contributed to. Replace _postgres_ with whatever dialect is specific to your changes.
+
+The basic structure is start, test, stop. Start the database, run the test script, stop the database.
+
+```bash
+yarn start-postgres            # start the database
+yarn test-integration-postgres # run the test
+yarn stop-postgres             # stop the database
+```
+
+### Different Tests
+
+Test | Description
+---- | ---
+`DIALECT=postgres yarn test` | **Type**: Full test of a database<br>**Requires**: specified dialect & running db
+`yarn test-integration` | **Type**: Partial test<br>**Requires**: running db
+`yarn test-unit` | **Type**: Partial test<br>**Requires**: no requirement can be run as-is
+`DIALECT=postgres npx mocha test/**/query-generator.test.js` | **Type**: Specific test<br>**Notice**: it calls `mocha` directly <br>**Requires**: varies based on test
+
+---
+
+## Debugging / Cheatsheet
+
+Command | Description
+--- | ---
+`console.log` | `node --inspect` is not currently utilized and would require modifications into the tests. In the meantime, use `console.log` for debugging
+`docker ps -a` | Display running containers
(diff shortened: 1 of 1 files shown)


mike-usa:
There was a slight deviation here from #14539 in that I changed the name, thinking that other dialects might want to have their own -- perhaps MySQL and phpMyAdmin?

This wasn't intended to be a book or fully comprehensive, but I hope it helps others.

CONTRIBUTING.CHEATSHEET.POSTGRESQL.md
```diff
@@ -0,0 +1,176 @@
...
+   ```yaml
+   Host: localhost
+   Port: 23010 (or whichever port revealed by `docker ps -a`)
+   Username: sequelize_test
+   Password: sequelize_test
+   ```
+---
+
+## Helpful Notes and Explanations
+1. Most scripts in _package.json_ use `yarn`, which is not a package.json dependency and must installed globally or available by some other means; for instance, it does not need to be specifically installed. Read on:
```
ephys, line 131: I think it would be simpler to link to yarn's install documentation which explains all of this, as otherwise we'll have to keep this in sync with theirs :)

CONTRIBUTING.CHEATSHEET.POSTGRESQL.md
```diff
@@ -0,0 +1,176 @@
...
+## Debugging / Cheatsheet
+
+Command | Description
+--- | ---
+`console.log` | `node --inspect` is not currently utilized and would require modifications into the tests. In the meantime, use `console.log` for debugging
+`docker ps -a` | Display running containers
+`docker exec -it <container name> psql -d sequelize_test -U sequelize_test` | Connect to a running database inside the container and run psql (psql is PostgreSQL-only)
+
+### PgAdmin4
+1. Running pgadmin4 as a container (pgadmin is PostgreSQL-only)
```
ephys, line 58: Setting up the web version of pgAdmin4 is pretty cumbersome, is there any reason to use it instead of the desktop client?

ephys (commented):
This is very thorough, really nice :)

A lot of it would work for the other dialects though, the only difference being the command name and which client needs to be installed

Do you think making this more generic (replacing pgAdmin4 Web with the desktop version so we can just link to their install page, and maybe mentioning other db clients like datagrip for the other dialects, and listing the possible commands for the different dialects) be still easy to understand?

CONTRIBUTING.CHEATSHEET.POSTGRESQL.md
mike-usa: The CONTRIBUTING.md already included the docker code to start up pgadmin4; however, I couldn't get it working.  That's why I included it here to mention how to get it working.

As for _why_ I imagine that's going to be a preference.  I know at my work, we are more limited on certain applications (or really the sockets) and think the web client would be a little less weight and bypass some of the more restrictive policies.

For my preference, I don't really use a GUI, just psql when necessary.  I do think the configuration steps here may be of use to setting up other interfaces as it demonstrates how to work between the web and docker.

CONTRIBUTING.CHEATSHEET.POSTGRESQL.md
mike-usa: Your call.  I didn't realize it until you told me about it :)  Since `yarn` is a dependency of the project (almost all npm scripts call it), but not included in the dev-dependencies, something needs to be stated as to why it wasn't included, or how to use it without installing it.

Linking to the documentation would be a good idea, but still need to be sure the links don't break in the future.

mike-usa:
> Do you think making this more generic (replacing pgAdmin4 Web with the desktop version so we can just link to their install page, and maybe mentioning other db clients like datagrip for the other dialects, and listing the possible commands for the different dialects) be still easy to understand?

- I think there's an opportunity to include other languages and nuances.  As size grows, it would certainly need to be refactored and re-organized.  I guess I kind of saw this as boilerplate to capture all that crucial information that you shared with this "NOOB".  
- For pgadmin4, I'm more inclined to want more info than to remove the things that were tricky for me to figure out -- much with your help :). I'm not familiar with some of the other databases and others I haven't used in 10 years, but yes I was hoping the power users would add their knowledge
- Aside from a few Postgres-specific things and the examples being in Postgres (convenience), my goals was for it to be generic notes for using Sequelize, not particular to Postgres, but with the added language nuances/frustrations -- simple things like running tests and in what order even that `prepare` will compile `src` into `lib` and `lib` is what you're testing against.  These are simple things to the experienced Sequelize developer, but it was a learning curve for me to look at it.

Note: I originally came here with a simple patch that I made to the files in `node_modules` and had to figure out the development process in order to copy that code to this repo. That's kind of what my goal was; to encourage contributions.

ephys:
@WikiRik as this is a bit outdated now, should I put it in draft until it's ready to be reviewed?

PR #14555, opened 2022-05-26
fix(sqlite): don't use mkdirp if the file already exists, to fix windows error

cuteinfo:
### Pull Request Checklist

### Description Of Change

On Windows, using fs.mkdir() on the root directory even with recursion will result in an error

https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fsmkdirpath-options-callback#:~:text=on%20the%20root%20directory%20even%20with%20recursion%20will%20result%20in%20an%20error

### DIFF
--- a/src/dialects/sqlite/connection-manager.js
+++ b/src/dialects/sqlite/connection-manager.js
@@ -67,7 +67,7 @@ export class SqliteConnectionManager extends ConnectionManager {
       return this.connections[options.inMemory || options.uuid];
     }
 
-    if (!options.inMemory && (options.readWriteMode & this.lib.OPEN_CREATE) !== 0) {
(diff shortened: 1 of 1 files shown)


src/dialects/sqlite/connection-manager.js
```diff
@@ -67,7 +67,7 @@ export class SqliteConnectionManager extends ConnectionManager {
       return this.connections[options.inMemory || options.uuid];
     }
 
-    if (!options.inMemory && (options.readWriteMode & this.lib.OPEN_CREATE) !== 0) {
+    if (!options.inMemory && (options.readWriteMode & this.lib.OPEN_CREATE) !== 0 && !fs.existsSync(path.dirname(options.storage)) ) {
```
ephys, line 70: could you extract `path.dirname(options.storage)` to a variable to avoid computing it twice?

ephys (changes requested):
There is a lint error, and a small tweak, but lgtm

WikiRik:
I'll rework this in a new PR

Issue #14161, opened 2022-02-24
Tests are flaky again


ephys:
This issue aims to document the various tests that fail randomly on the CI.

https://github.com/sequelize/sequelize/runs/5320675964?check_suite_focus=true
```
# mariadb 10.5 (node 16)
  1) [MARIADB] Pooling
       idle
         [MSSQL Flaky] should get new connection beyond idle range:

      AssertionError: expected 46 to not equal 46
      + expected - actual


      at Proxy.assertEqual (node_modules/chai/lib/chai/core/assertions.js:1038:12)
      at Proxy.methodWrapper (node_modules/chai/lib/chai/utils/addMethod.js:57:25)
      at doAsserterAsyncAndAddThen (node_modules/chai-as-promised/lib/chai-as-promised.js:289:22)
      at Proxy.<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:255:20)
      at Proxy.overwritingMethodWrapper (node_modules/chai/lib/chai/utils/overwriteMethod.js:78:33)
      at assertNewConnection (test/integration/pool.test.js:47:48)
      at Context.<anonymous> (test/integration/pool.test.js:213:7)
      at runMicrotasks (<anonymous>)
      at runNextTicks (node:internal/process/task_queues:61:5)
      at processTimers (node:internal/timers:499:9)
```

https://github.com/sequelize/sequelize/runs/5155830293?check_suite_focus=true

```
# postgres 9.5 (node 12) (minified aliases)
  1) [POSTGRES Specific] DAO
       enums
         ARRAY(ENUM)
           should be able to insert a new record even with an array of enums in a schema:
     schema "special_schema" does not exist
  Error
      at Query.run (lib/dialects/postgres/query.js:55:25)
      at /home/runner/work/sequelize/sequelize/lib/sequelize.js:315:28
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
      at async /home/runner/work/sequelize/sequelize/lib/dialects/postgres/query-interface.js:104:76
      at async /home/runner/work/sequelize/sequelize/lib/dialects/postgres/query-interface.js:104:96
      at async PostgresQueryInterface.ensureEnums (lib/dialects/postgres/query-interface.js:104:20)
      at async PostgresQueryInterface.createTable (lib/dialects/abstract/query-interface.js:85:5)
      at async Function.sync (lib/model.js:938:5)
[...]

PR #14560, opened 2022-05-27
fix(postgres-test): insert record with enums array into schema (#14161)

mike-usa:
### Pull Request Checklist

### Description Of Change

  - User.sync(), which references the schema, was executed before the schema exists
  - add an `await` before the schema is created to enforce sequencing/synchronocity
Subtle style change:
  - remove word 'even' from test description

Addresses only PostgreSQL-portion of issue #14161

### DIFF
--- a/test/integration/dialects/postgres/dao.test.js
+++ b/test/integration/dialects/postgres/dao.test.js
@@ -464,9 +464,10 @@ if (dialect.startsWith('postgres')) {
           expect(user.length).to.equal(1);
         });
 
-        it('should be able to insert a new record even with an array of enums in a schema', async function () {
+        it('should be able to insert a new record with an array of enums in a schema', async function () {
           const schema = 'special_schema';
-          this.sequelize.createSchema(schema);
+          await this.sequelize.createSchema(schema);
+
           const User = this.sequelize.define('UserEnums', {
             name: DataTypes.STRING,
             type: DataTypes.ENUM('A', 'B', 'C'),


WikiRik (approved):
Good catch! Let's hope that this will indeed solve the flaky test

ephys:
Hey :) Thanks for the PR!
It looks like this PR is only related to testing. The title of this PR used 'fix' as the commit type but 'fix' should only be used if it fixes a bug that is not internal (a bug that can impact users), as 'fix' commits are automatically listed in the changelog (same for 'feat').

NB: This is only about the PR title, as it's used as the name of the commit once we squash-merge the PR

Not to worry I'll update the commit message in our commit history :)

mike-usa:
TY @ephys that's good to know -- maybe more opportunity to update CONTRIBUTING.md 😏 -- sorry for creating more work. I scoured the AngularJS commit recommendation and tried looking at the commit history. Would you recommend this as `meta`? I see `ci`, `chore`, `build`, and `refactor(build)`. I imagine the more accurate I am, the less work you'll have to do with integrating/squashing commits.

## Tangent
The AngularJS commit message document (linked in CONTRIBUTING) includes contradictory examples to their recommendations.  For instance, it recommends to drop the past-tense like `added` in favor of `add`, but then use different verb tenses in the message-body. Small thing, but I found it amusing and thought others might too.

WikiRik:
@ephys woops, did not notice that before merging 😅

@mike-usa `test` is fine, the ones we allow are mentioned in the package.json file; https://github.com/sequelize/sequelize/blob/main/package.json#L172-L182

ephys:
I just updated the commit message
By the way, good catch on this!

ephys:
> maybe more opportunity to update CONTRIBUTING.md

Definitely :) I think our [note about it](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md#6-commit-your-modifications) is outdated. But there are also further changes I'd like to make:

- We don't really care about the *commit message* itself, because we squash all commits from a PR. Only the PR title matters. We should consider removing conventional commits from our git hooks, and use something like https://github.com/marketplace/actions/conventional-pr-title instead.
- Only `feat` and `fix` types are really important, because they are included in our changelog. All of the other commit types don't really have any purpose. We could consider simplifying them, as we have many of them and they overlap (`build`, `chore`, `ci`, `refactor`, `style`, `test`, `meta`). We could ban all of them except for `meta` (both short and generic). Maybe `meta` & `test`?
- Similarly, the commit scope is only important for `feat` & `fix` commits because they're included in our changelog. I try to limit them to `deps`, `types`, and our list of dialect names (`postgres`, `mysql`, etc…).

PR #14570, opened 2022-05-30
fix: fix `sync` duplicating existing foreign keys when a schema is used

Antamokhin:
### Pull Request Checklist

### Description Of Change
Sync functionality duplicates existing foreign key associations for per sync executions when schema is used.
Expected behaviour: It shouldn't duplicate foreign key association if exists.

Code block reproduces the issue can be found [here](https://stackoverflow.com/questions/64806845/sequelize-sync-duplicates-existing-foreign-key-associations).

Changes from PR [#12833](https://github.com/sequelize/sequelize/pull/12833)  in models.js (sync function, block about references checking) are not included in PR [#13600](https://github.com/sequelize/sequelize/pull/13600). 

References.model is object when schema is defined, not string, due addSchema method in query-generator. And because there is no chance to remove old foreign keys.

I just implement changes proposed by @denizkanmaz


ephys:
Thank you for the PR. Could you add an integration test that covers the issue, to prevent future regressions?

Antamokhin:
@ephys added.

test/integration/model/sync.test.js
```diff
@@ -403,6 +405,33 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           expect(results.filter(r => r.name === 'wow_my_index')).to.have.length(1);
         }
       });
+
+      it('should remove old foreign keys when sync with {alter: true} if custom schema', async function () {
+        if (current.dialect.supports.schemas) {
```
ephys, line 410: this condition should be outside of the test instead of inside

test/integration/model/sync.test.js
```diff
@@ -6,6 +6,8 @@ const { DataTypes } = require('@sequelize/core');
 const expect = chai.expect;
 const Support = require('../support');
 
+const current = Support.sequelize;
```
ephys, line 9: We're in the process of modernizing the codebase and instead of using `current` and `this.sequelize` (which are actually the same values), we're just using the `sequelize` variable

Example in this file: https://github.com/sequelize/sequelize/blob/main/test/unit/query-interface/decrement.test.ts#L19

Could you replace new usages of `current` and `this.sequelize` which just `sequelize`?

 ```suggestion
const { sequelize } = Support;
```

File: test/integration/model/sync.test.js [hunk 3]
```diff
@@ -403,6 +405,33 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           expect(results.filter(r => r.name === 'wow_my_index')).to.have.length(1);
         }
       });
+
+      it('should remove old foreign keys when sync with {alter: true} if custom schema', async function () {
+        if (current.dialect.supports.schemas) {
+          let schema = 'test';
+          if (['mysql', 'mariadb'].includes(current.dialect.name)) {
```
ephys, line 412: `mariadb` supports `createSchema`

test/integration/model/sync.test.js
```diff
@@ -403,6 +405,33 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+            schema = this.sequelize.config.database;
+          } else {
+            await this.sequelize.createSchema(schema);
+          }
+
+          const User = this.sequelize.define('User', {
+            user_id: {
```
ephys, line 419: this should be camelCase. Any reason to use this instead of the automatically generated `id` primary key?

test/integration/model/sync.test.js
```diff
@@ -403,6 +405,33 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+              type: DataTypes.STRING,
+            },
+          }, { schema });
+          const BelongsToUser = this.sequelize.define('BelongsToUser', {
+            email: {
```
ephys, line 424: is this `email` attribute relevant to this test? Otherwise it can be removed

test/integration/model/sync.test.js
```diff
@@ -403,6 +405,33 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+              type: DataTypes.STRING,
+            },
+          }, { schema });
+          BelongsToUser.belongsTo(User, { foreignKey: { targetKey: 'user_id', allowNull: false } });
```
ephys, line 428: There is no `targetKey` option in `foreignKey`. If you want to change the targetKey (which defaults to the primary key), it goes on the same level as `foreignKey`

test/integration/model/sync.test.js
```diff
@@ -403,6 +405,33 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+          await this.sequelize.sync({ alter: true });
```
ephys, line 429: If I understand correctly, this test ensures that `sync` doesn't create the foreign key again if it already exists, so this should `sync` with alter twice

test/integration/model/sync.test.js
```diff
@@ -403,6 +405,33 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
         }
       });
+
+      it('should remove old foreign keys when sync with {alter: true} if custom schema', async function () {
+        if (current.dialect.supports.schemas) {
+          let schema = 'test';
+          if (['mysql', 'mariadb'].includes(current.dialect.name)) {
+            schema = this.sequelize.config.database;
+          } else {
+            await this.sequelize.createSchema(schema);
```
ephys, line 415: the value of schema should be assigned here instead of outside the if-else

test/integration/model/sync.test.js
```diff
@@ -403,6 +405,33 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           expect(results.filter(r => r.name === 'wow_my_index')).to.have.length(1);
         }
       });
+
+      it('should remove old foreign keys when sync with {alter: true} if custom schema', async function () {
```
ephys, line 409:
```suggestion
      it('should not recreate a foreign key if it already exists when { alter: true } is used with a custom schema', async function () {
```

src/model.js
```diff
@@ -1283,16 +1283,21 @@ ${associationOwner._getAssociationDebugList()}`);
           const references = currentAttribute.references;
           if (currentAttribute.references) {
             const database = this.sequelize.config.database;
-            const schema = this.sequelize.config.schema;
+            const schema = tableName.schema;
+            const foreignReferenceSchema = currentAttribute.references.model.schema;
+            const foreignReferenceTableName = typeof references.model === 'object'
```
ephys, line 1288: `references.model` can also be a model, on which `getTableName` should be called to get the table name (could you add a test for this as well, can be done by defining an attribute with the `references` option set manually).

There is a function `isModelStatic` which can be used to determine whether `references.model` is a model class

src/model.js
```diff
@@ -1283,16 +1283,21 @@ ${associationOwner._getAssociationDebugList()}`);
           const references = currentAttribute.references;
           if (currentAttribute.references) {
             const database = this.sequelize.config.database;
-            const schema = this.sequelize.config.schema;
+            const schema = tableName.schema;
```
ephys, line 1286: shouldn't this default to `this.sequelize.config.schema` if it's not set? Maybe something to change in `Model.getTableName()`

src/model.js
```diff
@@ -1283,16 +1283,21 @@ ${associationOwner._getAssociationDebugList()}`);
...
             for (const foreignKeyReference of foreignKeyReferences) {
               const constraintName = foreignKeyReference.constraintName;
               if ((Boolean(constraintName)
                 && foreignKeyReference.tableCatalog === database
                 && (schema ? foreignKeyReference.tableSchema === schema : true)
-                && foreignKeyReference.referencedTableName === references.model
+                && foreignKeyReference.referencedTableName === foreignReferenceTableName
                 && foreignKeyReference.referencedColumnName === references.key
-                && (schema ? foreignKeyReference.referencedTableSchema === schema : true)
+                && (foreignReferenceSchema
```
ephys, line 1298: I think this will fail if the model has two foreign keys: one without foreignReferenceSchema, and one with it.

The one without would match either as this condition is skipped

ephys (changes requested):
Lots of comments, but given how many issue reports we have about `sync`, I want to make sure this issue is fixed for good and won't be back. Thank you very much for your help :)

ephys:
Note that tests are failing because we just added Node 18 to the CI, easily fixed by merging main in your branch :)

test/integration/model/sync.test.js
Antamokhin: Yes. But I need some time to unravel and understand the logic of ```getForeignKeyReferencesForTable``` and redefinitions in dialects and subfunctions callstack. Because it looks like for MySQL dialect (that MariaDB uses in that case) ```getForeignKeyReferencesForTable``` in result calls ```getForeignKeysQuery(tableName, this.sequelize.config.database)``` from abstract dialect, but signature for MySQL dialect redefinition is ```getForeignKeysQuery(table, schemaName)```. And as a result, we have ```schemaName``` takes ```this.sequelize.config.database```. This is exactly what I wanted to show with my code. And I can't understand if there are some mistakes in my logic and I don't understand MySQL dialect or if there is an error in MySQL dialect code. This is just my helicopter view :) If you have any comments about this I will appreciate you:)

Antamokhin: I never used MariaDB, but https://mariadb.com/kb/en/schema/ said that schema and database are synonyms. Will prepare a new commit.

Antamokhin:
@ephys update has been pushed

src/model.js
Antamokhin: We have ```attribute.references.model = attribute.references.model.getTableName()``` that replace model to object.

test/integration/model/sync.test.js
```diff
@@ -403,6 +404,56 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
       });
+
+      if (sequelize.dialect.supports.schemas) {
+        it('should not recreate a foreign key if it already exists when { alter: true } is used with a custom schema', async () => {
+          const schema = 'test';
+          await sequelize.createSchema(schema);
+
+          const User = sequelize.define('User', {}, { schema });
+          const BelongsToUser = sequelize.define('BelongsToUser', {}, { schema });
+          BelongsToUser.belongsTo(User, {  foreignKey: { targetKey: 'id', allowNull: false } });
```
ephys, line 415:
```suggestion
          BelongsToUser.belongsTo(User, { foreignKey: { targetKey: 'id', allowNull: false } });
```

test/integration/model/sync.test.js
```diff
@@ -403,6 +404,56 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+          await sequelize.dropSchema(schema);
+        });
+
+        it('should not recreate a foreign key if it already exists when { alter: true } is used with a custom schema (reference attribute is a Model)', async () => {
+          const schema = 'test';
+          await sequelize.createSchema(schema);
+
+          class User extends Sequelize.Model {}
+
+          class BelongsToUser extends Sequelize.Model {}
```
ephys, line 432: we're moving away from using the Sequelize class as a namespace, it has poor typing capabilities:

```js
const { DataTypes, Sequelize, Model } = require('@sequelize/core');
```

```suggestion
          class User extends Model {}

          class BelongsToUser extends Model {}
```

test/integration/model/sync.test.js
```diff
@@ -403,6 +404,56 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+
+          User.init({
+            id: {
+              type: DataTypes.STRING,
+              primaryKey: true,
+            },
```
ephys, line 438: is there any reason for making this a string, instead of the default auto-incremented integer id?

test/integration/model/sync.test.js
```diff
@@ -403,6 +404,56 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+          }, { sequelize, schema });
+          BelongsToUser.init({
+            user_id: {
+              type: DataTypes.STRING,
```
ephys, line 442: Same question here: is this being a string part of this test?

test/integration/model/sync.test.js
```diff
@@ -403,6 +404,56 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+          const BelongsToUser = sequelize.define('BelongsToUser', {}, { schema });
+          BelongsToUser.belongsTo(User, {  foreignKey: { targetKey: 'id', allowNull: false } });
+          await sequelize.sync({ alter: true });
+          await sequelize.sync({ alter: true });
+
+          const results = await sequelize
+            .getQueryInterface()
+            .getForeignKeyReferencesForTable(BelongsToUser.getTableName());
+          expect(results).to.have.length(1);
+          await sequelize.dropSchema(schema);
```
ephys, line 423: Not necessary to drop this schema, the `afterEach` hook will handle it

test/integration/model/sync.test.js
```diff
@@ -1,150 +1,151 @@
...
-      const User = this.sequelize.define('testSync', {
+    it('should remove a column if it exists in the databases schema but not the model', async () => {
+      const User = sequelize.define('testSync', {
         name: DataTypes.STRING,
         age: DataTypes.INTEGER,
         badgeNumber: { type: DataTypes.INTEGER, field: 'badge_number' },
       });
-      await this.sequelize.sync();
-      this.sequelize.define('testSync', {
+      await sequelize.sync();
```
ephys, line 27: I highly appreciate you helping us modernize the codebase :)

ephys (changes requested):
I could reproduce a bug that I think is caused by checking whether `foreignReferenceSchema` is set:

```js
it('does not mismatch foreign keys when referencing identically named tables in different schemas', async () => {
  const schema = 'test';
  await sequelize.createSchema(schema);

  const Project = sequelize.define('Project', {}, { schema });

  const SameSchemaUser = sequelize.define('SameSchemaUser', {}, { schema, tableName: 'users' });

  // the goal of this test is to ensure the code takes into account the unspecified default schema:
  const ForeignSchemaUser = sequelize.define('ForeignSchemaUser', {}, { tableName: 'users' });

  Project.belongsTo(SameSchemaUser);
  Project.belongsTo(ForeignSchemaUser);

  await sequelize.sync({ alter: true });
  await sequelize.sync({ alter: true });

  const results = await sequelize
    .getQueryInterface()
    .getForeignKeyReferencesForTable(Project.getTableName());

  results.sort((a, b) => a.columnName.localeCompare(b.columnName));

  expect(results).to.have.length(2);

  expect(results[0]).to.include({
    columnName: 'ForeignSchemaUserId',
    // TODO: public schema could be different per DB
    referencedTableSchema: 'public',
    referencedTableName: 'users',
    referencedColumnName: 'id',
  });

  expect(results[1]).to.include({
    columnName: 'SameSchemaUserId',
    referencedTableSchema: 'test',
    referencedTableName: 'users',
    referencedColumnName: 'id',
  });

  await sequelize.dropSchema(schema);
});
```

I think the best way to resolve this could be to normalize [`this._schema` in Model.js line 878](https://github.com/sequelize/sequelize/blob/main/src/model.js#L878) - but the default schema name can be different per database. What do you think?

Antamokhin:
> I could reproduce a bug that I think is caused by checking whether `foreignReferenceSchema` is set:
> 
> ```js
> it('does not mismatch foreign keys when referencing identically named tables in different schemas', async () => {
>   const schema = 'test';
>   await sequelize.createSchema(schema);
> 
>   const Project = sequelize.define('Project', {}, { schema });
> 
>   const SameSchemaUser = sequelize.define('SameSchemaUser', {}, { schema, tableName: 'users' });
> 
>   // the goal of this test is to ensure the code takes into account the unspecified default schema:
>   const ForeignSchemaUser = sequelize.define('ForeignSchemaUser', {}, { tableName: 'users' });
> 
>   Project.belongsTo(SameSchemaUser);
>   Project.belongsTo(ForeignSchemaUser);
> 
>   await sequelize.sync({ alter: true });
>   await sequelize.sync({ alter: true });
> 
>   const results = await sequelize
>     .getQueryInterface()
>     .getForeignKeyReferencesForTable(Project.getTableName());
> 
>   results.sort((a, b) => a.columnName.localeCompare(b.columnName));
> 
>   expect(results).to.have.length(2);
> 
>   expect(results[0]).to.include({
>     columnName: 'ForeignSchemaUserId',
>     // TODO: public schema could be different per DB
>     referencedTableSchema: 'public',
>     referencedTableName: 'users',
>     referencedColumnName: 'id',
>   });
> 
>   expect(results[1]).to.include({
>     columnName: 'SameSchemaUserId',
>     referencedTableSchema: 'test',
>     referencedTableName: 'users',
>     referencedColumnName: 'id',
>   });
> 
>   await sequelize.dropSchema(schema);
> });
> ```
> 
> I think the best way to resolve this could be to normalize [`this._schema` in Model.js line 878](https://github.com/sequelize/sequelize/blob/main/src/model.js#L878) - but the default schema name can be different per database. What do you think?

Sorry for the long break.
I tried to tackle the problem a week ago, but failed. Bug looks really funny and strange for me:) It seems that I need to understand the code a little more. 
And what do you mean about normalization? Usage "public"-like schemas name that depends on db?

Do you think the changes in the current PR contain a bug/error related/dependent on this test? For me it's not obvious :(. Still, it seems to me that there is no problem of double synchronization. If we leave only one sync, then the bug will remain. Maybe it's worth highlighting this in a separate issue and thinking there?

P.S. And also I have a little problem to run test on M1 mac, only MariaDB container is ok :)

ephys (approved):
Actually the bug I posted isn't introduced by your PR. I'll fix it in a separate PR :)

jdboris:
Is v6 going to get this fix? I'm still getting duplicate foreign keys every sync in 6.28.0

ephys:
This PR targets the v7 branch, but we welcome PRs that backport this to v6

PR #14572, opened 2022-05-30
feat: support cyclic foreign keys (v7)

ephys:
This PR is a port of #14499 to v7.

On top of the existing v6 changes, this PR:

- migrates `integration/support.ts` to typescript, as the new tests have been ported to TS.
- fixes the typings for options that accept `deferrable` classes


ephys:
How weird, the following is not valid in DB2:

```sql
INSERT INTO "custom_schema"."a"
```

DB2 supports creating schema but not specifying in which schema the table resides?

WikiRik (commented):
If you've solved the DB2 issue we should be good to merge

ephys:
Oh I thought it was not valid because of the schema, but it's not valid because you need to specify at least one column when inserting with db2

ephys:
I "fixed" the first part of the issue by setting a default attribute (instead of its default value). We'll need to add a test to resolve this problem in a separate PR.

This PR uncovered two more bugs in the DB2 implementation:
- DB2's `QueryInterface#createTable` silently marks all attributes that are part of an index as 'unique' (even if the index is not unique).
- DB2's `QueryInterface#changeColumn` produces invalid SQL if the attribute is marked as unique.

ephys:
PR opened: https://github.com/sequelize/sequelize/pull/14619

WikiRik (approved):
We do need to check after the various PRs have been merged if everything is still fine with DB2 but this looks good

ephys:
I think I'll merge this one after the 2 db2 ones, and we should then expect db2 to be green with this one

WikiRik:
@ephys can you look into the merge conflicts? I believe this one is up next for merging

WikiRik (approved):

PR #14574, opened 2022-05-30
fix: kill connection on commit/rollback error (v7)

ephys:
This PR ports PR #14535 to v7

This PR is dependent on https://github.com/sequelize/sequelize/pull/14572 being merged first, as it needs to use `integration/support.ts`


ephys:
Blocked by #14652

test/integration/sequelize.transaction.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        case 'postgres':
+          query = 'select pg_sleep(2);';
+          break;
+        case 'sqlite':
+          query = 'select sqlite3_sleep(2000);';
+          break;
+        case 'mssql':
+          query = 'WAITFOR DELAY \'00:00:02\';';
+          break;
+        default:
```
WikiRik, line 97: Why not use 'select sleep(2);' here?

WikiRik (approved):

WikiRik (changes requested):
Again I was too early with approving

ephys:
The CLS change I did is causing weird behaviors, I'm postponing it to when we move to the built-in AsyncLocalStorage

WikiRik:
Hmm, something goes wrong and I don't think it's connected to this PR

test/integration/sequelize.transaction.test.ts
ephys: It was already here in the .js version but looking through the docs I think mssql does not support `sleep()`

ephys:
Aah, nice. ibm_db is fixed :)

WikiRik (dismissed):

test/integration/sequelize.transaction.test.ts
WikiRik: I get that, but why not do `let query;` and add the default query under default? That way it's more clear what the actual query is instead of it being overwritten

WikiRik: But this is not a blocking issue

ephys: Aaah you pointed at `default`. I agree, I'll change it

ephys: Done

WikiRik (approved):

PR #14578, opened 2022-05-30
feat: show composite foreign keys and filter `showTables` by schema

mike-usa:
### Pull Request Checklist

### Description Of Change

1. **feat(schemas): show all tables in target schemas**

    Changes include:
    - accept `schema` as an argument to perform lookup for tables in a particular schema
    - accept 'all', '*', or '%' values to also retrieve system tables

    Note: this also fixes a dev-testing bug, which was struggling to
    drop tables in other schemas because showAllTables was not complete

1.  **fix(foreign keys): include multi-column FKs and pull from schema tables**

    PostgreSQL:
     - foreign keys were limited to public schema tables. Now they
       can search tables across schemas
     - includes a fix to allow for composite keys and renamed field
       from `tableColumnName` to `tableColumnNames`. The value
       will always be an array

3. **test: number of changes, new tests, and fixes to lingering test issues**

    Includes several fixes and test updates:
    - fix spelling error throughout "quoteIdenifiers" to "quoteIdentifiers"
    - fix test due to timezone errors
    - no longer use dialect checking and now use feature checking
    
    New tests:
    - postgres schemas
    - create table
    - foreign key names


mike-usa:
I don't know if I can debug the DB2 errors, but I'm looking into the MariaDB, which I was kind of shocked by because the MySQL had passed.

ephys:
> but I'm looking into the MariaDB, which I was kind of shocked by because the MySQL had passed.

For a reason I don't quite understand, our implementations for MySQL and MariaDB differ when it comes to schemas even though they should not.
I think previous developers could not agree on what a schema is in MySQL/MariaDB because they're synonymous with "database", but they're not "databases" in the postgres/mssql sense, they're really just schemas:

- https://dev.mysql.com/doc/refman/8.0/en/create-database.html
- https://mariadb.com/kb/en/create-database/

So that means that, when creating a table 'user' in a 'test' schema, 

- Postgres will create ``` "test"."user" ```
- MariaDB will create ``` `test`.`user` ```
- MySQL will create ``` `test.user` ``` (it prefixes the table name)

MySQL is wrong here, MariaDB is right

It also means that the first parameter in `showTablesQuery` is actually the table schema, even though it's named `database`. Even MariaDB/MySQL's INFORMATION_SCHEMA table calls is `schema`:

image

This impacts your code, as you added a second parameter which contains the schema. It should be deduplicated. I think `showTablesQuery` should have the following signature in all dialects:

```typescript
// I'm using an option bag to be future proof
type ShowTablesOptions = {
  schemaName?: string | string[],
};

showTablesQuery(options?: ShowTablesOptions): string
```

But I'm warning you, it's a bit of a rabbit hole of wrong decisions that need to be unified. 😓

src/dialects/postgres/query-generator.js
```diff
@@ -124,8 +124,39 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
+
+    if (['all', '*', '%'].includes(settings.schema.name?.toLowerCase())) {
+      settings.schema.op = 'LIKE';
+      settings.schema.name = '%';
+    }
+
+    return Utils.toSingleLine(`
+      SELECT table_name as "tableName", table_schema as schema
+      FROM information_schema.tables
+      WHERE table_schema ${settings.schema.op} '${settings.schema.name}'
```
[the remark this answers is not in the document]
mike-usa: Great catch! I just saw a v6 security patch and I was like, this seems like SQL-injection, but I didn't put back in my notes to come back to it.

Is `.escape` satisfy all platforms? I have some MariaDb patches that I'll make the change in before I submit.

src/dialects/abstract/query-interface.js
```diff
@@ -330,7 +330,8 @@ export class QueryInterface {
       type: QueryTypes.SHOWTABLES,
     };
 
-    const showTablesSql = this.queryGenerator.showTablesQuery(this.sequelize.config.database);
+    // Snowflake is the only dialect that expects database name
```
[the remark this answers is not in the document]
mike-usa: I just discovered this when looking into MariaDb -- don't know how I missed that 🤷.  I think an object should be passed in with named keys, instead of ordinal arguments to accommodate the various platforms, which might make it more flexible.  Maybe that's a bad pattern.

src/dialects/postgres/query-generator.js
```diff
@@ -869,13 +900,161 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
-  getForeignKeysQuery(tableName) {
-    return 'SELECT conname as constraint_name, pg_catalog.pg_get_constraintdef(r.oid, true) as condef FROM pg_catalog.pg_constraint r '
-      + `WHERE r.conrelid = (SELECT oid FROM pg_class WHERE relname = '${tableName}' LIMIT 1) AND r.contype = 'f' ORDER BY 1;`;
+  getForeignKeysQuery(tableName, options) {
+    const tableDetails = this.extractTableDetails(tableName, options);
+    const canParseIdent = semver.lt(this.sequelize.options.databaseVersion, '9.6.0');
+
+    // remove comments as they would interfere in a single-line statement
+    // make string single-line to execute correctly
+    const SQL = Utils.toSingleLine(`
```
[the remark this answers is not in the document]
mike-usa: Raw SQL queries seem to execute prematurely (or break) at newline. I didn't look into it deeply. It was quicker to flatten the query.

mike-usa:
>I think previous developers could not agree on what a schema is in MySQL/MariaDB because they're synonymous with "database", but they're not "databases" in the postgres/mssql sense, they're really just schemas:
Exactly. I have a background in database architecture, but I started with Oracle, which PostgreSQL seemed heavily modeled after at the time.  MySQL treating schemas as a database is close to Postgres's implementation because there is a permission wall.  Postgres, though, has more concepts with subtle differences in their fluidity and encapsulation.  Pg haa s _cluster_, _database_, _tablespace_, and _schema_.  In other databases, tablespace acted like PostgreSQL schemas.

It might be helpful to list out the different container schemes and their attributes that Sequelize might care about.  Or build out an alias tree. Sequelize shouldn't care that MySQL calls their database schema; if they have similar behavior (attributes and features), they can be treated the same way under the hood.

Concept      | PostgreSQL | MariaDb | MySQL | MSSQL | DB2 
------------- | ------------ | ---------- | -------- | -------- | --- 
Cluster        |                   |                 |              |             |  
 ~ Permissions |                   |                 |              |             |  
Database    |                   |                 |              |             | 
 ~ Permissions |                   |                 |              |             |  
Tablespace |                   |                 |              |             | 
 ~ Permissions |                   |                 |              |             |  
Schema      |                   |                 |              |             | 
 ~ Permissions |                   |                 |              |             |  
Table          |                   |                 |              |             | 
 ~ Permissions |                   |                 |              |             |  
Synonym   |                    |                 |              |             |
 ~ Permissions |                   |                 |              |             |  



>So that means that, when creating a table 'user' in a 'test' schema,
> 
> Postgres will create "test"."user"
> MariaDB will create `test`.`user`
> MySQL will create `test.user` (it prefixes the table name)
> MySQL is wrong here, MariaDB is right
> 
> It also means that the first parameter in showTablesQuery is actually the table schema, even though it's named database. Even MariaDB/MySQL's INFORMATION_SCHEMA table calls is schema:

Wowser @ephys you found the bit that I patched. I wish I would have seen your comment first! I don't want to break anything so I'm taking the _schema_ set by the options arg and using that in MariaDb if it's provided; otherwise using the database name.

While these may be synonymous, in the code I'm still trying to keep them semantically different.  Where `database` is the connected database (not a foreign database) and `schema` is a targeted namespace.  The implementation in MariaDb just happens to be a different database, but that can be obfuscated from the user.

src/dialects/postgres/query-generator.js
```diff
@@ -869,13 +900,161 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
+      SELECT
+          main.contypid::regclass  AS domain,
+
+          -- TABLE INFO
+          schema.nspname             AS "tableSchema",
+          -- strip off the schema prefix
+          --   NOTE: parse_ident available in 9.6 and later
```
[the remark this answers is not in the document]
mike-usa: That's going to simplify this a lot.  Also, it's not like `parse_ident` isn't available in 9.6.

test/integration/query-interface.test.js
```diff
@@ -53,7 +54,7 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
       await this.sequelize.query(sql);
       let tableNames = await this.queryInterface.showAllTables();
       await cleanup(this.sequelize);
-      if (tableNames[0] && tableNames[0].tableName) {
+      if (tableNames?.[0]?.tableName) {
```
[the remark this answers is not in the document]
mike-usa: The operator applies to the following property/value.  
In this case:
```javascript
tableNames?.[0]
```
Says "if tableNames has an element at index 0" (e.g., `[ {nestedObject} ]` vs `[ ]`) and then

```javascript
tableNames?.[0]?.tableName
```
says if first element contains a property called `tableName`, use it, otherwise return `undefined`.

test/integration/query-interface.test.js
```diff
@@ -454,29 +456,60 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
         },
       });
     });
+    afterEach(async function () {
+      await this.queryInterface.dropTable('hosts');
```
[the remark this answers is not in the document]
mike-usa: Is the goal to get off the global or would you prefer to rely on it? I thought your other post suggested doing it in chunks, starting by removing `dropAllDatabase` from the global and placing it in each file. Isn't the eventual state to have each test be explicit in what it builds and tears down (like this) or do you have a different vision in mind?

test/integration/support.js
```diff
@@ -11,7 +11,11 @@ const CLEANUP_TIMEOUT = Number.parseInt(process.env.SEQ_TEST_CLEANUP_TIMEOUT, 10
 
 let runningQueries = new Set();
 
+const _OriginalTz = process.env.TZ;
+
 before(function () {
+  // set code timezone to same as database to avoid TZ errors when comparing dates
+  process.env.TZ = 'UTC';
```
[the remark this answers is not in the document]
mike-usa: This is something I didn't know much about.  So the database is in UTC, but I think my node environment (e.g., `process.env`) was in EST. The difference between timezones broke datetime tests.  I initially placed it closer to the tests, but I thought it might be a better general dev configuration to be at the higher (highest) level.

test/integration/dialects/postgres/schema.test.js
```diff
@@ -0,0 +1,116 @@
...
+      });
+
+      it.skip('prevents schema-privileged user from creating public tables', async () => {
+        expect(() => {
+          this.restrictedUserQI.createTable(
+            'shouldnt_create',
+            { name: DataTypes.STRING },
+            { schema: 'public' },
+          );
+        }).to.throw(); // or to.be.rejected
```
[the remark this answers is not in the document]
mike-usa: Per your comments on Slack, I've updated this and feel so much better that these no longer have to be skipped :) The lines can be improved.  I'm using `rejectedWith(Error)`, but wish I knew what error so it could be more explicit.

test/integration/dialects/postgres/schema.test.js
```diff
@@ -0,0 +1,116 @@
...
+      /** create a new user that has restricted privileges, isolated to schema */
+      const restricted = {
+        username: 'restricted_user',
+        password: 'restricted_password',
+      };
+
+      beforeEach(async () => {
+        // create a new user that will have restricted privileges
+        await sequelize.query(`DROP USER IF EXISTS ${restricted.username};`);
+        await sequelize.query(`CREATE USER ${restricted.username} WITH PASSWORD '${restricted.password}';`);
```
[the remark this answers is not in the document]
mike-usa: @ephys could you elaborate?  I'm not sure what you mean by a replacement.

test/integration/dialects/postgres/schema.test.js
```diff
@@ -0,0 +1,116 @@
...
+        // Create table in schema
+        await this.restrictedUserQI.createTable(
+          'my_test_table',
+          { name: DataTypes.STRING },
+          { schema: 'testschema' },
+        );
+
+        // Retrieve tables in schema
+        let tableNames = await this.restrictedUserQI.showAllTables({ schema: 'testschema' });
+        if (tableNames?.[0]?.tableName) {
```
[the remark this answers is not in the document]
mike-usa: I'm thinking this relates to your other comment that sometimes tables names are returned as `schema.name` (really, no schema) and other times it's an object. I don't know if sqlite was one of the offenders.  Part of it was borrowing from other tests, but part of it was the fact-finding expedition into what `showAllTables` was doing.  I can review if you'd like.  I agree it should be standardized and consistent.

src/dialects/postgres/query-generator.js
```diff
@@ -124,8 +124,39 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
+
+    if (Array.isArray(settings.schema.name)) {
+      settings.schema.op = 'SIMILAR TO';
+      settings.schema.name = `(${settings.schema.name.join('|')})`;
+    }
+
+    if (['all', '*', '%'].includes(settings.schema.name?.toLowerCase())) {
+      settings.schema.op = 'LIKE';
+      settings.schema.name = '%';
+    }
```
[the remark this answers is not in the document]
mike-usa: Good idea! Can we create this as a new issue?  My PR is starting to get longer :)

WikiRik:
I'm starting to feel that the scope of this PR is not clear anymore. Can you move the support for postgres 12 and the test commit to different PRs?
And probably also the showing of the composite foreign keys could go to another PR. If the PR title contains 'and' it's a good indicator that you can split things up 😊

But thanks for working on it, we appreciate it!

src/dialects/postgres/query-interface.js
```diff
@@ -160,7 +160,7 @@ export class PostgresQueryInterface extends QueryInterface {
     const query = this.queryGenerator.getForeignKeyReferencesQuery(table.tableName || table, this.sequelize.config.database);
     const result = await this.sequelize.queryRaw(query, queryOptions);
 
-    return result.map(Utils.camelizeObjectKeys);
+    return result.map(Utils.camelizeKeys);
```
ephys, line 163: What is the reason behind this change? I would expect this method to continue returning an array of objects, so the previous code should be enough

src/dialects/abstract/query-interface.js
```diff
@@ -612,13 +611,21 @@ export class QueryInterface {
    * @returns {Promise}
    */
   async getForeignKeysForTables(tableNames, options) {
+    const generator = this.queryGenerator;
+    tableNames = Array.from([tableNames]).flat(); // force array
```
ephys, line 615: Consider using the `toArray` util we talked about, I see this code very often in your PR

src/dialects/abstract/query-interface.js
```diff
@@ -641,8 +648,8 @@ export class QueryInterface {
    * Get foreign key references details for the table
    *
    * Those details contains constraintSchema, constraintName, constraintCatalog
-   * tableCatalog, tableSchema, tableName, columnName,
-   * referencedTableCatalog, referencedTableCatalog, referencedTableSchema, referencedTableName, referencedColumnName.
+   * tableCatalog, tableSchema, tableName, tableColumnNames,
```
ephys, line 651: columns can only be in tables, I think this could be named `columnNames` and `referencedColumnNames`

test/integration/query-interface/removeColumn.test.js
```diff
@@ -54,7 +54,7 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
         expect(table).to.not.have.property('firstName');
       });
 
-      it('should be able to remove a column without default value', async function () {
+      it('should be able to remove a column without a default value', async function () {
```
ephys, line 57: consider moving this non-controversial change to another PR as a way to lighten this (pretty big) one

src/dialects/db2/query-generator.js
```diff
@@ -49,8 +45,12 @@ export class Db2QueryGenerator extends AbstractQueryGenerator {
   }
 
   showSchemasQuery() {
-    return 'SELECT SCHEMANAME AS "schema_name" FROM SYSCAT.SCHEMATA WHERE '
-      + '(SCHEMANAME NOT LIKE \'SYS%\') AND SCHEMANAME NOT IN (\'NULLID\', \'SQLJ\', \'ERRORSCHEMA\')';
+    return Utils.toSingleLine(`
```
ephys, line 48: `support.ts` has a function called `minifySql`. I think we could move it to utils instead of `toSingleLine`

src/dialects/db2/query-generator.js
```diff
@@ -751,59 +763,114 @@ export class Db2QueryGenerator extends AbstractQueryGenerator {
...
+   * @returns {string} The generated sql query.
    */
-  getForeignKeysQuery(table, schemaName) {
-    const tableName = table.tableName || table;
-    schemaName = table.schema || schemaName;
-    let sql = '';
-    if (tableName) {
-      sql = ` AND R.TABNAME = ${wrapSingleQuote(tableName)}`;
-    }
+  getForeignKeysQuery(table, _database, options) {
```
ephys, line 819: this should continue to be `getForeignKeysQuery(table, schemaName)`

The second parameter (used in MariaDB) is the schema name too, no need to pass it through options

src/dialects/db2/query-generator.js
```diff
@@ -751,59 +763,114 @@ export class Db2QueryGenerator extends AbstractQueryGenerator {
...
+      whereConditions.push(` AND R.TABSCHEMA = ${wrapSingleQuote(schemaName)}`);
     }
 
     if (columnName) {
-      sql += ` AND C.COLNAME = ${wrapSingleQuote(columnName)}`;
+      whereConditions.push(` AND C.COLNAME = ${wrapSingleQuote(columnName)}`);
     }
 
-    return this._getForeignKeysQuerySQL(sql);
+    return this._getForeignKeysQuerySQL(whereConditions.join(''));
```
ephys, line 873: I don't think this change is necessary, it does exactly the same thing as before and this PR is big enough already

src/dialects/postgres/query-generator.js
```diff
@@ -126,12 +126,12 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
-   * @param {object}    [options] Options
-   * @param {string}    [options.schema] Schema name to search for (defaults to 'public')
-   * @returns {string}  SQL statement to show all tables, including tableName and schema
+   * @param {*} _database Unused variable (database name)
+   * @param {object}      [options] Options
+   * @param {string}      [options.schema] Schema name to search for (defaults to 'public')
+   * @returns {string}    SQL statement to show all tables, including tableName and schema
    */
-  showTablesQuery(_, options) {
+  showTablesQuery(_database, options) {
```
ephys, line 134: I'd rather avoid complexifying our already complex API: the first parameter and the second parameter do the same thing, it was just incorrectly named 'database' in mariadb because it's the same thing in that dialect

Consider one of the two interfaces instead:

```typescript
showTablesQuery(schema?: string): string;
// or
showTablesQuery(options?: { schema?: string }): string;
```

src/dialects/db2/query-generator.js
```diff
@@ -820,7 +887,9 @@ export class Db2QueryGenerator extends AbstractQueryGenerator {
   }
 
   dropForeignKeyQuery(tableName, foreignKey) {
-    return _.template('ALTER TABLE <%= table %> DROP <%= key %>', this._templateSettings)({
+    const sql = 'ALTER TABLE <%= table %> DROP FOREIGN KEY <%= key %>;';
+
+    return _.template(sql, this._templateSettings)({
```
ephys, line 892: if you're touching this code, could you replace `_.template` with a regular template string?

src/dialects/mariadb/query-generator.js
```diff
@@ -25,6 +26,32 @@ export class MariaDbQueryGenerator extends MySqlQueryGenerator {
...
+    if (databaseVersion && semver.lt(this.sequelize.options.databaseVersion, '10.5.0')) {
+      SQL = SQL.replace(/JSON_ARRAYAGG/g, 'GROUP_CONCAT');
+    }
+
+    return SQL;
+  }
+
+  getForeignKeyQuery(table, columnName) {
+    // retrieve SQL from `super` (MySQL)
+    let SQL = super.getForeignKeyQuery(table, columnName);
```
ephys, line 44:
```suggestion
    let sql = super.getForeignKeyQuery(table, columnName);
```

src/dialects/db2/query-generator.js
```diff
@@ -181,8 +181,20 @@ export class Db2QueryGenerator extends AbstractQueryGenerator {
     });
   }
 
-  showTablesQuery() {
-    return 'SELECT TABNAME AS "tableName", TRIM(TABSCHEMA) AS "tableSchema" FROM SYSCAT.TABLES WHERE TABSCHEMA = USER AND TYPE = \'T\' ORDER BY TABSCHEMA, TABNAME';
+  showTablesQuery(_database, options) {
```
ephys, line 184: Same here, unify `database` (mariadb, but is actually schema) with `options.schema` into either of these two signatures:

```ts
showTablesQuery(schema?: string): string;
// or
showTablesQuery(options?: { schema?: string }): string;
```

src/dialects/postgres/query-generator.js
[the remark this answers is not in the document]
ephys: This would reduce the size of your PR actually, since it touches a change you've introduced here

src/dialects/postgres/query-generator.js
```diff
@@ -869,38 +900,170 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
+   *                                    object: { tableName: Name of table, schema: name of schema }
+   * @param   {object}        options   Pass to extractTableDetails
+   * @returns {string}                  The generated sql query.
    * @private
    */
-  getForeignKeysQuery(tableName) {
-    return 'SELECT conname as constraint_name, pg_catalog.pg_get_constraintdef(r.oid, true) as condef FROM pg_catalog.pg_constraint r '
-      + `WHERE r.conrelid = (SELECT oid FROM pg_class WHERE relname = '${tableName}' LIMIT 1) AND r.contype = 'f' ORDER BY 1;`;
-  }
+  getForeignKeysQuery(tableArg, options) {
```
ephys, line 909: IMO this function should not need a second argument, no need for `extractTableDetails`.

Instead, it should have the following signature:

```ts
getForeignKeysQuery(tableName: TableName): string;
```

`TableName` can be a string, or an object which includes the `tableName` and `schema` properties

If you want to filter by column too, pass columns in the option bag (not in the table name):

```ts
type GetForeignsKeyQueryOptions = {
  columns?: string[]
}

getForeignKeysQuery(tableName: TableName, options?: GetForeignsKeyQueryOptions): string;
```

src/dialects/postgres/query-generator.js
```diff
@@ -869,38 +900,170 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
+              "tableColumnNames" && {${this.escape(columnName)}}::text[]
+      ` : ''}
+
+      ORDER BY
+              "domain",
+              "tableSchema",
+              "tableName",
+              "constraintName"
+      ;
+    `.replace(/--.*/g, ''); // remove comments as they would interfere in a single-line statement
```
ephys, line 1040: removing comments can be done in `minifySql` instead (`minifySql` being the replacement for `toSingleLine` I proposed in another comment)

src/dialects/postgres/query-generator.js
```diff
@@ -913,19 +1076,18 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
    */
   getForeignKeyReferencesQuery(tableName, catalogName, schemaName) {
-    return `${this._getForeignKeyReferencesQueryPrefix()
-    }WHERE constraint_type = 'FOREIGN KEY' AND tc.table_name = '${tableName}'${
-      catalogName ? ` AND tc.table_catalog = '${catalogName}'` : ''
-    }${schemaName ? ` AND tc.table_schema = '${schemaName}'` : ''}`;
+    return this.getForeignKeysQuery({ tableName, schema: schemaName });
   }
 
   getForeignKeyReferenceQuery(table, columnName) {
```
ephys, line 1082: This method is never used and can be removed

src/dialects/postgres/query-generator.js
```diff
@@ -913,19 +1076,18 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
    * @param {string} schemaName
    */
   getForeignKeyReferencesQuery(tableName, catalogName, schemaName) {
-    return `${this._getForeignKeyReferencesQueryPrefix()
-    }WHERE constraint_type = 'FOREIGN KEY' AND tc.table_name = '${tableName}'${
-      catalogName ? ` AND tc.table_catalog = '${catalogName}'` : ''
-    }${schemaName ? ` AND tc.table_schema = '${schemaName}'` : ''}`;
+    return this.getForeignKeysQuery({ tableName, schema: schemaName });
```
ephys, line 1079: This method is only used by `PostgresQueryInterface#getForeignKeyReferencesForTable`. It would be interesting to look into whether that method and this one can be removed

src/dialects/sqlite/query-generator.js
```diff
@@ -303,8 +304,10 @@ export class SqliteQueryGenerator extends MySqlQueryGenerator {
           const referencesTable = this.quoteTable(dataType.references.model);
 
           let referencesKey;
-          if (dataType.references.key) {
-            referencesKey = this.quoteIdentifier(dataType.references.key);
+          const foreignKeys = Array.from([dataType.references.key]).flat().filter(v => !!v);
```
ephys, line 307:
```suggestion
          const foreignKeys = Array.from([dataType.references.key]).flat().filter(Boolean);
```

src/dialects/sqlite/query-interface.js
```diff
@@ -155,18 +155,46 @@ export class SqliteQueryInterface extends QueryInterface {
...
+    const foreignKeys = await this.sequelize.queryRaw(query, queryOptions);
+
+    // field name translation
+    const changeKey = new Map();
+    changeKey.set('from', 'tableColumnNames');
+    changeKey.set('table', 'referencedTableName');
+    changeKey.set('to', 'referencedTableColumnNames');
+
+    // convert Column field values to array
+    Array.from([foreignKeys]).flat().forEach(foreignKeyObject => {
```
ephys, line 174: this is a lot more complex than what we previously had. Any reason for using this code over the previous one?

src/dialects/sqlite/query-interface.js
```diff
@@ -231,10 +259,12 @@ export class SqliteQueryInterface extends QueryInterface {
 
       const foreignKeys = await this.getForeignKeyReferencesForTable(tableName, options);
       for (const foreignKey of foreignKeys) {
-        data[foreignKey.columnName].references = {
-          model: foreignKey.referencedTableName,
-          key: foreignKey.referencedColumnName,
-        };
+        foreignKey.tableColumnNames.forEach(fk => {
```
ephys, line 262: use for-of instead of .forEach

src/model.js
```diff
@@ -1284,14 +1282,19 @@ ${associationOwner._getAssociationDebugList()}`);
...
             const database = this.sequelize.config.database;
             const schema = this.sequelize.config.schema;
-            // Find existed foreign keys
+
+            // Find existing foreign keys
             for (const foreignKeyReference of foreignKeyReferences) {
               const constraintName = foreignKeyReference.constraintName;
+
+              // TODO: fix src/associations/helpers.ts::addForeignKeyConstraints to accept composite keys
+              const keys = new Set(Array.from([references.key]).flat());
```
ephys, line 1291: create this map outside of the for-of loop, to avoid re-creating it each time

test/integration/model/geometry.test.js
```diff
@@ -32,7 +36,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         await Pub.sync({ force: true });
         const pub = await Pub.create({ location: point });
         expect(pub).not.to.be.null;
-        expect(pub.location).to.be.deep.eql(point);
+        expect(pub.location).to.deep.include(point);
```
ephys, line 39: this shouldn't have changed. There shouldn't be additional properties

It also seems very unrelated to this PR. Can you revert all changes to this file and, if needed, move to another PR?

test/integration/query-interface.test.js
```diff
@@ -11,8 +11,9 @@ const current = Support.sequelize;
 const _ = require('lodash');
 
 describe(Support.getTestDialectTeaser('QueryInterface'), () => {
+
   beforeEach(function () {
-    this.sequelize.options.quoteIdenifiers = true;
+    this.sequelize.options.quoteIdentifiers = true;
```
ephys, line 16: Could you look into moving this `quoteIdentifiers` change to another PR? To lighten this PR

test/integration/query-interface.test.js
```diff
@@ -455,30 +473,81 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
-        expect(Object.keys(foreignKeys[2])).to.have.length(7);
-      } else if (['sqlite', 'db2'].includes(dialect)) {
-        expect(Object.keys(foreignKeys[0])).to.have.length(8);
-      } else if (dialect === 'ibmi') {
-        expect(Object.keys(foreignKeys[0])).to.have.length(9);
-      } else if (['mysql', 'mariadb', 'mssql'].includes(dialect)) {
-        expect(Object.keys(foreignKeys[0])).to.have.length(12);
-      } else {
-        throw new Error(`This test doesn't support ${dialect}`);
+      let firstForeignKeyProperties = Object.keys(foreignKeys[0]);
```
ephys, line 505: use `const` as much as possible, avoid re-assigning the same variable when they serve different purposes

test/integration/query-interface/changeColumn.test.js
```diff
@@ -215,8 +214,8 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
 
           const newForeignKeys = await this.queryInterface.getForeignKeyReferencesForTable('users');
           expect(firstForeignKeys.length).to.be.equal(newForeignKeys.length);
-          expect(firstForeignKeys[0].columnName).to.be.equal('level_id');
-          expect(firstForeignKeys[0].columnName).to.be.equal(newForeignKeys[0].columnName);
+          expect(firstForeignKeys[0].tableColumnNames).to.eql(['level_id']);
```
ephys, line 217: `.eql` is not declared in chai's typings

```suggestion
          expect(firstForeignKeys[0].tableColumnNames).to.equal(['level_id']);
```

test/integration/query-interface/getForeignKeyReferencesForTable.test.js
```diff
@@ -41,5 +44,86 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+      Book.belongsTo(Author);
+      Book.belongsToMany(Critic, { through: 'Reviews' });
+      Critic.belongsToMany(Book, { through: 'Reviews' });
+
+      const options = { force: true };
+
+      await Author.sync(options);
+      await Publisher.sync(options);
+      await Book.sync(options);
+      await Critic.sync(options);
```
ephys, line 66: any reason for not using this instead?

```suggestion
      await this.sequelize.sync({ force: true });
```

test/integration/query-interface/getForeignKeyReferencesForTable.test.js
```diff
@@ -41,5 +44,86 @@ describe(Support.getTestDialectTeaser('QueryInterface'), () => {
...
+        referencedTableName: 'Authors',
+      })).to.be.true;
+      expect(dataContainsSubset(bookRefs, {
+        tableColumnNames: ['PublisherId'],
+        referencedTableColumnNames: ['id'],
+        referencedTableName: 'Publishers',
+      })).to.be.true;
+
+      // same query called with different argument format
+      bookRefs = await this.queryInterface.getForeignKeyReferencesForTable('Books');
```
ephys, line 103: avoid reassigning to this variable, use `const`

test/integration/support.js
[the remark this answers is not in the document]
ephys: My timezone isn't UTC either and the tests aren't breaking on my end. That's strange.

But I don't see what uses that variable. It's not referenced anywhere else in our codebase and isn't a standard node env variable.

ephys (changes requested):
As with the previous review, I haven't reviewed everything yet. This is a huge PR that changes lots of things.

I concur with @WikiRik, some things should be moved to their own PR so we can properly focus on this (large) change. I think everything that was changed in /dev and package.json are better suited for a separate PR

This PR introduces some breaking changes in our APIs so it's important we get their API design right

Don't hesitate to click "re-request review" once you're ready for me to review this again :)

src/dialects/abstract/query-interface.js
mike-usa: I respectfully challenge :) I gave this some careful consideration (because it'd be a chore to update).  I wish we had a Slack channel (other than `sequelize-help`) where this kind of stuff can be discussed before begin worked on.

It's important to point out there are three changes taking place here:
1. the names are pluralized to accommodate an array (for composite keys)
2. `table` is added to both `columnNames` columns
3. `table` is added to all targeted table parts (it pre-existed on _catalog_ and _schema_)

My conclusions:
1. It makes sense to logically group the elements with a word (or absence of a word) to distinguish the "from" and "to" connections. 
   - The other objects were all qualified with "table", "referenced", or "constraint", leaving certain elements floating and not part of those buckets.  By not qualifying `columnName`, it makes it slightly confusing because it does not follow the pattern.
   - I'd argue if `table` is dropped from `tableColumnNames`, then it should be dropped from `tableCatalog` and `tableSchema` as well for consistency
1. A bit of preference/opinion, but when sorting/grouping it is much nicer to have all the parts sorted alphabetically and grouped together.  To reiterate a former point, this makes it easier to read and work with the code

Sidenotes:
1. I also wanted to change "referenced" to present tense, but that was going too far 😆
1. My understanding is this doesn't translate into the ORM and only used internally (or as JSON results), as an ORM property, I would add the `columnNames` alias to point to `tableColumnNames`

.github/workflows/ci.yml
```diff
@@ -191,7 +191,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [14, 18]
-        postgres-version: [9.5, 10]
+        postgres-version: [10, 12]
```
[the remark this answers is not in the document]
mike-usa: It should.  My understanding of GH workflows (and Matrix) is weak.  I thought the workflow would only run the listed version -- if I place `14` does it have its own instance of Postgres 14?

I only looked for database containers by the previous author to keep the scripts simple and because I'm also not a Sequelize/Docker expert yet -- I'm uncertain what configurations he did that differ from the official Postgres images.  I didn't find a Postgres 14 container by the same author, I only saw up to v12.  My thinking here was v14 could be added at a later point.

I imagine in an ideal world, each database version Sequelize supports would be included, but these tests are lengthy and it's good to limit.  For that reason, I submit that the minimum and latest versions should definitely be included and an optional popular database in between.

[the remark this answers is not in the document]
WikiRik: Yeah, we need to implement our own Docker scripts for postgres 14 so we're not dependent on Sushant. Let's add the support for postgres 12 in one PR and later one for postgres 14

src/dialects/abstract/query-generator.js
```diff
@@ -47,12 +47,24 @@ export class AbstractQueryGenerator {
     this._initQuoteIdentifier();
   }
 
+  defaultSchema(dialect = this.options.dialect) {
```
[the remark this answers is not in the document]
mike-usa: That is a great suggestion!

I looked at placing it elsewhere, but I'd need assistance. This requires a reference to the Sequelize object to access the connected config to retrieve username/databaseName details.  I'm not sure how that would be done.

Since Sequelize is not imported in the other places, I thought it could be an additional (unnecessary) performance hit to load/instantiate another library there.  I may be wrong on that front, obviously I need help 😄

src/dialects/abstract/query-interface.js
mike-usa: Can we look the other way on this? 😏 

Years ago, I used a project that used _underscore_ and then they switched to lodash, which wasn't a huge hit, but was a little work as not everything translated. Later on, they wanted to reduce their library dependency footprint and remove lodash -- you can imagine how that went, since it was integrated into their code.

I hesitate to use lodash where it can be avoided and encourage others to either create wrapper functions or create their custom functions for schemes like the above. But that's my preference and I can make the change if that's your preference and I haven't convinced you.

src/dialects/db2/query-generator.js
mike-usa: That would be great! I saw the SQLFragments, but it didn't help here.  This is more a DX problem it solves.  I find well-formatted SQL to be much easier to read and identify problems and performance issues.

src/dialects/db2/query-generator.js
mike-usa: this may benefit from the dialect `defaultSchema` call.  I imagine that could be a future refactor commit/PR.

src/dialects/db2/query-generator.js
mike-usa: I think _options_ should always be passed.  I'm came across these changes because CLI migration was not passing along the schema from the config.  I think I had to supply options.

Is there a reason why this couldn't be:
```javascript
getForeignKeysQuery(table, schemaName, options)
```

I'll still have to modify the code to not rely on options, I just don't want to remove it from the equation entirely. We also have some opportunity to fix the taxonomy.

I'm using `table` here because it can be an object or string -- but even that isn't good, I just don't know what's better.  In other places `tableName` is used, which is more specific and I would expect it to be a string, but it can be the object as well.

This is where TypeScript is good because it informs you on the arg types, but even these strings could have schema qualifiers (e.g., `public.users`). So there's multiple equivalent table configurations:
- `users`
- `public.users`
- `{ tableName: 'users' }`
- `{ tableName: 'users', schema: 'users' }`

We see `tableName` is used as a key in the object that points to a string, which is partly why I was confused when I saw it a parameter name that referenced the entire object.

src/dialects/db2/query-generator.js
mike-usa: It's not _necessary_.  It's a general-practice performance tweak when working with strings.

src/dialects/db2/query-generator.js
mike-usa: 😄 I debated that but I thought it was left for a reason (SQL injection or something).  Yes I can update it.

src/dialects/abstract/query-generator.js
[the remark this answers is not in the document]
ephys: You can access the dialect through `this._dialect`. It's already present so no extra performance hit there.

The dialect also already has access to the sequelize instance (example, db2: https://github.com/sequelize/sequelize/blob/main/src/dialects/db2/index.js#L32). So a `getDefaultSchema` method in these classes wouldn't have any perf impact either

src/dialects/abstract/query-interface.js
ephys: It's not a lodash util (I'm also removing lodash everywhere it's not needed anymore), it's a function we can add to our utilities, see here: https://github.com/sequelize/sequelize/pull/14578#discussion_r885318955

It has better performance (doesn't need to instanciate a few arrays and iterate them to flatten), and is more explicit. Given you've used this `Array.from([tableNames]).flat();` code a few times, I think the util is much needed for clarity

src/dialects/db2/query-generator.js
mike-usa: I think the issue is that the abstract `QueryInterface.getForeignKeysForTables()` passes databaseName to `getForeignKeysQuery`.  That QI function is a headache.  Maybe each dialect needs its own or instead of passing the databaseName, it should pass the `defaultSchema` created as pat of this commit?

src/dialects/abstract/query-interface.js
ephys: > the names are pluralized to accommodate an array

This change I'm 100% on board with, pluralize away

> The other objects were all qualified with "table", "referenced", or "constraint"

Not fully true, it's prefixed with `referenced` if it's the target element, and not prefixed otherwise (`tableCatalog` vs `referencedTableCatalog`, `tableSchema` vs `referencedTableSchema`, `tableName` vs `referencedTableName`, `columnName` vs `referencedColumnName`): it's always the exact same name + 'referenced'

I think for this PR, the change should be limited to only pluralizing the relevant names, we can look into having better names in a follow-up PR.

I would propose:

```
sourceCatalog
sourceSchema
sourceTable
sourceColumns

&

referencedCatalog
referencedSchema
referencedTable
referencedColumns
```

> only used internally

`queryInterface` is a public API (with some design issues), so API changes here must be considered breaking changes

src/dialects/postgres/query-generator.js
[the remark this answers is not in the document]
mike-usa: It's important to note, where I use `Array.from().flat()`, the flattening is important.  I'd still be using `toArray().flat()`. I'm starting to question what this might buy.

```javascript
const arr = [
  [
    { users: ['fk1','fk2'] }, 
    { admins: ['fk1', 'fk2'] }
  ],
  { meta: 'stuff'}
];

toArray(arr)         // unchanged: [ [ { users: ['fk1','fk2'] }, { admins: ['fk1','fk2'] } ], { meta: 'stuff' } ]
Array.from(arr)    // unchanged: [ [ { users: ['fk1','fk2'] }, { admins: ['fk1','fk2'] } ], { meta: 'stuff' } ]

toArray(arr).flat()  // desired:  [ { users: [ 'fk1', 'fk2' ] }, { admins: [ 'fk1', 'fk2' ] }, { meta: 'stuff' } ]
Array.from(arr).flat()  // desired:  [ { users: [ 'fk1', 'fk2' ] }, { admins: [ 'fk1', 'fk2' ] }, { meta: 'stuff' } ]
```

I'm not sure what `toArray` buys from the native `Array.from` and I'm not sure what it would clean up.

src/dialects/db2/query-generator.js
ephys: > I'm came across these changes because CLI migration was not passing along the schema from the config

This shouldn't prevent the fix from being made. The caller needs to provide the schema name, it shouldn't matter whether it's passed as part of the table name, or as part of option

> Is there a reason why this couldn't be `getForeignKeysQuery(table, schemaName, options)`

Just API design. Our other methods that accept a table name pass the schema as part of the table name (using the `TableName` object). Being able to pass the schema in 3 different ways (as part of the table name, as part of the second parameter, or in options) in very confusing. We should have a clean API, and the other places that call this must make sure they're calling it properly.

The issue with the configured default schema not being passed is separate. We need to normalize the `table` parameter at the beginning of this method. You could do it by creating this new method and calling it with the `table` param:

```typescript
normalizeTableName(tableIdentifier: TableName): TableNameWithSchema {
  if (typeof tableIdentifier === 'string') {
    // normalize to TableNameWithSchema
    tableIdentifier = { tableName: tableIdentifier };
  }

  if (!tableIdentifier.schema) {
    // user provided default value
    if (this.options.schema) {
      tableIdentifier.schema = this.options.schema;
    } else {
       tableIdentifier.schema = this._dialect.getDefaultSchema();
    }
  }

  return tableIdentifier;
}
```

then at the beginning of you method, you can do

```
tableIdentifier = normalizeTableName(tableIdentifier);
```

And it will be much easier to manage, and will have the correct schema name from the config (if set, default schema per dialect otherwise) :)

> I'm using table here because it can be an object or string -- but even that isn't good

`tableName` or `table` are both good. `tableIdentifier` is ok too if you want. Technically the schema is part of the table identifier. This will be clearer once we move to TypeScript

>  but even these strings could have schema qualifiers

They can't, if you pass `public.user`, it will be treated as `"public"."public.user"`. The only way to pass a schema is by doing `{ tableName: 'users', schema: 'users' }`. Passing `users` is strictly equivalent to passing `{ tableName: 'users' }`

src/dialects/postgres/query-generator.js
mike-usa: @ephys are you proposing to move `minifySql` to Utils class instead of Support? It's currently only available in the _test_ environment.

src/dialects/postgres/query-generator.js
mike-usa: There seems to be a lot of redundancy around constraints.  It's easier for me to think of retrieving foreignKeys and filtering through code from there.  I don't know about datawarehouses that might have tons of keys, but I don't think it'd be a large impact or much additional info to process locally.

src/dialects/postgres/query-generator.js
mike-usa: It's only used in the `getForeignKeyReferencesForTable` in the QueryInterface.

src/dialects/postgres/query-interface.js
mike-usa: The new function is recursive to traverse the argument.  `camelizeObjectKeys` doesn't camelize nested objects (or a nested array of objects).  

It handles a simple array of objects well using `.map(camelizeObjectKeys)`, but for anything more complex, there are gaps.  `camelizeObjectKeys` is also more efficient because it doesn't have the additional function calls, which is why it deserves to live on.

This:
```json
[
  [
    {
      "table_name": "stores",
      "table_column_names": ["customer_name", "customer_phone"],
      "references_table_name": "customers",
      "references_table_column_names": ["customer_name", "customer_phone"]
    },
    {
      "table_name": "cars",
      "table_column_names": ["part_id"],
      "references_table_name": "parts",
      "references_table_column_names": ["id"]
    }
  ],
  {
    "meta": {

    }
  }
]
 ```
Becomes:
```json
[
  [
    {
      "tableNames": "stores",
      "tableColumnNames": ["customer_name", "customer_phone"],
      "referencesTable": "customers",
      "referencesTableColumnNames": ["customer_name", "customer_phone"]
    },
    {
      "tableNames": "cars",
      "tableColumnNames": ["part_id"],
      "referencesTable": "parts",
      "referencesTableColumnNames": ["id"]
    }
  ],
  {
    "meta": {

    }
  }
]
 ```
 
I initially intended to camelize the values too, but I changed my mind as I thought this information should be raw; for instance, if building an ERD, you want to see the actual values stored in the DB.  If working through the ORM, you want to work with camelized values and that should happen elsewhere.

File: src/utils/object.ts [hunk 39]
```diff
@@ -216,6 +216,56 @@ export function camelizeObjectKeys(obj: { [key: string]: any }) {
...
+ * camelizeKeys recursively turns object keys camelCased, works on
+ *   arrays of objects, and object-key's array values, which may
+ *   contain column names
+ * This does not replace camelizedObjectKeys because of its
+ *   aggressive (recursive) nature, which is a little more greedy
+ *
+ * @param arg <Array|object|string>
+ * @returns An object with camelized keys or camelized string
+ */
+export function camelizeKeys(arg: any) {
```
ephys, line 229: Consider naming this `camelizeKeysDeep`

Avoid using any. Consider using a stricter type, similarly to what camelizeObjectKeys does

src/dialects/abstract/query-interface.js
mike-usa: The flattening needs to happen regardless as to whether the value is wrapped in an array.  It standardizes the return value and makes it iterable, removing the need for type detection. I only wrapped the value out of paranoia to handle `undefined`s. 

See other comments on `toArray`.  It seems to be equivalent to `Array.from()`, but handles `undefined` and `null`. Unless those are a popular use case, which can clutter the code, I think the custom `toArray` should be removed in favor of the native `Array.from()`.

Perhaps a new function should be created to:
```javascript
function toFlatArray( val ){
  Array.from([ val ]).filter(Boolean).flat()
}
```

src/utils/object.ts
```diff
@@ -216,6 +216,56 @@ export function camelizeObjectKeys(obj: { [key: string]: any }) {
...
+  let response: any;
+
+  if (typeof arg === 'string') {
+    return camelize(arg);
```
ephys, line 233: this method should only camelize object keys. I don't see a reason for camelizing strings. 

As such, this method should only expect plain objects or arrays, anything else should throw an error

src/utils/object.ts
```diff
@@ -216,6 +216,56 @@ export function camelizeObjectKeys(obj: { [key: string]: any }) {
...
+ *   arrays of objects, and object-key's array values, which may
+ *   contain column names
+ * This does not replace camelizedObjectKeys because of its
+ *   aggressive (recursive) nature, which is a little more greedy
+ *
+ * @param arg <Array|object|string>
+ * @returns An object with camelized keys or camelized string
+ */
+export function camelizeKeys(arg: any) {
+  let response: any;
```
ephys, line 230: avoid using `any`. Consider using a stricter type.
Avoid using `let` too. It's better to declare a variable in each if and have each if return that variable at the end of its block

src/utils/object.ts
```diff
@@ -216,6 +216,56 @@ export function camelizeObjectKeys(obj: { [key: string]: any }) {
...
+
+  if (typeof arg === 'string') {
+    return camelize(arg);
+  } else if (Array.isArray(arg)) {
+    response = [];
+
+    for (const [i, element] of arg.entries()) {
+      if (typeof element === 'string') {
+        response[i] = camelize(element);
```
ephys, line 239: if the element is not an array and not a plain object (isPlainObject), don't camelize it

src/utils/object.ts
```diff
@@ -216,6 +216,56 @@ export function camelizeObjectKeys(obj: { [key: string]: any }) {
...
+ */
+export function camelizeKeys(arg: any) {
+  let response: any;
+
+  if (typeof arg === 'string') {
+    return camelize(arg);
+  } else if (Array.isArray(arg)) {
+    response = [];
+
+    for (const [i, element] of arg.entries()) {
```
ephys, line 237: this can be simplified by using `return arg.map()` instead of a for-of

src/utils/object.ts
```diff
@@ -216,6 +216,56 @@ export function camelizeObjectKeys(obj: { [key: string]: any }) {
...
+      if (typeof element === 'string') {
+        response[i] = camelize(element);
+      } else { // array/object/etc
+        response[i] = camelizeKeys(element);
+      }
+    }
+  } else if (typeof arg === 'object') {
```
ephys, line 244: use `isPlainObject` instead. `null` has a typeof `object` and this method should only camelize plain objects (i.e. `{}`, `Object.create(null)`, but not `new Date()`)

src/utils/object.ts
```diff
@@ -216,6 +216,56 @@ export function camelizeObjectKeys(obj: { [key: string]: any }) {
...
+    for (const key of Object.keys(arg)) {
+      const value = arg[key];
+
+      if (typeof key === 'string') {
+        // TODO: consider camelizizing values that are columns and reference columns:
+        //    let isColumnKey = key.toLowerCase().includes('column')
+        //    response[camelize(key)] = isColumnKey ? camelizeKeys(value) : value
+        //    REM: camelizing values may have considerable impact on user expectations
+        //      this requires a lot of consideration
+        response[camelize(key)] = value;
```
ephys, line 256: `value` should be passed to `camelizeKeys` if it's a plain object or an array

src/utils/object.ts
```diff
@@ -216,6 +216,56 @@ export function camelizeObjectKeys(obj: { [key: string]: any }) {
...
+      } else {
+        // TODO: (cont.) consider camelizing values that are columns and reference columns:
+        //    response[camelizeKeys(key)] = typeof value !== 'string' ? camelizeKeys(value) : value
+        //    (Also see remark above)
+        response[camelizeKeys(key)] = value;
```
ephys, line 261: if `key` is not a string, then it's a symbol. You can't camelize symbols and it should be used as-is

src/utils/object.ts
```diff
@@ -216,6 +216,56 @@ export function camelizeObjectKeys(obj: { [key: string]: any }) {
...
+
+    for (const [i, element] of arg.entries()) {
+      if (typeof element === 'string') {
+        response[i] = camelize(element);
+      } else { // array/object/etc
+        response[i] = camelizeKeys(element);
+      }
+    }
+  } else if (typeof arg === 'object') {
+    response = {};
```
ephys, line 245:
```suggestion
    response = Object.create(null);
```

ephys (changes requested):

src/dialects/abstract/query-interface.js
mike-usa: > Not fully true, it's prefixed with referenced if it's the target element, and not prefixed otherwise (tableCatalog vs referencedTableCatalog, tableSchema vs referencedTableSchema, tableName vs referencedTableName, columnName vs referencedColumnName): it's always the exact same name + 'referenced'

Yes, but the `referenced` prefix is a qualifier and the scheme is not immediately evident because lacking a qualifier could mean they're related or unrelated.  I like your proposal.  This is going to require a lot of updating to revert.  Is it possible that I can just move towards your proposal instead?  It might be less work to attempt to get buy-in on that.

> considered breaking changes
I noted that in one of the git commits, so I don't know what I was referring to with _this_.  However, if this is used through a property, I would again recommend aliasing the property to the corresponding column:
```
const fk = await getFirstFk();  // pseduo-code
fk.columns                      // populated by tableColumns or sourceColumns
```

src/dialects/postgres/query-generator.js
mike-usa: https://github.com/sequelize/sequelize/pull/14578/files#r889674953

Yes. You already recommended that :) D'oh.  I need to get more sleep.

src/dialects/postgres/query-interface.js
ephys: > I initially intended to camelize the values too

Yes they definitely should not be camelized.

Still, why is this needed here? This method should return a flat array of flat objects. There is no need for a deep call here?

src/dialects/abstract/query-interface.js
ephys: > The flattening needs to happen regardless as to whether the value is wrapped in an array.

The input cannot be a nested array so the flattening does not need to happen.

> See other comments on `toArray`. It seems to be equivalent to `Array.from()`

It is not, it wraps the value in an array if it is not already an array:

```js
toArray([]) -> []
toArray(null) -> []
toArray(undefined) -> []
toArray(1) -> [1]
toArray([1,2]) -> [1,2]
```

src/dialects/abstract/query-interface.js
ephys: > Is it possible that I can just move towards your proposal instead?

I would very much prefer to do that in a separate PR as this one is growing too much in scope. Adding too many things to one PR makes it difficult to review and keep track of the breaking changes for our upgrade guide. You can have more than one PR open at a time if you want though

src/dialects/db2/query-generator.js
ephys: It's not more performant though. String concatenation is not faster than populating an array and joining it :)

test/integration/support.js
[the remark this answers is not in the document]
mike-usa: The `TZ` is used internally by node, which means anywhere it uses `Date()`, it's going to reference local timezone if set.

https://stackoverflow.com/a/46107140/10408280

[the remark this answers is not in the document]
ephys: Thanks, I did not know that :)

mike-usa:
> I'm starting to feel that the scope of this PR is not clear anymore. Can you move the support for postgres 12 and the test commit to different PRs? And probably also the showing of the composite foreign keys could go to another PR. If the PR title contains 'and' it's a good indicator that you can split things up 😊
> 
> But thanks for working on it, we appreciate it!

@WikiRik I'll try.  This did grow... unexpectedly.  I kept learning new things that were essential.  For instance, that 9.5 was dropped, but included in the CI workflow and 12 was required, but there was no way to test against it (other than Native) and testing against it revealed breakages, that I was uncertain were related to my previous changes -- some were, some weren't. The life of organic growth. 

I'll try to walk my way back through the diffs to make reviewing more consumable.  This is going to be a lot of work 😟

src/utils/object.ts
mike-usa: Good point.

Some questions:
1. is `null` ever passed in as a value?
2. is it ever a return value or set as the value of a key?  
 
I would think the database response converts nulls to string form (i.e. `'NULL'`).  If that's true, someone would have to explicitly pass in `null` for camelizeKeys to have in issue -- in which case they get what they deserve 😆

This seems to be overkill for what the function is doing.  I suspect you're future planning on how it could be used with values that are not strings or collections (arrays/objects); for instance, dates objects (or other objects) aren't returned by foreign key lookup queries.

That might be solving a future use case, which is beyond the scope of what the goal here was.

src/utils/object.ts
mike-usa: Are you sure that's what you'd want?  That would change `fk_id` to `fkId`.  Isn't the raw form desired?

WikiRik:
If it really proves too much work to move things to a different PR it also suffices to do a proper explanation of all changes in the description of the PR and in the associated PR on the website repo. Since v7 is still in alpha the generated changelog is not that important in my opinion. But that's my personal opinion, other maintainers might disagree

src/utils/object.ts
ephys: I don't understand the purpose of this method then. If it's not recursively camelizing keys, then the other method is sufficient

ephys:
> the generated changelog

The generated changelog is important. The non-alpha release will also use the generated changelog (a complete list of all changes from v6 to v7)

src/utils/object.ts
ephys: > I would think the database response converts nulls to string form (i.e. 'NULL')

I don't see why the database would convert a null value to `'null'`. We would have no way of differentiating these two values

src/dialects/postgres/query-generator.js
ephys: This specific `getForeignKeyReferenceQuery` method, I can't find any reference to it anywhere in the codebase

mike-usa:
Per the recommendation, I split the /dev and ci updates to a new PR: https://github.com/sequelize/sequelize/pull/14595

Should I create a new PR for the rest of these updates? I was thinking about splitting the schema/FK items up, but they're tightly coupled.  Meaning, in order to resolve of the FK issues, it required having schema specified, which mean writing tests to ensure they worked as intended.

I'm going to try and split out the Utils/support functions, though they're so small I don't know if it'll buy much.  The real question is what to do with this branch?  Should I create a new PR or keep this one alive :)

ephys:
> I was thinking about splitting the schema/FK items up, but they're tightly coupled.

If they're tightly coupled, it's not really worth it. I'll let you know if I see something that can be moved to another PR

> I'm going to try and split out the Utils/support functions

For now I think you can keep them here, we'll see as the discussion evolves :)

> Should I create a new PR or keep this one alive :)

I would keep this one alive

dev/postgres/9.5/check.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
```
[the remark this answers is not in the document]
ephys: > but I saw someone complaining about 9.5 in an issue somewhere

It's supported in Sequelize 6 (because dropping support is a breaking change), but not in v7 (which is what this branch is based on)

src/dialects/postgres/query-generator.js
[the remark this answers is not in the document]
ephys: > Does `.escape` satisfy all platforms?

Yep! It's dialect aware

test/integration/dialects/postgres/schema.test.js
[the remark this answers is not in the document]
ephys: I mean this:

```
await sequelize.query(`CREATE USER ${restricted.username} WITH PASSWORD :password;`, {
  replacements: {
    password: restricted.password,
  },
});
```

I know it's static, and a test, but let's use good practice by default. This way it continues working even if we change the password to something that includes a ' character :)

test/integration/dialects/postgres/schema.test.js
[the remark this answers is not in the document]
ephys: You can pass a string to check a message, I agree it would be nice to be more explicit. If you put a dummy string it should print what error message is used in your console when you run the test. You can then copy it here :)

test/integration/query-interface.test.js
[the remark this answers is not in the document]
ephys: > `tableNames?.[0]` Is testing that there's an element at index 0 of the array, not if tableNames is undefined or null. 

No I guarantee it's testing whether `tableNames` is nullish :)

image

`tableNames?.[0]` is equivalent to `tableNames == null ? tableNames : tableNames[0]`

test/integration/query-interface.test.js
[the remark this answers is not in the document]
ephys: Yes you're right. I commented because it slowed the test unnecessarily but we should move towards removing `dropAllDatabase` asap so this can stay in prevision of that

src/dialects/abstract/query-interface.js
```diff
@@ -618,17 +619,17 @@ export class QueryInterface {
...
     for (let [i, tableName] of tableNames.entries()) {
       if (_.isObject(tableName)) {
         tableName = `${tableName.schema}.${tableName.tableName}`;
       }
 
+      // sqlite has no constraintName so use `${id}_${seq}` as stand-in
       result[tableName] = Array.isArray(results[i])
-        ? results[i].map(r => r.constraint_name)
-        : [results[i] && results[i].constraint_name];
+        ? results[i].map(r => r?.constraint_name || r?.constraintName || [r?.id, r?.seq].join('_'))
```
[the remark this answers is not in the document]
ephys: Nope, [after](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#syntax) :)

src/dialects/postgres/query-generator.js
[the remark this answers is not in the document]
ephys: You shouldn't need to call flat: `toArray` does not wrap arrays in arrays.

The method in question here should definitely not accept a nested array input, so no flattening should be needed.

src/utils/object.ts
mike-usa: I think this would benefit from some test cases to demonstrate what's expected

I also might walk through the developmental process:
1. This originated after working on Foreign Keys and more specifically, composite foreign keys, which required moving from string values to array values
2. I was going to camelize everything -- keys and values -- which explains the recursive nature.  Consider the following example of data I think I came across:
   ```javascript
   [ { books_table: { column_names: ['fk_col1', 'fk_col2'] }, authors_table: { ... } ]
   ``` 

   I initially attempted to use `camelizeObjectKeys`, but here is what the results would be for above object:

   function | outcome
   --- | ---
   foreignKeys.map(camelizeObjectKeys) | `[ { booksTable: { column_names: ['fk_col1', 'fk_col2'] }, { authorsTable: { ... } }]` <br> The existing function would only update the first set of keys
   camelizeKeys(foreignKeys) <br>- or -<br> foreignKeys.map(foreignKeys) | `[ { booksTable: { columnNames: ['fkCol1', 'fkCol2'] }, { authorsTable: { ... } }]` <br> Iterates over everything.  Keys and values.
3. Then I thought it was a bad idea to convert the values, given that a user (or internal methods) may want to use the raw value. At this point I considered scrapping the entire function.
4. Then I came across dialects or differences between the format of getForeignKeysForTable vs getForeignKeysForTables vs getForeignKeyReferences, ..., and the results varied.  I think between a simple object `{constraintName: 'foo'}`, keyed objects `{ foreign_key: { tableName: 'users', schema: 'public'} }` , object that had table name as key and string values in an array (`constraints = { read_only_users: ['fk_1', 'fk_2', 'ref_3'] }`.  Sometimes the array of objects had a key (`[{ foo: { ... }, {bar: {...}}]`) or nested object-value, sometimes they didn't (`[ {...}, {...}]`).
   
   Instead of doing type/format detection, what I was really after was to camelizethe keys of all objects, which could be 2-3 levels deep and may or may not be in an array.  That's the landed code with results below:
   command | explanation
    --- | ---
   camelizeKeys(foreignKeys) <br>- or -<br> foreignKeys.map(foreignKeys) | iterates over everything but object-values <br> `[ { booksTable: { columnNames: ['fk_col1', 'fkcol2'] }, { authorsTable: { ... } }]`
   
   Versatile enough to handle arrays, arrays-of-objects, objects with object-values (e.g., `{ references: { table: {table_name: 'table', schema: 'public'} }`), but I consider this will be more useful in the time to come as I think the eventual state would be to have more layered data, which the API providing methods (or options) that might simplify the results if desired.  
   
---

**More**: I don't have a good understanding of the Sequelize ORM as a user to know how it behaves. In some ORMs you can use values to perform linkages, or the ORM might do a reverse translation.  For instance, they might do something like `orm.findTables({ where: fkId: 'parts_id' })` to retrieve tables where the foreign key is parts_id.  Notice that on the table the foreign key might be named `fk_id`, but the ORM would accept the camelized user input and convert it back to `fk_id`.  I don't know if Sequelize accomplishes those reverse-translation for column names.

---

Note: My last comment above was in response to your comment, which is suggesting a change to the code above (I confused myself when I looked at the code and saw my comment)

ephys: I see a first flaw then, if `books_table` is the name of the table, then it should not be camelized. It should be preserved as-is.

However, the `column_names` key (assuming it's a constant and not an actual table name) should be camelized.

> Notice that on the table the foreign key might be named `fk_id`, but the ORM would accept the camelized user input and convert it back to `fk_id`

We do support that, but that's a separate concern which does not impact this. This needs to return table names & column names as-is from the database. The Model will handle any necessary mapping separately.

src/dialects/postgres/query-interface.js
ephys: I just ran this code and it does return a flat array of flat objects. There is no need to make this recursive

image

ephys:
I compared the output of `getForeignKeyReferencesForTable`

Main:

image

This branch:

image

Ignore the casing differences, this is logged before `camelizeKeys`

A few questions:

- What is `domain`?
- is `constraintType` necessary if it always returns `FOREIGN KEY`?
- Thanks for adding `onUpdate` and `onDelete`
- What is `match`?
- Do we need `referencedTableConstraintName` for anything? I guess it can't hurt. Is it normal that it is quoted?

src/utils/object.ts
mike-usa: This is helpful! Tell me if I've got this wrong, but what I'm hearing is:
- the object names (e.g., `book_authors` table) should never be camelized

If that's the case, then yes, I absolutely need to review this and see where it'd be used. Oddly, I only see it in the Postgres QI and I thought I used it in a few places.  I also need to point out this should be a private function.  The intention was to use it primarily around the foreignKey/constraint lookup-queries to simplify the calling code.  

Are the rules/conditions/specs/conventions like this laid out anywhere for v7? That might save us some time in the future.

mike-usa:
@ephys 
>- What is domain?
_domain_ is refers to the domain the constraint is on (see: https://www.postgresql.org/docs/14/sql-createdomain.html). I would call this experimental information as other databases have it, I just don't know enough about retrieving them from other database platforms.

>- is constraintType necessary if it always returns FOREIGN KEY?
It's not necessary, but to the end-user it quickly gives scope and confidence that the constraint is a foreign key.  It's good metadata even if persistent.

>- Thanks for adding onUpdate and onDelete
I wasn't sure whether these could be camelized.  Later on, I think I discovered they can.

>- What is match?
It the foreign key match type.  https://dba.stackexchange.com/a/71508

>- Do we need referencedTableConstraintName for anything? I guess it can't hurt. Is it normal that it is quoted?
When I saw the different databases returning different lengths from the get foreign key data, I figured including some additional information was okay. I took some liberties, but I promise there was a lot more that I didn't include.  

What I was thinking was that each ORM database should have a standard set of keys, and the different dialects could include a collection of additional information that aren't part of the core sequelize-standard set.  Eventually as the databases become uniform those non-standard keys would pop-out into the standard set.

PR #14595, opened 2022-06-05
ci: remove PostgreSQL 9.5; add PostgreSQL 12

mike-usa:
### Pull Request Checklist

### Description Of Change
This commit also includes dev environment updates.
Details:
- Package.json: Add bulk scripts to start-all/stop-all databases and run-all tests:
   - `yarn start-all` spins up all database containers
   - `yarn stop-all` spins down all database containers
   - `yarn test-all` will cycle through all the different tests one-after-another
      - there is no special output redirection (logging), so need a long shell history and to watch for errors
- Dev: Remove PostgreSQL v9.5 Docker container that was added in a previous commit
- Dev: Add PostgreSQL v12 lifecycle scripts
- Dev: Update DB2 lifecycle scripts; use local directory for "Docker" instead of root
- Dev-docs: add DB2 README
- Dev: Fix check script (temporary workaround for TypeSript)

### Todos

- [ ] Add PostgreSQL 14
- [ ] Talk about compiling /test typescripts so `ts-node` can be used in /dev scripts


dev/postgres/12/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+services:
+  postgres-12:
+    container_name: sequelize-pg-12
+    image: sushantdhiman/postgres:12
+    environment:
+      POSTGRES_USER: sequelize_test
+      POSTGRES_PASSWORD: sequelize_test
+      POSTGRES_DB: sequelize_test
+    ports:
+      - 23012:5432
```
ephys, line 10: let's keep the same port as before, it's what [the config](https://github.com/sequelize/sequelize/blob/main/test/config/config.ts#L67) expects

```suggestion
      - 23010:5432
```

package.json
```diff
@@ -225,22 +225,30 @@
     "cover-integration": "cross-env COVERAGE=true nyc --reporter=lcovonly yarn mocha \"test/integration/**/*.test.[tj]s\" && node -e \"require('fs').renameSync('coverage/lcov.info', 'coverage/integration.info')\"",
     "cover-unit": "cross-env COVERAGE=true nyc --reporter=lcovonly yarn mocha \"test/unit/**/*.test.[tj]s\" && node -e \"require('fs').renameSync('coverage/lcov.info', 'coverage/unit.info')\"",
     "merge-coverage": "lcov-result-merger \"coverage/*.info\" \"coverage/lcov.info\"",
-    "----------------------------------------- local test dbs ------------------------------------------": "",
+    "----------------------------------------- bulk tests ------------------------------------------": "",
+    "start-all": "grep 'bash dev' package.json | grep start | grep -vP \"(all)\" | cut -d'\"' -f4 | sort | xargs -n1 | grep dev | xargs -n1 bash",
```
ephys, line 229: A `start-all` script doesn't really make sense as ports are only unique per database type, not per database version. You can't start postgres & postgres-10 at the same time.

A `start-all-minimum` (starts all minimum supported versions) and `start-all-maximum` (starts all maximum supported versions) would make sense, no port collision there, but right now we have commands such as `start-mysql` and `start-mysql-8` instead of `start-mysql-minimum` and `start-mysql-maximum`

package.json
```diff
@@ -225,22 +225,30 @@
...
+    "test-all": "grep 'test-' package.json | grep -vP \"(unit|integration|smoke|typings)\" | cut -d'\"' -f4 | xargs -n4 yarn ",
```
ephys, line 230: ibmi smoke tests should not be part of this, they take ages to run and require private credentials.

If typings are included, then `lint` could be included too

package.json
```diff
@@ -225,22 +225,30 @@
...
+    "stop-all": "grep 'bash dev' package.json | grep stop | grep -vP \"(all)\" | cut -d'\"' -f4 | sort | xargs -n1 | grep dev | xargs -n1 bash",
```
ephys, line 231: can you move this to a `stop-all.sh` script in `/dev` and make this call it?

package.json
ephys, line 230: I'm also not a fan of using grep to find the different commands. I feel it could eventually clash with something similarly named in the file.

Even though it's verbose, I prefer the `test-unit-all` approach.
You could add a `test-integration-all` that is written similarly to `test-unit-all`, then this command can be `yarn test-unit-all && yarn test-integration-all && yarn test-typings && yarn lint-no-fix`

.github/workflows/ci.yml
```diff
@@ -191,7 +191,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [14, 18]
-        postgres-version: [9.5, 10]
+        postgres-version: [10, 12]
```
ephys, line 194: This is fine until we fork the docker image to add pg 14 :)

ephys (changes requested):

dev/db-connection-check.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+/* // Commented out until we compile test/*.ts for /dev to use
+const sequelize = require('../../../test/support').createSequelizeInstance();
```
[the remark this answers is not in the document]
mike-usa: I'll need assistance with that. My understanding was that as a subset, JavaScript should just work inside `ts-node`, but when I called it against `check.js`, it complained. It'll be some time before I can get up to speed with TS -- I've got ServiceNow and have to re-buff on Python for work.

dev/postgres/12/start.sh
```diff
@@ -0,0 +1,17 @@
...
+cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" # https://stackoverflow.com/a/17744637
+
+
+docker-compose -p sequelize-pg-12 down --remove-orphans
+docker-compose -p sequelize-pg-12 up -d
+
+./../../wait-until-healthy.sh sequelize-pg-12
+
+# docker exec sequelize-pg-12 \
+#   bash -c "export PGPASSWORD=sequelize_test && psql -h localhost -p 5432 -U sequelize_test sequelize_test -c '\l'"
```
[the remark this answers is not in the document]
mike-usa: Are you sure? :) 

BTW after thinking about that CONTRIBUTING CHEATSHEET doc, I think I'm moving away from that, and going to move towards having a README in each platforms top level directory. I think we still need some Cliff's Notes higher in the project repo, but I'm starting to see how these can be organized and where they should live.

dev/mysql/8.0/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       MYSQL_PASSWORD: sequelize_test
       MYSQL_ROOT_PASSWORD: sequelize_test
     ports:
-      - 20057:3306
+      - 20080:3306
```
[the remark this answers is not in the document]
mike-usa: Good catch! That was a last-second change.  I'd really like to have each database run at the same time and then to cycle through them for my local testing.

The Config file might grow slightly, but in this test environment, this would save me time, especially as new versions are introduced.  I just don't know the best way to implement it.

I was thinking:
DIALECT=postgres
DIALECT=postgres12
DIALECT=postgres12.4

Where Config might functionize:
```javascript
function Config(dialect,options){
  const [_, platform, seekVersion] = dialect.match(/([a-z]+)(.*)/)
  const settings = {
    mssql: {
      default: {...}
    },
    postgres: {
      default: {...},
      '12.x': {...}
    },
    ...
  };
  
  const defaults = settings[dialect].default;
  const versions = Object.keys(settings[dialect]).filter(v=>v!=='default');
  if (seekVersion && versions.length) {
    // ordered by greatest semver first
    const semvers = semver.rsort(versions.map(semver.coerce));
    if (!semvers.length) { return defaults; }
    
    // find largest semver that matches
    const foundSemver = semvers.find(v=>semver.gte(semver.coerce(seekVersion), v)
    if (!foundSemver) { return defaults; }
    
    const settingVersion = versions.find(v=>semver.eq(semver.coerce(v),foundSemver));
    if (!settingVersion) { return defaults; }
    
    return Object.assign( defaults, settings[dialect][settingVersion] );
  }

  return defaults;
}
```

With settings looking almost the same as it currently does with a new tier for:

```javascript
  mysql: {
    default: { ... },
    '5.7': { port: ... },
    '8': { port: ... }
  },
  postgres: {
    default: {
      database: env.SEQ_PG_DB || env.SEQ_DB || 'sequelize_test',
      username: env.SEQ_PG_USER || env.SEQ_USER || 'sequelize_test',
      password: env.SEQ_PG_PW || env.SEQ_PW || 'sequelize_test',
      host: env.POSTGRES_PORT_5432_TCP_ADDR || env.SEQ_PG_HOST || env.SEQ_HOST || '127.0.0.1',
      port: env.POSTGRES_PORT_5432_TCP_PORT || env.SEQ_PG_PORT || env.SEQ_PORT || 23_010,
      pool: {
        max: Number(env.SEQ_PG_POOL_MAX || env.SEQ_POOL_MAX || 5),
        idle: Number(env.SEQ_PG_POOL_IDLE || env.SEQ_POOL_IDLE || 3000),
      },
      minifyAliases: Boolean(env.SEQ_PG_MINIFY_ALIASES),
    },
    '12.3': {
      port: env.POSTGRES_PORT_5432_TCP_PORT || env.SEQ_PG_PORT || env.SEQ_PORT || 23_012,
    }
  },
  ```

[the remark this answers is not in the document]
mike-usa: This is just an idea. I think there is type-detection on the dialects and it would mean modifying more than just support.ts/config.ts. I'd like to see it implemented, but I'm still thinking about the level of effort -- it was nice to see all the databases tested in sequence (when I thought it was all the databases).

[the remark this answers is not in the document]
mike-usa: Well, I have a proof of concept working.  Maybe I should just submit it and let others decide if they want it to be rejected.  It wasn't as difficult as I thought probably because this mostly affects dev environment.

package.json
mike-usa: Yeah I have my concerns. I weighed the frequency the dev containers would be upgraded against the potential collision.  If moving to a separate script, It'd be better to iterate over supported dialects.

package.json
WikiRik: Yes, if we're keeping these it might be better to move them all to `/dev`

ephys:
I'll close this issue as we've now included this through separate PRs :)

PR #14587, opened 2022-06-01
build(mssql): add support for MSSQL 2022

WikiRik:
### Pull Request Checklist

_Please make sure to review and check all of these items:_

- [X] Have you added new tests to prevent regressions?
- [X] Does `yarn test` or `yarn test-DIALECT` pass with this change (including linting)?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [X] Did you update the typescript typings accordingly (if applicable)?
- [X] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [X] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)?

### Description Of Change

Support for MSSQL 2022 has been added. MSSQL 2019 has not been removed yet since MSSQL 2022 is still in public preview; https://cloudblogs.microsoft.com/sqlserver/2022/05/24/announcing-sql-server-2022-public-preview-azure-enabled-with-continued-performance-and-security-innovation/

### Todos

- [x] Check website repo for updates to docs ( https://github.com/sequelize/website/pull/175 )
- [x] Remove CI check for MSSQL 2019 (in future PR)
- [x] Update required CI checks (in future PR)

PR #14604, opened 2022-06-08
refactor: config.js (and update scripts) + new start-up/check scripts

mike-usa:
### Pull Request Checklist

### Description Of Change

Includes:

  - yarn start-all/stop-all: start and stop all databases with start.sh scripts in their 
    dev/dialects/<version> sub-directories
    git 
  - yarn test-all runs `prepare` and `test-typings` once, then runs the other tests for
    each dialect and version

Details:

  1. Running tests against each database version (while running in paralled) requires a 
     couple things:
       (1) the databases must run on different ports 
       (2) there needs to be a way to pass opt-in versions to the test
     
     The minimally-intrusive (least amount of additional typing) way was to coerce the 
     dialect value to now include version numbers. The version numbers are now concatenated
     to the end of the dialect name. There are no delimiters except inside the version
     number itself, which follows semver.  Example: `DIALECT=mysql5.7 yarn test`

     The minor version isn't necessary as it as the Config function will perform a lookup
     to find best matches. $DIALECT is parsed by Support.js and Config.js with new functions
     to assist. Some dialects, like db2, end numeric.  That special case is manually checked.

  2. There are now fallbacks for *-native (e.g., postgres-native) to run on their default 
     ports, or can be updated via the environment variable (e.g., SEQ_PORT). Though not tested
     this may also open a way to test `mysql-native` and other variations, not previously
     included in package.json scripts.

  3. There is smart matching if a major version is supplied. Example: DIALECT=mysql5 will look through
     the list of database configurations and choose the highest that matches or equals it. If both 5.7
     and 5.2 configurations exist, it will choose 5.7.

  4. When database version is not supplied, for instance DIALECT=postgres, the config will attempt to
     retrieve the port from the most recent Docker container of that dialect. This means there is
     now a dependency on container names and order databases may be loaded. Because a developer
     may not have Docker isntalled, this code is placed in a try/catch block. Though, having
     Docker makes this much more accomodating to spin up a container and let Config do the work,
     especially as new versions are added.

  5. Config.js and the scripts in dev were written in TypeScript

  6. `dev/db-connection-check.ts`` replaces the previous check.js, which was broken for MySQL and
     other databases. This now works for all databases and uses the new Config modifications.

  7. A decent start to a README is added for db2 to help explain how to connect and run queries, given
     their online documentation all looks the same but may not apply.

  8. Use `DEBUG=1`` or `DEBUG=true` (really any truthy value) to show some additional 
     debug output for connection details.
    
### Todos

- [ ] test: native databases (postgres-native)


test/config/config.ts
```diff
@@ -1,98 +1,346 @@
...
-    host: env.SEQ_MSSQL_HOST || env.SEQ_HOST || 'localhost',
-    username: env.SEQ_MSSQL_USER || env.SEQ_USER || 'SA',
-    password: env.SEQ_MSSQL_PW || env.SEQ_PW || 'Password12!',
-    port: env.SEQ_MSSQL_PORT || env.SEQ_PORT || 22_019,
-    database: env.SEQ_MSSQL_DB || env.SEQ_DB || 'sequelize_test',
-    dialectOptions: {
-      options: {
-        encrypt: false,
-        requestTimeout: 25_000,
+    // When SQL Server has one instance it runs under 1433, with multiple instances
```
ephys, line 22: This looks much more complicated than it needs to be
If you're going to add support for having both oldest & latest run at the same time, I recommend you have an object that looks like this instead:

```ts
const configs: {
  base: {
    host: env.SEQ_MSSQL_HOST || env.SEQ_HOST || 'localhost',
    username: env.SEQ_MSSQL_USER || env.SEQ_USER || 'SA',
    password: env.SEQ_MSSQL_PW || env.SEQ_PW || 'Password12!',
    // no port!
    // ...etc
  },
  oldest: {
    port: env.SEQ_MSSQL_PORT || env.SEQ_PORT || 22_020,
  },
  latest: {
    port: env.SEQ_MSSQL_PORT || env.SEQ_PORT || 22_019,
  },
},
```

It's just an example, the values are probably not correct. Key points:

- `native` is not desired, the docker images are the way to go as they reflect what will be used on the CI (but if really needed can be set manually by changing the `SEQ_PORT` env variable).
- Use `oldest` and `latest` instead of `2019`, `2017` to reduce how much boilerplate must be updated when we upgrade

test/config/config.ts
```diff
@@ -1,98 +1,346 @@
...
+
+//   ██████╗ ██████╗ ███╗   ██╗███████╗██╗ ██████╗     ███████╗██╗   ██╗███╗   ██╗ ██████╗
+//  ██╔════╝██╔═══██╗████╗  ██║██╔════╝██║██╔════╝     ██╔════╝██║   ██║████╗  ██║██╔════╝
+//  ██║     ██║   ██║██╔██╗ ██║█████╗  ██║██║  ███╗    █████╗  ██║   ██║██╔██╗ ██║██║
+//  ██║     ██║   ██║██║╚██╗██║██╔══╝  ██║██║   ██║    ██╔══╝  ██║   ██║██║╚██╗██║██║
+//  ╚██████╗╚██████╔╝██║ ╚████║██║     ██║╚██████╔╝    ██║     ╚██████╔╝██║ ╚████║╚██████╗
+//   ╚═════╝ ╚═════╝ ╚═╝  ╚═══╝╚═╝     ╚═╝ ╚═════╝     ╚═╝      ╚═════╝ ╚═╝  ╚═══╝ ╚═════╝
+//
+
+export function Config(dialect: string | undefined): Options {
```
ephys, line 201: if `config` is a function it should not be Capitalized, it should also have a more explicit name such as `getConfig`, `buildConfig`

test/config/config.ts
```diff
@@ -1,98 +1,346 @@
...
+  const defaults: { [key: string]: any } = configs;
+  const config = defaults[dialectName].default as Options;
+  const versions = Object.keys(defaults[dialectName]).filter(v => !/(?:default|native)/.test(v));
+
+  //  ┌┬┐┌─┐┌─┐┬┌─┌─┐┬─┐  ┌─┐┌─┐┬─┐┌┬┐  ┬  ┌─┐┌─┐┬┌─┬ ┬┌─┐
+  //   │││ ││  ├┴┐├┤ ├┬┘  ├─┘│ │├┬┘ │   │  │ ││ │├┴┐│ │├─┘
+  //  ─┴┘└─┘└─┘┴ ┴└─┘┴└─  ┴  └─┘┴└─ ┴   ┴─┘└─┘└─┘┴ ┴└─┘┴
+
+  if (!dialectVersion) {
+    // attempt to get database port from newest running container for that dialect
```
ephys, line 254: Just set the port in the config instead, this is not necessary

test/config/config.ts
```diff
@@ -1,98 +1,346 @@
...
+    const versionOptions = semver.rsort(versions.map(v => semver.coerce(v) ?? ''));
+    if (!versionOptions.length) {
+      return postProcess(preReturnValues);
+    }
+
+    // find the largest semver that matches.  `.satisfies()` compares pattern with
+    //   the 'x' in the range pattern representing largest possible value:
+    //     `.satisfies('5.7.0','<=5.x.x')` is true; 5.x (infinity) is greater than 5.7
+    //     `.satisfies('8.2.3','<=5.x.x')` is false; 8 is greater than 5
+    const foundSemver = versionOptions.find((versionOption: any) => semver.satisfies(versionOption, versionMatch));
```
ephys, line 300: by using `oldest` and `latest`, you remove the need for all of this too

test/config/config.ts
```diff
@@ -1,98 +1,346 @@
...
+ * Perform any final maintenance before returning to caller
+ * @param args Post-processing arguments object
+ * @param args.config Database connection values
+ * @param args.defaults Default connection values for the database
+ * @param args.dialectName Name of the database dialect (type)
+ * @returns Dialect config details
+ */
+function postProcess({ config, defaults, dialectName }: {
+  config: Options,
+  defaults: { [key: string]: any },
```
ephys, line 332: this typing is too broad

test/integration/pool.test.js
```diff
@@ -73,7 +73,7 @@ function attachMSSQLUniqueId(connection) {
 }
 
 describe(Support.getTestDialectTeaser('Pooling'), () => {
-  if (dialect === 'sqlite' || process.env.DIALECT === 'postgres-native') {
+  if (dialect === 'sqlite' || Support.getDialectName(process.env.DIALECT) === 'postgres-native') {
```
ephys, line 76:
```suggestion
  if (dialect === 'sqlite' || dialect === 'postgres-native') {
```

test/support.ts
```diff
@@ -153,14 +153,15 @@ export async function prepareTransactionTest(sequelize: Sequelize) {
 }
 
 export function createSequelizeInstance(options: Options = {}) {
-  options.dialect = getTestDialect();
+  options.dialect = getTestDialect(); // only returns translated dialectName "mysql8"->"mysql"
```
ephys, line 156: This comment belongs in the JSDoc of getTestDialect instead

```suggestion
  options.dialect = getTestDialect();
```

test/support.ts
```diff
@@ -153,14 +153,15 @@ export async function prepareTransactionTest(sequelize: Sequelize) {
...
 
-  const config = Config[options.dialect];
+  // pass process.env.DIALECT b/c it may contain version "postgres12" config can use
+  const config = Config(process.env.DIALECT);
 
   const sequelizeOptions = defaults(options, {
     host: options.host || config.host,
     logging: process.env.SEQ_LOG ? console.debug : false,
-    dialect: options.dialect,
+    dialect: getDialectName(options.dialect),
```
ephys, line 164: why calling `getDialectName` here? `getTestDialect` is supposed to return a valid `Dialect` already

test/support.ts
```diff
@@ -245,32 +246,66 @@ export function getAbstractQueryGenerator(sequelize: Sequelize): unknown {
   return new ModdedQueryGenerator({ sequelize, _dialect: sequelize.dialect });
 }
 
+export function getDialectName(dialect: string): Dialect | string {
```
ephys, line 249: Almost all of this code is duplicated with `Config()`, it should be unified so config calls this method. Maybe rename this to `parseDialectEnv` and make it return `{ dialect: Dialect, version: 'latest' | 'oldest' }` so `Config()` can use it

ephys (changes requested):
Thanks for the PR :)

I agree that having the ability to have two versions of a DB run at the same time would be good

However if we're going to do this, we need to call these versions `oldest` and `latest` like we're doing in https://github.com/sequelize/sequelize/pull/14587#issuecomment-1148516022. I.e. `postgres-latest`, `postgres-oldest`. Calling them `postgres-10` and `postgres-14` means we'll have to touch many things once we drop pg 10 or add support for pg 15. That's the bulk of my comments

The rest is about how to reduce the size of your diff to avoid collision with other PRs, and refactors to reduce duplicate code

Other than that, I think this would be a good change!

package.json
```diff
@@ -65,6 +65,7 @@
     "@types/chai-datetime": "0.0.37",
     "@types/lodash": "4.14.182",
     "@types/mocha": "9.1.1",
+    "@types/semver": "^7.3.9",
```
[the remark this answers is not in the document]
WikiRik: And don't forget to commit the updated `yarn.lock` file

package.json
```diff
@@ -225,19 +226,25 @@
...
-    "stop-mysql": "bash dev/mysql/5.7/stop.sh",
-    "stop-mysql-8": "bash dev/mysql/8.0/stop.sh",
-    "stop-postgres": "bash dev/postgres/10/stop.sh",
-    "stop-mssql": "bash dev/mssql/2019/stop.sh",
-    "stop-db2": "bash dev/db2/11.5/stop.sh",
+    "----------------------------------------- bulk tests ------------------------------------------": "",
+    "start-all": "ts-node dev/start-all.ts",
+    "test-all": "ts-node dev/test-all.ts",
+    "stop-all": "ts-node dev/stop-all.ts",
+    "----------------------------------------- database containers -------------------------------------": "",
```
WikiRik, line 233: See what I did in #14587 and see if we can add that for other dialects as well

dev/dialects/db2/11.5/stop.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -Eeuxo pipefail # https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/
+cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" # https://stackoverflow.com/a/17744637
+
+sudo docker stop sequelize-db2
+docker rm sequelize-db2
+# docker-compose -p db2server down --remove-orphans
```
WikiRik, line 7: Why is this commented out? If it's not needed we can remove it

dev/dialects/db2/README.md
```diff
@@ -0,0 +1,84 @@
+# DB2 README
+
+There are many versions of DB2 products, which have different features and different documentation. For instance, in this version `JSON_ARRAYAGG` and `ARRAY_AGG` functions are not available, only `LISTAGG` is.  To confirm this query can be run:
```
WikiRik, line 3: What is 'this version'?

dev/dialects/db2/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+   ```bash
+   db2 "select distinct routinename from syscat.routines where routinename like 'JSON%' or routinename like '%ARRAY%' or routinename like '%AGG%' order by 1"
+   ```
+
+Types of DB2 Platforms
```
WikiRik, line 9: Is the DB2 dialect supporting all 4?

dev/start-all.ts
```diff
@@ -0,0 +1,31 @@
...
+    execSync(`${scriptFullFileName}`, {stdio: 'inherit'});
+  }
+  catch(e){
+    console.error(e);
+  }
+  console.groupEnd();
+});
+
+
+// NOTE: there is opportunity here for a future revision to replace the shell script with JavaSCript
```
WikiRik, line 31:
```suggestion
// NOTE: there is opportunity here for a future revision to replace the shell script with JavaScript
```

dev/stop-all.ts
```diff
@@ -0,0 +1,31 @@
...
+    execSync(`${scriptFullFileName}`, {stdio: 'inherit'});
+  }
+  catch(e){
+    console.error(e);
+  }
+  console.groupEnd();
+});
+
+
+// NOTE: there is opportunity here for a future revision to replace the shell script with JavaSCript
```
WikiRik, line 31:
```suggestion
// NOTE: there is opportunity here for a future revision to replace the shell script with JavaScript

```

Not sure if the empty line would display correctly, but end with a newline

.eslintrc.js
```diff
@@ -148,6 +148,15 @@ module.exports = {
       'no-prototype-builtins': 'off',
       'no-multi-spaces': 'off',
     },
+  }, {
```
WikiRik, line 151: Can you un-ignore the dev folder here? I think it shouldn't add too much noise to this PR. Otherwise make a separate PR  for that

dev/test-all.ts
```diff
@@ -0,0 +1,67 @@
...
+function runTest(command: string){
+  console.group();
+  console.log(command,"\n");
+  try { execSync(command, {stdio: 'inherit'}); }
+  catch(e) { console.error(e); }
+  console.groupEnd();
+}
+
+
+// NOTE: there is opportunity here for a future revision to replace the shell script with JavaSCript
```
WikiRik, line 67:
```suggestion
// NOTE: there is opportunity here for a future revision to replace the shell script with JavaScript
```

test/integration/pool.test.js
WikiRik:
```suggestion
  if (['sqlite','postgres-native'].includes(dialect)) {
```
This might be even nicer

WikiRik (commented):
Few more nitpicky stuff, but together with the suggestions of Zoé this is a nice PR to improve the DX. Thanks for your work on this and the other PRs!

.eslintrc.js
```diff
@@ -148,6 +148,15 @@ module.exports = {
...
+    files: ['test/**/*.ts'],
+    rules: {
+      // ignore the [_unusedVar, variable]
+      '@typescript-eslint/no-unused-vars': ['error', {
+        destructuredArrayIgnorePattern: '^_',
+        varsIgnorePattern: '[iI]gnore',
+      }],
+    },
```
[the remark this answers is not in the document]
mike-usa: I saw one recommendation for that. Don't you think comments in parameter lists clutter the code? It also goes against typical JS conventions, that use `_` for de-structured assignment and nebulous functional args. It's such a common pattern that `eslint` uses `^_` throughout its examples: https://eslint.org/docs/rules/no-unused-vars#destructuredarrayignorepattern

dev/dialects/db2/11.5/start.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
mike-usa: What's more difficult to merge? Are you suggesting back-porting to other versions?

While it may seem like more noise, it keeps convention -- no other place are there dialect folders (e.g., `/db2`, `/mysql`, `/postgres`) that aren't under a `/dialects` folder:

- /src/dialects/*
- /lib/dialects/*
- /test/integration/dialects/*
- /test/unit/dialects/*

The other places may have good reason to have that pattern, but would `/dev` be the _only_ one to break that convention? The directory should hardly be used, so consistency seems more important.

Currently the scripts are greping `start.sh` to get the dialect name.  What would be better is to iterate over the directory structure and not rely on grep or linux shell commands.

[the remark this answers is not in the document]
WikiRik: I think she is mostly referring to your other open PRs but I'm fine with doing it either way

dev/dialects/db2/README.md
mike-usa: Community.  Looks like 11.5.6

dev/dialects/db2/README.md
mike-usa: No idea. I was surprised there was a community edition.  Per something I read somewhere in Sequelize comments or maybe Contributing, it seems like its leaning on the community for supporting some of the features, for platforms that may not be broadly available to the public.

I would think the community edition is the lowest common denominator and that supporting it would support the other, but that's a guess.  If you want the others, you have to pay.

.eslintrc.js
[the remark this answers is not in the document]
mike-usa: I'll turn off the rule and go back through and undo the code, but submit this should be re-considered.

dev/db-connection-check.ts
```diff
@@ -0,0 +1,45 @@
+// Wanted: '@sequelize/core' in place of '..'
+// Issue: dev/tsconfig.json not configured correctly
+// import { Sequelize } from '@sequelize/core';
+
+import type { Dialect, Options } from '..';
```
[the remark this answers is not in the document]
mike-usa: I attempted that, but that was earlier on when I was learning TypeScript and how to run/configure it.  I thought I tested it again and it still broke, but I'll try one more time.

dev/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "extends": "../tsconfig.json",
```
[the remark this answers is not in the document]
mike-usa: I have no idea how to configure this file! 😆 That would make sense, but I wasn't sure about the directory structure and thought breaking the two would be easier to track when needing to customize.

parent -> dev
vs
parent -> test -> dev

That said I couldn't get `@sequelize/core` to work properly in dev, which maybe is more an issue with eslint?-- I don't know.  But that's why `'..'` is used everywhere.  Could use a TS expert to assist.

package.json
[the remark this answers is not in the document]
mike-usa: good catch -- both of you!

test/support.ts
mike-usa: Good question.  

I don't know if this was oversight or because I thought making it obvious that `options.dialect` is the dialect name.  Looking at `createSequelizeInstance` I'm curious why it takes in `options` that are then overwritten by the environment variable (i.e. `process.env.DIALECT` via _getTestDialect_).

Because `options` is a "reference", it's values are mutable in the caller.  The caller might not want the _dialect_ to change in its options it passed in -- it could iterate and pass in `mysql5` and then want to try (locally `options.dialect+'7'` to test `mysql57`, but with function changing the value it'd be `mysql7`). We could deep copy options, but that might be overkill.

I think maybe this is all that's needed.  Use supplied options, or get dialect from environment:

   ```javascript
   options.dialect = getDialectName(options?.dialect||'') || getTestDialect(),
   ```


Sidenote:
1. I think the _Test_ in _getTestDialect_ is a little confusing or might benefit from a change to  `getDialectNameFromEnvironment`, which would be a little less confusing. I'm punting on this for now as all tests call the function.
2. `getDialectName` is a parser that takes any _dialectal_ string argument and returns a dialect or empty string, which means `getDialectName` is a little clearer and shows intent in the code

test/support.ts
mike-usa: I like the way that sounds.  I'll take a look.  Really like what you all are doing with _oldest_ and _latest_.

dev/dialects/db2/11.5/start.sh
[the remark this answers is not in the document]
ephys: The reasoning is that changes with a large diff impact that don't have a purpose should be avoided to prevent unnecessary collisions with other PRs that may be touching the same files, as well as reducing diff size for review

In the examples you listed, they all mimic `src`, but there is no convention about a `dialects` directory. `lib` is the built version of `src`, and tests try to mimic the structure of src so tests are easier to find.

That being said, changing `start` to making it list the contents of the folder instead of a grep would be better, and is reason enough to introduce this `/dialect` folder

dev/db-connection-check.ts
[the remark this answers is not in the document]
ephys: Looking into it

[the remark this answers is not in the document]
ephys: Here you go :) https://github.com/sequelize/sequelize/pull/14608

.eslintrc.js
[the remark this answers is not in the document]
ephys: We can change the lint rules but that should happen in its own discussion thread / PR instead of here
Personally I would prefer to avoid declaring dummy variables as much as possible.
`/* ignore */` also has precedent as we have to put that in empty `catch` blocks by convention

dev/tsconfig.json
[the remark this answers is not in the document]
ephys: This will be resolved by https://github.com/sequelize/sequelize/pull/14608

package.json
ephys: > but if they're following semver, each major version includes breaking changes. So each major version should be tested.

Which is why we're testing against latest (https://github.com/sequelize/sequelize/pull/14587#issuecomment-1150355327 for more information)

> `*-<specific version>`

I would rather avoid adding `*-<specific version>` to our dev folder. Issues with these versions are going to be rare enough that the added maintenance in keeping our dev folders up to date is not going to be worth it. Having the CI warn every now and then if there is an issue, and manually booting a new container is sufficient

> There's so much convenience in not having to type the extra qualifier suffix 

Personally that's just `yarn sta<tab>` followed by selecting whichever one I want using arrow keys (thanks to https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/yarn):
image

> BTW we also have to keep in mind that `latest` and `oldest` are obfuscated.

True, that's something I considered in #14587, but ultimately could not come up with a better short name. Ideas welcome though!

.eslintrc.js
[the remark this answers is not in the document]
WikiRik: Earlier there was the suggestion to create a new Slack channel for this kind of discussions since #sequelize-help might not be the easiest. Maybe we should follow up on that

.eslintrc.js
ephys: `update-authors.js` is already un-ignored, `sscce-helpers.js` already follows the code style properly, and this PR modifies every other js/ts file present in dev. I agree it's time to un-ignore the dev folder 🥳

.eslintrc.js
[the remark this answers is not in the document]
ephys: `#sequelize-contributions`? (I was going to say `#sequelize-development` but could be confusing?)

package.json
WikiRik: Yeah, I think Zoé mentioned `minimum` and `maximum` but I disliked that so I went with `oldest` and `latest`. What we can do to improve is mention on the releases page which versions are checked by our CI so people know that the support of postgres 14 (in this example) is not tested

mike-usa: RE: @ephys 
---
> Which is why we're testing against latest (https://github.com/sequelize/sequelize/pull/14587#issuecomment-1150355327 for more information)
Yes, but I may not be explaining the problem well enough. The intermediary versions are also major versions.  _Latest_ may work, _Oldest_ may work, but there might be an intermediary major version that doesn't work.  

Consider version 6, 7, 8, which are all breaking changes.  7 (or a minor version) could have removed a feature. 8 might have added it back.  If you never test 7, there's no way of knowing.

> I would rather avoid adding `*-<specific version>` to our dev folder. Issues with these versions are going to be rare enough that the added maintenance in keeping our dev folders up to date is not going to be worth it.

The dev folder has already been out-of-date. I realized this when I saw the version number from `tree`. I don't think burying version numbers in files is going to make that issue any better.  It's your project, a more robust scheme would allow you to test against minor versions.  There's going to be time where 9.5 works and 9.6 doesn't (as with PostgreSQL).  You'll want an environment that accommodates testing/developing for those supported versions and ensure regression tests don't fail.

Keep the felxibility that the object can grown by version numbers -- specific versions in the directory structure help logging, parsing, and creating other scripts.  Create symbolic links to the versions (or implement dynamic lookups) if you don't want to maintain package.json.

I submit that the metadata of having the version numbers as directories is more beneficial than discarding them for cosmetics.

```
dev/dialects/
├── db2
│   ├── 11.5
│   │   ├── Docker
│   │   ├── start.sh
│   │   └── stop.sh
│   ├── latest -> 11.5
│   ├── oldest -> 11.5
│   └── README.md
├── mssql
│   ├── 2019
│   │   ├── docker-compose.yml
│   │   ├── start.sh
│   │   └── stop.sh
│   ├── 2022
│   │   ├── docker-compose.yml
│   │   ├── start.sh
│   │   └── stop.sh
│   ├── latest -> 2019
│   └── oldest -> 2022
```

>True, that's something I considered in https://github.com/sequelize/sequelize/pull/14587, but ultimately could not come up with a better short name. Ideas welcome though!

Unless you want to get more specific (with a qualifier), I think what you all have come up with are great! I'm not sure I'd change it.

As food for thought.  If you wanted to get more specific, it could be `mssql-latest-installed`, but that's somewhat redundant because they're all installed.  The only purpose is to serve as an overt reminder that to make you think there might be a version not installed.  At some point you might want to test against up-and-coming databases.  Something like `mssql-edge`.

@WikiRik
---
> Yeah, I think Zoé mentioned minimum and maximum but I disliked that so I went with oldest and latest. 
Minimum vs Maximum are more technically correct, but I think oldest and latest have better UX (or DX) and that'd be my preference too if I had a vote.

> What we can do to improve is mention on the releases page which versions are checked by our CI so people know that the support of postgres 14 (in this example) is not tested
I like that idea, especially if you can automate it. It's not so much about needing to test or develop for everything, but building the transparency.  If someone reports an issue where something is broken a certain version, though, that's where the flexibility of minor versions comes in.  You can do some of that natively, but why when you have such a good dev-environment process?  Lean on that, don't resist the version scheme 😆

.eslintrc.js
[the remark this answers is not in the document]
mike-usa: I asked Sacha if it would be worth creating #contribute: 

- Name: contribute
- Description: Contributing to Sequelize and Sequelize-CLI source code and documentation.

Padded with maintainers/contributors.  Just need a space to openly discuss ideas of the project with a lot of back-and-forth, and reduce the noise from those asking for help in #squelize-help.

test/config/config.ts
```diff
@@ -1,98 +1,346 @@
...
   },
 };
+
+//   ██████╗ ██████╗ ███╗   ██╗███████╗██╗ ██████╗     ███████╗██╗   ██╗███╗   ██╗ ██████╗
+//  ██╔════╝██╔═══██╗████╗  ██║██╔════╝██║██╔════╝     ██╔════╝██║   ██║████╗  ██║██╔════╝
+//  ██║     ██║   ██║██╔██╗ ██║█████╗  ██║██║  ███╗    █████╗  ██║   ██║██╔██╗ ██║██║
+//  ██║     ██║   ██║██║╚██╗██║██╔══╝  ██║██║   ██║    ██╔══╝  ██║   ██║██║╚██╗██║██║
+//  ╚██████╗╚██████╔╝██║ ╚████║██║     ██║╚██████╔╝    ██║     ╚██████╔╝██║ ╚████║╚██████╗
+//   ╚═════╝ ╚═════╝ ╚═╝  ╚═══╝╚═╝     ╚═╝ ╚═════╝     ╚═╝      ╚═════╝ ╚═╝  ╚═══╝ ╚═════╝
+//
```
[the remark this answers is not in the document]
mike-usa: That's just editor width. The text is 90-columns, you almost have to squish your editor to purposely wrap it like that, but who codes with word-wrapping?

I'm not married to this. Like you, I saw it done elsewhere and I had concerns about its accessibility.  I was concerned more with the seeing-impaired.  I later discovered it was a much more pleasurable experience when re-reading code or  reading other people's code -- like dramatically different.  For longer files, it helped break up the parts when scanning.  The big headers were even readable in editor sidebars, improving mouse-drag navigation. And screen readers can still read the function and their names; its additional, not mandatory.

Again, I'm not married to it, but there's more than meets the eye and want to make sure that's also considered.

test/config/config.ts
```diff
@@ -1,98 +1,346 @@
...
+    dialectName = dialectName.replace(/(.+?)-native/, '$1');
+  }
+
+  // TypeScript isn't intelligent enough to realize this will be assigned
+  dialectName = dialectName!;
+  dialectVersion = dialectVersion!;
+
+  const defaults: { [key: string]: any } = configs;
+  const config = defaults[dialectName].default as Options;
+  const versions = Object.keys(defaults[dialectName]).filter(v => !/(?:default|native)/.test(v));
```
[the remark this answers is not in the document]
mike-usa: >This whole special case thing can be avoided 

Agreed.  There are other schemes, it's just much better DX to type `DIALECT=mysql8` over `DIALECT=mysql-8` or `DIALECT=mysql DIALECT_VERSION=8` -- lots of us don't use _zsh_.

> if you name the env values postgres-oldest and postgres-latest instead of postgres10 and postgres12 

Yes, it can be avoided, but the goal is to be able to be able to accommodate ad-hoc development containers and be able to develop/test against targeted versions, not just the two that are present.

How do you ensure patches are tested?  As specific minor/patch versions have flaws, how do you ensure they're tested with changes? What's the regression testing process like if you support v10 and v18 and there's a v14-16 issue? -- I'd think you'd include them in the development environment to ensure all developers are capable of testing or running tests against them so that patches are maintained, and bugs aren't re-introduced.  Without versions between oldest-latest, you're not going to catch those flaws.

I'm wondering if you have a more preferred solution other than 2-defaults.  I agree maintaining the _package.json_ would be cumbersome, especially for every version.  What do you think of this? 

Forego environment variable entirely in favor of command-line arguments?
```
Format:
yarn test <dialect> <version>

Example:
yarn test mssql latest
yarn test mssql 2022
yarn test postgres native
```

Instead of any version or dialect being stored in the package.json, create a controller and a script in dev:

package.json
```
  "scripts": {
    "dialects": "yarn ts-node dev/dialects.ts",
    "test": "run(){ SEQ_DIALECT=\"${1}\" SEQ_DIALECT_VERSION=\"${2}\"; yarn prepare &&  ... }; run",
    "test-integration": "run(){ SEQ_DIALECT=\"${1}\" SEQ_DIALECT_VERSION=\"${2}\"; yarn prepare && ... }; run",
    "start": "db(){ SEQ_DIALECT=\"${1:}\"; SEQ_DIALECT_VERSION=\"${2:-latest}\"; $0 \"dev/start.js\" 2>/dev/null; }; db"
  }
```
- remove all the database-specific commands
- _start.js_ would be a controller/router to spin up the respective _/dev/dialects/<db>/<version>_
- no need to edit package.json
- dialects would return supported dialects
- `start` would also show supported dialects if one wasn't entered
- versions can be an enum: oldest, latest, native, <semver>

test/config/config.ts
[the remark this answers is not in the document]
WikiRik: I often look at PRs from my phone, and there the length is different. So the text does not show properly. It does indeed have some benefits but not enough to be used in Sequelize in my opinion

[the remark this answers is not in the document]
ephys: > but who codes with word-wrapping?

GitHub's split diff view uses word-wrapping

Consider using something like this if you need to split sections:

```
// ============================================
//  Special Cases
// ============================================
```

I've also seen people use `-` as a lower level of importance:

```
// --------------------------------------------
//  Special Cases - Part 1
// --------------------------------------------
```

I don't think the many = are going to be screen reader friendly either but the text inside of it can be read, and remains readable with word-wrapping

package.json
ephys: > Latest may work, Oldest may work, but there might be an intermediary major version that doesn't work.
>
> Consider version 6, 7, 8, which are all breaking changes. 7 (or a minor version) could have removed a feature. 8 might have added it back.

This is what I mentioned in my other comment. The probability of this happening is incredibly low. Databases are huge software designed slowly and carefully, breaking changes aren't done last minute to then realize it was a bad idea and revert it next release.

> If you never test 7, there's no way of knowing.

Again, this is planned: A once a week CI run on a broader range of databases.

> The dev folder has already been out-of-date. I realized this when I saw the version number from tree. I don't think burying version numbers in files is going to make that issue any better.

It's out of date because it's already a lot of work to keep them up to date, which is the whole reason behind the change to oldest/latest.

- `oldest` doesn't need to change at all until the next major.
- `latest` can be kept up to date automatically thanks to renovate, something we couldn't do before because version numbers were hardcoded everywhere.

Adding intermediary versions to the `dev` folder, to the test config, brings back the issue of having to regularly update everything, which won't be done because it's a lot of work for very little benefit.

An important note that may influence your opinion: The `dev` folder is not used on the CI. It's for local development. The only time you'll need an intermediary version locally is when the once-a-week CI run tells us something is broken in that version. You can then easily boot a temporary intermediary instance of the database.

> discarding them for cosmetics.

Not cosmetics, maintenance work that will not be done because the time necessary to keep it up to date is not worth the almost nil benefit it brings

> At some point you might want to test against up-and-coming databases.

That's already the case, `mssql-latest` is the upcoming 2022 version

ephys:
I think we can consider closing this now, the different changes have for the most part been done through other PRs

WikiRik:
Agree, the DB2 README can be added in the PR for the CONTRIBUTING docs

PR #14608, opened 2022-06-08
build: migrate db start scripts to TS

ephys:
### Description Of Change

This PR resolves the issue introduced in https://github.com/sequelize/sequelize/pull/14407 where database start commands would throw an error because they attempted to load a TypeScript file.

This PR should resolve this comment on this other PR: https://github.com/sequelize/sequelize/pull/14604#discussion_r891980284

### DIFF
--- a/dev/db2/11.5/start.sh
+++ b/dev/db2/11.5/start.sh
@@ -25,7 +25,6 @@ if [ ! "$(sudo docker ps -q -f name=db2server)" ]; then
 	  sleep 20
 	  let "count=count+1"
 	done
-  echo "Local DB2-11.5 instance is ready for Sequelize tests."
 fi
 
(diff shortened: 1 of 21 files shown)


WikiRik (dismissed):

WikiRik (approved):

dev/db2/11.5/start.sh
```diff
@@ -27,3 +27,7 @@ if [ ! "$(sudo docker ps -q -f name=db2server)" ]; then
 	done
   echo "Local DB2-11.5 instance is ready for Sequelize tests."
```
WikiRik, line 28: Wait, this is double now

ephys: Woops!
I removed the one that is not always printed, so the message is also displayed if the DB was already running

WikiRik (approved):

PR #14610, opened 2022-06-09
feat: exports types to support typescript >= 4.5 nodenext module

Beace:
### Description Of Change

typescript has released [version >=4.5.x](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/ )，when use the latest `typescript@4.7.3` and `nodenext` moduleResolution throw this error.

image

my  tsconfig.json 
```json
"compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "rootDir": ".",
    "outDir": "output"
 },
```

This is because the `types` field needs to be added to the `exports` field of package.json when module is `nodenext`. like this.

image

hope this new feature add in next v6 patch version and the we can enjoin the "nodenext".

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,9 +22,13 @@
   "exports": {
     ".": {
       "import": "./lib/index.mjs",
-      "require": "./lib/index.js"
+      "require": "./lib/index.js",
+      "types": "./types/index.d.ts"
     },
-    "./_non-semver-use-at-your-own-risk_/*": "./lib/*",
+    "./_non-semver-use-at-your-own-risk_/*": {
+      "default": "./lib/*",
+      "types": "./types/*.d.ts"
+    },
     "./package.json": "./package.json"
   },
   "engines": {


package.json
```diff
@@ -22,7 +22,8 @@
   "exports": {
     ".": {
       "import": "./lib/index.mjs",
-      "require": "./lib/index.js"
+      "require": "./lib/index.js",
+      "types": "./types/index.d.ts"
     },
     "./_non-semver-use-at-your-own-risk_/*": "./lib/*",
```
ephys, line 28: this needs to be updated too

```suggestion
    "./_non-semver-use-at-your-own-risk_/*": {
      "default": "./lib/*",
      "types": "./types/*.d.ts"
    },
```

ephys (changes requested):
Thank you for the PR!

I tried this PR against the other module resolutions in a test project and it doesn't cause any issue and does work with NodeNext

One note, I think the `./_non-semver-use-at-your-own-risk_/*` export needs to be updated to

Also note that this PR is against v7, another PR against v6 will also need to be done to resolve your issue

package.json
Beace: updated

ephys:
> How to create another PR against v6 ? Specify target branch as V6?

The exports have changed a lot between v6 and v7. You would need to create a new branch starting from our v6 branch (instead of main), do your changes, then open the PR against v6

package.json
```diff
@@ -22,9 +22,13 @@
   "exports": {
     ".": {
       "import": "./lib/index.mjs",
-      "require": "./lib/index.js"
+      "require": "./lib/index.js",
+      "types": "./types/index.d.ts"
+    },
+    "./_non-semver-use-at-your-own-risk_/*": {
+      "default": "./lib/*",
+      "types": "./types/*.d.ts"
```
ephys, line 30: I wrote this too quickly. Tested again with this code and `.d.ts` must not be specified here

```suggestion
      "types": "./types/*"
```

Beace: +1 done

ephys: I think something went wrong, I still have the old code in the diff 😅

Beace:
for v7 solution

https://github.com/Beace/sequelize-example  I have summarized several common module methods , and after testing, I feel that the following should be recommended exports in v7.

```json
"exports": {
    ".": {
      "import": "./lib/index.mjs",
      "require": "./lib/index.js",
      "types": "./types/index.d.ts"
    },
    "./_non-semver-use-at-your-own-risk_/*": {
      "default": "./lib/*",
      "types": "./types/*.d.ts"
    },
    "./package.json": "./package.json"
  },
```

ephys:
I've tested this again

I did the following import:

```ts
import { LOCK } from '@sequelize/core/_non-semver-use-at-your-own-risk_/transaction.js';
```

I get no error with the following config:

```json
"exports": {
  ".": {
    "import": "./lib/index.mjs",
    "require": "./lib/index.js",
    "types": "./types/index.d.ts"
  },
  "./_non-semver-use-at-your-own-risk_/*": {
    "default": "./lib/*",
    "types": "./types/*"
  },
  "./package.json": "./package.json"
}
```

But if I specify `.d.ts`, I get an error:

```json
"exports": {
  ".": {
    "import": "./lib/index.mjs",
    "require": "./lib/index.js",
    "types": "./types/index.d.ts"
  },
  "./_non-semver-use-at-your-own-risk_/*": {
    "default": "./lib/*",
    "types": "./types/*.d.ts"
  },
  "./package.json": "./package.json"
}
```

results in

> Could not find a declaration file for module '@sequelize/core/_non-semver-use-at-your-own-risk_/transaction.js'.

Tested with `moduleResolution` `NodeNext` and `Node16`

Beace:
Maybe `paths` needed when using a nonexist path https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping

Beace:
like this

```json

{
  "extends": "./tsconfig.base.json",
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@sequelize/core/_non-semver-use-at-your-own-risk_/*": ["node_modules/@sequelize/core/types/*"]
    }
  }
}

```

ephys:
Isn't the goal is to have a configuration that works without the user needing to configure `paths` (assuming the user uses Node16/NodeNext as they're the only resolvers that support node's `exports` config)?

You can already configure `paths` today. It bypasses `exports` so any test that uses it does not actually test the `exports` config

Also, reading the documentation, I noticed `default` must always come last in the configuration

Issue #12889, opened 2020-12-04
Duplicated unique constraints and indexes on sequelize.sync({ alter: true })


JekRock:
## Issue Description
When there is the `unique: true` property for any model field, sequelize creates a new unique constraint and index on each `sequelize.sync({ alter: true })`
### What are you doing?
Create any model with a unique field like that:
```typescript
  public initModel({ sequelize }: { sequelize: Sequelize }): ModelCtor<ExampleModel> {
    const exampleModel = sequelize.define<ExampleModel>(
      "ExampleModel",
      {
        id: {
          allowNull: false,
          defaultValue: DataTypes.UUIDV1,
          primaryKey: true,
          type: DataTypes.UUID,
        },
        exampleField: {
          unique: true,
          allowNull: false,
          type: DataTypes.STRING,
        },
    );
    return exampleModel;
  }
```
Then call `sequelize.sync({ alter: true })`.

### What do you expect to happen?
Only single unique constraint and index are created after the each `sequelize.sync({ alter: true })` call

### What is actually happening?
For each `sequelize.sync({ alter: true })` call there will be a new constraint named `ExampleModel_exampleField_key1` (the next will be `ExampleModel_exampleField_key2` and so on) and a new index called the same.
### Environment

- Sequelize version: 6.3.5 
- Node.js version: 14.15.0 
- Operating System: WSL2 Ubuntu 20.04
- If TypeScript related: TypeScript version: 4.0.2

### How does this problem relate to dialects?

- [ ] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [x] I don't know, I was using PostgreSQL, with connector library version 8.3.3 and database version 12.4

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [x] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

yusufkhan07:
Facing the same issue.

sandy912:
Facing the same issue with dialect Postgres.

tiagowippel:
Facing the same issue here too.

gkourpa:
Facing the same issue.

FuketsuBaka:
Facing the same issue.

EgoBrainProgrammer:
Facing the same issue. Postgres.

[...]

PR #14619, opened 2022-06-10
feat: fix multiple bugs related to `sync()` and indexes. Add `Model.getIndexes()`

ephys:
### Pull Request Checklist

### Description Of Change

Closes https://github.com/sequelize/sequelize/issues/12889

This PR fixes the following bugs:

- DB2 forced columns that had a unique index to be non-null, this has been removed (you still need to make your column non-null, but we don't silently make that happen).
- When defining an index through the `indexes` option on a model, then calling `sync()`, the dialect would *modify* the model to mark all attributes that are part of the index as unique. (DB2 only)
- Calling `sync({ alter: true })` with an attribute that has `unique: true` set would generate a query such as `ALTER TABLE users ALTER COLUMN emails SET VARCHAR UNIQUE`. DB2 and MSSQL do not allow `UNIQUE` to be specified there. This has been changed to use a separate `CREATE INDEX` query.
- SQLite supports specifying an index name during `CREATE TABLE`, but that name is ignored by sqlite. Causing subsequent `sync()` to try to recreate the index. The first `sync` does not try to create the index as part of `CREATE TABLE` anymore, but executes `CREATE INDEX` afterwards to create them (sqlite only, other dialects still do it during `CREATE TABLE`).
- This PR adds `Model.getIndexes()` which lists all indexes defined on the model (it regroups all indexes, the ones defined through `ModelOptions.indexes`, as well as the ones coming from `ModelAttributeColumnOptions.unique`)

This PR blocks https://github.com/sequelize/sequelize/pull/14572#issuecomment-1152578770

### TODO

- [x] Add tests for https://github.com/sequelize/sequelize/issues/12889
- [x] Add `sync(alter)` tests for *adding* unique constraints after the first `sync()`

### DIFF
--- a/src/dialects/sqlite/query-generator.js
+++ b/src/dialects/sqlite/query-generator.js
@@ -60,13 +60,26 @@ export class SqliteQueryGenerator extends MySqlQueryGenerator {
     let attrStr = attrArray.join(', ');
     const pkString = primaryKeys.map(pk => this.quoteIdentifier(pk)).join(', ');
 
-    if (options.uniqueKeys) {
-      _.each(options.uniqueKeys, columns => {
-        if (columns.customIndex) {
-          attrStr += `, UNIQUE (${columns.fields.map(field => this.quoteIdentifier(field)).join(', ')})`;
-        }
-      });
-    }
+    // sqlite has a bug where using CONSTRAINT constraint_name UNIQUE during CREATE TABLE
+    //  does not respect the provided constraint name
+    //  and uses sqlite_autoindex_ as the name of the constraint instead.
+    //  CREATE UNIQUE INDEX does not have this issue, so we're using that instead
+    //
+    // if (options.uniqueKeys) {
+    //   _.each(options.uniqueKeys, (columns, indexName) => {
+    //     if (columns.customIndex) {
+    //       if (typeof indexName !== 'string') {
+    //         indexName = Utils.generateIndexName(tableName, columns);
+    //       }
+    //
+    //       attrStr += `, CONSTRAINT ${
+    //         this.quoteIdentifier(indexName)
+    //       } UNIQUE (${
+    //         columns.fields.map(field => this.quoteIdentifier(field)).join(', ')
+    //       })`;
+    //     }
+    //   });
+    // }
 
     if (pkString.length > 0) {
       attrStr += `, PRIMARY KEY (${pkString})`;
--- a/src/model.js
+++ b/src/model.js
@@ -2730,12 +2730,6 @@ Specify a different name for either index to resolve this issue.`);
             }
           }
 
-          const firstUniqueKey = Object.values(model.uniqueKeys).find(c => c.fields.length > 0);
-
-          if (firstUniqueKey && firstUniqueKey.fields) {
-            upsertKeys.push(...firstUniqueKey.fields);
-          }
-
           options.upsertKeys = upsertKeys.length > 0
             ? upsertKeys
             : Object.values(model.primaryKeys).map(x => x.field);
--- a/src/model.js
+++ b/src/model.js
@@ -789,6 +789,10 @@ ${associationOwner._getAssociationDebugList()}`);
     ];
   }
 
+  static get _indexes() {
+    throw new Error('Model._indexes has been replaced with Model.getIndexes()');
+  }
+
   static _nameIndex(newIndex) {
     if (Object.prototype.hasOwnProperty.call(newIndex, 'name')) {
       return newIndex;
--- a/src/model.js
+++ b/src/model.js
@@ -783,12 +783,34 @@ ${associationOwner._getAssociationDebugList()}`);
 
   static getIndexes() {
     return [
-      ...this._manualIndexes,
-      ...this._attributeIndexes,
-      // TODO: (during TS migration) stop using 'uniqueKeys' internally in favor of '_attributeIndexes'
-      //       make 'uniqueKeys' a getter the filters getIndexes?
-      ...Object.values(this.uniqueKeys),
+      ...(this._manualIndexes ?? []),
+      ...(this._attributeIndexes ?? []),
+      ...(this.uniqueKeys ? Object.values(this.uniqueKeys) : []),
     ];
+  }
+
+  static _nameIndex(newIndex) {
+    if (Object.prototype.hasOwnProperty.call(newIndex, 'name')) {
+      return newIndex;
+    }
+
+    const newName = Utils.generateIndexName(this.getTableName(), newIndex);
+
+    // TODO: check for collisions on *all* models, not just this one, as index names are global.
+    for (const index of this.getIndexes()) {
+      if (index.name === newName) {
+        throw new Error(`Sequelize tried to give the name "${newName}" to index:
+${NodeUtil.inspect(newIndex)}
+on model "${this.name}", but that name is already taken by index:
+${NodeUtil.inspect(index)}
+
+Specify a different name for either index to resolve this issue.`);
+      }
+    }
+
+    newIndex.name = newName;
+
+    return newIndex;
   }
 
   /**
@@ -950,13 +972,12 @@ ${associationOwner._getAssociationDebugList()}`);
       return attribute;
     });
 
-    const tableName = this.getTableName();
     this._manualIndexes = this.options.indexes
-      .map(index => Utils.nameIndex(this._conformIndex(index), tableName));
+      .map(index => this._nameIndex(this._conformIndex(index)));
 
-    this.primaryKeys = {};
+    this.primaryKeys = Object.create(null);
     this._readOnlyAttributes = new Set();
-    this._timestampAttributes = {};
+    this._timestampAttributes = Object.create(null);
 
     // setup names of timestamp attributes
     if (this.options.timestamps) {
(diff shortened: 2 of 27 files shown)


ephys:
I've hit another rabbithole.
I've added tests to make sure the unique index is still created, but it isn't in mssql (which means it never was).

I think the best solution is going to be to unify indexes created by the `unique` option on attributes, with the indexes created by the `indexes` option in the table options.

Which leads to another issue: indexes created by `CREATE TABLE` and `ALTER COLUMN` are not explicitly named. The DBMS creates a random name. I don't see a way to specify the name.
When `sync({ alter: true })` tries to see if the index has already been created, it can't match the name generated by sequelize with the name generated by the DBMS.

I think we'll have to make `sync()` generate all indexes using `CREATE INDEX`, instead of sometimes using `CREATE TABLE`. Slightly slower, but the name is unified, which leads to less bugs.

ephys:
You're going to laugh. I've deadlocked my PRs.

One of the tests is failing because DB2 is recreating the table instead of altering it.
This is a DB2 bug I fixed in https://github.com/sequelize/sequelize/pull/14572
That PR is blocked because it relies on this one to fix DB2 bugs, and this one is blocked because it relies on that one.

This isn't a regression, so I'm requesting to force-merge this PR, and continue over at https://github.com/sequelize/sequelize/pull/14572

WikiRik:
Can we temporarily disable the failing unit tests so only DB2 checks fail for this PR and re-enable them in the other PR? Not a fan of force merging

Will review tomorrow

ephys:
> Can we temporarily disable the failing unit tests

I fixed them, these weren't supposed to happen.
The only tests that are 'supposed' to fail are the DB2 ones, which funnily enough I cannot reproduce locally so there is even more flakiness going on (still need to complete https://github.com/sequelize/sequelize/pull/14453)

Here is the complete list of failing DB2 tests:

```
  1) [DB2] Model.sync & Sequelize.sync
       does not alter table if data type does not change:
     TypeError: Cannot read properties of null (reading 'dataValues')
      at Context.<anonymous> (test/integration/model/sync.test.js:142:17)

==> Unsure what is going on there, can't reproduce locally. Will investigate more before merging

  2) [DB2] Model.sync & Sequelize.sync
       adds missing unique columns to existing tables:
      AssertionError: expected [] to have a length of 1 but got +0
      + expected - actual
      -0
      +1
      at Proxy.assertLength (node_modules/chai/lib/chai/core/assertions.js:2144:10)
      at doAsserterAsyncAndAddThen (node_modules/chai-as-promised/lib/chai-as-promised.js:289:22)
      at Proxy.<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:272:28)
      at Proxy.overwritingChainableMethodWrapper (node_modules/chai/lib/chai/utils/overwriteChainableMethod.js:60:34)
      at Proxy.chainableMethodWrapper (node_modules/chai/lib/chai/utils/addChainableMethod.js:113:49)
      at Context.<anonymous> (test/integration/model/sync.test.js:418:34)

==> Fixed by https://github.com/sequelize/sequelize/pull/14572, caused by DB2 overwriting tables when calling `sync()`

  3) [DB2] QueryInterface
       changeColumn
         should support schemas:
     [IBM][CLI Driver][DB2/LINUXX8664] SQL0205N  Column, attribute, or period "currency" is not defined in "archive.users".  SQLSTATE=42703
  Error: 
      at Db2Query._run (src/dialects/db2/query.js:45:22)
      at Db2Query.run (src/dialects/db2/query.js:189:23)
      at /home/runner/work/sequelize/sequelize/src/sequelize.js:638:28
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at Sequelize.queryRaw (src/sequelize.js:616:12)
      at Context.<anonymous> (test/integration/query-interface/changeColumn.test.js:37:7)

==> Typical flaky test, see https://github.com/sequelize/sequelize/pull/14453

  4) [DB2] QueryInterface
       removeColumn
         (with a schema)
           [Flaky] should be able to remove a column with a default value:
     [IBM][CLI Driver][DB2/LINUXX8664] SQL0205N  Column, attribute, or period "firstName" is not defined in "archive.users".  SQLSTATE=42703
  Error: 
      at Db2Query._run (src/dialects/db2/query.js:45:22)
      at Db2Query.run (src/dialects/db2/query.js:189:23)
      at /home/runner/work/sequelize/sequelize/src/sequelize.js:638:28
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at Sequelize.queryRaw (src/sequelize.js:616:12)
      at Context.<anonymous> (test/integration/query-interface/removeColumn.test.js:121:9)

==> Typical flaky test, see https://github.com/sequelize/sequelize/pull/14453

  5) [DB2] QueryInterface
       removeColumn
         (with a schema)
           should be able to remove a column without default value:
     [IBM][CLI Driver][DB2/LINUXX8664] SQL0205N  Column, attribute, or period "lastName" is not defined in "archive.users".  SQLSTATE=42703
  Error: 
      at Db2Query._run (src/dialects/db2/query.js:45:22)
      at Db2Query.run (src/dialects/db2/query.js:189:23)
      at /home/runner/work/sequelize/sequelize/src/sequelize.js:638:28
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at Sequelize.queryRaw (src/sequelize.js:616:12)
      at Context.<anonymous> (test/integration/query-interface/removeColumn.test.js:135:9)

==> Typical flaky test, see https://github.com/sequelize/sequelize/pull/14453
```

src/dialects/sqlite/query-generator.js
```diff
@@ -60,13 +60,26 @@ export class SqliteQueryGenerator extends MySqlQueryGenerator {
...
-          attrStr += `, UNIQUE (${columns.fields.map(field => this.quoteIdentifier(field)).join(', ')})`;
-        }
-      });
-    }
+    // sqlite has a bug where using CONSTRAINT constraint_name UNIQUE during CREATE TABLE
+    //  does not respect the provided constraint name
+    //  and uses sqlite_autoindex_ as the name of the constraint instead.
+    //  CREATE UNIQUE INDEX does not have this issue, so we're using that instead
+    //
+    // if (options.uniqueKeys) {
```
sdepold, line 68: What are we doing with this code?

src/model.js
```diff
@@ -771,6 +771,52 @@ ${associationOwner._getAssociationDebugList()}`);
...
+  static getIndexes() {
+    return [
+      ...(this._manualIndexes ?? []),
+      ...(this._attributeIndexes ?? []),
+      ...(this.uniqueKeys ? Object.values(this.uniqueKeys) : []),
+    ];
+  }
+
+  static get _indexes() {
+    throw new Error('Model._indexes has been replaced with Model.getIndexes()');
```
sdepold, line 793: This will land in v7, right? => Can we note this under breaking changes? Should we deprecate this in a new v6 release?

sdepold (commented):
I'm always surprised by how much you know about the library and in which corners you can stir :)

src/dialects/sqlite/query-generator.js
ephys: I'm typically against keeping dead code but I'm hesitant about deleting it, and keep it as a reminder that we should monitor every now and then whether the bug has been fixed in sqlite, and the uncomment this

src/model.js
ephys: Yep, v7 only

It's an undocumented, untyped, weakly private API. I've added the error in case people were using it, and we can of course add it in our list of breaking changes but I don't know how to go about deprecating this in v6 as we use it internally and us accessing the property would print the deprecation warning

src/dialects/sqlite/query-generator.js
WikiRik: Is there somewhere that we can track this bug?

src/model.js
```diff
@@ -771,6 +771,52 @@ ${associationOwner._getAssociationDebugList()}`);
...
+  }
+
+  static _nameIndex(newIndex) {
+    if (Object.prototype.hasOwnProperty.call(newIndex, 'name')) {
+      return newIndex;
+    }
+
+    const newName = Utils.generateIndexName(this.getTableName(), newIndex);
+
+    // TODO: check for collisions on *all* models, not just this one, as index names are global.
```
WikiRik, line 803: This is for a future PR, right?

src/model.js
```diff
@@ -1102,41 +1149,49 @@ ${associationOwner._getAssociationDebugList()}`);
...
 
-        const idx = this.uniqueKeys[idxName] || { fields: [] };
+        const index = definition.unique.name && this.uniqueKeys[definition.unique.name]
+          ? this.uniqueKeys[definition.unique.name]
+          : { fields: [] };
+
+        index.fields.push(definition.field);
+        index.msg = index.msg || definition.unique.msg || null;
+
+        // TODO: remove this 'column'? It does not work with composite indexes, and is only used by db2 which should use fields instead.
```
WikiRik, line 1167: If we remove this, that would be a breaking change right? Might be good to decide on that before releasing v7

test/integration/model/sync.test.js
```diff
@@ -1,408 +1,436 @@
...
+      badgeNumber: { type: DataTypes.INTEGER, field: 'badge_number' },
+    });
+
+    await sequelize.sync({ alter: true });
+    const data = await testSync.describe();
+    expect(data).to.have.ownProperty('badge_number');
+    expect(data).not.to.have.ownProperty('badgeNumber');
+  });
+
+  // IBM i can't alter INTEGER -> STRING
```
WikiRik, line 112: Does ibmi have something similar that we could test on?

src/dialects/sqlite/query-generator.js
ephys: I looked on [sqlite's issue tracker](https://www.sqlite.org/src/wiki?name=Bug+Reports) but couldn't find anything related to this

src/model.js
ephys: Yes, it's a lot more complicated than just checking on the same model because they can be scoped by schema, with different rules depending on the database

src/model.js
ephys: Maybe. It was added a long time ago (2016 (wow 2016 was a long time ago?)) in this commit https://github.com/sequelize/sequelize/commit/47dc46a7f4b00e06687c0bdea9a53a5cc2ba18fd#diff-305dbcc98ad8fc959dd0c0be51280893962af34c2d8b18abc871cffc2d80c5f9R858

It's possible some people rely on it, so we should consider it to be breaking. Not a huge priority for v7, keeping it doesn't really impact us at all

test/integration/model/sync.test.js
ephys: I don't know :/ Maybe it can alter between two different string types but can't check without a proper test DB
Note that this wasn't changed by this PR, it's been this way since the initial ibmi PR: https://github.com/sequelize/sequelize/pull/14619/files/77c2f7d81d97413f5a8c7e2ff7d43b2fd2b44ade#diff-81b952d2bbfce09ba3fc0712ea8248f597d3d3421d819a1c69eb06f68fcf318eL108

WikiRik: Ah woops, missed that part of the diff. Then we'll hopefully get another chance to look into it once we find somebody that can test on this

src/model.js
WikiRik: True, let's just leave it in for now until somebody wants to work on this

src/dialects/sqlite/query-generator.js
WikiRik: Let's keep this in for now and try to make a bug report for sqlite at a later stage

src/model.js
WikiRik: Technically we could move it to a new function in v6, reference that one internally and add a deprecation message to _indexes right?

ephys: We could, but then removing that function in V7 becomes the breaking change (or of we use `getIndexes` like in V7, then the breaking change is that it's behavior is not the same in V6 vs v7 because _indexes and getIndexes are not strictly equivalent)

WikiRik (approved):

ephys:
Let me handle the merging of this PR, I'll add a detailed message for the changelog

PR #14620, opened 2022-06-11
feat: exports types to support typescript >= 4.5 nodenext module(v6)

Beace:
### Description Of Change
v6 for #14610


ephys (changes requested):
Can you add them for the other exports too?

I have not tested this yet but it should be correct

```json
    "./lib/*": {
      "default": "./lib/*.js",
      "types": "./types/*.d.ts"
    },
    "./lib/errors": {
      "default": "./lib/errors/index.js",
      "types": "./types/errors/index.d.ts"
    },
    "./*": "./*"
```

(our v6 export config looks weird but, well, backward compatibility)

Beace:
> Thanks :) Can you add them for the other exports too?
> 
> I have not tested this yet but it should be correct
> 
> ```json
>     "./lib/*": {
>       "default": "./lib/*.js",
>       "types": "./types/*.d.ts"
>     },
>     "./lib/errors": {
>       "default": "./lib/errors/index.js",
>       "types": "./types/errors/index.d.ts"
>     },
>     "./*": "./*"
> ```
> 
> (our v6 export config looks weird but, well, backward compatibility)


updated. I always forget about types other than `index.d.ts` 😓

Beace:
@ephys   Sorry, How long will sequelize release a new version, my team want to use it in new projects.

package.json
```diff
@@ -22,10 +22,17 @@
   "exports": {
     ".": {
       "import": "./lib/index.mjs",
-      "require": "./lib/index.js"
+      "require": "./lib/index.js",
+      "types": "./types/index.d.ts"
```
WikiRik, line 26:
```suggestion
      "types": "./types/index"
```
This would probably also need to be without the extension as mentioned in #14610

package.json
```diff
@@ -22,10 +22,17 @@
...
+    },
+    "./lib/*": {
+      "default": "./lib/*.js",
+      "types": "./types/*.d.ts"
```
WikiRik, line 30:
```suggestion
      "types": "./types/*"
```
This would probably also need to be without the extension as mentioned in #14610

package.json
```diff
@@ -22,10 +22,17 @@
...
+    },
+    "./lib/errors": {
+      "default": "./lib/errors/index.js",
+      "types": "./types/errors/index.d.ts"
```
WikiRik, line 34:
```suggestion
      "types": "./types/errors/index"
```
This would probably also need to be without the extension as mentioned in #14610

package.json
ephys: I need to do one last test but the current version should be correct. Whether the extension is present or not should follow `default` (or `require` / `import`)

package.json
ephys: Here the extension must be specified because `default` specifies it (but I need to run a battery of tests on this PR to be 100% sure)

ephys:
> @ephys Sorry, How long will sequelize release a new version, my team want to use it in new projects.

As soon as this is merged (automatically), just need to take the time to test with a dummy project to be sure :)

package.json
Beace: i see the typescript [package-json-exports-imports-and-self-referencing](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#package-json-exports-imports-and-self-referencing) docs, the `types` always with `.d.ts`

ephys:
After further testing, this still breaks because of the `./` mapping

we can fix it by making it this:

```json
  "exports": {
    ".": {
      "import": "./lib/index.mjs",
      "require": "./lib/index.js",
      "types": "./types/index.d.ts"
    },
    "./lib/*": {
      "default": "./lib/*.js",
      "types": "./types/*.d.ts"
    },
    "./lib/errors": {
      "default": "./lib/errors/index.js",
      "types": "./types/errors/index.d.ts"
    },
    "./package.json": "./package.json"
  },
```

Beace:
> I'm looking into this right now
> 
> By the way, the v7 version is still outdated, could you apply the suggestion so the fix is available in v7 too? #14610

I haven't actually used v7. I will build an environment for V7 testing.

Beace:
> I'm looking into this right now
> 
> By the way, the v7 version is still outdated, could you apply the suggestion so the fix is available in v7 too? #14610

@ephys  https://github.com/Beace/sequelize-example I have summarized several common module methods, and after testing, I feel that the following should be recommended exports.

```json
"exports": {
    ".": {
      "import": "./lib/index.mjs",
      "require": "./lib/index.js",
      "types": "./types/index.d.ts"
    },
    "./_non-semver-use-at-your-own-risk_/*": {
      "default": "./lib/*",
      "types": "./types/*.d.ts"
    },
    "./package.json": "./package.json"
  },
```

Beace:
updated

> After further testing, this still breaks because of the `./` mapping
> 
> we can fix it by making it this:
> 
> ```json
>   "exports": {
>     ".": {
>       "import": "./lib/index.mjs",
>       "require": "./lib/index.js",
>       "types": "./types/index.d.ts"
>     },
>     "./lib/*": {
>       "default": "./lib/*.js",
>       "types": "./types/*.d.ts"
>     },
>     "./lib/errors": {
>       "default": "./lib/errors/index.js",
>       "types": "./types/errors/index.d.ts"
>     },
>     "./package.json": "./package.json"
>   },
> ```

updated

ephys (approved):
This looks good to me 👍

PR #14641, opened 2022-06-14
fix(typings): correct options for `bulkDelete()` (v6)

### LINKED ISSUE #14635: Missing `bulkDelete` typings
## Issue Creation Checklist

## Issue Description

According to the [documentation](https://sequelize.org/api/v6/class/src/dialects/abstract/query-interface.js~queryinterface#instance-method-bulkDelete), the `queryBuilder.bulkDelete` method accepts the boolean values `truncate`, `cascade` and `restartIdentity` in the `options` arguments. But this is not reflected in Typescript.

A `BulkDeleteOptions` interface extending `QueryOptions` with the parameters `truncate`, `cascade` and `restartIdentity` should be added.

A PR regarding this issue (#12750) was already submitted but it got closed by a bot.

### Additional context

image

image

- [ ] I don't know.

WikiRik:
### Pull Request Checklist

### Description Of Change

Recreation of #12750 which added typings for the `bulkDelete` options

Closes #12750 
Fixes #14635


src/dialects/abstract/query-interface.d.ts
```diff
@@ -214,6 +214,15 @@ export interface IndexesOptions {
 
 export interface QueryInterfaceIndexOptions extends IndexesOptions, QueryInterfaceOptions {}
 
+/**
+ * Interface for bulk delete options
+ */
+export interface QueryInterfaceBulkDeleteOptions {
+  truncate?: boolean
+  cascade?: boolean
+  restartIdentity?: boolean
```
ephys, line 223:
```suggestion
  truncate?: boolean;
  cascade?: boolean;
  restartIdentity?: boolean;
```

src/dialects/abstract/query-interface.d.ts
```diff
@@ -214,6 +214,15 @@ export interface IndexesOptions {
 
 export interface QueryInterfaceIndexOptions extends IndexesOptions, QueryInterfaceOptions {}
 
+/**
+ * Interface for bulk delete options
+ */
+export interface QueryInterfaceBulkDeleteOptions {
```
ephys, line 220: These options are passed to `Sequelize#query` so should inherit from `QueryOptions`

src/dialects/abstract/query-interface.d.ts
ephys, line 223: in non-truncate mode, `limit` and `offset` (db2) are accepted

ephys (changes requested):
In a future major release, we should consider splitting `bulkDelete` and `truncate`. They are different operations that accept different options

src/dialects/abstract/query-interface.d.ts
WikiRik: Is there some TypeScript magic that would be useful for only allowing `limit` and `offset` when `truncate` is false?

WikiRik:
Btw; we need these typings for main as well

src/dialects/abstract/query-interface.d.ts
WikiRik: That's a good one

src/dialects/abstract/query-interface.d.ts
WikiRik: Is this something we want to be picked up by eslint?

ephys: Is it now that .d.ts is linted? If not, that's probably a bug with the eslint rule because it's part of our config

ephys: Nothing very clean unfortunately. I would just add them and document in which mode they apply

We can fix this in the future by removing truncate ability from `.bulkDelete`, and have `.truncate` instead. It's two very different SQL operations after all

PR #14649, opened 2022-06-16
build: run eslint on `.d.ts`, `/dev` and `test/types`

ephys:
During #13930, we initially opted to not lint `.d.ts` because they were temporary files. 6 months later, they are still in our codebase because migrating to TS takes a long time, and we still get PRs for them. Applying our linter to these files caught a couple of minor typing bugs.

We also skipped `/dev` to reduce the size of the PR. I think it's time to update these files too.

Not sure why we did not lint `test/types`. Probably the same reason.


src/dialects/abstract/query.d.ts
```diff
@@ -1,9 +1,7 @@
-import { Connection } from './connection-manager';
-import { QueryTypes } from '../../query-types';
-import { Model, ModelStatic, IncludeOptions } from '../../model';
-import { Sequelize } from '../../sequelize';
-
-type BindOrReplacements = { [key: string]: unknown } | unknown[];
```
WikiRik, line 6: In this file some additional things are removed, such as the BindOrReplacement and replacementFuncType types and some TypeDoc is changed. Is that specific to the linting change?

WikiRik (commented):
One file where I was not so certain about but the others seem to look fine upon an initial scan

src/dialects/abstract/query.d.ts
ephys: They became unused after the implementation of the new bind parameters :)

WikiRik (approved):

PR #14652, opened 2022-06-16
feat: Migrate `ConnectionManager` to TypeScript

### LINKED ISSUE #13660: [Feature Request] Expose pool diagnostic information
## Issue Creation Checklist

## Feature Description

After upgrading to Sequelize 5/6 (from 4), we lost access to the `Sequelize.ConnectionManager`'s pool property. This makes sense, since the pool's implementation is now an internal detail of Sequelize. We were using this to access information about the pool, which we would periodically report to DataDog for diagnostic purposes. Here's how we do this after the upgrade (after adding sequelize-pool to our package.json):

```typescript
  const pool = (instance.connectionManager as any).pool as Pool<any>;
  registerGaugeCallback("sequelize.pool.size", {}, () => pool.size);
  registerGaugeCallback("sequelize.pool.available", {}, () => pool.available);
  registerGaugeCallback("sequelize.pool.borrowed", {}, () => pool.using);
  registerGaugeCallback("sequelize.pool.pending", {}, () => pool.waiting);
```

`registerGaugeCallback` sets up a timer that periodically polls those values and reports them to DataDog.

It would be great to expose these variables through some sort of `PoolDiagnostic` interface, so then the pool implementation details could continue to remain an internal detail of Sequelize.

It's not really a problem. Implementing this feature request would allow us to be less careful about upgrades, in the sense that we wouldn't need to re-check our usage of these internals (i.e. whether this usage is still valid) across upgrades.

Ideally, a new getter is added to Sequelize.ConnectionManager, maybe called something like poolDiagnostics, which just returns the following interface:

```typescript
export interface PoolDiagnostics {
    size: number;
    available: number;
    using: number;
    waiting: number;
    // any other diagnostics that others might use could be added on here
}
```

Here's a sample implementation, which would go in `Sequelize.ConnectionManager`:
```typescript
public get poolDiagnostics(): PoolDiagnostics {
    return {
        size: pool.size,
        available: pool.available,
        using: pool.using,
        waiting: pool.waiting,
    }
}
```

These examples are TypeScript, but they should translate well to JavaScript.

As of recent Sequelize versions, it hides the implementation of the pool. Adding this to Sequelize would allow us to continue to not care about the underlying pool implementation.

At the top of this, I laid out the hack we're using to currently work around this.

### Additional context

None that I can think of at the moment.

- [ ] Yes. This feature only applies to the following dialect(s): XXX, YYY, ZZZ

ephys:
## Pull Request Checklist

## Description Of Change

This PR migrates the `AbstractConnectionManager` to TypeScript. As usual it isn't just a type addition but a small refactor to work better with TypeScript.

### Notable Change 1: Pool Normalization

A first notable change is the unification of how the connection pool works with and without replication.
Before this PR, if read replication was disabled, `connectionManager.pool` was an instance of `sequelize-pool`. If read replication was enabled, `connectionManager.pool` was an object that mimicked `sequelize-pool` and used two pools internally (one for read, one for write).

After this PR, `connectionManager.pool` is always an instance of `ReplicationPool` (new class) which dispatches to its internal `read` and `write` pools. If read replication is disabled, the `read` pool is disabled and the `write` pool is used for reads too.

This change makes the type much easier to work with.

### Notable Change 2: Sequelize option priority

Which parameter takes priority in the Sequelize constructor is inconsistent:

```js
// The following resulted in 'sequelize.config.username' being equal to 'user1'

new Sequelize('db1', 'user1', 'pwd1', {
  username: 'user2',
});

// The following resulted in 'sequelize.config.username' being equal to 'user2'

new Sequelize('mysql://user1:pwd1@localhost:1234/dbname', {
  username: 'user2',
});
```

After this PR, both versions prioritize the option bag over the string parameters. Both result in `user2` being used.

### Notable Change 3: Fixing another MySQL schema bug

Moved to https://github.com/sequelize/sequelize/pull/14657. The code is still in this PR but that other PR should be merged first.

---

Closes #13660

## Squash Commit Message Body

If this is merged by someone else than me, here is the message body to use for the squash commit (don't forget the empty line at the start of the body, per conventional commits conventions):

```

BREAKING CHANGE: The signature of `connectionManager.pool` when read replication is disabled is now the same as when read replication is enabled. If read replication is disabled, you can access the actual pool through `connectionManager.pool.write`. If read replication is enabled, you can use both `connectionManager.pool.read` and `connectionManager.pool.write`.
```


ephys:
Moved to https://github.com/sequelize/sequelize/pull/14657

test/unit/configuration.test.ts
```diff
@@ -41,153 +44,296 @@ describe('Sequelize', () => {
...
       const config = sequelize.config;
 
-      expect(config.username).to.not.be.ok;
-      expect(config.password).to.be.null;
-    });
-
-    it('should work with no authentication options and passing additional options', () => {
-      const sequelize = new Sequelize('mysql://example.com:9821/dbname', {});
-      const config = sequelize.config;
+      const defaultPort: Record<Dialect, number> = {
```
WikiRik, line 107: Don't you want to use the getter for this?

test/integration/replication.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  getTestDialectTeaser,
+} from './support';
+
+const dialect = getTestDialect();
+describe(getTestDialectTeaser('Replication'), () => {
+  if (['sqlite', 'ibmi'].includes(dialect)) {
+    return;
+  }
+
+  describe('connection objects', () => {
```
WikiRik, line 18: Do we still want to test on connection strings?

test/integration/dialects/abstract/connection-manager.test.ts
```diff
@@ -0,0 +1,186 @@
...
+        .expect(stub.getCalls()[0].args[0])
+        .to.contain(
+          'This database engine version is not supported, please update your database server.',
+        );
+      stub.restore();
+    });
+  }
+
+  if (dialect !== 'sqlite') {
+    // sqlite overrides getConnection() and never uses _connect
```
WikiRik, line 136: Is this a change? Because the test did run on sqlite before right?

test/integration/dialects/abstract/connection-manager.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    chai.expect(connectStub.callCount).to.equal(4);
+
+    // First call is the get connection for DB versions - ignore
+    const calls = connectStub.getCalls();
+    chai.expect(calls[1].args[0].host).to.eql('slave1');
+    chai.expect(calls[2].args[0].host).to.eql('slave2');
+    chai.expect(calls[3].args[0].host).to.eql('slave1');
+  });
+
+  // sqlite does not have different database versions since it's bundled
```
WikiRik, line 105: Would still be nice to check, because maybe an older version of the connector might work but have an older database version. Or do we have a separate test for the minimum connector version?

src/dialects/abstract/index.ts
```diff
@@ -213,6 +215,10 @@ export abstract class AbstractDialect {
   }
 
   abstract createBindCollector(): BindCollector;
+
+  static getDefaultPort(): number {
```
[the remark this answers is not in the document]
ephys: It is tested here https://github.com/sequelize/sequelize/blob/c54745b7abd4e8adc8401fbec441d4b1f81c0b2f/test/unit/configuration.test.ts#L107-L116

test/unit/connection-manager.test.js
```diff
@@ -1,104 +0,0 @@
...
-      this.dialect = {
-        connectionManager: {
-          connect: sinon.stub().resolves(this.connection),
-        },
-      };
-
-      this.sequelize = Support.createSequelizeInstance();
-    });
-
-    it('should resolve connection on dialect connection manager', async function () {
```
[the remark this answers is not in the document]
ephys: This behavior has been removed because it doesn't really make sense. It checks that `ConnectionManager._connect` calls `connect` not on *itself* but on `dialect.connectionManager.connect`.

I think this is a legacy test from before Sequelize moved to using native classes

test/integration/pool.test.ts
```diff
@@ -1,33 +1,34 @@
...
-const sinon = require('sinon');
-
-const { Sequelize } = require('@sequelize/core');
-const delay = require('delay');
-
-function assertSameConnection(newConnection, oldConnection) {
+function assertSameConnection(newConnection: Connection, oldConnection: Connection) {
   switch (dialect) {
     case 'postgres':
+      // @ts-expect-error - processID not declared yet
```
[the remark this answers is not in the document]
ephys: I don't know how we'd type this. Each connection object is a completely different object for each dialect.

We could consider wrapping each `connection` object from dialects in our own `Connection` object that abstracts common elements (such as underlying ID) away

Definitely not in this PR, time will tell if it's worth the trouble

test/integration/dialects/mariadb/connection-manager.test.js
```diff
@@ -61,7 +61,15 @@ if (dialect === 'mariadb') {
...
-        const sequelize = new Sequelize('localhost', 'was', 'ddsd', Support.sequelize.options);
+        const sequelize = Support.createSequelizeInstance({
+          database: 'db',
+          username: 'was',
+          password: 'ddsd',
+        });
+
+        console.log(sequelize.config);
+        console.log(sequelize.options);
+
```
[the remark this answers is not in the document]
ephys: Weird that the linter didn't catch that

test/unit/configuration.test.ts
ephys: `dialect.getDefaultPort()`? No because it's an opportunity to test it

test/integration/replication.test.ts
ephys: It has been moved to unit tests. Unit tests check that connection strings are properly transformed into connection objects, then this suite tests that connection objects work properly

test/integration/dialects/abstract/connection-manager.test.ts
ephys: The old test suite [created its own ConnectionManager](https://github.com/sequelize/sequelize/blob/main/test/integration/dialects/abstract/connection-manager.test.js#L164-L167) instead of using the current dialect's.

So this was not really tested for any dialect except 'abstract', so no dialect

test/integration/dialects/abstract/connection-manager.test.ts
ephys: Let's keep it for another PR. It's not a bug introduced by this PR, but this PR uncovered it. This test was broken and always ran against AbstractConnectionManager instead of SQLiteConnectionManager - so this wasn't caught

src/dialects/abstract/index.ts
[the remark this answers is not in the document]
WikiRik: Yeah, I forgot to remove this comment. Thanks!

test/unit/configuration.test.ts
WikiRik: Yeah, ignore this. I was a bit confused before but this makes sense

test/integration/replication.test.ts
WikiRik: Ah, I probably forgot about those by the time I looked at this file 😅

WikiRik (approved):

WikiRik:
@ephys I'll leave the merging up to you

PR #14634, opened 2022-06-13
fix(postgres): use schema set in sequelize config by default

nholmes3:
### Pull Request Checklist

_Please make sure to review and check all of these items:_

- [x] Have you added new tests to prevent regressions?
- [x] Does `yarn test` or `yarn test-DIALECT` pass with this change (including linting)?
- [x] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [x] Did you update the typescript typings accordingly (if applicable)?
- [x] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)?

### Description Of Change

This is an attempt at fixing a bug where the Postgres Query Generator does not respect a custom schema name if it is provided.

[This comment](https://github.com/sequelize/cli/issues/284#issuecomment-573351082) on [Issue #248 of sequelize/cli](https://github.com/sequelize/cli/issues/284) describes the problem in detail.

PR #14657, opened 2022-06-17
fix(mysql,mariadb): remove extra `FROM` from showIndexes

ephys:
### Pull Request Checklist

### Description Of Change

Our implementation of MySQL schemas is very broken. MySQL does not have a concept of 'database', and when you see the word 'database' in the MySQL doc, it actually means what we understand as 'schema' in our codebase (`CREATE DATABASE` and `CREATE SCHEMA` are synonymous in MySQL).

This PR also fixes our MySQL implementation of `showIndexesQuery`.

[The query is defined as follows:](https://dev.mysql.com/doc/refman/8.0/en/show-index.html)

```
SHOW INDEX
  {FROM} tbl_name
  [{FROM} db_name]
```

Our implementation would generate this:

```sql
SHOW INDEX FROM `schema`.`tableName` FROM `databaseName`
```

Where `databaseName` is the name of the database from the connection credentials. The problem is that in MySQL, that is actually equivalent to doing this:

```sql
SHOW INDEX FROM `databaseName`.`tableName`
```

Which breaks because there is no schema called `databaseName`.

This PR fixes this by just removing the `` FROM `databaseName` `` part. The schema is already handled by our `quoteTable` function.


test/unit/dialects/mariadb/query-generator.test.js
```diff
@@ -750,9 +751,27 @@ if (dialect === 'mariadb') {
         {
           arguments: ['User'],
           expectation: 'SHOW INDEX FROM `User`',
-        }, {
+        },
+        {
+          arguments: [{ tableName: 'User', schema: 'schema' }],
+          expectation: 'SHOW INDEX FROM `schema`.`User`',
+        },
+        // FIXME: enable this test once fixed
```
WikiRik, line 759: When would we fix this?

test/unit/dialects/mysql/query-generator.test.js
```diff
@@ -700,9 +701,28 @@ if (dialect === 'mysql') {
         {
           arguments: ['User'],
           expectation: 'SHOW INDEX FROM `User`',
-        }, {
+        },
+        {
+          arguments: [{ tableName: 'User', schema: 'schema' }],
+          // FIXME: this is not the right way to handle schemas in MySQL, it should be `schema`.`User` like MariaDB
```
WikiRik, line 707: Same as in the MariaDB file, when do we want to fix this?

test/unit/dialects/mariadb/query-generator.test.js
ephys: Fixing it will like impact all query-generator methods so I want to do a follow up PR that fixes the issue everywhere, and adds tests for everything. Ideally before v7 is released, I would consider it to be a breaking change

There is also some overlap with #14634

WikiRik (approved):

PR #14668, opened 2022-06-21
fix(types): fix `findAll()` being treated as raw

ephys:
#12921 introduced a regression where `findAll()` without any parameter was understood as `findAll({ raw: true })` by TypeScript.

This PR fixes that


WikiRik (approved):

WikiRik (changes requested):
I should stop approving things if the tests look good but then fail 😅

WikiRik (approved):

PR #14670, opened 2022-06-21
meta: lint PR title

ephys:
## Description Of Change

Because we squash our PRs, it's important to make sure the PR title matches the conventional commits conventions.

I've tightly restricted our scopes. They're only useful for our changelog. I've limited them to only dialect names and `types`. That is:

- db2
- ibmi
- mariadb
- mssql
- mysql
- postgres
- snowflake
- sqlite
- types

Following the same principle, I've limited our commit types to only a handful of useful ones. I don't think it's very useful to determine whether a commit is build, style, ci, chore, refactor, etc. Only `fix` and `feat` are useful for semantic-release, the others are not very relevant. Here is what I propose to use:

- `fix`: A bug fix, used by semantic release for patch releases
- `feat`: A new feature, used by semantic release for minor releases
- `docs`: Documentation change, still its own commit type because it impacts our users (through jsdocs changes)
- `meta`: Anything else. Regroups everything that was previously in `build`, `style`, `ci`, `chore`, `refactor`, `test`. Doesn't impact our users.

If this is OK with everyone, the next steps (in this PR) are to:

- [ ] Figure out why renovate won't change the commit type
- [x] Remove commitlint from the project, we don't need to lint commits when we squash them


WikiRik:
Should we add an `abstract` scope for when it affects most/all dialects?

ephys:
The scope is not mandatory. If it affects a few dialects, `fix(mysql,mariadb)` should be valid (at least I saw a feature request for it). If it affects everything, not specifying the scope is better imo :)

WikiRik (changes requested):
Can you update https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md as well?

sdepold:
Good stuff. Also good remark from
Rik

sdepold:
Will the same logic be applied to commit messages?

ephys:
I would actually propose to disable commitlint on commit messages. I think the final squashed commit is the only interesting one. Intermediary commits can be trash as long as the final one is clean imo

sdepold:
Makes sense!

fzn0x:
> I would actually propose to disable commitlint on commit messages. I think the final squashed commit is the only interesting one. Intermediary commits can be trash as long as the final one is clean imo

Yeah, I agree, but how about we modify the squashed commit details too?

Because it will contain the list of bad commit messages.

fzn0x:
Capture

This is the details when we put null on squashed commit description, I think we must set rules to write anything related to avoid bad commit messages appearing on squashed commit details,

fzn0x:
Capture

This is the details when we put null on squashed commit description, I think we must set rules to write anything related to avoid non-standard commit messages appearing on squashed commit details,

ephys:
> Yeah, I agree, but how about we modify the squashed commit details too?

We can't enforce it (unfortunately there are no hooks to check the squashed commit message before merging), but our squash commit body should not include these intermediary messages. It should only be used to add more information about the changes, and breaking changes.

ephys:
pull_request_target does not seem to be the right target, I don't see the action running against the PR (even though it's listed here https://github.com/amannn/action-semantic-pull-request)

WikiRik:
In the CI we just do `on pull_request` or something like that right? That should work here as well, since with every extra commit it runs. But maybe we want to specify only conditions in which the name of the PR changes

WikiRik:
Or maybe this only runs after this PR is merged? Can you open a PR towards this branch to test that out?

WikiRik:
Btw; what do we want to use for when types are being improved?

ephys:
`fix(types)` or `feat(types)` depending on whether it's a bug fix or a new feature

WikiRik:
Woops, I missed the `types` scope. Nevermind my question, thanks for the quick response!

ephys:
It works! 

broken title type:
image

Correct type: 
image

WikiRik (approved):

WikiRik:
@ephys I got excited and merged, but we haven't fixed Renovate yet so we should do that soon. Can you try that out?

Issue #14260, opened 2022-03-18
Consider making `allowNull: false` the new default


ephys:
I don't think everyone is going to agree with me on this (feel free to express your opinion with the 👎 and 👍 reactions), but I think that nowadays we tend to be a lot stricter regarding what can and cannot be null.

My main reason for this change is that it would be better aligned with TypeScript. You'd only specify `allowNull` if you also made your type nullable:

```typescript
class User extends Model {
  firstName: string,
}

User.init({ firstName: string });

// &

class User extends Model {
  firstName: string | null,
}

User.init({ firstName: { type: string, allowNull: true } });
```

But I also think that in most cases, it's safer to make the application crash when null accidentally got in a field that was accidentally left nullable, than having incomplete data in the database.

To ease migration, I would also add an option in the Sequelize constructor to choose whether attributes are nullable by default .

---

TypeORM made this change a few years back, I think we could too.

Past issues related to this:

- https://github.com/sequelize/sequelize/issues/13439
- https://github.com/sequelize/sequelize/issues/11030
- https://github.com/sequelize/sequelize/issues/504

WikiRik:
Would be big breaking change, but I think this is for the best. Good idea to add an option in the constructor so if people want to keep the default nullable they can

javiermrz:
Hi! Is there any update on this?

ephys:
It's currently blocked by https://github.com/sequelize/sequelize/pull/14682, which is itself blocked by https://github.com/sequelize/sequelize/pull/14715, which is itself blocked by https://github.com/sequelize/sequelize/pull/14687, which is itself blocked by an issue with `describeTable` we need to resolve

We'll get there eventually :/

benasher44:
Hyped for this!

fpirsch:
Columns are nullable by default in databases. Making the opposite choice in the orm would be weird.
Why align with TypeScript ? Remember GraphQL fields are also nullable by default.

PR #14682, opened 2022-06-25
feat: return null if no defaultValue is specified

### LINKED ISSUE #14671: `Model.create()` in MySQL does not return `null` value when `defaultValue` hasn't defined
## Issue Creation Checklist

## Bug Description

In MySQL, if not provide explicit `defaultValue: null` to the model definition, the created result will not contain the column key.

But in PG it work right (with the `null` value in result);

### Reproducible Example

Here is the link to the SSCCE for this issue:

~~~js
const db = new Sequelize('mysql://root@localhost/test');

const Book = db.define('Books', {
  title: {
    type: Sequelize.STRING,
    defaultValue: null
  },
  price: {
    type: Sequelize.INTEGER
  }
});

await db.sync({ force: true });

const created = await Book.create();
// in pg here the result will contain `price` key as `null` value
console.log(created.get()); // without `price` key

const found = await Book.findOne();
console.log(found.get()); // with `price` key as `null` value
~~~

~~~javascript
{
  id: 1,
  title: null,
  price: null,
  createdAt: 2022-06-22T12:43:13.000Z,
  updatedAt: 2022-06-22T12:43:13.000Z
}
~~~

~~~javascript
{
  title: null,
  id: 1,
  updatedAt: 2022-06-22T12:43:13.333Z,
  createdAt: 2022-06-22T12:43:13.333Z
}
~~~

### Environment

- Sequelize version: 6.11.0
- Node.js version: v16.15.0
- If TypeScript related: TypeScript version:
- Database & Version: MySQL 8.0.28 & PostgresQL 14.3
- Connector library & Version: mysql2@2.3.3 & pg@8.7.3

- [x] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in resolving my issue.

---

WikiRik:
### Pull Request Checklist

### Description Of Change
This PR adds a defaultValue of `null` if there is no defaultValue specified when creating a model and the attribute allows `null`.
This fixes #14671.

@ephys I'm not so sure about this change. As you can see the unit tests fail because it also adds `DEFAULT NULL` to the SQL query. Is that wanted behaviour? This might even qualify as a breaking change

### Todos

- [ ] See what to do about the failing tests

### DIFF
--- a/src/sequelize.js
+++ b/src/sequelize.js
@@ -1271,6 +1271,10 @@ Use Sequelize#query if you wish to use replacements.`);
       }
     }
 
+    if (attribute.allowNull !== false && !attribute.defaultValue) {
+      attribute.defaultValue = null;
+    }
+
     return attribute;
   }
 }
--- a/test/integration/model/create.test.js
+++ b/test/integration/model/create.test.js
@@ -1395,6 +1395,12 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       expect(newTasks[3].title).to.equal('SAVE TRANSACTION');
     });
 
+    it('should return null value when defaultValue is not defined', async function () {
+      const user = await this.User.create();
+
+      expect(user.username).to.equal(null);
+    });
+
     describe('enums', () => {
       it('correctly restores enum values', async function () {
         const Item = this.sequelize.define('Item', {


src/sequelize.js
```diff
@@ -1271,6 +1271,10 @@ Use Sequelize#query if you wish to use replacements.`);
       }
     }
 
+    if (attribute.allowNull !== false && !attribute.defaultValue) {
```
ephys, line 1274: `defaultValue` could be 0, or `''`, a better check would be:

```suggestion
    if (attribute.allowNull !== false && attribute.defaultValue === undefined) {
```

It's also a good opportunity to normalize `attribute.allowNull` just before this so the test can be reduced to

```suggestion
    if (attribute.allowNull && attribute.defaultValue === undefined) {
```

ephys:
> As you can see the unit tests fail because it also adds DEFAULT NULL to the SQL query. Is that wanted behaviour? This might even qualify as a breaking change

The only element I was worried about was https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html, but after checking `DEFAULT NULL` is the default behavior for nullable columns there too

WikiRik:
@ephys looking at the failed integration tests, I think it would be better to do this change after https://github.com/sequelize/sequelize/issues/14260 so that the impact of this on the tests is smaller. Do you agree?

ephys:
Do we want to do #14260 in v7? I was thinking of postponing it to v8 to reduce how many impactful breaking changes we have, and maybe add the flag to change the default nullability in v7?

ephys:
I think this PR can be undrafted once https://github.com/sequelize/sequelize/pull/14715 is complete, that other PR should implement the missing pieces

WikiRik:
We still have an open issue connected to this, I'll close this for now and we can revisit with that issue later

Issue #9058, opened 2018-02-15
Alter queries generated by sequelize.sync with alter: true report syntax problems when columns use an explicit defaultValue


jsanta:
## What are you doing?
Syncing using alter: true for a table model using  a column with an explicit definiton for the defaultValue.

```js
// Model
const User = sequelize.define('User',
    {
        id: {
            type: Sequelize.INTEGER,
            autoIncrement: true,
            primaryKey: true,
            allowNull: false
        },
        email: {
            type: Sequelize.STRING,
            allowNull: false,
            unique: 'usr_email',
            set(val) {
                const email = (!val) ? this.getDataValue('username') : val;
                this.setDataValue('email', email.toLowerCase());
            },
            validate: {
                isEmail: true
            }
        },
        isValidated: {
            type: Sequelize.BOOLEAN,
            defaultValue: false
        }
    },
    {
        timestamps: true,
        freezeTableName:true,
        paranoid: true
    }
);

exports.User = User;
```

```
// DB Sync
var sequelize = db.sequelize;
    sequelize.sync({
        force: false,
        logging: dbLog,
        alter: true
    }).then(function () { /* ... */ });
```

## What do you expect to happen?
Alter queries executed without any syntax error

## What is actually happening?
Alter queries built with an invalid syntax.

_Output messages_
```
{
  SequelizeDatabaseError: Incorrect syntax near the keyword 'DEFAULT'. [...] ,
  sql: 'ALTER TABLE [User] ALTER COLUMN [isValidated] BIT DEFAULT 0;'
}
```

## Workaround
Modify the model definition to use a function that returns the default value the column instead of specifying the value itself:
```js
// Model
const User = sequelize.define('User',
    {
        id: {
            type: Sequelize.INTEGER,
            autoIncrement: true,
            primaryKey: true,
            allowNull: false
        },
        email: {
            type: Sequelize.STRING,
            allowNull: false,
            unique: 'usr_email',
            set(val) {
                const email = (!val) ? this.getDataValue('username') : val;
                this.setDataValue('email', email.toLowerCase());
            },
            validate: {
                isEmail: true
            }
        },
        isValidated: {
            type: Sequelize.BOOLEAN,
            defaultValue: () => false
        }
    },
    {
        timestamps: true,
        freezeTableName:true,
        paranoid: true
    }
);

exports.User = User;
```

__Dialect:__ mssql
__Dialect version:__ 4.29.2
[...]

PR #14477, opened 2022-05-02
MSSQL add changeColumn function

lohart13:
### Pull Request Checklist

_Please make sure to review and check all of these items:_

- [ ] Have you added new tests to prevent regressions?
- [x] Does `yarn test` or `yarn test-DIALECT` pass with this change (including linting)?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?
- [ ] Did you update the typescript typings accordingly (if applicable)?
- [ ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [x] Did you follow the commit message conventions explained in [CONTRIBUTING.md](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)?

### Description Of Change

The changeColumn function in the query interface does not currently work when constraints are defined (either default values or unique columns) for MSSQL. By checking if the constraint exists for the column and recreating if required, this allow the changeColumn function to work with MSSQL.

The test for the changeColumn function has been excluded for MSSQL as the new function calls two other SQL queries prior to the final query and the stub picks up the first query causing the test to fail.

### Todos

- [ ] Update docs if required
- [ ] Add test if required

PR #14687, opened 2022-06-26
feat: add `queryInterface.changeColumns`, fix many bugs with `changeColumn`

ephys:
## Description Of Change

The initial goal of this PR was to investigate [a bug reported here](https://github.com/sequelize/sequelize/issues/9058#issuecomment-1165615540), it quickly turned out that `changeColumn` had a completely broken implementation and I opted for doing a complete rewrite and fixing many bugs in the process.

### `changeColumns`

I've added `changeColumns` which is able to change more than one column from the same table in one `ALTER TABLE` query, reducing the number of necessary queries.

`changeColumns` follows our new API design, meaning it's possible to use it with a Model, a table name, a table object (table+schema), and it will default to the schema defined in the sequelize options if none is specified.

That method also throws clear error messages when the user tries to do something out of its scope. Previously, it would just crash with an obscure message.

`changeColumn` just defers to `changeColumns`, and inherits from its features.

### All dialects

- `changeColumns` *only* changes what has been provided. No more dropping the default value because it was not provided in the call. Instead `dropDefaultValue` has been added as an option. Similarly, the nullability of the column is *only* changed if `allowNull` is explicitly set to true or false (undefined just does not change the definition at all).
- The previous implementation was broken because it used regexes to modify another query, which broke very easily by using keywords in the default value.
- This method supports enabling/disabling `autoIncrement` & `autoIncrementIdentity`

### Postgres

- `changeColumns` is able to replace an enum with another, drastically simplifying the modification of enums in migrations. Before, it was impossible to use `changeColumn` to change the definition of a column's enum.

## Commit message

```

BREAKING CHANGE: `changeColumn` does not change a property if it is not specified. For instance, previously it would default `allowNull` to true and add `DROP NOT NULL` if unspecified.
postgres: `changeColumn` can replace a column's enum, as long as only that column is using that enum (which is the default behavior for tables generated by Sequelize).
`changeColumn` does not break if a SQL keyword is used in the default message anymore.
```

## Todos

- [x] postgres
- [ ] db2
- [ ] ibmi
- [ ] mariadb
- [ ] mssql
- [ ] mysql
- [ ] snowflake
- [ ] sqlite


WikiRik:
Does this replace #14477?

ephys:
Yep! It does fix the issue I mentioned but still checks whether a partially built query, that can include user-provided values (through `defaultValue`) contains some keywords, meaning the query will break if the `defaultValue` option includes one of those keywords

I'll merge the necessary changes from that PR into this one

ephys:
So I don't forget: This PR is blocked, we need to open another PR to clean up `queryInterface.describeTable` to ensure it returns consistent results accros all dialects

WikiRik:
Just to check; this is still blocked by `describeTable`, correct?

Any chance parts can be done already? Like migrating the enum functions to TS.

ephys:
Yep! I already introduced the internal query generator files & other similar refactors in a recent PR, the `canHaveDefaultValue` logic could be moved too, as well as the changes to `withUsageContext`, `getForeignKeyReferencesForTable`. The change to `Nullish` can also be done. Maybe the enum functions but I'm not sure. At least some of them can. Accepting `DbObjectId` in `Identifier` and other places would be great too, at `sql.identifier` could then directly specify the schema. Not using `.field` or `.fieldName` in our source code to remove deprecation warnings would be great too.

I added `MapView#getOrThrow` in #17077

Those would already be pretty great. As for the rest, it's indeed mostly blocked by `describeTable`

ephys:
(ignore notification, was demoing something)

lohart13:
Just wondering what the status of this PR is?

I notice the last comments mention the issues with `describeTable` but was wondering if some of the changes such as the datatype ones (especially the ENUM's for Postgres) extracted into a separate PR?

I'm working on a `addColumn` rewrite some of the changes in the PR would greatly help.

WikiRik:
> Just wondering what the status of this PR is?

It's on hold. I think this will land somewhere later in v7 as an experimental feature and becomes the default in v8.

> I notice the last comments mention the issues with `describeTable` but was wondering if some of the changes such as the datatype ones (especially the ENUM's for Postgres) extracted into a separate PR?

As far as I know, those have not been extracted yet but feel free to do so.

PR #14689, opened 2022-06-26
fix(mssql): insert values row limit is 1000

lohart13:
### Pull Request Checklist

### Description Of Change

The bulkInsertQuery sets the number of rows per insert query to 1000 with is the MSSQL limit as per [Table Value Constructor (Transact-SQL)](https://docs.microsoft.com/en-us/sql/t-sql/queries/table-value-constructor-transact-sql?view=sql-server-ver16)


WikiRik:
For the people reviewing this:

The following test was added with the original PR that introduced this behaviour;
```js
it('handles decent sized bulk creates', async function () {
  const User = this.sequelize.define('User', { username: DataTypes.STRING, num: DataTypes.INTEGER, status: DataTypes.STRING });
  const Task = this.sequelize.define('Task', { title: DataTypes.STRING });

  Task.hasMany(User);

  const users0 = _.range(1000).map(i => ({ username: `user${i}`, num: i, status: 'live' }));
  await User.bulkCreate(users0);
  await Task.create({ title: 'task' });
  const users = await User.findAll();
  expect(users).to.have.length(1000);
});
```

fzn0x (dismissed):

ephys (changes requested):
Thank you for the PR

I don't understand the goal of the PR.
Could you elaborate on what the issue was. Could you add an integration test that would fail without your change but succeeds now to prevent future regressions?

lohart13:
Hello @ephys,

The issue at the moment is the below code has a max rows per insert query of 126 and this decreases as the number of columns/attributes increases which in turn increases the number of insert queries to bulk insert data which increases the time take to complete a bulk insert operation.

According to the [Microsoft documentation](https://docs.microsoft.com/en-us/sql/t-sql/queries/table-value-constructor-transact-sql?view=sql-server-ver16), the max number of rows per insert query is 1000, therefore by maximising the number of rows per query, it speeds up the time take to complete a bulk insert operation.

Tests added now.
Due to the size of the results, I added regular expression checking to keep the test to a minimum. 
```javascript
const commands = [];
    let offset = 0;
    const batch = Math.floor(250 / (allAttributes.length + 1)) + 1;
    while (offset < Math.max(tuples.length, 1)) {
      const tupleStr = tuples.slice(offset, Math.min(tuples.length, offset + batch));
      let generatedQuery = allQueries.map(v => (typeof v === 'string' ? v : v(tupleStr))).join(';');
      if (needIdentityInsertWrapper) {
        generatedQuery = `SET IDENTITY_INSERT ${quotedTable} ON; ${generatedQuery}; SET IDENTITY_INSERT ${quotedTable} OFF;`;
      }

      commands.push(generatedQuery);
      offset += batch;
    }
```

src/dialects/mssql/query-generator.js
```diff
@@ -397,16 +397,15 @@ export class MsSqlQueryGenerator extends AbstractQueryGenerator {
 
     const commands = [];
     let offset = 0;
-    const batch = Math.floor(250 / (allAttributes.length + 1)) + 1;
     while (offset < Math.max(tuples.length, 1)) {
-      const tupleStr = tuples.slice(offset, Math.min(tuples.length, offset + batch));
+      const tupleStr = tuples.slice(offset, Math.min(tuples.length, offset + 1000));
```
ephys, line 401: Can you add the following comment here?

```suggestion
      // SQL Server can insert a maximum of 1000 rows at a time,
      // This splits the insert in multiple statements to respect that limit
      const tupleStr = tuples.slice(offset, Math.min(tuples.length, offset + 1000));
```

test/unit/query-interface/bulk-insert.test.ts
```diff
@@ -12,6 +13,22 @@ describe('QueryInterface#bulkInsert', () => {
     sinon.restore();
   });
 
+  it('uses minimal insert queries', async () => {
```
ephys, line 16: The test I had in mind was more an integration test where you try to bulkInsert 1001 rows, then count the results, to assert that it does work :)

test/support.ts
```diff
@@ -290,19 +290,22 @@ type PartialRecord<K extends keyof any, V> = Partial<Record<K, V>>;
 
 export function expectsql(
   query: { query: string, bind: unknown } | Error,
-  assertions: { query: PartialRecord<ExpectationKey, string | Error>, bind: PartialRecord<ExpectationKey, unknown> },
+  assertions: {
```
ephys, line 293: I'm very sorry, I made changes to our test tools that are going to conflict with this

I don't think the unit test is necessary if you add the integration test, but if you want to do the regex test anyway, I would propose to revert the changes you made to expectsql and instead use the new `expectPerDialect` function & create a new `toMatchRegex` assertion:

```ts
expectPerDialect(firstCall.args[0] as string, {
  default: toMatchRegex(/INSERT INTO (?:`|")Users(?:`|") \((?:`|")firstName(?:`|")\) VALUES (?:\('\w+'\),){999}\('\w+'\);/),
  ibmi: toMatchRegex(/SELECT \* FROM FINAL TABLE \(INSERT INTO "Users" \("firstName"\) VALUES (?:\('\w+'\),){999}\('\w+'\)\)/),
  mssql: toMatchRegex(/INSERT INTO \[Users\] \(\[firstName\]\) VALUES (?:\(N'\w+'\),){999}\(N'\w+'\);/),
})
```

`expectPerDialect` is a more versatile version of `expectsql`

ephys (changes requested):
Thank you for the explanation :)
This looks good to me then, just comments about tests

test/unit/query-interface/bulk-insert.test.ts
lohart13: Hello @ephys,

As @WikiRik mentioned above, there is already an integration test that is testing the bulkInsert query.
Does this test suffice?

The integration test to insert 1000 or 1001 rows should always pass as the previous code could complete the bulkInsert, the issue is that it uses multiple insert statements.

This is why I went with the unit test instead as this way I was able to ensure that only one insert statement is made.

test/support.ts
lohart13: Let me know if you are happy with the existing integration test otherwise I will update the regex test

WikiRik:
@ephys I re-requested your review to bring the comments back to your attention

lohart13:
@ephys wondering if you've had any further thoughts on this

WikiRik:
@ephys ping

ephys:
Thank you both, apologies for the very long delay

lohart13:
@ephys just wondering if you have any further comments on this PR

ephys:
I've uncovered a pretty important issue thanks to your PR: Since they're different queries, if one of them fails, the others are not rolled back

```ts
  // mssql can only insert 1k row at a time, so we split it in multiple queries
  it('rolls back changes after inserting 1000 rows', async function () {
      username: {
        type: DataTypes.STRING,
        allowNull: false,
      },
    });

    try {
      await User.bulkCreate([
        ...times(1000, () => ({ username: 'John' })),
        { username: null },
      ]);
    } catch {
      // ignore
    }

    const count = await User.count();
    expect(count).to.equal(0);
  });
```

It's not your PR's fault, it was already there, but I think we should throw an error if there are more than 1k rows and the query is not running in a transaction

Let me know if you're OK with doing it in this PR, otherwise I'll just open a bug report about it and we can merge this one

lohart13:
@ephys where can I check if the query is running under a transaction?
Does this need to be at the query interface or query generator?

ephys:
The best place to do it would be in MSSQL's queryInterface, but on second thought I'd love to see this merged and it's not caused by this PR, ok with you if we do that fix in a follow up PR? :)

ephys (approved):

WikiRik (approved):
@ephys can you make a bug report for that?

WikiRik:
@lohart13 thanks for working on all these mssql improvements!

PR #14691, opened 2022-06-27
docs: fix comment in data-types.d.ts

SeongJaeSong:
### Description Of Change

[The actual file(data-types.js)](https://github.com/sequelize/sequelize/blob/main/src/data-types.js#L585
) says JSON type available for MySQL, Postgres, and SQLite. But the `d.ts` file says JSON is only available in Postgres.

Is this too tiny problem to open PR?

### DIFF
--- a/src/data-types.d.ts
+++ b/src/data-types.d.ts
@@ -365,7 +365,7 @@ export interface DateOnlyDataType extends AbstractDataType {
 export const HSTORE: AbstractDataTypeConstructor;
 
 /**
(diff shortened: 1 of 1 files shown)


WikiRik (approved):

WikiRik:
No worries about the size of this PR. Thanks for making it!
@ephys do you mind if I merge this even though you are fixing it in #14505 ?

ephys:
Sure! Go ahead :)

PR #14699, opened 2022-06-29
fix: missing option typings and `raw` option in bulkCreate when using build method

### LINKED ISSUE #14639: `raw` option for bulkCreate()
## Issue Creation Checklist

### Describe the feature you'd like to see implemented

~~~javascript
const User = db.define('User', {
  name: {
    type: Sequelize.STRING,
    set(v) {
      this.setDataValue('name', v.toUpperCase());
    }
  }
});

await db.sync();

const [user] = await User.bulkCreate([
  { name: 'a' }
], { raw: true });

console.log(user.name); // got 'A' but not 'a'
~~~

Expect to get `"a"` as user.name when using `raw: true`.

### Describe why you would like this feature to be added to Sequelize

While `raw` is an option in `Model.create()` method, which could skip setters, but we cannot use it in `Model.bulkCreate()`. This is out of intuition, and very inconvenience to use in such scenario.

- [ ] Yes. This feature only applies to the following dialect(s): 

- [x] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in implementing my feature.

---

*Indicate your interest in the addition of this feature by adding the 👍 reaction. Comments such as "+1" will be removed.*

fzn0x:
### Pull Request Checklist

### Description Of Change

Closes #14639 I saw that we have missing options on bulkCreate `this.build` method when calling it. It's different when we see it in `create` method

### Todos

- [x] Add missing `raw` option in bulkCreate `this.build` method
- [x] Add types test for `raw` option

## Follow Up Tasks

- [ ] Add v6 backport


fzn0x:
For failed tests, i think it was flaky tests 👀 I'm unable to investigate it now.

WikiRik:
Yeah, that's our most flaky test. Don't worry about it

fzn0x:
> Yeah, that's our most flaky test. Don't worry about it

Thanks :)

test/unit/model/bulk-create.test.js
```diff
@@ -41,6 +47,16 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           { account_id: 42, id: null },
         ]);
       });
+
+      it('should not return uppercased A', async function () {
```
ephys, line 51: this test doesn't belong inside of the `validations` describe, can you place it outside of that block?

I also would name it more explicitly:

```suggestion
      it('skips setters if the raw option is true', async function () {
```

src/model.js
```diff
@@ -2571,7 +2571,13 @@ Specify a different name for either index to resolve this issue.`);
       }
     }
 
-    const instances = records.map(values => this.build(values, { isNewRecord: true, include: options.include }));
+    const instances = records.map(values => this.build(values, {
+      isNewRecord: true,
+      attributes: options.fields,
+      include: options.include,
+      raw: options.raw,
+      silent: options.silent,
```
ephys, line 2579: Can you add a test for `silent`?

src/model.js
```diff
@@ -2571,7 +2571,13 @@ Specify a different name for either index to resolve this issue.`);
       }
     }
 
-    const instances = records.map(values => this.build(values, { isNewRecord: true, include: options.include }));
+    const instances = records.map(values => this.build(values, {
+      isNewRecord: true,
+      attributes: options.fields,
```
ephys, line 2576: Can you add a test for `attributes`?

ephys (changes requested):
Thanks for the PR!

fzn0x:
I have no idea why stub keep failing on findAll :/

fzn0x:
Mark this PR as draft, all commits will be rebased soon.

src/model.js
fzn0x: Sorry, `silent` is not a valid missing option, I notice the test for silent is already passed in integration test.

src/model.js
fzn0x: Sorry, attributes is not a valid missing option, I notice the test for attributes is already passed in integration test. You can check the integration test for bulk-create.

fzn0x: Sorry, `attributes` is not a valid missing option, I notice the test for attributes is already passed in integration test. You can check the integration test for bulk-create.

ephys (dismissed):

ephys:
Do the typings support this option?

fzn0x:
> Is this option included in the typings of this method?

Will check it, thanks

test/unit/model/bulk-create.test.js
```diff
@@ -41,6 +47,16 @@ describe(Support.getTestDialectTeaser('Model'), () => {
           { account_id: 42, id: null },
         ]);
       });
+
+      it('skips setters if the raw option is true', async function () {
```
WikiRik, line 51: Should we add a unit test that checks that the name is changed to `A` when raw is not there/false?

fzn0x: Good catch, will add it

fzn0x:
> Feel free to broaden the scope of this PR and fix more typings here. If possible do limit it to typings of model.js in this PR

Thanks for the chance!

fzn0x:
> Feel free to broaden the scope of this PR and fix more typings here. If possible do limit it to typings of model.js in this PR

Hi @ephys @WikiRik @sdepold

Added the groupedLimit typing, I had few questions regarding this option because the option needs decision making.

1. Is it good to add a new TODO for the option to be part of private APIs in a separate PR and tag them as deprecated as public API in this PR?
2. Is it good to keep the option as public API?

fzn0x:
> Feel free to broaden the scope of this PR and fix more typings here. If possible do limit it to typings of model.js in this PR

Hi @ephys @WikiRik @sdepold

Added the groupedLimit typing, I had few questions regarding this option because the option needs decision making.

1. Is it good to add a new TODO for the option to be part of private APIs in a separate PR and tag them as deprecated soon as public API in this PR?
2. Is it good to keep the option as public API?

fzn0x:
```
Error: src/associations/has-many.ts(213,11): error TS2322: Type 'S[SourceKey][]' is not assignable to type 'string[]'.
```

No idea what is `S[SourceKey][]` 😄

fzn0x:
PR is ready for next review :) !

src/model.d.ts
```diff
@@ -881,8 +898,37 @@ export interface FindOptions<TAttributes = any>
...
+   *    SELECT * FROM "user" WHERE project_id = 2 LIMIT 3,
+   *    SELECT * FROM "user" WHERE project_id = 3 LIMIT 3
+   * );
+   * ```
+   *
+   * Since not every dialect supports lateral joins, this is currently implemented as a union.
+   *
+   * Currently groupedLimit is used by include.separate for hasMany includes (where we first load the main table, getting matching project ids, and then fetch a limited amount of users for each).
+   *
+   * Note: Using this option is not recommended. See https://github.com/sequelize/sequelize/issues/6899#issuecomment-262550213.
```
ephys, line 929: I feel like this should be the very first sentence in this JSDoc. This option is not considered stable and will likely go away in a future version

src/model.d.ts
```diff
@@ -881,8 +898,37 @@ export interface FindOptions<TAttributes = any>
...
+   * ```sql
+   * SELECT * FROM (
+   *   UNION ALL
+   *    SELECT * FROM "user" WHERE project_id = 1 LIMIT 3,
+   *    SELECT * FROM "user" WHERE project_id = 2 LIMIT 3,
+   *    SELECT * FROM "user" WHERE project_id = 3 LIMIT 3
+   * );
+   * ```
+   *
+   * Since not every dialect supports lateral joins, this is currently implemented as a union.
```
ephys, line 925: That's not the reason for the existence of `groupedLimit`, it's used internally when an association must be loaded in a separate query (`separate` option), you wouldn't do that with a lateral join (you'd use a separate join if `separate` were false)

```suggestion
```

ephys (changes requested):

test/unit/model/bulk-create.test.js
```diff
@@ -41,6 +47,26 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+        await this.Model.bulkCreate([
+          { accountId: 42, name: 'a' },
+        ], { raw: true });
+
+        expect(this.stub.getCall(0).args[1]).to.deep.equal([
+          { account_id: 42, id: null, name: 'a' },
+        ]);
+      });
+
+      it('should not skips setters if the raw option is false', async function () {
```
ephys, line 61:
```suggestion
      it('does not skip setters if the raw option is false', async function () {
```

WikiRik:
Putting this to draft until the following two comments have been resolved;
- Moving the note for not recommending groupedLimit to the top; https://github.com/sequelize/sequelize/pull/14699/files#r995583279
- Migrating test/unit/model/bulk-create to TypeScript; https://github.com/sequelize/sequelize/pull/14699/files#r995573078

PR #14719, opened 2022-07-05
feat: export hooks for types(v6)

Beace:
### Description Of Change

We solved the problem of typescript `nodenext` in the [previous PR](https://github.com/sequelize/sequelize/pull/14620).But I found that `types/index.d.ts` does not export `hooks`, which will make references such as `import { SequelizeHooks } from 'sequelize/types/hooks';` inaccessible in `nodenext`. Here is a example in `sequelize-typescript`

```ts
import { SequelizeHooks } from 'sequelize/types/hooks';
export interface HookMeta {
    hookType: keyof SequelizeHooks;
    methodName: string;
    options?: HookOptions;
}
```

image

### Todos

- [x] add `export * from './hooks'` in `index.d.ts`
- [ ] upgrade `sequelize-typescript`, change the import  `import { SequelizeHooks } from 'sequelize';`

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -14,6 +14,7 @@ export * from './model';
 export * from './dialects/abstract/query-interface';
 export * from './sequelize';
 export * from './transaction';
+export * from './hooks';
 export { useInflection } from './utils';
 export { Validator } from './utils/validator-extras';
 export { Utils, QueryTypes, Op, TableHints, IndexHints, DataTypes, Deferrable };


Beace:
Could someone help me look at it? After that, I have to ask PR for serialize-typescript.

Beace:
https://github.com/sequelize/sequelize-typescript/blob/master/src/hooks/shared/hook-meta.ts

ephys:
That's normal, `/types` is not intended to be used directly

In v6, you can import from `/lib` instead
In v7, you can import from `/_non-semver-use-at-your-own-risk_` instead

If sequelize-typescript is trying to import from `/types`, using one of the above should fix the issue

I don't want to expose `/lib/hooks` in the stable api endpoint. It's a bunch of internal code not designed for public use and exposing it means any changes made to it is going to be a breaking change

Beace:
Thanks for your explanation. I understand what you mean, I'm going to give pr to sequelize-typescript.

Beace:
@ephys  `sequelize/types/hooks` -> `sequelize/lib/hooks`. this will be a breaking update to `sequelize-typescript`. The module `commonjs` cannot found `import { SequelizeHooks } from 'sequelize/lib/hooks';`  I understand you don't want to expose /lib/hooks in the stable api endpoin, but what the difference between `import { SequelizeHooks } from 'sequelize/lib/hooks'` and `import { SequelizeHooks } from 'sequelize'`

ephys:
I'm closing this PR because https://github.com/sequelize/sequelize/pull/15123 should have fixed this. Let me know if the issue still exists though

PR #14731, opened 2022-07-08
fix(postgres): prevent crash if postgres connection fails immediately (v6)

schmod:
### Pull Request Checklist

### Description Of Change

We recently observed a crash due to an unhandled `Connection terminated unexpectedly` exception originating from the `pg` library.  As best as I've been able to tell, this _shouldn't_ be possible, and I've had difficulty reproducing it.  

However, upon reading through the current implementation of the Postgres connection-manager, I noticed that we don't immediately attach an error-handler to the `pg` connection.  

Based off of my reading of this code, it actually seems reasonably possible that an error emitted during either of the setup queries could trigger a hard crash.  Moving the error-handler to an earlier phase of the connection's lifecycle should, in theory, make this a bit safer, and guard against any errors that are emitted very early in the connection's lifecycle.

Given the difficulty associated with reproducing this issue, I haven't been able to write an automated test-case to reproduce the issue or validate the fix.  However, I'm reasonably confident that this fix is safe, and unlikely to make anything _worse_. 

### Todos

- [ ] Add an automated test (if possible)

### DIFF
--- a/src/dialects/postgres/connection-manager.js
+++ b/src/dialects/postgres/connection-manager.js
@@ -200,6 +200,13 @@ class ConnectionManager extends AbstractConnectionManager {
       });
     });
 
+    // Don't let a Postgres restart (or error) to take down the whole app
+    connection.once('error', error => {
+      connection._invalid = true;
+      debug(`connection error ${error.code || error.message}`);
+      this.pool.destroy(connection);
+    });
+
     let query = '';
 
     if (this.sequelize.options.standardConformingStrings !== false && connection['standard_conforming_strings'] !== 'on') {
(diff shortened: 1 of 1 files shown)


src/dialects/postgres/connection-manager.js
```diff
@@ -200,6 +200,13 @@ class ConnectionManager extends AbstractConnectionManager {
       });
     });
 
+    // Don't let a Postgres restart (or error) to take down the whole app
```
schmod, line 203: I also considered adding moving this even _further_ up in the lifecycle (prior to `.connect()`), but I was uncertain about whether or not it could potentially interfere with our _other_ error-handling code above.

(Additionally, I don't _think_ there's actually any room for an error to slip in between the establishment of the connection and time that this handler is attached)

ephys (approved):
This looks like a good change. Thank you for the PR!
I don't see how we would go about writing a test to prevent a regression either unfortunately

I see this is targeting v6. In order to avoid losing this PR in v7, could you also make a PR that targets `main`?

schmod:
v7/main PR: https://github.com/sequelize/sequelize/pull/14731

WikiRik:
Thanks for the two PRs!

PR #14737, opened 2022-07-10
meta: update all databases & use oldest/latest naming scheme

ephys:
## Description of changes

This PR does the following:

- Update all images in `/dev` to match the ones we support in v7
- Use the 'oldest/latest' naming scheme for images in `/dev`
- Replace the `sushantdhiman/postgres` docker image with `postgis/postgis`, so we don't have to maintain it ourselves
- The postgres-latest image has been updated to postgres 12.
  - Waiting on https://github.com/brianc/node-postgres/issues/1927 to be resolved before moving to postgres 14
- Add `yarn test-integration-all`, `start-oldest`, `start-latest`, and `stop-all` scripts
  - Useful to `.only` a single test suite and run it on all dialects
- Merged `stop-x-latest` and `stop-x-oldest` because both commands are fast enough to reduce how many commands we provide
- Made the CI use `/dev` for postgres to reduce duplication. Let's see if this works
- ~~Removed `yarn test-unit-postgres-native` as unit tests don't interact with pg-native so shouldn't behave differently from `test-unit-postgres` (I may be wrong, I can revert this)~~


WikiRik:
Thanks for this! I'm not so sure about the pg-native unit tests either. Might be better to revert that change for now until we've taken a better look at it

dev/mariadb/latest/start.sh
```diff
@@ -3,14 +3,12 @@ set -Eeuxo pipefail # https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo
...
-./../../wait-until-healthy.sh sequelize-mariadb-103
+./../../wait-until-healthy.sh sequelize-mariadb-latest
 
-docker exec sequelize-mariadb-103 \
+docker exec sequelize-mariadb-latest \
   mysql --host 127.0.0.1 --port 3306 -uroot -psequelize_test -e "GRANT ALL ON *.* TO 'sequelize_test'@'%' with grant option; FLUSH PRIVILEGES;"
 
 DIALECT=mariadb ts-node ../../check-connection.ts
-
-echo "Local MariaDB-10.3 instance is ready for Sequelize tests."
```
ephys, line 16: I removed this message because a similar one has been added to check-connection instead

ephys:
I fail to see what I changed in db2 that would cause this issue. The only change there was supposed to be a folder rename. Very strange. Do you see something weird @WikiRik?

ephys:
I downgraded postgres-latest to v12 until https://github.com/brianc/node-postgres/issues/1927 is fixed

.github/workflows/ci.yml
```diff
@@ -191,28 +191,16 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [14, 18]
-        postgres-version: [9.5, 10]
+        postgres-version: [oldest, latest]
```
WikiRik, line 194: Should we do this for the other dialects as well? It's a bit odd having this only for postgres

dev/mysql/oldest/docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
-  mysql-57:
-    container_name: sequelize-mysql-57
+  mysql-oldest:
+    container_name: sequelize-mysql-oldest
     image: mysql:5.7
```
WikiRik, line 4: Should we specify the patch version here as well? Like we did with mysql-latest

package.json
```diff
@@ -206,27 +207,24 @@
...
-    "stop-db2": "bash dev/db2/11.5/stop.sh",
-    "restart-mariadb": "yarn start-mariadb",
-    "restart-mysql": "yarn start-mysql",
-    "restart-postgres": "yarn start-postgres",
-    "restart-mssql-oldest": "yarn start-mssql-oldest",
-    "restart-mssql-latest": "yarn start-mssql-latest",
-    "restart-db2": "yarn start-db2",
-    "----------------------------------------- local tests ---------------------------------------------": "",
+    "start-db2-oldest": "bash dev/db2/oldest/start.sh",
+    "start-oldest": "concurrently \"npm:start-*(!db2)-oldest\" && npm run start-db2-oldest",
```
WikiRik, line 219:
```suggestion
    "start-oldest": "concurrently \"npm:start-*-oldest\"",
```
Any reason not to do this?

package.json
```diff
@@ -206,27 +207,24 @@
...
+    "start-latest": "concurrently \"npm:start-*(!db2)-latest\"",
+    "stop-mariadb": "bash dev/mariadb/oldest/stop.sh; bash dev/mariadb/latest/stop.sh",
+    "stop-mysql": "bash dev/mysql/oldest/stop.sh; bash dev/mysql/latest/stop.sh",
+    "stop-postgres": "bash dev/postgres/oldest/stop.sh; bash dev/postgres/latest/stop.sh",
+    "stop-mssql": "bash dev/mssql/oldest/stop.sh; bash dev/mssql/latest/stop.sh",
+    "stop-db2": "bash dev/db2/oldest/stop.sh",
+    "stop-all": "concurrently \"npm:stop-*(!all|db2)\" && npm run stop-db2",
```
WikiRik, line 226:
```suggestion
    "stop-all": "concurrently \"npm:stop-*(!all)\"",
```
Just like my previous suggestion

package.json
```diff
@@ -246,6 +245,8 @@
     "test-integration-db2": "cross-env DIALECT=db2 yarn test-integration",
     "test-integration-ibmi": "cross-env DIALECT=ibmi yarn test-integration",
     "test-integration-snowflake": "cross-env DIALECT=snowflake yarn test-integration",
+    "test-integration-all": "yarn test-integration-mariadb && yarn test-integration-mysql && yarn test-integration-postgres && yarn test-integration-postgres-native && yarn test-integration-sqlite && yarn test-integration-mssql && yarn test-integration-db2 && yarn test-integration-ibmi && yarn test-integration-snowflake",
+    "----------------------------------------- all tests ---------------------------------------------": "",
```
WikiRik, line 249: Do we want to add a `test-all-oldest` and `test-all-latest`?

package.json
```diff
@@ -246,6 +245,8 @@
     "test-integration-db2": "cross-env DIALECT=db2 yarn test-integration",
     "test-integration-ibmi": "cross-env DIALECT=ibmi yarn test-integration",
     "test-integration-snowflake": "cross-env DIALECT=snowflake yarn test-integration",
+    "test-integration-all": "yarn test-integration-mariadb && yarn test-integration-mysql && yarn test-integration-postgres && yarn test-integration-postgres-native && yarn test-integration-sqlite && yarn test-integration-mssql && yarn test-integration-db2 && yarn test-integration-ibmi && yarn test-integration-snowflake",
```
WikiRik, line 248: Don't we want to make this `test-integration-oldest` and `test-integration-latest`?

test/integration/dialects/postgres/dao.test.js
```diff
@@ -965,7 +965,7 @@ if (dialect.startsWith('postgres')) {
     it('should save geometry correctly', async function () {
       const point = { type: 'Point', coordinates: [39.807_222, -76.984_722] };
       const newUser = await this.User.create({ username: 'user', email: ['foo@bar.com'], location: point });
-      expect(newUser.location).to.deep.eql(point);
+      expect(newUser.location).to.deep.include(point);
```
WikiRik, line 968: Why this change?

package.json
ephys: DB2 requires root (with the current implementation at least), starting/stopping it in parallelp leads to the password prompt being lost in the noise

package.json
ephys: The databases use the same port, so the only difference is in which database version was started before running this command

test/integration/dialects/postgres/dao.test.js
ephys: Postgis was updated to V3 which includes more metadata, but V2 is still supported, which does not include it

dev/mysql/oldest/docker-compose.yml
ephys: I hesitated one which patch version to use:

- latest patch version v5.7.38
- or v5.7.0

WikiRik: Let's try out 5.7.0

WikiRik:
About the pg-native unit tests; I took another look at them and I think we can safely remove those after all.

.github/workflows/ci.yml
ephys: I started implementing these changes but I'd like to move it to its own PR that I can just discard if it doesn't work out

dev/mysql/oldest/docker-compose.yml
ephys: Done

ephys:
> we can safely remove [native postgres unit tests] after all.

Done

package.json
ephys: Same reason as above, because oldest & latest run on the same port, they can't run at the same time, so `test-integration-oldest` & `test-integration-latest` would do the same thing as `test-integration-all`

WikiRik:
What remains for this PR? And what are the next steps after merging this?

ephys:
For this specific PR, I think everything is done.

For future PRs:
- Update postgres to v14 (blocked by https://github.com/brianc/node-postgres/issues/1927)
- Make Github Actions use the `/dev` folder instead of duplicating the database configs

WikiRik (dismissed):

ephys:
@WikiRik conflict resolution dismissed your review :(

WikiRik (approved):

Issue #14562, opened 2022-05-28
Meta: `before` events are overwritten in integration tests 


mike-usa:
## Issue Description

some mocha tests _before_ events are overwritten by support _beforeEach_ events

### Additional context

test/integration/support.js includes:
```javascript
beforeEach(async function () {
  await Support.clearDatabase(this.sequelize);
});
```
Mocha follows the hierarchy:
- before
- beforeEach
- afterEach
- after
This applies even if a file is deeply nested. For example; a `before` block in a deep file, like _test/integration/dialect/**/*.test.js_, will run before the `support.ts::beforeEach`.

In such a nested test, I'm adding a user and schemas in a `before` block as it makes sense to do it once for efficiency reasons, but before it runs the test, the Support's `beforeEach` will execute to clear the database, including the newly generated schemas.

### Is this issue dialect-specific?

- [x] No. This issue is relevant to Sequelize as a whole.
- [x] Yes. This issue only applies to the following dialect(s): XXX, YYY, ZZZ
- [ ] I don't know.

My specific issue regarding custom schemas is relevant only to databases that support schemas; however, the `before` vs `beforeEach` overwriting applies to every database type.  Also note that the `clearDatabase` function only drops tables, schemas, and enums (postgres-only); but other created objects and entities (users, functions & stored procedures, fdws, etc) would still live on, so it's important to note this also isn't complete database cleansing.

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [x] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

I think a change of this significance may take some serious refactoring.  More importantly, the dev team needs to be made aware of such an architectural change. Dependencies and behavior some might rely on in the past may no longer be available and that needs to be communicated and understood to avoid potential confusion.

### Additional Considerations

- **containers**: if the idea is to isolate tests, one consideration is to containerize test and just clone/startup at each _describe_ step; perhaps with some provisioning to speed things up
- **environmental awareness**: isolation may make more sense on the server to ensure 
- **configuration**: perhaps setting a boolean value in the a test's describe block like `skipClearDatabase` to bypass the step might be helpfup; of course, the test would need to remember to reset it after and if they're doing that, they might as well be able to explicitly calls `clearDatabase` when they need to (instead of relying on Support) to do it.

[...]

PR #14751, opened 2022-07-13
meta: only clear database once per file

WikiRik:
### Pull Request Checklist

### Description Of Change

This PR improves the performance of our test suite. Initially it moves the global `beforeEach` to every test suite (as per [a comment](https://github.com/sequelize/sequelize/pull/14345#issuecomment-1133188810) of @ephys ) and replaced by a `before` so it only runs once per file.

The smoke test file seems to be outdated so is not part of this PR. The snowflake specific tests are also out of scope for further improvement since we cannot test on those.

Relates to #14562

### Todos

- [ ] Further limit the number of times the database will be cleared by doing it once and then only when necessary

### DIFF
--- a/test/integration/associations/alias.test.js
+++ b/test/integration/associations/alias.test.js
@@ -6,7 +6,7 @@ const expect = chai.expect;
 const Support = require('../support');
 
 describe(Support.getTestDialectTeaser('Alias'), () => {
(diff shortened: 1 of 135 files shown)


WikiRik:
Closing this in favor of #14754

Issue #14749, opened 2022-07-12
Change Column and it is always needed to inform the type


luizcalaca:
## Issue Creation Checklist

[X] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

### What was unclear/insufficient/not covered in the documentation

When we have to changeColumn, in my situation I needed to change the column to add a default value, but in [documentation](https://sequelize.org/docs/v6/other-topics/query-interface/) in the Query Interface -> _changeColumn Section_ is not describing about if is necessary to inform the type with its Datatypes.

It is works
```
queryInterface.changeColumn('Person', 'foo', {
  type: DataTypes.FLOAT,
  defaultValue: 3.14
});
```

It is not works and I got an error, because you need to inform the type with its DataTypes.
```
queryInterface.changeColumn('Person', 'foo', {
  defaultValue: 3.14
});
```
So the documentation should be inform about that.

### If possible: Provide some suggestion on how we can enhance the docs

Adding the obligatory parameters in the text. Maybe it's implicit because if I need to change de default value and I also need to inform the type of that. But would be good if had expressly writed on documentation. 

### Additional context

The steps I did here was: 
1 - `npx sequelize migration:create --name adddefault`
2 -   
```
async up (queryInterface, Sequelize) {
    queryInterface.changeColumn('users', 'email', {
      defaultValue: 'test@test.com'
    });
  },
```
3 - `npx sequelize db:migrate` and after I got an error below:
![Screen Shot 2022-07-12 at 09 30 25](https://user-images.githubusercontent.com/2284795/178490031-46d85270-5dab-461d-8610-2b27eae31094.png)
4 - If I add the type it works:
```
  async up (queryInterface, Sequelize) {
    queryInterface.changeColumn('users', 'email', {
      type: Sequelize.STRING,
      defaultValue: 'test@test.com'
    });
```

ephys:
I consider this to be a bug, we're working on a fix https://github.com/sequelize/sequelize/pull/14687

luizcalaca:
@ephys and @WikiRik let me know if I could help with that bug.

barraponto:
@ephys @luizcalaca I guess a good first step would be to add a test that tries to change a column without providing the type. There are no such tests in #14687 ... yet.

ephys:
[...]

PR #14781, opened 2022-07-21
changeColumn should not depend on column type

barraponto:
Following #14749, this PR adds a small change to #14687 to ensure `.changeColumn` does not need column type information if the column change is not changing the type itself.


ephys:
The `QueryGenerator` (responsible for generating the SQL), which the test you changed tests, still needs the type in some dialects (at least mysql/mariadb), because the query must specify it.

The `QueryInterface` (responsible for executing specific queries) doesn't need it, and has tests for it in the PR: https://github.com/sequelize/sequelize/blob/ephys/rewrite-changeColumn/test/integration/query-interface/change-column.test.ts#L83

PR #14784, opened 2022-07-22
fix(mssql): match aggregate errors

WikiRik:
### Pull Request Checklist

### Description Of Change

Adds support for aggregate errors as introduced in tedious 14.5.0
Unblocks #14484


sdepold (approved):
Should we have a test for this? Or will the fact that the referenced pr no longer fail, suffice?

WikiRik:
I was thinking about that as well but I couldn't come up with a proper new test. The referenced PR should be fine. Earlier I was also thinking if this code should only be used with the new version of tedious but I decided not to since no behaviour should be changed in the older versions of tedious

WikiRik:
@ephys check this out for the multiple tedious errors. Maybe we should not return the errors directly but first collect them all

ephys:
Definitely should :)

WikiRik:
Can you add that to #14505 or #14999 ? Or a third separate PR?

ephys:
Separate PR, it's not related to either :)

WikiRik:
I know, it's just that you're working on those PRs and might have encountered some issues because of this

Issue #13646, opened 2021-11-10
MariaDB JSON DataType - included / Joined fields are not parsed (incl. fix #13633)


LukasKubec:
## Bug Description
On Find queries with MariaDB, Sequelize currently have issue with parsing JSON datatype columns. That is fixed by https://github.com/sequelize/sequelize/pull/13633 which works just fine, but only for columns on the top level model. If any other model is joined via include:, then JSON columns in included model are not parsed. I had chance to check it only with eager loading.

lib/dialects/mariadb/query.js in this function (line 177+), this.model does not include fields of joined models (they are available in this.options.include? if this.options.hasJoin is true) - this is version of fnction from fix #13633 

```js
  handleJsonSelectQuery(rows) {
    if (!this.model || !this.model.fieldRawAttributesMap) {
      return;
    }
    for (const _field of Object.keys(this.model.fieldRawAttributesMap)) {
      const modelField = this.model.fieldRawAttributesMap[_field];
      if (modelField.type instanceof DataTypes.JSON) {
        // Value is returned as String, not JSON
        rows = rows.map(row => {
          if (row[modelField.fieldName] && typeof row[modelField.fieldName] === 'string') {
            row[modelField.fieldName] = JSON.parse(row[modelField.fieldName]);
          }
          if (DataTypes.JSON.parse) {
            return DataTypes.JSON.parse(modelField, this.sequelize.options,
              row[modelField.fieldName]);
          }
          return row;
        });
      }
    }
  }
```

so if:

```js

const A = sequelize.define('A', /* ... */);
const B = sequelize.define('B', { 
jsonField: { DataType.JSON },
});

A.hasMany(B);

// A['B.jsonField'] will not be parsed to JSON
A.findAll({include: B})
```

### What do you expect to happen?

Fields of included models should be parsed to JSON as well in case of DataType.JSON

### What is actually happening?

String is returned - e.g. this from DB: 
```js
{"key": value}
```
is returned as:
```
"\{\"key\": value\}"
```
in JSON fields of included models.
### Environment

- Sequelize version: 6.9.0.
- Node.js version: 14.7.5.

### How does this problem relate to dialects?

- [ ] I think this problem happens regardless of the dialect.
- [X] I think this problem happens only for the following dialect(s): MariaDB

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
[...]

PR #14800, opened 2022-07-25
fix(mariadb): do not automatically parse JSON fields

lesion:
### Pull Request Checklist

### Description Of Change

MariaDB connector automatically parse JSON fields when `autoJsonMap` is specified in dialectOptions ([this is the default value](https://github.com/mariadb-corporation/mariadb-connector-nodejs/blob/master/documentation/connection-options.md#other-options)) so there's no need to `handleJsonSelectQuery`:

Sequelize tries to re-parse it causing issues (#10946, #13646, #12583, #13027, #13273) especially when the real value is a string (not a JSON string), because `handleJsonSelectQuery` checks for it:  
https://github.com/sequelize/sequelize/blob/ecf49d0102fe0447935b884a1c34c00d294f31db/src/dialects/mariadb/query.js#L187

note that with this PR JSON fields from included model are parsed automatically (same behaviour as with other backends).

example code to play with:
https://gist.github.com/lesion/0b3e001f75c1fa89593ff365c2e931f7


lesion:
https://github.com/mariadb-corporation/mariadb-connector-nodejs/issues/136

lesion:
this test should pass also with other connectors, why is sqlite required?

https://github.com/sequelize/sequelize/blob/main/test/integration/json.test.js#L232

the point is that it's the only test with a string as a JSON value: `emergency_contact: 'Unknown'` so could not pass because of

https://github.com/sequelize/sequelize/blob/ecf49d0102fe0447935b884a1c34c00d294f31db/src/dialects/mariadb/query.js#L187

this line has to be changed but as mariadb connector behave differently based on mariadb version...

src/dialects/mariadb/query.js
```diff
@@ -184,7 +185,7 @@ class Query extends AbstractQuery {
       if (modelField.type instanceof DataTypes.JSON) {
         // Value is returned as String, not JSON
         rows = rows.map(row => {
-          if (row[modelField.fieldName] && typeof row[modelField.fieldName] === 'string') {
+          if (row[modelField.fieldName] && typeof row[modelField.fieldName] === 'string' && !this.connection.info.hasMinVersion(10, 5, 2)) {
```
sdepold, line 188: Could we add a link to the docs that describe this behavior?

sdepold (changes requested):

src/dialects/mariadb/query.js
lesion: I'm not sure if you mean to add documentation to sequelize indicating that JSON fields for mariadb < 10.5.2 will be returned as strings and not as data (unlike other db's) or you mean to add a link in the code that indicates a reason for the mariadb version check: this is not clearly specified anywhere but from testing it turns out that it is, the reference links are anyway:
https://github.com/mariadb-corporation/mariadb-connector-nodejs/issues/136 and https://jira.mariadb.org/browse/MDEV-17832

lesion: also this PR points to v6 but I think it's the same with v7

sdepold: Yeah a link that references the Maria documentation and explains that the parsing happens automatically from a specific version on

lesion: added a comment.
also note that version check is already used with mariadb-connector-nodejs to test this same thing:

https://github.com/mariadb-corporation/mariadb-connector-nodejs/blob/59da962ce4ae6338bb14ed8854a70526b429911f/test/integration/datatype/test-json.js#L101

https://github.com/mariadb-corporation/mariadb-connector-nodejs/blob/59da962ce4ae6338bb14ed8854a70526b429911f/types/index.d.ts#L108

sdepold (approved):

src/dialects/mariadb/query.js
```diff
@@ -184,7 +185,9 @@ class Query extends AbstractQuery {
       if (modelField.type instanceof DataTypes.JSON) {
         // Value is returned as String, not JSON
         rows = rows.map(row => {
-          if (row[modelField.fieldName] && typeof row[modelField.fieldName] === 'string') {
+          // JSON fields for MariaDB server 10.5.2+ already results in JSON format, skip JSON.parse
+          // this is due to this https://jira.mariadb.org/browse/MDEV-17832 and how mysql2 connector interacts with MariaDB and JSON fields
+          if (row[modelField.fieldName] && typeof row[modelField.fieldName] === 'string' && !this.connection.info.hasMinVersion(10, 5, 2)) {
```
lesion, line 190: this is not correct:

https://github.com/encrypt94/sequelize/commit/7e6c9f769e88c9184d6bda9778d0b9e5411f90c9

WikiRik: Do I understand correctly that the commit by @encrypt94 is more correct than the current code? If so; feel free to open a PR for this

lesion: I've investigated, that's true, that commit is more correct.
The type of field returned by mariadb does not depend on the version of mariadb, but on the version of mariadb WHEN the table was created!

https://mariadb.com/kb/en/result-set-packets/#field-types
For a [JSON](https://mariadb.com/kb/en/result-set-packets/json) column, the column type field will be MYSQL_TYPE_STRING, but the extended type will indicate 'json'.

the connector use this field to decide to parse or not the response:
https://github.com/mariadb-corporation/mariadb-connector-nodejs/blob/59da962ce4ae6338bb14ed8854a70526b429911f/lib/cmd/decoder/text-decoder.js#L98

during a select the connector put fields details inside a "meta" key that could be used to check if returned value from the connector is already parsed or not:
https://github.com/mariadb-corporation/mariadb-connector-nodejs/blob/f25081715c9e1f24a10017e70787108d0d2ced01/documentation/promise-api.md#array-result-sets

a gist to test things:
https://gist.github.com/lesion/af0c7f1370ab6eb05cafdabd5b39ac33

PR #14804, opened 2022-07-26
fix: minified aliases are now properly referenced in subqueries

### LINKED ISSUE #14723: Impossible to do literal order by with subQuery and minifyAliases enabled
## Issue Creation Checklist

## Bug Description

When using `findAll()` in a Postgres sequelize instance with `minifyAliases` set to `true` and with the flag `subQuery` set to `true` (either explicitly or implicitly because of includes+limit) it's not easily possible to order by a custom literal. There are basically two ways of ordering by a custom attribute:
Using a literal order expression:
```typescript
await Foo.findAll({order: sequelize.literal(`"Foo".my_name DESC`)})
```

Using an attribute and order by that attribute:
```typescript
await Foo.findAll({attributes: {include: [[sequelize.literal(`"Foo".my_name`), "customAttribute"]]}, order: ["customAttribute", "DESC"]})
```

However when `subQuery` is `true` and `minifyAliases` is enabled both options don't work anymore and there's no easy way to make them work (without introducing another subquery within the literals).

E.g.
```typescript
await Foo.findAll({subQuery: true, order: sequelize.literal(`"Foo".my_name DESC`)})
```
produces the following SQL query:
```sql
SELECT "Foo".* FROM (SELECT "id", "my_name" AS "_0" FROM "Foos" AS "Foo" ORDER BY "Foo".my_name) AS "Foo" ORDER BY "Foo".my_name;
-- error: column Foo.my_name does not exist
```

When we do the attribute variant:
```typescript
await Foo.findAll({subQuery: true, attributes: {include: [[sequelize.literal(`"Foo".my_name`), "customAttribute"]]}, order: ["customAttribute", "DESC"]})
```
it produces the following SQL query which is invalid too:
```sql
SELECT "Foo".* FROM (
  SELECT "id", "my_name" AS "_0", "Foo".my_name AS "_1"
  FROM "Foos" AS "Foo"
  ORDER BY "Foo"."customAttribute"
) AS "Foo"
ORDER BY "Foo"."customAttribute";
-- error: column Foo.customAttribute does not exist
```

### Reproducible Example

Here is the link to the SSCCE for this issue: https://github.com/sequelize/sequelize-sscce/pull/229

Ideally the last query would produce this:
```sql
SELECT "Foo".* FROM (
  SELECT "id", "my_name" AS "_0", "Foo".my_name AS "_1"
  FROM "Foos" AS "Foo"
  ORDER BY "Foo"."_1"
) AS "Foo"
ORDER BY "Foo"."_1";
```

The last query produces this SQL query which is invalid:
```sql
SELECT "Foo".* FROM (
  SELECT "id", "my_name" AS "_0", "Foo".my_name AS "_1"
  FROM "Foos" AS "Foo"
  ORDER BY "Foo"."customAttribute"
) AS "Foo"
ORDER BY "Foo"."customAttribute";
-- error: column Foo.customAttribute does not exist
```

### Environment

- Sequelize version: 6.21.2 & 7.0.0-alpha.10
- Node.js version: v16.14.2
- If TypeScript related: TypeScript version:  4.5.5
- Database & Version: 10
- Connector library & Version: pg 8.7.3

- [x] No, I don't have the time, but my company or I are [supporting Sequelize through donations on OpenCollective](https://opencollective.com/sequelize): https://opencollective.com/ashby
- [ ] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in resolving my issue.

---

evanrittenhouse:
### Pull Request Checklist

This closes #14723. 

The PR expands checks on what constitutes a `subQueryAlias`. One thing to note is that the successful tests hinge on `this.quote(...)` being included in the  alias generation block. I'm not sure why that's the case, and I'm sure that it can be refactored to be cleaner.


evanrittenhouse:
@WikiRik Do you know why the integration tests would pass on my local but not in the CI/CD pipeline? `yarn test-postgres` passes. CC @sdepold

WikiRik:
Can you try to delete your test database? Maybe you have something in there locally that's not being removed with `Support.clearDatabase`

evanrittenhouse:
Sure, how do I do that?

WikiRik:
That depends, are you running the database locally using Docker? If there is nothing else running on that database, you can just remove that container and volume.
However you are running it, you can always connect to the database using either `psql` in the terminal or something like PgAdmin and remove the database there. But that's less thorough

evanrittenhouse:
Hmm. I am running `docker system prune --all --volumes` before testing, but the same thing is happening. Any other ideas? I could also probably move this test to a unit test instead of an integration test, then just check that the query string is correct, but that would be a longer time to deployment.

WikiRik:
Yeah, in that case I wouldn't really know what to do without a proper look at the code which I probably won't be able to do in the next two weeks because of vacation. So a unit test would be better. I believe that @ephys mentioned before that she wanted to use unit tests more so we can test `snowflake` and `ibmi` better

evanrittenhouse:
@WikiRik I know you're on vacation so no pressure to respond. I think that the tests are failing because I didn't change the behavior in other tests involving renaming columns when `minifyAliases` and `subQuery` are true. I'm not sure why they're passing on my local, but I think the easiest way forward would be to change how those tests work to reflect the new behavior. That's the approach I'm taking now.

WikiRik:
Can't hurt to try that

src/dialects/abstract/query-generator.js
```diff
@@ -2143,13 +2143,23 @@ export class AbstractQueryGenerator {
           && !isModelStatic(order[0].model)
           && !(typeof order[0] === 'string' && model && model.associations !== undefined && model.associations[order[0]])
         ) {
-          subQueryOrder.push(this.quote(order, model, '->', options));
+          // TESTS ONLY PASS IF THIS IS INCLUDED
+          this.quote(order, model, '->', options);
```
sdepold, line 2147: It seems that quote is actually manipulating the first argument if it is an array. If it is, it automatically replaces the argument with the model field's `field` property. Can you replace this code with the following?

```
          // if the field is aliased, we want to push the alias instead of the real field
          const field = model.rawAttributes[order[0]]?.field || order[0];
          const subQueryAlias = this._getAliasForField(modelName, field, options);
```

I'm not entirely happy with this since we have now duplicated logic

src/dialects/abstract/query-generator.js
```diff
@@ -2143,13 +2143,23 @@ export class AbstractQueryGenerator {
...
+          const modelName = this.quoteIdentifier(model.name);
+
+          // if the field is aliased, we want to push the alias instead of the real field
+          const subQueryAlias = this._getAliasForField(modelName, order[0], options);
+          subQueryOrder.push(this.quote(subQueryAlias == null ? order : subQueryAlias, model, '->', options));
```
sdepold, line 2152: Please use triple equality

sdepold (changes requested):

test/integration/dialects/postgres/query.test.js
```diff
@@ -135,5 +135,44 @@ if (dialect.startsWith('postgres')) {
...
+      });
+
+      const Foo = sequelizeMinifyAliases.define('Foo', {
+        name: {
+          field: 'my_name',
+          type: DataTypes.TEXT,
+        },
+      }, { timestamps: false });
+      await sequelizeMinifyAliases.sync({ force: true });
+
```
sdepold, line 156: My inner monk asks if you could move this blank line up so that all the await statements are nicely aligned in its own block

evanrittenhouse:
@sdepold comments resolved!

sdepold (approved):

PR #14825, opened 2022-08-01
fix(types): add keepDefaultTimezone to Options interface

BeeMargarida:
### Pull Request Checklist

### Description Of Change

The `Options` interface did not include the attribute `keepDefaultTimezone`, which can be received as an option (by looking at the Sequelize constructor).
This PR adds that attribute to the `Options` interface.


sdepold (changes requested):
I think we should add a description on what it's doing and the default value too?

BeeMargarida:
> I think we should add a description on what it's doing and the default value too?

I don't wanna commit mistakes, is the following correct?
```
Uses the default timezone to convert dates from the database.

@default false
```

sdepold:
It actually seems that the option is undefined by default, which is a bit of a problem. Should we explicitly set it to false here? https://github.com/sequelize/sequelize/blob/main/src/sequelize.js#L213

BeeMargarida:
> It actually seems that the option is undefined by default, which is a bit of a problem. Should we explicitly set it to false here? https://github.com/sequelize/sequelize/blob/main/src/sequelize.js#L213

Seems good to me, already made that change 👍 I'll push the commit

sdepold (changes requested):
One last detail: Can you add the new property to this test? https://github.com/sequelize/sequelize/blob/main/test/types/sequelize.ts#L11

BeeMargarida:
> One last detail: Can you add the new property to this test? https://github.com/sequelize/sequelize/blob/main/test/types/sequelize.ts#L11

Yap, done!

sdepold (approved):
Thanks :)

Issue #9109, opened 2018-02-26
findAndCount returns array instead of count when the query uses GROUP BY


jharting:
## What are you doing?

I am fetching a list of records (User) and counting the number of Projects each is associated with. I am using `[sequelize.fn('COUNT', sequelize.col('projects.id')), 'project_count']` together with `GROUP BY` to do that in a single query without fetching whole project entities.

Full testcase can be found here: https://github.com/jharting/sequelize/commit/90c2f7028aaed598d25b4d5c37879c8135d9eb79

## What do you expect to happen?
I expect the count field in the result to return **3**.

## What is actually happening?
The count field contains an array: `[ { count: 1 }, { count: 1 }, { count: 1 } ]`

__Dialect:__ mysql
__Database version__: 10.1.26
__Sequelize version__: 4.34.0
__Tested with master branch__: Yes

notflip:
Any news on this?

keyiis:
@jharting i facing the same issue. are you fixed?

papb:
@jharting Thanks for the testcase. Sorry to take so long for a response.

I reproduced this in every dialect. Interestingly, every dialect gives a different failure.

* SQlite: `AssertionError: expected [ Array(3) ] to equal 3`
* MariaDB: `SequelizeDatabaseError: (conn=10, no: 1054, SQLState: 42S22) Unknown column 'user.id' in 'field list'`
* MySQL: `SequelizeDatabaseError: Unknown column 'user.id' in 'field list'`
* Postgres: `SequelizeDatabaseError: missing FROM-clause entry for table "user"`
* MSSQL: `SequelizeDatabaseError: Column 'Users.name' is invalid in the select list because it is not contained in either an aggregate function or the GROUP BY clause.`

jeebus3000:
Any progress on this?

papb:
@jeebus3000 none that I know of

bentedder:
Is there still no resolution on this?

Hwan-seok:
Still... :(

RahulRawat1994:
Any progress

RahulRawat1994:
? :(

bitcyber:
Just use "result.count.length/count.length" instead "result.count/count" when you use Group by in findAndCountAll function.

PR #14841, opened 2022-08-05
fix: count grouped rows

### LINKED ISSUE #6148: findAndCountAll with `GROUP BY` clause causes count to be an array of objects
## What you are doing?

I'm trying to get the number of rows returned in using `findAndCountAll` with a `GROUP BY` clause however, instead of a single integer, I get an array of `{ count: x }` objects.

``` js
const attributes = ['feed_id', [this.store.fn('COUNT', this.store.col('feed_id')), 'count']];
const where = { createdAt: { $gte: start, $lte: end } };
const group = ['feed_id'];
const order = [['count', 'DESC']];
const query = { where, attributes, order, group };

model.findAndCountAll(query).then(data => {
  console.log(data.count);
  console.log(data.rows);
});
```

I expected a single integer in `count` and an array of objects in `row`.

```
{
  count: [
    { count: '11' },
    { count: '20' },
    { count: '17' },
    { count: '17' },
    { count: '11' },
    ...
    { count: '12' },
    { count: '10' },
    { count: '18' } ],
  rows: [{
    dataValues: [Object],
    _previousDataValues: [Object],
    _changed: {},
    '$modelOptions': [Object],
    '$options': [Object],
    hasPrimaryKeys: true,
    __eagerlyLoadedAssociations: [],
    isNewRecord: false
  }, ...]
}
```

I got an array of count objects.

For me, the workaround was to stop using `findAndCountAll` and make an explicit `model.count` transaction before the `findAll`. I then removed the `GROUP BY` clause and replaced the `COUNT` attribute with `COUNT(DISTINCT field)`.

```
SELECT COUNT(feed_id) AS "count" FROM "history" AS "history" WHERE ("history"."created_at" >= '2016-04-22 07:24:26.038 +00:00' AND "history"."created_at" <= '2016-06-21 07:24:26.040 +00:00') GROUP BY "feed_id";
```

```
SELECT COUNT(DISTINCT feed_id) AS "count" FROM "history" AS "history" WHERE ("history"."created_at" >= '2016-04-22 07:24:26.038 +00:00' AND "history"."created_at" <= '2016-06-21 07:24:26.040 +00:00')
```

Is this something sequelize can do automatically when it sees a `GROUP BY` clause?

**Dialect:** postgres
**Database version:** 9.5.2
**Sequelize version:** 3.23.0

fratzinger:
### Pull Request Checklist

### Description Of Change

This is related to several issues #6148, #9109. 

#### Summary of problem: 
When  using `options.group` for `Model.count` or `Model.findAndCountAll`, then `result.count` as an array of items instead of a number.

#### Solution:
When there's `options.group` then wrap the query into another `SELECT COUNT(*) FROM ()` query.
See this: https://stackoverflow.com/a/5147021/13155919

Considering that this issue exists since 2016 for 6 years now, the go to workaround seems to be:
> Just use "result.count.length/count.length" instead "result.count/count" when you use Group by in findAndCountAll function.
> *(see https://github.com/sequelize/sequelize/issues/9109#issuecomment-753955578)*
Using `result.count.length` produces a type error because `result.count` is `typeof 'number'`.

So to not break this workaround this PR introduces a new option: `countGroupedRows: boolean`. When it's `false` (default), the old behavior persists. If it's `true` and there's `option.group` then the new behavior (wrap query) will be used.

### Todos

What I'm unsure about is, if the `countGroupedRows` really is necessary. Is there a use case where `options.group` is used and you don't want to count the grouped rows? If not, we can get rid of `countGroupedRows` and make the wrap query the default behavior.

If we introduce `options.countGroupedRows` this PR is considered a **Feature**. If we don't introduce it this PR is considered a **Fix**

- [ ] Maintainers: decide wether we need `countGroupedRows` or wrap query should be the default behavior
- [ ] @fratzinger: dependant on the decision: add passage to docs / remove `countGroupedRows` again

Closes #6148 ##9109


evanrittenhouse (approved):
LGTM - thanks for submitting!

fzn0x (approved):
Looks good to me, I believe this changes is so good, but lets wait for the core maintainers reviews.

mhuggins:
Awesome!  Would love to see this make into a future release. 🤞

ephys:
I somehow completely missed this PR, thanks for the notification :)

I'll do a final review asap

AustinRobinson:
Any Update on this?

ericbf:
What’s the hold-up here?

WikiRik:
> What’s the hold-up here?

This got lost after december 2022. I can do another review once the merge conflicts are resolved

ericbf:
@fratzinger can you resolve the merge conflicts?

seho0808:
I resolved the merge conflicts (seems to be a very simple type change)
should i open a new PR, or what should i do?
(since i can't push directly to `fratzinger/sequelize` i wanted to create PR from `seho0808/sequelize/fix/counted-group-rows` to  `fratzinger/sequelize/fix/counted-group-rows` but simply failed to find the buttons in github to do so)

PR #14852, opened 2022-08-07
fix: minified aliases are now properly referenced in subqueries (v6)

sdepold:
Backport of #14804 to v6 + linting fixes

### DIFF
--- a/src/dialects/abstract/query-generator.js
+++ b/src/dialects/abstract/query-generator.js
@@ -2083,13 +2083,21 @@ class QueryGenerator {
           && !(typeof order[0].model === 'function' && order[0].model.prototype instanceof Model)
           && !(typeof order[0] === 'string' && model && model.associations !== undefined && model.associations[order[0]])
         ) {
-          subQueryOrder.push(this.quote(order, model, '->'));
+          const field = model.rawAttributes[order[0]]?.field || order[0];
+          const subQueryAlias = this._getAliasForField(this.quoteIdentifier(model.name), field, options);
+
+          subQueryOrder.push(this.quote(subQueryAlias === null ? order : subQueryAlias, model, '->'));
         }
 
         if (subQuery) {
           // Handle case where sub-query renames attribute we want to order by,
           // see https://github.com/sequelize/sequelize/issues/8739
-          const subQueryAttribute = options.attributes.find(a => Array.isArray(a) && a[0] === order[0] && a[1]);
+          // need to check if either of the attribute options match the order
+
+          const subQueryAttribute = options.attributes.find(a => Array.isArray(a)
+             && a[1]
+             && (a[0] === order[0] || a[1] === order[0]));
+
           if (subQueryAttribute) {
             const modelName = this.quoteIdentifier(model.name);
 
--- a/test/integration/dialects/postgres/query.test.js
+++ b/test/integration/dialects/postgres/query.test.js
@@ -102,5 +102,44 @@ if (dialect.match(/^postgres/)) {
         }
       });
     });
+
+    it('orders by a literal when subquery and minifyAliases are enabled', async () => {
+      const sequelizeMinifyAliases = Support.createSequelizeInstance({
+        logQueryParameters: true,
+        benchmark: true,
+        minifyAliases: true,
+        define: {
+          timestamps: false
+        }
+      });
+
+      const Foo = sequelizeMinifyAliases.define('Foo', {
+        name: {
+          field: 'my_name',
+          type: DataTypes.TEXT
+        }
+      }, { timestamps: false });
+
+      await sequelizeMinifyAliases.sync({ force: true });
+      await Foo.create({ name: 'record1' });
+      await Foo.create({ name: 'record2' });
+
+      const thisWorks = (await Foo.findAll({
+        subQuery: false,
+        order: sequelizeMinifyAliases.literal('"Foo".my_name')
+      })).map(f => f.name);
+      expect(thisWorks[0]).to.equal('record1');
+
+      const thisShouldAlsoWork = (await Foo.findAll({
+        attributes: {
+          include: [
+            [sequelizeMinifyAliases.literal('"Foo".my_name'), 'customAttribute']
+          ]
+        },
+        subQuery: true,
+        order: ['customAttribute']
+      })).map(f => f.name);
+      expect(thisShouldAlsoWork[0]).to.equal('record1');
+    });
   });
 }
--- a/test/unit/dialects/postgres/query-generator.test.js
+++ b/test/unit/dialects/postgres/query-generator.test.js
@@ -12,7 +12,7 @@ const chai = require('chai'),
   _ = require('lodash');
 
 const customSequelize = Support.createSequelizeInstance({
-  schema: 'custom',
+  schema: 'custom'
 });
 
 if (dialect.startsWith('postgres')) {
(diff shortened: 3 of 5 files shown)


WikiRik:
Do we need the linting fixes? They're not in files related to this PR and I would prefer to limit the changes that we do in v6

sdepold:
> Do we need the linting fixes? They're not in files related to this PR and I would prefer to limit the changes that we do in v6

Well. I just ran lint --fix and this was the result. I wonder how it could have worked before since I expected the linter to fail if there are things to change

sdepold:
I suppose I can remove the linting commit if we feel strong about it

sdepold:
@WikiRik better?

WikiRik:
Linter still passes, so I'm fine with this. Thanks!

WikiRik:
@tomquist Any chance you can test this on your project to make sure it works for your use case?

tomquist:
@WikiRik Thanks! I tested it on our project. It all works fine if we enforce a subquery using `subQuery:true` 🎉.
However I ran into one bug now that when it isn't doing a subquery it doesn't minify the alias only at the place where the the order by column is used. 
E.g. this doesn't work:
```
await Foo.findAll({attributes: {include: [[sequelize.literal(`"Foo".my_name`), "order_0"]]}, order: ["order_0", "DESC"]})
```

The error is:
> column _Foo.order_0 does not exist

WikiRik:
@evanrittenhouse could you check this new case out?

tomquist:
I think the issue I mentioned is an existing issue and nothing introduced by this bugfix. I can also open a new issue for the `subQuery: false` case if you like?

sdepold:
> @WikiRik Thanks! I tested it on our project. It all works fine if we enforce a subquery using `subQuery:true` 🎉. However I ran into one bug now that when it isn't doing a subquery it doesn't minify the alias only at the place where the order by column is used. E.g. this doesn't work:
> 
> ```
> await Foo.findAll({attributes: {include: [[sequelize.literal(`"Foo".my_name`), "order_0"]]}, order: ["order_0", "DESC"]})
> ```
> 
> The error is:
> 
> > column _Foo.order_0 does not exist

Do I understand the code right, that you want to have a query that aliases Foo.myname as order_0 and then you want to sort by order_0?

With this code in place, we are currently generating

```
SELECT "id", "my_name" AS "_0", "Foo".my_name AS "_1" FROM "Foos" AS "Foo" ORDER BY "Foo"."order_0", "Foo"."DESC";
```

tomquist:
Ah I see some brackets are missing. So the issue is that this:
```typescript
await Foo.findAll({attributes: {include: [[sequelize.literal(`"Foo".my_name`), "order_0"]]}, order: [["order_0", "DESC"]]})
```
generates this query:
```sql
SELECT "id", "my_name" AS "_0", "Foo".my_name AS "_1" FROM "Foos" AS "Foo" ORDER BY "Foo"."order_0" DESC;
```

sdepold:
@tomquist what would you expect instead?

tomquist:
I'd expect this:
```
SELECT "id", "my_name" AS "_0", "Foo".my_name AS "_1" FROM "Foos" AS "Foo" ORDER BY "_1" DESC;
```

sdepold:
Excellent. Thanks for the clarification

sdepold:
Just pushed a change that might fix the issue. It's funny because I just removed a guard that was specifically targeting subQueries and enabled the logic in general.

sdepold:
@tomquist @WikiRik @evanrittenhouse let's see if this works :)

sdepold:
interesting. so mssql is additionally adding a sorting in it's query generator on top of the abstract order by fragments.

sdepold:
@tomquist can you pleas test the new version? I'll figure out the mssql issue meanwhile

src/dialects/mssql/query-generator.js
```diff
@@ -976,12 +976,39 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
...
+
+          primaryKey = new Utils.Col(alias || aliasedAttribute[1]);
+        }
+
         if (!options.order || !options.order.length) {
           fragment += ` ORDER BY ${tablePkFragment}`;
         } else {
-          const orderFieldNames = _.map(options.order, order => order[0]);
-          const primaryKeyFieldAlreadyPresent = _.includes(orderFieldNames, model.primaryKeyField);
+          const orderFieldNames = (options.order || []).map(order => {
```
sdepold, line 996: Do we have a function that is doing this already? Taking a name like thing and returns its string content?

src/dialects/mssql/query-generator.js
```diff
@@ -976,12 +976,39 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
       // TODO: document why this is adding the primary key of the model in ORDER BY
       //  if options.include is set
       if (!options.order || options.order.length === 0 || options.include && orders.subQueryOrder.length === 0) {
-        const tablePkFragment = `${this.quoteTable(options.tableAs || model.name)}.${this.quoteIdentifier(model.primaryKeyField)}`;
+        let primaryKey = model.primaryKeyField;
+
+        const tablePkFragment = `${this.quoteTable(options.tableAs || model.name)}.${this.quoteIdentifier(primaryKey)}`;
+        const aliasedAttribute = (options.attributes || []).find(attr => Array.isArray(attr)
```
sdepold, line 982: This is a copy from the generic query generator. I was about to refactor this function but then wondered if `_getAliasForField` should somehow take care of this logic. I'm still not 100% sure what we are even doing here but I supposed, that this piece of code combined with getAliasForField is basically catering for 1) model definition based aliases (using the field property?) and 2) aliases that are introduced during the query calls (e.g. findAll).

Maybe instead of doing this manually, we should have a bit of code inside of _getAliasForField that is checking the options and extracts the call alias from them?

sdepold:
I'm a little bit lost on why MSSQL is having additional logic wrt order but it's probably there for good reasons. Anyway, because of this, it becomes a bit more convoluted and messy. We could clean up MSSQL but I wonder if that should happen in v7 only?

tomquist:
@sdepold Just tested with af1a638, and it made all our tests pass 🥳 Thanks a lot!

sdepold:
> @sdepold Just tested with [af1a638](https://github.com/sequelize/sequelize/commit/af1a63893998c44811cf6fb8c6574a501bea9d39), and it made all our tests pass 🥳 Thanks a lot!

Sweet. Thanks for the feedback!

src/dialects/mssql/query-generator.js
```diff
@@ -976,12 +976,39 @@ class MSSQLQueryGenerator extends AbstractQueryGenerator {
...
+            }
+
+            return value;
+          });
+          const primaryKeyFieldAlreadyPresent = orderFieldNames.some(
+            fieldName => fieldName === (primaryKey.col || primaryKey)
+          );
 
           if (!primaryKeyFieldAlreadyPresent) {
             fragment += options.order && !isSubQuery ? ', ' : ' ORDER BY ';
```
sdepold, line 1014: It's a bit crazy that this is done manually. I was tempted to somehow inject this idea into the incoming order options so that this.getQueryOrders would do the heavy lifting for us

sdepold:
@WikiRik @evanrittenhouse ready for revie

WikiRik:
Let's review #14882 first and then get back to this one

sdepold:
@WikiRik @evanrittenhouse can you review this guy?

WikiRik (approved):
Yeah, it's not the prettiest but it seems to work for v6 so we should be good

sdepold:
Thanks Rik!

pascuflow:
Was this the only change from 6.20.1? Am getting "No context available. ns.run() or ns.bind() must be called first." error when using transactions.

ephys:
Sounds like a continuation local storage issue, but we haven't touched that part of the code in recent releases

bminer:
This PR seems to be causing an issue for me.  See #15422

PR #14853, opened 2022-08-07
meta: update updated_at on upsert

joker00777:
### Pull Request Checklist

### Todos

- [ ] 
- [ ] 
- [ ]

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -2517,7 +2517,7 @@ Specify a different name for either index to resolve this issue.`);
       insertValues[field] = this._getDefaultTimestamp(createdAtAttr) || now;
     }
 
-    if (updatedAtAttr && !insertValues[updatedAtAttr]) {
+    if (updatedAtAttr && !updateValues[updatedAtAttr]) {
       const field = this.rawAttributes[updatedAtAttr].field || updatedAtAttr;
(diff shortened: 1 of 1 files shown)


WikiRik:
Interesting that tests pass without any changed test. Might that be because of line 2522?

Either way, can you provide a test that is failing on `main` but passes now with this change?

joker00777:
Sure : )

joker00777:
After reading the code , I think they have made an assumption that the user can send updatedAt during upsert and if they have send a custom updatedAt in that case we will use that otherwise we will update updatedAt to current time.

sdepold:
I tend to recall that we have just recently changed this to cater for the insertattributes. Maybe it should check both attributes?

Anyway, a test case is a good idea

joker00777:
insertValues have all the values including primaryKey in the other hand updateValues have all values other than primary key.
So I don;t think there is a test case which will fail while checking with insertValues.

WikiRik:
So what would be the benefit of this PR? Just a simple refactor in case we do something with `insertValues` in the future that would break this?

joker00777:
I think updated_at should come from updateValues as it makes sense there.In future this could help if updatedValues and insertValues differ from more than just primary key

WikiRik:
Agree

WikiRik (approved):

Issue #14750, opened 2022-07-13
Allow returning values after performing delete query in postgres


vladPopovschii:
## Issue Creation Checklist

- [x] I understand that my issue will be automatically closed if I don't fill in the requested information
- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

### Describe the feature you'd like to see implemented

Postgresql has support for return in delete query, it could be implemented like with returning in create

### Describe why you would like this feature to be added to Sequelize

Sometimes we delete an entity and need to send back to frontend other entities linked to that, for this we perform an additional select query. 

### Is this feature dialect-specific?

- [ ] No. This feature is relevant to Sequelize as a whole.
- [x] Yes. This feature only applies to the following dialect(s): postgresql

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [x] Yes, I have the time but I will need guidance.
- [ ] No, I don't have the time, but my company or I are [supporting Sequelize through donations on OpenCollective](https://opencollective.com/sequelize).
- [ ] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in implementing my feature.

---

*Indicate your interest in the addition of this feature by adding the 👍 reaction. Comments such as "+1" will be removed.*

vladPopovschii:
If someone had found an workaround i can use it

Vigneshwaran16:
You can try executing raw queries, with `RETURNING` clause.
`
  await sequelize.query(
     'DELETE FROM :table_name WHERE id = :user_id RETURNING *',
     {
        replacements: { table_name: 'users', user_id: 100 },
        type: QueryTypes.DELETE
     }
  );
`

Vigneshwaran16:
Hi @ephys , will this feature be available in upcoming release? If not, can I raise a PR for this feature?

WikiRik:
@Vigneshwaran16 feel free to open a PR for this!

PR #14879, opened 2022-08-13
feat(postgres): add returning option in delete query to return destroyed rows

Vigneshwaran16:
### Pull Request Checklist

### Description Of Change

Postgresql has a facility to return the destroyed rows through `returning` clause. Why is this needed? Sometimes, we may need to update other tables after `deletion`. This PR has added `returning` option to deleteQuery which will eventually return the deleted records. This PR is applicable for `postgres` only . 

### Related issues 
#14750

### Todos

- [x]  Add test cases for the proposed feature.
- [x]  Code  cleanup.


src/model.d.ts
```diff
@@ -2762,12 +2767,19 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
...
    */
+  static destroy<M extends Model>(
+    this: ModelStatic<M>,
+    options: Omit<DestroyOptions<Attributes<M>>, 'returning'>
+      & { returning: Exclude<DestroyOptions<Attributes<M>>['returning'], undefined | false> }
+  ): Promise<[deletedCount: number, deletedRows: M[]]>;
   static destroy<M extends Model>(
     this: ModelStatic<M>,
     options?: DestroyOptions<Attributes<M>>
-  ): Promise<number>;
```
WikiRik, line 2770: Can we keep this as a number? I think that's a nicer solution.

Vigneshwaran16: Yeah, I was 50/50 on that. I was hoping to get a review on that. Will keep it as a number :+1:

src/dialects/postgres/query-generator.js
```diff
@@ -392,10 +394,39 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
+      returnValues = this.generateValuesToReturn(options);
+      query = `${baseQuery}${whereClause}${returnValues}`;
+
+      return query;
+    }
+
+    return `${baseQuery}${whereClause}`;
+  }
+
+  generateValuesToReturn(options) {
```
sdepold, line 416: I somehow have the feeling that something like this might exist somewhere already.

src/model.js
```diff
@@ -2950,6 +2950,7 @@ Specify a different name for either index to resolve this issue.`);
       force: false,
       cascade: false,
       restartIdentity: false,
+      returing: false,
```
sdepold, line 2953: Typo

src/dialects/postgres/query-generator.js
WikiRik: We do indeed have similar code in the abstract query generator for returning values for inserted/updates fields;
https://github.com/sequelize/sequelize/blob/2deee3c37ea3ff9e5d159538b3a1c067fc6d3ba2/src/dialects/abstract/index.ts#L26-L30
https://github.com/sequelize/sequelize/blob/2deee3c37ea3ff9e5d159538b3a1c067fc6d3ba2/src/dialects/abstract/query-generator.js#L1844-L1890

Vigneshwaran16: I used the updateQuery(postgres) as a template to structure the code for this deleteQuery.

src/model.js
Vigneshwaran16: My bad. Will fix it,  along with changes for review comments.

src/dialects/postgres/query-generator.js
sdepold: Cool. Can we utilize the existing function?

WikiRik:
Could you check out the unit tests before we do a further review?

src/dialects/postgres/query-generator.js
Vigneshwaran16: Updated code to utilize the existing one.

test/unit/query-interface/delete.test.ts
```diff
@@ -39,4 +39,88 @@ describe('QueryInterface#delete', () => {
...
+          id: '123',
+        },
+      },
+    );
+
+    expect(stub.callCount).to.eq(1);
+    const firstCall = stub.getCall(0);
+    expectsql(firstCall.args[0] as string, {
+      default: 'DELETE FROM `Users` WHERE `id` = \':id\'',
+      postgres: `DELETE FROM "Users" WHERE "id" = ':id' RETURNING "id","firstName"`,
```
Vigneshwaran16, line 63: Is it better to leave it as such to run for all dialects or bring these tests under an if condition to run only for postgres?

sdepold: I think something like you have is fine. Please make use of default however (so db2, ibmi, snowflake etc is not necessary it seems).

Vigneshwaran16:
I've added tests for the newly added functionality !

src/dialects/postgres/query-generator.js
```diff
@@ -392,10 +394,23 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
       primaryKeys = model.primaryKeyAttributes.length > 1 ? `(${pks})` : pks;
       primaryKeysSelection = pks;
 
-      return `DELETE FROM ${table} WHERE ${primaryKeys} IN (SELECT ${primaryKeysSelection} FROM ${table}${whereClause}${limit})`;
+      query = `${baseQuery} WHERE ${primaryKeys} IN (SELECT ${primaryKeysSelection} FROM ${table}${whereClause}${limit})`;
+      if (options.returning) {
+        returnValues = this.generateReturnValues(model.rawAttributes, options);
```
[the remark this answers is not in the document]
sdepold: Actually. Couldn't we just run this line where the variable is declared above and use the value here ?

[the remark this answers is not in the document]
Vigneshwaran16: Is there any other changes that needs to be done for this PR?

Vigneshwaran16:
@WikiRik Is there any other changes required for this PR?

test/integration/model/paranoid.test.js
```diff
@@ -39,7 +39,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       const count2 = await Account.count();
       expect(count2).to.be.equal(1);
       const result = await Account.destroy({ where: { ownerId: 12 } });
-      expect(result).to.be.equal(1);
+      expect(result).to.be.a('number');
```
WikiRik, line 42: I would add to this that the number equals 1

test/unit/query-interface/delete.test.ts
```diff
@@ -39,4 +39,116 @@ describe('QueryInterface#delete', () => {
...
+      },
+    );
+
+    expect(stub.callCount).to.eq(1);
+    const firstCall = stub.getCall(0);
+    expectsql(firstCall.args[0] as string, {
+      default: 'DELETE FROM `Users` WHERE `id` = \':id\'',
+      postgres: `DELETE FROM "Users" WHERE "id" = ':id' RETURNING *`,
+      mssql: `DELETE FROM [Users] WHERE [id] = N':id'; SELECT @@ROWCOUNT AS AFFECTEDROWS;`,
+      snowflake: `DELETE FROM "Users" WHERE "id" = ':id'`,
```
WikiRik, line 65: You should be able to delete the special cases of snowflake, db2 and ibmi since the default will handle the differences in the apostrophes etc

test/unit/query-interface/delete.test.ts
```diff
@@ -39,4 +39,116 @@ describe('QueryInterface#delete', () => {
...
+          id: '123',
+        },
+      },
+    );
+
+    expect(stub.callCount).to.eq(1);
+    const firstCall = stub.getCall(0);
+    expectsql(firstCall.args[0] as string, {
+      default: 'DELETE FROM `Users` WHERE `id` = \':id\'',
+      postgres: `DELETE FROM "Users" WHERE "id" = ':id'`,
```
WikiRik, line 147: That counts for postgres as well here

WikiRik (commented):
Few small remarks, other than this I think we should be good to merge. I'll give @sdepold some time to review this again as well

test/unit/query-interface/delete.test.ts
Vigneshwaran16: Differences in apostrophes couldn't be handled without special cases for dialects. In few tests(not in this PR), special cases are used for each dialects in assertions. For now, could I leave it as it is?

WikiRik: You need to update default to the following so that the difference in apostrophes are used.
```ts
default: `DELETE FROM [Users] WHERE [id] = ':id'`,
```

WikiRik:
@ephys don't think you've seen this PR before. I think it looks good, but maybe you have some suggestions regarding the types? Or maybe suggestions for future refactoring since you added dialect specific typing to the QueryGenerator and QueryInterface files in the DataTypes PR

test/unit/query-interface/delete.test.ts
Vigneshwaran16: Got it !!

ephys:
Indeed :) I'd like to take a look this evening before merging
One thing I can already point out is I think dialects that do not support "returning" on delete queries should throw if returning is specified.
Can be done by adding a flag for it in the dialect support table (in the different index.js files, but as far as I know only postgres supports this right?)

I know other returning options don't throw when unsupported (yet), but we're slowly making sequelize stricter

src/dialects/postgres/query-generator.js
```diff
@@ -401,7 +401,9 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
     const limit = options.limit ? ` LIMIT ${this.escape(options.limit, undefined, _.pick(options, ['replacements', 'bind']))}` : '';
     let primaryKeys = '';
     let primaryKeysSelection = '';
-
+    const baseQuery = `DELETE FROM ${table}`;
+    let query = '';
```
ephys, line 405: declare query on line 421 & 430 instead, it's not used anywhere else (we aim to declare variables as "late" as possible)

src/model.js
```diff
@@ -2971,6 +2972,7 @@ Specify a different name for either index to resolve this issue.`);
     }
 
     let result;
```
ephys, line 2974: declare result on line 2994 instead

test/integration/model.test.js
```diff
@@ -1436,7 +1436,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       await this.sequelize.sync({ force: true });
       await User.bulkCreate([{ username: 'user1' }, { username: 'user2' }]);
       const affectedRows = await User.destroy({ where: {} });
-      expect(affectedRows).to.equal(2);
+      expect(affectedRows).to.be.a('number').to.equal(2);
```
ephys, line 1439: There is no need to check that the value is a number if you already check that is equals a number value

```suggestion
      expect(affectedRows).to.equal(2);
```

Same comment for the rest of the changes in this file

test/integration/model.test.js
```diff
@@ -1698,10 +1698,81 @@ describe(Support.getTestDialectTeaser('Model'), () => {
 
       await this.User.bulkCreate(data);
       let affectedRows = await this.User.destroy({ where: { secretValue: '42' } });
-      expect(affectedRows).to.equal(2);
+      expect(affectedRows).to.be.a('number').to.equal(2);
       affectedRows = await this.User.destroy({ where: { secretValue: '44' } });
-      expect(affectedRows).to.equal(0);
+      expect(affectedRows).to.be.a('number').to.equal(0);
     });
+    if (dialect === 'postgres') {
```
ephys, line 1705: use the Support table instead:

```suggestion
    if (this.sequelize.dialect.supports.delete.returning) {
```

test/integration/model.test.js
```diff
@@ -1698,10 +1698,81 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+        [affectedRowsCount, affectedRows] = await this.User.destroy({ where: {}, returning: [] });
+        expect(affectedRowsCount).to.be.a('number');
+        expect(affectedRowsCount).to.equal(3);
+        expect(affectedRows).to.be.an('array').to.have.lengthOf(3);
+        expect(affectedRows[0]).to.be.instanceOf(this.User);
+        JSON.parse(JSON.stringify(affectedRows)).forEach(userObj => {
+          expect(userObj).to.include.keys('secretValue', 'username', 'id');
+        });
+      });
+    }
```
ephys, line 1775: Can you add a test that ensures an error is thrown if the dialect uses 'returning', but it isn't supported in that dialect?

```suggestion
    } else {
      it('throws an error when "returning" is used in an unsupported dialect', () => {
        await expect(this.User.destroy({ returning: true })).to.be.rejectedWith(`${dialect} does not support the "returning" option in DELETE queries.`);
      });
    }
```

test/integration/model/paranoid.test.js
```diff
@@ -39,7 +39,7 @@ describe(Support.getTestDialectTeaser('Model'), () => {
       const count2 = await Account.count();
       expect(count2).to.be.equal(1);
       const result = await Account.destroy({ where: { ownerId: 12 } });
-      expect(result).to.be.equal(1);
+      expect(result).to.be.a('number').equal(1);
```
ephys, line 42:
```suggestion
      expect(result).to.equal(1);
```

test/integration/model.test.js
```diff
@@ -1698,10 +1698,81 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+        { username: 'Rob', secretValue: '41' },
+        { username: 'Billy', secretValue: '40' },
+      ];
+      it('returns the affected rows if `options.returning` is set to true', async function () {
+        await this.User.bulkCreate(data);
+        let [affectedRowsCount, affectedRows] = await this.User.destroy({ where: { secretValue: '42' }, returning: true });
+        expect(affectedRowsCount).to.be.a('number');
+        expect(affectedRowsCount).to.equal(2);
+        expect(affectedRows).to.be.an('array').to.have.lengthOf(2);
+        [affectedRowsCount, affectedRows] = await this.User.destroy({ where: {}, returning: true });
```
ephys, line 1719: is this part of the test necessary?

Each test is executed with both a where & no where. If you want to test interactions with `where`, add a test that does that but don't add it to each one

test/integration/model.test.js
```diff
@@ -1698,10 +1698,81 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+      });
+
+      it('returns only the fields in affected rows specified in `options.returning`', async function () {
+        await this.User.bulkCreate(data);
+        let [affectedRowsCount, affectedRows] = await this.User.destroy({ where: { secretValue: '42' }, returning: ['username'] });
+        expect(affectedRowsCount).to.be.a('number');
+        expect(affectedRowsCount).to.equal(2);
+        expect(affectedRows).to.be.an('array').to.have.lengthOf(2);
+        expect(affectedRows[0]).to.be.instanceOf(this.User);
+        JSON.parse(JSON.stringify(affectedRows)).forEach(userObj => {
```
ephys, line 1742: why serialize / deserialize this value? You could just check that the value is or isn't undefined on the model

test/unit/dialects/postgres/query-generator.test.js
```diff
@@ -1071,6 +1071,28 @@ if (dialect.startsWith('postgres')) {
         },
       ],
 
+      deleteQuery: [
```
ephys, line 1074: Prefer adding these tests to `test/unit/query-generator/delete-query.test.ts` instead.
We're working on removing dialect specific tests. Instead we want to test the behavior on all dialects. 
In this case, all dialects should expect an error to be thrown, except PG

test/unit/dialects/postgres/query-generator.test.js
```diff
@@ -1071,6 +1071,28 @@ if (dialect.startsWith('postgres')) {
...
+          arguments: ['myTable', { name: 'foo' }, { returning: ['*'] }, 'myTable'],
+          expectation: `DELETE FROM "myTable" WHERE "name" = 'foo' RETURNING *`,
+        }, {
+          arguments: ['myTable', { name: 'foo' }, { returning: ['name', 'title'] }, 'myTable'],
+          expectation: `DELETE FROM "myTable" WHERE "name" = 'foo' RETURNING "name","title"`,
+        }, {
+          arguments: ['myTable', { name: 'foo' }, 'myTable'],
+          expectation: `DELETE FROM "myTable" WHERE "name" = 'foo'`,
+        }, {
+          arguments: ['myTable', { name: 'foo' }, { returning: [] }, 'myTable'],
```
ephys, line 1091: `returning: []` should be equivalent to `returning: false`, not `returning: true`

test/unit/dialects/postgres/query-generator.test.js
```diff
@@ -1071,6 +1071,28 @@ if (dialect.startsWith('postgres')) {
...
+      deleteQuery: [
+        {
+          arguments: ['myTable', { name: 'foo' }, { returning: true }, 'myTable'],
+          expectation: `DELETE FROM "myTable" WHERE "name" = 'foo' RETURNING *`,
+        }, {
+          arguments: ['myTable', { name: 'foo' }, { returning: false }, 'myTable'],
+          expectation: `DELETE FROM "myTable" WHERE "name" = 'foo'`,
+        }, {
+          arguments: ['myTable', { name: 'foo' }, { returning: ['*'] }, 'myTable'],
+          expectation: `DELETE FROM "myTable" WHERE "name" = 'foo' RETURNING *`,
```
ephys, line 1083: `*` should not have special meaning in the `returning` array. It's valid in PG to name a column "*" (unlikely, yes).

This query should produce this:

```suggestion
          expectation: `DELETE FROM "myTable" WHERE "name" = 'foo' RETURNING "*"`,
```

`returning: true` will be used for `RETURNING *`

test/unit/dialects/postgres/query-generator.test.js
```diff
@@ -1071,6 +1071,28 @@ if (dialect.startsWith('postgres')) {
...
+          arguments: ['myTable', { name: 'foo' }, { returning: ['name', 'title'] }, 'myTable'],
+          expectation: `DELETE FROM "myTable" WHERE "name" = 'foo' RETURNING "name","title"`,
+        }, {
+          arguments: ['myTable', { name: 'foo' }, 'myTable'],
+          expectation: `DELETE FROM "myTable" WHERE "name" = 'foo'`,
+        }, {
+          arguments: ['myTable', { name: 'foo' }, { returning: [] }, 'myTable'],
+          expectation: `DELETE FROM "myTable" WHERE "name" = 'foo' RETURNING *`,
+        },
+      ],
```
ephys, line 1094: Consider adding the following tests to check that literals are accepted:

```js
arguments: ['myTable', { name: 'foo' }, { returning: [literal('*'), literal('abc')] }, 'myTable'],
expectation: `DELETE FROM "myTable" WHERE "name" = 'foo' RETURNING *, abc`,
```

```js
arguments: ['myTable', { name: 'foo' }, { returning: literal('*') }, 'myTable'],
expectation: `DELETE FROM "myTable" WHERE "name" = 'foo' RETURNING *`,
```

test/unit/query-interface/delete.test.ts
```diff
@@ -39,4 +39,116 @@ describe('QueryInterface#delete', () => {
 
     expect(firstCall.args[1]?.bind).to.be.undefined;
   });
+
+  it('returns fields specified in `options.returning`', async () => {
```
ephys, line 43: prefer moving these tests to `unit/query-generator/delete-query.test.ts` instead. QueryInterface only tests that replacements and bind parameters are handled correctly, but they are not relevant to your PR

ephys (changes requested):
In `src/dialects/abstract/index.ts`, we have the `DialectSupports` object that is a map of supported features by dialect. It is a bit of a mess but you can add a new Property:

```
export type DialectSupports = {
  // ...
  delete: {
    /** This dialect supports returning deleted values in DELETE queries */
    returning: boolean,
  },
  // ...
};
```

Then set the default value in the AbstractDialect class:

```
  static readonly supports: DialectSupports = {
    // ...
    delete: { returning: false },
    // ...
  };
```

---

Then in `dialect/postgres/index.js`, you can tag it as supporting the feature:

```
  static supports = _.merge(_.cloneDeep(AbstractDialect.supports), {
    // ...
    delete: { returning: true },
    // ...
  });
```

Issue #14840, opened 2022-08-05
Bug with mixins when using with custom unique constraint in M:N relation


JaoodxD:
## Issue Creation Checklist

- [x] I understand that my issue will be automatically closed if I don't fill in the requested information
- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description

In short [association mixins](https://sequelize.org/docs/v6/core-concepts/assocs/#special-methodsmixins-added-to-instances) ignores custom rules of uniqueness.  
When we explicitly set [`unique: false`](https://sequelize.org/docs/v6/core-concepts/assocs/#options-2) we expect that every time we use `add` mixin, sequelize will generate new record inside associated table, but it doesn't.

In general the problem is that model mixins incorrectly interacts with unique constraints.

Let's say we have model `A` and `B`.

We need M:N relation between them.
We establish it through junction table `C`.

This relation should contain additional boolean field `flag`.
At our case we need to extend default unique constraint with this `flag` 
i.e. uniqueness should be determined by three fields `(Aid, BId, flag)`.

To do this we have to perform the following 2 steps:
1.  Turn off default constraint by passing `unique : false` inside `belongsToMany` relation definition.
2.  Define our own unique combined key with both `AId`, `BId` and `flag`.

Now we suggest that we can relate `A` with `B` twice (but with different `flag` value).
It works fine if we directly add two records inside `C` :

    const a = await A.create();
    const b = await B.create();
    await C.create({ 'Aid': a.Id, 'Bid': b.Id, 'flag': true });
    await C.create({ 'Aid': a.Id, 'Bid': b.Id, 'flag': false }); // <- flag is different
    console.log(await C.count()); // <- will display us 2 as expected

But if we do the same with the help of mixins `addB` we'll get different output:

    const a = await A.create();
    const b = await B.create();
    await a.addB(b, { through: { flag: true } });
    await a.addB(b, { through: { flag: false } }); // <- same different values
    console.log(await C.count()); // will display us 1 instead of 2.

### Reproducible Example

Here is the link to the SSCCE for this issue: https://github.com/JaoodxD/sequelize-sscce-m-m-bug

### What do you expect to happen?

When we turn off the unique constraint inside `belongsToMany` declaration
mixin `addB` should create two different records inside junction table `C`.

### What is actually happening?

Mixin `addB` still check only `(Aid, BId)` constraint and update previous record.

### Environment

[...]

PR #14890, opened 2022-08-16
fix: Association mixin bug with unique false 

Necromansr:
### Pull Request Checklist

### Description Of Change

Fixed association mixin bug `add` which ignores `unique: false` constraint. 
This pull request relates and fixes existing issue #14840

Previously `add` mixin completely ignores `unique: false` . 

In case where we want to implement own unique constraint/extend existing one
 we should use [`unique: false`](https://sequelize.org/docs/v6/core-concepts/assocs/#options-2) inside `belongsToMany` association. 
If we do this we expect that we can establish relation between 2 entities more than once.
Previously we could reach this only inside parallel querying:
```javascript
await Promise.all([
 a.addB(a, { through: { flag: false } }),
 a.addB(a, { through: { flag: true } })
]);
console.log((await a.getBs()).length); // will display 2
```
Which obviously gives us expected result only due to race condition (which is not good).

And If we do the same in sequential manner
```javascript
await a.addB(a, { through: { flag: false } });
await a.addB(a, { through: { flag: true } })


console.log((await a.getBs()).length); // will display 1
```
we will get different unexpected result.
The solution is to extend exising `add` method inside `belongs-to-many.ts` 
with additional if-statement check of `unique: false` `this.through` object.


sdepold:
Thanks for your change :) Can you add a test for this as well?

JaoodxD:
> Thanks for your change :) Can you add a test for this as well?

Added chai tests for this case.

src/associations/belongs-to-many.ts
```diff
@@ -646,17 +646,21 @@ Add your own primary key to the through model, on different attributes than the
...
-        [this.otherKey]: newInstances.map(newInstance => newInstance.get(this.targetKey)),
-        ...this.through.scope,
-      },
-      // force this option to be false, in case the user enabled
-      rejectOnEmpty: false,
-    });
+    let currentRows: any[] = [];
+    if (this.through?.unique ?? true) {
+      currentRows = await this.through.model.findAll({
+        // const currentRows = await this.through.model.findAll({
```
WikiRik, line 652:
```suggestion
```

test/integration/associations/belongs-to-many.test.js
```diff
@@ -3538,4 +3538,97 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), () => {
...
+      });
+    });
+
+    it('should create new relations in parallel', async function () {
+      await this.sequelize.sync({ force: true });
+
+      const { Order, Good } = this;
+
+      const order = await Order.create();
+      const good = await Good.create({ name: 'Coca-Cola' });
```
WikiRik, line 3580:
```suggestion
      const good = await Good.create({ name: 'Drink' });
```
Other suggestions are fine as well but I'd prefer not to include brand names

test/integration/associations/belongs-to-many.test.js
```diff
@@ -3538,4 +3538,97 @@ describe(Support.getTestDialectTeaser('BelongsToMany'), () => {
...
+        .equal(secondGood.orderGood.isUpsell);
+    });
+
+    it('should create new relations sequentialy', async function () {
+      await this.sequelize.sync({ force: true });
+
+      const { Order, Good } = this;
+
+      const order = await Order.create();
+      const good = await Good.create({ name: 'Coca-Cola' });
```
WikiRik, line 3611:
```suggestion
      const good = await Good.create({ name: 'Drink' });
```
Same as above

JaoodxD:
@WikiRik your fixes make sense, ty.
Waiting for the review to become a contributor of this community 😉.

PR #14901, opened 2022-08-18
feat: improved error reporting on associations

baileykue:
### Pull Request Checklist

    - Updated existing tests
    - Passing except for tests mentioned in #14161 which failed before changes were made

### Description Of Change

This improves the error reporting on associations by telling us what exactly is
not a proper subclass of the Sequelize Model. This assists in debugging and
should present little risk to the existing behavior.

Possibly due to #14161 I was not able to get the full test suite running
locally _before_ I made any local changes. However, the number of tests that
were failing (2) remained the same after my changes, and the failing tests had
more to do with SQLite than changes to the model structure.

Thanks for maintaining such an awesome open source project!


src/associations/helpers.ts
```diff
@@ -217,7 +271,11 @@ export function defineAssociation<
   construct: (opts: CleanOptions) => T,
 ): T {
   if (!isModelStatic(target)) {
-    throw new Error(`${source.name}.${lowerFirst(type.name)} called with something that's not a subclass of Sequelize.Model`);
+    throw new Error(
+      `${source.name}.${lowerFirst(
+        type.name,
+      )} called with ${target} that's not a subclass of Sequelize.Model`,
```
WikiRik, line 277: There's a lot of styling changes in this PR. Is it true that this is the only actual change?

baileykue: Yes, this was the only change. Sorry! The rest was auto formatting that happened on commit.

test/unit/associations/has-one.test.ts
```diff
@@ -12,7 +12,9 @@ describe(getTestDialectTeaser('hasOne'), () => {
     expect(() => {
       // @ts-expect-error
       User.hasOne();
-    }).to.throw('User.hasOne called with something that\'s not a subclass of Sequelize.Model');
+    }).to.throw(
+      'User.hasOne called with undefined that\'s not a subclass of Sequelize.Model',
```
WikiRik, line 16: I think this test is the reason why we used `something`. In my opinion `undefined` is less useful. Can we maybe make it that if it's undefined we still print `something` and if we know what it is we use that target?

test/unit/associations/has-many.test.ts
```diff
@@ -13,21 +17,28 @@ describe(getTestDialectTeaser('hasMany'), () => {
     expect(() => {
       // @ts-expect-error
       User.hasMany();
-    }).to.throw('User.hasMany called with something that\'s not a subclass of Sequelize.Model');
+    }).to.throw(
+      'User.hasMany called with undefined that\'s not a subclass of Sequelize.Model',
```
WikiRik, line 21: Same as my other comment

test/unit/associations/belongs-to.test.ts
```diff
@@ -12,7 +12,9 @@ describe(getTestDialectTeaser('belongsTo'), () => {
     expect(() => {
       // @ts-expect-error
       User.belongsTo();
-    }).to.throw('User.belongsTo called with something that\'s not a subclass of Sequelize.Model');
+    }).to.throw(
+      'User.belongsTo called with undefined that\'s not a subclass of Sequelize.Model',
```
WikiRik, line 16: And this one as well

test/unit/associations/belongs-to-many.test.ts
```diff
@@ -23,25 +36,31 @@ describe(getTestDialectTeaser('belongsToMany'), () => {
     expect(() => {
       // @ts-expect-error
       User.belongsToMany();
-    }).to.throw('User.belongsToMany called with something that\'s not a subclass of Sequelize.Model');
+    }).to.throw(
+      'User.belongsToMany called with undefined that\'s not a subclass of Sequelize.Model',
```
WikiRik, line 40: Same as my other comments, but I'll still do it so I don't have to go through all styling changes again

WikiRik (changes requested):
The four comments are about the same thing.

The styling changes have made it more difficult to review this PR, but I've went through it all and did not see any other changes so we should be okay

test/unit/associations/has-one.test.ts
LoganBarnett: Hi @WikiRik! I helped @baileykue put together this pull request. This is her first! Not that should motivate accepting changes :)

Our motivation for printing `target` in the error was because we were encountering `undefined` but had some confusion. Our first guess was "Was something wrong with how we setup our model?". Granted, if we were on TypeScript, it would've identified that specific error and we would've been fine. However jumping to TypeScript isn't something we're able to do the project at this time. While I'd not want to inflict our project's lack of advancement upon this project, our thought when we encountered this error message was "is there any benefit in withholding `target` here?". We thought the answer to that was a plain "no". But maybe there is benefit?

Re: Formatting changes - is there something we could've done better here? These seemed to run as part of the pre-commit hook. I assumed this meant y'all were doing a reformat-as-you-go approach. I apologize that this has caused some confusion as to the material changes of this pull request.

WikiRik: Hi! Thanks the message and the work of both of you on this.

Don't worry about the TypeScript usage, this is something we are working on as well. I do agree that adding the target can be worth it, but I am unsure on what developers might think if they see `undefined` here. Maybe we can do an adjusted message if it's undefined so it makes more sense that this is not an issue within Sequelize?

Regarding the formatting; the pre-commit hook should do the same as the lint check which passed before (and still does now) so it's not defined by our project. But overall I do like the changes so don't worry about it.

LoganBarnett: @WikiRik I think I misunderstood your original request - thanks for clarifying! To echo it back to make sure I understand:

You're thinking we should have two paths here for the error message: One for when the error is clearly incorrect (`null` or `undefined`) vs. another when the object passed is not a subclass of `Sequelize.Model`.

Something like this between the two error messages?

For `undefined` et. al:

```
`User.hasOne` called with undefined but needs to be an object that is a subclass of `Sequelize.Model`
```
For an object that's not the right subclass:

```
`User.hasOne` called with an object that's not a subclass of `Sequelize.Model`
```

Does that sound about right?

WikiRik: Yes, that would be great

sdepold: Sounds good!

test/unit/associations/belongs-to-many.test.ts
sdepold: btw. not sure if our linter will allow this but feel free to use backticks instead of single quotes to save the escaping inside of the string

src/associations/helpers.ts
ephys, line 277: Main thing was to use `inspect()`, but I did some rephrasing too to add some information:

```suggestion
      )} was called with ${NodeUtils.inspect(target)} as the target model, but it is not a subclass of Sequelize's Model class`,
```

ephys (changes requested):
Thank you for the PR. That error message definitely deserved some improvements

I'm sorry to ask this but could you revert the stylistic changes? They increase the risk of conflict with the other PRs, make this PR more difficult to read as well the the commit history. We'd like to restrict style changes to PRs dedicated to that.

test/unit/associations/belongs-to-many.test.ts
ephys: It should allow it

src/associations/helpers.ts
LoganBarnett: It's a very reasonable ask to revert the stylistic changes. I assumed there was a pre-commit hook at work here, but instead it sounds like it was just some aggressive settings in the editor that made the changes. We definitely want to minimize the diff as much as possible :)

test/unit/associations/has-many.test.ts
```diff
@@ -13,7 +13,7 @@ describe(getTestDialectTeaser('hasMany'), () => {
     expect(() => {
       // @ts-expect-error
       User.hasMany();
-    }).to.throw('User.hasMany called with something that\'s not a subclass of Sequelize.Model');
+    }).to.throw('User.hasMany was called with undefined as the target model, but it is not a subclass of Sequelize\'s Model class');
```
WikiRik, line 16:
```suggestion
    }).to.throw(`User.hasMany was called with undefined as the target model, but it is not a subclass of Sequelize's Model class`);
```

test/unit/associations/has-one.test.ts
```diff
@@ -12,7 +12,7 @@ describe(getTestDialectTeaser('hasOne'), () => {
     expect(() => {
       // @ts-expect-error
       User.hasOne();
-    }).to.throw('User.hasOne called with something that\'s not a subclass of Sequelize.Model');
+    }).to.throw('User.hasOne was called with undefined as the target model, but it is not a subclass of Sequelize\'s Model class');
```
WikiRik, line 15:
```suggestion
    }).to.throw(`User.hasOne was called with undefined as the target model, but it is not a subclass of Sequelize's Model class`);
```

WikiRik (commented):
Just a small readability suggestion.

Also, can you look into the previous comment as well?

> You're thinking we should have two paths here for the error message: One for when the error is clearly incorrect (`null` or `undefined`) vs. another when the object passed is not a subclass of `Sequelize.Model`.
> 
> Something like this between the two error messages?
> 
> For `undefined` et. al:
> 
> ```
> `User.hasOne` called with undefined but needs to be an object that is a subclass of `Sequelize.Model`
> ```
> 
> For an object that's not the right subclass:
> 
> ```
> `User.hasOne` called with an object that's not a subclass of `Sequelize.Model`
> ```

test/unit/associations/has-one.test.ts
baileykue: All the style changes have been reverted, and the error message has been revised. Thank you for your comments, let me know how it looks now!

src/associations/helpers.ts
baileykue: I have reverted the stylistic changes, should look good now. I also changed the error message to reflect what you had suggested. Thank you for your comments!

ephys (dismissed):
Once @WikiRik's style changes have been applied, this should be good to go :)

WikiRik (approved):
I've gone ahead and committed my suggestion so we should be good to go! Thanks for working on this!

LoganBarnett:
@WikiRik thanks so much for dragging this over the finish line! We got pretty caught up with some work duties.

PR #14963, opened 2022-09-07
fix: decode encoded value from `urlParts.pathname` for `database` option

### LINKED ISSUE #14962: Spaces are not encoded/decoded correctly in the Postgres database name when using connection string
## Issue Creation Checklist

## Bug Description

When I try connecting to a Postgres database that has a space inside the name, I get an error that is not there when using a config object. As can be seen in the example, everything is working when connecting using `pg` only, so the issue is in `sequelize` parameter handling. I tried debugging the issue and found a workaround, by using the `beforeConnect` hook. Apparently, the `database` property is correct inside `dialectOptions`, when it's there and correct in `config` when it's missing, so I simply pick the first one if it's there. Now, I'm not sure that it will work all of the time, so I would like to raise the issue for someone familiar with Sequelize internals to take a look at.

### Reproducible Example

Please note that I made modifications to the database name and `createSequelize6Instance`

Here is the link to the SSCCE for this issue:
https://github.com/wkrasnicki-vizlib/sequelize-sscce

I expect to be able to use database names with space, inside connection string.

It seems that the `url` lib from Node parses different parts of the URL differently. There is no issue with spaces in username or password, but in case of database name we get an URL encoded value, which is then used when connecting to Postgres, even though we want to have an unencoded value. Providing an encoded DB name also doesn't work.

### Environment

- Sequelize version: 6.21.4
- Node.js version: 7.0.0-alpha.10
- If TypeScript related: TypeScript version:
- Database & Version: PostgreSQL 10
- Connector library & Version: `pg@8.8.0

- [x] Yes, I have the time but I will need guidance.
- [ ] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in resolving my issue.

---

fzn0x:
### Pull Request Checklist

### Description Of Change

Close #14962

`options.database` returning bad value for database connection, this affected to all database except the `options.dialect` with `postgres` as the value. https://github.com/sequelize/sequelize/blob/a68b439fb3ea748d3f3d37356d9fe610f86184f6/src/utils/url.ts#L71

You can notice the strange thing by running this code with NodeJS:
```js
import url from 'url';

const urlParts = url.parse("pg://wpx%20ss:wpx%20ss@21.77.77:4001/database ss", true);

const options = {}

// adding decodeURI solve the database issue.
options.database = decodeURI(urlParts.pathname.replace(/^\//, ''));
console.log(urlParts.auth);

// urlParts.auth already decode the encoded URI value by default
if (urlParts.auth) {
  const authParts = urlParts.auth.split(':');
  options.username = authParts[0];
  if (authParts.length > 1) {
    options.password = authParts.slice(1).join(':');
  }
}

console.log(options)
```

### Todos

- [x] Add decodeURI for handling urlParts.pathname value
- [ ] Maybe adding tests?


src/utils/url.ts
```diff
@@ -26,7 +26,8 @@ export function parseConnectionString(connectionString: string): Options {
   }
 
   if (urlParts.pathname) {
-    options.database = urlParts.pathname.replace(/^\//, '');
+    // decode URI from urlParts.pathname value
+    options.database = decodeURI(urlParts.pathname.replace(/^\//, ''));
```
wkrasnicki-vizlib, line 30: Since we are not decoding the entire connection string, shouldn't we use `decodeURIComponent` here?

WikiRik: Looking at[ the difference between the two](https://stackoverflow.com/questions/747641/what-is-the-difference-between-decodeuricomponent-and-decodeuri), it indeed seems better to use `decodeURIComponent`
Maybe we can add a unit test for decoding the special characters?

fzn0x: I'd like to use the decodeURIComponent from start, but since we only want to decode the resulting encoding which are spaces and nonprintable characters I think using the decodeURIComponent is a bit overkill, as I'm not sure the database name will need the same thing as query strings.

fzn0x: > Since we are not decoding the entire connection string, shouldn't we use `decodeURIComponent` here?

I'd like to use the decodeURIComponent from the start, but since we only want to decode the resulting encoding which are spaces and nonprintable characters I think using the decodeURIComponent is a bit overkill, as I'm not sure the database name will need the same things as query strings.

fzn0x: > Looking at[ the difference between the two](https://stackoverflow.com/questions/747641/what-is-the-difference-between-decodeuricomponent-and-decodeuri), it indeed seems better to use `decodeURIComponent` Maybe we can add a unit test for decoding the special characters?

I'm not sure if the special characters is really allowed for database naming cause it may leading to unexpected errors.

fzn0x: But yeah if you guys have good reason to use decodeURIComponent, yep I would agree to move from decodeURI

fzn0x:
Do you have any suggestion for the unit test? @WikiRik 

Since the Utils does not include `url` functions on sequelize/core https://github.com/sequelize/sequelize/blob/a68b439fb3ea748d3f3d37356d9fe610f86184f6/test/unit/utils/utils.test.ts#L3 

We need to add a another PR that includes the `url` utility or all the other utilities too and release it to new v7 alpha version, you can take a part for this, or the alternative solution is I am to import the url utility directly.

fzn0x:
Or skip the unit test also fine for me, since the `decodeURIComponent` itself is the function coming from `nodejs` core, which is the part of the built-in we probably no doubt to use.

src/utils/url.ts
WikiRik: Ah true, special characters in database naming might not be smart. We could test that with an integration test as well

WikiRik:
> Do you have any suggestion for the unit test? @WikiRik 

I was thinking we could add unit tests for `parseConnectionString` to check if everything is parsing correctly. I do think that testing `decodeURIComponent` separately is not needed.

fzn0x:
Hi @WikiRik did you have any inputs left on this PR or we good to go?

WikiRik:
Looks good to me, but I'll wait with approval until further comment by @wkrasnicki-vizlib

wkrasnicki-vizlib:
Yes, this solves the issue. Thank you guys

wkrasnicki-vizlib (approved):

WikiRik (approved):

fzn0x:
Glad it helps, thank you guys!

Issue #14974, opened 2022-09-11
[BUG] v7-alpha.10 - using underscored makes createdAt/modifiedAt get ignored


nawlbergs:
## Issue Creation Checklist

- [x] I understand that my issue will be automatically closed if I don't fill in the requested information
- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description

TLDR: 
```js
const options = {
  define:{
    underscored: true, // <-- turning this on..
    createdAt: 'createdOn', // <-- ignores this and uses created_at
    updatedAt: 'modifiedOn', // <-- ignores this and uses updated_at
  }
}
const foo: any = new Sequelize(databaseURL, options);
```
This did not do this in older versions of sequelize. (pretty sure)

### What do you expect to happen?

I would expect it to use the names for the timestamps I provided.  Not generate them.

### What is actually happening?

It's ignoring the names provided and using defaults when `underscored: true` is enabled.

### Environment

- Sequelize version: `@sequelize/core@7.0.0-alpha.10`
- Node.js version: `16.17.0`
- TypeScript version: `typescript@4.8.3`
- Database & Version: `postgres:11.13`
- Connector library & Version: `pg-hstore@2.3.4, pg@8.8.0`

## Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I will need guidance.
- [ ] No, I don't have the time, but my company or I are [supporting Sequelize through donations on OpenCollective](https://opencollective.com/sequelize).
- [x] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in resolving my issue.

---

*Indicate your interest in the resolution of this issue by adding the 👍 reaction. Comments such as "+1" will be removed.*

fzn0x:
I'm not sure this bug started from `7.0.0-alpha.10`. Which older versions did you mean?

evanrittenhouse:
@fzn0x @nawlbergs I can take this on. Do you have any idea when the issue started occurring?

evanrittenhouse:
[...]

PR #14982, opened 2022-09-14
meta: add test based on #14974

WikiRik:
### Pull Request Checklist

### Description Of Change

Draft PR to test #14974. The added integration test is based on the following unit test;
https://github.com/sequelize/sequelize/blob/af0f024acdcb4f885b9969a958c5fb4752abfff8/test/unit/model/define.test.js#L15-L31

### Todos

- [ ] 
- [ ] 
- [ ]

### DIFF
--- a/test/integration/model.test.js
+++ b/test/integration/model.test.js
@@ -600,6 +600,24 @@ describe(Support.getTestDialectTeaser('Model'), () => {
         expect(idx2.unique).not.to.be.ok;
       }
     });
+
+    it('should allow custom timestamps with underscored: true', async function () {
+      const User = this.sequelize.define('UserCustomTimestampUnderscored', {}, {
+        createdAt: 'createdAt',
+        updatedAt: 'updatedAt',
+        timestamps: true,
+        underscored: true,
+      });
+
+      expect(User.rawAttributes).to.haveOwnProperty('createdAt');
+      expect(User.rawAttributes).to.haveOwnProperty('updatedAt');
+
+      expect(User._timestampAttributes.createdAt).to.equal('createdAt');
+      expect(User._timestampAttributes.updatedAt).to.equal('updatedAt');
+
+      expect(User.rawAttributes).not.to.have.property('created_at');
+      expect(User.rawAttributes).not.to.have.property('updated_at');
+    });
   });
 
   describe('build', () => {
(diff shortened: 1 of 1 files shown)


evanrittenhouse:
@WikiRik If this passes with the new name as `createdAt`/`updatedAt`, we may want to try again with `createdOn`/`updatedOn`. That's what the original issue mentioned and where I got the issue locally.

WikiRik:
I've updated the test, it still passes on sqlite locally but we'll see what postgres and other dialects will do

test/integration/model.test.js
```diff
@@ -600,6 +600,23 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+
+    it('should allow custom timestamps with underscored: true', async function () {
+      const User = this.sequelize.define('UserCustomTimestampUnderscored', {}, {
+        createdAt: 'createdOn',
+        updatedAt: 'modifiedOn',
+        underscored: true,
+      });
+
+      expect(User.rawAttributes).to.haveOwnProperty('createdOn');
+      expect(User.rawAttributes).to.haveOwnProperty('modifiedOn');
```
ephys, line 612: Maybe the issue is the DB name:

```suggestion
      expect(User.rawAttributes).to.haveOwnProperty('modifiedOn');
      expect(User.rawAttributes.modifiedOn.field).to.eq('modified_on');
```

WikiRik: What do you mean with the DB name? Or is that unrelated to your suggestion?

ephys: The key in `rawAttributes` is the JS name of the property. Since you mentioned this test passes in sqlite, it's possible that the issue is that the column name (db side) is incorrectly generated (as this is an issue with `underscored` apparently). `User.rawAttributes.modifiedOn.field` returns which name will be used for the column

WikiRik: sqlite was just the first dialect I tried but the test passed for other dialects as well. I'm not sure what the failing test is that @evanrittenhouse was able to come up with.

Even with the added `User.rawAttributes.modifiedOn.field` it seems to pass the test. So I'm not sure what's wrong here

evanrittenhouse: @WikiRik I saw it fail when I put this test https://github.com/sequelize/sequelize/blob/af0f024acdcb4f885b9969a958c5fb4752abfff8/test/unit/model/define.test.js#L15-L31 into an integration test. Unfortunately, I don't have more information as it was uncommitted work on the branch that I've been using to work on the new issue associated with #14723.

WikiRik: I think I used that test, possibly without `timestamp: true`, as well and it passed on CI. But I can try it again tomorrow

Either way, I think it might also be related to the usage of sequelize-typescript as I've asked in the original issue

WikiRik: Nope, that passes as well so I'll close this PR since I can't seem to reproduce this error

PR #14989, opened 2022-09-16
docs: add oracle in readme

fzn0x:
Add Oracle dialect, as it is supported now in version `6.22.0`.


sdepold (approved):

WikiRik:
We should add this to v6 as well. And maybe we should add in main that Oracle support is only in v6

fzn0x:
> We should add this to v6 as well. And maybe we should add in main that Oracle support is only in v6

Is something like `Oracle DB (v6 only)` and sync the readme to v6 and main a wise changes?

fzn0x:
> We should add this to v6 as well. And maybe we should add in main that Oracle support is only in v6

Is something like `Oracle DB (v6 only)` and sync the readme to v6 a wise changes?

PR #14991, opened 2022-09-16
fix(types): missing type for oracle dialect in v6

fzn0x:
The same as https://github.com/sequelize/sequelize/pull/14137

This is causing casting with any, just notice it when try the oracle DB.

Closes none but this is to prevent the future issues

### DIFF
--- a/src/sequelize.d.ts
+++ b/src/sequelize.d.ts
@@ -169,7 +169,7 @@ export interface Config {
   };
 }
 
-export type Dialect = 'mysql' | 'postgres' | 'sqlite' | 'mariadb' | 'mssql' | 'db2' | 'snowflake';
+export type Dialect = 'mysql' | 'postgres' | 'sqlite' | 'mariadb' | 'mssql' | 'db2' | 'snowflake' | 'oracle';
 
 export interface RetryOptions {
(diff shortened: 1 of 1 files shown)


fzn0x:
This changes target v6 only since Oracle DB dialect was only in v6 branch.

fzn0x:
This changes targets v6 only since Oracle DB dialect was only in v6 branch.

sdepold (approved):

sdepold:
👍

sdepold:
Please don't merge before answering this: Will the commit message cause a patch or a minor version bump?

fzn0x:
> Please don't merge before answering this: Will the commit message cause a patch or a minor version bump? We should aim for the patch

I'm agree with the patch option 👍

sdepold:
I think you need to change your commit message to fix instead of feat for that

fzn0x:
> I think you need to change your commit message to fix instead of feat for that

Changed, is there anything required for the patch?

sdepold:
Can you rename the commit message too?

fzn0x:
> Can you rename the commit message too?

Looks like the branch does not allow rebase, maybe I will create another PR.

PR #14994, opened 2022-09-16
meta: use default behaviour instead of force to lf

fzn0x:
#13377 is a great PR, but this changes is better since it will let the git installation on a user's machine decide which line ending to use when a repository is checked out rather than only force to LF, it will help file size smaller (with the default of github line endings).

This PR will help windows user will keep the correct line endings and cross-platform enabled.


fzn0x:
Since we using squash and merge this, what's better types to use for the PR title? I'm stick with `fix` for now

https://github.com/sequelize/sequelize/blob/0ea005a3cb83b3dc058d9a2e34db2aeb028f8a95/.github/workflows/pr-title.yml#L17

fzn0x:
Since we using squash and merge approach, I don't think we really mind about the commit messages, what's better types to use for the PR title? I'm stick with `fix` for now

https://github.com/sequelize/sequelize/blob/0ea005a3cb83b3dc058d9a2e34db2aeb028f8a95/.github/workflows/pr-title.yml#L17

WikiRik:
I would use `meta` since this should not affect other projects using sequelize

fzn0x:
> I would use `meta` since this should not affect other projects using sequelize

Thanks for the input :)

sdepold:
Does that mean that we have different like endings in different platforms? Isn't that causing problems?

sdepold:
Hm apparently I approved that other pr. I cannot remember ever seeing this other one 🙈

ephys:
From what I understand based on [this]https://git-scm.com/docs/gitattributes#_end_of_line_conversion), files should still be normalized to lf when pushed.

Is this an issue though? The only time I had issues with lf on windows was with nodepad. Haven't seen any other text editor or IDE that didn't support lf

Not sure about the filesize argument, cr+lf is larger than lf and we already use lf in the repo

Anyway, as long as it doesn't cause crlf to be pushed to the repo, I'm fine with it

.gitattributes
```diff
@@ -1 +1,7 @@
-* -lf
\ No newline at end of file
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Explicitly declare text files you want to always be normalized and converted
+# to native line endings on checkout.
```
ephys, line 5:
```suggestion
# to unix line endings on checkout.
```

.gitattributes
```diff
@@ -1 +1,7 @@
...
+*.js text
```
WikiRik, line 6: Why only `.js` and `.ts`?

fzn0x:
> From what I understand based on [this]https://git-scm.com/docs/gitattributes#_end_of_line_conversion), files should still be normalized to lf when pushed.
> 
> Is this an issue though? The only time I had issues with lf on windows was with nodepad. Haven't seen any other text editor or IDE that didn't support lf
> 
> Not sure about the filesize argument, cr+lf is larger than lf and we already use lf in the repo
> 
> Anyway, as long as it doesn't cause crlf to be pushed to the repo, I'm fine with it

I believe some editors like VSCode always using CRLF by default, actually Windows itself always using CRLF for the text files while linux and mac always using the LF one.

.gitattributes
fzn0x: > Why only `.js` and `.ts`?

Makes sense 👍

.gitattributes
fzn0x: Thanks

fzn0x: 👍

fzn0x:
Screenshot 2022-09-17 171525

Just checked my Windows and looks like VSCode using CRLF by default.

fzn0x:
This PR isn't very useful if developers know to always use LF for the line endings, but if they accidentally use CRLF, it really helps avoid conflicts.

ephys:
Yes but that's not the issue, vscode is perfectly able to view & edit files that use LF. Does it cause any issue to windows users to enforce the use of LF? Most IDEs will perfectly adapt to that

fzn0x:
> Yes but that's not the issue, vscode is perfectly able to view & edit files that use LF. Does it cause any issue to windows users to enforce the use of LF? Most IDEs will perfectly adapt to that

Nah it shouldn't, but it will great to avoid CRLF being pushed to the repo.

fzn0x:
> Anyway, as long as it doesn't cause crlf to be pushed to the repo, I'm fine with it

I agree 👍

ephys:
> but it will great to avoid CRLF being pushed to the repo.

I'm confused, isn't that what the current config already does? :o I thought that was the goal of https://github.com/sequelize/sequelize/pull/13377

fzn0x:
> Does that mean that we have different like endings in different platforms? Isn't that causing problems?

It will always use the default git line endings on checkout even we are on different platforms and environments

fzn0x:
> > but it will great to avoid CRLF being pushed to the repo.
> 
> I'm confused, isn't that what the current config already does? :o I thought that was the goal of #13377

The line endings are committed to the repository "normalized" as LF, but the checked out version depends on the OS default for line endings.

fzn0x:
This changes solves the issue for the checked out version

fzn0x:
> > but it will great to avoid CRLF being pushed to the repo.
> 
> I'm confused, isn't that what the current config already does? :o I thought that was the goal of #13377


The line endings are committed to the repository "normalized" as LF, but the checked in version depends on the OS default for line endings.

Source [Kernel.org](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitattributes.html#_end_of_line_conversion):

> When text is set to "auto", the path is marked for automatic end-of-line normalization. If git decides that the content is text, its line endings are normalized to LF on checkin.
If you want to interoperate with a source code management system that enforces end-of-line normalization, or you simply want all text files in your repository to be normalized, you should instead set the text attribute to "auto" for all files.
This ensures that all files that git considers to be text will have normalized (LF) line endings in the repository.

This changes solves the issue on checkin.

fzn0x:
But please note that 

> mac always using the LF one

This is not correct, Mac OS has always used the CR, but the Mac OSX using the LF one.

ephys:
Nowadays when people say Mac OS, they mean macOS, which uses LF. No one uses classic Mac OS, which did use CR, it's been dead since 2001

fzn0x:
> Nowadays when people say Mac OS, they mean macOS, which uses LF. No one uses classic Mac OS, which did use CR, it's been dead since 2001

Haha thanks for the information😃

WikiRik:
@ephys do you happen to know if eslint will fail when we try to check in CRLF lines?

ephys:
It will, we have `'linebreak-style': [2, 'unix'],`, but it will only check JS/TS files

WikiRik:
Hmm, that's also not good behaviour. Maybe we can do something with our pre-commit hook that converts all CRLF to LF?

fzn0x:
> Hmm, that's also not good behaviour. Maybe we can do something with our pre-commit hook that converts all CRLF to LF?

It is something sounds better, I would happy to close this PR if that's possible

ephys:
> Maybe we can do something with our pre-commit hook that converts all CRLF to LF?

So check out locally as CRLF then when committing convert to LF so eslint doesn't fail, then convert back to CRLF to avoid a local diff issue then have git convert back to LF when pushing?

I fail to see a good reason to use CRLF on windows, all editors (except notepad.exe) support LF. I think we should just enforce the use of LF, which from what I understand is what #13377 did.

@soryy708 as you worked on the original PR, do you have any insight on this issue?

WikiRik:
Yeah, I was thinking of allowing CRLF locally (if that's something we want in the end) and force converting to LF before committing. But I would be fine with enforcing LF everywhere locally.

ephys:
I don't think we can convert to LF during the pre-commit hook, then revert back when the pre-commit hook fails/succeeds without causing a lot of issues

WikiRik:
I think the conclusion is that the current behaviour works fine and changing it might require more work than we really have benefits from it. @fzn0x would you agree if we close it?

fzn0x:
> I think the conclusion is that the current behaviour works fine and changing it might require more work than we really have benefits from it. @fzn0x would you agree if we close it?

Sure, no doubt to close it

Issue #12449, opened 2020-07-01
Unable to specify full schema in MSSQL


Vaelek:
## Issue Description

when using the schema option of a Model, a database cannot specified.

### What are you doing?

On a table model defining `schema: "SomeDatabase.dbo"`

### What do you expect to happen?

The query to be executed as `[...] FROM [SomeDatabase].[dbo].[Tablename]`

### What is actually happening?

The query is executed as `[...] FROM [SomeDatabase.dbo].[Tablename]`
Naturally this fails.

### Environment

- Sequelize version: 6.2.0
- Node.js version: 13.8.0
- Operating System: Win10 Pro 64bit

### How does this problem relate to dialects?
- [X] I think this problem happens only for the following dialect(s): MSSQL

### Would you be willing to resolve this issue by submitting a Pull Request?
- [X] Yes, I have the time and know how to start

I have solved it by changing https://github.com/sequelize/sequelize/blob/945af927db60e964eba7e344665ed0594ee63231/lib/dialects/abstract/query-generator/helpers/quote.js#L68

to
 
```js
let cleanIdentifier = identifier.replace(/[[\]']+/g, '');
return cleanIdentifier.endsWith('.dbo') ? cleanIdentifier.split('.').map(part => `[${part}]`).join('.') : `[${cleanIdentifier}]`;
```

This allows me to specify `SomeDatabase.dbo` or `[SomeDatabase].[dbo]` or even `[SomeDatabase.dbo]` and it will appear in the query correctly as `[SomeDatabase].[dbo].[TableName]`
The ternary handles field names so `[blah].[id] AS [blah.id]` doesn't end up as `[blah].[id] AS [blah].[id]`

I'm sure there is a more elegant way, and this assumes all tables will be on the `dbo` schema.

I have seen several mentions in other issues as well as I think the docs which state both that the schema can be used to select a different database from the connection, as well as that 2 instances of Sequelize must be used to support 2 databases. Assuming everything about the connection is the same except for the database name, the former is true only because of this issue.

Vaelek:
After trying to do this "properly" and use 2 connections, I find that this was brought up several years ago and apparently swept under the rug in #6225. Making a real effort to not monkey patch but it is becoming increasingly difficult. Is there no legit way to properly join tables from 2 MSSQL databases with Sequelize?

hoangcao21:
I run into the same problem.

alex-jss:
#13277

bkozlik:
[...]

PR #14999, opened 2022-09-17
feat: implement proper schema support in mysql, snowflake, add `queryInterface.listDatabases`

ephys:
## Pull Request Checklist

## Description Of Change

This PR blocks https://github.com/sequelize/sequelize/pull/14687

This PR aims to fix the way we support schemas in mysql.

Before this PR, creating a table "users" in the schema "archive" would result in a table `archive.users` being created in the default schema.

After this PR, it would create the table `users` in the schema `archive`.

I believe the source of the issue is a confusion caused by the fact that in mysql/mariadb, `CREATE SCHEMA` and `CREATE DATABASE` are synonymous. However, the behavior of mysql's schema/database is that of a schema (as understood by the rest of our codebase and how other dialects define what a schema is), not a database.

Surprisingly, our MariaDB implementation already did this correctly. All I had to (more or less) was hoist the implementation from MariaDB to MySQL, as MariaDB extends MySQL.

## Other Changes

- Added an integration test for `createDatabase` & `dropDatabase`, did you know we had none?
- Added a new method `queryInterface.listDatabases` (needed it for integration tests)
- Added typings for `queryGenerator.{create,drop}DatabaseQuery`, as well as `queryGenerator.{create,drop}SchemaQuery`, `queryGenerator.listSchemasQuery`
- `createDatabaseQuery` and `createSchemaQuery`: now throws if it receives an option that specific dialect does not support (so people don't think the option is supported just because it's listed in our types)
- Moved their tests to group them by function rather than by dialect. Makes it much easier to determine whether one of them is doing something wrong.
- Disabled schema tests in dialects that silently ignored schemas, because they now throw.
- `QueryGenerator#{create,drop}Schema` are now named `QueryGenerator#{create,drop}SchemaQuery` to be aligned with other `QueryGenerator` methods. `showSchemasQuery` renamed to `listSchemasQuery`
- Added support for the `skip` option in all implementations of `listSchemasQuery`
- Fixed a few values not being escaped in ibmi
- Implemented schema methods in snowflake

## TODO

- [x] Finish proper schema support
- [x] Unify createSchema/dropSchema tests

## Squash Commit Message Body

```

BREAKING CHANGE: the 'schema' option in MySQL now references actual schemas instead of adding a prefix to table names.
```


src/dialects/mariadb/query-generator.js
```diff
@@ -1,83 +1,9 @@
 'use strict';
 
 const { MySqlQueryGenerator } = require('../mysql/query-generator');
-const Utils = require('./../../utils');
 const _ = require('lodash');
 
 export class MariaDbQueryGenerator extends MySqlQueryGenerator {
-  createSchema(schema, options) {
```
ephys, line 8: Everything deleted in this file has been moved to `MySqlQueryGenerator`

test/integration/dialects/mariadb/query-interface.test.js
```diff
@@ -1,24 +0,0 @@
...
-const expect = chai.expect;
-const Support = require('../../support');
-
-const dialect = Support.getTestDialect();
-
-if (dialect.startsWith('mariadb')) {
-  describe('QueryInterface', () => {
-
-    describe('databases', () => {
-      it('should create and drop database', async function () {
```
ephys, line 14: replaced with `create-drop-list-database.test.ts`, which runs on all dialects instead of just mariadb

WikiRik:
I haven't really looked at the code yet, just mainly the description;
- Do I understand correctly that people using MySQL have to update the names and schema of the database tables in order to upgrade to v7? So if a table is now called `archive.users` they need to update the name to `users` and set the schema to `archive`? Can you provide an example query to do this in the upgrade guide?
- Is there a backwards compatible config option for this? Probably not enabled by default since I can understand that people might want to use . in their table names

ephys:
> So if a table is now called archive.users they need to update the name to users and set the schema to archive? Can you provide an example query to do this in the upgrade guide?

That is correct, and they have two ways of migrating:

## Option 1: Rename & move the tables

Let's say you had a table called `archive.users` in the default schema, you can rename it & move it to the archive schema like this:

```sql
CREATE SCHEMA `archive`;
ALTER TABLE `archive.users` RENAME TO `archive`.`users`;
```

## Option 2: Rename your model

If you have a model that uses the schema option, like the following example,

```js
sequelize.define('Users', {/* attributes */}, {
  tableName: 'users',
  schema: 'archive',
});
```

you can continue using the old naming by changing its definition to this:

```js
sequelize.define('Users', {/* attributes */}, {
  tableName: 'archive.users',
});
```

> Is there a backwards compatible config option for this?

Because the migration is decently simple if you want to continue with the old fake schema (see option 2 above), and considering the complexity such an option would add, I don't think we should add one

WikiRik:
Ah yes, I forgot about option 2. But that's a very valid workaround and indeed does not require further backwards compatibility

test/support.ts
```diff
@@ -441,6 +443,18 @@ export function expectsql(
     }
   }
 
+  if (typeof query === 'function') {
+    try {
+      query = query();
+    } catch (error: unknown) {
+      if (!(error instanceof Error)) {
+        throw new TypeError('expectsql: function threw something that is not an instance of Error.');
```
fzn0x, line 451: This doesn't seems helpful

fzn0x: I think catch always throws an error, just the instance is different, so maybe we can provide the error based on the instance.

fzn0x (approved):
Other than that, looks good to me, many improvements on the syntax too.

src/dialects/db2/query-generator.js
```diff
@@ -9,7 +10,8 @@ const { AbstractQueryGenerator } = require('../abstract/query-generator');
 const randomBytes = require('crypto').randomBytes;
 const { Op } = require('../../operators');
 
-/* istanbul ignore next */
```
WikiRik, line 12: We keep this in other dialects (or at least in mssql), maybe we can keep it here as well?

src/dialects/mysql/query-generator.js
```diff
@@ -162,12 +172,12 @@ export class MySqlQueryGenerator extends AbstractQueryGenerator {
...
-  showTablesQuery(database) {
-    let query = 'SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = \'BASE TABLE\'';
-    if (database) {
-      query += ` AND TABLE_SCHEMA = ${this.escape(database)}`;
+  showTablesQuery(schemaName) {
+    let query = 'SELECT TABLE_NAME, TABLE_SCHEMA FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = \'BASE TABLE\'';
+    if (schemaName) {
+      query += ` AND TABLE_SCHEMA = ${this.escape(schemaName)}`;
     } else {
-      query += ' AND TABLE_SCHEMA NOT IN (\'MYSQL\', \'INFORMATION_SCHEMA\', \'PERFORMANCE_SCHEMA\', \'SYS\', \'mysql\', \'information_schema\', \'performance_schema\', \'sys\')';
```
WikiRik, line 170: Why has `SYS` been removed from here? And the lowercase versions? Since we do use those in `schemasToSkip`. I see that that's from the original MariaDB implementation but I'm not sure if that's correct for MySQL

src/dialects/mysql/query-generator.js
```diff
@@ -628,18 +639,6 @@ export class MySqlQueryGenerator extends AbstractQueryGenerator {
 
     return `json_unquote(json_extract(${quotedColumn},${pathStr}))`;
   }
-
-  _createBindParamCollector(bindContext /* : BindContext */) {
```
WikiRik, line 632: At first I was wondering why you removed this here as it doesn't relate to the scope of this PR, but I don't see it being used so good that we remove it

test/integration/model.test.js
```diff
@@ -2109,298 +2112,264 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
-        sqlite: 1,
-        ibmi: 3,
-      };
-      expect(schemas).to.have.length(expectedLengths[dialect]);
-    });
+        expect(schemas.sort()).to.deep.equal(expectedSchemas[dialectName].sort());
+      });
 
-    if (['mysql', 'sqlite'].includes(dialect)) {
-      it('should take schemaDelimiter into account if applicable', async function () {
```
WikiRik, line 2157: I haven't looked through all the test files yet when making this comment, but am I correct that this test has been removed?

test/unit/query-generator/create-schema-query.test.ts
```diff
@@ -0,0 +1,36 @@
+import { buildInvalidOptionReceivedError } from '@sequelize/core/_non-semver-use-at-your-own-risk_/utils/check.js';
+import { expectsql, getTestDialect, sequelize } from '../../support';
+
+const dialectName = getTestDialect();
+
+const notSupportedError = new Error(`Schemas are not supported in ${dialectName}.`);
+
+describe('QueryGenerator#createSchemaQuery', () => {
```
WikiRik, line 8: We had an unit test for both `collate` and `chatset` before, but it's not here. Is that something that we want to add or do we feel it's redundant?

test/unit/query-generator/create-database-query.test.ts
```diff
@@ -0,0 +1,61 @@
+import { buildInvalidOptionReceivedError } from '@sequelize/core/_non-semver-use-at-your-own-risk_/utils/check.js';
+import { expectsql, getTestDialect, sequelize } from '../../support';
+
+const dialectName = getTestDialect();
+
+const notSupportedError = new Error(`Databases are not supported in ${dialectName}.`);
+
+describe('QueryGenerator#createDatabaseQuery', () => {
```
WikiRik, line 8: We had an unit test that combined all options, but it's not here. Is that something that we want to add or do we feel it's redundant?

WikiRik, line 8: Snowflake had variants on these tests for when `quoteIdentifiers` is false. We should add those as well right?
Update; I see you do have it in the tests for `QueryGenerator#dropDatabaseQuery`

WikiRik (commented):
Very good PR as always, nice additions as well.

I didn't have time to look into why some dialects do not have native databases support, but I also want to do that soon for my own knowledge.
I also didn't have the time yet to properly take a look at if all new functions have tests with them which I want to do in the next few days.

src/utils/check.ts
```diff
@@ -82,3 +82,34 @@ export function defaultValueSchemable(value: DataType): boolean {
...
+ * For use in per-dialect implementation of methods to warn the user when they use an option that TypeScript declares as valid,
+ * but that the dialect they use does not support.
+ *
+ * @param methodName The name of the method that received the options
+ * @param dialectName The name of the dialect to which the implementation belongs
+ * @param allSupportableOptions All options that this method *can* support. The ones that are declared in TypeScript typings.
+ * @param supportedOptions The subset of options that this dialect *actually does* support.
+ * @param receivedOptions The user provided options that were passed to the method.
+ */
+export function rejectInvalidOptions(
```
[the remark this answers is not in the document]
fzn0x: > This is a great function, thanks for adding!

Agree this is indeed a good addition, this is amazing, I learn a lot from her

test/unit/sql/where.test.ts
```diff
@@ -306,10 +306,6 @@ describe(getTestDialectTeaser('SQL'), () => {
         })),
       }), {
         default: '[yolo].[User].[id] = 1',
-
-        // TODO: mysql, sqlite - this does not sound right.
```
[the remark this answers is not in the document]
ephys: I' conflicted.

Unlike mysql, sqlite doesn't support schemas (at all). If it were 100% up to me, I'd remove the fake schema feature altogether & tell user to prefix their model's `tableName` themselves instead. But the feature is present and I don't have a good enough reason to break existing code.

test/unit/sql/create-schema.test.js
```diff
@@ -1,43 +0,0 @@
...
-      });
-
-      it('9.2.0 or above', () => {
-        current.options.databaseVersion = '9.2.0';
-        expectsql(sql.createSchema('foo'), {
-          postgres: 'CREATE SCHEMA IF NOT EXISTS foo;',
-        });
-      });
-
-      it('below 9.2.0', () => {
```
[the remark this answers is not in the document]
ephys: Correct! It always produces `CREATE SCHEMA IF NOT EXISTS` now

src/dialects/abstract/query-interface.d.ts
```diff
@@ -245,6 +245,10 @@ export interface ColumnsDescription {
   [key: string]: ColumnDescription;
 }
 
+export interface DatabaseDescription {
```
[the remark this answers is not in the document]
ephys: It's possible that it's going to happen. The places where I get the information include all sort of metadata, such as encoding, collate, etc…

Example for postgres: https://www.postgresql.org/docs/current/catalog-pg-database.html
mssql: https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-databases-transact-sql?view=sql-server-ver16

I've limited it to `name` for this PR but now that it's available, I expect we'll eventually receive feature requests to expose more information. Might as well not introduce a breaking change when it happens

test/unit/sql/where.test.ts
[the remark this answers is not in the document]
ephys: Actually we could support schemas in sqlite (through `attach database`) but our codebase would need some adaptation.

src/dialects/abstract/index.ts
```diff
@@ -42,7 +42,10 @@ export type DialectSupports = {
   },
   /* Do we need to say DEFAULT for bulk insert */
   bulkDefault: boolean,
+  /** Whether this dialect has native support for schemas */
   schemas: boolean,
+  /** Whether this dialect has native support for databases (in the postgres or mssql sense) */
```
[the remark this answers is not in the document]
ephys: I think it will be clearer if I rename the property to `multiDatabases`.

They work similarly to schemas. You can think of them as a group of schemas, where schemas are a group of tables.

Some dialects support them, others do not. If I display all my locally running test databases, you can easily see which ones support what:

image png-mh

MariaDB not listed but works the same as MySQL.

Basically only postgres & mssql support having multiple databases, but all of them support schemas.

MySQL decided to make `CREATE DATABASE` and `CREATE SCHEMA` synonymous, but for our purposes, they behave like schemas (a collection of tables). MySQL also calls them schemas in their system tables, even though their documentation refers to them as databases.

Fixing this distinction is also a good step towards fixing https://github.com/sequelize/sequelize/issues/12449

src/dialects/mysql/query-generator.js
ephys: Accidental removal, but will need to not include it in MariaDB

MySQL:

image

MariaDB:

image

src/dialects/mysql/query-generator.js
ephys: It looks like a remnant of the "fix sql injection in replacements" PR that never got removed

test/integration/model.test.js
ephys: Yes but it shouldn't. sqlite still uses fake schemas. I still need at least a few tests that ensure "schema" prefixes the table name (but it won't run other schema tests because they're not schemas)

src/dialects/abstract/index.ts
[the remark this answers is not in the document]
WikiRik: I agree that 'multiDatabases' is a clearer distinction. The image makes it very clear, thanks for that explanation!

test/unit/query-generator/create-database-query.test.ts
ephys: `quoteIdentifiers` is handled by the same quoteIdentifier method every time, so I don't think it's necessary to test all of them. That being said `quoteIdentifiers` is an option I don't understand the purpose of. We need the table names to be quoted when they have characters that could be interpreted differently, such as differences in casing. I don't see the point of allowing to disable it

src/dialects/mysql/query-generator.js
WikiRik: Yeah indeed

test/unit/query-generator/create-database-query.test.ts
WikiRik: That makes sense. Shall we disable that in another PR?

ephys: Yes we'll investigate separately from this one

test/support.ts
ephys: You can throw anything, even primitives
image, hence why the type of caught errors is `unknown` (since TS 4.4 https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables)

The rest of `expectsql` expects errors to be instances of Error, and everything we throw in our codebase should inherit from `Error` (because throwing something else is bad practice), hence why we assert that the thrown error is an instance of Error

src/dialects/abstract/index.ts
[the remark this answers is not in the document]
ephys: Done

src/dialects/db2/query-generator.js
ephys: Done

src/dialects/mysql/query-generator.js
ephys: Done

test/integration/model.test.js
ephys: Re-added

test/unit/query-generator/create-database-query.test.ts
ephys: Done

test/unit/query-generator/create-schema-query.test.ts
ephys: Done

WikiRik (approved):
Looks like all my comments have been resolved. I think we're good to go!

fzn0x (approved):

test/support.ts
fzn0x: > You can throw anything, even primitives image, hence why the type of caught errors is `unknown` (since TS 4.4 https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables)
> 
> The rest of `expectsql` expects errors to be instances of Error, and everything we throw in our codebase should inherit from `Error` (because throwing something else is bad practice), which is why we assert that the thrown error is an instance of Error

Thanks for the explanation, lets merge this~!!

PR #15016, opened 2022-09-19
fix: Add support for Oracle DB 18c CI

nilabja-bhattacharya:
## Pull Request Checklist

## Description Of Change

PR to support the CI of the minimal supported Oracle DB version

## Todos

- [ ] 
- [ ] 
- [ ]


sdepold (approved):

sdepold:
Hm. Seems the failing tests are consistent across the both oracle versions?

nilabja-bhattacharya:
Yes, the tests that failed in the CI are flaky; they should be fixed with a re-run.

dev/oracle/18-slim/privileges.sql
```diff
@@ -0,0 +1,27 @@
+-- Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved
+
+create user sequelizetest identified by sequelizepassword;
```
WikiRik, line 3: Are these different from 21? Maybe we can reference the 21 version of this file instead of copying this

WikiRik: This also holds for the other files, try to minimize the copying of files. We want to do that for the other dialects as well later, so this is a good first step

src/dialects/oracle/index.js
```diff
@@ -54,7 +54,7 @@ OracleDialect.prototype.supports = _.merge(_.cloneDeep(AbstractDialect.prototype
   GEOMETRY: false
 });
 
-OracleDialect.prototype.defaultVersion = '18.4.0';
+OracleDialect.prototype.defaultVersion = '18.0.0';
```
WikiRik, line 57: With this, maybe we should test on the first version of 18 then? So the oldest minor version we support and the latest minor version we support.

WikiRik (commented):
I have a few comments but they are non blocking. Would be nice if you could work on these

src/dialects/oracle/index.js
nilabja-bhattacharya: The oldest minor version for Oracle DB 18 is 18.0.0 and we are testing against this version of the DB.  
The latest major version for Oracle DB is 21.3.0 and we're also testing against it.

dev/oracle/18-slim/privileges.sql
nilabja-bhattacharya: I have added a change does this look good?

nilabja-bhattacharya:
There has been a flaky failure in 2 DBs, it would be good to re-run to failed jobs. Thanks.

src/dialects/oracle/index.js
WikiRik: Thanks!

dev/oracle/18-slim/privileges.sql
WikiRik: Yes, that's great. Thanks!

package.json
```diff
@@ -245,13 +245,17 @@
     "start-mssql": "bash dev/mssql/2019/start.sh",
     "start-db2": "bash dev/db2/11.5/start.sh",
     "start-oracle": "bash dev/oracle/21-slim/start.sh",
```
WikiRik, line 247: Maybe we should remove this so it's clearer which version we are using. Also holds for the stop script

WikiRik (commented):
One more comment, but I think after that we should be good to merge

package.json
nilabja-bhattacharya: I left the start-oracle command as is because the term would-be uniform across all the dialects. Also, a user can call the `start-oracle-<DB version>` to set up a version-specific DB.

WikiRik: Yeah, on the `main` branch we have changed that already and might do so as well for `v6`. Check out https://github.com/sequelize/sequelize/pull/14737
So to be ahead of that change I would prefer to only have the Oracle DB version specific scripts defined

nilabja-bhattacharya: Okay thanks, should I rename these to start-oracle-oldest, start-oracle-latest, etc or start-oracle-<DB version> looks good!

WikiRik: If you could update to -oldest/-latest that would be great, thanks

nilabja-bhattacharya: Sure, I would update it.

nilabja-bhattacharya: Updated the package.json. Thanks.

WikiRik (approved):
This looks good to me. I do have some points for additional refactor but I'll take those up as part of backporting CI changes that we've done on `main` to `v6`.
Thanks for your work on this!

PR #15057, opened 2022-09-27
docs: improve first impression of Sequelize README

fzn0x:
Add features and quick start section


ephys:
> I think we should guide people to our website as soon as possible

I agree, I think the sooner they land on the website the better. There is too much information for a readme. The readme should remain minimal

Another reason why all of this was removed from the readme during the migration to the new website is that a lot of the information is specific to either v6 or v7. We don't know which version the user uses

fzn0x:
I think your reviews was right, this all about the versions and keeping the readme minimal, if that's the reason maybe I agree to close this PR.

Issue #9956, opened 2018-09-25
Add support for PG `?&` operator


acodercat:
## What are you doing?

client_ids is json field
["0a5dd8ef-380c-48df-94a2-801fd1256646", "123"]

I want select Contain "123" all rows.

```js
// code here
```

_Output, either JSON or SQL_

 SELECT "id", "client_ids" FROM "account" AS "account" WHERE client_ids ?& array[ '123'];

__Dialect: postgres
__Dialect version:last
__Database version:last
__Sequelize version:last
__Tested with latest release:__ No (If yes, specify that version)

**Note :** _Your issue may be ignored OR closed by maintainers if it's not tested against latest version OR does not follow issue template._

acodercat:
please！！！I need

sushantdhiman:
For now you can do this by something like this

```js
Model.findAll({
   where: {
      [Op.and]: [Sequelize.literal("client_ids ?& $ids")]
   },
   bind: {
      ids: ['123', '234', ....]
   }
})
```

PR #15073, opened 2022-10-01
feat: add new postgres JSONB operators `?|` and `?&`

### LINKED ISSUE #14844: Support for JSONB column operators (`?|`, `?&`, `?`)
## Issue Creation Checklist

## Feature Description
Support For JSONB Operators

### Describe the feature you'd like to see implemented
I have a use case with JSONB column in postgres. We use operator ?| to check if the jsonb column has any key from an array with text values.

PostgreSQL documentation: https://www.postgresql.org/docs/14/functions-json.html

Example query:
```sql
SELECT id, group_ids FROM contact WHERE group_ids ?| array['3367', '3368'];
```

- [ ] Yes. This feature only applies to the following dialect(s): 

- [x] Yes, I have the time but I will need guidance.
- [ ] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in implementing my feature.

---

*Indicate your interest in the addition of this feature by adding the 👍 reaction. Comments such as "+1" will be removed.*

fzn0x:
## Pull Request Checklist

## Description Of Change

Separated PR for `?` #14844
Closes https://github.com/sequelize/sequelize/issues/9956

Add new operators for

<html>
<body>

Operator | Right Operand Type | Description | Example
-- | -- | -- | --
?\| | text[] | Do any of these array strings exist as top-level keys? | '{"a":1, "b":2, "c":3}'::jsonb ?\| array['b', 'c']
?& | text[] | Do all of these array strings exist as top-level keys? | '["a", "b"]'::jsonb ?& array['a', 'b']

</body>
</html>

## Todos

- [x] Add types
- [x] Add tests
- [x] Add code documentation 

## Follow Up Tasks

- [ ] Add v6 backport
- [ ] Document to sequelize/website


fzn0x:
@ephys @WikiRik Did we are all good with the operators name or do you guys have any suggestion for the operators naming?

Currently the naming is:

```js
    [Op.hasKey]: '?',
    [Op.anyKeys]: '?|',
    [Op.allKeys]: '?&',
    [Op.concatenate]: '||',
```

ephys:
I would rename anyKeys & allKeys to anyKeyExists, allKeysExist

I also propose to not include concatenate in this PR. Instead let's look into providing a concatenate function that also works for strings. I think it'll be clearer than an operator:

```
User.update({
  myCol: concatenate('string', col('myCol'))
});
```

fzn0x:
Sure, will deprecate both

WikiRik:
I haven't given it much thought but maybe we should move the operators into their own section in DialectSupports? Not all, but just the ones that are only supported by some dialects, like REGEXP. With a little bit of further refactor we should be define OperatorMap completely in the abstract dialect instead of adding additional operators in the dialect's implementation.

Since this could mean that we migrate `abstract/query-generator/operators` to TypeScript as well this can also be done in a future PR. I just want to know your opinions in this

fzn0x:
> I haven't given it much thought but maybe we should move the operators into their own section in DialectSupports? Not all, but just the ones that are only supported by some dialects, like REGEXP. With a little bit of further refactor we should be define OperatorMap completely in the abstract dialect instead of adding additional operators in the dialect's implementation.
> 
> Since this could mean that we migrate `abstract/query-generator/operators` to TypeScript as well this can also be done in a future PR. I just want to know your opinions in this

It is really good, but I'm not sure if I can getting this done in this PR, but for a new PR it's fine.

 

for example like this `Op.postgres.anyKeyExists` or still `Op.anyKeyExists` but activated by default by the dialect choosen in the `SequelizeOptions`. Your thoughts? @ephys

fzn0x:
So even if we add an Op with the same name in the different dialects `query-generator`, we still can create a different `Op.[dialectBasedOpName]` implementation based on the dialect that is being used in the connection options.

WikiRik:
> So even if we add an Op with the same name in the different dialects `query-generator`, we still can create a different `Op.[dialectBasedOpName]` implementation based on the dialect that is being used in the connection options.

I would not do that. If an operator gives a different result based on the dialect it should also have a different name imo. Otherwise that would be more confusing and doesn't help with documentation.
It would also cause more code duplication for things like `Op.mysql.regexp` and `Op.mariadb.regexp` (or postgres and yugabyte or additional future dialects based on postgres)

fzn0x:
> > So even if we add an Op with the same name in the different dialects `query-generator`, we still can create a different `Op.[dialectBasedOpName]` implementation based on the dialect that is being used in the connection options.
> 
> I would not do that. If an operator gives a different result based on the dialect it should also have a different name imo. Otherwise that would be more confusing and doesn't help with documentation. It would also cause more code duplication for things like `Op.mysql.regexp` and `Op.mariadb.regexp` (or postgres and yugabyte or additional future dialects based on postgres)

That's makes sense, I have try pushing the integration tests, hope it pass

fzn0x:
> > So even if we add an Op with the same name in the different dialects `query-generator`, we still can create a different `Op.[dialectBasedOpName]` implementation based on the dialect that is being used in the connection options.
> 
> I would not do that. If an operator gives a different result based on the dialect it should also have a different name imo. Otherwise that would be more confusing and doesn't help with documentation. It would also cause more code duplication for things like `Op.mysql.regexp` and `Op.mariadb.regexp` (or postgres and yugabyte or additional future dialects based on postgres)

That's makes sense.

By the way, I have try to pushing the integration tests for this PR, hope it's pass.

fzn0x:
Need extra work for the postgres integration test, I will start tomorrow

fzn0x:
I think the Op didn't recognized because the operator imported from `@sequelize/core` do we need to separate the PR for integration test or just skip it? https://github.com/sequelize/sequelize/blob/ed730154689c1ad1028825ef8e658941af5cb294/test/integration/operators.test.js#L7

cc @ephys @WikiRik

fzn0x:
I think the Op didn't recognized because the operators is imported from `@sequelize/core` do we need to separate the PR for integration test or just skip it? https://github.com/sequelize/sequelize/blob/ed730154689c1ad1028825ef8e658941af5cb294/test/integration/operators.test.js#L7

cc @ephys @WikiRik

fzn0x:
If you guys have any options to fix the issue, kindly share it, because currently we can't test non-existing operators from `@sequelize/core`

fzn0x:
If you guys have any suggestions to fix the issue, kindly share it, because currently we can't test non-existing operators from `@sequelize/core`

ephys:
`@sequelize/core` points to the lib folder in tests. It's not the npm version. I think you have a different issue

ephys:
You haven't declared its typings in `WhereOptions`, there is a type used by `WhereOptions` that defines how each operator is used

I don't think an integration test is necessary here. Or if you add one, add a single one. The rest should go in unit tests. There is a `where.test.ts` file which tests the SQL & typing of every operator

fzn0x:
> You haven't declared its typings in `WhereOptions`, there is a type used by `WhereOptions` that defines how each operator is used
> 
> I don't think an integration test is necessary here. Or if you add one, add a single one. The rest should go in unit tests. There is a `where.test.ts` file which tests the SQL & typing of every operator

Thanks! very helping me to learn about the operators core, I'm still learning about the operators flow, I'm not sure I already understand yet but I will try.

fzn0x:
TODO: support using `Sequelize.literal` , will test with `jsonb_array_elements_text`.

fzn0x:
Woohoooo finally passing the tests!

fzn0x:
Guess what, I'm learning new things everyday.

To refine my confidence I will create the JSONB integration test in `test/integration/json.test.js`.

fzn0x:
PR is ready for next review!

src/dialects/abstract/query-generator.js
```diff
@@ -2861,6 +2861,22 @@ Only named replacements (:name) are allowed in literal() because we cannot guara
 
         return this._joinKeyValue(key, this.escape(pattern, undefined, options), comparator, options.prefix);
       }
+
+      case Op.anyKeyExists: {
```
WikiRik, line 2865: These two cases are the same right? We should combine these then to not have duplicated code

src/model.d.ts
```diff
@@ -570,6 +570,38 @@ export interface WhereOperators<AttributeType = any> {
...
+   * Check if any of these array strings exist as top-level keys.
+   *
+   * @example
+   * ```typescript
+   * { jsonbAttribute: { [Op.anyKeyExists]: ['a','b'] } }
+   * // results in
+   * // "jsonbAttribute" ?| ARRAY['a','b']
+   * ```
+   *
+   * https://www.postgresql.org/docs/9.5/functions-json.html
```
ephys, line 586:
```suggestion
   * https://www.postgresql.org/docs/current/functions-json.html
```

Or at least a more recent version than 9.5, we don't support anything < v10

src/model.d.ts
```diff
@@ -570,6 +570,38 @@ export interface WhereOperators<AttributeType = any> {
...
+   * Check if all of these array strings exist as top-level keys.
+   *
+   * @example
+   * ```typescript
+   * { jsonbAttribute: { [Op.allKeysExist]: ['a','b'] } }
+   * // results in
+   * // "jsonbAttribute" ?& ARRAY['a','b']
+   * ```
+   *
+   * https://www.postgresql.org/docs/9.5/functions-json.html
```
ephys, line 602:
```suggestion
   * https://www.postgresql.org/docs/current/functions-json.html
```

src/model.d.ts
```diff
@@ -570,6 +570,38 @@ export interface WhereOperators<AttributeType = any> {
...
+   * @example
+   * ```typescript
+   * { jsonbAttribute: { [Op.anyKeyExists]: ['a','b'] } }
+   * // results in
+   * // "jsonbAttribute" ?| ARRAY['a','b']
+   * ```
+   *
+   * https://www.postgresql.org/docs/9.5/functions-json.html
+   */
+  [Op.anyKeyExists]?: WhereOperators<StaticValues<NonNullable<string>>> | Literal;
```
ephys, line 588: `WhereOperators` refers to this type. The reason some types use `WhereOperators` is to re-use the same type as another operator.

The following makes `iRegexp` use the same type as `regexp`

```ts
[Op.iRegexp]?: WhereOperators<AttributeType>[typeof Op.regexp];
```

And `regexp` has the proper type definition:

```ts
[Op.regexp]?: AllowAnyAll<OperatorValues<Extract<AttributeType, string>>>;
```

When crafting the type you need to think about the following:

- are dynamic values allowed here (e.g. column references, functions, etc)
- is Op.any & Op.all usable in combination with this?

test/unit/sql/where.test.ts
```diff
@@ -2246,6 +2246,39 @@ describe(getTestDialectTeaser('SQL'), () => {
 
     if (sequelize.dialect.supports.JSONB) {
       describe('JSONB', () => {
+        if (dialectSupportsArray()) {
```
ephys, line 2249: this check is not necessary. Only postgres supports JSONB & it supports arrays.
If a dialect starts supporting these operators but they work with something else than arrays, we'll adapt the test suite

test/unit/sql/where.test.ts
```diff
@@ -2246,6 +2246,39 @@ describe(getTestDialectTeaser('SQL'), () => {
...
+            default: `[jsonbAttr] ?& ARRAY['a', 'b']`,
+          });
+
+          testSql({
+            jsonbAttr: {
+              [Op.anyKeyExists]: literal(`SELECT jsonb_array_elements_text('ARRAY["a","b"]')`),
+            },
+          }, {
+            default: `[jsonbAttr] ?| ARRAY(SELECT jsonb_array_elements_text('ARRAY["a","b"]'))`,
+          });
```
ephys, line 2272: When using `literal`, we don't assume what's in the literal. It's the responsibility of the user to use a valid literal.

```suggestion
          testSql({
            jsonbAttr: {
              [Op.anyKeyExists]: literal(`ARRAY(SELECT jsonb_array_elements_text('ARRAY["a","b"]'))`),
            },
          }, {
            default: `[jsonbAttr] ?| ARRAY(SELECT jsonb_array_elements_text('ARRAY["a","b"]'))`,
          });
```

test/unit/sql/where.test.ts
```diff
@@ -2246,6 +2246,39 @@ describe(getTestDialectTeaser('SQL'), () => {
...
+
+          testSql({
+            jsonbAttr: {
+              [Op.allKeysExist]: literal(`SELECT jsonb_array_elements_text('ARRAY["a","b"]')`),
```
ephys, line 2276: Same here

```suggestion
              [Op.allKeysExist]: literal(`ARRAY(SELECT jsonb_array_elements_text('ARRAY["a","b"]'))`),
```

test/unit/sql/where.test.ts
```diff
@@ -2246,6 +2246,39 @@ describe(getTestDialectTeaser('SQL'), () => {
...
+            },
+          }, {
+            default: `[jsonbAttr] ?& ARRAY(SELECT jsonb_array_elements_text('ARRAY["a","b"]'))`,
+          });
```
ephys, line 2280: Don't forget to test the following:

```typescript
// literal inside of an array
{ [Op.allKeysExist]: [literal()] }

// column reference
{ [Op.allKeysExist]: col('keys') }

// cast
{ [Op.allKeysExist]: cast(col('keys'), 'STRING[]') }

// function
{ [Op.allKeysExist]: fn('get_keys') }

// any & all -> check if valid in postgres first and add a test that rejects or succeeds depending
// context on any & all: https://www.postgresql.org/docs/current/functions-comparisons.html#id-1.5.8.30.16

// Op.any & Op.all + Op.values -> same as above
```

ephys (changes requested):

test/unit/sql/where.test.ts
fzn0x: Since we will wrap the literal within an array, I think we will skip this commit suggestion https://github.com/sequelize/sequelize/pull/15073#discussion_r990766377

test/unit/sql/where.test.ts
fzn0x: Since we will wrap the literal within an array, I think we will skip this commit suggestion https://github.com/sequelize/sequelize/pull/15073#discussion_r990766377

test/unit/sql/where.test.ts
fzn0x: Thanks

test/unit/sql/where.test.ts
ephys: You'll need to test both `[literal()]` and `literal`, both are valid

src/dialects/abstract/query-generator.js
fzn0x: Thanks

test/unit/sql/where.test.ts
fzn0x: I see

fzn0x:
Will continue working on the review midnight, I need to take a few hours for my free time with my family 🙏

test/unit/sql/where.test.ts
fzn0x: Thanks

test/unit/sql/where.test.ts
fzn0x: Thanks for the reminder

src/dialects/abstract/query-generator.js
```diff
@@ -2861,6 +2861,22 @@ Only named replacements (:name) are allowed in literal() because we cannot guara
...
+        if (value instanceof Utils.Literal) {
+          return this._joinKeyValue(key, `ARRAY(${value.val})`, comparator, options.prefix);
```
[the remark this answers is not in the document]
fzn0x: Thanks for the amazing review guys

src/dialects/abstract/query-generator.js
```diff
@@ -2861,6 +2861,15 @@ Only named replacements (:name) are allowed in literal() because we cannot guara
...
         return this._joinKeyValue(key, this.escape(pattern, undefined, options), comparator, options.prefix);
       }
+
+      case Op.anyKeyExists:
+      case Op.allKeysExist: {
+        if (value instanceof Utils.SequelizeMethod) {
+          return this._joinKeyValue(key, this.handleSequelizeMethod(value, undefined, undefined, options), comparator, options.prefix);
+        }
+
+        return this._joinKeyValue(key, `ARRAY[${value.map(item => this.escape(item, undefined, options)).join(', ')}]`, comparator, options.prefix);
```
ephys, line 2871: I just thought of a case we missed!

If `value` is empty, this will produce `jsonb ?| ARRAY[]`. Postgres won't be able to infer the type of the array. Luckily both `?|` and `?&` require a text array. If `value` is empty, could you make it produce `jsonb ?| ARRAY[]::text[]`? (don't forget to add a test for it). No need to add the cast the rest of the time

src/model.d.ts
```diff
@@ -570,6 +570,39 @@ export interface WhereOperators<AttributeType = any> {
...
+   * @example
+   * ```typescript
+   * { jsonbAttribute: { [Op.anyKeyExists]: ['a','b'] } }
+   * // results in
+   * // "jsonbAttribute" ?| ARRAY['a','b']
+   * ```
+   *
+   * https://www.postgresql.org/docs/current/functions-json.html
+   */
+  [Op.anyKeyExists]?: Array<NonNullable<string>>
```
ephys, line 588: string is already non-nullable

```suggestion
  [Op.anyKeyExists]?: Array<string>
```

src/model.d.ts
```diff
@@ -570,6 +570,39 @@ export interface WhereOperators<AttributeType = any> {
...
+    | DynamicValues<NonNullable<AttributeType>>;
```
ephys, line 589: This operator only accepts strings, so it would be this instead (all it does is exclude `where` since this isn't a boolean):

```suggestion
    | DynamicValues<string>;
```

test/integration/json.test.js
```diff
@@ -323,5 +323,99 @@ describe('model', () => {
...
+          emergency_contact: {
+            [Op.allKeysExist]: ['name', 'programmer', 'grade'],
+          },
+        },
+      });
+
+      expect(users.length).to.equal(1);
+    });
+
+    it('should be able to check any of these array strings exist as top-level keys (with literal)', async function () {
```
ephys, line 383: this test is already covered by our unit tests, can be removed

ephys (changes requested):
This is good work! There is one scenario we forgot but once that's handled, this will be good to go

fzn0x:
I just can say you are a beast and beautiful developer at the same time @ephys I learned a lot from you!

WikiRik:
What's the state of this PR, just [this comment](https://github.com/sequelize/sequelize/pull/15073#discussion_r1014663322) to implement and then it's ready for review right? Might be good to not merge `main` in here if there's no actual updates to the PR, since I (and probably others) will get a notification for that

fzn0x:
> What's the state of this PR, just [this comment](https://github.com/sequelize/sequelize/pull/15073#discussion_r1014663322) to implement and then it's ready for review right? Might be good to not merge `main` in here if there's no actual updates to the PR, since I (and probably others) will get a notification for that

The last merge only to test for the Utils export and it's fail as expected. 

Will merge once again to start working on the last review.

fzn0x:
This weekend I will settle all the drafted PRs

ephys:
Issue about deprecating `?` as a replacement so we can use as an operator: https://github.com/sequelize/sequelize/issues/15082

src/model.d.ts
```diff
@@ -575,6 +575,39 @@ export interface WhereOperators<AttributeType = any> {
...
+   * ```typescript
+   * { jsonbAttribute: { [Op.anyKeyExists]: ['a','b'] } }
+   * // results in
+   * // "jsonbAttribute" ?| ARRAY['a','b']
+   * ```
+   *
+   * https://www.postgresql.org/docs/current/functions-json.html
+   */
+  [Op.anyKeyExists]?: string[]
+    | DynamicValues<string>;
```
ephys, line 594:
```suggestion
    | DynamicValues<string[]>;
```

ephys (approved):

ephys (approved):
Good work :)

fzn0x:
Thank you! @WikiRik do you have inputs on this one? if all good, can you merge this one?

WikiRik (approved):
Looks good! Nice work

PR #15074, opened 2022-10-01
meta: Use crypto module instead of uuid

AlexHladin:
## Pull Request Checklist

## Description Of Change

This PR avoids using the uuid package and substitutes it with [crypto module](https://nodejs.org/docs/latest-v14.x/api/crypto.html) that is available in the Node.js package.
Suddenly the [dialect.ts](https://github.com/sequelize/sequelize/blob/main/src/utils/dialect.ts#L30) still uses uuid v1. This code blocks fully removing the uuid package from package.json.


ephys:
I'm very pro using native solutions instead of libraries, unfortunately Sequelize 7 still needs to support Node 14 and `crypto.randomUUID` is only available starting with node 16.7.0.

We will be able to accept this PR in the major that is released after 30 Apr 2023

AlexHladin:
> I'm very pro using native solutions instead of libraries, unfortunately Sequelize 7 still needs to support Node 14 and `crypto.randomUUID` is only available starting with node 16.7.0.
> 
> We will be able to accept this PR in the major that is released after 30 Apr 2023

@ephys `crypto.randomUUID` is available in Node 14. Take a look at [documentation](https://nodejs.org/docs/latest-v14.x/api/crypto.html#crypto_crypto_randomuuid_options)

WikiRik:
Indeed added in Node 14.17.0 which is the lowest version we plan to support for v7 so this should be fine

WikiRik:
It was indeed added in Node 14.17.0 which is the lowest version we support. If Node 16.0.0 has it as well we can use this

ephys:
MDN lied to me 😱 very nice then!

ephys (approved):

PR #15075, opened 2022-10-01
meta: move logWarnings method to AbstractQuery class

AlexHladin:
## Pull Request Checklist

## Description Of Change

This PR reduces the amount of duplicated code by moving the `logWarnings` function to the AbstractQuery class.

### DIFF
--- a/src/dialects/abstract/query.js
+++ b/src/dialects/abstract/query.js
@@ -35,7 +35,7 @@ export class AbstractQuery {
 
   async logWarnings(results) {
     const warningResults = await this.run('SHOW WARNINGS');
-    const warningMessage = `${_.capitalize(this.sequelize.dialect.name)} Warnings (${this.connection.uuid || 'default'}): `;
+    const warningMessage = `${this.sequelize.dialect.name} Warnings (${this.connection.uuid || 'default'}): `;
     const messages = [];
     for (const _warningRow of warningResults) {
       if (_warningRow === undefined || typeof _warningRow[Symbol.iterator] !== 'function') {
--- a/src/dialects/abstract/query.js
+++ b/src/dialects/abstract/query.js
@@ -35,7 +35,7 @@ export class AbstractQuery {
 
   async logWarnings(results) {
     const warningResults = await this.run('SHOW WARNINGS');
-    const warningMessage = `${this.options.name} Warnings (${this.connection.uuid || 'default'}): `;
+    const warningMessage = `${_.capitalize(this.sequelize.dialect.name)} Warnings (${this.connection.uuid || 'default'}): `;
     const messages = [];
     for (const _warningRow of warningResults) {
       if (_warningRow === undefined || typeof _warningRow[Symbol.iterator] !== 'function') {
--- a/src/dialects/abstract/query.js
+++ b/src/dialects/abstract/query.js
@@ -33,6 +33,31 @@ export class AbstractQuery {
     }
   }
 
+  async logWarnings(results) {
+    const warningResults = await this.run('SHOW WARNINGS');
+    const warningMessage = `${this.options.name} Warnings (${this.connection.uuid || 'default'}): `;
+    const messages = [];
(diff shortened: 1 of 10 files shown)


src/dialects/postgres/index.js
```diff
@@ -63,6 +63,9 @@ export class PostgresDialect extends AbstractDialect {
     this.queryGenerator = new PostgresQueryGenerator({
       dialect: this,
       sequelize,
+      options: {
```
ephys, line 66: QueryGenerator isn't responsible for the creation of Query objects, I don't think this option bag will reach it

The query object is created by `Sequelize#queryRaw`:

```
```

IMO the `logWarnings` could get the dialect name from `this.sequelize.dialect.name`. It would be fully lowercase but I think that's fine, that's already what our other messages do.

AlexHladin: @ephys `lodash.capitalize` could improve `this.sequelize.dialect.name` :)

WikiRik: I think we should consider styling the dialect name directly so `capitalize` is not needed. So we could use both `MariaDB`/`MySQL` and `mariadb`/`mysql`. The first one for displaying it to users, the second one internally

WikiRik (commented):
Does `SHOW WARNINGS` work for each dialect? If not, maybe we can define the supporting dialects in `DialectSupports`? And can we add integration tests for each supported dialect?

src/dialects/postgres/index.js
macluiggy: hola saludos desde ecuador, usamos sequelize en nuestra empresa :)

AlexHladin:
> Does `SHOW WARNINGS` work for each dialect? If not, maybe we can define the supporting dialects in `DialectSupports`? And can we add integration tests for each supported dialect?

@WikiRik the `logWarnings` method uses only in refactored classes, I am not sure that it makes sense to cover it with tests and define it in the supporting dialects.

src/dialects/abstract/query.js
```diff
@@ -33,6 +33,31 @@ export class AbstractQuery {
     }
   }
 
+  async logWarnings(results) {
+    const warningResults = await this.run('SHOW WARNINGS');
+    const warningMessage = `${this.sequelize.dialect.name} Warnings (${this.connection.uuid || 'default'}): `;
```
WikiRik, line 38:
```suggestion
    const warningMessage = `${this.sequelize.dialect.name} warnings (${this.connection.uuid || 'default'}): `;
```

test/integration/dialects/mysql/warning.test.js
```diff
@@ -35,7 +35,7 @@ describe(Support.getTestDialectTeaser('Warning'), () => {
         });
 
         // last log is warning message
-        expect(logger.args[logger.args.length - 1][0]).to.be.match(/^MySQL Warnings \(default\):.*/m);
+        expect(logger.args[logger.args.length - 1][0]).to.be.match(/^mysql Warnings \(default\):.*/m);
```
WikiRik, line 38:
```suggestion
        expect(logger.args[logger.args.length - 1][0]).to.be.match(/^mysql warnings \(default\):.*/m);
```

WikiRik (dismissed):
@AlexHladin ah yes, `logWarnings` is not a function to be used by users themselves. This is indeed fine for now. I have one small styling suggestion though, but it's a non-blocking issue to me.

Thanks for this PR!

src/dialects/postgres/index.js
WikiRik: We'll add a stylized name of dialect in a future refactor. For now the lowercase one is fine

WikiRik (dismissed):

WikiRik:
You can just merge the new commits from `main` in without rebasing so approvals are kept in place. We're squashing when merging so there is no difference in the end

WikiRik (approved):

PR #15093, opened 2022-10-06
fix: inject scope in findAndCountAll

### LINKED ISSUE #15067: Bad count query on distinct findAndCountAll when using a scope with an include
## Issue Creation Checklist

## Bug Description

I have two tables with a one-to-many relationship, let's call them Parent and Child, with Child belonging to one Parent, and I want to call `findAndCountAll` on my `Child` table with the included `Parent` model by association, and ordering results by parent then child.

When doing a regular `Child.findAndCountAll({ include, order, distinct: true })`, both `count` and `find` works (although the order is merged with default scope, but this is unrelated to my issue).

However, if I instead use a `scope` to call `Child.scope('includeParent').findAndCountAll({ distinct: true })` then the `count` part generates a wrong query with ambiguous `count(DISTINCT("id"))`, instead of the correct `count(DISTINCT("child"."id"))`. The rest of the `count` query is correct apart from the column name, and the `filter` query is correct as well.

### Reproducible Example

SSCCE example here: https://github.com/sequelize/sequelize-sscce/pull/238

When calling `Child.scope('includeParent').findAndCountAll({ distinct: true })`

I expect the `findAndCountAll` to generate a correct distinct `count` query with valid [tableName].[primaryKeyColumn] reference. (in this case: `count(DISTINCT("child"."id))`)

I noticed the issue only happens when using include in a `scope`: if the include is directly in the `findAndCountAll` arguments, correct reference `count(DISTINCT("child"."id))` is generated.

The `count` part of `findAndCountAll` incorrectly generates the distinct reference: filter from the scope is correctly used, but the reference is ambiguous `count(DISTINCT("id"))`, which raises an error because multiple tables with "id" column are used in the query.

```
SELECT count(DISTINCT("id")) AS "count" FROM "children" AS "child" LEFT OUTER JOIN "parents" AS "parent" ON "child"."parentId" = "parent"."id";
---> ambiguous column name: id
```

### Environment

- Sequelize version: 6.21.3 and 6.23.2
- Node.js version: 14.17.0
- Database & Version: PostgreSQL 12 (although all databases were affected in the SSCCE tests)
- Connector library & Version: pg

- [x] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in resolving my issue.

---

frostzt:
## Pull Request Checklist

## Description Of Change

`findAndCountAll` function if called with `scope` generates an ambiguous query clause. The issue was reported here #15067 and an reproduction was created here https://github.com/sequelize/sequelize-sscce/pull/238.

In the `findAndCountAll` we now call the `this._injectScope(options)` to inject the options provided by the scope for the query.

Closes #15067 

## Changes achieved

Previously if used with scope this was the query being generated:

```sql
SELECT count(DISTINCT("id")) AS "count" FROM "children" AS "child" LEFT OUTER JOIN "parents" AS "parent" ON "child"."parentId" = "parent"."id";
---> ambiguous column name: id
```

Query generated with the update:

```sql
SELECT count(DISTINCT("child"."id")) AS "count" FROM "children" AS "child" LEFT OUTER JOIN "parents" AS "parent" ON "child"."parentId" = "parent"."id";
---> ambiguous column name: id
```


evanrittenhouse (dismissed):

evanrittenhouse:
@frostzt Thanks for submitting this PR!

laurelgr:
I think you can actually check `Have you added new tests to prevent regressions?` by the way :)

frostzt:
@WikiRik are there some issues with the tests?

fzn0x:
I believe there' should be no issue since this changes only contains addition. AFAIK side effects from injectScope addition is okay.

fzn0x:
Please wait for the reviews, we are have a lot is going on in our plates. :))

WikiRik:
@frostzt I'm sorry, I'm not actually sure what happened with removing the review of @evanrittenhouse but it was on accident.
On initial look your change and tests look fine, I'll review more later.

fzn0x:
I'd undo my reviews because core maintainer already reach out your PR.

evanrittenhouse:
I'll re-approve - @ephys only change was moving `_injectScope` to run after `clone()`. The tests LGTM.

evanrittenhouse (dismissed):

src/model.js
```diff
@@ -2139,6 +2139,8 @@ Specify a different name for either index to resolve this issue.`);
 
     const countOptions = Utils.cloneDeep(options);
 
+    this._injectScope(options);
```
ephys, line 2142: This still mutates the `options` parameter, but more importantly, it doesn't seem like this solves the bug: The bug was related to the `count` query, but these options don't receive the scope anymore. Which indicates to me that the new test would pass even if we removed this line

Something else also concerns me: `findAll` already injects the scope. Injecting it here means it's injected twice for `findAll`, which will likely introduce bugs

ephys (changes requested):

src/model.js
laurelgr: Yeah, I guess you got distracted once working on it but I had indeed reported that the `findAll` part of the `findAndCountAll` was injected correctly, only the `count` was bad. :)

At a glance, I suppose the meaning was that you should make a clone of the `options` and inject on this clone, instead of literally mutating/modifying the existing, original options, which could have side effects in other places and make debugging harder.

Regular `count` with scope works correctly I believe, maybe you can have a look at how this part does the scope injection as guideline?

frostzt: @laurelgr rip dumb me, I didn't got that (that's why you should always read properly first) lemme get onto this!

src/model.js
```diff
@@ -2138,6 +2138,7 @@ Specify a different name for either index to resolve this issue.`);
     }
 
     const countOptions = Utils.cloneDeep(options);
+    this._injectScope(countOptions);
```
WikiRik, line 2141: If I remove this line and run the new test on sqlite it still passes. So I'm not sure if this fixes the issue or that the test is incorrect. Can you take a look at it?

frostzt: Oh lemme look into that, I do think that the tests are wrong since. Lemme see if it just that or am I messing something up.

frostzt: Hey @WikiRik seems like the tests are fine but I think what I lack in them is to produce the wanted case, tbh I am not able to come up with an example of the case. Here's why!

So the issue was that the query that was being generated was ambiguous, the example is the following:

```js
Player.scope('includeTeam').findAndCountAll({ distinct: true });
```

The query generated for the above is the following:

```sql
SELECT count(DISTINCT(`id`)) AS `count` FROM `Players` AS `Player` LEFT OUTER JOIN `Teams` AS `Team` ON `Player`.`TeamTeamId` = `Team`.`team_id`;
```

Notice above that the query contains ``DISTINCT(`id`)``, here `id` is ambiguous in the sense that if I were NOT to assign explicit primary key, the autogenerated field `id` can be of either of `Team` or `Player` table.

However post the update the query generated is the following!

```sql
SELECT count(DISTINCT(`Player`.`id`)) AS `count` FROM `Players` AS `Player` LEFT OUTER JOIN `Teams` AS `Team` ON `Player`.`TeamId` = `Team`.`id`;
```

Screenshot 2022-11-11 at 2 59 44 PM

This kinda says that the tests are kinda redundant but the query generated is specific. So I am kinda stuck with creating the exact case. Any feedback would be good!

@ephys sorry for the ping but what's your take on the above!?

frostzt:
@ephys can you check this too!

WikiRik:
@frostzt could you solve the merge conflicts? Then we can take another look at this one

frostzt:
Hey @WikiRik merged with main! I am really sorry for the time off got really busy with stuff! Will get back soon!

Issue #14402, opened 2022-04-16
Forbid using instance `save`, `update`, `reload`, `destroy`, `restore`, `decrement` and `increment` if the model has no PK


ephys:
Initial comment: https://github.com/sequelize/sequelize/pull/14173#issuecomment-1065388719

After discussing this issue internally, we reached the decision that instance methods that need to uniquely identify an entity in a stable way will throw if the model had no Primary Key.

This will make using models without Primary Keys a worse experience, but also a *much* safer one.

It will still be possible to use all static methods, of course.

frostzt:
Hey @ephys I was thinking of picking this up but I think it's already done? https://github.com/sequelize/sequelize/blob/main/src/model.js#L4009

frostzt:
You're awesome @ephys I will try my best and have a PR ready for this, and I'll be sure to get back to ask for help if I need any further guidance! Thanks!

frostzt:
Hey @ephys I believe the `update()` is already handled for this case. Since to create a model we would need to save it? Or my thinking is wrong? I tried to make a simple unit test for the `update()` method on instance and it throws since the save method is invoked first automatically. Or am I missing something, a bit of explanation would be awesome!

PR #15108, opened 2022-10-08
feat: restrict instance methods if no primary key

frostzt:
## Pull Request Checklist

## Description Of Change

Instance methods such as `save`, `update`, and others will now throw an error if no `primaryKey` was passed by the user.

Closes: #14402 
Closes https://github.com/sequelize/sequelize/issues/15113

## Todo

- Integration tests for `save` and `update` are failing.


src/model.js
```diff
@@ -3519,8 +3515,11 @@ Instead of specifying a Model, either:
       return result;
     }, {});
 
-    if (_.size(where) === 0) {
-      return this.constructor.options.whereCollection;
+    const primaryKeyAttribute = this.constructor.primaryKeyAttribute;
+    const isAutoGenerated = this.constructor.rawAttributes[primaryKeyAttribute]._autoGenerated;
+
+    if (primaryKeyAttribute && isAutoGenerated && this.isNewRecord === false) {
+      throw new Error('You attempted to query over a model that does not have a primary key explicitly defined this is not allowed');
```
ephys, line 3522: There is a misunderstanding regarding the "has no primary" part:

The goal is to ban using instance methods for models that have no primary at all. Not just that it hasn't been set on the model (i.e. not fetched), but that the table doesn't have a primary key column at all (`this.constructor.primaryKeyAttributes` is empty).

That being said you unearthed another case that needs to be handled too: If the user doesn't fetch the primary key when selecting an entity, then these methods cannot work either.

The fix can be done in two parts. First a check at the very start of this method that ensures a primary key is available:

```typescript
if (this.constructor.primaryKeyAttributes.length === 0) {
  throw new Error(`This model instance method needs to be able to identify the entity in a stable way, but the model does not have a primary key attribute definition. Either add a primary key to this model, or use one of the following alternatives:
  
- instance methods "save", "update", "decrement", "increment": Use the static "update" method instead.
- instance method "reload": Use the static "findOne" method instead.
- instance methods "destroy" and "restore": use the static "destroy" and "restore" methods instead.
  `.trim());
}
```

Then in `primaryKeyAttributes.reduce`: If any attribute is null or undefined, throw something along these lines:

```typescript
throw new Error(`This model instance method needs to be able to identify the entity in a stable way, but this model instance is missing the value of its primary key "${attribute}". Make sure that attribute was not excluded when retrieving the model from the database.`);
```

test/unit/instance/decrement.test.js
```diff
@@ -11,6 +11,13 @@ const sinon     = require('sinon');
 
 describe(Support.getTestDialectTeaser('Instance'), () => {
   describe('decrement', () => {
+    it('should disallow decrement if no primary key values is present', async () => {
+      const Model = current.define('User', {});
+      const instance = Model.build({}, { isNewRecord: false });
+
+      await expect(instance.decrement()).to.be.rejected;
+    });
```
ephys, line 19: A few notes:

- don't name the model "Model", it's the name of the base model, prefer naming it after the table (e.g. "User").
- We're slowly rewriting our tests and we're dropping the "should" to reduce noise
- Add at least part of the error message to make sure it's the right error that arrises, and not an unrelated one
- following the previous note I think this should be two different tests:

```suggestion
    // this one will likely need to go in integration tests
    it('is disallowed if no primary key attribute is defined', async () => {
      const User = current.define('User', {}, { noPrimaryKey: true });
      const instance = await Model.create({});

      await expect(instance.decrement()).to.be.rejectedWith('part of the error message here (no need to copy all of it, just something that identifies it)');
    });
    
    // but this one can remain in unit tests
    it('is disallowed if the instance doesn't have its primary key', async () => {
      const User = current.define('User', {});
      // this instance was built but not saved, so it doesn't have a PK yet.
      const instance = Model.build({});

      await expect(instance.decrement()).to.be.rejectedWith('part of the error message here (no need to copy all of it, just something that identifies it)');
    });
```

ephys (changes requested):
You actually discovered another issues related to PKs. They're closely related but not fully the same :)

frostzt:
@ephys I would love to do it! Thanks for the review lemme get this resolved!

src/model.js
```diff
@@ -3513,16 +3509,27 @@ Instead of specifying a Model, either:
    * @returns {object}
    */
   where(checkVersion) {
+    if (this.constructor.primaryKeyAttributes.length === 0) {
+      throw new Error(
+        `This model instance method needs to be able to identify the entity in a stable way, but the model does not have a primary key attribute definition. Either add a primary key to this model, or use one of the following alternatives:
+  
+      - instance methods "save", "update", "decrement", "increment": Use the static "update" method instead.
```
ephys, line 3516: Can you de-indent this? The log will include the extra whitespace at the beginning of this line

test/unit/instance/reload.test.js
```diff
@@ -11,6 +11,13 @@ const sinon     = require('sinon');
 
 describe(Support.getTestDialectTeaser('Instance'), () => {
   describe('reload', () => {
+    it('is not allowed if the instance does not have a primary key defined', async () => {
+      const User = current.define('User', {});
+      const instance = User.build({});
+
+      await expect(instance.increment()).to.be.rejectedWith('but this model instance is missing the value of its primary key');
```
ephys, line 18:
```suggestion
      await expect(instance.reload()).to.be.rejectedWith('but this model instance is missing the value of its primary key');
```

test/unit/instance/restore.test.js
```diff
@@ -11,6 +11,13 @@ const sinon     = require('sinon');
 
 describe(Support.getTestDialectTeaser('Instance'), () => {
   describe('restore', () => {
+    it('is not allowed if the instance does not have a primary key defined', async () => {
+      const User = current.define('User', {});
+      const instance = User.build({});
+
+      await expect(instance.increment()).to.be.rejectedWith('but this model instance is missing the value of its primary key');
```
ephys, line 18:
```suggestion
      await expect(instance.restore()).to.be.rejectedWith('but this model instance is missing the value of its primary key');
```

test/unit/instance/save.test.js
```diff
@@ -11,11 +11,18 @@ const sinon     = require('sinon');
 
 describe(Support.getTestDialectTeaser('Instance'), () => {
   describe('save', () => {
-    it('should disallow saves if no primary key values is present', async () => {
+    it('is not allowed if the instance does not have a primary key defined', async () => {
+      const User = current.define('User', {});
+      const instance = User.build({});
+
+      await expect(instance.increment()).to.be.rejectedWith('but this model instance is missing the value of its primary key');
```
ephys, line 18:
```suggestion
      await expect(instance.save()).to.be.rejectedWith('but this model instance is missing the value of its primary key');
```

test/unit/model/upsert.test.js
```diff
@@ -14,6 +14,11 @@ describe(Support.getTestDialectTeaser('Model'), () => {
     describe('method upsert', () => {
       before(function () {
         this.User = current.define('User', {
+          id: {
```
ephys, line 17: this is redundant, an auto-incremented integer primary key is added by default if the model doesn't define its own and doesn't use the "noPrimaryKey" option

src/model.js
```diff
@@ -3513,16 +3509,27 @@ Instead of specifying a Model, either:
...
+      - instance method "reload": Use the static "findOne" method instead.
+      - instance methods "destroy" and "restore": use the static "destroy" and "restore" methods instead.
+        `.trim(),
+      );
+    }
+
     const where = this.constructor.primaryKeyAttributes.reduce((result, attribute) => {
+      if (_.isNil(attribute)) {
```
ephys, line 3524: `attribute` is the name of the attribute, you need to check whether its value is nullish (which is returned by `this.get(attribute, { raw: true });`)

ephys (changes requested):

test/unit/instance/save.test.js
frostzt: Damn I don't even know how I missed these, thanks a lot for the patience!

test/unit/instance/save.test.js
```diff
@@ -11,11 +11,18 @@ const sinon     = require('sinon');
...
-      const instance = Model.build({}, { isNewRecord: false });
+    it('is not allowed if the instance does not have a primary key defined', async () => {
+      const User = current.define('User', {});
+      const instance = User.build({});
 
-      await expect(instance.save()).to.be.rejected;
+      await expect(instance.save()).to.be.rejectedWith('but this model instance is missing the value of its primary key');
+    });
+
+    it('should disallow updates if no primary key values is present', async () => {
```
WikiRik, line 21:
```suggestion
    it('should disallow saves if no primary key values is present', async () => {
```

test/unit/instance/save.test.js
```diff
@@ -11,11 +11,18 @@ const sinon     = require('sinon');
...
+      const User = current.define('User', {});
+      const instance = User.build({}, { isNewRecord: false });
+
+      await expect(instance.update()).to.be.rejected;
```
WikiRik, line 25:
```suggestion
      await expect(instance.save()).to.be.rejected;
```

test/integration/instance/update.test.js
```diff
@@ -182,6 +182,14 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
       expect(user.validateSideEffect).not.to.be.ok;
     });
 
+    it('is disallowed if no primary key is present', async function () {
```
WikiRik, line 185: Are there no unit tests for instance/update?

test/unit/instance/restore.test.js
```diff
@@ -11,6 +11,13 @@ const sinon     = require('sinon');
 
 describe(Support.getTestDialectTeaser('Instance'), () => {
   describe('restore', () => {
+    it('is not allowed if the instance does not have a primary key defined', async () => {
```
WikiRik, line 14: Are there no integration tests for instance/restore?

test/integration/instance/decrement.test.js
```diff
@@ -54,6 +54,9 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
       },
     });
 
+    this.Foo = this.sequelize.define('Foo', {}, { noPrimaryKey: true });
```
WikiRik, line 57: Why is this created here? Won't it be duplicated by the new test?

WikiRik (commented):
Few comments about the tests. There might be valid answers but this is what I noticed

test/integration/instance/restore.test.js
```diff
@@ -0,0 +1,36 @@
...
+const chai = require('chai');
+
+const expect = chai.expect;
+const Support = require('../support');
+const { DataTypes } = require('@sequelize/core');
+const sinon = require('sinon');
+
+const current = Support.sequelize;
+
+describe(Support.getTestDialectTeaser('Instance'), () => {
```
ephys, line 12: We have update conventions for our new test suites but we only require them to be applied on new files:

- Use TypeScript
- No need for 'use strict'
- Use import instead of require
- Name the sequelize instance "sequelize" instead of "current"
- Don't use `this`. Replace `this.sequelize` with sequelize. If you need to expose data from before & beforeEach, use `beforeAll2` and `beforeEach2`

test/integration/instance/restore.test.js
```diff
@@ -0,0 +1,36 @@
...
+  before(function () {
+    this.clock = sinon.useFakeTimers();
+  });
+
+  afterEach(function () {
+    this.clock.reset();
+  });
+
+  after(function () {
+    this.clock.restore();
```
ephys, line 22: are these necessary?

test/integration/instance/restore.test.js
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  describe('restore', () => {
```
ephys, line 25: since this test suite is only about restore, you can merge both describe into one named `Model#restore` (Instance used to be a separate class from model, now it's an instance of model)

test/integration/instance/save.test.js
```diff
@@ -129,6 +129,14 @@ describe(Support.getTestDialectTeaser('Instance'), () => {
       });
     });
 
+    it('is disallowed if no primary key is present', async function () {
+      const Foo = this.sequelize.define('Foo', {}, { noPrimaryKey: true });
+      await Foo.sync({ force: true });
+
+      const instance = await Foo.create({}, { isNewRecord: true });
```
ephys, line 136:
```suggestion
      const instance = await Foo.create({});
```

test/unit/instance/save.test.js
```diff
@@ -11,11 +11,11 @@ const sinon     = require('sinon');
 
 describe(Support.getTestDialectTeaser('Instance'), () => {
   describe('save', () => {
-    it('should disallow saves if no primary key values is present', async () => {
-      const Model = current.define('User', {});
-      const instance = Model.build({}, { isNewRecord: false });
+    it('is not allowed if the instance does not have a primary key defined', async () => {
+      const User = current.define('User', {});
+      const instance = User.build({});
```
ephys, line 16: This one should preserve its "isNewRecord: false", because `.save` does two different things based on that property:
- it updates the existing entity in the DB if isNewRecord is false (PK mandatory)
- it inserts a new one in the DB if isNewRecord is true (PK optional)

```suggestion
      const instance = User.build({}, { isNewRecord: false });
```

test/unit/instance/update.test.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const chai = require('chai');
+
+const expect = chai.expect;
+const Support   = require('../support');
+
+const current   = Support.sequelize;
+
+describe(Support.getTestDialectTeaser('Instance'), () => {
```
ephys, line 10: Same comments as the other new suite :)

ephys (dismissed):
There are failing tests related to upsert in the suite, I'll take a look when I can to see what it is about

frostzt:
Alright for some reason the `save`, `update`, and `restore` in the integration tests are being fulfilled when I expect them to throw!?

ephys:
Regarding the failing tests:

# Unit Tests

## Restore Test

The `restore` test is failing because it's calling `save`. Because it's a new model, `save` uses INSERT instead of UPDATE. Which means it doesn't call `.where()`.

How to fix it:

- `restore` should call `update` instead of `save`
- `update` should throw if `isNewRecord` is true (can you add a unit test for that too?)
- The unit test should have `isNewRecord` set to false

## The Upsert tests

I think `upsert` should be investigated deeper. Its current implementation sounds very strange to me. But let's keep that for its own PR.

In the meantime, instead of calling `instance.where()`, it could create the "where" object full of primary keys itself. Like what `where()` does but without throwing. It's a temporary measure until we can investigate `upsert` further.

# Integration Tests

## `restore`, `update`, `save`

None of these methods changes any value on the model (model is not deleted, restoring doesn't change the value of `deletedAt`), `update` and `save` don't change anything. The method aborts before it gets to `.where()`.

How to solve this: we should call `where()` sooner in `save()`. I prefer the consistency of it throwing even if there is nothing to save.

frostzt:
> Member

Actually I don't understand a few things `update should throw if isNewRecord is true (can you add a unit test for that too?)` I don't get what you mean by that? Should it be a straight up check for `isNewRecord`, if true it should throw?

Also this is something I am facing a little trouble with
```
we should call where() sooner in save() (remember: only in update mode, not in insert mode). I prefer the consistency of it throwing even if there is nothing to save.
```

ephys:
Exactly, `update` should check `isNewRecord` and throw if it's true, otherwise it's going to call `save` which is going to execute an insert because it's a new record :)

As for `save`: Sequelize keeps track of changes made locally to a model so that when you call `save`, the method can abort if there is nothing to save (to prevent a useless DB call)

The issue is that [`where()`](https://github.com/sequelize/sequelize/blob/main/src/model.js#L4109) (which checks whether a primary key exists on the model) is called after [the check to abort the save](https://github.com/sequelize/sequelize/blob/main/src/model.js#L4095). As a result, if you try to call `save` on a model that does not have a primary key, and which hasn't changed, the error about the missing primary key will not be thrown (even though it should)

frostzt:
Thanks for the review @ephys I have updated the PR, I don't know if that's exactly what you mentioned, seems like its gonna take me some time understanding the code and the way the repo works haha!

WikiRik:
@ephys ping

ephys:
I've pushed a commit that should fix our unit tests. Let's see whether this passes now

frostzt:
Thank you @ephys

ephys:
The failing test in postgres/mysql/mariadb/sqlite is an easy fix: [this line](https://github.com/sequelize/sequelize/blob/main/test/integration/instance/update.test.js#L177) must use `create` instead of `build`

The failing DB2/mssql should be fixed by changing this line https://github.com/sequelize/sequelize/blob/2e8e13fb2e3fc3b03ef1decf148b139ce0c9d44c/src/model.js#L2583 to `instance.where(false, true) ?? {},`

frostzt:
@ephys I'll get onto this post this week. Actually just got a lot work.

ephys:
Take your time, I took a month to get back to this 😅

WikiRik (approved):
Looks good to me. Not the biggest fan of how `rejectedWith` does not check if you've included the entire error message but that's not a big issue. The error messages we're testing for in this PR are specific enough

ephys:
I don't mind personally, my goal using `rejectedWith` is just to check that the right error is being thrown, not an unrelated one.

WikiRik:
Yeah, that goal is being fulfilled so we're all good

WikiRik:
@ephys I'll leave merging this up to you in case you want to wait for the review of @frostzt

frostzt:
Thank you @ephys and @WikiRik I'll check the commits since I wanna know why this was breaking. Just give me this week post that I'll be back!

frostzt:
Hey @ephys I got it now, do you think it is good to go?

ephys:
unless you see anything else, I think it's ready

frostzt:
Awesome @ephys I don't really notice anything aside the TODOs which I think I can take on in other PRs!

ephys:
@WikiRik I'm doing the merge (for the breaking changes)

PR #15109, opened 2022-10-08
feat: add retry deadlocks by default

fzn0x:
Since deadlocks are classic problem in transactional databases and able to be retried, we should retry it by default from the start or am i missing something?

Ability to wrap entire query process is added in this PR: https://github.com/sequelize/sequelize/pull/8991/files

### DIFF
--- a/src/sequelize.js
+++ b/src/sequelize.js
@@ -245,6 +245,7 @@ export class Sequelize {
         max: 5,
         match: [
           'SQLITE_BUSY: database is locked',
+          /Deadlock/i,
         ],
       },
       transactionType: TRANSACTION_TYPES.DEFERRED,


ephys:
I think doing this would be unsafe. Retrying is good for things like network errors and other benign issues. Retrying on a deadlock can lead to the wrong data being put in the database at worst, and likely won't work at best (transaction could have been rolled back by the database, and if it hasn't the deadlock will still be present).
Deadlocks are something the devs need to solve themselves by either reviewing their flow to prevent a deadlock, or by letting the request fail and let the end-user retry if it only happens occasionally, or other decisions we can't make for them

fzn0x:
> I think doing this would be unsafe. Retrying is good for things like network errors and other benign issues. Retrying on a deadlock can lead to the wrong data being put in the database at worst, and likely won't work at best (transaction could have been rolled back by the database, and if it hasn't the deadlock will still be present). Deadlocks are something the devs need to solve themselves by either reviewing their flow to prevent a deadlock, or by letting the request fail and let the end-user retry if it only happens occasionally, or other decisions we can't make for them
> 
> On that same note, I think retrying on sqlite_busy is a bad idea for the same reasons

Thanks for the clarification, I will trust about that issue to you for further.

PR #15111, opened 2022-10-09
docs: change jsdoc for findbypk

evanrittenhouse:
## Pull Request Checklist

## Todos
Closes https://github.com/sequelize/website/issues/202. Not sure if we want to add more documentation, but I didn't want to duplicate work when `FindOptions` already exists.


fzn0x:
This is a good approach to always sync findAll options with findAndCountAll, so we can avoid some missing options in the future.

fzn0x (dismissed):
LGTM

ephys (changes requested):
placeholder because I want to review before merging

evanrittenhouse:
@ephys Is there somewhere that we have the equivalent of `FindOptions` for v6? If not, would we need to rewrite all `<X>Options` interfaces for v6?

fzn0x:
AFAIK you can only use link https://jsdoc.app/tags-inline-link.html to reference something

fzn0x:
> @fzn0x You can't use link because the type does not exist in JSDoc, it only exists in TypeScript. This is why TypeDoc in v7 can use it but esdoc in v6 cannot
> 
> > Is there somewhere that we have the equivalent of FindOptions for v6
> 
> There isn't unfortunately. If you want to do this you'd have to copy each option here. This is what methods like `Sequelize#query` do: https://sequelize.org/api/v6/class/src/sequelize.js~sequelize#instance-method-query

I thought the same jsdoc could do 😞

evanrittenhouse:
See https://github.com/sequelize/sequelize/pull/15115

evanrittenhouse:
See https://github.com/sequelize/sequelize/pull/15115

Issue #14928, opened 2022-08-24
Support "ADD COLUMN IF NOT EXISTS"


Betree:
## Issue Creation Checklist

- [x] I understand that my issue will be automatically closed if I don't fill in the requested information
- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

### Describe the feature you'd like to see implemented

When creating a new column with `addColumn`, I'd like to be able to provide an option for the method to not crash if the column already exists. For example:

```es6
await queryInterface.addColumn('MyTable', 'new_col', {
  type: Sequelize.INTEGER,
}, {
  mustExist: false, // Prevents crashing if the column already exists
);
```

Using Postgres as a reference, the generated SQL would look like this:

```sql
ALTER TABLE "MyTable" ADD COLUMN IF NOT EXISTS "new_col";
```

### Describe why you would like this feature to be added to Sequelize

Because migrations [don't happen](https://github.com/sequelize/sequelize/issues/7902) in transactions, you can end up with a partially-executed migration that needs to be re-triggered. Adding support for this option would make it easier to have resiliant migrations that can be executed multiple times.

### Is this feature dialect-specific?

- [ ] No. This feature is relevant to Sequelize as a whole.
- [x] Yes. This feature only applies to the following dialect(s): 
  - Postgres: `ADD COLUMN IF NOT EXISTS` is [officially supported ](https://www.postgresql.org/docs/current/sql-altertable.html)
  - MySQL: Not officially supported. Possible [workarounds](https://thispointer.com/mysql-add-column-if-not-exists/)?
  - MSSQL: Not officially supported. Possible [workarounds](https://tableplus.com/blog/2018/08/ms-sql-server-how-to-check-and-add-new-column-if-it-does-not-exist.html)?
  - SQLite: Not officially supported. Possible [workarounds](https://stackoverflow.com/questions/3604310/alter-table-add-column-if-not-exists-in-sqlite)?

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [x] Yes, I have the time but I will need guidance.
- [ ] No, I don't have the time, but my company or I are [supporting Sequelize through donations on OpenCollective](https://opencollective.com/sequelize).
- [ ] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in implementing my feature.

## Context
[...]

PR #15119, opened 2022-10-12
feat(postgres): support for `ADD COLUMN IF NOT EXISTS` and `DROP COLUMN IF EXISTS`

frostzt:
## Pull Request Checklist

## Description Of Change

Its a feature request opened by #14928 
Postgres supports `IF NOT EXISTS` clause on adding columns. Adds support for Postgres dialect!

## Todos

- [ ] Add tests for `IF NOT EXISTS` both Unit and Integration
- [ ] Add dialect support check for all the other dialects


Betree (commented):
Nice work @frostzt, it's looking good so far! Thanks for getting the ball rolling on this one.

I'm not the best person to provide a comprehensive review, but I think we should keep https://github.com/sequelize/sequelize/issues/11774 in mind while implementing this.

frostzt:
@ephys I think if the above is correct I can push the changes for #11774 in this PR too?

WikiRik:
I think it would be good to tackle that in this PR as well

src/dialects/abstract/index.ts
```diff
@@ -23,6 +23,12 @@ export type DialectSupports = {
   skipLocked: boolean,
   finalTable: boolean,
 
+  /**
+   * Does this dialect support checking `IF NOT EXISTS` before adding column
+   * For instance, in Postgres, "ADD COLUMN IF NOT EXISTS" only adds the column if it does not exist
+   */
+  ifNotExists: boolean,
```
ephys, line 30: Multiple queries support "IF NOT EXISTS", but the support is not uniform across dialects. I recommend you name it like this instead:

```
addColumn: {
  ifNotExists: boolean,
},
```

src/dialects/postgres/query-generator.js
```diff
@@ -301,8 +301,9 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
     const definition = this.dataTypeMapping(table, key, dbDataType);
     const quotedKey = this.quoteIdentifier(key);
     const quotedTable = this.quoteTable(this.extractTableDetails(table));
+    const checkColumnExists = attribute.checkColumnExists;
```
ephys, line 304: IMO it shouldn't go on `ModelAttributeColumnOptions`. It doesn't make much sense to define it on the model because it won't ever have an impact when using models.

This is more for when you use `queryInterface` directly, such as in migrations.

Instead, prefer passing it as part of an option bag:

```ts
// abstract/query-generator.d.ts

export interface AddColumnQueryOptions {
  ifNotExists?: boolean;
}

addColumnQuery(
  table: TableName, 
  columnName: string, 
  columnDefinition: ModelAttributeColumnOptions | DataType,
  options?: AddColumnQueryOptions,
);
```

(could you add this to abstract/query-generator.d.ts?)

I also recommend using `ifNotExists` instead of `checkColumnExists`. It's closer to what you'd use in SQL so less prone to misunderstandings

---

In `abstract/query-interface.js`, you will need to pass the option bag to `addColumnQuery` & update its type definition:

```ts
// abstract/query-interface.d.ts

interface AddColumnOptions extends AddColumnQueryOptions, QueryRawOptions {}

addColumn(
  table: TableName,
  columnName: string,
  columnDefinition: ModelAttributeColumnOptions | DataType,
  options?: QiOptionsWithReplacements
): Promise<void>;
```

fzn0x:
Maybe you can rename this PR to includes `DROP COLUMN IF NOT EXIST` if you are sure want to tackle it in this PR, it would be great :) @frostzt

ephys (changes requested):
Nice work :)

Yes feel free to add `DROP COLUMN` it in this PR if you prefer, splitting it is fine too

---

Could you add a unit test in `unit/query-generator/add-column-query.test.ts`?

All dialects should be tested, but the ones that don't support the option should throw (like in `list-databases-query.test.ts`)

I don't think you need an integration test for this option

---

> I do think there are some workarounds for few of them should I try and implement them if possible

I think we can look into that if there is interest but I'd propose to keep it for a follow-up PR

fzn0x:
Better PR name :)

ephys:
Gotta be super explicit for the changelog :p

src/dialects/postgres/query-generator.js
frostzt: @ephys I had a question, you wanted to do this?

```ts
// abstract/query-interface.d.ts

interface AddColumnOptions extends AddColumnQueryOptions, QueryRawOptions {}

addColumn(
  table: TableName,
  columnName: string,
  columnDefinition: ModelAttributeColumnOptions | DataType,
  options?: AddColumnOptions // HERE?
): Promise<void>;
```

Second thing I wanted to ask was related to tests, so every database that does not support `IF NOT EXISTS` should throw, however I don't think implementing this in abstract will work so essentially I'll have to check in `addColumnQuery` for each db and then throw if that options is present and not supported. Is that correct or there is a better way around it?

src/dialects/abstract/query-interface.js
```diff
@@ -448,7 +448,13 @@ export class QueryInterface {
     options = options || {};
     attribute = this.sequelize.normalizeAttribute(attribute);
 
-    return await this.sequelize.queryRaw(this.queryGenerator.addColumnQuery(table, key, attribute), options);
+    const addColumnQueryOptions = {
+      ifNotExists: options.ifNotExists,
+    };
+
+    delete options.ifNotExists;
```
ephys, line 455: Preventing ifNotExists from going to queryRaw is good, but we must avoid mutating our inputs

Prefer this instead:

```
const { ifNotExists, ...queryRawOptions } = options;
```

src/dialects/postgres/query-generator.js
```diff
@@ -295,15 +295,20 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
 
-  addColumnQuery(table, key, attribute) {
+  addColumnQuery(table, key, attribute, options = {}) {
     const dbDataType = this.attributeToSQL(attribute, { context: 'addColumn', table, key });
     const dataType = attribute.type || attribute;
     const definition = this.dataTypeMapping(table, key, dbDataType);
     const quotedKey = this.quoteIdentifier(key);
     const quotedTable = this.quoteTable(this.extractTableDetails(table));
-    const checkColumnExists = attribute.checkColumnExists;
+    let ifNotExists = options.ifNotExists;
```
ephys, line 304: Avoid reassigning a variable if it's not necessary (especially if what it contains changes meaning). Prefer this:

```
const ifNotExists = options.ifNotExists ? 'IF NOT EXISTS ' : '';
```

src/dialects/postgres/query-generator.js
```diff
@@ -295,15 +295,20 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
...
+    if (ifNotExists) {
+      ifNotExists = 'IF NOT EXISTS';
+    } else {
+      ifNotExists = '';
+    }
 
-    let query = `ALTER TABLE ${quotedTable} ADD COLUMN ${checkColumnExists && 'IF NOT EXISTS'} ${quotedKey} ${definition};`;
```
ephys: You're going to have an extra space after ADD COLUMN if `ifNotExists` is false, one of the two spaces should be included in the ifNotExists variable to prevent this

src/dialects/postgres/query-generator.js
ephys: Q1: yep :) just like that

Q2: yes that's right. We already do that for some other methods if you want to do something similar: https://github.com/sequelize/sequelize/blob/main/src/dialects/postgres/query-generator.js#L34

test/unit/query-generator/add-column-query.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    firstName: DataTypes.STRING,
+  }, { timestamps: false });
+
+  it('generates a ADD COLUMN query in supported dialects', async () => {
+    const sql = queryGenerator.addColumnQuery(User.tableName, 'age', DataTypes.INTEGER);
+
+    expectsql(sql, {
+      default: `ALTER TABLE "public"."Users" ADD COLUMN "age" INTEGER;`,
+      postgres: `ALTER TABLE "public"."Users" ADD COLUMN "age" INTEGER;`,
+      mysql: 'ALTER TABLE `Users` ADD `age` INTEGER;',
```
WikiRik, line 20: The MariaDB implementation does not override `addColumnQuery`. Haven't looked into whether it should or not

test/unit/query-generator/add-column-query.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    });
+  });
+
+  it('generates a ADD COLUMN IF NOT EXISTS query in supported dialects', async () => {
+    const sql = queryGenerator.addColumnQuery(User.tableName, 'age', DataTypes.INTEGER, {
+      ifNotExists: true,
+    });
+
+    expectsql(sql, {
```
WikiRik, line 29:
```suggestion
    expectsql(() => sql, {
```
Maybe this works?

test/unit/query-generator/add-column-query.test.ts
```diff
@@ -0,0 +1,35 @@
...
+      default: buildInvalidOptionReceivedError('addColumnQuery', dialectName, ['ifNotExists']),
+      postgres: `ALTER TABLE "public"."Users" ADD COLUMN IF NOT EXISTS "age" INTEGER;`,
+      mysql: buildInvalidOptionReceivedError('addColumnQuery', dialectName, ['ifNotExists']),
```
WikiRik, line 32:
```suggestion
```

test/unit/query-generator/add-column-query.test.ts
```diff
@@ -0,0 +1,35 @@
...
+const dialectName = getTestDialect();
+
+describe('QueryGenerator#addColumnQuery', () => {
+  const queryGenerator = sequelize.getQueryInterface().queryGenerator;
+
+  const User = sequelize.define('User', {
+    firstName: DataTypes.STRING,
+  }, { timestamps: false });
+
+  it('generates a ADD COLUMN query in supported dialects', async () => {
```
WikiRik, line 14: I might be missing something, but why is this `async`?

WikiRik (commented):
Few comments about your tests, haven't looked into the rest yet

test/unit/query-generator/add-column-query.test.ts
frostzt: Actually I found that it was rather the `queryGenerator.addColumnQuery` that throws that error so instead of `() => sql` or simply `sql` I should have passed in that function's call.

test/unit/query-generator/add-column-query.test.ts
frostzt: Actually you're right lol. I write a lot of async tests at work so it has become a habit!

test/unit/query-generator/add-column-query.test.ts
```diff
@@ -0,0 +1,35 @@
...
+  it('generates a ADD COLUMN query in supported dialects', () => {
+    const sql = queryGenerator.addColumnQuery(User.tableName, 'age', {
+      type: DataTypes.INTEGER,
+    });
+
+    expectsql(() => sql, {
+      default: `ALTER TABLE [Users] ADD [age] INTEGER;`,
+      mssql: `ALTER TABLE [Users] ADD [age] INTEGER NULL;`,
+      postgres: `ALTER TABLE "public"."Users" ADD COLUMN "age" INTEGER;`,
+      mysql: 'ALTER TABLE `Users` ADD `age` INTEGER;',
```
WikiRik, line 23: This should be the same as the default since the ticks are added automatically

```suggestion
```

frostzt: Thanks a lot, yeah I actually didn't realize this. Will keep in mind from next time!

test/unit/query-generator/remove-column-query.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const User = sequelize.define('User', {
+    firstName: DataTypes.STRING,
+    age: DataTypes.INTEGER,
+  }, { timestamps: false });
+
+  it('generates a DROP COLUMN query in supported dialects', () => {
+    const sql = queryGenerator.removeColumnQuery(User.tableName, 'age');
+
+    expectsql(() => sql, {
```
WikiRik, line 18:
```suggestion
    expectsql(() => queryGenerator.removeColumnQuery(User.tableName, 'age'), {
```

test/unit/query-generator/remove-column-query.test.ts
```diff
@@ -0,0 +1,33 @@
...
+      default: `ALTER TABLE [Users] DROP COLUMN [age];`,
+      postgres: `ALTER TABLE "public"."Users" DROP COLUMN "age";`,
+      snowflake: `ALTER TABLE "Users" DROP "age";`,
+      sqlite: 'CREATE TABLE IF NOT EXISTS `Users_backup` (`0` a, `1` g, `2` e);INSERT INTO `Users_backup` SELECT `0`, `1`, `2` FROM `Users`;DROP TABLE `Users`;ALTER TABLE `Users_backup` RENAME TO `Users`;',
+      'mariadb mysql': 'ALTER TABLE `Users` DROP `age`;',
```
WikiRik, line 23:
```suggestion
      'mariadb mysql snowflake': `ALTER TABLE [Users] DROP [age];`,
      sqlite: 'CREATE TABLE IF NOT EXISTS `Users_backup` (`0` a, `1` g, `2` e);INSERT INTO `Users_backup` SELECT `0`, `1`, `2` FROM `Users`;DROP TABLE `Users`;ALTER TABLE `Users_backup` RENAME TO `Users`;',
```

test/unit/query-generator/add-column-query.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    expectsql(() => sql, {
+      default: `ALTER TABLE [Users] ADD [age] INTEGER;`,
+      mssql: `ALTER TABLE [Users] ADD [age] INTEGER NULL;`,
+      postgres: `ALTER TABLE "public"."Users" ADD COLUMN "age" INTEGER;`,
+    });
+  });
+
+  it('generates a ADD COLUMN IF NOT EXISTS query in supported dialects', () => {
+    expectsql(() => queryGenerator.addColumnQuery(User.tableName, 'age', {
```
WikiRik, line 27: Why is this not wrapped in a constant as well?

src/dialects/postgres/query-generator.js
```diff
@@ -295,14 +295,15 @@ export class PostgresQueryGenerator extends AbstractQueryGenerator {
     return super.handleSequelizeMethod.call(this, smth, tableName, factory, options, prepend);
   }
 
-  addColumnQuery(table, key, attribute) {
+  addColumnQuery(table, key, attribute, options) {
```
WikiRik, line 298:
```suggestion
  addColumnQuery(table, key, attribute, options = {}) {
```
Any reason why the options object is not defaulted to an empty object?

src/dialects/postgres/index.js
```diff
@@ -22,6 +22,9 @@ export class PostgresDialect extends AbstractDialect {
     bulkDefault: true,
     schemas: true,
     databases: true,
+    addColumn: {
+      ifNotExists: true,
+    },
     lock: true,
```
WikiRik, line 28: You forgot the `removeColumn: { ifExists: true }` here

src/dialects/abstract/query-interface.js
```diff
@@ -459,7 +462,12 @@ export class QueryInterface {
    * @param {object} [options]      Query options
    */
   async removeColumn(tableName, attributeName, options) {
```
WikiRik, line 464:
```suggestion
  async removeColumn(tableName, attributeName, options = {}) {
```
This might be a nice refactor, for `addColumn` as well I think. Or is there a reason not to do this?

src/dialects/abstract/query-generator.d.ts
```diff
@@ -120,6 +129,19 @@ export class AbstractQueryGenerator {
     columnDefinitions?: { [columnName: string]: BuiltModelAttributeColumOptions }
   ): string;
 
+  addColumnQuery(
```
WikiRik, line 132: @ephys I'll let you look at these

src/dialects/mariadb/index.js
```diff
@@ -24,6 +24,12 @@ export class MariaDbDialect extends AbstractDialect {
         ignoreDuplicates: ' IGNORE',
         updateOnDuplicate: ' ON DUPLICATE KEY UPDATE',
       },
+      addColumn: {
```
WikiRik, line 27: It seems that MariaDB does actually support this; https://mariadb.com/kb/en/alter-table/

src/dialects/mssql/index.js
```diff
@@ -25,6 +25,12 @@ export class MssqlDialect extends AbstractDialect {
       defaultValue: false,
       update: false,
     },
+    addColumn: {
```
WikiRik, line 28: It seems that mssql does actually support this; https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-2017

WikiRik (changes requested):
It seems that this is not a postgres exclusive feature

@ephys I'll let you look at the type definitions for {add,remove}ColumnQuery

test/unit/query-generator/add-column-query.test.ts
frostzt: Actually `addColumnQuery` throws an error when `ifNotExists` is passed to `options` therefore if I were to assign it to a constant the error would be thrown there.

```js
const sql = queryGenerator...
```

However to get the proper `default` behavior we invoke and expect it to equal `buildInvalidOptionReceivedError` call.

test/unit/query-generator/remove-column-query.test.ts
frostzt: Weirdly enough if I do this

```js
'mariadb mysql snowflake': `ALTER TABLE [Users] DROP [age];`
```
It is not able to replace the `[` with backticks `

src/dialects/postgres/query-generator.js
frostzt: I think I am doing it down below.

```js
options = options || {}
```

The reason for doing this (the above way) is that I am passing `options` explicitly as null in case the `ifNotExists` argument wasn't provided.

src/dialects/abstract/query-interface.js
frostzt: The same as above, also lemme give a reason as well why am I doing that `null`. The options is directly passed in the `rejectInvalidOptions` so as it skip it when no args are present I pass null.

And for double checking I am doing this.

```js
options = options || {};
```

src/dialects/mariadb/index.js
frostzt: bam bam this is awesome. @ephys how about I do it in subsequent PRs?

src/dialects/mssql/index.js
frostzt: same for this as mariadb?

test/unit/query-generator/remove-column-query.test.ts
WikiRik: That's strange. @ephys do you know why?

src/dialects/mariadb/index.js
WikiRik: I have a slight preference to do it in this PR directly, but I'm fine either way

src/dialects/abstract/query-interface.js
WikiRik: Maybe I'm missing something, but don't they do the same? So I want to set it by default when defining the function and you want to add it two lines later on.
In both cases `options` will be `{}` if it has not been set but in my suggestion it is done in the definition directly instead of having an additional line of code

src/dialects/postgres/query-generator.js
WikiRik: See my other comment about this, same holds for this one as well

test/unit/query-generator/remove-column-query.test.ts
frostzt: @WikiRik I think its because of [this](https://github.com/sequelize/sequelize/blob/main/test/support.ts#L431)
We are only doing it for `expectations.default`

WikiRik: That seems to be the case indeed. Then we'll leave this for now and we can fix it in a future refactor

src/dialects/mariadb/index.js
ephys: Same, I'm fine with either

ephys (approved):
This looks good to me. I'm OK with adding support for other dialects in follow-up PRs. :)

src/dialects/abstract/query-interface.js
ephys: > const options = ifExists ? { ifExists } : null;

Prefer setting it to undefined, so the default parameter operator can be used. I think it's cleaner

WikiRik (approved):
Yes, let's add support for additional dialects in future PRs and merge this one now. Would be nice if you could still work on those @frostzt :)

fzn0x:
Happy to see this is landed!

frostzt:
@WikiRik thanks I'll start working on those!

Issue #14519, opened 2022-05-18
PSA: Upgrade to Sequelize 6.19.2 & Our recent breaking change in v6


ephys:
Hi everyone 👋

We have recently released Sequelize 6.19.1 (since patched as 6.19.2) which includes a [breaking change](https://github.com/sequelize/sequelize/pull/14472). 

We normally want to fully respect semver rules, unfortunately a very serious [SQL injection exploit](https://github.com/sequelize/sequelize/issues/13817) has surfaced and it was impossible for us to fix it without introducing a breaking change. We considered the issue serious enough to break the rule and release the fix in V6. Most of you are going to be completely unaffected, but a small percentage of users may experience breakage with this version, more below

## The exploit

The SQL injection exploit is related to replacements. Here is such an example: 

In the following query, some parameters are passed through replacements, and some are passed directly through the `where` option.

```typescript
User.findAll({
  where: or(
    literal('soundex("firstName") = soundex(:firstName)'),
    { lastName: lastName },
  ),
  replacements: { firstName },
})
```

This is a very legitimate use case, but this query was vulnerable to SQL injection due to how Sequelize processed the query: Sequelize built a first query using the `where` option, then passed it over to `sequelize.query` which parsed the resulting SQL to inject all `:replacements`.

If the user passed values such as

```json
{
  "firstName": "OR true; DROP TABLE users;",
  "lastName": ":firstName"
}
```

Sequelize would first generate this query:

```sql
SELECT * FROM users WHERE soundex("firstName") = soundex(:firstName) OR "lastName" = ':firstName'
```

Then would inject replacements in it, which resulted in this:

```sql
SELECT * FROM users WHERE soundex("firstName") = soundex('OR true; DROP TABLE users;') OR "lastName" = ''OR true; DROP TABLE users;''
```

As you can see this resulted in arbitrary user-provided SQL being executed.

## The solution & breaking change

The way to fix this without breaking everyone's codebase was to use a smarter SQL parser that would only inject replacements in places where native SQL parameters are allowed (with some safe exceptions to stay as backward compatible as possible). 

This means that `:replacements` will not be treated as such anymore if they are quoted, as we treat it as being part of a SQL string.

Details here: https://github.com/sequelize/sequelize/blob/v6/test/unit/utils/sql.test.js

On top of this, in Sequelize 7, we rewrote this system completely to handle both steps at the same time. Removing all risk of SQL injection (https://github.com/sequelize/sequelize/pull/14447).

[...]

PR #15123, opened 2022-10-13
fix(types): expose legacy "types" folder in export alias

anderson-arlen:
## Description Of Problem

Imports of things like 'sequelize/types/hooks' (which is used by sequelize-typescript) is broken when using Typescript's "nodenext" or "node16" module resolution strategies. One could update the imports to 'sequelize/lib/hooks', but doing so breaks the default Typescript module resolution. So projects like sequelize-typescript can't easily make that change.

**Currently anyone using the newer module resolution strategies along with sequelize-typescript have no upgrade path to a version of sequelize without the vulnerability mentioned in #14519** 

## Description Of Change

This is a different approach to #14719.

Instead of adding hooks to the exports in index.d.ts, we can just add a  "/types/" alias to the exports in package.json. Doing so allows imports of 'sequelize/types/hooks' to properly resolve using both older and newer Typescript module resolution strategies. In essence, importing /types/* will work as it always has regardless of module resolution strategy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,10 @@
       "default": "./lib/errors/index.js",
       "types": "./types/errors/index.d.ts"
     },
-    "./package.json": "./package.json"
+    "./package.json": "./package.json",
+    "./types/*": {
+      "types": "./types/*.d.ts"
+    }
   },
   "engines": {
     "node": ">=10.0.0"


fzn0x:
Will the typings conflict?

ephys (approved):
I've tested this against a few configurations and it works, no conflicts, thank you :)

cage1618:
Cannot find module './data-types' after update to 6.25.1 😭

fzn0x:
> Cannot find module './data-types' after update to 6.25.1 😭 @ephys

Hi, just in case someone looking up about this types issue, this is already fixed by ephys :), refer to latest v6, thank you!

PR #15131, opened 2022-10-15
feat: migrate hooks to TypeScript

ephys:
## Pull Request Checklist

## Description Of Change

This PR migrates hooks to TypeScript. It uses composition instead of mixins since mixins can't be typed properly. Inspired by https://github.com/sequelize/sequelize/pull/12160

To avoid introducing an extra breaking change, legacy ways of hooking are still available, but are deprecated in favor of the `.hooks` property.

This PR mixes many, many errors surrounding hook typings, since there is only one source of truth for them now.

The new hooks are stricter. They'll throw if you try to return a promise from a sync hook, or if you try to listen to a hook that doesn't exist.

Also, because migrating model.js & sequelize.js is going to take a hell of a lot of work, I'm introducing a parent class to both of them so we can slowly migrate things instead of doing it all in the same PR

## Next steps

Update the documentation to use the `.hooks` property instead of the legacy methods


test/integration/hooks/hooks.test.js
```diff
@@ -310,96 +311,48 @@ describe(Support.getTestDialectTeaser('Hooks'), () => {
...
-      await this.User.create({ username: 'makunouchi' });
-      expect(sasukeHook).to.have.been.calledOnce;
-      expect(narutoHook).to.have.been.calledOnce;
-      this.User.removeHook('beforeCreate', sasukeHook);
-      await this.User.create({ username: 'sendo' });
-      expect(sasukeHook).to.have.been.calledOnce;
-      expect(narutoHook).to.have.been.calledTwice;
-    });
-
-    it('should be able to remove proxies', async function () {
```
WikiRik, line 389: Is this not possible anymore?

WikiRik:
> Also, because migrating model.js & sequelize.js is going to take a hell of a lot of work, I'm introducing a parent class to both of them so we can slowly migrate things instead of doing it all in the same PR

That's actually a very smart idea, should really help us with the migration!

---

Do you consider this PR ready? Both in the changes, but also in the tests. 

Also, can you check if the new version of `expect-type` works with this PR? Since it was complaining about hooks before.

test/integration/hooks/hooks.test.js
ephys: It is but the test is redundant because there is no concept of proxies anymore. Too much boilerplate for something that was used once. I replaced it with a regular hook

WikiRik: Sounds good!

ephys:
I do consider it ready. Ideally we'd migrate the hook tests too but I don't have the bandwidth for it right now. I didn't even plan on migrating this I just wanted to import the Hook Decorators from sequelize-typescript 😅

Expect-type 0.14.2 still fails with the hook option writability tests

WikiRik (approved):
Quite a difference in implementation but looks good! We should add tests for the new way later as well

src/hooks-legacy.ts
```diff
@@ -0,0 +1,130 @@
+import type { HookHandlerBuilder } from './hooks.js';
+import { hooksReworked } from './utils/deprecations.js';
+
+// TODO: delete this in Sequelize v8
```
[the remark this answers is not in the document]
ephys: It wouldn't work because we don't update the version property :( https://github.com/sequelize/sequelize/blob/main/package.json#L4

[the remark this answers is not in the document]
WikiRik: Oh yeah, that's true. But it would still be nice so we can manually set it locally and check the warnings. It's a non-blocking issue anyway

WikiRik:
Documentation PR is here; https://github.com/sequelize/website/pull/310

Issue #15133, opened 2022-10-15
Sequelize typings break w/ TS 4.9 beta


ethanresnick:
## Issue Creation Checklist

- [x] I understand that my issue will be automatically closed if I don't fill in the requested information
- [x] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description

Using TS 4.9's latest beta (`4.9.0-dev.20221015`) and Sequelize `6.25.1`, models that use `InferAttributes` now produce a type error, whereas they work in earlier TS versions.

### Reproducible Example

```ts
import sequelize, {  InferAttributes, InferCreationAttributes } from 'sequelize';

const { Model } = sequelize;

class TestModel extends Model<
  InferAttributes<TestModel>,
  InferCreationAttributes<TestModel>
> {
  public declare id: string;
}
```

### What do you expect to happen?

Code should compile.

### What is actually happening?

TS error, highlighting the expression: 

```
error TS2589: Type instantiation is excessively deep and possibly infinite.

6   InferAttributes<TestModel>,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### Environment

- Sequelize version: 6.25.1
- TypeScript version: 4.9.0-dev.20221015
- Node.js version: n/a
- Database & Version: n/a
- Connector library & Version: n/a

## Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I will need guidance.
- [ ] No, I don't have the time, but my company or I are [supporting Sequelize through donations on OpenCollective](https://opencollective.com/sequelize).
- [x] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in resolving my issue.

---

*Indicate your interest in the resolution of this issue by adding the 👍 reaction. Comments such as "+1" will be removed.*

ephys:
Thanks for the heads up. https://github.com/sequelize/sequelize/pull/15134 & https://github.com/sequelize/sequelize/pull/15135 should fix this

WikiRik:
Fixed in the latest releases

ethanresnick:
Thanks!

PR #15134, opened 2022-10-15
fix(types): fix TS 4.9 excessive depth error on `InferAttributes` (v7)

ephys:
## Pull Request Checklist

## Description Of Change

#15133: The latest TS build errors on `InferAttributes` with a `Type instantiation is excessively deep and possibly infinite.` error. Thankfully the fix is easy: Exclude keys inherited from Model before checking its values to reduce the type depth. Should improve perfs too.

No new test needed, but this wasn't caught by our CI yet. Will be once we add TypeScript 4.9

This should also get us ready for TS 4.9, I fixed all new typing issues

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "copyfiles": "2.4.1",
     "cross-env": "7.0.3",
     "delay": "5.0.0",
-    "esbuild": "0.15.10",
+    "esbuild": "0.15.11",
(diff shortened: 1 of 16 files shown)


ephys:
it works on 4.9 so of course it fails on all other versions 😄

WikiRik (approved):

PR #15139, opened 2022-10-16
fix: don't treat \ as escape in standard strings, support E-strings, support vars after ->> operator, treat lowercase e as valid e-string prefix

sdepold:
Backport of #14700 and #14733.

### DIFF
--- a/test/unit/utils/sql.test.js
+++ b/test/unit/utils/sql.test.js
@@ -1,17 +1,29 @@
 const { injectReplacements } = require('sequelize/lib/utils/sql');
 const { expect } = require('chai');
-const { expectsql, sequelize } = require('../../support');
+const {
+  expectsql,
+  sequelize,
+  createSequelizeInstance,
+  expectPerDialect,
+  toHaveProperties,
+  toMatchSql
+} = require('../../support');
 
 const dialect = sequelize.dialect;
 
 describe('injectReplacements (named replacements)', () => {
   it('parses named replacements', () => {
-    const sql = injectReplacements(`SELECT ${dialect.TICK_CHAR_LEFT}:id${dialect.TICK_CHAR_RIGHT} FROM users WHERE id = ':id' OR id = :id OR id = ''':id'''`, dialect, {
-      id: 1
-    });
+    const sql = injectReplacements(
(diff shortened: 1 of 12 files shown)


WikiRik (commented):
It feels a bit stupid, but can you undo the various style changes? Makes it easier to review the changes, especially in the tests.

Apart from that it looks good to me

sdepold:
> It feels a bit stupid, but can you undo the various style changes? Makes it easier to review the changes, especially in the tests.
> 
> Apart from that it looks good to me

to be completely honest with you, I think we should stop wasting time talking about formatting :D we have the linter in place to prevent syntax problems. it even auto-formats stuff as it see needs. it's all green. so why care about something that is legacy anyway. however, I've reverted the changes to best of my abilities, even preventing the auto-formatting of my editor to turn it all upside down again.

test/unit/utils/sql.test.js
```diff
@@ -44,6 +62,18 @@ describe('injectReplacements (named replacements)', () => {
...
+  it('parses bind parameters following JSONB indexing', () => {
+    const sql = injectReplacements(
+      'SELECT * FROM users WHERE json_col->>$key',
+      dialect,
+      { id: 1 }
+    );
+
+    expectsql(sql, {
+      default: 'SELECT * FROM users WHERE json_col->>$key'
+    });
```
WikiRik, line 74: Why are these commented out?

test/unit/utils/sql.test.js
```diff
@@ -109,14 +139,83 @@ describe('injectReplacements (named replacements)', () => {
...
+        default:
+          new Error(`The following SQL query includes an unterminated string literal:
+SELECT * FROM users WHERE id = E'\\' $id' OR id = $id`),
+        'mysql mariadb': toMatchSql(
+          "SELECT * FROM users WHERE id = E'\\' $id' OR id = $id"
+        ),
+        postgres: "SELECT * FROM users WHERE id = E'\\' $id' OR id = $id"
+      }
+    );
+  });
```
WikiRik, line 199: What is the difference between mysql/mariadb and postgres?

sdepold: probably none :) there was a diff in the original version but that was due to a difference in functionality where the function would return bind values. let me adjust this case

WikiRik (approved):

ephys (approved):

PR #15144, opened 2022-10-17
fix(postgres): invalidate connection after client-side timeout

harrykao:
## Pull Request Checklist

## Description Of Change

The `query_timeout` feature of the `pg` package helps handle stuck TCP connections more quickly and gracefully by implementing a client-side timeout:

https://github.com/brianc/node-postgres/issues/1713

Sequelize started passing this dialect-specific option through to `pg` here:

https://github.com/sequelize/sequelize/pull/13258

I believe we also want to invalidate the connection when a client-side timeout occurs. We shouldn't try to reuse the stuck connection because...it's stuck.

This PR updates the error handling code so that the connection is invalidated if the error matches the one thrown from here:

https://github.com/brianc/node-postgres/blob/5538df6b446f4b4f921947b460fe38acb897e579/packages/pg/lib/client.js#L529


harrykao:
@ephys @WikiRik Sorry to bug you. What's the right way to ask someone to review this?

WikiRik:
Hi @harrykao 
This is fine. We'll take a look at this soon. Thanks for the message

src/dialects/postgres/query.js
```diff
@@ -51,6 +51,13 @@ export class PostgresQuery extends AbstractQuery {
         || /Unable to set non-blocking to true/i.test(error)
         || /SSL SYSCALL error: EOF detected/i.test(error)
         || /Local: Authentication failure/i.test(error)
+
+        // If the query fails with a client-side timeout, invalidate the connection because
+        // it's probably stuck. (Note that query_timeout should be greater than
```
WikiRik, line 56: You mention that `query_timeout` should be greater than `statement_timeout` but what happens if it's not? Is this something we should document or is this already a documented best practice by https://github.com/brianc/node-postgres ?

WikiRik (commented):
Do you think we can make a test with a very low `query_timeout` and a query that does take a while? Not a high priority but would be nice to see.

Apart from that I think this should not be an issue, but I do have a question about your explanation. See my comment about that

src/dialects/postgres/query.js
harrykao: The purpose of `query_timeout`, I think, is to help the client know when the connection is stuck because it sometimes takes quite a long time for it to realize that on its own (15 minutes in our environment!). We know it's a stuck connection if the query takes longer than the statement timeout, because otherwise we would have seen a statement timeout.

So setting `query_timeout` < `statement_timeout` means that we'll never see statement timeouts and we'll invalidate connections unnecessarily.

This should probably be documented in node-postgres but isn't. (Or maybe it's considered obvious enough?)

harrykao:
@WikiRik Thanks for the review. I've added a test.

I'm not sure about coding style here and I'd be happy to do another round to make this more consistent with the rest of the code.

harrykao:
Hrm, it looks to me like the native bindings don't pass `statement_timeout`:

https://github.com/brianc/node-pg-native/blob/master/index.js#L14

Or implement `query_timeout`.

Would it make sense to skip these tests when testing against native?

WikiRik:
I think it would be fine to skip these on `postgres-native`. You can just add an if-statement at the top of the describe block that just returns if the dialect is postgres-native. That reduces nesting and is our new best practice

WikiRik (approved):
One small suggestion, but it's non blocking
I think it's fine having the new test functions here, especially since a developer might not directly expect that `getConnectionPid` also releases the connection.

WikiRik (approved):

harrykao:
@WikiRik Could I get a v6 release with this fix?

PR #15146, opened 2022-10-18
feat: allow attribute removal from model instance

mwaqar666:
## Pull Request Checklist

## Description Of Change

Added a method named `removeDataValue(key)` on the model instance to allow removing attributes from the instance after querying the database.

### Example

```
const jane = await User.create({ firstName: "Jane", lastName: "Doe" });
```

Using the `removeDataValue` method to remove `firstName` from the instance
```
console.log(jane.firstName); // Jane
jane.removeDataValue("firstName");
console.log(jane.firstName); // undefined
```

### DIFF
--- a/src/model.d.ts
+++ b/src/model.d.ts
@@ -3413,6 +3413,13 @@ export abstract class Model<TModelAttributes extends {} = any, TCreationAttribut
   setDataValue<K extends keyof TModelAttributes>(key: K, value: TModelAttributes[K]): void;
 
   /**
+   * Removes the underlying data value.
+   *
+   * @param key The name of the attribute to remove.
+   */
+  removeDataValue<K extends keyof TModelAttributes>(key: K): void;
+
+  /**
    * If no key is given, returns all values of the instance, also invoking virtual getters. If an object has child objects or if options.clone===true, the object will be a copy. Otherwise, it will reference instance.dataValues.
    *
    * If key is given and a field or virtual getter is present for the key it will call that getter - else it
--- a/src/model.js
+++ b/src/model.js
@@ -3567,6 +3567,17 @@ Instead of specifying a Model, either:
   }
 
   /**
+   * Removes the underlying data value
+   *
+   * @param {string} key The name of the attribute to remove.
+   */
+  removeDataValue(key) {
+    this.changed(key, true);
+
+    delete this.dataValues[key];
+  }
+
+  /**
    * If no key is given, returns all values of the instance, also invoking virtual getters.
    *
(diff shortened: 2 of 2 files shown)


mwaqar666:
I have a use case where after sending the response from the server, I remove all the primary and foreign keys from the result set through an interceptor. For that purpose, I created some static properties on base model class and override them on each model accordingly. Then, in the interceptor, I removed them using this function

```
1	private stripPrimaryAndForeignKeysFromEntity(data: BaseEntity<any>): any {
2		const sequelizeEntityModel = data.constructor as EntityType<BaseEntity<any>>;
3		const primaryKeyColumnName = sequelizeEntityModel.primaryKeyAttribute;
4
5		const [foreignKeys, relationShipProperties] = this.extractRelationShipKeysAndForeignKeys(sequelizeEntityModel);
6
7		data.removeDataValue(primaryKeyColumnName);
8		foreignKeys.forEach((foreignKey: string) => data.removeDataValue(foreignKey));
9
10		for (const [relationShipType, relationShipTypeProperties] of Object.entries(relationShipProperties)) {
11			relationShipTypeProperties.forEach((relationShipProperty: string) => {
12				const relationShip = data.getDataValue(relationShipProperty);
13
14				if (!relationShip) return;
15
16				const processedRelationShip = relationShipType === "singleRelationShipProperties"
17					? this.stripPrimaryAndForeignKeysFromEntity(relationShip)
18					: relationShip.map((eachMultiRelationship) => this.stripPrimaryAndForeignKeysFromEntity(eachMultiRelationship));
19
20				data.setDataValue(relationShipProperty, processedRelationShip);
21			});
22		}
23
24		return data;
25	}
```

The function I added, i.e `removeDataValue` is being used along with `setDataValue` in this function, specifically on line 7 and 8 to remove the primary key and foreign keys.

Side note: I'm removing them, because I do not expose these keys on front end, instead for record identification, I use `uuid` and send them on front end.

ephys:
Thank you for making this PR, and I'm sorry in advance, but I don't think this is something we should add. While it would help with your use case we have to be careful with our API design and how it interacts with the rest of our API

I do the same thing you do (internalId (serial pk, not exposed) & externalId (nanoid string, exposed)) but there are other ways to handle it, for instance:

- implement the `toJSON` method in your model to only return the data you need. If you use model inheritance you should even be able to write that just once in the call all of your models extend
- alternative, call `toJSON` in your interceptor and delete the keys you don't want to expose on the resulting object

WikiRik:
I'll close this PR, but feel free to comment on it if you do not agree or have additional questions/comments

PR #15159, opened 2022-10-22
meta: swap Slack links (v6)

sdepold:
As discussed, we have to get rid of Heroku and hence swap the invite links with the direct Slack powered feature.


WikiRik (approved):

sdepold:
retest

sdepold:
@WikiRik any idea why this would not build? is that a known thing?

WikiRik:
Maybe because you switched the branches? I haven't really seen this before. Try pushing an empty commit to restart the CI

sdepold:
ha. now it works :)

PR #15168, opened 2022-10-23
meta: add pnpm and npm lock files and remove .npmrc

fzn0x:
We are not using `package-lock.json` anymore, and replaced it with `yarn.lock`

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     "sqlite3": "5.1.2",
     "tedious": "15.1.0",
     "ts-node": "10.9.1",
-    "typedoc": "0.23.17",
+    "typedoc": "0.23.18",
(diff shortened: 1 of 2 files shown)


WikiRik (commented):
I'm not so sure about this change. Having this npmrc file will ensure that even when people install dependencies with npm when working on sequelize that it will not generate a package-lock file and therefore also cannot be commited accidentally. But I'm open to thoughts from others

WikiRik:
For sure would be good to add package-lock and the pnpm alternative (if it's different) to `.gitignore`

WikiRik (approved):

ephys (approved):
The ideal solution would be to have a way to make npm, pnpm etc throw but this is fine

PR #15169, opened 2022-10-23
meta: remove `sscce.js`

fzn0x:
This file is not relevant anymore since you just need to fork the [Sequelize SSCCE](https://github.com/sequelize/sequelize-sscce) and add the PR of your changes on [Source Folder](https://github.com/sequelize/sequelize-sscce/tree/main/src) from your fork.


WikiRik (changes requested):
I think we should keep this file, this way people can debug directly with the source code/`main` branch instead of doing other tricks using the SSCCE repo

PR #15174, opened 2022-10-24
fix(typings): remove `escape` & `quoteTable` from query-interface typings

### LINKED ISSUE #13225: QueryInterface.escape is not a function - but defined in Types
## Issue Description

Using Postgres adapter and have updated from 4.x to 6.5.x and this:

```ts
sequelize.getQueryInterface().escape('"hey"');
```

Gives the following error:

```
TypeError: getQueryInterface(...).escape is not a function
```

Method defined as being available here: 
https://github.com/sequelize/sequelize/blob/v6/src/dialects/abstract/query-interface.d.ts#L659-L662

- [x] I don't know.

- x ] Yes, I have the time but I don't know how to start, I would need guidance.

I can submit a PR but I'm not sure if this is dialect-specific or how much of QueryInterface type definition is true/false

WikiRik:
## Pull Request Checklist

## Description Of Change

It feels a bit stupid to just do this, but I think this is all we require. The escape function is part of the queryGenerator.

Backport of #14229

Fixes #13225

### DIFF
--- a/src/dialects/abstract/query-interface.d.ts
+++ b/src/dialects/abstract/query-interface.d.ts
@@ -657,11 +657,6 @@ export class QueryInterface {
   public quoteIdentifiers(identifiers: string): string;
 
   /**
-   * Escape a value (e.g. a string, number or date)
-   */
-  public escape(value?: string | number | Date): string;
-
-  /**
    * Set option for autocommit of a transaction
    */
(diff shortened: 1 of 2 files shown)


ephys:
It is, i don't think we should add quote to queryInterface (but we should eventually make it easier to access queryGenerator)

ephys (approved):

WikiRik:
It seems that @fzn0x actually already fixed this for `main` in #14229

PR #15175, opened 2022-10-24
feat: enable the ability to use queryGenerator in queryInterface v6

### LINKED ISSUE #13225: QueryInterface.escape is not a function - but defined in Types

fzn0x:
## Pull Request Checklist

## Description Of Change

Add ability to use `escape` and `quoteTable`  from `queryGenerator` and remove them from `queryInterface`

This also:
- Add missing SHOWCONTRAINTS in QueryTypes 

Closes #13225

### DIFF
--- a/src/dialects/abstract/query-interface.js
+++ b/src/dialects/abstract/query-interface.js
@@ -923,7 +923,7 @@ class QueryInterface {
 
     const sql = this.queryGenerator.updateQuery(tableName, values, identifier, options, attributes);
     const table = _.isObject(tableName) ? tableName : { tableName };
-    const model = _.find(this.sequelize.modelManager.models, { tableName: table.tableName });
+    const model = options.model ? options.model : _.find(this.sequelize.modelManager.models, { tableName: table.tableName });
 
     options.type = QueryTypes.BULKUPDATE;
     options.model = model;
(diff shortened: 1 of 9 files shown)


src/query-types.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @property UPSERT
+ * @property VERSION
+ * @property SHOWTABLES
+ * @property SHOWINDEXES
+ * @property DESCRIBE
+ * @property RAW
+ * @property FOREIGNKEYS
+ * @property SHOWCONSTRAINTS
+ */
+export enum QueryTypes {
```
ephys, line 21: we're not migrating files to TypeScript in v6, it caused too many problems

ephys (changes requested):

Issue #15227, opened 2022-11-03
ibmi Dialect Issues


hancockm:
## Issue Creation Checklist

- [x] I understand that my issue will be automatically closed if I don't fill in the requested information
- [ ] I have read the [contribution guidelines](https://github.com/sequelize/sequelize/blob/main/CONTRIBUTING.md)

## Bug Description

When using the ibmi dialect of Sequelize library in sequelize-automate, I noticed that the database object passed should have been an AS400 library/Schema not DSN. I have created an updated version of the plugin that conforms to the other dialects.

### Reproducible Example

Yes, I can provide one, but the version 7 doesn't include a test for the ibmi database.

Here is the link to the SSCCE for this issue:

### What do you expect to happen?

Update ibmi dialect plugin to have composability with other dialects and the sequelize object. 

### What is actually happening?

![image](https://user-images.githubusercontent.com/6729793/199761186-faec2927-9c5b-4797-96e8-fdec6f082cbb.png)

### Environment

- Sequelize version: sequelize@6.25.3
 Node.js version: 8.15.1 
 If TypeScript related: TypeScript version: N/A
- Database & Version: IBMi 7.2
- Connector library & Version: odbc@2.4.6

## Would you be willing to resolve this issue by submitting a Pull Request?

- [x] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I will need guidance.
- [ ] No, I don't have the time, but my company or I are [supporting Sequelize through donations on OpenCollective](https://opencollective.com/sequelize).
- [ ] No, I don't have the time, and I understand that I will need to wait until someone from the community or maintainers is interested in resolving my issue.

---

*Indicate your interest in the resolution of this issue by adding the 👍 reaction. Comments such as "+1" will be removed.*
👍

fzn0x:
I'm not sure I'm understand the screenshot you provide in **What is actually happening?**, can you explain about that?

WikiRik:
Thanks for raising a PR for this!

hancockm:
> I'm not sure I'm understand the screenshot you provide in **What is actually happening?**, can you explain about that?
@fzn0x 

This screenshot is from a library called sequelize-automate. I have tested it against mssql instance, but when I put in the arguments required for the ibmi dialect, this is the error that is returned. The ibmi dialect is initialized with the new Sequelize constructor, it substitutes a database with the DSN file.  

If you look at the "database options" object in the example, the DSN file name is the object assigned to the property "database". 

[...]

PR #15228, opened 2022-11-03
fix(ibmi): use CurrentSchema insteadof DSN

hancockm:
Co-authored-by: renovate[bot] <29139614+renovate[bot]@users.noreply.github.com>

## Pull Request Checklist

## Description Of Change

https://github.com/sequelize/sequelize/issues/15227

Updated the ibmi dialect constructor and connect method. I added the property "DSN" and mapped the IBMi Schema/Library to the "database" property.  There was also a small correction to the "describeTable" method by adding a valdiation to check to see if "tableName" parameter is an object.

## Todos

- [ ] 
- [ ] 
- [ ]

### DIFF
--- a/src/dialects/ibmi/connection-manager.js
+++ b/src/dialects/ibmi/connection-manager.js
@@ -35,8 +35,18 @@ export class IBMiConnectionManager extends AbstractConnectionManager {
 
     // 'database' doesn't make sense in this context, but it is mapped here to
     // DSN, which is a close fit
+    //  if (config.database) {
+    //    connectionKeywords.push(`DSN=${config.database}`);
+    //  }
+
+    if (config.DSN) {
+      connectionKeywords.push(`DSN=${config.DSN}`);
+    }
+
     if (config.database) {
-      connectionKeywords.push(`DSN=${config.database}`);
+      connectionKeywords.push(`CurrentSchema=${config.database}`);
+    } else {
+      connectionKeywords.push(`CurrentSchema=QGPL`);
     }
 
     if (config.username) {
--- a/src/dialects/ibmi/connection-manager.js
+++ b/src/dialects/ibmi/connection-manager.js
@@ -38,10 +38,10 @@ export class IBMiConnectionManager extends AbstractConnectionManager {
     //  if (config.database) {
     //    connectionKeywords.push(`DSN=${config.database}`);
     //  }
-  
-    if (config.dialectOptions && config.dialectOptions.DSN){
+
+    if (config.dialectOptions && config.dialectOptions.DSN) {
       connectionKeywords.push(`DSN=${config.dialectOptions.DSN}`);
-     }
+    }
 
     if (config.database) {
       connectionKeywords.push(`CurrentSchema=${config.database}`);
--- a/src/dialects/ibmi/connection-manager.js
+++ b/src/dialects/ibmi/connection-manager.js
@@ -38,10 +38,10 @@ export class IBMiConnectionManager extends AbstractConnectionManager {
     //  if (config.database) {
     //    connectionKeywords.push(`DSN=${config.database}`);
     //  }
-
-    if (config.DSN) {
-      connectionKeywords.push(`DSN=${config.DSN}`);
-    }
+  
+    if (config.dialectOptions && config.dialectOptions.DSN){
+      connectionKeywords.push(`DSN=${config.dialectOptions.DSN}`);
+     }
 
     if (config.database) {
       connectionKeywords.push(`CurrentSchema=${config.database}`);
--- a/src/dialects/ibmi/index.js
+++ b/src/dialects/ibmi/index.js
@@ -39,6 +39,10 @@ export class IBMiDialect extends AbstractDialect {
   constructor(sequelize) {
     super();
     this.sequelize = sequelize;
+
+    // Attach DSN configuation to the Sequelize object
+    this.sequelize.config.DSN = sequelize.options.DSN;
+
     this.connectionManager = new IBMiConnectionManager(this, sequelize);
     this.queryGenerator = new IBMiQueryGenerator({
       dialect: this,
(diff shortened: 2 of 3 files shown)


WikiRik:
Thanks for this PR! Can you point this PR towards `main` instead?

Tagging @markdirish in this as well since he worked on the implementation before

hancockm:
@WikiRik . Done

src/dialects/ibmi/connection-manager.js
```diff
@@ -35,8 +35,18 @@ export class IBMiConnectionManager extends AbstractConnectionManager {
 
     // 'database' doesn't make sense in this context, but it is mapped here to
```
WikiRik, line 36: This comment needs to be updated

src/dialects/ibmi/connection-manager.js
```diff
@@ -35,8 +35,18 @@ export class IBMiConnectionManager extends AbstractConnectionManager {
...
     // DSN, which is a close fit
+    //  if (config.database) {
+    //    connectionKeywords.push(`DSN=${config.database}`);
+    //  }
+
```
WikiRik, line 41:
```suggestion
```

src/dialects/ibmi/query-generator.js
```diff
@@ -130,21 +130,27 @@ export class IBMiQueryGenerator extends AbstractQueryGenerator {
   }
 
   dropTableQuery(tableName, options) {
-    let table = tableName;
+
```
WikiRik, line 133:
```suggestion
```

WikiRik (commented):
Few style comments, but I do not really know anything about the dialect internally so if @markdirish says it's good I will agree with him. Looking forward to see the documentation update as well

src/dialects/ibmi/connection-manager.js
hancockm: Yes, I agree.

src/dialects/ibmi/index.js
```diff
@@ -39,6 +39,10 @@ export class IBMiDialect extends AbstractDialect {
   constructor(sequelize) {
     super();
     this.sequelize = sequelize;
+
+    // Attach DSN configuation to the Sequelize object
+    this.sequelize.config.DSN = sequelize.options.DSN;
```
ephys, line 44: Instanciating the dialect shouldn't cause any side effect. If DSN is an option specific to ibmi, I think it should go in `dialectOptions` instead.

src/dialects/ibmi/query-generator.js
```diff
@@ -130,21 +130,26 @@ export class IBMiQueryGenerator extends AbstractQueryGenerator {
...
-    if (typeof table === 'object') {
-      schema = table.schema || undefined;
-      table = table.table;
-    } else if (options.schema) {
-      schema = options.schema;
-    }
 
-    return `DROP TABLE IF EXISTS ${schema ? `"${schema}".` : ''}"${table}"`;
+     if (typeof tableName === 'object') {
+      schema = tableName.schema || options.schema || undefined;
```
ephys, line 138: no other dialect accepts specifying the schema as part of the option bag, it's always part of `tableName`. You can drop `options` here.

(Not an issue for `describeTableQuery`, but we'll eventually need to drop the second parameter there too to be consistent)

ephys (changes requested):

src/dialects/ibmi/query-generator.js
WikiRik: I've gone ahead and fixed this in #15235

src/dialects/ibmi/query-generator.js
```diff
@@ -130,21 +130,25 @@ export class IBMiQueryGenerator extends AbstractQueryGenerator {
...
+      tableName = tableName.table;
     }
 
-    return `DROP TABLE IF EXISTS ${schema ? `"${schema}".` : ''}"${table}"`;
+    return `DROP TABLE IF EXISTS ${schema ? `"${schema}".` : ''}"${tableName}"`;
   }
 
   describeTableQuery(tableName, schema) {
 
+    if (typeof tableName === 'object') {
```
WikiRik, line 146: This is now also fixed in #15235

markdirish:
Finally got around to taking a look at this. A few thoughts:

1. Are you sure this is working as you expect? You map database to a `currentSchema` connection string keyword, but I don't see that as a valid keyword in our documentation: https://www.ibm.com/docs/en/i/7.5?topic=details-connection-string-keywords.  The closest thing appears to be `DBQ`, which specifies the libraries to add/replace values in the server job's library list. You say  that this fixed issues you had with `sequelize-automate`, but I'm not sure how if that keyword isn't valid. It might just be ignoring it?

2. If DBQ is the same behavior you are going for, we need to think about how to pass in the value in database. Note that in the documentation above, there are four different examples that all have different behavior:

> If UID=myuser;DBQ=mylib,mylib2,mylib3 is specified, then CURRENT SCHEMA special register is set to MYLIB and the library list would be MYLIB, MYLIB2, and MYLIB3.
> If UID=myuser;DBQ=,mylib,mylib2,mylib3;NAM=1 is specified, then CURRENT SCHEMA special register is set to *LIBL and the library list would be MYLIB, MYLIB2, and MYLIB3.
> If UID=myuser;DBQ=,mylib,mylib2,mylib3;NAM=0 is specified, then CURRENT SCHEMA special register is set to MYUSER and the library list would be MYLIB, MYLIB2, and MYLIB3.
> If UID=myuser;DBQ=mylib,*USRLIBL,mylib2 is specified, then CURRENT SCHEMA special register is set to MYLIB and the library list would be MYLIB, <existing user portion of the library list>, and MYLIB2.

Will run this through the sequelize integration tests to see what comes out the other end.

WikiRik:
Closing this due to inactivity and the comment above which seems to suggest this change is invalid

Issue #11415, opened 2019-09-10
Incorrect ordering for a findAll operation on a scoped model with predefined order


Jigar-Freightwalla:
### What are you doing?

I have a Model `TheModel` with a scope `scoped` containing `order` , say `order1`.
On invoking a findAll with an `order`, say `order2`, the resulting order is `order1`, `order2`

```
// MINIMAL, SELF-CONTAINED code here (SSCCE/MCVE/reprex)
TheModel.addScope('scoped', {
  order: [
    [associatedModel1, 'foo', 'ASC'],
    [associatedModel2, 'bar', 'ASC'],
  ]
});

scopedModel.findAll({
  order: [
    ['field1', 'ASC'],
    [associatedModel1, 'foo', 'ASC'],
  ]
});
```

### What do you expect to happen?

Resulting order: `field1` `ASC`, `associatedModel1`.`foo` `ASC`
(This was the actual order using sequelize version 3.30.4)

### What is actually happening?

Resulting order: `associatedModel1`.`foo` `ASC`, `associatedModel2`.`bar` `ASC`, `field1` `ASC`, `associatedModel1`.`foo` `ASC`

### Environment

- Sequelize version: 5.15.2
- Node.js version: 6.13.4
- Operating System: XXX
- If TypeScript related: TypeScript version: XXX

### How does this problem relate to dialects?

- [ ] I think this problem happens regardless of the dialect.
- [ ] I think this problem happens only for the following dialect(s): 
- [ ] I don't know, I was using PUT-YOUR-DIALECT-HERE, with connector library version XXX and database version XXX

### Would you be willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time and I know how to start.
- [ ] Yes, I have the time but I don't know how to start, I would need guidance.
- [ ] No, I don't have the time, although I believe I could do it if I had the time...
- [ ] No, I don't have the time and I wouldn't even know how to start.

Note: I was able to come to this [line](https://github.com/sequelize/sequelize/blob/d5e4a10f4e8dc1e37fe46678f04bbfa66b1299cc/lib/model.js#L798) which is merging the two orders. However, I think the community needs to collectively decide what should be the expected behaviour. I think that the order provided by `findAll` should always override the scope order.

papb:
[The docs](https://sequelize.org/master/manual/scopes.html#merging) claim that

[...]

Issue #10914, opened 2019-05-08
order in defaultScope merging rather than overwritten


disyam:
## What are you doing?
Docs say "The limit, offset, order, paranoid, lock and raw fields are overwritten, while where is shallowly merged (meaning that identical keys will be overwritten). The merge strategy for include will be discussed later on."
but when i user order in defaultScope, and i use order in my query, order is merging not overwritten

when using this model
```js
module.exports = (sequelize, DataTypes) => {
  const organizationMember = sequelize.define(
    "organizationMember",
    {
      id: {
        type: DataTypes.UUID,
        primaryKey: true,
        defaultValue: DataTypes.UUIDV4
      },
      organizationId: {
        type: DataTypes.UUID,
        allowNull: false
      },
      accountId: {
        type: DataTypes.UUID,
        allowNull: false
      },
      type: {
        type: DataTypes.STRING,
        allowNull: false
      },
      deletedAt: {
        type: DataTypes.DATE,
        allowNull: true
      }
    },
    {
      defaultScope: {
        order: [["createdAt", "ASC"]],
        where: { deletedAt: null }
      }
    }
  );
  return organizationMember;
};
```

then i query with
```js
const organizationMember = await OrganizationMember.findAll({
  limit: 10,
  offset: 0,
  order: [["createdAt", "DESC"]],
  where: { organizationId: "97b5d956-c740-4100-ad1a-ea2a1364f9df" }
});
```

## What do you expect to happen?
this sql query
```
SELECT "id", "organizationId", "accountId", "type", "deletedAt", "createdAt", "updatedAt" FROM "organizationMembers" AS "organizationMember" WHERE "organizationMember"."deletedAt" IS NULL AND "organizationMember"."organizationId" = '97b5d956-c740-4100-ad1a-ea2a1364f9df' ORDER BY "organizationMember"."createdAt" DESC LIMIT 10 OFFSET 0;
```

## What is actually happening?
but i got this query
```
SELECT "id", "organizationId", "accountId", "type", "deletedAt", "createdAt", "updatedAt" FROM "organizationMembers" AS "organizationMember" WHERE "organizationMember"."deletedAt" IS NULL AND "organizationMember"."organizationId" = '97b5d956-c740-4100-ad1a-ea2a1364f9df' ORDER BY "organizationMember"."createdAt" ASC, "organizationMember"."createdAt" DESC LIMIT 10 OFFSET 0;
```

## Environment
Dialect:
- [ ] mysql
[...]

PR #15230, opened 2022-11-04
fix: override order problem with default scope

DanielOverdevest:
## Pull Request Checklist

## Description Of Change

Make it possible to override defaultScope order. At the moment it's being added and that can lead to not overridable order bys or contradictory ASC and DESC of same field.

Fix this issue: #11415 & #10914

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -741,9 +741,10 @@ ${associationOwner._getAssociationDebugList()}`);
 
     if (Array.isArray(objValue) && Array.isArray(srcValue)) {
       if (key === 'order') {
-        return _.union(objValue, srcValue);
+        return srcValue;
       }
-      return srcValue;
+
+      return _.union(objValue, srcValue);
     }
 
     if (['where', 'having'].includes(key)) {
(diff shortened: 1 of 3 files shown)


WikiRik (changes requested):
Do not update sscce.js for new tests, but add it to https://github.com/sequelize/sequelize/blob/main/test/integration/model/scope/find.test.js

DanielOverdevest:
> Do not update sscce.js for new tests, but add it to https://github.com/sequelize/sequelize/blob/main/test/integration/model/scope/find.test.js

Consider it done. Can you review it?

ephys (changes requested):
I think this makes more sense than the current behavior

Let's keep in mind that this is a breaking change we'll need to document in our upgrade guide

Only note: Right now you've written an integration test. Can you write it as a unit test instead? We aim to move tests that don't need to access the database to our unit tests to speed up our test suites

The unit test could go here:

https://github.com/sequelize/sequelize/blob/main/test/unit/model/scope.test.ts#L749

Instead of checking what the database returns, we just need to check what `Model._injectScope` produces

PR #15231, opened 2022-11-04
fix: remove options.model overwrite on bulkUpdate

### LINKED ISSUE #11364: Update maps returned instance to the wrong model
## Issue Description

When multiple models are wired up to the same table. The model's static update method can incorrectly map the returned instance to the wrong model

### What are you doing?
```
class Foo extends Model {}
Foo.init({ /* bla */ }, {
  tableName: 'table_with_discriminator',
  sequelize
})

class Bar extends Model {}
Bar.init({ /* bla */ }, {
  tableName: 'table_with_discriminator',
  sequelize
})


const updates = await Bar.update(
      { prop: 'value' },
      {
        where: { id: 'xxx },
        returning: true
      }
    );
const updatedInstance = updates[1][0]; // this is an instance of Foo
```

The updated instance should be an instance of Bar

The updated instance is an instance of Foo

### Additional context
The root cause if this line:
https://github.com/sequelize/sequelize/blob/master/lib/query-interface.js#L1040
Which is called from the update method here:
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L3219

The returned instances are being mapped to a model based on the underlying table name.  The instance should be mapped to the same model used to call the update method

### Environment

- Sequelize version: 5.16.0
- Node.js version: 10.15.3
- Operating System: Mac

BigsonLvrocha:
## Pull Request Checklist

## Description Of Change

Closes #11364

This is a way that keeps the current behaviour of bulkUpdate while fixing the unexpected behaviour when calling `.update` when there are 2 or more models defined for the same table 

## Todos

- [x] #1 add test


ephys (approved):
This looks good to me

test/unit/model/update.test.js
```diff
@@ -53,4 +53,56 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+      }, {
+        tableName: 'model_table',
+      });
+
+      this.Model2 = current.define('Model2', {
+        value: DataTypes.INTEGER,
+        name: DataTypes.STRING,
+      }, {
+        tableName: 'model_table',
+        globalScope: {
```
ephys, line 73: just one note though: we don't have a globalScope option. Did you mean defaultScope? Either way it's not stricly necessary for this test :)

ephys (changes requested):

test/unit/model/update.test.js
BigsonLvrocha: oops, my mistake, it's there to make an example of an use case.
fixed in https://github.com/sequelize/sequelize/pull/15231/commits/af63e3d6a5dbb47d31d1ebabd994fec75aea211d

test/unit/model/update.test.js
```diff
@@ -53,4 +53,56 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
+      this.Model2 = current.define('Model2', {
+        value: DataTypes.INTEGER,
+        name: DataTypes.STRING,
+      }, {
+        tableName: 'model_table',
+        defaultScope: {
+          where: {
+            isModel2: true,
+          },
+        },
```
WikiRik, line 77:
```suggestion
```

BigsonLvrocha:
sorry for the span on request, I did not know only one review was requested per pr

WikiRik (approved):

WikiRik:
Thanks for this PR!

BigsonLvrocha:
Will this fix go into v6?
Can it go in there too?

BigsonLvrocha:
@WikiRik 
Will this change go into v6 too?
What should I do for the fix to get there?

WikiRik:
Good question! Just open a PR just like this one, but into the `v6` branch. We'll review it and then it will land in the next v6 release

PR #15235, opened 2022-11-05
fix: unify and fix tests `queryGenerator.arithmeticQuery`/`dropTableQuery`/`showIndexesQuery`/`describeTableQuery`

WikiRik:
## Pull Request Checklist

## Description Of Change

This PR unifies and fixes the following `queryGenerator` functions;
- arithmeticQuery
- dropTableQuery
- showIndexesQuery (added a FIXME for specifying the schema in the options, will be done in #14687)
- describeTableQuery (added a FIXME for specifying the schema in the options, will be done in #14687)

## Merge Squash Commit Message Body (for breaking changes)

```

BREAKING CHANGE: **ibmi** - dropTableQuery does not accept schema in the options anymore, use a tableName object
```


test/unit/dialects/mariadb/query-generator.test.js
```diff
@@ -680,34 +646,6 @@ if (dialect === 'mariadb') {
...
-        // FIXME: enable this test once fixed (in https://github.com/sequelize/sequelize/pull/14687)
-        // {
-        //   sequelizeOptions: {
-        //     schema: 'schema',
-        //   },
-        //   arguments: ['User'],
-        //   expectation: 'SHOW INDEX FROM `schema`.`User`',
-        // },
-        {
-          arguments: ['User', { database: 'sequelize' }],
```
WikiRik, line 701: I've removed this test for mariadb/mysql since it's not supported by the typings anymore

src/dialects/ibmi/query-generator.js
```diff
@@ -130,17 +136,17 @@ export class IBMiQueryGenerator extends AbstractQueryGenerator {
   }
 
   dropTableQuery(tableName, options) {
```
WikiRik, line 138: This is the just like the abstract `dropTableQuery` but the only difference is that there is no semicolon at the end. Not sure if this was the correct option or that we should update the tests

src/dialects/snowflake/query-generator.d.ts
```diff
@@ -1,3 +1,5 @@
 import { AbstractQueryGenerator } from '../abstract/query-generator.js';
 
-export class SnowflakeQueryGenerator extends AbstractQueryGenerator {}
+export class SnowflakeQueryGenerator extends AbstractQueryGenerator {
+  showIndexesQuery(): string;
```
WikiRik, line 4: Is this the correct way to overwrite the typings from `AbstractQueryGenerator` or does this now allow both?

test/unit/dialects/postgres/query-generator.test.js
```diff
@@ -223,47 +185,6 @@ if (dialect.startsWith('postgres')) {
...
-        {
-          arguments: ['myTable', { cascade: true }],
-          expectation: 'DROP TABLE IF EXISTS "myTable" CASCADE;',
-        },
-        {
-          arguments: [{ tableName: 'myTable', schema: 'mySchema' }, { cascade: true }],
-          expectation: 'DROP TABLE IF EXISTS "mySchema"."myTable" CASCADE;',
-        },
-
-        // Variants when quoteIdentifiers is false
```
WikiRik, line 244: I've not added tests for these since this function is being tested elsewhere and probably will be removed in a future PR. See https://github.com/sequelize/sequelize/pull/14999#discussion_r974055261

test/unit/dialects/sqlite/query-generator.test.js
```diff
@@ -13,42 +13,7 @@ const { SqliteQueryGenerator: QueryGenerator } = require('@sequelize/core/_non-s
 
 if (dialect === 'sqlite') {
   describe('[SQLITE Specific] QueryGenerator', () => {
-    beforeEach(function () {
```
WikiRik, line 16: I don't think this was used anymore so I removed it

src/dialects/abstract/query-interface.js
```diff
@@ -277,7 +277,14 @@ export class QueryInterface {
...
 
-    const sql = this.queryGenerator.dropTableQuery(tableName, options);
+    let sql;
+
+    // TODO: this should be refactored to probably use DialectSupports
+    if (this.queryGenerator.dialect.name === 'postgres') {
+      sql = this.queryGenerator.dropTableQuery(tableName, options);
+    } else {
+      sql = this.queryGenerator.dropTableQuery(tableName);
+    }
```
WikiRik, line 287: So this is messy and definitely should not be merged like this, but I'm unsure what the proper course of action is. Add something like `dialect.supports.drop.cascade`?
And what should we do with `DROP_TABLE_QUERY_SUPPORTABLE_OPTIONS`? Remove that and incorporate `dialect.supports.drop` into `rejectInvalidOptions`? Or keep both?

src/dialects/abstract/query-generator.d.ts
```diff
@@ -144,6 +149,10 @@ export class AbstractQueryGenerator {
     options?: ArithmeticQueryOptions,
   ): string;
 
+  showIndexesQuery(tableName: TableName): string;
+
+  dropTableQuery(tableName: TableName, options?: DropTableQueryOptions): string;
```
WikiRik, line 154: Do we want to keep this here as is or update to `dropTableQuery(tableName: TableName): string;`? Since the only option is `cascade` and that's only supported in postgres so we can also move that to the `PostgresQueryGenerator`.

src/dialects/abstract/query-generator.d.ts
```diff
@@ -166,6 +173,11 @@ export class AbstractQueryGenerator {
...
+  describeTableQuery(tableName: TableName, schema?: string, schemaDelimiter?: string): string;
```
WikiRik, line 179: `schemaDelimiter` is only supported in sqlite, do we want to keep this or drop this option?

src/dialects/abstract/query-generator.d.ts
ephys: I think it's better if all options are declared in abstract, that's the only type users are going to get when accessing `sequelize.queryGenerator`. Then each dialect can use `rejectInvalidOptions` to let the user know whether an option is actually supporter or not

src/dialects/abstract/query-generator.d.ts
ephys: The current design of `describeTableQuery` makes no sense to me. `TableName` is already capable of specifying a schema according to its definition:

```ts
export interface TableNameWithSchema {
  tableName: string;
  schema?: string;
  delimiter?: string;
  as?: string;
  name?: string;
}
export type TableName = string | TableNameWithSchema;
```

So this method doesn't need extra parameters, just a single `TableName` one:

```ts
describeTableQuery('users')
describeTableQuery({ tableName: 'users', schema: 'public' })
```

This one is inconsistent and should be removed (breaking change)

```ts
describeTableQuery('users', 'public')
```

src/dialects/abstract/query-interface.js
ephys: I vote for `dialect.supports.drop.cascade`

src/dialects/db2/query-generator.js
```diff
@@ -180,6 +181,11 @@ export class Db2QueryGenerator extends AbstractQueryGenerator {
   }
 
   describeTableQuery(tableName, schema) {
+    if (typeof tableName === 'object') {
```
ephys, line 184: this could call the updated `extractTableDetails` function instead, but it's stuck inside of the changeColumns PR https://github.com/sequelize/sequelize/pull/14687/files#diff-54c14e4744281a47d62196b5892c4004fec6e2f3c4d90f444aa3bd0a880f2209R88

src/dialects/ibmi/query-generator.js
ephys: ibmi doesn't want semicolons at the end of its queries but it could use lodash's `trimEnd(sql, ';')` instead

this would be a huge amount of work but we could also decide to remove all semicolons at the end of queries. Would certainly make using `queryGenerator` as a way to build subQueries easier

src/dialects/snowflake/query-generator.d.ts
ephys: Not sure you can change the signature

Does snowflake even implement `showIndexesQuery`? Its current implementation is `SELECT '' FROM DUAL` which is very strange.

Looks like not: https://dataedo.com/kb/query/snowflake/list-unique-keys-and-indexes-in-the-database

This is more a snowflake implem problem than a typing issue

test/unit/dialects/mariadb/query-generator.test.js
ephys: Yes it's testing something that's been removed (as per the comment), can be removed altogether

test/unit/dialects/postgres/query-generator.test.js
ephys: We should consider writing a PR to print a deprecation warning if `quoteIdentifiers` is set to false

test/unit/query-generator/arithmetic-query.test.ts
```diff
@@ -8,6 +8,86 @@ describe('QueryGenerator#arithmeticQuery', () => {
     firstName: DataTypes.STRING,
   }, { timestamps: false });
 
+  it('should use the plus operator', async () => {
+    const sql = queryGenerator.arithmeticQuery(
+      '+', 'myTable', {}, { foo: literal('bar') }, {}, {},
```
ephys, line 13: arithmeticQuery really needs a good redesign, these options are incomprehensible (unrelated to this PR of course)

test/unit/query-generator/arithmetic-query.test.ts
```diff
@@ -8,6 +8,86 @@ describe('QueryGenerator#arithmeticQuery', () => {
     firstName: DataTypes.STRING,
   }, { timestamps: false });
 
+  it('should use the plus operator', async () => {
```
ephys, line 11:
```suggestion
  it('uses the specified operator', async () => {
```

test/unit/query-generator/arithmetic-query.test.ts
```diff
@@ -8,6 +8,86 @@ describe('QueryGenerator#arithmeticQuery', () => {
...
+    expectsql(sql, {
+      default: `UPDATE [myTable] SET [foo]=[foo]+ bar`,
+      postgres: `UPDATE "myTable" SET "foo"="foo"+ bar RETURNING *`,
+      mssql: `UPDATE [myTable] SET [foo]=[foo]+ bar OUTPUT INSERTED.*`,
+    });
+  });
+
+  it('should use the plus operator with where clause', async () => {
+    const sql = queryGenerator.arithmeticQuery(
+      '+', 'myTable', { bar: 'biz' }, { foo: literal('bar') }, {}, {},
```
ephys, line 25: testing support for `literal` should be its own test

test/unit/query-generator/arithmetic-query.test.ts
```diff
@@ -8,6 +8,86 @@ describe('QueryGenerator#arithmeticQuery', () => {
...
+    );
+
+    expectsql(sql, {
+      default: `UPDATE [myTable] SET [foo]=[foo]+ bar WHERE [bar] = 'biz'`,
+      postgres: `UPDATE "myTable" SET "foo"="foo"+ bar WHERE "bar" = 'biz' RETURNING *`,
+      mssql: `UPDATE [myTable] SET [foo]=[foo]+ bar OUTPUT INSERTED.* WHERE [bar] = N'biz'`,
+    });
+  });
+
+  it('should use the plus operator without returning clause', async () => {
```
ephys, line 35:
```suggestion
  it('supports omitting the RETURNING clause', async () => {
```

test/unit/query-generator/arithmetic-query.test.ts
```diff
@@ -8,6 +8,86 @@ describe('QueryGenerator#arithmeticQuery', () => {
...
+    const sql = queryGenerator.arithmeticQuery(
+      '+', 'myTable', {}, { foo: literal('bar') }, {}, { returning: false },
+    );
+
+    expectsql(sql, {
+      default: `UPDATE [myTable] SET [foo]=[foo]+ bar`,
+    });
+  });
+
+  it('should use the minus operator', async () => {
```
ephys, line 45: should be merged with "uses the specified operator" test as a second assertion

test/unit/query-generator/arithmetic-query.test.ts
```diff
@@ -8,6 +8,86 @@ describe('QueryGenerator#arithmeticQuery', () => {
...
+    );
+
+    expectsql(sql, {
+      default: `UPDATE [myTable] SET [foo]=[foo]- bar`,
+      postgres: `UPDATE "myTable" SET "foo"="foo"- bar RETURNING *`,
+      mssql: `UPDATE [myTable] SET [foo]=[foo]- bar OUTPUT INSERTED.*`,
+    });
+  });
+
+  it('should use the minus operator with negative value', async () => {
```
ephys, line 57:
```suggestion
  it('supports negative values', async () => {
```

test/unit/query-generator/arithmetic-query.test.ts
```diff
@@ -8,6 +8,86 @@ describe('QueryGenerator#arithmeticQuery', () => {
...
+    );
+
+    expectsql(sql, {
+      default: `UPDATE [myTable] SET [foo]=[foo]- -1`,
+      postgres: `UPDATE "myTable" SET "foo"="foo"- -1 RETURNING *`,
+      mssql: `UPDATE [myTable] SET [foo]=[foo]- -1 OUTPUT INSERTED.*`,
+    });
+  });
+
+  it('should use the minus operator with where clause', async () => {
```
ephys, line 69: this test is redundant

test/unit/query-generator/arithmetic-query.test.ts
```diff
@@ -8,6 +8,86 @@ describe('QueryGenerator#arithmeticQuery', () => {
...
+    );
+
+    expectsql(sql, {
+      default: `UPDATE [myTable] SET [foo]=[foo]- bar WHERE [bar] = 'biz'`,
+      postgres: `UPDATE "myTable" SET "foo"="foo"- bar WHERE "bar" = 'biz' RETURNING *`,
+      mssql: `UPDATE [myTable] SET [foo]=[foo]- bar OUTPUT INSERTED.* WHERE [bar] = N'biz'`,
+    });
+  });
+
+  it('should use the minus operator without returning clause', async () => {
```
ephys, line 81: also redundant

test/unit/query-generator/describe-table-query.test.ts
```diff
@@ -0,0 +1,420 @@
+import { createSequelizeInstance, expectsql, getTestDialect, sequelize } from '../../support';
+
+const dialectName = getTestDialect();
+
+describe('QueryGenerator#describeTableQuery', () => {
+  const queryGenerator = sequelize.getQueryInterface().queryGenerator;
+
+  it('produces a DESCRIBE TABLE query', () => {
```
ephys, line 8: `DESCRIBE TABLE` makes it sound like there is such a SQL clause when it's actually reading the system tables

```suggestion
  it('produces a query to describe a table', () => {
```

test/unit/query-generator/describe-table-query.test.ts
```diff
@@ -0,0 +1,420 @@
...
+          LEFT JOIN
+            QSYS2.SYSCST
+          ON
+            QSYS2.SYSCSTCOL.CONSTRAINT_NAME = QSYS2.SYSCST.CONSTRAINT_NAME
+          WHERE QSYS2.SYSCOLUMNS.TABLE_SCHEMA = 'mySchema' AND QSYS2.SYSCOLUMNS.TABLE_NAME = 'myTable'`,
+    });
+  });
+
+  it('produces a DESCRIBE TABLE query with schema argument', () => {
+    expectsql(() => queryGenerator.describeTableQuery('myTable', 'mySchema'), {
```
ephys, line 171: we should drop this feature, it's redundant (breaking change)

test/unit/query-generator/describe-table-query.test.ts
```diff
@@ -0,0 +1,420 @@
...
+        LEFT JOIN
+          QSYS2.SYSCST
+        ON
+          QSYS2.SYSCSTCOL.CONSTRAINT_NAME = QSYS2.SYSCST.CONSTRAINT_NAME
+        WHERE QSYS2.SYSCOLUMNS.TABLE_SCHEMA = 'mySchema' AND QSYS2.SYSCOLUMNS.TABLE_NAME = 'myTable'`,
+    });
+  });
+
+  it('produces a DESCRIBE TABLE query with schema and schemaDelimiter argument', () => {
+    expectsql(() => queryGenerator.describeTableQuery('myTable', 'mySchema', '.'), {
```
ephys, line 341: may want to use a different schema delimiter since `'.'` is already the default

This test is only relevant in dialects that don't support schemas

test/unit/query-generator/drop-table-query.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  it('produces a DROP TABLE query with cascade', () => {
+    expectsql(() => queryGenerator.dropTableQuery('myTable', { cascade: true }), {
+      default: buildInvalidOptionReceivedError('dropTableQuery', dialectName, ['cascade']),
+      postgres: `DROP TABLE IF EXISTS "myTable" CASCADE;`,
+    });
+  });
+
+  it('produces a DROP TABLE query with schema', () => {
```
ephys, line 23: could add a test with a global schema option set

test/unit/query-generator/show-indexes-query.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        + `QSYS2.SYSCST.TABLE_NAME from QSYS2.SYSCSTCOL left outer join QSYS2.SYSCST on QSYS2.SYSCSTCOL.TABLE_SCHEMA = QSYS2.SYSCST.TABLE_SCHEMA and `
+        + `QSYS2.SYSCSTCOL.TABLE_NAME = QSYS2.SYSCST.TABLE_NAME and QSYS2.SYSCSTCOL.CONSTRAINT_NAME = QSYS2.SYSCST.CONSTRAINT_NAME where `
+        + `QSYS2.SYSCSTCOL.TABLE_SCHEMA = 'mySchema' and QSYS2.SYSCSTCOL.TABLE_NAME = 'myTable' union select QSYS2.SYSKEYS.INDEX_NAME AS NAME, `
+        + `QSYS2.SYSKEYS.COLUMN_NAME, CAST('INDEX' AS VARCHAR(11)), QSYS2.SYSINDEXES.TABLE_SCHEMA, QSYS2.SYSINDEXES.TABLE_NAME from QSYS2.SYSKEYS `
+        + `left outer join QSYS2.SYSINDEXES on QSYS2.SYSKEYS.INDEX_NAME = QSYS2.SYSINDEXES.INDEX_NAME where QSYS2.SYSINDEXES.TABLE_SCHEMA = 'mySchema' `
+        + `and QSYS2.SYSINDEXES.TABLE_NAME = 'myTable'`,
+    });
+  });
+
+  it('produces a SHOW INDEX query from a table and schema in options', () => {
```
ephys, line 53: IMO this phrasing makes it clearer that it's not an option of the `showIndexesQuery` method but of the Sequelize constructor (other phrasings welcome)

```suggestion
  it('produces a SHOW INDEX query from a table and the Sequelize schema options', () => {
```

ephys:
I'm looking forward to seeing this merged. It's a good step towards cleaning up our APIs!

src/dialects/ibmi/query-generator.js
WikiRik: A global way to add semicolons at the end of a query would be nice, but not really something I think we want to work on in the short term

src/dialects/snowflake/query-generator.d.ts
WikiRik: Yeah, I found the implementation strange but there is no way for me to really test it so I didn't doubt it for now. I can add a TODO to the implementation? And I'll just remove this signature then

test/unit/dialects/postgres/query-generator.test.js
WikiRik: For v6? Or deprecate in v7 and remove in v8?

src/dialects/abstract/query-generator.d.ts
WikiRik: Yeah, I wasn't sure if users can use typings of dialect specific query generators. But if that's more difficult I would agree that `rejectInvalidOptions` is the way to go

src/dialects/abstract/query-generator.d.ts
WikiRik: Yeah, that's something I noticed yesterday when we discussed the `extractTableDetails` issue. I think I'll split parts of this PR to another where we use the `TableName` type consistenly and not allow schema/delimiter to be passed seperately.

src/dialects/db2/query-generator.js
WikiRik: Yeah, I took a quick look at that yesterday but came across some failing tests. Will look into this

test/unit/query-generator/arithmetic-query.test.ts
WikiRik: Do you want me to make a TODO for that in the definition?

test/unit/query-generator/describe-table-query.test.ts
WikiRik: Yeah, I tried that and came across the fact that only sqlite does something with this

src/dialects/snowflake/query-generator.d.ts
ephys: Sure, TODO is fine. Especially since we can't test it

test/unit/dialects/postgres/query-generator.test.js
ephys: I'd say deprecate in v7

src/dialects/abstract/query-generator.d.ts
ephys: Sounds good!

test/unit/query-generator/arithmetic-query.test.ts
ephys: No we'll get to it when we migrate it to TS I think

test/unit/query-generator/describe-table-query.test.ts
ephys: Yup, it's the only one that doesn't support schemas

Honestly I think we should, at some point, drop this fake schema support.

WikiRik:
I might probably be missing something, but I think this PR should be good to go for now.
I'll update `describeTableQuery` with the new `extractTableDetails` in a different PR.

src/dialects/abstract/query-interface.js
ephys: You can do this:

```ts
options.cascade = options.cascade != null ? options.cascade
  : (options.force && dialect.supports.drop.cascade) ? true
  : undefined;
```

that way cascade is only set if it's supported

WikiRik: I think it should be fixed in https://github.com/sequelize/sequelize/pull/15235/commits/7d3099ddf03d961c8c5fe311d3a72a2d89c5db9e

src/dialects/ibmi/query-generator.js
```diff
@@ -138,20 +142,24 @@ export class IBMiQueryGenerator extends AbstractQueryGenerator {
...
-    let table = tableName;
-    let schema;
-
-    if (typeof table === 'object') {
-      schema = table.schema || undefined;
-      table = table.table;
-    } else if (options.schema) {
-      schema = options.schema;
```
ephys, line 148: don't forget to document this in the PR's breaking changes (don't forget CC requires a blank line at the start of the commit body)

test/unit/query-generator/arithmetic-query.test.ts
```diff
@@ -8,6 +8,74 @@ describe('QueryGenerator#arithmeticQuery', () => {
...
+    const sql = queryGenerator.arithmeticQuery(
+      '+', 'myTable', {}, { foo: 3 }, {}, { returning: false },
+    );
+
+    expectsql(sql, {
+      default: `UPDATE [myTable] SET [foo]=[foo]+ 3`,
+    });
+  });
+
+  it('does not cause a syntax error when a minor operator is used with a negative value', async () => {
```
ephys, line 67:
```suggestion
  it('does not cause a syntax error when the minus operator is used with a negative value', async () => {
```

ephys (changes requested):

test/unit/query-generator/arithmetic-query.test.ts
ephys: Looks like minor comes from my own type in the previous review 😄

src/dialects/abstract/query-generator.js
```diff
@@ -150,7 +152,19 @@ export class AbstractQueryGenerator {
     return `DESCRIBE ${table};`;
   }
 
-  dropTableQuery(tableName) {
+  dropTableQuery(tableName, options) {
```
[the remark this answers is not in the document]
WikiRik: That would be nice!

src/dialects/ibmi/query-generator.js
WikiRik: Good one, added to the description

src/dialects/abstract/query-interface.js
```diff
@@ -289,7 +290,7 @@ export class QueryInterface {
     for (const tableName of tableNames) {
       // if tableName is not in the Array of tables names then don't drop it
       if (!skip.includes(tableName.tableName || tableName)) {
-        await this.dropTable(tableName, { ...options, cascade: true });
+        await this.dropTable(tableName, options);
```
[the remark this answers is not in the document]
WikiRik: Something like this? https://github.com/sequelize/sequelize/pull/15235/commits/2f17e6e381afd0246bc6e0bf01fadf48649918f6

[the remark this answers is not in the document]
ephys: No because with this version if the user calls `queryInterface.dropTable(table, { cascade: true })` on a dialect that doesn't support cascade, they won't be warned

It's this that needs changing:

```ts
// line 293, _dropAllTables
await this.dropTable(tableName, { cascade: this.queryGenerator.dialect.supports.dropTable.cascade ? true : undefined, ...options });
```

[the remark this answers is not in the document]
WikiRik: Wasn't sure how to do the unit test with most dialects expecting an error and one dialect to return a query but I think I've done it okay; https://github.com/sequelize/sequelize/pull/15235/commits/11939bceb72a3bf5db2804415e0bec5b649a350a

src/dialects/abstract/query-interface.js
```diff
@@ -276,7 +276,7 @@ export class QueryInterface {
    * @returns {Promise}
    */
   async dropTable(tableName, options = {}) {
-    options.cascade = (options.cascade && this.queryGenerator.dialect.supports.dropTable.cascade) ? options.cascade
+    options.cascade = options.cascade !== null ? options.cascade
```
ephys, line 279:
```suggestion
    options.cascade = options.cascade != null ? options.cascade
```

Otherwise it doesn't cover undefined and the default value will not be applied unless cascade is set explicitly to null

ephys (changes requested):

test/integration/sequelize/drop.test.ts
```diff
@@ -41,28 +41,4 @@ describe('Sequelize#drop', () => {
     // drop both tables
     await sequelize.drop();
   });
-
-  it('supports truncating cyclic associations with { cascade: true }', async () => {
```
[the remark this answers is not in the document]
WikiRik: I removed this since I didn't think it added too much to the suite (since we have an unit test), but I've added something similar back; https://github.com/sequelize/sequelize/pull/15235/commits/ccbdf1eaaedb58bf3c92732baf78e200f73f7c3e
But maybe I should include some assertions to ensure that the tables are actually dropped?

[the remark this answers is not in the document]
ephys: > But maybe I should include some assertions to ensure that the tables are actually dropped?

That would be good

Note that the test is there because `sequelize.drop` changes its behavior when the cascade option is set

ephys (approved):
Up to you if you want to improve the test or not

WikiRik:
We'll improve it in a future refactor, for now it should be fine.

PR #15236, opened 2022-11-05
meta: unify `jsonPathExtractionQuery` tests

WikiRik:
## Pull Request Checklist

## Description Of Change

This PR unifies the tests on `jsonPathExtractionQuery`

## Todos

- [ ] Decide what to do with the jsdoc
- [ ] Rename function to `_jsonPathExtractionQuery`?
- [ ] Remove `_isJson` typings in abstract and only add for postgres?
- [ ] Throw when `_isJson` is given but not supported?


src/dialects/abstract/query-generator.js
```diff
@@ -1216,11 +1216,11 @@ export class AbstractQueryGenerator {
    * @param   {string}               _column   The JSON column
    * @param   {string|Array<string>} [_path]   The path to extract (optional)
    * @param   {boolean}              [_isJson] The value is JSON use alt symbols (optional)
-   * @returns {string}                        The generated sql query
+   * @returns {Error|string}                   The generated sql query
```
ephys, line 1219: It never returns an error

```suggestion
   * @returns {string}                   The generated sql query
```

WikiRik:
If it's easier I can just revert the changes in the query-generator files btw. Unifying the tests and adding types for `jsonPathExtractionQuery` would already be enough to merge this imo.

test/unit/query-generator/json-path-extraction-query.test.ts
```diff
@@ -0,0 +1,39 @@
...
+      mysql: 'json_unquote(json_extract(`profile`,\'$.\\"id\\"\'))',
+      postgres: `("profile"#>>'{id}')`,
+      sqlite: 'json_extract(`profile`,\'$.id\')',
+    });
+  });
+
+  it('should use default handling if isJson is not passed', () => {
+    expectsql(() => queryGenerator.jsonPathExtractionQuery('profile', 'id'), {
+      default: notSupportedError,
+      mariadb: 'json_unquote(json_extract(`profile`,\'$.id\'))',
```
ephys, line 33: mariadb is inconsistent with mysql: `id` is not quoted. What happens if it includes a character such as `.`, `$`, `'`, `"`, or `,`?

Note: this comment doesn't mean you have to fix it here, it can be a TODO

test/unit/query-generator/json-path-extraction-query.test.ts
```diff
@@ -0,0 +1,39 @@
...
+      mysql: 'json_unquote(json_extract(`profile`,\'$.\\"id\\"\'))',
+      postgres: `("profile"#>>'{id}')`,
```
ephys, line 35: we have too few tests for the path here. What happens if one of the entries in the path includes characters such as `'`, `}`, `,`, or `{`?

`'{id}'` is just an array. IMO we need to drop the [custom array building we have here](https://github.com/sequelize/sequelize/blob/main/src/dialects/postgres/query-generator.js#L1038) and use the stringifier implemented in the Array DataType. Having two separate implementations is going to lead to bugs

(also doesn't need to be implemented in this PR, can be a TODO)

test/unit/query-generator/json-path-extraction-query.test.ts
```diff
@@ -0,0 +1,39 @@
...
+      sqlite: 'json_extract(`profile`,\'$.id\')',
+    });
+  });
+
+  it('should use default handling if isJson is not passed', () => {
+    expectsql(() => queryGenerator.jsonPathExtractionQuery('profile', 'id'), {
+      default: notSupportedError,
+      mariadb: 'json_unquote(json_extract(`profile`,\'$.id\'))',
+      mysql: 'json_unquote(json_extract(`profile`,\'$.\\"id\\"\'))',
```
ephys, line 34: same idea as mariadb

test/unit/query-generator/json-path-extraction-query.test.ts
```diff
@@ -0,0 +1,39 @@
...
+      postgres: `("profile"#>>'{id}')`,
+      sqlite: 'json_extract(`profile`,\'$.id\')',
```
ephys, line 36: same idea as mariadb

ephys (changes requested):

test/unit/query-generator/json-path-extraction-query.test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const dialectName = getTestDialect();
+
+const notSupportedError = new Error(`JSON operations are not supported in ${dialectName}.`);
+
+describe('QueryGenerator#jsonPathExtractionQuery', () => {
+  const queryGenerator = sequelize.getQueryInterface().queryGenerator;
+
+  it('should handle isJson parameter true', () => {
+    expectsql(() => queryGenerator.jsonPathExtractionQuery('profile', 'id', true), {
```
ephys, line 11: other test I just thought of that we can // TODO:

if profile starts and ends with the `` ` `` character, or the `"` character (which is perfectly valid, even if an edge case), I'm pretty sure it breaks right now

src/dialects/abstract/query-generator.d.ts
```diff
@@ -106,6 +106,11 @@ export class AbstractQueryGenerator {
     prepend?: boolean,
   ): string;
 
+  /**
+   * @private
+   */
+  jsonPathExtractionQuery(_column: string, _path?: string | string[], _isJson?: boolean): Error | string;
```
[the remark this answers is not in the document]
WikiRik: I'll try out the protected

The _ prefix was already there in the abstract definition, I just unified that but maybe that's not the nicest solution.

src/dialects/abstract/query-generator.js
WikiRik: Yeah, I wasn't too sure about this change when I looked at it again. I think my original thought was that if it's not implemented it throws an Error, but that's not what it returns

test/unit/query-generator/json-path-extraction-query.test.ts
WikiRik: Yeah, working on these tests did already show some insights but I didn't really add TODOs yet. Will add those

src/dialects/abstract/query-generator.d.ts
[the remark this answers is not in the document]
ephys: Because it's unused in abstract but not in other places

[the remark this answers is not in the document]
WikiRik: Protected kinda works, but we get the following error; `Property 'jsonPathExtractionQuery' is protected and only accessible within class 'AbstractQueryGenerator' and its subclasses.`
Should I make a new class for the tests where I extend this one? Or is there another easier solution?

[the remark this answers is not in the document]
ephys: Ah right we use it in tests

Let's keep your solution then

WikiRik:
@ephys I'll leave it up to you to resolve the conversations, but I think they should all be dealt with now

ephys (approved):

PR #15245, opened 2022-11-06
fix(postgres): reuse `searchPath` in `extractTableDetails`

WikiRik:
## Pull Request Checklist

## Description Of Change

This PR is just a simple recreation of #13045 but for `main`. Additional information can be found there

Closes #13045

### DIFF
--- a/src/dialects/abstract/query-generator.js
+++ b/src/dialects/abstract/query-generator.js
@@ -86,9 +86,10 @@ export class AbstractQueryGenerator {
   extractTableDetails(tableName, options) {
     options = options || {};
     tableName = tableName || {};
+    const schema = tableName.schema || options.schema || this.options.schema || options.schema !== false && this.sequelize.options.searchPath || 'public';
 
     return {
-      schema: tableName.schema || options.schema || this.options.schema || 'public',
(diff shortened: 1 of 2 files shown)


ephys:
I think this PR would be a good opportunity to extract https://github.com/sequelize/sequelize/pull/14687/files#diff-54c14e4744281a47d62196b5892c4004fec6e2f3c4d90f444aa3bd0a880f2209R83 which fixes many schema issues and has a different schema per dialect instead of 'public'

ephys (changes requested):

WikiRik:
I just copied that PR over to `main` since it was the only one that was open on v5 and seemed to pass the tests. But I will work on the schema issues as well, that will also help with some skipped tests in #15235

ephys:
Should we reject #13045 based on my above comment (plus the fact that it's targeting v5)?

WikiRik:
> Should we reject #13045 based on my above comment (plus the fact that it's targeting v5)?

Yes, let's close both and mention in the original PR that `searchPath` should not be used but we will fix the mentioned issue in an upcoming PR.

PR #15249, opened 2022-11-07
meta: remove code/test of unsupported versions

WikiRik:
## Pull Request Checklist

## Description Of Change

We had some additional code and tests for Node/database versions that we do not support in v7. This PR removes those. It does not update `defaultVersion` since those were higher than the removed code/tests.


test/integration/data-types/geometry.test.ts
```diff
@@ -86,25 +85,6 @@ describe(getTestDialectTeaser('DataTypes'), () => {
     });
 
     it('correctly parses an empty GEOMETRY field', async () => {
```
WikiRik, line 87: Okay, this test seems to be broken.

---

For `mariadb` it indeed returns `null`.

For `mysql` it returns `Invalid GIS data provided to function st_geometryfromtext.`

For `postgres oldest` it returns `Too few ordinates in GeoJSON`

For `postgres latest` it returns the following;
```
{
  "coordinates": [
    NaN
    NaN
  ]
  "crs": {
    "properties": {
      "name": "EPSG:4326"
    }
    "type": "name"
  }
  "type": "Point"
}
```

WikiRik:
Since the broken test is not in the scope of this PR I've skipped it and added a TODO

WikiRik:
This PR should make #15346 a bit easier as well

frostzt:
Ok this did make something come to my mind @ephys so in #15346 I actually changed initial version to `null`. The issue with this approach rises with tests especially unit tests that's why I did [this](https://github.com/sequelize/sequelize/pull/15346/files#diff-c791944fac0aba50f97b21713ea33d2f71679c834c9e87ffad79db3885874369). I don't know if that is a good approach!

ephys (approved):

ephys (approved):

frostzt (approved):

ephys (approved):

frostzt (approved):

PR #15268, opened 2022-11-10
meta: move used types deps to prod deps

WikiRik:
## Pull Request Checklist

## Description Of Change

This PR moves the types dependencies referenced in our `types` folder to the deps


ephys:
`debug`, `inflection` and `validator` are referenced by our `types` folder which is why they were moved to deps instead of devDeps (https://github.com/sequelize/sequelize/pull/14159)

The following packages are also referenced in types but are in devDeps instead of deps:

- ibm_db
- pg
- snowflake-sdk
- tedious

They probably should not be in devDeps since the project will error without them

Lodash, semver and uuid are not used in our types folder so can remain in dev deps.

WikiRik:
It seems like this is mostly a docs issue then. Because the types for the dialect connectors are not used outside of the dialect, so users should install it themselves (docs [here](https://sequelize.org/docs/v7/other-topics/typescript/#installation)). Maybe we can put them in `peerDependenciesMeta`?

ephys:
I don't know for sure. We can lazy load libraries, but we can't lazy load their type. I wouldn't be surprised if the currently unreleased version causes will cause ts errors in user projects unless they've installed all missing dialect types

WikiRik:
Okay, in that case I move the dialect types to the deps as well. We can see later if we can change this behaviour but for now I'd rather be on the safe side.

ephys:
We'll still need to test to make sure: some dialects, like mysql2, bundle their types and we don't really want to bundle mysql2

WikiRik:
Ah right. I forgot about that.

We could make a new `Upload install and build artifact` step in CI where we only install the deps and run a test file on that? Trying to see if we can import things properly? Not sure what exact file we would like to run

ephys:
It's tricky. Same issue with testing our TS import config. I'm not quite sure how to handle it automatically. I don't really know how to test it without doing a release either

WikiRik:
> The main blocking point is: how do we install a yet unreleased version of sequelize? Can yarn/npm emulate that properly? If yes we could have a local "test project" that just installs sequelize and tries to compile itself with ts

Yes, you can just [tag a branch or commit](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#dependencies)

ephys:
we can also try file:

WikiRik:
Ah yes, file works as well.

So we should add a step in the CI (after 'Upload install and build artifact') where we do the following;
- Remove installed dependencies (since we will use the built sequelize code from the previous step)
- Install some dependencies and use the built sequelize in a dedicated package.json for this test
- Try to execute a very basic test, like connecting to a sqlite instance.

Would that already be enough? Or do we need a dialect specific test as well?

ephys:
IMO no. No need to touch Sequelize itself during the test.

Have a folder in the repository that contains basic code, like connecting to a sqlite instance. That folder has its own package.json which points to sequelize using `file:` + sqlite3 + typescript. Run `yarn` in that project, then run `tsc --dryRun` in that project to ensure there are no typing issues (which is all we want to test)

WikiRik:
I think we're on the same track. It's just that we need sequelize to be compiled so we can point `file:` to the JS code instead of TS code, so therefore we need to start with the `Upload install and build artifact` step (or a different step that does the packaging in a separate artifact to be used by our 'test'). With that we can indeed make a small 'project' with basic code and run `yarn` and `tsc --dryRun`

ephys:
Removing myself from review request since it's back to draft to remove it from my TODO list but of course don't hesitate to request review again when needed

WikiRik:
Quickly going through this PR I see that the disucssion was about the types deps for the dialect adapters. Since we will extract those, this is not something anymore that we want to resolve in this PR

Issue #5252, opened 2016-01-18
sync() with schema option fails if a table has foreign key


bow-fujita:
I wonder if this is the same issue as #4898 though, the following test codes fails with sequelize 3.17.3:

```
// app.js

var argv = require('yargs')
      .usage('Usage: $0 -d dbname -u username -p password -s schema')
      .demand(['d', 'u', 'p', 's'])
      .alias('d', 'dbname')
      .alias('u', 'username')
      .alias('p', 'password')
      .alias('s', 'schema')
      .argv
  , Sequelize = require('sequelize')
  , sequelize = new Sequelize(argv.dbname, argv.username, argv.password, { dialect: 'postgres' })
  , schema = argv.schema
;

(function() {
  return sequelize.authenticate()
    .then(function() {
      return sequelize.createSchema(schema)
        .then(function() {
          var Role = sequelize.import('./models/role.js')
            , User = sequelize.import('./models/user.js')
          ;

          User.belongsTo(Role);

          return sequelize.sync({
              schema: schema
            , force: true
            })
            .catch(function(err) {
              sequelize.dropSchema(schema);
              throw err;
            });
        });
    });
})()
.then(function() {
  process.exit();
})
.catch(function(err) {
  console.error(err);
  process.exit(1);
});

// models/role.js
module.exports = function(sequelize, DataTypes) {
  return sequelize.define('Role', {
    name: {
      type: DataTypes.STRING
    , allowNull: false
    }
  });
};

// models/user.js
module.exports = function(sequelize, DataTypes) {
  return sequelize.define('User', {
    email: {
      type: DataTypes.STRING
    , validate: {
        isEmail: true
      }
    }
  , password: {
      type: DataTypes.STRING(32)
    , get: function() {
        return undefined;
      }
    }
  });
};
```

Error thrown as:

```
{ [SequelizeDatabaseError: relation "Roles" does not exist]
  name: 'SequelizeDatabaseError',
  message: 'relation "Roles" does not exist',
  parent:
   { [error: relation "Roles" does not exist]
     name: 'error',
     length: 96,
     severity: 'ERROR',
     code: '42P01',
     detail: undefined,
     hint: undefined,
     position: undefined,
     internalPosition: undefined,
     internalQuery: undefined,
     where: undefined,
     schema: undefined,
     table: undefined,
     column: undefined,
     dataType: undefined,
     constraint: undefined,
     file: 'namespace.c',
     line: '418',
[...]

PR #15274, opened 2022-11-11
fix: cleanup query generator

ephys:
## Description Of Change

This PR is yet-another preliminary PR for https://github.com/sequelize/sequelize/pull/14687 in order to prevent that other PR from growing endlessly.

The goal of this PR is to:

- Introduce `AbstractQueryGeneratorTypeScript` and migrate some base QueryGenerator methods to TypeScript
- Normalize which schema is used by introducing `getDefaultSchema`, and rewriting `extractTableDetails`. We should eventually arrive at a state where all query generator methods default the table schema to either the schema set through Sequelize's options, or the per-dialect default schema.
- change `quoteIdentifier` with one very important change:
  `queryGenerator.quoteIdentifier(`"abc"`)` previously *unquoted* then quoted its input. Quotes are actually valid as identifiers, as long as they're escaped. Instead of returning `"abc"`, this method will now return `"""abc"""` (triple " because one is the actual identifier delimiter, and two consecutive `""` is how we escape `"`).
- deprecate `addTicks`, `removeTicks`, and `TICK_CHAR` for the above reason

### DIFF
--- a/src/dialects/abstract/query-generator-typescript.ts
+++ b/src/dialects/abstract/query-generator-typescript.ts
@@ -1,0 +1,111 @@
+import NodeUtil from 'node:util';
+import isObject from 'lodash/isObject';
+import type { ModelStatic } from '../../model.js';
+import type { Sequelize } from '../../sequelize.js';
+import { isPlainObject, isString } from '../../utils/index.js';
+import { isModelStatic } from '../../utils/model-utils.js';
+import type { TableName } from './query-interface.js';
+import type { AbstractDialect } from './index.js';
+
+export type TableNameOrModel = TableName | ModelStatic;
+
+export interface QueryGeneratorOptions {
+  sequelize: Sequelize;
+  dialect: AbstractDialect;
+}
+
+/**
+ * Temporary class to ease the TypeScript migration
+ */
+export class AbstractQueryGeneratorTypeScript {
+
+  protected readonly dialect: AbstractDialect;
+  protected readonly sequelize: Sequelize;
+
+  constructor(options: QueryGeneratorOptions) {
+    if (!options.sequelize) {
+      throw new Error('QueryGenerator initialized without options.sequelize');
+    }
+
+    if (!options.dialect) {
+      throw new Error('QueryGenerator initialized without options.dialect');
+    }
+
+    this.sequelize = options.sequelize;
+    this.dialect = options.dialect;
+  }
+
+  get options() {
+    return this.sequelize.options;
+  }
+
+  extractTableDetails(tableNameOrModel: TableNameOrModel, options?: { schema?: string, delimiter?: string }) {
+    const tableNameObject = isModelStatic(tableNameOrModel) ? tableNameOrModel.getTableName()
+      : isString(tableNameOrModel) ? { tableName: tableNameOrModel }
+      : tableNameOrModel;
+
+    if (!isPlainObject(tableNameObject)) {
+      throw new Error(`Invalid input received, got ${NodeUtil.inspect(tableNameOrModel)}, expected a Model Class, a TableNameWithSchema object, or a table name string`);
+    }
+
+    return {
+      ...tableNameObject,
+      schema: options?.schema || tableNameObject.schema || this.options.schema || this.dialect.getDefaultSchema(),
+      delimiter: options?.delimiter || tableNameObject.delimiter || '.',
+    };
+  }
+
+  /**
+   * Quote table name with optional alias and schema attribution
+   *
+   * @param param table string or object
+   * @param alias alias name
+   */
+  quoteTable(param: TableNameOrModel, alias: boolean | string = false): string {
+    if (isModelStatic(param)) {
+      param = param.getTableName();
+    }
+
+    const tableName = this.extractTableDetails(param);
+
+    if (isObject(param) && ('as' in param || 'name' in param)) {
+      throw new Error('parameters "as" and "name" are not allowed in the first parameter of quoteTable, pass them as the second parameter.');
+    }
+
+    if (alias === true) {
+      alias = tableName.tableName;
+    }
+
+    let sql = '';
+
+    if (this.dialect.supports.schemas) {
+      if (tableName.schema && tableName.schema !== this.dialect.getDefaultSchema()) {
+        sql += `${this.quoteIdentifier(tableName.schema)}.`;
+      }
+
+      sql += this.quoteIdentifier(tableName.tableName);
+    } else {
+      if (tableName.schema && tableName.schema !== this.dialect.getDefaultSchema()) {
+        sql += tableName.schema + (tableName.delimiter || '.');
+      }
+
+      sql += tableName.tableName;
+    }
+
+    if (alias) {
+      sql += ` AS ${this.quoteIdentifier(alias)}`;
+    }
+
+    return sql;
+  }
+
+  /**
+   * Adds quotes to identifier
+   *
+   * @param _identifier
+   * @param _force
+   */
+  quoteIdentifier(_identifier: string, _force?: boolean): string {
+    throw new Error(`quoteIdentifier for Dialect "${this.dialect.name}" is not implemented`);
+  }
+}
--- a/src/dialects/abstract/query-generator.js
+++ b/src/dialects/abstract/query-generator.js
@@ -6,6 +6,7 @@ import { isModelStatic } from '../../utils/model-utils';
 import { injectReplacements } from '../../utils/sql';
 import { AbstractDataType } from './data-types';
 import { attributeTypeToSql, validateDataType } from './data-types-utils';
+import { AbstractQueryGeneratorTypeScript } from './query-generator-typescript';
 
 const util = require('util');
 const _ = require('lodash');
@@ -41,19 +42,9 @@ export const REMOVE_COLUMN_QUERY_SUPPORTABLE_OPTION = new Set(['ifExists']);
  *
  * @private
  */
-export class AbstractQueryGenerator {
+export class AbstractQueryGenerator extends AbstractQueryGeneratorTypeScript {
   constructor(options) {
-    if (!options.sequelize) {
-      throw new Error('QueryGenerator initialized without options.sequelize');
-    }
-
-    if (!options.dialect) {
-      throw new Error('QueryGenerator initialized without options.dialect');
-    }
-
-    this.sequelize = options.sequelize;
-    this.options = options.sequelize.options;
-    this.dialect = options.dialect;
+    super(options);
 
     // wrap quoteIdentifier with common logic
     this._initQuoteIdentifier();
@@ -83,18 +74,12 @@ export class AbstractQueryGenerator {
     throw new Error(`Databases are not supported in ${this.dialect.name}.`);
   }
 
-  extractTableDetails(tableName, options) {
-    options = options || {};
-    tableName = tableName || {};
-
-    return {
-      schema: tableName.schema || options.schema || this.options.schema || 'public',
-      tableName: _.isPlainObject(tableName) ? tableName.tableName : tableName,
-      delimiter: tableName.delimiter || options.delimiter || '.',
-    };
-  }
-
   // TODO: always return an object, instead of sometimes an object, sometimes a string
+  /**
+   * @deprecated use extractTableDetails instead
+   *
+   * @param {unknown} param
+   */
   addSchema(param) {
     if (!param._schema) {
       return param.tableName || param;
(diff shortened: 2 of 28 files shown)


src/dialects/abstract/query-generator-typescript.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    if (alias === true) {
+      alias = tableName.tableName;
+    }
+
+    let sql = '';
+
+    if (this.dialect.supports.schemas) {
+      if (tableName.schema && tableName.schema !== this.dialect.getDefaultSchema()) {
```
ephys, line 82: this is a change I'm unsure about.

Now that we always have a default schema per dialect, should we:

1. always add the schema to the query
2. or omit the schema if it's equal to the default schema, even if the user specified it manually?

test/unit/sql/select.test.js
```diff
@@ -706,11 +715,12 @@ describe(Support.getTestDialectTeaser('SQL'), () => {
               .replace(/\]/g, Support.sequelize.dialect.TICK_CHAR_RIGHT),
           ],
         }), {
-          default: 'SELECT \'* FROM [User]; DELETE FROM [User];SELECT [id]\' FROM [User];',
-          ibmi: 'SELECT \'* FROM "User"; DELETE FROM "User";SELECT "id"\' FROM "User"',
-          db2: 'SELECT \'* FROM "User"; DELETE FROM "User";SELECT "id"\' FROM "User";',
-          snowflake: 'SELECT \'* FROM "User"; DELETE FROM "User";SELECT "id"\' FROM "User";',
-          mssql: 'SELECT [* FROM User; DELETE FROM User;SELECT id] FROM [User];',
+          // TODO: the attribute should be escaped as an identifier, not a string
```
ephys, line 718: while this is inconsistent and needs to be fixed, I'm not going to do it in this PR, it's big enough and unrelated

src/dialects/abstract/query-generator-typescript.ts
ephys: One of the reasons for omitting the schema if it's the default schema is that some users use the `schema` option in `sequelize.sync` to generate the same set of tables in different schemas, then use the `searchPath` option to execute code in different schemas: https://github.com/sequelize/sequelize/issues/5252#issuecomment-172666833

We'll make specifying the default schema equivalent to not specifying a schema, so this use case can continue

src/dialects/abstract/query-generator.js
```diff
@@ -83,37 +74,6 @@ export class AbstractQueryGenerator {
...
-
-    return {
-      schema: tableName.schema || options.schema || this.options.schema || 'public',
-      tableName: _.isPlainObject(tableName) ? tableName.tableName : tableName,
-      delimiter: tableName.delimiter || options.delimiter || '.',
-    };
-  }
-
-  // TODO: always return an object, instead of sometimes an object, sometimes a string
-  addSchema(param) {
```
ephys, line 98: this has been permanently removed. It had a weird behavior and was mostly used when schemas were passed separately from the table name, which is something we're removing (from queryInterface/queryGenerator).

src/dialects/abstract/query-generator.js
```diff
@@ -712,15 +668,9 @@ export class AbstractQueryGenerator {
   }
 
   addConstraintQuery(tableName, options) {
-    if (typeof tableName === 'string') {
-      tableName = this.quoteIdentifiers(tableName);
```
ephys, line 716: something else that's been removed:

You used to be able to call `addConstraintQuery('schema.table')` and it'd split the string in two.

The thing is that `.` is a valid character in table names, so now functions like this one produce `"schema.table"` instead of `"schema"."table"`. If you want to specify a schema, pass `addConstraintQuery({ schema, table })` instead of a string

src/dialects/abstract/query-interface.d.ts
```diff
@@ -86,9 +86,8 @@ export interface TableNameWithSchema {
   tableName: string;
   schema?: string;
   delimiter?: string;
-  as?: string;
-  name?: string;
```
ephys, line 90: noise added by `addSchema`. I removed it altogether and it has not been a problem. Probably legacy?

test/integration/model/create.test.js
```diff
@@ -1077,16 +1077,15 @@ describe(Support.getTestDialectTeaser('Model'), () => {
...
         const User = this.sequelize.define('UserWithUniqueUsernameFunctionIndex', {
           username: DataTypes.STRING,
           email: { type: DataTypes.STRING, unique: true },
         });
 
         try {
           await User.sync({ force: true });
-          const tableName = User.getTableName();
-          await this.sequelize.query(`CREATE UNIQUE INDEX lower_case_username ON "${tableName}" ((lower("username")))`);
+          await this.sequelize.query(`CREATE UNIQUE INDEX lower_case_username ON ${this.sequelize.queryInterface.queryGenerator.quoteTable(User)} ((lower("username")))`, { logging: console.log });
```
WikiRik, line 1088: I don't think we need the logging here, right?

src/dialects/abstract/query-generator-typescript.ts
```diff
@@ -0,0 +1,129 @@
...
+    return sql;
+  }
+
+  /**
+   * Adds quotes to identifier
+   *
+   * @param identifier
+   * @param _force
+   */
+  quoteIdentifier(identifier: string, _force?: boolean) {
```
WikiRik, line 115: Do we have unit tests for these?

src/dialects/abstract/query-generator.js
```diff
@@ -1249,6 +1156,7 @@ export class AbstractQueryGenerator {
       name: tableName,
       quotedName: null,
       as: null,
+      quotedAa: null,
```
WikiRik, line 1159:
```suggestion
      quotedAs: null,
```

WikiRik (commented):
Thanks for this PR!

src/dialects/abstract/query-interface.d.ts
ephys: I think they were used for the alias (`FROM "table" AS "t"`). The only function that was using these properties was `quoteTable`, I made it throw if either property was specified and ran all tests - no issues at all.

test/unit/dialects/mssql/query-generator.test.js
```diff
@@ -69,21 +69,21 @@ if (current.dialect.name === 'mssql') {
 
     it('createTableQuery', function () {
       expectsql(this.queryGenerator.createTableQuery('myTable', { int: 'INTEGER' }, {}), {
-        mssql: 'IF OBJECT_ID(\'[myTable]\', \'U\') IS NULL CREATE TABLE [myTable] ([int] INTEGER);',
+        mssql: `IF OBJECT_ID(N'[myTable]', 'U') IS NULL CREATE TABLE [myTable] ([int] INTEGER);`,
```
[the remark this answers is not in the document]
ephys: Yes, `quotes` includes an `avoidEscape` option. Feel free to open a PR against our eslint config if you want (https://github.com/ephys/eslint-config) :)

There is a bunch of options I'd like to tweak in our config. I'll eventually take the time to update it 😶‍🌫️

src/dialects/abstract/query-interface.js
```diff
@@ -204,6 +204,7 @@ export class QueryInterface {
    *
    * @returns {Promise}
    */
+  // TODO: remove "schema" option from the option bag, it must be passed as part of "tableName" instead
```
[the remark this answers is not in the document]
ephys: It would certainly make all of this cleaner, but we'll need to stop somewhere so fine by me if it's delayed to v8

src/dialects/db2/index.ts
```diff
@@ -72,6 +72,11 @@ export class Db2Dialect extends AbstractDialect {
     return `BLOB(${this.queryGenerator.escape(buffer.toString())})`;
   }
 
+  getDefaultSchema(): string {
+    // TODO: what is the default schema in DB2?
```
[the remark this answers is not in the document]
ephys: I assume authorization ID means their username?

We could also decide to leave it blank if we're unsure. It shouldn't break anything

src/dialects/db2/query-generator.js
```diff
@@ -104,7 +104,7 @@ export class Db2QueryGenerator extends AbstractQueryGenerator {
           if (commentMatch && commentMatch.length > 2) {
             const commentText = commentMatch[2].replace(/COMMENT/, '').trim();
             commentStr += _.template(commentTemplate, this._templateSettings)({
-              table: this.quoteIdentifier(tableName),
+              table: this.quoteTable(tableName),
               comment: this.escape(commentText, undefined, { replacements: options.replacements }),
               column: this.quoteIdentifier(attr),
```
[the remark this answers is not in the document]
ephys: `quoteIdentifier` the pure function is just a utility function to ease the implementation of custom `QueryGenerator#quoteIdentifier`, but `QueryGenerator#quoteIdentifier` should be used when possible. I remove the unused import

src/dialects/mssql/query-generator.js
```diff
@@ -218,7 +218,7 @@ export class MsSqlQueryGenerator extends AbstractQueryGenerator {
     const quotedTableName = this.quoteTable(tableName);
 
     return Utils.joinSQLFragments([
-      `IF OBJECT_ID('${quotedTableName}', 'U') IS NULL`,
+      `IF OBJECT_ID(${this.escape(quotedTableName)}, 'U') IS NULL`,
       `CREATE TABLE ${quotedTableName} (${attributesClauseParts.join(', ')})`,
```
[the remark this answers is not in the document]
ephys: It's a function call, this will produce

```sql
-- tableName is "users"
IF OBJECT_ID(N'[users]', 'U')
```

src/utils/dialect.ts
```diff
@@ -46,14 +46,48 @@ export function toDefaultValue(value: unknown, dialect: AbstractDialect): unknow
...
+ * @param s
+ * @param tickChar
+ * @returns
+ */
 export function removeTicks(s: string, tickChar: string = TICK_CHAR): string {
   return s.replace(new RegExp(tickChar, 'g'), '');
 }
+
+export function quoteIdentifier(identifier: string, leftTick: string, rightTick: string): string {
+  // TODO: node 15+ : drop regexp, use replaceAll with a string instead.
```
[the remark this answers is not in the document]
ephys: I propose `[engine:node@>14]` because Sequelize 7 will continue to support node 14 after April.

test/integration/model/create.test.js
ephys: Anytime you see `console.log`, definitely assume I forgot to remove it 😅

src/dialects/abstract/query-generator-typescript.ts
ephys: A single one in `mssql/query-generator.test.js`. I've moved it to our common query-generator tests

src/dialects/abstract/query-generator.js
ephys: Done

src/utils/dialect.ts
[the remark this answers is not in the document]
ephys: Done

src/dialects/snowflake/index.ts
```diff
@@ -67,6 +67,11 @@ export class SnowflakeDialect extends AbstractDialect {
     return createUnspecifiedOrderedBindCollector();
   }
 
+  getDefaultSchema(): string {
+    // TODO: what is the default schema in Snowflake?
```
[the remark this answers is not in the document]
ephys: Done

src/dialects/abstract/query-interface.js
```diff
@@ -242,8 +242,6 @@ export class QueryInterface {
       table: tableName,
       context: 'createTable',
       withoutForeignKeyConstraints: options.withoutForeignKeyConstraints,
-      // schema override for multi-tenancy
-      schema: options.schema,
```
WikiRik: This has broken a postgres test, but I think we should update the test rather than leaving this in here

ephys: I think the test is fine but this implementation needs to be changed slightly

ephys: Surprisingly it's not failing when I run it locally 🤔

WikiRik:
SQLite unit tests fail since I included a test for the schema, but we've said that sqlite does not support schemas so we'll need to see how to deal with that

src/dialects/abstract/query-interface.js
ephys: wait why did I remove that it's super important until we've actually rewritten this API 🤦‍♀️

WikiRik (approved):

PR #15283, opened 2022-11-15
fix(postgres): invalidate connection after client-side timeout

harrykao:
Merge ff43e8d32d8069f4071df2ca57931c74f6556a2e from main:

The `query_timeout` feature of the `pg` package helps handle stuck TCP connections more quickly and gracefully by implementing a client-side timeout:

https://github.com/brianc/node-postgres/issues/1713

Sequelize started passing this dialect-specific option through to `pg` here:

https://github.com/sequelize/sequelize/pull/13258

I believe we also want to invalidate the connection when a client-side timeout occurs. We shouldn't try to reuse the stuck connection because...it's stuck.

This PR updates the error handling code so that the connection is invalidated if the error matches the one thrown from here:

https://github.com/brianc/node-postgres/blob/5538df6b446f4b4f921947b460fe38acb897e579/packages/pg/lib/client.js#L529

## Pull Request Checklist

## Todos

- [ ] 
- [ ] 
- [ ]

### DIFF
--- a/src/dialects/postgres/query.js
+++ b/src/dialects/postgres/query.js
@@ -85,6 +85,8 @@ class Query extends AbstractQuery {
         || /Unable to set non-blocking to true/i.test(error)
         || /SSL SYSCALL error: EOF detected/i.test(error)
         || /Local: Authentication failure/i.test(error)
+        // https://github.com/sequelize/sequelize/pull/15144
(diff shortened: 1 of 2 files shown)


harrykao:
> Assuming the other tests pass I think we should be good. Thanks for backporting!

No problem. Thanks for reviewing!

PR #15284, opened 2022-11-15
fix: support TS 4.9 and drop support TS 4.5

WikiRik:
## Pull Request Checklist

## Description Of Change

Now TS 4.9 is out we need to add that to our CI. We also drop support for TS 4.5 per our version policy since it has been out for a year.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ts-version: ["4.5", "4.6", "4.7", "4.8"]
+        ts-version: ["4.6", "4.7", "4.8", "4.9"]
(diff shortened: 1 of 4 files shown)


ephys (approved):
If tests are green, let's go

Do you plan on adding 4.9 to our v6 CI?

WikiRik:
> Do you plan on adding 4.9 to our v6 CI?

Yes, in #15059 probably. But I need to figure out what to do with the Node 10 tests there, maybe I just need to undo some TS migration things

ephys:
I don't see ts-node listed in the project's dev dependencies but ts-node 10 only supports Node 12+.

I think v6 should avoid using ts-node, it's supporting versions that are just too outdated.
Issues like this are part of why we should avoid keeping a major version alive for too long

PR #15286, opened 2022-11-16
meta: remove CI badge

fzn0x:
Lets remove CI badge, this is somehow a mistake.

CI badge was checking all the PRs from the CI action: https://github.com/sequelize/sequelize/actions/workflows/ci.yml so instead of success it will always show `failing`

The correct way is to replace it by https://github.com/sequelize/sequelize/actions/workflows/notify.yml but the badge was looking like this after https://github.com/sequelize/sequelize/workflows/Notify%20release%20channels/badge.svg so I think we can remove it instead

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@
 </p>
 
 [![npm version](https://badgen.net/npm/v/@sequelize/core)](https://www.npmjs.com/package/@sequelize/core)
-[![Build Status](https://github.com/sequelize/sequelize/workflows/CI/badge.svg)](https://github.com/sequelize/sequelize/actions?query=workflow%3ACI)
 [![npm downloads](https://badgen.net/npm/dm/@sequelize/core)](https://www.npmjs.com/package/@sequelize/core)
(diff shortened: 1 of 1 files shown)


fzn0x:
More badges to delete is welcome

fzn0x:
We can also rename `notify release channel` to `Release CI`, what do you think?

WikiRik:
The release notifier does not really do anything with the code. It is triggered when a release is happening, but only tweets and informs OpenCollective. We can also do that manually without changing our source code.

I think that we should think about running our CI once a week on `main` so that the badge will pass. It's useful information to have on your README but it's failing now because it only runs on PRs (which have to be up to date with `main` making CI on `main` basically redundant)

ephys (approved):
I'm OK with this change. The CI badge is a bit pointless

> I think that we should think about running our CI once a week on main so that the badge will pass.

Since nothing can change without first going through a PR, running the CI on main once a week would never catch any issue. If it's just to have the badge be green, I say we just remove the badge

WikiRik (approved):
I'm fine either way. We can always place this badge back when we do a more extensive test suite that runs on `main`. Let's merge this

